PR #988, opened 2020-07-11
Configurable cluster domain

spoonofpower:
**Issue number:**
N/A

**Description of changes:**
This change allows users to set the value for the kubelet's clusterDomain setting. By default the clusterDomain is set to `cluster.local`, but users can set it to any valid domain. A description of the clusterDomain setting can be found [here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L288-L295).

We have multiple EKS clusters and each uses a different clusterDomain. We'd prefer not to have all our EKS clusters using `cluster.local`.

**Testing done:**
Tested using the aws-k8s-1.15 variant on our EKS clusters. If no value is set for `settings.kubernetes.cluster-domain` in the user data, the kubelet will be configured to use `cluster.local` for the clusterDomain. This is the current behaviour. By default this PR doesn't change the current behaviour. If there is a value for `settings.kubernetes.cluster-domain`, the kubelet will be configure to use that value for the clusterDomain.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,7 @@ The following settings can be optionally set to customize the node labels and ta
 The following settings are set for you automatically by [pluto](sources/api/) based on runtime instance information, but you can override them if you know what you're doing!
 * `settings.kubernetes.max-pods`: The maximum number of pods that can be scheduled on this node (limited by number of available IPv4 addresses)
 * `settings.kubernetes.cluster-dns-ip`: The CIDR block of the primary network interface.
+* `settings.kubernetes.cluster-domain`: The cluster domain is set to `cluster.local` by default.
 * `settings.kubernetes.node-ip`: The IPv4 address of this node.
 * `settings.kubernetes.pod-infra-container-image`: The URI of the "pause" container.
 
--- a/README.md
+++ b/README.md
@@ -347,10 +347,12 @@ The following settings can be optionally set to customize the node labels and ta
     special = "true:NoSchedule"
     ```
 
+The following settings are optional and allow you to further configure your cluster.
+* `settings.kubernetes.cluster-domain`: The DNS domain for this cluster, allowing all Kubernetes-run containers to search this domain before the host's search domains.  Defaults to `cluster.local`.
+
 The following settings are set for you automatically by [pluto](sources/api/) based on runtime instance information, but you can override them if you know what you're doing!
 * `settings.kubernetes.max-pods`: The maximum number of pods that can be scheduled on this node (limited by number of available IPv4 addresses)
 * `settings.kubernetes.cluster-dns-ip`: The CIDR block of the primary network interface.
-* `settings.kubernetes.cluster-domain`: The cluster domain is set to `cluster.local` by default.
 * `settings.kubernetes.node-ip`: The IPv4 address of this node.
 * `settings.kubernetes.pod-infra-container-image`: The URI of the "pause" container.
 
--- a/sources/models/defaults.toml
+++ b/sources/models/defaults.toml
@@ -92,8 +92,3 @@ template-path = "/usr/share/templates/chrony-conf"
 
 [metadata.settings.ntp]
 affected-services = ["chronyd"]
-
-# Kubernetes
-
-[settings.kubernetes]
-cluster-domain = "cluster.local"
--- a/sources/models/defaults.toml
+++ b/sources/models/defaults.toml
@@ -92,3 +92,8 @@ template-path = "/usr/share/templates/chrony-conf"
 
 [metadata.settings.ntp]
 affected-services = ["chronyd"]
+
+# Kubernetes
+
+[settings.kubernetes]
+cluster-domain = "cluster.local"
--- a/sources/models/src/modeled_types/shared.rs
+++ b/sources/models/src/modeled_types/shared.rs
@@ -9,6 +9,7 @@ use std::convert::TryFrom;
 use std::fmt;
 use std::ops::Deref;
 use std::str::FromStr;
+use url::Host;
 
 /// ValidBase64 can only be created by deserializing from valid base64 text.  It stores the
 /// original text, not the decoded form.  Its purpose is input validation, namely being used as a
@@ -365,25 +366,36 @@ impl TryFrom<&str> for DNSDomain {
     type Error = error::Error;
 
     fn try_from(input: &str) -> Result<Self, Self::Error> {
-        ensure!(input.len() <= 253, error::InvalidDomainName { input });
-
-        let labels = input.split(".");
-        for (i, label) in labels.enumerate() {
-            ensure!(i < 127, error::InvalidDomainName { input });
-            ensure!(label.len() <= 63, error::InvalidDomainName { input });
-            let last = label.len() - 1;
-            for (j, c) in label.chars().enumerate() {
-                ensure!(
-                    (j != 0 && j != last && (c.is_alphanumeric() || c == '-'))
-                        || c.is_alphanumeric(),
-                    error::InvalidDomainName { input }
-                );
+        if input.starts_with('.') {
+            return error::InvalidDomainName {
+                input: input,
+                msg: "must not start with '.'",
             }
+            .fail();
         }
 
-        Ok(Self {
-            inner: input.to_string(),
-        })
+        match Host::parse(input) {
+            Err(e) => error::InvalidDomainName {
+                input: input,
+                msg: e.to_string(),
+            }
+            .fail(),
+            Ok(h) => match h {
+                Host::Ipv4(_) => error::InvalidDomainName {
+                    input: input,
+                    msg: "IP address is not a valid domain name",
+                }
+                .fail(),
+                Host::Ipv6(_) => error::InvalidDomainName {
+                    input: input,
+                    msg: "IP address is not a valid domain name",
+                }
+                .fail(),
+                Host::Domain(_) => Ok(Self {
+                    inner: input.to_string(),
+                }),
+            },
+        }
     }
 }
 
@@ -396,15 +408,7 @@ mod test_dns_domain {
 
     #[test]
     fn valid_dns_domain() {
-        for ok in &[
-            "cluster.local",
-            "dev.eks",
-            "stage.eks",
-            "prod.eks",
-            "1-good.local",
-            &"a".repeat(63),
-            &format!("a{}", ".a".repeat(126)),
-        ] {
+        for ok in &["cluster.local", "dev.eks", "stage.eks", "prod.eks"] {
             assert!(DNSDomain::try_from(*ok).is_ok());
         }
     }
@@ -412,12 +416,10 @@ mod test_dns_domain {
     #[test]
     fn invalid_dns_domain() {
         for err in &[
-            "-bad.com",
-            "bad-.com",
-            "-bad-.com",
-            "b_ad.com",
-            &"a".repeat(64),
-            &format!("a{}", ".a".repeat(127)),
+            "foo/com",
+            ".a",
+            "123.123.123.123",
+            "[2001:db8::ff00:42:8329]",
         ] {
             assert!(DNSDomain::try_from(*err).is_err());
         }
--- a/sources/models/src/modeled_types/shared.rs
+++ b/sources/models/src/modeled_types/shared.rs
@@ -365,36 +365,31 @@ pub struct DNSDomain {
 impl TryFrom<&str> for DNSDomain {
     type Error = error::Error;
 
-    fn try_from(input: &str) -> Result<Self, Self::Error> {
-        if input.starts_with('.') {
-            return error::InvalidDomainName {
+    fn try_from(input: &str) -> Result<Self, error::Error> {
+        ensure!(
+            !input.starts_with('.'),
+            error::InvalidDomainName {
                 input: input,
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -350,6 +350,7 @@ The following settings can be optionally set to customize the node labels and ta
 The following settings are set for you automatically by [pluto](sources/api/) based on runtime instance information, but you can override them if you know what you're doing!
 * `settings.kubernetes.max-pods`: The maximum number of pods that can be scheduled on this node (limited by number of available IPv4 addresses)
 * `settings.kubernetes.cluster-dns-ip`: The CIDR block of the primary network interface.
+* `settings.kubernetes.cluster-domain`: The cluster domain is set to `cluster.local` by default.
```
tjkirch: We currently have three subsections related to Kubernetes settings:
* Required settings  (this isn't required)
* Optional settings, but specifically narrowed down to node labels/taints
* This section, which is for for settings automatically set by `pluto` based on runtime information  (this isn't discovered at runtime)

I think instead of being here, we should make another section about optional settings, above the pluto section, since most people won't override the settings set by `pluto`. Perhaps like this:
```
The following settings are optional and allow you to further configure your cluster:
```

I also think it would be helpful to have a description of this setting before its default value, perhaps like this:
```
```

sources/models/defaults.toml
```diff
@@ -92,3 +92,8 @@ template-path = "/usr/share/templates/chrony-conf"
 
 [metadata.settings.ntp]
 affected-services = ["chronyd"]
+
+# Kubernetes
+
+[settings.kubernetes]
+cluster-domain = "cluster.local"
```
tjkirch: Kubernetes settings need to go into the Kubernetes override files, rather than this file for all variants, because they don't apply to all variants.  (If this is here and you try a non-Kubernetes variant (currently just `aws-dev`) you'll see the API fail to deserialize the data store.)

Currently that means `src/aws-k8s-1.15/override-defaults.toml`, since 1.16 and 1.17 models are symlinked.  Would you mind moving this there?

sources/models/src/modeled_types/shared.rs
```diff
@@ -349,3 +349,77 @@ mod test_version {
...
+/// config value.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct DNSDomain {
+    inner: String,
+}
+
+impl TryFrom<&str> for DNSDomain {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
```
tjkirch: I'm wondering if it would be simpler / more resilient to use `url::Host::parse` on the input string and verify the result is the `Host::Domain` variant (or something similar with another library, but we already have `url`) rather than parsing and implementing domain name checking rules ourselves.  Then, if there are further restrictions that Kubernetes has, we could check those afterward.  Or is this already pretty specific to what Kubernetes will accept?

(My comments on the implementation wouldn't apply if we change this approach)

sources/models/src/modeled_types/shared.rs
```diff
@@ -349,3 +349,77 @@ mod test_version {
...
+        ensure!(input.len() <= 253, error::InvalidDomainName { input });
```
tjkirch: I agree that a single error variant can keep things tidy, but there are a few different failure cases implemented here; perhaps a `msg: String` field in the variant that allows us to add some detail here, and include that in the error `display`?

sources/models/src/modeled_types/shared.rs
```diff
@@ -349,3 +349,77 @@ mod test_version {
...
+
+        let labels = input.split(".");
+        for (i, label) in labels.enumerate() {
+            ensure!(i < 127, error::InvalidDomainName { input });
+            ensure!(label.len() <= 63, error::InvalidDomainName { input });
+            let last = label.len() - 1;
+            for (j, c) in label.chars().enumerate() {
+                ensure!(
+                    (j != 0 && j != last && (c.is_alphanumeric() || c == '-'))
```
tjkirch: I'm having trouble understanding this bit, how does this work?  It seems like `j` will be 0 in the first iteration, since `enumerate` starts at 0, and that would short-circuit the `j != 0` line to be false, and the `ensure` would fail... same when we hit `j == last`.  What am I missing?  :)

I was naively thinking we'd need separate checks, like:
```rust
if j == 0 || j == last {
    ensure!(c.is_alphanumeric(), error::bla);
} else {
    ensure!(c.is_alphanumeric() || c == '-', error::bla);
}
```

Maybe something like that would be more obvious to readers in any case?

sources/models/src/modeled_types/shared.rs
```diff
@@ -349,3 +349,77 @@ mod test_version {
...
+}
+
+impl TryFrom<&str> for DNSDomain {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        ensure!(input.len() <= 253, error::InvalidDomainName { input });
+
+        let labels = input.split(".");
+        for (i, label) in labels.enumerate() {
```
tjkirch: nit: Can we change `i` and `j` to `label_index` and `char_index` or something similar?  I know, bucking decades of research, but I think it's clearer and faster if you don't have to look back to where the letter was defined...

sources/models/src/modeled_types/shared.rs
```diff
@@ -349,3 +349,77 @@ mod test_version {
...
+            &"a".repeat(63),
+            &format!("a{}", ".a".repeat(126)),
+        ] {
+            assert!(DNSDomain::try_from(*ok).is_ok());
+        }
+    }
+
+    #[test]
+    fn invalid_dns_domain() {
+        for err in &[
```
tjkirch, line 413: `.a` is another interesting one that should fail; `a.` is interesting but might pass, depending on expectations :D

tjkirch (commented):
Thanks for your contribution!  We'll also test the other variants (and respond here) since this is something that affects runtime behavior.

tjkirch:
@etungsten was kind enough to verify the functionality on all of the Kubernetes variants. :)

README.md
spoonofpower: Yes. Makes sense.

sources/models/defaults.toml
spoonofpower: Ah. I was only thinking about the kubernetes case. I'll move the default setting.

sources/models/src/modeled_types/shared.rs
spoonofpower: The kubelet doesn't do any validation on the clusterDomain setting. It uses whatever string is supplied. I wanted to add some basic sanity checking.

Using url::Host:parse to validate the clusterDomain makes sense to me. I'll make the change. I'm new to rust and wasn't sure what the best practice was for validating a DNS domain.

sources/models/src/modeled_types/shared.rs
spoonofpower: I'll make sure to add a msg field.

sources/models/src/modeled_types/shared.rs
spoonofpower: Yup. Not very clear.  That code will be removed with the switch to url::Host::parse.

For what it's worth... There were two lines in that || condition. If that first line that you mentioned was false, then the second would have to be true.

sources/models/src/modeled_types/shared.rs
spoonofpower: It doesn't look like url::Host:Parse handles the `.a` case. I'll make sure to add a check for it.

spoonofpower:
@tjkirch Thanks for the review. I think I addressed all your concerns. Let me know if there's more I can do.

sources/models/src/modeled_types/shared.rs
```diff
@@ -396,28 +408,18 @@ mod test_dns_domain {
...
     #[test]
     fn valid_dns_domain() {
-        for ok in &[
-            "cluster.local",
-            "dev.eks",
-            "stage.eks",
-            "prod.eks",
-            "1-good.local",
-            &"a".repeat(63),
-            &format!("a{}", ".a".repeat(126)),
```
tjkirch: Do these no longer pass as valid domains?  It seems like "1-good.local" and "a.a.a" should, though maybe not "aaa"...

sources/models/src/modeled_types/shared.rs
```diff
@@ -365,25 +366,36 @@ impl TryFrom<&str> for DNSDomain {
...
-                        || c.is_alphanumeric(),
-                    error::InvalidDomainName { input }
-                );
+        if input.starts_with('.') {
+            return error::InvalidDomainName {
+                input: input,
+                msg: "must not start with '.'",
             }
+            .fail();
         }
```
tjkirch:
```suggestion
        ensure!(!input.starts_with('.'), error::InvalidDomainName {
            input: input,
            msg: "must not start with '.'",
        });
```

sources/models/src/modeled_types/shared.rs
```diff
@@ -365,25 +366,36 @@ impl TryFrom<&str> for DNSDomain {
...
+                Host::Ipv6(_) => error::InvalidDomainName {
+                    input: input,
+                    msg: "IP address is not a valid domain name",
+                }
+                .fail(),
+                Host::Domain(_) => Ok(Self {
+                    inner: input.to_string(),
+                }),
+            },
+        }
```
tjkirch:
```suggestion
        let host = Host::parse(input).or_else(|e| {
            error::InvalidDomainName {
                input: input,
                msg: e.to_string(),
            }
            .fail()
        })?;
        match host {
            Host::Ipv4(_) | Host::Ipv6(_) => error::InvalidDomainName {
                input: input,
                msg: "IP address is not a valid domain name",
            }
            .fail(),
            Host::Domain(_) => Ok(Self {
                inner: input.to_string(),
            }),
        }
```

tjkirch (approved):
Looks good to me, just a couple simplification suggestions.  I'll poke someone else to review as well.

sources/models/src/modeled_types/shared.rs
spoonofpower: Yes. Those will pass. Obviously invalid ones like `&format!("a{}", ".a".repeat(500))`, `&"a".repeat(500)`, and `_b_a_d_.com` will also pass. The `url::Host::parse()` checks are not as strict as what I had earlier. I think that's fine though. In this case I think it's better to have a standard library do the checks instead of custom code. Especially since the kubelet doesn't even do any checks on this. Since just about everything passes, I removed some of the test cases and used it more as documentation for what likely uses would look like.

sources/models/src/modeled_types/shared.rs
spoonofpower: I'll make the change.

etungsten (approved):
Nice!

sources/models/src/modeled_types/shared.rs
spoonofpower: Thanks. Like I mentioned, I'm new to rust. This helps a lot. I don't have any sense of the best way to do things yet.

tjkirch: No problem!  This is a little more idiomatic, I think.  I hope you're enjoying Rust - it's been great for Bottlerocket.

We can merge the PR as soon as you're happy with these last changes.

spoonofpower: I hate to bug you with language questions. But I can't seem to find a good way to deal with this:

```rust
        let host = Host::parse(input).map_err(|e| error::InvalidDomainName {
            input: input,
            msg: e.to_string(),
        })?;
```

The `?` needs to use the From trait to convert from a `modeled_types::error::InvalidDomainName` to a `modeled_types::error::Error`, but that trait isn't implemented. What's the best way to handle that?

tjkirch: No problem, I'm sorry I gave an incomplete suggestion!  I've updated it above, and it compiles now.

The problem you ran into is related to how `snafu`, the error handling library we use, works.  The details are [here](https://docs.rs/snafu/0.6.8/snafu/guide/the_macro/index.html), but the short version is: snafu provides a nice interface to library developers using a procedural macro that adds to the code at compile time.  The library developer can make a nice enum of their error cases, and they don't have to define everything about error handling that's actually needed.  snafu takes that and generates structs for each of the error variants, which are what's actually used behind the scenes, and makes error construction easier.

My mistake was suggesting `map_err` and creating the struct form (`modeled_types::error::InvalidDomain`) inside, when the return type of the function needs the enum form (`modeled_types::error::Error`).  Rust helpfully tried to convert from one to the other with `From` as part of using `?` for an early return, but it doesn't know how.  What I did in my revised suggestion is to convert from the struct form to the enum form with `.fail()` so we can return it.  (This also means I needed to use `or_else` instead of `map_err`; they're identical except that `or_else` takes a full `Result` and not just the `Err` part inside a result, which we need because `fail()` returns a full `Result`, for reasons related to ease of use in more common scenarios.)

Hope this helps, and let me know if you have any other trouble.

spoonofpower: Thanks for the explanation! This has been very helpful for me.

spoonofpower:
I'm happy with the changes, and all my testing looks good. Let me know if you need anything else changed.

PR #992, opened 2020-07-15
updog: change heuristics for determining update availability

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
Author: Erikson Tung <etung@amazon.com>
Date:   Tue Jul 14 15:58:23 2020 -0700

    updog: remove jitters, calculate wave position to determine availability
    
    This removes jittering in updog and instead use the seed to find a
    target time position within the wave to determine whether an update is
    available
```

This also removes the `--timestamp` option since it's no longer necessary without updog returning the jittered time.

As result of this change, updates will only be visible if they are available (host is past their defined update wave). To see all updates regardless of update waves, either `--ignore-waves` needs to be specified or `settings.updates.ignore-waves` needs to be set to true.

**Testing done:**
`updog` unit tests pass, newly created unit test for `update_metadata` passes.

Testing in an instance:

Confirmed that the seed value places the host in the last wave which starts on 7/19
```
bash-5.0# cat /etc/updog.toml 
metadata_base_url = "https://updates.bottlerocket.aws/2020-02-02/aws-k8s-1.15/x86_64/"
targets_base_url = "https://updates.bottlerocket.aws/targets/"
seed = 887
version_lock = "latest"
ignore_waves = false

```

Check update without ignoring waves, v0.4.1 does not show up since our wave hasn't started:
```
bash-5.0# updog check-update                                           aws-k8s-1.15 0.4.0
bash-5.0# updog check-update -a --json
[
  {
    "variant": "aws-k8s-1.15",
    "arch": "x86_64",
    "version": "0.4.0",
    "max_version": "0.4.1",
    "waves": {
      "0": "2020-06-29T20:00:00Z",
      "20": "2020-06-29T23:00:00Z",
      "102": "2020-06-30T19:00:00Z",
      "204": "2020-07-02T19:00:00Z",
      "512": "2020-07-05T19:00:00Z"
    },
    "images": {
      "boot": "bottlerocket-aws-k8s-1.15-x86_64-0.4.0-7303622-boot.ext4.lz4",
      "root": "bottlerocket-aws-k8s-1.15-x86_64-0.4.0-7303622-root.ext4.lz4",
      "hash": "bottlerocket-aws-k8s-1.15-x86_64-0.4.0-7303622-root.verity.lz4"
    }
  },
  {
    "variant": "aws-k8s-1.15",
    "arch": "x86_64",
    "version": "0.3.4",
    "max_version": "0.4.1",
    "waves": {
[...]

PR #993, opened 2020-07-15
updog: change method of determining update availability

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
Date:   Tue Jul 14 15:58:23 2020 -0700

    updog: remove jitters, calculate wave position to determine availability
    
    This removes jittering in updog and instead use the seed to find a
    target time position within the wave to determine whether an update is
    available.
    
    This also removes the --timestamp option since it's no longer necessary
    without updog returning the jittered time.
    
    As a result of this change, updates will only be visible if they are
    available (host is past their defined update wave). To see all updates
    regardless of update waves, either --ignore-waves needs to be specified
    or settings.updates.ignore-waves needs to be set to true.
```
```
Date:   Wed Jul 15 17:54:46 2020 -0700

    waves: define a wave duration for the "last" wave
    
    To avoid having all the hosts in the last wave all updating at once when
    the last wave starts, we define an intermediary wave before the actual
    last wave to define its wave duration so updog can evenly spread the
    hosts throughout the "last" wave.
```

**Testing done:**
`updog` unit tests pass, newly created unit test for `update_metadata` passes.

Testing in an instance:

Confirmed that the seed value places the host in the last wave for v0.4.1 which starts on 7/19
```
bash-5.0# cat /etc/updog.toml 
metadata_base_url = "https://updates.bottlerocket.aws/2020-02-02/aws-k8s-1.15/x86_64/"
targets_base_url = "https://updates.bottlerocket.aws/targets/"
seed = 887
version_lock = "latest"
ignore_waves = false

```

Check update without ignoring waves, v0.4.1 does not show up since our wave hasn't started:
```
bash-5.0# updog check-update                                           aws-k8s-1.15 0.4.0
bash-5.0# updog check-update -a --json
[
  {
    "variant": "aws-k8s-1.15",
    "arch": "x86_64",
    "version": "0.4.0",
    "max_version": "0.4.1",
    "waves": {
      "0": "2020-06-29T20:00:00Z",
      "20": "2020-06-29T23:00:00Z",
      "102": "2020-06-30T19:00:00Z",
      "204": "2020-07-02T19:00:00Z",
      "512": "2020-07-05T19:00:00Z"
    },
    "images": {
      "boot": "bottlerocket-aws-k8s-1.15-x86_64-0.4.0-7303622-boot.ext4.lz4",
      "root": "bottlerocket-aws-k8s-1.15-x86_64-0.4.0-7303622-root.ext4.lz4",
      "hash": "bottlerocket-aws-k8s-1.15-x86_64-0.4.0-7303622-root.verity.lz4"
    }
  },
  {
    "variant": "aws-k8s-1.15",
    "arch": "x86_64",
    "version": "0.3.4",
    "max_version": "0.4.1",
    "waves": {
      "0": "2020-05-28T18:00:00Z",
      "20": "2020-05-28T21:00:00Z",
      "102": "2020-05-29T17:00:00Z",
      "204": "2020-05-31T17:00:00Z",
      "512": "2020-06-03T17:00:00Z"
    },
    "images": {
      "boot": "bottlerocket-aws-k8s-1.15-x86_64-0.3.4-85d09a8-boot.ext4.lz4",
      "root": "bottlerocket-aws-k8s-1.15-x86_64-0.3.4-85d09a8-root.ext4.lz4",
      "hash": "bottlerocket-aws-k8s-1.15-x86_64-0.3.4-85d09a8-root.verity.lz4"
    }
  }, 
 ...
]
```

Calling `updog update` gives me 0.4.0:
```
bash-5.0# updog update
Starting update to 0.4.0
^C
```

With `--ignore-waves` set, 0.4.1 becomes visible

```
bash-5.0# updog check-update --ignore-waves
aws-k8s-1.15 0.4.1
bash-5.0# updog check-update -a --ignore-waves       
aws-k8s-1.15 0.4.1
aws-k8s-1.15 0.4.0
aws-k8s-1.15 0.3.4
aws-k8s-1.15 0.3.3
aws-k8s-1.15 0.3.2
aws-k8s-1.15 0.3.1
aws-k8s-1.15 0.3.0
```

Calling `updog update --ignore-waves` gives me v0.4.1:
```
bash-5.0# updog update --ignore-waves         
Starting update to 0.4.1

``` 

I manually edited `/etc/updog.toml` to lower the seed value to fit a wave that has already passed:
```
[ec2-user@ip-192-168-21-225 ~]$ sudo vi /.bottlerocket/rootfs/etc/updog.toml 
[ec2-user@ip-192-168-21-225 ~]$ sudo sheltie
bash-5.0# cat /etc/updog.toml 
metadata_base_url = "https://updates.bottlerocket.aws/2020-02-02/aws-k8s-1.15/x86_64/"
targets_base_url = "https://updates.bottlerocket.aws/targets/"
seed = 200 
version_lock = "latest"
ignore_waves = false
```

And now v0.4.1 shows up as an available update even without the `--ignore-waves` flag.
```
bash-5.0# updog check-update --json
{
  "variant": "aws-k8s-1.15",
  "arch": "x86_64",
  "version": "0.4.1",
  "max_version": "0.4.1",
  "waves": {
    "0": "2020-07-13T19:00:00Z",
    "20": "2020-07-13T22:00:00Z",
    "102": "2020-07-14T18:00:00Z",
    "204": "2020-07-16T18:00:00Z",
    "512": "2020-07-19T18:00:00Z"
  },
  "images": {
    "boot": "bottlerocket-aws-k8s-1.15-x86_64-0.4.1-a29a1450-boot.ext4.lz4",
    "root": "bottlerocket-aws-k8s-1.15-x86_64-0.4.1-a29a1450-root.ext4.lz4",
    "hash": "bottlerocket-aws-k8s-1.15-x86_64-0.4.1-a29a1450-root.verity.lz4"
  }
}
bash-5.0# updog update
Starting update to 0.4.1
```

**Terms of contribution:**

### DIFF
--- a/sources/updater/README.md
+++ b/sources/updater/README.md
@@ -41,8 +41,8 @@ Updog will ensure that appropriate migration files are available to safely trans
 
 ### Update wave
 Updates may include "wave" information which provides a way for updates to be scheduled over time for groups of Bottlerocket hosts.
-Updog will find the update wave the host belongs to and jitter its update time within that range.
-If the calculated time has not passed yet, Updog returns the update timestamp to the caller so it can be called again at the correct time.
+Updog will find the update wave the host belongs to and calculate its time position within the wave based on its `settings.updates.seed` value.
+If the calculated time has not passed, Updog will not report an update as being available.
 
 Assuming all the requirements are met, Updog requests the update images from the TUF repository and writes them to the "inactive" partition.
 
--- a/sources/updater/update_metadata/src/lib.rs
+++ b/sources/updater/update_metadata/src/lib.rs
@@ -5,9 +5,8 @@ pub mod error;
 mod se;
 
 use crate::error::Result;
-use chrono::{DateTime, Duration, Utc};
+use chrono::{DateTime, Utc};
 use parse_datetime::parse_offset;
-use rand::{thread_rng, Rng};
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use snafu::{ensure, OptionExt, ResultExt};
@@ -23,30 +22,34 @@ pub const MAX_SEED: u32 = 2048;
 #[derive(Debug, PartialEq, Eq)]
 pub enum Wave {
     Initial {
-        end: DateTime<Utc>,
+        end_time: DateTime<Utc>,
+        end_seed: u32,
     },
     General {
-        start: DateTime<Utc>,
-        end: DateTime<Utc>,
+        start_time: DateTime<Utc>,
+        end_time: DateTime<Utc>,
+        start_seed: u32,
+        end_seed: u32,
     },
     Last {
-        start: DateTime<Utc>,
+        start_time: DateTime<Utc>,
+        start_seed: u32,
     },
 }
 
 impl Wave {
-    pub fn has_started(&self) -> bool {
+    pub fn has_started(&self, time: DateTime<Utc>) -> bool {
         match self {
             Self::Initial { .. } => true,
-            Self::General { start, .. } | Self::Last { start } => *start <= Utc::now(),
+            Self::General { start_time, .. } | Self::Last { start_time, .. } => *start_time <= time,
         }
     }
 
-    pub fn has_passed(&self) -> bool {
+    pub fn has_passed(&self, time: DateTime<Utc>) -> bool {
         match self {
-            Self::Initial { end } => *end <= Utc::now(),
-            Self::General { end, .. } => *end <= Utc::now(),
-            Self::Last { start } => *start <= Utc::now(),
+            Self::Initial { end_time, .. } => *end_time <= time,
+            Self::General { end_time, .. } => *end_time <= time,
+            Self::Last { start_time, .. } => *start_time <= time,
         }
     }
 }
@@ -271,58 +274,92 @@ impl Manifest {
 impl Update {
     /// Returns the update wave that Updog belongs to, based on the seed value.
     /// Depending on the waves described in the update, the possible results are
-    /// - Some wave described by a start and end time.
-    /// - The "0th" wave, which has an "end" time but no specified start time.
-    /// - The last wave, which has a start time but no specified end time.
+    /// - Some wave described by a start and end time, and the starting seed and ending seed.
+    /// - The "0th" wave, which has an "end" time but no specified start time, and the ending seed.
+    /// - The last wave, which has a start time but no specified end time, and the starting seed.
     /// - Nothing, if no waves are configured.
+    #[must_use]
     pub fn update_wave(&self, seed: u32) -> Option<Wave> {
-        let start = self
+        let start_wave = self
             .waves
             .range((Included(0), Excluded(seed)))
-            .last()
-            .map(|(_, wave)| *wave);
-        let end = self
+            .map(|(k, v)| (*k, *v))
+            .last();
+        let end_wave = self
             .waves
             .range((Included(seed), Included(MAX_SEED)))
-            .next()
-            .map(|(_, wave)| *wave);
+            .map(|(k, v)| (*k, *v))
+            .next();
 
-        match (start, end) {
-            (None, Some(end)) => Some(Wave::Initial { end }),
-            (Some(start), Some(end)) => Some(Wave::General { start, end }),
-            (Some(start), None) => Some(Wave::Last { start }),
+        match (start_wave, end_wave) {
+            // Note that the key for each wave entry is the starting seed for that wave, the value is the DateTime
+            (None, Some((end_seed, end_time))) => Some(Wave::Initial { end_seed, end_time }),
+            (Some((start_seed, start_time)), Some((end_seed, end_time))) => Some(Wave::General {
+                start_time,
+                end_time,
+                start_seed,
+                end_seed,
+            }),
+            (Some((start_seed, start_time)), None) => Some(Wave::Last {
+                start_time,
+                start_seed,
+            }),
             _ => None,
         }
     }
 
-    pub fn update_ready(&self, seed: u32) -> bool {
-        // Has this client's wave started
+    /// Returns whether the update is available. An update is said to be 'ready/available' if the wave
+    /// this host belongs to has fully passed, or if the host's position in the wave has passed, or
+    /// if there are no waves.
+    /// The position of the host within the wave is determined by the seed value.
+    #[must_use]
+    pub fn update_ready(&self, seed: u32, time: DateTime<Utc>) -> bool {
+        // If this host is part of some update wave
         if let Some(wave) = self.update_wave(seed) {
-            return wave.has_started();
-        }
-
-        // Or there are no waves
-        true
-    }
-
-    pub fn jitter(&self, seed: u32) -> Option<DateTime<Utc>> {
-        if let Some(wave) = self.update_wave(seed) {
-            if wave.has_passed() {
-                return None;
+            // If the wave has passed, the update is available (this includes passing the last wave start time)
+            if wave.has_passed(time) {
+                return true;
+            } else if !wave.has_started(time) {
+                return false;
             }
-            let bounds = match self.update_wave(seed) {
-                Some(Wave::Initial { end }) => Some((Utc::now(), end)),
-                Some(Wave::General { start, end }) => Some((start, end)),
-                Some(Wave::Last { start: _ }) | None => None,
+            let bound = match wave {
+                // Hosts should not wind up in the special "initial" wave with no start time, but if they do,
+                // we consider the update as being available immediately.
+                Wave::Initial { .. } => None,
+                Wave::General {
+                    start_time,
+                    end_time,
+                    start_seed,
+                    end_seed,
+                } => Some((start_time, Some(end_time), start_seed, end_seed)),
+                // Last wave has no end time nor end seed; Let end seed be `MAX_SEED` since all the
+                // remaining hosts are in this last wave
+                Wave::Last {
+                    start_time,
+                    start_seed,
+                } => Some((start_time, None, start_seed, MAX_SEED)),
             };
-            if let Some((start, end)) = bounds {
-                let mut rng = thread_rng();
-                if let Some(range) = end.timestamp().checked_sub(start.timestamp()) {
-                    return Some(start + Duration::seconds(rng.gen_range(1, range)));
+            if let Some((start_time, maybe_end_time, start_seed, end_seed)) = bound {
+                if let Some(end_time) = maybe_end_time {
+                    // This host is not part of last wave
+                    // Determine the duration of this host's wave
+                    let wave_duration = end_time - start_time;
+                    let num_seeds_allocated_to_wave = (end_seed - start_seed) as i32;
+                    if num_seeds_allocated_to_wave == 0 {
+                        // Empty wave, no host should have been allocated to it
+                        return true;
+                    }
+                    let time_per_seed = wave_duration / num_seeds_allocated_to_wave;
+                    // Derive the target time position within the wave given the host's seed.
+                    let target_time = start_time + (time_per_seed * (seed as i32));
+                    // If the current time is past the target time position in the wave, the update is
+                    // marked available
+                    return time >= target_time;
                 }
             }
         }
-        None
+        // There are no waves, so we consider the update available
+        true
     }
 }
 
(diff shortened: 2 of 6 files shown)


sources/updater/update_metadata/src/lib.rs
```diff
@@ -320,34 +319,40 @@ impl Update {
...
-        if let Some(wave) = self.update_wave(seed) {
-            return wave.has_started();
-        }
-
-        // Or there are no waves
-        true
-    }
-
-    pub fn jitter(&self, seed: u32) -> Option<DateTime<Utc>> {
+    /// Returns whether the update is truly available. A update is said to be 'ready/available' if the wave
```
webern:
```suggestion
    /// Returns whether the update is available. An update is said to be 'ready/available' if the wave
```

sources/updater/update_metadata/src/lib.rs
```diff
@@ -320,34 +319,40 @@ impl Update {
...
+    /// this host belongs to has fully passed, or if the host's position in the wave has passed.
+    /// The position of the host within the wave is determined by the seed value.
+    pub fn update_ready(&self, seed: u32, time: DateTime<Utc>) -> bool {
+        // If this host is part of some update wave
         if let Some(wave) = self.update_wave(seed) {
             if wave.has_passed() {
```
webern: Sorry, my previous suggestion to make `update_ready` a pure function was not complete. `has_passed` would need to become `has_passed(time)`. I have the philosophy of pushing impurity upward, but others may not like it. You might want to wait for other reviewers before making such changes! Thanks.

```rust
            if wave.has_passed() {
```

sources/updater/update_metadata/src/lib.rs
```diff
@@ -320,34 +319,40 @@ impl Update {
...
+                    let wave_duration = end - start;
+                    let time_increment = wave_duration / MAX_SEED as i32;
+                    // Derive the target time position within the wave given the host's seed.
+                    let target_time = start + (time_increment * (seed as i32));
+                    // If the current time is past the target time position in the wave, the update is
+                    // marked available
+                    return time >= target_time;
+                } else {
+                    // Check if last wave has started
+                    return time >= start;
```
webern: The last wave does not have any jittering or ramping. We could avoid this by making our last wave always start with MAX_SEED in the manifest (I think). Otherwise I'm not sure how we could avoid all hosts in the last wave hitting at the same time.

sources/updater/README.md
```diff
@@ -41,8 +41,8 @@ Updog will ensure that appropriate migration files are available to safely trans
 
 ### Update wave
 Updates may include "wave" information which provides a way for updates to be scheduled over time for groups of Bottlerocket hosts.
-Updog will find the update wave the host belongs to and jitter its update time within that range.
-If the calculated time has not passed yet, Updog returns the update timestamp to the caller so it can be called again at the correct time.
+Updog will find the update wave the host belongs to and calculate the time position allocated to the host given its seed.
```
tjkirch:
```suggestion
Updog will find the update wave the host belongs to and calculate its time position within the wave based on its `settings.updates.seed` value.
```

sources/updater/update_metadata/src/lib.rs
```diff
@@ -419,6 +424,33 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
...
+        version: Version::parse("1.1.1").unwrap(),
+        max_version: Version::parse("1.1.1").unwrap(),
+        waves,
+        images: Images {
+            boot: String::from("boot"),
+            root: String::from("boot"),
+            hash: String::from("boot"),
+        },
+    };
+    assert!(!update.update_ready(100, Utc::now()), "100 milliseconds hasn't passed yet");
```
tjkirch: Can we generate a single `Utc::now` value at the top of the function to be more consistent with timing?  It shouldn't matter how long the test function takes to run, and it shouldn't fail if the kernel swaps it out for 100 milliseconds before it runs this code, etc.

I think it'd be even better to use a specific timestamp instead of "now", so the millisecond portion can be 0, and we can make increments / differences more obvious.

sources/updater/update_metadata/src/lib.rs
```diff
@@ -419,6 +424,33 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
     .context(error::ManifestParse)
 }
 
+#[test]
+fn test_update_ready_with_seeds() {
+    use chrono::Duration;
+    let mut waves = BTreeMap::new();
+    // One single wave (0th wave does not count) for every update that spans over 2048 millisecond,
```
tjkirch: Wave timing is super important and I want to be sure the math is correct, and we don't break it later

Can we do a few more checks in this function, both positive and negative, and have another test function that checks with multiple waves, ideally ones that aren't linear?

sources/updater/update_metadata/src/lib.rs
```diff
@@ -419,6 +424,33 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
...
+    let update = Update {
+        variant: "".to_string(),
+        arch: "".to_string(),
+        version: Version::parse("1.1.1").unwrap(),
+        max_version: Version::parse("1.1.1").unwrap(),
+        waves,
+        images: Images {
+            boot: String::from("boot"),
+            root: String::from("boot"),
+            hash: String::from("boot"),
```
tjkirch:
```suggestion
            root: String::from("root"),
            hash: String::from("hash"),
```
:)

sources/updater/update_metadata/src/lib.rs
```diff
@@ -320,34 +319,40 @@ impl Update {
...
+                    // marked available
+                    return time >= target_time;
+                } else {
+                    // Check if last wave has started
+                    return time >= start;
                 }
             }
         }
-        None
+        true
```
tjkirch, line 362:
```suggestion
        // There are no waves, so we consider the update available
        true
```

sources/updater/update_metadata/src/lib.rs
```diff
@@ -320,34 +319,40 @@ impl Update {
...
-            return wave.has_started();
-        }
-
-        // Or there are no waves
-        true
-    }
-
-    pub fn jitter(&self, seed: u32) -> Option<DateTime<Utc>> {
+    /// Returns whether the update is truly available. A update is said to be 'ready/available' if the wave
+    /// this host belongs to has fully passed, or if the host's position in the wave has passed.
```
tjkirch:
```suggestion
    /// Returns whether the update is available. A update is said to be 'ready/available' if the wave
    /// this host belongs to has fully passed, or if the host's position in the wave has passed, or if there are no waves.
```

sources/updater/update_metadata/src/lib.rs
```diff
@@ -320,34 +319,40 @@ impl Update {
...
             }
-            let bounds = match self.update_wave(seed) {
-                Some(Wave::Initial { end }) => Some((Utc::now(), end)),
-                Some(Wave::General { start, end }) => Some((start, end)),
-                Some(Wave::Last { start: _ }) | None => None,
+            let bound = match wave {
+                // For all intent and purposes, with the way we define update wave structures,
+                // the initial wave (0th wave) never has any updates.
+                // No seed value would ever land the host in this initial wave. But if for whatever reason
+                // this host is part of the initial (0th) wave, the update is going to be available
```
tjkirch:
```suggestion
                // Hosts should not wind up in the special "initial" wave with no start time, but if they do,
                // we consider the update as being available immediately.
```
I'm not sure I understand the initial wave timing, and it feels like something we're liable to change (I'd really like to get rid of the unused special case) so I'd prefer to simplify this explanation.

sources/updater/update_metadata/src/lib.rs
```diff
@@ -320,34 +319,40 @@ impl Update {
...
+                Wave::Initial { .. } => None,
+                Wave::General { start, end } => Some((start, Some(end))),
+                Wave::Last { start } => Some((start, None)),
             };
-            if let Some((start, end)) = bounds {
-                let mut rng = thread_rng();
-                if let Some(range) = end.timestamp().checked_sub(start.timestamp()) {
-                    return Some(start + Duration::seconds(rng.gen_range(1, range)));
+            if let Some((start, end)) = bound {
+                if let Some(end) = end {
```
tjkirch:
```suggestion
            if let Some((start, maybe_end)) = bound {
                if let Some(end) = maybe_end {
```
Just helps to clarify the difference, I think.

sources/updater/update_metadata/src/lib.rs
```diff
@@ -320,34 +319,40 @@ impl Update {
...
-            let bounds = match self.update_wave(seed) {
-                Some(Wave::Initial { end }) => Some((Utc::now(), end)),
-                Some(Wave::General { start, end }) => Some((start, end)),
-                Some(Wave::Last { start: _ }) | None => None,
+            let bound = match wave {
+                // For all intent and purposes, with the way we define update wave structures,
+                // the initial wave (0th wave) never has any updates.
+                // No seed value would ever land the host in this initial wave. But if for whatever reason
+                // this host is part of the initial (0th) wave, the update is going to be available
+                Wave::Initial { .. } => None,
```
webern: I think this comment could use another pass. The real issue is that this is impossibly hard to understand, no matter how well you write the comment! I'm going to move on, but I was just looking at whether `Wave::Initial` should exist at all. And if it doesn't serve any purpose, then I wonder what is the purpose of the seed value in the first 'wave' listed in the manifest?

Anyway for this PR, I would see if this comment could be tightened up, and maybe add doc comments to the Wave enum.

```rust
                // For all intent and purposes, with the way we define update wave structures,
                // the initial wave (0th wave) never has any updates.
                // No seed value would ever land the host in this initial wave. But if for whatever reason
                // this host is part of the initial (0th) wave, the update is going to be available
                Wave::Initial { .. } => None,
```

webern (approved):

sources/updater/update_metadata/src/lib.rs
```diff
@@ -320,34 +319,40 @@ impl Update {
...
+                Wave::Last { start } => Some((start, None)),
             };
-            if let Some((start, end)) = bounds {
-                let mut rng = thread_rng();
-                if let Some(range) = end.timestamp().checked_sub(start.timestamp()) {
-                    return Some(start + Duration::seconds(rng.gen_range(1, range)));
+            if let Some((start, end)) = bound {
+                if let Some(end) = end {
+                    let wave_duration = end - start;
+                    let time_increment = wave_duration / MAX_SEED as i32;
```
etungsten: @tjkirch rightfully pointed out this should be the number of seeds allocated to the wave instead of MAX_SEED.

Fixing this now...

tjkirch: We talked about this offline - I think we need to get the number of seed values in *this* wave and not the total number of seeds, otherwise we're spreading out across the whole deployment and not just this wave.  We had talked about extending `Wave` to include start/end seeds and not just times, so that this type of math can be done more easily.

tjkirch (commented):
Some comments on update_metadata; I'll go through updog soon!

> This also removes the --timestamp option since it's no longer necessary without updog returning the jittered time.

This is a breaking change, would you please add this part to the commit message too?  (Personally I'd put the next paragraph of the PR description into the commit message, too; people sometimes look there for reasoning and usage of changes.)

Because of that and the other changes to `pub` functions/structs, we need to make sure it goes into a breaking release like 0.5.0.

sources/updater/update_metadata/src/lib.rs
etungsten: I think it makes sense to pass it down through `has_passed` as well.

Reference to @webern 's previous comment: https://github.com/bottlerocket-os/bottlerocket/pull/992#discussion_r455079813

sources/updater/update_metadata/src/lib.rs
etungsten: Yeah, even before this change, the hosts part of the last wave will always all update at once.

That would mean we would always need to define the duration of the last wave in our `waves.toml` which currently we do not. 

I think I'll go ahead and define the duration of the last wave to be 1 day.

So for example, if the last wave has 2048-512=1536 seeds allocated to it (going off of our [`default-waves.toml`](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/updater/waves/default-waves.toml)), then the time increment for each succeeding seed value would be (86400 seconds / 1536) ~= 1 minute. Does that sound reasonable? @tjkirch

sources/updater/update_metadata/src/lib.rs
etungsten: This comment block might help with understanding the `Initial` wave:
https://github.com/bottlerocket-os/bottlerocket/blob/55aee0af4a0b8f7df80e58a9d49165d00d4912ff/sources/updater/update_metadata/src/lib.rs#L232-L254

I agree it's something we should look at removing (will probably have to completely restructure the wave representation in the manifest if we do). But I don't think it's appropriate for this PR.

webern: I didn't see that comment, but maybe those lines 241-245 should be more explicit about what the resulting map is. It feels like something isn't right about the design/implementation, but I'm not quite sure. (And if there is, fixing/simplifying may be beyond the current scope.)

So... 
```
// [ 
 //   UpdateWave { start_after: "1 hour", fleet_percentage: 1 }, 
 //   UpdateWave { start_after: "1 day", fleet_percentage: 100}, 
 // ] 
```

becomes?

```
{
    0 => Wave::Initial{ end: <UTC datetime of 1 hour from now> },
   20 => Wave::General( start: <UTC datetime of 1 hour from now>, end: <UTC datetime of 1 day from now> ),
}
```

etungsten:
[Push above](https://github.com/bottlerocket-os/bottlerocket/compare/7dcc339ca91c72c2d2fb2675b12f7677d0118ed2..3b88ec5fd72cab228cf14cd6b0ec50b68bc37816) addresses @webern and @tjkirch 's comments.


- Moved `test_update_ready` and `test_final_wave` from `updog/src/main.rs` to `update_metadata/src/lib.rs`.
- Updated `Wave` enum to include the seed of the wave, and I use that to calculate the seeds allocated to a particular wave.
- Fixed https://github.com/bottlerocket-os/bottlerocket/pull/993#discussion_r455279741
- For the purpose of determining a host's position in the last wave, the last wave duration is set to 1 day by default.
- Updated commit message to include more details about the changes.

etungsten:
https://github.com/bottlerocket-os/bottlerocket/compare/3b88ec5fd72cab228cf14cd6b0ec50b68bc37816..0dcc5235684461cf1d08170e3ac3976b32a980dd

This force push fixes `test_update_ready`. Hopefully it's more clear.

etungsten:
Made some small changes to the `test_update_ready`.

sources/updater/update_metadata/src/lib.rs
tjkirch: I don't love the idea of using an arbitrary time period for the last wave - I'd rather clean up our understanding of wave definitions, which is currently fairly murky.  I've already seen people interpreting the wave times differently.  The "initial" wave is weird because it shouldn't actually get used, the final wave is weird because of this jump, intermediate waves still have questions about whether the time is start vs. end, etc.

Until we fix that, I think we could set an explicit ramp for the last wave in our wave TOML files by adding another wave at the end for 99/100%.

sources/updater/update_metadata/src/lib.rs
tjkirch: I also suggested a simplification of this comment in my review.  I don't think more detail is what's needed; we'd just continue to struggle to explain a weird, complicated thing.  I think we're moving toward fixing it with Erikson's other changes, and we have more to do, but I'd rather simplify this comment on the road to removing it.

sources/updater/update_metadata/src/lib.rs
```diff
@@ -419,35 +452,167 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
...
+                "seed: {}, time: {}, wave start time: {}, wave start seed: {}, update should be ready",
+                seed,
+                time + Duration::milliseconds(100),
+                time,
+                0,
+            );
+        }
+
+        // Test multiple waves
+        update.waves.clear();
```
tjkirch: It seems like there are conceptually separate tests combined into one, can we split them?  Could use a function/constant for the shared setup.

sources/updater/update_metadata/src/lib.rs
```diff
@@ -419,35 +452,167 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
...
-    assert!(i.next().unwrap() == "migration_1.1.0_a");
+        // First wave starts immediately
+        update.waves.insert(0, time);
+        // First wave ends 200 milliseconds after it starts and has seeds 0 - 50
+        update.waves.insert(50, time + Duration::milliseconds(200));
+        // Second wave ends 1024 milliseconds after it starts and has seeds 50 - 100
+        update
+            .waves
+            .insert(100, time + Duration::milliseconds(1024));
+        // Third wave ends 4096 milliseconds after it starts and has seeds 100 - 1024
```
etungsten: "it" here refers to start of the 0th wave (creation of update)... I should make that more clear ugh

webern (approved):

sources/updater/update_metadata/src/lib.rs
etungsten: Ah sorry, I missed this.

etungsten:
Force push above addresses the following:

- https://github.com/bottlerocket-os/bottlerocket/pull/993#discussion_r455262245
- Removes the default 1 day duration for the last wave. 
- Takes the suggestion from https://github.com/bottlerocket-os/bottlerocket/pull/993#discussion_r455425434 to update our preset wave structures so the last wave has a defined wave duration.
- Updates `test_update_ready` again... Tried to make it more robust.

etungsten:
Push above adds a few more test cases
Push below has minor fixes for those test cases

sources/updater/waves/accelerated-waves.toml
```diff
@@ -15,4 +15,8 @@ fleet_percentage = 50
 
 [[waves]]
 start_after = '1 day'
+fleet_percentage = 99
+
+[[waves]]
+start_after = '2 days'
 fleet_percentage = 100
```
tjkirch: I think we can smooth these out; looks a little weird with the 99 after I see it like this.  Instead of 99, I think it's fine if the last X% release at once - maybe 20% here in accelerated, 10% for default, 30% for ohno?

sources/updater/update_metadata/src/lib.rs
```diff
@@ -296,58 +299,86 @@ impl Manifest {
 impl Update {
     /// Returns the update wave that Updog belongs to, based on the seed value.
     /// Depending on the waves described in the update, the possible results are
-    /// - Some wave described by a start and end time.
-    /// - The "0th" wave, which has an "end" time but no specified start time.
-    /// - The last wave, which has a start time but no specified end time.
+    /// - Some wave described by a start and end time, and the starting seed and ending seed.
+    /// - The "0th" wave, which has an "end" time but no specified start time, and the starting seed.
```
tjkirch:
```suggestion
    /// - The "0th" wave, which has an "end" time but no specified start time, and the ending seed.
```

sources/updater/update_metadata/src/lib.rs
```diff
@@ -296,58 +299,86 @@ impl Manifest {
...
-            (Some(start), Some(end)) => Some(Wave::General { start, end }),
-            (Some(start), None) => Some(Wave::Last { start }),
+            .next();
+
+        match (start_wave, end_wave) {
+            // Note that the key for each wave entry is the starting seed for that wave, the value is the DateTime
+            (None, Some(end_wave)) => Some(Wave::Initial {
+                end_time: *end_wave.1,
+                end_seed: *end_wave.0,
+            }),
```
tjkirch: I'm hoping we can bring together the two wave representations at some point, but for now, can we use pattern matching instead of tuple indexing?  Something like this, and similar below...
```suggestion
            (None, Some((end_seed, end_time))) => Some(Wave::Initial { end_seed, end_time }),
```

sources/updater/update_metadata/src/lib.rs
```diff
@@ -296,58 +299,86 @@ impl Manifest {
...
             };
-            if let Some((start, end)) = bounds {
-                let mut rng = thread_rng();
-                if let Some(range) = end.timestamp().checked_sub(start.timestamp()) {
-                    return Some(start + Duration::seconds(rng.gen_range(1, range)));
+            if let Some((start_time, maybe_end_time, start_seed, end_seed)) = bound {
+                // If the wave has passed, the update is available (this includes passing the last wave start time)
+                if wave.has_passed(time) {
+                    return true;
+                }
```
tjkirch: We could move this up right after we have `wave` to save a little time and make the obvious cases come first, right?  This and `} else if !wave.has_started(time) { return false }`?

etungsten:
Push above adds some comments to the wave toml files.
Fixes a comment in `update_metadata`

sources/updater/update_metadata/src/lib.rs
```diff
@@ -296,58 +299,86 @@ impl Manifest {
...
+                    // Determine the duration of this host's wave
+                    let wave_duration = end_time - start_time;
+                    let num_seeds_allocated_to_wave = (end_seed - start_seed) as i32;
+                    if num_seeds_allocated_to_wave == 0 {
+                        // Empty wave, no host should have been allocated to it
+                        return true;
+                    }
+                    let time_increment = wave_duration / num_seeds_allocated_to_wave;
+                    // Derive the target time position within the wave given the host's seed.
+                    let target_time = start_time + (time_increment * (seed as i32));
```
tjkirch:
```suggestion
                    let time_per_seed = wave_duration / num_seeds_allocated_to_wave;
                    // Derive the target time position within the wave given the host's seed.
                    let target_time = start_time + (time_per_seed * (seed as i32));
```
Maybe it's a slightly clearer name?

sources/updater/update_metadata/src/lib.rs
```diff
@@ -419,35 +450,207 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
...
+            images: Images {
+                boot: String::from("boot"),
+                root: String::from("root"),
+                hash: String::from("hash"),
+            },
+        };
+
+        for seed in (100..500).step_by(100) {
+            assert!(
+                !update.update_ready(seed, time),
```
tjkirch:
```suggestion
                !update.update_ready(seed, time + Duration::milliseconds((seed as i64) - 1)),
```
A bit more specific, proving we become available at the exact time we intend...

sources/updater/update_metadata/src/lib.rs
```diff
@@ -419,35 +450,207 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
...
+        }
+
+        for duration in (200..1024).step_by(4) {
+            assert!(
+                !update.update_ready(seed, time + Duration::milliseconds(duration)),
+                "seed should not part of second wave",
+            );
+        }
+
+        for duration in (1024..((4096 - 1024) / (1024 - 100) * seed)).step_by(4) {
```
tjkirch: Can we pull this out into a variable and explain what it represents?  Same below?  Clarifying exactly when the seed should become available would be helpful.

sources/updater/update_metadata/src/lib.rs
```diff
@@ -419,35 +450,207 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
...
+                    seed,
+                    time + Duration::milliseconds(1024 + 200)
+                        + Duration::milliseconds(duration as i64)
+                ),
+                "update should be ready now that we're passed the allocated time position within the third wave: {}", duration,
+            );
+        }
+
+        // Now we should be in the second wave
+        seed = 60;
```
tjkirch: Also check that it's available at times in the third wave?

sources/updater/waves/accelerated-waves.toml
```diff
@@ -1,5 +1,5 @@
 # The following represents an "accelerated" set of update waves for a much
-# quicker deployment. The deployment lasts for 1 day, and quickly increases the
+# quicker deployment. The deployment lasts for 1.5 days, and quickly increases the
```
tjkirch: Instead of changing our deployment windows, can we add the extra wave a bit before the existing final time?

sources/updater/updog/src/main.rs
```diff
@@ -89,7 +89,6 @@ SUBCOMMANDS:
         [ -i | --image version ]      Update to a specfic image version
```
tjkirch, line 88: It's not let me commenting on unchanged lines, but I think we could improve the usage message for `-a` now that we're clarifying the logic.  `-a` could be "Output all available updates, even if they're not upgrades"?

sources/updater/updog/src/main.rs
```diff
@@ -897,56 +846,66 @@ mod tests {
...
+                },
             "Expected to be 0th wave"
         );
-        assert!(u.jitter(1).is_none(), "Expected immediate update");
         assert!(
-            u.update_wave(101).unwrap() == Wave::Last { start: first_bound },
+            u.update_wave(101).unwrap()
+                == Wave::Last {
+                    start_time: first_bound,
+                    start_seed: 100
```
tjkirch: Should the 100 match the 101?  (Not sure I fully understand what this is doing)

(Same below with 200/100, the 150, the 201/200...)

sources/updater/updog/src/main.rs
```diff
@@ -968,28 +927,49 @@ mod tests {
...
-
         assert!(
-            potential_update.update_ready(512),
-            "0th wave doesn't appear ready"
+            update_required(
+                &manifest,
+                &current_version,
+                &variant,
+                config.ignore_waves,
+                0,
```
tjkirch: This 0 is intended to match the seed from above, which we changed to `Default::default()`?  Is the value important?  If not, can we use `Default::default()` here too?  If it is, can we use a variable or something?

tjkirch (commented):
Woohoo!

One other thing I'd be interested in seeing, in your PR testing, is that the update becomes available when you adjust your seed downward to fit into the current wave.  It's not quite 7/19 yet, so there's a little window to check :)

sources/updater/updog/src/main.rs
etungsten: 101 places it in the last wave. The test is making sure the update belongs to the last wave if seed is 101 and the initial wave goes up to seed 100. Same with the others.

etungsten: After some offline discussion, we decided that the `early_wave` test is redundant because tests in `update_metadata` already cover testing wave ownership of seeds.

etungsten:
Rebased onto develop to fix some minor conflicts.

Addresses @tjkirch 's review comments.

> One other thing I'd be interested in seeing, in your PR testing, is that the update becomes available when you adjust your seed downward to fit into the current wave. It's not quite 7/19 yet, so there's a little window to check :)

Updated PR description to include this testing.

sources/updater/waves/default-waves.toml
```diff
@@ -1,5 +1,5 @@
 # The following represents a "normal" set of update waves for a typical
-# deployment. The deployment lasts for 6 days, and gradually increases the
+# deployment. The deployment lasts for 7 days, and gradually increases the
```
tjkirch: Can we do the same here and not extend the existing deployment schedule?

The way I'm understanding the current schedule, we start opening up from 25% to 90% of hosts between 6 days and 7 days, which is a huge jump.  I think we were misunderstanding the `fleet_percentage` as the target percentage of hosts that have the update available at the related time, when in fact we've found that it's the percentage of hosts that will have the update by the time of the next wave, hence the confusion that led to the jump in this file.  (If that's not right, please correct me!)

How about:
* 1% at 1 hour  (increase of 1% over next 3 hours)
* 5% at 4 hours  (increase of 4% over next 20 hours)
* 15% at 1 day  (increase of 10% over next 24 hours)
* 40% at 2 days  (increase of 25% over next 48 hours)
* 60% at 4 days  (increase of 20% over next 24 hours)
* 90% at 5 days  (increase of 30% over next 24 hours)
* 100% at 6 days  (final 10% at once)

sources/updater/waves/ohno.toml
```diff
@@ -1,14 +1,19 @@
 # The following represents an "emergency" set of update waves for a rapid
-# deployment. The deployment lasts for 3 hours, with a small initial wave,
-# and then all nodes will be updated after the first hour.
+# deployment. The deployment lasts for 3 hours, with a small initial wave.
 [[waves]]
-start_after = '1 hour'
+start_after = '60 minutes'
```
tjkirch: I think the ohno timing is a little odd in that we're waiting a relatively long time to start (for an emergency) and the "big jump" isn't much faster than the previous jump.  This is what happens now:
* 5% at 60 minutes  (increase of 5% over 30 minutes)
* 25% at 90 minutes  (increase of 20% over 30 minutes)
* 70% at 120 minutes  (increase of 45% over 60 minutes)
* 100% at 180 minutes  (final 30% at once)

It won't let me leave a suggestion, but here's what I'd do to make it a bit more like a normal curve, and start a bit sooner:
* 5% at 30 minutes  (increase of 5% over next 30 minutes)
* 25% at 60 minutes  (increase of 20% over next 60 minutes)
* 50% at 120 minutes  (increase of 25% over next 30 minutes)
* 90% at 150 minutes  (increase of 40% over next 30 minutes)
* 100% at 180 minutes  (final 10% at once)

sources/updater/waves/accelerated-waves.toml
```diff
@@ -13,6 +13,12 @@ fleet_percentage = 12
 start_after = '8 hours'
 fleet_percentage = 50
 
+[[waves]]
+start_after = '16 hours'
+fleet_percentage = 80
```
tjkirch, line 18: Similar to the other files, here's what I'd suggest for the accelerated wave structure:

* 3% at 1 hour (increase of 3% over next 3 hours)
* 12% at 4 hours (increase of 9% over next 4 hours)
* 40% at 8 hours (increase of 28% over next 8 hours)
* 80% at 16 hours (increase of 40% over next 8 hours)
* 100% at 24 hours (final 20% at once)

sources/updater/update_metadata/src/lib.rs
```diff
@@ -394,35 +427,231 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
...
+                "seed: {}, time: {}, wave start time: {}, wave start seed: {}, update should be ready",
+                seed,
+                time + Duration::milliseconds(100),
+                time,
+                0,
+            );
+        }
+
+        // Test multiple waves
+        update.waves.clear();
```
tjkirch:
```suggestion
```
Don't need this since we have a separate test now, right?

sources/updater/update_metadata/src/lib.rs
```diff
@@ -271,58 +274,88 @@ impl Manifest {
...
-            (None, Some(end)) => Some(Wave::Initial { end }),
-            (Some(start), Some(end)) => Some(Wave::General { start, end }),
-            (Some(start), None) => Some(Wave::Last { start }),
+            .next();
+
+        match (start_wave, end_wave) {
+            // Note that the key for each wave entry is the starting seed for that wave, the value is the DateTime
+            (None, Some((end_seed, end_time))) => Some(Wave::Initial {
+                end_seed: *end_seed,
+                end_time: *end_time,
```
tjkirch: Can we avoid the repetition and dereferencing if we clone `start_wave` and `end_wave` above, so they're not references?

sources/updater/update_metadata/src/lib.rs
```diff
@@ -394,35 +427,231 @@ pub fn load_manifest<T: tough::Transport>(repository: &tough::Repository<T>) ->
...
+                    seed,
+                    time + Duration::milliseconds(1024 + 200)
+                        + Duration::milliseconds(duration as i64)
+                ),
+                "update should be ready now that we're passed the allocated time position within the third wave: {}", duration,
+            );
+        }
+
+        // Now we should be in the second wave
+        seed = 60;
```
tjkirch: This feels like the starting point for a new test too?

etungsten:

Separated out the `test_update_ready...` test into multiple ones.

Push below removes some redundant wave set up.

etungsten:
Push above refactors the tests a little.

tjkirch (approved):

zmrow (approved):
:neckbeard:

PR #996, opened 2020-07-18
buildsys: exclude README.md from change tracking

bcressey:
**Issue number:**
N/A

**Description of changes:**
This is a follow-up to cd7e66db that also excludes `README.md` from change tracking for buildsys.

**Testing done:**
Confirmed that the `os` package is not rebuilt after changing the mtime of `sources/api/README.md`, but is rebuilt after changing other sources.

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/project.rs
+++ b/tools/buildsys/src/project.rs
@@ -45,7 +45,7 @@ impl ProjectInfo {
         entry
             .file_name()
             .to_str()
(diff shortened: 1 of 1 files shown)


tjkirch (approved):
Ah, neat!  I had forgotten about this walk.

webern (approved):

Issue #815, opened 2020-03-02
ECS Support


jhaynes:
**What I'd like:**

An ECS Variant of Bottlerocket. ECS today uses Docker as the container runtime and packages a go agent as a container along with a package to handle running the agent container correctly.

samuelkarp:
Hey everyone!

I wanted to let you all know what our plan is for the ECS variant.  As of now, we’ll be closing this issue with the next release of Bottlerocket, which will be the first one to include AMIs for the new `aws-ecs-1` variant!   At this initial release, Bottlerocket will support all the following features:

* Logging drivers: awslogs, json-file, journald, none
* Volumes: Local and unencrypted/unauthenticated EFS
* Networking: bridge, host, and none
* IAM roles (both task roles and execution roles)
* Agent introspection and task metadata
* SELinux (Bottlerocket’s container_t, control_t, and super_t labels can be assigned in the task definition)
* Docker’s default seccomp profile
* Privileged mode (disabled by default)
* Troubleshooting log collector (logdog - similar to ecs-logs-collector)
* Custom attributes (set in user-data)
* Intel, AMD, and Graviton/Graviton2 instance types

I’ve created tracking issues for the features that are not currently supported in Bottlerocket.  Please follow those issues and vote with reactions (:+1:) to let us know which you think we should prioritize first:

* Devices
    * [GPU support](https://github.com/bottlerocket-os/bottlerocket/issues/1074)
    * [Inferentia support](https://github.com/bottlerocket-os/bottlerocket/issues/1075)
    * [Elastic Inference support](https://github.com/bottlerocket-os/bottlerocket/issues/1076)
* Networking
    * [awsvpc mode](https://github.com/bottlerocket-os/bottlerocket/issues/1077)
    * [App Mesh support](https://github.com/bottlerocket-os/bottlerocket/issues/1078)
* Logging
    * [FireLens support](https://github.com/bottlerocket-os/bottlerocket/issues/1079)
* Volumes
    * [EFS encryption and IAM authentication support](https://github.com/bottlerocket-os/bottlerocket/issues/1080)

If I’ve missed anything that you’d like to see supported, please feel free to create an issue!

Thanks!
Sam

mwarkentin:
Cool!

samuelkarp:
[...]

PR #1005, opened 2020-07-29
aws-ecs-1: add configurable logging drivers

samuelkarp:
**Issue number:**

**Description of changes:**

The new ecs.logging-drivers setting controls the list of logging drivers available for use with ECS.  The defaults are "json-file", "awslogs", and "none".  Other logging drivers (such as those installed via a plugin) can be enabled.  The default set of logging drivers can be reduced or removed to implement policy for allowable logging drivers in a cluster.

**Testing done:**

Launched an instance, saw it register with the appropriate capabilities.  Ran a task using the awslogs driver, validated that logs appeared in CloudWatch Logs with the correct group and stream.  Used apiclient to mutate the list of logging drivers, saw changes rendered in the ECS config file and reflected in DescribeContainerInstances.

**Terms of contribution:**

### DIFF
--- a/sources/api/ecs-settings-applier/src/main.rs
+++ b/sources/api/ecs-settings-applier/src/main.rs
@@ -30,6 +30,9 @@ struct ECSConfig {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     privileged_disabled: Option<bool>,
+
+    #[serde(skip_serializing_if = "std::vec::Vec::is_empty")]
+    available_logging_drivers: Vec<String>,
 }
 
 // Returning a Result from main makes it print a Debug representation of the error, but with Snafu
@@ -58,6 +61,12 @@ fn run() -> Result<()> {
     let mut config = ECSConfig {
         cluster: ecs.cluster,
         privileged_disabled: ecs.allow_privileged_containers.map(|s| !s),
+        available_logging_drivers: ecs
+            .logging_drivers
+            .unwrap_or_default()
+            .iter()
+            .map(|s| s.to_string())
+            .collect(),
         ..Default::default()
     };
     if let Some(os) = settings.os {
--- a/sources/models/src/aws-ecs-1/override-defaults.toml
+++ b/sources/models/src/aws-ecs-1/override-defaults.toml
@@ -12,3 +12,4 @@ affected-services = ["ecs"]
 
 [settings.ecs]
 allow-privileged-containers = false
+logging-drivers = ["json-file", "awslogs", "none"]
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -111,6 +111,7 @@ struct ECSSettings {
     cluster: String,
     instance_attributes: HashMap<ECSAttributeKey, ECSAttributeValue>,
     allow_privileged_containers: bool,
+    logging_drivers: Vec<SingleLineString>,
 }
 
 // Update settings. Taken from userdata. The 'seed' setting is generated


sources/api/ecs-settings-applier/src/main.rs
```diff
@@ -30,6 +30,9 @@ struct ECSConfig {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     privileged_disabled: Option<bool>,
+
+    #[serde(skip_serializing_if = "std::vec::Vec::is_empty")]
+    available_logging_drivers: Vec<String>,
```
bcressey, line 35: Is there a complete set of known logging drivers, such that we can use an Enum rather than a string? Or does the ability to add logging drivers via plugin make that impossible?

samuelkarp: There's a complete set of built-in logging drivers, so we could do that.  But adding via plugin is also a supported scenario with ECS and making it available requires passing the appropriate value to the ECS agent.  We could decide to disallow plugins for Bottlerocket right now and add it later if that ends up being the preference; I don't have a strong opinion.

webern (approved):

sources/api/ecs-settings-applier/src/main.rs
jahkeup: +1 to keeping them as strings for the time being.

IIRC, if users were to load a log driver plugin at runtime (with a task or otherwise) then they may want to run other tasks using their driver (which has to be listed here).

sources/models/src/aws-ecs-1/override-defaults.toml
```diff
@@ -12,3 +12,4 @@ affected-services = ["ecs"]
 
 [settings.ecs]
 allow-privileged-containers = false
+logging-drivers = ["json-file", "journald", "awslogs", "none"]
```
jahkeup: I'm wondering if `journald` should be listed here if this is a suggested production setting: journal access is limited without the use of the admin container. I suspect we'll see many more users running with log-shipper drivers or transports. Thoughts?

jahkeup (approved):
Looks great! Thanks for exposing the logging settings.

I have one question regarding the default set, see review.

sources/models/src/aws-ecs-1/override-defaults.toml
samuelkarp: > I'm wondering if `journald` should be listed here if this is a suggested production setting: journal access is limited without the use of the admin container.

This is true for `json-file` as well, though you can get access to that by mounting the Docker socket.

> I suspect we'll see many more users running with log-shipper drivers or transports.

Agreed.

> Thoughts?

I don't have a strong opinion on it; I lean toward leaving it in but am happy to remove if you think that's better.

jahkeup: > This is true for `json-file` as well, though you can get access to that by mounting the Docker socket.

This pattern, in my experience, appears to be understood and widely accepted - even if it blurs security boundaries in the process. I've seen logging tools (such as [elastic's `beat` docker log input](https://github.com/elastic/beats/blob/aa9c1eab64686203b9f1e606f304d8f35613d976/filebeat/docs/inputs/input-docker.asciidoc#docker-input)) that use the buffered stream from `json-file` that then ship off or handle events by listening in on the log using `json-file` as the driver.

I'm in favor of removing `journald` from the default set. However, `json-file` should remain in the set; `json-file` is *docker*'s default log driver and not including it would prevent tasks from scheduling even if they don't have LogConfiguration set.

bcressey (approved):

sources/models/src/aws-ecs-1/override-defaults.toml
samuelkarp: > I'm in favor of removing `journald` from the default set.

Done

> However, `json-file` should remain in the set; `json-file` is _docker_'s default log driver and not including it would prevent tasks from scheduling even if they don't have LogConfiguration set.

This isn't quite accurate.  If LogConfiguration isn't set, no explicit or implicit placement constraints are applied; the task can be placed anywhere in the cluster without respect to log drivers and it will use whatever the default driver is on the instance where it is placed (`json-file` is the default, but that can be configured in Docker's config file).

jahkeup (approved):

PR #1007, opened 2020-08-03
aws-ecs-1: enable task IAM role support

samuelkarp:
**Issue number:**

**Description of changes:**
This change enables the task role capabilities so Bottlerocket instances become eligible for placement for tasks with roles.

**Testing done:**
Launched a new Bottlerocket instance.  Saw the following capabilities registered successfully:
* `com.amazonaws.ecs.capability.task-iam-role-network-host`
* `com.amazonaws.ecs.capability.task-iam-role`

Ran two tasks using task roles, one in the `bridge` (default) network mode and the other in `host` (requires the `task-iam-role-network-host` capability).  Observed that the containers were able to use the role and successfully make API calls requiring permissions that were not in the instance's role.

**Terms of contribution:**

### DIFF
--- a/sources/api/ecs-settings-applier/src/main.rs
+++ b/sources/api/ecs-settings-applier/src/main.rs
@@ -33,6 +33,12 @@ struct ECSConfig {
 
     #[serde(skip_serializing_if = "std::vec::Vec::is_empty")]
     available_logging_drivers: Vec<String>,
+
+    #[serde(rename = "TaskIAMRoleEnabled")]
+    task_iam_role_enabled: bool,
+
+    #[serde(rename = "TaskIAMRoleEnabledForNetworkHost")]
+    task_iam_role_enabled_for_network_host: bool,
 }
 
 // Returning a Result from main makes it print a Debug representation of the error, but with Snafu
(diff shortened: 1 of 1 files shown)


sources/api/ecs-settings-applier/src/main.rs
```diff
@@ -67,6 +73,8 @@ fn run() -> Result<()> {
             .iter()
             .map(|s| s.to_string())
             .collect(),
+        task_iam_role_enabled: true,
+        task_iam_role_enabled_for_network_host: true,
```
jahkeup, line 79: If we're expecting to lift these into settings (in the near future), then let's add a comment saying that. Otherwise, the static tuning should be commented with our rationale behind em.

samuelkarp: I can add a comment.  I have no plans to lift these into settings unless someone actually wants to disable the feature.

webern (approved):

sources/api/ecs-settings-applier/src/main.rs
samuelkarp: @jahkeup Done

jahkeup (approved):

PR #1008, opened 2020-08-05
ecs-settings-applier: conditional compilation

samuelkarp:
**Description of changes:**
This commit will compile ecs-settings-applier with a no-op body when any variant other than aws-ecs-1 is used.  This allows for the top-level `cargo make` targets like `unit-tests` to function properly for all variants.

**Testing done:**
* `cargo check` with a variety of different `VARIANT` values provided, ensuring that cargo only rebuilds when the variant changes and builds the appropriate code
* `cargo make unit-tests` with `-e BUILDSYS_VARIANT` set to different variants, ensuring that compilation does not fail.

Note: `rustc` warns for unused imports, constants, structs, and functions when a variant other than `aws-ecs-1` is used.  If we want to reduce the set of warnings, I'd appreciate suggestions on the best way to conditionally define those items (other than sprinkling `#[cfg(variant="aws-ecs-1")]` all over the place).

**Terms of contribution:**


tjkirch (commented):
Nice find with rustc-cfg!  Is it possible to do the conditionalizing at a higher level than the run function, so the whole crate isn't compiled?  We know we're not including this binary, so it feels silly to build it, even if it wouldn't do anything when run.  (It'd get rid of the warnings too...)

webern (approved):
Makes sense.

webern:
> Nice find with rustc-cfg! Is it possible to do the conditionalizing at a higher level than the run function, so the whole crate isn't compiled? We know we're not including this binary, so it feels silly to build it, even if it wouldn't do anything when run. (It'd get rid of the warnings too...)

I was trying to find something like this, too. I was thinking at the workspace level, but I did not find anything that could conditionally leave the crate out of the workspace.

tjkirch:
> I was trying to find something like this, too. I was thinking at the workspace level, but I did not find anything that could conditionally leave the crate out of the workspace.

I couldn't find anything with a quick look, but what if we put all of the code in a module and `#[cfg]` that, rather than just having `#[cfg]` on `run`?  Then none of the other structs/functions/etc. would compile, either.

A couple thoughts around options:
* src/ecs.rs, with all the code from main.rs, and have main.rs just do something like `#[cfg(ecs)] use ecs::*`?
* symlink strategy, like models, so it'd be based on variant, and we could actually just have an empty file for any other variant.  Downside: still need directories/etc. for any new variant, even though they wouldn't care about this crate.

samuelkarp:
> src/ecs.rs, with all the code from main.rs, and have main.rs just do something like `#[cfg(ecs)] use ecs::*`?

I think I like this one better than symlinks.  We'll still end up building the crate as empty, but it'll get rid of the warnings at least.

samuelkarp:
@tjkirch GitHub doesn't show the diff well, but `src/main.rs` was moved to `src/ecs.rs`, slightly modified to remove the `#[cfg]` lines and make `main()` public, and a new `src/main.rs` was added that calls the old one.

tjkirch (approved):

bcressey (approved):

PR #1009, opened 2020-08-06
aws-ecs-1: enable SELinux capability

samuelkarp:
**Issue number:**

**Description of changes:**
Configure the ECS agent to register the SELinux capability so SELinux options can be set in the task definition:

```
"dockerSecurityOptions": [
    "label:type:super_t"
]
```

**Testing done:**
* Ran a task through ECS with a restricted host path (`/etc`) mounted and no explicit SELinux options.  Observed that the container received the default `container_t` label and that it cannot write to the `/etc` bind-mount.
* Ran a task through ECS with a restricted host path (`/etc`) mounted and the `super_t` label assigned in the task definition.  Observed that the container received the `super_t` label and that it can write to the `/etc` bind-mount.

**Terms of contribution:**

### DIFF
--- a/sources/api/ecs-settings-applier/src/ecs.rs
+++ b/sources/api/ecs-settings-applier/src/ecs.rs
@@ -4,7 +4,8 @@
 ecs-settings-applier generates a configuration file for the ECS agent from Bottlerocket settings.
 
 The configuration file for ECS is a JSON-formatted document with conditionally-defined keys and
-embedded lists.
+embedded lists.  The structure and names of fields in the document can be found
+[here](https://github.com/aws/amazon-ecs-agent/blob/a250409cf5eb4ad84a7b889023f1e4d2e274b7ab/agent/config/types.go).
 */
 use log::debug;
 use serde::Serialize;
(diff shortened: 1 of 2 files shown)


webern (approved):

sources/api/ecs-settings-applier/src/ecs.rs
```diff
@@ -38,6 +38,9 @@ struct ECSConfig {
 
     #[serde(rename = "TaskIAMRoleEnabledForNetworkHost")]
     task_iam_role_enabled_for_network_host: bool,
+
+    #[serde(rename = "SELinuxCapable")]
```
jahkeup, line 43: Let's add a link to the [`types.go`](https://github.com/aws/amazon-ecs-agent/blob/a250409cf5eb4ad84a7b889023f1e4d2e274b7ab/agent/config/types.go) package that define the field names in the doc string (so it appears in the readme). The JSON structure is defined [there](https://github.com/aws/amazon-ecs-agent/blob/a250409cf5eb4ad84a7b889023f1e4d2e274b7ab/agent/config/types.go) and is authoritative for the "pass through" names we are seeing here.

jahkeup (approved):
Looks good! :closed_lock_with_key: :whale:

bcressey (approved):

sources/api/ecs-settings-applier/src/ecs.rs
samuelkarp: Done

jahkeup (approved):

PR #1014, opened 2020-08-07
Update SDK to 0.12.0

iliana:
**Issue number:**

n/a

**Description of changes:**

Move to SDK v0.12.0 (https://github.com/bottlerocket-os/bottlerocket-sdk/pull/30). Two changes on the Bottlerocket side:

* [The new `embed-bitcode` flag](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1450-2020-07-16) conflicts with `-C lto`. https://github.com/rust-lang/rust/issues/71848 suggests using Cargo-specific environment variables to get correct behavior, as "Cargo has much more intrusive knowledge of LTO now for compile time speedups".
* ~~Updated to glibc 2.32 in this PR, otherwise our Rust code doesn't build on aarch64 due to missing symbol versions. This will probably conflict with @tjkirch's version-bumps-in-progress, and I can cherry-pick his commit over here to replace mine to get rid of the conflict (I think).~~ Looks like glibc updates need to be lockstep across the SDK and the OS.

**Testing done:**

It builds, need to do a boot test.

CI will fail until I actually perform an SDK release.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -55,7 +55,7 @@ DOCKER_BUILDKIT = "1"
 # on the command line.
 
 # Depends on ${BUILDSYS_ARCH}.
-BUILDSYS_SDK_IMAGE = "bottlerocket/sdk-${BUILDSYS_ARCH}:v0.11.0"
+BUILDSYS_SDK_IMAGE = "bottlerocket/sdk-${BUILDSYS_ARCH}:v0.12.0"
 
 # Depends on ${BUILDSYS_JOBS}.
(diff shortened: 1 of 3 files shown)


tjkirch (approved):
Presuming we have an answer for host/target glibc symbol mismatches -- which this lockstep update might address, but I don't fully understand, so help me @bcressey...

[edit] :white_check_mark:

bcressey (approved):

PR #1025, opened 2020-08-11
docker: upgrade to 19.03.12

samuelkarp:
**Description of changes:**
Upgrade docker-engine and docker-cli to 19.03.12.

**Testing done:**
Built and ran both the `aws-dev` and `aws-ecs-1` variants.  Both variants are able to run containers.  `docker version` output looks reasonable.

**Terms of contribution:**

### DIFF
--- a/packages/docker-cli/clarify.toml
+++ b/packages/docker-cli/clarify.toml
@@ -5,8 +5,8 @@ license-files = [
 ]
 skip-files = ["license.go", "model/license.go"]
 
-[clarify."github.com/ghodss/yaml"]
+[clarify."sigs.k8s.io/yaml"]
 expression = "MIT AND BSD-3-Clause"
 license-files = [
-    { path = "LICENSE", hash = 0xcdf3ae00 },
+    { path = "LICENSE", hash = 0xcdf3ae00},
 ]
--- a/packages/docker-engine/docker-engine.spec
+++ b/packages/docker-engine/docker-engine.spec
@@ -7,8 +7,11 @@
 %global dorepo docker
 %global doimport %{goproject}/%{dorepo}
 
-%global gover 18.09.9
+%global gover 19.03.12
 %global rpmver %{gover}
+%global gitrev 9dc6525e6118a25fab2be322d1914740ea842495
+
+%global source_date_epoch 1363394400
 
 %global _dwz_low_mem_die_limit 0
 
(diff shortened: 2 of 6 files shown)


packages/docker-cli/clarify.toml
```diff
@@ -5,8 +5,8 @@ license-files = [
...
 skip-files = ["license.go", "model/license.go"]
 
-[clarify."github.com/ghodss/yaml"]
+[clarify."sigs.k8s.io/yaml"]
 expression = "MIT AND BSD-3-Clause"
 license-files = [
-    { path = "LICENSE", hash = 0xcdf3ae00 },
-]
+    { path = "LICENSE", hash = 0xcdf3ae00},
+]
```
tjkirch, line 12: nits: lost final newline, and space before `}`

tjkirch (approved):

bcressey (approved):

iliana (approved):

packages/docker-engine/docker-engine.spec
```diff
@@ -7,8 +7,11 @@
 %global dorepo docker
 %global doimport %{goproject}/%{dorepo}
 
-%global gover 18.09.9
+%global gover 19.03.12
 %global rpmver %{gover}
+%global gitrev 9dc6525e6118a25fab2be322d1914740ea842495
+
+%global source_date_epoch 1363394400
```
jamieand, line 14: Where does this timestamp come from?

samuelkarp: This is, as near as I can tell, Docker's birthday.  (The date and time of [this talk](https://pyvideo.org/pycon-us-2013/the-future-of-linux-containers.html).)

webern (approved):

samuelkarp:
`--security-opt label:disable` appears to have different behavior now; in the previous version of Docker it would successfully disable all process and mount labeling.  With this upgraded version of Docker, I'm still seeing the `system_u:system_r:container_t:s0` process label and `system_u:object_r:local_t:s0:c322,c990` mount label applied.  `--security-opt label:type:super_t` behaves as expected.

samuelkarp:
Discussed with @bcressey offline; we believe the change in behavior is fine for now, it ends up with strictly more-compatible behavior for the case where `label:disable` is provided (the container can actually write to its rootfs).  We'll need to revisit a little when we revise our SELinux policy for MCS enforcement as the rootfs mount currently has an MCS pair but the process does not.

PR #1028, opened 2020-08-11
Address feedback about `cargo make ami`

zmrow:
```
commit 89e2a37afafb7d2b088572bed723971467cb8730
Date:   Tue Aug 11 17:41:16 2020 +0000

    pubsys: default AwsConfig so it's not required

    Have serde default `aws.region` so the user does not have to have an empty
    `aws.region` table if they have no region-specific configuration.

    Have serde default `aws.regions` in case the user only wants to specify regions
    on the command line with `PUBLISH_REGIONS`.

    Derive Default on AwsConfig now that all sections of `aws` in Infra.toml are
    optional, so the user doesn't need an empty `[aws]` section if they intend to
    use default credential mechanisms and specify `PUBLISH_REGIONS`.
```

```
commit b69d2a218178914d8081c7e0694318da98a439ae
Date:   Tue Aug 11 17:19:42 2020 +0000

    Separate publish tasks

    If you just want to build images, you don't need to build the pubsys tool
    first, only buildsys.  This change splits the tasks so that builds only need to
    depend on buildsys.

    Similarly, AMI builds only need to depend on pubsys, not buildsys.  `cargo
    make` currently rebuilds images each run.  If you want to build or copy AMIs
    based on a previous image build, you don't want to wait for image rebuilds, and
    you don't want to use different files - you want to use the same image files
    you had.  This change makes the ami task depend only on pubsys and sources, and
    checks for image file existence internally.  (The image file names include
    version and commit, so you can't accidentally use an old build.)

    This will also apply for `cargo make ssm`, where timing is even more critical.
```

**Testing done:**

Before: `cargo make` built pubsys unnecessarily.
After: `cargo make` does not build pubsys; `cargo make ami` still does.

Before: images rebuilt every time before `cargo make ami` when it should use existing images.
After: `cargo make ami` fails appropriately if no images exist, and runs quickly and correctly when they do.

Before: aws.region was required in Infra.toml even when you don't have region-specific config.
Before: aws.regions was required in Infra.toml even if you wanted to override with PUBLISH_REGIONS.
After: `cargo make ami -e PUBLISH_REGIONS=bla1,bla2` works, even with no `aws` section in Infra.toml.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -181,7 +181,13 @@ cargo install \
   --root tools \
   --force \
   --quiet
+'''
+]
 
+[tasks.publish-tools]
+dependencies = ["setup", "fetch-sources"]
+script = [
+'''
 cargo install \
   ${CARGO_MAKE_CARGO_ARGS} \
   --path tools/pubsys \
@@ -281,7 +287,10 @@ alias = "build"
 # to create a repo under /build/repos, named after the arch/variant/version,
 # containing subdirectories for the repo metadata and targets.
 [tasks.repo]
-dependencies = ["build"]
+# Rather than depend on "build", which currently rebuilds images each run, we
+# check for the image files below to save time.  This does mean that `cargo
+# make` must be run before `cargo make repo`.
+dependencies = ["publish-tools"]
 script_runner = "bash"
 script = [
 '''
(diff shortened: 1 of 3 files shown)


tjkirch (approved):

etungsten:
> Before: images rebuilt every time before cargo make ami when it should use existing images.
After: cargo make ami fails appropriately if no images exist, and runs quickly and correctly when they do.

Can we give the same treatment to `cargo make repo`? The `repo` target currently depends on the `build` step.

tjkirch:
> Can we give the same treatment to `cargo make repo`? The `repo` target currently depends on the `build` step.

I think we can, yeah.

tjkirch:
^ This push addresses the comment from @etungsten by making `cargo make repo` also check for image files internally, like AMI.  We confirmed that missing image files (or migration tarball) error, and with them a repo is built correctly.

etungsten (approved):

bcressey (approved):

PR #1030, opened 2020-08-11
defaults, migrations: host-containers default version bumps and migrations

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
Author: Erikson Tung <etung@amazon.com>
Date:   Tue Aug 11 15:40:09 2020 -0700

    defaults: update default versions for host containers
    
    Set default admin container version to v0.5.1
    Set default control container version to v0.4.1
```

```
Author: Erikson Tung <etung@amazon.com>
Date:   Tue Aug 11 13:04:30 2020 -0700

    migrations: host-containers default version migrations
    
    Migrates default version for the admin container from v0.5.0 to v0.5.1
    Migrates default version for the control container from v0.4.0 to v0.4.1
```

**Testing done:**
Created a local datastore for testing purposes with `storewolf`.
For the admin container migration:

Initial `admin.source.template` and `admin.source`
```
$ cat ~/thar/testing/ds/current/live/settings/host-containers/admin/source
"328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.5.0"
$ cat ~/thar/testing/ds/current/live/settings/host-containers/admin/source.template 
"328549459982.dkr.ecr.{{ settings.aws.region }}.amazonaws.com/bottlerocket-admin:v0.5.0"
```

Run the admin container forward migration:
```
     Running `migrate-admin-container-v0-5-1 --source-datastore ~/thar/testing/ds/current --target-datastore ~/thar/testing/ds/next --forward`
Updating template and value of 'settings.host-containers.admin.source' on upgrade
Changing template of 'settings.host-containers.admin.source' from '328549459982.dkr.ecr.{{ settings.aws.region }}.amazonaws.com/bottlerocket-admin:v0.5.0' to '328549459982.dkr.ecr.{{ settings.aws.region }}.amazonaws.com/bottlerocket-admin:v0.5.1'
Changing value of 'settings.host-containers.admin.source' from '328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.5.0' to '328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.5.1'
Found no 'settings.host-containers.admin.source' to change on upgrade
```

See that the `admin.source.template` and `admin.source` got updated:
```
$ cat ~/thar/testing/ds/next/live/settings/host-containers/admin/source
"328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.5.1"
[...]

PR #1032, opened 2020-08-12
schnauzer: add a helper for ecr repos

webern:
**Issue number:**

N/A

**Description of changes:**

In order to expand Bottlerocket to certain AWS regions, we need to change the way the ECR URL is constructed. Initially we intended to use a single ECR registry ID. For some regions this will not work. So we added a helper function and a map to Schnauzer so that we can look up the registry ID from a map.

Additionally, this allows us to introduce the concept of a fallback ID/region pair. This will prove useful if a new region is added and, for example, an older version of Bottlerocket tries to run on it.

*NOTE* we need to update the migration versionized-directrory and the container tag versions that it hard-codes based on where things stand when we release these changes.

**Testing done:**

Ran an instance in me-south-1, the admin container worked, and I found these container URLs in the settings:

```
509306038620.dkr.ecr.me-south-1.amazonaws.com/bottlerocket-control:v0.4.1
509306038620.dkr.ecr.me-south-1.amazonaws.com/bottlerocket-admin:v0.5.2
```

Created a fake future version of Bottlerocket and ran an upgrade in us-west-2. Observed that the migrations ran and the URLs were still good.

**Terms of contribution:**

### DIFF
--- a/Release.toml
+++ b/Release.toml
@@ -7,3 +7,5 @@ version = "0.5.0"
 "(0.3.4, 0.4.0)" = []
 "(0.4.0, 0.4.1)" = ["migrate_v0.4.1_add-version-lock-ignore-waves.lz4", "migrate_v0.4.1_pivot-repo-2020-07-07.lz4"]
 "(0.4.1, 0.5.0)" = ["migrate_v0.5.0_add-cluster-domain.lz4", "migrate_v0.5.0_admin-container-v0-5-2.lz4", "migrate_v0.5.0_control-container-v0-4-1.lz4"]
+# TODO - add ecr-helper-admin and ecr-helper-control migrations to next release.
+# NOTE - these expect container versions v0.5.2 and v0.4.1
--- a/sources/api/migration/migrations/v0.5.1/ecr-helper-admin/src/main.rs
+++ b/sources/api/migration/migrations/v0.5.1/ecr-helper-admin/src/main.rs
@@ -1,0 +1,29 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::ReplaceTemplateMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+const OLD_ADMIN_CTR_TEMPLATE: &str =
+    "328549459982.dkr.ecr.{{ settings.aws.region }}.amazonaws.com/bottlerocket-admin:v0.5.2";
+const NEW_ADMIN_CTR_TEMPLATE: &str =
+    "{{ ecr-prefix settings.aws.region }}/bottlerocket-admin:v0.5.2";
+
+/// We added a helper to lookup an ECR registry number by region.
+fn run() -> Result<()> {
+    migrate(ReplaceTemplateMigration {
+        setting: "settings.host-containers.admin.source",
+        old_template: OLD_ADMIN_CTR_TEMPLATE,
+        new_template: NEW_ADMIN_CTR_TEMPLATE,
+    })
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/api/schnauzer/src/helpers.rs
+++ b/sources/api/schnauzer/src/helpers.rs
@@ -3,8 +3,45 @@
 // text at render time.
 
 use handlebars::{Context, Handlebars, Helper, Output, RenderContext, RenderError};
+use lazy_static::lazy_static;
 use serde_json::value::Value;
 use snafu::{OptionExt, ResultExt};
+use std::borrow::Borrow;
+use std::collections::HashMap;
+
+lazy_static! {
+    /// A map to tell us which registry to pull ECR images from for a given region.
+    static ref ECR_MAP: HashMap<&'static str, &'static str> = {
+        let mut m = HashMap::new();
+        m.insert("af-south-1", "917644944286");
+        m.insert("ap-east-1", "375569722642");
+        m.insert("ap-northeast-1", "328549459982");
+        m.insert("ap-northeast-2", "328549459982");
+        m.insert("ap-south-1", "328549459982");
+        m.insert("ap-southeast-1", "328549459982");
+        m.insert("ap-southeast-2", "328549459982");
+        m.insert("ca-central-1", "328549459982");
+        m.insert("eu-central-1", "328549459982");
+        m.insert("eu-north-1", "328549459982");
+        m.insert("eu-south-1", "586180183710");
+        m.insert("eu-west-1", "328549459982");
+        m.insert("eu-west-2", "328549459982");
+        m.insert("eu-west-3", "328549459982");
+        m.insert("me-south-1", "509306038620");
+        m.insert("sa-east-1", "328549459982");
+        m.insert("us-east-1", "328549459982");
+        m.insert("us-east-2", "328549459982");
+        m.insert("us-west-1", "328549459982");
+        m.insert("us-west-2", "328549459982");
+        m
+    };
+}
+
+/// But if there is a region that does not exist in our map (for example a new
+/// region is created or being tested), then we will fallback to pulling ECR
+/// containers from here.
+const ECR_FALLBACK_REGION: &str = "us-east-1";
+const ECR_FALLBACK_REGISTRY: &str = "328549459982";
 
 /// Potential errors during helper execution
 mod error {
@@ -14,6 +51,18 @@ mod error {
     #[derive(Debug, Snafu)]
     #[snafu(visibility = "pub(super)")]
     pub(super) enum TemplateHelperError {
+        #[snafu(display(
+            "Expected ecr helper to be called with either 'registry' or 'region', got '{}'",
+            value,
+        ))]
+        EcrParam { value: String },
+
+        #[snafu(display("Expected an AWS region, got '{}' in template {}", value, template))]
+        EcrRegion {
+            value: handlebars::JsonValue,
+            template: String,
+        },
+
         #[snafu(display(
             "Incorrect number of params provided to helper '{}' in template '{}' - {} expected, {} received",
             helper,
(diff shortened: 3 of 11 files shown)


sources/api/migration/migrations/v0.5.1/ecr-helper-admin/src/main.rs
```diff
@@ -0,0 +1,29 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::ReplaceTemplateMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+const OLD_ADMIN_CTR_TEMPLATE: &str =
+    "328549459982.dkr.ecr.{{ settings.aws.region }}.amazonaws.com/bottlerocket-admin:v0.5.0";
```
tjkirch: I think we want the versions to be 0.5.1 (admin) and 0.4.1 (control), after https://github.com/bottlerocket-os/bottlerocket/pull/1030/, in the migrations and the defaults...

sources/api/schnauzer/src/helpers.rs
```diff
@@ -422,6 +400,163 @@ pub fn default(
...
+///
+/// # Example
+///
+/// In this example the account number for the region will be returned.
+/// `{{ ecr "account" settings.aws.region }}`
+///
+/// In this example the region will either be returned, or if it happens to be
+/// a region for which we don't have an ECR repo, the fallback region will be
+/// returned.
+/// `{{ ecr "region" settings.aws.region }}`
```
tjkirch: I think it'd be simpler to have a single helper call that returned the whole ECR prefix (for now, that means the domain) rather than using two calls, and in particular the region call that seems a little awkward to me.  Are there any cases where we'd want the account or region but not the whole domain?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -422,6 +400,163 @@ pub fn default(
...
 }
 
+/// Gets the value at `idx` and unwraps it. Assumes you have already checked that `idx` is in range
+/// and says so in an error message if the value cannot be obtained.
+fn get_param<'a>(helper: &'a Helper<'_, '_>, idx: usize) -> Result<&'a Value, RenderError> {
+    Ok(helper
+        .param(idx)
+        .map(|v| v.value())
+        .context(error::Internal {
+            msg: "Missing param after confirming there are enough",
```
tjkirch:
```suggestion
        .context(error::MissingParam {
            number: idx,
            helper_name: helper.name(),
```
We can't assume they've checked for enough parameters anymore, so I think this makes the failure clearer.

(Also, should the get_param/template_name refactor be a separate, earlier commit?)

sources/api/schnauzer/src/helpers.rs
```diff
@@ -422,6 +400,163 @@ pub fn default(
...
+
+/// Get the template name if there is one, otherwise return "dynamic template"
+fn template_name<'a>(renderctx: &'a mut RenderContext<'_, '_>) -> &'a str {
+    match renderctx.get_root_template_name() {
+        Some(s) => s.as_str(),
+        None => "dynamic template",
+    }
+}
+
+/// The `ecr` helper is used to map an aws-region to the correct ECR account.
```
tjkirch:
```suggestion
/// The `ecr` helper is used to map an AWS region to the correct Bottlerocket ECR account.
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -422,6 +400,163 @@ pub fn default(
...
+    _: &Context,
+    renderctx: &mut RenderContext<'_, '_>,
+    out: &mut dyn Output,
+) -> Result<(), RenderError> {
+    trace!("Starting ecr helper");
+    let template_name = template_name(renderctx);
+    if helper.params().len() != 2 {
+        return Err(RenderError::from(
+            error::TemplateHelperError::IncorrectNumberOfParams {
+                expected: 4,
```
tjkirch:
```suggestion
                expected: 2,
```
Maybe we need a proc macro that gets the maximum get_param number requested and checks it at the start of the function :) :black_joker:

sources/api/schnauzer/src/helpers.rs
```diff
@@ -422,6 +400,163 @@ pub fn default(
...
+                received: helper.params().len(),
+                helper: helper.name().to_string(),
+                template: template_name.to_owned(),
+            },
+        ));
+    }
+
+    // get the first parameter, which should be either 'account' or 'region'.
+    // convert this to a boolean is_account. error if param is bad.
```
tjkirch: I think an `enum EcrMode { Account, Region }` (terrible name, but you get the idea) would be clearer than the boolean.

(See also where I suggested removing this parameter entirely...)

sources/api/schnauzer/src/helpers.rs
```diff
@@ -422,6 +400,163 @@ pub fn default(
...
+            return Err(RenderError::from(error::TemplateHelperError::EcrParam {
+                value: account_or_region.to_owned(),
+            }))
+        }
+    };
+
+    // get the second parameter, which is probably given by the template value
+    // settings.aws.region. regardless, we expect it to be a string.
+    let aws_region = get_param(helper, 1)?;
+    let aws_region = aws_region.as_str().ok_or_else(|| {
```
tjkirch: Why `ok_or_else` rather than `with_context` like on line 474's `as_str`?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -422,6 +400,163 @@ pub fn default(
...
+
+/// If there is no region setting at all, then we will error. But if there is a
+/// region setting that does not exist in our map (for example a new region is
+/// created or being tested), then we will fallback to pulling ECR containers
+/// from us-east-1.
+const ECR_FALLBACK_REGION: &str = "us-east-1";
+const ECR_FALLBACK_ACCOUNT: &str = "328549459982";
+
+/// A map to tell us which account to pull ECR images from for a given region.
+fn ecr_map() -> &'static HashMap<&'static str, &'static str> {
```
tjkirch:
```suggestion
```
Can we just do the `lazy_static!` at the top level rather than in a function?  It still won't be constructed until needed, and keeping it at the top level is more common.  Their example does a hashmap like this: https://docs.rs/lazy_static/1.4.0/lazy_static/#example

sources/api/schnauzer/src/helpers.rs
```diff
@@ -422,6 +400,163 @@ pub fn default(
...
+        RenderError::from(error::TemplateHelperError::EcrRegion {
+            value: aws_region.to_owned(),
+            template: template_name.to_owned(),
+        })
+    })?;
+
+    // get a mapping of regions to their ECR accounts. find the pair for our
+    // region, or the fallback account and region if our region is not found.
+    let ecr_map = ecr_map();
+    let ecr_region_and_account = match ecr_map.get(aws_region) {
```
tjkirch:
```suggestion
    let (region, account) = match ecr_map.get(aws_region) {
```
Then we shouldn't need the `.0` and `.1` below.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -422,6 +400,163 @@ pub fn default(
...
+        None => (ECR_FALLBACK_REGION, ECR_FALLBACK_ACCOUNT),
+        Some(account) => (aws_region, *account),
+    };
+
+    // write either the resultant account or resultant region to the template.
+    if is_account {
+        out.write(ecr_region_and_account.1)
+            .context(error::TemplateWrite {
```
tjkirch:
```suggestion
            .with_context(|| error::TemplateWrite {
```
In places where we have to do something to create the error (often, cloning) this will remove that from the success case.  (Often, though, snafu will handle things like this for you; it's worth trying to pass references or just not calling `to_owned` rather than switching to `with_context`.)

iliana:
Can/should we combine this migration with #1030?

webern:
> Can/should we combine this migration with #1030?

It's possible, but I'm not sure I'll be ready in time. I need to test this with confidence, which would be today (8/12) EOD/late at best.

webern:
> force-pushed the webern:ecr-repos branch from 95538ce to 1961962

Fixed some wording, have not addressed PR comments yet.

sources/api/schnauzer/src/helpers.rs
webern: Makes sense, but I want to make sure I understand what you mean. So, would `defaults.toml` look like this?

```toml
[metadata.settings.host-containers.admin.source]
setting-generator = "schnauzer settings.host-containers.admin.source"
template = "{{ ecr settings.aws.region }}.amazonaws.com/bottlerocket-admin:v0.5.0"
```

tjkirch: `amazonaws.com` can change too, so I'd have:
```
[metadata.settings.host-containers.admin.source]
setting-generator = "schnauzer settings.host-containers.admin.source"
template = "{{ ecr-prefix settings.aws.region }}/bottlerocket-admin:v0.5.0"
```

webern:
> force-pushed the webern:ecr-repos branch from 1961962 to 7b5acbf now

separates the refactors from the new code (two commits)

webern:
> force-pushed the webern:ecr-repos branch from 7b5acbf to 3ee106c

Rebase develop with conflict resolution over the container version bumps.

webern:
>  webern force-pushed the webern:ecr-repos branch from 3ee106c to 356423d 

Change design to return the whole fqdn instead of registry and region separately. Also addressed all other PR comments to-date.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -656,9 +753,73 @@ mod test_default {
...
+        T: Serialize,
+    {
+        let mut registry = Handlebars::new();
+        registry.register_helper("ecr-prefix", Box::new(ecr_prefix));
+
+        registry.render_template(tmpl, data)
+    }
+
+    const ADMIN_CONTAINER_TEMPLATE: &str =
+        "{{ ecr-prefix settings.aws.region }}/bottlerocket-admin:v0.5.0";
```
tjkirch: nit: change tests to match new version too?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -14,6 +51,18 @@ mod error {
...
+            "Expected ecr helper to be called with either 'registry' or 'region', got '{}'",
+            value,
+        ))]
+        EcrParam { value: String },
+
+        #[snafu(display("Expected an AWS region, got '{}' in template {}", value, template))]
+        EcrRegion {
+            value: handlebars::JsonValue,
+            template: String,
+        },
```
tjkirch, line 64: This is currently in the refactor commit.

tjkirch (approved):

webern:
>  webern force-pushed the webern:ecr-repos branch from 356423d to a2cba99
> webern force-pushed the webern:ecr-repos branch from a2cba99 to dc87e69

Addresses @tjkirch comments: fixup refactor commit to remove unrelated errors, use 0.5.1 in the test string to make it more clear when this change took place.

tjkirch (approved):

webern:
> webern force-pushed the webern:ecr-repos branch from dc87e69 to 92a8348

Rebase, resolve conflict for admin container version bump.

webern:
> webern force-pushed the webern:ecr-repos branch from 92a8348 to 783374d

Missed a spot with the admin container bump.

Release.toml
```diff
@@ -7,3 +7,5 @@ version = "0.5.0"
 "(0.3.4, 0.4.0)" = []
 "(0.4.0, 0.4.1)" = ["migrate_v0.4.1_add-version-lock-ignore-waves.lz4", "migrate_v0.4.1_pivot-repo-2020-07-07.lz4"]
 "(0.4.1, 0.5.0)" = ["migrate_v0.5.0_add-cluster-domain.lz4", "migrate_v0.5.0_admin-container-v0-5-2.lz4", "migrate_v0.5.0_control-container-v0-4-1.lz4"]
+# TODO - add ecr-helper-admin and ecr-helper-control migrations to next release.
+# NOTE - these expect container versions v0.5.1 and v0.4.1
```
tjkirch: Is this 0.5.2 now?

webern: Good catch, thanks.

webern:
> webern force-pushed the webern:ecr-repos branch from 783374d to b0944cf 

Bump the container version in a comment that was missed.

webern:
Finished testing, exiting draft status...

etungsten (approved):
:whale:

tjkirch (approved):

bcressey (approved):

PR #1033, opened 2020-08-12
Add `cargo make ami-public` and `ami-private` targets

zmrow:
**Description of changes:**
```
dc4d53b5 Add `cargo make ami-public` and `ami-private` targets
1d806774 pubsys: Move region_from_string to common aws module
e0988d67 Allow JSON output of registered AMIs
293d5b24 rustfmt pubsys client module
```

**Testing done:**
* Made multiple AMIs public and private
  * @tjkirch launched an AMI made public from @zmrow 's account
  * Checked permissions on all AMIs before and after `cargo make ami-public/private`
* Verified that a missing AMI will cause an error

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/publish_ami/mod.rs
+++ b/tools/pubsys/src/aws/publish_ami/mod.rs
@@ -1,0 +1,459 @@
+//! The publish_ami module owns the 'publish-ami' subcommand and controls the process of granting
+//! and revoking public access to EC2 AMIs.
+
+use crate::aws::{client::build_client, region_from_string};
+use crate::config::InfraConfig;
+use crate::Args;
+use futures::future::{join, ready};
+use futures::stream::{self, StreamExt};
+use log::{debug, error, info, trace};
+use rusoto_core::{Region, RusotoError};
+use rusoto_ec2::{
+    DescribeImagesError, DescribeImagesRequest, DescribeImagesResult, Ec2, Ec2Client,
+    ModifyImageAttributeError, ModifyImageAttributeRequest, ModifySnapshotAttributeError,
+    ModifySnapshotAttributeRequest,
+};
+use snafu::{ensure, OptionExt, ResultExt};
+use std::collections::{HashMap, HashSet};
+use std::fs::File;
+use std::iter::FromIterator;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+/// Grants or revokes permissions to Bottlerocket AMIs
+#[derive(Debug, StructOpt)]
+#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+#[structopt(group = clap::ArgGroup::with_name("mode").required(true).multiple(false))]
+pub(crate) struct PublishArgs {
+    /// Path to the JSON file containing regional AMI IDs to modify
+    #[structopt(long)]
+    ami_input: PathBuf,
+
+    /// Comma-separated list of regions to publish in, overriding Infra.toml; given regions must be
+    /// in the --ami-input file
+    #[structopt(long, use_delimiter = true)]
+    regions: Vec<String>,
+
+    /// Make the AMIs public
+    #[structopt(long, group = "mode")]
+    make_public: bool,
+    /// Make the AMIs private
+    #[structopt(long, group = "mode")]
+    make_private: bool,
+}
+
+/// Common entrypoint from main()
+pub(crate) async fn run(args: &Args, publish_args: &PublishArgs) -> Result<()> {
+    let (operation, mode) = if publish_args.make_public {
+        ("add".to_string(), "public")
+    } else if publish_args.make_private {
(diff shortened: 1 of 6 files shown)


tjkirch (approved):

tjkirch:
^ This push rebases to fix the conflict from #1034, and rewords the commit messages per @bcressey.

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -0,0 +1,419 @@
...
+            path: &publish_args.ami_input
+        }
+    );
+
+    // Parse region names, adding endpoints from InfraConfig if specified
+    let mut amis = HashMap::with_capacity(ami_input.len());
+    for (name, ami_id) in ami_input {
+        let region = region_from_string(&name, &aws).context(error::ParseRegion)?;
+        amis.insert(region, ami_id);
+    }
```
bcressey: Can we add another check that all the regions we expect from `Infra.toml` are present in the input file?

tjkirch:
^ Addresses @bcressey's comment above; per offline discussion, rather than an exact match, we check that the requested regions are a subset of the ones in the JSON file, in case a user only wants to publish in some of their regions at first.

bcressey (approved):

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+#[structopt(group = clap::ArgGroup::with_name("mode").required(true).multiple(false))]
+pub(crate) struct PublishArgs {
+    /// Path to the JSON file containing regional AMI IDs to modify
+    #[structopt(long)]
+    ami_input: PathBuf,
+
+    /// List of regions to publish in, overriding Infra.toml; must match regions in --ami-input
+    /// file.
+    #[structopt(long, use_delimiter = true)]
+    regions: Vec<String>,
```
webern: Seems strange. If it must match the --ami-input file, then why not just a bool to use the regions listed in that file (instead of infra.toml)?

Edit: from below, I think it's just slightly unclear that it means 'must match or be a subset of', or 'all regions must exist in the --ami-input file'.

```
    /// List of regions to publish in, overriding Infra.toml; must match regions in --ami-input
    /// file.
    #[structopt(long, use_delimiter = true)]
    regions: Vec<String>,
```

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+
+    /// Make the AMIs public
+    #[structopt(long, group = "mode")]
+    make_public: bool,
+    /// Make the AMIs private
+    #[structopt(long, group = "mode")]
```
webern, line 41: Nit: matches the spacing of the other arguments.
```suggestion
    make_public: bool,
    
    /// Make the AMIs private
    #[structopt(long, group = "mode")]
```

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -0,0 +1,459 @@
...
+                ec2_client.modify_snapshot_attribute(modify_snapshot_request);
+
+            // Store the region and snapshot ID so we can include it in errors
+            let info_future = ready((region.name().to_string(), snapshot_id.clone()));
+            modify_snapshot_requests.push(join(info_future, modify_snapshot_future));
+        }
+    }
+
+    // Send requests in parallel and wait for responses, collecting results into a list.
+    let request_stream = stream::iter(modify_snapshot_requests).buffer_unordered(4);
```
webern, line 259: Not 3 or 5? :)

```
    let request_stream = stream::iter(modify_snapshot_requests).buffer_unordered(4);
```

webern (commented):
Nice!

tools/pubsys/src/aws/publish_ami/mod.rs
tjkirch: Yeah, looks like we forgot to change this comment after changing to the subset strategy.

tools/pubsys/src/aws/publish_ami/mod.rs
tjkirch: public/private are tied and should be read together, that's why we kept them together without a blank line.

tools/pubsys/src/aws/publish_ami/mod.rs
tjkirch: There's a much longer comment explaining this in `ami/mod.rs`, because it matters a lot more there; here, we just need to turn the stream iter into something we can wait on, and you have to pick a number, and it really doesn't matter because these requests are so quick.  :|

tjkirch:
^ This push addresses @webern's comment about the inaccurate description of the regions parameter.  I *think* the other comments are OK.

webern (approved):

Issue #1039, opened 2020-08-13
logdog: support the ecs variant


webern:
`logdog` needs to output different information for `ecs` than what is currently hard-coded.

Small complication:

Currently `logdog` assumes that everything we need can be collected by running a shell command with `std::process::Command`. However, the ECS requires something like `curl localhost:51678/v1/tasks` and we don't ship curl on Bottlerocket. A redesign of logdog's abstractions will be needed.

tjkirch:
If we move inputs to a file rather than being hardcoded, we can structure them a bit more.  Straw-man: still plain text, but prefixed with a "mode", like `sh echo hi` to get the output of a shell command and `http localhost:51678/v1/tasks` to get the body of an HTTP response.

PR #1042, opened 2020-08-14
logdog: support ecs variant

webern:
**Issue number:**

#815

**Description of changes:**

Conditionally compile a different set of `logdog` commands for ECS.

**Testing done:**

Ran an ECS variant, got the ECS logs.
Ran a k8s variant, got the k8s logs.

** More work to be Done **

I wanted to get the low hanging fruit handled now, but I still need to do #1039 which requires a more substantial change in `logdog`.

**Terms of contribution:**

### DIFF
--- a/sources/logdog/build.rs
+++ b/sources/logdog/build.rs
@@ -5,7 +5,20 @@ use std::fs::File;
 use std::io::Write;
 use std::path::PathBuf;
 
+const VARIANT_GROUPS: &[&str] = &["aws-k8s", "aws-ecs"];
+
 fn main() {
+    println!("cargo:rerun-if-env-changed=VARIANT");
+    if let Ok(variant) = env::var("VARIANT") {
+        println!("cargo:rustc-cfg=variant=\"{}\"", variant);
+        for &variant_group in VARIANT_GROUPS {
+            if variant.starts_with(variant_group) {
+                println!("cargo:rustc-cfg=variant_group=\"{}\"", variant_group);
+                break;
+            }
+        }
+    }
+
     // Check for environment variable "SKIP_README". If it is set,
     // skip README generation
     if env::var_os("SKIP_README").is_some() {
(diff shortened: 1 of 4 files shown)


sources/logdog/src/log_request.rs
```diff
@@ -6,24 +6,85 @@ pub(crate) struct LogRequest<'a> {
...
+            ("signpost", "signpost status"),
+            ("wicked", "wicked show all"),
+            // TODO - https://github.com/bottlerocket-os/bottlerocket/issues/1039
+            // ("ecs-tasks", "curl localhost:51678/v1/tasks"),
+        ]
+        .iter()
+        .map(|(filename, command)| LogRequest { filename, command })
+        }
+    }
+    else {
```
tjkirch, line 64: There's a lot of repetition here, and it's not the kind that's easy to visually compare and verify.  Can we store common entries in a common table and append them to the variant-specific ones, maybe with `chain` or similar on the iterator?

sources/logdog/build.rs
```diff
@@ -5,7 +5,20 @@ use std::fs::File;
...
 use std::path::PathBuf;
 
+const VARIANT_GROUPS: &[&str] = &["aws-k8s", "aws-ecs"];
+
 fn main() {
+    println!("cargo:rerun-if-env-changed=VARIANT");
+    if let Ok(variant) = env::var("VARIANT") {
+        println!("cargo:rustc-cfg=variant=\"{}\"", variant);
+        for &variant_group in VARIANT_GROUPS {
+            if variant.starts_with(variant_group) {
```
tjkirch, line 15: I'd rather avoid the idea of variant groups/families until we have time for more thought and design on it.  I think we can do a relatively simple change at the buildsys level that will enable this information for all builds, not just this one, and possibly with a better, more granular mechanism like features/labels for k8s, aws, etc.  Or, at the least, move the decision out of the crate and into a place that already has to care about variant, for example using a data file to list the commands we run and installing a different data file per variant in the os package, or something similar.  In my mind, this feature isn't critical for GA, and waiting is a better option than a one-off, given the complexity involved.

webern:
Closing in favor of a different approach.

PR #1043, opened 2020-08-14
ecs-agent: upgrade to v1.43.0

samuelkarp:
**Issue number:**

**Description of changes:**
Upgrade to the current version of the ECS agent.  One patch removed, the other two refreshed.

**Testing done:**
Built an AMI, ran a task.

**Terms of contribution:**


bcressey (approved):

packages/ecs-agent/0002-bottlerocket-version-values-settable-with-linker.patch
```diff
@@ -1,30 +1,30 @@
...
 --- a/agent/version/version.go
 +++ b/agent/version/version.go
 @@ -22,10 +22,10 @@ package version
  // repository. Only the 'Version' const should change in checked-in source code
  
  // Version is the version of the Agent
--const Version = "1.41.0"
-+var Version = "1.41.0"
+-const Version = "1.43.0"
++var Version = ""
```
tjkirch, line 19: Neat, what happened so that we don't need this anymore?

tjkirch (approved):

packages/ecs-agent/0002-bottlerocket-version-values-settable-with-linker.patch
samuelkarp: We never actually needed the value in the patch before, since we're overriding all the values via `-ldflags`.  I had previously left the version in because I had assumed it would make it easier to refresh the patch going forward, but it just caused conflicts; I'm expecting that this patch is _still_ going to cause conflicts but at least it's more clear that we don't need to preserve the value in the patch itself.

PR #1045, opened 2020-08-15
Fix rust build caching and tune rpm compression

bcressey:
**Issue number:**
N/A

**Description of changes:**
Use a separate cache for gnu and musl targets, so that they don't overwrite each other. Use fixed `rustc` flags to avoid letting `cargo` pick, which can invalidate the musl cache between builds.

Disable debuginfo compression since we don't need it. Optimize the compression of the cpio archive inside the rpms for our use case.

**Testing done:**
While hacking on `growpart` I noticed that each build was taking over five minutes which was pretty painful for iterative testing.

Changing the Rust macros to better preserve the cache dropped the time to two minutes, while adjusting the compression for debuginfo and cpio archives dropped it to one minute. I verified that all crates are marked as "fresh" in the output for subsequent runs.

Rebuild time when switching variants is now around five minutes. 

**Terms of contribution:**

### DIFF
--- a/macros/cargo
+++ b/macros/cargo
@@ -16,9 +16,10 @@
 %__cargo_cross_opts %{__cargo_common_opts} --target %{__cargo_target}
 %__cargo_cross_opts_static %{__cargo_common_opts} --target %{__cargo_target_static}
 %__cargo_env CARGO_TARGET_DIR="${HOME}/.cache" SKIP_README="true"
+%__cargo_env_static CARGO_TARGET_DIR="${HOME}/.cache/.static" SKIP_README="true"
 %__cargo_cross_pkg_config PKG_CONFIG_PATH="%{_cross_pkgconfigdir}" PKG_CONFIG_ALLOW_CROSS=1
 %__cargo_cross_env %{__cargo_env} %{__cargo_cross_pkg_config} TARGET_CC="%{_cross_triple}-gnu-gcc"
-%__cargo_cross_env_static %{__cargo_env} %{__cargo_cross_pkg_config} TARGET_CC="%{_cross_triple}-musl-gcc" CARGO_PROFILE_RELEASE_LTO=thin
+%__cargo_cross_env_static %{__cargo_env_static} %{__cargo_cross_pkg_config} TARGET_CC="%{_cross_triple}-musl-gcc"
 
 %cargo_prep (\
 %{__mkdir} -p %{_builddir}/.cargo \
--- a/macros/shared
+++ b/macros/shared
@@ -222,6 +222,13 @@ CROSS_COMPILATION_CONF_EOF\
 # failures when rewriting "shebangs" in crate files.
 %__brp_mangle_shebangs /bin/true
 
+# Do not compress debuginfo.
+%_find_debuginfo_dwz_opts %{nil}
+
+# Compress RPM payloads with zstd level 1 in single-threaded mode.
+%_source_payload w1T0.zstdio
+%_binary_payload w1T0.zstdio
+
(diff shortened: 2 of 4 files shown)


samuelkarp (approved):

macros/shared
```diff
@@ -222,6 +222,13 @@ CROSS_COMPILATION_CONF_EOF\
 # failures when rewriting "shebangs" in crate files.
 %__brp_mangle_shebangs /bin/true
 
+# Do not compress debuginfo.
+%_find_debuginfo_dwz_opts %{nil}
+
+# Compress RPM payloads with zstd level 19 using 9 threads.
+%_source_payload w19T9.zstdio
+%_binary_payload w19T9.zstdio
```
tjkirch: Can/should we use T0 to match the number of CPU cores instead?  The zstd CLI supports that, but I'm not sure if RPM passes it through directly; it seems to?

Out of curiosity, did you try different compression levels?  19 is near the max, and (naively) it seems like .rpm transfer out wouldn't be as slow as the CPU hit from near-max compression.

macros/cargo
```diff
@@ -16,9 +16,10 @@
 %__cargo_cross_opts %{__cargo_common_opts} --target %{__cargo_target}
 %__cargo_cross_opts_static %{__cargo_common_opts} --target %{__cargo_target_static}
 %__cargo_env CARGO_TARGET_DIR="${HOME}/.cache" SKIP_README="true"
+%__cargo_env_static CARGO_TARGET_DIR="${HOME}/.cache/static" SKIP_README="true"
```
tjkirch: I'm slightly worried about the possibility of cargo using a directory named "static" under its target dir someday, though it's safe for now.  ".cache_static" to avoid any worry?

Similarly, we know skipping readme builds here is safe because we're only calling it for two things after our normal builds are done, but that may not always be true.  How about at least adding a comment here, like "# Take advantage of the fact that cargo_build_static is only used after cargo_build, so we don't need to rebuild READMEs" ?

tjkirch (approved):

macros/cargo
bcressey: We skip READMEs because the build scripts will attempt to write them to the same directory as `build.rs` and `Cargo.toml`, which is immutable at this point.

We have to use a subdirectory of `.cache` or else we'd need a second cache layer to wire up, which feels excessive. Maybe `.cache/.static` instead?

macros/shared
bcressey: The "threads" value is set via `ZSTD_CCtx_setParameter(_stream, ZSTD_c_nbWorkers, threads)`. The zstd manual says that 0 is the default value and means single-threaded mode.

I experimented with compression levels from 0 to 19 in steps of 5 or so. "0" was the fastest, and all the others were approximately one second slower.

I've now experimented with threads and found that single-threaded mode seems to be the fastest.

macros/cargo
tjkirch: > We skip READMEs because the build scripts will attempt to write them to the same directory as build.rs and Cargo.toml, which is immutable at this point.

Sorry, I was silly and blind, it's also on the line above. :)

> We have to use a subdirectory of .cache or else we'd need a second cache layer to wire up, which feels excessive. Maybe .cache/.static instead?

Ah, right, I forgot about the cache layer being mounted there.  I don't think it's a big risk, but we could be more specific if you want.  `.static`, or `.buildsys_static` to be fully safe perhaps.  I'm also totally fine with it as-is.

tjkirch (approved):

webern (approved):

PR #1051, opened 2020-08-19
aws-ecs-1: constrain ephemeral port range

samuelkarp:
**Issue number:**

**Description of changes:**
ECS documents a smaller ephemeral port range than is default on Bottlerocket.  Constraining our range to the documented ECS range should avoid surprising customers who expect ECS's documented range and have configured their security groups accordingly.

**Testing done:**
* Ran containers with `docker run` locally and saw the correct ephemeral port range being used.
* Ran a task with `"portMappings": [{"containerPort": 80}]` and saw the correct ephemeral port range being used.

**Terms of contribution:**

### DIFF
--- a/packages/ecs-agent/ecs-sysctl.conf
+++ b/packages/ecs-agent/ecs-sysctl.conf
@@ -1,2 +1,6 @@
 # ECS task role endpoint
 net.ipv4.conf.all.route_localnet = 1
+
+# Constrain ephemeral ports to the range documented for ECS
+# https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html
+net.ipv4.ip_local_port_range = 32768 60999


bcressey (approved):

webern (approved):

jahkeup (approved):

packages/ecs-agent/ecs-sysctl.conf
```diff
@@ -1,2 +1,6 @@
 # ECS task role endpoint
 net.ipv4.conf.all.route_localnet = 1
+
+# Constrain ephemeral ports to the range documented for ECS
+# https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html
+net.ipv4.ip_local_port_range = 32768 60999
```
jahkeup, line 6: Didn't see `60999` referenced in the linked docs, is `60999` a value that was chosen over`65535`?

samuelkarp: I copied this without noticing it was different from the ECS documentation, but 60999 is the upper bound I see on Amazon Linux, Debian, and Ubuntu.  I can add that to the comments if you want.

jahkeup: I think we should keep this aligned with documented and current usages of the constrained port range: so yes, I'd prefer to see this as `65535`. 

An example: [the current port range used by the CDK is 32768-65535](https://github.com/aws/aws-cdk/blob/366c7819321dfd23d4ce638fb218fe38072e90aa/packages/%40aws-cdk/aws-ecs/lib/base/task-definition.ts#L505-L508) (though this [range might be opened up in the near future](https://github.com/aws/aws-cdk/issues/9778)).

jahkeup: 1. Discussed offline with @samuelkarp : common distributions are using the kernel's default of `60999`
2. digging into the [kernel history on `60999`](https://github.com/torvalds/linux/commit/07f4c90062f8fc7c8c26f8f95324cbe8fa3145a5) (and the [move to `61000`](https://github.com/torvalds/linux/commit/3f196eb519a419bf83ecc22753943fd0a0de4f8f)) show no concerns surrounding `65535` nor `60999` (other than for port assignment efficacy)
3. I'm on board with leaving this at `60999` :+1:

PR #1054, opened 2020-08-20
Logdog ecs

### LINKED ISSUE #1039: logdog: support the ecs variant

webern:
**Issue number:**

Closes #1039

**Description of changes:**

Statically compile variant-specific logdog behavior by reading from two text files at build time. A symlink is used to point to the correct file for variant-specific commands.

**Testing done:**

Ran an ecs variant, got ecs logs. Ran a k8s variant, got k8s logs.

**Terms of contribution:**


sources/logdog/static/aws-ecs-1
```diff
@@ -0,0 +1,5 @@
+http get http://localhost:51678/v1/tasks
```
webern:
```suggestion
http ecs-tasks http://localhost:51678/v1/tasks
```

sources/logdog/src/commands.rs
```diff
@@ -0,0 +1,154 @@
...
+        .lines()
+        .filter(|&command| !command.is_empty())
+        .collect();
+    commands.append(&mut variant_commands);
+    commands
+}
+
+/// A logdog `Command` is in the format `mode filename [args...]`. `mode` specifies what type of
+/// command it is, e.g. `sh` for a shell command or `http` for an HTTP get request. `filename` is
+/// the name of the output file. `args` is any number of shell-like arguments. Putting this all
```
tjkirch:
```suggestion
/// the name of the output file. `args` is any number of shell-like arguments, for `sh`, or a URL, for `http`. Putting this all
```
I guess "shell-like" doesn't apply to the http arguments...

sources/logdog/Cargo.toml
```diff
@@ -10,10 +10,12 @@ exclude = ["README.md"]
 
 [dependencies]
 flate2 = "1.0"
+reqwest = { version = "0.10.1", default-features = false, features = ["blocking"] }
```
tjkirch: Should we add `rustls-tls` to the feature list to support https URLs?

samuelkarp (approved):

sources/logdog/static/common
```diff
@@ -0,0 +1,15 @@
+sh containerd-config containerd config dump
```
tjkirch: nit, because it's really a question about the existing command: do we need to do something to talk to the host-containerd specifically, rather than the containerd used for k8s?

sources/logdog/static/aws-ecs-1
```diff
@@ -0,0 +1,5 @@
...
+sh docker-daemon.json cat /etc/docker/daemon.json
+sh docker-info docker info
+sh ecs-agent-state.json cat /var/lib/ecs/data/ecs_agent_data.json
+sh ecs-config.json cat /etc/ecs/ecs.config.json
```
tjkirch: Maybe a `file` mode to handle the common case of copying files like this?

sources/logdog/build.rs
```diff
@@ -30,3 +39,36 @@ fn main() {
...
+/// Creates a file named `variant` which is a symlink to a file with `logdog` commands for the
+/// current variant. Whatever the value of the `VARIANT` environment variable is, this function
+/// requires a file of the same name to exist in ./static, and points to it from a symlink.
+fn symlink_variant() {
+    let variant = env::var(ENV_VARIANT).unwrap_or_else(|_| {
+        eprintln!("For local builds, you must set the {} environment variable so we know which logdog commands to \
+        build. Valid values are the directories in models/src/variants/, for example 'aws-k8s-1.17'.", ENV_VARIANT);
+        process::exit(1);
+    });
+    if !PathBuf::from(format!("static/{}", variant)).is_file() {
```
tjkirch: I'm a little confused about this directory; let me suggest something...

We make `sources/logdog/etc` (or `conf` or something if we hate `etc`) where we keep our variant-specific files, which we'd name `logdog.VARIANT.conf`.  We have our writable directory that we mount in via the Dockerfile, and we put it at `soures/logdog/etc/current`, and we symlink `current/logdog.conf` -> `../logdog.VARIANT.conf`.  Then I think we don't need the copies in the Dockerfile, and (imo) it turns out a little simpler.  (And we could still have `sources/logdog/etc/logdog.common.conf` or similar.)

sources/logdog/build.rs
```diff
@@ -1,11 +1,20 @@
...
+// The VARIANT variable is originally BUILDSYS_VARIANT, set in the top-level Makefile.toml,
+// and is passed through as VARIANT by the top-level Dockerfile.  It represents which OS
+// variant we're building, and therefore which API model to use.
+const ENV_VARIANT: &str = "VARIANT";
 
 fn main() {
+    println!("cargo:rerun-if-env-changed={}", ENV_VARIANT);
+    symlink_variant();
+
     // Check for environment variable "SKIP_README". If it is set,
```
tjkirch: nit: can we do like `models` and put variant/readme code in separate functions, and just call those from main?

Dockerfile
```diff
@@ -79,6 +81,7 @@ USER builder
 RUN --mount=source=.cargo,target=/home/builder/.cargo \
     --mount=type=cache,target=/home/builder/.cache,from=cache,source=/cache \
     --mount=type=cache,target=/home/builder/rpmbuild/BUILD/sources/models/src/variant,from=variantcache,source=/variantcache \
+    --mount=type=cache,target=/home/builder/rpmbuild/BUILD/sources/logdog/static,from=variantcache,source=/variantcache/logdog \
```
tjkirch: I think we should consider doing `/variantcache/models` like you did with `/variantcache/logdog` so we don't have any potential interaction between the two crates' storage.

tjkirch (commented):
I haven't been through the main Rust code yet, but these are my comments about the outlying bits.

sources/logdog/build.rs
webern: How's this?

```
├── .gitignore # (ignore conf/current/)
├── Cargo.toml
├── README.md
├── README.tpl
├── build.rs
├── conf
│   ├── aws-k8s.conf
│   ├── current
│   │   ├── logdog.common.conf -> ../logdog.common.conf
│   │   └── logdog.conf -> ../logdog.aws-k8s-1.17.conf
│   ├── logdog.aws-dev.conf
│   ├── logdog.aws-ecs-1.conf
│   ├── logdog.aws-k8s-1.15.conf -> aws-k8s.conf
│   ├── logdog.aws-k8s-1.16.conf -> aws-k8s.conf
│   ├── logdog.aws-k8s-1.17.conf -> aws-k8s.conf
│   └── logdog.common.conf
└── src
    ├── commands.rs
    ├── create_tarball.rs
    ├── error.rs
    └── main.rs
```

webern:
> webern force-pushed the webern:logdog-ecs branch from 740b6eb to 6eb611d

Rebase to fix the Dockerfile conflict, no other changes yet.

sources/logdog/build.rs
tjkirch: Only logdog itself needs to read these files, so I don't think the common file needs to be linked into the writable directory, it'd just complicate build.rs.  Looks good though. :+1:

webern: Oh yeah, good point, thanks.

sources/logdog/static/common
webern: I'm not sure, both of these commands have the exact same output in sudo sheltie:

```
containerd --address /run/host-containerd/containerd.sock config dump
containerd --address /run/containerd/containerd.sock config dump
```

Both of these report:

```
version = 2
root = "/var/lib/containerd"
state = "/run/containerd"
plugin_dir = ""
disabled_plugins = ["io.containerd.internal.v1.opt", "io.containerd.snapshotter.v1.aufs", "io.containerd.snapshotter.v1.devmapper", "io.containerd.snapshotter.v1.native", "io.containerd.snapshotter.v1.zfs"]
required_plugins = []
oom_score = 0
imports = ["/etc/containerd/config.toml"]

[grpc]
  address = "/run/dockershim.sock"
  tcp_address = ""
  tcp_tls_cert = ""
  tcp_tls_key = ""
  uid = 0
  gid = 0
  max_recv_message_size = 16777216
  max_send_message_size = 16777216
etc...
```

It seems like both of these report the non-host-containerd config (not sure though). I'm hoping we can avoid blocking on this.

samuelkarp: You need `containerd --config /etc/host-containerd/config.toml config dump` to get the host-containerd config data.

webern:
> force-pushed the webern:logdog-ecs branch from 6eb611d to afdc876

Change the dockerfile and directory structure of the command files. Also fix a couple of smaller issues.

sources/logdog/static/aws-ecs-1
webern: [afdc876](https://github.com/bottlerocket-os/bottlerocket/compare/6eb611d17506ab4aad98e5db4cd17b00bacd0839..afdc8765a9e97bb1d4436eda9b0c964fce24ab95)

sources/logdog/src/commands.rs
webern: [afdc876](https://github.com/bottlerocket-os/bottlerocket/compare/6eb611d17506ab4aad98e5db4cd17b00bacd0839..afdc8765a9e97bb1d4436eda9b0c964fce24ab95)

sources/logdog/Cargo.toml
webern: [afdc876](https://github.com/bottlerocket-os/bottlerocket/compare/6eb611d17506ab4aad98e5db4cd17b00bacd0839..afdc8765a9e97bb1d4436eda9b0c964fce24ab95)

sources/logdog/build.rs
webern: [afdc876](https://github.com/bottlerocket-os/bottlerocket/compare/6eb611d17506ab4aad98e5db4cd17b00bacd0839..afdc8765a9e97bb1d4436eda9b0c964fce24ab95)

Dockerfile
webern: [afdc876](https://github.com/bottlerocket-os/bottlerocket/compare/6eb611d17506ab4aad98e5db4cd17b00bacd0839..afdc8765a9e97bb1d4436eda9b0c964fce24ab95)

webern:
> webern force-pushed the webern:logdog-ecs branch from afdc876 to 7d1f985

Oops, create the ignored output directory if it does not exist.

sources/logdog/conf/logdog.aws-ecs-1.conf
```diff
@@ -0,0 +1,5 @@
+http ecs-tasks http://localhost:51678/v1/tasks
+sh docker-daemon.json cat /etc/docker/daemon.json
+sh docker-info docker info
+sh ecs-agent-state.json cat /var/lib/ecs/data/ecs_agent_data.json
+sh ecs-config.json cat /etc/ecs/ecs.config.json
```
bcressey: Can we do `exec` instead of `sh` since we don't have a shell?

sources/logdog/src/commands.rs
```diff
@@ -0,0 +1,154 @@
...
+        .lines()
+        .filter(|&command| !command.is_empty())
+        .collect();
+    let mut variant_commands: Vec<&str> = VARIANT_COMMANDS
+        .lines()
+        .filter(|&command| !command.is_empty())
+        .collect();
+    commands.append(&mut variant_commands);
+    commands
+}
```
bcressey: You could also filter out lines that start with `#` so we could add comments for the commands explaining the relevance.

Dockerfile
```diff
@@ -35,7 +35,8 @@ ARG ARCH
 ARG VARIANT
 ARG TOKEN
 # We can't create directories via RUN in a scratch container, so take an existing one.
-COPY --chown=1000:1000 --from=sdk /tmp /variantcache
+COPY --chown=1000:1000 --from=sdk /tmp /variantcache/model
+COPY --chown=1000:1000 --from=sdk /tmp /variantcache/logdog
```
tjkirch: I know I suggested the subdirectories, but Ben told me that each usage of variantcache should actually be a new copy of the layer based on the earlier definition, and so subdirectories shouldn't actually be needed.  I'm not sure how to test that (`ls` in the models and logdog build scripts?) but it'd be cleaner not to use the subdirectories if not needed...

sources/logdog/.gitignore
```diff
@@ -0,0 +1,2 @@
+# ignore the symlink generated by build.rs
+conf/current/
```
tjkirch:
```suggestion
/conf/current/
```
nit: It shouldn't actually make a difference here, but the starting slash at least makes clear to future readers that this directory lives at the top level.

sources/logdog/build.rs
```diff
@@ -30,3 +39,53 @@ fn main() {
...
+        process::exit(1);
+    }
+    // create the conf/current directory if it is missing
+    fs::create_dir_all("conf/current").unwrap();
+    // create the symlink from conf/current/logdog.conf to the variant-specific file
+    symlink_force(
+        format!("../{}", variant_filename),
+        "conf/current/logdog.conf",
+    )
+    .unwrap();
```
tjkirch: Can we give error messages the way `models` does rather than unwrapping?  I don't think the unwrap message will be very helpful, it usually doesn't include a path, and it won't explain why were trying to do this.

sources/logdog/build.rs
```diff
@@ -30,3 +39,53 @@ fn main() {
...
+        eprintln!(
+            "There is no file named '{}' in the 'conf' directory for the current variant (given \
+            by the '{}' environment variable) Each variant must have a file representing the \
+            variant-specific commands that logdog will run.",
+            variant, ENV_VARIANT
+        );
+        process::exit(1);
+    }
+    // create the conf/current directory if it is missing
+    fs::create_dir_all("conf/current").unwrap();
```
tjkirch: I think this might be simpler the way `models` does it, where there's an empty `.keep` file in the directory where we mount the writable cache (`conf/current` here) rather than needing to make the directory here.

sources/logdog/src/error.rs
```diff
@@ -3,92 +3,164 @@
...
+
+    #[snafu(display("The output directory '{}' could not be created: {}.", path.display(), source))]
+    CreateOutputDirectory {
         source: io::Error,
         path: PathBuf,
         backtrace: Backtrace,
     },
+
+    #[snafu(display("Encountered an empty command."))]
+    EmptyCommand { backtrace: Backtrace },
```
tjkirch: It seems this is unused now?

sources/logdog/src/error.rs
```diff
@@ -3,92 +3,164 @@
...
+    },
+
+    #[snafu(display("Unable to read config file '{}': {}", path.display(), source))]
+    ConfigRead { source: io::Error, path: PathBuf },
+
+    #[snafu(display("Config file is not valid UTF-8 '{}': {}", path.display(), source))]
+    ConfigUtf8 {
+        source: std::str::Utf8Error,
+        path: PathBuf,
+    },
```
tjkirch: These seem unused?

sources/logdog/src/main.rs
```diff
@@ -18,22 +18,20 @@ For the commands used to gather logs, please see [log_request](src/log_request.r
...
 use std::path::{Path, PathBuf};
-use std::process::{Command, Stdio};
 use std::{env, process};
-
-use create_tarball::create_tarball;
-use error::Result;
-use log_request::{log_requests, LogRequest};
-use snafu::{ErrorCompat, OptionExt, ResultExt};
 use tempfile::TempDir;
+use From;
```
tjkirch: I've never seen something like this before, what is this doing?  Is it referencing `std::convert::From`?  It has to already be in scope to reference it like this, so I'm not sure what the additional `use` would do...

sources/logdog/src/commands.rs
```diff
@@ -0,0 +1,154 @@
...
+
+/// Runs a `logdog` command and writes its output to a file in `tempdir`.
+pub(crate) fn handle_command<S, P>(request: S, tempdir: P) -> Result<()>
+where
+    S: AsRef<str>,
+    P: AsRef<Path>,
+{
+    let request = request.as_ref();
+    let tokens = shell_words::split(request).context(error::CommandParse { command: request })?;
```
tjkirch: I'm not sure we want to split all command types using shell rules; it might not have the intended effect for other modes.  Can we do this inside `handle_shell_request` like we used to in `run_command`?

You could use `request.splitn(3, ' ')` to get an iterator of the mode, filename, and whatever's left, and use `.next().context(bla)` twice to set the mode and filename, and `.collect()` to get whatever's left into a string to pass to the handler, where `shell_words.split` (or whatever's appropriate for the mode) would come into play.

sources/logdog/src/commands.rs
```diff
@@ -0,0 +1,154 @@
...
+    args: &'a [String],
+}
+
+/// This is used in error construction.
+impl ToString for Command<'_> {
+    fn to_string(&self) -> String {
+        if self.args.is_empty() {
+            format!("{} {}", self.mode, self.filename)
+        } else {
+            format!("{} {} {}", self.mode, self.filename, self.args.join(" "))
```
tjkirch: If we use shell_words to split, we should probably use it to join as well?

sources/logdog/src/commands.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/// A logdog `Command` is in the format `mode filename [args...]`. `mode` specifies what type of
+/// command it is, e.g. `sh` for a shell command or `http` for an HTTP get request. `filename` is
+/// the name of the output file. `args` is any number of shell-like arguments, for `sh`, or a
+/// URL, for `http`. Putting this all together, `sh hello.txt echo hello world` will run `echo
+/// hello world` and write the output to `hello.txt`. `http example http://example.com` will GET
+/// `http://example.com` and write the response body to `example`.
+#[derive(Debug, Clone)]
+struct Command<'a> {
```
tjkirch:
```suggestion
struct LogRequest<'a> {
```
So that people used to `std::process::Command` don't get confused, and we don't need the full referencing of `std::process::Command` below.  And this was the terminology being used elsewhere :)

sources/logdog/src/commands.rs
```diff
@@ -0,0 +1,154 @@
...
+    let mut commands: Vec<&str> = COMMON_COMMANDS
+        .lines()
+        .filter(|&command| !command.is_empty())
+        .collect();
+    let mut variant_commands: Vec<&str> = VARIANT_COMMANDS
+        .lines()
+        .filter(|&command| !command.is_empty())
+        .collect();
+    commands.append(&mut variant_commands);
+    commands
```
tjkirch:
```suggestion
    COMMON_COMMANDS
        .lines()
        .chain(VARIANT_COMMANDS
        .lines())
        .filter(|&command| !command.is_empty() && !command.trim_start().starts_with('#'))
        .collect()
```
Shouldn't need to construct/expand two vecs and filter twice.

(Also fixes Ben's request)

sources/logdog/src/commands.rs
```diff
@@ -0,0 +1,154 @@
...
+            .context(error::IncompleteRequest { request })?,
+        args: if tokens.len() > 2 { &tokens[2..] } else { &[] },
+    };
+    match req.mode {
+        "sh" => handle_shell_request(&req, tempdir)?,
+        "http" => handle_http_request(&req, tempdir)?,
+        unmatched => {
+            return Err(error::Error::UnhandledRequest {
+                mode: unmatched.into(),
+                request: request.into(),
```
tjkirch: I don't think you need the manual `into` calls here; snafu should convert into owned strings if needed by the error variant, and only if there was a failure.

sources/logdog/src/commands.rs
```diff
@@ -0,0 +1,154 @@
+//! Provides the list of commands that `logdog` will run and provides the handler function to run
+//! them.
+//!
+//! # Static Commands
+//!
+//! At build time, the `build.rs` file checks which variant is being built and creates a symlink
+//! file which points to the commands for the current variant. This file is named `variant`. We load
+//! the `variant` and `common` files into static strings at compile time, and these provide the list
```
tjkirch: I think these filenames are now out of date.  Same above `fn commands()`.

sources/logdog/src/commands.rs
webern: Not in this case since I am constructing the error myself. I'm fairly certain the into() calls will only execute when the code branch is executed.

webern:
> webern force-pushed the webern:logdog-ecs branch from 7d1f985 to 60e5d3c

Rebase develop without any other changes.

webern:
> webern force-pushed the webern:logdog-ecs branch from 60e5d3c to aa7d592 

Un-renamed `log_request.rs` / `command.rs`. Re-ordered the error re-ordering commit to happen first. Addressed all PR comments.

webern:
> webern force-pushed the webern:logdog-ecs branch from aa7d592 to cb6b585

A fix in the readme.

sources/logdog/static/common
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/static/aws-ecs-1
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/build.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/conf/logdog.aws-ecs-1.conf
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/commands.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

Dockerfile
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/.gitignore
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/build.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/build.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/error.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/error.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/main.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/commands.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/commands.rs
webern: No longer applicable: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/commands.rs
webern: Scrubbed it all back to 'log request' terminology: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/commands.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/commands.rs
webern: [aa7d592](https://github.com/bottlerocket-os/bottlerocket/compare/60e5d3c8e65c58170e8f72d944a80be11ea098e0..aa7d59202dad5161a9dc8aff3995f1fbe0d4c706)

sources/logdog/src/main.rs
```diff
@@ -12,7 +12,12 @@ logs are at: /tmp/bottlerocket-logs.tar.gz
 
 # Logs
 
-For the commands used to gather logs, please see [log_request](src/log_request.rs).
+For the log requests used to gather logs, please see the following:
+
+* [log_requests](src/log_requests.rs)
```
tjkirch: Looks like the last update might have fixed the README directly but not the source here.

sources/logdog/build.rs
```diff
@@ -30,3 +89,9 @@ fn main() {
     let mut readme = File::create("README.md").unwrap();
     readme.write_all(content.as_bytes()).unwrap();
 }
+
+fn main() {
+    println!("cargo:rerun-if-env-changed={}", ENV_VARIANT);
+    symlink_variant();
```
tjkirch: Can we do the print inside the function so it's clearer why it's relevant?

Dockerfile
tjkirch: Did you find a way to check that the directories are actually empty at the start of the logdog and models build scripts?

sources/logdog/src/log_request.rs
```diff
@@ -1,29 +1,238 @@
...
+    instructions: &'a str,
+}
+
+/// This is used in error construction.
+impl ToString for LogRequest<'_> {
+    fn to_string(&self) -> String {
+        if self.instructions.is_empty() {
+            format!("{} {}", self.mode, self.filename)
+        } else {
+            format!("{} {} {}", self.mode, self.filename, self.instructions,)
```
```suggestion
            format!("{} {} {}", self.mode, self.filename, self.instructions)
```

webern:
> webern force-pushed the webern:logdog-ecs branch from cb6b585 to 3ed71ad

I missed this readme change in the previous push.

sources/logdog/src/log_request.rs
```diff
@@ -1,29 +1,238 @@
...
+use std::fs;
+use std::fs::File;
+use std::path::Path;
+use std::process::Stdio;
+use url::Url;
+
+/// The `logdog` log requests that all variants have in common.
+const COMMON_REQUESTS: &str = std::include_str!("../conf/logdog.common.conf");
+/// The `logdog` log requests that are specific to the current variant.
+const VARIANT_REQUESTS: &str = std::include_str!("../conf/current/logdog.conf");
```
```suggestion
const COMMON_REQUESTS: &str = include_str!("../conf/logdog.common.conf");
/// The `logdog` log requests that are specific to the current variant.
const VARIANT_REQUESTS: &str = include_str!("../conf/current/logdog.conf");
```

sources/logdog/src/log_request.rs
```diff
@@ -1,29 +1,238 @@
...
+    })?;
+    let (command, args) = split.split_first().context(error::CommandMissing {
+        request: request.to_string(),
+    })?;
+    let outpath = tempdir.as_ref().join(request.filename);
+    let ofile = File::create(&outpath).context(error::CommandOutputFile { path: &outpath })?;
+    let stderr_file = ofile
+        .try_clone()
+        .context(error::CommandErrFile { path: &outpath })?;
+    std::process::Command::new(command)
```
tjkirch: Can we `use` this instead of using the full path now that there's no naming conflict?

sources/logdog/src/log_request.rs
```diff
@@ -1,29 +1,238 @@
...
+        error::HttpMissingUrl {
+            request: request.to_string(),
+        }
+    );
+    let outpath = tempdir.as_ref().join(request.filename);
+    let response = send_get_request(request.instructions)?;
+    let data = response.bytes().context(error::HttpResponseBytes {
+        request: request.to_string(),
+    })?;
+    std::fs::write(&outpath, &data).context(error::HttpWriteBytes {
```
```suggestion
    fs::write(&outpath, &data).context(error::HttpWriteBytes {
```

sources/logdog/src/log_request.rs
```diff
@@ -1,29 +1,238 @@
...
+{
+    ensure!(
+        !request.instructions.is_empty(),
+        error::FileFromEmpty {
+            request: request.to_string()
+        }
+    );
+    let dest = tempdir.as_ref().join(request.filename);
+    let _ = fs::copy(&request.instructions, &dest).context(error::FileCopy {
+        request: request.to_string(),
```
tjkirch, line 201: Do you need the manual `to_string` here?  If so, can we do a `with_context`?

tjkirch (approved):

Dockerfile
webern: I'll work on it this morning.

sources/logdog/src/log_request.rs
webern: It is needed, so switched to with_context.

webern:
> webern force-pushed the webern:logdog-ecs branch from 3ed71ad to f8f85c6

Final fixes and cleanups.

sources/logdog/src/log_request.rs
webern: A bit disappointed that fmt didn't fix this.

sources/logdog/src/log_request.rs
tjkirch: Same thing on the line above?

Dockerfile
webern: I added this to `logdog/build.rs`:

```rust
    if std::path::PathBuf::from("conf/current/models-foo").is_file() {
        panic!("models-foo exists!");
    }
    std::fs::write("conf/current/logdog-foo", "").unwrap();
```

And I added this to `models/build.rs`:

```rust
    if std::path::PathBuf::from("src/variant/logdog-foo").is_file() {
        panic!("logdog-foo exists!");
    }
    std::fs::write("src/variant/models-foo", "foo").unwrap();
```

`cargo make` succeeded, which I think shows that the two cache directories are independent.

webern:
>  webern force-pushed the webern:logdog-ecs branch from f8f85c6 to ab8692e

Avoid allocations when not-needed by using `with_context`.

webern:
> webern force-pushed the webern:logdog-ecs branch from ab8692e to a01449d 

File copy doesn't work for /proc/mounts so using cat command instead. Also /etc/mtab (which also can not be copied with a file request) is duplicative of /proc/mounts, so removed it.

tjkirch (approved):

bcressey (approved):

PR #1055, opened 2020-08-20
Added patch for EC2 IMDSv2 support in Docker

srgothi92:
**Issue number:**
N/A

**Description of changes:**
Used Look aside cache for caching awslogs_update_aws-sdk-go_to_support_imdsv2.patch 
 Applied the patch to docker-engine for EC2 IMDSv2 support in Docker

**Testing done:**
1. Created an ECS taskDefinition that uses `awslogs` Log driver. 
2. Ran the task on two instance : 1. With Patch 2. Without Patch . Task Ran Successfully on Both the instance becasue imdsv1 was enabled. 
2. Ran the task on two instances 1. With Patch 2. Without patch, with `--metadata-options "HttpEndpoint=enabled,HttpTokens=required"` v1 disabled. Task Stopped on 1st instance but ran successfully on 2nd instance.
3. Verified CloudWatch metric `MetadataNoToken` to zero for 2nd Instance.

**Terms of contribution:**


packages/docker-engine/Cargo.toml
```diff
@@ -12,6 +12,11 @@ path = "pkg.rs"
 url = "https://github.com/docker/engine/archive/v19.03.12/engine-19.03.12.tar.gz"
 sha512 = "51632ca8cff03440e3b76ad3e7f06b9c6275c776f4c4f9dc88cd7f5aa9b8aa8cb16ce2b3702426cb93f3714265a6cd28d4ba31612db37199e2915821b9ee8fd4"
 
+[[package.metadata.build-package.external-files]]
```
samuelkarp, line 15: A note here for why the patch is not checked in to the repo (like [this](https://github.com/bottlerocket-os/bottlerocket/blob/23a20e5e91cadc007a37f1c0471822c2fd2bf0d3/packages/kubernetes-1.17/Cargo.toml#L20-L22)) would be useful.

packages/docker-engine/docker-engine.spec
```diff
@@ -32,6 +32,8 @@ Source1000: clarify.toml
 # Bottlerocket-specific - Privileged containers should receive SELinux labels
 # https://github.com/bottlerocket-os/bottlerocket/issues/1011
 Patch0001: 0001-bottlerocket-privileged-shouldn-t-disable-SELinux.patch
+# Update aws-sdk-go for IMDSv2 support
+Patch0100: awslogs_update_aws-sdk-go_to_support_imdsv2.patch.bz2
```
samuelkarp, line 37: Why did you skip to 100 instead of going to 2?

packages/docker-engine/Cargo.toml
srgothi92: I am not sure where should I check-in ?

packages/docker-engine/docker-engine.spec
srgothi92: Tom's suggestion. To distinguish between two patches to different package.

tjkirch: I assume 1-N will be a series of patches you (@samuelkarp) write for Bottlerocket-specific changes, and generally patch series from different sources will be in different numbered chunks, like jumping up to 100 or 1,000.

packages/docker-engine/Cargo.toml
tjkirch: Sam just means adding a comment above this line explaining why the file isn't checked into the repo directly.  You can copy the comment he linked to start, and update it for these circumstances.  (Obviously the link in that comment doesn't apply directly, it's for a different package.)

samuelkarp: "This is a large patch, so we don't want to check it into the repo.  It's like https://github.com/docker/engine/commit/44a8e10bfc794dbbc4011a62fa8ba71d948d13cf and applies to 19.03.12 without changes."

packages/docker-engine/.gitignore
```diff
@@ -0,0 +1 @@
+/*.patch.bz2
```
tjkirch: This ignore isn't specific to docker-engine, it would apply to any package, so I think this file should be at `/packages/.gitignore` instead.

packages/docker-engine/docker-engine.spec
```diff
@@ -32,6 +32,8 @@ Source1000: clarify.toml
 # Bottlerocket-specific - Privileged containers should receive SELinux labels
 # https://github.com/bottlerocket-os/bottlerocket/issues/1011
 Patch0001: 0001-bottlerocket-privileged-shouldn-t-disable-SELinux.patch
+# Update aws-sdk-go for IMDSv2 support
```
tjkirch:
```suggestion
Patch0001: 0001-bottlerocket-privileged-shouldn-t-disable-SELinux.patch

# Update aws-sdk-go for IMDSv2 support
```
This isn't Bottlerocket-specific, so we want it to visually appear as a different section.

packages/docker-engine/Cargo.toml
tjkirch: I confirmed it's exactly the same patch, so we don't need to say "like".  Perhaps:

```
# This is a large patch, so we don't want to check it into the repo. Source:
# https://github.com/docker/engine/commit/44a8e10bfc794dbbc4011a62fa8ba71d948d13cf
```

packages/docker-engine/.gitignore
srgothi92: I saw it was there in Kubernetes package, so i added it here. Do you want me to remove form other packages and add one in packages root?

tjkirch: Yeah, we can move it up to the root of packages. :)

packages/.gitignore
```diff
@@ -0,0 +1 @@
+**/*.patch.bz2
```
tjkirch:
```suggestion
*.patch.bz2
```
I think this simpler version will work without having to include the wildcard for directories.

tjkirch (commented):
Looks good - would you mind splitting this into a commit for the patch and a commit for moving the gitignore?

packages/docker-engine/Cargo.toml
```diff
@@ -12,6 +12,13 @@ path = "pkg.rs"
 url = "https://github.com/docker/engine/archive/v19.03.12/engine-19.03.12.tar.gz"
 sha512 = "51632ca8cff03440e3b76ad3e7f06b9c6275c776f4c4f9dc88cd7f5aa9b8aa8cb16ce2b3702426cb93f3714265a6cd28d4ba31612db37199e2915821b9ee8fd4"
 
+# This is a large patch, so we don't want to check it into the repo. Source:
+# https://github.com/docker/engine/commit/44a8e10bfc794dbbc4011a62fa8ba71d948d13cf
```
tjkirch: Seeing this in context made me realize something... :)

This patch has an externally accessible URI, so it's not the same case as with k8s.

```
https://github.com/docker/engine/commit/44a8e10bfc794dbbc4011a62fa8ba71d948d13cf.patch
```

Because of that, we can reference that URL in the `url` field below, instead of using a `file:///` URL, to make it clearer to others where the patch comes from.  We don't need this comment then, either, because the source becomes obvious.

I'm also pretty sure you can use the http URL in the `Patch100:` line in the spec file, instead of just the filename, the same way we do for `Source` lines.

Sorry for the wild goose chase on this; I didn't originally know that the patch was public.

srgothi92: I am little confused here, I don't know how this parameters are used.  Just changing the `url` is enough ? Or I should change the `sha512` remove `path` and upload again to `lookaside cache`. Wouldn't the name of patch change to `4a8e10bfc794dbbc4011a62fa8ba71d948d13cf.patch` should I change the name in `docker-engine.spec` file as well. 

I tried example similar to `engine-19.03.12.tar.gz` but build failed with `Failed to fetch 'https://cache.bottlerocket.aws/44a8e10bfc794dbbc4011a62fa8ba71d948d13cf.patch/`sha512/44a8e10bfc794dbbc4011a62fa8ba71d948d13cf.patch``

tjkirch: If you add a `path` key to the `external-files` section, you can rename the file however you like.  For example, in `packages/iputils/Cargo.toml` we rename `s20190709.tar.gz` (the end of the `url`) to `iputils-s20190709.tar.gz` so the naming is more clear.  So we could use a `path = "awslogs_update_aws-sdk-go_to_support_imdsv2.patch.bz2"` entry, in this case, to get the filename back to what we expect, and then the lookaside cache will work without needing to know a new name.

srgothi92: Done, Changed url and tested again.

tjkirch (approved):

samuelkarp (approved):

srgothi92:
Changed the URL parameter in Cargo.toml to actual external patch URL. Retested with the changes.

PR #1056, opened 2020-08-20
isolate host containers and limit access to API socket

bcressey:
**Issue number:**
N/A

**Description of changes:**
Create a distinction between host containers and orchestrated containers in the SELinux policy, and update host-ctr to match. Now host containers will run with one of two privileged labels, and the files they write will receive a different label. 

Limit access to the API socket to processes with one of the privileged labels. All host containers will have access to the API, but orchestrated containers will need to opt-in to its use by running as either `control_t` or `super_t`.

**Testing done:**
Confirmed that `apiclient` in the control container and admin container still works, and that the files they create are correctly labeled.

**Terms of contribution:**


tjkirch (approved):

samuelkarp (approved):

zmrow (approved):
🦺

PR #1058, opened 2020-08-21
fix growpart to avoid race in partition table reload

### LINKED ISSUE #845: occasional failure to prepare local storage on boot
**Image I'm using:**
v0.3.1

All instances come up in a healthy state.

**What actually happened:**
Some instances fail to boot successfully.

```
[    5.686707]  nvme1n1: p1
[    5.711577]  nvme1n1: p1
[ FAILED ] Failed to start Prepare Local Directory (/local). See 'systemctl status prepare-local.service' for details.
[    4.588335] growpart[2042]: Failed to canonicalize link for '/dev/disk/by-partlabel/BOTTLEROCKET-DATA': No such file or directory (os error 2)
```

Launch instances.

bcressey:
**Issue number:**
#845 

**Description of changes:**
udevd watches files in /dev via inotify, and when a block disk device is closed after writing, it tells the kernel to reload the partition table.

This was racing with the reload code in growpart; if we were unlucky in our timing, the kernel would be busy with udevd's reload and our call would fail.

To fix this, we stop reloading the partition table in growpart, and use inotify to determine when udevd has deleted and recreated the partition links.

**Testing done:**
Tested launches and reboots of 250 instances in a cluster. All came up successfully and resized the partition to the expected 20 GiB size.

```
# systemctl status prepare-local
● prepare-local.service - Prepare Local Directory (/local)
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/prepare-local.service; enabled; vendor preset: enabled)
     Active: active (exited) since Fri 2020-08-21 19:05:08 UTC; 22min ago
   Main PID: 2049 (code=exited, status=0/SUCCESS)
      Tasks: 0 (limit: 9185)
     Memory: 0B
     CGroup: /system.slice/prepare-local.service

Aug 21 19:05:07 localhost growpart[1911]: read partition table from nvme1n1
Aug 21 19:05:07 localhost growpart[1911]: wrote partition table to /dev/nvme1n1
Aug 21 19:05:07 localhost growpart[1911]: saw BOTTLEROCKET-DATA link deleted
Aug 21 19:05:07 localhost growpart[1911]: saw BOTTLEROCKET-DATA link created
Aug 21 19:05:07 localhost growpart[2008]: read partition table from nvme1n1
Aug 21 19:05:07 localhost growpart[2008]: wrote partition table to /dev/nvme1n1
Aug 21 19:05:07 localhost growpart[2008]: saw BOTTLEROCKET-DATA link deleted
Aug 21 19:05:07 localhost growpart[2008]: saw BOTTLEROCKET-DATA link created
Aug 21 19:05:08 localhost systemd-growfs[2026]: Successfully resized "/local" to 19.9G bytes.
Aug 21 19:05:08 localhost systemd[1]: Finished Prepare Local Directory (/local).
```

**Terms of contribution:**

### DIFF
--- a/sources/growpart/src/diskpart.rs
+++ b/sources/growpart/src/diskpart.rs
@@ -8,14 +8,15 @@ use error::Result;
 
 use block_party::BlockDevice;
 use gptman::{GPTPartitionEntry, GPT};
+use inotify::{EventMask, Inotify, WatchMask};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::fs;
-use std::os::linux::fs::MetadataExt;
 use std::path::{Path, PathBuf};
 
 pub struct DiskPart {
     gpt: GPT,
     device: PathBuf,
+    watcher: WatchPart,
 }
 
 impl DiskPart {
@@ -28,6 +29,8 @@ impl DiskPart {
         let disk = Self::find_disk(path)?;
         let device = disk.path();
         let gpt = Self::load_gpt(&device)?;
+        println!("read partition table from {}", disk);
+
         let used_partitions = gpt.iter().filter(|(_num, part)| part.is_used()).count();
         ensure!(
             used_partitions == 1,
@@ -36,7 +39,14 @@ impl DiskPart {
                 count: used_partitions
             }
         );
-        Ok(Self { device, gpt })
+
+        let watcher = WatchPart::new(path)?;
+
+        Ok(Self {
+            device,
+            gpt,
+            watcher,
+        })
     }
 
     /// Grow a single partition to fill the available capacity on the device.
@@ -47,7 +57,6 @@ impl DiskPart {
         let partition_name = current.partition_name.clone();
         let partition_type_guid = current.partition_type_guid;
         let unique_parition_guid = current.unique_parition_guid;
-
         let path = &self.device;
 
         // Remove all existing partitions so that the space shows up as free.
@@ -79,8 +88,7 @@ impl DiskPart {
         Ok(())
     }
 
-    /// Write the GPT label back to the device. If this is a block device, tell
-    /// the kernel to reload the partition table.
+    /// Write the GPT label back to the device.
     pub(crate) fn write(&mut self) -> Result<()> {
         let path = &self.device;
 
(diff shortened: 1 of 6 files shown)


sources/growpart/src/diskpart.rs
```diff
@@ -140,3 +148,63 @@ impl DiskPart {
...
+       // When the kernel reloads the partition table, we expect two events, when udev deletes and
+       // then recreates the path. This isn't synchronized with our code, so to avoid races we need
+       // to watch for both events.
+        let mut inotify = Inotify::init().context(error::InitInotify)?;
+        inotify
+            .add_watch(&dirname, WatchMask::CREATE | WatchMask::DELETE)
+            .context(error::AddInotifyWatch)?;
+
+        let filename = path.file_name().context(error::FindFileName { path })?;
+        let filename = Path::new(filename).to_path_buf();
```
tjkirch: nit: maybe do these before doing inotify, in case these fail for a silly reason?

sources/growpart/src/diskpart.rs
```diff
@@ -140,3 +148,63 @@ impl DiskPart {
...
+            let events = self
+                .inotify
+                .read_events_blocking(&mut buf)
+                .context(error::ReadInotifyEvents)?;
+
+            for event in events {
+                if let Some(event_file) = event.name {
+                    if self.filename != Path::new(event_file) {
+                        continue;
+                    } else if event.mask == EventMask::DELETE {
```
tjkirch:
```suggestion
                        continue;
                    }

                    if event.mask == EventMask::DELETE {
```
nit: these are logically separate conditionals; this would have to run the same way since the first one breaks out.

tjkirch (approved):

sources/growpart/src/diskpart.rs
```diff
@@ -140,3 +148,63 @@ impl DiskPart {
...
+    fn new<P>(path: P) -> Result<Self>
+    where
+        P: AsRef<Path>,
+    {
+        let path = path.as_ref();
+        let dirname = path.parent().context(error::FindParentDirectory { path })?;
+
+       // When the kernel reloads the partition table, we expect two events, when udev deletes and
+       // then recreates the path. This isn't synchronized with our code, so to avoid races we need
+       // to watch for both events.
```
samuelkarp: Do we need to start this watch prior to writing the new partition table to disk?  It looks like there's still a race here: the work to reload the partition table and rewrite the symlinks could complete prior to us setting up the inotify watch.

bcressey: We do; but the watch is set up in `Diskpart::new`, we just wait to ask the kernel for corresponding events until after the first write.

sources/growpart/src/diskpart.rs
bcressey: True, and I had it the other way at first, but I thought this was slightly more readable since there are really only these three branches to consider.

sources/growpart/src/diskpart.rs
samuelkarp: I missed that, thanks!

samuelkarp (approved):

tjkirch (approved):

PR #1059, opened 2020-08-21
Updated Readme to provide more context to admin container access

srgothi92:
**Issue number:**
N/A

**Description of changes:**
Updated `README.md` to add more context to admin container access. 

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,8 +128,9 @@ If you're using a custom control container, or want to make the API calls direct
 apiclient -u /settings -m PATCH -d '{"host-containers": {"admin": {"enabled": true}}}'
 apiclient -u /tx/commit_and_apply -m POST
 ```
-
-Once you're in the admin container, you can run `sheltie` to get a full root shell in the Bottlerocket host.
+  
+Once you have enabled admin container, you can ssh with `ec2-user` using your EC2-registered SSH key to enter admin container.
+In the admin container, you can run `sheltie` to get a full root shell in the Bottlerocket host.
 Be careful; while you can inspect and change even more as root, Bottlerocket's filesystem and dm-verity setup will prevent most changes from persisting over a restart - see [Security](#security).
 
 ## Updates


README.md
```diff
@@ -128,8 +128,9 @@ If you're using a custom control container, or want to make the API calls direct
 apiclient -u /settings -m PATCH -d '{"host-containers": {"admin": {"enabled": true}}}'
 apiclient -u /tx/commit_and_apply -m POST
 ```
-
-Once you're in the admin container, you can run `sheltie` to get a full root shell in the Bottlerocket host.
+  
+Once you have enabled admin container, you can ssh with `ec2-user` using your EC2-registered SSH key to enter admin container.
```
bcressey, line 132: This duplicates the text on line 109 - not sure I see the value in repeating the information.

srgothi92: I agree. But, It gets ignored for someone trying for the 1st time. Since the workflow is not straightforward like any other operating system. 
My experience: We follow document top down, document lists steps how to enable admin container from control container and directly jumps to once you are in admin container. 
 Additionally control container logs `The admin container is now enabled - it should pull and start soon, and then you can SSH in`, so it feels that admin container is getting pulled in control container and we can start using `sudo sheltie`. But, that does not work and you start wondering what's wrong.  I was just looking to add information in order required.
We can change the format how this information is added, but I feel adding something there on "How to enter admin container make sense".  Open to suggestion to edit the section if you agree. Or I can delete the PR :)

srgothi92: Just saw an example in [Log](https://github.com/bottlerocket-os/bottlerocket#logs), may be we can add something like that. It's one of the important step and least info is provided.
```bash
ssh -i YOUR_KEY_FILE \
    ec2-user@YOUR_HOST \
    "cat /.bottlerocket/rootfs/tmp/bottlerocket-logs.tar.gz" > bottlerocket-logs.tar.gz
```

srgothi92:
PR duplicates the information

PR #1060, opened 2020-08-21
pubsys: add SSM parameter publishing

zmrow:
`cargo make ssm` will use the amis.json from `cargo make ami` to populate
parameter name/value templates from files in `policies/ssm`.  The parameters
are set in SSM using the full build version, e.g. "0.5.0-abcdef", and it won't
overwrite by default.

`cargo make promote-ssm -e SSM_TARGET=VERSION` will promote (copy) those values
from the full-build-version name to a more general one.  It's recommended to
promote from the full version to a short version like "0.5.0", then to a
well-known pointer like "latest".  This allows for easy rollback by promoting
from an older version, using `-e SSM_SOURCE=OLD_VERSION`.

**Testing done:**
* Tested setting more than 100 parameters in a region - request chunking and throttling worked, parameters got through                                                                                         
* Tested the "new namespace" issue in opt-in region - error was caught properly and it was treated as a new parameter                                                                                          
* Tested many regions at once                                                                                                                                                                                  
* Confirmed no overwrite allowed unless ALLOW_CLOBBER                                                                                                                                                          
* Confirmed that adding regions / parameters will add the new ones without considering it a clobber  (existing parameters show no change; idempotent)                                                          
* Confirmed that override templates work, and it doesn't run overrides not matching current values
* Confirmed that set_parameter retries work, or at least that bad requests are retried and eventually fail, and good requests pass, and the output is clear about failures and failure count.

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/Infra.toml.example
+++ b/tools/pubsys/Infra.toml.example
@@ -38,6 +38,8 @@ regions = ["us-west-2", "us-east-1", "us-east-2"]
 profile = "my-profile"
 # If specified, we assume this role before making any API calls.
 role = "arn:aws:iam::012345678901:role/assume-global"
+# If specified, this string will be prefixed on all parameter names published to SSM.
+ssm_prefix = "/your/prefix/here"
 
 [aws.region.us-west-2]
 # If specified, we assume this role before making any API calls in this region.
--- a/tools/pubsys/policies/ssm/defaults.toml
+++ b/tools/pubsys/policies/ssm/defaults.toml
@@ -1,0 +1,7 @@
+[[parameter]]
+name = "{variant}/{arch}/{image_version}/image_id"
+value = "{image_id}"
+
+[[parameter]]
+name = "{variant}/{arch}/{image_version}/image_version"
+value = "{image_version}"
--- a/tools/pubsys/src/aws/ami/mod.rs
+++ b/tools/pubsys/src/aws/ami/mod.rs
@@ -15,6 +15,7 @@ use register::{get_ami_id, register_image};
 use rusoto_core::{Region, RusotoError};
 use rusoto_ebs::EbsClient;
 use rusoto_ec2::{CopyImageError, CopyImageRequest, CopyImageResult, Ec2, Ec2Client};
+use serde::{Deserialize, Serialize};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::{HashMap, VecDeque};
 use std::fs::File;
@@ -70,11 +71,11 @@ pub(crate) struct AmiArgs {
 /// Common entrypoint from main()
 pub(crate) async fn run(args: &Args, ami_args: &AmiArgs) -> Result<()> {
     match _run(args, ami_args).await {
-        Ok(ami_ids) => {
+        Ok(amis) => {
             // Write the AMI IDs to file if requested
             if let Some(ref path) = ami_args.ami_output {
                 let file = File::create(path).context(error::FileCreate { path })?;
-                serde_json::to_writer_pretty(file, &ami_ids).context(error::Serialize { path })?;
+                serde_json::to_writer_pretty(file, &amis).context(error::Serialize { path })?;
                 info!("Wrote AMI data to {}", path.display());
             }
             Ok(())
@@ -83,8 +84,8 @@ pub(crate) async fn run(args: &Args, ami_args: &AmiArgs) -> Result<()> {
     }
 }
 
-async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, String>> {
-    let mut ami_ids = HashMap::new();
+async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>> {
+    let mut amis = HashMap::new();
 
     info!(
         "Using infra config from path: {}",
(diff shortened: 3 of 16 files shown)


tjkirch (approved):

tjkirch:
^ This push fixes an issue Ben found.  There's a convenience function to join together SSM prefix and parameter name, making sure there's only one slash between them, and we removed a character in the wrong branch.

bcressey (approved):

tjkirch:
:arrow_up: This force-push rebases on #1064 to get the STS region fix.

:arrow_down: The force-push below incorporates the changes from #1064 into the SSM code from this PR.

tools/pubsys/Infra.toml.example
```diff
@@ -38,6 +38,8 @@ regions = ["us-west-2", "us-east-1", "us-east-2"]
 profile = "my-profile"
 # If specified, we assume this role before making any API calls.
 role = "arn:aws:iam::012345678901:role/assume-global"
+# If specified, this string will be prefixed on all parameter names published to SSM.
+ssm_prefix = "/your/prefix/here"
```
webern, line 42: The path to the ssm policy is not specified in Infra.toml?

tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -287,7 +294,25 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, String>
 
     ensure!(!saw_error, error::AmiCopy);
 
-    Ok(ami_ids)
+    Ok(amis)
+}
+
+/// If JSON output was requested, we serialize out a mapping of region to AMI information; this
```
webern, line 300: Nice, this should be helpful. Currently I grep for the ami ids.

tools/pubsys/src/aws/ssm/template.rs
```diff
@@ -0,0 +1,216 @@
...
+use tinytemplate::TinyTemplate;
+
+/// Represents a single SSM parameter
+#[derive(Debug, Deserialize)]
+pub(crate) struct TemplateParameter {
+    pub(crate) name: String,
+    pub(crate) value: String,
+}
+
+/// Represents a set of of SSM parameters, in a format that allows for clear definition of
```
webern:
```suggestion
/// Represents a set of SSM parameters, in a format that allows for clear definition of
```

tools/pubsys/src/aws/ssm/template.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub(crate) enum Error {
+        #[snafu(display("Error building template from '{}': {}", template, source))]
+        AddTemplate {
+            template: String,
+            source: tinytemplate::error::Error,
+        },
+
+        #[snafu(display("Failed to {} '{}': {}", op, path.display(), source))]
+        File {
+            op: String,
```
webern, line 192: Nice solution to the profusion of error variants problem.

tools/pubsys/src/aws/ssm/ssm.rs
```diff
@@ -0,0 +1,377 @@
...
+        std::result::Result<GetParametersResult, RusotoError<GetParametersError>>,
+    )> = request_stream.collect().await;
+
+    // If you're checking parameters in a region you haven't pushed to before, you can get an
+    // error here about the parameter's namespace being new.  We want to treat these as new
+    // parameters rather than failing.  Unfortunately, we don't know which parameter in the region
+    // was considered new, but we expect that most people are publishing all of their parameters
+    // under the same namespace, so treating the whole region as new is OK.  We use this just to
+    // warn the user.
+    let mut new_regions = HashSet::new();
```
webern, line 70: 😢  Nice solution.

tools/pubsys/policies/ssm/defaults.toml
```diff
@@ -0,0 +1,7 @@
+[[parameter]]
+name = "{variant}/{arch}/{image_version}/image_id"
+value = "{image_id}"
+
+[[parameter]]
```
webern, line 5: Don't we have one of these for `current`? Should that be in the defaults.toml as well?

webern (approved):
Nice work!

etungsten (approved):

tools/pubsys/Infra.toml.example
tjkirch: Yeah, though we could consider moving it, and there's also some separate discussion about combining some of our configuration sources...

The way we were thinking about the configuration sources is that the Makefile would have default values/paths for things that should work for most people and rarely need to change; Infra.toml would have user-specific settings that rarely need to change; parameters handle things that always need to be specified or frequently change; and JSON files as a "runtime store" for things that change each time, like the AMI IDs.

tools/pubsys/policies/ssm/defaults.toml
tjkirch: For `cargo ssm`, the `image_version` in the template is replaced by the version you're building, and for `promote-ssm`, by the version you specify with `-e SSM_TARGET=`.  So when we want to publish `latest`, we'd do `cargo make promote-ssm -e SSM_TARGET=latest`.  In our release process we're planning to promote once from the full version (e.g. `0.5.0-abcdef`) to the short version (`0.5.0`) and again to `latest`.

tjkirch:
^ This force-push fixes a typo caught by @webern.

PR #1053, opened 2020-08-19
Add documentation for the aws-ecs-1 variant

samuelkarp:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket/issues/815

**Description of changes:**
* New `QUICKSTART-ECS.md` setup guide for ECS
* Updates to `README.md` to document the variant
* Updates to `BUILDING.md` to link to the new setup guide

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #1062, opened 2020-08-22
aws-ecs-1: add ecs.loglevel setting

samuelkarp:
**Issue number:**

(Note: this overlaps with https://github.com/bottlerocket-os/bottlerocket/pull/1053 as it adds a new setting.  I will update either this one or the other one depending on which gets merged first.)

**Description of changes:**
The log level for the ECS agent cannot be set using ecs.config.json.  Instead, for this setting we use a template file which is then sourced as an EnvironmentFile in ecs.service.

This change also reduces the default logging level from "debug" (which is quite chatty) to "info" (which is the default on the ECS-optimized AMIs).

**Testing done:**
Built a new AMI.  Observed that the correct default value was populated to `/etc/ecs/ecs.config` and that the ECS agent was logging at the correct level.  Used `apiclient` to patch the value to "debug", committed the change, and observed the correct change to `/etc/ecs/ecs.config` as well as the ECS agent's logging behavior.  Used `apiclient` to try and set an invalid value, received an appropriate error response.

**Terms of contribution:**

### DIFF
--- a/packages/ecs-agent/ecs.service
+++ b/packages/ecs-agent/ecs.service
@@ -11,7 +11,6 @@ Restart=on-failure
 RestartPreventExitStatus=5
 RestartSec=1s
 EnvironmentFile=-/etc/ecs/ecs.config
-Environment=ECS_LOGFILE=/var/log/ecs/ecs-agent.log ECS_LOGLEVEL=debug
 Environment=ECS_CHECKPOINT=true
 ExecStartPre=/sbin/iptables -t nat -A PREROUTING -d 169.254.170.2/32 \
  -p tcp -m tcp --dport 80 -j DNAT --to-destination 127.0.0.1:51679
--- a/sources/models/src/modeled_types/ecs.rs
+++ b/sources/models/src/modeled_types/ecs.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 // Just need serde's Error in scope to get its trait methods
 use super::error;
 use serde::de::Error as _;
-use snafu::ensure;
+use snafu::{ensure, ResultExt};
 use std::borrow::Borrow;
 use std::convert::TryFrom;
 use std::fmt;
@@ -172,6 +172,59 @@ mod test_ecs_attribute_value {
             "trailing space ",
         ] {
             ECSAttributeValue::try_from(*val).unwrap_err();
+        }
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// ECSAgentLogLevel represents a string that contains a valid ECS log level for the ECS agent.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ECSAgentLogLevel {
+    inner: String,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum ECSLogLevel {
+    Debug,
+    Info,
+    Warn,
+    Error,
+    Crit,
+}
+
+string_impls_for!(ECSAgentLogLevel, "ECSAgentLogLevel");
+
+impl TryFrom<&str> for ECSAgentLogLevel {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        serde_plain::from_str::<ECSLogLevel>(&input).context(error::InvalidPlainValue {
+            field: "ecs.loglevel",
+        })?;
+        Ok(ECSAgentLogLevel {
+            inner: input.to_string(),
+        })
+    }
+}
+
+#[cfg(test)]
+mod test_ecs_agent_log_level {
+    use super::ECSAgentLogLevel;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn good_vals() {
+        for val in &["debug", "info", "warn"] {
+            ECSAgentLogLevel::try_from(*val).unwrap();
+        }
+    }
+
+    #[test]
+    fn bad_vals() {
+        for val in &["", "warning", "errors", " "] {
+            ECSAgentLogLevel::try_from(*val).unwrap_err();
         }
     }
 }
--- a/sources/models/src/modeled_types/mod.rs
+++ b/sources/models/src/modeled_types/mod.rs
@@ -47,6 +47,12 @@ pub mod error {
 
         #[snafu(display("Invalid domain name '{}': {}", input, msg))]
         InvalidDomainName { input: String, msg: String },
+
+        #[snafu(display("Invalid input for field {}: {}", field, source))]
+        InvalidPlainValue {
+            field: String,
+            source: serde_plain::Error,
+        },
     }
 }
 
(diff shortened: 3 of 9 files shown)


bcressey (approved):

packages/ecs-agent/ecs.service
```diff
@@ -11,7 +11,7 @@ Restart=on-failure
 RestartPreventExitStatus=5
 RestartSec=1s
 EnvironmentFile=-/etc/ecs/ecs.config
-Environment=ECS_LOGFILE=/var/log/ecs/ecs-agent.log ECS_LOGLEVEL=debug
+Environment=ECS_LOGFILE=/var/log/ecs/ecs-agent.log
```
tjkirch: Should this move to the new env file too, just hardcoded in the template?  It's easier to change and easier to find (on the instance) than the service file, imo.

sources/models/src/modeled_types/mod.rs
```diff
@@ -47,6 +47,12 @@ pub mod error {
 
         #[snafu(display("Invalid domain name '{}': {}", input, msg))]
         InvalidDomainName { input: String, msg: String },
+
+        #[snafu(display("Invalid input '{}': {}", input, source))]
+        InvalidPlainValue {
+            input: String,
+            source: serde_plain::Error,
```
tjkirch: This seems like it'd be useful for other modeled types as well, since serde_plain seems like a good fit for things like this.  Perhaps it could include another string for the intended purpose of the input, e.g. "ECS loglevel" or "loglevel" here, to make it clearer what was being checked?  I'm not sure the serde_plain error will always be enough.

sources/models/src/modeled_types/ecs.rs
```diff
@@ -175,3 +176,52 @@ mod test_ecs_attribute_value {
...
+    inner: String,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum LogLevel {
+    Debug,
+    Info,
+    Warn,
+}
```
tjkirch: Should we include all of the [values the agent supports](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html)?

```
Example values: crit, error, warn, info, debug
```

sources/models/src/modeled_types/ecs.rs
```diff
@@ -175,3 +176,52 @@ mod test_ecs_attribute_value {
...
+
+/// ECSAgentLogLevel represents a string that contains a valid ECS log level for the ECS agent.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ECSAgentLogLevel {
+    inner: String,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum LogLevel {
```
tjkirch:
```suggestion
enum ECSLogLevel {
```
Seems reasonable that we'd have other log level settings.

This does make me think that `ECSAgentLogLevel` and `ECSLogLevel` are confusingly similar.  Perhaps we could just use the enum (and serde_plain's `to_string`) rather than the struct with original `input` string if we don't think the exact input string matters... in this case, I don't think it does, since it could only differ in capitalization and (by virtue of the `lowercase`) we're saying that doesn't matter.

tjkirch (approved):

sources/models/src/modeled_types/ecs.rs
samuelkarp: That's what I had in my first attempt, but I don't know how common it'd be to set to a level like "error" or "critical".  I can add them now, or we can always add later if people actually do want even less verbosity.

(The agent also supports ["none"](https://github.com/aws/amazon-ecs-agent/blob/478c4ec00a43b4713a0acc811cfd1e2d70307f63/agent/logger/log.go#L135) which disables all logging...)

sources/models/src/modeled_types/ecs.rs
samuelkarp: I had trouble getting that to work before, but I can take another attempt at it.  Since this is internal to the model, maybe we can merge this roughly as-is (after I change LogLevel to ECSLogLevel) and I can ask you for help with removing the struct later?

packages/ecs-agent/ecs.service
samuelkarp: Done.

sources/models/src/modeled_types/mod.rs
samuelkarp: I added a "field" string and removed the "input" string, as `serde_plain::Error` includes the invalid input.

sources/models/src/modeled_types/ecs.rs
samuelkarp: Did the rename, was unable to get the enum-sans-struct working.

sources/models/src/modeled_types/ecs.rs
tjkirch: I'd probably include them all because we don't know how the output will change in the future (could get more verbose) or what people want to see in their logs/console.

sources/models/src/modeled_types/ecs.rs
```diff
@@ -5,6 +5,7 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use super::error;
 use serde::de::Error as _;
 use snafu::ensure;
+use snafu::ResultExt;
```
tjkirch:
```suggestion
use snafu::{ensure, ResultExt};
```

tjkirch (approved):

sources/models/src/modeled_types/ecs.rs
tjkirch: I think it will work like this, if you remove the `string_impls_for` that doesn't apply to an enum.  (As long as you can get/set the setting through the API, it should be fine; it compiles!)

```
#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Serialize)]
#[serde(rename_all = "lowercase")]
pub enum ECSAgentLogLevel {
```

samuelkarp: I tried that and got some errors I don't fully understand from the API server:

```
$ apiclient -u /settings -m PATCH -d '{"ecs":{"loglevel":"warn"}}'
Status 500 when PATCHing unix://2f72756e2f6170692e736f636b:0/settings: Error serializing Settings: 'unit variant' not allowed by Serializer 
```

sources/models/src/modeled_types/ecs.rs
samuelkarp: Done

sources/models/src/modeled_types/ecs.rs
samuelkarp: Done

tjkirch: Don't want `none`?  :)

samuelkarp: Not really; `crit` should be very quiet and if something comes up at that log level there is something seriously wrong with the ECS agent.  That log level isn't included in [the documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html) either.

sources/models/src/modeled_types/ecs.rs
tjkirch: Hmm.  It's something we could work on as an improvement to apiserver (datastore serialization) to support [this](https://github.com/bottlerocket-os/bottlerocket/blob/ff88228b88ac0093b22927a3f013db51c8f93b78/sources/api/apiserver/src/datastore/serialization/pairs.rs#L179).  I think the deserialization would probably work, we'd just have to decide how to consistently serialize enum variants.  Since they're unit variants (no internal data) I think just serializing the name of the variant would work, it's something we could try.

PR #1092, opened 2020-08-30
switch to SDK v0.13.0

bcressey:
**Issue number:**
N/A

**Description of changes:**
This updates to SDK v0.13.0, and includes fixes for building Bottlerocket on a Graviton instance.

The major change is the switch to using our vendor Rust targets, which unblocks the glibc update.

I found the zstd compression warning when inspecting the build logs for grub, which needed a fix to avoid building x86_64 tools on an aarch64 host.

**Testing done:**
Verified that an x86_64 host could build the OS for both targets, and likewise that an aarch64 host could build the OS for both targets. Verified that the resulting artifacts worked as expected when registered as AMIs.

Note: CI will fail until bottlerocket-os/bottlerocket-sdk/pull/33 is merged and the SDKs are built.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -100,7 +100,7 @@ mkdir -p ${BUILDSYS_OUTPUT_DIR}
 mkdir -p ${BUILDSYS_PACKAGES_DIR}
 mkdir -p ${GO_MOD_CACHE}
 
-for cmd in bash curl docker gunzip ; do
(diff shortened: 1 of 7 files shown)


tjkirch (approved):
This is a huge win for the build process, thank you Ben!

webern (approved):

iliana (approved):

PR #1100, opened 2020-09-02
Added option to enable spot instance draining

### LINKED ISSUE #1099: [ECS] SpotInstanceDraining Agent Configuration
**What I'd like:**

I need to be able to set this parameter to `true` via user data, to evict workload safely from to-be-terminated spot instances in my cluster

This agent functionality could be implemented in bottlerocket (see https://github.com/reach-now/spot-termination-handler), but I would not see a good reason why this would be preferable.

mkulke:
**Issue number:** 

 fixes #1099

**Description of changes:**

Added the option to have instance draining on spot instances, I used https://github.com/aws/amazon-ecs-agent/blob/a250409cf5eb4ad84a7b889023f1e4d2e274b7ab/agent/config/types.go as reference

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -389,6 +389,7 @@ These settings can be changed at any time.
   Bottlerocket sets this value to false by default. 
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
 
 #### Updates settings
 
--- a/sources/api/ecs-settings-applier/src/ecs.rs
+++ b/sources/api/ecs-settings-applier/src/ecs.rs
@@ -45,6 +45,9 @@ struct ECSConfig {
 
     #[serde(rename = "OverrideAWSLogsExecutionRole")]
     override_awslogs_execution_role: bool,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    spot_instance_draining_enabled: Option<bool>,
 }
 
 // Returning a Result from main makes it print a Debug representation of the error, but with Snafu
@@ -79,6 +82,7 @@ fn run() -> Result<()> {
             .iter()
             .map(|s| s.to_string())
             .collect(),
+        spot_instance_draining_enabled: ecs.enable_spot_instance_draining,
 
         // Task role support is always enabled
         task_iam_role_enabled: true,
--- a/Release.toml
+++ b/Release.toml
@@ -1,4 +1,4 @@
-version = "1.0.1"
+version = "1.0.2"
 
 [migrations]
 "(0.3.1, 0.3.2)" = ["migrate_v0.3.2_admin-container-v0-5-0.lz4"]
@@ -9,3 +9,4 @@ version = "1.0.1"
 "(0.4.1, 0.5.0)" = ["migrate_v0.5.0_add-cluster-domain.lz4", "migrate_v0.5.0_admin-container-v0-5-2.lz4", "migrate_v0.5.0_control-container-v0-4-1.lz4"]
 "(0.5.0, 1.0.0)" = ["migrate_v1.0.0_ecr-helper-admin.lz4", "migrate_v1.0.0_ecr-helper-control.lz4"]
 "(1.0.0, 1.0.1)" = []
+"(1.0.1, 1.0.2)" = ["migrate_v1.0.2_add-enable-spot-instance-draining.lz4"]
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -250,6 +250,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "add-enable-spot-instance-draining"
+version = "0.1.0"
+dependencies = [
+ "migration-helpers 0.1.0",
+]
+
+[[package]]
 name = "add-version-lock-ignore-waves"
 version = "0.1.0"
 dependencies = [
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -27,6 +27,7 @@ members = [
     "api/migration/migrations/v0.5.0/add-cluster-domain",
     "api/migration/migrations/v1.0.0/ecr-helper-admin",
     "api/migration/migrations/v1.0.0/ecr-helper-control",
+    "api/migration/migrations/v1.0.2/add-enable-spot-instance-draining",
 
     "bottlerocket-release",
 
--- a/sources/api/migration/migrations/v1.0.2/add-enable-spot-instance-draining/Cargo.toml
+++ b/sources/api/migration/migrations/v1.0.2/add-enable-spot-instance-draining/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "add-enable-spot-instance-draining"
+version = "0.1.0"
+authors = ["Magnus Kulke <mkulke@gmail.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2018"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers" }
--- a/sources/api/migration/migrations/v1.0.2/add-enable-spot-instance-draining/src/main.rs
+++ b/sources/api/migration/migrations/v1.0.2/add-enable-spot-instance-draining/src/main.rs
@@ -1,0 +1,22 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddSettingsMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting, `ecs.enable-spot-instance-draining`
+fn run() -> Result<()> {
+    migrate(AddSettingsMigration(&[
+        "settings.ecs.enable-spot-instance-draining",
+    ]))
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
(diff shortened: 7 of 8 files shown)


README.md
```diff
@@ -389,6 +389,8 @@ These settings can be changed at any time.
   Bottlerocket sets this value to false by default. 
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.spot-instance-draining-enabled`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
```
bcressey: I'd prefer the imperative form here, something like:
```suggestion
* `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
```

README.md
```diff
@@ -389,6 +389,8 @@ These settings can be changed at any time.
...
+  Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
```
zmrow: I believe this sentence was added by mistake and isn't applicable to this setting. :)

mkulke: sry :) yes

sources/api/ecs-settings-applier/src/ecs.rs
```diff
@@ -45,6 +45,12 @@ struct ECSConfig {
 
     #[serde(rename = "OverrideAWSLogsExecutionRole")]
     override_awslogs_execution_role: bool,
+
+    #[serde(
+        rename = "SpotInstanceDrainingEnabled",
+        skip_serializing_if = "Option::is_none"
+    )]
+    enable_spot_instance_draining: Option<bool>,
```
samuelkarp: This struct should match the generated JSON, so we can rely on the default serde transformation except for the cases where the expected JSON doesn't match that.  (The ECS agent generally follows the [Go convention for initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) that I don't believe serde knows how to handle.)

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    spot_instance_draining_enabled: Option<bool>,
```

samuelkarp (commented):
Thanks so much for opening this pull request!  We’re super excited to see the interest in the ECS variant of Bottlerocket, and adding new settings is a great way to help out!  There are a few more things that will need to be done to take this PR across the finish line.  I’m happy to describe them here; if you want to handle implementing them that would be awesome, but if you’d prefer to have us do it that’s also fine (though I can’t guarantee _when_ we’ll be able to get to it).  Let us know which you’d prefer!

You can take a look at [this commit](https://github.com/bottlerocket-os/bottlerocket/commit/9b7955ac4f4822800a83bcc944df30aa2a990bc9) for inspiration, as it has most of the things you’ll want to do.  Specifically:

* Add the new setting to `sources/models/src/lib.rs` in the `ECSSettings` struct.  This is how we define the settings that show up in the Bottlerocket API and hook it up with the component that can read settings from user-data.
* Read the new setting and inject it into the generated `ECSConfig` struct in `sources/api/ecs-settings-applier/src/ecs.rs`; you can see that on lines 79-100 of the file
* (Not shown in the commit I linked you to) Bottlerocket settings are effectively stored in a database with a strict schema; adding new settings changes that schema.  We’ll need a migration program (like [in here](https://github.com/bottlerocket-os/bottlerocket/tree/develop/sources/api/migration), [this](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/migration/migrations/v0.5.0/add-cluster-domain/src/main.rs) is a good example) to handle adding the new setting to the schema.
* You _don’t_ need to change `sources/models/src/aws-ecs-1/override-defaults.toml` even though it's show in the other commit; the default can remain unset.

mkulke:
@samuelkarp Thanks a ton for the kind words, a very comprehensive review and concrete pointers! I suppose I missed a layer of indirection, things makes more sense now. I _think_ I covered the points you listed, however i'm not sure about the versioning scheme of the project: I just put the migration in a `v1.0.1` folder.

sources/api/ecs-settings-applier/src/ecs.rs
```diff
@@ -90,6 +93,7 @@ fn run() -> Result<()> {
         // Always supported with Docker newer than v17.11.0
         // See https://github.com/docker/engine/commit/c7cc9d67590dd11343336c121e3629924a9894e9
         override_awslogs_execution_role: true,
+        spot_instance_draining_enabled: ecs.enable_spot_instance_draining,
```
samuelkarp: nit: I'd move this up above line 86 so that the constant values are still at the end

samuelkarp (commented):
Thanks so much, this looks super good!  The last step I think is just for us to test and make sure the following scenarios are all working properly:

* upgrade from 1.0.0 and make sure the new setting is now available in the API
* downgrade from 1.0.1 to 1.0.0 and make sure the new setting is removed
* when enabling spot instance draining through the API, make sure the right codepath in the agent is executing

This is work that we can take care of; staging updates in a repo to execute upgrade-downgrade testing is not super trivial yet.  But if you're able to build an AMI (`cargo make && cargo make ami`) you can test whether the spot instance draining feature functions on a spot instance.

mkulke:
@samuelkarp

Spot termination notices are hard to test, i'm not sure I can trigger such a notification myself. However, i built an AMI, started an EC2 instance w/ user data:

```
enabled = true

[settings.ecs]
cluster = "redacted"
enable-spot-instance-draining = true
```

...and it registered succesfully as a cluster instance, so the agent works in principle. The agent's introspection api does not seem to provide configuration details (possibly because there might be secrets in there):

```bash
[ec2-user@ip-redacted ~]$ curl -s http://localhost:51678/v1/metadata
{"Cluster":"redacted","ContainerInstanceArn":"arn:aws:ecs:eu-west-1:redacted","Version":"Amazon ECS Agent - v1.43.0 (1ebf0604)"}
```

However I checked whether the param is properly written to the config and that seems to be the case:

```bash
[ec2-user@ip-redacted ~]$ sudo sheltie
bash-5.0# cat /etc/ecs/ecs.config.json
{"Cluster":"redacted","InstanceAttributes":{"bottlerocket.version":"1.0.0","bottlerocket.variant":"aws-ecs-1"},"PrivilegedDisabled":true,"AvailableLoggingDrivers":["json-file","awslogs","none"],"TaskIAMRoleEnabled":true,"TaskIAMRoleEnabledForNetworkHost":true,"SELinuxCapable":true,"OverrideAWSLogsExecutionRole":true,"SpotInstanceDrainingEnabled":true}
```

samuelkarp:
> Spot termination notices are hard to test, i'm not sure I can trigger such a notification myself.

I didn't know this until today, but [Spot blocks](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances) allow specifying a duration for your instance to run, and termination notices are delivered at the end of the block.  I was able to test an AMI based on your code, and correctly saw the ECS agent flip the container instance's state to `DRAINING` roughly two minutes before it was terminated.

I haven't had a chance to run the other test I want, which is upgrade-downgrade to check the migration.  (I'm pretty confident that it'll work, but I still want to run the test.)

The other thing that needs to be taken care of (and we can take care of this if you don't want to) is that the migration needs to be moved to 1.0.2 since we've [released 1.0.1](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.0.1).

mkulke:
> I didn't know this until today, but [Spot blocks](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances) allow specifying a duration for your instance to run, and termination notices are delivered at the end of the block. I was able to test an AMI based on your code, and correctly saw the ECS agent flip the container instance's state to `DRAINING` roughly two minutes before it was terminated.

Ah, wasn't aware, either. This will be helpful for testing spot tools.

> I haven't had a chance to run the other test I want, which is upgrade-downgrade to check the migration. (I'm pretty confident that it'll work, but I still want to run the test.)

👍

> The other thing that needs to be taken care of (and we can take care of this if you don't want to) is that the migration needs to be moved to 1.0.2 since we've [released 1.0.1](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.0.1).

will do

samuelkarp:
I was able to test upgrade-downgrade and exercise the migration successfully, but there's one change I had to make: the `Release.toml` file needs to list the migration in a new line at the end like this:

```
"(1.0.1, 1.0.2)" = ["migrate_v1.0.2_add-enable-spot-instance-draining.lz4"]
```

Can you add that to the PR?  I'll be happy to approve it after. :smile:

samuelkarp (approved):
Thank you!

zmrow (approved):
:unicorn: 

Thanks!

README.md
```diff
@@ -389,6 +389,7 @@ These settings can be changed at any time.
   Bottlerocket sets this value to false by default. 
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
```
tjkirch, line 392:
```suggestion
* `settings.ecs.enable-spot-instance-draining`: If the instance receives a [Spot](https://aws.amazon.com/ec2/spot/) termination notice, the agent will set the instance's state in ECS to `DRAINING` so the workload can be moved gracefully.  Defaults to `false`.  (The instance will need to have permission to call the UpdateContainerInstancesState API, per [ECS documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-instance-draining.html).)
```
I think we should mention the permission that the instance has to have, so the user knows what they have to set up beforehand.

tjkirch (approved):
Thanks!

tjkirch (approved):

samuelkarp:
Hi @mkulke! Would you mind squashing your commits into a single commit? Then we can merge from there.

(If we don't hear back from you in a couple days, we'll take care of the squash and merge.)

PR #1121, opened 2020-09-15
pass Go module proxy variables through docker-go

bcressey:
**Issue number:**
N/A

**Description of changes:**
Previously, we always set `GOPRIVATE=*`, which blocked any use of the public Go module mirror.

Now we pass the module proxy variables through from the environment, so that the behavior is under the developer's control.

**Testing done:**
None, yet.

**Terms of contribution:**

### DIFF
--- a/tools/docker-go
+++ b/tools/docker-go
@@ -54,18 +54,31 @@ DOCKER_RUN_ARGS="--network=host"
 
 parse_args "${@}"
 
+# Pass through relevant Go variables, from the config or environment.
+go_env=( )
+for i in GOPROXY GONOPROXY GOPRIVATE ; do
+  if command -v go >/dev/null 2>&1 ; then
+    govar="$(go env ${i})"
+    if [ -n "${govar}" ] ; then
+      go_env[${#go_env[@]}]="--env=${i}=${govar}"
+    fi
+  elif [ -n "${!i}" ] ; then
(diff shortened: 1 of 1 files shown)


tools/docker-go
```diff
@@ -56,14 +56,15 @@ parse_args "${@}"
 
 # Go accepts both lower and uppercase proxy variables, pass both through.
 proxy_env=( )
-for i in http_proxy https_proxy no_proxy HTTP_PROXY HTTPS_PROXY NO_PROXY; do
-   if [ -n "${!i}" ]; then
-      proxy_env[${#proxy_env[@]}]="--env=$i=${!i}"
-   fi
+for i in \
+    http_proxy https_proxy no_proxy HTTP_PROXY HTTPS_PROXY NO_PROXY \
+    goproxy gonoproxy goprivate GOPROXY GONOPROXY GOPRIVATE ; do
```
samuelkarp: If a `go` binary is present on the host where this is running, the values for these variables should be read from `go env`.  This enables the configuration values to be defined in a session-independent way in the Go tool's config file.  (It's also where I personally configure my value for `GOPROXY`, so reading just the environment variable won't pick up the value I've configured.)

samuelkarp (approved):

iliana (approved):

PR #1133, opened 2020-09-25
pubsys: automate setup of role and key

zmrow:
Reviewing these commits separately will probably be easiest.

**Description of changes:**
```
commit dafe3b16c7b8f3301ce03dfe40703e2bad5c0bc0 
Date:   Thu Sep 24 20:35:12 2020 +0000

    pubsys-config: disallow unknown fields

    Without this, a user could have an old-format Infra.toml files with root role
    and key defined at the top level, and pubsys wouldn't error, so the user might
    think they're being used.  With this change, we give a clear error so the user
    knows what to update.

    Co-authored-by: Zac Mrowicki <mrowicki@amazon.com>
    Co-authored-by: Tom Kirchner <tjk@amazon.com>

commit a04cd78f6e23665e434741a9a671007aa8453674
Date:   Thu Sep 24 18:40:32 2020 +0000

    Remove root.json from repo

    The official Bottlerocket root.json will live at:
    https://cache.bottlerocket.aws/root.json
    and be used by the Bottlerocket team for public releases.

    Note: if you don't have access to the official Bottlerocket signing keys, you
    could still reference that URL in your Infra.toml config file and build a
    Bottlerocket image that's able to update to official images, but you wouldn't
    be able to create a repo.  For your own repos, you can leave root role
    unspecified in your Infra.toml, and pubsys-setup will automatically be run to
    create one for you.

    Co-authored-by: Zac Mrowicki <mrowicki@amazon.com>
    Co-authored-by: Tom Kirchner <tjk@amazon.com>

commit 165b0fe326d020822797cc8fb26a4dceed16b380
Date:   Thu Sep 24 18:36:15 2020 +0000

    Add pubsys-setup to ease creation of root role and signing key

    If the user doesn't define a root role and key themselves in Infra.toml, we'll
    generate them so it's easier to get started building images and repos.  The
    user should update Infra.toml with more permanent resources and locations, but
    this makes it much easier to get started handling your own builds.

    Note: this changes the format of Infra.toml to make root roles and keys
    specific to a repo.  Roles and keys are necessarily tied to a repo, so this
    is simpler than having to specify PUBLISH_REPO *and* PUBLISH_KEY (and manually
    changing the root.json file, which wasn't configurable before).

    Co-authored-by: Zac Mrowicki <mrowicki@amazon.com>
    Co-authored-by: Tom Kirchner <tjk@amazon.com>

commit ff46293d469d544a437af1843cea5027e9f62144
Date:   Thu Sep 24 18:21:15 2020 +0000

    pubsys: split pubsys-config into new library crate

    Co-authored-by: Zac Mrowicki <mrowicki@amazon.com>
    Co-authored-by: Tom Kirchner <tjk@amazon.com>
```

**Testing done:**
* Ran a build OK and saw it generate key and role                                                                                                                                                               
* Ran again, it found the existing ones                                                                                                                                                                         
* Deleted the role, ran again, it created a new role from the existing key                                                                                                                                      
* Ran with Infra.toml pointing to the URL of the production root role, it downloaded/verified/used it OK                                                                                                        
                                                                                                                                                                                                                
* Put the key in SSM, deleted it locally, set SSM key in Infra.toml, and confirmed that it could create/sign a new role with it                                                                                 
* Put the key in KMS, deleted it locally, set KMS key in Infra.toml, and confirmed that it could create/sign a new role with it                                                                                 
                                                                                                                                                                                                                
* Tested that we got a hash mismatch with a local root role not matching the URL                                                                                                                                
* Tested that Infra.toml role and key override local files                                                                                                                                                      
                                                                                                                                                                                                                
* Tested a repo build                                                                                                                                                                                           
* Tested changing PUBLISH_REPO and seeing a new key/role being built for it, and confirmed updog.rpm has the new root.json                                                                                      
* Tested touching root.json and seeing it rebuild the os package                                                                                                                                                
                                                                                                                                                                                                                
* Tested building repos with KMS/SSM keys


**Terms of contribution:**


etungsten (approved):
:seedling: :deciduous_tree:

tjkirch (approved):

bcressey:
Can you expand the commit message for `pubsys-config: disallow unknown fields`?

Did we run into a problem? Should we be adding `#[serde(deny_unknown_fields)]` in other places?

.gitignore
```diff
@@ -7,3 +7,6 @@
 /html
 /Infra.toml
 /*.pem
+/tools/bin
```
bcressey: Should we remove the equivalent exclusion from `tools/.gitignore`, or move this one there?

Makefile.toml
```diff
@@ -87,6 +85,12 @@ BUILDSYS_VERSION_FULL="${BUILDSYS_VERSION_IMAGE}-${BUILDSYS_VERSION_BUILD}"
 BUILDSYS_NAME_VARIANT="${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH}"
 BUILDSYS_NAME_VERSION="${BUILDSYS_NAME}-${BUILDSYS_VERSION_FULL}"
 BUILDSYS_NAME_FULL="${BUILDSYS_NAME_VARIANT}-${BUILDSYS_VERSION_FULL}"
+# Relative path to repo-specific root role.  This gets mapped into the package
+# build environment so it can be included in the image, and Docker only accepts
+# relative paths for bind mounts.
+PUBLISH_REPO_ROOT_JSON_RELATIVE = "roles/${PUBLISH_REPO}.root.json"
```
bcressey: Comment is a bit confusing.

Is the issue that Docker wants files for `COPY` to come from the build context so it can do change tracking?

If so, I expect Docker would also be unhappy about a "../" style relative path.

I don't love the variable name, so part of this is seeing if we can chop off the "_RELATIVE" suffix.

Makefile.toml
```diff
@@ -194,6 +198,21 @@ cargo install \
 '''
 ]
 
+# Note: this is separate from publish-tools because publish-tools takes a while
+# to build and isn't needed to build an image.
```
bcressey, line 200: How much time does this save? Several minutes?

tjkirch:
> Can you expand the commit message for `pubsys-config: disallow unknown fields`?
> 
> Did we run into a problem?

That was to catch people running with old-format Infra.toml files, so that pubsys would stop with a clear error, rather than continuing and letting them think that the root role and key they defined at the top level of the config was being used.  We can expand the commit message to explain.

> Should we be adding `#[serde(deny_unknown_fields)]` in other places?

It's a good thing to consider on Deserialize structures in general.  In this case it seemed important to let users know when we're not using fields from their Infra.toml that they need used.  We also use it in API structures, like settings, for the same reason.  I don't think it's needed on every Deserialize, but it's important to know it's an option.

tools/pubsys/Infra.toml.example
```diff
@@ -2,30 +2,32 @@
...
+# You can have any number of repos defined and build a specific one by running like this:
+#     cargo make repo -e PUBLISH_REPO=myrepo
+[repo.default]
+# URL to your root role JSON file; can be a file:// URL for local files.  If
+# you don't specify one here, a file will be generated for you under /roles.
+# For production use, you should store them somewhere safer.
+# (The current Bottlerocket root.json lives at https://cache.bottlerocket.aws/root.json)
+root_role_url = "https://example.com/root.json"
+# SHA512 checksum of your root role JSON file.
+root_role_sha512 = "0123456789abcdef"
```
bcressey: Maybe add our root.json and the current hash in a commented-out example?

tools/pubsys/Infra.toml.example
```diff
@@ -2,30 +2,32 @@
...
 
-# You would normally create repo signing keys using `tuftool root gen-rsa-key`
-# as part of the initial setup of your TUF roles.  pubsys assumes a single
-# publication key that lives in the snapshot, targets, and timestamp roles.
-# Here you specify where that key lives so we can sign the created repo.
-# You can specify multiple keys, if you like, and select one by name:
-#     cargo make repo -e PUBLISH_KEY=mysigningkey
-# (Don't keep your keys in the repo!)
-# You can keep your private key in a file or an SSM parameter; pick one:
+# pubsys assumes a single publication key that lives in the snapshot, targets,
```
bcressey:
```suggestion
# pubsys assumes a single publication key that signs for the snapshot, targets,
```

tools/pubsys-config/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+    type Error = ();
+    fn try_from(key: SigningKeyConfig) -> std::result::Result<Self, Self::Error> {
+        match key {
+            SigningKeyConfig::file { path } => Url::from_file_path(path),
+            SigningKeyConfig::kms { key_id } => {
+                // Note: we add a third slash here because we expect most people will enter a KMS
+                // key ID without a slash; tough assumes anything before an initial slash is a
+                // profile name, which we don't support here.
+                Url::parse(&format!("aws-kms:///{}", key_id)).map_err(|_| ())
+            }
```
bcressey: Profile here means an `~/.aws/config` profile?

If we end up adding profile support for KMS then it'll be a breaking change - I'd like a better story here, even if it's just that we don't support profiles this way and you can only set them through the `AWS_PROFILE` environment variable.

tools/pubsys-config/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+                // Note: we don't add a third slash here because we expect most people will enter
+                // an SSM parameter name with a leading slash.  (They could specify a profile name
+                // before the slash if they wanted, but that's not recommended here.)
+                Url::parse(&format!("aws-ssm://{}", parameter)).map_err(|_| ())
+            }
+        }
+    }
+}
+
+/// Location of existing published repo
```
bcressey: This comment is out of date.

tools/pubsys-setup/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+                info!("Didn't find root role in Infra.toml, generating...");
+            } else {
+                info!("Didn't find root role or signing key in Infra.toml, generating...");
+            }
+
+            let role_dir = args.root_role_path.parent().context(error::Path {
+                path: &args.root_role_path,
+                thing: "root role",
+            })?;
+            fs::create_dir_all(role_dir).context(error::Mkdir { path: role_dir })?;
```
bcressey: Might be cleaner to create this at the outset and pass `role_dir` to `find_root_role_and_key`, rather than duplicating this bit.

tools/pubsys-setup/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+
+            let temp_path = NamedTempFile::new_in(&role_dir).context(error::TempFileCreate {
+                purpose: "root role",
+            })?;
+            let temp_path_display = temp_path.path().display();
```
bcressey: Maybe `temp_root_role` ?

`temp_path_display` is pretty meaningless for a variable that gets used so often.

tjkirch:
@bcressey we've updated the commit message to explain why we added those attributes.

tools/pubsys-setup/src/main.rs
```diff
@@ -0,0 +1,384 @@
...
+                tuftool!("root add-key '{}' '{}' --role root --role snapshot --role targets --role timestamp",
+                         temp_path_display, key_url);
+                key_url
+            } else {
+                // If the user has no key, build one and add it to each role.
+                let key_dir = args.default_key_path.parent().context(error::Path {
+                    path: &args.default_key_path,
+                    thing: "key",
+                })?;
+                fs::create_dir_all(key_dir).context(error::Mkdir { path: key_dir })?;
```
bcressey: nit: if we did this `mkdir` somewhere else, it would make the two branches a bit more symmetrical.

Also, I know it's wasteful, but there'd be a lot less branching if we generated the key in `find_root_role_and_key`, possibly by calling `tuftool root gen-rsa-key ...` with a temp file and throwing away the root.json we don't need.

We could wrap the initial few `tuftool` commands in a function.

tools/pubsys/src/repo.rs
```diff
@@ -430,30 +437,34 @@ pub(crate) fn run(args: &Args, repo_args: &RepoArgs) -> Result<()> {
...
+        Some(SigningKeyConfig::ssm { parameter }) => Box::new(SsmKeySource {
             profile: None,
             parameter_name: parameter.clone(),
             key_id: None,
         }),
+        None => {
+            ensure!(
+                repo_args.default_key_path.exists(),
+                error::MissingConfig {
+                    missing: "signing_keys in repo config, and we found no local key",
```
bcressey, line 461: Should this be "no signing_keys in repo config..."?

.gitignore
tjkirch: Sorry, good catch; forgot about the separate .gitignore file.

Makefile.toml
tjkirch: Good catch; I'm not sure we need this to be relative anymore, now that we're COPYing the file rather than mapping.  It also really bugged me, but I forgot to go back to it after switching back to COPY.

[edit] this isn't correct, it still has to be a relative path for COPY.  Thinking.

Makefile.toml
tjkirch: On a laptop, maybe 10 minutes - it's enough of a harm to the new-user experience that I thought the separate binary was worth it.

tools/pubsys-config/src/lib.rs
tjkirch: Yeah, a local profile.  I do think the story is "that we don't support profiles this way", which is what I meant by `which we don't support here`.  (tough and pubsys just handle profiles differently, and profiles aren't as important to pubsys.)

tools/pubsys-setup/src/main.rs
tjkirch: `temp_root_role` and `temp_root_role_path`?

tools/pubsys/src/repo.rs
tjkirch: I don't think so, because we're saying what's missing, not what's present... "*missing*: signing_keys ..."

[edit] To be clear, the printed error message also clarifies - this would print `Infra.toml is missing signing_keys in repo config, and we found no local key`

webern:
> old-format Infra.toml files

Back when we first thought about Infra.toml we had discussed versioning its schema. Does it make sense for us to think about adding in an Infa.toml schema version to the file? (Not as part of this PR, of course)

webern (approved):

tools/pubsys-config/src/lib.rs
tjkirch: After discussion, we're going to make these consistent by prepending a slash to both ssm and kms names if one isn't given.  Profiles won't be supported until we figure out a better way to interface them with tough.

Makefile.toml
tjkirch: After discussion, we're going to pass through PUBLISH_REPO instead, which we can use in the COPY line, and we'll form the (relative) path in the COPY call instead of passing the relative path through the whole way.

tools/pubsys-setup/src/main.rs
tjkirch: Our push moved the mkdirs, which does clarify the code, but we didn't think it made as much sense to generate a key inside the find function.  It seems unexpected for that to happen given the purpose of that function.  The conditional in the `run` function that handles key generation is relatively small, especially now that the mkdir is moved.  It seems to be pretty clear now.

tjkirch:
^ This push addresses @bcressey's comments:
* Moved .gitignore of `/tools/bin` into `tools/.gitignore`
* Removed `_RELATIVE` variable by passing REPO through instead
* Added commented-out values for Bottlerocket root.json in `Infra.toml.example`
* Fixed comment about key "living in" a role
* Prefix SSM parameters / KMS keys with slashes rather than inconsistently allowing profile names
* Updated comment about repo structure
* Create roles/keys directories once rather than spread through pubsys-seutp
* Clarify variable names referring to temp file for root role

tjkirch:
> Back when we first thought about Infra.toml we had discussed versioning its schema. Does it make sense for us to think about adding in an Infa.toml schema version to the file? (Not as part of this PR, of course)

@webern I think it could make sense for some future cases, but I don't think we'd want to now.  Pretty soon we'll have all the main components of pubsys implemented, and it may not need to change much after.  We had some discussion of using Infra.toml for other tools as well, but I don't expect it to be widespread, or for the file to have rapidly changing structure after building them.  Versioning is relatively easy to add later, given that we'd probably have one breaking change anyway, so I don't think we have to bake it in now.

bcressey (approved):

PR #1137, opened 2020-09-30
Update README.md

PrashantMehta-coder:
**Issue number:**

**Description of changes:**

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Some notable features include:
 
 ## Contact us
 
-If you find a security issue, please [contact our security team](https://github.com/bottlerocket-os/bottlerocket/security/policy) rather than opening an issue.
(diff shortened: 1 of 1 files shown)


tjkirch:
Hi @PrashantMehta-coder, thanks for your interest!  Would you please explain your thoughts around this change?  I think the original is a more common wording, but I'm interested to hear what you think.

PrashantMehta-coder:
Thanks for the response, common words are used by everyone.. But, we should use some complex words which will increase our knowledge of words and grammar level.

tjkirch:
In Bottlerocket docs, we aim for comfortable, common language that's easier to understand, so the reader can focus on the important parts.  So, in this case we're going to stick with the more common word.  Thanks though!

PrashantMehta-coder:
Yahh.. You are 100% right.

PR #1138, opened 2020-10-01
Add PUBLISHING.md guide explaining pubsys and related tools

zmrow:
**Testing done:**
* Checked all the links 
* Read through multiple times

**Terms of contribution:**

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -64,7 +64,10 @@ cargo make
 ```
 
 This will build an image for the default variant, `aws-k8s-1.17`.
-All packages will be built in turn, and then compiled into an `img` file in the `build/` directory.
+All packages will be built in turn, and then compiled into an `img` file in the `build/images/` directory.
+
+The version number in [Release.toml](Release.toml) will be used in naming the file, and will be used inside the image as the release version.
+If you're planning on [publishing your build](PUBLISHING.md), you may want to change the version.
 
 To build an image for a different variant, run:
 
@@ -72,36 +75,38 @@ To build an image for a different variant, run:
 cargo make -e BUILDSYS_VARIANT=my-variant-here
 ```
 
+To build an image for a different architecture, run:
+
+```
+cargo make -e BUILDSYS_ARCH=my-arch-here
+```
+
+(You can use variant and arch arguments together, too.)
+
 ### Register an AMI
 
 To use the image in Amazon EC2, we need to register the image as an AMI.
-The `bin/amiize.sh` script does this for you.
 
-The script has some assumptions about your setup, in particular that you:
-  * have [aws-cli](https://aws.amazon.com/cli/) set up, and that its default profile can create and control EC2 resources
-  * have [coldsnap](https://github.com/awslabs/coldsnap/) installed to upload snapshots
-  * have a few other common tools installed, like `jq` and `du`
-
-First, decompress the images.
-(Note: these filenames assume an `x86_64` architecture and `aws-k8s-1.17` [variant](README.md).)
+For a simple start, pick an [EC2 region](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions), then run:
 
 ```
-lz4 -d build/images/x86_64-aws-k8s-1.17/latest/bottlerocket-aws-k8s-1.17-x86_64.img.lz4 && \
-lz4 -d build/images/x86_64-aws-k8s-1.17/latest/bottlerocket-aws-k8s-1.17-x86_64-data.img.lz4
+cargo make ami -e PUBLISH_REGIONS=your-region-here
 ```
 
-Next, register an AMI:
+Your new AMI ID will be printed after it's registered.
+
+If you built your image for a different architecture or variant, just use the same arguments here:
 
 ```
-bin/amiize.sh --name YOUR-AMI-NAME-HERE \
-              --arch x86_64 \
-              --region us-west-2 \
-              --root-image build/images/x86_64-aws-k8s-1.17/latest/bottlerocket-aws-k8s-1.17-x86_64.img \
-              --data-image build/images/x86_64-aws-k8s-1.17/latest/bottlerocket-aws-k8s-1.17-x86_64-data.img
+cargo make ami -e PUBLISH_REGIONS=your-region-here -e BUILDSYS_VARIANT=my-variant-here
 ```
 
-Your new AMI ID will be printed at the end.
+(There's a lot more detail on building and managing AMIs in the [PUBLISHING](PUBLISHING.md) guide.)
 
 ## Use your image
 
 See the [setup guide for Kubernetes](QUICKSTART-EKS.md) or the [setup guide for Amazon ECS](QUICKSTART-ECS.md) for information on running Bottlerocket images.
+
+## Publish your image
+
+See the [PUBLISHING](PUBLISHING.md) guide for information on deploying Bottlerocket images and repositories.
--- a/PUBLISHING.md
+++ b/PUBLISHING.md
@@ -1,0 +1,396 @@
+# Publishing Bottlerocket
+
+This guide will walk you through deploying a Bottlerocket image, and if desired, sharing it with others.
+It currently focuses on deploying to AWS, though the tooling is built to support other platforms in the future.
+
+Remember to look at the [TRADEMARKS](TRADEMARKS.md) guide to understand naming concerns.
+You can pass `-e BUILDSYS_NAME=my-name` to `cargo make` commands to change the default name, which is used in file and AMI names.
+
+We'll assume you've been through the [BUILDING](BUILDING.md) guide to make an image.
+
+### Configuring the publishing process
+
+The publishing process uses a configuration file called `Infra.toml`.
+The relevant sections of this file will be introduced as needed below.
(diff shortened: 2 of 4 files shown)


tjkirch (approved):

BUILDING.md
```diff
@@ -64,44 +64,48 @@ cargo make
 ```
 
 This will build an image for the default variant, `aws-k8s-1.17`.
-All packages will be built in turn, and then compiled into an `img` file in the `build/` directory.
+All packages will be built in turn, and then compiled into an `img` file in the `build/images/` directory.
+The image will receive the version number from the `Release.toml` file.
```
etungsten:
```suggestion
The image file names will include the version number from the `Release.toml` file.
```

etungsten: I guess I'm just unsure what "receive" means in this context.

PUBLISHING.md
```diff
@@ -0,0 +1,392 @@
...
+```
+
+### Roles and keys
+
+#### Background on roles and keys
+
+If you run the `cargo make repo` command above without any configuration, it will generate a root role file and a signing key for you.
+The key is used to sign the metadata that lives in the repo, and the root role is included in the image and used to verify that trusted keys are being used to sign the repo in the future.
+
+The generated role and key are functional, but a bit basic.
```
etungsten, line 120: I think maybe a reference to the TUF spec might be helpful here to help people understand the context?

tjkirch: The spec itself is pretty unfriendly, so I was trying to avoid that.  We link to their official docs below, and to our updater docs above, which explain TUF and how we use it.   How do you think we could best incorporate a link in this section, and where would it point?

BUILDING.md
tjkirch: It's just not the image file names, though - it propagates through buildsys, to Dockerfile VERSION_ID, into rpm2img, and from there into os-release, which shows up in the API and some other places in the OS.  "receive" was a generic way of saying that the version of the image (wherever that version is used) is this one.  What do you think, how can we clarify?

etungsten: I see, then can we explain just a little bit more about how the version number is used within the image? Perhaps,
 ```suggestion
The image's OS release version will be set to the `version` from the [`Release.toml` file](https://github.com/bottlerocket-os/bottlerocket/blob/develop/Release.toml).
```
(I can't think of anywhere else in Bottlerocket besides /etc/os-release and `apiclient -u /os` that uses the version number)

File: PUBLISHING.md (hunk 2 above)
Comment (etungsten) [reply PUBLISHING.md:120]: I think we should maybe move the following line to the beginning of this section:
 https://github.com/bottlerocket-os/bottlerocket/blob/eb855a2de16e7f1fc16469c1de7cfaccba5a92f0/PUBLISHING.md#L276
This is where we first reference TUF metadata if users are reading from top to bottom.

In fact, maybe the whole "Expiration Policy" section could be moved upwards. Is there any reason why it's explained after the "Waves" section?

Comment (tjkirch) [reply PUBLISHING.md:120]: > Is there any reason why it's explained after the "Waves" section?

It's more straightforward, and I find it useful to explain the "interesting" concepts first. :)

We'll look at some options for reorganizing this bit.

**Comment by tjkirch** [comment]
^ This push clarifies some issues raised by @etungsten.  Thanks!

**Review by etungsten — APPROVED** [review]
:bee:

File: PUBLISHING.md [hunk 3]
```diff
@@ -0,0 +1,393 @@
...
+
+The publishing process uses a configuration file called `Infra.toml`.
+The relevant sections of this file will be introduced as needed below.
+You can also see an [example file](tools/pubsys/Infra.toml.example) where each section is commented.
+
+When you make your own `Infra.toml`, you put it in the root of the Bottlerocket code repo, wherever you have it checked out.
+(If you want to keep it elsewhere, you can pass `-e "PUBLISH_INFRA_CONFIG_PATH=/my/path"` to subsequent `cargo make` commands.)
+
+Note: several commands work with AWS services, so there's some shared configuration related to AWS accounts and AWS IAM roles.
+For example, you can say to assume a role before any API calls are made, and to assume a role for any API calls in a specific region.
```
webern: "you can say to assume" felt a little awkward on first reading. You can take-or-leave the suggestion.

```suggestion
For example, you can specify a role to assume before any API calls are made, and a role to assume for any API calls in a specific region.
```

PUBLISHING.md
webern: Seems fine. Using "root.json file" instead of "root role file" might make it more clear that we are no longer talking about AWS roles. I could see it being confusing due to the overload of "role".

webern (approved):
👟

zmrow:
:arrow_up: Addresses @webern 's wording concern.

PUBLISHING.md
```diff
@@ -0,0 +1,393 @@
...
+The variant and architecture are automatically added onto the metadata URL, matching the format of the directories inside `build/repos/default/latest`.
+(The targets directories is shared for all variants and architectures, since target files are prefixed with a checksum.)
+
+### Using your repo from a Bottlerocket host
+
+By default, Bottlerocket hosts talk to the project's official repos.
+There are two ways to point your hosts at your own repo - at build time or at run time.
+
+If you're maintaining your own fork of Bottlerocket, you'd probably want to change the settings at build time, so you don't have to change settings for every host you launch.
+If you're just running a few hosts, or don't want to maintain a fork, then it's easier to change settings at run time.
```
samuelkarp, line 232: If you're running your own repo, won't you need to make sure that the hosts have your keys in the root.json on the filesystem instead of the project's official keys?

PUBLISHING.md
```diff
@@ -0,0 +1,393 @@
...
+
+```shell
+cargo make repo -e PUBLISH_WAVE_POLICY_PATH=sources/updater/waves/accelerated-waves.toml
+```
+
+To learn more about waves, check out the [README](sources/updater/waves).
+
+### Expiration policy
+
+Each piece of signed metadata in a TUF repo expires after a specific length of time, meaning that repos need to resigned regularly.
```
samuelkarp:
```suggestion
Each piece of signed metadata in a TUF repo expires after a specific length of time, meaning that repos need to re-signed regularly.
```

("resigned" is a word that means something different from what we want here)

PUBLISHING.md
```diff
@@ -0,0 +1,393 @@
...
+```
+/your/prefix/here/aws-k8s-1.17/x86_64/1.0.1-dafe3b16/image_id
+```
+
+...to a simpler parameter name:
+```
+/your/prefix/here/aws-k8s-1.17/x86_64/latest/image_id
+```
+
+You can then use (and share) this parameter name to get the latest AMI ID.
```
samuelkarp: Outside of the `/aws` prefix, I don't believe parameters can be shared outside the account that owns the parameter.

PUBLISHING.md
tjkirch: The only way you get the official Bottlerocket root.json in your image now is if you specifically list its URL in your Infra.toml file - by default, it'll generate a role for you and include that, or use whatever you've pointed to with a URL.  This section is trying to describe how the settings in the model will still be pointing at the Bottlerocket repos, which will have used a different root role, and the options you have for changing that.  Does that make sense, am I misunderstanding?

samuelkarp: That makes a lot of sense, thanks for the explanation!

In the future, should we change the default settings so that it doesn't include the project's official repository unless you've indicated that in your Infra.toml?

PUBLISHING.md
tjkirch: There's no hard rule, but the general guideline ([1](http://www.gatehousenewsroom.com/2017/08/17/ap-style-guidelines-for-hyphen-usage/) [2](https://blog.apastyle.org/apastyle/2016/10/hyphenation-station-the-hyphenation-of-prefixes-in-apa-style.html)) is to use a hyphen if the prefixed form (`re signed`) could be misunderstood as another word (past tense of `resign`).  I don't think anyone expects that a repo will resign from its job.  Particularly when we're in a paragraph about signing.  And you favor the unhyphenated where possible, per [AP](http://www.gatehousenewsroom.com/2017/08/17/ap-style-guidelines-for-hyphen-usage/).  :P

PUBLISHING.md
tjkirch: Yeah, but it's still useful to share within the account, e.g. to separate launch mechanisms from output of the deployment process.

PUBLISHING.md
tjkirch: Yeah, that makes sense.  We'll need fancier ways of plumbing things together to make it work, and it may be easiest after we have out-of-tree variants going.

PUBLISHING.md
tjkirch: Chicago style could go either way. :P

tjkirch: (happy to change it, just having fun)

etungsten: English is hard

zmrow:
:arrow_up_small: Added the hyphen to `re-sign`. :)

samuelkarp (approved):

zmrow:
:arrow_up: Addressed @samuelkarp 's comment re: SSM parameters.

samuelkarp (approved):

Issue #1139, opened 2020-10-02
[ECS] Can't connect to instance via Aws Session Manager


namiops:
I'm follow QUICKSTART-ECS with the role setting for lauching EC2 instances, but I can't connect to any Bottlerocket instance with SSM (Session Manager).

But It work, until I attached additional Policy "AmazonEC2ContainerRegistryReadOnly" to the running EC2's role.

**Platform I'm building on:**
AWS ECS

**What I expected to happen:**
Connect to EC2 Instances with Session Manager

**What actually happened:**
Can't connect without the policy AmazonEC2ContainerRegistryReadOnly

**How to reproduce the problem:**
Attach addtional policy AmazonEC2ContainerRegistryReadOnly to EC2's Role

tjkirch:
Thanks for the input!  We do have a note in the [README](https://github.com/bottlerocket-os/bottlerocket/blob/17c1e29ec3fa5cade1ab631e899141267118fecc/README.md#exploration), but it probably also belongs in the QUICKSTART guides.

PR #1144, opened 2020-10-04
ECS README: Add ECR permission note

ghost:
**Issue number:**
#1139

**Description of changes:**
Additional note to QUICKSTART-ECS.md about ECR permision.

### DIFF
--- a/QUICKSTART-ECS.md
+++ b/QUICKSTART-ECS.md
@@ -144,6 +144,8 @@ aws iam attach-role-policy \
    --policy-arn arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore
 ```
 
+**Note:** The instance also need permission to access ECR repository where [default control container](https://github.com/bottlerocket-os/bottlerocket-control-container) live; the appropriate policy to add to INSTANCE_ROLE_NAME role is `AmazonEC2ContainerRegistryReadOnly`
+
 Next, to retrieve the instance profile name used to launch instances, run this:
 
 ```


QUICKSTART-ECS.md
```diff
@@ -144,6 +144,8 @@ aws iam attach-role-policy \
    --policy-arn arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore
 ```
 
+**Note:** The instance also need permission to access ECR repository where [default control container](https://github.com/bottlerocket-os/bottlerocket-control-container) live; the appropriate policy to add to INSTANCE_ROLE_NAME role is `AmazonEC2ContainerRegistryReadOnly`
```
zmrow, line 147: Since this policy is required to fetch the container to make this feature work, let's add it to the `aws iam...` command directly above.

zmrow (commented):
Thanks for the contribution!

QUICKSTART-ECS.md
tjkirch: That sounds good to me.  We can just add another `--policy-arn` line, and then we shouldn't need any extra wording.

tjkirch: Actually, @samuelkarp pointed out that the ECS policy referenced in the previous step, which is a requirement before getting here, includes the read-only permissions we need for ECR.

@saigoneer, can you please confirm that you ran the previous step, and created the initial policy as per [the ECS instructions](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/instance_IAM_role.html)?

namiops: I confirm it worked, the ecsInstanceRole actually had the policy AmazonEC2ContainerServiceforEC2Role, which covers AmazonEC2ContainerRegistryReadOnly.

Anyway, we still need to instruct somewhere about the two necessary policies for INSTANCE_ROLE_NAME: AmazonEC2ContainerServiceforEC2Role and AmazonSSMManagedInstanceCore.

@tjkirch adding another line ```--policy-arn``` with AmazonEC2ContainerServiceforEC2Role should be fine?

tjkirch: > Anyway, we still need to instruct somewhere about the two necessary policies for INSTANCE_ROLE_NAME: AmazonEC2ContainerServiceforEC2Role and AmazonSSMManagedInstanceCore.

@saigoneer I'm sorry, I'm not sure I follow.  The command just above this adds AmazonSSMManagedInstanceCore, and the step before that describes following the ECS docs to set up AmazonEC2ContainerServiceforEC2Role.  I don't think we want to repeat the ECS instructions in here.  Can you please describe what's missing?

namiops: @tjkirch Actually, the policy AmazonEC2ContainerServiceforEC2Role which need for INSTANCE_ROLE_NAME wasn't mentioned in QUICKSTART. (Until I opened the role ```ecsInstanceRole``` in AWS)

tjkirch: Yeah, it's there from following the ECS guide that we link to in the "IAM role" section above.  I don't think we don't need to call out AmazonEC2ContainerServiceforEC2Role by name here, we just need to make sure the user is set up.  We could copy some of the instructions out of the ECS guide and into this quickstart, but I'm hesitant because I think the ECS team is a better owner for that information.  Is that your concern in general, users not having to look at other resources?  I think that's going to be unavoidable, given that we have to have some other tools set up as well.  We need to strike a balance between a concise guide and a self-contained guide, and I think linking to ECS docs is a good thing here.

namiops: > We could copy some of the instructions out of the ECS guide and into this quickstart, but I'm hesitant because I think the ECS team is a better owner for that information. Is that your concern in general, users not having to look at other resources? I think that's going to be unavoidable, given that we have to have some other tools set up as well.

@tjkirch I see, that makes sense. Thank you for explaining, I'm closing the PR.

PR #1155, opened 2020-10-09
align opt-level for crate build deps

### LINKED ISSUE #1153: Rust 1.47.0 is causing CI to fail
Related to https://github.com/bottlerocket-os/bottlerocket-sdk/issues/36

**Platform I'm building on:**
GitHub Actions

Bottlerocket to build successfully in CI

**What actually happened:**
In CI, the build fails with errors similar to:
```
2020-10-09T18:42:05.0204440Z   Error response from daemon: Conflict. The container name "/buildsys-pkg-glibc-x86_64-2fe95a9c2d67" is already in use by container "e578bf40fc498b9cbe4b13b1a866d023c702cce35d44e27103293c5de852f2fc". You have to remove (or rename) that container to be able to reuse that name.
```

Use Rust 1.47.0 and attempt to build Bottlerocket via `cargo make`

**I'm able to build Bottlerocket successfully with Rust 1.46.0**

bcressey:
**Issue number:**
Fixes #1153

**Description of changes:**
Our OS packages are built as a byproduct of build script execution, while the final Rust artifacts are largely ignored.

Since Rust 1.47.0, cargo builds host dependencies like build scripts with opt-level "0" by default, which did not match the "z" level we picked to minimize the size of the output artifacts.

This caused package build dependencies to be built more than once, with concurrent Docker builds for the same package. This is wasteful and creates races that our build tool does not handle gracefully.

By aligning the opt-level we restore the previous behavior.

**Testing done:**
Built locally using Rust 1.47, with and without the change applied.

Without the change, packages like glibc were built more than once. With the change, they are only built once.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,6 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v2
-      - run: rustup toolchain install 1.46.0 && rustup default 1.46.0
       - run: cargo install --version 0.30.0 cargo-make
       - run: cargo install --version 0.6.6 cargo-deny --no-default-features
       - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} unit-tests
--- a/packages/Cargo.toml
+++ b/packages/Cargo.toml
@@ -76,3 +76,6 @@ members = [
 [profile.dev]
 debug = false
 opt-level = 'z'
+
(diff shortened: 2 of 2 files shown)


webern (approved):
it works 🍪

zmrow (approved):
Hey look at that! We can build again!

:medal_military:

Issue #1159, opened 2020-10-16
Add support for BPFFS mount.


izaakschroeder:
**What I'd like:**

To be able to install Cilium on top of bottlerocket-based EKS clusters. Since the rootfs is immutable, CNI plugins that rely on bootstrapping to mount `bpffs` won't work (i.e. Cilium).

**Any alternatives you've considered:**

There aren't any really… the underlying FS can't be modified so there isn't really any way around this 😢

samuelkarp:
Per https://github.com/bottlerocket-os/bottlerocket/pull/1160#issuecomment-725692670, the bpffs mount is already present at `/sys/fs/bpf` in our 1.0.2 release.  I'm going to go ahead and close this, but please feel free to re-open or open a new issue if you run into further issues.  Thanks for using Bottlerocket!

Issue #1063, opened 2020-08-22
Enabling bpf-lsm security module and both CAP_BPF and CAP_PERFMON security capabilities


danmx:
**What I'd like:**

I'd like to see `bpf-lsm` (BPF-based Linux Security Module, introduced in kernel 5.7) and both `CAP_BPF` and CAP_PERFMON security capabilities (introduced in kernel 5.8) enabled.

**Any alternatives you've considered:**

webern:
Good ideas, thank you for bringing these to our attention. It may be some time before we move to these kernel versions, but we'll take a closer look at this.

danmx:
We could port those patches to the current kernel. I don’t have much experience with it but with some guidance I could try contributing.

WilboMo:
Hello danmx. Since our last correspondence on this issue, the features you’re looking for have been added in newer [variants](https://github.com/bottlerocket-os/bottlerocket/tree/develop/variants#bottlerocket-variants) of Bottlerocket like aws-k8s-1.20 and aws-k8s-1.21 through use of the 5.10 kernel. You can find AMIs for these variants through the web console [here](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami-bottlerocket.html) or via command line as shown [here](https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-EKS.md#finding-an-ami). Instructions on launching Bottlerocket nodes can be found [here](https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-EKS.md#using-a-bottlerocket-ami-with-amazon-eks). Please let us know if you have any other issues or concerns.

kdaula:
Resolving, since bpf-lsm and the new capabilities are available in newer variants.

PR #1160, opened 2020-10-16
feat: Add `bpffs` mount to default installation.

izaakschroeder:
Since the rootfs is immutable, CNI plugins that rely on bootstrapping to mount `bpffs` won't work (i.e. Cilium). There is little overhead in adding this functionality. I'm uncertain what security concerns would arise from having this mounted, as only privileged process should be able to access it. There are other hardening options available which one might consider, e.g. https://wiki.archlinux.org/index.php/Security#BPF_hardening (perhaps also related to https://github.com/bottlerocket-os/bottlerocket/issues/1063?).

**Issue number:**

See: https://github.com/bottlerocket-os/bottlerocket/issues/1159

**Description of changes:**

Add a `bpffs` to `release` package that adds a `bpffs.mount` entry for `systemd` using reference from the Cilium codebase here: https://github.com/cilium/cilium/blob/e9cb43c03179af6f7cd5d0abb9974daeb5a21e5a/install/kubernetes/cilium/templates/cilium-nodeinit-daemonset.yaml#L111.

**Testing done:**

Built a new AMI in `us-west-2` (`ami-060ac8bae2db28b67`) with the changes in this PR and ran an EKS cluster using this AMI. Installed `cilium@1.8.3` using `terraform` + `helm` with the following settings:

```
  set {
    name  = "global.kubeProxyReplacement"
    value = "strict"
  }
  set {
    name  = "global.k8sServiceHost"
    value = var.k8s_host
  }
  set {
    name  = "global.k8sServicePort"
    value = var.k8s_port
  }
  set {
    name  = "global.egressMasqueradeInterfaces"
    value = "eth0"
  }
  set {
    name  = "global.proxy.sidecarImageRegex"
    value = "cilium/proxyv2"
  }
  set {
    name  = "global.prometheus.enabled"
    value = "true"
  }
```

Which is pretty close to the recommendations they have for installation on EKS here: https://docs.cilium.io/en/v1.8/gettingstarted/k8s-install-eks/. The cilium pods seem to run fine:

```
NAME                                  READY   STATUS             RESTARTS   AGE
pod/cilium-j5qz5                      1/1     Running            0          27m
pod/cilium-kbw9h                      1/1     Running            0          27m
pod/cilium-operator-c9f6c4577-cv6s8   1/1     Running            0          12m
pod/cilium-operator-c9f6c4577-t6vts   1/1     Running            0          21m
```

The logs for `cilium` indicate:

```
level=info msg="linking environment: OK!" subsys=linux-datapath
level=info msg="Detected mounted BPF filesystem at /sys/fs/bpf" subsys=bpf
```

Which leads me to believe these changes are at least necessary if not sufficient to allow Cilium to work.

**Other notes:**

Not related to this PR, but the startup logs also indicate:

```
level=warning msg="Session affinity for host reachable services needs kernel 5.7.0 or newer to work properly when accessed from inside cluster: the same service endpoint will be selected from all network namespaces on the host." subsys=daemon
```

The current kernel according to logs:

```
level=info msg="clang (10.0.0) and kernel (5.4.50) versions: OK!" subsys=linux-datapath
```

Does `bottlerocket` have a recommended or required kernel? I haven't tried changing it through my launch template yet, but was just wondering if this is supported and if there are any gotchas I need to be aware of?

**Terms of contribution:**

### DIFF
--- a/packages/release/bpffs.mount
+++ b/packages/release/bpffs.mount
@@ -1,0 +1,14 @@
+[Unit]
+Description=Mount BPF filesystem
+DefaultDependencies=no
+Before=local-fs.target umount.target
+After=swap.target
+
+[Mount]
+What=bpffs
+Where=/sys/fs/bpf
+Type=bpf
+Options=rw,nosuid,nodev,noexec,relatime,mode=700
+
+[Install]
+WantedBy=multi-user.target
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -23,6 +23,7 @@ Source1009: usr-src-kernels.mount.in
 Source1010: var-lib-bottlerocket.mount
 Source1011: usr-share-licenses.mount.in
 Source1012: etc-cni.mount
+Source1013: bpffs.mount
 
 BuildArch: noarch
 Requires: %{_cross_os}acpid
@@ -97,7 +98,7 @@ EOF
 
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 \
-  %{S:1002} %{S:1006} %{S:1007} %{S:1008} %{S:1010} %{S:1012} \
+  %{S:1002} %{S:1006} %{S:1007} %{S:1008} %{S:1010} %{S:1012} %{S:1013} \
   %{buildroot}%{_cross_unitdir}
 # Mounting on usr/src/kernels requires using the real path: %{_cross_usrsrc}/kernels
 KERNELPATH=$(systemd-escape --path %{_cross_usrsrc}/kernels)
@@ -127,6 +128,7 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_templatedir}/motd
 %{_cross_unitdir}/*-kernels.mount
 %{_cross_unitdir}/*-licenses.mount
 %{_cross_unitdir}/var-lib-bottlerocket.mount
+%{_cross_unitdir}/bpffs.mount
 %dir %{_cross_templatedir}
 %{_cross_templatedir}/motd
 


packages/bpffs/bpffs.mount
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Mount BPF filesystem
+DefaultDependencies=no
+Before=local-fs.target umount.target
+After=swap.target
```
bcressey: I appreciate the attention to detail in setting this up as a separate package. That would be valuable if for some reason it's not safe for all variants of Bottlerocket.

I need to spend some time looking at the implications of bpffs and whether additional controls (e.g. via SELinux policy) are needed, but my initial take is that we'll want this file to go into `release` directly.

izaakschroeder: Ok great! Let me know what you find out and what you'd like from me and I'll make it happen! Thanks so much for looking into this 😄

bcressey: Let's add this file as `packages/release/bpffs.mount` and update `release.spec` with:
```
Source1013: bpffs.mount
```

Then it can be installed along with other units like `Source1012` / `%{S:1012}`.

bcressey:
> Does bottlerocket have a recommended or required kernel? I haven't tried changing it through my launch template yet, but was just wondering if this is supported and if there are any gotchas I need to be aware of?

We're currently on 5.4 which is the most recent LTS kernel supported by the Amazon Linux team. We expect to move to the newer LTS kernel, likely 5.10, assuming one is picked in December. That could happen in late January or early February.

For Kubernetes variants our thinking is that we'd add the newer kernel to the most recent variant, see how that goes, and introduce it to older variants once we're confident there are no major regressions for popular workloads.

Depending on how the releases line up that might land in our k8s-1.19 or k8s-1.20 variants.

Custom builds should be able to use kernel builds newer than 5.4. There may be minor compatibility issues though. For example, in places where new SELinux checks have been added, the policy will deny those operations by default.

packages/bpffs/bpffs.mount
izaakschroeder: On it now! 😄

izaakschroeder: All set @bcressey! I haven't done any further testing after this change – do you have a process you follow? There are quite a few image variants and it would probably take considerable time for me to test them all… I can re-publish an AMI and re-run my EKS smoke test with Cilium if you'd like though! Let me know what else you need from me! 🎉

bcressey: We run more extensive tests on all variants prior to release, so you don't need to worry about testing all of them.

It'd be good to confirm at least that `/sys/fs/bpf` is mounted as you expect on the host.

You could also run through the [Cilium connectivity test](https://docs.cilium.io/en/v1.8/gettingstarted/k8s-install-eks/#deploy-the-connectivity-test) and confirm that it passes. Checking `dmesg` afterwards for errors, especially SELinux denials, also wouldn't hurt.

(Any problems from the more thorough testing wouldn't be blockers here, as long as we had issues to track them.)

izaakschroeder: Ok sure I'll do that! It might take a little bit of time here yet but we'll make it happen!

Also what is your release process like? After we (hopefully 🤞 ) merge this should I expect about a week? month? until this functionality is usable in the released bottlerocket AMIs that are listed by `"/aws/service/bottlerocket/aws-k8s-${var.eks_cluster.version}/x86_64/latest/image_id"`?

bcressey: We don't have a fixed release cadence but roughly 4-6 weeks has been typical.

bcressey:
I went to test this today, and discovered that `/sys/fs/bpf` is already mounted. It was added to the set of automatic mounts in systemd 238.

So it seems like we don't need this change, unless you've found otherwise in your testing.

bcressey:
I'll go ahead and close this, since the expected functionality is present.

Just want to also say that I really appreciate the effort you put in to learn the build system and tools, to tell us what didn't work, and to propose fixes.

izaakschroeder:
Ah sorry! Thanks for verifying this, and no problem! 😄 Work has kind of sidelined this stuff for a while 😢 but I'm glad it's working. Your team is super responsive and friendly and I'm excited to see this project continue! Thanks for being patient 🎉

PR #1161, opened 2020-10-16
chore: Refactor `amiize` script.

izaakschroeder:
When developing https://github.com/bottlerocket-os/bottlerocket/pull/1160 running the `amiize` script failed on my machine repeatedly (macOS 10.15.7) using `aws-vault` in the `us-west-2` region. The script itself failed in then tried to continue due to lack of `set -e` and so a set of more strict shell options has been added as a prologue. Because of this, `check_return` isn't needed per se since the script just fails if the command fails. The file size calculator also fails to work on macOS because the `du` command is missing some (presumably GNU) flags; so instead we use `awk`, `ls` and `bc` to pull out the size in GiB. When trying to make the AMI `coldsnap` also always uploaded to the default region instead of the region specified by `--region`. This has been fixed.

These refactorings are not at all required and were just useful for my development and testing and I am in no way suggesting or asking you incorporate them, but if there's something useful you find in here feel free to take it (in particular the `--region` flag to `coldsnap`)!

**Issue number:**

 * https://github.com/bottlerocket-os/bottlerocket/issues/1162 – Bug with `--region`.

**Description of changes:**

 * Use `set -euo pipefail`
 * Use `awk`/`ls`/`bc` instead of `du`
 * Remove `check_return`
 * Add `--region` to `coldsnap`

**Testing done:**

I ran this locally and used the AMI built as part of https://github.com/bottlerocket-os/bottlerocket/pull/1160.

**Terms of contribution:**

### DIFF
--- a/bin/amiize.sh
+++ b/bin/amiize.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 # Register partitioned root and data images as an AMI in EC2.
 # Only registers with HVM virtualization type and GP2 EBS volume type.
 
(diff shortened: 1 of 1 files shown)


izaakschroeder:
Closing per conversation in https://github.com/bottlerocket-os/bottlerocket/issues/1162 🎉

Issue #1162, opened 2020-10-16
`cargo make ami` fails on macOS around `BUILDSYS_ARCH`.


izaakschroeder:
`cargo make ami` doesn't work properly on macOS.

**Platform I'm building on:**

macOS

**What I expected to happen:**

`cargo make ami` to build without errors.

**What actually happened:**

```
/var/folders/_8/37v4m93x6xg162jyl200y5nw0000gn/T/izaakschroeder/fsio/v0FfHrk3Iz.sh: line 26: ${BUILDSYS_ARCH,,}: bad substitution
```

**How to reproduce the problem:**

Run `cargo make ami` on macOS.

tjkirch:
Hi @izaakschroeder - I'm afraid you caught us in an awkward time, since we just replaced the amiize.sh script and were about to remove it from the repo.  Would you be up for trying the replacement?  It's documented in our [PUBLISHING](https://github.com/bottlerocket-os/bottlerocket/blob/develop/PUBLISHING.md) guide.

izaakschroeder:
Oh 😢 no worries! Sorry about that, I didn't realize. I can give it a whirl and see if it suffers from the same problem though!

izaakschroeder:
This is where I'm at right now:

```sh
[cargo-make] INFO - Running Task: link-variant
[cargo-make] INFO - Build Done in 2577 seconds.
❯ cargo make ami -e PUBLISH_REGIONS="us-west-2" -e PUBLISH_AMI_NAME="bottlerocket-aws-k8s-1.17-x86_64-bpffs-test"
[cargo-make] INFO - cargo make 0.32.5
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: ami
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Running Task: setup
[cargo-make] INFO - Running Task: fetch-sources
[cargo-make] INFO - Running Task: publish-tools
[cargo-make] INFO - Running Task: ami
/Users/izaakschroede : decoded 2147483648 bytes
/Users/izaakschroede : decoded 1073741824 bytes
/var/folders/_8/37v4m93x6xg162jyl200y5nw0000gn/T/izaakschroeder/fsio/v0FfHrk3Iz.sh: line 26: ${BUILDSYS_ARCH,,}: bad substitution
```

I'll continue to investigate!

izaakschroeder:
Alright so patching

```sh
[...]

PR #1167, opened 2020-10-16
Makefile: Check that $BUILDSYS_ARCH is a supported architecture

zmrow:
**Issue number:**
Related to #1162 

**Description of changes:**
```
$BUILDSYS_ARCH is used all over the place in the Makefile, and if we don't
check that it's a valid supported architecture first, spurious directories get
created and commands further down the line will fail.  This change checks
$BUILDSYS_ARCH first before anything happens and fails.

    
It doesn't use any parameter expansion like the previous check did, meaning it
will be compatible with all versions of bash.
```

**Testing done:**
* `cargo make` commands fail if an invalid arch is supplied via `-e BUILDSYS_ARCH=`
```
...
[cargo-make] INFO - Running Task: setup
Unrecognized architecture 'bar'; please use 'x86_64 or 'aarch64'
...
```

* `cargo make` commands with valid architectures work as expected.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -98,6 +98,15 @@ PUBLISH_AMI_NAME_DEFAULT = "${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH
 [tasks.setup]
 script = [
 '''
+# Ensure we use a supported architecture
+case "${BUILDSYS_ARCH}" in
+   x86_64|aarch64) ;;
+   *)
+      echo "Unrecognized architecture '${BUILDSYS_ARCH}'; please use 'x86_64 or 'aarch64'"
+      exit 1
+      ;;
+esac
+
 mkdir -p ${BUILDSYS_BUILD_DIR}
 mkdir -p ${BUILDSYS_OUTPUT_DIR}
 mkdir -p ${BUILDSYS_PACKAGES_DIR}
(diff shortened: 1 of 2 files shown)


zmrow:
^ Fixed a typo

Makefile.toml
```diff
@@ -98,6 +98,15 @@ PUBLISH_AMI_NAME_DEFAULT = "${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH
 [tasks.setup]
 script = [
 '''
+# Ensure we use a supported architecture
+case "${BUILDSYS_ARCH}" in
+   x86_64|aarch64)
+   ;;
+   *)
+      echo "Unrecognized architecture '${BUILDSYS_ARCH}'; please use 'x86_64 or 'aarch64'"
+      exit 1 ;;
```
```suggestion
   x86_64|aarch64) ;;
   *)
      echo "Unrecognized architecture '${BUILDSYS_ARCH}'; please use 'x86_64 or 'aarch64'"
      exit 1
      ;;
```
nit: a more standard style.

tjkirch (approved):

zmrow:
^ style points

izaakschroeder:
Exciting, thank-you! ❤️

tjkirch (dismissed):

Makefile.toml
```diff
@@ -437,16 +446,6 @@ fi
...
 
-# Canonicalize architecture identifier to the value recognized by EC2.
-case "${BUILDSYS_ARCH,,}" in
-   arm64|aarch64)
-      arch=arm64 ;;
-   x86_64|amd64)
-      arch=x86_64 ;;
-   *)
-      arch="${BUILDSYS_ARCH}" ;;
-esac
```
bcressey, line 448: If we don't already, we need to handle at least the `aarch64` -> `arm64` canonicalization in pubsys, since it's going away here.

zmrow: This is fixed by using the `parse_arch` function the SSM code already uses to automatically canonicalize the arch to the correct value.

zmrow: Fixed using the `parse_arch` function that automatically canonicalizes the architecture to the proper value.

zmrow:
Fixes @bcressey 's concern - when we wrote the SSM publishing code we added a function to canonicalize architecture but hadn't gone back to add it to the AMI code.

bcressey (approved):

webern (approved):

tjkirch (approved):

Issue #1088, opened 2020-08-28
Read/Write access to NVMe volumes


chonton:
**What I'd like:**
Many AWS instance types offer "[Instance store volumes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/InstanceStorage.html#instance-store-volumes)".  When running stateful services inside kubernetes, local storage would increase performance (and decrease cost). Unfortunately, most kubernetes cluster management terminate node instances for upgrades, causing the local storage to be lost.   The bottlerocket upgrade scheme is appealing because it provides half of the solution.

How do I make the local volumes available to pods?  The boot volume is read-only, any pods only have access to boot volume.

**Any alternatives you've considered:**
- [storage-local-static-provisioner](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/getting-started.md#option-5-eks-experimental)
- [eks ssd provisioner](https://github.com/brunsgaard/eks-nvme-ssd-provisioner) - fails with ```failed to mkdir "/pv-disks": mkdir /pv-disks: read-only file system```

bcressey:
It's not a complete answer, but modifying the EKS SSD provisioner to use `/var/lib/pv-disks` instead of `/pv-disks` would work around the read-only filesystem error.

bcressey:
We landed the OS side of this feature in #1173; NVMe volumes will show up under `/dev/disk/ephemeral` and can be used as persistent volumes.

I've used Helm to install the local volume provisioner using these steps:
https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/tree/master/helm#install-with-helm-install-directly
```
$ helm install \
  -f ./bottlerocket.yaml local-volume-provisioner \
  -n kube-system ./helm/provisioner
```

Here's my `bottlerocket.yaml`:
```
 common:
  useNodeNameOnly: true
classes:
- name: local-storage
  hostDir: "/dev/disk/ephemeral"
  mountDir: "/dev/disk/ephemeral"
  storageClass: true
```

Please try it out and let us know if you run into problems.

ei-grad:
Is it possible to use ephemeral disks for both - local-volume-provisioner and [{/var/lib/containerd,/var/lib/kubelet,/var/log/pods}](https://bottlerocket.dev/en/os/1.22.x/concepts/bootstrap-commands/#configure-ephemeral-disks) directories?
```
[...]

PR #1173, opened 2020-10-19
create links for ephemeral storage devices

bcressey:
**Issue number:**
#1088

**Description of changes:**
Populate `/dev/disk/ephemeral` with links to ephemeral storage devices.

Update security guidance to reflect the risks of mapping a block device into an untrusted container.

**Testing done:**

Verified that the two ephemeral disks were linked on an `m5ad.4xlarge`:
```
# ls -latr /dev/disk/ephemeral/

lrwxrwxrwx. 1 root root  13 Oct 18 04:11 nvme-Amazon_EC2_NVMe_Instance_Storage_AWS27A87D8ECFBD312B5 -> ../../nvme3n1
lrwxrwxrwx. 1 root root  13 Oct 18 04:11 nvme-Amazon_EC2_NVMe_Instance_Storage_AWS220C24D1769700C7B -> ../../nvme2n1
```

Verified that other disks were classified as "system":
```
# ghostdog scan /dev/nvme0n1
BOTTLEROCKET_DEVICE_TYPE=system

# ghostdog scan /dev/nvme1n1
BOTTLEROCKET_DEVICE_TYPE=system

# ghostdog scan /dev/nvme2n1
BOTTLEROCKET_DEVICE_TYPE=ephemeral

# ghostdog scan /dev/nvme3n1
BOTTLEROCKET_DEVICE_TYPE=ephemeral
```

Verified that the external static provisioner could discover the ephemeral disks:
```
$ kubectl describe pv local-pv-eddb9f1f
...
Source:
    Type:  LocalVolume (a persistent volume backed by local storage on a node)
    Path:  /dev/disk/ephemeral/nvme-Amazon_EC2_NVMe_Instance_Storage_AWS220C24D1769700C7B
Events:    <none>
```

**Terms of contribution:**

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -8,6 +8,7 @@
 * [**early-boot-config**](sources/api/early-boot-config): A program run at boot to read platform-specific data, such as EC2 user data, and send requested configuration to the API.
 * **gptprio:** A structure of bits in GPT partition headers that specifies priority, tries remaining, and whether the partition booted successfully before.
   signpost sets these and GRUB uses them to determine which partition set to boot.
+* [**ghostdog**](sources/ghostdog): A program used to manage ephemeral disks.
 * [**growpart**](sources/growpart): A program used to expand disk partitions upon boot.
 * **host containers**: Containers that run in a separate instance of containerd than "user" containers spawned by an orchestrator (e.g. Kubernetes).
   Used for system maintenance and connectivity.
--- a/packages/os/ephemeral-storage.rules
+++ b/packages/os/ephemeral-storage.rules
@@ -1,0 +1,27 @@
+# ephemeral storage links: /dev/disk/ephemeral
+
+ACTION=="remove", GOTO="ephemeral_storage_end"
+SUBSYSTEM!="block", GOTO="ephemeral_storage_end"
+ENV{DEVTYPE}!="disk", GOTO="ephemeral_storage_end"
+
+# Known sources of ephemeral disks:
+# - EC2 instance types with instance storage volumes
+# - KVM VMs with additional virtio disks
+KERNEL!="nvme*|vd*", GOTO="ephemeral_storage_end"
+
+# We can't be sure that devices with similar kernel names are all ephemeral.
+# Our udev helper checks for known system partitions to classify the device.
+IMPORT{program}=="/usr/bin/ghostdog scan $devnode"
+ENV{BOTTLEROCKET_DEVICE_TYPE}!="ephemeral", GOTO="ephemeral_storage_end"
+
+# EBS volumes will show up as "Amazon Elastic Block Store", but we expect these
+# to be either system volumes or managed by the EBS CSI driver.
+KERNEL=="nvme*", ATTRS{model}=="Amazon EC2 NVMe Instance Storage*", ENV{ID_SERIAL}=="?*", \
+  SYMLINK+="disk/ephemeral/nvme-$env{ID_SERIAL}"
+
+# Fow now, virtio support is meant to enable local development, so only create
+# links if the serial matches a specific prefix.
+KERNEL=="vd*", ENV{ID_SERIAL}=="eph*", \
+  SYMLINK+="disk/ephemeral/virtio-$env{ID_SERIAL}"
+
+LABEL="ephemeral_storage_end"
--- a/sources/ghostdog/src/main.rs
+++ b/sources/ghostdog/src/main.rs
@@ -1,0 +1,181 @@
+/*!
+ghostdog is a tool to manage ephemeral disks.
+It can be called as a udev helper program to identify ephemeral disks.
+*/
+
+use argh::FromArgs;
+use gptman::GPT;
+use hex_literal::hex;
+use lazy_static::lazy_static;
+use signpost::uuid_to_guid;
+use snafu::ResultExt;
+use std::collections::HashSet;
+use std::fs;
+use std::io::{Read, Seek};
+use std::path::PathBuf;
+
+#[derive(FromArgs, PartialEq, Debug)]
+/// Manage ephemeral disks.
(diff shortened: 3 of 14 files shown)


packages/os/ephemeral-storage.rules
```diff
@@ -0,0 +1,27 @@
+# ephemeral storage links: /dev/disk/ephemeral
+
+ACTION=="remove", GOTO="ephemeral_storage_end"
```
tjkirch, line 3: Do we want to continue for `change` actions?  `ACTION!="add"` may be more accurate?

sources/ghostdog/src/main.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+    Ok(())
+}
+
+/// Find the device type by examining the partition table, if present.
+fn find_device_type<R>(reader: &mut R) -> Result<String>
+where
+    R: Read + Seek,
+{
+    // Assume it's an ephemeral unless we find evidence to the contrary.
```
tjkirch: May be worth saying that we lean this way because we expect to be called by udev rules where there's already some level of evidence that the device is ephemeral?

tjkirch (approved):

packages/os/ephemeral-storage.rules
bcressey: We need `change` since that fires when the partition table is rewritten, which can influence whether a device is counted as an ephemeral.

There are a lot of other actions that we don't care about. `60-persistent-storage.rules` also uses this `ACTION=="remove"` guard, so I think it's still the right choice.

GLOSSARY.md
```diff
@@ -8,6 +8,7 @@
 * [**early-boot-config**](sources/api/early-boot-config): A program run at boot to read platform-specific data, such as EC2 user data, and send requested configuration to the API.
 * **gptprio:** A structure of bits in GPT partition headers that specifies priority, tries remaining, and whether the partition booted successfully before.
   signpost sets these and GRUB uses them to determine which partition set to boot.
+* [**ghostdog**](sources/ghostdog): A program used to manage ephemeral disks.
```
webern, line 11: yay

webern (approved):
💾

tjkirch (approved):

sources/ghostdog/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+    Ok(device_type.to_string())
+}
+
+/// Print the device type in the environment key format udev expects.
+fn emit_device_type(device_type: &str) -> () {
+    println!("BOTTLEROCKET_DEVICE_TYPE={}", device_type);
+}
+
+// Known system partition types for Bottlerocket.
+lazy_static! {
```
zmrow, line 84: super nit: I might put this constant at the top.

zmrow (approved):
Nice!  :tractor:

sources/ghostdog/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/// Find the device type by examining the partition table, if present.
+fn find_device_type<R>(reader: &mut R) -> Result<String>
+where
+    R: Read + Seek,
+{
+    // We expect the udev rules to only match block disk devices, so it's fair
+    // to assume it could have a partition table, and that it's probably an
+    // unformatted ephemeral disk if it doesn't.
```
samuelkarp, line 59: Will this tag an unformatted EBS volume as an ephemeral disk?

packages/os/ephemeral-storage.rules
```diff
@@ -0,0 +1,27 @@
...
+KERNEL!="nvme*|vd*", GOTO="ephemeral_storage_end"
+
+# We can't be sure that devices with similar kernel names are all ephemeral.
+# Our udev helper checks for known system partitions to classify the device.
+IMPORT{program}=="/usr/bin/ghostdog scan $devnode"
+ENV{BOTTLEROCKET_DEVICE_TYPE}!="ephemeral", GOTO="ephemeral_storage_end"
+
+# EBS volumes will show up as "Amazon Elastic Block Store", but we expect these
+# to be either system volumes or managed by the EBS CSI driver.
+KERNEL=="nvme*", ATTRS{model}=="Amazon EC2 NVMe Instance Storage*", ENV{ID_SERIAL}=="?*", \
```
samuelkarp, line 19: Is this only NVME EBS volumes?  What about instance types that do not support NVME devices or EBS volumes that might present differently (like the `st1`, `sc1`, or magnetic types)?

sources/ghostdog/src/main.rs
bcressey: It will, though we won't add a symlink in udev unless the model matches "Amazon EC2 NVMe Instance Storage".

I wanted to keep the code here pretty simple. At least some of the filtering has to happen in udev, or else we would execute a `ghostdog` process for all sorts of devices, so I thought it made sense to prefer udev rules where possible.

packages/os/ephemeral-storage.rules
bcressey: EBS volumes of all types won't match this rule, so we won't write the symlink.

Older instance types where the xen-blkfront driver is used for both EBS and ephemeral disks are a problem, since there's no easy way to tell them apart.

We could query IMDS to find the block device mapping and then try to match devices that way. This is complicated because the devices show up before the network is up. We'd have to do the query at a later point during boot, save the mapping somewhere, and trigger a change event on the devices so they get processed again.

I can backlog an issue for this, though I'd prefer to wait until someone says they need it.

samuelkarp (approved):

packages/os/ephemeral-storage.rules
bcressey: Opened #1181 so we can track any requests for this feature on older instance types, and to capture my design thoughts.

PR #1174, opened 2020-10-20
Makefile: split tasks.setup into tasks.setup and tasks.setup-build

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
Date:   Mon Oct 19 14:33:13 2020 -0700

    Makefile: add `setup-build` task
    
    Not all cargo-make tasks need docker, lz4 or curl (i.e. pubsys related tasks).
    This separates original setup task into `setup` and `setup-build`.
    `setup` sets up the build/output directories.
    `setup-build` checks for all the necessary tools to build and release
    images.

```

**Testing done:**
I'm still able to build and create AMIs.
I can `cargo make validate-repo` without needing docker.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -108,8 +108,14 @@ mkdir -p ${BUILDSYS_BUILD_DIR}
 mkdir -p ${BUILDSYS_OUTPUT_DIR}
 mkdir -p ${BUILDSYS_PACKAGES_DIR}
 mkdir -p ${GO_MOD_CACHE}
+'''
+]
 
-for cmd in bash curl docker gunzip lz4 ; do
+[tasks.setup-build]
+dependencies = ["setup"]
+script = [
+'''
+for cmd in curl docker gunzip lz4; do
   if ! command -v ${cmd} >/dev/null 2>&1 ; then
     echo "required program '${cmd}' not found" >&2
     exit 1
@@ -126,7 +132,7 @@ dependencies = [
 ]
 
 [tasks.fetch-sdk]
-dependencies = ["setup"]
+dependencies = ["setup-build"]
 script_runner = "bash"
 script = [
 '''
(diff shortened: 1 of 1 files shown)


Makefile.toml
```diff
@@ -157,7 +163,7 @@ chmod -R o+r ${CARGO_HOME}
 ]
 
 [tasks.fetch-vendored]
-dependencies = ["setup", "fetch-sdk"]
+dependencies = ["build-preflight-check", "fetch-sdk"]
```
tjkirch: You don't need to list all transitive dependencies; fetch-sdk already pulls in the check.  Same below.  I think we just change the "setup" dependency to the new one in most places below (other than your new non-build tasks) since "setup" should already already be pulled into almost everything transitively.

An alternative is splitting up the tool checks into individual tasks, and requiring them specifically below, e.g. a "docker" task, "lz4", etc.  It may be clearer?  You can no longer assume setup is done, but I guess we can't assume that now because of the new non-build tasks.

Makefile.toml
```diff
@@ -108,8 +108,14 @@ mkdir -p ${BUILDSYS_BUILD_DIR}
 mkdir -p ${BUILDSYS_OUTPUT_DIR}
 mkdir -p ${BUILDSYS_PACKAGES_DIR}
 mkdir -p ${GO_MOD_CACHE}
+'''
+]
 
-for cmd in bash curl docker gunzip lz4 ; do
+[tasks.build-preflight-check]
```
```suggestion
[tasks.setup-build]
```
Might need other setup tasks later, not just checks, and this is more consistent with the other task name.

etungsten:
Push above changes the task name to `setup-build` and cleaned up task dependencies to address @tjkirch 's comments.

etungsten:
Realized I could trim the dependencies further.

zmrow:
Where would the changes from #1167 fit in?

etungsten:
> Where would the changes from #1167 fit in?

I think those changes will still fit in `tasks.setup`.

webern (approved):

tjkirch (approved):

zmrow (approved):
:yarn:

PR #1176, opened 2020-10-21
update to systemd 246 and dbus-broker 24

### LINKED ISSUE #1015: Update to systemd 246
https://github.com/systemd/systemd/blob/master/NEWS

I tried updating from systemd 245 to 246, but the wicked nanny service fails to start about half of the time, maybe more - maybe a timing issue.  So far all I have is the console message saying "Failed to start wicked network nanny service", which prevents most networking from starting up.

### LINKED ISSUE #1175: update to dbus-broker 24
[dbus-broker 24 is out!](https://github.com/bus1/dbus-broker/releases/tag/v24)

bcressey:
**Issue number:**
Fixes #1015, fixes #1175 

**Description of changes:**
Update systemd and dbus-broker to the latest release.

Work around a behavior change in systemd-notify that caused an SELinux denial when starting apiserver.

**Testing done:**
Launched x86_64 and aarch64 nodes. Confirmed that the nodes joined the cluster, ran pods OK.

**Terms of contribution:**

### DIFF
--- a/packages/dbus-broker/Cargo.toml
+++ b/packages/dbus-broker/Cargo.toml
@@ -9,12 +9,11 @@ build = "build.rs"
 path = "pkg.rs"
 
 [[package.metadata.build-package.external-files]]
-url = "https://github.com/bus1/dbus-broker/releases/download/v23/dbus-broker-23.tar.xz"
-sha512 = "b9431a9733342d4b6214a6a35ad05d3278ab2c8cfdc35537dba86ba9ee3f59eef4e31687c2fe8a6444072eaf8384bc71a41617fec0539a8c00be9eae0efa9832"
+url = "https://github.com/bus1/dbus-broker/releases/download/v24/dbus-broker-24.tar.xz"
+sha512 = "7fbe6c689eff27ec842bb9d839a418abc356b026bb0a54dfa8b680a655409aa1ea4cb90655a6b04561a88c6e703cacf8800fddfe1abc21e7ee60db2dac1c2db9"
 
 [build-dependencies]
 glibc = { path = "../glibc" }
-libcap = { path = "../libcap" }
 libexpat = { path = "../libexpat" }
 libselinux = { path = "../libselinux" }
 systemd = { path = "../systemd" }
--- a/sources/api/apiserver/src/server/mod.rs
+++ b/sources/api/apiserver/src/server/mod.rs
@@ -36,6 +36,7 @@ fn notify_unix_socket_ready() -> Result<()> {
         ensure!(
             Command::new("systemd-notify")
                 .arg("--ready")
+                .arg("--no-block")
                 .status()
                 .context(error::SystemdNotify)?
                 .success(),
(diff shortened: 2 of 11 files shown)


packages/dbus-broker/Cargo.toml
```diff
@@ -9,12 +9,11 @@ build = "build.rs"
 path = "pkg.rs"
 
 [[package.metadata.build-package.external-files]]
-url = "https://github.com/bus1/dbus-broker/releases/download/v23/dbus-broker-23.tar.xz"
-sha512 = "b9431a9733342d4b6214a6a35ad05d3278ab2c8cfdc35537dba86ba9ee3f59eef4e31687c2fe8a6444072eaf8384bc71a41617fec0539a8c00be9eae0efa9832"
+url = "https://github.com/bus1/dbus-broker/releases/download/v23/dbus-broker-24.tar.xz"
```
tjkirch:
```suggestion
url = "https://github.com/bus1/dbus-broker/releases/download/v24/dbus-broker-24.tar.xz"
```
Otherwise this actually downloads 23 (eek) and has the wrong checksum.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -36,6 +36,7 @@ fn notify_unix_socket_ready() -> Result<()> {
         ensure!(
             Command::new("systemd-notify")
                 .arg("--ready")
+                .arg("--no-block")
```
tjkirch, line 39: Do we need to worry about any of the service units defined with `Type=notify` having the issue you ran into that required this, or does systemd handle it OK internally?  I don't have enough context.

packages/dbus-broker/Cargo.toml
bcressey: 😞

sources/api/apiserver/src/server/mod.rs
bcressey: The out-of-tree ones would need to be updated to use the new barrier function, and that doesn't seem to have happened yet.

I expect we'd hit the same SELinux denials in that case. I'm OK with changing the policy if we do.

tjkirch: By out-of-tree you mean not shipped with systemd/dbus?  I think that's all of our `Type=notify` cases except for dbus-broker itself, so 8 service units.  I'm assuming you didn't see SELinux denials, though, so I may be misunderstanding.

bcressey: By out-of-tree I meant not part of systemd:
* apiserver
* containerd
* dbus-broker
* wickedd

In-tree ones could have been updated to use the barrier function. I doubt the others would have since the systemd release is so new and they would need a release cycle to adopt it, and also need a reason to worry about the race condition the feature was meant to address.

`apiserver` is unusual in that it uses the `systemd-notify` client directly, which now defaults to blocking via the barrier function.

tjkirch (approved):

webern (approved):

Issue #1136, opened 2020-09-30
MySQL 5.7 container OOM kill / changing the NOFILE limit


zlangbert:
I am running a 1.17 EKS cluster that was we recently tried to switch to bottlerocket nodes. We ran into an issue where MySQL 5.7 containers were being OOM killed. This does not happen on the latest AL2 EKS AMI. 

After some searching I came across these issues / PRs:
https://github.com/docker-library/mysql/issues/579
https://github.com/kubernetes-sigs/kind/pull/760
https://github.com/containerd/containerd/issues/3201
https://github.com/containerd/containerd/pull/3202 (this one has since been [reverted](https://github.com/containerd/containerd/commit/c691c36614622b205a859ae7e656badb4a553076#diff-5f2ec709a3860c8863a703cffa7b4921) without addressing the reason it was changed in the first place)

I am having trouble actually testing this change on bottlerocket, but it appears that setting [LimitNOFILE](https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/containerd/containerd.service#L17) to something like `1048576` should fix this issue.

I also found that the `mysql:8` image does not exhibit this issue.

So couple questions:
1. This doesn't seem like a bottlerocket bug necessarily, but wanted to at least report the issue and see if it something that could be addressed, especially given the behavior change compared to the AL2 EKS AMI.
2. Is there any friendly way of making that kind of modification to the containerd unit file without rebuilding the image myself? Even on an already running box I was having trouble trying to test this because I couldn't find a way to edit the file after `sudo sheltie`

**What I expected to happen:**
I would expect the mysql container to start and use a resonable amount of memory.

**What actually happened:**
The mysql container attempts to consume many gigabytes of ram and is OOM killed when there are resource limits placed on it.

**How to reproduce the problem:**

Run an EKS cluster with bottlerocket nodes and apply this manifest. The mysql pod will go into a restart loop being OOM killed over and over. If you remove the resource limits you can see the pod will take up to 16gb of RAM doing nothing.

```
apiVersion: v1
kind: Pod
metadata:
  name: mysql-test
  namespace: default
spec:
  containers:
    - name: mysql
      imagePullPolicy: Always
      image: 'mysql:5.7'
      env:
        - name: MYSQL_DATABASE
          value: test
        - name: MYSQL_ROOT_PASSWORD
          value: test
      resources:
        limits:
          cpu: 500m
          memory: 500Mi
        requests:
          cpu: 500m
[...]

PR #1180, opened 2020-10-23
cri: set default RLIMIT_NOFILE

zmrow:
**Issue number:**
Related to #1136 

**Description of changes:**
```
The `cri` plugin currently inherits the limit from the default OCI spec or the
containerd process.  This change sets the default RLIMIT_NOFILE to 1048576 in
the OCI spec for any container spawned using `cri`.
```
The intent is to upstream this patch, but there is additional work to be done to make it compatible with `containerd` >1.3.  We also don't have some of the upstream unit tests in our `containerd` package; we'll need to write one for this change when we submit upstream. 

Given that the AL2 AMI uses this limit, albeit on the  `containerd` process, it seems a reasonable number to use for container processes spawned by `cri`.

**Testing done:**
Ran a busybox pod on a Bottlerocket AMI built using this patch:
```
~/bottlerocket/bottlerocket $ kubectl run -i -t busybox --image=busybox --restart=Never                                                  
If you don't see a command prompt, try pressing enter.
/ # ulimit -a
...
open files                      (-n) 65536
...
```

Ran a busybox pod on the latest released Bottlerocket AMI:
```
~/bottlerocket/bottlerocket $ kubectl run -i -t busybox --image=busybox --restart=Never
If you don't see a command prompt, try pressing enter.
/ # ulimit -a
...
open files                      (-n) 1073741816
...
```

Ran the MySQL manifest from #1136 with an AMI built using this patch and confirmed the pod runs just fine without any OOM issues that I could see. (I also ran the pod with the current released Bottlerocket AMI and confirmed that it does indeed OOM)

**Terms of contribution:**

### DIFF
--- a/packages/containerd/3001-cri-set-default-RLIMIT_NOFILE.patch
+++ b/packages/containerd/3001-cri-set-default-RLIMIT_NOFILE.patch
@@ -1,0 +1,88 @@
+From 64deef11da4cf532ea9d82cc499f5174a8907e29 Mon Sep 17 00:00:00 2001
+From: Zac Mrowicki <mrowicki@amazon.com>
+Date: Thu, 22 Oct 2020 20:44:38 +0000
+Subject: [PATCH] cri: set default RLIMIT_NOFILE
+
+The `cri` plugin currently inherits the limit from the default OCI spec
+or the containerd process.  This change sets the default hard
+RLIMIT_NOFILE to 1048576 and the soft limit to 65536 in the OCI spec for
+any container spawned using `cri`.
+---
+ vendor/github.com/containerd/cri/pkg/config/config.go |  8 ++++++++
+ .../containerd/cri/pkg/containerd/opts/spec.go        | 11 +++++++++++
+ .../containerd/cri/pkg/server/container_create.go     | 11 +++++++++++
+ 3 files changed, 30 insertions(+)
+
+diff --git a/vendor/github.com/containerd/cri/pkg/config/config.go b/vendor/github.com/containerd/cri/pkg/config/config.go
+index 7c5f9eb..772bf28 100644
+--- a/vendor/github.com/containerd/cri/pkg/config/config.go
++++ b/vendor/github.com/containerd/cri/pkg/config/config.go
+@@ -204,6 +204,12 @@ type PluginConfig struct {
+ 	// DisableProcMount disables Kubernetes ProcMount support. This MUST be set to `true`
+ 	// when using containerd with Kubernetes <=1.11.
+ 	DisableProcMount bool `toml:"disable_proc_mount" json:"disableProcMount"`
++	// ProcessRLimitNoFileSoft sets the soft limit of maximum file
++	// descriptors each container process can use.
++	ProcessRLimitNoFileSoft int `toml:"process_rlimit_no_file_soft" json:"process_rlimit_no_file_soft"`
++	// ProcessRLimitNoFileHard sets the hard limit of maximum file
++	// descriptors each container process can use.
++	ProcessRLimitNoFileHard int `toml:"process_rlimit_no_file_hard" json:"process_rlimit_no_file_hard"`
+ }
+ 
+ // X509KeyPairStreaming contains the x509 configuration for streaming
+@@ -271,6 +277,8 @@ func DefaultConfig() PluginConfig {
+ 		},
+ 		MaxConcurrentDownloads: 3,
+ 		DisableProcMount:       false,
++		ProcessRLimitNoFileSoft:    65536,
++		ProcessRLimitNoFileHard:    1048576,
+ 	}
+ }
+ 
+diff --git a/vendor/github.com/containerd/cri/pkg/containerd/opts/spec.go b/vendor/github.com/containerd/cri/pkg/containerd/opts/spec.go
+index 0da421a..445c279 100644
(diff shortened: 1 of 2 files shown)


packages/containerd/3001-cri-set-a-default-RLIMIT_NOFILE-of-1048576.patch
```diff
@@ -0,0 +1,83 @@
...
++        // process can use.
++        ProcessRLimitNoFile int `toml:"process_rlimit_no_file" json:"process_rlimit_no_file"`
+ }
+ 
+ // X509KeyPairStreaming contains the x509 configuration for streaming
+@@ -271,6 +274,7 @@ func DefaultConfig() PluginConfig {
+ 		},
+ 		MaxConcurrentDownloads: 3,
+ 		DisableProcMount:       false,
++                ProcessRLimitNoFile:    1048576,
```
samuelkarp: Should we match what we're doing for [Docker](https://github.com/bottlerocket-os/bottlerocket/blob/1b8721a72784bfd93f25c95a56f62ee8c7649616/packages/docker-engine/daemon.json#L9) (soft: 1024, hard: 4096) or [containerd's default](https://github.com/containerd/containerd/blob/656b487d333db8ad7134d26cfe12358c46c2b7f8/oci/spec.go#L158-L160) (1024 for both) instead?

samuelkarp (commented):
Code looks good to me, just one question about what the value should be.

packages/containerd/3001-cri-set-a-default-RLIMIT_NOFILE-of-1048576.patch
bcressey: Those limits seem too low for databases, and there's a report that [Elasticsearch doesn't work without more file descriptors](https://github.com/pires/kubernetes-elasticsearch-cluster/issues/215).

I would also tend toward a larger number for Docker, and perhaps override this in `host-ctr` for containers launched in that path.

Maybe `65536` soft and `1048576` hard?

samuelkarp: > I would also tend toward a larger number for Docker

For the ECS variant, the limits can be adjusted in the task definition.

webern (approved):
Wow, nice job.

packages/containerd/3001-cri-set-a-default-RLIMIT_NOFILE-of-1048576.patch
zmrow: Would we want to provide config struct members for soft/hard?
```
...
ProcessRLimitNoFileSoft:    65536,
ProcessRLimitNoFileHard:    1048576,
...
```

zmrow:
^ added a soft limit of 65536 and confirmed this is what the node sees. 
```
open files                      (-n) 65536
```

packages/containerd/3001-cri-set-a-default-RLIMIT_NOFILE-of-1048576.patch
zmrow: Added config struct members for soft/hard - but since we're using these default limits and not overriding them anywhere, the node shows `65536` for it's open files limit.  Will this be good enough?

packages/containerd/3001-cri-set-default-RLIMIT_NOFILE.patch
```diff
@@ -0,0 +1,87 @@
...
+ // X509KeyPairStreaming contains the x509 configuration for streaming
+@@ -271,6 +277,8 @@ func DefaultConfig() PluginConfig {
+ 		},
+ 		MaxConcurrentDownloads: 3,
+ 		DisableProcMount:       false,
++                ProcessRLimitNoFileSoft:    65536,
++                ProcessRLimitNoFileHard:    1048576,
+ 	}
+ }
+
```
bcressey: The existing code is indented with tabs, but you've used spaces. It'd be good to match the whitespace convention in case this is headed upstream in some form.

zmrow:
I ran the MySQL manifest from #1136 with an AMI built using this patch and confirmed the pod runs just fine without any OOM issues that I could see.  (I also ran the pod with the current released Bottlerocket AMI and confirmed that it does indeed OOM)
```
apiVersion: v1
kind: Pod
metadata:
  name: mysql-test
  namespace: default
spec:
  containers:
    - name: mysql
      imagePullPolicy: Always
      image: 'mysql:5.7'
      env:
        - name: MYSQL_DATABASE
          value: test
        - name: MYSQL_ROOT_PASSWORD
          value: test
```

packages/containerd/3001-cri-set-default-RLIMIT_NOFILE.patch
zmrow: Good catch!

zmrow:
^ the above force push fixes up the spacing issues @bcressey caught 

(Thanks `:set list` !)

bcressey (approved):

webern (approved):

packages/containerd/3001-cri-set-default-RLIMIT_NOFILE.patch
samuelkarp: For Go changes, it's a good idea to run them through `go fmt` and `goimports`.  This will fix issues like tabs vs spaces.

samuelkarp (approved):

PR #1184, opened 2020-10-26
Adds rtcsync directive to chrony config file

### LINKED ISSUE #1183: node_timex_sync_status metric indicates that NTP is not syncing
**Image I'm using:**

bottlerocket-aws-k8s-1.18-x86_64-v1.0.2-ddeb03c8 - ami-00e1ed6a3ebe06fd0

We are running prometheus node exporter as a Kubernetes DemonSet.

We expected the `node_timex_sync_status` metric to be 1 if NTP is working properly.

**What actually happened:**

The `NodeClockNotSynchronising` alert on our prometheus instance fires: https://github.com/prometheus/node_exporter/blob/v1.0.1/docs/node-mixin/alerts/alerts.libsonnet#L237-L249 because `node_timex_sync_status` is 0

It seems like the clock on our instances is set correctly (but this doesn't prove that NTP is working correctly)

There are no logs indicating a problem either on the node exporter, or on chronyd on the host.

It's hard to debug this issue as there is no chronyc available when using the admin container to debug the host!

Run the prometheus node exporter e.g. https://github.com/prometheus-operator/kube-prometheus/blob/v0.6.0/manifests/node-exporter-daemonset.yaml

And check the `node_timex_sync_status` metric

errm:
**Issue number:**

Fixes #1183

**Description of changes:**

Adds the `rtcsync` directive to the chrony config file.

I think this is a resonable change to make, as it allready seems to be the default in Amazon Linux 2.

**Testing done:**

I have added this directive to the config file on a running instance and restarted chronyd, and observed that the value of `node_timex_sync` exported by the prometheus node exporter is 1.

**Terms of contribution:**

### DIFF
--- a/packages/chrony/chrony-conf
+++ b/packages/chrony/chrony-conf
@@ -6,3 +6,4 @@ makestep 1.0 3
 dumponexit
 dumpdir /var/lib/chrony
 user chrony
+rtcsync


zmrow:
This looks reasonable to me, but... paging @bcressey

bcressey (approved):

bcressey:
LGTM, thanks!

webern (approved):

PR #1204, opened 2020-11-11
settings.network: add new proxy settings

### LINKED ISSUE #1178: Add support for HTTP_PROXY environment variable configuration into various services
**What I'd like:**
I would like to add  HTTP_PROXY related environment variables to the various services (example - kubelet & containerd/docker) configuration files so that they can connect to the internet through the proxy server.

May be global configuration that allows HTTP_PROXY environment variables and parameter in each service to enable global config.

No, as I understand, by design it is not possible to dynamically add any configuration that does not exist into the [data models](https://github.com/bottlerocket-os/bottlerocket/tree/develop/sources/models)

etungsten:
**Issue number:**
Addresses https://github.com/bottlerocket-os/bottlerocket/issues/1178

**Description of changes:**

```
Date:   Wed Nov 11 00:06:23 2020 -0800

    settings.network: add new proxy settings
    
    Add new settings for configuring HTTPS_PROXY/https_proxy and
    NO_PROXY/no_proxy environment variables.
    
    These environment variables will only be exposed to a limited set
    of services that need them.

```

```
Date:   Tue Nov 10 23:49:38 2020 -0800

    schnauzer: add `host` helper
    
    Adds a new rendering-template helper that trims a fully qualified URL
    setting down to just its host portion.

```

**Testing done:**
Set up a simple HTTPS proxy server with [tinyproxy](http://tinyproxy.github.io/) with mostly default settings.

- [x] Kubernetes Variant testing

Built aws-k8s-1.17/arm64 AMI and launched several instances with the following settings in my userdata:
```
[settings.network]
https-proxy="myhttpserver:9898"
no-proxy=[".test"]
```
Host containers, both admin and control, come up fine.

All of the nodes came up fine in my EKS cluster. Can pull images and run pods.

Checked the HTTPS proxy server access logs and observed traffic from the instances I launched:
```
INFO      Dec 11 21:23:38 [18613]: Starting main loop. Accepting connections.
CONNECT   Dec 11 21:26:53 [18625]: Connect (file descriptor 7): ec2-18-236-101-17.us-west-2.compute.amazonaws.com [18.236.101.17]
CONNECT   Dec 11 21:26:53 [18625]: Request (file descriptor 7): CONNECT api.ecr.us-west-2.amazonaws.com:443 HTTP/1.1
INFO      Dec 11 21:26:53 [18625]: No upstream proxy for api.ecr.us-west-2.amazonaws.com
INFO      Dec 11 21:26:53 [18625]: opensock: opening connection to api.ecr.us-west-2.amazonaws.com:443
INFO      Dec 11 21:26:53 [18625]: opensock: getaddrinfo returned for api.ecr.us-west-2.amazonaws.com:443
CONNECT   Dec 11 21:26:53 [18625]: Established connection to host "api.ecr.us-west-2.amazonaws.com" using file descriptor 8.
INFO      Dec 11 21:26:53 [18625]: Not sending client headers to remote machine
INFO      Dec 11 21:26:54 [18625]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:26:54 [18624]: Connect (file descriptor 7): ec2-18-236-101-17.us-west-2.compute.amazonaws.com [18.236.101.17]
CONNECT   Dec 11 21:26:54 [18624]: Request (file descriptor 7): CONNECT ec2.us-west-2.amazonaws.com:443 HTTP/1.1
INFO      Dec 11 21:26:54 [18624]: No upstream proxy for ec2.us-west-2.amazonaws.com
INFO      Dec 11 21:26:54 [18624]: opensock: opening connection to ec2.us-west-2.amazonaws.com:443
INFO      Dec 11 21:26:54 [18624]: opensock: getaddrinfo returned for ec2.us-west-2.amazonaws.com:443
CONNECT   Dec 11 21:26:54 [18624]: Established connection to host "ec2.us-west-2.amazonaws.com" using file descriptor 8.
INFO      Dec 11 21:26:54 [18624]: Not sending client headers to remote machine
INFO      Dec 11 21:26:54 [18624]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:26:54 [18625]: Connect (file descriptor 7): ec2-18-236-101-17.us-west-2.compute.amazonaws.com [18.236.101.17]
CONNECT   Dec 11 21:26:54 [18625]: Request (file descriptor 7): CONNECT api.ecr.us-west-2.amazonaws.com:443 HTTP/1.1
INFO      Dec 11 21:26:54 [18625]: No upstream proxy for api.ecr.us-west-2.amazonaws.com
INFO      Dec 11 21:26:54 [18625]: opensock: opening connection to api.ecr.us-west-2.amazonaws.com:443
INFO      Dec 11 21:26:54 [18625]: opensock: getaddrinfo returned for api.ecr.us-west-2.amazonaws.com:443
CONNECT   Dec 11 21:26:54 [18625]: Established connection to host "api.ecr.us-west-2.amazonaws.com" using file descriptor 8.
INFO      Dec 11 21:26:54 [18625]: Not sending client headers to remote machine
INFO      Dec 11 21:26:54 [18625]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:26:54 [18623]: Connect (file descriptor 7): ec2-18-236-101-17.us-west-2.compute.amazonaws.com [18.236.101.17]
CONNECT   Dec 11 21:26:54 [18623]: Request (file descriptor 7): CONNECT ec2.us-west-2.amazonaws.com:443 HTTP/1.1
INFO      Dec 11 21:26:54 [18623]: No upstream proxy for ec2.us-west-2.amazonaws.com
INFO      Dec 11 21:26:54 [18623]: opensock: opening connection to ec2.us-west-2.amazonaws.com:443
INFO      Dec 11 21:26:54 [18623]: opensock: getaddrinfo returned for ec2.us-west-2.amazonaws.com:443
CONNECT   Dec 11 21:26:54 [18623]: Established connection to host "ec2.us-west-2.amazonaws.com" using file descriptor 8.
INFO      Dec 11 21:26:54 [18623]: Not sending client headers to remote machine
INFO      Dec 11 21:28:02 [18623]: Closed connection between local client (fd:7) and remote client (fd:8)

```

Ran Kubernetes certified-conformance tests and all of them pass.

To make sure that no_proxy'ing the Kubernetes API server works, I limited the API endpoint access to within the cluster's private VPC and the nodes fail to connect to the cluster as expected if I do not no_proxy the API server endpoint. Once I do (by rendering the `proxy-env` template with `settings.kubernetes.apiserver`), the node comes up fine.

- [x] **ECS variant testing**

Launched ECS variant with https-proxy set to my https proxy server. Started a simple nginx task and it runs fine. I'm able curl localhost:80 and get the nginx welcome page.
Observed traffic on my https server:

<details>

```
CONNECT   Dec 11 21:51:47 [19062]: Connect (file descriptor 7): ec2-52-36-254-125.us-west-2.compute.amazonaws.com [52.36.254.125]
CONNECT   Dec 11 21:51:47 [19062]: Request (file descriptor 7): CONNECT registry-1.docker.io:443 HTTP/1.1
INFO      Dec 11 21:51:47 [19062]: No upstream proxy for registry-1.docker.io
INFO      Dec 11 21:51:47 [19062]: opensock: opening connection to registry-1.docker.io:443
INFO      Dec 11 21:51:47 [19062]: opensock: getaddrinfo returned for registry-1.docker.io:443
CONNECT   Dec 11 21:51:47 [19062]: Established connection to host "registry-1.docker.io" using file descriptor 8.
INFO      Dec 11 21:51:47 [19062]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:47 [18625]: Connect (file descriptor 7): ec2-52-36-254-125.us-west-2.compute.amazonaws.com [52.36.254.125]
CONNECT   Dec 11 21:51:47 [18625]: Request (file descriptor 7): CONNECT auth.docker.io:443 HTTP/1.1
INFO      Dec 11 21:51:47 [18625]: No upstream proxy for auth.docker.io
INFO      Dec 11 21:51:47 [18625]: opensock: opening connection to auth.docker.io:443
INFO      Dec 11 21:51:47 [18625]: opensock: getaddrinfo returned for auth.docker.io:443
CONNECT   Dec 11 21:51:47 [18625]: Established connection to host "auth.docker.io" using file descriptor 8.
INFO      Dec 11 21:51:47 [18625]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:47 [19056]: Connect (file descriptor 7): ec2-52-36-254-125.us-west-2.compute.amazonaws.com [52.36.254.125]
CONNECT   Dec 11 21:51:47 [19056]: Request (file descriptor 7): CONNECT registry-1.docker.io:443 HTTP/1.1
INFO      Dec 11 21:51:47 [19056]: No upstream proxy for registry-1.docker.io
INFO      Dec 11 21:51:47 [19056]: opensock: opening connection to registry-1.docker.io:443
INFO      Dec 11 21:51:47 [19056]: opensock: getaddrinfo returned for registry-1.docker.io:443
CONNECT   Dec 11 21:51:47 [19056]: Established connection to host "registry-1.docker.io" using file descriptor 8.
INFO      Dec 11 21:51:47 [19056]: Not sending client headers to remote machine
INFO      Dec 11 21:51:48 [19056]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:51:48 [19061]: Connect (file descriptor 7): ec2-52-36-254-125.us-west-2.compute.amazonaws.com [52.36.254.125]
CONNECT   Dec 11 21:51:48 [19061]: Request (file descriptor 7): CONNECT auth.docker.io:443 HTTP/1.1
INFO      Dec 11 21:51:48 [19061]: No upstream proxy for auth.docker.io
INFO      Dec 11 21:51:48 [19061]: opensock: opening connection to auth.docker.io:443
INFO      Dec 11 21:51:48 [19061]: opensock: getaddrinfo returned for auth.docker.io:443
CONNECT   Dec 11 21:51:48 [19061]: Established connection to host "auth.docker.io" using file descriptor 8.
INFO      Dec 11 21:51:48 [19061]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:48 [19066]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:48 [19066]: Request (file descriptor 7): CONNECT registry.hub.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:48 [19066]: No upstream proxy for registry.hub.docker.com
INFO      Dec 11 21:51:48 [19066]: opensock: opening connection to registry.hub.docker.com:443
INFO      Dec 11 21:51:48 [19066]: opensock: getaddrinfo returned for registry.hub.docker.com:443
CONNECT   Dec 11 21:51:48 [19066]: Established connection to host "registry.hub.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:48 [19066]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:48 [18624]: Connect (file descriptor 7): ec2-52-36-254-125.us-west-2.compute.amazonaws.com [52.36.254.125]
CONNECT   Dec 11 21:51:48 [18624]: Request (file descriptor 7): CONNECT registry-1.docker.io:443 HTTP/1.1
INFO      Dec 11 21:51:48 [18624]: No upstream proxy for registry-1.docker.io
INFO      Dec 11 21:51:48 [18624]: opensock: opening connection to registry-1.docker.io:443
INFO      Dec 11 21:51:48 [18624]: opensock: getaddrinfo returned for registry-1.docker.io:443
CONNECT   Dec 11 21:51:48 [18624]: Established connection to host "registry-1.docker.io" using file descriptor 8.
INFO      Dec 11 21:51:48 [18624]: Not sending client headers to remote machine
INFO      Dec 11 21:51:48 [19066]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:51:48 [19066]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:48 [19066]: Request (file descriptor 7): CONNECT auth.docker.io:443 HTTP/1.1
INFO      Dec 11 21:51:48 [19066]: No upstream proxy for auth.docker.io
INFO      Dec 11 21:51:48 [19066]: opensock: opening connection to auth.docker.io:443
INFO      Dec 11 21:51:48 [19066]: opensock: getaddrinfo returned for auth.docker.io:443
CONNECT   Dec 11 21:51:48 [19066]: Established connection to host "auth.docker.io" using file descriptor 8.
INFO      Dec 11 21:51:48 [19066]: Not sending client headers to remote machine
INFO      Dec 11 21:51:48 [19066]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:51:48 [19072]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:48 [19072]: Request (file descriptor 7): CONNECT registry.hub.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:48 [19072]: No upstream proxy for registry.hub.docker.com
INFO      Dec 11 21:51:48 [19072]: opensock: opening connection to registry.hub.docker.com:443
INFO      Dec 11 21:51:48 [19072]: opensock: getaddrinfo returned for registry.hub.docker.com:443
CONNECT   Dec 11 21:51:48 [19072]: Established connection to host "registry.hub.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:48 [19072]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:49 [19058]: Connect (file descriptor 7): ec2-52-36-254-125.us-west-2.compute.amazonaws.com [52.36.254.125]
CONNECT   Dec 11 21:51:49 [19058]: Request (file descriptor 7): CONNECT registry-1.docker.io:443 HTTP/1.1
INFO      Dec 11 21:51:49 [19058]: No upstream proxy for registry-1.docker.io
INFO      Dec 11 21:51:49 [19058]: opensock: opening connection to registry-1.docker.io:443
CONNECT   Dec 11 21:51:49 [18619]: Connect (file descriptor 7): ec2-52-36-254-125.us-west-2.compute.amazonaws.com [52.36.254.125]
INFO      Dec 11 21:51:49 [19058]: opensock: getaddrinfo returned for registry-1.docker.io:443
CONNECT   Dec 11 21:51:49 [18619]: Request (file descriptor 7): CONNECT registry-1.docker.io:443 HTTP/1.1
INFO      Dec 11 21:51:49 [18619]: No upstream proxy for registry-1.docker.io
INFO      Dec 11 21:51:49 [18619]: opensock: opening connection to registry-1.docker.io:443
INFO      Dec 11 21:51:49 [18619]: opensock: getaddrinfo returned for registry-1.docker.io:443
CONNECT   Dec 11 21:51:49 [19058]: Established connection to host "registry-1.docker.io" using file descriptor 8.
INFO      Dec 11 21:51:49 [19058]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:49 [18619]: Established connection to host "registry-1.docker.io" using file descriptor 8.
INFO      Dec 11 21:51:49 [18619]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:49 [19065]: Connect (file descriptor 7): ec2-52-36-254-125.us-west-2.compute.amazonaws.com [52.36.254.125]
CONNECT   Dec 11 21:51:49 [19065]: Request (file descriptor 7): CONNECT production.cloudflare.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:49 [19065]: No upstream proxy for production.cloudflare.docker.com
INFO      Dec 11 21:51:49 [19065]: opensock: opening connection to production.cloudflare.docker.com:443
INFO      Dec 11 21:51:49 [19065]: opensock: getaddrinfo returned for production.cloudflare.docker.com:443
CONNECT   Dec 11 21:51:49 [19065]: Established connection to host "production.cloudflare.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:49 [19065]: Not sending client headers to remote machine
INFO      Dec 11 21:51:49 [19072]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:51:49 [18623]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:49 [18623]: Request (file descriptor 7): CONNECT registry.hub.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:49 [18623]: No upstream proxy for registry.hub.docker.com
INFO      Dec 11 21:51:49 [18623]: opensock: opening connection to registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [18623]: opensock: getaddrinfo returned for registry.hub.docker.com:443
CONNECT   Dec 11 21:51:49 [19063]: Connect (file descriptor 7): ec2-52-36-254-125.us-west-2.compute.amazonaws.com [52.36.254.125]
CONNECT   Dec 11 21:51:49 [19063]: Request (file descriptor 7): CONNECT production.cloudflare.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:49 [19063]: No upstream proxy for production.cloudflare.docker.com
INFO      Dec 11 21:51:49 [19063]: opensock: opening connection to production.cloudflare.docker.com:443
INFO      Dec 11 21:51:49 [19063]: opensock: getaddrinfo returned for production.cloudflare.docker.com:443
CONNECT   Dec 11 21:51:49 [19063]: Established connection to host "production.cloudflare.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:49 [19063]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:49 [18623]: Established connection to host "registry.hub.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:49 [18623]: Not sending client headers to remote machine
INFO      Dec 11 21:51:49 [18619]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:49 [18623]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:51:49 [19071]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:49 [18623]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:49 [19059]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:49 [18621]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:49 [19067]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:49 [19070]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:49 [18623]: Request (file descriptor 7): CONNECT registry.hub.docker.com:443 HTTP/1.1
CONNECT   Dec 11 21:51:49 [19071]: Request (file descriptor 7): CONNECT registry.hub.docker.com:443 HTTP/1.1
CONNECT   Dec 11 21:51:49 [18621]: Request (file descriptor 7): CONNECT registry.hub.docker.com:443 HTTP/1.1
CONNECT   Dec 11 21:51:49 [19059]: Request (file descriptor 7): CONNECT registry.hub.docker.com:443 HTTP/1.1
CONNECT   Dec 11 21:51:49 [19067]: Request (file descriptor 7): CONNECT registry.hub.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:49 [18621]: No upstream proxy for registry.hub.docker.com
CONNECT   Dec 11 21:51:49 [19070]: Request (file descriptor 7): CONNECT registry.hub.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:49 [19059]: No upstream proxy for registry.hub.docker.com
INFO      Dec 11 21:51:49 [19071]: No upstream proxy for registry.hub.docker.com
INFO      Dec 11 21:51:49 [18623]: No upstream proxy for registry.hub.docker.com
INFO      Dec 11 21:51:49 [18621]: opensock: opening connection to registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [19067]: No upstream proxy for registry.hub.docker.com
INFO      Dec 11 21:51:49 [19070]: No upstream proxy for registry.hub.docker.com
INFO      Dec 11 21:51:49 [19071]: opensock: opening connection to registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [19059]: opensock: opening connection to registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [19067]: opensock: opening connection to registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [18623]: opensock: opening connection to registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [18621]: opensock: getaddrinfo returned for registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [19071]: opensock: getaddrinfo returned for registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [19070]: opensock: opening connection to registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [19067]: opensock: getaddrinfo returned for registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [19059]: opensock: getaddrinfo returned for registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [18623]: opensock: getaddrinfo returned for registry.hub.docker.com:443
INFO      Dec 11 21:51:49 [19070]: opensock: getaddrinfo returned for registry.hub.docker.com:443
CONNECT   Dec 11 21:51:49 [19071]: Established connection to host "registry.hub.docker.com" using file descriptor 8.
CONNECT   Dec 11 21:51:49 [19059]: Established connection to host "registry.hub.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:49 [19059]: Not sending client headers to remote machine
INFO      Dec 11 21:51:49 [19071]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:49 [18621]: Established connection to host "registry.hub.docker.com" using file descriptor 8.
CONNECT   Dec 11 21:51:49 [18623]: Established connection to host "registry.hub.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:49 [18621]: Not sending client headers to remote machine
INFO      Dec 11 21:51:49 [18623]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:49 [19070]: Established connection to host "registry.hub.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:49 [19070]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:49 [19067]: Established connection to host "registry.hub.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:49 [19067]: Not sending client headers to remote machine
INFO      Dec 11 21:51:50 [19059]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:51:50 [19056]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:50 [19056]: Request (file descriptor 7): CONNECT production.cloudflare.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:50 [19070]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:50 [19056]: No upstream proxy for production.cloudflare.docker.com
CONNECT   Dec 11 21:51:50 [19070]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
INFO      Dec 11 21:51:50 [19071]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:51:50 [19070]: Request (file descriptor 7): CONNECT production.cloudflare.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:50 [19056]: opensock: opening connection to production.cloudflare.docker.com:443
CONNECT   Dec 11 21:51:50 [19066]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
INFO      Dec 11 21:51:50 [19070]: No upstream proxy for production.cloudflare.docker.com
INFO      Dec 11 21:51:50 [19056]: opensock: getaddrinfo returned for production.cloudflare.docker.com:443
INFO      Dec 11 21:51:50 [19070]: opensock: opening connection to production.cloudflare.docker.com:443
CONNECT   Dec 11 21:51:50 [19066]: Request (file descriptor 7): CONNECT production.cloudflare.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:50 [19066]: No upstream proxy for production.cloudflare.docker.com
INFO      Dec 11 21:51:50 [19070]: opensock: getaddrinfo returned for production.cloudflare.docker.com:443
INFO      Dec 11 21:51:50 [18621]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:50 [19066]: opensock: opening connection to production.cloudflare.docker.com:443
CONNECT   Dec 11 21:51:50 [19059]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:50 [19059]: Request (file descriptor 7): CONNECT production.cloudflare.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:50 [19066]: opensock: getaddrinfo returned for production.cloudflare.docker.com:443
INFO      Dec 11 21:51:50 [19059]: No upstream proxy for production.cloudflare.docker.com
CONNECT   Dec 11 21:51:50 [19056]: Established connection to host "production.cloudflare.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:50 [19059]: opensock: opening connection to production.cloudflare.docker.com:443
CONNECT   Dec 11 21:51:50 [19070]: Established connection to host "production.cloudflare.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:50 [19056]: Not sending client headers to remote machine
INFO      Dec 11 21:51:50 [19059]: opensock: getaddrinfo returned for production.cloudflare.docker.com:443
INFO      Dec 11 21:51:50 [19070]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:50 [19066]: Established connection to host "production.cloudflare.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:50 [19066]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:50 [19059]: Established connection to host "production.cloudflare.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:50 [19059]: Not sending client headers to remote machine
INFO      Dec 11 21:51:50 [19067]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:51:50 [18619]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:50 [18619]: Request (file descriptor 7): CONNECT production.cloudflare.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:50 [18619]: No upstream proxy for production.cloudflare.docker.com
INFO      Dec 11 21:51:50 [18623]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:50 [18619]: opensock: opening connection to production.cloudflare.docker.com:443
CONNECT   Dec 11 21:51:50 [19067]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:50 [19067]: Request (file descriptor 7): CONNECT production.cloudflare.docker.com:443 HTTP/1.1
INFO      Dec 11 21:51:50 [18619]: opensock: getaddrinfo returned for production.cloudflare.docker.com:443
INFO      Dec 11 21:51:50 [19067]: No upstream proxy for production.cloudflare.docker.com
INFO      Dec 11 21:51:50 [19067]: opensock: opening connection to production.cloudflare.docker.com:443
INFO      Dec 11 21:51:50 [19067]: opensock: getaddrinfo returned for production.cloudflare.docker.com:443
CONNECT   Dec 11 21:51:50 [18619]: Established connection to host "production.cloudflare.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:50 [18619]: Not sending client headers to remote machine
CONNECT   Dec 11 21:51:50 [19067]: Established connection to host "production.cloudflare.docker.com" using file descriptor 8.
INFO      Dec 11 21:51:50 [19067]: Not sending client headers to remote machine
INFO      Dec 11 21:51:50 [19056]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:50 [19070]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:50 [19066]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:50 [18619]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:50 [19059]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:50 [19067]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:52 [18625]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:52 [19062]: Closed connection between local client (fd:7) and remote client (fd:8)
CONNECT   Dec 11 21:51:53 [19062]: Connect (file descriptor 7): ec2-44-242-151-120.us-west-2.compute.amazonaws.com [44.242.151.120]
CONNECT   Dec 11 21:51:53 [19062]: Request (file descriptor 7): CONNECT ecs.us-west-2.amazonaws.com:443 HTTP/1.1
INFO      Dec 11 21:51:53 [19062]: No upstream proxy for ecs.us-west-2.amazonaws.com
INFO      Dec 11 21:51:53 [19062]: opensock: opening connection to ecs.us-west-2.amazonaws.com:443
INFO      Dec 11 21:51:53 [19062]: opensock: getaddrinfo returned for ecs.us-west-2.amazonaws.com:443
CONNECT   Dec 11 21:51:53 [19062]: Established connection to host "ecs.us-west-2.amazonaws.com" using file descriptor 8.
INFO      Dec 11 21:51:53 [19062]: Not sending client headers to remote machine
INFO      Dec 11 21:51:53 [19061]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:54 [19058]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:54 [18624]: Closed connection between local client (fd:7) and remote client (fd:8)
INFO      Dec 11 21:51:59 [19062]: Closed connection between local client (fd:7) and remote client (fd:8)

``` 

</details>

- [x] Testing the restart commands
Launched instance without specifying proxy settings. 
After host comes up, change proxy settings to route traffic through my https proxy server:
```bash
apiclient -u /settings -m PATCH -d '{"network": {"https-proxy": "myhttpsserver:9898", "no-proxy":[".local"]}}'
apiclient -u /tx/commit_and_apply -m POST
```

Observed that containerd, kubelet, host-containerd all restarted successfully and pod workloads continued without problems.

Please let me know if there is anything else that needs to be tested.

**Note: The migration for the new settings will fast-follow this PR.**

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,27 @@ These settings can be changed at any time.
 * `settings.updates.version-lock`: Controls the version that will be selected when you issue an update request.  Can be locked to a specific version like `v1.0.0`, or `latest` to take the latest available version.  Defaults to `latest`.
 * `settings.updates.ignore-waves`: Updates are rolled out in waves to reduce the impact of issues.  For testing purposes, you can set this to `true` to ignore those waves and update immediately.
 
+#### Network settings
+
+##### Proxy settings
+
+These settings will configure the proxying behavior of the following services:
+* For all variants:
+    * [containerd.service](packages/containerd/containerd.service)
+    * [host-containerd.service](packages/host-ctr/host-containerd.service)
+* For Kubernetes variants:
+    * [kubelet.service](packages/kubernetes-1.18/kubelet.service)
+* For the ECS variant:
+    * [docker.service](packages/docker-engine/docker.service)
+    * [ecs.service](packages/ecs-agent/ecs.service)
+
+* `settings.network.https-proxy`: The HTTPS proxy server to be used by services listed above.
+* `settings.network.no-proxy`: A list of hosts that are excluded from proxying.
+
+The no-proxy list will automatically include entries for localhost.
+
+If you're running a Kubernetes variant, the no-proxy list will automatically include the Kubernetes API server endpoint and other commonly used Kubernetes DNS suffixes to facilitate intra-cluster networking.
+
 #### Time settings
 
 * `settings.ntp.time-servers`: A list of NTP servers used to set and verify the system time.
--- a/packages/docker-engine/docker.service
+++ b/packages/docker-engine/docker.service
@@ -10,6 +10,7 @@ StartLimitIntervalSec=60s
 
 [Service]
 Type=notify
+EnvironmentFile=/etc/network/proxy.env
 ExecStart=/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock
 ExecReload=/bin/kill -s HUP $MAINPID
 Delegate=yes
--- a/packages/release/proxy-env
+++ b/packages/release/proxy-env
@@ -1,0 +1,6 @@
+{{#if settings.network.https-proxy}}
+HTTPS_PROXY={{settings.network.https-proxy}}
+https_proxy={{settings.network.https-proxy}}
+{{/if}}
+NO_PROXY={{#each settings.network.no-proxy}}{{this}},{{/each}}localhost,127.0.0.1{{#if settings.kubernetes.api-server}},{{host settings.kubernetes.api-server}}{{/if}}{{#if settings.kubernetes.cluster-domain}},.{{settings.kubernetes.cluster-domain}}{{/if}}
+no_proxy={{#each settings.network.no-proxy}}{{this}},{{/each}}localhost,127.0.0.1{{#if settings.kubernetes.api-server}},{{host settings.kubernetes.api-server}}{{/if}}{{#if settings.kubernetes.cluster-domain}},.{{settings.kubernetes.cluster-domain}}{{/if}}
--- a/sources/api/schnauzer/Cargo.toml
+++ b/sources/api/schnauzer/Cargo.toml
@@ -23,6 +23,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
 snafu = "0.6"
 tokio = { version = "0.2", default-features = false, features = ["macros", "rt-threaded"] }
+url = "2.1"
 # When hyper updates to tokio 0.3:
 #tokio = { version = "0.3", default-features = false, features = ["macros", "rt-multi-thread"] }
 
--- a/sources/api/schnauzer/src/helpers.rs
+++ b/sources/api/schnauzer/src/helpers.rs
@@ -8,6 +8,7 @@ use serde_json::value::Value;
 use snafu::{OptionExt, ResultExt};
 use std::borrow::Borrow;
 use std::collections::HashMap;
+use url::Url;
 
 lazy_static! {
     /// A map to tell us which registry to pull ECR images from for a given region.
@@ -132,6 +133,21 @@ mod error {
             template: String,
             source: std::io::Error,
         },
+
+        #[snafu(display(
+            "Expected an absolute URL, got '{}' in template '{}': '{}'",
+            url_str,
+            template,
+            source
+        ))]
+        UrlParse {
+            url_str: String,
+            template: String,
+            source: url::ParseError,
+        },
+
+        #[snafu(display("URL '{}' is missing host component", url_str))]
+        UrlHost { url_str: String },
     }
 
     // Handlebars helpers are required to return a RenderError.
@@ -459,6 +475,40 @@ pub fn ecr_prefix(
         .with_context(|| error::TemplateWrite {
             template: template_name.to_owned(),
         })?;
+
+    Ok(())
+}
+
+/// `host` takes an absolute URL and trims it down and returns its host.
+pub fn host(
+    helper: &Helper<'_, '_>,
+    _: &Handlebars,
+    _: &Context,
+    renderctx: &mut RenderContext<'_, '_>,
+    out: &mut dyn Output,
+) -> Result<(), RenderError> {
+    trace!("Starting host helper");
+    let template_name = template_name(renderctx);
+    check_param_count(helper, template_name, 1)?;
+
+    let url_val = get_param(helper, 0)?;
+    let url_str = url_val
+        .as_str()
+        .with_context(|| error::InvalidTemplateValue {
+            expected: "string",
+            value: url_val.to_owned(),
+            template: template_name.to_owned(),
+        })?;
+    let url = Url::parse(url_str).context(error::UrlParse {
+        url_str,
+        template: template_name,
+    })?;
+    let url_host = url.host_str().context(error::UrlHost { url_str })?;
+
+    // write it to the template
+    out.write(&url_host).with_context(|| error::TemplateWrite {
+        template: template_name.to_owned(),
+    })?;
 
     Ok(())
 }
@@ -821,5 +871,64 @@ mod test_ecr_registry {
         )
         .unwrap();
         assert_eq!(result, EXPECTED_URL_XY_ZTOWN_1);
+    }
+}
+
+#[cfg(test)]
+mod test_host {
+    use super::*;
+    use handlebars::TemplateRenderError;
+    use serde::Serialize;
+    use serde_json::json;
+
+    // A thin wrapper around the handlebars render_template method that includes
+    // setup and registration of helpers
+    fn setup_and_render_template<T>(tmpl: &str, data: &T) -> Result<String, TemplateRenderError>
+    where
+        T: Serialize,
+    {
+        let mut registry = Handlebars::new();
+        registry.register_helper("host", Box::new(host));
+
+        registry.render_template(tmpl, data)
+    }
+
+    #[test]
+    fn not_absolute_url() {
+        assert!(setup_and_render_template(
+            "{{ host url_setting }}",
+            &json!({"url_setting": "example.com"}),
+        )
+        .is_err());
+    }
+
+    #[test]
+    fn https() {
+        let result = setup_and_render_template(
+            "{{ host url_setting }}",
+            &json!({"url_setting": "https://example.example.com/example/example"}),
+        )
+        .unwrap();
+        assert_eq!(result, "example.example.com");
+    }
+
+    #[test]
+    fn http() {
+        let result = setup_and_render_template(
+            "{{ host url_setting }}",
+            &json!({"url_setting": "http://example.com"}),
+        )
+        .unwrap();
+        assert_eq!(result, "example.com");
+    }
+
+    #[test]
+    fn unknown_scheme() {
+        let result = setup_and_render_template(
+            "{{ host url_setting }}",
+            &json!({"url_setting": "foo://example.com"}),
+        )
+        .unwrap();
+        assert_eq!(result, "example.com");
     }
 }
--- a/sources/api/schnauzer/src/lib.rs
+++ b/sources/api/schnauzer/src/lib.rs
@@ -122,6 +122,7 @@ pub fn build_template_registry() -> Result<handlebars::Handlebars<'static>> {
     template_registry.register_helper("join_map", Box::new(helpers::join_map));
     template_registry.register_helper("default", Box::new(helpers::default));
     template_registry.register_helper("ecr-prefix", Box::new(helpers::ecr_prefix));
+    template_registry.register_helper("host", Box::new(helpers::host));
 
     Ok(template_registry)
 }
--- a/sources/models/defaults.toml
+++ b/sources/models/defaults.toml
@@ -22,7 +22,7 @@ template-path = "/usr/share/templates/motd"
 # Container runtime.
 
 [services.containerd]
-configuration-files = ["containerd-config-toml"]
+configuration-files = ["containerd-config-toml", "proxy-env"]
 restart-commands = ["/bin/systemctl try-restart containerd.service"]
 
 [configuration-files.containerd-config-toml]
@@ -32,7 +32,7 @@ template-path = "/usr/share/templates/containerd-config-toml"
 # Host-container runtime
 
 [services.host-containerd]
-configuration-files = []
+configuration-files = ["proxy-env"]
 restart-commands = ["/bin/systemctl try-restart host-containerd.service"]
 
 # Updates.
@@ -82,6 +82,15 @@ restart-commands = ["/usr/bin/host-containers"]
 
 [metadata.settings.host-containers]
 affected-services = ["host-containers"]
+
+# Network
+
+[configuration-files.proxy-env]
+path = "/etc/network/proxy.env"
+template-path = "/usr/share/templates/proxy-env"
+
+[metadata.settings.network]
+affected-services = ["containerd", "host-containerd"]
 
 # NTP
 
--- a/sources/models/src/aws-ecs-1/override-defaults.toml
+++ b/sources/models/src/aws-ecs-1/override-defaults.toml
@@ -5,7 +5,7 @@ template-path = "/usr/share/templates/containerd-config-toml_aws-ecs-1"
 # Docker
 [services.docker]
 restart-commands = ["/bin/systemctl try-restart docker.service"]
-configuration-files = []
+configuration-files = ["proxy-env"]
 
 # ECS
 [services.ecs]
@@ -23,3 +23,7 @@ affected-services = ["ecs"]
 allow-privileged-containers = false
 logging-drivers = ["json-file", "awslogs", "none"]
 loglevel = "info"
+
+# Network
+[metadata.settings.network]
+affected-services = ["containerd", "docker", "ecs", "host-containerd"]
--- a/sources/models/src/aws-k8s-1.15/override-defaults.toml
+++ b/sources/models/src/aws-k8s-1.15/override-defaults.toml
@@ -5,7 +5,7 @@ template-path = "/usr/share/templates/containerd-config-toml_aws-k8s"
 # Kubernetes.
 
 [services.kubernetes]
-configuration-files = ["kubelet-env", "kubelet-config", "kubelet-kubeconfig", "kubernetes-ca-crt"]
+configuration-files = ["kubelet-env", "kubelet-config", "kubelet-kubeconfig", "kubernetes-ca-crt", "proxy-env"]
 restart-commands = ["/bin/systemctl try-restart kubelet.service"]
 
 [configuration-files.kubelet-env]
@@ -36,3 +36,8 @@ affected-services = ["kubernetes", "containerd"]
 
 [settings.kubernetes]
 cluster-domain = "cluster.local"
+
+
+# Network
+[metadata.settings.network]
+affected-services = ["containerd", "kubernetes", "host-containerd"]
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -140,6 +140,14 @@ struct ContainerImage {
     superpowered: bool,
 }
 
+// Network settings. These settings will affect host service components' network behavior
+#[model]
+struct NetworkSettings {
+    https_proxy: Url,
+    // We allow some flexibility in NO_PROXY values because different services support different formats.
+    no_proxy: Vec<SingleLineString>,
+}
+
 // NTP settings
 #[model]
 struct NtpSettings {
--- a/sources/models/src/modeled_types/shared.rs
+++ b/sources/models/src/modeled_types/shared.rs
@@ -240,6 +240,9 @@ mod test_url {
             "http://localhost",
             "localhost/path",
             "localhost",
+            "localhost:8080",
+            ".internal",
+            ".cluster.local"
         ] {
             Url::try_from(*ok).unwrap();
         }
--- a/packages/containerd/containerd.service
+++ b/packages/containerd/containerd.service
@@ -5,6 +5,7 @@ After=network-online.target configured.target
 Wants=network-online.target configured.target
 
 [Service]
+EnvironmentFile=/etc/network/proxy.env
 ExecStart=/usr/bin/containerd
 Type=notify
 Delegate=yes
--- a/packages/ecs-agent/ecs.service
+++ b/packages/ecs-agent/ecs.service
@@ -11,6 +11,7 @@ Restart=on-failure
 RestartPreventExitStatus=5
 RestartSec=1s
 EnvironmentFile=-/etc/ecs/ecs.config
+EnvironmentFile=/etc/network/proxy.env
 Environment=ECS_CHECKPOINT=true
 ExecStartPre=/sbin/iptables -t nat -A PREROUTING -d 169.254.170.2/32 \
  -p tcp -m tcp --dport 80 -j DNAT --to-destination 127.0.0.1:51679
--- a/packages/host-ctr/host-containerd.service
+++ b/packages/host-ctr/host-containerd.service
@@ -5,6 +5,7 @@ After=network-online.target configured.target
 Wants=network-online.target configured.target
 
 [Service]
+EnvironmentFile=/etc/network/proxy.env
 ExecStart=/usr/bin/containerd --config /etc/host-containerd/config.toml
 Type=notify
 Delegate=yes
--- a/packages/kubernetes-1.15/kubelet.service
+++ b/packages/kubernetes-1.15/kubelet.service
@@ -7,6 +7,7 @@ BindsTo=containerd.service
 
 [Service]
 Type=notify
+EnvironmentFile=/etc/network/proxy.env
 EnvironmentFile=/etc/kubernetes/kubelet/env
 ExecStartPre=/sbin/iptables -P FORWARD ACCEPT
 # Pull the pause container image before starting `kubelet` so `containerd/cri` wouldn't have to
--- a/packages/kubernetes-1.16/kubelet.service
+++ b/packages/kubernetes-1.16/kubelet.service
@@ -7,6 +7,7 @@ BindsTo=containerd.service
 
 [Service]
 Type=notify
+EnvironmentFile=/etc/network/proxy.env
 EnvironmentFile=/etc/kubernetes/kubelet/env
 ExecStartPre=/sbin/iptables -P FORWARD ACCEPT
 # Pull the pause container image before starting `kubelet` so `containerd/cri` wouldn't have to
--- a/packages/kubernetes-1.17/kubelet.service
+++ b/packages/kubernetes-1.17/kubelet.service
@@ -7,6 +7,7 @@ BindsTo=containerd.service
 
 [Service]
 Type=notify
+EnvironmentFile=/etc/network/proxy.env
 EnvironmentFile=/etc/kubernetes/kubelet/env
 ExecStartPre=/sbin/iptables -P FORWARD ACCEPT
 # Pull the pause container image before starting `kubelet` so `containerd/cri` wouldn't have to
--- a/packages/kubernetes-1.18/kubelet.service
+++ b/packages/kubernetes-1.18/kubelet.service
@@ -7,6 +7,7 @@ BindsTo=containerd.service
 
 [Service]
 Type=notify
+EnvironmentFile=/etc/network/proxy.env
 EnvironmentFile=/etc/kubernetes/kubelet/env
 ExecStartPre=/sbin/iptables -P FORWARD ACCEPT
 # Pull the pause container image before starting `kubelet` so `containerd/cri` wouldn't have to
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -13,6 +13,7 @@ Source98: release-systemd-system.conf
 Source99: release-tmpfiles.conf
 
 Source200: motd.template
+Source201: proxy-env
 
 Source1000: eth0.xml
 Source1002: configured.target
@@ -122,6 +123,7 @@ install -p -m 0644 ${LICENSEPATH}.mount %{buildroot}%{_cross_unitdir}
 
 install -d %{buildroot}%{_cross_templatedir}
 install -p -m 0644 %{S:200} %{buildroot}%{_cross_templatedir}/motd
+install -p -m 0644 %{S:201} %{buildroot}%{_cross_templatedir}/proxy-env
 
 %files
 %{_cross_factorydir}%{_cross_sysconfdir}/hosts
@@ -142,5 +144,6 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_templatedir}/motd
 %{_cross_unitdir}/var-lib-bottlerocket.mount
 %dir %{_cross_templatedir}
 %{_cross_templatedir}/motd
+%{_cross_templatedir}/proxy-env
 
 %changelog
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -976,6 +976,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
+name = "form_urlencoded"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ece68d15c92e84fa4f19d3780f1294e5ca82a78a6d515f1efaabcc144688be00"
+dependencies = [
+ "matches",
+ "percent-encoding",
+]
+
+[[package]]
 name = "fs2"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2386,6 +2396,7 @@ dependencies = [
  "serde_json",
  "snafu",
  "tokio",
+ "url",
 ]
 
 [[package]]
@@ -3297,10 +3308,11 @@ dependencies = [
 
 [[package]]
 name = "url"
-version = "2.1.1"
+version = "2.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "829d4a8476c35c9bf0bbce5a3b23f4106f79728039b726d292bb93bc106787cb"
+checksum = "5909f2b0817350449ed73e8bcd81c8c3c8d9a7a5d8acba4b27db277f1868976e"
 dependencies = [
+ "form_urlencoded",
  "idna",
  "matches",
  "percent-encoding",
--- a/sources/models/src/aws-dev/mod.rs
+++ b/sources/models/src/aws-dev/mod.rs
@@ -3,7 +3,9 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 use crate::modeled_types::Identifier;
-use crate::{AwsSettings, ContainerImage, KernelSettings, NtpSettings, UpdatesSettings};
+use crate::{
+    AwsSettings, ContainerImage, KernelSettings, NetworkSettings, NtpSettings, UpdatesSettings,
+};
 
 // Note: we have to use 'rename' here because the top-level Settings structure is the only one
 // that uses its name in serialization; internal structures use the field name that points to it
@@ -13,6 +15,7 @@ struct Settings {
     updates: UpdatesSettings,
     host_containers: HashMap<Identifier, ContainerImage>,
     ntp: NtpSettings,
+    network: NetworkSettings,
     kernel: KernelSettings,
     aws: AwsSettings,
 }
--- a/sources/models/src/aws-dev/override-defaults.toml
+++ b/sources/models/src/aws-dev/override-defaults.toml
@@ -5,4 +5,8 @@ template-path = "/usr/share/templates/containerd-config-toml_aws-dev"
 # Docker
 [services.docker]
 restart-commands = ["/bin/systemctl try-restart docker.service"]
-configuration-files = []
+configuration-files = ["proxy-env"]
+
+# Network
+[metadata.settings.network]
+affected-services = ["containerd", "docker", "host-containerd"]
--- a/sources/models/src/aws-ecs-1/mod.rs
+++ b/sources/models/src/aws-ecs-1/mod.rs
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 
 use crate::modeled_types::Identifier;
 use crate::{
-    AwsSettings, ContainerImage, ECSSettings, KernelSettings, NtpSettings, UpdatesSettings,
+    AwsSettings, ContainerImage, ECSSettings, KernelSettings, NetworkSettings, NtpSettings, UpdatesSettings,
 };
 
 // Note: we have to use 'rename' here because the top-level Settings structure is the only one
@@ -15,6 +15,7 @@ struct Settings {
     updates: UpdatesSettings,
     host_containers: HashMap<Identifier, ContainerImage>,
     ntp: NtpSettings,
+    network: NetworkSettings,
     kernel: KernelSettings,
     aws: AwsSettings,
     ecs: ECSSettings,
--- a/sources/models/src/aws-k8s-1.15/mod.rs
+++ b/sources/models/src/aws-k8s-1.15/mod.rs
@@ -4,7 +4,8 @@ use std::collections::HashMap;
 
 use crate::modeled_types::Identifier;
 use crate::{
-    AwsSettings, ContainerImage, KernelSettings, KubernetesSettings, NtpSettings, UpdatesSettings,
+    AwsSettings, ContainerImage, KernelSettings, KubernetesSettings, NetworkSettings, NtpSettings,
+    UpdatesSettings,
 };
 
 // Note: we have to use 'rename' here because the top-level Settings structure is the only one
@@ -16,6 +17,7 @@ struct Settings {
     updates: UpdatesSettings,
     host_containers: HashMap<Identifier, ContainerImage>,
     ntp: NtpSettings,
+    network: NetworkSettings,
     kernel: KernelSettings,
     aws: AwsSettings,
 }


packages/ecs-agent/ecs.config
```diff
@@ -1,2 +1,6 @@
 ECS_LOGFILE=/var/log/ecs/ecs-agent.log
 ECS_LOGLEVEL="{{settings.ecs.loglevel}}"
+{{#if settings.network.http-proxy}}
+HTTP_PROXY={{settings.network.http-proxy}}
+{{/if}}
+NO_PROXY={{#each settings.network.no-proxy}}{{#if @index}},{{/if}}{{this}}{{/each}}{{#if settings.network.no-proxy}},{{/if}}169.254.169.254,169.254.170.2,/var/run/docker.sock
```
webern: I suspect we will want a helper for this. When I used the built-in handlebars functionality to create a comma-delimited and quoted list, it was observed that a helper would be easier to understand.

sources/models/defaults.toml
```diff
@@ -77,6 +77,15 @@ restart-commands = ["/usr/bin/host-containers"]
...
 affected-services = ["host-containers"]
 
+# Network
+
+[configuration-files.proxy-env]
+path = "/etc/shared/services/network/proxy/env"
+template-path = "/usr/share/templates/proxy-env"
+
+[metadata.settings.network]
+affected-services = ["containerd", "ecs", "kubernetes", "host-containerd"]
```
webern: This list strikes me as odd since ecs and kubernetes are mutually exclusive. I guess it doesn't matter since since, presumably, the presence of non-existent services in the list is harmless.

sources/api/schnauzer/src/lib.rs
```diff
@@ -122,6 +122,7 @@ pub fn build_template_registry() -> Result<handlebars::Handlebars<'static>> {
     template_registry.register_helper("join_map", Box::new(helpers::join_map));
     template_registry.register_helper("default", Box::new(helpers::default));
     template_registry.register_helper("ecr-prefix", Box::new(helpers::ecr_prefix));
+    template_registry.register_helper("host-and-port", Box::new(helpers::host_and_port));
```
webern: My `ecr-prefix` unfortunately failed to observe the naming convention started by `join_map`, and the long standing metricdog PR will continue with the underscore convention instead of snake case.

```suggestion
    template_registry.register_helper("host_and_port", Box::new(helpers::host_and_port));
```

sources/models/src/aws-k8s-1.15/override-defaults.toml
```diff
@@ -5,7 +5,7 @@ template-path = "/usr/share/templates/containerd-config-toml_aws-k8s"
 # Kubernetes.
 
 [services.kubernetes]
-configuration-files = ["kubelet-env", "kubelet-config", "kubelet-kubeconfig", "kubernetes-ca-crt"]
+configuration-files = ["kubelet-env", "kubelet-config", "kubelet-kubeconfig", "kubernetes-ca-crt", "proxy-env"]
```
webern, line 8: Does the configuration file need to be added to aws-dev?

sources/models/src/lib.rs
```diff
@@ -140,6 +140,14 @@ struct ContainerImage {
     superpowered: bool,
 }
 
+// Network settings. These settings will affect host service components' network behavior
+#[model]
+struct NetworkSettings {
+    http_proxy: Url,
+    https_proxy: Url,
```
webern: These don't need to be optional?

packages/release/proxy-env
```diff
@@ -0,0 +1,6 @@
+{{#if settings.network.http-proxy}}HTTP_PROXY={{settings.network.http-proxy}}
+http_proxy={{settings.network.http-proxy}}{{/if}}
+{{#if settings.network.https-proxy}}HTTPS_PROXY={{settings.network.https-proxy}}
+https_proxy={{settings.network.https-proxy}}{{/if}}
```
tjkirch:
```suggestion
{{#if settings.network.http-proxy}}
HTTP_PROXY={{settings.network.http-proxy}}
http_proxy={{settings.network.http-proxy}}
{{/if}}
{{#if settings.network.https-proxy}}
HTTPS_PROXY={{settings.network.https-proxy}}
https_proxy={{settings.network.https-proxy}}
{{/if}}
```

sources/models/src/lib.rs
etungsten: They don't. With how the template is written right now, `HTTP_PROXY` and `HTTPS_PROXY` are only rendered if their corresponding settings are set.

packages/release/proxy-env
```diff
@@ -0,0 +1,6 @@
...
+NO_PROXY={{#each settings.network.no-proxy}}{{#if @index}},{{/if}}{{this}}{{/each}}{{#if settings.kubernetes.api-server}}{{#if settings.network.no-proxy}},{{/if}}{{host-and-port settings.kubernetes.api-server}},.svc,.default,.local,.cluster.local{{/if}}
```
tjkirch: I don't think we need to care about port for no-proxying the api-server.  Can we simplify and have a `host` function instead?

packages/ecs-agent/ecs.config
tjkirch: Is this our proper docker socket path?

tjkirch: Separately - I think handlebars has `@first` and `@last` in addition to `@index`, so you might be able to clarify why we're checking `@index` by using `@first` instead, and instead of having a separate `#if settings.network.no-proxy`, use `@last`?  (same in proxy-env)

sources/api/schnauzer/Cargo.toml
```diff
@@ -23,6 +23,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
 snafu = "0.6"
 tokio = { version = "0.2", default-features = false, features = ["macros", "rt-threaded"] }
+url = "2.2.0"
```
tjkirch: nit: looks like 2.2 *just* came out.  Can we stick with 2.1 to match other packages, until we do our next bump?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -132,6 +133,21 @@ mod error {
             template: String,
             source: std::io::Error,
         },
+
+        #[snafu(display(
+            "Expected an absolute URL, got '{}' in template '{}': '{}'",
```
tjkirch, line 138: The proxy variables typically allow just a hostname, and don't require a full URL, no?  I'm wondering if we can just use url::Host to validate.  It accepts IPs.  (We use it for DNSDomain, and explicitly reject IPs.)

packages/ecs-agent/ecs.config
etungsten: `/var/run/docker.sock` should also just be `/run/docker.sock` because that's where the docker socket is in Bottlerocket

sources/models/defaults.toml
```diff
@@ -77,6 +77,15 @@ restart-commands = ["/usr/bin/host-containers"]
 [metadata.settings.host-containers]
 affected-services = ["host-containers"]
 
+# Network
+
+[configuration-files.proxy-env]
+path = "/etc/shared/services/network/proxy/env"
```
tjkirch: This seems really deep.  Can we simplify to `/etc/network/proxy.env` or something?

sources/models/src/aws-ecs-1/override-defaults.toml
```diff
@@ -5,7 +5,7 @@ template-path = "/usr/share/templates/containerd-config-toml_aws-ecs-1"
 # ECS
 [services.ecs]
 restart-commands = ["/usr/bin/ecs-settings-applier", "/bin/systemctl try-reload-or-restart ecs.service"]
-configuration-files = ["ecs-config"]
+configuration-files = ["ecs-config", "proxy-env"]
```
tjkirch: I'm not sure I understand tying the proxy-env configuration file to the ecs and kubernetes services.  It's more of a system-wide configuration file.  Can we have a new service defined in defaults.toml instead?  (That would handle aws-dev as well, which I'm not sure is handled now?)

etungsten:
I should look into supporting CIDR notations for the `no-proxy` setting seeing as docker and containerd support it.

sources/models/src/lib.rs
```diff
@@ -140,6 +140,14 @@ struct ContainerImage {
...
+    no_proxy: Vec<Url>,
```
tjkirch: You had mentioned making this support CIDR notation, but it doesn't seem like there's consistent support across the services we care about for CIDRs/wildcards/suffixes/etc.  Perhaps this should just be `Vec<SingleLineString>`?

README.md
```diff
@@ -344,6 +344,30 @@ These settings can be changed at any time.
...
+* For ECS variants:
+    * [docker.service](packages/docker-engine/docker.service)
+    * [ecs.service](packages/ecs-agent/ecs.service)
+
+* `settings.network.http-proxy`: The HTTP proxy server to be used by services listed above.
+* `settings.network.https-proxy`: The HTTPS proxy server to be used by services listed above.
+* `settings.network.no-proxy`: A list of hosts that are excluded from proxying.
+Here are a few values you normally want to set for this setting when setting up proxies:
+    * "localhost" and "127.0.0.1" for loopback
+    * Instance metadata endpoint (e.g. "169.254.169.254" for Amazon EC2 Instance metadata data service)
```
tjkirch: Let's make these defaults, since there are almost no cases in which someone *wouldn't* want these, and if they don't, they can set no-proxy.

packages/ecs-agent/ecs.config
samuelkarp: `/var/run` is a symlink to `/run`.  The ECS agent [dials `/var/run/docker.sock`](https://github.com/aws/amazon-ecs-agent/blob/1489adfaec05c559ea4e1490774d1f86261cf936/agent/config/config_unix.go#L48), so this value is correct.

etungsten: Nevermind, `/var/run` is symlinked to `/run`, so this is correct as is

sources/models/src/lib.rs
tjkirch: (The `#[model]` annotation makes all fields `Option`al.)

packages/ecs-agent/ecs.config
etungsten: See https://github.com/bottlerocket-os/bottlerocket/pull/1204#discussion_r521560554.

sources/api/schnauzer/src/helpers.rs
etungsten: My intention for this helper was to trim `settings.kubernetes.api-server` which should always be a fully qualified URL (otherwise kubelet complains).

I'll go ahead and make it more flexible.

packages/ecs-agent/ecs.config
etungsten: Oh I didn't know `@first` and `@last` are a thing.. that's such a weird construct. Thanks!

etungsten: For `proxy-env` it's slightly different because I cannot guarantee I'll be adding additional elements to the list. The additional elements are added on the condition that `settings.kubernetes.api-server` is set (meaning it's running kubelet) since only kubernetes needs them.

etungsten: I thought about this for a while and I think implementing a custom helper to do this would probably bit more work than it's worth. To fulfil my desired use-case, the helper would also need to understand how to concatenate both strings and collections.

If the helper is just for generating comma-delimited list out of an array. Then essentially the only part we're replacing is `{{#each settings.network.no-proxy}}{{#if @index}},{{/if}}{{this}}{{/each}}`, which on its own doesn't seem too bad in terms of readability.

I'm willing to be convinced though. Maybe if there are more instances of us doing this, I would probably think a custom helper is necessary.

sources/api/schnauzer/src/helpers.rs
etungsten: After some tinkering and thinking, I think I'm going to keep it as is. This helper is really just meant for trimming the URI scheme and paths off of absolute URLs stored in settings (e.g. kubernetes.api-server) so they can be included in the `no-proxy` setting.

packages/ecs-agent/ecs.config
```diff
@@ -1,2 +1,6 @@
 ECS_LOGFILE=/var/log/ecs/ecs-agent.log
 ECS_LOGLEVEL="{{settings.ecs.loglevel}}"
+{{#if settings.network.http-proxy}}
+HTTP_PROXY={{settings.network.http-proxy}}
+{{/if}}
+NO_PROXY={{#each settings.network.no-proxy}}{{#if @first}},{{/if}}{{this}}{{#if @last}},{{/if}}{{/each}}169.254.169.254,169.254.170.2,/run/docker.sock
```
tjkirch: My `@first` suggestion was bad, we want exactly the opposite, so `@index` was right.  Sorry.

packages/docker-engine/docker.service
```diff
@@ -10,6 +10,7 @@ StartLimitIntervalSec=60s
 
 [Service]
 Type=notify
+EnvironmentFile=/etc/network/proxy/env
```
tjkirch:
```suggestion
EnvironmentFile=/etc/network/proxy.env
```
I don't think we expect to have other proxy files, no need for more hierarchy.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -463,6 +479,43 @@ pub fn ecr_prefix(
...
 
+/// `host` takes an absolute URL and trims it down and returns its host.
+pub fn host(
+    helper: &Helper<'_, '_>,
+    _: &Handlebars,
+    _: &Context,
+    renderctx: &mut RenderContext<'_, '_>,
+    out: &mut dyn Output,
+) -> Result<(), RenderError> {
+    trace!("Starting host and port helper");
```
tjkirch:
```suggestion
    trace!("Starting host helper");
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -463,6 +479,43 @@ pub fn ecr_prefix(
...
+    let url_val = get_param(helper, 0)?;
+    let url_str = url_val
+        .as_str()
+        .with_context(|| error::InvalidTemplateValue {
+            expected: "string",
+            value: url_val.to_owned(),
+            template: template_name.to_owned(),
+        })?;
+    let url = Url::parse(url_str).with_context(|| error::UrlParse {
+        value: url_val.to_owned(),
```
tjkirch:
```suggestion
        value: url_str,
```
No need to clone the larger JSON object, I think.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -463,6 +479,43 @@ pub fn ecr_prefix(
...
+        template: template_name,
+    })?;
+    let url_host = url.host_str().context(error::UrlHost {
+        url: url.to_owned(),
```
tjkirch: I'd recommend `with_context` if we *really* have to clone the url, but... do we?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -823,3 +876,62 @@ mod test_ecr_registry {
...
+        T: Serialize,
+    {
+        let mut registry = Handlebars::new();
+        registry.register_helper("host", Box::new(host));
+
+        registry.render_template(tmpl, data)
+    }
+
+    #[test]
+    fn not_absolute_url() {
```
tjkirch, line 897: It seems really weird to me to error when an input isn't an absolute URL when we're trimming it down to just a host anyway.  Maybe we could do like the `Url` modeled type and temporarily add a scheme if it fails to parse?

etungsten:
Push above and below addresses @tjkirch and @webern 's comments.

Tested changes in EKS and they work as described in the PR description still.

sources/models/defaults.toml
```diff
@@ -77,6 +77,25 @@ restart-commands = ["/usr/bin/host-containers"]
...
+[metadata.settings.network]
+affected-services = ["containerd", "ecs", "kubernetes", "host-containerd"]
+
+# Network
+
+[services.network]
+configuration-files = ["proxy-env"]
+restart-commands = []
+
+[settings.network]
```
tjkirch: nit: can we put the settings before their metadata?

sources/models/src/lib.rs
```diff
@@ -140,6 +140,15 @@ struct ContainerImage {
     superpowered: bool,
 }
 
+// Network settings. These settings will affect host service components' network behavior
+#[model]
+struct NetworkSettings {
+    http_proxy: Url,
+    https_proxy: Url,
+    // We're allowing more flexibility for setting NO_PROXY values
```
tjkirch:
```suggestion
    // We allow some flexibility in NO_PROXY values because different services support different formats.
```

packages/ecs-agent/ecs.config
```diff
@@ -1,2 +1,6 @@
 ECS_LOGFILE=/var/log/ecs/ecs-agent.log
 ECS_LOGLEVEL="{{settings.ecs.loglevel}}"
+{{#if settings.network.http-proxy}}
+HTTP_PROXY={{settings.network.http-proxy}}
+{{/if}}
+NO_PROXY={{#each settings.network.no-proxy}}{{#if @index}},{{/if}}{{this}}{{#if @last}},{{/if}}{{/each}}169.254.169.254,169.254.170.2,/var/run/docker.sock
```
tjkirch:
```suggestion
NO_PROXY={{#each settings.network.no-proxy}}{{this}},{{/each}}169.254.169.254,169.254.170.2,/var/run/docker.sock
```
Wouldn't this wind up being the same?  Can simplify because we know there are values after.

README.md
```diff
@@ -344,6 +344,28 @@ These settings can be changed at any time.
...
+* For Kubernetes variants:
+    * [kubelet.service](packages/kubernetes-1.18/kubelet.service)
+* For ECS variants:
+    * [docker.service](packages/docker-engine/docker.service)
+    * [ecs.service](packages/ecs-agent/ecs.service)
+
+* `settings.network.http-proxy`: The HTTP proxy server to be used by services listed above.
+* `settings.network.https-proxy`: The HTTPS proxy server to be used by services listed above.
+* `settings.network.no-proxy`: A list of hosts that are excluded from proxying.
+By default, this is set to `["localhost", "127.0.0.1", "169.254.169.254" (Instance Metadata Service endpoint)]`.
```
tjkirch: I think having these in `defaults.toml` might be overly complicating things, actually.  Could there be a case where someone wants to proxy these?  (We shouldn't encourage proxying IMDS, and proxying localhost seems... breaky.)

Can we just put these in the EnvironmentFile?  Then the user doesn't have to repeat all these to set their own no-proxy values, which would be a bad experience.  And we can simplify the README here a bit.  (And the handlebars logic in proxy-env, similar to my suggestion in ecs.config.)

sources/api/schnauzer/src/helpers.rs
etungsten: Apparently anything can be a valid URL as long as it has a scheme and the string does not contain any special characters, even if that scheme is something random. 

Going to opt for more stricter checking since otherwise anything can qualify as an URL host.

sources/api/schnauzer/src/helpers.rs
etungsten: I have to clone here because `.host_str()` borrows `url`

tjkirch: Normally Snafu handles ownership/cloning of referenced objects, though.  You can pass `&str` in `context` with `String` in the error type, for the most common example.  Is that not working here?  (I'm not sure of the exact type bound required.)

etungsten: When I try that I get the following, note that `url` is a `url::Url`:
```
error[E0505]: cannot move out of `url` because it is borrowed
   --> api/schnauzer/src/helpers.rs:506:48
    |
506 |     let url_host = url.host_str().with_context(|| error::UrlHost {
    |                    ---            ------------ ^^ move out of `url` occurs here
    |                    |              |
    |                    |              borrow later used by call
    |                    borrow of `url` occurs here
507 |         url: url,
    |              --- move occurs due to use in closure

```

etungsten: I guess I can just make it take a String, and just pass `url_str`

etungsten:
Push above and below addresses @tjkirch 's comments. 
- Removed defaults for `no-proxy`. 
- Added localhost, 127.0.0.1 and imds endpoint directly to the `no_proxy` list in the proxy-env file template
- Other minor tweaks and fixes

etungsten:
Updated PR description with additional ECS testing info.

sources/models/src/aws-k8s-1.15/override-defaults.toml
etungsten: I removed `"proxy_env"` from all `services.` metadata and created a separate service metadata table to house it

sources/api/schnauzer/src/lib.rs
etungsten: Now the helper is just called `host`

README.md
```diff
@@ -344,6 +344,31 @@ These settings can be changed at any time.
...
+    * [host-containerd.service](packages/host-ctr/host-containerd.service)
+* For Kubernetes variants:
+    * [kubelet.service](packages/kubernetes-1.18/kubelet.service)
+* For ECS variants:
+    * [docker.service](packages/docker-engine/docker.service)
+    * [ecs.service](packages/ecs-agent/ecs.service)
+
+* `settings.network.http-proxy`: The HTTP proxy server to be used by services listed above.
+* `settings.network.https-proxy`: The HTTPS proxy server to be used by services listed above.
+* `settings.network.no-proxy`: A list of hosts that are excluded from proxying.
```
tjkirch, line 362:
```suggestion
* `settings.network.no-proxy`: A comma-separated list of hosts that are excluded from proxying.
```

README.md
```diff
@@ -344,6 +344,31 @@ These settings can be changed at any time.
...
+
+`NO_PROXY` and `no_proxy` will always include `localhost`, `127.0.0.1`, and `169.254.169.254`
+(Instance Metadata Service endpoint)
+
+If you are running Kubernetes variants, `NO_PROXY` and `no_proxy` will be appended with the Kubernetes API server endpoint
+and other commonly used Kubernetes DNS suffixes to facilitate intra-cluster networking.
+
+If you are running the Amazon ECS variants, `NO_PROXY` will be appended with the IMDS endpoint, the Task Metadata Endpoint
+ and the docker socket path.
```
```suggestion
If you're running the Amazon ECS variant, the no-proxy list will automatically include the task metadata endpoint and the Docker socket path.
```

README.md
```diff
@@ -344,6 +344,31 @@ These settings can be changed at any time.
...
+* For ECS variants:
+    * [docker.service](packages/docker-engine/docker.service)
+    * [ecs.service](packages/ecs-agent/ecs.service)
+
+* `settings.network.http-proxy`: The HTTP proxy server to be used by services listed above.
+* `settings.network.https-proxy`: The HTTPS proxy server to be used by services listed above.
+* `settings.network.no-proxy`: A list of hosts that are excluded from proxying.
+
+`NO_PROXY` and `no_proxy` will always include `localhost`, `127.0.0.1`, and `169.254.169.254`
+(Instance Metadata Service endpoint)
```
```suggestion
The no-proxy list will automatically include entries for localhost and the instance metadata service.
```

README.md
```diff
@@ -344,6 +344,31 @@ These settings can be changed at any time.
...
+
+If you are running Kubernetes variants, `NO_PROXY` and `no_proxy` will be appended with the Kubernetes API server endpoint
+and other commonly used Kubernetes DNS suffixes to facilitate intra-cluster networking.
```
```suggestion
If you're running a Kubernetes variant, the no-proxy list will automatically include the Kubernetes API server endpoint and other commonly used Kubernetes DNS suffixes to facilitate intra-cluster networking.
```

README.md
tjkirch: We already mentioned instance metadata.  It raises another question, though - should we include localhost in the ECS list?  We (rightfully) say it'll be on the list, but it's not in ecs.config.

Maybe a better question - does ecs-agent respect no-proxy environment variables?  Can we have ecs.service source the environment file instead of setting up proxying in ecs.config?  (It'd set _additional_ variables, perhaps, but that shouldn't matter.)

If we can't do that, maybe we should add localhost to ecs.config, for consistency.

README.md
```diff
@@ -344,6 +344,31 @@ These settings can be changed at any time.
 * `settings.updates.version-lock`: Controls the version that will be selected when you issue an update request.  Can be locked to a specific version like `v1.0.0`, or `latest` to take the latest available version.  Defaults to `latest`.
 * `settings.updates.ignore-waves`: Updates are rolled out in waves to reduce the impact of issues.  For testing purposes, you can set this to `true` to ignore those waves and update immediately.
 
+#### Network settings
```
tjkirch, line 347:
```suggestion
#### Network settings

##### Proxy settings
```

README.md
samuelkarp: > Maybe a better question - does ecs-agent respect no-proxy environment variables?

Yes

> Can we have ecs.service source the environment file instead of setting up proxying in ecs.config?

Yes, the `ecs.config` file is just sourced as an environment file in the systemd unit anyway.

README.md
etungsten: Specifying "comma-separated" gives the impression that this setting is a single line string when in fact it should be specified with a toml array of strings. I think "A list" is appropriate here. 

(The description for `settings.ecs.logging-drivers` also says "list")

Now if I think about it. Is it a bad idea to make `no-proxy` a single line string setting and say it should be a comma-separated list?... I'm somewhat inclined to say we shouldn't.

README.md
etungsten: Is there any harm in adding the task metadata endpoint to the no-proxy list even when it's a Kubernetes variant or any variant other than ECS?

Letting `ecs.config` set the proxy environment variables has the nice property of separating out the ECS specific no-proxy entries to just the ECS agent service.

README.md
tjkirch: Ah, right, sorry.  List!  We should definitely represent a list as a list, not a weird string.

README.md
tjkirch: There could possibly be harm if someone is using that link-local address for their own purposes, but you'd have to go out of your way.  I think it'd be reasonable to have everything in one list.

etungsten:
Mostly documentation changes.
Added a few more default no_proxy entries to proxy.env for ECS.
Instead of setting up the proxy env vars in `ecs.config`, ecs-agent.service will just source `proxy.env`.

packages/release/proxy-env
```diff
@@ -0,0 +1,10 @@
+{{#if settings.network.http-proxy}}
+HTTP_PROXY={{settings.network.http-proxy}}
+http_proxy={{settings.network.http-proxy}}
+{{/if}}
+{{#if settings.network.https-proxy}}
+HTTPS_PROXY={{settings.network.https-proxy}}
+https_proxy={{settings.network.https-proxy}}
+{{/if}}
+NO_PROXY={{#each settings.network.no-proxy}}{{this}},{{/each}}localhost,127.0.0.1,169.254.169.254,169.254.170.2,/var/run/docker.sock{{#if settings.kubernetes.api-server}},{{host settings.kubernetes.api-server}},.svc,.default,.local,.cluster.local{{/if}}
```
tjkirch: My only remaining concern is that we list /var/run/docker.sock but not /run/docker.sock, even though our docker.socket unit lists /run/docker.sock.  Is there a chance we wouldn't proxy requests because we're using one instead of the other, even though they're symlinked?  Do we need to list both?

README.md
```diff
@@ -344,6 +344,28 @@ These settings can be changed at any time.
...
+    * [kubelet.service](packages/kubernetes-1.18/kubelet.service)
+* For ECS variants:
+    * [docker.service](packages/docker-engine/docker.service)
+    * [ecs.service](packages/ecs-agent/ecs.service)
+
+* `settings.network.http-proxy`: The HTTP proxy server to be used by services listed above.
+* `settings.network.https-proxy`: The HTTPS proxy server to be used by services listed above.
+* `settings.network.no-proxy`: A list of hosts that are excluded from proxying.
+
+The no-proxy list will automatically include entries for localhost, the instance metadata service, the Amazon ECS task metadata endpoint and the Docker socket path.
```
```suggestion
The no-proxy list will automatically include entries for localhost, the instance metadata service, the Amazon ECS task metadata endpoint, and the Docker socket path.
```
(this is my hill and I will die on it)

tjkirch (approved):

packages/release/proxy-env
etungsten: All I know is ECS agent explicitly dials `/var/run/docker.sock`. As for how that gets resolved is unclear to me. I do know from testing that the ECS variant currently works with what's currently there.

tjkirch: @samuelkarp any thoughts?

sources/models/defaults.toml
bcressey: I suppose I would expect to override this within the models for `aws-k8s-*` and `aws-ecs`, as you did for `settings.metrics`.

packages/release/proxy-env
samuelkarp: I haven't seen any problems with using `/var/run/docker.sock` on other platforms with symlinks (like Amazon Linux 2).

We should have test coverage for the ECS agent in the agent's test suite, so I'm comfortable relying on that for validation.

README.md
```diff
@@ -344,6 +344,28 @@ These settings can be changed at any time.
...
+
+##### Proxy settings
+
+These settings will configure the proxying behavior of the following services:
+* For all variants:
+    * [containerd.service](packages/containerd/containerd.service)
+    * [host-containerd.service](packages/host-ctr/host-containerd.service)
+* For Kubernetes variants:
+    * [kubelet.service](packages/kubernetes-1.18/kubelet.service)
+* For ECS variants:
```
samuelkarp: Since there's only one ECS variant, should we say "For the ECS variant"?

packages/release/proxy-env
tjkirch: I know that we're seeing the basics work, I guess my concern is that I'm not sure whether there could be any other components called by Docker (or whatever) that are assuming `/run/docker.sock`, and they've just worked for us because of the symlink, but suddenly they'd fail (or do horrible things) because they're being proxied.  (I don't know of anything, and we can fix it if we find it, hence my ship-it)

README.md
```diff
@@ -344,6 +344,28 @@ These settings can be changed at any time.
...
+    * [docker.service](packages/docker-engine/docker.service)
+    * [ecs.service](packages/ecs-agent/ecs.service)
+
+* `settings.network.http-proxy`: The HTTP proxy server to be used by services listed above.
+* `settings.network.https-proxy`: The HTTPS proxy server to be used by services listed above.
```
samuelkarp: Do we want these to be separately-configurable at this point?

packages/release/proxy-env
```diff
@@ -0,0 +1,10 @@
+{{#if settings.network.http-proxy}}
+HTTP_PROXY={{settings.network.http-proxy}}
+http_proxy={{settings.network.http-proxy}}
+{{/if}}
+{{#if settings.network.https-proxy}}
+HTTPS_PROXY={{settings.network.https-proxy}}
+https_proxy={{settings.network.https-proxy}}
```
samuelkarp: If we're targeting Go programs, either the capitalized or lower-case environment variable will work.  It should be fine to just use `HTTP_PROXY`/`HTTPS_PROXY` without duplication.

README.md
etungsten: I can't really think of a use-case for configuring `http_proxy` and `https_proxy` differently. But then again there really is no harm in letting them be separately-configurable.

packages/release/proxy-env
etungsten: I wish there is some universal standard for how these should be named and specified. I think it would be best to specify both just in case if we ever have another non-Go based service use these environment variables.

README.md
samuelkarp: It's more overhead for someone who wants to enable a proxy; now they have to configure both.  My preference is to have just one right now and we can always add a second one later if it becomes desirable.

etungsten: Would we still be setting the `HTTPS_PROXY` env var? Setting `HTTPS_PROXY` to the same contents as what's being set for `HTTP_PROXY`?

packages/docker-engine/docker.service
```diff
@@ -10,6 +10,7 @@ StartLimitIntervalSec=60s
 
 [Service]
 Type=notify
+EnvironmentFile=/etc/network/proxy.env
```
tjkirch, line 13: I just realized that if docker is now depending on settings, we need to restart it on changes, i.e. make a `docker` service in the model.

I guess the service has to be defined separately in the aws-dev and aws-ecs-1 defaults-overrides, since we probably shouldn't have it in the defaults..?

It should probably be restarted before ecs, in the ecs model, though, and we don't have dependencies between services.  Maybe in the ecs model, docker is just a restart-command in the ecs service, rather than being its own?

README.md
tjkirch: > Would we still be setting the HTTPS_PROXY env var? Setting HTTPS_PROXY to the same contents as what's being set for HTTP_PROXY?

Yep.

Perhaps the hardest question - what to call the setting?  :P

packages/docker-engine/docker.service
samuelkarp: > It should probably be restarted before ecs, in the ecs model, though, and we don't have dependencies between services.

There's maybe a race while you're configuring/restarting, but it shouldn't really matter?  The ECS agent should tolerate Docker going down and coming back up already.  There's also a dependency listed in the unit itself, though I know that our model isn't reading that.

README.md
samuelkarp: Yeah, same value for both.  I'd probably call the setting "https-proxy" to imply that the proxy needs to be able to transparently handle TLS traffic.

packages/docker-engine/docker.service
tjkirch: Ugh, this also applies to containerd, host-containerd, and kubelet, I believe.  We haven't dynamically affected the configuration of {host-,}containerd before.  kubelet we technically could, and it seems like it wouldn't have worked without a reboot, but they're mostly dynamic settings generated at boot that you wouldn't change.  (Maybe labels/taints are the worst offender?)

README.md
etungsten: What if the user wants to use a http proxy that can't handle https traffic though? Wouldn't `https-proxy` be misleading and discourage the user from using the setting?

packages/docker-engine/docker.service
tjkirch: @samuelkarp would containerd be similarly safe to restart?  No lost containers, logs, etc.?

tjkirch: @bcressey found that restarting host-containerd kills SSH sessions, implying that the admin container is stopped.  Not sure if it's a necessary kill, or if our setup is wrong.

samuelkarp: > would containerd be similarly safe to restart? No lost containers, logs, etc.?

Containerd when used either directly through its API or with Docker does no networking and wouldn't need to be restarted.

Containerd when used with Kubernetes (through cri-containerd) does networking to pull images, and would need to be restarted.

By default, containerd will not stop containers when it exits, but systemd might be killing them.

etungsten:
Push above adds restart commands for all services ingesting the templated `proxy.env` `EnvironmentFile`.

Here's how I tested the changes for Kubernetes:
1. Launched several nodes, with the `http-proxy` setting unset (no proxying).
2. I applied the bottlerocket update operator on all nodes.
3. Waited for all the agent pods and the controller pods to start running
4. Used `apiclient` from the admin container to set `settings.network.http-proxy` to my squid http proxy server.
5. Re-logged into the admin container and saw containerd.service and kubelet.service both restarted successfully.
6. Checked the agent pods and controller pod and saw them all still running with the same name (meaning they didn't die)

For testing ECS, I did the following:
1. Launched an ECS instance with the `http-proxy` setting unset (no proxying).
2. Ran a simple ECS task that simply started an nginx server with host networking.
3. Got into the admin container as curl'ed `localhost:80` and saw nginx running successfully.

<details>

```
[ec2-user@ip- ~]$ curl localhost:80
<!DOCTYPE html>
<html>
<head>
<style>
    body {
        width: 35em;
        margin: 0 auto;
        font-family: Tahoma, Verdana, Arial, sans-serif;
    }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>

</body>
</html>
```
</details>

4. Used `apiclient` from the admin container to set `settings.network.http-proxy` to my squid http proxy server.
5. Re-logged into admin container saw both docker.service and ecs.service both restarted successfully
6. curl'd `localhost:80` and saw nginx is still running fine.
7. Checked ECS console and saw the task still running fine.

Let me know if there's anything else I should test.

README.md
samuelkarp: If we're using the same setting for both, the proxy needs to be able to handle HTTPS traffic.  I think the naming should reflect that.

Most of the traffic that will be generated from the software in Bottlerocket (pulling container images, log forwarding, talking to API endpoints) is going to be over HTTPS.  cri-containerd and Docker both require HTTPS when pulling images (without manual overrides that we don't enable in Bottlerocket), and the Kubelet and ECS agent talk to their respective API endpoints over HTTPS as well.

sources/models/defaults.toml
```diff
@@ -22,13 +22,19 @@ template-path = "/usr/share/templates/motd"
...
 [configuration-files.containerd-config-toml]
 path = "/etc/containerd/config.toml"
 template-path = "/usr/share/templates/containerd-config-toml"
 
+# Host-container runtime
+
+[services.host-containerd]
+configuration-files = ["proxy-env"]
+restart-commands = ["/bin/systemctl try-restart host-containerd.service"]
+
```
bcressey: I am not comfortable adding restarts of containerd as part of this PR. Let's assume that it's disruptive to workloads until we prove otherwise. We can track it in an issue and fix it as a follow-up.

etungsten:
Converting this pull request to draft as it now depends on https://github.com/bottlerocket-os/bottlerocket/pull/1230 and https://github.com/bottlerocket-os/bottlerocket/pull/1231

etungsten:

etungsten:
Pushes above removes `http_proxy` as a settings option. Some minor formatting changes and commit message change.

PR description updated with the latest testing information.

packages/release/proxy-env
```diff
@@ -0,0 +1,6 @@
...
+{{/if}}
+NO_PROXY={{#each settings.network.no-proxy}}{{this}},{{/each}}localhost,127.0.0.1{{#if settings.kubernetes.api-server}},{{host settings.kubernetes.api-server}},.svc,.default,.local,.{{settings.kubernetes.cluster-domain}}{{/if}}
```
bcressey: Are we making queries to a `.svc`, `.default`, or `.local` TLD? 

I'm wondering why we don't have this instead:
```suggestion
NO_PROXY={{#each settings.network.no-proxy}}{{this}},{{/each}}localhost,127.0.0.1{{#if settings.kubernetes.api-server}},{{host settings.kubernetes.api-server}}{{/if}}{{#if settings.kubernetes.cluster-domain}},.{{settings.kubernetes.cluster-domain}}{{/if}}
```

sources/models/src/modeled_types/shared.rs
```diff
@@ -240,6 +240,9 @@ mod test_url {
             "http://localhost",
             "localhost/path",
             "localhost",
+            "localhost:8080",
+            ".internal",
+            ".cluster.local"
```
bcressey, line 245: Should these last two really be good URLs?

packages/release/proxy-env
etungsten: You're right. I probably don't need those. I originally thought I had to `no_proxy` kubernetes service domains with `.svc`, `.default` but they're all suffixed with the cluster domain anyways (at least according to https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/).

I'll go ahead and remove those.

sources/models/src/modeled_types/shared.rs
etungsten: Since the `no_proxy` settings is passed as a list of URLs (the modelled type), I want to allow users to specify `no_proxy` for everything under a domain name suffix and apparently any string without special UTF characters can be a valid URL if you prefix it with a scheme.

etungsten:
Push above addresses https://github.com/bottlerocket-os/bottlerocket/pull/1204#discussion_r523075152 and https://github.com/bottlerocket-os/bottlerocket/pull/1204#discussion_r544531776

sources/models/defaults.toml
bcressey: This was an older comment that somehow got resurrected in my review. I didn't catch it before I hit publish.

I'm fine with the testing you've done elsewhere, especially after the fix to `host-ctr`, so these restart commands can come back. Sorry for the churn!

etungsten:
Push above recovers the restart commands for containerd and host-containerd

bcressey (approved):

zmrow (approved):
:microbe:

samuelkarp (approved):

PR #1207, opened 2020-11-12
add compatibility symlink for cert bundle

### LINKED ISSUE #1202: Cluster autoscaler is not working with EKS because of missing certificate
I tried to use Bottlerocket in my EKS node group(s) as shown here: https://docs.aws.amazon.com/cdk/api/latest/docs/aws-eks-readme.html#bottlerocket

The worker nodes are just fine but I cannot get the cluster autoscaler working as is misses the certificate. I followed the official docs: https://docs.aws.amazon.com/eks/latest/userguide/cluster-autoscaler.html

The applied volumes expects a cert in `/etc/ssl/certs/ca-bundle.crt` but it's missing. See: https://raw.githubusercontent.com/kubernetes/autoscaler/master/cluster-autoscaler/cloudprovider/aws/examples/cluster-autoscaler-autodiscover.yaml

Amazon Linux 2 workers have it in place.

I found an issue in the autoscaler project. But the solution did not work out for me. See: https://github.com/kubernetes/autoscaler/issues/3513

bcressey:
**Issue number:**
Fixes #1202

**Description of changes:**
Add a symlink for the path where the cluster autoscaler expects to find the CA certificate bundle.

Also renames the tmpfiles snippet to follow the convention we've since adopted elsewhere, to distinguish it from other "conf" files.

**Testing done:**
Verified that the symlink is created.

```
# ls -latr /etc/ssl/certs
lrwxrwxrwx. 1 root root 16 Nov 12 16:52 /etc/ssl/certs -> ../pki/tls/certs

# ls -latr /etc/ssl/certs/ca-bundle.crt
-rw-r--r--. 1 root root 221418 Nov  6 18:35 /etc/ssl/certs/ca-bundle.crt
```

**Terms of contribution:**

### DIFF
--- a/packages/ca-certificates/ca-certificates-tmpfiles.conf
+++ b/packages/ca-certificates/ca-certificates-tmpfiles.conf
@@ -1,1 +1,2 @@
 C /etc/pki/tls/certs/ca-bundle.crt - - - -
+L /etc/ssl/certs - - - - ../pki/tls/certs
--- a/packages/ca-certificates/ca-certificates.spec
+++ b/packages/ca-certificates/ca-certificates.spec
@@ -7,7 +7,7 @@ License: MPL-2.0
 # https://hg.mozilla.org/projects/nss/log/tip/lib/ckfw/builtins/certdata.txt
 URL: https://curl.haxx.se/docs/caextract.html
 Source0: https://curl.haxx.se/ca/cacert-2020-10-14.pem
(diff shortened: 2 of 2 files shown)


samuelkarp (approved):

tjkirch (approved):

zmrow (approved):
:lab_coat:

Issue #1212, opened 2020-11-14
`thar-be-updates refresh` hangs with async apiclient


tjkirch:
#1186 moved apiclient to async/await using tokio.  One thing that missed testing is `thar-be-updates refresh`, i.e. the command invoked by the `/actions/refresh-updates` API.

The [relevant function](https://github.com/bottlerocket-os/bottlerocket/blob/bf5915a5ccf7720c6bf15edce0f0f73317a3cede/sources/api/thar-be-updates/src/main.rs#L185) uses `fork_and_return`, runs updog to get information about available updates, and then _tries_ to request settings from the API to determine the version-lock setting, so we can determine the chosen update.

The problem is that the tokio runtime was in a thread in the original process, and the forked process has no runtime.  It can't `await` any (now async) calls to apiclient.

tjkirch:
Similar issues:
* https://github.com/tokio-rs/tokio/issues/1541
* https://users.rust-lang.org/t/tokio-0-2-and-daemonize-process/42427

tjkirch:
Some options we've discussed:
1. Just revert #1186.  This means async applications can't use the apiclient library, because it would spawn a runtime inside a runtime and panic.
2. Manage tokio runtimes more directly.  We could perhaps create a runtime in the `fork_and_return` macro that would be available to the block it's passed, perhaps wrapping the block in `tokio::spawn(async move {` or something similar.
3. Fix this specific issue by moving the apiclient call before the fork, since it doesn't depend on updog output, and add documentation to `fork_and_return` explaining that if you need an async runtime in your forked process you'll have to manage it.

At the moment I lean toward (3), for simplicity, and it could lead into something like (2) in the future.  I don't think (1) is a good experience, and it's a time bomb for future apiclient users.

tjkirch:
I tried (3) and found that it's just not safe to use tokio::main if you're going to fork, even if your async calls are complete before the fork.  (The runtime hangs when shutting down; perhaps you could get around it by judicious use of `exit`, but eww.)  If you wanted to use tokio and fork, then you'd want to shut down the runtime before forking, which (as far as I know) means manual runtime management rather than tokio::main.

Instead, since we're not making any async calls in the parent process anyway, it seems easier and clearer to move thar-be-updates away from tokio::main and to briefly create a runtime (in the child) when we need it for the single async call.  You could think of it like (2) or #1213 but simplified because we can undo the async/await additions from #1186.

[...]

PR #1213, opened 2020-11-14
thar-be-updates: reduce tokio runtime scope

webern:
**Issue number:**

#1212

**Description of changes:**

Demonstrates a possible fix for the issue where the tokio runtime cannot survive forking and hangs the process. Consider this 'option 1'.

**Testing done:**

None yet.

**Terms of contribution:**

### DIFF
--- a/sources/api/thar-be-updates/src/main.rs
+++ b/sources/api/thar-be-updates/src/main.rs
@@ -196,9 +196,11 @@ async fn refresh(status: &mut UpdateStatus, socket_path: &str) -> Result<bool> {
         }
         let update_info: Vec<update_metadata::Update> =
             serde_json::from_slice(&output.stdout).context(error::UpdateInfo)?;
-        status
-            .update_available_updates(socket_path, update_info)
-            .await
+        // scope the tokio runtime to this block
+        {
+            let mut rt = tokio::runtime::Runtime::new().context(error::Runtime)?;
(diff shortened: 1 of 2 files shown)


sources/api/thar-be-updates/src/main.rs
```diff
@@ -277,7 +280,11 @@ async fn drive_state_machine(
         }
         // Refreshing the list of updates is allowed under every update state
         (UpdateCommand::Refresh, _) => {
-            refresh(update_status, socket_path).await?;
+            // scope for the tokio runtime to this block
```
tjkirch, line 283: This is one thing that bugs me about this option. You have to be really careful about the scoping, or you could create an extra runtime and panic. It's also uncommon to create runtimes in multiple places. If we were going to create it, I think it might make more sense to make sure we always have one, by using the normal tokio::main annotation and then creating one when we fork. It requires awkward wrapping, but only in one place... (I'm hoping we could come up with something less awkward later, just using the simple reordering solution for now...)

webern: > then creating one when we fork

I don't quite understand that. Do you also have to shut down the existing runtime before the call to fork?

tjkirch: I think it'd effectively disappear, and you just start a new one in the forked process like it was the first time.

Issue #1141, opened 2020-10-02
Kernel Module Support: Publish kernel-devel & /proc/config.gz artifacts for each bottlerocket release


faarshad:
**What I'd like:**
- For building out-of-tree kernel modules (e.g. [falco kernel module](https://github.com/falcosecurity/falco-website/blob/e10197ab6eb57939456c08c7fa121c476e3145ed/content/en/docs/source.md#build-kernel-module-driver)), it would benefit the community to have public access to `kernel-devel-*.rpm` and `/proc/config.gz` for each released bottlerocket-os version(e.g. `bottlerocket-aws-k8s-1.16-x86_64-v1.0.1-2a181156`) 
- Please also consider backporting i.e. publishing `kernel-devel.*.rpm` for already released versions

**Any alternatives you've considered:**
- Using a running host admin container is inconvenient since it requires to standup a runtime os and has manual steps involved.
- Also, bottlerocket-sdk versions are not aligned 1-1 with a particular bottlerocket-os release which results in difficulty correlating the two.

cc: @bcressey @jhaynes

bcressey:
The kernel config is included in the devel archive as `/usr/src/kernels/%{version}/.config`, so we probably don't need a separate `config.gz` artifact.

Retroactively publishing archives for the existing releases is a great idea.

Progress on getting a portable artifact out of the SDK can be tracked in bottlerocket-os/bottlerocket-sdk/issues/7. We might end up publishing a toolchain artifact here along with each release.

bcressey:
I have not tested it, but these toolchains from [toolchains.bootlin.com](https://toolchains.bootlin.com) should work in the interim:
* [x86_64](https://toolchains.bootlin.com/downloads/releases/toolchains/x86-64-core-i7/tarballs/x86-64-core-i7--glibc--stable-2020.08-1.tar.bz2)
* [aarch64](https://toolchains.bootlin.com/downloads/releases/toolchains/aarch64/tarballs/aarch64--glibc--stable-2020.08-1.tar.bz2)

We're currently on GCC 9.3.0 and binutils 2.33.1, which matches up with their stable release.

bcressey:
#1275 is our current plan to integrate this into the release process.

arnaldo2792:
[...]

PR #1220, opened 2020-11-19
improve support for out-of-tree kernel modules

bcressey:
**Issue number:**
#1141 

**Description of changes:**
This overhauls our packaging of kernel development sources, so that they can be used on an architecture that differs from the build host, and can be extracted for use outside of a running host.

**Testing done:**
Tests were repeated across this matrix:
* x86_64 image and kmod kit built on an x86_64 host
* x86_64 image and kmod kit built on an aarch64 host
* aarch64 image and kmod kit built on an x86_64 host
* aarch64 image and kmod kit built on an aarch64 host

Images were tested by launching nodes, logging into a Fedora-based admin container, building the Wireguard out-of-tree module, and then running `make prepare`.

x86_64 images were additionally tested by using Helm to install Falco, and verifying that the module could be built by DKMS and loaded:
```
helm install falco falcosecurity/falco \
  --namespace kube-system \
  --set docker.enabled=false
```

The kmod kits were tested on Fedora 32 and Amazon Linux 2, by building the Wireguard and ZFS out-of-tree modules, using snippets like this:
```
FROM base as wireguard-build
ARG VARIANT="aws-k8s-1.18"
ARG ARCH="x86_64"
ARG VERSION="1.0.3"
ARG KIT="${VARIANT}-${ARCH}-kmod-kit-v${VERSION}"
ARG KERNELDIR="/tmp/${KIT}/kernel-devel"
ARG CROSS_COMPILE="${ARCH}-bottlerocket-linux-musl-"
ARG INSTALL_MOD_STRIP=1
RUN \
  export PATH="/tmp/${KIT}/toolchain/usr/bin:${PATH}" && \
  make -C wireguard-linux-compat/src module
```

The modules were then loaded on a running instance to confirm they were built correctly.

x86_64 kmod kits were additionally tested by building the Falco module.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -7,6 +7,7 @@ BUILDSYS_ROOT_DIR = "${CARGO_MAKE_WORKING_DIRECTORY}"
 BUILDSYS_BUILD_DIR = "${BUILDSYS_ROOT_DIR}/build"
 BUILDSYS_PACKAGES_DIR = "${BUILDSYS_BUILD_DIR}/rpms"
 BUILDSYS_IMAGES_DIR = "${BUILDSYS_BUILD_DIR}/images"
+BUILDSYS_ARCHIVES_DIR = "${BUILDSYS_BUILD_DIR}/archives"
 BUILDSYS_TOOLS_DIR = "${BUILDSYS_ROOT_DIR}/tools"
 BUILDSYS_SOURCES_DIR = "${BUILDSYS_ROOT_DIR}/sources"
 BUILDSYS_TIMESTAMP = { script = ["date +%s"] }
@@ -80,8 +81,9 @@ DOCKER_BUILDKIT = "1"
 # Certain variables are defined here to allow us to override a component value
 # on the command line.
 
-# Depends on ${BUILDSYS_ARCH}.
+# Depends on ${BUILDSYS_ARCH} and ${BUILDSYS_SDK_VERSION}.
 BUILDSYS_SDK_IMAGE = { script = [ "echo bottlerocket/sdk-${BUILDSYS_ARCH}:v${BUILDSYS_SDK_VERSION}-$(uname -m)" ] }
+BUILDSYS_TOOLCHAIN = { script = [ "echo bottlerocket/toolchain-${BUILDSYS_ARCH}:v${BUILDSYS_SDK_VERSION}-${BUILDSYS_ARCH}" ] }
 
 # Depends on ${BUILDSYS_JOBS}.
 CARGO_MAKE_CARGO_ARGS = "--jobs ${BUILDSYS_JOBS} --offline --locked"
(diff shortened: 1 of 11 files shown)


packages/kernel/kernel.spec
```diff
@@ -84,35 +90,83 @@ find %{buildroot}%{_cross_prefix} \
...
 (
-  find * -name Kbuild\* -type f -print  \
-    -o -name Kconfig\* -type f -print \
-    -o -name Makefile\* -type f -print \
-    -o -name module.lds -type f -print \
-    -o -name Platform -type f -print
-  find arch/*/include/ include/ -type f -o -type l
-  find scripts/ -executable -type f
-  find scripts/ ! \( -name Makefile\* -o -name Kbuild\* \) -type f
+  find * \
```
tjkirch, line 123: This section is serious magic / painstaking effort.  Is there anything that could be documented so that others know how to determine what should be included, even if it's pointing at similar examples elsewhere, perhaps?

Makefile.toml
```diff
@@ -7,6 +7,7 @@ BUILDSYS_ROOT_DIR = "${CARGO_MAKE_WORKING_DIRECTORY}"
 BUILDSYS_BUILD_DIR = "${BUILDSYS_ROOT_DIR}/build"
 BUILDSYS_PACKAGES_DIR = "${BUILDSYS_BUILD_DIR}/rpms"
 BUILDSYS_IMAGES_DIR = "${BUILDSYS_BUILD_DIR}/images"
+BUILDSYS_ARCHIVES_DIR = "${BUILDSYS_BUILD_DIR}/archives"
```
tjkirch, line 10: Hmm, archives sounds like old files, or random tar files, or perhaps packages.  Maybe we should call them build... kits?  ;)

Makefile.toml
```diff
@@ -262,6 +264,57 @@ cargo build \
...
+kmod_kit="${BUILDSYS_VARIANT}-${BUILDSYS_ARCH}-kmod-kit-v${BUILDSYS_VERSION_IMAGE}"
+rm -f "${BUILDSYS_ARCHIVES_DIR}/${kmod_kit}.tar.xz"
+
+prepare_kmod_kit="
+set -e -o pipefail
+
+mkdir /tmp/${kmod_kit}
+
+# Retrieve the toolchain and kernel archives.
+curl --silent --output /tmp/${kmod_kit}/toolchain.tar.xz \
```
```suggestion
curl --silent --show-error --output /tmp/${kmod_kit}/toolchain.tar.xz \
```
Any reason not to expose errors?

Makefile.toml
```diff
@@ -262,6 +264,57 @@ cargo build \
...
+  https://${BUILDSYS_SDK_SITE}/${BUILDSYS_TOOLCHAIN}.tar.xz
+find ./rpms -name '*-${BUILDSYS_ARCH}-kernel-archive-*.rpm' \
+  -exec rpm2cpio {} \; | cpio -idmu --quiet
```
tjkirch: This brings to mind times I've been bitten by old rpms... should we be a little more selective somehow?  Or just take the latest?

tjkirch (approved):

sam-aws (approved):
:christmas_tree: 

It's times like these that make me look fondly back on Buildroot

zmrow (approved):
The `kernel.spec` changes are over my head a bit, but everything else looks good!

:turkey:

packages/kernel/kernel.spec
bcressey: I'm not sure how to do better than the comment that says that this is what we need for the three targets (scripts, prepare, modules). The set of files is subject to change over time since the authoritative source is the kernel tree for a specific version.

"Keep adding things until those targets work" and "stop adding things once they do" is unsatisfying advice, but it pretty accurately reflects the workflow.

Pointing to a different packaging effort can be misleading since the set of files changes over time and can be influenced by the kernel configuration. We also need more of the sources than others might because we can't include the host tools.

Makefile.toml
bcressey: That's a good call out - I've been testing on a clean tree, but the result of this would be ambiguous if two kernel-archive packages were present, which could happen when testing an update to a newer kernel.

tjkirch (approved):

PR #1230, opened 2020-12-03
host-ctr, host-containers: proper restarts

### LINKED ISSUE #1229: host-containers get restarted when host-containerd restarts
**Image I'm using:**
Bottlerocket v1.0.4 aws-k8s-1.17

host-containers, like the admin container, to stay running when the host-containerd daemon restarts.

**What actually happened:**
The admin container task gets killed, terminating my SSH connection to the host.

Restart `host-containerd.service` via the admin container.

`host-containers@` services should change their dependency on `host-containerd` from `BindsTo` to `Wants` to keep them running if `host-containerd` stops or gets restarted. When `host-ctr` eventually exits non-zero due to losing the connection to host-containerd, `host-containers@` services should restart and let `host-ctr` re-bind to the existing host container task.

`host-ctr` needs to learn to not kill the existing container task https://github.com/bottlerocket-os/bottlerocket/blob/cd66f45db17fcd4da50f589de35bee74cf286e6f/sources/host-ctr/cmd/host-ctr/main.go#L167-L171 and instead try to take the helm over the container task again.

etungsten:
**I recommend reviewing each commit separately**

**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/1229

**Description of changes:**
```
Date:   Sun Dec 6 20:54:03 2020 -0800

    host-containers: add safeguards against lingering host containers
    
    Now that host-ctr has the ability to rebind to existing host containers.
    We want to ensure whenever we enable host containers the container will
    be running with its latest configuration.
    
    We utilize `host-ctr`'s clean-up command to clean up any potential
    lingering host-container when we're enabling a previously disabled
    host-container and whenever a host-container is disabled.
```

```
Date:   Fri Dec 4 20:02:37 2020 -0800

    host-ctr: add new subcommand `clean-up`
    
    Adds a new subcommand `clean-up` that checks if a given container
    exists, if it does, `host-ctr` will attempt to kill the container task
    and delete the container.
```

```
Date:   Fri Dec 4 18:34:07 2020 -0800

    host-ctr: refactoring
    
    Refactors `host-ctr`.
    Categorizes functionality into subcommands.

```

```
Date:   Fri Dec 4 14:31:26 2020 -0800

    host-containers@: remove KillMode=mixed
    
    We don't need systemd to go and actively try kill all processes
    of the unit's cgroup.

```

```
Date:   Wed Dec 2 18:18:57 2020 -0800

    host-ctr: do not kill existing container, take over it
    
    If the host-container already exists, we should just take over the
    helm and not try to replace it with a new container. This is so that
    even if we temporarily lose connection with host-containerd, we can
    still eventually get the task status when containerd comes back up.

commit 0dbc03cdcde435aa259f34d80f6ec66a202b45a4
```
```
Date:   Wed Dec 2 14:45:44 2020 -0800

    host-containers: 'Wants' host-containerd instead of 'BindsTo'
    
    host-containers@ systemd units should not stop when host-containerd is
    restarted or killed.
    
    By changing host-containers' dependency on host-containerd.service from
    `BindsTo=` to `Wants=` we ensure host containers tasks won't be killed if
     host-containerd temporarily stops.
    
    host-ctr then has a chance to reclaim the container task when
    host-containerd comes back up.

```

**Testing done:**
- Built AMI, launched instance
- `sudo sheltie` into the host via the admin container
- Restarted host-containerd, and my ssh connection to the admin container was still alive
- Checked the status of `host-containers@admin` and saw that it exited and restarted successfully.

`host-containers@admin` initial starts successfully.
```
Dec 03 03:09:30 host-ctr[3077]: Server listening on 0.0.0.0 port 22.
Dec 03 03:09:30 host-ctr[3077]: Server listening on :: port 22.
Dec 03 03:10:48 host-ctr[3077]: Accepted publickey for ec2-user from 123.123.123.123 port 5417 ssh2
```

This is where I restarted `host-containerd`. `host-ctr` loses connection to the containerd server and exits.
```
Dec 03 03:11:28 host-ctr[3077]: time="2020-12-03T03:11:28Z" level=error msg="failed to get container task
 exit status" error="rpc error: code = Unavailable desc = transport is closing"
Dec 03 03:11:28 host-ctr[3077]: time="2020-12-03T03:11:28Z" level=error msg="failed to delete container t
ask" error="task must be stopped before deletion: running: failed precondition"
Dec 03 03:11:28 host-ctr[3077]: time="2020-12-03T03:11:28Z" level=error msg="failed to cleanup container"
 error="cannot delete running task admin: failed precondition"
Dec 03 03:11:28 systemd[1]: host-containers@admin.service: Main process exited, c
ode=exited, status=1/FAILURE
Dec 03 03:11:28 systemd[1]: host-containers@admin.service: Failed wit
h result 'exit-code'.
```

`host-containers@admin` restarts and `host-ctr` successfully rebinds to the admin container task that's already running.
```
Dec 03 03:12:14 systemd[1]: host-containers@admin.service: Scheduled restart job, restart counter is at 1
.
Dec 03 03:12:14 systemd[1]: Stopped Host container: admin.
Dec 03 03:12:14 systemd[1]: Starting Host container: admin...
Dec 03 03:12:14 systemd[1]: Started Host container: admin.
Dec 03 03:12:14 host-ctr[5096]: time="2020-12-03T03:12:14Z" level=info msg="Pulling with Amazon ECR Resol
ver" ref="ecr.aws/arn:aws:ecr:us-west-2:328549459982:repository/bottlerocket-admin:v0.5.2"
Dec 03 03:12:14 host-ctr[5096]: time="2020-12-03T03:12:14Z" level=info msg="Pulled successfully" img="ecr
.aws/arn:aws:ecr:us-west-2:328549459982:repository/bottlerocket-admin:v0.5.2"
Dec 03 03:12:14 host-ctr[5096]: time="2020-12-03T03:12:14Z" level=info msg=Unpacking... img="ecr.aws/arn:
aws:ecr:us-west-2:328549459982:repository/bottlerocket-admin:v0.5.2"
Dec 03 03:12:14 host-ctr[5096]: time="2020-12-03T03:12:14Z" level=info msg="Tagging image" imageName="328
549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.5.2"
Dec 03 03:12:14 host-ctr[5096]: time="2020-12-03T03:12:14Z" level=info msg="Container task is still runni
ng, proceeding to monitor it"
```

Testing for when both host-containers and host-containerd are restarted due to a single API transaction. We ensure the restarted host-ctr runs an up-to-date host container that reflects the setting changes:
See https://github.com/bottlerocket-os/bottlerocket/pull/1230#issuecomment-740085820

**Terms of contribution:**

### DIFF
--- a/packages/os/host-containers@.service
+++ b/packages/os/host-containers@.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Host container: %i
 After=host-containerd.service
-BindsTo=host-containerd.service
+Wants=host-containerd.service
 
 [Service]
 Type=simple
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -129,6 +129,21 @@ func App() *cli.App {
 				return pullImageOnly(containerdSocket, namespace, source)
 			},
 		},
+		{
+			Name:  "clean-up",
+			Usage: "delete specified container's resources if it exists",
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:        "container-id",
+					Usage:       "the id of the container to clean up",
+					Destination: &containerID,
+					Required:    true,
+				},
+			},
+			Action: func(c *cli.Context) error {
+				return cleanUp(containerdSocket, namespace, containerID)
+			},
+		},
 	}
 
 	return app
@@ -404,6 +419,58 @@ func pullImageOnly(containerdSocket string, namespace string, source string) err
 			log.G(ctx).WithField("ref", ref).Error(err)
 			return err
 		}
+	}
+
+	return nil
+}
+
+// cleanUp checks if the specified container exists and attempts to clean it up
+func cleanUp(containerdSocket string, namespace string, containerID string) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	ctx = namespaces.WithNamespace(ctx, namespace)
+
+	client, err := newContainerdClient(ctx, containerdSocket, namespace)
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	container, err := client.LoadContainer(ctx, containerID)
+	if err != nil {
+		if errdefs.IsNotFound(err) {
+			log.G(ctx).WithField("container-id", containerID).Info("container does not exist, no clean up necessary")
+			return nil
+		}
+		log.G(ctx).WithField("container-id", containerID).WithError(err).Error("failed to retrieve list of containers")
+		return err
+	}
+	if container != nil {
+		log.G(ctx).WithField("container-id", containerID).Info("container exists, deleting")
+		// Kill task associated with existing container if it exists
+		task, err := container.Task(ctx, nil)
+		if err != nil {
+			// No associated task found, proceed to delete existing container
+			if errdefs.IsNotFound(err) {
+				log.G(ctx).WithField("container-id", containerID).Info("no task associated with existing container")
+			} else {
+				log.G(ctx).WithField("container-id", containerID).WithError(err).Error("failed to retrieve task associated with existing container")
+				return err
+			}
+		}
+		if task != nil {
+			_, err := task.Delete(ctx, containerd.WithProcessKill)
+			if err != nil {
+				log.G(ctx).WithField("container-id", containerID).WithError(err).Error("failed to delete existing container task")
+				return err
+			}
+			log.G(ctx).WithField("container-id", containerID).Info("killed existing container task")
+		}
+		if err := container.Delete(ctx, containerd.WithSnapshotCleanup); err != nil {
+			log.G(ctx).WithField("container-id", containerID).WithError(err).Error("failed to delete existing container")
+			return err
+		}
+		log.G(ctx).WithField("container-id", containerID).Info("deleted existing container")
 	}
 
 	return nil
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -173,65 +173,73 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 		}
 	}
 
-	// Clean up target container if it already exists before starting container
-	// task.
-	if err := deleteCtrIfExists(ctx, client, containerID); err != nil {
-		return err
+	// Check if the target container already exists. If it does, take over the helm to manage it.
+	container, err := client.LoadContainer(ctx, containerID)
+	if err != nil {
+		if errdefs.IsNotFound(err) {
+			log.G(ctx).WithField("ctr-id", containerID).Info("Container does not exist, proceeding to create it")
+		} else {
+			log.G(ctx).WithField("ctr-id", containerID).WithError(err).Error("failed to retrieve list of containers")
+			return err
+		}
 	}
 
-	// Get the cgroup path of the systemd service
-	cgroupPath, err := cgroups.GetOwnCgroup("name=systemd")
-	if err != nil {
-		log.G(ctx).WithError(err).Error("failed to discover systemd cgroup path")
-		return err
-	}
+	// If the container doesn't already exist, create it
+	if container == nil {
+		// Get the cgroup path of the systemd service
+		cgroupPath, err := cgroups.GetOwnCgroup("name=systemd")
+		if err != nil {
+			log.G(ctx).WithError(err).Error("failed to discover systemd cgroup path")
+			return err
+		}
 
-	// Set up the container spec. See `withSuperpowered` for conditional options
-	// set when configured as superpowered.
-	ctrOpts := containerd.WithNewSpec(
-		oci.WithImageConfig(img),
-		oci.WithHostNamespace(runtimespec.NetworkNamespace),
-		oci.WithHostHostsFile,
-		oci.WithHostResolvconf,
-		// Launch the container under the systemd unit's cgroup
-		oci.WithCgroup(cgroupPath),
-		// Mount in the API socket for the Bottlerocket API server, and the API
-		// client used to interact with it
-		oci.WithMounts([]runtimespec.Mount{
-			{
-				Options:     []string{"bind", "rw"},
-				Destination: "/run/api.sock",
-				Source:      "/run/api.sock",
-			},
-			// Mount in the apiclient to make API calls to the Bottlerocket API server
-			{
-				Options:     []string{"bind", "ro"},
-				Destination: "/usr/local/bin/apiclient",
-				Source:      "/usr/bin/apiclient",
-			},
-			// Mount in the persistent storage location for this container
-			{
-				Options:     []string{"rbind", "rw"},
-				Destination: "/.bottlerocket/host-containers/" + containerID,
-				Source:      "/local/host-containers/" + containerID,
-			}}),
-		// Mount the rootfs with an SELinux label that makes it writable
-		withMountLabel("system_u:object_r:state_t:s0"),
-		// Include conditional options for superpowered containers.
-		withSuperpowered(superpowered),
-	)
+		// Set up the container spec. See `withSuperpowered` for conditional options
+		// set when configured as superpowered.
+		ctrOpts := containerd.WithNewSpec(
+			oci.WithImageConfig(img),
+			oci.WithHostNamespace(runtimespec.NetworkNamespace),
+			oci.WithHostHostsFile,
+			oci.WithHostResolvconf,
+			// Launch the container under the systemd unit's cgroup
+			oci.WithCgroup(cgroupPath),
+			// Mount in the API socket for the Bottlerocket API server, and the API
+			// client used to interact with it
+			oci.WithMounts([]runtimespec.Mount{
+				{
+					Options:     []string{"bind", "rw"},
+					Destination: "/run/api.sock",
+					Source:      "/run/api.sock",
+				},
+				// Mount in the apiclient to make API calls to the Bottlerocket API server
+				{
+					Options:     []string{"bind", "ro"},
+					Destination: "/usr/local/bin/apiclient",
+					Source:      "/usr/bin/apiclient",
+				},
+				// Mount in the persistent storage location for this container
+				{
+					Options:     []string{"rbind", "rw"},
+					Destination: "/.bottlerocket/host-containers/" + containerID,
+					Source:      "/local/host-containers/" + containerID,
+				}}),
+			// Mount the rootfs with an SELinux label that makes it writable
+			withMountLabel("system_u:object_r:state_t:s0"),
+			// Include conditional options for superpowered containers.
+			withSuperpowered(superpowered),
+		)
 
-	// Create and start the container.
-	container, err := client.NewContainer(
-		ctx,
-		containerID,
-		containerd.WithImage(img),
-		containerd.WithNewSnapshot(containerID+"-snapshot", img),
-		ctrOpts,
-	)
-	if err != nil {
-		log.G(ctx).WithError(err).WithField("img", img.Name).Error("failed to create container")
-		return err
+		// Create the container.
+		container, err = client.NewContainer(
+			ctx,
+			containerID,
+			containerd.WithImage(img),
+			containerd.WithNewSnapshot(containerID+"-snapshot", img),
+			ctrOpts,
+		)
+		if err != nil {
+			log.G(ctx).WithError(err).WithField("img", img.Name).Error("failed to create container")
+			return err
+		}
 	}
 	defer func() {
 		// Clean up the container as program wraps up.
@@ -243,11 +251,52 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 		}
 	}()
 
-	// Create the container task
-	task, err := container.NewTask(ctx, cio.NewCreator(cio.WithStdio))
+	// Check if the container task already exists. If it does, try to manage it.
+	task, err := container.Task(ctx, cio.NewAttach(cio.WithStdio))
 	if err != nil {
-		log.G(ctx).WithError(err).Error("failed to create container task")
-		return err
+		if errdefs.IsNotFound(err) {
+			log.G(ctx).WithField("container-id", containerID).Info("container task does not exist, proceeding to create it")
+		} else {
+			log.G(ctx).WithField("container-id", containerID).WithError(err).Error("failed to retrieve container task")
+			return err
+		}
+	}
+	// If the container doesn't already exist, create it
+	taskAlreadyRunning := false
+	if task == nil {
+		// Create the container task
+		task, err = container.NewTask(ctx, cio.NewCreator(cio.WithStdio))
+		if err != nil {
+			log.G(ctx).WithError(err).Error("failed to create container task")
+			return err
+		}
+	} else {
+		// Check the container task process status and see if it's already running.
+		taskStatus, err := task.Status(ctx)
+		if err != nil {
+			log.G(ctx).WithError(err).Error("failed to retrieve container task status")
+			return err
+		}
+		log.G(ctx).WithField("task status", taskStatus.Status).Info("found existing container task")
+
+		// If the task isn't running (it's in some weird state like `Paused`), we should replace it with a new task.
+		if taskStatus.Status != containerd.Running {
+			_, err := task.Delete(ctx, containerd.WithProcessKill)
+			if err != nil {
+				log.G(ctx).WithError(err).Error("failed to delete existing container task")
+				return err
+			}
+			log.G(ctx).Info("killed existing container task to replace it with a new task")
+			// Recreate the container task
+			task, err = container.NewTask(ctx, cio.NewCreator(cio.WithStdio))
+			if err != nil {
+				log.G(ctx).WithError(err).Error("failed to create container task")
+				return err
+			}
+		} else {
+			log.G(ctx).Info("container task is still running, proceeding to monitor it")
+			taskAlreadyRunning = true
+		}
 	}
 	defer func() {
 		// Clean up the container's task as program wraps up.
@@ -259,20 +308,20 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 		}
 	}()
 
-	// Call Wait before calling Start to ensure the task's status notifications
-	// are received.
+	// Call `Wait` to ensure the task's status notifications are received.
 	exitStatusC, err := task.Wait(context.TODO())
 	if err != nil {
 		log.G(ctx).WithError(err).Error("unexpected error during container task setup")
 		return err
 	}
-
-	// Execute the target container's task.
-	if err := task.Start(ctx); err != nil {
-		log.G(ctx).WithError(err).Error("failed to start container task")
-		return err
+	if !taskAlreadyRunning {
+		// Execute the target container's task.
+		if err := task.Start(ctx); err != nil {
+			log.G(ctx).WithError(err).Error("failed to start container task")
+			return err
+		}
+		log.G(ctx).Info("successfully started container task")
 	}
-	log.G(ctx).Info("successfully started container task")
 
 	// Block until an OS signal (e.g. SIGTERM, SIGINT) is received or the
 	// container task finishes and exits on its own.
@@ -420,48 +469,6 @@ func newContainerdClient(ctx context.Context, containerdSocket string, namespace
 	}
 
 	return client, nil
-}
-
-// deleteCtrIfExists cleans up an existing container. This involves killing its
-// task then deleting it and its snapshot when any exist.
-func deleteCtrIfExists(ctx context.Context, client *containerd.Client, targetCtr string) error {
-	existingCtr, err := client.LoadContainer(ctx, targetCtr)
-	if err != nil {
-		if errdefs.IsNotFound(err) {
-			log.G(ctx).WithField("container-id", targetCtr).Info("no clean up necessary, proceeding")
-			return nil
-		}
-		log.G(ctx).WithField("container-id", targetCtr).WithError(err).Error("failed to retrieve list of containers")
-		return err
-	}
-	if existingCtr != nil {
-		log.G(ctx).WithField("container-id", targetCtr).Info("container already exists, deleting")
-		// Kill task associated with existing container if it exists
-		existingTask, err := existingCtr.Task(ctx, nil)
-		if err != nil {
-			// No associated task found, proceed to delete existing container
-			if errdefs.IsNotFound(err) {
-				log.G(ctx).WithField("container-id", targetCtr).Info("no task associated with existing container")
-			} else {
-				log.G(ctx).WithField("container-id", targetCtr).WithError(err).Error("failed to retrieve task associated with existing container")
-				return err
-			}
-		}
-		if existingTask != nil {
-			_, err := existingTask.Delete(ctx, containerd.WithProcessKill)
-			if err != nil {
-				log.G(ctx).WithField("container-id", targetCtr).WithError(err).Error("failed to delete existing container task")
-				return err
-			}
-			log.G(ctx).WithField("container-id", targetCtr).Info("killed existing container task")
-		}
-		if err := existingCtr.Delete(ctx, containerd.WithSnapshotCleanup); err != nil {
-			log.G(ctx).WithField("container-id", targetCtr).WithError(err).Error("failed to delete existing container")
-			return err
-		}
-		log.G(ctx).WithField("container-id", targetCtr).Info("deleted existing container")
-	}
-	return nil
 }
 
 // withMountLabel configures the mount with the provided SELinux label.
--- a/sources/api/host-containers/src/main.rs
+++ b/sources/api/host-containers/src/main.rs
@@ -30,6 +30,7 @@ const API_SETTINGS_URI: &str = "/settings";
 const ENV_FILE_DIR: &str = "/etc/host-containers";
 
 const SYSTEMCTL_BIN: &str = "/bin/systemctl";
+const HOST_CTR_BIN: &str = "/bin/host-ctr";
 
 mod error {
     use http::StatusCode;
@@ -90,9 +91,9 @@ mod error {
             source: std::io::Error,
         },
 
-        #[snafu(display("Systemd command failed - stderr: {}",
-                        std::str::from_utf8(&output.stderr).unwrap_or_else(|_| "<invalid UTF-8>")))]
-        SystemdCommandFailure { output: Output },
+        #[snafu(display("'{}' failed - stderr: {}",
+                        bin_path, std::str::from_utf8(&output.stderr).unwrap_or_else(|_| "<invalid UTF-8>")))]
+        CommandFailure { bin_path: String, output: Output },
 
         #[snafu(display("Failed to manage {} of {} host containers", failed, tried))]
         ManageContainersFailed { failed: usize, tried: usize },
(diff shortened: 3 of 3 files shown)


sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -164,106 +164,153 @@ func _main() int {
...
-
-	// Get the cgroup path of the systemd service
-	cgroupPath, err := cgroups.GetOwnCgroup("name=systemd")
+	// Check if the target container already exists. If it does, take over the helm to manage it.
+	container, err := client.LoadContainer(ctx, containerID)
 	if err != nil {
-		log.G(ctx).WithError(err).Error("Failed to discover systemd cgroup path")
-		return 1
+		if errdefs.IsNotFound(err) {
+			log.G(ctx).WithField("ctr-id", containerID).Info("Container does not exists, proceeding to create it")
```
zmrow: s/exists/exist

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -312,55 +359,13 @@ func _main() int {
...
+		log.G(ctrCtx).WithError(err).Error("failed to get container task exit status")
 		return 1
 	}
 	log.G(ctrCtx).WithField("code", code).Info("Container task exited")
 	return int(code)
 }
 
-// deleteCtrIfExists cleans up an existing container. This involves killing its
-// task then deleting it and its snapshot when any exist.
-func deleteCtrIfExists(ctx context.Context, client *containerd.Client, targetCtr string) error {
```
zmrow, line 427: Yay so much red! :red_square:

zmrow (approved):
Just a spelling nit.

:snowman_with_snow:

etungsten:
Push above adds a condition during clean up to not delete the task and container if we're returning due to host-containerd closing its connection. The deletions will fail anyways so we should not even attempt deletion.

etungsten:
Push above reverts the previous force push. I realized that the change wasn't working in the way I expected it to.

Also addresses typo pointed out by @zmrow 's comment.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -164,106 +164,153 @@ func _main() int {
...
-
-	// Get the cgroup path of the systemd service
-	cgroupPath, err := cgroups.GetOwnCgroup("name=systemd")
+	// Check if the target container already exists. If it does, take over the helm to manage it.
+	container, err := client.LoadContainer(ctx, containerID)
 	if err != nil {
-		log.G(ctx).WithError(err).Error("Failed to discover systemd cgroup path")
-		return 1
+		if errdefs.IsNotFound(err) {
+			log.G(ctx).WithField("ctr-id", containerID).Info("Container does not exist, proceeding to create it")
```
webern, line 195: Nit, seems like you went to the trouble of unifying on lowercase elsewhere?

```suggestion
			log.G(ctx).WithField("ctr-id", containerID).Info("container does not exist, proceeding to create it")
```

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -164,106 +164,153 @@ func _main() int {
...
 		if err != nil {
-			log.G(cleanup).WithError(err).Error("Failed to cleanup container")
+			log.G(cleanup).WithError(err).Error("failed to cleanup container")
 		}
 	}()
 
-	// Create the container task
-	task, err := container.NewTask(ctx, cio.NewCreator(cio.WithStdio))
+	// Check if the container task already exists. If it does, try to manage it.
+	task, err := container.Task(ctx, cio.NewAttach(cio.WithStdio))
```
webern, line 270: Is there any other error case that should be checked? For example, Below where you say

> If the task isn't running (it's in some weird state like `Paused`), we should replace it with a new task.

Is it possible that such a container could have caused an err at line 246?

packages/os/host-containers@.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=Host container: %i
 After=host-containerd.service
-BindsTo=host-containerd.service
+Wants=host-containerd.service
```
samuelkarp, line 4: Why `Wants` instead of `Requires`?

webern (approved):
Looks good. One probing question.

sources/host-ctr/cmd/host-ctr/main.go
etungsten: I was trying to follow the golang style guide regarding error strings https://github.com/golang/go/wiki/CodeReviewComments#error-strings. 

It also explicit says that this doesn't apply to normal log messages. So I kept info logs the same.

sources/host-ctr/cmd/host-ctr/main.go
etungsten: If the task exists, then `container.Task` shouldn't error. The task has a self-contained task state that does not affect the container itself.

If the task does not exist, line 248 branches on that to proceed to continue to create the task.

samuelkarp:
@etungsten wrote

> I was trying to follow the golang style guide regarding error strings golang/go/wiki/CodeReviewComments#error-strings.
>
> It also explicit says that this doesn't apply to normal log messages. So I kept info logs the same.

(it won't let me reply inline for some reason)

This guidance applies to the `error` type, not to the log output (thought it's intended to improve log output when errors are appended onto log lines for context).  I don't have any problem with the changes you made, but I did want to clarify that.

sources/host-ctr/cmd/host-ctr/main.go
samuelkarp: Have you tested that `cio.NewAttach(cio.WithStdio)` works for a task that was launched by a prior copy of host-ctr?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -164,106 +164,153 @@ func _main() int {
...
 	if err != nil {
-		log.G(ctx).WithError(err).Error("Failed to create container task")
-		return 1
+		if errdefs.IsNotFound(err) {
+			log.G(ctx).WithField("ctr-id", containerID).Info("Container task does not exists, proceeding to create it")
```
samuelkarp: nit:

```suggestion
			log.G(ctx).WithField("ctr-id", containerID).Info("Container task does not exist, proceeding to create it")
```

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -164,106 +164,153 @@ func _main() int {
...
+		}
+	} else {
+		// Check the container task process status and see if it's already running.
+		taskStatus, err := task.Status(ctx)
+		if err != nil {
+			log.G(ctx).WithError(err).Error("failed to retrieve container task status")
+			return 1
+		}
+		// If the task isn't running (it's in some weird state like `Paused`), we should replace it with a new task.
+		if taskStatus.Status != containerd.Running {
```
samuelkarp, line 298: It's worth logging the current `taskStatus.Status`.

samuelkarp (commented):
Generally LGTM with one small requested change around adding another log line.  With the introduction of this PR, the next change is about time to break up the `_main` function into smaller bits.

sources/host-ctr/cmd/host-ctr/main.go
etungsten: Yes, I am able to see the stdout in host-containers@admin's journal log even after a restart (meaning a new host-ctr instance)

Making a new ssh connection after host-containers@admin restarts.
```
....
Dec 04 01:07:35 host-ctr[4773]: time="2020-12-04T
01:07:35Z" level=info msg="Container task is still running, proceeding to monitor it"
Dec 04 01:07:51 host-ctr[4773]: Accepted publicke
y for ec2-user from 123.123.123.123 port 2682 ssh2: RSA SHA256:
```

packages/os/host-containers@.service
etungsten: I don't want systemd to stop `host-containers@.service` if `host-containerd` stops for any reason. If systemd stops `host-containers@service`, systemd will terminate everything under its cgroup, which includes the container process.

From https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Requires=
> Besides, with or without specifying After=, this unit will be stopped if one of the other units is explicitly stopped.

samuelkarp (approved):

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -164,106 +164,155 @@ func _main() int {
...
-	// Clean up target container if it already exists before starting container
-	// task.
-	if err := deleteCtrIfExists(ctx, client, containerID); err != nil {
-		return 1
-	}
-
-	// Get the cgroup path of the systemd service
-	cgroupPath, err := cgroups.GetOwnCgroup("name=systemd")
+	// Check if the target container already exists. If it does, take over the helm to manage it.
+	container, err := client.LoadContainer(ctx, containerID)
```
bcressey, line 192: Previously we handled the case where we modify the image for a host container that's currently running, since we'd always kill it and restart afterwards.

Now it looks like if we already have "admin" running, we'll continue using it even if the image has changed. I don't think that's the behavior we want.

bcressey (changes requested):

sources/host-ctr/cmd/host-ctr/main.go
bcressey: We should also handle toggling superpowered on and off.

etungsten: If the image for the host container is changed via settings. Then the corresponding `host-containers@` service will be restarted by systemd via [`restart-commands = ["/usr/bin/host-containers"]`](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/host-containers/src/main.rs) . This means `host-ctr` will actually receive a termination signal and proceed to try and terminate the container task and restart. That hasn't changed.

What this is trying to address is `host-ctr`'s connection to containerd being closed off suddenly and losing track of the container task status and coming back up to try and reclaim the original running host container task.

etungsten: Oh I guess looking at the code, it seems like to actually apply the new host containers setting (whether it be toggling superpowered or changing the image URL) users would have to toggle the `enable` setting as well. `host-containers` (the binary doesn't actually handle restarts?)

Is that the intended workflow? @zmrow @tjkirch ?

bcressey: It doesn't seem robust to the case where `host-ctr` itself dies for some reason, gets restarted, and finds a running container - which may not be running with the right settings. Previously we did handle this correctly, at the cost of tearing down a running container that was otherwise fine.

Can we inspect the running container and determine that it's correct based on the image and one of the superpowered properties from the spec?

bcressey: You're right, changing the image or toggling superpowered doesn't restart the affected host container today. 😞

bcressey: There might be another edge case to consider - if settings for host-containerd are changed in the same transaction that disables a running host container, then `host-ctr` will exit after losing the connection, and not be restarted afterwards to clean up the running task.

One approach might be for `host-ctr` to have a cleanup mode so that it can be invoked by `host-containers` to delete the running task, if present.

That would handle the edge case and avoid the need to inspect the running container, because it wouldn't be running if the settings had changed.

packages/os/host-containers@.service
bcressey, line 4: `KillMode=mixed` in this unit file is misleading because there's only the one `host-ctr` process. Maybe a separate commit to clean that up, since it's our last use of that KillMode option.

sources/host-ctr/cmd/host-ctr/main.go
tjkirch: > Is that the intended workflow? @zmrow @tjkirch ?

(Just for historical perspective - it was *intended*, but we didn't *like it* :)  We didn't have the tools at the time to restart properly on those settings changes and considered it a weakness.)

etungsten:
Push above adds additional commits to address concerns about edge cases with host-ctr potentially rebinding to an out-of-date host container when both host-containers and host-containerd are restarted.

* A new clean-up mode has been added to `host-ctr`
* Refactored `host-ctr` to accommodate the new subcommand.
* `host-containers` (the rust binary) will now call `host-ctr clean-up` when appropriate to ensure restarted host-containers are running with the latest settings/configuration.
* Removes `KillMode=mixed` from host-containers@ units

etungsten:
Testing done for verifying the fix for the edge case:

`host-ctr` no longer rebinds to lingerering out-of-date host-containers when both host-containers and host-containerd are restarted.

<details>

```
bash-5.0# 
bash-5.0# ############# Disable host-containerd
bash-5.0# 
bash-5.0# systemctl restart host-containerd
bash-5.0# sleep 1
bash-5.0# 
bash-5.0# 
bash-5.0# ############# host-ctr loses connection and there is now a lingering control host container running
bash-5.0# 
bash-5.0# systemctl status host-containers@control
● host-containers@control.service - Host container: control
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/host-containers@.service;
 enabled; vendor preset: enabled)
     Active: activating (auto-restart) (Result: exit-code) since Mon 2020-12-07 06:35:21 UTC; 992ms ago
    Process: 3020 ExecStartPre=/usr/bin/mkdir -m 1777 -p ${LOCAL_DIR}/host-containers/control (code=exited, s
tatus=0/SUCCESS)
    Process: 3080 ExecStart=/usr/bin/host-ctr run -container-id=control -source=${CTR_SOURCE} -superpowered=$
{CTR_SUPERPOWERED} (code=exited, status=1/FAILURE)
   Main PID: 3080 (code=exited, status=1/FAILURE)

Dec 07 06:35:21 ip-192-168-28-193.us-west-2.compute.internal host-ctr[3080]: time="2020-12-07T06:35:21Z" leve
l=error msg="failed to cleanup container" error="cannot delete running task control: failed precondition"
Dec 07 06:35:21 ip-192-168-28-193.us-west-2.compute.internal systemd[1]: host-contain
ers@control.service: Main process exited, code=exited, status=1/FAILURE
Dec 07 06:35:21 ip-192-168-28-193.us-west-2.compute.internal systemd[1]: 
host-containers@control.service: Failed with result 'exit-code'.
bash-5.0# 
bash-5.0# ctr -a /run/host-containerd/containerd.sock task ls
TASK       PID     STATUS    
control    3739    RUNNING
admin      3823    RUNNING
bash-5.0# 
bash-5.0# ctr -a /run/host-containerd/containerd.sock container ls
CONTAINER    IMAGE                                                                                RUNTIME                  
admin        ecr.aws/arn:aws:ecr:us-west-2:328549459982:repository/bottlerocket-admin:v0.5.2      io.containerd.runc.v2    
control      ecr.aws/arn:aws:ecr:us-west-2:328549459982:repository/bottlerocket-control:v0.4.1    io.containerd.runc.v2    
bash-5.0# 
bash-5.0# 
bash-5.0# 
bash-5.0# ############# Disable host-containers@control through settings
bash-5.0# 
bash-5.0# apiclient -u /settings -m PATCH -d '{"host-containers": {"control": {"enabled": false}}}'
bash-5.0# apiclient -u /tx/commit_and_apply -m POST
["settings.host-containers.control.enabled"]
bash-5.0# sleep 3

bash-5.0# 
bash-5.0# 
bash-5.0# 
bash-5.0# ############# Restart commands succeeded 
bash-5.0# 
bash-5.0# journalctl -u apiserver -n 50
-- Logs begin at Mon 2020-12-07 06:30:11 UTC, end at Mon 2020-12-07 06:35:22 UTC. --
Dec 07 06:30:13 localhost systemd[1]: Starting Bottlerocket API server...
Dec 07 06:30:13 localhost apiserver[2487]: 06:30:13 [INFO] Starting server at /run/api.sock with 1 thread and
 datastore at /var/lib/bottlerocket/datastore/current
Dec 07 06:30:13 localhost systemd[1]: Started Bottlerocket API server.
Dec 07 06:30:13 localhost apiserver[2487]: 06:30:13 [INFO] Starting 1 workers
Dec 07 06:30:13 localhost apiserver[2487]: 06:30:13 [INFO] Starting "actix-web-service-"/run/api.sock"" servi
ce on "/run/api.sock" (pathname)
Dec 07 06:35:22 apiserver[5871]: 06:35:22 [INFO] thar-be-setting
s started
Dec 07 06:35:22 apiserver[5871]: 06:35:22 [INFO] Parsing stdin f
or updated settings
Dec 07 06:35:22 apiserver[5871]: 06:35:22 [INFO] Requesting affe
cted services for settings: {"settings.host-containers.control.enabled"}
Dec 07 06:35:22 apiserver[5871]: 06:35:22 [INFO] Restarting affe
cted services...
bash-5.0# sleep 5
bash-5.0# 
bash-5.0# 
bash-5.0# ############## host-containers@control is disabled at this point
bash-5.0# 
bash-5.0# systemctl status host-containers@control
● host-containers@control.service - Host container: control
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/host-containers@.service;
 disabled; vendor preset: enabled)
     Active: inactive (dead)

....
Dec 07 06:35:21 host-ctr[3080]: time="2020-12-07T06:35:21Z" leve
l=error msg="failed to get container task exit status" error="rpc error: code = Unavailable desc = transport 
is closing"
...
Dec 07 06:35:21 host-ctr[3080]: time="2020-12-07T06:35:21Z" level=error msg="failed to cleanup container" error="cannot delete running task control: failed precondition"
Dec 07 06:35:21 systemd[1]: host-containers@control.service: Main process exited, code=exited, status=1/FAILURE
Dec 07 06:35:21 systemd[1]: host-containers@control.service: Failed with result 'exit-code'.
Dec 07 06:35:22 systemd[1]: Stopped Host container: control.
bash-5.0# 
bash-5.0# 
bash-5.0# 
bash-5.0# ############## The lingering control host container got cleaned up by the restart command
bash-5.0# 
bash-5.0# ctr -a /run/host-containerd/containerd.sock task ls
TASK     PID     STATUS    
admin    3823    RUNNING
bash-5.0# 
bash-5.0# ctr -a /run/host-containerd/containerd.sock container ls
CONTAINER    IMAGE                                                                              RUNTIME                  
admin        ecr.aws/arn:aws:ecr:us-west-2:328549459982:repository/bottlerocket-admin:v0.5.2    io.containerd.runc.v2    
bash-5.0# 
bash-5.0# 
bash-5.0# ############## Re-enable control host-container and use a different image source for differentiation
bash-5.0# 
<459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-control:v0.4.0"}}}'
bash-5.0# apiclient -u /tx/commit_and_apply -m POST
["settings.host-containers.control.enabled","settings.host-containers.control.source"]
bash-5.0# sleep 5
bash-5.0# 
bash-5.0# 
bash-5.0# 
bash-5.0# ############# Restart commands succeeded 
bash-5.0# 
bash-5.0# journalctl -u apiserver -n 50
-- Logs begin at Mon 2020-12-07 06:30:11 UTC, end at Mon 2020-12-07 06:35:35 UTC. --
Dec 07 06:30:13 localhost systemd[1]: Starting Bottlerocket API server...
Dec 07 06:30:13 localhost apiserver[2487]: 06:30:13 [INFO] Starting server at /run/api.sock with 1 thread and datastore at /var/lib/bottlerocket/datastore/current
Dec 07 06:30:13 localhost systemd[1]: Started Bottlerocket API server.
Dec 07 06:30:13 localhost apiserver[2487]: 06:30:13 [INFO] Starting 1 workers
Dec 07 06:30:13 localhost apiserver[2487]: 06:30:13 [INFO] Starting "actix-web-service-"/run/api.sock"" service on "/run/api.sock" (pathname)
Dec 07 06:35:22 apiserver[5871]: 06:35:22 [INFO] thar-be-settings started
Dec 07 06:35:22 apiserver[5871]: 06:35:22 [INFO] Parsing stdin for updated settings
Dec 07 06:35:22 apiserver[5871]: 06:35:22 [INFO] Requesting affected services for settings: {"settings.host-containers.control.enabled"}
Dec 07 06:35:22 apiserver[5871]: 06:35:22 [INFO] Restarting affected services...
Dec 07 06:35:30 apiserver[6037]: 06:35:30 [INFO] thar-be-settings started
Dec 07 06:35:30 apiserver[6037]: 06:35:30 [INFO] Parsing stdin for updated settings
Dec 07 06:35:30 apiserver[6037]: 06:35:30 [INFO] Requesting affected services for settings: {"settings.host-containers.control.source", "settings.host-containers.cont
rol.enabled"}
Dec 07 06:35:30 apiserver[6037]: 06:35:30 [INFO] Restarting affected services...
bash-5.0# sleep 10
bash-5.0# 
bash-5.0# 
bash-5.0# 
bash-5.0# ############# control host container running again
bash-5.0# systemctl status host-containers@control
● host-containers@control.service - Host container: control
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/host-containers@.service; enabled; vendor preset: enabled)
     Active: active (running) since Mon 2020-12-07 06:35:30 UTC; 14s ago
    Process: 6070 ExecStartPre=/usr/bin/mkdir -m 1777 -p ${LOCAL_DIR}/host-containers/control (code=exited, status=0/SUCCESS)
   Main PID: 6088 (host-ctr)
      Tasks: 19 (limit: 9185)
     Memory: 42.5M
     CGroup: /system.slice/system-host\x2dcontainers.slice/host-containers@control.service
             └─6088 /usr/bin/host-ctr run -container-id=control -source=328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-control:v0.4.0 -superpowered=false

.....
bash-5.0# 
bash-5.0# 
bash-5.0# 
bash-5.0# ############# Notice that the new control host container is using the new image source.
bash-5.0# 
bash-5.0# ctr -a /run/host-containerd/containerd.sock task ls
TASK       PID     STATUS    
admin      3823    RUNNING
control    6209    RUNNING
bash-5.0# 
bash-5.0# ctr -a /run/host-containerd/containerd.sock container ls
CONTAINER    IMAGE                                                                                RUNTIME                  
admin        ecr.aws/arn:aws:ecr:us-west-2:328549459982:repository/bottlerocket-admin:v0.5.2      io.containerd.runc.v2    
control      ecr.aws/arn:aws:ecr:us-west-2:328549459982:repository/bottlerocket-control:v0.4.0    io.containerd.runc.v2    
bash-5.0# 
bash-5.0# #
```

</details>

etungsten:
I noticed while making these new changes that `host-containers`, the rust binary responsible for managing host container systemd units, unconditionally loops through all host containers to apply their current settings and tries to enables or disables each of their systemd unit. This means that I can't unconditionally try to clean-up host containers with `host-ctr` before `host-containers` attempt to enable any particular host-container unit, because it'll do it for every host container regardless of its current running state and whether it's settings have changed or not. 

`host-containers` (the binary) currently does not have the ability to detect settings changes and to only enact them upon impacted host containers. We can however implement this by making `host-containers` check whether the environment file its writing to for each container is being changed. If the environment file is changed, that means the host container that's being handled right now has to be restarted if it's currently running (enabled=true). If the environment file hasn't changed, then that means we don't need to do anything for the host container being handled. If the environment file does not exist, that means we're going through first boot and we should apply the settings and enact them unconditionally.

I did not try to implement this here because it's not strictly within the scope of this PR. But I can make an issue to follow up with this if people think the approach above is sensible. Please let me know what you think @tjkirch, @bcressey .

tjkirch:
^ I think it could be better to improve the way restart-commands are run so that they're always given the list of settings that have changed, rather than the setting name having to be in the restart-command itself.  That way we can handle dynamic settings, and not have to inspect the system to see what changed.

For example, right now, `services.motd` could have a restart-command like `my-command settings.motd` because we know `services.motd` is only associated with one setting, but `services.host-containers` is associated with any host container name under `settings.host-containers`, so we can't pass (hardcode) a single setting.  Instead, if the command was given the changed settings, it could alter its behavior based on what changed, like in this case where we need to know which host container changed.  (It could also give us the ability to re-use restart command helpers more often, if they can branch on setting.)

(I just mean this as a different potential follow-up, not as a blocker for this PR)

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -45,47 +46,60 @@ func init() {
...
+	fs.BoolVar(&superpowered, "superpowered", false, "specifies whether to create the container with `superpowered` privileges or not")
+	fs.StringVar(&containerID, "container-id", "", "the id of the container to manage")
+	if len(os.Args) < 2 {
+		fmt.Fprintln(os.Stderr, "expected one of the following subcommands: 'run', 'pull-image'")
+		flag.Usage()
+		os.Exit(2)
+	}
+	fs.Parse(os.Args[2:])
+
+	var subcommand = os.Args[1]
```
samuelkarp: I think that the built-in `flag` library isn't very well-suited to this subcommand-style logic.  While it'd increase the total code size to use a different library here, I think one like that will help reduce little errors that can creep in.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -45,47 +46,60 @@ func init() {
...
+	switch subcommand {
+	case "run":
+		os.Exit(runCtr(containerdSocket, namespace, containerID, source, superpowered))
+	case "pull-image":
+		os.Exit(pullImageOnly(containerdSocket, namespace, source))
+	default:
+		fmt.Errorf("expected one of the following subcommands: 'run', 'pull-image'")
```
samuelkarp: `fmt.Errorf` isn't the right approach here; that will just return an `error` but since it's not assigned it gets dropped on the floor.

`fmt.Fprintf` is what you were looking for here, but probably a better pattern is similar to what you had prior to this refactoring change:

```go
func main() {
	if err := _main(); err != nil {
		fmt.Fprintln(err)
		os.Exit(1)
	}
}

func _main() error {
	// main body here
}
```

samuelkarp (commented):
Can we separate the refactoring change from the rest of this PR?

etungsten:
> Can we separate the refactoring change from the rest of this PR?

I'm inclined to keep it as part of this PR since we're adding an additional "subcommand-like" functionality to `host-ctr` here. It feels like the right thing to do as opposed to having to add another flag that completely changes what `host-ctr` does. I very much regret adding `-pull-image-only` as a flag instead of a subcommand.

Hopefully the refactoring changes aren't too controversial and produce too much churn. It's mostly refactoring stuff into functions.

samuelkarp:
> I'm inclined to keep it as part of this PR since we're adding an additional "subcommand-like" functionality to host-ctr here. It feels like the right thing to do as opposed to having to add another flag that completely changes what host-ctr does. I very much regret adding -pull-image-only as a flag instead of a subcommand.
>
> Hopefully the refactoring changes aren't too controversial and produce too much churn. It's mostly refactoring stuff into functions.

It might be easier to pull it out into a separate PR that we merge ahead of this one.  I don't expect breaking into functions to be controversial, but I'm expecting that I won't be the only one with an opinion on the subcommand implementation and it might reduce churn/rebasing effort for you to do it that way.  And separating the functional changes (proper restart behavior) from the refactor should make both easier to review.

etungsten:
This PR now depends on https://github.com/bottlerocket-os/bottlerocket/pull/1235 being merged before proceeding. Will rebase once it does.

etungsten:
Push above rebases upon develop to pull in the refactor for host-ctr.

Tested things and the things still work as expected, the previous test results still stands.

zmrow (approved):
:medal_military:

samuelkarp (approved):

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -173,65 +188,73 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
...
-		return err
+		// Set up the container spec. See `withSuperpowered` for conditional options
+		// set when configured as superpowered.
+		ctrOpts := containerd.WithNewSpec(
+			oci.WithImageConfig(img),
+			oci.WithHostNamespace(runtimespec.NetworkNamespace),
+			oci.WithHostHostsFile,
+			oci.WithHostResolvconf,
+			// Launch the container under the systemd unit's cgroup
+			oci.WithCgroup(cgroupPath),
```
bcressey, line 219: Was this the reason behind `KillMode=mixed` in the unit file - so they'd share a cgroup and systemd could clean them up together?

etungsten: Ah, you're right. I had forgotten. It's so that the host container task will get cleaned up if `host-containers@.service` gets stopped by systemd.

I'll go ahead drop the commit that removed `KillMode=mixed`

etungsten:
Push above drops the commit for removing `KillMode=mixed` from the host containers unit files.

sources/host-ctr/cmd/host-ctr/main.go
etungsten: That doesn't necessarily mean the container resources are cleaned up though (the container itself, and its snapshot in ctrd), so we still the need the host-ctr clean-up mechanism.

bcressey (approved):

sources/host-ctr/cmd/host-ctr/main.go
bcressey: It seems like the cgroup piece isn't working as intended, but that's not a new problem.

We can fix that and clean up `KillMode=mixed` again at the same time. My concern is really that we'd "fix" this code so the tasks landed in the `host-containers@...` cgroup, and then a service restart would cause them to be killed again.

PR #1233, opened 2020-12-06
tools: Fix message in the partition size check condition

pranavek:
**Description of changes:**
Fix message in the partition size check condition. The size was doubled earlier, however, the message was not changed accordingly.

**Terms of contribution:**


tjkirch (approved):
Thank you for catching this and submitting a fix!

webern (approved):

PR #1235, opened 2020-12-09
host-ctr: refactoring

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
Date:   Fri Dec 4 18:34:07 2020 -0800

    host-ctr: refactoring
    
    Refactors `host-ctr`.
    Organizes functionality into subcommands.

```

**Testing done:**
Build image and launched several instances and all host containers came up fine. kubelet running fine, meaning pause container got pulled successfully.

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"flag"
 	"io/ioutil"
 	"math/rand"
 	"os"
@@ -25,6 +24,7 @@ import (
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
 )
 
 // Expecting to match ECR image names of the form:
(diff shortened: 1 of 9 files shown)


sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -45,47 +45,99 @@ func init() {
...
 func main() {
-	os.Exit(_main())
+	if err := _main(); err != nil {
+		log.L.Fatalf("%v", err)
+	}
 }
 
-func _main() int {
-	// Parse command-line arguments
+func _main() error {
```
jahkeup: Instead of returning the result of having called the command, the `cli.App` type should be returned for direct use in testing and have that `cli.App.Run` called from the real `main`. This gives a little more control to testing entrypoints where `Run` can optionally be called.

If that were changed, I'd rename this to `func App() cli.App` (`App` because it stands out a bit more than `app`, not because it needs to exported and because I personally like it :P)

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -45,47 +45,99 @@ func init() {
...
+	// Command-line arguments
 	var (
 		containerID      string
 		source           string
 		containerdSocket string
 		namespace        string
 		superpowered     bool
-		pullImageOnly    bool
 	)
```
jahkeup, line 63: I want to say change this to not use predeclared vars, but I can't come up with a good reason right now to make it otherwise (using `cli.Context.String()` and friends instead). I'm in favor of leaving this as-is for now (Sam and I have debated this before ourselves, I'm ambivalent).

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -98,172 +150,177 @@ func _main() int {
...
 		}
-		ref = ecrRef.Canonical()
-		log.G(ctx).
-			WithField("source", source).
-			WithField("ref", ref).
-			Debug("Parsed ECR reference from URI")
 	}
 
-	img, err := pullImage(ctx, ref, client)
+	// Check if the target container already exists. If it does, take over the helm to manage it.
```
samuelkarp: Can we separate the functional changes from the refactoring changes?

sources/host-ctr/cmd/host-ctr/main.go
samuelkarp: I like the predeclared vars.

sources/host-ctr/cmd/host-ctr/main.go
etungsten: Oh oops. I totally forgot that this came along with my cherry pick.

etungsten:
Push above removes some of the changes I accidentally picked up from https://github.com/bottlerocket-os/bottlerocket/pull/1230

Tested things and they still work

Also addresses @jahkeup's comments.

samuelkarp (approved):

sources/host-ctr/cmd/host-ctr/main.go
jahkeup: > I like the predeclared vars.

I know, but it's less clear in these circumstances because the two subcommands are each using them. There's not a lot of them and so the scope remains clear for now - I would push on this if there were more at play. For now, I think its reasonable and good to move forward with.

jahkeup (approved):

PR #1242, opened 2020-12-14
host-ctr: remove KillMode=mixed and don't create the container task under the unit's cgroup

### LINKED ISSUE #1237: host-ctr: don't try to create the container under the systemd unit's cgroup
>It seems like the cgroup piece isn't working as intended, but that's not a new problem.

>We can fix that and clean up `KillMode=mixed` again at the same time. My concern is really that we'd "fix" this code so the tasks landed in the `host-containers@...` cgroup, and then a service restart would cause them to be killed again.

_Originally posted by @bcressey in https://github.com/bottlerocket-os/bottlerocket/pull/1230#discussion_r540597593_

We should remove https://github.com/bottlerocket-os/bottlerocket/blob/294f7af9e3e9992183e549759c8bd07a611e2c89/sources/host-ctr/cmd/host-ctr/main.go#L218-L219 since the container task doesn't end up in the systemd unit's cgroup anyway and fixing it would introduce problems. We're handling the service restarts ourselves now.

etungsten:
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/1237

**Description of changes:**
```
Date:   Fri Dec 4 14:31:26 2020 -0800

    host-containers@: remove KillMode=mixed
    
    We don't need systemd to go and actively try kill all processes
    of the unit's cgroup.

```

```
Date:   Mon Dec 14 13:47:04 2020 -0800

    host-ctr: don't create the container task under the service's cgroup
    
    We no longer need to launch the host container task under the systemd
    service unit's cgroup. We're managing the life-cycle of the host
    container directly without the help of systemd
```

**Testing done:**
Built aws-k8s-1.17 x86 ami and launched instance.
Restarting `host-containerd` doesn't kill any running host-containers as expected.

**Terms of contribution:**

### DIFF
--- a/packages/os/host-containers@.service
+++ b/packages/os/host-containers@.service
@@ -16,7 +16,6 @@ ExecStart=/usr/bin/host-ctr run \
 Restart=always
 RestartSec=45
 TimeoutStopSec=60
-KillMode=mixed
 StandardError=journal+console
 
 [Install]
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -20,7 +20,6 @@ import (
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/oci"
-	"github.com/opencontainers/runc/libcontainer/cgroups"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -201,13 +200,6 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 
 	// If the container doesn't already exist, create it
 	if container == nil {
-		// Get the cgroup path of the systemd service
-		cgroupPath, err := cgroups.GetOwnCgroup("name=systemd")
-		if err != nil {
-			log.G(ctx).WithError(err).Error("failed to discover systemd cgroup path")
-			return err
-		}
-
 		// Set up the container spec. See `withSuperpowered` for conditional options
 		// set when configured as superpowered.
 		ctrOpts := containerd.WithNewSpec(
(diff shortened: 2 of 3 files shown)


samuelkarp (approved):

bcressey:
Which cgroup are the tasks created under now? Still host-containerd's, or something else?

etungsten:
> Which cgroup are the tasks created under now? Still host-containerd's, or something else?

They're all launched under host-containerd's cgroup:
```
● host-containerd.service - containerd runtime for host containers
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/host-containerd.service; enabled; vendor preset: enabled)
     Active: active (running) since Tue 2020-12-15 06:13:18 UTC; 1min 8s ago
       Docs: https://containerd.io
   Main PID: 2928 (containerd)
      Tasks: 34
     Memory: 685.3M
     CGroup: /system.slice/host-containerd.service
             ├─2928 /usr/bin/containerd --config /etc/host-containerd/config.toml
             ├─3660 /x86_64-bottlerocket-linux-gnu/sys-root/usr/bin/containerd-shim-runc-v2 -namespace default -id control -address /run/host-containerd/containerd.sock
             ├─3736 /usr/bin/amazon-ssm-agent
             ├─3836 /x86_64-bottlerocket-linux-gnu/sys-root/usr/bin/containerd-shim-runc-v2 -namespace default -id admin -address /run/host-containerd/containerd.sock
             ├─3858 /usr/sbin/sshd -e -D
             ├─3961 sshd: ec2-user [priv]
             ├─3972 sshd: ec2-user@pts/0
             ├─3974 -bash
             ├─4017 sudo sheltie
             ├─4018 nsenter -t 1 -a /proc/4017/root/opt/bin/bash
             ├─4019 /proc/4017/root/opt/bin/bash
             ├─4401 systemctl status host-containerd
             └─4402 more

```

zmrow (approved):
:volleyball:

bcressey (approved):

PR #1246, opened 2020-12-17
Support for awsvpc networking mode in ECS

### LINKED ISSUE #1077: [ECS] awsvpc mode
This issue is tracking adding support for the [awsvpc network mode](﻿https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-networking.html﻿) to the ECS variant of Bottlerocket.

This will require adding the [ECS CNI plugin](﻿https://github.com/aws/amazon-ecs-cni-plugins/﻿).

samuelkarp:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket/issues/1077

**Description of changes:**
This pull request adds support for the `awsvpc` networking mode to the Bottlerocket `aws-ecs-1` variant.

The `awsvpc` networking mode is enabled through the use of several [CNI plugins](https://github.com/containernetworking/cni) that are used in combination.  While the CNI plugins are stored in two separate git repositories ([amazon-ecs-cni-plugins](https://github.com/aws/amazon-ecs-cni-plugins) and [amazon-vpc-cni-plugins](https://github.com/aws/amazon-vpc-cni-plugins)), they are effectively tightly coupled to the ECS agent; the ECS agent repository [includes the plugins as git submodules](https://github.com/aws/amazon-ecs-agent/blob/master/.gitmodules) and is only ever tested with the exact versions of the plugins included as submodules.  Because of this tight coupling, the CNI plugins are included as part of the `ecs-agent` spec file; when any one of these source files are updated all should be.

**Testing done:**
* `cargo make -e BUILDSYS_VARIANT=aws-ecs-1`
* Built an AMI, ensured that the files were installed to the expected locations, and ran each of the plugins with the `-version` argument
* Ran a task with `awsvpc` mode on a `t2.micro`
* Ran 10 tasks with `awsvpc` mode on a `c5.large`, exercising [ENI trunking](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-instance-eni.html)
* Ran ECS qualification test suite, including newly-enabled tests for awsvpc and IPv6 support

**Terms of contribution:**


packages/ecs-agent/0001-TEST-PATCH.patch
```diff
@@ -0,0 +1,22 @@
...
+---
+ README.md | 2 ++
+ 1 file changed, 2 insertions(+)
+
+diff --git a/README.md b/README.md
+index 0b95444..0380e28 100644
+--- a/README.md
++++ b/README.md
+@@ -1,3 +1,5 @@
++TEST PATCH!
```
jahkeup: 🧪

packages/ecs-agent/ecs-agent.spec
```diff
@@ -1,31 +1,43 @@
...
+Source104: ecs.config
+Source105: pause-image-VERSION
+Source106: pause-config.json
+Source107: pause-manifest.json
+Source108: pause-repositories
+
+# Patches are numbered according to which source they apply to
+# Patches 0000 - 0999 apply to Source0
+# Patches 1000 - 1999 apply to Source1
+# Patches 2000 - 2999 apply to Source2
```
srgothi92, line 48: this comment nicely describes how patches are numbered, however `Source2` does not exist, would it confuse the reader ?

samuelkarp:
> samuelkarp force-pushed the samuelkarp:awsvpc branch from 4941409 to d878f22 

This push rebases on top of the current tip of `develop` and updates the CNI plugins to use the versions included in ECS agent 1.50.2's submodules.

samuelkarp:
> samuelkarp force-pushed the samuelkarp:awsvpc branch from d878f22 to e07bbc9

* Enabled awsvpc mode in the ECS agent
* Fixed filesystem paths in the ECS agent for non-containerized operation
* Fixed filesystem paths in the CNI plugins for non-containerized operation
* Disabled unsupported CNI plugins (App Mesh and branch ENIs) for now
* Successfully ran a task using awsvpc mode on a t2.micro.

packages/ecs-agent/ecs-agent.spec
samuelkarp: `Source2` will be added before this PR is ready to merge.

packages/ecs-agent/0005-bottlerocket-fix-procfs-path-on-host.patch
```diff
@@ -0,0 +1,25 @@
+From aa1f344f6a13f1533a1eb93d7de419739bf3b8c3 Mon Sep 17 00:00:00 2001
+From: Samuel Karp <skarp@amazon.com>
+Date: Fri, 26 Mar 2021 17:48:28 -0700
+Subject: [PATCH 5/5] bottlerocket: fix procfs path on host
+
```
jahkeup, line 5: This is because we're not running the agent in a container, right? Can you add a comment to this patch to explain why the path is fixed up this way?

packages/ecs-agent/Cargo.toml
```diff
@@ -13,10 +13,17 @@ path = "pkg.rs"
...
-# the source of those plugins.
+# the source of those plugins.  Instead, we include the CNI plugin source
+# directly.
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/aws/amazon-ecs-cni-plugins/archive/55b2ae77ee0bf22321b14f2d4ebbcc04f77322e1/amazon-ecs-cni-plugins.tar.gz"
+sha512 = "42e2dad0741f02a3b0d675c0f512557a118cef6df7e9d989a35c73d02484912d800453148abbf6d8424c5e40d2662745a9aef4a68cb2f610d9defc976e394f1f"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/aws/amazon-vpc-cni-plugins/archive/a21d3a41f922e14c19387713df66be3e4ee1e1f6/amazon-vpc-cni-plugins.tar.gz"
+sha512 = "86011a1d7a8d4a2fe824c9eea6ad2ef9394059475ffa421c3636afd4f4031d60898992022e34ce2b1e9594fa79766f747a6c239e26be5d71fe96addc46bf7f54"
```
jahkeup, line 26: Is it possible to break the ecs-agent and the CNI plugins into two packages? Are there cross dependencies in the code itself?

I'd prefer to have a simpler build of the two where patches and such were well scoped to the agent and CNI plugins over having them combined and do a dance to build them together in the same package.

packages/ecs-agent/0005-bottlerocket-fix-procfs-path-on-host.patch
samuelkarp: Yes, can add a comment.

packages/ecs-agent/Cargo.toml
samuelkarp: I thought about this a lot as I was working on this.  While the CNI plugins exist in a separate git repository, the reality of how they're used in ECS is that they're incredibly tightly-coupled; there is no support for running versions of the CNI plugins that are out of sync with the versions expected to be used by the ECS agent (controlled via git submodules in the ECS agent source repository).  While the overall spec is more complicated with the patching work, the approach here makes it very obvious when one changes without the other and my hope is that it will be an overall improvement in our ability to effectively maintain the ECS agent + the plugins over time.

packages/ecs-agent/ecs-agent.spec
samuelkarp: The latest push now includes `Source2`.

packages/ecs-agent/ecs-agent.spec
```diff
@@ -48,23 +76,92 @@ Requires: %{_cross_os}iptables
...
-cp %{S:9} "agent/version/version.go"
+cp %{S:109} "agent/version/version.go"
+
+# Extract Source1, which has a top-level directory of
+# %{ecscni_gorepo}-%{ecscni_gitrev}
+# -T: Do not perform default archive unpack (i.e., skip Source0)
+# -D: Do not delete directory before unpacking sources (i.e., don't delete
+#     unpacked Source0)
+# -a: Unpack after changing into the directory
+# -q: Unpack quietly
```
webern, line 118: Nice, thank you for documenting this.

webern (approved):

packages/ecs-agent/ecs-agent.spec
```diff
@@ -48,23 +76,92 @@ Requires: %{_cross_os}iptables
...
+%cross_go_setup %{name}-%{version}/%{ecscni_gorepo}-%{ecscni_gitrev} %{ecscni_goproject} %{ecscni_goimport}
+# Symlink amazon-vpc-cni-plugins-%{vpccni_gitrev} to the GOPATH location
+%cross_go_setup %{name}-%{version}/%{vpccni_gorepo}-%{vpccni_gitrev} %{vpccni_goproject} %{vpccni_goimport}
 
 %build
+pushd $(pwd)
 # Build the agent
-%cross_go_configure %{goimport}
+# cross_go_configure cd's to the correct GOPATH location
+%cross_go_configure %{agent_goimport}
```
jahkeup, line 158: Are the `pushd` and `popd` working around a limitation in `%cross_go_configure` where it can't `cd` to the right path *unless* it is at the top of the unpacked tree?

I'd prefer to capture that path and later `cd` *directly* to the path. This avoids later edits that would need to consider how they might affect working directory navigation.

```specfile
%build
BUILD_TOP=$(pwd -P)

# build a
cd $BUILD_TOP
%cross_go_configure %{a_goimport}

# build a/child/artifact
cd child/artifact
%make

# build b
cd $BUILD_TOP
%cross_go_configure %{b_goimport}

...

cd $BUILD_TOP
```

jahkeup (commented):
Very cool! 

I'm gonna build and test this out myself before hitting the green button 🔴 🟡 🟢 ☁️ 🚀 🐳

packages/ecs-agent/ecs-agent.spec
samuelkarp: > Are the `pushd` and `popd` working around a limitation in `%cross_go_configure` where it can't `cd` to the right path unless it is at the _top_ of the unpacked tree?

Yes

> I'd prefer to capture that path and later `cd` _directly_ to the path. This avoids later edits that would need to consider how they might affect working directory navigation.

I like that suggestion.  I'll make sure it works and then update here.

packages/ecs-agent/ecs-agent.spec
```diff
@@ -48,23 +76,92 @@ Requires: %{_cross_os}iptables
...
+# -a: Unpack after changing into the directory
+# -q: Unpack quietly
+# See http://ftp.rpm.org/max-rpm/s1-rpm-inside-macros.html
+%setup -T -D -a 2 -q
+# Change to the directory that we unpacked
+cd %{vpccni_gorepo}-%{vpccni_gitrev}
+# Set up git so we can apply patches
+# This is included in autosetup, but not autopatch
+%__scm_setup_git
+# Apply patches from 1000 to 1999
```
samuelkarp:
```suggestion
# Apply patches from 2000 to 2999
```

Bad copy-paste, will fix in the next push

packages/ecs-agent/ecs-agent.spec
```diff
@@ -48,23 +76,92 @@ Requires: %{_cross_os}iptables
...
-%autosetup -Sgit -n %{gorepo}-%{gover} -p1
-%cross_go_setup %{gorepo}-%{gover} %{goproject} %{goimport}
+# After prep runs, the directory setup looks like this:
+# %{_builddir} [root]
+# └── %{name}-%{version} [created by setup]
+#     ├── amazon-ecs-agent-%{agent_gover} [top level of Source0]
+#     │   └── [unpacked sources]
+#     ├── amazon-ecs-cni-plugins-%{ecscni_gitrev} [top level of Source1]
+#     │   └── [unpacked sources]
+#     ├── amazon-vpc-cni-plugins-%{vpccni_gitrev} [top level of Source1]
```
samuelkarp:
```suggestion
#     ├── amazon-vpc-cni-plugins-%{vpccni_gitrev} [top level of Source2]
```

Bad copy-paste, will fix in the next push.

packages/ecs-agent/ecs-agent.spec
```diff
@@ -84,30 +181,108 @@ go build -a \
...
+%cross_scan_attribution go-vendor go-vendor
 
 %files
+# License and attribution files are installed into /usr/share/licenses with a
+# directory structure as follows:
+# /usr/share/licenses/ecs-agent/
+# ├── attribution.txt
+# ├── LICENSE
+# ├── LICENSE.amazon-ecs-cni-plugins
+# ├── NOTICE
```
samuelkarp:
```suggestion
# ├── LICENSE
# ├── LICENSE.amazon-ecs-cni-plugins
# ├── LICENSE.amazon-vpc-cni-plugins
# ├── NOTICE
```

Will update with the next push.

packages/ecs-agent/0005-bottlerocket-fix-procfs-path-on-host.patch
samuelkarp: Done

packages/ecs-agent/ecs-agent.spec
samuelkarp: Done

bcressey (approved):

jahkeup (approved):
Built and ran an instance for myself, looks great 👍🏽

PR #1248, opened 2020-12-17
containerd: set version and revision strings

### LINKED ISSUE #1241: Bottlerocket containerd runtime version has unknown string
When pulling the containerd runtime from EKS 1.18 variant, an "unknown" string is appended to version that gets returned as shown below

Container runtime
containerd://1.3.7+unknown

**Image I'm using:**
Bottlerocket v1.18 EKS

Should return containerd version without "unknown"

View Bottlerocket node group in EKS console, or use kubelet to pull runtime version directly

samuelkarp:
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/1241

**Description of changes:**
Set version and revision strings for containerd

**Testing done:**
```
bash-5.0# ctr -a /run/dockershim.sock version
Client:
  Version:  1.3.7+bottlerocket
  Revision: 8fba4e9a7d01810a393d5d25a3621dc101981175

Server:
  Version:  1.3.7+bottlerocket
  Revision: 8fba4e9a7d01810a393d5d25a3621dc101981175
  UUID: 5278435f-2098-4f42-8ee1-d55ec6f511ec
```

```
$ kubectl get nodes -o custom-columns=NAME:.metadata.name,OS-IMAGE:status.nodeInfo.osImage,CONTAINER-RUNTIME:status.nodeInfo.containerRuntimeVersion
NAME                                           OS-IMAGE                CONTAINER-RUNTIME
ip-192-168-58-191.us-west-2.compute.internal   Bottlerocket OS 1.0.4   containerd://1.3.7+bottlerocket
```

**Terms of contribution:**

### DIFF
--- a/packages/containerd/containerd.spec
+++ b/packages/containerd/containerd.spec
@@ -4,6 +4,7 @@
 
 %global gover 1.3.7
 %global rpmver %{gover}
+%global gitrev 8fba4e9a7d01810a393d5d25a3621dc101981175
 
 %global _dwz_low_mem_die_limit 0
 
@@ -65,6 +66,8 @@ Requires: %{_cross_os}systemd
 %build
 %cross_go_configure %{goimport}
 export BUILDTAGS="no_btrfs seccomp selinux"
+export LD_VERSION="-X github.com/containerd/containerd/version.Version=%{gover}+bottlerocket"
+export LD_REVISION="-X github.com/containerd/containerd/version.Revision=%{gitrev}"
 for bin in \
   containerd \
   containerd-shim \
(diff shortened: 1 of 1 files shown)


tjkirch (approved):
Is the `+suffix` traditional/expected? Do we need it at all?

webern (approved):

samuelkarp:
> Is the `+suffix` traditional/expected? Do we need it at all?

I had reached out to one of the containerd maintainers to ask about their preferences here.  The output of `ctr version` is helpful for them to know what version of the code is in-use in the context of a bug report, and an adjusted output here lets them know that there are differences between the upstream code and the version exhibiting a bug.  Normally, their preference is for the `Release` field to be adjusted to indicate the patches applied (for example, an RPM NVR so they can find the SRPM) but in the context of Bottlerocket we have a single version for the whole OS and they thought a `+bottlerocket` suffix was sufficient.

bcressey (approved):

PR #1265, opened 2021-01-07
Update to systemd v247 and dbus-broker 26

etungsten:
**Issue number:**
N/A

**Description of changes:**

- Updates systemd to v247
https://github.com/systemd/systemd/blob/v247/NEWS

systemd-oomd is a new thing that's in preview phase. I decided to exclude it for now as it was throwing errors in the system journal:

<details>

```
Jan 07 07:27:29 systemd[1]: systemd-oomd.service: Scheduled restart job, restart counter is at 2.
Jan 07 07:27:29 systemd[1]: Stopped Userspace Out-Of-Memory (OOM) Killer.
Jan 07 07:27:29 systemd[1]: Starting Userspace Out-Of-Memory (OOM) Killer...
Jan 07 07:27:29 systemd-oomd[359]: Requires swap to operate
Jan 07 07:27:29 systemd[1]: systemd-oomd.service: Main process exited, code=exited, status=1/FAILURE
Jan 07 07:27:29 systemd[1]: systemd-oomd.service: Failed with result 'exit-code'.
Jan 07 07:27:29 systemd[1]: Failed to start Userspace Out-Of-Memory (OOM) Killer.
Jan 07 07:27:29 storewolf[357]: 07:27:29 [INFO] Datastore populated
Jan 07 07:27:29 systemd[1]: Finished Datastore creator.
Jan 07 07:27:29 systemd[1]: systemd-oomd.service: Scheduled restart job, restart counter is at 3.
Jan 07 07:27:29 systemd[1]: Starting Bottlerocket API server...
Jan 07 07:27:29 systemd[1]: Stopped Userspace Out-Of-Memory (OOM) Killer.
Jan 07 07:27:29 systemd[1]: Starting Userspace Out-Of-Memory (OOM) Killer...
Jan 07 07:27:29 kernel: random: crng init done
Jan 07 07:27:29 kernel: random: 7 urandom warning(s) missed due to ratelimiting
Jan 07 07:27:29 systemd[1]: Finished Load/Save Random Seed.
Jan 07 07:27:29 systemd[1]: Reached target First Boot Complete.
Jan 07 07:27:29 systemd[1]: Starting Commit a transient machine-id on disk...
Jan 07 07:27:29 systemd-oomd[363]: Requires swap to operate
Jan 07 07:27:29 systemd[1]: systemd-oomd.service: Main process exited, code=exited, status=1/FAILURE
Jan 07 07:27:29 systemd[1]: systemd-oomd.service: Failed with result 'exit-code'.
Jan 07 07:27:29 systemd[1]: Failed to start Userspace Out-Of-Memory (OOM) Killer.
```
</details>

With regards to the udev event changes, the only udev rule we're maintaining is for ephemeral storage and it's still compatible with systemd  v247's changes.

No other systemd changes should impact us.

I updated the 9003 patch for systemd to accommodate new version changes, namely https://github.com/systemd/systemd/commit/53171c0453b564fa0bdc59da9e1bd9060575cf59.

I added a new patch to restore `prefix=@prefix@` in systemd's pkg-config to allow prefix to be overridable again so that we don't install things under `/usr` directly.

- Updates dbus-broker to v25

**Testing done:**
Launched aws-k8s-1.17 x86_64 and aws-k8s-1.18 aarch64 nodes. 
Both nodes are able to run pods OK.
System journal shows no abnormalities. 
I also confirmed that I was able to open the system journal produced by v247 with older versions of systemd.

**Terms of contribution:**

### DIFF
--- a/packages/dbus-broker/dbus-broker.spec
+++ b/packages/dbus-broker/dbus-broker.spec
@@ -1,5 +1,5 @@
 Name: %{_cross_os}dbus-broker
-Version: 24
+Version: 26
 Release: 1%{?dist}
 Summary: D-BUS message broker
 License: Apache-2.0
--- a/packages/dbus-broker/Cargo.toml
+++ b/packages/dbus-broker/Cargo.toml
@@ -9,8 +9,8 @@ build = "build.rs"
 path = "pkg.rs"
 
 [[package.metadata.build-package.external-files]]
-url = "https://github.com/bus1/dbus-broker/releases/download/v24/dbus-broker-24.tar.xz"
-sha512 = "7fbe6c689eff27ec842bb9d839a418abc356b026bb0a54dfa8b680a655409aa1ea4cb90655a6b04561a88c6e703cacf8800fddfe1abc21e7ee60db2dac1c2db9"
+url = "https://github.com/bus1/dbus-broker/releases/download/v26/dbus-broker-26.tar.xz"
+sha512 = "51062edba619cdd7242c89422cf232f269feda2ad2fa221f66b09a616a3a76b90b25646b9b7f548290cddfbdac0e0da0c7347270ac3807cae2db056a1d4f480d"
 
 [build-dependencies]
 glibc = { path = "../glibc" }
--- a/packages/systemd/9003-repart-always-use-random-UUIDs.patch
+++ b/packages/systemd/9003-repart-always-use-random-UUIDs.patch
@@ -1,4 +1,4 @@
-From 42b00096598978558b8abb3e03daab9e3315b98d Mon Sep 17 00:00:00 2001
+From d80f5878542dda9389fc5c491288a2e6b35a0c32 Mon Sep 17 00:00:00 2001
 From: Ben Cressey <bcressey@amazon.com>
 Date: Thu, 16 Apr 2020 15:10:41 +0000
 Subject: [PATCH 9003/9006] repart: always use random UUIDs
@@ -18,15 +18,15 @@ and disk UUIDs.
 Signed-off-by: Ben Cressey <bcressey@amazon.com>
 ---
  meson.build            |   3 +-
- src/partition/repart.c | 101 ++++++-----------------------------------
- 2 files changed, 14 insertions(+), 90 deletions(-)
+ src/partition/repart.c | 103 ++++++-----------------------------------
+ 2 files changed, 14 insertions(+), 92 deletions(-)
 
 diff --git a/meson.build b/meson.build
-index dbbddb6..09fe7b6 100644
+index f406d595e6..c16b50c1e8 100644
 --- a/meson.build
 +++ b/meson.build
-@@ -1357,8 +1357,7 @@ substs.set('DEFAULT_LLMNR_MODE', default_llmnr)
- 
+@@ -1374,8 +1374,7 @@ substs.set('DEFAULT_LLMNR_MODE', default_llmnr)
+
  want_repart = get_option('repart')
  if want_repart != 'false'
 -        have = (conf.get('HAVE_OPENSSL') == 1 and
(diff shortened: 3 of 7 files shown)


etungsten:
Moving PR to draft mode as there are some unanswered questions about how the udev rules changes impacts us.

packages/dbus-broker/dbus-broker.spec
```diff
@@ -53,11 +53,11 @@ install -p -m 0644 %{S:3} %{buildroot}%{_cross_sysusersdir}/dbus.conf
...
 %{_cross_datadir}/dbus-1/*
-%{_cross_journalcatalogdir}/dbus-broker.catalog
-%{_cross_journalcatalogdir}/dbus-broker-launch.catalog
+%{_prefix}/lib/systemd/catalog/dbus-broker.catalog
+%{_prefix}/lib/systemd/catalog/dbus-broker-launch.catalog
 %{_cross_sysusersdir}/dbus.conf
 %{_cross_unitdir}/dbus-broker.service
 %{_cross_unitdir}/dbus.socket
-%exclude %{_cross_userunitdir}/dbus-broker.service
+%exclude %{_prefix}/lib/systemd/user/dbus-broker.service
```
bcressey: The previous macros are correct; we need to make sure we install files under the sys-root path rather than directly under `/usr`.

Looking at dbus-broker's meson.build file, it uses pkg-config to find these paths from systemd.pc. I'd guess something changed in the systemd update, and I expect we'll have to fix it there.

In the `build/rpms` directory, you can use these commands to inspect the file:
```
rpm2cpio bottlerocket-aarch64-systemd-devel-247-1.aarch64.rpm |cpio -idmv
cat ./aarch64-bottlerocket-linux-gnu/sys-root/usr/lib/pkgconfig/systemd.pc
```

The current file has these lines:
```
prefix=/aarch64-bottlerocket-linux-gnu/sys-root/usr
root_prefix=/aarch64-bottlerocket-linux-gnu/sys-root/usr
rootprefix=${root_prefix}
```

If the prefix is just `/usr` now then that would explain this behavior.

bcressey:
> Moving PR to draft mode as there are some unanswered questions about how the udev rules changes impacts us.

In ephemeral-storage.rules we're already using the recommended style of header guard:
```
ACTION=="remove", GOTO="ephemeral_storage_end"
```

As far as I know the only other udev rules we ship come from the systemd package. I'd expect those have all been updated but it wouldn't hurt to run `head /usr/lib/udev/rules.d/*.rules` and look at the first `ACTION` line.

etungsten:
Push above rebases onto develop to pull in latest package updates and aws-k8s-1.19 variant

packages/dbus-broker/dbus-broker.spec
etungsten: prefix is hard-coded to `/usr` in v247. see https://github.com/systemd/systemd/commit/6e65df89c348242dbd10036abc7dd5e8181cf733 :facepalm:

etungsten: This has been partially reverted in upstream main branch https://github.com/systemd/systemd/commit/60bce7c6d9606185114df1bdcd5ea100407688b8.

bcressey: I appreciate upstream's point but it doesn't really apply to us: we ensure `/usr` points to the sys-root path at runtime, but for our own packaging we need to be careful to avoid dependencies on the host OS.

Either we should just do a straight revert patch of https://github.com/systemd/systemd/commit/6e65df89c348242dbd10036abc7dd5e8181cf733, or else add https://github.com/systemd/systemd/commit/60bce7c6d9606185114df1bdcd5ea100407688b8 and add a local patch to restore `prefix=@prefix@` (if needed, let's try it without).

etungsten: I'll go ahead and make a revert patch of https://github.com/systemd/systemd/commit/6e65df89c348242dbd10036abc7dd5e8181cf733.

etungsten:
Push above addresses https://github.com/bottlerocket-os/bottlerocket/pull/1265#discussion_r554238856

etungsten:
>I'd expect those have all been updated but it wouldn't hurt to run head /usr/lib/udev/rules.d/*.rules and look at the first ACTION line.

Here is a gist of `head /usr/lib/udev/rules.d/*.rules`: https://gist.github.com/etungsten/270558834de44668bd8e5d8adf881304
They look fine to me.

The only one we maintain ourselves is `ephemeral-storage.rules` which is following the new guidance for compatibility.

etungsten:

bcressey (approved):

packages/dbus-broker/dbus-broker.spec
```diff
@@ -1,5 +1,5 @@
 Name: %{_cross_os}dbus-broker
-Version: 24
+Version: 25
```
tjkirch: Should we go with 26?

packages/systemd/9007-Revert-pkg-config-prefix-is-not-really-configurable-.patch
```diff
@@ -0,0 +1,84 @@
+From ed49352fa7cf7d9c2e371849dca3ebf906e800bc Mon Sep 17 00:00:00 2001
+From: Erikson Tung <etung@amazon.com>
+Date: Mon, 11 Jan 2021 11:01:57 -0800
+Subject: [PATCH] Revert "pkg-config: prefix is not really configurable, don't
```
tjkirch: It looks like we could get what we need from upstream's revert rather than doing it ourselves, possibly with the addition of a small `prefix=@prefix@` patch, if that's needed.  https://github.com/systemd/systemd/commit/60bce7c6d9606185114df1bdcd5ea100407688b8

tjkirch (approved):

packages/systemd/9007-Revert-pkg-config-prefix-is-not-really-configurable-.patch
etungsten: We still need the `prefix=@prefix@`. It doesn't work with `prefix` hardcoded to `/usr`.

I'll go ahead and pick that commit and add a small patch.

etungsten:
Please ignore me failing miserably in trying to rename the branch.

Push above addresses both of @tjkirch 's comments:
Updates dbus-broker to 26
Adds a small patch on top of a patch that adds https://github.com/systemd/systemd/commit/60bce7c6d9606185114df1bdcd5ea100407688b8 to restore prefix path configurability.

Tested things and they still work as described in the PR description.

tjkirch (approved):

bcressey (approved):

PR #1267, opened 2021-01-07
update cargo dependencies

webern:
**Issue number:**

N/A

**Description of changes:**

cargo update, once in tools and once in sources.
Needed to bump the version of webpki-roots-shim to match what reqwest is using.

**Testing done:**

cargo make, cargo make ami and cargo make repo. Ran k8s 1.17, ran a pod. Also ssh'ed in and updog check-update worked.

Also checked `apiclient` with the following commands, which successfully changed the `motd` setting.

```sh
apiclient -m PATCH -u /settings -d '{"motd": "my own value!"}'
apiclient -m POST -u /tx/commit_and_apply
apiclient -u /settings | jq .motd
# "my own value!"
```

**Terms of contribution:**

### DIFF
--- a/sources/api/migration/migrator/Cargo.toml
+++ b/sources/api/migration/migrator/Cargo.toml
@@ -15,10 +15,10 @@ log = "0.4"
 lz4 = "1.23.1"
 nix = "0.19"
 pentacle = "1.0.0"
-rand = { version = "0.7", default-features = false, features = ["std"] }
+rand = { version = "0.8", default-features = false, features = ["std", "std_rng"] }
 regex = "1.1"
 semver = "0.11"
-simplelog = "0.8"
+simplelog = "0.9"
 snafu = "0.6"
 tempfile = "3.1.0"
 tough = "0.8"
--- a/sources/api/migration/migrator/src/main.rs
+++ b/sources/api/migration/migrator/src/main.rs
@@ -185,7 +185,11 @@ pub(crate) fn run(args: &Args) -> Result<()> {
 /// Generates a random ID, affectionately known as a 'rando', that can be used to avoid timing
 /// issues and identify unique migration attempts.
 fn rando() -> String {
-    thread_rng().sample_iter(&Alphanumeric).take(16).collect()
+    thread_rng()
+        .sample_iter(&Alphanumeric)
+        .take(16)
+        .map(char::from)
+        .collect()
 }
 
 /// Generates a path for a new data store, given the path of the existing data store,
--- a/sources/api/storewolf/Cargo.toml
+++ b/sources/api/storewolf/Cargo.toml
@@ -14,9 +14,9 @@ bottlerocket-release = { path = "../../bottlerocket-release" }
 datastore = { path = "../datastore" }
 log = "0.4"
 models = { path = "../../models" }
-rand = { version = "0.7", default-features = false, features = ["std"] }
+rand = { version = "0.8", default-features = false, features = ["std", "std_rng"] }
 semver = "0.11"
-simplelog = "0.8"
+simplelog = "0.9"
 snafu = "0.6"
 toml = "0.5"
 
--- a/tools/pubsys-setup/Cargo.toml
+++ b/tools/pubsys-setup/Cargo.toml
@@ -13,7 +13,7 @@ pubsys-config = { path = "../pubsys-config/" }
 reqwest = { version = "0.10.1", default-features = false, features = ["rustls-tls", "blocking"] }
 sha2 = "0.9"
 shell-words = "1.0"
-simplelog = "0.8"
+simplelog = "0.9"
 snafu = "0.6"
 structopt = { version = "0.3", default-features = false  }
 tempfile = "3.1"
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -7,7 +7,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "78d1833b3838dbe990df0f1f87baf640cf6146e898166afe401839d1b001e570"
 dependencies = [
  "bitflags",
- "bytes",
+ "bytes 0.5.6",
  "futures-core",
  "futures-sink",
  "log",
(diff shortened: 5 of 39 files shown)


tjkirch (approved):
Would you please also check (1) unit tests and (2) that apiclient can change/commit/apply settings?

Does https://github.com/kbknapp/cargo-outdated show any bigger updates we need to apply?  (modulo tokio, gptman)  Does https://github.com/est31/cargo-udeps show anything unused?  (careful with that one...)

webern:
Expanded to include all reasonable major updates. No udeps found. Unit tests pass (btw they run in CI as well).

I still need to repeat the testing that I listed in the PR description and add to that the apiclient tests suggested above.

Created two follow-up issues #1268 and #1269 for major dep upgrades that require follow-up.

webern:
Update: I repeated the testing (and added a test of apiclient) after dep major updates in c76670b and everything seems to be working.

etungsten (approved):

sources/api/storewolf/Cargo.toml
```diff
@@ -14,9 +14,9 @@ bottlerocket-release = { path = "../../bottlerocket-release" }
 datastore = { path = "../datastore" }
 log = "0.4"
 models = { path = "../../models" }
-rand = { version = "0.7", default-features = false, features = ["std"] }
+rand = { version = "0.8", default-features = false, features = ["std"] }
```
srgothi92: any particular reason why `std_rng` feature is not included here and included in `netdog`. according to [docs](https://docs.rs/rand/0.8.0/rand/index.html#functions) both `std and std_rng` feature is needed to use `thread_rng`

sources/api/migration/migrator/src/main.rs
```diff
@@ -185,7 +185,11 @@ pub(crate) fn run(args: &Args) -> Result<()> {
 /// Generates a random ID, affectionately known as a 'rando', that can be used to avoid timing
 /// issues and identify unique migration attempts.
 fn rando() -> String {
-    thread_rng().sample_iter(&Alphanumeric).take(16).collect()
+    thread_rng()
+        .sample_iter(&Alphanumeric)
+        .take(16)
+        .map(|c| char::from(c))
```
srgothi92: Just curious: was syntax ` .map(char::from)`
    not working.  Reference from their doc: https://docs.rs/rand/0.8.0/rand/distributions/trait.Distribution.html#example

sources/api/storewolf/Cargo.toml
webern: I guess the compiler did not compel me to do so since the feature had been enabled elsewhere.

sources/api/migration/migrator/src/main.rs
webern: You are correct, I know that clippy would have compelled me to do so had I used it.

webern:
> webern force-pushed the webern:bump branch from c76670b to d4ad267 now

Addresses @srgothi92 's findings.

sources/api/storewolf/Cargo.toml
srgothi92: Does that mean we don't need to explicitly enable over here ?

srgothi92 (approved):

sources/api/migration/migrator/Cargo.toml
```diff
@@ -15,13 +15,13 @@ log = "0.4"
 lz4 = "1.23.1"
 nix = "0.19"
 pentacle = "1.0.0"
-rand = { version = "0.7", default-features = false, features = ["std"] }
+rand = { version = "0.8", default-features = false, features = ["std"] }
```
tjkirch: Same thing with std_rng here?

sources/api/migration/migrator/src/main.rs
```diff
@@ -185,7 +185,11 @@ pub(crate) fn run(args: &Args) -> Result<()> {
 /// Generates a random ID, affectionately known as a 'rando', that can be used to avoid timing
 /// issues and identify unique migration attempts.
 fn rando() -> String {
-    thread_rng().sample_iter(&Alphanumeric).take(16).collect()
+    thread_rng()
+        .sample_iter(&Alphanumeric)
+        .take(16)
+        .map(char::from)
+        .collect()
```
tjkirch, line 192: Would you mind testing this bit?  Independently of migrator would be fine.  There were a few changes to Alphanumeric / char handling, and I just want to be sure this still generates a rando OK.

sources/growpart/Cargo.toml
```diff
@@ -13,7 +13,7 @@ gptman = { version = "0.6.1", default-features = false }
 snafu = "0.6"
 libc = "0.2"
 block-party = { path = "../updater/block-party" }
-inotify = "0.8.3"
+inotify = "0.9"
```
tjkirch: This brings in tokio 0.3 - should it wait?

tools/pubsys-setup/Cargo.toml
```diff
@@ -10,10 +10,10 @@ publish = false
 hex = "0.4.0"
 log = "0.4"
 pubsys-config = { path = "../pubsys-config/" }
-reqwest = { version = "0.10.1", default-features = false, features = ["rustls-tls", "blocking"] }
+reqwest = { version = "0.11", default-features = false, features = ["rustls-tls", "blocking"] }
```
tjkirch: Can we hold off on this (and in buildsys) until we do a bigger tokio-dep update?

tjkirch (commented):
I'm a little hesitant about updating tough, just to change the migrator code back for the re-removed lifetime stuff in 0.10, but I don't mind if you prefer it.

sources/api/migration/migrator/src/main.rs
webern: ~~Is a unit test OK? What's the assertion? Just that len==16 and is_alphanumeric?~~

webern:
> webern force-pushed the webern:bump branch from d4ad267 to f2d56ed 28 seconds ago

Undo:
- fmt of sources/api/datastore/src/memory.rs
- tough version changes
- inotify version change
- reqwest version change

sources/api/migration/migrator/src/main.rs
webern: Looks good to me:

```rust
use rand::distributions::Alphanumeric;
use rand::{thread_rng, Rng};

fn main() {
    for _ in 0..10 {
        let random_id: String = thread_rng()
            .sample_iter(&Alphanumeric)
            .take(16)
            .map(char::from)
            .collect();
        println!("{}", random_id)
    }
}
```

```text
TakPKvEYjfUTActX
fnqlC8k9vjeLXBH1
y1fHGADCkFboNh5l
eF3FqK9L0ORGIrH8
wXU5GXogiTT9W0Wd
kAVPs4GRsuDj3p4q
0WMbnQkdXgF3WNLs
HW4ciPwl11guLLn9
WxIknhrWN2MmcAMP
4iw9Y9sADf8WLxTG
```

tjkirch (approved):

PR #1274, opened 2021-01-11
Fix build.yml to ignore changes that include only .md files

gregdek:
That's a lot of CI to run for a PR to a markdown file. This PR should allow build.yml to say "hey, if it's all md files, just, y'know, don't do all that stuff."

I can't think of any cases in which we'd want .md file changes to kick off CI runs, but if you can, feel free to reject this PR.

Also, cheers to the GitHub built-in editor for making sure the yaml is right.

**Issue number:**

**Description of changes:**

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,8 @@ name: Build
 on:
   pull_request:
     branches: [develop]
+    paths-ignore: 
+      - '**.md'
 jobs:
   build:
     runs-on: [self-hosted, linux, x64]


gregdek:
(See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags and https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet for how this works)

tjkirch (approved):
Looks good to me; I imagine that if we want CI to run on md files, we'd probably want a rather different Action.

samuelkarp:
~It looks like CI did run for this PR.  Is that expected?~

I missed the obvious.  See @etungsten's response.

webern:
It looks like it *should* work, based on the documentation. However CI ran for this PR which means it did not work as desired.

etungsten:
The CI ran for this PR because the change is against a `.yml` file and the we're ignoring `.md` files

samuelkarp (approved):

webern:
> The CI ran for this PR because the change is against a `.yml` file and the we're ignoring `.md` files

🎓

webern (approved):

PR #1281, opened 2021-01-14
docs: add lockdown notes to SECURITY_GUIDANCE.md

### LINKED ISSUE #1224: Update security documentation for kernel lockdown mode
The [SECURITY_FEATURES](https://github.com/bottlerocket-os/bottlerocket/blob/develop/SECURITY_FEATURES.md) and/or [SECURITY_GUIDANCE](https://github.com/bottlerocket-os/bottlerocket/blob/develop/SECURITY_GUIDANCE.md) documents could probably use an update now that #1223 is merged and added lockdown capability.

bcressey:
**Issue number:**
Fixes #1224

**Description of changes:**
Add notes about the Lockdown LSM to security guidance.

**Testing done:**
Verified that the example works correctly as user-data and disables unsigned kernel modules from being loaded.

**Terms of contribution:**

### DIFF
--- a/SECURITY_GUIDANCE.md
+++ b/SECURITY_GUIDANCE.md
@@ -13,6 +13,7 @@ We provide these recommendations, along with [details](#details) and [examples](
 | [Restrict access to the host API socket](#restrict-access-to-the-host-api-socket)                   | Critical  |
 | [Restrict access to the container runtime socket](#restrict-access-to-the-container-runtime-socket) | Critical  |
 | [Design for host replacement](#design-for-host-replacement)                                         | Important |
+| [Enable kernel lockdown](#enable-kernel-lockdown)                                                   | Important |
 | [Limit use of host containers](#limit-use-of-host-containers)                                       | Important |
 | [Limit use of privileged SELinux labels](#limit-use-of-privileged-selinux-labels)                   | Important |
 | [Limit access to system mounts](#limit-access-to-system-mounts)                                     | Important |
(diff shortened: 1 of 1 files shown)


SECURITY_GUIDANCE.md
```diff
@@ -104,6 +105,28 @@ If the kernel is ever compromised through a local exploit, then other defenses m
...
+### Enable kernel lockdown
+
+The security mechanisms in Bottlerocket ultimately depend on the kernel for enforcement.
+This includes access controls such as capabilities and SELinux, and integrity checks such as dm-verity.
+Modifications to the running kernel could bypass or subvert these mechanisms.
+
+Bottlerocket enables the Lockdown security module and offers settings to choose from one of three modes.
+
+The first mode, "none", effectively disables the protection.
+This is the default in today's [variants](variants/) of Bottlerocket, for compatibility with existing deployments.
```
tjkirch, line 117: Do you expect the default to change for some variants?  I see it as more of a global Bottlerocket security offering.

SECURITY_GUIDANCE.md
```diff
@@ -104,6 +105,28 @@ If the kernel is ever compromised through a local exploit, then other defenses m
...
+
+The second mode, "integrity", blocks most ways to overwrite the kernel's memory and modify its code.
+This will become the default in future variants.
```
tjkirch, line 120: Similar question here - some variants are longer-lived, like aws-ecs-1, and I believe we should aim for changing the setting for existing variants rather than segmenting our security story by variant.

SECURITY_GUIDANCE.md
```diff
@@ -223,6 +246,11 @@ These settings can passed as [user data](https://docs.aws.amazon.com/AWSEC2/late
 They apply to any Bottlerocket variant.
```
tjkirch, line 246: Does this comment still apply if you're talking about segmenting this setting's default by variant?

bcressey: It's a correct statement today, and would be a no-op if we change the default.

SECURITY_GUIDANCE.md
bcressey: We could change the default for new launches but migrate to "none" on upgrade, which might be the current behavior.

Today we have the property that you can continue launching new versions of the same variant with no user-data changes. I worry about discarding that property; we don't want moving to new versions to become a scary proposition.

SECURITY_GUIDANCE.md
bcressey: For aws-ecs-1, I'd like to change the default prior to GA.

The aws-k8s-* variants will age out of support over the next year, so making the changes in newer variants would let us converge on the new default without risk of disruption.

SECURITY_GUIDANCE.md
samuelkarp: I'd worry about anything where we change behavior for new launches as compared to upgrades because it introduces a risk for drift within a customer's cluster.  For example, if a customer operates an auto-scaling group and changes the ASG to launch a new AMI while upgrading the existing instances, some will have "none" (the upgraded ones) while others won't.

SECURITY_GUIDANCE.md
samuelkarp: I'm :+1: on changing the default for `aws-ecs-1` prior to GA.  I'm also :+1: on introducing this to the new `aws-k8s-*` variants as we add them.

samuelkarp (approved):

jhaynes (approved):

zmrow (approved):

webern (approved):

Issue #1218, opened 2020-11-17
Platform-agnostic host metadata for early-boot-config and pluto


zmrow:
## Problem

`early-boot-config` and `pluto` use IMDS directly.  In order to start testing Bottlerocket on other platforms, these services must retrieve metadata and user data from other sources.

## Proposed Solution

For the short term, we will extend `early-boot-config` for other platforms via the `PlatformDataProvider` trait provided in the program and conditionally compile it based on the variant.  `pluto` will be similarly refactored and extended, adding additional arguments for other platforms if required.  Doing so removes the first roadblock, allows us to begin testing on other platforms, and start gaining knowledge.  While conditional compilation isn’t our favorite solution, the code can be factored out into a more unified solution should we find one in the future.  Given the fact that there are only two first party Rust programs to extend makes this low risk and a two-way door.

## Alternatives

*Metadata translation service*
This service would run on the Bottlerocket host and programs would call it rather than IMDS.  It would translate any calls to the appropriate source for metadata.  This is a compelling idea, but given how little we know about running Bottlerocket on other platforms, it feels a bit too early to commit to such a service.   It is also yet another service to run at boot.  The complexity it would add feels a bit unnecessary at this point. 

*One-shot program to query for all metadata*
This idea is similar to the above metadata translation service, except it wouldn’t be a long running program.  This program would run once at boot, query for the required metadata, and populate a file or data store with a commonly agreed upon format.  Other programs would query this file or data store.  This idea solves the issue of unifying the logic for querying various metadata services.  However, it does raise a few problems of its own.  This one-shot program must shape data in a format that other programs must agree on, which means a common and conditionally compiled library.  We also would need to agree on a secure place to store this data.  It is also another service to run on boot.  Once again, the additional complexity doesn’t seem worth it for the short term.

Related to #968, #1097, #1114

zmrow:
After thinking about this a bit more, I had a few more thoughts about the conditional compilation piece of this puzzle.  I _do_ think that conditionally compiling `early-boot-config` is the correct path, but given the current state of variants and our build system, I don't think that _now_ is the right time to do so.  

Here's why:
* our build system doesn't have the concept of "families"; meaning that we don't have a good way of grouping "aws" variants, or "k8s" variants, etc.  This means that every single time we add a variant, we'll need to remember to revisit this code and add an additional `cfg` attribute.  This in itself is some code smell that can't be ignored.
[...]

Issue #1260, opened 2021-01-05
Composable Variants (or Richer Variant Semantics)


webern:
Variants are defined as if they were independent of any other variant. However we see a pattern emerging where variants share traits with other variants (Several Kubernetes variants that differ only in the version of Kubernetes, for example).

Looking to a future where many more variants may exist, we should consider whether the current mechanisms for defining variants are sufficient, or if some mechanisms for composing variants from re-usable building blocks should be introduced.

Edit: There are a couple of places in the codebase that are referencing this issue. In particular, in places that we want to interact with the `models::Settings` struct, we need to do conditional compilation. In a variant build using buildsys, we can prevent compilation of crates that aren't compatible for the current variant. But we want to preserve the usability of the `sources` workspace for developer workflow and unit-tests (e.g. `cargo make unit-tests`). Thus we end up with conditional compilation (using `#[cfg(something)]`) to avoid compiling incompatible bits of the sources workspace when the 'wrong' variant is being compiled.

webern:
Things have changed a lot in our thinking since I opened this issue. We are now pursuing #2669 which includes changes throughout the build system. Although #2669 doesn't address this directly, it does make this issue less relevant. Closing in favor of issues that we will right that solve problems more directly as they come up.

PR #1284, opened 2021-01-19
Add (short term) conditional compilation to early-boot-config

zmrow:
**Issue number:**
Related to #1218

**Description of changes:**
```
This change adds a simple (short term) solution for conditionally
compiling early-boot-config based on the current variant.  As part of
the change, local file handling (which was previously broken) is split
out into its own trait implementation and only compiled into the program
for "aws-dev" variants.
```

**Testing done:**
* Built and ran the `aws-dev` variant both with and without local files.  The system boots and correctly uses local files if they exist.
* Build the `aws-k8s-1.17` variant and made sure it correctly fetches user data from IMDS on a new instance.

**Terms of contribution:**

### DIFF
--- a/sources/api/early-boot-config/build.rs
+++ b/sources/api/early-boot-config/build.rs
@@ -4,8 +4,30 @@ use std::env;
 use std::fs::File;
 use std::io::Write;
 use std::path::PathBuf;
+use std::process;
 
 fn main() {
+    // The code below emits `cfg` operators to conditionally compile this program based on the
+    // current variant.  This approach is only meant to be in use for the short term; when the
+    // build system supports ideas like "variant families" we should be able to drive the
+    // conditional compilation in a less brittle way.
+    println!("cargo:rerun-if-env-changed=VARIANT");
+    if let Ok(variant) = env::var("VARIANT") {
+        // The aws-dev variant includes the ability to read user data from local file
+        if variant == "aws-dev" {
+            println!("cargo:rustc-cfg=bottlerocket_platform=\"aws-dev\"");
+        } else if variant.starts_with("aws") {
+            println!("cargo:rustc-cfg=bottlerocket_platform=\"aws\"");
+        } else {
+            eprintln!(
+            "For local builds, you must set the 'VARIANT' environment variable so we know which data \
+            provider to build. Valid values are the directories in models/src/variants/, for \
+            example 'aws-k8s-1.17'."
+            );
+            process::exit(1);
+        }
+    }
+
     // Check for environment variable "SKIP_README". If it is set,
     // skip README generation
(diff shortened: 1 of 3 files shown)


sources/api/early-boot-config/src/main.rs
```diff
@@ -306,11 +307,59 @@ impl PlatformDataProvider for AwsDataProvider {
     }
 }
 
-/// This function determines which provider we're currently running on.
+#[cfg(bottlerocket_platform = "aws-dev")]
+struct LocalFileDataProvider;
```
tjkirch, line 311: I haven't looked through the rest yet, but why restrict the file provider to aws-dev?  It's there for testing in odd situations, I'd say, and odd situations can arise in any variant.  Is there a reason not to allow it in other variants?

sources/api/early-boot-config/build.rs
```diff
@@ -4,8 +4,30 @@ use std::env;
 use std::fs::File;
 use std::io::Write;
 use std::path::PathBuf;
+use std::process;
 
 fn main() {
+    // The code below emits `cfg` operators to conditionally compile this program based on the
+    // current variant.  This approach is only meant to be in use for the short term; when the
```
webern, line 11: This caveat might be better provided with `// TODO - replace this short term approach blah blah` so that it sticks out as something that should be deleted when no longer true.

bcressey:
Tell me more about the "short term" part. 🤔

webern (approved):
One could emit `aws` as `bottlerocket_platform` and emit an additional `include_file_provider=true` (or something) to conditionally add the file provider. It seems like the concerns are separate whereby `aws_dev` is still an `aws` platform, and the file provider is a different concern.

That being said, what you have here gets the job done as you need, is easy to understand, and does no harm IMO.

webern:
> Tell me more about the "short term" part. 🤔

I think there's some hope that rethinking of variants and build logic will make this kind of thing unnecessary (or at least different) in the future.

bcressey:
> I think there's some hope that rethinking of variants and build logic will make this kind of thing unnecessary (or at least different) in the future.

OK, that makes sense. I'd like to remove that phrase from the commit and pull request title, and just link over to an issue to track that if we have one. If we merge this then it could be around for a while.

webern:
> > I think there's some hope that rethinking of variants and build logic will make this kind of thing unnecessary (or at least different) in the future.
> 
> OK, that makes sense. I'd like to remove that phrase from the commit and pull request title, and just link over to an issue to track that if we have one. If we merge this then it could be around for a while.

That seems like a good idea, the issue in question is probably #1260 (though the scope is quite broad currently).

zmrow:
Closing in favor of #1288

PR #1288, opened 2021-01-21
Add conditional compilation to early-boot-config

zmrow:
**Issue number:**
Related to #1218

**Description of changes:**
This change adds a simple solution for conditionally compiling
early-boot-config based on the current variant.  As part of the change,
local file handling (which was previously broken) is split into its own
trait implementation and only compiled into the program for "aws-dev"
variants.  We also add the ability to read a mounted cdrom for user
data.

Most of the code in `main.rs` has been moved to modules, but the logic is still very largely the same.

A separate PR will contain the relevant mount units to correctly mount a CD-ROM.

**Testing done:**
* Built and ran the aws-dev variant both with and without local files. The system boots and correctly uses local files if they exist.
* Build the aws-k8s-1.17 variant and made sure it correctly fetches user data from IMDS on a new instance.
* Built and ran a fake variant (based on `aws-dev`) locally to test the cd-rom user data capability.  Both `xml` and `user-data` files worked as expected.

**Terms of contribution:**

### DIFF
--- a/sources/api/early-boot-config/build.rs
+++ b/sources/api/early-boot-config/build.rs
@@ -4,8 +4,31 @@ use std::env;
 use std::fs::File;
 use std::io::Write;
 use std::path::PathBuf;
+use std::process;
 
 fn main() {
+    // The code below emits `cfg` operators to conditionally compile this program based on the
+    // current variant.
+    // TODO: Replace this approach when the build system supports ideas like "variant
+    // tags": https://github.com/bottlerocket-os/bottlerocket/issues/1260
+    println!("cargo:rerun-if-env-changed=VARIANT");
+    if let Ok(variant) = env::var("VARIANT") {
+        if variant == "aws-dev" {
+            println!("cargo:rustc-cfg=bottlerocket_platform=\"aws-dev\"");
+        } else if variant.starts_with("aws") {
+            println!("cargo:rustc-cfg=bottlerocket_platform=\"aws\"");
+        } else if variant.starts_with("vmware") {
+            println!("cargo:rustc-cfg=bottlerocket_platform=\"vmware\"");
+        } else {
+            eprintln!(
+            "For local builds, you must set the 'VARIANT' environment variable so we know which data \
+            provider to build. Valid values are the directories in models/src/variants/, for \
+            example 'aws-k8s-1.17'."
+            );
+            process::exit(1);
+        }
+    }
+
     // Check for environment variable "SKIP_README". If it is set,
     // skip README generation
     if env::var_os("SKIP_README").is_some() {
--- a/sources/api/early-boot-config/src/main.rs
+++ b/sources/api/early-boot-config/src/main.rs
@@ -15,15 +15,15 @@ Currently, Amazon EC2 is supported through the IMDSv1 HTTP API.  Data will be ta
 #[macro_use]
 extern crate log;
 
-use http::StatusCode;
-use reqwest::blocking::Client;
-use serde::Serialize;
-use serde_json::json;
 use simplelog::{Config as LogConfig, LevelFilter, TermLogger, TerminalMode};
-use snafu::{ensure, OptionExt, ResultExt};
-use std::path::Path;
+use snafu::{ensure, ResultExt};
+use std::fs;
 use std::str::FromStr;
-use std::{env, fs, process};
+use std::{env, process};
+
+mod provider;
+mod settings;
+use crate::provider::PlatformDataProvider;
 
 // TODO
 // Tests!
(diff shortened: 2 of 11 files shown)


zmrow:
:arrow_double_up: Fix the conflicts with the base branch

sources/api/early-boot-config/build.rs
```diff
@@ -4,8 +4,31 @@ use std::env;
 use std::fs::File;
 use std::io::Write;
 use std::path::PathBuf;
+use std::process;
 
 fn main() {
+    // The code below emits `cfg` operators to conditionally compile this program based on the
+    // current variant.
+    // TODO: Replace this approach when the build system supports ideas like "variant
+    // families": https://github.com/bottlerocket-os/bottlerocket/issues/1260
```
webern:
```suggestion
    // TODO: Replace this approach when the build system supports ideas like "variant
    // tags": https://github.com/bottlerocket-os/bottlerocket/issues/1260
```

I think we are preferring a more general, or flatter, conception currently.

sources/api/early-boot-config/src/main.rs
```diff
@@ -39,302 +48,26 @@ const TRANSACTION: &str = "bottlerocket-launch";
...
-            serde_json::from_str(&iid_str).context(error::DeserializeJson)?;
-        let region = iid
-            .get("region")
-            .context(error::IdentityDocMissingData { missing: "region" })?;
-        let val = json!({ "aws": {"region": region} });
-
-        SettingsJson::from_val(&val, desc).map(|s| Some(s))
+/// This function returns the appropriate data provider for this variant. It exists primarily to
+/// keep the ugly bits of conditional compilation out of the main function.
+fn find_provider() -> Result<Box<dyn PlatformDataProvider>> {
```
webern: Since the 'finding' is compile-time logic, I think I would name the function something like `make_provider`, `construct_provider`, or `create_provider`.

sources/api/early-boot-config/src/main.rs
```diff
@@ -454,3 +195,35 @@ async fn main() {
...
     }
 }
+
+mod error {
+    use http::StatusCode;
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility = "pub(super)")]
+    pub(super) enum Error {
```
webern, line 197: I'm not fully convinced by this error approach. The use of an Enum doesn't make sense to me here given that it contains a different variant for each provider type, where only one of these variant choices is present in the compiled code.

I think the approach used by `std::io::Error` which I followed in `tough::transport::TransportError` might be better. In that approach the underlying error type can be anything, and an enum is used to convey information that is generally useful regardless of the underlying provider/error type. The challenge with that approach is that you can't impl `From` or `Into` like you have here.

it would look something like:

```rust
pub enum ProviderErrorKind {
    IoError,
    ApiError,
    MalformedSettings,
    Other,
}

#[derive(Debug)]
pub struct ProviderError {
    kind: ProviderErrorKind,
    source: Option<Box<dyn Error + Send + Sync>>,
}

impl ProviderError {
    pub fn new_without_error<S>(kind: ProviderErrorKind) -> Self
    where
        S: AsRef<str>,
    {
        Self {
            kind,
            url: url.as_ref().into(),
            source: None,
        }
    }

    pub fn new_with_error<S, E>(kind: source: E) -> Self
    where
        E: Into<Box<dyn Error + Send + Sync>>,
        S: AsRef<str>,
    {
        Self {
            kind,
            url: url.as_ref().into(),
            source: Some(source.into()),
        }
    }

    pub fn kind(&self) -> ProviderErrorKind {
        self.kind
    }

}

```

The 'problem' here is that Snafu becomes less useful (incompatible?)

This pattern presents a better abstraction for the code that is using the Provider trait at the expense of having to type:

`.map_err(|e| ProviderError::new_with_error(ProviderErrorKind::MalformedSettings, e))?`

The advantages:
- No conditional compilation in this error type.
- No need to add a new type when a new provider is added.

sources/api/early-boot-config/src/provider.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// natural; you can also send all changes in one entry if you like.
+    fn platform_data(&self) -> Result<Vec<SettingsJson>>;
+}
+
+mod error {
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility = "pub(super)")]
+    pub(crate) enum Error {
```
webern: This error enum makes my convictions above even stronger! I think we should try to reduce the conditional compilation for the sake of error types.

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,212 @@
...
+            error::UserDataFileCount {
+                location: Self::CD_ROM_MOUNT
+            }
+        );
+
+        // Read the file; XML files require extra processing.
+        info!("'{}' exists, using it", user_data_file.display());
+        let user_data_str = match user_data_file.extension().and_then(OsStr::to_str) {
+            Some("xml") | Some("XML") => Self::ovf_user_data(user_data_file)?,
+            Some(extension) => return error::UserDataFileExtension { extension }.fail(),
```
webern: Slightly weird since this can never happen. I don't have a better idea though. (It could be `_ =>` but I don't see why that's any better).

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,212 @@
...
+        let table = val.as_table_mut().context(error::UserDataNotTomlTable)?;
+        let inner = table
+            .remove("settings")
+            .context(error::UserDataMissingSettings)?;
+
+        let json = SettingsJson::from_val(&inner, "user data").context(error::SettingsToJSON)?;
+        Ok(Some(json))
+    }
+
+    /// Read and base64 decode user data contained in an OVF file
```
webern, line 77: Did you mean for the first line to be a doc comment and the rest to be `//`?

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,212 @@
...
+        let file = File::open(path).context(error::InputFileRead { path })?;
+        let reader = BufReader::new(file);
+
+        // Deserialize the OVF file, dropping everything we don't care about
+        let ovf: Environment =
+            serde_xml_rs::from_reader(reader).context(error::XMLDeserialize { path })?;
+
+        // We have seen the keys in the `Property` section be "namespaced" like
+        // "oe:key" or "of:key". We haven't been able to determine why certain files are
+        // this way, or where it comes from.  However, the namespacing doesn't seem to
```
webern: This is how XML Schema (xsd) works. It associates the elements with a declaration at the outermost element where an XSD schema is referenced and bound to the namespace. Since you're not trying to validate that a schema is being adhered to (beyond the presence of elements and data you care about), it is appropriate to ignore the namespace prefixes.

Somewhere, probably at the root element in those files that have prefixes, there is probably something like this.

```xml
<whatever xmlns:oe="http://something.com/OVF.xsd">
    <oe:whateverelement/>
</whatever>
```


- https://www.w3.org/TR/xmlschema11-1/
- https://www.w3.org/TR/xml-names/

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,212 @@
...
+        // mean anything or change the way that user data is contained in the files.
+        // `serde_xml_rs` effectively ignores these namespaces and returns "key" / "value":
+        // https://github.com/Rreverser/serde-xml-rs/issues/64#issuecomment=540448434
+        let mut base64_str = String::new();
+        let user_data_key = "user-data";
+        for property in ovf.property_section.properties {
```
webern, line 100: There's an `iter().find(|xxx| xxx==yyy)` syntax if you prefer it.

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,212 @@
...
+
+#[derive(Debug, Deserialize)]
+struct Property {
+    pub key: String,
+    pub value: String,
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=
+
+mod error {
```
webern, line 157: In the error pattern I've suggested above, you could

```rust
impl Into<ProviderError> for cdrom::error::Error {
    /// ...
}
```

Which would avoid `map_err` calls.

sources/api/early-boot-config/src/provider/local_file.rs
```diff
@@ -0,0 +1,68 @@
...
+        let user_data_str =
+            fs::read_to_string(Self::USER_DATA_FILE).context(error::InputFileRead {
+                path: Self::USER_DATA_FILE,
+            })?;
+
+        if user_data_str.is_empty() {
+            return Ok(output);
+        }
+
+        // Remove outer "settings" layer before sending to API
```
webern, line 28: Repeated code, maybe keep an eye out for a third appearance and pull it if that happens.

sources/api/early-boot-config/src/provider/cdrom.rs
zmrow: I did indeed!

sources/api/early-boot-config/src/provider/cdrom.rs
zmrow: Ah - that's a great piece of information!  Thanks for enlightening me.

I'll update the comment.

sources/api/early-boot-config/src/main.rs
webern: We got together and implemented this to see what it would look like. It looks like this:

https://github.com/zmrow/bottlerocket/compare/ebc_refactor...webern:ebc_error

There's some cleanup, but the basic premise is that one doesn't not have to introduce any conditional compilation into the provider's error type. This would keep the surface area of conditional compilation restricted to one place, (the instantiation of the provider).

sources/api/early-boot-config/src/provider/cdrom.rs
zmrow: Yeah, it's slightly weird.  I'll change it to be `unreachable!()`

zmrow:
The above force push should address outstanding comments.

The tl;dr is: "removed any odd conditional compilation in error handling."

After multiple offline conversations I think there's a much better solution in place for error handling here.  The `PlatformDataProvider` trait was changed to return `Box<std::error::Error` which eliminates the error type that was previously defined in that module.  Doing so eliminates _all_ of the conditional compilation that was previously converting different provider's errors to a single error type so `main.rs` could handle it.  Now `main` has a single error variant with a source of `Box<std::error::Error>`.  Boom! Roasted!

@tjkirch brought up the valid point that we should probably separate "platform" from "data provider", and let the platform (AWS, VMWare, etc) choose the provider (local file, IMDS, etc.).  I'll open an issue separately; it shouldn't block this PR.

sources/api/early-boot-config/src/main.rs
zmrow: Addressed in my comments below!

sources/api/early-boot-config/src/provider.rs
```diff
@@ -0,0 +1,23 @@
+//! The provider module owns the `PlatformDataProvider` trait as well as the methods needed to
+//! convert submodule errors to a single error type for further consumption in `main()`.
```
etungsten: >`//! convert submodule errors to a single error type for further consumption in `main()`.`

This is no longer true I think.

etungsten (approved):

sources/api/early-boot-config/src/provider.rs
zmrow: Good catch - fixed.

zmrow:
^ Address @etungsten 's comment

webern:
> trait was changed to return `Box<std::error::Error` which eliminates the error type that was previously defined in that module.

Yes, I agree that is fine. The wrapper that I wrote basically amounts to implementing `Error` for `Box<dyn Error>`. That would probably be nicer if a this were a library to be used externally, but here we don't care that our error type isn't an `Error`.

webern (approved):

sources/api/early-boot-config/src/provider/local_file.rs
```diff
@@ -0,0 +1,66 @@
...
+//! local file
+
+use super::{PlatformDataProvider, SettingsJson};
+use snafu::{OptionExt, ResultExt};
+use std::fs;
+
+pub(crate) struct LocalFileDataProvider;
+
+impl LocalFileDataProvider {
+    pub(crate) const USER_DATA_FILE: &'static str = "/etc/early-boot-config/user-data";
```
bcressey, line 11: Does this need to be `pub(crate)`?

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,209 @@
...
+            None => return Ok(None),
+        };
+
+        // There should only be 1 file, if more exist something is wonky
+        ensure!(
+            user_data_files.next().is_none(),
+            error::UserDataFileCount {
+                location: Self::CD_ROM_MOUNT
+            }
+        );
```
bcressey: nit: for me, all three of these comments are about "what" which is pretty obvious. A "why" comment could be helpful for building a list of the files, since it's not obvious why there wouldn't be a canonical file name.

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,209 @@
...
+            Some(_) => unreachable!(),
+            None => fs::read_to_string(&user_data_file).context(error::InputFileRead {
+                path: user_data_file,
+            })?,
+        };
+
+        // If the file was empty, return
+        if user_data_str.is_empty() {
+            return Ok(None);
+        }
```
bcressey: same nit re: obvious comment

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,209 @@
...
+
+        // There should only be 1 file, if more exist something is wonky
+        ensure!(
+            user_data_files.next().is_none(),
+            error::UserDataFileCount {
+                location: Self::CD_ROM_MOUNT
+            }
+        );
+
+        // Read the file; XML files require extra processing.
```
bcressey:
```suggestion
        // XML files require extra processing, while a user-supplied file should already be in the TOML format.
```

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,209 @@
...
+        UserDataFileCount { location: String },
+
+        #[snafu(display("TOML data did not contain 'settings' section"))]
+        UserDataMissingSettings,
+
+        #[snafu(display("Data is not a TOML table"))]
+        UserDataNotTomlTable,
+
+        #[snafu(display("Unable to deserialize XML from: '{}': {}", path.display(), source))]
+        XMLDeserialize {
```
bcressey: Why not `XmlDeserialize`?

sources/api/early-boot-config/src/provider/cdrom.rs
```diff
@@ -0,0 +1,209 @@
...
+        let reader = BufReader::new(file);
+
+        // Deserialize the OVF file, dropping everything we don't care about
+        let ovf: Environment =
+            serde_xml_rs::from_reader(reader).context(error::XMLDeserialize { path })?;
+
+        // We have seen the keys in the `Property` section be "namespaced" like "oe:key" or
+        // "of:key".  Since we aren't trying to validate the schema beyond the presence of the
+        // elements we care about, we can ignore the namespacing.  `serde_xml_rs` effectively
+        // ignores these namespaces and returns "key" / "value":
```
bcressey: A document with an example of this kind of "namespaced", ideally in a unit test, would help with understanding.

sources/api/early-boot-config/src/provider/local_file.rs
zmrow: I believe so - we use the existence of this file in `main.rs` when deciding which data provider to return from `create_provider()`

zmrow:
^ Addressed @bcressey 's comments

sources/api/early-boot-config/src/provider/cdrom.rs
zmrow: Added a few unit tests with sample data.

bcressey (approved):

Issue #1027, opened 2020-08-11
newer versions of RPMs will be used instead of the most recent builds


bcressey:
**Platform I'm building on:**
EC2

**What I expected to happen:**
After reverting a glibc package update, I expected that other package builds and the final image would use the older versions of glibc RPMs once they were rebuilt.

**What actually happened:**
Package builds and the final image used the newer versions of the glibc packages, since they were still present in `build/rpms` and available in the local repository that we generate.

**How to reproduce the problem:**
Bump the version in a spec file; build RPMs; revert the change; rebuild RPMs; check packages installed in `rpm2img` output.

bcressey:
A workaround could be to inject a `Epoch: $(date +%s)` into the spec file during build.

One problem is that we don't do much in the way of spec file parsing in `buildsys`, so we don't have a clear map of inputs to outputs, such that we could clean them up whenever we start the new build. We also support sub-packages with a name that doesn't start with the same prefix as the main package, so a wildcard-based cleanup won't work.

The set of sub-packages can also change across an update, so parsing the current version of the spec doesn't necessarily tell us what the older outputs were. This is probably not a big deal. Querying the source RPM for all RPMs in the output directory would let us be more precise, but is relatively expensive.

PR #1289, opened 2021-01-22
Add the ability to output vmdk via qemu-img

jpculp:
**Issue number:**
N/A

**Description of changes:**
This change allows users to edit the metadata in a variant's cargo.toml to output an OVA compatible vmdk through buildsys via qemu-img.

Double quotes were also added to several places in rpm2img to prevent unwanted globbing and word splitting (SC2086).

**Testing done:**

- Built aws-k8s-1.18 variant with no toml changes and received the standard img.lz4 output.

- Built aws-k8s-1.18 variant with ```image-format = "vmdk"``` added to the toml and received stream-optimized vmdk output for bottlerocket-aws-k8s-1.18-aarch64 and bottlerocket-aws-k8s-1.18-aarch64-data.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -383,18 +383,11 @@ dependencies = ["build-variant"]
 script = [
 '''
 mkdir -p ${BUILDSYS_OUTPUT_DIR}/latest
-ln -snf ../${BUILDSYS_NAME_FULL}.img.lz4 \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}.img.lz4
-ln -snf ../${BUILDSYS_NAME_FULL}-data.img.lz4 \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-data.img.lz4
-ln -snf ../${BUILDSYS_NAME_FULL}-boot.ext4.lz4 \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-boot.ext4.lz4
-ln -snf ../${BUILDSYS_NAME_FULL}-root.ext4.lz4 \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-root.ext4.lz4
-ln -snf ../${BUILDSYS_NAME_FULL}-root.verity.lz4 \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-root.verity.lz4
-ln -snf ../${BUILDSYS_NAME_FULL}-migrations.tar \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-migrations.tar
+for artifact in ${BUILDSYS_OUTPUT_DIR}/${BUILDSYS_NAME_FULL}*; do
+  file_name="${artifact##*/}"
+  link_name="${file_name/${BUILDSYS_NAME_FULL}/${BUILDSYS_NAME_VARIANT}}"
+  ln -snf "../${file_name}" "${BUILDSYS_OUTPUT_DIR}/latest/${link_name}"
+done
 '''
 ]
 
--- a/tools/buildsys/src/builder.rs
+++ b/tools/buildsys/src/builder.rs
@@ -19,6 +19,8 @@ use std::path::{Path, PathBuf};
 use std::process::Output;
 use walkdir::{DirEntry, WalkDir};
 
+use crate::manifest::ImageFormat;
+
 /*
 There's a bug in BuildKit that can lead to a build failure during parallel
 `docker build` executions:
@@ -46,7 +48,7 @@ impl PackageBuilder {
         let arch = getenv("BUILDSYS_ARCH")?;
         let output_dir: PathBuf = getenv("BUILDSYS_PACKAGES_DIR")?.into();
 
-        // We do *not* want to rebuild most packages when the variant changes, becauses most aren't
+        // We do *not* want to rebuild most packages when the variant changes, because most aren't
         // affected; packages that care about variant should "echo cargo:rerun-if-env-changed=VAR"
         // themselves in the package's spec file.
         let var = "BUILDSYS_VARIANT";
@@ -82,7 +84,7 @@ pub(crate) struct VariantBuilder;
 
 impl VariantBuilder {
     /// Build a variant with the specified packages installed.
-    pub(crate) fn build(packages: &[String]) -> Result<Self> {
+    pub(crate) fn build(packages: &[String], image_format: Option<&ImageFormat>) -> Result<Self> {
         // We want PACKAGES to be a value that contains spaces, since that's
         // easier to work with in the shell than other forms of structured data.
         let packages = packages.join("|");
@@ -91,6 +93,10 @@ impl VariantBuilder {
         let version_image = getenv("BUILDSYS_VERSION_IMAGE")?;
         let version_build = getenv("BUILDSYS_VERSION_BUILD")?;
         let output_dir: PathBuf = getenv("BUILDSYS_OUTPUT_DIR")?.into();
+        let image_format = match image_format {
+            Some(ImageFormat::Raw) | None => String::from("raw"),
+            Some(ImageFormat::Vmdk) => String::from("vmdk"),
+        };
 
         // Always rebuild variants since they are located in a different workspace,
         // and don't directly track changes in the underlying packages.
@@ -101,12 +107,14 @@ impl VariantBuilder {
              --build-arg ARCH={arch} \
              --build-arg VARIANT={variant} \
              --build-arg VERSION_ID={version_image} \
-             --build-arg BUILD_ID={version_build}",
+             --build-arg BUILD_ID={version_build} \
+             --build-arg IMAGE_FORMAT={image_format}",
             packages = packages,
             arch = arch,
             variant = variant,
             version_image = version_image,
             version_build = version_build,
+            image_format = image_format,
         );
         let tag = format!(
             "buildsys-var-{variant}-{arch}",
--- a/tools/buildsys/src/manifest.rs
+++ b/tools/buildsys/src/manifest.rs
@@ -110,6 +110,11 @@ impl ManifestInfo {
             .and_then(|b| b.included_packages.as_ref())
     }
 
+    /// Convenience method to return the image format override, if any.
+    pub(crate) fn image_format(&self) -> Option<&ImageFormat> {
+        self.build_variant().and_then(|b| b.image_format.as_ref())
+    }
+
     /// Helper methods to navigate the series of optional struct fields.
     fn build_package(&self) -> Option<&BuildPackage> {
         self.package
@@ -152,6 +157,14 @@ pub(crate) struct BuildPackage {
 #[serde(rename_all = "kebab-case")]
 pub(crate) struct BuildVariant {
     pub(crate) included_packages: Option<Vec<String>>,
+    pub(crate) image_format: Option<ImageFormat>,
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "lowercase")]
+pub(crate) enum ImageFormat {
+    Raw,
+    Vmdk,
 }
 
 #[derive(Deserialize, Debug)]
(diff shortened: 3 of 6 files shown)


Makefile.toml
```diff
@@ -381,18 +381,11 @@ dependencies = ["build-variant"]
...
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-data.img.lz4
-ln -snf ../${BUILDSYS_NAME_FULL}-boot.ext4.lz4 \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-boot.ext4.lz4
-ln -snf ../${BUILDSYS_NAME_FULL}-root.ext4.lz4 \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-root.ext4.lz4
-ln -snf ../${BUILDSYS_NAME_FULL}-root.verity.lz4 \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-root.verity.lz4
-ln -snf ../${BUILDSYS_NAME_FULL}-migrations.tar \
-  ${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_NAME_VARIANT}-migrations.tar
+for artifact in ${BUILDSYS_OUTPUT_DIR}/${BUILDSYS_NAME_FULL}*; do
```
jpculp, line 386: One pitfall of using a loop here is if a user builds a variant as an image, and then later rebuilds that same variant as a vmdk, if the build-id does not change it will link the old artifacts img.lz4 as well.

tools/buildsys/src/builder.rs
```diff
@@ -89,6 +95,19 @@ impl VariantBuilder {
         let version_image = getenv("BUILDSYS_VERSION_IMAGE")?;
         let version_build = getenv("BUILDSYS_VERSION_BUILD")?;
         let output = getenv("BUILDSYS_OUTPUT_DIR")?;
+        let image_format = match image_format {
+            Some(ImageFormat::IMG) | None => String::from("img"),
+            Some(ImageFormat::VMDK) => String::from("vmdk"),
+        };
+        let image_options = match image_options {
+            Some(options_string) => options_string.join(","),
```
jahkeup: nit: `options_string` isn't a string here, let's avoid suggesting it is:
```suggestion
            Some(opts) => opts.join(","),
```

tools/buildsys/src/manifest.rs
```diff
@@ -152,6 +162,15 @@ pub(crate) struct BuildPackage {
...
 pub(crate) struct BuildVariant {
     pub(crate) included_packages: Option<Vec<String>>,
+    pub(crate) image_format: Option<ImageFormat>,
+    pub(crate) image_options: Option<Vec<String>>,
+}
+
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "lowercase")]
+pub(crate) enum ImageFormat {
+    IMG,
```
jahkeup: Can we refer to `IMG` instead as `Raw`? The sort of files that `IMG` is naming are typically called `raw` by tools and the various virtualization communities.

jahkeup (changes requested):

Makefile.toml
jahkeup: That might be desirable in some situations, for example: to build both a raw disk image *and* a VMDK.

tools/rpm2img
```diff
@@ -90,43 +95,43 @@ rm -rf "${ROOT_MOUNT}"/var/lib "${ROOT_MOUNT}"/usr/share/licenses/*
...
 
 # Set the Bottlerocket variant, version, and build-id
-echo "PRETTY_NAME=\"Bottlerocket OS ${VERSION_ID}\"" >> ${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release
-echo "VARIANT_ID=${VARIANT}" >> ${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release
-echo "VERSION_ID=${VERSION_ID}" >> ${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release
-echo "BUILD_ID=${BUILD_ID}" >> ${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release
+echo "PRETTY_NAME=\"Bottlerocket OS ${VERSION_ID}\"" >> "${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release"
+echo "VARIANT_ID=${VARIANT}" >> "${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release"
+echo "VERSION_ID=${VERSION_ID}" >> "${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release"
+echo "BUILD_ID=${BUILD_ID}" >> "${ROOT_MOUNT}/${SYS_ROOT}/usr/lib/os-release"
```
bcressey, line 142: I'd like to have all these unrelated shellcheck fixes in a separate commit.

tools/buildsys/src/builder.rs
bcressey: `image_opts` feels like the wrong abstraction boundary here - it's not consistent with the other handling of low-level details that takes place in `rpm2img`.

Let's hardcode the stream-optimized options in `rpm2img` when building VMDKs. We can name the format `stream-vmdk` if we need to distinguish it from ordinary VMDK for some reason, but I don't think we do.

Makefile.toml
bcressey: I have a fix coming soon for #1027 that should address this; artifacts from a previous `buildsys` invocation will be cleaned up prior to the next, so we shouldn't hit the situation where we have both types at once.

bcressey (changes requested):

Makefile.toml
zmrow: @jpculp Let's remove this loop.  If another solution is coming we don't need to fix the "problem" here.

tools/rpm2img
```diff
@@ -3,22 +3,26 @@
 set -eu -o pipefail
 shopt -qs failglob
 
+OUTPUT_FMT="raw"
+OUTPUT_OPTS="none"
```
zmrow: Let's eliminate this option for now

tools/rpm2img
```diff
@@ -206,16 +210,34 @@ dd if="${BOTTLEROCKET_DATA}" of="${DATA_IMAGE}" conv=notrunc bs=1M seek=1
...
+if [[ ${OUTPUT_FMT} != "raw" ]]; then
+  QEMU_PARAMS="-f raw -O ${OUTPUT_FMT}"
+
+  # Stream optimization is required for creating an Open Virtual Appliance (OVA).
+  if [[ ${OUTPUT_FMT} == "vmdk" ]]; then
+    OUTPUT_OPTS="subformat=streamOptimized"
+  fi
+
+  if [[ ${OUTPUT_OPTS} != "none" ]]; then
+    QEMU_PARAMS="${QEMU_PARAMS} -o ${OUTPUT_OPTS}"
```
zmrow: I think this is a bit over complicated.  We can simplify it a bit by eliminating the multiple extra variables and having the qemu command hardcoded for now.  If you want to encapsulate all qemu args in a single variable that can be re-used, that's fine too.

IMHO it also might help to match on a positive case first to make it simpler to read:

```
if [[ ${OUTPUT_FMT} == "vmdk" ]]; then
    # Don't need to use this variable, could just hard code the commands
    QEMU_OPTS="convert -f raw -O vmdk -o subformat=streamOptimized"

    qemu-img "${QEMU_OPTS}" .....

# Can use elif's later for additional output formats
else
    ....
fi
```__

**Review by zmrow — CHANGES_REQUESTED** [review]

File: Makefile.toml (hunk 1 above)
Comment (jpculp) [reply Makefile.toml:386]: @bcressey  can correct me, but I think his solution is just for cleaning up old artifacts, such as in the situation I mentioned above. This loop would still be necessary for generating the proper symlinks under the build/images/latest directory.

File: tools/buildsys/src/builder.rs (hunk 2 above)
Comment (jpculp) [reply tools/buildsys/src/builder.rs]: Makes sense to me, but it looks like we're just going to remove image_options for now.

File: tools/rpm2img (hunk 6 above)
Comment (jpculp) [reply tools/rpm2img]: I've updated my commits with something in-between your suggestion and the original commit.

File: Makefile.toml (hunk 1 above)
Comment (zmrow) [reply Makefile.toml:386]: Ah yeah you're right!

**Review by zmrow — APPROVED** [review]
:hamburger:

File: tools/rpm2img [hunk 7]
```diff
@@ -3,22 +3,25 @@
...
-DISK_IMAGE_NAME="${FILENAME_PREFIX}.img.lz4"
-BOOT_IMAGE_NAME="${FILENAME_PREFIX}-boot.ext4.lz4"
-VERITY_IMAGE_NAME="${FILENAME_PREFIX}-root.verity.lz4"
-ROOT_IMAGE_NAME="${FILENAME_PREFIX}-root.ext4.lz4"
-DATA_IMAGE_NAME="${FILENAME_PREFIX}-data.img.lz4"
+DISK_IMAGE_BASENAME="${FILENAME_PREFIX}"
+BOOT_IMAGE_BASENAME="${FILENAME_PREFIX}-boot"
+VERITY_IMAGE_BASENAME="${FILENAME_PREFIX}-root"
+ROOT_IMAGE_BASENAME="${FILENAME_PREFIX}-root"
+DATA_IMAGE_BASENAME="${FILENAME_PREFIX}-data"
```
bcressey: It's weird to me that `VERITY_IMAGE_BASENAME` and `ROOT_IMAGE_BASENAME` have the same value. Maybe incorporate part of the extension?

```suggestion
DISK_IMAGE_BASENAME="${FILENAME_PREFIX}"
DATA_IMAGE_BASENAME="${FILENAME_PREFIX}-data"

BOOT_IMAGE_BASENAME="${FILENAME_PREFIX}-boot.ext4"
ROOT_IMAGE_BASENAME="${FILENAME_PREFIX}-root.ext4"
VERITY_IMAGE_BASENAME="${FILENAME_PREFIX}-root.verity"
```

Then the final extension (`.img.lz4`, `.vmdk`, `.lz4`) would be provided on the same line where we invoke the tool that converts it to that format, which would keep the clarity improvement you're after.

jpculp: I can see how that would be strange. I've gone ahead and amended the commit so that we only use BASENAME for the disk and data images.

tools/rpm2img
```diff
@@ -206,14 +211,28 @@ dd if="${BOTTLEROCKET_DATA}" of="${DATA_IMAGE}" conv=notrunc bs=1M seek=1
...
+  qemu-img convert -f raw -O vmdk -o subformat=streamOptimized "${DISK_IMAGE}" "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.vmdk"
+  qemu-img convert -f raw -O vmdk -o subformat=streamOptimized "${DATA_IMAGE}" "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.vmdk"
+  chown 1000:1000 "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.vmdk" \
+    "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.vmdk"
+else
+  qemu-img convert -f raw -O "${OUTPUT_FMT}" "${DISK_IMAGE}" "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.${OUTPUT_FMT}"
+  qemu-img convert -f raw -O "${OUTPUT_FMT}" "${DATA_IMAGE}" "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.${OUTPUT_FMT}"
+  chown 1000:1000 "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.${OUTPUT_FMT}" \
+    "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.${OUTPUT_FMT}"
+fi
```
bcressey: If we don't currently expect to take this path, I'd prefer to just error out:

```suggestion
else
  echo "unexpected image format '${OUTPUT_FMT}'" >&2
  exit 1
fi
```

bcressey (commented):
LGTM, just one small concern.

tools/rpm2img
jpculp: After some further discussion, I removed the else entirely and added a case statement to check for valid output formats immediately after the argument parsing. To support additional formats in the future, they will have to be added to the case statement and an additional elif will have to be appended to this block.

jahkeup (approved):

zmrow (approved):
:medal_military:

bcressey (approved):

PR #1290, opened 2021-01-22
Add a systemd mount unit for CD-ROM devices

zmrow:
**Issue number:**
Related to #1288 

**Description of changes:**
This change adds a mount unit that will mount a CD-ROM device located at
`/dev/cdrom` to `/media/cdrom` if that device exists.  systemd includes a
udev rule to symlink the first suspected CD-ROM device to `/dev/cdrom`.  The
primary use of said device to supply user data to a Bottlerocket host.

**Testing done:**
* Build and ran a Bottlerocket AMI to ensure that this unit doesn't interfere with or delay boot
* Built and ran a fake variant using the changes from #1288 (based on aws-dev) locally to test the cd-rom user data capability. Both xml and user-data files worked as expected.
* Built and ran the same fake variant without cd-rom connected and the instance booted as expected, did not delay, and `early-boot-config` properly exited gracefully after no user data was found at the cd-rom location.

**Terms of contribution:**

### DIFF
--- a/packages/release/media-cdrom.mount
+++ b/packages/release/media-cdrom.mount
@@ -1,0 +1,21 @@
+[Unit]
+Description=CD-ROM mount (/media/cdrom)
+# Only run this unit if /dev/cdrom exists and is tracked via systemd. (systemd
+# ships with a udev rule to tag and symlink the first suspected cdrom device to
+# /dev/cdrom)
+Requires=dev-cdrom.device
+After=dev-cdrom.device
+DefaultDependencies=no
+Conflicts=umount.target
+Before=local-fs.target umount.target
+
+[Mount]
+What=/dev/cdrom
+Where=/media/cdrom
+Type=iso9660
+Options=defaults,noexec
+
+[Install]
+# This dependency ensures that systemd attempts to run this unit if the device
+# exists.
+WantedBy=dev-cdrom.device
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -25,6 +25,9 @@ Source1008: opt.mount
 Source1009: var-lib-bottlerocket.mount
 Source1010: etc-cni.mount
 
+# CD-ROM mount
+Source1015: media-cdrom.mount
+
 # Mounts that require build-time edits.
 Source1020: var-lib-kernel-devel-lower.mount.in
 Source1021: usr-src-kernels.mount.in
@@ -104,7 +107,7 @@ EOF
 
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 \
-  %{S:1002} %{S:1006} %{S:1007} %{S:1008} %{S:1009} %{S:1010} \
+  %{S:1002} %{S:1006} %{S:1007} %{S:1008} %{S:1009} %{S:1010} %{S:1015} \
   %{buildroot}%{_cross_unitdir}
 
 LOWERPATH=$(systemd-escape --path %{_cross_sharedstatedir}/kernel-devel/lower)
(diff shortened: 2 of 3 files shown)


samuelkarp (approved):

packages/release/media-cdrom.mount
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=CDROM mount(/media/cdrom)
```
tianon: I'm nothing but a causal bystander who enjoys the thought of this PR, but IMO this would probably look better (if someone were watching the boot logs) with a little extra punctuation, ala `Description=CD-ROM mount (/media/cdrom)` :innocent:

packages/release/media-cdrom.mount
```diff
@@ -0,0 +1,13 @@
...
+Requires=dev-cdrom.device
+After=dev-cdrom.device
```
jahkeup, line 7: It's worth adding a brief explanation of *why* these (`Requires`, `After`, & `WantedBy`) have to be written here, where they otherwise don't usually need to be. This is a slightly odd unit to begin with :)

jahkeup (approved):
I'd like to see a comment for the odd `.mount`, but otherwise looks great!

packages/release/media-cdrom.mount
bcressey: I would expect this mount unit to have the usual set of mount dependencies:
```
DefaultDependencies=no
Conflicts=umount.target
Before=local-fs.target umount.target
```

packages/os/early-boot-config.service
```diff
@@ -3,6 +3,9 @@ Description=Bottlerocket userdata configuration system
 # Need network online to talk to IMDS.
 After=network-online.target apiserver.service
 Requires=network-online.target apiserver.service
+# This is for ordering purposes, systemd will ignore it if the mount unit
+# never runs because the device doesn't exist
+After=media-cdrom.mount
```
bcressey: Our config services should be able to rely on this having run by this point without specifying it.

It's leaking through the `local-fs.target` abstraction, which I'd guess is because systemd doesn't know it's supposed to have started it. See my suggestion for the mount unit.

packages/release/release.spec
```diff
@@ -25,6 +25,9 @@ Source1008: opt.mount
 Source1009: var-lib-bottlerocket.mount
 Source1010: etc-cni.mount
 
+# CDROM mount
```
bcressey:
```suggestion
# CD-ROM mount
```

packages/release/media-cdrom.mount
bcressey, line 7: This should have the usual set of dependencies for mount units:
```
DefaultDependencies=no
Conflicts=umount.target
Before=local-fs.target umount.target
```

packages/release/media-cdrom.mount
```diff
@@ -0,0 +1,13 @@
...
+
+[Mount]
+What=/dev/cdrom
+Where=/media/cdrom
+Type=iso9660
+Options=defaults,noexec,context=system_u:object_r:private_t:s0
```
zmrow: @bcressey : does the `context` look correct here?

zmrow:
Addressed all comments!

jahkeup (approved):

packages/release/media-cdrom.mount
bcressey: What label does `ls -lZ` show if you don't specify this option?

I'd say `external_t` is more correct and possibly that's the default.

zmrow: Without the option the label is `unlabeled_t`

zmrow: Per offline conversation, `unlabeled_t` should be fine. I removed the `context=` from the line.

zmrow:
Per offline conversation, removed `context=` bit from the `Options` line.

bcressey (approved):

PR #1291, opened 2021-01-23
track and clean output files for builds

### LINKED ISSUE #1027: newer versions of RPMs will be used instead of the most recent builds

bcressey:
**Issue number:**
Fixes #1027

**Description of changes:**
Add a layer of indirection between the build artifacts and the final output directory, so we can keep track of what we've previously built for a given package or variant.

That allows us to remove the files before the next build, so they do not interact with other builds in unexpected ways.

**Testing done:**
Verified that the `state` directory is populated with marker files.
```
❯ find build/state/variants -type f
build/state/variants/aws-k8s-1.17/bottlerocket-aws-k8s-1.17-aarch64-1.0.5-e3234419-data.img.lz4.marker
build/state/variants/aws-k8s-1.17/bottlerocket-aws-k8s-1.17-aarch64-1.0.5-e3234419-root.ext4.lz4.marker
build/state/variants/aws-k8s-1.17/bottlerocket-aws-k8s-1.17-aarch64-1.0.5-e3234419-migrations.tar.marker
build/state/variants/aws-k8s-1.17/bottlerocket-aws-k8s-1.17-aarch64-1.0.5-e3234419-boot.ext4.lz4.marker
build/state/variants/aws-k8s-1.17/bottlerocket-aws-k8s-1.17-aarch64-1.0.5-e3234419.img.lz4.marker
build/state/variants/aws-k8s-1.17/bottlerocket-aws-k8s-1.17-aarch64-1.0.5-e3234419-root.verity.lz4.marker
```

Confirmed that the corresponding files in `packages` and `images` are removed before another build. For example, old variant images are cleaned up on each build, so we no longer accumulate artifacts named for older commits. I also bumped the release package to version 1000.0, built it, then reverted it to 0.0. The older package with the higher version number was cleaned up as expected.

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/builder.rs
+++ b/tools/buildsys/src/builder.rs
@@ -11,10 +11,13 @@ use duct::cmd;
 use nonzero_ext::nonzero;
 use rand::Rng;
 use sha2::{Digest, Sha512};
-use snafu::{ensure, ResultExt};
+use snafu::{ensure, OptionExt, ResultExt};
 use std::env;
+use std::fs::{self, File};
 use std::num::NonZeroU16;
+use std::path::{Path, PathBuf};
 use std::process::Output;
+use walkdir::{DirEntry, WalkDir};
 
 /*
 There's a bug in BuildKit that can lead to a build failure during parallel
@@ -41,7 +44,7 @@ impl PackageBuilder {
     /// Build RPMs for the specified package.
     pub(crate) fn build(package: &str) -> Result<Self> {
         let arch = getenv("BUILDSYS_ARCH")?;
-        let output = getenv("BUILDSYS_PACKAGES_DIR")?;
+        let output_dir: PathBuf = getenv("BUILDSYS_PACKAGES_DIR")?.into();
 
         // We do *not* want to rebuild most packages when the variant changes, becauses most aren't
         // affected; packages that care about variant should "echo cargo:rerun-if-env-changed=VAR"
@@ -53,7 +56,6 @@ impl PackageBuilder {
         let var = "PUBLISH_REPO";
         let repo = env::var(var).context(error::Environment { var })?;
 
-        let target = "package";
         let build_args = format!(
             "--build-arg PACKAGE={package} \
              --build-arg ARCH={arch} \
@@ -70,7 +72,7 @@ impl PackageBuilder {
             arch = arch,
         );
 
-        build(&target, &build_args, &tag, &output)?;
+        build(BuildType::Package, &package, &build_args, &tag, &output_dir)?;
 
         Ok(Self)
     }
@@ -88,13 +90,12 @@ impl VariantBuilder {
         let variant = getenv("BUILDSYS_VARIANT")?;
         let version_image = getenv("BUILDSYS_VERSION_IMAGE")?;
         let version_build = getenv("BUILDSYS_VERSION_BUILD")?;
-        let output = getenv("BUILDSYS_OUTPUT_DIR")?;
+        let output_dir: PathBuf = getenv("BUILDSYS_OUTPUT_DIR")?.into();
 
         // Always rebuild variants since they are located in a different workspace,
         // and don't directly track changes in the underlying packages.
         getenv("BUILDSYS_TIMESTAMP")?;
 
-        let target = "variant";
         let build_args = format!(
             "--build-arg PACKAGES={packages} \
              --build-arg ARCH={arch} \
@@ -113,17 +114,28 @@ impl VariantBuilder {
             arch = arch
         );
 
-        build(&target, &build_args, &tag, &output)?;
+        build(BuildType::Variant, &variant, &build_args, &tag, &output_dir)?;
 
         Ok(Self)
     }
 }
 
+enum BuildType {
+    Package,
+    Variant,
+}
+
 /// Invoke a series of `docker` commands to drive a package or variant build.
-fn build(target: &str, build_args: &str, tag: &str, output: &str) -> Result<()> {
+fn build(
+    kind: BuildType,
+    what: &str,
+    build_args: &str,
+    tag: &str,
+    output_dir: &PathBuf,
+) -> Result<()> {
     // Our Dockerfile is in the top-level directory.
     let root = getenv("BUILDSYS_ROOT_DIR")?;
-    std::env::set_current_dir(&root).context(error::DirectoryChange { path: &root })?;
+    env::set_current_dir(&root).context(error::DirectoryChange { path: &root })?;
 
     // Compute a per-checkout prefix for the tag to avoid collisions.
     let mut d = Sha512::new();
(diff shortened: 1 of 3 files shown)


tools/buildsys/src/builder.rs
```diff
@@ -161,7 +184,7 @@ fn build(target: &str, build_args: &str, tag: &str, output: &str) -> Result<()>
     ));
 
     let create = args(format!("create --name {tag} {tag} true", tag = tag));
-    let cp = args(format!("cp {}:/output/. {}", tag, output));
+    let cp = args(format!("cp {}:/output/. {}", tag, build_dir.display()));
```
webern, line 187: Would it be better to use rust instead of cp to do a copy here?

tools/buildsys/src/builder.rs
```diff
@@ -255,6 +281,106 @@ where
...
 }
 
+/// Create a directory for build artifacts.
+fn create_build_dir(kind: &BuildType, name: &str) -> Result<PathBuf> {
+    let prefix = match kind {
+        BuildType::Package => "packages",
+        BuildType::Variant => "variants",
+    };
+
+    let path = [&getenv("BUILDSYS_STATE_DIR")?, prefix, name]
```
webern, line 291: At some point it might be nice to move all of these variables to a StructOpt, which can get its values from environment variables. This would make it easier to understand all of the inputs to the program by centralizing them.

webern (commented):
So, because we rely on Cargo to detect changes, buildsys knows that it must rebuild, but sometimes it doesn't because rpmbuild sees the output files? Therefore you we now always delete those output RPM files? Do I understand correctly? Thanks.

tools/buildsys/src/builder.rs
bcressey: This is a `docker cp` rather than a regular `cp`.

bcressey:
> So, because we rely on Cargo to detect changes, buildsys knows that it must rebuild, but sometimes it doesn't because rpmbuild sees the output files? Therefore you we now always delete those output RPM files? Do I understand correctly? 

`buildsys` would always rebuild, but the output files from previous builds were not cleaned up, and accumulated in the `build/rpms` directory, and continued to be added to a repository for `dnf` to consume.

In the related issue, we observed that if we updated `glibc` to a newer version, then reverted that change because of a problem we found during testing, the newer versioned RPMs would still be there, and `dnf` would continue to install them instead of the most recently built RPMs, since package managers care about versions rather than timestamps.

Deleting the RPMs avoids the mismatch between what we expect the build system to do - packages and images should always use the most recently built dependencies - and what it actually does.

tools/buildsys/src/builder.rs
webern: oh yeah

webern (approved):

tools/buildsys/src/builder/error.rs
```diff
@@ -16,6 +16,36 @@ pub(crate) enum Error {
...
+
+    #[snafu(display("Failed to remove file '{}': {}", path.display(), source))]
+    FileRemove {
+        path: PathBuf,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("Failed to rename file '{}': {}", path.display(), source))]
+    FileRename {
+        path: PathBuf,
```
tjkirch: The output path would be useful here too, in case it's a problem with the target directory.

tools/buildsys/src/builder.rs
```diff
@@ -11,10 +11,13 @@ use duct::cmd;
 use nonzero_ext::nonzero;
 use rand::Rng;
 use sha2::{Digest, Sha512};
-use snafu::{ensure, ResultExt};
+use snafu::{ensure, OptionExt, ResultExt};
 use std::env;
+use std::fs;
```
tjkirch:
```suggestion
use std::fs::{File, self};
```

nit: `File` is usually used without the module prefix since it's so common.

tools/buildsys/src/builder.rs
```diff
@@ -255,6 +281,106 @@ where
...
+/// Remove build artifacts from the output directory.
+fn clean_build_files<P>(build_dir: P, output_dir: P) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    fn is_marker(entry: &DirEntry) -> bool {
+        entry
+            .file_name()
+            .to_str()
+            .map(|s| s.ends_with(".marker"))
```
tjkirch:
```suggestion
            .map(|s| s.ends_with(".buildsys_marker"))
```
In case a package has a real `bla.marker` file...

On that note, it took me a while to understand what we're tracking here and the assumptions we're making.  I think we can only handle one level of files, no subdirectories, because we're not tracking anything about the path within the build / output directories.  I think it'd be worth documenting more of what's going on, example use cases, etc.

tjkirch (approved):
I don't want to block, but I do think documentation is important here.  The process makes some assumptions about the outputs of builds, and is something that could have subtle effects we're not anticipating.

Personally, I'm also worried about the removal of old artifacts, because of the impact to developer build time.  I don't think working on different branches should mean abandoning prior artifacts if relevant sources on the given branch haven't changed.  We _should_ be able to figure out use of the latest artifacts, but I know it's a hard problem, and we're solving something else here.

bcressey:
> Personally, I'm also worried about the removal of old artifacts, because of the impact to developer build time. I don't think working on different branches should mean abandoning prior artifacts if relevant sources on the given branch haven't changed. We _should_ be able to figure out use of the latest artifacts, but I know it's a hard problem, and we're solving something else here.

Essentially this code only runs when `cargo` has already decided that a package needs to be rebuilt. That should always happen today when switching between branches where the files for a package have changed. If they haven't changed, then `cargo` should not see any modifications and won't start a rebuild.

If you're seeing spurious package rebuilds, it's possible we have another issue like the one in #996 that needs to be root caused.

tools/buildsys/src/builder.rs
```diff
@@ -255,6 +281,114 @@ where
...
+    let path = [&getenv("BUILDSYS_STATE_DIR")?, prefix, name]
+        .iter()
+        .collect();
+
+    fs::create_dir_all(&path).context(error::DirectoryCreate { path: &path })?;
+
+    Ok(path)
+}
+
+static MARKER_EXTENSION: &str = ".buildsys_marker";
```
tjkirch:
```suggestion
const MARKER_EXTENSION: &str = ".buildsys_marker";
```

tjkirch (approved):

tjkirch (approved):

PR #1292, opened 2021-01-27
Add a vmware-dev variant

zmrow:
**Issue number:**
N/A

**Description of changes:**
This change adds the necessary files for a VMWare development variant.  The files are currently a copy of the `aws-dev` variant but will diverge over time.

This variant currently includes some AWS settings that are required to fully boot, but those settings will slowly disappear as we eliminate any AWS dependencies in the software.

**Testing done:**
* Build and run the variant locally (including the changes from #1288 and #1290 ) to ensure that the variant boots and correctly looks at `/media/cdrom` for userdata

**Terms of contribution:**


sources/models/src/vmware-dev/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+// that uses its name in serialization; internal structures use the field name that points to it
+#[model(rename = "settings", impl_default = true)]
+struct Settings {
+    motd: String,
+    updates: UpdatesSettings,
+    host_containers: HashMap<Identifier, ContainerImage>,
+    ntp: NtpSettings,
+    network: NetworkSettings,
+    kernel: KernelSettings,
+    aws: AwsSettings,
```
samuelkarp: might be worth a comment that we intend to remove this setting

samuelkarp (approved):

sources/models/src/vmware-dev/override-defaults.toml
```diff
@@ -0,0 +1,12 @@
+[configuration-files.containerd-config-toml]
+# No override to path
+template-path = "/usr/share/templates/containerd-config-toml_aws-dev"
```
tjkirch: Can we make a vmware-dev file instead?  If it's the same, it could be a symlink in the packages directory, but having the wrong variant name on the system at runtime feels wrong...

sources/logdog/conf/logdog.vmware-dev.conf
```diff
@@ -0,0 +1,2 @@
+exec docker-info docker info
```
tjkirch: I think we could have this be a link to the aws-dev file to reflect that we expect no differences.  (We don't expect any, right?)

tjkirch (commented):
I think we should add this to `.github/workflows/build.yml`, as an optional build like aws-dev, to make sure we don't break it as we develop.

I guess we can wait on `sources/models/src/lib.rs` (for README generation) and `variants/README.md` until there's some useful difference, but let's not forget...

[edit] Actually, the cdrom difference is there, so it might be worth doing this now?

sources/models/src/vmware-dev/override-defaults.toml
zmrow: I agree.  A symlink seems fine.

zmrow:
> I guess we can wait on sources/models/src/lib.rs (for README generation) and variants/README.md until there's some useful difference, but let's not forget...
[edit] Actually, the cdrom difference is there, so it might be worth doing this now?

Good catch - I completely missed this.

sources/logdog/conf/logdog.vmware-dev.conf
zmrow: I'll symlink it for now.  If we find differences later, we'll rectify it then.

sources/models/src/vmware-dev/mod.rs
zmrow: Per some offline conversation - we removed this completely.  :) 

Even though some system services end up failing because it's missing, it's probably important to remove the crutch early and work towards fixing whatever is still broken rather than leaving it in and having it be a pain to dig out later.

sources/models/src/vmware-dev/override-defaults.toml
zmrow: I couldn't use a symlink for this file and still reference it in the `containerd.spec` so I opted to just create an actual file.

zmrow:
^ Addressed all the comments!

sources/models/src/vmware-dev/mod.rs
tjkirch: I'm hesitant to support merging a new variant that's known to be broken.  We're definitely working toward removing this crutch, and it's easy to test what happens if we remove it...

sources/models/src/vmware-dev/override-defaults.toml
tjkirch: We should be able to use a symlink; what happened when you tried?  Can we dig in a bit?  If you post or share the changes you tried, I bet we can figure it out.

sources/models/src/lib.rs
```diff
@@ -52,6 +52,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/aws-dev/mod.rs)
```
tjkirch, line 54: Can we also update the aws-dev header to say "AWS development build" to clarify the distinction?  It won't let me make a suggestion.

Same in `variants/README.md` - the headers are currently the same there, should probably make it match this wording.  And in that file, there are clarifications for VMWare that didn't make it back into the aws-dev paragraph.  I think it'd be cleaner to have each describe its setup, instead of only saying the differences in the VMWare entry.

sources/models/src/lib.rs
```diff
@@ -52,6 +52,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
...
 * [Overridden defaults](src/aws-dev/override-defaults.toml)
 
+## vmware-dev: VMWare Development build
```
tjkirch:
```suggestion
## vmware-dev: VMWare development build
```

variants/README.md
```diff
@@ -69,6 +69,12 @@ container instance in AWS.
 The [aws-dev](aws-dev/Cargo.toml) variant has useful packages for local development of the OS.
 It includes tools for troubleshooting as well as Docker for running containers.
 
+### vmware-dev: Development build
+
+The [vmware-dev](vmware-dev/Cargo.toml) variant has useful packages for local development of the OS for running as a VMWare guest.
```
tjkirch:
```suggestion
The [vmware-dev](vmware-dev/Cargo.toml) variant has useful packages for local development of the OS, and is intended to run as a VMWare guest.
```

variants/README.md
```diff
@@ -69,6 +69,12 @@ container instance in AWS.
...
+It includes tools for troubleshooting as well as Docker for running containers, and doesn't depend on AWS or IMDS for things like user data.
+User data will be read from a mounted CD-ROM, either from a file named "user-data" or from an OVF file.
```
tjkirch:
```suggestion
It includes tools for troubleshooting as well as Docker for running containers.
User data will be read from a mounted CD-ROM, either from a file named "user-data" or from an OVF file.
```

sources/models/src/vmware-dev/mod.rs
zmrow: I can see where you're coming from, but this variant is to facilitate development.  It's hard to support fixing the variant if there's not a variant to build. :)

tjkirch: > It's hard to support fixing the variant if there's not a variant to build.

I don't understand.  The new variant would be available, would have `settings.aws` so it's not broken, and development could continue toward removing `settings.aws`.

zmrow:
^ Addressed @tjkirch 's comments

sources/models/src/vmware-dev/override-defaults.toml
bcressey: Either way we should update the path here to point to `containerd-config-toml_vmware-dev`.

sources/models/src/vmware-dev/override-defaults.toml
```diff
@@ -0,0 +1,12 @@
...
+
+# Docker
+[services.docker]
+restart-commands = ["/bin/systemctl try-restart docker.service"]
+configuration-files = ["proxy-env"]
+
+# Network
+[metadata.settings.network]
+affected-services = ["containerd", "docker", "host-containerd"]
```
bcressey: Should we also override the default host containers, since they aren't expected to work?

zmrow:
^ The above push:
* Addresses the remaining comments
* Takes advantage of the default settings refactor from #1303 
* Add a commit that fixes a small bug in `host-containers`.  Previously if a user didn't define any host containers, the program would fail.  The change ensures that the program continues on gracefully since it has nothing to do.

bcressey (approved):

sources/models/src/vmware-dev/defaults.d/50-vmware-dev.toml
```diff
@@ -0,0 +1,18 @@
...
+[services.docker]
+restart-commands = ["/bin/systemctl try-restart docker.service"]
+configuration-files = ["proxy-env"]
+
+# Network
+[metadata.settings.network]
+affected-services = ["containerd", "docker", "host-containerd"]
+
+# NTP
+# Override the 169. endpoint as we can't trust it off of AWS
```
tjkirch:
```suggestion
# Use a public endpoint, don't assume any local ones.
```

tjkirch (approved):

zmrow:
^ Integrate @tjkirch 's wording suggestion

Issue #1294, opened 2021-01-28
EKS - Support kube-bench CIS scan


felipeac:
## Expected Behavior

kube-bench report should return valid results when scanning Bottlerocket OS nodes.
```
== Summary total ==
14 checks PASS
0 checks FAIL
1 checks WARN
0 checks INFO
```
## Actual Behavior

The kube-bench scan is throwing invalid results because Bottlerocket OS places the kubelet config files in a different location and Bottlerocket doesn't have some missing properties.

**Report before applying any change to either kube-bench or Bottlerocket:**

```
== Summary node ==
1 checks PASS
13 checks FAIL
1 checks WARN
0 checks INFO

== Summary total ==
1 checks PASS
13 checks FAIL
1 checks WARN
0 checks INFO
```

**Report after applying configs changes (add right kubelet path) to kube-bench:**
**Issue on kube-bench: https://github.com/aquasecurity/kube-bench/issues/808**

```
== Summary node ==
12 checks PASS
2 checks FAIL
1 checks WARN
0 checks INFO

== Summary total ==
12 checks PASS
2 checks FAIL
1 checks WARN
0 checks INFO
```

**Report after applying the requested changes to Bottlerocket:**

```
== Summary node ==
14 checks PASS
0 checks FAIL
1 checks WARN
0 checks INFO

== Summary total ==
14 checks PASS
0 checks FAIL
1 checks WARN
0 checks INFO
```

## Steps to Reproduce the Problem

  1. Deploy an EKS cluster on AWS with a node group using Bottlerocket OS
  2. Deploy kube-bench Docker image to a container registry 
  3. Deploy job-eks.yaml to the EKS Cluster

## Specifications

  - Version: k8s 1.18
  - Platform: AWS EKS

bcressey:
Done.

PR #1295, opened 2021-01-28
Change/add params to support CIS scan

felipeac:
**Summary**

**Issue number #1294** 

This PR update configs to achieve good results on a CIS scan using kube-bench.

These were my results:

**After update the kubelet location on kube-bench**
**Issue opened on kube-bench: https://github.com/aquasecurity/kube-bench/issues/808**

```
== Summary node ==
12 checks PASS
2 checks FAIL
1 checks WARN
0 checks INFO

== Summary total ==
12 checks PASS
2 checks FAIL
1 checks WARN
0 checks INFO
```

**After building a new Bottlerocket AMI with the changes requested in this PR**
```
== Summary node ==
14 checks PASS
0 checks FAIL
1 checks WARN
0 checks INFO

== Summary total ==
14 checks PASS
0 checks FAIL
1 checks WARN
0 checks INFO
```

**Testing done:**
I've only tested it by deploying a new image and running the CIS scan, but I'm more than happy to run any required tests.

**Important: I've only validated those changes in an EKS cluster running Kubernetes 1.18. Eventually, if this PR gets approved, I'll apply and validate the same change for different versions.**

**Terms of contribution:**

PS: Thanks to @gregdek for the support

### DIFF
--- a/packages/kubernetes-1.15/kubelet-sysctl.conf
+++ b/packages/kubernetes-1.15/kubelet-sysctl.conf
@@ -1,0 +1,2 @@
+# Overcommit handling mode - 1: Always overcommit
+vm.overcommit_memory = 1
--- a/packages/kubernetes-1.15/kubernetes-1.15.spec
+++ b/packages/kubernetes-1.15/kubernetes-1.15.spec
@@ -23,6 +23,7 @@ Source5: kubernetes-ca-crt
 Source6: kubelet-exec-start-conf
 Source7: kubelet-bootstrap-kubeconfig
 Source8: kubernetes-tmpfiles.conf
+Source9: kubelet-sysctl.conf
 Source1000: clarify.toml
 Patch1: 0001-always-set-relevant-variables-for-cross-compiling.patch
 
@@ -92,6 +93,9 @@ ln -rs \
   %{buildroot}%{_sharedstatedir}/kubelet/plugins \
   %{buildroot}%{_cross_libexecdir}/kubernetes/kubelet-plugins
 
+mkdir -p %{buildroot}%{_cross_sysctldir}
+install -p -m 0644 %{S:9} %{buildroot}%{_cross_sysctldir}/90-kubelet.conf
+
 %cross_scan_attribution --clarify %{S:1000} go-vendor vendor
 
 %files -n %{_cross_os}kubelet-1.15
@@ -110,5 +114,6 @@ ln -rs \
 %{_cross_tmpfilesdir}/kubernetes.conf
 %dir %{_cross_libexecdir}/kubernetes
 %{_cross_libexecdir}/kubernetes/kubelet-plugins
+%{_cross_sysctldir}/90-kubelet.conf
 
 %changelog
--- a/packages/kubernetes-1.18/kubelet-config
+++ b/packages/kubernetes-1.18/kubelet-config
@@ -44,12 +44,14 @@ kubeReserved:
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
+readOnlyPort: 0
 cgroupDriver: systemd
 cgroupRoot: "/"
 runtimeRequestTimeout: 15m
 featureGates:
   RotateKubeletServerCertificate: true
   CSIMigration: false
+protectKernelDefaults: true
 serializeImagePulls: false
 serverTLSBootstrap: true
 configMapAndSecretChangeDetectionStrategy: Cache
--- a/packages/release/release-sysctl.conf
+++ b/packages/release/release-sysctl.conf
@@ -2,8 +2,11 @@
 # Maximize console logging level for kernel printk messages
 kernel.printk = 8 4 1 7
 
-# Wait 30 seconds and then reboot
-kernel.panic = 30
+# Wait 10 seconds and then reboot
+kernel.panic = 10
+
+# Controls the kernel's behaviour when an oops or BUG is encountered
+kernel.panic_on_oops = 1
 
 # Allow neighbor cache entries to expire even when the cache is not full
 net.ipv4.neigh.default.gc_thresh1 = 0
--- a/packages/kubernetes-1.15/kubelet-config
+++ b/packages/kubernetes-1.15/kubelet-config
@@ -44,11 +44,13 @@ kubeReserved:
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
+readOnlyPort: 0
 cgroupDriver: systemd
 cgroupRoot: "/"
 runtimeRequestTimeout: 15m
 featureGates:
   RotateKubeletServerCertificate: true
+protectKernelDefaults: true
 serializeImagePulls: false
 serverTLSBootstrap: true
 configMapAndSecretChangeDetectionStrategy: Cache
--- a/packages/kubernetes-1.16/kubelet-config
+++ b/packages/kubernetes-1.16/kubelet-config
@@ -44,11 +44,13 @@ kubeReserved:
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
+readOnlyPort: 0
 cgroupDriver: systemd
 cgroupRoot: "/"
 runtimeRequestTimeout: 15m
 featureGates:
   RotateKubeletServerCertificate: true
+protectKernelDefaults: true
 serializeImagePulls: false
 serverTLSBootstrap: true
 configMapAndSecretChangeDetectionStrategy: Cache
--- a/packages/kubernetes-1.16/kubelet-sysctl.conf
+++ b/packages/kubernetes-1.16/kubelet-sysctl.conf
@@ -1,0 +1,2 @@
+# Overcommit handling mode - 1: Always overcommit
+vm.overcommit_memory = 1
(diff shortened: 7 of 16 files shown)


packages/kubernetes-1.18/kubelet-config
```diff
@@ -20,12 +20,14 @@ clusterDNS:
...
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
+readOnlyPort: 0
 cgroupDriver: systemd
 cgroupRoot: "/"
 runtimeRequestTimeout: 15m
 featureGates:
   RotateKubeletServerCertificate: true
   CSIMigration: false
+protectKernelDefaults: true
```
tjkirch, line 54: I think @bcressey should comment on the specific settings in this PR, but separate from that, I'm almost positive we'd want to apply them to all Kubernetes versions, not just 1.18.  I don't think we should separate those changes into different PRs; we don't intend for the different versions to have different behaviors.

packages/release/release-sysctl.conf
```diff
@@ -2,8 +2,14 @@
...
-# Wait 30 seconds and then reboot
-kernel.panic = 30
+# Wait 10 seconds and then reboot
+kernel.panic = 10
+
+# Controls the kernel's behaviour when an oops or BUG is encountered
+kernel.panic_on_oops = 1
+
+# Overcommit handling mode - 1: Always overcommit
+vm.overcommit_memory = 1
```
samuelkarp: Do you have any information on why the [overcommit mode](https://www.kernel.org/doc/Documentation/vm/overcommit-accounting) would need to be changed from the default of 0 to 1?

I don't know if this would be appropriate for all variants.  In particular, this would be an unexpected change for the `aws-ecs-1` variant, as the AL2-based ECS-optimized AMIs inherit AL2's default setting of 0.

bcressey: > Do you have any information on why the [overcommit mode](https://www.kernel.org/doc/Documentation/vm/overcommit-accounting) would need to be changed from the default of 0 to 1?

Per https://github.com/kubernetes/kubernetes/issues/14935 it's done to disable the heuristics that could let a "best effort" container cause malloc to fail for a "guaranteed" container.

> I don't know if this would be appropriate for all variants.

If ECS doesn't have different QoS classes then it may not apply. I tend to want consistent behavior across variants here, though. The kernel's heuristics can introduce some surprising behaviors - running containers as root or not can give different results; applications can see unexpected malloc failures even if they're within their limits.

Allowing containers to exceed their reservation and killing them when they do is often much easier for operators to reason about.

samuelkarp: > If ECS doesn't have different QoS classes then it may not apply.

ECS does not have different QoS classes; there are no "best effort" or "guaranteed" containers.  Instead, ECS controls memory limits on containers via cgroups; every container has at least a memory limit or a memory reservation.

samuelkarp: @bcressey and I dug into this a bit more together.  The kernel behavior for `OVERCOMMIT_GUESS` (`0`, the default) [changed](https://github.com/torvalds/linux/commit/8c7829b04c523cdc732cb77f59f03320e09f3386) in Linux 5.2.  The new behavior is that allocations only fail if the requested virtual memory allocation exceeds the physical ram + swap space rather than using more complicated heuristics to guess.  This brings the behavior closer to `OVERCOMMIT_ALWAYS` (`1`), except that "wild allocations" will still fail.

Bottlerocket's kernel is 5.4, meaning that it includes this change.  With that said, my opinion is that we should leave the kernel configured with the default behavior (`OVERCOMMIT_GUESS`) in all variants; the QoS issues described in https://github.com/kubernetes/kubernetes/issues/14935 shouldn't happen in Bottlerocket.

bcressey: @samuelkarp found that my understanding is out-of-date, and the kernel documentation is no longer accurate since https://github.com/torvalds/linux/commit/8c7829b04c523cdc732cb77f59f03320e09f3386. There's no special treatment of root or processes with `CAP_SYS_ADMIN` since kernel 5.2+.

The difference is now:
 * `overcommit_memory = 0` only blocks requests for more memory than is physically present in ram + swap
 * `overcommit_memory = 1` allows all requests

I doubt the Kubernetes check is still relevant for newer kernels. It could be relaxed upstream to allow either `0` or `1` - but the application code would need to account for the running kernel version which is not recommended as a general practice.

The two paths I see are to patch our kubelets so they expect `0` instead of `1`, or ship a sysctl snippet for kubelet so that we can apply the expected value, and take advantage of the other checks for protected settings. I prefer the second approach.

bcressey:
@felipeac - can you split this into a few commits based on discussion?
1. "release: update kernel panic settings" (for the panic tuning)
2. "kubernetes: adjust kernel overcommit settings" (for the overcommit tuning)
3. "kubernetes: set readOnlyPort and protectKernelDefaults" (for the kubelet config changes)

The second one would add `packages/kubernetes-*/kubelet-sysctl.conf` and include it in all the spec files, and the third one would modify `packages/kubernetes-*/kubelet-config` with the new values.

gregdek:
Hi @felipeac just checking in to see if you need any help with the split.

felipeacr:
> Hi @felipeac just checking in to see if you need any help with the split.
Thanks, @gregdek . I've been busy at work. I'll split this PR later today.
CC: @bcressey

felipeac:
> Hi @felipeac just checking in to see if you need any help with the split.
Thanks, @gregdek . I've been busy at work. I'll split this PR later today.
CC: @bcressey

packages/release/release-sysctl.conf
```diff
@@ -2,8 +2,14 @@
 # Maximize console logging level for kernel printk messages
 kernel.printk = 8 4 1 7
 
-# Wait 30 seconds and then reboot
-kernel.panic = 30
+# Wait 10 seconds and then reboot
+kernel.panic = 10
```
sftim, line 6: I believe `kube-bench` complains if this is higher than 10s (based on https://rancher.com/docs/rancher/v2.x/en/security/rancher-2.4/hardening-2.4/#configure-kernel-runtime-parameters perhaps?)

Would a lower setting be OK; for example, 2 seconds or 5 seconds?

packages/kubernetes-1.15/kubelet-sysctl.conf
```diff
@@ -0,0 +1,2 @@
+# Overcommit handling mode - 1: Always overcommit
+vm.overcommit_memory = 1
```
bcressey: nit: should have a newline at end of file

packages/kubernetes-1.15/kubernetes-1.15.spec
```diff
@@ -79,6 +80,7 @@ install -m 0644 %{S:2} %{buildroot}%{_cross_templatedir}/kubelet-env
 install -m 0644 %{S:3} %{buildroot}%{_cross_templatedir}/kubelet-config
 install -m 0644 %{S:4} %{buildroot}%{_cross_templatedir}/kubelet-kubeconfig
 install -m 0644 %{S:5} %{buildroot}%{_cross_templatedir}/kubernetes-ca-crt
+install -m 0644 %{S:6} %{buildroot}%{_cross_templatedir}/kubelet-sysctl.conf
```
bcressey: The template directory is for templates that we render with settings.

In order for the sysctl.conf drop-in to be used directly, you'll want something like this instead:
```suggestion

mkdir -p %{buildroot}%{_cross_sysctldir}
install -p -m 0644 %{S:6} %{buildroot}%{_cross_sysctldir}/90-kubelet.conf
```

packages/kubernetes-1.15/kubernetes-1.15.spec
```diff
@@ -93,5 +95,6 @@ install -m 0644 %{S:5} %{buildroot}%{_cross_templatedir}/kubernetes-ca-crt
 %{_cross_templatedir}/kubelet-config
 %{_cross_templatedir}/kubelet-kubeconfig
 %{_cross_templatedir}/kubernetes-ca-crt
+%{_cross_templatedir}/kubelet-sysctl.conf
```
bcressey:
```suggestion
%{_cross_sysctldir}/90-kubelet.conf
```

bcressey (dismissed):

packages/release/release-sysctl.conf
bcressey: I expect it would be OK, but I have trouble picturing a scenario where it would make a big difference. The node will still take 10+ seconds to come back up from a reboot, which will usually mean a noticeable service impact.

packages/kubernetes-1.15/kubelet-sysctl.conf
felipeac: my bad.
just added it

bcressey:
@felipeac - Thanks again for your contribution here!

You've had the bad luck of touching some files - the various Kubernetes specs - that have seen more churn over the last few weeks as we've added new features.

If it helps, I'm happy to do the rebase to solve the merge conflicts, and make that last tweak from `%{_cross_templatedir}` to `%{_cross_sysctldir}`. I'd open a new pull request for it to avoid pushing to your fork, but you'd still get the credit.

Let me know if you have any objections, otherwise I'll plan to pick this up next week.

felipeac:
@bcressey - Sorry about the delay. No worries. I'll push those changes tomorrow. Thanks!

bcressey:
@felipeac - thanks! This looks good to me.

My lingering concern is around `kernel.panic_on_oops=1` in combination with the known oops in #1435, where we might convert a soft, "harmless" failure into a more permanent one.

I need to do some more testing to see if this is going to be a problem. Worst case, this would end up blocked on the backport of the newer BPF helpers to the 5.4 kernel, which is currently in progress.

bcressey (approved):

bcressey:
> My lingering concern is around kernel.panic_on_oops=1 in combination with the known oops in #1435, where we might convert a soft, "harmless" failure into a more permanent one.

Ah, that issue is a kernel "warning" so it'd be controlled by `panic_on_warn` instead. I've tested locally and confirmed that the warning doesn't cause a panic.

webern (approved):

etungsten (approved):

PR #1296, opened 2021-01-29
host-ctr: add support for authenticated image pulls from ECR Public

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
Date:   Thu Jan 28 15:25:47 2021 -0800

    host-ctr: add support for authenticated image pulls from ECR Public
    
    Add support for authenticated image pulls from ECR Public registries.
    Unlike ECR private registries, we have to use Docker's registry API for
    resolving ECR Public image references. This means having to manage
    ECR Public registry auth credentials ourselves

```

**Testing done:**

Verified that `ecrpublic::GetAuthorizationToken` calls from host-ctr are successful and I get credentials back without problems.

Verified the request header includes the base64-encoded authorization token and the image gets pulled successfully:
```
time="2021-02-01T10:02:25-08:00" level=debug msg="do request" digest="sha256:6c72cf34ed39ff40a5ad8abd57c66e6766897f311a28f6b993f4b87545195204" mediatype=application/vnd.docker.distribution.manifest.list.v2+json 
request.header.accept="application/vnd.docker.distribution.manifest.list.v2+json, */*" 
request.header.authorization="Bearer  <BASE64-ENCODED AUTHORIZATION TOKEN>"
request.header.user-agent=containerd/1.3.7+unknown 
request.method=GET 
size=743 
url="https://public.ecr.aws/v2/nginx/nginx/manifests/sha256:6c72cf34ed39ff40a5ad8abd57c66e6766897f311a28f6b993f4b87545195204"
time="2021-02-01T10:02:25-08:00" level=debug msg="fetch response received" host=public.ecr.aws response.header.content-length=743 response.header.content-type=application/vnd.docker.distribution.manifest.list.v2+json response.header.date="Mon, 01 Feb 2021 18:02:25 GMT" response.header.docker-content-digest="sha256:6c72cf34ed39ff40a5ad8abd57c66e6766897f311a28f6b993f4b87545195204" response.header.docker-distribution-api-version=registry/2.0 response.status="200 OK" url="https://public.ecr.aws/v2/nginx/nginx/manifests/latest"
....
```

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/base64"
 	"io/ioutil"
 	"math/rand"
 	"os"
@@ -11,6 +12,9 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ecrpublic"
 	"github.com/awslabs/amazon-ecr-containerd-resolver/ecr"
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/cio"
@@ -20,6 +24,7 @@ import (
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/oci"
+	"github.com/containerd/containerd/remotes/docker"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -664,19 +669,68 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
 
 // withDynamicResolver provides an initialized resolver for use with ref.
 func withDynamicResolver(ctx context.Context, ref string) containerd.RemoteOpt {
-	if !strings.HasPrefix(ref, "ecr.aws/") {
-		// not handled here
-		return func(_ *containerd.Client, _ *containerd.RemoteContext) error { return nil }
-	}
+	noOp := func(_ *containerd.Client, _ *containerd.RemoteContext) error { return nil }
 
-	return func(_ *containerd.Client, c *containerd.RemoteContext) error {
-		// Create the ECR resolver
-		resolver, err := ecr.NewResolver()
-		if err != nil {
-			return errors.Wrap(err, "Failed to create ECR resolver")
+	switch {
+	// For ECR registries, we need to use the Amazon ECR resolver
+	case strings.HasPrefix(ref, "ecr.aws/"):
+		return func(_ *containerd.Client, c *containerd.RemoteContext) error {
+			// Create the Amazon ECR resolver
+			resolver, err := ecr.NewResolver()
+			if err != nil {
+				return errors.Wrap(err, "Failed to create ECR resolver")
+			}
+			log.G(ctx).WithField("ref", ref).Info("pulling with Amazon ECR Resolver")
+			c.Resolver = resolver
+			return nil
 		}
-		log.G(ctx).WithField("ref", ref).Info("pulling with Amazon ECR Resolver")
-		c.Resolver = resolver
-		return nil
+	// For Amazon ECR Public registries, we should try and fetch credentials before resolving the image reference
+	case strings.HasPrefix(ref, "public.ecr.aws/"):
+		// Try to get credentials for authenticated pulls from ECR Public
+		session := session.Must(session.NewSession())
+		// The ECR Public API is only available in us-east-1 today
+		publicConfig := aws.NewConfig().WithRegion("us-east-1")
+		client := ecrpublic.New(session, publicConfig)
+		output, err := client.GetAuthorizationToken(&ecrpublic.GetAuthorizationTokenInput{})
+		if err != nil {
+			log.G(ctx).Warn("ecr-public: failed to get authorization token, falling back to default resolver (unauthenticated pull)")
+			return noOp
+		}
+		if output == nil || output.AuthorizationData == nil {
+			log.G(ctx).Warn("ecr-public: missing AuthorizationData in ECR Public GetAuthorizationToken response, falling back to default resolver (unauthenticated pull)")
+			return noOp
+		}
+		authToken, err := base64.StdEncoding.DecodeString(aws.StringValue(output.AuthorizationData.AuthorizationToken))
+		if err != nil {
+			log.G(ctx).Warn("ecr-public: unable to decode authorization token, falling back to default resolver (unauthenticated pull)")
+			return noOp
+		}
+		tokens := strings.SplitN(string(authToken), ":", 2)
+		if len(tokens) != 2 {
+			log.G(ctx).Warn("ecr-public: invalid credentials decoded from authorization token, falling back to default resolver (unauthenticated pull)")
+			return noOp
+		}
+		// Use the fetched authorization credentials to resolve the image
+		authOpt := docker.WithAuthCreds(func(host string) (string, string, error) {
+			// Double-check to make sure the we're doing this for an ECR Public registry
+			if host != "public.ecr.aws" {
+				return "", "", errors.New("ecr-public: expected image to start with public.ecr.aws")
+			}
+			return tokens[0], tokens[1], nil
+		})
+		authorizer := docker.NewDockerAuthorizer(authOpt)
+		resolverOpt := docker.ResolverOptions{
+			Hosts: docker.ConfigureDefaultRegistries(docker.WithAuthorizer(authorizer)),
+		}
+
+		return func(_ *containerd.Client, c *containerd.RemoteContext) error {
+			resolver := docker.NewResolver(resolverOpt)
+			log.G(ctx).WithField("ref", ref).Info("pulling from ECR Public")
+			c.Resolver = resolver
+			return nil
+		}
+	default:
+		// For all other registries
+		return noOp
 	}
 }
--- a/sources/host-ctr/go.mod
+++ b/sources/host-ctr/go.mod
@@ -5,7 +5,7 @@ go 1.12
 require (
 	github.com/Microsoft/go-winio v0.4.15 // indirect
 	github.com/Microsoft/hcsshim v0.8.10 // indirect
-	github.com/aws/aws-sdk-go v1.35.25 // indirect
+	github.com/aws/aws-sdk-go v1.37.0
 	github.com/awslabs/amazon-ecr-containerd-resolver v0.0.0-20200922205237-bbd7175f7bd0
 	github.com/containerd/cgroups v0.0.0-20201109155418-13abef5d31ec // indirect
 	github.com/containerd/containerd v1.3.7
@@ -24,7 +24,6 @@ require (
 	github.com/willf/bitset v1.1.11 // indirect
 	go.etcd.io/bbolt v1.3.5 // indirect
 	go.opencensus.io v0.22.5 // indirect
-	golang.org/x/net v0.0.0-20201110031124-69a78807bb2b // indirect
 	golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9 // indirect
 	golang.org/x/sys v0.0.0-20201109165425-215b40eba54c // indirect
 	golang.org/x/text v0.3.4 // indirect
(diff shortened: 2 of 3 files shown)


sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,61 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
...
+			// Create the Amazon ECR resolver
+			resolver, err := ecr.NewResolver()
+			if err != nil {
+				return errors.Wrap(err, "Failed to create ECR resolver")
+			}
+			log.G(ctx).WithField("ref", ref).Info("pulling with Amazon ECR Resolver")
+			c.Resolver = resolver
+			return nil
+		}
+	} else if strings.HasPrefix(ref, "public.ecr.aws") {
```
samuelkarp:
```suggestion
	} else if strings.HasPrefix(ref, "public.ecr.aws/") {
```

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,61 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
...
+		// For Amazon ECR Public registries, we need to fetch credentials before resolving the image reference
+		// TODO Potentially need to implement credentials caching to avoid throttling during large deployments
```
samuelkarp: If we do credential caching, we'll need a bit more of a design for handling those credentials securely.  It might be worth removing this comment and opening an issue for it instead.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,61 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
...
+		authOpt := docker.WithAuthCreds(func(host string) (string, string, error) {
+			session := session.Must(session.NewSession())
```
samuelkarp: I think it's worth re-validating that the provided input matches ECR Public.
```suggestion
		authOpt := docker.WithAuthCreds(func(host string) (string, string, error) {
			if !strings.HasPrefix(host, "public.ecr.aws/") {
				return "", "", errors.New("ecr-public: expected image to start with public.ecr.aws")
			}
			session := session.Must(session.NewSession())
```

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,61 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
...
+			log.G(ctx).WithField("ref", ref).Info("pulling from ECR Public")
+			c.Resolver = resolver
+			return nil
 		}
-		log.G(ctx).WithField("ref", ref).Info("pulling with Amazon ECR Resolver")
-		c.Resolver = resolver
-		return nil
+
 	}
+	// TODO Handle ECR Public registries with custom domains
```
samuelkarp: This doesn't exist, so I'd remove the comment.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,61 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
 
 // withDynamicResolver provides an initialized resolver for use with ref.
 func withDynamicResolver(ctx context.Context, ref string) containerd.RemoteOpt {
-	if !strings.HasPrefix(ref, "ecr.aws/") {
-		// not handled here
-		return func(_ *containerd.Client, _ *containerd.RemoteContext) error { return nil }
-	}
+	if strings.HasPrefix(ref, "ecr.aws/") {
```
samuelkarp: nit: I have a slight bias toward using `switch` statements instead of if/else chains

```go
switch {
case strings.HasPrefix(ref, "ecr.aws/"):
	// body
case strings.HasPrefix(ref, "public.ecr.aws/"):
	// body
default:
	// body
}
```

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,61 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
...
+		// For Amazon ECR Public registries, we need to fetch credentials before resolving the image reference
+		// TODO Potentially need to implement credentials caching to avoid throttling during large deployments
+		authOpt := docker.WithAuthCreds(func(host string) (string, string, error) {
+			session := session.Must(session.NewSession())
+			// The ECR Public API is only available in us-east-1 today
+			publicConfig := aws.NewConfig().WithRegion("us-east-1")
+			client := ecrpublic.New(session, publicConfig)
+
+			var input *ecrpublic.GetAuthorizationTokenInput
+			output, err := client.GetAuthorizationToken(input)
```
samuelkarp:
```suggestion
			output, err := client.GetAuthorizationToken(&ecrpublic.GetAuthorizationTokenInput{})
```

or

```suggestion
			output, err := client.GetAuthorizationToken(nil)
```

etungsten:
Push above and below addresses all of @samuelkarp 's comments.

etungsten:
Push above fixes a small issue with double checking the host part of the url

zmrow (approved):
:hugs:

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,62 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
...
+			return nil
+		}
+	// For Amazon ECR Public registries, we need to fetch credentials before resolving the image reference
+	case strings.HasPrefix(ref, "public.ecr.aws/"):
+		authOpt := docker.WithAuthCreds(func(host string) (string, string, error) {
+			// Double-check to make sure the we're doing this for an ECR Public registry
+			if !strings.HasPrefix(host, "public.ecr.aws") {
+				return "", "", errors.New("ecr-public: expected image to start with public.ecr.aws")
+			}
+			session := session.Must(session.NewSession())
```
webern: What are the pre-conditions for this to work? The instance needs an instance role? Same as for the old resolver, right?

srgothi92 (approved):

sources/host-ctr/cmd/host-ctr/main.go
etungsten: Yes, the instance needs to be launched with an instance role profile that has the ECRPublic:GetAuthorizationToken permission. For the normal resolver, the role just needs the managed policy for ECR (not public). All other registries just uses the default resolver.

I think this brings up a good point on how we should make host-ctr more cloud-agnostic. I don't think it's easy to make a determination at runtime to check if we're running in EC2 (or at least not in a hacky way). So this is something that needs to be conditionally compiled... I can follow up with a issue to investigate that further if that's alright.

etungsten: Or maybe it just makes sense to fallback to using the default resolver if we fail a call to ecrpublic:GetAuthorizationToken. I think that's probably better and makes things easier. I'll go ahead and do that.

samuelkarp: Since ECR Public specifically allows unauthenticated requests (albeit with more throttling), a lack of credentials/permission should fall-through to an anonymous pull.

etungsten:
Push above and below makes it so that we fallback to using the default resolver with no creds if we can't get creds for ECR Public.

Tested things and they work as they should. If I don't have permission to get authorization creds, host-ctr correctly falls back to using the default resovler:
```
Feb 02 00:22:51 host-ctr[4640]: time="2021-02-02T00:22:51Z" level=warning msg="ecr-public: failed to get authorization token, falling back to default resolver (unauthenticated pull)"
Feb 02 00:22:52 host-ctr[4640]: time="2021-02-02T00:22:52Z" level=info msg="pulled image successfully" img="public.ecr.aws/nginx/nginx:latest"
```

samuelkarp:
> If I don't have permission to get authorization creds, host-ctr correctly falls back to using the default resovler

Can you also explicitly test the case where no AWS credentials are available as opposed to the case where AWS credentials are available but don't have permission?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,68 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
...
+		}
+		tokens := strings.SplitN(string(authToken), ":", 2)
+		if len(tokens) != 2 {
+			log.G(ctx).Warn("ecr-public: got invalid authorization token, expecting token of the form 'username:password', falling back to default resolver (unauthenticated pull)")
+			return noOpt
+		}
+		// Use the fetched authorization credentials to resolve the image
+		authOpt := docker.WithAuthCreds(func(host string) (string, string, error) {
+			// Double-check to make sure the we're doing this for an ECR Public registry
+			if !strings.HasPrefix(host, "public.ecr.aws") {
```
samuelkarp: Do you just get the hostname as a parameter here or is it the full image?

If it's just the hostname, this would be better:
```suggestion
			if host != "public.ecr.aws" {
```

If it's the full image name, this would be better:
```suggestion
			if !strings.HasPrefix(host, "public.ecr.aws/") {
```

etungsten: It's just the hostname. I'll go ahead and fix it.

etungsten:
Push above addresses @samuelkarp 's latest comments.

> > If I don't have permission to get authorization creds, host-ctr correctly falls back to using the default resovler
> 
> Can you also explicitly test the case where no AWS credentials are available as opposed to the case where AWS credentials are available but don't have permission?

I moved my `~/.aws/` directory to a random location and tried running host-ctr locally. It was not able to fetch credentials and fell back using the default resolver as expected:
```
time="2021-02-01T20:33:17-08:00" level=warning msg="ecr-public: failed to get authorization token, falling back to default resolver (unauthenticated pull)"
time="2021-02-01T20:33:17-08:00" level=debug msg=resolving host=public.ecr.aws
time="2021-02-01T20:33:17-08:00" level=debug msg="do request" host=public.ecr.aws request.header.accept="application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, */*" request.header.user-agent=containerd/1.3.7+unknown request.method=HEAD url="https://public.ecr.aws/v2/nginx/nginx/manifests/latest"
```

samuelkarp (approved):
:ship:

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,68 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
...
 func withDynamicResolver(ctx context.Context, ref string) containerd.RemoteOpt {
-	if !strings.HasPrefix(ref, "ecr.aws/") {
-		// not handled here
-		return func(_ *containerd.Client, _ *containerd.RemoteContext) error { return nil }
-	}
-
-	return func(_ *containerd.Client, c *containerd.RemoteContext) error {
-		// Create the ECR resolver
-		resolver, err := ecr.NewResolver()
+	noOpt := func(_ *containerd.Client, _ *containerd.RemoteContext) error { return nil }
```
jahkeup: *nit*

Was this name supposed to be "no-op" or no opt? No-op seems more apt in this case.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -664,19 +669,68 @@ func tagImage(ctx context.Context, imageName string, newImageName string, client
...
+		authToken, err := base64.StdEncoding.DecodeString(aws.StringValue(output.AuthorizationData.AuthorizationToken))
+		if err != nil {
+			log.G(ctx).Warn("ecr-public: got invalid authorization token, falling back to default resolver (unauthenticated pull)")
+			return noOpt
+		}
+		tokens := strings.SplitN(string(authToken), ":", 2)
+		if len(tokens) != 2 {
+			log.G(ctx).Warn("ecr-public: got invalid authorization token, expecting token of the form 'username:password', falling back to default resolver (unauthenticated pull)")
+			return noOpt
+		}
```
jahkeup: > ecr-public: got invalid authorization token, expecting token of the form 'username:password', 

To me, this error suggests that - as a user - I am able to resolve the issue of a poorly formatted set of credentials (ie: a problem the user might be able to tweak and fix). However, the services API's response would be the concern at this point so I'd rather that be where we point to.

```suggestion
		authToken, err := base64.StdEncoding.DecodeString(aws.StringValue(output.AuthorizationData.AuthorizationToken))
		if err != nil {
			log.G(ctx).Warn("ecr-public: unable to decode authorization token, falling back to default resolver (unauthenticated pull)")
			return noOpt
		}
		tokens := strings.SplitN(string(authToken), ":", 2)
		if len(tokens) != 2 {
			log.G(ctx).Warn("ecr-public: invalid credentials decoded from authorization token, falling back to default resolver (unauthenticated pull)")
			return noOpt
		}
```

jahkeup (changes requested):
Looks great! 👍🏽 

I would like to see the error messages slightly tweaked to avoid the potential of misleading users in some situations before merging - let me know what you think.

webern (approved):

sources/host-ctr/cmd/host-ctr/main.go
etungsten: I was originally going for no-option, but I think you're right that no op is more befitting in this case.

etungsten:
Push above addresses @jahkeup 's comments.

jahkeup (approved):

Issue #1307, opened 2021-02-09
Support a separate ENI/Instance-Type Map list for Custom Networking modes


diranged:
**What I'd like:**
For users running the EKS AWS VPC CNI in "[Custom Networking](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html)" mode, the supplied [packages/os/eni-max-pods](https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/os/eni-max-pods) file does not work. The mappings are incorrect because you have to remove an ENI from the math.

Ideally either BottleRocket would be smart enough to do that math for us (unlikely), or at a minimum it could include a second ENI-to-InstanceType map file and give us the option to select that file instead.

bcressey:
I suppose the way I'd approach this is by having one ASG or managed node group per instance type, and then just overriding `settings.kubernetes.max-pods` via user-data with the right value for that instance type within the launch template.

Is that unworkable for some reason?

diranged:
@bcressey,
  So - if you operate with the Kubernetes cluster-autoscaler, they specifically encourage you to use ASGs exactly as you described. However, if you operate with a third party management tool like Spot.io, you are encouraged to have a wide range of instance types in your mix so that Spot.io's Ocean product can scale up the "right" sized hosts for the workloads. We operate in both - we use ASGs for our "core system" pods, but we use Ocean to scale out our developers workloads.

  I sort of feel like if we're already willing to bake in a "max pods" list into the EKS and BottlerocketEKS AMIs, baking in a second "max pods in custom networking mode" list is very reasonable to do. I've prototyped it locally by building our own image with the new list (generated in https://github.com/bottlerocket-os/bottlerocket/pull/1309) and it works.

  I would like to build a better PR where I create a flag that lets you define the "eni pod list" source.. (even optionally allowing you to curl it or something) .. but I am finding the rust templating system really tough to understand for some reason, so I got a bit lost. I would happily help finish up the PR though with a bit of guidance on how to add in options?

  By the way.. here's the bash-code we implemented on our non-bottlerocket EKS AMIs so that we would dynamically generate the calculation on-bootup:
```
# https://github.com/awslabs/amazon-eks-ami/pull/494/files#diff-049390d14bc3ea2d7882ff0f108e2802ad9b043336c5fa637e93581d9a7fdfc2
# * First IP on each ENI is not used for pods
[...]

PR #1308, opened 2021-02-09
ND: Implement a separate ENI MaxPods list for clusters using AWS VPC CNI Networking with Custom Subnets

diranged:
**Issue number:**

https://github.com/bottlerocket-os/bottlerocket/issues/1307

**Description of changes:**

This is a partial change. Basically we are trying to solve the issue in #1307 by creating a new ENI mapping file. We used a bit of Python to generate the list from the published data on the AWS site, and I have included that script here. The reason this is a partial change is that I am not familiar enough with the templating setup to understand how to add in my own option to be able to select which file to use. I could use some suggestions on that.

**Testing done:**

None yet - this is just a test so that we can discuss the solution.

**Terms of contribution:**


diranged:
(Closing this PR and opening on a new branch where I can more easily develop. Sorry for the spam.)

PR #1264, opened 2021-01-07
Package updates

etungsten:
**Issue number:**
N/A

**Description of changes:**
Updates third-party dependencies in `packages/`.

Excluding ecs-agent as that is being covered by https://github.com/bottlerocket-os/bottlerocket/pull/1201
Excluding containerd 1.4 which is a major update that requires further inspection and testing.
systemd v247 and dbus-broker 25 update is in https://github.com/bottlerocket-os/bottlerocket/pull/1265

```
* 1d7baa43 - kubernetes-1.18: update to 1.18.14 <Erikson Tung>
* 3edc27cc - kubernetes-1.17: update to 1.17.13 <Erikson Tung>
* 2a7cdcf1 - docker-proxy: update to 19.03.14 <Erikson Tung>
* 2c8c9e72 - docker-engine: update to 19.03.14 <Erikson Tung>
* 5351e16a - docker-cli: update to 19.03.14 <Erikson Tung>
* 944d54b3 - tcpdump: update to 4.99.0 <Erikson Tung>
* c90749f7 - socat: update to 1.7.4.0 <Erikson Tung>
* d765c8c3 - readline: update to 8.1 <Erikson Tung>
* e432874e - libseccomp: update to 2.5.1 <Erikson Tung>
* 0289a32f - libpcre: update to 10.36 <Erikson Tung>
* d29c8f6e - libcap: update to 2.44 <Erikson Tung>
* b4b9c360 - libaudit: update to 3.0 <Erikson Tung>
* 65f4af13 - cni-plugins: update to 0.9.0 <Erikson Tung>
* 30c58e3e - ca-certificates: update to 2020.12.08 <Erikson Tung>
```

**Testing done:**

- [x] aws-k8s-1.15, aws-k8s-1.16, aws-k8s-1.17, aws-k8s-1.18, aws-k8s-1.19 on x86 and arm64
- Can run pods OK
- kubelet status look good, journal looks good
- journal in general looks normal

- [x] aws-ecs-1
- Runs tasks ok

- [x] aws-dev
- Runs containers ok

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #1315, opened 2021-02-10
drop Docker patch for privileged and SELinux

bcressey:
**Issue number:**
Follow-up to #1013

**Description of changes:**
This removes a patch that's no longer needed on newer versions of Docker, which vendor `go-selinux` after v1.0.0 and include the fix from https://github.com/opencontainers/selinux/commit/fd7b61c28f9e134d36a5911eb1550f2704e4bf8e. That ensures that root filesystems of privileged containers still end up with a writable mount label. 

**Testing done:**
Built the `aws-dev` variant and exercised the cases from #1011.

**Privileged:**
```
docker run -it --privileged amazonlinux:2 bash

bash-4.2# touch test
bash-4.2# head -n1 /proc/mounts
overlay / overlay rw,context="system_u:object_r:local_t:s0:c159,c466",...
bash-4.2# cat /proc/self/attr/current
system_u:system_r:container_t:s0
```
 
**Label disabled:**
```
docker run -it --security-opt label:disable amazonlinux:2 bash

bash-4.2# touch test
bash-4.2# head -n1 /proc/mounts
overlay / overlay rw,context="system_u:object_r:local_t:s0:c12,c282",...
bash-4.2# cat /proc/self/attr/current
system_u:system_r:container_t:s0
```
 
**Normal:**
```
docker run -it amazonlinux:2 bash

bash-4.2# touch test
bash-4.2# head -n1 /proc/mounts
overlay / overlay rw,context="system_u:object_r:local_t:s0:c17,c466",...
bash-4.2# cat /proc/self/attr/current
system_u:system_r:container_t:s0:c17,c466
```

Note that the `system_u:system_r:container_t:s0` label assigned to the "privileged" and "label disabled" containers are in theory less privileged than a label like `system_u:system_r:container_t:s0:c17,c466`.

This isn't a problem yet since we don't enforce MCS separation, but will need to be corrected separately by adjusting the SELinux policy.

**Terms of contribution:**


samuelkarp:
This patch is no longer needed after https://github.com/bottlerocket-os/bottlerocket/pull/1264?

bcressey:
> This patch is no longer needed after #1264?

It wasn't needed after #1025 but we didn't realize it at the time, since it had just been added in #1013.

samuelkarp (approved):

tjkirch (approved):

PR #1316, opened 2021-02-11
add aliases for container-selinux types

bcressey:
**Issue number:**
N/A

**Description of changes:**
This aliases `container-selinux` policy types such as `spc_t` to our own policy types.

That lets us drop one of the Kubernetes patches we've been carrying for a while, and makes it easier to port pod and task definitions from Amazon Linux to Bottlerocket.

**Testing done:**
Verified that a pod with `spc_t` set in its security context ended up running as `control_t`.

Confirmed that there were no warnings about an invalid `container_file_t` context with the Kubernetes patch reverted, which was what it was originally intended to fix per 2e111e355443da140b7c5aa31c3f18e779f04af4.

**Terms of contribution:**


samuelkarp (approved):
I'm not 100% wild about this as now we've made our labels tied to the container-selinux labels, but it does provide a better experience for moving workloads to Bottlerocket.

tjkirch (approved):

PR #1317, opened 2021-02-11
k8s static pods support

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
Date:   Tue Feb 9 13:20:22 2021 -0800

    settings, static-pods: new settings and helper for k8s static pods
    
    Adds a set of new settings for managing k8s static pods.
    Adds a new binary helper to manage k8s static pod manifests.

```

```
Date:   Tue Feb 9 13:27:48 2021 -0800

    aws-k8s-*: specify 'staticPodPath' in kubelet config
    
    Specify pod manifest path in kubelet config for all kubernetes variants.

```

```
Date:   Fri Feb 12 12:17:29 2021 -0800

    pluto: only build for kubernetes variants
    
    We only need to build pluto for kubernetes variants

```

```
Date:   Fri Feb 12 13:23:43 2021 -0800

    Makefile: conditionally exclude packages from cargo test
    
    In tasks.unit-test, we need to conditionally exclude variant specific
    packages from being compiled if the variant does not match up

```

```
Date:   Tue Feb 16 10:41:34 2021 -0800

    migrations: add `add-static-pods` migration
    
    Adds a new migration for new k8s static-pods settings

```

**Testing done:**

Verified that buildsys does not build the package for non-k8s variants.

Build aws-k8s-1.15, aws-k8s-1.16, aws-k8s-1.17, aws-k8s-1.18, aws-k8s-1.19 images and launched instances for all of them

Kubelet started up fine.

Then I specified a static pod that ran nginx on port 80 with host networking:
```
apiclient set kubernetes.static-pods.control-plane-stuff.manifest=YXBpVmVyc2lvbjogdjEKa2luZDogUG9kCm1ldGFkYXRhOgogICBuYW1lOiBzdGF0aWMtcG9kLW5naW54LXRlc3QKc3BlYzoKICAgaG9zdE5ldHdvcms6IHRydWUKICAgY29udGFpbmVyczoKICAgICAgLSBuYW1lOiBuZ2lueAogICAgICAgIGltYWdlOiBuZ2lueDoxLjE0LjIKICAgICAgICBwb3J0czoKICAgICAgICAgICAtIGNvbnRhaW5lclBvcnQ6IDgwCg==
apiclient set kubernetes.static-pods.control-plane-stuff.enabled=true
```

The base64 blob decodes to the following:
```
apiVersion: v1
kind: Pod
metadata:
   name: static-pod-nginx-test
spec:
   hostNetwork: true
   containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
           - containerPort: 80

```

Observed that the manifest gets created successfully in the pod manifest path:
```
bash-5.0# cat /etc/kubernetes/static-pods/control-plane-stuff 
apiVersion: v1
kind: Pod
metadata:
   name: static-pod-nginx-test
spec:
   hostNetwork: true
   containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
           - containerPort: 80
```

Then curling localhost:80 from within the admin container I can see nginx is running as expected:
```
[ec2-user@ip-192-168-3-187 ~]$ curl localhost:80
<!DOCTYPE html>
<html>
<head>
<style>
    body {
        width: 35em;
        margin: 0 auto;
        font-family: Tahoma, Verdana, Arial, sans-serif;
    }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>

</body>
</html>
[ec2-user@ip-192-168-3-187 ~]$ 
```

Checking pods via kubectl, I can see mirror pod entries in the k8s API:
```
$ kubectl get pods -A
NAMESPACE      NAME                                                                READY   STATUS    RESTARTS   AGE
default        static-pod-nginx-test-ip-192-168-3-187.us-west-2.compute.internal   1/1     Running   0          6m16s
default        static-pod-nginx-test-ip-192-168-8-47.us-west-2.compute.internal    1/1     Running   0          20m
....
```

Disabling the static pod by setting `enabled` to `false` also works as expected. The manifest file gets correctly deleted.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,10 @@ The following settings can be optionally set to customize the node labels and ta
 The following settings are optional and allow you to further configure your cluster.
 * `settings.kubernetes.cluster-domain`: The DNS domain for this cluster, allowing all Kubernetes-run containers to search this domain before the host's search domains.  Defaults to `cluster.local`.
 
+You can also optionally specify static pods for your node with the following settings.
+* `settings.kubernetes.static-pods.<custom identifier>.manifest`: A base64-encoded pod manifest.
+* `settings.kubernetes.static-pods.<custom identifier>.enabled`: Whether the static pod is enabled.
+
 The following settings are set for you automatically by [pluto](sources/api/) based on runtime instance information, but you can override them if you know what you're doing!
 * `settings.kubernetes.max-pods`: The maximum number of pods that can be scheduled on this node (limited by number of available IPv4 addresses)
 * `settings.kubernetes.cluster-dns-ip`: The CIDR block of the primary network interface.
--- a/packages/kubernetes-1.15/kubernetes-tmpfiles.conf
+++ b/packages/kubernetes-1.15/kubernetes-tmpfiles.conf
@@ -1,0 +1,2 @@
+d /etc/kubernetes/static-pods - - - -
+L /etc/kubernetes/manifests - - - - static-pods
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -74,7 +74,9 @@ Requires: %{_cross_os}apiserver = %{version}-%{release}
 Summary: Updates settings dynamically based on user-specified generators
 Requires: %{_cross_os}apiserver = %{version}-%{release}
 Requires: %{_cross_os}schnauzer = %{version}-%{release}
+%if %{_is_k8s_variant}
 Requires: %{_cross_os}pluto = %{version}-%{release}
+%endif
 Requires: %{_cross_os}bork = %{version}-%{release}
 %description -n %{_cross_os}sundog
 %{summary}.
@@ -93,11 +95,6 @@ Requires: %{_cross_os}apiserver = %{version}-%{release}
 %package -n %{_cross_os}schnauzer
 Summary: Setting generator for templated settings values.
 %description -n %{_cross_os}schnauzer
-%{summary}.
-
-%package -n %{_cross_os}pluto
-Summary: Dynamic setting generator for kubernetes
-%description -n %{_cross_os}pluto
 %{summary}.
 
 %package -n %{_cross_os}thar-be-settings
@@ -184,6 +181,11 @@ Summary: Settings generator for ECS
 %endif
 
 %if %{_is_k8s_variant}
+%package -n %{_cross_os}pluto
+Summary: Dynamic setting generator for kubernetes
+%description -n %{_cross_os}pluto
+%{summary}.
+
 %package -n %{_cross_os}static-pods
 Summary: Manages user-defined K8S static pods
 Requires: %{_cross_os}apiserver = %{version}-%{release}
@@ -203,7 +205,6 @@ mkdir bin
     -p netdog \
     -p sundog \
     -p schnauzer \
-    -p pluto \
     -p bork \
     -p thar-be-settings \
     -p thar-be-updates \
@@ -223,6 +224,7 @@ mkdir bin
     -p ecs-settings-applier \
 %endif
 %if %{_is_k8s_variant}
+    -p pluto \
     -p static-pods \
 %endif
     %{nil}
@@ -246,7 +248,7 @@ done
 install -d %{buildroot}%{_cross_bindir}
 for p in \
   apiserver \
-  early-boot-config netdog sundog schnauzer pluto bork corndog \
+  early-boot-config netdog sundog schnauzer bork corndog \
   thar-be-settings thar-be-updates servicedog host-containers \
   storewolf settings-committer \
   migrator \
@@ -256,7 +258,7 @@ for p in \
   ecs-settings-applier \
 %endif
 %if %{_is_k8s_variant}
-  static-pods \
+  pluto static-pods \
 %endif
 ; do
   install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/${p} %{buildroot}%{_cross_bindir}
@@ -292,8 +294,10 @@ install -d %{buildroot}%{_cross_datadir}/bottlerocket
 install -d %{buildroot}%{_cross_sysusersdir}
 install -p -m 0644 %{S:2} %{buildroot}%{_cross_sysusersdir}/api.conf
 
+%if %{_is_k8s_variant}
 install -d %{buildroot}%{_cross_datadir}/eks
 install -p -m 0644 %{S:3} %{buildroot}%{_cross_datadir}/eks
+%endif
 
 install -d %{buildroot}%{_cross_datadir}/updog
 install -p -m 0644 %{_cross_repo_root_json} %{buildroot}%{_cross_datadir}/updog
@@ -346,11 +350,6 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 
 %files -n %{_cross_os}schnauzer
 %{_cross_bindir}/schnauzer
-
-%files -n %{_cross_os}pluto
-%{_cross_bindir}/pluto
-%dir %{_cross_datadir}/eks
-%{_cross_datadir}/eks/eni-max-pods
 
 %files -n %{_cross_os}bork
 %{_cross_bindir}/bork
@@ -419,6 +418,11 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 %endif
 
 %if %{_is_k8s_variant}
+%files -n %{_cross_os}pluto
+%{_cross_bindir}/pluto
+%dir %{_cross_datadir}/eks
+%{_cross_datadir}/eks/eni-max-pods
+
 %files -n %{_cross_os}static-pods
 %{_cross_bindir}/static-pods
 %endif
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -1,4 +1,5 @@
 %global _cross_first_party 1
+%global _is_k8s_variant %(if echo %{_cross_variant} | grep -q "k8s"; then echo 1; else echo 0; fi)
 %undefine _debugsource_packages
 
 Name: %{_cross_os}os
@@ -182,6 +183,14 @@ Summary: Settings generator for ECS
 %{summary}.
 %endif
 
+%if %{_is_k8s_variant}
+%package -n %{_cross_os}static-pods
+Summary: Manages user-defined K8S static pods
+Requires: %{_cross_os}apiserver = %{version}-%{release}
+%description -n %{_cross_os}static-pods
+%{summary}.
+%endif
+
 %prep
 %setup -T -c
 %cargo_prep
@@ -213,6 +222,9 @@ mkdir bin
 %if "%{_cross_variant}" == "aws-ecs-1"
     -p ecs-settings-applier \
 %endif
+%if %{_is_k8s_variant}
+    -p static-pods \
+%endif
     %{nil}
 
 # Next, build components that should be static.
@@ -242,6 +254,9 @@ for p in \
   ghostdog \
 %if "%{_cross_variant}" == "aws-ecs-1"
   ecs-settings-applier \
+%endif
+%if %{_is_k8s_variant}
+  static-pods \
 %endif
 ; do
   install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/${p} %{buildroot}%{_cross_bindir}
@@ -401,6 +416,11 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 %if "%{_cross_variant}" == "aws-ecs-1"
 %files -n %{_cross_os}ecs-settings-applier
 %{_cross_bindir}/ecs-settings-applier
+%endif
+
+%if %{_is_k8s_variant}
+%files -n %{_cross_os}static-pods
+%{_cross_bindir}/static-pods
 %endif
 
 %changelog
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -63,7 +63,6 @@ Requires: %{_cross_os}selinux-policy
 Requires: %{_cross_os}policycoreutils
 Requires: %{_cross_os}signpost
 Requires: %{_cross_os}sundog
-Requires: %{_cross_os}pluto
 Requires: %{_cross_os}storewolf
 Requires: %{_cross_os}host-containers
 Requires: %{_cross_os}settings-committer
--- a/sources/api/static-pods/src/main.rs
+++ b/sources/api/static-pods/src/main.rs
@@ -1,0 +1,16 @@
+#![deny(rust_2018_idioms)]
+
+#[cfg(k8s_variant)]
+mod static_pods;
+#[cfg(k8s_variant)]
+#[macro_use]
+extern crate log;
+
+#[cfg(k8s_variant)]
+#[tokio::main]
+async fn main() {
+    static_pods::main().await
+}
+
+#[cfg(not(k8s_variant))]
+fn main() {}
--- a/sources/api/static-pods/src/static_pods.rs
+++ b/sources/api/static-pods/src/static_pods.rs
@@ -1,0 +1,281 @@
+/*!
+# Background
+
+static-pods ensures static pods are running as defined in settings.
+
+It queries for all existing static pod settings, then configures the system as follows:
+* If the pod is enabled, it creates the manifest file in the pod manifest path that kubelet is
+  configured to read from and populates the file with the base64-decoded manifest setting value.
+* If the pod is enabled and the manifest file already exists, it overwrites the existing manifest
+  file with the base64-decoded manifest setting value.
+* If the pod is disabled, it ensures the manifest file is removed from the pod manifest path.
+*/
+
+use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
+use snafu::{ensure, OptionExt, ResultExt};
+use std::collections::HashMap;
+use std::env;
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process;
+use std::str::FromStr;
+
+use model::modeled_types::Identifier;
+
+// FIXME Get from configuration in the future
+const DEFAULT_API_SOCKET: &str = "/run/api.sock";
+
+const STATIC_POD_DIR: &str = "/etc/kubernetes/static-pods";
+
+type Result<T> = std::result::Result<T, error::Error>;
+
+/// Query the API for the currently defined static pods
+async fn get_static_pods<P>(socket_path: P) -> Result<Option<HashMap<Identifier, model::StaticPod>>>
+where
+    P: AsRef<Path>,
+{
+    debug!("Requesting settings values");
+    let settings = schnauzer::get_settings(socket_path)
+        .await
+        .context(error::RetrieveSettings)?
+        .settings
+        .context(error::MissingSettings)?;
+
+    Ok(settings
+        .kubernetes
+        .context(error::MissingSettings)?
+        .static_pods)
+}
+
+/// Write out the manifest file to the pod manifest path with a given filename
+fn write_manifest_file<S1, S2>(name: S1, manifest: S2) -> Result<()>
+where
+    S1: AsRef<str>,
+    S2: AsRef<[u8]>,
+{
+    let name = name.as_ref();
+
+    let dir = Path::new(STATIC_POD_DIR);
+    fs::create_dir_all(&dir).context(error::Mkdir { dir: &dir })?;
+    let path = dir.join(name);
+    // Create the file if it does not exist, completely replace its contents
+    // if it does (constitutes an update).
+    fs::write(path, manifest).context(error::ManifestWrite { name })?;
+
+    Ok(())
+}
+
+/// Deletes the named manifest file if it exists
+fn delete_manifest_file<S1>(name: S1) -> Result<()>
+where
+    S1: AsRef<str>,
+{
+    let name = name.as_ref();
+    let path = Path::new(STATIC_POD_DIR).join(name);
+    if path.exists() {
+        fs::remove_file(path).context(error::ManifestDelete { name })?;
+    }
+
+    Ok(())
+}
+
+fn handle_static_pod<S>(name: S, pod_info: &model::StaticPod) -> Result<()>
+where
+    S: AsRef<str>,
+{
+    // Get basic settings, as retrieved from API.
+    let name = name.as_ref();
+    let enabled = pod_info.enabled.context(error::MissingField {
+        name,
+        field: "enabled",
+    })?;
+
+    if enabled {
+        let manifest = pod_info.manifest.as_ref().context(error::MissingField {
+            name,
+            field: "manifest",
+        })?;
+
+        let manifest = base64::decode(manifest.as_bytes()).context(error::Base64Decode {
+            base64_string: manifest.as_ref(),
+        })?;
+
+        info!("Writing static pod '{}' to '{}'", name, STATIC_POD_DIR);
+
+        // Write the manifest file for this static pod
+        write_manifest_file(name, manifest)?;
+    } else {
+        info!("Removing static pod '{}' from '{}'", name, STATIC_POD_DIR);
+
+        // Delete the manifest file so the static pod no longer runs (disabled)
+        delete_manifest_file(name)?;
+    }
+
+    Ok(())
+}
+
+async fn run() -> Result<()> {
+    let args = parse_args(env::args())?;
+
+    // SimpleLogger will send errors to stderr and anything less to stdout.
+    SimpleLogger::init(args.log_level, LogConfig::default()).context(error::Logger)?;
+
+    info!("static-pods started");
+
+    let mut failed = 0u32;
+    if let Some(static_pods) = get_static_pods(args.socket_path).await? {
+        for (name, pod) in static_pods.iter() {
+            // Continue to handle other static pods if we fail one
+            if let Err(e) = handle_static_pod(name, pod) {
+                failed += 1;
+                error!("Failed to handle static pod '{}': {}", &name, e);
+            }
+        }
+
+        ensure!(
+            failed == 0,
+            error::ManageStaticPodsFailed {
+                failed,
+                tried: static_pods.len()
+            }
+        );
+    }
+
+    Ok(())
+}
+
+/// Store the args we receive on the command line
+struct Args {
+    log_level: LevelFilter,
+    socket_path: PathBuf,
+}
+
+/// Print a usage message in the event a bad arg is passed
+fn usage() {
+    let program_name = env::args().next().unwrap_or_else(|| "program".to_string());
+    eprintln!(
+        r"Usage: {}
+            [ --socket-path PATH ]
+            [ --log-level trace|debug|info|warn|error ]
+
+    Socket path defaults to {}",
+        program_name, DEFAULT_API_SOCKET,
+    );
+}
+
+/// Parse the args to the program and return an Args struct
+fn parse_args(args: env::Args) -> Result<Args> {
+    let mut log_level = None;
+    let mut socket_path = None;
+
+    let mut iter = args.skip(1);
+    while let Some(arg) = iter.next() {
+        match arg.as_ref() {
+            "--log-level" => {
+                let log_level_str = iter.next().ok_or_else(|| error::Error::Usage {
+                    message: "Did not give argument to --log-level".into(),
+                })?;
+                log_level = Some(LevelFilter::from_str(&log_level_str).map_err(|_| {
+                    error::Error::Usage {
+                        message: format!("Invalid log level '{}'", log_level_str),
+                    }
+                })?);
+            }
+
+            "--socket-path" => {
+                socket_path = Some(
+                    iter.next()
+                        .ok_or_else(|| error::Error::Usage {
+                            message: "Did not give argument to --socket-path".into(),
+                        })?
+                        .into(),
+                )
+            }
+
+            _ => {
+                return Err(error::Error::Usage {
+                    message: "unexpected argument".into(),
+                })
+            }
+        }
+    }
+
+    Ok(Args {
+        log_level: log_level.unwrap_or_else(|| LevelFilter::Info),
+        socket_path: socket_path.unwrap_or_else(|| DEFAULT_API_SOCKET.into()),
+    })
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+pub(crate) async fn main() -> () {
+    if let Err(e) = run().await {
+        match e {
+            error::Error::Usage { .. } => {
+                eprintln!("{}", e);
+                usage();
+                process::exit(2);
+            }
+            _ => {
+                eprintln!("{}", e);
+                process::exit(1);
+            }
+        }
+    }
+}
+
+mod error {
+    use snafu::Snafu;
+    use std::path::PathBuf;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility = "pub(super)")]
+    pub(super) enum Error {
+        #[snafu(display("{}", message))]
+        Usage { message: String },
+
+        #[snafu(display("Failed to retrieve settings: {}", source))]
+        RetrieveSettings { source: schnauzer::Error },
+
+        #[snafu(display("settings.kubernetes.static_pods missing in API response"))]
+        MissingSettings {},
+
+        #[snafu(display("Static pod '{}' missing field '{}'", name, field))]
+        MissingField { name: String, field: String },
+
+        #[snafu(display("Failed to manage {} of {} static pods", failed, tried))]
+        ManageStaticPodsFailed { failed: u32, tried: usize },
+
+        #[snafu(display("Logger setup error: {}", source))]
+        Logger { source: log::SetLoggerError },
+
+        #[snafu(display("Unable to base64 decode manifest '{}': '{}'", base64_string, source))]
+        Base64Decode {
+            base64_string: String,
+            source: base64::DecodeError,
+        },
+
+        #[snafu(display("Failed to create directory '{}': '{}'", dir.display(), source))]
+        Mkdir {
+            dir: PathBuf,
+            source: std::io::Error,
+        },
+
+        #[snafu(display("Failed to write manifest for static pod '{}': {}", name, source))]
+        ManifestWrite {
+            name: String,
+            source: std::io::Error,
+        },
+
+        #[snafu(display(
+            "Failed to delete manifest file for static pod '{}': {}'",
+            name,
+            source
+        ))]
+        ManifestDelete {
+            name: String,
+            source: std::io::Error,
+        },
+    }
+}
--- a/sources/models/src/aws-k8s-1.15/defaults.d/50-aws-k8s.toml
+++ b/sources/models/src/aws-k8s-1.15/defaults.d/50-aws-k8s.toml
@@ -41,6 +41,13 @@ cluster-domain = "cluster.local"
 [settings.metrics]
 service-checks = ["apiserver", "chronyd", "containerd", "host-containerd", "kubelet"]
 
+[services.static-pods]
+configuration-files = []
+restart-commands = ["/usr/bin/static-pods"]
+
+[metadata.settings.kubernetes.static-pods]
+affected-services = ["static-pods"]
+
 # Network
 [metadata.settings.network]
 affected-services = ["containerd", "kubernetes", "host-containerd"]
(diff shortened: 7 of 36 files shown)


etungsten:
I need to figure out how to only build the `api/static-pods` crate if variant is set to *-k8s-*

etungsten:
Push above adds some documentation to the settings section of the top-level README.

README.md
```diff
@@ -304,6 +304,24 @@ The following settings can be optionally set to customize the node labels and ta
...
+```
+apiclient set kubernetes.static-pods.nginx.manifest=YXBpVmVyc2lvbjogdjEKa2luZDogUG9kCm1ldGFkYXRhOgogICBuYW1lOiBzdGF0aWMtcG9kLW5naW54LXRlc3QKc3BlYzoKICAgaG9zdE5ldHdvcms6IHRydWUKICAgY29udGFpbmVyczoKICAgICAgLSBuYW1lOiBuZ2lueAogICAgICAgIGltYWdlOiBuZ2lueDoxLjE0LjIKICAgICAgICBwb3J0czoKICAgICAgICAgICAtIGNvbnRhaW5lclBvcnQ6IDgwCg==
+apiclient set kubernetes.static-pods.nginx.enabled=true
+```
+
+Here's the same example, but with the settings you'd add to user data:
+```
+[settings.static-pods.nginx]
+enabled = true
+manifest = YXBpVmVyc2lvbjogdjEKa2luZDogUG9kCm1ldGFkYXRhOgogICBuYW1lOiBzdGF0aWMtcG9kLW5naW54LXRlc3QKc3BlYzoKICAgaG9zdE5ldHdvcms6IHRydWUKICAgY29udGFpbmVyczoKICAgICAgLSBuYW1lOiBuZ2lueAogICAgICAgIGltYWdlOiBuZ2lueDoxLjE0LjIKICAgICAgICBwb3J0czoKICAgICAgICAgICAtIGNvbnRhaW5lclBvcnQ6IDgwCg==
```
```suggestion
manifest = "YXBpVmVyc2lvbjogdjEKa2luZDogUG9kCm1ldGFkYXRhOgogICBuYW1lOiBzdGF0aWMtcG9kLW5naW54LXRlc3QKc3BlYzoKICAgaG9zdE5ldHdvcms6IHRydWUKICAgY29udGFpbmVyczoKICAgICAgLSBuYW1lOiBuZ2lueAogICAgICAgIGltYWdlOiBuZ2lueDoxLjE0LjIKICAgICAgICBwb3J0czoKICAgICAgICAgICAtIGNvbnRhaW5lclBvcnQ6IDgwCg=="
```

packages/kubernetes-1.15/kubernetes-tmpfiles.conf
```diff
@@ -0,0 +1 @@
+d /etc/kubernetes/static-pods - - - -
```
bcressey, line 1: `/etc/kubernetes/manifests` is the path used by kubeadm, so it might make sense to have that for consistency with the rest of the ecosystem.

Rather than change anything elsewhere I'd just add a symlink here:
```suggestion
d /etc/kubernetes/static-pods - - - -
L /etc/kubernetes/manifests - - - - static-pods
```

packages/os/os.spec
```diff
@@ -234,6 +246,9 @@ for p in \
 %if "%{_cross_variant}" == "aws-ecs-1"
   ecs-settings-applier \
 %endif
+%if "%{_is_k8s_variant}"
+  static-pods \
+%endif
```
bcressey: We should probably also move `pluto` here in a separate commit; it doesn't belong on the ECS variant.

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,16 @@
...
+extern crate log;
+
+#[cfg(k8s_variant)]
+#[tokio::main]
+async fn main() {
+    static_pods::main().await
+}
+
+#[cfg(not(k8s_variant))]
+fn main() {}
```
bcressey, line 16: Do we still want to no-op the build if we're not building it in the spec file for the non-Kubernetes case?

sources/api/static-pods/src/static_pods.rs
```diff
@@ -0,0 +1,320 @@
...
+const DEFAULT_API_SOCKET: &str = "/run/api.sock";
+const API_SETTINGS_URI: &str = "/settings";
+
+const STATIC_POD_DIR: &str = "/etc/kubernetes/static-pods";
+
+mod error {
+    use http::StatusCode;
+    use snafu::Snafu;
+    use std::path::PathBuf;
+    use std::process::{Command, Output};
```
bcressey: nit: elsewhere we've started moving this module to the end so that the more interesting code is first.

sources/api/static-pods/src/static_pods.rs
```diff
@@ -0,0 +1,320 @@
...
+    S2: AsRef<[u8]>,
+{
+    let name = name.as_ref();
+
+    let dir = Path::new(STATIC_POD_DIR);
+    fs::create_dir_all(&dir).context(error::Mkdir { dir: &dir })?;
+    let path = dir.join(name);
+    // Create the file if it does not exist, completely replace its contents
+    // if it does (constitutes an update).
+    fs::write(path, manifest).context(error::ManifestWrite { name })?;
```
bcressey, line 63: Does re-applying the same manifest to a static pod, and rewriting it to the file, cause kubelet to restart it because mtime changed? Or does it do a deeper comparison of file contents and leave the pod running?

If it's not a deep check, we could wind up restarting all static pods whenever a new one is added, enabled, or disabled, which would be overly disruptive.

sources/api/static-pods/src/static_pods.rs
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    Ok(())
+}
+
+/// Store the args we receive on the command line
+struct Args {
+    log_level: LevelFilter,
+    socket_path: PathBuf,
+}
```
bcressey: nit: we can move Args, usage(), usage_msg(), and parse_args() down below run() since they're mostly boilerplate.

sources/api/static-pods/src/static_pods.rs
```diff
@@ -0,0 +1,320 @@
...
+        field: "manifest",
+    })?;
+
+    if enabled {
+        let manifest = base64::decode(manifest.as_bytes()).context(error::Base64Decode {
+            base64_string: manifest.as_ref(),
+        })?;
+
+        info!(
+            "Handling static pod '{}' which is enabled: {}",
```
bcressey: Might be better to say "Writing static pod '{}' to $PATH" here, and "Removing static pod '{}' from $PATH" in the else branch - we can infer the value of the boolean from the verb choice.

sources/api/static-pods/src/static_pods.rs
```diff
@@ -0,0 +1,320 @@
...
+    // Get basic settings, as retrieved from API.
+    let name = name.as_ref();
+    let enabled = pod_info.enabled.context(error::MissingField {
+        name,
+        field: "enabled",
+    })?;
+    let manifest = pod_info.manifest.as_ref().context(error::MissingField {
+        name,
+        field: "manifest",
+    })?;
```
bcressey: nit: we don't technically need to do this check unless the pod is enabled.

This is weird but arguably harmless input:
```
[settings.kubernetes.static-pods.my-empty-pod]
enabled = false
```

packages/release/release.spec
```diff
@@ -76,6 +77,9 @@ Requires: %{_cross_os}logdog
 Requires: %{_cross_os}util-linux
 Requires: %{_cross_os}wicked
 Requires: %{_cross_os}os
+%if "%{_is_k8s_variant}"
+Requires: %{_cross_os}static-pods
+%endif
```
bcressey: This should be listed in the variant definitions, like `ecs-settings-applier` is for the ecs variant. `release` is meant to be the keystone package that's common across all variants.

It's also not marked as `variant-sensitive = true` in its Cargo.toml, so we're not guaranteed that it will be rebuilt when the variant changes.

packages/os/os.spec
```diff
@@ -1,4 +1,5 @@
 %global _cross_first_party 1
+%global _is_k8s_variant %(if echo %{_cross_variant} | grep -q "k8s"; then echo 1; else echo 0; fi)
```
bcressey, line 2: 🥇

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,16 @@
...
+#[macro_use]
+extern crate log;
+
+#[cfg(k8s_variant)]
+#[tokio::main]
+async fn main() {
+    static_pods::main().await
+}
+
+#[cfg(not(k8s_variant))]
```
zmrow, line 15: Not that it hurts anything, but if we are only going to include this binary in k8s variants, do we need the extra protection of conditional compilation?

sources/api/static-pods/src/static_pods.rs
```diff
@@ -0,0 +1,320 @@
...
+
+use model::modeled_types::Identifier;
+
+// FIXME Get from configuration in the future
+const DEFAULT_API_SOCKET: &str = "/run/api.sock";
+const API_SETTINGS_URI: &str = "/settings";
+
+const STATIC_POD_DIR: &str = "/etc/kubernetes/static-pods";
+
+mod error {
```
zmrow: FYI we've been moving this module to the bottom in more recent code. :)

sources/api/static-pods/src/static_pods.rs
```diff
@@ -0,0 +1,320 @@
...
+    let path = Path::new(STATIC_POD_DIR).join(name);
+    if path.exists() {
+        fs::remove_file(path).context(error::ManifestDelete { name })?;
+    }
+
+    Ok(())
+}
+
+/// Store the args we receive on the command line
+struct Args {
```
zmrow: nit: these arg parsing functions seem to split up the main logic functions of the program.  `handle_static_pod()` is all by its lonesome below!  I might consider grouping the arg parsing and other logic functions together.

sources/api/static-pods/src/static_pods.rs
```diff
@@ -0,0 +1,320 @@
...
+        info!(
+            "Handling static pod '{}' which is enabled: {}",
+            name, enabled
+        );
+
+        // Write the manifest file for this static pod
+        write_manifest_file(name, manifest)?;
+    } else {
+        // Delete the manifest file so the static pod no longer runs (disabled)
+        delete_manifest_file(name)?;
```
zmrow, line 111: Does kubelet monitor these files?  If you delete the file, is kubelet smart enough to stop the pod?  Same for file updates; is kubelet smart enought to stop/start the pod with new configuration, etc.?

sources/api/static-pods/src/static_pods.rs
etungsten: Kubelet doesn't restart, nor does the pod if the manifest doesn't change (including after applying the same manifest). The pod only restarts when kubelet sees the manifest change for the pod after a periodic scan (it's a deeper comparison). When adding a new static pod, the other pods aren't disrupted either.

sources/api/static-pods/src/static_pods.rs
etungsten: Yes, kubelet monitors the pod manifest path for changes. kubelet does stop the pod if the corresponding manifest is deleted. It also knows when to restart the pod if the manifest changes.

sources/api/static-pods/src/main.rs
etungsten: aws-ecs-1 and aws-dev were building this package and failing the build if I didn't put this in. Because the settings model is variant specific and will not include `KuberneteSettings` which `static-pods` depends on as an import.

packages/release/release.spec
etungsten: Should I go ahead and remove pluto from this list as well?

sources/api/static-pods/src/main.rs
etungsten: See my response here: https://github.com/bottlerocket-os/bottlerocket/pull/1317#discussion_r575451214

sources/api/static-pods/src/main.rs
etungsten: Turns out `_is_k8s_variant` is not working as it should...

etungsten: Ok, it's actually the `unit-test` make target that's failing the build check which makes more sense. Not sure how to fix that though without conditional compilation.

zmrow (commented):
:musical_score:

zmrow (approved):

srgothi92 (commented):
Nice 🎉 

Just out of curiosity and may not be relevant at all to the changes.
I was wondering what would happen if someone sends incorrect manifest file ? I think Kubernetes would not launch a pod for it. However, how can someone debug such cases.

etungsten:
> Nice tada
> 
> Just out of curiosity and may not be relevant at all to the changes.
> I was wondering what would happen if someone sends incorrect manifest file ? I think Kubernetes would not launch a pod for it. However, how can someone debug such cases.

If a user sends a poorly formatted manifest file, kubelet would just ignore that manifest. To debug, by virtue of not seeing the pod come up, the user would need to recheck their manifest. If they would like to see kubelet's log, then they'd have to go through logdog.

sources/api/static-pods/src/main.rs
etungsten: I fixed this in https://github.com/bottlerocket-os/bottlerocket/pull/1317/commits/a4b06089235de53c775d0545e2eb6de5eb54836a by conditionally excluding packages based on the variant.

etungsten:
The force push  from fda4739 to e3e9e1a addresses @zmrow and @bcressey 's comments.
https://github.com/bottlerocket-os/bottlerocket/pull/1317/commits/a4b06089235de53c775d0545e2eb6de5eb54836a above fixes the build check by modifying `tasks.unit-test` to conditionally exclude variant specific packages if the variant isn't aligned.

Tested and things work as expected.

etungsten:
I'll remove the conditional compilation from ecs-settings-applier and do the same for it in `tasks.unit-tests` in a separate PR.

packages/release/release.spec
etungsten: I went ahead and removed pluto from `release` and added them to k8s variant's Cargo.toml

README.md
```diff
@@ -304,6 +304,24 @@ The following settings can be optionally set to customize the node labels and ta
...
+
+Here's an example of adding a nginx static pod with API calls:
+```
+apiclient set kubernetes.static-pods.nginx.manifest=YXBpVmVyc2lvbjogdjEKa2luZDogUG9kCm1ldGFkYXRhOgogICBuYW1lOiBzdGF0aWMtcG9kLW5naW54LXRlc3QKc3BlYzoKICAgaG9zdE5ldHdvcms6IHRydWUKICAgY29udGFpbmVyczoKICAgICAgLSBuYW1lOiBuZ2lueAogICAgICAgIGltYWdlOiBuZ2lueDoxLjE0LjIKICAgICAgICBwb3J0czoKICAgICAgICAgICAtIGNvbnRhaW5lclBvcnQ6IDgwCg==
+apiclient set kubernetes.static-pods.nginx.enabled=true
+```
+
+Here's the same example, but with the settings you'd add to user data:
+```
+[settings.static-pods.nginx]
```
etungsten: missing kubernetes in the settings path...

etungsten:
Push above addresses https://github.com/bottlerocket-os/bottlerocket/pull/1317#discussion_r575565954

etungsten:
Push above rebases onto develop to fix merging conflicts.

etungsten:
Push above changes the logging facility from TermLogger to SimpleLogger. Tested things and they still work as expected.

README.md
```diff
@@ -304,6 +304,24 @@ The following settings can be optionally set to customize the node labels and ta
...
+apiclient set kubernetes.static-pods.nginx.manifest=YXBpVmVyc2lvbjogdjEKa2luZDogUG9kCm1ldGFkYXRhOgogICBuYW1lOiBzdGF0aWMtcG9kLW5naW54LXRlc3QKc3BlYzoKICAgaG9zdE5ldHdvcms6IHRydWUKICAgY29udGFpbmVyczoKICAgICAgLSBuYW1lOiBuZ2lueAogICAgICAgIGltYWdlOiBuZ2lueDoxLjE0LjIKICAgICAgICBwb3J0czoKICAgICAgICAgICAtIGNvbnRhaW5lclBvcnQ6IDgwCg==
+apiclient set kubernetes.static-pods.nginx.enabled=true
+```
+
+Here's the same example, but with the settings you'd add to user data:
+```
+[settings.kubernetes.static-pods.nginx]
+enabled = true
+manifest = "YXBpVmVyc2lvbjogdjEKa2luZDogUG9kCm1ldGFkYXRhOgogICBuYW1lOiBzdGF0aWMtcG9kLW5naW54LXRlc3QKc3BlYzoKICAgaG9zdE5ldHdvcms6IHRydWUKICAgY29udGFpbmVyczoKICAgICAgLSBuYW1lOiBuZ2lueAogICAgICAgIGltYWdlOiBuZ2lueDoxLjE0LjIKICAgICAgICBwb3J0czoKICAgICAgICAgICAtIGNvbnRhaW5lclBvcnQ6IDgwCg=="
+```
```
bcressey: I'd prefer not to include examples for this; the very long lines of base64 are somewhat jarring on in our main README.

It's also a bit weird showing both user-data and `apiclient` examples, since we don't do that for other settings.

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,323 @@
...
+fn write_manifest_file<S1, S2>(name: S1, manifest: S2) -> Result<()>
+where
+    S1: AsRef<str>,
+    S2: AsRef<[u8]>,
+{
+    let name = name.as_ref();
+
+    let dir = Path::new(STATIC_POD_DIR);
+    fs::create_dir_all(&dir).context(error::Mkdir { dir: &dir })?;
+    let path = dir.join(name);
```
bcressey: nit: should we append a `.yaml` suffix?

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,323 @@
...
+            base64_string: manifest.as_ref(),
+        })?;
+
+        info!("Writing static pod '{}' to '{}'", name, STATIC_POD_DIR);
+
+        // Write the manifest file for this static pod
+        write_manifest_file(name, manifest)?;
+    } else {
+        // Delete the manifest file so the static pod no longer runs (disabled)
+        delete_manifest_file(name)?;
```
bcressey:
```suggestion
        info!("Removing static pod '{}' from '{}'", name, STATIC_POD_DIR);

        // Delete the manifest file so the static pod no longer runs (disabled)
        delete_manifest_file(name)?;
```

sources/api/static-pods/src/main.rs
etungsten: Manifests can be in JSON as well.

etungsten:

sources/api/static-pods/src/main.rs
bcressey: 😧

sources/models/src/aws-k8s-1.15/defaults.d/50-aws-k8s.toml
```diff
@@ -41,6 +41,13 @@ cluster-domain = "cluster.local"
 [settings.metrics]
 service-checks = ["apiserver", "chronyd", "containerd", "host-containerd", "kubelet"]
 
+[services.static-pods]
+configuration-files = []
+restart-commands = ["/usr/bin/static-pods"]
+
+[metadata.settings.kubernetes.static-pods]
+affected-services = ["static-pods"]
+
```
bcressey, line 50: We'll need migrations to clean these up on downgrade.

README.md
```diff
@@ -304,6 +304,10 @@ The following settings can be optionally set to customize the node labels and ta
 The following settings are optional and allow you to further configure your cluster.
 * `settings.kubernetes.cluster-domain`: The DNS domain for this cluster, allowing all Kubernetes-run containers to search this domain before the host's search domains.  Defaults to `cluster.local`.
 
+You can also optionally specify static pods for your node with the following settings.
+* `settings.kubernetes.static-pods.<custom identifier>.manifest`: Base64-blob of the pod's manifest.
```
samuelkarp: nit: "base64-encoded blob"

README.md
```diff
@@ -304,6 +304,10 @@ The following settings can be optionally set to customize the node labels and ta
...
+* `settings.kubernetes.static-pods.<custom identifier>.enabled`: Whether the static pod is enabled or not.
```
samuelkarp: nit: drop "or not" to match the description of host container settings

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,325 @@
...
+use std::collections::HashMap;
+use std::env;
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process;
+use std::str::FromStr;
+
+use model::modeled_types::Identifier;
+
+// FIXME Get from configuration in the future
```
samuelkarp: Are you planning to address this as part of this PR or as a follow-up?

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,325 @@
...
+/// Query the API for the currently defined static pods
+async fn get_static_pods<P>(socket_path: P) -> Result<Option<HashMap<Identifier, model::StaticPod>>>
+where
+    P: AsRef<Path>,
+{
+    debug!("Querying the API for settings");
+
+    let method = "GET";
+    let uri = API_SETTINGS_URI;
+    let (code, response_body) = apiclient::raw_request(&socket_path, uri, method, None)
```
samuelkarp: Why this instead of `schnauzer::get_settings`?

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,325 @@
...
+
+/// Deletes the named manifest file if it exists
+fn delete_manifest_file<S1>(name: S1) -> Result<()>
+where
+    S1: AsRef<str>,
+{
+    let name = name.as_ref();
+    let path = Path::new(STATIC_POD_DIR).join(name);
+    if path.exists() {
+        fs::remove_file(path).context(error::ManifestDelete { name })?;
```
samuelkarp: Do we care about the race here?  Do we expect that there will only be one copy of `static-pods` running at a time?  It looks like `run` just logs failures and proceeds through the rest of the static pods, which looks to be the safe approach rather than bailing on the first error.

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,325 @@
...
+
+async fn run() -> Result<()> {
+    let args = parse_args(env::args());
+
+    // SimpleLogger will send errors to stderr and anything less to stdout.
+    SimpleLogger::init(args.log_level, LogConfig::default()).context(error::Logger)?;
+
+    info!("static-pods started");
+
+    let mut failed = 0usize;
```
samuelkarp: Why `usize` here?  Do we want this to vary on different architectures based on the pointer size?

sources/api/static-pods/src/main.rs
etungsten: As a follow up, this FIXME shows up in other places where we set the same const vars for common file paths.

README.md
webern: Or, "a base64-encoded pod manifest"?

sources/api/static-pods/src/main.rs
etungsten: There should be only one copy of static-pods running at one time.

variants/aws-k8s-1.15/Cargo.toml
```diff
@@ -16,6 +16,8 @@ included-packages = [
     "cni-plugins",
     "kubelet-1.15",
     "release",
+    "static-pods",
+    "pluto"
```
webern, line 20: Interesting. Not now, but I wonder if there's a `k8s.spec` and `k8s` package for these in our future.

Makefile.toml
```diff
@@ -201,11 +201,17 @@ script = [
 '''
 export VARIANT="${BUILDSYS_VARIANT}"
 
+# Exclude Kubernetes specific packages from being built if the variant isn't Kubernetes specific
+if ! echo "${VARIANT}" | grep -q "k8s"; then
+    EXCLUDE="--exclude static-pods pluto"
+fi
+
```
webern: This feels so far away from other conditionals. :(
I think this also means that I can no longer run:

```sh
cd sources
export VARIANT=aws-ecs-1 && cargo build
```

Which makes me sad (and might have IDE ramifications, not sure). Are we sure this is how we want to do this?

sources/api/static-pods/src/main.rs
etungsten: I'm not sure. I'll go ahead and change it to a regular unsigned int.

sources/api/static-pods/src/main.rs
etungsten: `schnauzer::get_settings` returns the whole `Model` for template rendering purposes. The fields in `Model` are private so I have to go through an request to retrieve the field I need.

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,325 @@
...
+            [ --log-level trace|debug|info|warn|error ]
+
+    Socket path defaults to {}",
+        program_name, DEFAULT_API_SOCKET,
+    );
+    process::exit(2);
+}
+
+/// Prints a more specific message before exiting through usage().
+fn usage_msg<S: AsRef<str>>(msg: S) -> ! {
```
webern:
```suggestion
fn usage_msg<S: AsRef<str>>(msg: S) {
```

I wish we would get away from this untestable arg parsing pattern and make `main` the only function that can exit the program. I realize most other components follow this pattern, but here is an alternative approach with arg parsing under test:

```rust
/// Print a usage message in the event a bad arg is passed
fn usage() {
    let program_name = env::args().next().unwrap_or_else(|| "program".to_string());
    eprintln!(
        r"Usage: {}
            [ --socket-path PATH ]
            [ --log-level trace|debug|info|warn|error ]
    Socket path defaults to {}",
        program_name, "whatever",
    );
}

/// Prints a more specific message before exiting through usage().
fn usage_msg<S: AsRef<str>>(msg: S) {
    eprintln!("{}\n", msg.as_ref());
    usage();
}

/// Parse the args to the program and return an Args struct
fn parse_args<I>(args: I) -> Result<Args>
where
    I: Iterator<Item = String>,
{
    let mut log_level = None;
    let mut socket_path = None;

    let mut iter = args.skip(1);
    while let Some(arg) = iter.next() {
        match arg.as_ref() {
            "--log-level" => {
                let log_level_str = iter
                    .next()
                    .ok_or_else(|| Error::Usage("Did not give argument to --log-level".into()))?;
                log_level =
                    Some(LevelFilter::from_str(&log_level_str).map_err(|_| {
                        Error::Usage(format!("Invalid log level '{}'", log_level_str))
                    })?);
            }

            "--socket-path" => {
                socket_path = Some(
                    iter.next()
                        .ok_or_else(|| {
                            Error::Usage("Did not give argument to --socket-path".into())
                        })?
                        .into(),
                )
            }

            _ => return Err(Error::Usage("unexpected argument".into())),
        }
    }

    Ok(Args {
        log_level: log_level.unwrap_or_else(|| LevelFilter::Info),
        socket_path: socket_path.unwrap_or_else(|| "socket".into()),
    })
}

fn run<I>(args: I) -> Result<()>
where
    I: Iterator<Item = String>,
{
    let _args_struct = parse_args(args);
    Ok(())
}

fn main() {
    if let Err(e) = run(env::args()) {
        match e {
            Error::Usage(message) => {
                usage_msg(message);
                process::exit(2);
            }
            _ => {
                eprintln!("{}", e);
                process::exit(1);
            }
        }
    }
}

type Result<T> = std::result::Result<T, Error>;

#[derive(Debug)]
enum Error {
    Usage(String),
    EverythingElse,
}

#[test]
fn parse_args_mising_level() {
    let args = ["whatever", "--level-filter"].iter().map(|s| s.to_string());
    let result = parse_args(args);
    assert!(result.is_err());
    assert!(matches!(result.err().unwrap(), Error::Usage(_)))
}
```

sources/api/static-pods/src/main.rs
etungsten: I see why now. For some reason the `ensure!` macro needs the predicate expression to use usize comparison...

<details>

```
error[E0277]: the trait bound `usize: std::convert::From<u32>` is not satisfied
   --> api/static-pods/src/main.rs:155:9
    |
155 | /         ensure!(
156 | |             failed == 0,
157 | |             error::ManageStaticPodsFailed {
158 | |                 failed,
159 | |                 tried: static_pods.len()
160 | |             }
161 | |         );
    | |__________^ the trait `std::convert::From<u32>` is not implemented for `usize`
    |
    = help: the following implementations were found:
              <usize as std::convert::From<bool>>
              <usize as std::convert::From<mio::poll::ReadinessState>>
              <usize as std::convert::From<mio::token::Token>>
              <usize as std::convert::From<std::num::NonZeroUsize>>
            and 4 others
    = note: required because of the requirements on the impl of `std::convert::Into<usize>` for `u32`
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

</details>

etungsten: Nvm I'm dumb! Fixed it

etungsten:
d409af8adac486f904bc4216bdf8328fa2a31bb9 adds the migration for migrating the new static-pods settings.
The force push above addresses @samuelkarp 's comments.

Tested the migration and it works as expected on downgrades and upgrades.

sources/models/src/aws-k8s-1.15/defaults.d/50-aws-k8s.toml
etungsten: Added the migration in https://github.com/bottlerocket-os/bottlerocket/pull/1317/commits/d409af8adac486f904bc4216bdf8328fa2a31bb9

Makefile.toml
etungsten: There has to be a conditional somewhere for this unit test task to work. Either we continue conditionally compiling variant specific binaries, or we do this.

webern: Yeah, I realize that's the problem we're solving for. With ecs-settings-applier, the build.rs exports a cfg and the 'guts' are not compiled unless its ecs.

The problem with your approach is that the rust workspace is no longer usable on its own and we're moving toward a reliance on the makefile for things to work at all.

Both solutions are undesirable, so I'm not demanding a change here, but inviting discussion.

sources/api/static-pods/src/main.rs
etungsten: Would you like me to do this in this PR or follow up to fix all instances of this in Bottlerocket?

webern: I think changing it in this PR would be nice, and I would suggest we follow it for future binaries. I don't know that retroactively changing all the other binaries is necessary, though I would recommend it for some of the more complicated arg-parsing functions.

My implementation above isn't precisely correct, it requires some finessing to get the message right, but the basic idea is that only main should exit the program. Thanks!

Makefile.toml
srgothi92: Sorry for jumping in between the conversation.
```
cd sources
export VARIANT=aws-ecs-1 && cargo build
```
@webern  Why won't this work ? This conditional is not inside `sources` dir packages right ? Not able to figure out how it affects.

webern: I'm assuming that the reason @etungsten added this to the makefile is because the `pluto` and `static-pods` packages will fail to compile unless the `model` package is set to use `k8s` settings. If we try to compile the workspace for ECS, the `model` will have the wrongs structs and our build will fail.

etungsten: Specifically `tasks.unit-tests` will fail because it just compiles everything in the workspace with the same VARIANT env var.

Regarding the `cd sources; export VARIANT=aws-ecs-1 && cargo build` use case. I feel like that's somewhat unusual. What reasons are there for people wanting to build the entire workspace without going through the Makefile? I would understand if people wanted to build specific workspace components for testing or w.e. But if they want to build everything in the workspace for use, shouldn't they just go through building the OS package via the Makefile?

etungsten:
Push above addresses @webern 's comment https://github.com/bottlerocket-os/bottlerocket/pull/1317#discussion_r577095160

Revamps arg parsing error handling. Still works as expected:

<details>

```
etung in api/static-pods/debug on 🌱 static-pods [$!?] 
$ ./static-pods --log-level asdasd
Invalid log level 'asdasd'
Usage: ./static-pods
            [ --socket-path PATH ]
            [ --log-level trace|debug|info|warn|error ]

    Socket path defaults to /run/api.sock
etung in api/static-pods/debug on 🌱 static-pods [$!?] 
$ ./static-pods --socket-path
Did not give argument to --socket-path
Usage: ./static-pods
            [ --socket-path PATH ]
            [ --log-level trace|debug|info|warn|error ]

    Socket path defaults to /run/api.sock
etung in api/static-pods/debug on 🌱 static-pods [$!?] 
$ ./static-pods --log-level
Did not give argument to --log-level
Usage: ./static-pods
            [ --socket-path PATH ]
            [ --log-level trace|debug|info|warn|error ]

    Socket path defaults to /run/api.sock

```

</details>

sources/api/static-pods/src/main.rs
webern: Thanks. LGTM.

sources/api/static-pods/src/main.rs
samuelkarp: I think I'm not understanding still.  We use `schnauzer::get_settings` to get [all the ECS settings](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/ecs-settings-applier/src/ecs.rs#L68-L76) in `ecs-settings-applier`.

etungsten: Hmm, I was mistaken in thinking the fields were not public. I'll go ahead and fix it.

sources/api/static-pods/src/main.rs
```diff
@@ -0,0 +1,338 @@
...
+        .context(error::APIRequest { method, uri })?;
+    ensure!(
+        code.is_success(),
+        error::APIResponse {
+            method,
+            uri,
+            code,
+            response_body,
+        }
+    );
```
arnaldo2792: Do you actually need this? Both [`apiclient::raw_request`](https://github.com/bottlerocket-os/bottlerocket/blob/8ea251400615cd275d830b690a445414451516d0/sources/api/apiclient/src/lib.rs#L85) and `schnauzer::get_settings` (through [`schnauzer::get_json`](https://github.com/bottlerocket-os/bottlerocket/blob/8ea251400615cd275d830b690a445414451516d0/sources/api/schnauzer/src/lib.rs#L59:14)) have this check already

etungsten:
Push above addresses @samuelkarp  @arnaldo2792's comment.
Use schnauzer instead of raw api requests.
Cleans up a bunch of unused errors, oops

Things still work as expected.

etungsten:
Push above sheds a few unnecessary imports.

samuelkarp (approved):

variants/aws-k8s-1.15/Cargo.toml
bcressey: Or a `k8s-release` - but `pluto` is AWS-specific where `static-pods` is not.

Makefile.toml
bcressey: I'm OK with your original conditional compilation for static-pods if it makes the overall workspace easier to build. I like the property that these crates can be built without bringing the entire build system to bear.

I'd prefer to get to the point where the entire sources tree can be built for each model simultaneously, and although the `Makefile.toml` workaround doesn't preclude that, it may invite us to continue segmenting the tree in ad-hoc ways rather than driving toward that goal.

webern: I use the workspace without bringing the build system to bear all the time. I'm not sure how to describe the use case, but having a workspace that can't be compiled with `cargo build` or `cargo test` seems to defeat the purpose of a workspace. Right now the only "gotchya" is that the `VARIANT` env is needed, and the error messages document this very well (I often forget, read the error then export a VARIANT).

I think the options are:

1) Merge this as is with the  `Makefile.toml` changes, and no longer can we `cargo build` the workspace (without adding the `--exclude` statements depending on `VARIANT`), or
2) Change to the `cfg` solution that we used with `ecs-settings-applier` so that we can always build all of the workspace, but the incompatible code is 'commented out'.

@bcressey are you saying that you choose `#1` and that, because the workspace no longer builds, we may be more motivated to fix it?

etungsten: I'll revert back to conditionally compiling static-pods since that's what people seem to prefer so far.

etungsten:
Push above reverts back to conditionally compiling static-pods based on the VARIANT env var. Drops changes to tasks.unit-tests.

webern (approved):
I'm happier with the cfg approach than the Makefile approach for conditional compilation, but I'll survive either way :)

arnaldo2792 (approved):

Makefile.toml
bcressey: >@bcressey are you saying that you choose `#1` and that, because the workspace no longer builds, we may be more motivated to fix it?

No, I was agreeing with you and I prefer `#2`.

bcressey (approved):

PR #1318, opened 2021-02-11
modify default label for containers

bcressey:
**Issue number:**
N/A

**Description of changes:**
The goal here is to better align our SELinux policy transitions with the behavior of Docker and containerd's CRI plugin, which expect that privileged containers will take the default path and end up with a more powerful label, while unprivileged ones will allocate an MCS pair that facilitates isolation between containers.

Although we don't yet support MCS isolation, this lays the groundwork for that by introducing a distinct label for privileged containers. That will let us add more restrictions to the `container_t` type, such as MCS constraints, without simultaneously restricting the `control_t` type in ways that are incompatible with expectations in the ecosystem.

Note that `control_t` and privileged are already quite similar in terms of their ability to affect the system: `control_t` offers highly privileged access to the host by means of the API, while privileged containers can bypass many access controls without using the API. 

Instead of repurposing the `control_t` type and granting the additional permissions to privileged containers, we could define a new type, such as `privileged_t`. Maintaining a separate but similar label would require duplicating many additional rules over time; for example, both labels would need access to host sockets, and both would need to be able to bypass MCS constraints.

**Testing done:**
I verified that Docker now uses `control_t` for "privileged" and "label disabled" containers.

As part of the work to update to containerd 1.4, I also verified that `control_t` is used for privileged containers launched via CRI. Until then, `container_t` will continue to be used because of our SELinux patches on containerd 1.3.

**Terms of contribution:**

### DIFF
--- a/packages/selinux-policy/lxc_contexts
+++ b/packages/selinux-policy/lxc_contexts
@@ -1,3 +1,12 @@
+# Runtimes that use the Go SELinux implementation, such as Docker and
+# the containerd CRI plugin, will apply the 'process' label to the
+# initial process for unprivileged containers, unless the option for
+# automatic labeling is disabled.
 process = "system_u:system_r:container_t:s0"
+
+# The 'file' label should always be applied to the container's root
+# filesystem, regardless of privileged status or automatic labeling.
 file = "system_u:object_r:local_t:s0"
+
+# The 'ro_file' label is not currently used by the above runtimes.
 ro_file = "system_u:object_r:cache_t:s0"
--- a/packages/selinux-policy/rules.cil
+++ b/packages/selinux-policy/rules.cil
@@ -53,13 +53,16 @@
 (allow init_t runtime_t (processes (transform)))
 (allow runtime_t runtime_exec_t (file (entrypoint)))
 
-; `runc` starts container processes as "container_t" by default, but
-; it can use other "container" subject labels like "super_t".
-; Unlike the above transitions, this depends on correct labeling for
-; objects on local storage.
-(typetransition runtime_t local_t process container_t)
-(typetransition runtime_t cache_t process container_t)
-(typetransition runtime_t state_t process container_t)
+; `runc` starts container processes as "control_t" by default, but it
+; can use other "container" subject labels like "container_t". This
+; depends on correct labeling for objects on local storage.
+;
+; Runtimes that use the Go SELinux library will override this label
+; with the "process" label from the `lxc_contexts` when launching
+; unprivileged containers, unless automatic labeling is disabled.
+(typetransition runtime_t local_t process control_t)
+(typetransition runtime_t cache_t process control_t)
+(typetransition runtime_t state_t process control_t)
 (allow runtime_t container_s (processes (transform)))
 (allow container_s local_t (file (entrypoint)))
 (allow container_s cache_t (file (entrypoint)))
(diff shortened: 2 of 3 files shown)


packages/selinux-policy/rules.cil
```diff
@@ -53,13 +53,13 @@
 (allow init_t runtime_t (processes (transform)))
 (allow runtime_t runtime_exec_t (file (entrypoint)))
 
-; `runc` starts container processes as "container_t" by default, but
-; it can use other "container" subject labels like "super_t".
+; `runc` starts container processes as "control_t" by default, but
+; it can use other "container" subject labels like "container_t".
```
tjkirch: I think it could be worth expanding this a little bit, because (naively read) it seems to imply that privileged is the default, since here it says control_t is default, and the docs above say "control_t (the default for privileged containers)".  This says "can use other [..] labels" but (again, naively) it's not clear when or why it would.  Offline you mentioned a process that automatically applies process labels, including making the default label container_t, down from control_t -- could there be a few words about that here, perhaps?  I'm totally willing to accept that this is my naivete on SELinux and that it's clear enough if you have context.

tjkirch (approved):

jhaynes (approved):
I'll echo @tjkirch's comment on expanding the explanation of the expected behavior.

bcressey:
^^ rebase for overall testing with containerd 1.4 update

bcressey:
The two force pushes together are what I intended to be a rebase with no changes. I'll push the actual changes for this PR shortly.

packages/selinux-policy/lxc_contexts
```diff
@@ -1,3 +1,13 @@
+# Runtimes that use the Go SELinux implementation, such as Docker and
+# the containerd CRI plugin, will apply the 'process' label to the
+# initial process for unprivileged containers, unless the option for
+# automatic labeling is disabled.
 process = "system_u:system_r:container_t:s0"
+
+# The 'file' label should always be applied to the container's root
+# filesystem, whether or not the container is privileged or automatic
+# labeling is disabled.
```
webern: Grammatically unclear?

Does it mean?

```
# The 'file' label should always be applied to the container's root
# filesystem, whether or not the container is privileged, or else
# automatic labeling will be disabled.
```

Could be my lack of context...

webern (approved):

packages/selinux-policy/lxc_contexts
samuelkarp: "The file label should always be applied to the container's root, whether or not the container is privileged, unless automatic labeling is disabled."

?

bcressey: Better like this?

```
# The 'file' label should always be applied to the container's root
# filesystem, regardless of whether the container is privileged or
# automatic labeling is disabled.
```

Or this?
```
# The 'file' label should always be applied to the container's root
# filesystem, even if the container is privileged and even when
# automatic labeling is disabled.
```

webern: I think A is good.

tjkirch: Or this?  :bike: :house:
```
# The 'file' label should always be applied to the container's root
# filesystem, regardless of privileged status or automatic labeling.
```

samuelkarp (approved):

PR #1282, opened 2021-01-17
Switched from TermLogger to SimpleLogger

hencrice:
**Issue number:**
Resolves #1157.

**Description of changes:**
Refactored the code to switch from TermLogger to SimpleLogger

**Testing done:**
`cargo test` with `VARIANT=aws-k8s-1.19`

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #1331, opened 2021-02-18
Add "shibaken" generator to populate user-data with public keys from IMDS

jpculp:
**Issue number:**
N/A

**Description of changes:**

Add "shibaken" generator to populate user-data with public keys from IMDS

This populates the host container's user-data setting with public keys
available from IMDS in the event that user-data has not been populated
by the user.

Adds a new migration helper `AddMetadataMigration` to remove setting
metadata (like setting-generators).

Adds a new migration for shibaken setting-generator.

**Testing done:**

Build aws-ecs-1 image and launched instance.

Verified that `host-containers.admin.user-data` contained a base64-encoded block.

Enabled and launched admin container, which started fine.

Verified that `/.bottlerocket/host-containers/admin/user-data` contained JSON with the following structure:

```
{
   "ssh":{
      "authorized_keys":[
         "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQClKsfkNkuSevGj3eYhCe53pcjqP3maAhDFcvBS7O6V
          hz2ItxCih+PnDSUaw+WNQn/mZphTk/a/gU8jEzoOWbkM4yxyb/wB96xbiFveSFJuOp/d6RJhJOI0iBXr
          lsLnBItntckiJ7FbtxJMXLvvwJryDUilBMTjYtwB+QhYXUMOzce5Pjz5/i8SeJtjnV3iAoG/cQk+0FzZ
          qaeJAAHco+CY/5WrUBkrHmFJr6HcXkvJdWPkYQS3xqC0+FmUZofz221CBt5IMucxXPkX4rWi+z7wB3Rb
          BQoQzd8v7yeb7OzlPnWOyN0qFU0XA246RA8QFYiCNYwI3f05p6KLxEXAMPLE my-key-pair"
      ]
   }
}
```

Ran `sudo sheltie` to verify root shell was still available.

@etungsten verified the migration between (1.0.5, 1.0.6)

**Terms of contribution:**

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -14,13 +14,14 @@
   Used for system maintenance and connectivity.
 * [**host-ctr**](sources/host-ctr): The program started by `host-containers@.service` for each host container.
   Its job is to start the specified host container on the “host” instance of containerd, which is separate from the “user” instance of containerd used for Kubernetes pods.
-* [**logdog**](sources/logdog): A program that one can use to collect logs when things go wrong. 
+* [**logdog**](sources/logdog): A program that one can use to collect logs when things go wrong.
 * [**metricdog**](sources/metricdog): A program that sends anonymous health pings.
 * [**model**](sources/models): The API system has a data model defined for each variant, and this model is used by other programs to serialize and deserialize requests while maintaining safety around data types.
 * [**netdog**](sources/api/netdog): A program called by wicked to retrieve and write out network configuration from DHCP.
 * [**pluto**](sources/api/pluto): A setting generator called by sundog to find networking settings required by Kubernetes.
 * [**schnauzer**](sources/api/schnauzer): A setting generator called by sundog to build setting values that contain template variables referencing other settings.
 * **setting generator**: A binary that generates the default value of a setting.
+* [**shibaken**](sources/api/shibaken): A setting generator called by sundog to populate the admin container's user-data with public keys from IMDS, when running in AWS.
 * [**signpost**](sources/updater/signpost): A program used to manipulate the GPT header of the OS disk; fields in the header are used by GRUB to determine the partition set we should boot from.
 * [**storewolf**](sources/api/storewolf): A program that sets up the data store for the API upon boot.
 * [**sundog**](sources/api/sundog): A program run during boot that generates any settings that depend on runtime system information.
--- a/Release.toml
+++ b/Release.toml
@@ -18,13 +18,13 @@ version = "1.0.5"
     "migrate_v1.0.5_add-user-data.lz4",
     "migrate_v1.0.5_add-network-settings.lz4",
     "migrate_v1.0.5_add-proxy-restart.lz4",
-    "migrate_v1.0.5_add-proxy-services.lz4"
+    "migrate_v1.0.5_add-proxy-services.lz4",
 ]
 "(1.0.5, 1.0.6)" = [
     "migrate_v1.0.6_metricdog-init.lz4",
     "migrate_v1.0.6_add-static-pods.lz4",
     "migrate_v1.0.6_kubelet-standalone-tls-settings.lz4",
     "migrate_v1.0.6_kubelet-standalone-tls-services.lz4",
-    "migrate_v1.0.6_control-container-v0-4-2.lz4"
+    "migrate_v1.0.6_control-container-v0-4-2.lz4",
+    "migrate_v1.0.6_add-shibaken.lz4",
 ]
-
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -78,6 +78,7 @@ Requires: %{_cross_os}schnauzer = %{version}-%{release}
 Requires: %{_cross_os}pluto = %{version}-%{release}
 %endif
 Requires: %{_cross_os}bork = %{version}-%{release}
+Requires: %{_cross_os}shibaken = %{version}-%{release}
 %description -n %{_cross_os}sundog
 %{summary}.
 
@@ -95,6 +96,11 @@ Requires: %{_cross_os}apiserver = %{version}-%{release}
 %package -n %{_cross_os}schnauzer
 Summary: Setting generator for templated settings values.
 %description -n %{_cross_os}schnauzer
+%{summary}.
+
+%package -n %{_cross_os}shibaken
+Summary: Setting generator for populating admin container user-data from IMDS.
+%description -n %{_cross_os}shibaken
 %{summary}.
 
 %package -n %{_cross_os}thar-be-settings
@@ -206,6 +212,7 @@ mkdir bin
     -p sundog \
     -p schnauzer \
     -p bork \
+    -p shibaken \
     -p thar-be-settings \
     -p thar-be-updates \
     -p servicedog \
@@ -248,7 +255,7 @@ done
 install -d %{buildroot}%{_cross_bindir}
 for p in \
   apiserver \
-  early-boot-config netdog sundog schnauzer bork corndog \
+  early-boot-config netdog sundog schnauzer bork shibaken corndog \
   thar-be-settings thar-be-updates servicedog host-containers \
   storewolf settings-committer \
   migrator \
@@ -353,6 +360,9 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 
 %files -n %{_cross_os}bork
 %{_cross_bindir}/bork
+
+%files -n %{_cross_os}shibaken
+%{_cross_bindir}/shibaken
 
 %files -n %{_cross_os}thar-be-settings
 %{_cross_bindir}/thar-be-settings
--- a/sources/api/migration/migration-helpers/src/common_migrations.rs
+++ b/sources/api/migration/migration-helpers/src/common_migrations.rs
@@ -739,3 +739,106 @@ impl Migration for ReplaceTemplateMigration {
         Ok(input)
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// We use this migration when we add metadata and want to make sure they're removed before we go
+/// back to old versions that don't understand them.
+#[derive(Debug)]
+pub struct SettingMetadata {
+    pub setting: &'static str,
+    pub metadata: &'static [&'static str],
+}
+
+pub struct AddMetadataMigration(pub &'static [SettingMetadata]);
+
+impl Migration for AddMetadataMigration {
+    /// New versions must have the metadata already defined in defaults.
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        println!(
+            "AddMetadataMigration({:?}) has no work to do on upgrade.",
+            &self.0
+        );
+        Ok(input)
+    }
+
+    /// Older versions might break with certain settings metadata (such as with setting-generators)
+    /// so we need to remove them.
+    fn backward(&mut self, mut input: MigrationData) -> Result<MigrationData> {
+        for setting_metadata in self.0 {
+            if let Some(found_metadata) = input.metadata.get_mut(setting_metadata.setting) {
+                for metadata in setting_metadata.metadata {
+                    if let Some(metadata_value) = found_metadata.remove(*metadata) {
+                        println!(
+                            "Removed {}, which was set to '{}'",
+                            metadata, metadata_value
+                        );
+                    } else {
+                        println!(
+                            "Found no metadata '{}' to remove on setting '{}'",
+                            metadata, setting_metadata.setting
+                        );
+                    }
+                }
+            } else {
+                println!(
+                    "Found no metadata for '{}' setting",
+                    setting_metadata.setting
+                );
+            }
+        }
+        Ok(input)
+    }
+}
+
+#[cfg(test)]
+mod test_add_metadata {
+    use super::{AddMetadataMigration, SettingMetadata};
+    use crate::{Migration, MigrationData};
+    use maplit::hashmap;
+    use std::collections::HashMap;
+
+    #[test]
+    fn backward() {
+        let data = MigrationData {
+            data: HashMap::new(),
+            metadata: hashmap! {
+                "hi".into() => hashmap!{"there".into() => "whatever".into()},
+            },
+        };
+        let result = AddMetadataMigration(&[SettingMetadata {
+            setting: "hi",
+            metadata: &["there"],
+        }])
+        .backward(data)
+        .unwrap();
+        assert_eq!(
+            result.metadata,
+            hashmap! {
+                "hi".into() => HashMap::new().into(),
+            }
+        );
+    }
+
+    #[test]
+    fn backward_noop() {
+        let data = MigrationData {
+            data: HashMap::new(),
+            metadata: hashmap! {
+                "hi".into() => hashmap!{"sup".into() => "wassup".into()},
+            },
+        };
+        let result = AddMetadataMigration(&[SettingMetadata {
+            setting: "hi",
+            metadata: &["there"],
+        }])
+        .backward(data)
+        .unwrap();
+        assert_eq!(
+            result.metadata,
+            hashmap! {
+                "hi".into() => hashmap!{"sup".into() => "wassup".into()},
+            }
+        );
+    }
+}
--- a/sources/api/migration/migrations/v1.0.6/add-shibaken/src/main.rs
+++ b/sources/api/migration/migrations/v1.0.6/add-shibaken/src/main.rs
@@ -1,0 +1,23 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::{AddMetadataMigration, SettingMetadata};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting metadata, `host-containers.admin.user-data.setting-generator`
+fn run() -> Result<()> {
+    migrate(AddMetadataMigration(&[SettingMetadata {
+        metadata: &["setting-generator"],
+        setting: "settings.host-containers.admin.user-data",
+    }]))
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/api/shibaken/src/main.rs
+++ b/sources/api/shibaken/src/main.rs
@@ -1,0 +1,285 @@
+/*!
+# Introduction
+
+shibaken is called by sundog as a setting generator.
+
+shibaken will fetch and populate the admin container's user-data with authorized ssh keys from the
+AWS instance metadata service (IMDS).
+
+(The name "shibaken" comes from the fact that Shiba are small, but agile, hunting dogs.)
+*/
+
+#![deny(rust_2018_idioms)]
+
+use log::{debug, info, warn};
+use reqwest::blocking::Client;
+use serde::Serialize;
+use simplelog::{Config as LogConfig, LevelFilter, TermLogger, TerminalMode};
+use snafu::{OptionExt, ResultExt};
+use std::str::FromStr;
+use std::{env, process};
+
+// Instance Meta Data Service.
+//
+// Currently only able to get fetch session tokens from `latest`
+// FIXME Pin to a date version that supports IMDSv2 once such a date version is available.
+const IMDS_PUBLIC_KEY_BASE_URI: &str = "http://169.254.169.254/latest/meta-data/public-keys";
+const IMDS_SESSION_TOKEN_URI: &str = "http://169.254.169.254/latest/api/token";
+
+#[derive(Serialize)]
+struct UserData {
+    ssh: Ssh,
+}
+
+#[derive(Serialize)]
+struct Ssh {
+    authorized_keys: Vec<String>,
+}
+impl UserData {
+    fn new(public_keys: Vec<String>) -> Self {
+        UserData {
+            ssh: Ssh {
+                authorized_keys: public_keys,
+            },
+        }
+    }
+}
+
+/// Helper to fetch an IMDSv2 session token that is valid for 60 seconds.
+fn fetch_imds_session_token(client: &Client) -> Result<String> {
+    let uri = IMDS_SESSION_TOKEN_URI;
+    let imds_session_token = client
+        .put(uri)
+        .header("X-aws-ec2-metadata-token-ttl-seconds", "60")
+        .send()
+        .context(error::ImdsRequest { method: "PUT", uri })?
+        .error_for_status()
+        .context(error::ImdsResponse { uri })?
+        .text()
+        .context(error::ImdsText { uri })?;
+    Ok(imds_session_token)
+}
+
+/// Helper to fetch data from IMDS. Taken from pluto.
+fn get_text_from_imds(client: &Client, uri: &str, session_token: &str) -> Result<String> {
+    client
+        .get(uri)
+        .header("X-aws-ec2-metadata-token", session_token)
+        .send()
+        .context(error::ImdsRequest { method: "GET", uri })?
+        .error_for_status()
+        .context(error::ImdsResponse { uri })?
+        .text()
+        .context(error::ImdsText { uri })
+}
+
+/// Returns a list of public keys.
+fn fetch_public_keys_from_imds() -> Result<Vec<String>> {
+    info!("Fetching IMDS session token");
+    let client = Client::new();
+    let imds_session_token = fetch_imds_session_token(&client)?;
+
+    info!("Fetching list of available public keys from IMDS");
+    // Returns a list of available public keys as '0=my-public-key'
+    let public_key_list =
+        get_text_from_imds(&client, IMDS_PUBLIC_KEY_BASE_URI, &imds_session_token)?;
+    debug!("available public keys '{}'", &public_key_list);
+
+    info!("Generating uris to fetch text of available public keys");
+    let public_key_uris = build_public_key_uris(&public_key_list);
+
+    info!("Fetching public keys from IMDS");
+    let mut public_keys = Vec::new();
+    for uri in public_key_uris {
+        let public_key_text = get_text_from_imds(&client, &uri, &imds_session_token)?;
+        let public_key = public_key_text.trim_end();
+        // Simple check to see if the text is probably an ssh key.
+        if public_key.starts_with("ssh") {
+            debug!("{}", &public_key);
+            public_keys.push(public_key.to_string())
+        } else {
+            warn!(
+                "'{}' does not appear to be a valid key. Skipping...",
+                &public_key_text
+            );
+            continue;
+        }
+    }
+    if public_keys.is_empty() {
+        warn!("No valid keys found");
+    }
+    Ok(public_keys)
+}
+
+/// Returns a list of public key uris strings for the public keys in IMDS. Since IMDS returns the
+/// list of available public keys as '0=my-public-key', we need to strip the index from the list and
+/// insert it into the key uri.
+fn build_public_key_uris(public_key_list: &str) -> Vec<String> {
+    let mut public_key_uris = Vec::new();
+    for available_key in public_key_list.lines() {
+        let f: Vec<&str> = available_key.split('=').collect();
+        // If f[0] isn't a number, then it isn't a valid index.
+        if f[0].parse::<u32>().is_ok() {
+            let public_key_uri = format!("{}/{}/openssh-key", IMDS_PUBLIC_KEY_BASE_URI, f[0]);
+            public_key_uris.push(public_key_uri);
+        } else {
+            warn!(
+                "'{}' does not appear to be a valid index. Skipping...",
+                &f[0]
+            );
+            continue;
+        }
+    }
+    if public_key_uris.is_empty() {
+        warn!("No valid key uris found");
+    }
+    public_key_uris
+}
+
+/// Store the args we receive on the command line.
+struct Args {
+    log_level: LevelFilter,
+}
+
+/// Print a usage message in the event a bad arg is passed
+fn usage() {
+    let program_name = env::args().next().unwrap_or_else(|| "program".to_string());
+    eprintln!(
+        r"Usage: {}
+            [ --log-level trace|debug|info|warn|error ]",
+        program_name
+    );
+}
+
+/// Parse the args to the program and return an Args struct
+fn parse_args(args: env::Args) -> Result<Args> {
+    let mut log_level = None;
+
+    let mut iter = args.skip(1);
+    while let Some(arg) = iter.next() {
+        match arg.as_ref() {
+            "--log-level" => {
+                let log_level_str = iter.next().context(error::Usage {
+                    message: "Did not give argument to --log-level",
+                })?;
+                log_level = Some(
+                    LevelFilter::from_str(&log_level_str)
+                        .context(error::LogLevel { log_level_str })?,
+                );
+            }
+
+            x => {
+                return error::Usage {
+                    message: format!("unexpected argument '{}'", x),
+                }
+                .fail()
+            }
+        }
+    }
+
+    Ok(Args {
+        log_level: log_level.unwrap_or_else(|| LevelFilter::Info),
+    })
+}
+
+fn run() -> Result<()> {
+    let args = parse_args(env::args())?;
+
+    // TerminalMode::Stderr will send all logs to stderr, as sundog only expects the json output of
+    // the setting on stdout.
+    TermLogger::init(args.log_level, LogConfig::default(), TerminalMode::Stderr)
+        .context(error::Logger)?;
+
+    info!("shibaken started");
+
+    let public_keys = fetch_public_keys_from_imds()?;
+
+    let user_data = UserData::new(public_keys);
+
+    info!("Generating user-data");
+    // Serialize user_data to a JSON string that can be read by the admin container.
+    let user_data_json = serde_json::to_string(&user_data).context(error::SerializeJson)?;
+    debug!("{}", &user_data_json);
+
+    info!("Encoding user-data");
+    // admin container user-data must be base64-encoded to be passed through to the admin container
+    // using a setting, rather than another arbitrary storage mechanism. This approach allows the
+    // user to bypass shibaken and use their own user-data if desired.
+    let user_data_base64 = base64::encode(&user_data_json);
+
+    info!("Outputting user-data");
+    // sundog expects JSON-serialized output so that many types can be represented, allowing the
+    // API model to use more accurate types.
+    let output = serde_json::to_string(&user_data_base64).context(error::SerializeJson)?;
+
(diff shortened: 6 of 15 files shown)


packages/os/os.spec
```diff
@@ -97,6 +98,11 @@ Summary: Setting generator for templated settings values.
 %description -n %{_cross_os}schnauzer
 %{summary}.
 
+%package -n %{_cross_os}shibaken
+Summary: Setting generator for IMDS keys to host-containers.admin.user-data.
```
bcressey: nit: it's somewhat inaccurate to mention the specific setting here; the model could default to using the generator on a different key.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+            public_keys.push(public_key_text.trim_end().to_string())
+        } else {
+            warn!(
+                "'{}' does not appear to be a valid key. Skipping...",
+                &public_key_text.trim_end()
+            );
+            continue;
+        }
+    }
+    ensure!(!public_keys.is_empty(), error::NoValidImdsKeys);
```
bcressey: It's worth handling the case of an instance with no key pair, to simplify the implementation for the admin container.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+    let public_key_endpoints = build_public_key_endpoints_from_imds(&public_key_list)?;
+
+    info!("Fetching public keys from IMDS");
+    let mut public_keys = Vec::new();
+    for endpoint in public_key_endpoints.iter() {
+        let public_key_text = fetch_data_from_imds(&client, endpoint, &imds_session_token)?;
+        // Simple check to see if the text is probably an ssh key.
+        if &public_key_text[0..3] == "ssh" {
+            debug!("{}", &public_key_text.trim_end());
+            public_keys.push(public_key_text.trim_end().to_string())
```
bcressey: nit: we should store `public_key_text.trim_end()` in a variable so we don't repeat the calculation so many times.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+use simplelog::{Config as LogConfig, LevelFilter, TermLogger, TerminalMode};
+use snafu::{ensure, ResultExt};
+use std::str::FromStr;
+use std::{env, process};
+
+// Instance Meta Data Service.
+//
+// Currently only able to get fetch session tokens from `latest`
+// FIXME Pin to a date version that supports IMDSv2 once such a date version is available.
+const IMDS_PUBLIC_KEY_BASE_URL: &str = "http://169.254.169.254/latest/meta-data/public-keys/";
```
bcressey: nit: can omit the trailing slash here and append it in context later so that the format string is more readable

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+
+/// Returns a list of endpoint urls strings for the public keys in IMDS. Since IMDS returns the list
+/// of available public keys as '0=my-public-key', we need to strip the index from the list and
+/// embed it into the endpoint url.
+fn build_public_key_endpoints_from_imds(public_key_list: &str) -> Result<Vec<String>> {
+    let mut public_key_endpoints = Vec::new();
+    for available_key in public_key_list.lines() {
+        let f: Vec<&str> = available_key.split('=').collect();
+        // If f[0] isn't a number, then it isn't a valid index.
+        if f[0].parse::<i32>().is_ok() {
```
bcressey: nit: `u32` instead? Doesn't seem like we'd get negative key indices.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+    }
+    ensure!(!public_keys.is_empty(), error::NoValidImdsKeys);
+
+    Ok(public_keys)
+}
+
+/// Returns a list of endpoint urls strings for the public keys in IMDS. Since IMDS returns the list
+/// of available public keys as '0=my-public-key', we need to strip the index from the list and
+/// embed it into the endpoint url.
+fn build_public_key_endpoints_from_imds(public_key_list: &str) -> Result<Vec<String>> {
```
bcressey: nit: we're not talking to IMDS in this function, so perhaps just
```suggestion
fn build_public_key_endpoints(public_key_list: &str) -> Result<Vec<String>> {
```

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+    info!("Fetching IMDS session token");
+    let client = Client::new();
+    let imds_session_token = fetch_imds_session_token(&client)?;
+
+    info!("Fetching list of available public keys from IMDS");
+    // Returns a list of available public keys as '0=my-public-key'
+    let public_key_list =
+        fetch_data_from_imds(&client, IMDS_PUBLIC_KEY_BASE_URL, &imds_session_token)?;
+    debug!("available public keys '{}'", &public_key_list);
+
```
bcressey: My misgiving here is that we're essentially parsing an index page. It might be cleaner and simpler just to start from `0` and retrieve the data until we get a 404.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+fn usage() -> ! {
+    let program_name = env::args().next().unwrap_or_else(|| "program".to_string());
+    eprintln!(
+        r"Usage: {}
+            [ --log-level trace|debug|info|warn|error ]",
+        program_name
+    );
+    process::exit(2);
+}
+
```
bcressey: @webern wants this handled differently to simplify testing.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+        log_level: log_level.unwrap_or_else(|| LevelFilter::Error),
+    }
+}
+
+fn run() -> Result<()> {
+    let args = parse_args(env::args());
+
+    // TerminalMode::Mixed will send errors to stderr and anything less to stdout.
+    TermLogger::init(args.log_level, LogConfig::default(), TerminalMode::Mixed)
+        .context(error::Logger)?;
```
bcressey: This might need to be aligned with the SimpleLogger convention from #1282.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    info!("shibaken started");
```
bcressey, line 193: Are these `info!` uses correct? It seems odd to mix `info!` and `println!` if sundog is expecting structured data on stdout.

jpculp: It's a tad tricky. shibaken is called by sundog which is expecting a JSON answer on **stdout**. It seems like it would be nice if this dog had a logging mechanism, but in `TerminalMode::Mixed` anything less than error would go to **stdout**. One alternative would be to set `TerminalMode::Stderr` which would send all logs to **stderr**, but it feels a little wrong. In this case, the default log level is set to Error, but then there isn't much point to having the info and warn logs. Would it be better to remove the info logs all together and send warn/error to **stderr** or keep the logging as is and send it all to **stderr**?

sources/api/shibaken/src/main.rs
jpculp: I'll go ahead and mirror the logic in https://github.com/bottlerocket-os/bottlerocket/pull/1317

sources/api/shibaken/src/main.rs
jpculp: That is certainly a valid approach. I would rather keep the current logic as it more closely reflects the logic in the [admin container's entrypoint](https://github.com/bottlerocket-os/bottlerocket-admin-container/blob/eec7430012182bb77a68a327f10065abdcc73a3f/start_admin_sshd.sh#L21) and the [examples for retrieving public keys](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#instance-metadata-ex-3) in the AWS documentation.

sources/api/shibaken/src/main.rs
jpculp: After some testing, it turns out that we won't be able to use SimpleLogger as it logs all messages to stdout. As detailed in another conversation, since sundog is only expecting the JSON setting on stdout, we should send everything to stderr or use another mechanism for logging. For now, I'll set the logger to TermLogger with a TerminalMode of Stderr.

GLOSSARY.md
```diff
@@ -14,13 +14,14 @@
...
-* [**logdog**](sources/logdog): A program that one can use to collect logs when things go wrong. 
+* [**logdog**](sources/logdog): A program that one can use to collect logs when things go wrong.
 * [**metricdog**](sources/metricdog): A program that sends anonymous health pings.
 * [**model**](sources/models): The API system has a data model defined for each variant, and this model is used by other programs to serialize and deserialize requests while maintaining safety around data types.
 * [**netdog**](sources/api/netdog): A program called by wicked to retrieve and write out network configuration from DHCP.
 * [**pluto**](sources/api/pluto): A setting generator called by sundog to find networking settings required by Kubernetes.
 * [**schnauzer**](sources/api/schnauzer): A setting generator called by sundog to build setting values that contain template variables referencing other settings.
 * **setting generator**: A binary that generates the default value of a setting.
+* [**shibaken**](sources/api/shibaken): A setting generator called by sundog to build user-data with public keys from IMDS.
```
zmrow: nit: `...to populate host container user-data...`

packages/os/os.spec
```diff
@@ -97,6 +98,11 @@ Summary: Setting generator for templated settings values.
 %description -n %{_cross_os}schnauzer
 %{summary}.
 
+%package -n %{_cross_os}shibaken
+Summary: Setting generator for user-data with IMDS public keys.
```
zmrow: nit: I might just leave this as:
`Setting generator for IMDS public key host-container user-data`

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -739,3 +739,43 @@ impl Migration for ReplaceTemplateMigration {
...
+        println!(
+            "AddMetadataMigration({},{:?}) has no work to do on upgrade.",
+            self.setting, self.metadata
+        );
+        Ok(input)
+    }
+
+    /// Older versions might break with certain settings metadata (such as with setting-generators)
+    /// so we need to remove them.
+    fn backward(&mut self, mut input: MigrationData) -> Result<MigrationData> {
```
zmrow, line 767: Paging @tjkirch for a second set of eyes on this new migration type.

sources/api/shibaken/src/main.rs
zmrow: Yeah, this is a bit of a new situation for us.  None of the other setting generators log anything precisely for this reason.

However, since this generator is mission-critical it seems like it should probably log what it's doing to make for easier troubleshooting in the event something bad happens.  It _is_ weird to log everything to stderr, but we don't have another option as far as I know.  We could remove all logging, but that seems like a poor option as well.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    let user_data = UserData::new(public_keys);
+
+    info!("Generating user-data");
+    // Serialize user_data to a JSON string that can be read by the admin container.
+    let user_data_json = serde_json::to_string(&user_data).context(error::SerializeJson)?;
+    debug!("{}", &user_data_json);
+
+    info!("Encoding user-data");
+    // Encode user_data as base64 to represent it in the base64 `user-data` setting, which allows
```
zmrow: nit: This comment could probably be reduced to something like `Host container user-data must be base64 encoded`

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        match e {
+            error::Error::Usage { .. } => {
+                eprintln!("{}", e);
+                usage();
+                process::exit(2);
```
zmrow: Sundog matches on the exit code of these generators, and "2" means it only logs that something bad happened.  Since this is a crucial piece of the operating system, I'd say we need to still need to return "1" here to make sure it's treated like a failure.

zmrow (commented):
Also - could we update the testing of this PR to reflect that the migration up/down was tested and the settings appeared as we expected them to and the admin host container works in both versions?

sources/api/shibaken/src/main.rs
tjkirch: >  None of the other setting generators log anything precisely for this reason

I think we just weren't thinking about it.  We should probably log to stderr.  All generators are mission critical!  :)

packages/os/os.spec
jahkeup: Or even more general: `Settings generator for populating host-container user-data vended by IMDS`

We'll expect to handle other data and the `Summary` tag isn't made visible. Making this general means we won't have to remember to update the description as features are added elsewhere in the tree (and the suggested wording is not inaccurate as-is).

sources/api/shibaken/src/main.rs
jahkeup: Agree with @zmrow here - this is a constraint from the model, not a choice of the generator.

sources/api/shibaken/src/main.rs
jahkeup: I'd want to see these lifted into a `generator_helper::ExitStatus` enum or something. Convention and discipline is on our side, but a shared definition makes this clearer and harder to diverge from.

packages/os/os.spec
jpculp: I'll change this to:
`Summary: Setting generator for populating host container user-data with IMDS.`

That way it will remain general and is also <80 chars.

sources/api/shibaken/src/main.rs
bcressey: Fair enough!

sources/api/shibaken/src/main.rs
zmrow: Let's open that as a separate issue - I'd rather not block on that here.

zmrow (approved):
Looks good!

:parachute:

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+    let imds_session_token = fetch_imds_session_token(&client)?;
+
+    info!("Fetching list of available public keys from IMDS");
+    // Returns a list of available public keys as '0=my-public-key'
+    let public_key_list =
+        fetch_data_from_imds(&client, IMDS_PUBLIC_KEY_BASE_URL, &imds_session_token)?;
+    debug!("available public keys '{}'", &public_key_list);
+
+    info!("Generating endpoint urls to fetch text of available public keys");
+    let public_key_endpoints = build_public_key_endpoints(&public_key_list);
```
bcressey: nit: `endpoint` seems like not quite the right name, since I think of an endpoint as a hostname + port.
```suggestion
    info!("Generating urls to fetch text of available public keys");
    let public_key_urls = build_public_key_urls(&public_key_list);
```

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -739,3 +739,43 @@ impl Migration for ReplaceTemplateMigration {
...
+
+impl Migration for AddMetadataMigration<'_> {
+    /// New versions must have the metadata already defined in defaults.
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        println!(
+            "AddMetadataMigration({},{:?}) has no work to do on upgrade.",
+            self.setting, self.metadata
+        );
+        Ok(input)
+    }
```
bcressey: This looks right to me, but it'd be good to add a unit test.

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -739,3 +739,43 @@ impl Migration for ReplaceTemplateMigration {
...
+                    println!(
+                        "Removed {}, which was set to '{}'",
+                        metadata, metadata_value
+                    );
+                } else {
+                    println!("Found no {} to remove", metadata);
+                }
+            }
+        } else {
+            println!("Found no '{}' setting", self.setting);
```
tjkirch:
```suggestion
            println!("Found no metadata for '{}' setting", self.setting);
```

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -739,3 +739,43 @@ impl Migration for ReplaceTemplateMigration {
...
+    fn backward(&mut self, mut input: MigrationData) -> Result<MigrationData> {
+        if let Some(found_metadata) = input.metadata.get_mut(self.setting) {
+            for metadata in self.metadata {
+                if let Some(metadata_value) = found_metadata.remove(*metadata) {
+                    println!(
+                        "Removed {}, which was set to '{}'",
+                        metadata, metadata_value
+                    );
+                } else {
+                    println!("Found no {} to remove", metadata);
```
tjkirch:
```suggestion
                    println!("Found no metadata '{}' to remove on setting '{}'", metadata, self.setting);
```

GLOSSARY.md
```diff
@@ -14,13 +14,14 @@
...
-* [**logdog**](sources/logdog): A program that one can use to collect logs when things go wrong. 
+* [**logdog**](sources/logdog): A program that one can use to collect logs when things go wrong.
 * [**metricdog**](sources/metricdog): A program that sends anonymous health pings.
 * [**model**](sources/models): The API system has a data model defined for each variant, and this model is used by other programs to serialize and deserialize requests while maintaining safety around data types.
 * [**netdog**](sources/api/netdog): A program called by wicked to retrieve and write out network configuration from DHCP.
 * [**pluto**](sources/api/pluto): A setting generator called by sundog to find networking settings required by Kubernetes.
 * [**schnauzer**](sources/api/schnauzer): A setting generator called by sundog to build setting values that contain template variables referencing other settings.
 * **setting generator**: A binary that generates the default value of a setting.
+* [**shibaken**](sources/api/shibaken): A setting generator called by sundog to populate host container user-data with public keys from IMDS.
```
tjkirch:
```suggestion
* [**shibaken**](sources/api/shibaken): A setting generator called by sundog to populate host container user-data with public keys from IMDS, when running in AWS.
```

sources/api/migration/migrations/v1.0.6/add-shibaken/src/main.rs
```diff
@@ -0,0 +1,23 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddMetadataMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting metadata, `host-containers.admin.user-data.setting-generator`
+fn run() -> Result<()> {
+    migrate(AddMetadataMigration {
```
tjkirch: We've learned in the past that we almost always want to handle multiple settings in the same migration, since our current migration system only lets you have one `Migration` per binary, and needing multiple binaries always feels a bit silly.  From that perspective, this could take a list of things, each of which has a setting and associated list of metadata.  Erikson did that for a similar case with the `ListReplacement` struct that you pass multiple times in the Migration constructor.  (I don't think you'd need to rename the migration, just based on the fact that `Metadatas` is awkward, and doesn't describe the change anyway.)

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+        UserData {
+            ssh: Ssh {
+                authorized_keys: public_keys,
+            },
+        }
+    }
+}
+
+/// Helper to fetch an IMDSv2 session token that is valid for 60 seconds.
+fn fetch_imds_session_token(client: &Client) -> Result<String> {
```
tjkirch, line 49: _tangent:_ It is a bit unfortunate to have three implementations of this now, each slightly different.  We should consider making an imds crate as a follow-up.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+shibaken will fetch and populate the host container's user-data with authorized ssh keys from the
+AWS instance metadata service (IMDS).
+
+(The name "shibaken" comes from the fact that Shiba are small, but agile, hunting dogs.)
+*/
+
+#![deny(rust_2018_idioms)]
+
+#[macro_use]
+extern crate log;
```
tjkirch: This is an older style - now you can just `use log::info;` and not need the extern syntax.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+    let public_key_list =
+        fetch_data_from_imds(&client, IMDS_PUBLIC_KEY_BASE_URL, &imds_session_token)?;
+    debug!("available public keys '{}'", &public_key_list);
+
+    info!("Generating endpoint urls to fetch text of available public keys");
+    let public_key_endpoints = build_public_key_endpoints(&public_key_list);
+
+    info!("Fetching public keys from IMDS");
+    let mut public_keys = Vec::new();
+    for endpoint in public_key_endpoints.iter() {
```
tjkirch:
```suggestion
    for endpoint in public_key_endpoints {
```
You should be able to use the default behavior that automatically calls`.into_iter()`, I think...

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+        let public_key_text = fetch_data_from_imds(&client, endpoint, &imds_session_token)?;
+        let public_key = public_key_text.trim_end();
+        // Simple check to see if the text is probably an ssh key.
+        if &public_key[0..3] == "ssh" {
```
tjkirch: This will panic if the text we got back has fewer than 3 characters.

```suggestion
        if public_key.starts_with("ssh") {
```

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+                log_level = Some(LevelFilter::from_str(&log_level_str).map_err(|_| {
+                    error::Error::Usage {
+                        message: format!("Invalid log level '{}'", log_level_str),
+                    }
+                })?);
+            }
+
+            _ => {
+                return Err(error::Error::Usage {
+                    message: "unexpected argument".into(),
```
tjkirch: This wouldn't say which argument was unexpected.  Can include that in the struct, if you use `x` instead of `_` to capture it, for example.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    let user_data = UserData::new(public_keys);
+
+    info!("Generating user-data");
+    // Serialize user_data to a JSON string that can be read by the admin container.
+    let user_data_json = serde_json::to_string(&user_data).context(error::SerializeJson)?;
+    debug!("{}", &user_data_json);
+
+    info!("Encoding user-data");
+    // Host container user-data must be base64-encoded.
```
tjkirch: nit: I think it'd be useful to say _why_ it needs to be base64-encoded... the reader can assume it needs to be based on the fact that we're doing it.  We encode it because we want to pass the data through to the container using a setting, rather than another arbitrary storage mechanism, so that the user has control to override it if desired.  And the field has to be base64 to allow that user-entered data to be arbitrary.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+
+mod error {
+    use snafu::Snafu;
+    fn code(source: &reqwest::Error) -> String {
+        source
+            .status()
+            .as_ref()
+            .map(|i| i.as_str())
+            .unwrap_or("Unknown")
+            .to_string()
```
tjkirch, line 249: Does this have to be converted to an owned String?  I would have thought snafu would do that for you.  If not, no problem!

sources/api/shibaken/src/main.rs
tjkirch: Can we say _why_ we want it handled differently or how it simplifies testing?  This new style using snafu and pushing the work into `main` seems more complicated.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,286 @@
+/*!
+# Introduction
+
+shibaken is called by sundog as a setting generator.
+
+shibaken will fetch and populate the host container's user-data with authorized ssh keys from the
```
samuelkarp: nit

```suggestion
shibaken will fetch and populate the admin container's user-data with authorized ssh keys from the
```

sources/api/shibaken/src/main.rs
jpculp: I definitely agree. I can take that on in another PR if folks are interested.

sources/api/shibaken/src/main.rs
jpculp: I'm not positive. This block came from pluto so I'm hesitant to change it. That's certainly an optimization we could look at in a follow-up imds crate.

sources/api/shibaken/src/main.rs
jpculp: @zmrow in several places you asked me you use host container's user-data previously. Should I go back and switch all of those to admin container's user-data? I'm kind of with Sam here in feeling that admin container is more correct.

sources/api/shibaken/src/main.rs
jpculp: My understanding is this makes it so only `main` can exit the program to simplify testing. I think it would make sense for us to discuss as a team what our standard approach should be as there are pros/cons to both styles.

sources/api/shibaken/src/main.rs
tjkirch: It's the kind of thing that wouldn't compile if it was wrong, so it's safe to test by removing the `to_string()` and changing the return type to `&str`.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+use snafu::ResultExt;
+use std::str::FromStr;
+use std::{env, process};
+
+// Instance Meta Data Service.
+//
+// Currently only able to get fetch session tokens from `latest`
+// FIXME Pin to a date version that supports IMDSv2 once such a date version is available.
+const IMDS_PUBLIC_KEY_BASE_URL: &str = "http://169.254.169.254/latest/meta-data/public-keys";
+const IMDS_SESSION_TOKEN_ENDPOINT: &str = "http://169.254.169.254/latest/api/token";
```
tjkirch:
```suggestion
const IMDS_SESSION_TOKEN_URI: &str = "http://169.254.169.254/latest/api/token";
```
To match changes below.

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+    let public_key_list =
+        get_text_from_imds(&client, IMDS_PUBLIC_KEY_BASE_URL, &imds_session_token)?;
+    debug!("available public keys '{}'", &public_key_list);
+
+    info!("Generating urls to fetch text of available public keys");
+    let public_key_urls = build_public_key_urls(&public_key_list);
+
+    info!("Fetching public keys from IMDS");
+    let mut public_keys = Vec::new();
+    for url in public_key_urls.into_iter() {
```
tjkirch:
```suggestion
    for url in public_key_urls {
```

You shouldn't have to say `.into_iter()`, it should be the default in a for loop.

jpculp:
For some context on the recent pushes that squashed together...

Rebased and addressed comments/concerns with the shibaken crate itself:
https://github.com/bottlerocket-os/bottlerocket/compare/e617e9ec..0d9fb467

Replaced instances of host container with admin container and enhanced the AddMetadataMigration to accept multiple settings (+ unit test):
https://github.com/bottlerocket-os/bottlerocket/compare/0d9fb467..30d61d14

Attempt to manually fix merge conflict:
https://github.com/bottlerocket-os/bottlerocket/compare/30d61d14..925e9f1c

Went back and just rebased instead:
https://github.com/bottlerocket-os/bottlerocket/compare/925e9f1c..11e5885fc2463e2495bcdff0ca7a61d6dd730df0

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+                    error::Error::Usage {
+                        message: format!("Invalid log level '{}'", log_level_str),
+                    }
+                })?);
+            }
+
+            x => {
+                return Err(error::Error::Usage {
+                    message: format!("unexpected argument '{}'", x).into(),
+                })
```
tjkirch:
```suggestion
                return error::Usage {
                    message: format!("unexpected argument '{}'", x),
                }.fail()
```
A few things for this and similar spots.  With snafu, you probably shouldn't be using the `error::Error` enum directly, rather the structs that snafu generates based on them, hence the removal of `Error::`.  It gives you the `fail()` helper to handle the conversion and creation of `Err` around it.  Then, you shouldn't need to use `into()` when creating a snafu error type, since it does that for you.

This situation, where you need to return an error but can't use `ensure!`, is probably where snafu is the least helpful, but it's still best to fit to convention, and make use of snafu's better modes elsewhere.

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -739,3 +739,106 @@ impl Migration for ReplaceTemplateMigration {
         Ok(input)
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// We use this migration when we add metadata and want to make sure they're removed before we go
+/// back to old versions that don't understand them.
+#[derive(Debug)]
+pub struct SettingsMetadata<'a> {
```
tjkirch:
```suggestion
pub struct SettingMetadata<'a> {
```
It's just one setting :)

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -739,3 +739,106 @@ impl Migration for ReplaceTemplateMigration {
...
+    pub setting: &'static str,
+    pub metadata: &'a [&'static str],
```
tjkirch: This list is always going to be static in the code, too, so I think we can avoid the `'a` lifetimes and make it `'static` as well.

packages/os/os.spec
```diff
@@ -97,6 +98,11 @@ Summary: Setting generator for templated settings values.
 %description -n %{_cross_os}schnauzer
 %{summary}.
 
+%package -n %{_cross_os}shibaken
+Summary: Setting generator for populating admin container user-data with IMDS.
```
tjkirch:
```suggestion
Summary: Setting generator for populating admin container user-data from IMDS.
```

Release.toml
```diff
@@ -25,6 +25,6 @@ version = "1.0.5"
     "migrate_v1.0.6_add-static-pods.lz4",
     "migrate_v1.0.6_kubelet-standalone-tls-settings.lz4",
     "migrate_v1.0.6_kubelet-standalone-tls-services.lz4",
-    "migrate_v1.0.6_control-container-v0-4-2.lz4"
+    "migrate_v1.0.6_control-container-v0-4-2.lz4",
```
tjkirch, line 28: I believe this is pointing to the need for a rebase - we should only see shibaken in the diff for the migration list.

sources/api/migration/migration-helpers/src/common_migrations.rs
tjkirch: Similarly, I think the `Vec` in `AddMetadataMigration` could be a `&'static []`, just like this, for the same reason?

sources/api/shibaken/src/main.rs
tjkirch: Oh, this wouldn't work because `request::Error.status()` returns an owned status, not a reference, and there's nowhere for that to live when the function returns, if we're trying to return a reference.  Sorry!

sources/models/shared-defaults/aws-host-containers.toml
```diff
@@ -13,3 +13,6 @@ superpowered = false
 [metadata.settings.host-containers.control.source]
 setting-generator = "schnauzer settings.host-containers.control.source"
 template = "{{ ecr-prefix settings.aws.region }}/bottlerocket-control:v0.4.2"
+
+[metadata.settings.host-containers.admin.user-data]
+setting-generator = "shibaken"
```
bcressey: This should probably move up so it comes after `[metadata.settings.host-containers.admin.source]`.

sources/api/migration/migrations/v1.0.6/add-shibaken/src/main.rs
```diff
@@ -0,0 +1,23 @@
...
+use migration_helpers::common_migrations::{AddMetadataMigration, SettingsMetadata};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting metadata, `host-containers.admin.user-data.setting-generator`
+fn run() -> Result<()> {
+    migrate(AddMetadataMigration(vec![SettingsMetadata {
+        metadata: &["setting-generator"],
+        setting: "settings.host-containers.admin.user-data",
+    }]))
```
bcressey, line 12: @tjkirch - if we may eventually need a `ReplaceMetadataMigration` to complement this one, would it be better to anticipate that with an interface like this?
```
    migrate(ReplaceMetadataMigration(vec![SettingsMetadata {
        metadata: &["setting-generator"],
        new_setting: Some("settings.host-containers.admin.user-data"),
        old_setting: None.
    }]))
```

Having a family of `AddMetadataMigration`, `ReplaceMetadataMigration`, `RemoveMetadataMigration` isn't the worst thing - it has the benefit of being explicit - but it does mean we'd need a second migration binary whenever we're adding as well as replacing.

jpculp:
- Changed `SettingsMetadata` to `SettingMetadata`
- Simplified `parse_args` function logic
- Replaced `'a` lifetimes with `static` in `AddMetadataMigration` code
- Additional clean-up

sources/api/shibaken/src/main.rs
```diff
@@ -0,0 +1,285 @@
...
+use snafu::{OptionExt, ResultExt};
+use std::str::FromStr;
+use std::{env, process};
+
+// Instance Meta Data Service.
+//
+// Currently only able to get fetch session tokens from `latest`
+// FIXME Pin to a date version that supports IMDSv2 once such a date version is available.
+const IMDS_PUBLIC_KEY_BASE_URL: &str = "http://169.254.169.254/latest/meta-data/public-keys";
+const IMDS_SESSION_TOKEN_URI: &str = "http://169.254.169.254/latest/api/token";
```
bcressey: super-nit: it's a little odd to mix `URI` and `URL`, should use one or the other throughout

bcressey (approved):

jpculp:
- Changed `SettingMetadata` from a vec to a slice in `AddMetadataMigration`
- Replaced instances of `url` with `uri` in `shibaken`
- Removed comma from the last element of the (1.0.5, 1.0.6) migration list for consistency

tjkirch (dismissed):

zmrow (approved):
:safety_vest:

jpculp:
- Changed `SettingMetadata` from a vec to a slice in `add-shibaken`
- Added additional commas to migration lists in `Release.toml`. The hope is we can be consistent in adding them to make it easier to append additional migrations between PRs.

tjkirch (approved):

bcressey (approved):

PR #1338, opened 2021-02-21
add support for kubelet in standalone mode and TLS auth

bcressey:
**Issue number:**
N/A

**Description of changes:**
Refactored handling of kubelet args so they can be templated. Added support for "standalone" mode, where the kubelet is started without a kubeconfig file and does not connect to an API server. 

**Testing done:**
Verified that 1.18 and 1.19 nodes came up with the expected defaults: standalone mode off, AWS authentication on.

Launched 1.15, 1.16, 1.17, 1.18, and 1.19 in standalone mode running a static redis pod. Confirmed that kubelet started correctly and brought up redis.

```
[settings.kubernetes]
standalone-mode = true

[settings.kubernetes.static-pods.redis]
manifest = "YXBpVmVyc2lvbjogdjEKa2luZDogUG9kCm1ldGFkYXRhOgogIG5hbWU6IHJlZGlzCnNwZWM6CiAgaG9zdE5ldHdvcms6IHRydWUKICBjb250YWluZXJzOgogIC0gbmFtZTogcmVkaXMKICAgIGltYWdlOiByZWRpczpsYXRlc3QKICAgIHBvcnRzOgogICAgLSBjb250YWluZXJQb3J0OiA2Mzc5CiAgICByZXNvdXJjZXM6CiAgICAgIGxpbWl0czoKICAgICAgICBjcHU6ICIwLjEiCg=="
enabled = true
```

Launched 1.15, 1.16, 1.17, 1.18, and 1.19 in TLS authentication mode with a bootstrap token to join a cluster provisioned by the Cluster API Provider for AWS (CAPA). All nodes joined.
```
❯ kubectl --kubeconfig snowflake.kubeconfig get nodes
NAME                                         STATUS     ROLES    AGE     VERSION
ip-10-0-116-63.us-west-2.compute.internal    NotReady   <none>   3m37s   v1.17.16
ip-10-0-121-4.us-west-2.compute.internal     NotReady   <none>   2m12s   v1.15.12
ip-10-0-123-129.us-west-2.compute.internal   NotReady   <none>   11m     v1.18.14
ip-10-0-73-82.us-west-2.compute.internal     NotReady   <none>   3m43s   v1.16.15
ip-10-0-81-70.us-west-2.compute.internal     NotReady   <none>   73s     v1.19.6
```

(Nodes are in `NotReady` status because I haven't set up a CNI plugin.)

Tested upgrade and downgrade from 1.0.5 into 1.0.6 with the migrations. Verified that new settings were removed and old lists restored on downgrade.

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.15/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.15/kubelet-exec-start-conf
@@ -1,0 +1,17 @@
+[Service]
+ExecStart=
+ExecStart=/usr/bin/kubelet \
+    --cloud-provider aws \
+    --config /etc/kubernetes/kubelet/config \
+    --kubeconfig /etc/kubernetes/kubelet/kubeconfig \
+    --container-runtime=remote \
+    --container-runtime-endpoint=unix:///run/dockershim.sock \
+    --containerd=/run/dockershim.sock \
+    --network-plugin cni \
+    --root-dir /var/lib/kubelet \
+    --cert-dir /var/lib/kubelet/pki \
+    --volume-plugin-dir /var/lib/kubelet/plugins/volume/exec \
+    --node-ip ${NODE_IP} \
+    --node-labels "${NODE_LABELS}" \
+    --register-with-taints "${NODE_TAINTS}" \
+    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE}
--- a/packages/kubernetes-1.15/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.15/kubelet-exec-start-conf
@@ -4,6 +4,9 @@ ExecStart=/usr/bin/kubelet \
 {{~#unless settings.kubernetes.standalone-mode}}
     --cloud-provider aws \
     --kubeconfig /etc/kubernetes/kubelet/kubeconfig \
+{{~#if (eq settings.kubernetes.authentication-mode "tls")}}
+    --bootstrap-kubeconfig /etc/kubernetes/kubelet/bootstrap-kubeconfig \
+{{~/if}}
 {{~else}}
     --cloud-provider "" \
 {{~/unless}}
--- a/packages/kubernetes-1.15/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.15/kubelet-exec-start-conf
@@ -1,9 +1,13 @@
 [Service]
 ExecStart=
 ExecStart=/usr/bin/kubelet \
+{{~#unless settings.kubernetes.standalone-mode}}
     --cloud-provider aws \
-    --config /etc/kubernetes/kubelet/config \
     --kubeconfig /etc/kubernetes/kubelet/kubeconfig \
+{{~else}}
+    --cloud-provider "" \
+{{~/unless}}
+    --config /etc/kubernetes/kubelet/config \
     --container-runtime=remote \
     --container-runtime-endpoint=unix:///run/dockershim.sock \
     --containerd=/run/dockershim.sock \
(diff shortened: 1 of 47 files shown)


webern:
Now I see why you knew about the `~` to avoid a newline. I was like wut? @bcressey is a mustache expert?

zmrow:
He does grow a pretty healthy mustache.... :thinking:

packages/kubernetes-1.15/kubelet-exec-start-conf
```diff
@@ -0,0 +1,24 @@
+[Service]
+ExecStart=
+ExecStart=/usr/bin/kubelet \
+{{~#unless settings.kubernetes.standalone-mode}}
```
zmrow, line 4: Any reason not to use an `if` here vs. an `unless/else`?

packages/kubernetes-1.15/kubelet.service
```diff
@@ -16,21 +16,8 @@ ExecStartPre=/usr/bin/host-ctr \
...
-    --network-plugin cni \
-    --root-dir /var/lib/kubelet \
-    --cert-dir /var/lib/kubelet/pki \
-    --volume-plugin-dir /var/lib/kubelet/plugins/volume/exec \
-    --node-ip ${NODE_IP} \
-    --node-labels "${NODE_LABELS}" \
-    --register-with-taints "${NODE_TAINTS}" \
-    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE}
+# Must be overridden by a drop-in file or `kubelet` won't start
+ExecStart=/usr/bin/false
```
zmrow, line 20: Maybe I missed it, but where does this get overridden?

bcressey: The first `ExecStart=` line in `kubelet-exec-start.conf` clears out this value, and the second one sets it.

packages/kubernetes-1.15/kubelet-exec-start-conf
bcressey: At first I wasn't conditionalizing the `--cloud-provider` argument, so there was no `else` clause and I thought it read better.

I kept it since I expected the top branch to continue expanding while the bottom branch should always just be turning off the cloud provider, and it can go away in 1.23 when support for cloud providers in kubelet goes away.

zmrow: Ah, ok - seems good!

zmrow (approved):
:coffee:

etungsten (approved):

sources/models/src/lib.rs
```diff
@@ -121,6 +122,11 @@ struct KubernetesSettings {
     node_labels: HashMap<KubernetesLabelKey, KubernetesLabelValue>,
     node_taints: HashMap<KubernetesLabelKey, KubernetesTaintValue>,
     static_pods: HashMap<Identifier, StaticPod>,
+    bootstrap_token: KubernetesBootstrapToken,
+
+    // Settings with useful default values that can be overridden.
```
tjkirch: nit: I think these headers are getting confusing the more settings we add, because we don't "require the user to specify" bootstrap token, it's just for one mode of operation.  Rather than adding another header here and codifying it, I'd vote for getting rid of / changing the first two headers, and keeping / improving the "dynamic settings" one, to make it clear that those are actually the special ones that users probably don't want to set themselves.

tjkirch (approved):

tjkirch (approved):

PR #1339, opened 2021-02-22
Clarify use case for update repos

gregdek:
Just looking to make it a little more clear that this step isn't necessary for users who don't intend to use update functionality.

**Terms of contribution:**

### DIFF
--- a/PUBLISHING.md
+++ b/PUBLISHING.md
@@ -79,7 +79,7 @@ cargo make revoke-ami -e REVOKE_FROM_USERS=0123456789,9876543210
 
 ## Build a repo
 
-__NOTE: If you intend to replace hosts rather than updating them, you don't need to build an update repository.__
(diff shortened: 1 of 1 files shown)


PUBLISHING.md
```diff
@@ -79,10 +79,11 @@ cargo make revoke-ami -e REVOKE_FROM_USERS=0123456789,9876543210
 
 ## Build a repo
 
+__NOTE: If you intend to replace hosts rather than updating them, you don't need to build an update repository.__
```
tjkirch: Can we use the same note format as elsewhere in the doc, e.g. a few lines up?

```suggestion
> NOTE: If you intend to replace hosts rather than update them, you don't need to build an update repository.
```

tjkirch (approved):

samuelkarp (approved):

zmrow (approved):
Thanks!

:cocktail:

srgothi92 (approved):

Issue #1140, opened 2020-10-02
Pass additional arguments to the kubelet


itomaldonado:
**What I'd like:**
Running on AWS + EKS, the bottlerocket worker nodes are in an ASG and currently start properly and become ready. BBut we would like to be able to pass in extra args to the kubelet (similar to what the eks images can do), for example:
```
--eviction-hard=memory.available<560Mi
--allowed-unsafe-sysctls net.core.somaxconn,net.ipv4.ip_local_port_range
```

**Any alternatives you've considered:**
Building our own image, although seems a bit cumbersome when all that is needed is to add one or one or two extra args to kubectl.

zmrow:
Thanks for opening this issue! We don't currently have a good solution for arbitrary arguments like this, but it's something that's on our radar.

ugurarpaci:
This will be pretty useful! Adding these flags will satisfy the benchmark tools like [kube-bench](https://www.eksworkshop.com/intermediate/300_cis_eks_benchmark/run-as-job/)

itomaldonado:
@zmrow that's fair, I would love to follow along an issue or something that tracks that functionality, if there is one (and could be linked here). If so you can also feel free to close this. If there isn't one, could this one track this functionality?

As @ugurarpaci mentioned there are quite a few things we could control in our clusters by just changing kubelet flags instead of having to build an image for each.

bcressey:
We don't have plans to allow arbitrary arguments to kubelet, but we can and should model arguments like this as settings.

The reason we'd like to avoid arbitrary arguments is so that we can better test and predict kubelet behavior across upgrades and downgrades; otherwise removing the argument or changing how it's interpreted could break the system in ways that we can't anticipate.

This is especially a concern when we introduce support for cross-variant upgrades, for example from Kubernetes 1.18 to 1.19. Having a model for the settings also lets us specify a migration path for them - causing the setting to go away on upgrade and reappear on downgrade, or moving it from a command-line argument into the kubelet config.

Are these the only two flags you're looking for, or are there others?

svedatbegar:
As an addition to parameters mentioned above, we use kube-reserved parameters to reduce allocatable resources on nodes so there's always enough CPU, memory, ephemeral storage and pid for stability. So far I couldn't see a way to set them.

supasteev0:
Just wanted to share my experience in testing bottlerocket AMI: 
We currently use the following kubelet args with the standard EKS AMI: 
[...]

PR #1341, opened 2021-02-23
[Draft] Add new setting: eviction-hard

gthao313:
**Get some feedbacks from this draft, which can help me in implementing another three settings**

**Issue number:**
#1140

**Description of changes:**
Add new setting `eviction-hard` to k8s. User can specify this setting in userdata.
* example: 
     "memory.available" = "10Gi"
     "nodefs.inodesFree" = “15%”

**Testing done:**
Build aws-eks-* images and launched instance.

***Test step1***:
Go to control container run `apiclient -u /settings` to check if expected setting is there.
`eviction-hard":{"memory.available":"99.999%"} `

***Test step2***:
ssh to admin contianer and look `etc/kubernetes/kubelet/config` to check if expected config is there.
```
evictionHard:
  memory.available: 15%
```
***Test step3*** - test effected/behavior:
set up eviction threshold to a high value for triggering eviction.
```
[evictionHard.memory]
"memory.available" = "99.99%" (high percentage to trigger eviction)
```
a) Go to EKS to check node condition 

MemoryPressure | True | kubelet has insufficient memory available
-- | -- | --

b) In admin contianer, check kubelet log by `journalctl -u kubelet`
```
eviction_manager.go:335] eviction manager: attempting to reclaim memory
eviction_manager.go:346] eviction manager: must evict pod(s) to reclaim memory
```

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,6 +306,16 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.standalone-mode`: Whether to run the kubelet in standalone mode, without connecting to an API server.  Defaults to `false`.
 * `settings.kubernetes.authentication-mode`: Which authentication method the kubelet should use to connect to the API server, and for incoming requests.  Defaults to `aws` for AWS variants, and `tls` for other variants.
 * `settings.kubernetes.bootstrap-token`: The token to use for [TLS bootstrapping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/).  This is only used with the `tls` authentication mode, and is otherwise ignored.
+* `settings.kubernetes.eviction-hard`: The hard eviction Signal and Thresholds you set up to reclaim the associated starved resource.
+  Remember to quote keys (since they often contain ".") and to quote all values.
+  * Example user data for setting up eviction hard:
+  ```
+    [settings.kubernetes.eviction-hard]
+    "memory.available" = "15%"
+    "nodefs.inodesFree" = "15Mi"
+    "pid.available" = "15%"
+    ```
+
 
 You can also optionally specify static pods for your node with the following settings.
 Static pods can be particularly useful when running in standalone mode.
--- a/packages/kubernetes-1.18/kubelet-config
+++ b/packages/kubernetes-1.18/kubelet-config
@@ -29,6 +29,8 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 clusterDNS:
 - {{settings.kubernetes.cluster-dns-ip}}
+evictionHard:
+  {{join_map ": " "\n  " "no-fail-if-missing" settings.kubernetes.eviction-hard}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 cgroupDriver: systemd
--- a/sources/models/src/modeled_types/kubernetes.rs
+++ b/sources/models/src/modeled_types/kubernetes.rs
@@ -437,3 +437,120 @@ mod test_kubernetes_bootstrap_token {
         }
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// EvictionHardKey represents a string that contains a valid Kubernetes eviction hard
+/// signal. There are few valid eviction hard signals [memory.available], [nodefs.available],
+/// [imagefs.available], and [nodefs.inodesFree].
+/// https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/
+
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct EvictionHardKey {
+    inner: String,
+}
+
+impl TryFrom<&str> for EvictionHardKey {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        let evitionsignal = vec![
+            "memory.available","nodefs.available",
+            "nodefs.inodesFree","imagefs.available",
+            "imagefs.inodesFree","pid.available"
+            ];
+
+        ensure!(
+            evitionsignal.contains(&input),
+            error::InvalideEvictionHard {
+                input,
+                msg: format!("must be one of designated signals"),
+            }
+        );
+
+        Ok(EvictionHardKey {
+            inner: input.to_string(),
+        })
+    }
+}
+string_impls_for!(EvictionHardKey, "EvictionHardKey");
+
+#[cfg(test)]
+mod test_kubernetes_eviction_hard_key {
+    use super::EvictionHardKey;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn good_eviction_hard_key() {
+        for ok in &[
+            "memory.available",
+            "nodefs.available",
+            "nodefs.inodesFree",
+            "imagefs.available",
+            "imagefs.inodesFree",
+            "pid.available",
+            ] {
+            EvictionHardKey::try_from(*ok).unwrap();
+        }
+    }
+
+    #[test]
+    fn bad_eviction_hard_key() {
+        for err in &["", "storage.available", ".bad", "bad.", &"a".repeat(64)] {
+            EvictionHardKey::try_from(*err).unwrap_err();
+        }
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// EvictionHardValue represents a string that contains a valid Kubernetes eviction threshold quantity
+/// An eviction threshold can be expressed as Gi/Mi or a percentage using the % token
+/// https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/
+
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct EvictionHardValue {
+    inner: String,
+}
+
+impl TryFrom<&str> for EvictionHardValue {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+
+        ensure!(
+            input.ends_with("Gi") || input.ends_with("Mi") || input.ends_with("%"),
+            error::InvalideEvictionHard {
+                input,
+                msg: format!("must be ends with Gi, Mi, or %"),
+            }
+        );
+
+        Ok(EvictionHardValue {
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -306,6 +306,16 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.standalone-mode`: Whether to run the kubelet in standalone mode, without connecting to an API server.  Defaults to `false`.
 * `settings.kubernetes.authentication-mode`: Which authentication method the kubelet should use to connect to the API server, and for incoming requests.  Defaults to `aws` for AWS variants, and `tls` for other variants.
 * `settings.kubernetes.bootstrap-token`: The token to use for [TLS bootstrapping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/).  This is only used with the `tls` authentication mode, and is otherwise ignored.
+* `settings.kubernetes.eviction-hard`: The hard eviction Signal and Thresholds you set up to reclaim the associated starved resource.
+  Remember to quote keys (since they often contain ".") and to quote all values.
+  * Example user data for setting up eviction hard:
+  ```
```
gthao313, line 312: I'll fix this indent.

README.md
```diff
@@ -306,6 +306,16 @@ The following settings are optional and allow you to further configure your clus
...
+    [settings.kubernetes.eviction-hard]
+    "memory.available" = "15%"
+    "nodefs.inodesFree" = "15Mi"
+    "pid.available" = "15%"
+    ```
+
 
 You can also optionally specify static pods for your node with the following settings.
```
gthao313, line 320: Question: There are four settings, and wonder if it's better to make four PRs for them (one PR for each setting.) Ben mentioned that we should make them in different commits and same PR. I think maybe it's better to have four PR, because all of them have different testing and implementation requirement. Can I get some suggestions on that? thanks!

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,120 @@ mod test_kubernetes_bootstrap_token {
...
+        }
+    }
+
+    #[test]
+    fn bad_eviction_hard_value() {
+        for err in &["", "bad", "100", &"a".repeat(64)] {
+            EvictionHardValue::try_from(*err).unwrap_err();
+        }
+    }
+}
```
srgothi92, line 556: nit: new line

sources/models/src/modeled_types/mod.rs
```diff
@@ -62,6 +62,8 @@ pub mod error {
             field: String,
             source: serde_plain::Error,
         },
+        #[snafu(display("Invalid eviction hard '{}': {}", input, msg))]
+        InvalideEvictionHard { input: String, msg: String },
```
srgothi92, line 66: we can create two different error enums, one for Key and other for value. So, that you don't have to specifically pass a msg string and we can include msg string as part of `#[snafu(display` to create string you want.
Did you refer it from somewhere, where we are using same enum ? I checked `ECSAttributeKey` and `ECSAttributeValue` we are using different error enum there.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,120 @@ mod test_kubernetes_bootstrap_token {
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct EvictionHardKey {
+    inner: String,
+}
+
+impl TryFrom<&str> for EvictionHardKey {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        let evitionsignal = vec![
```
srgothi92, line 458: we can also use enum here for example:  https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/src/modeled_types/ecs.rs#L189
However, I don't know which aligns more to our existing style, I would let other comment on what we should use.

packages/kubernetes-1.18/kubelet-config
```diff
@@ -29,6 +29,8 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 clusterDNS:
 - {{settings.kubernetes.cluster-dns-ip}}
+evictionHard:
```
srgothi92, line 32: what would happen if user does not set this setting ? I guess this setting would still exist in kubelet config without any value. Should we wrap setting in if ? for example:
```
{{~#if settings.kubernetes.evictionHard }}
evictionHard:
      {{join_map ": " "\n  " "no-fail-if-missing" settings.kubernetes.eviction-hard}}
{{~/if}}
```
I would let others weigh in, I am also learning in this commit :)

srgothi92 (commented):
Nice ! 🎉

packages/kubernetes-1.18/kubelet-config
gthao313: Correct. The `eviction-hard` setting would still there and would not effect `kubelet start`. Your example is a good suggestion to me. Thanks 👍

sources/models/src/modeled_types/kubernetes.rs
gthao313: I was trying to use enum to handle this part; however, I think enum could be not that efficient and clear there because too many works on dealing with variables which contains dots like `memory.available`.  I got some ideas that if we can use rename attributes for each enum variant `#[serde(rename = "memory.available")] `, and I'm not sure if that worths. Appreciate for the comments! :)

README.md
etungsten: If the commits have no inter-dependencies and can stand alone by themselves I don't think we have to have them in the same PR. That being said, if you already have them ready, and they're similar in scope to this commit you have out right now, it'd be nice to have them all in the PR and reviewers can just review commit by commit instead of having to jump through 4 PRs that might end up with similar review comments.

sources/models/src/modeled_types/mod.rs
gthao313: Your suggestion could be a way as well. The reason why I make them as one is trying to make simple and clear. and I agree with you that different error enums could be a better way. Btw, I got this idea from `KubernetesLableKey` and `KubernetesLabelValue`.

sources/models/src/modeled_types/kubernetes.rs
etungsten: Another plus to just having it as an enum is being able to reuse it elsewhere (e.g. in your tests) for matching/validation etc.

I also wonder if we even need any of the traits here for converting to/from strings if `EvictionHardKey` itself was the enum. We've never tried that before but I don't see a clear reason why we can't just rely on serde's deserialization trait (with the appropriate attributes like you mentioned) for validation.

sources/models/src/modeled_types/mod.rs
etungsten: The reason why we have error variants like this is because there might be a lot of special cases we check for and capturing each one as a separate error variant would be too much when all we want to convey is the value is invalid for so-and-so reason. An example of this is `InvalidSysctlKey`.

If we only return this error in one location I would use `#[snafu(display)]`. If there's more than one check that check for different things that might return this, then I think having it like this is fine.

README.md
zmrow: +1 to that

packages/kubernetes-1.18/kubelet-config
zmrow: Yes - I think we should wrap this in an `if`.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,120 @@ mod test_kubernetes_bootstrap_token {
         }
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// EvictionHardKey represents a string that contains a valid Kubernetes eviction hard
+/// signal. There are few valid eviction hard signals [memory.available], [nodefs.available],
```
zmrow, line 444: nit: Rather than naming the currently available values, let's just point to the relevant k8s page so if more are added or if they change, we don't need to remember to update this comment.

`Available eviction hard signals: <link here>`

sources/models/src/modeled_types/kubernetes.rs
zmrow: @gthao313 I think the rename attributes you mentioned above _should_ allow us to use an enum.  If so, that's probably the preferred route as other folks mentioned.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,120 @@ mod test_kubernetes_bootstrap_token {
...
+        for err in &["", "storage.available", ".bad", "bad.", &"a".repeat(64)] {
+            EvictionHardKey::try_from(*err).unwrap_err();
+        }
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// EvictionHardValue represents a string that contains a valid Kubernetes eviction threshold quantity
+/// An eviction threshold can be expressed as Gi/Mi or a percentage using the % token
```
zmrow, line 509: Same down here - let's just point to the docs rather than reiterating them.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,120 @@ mod test_kubernetes_bootstrap_token {
...
+    inner: String,
+}
+
+impl TryFrom<&str> for EvictionHardValue {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+
+        ensure!(
+            input.ends_with("Gi") || input.ends_with("Mi") || input.ends_with("%"),
```
zmrow, line 524: Let's make sure that these are the only values that are possible (`Gi`, `Mi`, `%`)  I took a peek at the doc page, but perhaps I missed the listing of acceptable values.

sources/models/src/modeled_types/mod.rs
zmrow, line 66: `s/InvalideEvictionHard/InvalidEvictionHard`  (There's an extra "e" after Invalid)

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,120 @@ mod test_kubernetes_bootstrap_token {
...
+            error::InvalideEvictionHard {
+                input,
+                msg: format!("must be ends with Gi, Mi, or %"),
```
zmrow, line 527: "must end with ..."

PR #1357, opened 2021-02-26
fix build-kmod-kit build failure

sanu11:
**Issue number:**
NA

**Description of changes:**
The build was failing on the `build-kmod-kit` task on one host with the following error 

`WARNING: IPv4 forwarding is disabled. Networking will not work.       
curl: (6) Could not resolve host: cache.bottlerocket.aws`

Added a fix `--network=host` to the build-kmod-kit task's `docker run`  command and it worked..

**Testing done:**
`cargo make build-kmod-kit` worked

**Terms of contribution:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 ## Build Changes
 
 * Add the ability to output vmdk via qemu-img ([#1289])
-* Add support for kmod kits to ease building of third-party kernel modules ([#1287], [#1286], [#1285])
+* Add support for kmod kits to ease building of third-party kernel modules ([#1287], [#1286], [#1285],[#1357])
 * storewolf: Declare dependencies on model and defaults files ([#1319])
 * storewolf: Refactor default settings files to allow sharing ([#1303], [#1329])
 * Switch from TermLogger to SimpleLogger ([#1282], **thanks @hencrice!**)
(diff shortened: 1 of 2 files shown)


bcressey (approved):

samuelkarp (approved):
It might be worth updating the commit message to have a bit more data.  For reference, some hosts might have the `net.ipv4.ip_forward` sysctl set to `0`, but we still want those hosts to be able to build Bottlerocket.

Given that the image in question is the SDK image, the script isn't doing anything funky, and we're running inside the container as non-root this looks fine to me.

sanu11:
> It might be worth updating the commit message to have a bit more data. For reference, some hosts might have the `net.ipv4.ip_forward` sysctl set to `0`, but we still want those hosts to be able to build Bottlerocket.
> 
> Given that the image in question is the SDK image, the script isn't doing anything funky, and we're running inside the container as non-root this looks fine to me.

Sure, Would update the commit message. Thanks!

webern (approved):
Good when CI passes.

sanu11:
> @sanu11 sanu11 force-pushed the sanu11:develop branch from 113e44a to 24c31df now

Added description to commit message

samuelkarp (approved):

tjkirch (approved):
Would you please add a commit that updates the changelog and adds this PR ID/link to the entry for kmod kits?

sanu11:
> @sanu11 sanu11 force-pushed the sanu11:develop branch from 6fd0845 to 2c188c0
 Resolved conflicts, pulled with rebase.

PR #1358, opened 2021-02-26
shibaken: return empty key list if IMDS returns 404

jpculp:
**Issue number:**

N/A

**Description of changes:**

In the case where there are no keys available in IMDS, shibaken will create an empty public key list instead of returning an error.

This addresses a rare scenario where a user launches an instance of Bottlerocket without attaching a key to the EC2 instance. Rather than returning an empty string, the IMDS request for available keys returns a 404. The 404 status causes shibaken to return an error and prevents Bottlerocket from booting.

**Testing done:**

Set `IMDS_PUBLIC_KEY_BASE_URI` to an invalid URI and cargo ran `shibaken`
> @webern
I have tested these changes. Before these changes an instance without a key failed to boot. With these changes the instance came up and I ran an ECS task. 

**Terms of contribution:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Add a vmware-dev variant ([#1292], [#1288], [#1290])
 * Add Kubernetes static pods support ([#1317])
 * Add high-level 'set' subcommand for changing settings using apiclient ([#1278])
-* Allow admin container to use SSH public keys from user data ([#1331], [#19])
+* Allow admin container to use SSH public keys from user data ([#1331], [#1358], [#19])
 * Add support for kubelet in standalone mode and TLS auth ([#1338])
 * Add https-proxy and no-proxy settings to updog ([#1324])
 * Add support for pulling host-containers from ECR Public ([#1296])
@@ -80,6 +80,7 @@
 [#1353]: (https://github.com/bottlerocket-os/bottlerocket/pull/1353)
 [#1356]: (https://github.com/bottlerocket-os/bottlerocket/pull/1356)
 [#1357]: (https://github.com/bottlerocket-os/bottlerocket/pull/1357)
+[#1358]: (https://github.com/bottlerocket-os/bottlerocket/pull/1358)
 [#19]: (https://github.com/bottlerocket-os/bottlerocket-admin-container/pull/19)
 
 # v1.0.5 (2021-01-15)
--- a/sources/api/shibaken/src/main.rs
+++ b/sources/api/shibaken/src/main.rs
@@ -60,17 +60,23 @@ fn fetch_imds_session_token(client: &Client) -> Result<String> {
     Ok(imds_session_token)
 }
 
-/// Helper to fetch data from IMDS. Taken from pluto.
-fn get_text_from_imds(client: &Client, uri: &str, session_token: &str) -> Result<String> {
-    client
+/// Helper to fetch data from IMDS. Inspired by pluto.
+fn fetch_from_imds(client: &Client, uri: &str, session_token: &str) -> Result<Option<String>> {
+    let response = client
         .get(uri)
         .header("X-aws-ec2-metadata-token", session_token)
         .send()
-        .context(error::ImdsRequest { method: "GET", uri })?
-        .error_for_status()
-        .context(error::ImdsResponse { uri })?
-        .text()
-        .context(error::ImdsText { uri })
+        .context(error::ImdsRequest { method: "GET", uri })?;
+    if response.status().as_u16() == 404 {
+        return Ok(None);
+    }
+    Ok(Some(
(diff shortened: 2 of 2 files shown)


tjkirch (approved):
Would you please also add a commit that adds this PR ID/link to the changelog entry for the shibaken set of changes?

sources/api/shibaken/src/main.rs
```diff
@@ -60,17 +60,23 @@ fn fetch_imds_session_token(client: &Client) -> Result<String> {
...
+    if response.status().as_u16() == 404 {
+        return Ok(None);
+    }
+    Ok(Some(
+        response
+            .error_for_status()
+            .context(error::ImdsResponse { uri })?
+            .text()
+            .context(error::ImdsText { uri })?,
+    ))
```
srgothi92, line 79: clippy should have suggested this. I guess. 
```suggestion
    if response.status().as_u16() == 404 {
         Ok(None)
    } else {
        Ok(Some(
        response
            .error_for_status()
            .context(error::ImdsResponse { uri })?
            .text()
            .context(error::ImdsText { uri })?
          )))
    }
```

bcressey (approved):

sources/api/shibaken/src/main.rs
```diff
@@ -81,17 +87,24 @@ fn fetch_public_keys_from_imds() -> Result<Vec<String>> {
...
-    debug!("available public keys '{}'", &public_key_list);
+    let public_key_list = if let Some(public_key_list) =
+        fetch_from_imds(&client, IMDS_PUBLIC_KEY_BASE_URI, &imds_session_token)?
+    {
+        debug!("available public keys '{}'", &public_key_list);
+        public_key_list
+    } else {
+        debug!("no available public keys");
+        return Ok(Vec::new());
+    };
```
srgothi92, line 98: May be this is more rust idiomatic ? You can workout the var names sorry for that.
```suggestion
let resp =  fetch_from_imds(&client, IMDS_PUBLIC_KEY_BASE_URI, &imds_session_token)?;
        match resp {
            None => {
                    debug!("no available public keys");
                   Ok(Vec::new())
            },
            Some(key_list) => {
                        debug!("available public keys '{}'", &public_key_list);
                      key_list
            }
        }
    
```

bcressey: `if let / else` feels more idiomatic to me, or at least it's how I'd write it in this case.

sources/api/shibaken/src/main.rs
tjkirch: It's a common pattern to return early during an error check, and to keep the happy path clear.  It makes it easier and clearer during later changes.

webern:
I have tested these changes. Before these changes an instance without a key failed to boot. With these changes the instance came up and I ran an ECS task.

sources/api/shibaken/src/main.rs
tjkirch: It's common and shorter to use if/else for short conditionals like this, particularly with `if let Some(x)`.

srgothi92: Okay. I was more worried about return, because clippy complains me for that. It also complains about `let public_key_list = if let Some(public_key_list) =
        fetch_from_imds(&client, IMDS_PUBLIC_KEY_BASE_URI, &imds_session_token)?` 
        and asks to remove from If closure block something.
 if you feel it is okay then we are good :). I have written lot less rust.

jpculp: I actually had `Some(s)` originally, but it looked a little bizarre having a single character on Line 94. Using `public_key_list` looked a little nicer and seemed to provide more context, but I can change it back if folks prefer.

srgothi92 (approved):

webern (approved):
Upon _most_ variant arches passing CI. :)

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 * Add a vmware-dev variant ([#1292], [#1288], [#1290])
 * Add Kubernetes static pods support ([#1317])
 * Add high-level 'set' subcommand for changing settings using apiclient ([#1278])
-* Allow admin container to use SSH public keys from user data ([#1331], [#19])
+* Allow admin container to use SSH public keys from user data ([#1331], [#1358], [#19])
```
tjkirch, line 9: Also need to include the full link for this down below in the list, or GitHub won't render it the same way.

tjkirch (approved):

jpculp:
- Added full link to the PR in the changelog.

jpculp:
- Rebased and addressed merge conflict in `CHANGELOG.md`

bcressey (approved):

tjkirch (approved):

PR #1360, opened 2021-03-01
docs: update quickstarts to have arm64 examples

samuelkarp:
**Description of changes:**
Update the quickstart docs to have examples showing how to find `arm64` AMI IDs.

**Testing done:**
None

**Terms of contribution:**

### DIFF
--- a/QUICKSTART-ECS.md
+++ b/QUICKSTART-ECS.md
@@ -59,6 +59,12 @@ aws ssm get-parameters --region us-west-2 \
 You can replace the architecture (`x86_64`) and region (`us-west-2`) to look for other images.
 If you know a specific Bottlerocket version you'd like to use, you can replace `latest` with that version.
 
+For example, to find the `1.0.0` version of the `aws-ecs-1` variant for the `arm64` architecture (also known as `aarch64`) in the `eu-west-1` region, you can use the following command:
+
+```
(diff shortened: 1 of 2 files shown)


etungsten (approved):

zmrow (approved):
:singer:

srgothi92 (approved):

QUICKSTART-ECS.md
```diff
@@ -59,6 +59,12 @@ aws ssm get-parameters --region us-west-2 \
 You can replace the architecture (`x86_64`) and region (`us-west-2`) to look for other images.
 If you know a specific Bottlerocket version you'd like to use, you can replace `latest` with that version.
 
+For example, to find the `1.0.0` version of the `aws-ecs-1` variant for the `arm64` architecture in the `eu-west-1` region, you can use the following command:
```
tjkirch:
```suggestion
For example, to find the `1.0.0` version of the `aws-ecs-1` variant for the `arm64` architecture (also known as `aarch64`) in the `eu-west-1` region, you can use the following command:
```

(And in the other file.)

This might help with the confusion that resulted in this change.

It might also help to explain _why_ we use that term here, but I think that could be a step too far, and we should (separately) consider putting out parameters for both arch names (and `x86_64`/`amd64`) so that doesn't even come up.

tjkirch (approved):

QUICKSTART-ECS.md
samuelkarp: Done.  Agreed about putting out parameters for multiple arch names, but that's separate from this docs change.

tjkirch (approved):

PR #1366, opened 2021-03-05
Support gzipped user data

### LINKED ISSUE #1320: add support for gzip compressed user data to early-boot-config
**What I'd like:**
`early-boot-config` should support gzip compressed user data, in addition to uncompressed TOML.

With #1244 and #1317 we could get enough input that we'd otherwise run into the [16 KiB limit](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-add-user-data.html) for EC2 instance user data. Compression will make this channel more useful.

Continue to accept plain text only. This would require people to switch to host containers or post-boot configuration mechanisms such as SSM workflows for more scenarios, which is relatively more difficult to coordinate.

tjkirch:
**Issue number:**

Fixes #1320.  The goal is adding an easy way to handle larger amounts of user data, now that we support adding static pods, host container user data, and other settings that could potentially be long.  (This isn't the only solution or the final one, but it's an easy use case.)

**Description of changes:**

```
778760a7 early-boot-config: add module to handle maybe-gzip inputs
7a558d9c early-boot-config: allow gzip compression of user data
```

Note: this is based on #1365 because I needed the model fix to properly run the vmware-dev variant.

**Testing done:**

#### aws-k8s-1.17

Started an instance with aws-cli's `--user-data fileb://path` argument pointing to a gzipped file with >16KiB of text uncompressed, saw those settings take effect (very long motd), and the instance was healthy.

Started the same AMI with a (smaller) uncompressed user data file and was still successful/healthy.

#### aws-dev

Used an uncompressed local user data file OK:

```
$ sudo sheltie
bash-5.0# mkdir /etc/early-boot-config
bash-5.0# echo -e '[settings]\nmotd = "woo"' > /etc/early-boot-config/user-data
bash-5.0# early-boot-config
bash-5.0# settings-committer
bash-5.0# thar-be-settings --all
bash-5.0# cat /etc/motd
woo
```

Then I did the same on a new instance, but outside of sheltie (so gzip is available) with `echo -e '[settings]\nmotd = "compressed"' | gzip | sudo tee user-data`.  Still saw the motd update OK.

#### vmware-dev

I ran vmware-dev in qemu, and tested changing the motd with a plain uncompressed TOML file in a ISO user data; worked fine.

I tested with a compressed user-data file OK, confirming that it's compressed at /media/cdrom/user-data but uncompressed OK into settings.motd and the motd file.

Then I repeated the tests with an OVF file inside the ISO.  Plain text was OK, I confirmed /media/cdrom/ovf_env.xml has the plain text base64, and the API/motd shows the text.  Compressed was OK, I confirmed /media/cdrom/ovf_env.xml has the compressed text base64, and the API/motd shows uncompressed text.

**Terms of contribution:**


sources/api/early-boot-config/src/compression.rs
```diff
@@ -0,0 +1,205 @@
+//! This module supports reading from an input source that could be compressed or plain text.
+//!
+//! Currently gzip compression is supported.
+
+use flate2::read::GzDecoder;
+use std::fs::File;
+use std::io::{BufReader, Chain, Cursor, ErrorKind, Read, Result, Take};
+use std::path::Path;
+
+/// "File magic" that indicates file type is typically two bytes at the start.
```
zmrow: Could we clarify this a little bit more?  Is this mostly true for all file types?

sources/api/early-boot-config/src/compression.rs
```diff
@@ -0,0 +1,205 @@
...
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use hex_literal::hex;
+    use lazy_static::lazy_static;
+    use std::io::Cursor;
+
+    lazy_static! {
```
zmrow, line 166: Nice!

sources/api/early-boot-config/src/compression.rs
tjkirch: It's not always true, for example bzip2 is 3 bytes, but at least a couple other compression/packaging formats are two bytes.  ([reference](https://en.wikipedia.org/wiki/List_of_file_signatures))  I expect this to have to change.  Probably we'd read in the max number of bytes needed for any compression format we care about, and then check the appropriate length prefix of it.  I'll try to clarify the comment.

tjkirch:
^ This push clarifies the comment about file magic, and explicitly changes the gzip const length to 2, to represent the fact that it may differ in the future.

tjkirch:
^ @bcressey gave the nice suggestion of a unit test that confirms giving _part_ of the magic string works, in that it just gives back the input, it's not falsely recognized as gzip or anything.  Added that in this push; test passes.

bcressey (approved):

sources/api/early-boot-config/src/compression.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+impl<R: Read> RetryRead<R> for R {
+    // This implementation is based on stdlib Read::read_exact, but hitting EOF isn't a failure, we
+    // just want to return the number of bytes we could read.
+    fn retry_read(&mut self, mut buf: &mut [u8]) -> Result<usize> {
+        let mut count = 0;
+
+        // Read until we have no more space in the output buffer
+        while !buf.is_empty() {
```
zmrow, line 136: Is there a chance we could get stuck in an infinite loop here?  Should we have an upper bound for retries?  We could set it decently high to give us plenty of chances but ensure we never get stuck on this.

I think the chances of this happening are pretty slim, but thought it's worth calling out.

sources/api/early-boot-config/src/compression.rs
```diff
@@ -0,0 +1,222 @@
...
+/// https://en.wikipedia.org/wiki/List_of_file_signatures
+const MAGIC_LEN: usize = 2;
+
+// We currently only support gzip, but it shouldn't be hard to add more.
+/// These bytes are at the start of any gzip-compressed data.
+const GZ_MAGIC: [u8; 2] = [0x1f, 0x8b];
+
+/// This helper takes a slice of bytes representing UTF-8 text, which can optionally be
+/// compressed, and returns an uncompressed string.
+pub fn expand_slice_maybe(input: &[u8]) -> Result<String> {
```
webern, line 23: maybe `decode_slice` and `decode_file`? Either way you are decoding, 1) UTF-8 bytes -> `String`, or 2) gzipped bytes -> UTF-8 bytes -> `String`.

sources/api/early-boot-config/src/compression.rs
tjkirch: No, I don't think there's a chance of an infinite loop.  Either we read some bytes, in which case we move closer to our goal of 2, or we read no bytes and we break out of the loop.  (`buf` itself is changed on each read, sliding forward by the number of bytes read.)

It's the same logic as `read_exact` in the standard library: https://doc.rust-lang.org/src/std/io/mod.rs.html#770-787

sources/api/early-boot-config/src/compression.rs
```diff
@@ -0,0 +1,222 @@
...
+    let mut reader = OptionalCompressionReader::new(BufReader::new(file));
+    reader.read_to_string(&mut output)?;
+    Ok(output)
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// This type lets you wrap a `Read` whose data may or may not be compressed, and its `read()`
+/// calls will uncompress the data if needed.
+pub enum OptionalCompressionReader<R> {
```
webern: I suppose this weird enum could be hidden?

```suggestion
pub struct OptionalCompressionReader<R>(WeirdEnum);

pub enum WeirdEnum<R> {
```

sources/api/early-boot-config/src/compression.rs
tjkirch: I don't think `decode` gives any clue to the reader that it deals with compression.  You'd typically import the function with a `use` and not have the module name as a clue.

tjkirch:
^ This push moves the main type to a struct, per @webern.

webern (approved):

zmrow (approved):
:microscope:

Issue #1269, opened 2021-01-09
update to tough v0.11.0, tokio 1.0, rusoto 0.46, reqwest v0.11.


webern:
The rust packages listed in the title are intertwined in various ways.

Multiple major-versions of tokio can not be in use at a given time. rusoto 0.46, hyper 14 and reqwest v0.11 all bump to tokio v1.0.  Meanwhile, we cannot have multiple versions of reqwest in the sources workspace because of the way webpki-root-shim works. So, basically what needs to happen is this https://github.com/awslabs/tough/issues/269 needs to land and get published (I'm calling it v0.11.0 since v0.10.0 is already a huge diff), and then all the Bottlerocket stuff can update.

Edit: remember to remove the https-proxy workarounds in metricdog and updog, see #1332

tjkirch:
inotify is another dependency that updated because of tokio.

tjkirch:
Also httptest.

tjkirch:
Also coldsnap, which needs to do a new release that includes the change it made to use tokio 1.

webern:
> Also coldsnap, which needs to do a new release that includes the change it made to use tokio 1.

Good catch, opened this PR to get it done https://github.com/awslabs/coldsnap/pull/46

webern:
OK, `coldsnap` and `tough` dependencies are published so this should be unblocked.

webern:
`tools` has been updated to use tokio v1. Bottlerocket (i.e. `sources`) has not because we need `actix-web 4.0`.

webern:
Here are a few useful tracking links
https://github.com/actix/actix-web/milestone/6
https://github.com/actix/actix-web/blob/web-v3.3.2/actix-http/Cargo.toml#L47
https://github.com/actix/actix-net/blob/1.0/actix-rt/Cargo.toml#L23

PR #1370, opened 2021-03-09
tools: update to tokio v1 reqwest v0.11 tough v0.11

webern:
**Issue number:**

Half of #1269

**Description of changes:**

Update the tools workspace to tokio v1

**Testing done:**

- Checked Cargo.lock to make sure only one version of tokio is in use.
- Ran cargo make ami (with multiple regions and accounts in the Infra.toml).
- Ran cargo make repo, ran an instance and checked the repo.
- Ran cargo make ssm(with multiple regions and accounts in the Infra.toml). 

**Terms of contribution:**


tools/Cargo.lock
```diff
@@ -1291,8 +1100,9 @@ dependencies = [
  "tempfile",
  "tinytemplate",
  "tokio",
+ "tokio-stream",
  "toml",
- "tough",
+ "tough 0.11.0",
```
tjkirch, line 1105: It's unfortunate that we need tough 0.10 as well because we link to update_metadata from sources/.  Could that potentially cause a problem?  I think I trust the compiler here, in that we know we're not intermingling different versions of tough types, but as we know from tokio runtimes, it's possible for there to be failure modes that only show up with a bang.  Do you feel OK here?

tools/pubsys/src/aws/ssm/ssm.rs
```diff
@@ -220,7 +219,10 @@ pub(crate) async fn set_parameters(
         // need the region again here.)
         let mut throttled_streams = Vec::new();
         for (_region, request_list) in regional_requests {
-            throttled_streams.push(throttle(request_interval, stream::iter(request_list)));
+            throttled_streams.push(Box::pin(tokio_stream::StreamExt::throttle(
```
tjkirch, line 222: Nice job figuring out pinning.

nit: can we `use` `throttle` like we used to?

tjkirch (approved):
Nice.  (Un?)fortunately tokio 1.3 and reqwest 0.11.2 came out today.  It should be an easy update on top of this..?

tools/Cargo.lock
webern: I took a look at `update_metadata` and I don't think `pubsys` is touching `tough` (or `tokio`) via `update_metadata`. It's kind of a weird crate.

I think the only use of `tough` in `update_metadata` is this function:

```rust
pub fn load_manifest(repository: &tough::Repository) -> Result<Manifest> {
    let target = "manifest.json";
    serde_json::from_reader(
        repository
            .read_target(target)
            .context(error::ManifestLoad)?
            .context(error::ManifestNotFound)?,
    )
    .context(error::ManifestParse)
}
```

That should probably be pasted into `updog` and `migrator` to eliminate the `tough` dependency from `update_metadata`. (I'm the one that must have created this dependency when working on signed migrations).

tools/pubsys/src/aws/ssm/ssm.rs
webern: I tried again, and we can't. The best we could do is this:

```rust
use tokio_stream::StreamExt as TokioStreamExt;
```

```rust
        for (_region, request_list) in regional_requests {
            throttled_streams.push(Box::pin(TokioStreamExt::throttle(
                stream::iter(request_list),
                request_interval,
            )));
        }
```

I think what I have is better:

```rust
        for (_region, request_list) in regional_requests {
            throttled_streams.push(Box::pin(tokio_stream::StreamExt::throttle(
                stream::iter(request_list),
                request_interval,
            )));
        }
```

webern:
> webern force-pushed the webern:tools-tokio-v1 branch from 5427ac7 to 18e9db3

Update tokio and reqwest to the latest versions.

webern:
> webern force-pushed the webern:tools-tokio-v1 branch from 18e9db3 to a7f0037

Rebase

tjkirch (approved):

etungsten (approved):

tools/buildsys/Cargo.toml
```diff
@@ -12,7 +12,7 @@ exclude = ["README.md"]
 duct = "0.13.0"
 hex = "0.4.0"
 rand = { version = "0.8", default-features = false, features = ["std", "std_rng"] }
-reqwest = { version = "0.10.1", default-features = false, features = ["rustls-tls", "blocking"] }
+reqwest = { version = "0.11.1", default-features = false, features = ["rustls-tls", "blocking"] }
 serde = { version = "1.0", features = ["derive"] }
```
srgothi92, line 16: even `0.11.2` is available, are we not updating directly to that ? Or you are ignoring here because minor version change does not matter and `Cargo.lock` will include `0.11.2`

Any reason why we are not removing minor version and change it to "0.11" for all new dependency updates ? We know minor version any way does not matter.

tools/pubsys-setup/Cargo.toml
```diff
@@ -10,7 +10,7 @@ publish = false
 hex = "0.4.0"
 log = "0.4"
 pubsys-config = { path = "../pubsys-config/" }
-reqwest = { version = "0.10.1", default-features = false, features = ["rustls-tls", "blocking"] }
+reqwest = { version = "0.11.1", default-features = false, features = ["rustls-tls", "blocking"] }
```
srgothi92, line 13: same reqwest version comment here.

tools/buildsys/Cargo.toml
webern: we are using 0.11.2. 0.11.1 is the minimum acceptable version

tjkirch: If the micro version is specified, it can generally be taken to mean something like "this is the minimum version we've tested with, so you shouldn't use 0.11.0 because we can't guarantee it'll work, but later versions probably will work."

srgothi92 (approved):

PR #1372, opened 2021-03-09
Add imdsclient library for querying IMDS

jpculp:
**Issue number:**

N/A

**Description of changes:**

This adds an imdsclient library for tools and setting generators to leverage for fetching user-data, meta-data, and dynamic data from the AWS Instance Metadata Service.

`shibaken` was also updated by removing the stand-alone IMDS logic in favor of the shared imdsclient library.
Some of the log messages in `shibaken` were also tweaked for more clarity.

`pluto` and `early-boot-config` have also been updated by removing the stand-alone IMDS logic in favor of the shared imdsclient library.

**Packages that need to be updated:**
- [x] shibaken
- [x] pluto
- [x] early-boot-config

**Issues:**
- [x] Add testing (https://github.com/bottlerocket-os/bottlerocket/issues/1398)

**Testing done:**

- Built `aws-k8s-1.19` ami and launched instance.

Also built and launched `vmware-k8s-1.20` (x86_64) and `aws-ecs-1` (aarch64).
Verified that they booted correctly and connected to their respective clusters.

**Terms of contribution:**

### DIFF
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -857,14 +857,15 @@ name = "early-boot-config"
 version = "0.1.0"
 dependencies = [
  "apiclient",
+ "async-trait",
  "base64",
  "cargo-readme",
  "flate2",
  "hex-literal",
  "http",
+ "imdsclient",
  "lazy_static",
  "log",
- "reqwest",
  "serde",
  "serde-xml-rs",
  "serde_json",
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -2781,12 +2781,13 @@ version = "0.1.0"
 dependencies = [
  "base64",
  "cargo-readme",
+ "imdsclient",
  "log",
- "reqwest",
  "serde",
  "serde_json",
  "simplelog",
  "snafu",
+ "tokio",
 ]
 
 [[package]]
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -329,6 +329,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "781f336cc9826dbaddb9754cb5db61e64cab4f69668bd19dcc4a0394a86f4cb1"
 
 [[package]]
+name = "async-stream"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0a26cb53174ddd320edfff199a853f93d571f48eeb4dde75e67a9a3dbb7b7e5e"
+dependencies = [
+ "async-stream-impl",
+ "futures-core",
+]
+
+[[package]]
+name = "async-stream-impl"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "db134ba52475c060f3329a8ef0f8786d6b872ed01515d4b79c162e5798da1340"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "async-trait"
 version = "0.1.42"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1397,6 +1418,24 @@ dependencies = [
  "matches",
  "unicode-bidi",
  "unicode-normalization",
+]
+
+[[package]]
+name = "imdsclient"
+version = "0.1.0"
+dependencies = [
+ "cargo-readme",
+ "http",
+ "httptest",
+ "log",
+ "reqwest",
+ "serde",
+ "serde_json",
+ "simplelog",
+ "snafu",
+ "tokio",
+ "tokio-test",
+ "url",
 ]
 
 [[package]]
@@ -3249,6 +3288,30 @@ dependencies = [
  "rustls",
  "tokio",
  "webpki",
+]
+
+[[package]]
+name = "tokio-stream"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e177a5d8c3bf36de9ebe6d58537d8879e964332f93fb3339e43f618c81361af0"
+dependencies = [
+ "futures-core",
+ "pin-project-lite",
+ "tokio",
+]
+
+[[package]]
+name = "tokio-test"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f58403903e94d4bc56805e46597fced893410b2e753e229d3f7f22423ea03f67"
+dependencies = [
+ "async-stream",
+ "bytes",
+ "futures-core",
+ "tokio",
+ "tokio-stream",
 ]
 
 [[package]]
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -2144,8 +2144,8 @@ version = "0.1.0"
 dependencies = [
  "apiclient",
  "cargo-readme",
+ "imdsclient",
  "models",
- "reqwest",
  "rusoto_core",
  "rusoto_eks",
  "serde_json",
--- a/sources/api/pluto/src/main.rs
+++ b/sources/api/pluto/src/main.rs
@@ -36,7 +36,7 @@ reasonable default is available.
 mod api;
 mod eks;
 
-use reqwest::Client;
+use imdsclient::ImdsClient;
 use snafu::{ensure, OptionExt, ResultExt};
 use std::fs::File;
 use std::io::{BufRead, BufReader};
@@ -48,32 +48,11 @@ const DEFAULT_DNS_CLUSTER_IP: &str = "10.100.0.10";
 // If our CIDR block begins with "10." this is our DNS.
 const DEFAULT_10_RANGE_DNS_CLUSTER_IP: &str = "172.20.0.10";
 
-// Instance Meta Data Service
-const IMDS_BASE_URL: &str = "http://169.254.169.254/2018-09-24";
-// Currently only able to get fetch session tokens from `latest`
-// FIXME Pin to a date version that supports IMDSv2 once such a date version is available.
-const IMDS_SESSION_TOKEN_ENDPOINT: &str = "http://169.254.169.254/latest/api/token";
-const IMDS_NODE_IPV4_ENDPOINT: &str = "http://169.254.169.254/2018-09-24/meta-data/local-ipv4";
-const IMDS_MAC_ENDPOINT: &str =
-    "http://169.254.169.254/2018-09-24/meta-data/network/interfaces/macs";
-const IMDS_INSTANCE_TYPE_ENDPOINT: &str =
-    "http://169.254.169.254/2018-09-24/meta-data/instance-type";
-
 const ENI_MAX_PODS_PATH: &str = "/usr/share/eks/eni-max-pods";
 
 mod error {
     use crate::eks;
     use snafu::Snafu;
-
-    // Taken from sundog.
-    fn code(source: &reqwest::Error) -> String {
-        source
-            .status()
-            .as_ref()
-            .map(|i| i.as_str())
-            .unwrap_or("Unknown")
-            .to_string()
-    }
 
     #[derive(Debug, Snafu)]
     #[snafu(visibility = "pub(super)")]
@@ -81,18 +60,14 @@ mod error {
         #[snafu(display("Unable to parse CIDR '{}': {}", cidr, reason))]
         CidrParse { cidr: String, reason: String },
 
-        #[snafu(display("Error {}ing '{}': {}", method, uri, source))]
-        ImdsRequest {
-            method: String,
-            uri: String,
-            source: reqwest::Error,
-        },
+        #[snafu(display("IMDS request failed: {}", source))]
+        ImdsRequest { source: imdsclient::Error },
 
-        #[snafu(display("Error '{}' from '{}': {}", code(&source), uri, source))]
-        ImdsResponse { uri: String, source: reqwest::Error },
+        #[snafu(display("IMDS client failed: {}", source))]
+        ImdsClient { source: imdsclient::Error },
 
-        #[snafu(display("Error getting text response from {}: {}", uri, source))]
-        ImdsText { uri: String, source: reqwest::Error },
+        #[snafu(display("IMDS request failed: No '{}' found", what))]
+        ImdsNone { what: String },
 
         #[snafu(display("Error deserializing response into JSON from {}: {}", uri, source))]
         ImdsJson {
@@ -109,12 +84,6 @@ mod error {
             output: String,
             source: serde_json::error::Error,
         },
-
-        #[snafu(display("Missing MAC address from IMDS: {}", uri))]
-        MissingMac { uri: String },
-
-        #[snafu(display("Invalid machine architecture, not one of 'x86_64' or 'aarch64'"))]
-        UnknownArchitecture,
 
         #[snafu(display("{}", source))]
         EksError { source: eks::Error },
@@ -146,23 +115,14 @@ use error::PlutoError;
 
 type Result<T> = std::result::Result<T, PlutoError>;
 
-async fn get_text_from_imds(client: &Client, uri: &str, session_token: &str) -> Result<String> {
-    client
-        .get(uri)
-        .header("X-aws-ec2-metadata-token", session_token)
-        .send()
+async fn get_max_pods(client: &mut ImdsClient) -> Result<String> {
+    let instance_type = client
+        .fetch_identity_document()
         .await
-        .context(error::ImdsRequest { method: "GET", uri })?
-        .error_for_status()
-        .context(error::ImdsResponse { uri })?
-        .text()
-        .await
-        .context(error::ImdsText { uri })
-}
+        .context(error::ImdsRequest)?
+        .instance_type()
+        .to_string();
 
-async fn get_max_pods(client: &Client, session_token: &str) -> Result<String> {
-    let instance_type =
-        get_text_from_imds(&client, IMDS_INSTANCE_TYPE_ENDPOINT, session_token).await?;
     // Find the corresponding maximum number of pods supported by this instance type
     let file = BufReader::new(
         File::open(ENI_MAX_PODS_PATH).context(error::EniMaxPodsFile {
@@ -187,7 +147,7 @@ async fn get_max_pods(client: &Client, session_token: &str) -> Result<String> {
 /// the `serviceIPv4CIDR`. If that works, it returns the expected cluster DNS IP address which is
 /// obtained by substituting `10` for the last octet. If the EKS call is not successful, it falls
 /// back to using IMDS MAC CIDR blocks to return one of two default addresses.
-async fn get_cluster_dns_ip(client: &Client, session_token: &str) -> Result<String> {
+async fn get_cluster_dns_ip(client: &mut ImdsClient) -> Result<String> {
     // try calling eks describe-cluster to figure out the dns cluster ip
     if let Some(dns_ip) = get_dns_from_eks().await {
         // we were able to calculate the dns ip from the cidr range we received from eks
@@ -196,7 +156,7 @@ async fn get_cluster_dns_ip(client: &Client, session_token: &str) -> Result<Stri
 
     // we were unable to obtain or parse the cidr range from eks, fallback to one of two default
     // values based on the cidr range of our primary network interface
-    get_cluster_dns_from_imds_mac(client, session_token).await
+    get_cluster_dns_from_imds_mac(client).await
 }
 
 /// Gets the Service IPV4 CIDR setting from EKS and parses it to calculate the cluster DNS IP.
@@ -242,20 +202,31 @@ fn get_dns_from_cidr(cidr: &str) -> Result<String> {
 
 /// Gets gets the the first VPC IPV4 CIDR block from IMDS. If it starts with `10`, returns
 /// `10.100.0.10`, otherwise returns `172.20.0.10`
-async fn get_cluster_dns_from_imds_mac(client: &Client, session_token: &str) -> Result<String> {
-    let uri = IMDS_MAC_ENDPOINT;
-    let macs = get_text_from_imds(&client, uri, session_token).await?;
-    // Take the first (primary) MAC address. Others will exist from attached ENIs.
-    let mac = macs.split('\n').next().context(error::MissingMac { uri })?;
+async fn get_cluster_dns_from_imds_mac(client: &mut ImdsClient) -> Result<String> {
+    // Take the first (primary) MAC address. Others may exist from attached ENIs.
+    let mac = client
+        .fetch_mac_addresses()
+        .await
+        .context(error::ImdsRequest)?
+        .first()
+        .context(error::ImdsNone {
+            what: "mac addresses",
+        })?
+        .clone();
 
-    // Infer the cluster DNS based on our CIDR blocks.
-    let mac_cidr_blocks_uri = format!(
-        "{}/meta-data/network/interfaces/macs/{}/vpc-ipv4-cidr-blocks",
-        IMDS_BASE_URL, mac
-    );
-    let mac_cidr_blocks = get_text_from_imds(&client, &mac_cidr_blocks_uri, session_token).await?;
+    // Take the first CIDR block for the primary MAC.
+    let cidr_block = client
+        .fetch_cidr_blocks_for_mac(&mac)
+        .await
+        .context(error::ImdsRequest)?
+        .first()
+        .context(error::ImdsNone {
+            what: "CIDR blocks",
+        })?
+        .clone();
 
-    let dns = if mac_cidr_blocks.starts_with("10.") {
+    // Infer the cluster DNS based on the CIDR block.
+    let dns = if cidr_block.starts_with("10.") {
         DEFAULT_10_RANGE_DNS_CLUSTER_IP
     } else {
         DEFAULT_DNS_CLUSTER_IP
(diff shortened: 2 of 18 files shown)


jpculp:
Seems like the examples in the Doc strings aren't tight enough to pass `Doc-tests`. Will post another revision.

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /// Helper to fetch `dynamic` targets from IMDS, preferring an override file if present.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use imdsclient::ImdsClient;
+    /// # let imdsclient=ImdsClient::new().unwrap();
+    /// # let end_target="";
```
jpculp: Would it improve readability to add some example values to the hidden pieces of the doc string?

sources/api/shibaken/src/main.rs
```diff
@@ -191,7 +158,7 @@ fn parse_args(args: env::Args) -> Result<Args> {
     }
 
     Ok(Args {
-        log_level: log_level.unwrap_or_else(|| LevelFilter::Info),
+        log_level: log_level.unwrap_or(LevelFilter::Info),
```
jpculp, line 94: Since `sundog` only prints `stderr` in the event of an error, would it make sense for me to change default log level to Debug? It would make it easier to troubleshoot, but might be somewhat ugly if someone ran `shibaken` in a shell.

sources/imdsclient/build.rs
```diff
@@ -0,0 +1,45 @@
...
+        false, // add license
+        true,  // indent headings
+    )
+    .unwrap();
+
+    let mut readme = File::create("README.md").unwrap();
+    readme.write_all(content.as_bytes()).unwrap();
+}
+
+fn generate_constants() {
```
zmrow: Can we add a comment here explaining what this is for?

sources/imdsclient/README.md
```diff
@@ -0,0 +1,13 @@
+# imdsclient
+
+Current version: 0.1.0
+
+The imdsclient library provides high-level methods to interact with the AWS Instance Metadata Service.
+See the documentation for submodules [`fetch_dynamic`], [`fetch_metadata`], and [`fetch_userdata`] for high-level helpers.
```
zmrow: These aren't really submodules, just methods. :)  Probably safe to just say "methods".

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+
+#![deny(rust_2018_idioms)]
+
+use http::StatusCode;
+use log::{debug, info, trace};
+use reqwest::blocking::Client;
+use snafu::ResultExt;
+
+const IMDS_BASE_URI: &str = "http://169.254.169.254";
+const IMDS_PINNED_DATE: &str = "2020-10-27";
```
zmrow: I'm assuming this is the latest date available as of writing?

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+            session_token,
+        })
+    }
+
+    /// Helper to fetch `dynamic` targets from IMDS, preferring an override file if present.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use imdsclient::ImdsClient;
```
zmrow: I'm assuming these doctests will fail if you try to build on anything but an EC2 instance?

If that's the case, let's just move them to the docstring at the top of the lib.

sources/imdsclient/src/lib.rs
zmrow: If we end up moving these examples to the docstring, I would vote we use some "real" values to make them more applicable and run-able by a user.

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+const IMDS_PINNED_DATE: &str = "2020-10-27";
+
+// Currently only able to get fetch session tokens from `latest`
+const IMDS_SESSION_TOKEN_URI: &str = "http://169.254.169.254/latest/api/token";
+
+/// A client for making IMDSv2 queries.
+/// It obtains a session token when it is first instantiated and is reused between helper functions.
+pub struct ImdsClient {
+    client: Client,
+    session_token: String,
```
zmrow, line 30: I'm a little wary about storing the token without giving the user the option to specify how long it will live.  A user may try to use the client in a process that lasts longer than 60 seconds.

I think a default value of 60 seconds is fine as it covers most of the use cases we have (short quick programs that make a request or three quickly), but I'd like to have an additional `new..`  and/or `fetch...` method that gives the user the opportunity to define the token TTL.

In Rust-land, this could be implemented as an additional method like `ImdsClient::new_with_token_ttl()`.  Under the covers it could call the `fetch_token_with_ttl()` method.  Doing it this way avoids `Option`s in the method signature, which requires a user to pass a `Some` or `None` (which IMHO in this case is not a very friendly API).

zmrow (commented):
Posting what I have for now, will be back with a few more!

sources/imdsclient/src/lib.rs
webern: They will run as tests even there. You can tell rust not to run examples though https://stackoverflow.com/a/32429385/2779792

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+                    method: "GET",
+                    uri: &uri,
+                    code,
+                })?;
+                trace!("Response text: {:?}", &response_body);
+
+                Ok(Some(String::from(response_body.trim_end())))
+            }
+
+            // IMDS returns 404 if no user data is given, or if IMDS is disabled, for example
```
webern:
```suggestion
            // IMDS returns 404 if no user data is given, or if IMDS is disabled
```

webern (commented):
Can tests be written with a mock server? You would need another constructor (which could be private) that takes the URLs. If this constructor were public, you could also use it in your doc example for a passing doc test. You could then implement `Default` for the real clients to use.

sources/imdsclient/src/lib.rs
jpculp: I'm leaning towards removing the examples altogether. The functions themselves are fairly simple and having the examples be Docs-test compatible feels like it's reducing the code readability outside of an IDE.

sources/imdsclient/build.rs
jpculp: I copied the `build.rs` to be consistent with the `bottlerocket-release` and `parse-datetime` libs, but after looking at it closer and talking with @tjkirch it seems that only `bottlerocket-release` needs this logic and it should be removed from `imdsclient` and `parse-datetime`

sources/imdsclient/src/lib.rs
jpculp: Sure is!

sources/imdsclient/src/lib.rs
jpculp: Would it be preferable to tell rust not to run the examples, or just remove them entirely?

sources/api/shibaken/src/main.rs
tjkirch: Do we not have enough information in error/warn/info to get a sense of the problem when it fails?  It seems like we should be adding information there, not just promoting all of the debug messages, if you think there's not enough to understand the problem.

sources/imdsclient/src/lib.rs
tjkirch: I'd probably remove them; the functions are pretty simple.

sources/imdsclient/src/lib.rs
tjkirch: Or should we implement automatic token refresh?

zmrow: ^ also a good option. :)

sources/imdsclient/src/lib.rs
zmrow: I'm on board with either.  Even though this isn't a public crate, I appreciate examples in crates I work with.  They're nice for newcomers/beginners as well.

sources/imdsclient/README.md
jpculp: I'm going to reword this piece and take out the "submodules" while I'm at it.

jpculp:
- Removed examples from docstrings.
- Made the `ImdsClient` struct mutable.
- Added `refresh_token` function. When receiving a 401 response from IMDS, the token will be refreshed, and the IMDS fetch be re-attempted 1 time.
- Tweaked `README` wording.
- Removed extraneous `generate_constants` from build.rs

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+                        code,
+                        response_body,
+                    }
+                    .fail();
+                }
+            }
+        }
+    }
+
+    pub fn refresh_token(&mut self) -> Result<()> {
```
jpculp: I'll add a docstring to this before I pull this PR out of draft.

sources/api/shibaken/src/main.rs
jpculp: It's kind of an edge-case, but the current IMDS logic in Bottlerocket treats a 404 as an `Ok(None)` and that's what we want 99% of the time. However, in `shibaken` we decided to actually error out if it receives a 404 when it is collecting the data from the individual keys in IMDS. Since the `imdsclient` isn't returning an error on 404, some context is lost.

At an error level `shibaken` would return something like:
```
Error retrieving 'public-keys/1/openssh-key' from IMDS metadata
```
but the debug message would show the full uri like this:
```
(2) reqwest::async_impl::client: response '404 Not Found' for http://169.254.169.254/2020-10-27/meta-data/public-keys/1/openssh-key
```

I can update the error to say 404 for clarity, but I'm not quite sure if there is a simple way to pass the full uri. One option is to re-construct the uri via public constants, but I'm not a fan of that since it's not a guarantee that that was the uri that was `reqwest`'d

tjkirch: I don't think the full URI is super helpful, since there's only one thing it could possibly be, given you know the version you're running.  And this case could seemingly only crop up if EC2 listed erroneous keys, or if the user changed the registered keys after the list was retrieved.  I don't think it's possible to change the keys associated in EC2 while an instance is running, so I don't expect that particular issue to come up.  The error seems fine to me...

jpculp:
- Add docstring to `refresh_token`
- Tweaked `ImdsData` error in `shibaken` to be more visually consistent with `ImdsClient` error.

sources/api/shibaken/src/main.rs
```diff
@@ -45,75 +38,56 @@ impl UserData {
...
 
     info!("Fetching public keys from IMDS");
     let mut public_keys = Vec::new();
-    for uri in public_key_uris {
-        let public_key_text = fetch_from_imds(&client, &uri, &imds_session_token)?
-            .context(error::KeyNotFound { uri })?;
-        let public_key = public_key_text.trim_end();
+    let target_count: u32 = 0;
+    for target in &public_key_targets {
+        let target_count = target_count + 1;
```
zmrow: If I'm understanding this correctly, this variable should match up with the index of the key currently being fetched?  If that's the case, rather than keeping a count yourself you can use the [`enumerate()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate) method on `public_key_targets` to avoid any "off by one" errors.

sources/api/shibaken/src/main.rs
webern: I think the error message `Error retrieving 'public-keys/1/openssh-key' from IMDS metadata` is clear enough. If you really do have a reproducible case where you need to know whether it was a 404 or something else, then you are really in the weeds anyway and it seems reasonable to expect the user/developer to turn on debugging for something that deep.

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// Fetches a new session token and adds it to the current ImdsClient.
+    pub fn refresh_token(&mut self) -> Result<()> {
+        self.session_token = fetch_token(&self.client)?;
+        Ok(())
+    }
+}
+
+/// Helper to fetch an IMDSv2 session token that is valid for 60 seconds.
+pub fn fetch_token(client: &Client) -> Result<String> {
```
webern: Is fetch_token actually part of the public interface for this library? I don't see it being used in the clients. If it's not used in the clients maybe we should make it private.

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+        let client = Client::new();
+        let session_token = fetch_token(&client)?;
+        Ok(Self {
+            client,
+            session_token,
+        })
+    }
+
+    /// Helper to fetch `dynamic` targets from IMDS, preferring an override file if present.
+    pub fn fetch_dynamic(&mut self, end_target: &str, description: &str) -> Result<Option<String>> {
```
webern: Is fetch_dynamic a local helper or something we expect client programs to use?

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+
+                    return Ok(Some(String::from(response_body.trim_end())));
+                }
+
+                // IMDS returns 404 if no user data is given, or if IMDS is disabled
+                StatusCode::NOT_FOUND => return Ok(None),
+
+                // IMDS returns 401 if the session token is expired or invalid
+                StatusCode::UNAUTHORIZED => {
+                    warn!("Session token is invalid or expired");
```
webern: Maybe this is actually `info!` since it's not actually a problem and we are going to do the refresh. Sometimes when you're debugging something hard, warnings can lead you astray.

sources/imdsclient/src/lib.rs
tjkirch: In my mind, ideally, the user doesn't know or care if they're using tokens or IMDSv2 or IMDSv1, unless they ask...

sources/imdsclient/src/lib.rs
jpculp: The only case I know of is that `early-boot-config` needs it to [fetch the region](https://github.com/bottlerocket-os/bottlerocket/blob/bf2a78ab8c8e628d8b2c6365bdce3958b0635d4f/sources/api/early-boot-config/src/provider/aws.rs#L149) for the API.

etungsten (approved):
LGTM assuming all the existing comments are addressed.

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /// Helper to fetch `dynamic` targets from IMDS, preferring an override file if present.
+    pub fn fetch_dynamic(&mut self, end_target: &str, description: &str) -> Result<Option<String>> {
+        let dynamic_target = format!("dynamic/{}", end_target);
+        self.fetch_imds(IMDS_PINNED_DATE, &dynamic_target, description)
+    }
+
+    /// Helper to fetch `meta-data` targets from IMDS, preferring an override file if present.
+    pub fn fetch_metadata(
```
zmrow: I'm a bit worried about this API since we're not handling these strings as URLs and we're not doing any handling of slashes, etc.  Ideally we'd at least try to parse the string as a URL so we at least know it's valid.

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Helper to fetch `user-data` from IMDS, preferring an override file if present.
+    pub fn fetch_userdata(&mut self) -> Result<Option<String>> {
+        self.fetch_imds(IMDS_PINNED_DATE, "user-data", "user-data")
+    }
+
+    /// Fetch data from IMDS, preferring an override file if present.
+    pub fn fetch_imds(
+        &mut self,
+        schema_version: &str,
+        target: &str,
```
zmrow: Any reason these `&str`s shouldn't be `AsRef<str>`?

sources/api/shibaken/src/main.rs
zmrow: But perhaps I'm misunderstanding the purpose here.

sources/imdsclient/src/lib.rs
webern: OK, if it's being used then pub is fine. I was just looking for pub lock-down opportunities.

sources/api/shibaken/src/main.rs
jpculp: This is just for some extra status context, taking advantage of the description variable that wasn't in the original `shibaken` IMDS logic, but is present in `early-boot-config`. This count is unrelated to the index number (for example if index `0` is the only available key then it will be `public key (1/1)`)

The generated log messages that look something like:
`Requesting public key (1/2) from http://169.254.169.254/latest/meta-data/public-keys/0/openssh-key`

sources/imdsclient/src/lib.rs
jpculp: I didn't think about it, but I think it's a good idea to make that change.

sources/imdsclient/src/lib.rs
jpculp: I was going to rely on the `str` -> `Url` parse in `Reqwest`'s `get` function, but it probably makes sense to parse the `uri`  just before the loop.

sources/imdsclient/src/lib.rs
jpculp: It was in the first iteration, but now it's no longer necessary. I'll remove it.

jpculp:
> Can tests be written with a mock server? You would need another constructor (which could be private) that takes the URLs. If this constructor were public, you could also use it in your doc example for a passing doc test. You could then implement `Default` for the real clients to use. -@webern

I've gone ahead and added an issue (https://github.com/bottlerocket-os/bottlerocket/issues/1398) for this so that we can revisit it at a later date.

jpculp:
- Replaced `&str` with `AsRef<str>` for public functions
- Changed `attempt` from `u16` to `u8`
- Created `max_attempts` variable
- Changed session token expired message from `warn` to `info`
- Tweaked a description string in `shibaken` for a clearer log message.

webern (approved):

zmrow (approved):
:key:

jpculp:
Since this is an optimization and there is no rush, I downgraded this to a draft so we can tackle the testing described in [#1398](https://github.com/bottlerocket-os/bottlerocket/issues/1398) and integrate with `pluto` and `early-boot-config` in the same PR.

jpculp:
Rebased on v1.1.0

jpculp:
Added async functionality in preparation for on-boarding `pluto`.

jpculp:
Added a commit on-boarding `pluto`

sources/api/pluto/src/main.rs
```diff
@@ -146,23 +125,22 @@ use error::PlutoError;
...
+        Ok(Some(instance_type)) => instance_type,
+        Ok(None) => {
+            return error::ImdsData {
+                target: instance_type_target,
+                target_type: "meta-data",
+            }
+            .fail()
+        }
+        Err(e) => Err(e).context(error::ImdsRequest)?,
+    };
```
etungsten: I'm pretty sure this can be simplified a bit.
Can you try `let instance_type = imdsclient.fetch_metadata(&instance_type_target, "instance type").context(error::ImdsRequest)?.context(error::ImdsData{target : instance_type_target, target_type: "meta-data"})?;`

Same for all other similar constructs in the PR.

sources/api/pluto/src/main.rs
```diff
@@ -187,7 +165,7 @@ async fn get_max_pods(client: &Client, session_token: &str) -> Result<String> {
 /// the `serviceIPv4CIDR`. If that works, it returns the expected cluster DNS IP address which is
 /// obtained by substituting `10` for the last octet. If the EKS call is not successful, it falls
 /// back to using IMDS MAC CIDR blocks to return one of two default addresses.
-async fn get_cluster_dns_ip(client: &Client, session_token: &str) -> Result<String> {
+async fn get_cluster_dns_ip(imdsclient: &mut ImdsClient) -> Result<String> {
```
etungsten: Why does this have to be a mutable reference? Is it because we might need to refresh the auth token?

jpculp: Correct

sources/api/pluto/src/main.rs
```diff
@@ -146,23 +125,22 @@ use error::PlutoError;
...
-    client
-        .get(uri)
-        .header("X-aws-ec2-metadata-token", session_token)
-        .send()
-        .await
-        .context(error::ImdsRequest { method: "GET", uri })?
-        .error_for_status()
-        .context(error::ImdsResponse { uri })?
-        .text()
+async fn get_max_pods(imdsclient: &mut ImdsClient) -> Result<String> {
```
zmrow: Nit: It's typically good Rust style to split words in variables with underscore: `imds_client`

jpculp: What are your thoughts on the variable just being `client`?

jpculp:
**imdsclient**
- Replaced output type from `String` to `Vec<u8>` in `fetch_imds`, `fetch_userdata`, and `fetch_dynamic` functions.
- Updated `IMDS_PINNED_DATE` to latest date (2021-01-03).
- Added logic from `early-boot-config` to support compression and shorter log output.

**shibaken & pluto**
- Replaced instances of `imdsclient` with `client`.
- Updated match logic based on @etungsten's suggestion.

**early-boot-config**
- Added async to aws provider and replaced IMDS logic with shared imdsclient library.

jpculp:
Fixed and tested `vmware` and `aws-ecs` builds.

jpculp:
Added tests to `imdsclient`.

jpculp:
- In collaboration with @webern, moved IMDS functions from `shibaken`, `pluto`, and `early-boot-config`, and replaced them with more abstracted functions like `fetch_mac_addresses`.
- Replaced `fetch_metadata` / `fetch_dynamic` with `fetch_string` / `fetch_bytes`.
- Made `refresh_token` private.
- Removed requirement to always add a `description` value.
- @webern added some additional tests for printable strings and parsing public keys.

sources/api/pluto/src/main.rs
```diff
@@ -48,51 +48,26 @@ const DEFAULT_DNS_CLUSTER_IP: &str = "10.100.0.10";
...
+        ImdsRequest { source: imdsclient::Error },
 
-        #[snafu(display("Error '{}' from '{}': {}", code(&source), uri, source))]
-        ImdsResponse { uri: String, source: reqwest::Error },
+        #[snafu(display("IMDS client failed: {}", source))]
+        ImdsClient { source: imdsclient::Error },
 
-        #[snafu(display("Error getting text response from {}: {}", uri, source))]
-        ImdsText { uri: String, source: reqwest::Error },
+        #[snafu(display("IMDS client failed: Response '404' while fetching '{}'", what))]
```
webern: This is not actually related to 404's anymore.

```suggestion
        #[snafu(display("No '{}' found", what))]
```

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+        #[snafu(display("IMDS fetch failed after {} attempts", attempt))]
+        FailedFetch { attempt: u8 },
+
+        #[snafu(display("IMDS session failed: {}", source))]
+        FailedSession { source: reqwest::Error },
+
+        #[snafu(display("Response was not UTF-8: {}", source))]
+        NonUtf8Response { source: std::string::FromUtf8Error },
+
+        #[snafu(display("404 file not found fetching {}", uri))]
```
webern:
```suggestion
        #[snafu(display("404 file not found fetching '{}'", uri))]
```

webern (approved):
Nice refactor with new test code. 👍

etungsten (approved):

jpculp:
Tweaked the two error messages from @webern's feedback.

webern (approved):

sources/Cargo.lock
```diff
@@ -2742,12 +2782,13 @@ version = "0.1.0"
 dependencies = [
  "base64",
  "cargo-readme",
+ "imdsclient",
```
bcressey, line 2785: This reminds me that we should conditionalize shibaken in `os.spec`, like we do for pluto, since we don't want to run it on non-AWS platforms.

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+            self.imds_base_uri,
+            schema_version.as_ref(),
+            target.as_ref()
+        );
+        debug!("Requesting {}", &uri);
+        let mut attempt: u8 = 1;
+        let max_attempts: u8 = 3;
+        loop {
+            attempt += 1;
+            ensure!(attempt <= max_attempts, error::FailedFetch { attempt });
```
bcressey: Seems like we'll only try twice?

```suggestion
        let mut attempt: u8 = 0;
        let max_attempts: u8 = 3;
        loop {
            attempt += 1;
            ensure!(attempt <= max_attempts, error::FailedFetch { attempt });
```

sources/imdsclient/src/lib.rs
```diff
@@ -0,0 +1,726 @@
...
+                // IMDS returns 404 if no user data is given, or if IMDS is disabled
+                StatusCode::NOT_FOUND => return Err(error::Error::NotFound { uri }),
+
+                // IMDS returns 401 if the session token is expired or invalid
+                StatusCode::UNAUTHORIZED => {
+                    info!("Session token is invalid or expired");
+                    self.refresh_token().await?;
+                    info!("Refreshed session token");
+                    continue;
+                }
```
bcressey, line 233: Do we still want to burn one of our retry attempts in this case?

sources/Cargo.lock
jpculp: That makes sense to me, although I'd want to do that in a separate PR. I'll make an issue so I don't forget. (https://github.com/bottlerocket-os/bottlerocket/issues/1590)

sources/imdsclient/src/lib.rs
webern: ~~I think it's correct `attempt <= max_attempts`~~

edit: Disregard

jpculp:
**imdsclient**
- Changed number of retries.
- Added 100ms delay between retries.
- Added logic to retry on request timeouts (instead of just expired tokens).

webern (approved):

jpculp:
**imdsclient**
- Removed extraneous semicolon.

webern (approved):

bcressey (approved):

PR #1388, opened 2021-03-12
k8s settings: kube-reserved, eviction-hard, cpu-manager-policy, allow-unsafe-sysctls

gthao313:
**My fork draft PR**: https://github.com/gthao313/bottlerocket/pull/1

### **Description of changes:**

- Add new settings `eviction-hard`, `kube-reserved`, `allowed-unsafe-sysctls` to k8s. User can specify this setting in userdata.
- Enable `cpu-manager-policy` to static

- Add migration for `eviction-hard`, `kube-reserved`, `allowed-unsafe-sysctls` 

**example**: 

```
"memory.available" = "10Gi"
"nodefs.inodesFree" = “15%”
[settings.kubernetes.kube-reserved]
cpu = "70Mi"
memory = "1Gi"
ephemeral-storage = "1Gi"
[settings.kubernetes]
allowed-unsafe-sysctls = ["net.core.somaxconn", ...]
```

### **Testing done:**

- Build aws-eks-* images and launched instance.
- Migration test: tested the migration via upgrade and downgrade testing
- Unit tests for new model types `EvictionHardKey`, `KubernetesThresholdValue`, `ReservedResourcesKey`, `KubernetesQuantityValue`, and `CpuManagerPolicy` (all passes)

### Eviction-hard test
***Test step1***:
Go to control container run `apiclient -u /settings` to check if expected setting is there.

***Test step2***:
ssh to admin contianer and look `etc/kubernetes/kubelet/config` to check if expected config is there.
```
evictionHard:
  memory.available: 15%
```
***Test step3*** - test effected/behavior:
set up eviction threshold to a high value for triggering eviction.
```
[evictionHard.memory]
"memory.available" = "99.99%" (high percentage to trigger eviction)
```
a) Go to EKS to check node condition 

MemoryPressure | True | kubelet has insufficient memory available
-- | -- | --

b) In admin contianer, check kubelet log by `journalctl -u kubelet`
```
eviction_manager.go:335] eviction manager: attempting to reclaim memory
eviction_manager.go:346] eviction manager: must evict pod(s) to reclaim memory
```
### kube-reservd test
***Test step1***:
Go to control container run apiclient -u /settings to check if expected setting is there.

***Test step2***:
_NodeAllocatable = NodeCapacity - Kube-reserved - system-reserved - eviction-threshold_
**cpu**
[settings.kubernetes.kube-reserved.cpu]
cpu = 1
Allocatable cpu = 2 - 1 = 1
	
```
EKS Resource allocation:
Name Capacity Allocatable
CPU		2		1
```

**memory**
[settings.kubernetes.kube-reserved.Memory]
memory = 1Gi

```
EKS Resource allocation:
Name    Capacity    Allocatable
Memory	7844780Ki	6693804Ki
```

**ephemeral-storage**
[settings.kubernetes.kube-reserved.ephemeral-storage]
ephemeral-storage = 1Gi

```
Capacity:
  attachable-volumes-aws-ebs:  25
  cpu:                         2
  ephemeral-storage:           20624592Ki
  hugepages-1Gi:               0
  hugepages-2Mi:               0
  memory:                      7930796Ki
  pods:                        29
Allocatable:
  attachable-volumes-aws-ebs:  25
  cpu:                         1
  ephemeral-storage:           19007623956
  hugepages-1Gi:               0
  hugepages-2Mi:               0
  memory:                      7828396Ki
  pods:                        29˜
```
### cpu-manager-policy test
***Test step1***:
Go to host find `cpu_manager_state`
```
bash-5.0# cat /var/lib/kubelet/cpu_manager_state
{"policyName":"static","defaultCpuSet":"0-1","checksum":3945352861}bash-5.0#
```

```
journalctl -u kubelet
cpu_manager.go:173] [cpumanager] starting with static policy
```
### allowed-unsafe-sysctls test
***Test step1***:
Go to control container run apiclient -u /settings to check if expected setting is there.

***Test step2***:
Go to kubelet-config to check if desired result is there.
```
cat eks/kubernetes/kubelet/config

allowed-unsafe-sysctls = ["net.core.somaxconn", ]
```

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,6 +318,13 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.standalone-mode`: Whether to run the kubelet in standalone mode, without connecting to an API server.  Defaults to `false`.
 * `settings.kubernetes.authentication-mode`: Which authentication method the kubelet should use to connect to the API server, and for incoming requests.  Defaults to `aws` for AWS variants, and `tls` for other variants.
 * `settings.kubernetes.bootstrap-token`: The token to use for [TLS bootstrapping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/).  This is only used with the `tls` authentication mode, and is otherwise ignored.
+* `settings.kubernetes.eviction-hard`: The signals and thresholds that trigger pod eviction.
+  Remember to quote signals (since they all contain ".") and to quote all values.
+  * Example user data for setting up eviction hard:
+    ```
+    [settings.kubernetes.eviction-hard]
+    "memory.available" = "15%"
+    ```
 
 You can also optionally specify static pods for your node with the following settings.
 Static pods can be particularly useful when running in standalone mode.
--- a/README.md
+++ b/README.md
@@ -325,6 +325,11 @@ The following settings are optional and allow you to further configure your clus
     [settings.kubernetes.eviction-hard]
     "memory.available" = "15%"
     ```
+* `settings.kubernetes.allowed-unsafe-sysctls`: Enables specified list of unsafe sysctls.
+  * Example user data for setting up allowed unsafe sysctls:
+    ```
+    allowed-unsafe-sysctls = ["net.core.somaxconn", "net.ipv4.ip_local_port_range"]
+    ```
 
 You can also optionally specify static pods for your node with the following settings.
 Static pods can be particularly useful when running in standalone mode.
--- a/README.md
+++ b/README.md
@@ -336,6 +336,11 @@ The following settings are set for you automatically by [pluto](sources/api/) ba
 * `settings.kubernetes.cluster-dns-ip`: The CIDR block of the primary network interface.
 * `settings.kubernetes.node-ip`: The IPv4 address of this node.
 * `settings.kubernetes.pod-infra-container-image`: The URI of the "pause" container.
+* `settings.kubernetes.kube-reserved`: Resources reserved for node components.
+  * Bottlerocket provides default values for the resources by [schnauzer](sources/api/):
+    * `cpu`: in millicores from the total number of vCPUs available on the instance.
+    * `memory`: in mebibytes from the max num of pods on the instance. `memory_to_reserve = max_num_pods * 11 + 255`.
+    * `ephemeral-storage`: defaults to `1Gi`.
 
 #### Amazon ECS settings
 
--- a/Release.toml
+++ b/Release.toml
@@ -30,3 +30,7 @@ version = "1.0.7"
     "migrate_v1.0.6_admin-container-v0-6-0.lz4",
 ]
 "(1.0.6, 1.0.7)" = []
+"(1.0.7, 1.0.8)" = [
+    "migrate_v1.0.8_kubelet-eviction-hard.lz4",
+    "migrate_v1.0.8_kubelet-unsafe-sysctl-kube-reserved.lz4",
+]
--- a/packages/kubernetes-1.15/kubelet-config
+++ b/packages/kubernetes-1.15/kubelet-config
@@ -35,6 +35,10 @@ evictionHard:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
+kubeReserved:
+  cpu: "{{kube_reserve_cpu settings.kubernetes.kube-reserved.cpu}}"
+  memory: "{{kube_reserve_memory settings.kubernetes.max-pods settings.kubernetes.kube-reserved.memory}}"
+  ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 cgroupDriver: systemd
--- a/packages/kubernetes-1.15/kubelet-config
+++ b/packages/kubernetes-1.15/kubelet-config
@@ -29,6 +29,12 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 clusterDNS:
 - {{settings.kubernetes.cluster-dns-ip}}
+{{~#if settings.kubernetes.eviction-hard}}
+evictionHard:
+  {{~#each settings.kubernetes.eviction-hard}}
+  {{@key}}: "{{this}}"
+  {{~/each}}
+{{~/if}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 cgroupDriver: systemd
--- a/packages/kubernetes-1.15/kubelet-config
+++ b/packages/kubernetes-1.15/kubelet-config
@@ -35,6 +35,9 @@ evictionHard:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
+{{~#if settings.kubernetes.allowed-unsafe-sysctls}}
+allowedUnsafeSysctls: {{settings.kubernetes.allowed-unsafe-sysctls}}
+{{~/if}}
 kubeReserved:
   cpu: "{{kube_reserve_cpu settings.kubernetes.kube-reserved.cpu}}"
   memory: "{{kube_reserve_memory settings.kubernetes.max-pods settings.kubernetes.kube-reserved.memory}}"
--- a/packages/kubernetes-1.15/kubelet-config
+++ b/packages/kubernetes-1.15/kubelet-config
@@ -39,6 +39,7 @@ kubeReserved:
   cpu: "{{kube_reserve_cpu settings.kubernetes.kube-reserved.cpu}}"
   memory: "{{kube_reserve_memory settings.kubernetes.max-pods settings.kubernetes.kube-reserved.memory}}"
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
+cpuManagerPolicy: "static"
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 cgroupDriver: systemd
--- a/packages/kubernetes-1.19/kubelet-config
+++ b/packages/kubernetes-1.19/kubelet-config
@@ -35,6 +35,10 @@ evictionHard:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
+kubeReserved:
+  cpu: "{{kube_reserve_cpu settings.kubernetes.kube-reserved.cpu}}"
+  memory: "{{kube_reserve_memory settings.kubernetes.max-pods settings.kubernetes.kube-reserved.memory}}"
+  ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 cgroupDriver: systemd
--- a/packages/kubernetes-1.19/kubelet-config
+++ b/packages/kubernetes-1.19/kubelet-config
@@ -29,6 +29,12 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 clusterDNS:
 - {{settings.kubernetes.cluster-dns-ip}}
+{{~#if settings.kubernetes.eviction-hard}}
+evictionHard:
+  {{~#each settings.kubernetes.eviction-hard}}
+  {{@key}}: "{{this}}"
+  {{~/each}}
+{{~/if}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 cgroupDriver: systemd
--- a/packages/kubernetes-1.19/kubelet-config
+++ b/packages/kubernetes-1.19/kubelet-config
@@ -35,6 +35,9 @@ evictionHard:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
+{{~#if settings.kubernetes.allowed-unsafe-sysctls}}
+allowedUnsafeSysctls: {{settings.kubernetes.allowed-unsafe-sysctls}}
+{{~/if}}
 kubeReserved:
   cpu: "{{kube_reserve_cpu settings.kubernetes.kube-reserved.cpu}}"
   memory: "{{kube_reserve_memory settings.kubernetes.max-pods settings.kubernetes.kube-reserved.memory}}"
--- a/packages/kubernetes-1.19/kubelet-config
+++ b/packages/kubernetes-1.19/kubelet-config
@@ -39,6 +39,7 @@ kubeReserved:
   cpu: "{{kube_reserve_cpu settings.kubernetes.kube-reserved.cpu}}"
   memory: "{{kube_reserve_memory settings.kubernetes.max-pods settings.kubernetes.kube-reserved.memory}}"
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
+cpuManagerPolicy: "static"
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 cgroupDriver: systemd
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -46,6 +46,8 @@ members = [
     "api/migration/migrations/v1.0.6/control-container-v0-4-2",
     "api/migration/migrations/v1.0.6/add-shibaken",
     "api/migration/migrations/v1.0.6/admin-container-v0-6-0",
+    "api/migration/migrations/v1.0.8/kubelet-eviction-hard",
+    "api/migration/migrations/v1.0.8/kubelet-unsafe-sysctl-kube-reserved",
 
     "bottlerocket-release",
 
--- a/sources/api/schnauzer/src/helpers.rs
+++ b/sources/api/schnauzer/src/helpers.rs
@@ -4,10 +4,12 @@
 
 use handlebars::{Context, Handlebars, Helper, Output, RenderContext, RenderError};
 use lazy_static::lazy_static;
+use num_cpus;
 use serde_json::value::Value;
 use snafu::{OptionExt, ResultExt};
 use std::borrow::Borrow;
 use std::collections::HashMap;
+use std::convert::TryFrom;
 use url::Url;
 
 lazy_static! {
@@ -43,6 +45,19 @@ lazy_static! {
 /// containers from here.
 const ECR_FALLBACK_REGION: &str = "us-east-1";
 const ECR_FALLBACK_REGISTRY: &str = "328549459982";
+
+/// The amount of CPU to reserve
+/// We are using these CPU ranges from GKE
+/// (https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#node_allocatable):
+/// 6% of the first core
+/// 1% of the next core (up to 2 cores)
+/// 0.5% of the next 2 cores (up to 4 cores)
+/// 0.25% of any cores above 4 cores
+const KUBE_RESERVE_1_CORE: f32 = 60.0;
+const KUBE_RESERVE_2_CORES: f32 = KUBE_RESERVE_1_CORE + 10.0;
+const KUBE_RESERVE_3_CORES: f32 = KUBE_RESERVE_2_CORES + 5.0;
+const KUBE_RESERVE_4_CORES: f32 = KUBE_RESERVE_3_CORES + 5.0;
+const KUBE_RESERVE_ADDITIONAL: f32 = 2.5;
 
 /// Potential errors during helper execution
 mod error {
@@ -160,6 +175,19 @@ mod error {
 
         #[snafu(display("URL '{}' is missing host component", url_str))]
         UrlHost { url_str: String },
+
+        #[snafu(display("Failed to convert {} {} to {}", what, number, target))]
+        ConvertNumber {
+            what: String,
+            number: String,
+            target: String,
+        },
+
+        #[snafu(display("Failed to convert usize {} to u16: {}", number, source))]
+        ConvertUsizeToU16 {
+            number: usize,
+            source: std::num::TryFromIntError,
+        },
     }
 
     // Handlebars helpers are required to return a RenderError.
@@ -170,6 +198,8 @@ mod error {
         }
     }
 }
+
+use error::TemplateHelperError;
 
 /// `base64_decode` decodes base64 encoded text at template render time.
 /// It takes a single variable as a parameter: {{base64_decode var}}
@@ -592,6 +622,120 @@ pub fn join_array(
     Ok(())
 }
 
+/// kube_reserve_memory and kube_reserve_cpu are taken from EKS' calculations.
+/// https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh
+
+/// Calculates the amount of memory to reserve for kubeReserved in mebibytes.
+/// Formula: memory_to_reserve = max_num_pods * 11 + 255 is taken from
+/// https://github.com/awslabs/amazon-eks-ami/pull/419#issuecomment-609985305
+pub fn kube_reserve_memory(
+    helper: &Helper<'_, '_>,
+    _: &Handlebars,
+    _: &Context,
+    renderctx: &mut RenderContext<'_, '_>,
+    out: &mut dyn Output,
+) -> Result<(), RenderError> {
+    trace!("Starting kube_reserve_memory helper");
+    let template_name = template_name(renderctx);
+    trace!("Template name: {}", &template_name);
+
+    trace!("Number of params: {}", helper.params().len());
+    check_param_count(helper, template_name, 2)?;
+
+    let max_num_pods_val = get_param(helper, 0)?;
+    let max_num_pods = match max_num_pods_val {
+        Value::Number(n) => n,
+
+        _ => {
+            return Err(RenderError::from(
+                error::TemplateHelperError::InvalidTemplateValue {
+                    expected: "number",
+                    value: max_num_pods_val.to_owned(),
+                    template: template_name.to_owned(),
+                },
+            ))
+        }
+    };
+    let max_num_pods = max_num_pods
+        .as_u64()
+        .with_context(|| error::ConvertNumber {
+            what: "number of pods",
+            number: max_num_pods.to_string(),
+            target: "u64",
+        })?;
+
+    // Calculates the amount of memory to reserve
+    let memory_to_reserve_value = get_param(helper, 1)?;
+    let memory_to_reserve = match memory_to_reserve_value {
+        Value::Number(n) => n.to_string(),
+        Value::String(s) => s.to_string(),
+        // If no value is set, use the given default.
+        Value::Null => { format!("{}Mi", (max_num_pods * 11 + 255).to_string()) }.to_string(),
+        // composite types unsupported
+        _ => {
+            return Err(RenderError::from(
+                error::TemplateHelperError::InvalidTemplateValue {
+                    expected: "scalar",
+                    value: memory_to_reserve_value.to_owned(),
+                    template: template_name.to_owned(),
+                },
+            ))
+        }
+    };
+
+    // write it to the template
+    out.write(&memory_to_reserve)
+        .with_context(|| error::TemplateWrite {
+            template: template_name.to_owned(),
+        })?;
+
+    Ok(())
+}
+
+/// Get the amount of CPU to reserve for kubeReserved in millicores
+pub fn kube_reserve_cpu(
+    helper: &Helper<'_, '_>,
+    _: &Handlebars,
+    _: &Context,
+    renderctx: &mut RenderContext<'_, '_>,
+    out: &mut dyn Output,
+) -> Result<(), RenderError> {
+    trace!("Starting kube_reserve_cpu helper");
+    let template_name = template_name(renderctx);
+    trace!("Template name: {}", &template_name);
+
+    trace!("Number of params: {}", helper.params().len());
+    check_param_count(helper, template_name, 1)?;
+
+    // Calculates the amount of CPU to reserve
+    let num_cores = num_cpus::get();
+    let cpu_to_reserve_value = get_param(helper, 0)?;
+    let cpu_to_reserve = match cpu_to_reserve_value {
+        Value::Number(n) => n.to_string(),
+        Value::String(s) => s.to_string(),
+        // If no value is set, use the given default.
+        Value::Null => kube_cpu_helper(num_cores)?,
+        // composite types unsupported
+        _ => {
+            return Err(RenderError::from(
+                error::TemplateHelperError::InvalidTemplateValue {
+                    expected: "scalar",
+                    value: cpu_to_reserve_value.to_owned(),
+                    template: template_name.to_owned(),
+                },
+            ))
+        }
+    };
+
+    // write it to the template
+    out.write(&cpu_to_reserve)
+        .with_context(|| error::TemplateWrite {
+            template: template_name.to_owned(),
+        })?;
+
+    Ok(())
+}
+
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 // helpers to the helpers
 
@@ -644,6 +788,24 @@ fn ecr_registry<S: AsRef<str>>(region: S) -> String {
         Some(registry_id) => (region.as_ref(), *registry_id),
     };
     format!("{}.dkr.ecr.{}.amazonaws.com", registry_id, region)
+}
+
+/// Calculates and returns the amount of CPU to reserve
+fn kube_cpu_helper(num_cores: usize) -> Result<String, TemplateHelperError>{
+    let num_cores = u16::try_from(num_cores).context(error::ConvertUsizeToU16 { number: num_cores })?;
+    let millicores_unit = "m";
+    let cpu_to_reserve = match num_cores {
+        0 => 0.0,
+        1 => KUBE_RESERVE_1_CORE,
+        2 => KUBE_RESERVE_2_CORES,
+        3 => KUBE_RESERVE_3_CORES,
+        4 => KUBE_RESERVE_4_CORES,
+        _ => {
+            let num_cores = f32::from(num_cores);
+            KUBE_RESERVE_4_CORES + ((num_cores - 4.0) * KUBE_RESERVE_ADDITIONAL)
+        }
+    };
+    Ok(format!("{}{}", cpu_to_reserve.floor().to_string(), millicores_unit))
 }
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
@@ -1077,5 +1239,123 @@ mod test_join_array {
                 .unwrap();
         let expected = r#""a", "", "c""#;
         assert_eq!(result, expected);
+    }
+}
+
+#[cfg(test)]
+mod test_kube_reserve_memory {
+    use super::*;
+    use handlebars::TemplateRenderError;
+    use serde::Serialize;
+    use serde_json::json;
+
+    // A thin wrapper around the handlebars render_template method that includes
+    // setup and registration of helpers
+    fn setup_and_render_template<T>(tmpl: &str, data: &T) -> Result<String, TemplateRenderError>
+    where
+        T: Serialize,
+    {
+        let mut registry = Handlebars::new();
+        registry.register_helper("kube_reserve_memory", Box::new(kube_reserve_memory));
+
+        registry.render_template(tmpl, data)
+    }
+
+    const TEMPLATE: &str = r#""{{kube_reserve_memory  max-pods kube-reserved-memory}}""#;
+
+    #[test]
+    fn have_settings_1024_mi() {
+        let result = setup_and_render_template(
+            TEMPLATE,
+            &json!({"max-pods": 29, "kube-reserved-memory": "1024Mi"}),
+        )
+        .unwrap();
+        assert_eq!(result, "\"1024Mi\"");
+    }
+
+    #[test]
+    fn no_settings_max_pods_0() {
+        let result =
+            setup_and_render_template(TEMPLATE, &json!({"max-pods": 0, "no-settings": "hi"}))
+                .unwrap();
+        assert_eq!(result, "\"255Mi\"");
+    }
+
+    #[test]
+    fn no_settings_max_pods_29() {
+        let result =
+            setup_and_render_template(TEMPLATE, &json!({"max-pods": 29, "no-settings": "hi"}))
+                .unwrap();
+        assert_eq!(result, "\"574Mi\"");
+    }
+
+    #[test]
+    fn max_pods_not_number() {
+        setup_and_render_template(
+            TEMPLATE,
+            &json!({"settings": {"kubernetes": {"max-pods": "ten"}}}),
+        )
+        .unwrap_err();
+    }
+}
+
+#[cfg(test)]
+mod test_kube_reserve_cpu {
+    use super::*;
+    use handlebars::TemplateRenderError;
+    use serde::Serialize;
+    use serde_json::json;
+
+    // A thin wrapper around the handlebars render_template method that includes
+    // setup and registration of helpers
+    fn setup_and_render_template<T>(tmpl: &str, data: &T) -> Result<String, TemplateRenderError>
+    where
+        T: Serialize,
+    {
+        let mut registry = Handlebars::new();
+        registry.register_helper("kube_reserve_cpu", Box::new(kube_reserve_cpu));
+
+        registry.render_template(tmpl, data)
+    }
+
+    const TEMPLATE: &str = r#"{{kube_reserve_cpu settings.kubernetes.kube-reserved.cpu}}"#;
+
+    #[test]
+    fn kube_reserve_cpu_ok() {
+        setup_and_render_template(TEMPLATE, &json!({"not-the-setting": "hi"})).unwrap();
+        assert!(true);
+    }
+
+    #[test]
+    fn kube_reserve_cpu_30_m() {
+        let result = setup_and_render_template(
+            TEMPLATE,
+            &json!({"settings": {"kubernetes": {"kube-reserved": {"cpu": "30m"}}}}),
+        )
+        .unwrap();
+        assert_eq!(result, "30m");
+    }
+}
+#[cfg(test)]
+mod test_kube_cpu_helper {
+    use crate::helpers::kube_cpu_helper;
+    use std::collections::HashMap;
+
+    #[test]
+    fn kube_cpu_helper_ok() {
+        let mut cpu_reserved: HashMap<usize, &str> = HashMap::new();
+        cpu_reserved.insert(0, "0m");
+        cpu_reserved.insert(1, "60m");
+        cpu_reserved.insert(2, "70m");
+        cpu_reserved.insert(3, "75m");
+        cpu_reserved.insert(4, "80m");
+        cpu_reserved.insert(5, "82m");
+        cpu_reserved.insert(6, "85m");
+        cpu_reserved.insert(47, "187m");
+        cpu_reserved.insert(48, "190m");
+
+        for (num_cpus, expected_millicores) in cpu_reserved.into_iter() {
+            assert_eq!(kube_cpu_helper(num_cpus).unwrap(), expected_millicores);
+        }
     }
 }
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -100,8 +100,9 @@ use std::net::Ipv4Addr;
 use crate::modeled_types::{
     DNSDomain, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue, FriendlyVersion, Identifier,
     KubernetesAuthenticationMode, KubernetesBootstrapToken, KubernetesClusterName,
-    KubernetesLabelKey, KubernetesLabelValue, KubernetesTaintValue,
-    Lockdown, SingleLineString, SysctlKey, Url, ValidBase64, KubernetesEvictionHardKey, KubernetesThresholdValue,
+    KubernetesEvictionHardKey, KubernetesLabelKey, KubernetesLabelValue, KubernetesQuantityValue,
+    KubernetesReservedResourceKey, KubernetesTaintValue, KubernetesThresholdValue, Lockdown,
+    SingleLineString, SysctlKey, Url, ValidBase64,
 };
 
 // Kubernetes static pod manifest settings
@@ -128,6 +129,7 @@ struct KubernetesSettings {
     bootstrap_token: KubernetesBootstrapToken,
     standalone_mode: bool,
     eviction_hard: HashMap<KubernetesEvictionHardKey, KubernetesThresholdValue>,
+    kube_reserved: HashMap<KubernetesReservedResourceKey, KubernetesQuantityValue>,
 
     // Settings where we generate a value based on the runtime environment.  The user can specify a
     // value to override the generated one, but typically would not.
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -101,7 +101,7 @@ use crate::modeled_types::{
     DNSDomain, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue, FriendlyVersion, Identifier,
     KubernetesAuthenticationMode, KubernetesBootstrapToken, KubernetesClusterName,
     KubernetesLabelKey, KubernetesLabelValue, KubernetesTaintValue,
-    Lockdown, SingleLineString, SysctlKey, Url, ValidBase64,
+    Lockdown, SingleLineString, SysctlKey, Url, ValidBase64, KubernetesEvictionHardKey, KubernetesThresholdValue,
 };
 
 // Kubernetes static pod manifest settings
@@ -127,6 +127,7 @@ struct KubernetesSettings {
     authentication_mode: KubernetesAuthenticationMode,
     bootstrap_token: KubernetesBootstrapToken,
     standalone_mode: bool,
+    eviction_hard: HashMap<KubernetesEvictionHardKey, KubernetesThresholdValue>,
 
     // Settings where we generate a value based on the runtime environment.  The user can specify a
     // value to override the generated one, but typically would not.
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -130,6 +130,7 @@ struct KubernetesSettings {
     standalone_mode: bool,
     eviction_hard: HashMap<KubernetesEvictionHardKey, KubernetesThresholdValue>,
     kube_reserved: HashMap<KubernetesReservedResourceKey, KubernetesQuantityValue>,
+    allowed_unsafe_sysctls: Vec<SingleLineString>,
 
     // Settings where we generate a value based on the runtime environment.  The user can specify a
     // value to override the generated one, but typically would not.
--- a/sources/models/src/modeled_types/kubernetes.rs
+++ b/sources/models/src/modeled_types/kubernetes.rs
@@ -2,13 +2,13 @@ use lazy_static::lazy_static;
 use regex::Regex;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 // Just need serde's Error in scope to get its trait methods
+use super::error;
 use serde::de::Error as _;
-use snafu::ensure;
+use snafu::{ensure, ResultExt};
 use std::borrow::Borrow;
 use std::convert::TryFrom;
 use std::fmt;
 use std::ops::Deref;
-use super::error;
 
 /// KubernetesName represents a string that contains a valid Kubernetes resource name.  It stores
 /// the original string and makes it accessible through standard traits.
@@ -365,7 +365,7 @@ impl TryFrom<&str> for KubernetesAuthenticationMode {
 
     fn try_from(input: &str) -> Result<Self, error::Error> {
         ensure!(
-            matches!(input, "aws" | "tls" ),
+            matches!(input, "aws" | "tls"),
             error::InvalidAuthenticationMode { input }
         );
         Ok(KubernetesAuthenticationMode {
@@ -407,8 +407,8 @@ pub struct KubernetesBootstrapToken {
 }
 
 lazy_static! {
-    pub(crate) static ref KUBERNETES_BOOTSTRAP_TOKEN: Regex = Regex::new(
-        r"^[a-z0-9]{6}\.[a-z0-9]{16}$").unwrap();
+    pub(crate) static ref KUBERNETES_BOOTSTRAP_TOKEN: Regex =
+        Regex::new(r"^[a-z0-9]{6}\.[a-z0-9]{16}$").unwrap();
 }
 
 impl TryFrom<&str> for KubernetesBootstrapToken {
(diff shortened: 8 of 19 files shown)


gthao313:
Oush above add new model types `KubernetesThresholdValue ` and `KubernetesQuantityValue`, and those passed unit tests.

gthao313:
push above fix typo

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,278 @@ mod test_kubernetes_bootstrap_token {
...
+    inner: String,
+}
+
+impl TryFrom<&str> for KubernetesQuantityValue {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+            ensure!(
+                KUBERNETES_QUANTITY.is_match(input),
+                error::InvalidKubernetesQuantity { input }
```
etungsten: Instead of making a new error variant for this, let's use the existing `error::Pattern` here.
```suggestion
            error::Pattern {
                thing: "Kubernetes quantity",
                pattern: KUBERNETES_QUANTITY.clone(),
                input
            }
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,278 @@ mod test_kubernetes_bootstrap_token {
...
+lazy_static! {
+    pub(crate) static ref KUBERNETES_QUANTITY: Regex = Regex::new(r"^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$").unwrap();
+}
+
+impl TryFrom<&str> for KubernetesThresholdValue {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        if input.ends_with("%") {
+            let input_f32 = input.replace("%", "").parse::<f32>().context(error::ParseToF32 { input })?;
```
etungsten:
```suggestion
            let input_f32 = input[..input.len()-1].parse::<f32>().context(error::ParseToF32 { input })?;
```

sources/models/src/modeled_types/mod.rs
```diff
@@ -62,6 +62,21 @@ pub mod error {
...
         },
+
+        #[snafu(display("Invalid Kubernetes Quantity '{}'", input))]
+        InvalidKubernetesQuantity { input: String },
+
+        #[snafu(display("Failed to parse {} as f32: {}", input, source))]
+        ParseToF32 {
+            input: String,
+            source: std::num::ParseFloatError,
+         },
```
etungsten: We're only using this for when it's an invalid threshold percentage value. So let's make the error reflect that.
```suggestion
        #[snafu(display("Invalid percentage value '{}': {}", input, source))]
        InvalidPercentage {
            input: String,
            source: std::num::ParseFloatError,
         },
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,278 @@ mod test_kubernetes_bootstrap_token {
...
+    #[test]
+    fn good_kubernetes_threshold_value() {
+        for ok in &[ "10%", "10Mi", "1024M", "1Gi" ] {
+            KubernetesThresholdValue::try_from(*ok).unwrap();
+        }
+    }
+
+    #[test]
+    fn bad_kubernetes_threshold_value() {
+        for err in &["", "anything%", ".bad", "bad.", &"a".repeat(64)] {
```
etungsten: Looks like you need to run `cargo fmt`. There are also some other places that need formatting.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,278 @@ mod test_kubernetes_bootstrap_token {
...
+    fn bad_reserved_resources_key() {
+        for err in &["", "cpa", ".bad", "bad.", &"a".repeat(64)] {
+            ReservedResourcesKey::try_from(*err).unwrap_err();
+        }
+    }
+}
+/// // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesQuantityValue represents a string that contains a valid kubernetes quantity value.
+/// /// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
```
etungsten:
```suggestion
/// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,278 @@ mod test_kubernetes_bootstrap_token {
...
+
+/// KubernetesThresholdValue represents a string that contains a valid kubernetes threshold value.
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesThresholdValue {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref KUBERNETES_QUANTITY: Regex = Regex::new(r"^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$").unwrap();
```
etungsten: Maybe add a comment listing examples of strings this would match against.

gthao313:
Push above improve `error` mod  and syntax.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,288 @@ mod test_kubernetes_bootstrap_token {
...
+/// Available eviction hard signals: https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct EvictionHardKey {
+    inner: String,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum EvitionSignal {
```
```suggestion
enum EvictionSignal {
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,288 @@ mod test_kubernetes_bootstrap_token {
...
+mod test_kubernetes_threshold_value {
+    use super::KubernetesThresholdValue;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn good_kubernetes_threshold_value() {
+        for ok in &["10%", "129e6", "10Mi", "1024M", "1Gi"] {
+            KubernetesThresholdValue::try_from(*ok).unwrap();
+        }
+    }
```
bcressey: This doesn't seem to support operators, e.g. `memory.available<100Mi`.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,288 @@ mod test_kubernetes_bootstrap_token {
...
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        serde_plain::from_str::<ValidCpuManagerPolicy>(&input)
+            .context(error::InvalideCpuManagerPolicy { input })?;
+        Ok(CpuManagerPolicy {
+            inner: input.to_string(),
+        })
+    }
+}
+string_impls_for!(CpuManagerPolicy, "CpuManagerPolicy");
```
bcressey: There was some discussion of always enabling the `static` CPU manager policy. Where did that end up?

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,288 @@ mod test_kubernetes_bootstrap_token {
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesThresholdValue {
+    inner: String,
+}
+
+// Regular expression of Kubernetes quantitiy. i.e. 128974848, 129e6, 129M, 123Mi
+lazy_static! {
+    pub(crate) static ref KUBERNETES_QUANTITY: Regex =
+        Regex::new(r"^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$").unwrap();
+}
```
bcressey: This accepts nonsense like `50KiGTPeG911`.

sources/models/src/modeled_types/kubernetes.rs
gthao313: I make `cpu-manager-policy` as configurable setting, which customers can configure it to `none` and `static`, and  I default `cpu-manager-policy` to `none` in `50-aws-k8s.toml`. 
```
cpu-manager-policy = "none"
```
I'll change it to `static`. Thanks

sources/models/src/modeled_types/kubernetes.rs
gthao313: I take this regex from kubernetes quantity resource, and kubernetes used this regex to validate input value for `eviction-hard`, `kube-reserved` and some settings else. So should I still keep it or change it to more meaningful regex. https://github.com/kubernetes/kubernetes/blob/fcee7a01050652e54d2819b1942533d96e40f455/staging/src/k8s.io/apimachinery/pkg/api/resource/quantity.go#L133-L137

Btw, I had a discussion with Erikson around this validation about if we should be a strict gatekeeper here or if we just need to do a simple validation for that like it has to be a `SingleLineString`.

sources/models/src/modeled_types/kubernetes.rs
gthao313: Yeah. Shailesh left a pretty similar  comment in my fork PR. And my answer was 
" I do not think It's necessary to use format  `[eviction-signal][operator][quantity]` there. Kubernetes used to have flag `--eviction-hard`, and it takes parameters like this `--eviction-hard=memory.available<500Mi,nodefs.available<1Gi,imagefs.available<100Gi`, and right now this flag has been deprecated, `eviction-hard` parameters should be configured though `kublete-config`, and in `kublete-config`the formate is 
```
kind: KubeletConfiguration
evictionHard:
    memory.available:  "200Mi"
```
So I think `"memory.available" = "15%"`  is fine here."

What do you think? :)

sources/api/migration/migrations/v1.1.0/kubelet-unsafe-sysctl-and-cpu-manager-policy/src/main.rs
```diff
@@ -0,0 +1,24 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddSettingsMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+///We added two new settings for configuring kubelet, `kubernetes.allowed-unsafe-sysctls`
```
webern: nit

```suggestion
/// We added two new settings for configuring kubelet, `kubernetes.allowed-unsafe-sysctls`
```

sources/api/pluto/src/main.rs
```diff
@@ -270,11 +271,76 @@ fn pause_container_uri(region: &str) -> Result<String> {
...
+/// Calculates the amount of memory to reserve for kubeReserved in mebibytes.
+/// This function is taken from:
+/// https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh#L228-L239
+fn get_memory_mebibytes_to_reserve(client: &Client, session_token: &str) -> Result<String> {
+    let max_num_pods = get_max_pods(&client, &session_token)?;
+    let mebibytes_unit = "Mi";
+    let max_num_pods_u32 = max_num_pods.parse::<u32>().context(error::ParseToU32 {
+        object: max_num_pods,
+    })?;
+    let memory_to_reserve = max_num_pods_u32 * 11 + 255;
```
webern: Can you explain the magic numbers?

```rust
    let memory_to_reserve = max_num_pods_u32 * 11 + 255;
```

sources/api/pluto/src/main.rs
```diff
@@ -270,11 +271,76 @@ fn pause_container_uri(region: &str) -> Result<String> {
     ))
 }
 
+/// Calculates the amount of memory to reserve for kubeReserved in mebibytes.
+/// This function is taken from:
+/// https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh#L228-L239
```
webern: Do we want this attribution? Same question for the others? Seems like the kind of thing that could someday be meaningless and return a 404.

```rust
/// This function is taken from:
/// https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh#L228-L239
```

sources/api/pluto/src/main.rs
```diff
@@ -270,11 +271,76 @@ fn pause_container_uri(region: &str) -> Result<String> {
...
+        object: max_num_pods,
+    })?;
+    let memory_to_reserve = max_num_pods_u32 * 11 + 255;
+
+    Ok(memory_to_reserve.to_string() + mebibytes_unit)
+}
+
+/// Calculates the amount of CPU to reserve for kubeReserved in millicores
+/// from the total number of vCPUs available on the instance.
+/// We are using these CPU ranges from GKE
```
webern: Does EKS give its own recommendation?

sources/api/pluto/src/main.rs
```diff
@@ -270,11 +271,76 @@ fn pause_container_uri(region: &str) -> Result<String> {
...
+/// (https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#node_allocatable):
+///  6% of the first core
+///  1% of the next core (up to 2 cores)
+///  0.5% of the next 2 cores (up to 4 cores)
+///  0.25% of any cores above 4 cores
+/// This function is taken from:
+/// https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh#L241-L251
+fn get_cpu_millicores_to_reserve() -> Result<String> {
```
webern: These functions: `get_cpu_millicores_to_reserve` and `get_resource_to_reserve_in_range` are kind of hard to follow. I realize we took them from elsewhere, but is it just me, or does the algorithm described in the function's documentation seem simpler than what the function is doing?

The description sounds like this (possibly but with constants for readability)

```rust
match num_cores {
    1 => 600,
    2 => 600 + 100,
    3 => 600 + 100 + 50,
    4 => 600 + 100 + 50 + 50,
    _ => 600 + 100 + 50 + 50 + ((num_cores - 4) * 25),
}
```

Maybe I just don't get it, and maybe that's OK.

sources/models/src/aws-k8s-1.15/defaults.d/50-aws-k8s.toml
```diff
@@ -58,6 +63,8 @@ affected-services = ["kubernetes", "containerd"]
 cluster-domain = "cluster.local"
 standalone-mode = false
 authentication-mode = "aws"
+kube-reserved.ephemeral-storage = "1Gi"
+cpu-manager-policy = "none"
```
webern: Are these values that we were already implicitly defaulting to?

sources/api/pluto/src/main.rs
```diff
@@ -383,3 +451,24 @@ mod test_pause_container_account {
         );
     }
 }
+#[cfg(test)]
+mod test_get_resource_to_reserve_in_range {
+    use crate::get_resource_to_reserve_in_range;
+    #[test]
+    fn get_resource_to_reserve_in_range_ok() {
```
webern: Test more inputs?

```
    fn get_resource_to_reserve_in_range_ok() {
```

sources/models/src/modeled_types/kubernetes.rs
bcressey: Please check with @anguslees about https://github.com/bottlerocket-os/bottlerocket/issues/1140#issuecomment-781848321 - it's not clear to me that we need to expose a setting here, if "static" is good enough as a default.

sources/models/src/modeled_types/kubernetes.rs
bcressey: I'd prefer to do stricter validation, because kicking it out at the API level results in a more immediately actionable error message.

Giving kubelet bad input will either cause it not to start, or it will ignore the setting. That's harder to diagnose because the error or warning will be buried in the journal.

sources/api/pluto/src/main.rs
gthao313: Good catch. I think we do not need that.

sources/api/pluto/src/main.rs
gthao313: No, EKS does not have it own one, and amazon-eks-ami uses GKE idea. https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh

sources/api/pluto/src/main.rs
gthao313: Yeah. I think your idea is more readable. But I want to make it consistent with `get_cpu_millicores_to_reserve ` and `get_resource_to_reserve_in_range ` in amazon-eks-ami, where I got the idea from.  What do you think? still keep it same as in amazon-eks-ami or make it more readable?

sources/models/src/aws-k8s-1.15/defaults.d/50-aws-k8s.toml
gthao313: No. I just add those values for those new settings `kube-reserved` and `cpu-manager-policy`

sources/api/pluto/src/main.rs
gthao313: sure thing. sounds good to me

sources/api/pluto/src/main.rs
gthao313: Those numbers were already set in `amazon-eks-ami`, and from our discussion before, we can use the same strategy to calculate the default of `memory_to_reserve`. So I did not dig too much into the numbers. https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh#L228-L239.

sources/api/pluto/src/main.rs
etungsten: I think we should keep the attribution, if we're sourcing calculations and magic numbers from elsewhere, it's nice to have a direct reference to where they are from. That link should be valid as long as that repo exists. If we're worried about the `aws-eks-ami` repo going away or that commit getting force-pushed over for whatever reason, we should still at least have a comment saying it's from EKS' calculations.

packages/kubernetes-1.15/kubelet-config
```diff
@@ -29,6 +29,20 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 clusterDNS:
 - {{settings.kubernetes.cluster-dns-ip}}
+{{~#if settings.kubernetes.eviction-hard}}
+evictionHard:
+  {{~#each settings.kubernetes.eviction-hard}}
+  {{@key}}: "{{this}}"
+  {{~/each}}
+{{~/if}}
+kubeReserved:
```
zmrow, line 41: Does this key need to be wrapped in the "if" also?

Release.toml
```diff
@@ -29,3 +29,7 @@ version = "1.0.6"
     "migrate_v1.0.6_add-shibaken.lz4",
     "migrate_v1.0.6_admin-container-v0-6-0.lz4",
 ]
+"(1.0.6, 1.1.0)" = [
+    "migrate_v1.1.0_kubelet-kube-reserved-and-eviction-hard.lz4",
```
zmrow: Other than not having a few extra migrations - is there any reason to bundle these settings into a single migration?

sources/api/pluto/src/main.rs
webern: I think it's fine to keep the logic that we have copied from elsewhere. However I would like to see it tested. There's one small change we can make to this function that will make it testable.

Instead of calling `num_cpus::get()` *inside* this function, move it *outside* this function and take num_cpus as an input argument:

```rust
fn get_cpu_millicores_to_reserve(num_cpus: usize) -> Result<String> {
  // ...
}
```

Now write tests for this function given the following inputs
0, 1, 2, 3, 4, 5 and 6.

sources/api/pluto/src/main.rs
webern: Hmmm, if anybody know the answer, it would be a good thing to document.

sources/api/pluto/src/main.rs
webern: Sounds good, resolving.

sources/api/pluto/src/main.rs
webern: My suggestion above to test `get_cpu_millicores_to_reserve` makes tests of `test_get_resource_to_reserve_in_range` less important. But it never hurts...

sources/api/pluto/src/main.rs
gthao313: Yeah. Definitely!

packages/kubernetes-1.15/kubelet-config
gthao313: We defaults values (like `ephemeral-storage = 1Gi`) to `kubeReserved`, so kubeReserved gonna always be there. No need for `if`. 👍

Release.toml
gthao313: The reason why I bundle them is that the settings which are bundled into one migration use same `migration`. For example. `settings.kubernetes.kube-reserved` and `settings.kubernetes.eviction-hard` use `AddPrefixesMigration`. I think that could be much easier to review and maintain without losing functionality.

sources/models/src/modeled_types/kubernetes.rs
gthao313: Yesterday I talked with Angus and we reached an agreement that `cpu-manager-policy` is really enabling a feature.  The default (none) is the legacy behaviour, and `static` turns on the feature. Therefore, I think we just need always enable the static CPU manager policy. I'll update `cpu-manager-policy`.

Release.toml
etungsten: We can't use two different migration helpers in the same migration at the moment (since one `migrate` call will overwrite the other) so they have to be separate migrations.

Edit: I think I must have misunderstood the question. Hiding.

gthao313:
Push above enable `cpu-manager-policy` to `static`, improve input validation for `eviction-hard` and `kube-reserved`, and add more unit test for  `get_resource_to_reserve_in_range` `get_cpu_millicores_to_reserve`.

gthao313:
push above fix migration version.

sources/api/pluto/src/main.rs
webern: I see your tests and it seems to suggest that my interpretation of the algo was correct.

By the way, I actually prefer representing test cases independently, like you did when they looked like this:

```rust
#[test]
    fn get_cpu_millicores_to_reserve_0_ok() {
        let num_cpus: usize = 0;
        assert_eq!(get_cpu_millicores_to_reserve(num_cpus).unwrap(), "0m");
    }
```

Instead of building a map or a vec and looping through test cases. (I realize that test case loops are the Go way, but I never liked it). My reasoning is
- In a loop, you don't run the test cases after the one that failed. Sometimes seeing the complete PASS/FAIL picture is helpful.
- When debugging, you have to comment out previous test cases if you're trying to isolate the failure case.

I can't make you change it though since we definitely have precedent for test loops in the project! :)

gthao313:
push above improve `KubernetesQuantity` regex and improve `cpu-manager-policy`.

packages/kubernetes-1.15/kubelet-config
```diff
@@ -29,6 +29,20 @@ authorization:
...
+  {{~#each settings.kubernetes.kube-reserved}}
+  {{@key}}: "{{this}}"
+  {{~/each}}
+cpuManagerPolicy: "static"
+{{~#if settings.kubernetes.allowed-unsafe-sysctls}}
+allowedUnsafeSysctls: {{settings.kubernetes.allowed-unsafe-sysctls}}
+{{~/if}}
```
etungsten: nit: let's move the `cpuManagerPolicy` specification below all the handlebar expressions for clarity
```suggestion
{{~#if settings.kubernetes.allowed-unsafe-sysctls}}
allowedUnsafeSysctls: {{settings.kubernetes.allowed-unsafe-sysctls}}
{{~/if}}
cpuManagerPolicy: "static"
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,272 @@ mod test_kubernetes_bootstrap_token {
...
+    inner: String,
+}
+
+// Regular expression of Kubernetes quantitiy. i.e. 128974848, 129e6, 129M, 123Mi
+lazy_static! {
+    pub(crate) static ref KUBERNETES_QUANTITY: Regex = Regex::new(
+        r"(?x)
+        # format1 for scientific notations e.g. 123e4 or:
+        ^([+-]?[0-9.]+)((e)?[0-9]*)$ |
+        # format2 for those suffixes [EPTGMK] and [EiPiTiGiMiKi] e.g. 100K or 100Ki, or:
```
etungsten: nit: 
```suggestion
        # format2 for values with unit suffixes [EPTGMK] and [EiPiTiGiMiKi] (e.g. 100K or 100Ki), or no units (e.g. 100):
```

packages/kubernetes-1.15/kubelet-config
etungsten: If we're defaulting `ephemeral-storage` to some value, we should document that in the README somehow (probably good to explain why as well). I think we should also still wrap `kubeReserved` in a "if" block like the other fields just in case since it doesn't hurt.

gthao313: Yeah. I'll document that, and I take this idea from EKS as well same as `cpu_reserved` and `memory_reserved`. right, it does not hurt, but I think since we default  `kubeReserved` in bottlerocket, `kubeReserved` will always have values there.
If it's unnecessary to wrap it in a `if` block?

zmrow: @etungsten If we're defaulting values for this - is it necessary to wrap in an "if" block?  I'm just trying to understand the reasons why we would

zmrow (approved):
I think we're almost to the finish line here.

Other than my comment below, I'd like to make sure that any defaults we set are documented.

packages/kubernetes-1.15/kubelet-config
etungsten: It certainly isn't necessary. It just seems more consistent with the other fields that way. It also means the kubeconfig will still be valid even if we're somehow missing that setting for whatever reason.
But I'll relent and agree that if we know we're keeping that default for sure, we can just not have the if block. I'm ok with keeping it as is.

README.md
```diff
@@ -320,6 +334,11 @@ The following settings are set for you automatically by [pluto](sources/api/) ba
 * `settings.kubernetes.cluster-dns-ip`: The CIDR block of the primary network interface.
 * `settings.kubernetes.node-ip`: The IPv4 address of this node.
 * `settings.kubernetes.pod-infra-container-image`: The URI of the "pause" container.
+* `settings.kubernetes.kube-reserved`: Resources reserved for node components.
+  * Bottlerocket provides default values for the resources. Default values are taken from [EKS' calculations](https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh):
+    `cpu`: in millicores from the total number of vCPUs available on the instance.
```
etungsten: If you intend for these to be bullets, you need to add a "*" in front of each bullet point. You can check the richdiff by clicking on the page icon in the upper right corner when viewing diffs.

gthao313:
push above add document for `kube-reserved` default

README.md
gthao313: yeah. I will fix in next push

packages/kubernetes-1.15/kubelet-config
webern: My concern is that we are adding settings with our own defaults that were not there before? @etungsten is there any chance that these changes would result in different behavior for a user that hasn't set these settings (ephemeral-storage, for example)? I tried to figure out the answer but could not.

etungsten: The change would reserve some system resources by default for kubelet (just kubelet right now I think) (e.g. 1 gibibyte of ephemeral storage). Users shouldn't be seeing any behavioral changes unless they're squeezing the system to the limit so much so that kubelet isn't operating normally. These values are aligned with what EKS does for their EKS-optimized AMIs so I think they make sense here for our EKS-specific K8S variants.

gthao313:
push above improve README

webern (approved):

etungsten (approved):

zmrow (approved):
:waffle:

packages/kubernetes-1.19/kubelet-config
```diff
@@ -29,6 +29,20 @@ authorization:
...
+  {{~/each}}
+{{~/if}}
+{{~#if settings.kubernetes.allowed-unsafe-sysctls}}
+allowedUnsafeSysctls: {{settings.kubernetes.allowed-unsafe-sysctls}}
+{{~/if}}
+kubeReserved:
+  {{~#each settings.kubernetes.kube-reserved}}
+  {{@key}}: "{{this}}"
+  {{~/each}}
+cpuManagerPolicy: "static"
```
bcressey, line 45: The "Add new setting: Cpu Manager Policy" commit should be renamed since we're not adding a setting.

"kubelet: use static CPU manager policy by default"

README.md
```diff
@@ -309,6 +309,20 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.standalone-mode`: Whether to run the kubelet in standalone mode, without connecting to an API server.  Defaults to `false`.
 * `settings.kubernetes.authentication-mode`: Which authentication method the kubelet should use to connect to the API server, and for incoming requests.  Defaults to `aws` for AWS variants, and `tls` for other variants.
 * `settings.kubernetes.bootstrap-token`: The token to use for [TLS bootstrapping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/).  This is only used with the `tls` authentication mode, and is otherwise ignored.
+* `settings.kubernetes.eviction-hard`: The hard eviction Signal and Thresholds you set up to reclaim the associated starved resource.
```
```suggestion
* `settings.kubernetes.eviction-hard`: The signals and thresholds that trigger pod eviction.
```

README.md
```diff
@@ -320,6 +334,11 @@ The following settings are set for you automatically by [pluto](sources/api/) ba
 * `settings.kubernetes.cluster-dns-ip`: The CIDR block of the primary network interface.
 * `settings.kubernetes.node-ip`: The IPv4 address of this node.
 * `settings.kubernetes.pod-infra-container-image`: The URI of the "pause" container.
+* `settings.kubernetes.kube-reserved`: Resources reserved for node components.
+  * Bottlerocket provides default values for the resources. Default values are taken from [EKS' calculations](https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh):
```
bcressey: I would prefer to link to our own implementation within the Bottlerocket repo.

README.md
```diff
@@ -309,6 +309,20 @@ The following settings are optional and allow you to further configure your clus
...
 * `settings.kubernetes.authentication-mode`: Which authentication method the kubelet should use to connect to the API server, and for incoming requests.  Defaults to `aws` for AWS variants, and `tls` for other variants.
 * `settings.kubernetes.bootstrap-token`: The token to use for [TLS bootstrapping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/).  This is only used with the `tls` authentication mode, and is otherwise ignored.
+* `settings.kubernetes.eviction-hard`: The hard eviction Signal and Thresholds you set up to reclaim the associated starved resource.
+  Remember to quote keys (since they often contain ".") and to quote all values.
+  * Example user data for setting up eviction hard:
+    ```
+    [settings.kubernetes.eviction-hard]
+    "memory.available" = "15%"
+    "nodefs.inodesFree" = "15Mi"
+    "pid.available" = "15%"
```
bcressey: It might be better to have just one example document, potentially in a different file, to avoid cluttering the main README with specifics about settings that aren't widely used.

sources/Cargo.toml
```diff
@@ -46,6 +46,8 @@ members = [
     "api/migration/migrations/v1.0.6/control-container-v0-4-2",
     "api/migration/migrations/v1.0.6/add-shibaken",
     "api/migration/migrations/v1.0.6/admin-container-v0-6-0",
+    "api/migration/migrations/v1.1.0/kubelet-kube-reserved-and-eviction-hard",
+    "api/migration/migrations/v1.1.0/kubelet-unsafe-sysctl",
```
bcressey: Let's use `v1.0.8` until we know what our next release version will be.

sources/models/src/lib.rs
```diff
@@ -127,6 +128,9 @@ struct KubernetesSettings {
     authentication_mode: KubernetesAuthenticationMode,
     bootstrap_token: KubernetesBootstrapToken,
     standalone_mode: bool,
+    eviction_hard: HashMap<EvictionHardKey, KubernetesThresholdValue>,
+    kube_reserved: HashMap<ReservedResourcesKey, KubernetesQuantityValue>,
```
bcressey: Let's prefix the types that are specific to Kubernetes.

```suggestion
    eviction_hard: HashMap<KubernetesEvictionHardKey, KubernetesThresholdValue>,
    kube_reserved: HashMap<KubernetesReservedResourceKey, KubernetesQuantityValue>,
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,273 @@ mod test_kubernetes_bootstrap_token {
         }
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// EvictionHardKey represents a string that contains a valid Kubernetes eviction hard signal.
+/// Available eviction hard signals: https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/
```
bcressey: "eviction hard signal" is not a term that can be found anywhere in the Kubernetes documentation, so I'd rather not introduce it in our own docs.

Maybe `KubernetesEvictionKey` + `KubernetesEvictionSignal` for the two types here.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,273 @@ mod test_kubernetes_bootstrap_token {
...
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesThresholdValue represents a string that contains a valid kubernetes threshold value.
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesThresholdValue {
+    inner: String,
+}
+
+// Regular expression of Kubernetes quantitiy. i.e. 128974848, 129e6, 129M, 123Mi
```
```suggestion
// Regular expression of Kubernetes quantity. i.e. 128974848, 129e6, 129M, 123Mi
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,273 @@ mod test_kubernetes_bootstrap_token {
...
+lazy_static! {
+    pub(crate) static ref KUBERNETES_QUANTITY: Regex = Regex::new(
+        r"(?x)
+        # format1 for scientific notations (e.g. 123e4) or:
+        ^([+-]?[0-9.]+)((e)?[0-9]*)$ |
+        # format2 for values with unit suffixes [EPTGMK] and [EiPiTiGiMiKi] (e.g. 100K or 100Ki),
+        # or no units (e.g. 100) or:
+        ^([+-]?[0-9.]+)((E|P|T|G|M|K)i?)?$ |
```
bcressey, line 539: Does kubelet reject `123e4M`?

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -437,3 +437,273 @@ mod test_kubernetes_bootstrap_token {
...
+    #[test]
+    fn good_eviction_hard_key() {
+        for ok in &[
+            "memory.available",
+            "nodefs.available",
+            "nodefs.inodesFree",
+            "imagefs.available",
+            "imagefs.inodesFree",
+            "pid.available",
+        ] {
```
bcressey, line 509: Ideally we'd go a step further and ensure that each key could only be paired with the expected kind of value, so we wouldn't accept `memory.available = 5u`. However, I don't see a straightforward way to achieve that.

sources/api/pluto/src/main.rs
```diff
@@ -270,11 +271,80 @@ fn pause_container_uri(region: &str) -> Result<String> {
...
+/// Returns the number of CPUs on current instance.
+fn get_num_cpus() -> usize {
+    num_cpus::get()
+}
+
+/// Calculates the amount of memory to reserve for kubeReserved in mebibytes.
+/// This function is taken from EKS' calculations.
+/// https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh#L228-L239
+fn get_memory_mebibytes_to_reserve(client: &Client, session_token: &str) -> Result<String> {
+    let max_num_pods = get_max_pods(&client, &session_token)?;
```
bcressey: A few issues here:
* others have mentioned the magic number, we need a comment explaining it.
* This results in a second query to IMDS, which is unfortunate.
* Doing this in `pluto` ties it to AWS.

Ideally we would update the memory reservation whenever `max-pods` was updated, or whenever the physical resources on the machine changed.

What would be really great here is if in the kubelet-config template, we could do something like this:
```
kubeReserved:
  {{~#if settings.kubernetes.kube-reserved.cpu}}
  cpu: {{settings.kubernetes.kube-reserved.cpu}}
  {{~else}
  cpu: {{calculate-cpu-reserve}}
  {{~/if}}
  {{~#if settings.kubernetes.kube-reserved.memory}}
  memory: {{settings.kubernetes.kube-reserved.memory}}
  {{~else}
  {{~#if settings.kubernetes.max-pods}}
  memory: {{calculate-mem-reserve settings.kubernetes.max-pods}}
  {{~/if}}
  {{~/if}}
```

Then if we were given a specific value, we'd always use it, but otherwise we'd run the calculation when the template was rendered.

sources/api/pluto/src/main.rs
```diff
@@ -270,11 +271,80 @@ fn pause_container_uri(region: &str) -> Result<String> {
...
+
+        cpu_to_reserve += get_resource_to_reserve_in_range(
+            total_cpu_on_instance,
+            start_range,
+            end_range,
+            percentage_to_reserve_for_range,
+        )
+    }
+    Ok(cpu_to_reserve.to_string() + millicores_unit)
+}
```
bcressey: It would be more idiomatic to write this with iterators.

For CPUs, have an iterator that repeats 1000 and take num_cpus elements from it.

For percentages, chain an iterator that produces 0.06, 0.01, 0.005, with an iterator that repeats 0.0025.

Then zip the two together, multiply each pair, and sum up the result.

sources/api/pluto/src/main.rs
```diff
@@ -270,11 +271,80 @@ fn pause_container_uri(region: &str) -> Result<String> {
...
+
+/// This function is taken from EKS' calculations.
+/// https://github.com/awslabs/amazon-eks-ami/blob/db28da15d2b696bc08ac3aacc9675694f4a69933/files/bootstrap.sh#L203-L226
```
bcressey: We don't need to repeat this comment everywhere, once is probably enough.

sources/api/pluto/src/main.rs
```diff
@@ -383,3 +455,101 @@ mod test_pause_container_account {
...
+    use crate::get_cpu_millicores_to_reserve;
+    #[test]
+    fn get_cpu_millicores_to_reserve_ok() {
+        let mut cpu_reserved: HashMap<usize, &str> = HashMap::new();
+        cpu_reserved.insert(0, "0m");
+        cpu_reserved.insert(1, "60m");
+        cpu_reserved.insert(2, "70m");
+        cpu_reserved.insert(3, "75m");
+        cpu_reserved.insert(4, "80m");
+        cpu_reserved.insert(5, "82.5m");
```
bcressey: Fractional millicores don't make a ton of sense to me. Even if kubelet accepts values like this - does it? - we should round up or down to an integer value.

bcressey (changes requested):

sources/models/src/modeled_types/kubernetes.rs
gthao313: Yeah. If we pass `123e4M` to `kubeReserved` or `evictionHard`, it will fail to run Kubelet because quantities must match the regular expression .

sources/models/src/modeled_types/kubernetes.rs
gthao313: If we do that each key could only be paired with the expected kind of value, I think that will be complicated and I haven't came up with a good solution yet. And my understanding for setting input validation is that we validate the input to make sure the input follows `kubelet` rules. If the input like `memory.available = 5u` can be accepted by `kubelet`, I think that's a valid input. Please let me know what do you think? :grin:

sources/api/pluto/src/main.rs
gthao313: Yeah. `kubelet` accepts  fractional millicores like `82.5m` and I will round down. Can you explain why fractional millicores don't make sense?

README.md
gthao313: Sure thing!

sources/api/pluto/src/main.rs
gthao313: The reason why I repeat is those comments are slightly different. The link points to the place where we take the function from (e.g. `...bootstrap.sh#L203-L226`).  Do you think we should comment once or ?

sources/api/pluto/src/main.rs
gthao313: That's a good idea. The reason why I do this way is that I want to make it consistent with `get_cpu_millicores_to_reserve` and `get_resource_to_reserve_in_range` functions in `amazon-eks-ami`, where I got the idea from. Do you think I should optimize it or keep this?

sources/Cargo.toml
gthao313: Sure.

sources/api/pluto/src/main.rs
bcressey: Definitely optimize it - the comments from the EKS AMI implementation are valuable, but translating the implementation directly from shell script to Rust makes for awkward code.

webern: Yeah see my previous comment here: https://github.com/bottlerocket-os/bottlerocket/pull/1388#discussion_r594746238

The algorithm is very simple and can be done with a match statement and a few constants. I had to request tests first to make sure I even understood it!

I agree with Ben on this, simplify it down to a match statement that returns the correct value.

anguslees:
Fyi for code archeologists:  Any particular PR is of course only a snapshot, but the data/discussion in https://github.com/awslabs/amazon-eks-ami/pull/419 is particularly relevant to the EKS reserved CPU/memory calculation.

gthao313:
push above improve `kube-reserved`'s `calculate-memory-reserved` and `get-cpu-reserved` functions, and improve format and comments.

packages/kubernetes-1.15/kubelet-config
```diff
@@ -29,6 +29,34 @@ authorization:
...
+  {{~/each}}
+{{~/if}}
+{{~#if settings.kubernetes.allowed-unsafe-sysctls}}
+allowedUnsafeSysctls: {{settings.kubernetes.allowed-unsafe-sysctls}}
+{{~/if}}
+kubeReserved:
+  {{~#if settings.kubernetes.kube-reserved.cpu}}
+  cpu: "{{settings.kubernetes.kube-reserved.cpu}}"
+  {{~else}}
+  cpu: "{{get_cpu_millicores_to_reserve}}"
```
webern: Instead of an if/else here, should we consider putting the switching logic in the Schnauzer helper instead? So Schnauzer would either use the supplied argument `settings.kubernetes.kube-reserved.cpu` or calculate the backup value if it is empty.

I'm assuming that `get_cpu_millicores_to_reserve  settings.kubernetes.kube-reserved.cpu` would *not* error if the user did not set `kube-reserved.cpu`.

It would cut down on the unsightliness of the template file but I'm not 100% sure it's a better solution. Do others have thoughts?

webern: I'm also wondering why one of the names is `get_` and the other is `calculate_`

gthao313: Oh I have a function in `Schnauzer ` named `calculate_cpu_..`, so I use `get_` here. I think I should change it to `calculate_` to be more consistent.

etungsten: > I'm assuming that get_cpu_millicores_to_reserve settings.kubernetes.kube-reserved.cpu would not error if the user did not set kube-reserved.cpu.

I think we can treat it like the `default` helper. 

https://github.com/bottlerocket-os/bottlerocket/blob/513c8a2033a7dee90a31061c9f67895e5b685b5d/sources/api/schnauzer/src/helpers.rs#L411-L418

Instead of the `default.to_string()`, it would be something like `calculate_cpu_reserve(num_cores).to_string();`

gthao313:
push above improve `calculate_memory` `calculate_cpu` in schnauzer and `kube-config` template

webern (approved):

etungsten (approved):

packages/kubernetes-1.15/kubelet-config
```diff
@@ -29,6 +29,20 @@ authorization:
...
+  {{~#each settings.kubernetes.eviction-hard}}
+  {{@key}}: "{{this}}"
+  {{~/each}}
+{{~/if}}
+{{~#if settings.kubernetes.allowed-unsafe-sysctls}}
+allowedUnsafeSysctls: {{settings.kubernetes.allowed-unsafe-sysctls}}
+{{~/if}}
+kubeReserved:
+  cpu: "{{calculate_cpu_millicores_to_reserve settings.kubernetes.kube-reserved.cpu}}"
+  memory: "{{calculate_memory_mebibytes_to_reserve settings.kubernetes.max-pods settings.kubernetes.kube-reserved.memory}}"
```
bcressey: nit: could make these a bit less wordy:
```suggestion
  cpu: "{{kube_reserve_cpu settings.kubernetes.kube-reserved.cpu}}"
  memory: "{{kube_reserve_mem settings.kubernetes.max-pods settings.kubernetes.kube-reserved.memory}}"
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -592,6 +612,123 @@ pub fn join_array(
...
+                    expected: "scalar",
+                    value: max_num_pods_val.to_owned(),
+                    template: template_name.to_owned(),
+                },
+            ))
+        }
+    };
+    let max_num_pods_u32 = max_num_pods.parse::<u32>().context(error::ParseToU32 {
+        object: max_num_pods,
+    })?;
```
bcressey: nit: can just reuse the variable name, since we don't need the unconverted form again:
```suggestion
    let max_num_pods = max_num_pods.parse::<u32>().context(error::ParseToU32 {
        object: max_num_pods,
    })?;
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -44,6 +45,19 @@ lazy_static! {
...
+/// (https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture#node_allocatable):
+/// 6% of the first core
+/// 1% of the next core (up to 2 cores)
+/// 0.5% of the next 2 cores (up to 4 cores)
+/// 0.2% of any cores above 4 cores
+const RESERVE_1_CORE: usize = 60;
+const RESERVE_2_CORES: usize = RESERVE_1_CORE + 10;
+const RESERVE_3_CORES: usize = RESERVE_2_CORES + 5;
+const RESERVE_4_CORES: usize = RESERVE_3_CORES + 5;
+const RESERVE_ADDITIONAL: usize = 2;
```
bcressey: nit:
```suggestion
const KUBE_RESERVE_1_CORE: usize = 60;
const KUBE_RESERVE_2_CORES: usize = RESERVE_1_CORE + 10;
const KUBE_RESERVE_3_CORES: usize = RESERVE_2_CORES + 5;
const KUBE_RESERVE_4_CORES: usize = RESERVE_3_CORES + 5;
const KUBE_RESERVE_ADDITIONAL: usize = 2;
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -592,6 +612,123 @@ pub fn join_array(
...
+        Value::Null | Value::Array(_) | Value::Object(_) => {
+            return Err(RenderError::from(
+                error::TemplateHelperError::InvalidTemplateValue {
+                    expected: "scalar",
+                    value: max_num_pods_val.to_owned(),
+                    template: template_name.to_owned(),
+                },
+            ))
+        }
+    };
```
bcressey: Can we reject everything but the `Value::Number(n)` case, and convert it directly to a `u32`?

We do that elsewhere with keys that we know should be strings.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -592,6 +612,123 @@ pub fn join_array(
...
+    let memory_to_reserve_value = get_param(helper, 1)?;
+    let memory_to_reserve = match memory_to_reserve_value {
+        // these ones Display as their simple scalar selves
+        Value::Bool(b) => b.to_string(),
+        Value::Number(n) => n.to_string(),
+        Value::String(s) => s.to_string(),
+        // If no value is set, use the given default.
+        Value::Null => {
+            format!("{}Mi", (max_num_pods_u32 * 11 + 255).to_string())
+        }.to_string(),
```
bcressey: I could imagine how we'd get `Number`, `String`, or `Null` back, but probably not `Bool`, so it should also be unsupported.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -592,6 +612,123 @@ pub fn join_array(
...
+    check_param_count(helper, template_name, 1)?;
+
+    // Calculates the amount of CPU to reserve
+    let num_cores = num_cpus::get();
+    let cpu_to_reserve_value = get_param(helper, 0)?;
+    let cpu_to_reserve = match cpu_to_reserve_value {
+        Value::Bool(b) => b.to_string(),
+        Value::Number(n) => n.to_string(),
+        Value::String(s) => s.to_string(),
+        // If no value is set, use the given default.
```
bcressey: Same note here about restricting the types we'll accept.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -646,6 +783,20 @@ fn ecr_registry<S: AsRef<str>>(region: S) -> String {
     format!("{}.dkr.ecr.{}.amazonaws.com", registry_id, region)
 }
 
+/// Calculates and returns the amount of CPU to reserve
+fn cpu_reserve(num_cores: usize) -> String {
```
bcressey: If you take the suggestion to rename the template-level helpers, this could be renamed to `kube_cpu_helper` so it doesn't read like it applies universally.

sources/models/src/lib.rs
```diff
@@ -101,7 +101,8 @@ use crate::modeled_types::{
     DNSDomain, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue, FriendlyVersion, Identifier,
     KubernetesAuthenticationMode, KubernetesBootstrapToken, KubernetesClusterName,
     KubernetesLabelKey, KubernetesLabelValue, KubernetesTaintValue,
-    Lockdown, SingleLineString, SysctlKey, Url, ValidBase64,
+    Lockdown, SingleLineString, SysctlKey, Url, ValidBase64, KubernetesEvictionHardKey, KubernetesThresholdValue,
+    KubernetesReservedResourceKey, KubernetesQuantityValue,
```
bcressey: super-nit: put these `Kubernetes` imports up with the other ones, optionally in alphabetical order

bcressey (commented):
As an overall test, can you make sure that `kubelet-config` is still the same for 1.15 + 1.16, 1.17 + 1.18, with 1.19 as an outlier?

```
❯ sha1sum packages/kubernetes-1.1*/kubelet-config
...

❯ diff -u -p packages/kubernetes-1.1{6,7}/kubelet-config
--- packages/kubernetes-1.16/kubelet-config	2021-03-29 19:27:33.165289932 +0000
+++ packages/kubernetes-1.17/kubelet-config	2021-03-29 19:27:33.165289932 +0000
@@ -36,6 +36,7 @@ cgroupRoot: "/"
 runtimeRequestTimeout: 15m
 featureGates:
   RotateKubeletServerCertificate: true
+  CSIMigration: false
 serializeImagePulls: false
 serverTLSBootstrap: true
 configMapAndSecretChangeDetectionStrategy: Cache

❯ diff -u -p packages/kubernetes-1.1{8,9}/kubelet-config
--- packages/kubernetes-1.18/kubelet-config	2021-03-29 19:27:33.166289937 +0000
+++ packages/kubernetes-1.19/kubelet-config	2021-03-29 19:27:33.166289937 +0000
@@ -42,5 +42,6 @@ serverTLSBootstrap: true
 configMapAndSecretChangeDetectionStrategy: Cache
 tlsCipherSuites:
+volumePluginDir: "/var/lib/kubelet/plugins/volume/exec"
 maxPods: {{default 110 settings.kubernetes.max-pods}}
 staticPodPath: "/etc/kubernetes/static-pods/"
```

sources/models/src/modeled_types/kubernetes.rs
bcressey: I'm not sure what `kubelet` would do with input like that, but in the absence of a better way to do the input validation, I'm also not inclined to block on this.

sources/api/schnauzer/src/helpers.rs
gthao313: Yeah we can reject everthing but `Value::Number(n)`; however, we can not convert it to `u32`. Because the type of this value is `&serde_json::Number`, so I can not phase it to `u32` directly. My solution is about converting it to `string` here, and then phase it to `u32`. If my solution is not that efficient, please correct me.

gthao313:
push above improve variable name and improve `kube_reserve_cpu` `kube_reserve_memory` in `schnauzer`.

README.md
```diff
@@ -318,6 +318,18 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.standalone-mode`: Whether to run the kubelet in standalone mode, without connecting to an API server.  Defaults to `false`.
 * `settings.kubernetes.authentication-mode`: Which authentication method the kubelet should use to connect to the API server, and for incoming requests.  Defaults to `aws` for AWS variants, and `tls` for other variants.
 * `settings.kubernetes.bootstrap-token`: The token to use for [TLS bootstrapping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/).  This is only used with the `tls` authentication mode, and is otherwise ignored.
+* `settings.kubernetes.eviction-hard`: The signals and thresholds that trigger pod eviction.
+  Remember to quote keys (since they often contain ".") and to quote all values.
```
bcressey: It looks like all signals are named in "x.y" form, so:

```suggestion
  Remember to quote signals (since they all contain ".") and to quote all values.
```

sources/api/pluto/src/main.rs
gthao313: I have document it. Thanks https://github.com/awslabs/amazon-eks-ami/pull/419

gthao313:
Push above improve number conversion process in `kube_reserve_memory` and improve `eviction-hard` in README

Additional Test:
Launching two m5.xlarge instances(one Bottlerocket with my changes & one EKS AL2) and two c5.12xlarge instances(one Bottlerocket with your changes & one EKS AL2) to confirm that we calculate the same values across max pods, CPU reserve, and memory reserve.
```
Bottlerocket m5 xlarge
kubeReserved:
  cpu: "80m"
  memory: "893Mi"
  ephemeral-storage: "1Gi"

EKS AL2 m5 xlargs
"kubeReserved": {
    "cpu": "80m",
    "ephemeral-storage": "1Gi",
    "memory": "893Mi"
  }

Bottlerocket c5 12.xlarge
kubeReserved:
  cpu: "190m"
  memory: "2829Mi"
  ephemeral-storage: "1Gi"

EKS AL2 c5 12.xlarge
"kubeReserved": {
	"cpu": "190m",
	"ephemeral-storage": "1Gi",
	"memory": "2829Mi"
}
```

bcressey (approved):

etungsten (approved):

zmrow (approved):
:mushroom:

packages/kubernetes-1.15/kubelet-config
```diff
@@ -29,6 +29,20 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 clusterDNS:
 - {{settings.kubernetes.cluster-dns-ip}}
+{{~#if settings.kubernetes.eviction-hard}}
+evictionHard:
```
diranged, line 33: This is awesome - we've built an AMI with this patch and its working wonderfully. In a followup PR, we'd like to see the `evictionSoft` settings exposed, as well as the `eviction-max-pod-grace-period` setting. Overall though, this is great!

gthao313: Hi @diranged. We have opened a new issue about supporting `eviction soft` related settings. You can follow up here https://github.com/bottlerocket-os/bottlerocket/issues/1445. Thanks!

webern (approved):

PR #1393, opened 2021-03-15
early-boot-config: Add guestinfo to available VMWare user data retrieval methods

### LINKED ISSUE #1369: early-boot-config: Add "guestinfo" datasource (host-guest protocol) for VMware variants
In the process of researching [ClusterAPI](https://cluster-api.sigs.k8s.io/), and more specifically the [vSphere data provider](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere), we discovered that CAPV [uses the VMWare host-guest protocol](https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/7a529442e2102159b1d51314c9b3dffba6ed0280/pkg/services/govmomi/vcenter/clone.go#L55) to provide user data to hosts built from templates.

`early-boot-config` should have the ability to read user data from this interface.

Related to #1218

zmrow:
**Issue number:**
Closes #1369 

**Description of changes:**
The changes aren't vast here, but it's probably easiest to read the commits separately.  
```
    early-boot-config: Rename CdRomDataProvider to VmwareDataProvider
    
    Given that the CdRomDataProvider contained VMware-specific logic and
    more will continue to be added to it, naming it as such makes it clearer
    to the caller and still allows for the underlying logic to be factored
    out in the future should we deem necessary.
```
```
    early-boot-config: Add `from_toml_str` to SettingsJson
    
    This adds a method to SettingsJson that takes in a string (expected to
    be TOML data), and attempts to parse it, remove the "settings" layer,
    and create a `SettingsJson` from it.  This logic was previously
    sprinkled around in a few places, creating multiple identical error
    variants, etc.
```
```    
    early-boot-config: Add guestinfo to available VMWare retrieval methods
    
    This change adds an additional user data source to the VMWare provider;
    the guestinfo (host-guest) interface.  This source will be checked
    after the CD-ROM, and override any settings set via CD-ROM.
```

I do think that there's some refactoring to be done in this program, namely to separate "platform" from "data provider", but that will be done in another PR.

**Testing done:**
* Boot an `aws-k8s-1.17` AMI to ensure I didn't break anything over there
* Boot a `vmware-dev` image with a CD-ROM attached containing an `ovf-env.xml` file w/ user data 
* Boot a `vmware-dev` image with a CD-ROM attached containing a TOML `user-data` file
* In vSphere with guestinfo:
  * Traverse the entire matrix of encodings `base64`, `b64`, `B64`, `Base64` and `gz+b64`, `Gz+B64`, `gzip+base64`, and `Gzip+Base64` with user data set and unset
  * Ensure that raw TOML user data passed via guestinfo still works
  * Ensure that settings passed via guestinfo override setting from an attached CD-ROM

**Terms of contribution:**

### DIFF
--- a/sources/api/early-boot-config/Cargo.toml
+++ b/sources/api/early-boot-config/Cargo.toml
@@ -18,6 +18,7 @@ log = "0.4"
 reqwest = { version = "0.10", default-features = false, features = ["blocking"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
+serde_plain = "0.3"
 serde-xml-rs = "0.4.1"
 simplelog = "0.9"
 snafu = "0.6"
@@ -25,6 +26,10 @@ tokio = { version = "0.2", default-features = false, features = ["macros", "rt-t
 # When hyper updates to tokio 0.3:
 #tokio = { version = "0.3", default-features = false, features = ["macros", "rt-multi-thread"] }
 toml = "0.5"
+
+[target.'cfg(target_arch = "x86_64")'.dependencies]
+# vmw_backdoor includes x86_64 assembly, prevent it from building for ARM
+vmw_backdoor = "0.2"
 
 [build-dependencies]
 cargo-readme = "3.1"
--- a/sources/api/early-boot-config/src/provider/vmware.rs
+++ b/sources/api/early-boot-config/src/provider/vmware.rs
@@ -4,7 +4,7 @@
 use super::{PlatformDataProvider, SettingsJson};
 use crate::compression::{expand_file_maybe, expand_slice_maybe, OptionalCompressionReader};
 use serde::Deserialize;
-use snafu::{ensure, OptionExt, ResultExt};
+use snafu::{ensure, ResultExt};
 use std::ffi::OsStr;
 use std::fs::File;
 use std::io::BufReader;
@@ -50,7 +50,7 @@ impl VmwareDataProvider {
         // format
         info!("'{}' exists, using it", user_data_file.display());
         let user_data_str = match user_data_file.extension().and_then(OsStr::to_str) {
-            Some("xml") | Some("XML") => Self::ovf_user_data(user_data_file)?,
+            Some("xml") | Some("XML") => Self::ovf_user_data(&user_data_file)?,
             // Since we only look for a specific list of file names, we should never find a file
             // with an extension we don't understand.
             Some(_) => unreachable!(),
@@ -78,15 +78,12 @@ impl VmwareDataProvider {
             );
         }
 
-        // Remove outer "settings" layer before sending to API
-        let mut val: toml::Value =
-            toml::from_str(&user_data_str).context(error::TOMLUserDataParse)?;
-        let table = val.as_table_mut().context(error::UserDataNotTomlTable)?;
-        let inner = table
-            .remove("settings")
-            .context(error::UserDataMissingSettings)?;
+        let json = SettingsJson::from_toml_str(&user_data_str, "user data").context(
+            error::SettingsToJSON {
+                from: user_data_file.display().to_string(),
+            },
+        )?;
 
-        let json = SettingsJson::from_val(&inner, "user data").context(error::SettingsToJSON)?;
         Ok(Some(json))
     }
 
@@ -190,20 +187,14 @@ mod error {
         #[snafu(display("Unable to read input file '{}': {}", path.display(), source))]
         InputFileRead { path: PathBuf, source: io::Error },
 
-        #[snafu(display("Error serializing TOML to JSON: {}", source))]
-        SettingsToJSON { source: serde_json::error::Error },
-
-        #[snafu(display("Error parsing TOML user data: {}", source))]
-        TOMLUserDataParse { source: toml::de::Error },
+        #[snafu(display("Unable to serialize settings from {}: {}", from, source))]
+        SettingsToJSON {
+            from: String,
+            source: crate::settings::Error,
+        },
 
         #[snafu(display("Found multiple user data files in '{}', expected 1", location))]
         UserDataFileCount { location: String },
-
-        #[snafu(display("TOML data did not contain 'settings' section"))]
-        UserDataMissingSettings,
-
-        #[snafu(display("Data is not a TOML table"))]
-        UserDataNotTomlTable,
 
         #[snafu(display("Unable to deserialize XML from: '{}': {}", path.display(), source))]
         XmlDeserialize {
--- a/sources/api/early-boot-config/src/provider/vmware.rs
+++ b/sources/api/early-boot-config/src/provider/vmware.rs
@@ -1,5 +1,5 @@
-//! The cdrom module implements the `PlatformDataProvider` trait for gathering userdata from a
-//! mounted CDRom.
+//! The vmware module implements the `PlatformDataProvider` trait for gathering userdata on VMWare
+//! via mounted CDRom or the guestinfo interface
 
 use super::{PlatformDataProvider, SettingsJson};
 use crate::compression::{expand_file_maybe, expand_slice_maybe, OptionalCompressionReader};
@@ -11,9 +11,9 @@ use std::io::BufReader;
 use std::iter::FromIterator;
 use std::path::Path;
 
-pub(crate) struct CdromDataProvider;
+pub(crate) struct VmwareDataProvider;
 
-impl CdromDataProvider {
+impl VmwareDataProvider {
     // This program expects that the CD-ROM is already mounted.  Mounting happens elsewhere in a
     // systemd unit file
     const CD_ROM_MOUNT: &'static str = "/media/cdrom";
@@ -134,7 +134,7 @@ impl CdromDataProvider {
     }
 }
 
-impl PlatformDataProvider for CdromDataProvider {
+impl PlatformDataProvider for VmwareDataProvider {
     fn platform_data(&self) -> std::result::Result<Vec<SettingsJson>, Box<dyn std::error::Error>> {
         let mut output = Vec::new();
 
@@ -229,7 +229,7 @@ mod test {
         let xml = test_data().join("namespaced_keys.xml");
         let expected_user_data = "settings.motd = \"hello\"";
 
-        let actual_user_data = CdromDataProvider::ovf_user_data(xml).unwrap();
+        let actual_user_data = VmwareDataProvider::ovf_user_data(xml).unwrap();
 
         assert_eq!(actual_user_data, expected_user_data)
     }
@@ -239,7 +239,7 @@ mod test {
         let xml = test_data().join("ovf-env.xml");
         let expected_user_data = "settings.motd = \"hello\"";
 
-        let actual_user_data = CdromDataProvider::ovf_user_data(xml).unwrap();
+        let actual_user_data = VmwareDataProvider::ovf_user_data(xml).unwrap();
 
         assert_eq!(actual_user_data, expected_user_data)
     }
--- a/sources/api/early-boot-config/src/provider/vmware.rs
+++ b/sources/api/early-boot-config/src/provider/vmware.rs
@@ -8,8 +8,11 @@ use snafu::{ensure, ResultExt};
 use std::ffi::OsStr;
 use std::fs::File;
 use std::io::BufReader;
+use std::io::Cursor;
+use std::io::Read;
 use std::iter::FromIterator;
 use std::path::Path;
+use std::str;
 
 pub(crate) struct VmwareDataProvider;
 
@@ -26,9 +29,15 @@ impl VmwareDataProvider {
         "OVF_ENV.XML",
     ];
 
-    /// Given the list of acceptable filenames, ensure only 1 exists and parse
-    /// it for user data
-    fn user_data() -> Result<Option<SettingsJson>> {
+    // The fields in which user data and its encoding are stored in guestinfo
+    const GUESTINFO_USERDATA: &'static str = "guestinfo.userdata";
+    const GUESTINFO_USERDATA_ENCODING: &'static str = "guestinfo.userdata.encoding";
+
+    /// Read and decode user data from files via mounted CD-ROM
+    fn cdrom_user_data() -> Result<Option<SettingsJson>> {
+        // Given the list of acceptable filenames, ensure only 1 exists and parse
+        // it for user data
+        info!("Attempting to retrieve user data from mounted CD-ROM");
         let mut user_data_files = Self::USER_DATA_FILENAMES
             .iter()
             .map(|filename| Path::new(Self::CD_ROM_MOUNT).join(filename))
@@ -78,7 +87,7 @@ impl VmwareDataProvider {
             );
         }
 
-        let json = SettingsJson::from_toml_str(&user_data_str, "user data").context(
+        let json = SettingsJson::from_toml_str(&user_data_str, "user data from CD-ROM").context(
             error::SettingsToJSON {
                 from: user_data_file.display().to_string(),
             },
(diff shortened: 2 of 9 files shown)


zmrow:
^ Fixed up a failing test I missed... :)

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    /// Read and decode user data based on values retrieved from the guestinfo interface
+    fn guestinfo_user_data() -> Result<Option<SettingsJson>> {
+        info!("Attempting to retrieve user data via guestinfo interface");
+
+        // Request a key's value from guestinfo
+        // TODO: Figure out if it's possible to re-use the backdoor or ERPC channel (mutable
+        // references everywhere...)
+        fn backdoor_get_bytes(key: &str) -> Result<Option<Vec<u8>>> {
```
webern: I don't quite understand the TODO. Is there a reason this function is nested? I don't see it capturing anything.

zmrow: The function is nested because I would never expect (or want to) use it outside of the current function.  I think when some of these "data provider" methods are factored out a bit that will change, but until then it's safely located where it can't be misused. :)

Re: the TODO: My first iteration of this code created a backdoor and ERPC channel, then attempted to use it to pull the encoding and userdata.  The first call would succeed, the second call would fail, almost like the channel was "used up".  As soon as I instantiated a new backdoor/channel for the second call, everything was happy.  I've never used an ERPC channel like this before and didn't get too far into the source code to follow all the mutable references and see what's happening, so I left the TODO there as a reminder that it may not be necessary to create a backdoor/channel for each call.  There are only 2 calls, so I think it's ok for now.

webern: It is common to break out private functions even if they are only called in one place. It makes it easier to read. Nesting the function seems very unidiomatic to me. I very strongly advocate for moving it outside of its calling function.

For precedent, look at some of our main.rs files, where we have functions that are called only once! i.e. They exist purely for human readability and don't even facilitate code re-use.

The human readability factor works like this `backdoor_get_bytes` is named clearly enough for me to understand that I'm getting some bytes from somewhere, so I can follow the logic of `guestinfo_user_data` without concerning myself about where the bytes are coming from.  Unless, I do care, in which case I go look at `backdoor_get_bytes` and see a crisp isolated function that tells me where the bytes come from.

sources/api/early-boot-config/Cargo.toml
```diff
@@ -18,13 +18,15 @@ log = "0.4"
...
 serde_json = "1"
+serde_plain = "0.3"
 serde-xml-rs = "0.4.1"
 simplelog = "0.9"
 snafu = "0.6"
 tokio = { version = "0.2", default-features = false, features = ["macros", "rt-threaded"] }
 # When hyper updates to tokio 0.3:
 #tokio = { version = "0.3", default-features = false, features = ["macros", "rt-multi-thread"] }
 toml = "0.5"
+vmw_backdoor = "0.2"
```
tjkirch, line 32: vmw_backdoor only works on x86_64 (using x86 asm) and I believe VMware only supports x86 virtualization (though it can now run on ARM hosts), so we might want to do something to stop BUILDSYS_ARCH=aarch64 BUILDSYS_VARIANT=vmware-dev.  Maybe we should check the arch/variant combo in Makefile.toml so we don't get this far; it seems like any solution localized to early-boot-config could raise issues regarding whether conditional compilation / arch-specific errors would break unit tests, etc.  And maybe just a comment here?

zmrow: Good point - I think a solution in `Makefile.toml` is the way to go.

zmrow: I'll probably have to create a platform specific dependency for `vmw_backdoor` in `Cargo.toml` anyway, just to keep `early-boot-config` from failing to build for ARM.

webern: > Good point - I think a solution in Makefile.toml is the way to go.

Makefile solutions mean that we can't use our workspace for development anymore :(
I guess it's fine as long as you are on x86_64.

webern: P.S. variants that only work for some arches seems like a new concept.

tjkirch: > Makefile solutions mean that we can't use our workspace for development anymore

What does this mean?  What workspace?  What would break?

> variants that only work for some arches seems like a new concept.

It is, one that we have to tackle somehow until VMware supports ARM guests...

> I'll probably have to create a platform specific dependency for vmw_backdoor in Cargo.toml anyway, just to keep early-boot-config from failing to build for ARM.

Having an early-boot-config on VMware that doesn't support guestinfo would be kind of unhelpful, right?  Seems like we should prevent that from existing at all, with a higher level solution..?

zmrow: I think we're going to have to do some combination of both a `Makefile` solution and a crate level solution at the `early-boot-config` level.  If you look at CI (and I'm able to reproduce this locally), _every_ variant fails to build for ARM because `vmw_backdoor` is a dependency of `early-boot-config` and includes x86 assembly.  Even though `vmw_backdoor` crate [compiles to an empty crate if not on x86](https://github.com/lucab/vmw_backdoor-rs/blob/ff94a690bf813a1e1915ff774a54ea88fe9b6f51/src/lib.rs#L51) it's still getting compiled.  See below error messages from the build.

```
   #17 11.78 warning: src/asm/x86_64-linux.s:152: Error: unknown mnemonic `movq' -- `movq %r9,%rbp'
  #17 11.78 warning: src/asm/x86_64-linux.s:155: Error: unknown mnemonic `xor' -- `xor %rax,%rax'
  #17 11.78 
  #17 11.78 error: failed to run custom build command for `vmw_backdoor v0.2.0`
```

My suggested solution:
* In `Makefile.toml`, prevent users from building the `vmware*` / `aarch64` combo.
* In `early-boot-config/Cargo.toml`, prevent the `vmw_backdoor` crate from being build for `aarch64` with the addition of below.
```
+[target.'cfg(target_arch = "x86_64")'.dependencies]                                                                                              
 vmw_backdoor = "0.2"
```

webern: > What does this mean? What workspace? What would break?

If I understand the problem correctly, this would not work on an ARM developer machine:

```sh
cd sources && export VARIANT=anyvariant? && cargo build
```

What's not clear in my mind is the list of VARIANT values that would fail to build (only VMWARE? in that case maybe its fine since it would be reasonable to assume you can't compile VMWare bits on an arm machine in that way.

webern: Does this help? https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies

webern: Sorry, Zack already found the answer before my previous two comments:

> [target.'cfg(target_arch = "x86_64")'.dependencies]

zmrow: @bcressey made a good point: Another (probably more preferable) solution would be to specify the supported architectures for a variant in the variant's `Cargo.toml`.   Then we can avoid extra cruft in the `Makefile` as well.  `buildsys` would just refuse to build a variant unless the current arch is in the list of supported arches.

zmrow:
^ adds the platform dependency to `Cargo.toml` in `early-boot-config`. CI should all pass this time.

sources/api/early-boot-config/Cargo.toml
zmrow: Opened #1404 to track this feature for `buildsys`

zmrow:
^ un-nested the `backdoor_get_bytes` function per @webern

webern (approved):

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -132,23 +134,146 @@ impl CdromDataProvider {
...
+        // the same data so we fall back to "unprivileged" if the first call fails.
+        // https://github.com/lucab/vmw_backdoor-rs/issues/6
+        let mut backdoor = vmw_backdoor::probe_backdoor_privileged()
+            .or_else(|e| {
+                warn!(
+                    "Unable to access backdoor via privileged mode, using unprivileged: {}",
+                    e
+                );
+                vmw_backdoor::probe_backdoor()
+            })
```
bcressey: Did you test that the fallback path works with `settings.kernel.lockdown = "integrity"`?

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -132,23 +134,146 @@ impl CdromDataProvider {
...
+        Ok(Some(json))
+    }
+
+    /// Request a key's value from guestinfo
+    // TODO: Figure out if it's possible to re-use the backdoor or ERPC channel (mutable
+    // references everywhere...)
+    fn backdoor_get_bytes(key: &str) -> Result<Option<Vec<u8>>> {
+        // Probe and access the VMWare backdoor.  `kernel lockdown(7)` may block "privileged"
+        // mode because of its use of `iopl()`.  According the the bug report below, in theory
+        // "privileged" mode is more reliable than "unprivileged" but both provide access to
```
bcressey, line 214: Wouldn't hurt to check `vmw_backdoor::is_vmware_cpu()`.

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -132,23 +134,146 @@ impl CdromDataProvider {
...
+                })?
+            }
+        };
+
+        let json = SettingsJson::from_toml_str(&user_data_string, "user data")?;
+        Ok(Some(json))
+    }
+
+    /// Request a key's value from guestinfo
+    // TODO: Figure out if it's possible to re-use the backdoor or ERPC channel (mutable
```
bcressey: Is this important? I'm not sure why it's a `TODO` vs. either implementing it or not implementing it if we don't need it.

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -190,21 +183,22 @@ mod error {
...
+        ))]
+        InvalidUTF8 {
+            base64_string: String,
+            source: std::str::Utf8Error,
+        },
 
-        #[snafu(display("Error parsing TOML user data: {}", source))]
-        TOMLUserDataParse { source: toml::de::Error },
+        #[snafu(context(false))]
+        SettingsToJSON { source: crate::settings::Error },
```
tjkirch: I'm less confident about removing context from this error variant.  VMware has multiple sources of user data.  Is it going to be clear which failed?

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -190,21 +183,22 @@ mod error {
...
         InputFileRead { path: PathBuf, source: io::Error },
 
-        #[snafu(display("Error serializing TOML to JSON: {}", source))]
-        SettingsToJSON { source: serde_json::error::Error },
+        #[snafu(display(
+            "Invalid (non-utf8) output from base64 string '{}': {}",
+            base64_string,
+            source
+        ))]
+        InvalidUTF8 {
```
tjkirch: Looks like this is being added in the wrong commit; it's not used until the guestinfo commit.

sources/api/early-boot-config/Cargo.toml
```diff
@@ -26,6 +27,10 @@ tokio = { version = "0.2", default-features = false, features = ["macros", "rt-t
 #tokio = { version = "0.3", default-features = false, features = ["macros", "rt-multi-thread"] }
 toml = "0.5"
 
+[target.'cfg(target_arch = "x86_64")'.dependencies]
```
tjkirch, line 30: So if someone is doing a `cargo build` (so, separate from #1404 solutions), what happens here?  This dependency here would be skipped, but the code still references `vmw_backdoor::bla`, so I think it'd have compile errors about unknown references, which could be pretty confusing.  ARM developer machines are getting more common now, and vmware-dev is our first "local" variant, so I don't think this is "weird" enough to ignore.  Do we need to conditionalize the VmwareDataProvider on target_arch?

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -125,23 +134,146 @@ impl VmwareDataProvider {
...
-            Ok(None) => warn!("No user data found."),
+        match Self::guestinfo_user_data() {
+            Err(e) => {
+                error!("Error retrieving user data from guestinfo: {}", e);
+            }
+            Ok(None) => warn!("No user data found via guestinfo"),
             Ok(Some(s)) => output.push(s),
         }
+
+        // If we didn't find any user data in guestinfo, try CD-ROM
```
tjkirch: I'm curious what others think here.  If I specified user data in two locations, I'd expect both of them to be obeyed (in some order, but that's a different problem) rather than one taking precedence and ignoring the other.  Perhaps I have a common user data CD that I use in all launches, and then I use guestinfo for more specific settings in some launches, or whatever.

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -125,23 +134,146 @@ impl VmwareDataProvider {
...
 
 // =^..^=   =^..^=   =^..^=   =^..^=
 
+// Acceptable user data encodings
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum UserDataEncoding {
+    #[serde(alias = "b64")]
+    #[serde(alias = "base64")]
+    Base64,
```
tjkirch, line 269: I'm curious about the alias of "base64"; does that mean "Base64" and "base64" would both be accepted?  Should it be a rename instead of an alias, so only "base64" is accepted, or does VMware really allow those two alternate capitalizations?  I'm not sure how it interacts with `rename_all = "lowercase"` above - I would have expected that you wouldn't have needed either one, alias or rename, since the rename_all seems like it'd do the job on its own.

(Same question for the gzip version)

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -125,23 +134,146 @@ impl VmwareDataProvider {
...
+        let user_data_string = match user_data_encoding {
+            // gzip+base64 is gzip'ed user data that is base64 encoded
+            UserDataEncoding::Base64 | UserDataEncoding::GzipBase64 => {
+                info!("Decoding user data");
+                let mut reader = Cursor::new(user_data_bytes);
+                let mut decoder = base64::read::DecoderReader::new(&mut reader, base64::STANDARD);
+
+                let mut decoded_bytes = Vec::new();
+                decoder
+                    .read_to_end(&mut decoded_bytes)
```
tjkirch: It seems like it'd be simpler and more efficient to wrap an OptionalCompressionReader around a DecoderReader, rather than decoding base64 into a string and then expanding that?

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -125,23 +134,146 @@ impl VmwareDataProvider {
...
+            }
+            Ok(None) => warn!("No user data found via guestinfo"),
             Ok(Some(s)) => output.push(s),
         }
+
+        // If we didn't find any user data in guestinfo, try CD-ROM
+        if output.is_empty() {
+            match Self::cdrom_user_data() {
+                Err(e) => return Err(e).map_err(Into::into),
+                Ok(None) => warn!("No user data found."),
```
tjkirch: nit: might help to clarify that nothing was found via cdrom, like above it clarifies that nothing was found via guestinfo.

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -125,23 +134,146 @@ impl VmwareDataProvider {
...
+                error!("Error retrieving user data from guestinfo: {}", e);
+            }
+            Ok(None) => warn!("No user data found via guestinfo"),
             Ok(Some(s)) => output.push(s),
         }
+
+        // If we didn't find any user data in guestinfo, try CD-ROM
+        if output.is_empty() {
+            match Self::cdrom_user_data() {
+                Err(e) => return Err(e).map_err(Into::into),
```
tjkirch: It seems like you could `cdrom_user_data()?` (with .context if needed) and just handle None vs. Some rather than needing to separate out the Err case here and manually call Into when snafu could do that for you?

sources/api/early-boot-config/Cargo.toml
```diff
@@ -26,6 +27,10 @@ tokio = { version = "0.2", default-features = false, features = ["macros", "rt-t
...
+# vmw_backdoor includes x86_64 assembly, prevent it from building for ARM
+vmw_backdoor = "0.2"
```
tjkirch, line 32: Are you planning to do #1404 before merging this?  It'd be nice not to have a known (and awkward) build failure, and I don't think this is blocking anything else yet..?

sources/api/early-boot-config/src/provider/vmware.rs
zmrow: The cloud-init VMWare data provider accepts `base64` and `b64`, and the same goes for the `gzip` stuff.  The `alias` allows us to accept both.

I only added the `rename_all = lowercase` to be friendly and if a user had `Base64` we would rename it to `base64` and accept it.

sources/api/early-boot-config/src/provider/vmware.rs
zmrow: I think you're right - I was simply following an earlier pattern.

sources/api/early-boot-config/Cargo.toml
zmrow: #1431 is out to take care of #1404 and I plan to merge it before I merge this PR.

sources/api/early-boot-config/src/provider/vmware.rs
zmrow: I'll double check that

sources/api/early-boot-config/src/provider/vmware.rs
tjkirch: I understand the `b64` alias because it's a different name, but `base64`/`Base64` are the same other than capitalization, so it seems like the `rename_all` would take care of it for you without another alias here?  I'm wondering if line 266 can go, basically.  (and same for gzip)

zmrow: The `rename_all` happens first, so if a user entered `Base64` it would be lowercased to `base64`, and without the alias `base64` it wouldn't match any variants in the enum.

We could flip the logic around and remove the `rename_all` and just have aliases `base64` and `b64`.  That way we'll match the upper _and_ lowercase versions, as well as `b64`, but then we'd also need an alias for `B64`... haha :)

I think the way it is currently is safest and allows us to match everything with the least amount of `alias`es.

zmrow: Ok, I found that ^ wasn't true, `Base64` doesn't get accepted.  I'm going to remove the `rename_all` and add aliases.  That behavior feels the most straightforward to me and has the fewest surprises.

sources/api/early-boot-config/src/provider/vmware.rs
zmrow: I was able to check this by hand and the fallback path works as expected!

sources/api/early-boot-config/src/provider/vmware.rs
zmrow: Added some additional context

sources/api/early-boot-config/src/provider/vmware.rs
zmrow: Fixed

sources/api/early-boot-config/src/provider/vmware.rs
zmrow: @bcressey and I hacked around this a bit, and ended up removing the `rename_all` and going with aliases.  Once "case-insensitive" de/serialization is finalized, we should use that: https://github.com/serde-rs/serde/pull/1902 .  Added a comment to that effect.

zmrow:
^ Addressed all the comments and updated the testing!

sources/api/early-boot-config/Cargo.toml
tjkirch: With #1431 in, this is a lot less important.  I'm assuming we'd still have undefined references if `cargo build`ing this on a non-x86 host, so It'd be nice to do something to clean that up and make it clearer that it's not supported, but it's not as big an issue now since we'd expect the person to use `cargo make` to handle cross-compilation automatically.

sources/api/early-boot-config/src/provider/vmware.rs
tjkirch: This seems good now, but I'm wondering where you think we should document this?

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -0,0 +1,397 @@
...
+            }
+        }
+
+        // Base64 decode the &str
+        let decoded_bytes = base64::decode(&base64_str).context(error::Base64Decode {
+            base64_string: base64_str.to_string(),
+        })?;
+
+        // Decompress the data if it's compressed
+        let decoded = expand_slice_maybe(&decoded_bytes).context(error::Decompression {
```
tjkirch, line 135: Here's another spot we could chain the OptionalCompressionReader and DecoderReader.

One note though - the error is no longer strictly related to decompression - it could fail base64 too - so I think we'd want a new variant that's a little more generic, maybe UserDataInput or something?

sources/api/early-boot-config/src/provider/vmware.rs
zmrow: Creating an issue for that now.  We can probably comment on it here, but I the "main" documentation should happen in the variant's top level `README` unless you think it belongs somewhere else.

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -0,0 +1,397 @@
...
+        } else {
+            trace!(
+                "Received long user data, starts with: {}",
+                // (this isn't perfect because chars aren't grapheme clusters, but will error
+                // toward printing the whole input, which is fine)
+                String::from_iter(user_data_str.chars().take(2048))
+            );
+        }
+
+        let json = SettingsJson::from_toml_str(&user_data_str, "user data from CD-ROM").context(
```
tjkirch, line 90:
```suggestion
        let json = SettingsJson::from_toml_str(&user_data_str, "user data from CD-ROM").with_context(||
```
nit: so we don't do the string conversion unnecessarily.

tjkirch (approved):

bcressey (approved):

PR #1397, opened 2021-03-16
vmware-dev: Change image format to vmdk

zmrow:
**Description of changes:**
This change makes the default image format for the `vmware-dev` variant VMDK. 

**Testing done:**
Built and booted said VMDK images. :)

**Terms of contribution:**

### DIFF
--- a/variants/vmware-dev/Cargo.toml
+++ b/variants/vmware-dev/Cargo.toml
@@ -8,6 +8,7 @@ build = "build.rs"
 exclude = ["README.md"]
 
 [package.metadata.build-variant]
+image-format = "vmdk"
 included-packages = [
 # core
     "release",


webern (approved):

variants/vmware-dev/Cargo.toml
```diff
@@ -8,6 +8,7 @@ build = "build.rs"
 exclude = ["README.md"]
 
 [package.metadata.build-variant]
+image-format = "vmdk"
```
tjkirch, line 11: I think this means that local testing in qemu would need a conversion to raw/qcow2 first, since you mentioned that VMDKs only run in read-only mode.  (I'm thinking it's easier for some people to run in qemu than vmware locally.)

I'm wondering if we should make this a list instead of a single entry, so we could say `["raw", "vmdk"]` here and get the benefits of both?

bcressey: The alternative format is `img.lz4`, which also needs to be "converted" before it can be used under QEMU. I'd rather keep this as the primary format and let people convert to others as needed.

tjkirch: > I'd rather keep this as the primary format and let people convert to others as needed.

Including those that need vmdk?  (Or does "this" mean vmdk?)

jpculp: If we went with a list, we could add a third option `raw_uncompressed` to represent `img` for testing.

jpculp (approved):

variants/vmware-dev/Cargo.toml
bcressey: > Including those that need vmdk?  (Or does "this" mean vmdk?)

I did mean "vmdk" - we should output the (compressed) format that pubsys expects, and let other use cases manipulate it as needed.

Issue #1302, opened 2021-02-06
kubernetes: bottlerocket doesn't understand the "Exists" taint


diranged:
**Platform I'm building on:**
Bottlerocket 1.0.5 on EKS 1.18

**What I expected to happen:**
Kubernetes has an `Exists` taint that has no value key in it. So the CLI argument looks like this:

```
kubelet ... --register-with-taints=MyKey:NoSchedule
```

and a matching Toleration would look like this:

```
tolerations:
  - operator: Exists
    key: MyKey
    effect: NoSchedule
```

However, there is no way to put that into the `settings.kubernetes.node-taints` configuration because the pattern is hard-coded to expect an `Equals` taint: https://github.com/bottlerocket-os/bottlerocket/blob/b5fd55f22d31eea87f2fa4b9491aeb73c12498a2/packages/kubernetes-1.18/kubelet-env#L3

**What actually happened:**
I could not boot up nodes - they failed to parse the key and threw an error.

etungsten:
Hi @diranged, 

The `--register-with-taints` argument to kubelet takes a list of taints of the form ` <key1>=<value1>:<effect1>,<key2>=<value2>:<effect2>`. The equal sign is a necessary part of the separators. (See documentation for this argument [here](https://kubernetes.io/docs/concepts/architecture/nodes/#self-registration-of-nodes) and [here](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/)).

The problem is that the validation we have for KubernetesTaintValue currently doesn’t allow for an empty taint value. 
https://github.com/bottlerocket-os/bottlerocket/blob/fdb0353df12fa3f710628c81fe3a79aa98d42e6c/sources/models/src/modeled_types/kubernetes.rs#L206-L236

This is being fixed in #1406. Once that’s merged, you should be able to specify taints without values and use `Exists` tolerations. For example, in userdata:

```
[settings.kubernetes.node-taints]
MyKey = ":NoSchedule"
SomeOtherKey = "SomeValue:NoSchedule"
```

PR #1406, opened 2021-03-19
models: make kubernetes taint values optional

etungsten:
**Issue number:**
Addresses https://github.com/bottlerocket-os/bottlerocket/issues/1302

**Description of changes:**
```
Date:   Fri Mar 19 11:55:17 2021 -0700

    models: make kubernetes taint values optional
    
    This modifies the validation for KubernetesTaintValue to make the taint
    value optional for 'Exists' toleration operators.

```

**Testing done:**
Built an aws-k8s-1.19 x86_64 AMI and launched it with the following specified in my userdata:
```
[settings.kubernetes.node-labels]
label1 = "foo"
label2 = "bar"
[settings.kubernetes.node-taints]
testMe = ":NoSchedule"
```

The node comes up successfully in my cluster with the specified labels and taints:
```
$ kubectl get nodes --output json
...
                "taints": [
                    {
                        "effect": "NoSchedule",
                        "key": "testMe"
                    }
                ]

```

I'm also able to create tolerations that uses the `Exists` operator in my pod specs:
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    env: test
spec:
  containers:
  - name: nginx
    image: nginx
    imagePullPolicy: IfNotPresent
  tolerations:
  - key: "MyKey"
    operator: "Exists"
    effect: "NoSchedule"
```

...and the nginx pod does not get scheduled on the node that has the `testMe` taint as expected.
```
$ kubectl get pod nginx --output json
...
    "status": {
        "conditions": [
            {
                "lastProbeTime": null,
                "lastTransitionTime": "2021-03-19T20:35:20Z",
                "message": "0/1 nodes are available: 1 node(s) had taint {testMe: }, that the pod didn't tolerate.",
                "reason": "Unschedulable",
                "status": "False",
                "type": "PodScheduled"
            }
        ],
        "phase": "Pending",
        "qosClass": "BestEffort"
    }

```

I launched another node without the taint, and the nginx pod gets scheduled onto that node as soon as it became Ready.

If I launch another node with a `MyKey=:NoSchedule` taint, the pod correctly tolerates it and is able to schedule onto that node as well.

**Terms of contribution:**

### DIFF
--- a/sources/models/src/modeled_types/kubernetes.rs
+++ b/sources/models/src/modeled_types/kubernetes.rs
@@ -223,12 +223,14 @@ pub struct KubernetesTaintValue {
 lazy_static! {
     pub(crate) static ref KUBERNETES_TAINT_VALUE: Regex = Regex::new(
         r"(?x)^
-       [[:alnum:]]  # at least one alphanumeric
        (
-           ([[:alnum:]._-]{0,61})?  # more characters allowed in middle
-           [[:alnum:]]  # have to end with alphanumeric
-       )?
-       :  # separate the label value from the effect
+          [[:alnum:]]  # values have to start with alphanumeric if they're specified
+          (
+             ([[:alnum:]._-]{0,61})?  # more characters allowed in middle
+             [[:alnum:]]  # values have to end with alphanumeric
+          )?  # only the first alphanumeric is required, further chars optional
+       )? # the taint value is optional
+       :  # separate the taint value from the effect
        [[:alnum:]]{1,253}  # effect
    $"
     )
@@ -266,6 +268,9 @@ mod test_kubernetes_taint_value {
             "value:NoSchedule",
             "value:PreferNoSchedule",
             "value:NoExecute",
+            ":NoSchedule",
+            "a:NoSchedule",
+            "a-b:NoSchedule",
         ] {
             KubernetesTaintValue::try_from(*ok).unwrap();
         }
@@ -273,7 +278,15 @@ mod test_kubernetes_taint_value {
 
     #[test]
     fn bad_values() {
-        for err in &[".bad", "bad.", &"a".repeat(254), "value:", ":effect"] {
+        for err in &[
+            ".bad",
+            "bad.",
+            &"a".repeat(254),
+            "value:",
+            ":",
+            "-a:NoSchedule",
+            "a-:NoSchedule",
+        ] {
             KubernetesTaintValue::try_from(*err).unwrap_err();
         }
     }


zmrow (approved):
:volleyball:

tjkirch (commented):
I believe this would now mistakenly require the value to be two characters minimum, but I think one is intended to be allowed. Would you please add a test using a single character?

etungsten:
> I believe this would now mistakenly require the value to be two characters minimum, but I think one is intended to be allowed. Would you please add a test using a single character?

You're absolutely right. :facepalm:. It'll require minimum 2 alphanumeric characters if the value is specified. I'll fix it.

etungsten:
Push above addresses @tjkirch 's comment https://github.com/bottlerocket-os/bottlerocket/pull/1406#pullrequestreview-616841936:

- Fixes regex so that if a taint value is specified, it needs at least 1 alphanumeric character, and end with an alphanumeric character.
- Adds more unit tests

All unit tests pass.

zmrow (approved):
:hugs:

bcressey (approved):

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -223,12 +223,12 @@ pub struct KubernetesTaintValue {
 lazy_static! {
     pub(crate) static ref KUBERNETES_TAINT_VALUE: Regex = Regex::new(
         r"(?x)^
-       [[:alnum:]]  # at least one alphanumeric
+       ([[:alnum:]]  # values have to start with alphanumeric if they're specified
```
tjkirch: nit: I think the existing pattern of formatting is clearer, i.e. grouping on its own line, because it's quicker to see which things are tied together and which are required, and you can document each grouping clearly.  Something like:
```
(
   [[:alnum:]]  # values have to start with alphanumeric if they're specified
   (
      ([[:alnum:]._-]{0,61})?  # characters allowed in middle
      [[:alnum:]]  # values have to end with alphanumeric
   )?  # only the first alphanumeric is required, further chars optional
)?  # the whole taint value is optional
:
```

tjkirch (approved):

etungsten:
Push above addresses @tjkirch 's comment: https://github.com/bottlerocket-os/bottlerocket/pull/1406#discussion_r598935557

Ran unit tests again and they passed.

webern (approved):

tjkirch (approved):

PR #1408, opened 2021-03-23
build only the packages needed for the current variant

### LINKED ISSUE #1361: only build the packages needed for the current variant
**What I'd like:**

We have a build-packages phase that builds all packages, even those that aren't needed for the variant currently being built. See if we can build only the packages that we need for the current variant.

Leaving it the way it is... but this becomes more problematic as we have more large packages (multiple kubernetes versions, or kernels, for example).

webern:
**Issue number:**

Closes #1361

**Description of changes:**

Utilize cargo dependency graph to build only the packages we need for the current variant.

Packages express a dependency on anything that the rpm spec file declares a dependency on. (We already did this for `BuildRequires` but now we do it for `Requires` as well.) Additionally, the variant's `Cargo.toml` needs to express the things the package crates it depends on.

**Note:** We now rely on a change in behavior in cargo 1.51.0 to allow us to build a package that is a dependency of a workspace, but which is not a workspace member. You will need to `rustup update` if you have an older version.

```
    build: refactor os package dependencies

    Signed-off-by: Ben Cressey <bcressey@amazon.com>

    Change os.spec such that all first party code is installed when the os
    package is installed. Change release.spec such that it requires only the
    os package for first-party code instead of requiring each individual
    first-party package.

    This consolidates the package dependencies for first-party code into
    one place, instead of picking some of the packages in "release" and
    others through the variant definition.

    This is part of #1361. We want to better manage package dependencies
    with cargo so that we can build only the packages needed for a given
    variant, and this change eliminates some one-off first-party package
    dependencies that needed to be expressed at the variant level.
```

```
    packages: express installation dependencies in cargo

    Previously we expressed RPM BuildRequires dependencies in Cargo's
    dependency graph to ensure the necessary RPM's exist before we build
    a package. If we add to this RPM Requires dependencies (dependencies
    that are needed when software packages are installed), then we can use
    Cargo to select only the packages that are needed for a set of desired
    install packages.
```

```
    build: only build packages for current variant

    Adds the packages that are needed to build a variant to the variant's
    Cargo dependencies. Previously the build-variant makefile target assumed
    that all packages were pre-built. Instead we now tell Cargo to build the
    packages we need for the variant.

    Creates a workspace in the variants directory and removes the workspace
    in the packages directory. Updates the makefile accordingly. Now only
    the packages that are part of the dependency tree starting with the
    variant will be built.

    The build-kmod-kit target previously depended on build-packages, which
    is no longer available. So a new target, build-package, and a
    specialization of it, build-kernel, have been added. These use the same
    workspace context that is used when building variants to ensure that the
    same rpm is used by build-kmod-kit.
```

```
    makefile: remove world targets

    cargo make world targets were vestigial. Probably nobody is using them
    so we cleaned them up.
```

**Testing done:**

Starting from a clean build state:

- ran `cargo make -e BUILDSYS_VARIANT=aws-k8s-1.18`
  - Observed that only k8s 1.18 was built (other k8s versions were not)
  - build succeeded
- Without cleaning, ran `cargo make -e BUILDSYS_VARIANT=aws-k8s-1.19`
  - Observed that existing packages (other than os and release) were not rebuilt.
  - Saw that k8s 1.19 was built.
- Without cleaning, ran `cargo make -e BUILDSYS_VARIANT=aws-k8s-1.18 repo`
  - Observed nothing was rebuilt.
- Without cleaning, ran `cargo make -e BUILDSYS_VARIANT=aws-k8s-1.19 repo`
  - Observed that nothing was rebuilt.
- `find . -type d -name target`
  - `./tools/target`
  - `./variants/target`

Starting from a clean build state:

- ran `cargo make build-kmod-kit`
- Observed the kernel being built.
- target succeeded
- `find . -type d -name target`
  - `./tools/target`
  - `./variants/target`

Starting from the above kmod-kit build state:

- Ran an x86_64 aws-k8s-1.17 variant and ran a pod.
- Ran an aarch64 aws-ecs-1 variant and ran a task.
- Ran an aarch64 aws-dev variant and logged in.

**Terms of contribution:**


webern:
I'll do some squashing and try to figure out the best thing for the kmod kit target today.

Makefile.toml
```diff
@@ -261,21 +261,10 @@ cargo install \
...
-  --manifest-path packages/Cargo.toml \
-  --all
-'''
-]
-
 [tasks.build-kmod-kit]
-dependencies = ["build-packages"]
+# FIXME - oops, it's weird to build a variant just to get the kernel package built
+# FIXME - maybe a variant's kernel package should become a first-class concept so it could be built here
+dependencies = ["build-variant"]
```
webern: I think what I want to do here is create a `build-kernel` target and call it here instead of `build-packages` or `build-variant`.

webern:
> webern force-pushed the webern:package-build branch from e428c02 to c49af80

This squashes things into a reasonable set of commits.

TODOs
- It looks like the above push has an unintentional Cargo.lock change
- Still need to do the kmod-kit fixme

webern:
> webern force-pushed the webern:package-build branch from c49af80 to 8779092

Fix an errant lock file.

webern:
> webern force-pushed the webern:package-build branch from 8779092 to 98a9439

Builds the kernel as a dependency of build-kmod-kits, which previously relied on less specific prerequisites.

Makefile.toml
```diff
@@ -354,9 +366,6 @@ cargo build \
 '''
 ]
 
-[tasks.world-packages]
-alias = "build-packages"
-
 [tasks.world-variant]
 alias = "build-variant"
```
bcressey: We should clean up the other "world" aliases; they were for muscle-memory compatibility for the Makefile from aeons past.

Makefile.toml
```diff
@@ -261,21 +261,33 @@ cargo install \
...
 '''
 ]
 
+[tasks.build-kernel]
+env = { "PACKAGE" = "kernel" }
+run_task = "build-package"
+
 [tasks.build-kmod-kit]
-dependencies = ["build-packages"]
+dependencies = ["build-kernel"]
```
bcressey, line 295: This is a bit of a mess and it's my fault.

When we introduce the 5.10 kernel, we'll need two of these kmod kits:
* one that has the toolchain + 5.4 kernel files
* another that has the toolchain + 5.10 kernel files

In this new approach, we'd either need a way to express a dependency on something cargo should build that is not a package or an image, or else we'd need to create the kmod kit as a side effect of building the variant - like we do today with migrations.

Maybe a top-level `archives` directory, an ArchiveBuilder component in `buildsys`, where we expect an `archives/{name}/{name}` script?

I didn't use the `rpm2migrations` approach for the kmod kit because I think it's ugly, and also because it requires network access and `buildsys` invoked stuff is forbidden to touch the network.

There are three related use cases, all of which involve post-processing RPMs:
* find the kernel-devel RPM, unpack it, repackage its contents
* find the migrations RPM and do the same thing
* find all debuginfo RPMs and extract the symbols

The debuginfo one has the potential to be pretty slow; we might implement that as a "side effect" for building a package so that we don't batch it all up at the end.

That would incline me toward the "side effect" treatment of the kmod kit also. I'm planning a separate change to package the toolchain archives as containers in their own right, which would remove the network dependency issue.

packages/cni-plugins/Cargo.toml
```diff
@@ -14,3 +14,4 @@ sha512 = "24e8fcedbff2ae7a83aa96085b546b164de6a0884d593e3b5386e9d2de3c4d9a215db9
 
 [build-dependencies]
 glibc = { path = "../glibc" }
+iptables = { path = "../iptables" }
```
bcressey, line 21: Would it work to make these regular `dependencies`, and to use that instead of `build-dependencies` for the variant also?

That might be more clear. Otherwise we lose some information here about the `BuildRequires` vs `Requires` distinction.

packages/release/Cargo.toml
```diff
@@ -7,3 +7,26 @@ build = "build.rs"
...
+chrony = { path = "../chrony" }
+coreutils = { path = "../coreutils" }
+dbus-broker = { path = "../dbus-broker" }
+e2fsprogs = { path = "../e2fsprogs" }
+filesystem = { path = "../filesystem" }
+glibc = { path = "../glibc" }
+grub = { path = "../grub" }
+host-ctr = { path = "../host-ctr" }
+iproute = { path = "../iproute" }
+kernel = { path = "../kernel" }
```
bcressey, line 27: This can wait for the 5.10 kernel to come in, but we'd want to move this to the variant's dependencies rather than release, since it'll no longer be the same for all variants.

variants/aws-dev/Cargo.toml
```diff
@@ -28,3 +28,21 @@ included-packages = [
```
bcressey, line 28: I am sad about the duplication here. Could we find the list of direct dependencies somehow instead?

variants/aws-dev/Cargo.toml
```diff
@@ -28,3 +28,21 @@ included-packages = [
 
 [lib]
 path = "lib.rs"
+
+[build-dependencies]
+# core
+os = { path = "../../packages/os" }
```
bcressey: It's weird that `os` shows up in this list now. Isn't it pulled in by the dependency on `release`?

variants/aws-dev/Cargo.toml
webern: You mean the duplication between our own metadata in `included-packages` and Cargo's dependency list? Unfortunately it's not exact duplication. In `included-packages` we're telling buildsys what to install, and in Cargo's list we're telling Cargo what has to be built in order for those packages to be available. It's not always 1:1.

There are some designs we could consider to cut down on the duplication.
- Introduce a package/spec for each variant, e.g. `aws-dev.spec` where Cargo.toml would give the dependencies for Cargo and the spec would say `Requires` for each of the dependencies. (?)
- Some kind of preprocessing of the variant's Cargo.toml to figure out what to build during the variant's `build.rs`? I'm not sure how this would work, and I'm not sure how buildsys would understand when a subpackage comes from a parent package.

Any ideas?

packages/cni-plugins/Cargo.toml
webern: Yeah, that makes sense. It also 'relaxes' our instructions to Cargo about the build graph to something more representative of reality.

webern:
> webern force-pushed the webern:package-build branch from 98a9439 to c9684e9

Remove os from the variant dependency lists since it is a dependency of release. (Leaning on CI to prove it.)

webern:
> webern force-pushed the webern:package-build branch from c9684e9 to 2c415ff

Separate package dependencies between `build-dependencies` and `dependencies` (to represent RPM `BuildRequires` and `Requires`).

webern:
> webern force-pushed the webern:package-build branch from 2c415ff to 6a000ce

Fix a couple of mistakes in the `Requires` vs `BuildRequires` distinctions.

webern:
Added a commit to remove `cargo make world` and `cargo make world-variant`.

packages/release/Cargo.toml
webern: Agreed. The release package has consistently struck me as an abstraction worth poking at/questioning as I've been getting more adept with the build. I have a feeling that, as variants become more specialized, the number of things in release may reduce over time.

Makefile.toml
webern: Should we create an issue for this? I think I would like to *not* build the kmod-kits and debug symbols most of the time. I agree that `rpm2migrations` is not something we want to emulate for future 'archives' that don't go into the Bottlerocket image.

variants/aws-dev/Cargo.toml
bcressey: It looks like the exceptions all come from subpackages of `os.spec`, so one alternative would be for the `os` package to conditionally include its own subpackages based on the variant.

I didn't like the idea at first, but we're already conditionalizing the building of those components and their subpackages within `os.spec`, so it isn't making things any more weird than they already are. It would also thin out dependencies in `release.spec` quite a bit - it'd just depend on `os` - and remove some of the coupling between those two.

We'd end up with an implicit rule like "all main packages must require all of their own subpackages that they need at runtime."

bcressey:
One of the side effects of this change is that modifying certain packages (systemd) will trigger many more package rebuilds than they previously did. It might be worth exploring the graph with `cargo tree` to see if this is going to be painful for the developer workflow in common cases.

I'd say we might want to consider removing containerd's dependency on systemd, and the kubelet + docker-engine dependencies on containerd. We know both will end up on the final image by way of the release package, and they aren't needed at build time.

Makefile.toml
```diff
@@ -261,21 +261,33 @@ cargo install \
...
 dependencies = ["build-tools", "publish-setup"]
+script_runner = "bash"
 script = [
 '''
+set -e
+if [ -z "${PACKAGE}" ]; then
+    echo "the PACKAGE environment variable must be set. for example:"
+    echo "cargo make build-package -e PACKAGE=kernel"
+    exit 1
+fi
```
bcressey: Consider this scenario. Note that I haven't tested this, so I may be wrong in some of the specifics, but I expect I'm right about the larger issue.
```
# Get to a fresh state
cargo make clean

# builds the kernel and os package for the 1.18 variant crate
cargo make -e BUILDSYS_VARIANT=aws-k8s-1.18

# rebuilds the kernel and os packages for the 1.19 variant crate,
# since cargo doesn't know the RPMs exist yet, even though the
# kernel package is not variant-specific.
cargo make -e BUILDSYS_VARIANT=aws-k8s-1.19

# builds the kernel package for the packages workspace,
# resulting in a new RPM build that doesn't match what we
# used for either 1.18 or 1.19.
cargo make -e BUILDSYS_VARIANT=aws-k8s-1.18 repo

# reuses the kernel package build from the packages workspace.
cargo make -e BUILDSYS_VARIANT=aws-k8s-1.19 repo

# reuses the kernel and os package builds for the 1.19 crate,
# which is incorrect since "kernel" has been rebuilt, possibly
# with a different config, and more critically because "os" has
# been rebuilt. cargo won't trigger a rebuild since none of
# the source files have changed and the environment matches
# what it was for the last 1.18 build.
cargo make -e BUILDSYS_VARIANT=aws-k8s-1.18

# reuses the kernel and os package builds for the 1.19 crate,
# which is incorrect since "kernel" has been rebuilt, possibly
# with a different config.
cargo make -e BUILDSYS_VARIANT=aws-k8s-1.19
```

The underlying problem is that we have only a single cache directory for RPMs, but `cargo` has separate views of that cache depending on which workspace or crate was last built.

I am not sure of the best way to fix it. The `build-package` problem may not be solvable if we take the position that the kernel package that we use to generate the kmod kit - or more generally, any package that contributes a repo artifact - must exactly match the package we used to build the image. Essentially this means the separate packages workspace might have to go away, leaving us with just the variant crates.

For detecting that "os" RPMs have been silently rebuilt, it might work to emit `rerun-if-changed` directives on the RPM and image outputs. I am not sure it's sufficient but it's definitely an improvement.

Otherwise we might have to unify all variants back into a single workspace, so that there's no mismatch between cargo's view of the most recent builds and the contents of our `build` directory.

webern: Ugh, I don't quite follow but I'll look more closely at this. It's hard to wrap my head around which system governs rebuilding since there are four (?) competing systems governing build caching vs rebuilding.
- cargo
- rpm
- docker buildkit
- buildsys

In my mind everything would 'just work' because cargo doesn't know about rpms and only rebuilds based on a variant change if we tell it to. Cargo has the ultimate veto power in terms of re-building something, no? I mean if cargo doesn't need to rebuild the empty lib.rs (actually pkg.rs) then it won't rebuild it right?

webern: P.S. rpms that change how/what they build based on variant wasn't something I was considering.

webern: I think I understand the problem now.

The flaw in my thinking was that I didn't account for the fact that we used to have a shared `target` directory for all packages because we were using the `packages` workspace. In my current solution we have different `target` directories for each variant, and a different one still when we use the packages workspace to `build-kernel` for the kmod kit.

webern:
> webern force-pushed the webern:package-build branch from 46d1fce to 35e636e

A rebase that git was surprisingly angry about and therefore might need more work. We'll see.

webern:
I'm going to switch this back to draft for now since my next push will be somewhat of a proposed solution.

webern:
After a push that didn't work out, then a revert... this push takes care of a surly rebase over develop

> webern force-pushed the webern:package-build branch from 35e636e to ea32575

webern:
It seems the GitHub UI doesn't show the fact that I just pushed a new commit: 0caaf138909e83b84a7c264083523f725c133e51

In this commit I combine package and variant workspaces into a single workspace so that they have a shared target directory. This should solve the problems @bcressey detected here https://github.com/bottlerocket-os/bottlerocket/pull/1408#discussion_r602908420

There's something no one is going to like... I had to `mv packages variants/packages`. I tried the following alternate solutions which did not work:
- A Cargo.toml workspace file in variants that reaches up and over to packages with `../packages/acip`. Cargo didn't like this, the error was "workspace member  is not hierarchically below the workspace root".
- I tried creating a symlink at `variants/packages` that points to `packages`. Cargo didn't like this because it produced conflicting file paths for its lock file: "error: package collision in the lockfile: packages acpid v0.1.0 (/Users/someone/repos/br/packages/acpid) and acpid v0.1.0 (/Users/someone/repos/br/variants/packages/acpid) are different, but only one can be written to lockfile unambiguously"
- I tried creating the workspace at the top level of the git repo, i.e. 'above' both variants and packages, but this didn't work because `cargo make` has a feature that detects workspaces and behaves differently when they are detected: https://sagiegurari.github.io/cargo-make/#usage-workspace-support

So for now what I have pushed is something that would work if we could live with the moving of the packages directory.

webern:
> webern force-pushed the webern:package-build branch from 0caaf13 to eca1de3

In this version packages are no longer members of a workspace. This allows us to leave the packages directory alone instead of moving it into the variants workspace.

Makefile.toml
webern: This is now fixed in 76cb1ad by pointing a cargo build to the workspace Cargo.toml

webern:
@bcressey this is ready for re-review. I think there might be more you wanted to do with moving some deps from the variant's cargo.toml to common release.spec dependencies. But everything else is done.

packages/release/Cargo.toml
```diff
@@ -7,3 +7,31 @@ build = "build.rs"
...
+acpid = { path = "../acpid" }
+ca-certificates = { path = "../ca-certificates" }
+chrony = { path = "../chrony" }
+coreutils = { path = "../coreutils" }
+dbus-broker = { path = "../dbus-broker" }
+e2fsprogs = { path = "../e2fsprogs" }
+filesystem = { path = "../filesystem" }
+glibc = { path = "../glibc" }
+grub = { path = "../grub" }
+host-ctr = { path = "../host-ctr" }
```
bcressey: nit: we get this transitively via `os`:
```suggestion
```

variants/aws-dev/Cargo.lock
```diff
@@ -1,5 +1,504 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
+[[package]]
+name = "acpid"
+version = "0.1.0"
```
bcressey: We can delete these individual Cargo.lock files from the variant directories, since they should be superseded by the workspace lock.

Makefile.toml
```diff
@@ -266,21 +266,33 @@ cargo install \
...
 export PATH="${BUILDSYS_TOOLS_DIR}/bin:${PATH}"
 cargo build \
   ${CARGO_BUILD_ARGS} \
   ${CARGO_MAKE_CARGO_ARGS} \
-  --manifest-path packages/Cargo.toml \
-  --all
+  --manifest-path "variants/Cargo.toml" \
+  --package "${PACKAGE}"
```
bcressey, line 286: Does this work? In my scratch workspace, it seems like only packages that are part of the workspace are valid targets.

```
fedora on ip-10-0-60-129 in ~/scratch/ugly 
❯ cargo build --manifest-path variants/Cargo.toml --package foo
error: package ID specification `foo` matched no packages

fedora on ip-10-0-60-129 in ~/scratch/ugly 
❯ cargo build --manifest-path variants/Cargo.toml --package bar
error: package ID specification `bar` matched no packages

fedora on ip-10-0-60-129 in ~/scratch/ugly 
❯ cargo build --manifest-path variants/Cargo.toml --package baz
   Compiling foo v0.1.0 (/home/fedora/scratch/ugly/packages/foo)
   Compiling bar v0.1.0 (/home/fedora/scratch/ugly/packages/bar)
   Compiling baz v0.1.0 (/home/fedora/scratch/ugly/variants/baz)
    Finished dev [unoptimized + debuginfo] target(s) in 0.76s
```

webern: I think it did work when I ran cargo make build-package -ePACKAGE=systemd and cargo make build-kernel. Though I can’t explain why that would work or why I expected it to. I’ll take a closer look. Maybe it works so long as —package is a transient dependency of a workspace? 🤔 (edit: looks like your scratch workspace disproves the transient dependency theory)

webern:
> webern force-pushed the webern:package-build branch from 0d27970 to f684ba7

Remove the individual variant lock files since they are part of a workspace now.

Makefile.toml
bcressey: This ended up being a difference in behavior between `cargo` from 1.49.0 and 1.51.0 - the newer version works fine.

webern:
>  webern force-pushed the webern:package-build branch from f684ba7 to 53bda81

- Remove host-ctr from release's Cargo.toml dependencies since it is obtained through os
- Note in BUILDING.md that you now need Rust 1.51.0 or higher.

webern:
> webern force-pushed the webern:package-build branch from 53bda81 to 2037bd3

rebase

oops, unsquashed a couple of fixup commits

webern:
> webern force-pushed the webern:package-build branch from 2037bd3 to 5208962

resquash a couple of fixups

webern:
Failed builds, I guess I need to add host-ctr to os's Cargo.toml.

webern:
> webern force-pushed the webern:package-build branch from 5208962 to bc53009

Add host-ctr to os Cargo.toml 🤞

webern:
> webern force-pushed the webern:package-build branch from bc53009 to 5fcc3cf 

Fix Cargo.lock files in all affected commits 🤞

packages/os/os.spec
```diff
@@ -197,7 +211,6 @@ Summary: Dynamic setting generator for kubernetes
 
 %package -n %{_cross_os}static-pods
 Summary: Manages user-defined K8S static pods
-Requires: %{_cross_os}apiserver = %{version}-%{release}
 %description -n %{_cross_os}static-pods
 %{summary}.
 %endif
```
webern, line 217: Can't comment on line 220... rebase pulled in bootstrap-containers which breaks the pattern.

bcressey: >Requires: %{_cross_os}bootstrap-containers

This should move from `release.spec` to the main package in `os.spec`, as part of our co-authored commit.

webern:
> webern force-pushed the webern:package-build branch from 5fcc3cf to 1ca16f1

Add bootstrap-containers to the os/release package dependency refactor.

Edit: oops bad push

webern:
> webern force-pushed the webern:package-build branch from 1ca16f1 to 5fcc3cf

undo bad push

webern:
> webern force-pushed the webern:package-build branch from 5fcc3cf to 4477e17

Add bootstrap-containers to the os/release package dependency refactor (without undoing other work this time).

packages/os/os.spec
webern: [fixed](https://github.com/bottlerocket-os/bottlerocket/compare/5fcc3cf03741eae766490d027398585e21a5701f..4477e171823e0947f5b085a6ccf795c337832b99)

bcressey (approved):

webern:
> webern force-pushed the webern:package-build branch from 4477e17 to c0b6d76

Very minor rebase

webern:
> webern force-pushed the webern:package-build branch from c0b6d76 to 59dd076 

Attempt at README correctness.

packages/README.md
```diff
@@ -42,7 +42,8 @@ Bottlerocket uses Docker and containers to accomplish this instead.
 Every package build starts from a container with the [Bottlerocket SDK](https://github.com/bottlerocket-os/bottlerocket-sdk) and zero or more other packages installed as dependencies.
 Any source archives and patches needed for the build are copied in, and the binary RPMs are copied out once the build is complete.
 
-Some packages depend on building other packages first.
+Some packages depend on building other packages first and some packages depend on other packages being available when they are installed.
+In RPM spec (see example below) these expressed as `BuildRequires:` and `Requires:` respectively.
```
tjkirch:
```suggestion
In RPM spec (see example below) these are expressed as `BuildRequires:` and `Requires:` respectively.
```

packages/README.md
```diff
@@ -42,7 +42,8 @@ Bottlerocket uses Docker and containers to accomplish this instead.
 Every package build starts from a container with the [Bottlerocket SDK](https://github.com/bottlerocket-os/bottlerocket-sdk) and zero or more other packages installed as dependencies.
 Any source archives and patches needed for the build are copied in, and the binary RPMs are copied out once the build is complete.
 
-Some packages depend on building other packages first.
+Some packages depend on building other packages first and some packages depend on other packages being available when they are installed.
```
tjkirch:
```suggestion
Some packages depend on building other packages first, because they're used directly by the build process.
Some packages depend on other packages being available when they're installed, because they're used dynamically at runtime.
```

packages/README.md
```diff
@@ -101,10 +107,17 @@ These files are, by default, only fetched from our upstream source mirror, using
...
 
-The [build-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#build-dependencies) table is used to declare a dependency on another crate.
+The [build-dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#build-dependencies) table is used to declare a build-time dependency on another package.
 In this case, `libwoof` depends on `glibc` and `libseccomp`.
 We want those libraries to be built first, and for this one to be rebuilt when they are modified.
 
+We use the [dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) section to specify any packages that are needed at runtime (that is, RPM `Requires:`).
+Although we don't need these to be built *before* the current package, we do need them to be built before we construct the Bottlerocket image.
+With build-dependencies (`BuildRequires:`) it's important that they be built *before* the current package is built.
+Runtime dependencies (`Requires`) can be built at any time.
```
tjkirch:
```suggestion
```
I think it's better to give a brief introduction to these concepts earlier, as I suggested above, so the reader isn't confused about why there's a distinction.  Then it wouldn't need to be repeated here.

packages/README.md
```diff
@@ -101,10 +107,17 @@ These files are, by default, only fetched from our upstream source mirror, using
...
+The separation between `build-dependencies` and `dependencies` in Cargo.toml is not necessary but helps us distinguish between the two dependency types.
```
tjkirch:
```suggestion

The separation between `build-dependencies` and `dependencies` in Cargo.toml is not strictly necessary, because either way we'll get the package builds we need, but it helps readers distinguish between the two dependency types and correlate them to the RPM .spec file.
```
I think a blank line helps separate the facts above from this background material, and I suggested some more background.

variants/README.md
```diff
@@ -121,11 +121,17 @@ included-packages = [
...
+[build-dependencies]
+"my-agent" = { path = "../../packages/my-agent" }
+"release" = { path = "../../packages/release" }
 ```
 
 The [package.metadata](https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table-optional) table is ignored by Cargo and interpreted by our `buildsys` tool.
 
 It contains an `included-packages` list which specifies the packages to install when building the image.
+In the `[build-dependencies]` section, we specify the packages that need to be built, which is sometimes slightly different than `included-packages`.
+This populates the Cargo build graph with all of the RPM packages that need to be built before the variant can be constructed.
```
tjkirch, line 135: I'm finding these new sentences a bit confusing.  build-dependencies are rather different than package.metadata, but they're stuck in the middle of that paragraph - can we separate it?  (Unless they're always intended to match, in which we should probably make *less* of a distinction.)  It also has the effect of making it unclear whether the sentence just after this about including the release package applies to included-packages, build-dependencies, or both.  (Either way... why?)

And I have to ask... if they need to match, can we get rid of included-packages and have buildsys read the names of the build-dependencies instead?  I see there are a couple package names that differ, but I bet we can come up with a solution for that rather than entirely duplicating it.  For example, I just tested and found that you can add extra keys to dependencies for your own purposes; we could do something like `chrony = { path = "../../packages/chrony", included_rpms = "chrony-tools" }` for those special cases.

I'm not sure we need the technical wording here about the build graph, unless it's important to explain the difference between included-packages and build-dependencies.  I think the audience for this doc is someone trying to build a variant, not work on the build system.

packages/README.md
```diff
@@ -101,10 +107,17 @@ These files are, by default, only fetched from our upstream source mirror, using
...
+Note that we do not include `Requires:` dependencies if they are already satisfied by either the current package, or the build-dependencies.
```
tjkirch: I'm not sure I understand this.  What do you mean by satisfied by the current package?

I think the biggest potential confusion we have is from someone adding or updating a package who doesn't understand what needs to go in Requires vs. dependencies, BuildRequires vs. build-dependencies, and dependencies vs. build-dependencies.  Several comments here imply that dependencies = Requires, then we walk it back a bit here; I think it could be clearer to remove the "# RPM Requires" comments in the Cargo.toml files (since they're not really true) and have people read this doc instead, and in this doc, also remove the equivalence statements above, and have one section that clearly explains what `[dependencies]` should be and how/when they differ from Requires.

Part of the confusion is that we don't list all of the Requires as dependencies, even if they're not satisfied by build-dependencies, in special cases where we just know it'll cause problems.  (unnecessary systemd rebuilds, glibc, etc.)  If that's always going to be a small set, we could just list them, but either way it's probably worth mentioning.

packages/containerd/Cargo.toml
```diff
@@ -12,6 +12,13 @@ path = "pkg.rs"
...
+# RPM BuildRequires
 [build-dependencies]
 glibc = { path = "../glibc" }
 libseccomp = { path = "../libseccomp" }
+
+# RPM Requires
+[dependencies]
+cni-plugins = { path = "../cni-plugins" }
+runc = { path = "../runc" }
+systemd = { path = "../systemd" }
```
tjkirch, line 24: I'm having a hard time following the discussion at the top level of the PR, but are we sure we still want to include "global" `[dependencies]` like systemd that would be unnecessarily rebuilt?  Per https://github.com/bottlerocket-os/bottlerocket/pull/1408#issuecomment-808767558

Makefile.toml
```diff
@@ -184,7 +184,7 @@ dependencies = ["setup"]
 script_runner = "bash"
```
tjkirch, line 184: For lack of a better spot to leave a comment: I did an audit of dependencies, checking that everything needed is present and everything present is needed, taking into account the proposed policy of not having Requires in dependencies if they're already in build-dependencies, and the couple special cases.

For unneeded, I found that procps has a build-dependency on ncurses that isn't declared in the .spec file - but it does actually need ncurses, so we should probably fix the .spec file.

~For missing, I found kernel Requires filesystem, host-ctr Requires containerd, and ecs-agent Requires docker-engine and iptables, but none of those are listed in Cargo.toml files.~ [edit: fixed]

Makefile.toml
```diff
@@ -266,21 +266,33 @@ cargo install \
...
-[tasks.build-packages]
+# Builds a package including its build-time and runtime dependency packages.
+[tasks.build-package]
 dependencies = ["build-tools", "publish-setup"]
+script_runner = "bash"
 script = [
 '''
+set -e
+if [ -z "${PACKAGE}" ]; then
+    echo "the PACKAGE environment variable must be set. for example:"
```
```suggestion
    echo "The PACKAGE environment variable must be set.  For example:"
```

tjkirch (commented):
This is a great improvement!

webern:
> webern force-pushed the webern:package-build branch from 59dd076 to 07a1b1e

Fixes to readmes and missing dependencies.

webern:
> webern force-pushed the webern:package-build branch from 07a1b1e to 172c298

rebase

webern:
> webern force-pushed the webern:package-build branch from 172c298 to f72a018

Remove k8s-1.15 from variants workspace.

packages/README.md
webern: > I think it could be clearer to remove the "# RPM Requires" comments in the Cargo.toml

The closest thing we have to a mapping is: `Requires:` -> `dependencies` and `BuildRequires:` -> `build-dependencies`. The mapping is true even if some are omitted.

I definitely think it's more helpful than not to have the `# RPM Requires` comment in Cargo.toml even if you need the README to fully understand what's going on.

I'll take another stab at explaining it better in the README.

variants/README.md
webern: > What do you mean by satisfied by the current package

For example os.spec `Requires: apiserver`, but os.spec defines/builds the apiserver package, so there is nothing to add to `[dependencies]`.

> if they need to match, can we get rid of included-packages and have buildsys read the names of the build-dependencies instead

That seems like a reasonable enhancement to consider. I think it should be outside of the scope of this PR, but we could open an issue.

Makefile.toml
webern: > procps has a build-dependency on ncurses that isn't declared in the .spec file - but it does actually need ncurses, so we should probably fix the .spec file.

The only mention I see of ncurses is `--without-ncurses`. Are you sure it's needed?

packages/containerd/Cargo.toml
webern: > are we sure we still want to include "global" [dependencies] like systemd that would be unnecessarily rebuilt?

My preference would be to go ahead with the current version which does not have the global dependency removal optimization. It's a two-way door, but it seems better to me to start with fewer exceptions.

webern:
> webern force-pushed the webern:package-build branch from f72a018 to e42c238

rebase (k8s 1.15 package removal)

webern:
> webern force-pushed the webern:package-build branch from e42c238 to 26c140a 

Another stab at explaining dependencies in packages/README

tjkirch (approved):

bcressey (approved):

PR #1416, opened 2021-03-24
host-ctr: Add "current" generic persistent storage location

jpculp:
**Issue number:**

N/A

**Description of changes:**

This adds an additional mount of the `/local/host-containers/NAME` directory as `/.bottlerocket/host-containers/current` to provide a generic persistent storage location that isn't tied to the name of the container. This makes it easier for scripts to utilize the storage.

**Testing done:**

- Built `aws-ecs-1` ami and launched instance.
- Instance connected to ecs cluster.
- Test task deployed successfully.
- Verified that `/.bottlerocket/host-containers/` contained both a `control` directory and a `current` directory.
- Enabled and launched admin container.
- Verified that `/.bottlerocket/host-containers/` contained both an `admin` directory and a `current` directory.
- Verified that `/.bottlerocket/host-containers/admin/` and `/.bottlerocket/host-containers/current/` had the same contents.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,7 +421,7 @@ If you're running a Kubernetes variant, the no-proxy list will automatically inc
 Beyond just changing the settings above to affect the `admin` and `control` containers, you can add and remove host containers entirely.
 As long as you define the three fields above -- `source` with a URI, and `enabled` and `superpowered` with true/false -- you can add host containers with an API call or user data.
 
-You can optionally define a `user-data` field with arbitrary base64-encoded data, which will be made available in the container at `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME/user-data`.
+You can optionally define a `user-data` field with arbitrary base64-encoded data, which will be made available in the container at `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME/user-data` and (since Bottlerocket v1.0.8) `/.bottlerocket/host-containers/current/user-data`.
 (It was inspired by instance user data, but is entirely separate; it can be any data your host container feels like interpreting.)
 
 Keep in mind that the default admin container (since Bottlerocket v1.0.6) relies on `user-data` to store SSH keys.  You can set `user-data` to [customize the keys](https://github.com/bottlerocket-os/bottlerocket-admin-container/#authenticating-with-the-admin-container), or you can use it for your own purposes in a custom container.
@@ -444,7 +444,8 @@ If the `enabled` flag is `true`, it will be started automatically.
 
 All host containers will have the `apiclient` binary available at `/usr/local/bin/apiclient` so they're able to [interact with the API](#using-the-api-client).
 
-In addition, all host containers come with persistent storage at `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME` that is persisted across reboots and container start/stop cycles.
+In addition, all host containers come with persistent storage that survives reboots and container start/stop cycles.
+It's available at `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME` and (since Bottlerocket v1.0.8) `/.bottlerocket/host-containers/current`.
 The default `admin` host-container, for example, stores its SSH host keys under `/.bottlerocket/host-containers/admin/etc/ssh/`.
 
 There are a few important caveats to understand about host containers:
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -226,13 +226,9 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 					Options:     []string{"bind", "ro"},
 					Destination: "/usr/local/bin/apiclient",
 					Source:      "/usr/bin/apiclient",
-				},
-				// Mount in the persistent storage location for this container
-				{
-					Options:     []string{"rbind", "rw"},
-					Destination: "/.bottlerocket/host-containers/" + containerID,
-					Source:      "/local/host-containers/" + containerID,
 				}}),
+			// Mount in the persistent storage location for this container
+			withPersistentStorage(containerID),
 			// Mount the rootfs with an SELinux label that makes it writable
 			withMountLabel("system_u:object_r:state_t:s0"),
 			// Include conditional options for superpowered containers.
@@ -588,6 +584,26 @@ func withSuperpowered(superpowered bool) oci.SpecOpts {
 				Source:      "/sys/kernel/debug",
 			}}),
 	)
+}
+
+// withPersistentStorage add persistent storage location that matches the container name
+// (legacy location) and a generically named `current` dir. The `current` dir was added for easier
+// referencing in Dockerfiles and scripts. If a host container is also named `current` this function
+// will only add a single `current` mount to the spec.
+func withPersistentStorage(containerID string) oci.SpecOpts {
+	var persistentMounts = []runtimespec.Mount{{
+		Options:     []string{"rbind", "rw"},
+		Destination: "/.bottlerocket/host-containers/" + containerID,
+		Source:      "/local/host-containers/" + containerID,
+	}}
+	if containerID != "current" {
+		persistentMounts = append(persistentMounts, runtimespec.Mount{
+			Options:     []string{"rbind", "rw"},
+			Destination: "/.bottlerocket/host-containers/current",
+			Source:      "/local/host-containers/" + containerID,
+		})
+	}
+	return oci.Compose(oci.WithMounts(persistentMounts))
 }
 
 // pullImage pulls an image from the specified source.


sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -227,11 +227,19 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
...
 					Options:     []string{"rbind", "rw"},
 					Destination: "/.bottlerocket/host-containers/" + containerID,
 					Source:      "/local/host-containers/" + containerID,
+				},
+				// Mount in a generic persistent storage location for this container.
+				// This mount has the same source as the named persistent storage location above,
+				// but the destination is not tied to the containerID.
+				{
+					Options:     []string{"rbind", "rw"},
+					Destination: "/.bottlerocket/host-containers/current",
```
tjkirch: What happens if the container is named "current"?  If the mount fails, I think we'll need to conditionally add this item to the list.

README.md
```diff
@@ -421,7 +421,7 @@ If you're running a Kubernetes variant, the no-proxy list will automatically inc
 Beyond just changing the settings above to affect the `admin` and `control` containers, you can add and remove host containers entirely.
 As long as you define the three fields above -- `source` with a URI, and `enabled` and `superpowered` with true/false -- you can add host containers with an API call or user data.
 
-You can optionally define a `user-data` field with arbitrary base64-encoded data, which will be made available in the container at `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME/user-data`.
+You can optionally define a `user-data` field with arbitrary base64-encoded data, which will be made available in the container at `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME/user-data` and (since Bottlerocket v1.0.8) `/.bottlerocket/host-containers/current/user-data`.
 (It was inspired by instance user data, but is entirely separate; it can be any data your host container feels like interpreting.)
```
tjkirch, line 425: Would you please also update line 447 that talk about this persistent storage directory?  Perhaps:

```markdown
In addition, all host containers come with persistent storage that survives reboots and container start/stop cycles.
It's available at `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME` and (since Bottlerocket v1.0.8) `/.bottlerocket/host-containers/current`.
```

sources/host-ctr/cmd/host-ctr/main.go
webern: I don't see the need for the path to 'look like' the host path. Why not?

```
Destination: "/host,
```

tjkirch: The point of `/.bottlerocket` is that it's a namespace unlikely to be taken by any container image.  (`/host` could legitimately be in an image, and would be shadowed by our data.)  Then, `/host-containers` to segregate the different types of data we insert from the host, this being the persistent store per host container, as opposed to `/rootfs`, etc.

README.md
webern: Should we call `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME` deprecated and encourage `/.bottlerocket/host-containers/current/user-data`?

samuelkarp: ~If we add `/.bottlerocket/host-containers/current`, we'd create a collision with any host container called "current".~

@tjkirch already mentioned this above, you can ignore my additional comment.

tjkirch: > Should we call `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME` deprecated and encourage `/.bottlerocket/host-containers/current/user-data`?

I think we should definitely consider it, but I'm not sure we should confuse things right away by doing it in this PR.  It'd be useful to have a period of overlap to see if there are any issues, and deprecate later.  (This one is fairly low stakes.)

sources/host-ctr/cmd/host-ctr/main.go
jpculp: @tjkirch I just launched an instance with a custom "current" control container. There were no errors to speak of and `/.bottlerocket/host-containers/` only contained one directory `current`. We can certainly add a check if you'd like, but nothing breaks as is.

@webern If we changed the location then we definitely wouldn't need to account for a container named "current" described above. I don't mind having the "current" location be somewhere else, but I think we need more consensus of where that somewhere else would be.

tjkirch: > @tjkirch I just launched an instance with a custom "current" control container. There were no errors to speak of and `/.bottlerocket/host-containers/` only contained one directory `current`. We can certainly add a check if you'd like, but nothing breaks as is.

What mounts show up?  I just tested locally and found that it kinda-sorta recursively mounts.  I bind-mounted a at b, then a at b again, and see _three_ mounts, two for b and one for a.

It seems safest to me to conditionally include this extra entry based on name.

etungsten:
If the goal is to let scripts know where the host-container's persistent storage location is, maybe we can utilize environment variables instead of trying to mount the same thing twice.
https://pkg.go.dev/github.com/ktock/containerd/oci#WithEnv
So something like, 
```
oci.WithEnv([]string{"HOST_CONTAINER_STORAGE=/.bottlerocket/host-containers/" + containerID})
```

tjkirch:
> If the goal is to let scripts know where the host-container's persistent storage location is, maybe we can utilize environment variables instead of trying to mount the same thing twice.
> https://pkg.go.dev/github.com/ktock/containerd/oci#WithEnv
> So something like,
> 
> ```
> oci.WithEnv([]string{"HOST_CONTAINER_STORAGE=/.bottlerocket/host-containers/" + containerID})
> ```

Or just set `"HOST_CONTAINER_NAME="containerID` so the examples we have in the README using that variable would work literally.  It feels _slightly_ more troublesome to me than just having a fixed directory name, because variables add other complications, but not huge.

jpculp:
There are definitely pros and cons to both methods. I'm leaning more towards what @etungsten is saying with the ENV, but using `oci.WithEnv([]string{"HOST_CONTAINER_NAME"=containerID})"`. It keeps our current documentation correct and users don't need to unlearn anything. Is anyone strongly opposed?

tjkirch:
> There are definitely pros and cons to both methods. I'm leaning more towards what @etungsten is saying with the ENV, but using `oci.WithEnv([]string{"HOST_CONTAINER_NAME"=containerID})"`. It keeps our current documentation correct and users don't need to unlearn anything. Is anyone strongly opposed?

Is it going to be possible to symlink `/var/lib/amazon/ssm` to the correct host-containers/FOO in https://github.com/bottlerocket-os/bottlerocket-control-container/pull/12 in Dockerfile if this uses a variable instead of a mount?  The variable wouldn't be present during the container build.  It could be done in `start_control_ssm.sh` but it's better to do that statically where possible.

jpculp:
Right, the symlink would have to be created in `start_control_ssm.sh`, but I don't think that is necessarily a bad thing. The folder it is symlinking to is created in `start_control_ssm.sh`, so it may be better to keep that logic closer together.

jpculp:
I'm switching this PR to draft and proposing #1422 as an alternative.

sources/host-ctr/cmd/host-ctr/main.go
tjkirch: There are a few people opposed to the variable approach because it makes it impossible to construct directories or even know paths at build time, in a Dockerfile, which seems like a fair argument.  I think we should fix this ^ with the conditional, and probably continue with this PR.

README.md
zmrow: Probably good to call out `current` as a "reserved word/name" in the docs, even though we're mentioning it as the location for user data.  Not everyone will make that connection immediately.

zmrow: Oh - maybe I didn't read far enough.  A conditional mount on `container name != "current` seems fine too

jpculp:
- Closed alternative solution #1422.
- Moved persistent mount logic into it's own function. This function mounts the legacy persistant storage location and mounts a `current` storage location only if the host container does not also share the name `current`.

jpculp:
- Small change to another section of the `README`.

webern (approved):

bcressey:
Inspired by the discussion here, I opened #1424 for environment variables we should definitely set: `HTTPS_PROXY` and `NO_PROXY`.

README.md
```diff
@@ -444,7 +444,8 @@ If the `enabled` flag is `true`, it will be started automatically.
 
 All host containers will have the `apiclient` binary available at `/usr/local/bin/apiclient` so they're able to [interact with the API](#using-the-api-client).
 
-In addition, all host containers come with persistent storage at `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME` that is persisted across reboots and container start/stop cycles.
+In addition, all host containers come with persistent storage that survives reboots and container start/stop cycles.
+It's available at `/.bottlerocket/host-containers/$HOST_CONTAINER_NAME` and (since Bottlerocket v1.0.8) `/.bottlerocket/host-containers/current`.
```
bcressey, line 448: I don't like putting a specific version number here - we may need to cut a 1.0.8 release for an emergency security or operational fix, and it'd be easy to overlook references like this.

tjkirch: I suggested this because I didn't like having a documented feature that wasn't available yet, and we've used similar terminology elsewhere. We can update this if there's an emergency. I'd you don't like this, what would you suggest?

bcressey: I don't love it, but it looks like it's consistent with our documentation for the `user-data` feature in v1.0.6, so I'm OK with keeping it.

tjkirch (approved):

zmrow (approved):
:rice:

bcressey (approved):

PR #1417, opened 2021-03-24
add compat symlink for kubelet volume plugins

### LINKED ISSUE #1354: Way to install kubelet plugins?
**What I'd like:**

I'm trying to get Calico to install onto an EKS Bottlerocket system.  By default, Calico tries to install a flexvol driver by mounting `/usr/libexec/kubernetes` in from the host and dropping the binary in there.  However, on Bottlerocket systems, that directory is read-only.  I was wondering if the path is different on Bottlerocket or if kubelet plugins simply are not supported.

* Add workarounds to our docs "remove this part of the YAML"; not a great UX.
* Provide yet-another axis of manifests; so the user has to select "calico-with-vxlan-for-bottlerocket.yaml".  Again, this gives a poor UX where the user has to make lots of decisions.
* Replace our flexvol driver with something else, maybe installing a CSI driver is supported?
* Don't support bottlerocket.

bcressey:
**Issue number:**
Fixes #1354

**Description of changes:**
This allows existing manifests that try to install a flexvol plugin to the default `/usr/libexec/kubernetes` path to work, rather than failing because the root filesystem is read-only.

**Testing done:**
Followed the steps to "Install EKS with Calico networking":
https://docs.projectcalico.org/getting-started/kubernetes/managed-public-cloud/eks

The symlink worked as expected from both the `/usr` path and the target prefix:
```
# ls -latr /usr/libexec/kubernetes/kubelet-plugins
lrwxrwxrwx. 1 root root 38 Mar 24 23:19 /usr/libexec/kubernetes/kubelet-plugins -> ../../../../../var/lib/kubelet/plugins

# ls -latr /usr/libexec/kubernetes/kubelet-plugins/
drwxr-xr-x. 3 root root 4096 Mar 24 23:55 volume
drwxr-xr-x. 2 root root 4096 Mar 24 23:56 ebs.csi.aws.com
drwxr-xr-x. 2 root root 4096 Mar 24 23:56 efs.csi.aws.com

# ls -latr /x86_64-bottlerocket-linux-gnu/sys-root/usr/libexec/kubernetes/kubelet-plugins
lrwxrwxrwx. 1 root root 38 Mar 24 23:19 /x86_64-bottlerocket-linux-gnu/sys-root/usr/libexec/kubernetes/kubelet-plugins -> ../../../../../var/lib/kubelet/plugins

# ls -latr /x86_64-bottlerocket-linux-gnu/sys-root/usr/libexec/kubernetes/kubelet-plugins/
drwxr-xr-x. 3 root root 4096 Mar 24 23:55 volume
drwxr-xr-x. 2 root root 4096 Mar 24 23:56 ebs.csi.aws.com
drwxr-xr-x. 2 root root 4096 Mar 24 23:56 efs.csi.aws.com
```

The node agent plugin showed up in the expected location:
```
# find /var/lib/kubelet/plugins -type s -o -type f
/var/lib/kubelet/plugins/efs.csi.aws.com/csi.sock
/var/lib/kubelet/plugins/volume/exec/nodeagent~uds/uds
/var/lib/kubelet/plugins/ebs.csi.aws.com/csi.sock
```

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.15/kubernetes-1.15.spec
+++ b/packages/kubernetes-1.15/kubernetes-1.15.spec
@@ -87,6 +87,11 @@ install -m 0644 %{S:7} %{buildroot}%{_cross_templatedir}/kubelet-bootstrap-kubec
 install -d %{buildroot}%{_cross_tmpfilesdir}
 install -p -m 0644 %{S:8} %{buildroot}%{_cross_tmpfilesdir}/kubernetes.conf
 
+install -d %{buildroot}%{_cross_libexecdir}/kubernetes
+ln -rs \
+  %{buildroot}%{_sharedstatedir}/kubelet/plugins \
+  %{buildroot}%{_cross_libexecdir}/kubernetes/kubelet-plugins
+
 %cross_scan_attribution --clarify %{S:1000} go-vendor vendor
 
 %files -n %{_cross_os}kubelet-1.15
@@ -103,5 +108,7 @@ install -p -m 0644 %{S:8} %{buildroot}%{_cross_tmpfilesdir}/kubernetes.conf
 %{_cross_templatedir}/kubelet-exec-start-conf
 %{_cross_templatedir}/kubernetes-ca-crt
 %{_cross_tmpfilesdir}/kubernetes.conf
+%dir %{_cross_libexecdir}/kubernetes
+%{_cross_libexecdir}/kubernetes/kubelet-plugins
 
 %changelog
--- a/packages/kubernetes-1.16/kubernetes-1.16.spec
+++ b/packages/kubernetes-1.16/kubernetes-1.16.spec
@@ -83,6 +83,11 @@ install -m 0644 %{S:7} %{buildroot}%{_cross_templatedir}/kubelet-bootstrap-kubec
 install -d %{buildroot}%{_cross_tmpfilesdir}
 install -p -m 0644 %{S:8} %{buildroot}%{_cross_tmpfilesdir}/kubernetes.conf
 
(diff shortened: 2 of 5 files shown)


etungsten (approved):

packages/kubernetes-1.15/kubernetes-1.15.spec
```diff
@@ -87,6 +87,9 @@ install -m 0644 %{S:7} %{buildroot}%{_cross_templatedir}/kubelet-bootstrap-kubec
 install -d %{buildroot}%{_cross_tmpfilesdir}
 install -p -m 0644 %{S:8} %{buildroot}%{_cross_tmpfilesdir}/kubernetes.conf
 
+install -d %{buildroot}%{_cross_libexecdir}/kubernetes
+ln -rs %{_sharedstatedir}/kubelet/plugins %{buildroot}%{_cross_libexecdir}/kubernetes/kubelet-plugins
```
tjkirch: You mentioned offline that this results in 10 levels of `../` in the link target, which may be due to expanding `%{buildroot}` on one side and not the other, or something like that.  (There are 5 levels in the expansion of `%{_cross_libexecdir}`, so maybe that's telling?)  I'm curious if it could be simplified somehow.  It might help to show the links in the testing description so it's easier to understand the structures being linked together.

tjkirch (approved):

bcressey:
Fixed excessive use of `../`.

tjkirch (approved):

samuelkarp (approved):

PR #1422, opened 2021-03-25
host-ctr: export containerID as HOST_CONTAINER_NAME env variable

jpculp:
**Issue number:**

N/A

**Description of changes:**

This adds a `HOST_CONTAINER_NAME` environment variable to host containers so that they are able to reference their own names. This is useful for scripts that want to take advantage of the persistent storage location `/.bottlerocket/host-containers/NAME`.

**Testing done:**

- Built `aws-ecs-1` ami and launched instance.
- Instance connected to ecs cluster.
- Test task deployed successfully.
- Replaced control and admin containers with patched containers that take advantage of the HOST_CONTAINER_NAME environment variable.
- Verified that `/.bottlerocket/host-containers/` contained a `control` directory.
- Enabled and launched admin container.
- Verified that `/.bottlerocket/host-containers/` contained an `admin`.
- Disabled the control and admin containers and replaced them with containers named `custom-control` and `custom-admin`
- Connected to each of the containers and verified they were still functioning.
- Verified that `/.bottlerocket/host-containers/` contained a single directory with a name that matched the `containerID`.

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -213,6 +213,9 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 			oci.WithHostNamespace(runtimespec.NetworkNamespace),
 			oci.WithHostHostsFile,
 			oci.WithHostResolvconf,
+			// Set environment variable for host containers to reference their name.
+			// This is essential for containers to reference their persistent storage location.
+			oci.WithEnv([]string{"HOST_CONTAINER_NAME=" + containerID}),
 			// Mount in the API socket for the Bottlerocket API server, and the API
 			// client used to interact with it
 			oci.WithMounts([]runtimespec.Mount{


jpculp:
This is an alternative solution to #1416 to enable custom-named containers in [bottlerocker-control-container #12](https://github.com/bottlerocket-os/bottlerocket-control-container/pull/12)

jpculp:
After further discussion, we're going to stay the course with the solution outlined in #1416.

webern:
This could be a nice additional feature. Suppose I run two instances of the same container image, I could use the name to differentiate their logic, or in their logging.

But, treating this as YAGNI until someone asks for it makes sense. No need to invent features that don't yet have use cases.

tjkirch:
^ I agree, and when we have more urgent uses, I think it's worth considering what other type of data we'd want to expose between host and host-container, so we can decide if environment variables are the best option.

Issue #1392, opened 2021-03-15
add a mechanism for containers to configure the host on first boot


bcressey:
**What I'd like:**
I'd like a way to extend first boot setup by bringing my own custom logic in a container.

Some use cases:
* creating paths on the host before any orchestrated container starts (#1208)
* generating a setting on boot based on the running environment (#1307)
* applying settings in bulk from a remote source (#1391)

**Any alternatives you've considered:**
In #1208 we looked at extending the settings-based API to create directories and files. The goal was to compose this functionality with other settings, such as #1209 for setting up filesystem mounts.

However, this introduces a lot of complexity in the general case. Declaring that a file should be created in a directory on a specific mounted filesystem means that we have to account for the filesystem's devices not always being available. When settings are changed, reconciliation can be challenging. The directory may be declared one way and later moved to a different mount; the directory may contain other, unmanaged host files; a declared file may have its contents modified outside of the API.

We can constrain this with arbitrary rules, such as: files created through the API are immutable; directories will be reconciled at boot, with permissions restored and unmanaged files removed. We can also limit composition: not allowing host files to be related to host directories; or not allowing host directories to be created on ephemeral disks.

In both cases we end up with an API that might suffice for the specific use cases we've heard about, but which is unlikely to generalize. The most powerful solution is to bring a host container with its own tools and logic, and rely on that instead of any host API features.

Host containers as they stand today fall short in a few ways. They run at an unspecified point in the boot sequence, but generally at the same time as orchestrated containers. They're meant to run as long-lived services rather than one-shots. They're also overly privileged for this use case; a super-powered bootstrap container could undermine many of the host protections in subtle ways.

bcressey:
Tracking the implementation in #1387.

PR #1423, opened 2021-03-26
packages: change systemd boot sequence

arnaldo2792:
**Issue number:**
#1392

**Description of changes:**

```
packages: change systemd boot sequence
```

The current systemd boot sequence is error prone on isolated scenarios like sending invalid configurations through user data. These scenarios could cause inconsistent states at the end of the boot process, which affect the mechanism to determine if a boot was successful after applying new configurations or updates.

With the current boot sequence, it is difficult to implement features that require a strict services' initialization order (a.k.a run levels).

In order to fix the problems presented above, this commit defines a new boot sequence, with three main systemd targets: preconfigured, configured and multi-user. The "host-containers" crate was modified to enable the host-container service units only when multi-user is the current default target.

The "mark-successful-boot" service unit was moved from the "multi-user" target to the "preconfigured" target, and it only depends on the "migrator" oneshot service unit since it is the only service required (for now) to determine if a boot was successful. The "mark-successful-boot" service unit used to send metrics after it was executed. However, doing so indirectly declared a strong dependency on "settings-applier" since the unit was setup to use the "proxy.env" file. The "metricdog-successful-boot" oneshot service unit was created to remove such dependency from "mark-successful-boot".

#### Preconfigured
This target is used to start the boot process (a.k.a default target), it is based on the "multi-user" target provided by systemd so that other basic systemd units are started during this stage. Failures in the required "oneshot" service units will cause the target to fail, stopping the boot process. The boot is marked as successful during the execution of this target, only if the migrator "oneshot" service exits successfully. Services initialized during this phase include:

* acpid
* chrony
* dbus-broker
* audit-rules
* migrator (required)
* metricdog-successful-boot
* storewolf (required)
* early-boot-config (required)
* sundog (required)
* settings-applier(required)

Once the target is reached (completed), the "activate-configured.service" unit will set the configured target as the default target and start it.

#### Configured
This target should be used to setup additional configurations in the host before services like kubernetes/ecs start. Services initialized during this phase include:

* chronyd
* host-containerd

Once the target is reached (completed) the "activate-multi-user.service" unit will set the multi-user target as the default target, and start it.

#### Multi-user
This is the final target enabled in the boot sequence. Services initialized during this phase include:

* host-containers@*
* docker
* ecs
* kubernetes

**Testing done:**
* `systemctl status` didn't show any failures on successful boots
* Run nginx pod/task/container
* Admin/Control containers were enabled once the `multi-user` target was reached
* Custom host container sent as user data was enabled once  the `multi-user`  target was reached
* Custom host container created after boot was enabled after the new settings were applied

```sh
apiclient set --json \
    '{"host-containers": { "test": { "source": "docker.io/arnaldo2792/blocker:latest", "superpowered": false, "enabled": true}}}'
```

```sh
ps ax -o pid,args | grep host-ctr
479 /usr/bin/host-ctr run --container-id=admin --source=328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.6.0 --superpowered=true
480 /usr/bin/host-ctr run --container-id=control --source=328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-control:v0.4.2 --superpowered=false
481 /usr/bin/host-ctr run --container-id=custom-at-boot --source=docker.io/arnaldo2792/blocker:latest --superpowered=false
16546 /usr/bin/host-ctr run --container-id=test --source=docker.io/arnaldo2792/blocker:latest --superpowered=false
```
* Failed boot with bad configurations

```toml
enabled = true

[settings.bad]
bad = "data"
```

```sh
[   11.365797] early-boot-config[366]: Error PATCHing '/settings?tx=bottlerocket-launch': Status 400 when PATCHing /settings?tx=bottlerocket-launch: Json deserialize error: unknown field `bad`, expected one of `source`, `enabled`, `superpowered`, `user-data` at line 1 column 85
[DEPEND] Dependency failed for Bottlerocket dynamic pre-configuration.
```

**Terms of contribution:**

### DIFF
--- a/packages/chrony/chronyd.service
+++ b/packages/chrony/chronyd.service
@@ -1,12 +1,12 @@
 [Unit]
 Description=A versatile implementation of the Network Time Protocol
 Documentation=https://chrony.tuxfamily.org
-After=network-online.target configured.target
-Requires=network-online.target configured.target
+After=network-online.target preconfigured.target
+Wants=network-online.target preconfigured.target
 
 [Service]
 Type=simple
 ExecStart=/usr/sbin/chronyd -d -F -1
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=configured.target
--- a/packages/libaudit/audit-rules.service
+++ b/packages/libaudit/audit-rules.service
@@ -12,4 +12,4 @@ ExecStart=/sbin/auditctl -R /usr/share/audit/audit.rules
 ExecStop=-/sbin/auditctl -D
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=preconfigured.target
--- a/packages/os/early-boot-config.service
+++ b/packages/os/early-boot-config.service
@@ -1,11 +1,18 @@
 [Unit]
 Description=Bottlerocket userdata configuration system
 # Need network online to talk to IMDS.
-After=network-online.target apiserver.service
-Requires=network-online.target apiserver.service
+After=network-online.target apiserver.service storewolf.service
+# Don't restart the unit if the network goes offline or apiserver restarts
+Wants=apiserver.service network-online.target
+# Don't start the unit if storewolf.service fails
+Requires=storewolf.service
 # We only want to run once, at first boot.  This file is created by early-boot-config
 # after a successful run.
 ConditionPathExists=!/var/lib/bottlerocket/early-boot-config.ran
+# Block manual interactions with this service, since it could leave the system in an
+# unexpected state
+RefuseManualStart=true
+RefuseManualStop=true
 
 [Service]
 Type=oneshot
@@ -14,4 +21,4 @@ RemainAfterExit=true
 StandardError=journal+console
 
 [Install]
-WantedBy=multi-user.target
+RequiredBy=preconfigured.target
--- a/packages/os/host-containers@.service
+++ b/packages/os/host-containers@.service
@@ -7,8 +7,6 @@ Wants=host-containerd.service
 Type=simple
 EnvironmentFile=/etc/host-containers/%i.env
 Environment=LOCAL_DIR=/local
-# Create directories for container persistent storage
-ExecStartPre=/usr/bin/mkdir -m 1777 -p ${LOCAL_DIR}/host-containers/%i
 ExecStart=/usr/bin/host-ctr run \
     --container-id='%i' \
     --source='${CTR_SOURCE}' \
--- a/packages/os/mark-successful-boot.service
+++ b/packages/os/mark-successful-boot.service
@@ -1,15 +1,18 @@
 [Unit]
 Description=Call signpost to mark the boot as successful after all required targets are met.
-After=multi-user.target
-# Each service that must start correctly in order for a boot to be successful should be of type "notify"
-# and include "RequiredBy=mark-successful-boot.service" in its [Install] section.
+# This unit is in charge of updating the partitions on successful boots. Use other service
+# units instead of adding more `ExecStart*` lines to prevent indirect dependencies on
+# other units not listed in the `RequiredBy` section.
+Requires=migrator.service
+# Block manual interactions with this service, manually running it could leave the system in an
+# unexpected state
+RefuseManualStart=true
+RefuseManualStop=true
 
 [Service]
-EnvironmentFile=/etc/network/proxy.env
 Type=oneshot
 RemainAfterExit=true
 ExecStart=/bin/signpost mark-successful-boot
-ExecStartPost=-/usr/bin/metricdog send-boot-success
 
 [Install]
-WantedBy=multi-user.target
+RequiredBy=preconfigured.target
--- a/packages/os/migrator.service
+++ b/packages/os/migrator.service
@@ -1,11 +1,18 @@
 [Unit]
 Description=Bottlerocket data store migrator
+RefuseManualStart=true
+RefuseManualStop=true
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/migrator --datastore-path /var/lib/bottlerocket/datastore/current --migration-directory /var/lib/bottlerocket-migrations --root-path /usr/share/updog/root.json --metadata-directory /var/cache/bottlerocket-metadata --migrate-to-version-from-os-release
+ExecStart=/usr/bin/migrator \
+  --datastore-path /var/lib/bottlerocket/datastore/current \
+  --migration-directory /var/lib/bottlerocket-migrations \
+  --root-path /usr/share/updog/root.json \
+  --metadata-directory /var/cache/bottlerocket-metadata \
+  --migrate-to-version-from-os-release
 RemainAfterExit=true
 StandardError=journal+console
 
 [Install]
-WantedBy=multi-user.target
+RequiredBy=preconfigured.target
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -34,6 +34,7 @@ Source107: host-containers@.service
 Source110: mark-successful-boot.service
 Source111: metricdog.service
 Source112: metricdog.timer
+Source113: send-boot-success.service
 
 # 2xx sources: tmpfilesd configs
 Source200: migration-tmpfiles.conf
@@ -337,7 +338,7 @@ install -p -m 0644 %{S:5} %{S:6} %{buildroot}%{_cross_templatedir}
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 \
   %{S:100} %{S:101} %{S:102} %{S:103} %{S:105} \
-  %{S:106} %{S:107} %{S:110} %{S:111} %{S:112} \
+  %{S:106} %{S:107} %{S:110} %{S:111} %{S:112} %{S:113}\
   %{buildroot}%{_cross_unitdir}
 
 install -d %{buildroot}%{_cross_tmpfilesdir}
@@ -440,6 +441,7 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 %{_cross_templatedir}/metricdog-toml
 %{_cross_unitdir}/metricdog.service
 %{_cross_unitdir}/metricdog.timer
+%{_cross_unitdir}/send-boot-success.service
 
 %files -n %{_cross_os}logdog
 %{_cross_bindir}/logdog
--- a/packages/os/send-boot-success.service
+++ b/packages/os/send-boot-success.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Send boot success
+# The unit depends on 'configured.target' since Metricdog indirectly
+# depends on the proxy.env file created by settings-applier in the
+# preconfigured target
+After=network-online.target configured.target
+Wants=network-online.target configured.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+EnvironmentFile=/etc/network/proxy.env
+ExecStart=-/usr/bin/metricdog send-boot-success
+
+[Install]
+WantedBy=multi-user.target
--- a/packages/os/settings-applier.service
+++ b/packages/os/settings-applier.service
@@ -1,8 +1,13 @@
 [Unit]
 Description=Applies settings to create config files
 After=storewolf.service sundog.service early-boot-config.service apiserver.service
-Requires=apiserver.service
-Wants=storewolf.service sundog.service early-boot-config.service
+Requires=storewolf.service sundog.service early-boot-config.service
+# We don't want to restart the unit if apiserver restarts
+Wants=apiserver.service
+# Block manual interactions with this service, since it could leave the system in an
+# unexpected state
+RefuseManualStart=true
+RefuseManualStop=true
 
 [Service]
 Type=oneshot
@@ -12,4 +17,4 @@ RemainAfterExit=true
 StandardError=journal+console
 
 [Install]
-WantedBy=multi-user.target
+RequiredBy=preconfigured.target
--- a/packages/os/sundog.service
+++ b/packages/os/sundog.service
@@ -1,8 +1,14 @@
 [Unit]
 Description=User-specified setting generators
-# Need network access to support commands talking to IMDS.
+# Needs network access to support commands talking to IMDS.
 After=network-online.target apiserver.service early-boot-config.service
-Requires=network-online.target apiserver.service
+Requires=early-boot-config.service
+# We don't want to restart the unit if the network goes offline or apiserver restarts
+Wants=network-online.target apiserver.service
+# Block manual interactions with this service, since it could leave the system in an
+# unexpected state
+RefuseManualStart=true
+RefuseManualStop=true
 
 [Service]
 Type=oneshot
@@ -13,4 +19,4 @@ RemainAfterExit=true
 StandardError=journal+console
 
 [Install]
-WantedBy=multi-user.target
+RequiredBy=preconfigured.target
--- a/packages/release/activate-configured.service
+++ b/packages/release/activate-configured.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Isolates configured.target
+After=preconfigured.target
+Requires=preconfigured.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/systemctl set-default configured
+ExecStart=/usr/bin/systemctl isolate default
+RemainAfterExit=true
+StandardError=journal+console
+
+[Install]
+WantedBy=preconfigured.target
--- a/packages/release/configured.target
+++ b/packages/release/configured.target
@@ -1,5 +1,8 @@
 [Unit]
-Description=Bottlerocket user and dynamic configuration complete
-After=settings-applier.service
-Requires=settings-applier.service early-boot-config.service
+Description=Bottlerocket final configuration complete
+After=preconfigured.target
+Requires=preconfigured.target
 AllowIsolate=yes
+
+[Install]
+RequiredBy=multi-user.target
--- a/packages/release/preconfigured.target
+++ b/packages/release/preconfigured.target
@@ -1,0 +1,11 @@
+[Unit]
+Description=Bottlerocket initial configuration complete
+AllowIsolate=yes
+After=basic.target
+Requires=basic.target
+# Prevent manually starting/stopping the target
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Install]
+RequiredBy=configured.target multi-user.target
--- a/sources/api/host-containers/src/main.rs
+++ b/sources/api/host-containers/src/main.rs
@@ -170,7 +170,7 @@ impl<'a> SystemdUnit<'a> {
 
     fn is_enabled(&self) -> Result<bool> {
         match command(SYSTEMCTL_BIN, &["is-enabled", &self.unit]) {
-            Ok(()) => Ok(true),
+            Ok(_) => Ok(true),
             Err(e) => {
                 // If the systemd unit is not enabled, then `systemctl is-enabled` will return a
                 // non-zero exit code.
@@ -187,7 +187,7 @@ impl<'a> SystemdUnit<'a> {
 
     fn is_active(&self) -> Result<bool> {
         match command(SYSTEMCTL_BIN, &["is-active", &self.unit]) {
-            Ok(()) => Ok(true),
+            Ok(_) => Ok(true),
             Err(e) => {
                 // If the systemd unit is not active(running), then `systemctl is-active` will
                 // return a non-zero exit code.
@@ -202,23 +202,33 @@ impl<'a> SystemdUnit<'a> {
         }
     }
 
+    fn enable(&self) -> Result<()> {
+        command(SYSTEMCTL_BIN, &["enable", &self.unit])?;
+
+        Ok(())
+    }
+
     fn enable_and_start(&self) -> Result<()> {
-        // We enable/start units with --no-block to work around cyclic dependency issues at boot
-        // time.  It would probably be better to give systemd more of a chance to tell us that
-        // something failed to start, if dependencies can be resolved in another way.
         command(
             SYSTEMCTL_BIN,
             &["enable", &self.unit, "--now", "--no-block"],
-        )
+        )?;
+
+        Ok(())
     }
 
     fn disable_and_stop(&self) -> Result<()> {
-        command(SYSTEMCTL_BIN, &["disable", &self.unit, "--now"])
+        command(
+            SYSTEMCTL_BIN,
+            &["disable", &self.unit, "--now", "--no-block"],
+        )?;
+
+        Ok(())
     }
 }
 
 /// Wrapper around process::Command that adds error checking.
-fn command<I, S>(bin_path: &str, args: I) -> Result<()>
+fn command<I, S>(bin_path: &str, args: I) -> Result<String>
 where
     I: IntoIterator<Item = S>,
     S: AsRef<OsStr>,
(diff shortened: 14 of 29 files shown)


packages/chrony/chronyd.service
```diff
@@ -1,12 +1,12 @@
...
+After=network-online.target
+Requires=network-online.target preconfigured.target
 
 [Service]
 Type=simple
 ExecStart=/usr/sbin/chronyd -d -F -1
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=preconfigured.target
```
bcressey: This seems like it should be `WantedBy=configured.target` since that's the next one in the chain.

packages/libaudit/audit-rules.service
```diff
@@ -12,4 +12,4 @@ ExecStart=/sbin/auditctl -R /usr/share/audit/audit.rules
 ExecStop=-/sbin/auditctl -D
 
 [Install]
-WantedBy=multi-user.target
+RequiredBy=preconfigured.target
```
bcressey: This can be `WantedBy=preconfigured.target` since even though it's a one-shot, it's not related to settings.

packages/os/host-containers@.service
```diff
@@ -1,14 +1,12 @@
 [Unit]
 Description=Host container: %i
 After=host-containerd.service
-Wants=host-containerd.service
+Requires=host-containerd.service
```
bcressey: This would revert e1a051c81 and cause host container units to be stopped whenever host-containerd restarts.

packages/os/migrator.service
```diff
@@ -3,9 +3,13 @@ Description=Bottlerocket data store migrator
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/migrator --datastore-path /var/lib/bottlerocket/datastore/current --migration-directory /var/lib/bottlerocket-migrations --root-path /usr/share/updog/root.json --metadata-directory /var/cache/bottlerocket-metadata --migrate-to-version-from-os-release
+ExecStart=/usr/bin/migrator --datastore-path /var/lib/bottlerocket/datastore/current \
```
bcressey: nit:
```suggestion
ExecStart=/usr/bin/migrator \
  --datastore-path /var/lib/bottlerocket/datastore/current \
```

packages/os/sundog.service
```diff
@@ -1,8 +1,8 @@
 [Unit]
 Description=User-specified setting generators
-# Need network access to support commands talking to IMDS.
-After=network-online.target apiserver.service early-boot-config.service
-Requires=network-online.target apiserver.service
+# Needs network access to support commands talking to IMDS.
+After=network-online.target early-boot-config.service
+Requires=network-online.target early-boot-config.service
```
bcressey: `sundog` has a direct dependency on the API being up, so it seems like it makes sense to keep the `apiserver.service` requirement.

packages/release/activate-configured.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Isolates configured.target
+After=preconfigured.target
+Requires=preconfigured.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/systemctl isolate configured
```
bcressey: For consistency, should we also set the default?
```suggestion
ExecStart=/usr/bin/systemctl set-default configured
ExecStart=/usr/bin/systemctl isolate configured
```

packages/release/configured.target
```diff
@@ -1,5 +1,9 @@
 [Unit]
 Description=Bottlerocket user and dynamic configuration complete
-After=settings-applier.service
-Requires=settings-applier.service early-boot-config.service
+Before=multi-user.target
```
bcressey: Is this effectively serving as a sort of documentation? I don't think we strictly need to specify this, unless something is queuing up the `multi-user.target` ahead of time.

I think it'd be more consistent to give `multi-user.target` -> `After=configured.target`, so that none of the targets specify "Before".

packages/release/preconfigured.target
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Bottlerocket dynamic pre-configuration
+AllowIsolate=yes
+After=basic.target
+Requires=basic.target settings-applier.service
```
bcressey: Seems weird to specifically list `settings-applier.service` here; we've got that information in the settings applier unit.

sources/api/host-containers/src/main.rs
```diff
@@ -383,7 +393,19 @@ where
         if host_containerd_unit.is_active()? && !systemd_unit.is_enabled()? {
             command(HOST_CTR_BIN, &["clean-up", "--container-id", name])?;
         }
-        systemd_unit.enable_and_start()?;
+
+        // Only start the host container if the systemd target is 'multi-user'
```
bcressey:
```suggestion
        // Only start the host container if the systemd target is 'multi-user', otherwise it will start before the system is fully configured.
```

packages/docker-engine/docker.service
```diff
@@ -2,8 +2,8 @@
 Description=docker engine
 Documentation=https://docs.docker.com
 BindsTo=containerd.service
-After=network-online.target containerd.service configured.target
-Wants=network-online.target configured.target
+After=network-online.target containerd.service
+Wants=network-online.target
```
bcressey: I'd prefer to keep `configured.target` in `After` / `Wants`, even if it may be overspecified. Does it cause a problem?

(Same with the similar changes for ecs-agent and kubelet.)

packages/docker-engine/docker.service
```diff
@@ -25,3 +25,4 @@ TasksMax=infinity
 
 [Install]
 WantedBy=multi-user.target
+RequiredBy=mark-successful-boot.service
```
bcressey: We may also want a `Before=mark-successful-boot.service` in the unit section to specify ordering as well as dependency.

(Same with the similar changes for ecs-agent and kubelet.)

packages/release/preconfigured.target
```diff
@@ -0,0 +1,10 @@
...
+RefuseManualStart=true
+RefuseManualStop=true
```
bcressey, line 8: I had the sense you said we couldn't add these to the targets or it would break `systemctl isolate`, but it doesn't seem like it does.

Can we add them to the other targets, and potentially all of our required one-shots?

packages/os/settings-applier.service
```diff
@@ -1,8 +1,7 @@
 [Unit]
 Description=Applies settings to create config files
-After=storewolf.service sundog.service early-boot-config.service apiserver.service
-Requires=apiserver.service
-Wants=storewolf.service sundog.service early-boot-config.service
```
bcressey: Why remove all of the other dependency information? `sundog` is not the only service that's needed for `settings-applier.service` to succeed.

bcressey (changes requested):

bcressey:
Re: preconfigured section - `apiserver` is not a one-shot.

One question is whether `systemctl restart apiserver` or a stop & start causes the system to fail catastrophically because all the one-shots in "preconfigured" require it. If so, we might want to relax those dependencies to "Wants" instead.

packages/chrony/chronyd.service
arnaldo2792: I used `WantedBy=preconfigured.target` because I thought that chrony should is one of the services required at the very early stages of the boot (during the preconfigured target). I'll make the changes to be started in the `configured` target.

packages/docker-engine/docker.service
arnaldo2792: It doesn't cause any problems, I can make the changes.

packages/docker-engine/docker.service
arnaldo2792: Based on our offline conversation, we will only use `RequiredBy=migrator.service` in the `mark-successful-boot.service`

packages/os/settings-applier.service
arnaldo2792: `sundog` requires `early-boot-config.service`, which itself requires `apiserver.service`. So by specifying just `sundog`, we get the exact same behavior.

packages/os/sundog.service
arnaldo2792: `early-boot-config` requires `apiserver`, so by just specifying `early-boot-config` we get the exact same behavior as specifying all the removed units.

packages/release/activate-configured.service
arnaldo2792: I could but I wanted to minimize the calls to `set-default` because it will effectively create/remove symlinks.

packages/release/configured.target
arnaldo2792: Yeah, it was some sort of "documentation", I wanted the target to be very specific about when in the boot process it runs.

packages/release/preconfigured.target
arnaldo2792: `preconfigured` is queued by systemd itself, so _that_ call isn't manual and using those two configurations here works

I could try to add the configuration to the oneshots, I don't see why they couldn't use these settings, but the other targets can't use them.

packages/release/preconfigured.target
arnaldo2792: I used `settings-applier` since it is the last "leaf" in the dependency tree, all the services required by `preconfigured.target` are effectively "successful" if `settings-applier` exits successfully

packages/release/activate-configured.service
bcressey: I'd prefer the consistency and the ability to use command-line tools to inspect what state the host system thinks it's reached.

packages/chrony/chronyd.service
bcressey: It'd be fine to put it in the "preconfigured" stage, although a bit tricky since chrony itself needs to be configured by settings. I'm OK with "configured" for now.

packages/os/sundog.service
bcressey: I'd prefer units to more fully specify their dependency chain, at least within a stage and for our own programs, rather than relying on transitive dependencies. That makes it easier to grep for dependencies in the source tree, and easier to reason about the impact of moving or reordering any particular service.

packages/release/preconfigured.target
bcressey: This goes along with my comment about disliking transitive dependencies: it's hard to trust and reason about `settings-applier` being correct here. What if we add some other service that starts after settings-applier?

I'd prefer that targets mostly just express dependencies on other targets in their own units, and that we fill in the dependency graph in service units instead.

packages/chrony/chronyd.service
```diff
@@ -1,12 +1,12 @@
 [Unit]
 Description=A versatile implementation of the Network Time Protocol
 Documentation=https://chrony.tuxfamily.org
-After=network-online.target configured.target
-Requires=network-online.target configured.target
+After=network-online.target
```
bcressey: This should still come after `preconfigured.target`:

```suggestion
After=network-online.target preconfigured.target
```

packages/os/metricdog-successful-boot.service
```diff
@@ -0,0 +1,14 @@
...
+Requires=settings-applier.service mark-successful-boot.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+EnvironmentFile=/etc/network/proxy.env
+ExecStart=-/usr/bin/metricdog send-boot-success
+
+[Install]
+WantedBy=preconfigured.target
```
bcressey: If we're sending a ping that includes service health, this should happen in `multi-user.target` when the relevant services have started.

packages/os/mark-successful-boot.service
```diff
@@ -1,15 +1,14 @@
...
 [Service]
-EnvironmentFile=/etc/network/proxy.env
 Type=oneshot
 RemainAfterExit=true
 ExecStart=/bin/signpost mark-successful-boot
-ExecStartPost=-/usr/bin/metricdog send-boot-success
 
 [Install]
-WantedBy=multi-user.target
+WantedBy=preconfigured.target
```
bcressey: If we don't mark the boot as successful, we could potentially run for a very long time with no issues, and then fail back to a previous release on reboot.

```suggestion
RequiredBy=preconfigured.target
```

packages/os/mark-successful-boot.service
```diff
@@ -1,15 +1,14 @@
 [Unit]
 Description=Call signpost to mark the boot as successful after all required targets are met.
-After=multi-user.target
+After=migrator.service
+Requires=migrator.service
```
bcressey, line 6:
```suggestion
Requires=migrator.service
RefuseManualStart=true
RefuseManualStop=true
```

packages/os/mark-successful-boot.service
```diff
@@ -1,15 +1,14 @@
...
 # Each service that must start correctly in order for a boot to be successful should be of type "notify"
 # and include "RequiredBy=mark-successful-boot.service" in its [Install] section.
```
bcressey: We should get rid of this comment.

packages/os/early-boot-config.service
```diff
@@ -2,10 +2,15 @@
...
 # Need network online to talk to IMDS.
 After=network-online.target apiserver.service
-Requires=network-online.target apiserver.service
+Requires=network-online.target
+# We don't want to restart the unit if apiserver restarts
+Wants=apiserver.service
 # We only want to run once, at first boot.  This file is created by early-boot-config
 # after a successful run.
 ConditionPathExists=!/var/lib/bottlerocket/early-boot-config.ran
+# Prevent manually starting/stoping the service
```
bcressey: typo:
```suggestion
# Prevent manually starting/stopping the service
```

There are a couple of other cases of this.

packages/os/send-boot-success.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Send successful boot metrics
```
bcressey: nit:
```suggestion
Description=Send boot success
```

packages/os/early-boot-config.service
```diff
@@ -2,10 +2,15 @@
 Description=Bottlerocket userdata configuration system
 # Need network online to talk to IMDS.
 After=network-online.target apiserver.service
-Requires=network-online.target apiserver.service
+Requires=network-online.target
+# We don't want to restart the unit if apiserver restarts
+Wants=apiserver.service
```
bcressey: `network-online.target` shouldn't be fallible but the same rationale would apply to it:
```suggestion
# We don't want to restart the unit if the network goes offline or apiserver restarts
Wants=network-online.target apiserver.service
```

packages/os/os.spec
```diff
@@ -34,6 +34,7 @@ Source107: host-containers@.service
 Source110: mark-successful-boot.service
 Source111: metricdog.service
 Source112: metricdog.timer
+Source113: metricdog-successful-boot.service
```
bcressey:
```suggestion
Source113: send-boot-success.service
```

packages/os/send-boot-success.service
```diff
@@ -0,0 +1,14 @@
...
+# Metricdog depends on the proxy.env file created by settings-applier.service
+After=settings-applier.service
+Requires=settings-applier.service
```
bcressey: Seems like we can just treat this the same as the other multi-user services:
```suggestion
After=network-online.target configured.target
Wants=network-online.target configured.target
```

packages/os/sundog.service
```diff
@@ -1,8 +1,13 @@
 [Unit]
 Description=User-specified setting generators
-# Need network access to support commands talking to IMDS.
+# Needs network access to support commands talking to IMDS.
 After=network-online.target apiserver.service early-boot-config.service
-Requires=network-online.target apiserver.service
+Requires=network-online.target early-boot-config.service
+# We don't want to restart the unit if apiserver restarts
+Wants=apiserver.service
```
bcressey:
```suggestion
Requires=early-boot-config.service
# We don't want to restart the unit if the network goes offline or apiserver restarts
Wants=network-online.target apiserver.service
```

packages/release/preconfigured.target
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Bottlerocket dynamic pre-configuration
```
bcressey:
```suggestion
Description=Bottlerocket initial configuration complete
```

packages/release/configured.target
```diff
@@ -1,5 +1,8 @@
 [Unit]
 Description=Bottlerocket user and dynamic configuration complete
```
bcressey:
```suggestion
Description=Bottlerocket final configuration complete
```

bcressey (commented):
Nice! Just a handful of wording and consistency fixes.

arnaldo2792:
In the last forced pushed I addressed the comments left by @bcressey:

* Typo in `stoping` in multiple units
* Fixed os.spec to include "send-boot-success.service"
* Fixed "send-boot-success.service" dependencies
* Fixed "sundog.service" dependencies
* Fixed "preconfigured.target" and "configured.target" descriptions

packages/os/early-boot-config.service
```diff
@@ -2,10 +2,14 @@
 Description=Bottlerocket userdata configuration system
 # Need network online to talk to IMDS.
 After=network-online.target apiserver.service
-Requires=network-online.target apiserver.service
+# We don't want to restart the unit the network goes offline or apiserver restarts
```
bcressey:
```suggestion
# We don't want to restart the unit if the network goes offline or apiserver restarts
```

packages/os/early-boot-config.service
```diff
@@ -2,10 +2,14 @@
...
+Wants=apiserver.service network-online.target
 # We only want to run once, at first boot.  This file is created by early-boot-config
 # after a successful run.
 ConditionPathExists=!/var/lib/bottlerocket/early-boot-config.ran
+# Prevent manually starting/stopping the service
```
bcressey: We should say "why" rather than "what":

```suggestion
# Block manual interactions with this service, since it could leave the system in an
# unexpected state.
```

arnaldo2792:
In the last forced pushed I addressed the comments left by @bcressey:

* Typo in "early-boot-config.service"
* Improved documentation on why we switched from `Requires` to `Wants` in "early-boot-config.service"
* Improved documentation on why we added the new settings in the "oneshot" units

bcressey:
I'm happy with the code changes, but I'm planning to pull down the branch and try some contrived scenarios to see if the system behaves as expected.

arnaldo2792:
* Updated commit message
* Added comments in `mark-successful-boot` and `send-boot-success`

arnaldo2792:
* Rebase upstream
* Add `Requires=storewolf.service` in `early-boot-config.service`

packages/os/early-boot-config.service
```diff
@@ -2,10 +2,17 @@
 Description=Bottlerocket userdata configuration system
 # Need network online to talk to IMDS.
 After=network-online.target apiserver.service
```
bcressey:
```suggestion
After=network-online.target apiserver.service storewolf.service
```

bcressey (approved):

packages/os/early-boot-config.service
arnaldo2792: Fixed :+1:

arnaldo2792:
* Added `storewolf.service` in `After` for `early-boot-config`

sources/api/host-containers/src/main.rs
```diff
@@ -202,23 +202,30 @@ impl<'a> SystemdUnit<'a> {
...
+
     fn enable_and_start(&self) -> Result<()> {
-        // We enable/start units with --no-block to work around cyclic dependency issues at boot
-        // time.  It would probably be better to give systemd more of a chance to tell us that
-        // something failed to start, if dependencies can be resolved in another way.
         command(
             SYSTEMCTL_BIN,
-            &["enable", &self.unit, "--now", "--no-block"],
-        )
+            &["enable", &self.unit, "--now"]
```
webern: Is the removal of --no-block intentional?

webern (approved):
`/usr/bin/metricdog send-boot-success` will return zero even if the network is not ready. If that is the case then you would see an error in the journal. If it succeeds there would be no output in the journal unless you run it like this `/usr/bin/metricdog --log-level debug send-boot-success`. Other than making sure send-boot-success works, looks good to me. Nice work!

sources/api/host-containers/src/main.rs
bcressey: Before the refactor, this would hold up the boot because of a circular dependency - we couldn't start the container until containerd started, but containerd couldn't start until we were finished applying settings.

I expect it's better to keep `--no-block`, because nothing is waiting on this process and so it may as well exit quickly, and because it could cause trouble later if we start paying more attention to processes spawned by apiserver.

arnaldo2792: I added `--no-block` to the command in the last forced push I did, as well as the `log-level` flag.

arnaldo2792:
* Added `--no-block` to the `enable_and_start` command
* Added `--log-level debug` to `send-boot-success`

arnaldo2792:
* @webern  Removed `--log-level debug` from `send-boot-success` since we don't want the entire debug logs here, I verified that `send-boot-success` works.

sources/api/host-containers/src/main.rs
```diff
@@ -229,14 +239,16 @@ where
...
+
+    trace!("stdout: {}", stdout);
     trace!("stderr: {}", String::from_utf8_lossy(&output.stderr));
 
     ensure!(
         output.status.success(),
         error::CommandFailure { bin_path, output }
     );
-    Ok(())
+    Ok(stdout)
```
jahkeup, line 251: Is there a near-future use case where stdout is handled by callers? If not, then I'd rather this stay as a unit `()` return. The `trace!` can keep printing the value, but I'm not sure it ought be returned to callers based on the changes at call sites in this PR.

packages/release/activate-configured.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
```
jahkeup, line 1: I'm a bit confused by the use of procedural units to provide what I think are intended to be "hard" synchronization points. systemd provides the means to express the same graph without intermediate `activate-<next>.service` units by way of `<next>.target`s and job ordering. 

To me, the use of `systemctl isolate` seems error prone as unexpressed dependencies may result in a half-working system where a modification of transitive dependency could actually drop a unit from the final set of running units.

This change is somewhat new to me so I might be missing some context - so please forgive me if this was already discussed:

* Did we determine that a `configured-pre.target` and `configured.target` would be unable to express the desired ordering?

* Does `systemctl isolate` provide desired functionality that unit ordering and dependency expression does not? What do we gain by injecting a corrective set of jobs (with `systemctl isolate`) at runtime vs. using dependency ordering?

* Can we avoid replacing the default unit and instead order our units within the default graph?

sources/api/host-containers/src/main.rs
arnaldo2792: I used the `stdout` a little down as part of this change. I use the output to check what's the current default target, and force enable the host containers.

packages/release/activate-configured.service
arnaldo2792: > Did we determine that a `configured-pre.target` and `configured.target` would be unable to express the desired ordering?

Yes, since the problem that we have is that systemd doesn't start runtime created units (like host containers) unless we force them to start (a.k.a `systemctl enable <> --now`). Force starting units results in units starting/stopping multiple times if any of their dependencies fails to start.

>Does `systemctl isolate` provide desired functionality that unit ordering and dependency expression does not? What do we gain by injecting a corrective set of jobs (with `systemctl isolate`) at runtime vs. using dependency ordering?

Yes, since we can dynamically update the jobs graph at runtime which allows us to add more dependencies to the targets if needed (a.k.a. bootstrap containers). And we gain the ability to dynamically "update" the dependency graph and load the units at the correct time, without forcing starts with `systemctl enable --now`

> Can we avoid replacing the default unit and instead order our units within the default graph?

No, we need `preconfigured` to be the default target for the boot, and transition along the targets to load the systemd services (potentially created during `preconfigured`) without force start them (like we used to do with `host-contaieners`).

jahkeup: Thanks for writing back here. The details will be helpful to maybe-future us scratching our heads reading this. 

We discussed this a bit more offline - this all is roughly what I suspected. I think we should stick with the solution you have implemented here. If we run into issues later on then we can revisit it.

I'm not a fan of `systemctl isolate` by any means, but really this incidental complexity is `systemd`'s fault 😆

sources/api/host-containers/src/main.rs
jahkeup: Thanks, I missed that usage!

jahkeup (approved):
I'm okay moving forward with this - though I'd like us to use another approach (if that's possible with what systemd provides) if we run into significant issues with `systemctl isolate` down the road.

:shipit:

PR #1425, opened 2021-03-29
Add support for qcow2 as an image format.

mikalstill:
**Issue number:**

None, should every PR have an associated issue? CONTRIBUTING.md appears ambigious on this point.

**Description of changes:**

Add support for qcow2 as an output format for iamges. qcow2 is pretty common in terms of image formats. Its what OpenStack, KVM, and so on use as their VM image format. This patch adds support
for creating images in that format. 

**Testing done:**

I have verified the output images by booting them on a KVM system.

**Terms of contribution:**

### DIFF
--- a/tools/rpm2img
+++ b/tools/rpm2img
@@ -15,7 +15,7 @@ for opt in "$@"; do
 done
 
 case "${OUTPUT_FMT}" in
-   raw|vmdk) ;;
+   raw|qcow2|vmdk) ;;
    *)
       echo "unexpected image output format '${OUTPUT_FMT}'" >&2
       exit 1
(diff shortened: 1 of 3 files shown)


tools/rpm2img
```diff
@@ -219,18 +219,22 @@ dd if="${BOTTLEROCKET_DATA}" of="${DATA_IMAGE}" conv=notrunc bs=1M seek=1
...
 elif [[ ${OUTPUT_FMT} == "vmdk" ]]; then
   # Stream optimization is required for creating an Open Virtual Appliance (OVA)
   qemu-img convert -f raw -O vmdk -o subformat=streamOptimized "${DISK_IMAGE}" "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.vmdk"
   qemu-img convert -f raw -O vmdk -o subformat=streamOptimized "${DATA_IMAGE}" "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.vmdk"
-  chown 1000:1000 "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.vmdk" \
-    "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.vmdk"
+  IMAGE_EXTENSION="vmdk"
 fi
+chown 1000:1000 "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.${IMAGE_EXTENSION}" \
+  "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.${IMAGE_EXTENSION}"
```
bcressey: We can just `chown` everything under `${OUTPUT_DIR}` - then we don't need to keep track of `IMAGE_EXTENSION`, and we only need to run it once at the very end.

```suggestion
find "${OUTPUT_DIR}" -type f -print -exec chown 1000:1000 {} \;
```

webern (approved):

tools/rpm2img
```diff
@@ -219,18 +219,22 @@ dd if="${BOTTLEROCKET_DATA}" of="${DATA_IMAGE}" conv=notrunc bs=1M seek=1
 sgdisk -v "${DISK_IMAGE}"
 sgdisk -v "${DATA_IMAGE}"
 
+IMAGE_EXTENSION="img.lz4"
 if [[ ${OUTPUT_FMT} == "raw" ]]; then
   lz4 -vc "${DISK_IMAGE}" >"${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.img.lz4"
   lz4 -vc "${DATA_IMAGE}" >"${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.img.lz4"
-  chown 1000:1000 "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.img.lz4" \
-    "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.img.lz4"
+elif [[ ${OUTPUT_FMT} == "qcow2" ]]; then
```
jpculp, line 225: Rather than having an `elif` for each format, could we just do an `else`? `raw` and `vmdk` are special cases since `raw`'s extension ends up being `img.lz4` and `vmdk` requires some extra parameters, but other formats like `qcow2` and `vdi` could be in a more simple catch-all else statement like:

```
qemu-img convert -f raw -O "${OUTPUT_FMT}" "${DISK_IMAGE}" "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.${OUTPUT_FMT}"
qemu-img convert -f raw -O "${OUTPUT_FMT}" "${DATA_IMAGE}" "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.${OUTPUT_FMT}"
```
Users would still be limited to formats in the enum in `manifest.rs`, but we could reduce future code changes to just edge-cases here in `rpm2img`.

mikalstill:
Thanks for the comments. I was going for the minimum change, I'll take a look at the proposed comments tomorrow and incorporate them.

tools/rpm2img
bcressey: I'd rather error out in the "else" branch than pick a format that wasn't requested and somehow we made it down here.

I don't think it's safe to assume that we can invoke 'qemu-img' and get the desired format with no additional options, or even that 'qemu-img' is the correct tool.

I like the "elif" branch and the overall approach here.

tools/rpm2img
mikalstill: Done.

tools/rpm2img
mikalstill: I'm going to assume that the emojis win here.

tools/rpm2img
```diff
@@ -222,16 +222,17 @@ sgdisk -v "${DATA_IMAGE}"
...
   qemu-img convert -f raw -O vmdk -o subformat=streamOptimized "${DATA_IMAGE}" "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.vmdk"
-  chown 1000:1000 "${OUTPUT_DIR}/${DISK_IMAGE_BASENAME}.vmdk" \
-    "${OUTPUT_DIR}/${DATA_IMAGE_BASENAME}.vmdk"
 fi
 
+find "${OUTPUT_DIR}" -type f -print -exec chown 1000:1000 {} \;
+
 lz4 -9vc "${BOOT_IMAGE}" >"${OUTPUT_DIR}/${BOOT_IMAGE_NAME}"
 lz4 -9vc "${VERITY_IMAGE}" >"${OUTPUT_DIR}/${VERITY_IMAGE_NAME}"
 lz4 -9vc "${ROOT_IMAGE}" >"${OUTPUT_DIR}/${ROOT_IMAGE_NAME}"
```
bcressey, line 236: nit: we can move the `find` command below these lz4 steps, and delete the other `chown` command since it's redundant.

bcressey:
LGTM apart from the nitpick. Can you squash your two commits together before we merge?

tools/rpm2img
mikalstill: Done.

bcressey (approved):
Looks great, thank you!

jpculp (approved):

webern (approved):

zmrow (approved):
:cake:

zmrow:
Thanks again for the contribution!  Nice work.

Just for future reference and since you had a question about it: `CONTRIBUTING.md` [mentions opening an issue](https://github.com/bottlerocket-os/bottlerocket/blob/develop/CONTRIBUTING.md#contributing-via-pull-requests) before a PR.  This helps us discuss the changes before you start working on them, and avoids any wasted work if we're already working on it!

> You open an issue first to discuss any significant work - we would hate for your time to be wasted.

PR #1431, opened 2021-03-30
Add support for specifying a variant's supported architectures

### LINKED ISSUE #1404: buildsys: Add ability for variants to support specific architectures
**What I'd like:**
Authors of variants should be able to choose whether or not they will support an architecture.  The variant should only build for architectures specified.  Ideally, this would be something that could be specified in the variant's `Cargo.toml`, similar to the `image-format`.  

```
supported_arches = ["x86_64", "aarch64"]
```

We could add additional logic into `Makefile.toml` to outlaw certain combinations of arch/variant, but this would get out of control quickly and we certainly don't need that many more lines of bash. :smile:

zmrow:
**Issue number:**
Fixes #1404 
Related to #1393 

**Description of changes:**
```
buildsys: Add ability for variants to specify supported architectures

This change adds an additional key that can be specified in a variant's
`Cargo.toml`, `supported-arches`.  It is a list and the supported enum
values are `x86_64` and `aarch64`.  If `supported-arches` is specified,
the current `BUILDSYS_ARCH` is checked against the list.  If
`supported-arches` is not specified, the build continues as before.
```

```
 vmware-dev: Support x86_64 only

This change adds the `supported-arch` key to `vmware-dev`'s `Cargo.toml`
and ensure's the variant will only be built for `x86_64`.
```

**Testing done:**
* Build an `aws-k8s-1.18` image for both `x86_64` and `aarch64`
* Successfully build `vmware-dev` for `x86_64`
* Attempt to build `vmware-dev` for `aarch64` and it correctly fails

```Caused by:
  process didn't exit successfully: `<long path>/bottlerocket/bottlerocket/variants/vmware-dev/target/debug/build/vmware-dev-bce40105c9e5560f/b
...
  --- stderr
  BuildAttempt: Unsupported architecture 'aarch64', check the variant's Cargo.toml for supported architectures
```

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/main.rs
+++ b/tools/buildsys/src/main.rs
@@ -16,10 +16,10 @@ mod spec;
 
 use builder::{PackageBuilder, VariantBuilder};
 use cache::LookasideCache;
-use manifest::ManifestInfo;
+use manifest::{ManifestInfo, SupportedArch};
 use project::ProjectInfo;
 use serde::Deserialize;
-use snafu::ResultExt;
+use snafu::{ensure, ResultExt};
 use spec::SpecInfo;
 use std::env;
 use std::path::PathBuf;
@@ -55,6 +55,22 @@ mod error {
         Environment {
             var: String,
             source: std::env::VarError,
+        },
+
+        #[snafu(display("Unknown architecture: '{}'", arch))]
+        UnknownArch {
+            arch: String,
+            source: serde_plain::Error,
+        },
+
+        #[snafu(display(
+            "Unsupported architecture {}, this variant supports {}",
+            arch,
+            supported_arches.join(", ")
+        ))]
+        UnsupportedArch {
+            arch: String,
+            supported_arches: Vec<String>,
         },
     }
 }
@@ -102,10 +118,17 @@ fn run() -> Result<()> {
 }
 
 fn build_package() -> Result<()> {
-    let manifest_dir: PathBuf = getenv("CARGO_MANIFEST_DIR")?.into();
     let manifest_file = "Cargo.toml";
     println!("cargo:rerun-if-changed={}", manifest_file);
 
+    let root_dir: PathBuf = getenv("BUILDSYS_ROOT_DIR")?.into();
+    let variant = getenv("BUILDSYS_VARIANT")?;
+    let variant_manifest_path = root_dir.join("variants").join(variant).join(manifest_file);
+    let variant_manifest =
+        ManifestInfo::new(variant_manifest_path).context(error::ManifestParse)?;
+    supported_arch(&variant_manifest)?;
+
+    let manifest_dir: PathBuf = getenv("CARGO_MANIFEST_DIR")?.into();
     let manifest =
         ManifestInfo::new(manifest_dir.join(manifest_file)).context(error::ManifestParse)?;
 
@@ -165,6 +188,8 @@ fn build_variant() -> Result<()> {
     let manifest =
         ManifestInfo::new(manifest_dir.join(manifest_file)).context(error::ManifestParse)?;
 
+    supported_arch(&manifest)?;
+
     if let Some(packages) = manifest.included_packages() {
         let image_format = manifest.image_format();
         VariantBuilder::build(&packages, image_format).context(error::BuildAttempt)?;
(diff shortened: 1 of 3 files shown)


tools/buildsys/src/manifest.rs
```diff
@@ -158,6 +164,7 @@ pub(crate) struct BuildPackage {
 pub(crate) struct BuildVariant {
     pub(crate) included_packages: Option<Vec<String>>,
     pub(crate) image_format: Option<ImageFormat>,
+    pub(crate) supported_arches: Option<Vec<SupportedArch>>,
```
bcressey: If we deserialize this into an `Option<HashSet<SupportedArch>>`, it might make the existence test easier.

tools/buildsys/src/main.rs
```diff
@@ -167,7 +167,9 @@ fn build_variant() -> Result<()> {
 
     if let Some(packages) = manifest.included_packages() {
         let image_format = manifest.image_format();
-        VariantBuilder::build(&packages, image_format).context(error::BuildAttempt)?;
+        let supported_arches = manifest.supported_arches();
+        VariantBuilder::build(&packages, image_format, supported_arches)
+            .context(error::BuildAttempt)?;
```
bcressey: We also want the supported arches check when building packages.

tools/buildsys/src/manifest.rs
zmrow: Oh great point!  I always forget about that handy `contains()` method. :)

tools/buildsys/src/main.rs
webern: But we don't have that information when building packages, do we, since it comes from the metadata in the variant's cargo.toml?

tools/buildsys/src/manifest.rs
```diff
@@ -167,6 +174,13 @@ pub(crate) enum ImageFormat {
     Vmdk,
 }
 
+#[derive(Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub(crate) enum SupportedArch {
+    X86_64,
```
webern, line 182: That's a suspicious variant name wrt Rust style. Not sure what to suggest though.

webern (approved):

tools/buildsys/src/main.rs
zmrow: We are able to get the information we need from the environment! :)

zmrow:
^ addresses @bcressey 's comments

zmrow:
The build now stops before it tries to build packages if the current variant's supported architectures doesn't support the current architecture.

tools/buildsys/src/manifest.rs
jpculp: Amd64 and Arm64 are always options, but they are annoyingly similar if you aren't paying attention.

jpculp (approved):

tools/buildsys/src/manifest.rs
tjkirch: > That's a suspicious variant name wrt Rust style. Not sure what to suggest though.

I think we can use lowercase names and tell rust/clippy to ignore the warning?

> Amd64 and Arm64 are always options

We may want to use serde aliases to support these alternate names without having extra variants.

tools/buildsys/src/builder.rs
```diff
@@ -44,10 +45,23 @@ pub(crate) struct PackageBuilder;
...
+        if let Some(supported_arches) = supported_arches {
+            let current_arch: SupportedArch =
+                serde_plain::from_str(&arch).context(error::UnknownArch { arch: &arch })?;
+
+            ensure!(
+                supported_arches.contains(&current_arch),
+                error::UnsupportedArch { arch: &arch }
+            )
+        }
+
```
bcressey: Let's extract this into a helper function in `main.rs` that we can call from `build_package` or `build_variant` before we get to this point. Otherwise we're complicating the function signature and duplicating the logic here.

tools/buildsys/src/builder/error.rs
```diff
@@ -52,6 +52,18 @@ pub(crate) enum Error {
...
     },
+
+    #[snafu(display("Unknown architecture: '{}'", arch))]
+    UnknownArch {
+        arch: String,
+        source: serde_plain::Error,
+    },
+
+    #[snafu(display(
+        "Unsupported architecture '{}', check the variant's Cargo.toml for supported architectures",
```
tjkirch: We could include the list of supported architectures here, since we have it in the check, and that might be more helpful / immediate...

tools/buildsys/src/manifest.rs
zmrow: I feel like the names are fine, I'd rather not ignore clippy warnings just for a lowercase letter.

I'm leaning towards not adding the aliases.  Our Makefile (which calls `buildsys`) only allows `x86_64` and `aarch64`.  In order to even use an alias, you'd need to be calling `buildsys` directly which is probably pretty unlikely.

zmrow:
^ Addresses the comments!

tjkirch (approved):

bcressey (approved):

PR #1437, opened 2021-04-01
Add aws-k8s-1.20 variant with Kubernetes 1.20 support

### LINKED ISSUE #1386: Setting for local_port_range overlaps with very common listen ports
**Image I'm using:**

Bottlerocket appears to set the local port range to 
```
net.ipv4.ip_local_port_range = 1025 65000
```
In this file:
https://github.com/bottlerocket-os/bottlerocket/blob/v1.0.5/packages/release/release-sysctl.conf#L23

Ports in the 8000-10000 range should be available for services to listen on (e.g. HTTP servers on 8081, common healthcheck ports etc).

**What actually happened:**

Outbound sockets can claim those ports resulting in sporadic failures to bind.  I work on a project that opens port 9090; one of our users reported sporadic failures on bottlerocket because port 9090 gets used at random for outbound connections. 

Make lots of outbound connections, try to bind to say port 8080; it may fail at random if an outbound connection happened to use that port.

### LINKED ISSUE #1429: Kubernetes 1.20 support
Kubernetes 1.20 was released a while ago: https://kubernetes.io/blog/2020/12/08/kubernetes-1-20-release-announcement/

etungsten:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket/issues/1429
https://github.com/bottlerocket-os/bottlerocket/issues/1386

**Description of changes:**

For reference, the addition of 1.19 was in https://github.com/bottlerocket-os/bottlerocket/pull/1256.

Version of kubelet packaged: v1.20.6

```
Date:   Wed Mar 31 17:37:47 2021 -0700

    Add aws-k8s-1.20 variant with Kubernetes 1.20
```

```
Date:   Thu Apr 1 10:59:47 2021 -0700

    kubernetes-1.20: constrain ip_local_port_range
    
    Constrains ephemeral ports for k8s-1.20 variants

```

Regenerated the `0001-always-set-relevant-variables-for-cross-compiling.patch` patch.

Added license clarification for "github.com/storageos/go-api"

**Testing done:**
- [x] Pods run OK

- [x]  Conformance testing with 1.20 control plane and x86_64 aws-k8s-1.20 nodes - all passed

- [x]  Conformance testing with 1.20 control plane and aarch64 aws-k8s-1.20 nodes - all passed

- [x] Confirm the EBS CSI driver still works.

- [x] Check kubelet logs and system logs in general, compared it with the aws-k8s-1.19 variant kubelet logs and found no abnormalities.

Confirmed that the ephemeral port range has been constrained:
```
[ec2-user@ ~]$ sysctl -a | grep net.ipv4.ip_local_port_range
net.ipv4.ip_local_port_range = 32768	60999
```

**Terms of contribution:**


packages/Cargo.toml
```diff
@@ -32,6 +32,7 @@ members = [
     "kubernetes-1.17",
```
tjkirch: For lack of a better place to leave a comment - we also need to update the top-level README.md to list aws-k8s-1.20 in the variant list in `## Variants`.

packages/kubernetes-1.20/kubelet-sysctl.conf
```diff
@@ -0,0 +1,4 @@
+# Ports below 32768 are generally outside of the ephemeral port range
+# Aligning with the port range documented for ECS
+# https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html
```
tjkirch:
```suggestion
# This is generally considered a safe ephemeral port range
```
I'm thinking it may be confusing to point to ECS documentation in a k8s variant, and it seems fine to just have a comment about choosing a safe range?  (Also, based on the linked doc, it seems the range has changed, and it may be confusing what's intended.)

packages/kubernetes-1.20/kubernetes-1.20.spec
```diff
@@ -0,0 +1,109 @@
...
+%{_cross_unitdir}/kubelet.service
+%dir %{_cross_templatedir}
+%{_cross_templatedir}/kubelet-env
+%{_cross_templatedir}/kubelet-config
+%{_cross_templatedir}/kubelet-kubeconfig
+%{_cross_templatedir}/kubelet-bootstrap-kubeconfig
+%{_cross_templatedir}/kubelet-exec-start-conf
+%{_cross_templatedir}/kubernetes-ca-crt
+%{_cross_tmpfilesdir}/kubernetes.conf
+%{_cross_sysctldir}/99-k8s.conf
```
tjkirch:
```suggestion
%{_cross_sysctldir}/90-k8s.conf
```
Suggesting `90` because the highest numbers are typically meant for users (so they can override) and it matches what we did for the ECS variant.  (the agent)

packages/kubernetes-1.20/0001-always-set-relevant-variables-for-cross-compiling.patch
```diff
@@ -0,0 +1,77 @@
...
+   export GOARCH=${platform##*/}
+ 
+-  # Do not set CC when building natively on a platform, only if cross-compiling from linux/amd64
+-  if [[ $(kube::golang::host_platform) == "linux/amd64" ]]; then
+-    # Dynamic CGO linking for other server architectures than linux/amd64 goes here
+-    # If you want to include support for more server platforms than these, add arch-specific gcc names here
+-    case "${platform}" in
+-      "linux/arm")
+-        export CGO_ENABLED=1
+-        export CC=${KUBE_LINUX_ARM_CC:-arm-linux-gnueabihf-gcc}
```
tjkirch, line 26: Hmm.  In previous versions, these were hardcoded, so it made sense to unilaterally override them with the `-bottlerocket-` versions we need.  Is there any reason we'd want to allow the `$KUBE_LINUX_ARM_CC`-type overrides in our replacement code?  I don't have background on these.

packages/kubernetes-1.20/clarify.toml
```diff
@@ -0,0 +1,62 @@
...
+expression = "MIT AND BSD-3-Clause AND Apache-2.0"
+license-files = [
+    { path = "LICENSE", hash = 0xad378ed2 },
+    { path = "LICENSE-THIRD-PARTY", hash = 0x546425eb },
+    { path = "lint/LICENSE", hash = 0xc6b58232 },
+    { path = "ssa/LICENSE", hash = 0xe656fb62 },
+]
+
+[clarify."github.com/storageos/go-api"]
+expression = "MIT AND BSD-3-Clause"
```
tjkirch:
```suggestion
expression = "MIT AND BSD-2-Clause"
```
https://github.com/storageos/go-api/blob/master/LICENCE#L30-L34

packages/kubernetes-1.20/kubelet-config
```diff
@@ -0,0 +1,47 @@
...
+authorization:
+  mode: Webhook
+  webhook:
+    cacheAuthorizedTTL: 5m0s
+    cacheUnauthorizedTTL: 30s
+{{~/if}}
+clusterDomain: {{settings.kubernetes.cluster-domain}}
+clusterDNS:
+- {{settings.kubernetes.cluster-dns-ip}}
+resolvConf: "/etc/resolv.conf"
```
tjkirch: We'll need to merge in the settings from #1388 here.

packages/kubernetes-1.20/kubelet.service
```diff
@@ -0,0 +1,32 @@
...
+RestartForceExitStatus=SIGPIPE
+RestartSec=5
+Delegate=yes
+KillMode=process
+CPUAccounting=true
+MemoryAccounting=true
+
+[Install]
+WantedBy=multi-user.target
+RequiredBy=mark-successful-boot.service
```
tjkirch: This went away in #1423.

sources/models/src/aws-k8s-1.20
```diff
@@ -0,0 +1 @@
+aws-k8s-1.15
```
tjkirch:
```suggestion
aws-k8s-1.19
```
To match the existing pattern; I'm not opposed to changing the pattern, but while it's there we should be consistent.  (I think the pattern does help in case we determine we need to change the model for 1.18+, for example; we can just change 1.18 and not need to repoint the links following it.)

packages/kubernetes-1.20/kubelet-config
```diff
@@ -0,0 +1,47 @@
+---
```
tjkirch, line 1: We'll need to make sure we get the changes from:
* https://github.com/bottlerocket-os/bottlerocket/pull/1388
* https://github.com/bottlerocket-os/bottlerocket/pull/1465
* https://github.com/bottlerocket-os/bottlerocket/pull/1472

webern: And #1408 changes variant Cargo.toml and package Cargo.toml patterns a bit.

sources/models/src/aws-k8s-1.20
arnaldo2792: k8s 1.20 will use `settings.kernel.lockdown=integrity`, so you won't be able to just symlink 1.19, we will need to write its own `defaults.d` folder.

tjkirch: Possibly with symlinks inside, and an override / customization file, rather than duplicating the rest.

etungsten:
Push above rebases onto develop.
Create separate models files for aws-k8s-1.20.

sources/models/src/aws-k8s-1.20
etungsten: The files in `defaults.d` were already symlinked (except for `50-aws-k8s.toml`). I symlinked `mods.rs` to aws-k8s-1.19's `mods.rs`

etungsten:
Push above addresses majority of @tjkirch comments.
  - Replays several changes:
    - #1388
    - #1472
  - Fixes the license info in `clarify.toml`
  - Adds aws-k8s-1.20 to the list of variants in the README
  - Updates documentation for the sysctl file.
  - Changes the prefix of the destination filename for the sysctl-conf file.
  
For https://github.com/bottlerocket-os/bottlerocket/pull/1465, we decided offline that this will be the first variant where we introduce `cpuManagerPolicy: static` as a hardcoded default.

packages/kubernetes-1.20/0001-always-set-relevant-variables-for-cross-compiling.patch
etungsten: @bcressey can you take a look at this?

etungsten:
Push above rebases onto develop to get changes from https://github.com/bottlerocket-os/bottlerocket/pull/1408 and https://github.com/bottlerocket-os/bottlerocket/pull/1494

etungsten:
Push above applies changes from  https://github.com/bottlerocket-os/bottlerocket/pull/1408 to the `aws-k8s-1.20` variant.

etungsten:
Push above adds the missing Cargo.lock changes.

etungsten:
Push above applies changes from https://github.com/bottlerocket-os/bottlerocket/pull/1494 to the aws-k8s-1.20 variant.

etungsten:
Push above adopts https://github.com/bottlerocket-os/bottlerocket/pull/1295 for aws-k8s-1.20

etungsten:
Push above updates the kubernetes-1.20 package to Kubernetes v1.20.6

variants/README.md
```diff
@@ -52,6 +52,13 @@ It supports self-hosted clusters and clusters managed by [EKS](https://aws.amazo
 
 This variant is compatible with Kubernetes 1.19, 1.20, and 1.21 clusters.
 
+### aws-k8s-1.20: Kubernetes 1.20 node
```
webern, line 55: Should we say this is a preview and not released yet? Then change it when we release an AMI?

etungsten: I don't think preview is the right word since it's something we've tested and it's intended to work just like the other k8s variants. 

As for "not released yet", I can go ahead and add that, but I feel like people should be able to look at our releases and changelogs to know what each official release contains. This applies not just for new variants but new functionality as well (going through our README and marking every unreleased change and then un-marking them after the release seems hard to keep track of).

sources/models/src/aws-k8s-1.20/mod.rs
```diff
@@ -0,0 +1 @@
+../aws-k8s-1.19/mod.rs
```
tjkirch, line 1: Now that these link up rather than down, I think we should consider moving 19 to 20, and linking 19 to 20, to keep a consistent pattern, otherwise 19 looks like some special base of operations.  (It's currently the default variant, but that's irrelevant with respect to the model.)

sources/models/README.md
```diff
@@ -42,6 +42,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/aws-k8s-1.19/mod.rs)
 * [Default settings](src/aws-k8s-1.19/defaults.d/)
 
+### aws-k8s-1.20: Kubernetes 1.20
+
+* [Model](src/aws-k8s-1.15/mod.rs)
+* [Default settings](src/aws-k8s-1.15/defaults.d/)
```
tjkirch: These should point at 1.19 now that 1.15 is removed.

sources/models/src/lib.rs
```diff
@@ -39,6 +39,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/aws-k8s-1.19/mod.rs)
 * [Default settings](src/aws-k8s-1.19/defaults.d/)
 
+## aws-k8s-1.20: Kubernetes 1.20
+
+* [Model](src/aws-k8s-1.15/mod.rs)
+* [Default settings](src/aws-k8s-1.15/defaults.d/)
```
tjkirch: (mentioned in the generated README file, but this is the source - need 1.19 here)

packages/kubernetes-1.20/0001-always-set-relevant-variables-for-cross-compiling.patch
bcressey: Upstream is now a bit closer to what we want, which is to always use our known cross compilers. We're still missing a way to express "always cross-compile, even if we're on linux/amd64". I expect we could come up with a patch to bridge what's left.

For now, your port of the patch looks good.

sources/models/src/aws-k8s-1.20/mod.rs
etungsten: Currently we're symlinking the entire aws-k8s-1.19 directory. We can't symlink the entire aws-k8s-1.20 for other variants since aws-k8s-1,20 has it's own overrides (specifically `defaults.d/50-aws-k8s.toml`). If we want to go that route, then we would be symlinking individual files. So for aws-k8s-1.16, aws-k8s-1.17, and aws-k8s-1.18, their `defaults.d/50-aws-k8s.toml` would have to be symlinked to aws-k8s-1.19.

Maybe each variant should just have its own `defaults.d/50-aws-k8s.toml`?

tjkirch: Oh, no, I was just thinking at the wrong level, sorry.  No need to duplicate configs on all the old variants.

What I might suggest instead is refactoring 50-aws-k8s.toml so that we have pieces that apply to every variant, and then just some smaller overrides in variants as required.  We could even do a simpler strategy for now, where 1.20 links to the existing 50-aws-k8s.toml, plus has a 51 link to a new lockdown-integrity.toml file.

etungsten:

- Splits out kernel lockdown default setting into its own separate shared default toml file.
- Symlinks aws-k8s-1.20's `50-aws-k8s.toml` to aws-k8s-1.19's `50-aws-k8s.toml`
- Fixes documentation

I ran sonobuoy conformance testing for kubelet v1.20.6 and everything still passed. The testing description in the PR is up to date.

tjkirch (approved):

webern (approved):

etungsten:
Push above adds a commit that adds `systemd.unified_cgroup_hierarchy=1` to `aws-k8s-1.20`'s kernel parameters.

I've built and image and verified that it works:

```
bash-5.0# dmesg
[    0.000000] Linux version 5.4.105 (builder@buildkitsandbox) (gcc version 9.3.0 (Buildroot 2020.02.2)) #1 SMP Tue Apr 20 00:03:24 UTC 2021
[    0.000000] Command line: BOOT_IMAGE=(hd0,gpt2)/vmlinuz systemd.unified_cgroup_hierarchy=1 ....
```

The host is running with a unified cgroup hierarchy (cgroupv2)
```
bash-5.0# mount -l 
cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime,seclabel)
```


kubelet is running fine, the node joins the cluser fine, can run pods.

variants/aws-k8s-1.20/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+
+[package.metadata.build-variant]
+included-packages = [
+    "aws-iam-authenticator",
+    "cni",
+    "cni-plugins",
+    "kubelet-1.20",
+    "release",
+]
+kernel-parameters = [
```
tjkirch, line 20: Now also have #1513 to integrate, if it merges first...

bcressey (approved):

arnaldo2792 (approved):
:boot:

etungsten:
Push above integrates changes from https://github.com/bottlerocket-os/bottlerocket/pull/1513

etungsten:
Push above rebases onto current develop to be on top of the changes from https://github.com/bottlerocket-os/bottlerocket/pull/1513.

Tested things they still work.

tjkirch (approved):

Issue #1389, opened 2021-03-12
Improve restart-commands: pass list of changed settings that's triggering the restart


etungsten:
> ^ I think it could be better to improve the way restart-commands are run so that they're always given the list of settings that have changed, rather than the setting name having to be in the restart-command itself.  That way we can handle dynamic settings, and not have to inspect the system to see what changed.
>For example, right now, `services.motd` could have a restart-command like `my-command settings.motd` because we know `services.motd` is only associated with one setting, but `services.host-containers` is associated with any host container name under `settings.host-containers`, so we can't pass (hardcode) a single setting.  Instead, if the command was given the changed settings, it could alter its behavior based on what changed, like in this case where we need to know which host container changed.  (It could also give us the ability to re-use restart command helpers more often, if they can branch on setting.)

_Originally posted by @tjkirch in https://github.com/bottlerocket-os/bottlerocket/issues/1230#issuecomment-740107225_

**What I'd like:**
For restart-command helpers like[ `host-containers`](https://github.com/bottlerocket-os/bottlerocket/tree/develop/sources/api/host-containers) to know which settings are being changed so it can determine which host-containers needs to be restarted. 
Currently host-containers are not automatically restarted if their settings changed. Users have to explicitly disable and re-enable their host-container via the `enabled` setting to see their new host-container settings enacted.

It would be great if [`thar-be-settings` ](https://github.com/bottlerocket-os/bottlerocket/blob/b27869e77c3458cf2ca6d4258f4b0276cc491979/sources/api/thar-be-settings/src/service.rs#L123) can optionally pass the settings being changed to the restart-command being invoked. The restart-command helper would then be able to determine exactly what needs to be done to enact the new setting changes.

**Any alternatives you've considered:**
`host-containers` can try to infer what settings has changed by checking current settings against previously generated environment files. But this approach is potentially nondeterministic and unelegant.

bcressey:
@sanu11 is this done, or if not what's left to do?

etungsten:
I believe this issue has been resolved. Restart command helpers now have access to the list of changed settings via environment variables. The additional work for handling host container restarts when settings change is being tracked in https://github.com/bottlerocket-os/bottlerocket/issues/1531.

PR #1480, opened 2021-04-12
Improve restart commands by passing changed settings

sanu11:
**Issue number:**
#1389 

**Description of changes:**
Pass a space separated string of changed settings to the restart command
Here, I have added a function to return a map of `service, list of changed settings`  so that we get the list of changed settings for a particular service and that list can be passed as a env to the restart command fo that service. Here I have used `env` because some commands are `systemctl` based and passing them as arguments to them wouldn't be right.

```
Author: Sanika <shasanik@amazon.com>
Date:   Sat Apr 10 00:29:23 2021 +0000

    thar-be-settings: pass changed settings to restart commands
    
    When the changed settings affect any services, the changed settings are passed to the restart command of that service.
    The changed settings are passed as space separated strings to the env of the restart command.
    The env is only set when there are any changed settings that affects the service.
```
Handle the changed settings for host-containers. Handle only the host-containers whose setting is changed. 

```
Author: Sanika <shasanik@amazon.com>
Date:   Sat Apr 10 00:43:09 2021 +0000

    host-containers: handle host-containers based on changed settings passed to the restart command
    
    if there are no changed settings passed, say during startup, handle all the host containers.
    if the settings are changed for a particular host container, handle only that host container.
    if the settings are not host-container settings, but do affect host-containers, handle all the host-containers.
```

**Testing done:**
1. Verified that on startup, all the host containers are handled
```
Apr 29 19:29:50 ip-172-31-44-44.us-west-2.compute.internal thar-be-settings[3581]: 19:29:50 [INFO] Command stdout: 19:29:50 [INFO] host-containers started
Apr 29 19:29:50 ip-172-31-44-44.us-west-2.compute.internal thar-be-settings[3581]: 19:29:50 [INFO] Handling host container 'control' during full configuration process
Apr 29 19:29:50 ip-172-31-44-44.us-west-2.compute.internal thar-be-settings[3581]: 19:29:50 [INFO] Host container 'control' is enabled: true, superpowered: false, with source: 328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-control:v0.5.0
Apr 29 19:29:50 ip-172-31-44-44.us-west-2.compute.internal thar-be-settings[3581]: 19:29:50 [INFO] Handling host container 'admin' during full configuration process
Apr 29 19:29:50 ip-172-31-44-44.us-west-2.compute.internal thar-be-settings[3581]: 19:29:50 [INFO] Host container 'admin' is enabled: true, superpowered: true, with source: 328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.7.0

```
2. Verified that if a particular host-container setting is changed, only that host-container is handled
```
apiclient set  settings.host-containers.admin.source=328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.7.0
 
Apr 29 19:33:10 ip-172-31-44-44.us-west-2.compute.internal apiserver[4784]: 19:33:10 [TRACE] (1) thar_be_settings: [api/thar-be-settings/src/main.rs:191] Found services: Services({"host-containers": Service { changed_settings: Some({"settings.host-containers.admin.source"}), model: Service { configuration_files: [], restart_commands: ["/usr/bin/host-containers"] } }})

Apr 29 19:33:10 ip-172-31-44-44.us-west-2.compute.internal apiserver[4784]: 19:33:10 [INFO] Command stdout: 19:33:10 [INFO] host-containers started
Apr 29 19:33:10 ip-172-31-44-44.us-west-2.compute.internal apiserver[4784]: 19:33:10 [INFO] Handling host container 'admin' because it's directly affected by changed setting 'settings.host-containers.admin.source' (and maybe others)
Apr 29 19:33:10 ip-172-31-44-44.us-west-2.compute.internal apiserver[4784]: 19:33:10 [INFO] Host container 'admin' is enabled: true, superpowered: true, with source: 328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.7.0
Apr 29 19:33:10 ip-172-31-44-44.us-west-2.compute.internal apiserver[4784]: 19:33:10 [INFO] Not handling host container 'control', no changed settings affect it


apiclient set  settings.host-containers.control.enabled=true

Apr 29 19:34:18 ip-172-31-44-44.us-west-2.compute.internal apiserver[4811]: 19:34:18 [TRACE] (1) thar_be_settings: [api/thar-be-settings/src/main.rs:191] Found services: Services({"host-containers": Service { changed_settings: Some({"settings.host-containers.control.enabled"}), model: Service { configuration_files: [], restart_commands: ["/usr/bin/host-containers"] } }})

Apr 29 19:34:18 ip-172-31-44-44.us-west-2.compute.internal apiserver[4811]: 19:34:18 [INFO] Command stdout: 19:34:18 [INFO] host-containers started
Apr 29 19:34:18 ip-172-31-44-44.us-west-2.compute.internal apiserver[4811]: 19:34:18 [INFO] Handling host container 'control' because it's directly affected by changed setting 'settings.host-containers.control.enabled' (and maybe others)
Apr 29 19:34:18 ip-172-31-44-44.us-west-2.compute.internal apiserver[4811]: 19:34:18 [INFO] Host container 'control' is enabled: true, superpowered: false, with source: 328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-control:v0.5.0
Apr 29 19:34:18 ip-172-31-44-44.us-west-2.compute.internal apiserver[4811]: 19:34:18 [INFO] Not handling host container 'admin', no changed settings affect it

```

3. Verified that if a setting that doesn't belong to `settings.host-containers` but affects the host-containers, all the host-containers are handled
```
apiclient set settings.network.https-proxy="0.0.0.0:9898"

Apr 29 19:34:57 ip-172-31-44-44.us-west-2.compute.internal apiserver[4836]: 19:34:57 [TRACE] (1) thar_be_settings: [api/thar-be-settings/src/main.rs:191] Found services: Services({"docker": Service { changed_settings: Some({"settings.network.https-proxy"}), model: Service { configuration_files: [SingleLineString { inner: "proxy-env" }], restart_commands: ["/bin/systemctl try-restart docker.service"] } }, "host-containers": Service { changed_settings: Some({"settings.network.https-proxy"}), model: Service { configuration_files: [], restart_commands: ["/usr/bin/host-containers"] } }, "containerd": Service { changed_settings: Some({"settings.network.https-proxy"}), model: Service { configuration_files: [SingleLineString { inner: "containerd-config-toml" }, SingleLineString { inner: "proxy-env" }], restart_commands: ["/bin/systemctl try-restart containerd.service"] } }, "host-containerd": Service { changed_settings: Some({"settings.network.https-proxy"}), model: Service { configuration_files: [SingleLineString { inner: "proxy-env" }], restart_commands: ["/bin/systemctl try-restart host-containerd.service"] } }, "ecs": Service { changed_settings: Some({"settings.network.https-proxy"}), model: Service { configuration_files: [SingleLineString { inner: "ecs-config" }], restart_commands: ["/usr/bin/ecs-settings-applier", "/bin/systemctl try-reload-or-restart ecs.service"] } }})

Apr 29 19:34:59 ip-172-31-44-44.us-west-2.compute.internal apiserver[4836]: 19:34:59 [INFO] Command stdout: 19:34:58 [INFO] host-containers started
Apr 29 19:34:59 ip-172-31-44-44.us-west-2.compute.internal apiserver[4836]: 19:34:58 [INFO] Handling host container 'admin' because it's indirectly affected by changed setting 'settings.network.https-proxy' (and maybe others)
Apr 29 19:34:59 ip-172-31-44-44.us-west-2.compute.internal apiserver[4836]: 19:34:58 [INFO] Host container 'admin' is enabled: true, superpowered: true, with source: 328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.7.0
Apr 29 19:34:59 ip-172-31-44-44.us-west-2.compute.internal apiserver[4836]: 19:34:58 [INFO] Handling host container 'control' because it's indirectly affected by changed setting 'settings.network.https-proxy' (and maybe others)
Apr 29 19:34:59 ip-172-31-44-44.us-west-2.compute.internal apiserver[4836]: 19:34:58 [INFO] Host container 'control' is enabled: true, superpowered: false, with source: 328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-control:v0.5.0
```

4. Verified that code doesn't break if we pass in setting that has no affected services

```
 apiclient set settings.metrics.metrics-url=https://metrics.bottlerocket.aws/v1/metrics

Apr 29 19:36:48 ip-172-31-44-44.us-west-2.compute.internal apiserver[6307]: 19:36:48 [TRACE] (1) thar_be_settings: [api/thar-be-settings/src/main.rs:191] Found services: Services({})
Apr 29 19:36:48 ip-172-31-44-44.us-west-2.compute.internal apiserver[6307]: 19:36:48 [INFO] No services are affected, exiting...

```
5. Verified that corresponding services are handled, if we pass in a list of changed_settings that affects different services

```
apiclient set settings.host-containers.admin.source=328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.7.0 settings.motd="helllo"

Apr 29 00:03:45 ip-172-31-33-3.us-west-2.compute.internal apiserver[4812]: 00:03:45 [TRACE] (1) thar_be_settings: [api/thar-be-settings/src/main.rs:191] Found services: Services({"motd": Service { changed_settings: Some({"settings.motd"}), model: Service { configuration_files: [SingleLineString { inner: "motd" }], restart_commands: [] } }, "host-containers": Service { changed_settings: Some({"settings.host-containers.admin.source"}), model: Service { configuration_files: [], restart_commands: ["/usr/bin/host-containers"] } }})

Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [INFO] Restarting affected services...
Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [DEBUG] (1) thar_be_settings::service: Checking for restart-commands for host-containers
Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [INFO] restart commands ["/usr/bin/host-containers"]
Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [DEBUG] (1) thar_be_settings::service: Restart command: "/usr/bin/host-containers"

Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [INFO] Command stdout: 19:37:03 [INFO] host-containers started
Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [INFO] Handling host container 'admin' because it's directly affected by changed setting 'settings.host-containers.admin.source' (and maybe others)
Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [INFO] Host container 'admin' is enabled: true, superpowered: true, with source: 328549459982.dkr.ecr.us-west-2.amazonaws.com/bottlerocket-admin:v0.7.0
Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [INFO] Not handling host container 'control', no changed settings affect it

Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [DEBUG] (1) thar_be_settings::service: Checking for restart-commands for motd
Apr 29 19:37:03 ip-172-31-44-44.us-west-2.compute.internal apiserver[6316]: 19:37:03 [INFO] restart commands []

```
6. Ran upgrade downgrade testing for k8s variants to verify the instances boot up correctly after upgrade/downgrade
7. Verified the above 5 usecases on k8s and ecs variants both.

**Terms of contributiaon:**

### DIFF
--- a/sources/api/host-containers/src/main.rs
+++ b/sources/api/host-containers/src/main.rs
@@ -372,10 +372,8 @@ where
         field: "superpowered",
     })?;
 
-    info!(
-        "Handling host container '{}' which is enabled: {}",
-        name, enabled
-    );
+    info!("Host container '{}' is enabled: {}, superpowered: {}, with source: {}",
+          name, enabled, superpowered, source);
 
     // Create the directory regardless if user data was provided for the container
     let dir = Path::new(PERSISTENT_STORAGE_BASE_DIR).join(name);
@@ -437,8 +435,36 @@ where
     Ok(())
 }
 
+fn is_container_affected(settings: &[&str], container_name: &str) -> bool {
+    if settings.is_empty() {
+        // it means that Bottlerocket is booting - all containers need to be started
+        info!("Handling host container '{}' during full configuration process", container_name);
+        return true;
+    }
+
+    let setting_prefix = "settings.host-containers.";
+    let container_prefix = format!("{}{}.", setting_prefix, container_name);
+
+    for setting in settings {
+        if setting.starts_with(&container_prefix) {
+            info!("Handling host container '{}' because it's directly affected by changed setting '{}' (and maybe others)", container_name, setting);
+            return true;
+        }
+        if !setting.starts_with(&setting_prefix) {
+            // if its some other setting, return true for all host-containers, example: network
+            info!("Handling host container '{}' because it's indirectly affected by changed setting '{}' (and maybe others)", container_name, setting);
+            return true;
+        }
+    }
+    info!("Not handling host container '{}', no changed settings affect it", container_name);
+    return false;
+}
+
 async fn run() -> Result<()> {
     let args = parse_args(env::args());
+    // this env var is passed by thar-be-settings
+    let changed_settings_env = env::var("CHANGED_SETTINGS").unwrap_or("".to_string());
+    let changed_settings: Vec<&str> = changed_settings_env.split_whitespace().collect();
 
     // SimpleLogger will send errors to stderr and anything less to stdout.
     SimpleLogger::init(args.log_level, LogConfig::default()).context(error::Logger)?;
@@ -448,10 +474,13 @@ async fn run() -> Result<()> {
     let mut failed = 0usize;
     let host_containers = get_host_containers(args.socket_path).await?;
     for (name, image_details) in host_containers.iter() {
-        // Continue to handle other host containers if we fail one
-        if let Err(e) = handle_host_container(name, image_details) {
-            failed += 1;
-            error!("Failed to handle host container '{}': {}", &name, e);
+        // handle all host containers during startup
+        // handle the host container that has settings changed during restart
+        if is_container_affected(&changed_settings, name.as_ref()) {
+            if let Err(e) = handle_host_container(name, image_details) {
+                failed += 1;
+                error!("Failed to handle host container '{}': {}", &name, e);
+            }
         }
     }
 
--- a/sources/api/thar-be-settings/src/main.rs
+++ b/sources/api/thar-be-settings/src/main.rs
@@ -189,7 +189,7 @@ async fn run(args: Args) -> Result<(), Box<dyn std::error::Error>> {
             let services =
                 service::get_affected_services(&args.socket_path, Some(changed_settings)).await?;
             trace!("Found services: {:?}", services);
-            if services.is_empty() {
+            if services.0.is_empty() {
                 info!("No services are affected, exiting...");
                 process::exit(0)
             }
--- a/sources/api/thar-be-settings/src/service.rs
+++ b/sources/api/thar-be-settings/src/service.rs
@@ -1,46 +1,94 @@
-use snafu::{ensure, OptionExt, ResultExt};
-use std::collections::HashMap;
-use std::collections::HashSet;
-use std::path::Path;
-use std::process;
-
-use itertools::join;
-
 use crate::{error, Result};
+use itertools::join;
+use snafu::{ensure, OptionExt, ResultExt};
+use std::collections::{HashMap, HashSet};
+use std::path::Path;
+use std::process::Command;
 
-/// Wrapper for the multiple functions needed to go from
-/// a list of changed settings to a Services map
-#[allow(clippy::implicit_hasher)]
+// TODO: thar-be-settings isn't used as a library; declare its modules in main rather than lib so
+// we don't have to expose helper types like this just so we can call related functions in main.
+/// The `Services` type is an augmented version of `model::Services` that also stores the list of
+/// settings that were changed and that are relevant to each service.
+#[derive(Debug, Default)]
+pub struct Services(pub HashMap<String, Service>);
+/// The `Service` type stores the original `model::Service` and the list of settings that have
+/// changed that are relevant to that service.
+#[derive(Debug)]
+pub struct Service {
+    /// The specific list of settings that changed and are relevant to this service.  Will be None
+    /// if the program is running for *all* services, like at startup.
+    pub changed_settings: Option<HashSet<String>>,
+    /// The model's representation of this service.
+    pub model: model::Service,
+}
+
+impl Services {
+    /// Convert from the model's representation of all services to our own, adding in the lists of
+    /// changed settings, if appropriate; pass None if the program is running for *all* services.
+    pub fn from_model_services(
+        input: model::Services,
+        mut all_changed_settings: Option<HashMap<String, HashSet<String>>>,
+    ) -> Self {
+        let mut output = HashMap::new();
+        for (name, model) in input {
+            let changed_settings = all_changed_settings.as_mut().and_then(|s| s.remove(&name));
+            output.insert(
+                name,
+                Service {
+                    changed_settings,
+                    model,
+                },
+            );
+        }
+        Self(output)
+    }
+}
+
+/// Returns a `Services` reflecting the set of services affected by the given changed settings in
+/// `settings_limit`.  If `settings_limit` is None, reflects all known services.
 pub async fn get_affected_services<P>(
     socket_path: P,
     settings_limit: Option<HashSet<String>>,
-) -> Result<model::Services>
+) -> Result<Services>
 where
     P: AsRef<Path>,
 {
-    let service_limit = if let Some(settings_limit) = settings_limit {
-        let setting_to_service_map =
-            get_affected_service_map(socket_path.as_ref(), settings_limit).await?;
-        if setting_to_service_map.is_empty() {
-            return Ok(HashMap::new());
+    let services: Services;
+    if let Some(settings_limit) = settings_limit {
+        // Get the list of affected services for each setting
+        let affected_services =
+            get_affected_service_metadata(socket_path.as_ref(), settings_limit).await?;
+        if affected_services.is_empty() {
+            return Ok(Services::default());
         }
 
-        let service_names = get_affected_service_names(setting_to_service_map);
-        Some(service_names)
-    } else {
-        // No limit, we want all services.
-        None
-    };
+        // Pull out the names of the services so we can ask the API about them.
+        let service_names = affected_services.values().flatten().collect();
+        // Ask the API for its metadata about the affected services.
+        let service_meta = get_service_metadata(socket_path.as_ref(), Some(service_names)).await?;
 
-    let services = get_service_metadata(socket_path.as_ref(), service_limit).await?;
+        // Reverse the mapping, getting the list of changed settings for each service
+        let mut changed_settings = HashMap::new();
+        for (setting, services) in affected_services {
+            for service in services {
+                let settings = changed_settings.entry(service).or_insert_with(HashSet::new);
+                settings.insert(setting.clone());
+            }
+        }
+
+        services = Services::from_model_services(service_meta, Some(changed_settings));
+    } else {
+        // If there was no settings limit, get data for all services.
+        let service_meta = get_service_metadata(socket_path.as_ref(), None).await?;
+        services = Services::from_model_services(service_meta, None);
+    }
 
     Ok(services)
 }
 
-/// Gather the services affected for each setting into a map, or if `settings_limit` is None, all
-/// services
+/// Ask the API which services are affected by the given list of settings.
 #[allow(clippy::implicit_hasher)]
-async fn get_affected_service_map<P>(
+async fn get_affected_service_metadata<P>(
     socket_path: P,
     settings: HashSet<String>,
 ) -> Result<HashMap<String, Vec<String>>>
@@ -62,29 +110,11 @@ where
     Ok(setting_to_services_map)
 }
 
-/// Given a map of Setting to affected Service name, return a
-/// HashSet of affected service names
-fn get_affected_service_names(
-    setting_to_service_map: HashMap<String, Vec<String>>,
-) -> HashSet<String> {
-    // Build a HashSet of names of affected services
-    debug!("Building set of affected services");
-    let mut service_set: HashSet<String> = HashSet::new();
-    for (_, service_list) in setting_to_service_map {
-        for service_name in service_list {
-            debug!("Found {}", &service_name);
-            service_set.insert(service_name);
-        }
-    }
-
-    trace!("Affected service names: {:?}", service_set);
-    service_set
-}
-
-/// Gather the metadata for each Service affected
+/// Ask the API for metadata about the given list of services, or all services if `services_limit`
+/// is None.
 async fn get_service_metadata<P>(
     socket_path: P,
-    services_limit: Option<HashSet<String>>,
+    services_limit: Option<HashSet<&String>>,
 ) -> Result<model::Services>
 where
     P: AsRef<Path>,
@@ -104,8 +134,8 @@ where
 }
 
 /// Call the `restart()` method on each Service in a Services object
-pub fn restart_services(services: model::Services) -> Result<()> {
-    for (name, service) in services {
+pub fn restart_services(services: Services) -> Result<()> {
+    for (name, service) in services.0 {
         debug!("Checking for restart-commands for {}", name);
         service.restart()?;
     }
@@ -119,9 +149,11 @@ trait ServiceRestart {
     fn restart(&self) -> Result<()>;
 }
 
-impl ServiceRestart for model::Service {
+impl ServiceRestart for Service {
     fn restart(&self) -> Result<()> {
-        for restart_command in self.restart_commands.iter() {
+        let restart_commands = &self.model.restart_commands;
+        info!("restart commands {:?}", restart_commands);
+        for restart_command in restart_commands {
             // Split on space, assume the first item is the command
             // and the rest are args.
             debug!("Restart command: {:?}", &restart_command);
@@ -135,8 +167,14 @@ impl ServiceRestart for model::Service {
             trace!("Args: {:?}", &command_strings);
 
             // Go execute the restart command
-            let result = process::Command::new(command)
-                .args(command_strings)
+            let mut process_command = Command::new(command);
+            process_command.args(command_strings);
+            if let Some(ref changed_settings) = self.changed_settings {
+                if !changed_settings.is_empty() {
+                    process_command.env("CHANGED_SETTINGS", join(changed_settings, " "));
+                }
+            }
+            let result = process_command
                 .output()
                 .context(error::CommandExecutionFailure {
                     command: restart_command.as_str(),
@@ -160,28 +198,5 @@ impl ServiceRestart for model::Service {
             );
         }
         Ok(())
-    }
-}
-
-#[cfg(test)]
-mod test {
-    use super::*;
-    use maplit::{hashmap, hashset};
-
-    #[test]
-    fn test_get_affected_service_names() {
-        let input_map = hashmap!(
-            "settings.example".to_string() => vec![
-                "example".to_string(),
-            ],
-            "settings.foobar".to_string() => vec![
-                "example".to_string(),
-                "barbaz".to_string()
-            ]
-        );
-
-        let expected_output = hashset! {"example".to_string(), "barbaz".to_string()};
-
-        assert_eq!(get_affected_service_names(input_map), expected_output)
     }
 }
--- a/sources/api/thar-be-settings/src/config.rs
+++ b/sources/api/thar-be-settings/src/config.rs
@@ -1,3 +1,4 @@
+use crate::service::Services;
 use crate::{error, Result};
 use itertools::join;
 use snafu::ResultExt;
@@ -28,11 +29,11 @@ where
 
 /// Given a map of Service objects, return a HashSet of
 /// affected configuration file names
-pub fn get_config_file_names(services: &model::Services) -> HashSet<String> {
+pub fn get_config_file_names(services: &Services) -> HashSet<String> {
     debug!("Building set of affected configuration file names");
     let mut config_file_set = HashSet::new();
-    for service in services.values() {
-        for file in service.configuration_files.iter() {
+    for service in services.0.values() {
+        for file in &service.model.configuration_files {
             config_file_set.insert(file.to_string());
         }
     }
@@ -117,6 +118,7 @@ impl RenderedConfigFile {
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::service::Services;
     use maplit::{hashmap, hashset};
     use std::convert::TryInto;
 
@@ -132,9 +134,10 @@ mod test {
                 restart_commands: vec!["echo hi".to_string()]
             },
         );
+        let services = Services::from_model_services(input_map, None);
 
         let expected_output = hashset! {"file1".to_string(), "file2".to_string() };
 
-        assert_eq!(get_config_file_names(&input_map), expected_output)
+        assert_eq!(get_config_file_names(&services), expected_output)
     }
 }


sanu11:
This force push addresses the unit test in `thar-be-settings`, `cargo fmt` check and removal of `unwrap` function wherever used

sources/api/host-containers/src/main.rs
```diff
@@ -434,8 +434,21 @@ where
     Ok(())
 }
 
+fn verify_change_for_container(changed_settings: &String, container_name: &str) -> bool {
+    let settings: Vec<&str> = changed_settings.split_whitespace().collect();
+    for setting in settings.iter() {
+        // if its a host-container setting, check for specific host-container
+        // if its some other setting, return true for all host-containers, example: network
+        if setting.contains(container_name) == true || ! setting.contains("host-containers") {
```
sanu11: I just realized may be we should check for `"host-containers.<container_name>"` instead of just `container_name` to be safer, since it guarantees that we are only looking for host-containers with `container_name` name and  not any other substring containing `container_name`

webern: BTW, You can get rid of `== true` since contains returns a bool.

Like you said above, maybe you can create `let setting_prefix = format!("host-containers.{}", container_name")` and then check that `setting.starts_with(setting_prefix)`.

And it might be clearer to break up the conditional:

```rust
if !settings.starts_with("host-container") {
   // we found a setting that might affect all host-containers
  return true;
}

if setting.starts_with(setting_prefix) {
  // we found a setting for our container
  return true;
}

```

sources/api/host-containers/src/main.rs
```diff
@@ -434,8 +434,21 @@ where
     Ok(())
 }
 
+fn verify_change_for_container(changed_settings: &String, container_name: &str) -> bool {
```
webern: I think this function could have a better name.

```suggestion
/// Checks the list of changed settings to determine whether `container_name` needs to be restarted.
fn is_restart_required(changed_settings: &String, container_name: &str) -> bool {
```

sources/api/host-containers/src/main.rs
```diff
@@ -446,10 +459,15 @@ async fn run() -> Result<()> {
     let mut failed = 0usize;
     let host_containers = get_host_containers(args.socket_path).await?;
     for (name, image_details) in host_containers.iter() {
-        // Continue to handle other host containers if we fail one
-        if let Err(e) = handle_host_container(name, image_details) {
-            failed += 1;
-            error!("Failed to handle host container '{}': {}", &name, e);
+        // handle all host containers during startup
+        // handle the host container that has settings changed during restart
+        if changed_settings.len() == 0
```
webern: This check (BTW you can use is_empty()) could be moved into the `verify_change_for_container` function (which I suggest renaming to `is_restart_required`)... so that function would start like this:

```rust
fn is_restart_required(changed_settings: &String, container_name: &str) -> bool {
if changed_settings.empty() {
  // if change_settings is empty, it means that Bottlerocket is booting.
  // all containers need to be started
  return true;
}
// ...

```

It would eliminate another conditional that takes some time to understand when reading.

sources/api/thar-be-settings/src/service.rs
```diff
@@ -4,36 +4,49 @@ use std::collections::HashSet;
...
     } else {
         // No limit, we want all services.
         None
     };
 
-    let services = get_service_metadata(socket_path.as_ref(), service_limit).await?;
+    Ok(settings_limit)
+}
 
+pub async fn get_affected_services<P>(
```
webern: Can you try moving this function above `get_affected_services_settings_map` so that it starts around line 12 like it did before? I think it might help Git produce a better diff.

webern (approved):
I didn't finish, but had a few suggestions.

sources/api/host-containers/src/main.rs
sanu11: Yeah, cool. that would make it clearer. Thanks!

sources/api/thar-be-settings/src/service.rs
sanu11: Yeah sure

sources/api/thar-be-settings/src/service.rs
```diff
@@ -4,36 +4,49 @@ use std::collections::HashSet;
...
     settings_limit: Option<HashSet<String>>,
-) -> Result<model::Services>
+) -> Result<Option<HashMap<String, Vec<String>>>>
 where
     P: AsRef<Path>,
 {
-    let service_limit = if let Some(settings_limit) = settings_limit {
+    let settings_limit = if let Some(settings_limit) = settings_limit {
         let setting_to_service_map =
             get_affected_service_map(socket_path.as_ref(), settings_limit).await?;
```
sanu11: I will refactor function name and some variable names here, since it seems similar and confusing

sanu11:
^Addressed comments by Matt and refactored some function names to be more clear.

sources/api/thar-be-settings/src/service.rs
```diff
@@ -104,10 +114,18 @@ where
 }
 
 /// Call the `restart()` method on each Service in a Services object
-pub fn restart_services(services: model::Services) -> Result<()> {
+/// used &Option, without it I was getting move occured error
```
etungsten:
```suggestion

```

sources/api/thar-be-settings/src/service.rs
```diff
@@ -135,8 +155,21 @@ impl ServiceRestart for model::Service {
...
             // Go execute the restart command
-            let result = process::Command::new(command)
-                .args(command_strings)
+            let mut process_command = process::Command::new(command);
+            process_command.args(command_strings);
+            match changed_settings {
+                Some(value) => {
+                    let changed_settings_string = value.iter().join(" ");
+                    info!(
+                        "restarting service  with changed settings {:?}",
```
etungsten: nit: I think we should move this log message one level up to `restart_services` right before we call `restart` where we have the service name and the list of changed settings. Let's remove it from here.

sources/api/thar-be-settings/src/service.rs
```diff
@@ -166,10 +199,24 @@ impl ServiceRestart for model::Service {
...
+            None => HashSet::new(),
+            Some(b_value) => b_value.iter().collect(),
+        };
+
+        a_value == b_value
+    }
 
     #[test]
-    fn test_get_affected_service_names() {
+    fn test_get_service_setting_map() {
```
etungsten: nit: Might be just me. But I think this test isn't very clear on what it's testing for. Maybe a few comments might help here.

sources/api/host-containers/src/main.rs
```diff
@@ -434,22 +434,47 @@ where
     Ok(())
 }
 
+fn is_restart_required(changed_settings: &String, container_name: &str) -> bool {
```
etungsten: nit: Just a preference and better consistency with the rest of the code-base. 
`&String` can be coerced into `&str` but not the other way around I think?
```suggestion
fn is_restart_required(changed_settings: &str, container_name: &str) -> bool {
```

etungsten (approved):

sources/api/thar-be-settings/src/service.rs
```diff
@@ -4,43 +4,56 @@ use std::collections::HashSet;
...
+        None => None,
+    };
+    // if no services are affected by changed_settings, metadata for all the services is returned
+    let services = get_service_metadata(socket_path.as_ref(), service_names).await?;
+    Ok(services)
+}
 
-        let service_names = get_affected_service_names(setting_to_service_map);
-        Some(service_names)
+pub async fn get_affected_service_setting_map<P>(
```
etungsten: Let's also add a docstring comment to explain what this public function does.

arnaldo2792:
With your changes you addressed this part of #1389:

> It would be great if `thar-be-settings` can optionally pass the settings being changed to the restart-command being invoked. The restart-command helper would then be able to determine exactly what needs to be done to enact the new setting changes.

But @etungsten mentioned:

> Currently host-containers are not automatically restarted if their settings changed. Users have to explicitly disable and re-enable their host-container via the enabled setting to see their new host-container settings enacted.

Should that use case be handled as well in your PR? If so, you need to change `handle_host_container` to restart the systemd units without requiring users to explicitly disable/enable the containers.

sources/api/host-containers/src/main.rs
```diff
@@ -434,22 +434,47 @@ where
...
+            // if its some other setting, return true for all host-containers, example: network
+            return true;
+        }
+    }
+    return false;
+}
+
 async fn run() -> Result<()> {
     let args = parse_args(env::args());
+    let changed_settings = env::var("changed_settings").unwrap_or("".to_string());
```
arnaldo2792: Nit: It will be good to document that this environment variable is passed by `thar-be-settings`

etungsten:
> But @etungsten mentioned:
> 
> > Currently host-containers are not automatically restarted if their settings changed. Users have to explicitly disable and re-enable their host-container via the enabled setting to see their new host-container settings enacted.
> 
> Should that use case be handled as well in your PR? If so, you need to change `handle_host_container` to restart the systemd units without requiring users to explicitly disable/enable the containers.

Ah, that's correct. There needs to be a conditional inside `handle_host_container` to check if the host-container is enabled. If it is, we need to `systemctl try-restart` the unit.

sources/api/thar-be-settings/src/service.rs
```diff
@@ -135,8 +155,21 @@ impl ServiceRestart for model::Service {
...
+                Some(value) => {
+                    let changed_settings_string = value.iter().join(" ");
+                    info!(
+                        "restarting service  with changed settings {:?}",
+                        changed_settings_string
+                    );
+                    process_command.env("changed_settings", changed_settings_string);
+                }
+                None => {}
+            }
```
arnaldo2792: Can you try?

```rust
let process_command = if let Some(changed_settings_string) = changed_settings_string {
    process::Command::new(command)
         .args(command_strings)
         .env("changed_settings", changed_settings_string.iter().join(" "));
} else {
    process::Command::new(command).args(command_strings)
}
```

sources/api/thar-be-settings/src/service.rs
sanu11: ops, that was for my understanding, forgot to remove it haha
thanks!

sanu11:
> > But @etungsten mentioned:
> > > Currently host-containers are not automatically restarted if their settings changed. Users have to explicitly disable and re-enable their host-container via the enabled setting to see their new host-container settings enacted.
> > 
> > 
> > Should that use case be handled as well in your PR? If so, you need to change `handle_host_container` to restart the systemd units without requiring users to explicitly disable/enable the containers.
> 
> Ah, that's correct. There needs to be a conditional inside `handle_host_container` to check if the host-container is enabled. If it is, we need to `systemctl try-restart` the unit.

Okay, I was going to do it, but after our discussion offline, I thought we are supposed to handle only the `what` part (changed_settings) and not the `when` (restart host-containers) part. I will add it here in my next push.

sources/api/thar-be-settings/src/service.rs
sanu11: Yeah sure. I had moved the common part to the top, but with this approach, the empty `None` would go away. I will try it. Thanks!

sources/api/thar-be-settings/src/main.rs
```diff
@@ -179,13 +178,27 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
                 "Requesting affected services for settings: {:?}",
                 &changed_settings
             );
-            let services =
-                service::get_affected_services(&args.socket_path, Some(changed_settings)).await?;
-            trace!("Found services: {:?}", services);
-            if services.is_empty() {
+
+            let service_settings_map = service::get_affected_service_setting_map(
```
webern: This should not take an option and should not return an option (since the only invocation of this function always give Some(x)).

sources/api/thar-be-settings/src/service.rs
```diff
@@ -4,43 +4,56 @@ use std::collections::HashSet;
...
 use crate::{error, Result};
 
-/// Wrapper for the multiple functions needed to go from
-/// a list of changed settings to a Services map
-#[allow(clippy::implicit_hasher)]
+//returns changed settings corresponding to the services
 pub async fn get_affected_services<P>(
     socket_path: P,
-    settings_limit: Option<HashSet<String>>,
+    service_settings_map: &Option<HashMap<String, Vec<String>>>,
```
webern: Please explain this parameter. Why is it an option? Why is it a map?

webern (approved):

sources/api/thar-be-settings/src/main.rs
sanu11: Yeah, would update that. I passed it as an option because earlier `some` and `none` were used for two specific modes, but now since the function is broken down into two `get_affected_service_setting_map` and `get_affected_services` in the first mode, `get_affected_service_setting_map` should not need option I believe and would make change accordingly

sources/api/thar-be-settings/src/service.rs
sanu11: Also, do you think it would be good idea to set empty string value to  key `changed_settings` key for `.env`  if there are no `changed_settings` so that we can get rid of `Option` for `changed_settings` in this function signature. 
Currently I am not setting `.env` for the services if `changed_settings` is empty

or we can just check if string is empty in this function, if so do not set `env`

etungsten: I think that makes sense. We can move that check up one level.

sanu11:
Above force push addresses comments from @webern, @etungsten and @arnaldo2792 

  - Removes option from wherever possible
  - Splits `get_affected_services` into 2 functions, one for getting all services - `get_all_services` and other for getting affected services when there are changed settings - `get_affected_services` and added a `get_services_impl` function which will be called by both the mentioned functions.  This makes code more readable
  - Passes string of `changed_settings` to restart function instead of passing vector
  - Adds comments wherever necessary
  
Will push the changes for checking if the host-container is enabled in new commit

sanu11:
Above force push shortens the commit message

sources/api/thar-be-settings/src/service.rs
```diff
@@ -103,11 +124,37 @@ where
     Ok(service_map)
 }
 
+/// Returns string of space separated settings for given service
+fn get_changed_settings_for_service(
```
tjkirch: I'm having a hard time understanding something here.  I know we only set the environment variable if a service was affected, but does that variable only include settings that are relevant to that service?  It's important because the host-containers code change includes `if !setting.contains("host-containers") { true }` which means we restart a host container if it sees any non-host-container setting.

For example, if the user changed settings.host-containers.x and settings.kubernetes.y in the same commit, and the host-containers service says it's affected by settings.host-containers, and then thar-be-settings runs, will the environment variable contain settings.kubernetes?  If so, then even host container "y" would be restarted.

sanu11: no, it wouldn't contain `settings.kubernetes` since the map of service_setting would be something like `{ host-containers:  settings.host-containers.x, kubernetes: settings.kubernetes} ` so the host container service will only have the settings that affects it which is `settings.host-containers.x`

sanu11: Regrading this ```It's important because the host-containers code change includes if !setting.contains("host-containers") { true } which means we restart a host container if it sees any non-host-container setting.```
So, the settings that are sent to the host-containers are only those that affects the host-containers, say `settings.host-containers.x` or` settings.network.proxy ` The use case for the check you mentioned - `if !setting.contains("host-containers") { true }` is `settings.network`  

With `settings.kubernetes` since host-containers isn't affected by this settings, it would not be passed to the restart command of host-containers thus the check wouldn't affect them.

tjkirch: Ok, I think I understand now.  This function is taking in a map where the keys are service names ("sysctl", "lockdown", "containerd"... all the "services.bla" defined in the model) and the values are the settings prefixes that the model says affect those services.  For example, the model has `services.motd`, and `metadata.settings.motd` says `affected-services = ["motd"]`, so the map would have `"motd": ["settings.motd"]` as a key/value pair.  (It could possibly have other settings prefixes in the list, too, if they also affected the motd service, but that's all we have that touches motd for now.)

So, the environment variable would only contain settings prefixes from that list, i.e. directly from our model.

I think `changed_settings` was throwing me a little bit, because they're not exactly changed settings, they're the settings metadata prefixes from the model.  I wonder if we should list the changed settings instead, since services might only care about particular keys.

sanu11: Yeah, thanks for putting it up together, that's correct.  I wanted to clarify the `value` part here.
So, these are the `changed_settings` that user changes using apiclient, so if a user runs ```apiclient set settings.network.https-proxy="0.0.0.0:9898"   settings.host-containers.admin.control.enabled=false``` for example 

The map would be - { "host-containers" :  [ settings.host-containers.admin.source,   settings.network.https-proxy="0.0.0.0:9898" ]}  
which means ```settings.network.https-proxy="0.0.0.0:9898" and  settings.host-containers.admin.source``` are the `changed_settings` that affects host-containers and those would be passed as parameter to restart_command

tjkirch: Ok, I see... the real data source for the map is the `/metadata/affected-services` API, which returns keys that are full settings names (like `settings.host-containers.admin.source`) and not the metadata prefixes from the model.

sources/api/host-containers/src/main.rs
```diff
@@ -434,22 +434,48 @@ where
     Ok(())
```
tjkirch, line 435: There are some minor typos in the commit messages; I'd recommend the following to fix those and simplify a bit:
```
thar-be-settings: pass changed settings to restart commands

If a service is affected by changed settings, the changed settings are passed
to the restart commands of that service in an environment variable, separated
by spaces.  The variable is only set if there are changed settings that affect
the service, and only the settings relevant to that service are included.
```
and
```
host-containers: restart only modified host containers

If no settings have changed, say during startup, handle all host containers.
If settings are changed for particular host containers, handle only those
host containers.  If the settings are not host-container-specific, but affect
host containers overall, handle all host containers.
```

sources/api/host-containers/src/main.rs
```diff
@@ -434,22 +434,48 @@ where
...
+fn is_restart_required(changed_settings: &str, container_name: &str) -> bool {
+    if changed_settings.is_empty() {
+        // it means that Bottlerocket is booting - all containers need to be started
+        return true;
+    }
+
+    let settings: Vec<&str> = changed_settings.split_whitespace().collect();
+    let setting_prefix = format!("host-containers.{}", container_name);
+
+    for setting in settings.iter() {
```
tjkirch:
```suggestion
    for setting in settings {
```

sources/api/host-containers/src/main.rs
```diff
@@ -434,22 +434,48 @@ where
...
+        if setting.contains(&setting_prefix) {
```
tjkirch: I think we could be a little more careful here to match the string we know will be present.  Something like:
```rust
     let setting_prefix = "settings.host-containers.";
     let container_prefix = format!("{}.{}.", setting_prefix, container_name);
...
     if setting.starts_with(&container_prefix) {
```
(I'm suggesting two variables so I can use `setting_prefix` in a later comment...)

sources/api/host-containers/src/main.rs
```diff
@@ -434,22 +434,48 @@ where
...
+            info!("Found a changed setting for {} container", container_name);
+            return true;
+        }
+        if !setting.contains("host-containers") {
```
tjkirch:
```suggestion
        if !setting.starts_with(&setting_prefix) {
```
(if we split up the settings, per previous suggestion)

sources/api/host-containers/src/main.rs
```diff
@@ -434,22 +434,48 @@ where
     Ok(())
 }
 
+fn is_restart_required(changed_settings: &str, container_name: &str) -> bool {
+    if changed_settings.is_empty() {
+        // it means that Bottlerocket is booting - all containers need to be started
+        return true;
+    }
+
+    let settings: Vec<&str> = changed_settings.split_whitespace().collect();
```
tjkirch: I think it'd be good to do this split right after we get the value from the environment; we know any user of that data is going to want separated strings.

tjkirch (commented):
I'm just leaving a few comments on host-containers because I want to put some more thought into thar-be-settings and see if we can simplify what's going on a bit.

sources/api/thar-be-settings/src/service.rs
sanu11: Yeah right. It will return the affected services for `changed_settings` thus our map would be `affected_services` to `changed_settings`

tjkirch:
>> Should that use case be handled as well in your PR? If so, you need to change `handle_host_container` to restart the systemd units without requiring users to explicitly disable/enable the containers.

> Ah, that's correct. There needs to be a conditional inside `handle_host_container` to check if the host-container is enabled. If it is, we need to `systemctl try-restart` the unit.

I think we should do that in a separate PR.  This PR isn't changing behavior that much, just fixing a bug where we impact unrelated host containers.  Whether impacted host containers are restarted is a separate problem with different implications and testing requirements.

sanu11:
Addressed @tjkirch's comments

sources/api/host-containers/src/main.rs
```diff
@@ -434,22 +434,48 @@ where
     Ok(())
 }
 
+fn is_restart_required(settings: &Vec<&str>, container_name: &str) -> bool {
```
tjkirch:
```suggestion
fn is_restart_required(settings: &[&str], container_name: &str) -> bool {
```
Rust allows you to pass a `Vec` to a function taking a slice `&[]`, but this way you can also pass a real slice, so it's more flexible.

sources/api/thar-be-settings/src/service.rs
```diff
@@ -62,23 +85,21 @@ where
...
+) -> HashMap<String, Vec<String>> {
+    let mut service_settings_map: HashMap<String, Vec<String>> = HashMap::new();
+
+    for (settings_name, service_list) in setting_to_service_map {
         for service_name in service_list {
-            debug!("Found {}", &service_name);
-            service_set.insert(service_name);
+            let settings = service_settings_map
+                .entry(service_name.clone())
+                .or_insert(vec![]);
```
tjkirch:
```suggestion
                .or_insert_with(Vec::new);
```
This way it won't create the Vec unless it's needed for a new entry.

sources/api/thar-be-settings/src/service.rs
```diff
@@ -62,23 +85,21 @@ where
...
+    setting_to_service_map: &HashMap<String, Vec<String>>,
+) -> HashMap<String, Vec<String>> {
+    let mut service_settings_map: HashMap<String, Vec<String>> = HashMap::new();
+
+    for (settings_name, service_list) in setting_to_service_map {
         for service_name in service_list {
-            debug!("Found {}", &service_name);
-            service_set.insert(service_name);
+            let settings = service_settings_map
+                .entry(service_name.clone())
```
tjkirch:
```suggestion
                .entry(service_name)
```

Since we only use this function to transform a map and we don't need the original afterward, if we change the parameter to take ownership of the hashmap rather than reference `&` it, then we can remove this clone.

sources/api/thar-be-settings/src/service.rs
```diff
@@ -62,23 +85,21 @@ where
...
-/// HashSet of affected service names
-fn get_affected_service_names(
-    setting_to_service_map: HashMap<String, Vec<String>>,
-) -> HashSet<String> {
-    // Build a HashSet of names of affected services
-    debug!("Building set of affected services");
-    let mut service_set: HashSet<String> = HashSet::new();
-    for (_, service_list) in setting_to_service_map {
+/// Converts map of changed setting to affected service into affected service to changed setting
+fn get_service_setting_map(
```
tjkirch: I think we could move this code into `get_affected_service_setting_map` to simplify things, with a comment above like "Invert the hashmap so we can index by service name."

sources/api/thar-be-settings/src/service.rs
```diff
@@ -4,43 +4,66 @@ use std::collections::HashSet;
...
-use itertools::join;
+use itertools::{join, Itertools};
 
 use crate::{error, Result};
 
-/// Wrapper for the multiple functions needed to go from
-/// a list of changed settings to a Services map
-#[allow(clippy::implicit_hasher)]
+/// Returns services from the services to setting map
 pub async fn get_affected_services<P>(
```
tjkirch: I'm not sure the two helper functions calling `get_services_impl` are pulling their weight - I'd recommend getting rid of them to reduce the level of function calls.  Then we don't need an `_impl`-named function, which can be a little confusing.

sources/api/thar-be-settings/src/main.rs
```diff
@@ -179,13 +179,20 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
                 "Requesting affected services for settings: {:?}",
                 &changed_settings
             );
-            let services =
-                service::get_affected_services(&args.socket_path, Some(changed_settings)).await?;
-            trace!("Found services: {:?}", services);
-            if services.is_empty() {
+
+            let service_settings_map =
+                service::get_affected_service_setting_map(&args.socket_path, changed_settings)
```
tjkirch: I think my confusion mainly comes from the different ways we're storing settings and services, since some are duplicated, and what "services" actually means in different places.  Before, services was a `model::Services` object, but now we also have maps with just their names so we can find related settings.

Let me throw out an idea, because it might help make my confusion clearer, and at least give us something to start with.  What if we get rid of the change on this line, and change `get_affected_services` to return a new type that has all of that information together.  The new type, perhaps just called `Services`, would effectively be a map where the keys are service names, just like `service_settings_map` and `services` are now, and the values would be another new type (`Service`?) with one field for the relevant setting names and one field for the `Service` objects used in the model.  Then we'd have everything we need, and we can just pass that around instead of needing two different representations of services.  It'd be easy to convert for places where we need a `model::Services` or just the map of services to settings, or we could change those places to understand the new type instead of converting at all.

I think this could simplify quite a bit because we wouldn't have as many conversions between different representations, there'd be fewer maps, fewer naming issues, etc.  I can work on a prototype if it'd help clarify what I mean.

sources/api/thar-be-settings/src/service.rs
sanu11: Yeah okay. Initially I had written it there, but I thought it would be easier to understand if separated in a different function

sources/api/thar-be-settings/src/service.rs
sanu11: Okay, Me and Matt had thought of doing this just to get more readability as we can call `get_all_services` instead of `get_affected_services` with `None`
But if its not required and adds to confusion I can make it as it was earlier.

sources/api/thar-be-settings/src/service.rs
sanu11: Oh nice! I could remove the & from the parameter. I didn't notice this wasn't used later on. : )

sources/api/thar-be-settings/src/main.rs
webern: I agree using a type here could help a lot. The type could also have a member function for returning a unique HashSet<String> of whatever it is we need that for.

tjkirch: Ok, I did [a prototype](https://gist.github.com/tjkirch/dbfc3be1b20ec4c6c52bf6d1e9d33139) of the `Services` thing I mentioned.  (prototype because it's untested, commits aren't split nicely, etc.)  @webern and @sanu11, what do you think, does this make the flow clearer?

sanu11: Thanks for putting it up @tjkirch  : ) Will go through it

webern: - `pub fn get_config_file_names(services: &Services)` could be a member function in `impl Services`.
- creating a member function for `services.0.is_empty()` improves readability slightly.

In general in both the original and @sanu11 's PR, I found the propagation of an `Option` to reflect `all services` very confusing. I suggested the change where @sanu11 introduced:
- `service::get_affected_services`
- `service::get_all_services`

I think hiding this implementation detail helps a bit instead of having mysterious options in `async fn run()`.

tjkirch: I think there are too many functions and it's too hard to trace.  If the Option is what's confusing, maybe we could make our own enum..? `get_services(Services::All)`, `get_services(Services::AffectedBySettings(list))`, would that help?

sanu11:
Addressed @tjkirch's comments except the prototype 

1. Passed ` &[&str]` to `is_restart_required` instead of `Vec`
2. Used `or_insert_with(Vec::new)` instead of ` .or_insert(vec![])` in `thar-be-settings/src/service.rs`
3. Removed `service_name.clone()` from  `thar-be-settings/src/service.rs`
4. Removed  function `get_affected_service_setting_map` and moved the code to the caller function, thus removed the test for the same 
5. The function `is_restart_required` wasn't called after changing `host-containers` setting, fixed it by removing the extra dot from `setting_prefix` in `host-containers`

PS: I have removed the testcase since the function is removed, would add it later based on the prototype changes

sources/api/host-containers/src/main.rs
```diff
@@ -437,8 +437,32 @@ where
     Ok(())
 }
 
+fn is_restart_required(settings: &[&str], container_name: &str) -> bool {
+    if settings.is_empty() {
+        // it means that Bottlerocket is booting - all containers need to be started
+        return true;
+    }
+    let setting_prefix = "settings.host-containers";
```
tjkirch: I don't understand the change to remove the final `.`.  thar-be-settings gets a list of concrete settings that changed, and `settings.host-containers` isn't a setting you can change on its own, it's only a container for other settings, so everything we receive should start with `settings.host-containers.` if it's related to host containers.  What am I missing?

sanu11: We are appending another dot below here for container_name` let container_prefix = format!("{}.{}.", setting_prefix, container_name);` so the container prefix was `settings.admin..enabled` earlier, because of which it was returning false. Do you want me to change container_prefix rather ?

tjkirch: Yeah, I think we should remove the inner `.` from `container_prefix` so we can check `setting_prefix` correctly.

sanu11:
Above push addresses
1. @tjkirch's prototype
2. Modify the commit message to describe newly added Services type  
3. Add the dot to the `settings_prefix` and removed it from `container_prefix` in `host-containers`
4. Rename function `is_restart_required` to `is_container_affected` in `host-containers` since the former was misleading as we do not restart containers on change yet. This is to be addressed as another issue/PR

I have tested the changes for the usecases mentioned in the PR description. I can update the description to contain the new logs once this  change is approved/reviewed

sources/api/host-containers/src/main.rs
```diff
@@ -437,8 +437,32 @@ where
...
+    if settings.is_empty() {
+        // it means that Bottlerocket is booting - all containers need to be started
+        return true;
+    }
+    let setting_prefix = "settings.host-containers.";
+    let container_prefix = format!("{}{}.", setting_prefix, container_name);
+
+    for setting in settings {
+        if setting.starts_with(&container_prefix) {
+            info!("Found a changed setting for {} container", container_name);
```
tjkirch: It might be useful to have similar messages above the other returns so that it's clear why (or why not) something was done with each container.

sources/api/thar-be-settings/src/service.rs
```diff
@@ -119,9 +168,11 @@ trait ServiceRestart {
     fn restart(&self) -> Result<()>;
 }
 
-impl ServiceRestart for model::Service {
+impl ServiceRestart for Service {
     fn restart(&self) -> Result<()> {
-        for restart_command in self.restart_commands.iter() {
+        let restart_commands = &self.model.restart_commands;
+        info!("restart commands {:?}", restart_commands);
```
tjkirch, line 155:
```suggestion
        info!("Found restart commands: {:?}", restart_commands);
```
Hmm, it's kind of unfortunate that we don't have the service name for use in messages like this.  We could change the HashMap in Services to a HashSet and store name as a field in Service instead... then we could easily use it in places like this without passing it around everywhere.

sources/api/thar-be-settings/src/service.rs
```diff
@@ -65,12 +113,12 @@ where
 /// Given a map of Setting to affected Service name, return a
 /// HashSet of affected service names
 fn get_affected_service_names(
```
tjkirch: I just realized this whole function can be replaced by a simple iterator function call :)

So above, we should be able to do this, and then remove the function entirely.
```diff
-let service_names = get_affected_service_names(&affected_services);
+let service_names = affected_services.values().flatten().collect();
```

tjkirch (commented):
Looking good to me, though I want to make sure @webern and @etungsten are happy with the ergonomics based on the open comments above.

sources/api/host-containers/src/main.rs
sanu11: I didn't write it for the settings that affects all host-containers since it would print that  `number of container` of times

tjkirch: That seems fine - it seems like it'd be useful to know why a container is being handled.  The user might not have changed any settings with that container in the name, for example.  They might not realize that network settings affect it.  Or that sometimes we have to handle all containers.  Or maybe they expected a container to be affected and it wasn't, so clarifying that could help.  When debugging, consistent output is good :)

sources/api/thar-be-settings/src/service.rs
tjkirch: @sanu11 raised a good point offline, which is that `Service` would have to implement `Hash` and `Eq` to be in a `HashSet`.  `model::Service` is stored in a field, and doesn't implement those traits, so it'd be a manual implementation for both... and at that point, I think it's too much code for too little benefit.

sources/api/host-containers/src/main.rs
```diff
@@ -437,8 +437,32 @@ where
...
+            return true;
+        }
+    }
+    return false;
+}
+
 async fn run() -> Result<()> {
     let args = parse_args(env::args());
+    // this env var is passed by thar-be-settings
+    let changed_settings_env = env::var("changed_settings").unwrap_or("".to_string());
```
webern: Should the env var be uppercase?

sources/api/thar-be-settings/src/service.rs
```diff
@@ -1,46 +1,94 @@
...
+impl Services {
+    /// Convert from the model's representation of all services to our own, adding in the lists of
+    /// changed settings, if appropriate; pass None if the program is running for *all* services.
+    pub fn from_model_services(
+        input: model::Services,
+        mut changed_settings: Option<HashMap<String, HashSet<String>>>,
+    ) -> Self {
+        let mut output = HashMap::new();
+        for (name, model) in input {
+            let changed_settings = changed_settings.as_mut().and_then(|s| s.remove(&name));
```
webern: Changing the name of this inner variable to anything other than `changed_settings` would help. I found the function confusing because I didn't quite get the nuance of what was happening with variable shadowing and scope.

```suggestion
            let current_changed_settings = changed_settings.as_mut().and_then(|s| s.remove(&name));
```

sanu11:
^ Above push addresses @tjkirch's comments
1.  Remove `get_affected_service_names` function and added iterator function call
2. Remove unit test since the function is no more present
3. Add log for services that affects all host-containers in `is_container_affected`

Made changes to the PR description by adding recent logs with changed info statements

sources/api/thar-be-settings/src/service.rs
tjkirch: The field in the service structure is `changed_settings`, and I think that makes sense since they're the changed settings relevant to that service, and if this variable name matches the field then the code is slightly simpler...  An alternate that I think achieves the same goal would be to rename the function parameter from `changed_settings` to `all_changed_settings` or something similar to indicate that it includes all changed settings we know about for all services, which we need to scope down here...

sources/api/host-containers/src/main.rs
tjkirch: I think it'd be helpful to include in all four return locations for consistency.  It'd also help to include the setting name that triggered the change in the log messages.

sources/api/host-containers/src/main.rs
sanu11: oh yeah, but this is something temporary, its fine to make it uppercase right ?

sources/api/thar-be-settings/src/service.rs
webern: OK, yeah that makes more sense.

sources/api/host-containers/src/main.rs
sanu11: okay, about the setting name, I was thinking if there are multiple settings for same container, only first one would be printed since we are returning after a match, thus I didn't add it

tjkirch: We could just include the first, if the messaging was clear, or maybe instead of returning true/false immediately we could build a list of settings that were relevant and include them in the log...  or leave it as-is :)

sanu11:
Above force push addresses
1. add log messages in all four return locations for consistency in the `is_container_affected` function
2. change env for `changed_settings` to uppercase 

Changed the PR description for tests according to the changed logs

sources/api/host-containers/src/main.rs
```diff
@@ -437,8 +437,36 @@ where
     Ok(())
 }
 
+fn is_container_affected(settings: &[&str], container_name: &str) -> bool {
+    if settings.is_empty() {
+        // it means that Bottlerocket is booting - all containers need to be started
+        info!("{} container will be handled since bottlerocket is booting", container_name);
```
tjkirch:
```suggestion
        info!("Handling host container '{}'  during full configuration process", container_name);
```

sources/api/host-containers/src/main.rs
```diff
@@ -437,8 +437,36 @@ where
...
+        return true;
+    }
+
+    let setting_prefix = "settings.host-containers.";
+    let container_prefix = format!("{}{}.", setting_prefix, container_name);
+
+    for setting in settings {
+        if setting.starts_with(&container_prefix) {
+            info!("Found one of the changed settings '{}' that affects '{}' container, thus handling this container", setting, container_name);
```
tjkirch:
```suggestion
            info!("Handling host container '{}' because it's directly affected by changed setting '{}' (and maybe others)", container_name, setting);
```

sources/api/host-containers/src/main.rs
```diff
@@ -437,8 +437,36 @@ where
...
+            return true;
+        }
+        if !setting.starts_with(&setting_prefix) {
+            // if its some other setting, return true for all host-containers, example: network
+            info!("Found one of the changed settings '{}' that affects '{}' container, thus handling this container", setting, container_name);
```
tjkirch:
```suggestion
            info!("Handling host container '{}' because it's indirectly affected by changed setting '{}' (and maybe others)", container_name, setting);
```

sources/api/host-containers/src/main.rs
```diff
@@ -437,8 +437,36 @@ where
...
+            return true;
+        }
+    }
+    info!("No changed settings affect {} container, thus no handling required", container_name);
```
tjkirch:
```suggestion
    info!("Not handling host container '{}', no changed settings affect it", container_name);
```

tjkirch (commented):
Looks like what's remaining is:
* The member functions suggested in https://github.com/bottlerocket-os/bottlerocket/pull/1480#discussion_r620479668, and possibly a change to `get_affected_services`, depending on how @webern is feeling in that thread now
* The confusing argument shadowing mentioned in https://github.com/bottlerocket-os/bottlerocket/pull/1480#discussion_r622561258  (sorry)

sources/api/host-containers/src/main.rs
sanu11: If we write it here, should we remove it from the `handle_host_container` function. ? 
it says - `  info!(
        "Handling host container '{}' which is enabled: {}",
        name, enabled
    );
` or is it fine to print twice

tjkirch: Maybe that one should be repurposed to show all the relevant fields:
```
info!("Host container '{}' is enabled: {}, superpowered: {}, with source: {}",
    name, enabled, superpowered, source);
```

sources/api/thar-be-settings/src/service.rs
sanu11: I'll rename the function parameter then

webern (commented):
My approval is with [this one request](https://github.com/bottlerocket-os/bottlerocket/pull/1480#discussion_r622561258) outstanding (the shadowed variable name). Thanks for all the work on this @sanu11!

webern (approved):
My approval is with [this one request](https://github.com/bottlerocket-os/bottlerocket/pull/1480#discussion_r622561258) outstanding (the shadowed variable name). Thanks for all the work on this @sanu11!

(with actual Approve radio button this time, I'm getting bad at using this correctly)

sanu11:
> My approval is with [this one request](https://github.com/bottlerocket-os/bottlerocket/pull/1480#discussion_r622561258) outstanding (the shadowed variable name). Thanks for all the work on this @sanu11!
> 
> (with actual Approve radio button this time, I'm getting bad at using this correctly)

sure, I will change it. Thanks! 😄

sanu11:
Above force push addresses
1. Rename function parameter as suggested by @webern 
2. Add log messages for all conditions in `is_container_affected` as suggested by @tjkirch 

Updated PR description to include recent logs for test cases

etungsten (approved):

tjkirch (approved):

arnaldo2792 (approved):

Issue #1453, opened 2021-04-05
`settings.kubernetes.cluster-dns-ip` should be optional


zmrow:
Outside of EKS, we cannot compute a value for this, and there's not a reasonable default.  We also need to account for it being optional for kubelet in standalone mode.

We will need to:
* Make it an `Option<>` in the model
* Add an `{{#if ... }}` statement to the `kubelet-config` files for that setting

tjkirch:
> Make in an `Option<>` in the model

Everything in the model is made `Option` by `#[model]`, so we should just need the conditional.

zmrow:
Fixed via #1482

PR #1482, opened 2021-04-13
kubelet: make cluster-dns-ip optional

gthao313:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket/issues/1453

**Description of changes:**
Outside of EKS, we cannot compute a value for this, and there's not a reasonable default. We also need to account for it being optional for kubelet in standalone mode.

**Testing done:**
Since Those changes in different version kube-config are an identical change, and it works the same way in all versions, so we just need build aws-k8s-1.19-x86_64 and aws-k8s-1.19-aarch64 images to do test.

With `cluster-dns-ip` data
`bottlerocket-aws-k8s-1.19-x86_64-v1.0.8-b20d9ca6`
`bottlerocket-aws-k8s-1.19-aarch64-v1.0.8-b20d9ca6`

Launching instances to validate if `clusterDNS` works as expectation. 

```
clusterDNS:
- 10.100.0.10
```

Without `cluster-dns-ip` data (remove `cluster-dns-ip` value in `50-aws-k8s.toml`)
`bottlerocket-aws-k8s-1.19-x86_64-v1.0.8-ad8c4812-dirty`
`bottlerocket-aws-k8s-1.19-aarch64-v1.0.8-ad8c4812-dirty`

Launching instances to validate if `clusterDNS` is not in `kubelet-config`. 

```
clusterDomain: cluster.local
kubeReserved:
```
(no `clusterDNS` value there)

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.15/kubelet-config
+++ b/packages/kubernetes-1.15/kubelet-config
@@ -27,8 +27,10 @@ authorization:
     cacheUnauthorizedTTL: 30s
 {{~/if}}
 clusterDomain: {{settings.kubernetes.cluster-domain}}
+{{~#if settings.kubernetes.cluster-dns-ip}}
 clusterDNS:
 - {{settings.kubernetes.cluster-dns-ip}}
+{{~/if}}
 {{~#if settings.kubernetes.eviction-hard}}
 evictionHard:
   {{~#each settings.kubernetes.eviction-hard}}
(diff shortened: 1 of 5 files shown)


bcressey:
Nits on commit message:
* "kubelet: make cluster-dns-ip optional" would be a better title
* it would be good to put the description in this PR into the actual commit message

zmrow (commented):
I'm happy once @bcressey 's suggestions re: the commit message are addressed. :)

Were you able to confirm that instances' config was correct with _and_ without the setting?  Could you add that to the testing section in the description?

zmrow (approved):
:microscope:

bcressey (approved):

Issue #1452, opened 2021-04-05
Add metrics to VMWare variants


zmrow:
Currently VMWare variants don't include metrics as the `metricdog` config file depends on `settings.aws.region`.  Since we will not have `aws.region` in VMWare, let's default to having "`global`" as the region for now.

jpculp:
Merged #1486.

PR #1486, opened 2021-04-13
Enable metrics for vmware-dev variant

jpculp:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket/issues/1452

**Description of changes:**

This adds a conditional for the region. If an aws region is not available, the region is set to global. This is required to enable metrics on non-aws variants like `vmware-dev`.

This also enables metrics on the `vmware-dev` variant by adding the `metrics.toml` to the variant's `defaults.d` directory and adding`MetricsSettings` to the `mod.rs`.

**Testing done:**

- [x] Built vmware-dev variant.
- [x] Built OVA.
- [x] Imported into vSphere.
- [x] Set motd in user data.
- [x] Started VM.
- [x] Logged into console and ran `metricdog --log-level debug send-health-ping`.
- [x] Checked for health ping record.
- [x] Verified the data inside the health ping record was correct _(ie. region = global)_.

**Terms of contribution:**

### DIFF
--- a/packages/os/metricdog-toml
+++ b/packages/os/metricdog-toml
@@ -1,7 +1,11 @@
 metrics_url = "{{settings.metrics.metrics-url}}"
 send_metrics = {{settings.metrics.send-metrics}}
 service_checks = [{{join_array ", " settings.metrics.service-checks}}]
-region = "{{settings.aws.region}}"
 seed = {{settings.updates.seed}}
 version_lock = "{{settings.updates.version-lock}}"
 ignore_waves = {{settings.updates.ignore-waves}}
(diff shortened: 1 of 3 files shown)


zmrow (commented):
I believe you'll also need to add MetricsSettings to the `vmware-dev` model

jpculp:
- Added `MetricsSettings` to variant's `mod.rs`.
- Tested and updated PR notes.

zmrow (approved):
:bear:

webern (approved):

zmrow (approved):

PR #1487, opened 2021-04-14
models: remove aws-k8s-1.15 files/references, redo symlinks, add aws-k8s-1.19 defaults

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
Date:   Tue Apr 13 17:46:52 2021 -0700

    models: redo k8s variant symlinks, add aws-k8s-1.19 model/defaults
    
    Create default and model files for the aws-k8s-1.19 variant.
    Relink the symlink for all k8s variants so they link to aws-k8s-1.19's
    files.


```

```
Date:   Wed Apr 14 14:09:58 2021 -0700

    variants, models: remove aws-k8s-1.15 files and references
    
    This deletes aws-k8s-1.15 models and defaults and the variant directory.
    The aws-k8s-1.15 can no longer be built.

```

**Testing done:**
Built aws-k8s-1.16, aws-k8s-1.17, aws-k8s-1.18, aws-k8s-1.19 image successfully. Launched instances with said images and they all come up successfully and show up ready in my cluster.

**Terms of contribution:**


jpculp:
Rather than having everything symlink to 1.16, would it make more sense to make the symlinks point to 1.19? That way we wouldn't have to make this change _each_ time a version is deprecated.

jpculp (approved):

etungsten:
Push above removes files and directories related to aws-k8s-1.15. Truly deprecates the aws-k8s-1.15 variant.

etungsten:
> Rather than having everything symlink to 1.16, would it make more sense to make the symlinks point to 1.19? That way we wouldn't have to make this change _each_ time a version is deprecated.

I think I'd prefer keeping the symlinks as they are now since we're likely to change something about newer k8s variants that older k8s variants might not support (necessitating a different set of defaults, setting models, etc) before we deprecate another k8s variant. So I think it makes sense to symlink back to older k8s versions as they are more "stable" in a sense.

tjkirch:
> > Rather than having everything symlink to 1.16, would it make more sense to make the symlinks point to 1.19? That way we wouldn't have to make this change _each_ time a version is deprecated.
> 
> I think I'd prefer keeping the symlinks as they are now since we're likely to change something about newer k8s variants that older k8s variants might not support (necessitating a different set of defaults, setting models, etc) before we deprecate another k8s variant. So I think it makes sense to symlink back to older k8s versions as they are more "stable" in a sense.

I don't think we'd do that with an existing variant, though, only a new one like 1.20, since we don't want to change behavior in any breaking way for existing users.

etungsten:
Adopts https://github.com/bottlerocket-os/bottlerocket/pull/1487#issuecomment-819681491's suggestion.
Push above reverses the symlink's direction so aws-k8s-1.19's directory is the source of all the symlinks.

variants/README.md
```diff
@@ -73,7 +73,7 @@ User data will be read from a mounted CD-ROM, either from a file named "user-dat
 
 #### aws-k8s-1.15: Kubernetes 1.15 node
 
-The [aws-k8s-1.15](aws-k8s-1.15/Cargo.toml) variant includes the packages needed to run a Kubernetes node in AWS.
+The aws-k8s-1.15 variant includes the packages needed to run a Kubernetes node in AWS.
 It supports self-hosted clusters and clusters managed by [EKS](https://aws.amazon.com/eks/).
 
 This variant is compatible with Kubernetes 1.15, 1.16, and 1.17 clusters.
```
bcressey: Why not just delete this section?

etungsten: This is under the `### Deprecated variants` section of the README. I thought it's meant to keep historical records of deprecated variants.

tjkirch: At least for some period of time, I thought it'd be good to keep the listing here, in the deprecated section.  Users running 1.15 who suddenly can't find any mention of their product anywhere could be understandably peeved.  :)

jpculp (approved):

webern (approved):

variants/README.md
bcressey: Should we change it to past tense, since it's gone from the tree?

sources/models/README.md
```diff
@@ -22,30 +22,25 @@ Entries are sorted by filename, and later entries take precedence.
...
-
-* [Model](src/aws-k8s-1.15/mod.rs)
-* [Default settings](src/aws-k8s-1.15/defaults.d/)
-
 ### aws-k8s-1.16: Kubernetes 1.16
 
-* [Model](src/aws-k8s-1.15/mod.rs)
-* [Default settings](src/aws-k8s-1.15/defaults.d/)
+* [Model](src/aws-k8s-1.16/mod.rs)
+* [Default settings](src/aws-k8s-1.16/defaults.d/)
```
bcressey: It looks like this README needs to be regenerated.

variants/README.md
bcressey:
```suggestion
The aws-k8s-1.15 variant included the packages needed to run a Kubernetes node in AWS.
It supported self-hosted clusters and clusters managed by [EKS](https://aws.amazon.com/eks/).

This variant was compatible with Kubernetes 1.15, 1.16, and 1.17 clusters.
It reached end-of-life on May 3, 2021.
```

etungsten:

bcressey (approved):

Issue #1459, opened 2021-04-06
All variants: update `cloud-provider` argument for `kubelet`


zmrow:
### Problem
For our existing AWS Kubernetes variants we hard-code `--cloud-provider "aws"` in the `kubelet` args.

Clusters provisioned by ClusterAPI Provider for vSphere must use `--cloud-provider "external"`.  At the time of writing, clusters provisioned by Cluster API Provider for AWS don't default to `--cloud-provider "external"` yet, [but might later](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/964).

Upstream Kubernetes [is in the process](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cloud-provider/2395-removing-in-tree-cloud-providers/README.md) of removing all in-tree cloud provider code.  This is slated to be GA in 1.25, at which point `--cloud-provider "external"` will be mandatory.

### Recommended solution
We should add a new setting: `settings.kubernetes.cloud-provider` that will only accept the values "aws" and "external".  When `--cloud-provider "external"` becomes mandatory in the future, and all providers are removed from in-tree upstream, we can remove the setting for the Kubernetes variants that are on at least that version.  Unfortunately, this means we'll need to carry the setting until older Kubernetes versions "age out".

### Alternatives
* We could conditionalize the `--cloud-provider` argument in the `kubelet-exec-start-conf` template based on some other factor, for example `settings.aws.region`.  This would be unreliable in DIY Kubernetes installs, or when using Cluster API Provider for AWS.

tjkirch:
Something like `{{#if settings.aws}}aws{{else}}external{{/if}}`, perhaps, to remove the need for a setting?

>  This would be unreliable in DIY Kubernetes installs, or when using Cluster API Provider for AWS.

Can you expand on this a bit?  Is the issue that it would be detected as AWS because of IMDS, but we _wouldn't_ want to use `--cloud-provider aws`?  (Why?)

bcressey:
Some people will want to use the [external AWS cloud provider](https://github.com/kubernetes/cloud-provider-aws) - I'd expect more interest here as the date for deleting the in-tree provider code looms closer.

PR #1494, opened 2021-04-16
kubelet: add setting for configuring cloudProvider

jpculp:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket/issues/1459

**Description of changes:**

Adds a new setting `kubernetes.cloud-provider` for configuring whether the cloud provider is `aws` or `external`.
Prior to this, the argument was hard-coded to `aws`.

Adds a migration for the new `kubernetes.cloud-provider` setting.

**Testing done:**
- [x] Build `aws-k8s-1.19` ami.
- [x] Launch and connect to EKS cluster.
- [x] Deploy something to the node.
- [x] Connect to admin container and `sudo sheltie`.
- [x] Verify that there are no kubelet errors.
- [x] Repeat the same steps for `aws-k8s-1.18`.
- [x] Repeat the same steps for `aws-k8s-1.17`.
- [x] Repeat the same steps for `aws-k8s-1.16`.
- [x] Verify the migration works as expected.
- [x] Verify variant without a default cloud provider specified defaults to external.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,6 +308,7 @@ The following settings can be optionally set to customize the node labels and ta
 The following settings are optional and allow you to further configure your cluster.
 * `settings.kubernetes.cluster-domain`: The DNS domain for this cluster, allowing all Kubernetes-run containers to search this domain before the host's search domains.  Defaults to `cluster.local`.
 * `settings.kubernetes.standalone-mode`: Whether to run the kubelet in standalone mode, without connecting to an API server.  Defaults to `false`.
+* `settings.kubernetes.cloud-provider`: The cloud provider for this cluster. Defaults to `aws` for AWS variants, and `external` for other variants.
 * `settings.kubernetes.authentication-mode`: Which authentication method the kubelet should use to connect to the API server, and for incoming requests.  Defaults to `aws` for AWS variants, and `tls` for other variants.
 * `settings.kubernetes.server-tls-bootstrap`: Enables or disables server certificate bootstrap.  When enabled, the kubelet will request a certificate from the certificates.k8s.io API.  This requires an approver to approve the certificate signing requests (CSR).  Defaults to `true`.
 * `settings.kubernetes.bootstrap-token`: The token to use for [TLS bootstrapping](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/).  This is only used with the `tls` authentication mode, and is otherwise ignored.
--- a/packages/kubernetes-1.16/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.16/kubelet-exec-start-conf
@@ -2,7 +2,7 @@
 ExecStart=
 ExecStart=/usr/bin/kubelet \
 {{~#unless settings.kubernetes.standalone-mode}}
-    --cloud-provider aws \
+    --cloud-provider {{default "external" settings.kubernetes.cloud-provider}} \
     --kubeconfig /etc/kubernetes/kubelet/kubeconfig \
 {{~#if (eq settings.kubernetes.authentication-mode "tls")}}
     --bootstrap-kubeconfig /etc/kubernetes/kubelet/bootstrap-kubeconfig \
--- a/packages/kubernetes-1.18/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.18/kubelet-exec-start-conf
@@ -2,7 +2,7 @@
 ExecStart=
 ExecStart=/usr/bin/kubelet \
 {{~#unless settings.kubernetes.standalone-mode}}
-    --cloud-provider aws \
+    --cloud-provider {{default "external" settings.kubernetes.cloud-provider}} \
     --kubeconfig /etc/kubernetes/kubelet/kubeconfig \
 {{~#if (eq settings.kubernetes.authentication-mode "tls")}}
     --bootstrap-kubeconfig /etc/kubernetes/kubelet/bootstrap-kubeconfig \
--- a/sources/models/src/modeled_types/kubernetes.rs
+++ b/sources/models/src/modeled_types/kubernetes.rs
@@ -723,3 +723,48 @@ mod test_kubernetes_quantity_value {
         }
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesCloudProvider represents a string that is a valid cloud provider for the
+/// kubelet.  It stores the original string and makes it accessible through standard traits.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesCloudProvider {
+    inner: String,
+}
+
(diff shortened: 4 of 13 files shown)


packages/kubernetes-1.16/kubelet-exec-start-conf
```diff
@@ -2,14 +2,12 @@
...
 {{~#unless settings.kubernetes.standalone-mode}}
-    --cloud-provider aws \
     --kubeconfig /etc/kubernetes/kubelet/kubeconfig \
 {{~#if (eq settings.kubernetes.authentication-mode "tls")}}
     --bootstrap-kubeconfig /etc/kubernetes/kubelet/bootstrap-kubeconfig \
 {{~/if}}
-{{~else}}
-    --cloud-provider "" \
 {{~/unless}}
+    --cloud-provider "${CLOUD_PROVIDER}" \
```
webern: Is there a reason to use an environment variable instead of templating the setting value itself?

packages/kubernetes-1.17/kubelet-env
```diff
@@ -2,3 +2,4 @@ NODE_IP={{settings.kubernetes.node-ip}}
 NODE_LABELS={{join_map "=" "," "no-fail-if-missing" settings.kubernetes.node-labels}}
 NODE_TAINTS={{join_map "=" "," "no-fail-if-missing" settings.kubernetes.node-taints}}
 POD_INFRA_CONTAINER_IMAGE={{settings.kubernetes.pod-infra-container-image}}
+CLOUD_PROVIDER={{settings.kubernetes.cloud-provider}}
```
webern: Same question as above. Curious whether this needs to be an environment variable whether its value should be written to the conf.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -723,3 +723,48 @@ mod test_kubernetes_quantity_value {
...
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesCloudProvider represents a string that is a valid cloud provider for the
+/// kubelet.  It stores the original string and makes it accessible through standard traits.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesCloudProvider {
+    inner: String,
```
webern, line 733: Maybe an enum would be better. Not a strong preference but it looks like we have enums for other settings.

packages/kubernetes-1.17/kubelet-env
bcressey: Everything but `POD_INFRA_CONTAINER_IMAGE` can be in the conf template now. The environment variable approach was all we had before ef41a6b1, but isn't as flexible.

We don't have to move the other settings here, but shouldn't add more.

README.md
```diff
@@ -308,6 +308,7 @@ The following settings can be optionally set to customize the node labels and ta
 The following settings are optional and allow you to further configure your cluster.
 * `settings.kubernetes.cluster-domain`: The DNS domain for this cluster, allowing all Kubernetes-run containers to search this domain before the host's search domains.  Defaults to `cluster.local`.
 * `settings.kubernetes.standalone-mode`: Whether to run the kubelet in standalone mode, without connecting to an API server.  Defaults to `false`.
+* `settings.kubernetes.cloud-provider`: The cloud platform for this cluster. Defaults to `aws` for AWS variants, and `external` for other variants.
```
zmrow: We should probably say "cloud provider" rather than "cloud platform".

sources/models/src/modeled_types/kubernetes.rs
zmrow: I think I'd reach for an enum as well. :)

packages/kubernetes-1.17/kubelet-env
jpculp: Updated.

packages/kubernetes-1.16/kubelet-exec-start-conf
jpculp: I was following some existing code, but since it's not necessary I've removed the environment variable.

sources/models/src/modeled_types/kubernetes.rs
jpculp: I'm not crazy about the enum for something this small. This block of code used [KubernetesAuthenticationMode](https://github.com/bottlerocket-os/bottlerocket/blob/64e679747e3af8a029df7e66ae5dfc8b6c220e0c/sources/models/src/modeled_types/kubernetes.rs#L356) as a reference.

README.md
jpculp: I was trying to avoid repeating the name of the setting, but if folks feel strongly or have a less redundant suggestion I'm open to changing it.

jpculp:
- Removed the environment variable logic.

jpculp:
- Tested all other support kubernetes versions.
- Performed migration test with 1.19 instance.

README.md
etungsten: I think we should keep it in line with the term that the k8s official doc uses to remove any ambiguity on what this is for. See https://v1-18.docs.kubernetes.io/docs/concepts/cluster-administration/cloud-providers/

packages/kubernetes-1.18/kubelet-exec-start-conf
```diff
@@ -2,14 +2,12 @@
...
 {{~#unless settings.kubernetes.standalone-mode}}
-    --cloud-provider aws \
     --kubeconfig /etc/kubernetes/kubelet/kubeconfig \
 {{~#if (eq settings.kubernetes.authentication-mode "tls")}}
     --bootstrap-kubeconfig /etc/kubernetes/kubelet/bootstrap-kubeconfig \
 {{~/if}}
-{{~else}}
-    --cloud-provider "" \
 {{~/unless}}
+    --cloud-provider {{settings.kubernetes.cloud-provider}} \
```
bcressey: We need the old logic here so that we pass an empty string to cloud-provider if standalone mode is enabled.

jpculp:
- Reverted back to unless+else logic in `kubelet-exec-start-conf`.
- Added default setting of `external` if `settings.kubernetes.cloud-provider` is absent from the variant's defaults.
- Replaced the word `platform` with `provider`.

sources/models/src/modeled_types/kubernetes.rs
jpculp: I'm open to replacing it with an enum later, but we would want to address https://github.com/bottlerocket-os/bottlerocket/issues/1501 first.

bcressey (approved):

zmrow (approved):
:safety_vest: 

Code looks good.  Before you merge - let's make sure to build an image without the `cloud-provider` setting and make sure that it defaults to "external" like we think it should.

jpculp:
Tested migration with a variant that did not have a default cloud provider setting set. Launched as `--cloud-provider aws` with 1.0.8 and after upgrading to a modified 1.1.0 variant launched as `--cloud-provider external`.

PR #1496, opened 2021-04-16
restrict bootstrap container user data

bcressey:
**Issue number:**
N/A

**Description of changes:**
This is a follow-up ea35f1bb, which added similar restrictions to host container user data.

**Testing done:**
Confirmed that a bootstrap container's user data had the expected label:
```
# ls -latrZ /local/bootstrap-containers/boot-label/user-data
-rw-r--r--. 1 root root system_u:object_r:secret_t:s0 11 Apr 16 18:36 /local/bootstrap-containers/boot-label/user-data
```

**Terms of contribution:**

### DIFF
--- a/packages/os/bootstrap-containers-tmpfiles.conf
+++ b/packages/os/bootstrap-containers-tmpfiles.conf
@@ -1,3 +1,3 @@
 d /etc/bootstrap-containers 0750 root root -
 d /local/bootstrap-containers 0700 root root -
-T /local/bootstrap-containers - - - - security.selinux=system_u:object_r:state_t:s0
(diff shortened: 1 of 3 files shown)


packages/os/bootstrap-containers-tmpfiles.conf
```diff
@@ -1,3 +1,3 @@
 d /etc/bootstrap-containers 0750 root root -
```
tjkirch, line 1: I think we also need to add filecon entries in fs.cil and typetransition in rules.cil, like host-containers.

tjkirch (approved):

samuelkarp (approved):

Issue #1458, opened 2021-04-06
VMWare: figure out where to get `settings.kubernetes.node-ip`


zmrow:
Related to #1451 

For AWS variants we depend on IMDS to retrieve the node's static IP address.  We'll need to figure out the best way of determining this IP in VMWare.

This will require a little bit of investigation, but some options:
* Potentially use the VMWare backdoor interface?
  * If this is possible, we'll write a small settings generator (`charon`) to retrieve this information at runtime
* Query DHCP lease information 
  * This could be done via a new subcommand of netdog (`netdog node-ip`) and would need to happen at boot time before kubelet.  `netdog node-ip` would be our settings generator in this case.
  * This assumes we have a single interface that has the proper DHCP-assigned IP address to use (could be problematic later?)

zmrow:
The static IP address isn't provided via the VMWare backdoor.  We will probably move forward with querying the DHCP lease info via a new subcommand of `netdog`.  This subcommand will be used as the settings generator.

zmrow:
Implemented in #1502

Issue #1451, opened 2021-04-05
Add a `vmware-k8s-1.x` variant


zmrow:
Create a "real" or non-dev `vmware` Kubernetes variant.  This will require at the least:

* a new model for this variant
* a new variant definition
* containerd config
* adding metrics to the variant (metricdog settings)
* logdog config

PR #1502, opened 2021-04-19
netdog: Persist current IP and add subcommand `node-ip` for retrieval

zmrow:
**Issue number:**
Related to #1458 

**Description of changes:**
```
With this change, netdog will persist the node's current IP (read from
it's DHCP leases file) to /var/lib/netdog/current_ip.  It does this as
part of its "install" subcommand, which is run each time that wickedd
receives an update.  A new subcommand has also been added, which
retrieves the current IP from said file and returns it as JSON.  This
subcommand is intended for use as a settings generator.
```

This change is meant to support the VMWare variant #1451, specifically the `kubernetes.node-ip` setting and will receive further exercise with that PR.

**Testing done:**
* Build and run an `aws-dev` variant to ensure that the file containing the IP exists with the correct data and the subcommand works:
```
bash-5.0# ls -lahZ /var/lib/netdog/current_ip 
-rw-r--r--. 1 root root system_u:object_r:lease_t:s0 11 Apr 19 20:15 /var/lib/netdog/current_ip

bash-5.0# cat /var/lib/netdog/current_ip 
10.0.60.220

bash-5.0# netdog node-ip
"10.0.60.220"
```
* Build and run an `aws-k8s-1.19` node, started sheltie and checked files.  Node joins the cluster just fine and I can run pods on it.

**Terms of contribution:**

### DIFF
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -2088,6 +2088,7 @@ dependencies = [
  "rand 0.8.3",
  "regex",
  "serde",
+ "serde_json",
  "serde_plain",
  "snafu",
 ]
--- a/sources/api/netdog/src/main.rs
+++ b/sources/api/netdog/src/main.rs
@@ -1,9 +1,11 @@
 /*!
 # Introduction
 
-netdog is a small helper program for wicked, to apply network settings received from DHCP.
+netdog is a small helper program for wicked, to apply network settings received from DHCP.  It also
+contains a subcommand `node-ip` that returns the node's current IP address in JSON format; this
+subcommand is intended for use as a settings generator.
 
-It generates `/etc/resolv.conf` and sets the hostname.
+It generates `/etc/resolv.conf`, sets the hostname, and persists the current IP to file.
 */
 
 // TODO:
@@ -25,7 +27,7 @@ use regex::Regex;
 use serde::Deserialize;
 use snafu::ResultExt;
 use std::collections::BTreeSet;
-use std::fmt::Write;
+use std::fmt::{self, Write};
 use std::fs::{self, File};
 use std::io::{BufRead, BufReader};
 use std::net::IpAddr;
@@ -34,6 +36,7 @@ use std::{env, process};
 
 static RESOLV_CONF: &str = "/etc/resolv.conf";
 static KERNEL_HOSTNAME: &str = "/proc/sys/kernel/hostname";
+static CURRENT_IP: &str = "/var/lib/netdog/current_ip";
 
 // Matches wicked's shell-like syntax for DHCP lease variables:
 //     FOO='BAR' -> key=FOO, val=BAR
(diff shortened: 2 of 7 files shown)


sources/Cargo.lock
```diff
@@ -2088,6 +2085,7 @@ dependencies = [
  "rand 0.8.3",
  "regex",
  "serde",
+ "serde_json",
```
bcressey, line 2091: I expect it'd work with just this change to `Cargo.lock` and not all the rest of it. Can you try that?

sources/api/netdog/src/main.rs
```diff
@@ -299,11 +327,27 @@ fn remove(args: &Args) -> Result<()> {
...
+    match parse_args(env::args())? {
+        (SubCommand::NodeIp, None) => node_ip()?,
+        (subcommand, Some(args)) => match subcommand {
+            SubCommand::Install => install(&args)?,
+            SubCommand::Remove => remove(&args)?,
+            SubCommand::NodeIp => {
+                usage_msg(format!("Subcommand 'node-ip' doesn't support arguments"))
+            }
+        },
+        (_, None) => usage_msg(format!("Subcommands 'install/remove' require arguments")),
```
bcressey: nit: this might be clearer overall, since it doesn't mix `NodeIp` in with the handling of `Install` and `Remove`:

```suggestion
    match parse_args(env::args())? {
        (SubCommand::NodeIp, None) => node_ip()?,
        (SubCommand::NodeIp, Some(_) => usage_msg(format!("Subcommand 'node-ip' doesn't support arguments")).
        (subcommand, Some(args)) => match subcommand {
            SubCommand::Install => install(&args)?,
            SubCommand::Remove => remove(&args)?,
        },
        (_, None) => usage_msg(format!("Subcommands 'install/remove' require arguments")),
```

bcressey (commented):
Can you also verify that `/etc/resolv.conf` and `/proc/sys/kernel/hostname` are still set correctly?

sources/api/netdog/src/main.rs
bcressey:
```
(_, None) => usage_msg(format!("Subcommands 'install/remove' require arguments"))
```

For this it might be better to use the `subcommand` parameter so you can add it to the format string.

sources/Cargo.lock
webern: Was this a cargo update?

webern (commented):
Is persisting the IP to a file an optimization (i.e. for faster queries)? Or is the persistence needed across reboots?

zmrow:
@webern Persisting the IP to file means that we can use it as the static IP for the setting `kubernetes-node-ip`, specifically for VMWare (since we don't have the luxury of IMDS).

sources/Cargo.lock
zmrow: It wasn't meant to be.  I'm going to implement @bcressey 's ask above to avoid it.

zmrow:
^ Addressed @bcressey 's comments and verified that `/etc/resolv.conf` and `/proc/sys/kernel/hostname` are both still set correctly.

sources/api/netdog/src/main.rs
webern: I suggest:

```rust
enum Subcommand {
    Install(Args),
    Remove(Args),
    NodeIp,
}
```

And

```rust
fn parse_args(args: env::Args) -> Result<SubCommand> {
   // ...
}
```

webern: Actually since, since this would require us to stop using serde_plain to 'deserialize' the subcommand string, it's probably not worth it.

webern (approved):

sources/api/netdog/src/main.rs
zmrow: @bcressey - The compiler wasn't smart enough to do your suggestion, it still wanted an exhaustive match of the enum variants inside the second `match`.  I opted to split out the options a bit further and just avoid a nested match altogether.

bcressey (approved):

PR #1503, opened 2021-04-19
makefile: check cargo version

webern:
**Issue number:**

Follow up to #1408
Follow up to #1361 

**Description of changes:**

    makefile: check cargo version

    For the build-package target (used by cargo make repo), we need Cargo
    version 1.51 or higher to build a dependency of the variants workspace.
    To make this clear, we check the Cargo version before build-package and
    provide a clear error message.

**Testing done:**

```sh
rustup default 1.50.0
cargo make repo
# Error: Cargo 1.51.0 or greater is required, your version is 1.50.0
# [cargo-make] ERROR - Error while executing command, exit code: 1
# [cargo-make] WARN - Build Failed.
# FAIL: 1
```

```sh
rustup default 1.51.0
cargo make repo
# Works!
```

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -266,9 +266,31 @@ cargo install \
 '''
 ]
 
+# We need Cargo version 1.51 or higher in order to build a workspace's
+# dependency during build-package
+[tasks.check-cargo-version]
+script_runner = "bash"
+script = [
+'''
+set -euo pipefail
+cargo_version=$(cargo --version | awk '{print $2}')
+strarr=(${cargo_version//./ })
+cargo_major="${strarr[0]}"
+cargo_minor="${strarr[1]}"
+if [ "${cargo_major}" -gt "1" ] ; then
+  # cargo is version 2 or higher, so it's higher than 1.51
+  exit 0
+fi
+if [ "${cargo_minor}" -lt "51" ] ; then
+  echo "Error: Cargo 1.51.0 or greater is required, your version is ${cargo_version}" >&2
+  exit 1
+fi
+'''
+]
+
 # Builds a package including its build-time and runtime dependency packages.
 [tasks.build-package]
-dependencies = ["build-tools", "publish-setup"]
+dependencies = ["check-cargo-version", "build-tools", "publish-setup"]
 script_runner = "bash"
 script = [
 '''


Makefile.toml
```diff
@@ -266,9 +266,31 @@ cargo install \
...
 ]
 
+# We need Cargo version 1.51 or higher in order to build a workspace's transient dependency
+# during build-package
+[tasks.check-cargo-version]
+script_runner = "bash"
+script = [
+'''
+set -e
+cargo_version=$(cargo -V | awk -F '[() ]' '{print $2}')
```
```suggestion
cargo_version=$(cargo --version | awk '{print $2}')
```
Don't need the regex field separator since this version doesn't use the part in the parens.  (And `--version` to be a little clearer about what we're asking from cargo...)

Or if you want to simplify the next bits too:
```
IFS=. read -r major minor _ <<< "$(cargo --version | awk '{print $2}')"
```

Makefile.toml
```diff
@@ -266,9 +266,31 @@ cargo install \
 '''
 ]
 
+# We need Cargo version 1.51 or higher in order to build a workspace's transient dependency
+# during build-package
+[tasks.check-cargo-version]
+script_runner = "bash"
+script = [
+'''
+set -e
```
```suggestion
set -euo pipefail
```
Since we're relying on data from a non-final pipeline component.

tjkirch (approved):

Makefile.toml
webern: I avoided `read -r` to stay compatible with old macbook bash. I wouldn't strive for that compatibility unless it was somewhat trivial to accomplish... but currently I think everything in Makefile works on stock/old bash that comes with 10.X.

I made the other changes.

tjkirch (approved):

Makefile.toml
```diff
@@ -266,9 +266,31 @@ cargo install \
...
+[tasks.check-cargo-version]
+script_runner = "bash"
+script = [
+'''
+set -euo pipefail
+cargo_version=$(cargo --version | awk '{print $2}')
+strarr=(${cargo_version//./ })
+cargo_major="${strarr[0]}"
+cargo_minor="${strarr[1]}"
+if [ "${cargo_major}" -gt "1" ] ; then
```
jpculp, line 280: nit. It would be nice if we could standardize on `[[...]]`.
nit. I don't think you need quotes for things without spaces (ie. numbers)

jpculp (approved):
Approved if the checks come back green.

Makefile.toml
tjkirch: This is typically done to maintain some level of compatibility with other shells, and then only use `[[ ]]` where you know you need bash specifics.

etungsten (approved):

bcressey (approved):

Makefile.toml
jpculp: If we are worried about compatibility with other shells, then shouldn't we avoid _all_ instances of bash-isms?

webern:
> webern force-pushed the webern:check-cargo branch from a8a10b1 to afd9c4c

I pushed @jpculp's suggestion, but I based on the discussion I think I'm going to undo it and merge.

webern:
> webern force-pushed the webern:check-cargo branch from afd9c4c to a8a10b1

Revert back to where we were when we had a bunch of green check marks :)

Makefile.toml
tjkirch: No, we do need some of them.  It's a balance - sometimes bash-isms are the best way to do something, but otherwise staying compatible with POSIX sh and BSD userland is a good habit.  We've seen both of those things bite us in Makefile.toml, for example.  Even if your shebang is bash, it's nice to use sh basics when they're essentially identical, like in this case, in case you need the script to run elsewhere someday.  (I'm old, and so have run into this many times.)

Issue #1481, opened 2021-04-12
logdog should collect all available log files for ECS CNI plugins


samuelkarp:
**What I'd like:**
As part of enabling the ECS `awsvpc` networking mode for Bottlerocket (https://github.com/bottlerocket-os/bottlerocket/pull/1246), several new ECS-related components (CNI plugins) are added to Bottlerocket.  As independent components, they do not share the log file that is used by the ECS agent itself.  These components do their own log file rotation by writing files with hourly timestamp suffixes (and one of them even uses the hourly timestamp suffixed file name for the current log file), with file names like `vpc-branch-eni.log.2021-04-09-00`.  Today, there is no mechanism to specify a non-static log file name with logdog, so logdog cannot collect any log files with hourly timestamp suffixes.

**Any alternatives you've considered:**
Here are some potential approaches

1. Enable logdog to specify file paths with *wildcards*.  Wildcards would enable us to specify a pattern of files to collect and is very flexible.  However, the flexibility makes wildcards somewhat complicated (we'd need to define the exact rules we want to use for expansion) and we may not really need that level of flexibility.
3. Enable file *prefixes* with logdog.  Rather than an arbitrary pattern, prefixes are less complicated and easier to reason about.  Prefixes adequately cover all of the ECS-related use-cases; none of the log files ever change the beginning of their names.
4. Collect whole *directories* rather than just files.  We could add a new directory directive to logdog’s configuration file.  When a directory is specified, logdog includes the contents of the whole directory.  For ECS, all of the log files are contained in /var/log/ecs, even rotated ones.

tjkirch:
>  However, the flexibility makes wildcards somewhat complicated (we'd need to define the exact rules we want to use for expansion)

I think we can use the glob crate and get a syntax that's widely understood, since it mimics shell wildcards.  It's strictly more powerful, and easier to implement, too.

srgothi92:
Fixed in [PR-1509](https://github.com/bottlerocket-os/bottlerocket/pull/1509)

PR #1509, opened 2021-04-21
Supported file pattern for logdog log files

srgothi92:
**Issue number:**
#1481 

**Description of changes:**
 Ecs variant awsvpc mode feature's log file has dynamic names, so it was difficult to export such files with current `logdog` capabilities. This change extends logdog to support a new request type `glob`, which takes pattern for file matching. For this case, destination filename and path in tarball will be same as source file.

**Testing done:**
1. Added unit tests to test different combination of patterns.
2. Tested `logdog` in an image:
     * Build an ecs variant image
     * Launced ami in ECS cluster
     * Started eni trunking task to generate `ecs-cni-eni-plugin.log`, `ecs-cni-bridge-plugin.log` and `vpc-branch-eni.log` log files. 
     * Waited an hour and re-ran task so that timestamp are attached to log files.
     * Followed steps [here](https://github.com/bottlerocket-os/bottlerocket#logs) to export log files.
     * verified all the log files are present in expected directory. 

**Terms of contribution:**

### DIFF
--- a/sources/logdog/src/error.rs
+++ b/sources/logdog/src/error.rs
@@ -124,6 +124,15 @@ pub(crate) enum Error {
     #[snafu(display("Empty command."))]
     ModeMissing {},
 
+    #[snafu(display("Error parsing glob pattern '{}': {}", pattern, source))]
+    ParseGlobPattern {
+        pattern: String,
+        source: glob::PatternError,
+    },
+
+    #[snafu(display("The logdog configuration has a 'glob' line with no glob instructions."))]
+    PatternMissing {},
+
     #[snafu(display("Cannot write to / as a file."))]
     RootAsFile { backtrace: Backtrace },
 
--- a/sources/logdog/src/log_request.rs
+++ b/sources/logdog/src/log_request.rs
@@ -9,13 +9,16 @@
 //! these provide the list of log requests that `logdog` will run.
 
 use crate::error::{self, Result};
+use glob::glob;
 use reqwest::blocking::{Client, Response};
 use snafu::{ensure, OptionExt, ResultExt};
+use std::collections::HashSet;
 use std::fs;
 use std::fs::File;
 use std::path::Path;
 use std::process::{Command, Stdio};
 use url::Url;
+use walkdir::WalkDir;
 
 /// The `logdog` log requests that all variants have in common.
 const COMMON_REQUESTS: &str = include_str!("../conf/logdog.common.conf");
@@ -33,11 +36,12 @@ pub(crate) fn log_requests() -> Vec<&'static str> {
         .collect()
 }
 
-/// A logdog `LogRequest` is in the format `mode filename instructions`. `mode` specifies what type
-/// of command it is, e.g. `exec ` for a command or `http` for an HTTP get request. `filename` is
-/// the name of the output file. `instructions` is any additional information needed.  For example,
-/// an `exec` request'ss instructions will include the program and program arguments. An `http`
-/// request's instructions will be the URL.
+/// A logdog `LogRequest` represents a line from the config file. It starts with a "mode" that
+/// specifies what type of request it is, e.g. `exec ` for a command or `http` for an HTTP get
+/// request. Some modes then require a `filename` that determines where the data will be saved in
+/// the output tarball.  The final field is `instructions` which is any additional information
+/// needed by the given mode.  For example, an `exec` requests' instructions will include the
+/// program and program arguments. An `http` request's instructions will be the URL.
 ///
 /// # Examples
 ///
@@ -61,11 +65,18 @@ pub(crate) fn log_requests() -> Vec<&'static str> {
 /// ```text
 /// file some-conf /etc/some/conf
 /// ```
+///
+/// This request will copy files with a known prefix into the tarball; this can be useful for dated
+/// log files, for example.
+///
+/// ```text
+/// glob /var/log/my-app.log*
+/// ```
 #[derive(Debug, Clone)]
 struct LogRequest<'a> {
-    /// The log request mode. For example `exec`, `http`, or `file`.
+    /// The log request mode. For example `exec`, `http`, `file`, or `glob`.
     mode: &'a str,
-    /// The filename that the logs will be written to.
+    /// The filename that the logs will be written to, if appropriate for the mode.
     filename: &'a str,
     /// Any additional instructions or commands needed to fulfill the log request. For example, with
     /// `exec` this will be a program invocation like `echo hello world`. For an `http` request this
@@ -91,19 +102,30 @@ where
     P: AsRef<Path>,
 {
     let request = request.as_ref();
-    // get the first and second token: i.e. mode and output filename, put the remainder of the
-    // log request into the instructions field (or default to an empty string).
     let mut iter = request.splitn(3, ' ');
-    let req = LogRequest {
-        mode: iter.next().context(error::ModeMissing)?,
-        filename: iter.next().context(error::FilenameMissing { request })?,
-        instructions: iter.next().unwrap_or(""),
+    let mode = iter.next().context(error::ModeMissing)?;
+    let req = if mode == "glob" {
+        // for glob request format is "glob <pattern>"
+        LogRequest {
+            mode,
+            filename: "",
+            instructions: iter.next().context(error::PatternMissing)?,
+        }
+    } else {
+        // Get the second token (output filename) and put the remainder of the
+        // log request into the instructions field (or default to an empty string).
+        LogRequest {
+            mode,
+            filename: iter.next().context(error::FilenameMissing { request })?,
+            instructions: iter.next().unwrap_or(""),
+        }
     };
     // execute the log request with the correct handler based on the mode field.
     match req.mode {
         "exec" => handle_exec_request(&req, tempdir)?,
         "http" | "https" => handle_http_request(&req, tempdir)?,
         "file" => handle_file_request(&req, tempdir)?,
+        "glob" => handle_glob_request(&req, tempdir)?,
         unmatched => {
             return Err(error::Error::UnhandledRequest {
                 mode: unmatched.into(),
@@ -205,11 +227,105 @@ where
     Ok(())
 }
 
+/// Copies all files matching the glob pattern given by `request.instructions` to the tempdir with filename and path
+/// same as source file.
+fn handle_glob_request<P>(request: &LogRequest<'_>, tempdir: P) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    let mut files = HashSet::new();
+    let glob_paths = glob(request.instructions).context(error::ParseGlobPattern {
+        pattern: request.instructions,
+    })?;
+    for entry in glob_paths {
+        if let Ok(path) = entry {
+            if path.is_dir() {
+                // iterate the directory and sub-directory to get all file paths
+                for candidate in WalkDir::new(&path) {
+                    if let Ok(e) = candidate {
+                        if e.path().is_file() {
+                            files.insert(e.into_path());
+                        }
+                    }
+                }
+            } else {
+                files.insert(path);
+            }
+        }
+    }
+    for src_filepath in &files {
+        // with glob pattern there are chances of multiple targets with same name, therefore
+        // we maintain source file path and name in destination directory.
+        // Eg. src file path "/a/b/file" will be converted to "dest_dir/a/b/file"
+        let relative_path = src_filepath
+            .strip_prefix("/")
+            .unwrap_or(src_filepath.as_path());
+        let dest_filepath = tempdir.as_ref().join(relative_path);
+        let dest_dir_path = dest_filepath.parent().context(error::RootAsFile)?;
+        // create directories in dest file path if it does not exist
+        fs::create_dir_all(dest_dir_path).context(error::CreateOutputDirectory {
+            path: dest_dir_path,
+        })?;
+        let _ = fs::copy(&src_filepath, &dest_filepath).with_context(|| error::FileCopy {
+            request: request.to_string(),
+            from: src_filepath.to_str().unwrap_or("<unknown>"),
+            to: &dest_filepath,
+        })?;
+    }
+    Ok(())
+}
+
 #[cfg(test)]
 mod test {
     use crate::log_request::handle_log_request;
+    use std::fs;
     use std::fs::write;
+    use std::path::PathBuf;
     use tempfile::TempDir;
+
+    // adds a sub directory and some files to temp directory for file request tests
+    fn create_source_dir(dir: &TempDir) {
+        let filenames_content = [
+            ("foo.source", "1"),
+            ("bar.source", "2"),
+            ("for-bar.log", "3"),
+        ];
+        // add files to temp directory
+        for entry in filenames_content.iter() {
+            let filepath = dir.path().join(entry.0);
+            write(&filepath, entry.1).unwrap();
+        }
+
+        let subdir_name_depth1 = "depth1";
+        // create sub directory
+        let subdir_path_depth1 = dir.path().join(subdir_name_depth1);
+        fs::create_dir(&subdir_path_depth1).unwrap();
+        // Add files to sub directory
+        for entry in filenames_content.iter() {
+            let filepath = subdir_path_depth1.join(entry.0);
+            write(&filepath, entry.1).unwrap();
+        }
+
+        let subdir_name_depth2 = "depth2";
+        // create sub directory
+        let subdir_path_depth2 = subdir_path_depth1.join(subdir_name_depth2);
+        fs::create_dir(&subdir_path_depth2).unwrap();
+        // Add files to sub directory
+        for entry in filenames_content.iter() {
+            let filepath = subdir_path_depth2.join(entry.0);
+            write(&filepath, entry.1).unwrap();
+        }
+    }
+
+    fn get_dest_filepath(src_dir: &TempDir, filepath: &str) -> PathBuf {
+        src_dir.path().strip_prefix("/").unwrap().join(filepath)
+    }
+
+    fn assert_file_match(dest_dir: &TempDir, filepath: PathBuf, want: &str) {
+        let outfile = dest_dir.path().join(filepath);
+        let got = std::fs::read_to_string(&outfile).unwrap();
+        assert_eq!(got, want);
+    }
 
     #[test]
     fn file_request() {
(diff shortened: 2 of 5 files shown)


sources/logdog/src/log_request.rs
```diff
@@ -196,21 +198,72 @@ where
...
+    let multiple_files = glob_paths.count() > 1;
+    // iterator gets consumed in count, therefore re-create
+    glob_paths = glob(request.instructions).context(error::ParseGlobPattern {
+        pattern: request.instructions.to_string(),
     })?;
+    for entry in glob_paths {
+        if let Ok(path) = entry {
+            ensure!(path.is_file(), error::Directory { path });
+            // if there are more than one file then all the files will be written to same file,
+            // therefore keep destination filename same as source filename
```
tjkirch: I think trying to fit globs into the `file` mode is causing more trouble than it's worth, with the iterator count and file naming issues.  I'd suggest we switch to a new `glob /bla/*` log request format.

Either way, we should handle the possibility of multiple targets with the same name, which would currently overwrite each other, and wasn't a problem before because the target names were all written in the config file.  I think the simplest, clearest solution is probably to keep the full paths to the globbed files, i.e. /a/b/file would show up in the tarball at /a/b/file.

sources/logdog/src/error.rs
```diff
@@ -124,6 +127,12 @@ pub(crate) enum Error {
     #[snafu(display("Empty command."))]
     ModeMissing {},
 
+    #[snafu(display("Error parsing `glob` pattern {}: '{}'", pattern, source))]
```
tjkirch:
```suggestion
    #[snafu(display("Error parsing glob pattern '{}': {}", pattern, source))]
```

sources/logdog/src/log_request.rs
```diff
@@ -196,21 +198,72 @@ where
...
+        pattern: request.instructions.to_string(),
+    })?;
+    let multiple_files = glob_paths.count() > 1;
+    // iterator gets consumed in count, therefore re-create
+    glob_paths = glob(request.instructions).context(error::ParseGlobPattern {
+        pattern: request.instructions.to_string(),
     })?;
+    for entry in glob_paths {
+        if let Ok(path) = entry {
+            ensure!(path.is_file(), error::Directory { path });
```
tjkirch: This feels like a bit of a harsh restriction because globs can wind up with directories fairly easily, and it can be hard to avoid them; you might need to use multiple overly specific globs just to avoid a directory when you actually want its contents anyway.

I think it'd be pretty easy to remove this restriction by using `walkdir` if the current glob result is a directory.  Something like:
```rust
for entry in WalkDir::new(&path) {
   let _ = fs::copy ...etc...
}
```

samuelkarp:
> Trying to build an image with conf that exports awsvpc mode logs and make sure it works.

Do you want to actually update the config for the `aws-ecs-1` variant to use the new functionality here in a second commit on top of the one you already have?  https://github.com/bottlerocket-os/bottlerocket/issues/1481 calls out one of the log files, the other two that should be adjusted are [here](https://github.com/bottlerocket-os/bottlerocket/blob/dde4a12918cb7fc7bde8b25715125aafe024e087/sources/logdog/conf/logdog.aws-ecs-1.conf#L6-L7).

srgothi92:
> Do you want to actually update the config for the `aws-ecs-1` variant to use the new functionality here in a second commit on top of the one you already have? #1481 calls out one of the log files, the other two that should be adjusted are [here](https://github.com/bottlerocket-os/bottlerocket/blob/dde4a12918cb7fc7bde8b25715125aafe024e087/sources/logdog/conf/logdog.aws-ecs-1.conf#L6-L7).

Yes, I am planning to do that. Thanks for the pointer.

I raised draft PR to get team's feedback on two things:
1. If we would like to create a new request type that takes file pattern instead of single file. As of now conf pattern is `mode filename instruction`, which means it expects only one file per request. So i was not sure how I should handle the middle parameter `filename`. I decided on using file name provided if there is only one file matching the pattern, otherwise use source filename.
2. If we would like to support directory pattern, as of now I am throwing error for it.

Edit: I did not refresh the PR, now I see tjk@ feedback exactly for for this two :)

sources/logdog/src/log_request.rs
srgothi92: Yeah I thought of this, there were two reason why i decided on throwing error for directory pattern:
1. We do not need directory based pattern as of now. 
2. I did not want to introduce additional dependency `WalkDir` unless it was needed.

I am happy to handle this requirement if you feel we should.

sources/logdog/src/log_request.rs
srgothi92: My reasoning behind modifying file mode was: 
1. The glob in the end is going to target files, so it is a file based request.
2. Current request conf format is `mode filename instruction`, so i was not sure how to fit glob based requirement in this. Your suggestion of `glob /bla/*` will consider `/bla/*` pattern as filename in request argument which I thought was weird because `/bla/*` is actually an instruction of the source file. I had multiple thought of using `glob * /bla/*` all seemed weird to me. So, I finally settled on counting the number of files and making decision and ignore filename argument for multiple files.

Let me know what you think about this.

`/a/b/file` could be useful. I can do this, regardless of above conversation.

sources/logdog/src/log_request.rs
tjkirch: > We do not need directory based pattern as of now.

We expect it, though, and allowing directories (instead of globs) was another primary option to address this problem, so I think it makes sense.

> I did not want to introduce additional dependency WalkDir unless it was needed.

We use walkdir in a number of places, so we're already building and trusting it, and it's quite nice to use! :)

srgothi92: Okay, I will get started with this.

sources/logdog/src/log_request.rs
tjkirch: > Current request conf format is `mode filename instruction`, so i was not sure how to fit glob based requirement in this.

We can make `handle_log_request` check the first key in the line before parsing the request, to make the number of fields dynamic.  They don't *have* to have three fields per line, that was just convention for existing modes.

srgothi92: makes sense 👍

srgothi92:
> srgothi92 force-pushed the srgothi92:develop branch from 9654e34 to d0a24af 2 minutes ago

Addressed @tjk comments

sources/logdog/src/log_request.rs
webern: I agree with @tjkirch here. I think a separate 'command' for glob is more intuitive.

srgothi92: changed :)

sources/logdog/src/log_request.rs
```diff
@@ -205,12 +218,114 @@ where
     Ok(())
 }
 
+/// Copies all file matching the glob pattern given by `request.instructions` to the tempdir with filename and path
```
webern: Grammar nit

```suggestion
/// Copies all files matching the glob pattern given by `request.instructions` to the tempdir with filename and path
```

sources/logdog/src/log_request.rs
```diff
@@ -205,12 +218,114 @@ where
...
+    P: AsRef<Path>,
+{
+    let glob_paths = glob(request.instructions).context(error::ParseGlobPattern {
+        pattern: request.instructions.to_string(),
+    })?;
+    for entry in glob_paths {
+        if let Ok(path) = entry {
+            if path.is_dir() {
+                // iterate the directory and copy files in the directory, nested
+                // sub-directories files will be copied only if matched by glob pattern.
```
webern: Grammar nit, but I also kind of don't understand what it means.

```suggestion
                // sub-directory files will be copied only if matched by glob pattern.
```

sources/logdog/src/log_request.rs
```diff
@@ -205,12 +218,114 @@ where
...
+                for candidate in WalkDir::new(&path)
+                    .max_depth(1)
```
webern: I don't quite understand the behavior for directories.
If my glob is `/etc/foo/*`, would both of these files be picked up?

```
/etc/foo/file.txt
/etc/foo/dir1/dir2/dir3/file.txt
```

Maybe you can just copy directories that you encounter with something like `fs_extra`?

sources/logdog/src/log_request.rs
```diff
@@ -205,12 +218,114 @@ where
...
+    Ok(())
+}
+
+fn file_copy<P>(request: &LogRequest<'_>, destdir: P, src_filepath: &PathBuf) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    // with glob pattern there are chances of multiple targets with same name, therefore
+    // we maintain source file path and name in destination directory.
+    // Eg. src file path "/a/b/file" will be converted to "dest_dir/a/b/file"
```
webern: It's too bad that I didn't do this with all the files we already collect, but this behavior makes sense. 

Unfortunately it might be a little confusing when you unpack that tarball and some things are at the top and other things replicate their on-host paths. One possibility would be to put all of these paths into a dir at the top of the tarball like `/glob`, but I don't think that's any better.

sources/logdog/src/log_request.rs
srgothi92: > If my glob is /etc/foo/*, would both of these files be picked up?

This is what I tried explaining by comment : "sub-directories files will be copied only if matched by glob pattern". :P. Basically `glob` and `walkdir` both can iterate sub-directories, if we let both of them iterate we will have conflict. So, I am only copying sub-directories files if it matches `glob pattern`. In this case, `/etc/**/` would match for all directories and sub directories of `/etc` and `/etc/foo/*` only matches `/etc/foo` dir.

sources/logdog/src/log_request.rs
tjkirch: It's not perfect, but I think it works out OK, because the file/command entries are probably the most important ones, and you can give them particular names and have easier access to them, and the glob entries can grab any extras...

sources/logdog/src/log_request.rs
tjkirch: I'm not sure we need to care about that, or the max_depth.  If we hit a directory, then the user specified that directory in their glob, so I think we should copy all of it with a plain walkdir, no max_depth.  If there are duplicates, we could either copy again (this isn't performance critical) or make a hashset of files to copy and do it once at the end.  (With the set, we could get rid of the file_copy function, which I think raises questions about why we're not using it for "file" mode...)

sources/logdog/src/error.rs
```diff
@@ -57,6 +57,9 @@ pub(crate) enum Error {
         backtrace: Backtrace,
     },
 
+    #[snafu(display("Failed to walk directory to find log files: {}", source))]
+    DirectoryWalk { source: walkdir::Error },
```
tjkirch: We should include which directory we failed to walk.

sources/logdog/src/log_request.rs
```diff
@@ -91,19 +93,30 @@ where
...
 {
     let request = request.as_ref();
-    // get the first and second token: i.e. mode and output filename, put the remainder of the
-    // log request into the instructions field (or default to an empty string).
-    let mut iter = request.splitn(3, ' ');
-    let req = LogRequest {
-        mode: iter.next().context(error::ModeMissing)?,
-        filename: iter.next().context(error::FilenameMissing { request })?,
-        instructions: iter.next().unwrap_or(""),
+    let req = if request.starts_with("glob") {
```
tjkirch: This would match "globlksjdlfk" too, and the error check below is a little awkward.  I think we can `splitn(2, ' ')` to get the mode and the remainder, and then check the first field exactly.  Then we can split again if we're not doing a glob.

sources/logdog/src/log_request.rs
```diff
@@ -91,19 +93,30 @@ where
...
+        // For glob conf format is "glob pattern"
+        let mut iter = request.splitn(2, ' ');
+        LogRequest {
+            mode: iter.next().context(error::ModeMissing)?,
+            filename: "",
+            instructions: iter.next().unwrap_or(""),
```
tjkirch: I think we should `.context` this and fail earlier if it's missing; trying to parse it as an empty glob could be confusing.  (The LogRequest below has an `unwrap_or` because not all types would need the instructions, but this code is for a specific mode that we know needs them.)

sources/logdog/src/log_request.rs
```diff
@@ -205,12 +218,114 @@ where
...
+    for entry in glob_paths {
+        if let Ok(path) = entry {
+            if path.is_dir() {
+                // iterate the directory and copy files in the directory, nested
+                // sub-directories files will be copied only if matched by glob pattern.
+                for candidate in WalkDir::new(&path)
+                    .max_depth(1)
+                    .into_iter()
+                    .flat_map(|e| e.context(error::DirectoryWalk))
+                    .map(|e| e.into_path())
```
```suggestion
                    .into_iter()
                    .filter_map(|e| e.ok())
```
For the glob piece, we do `if let Ok` to ignore errors; I think it'd be fine to have the same behavior here.  (The suggestion is the standard way to do that with walkdir.)  We're trying to get as much information as possible, and we wouldn't want to stop collecting more files because of issues with a prior one.

sources/logdog/src/log_request.rs
```diff
@@ -205,12 +218,114 @@ where
...
+    P: AsRef<Path>,
+{
+    // with glob pattern there are chances of multiple targets with same name, therefore
+    // we maintain source file path and name in destination directory.
+    // Eg. src file path "/a/b/file" will be converted to "dest_dir/a/b/file"
+    let relative_path = &src_filepath
+        .strip_prefix("/")
+        .unwrap_or(src_filepath);
+    let dest_filepath = destdir.as_ref().join(relative_path);
+    let dest_dir_path = dest_filepath.parent().unwrap_or(&Path::new(""));
```
tjkirch: I don't think we should continue and try to create_dir_all a path of `""` - we should probably fail here if we somehow came up with a `dest_filepath` of `/`.

sources/logdog/src/log_request.rs
```diff
@@ -205,12 +218,114 @@ where
...
+    // create directories in dest file path if it does not exist
+    fs::create_dir_all(dest_dir_path).context(error::CreateOutputDirectory {
+        path: dest_dir_path,
+    })?;
+    let _ = fs::copy(&src_filepath, &dest_filepath).with_context(|| error::FileCopy {
+        request: request.to_string(),
+        from: src_filepath.to_str().unwrap_or_default(),
```
```suggestion
        from: src_filepath.to_str().unwrap_or("<unknown>"),
```
Gives slightly more context in the error message, rather than having a blank.

sources/logdog/src/log_request.rs
```diff
@@ -205,12 +218,114 @@ where
...
 }
 
+/// Copies all file matching the glob pattern given by `request.instructions` to the tempdir with filename and path
+/// same as source file.
+fn handle_glob_request<P>(request: &LogRequest<'_>, tempdir: P) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    let glob_paths = glob(request.instructions).context(error::ParseGlobPattern {
+        pattern: request.instructions.to_string(),
```
```suggestion
        pattern: request.instructions,
```
Shouldn't need to convert `&str` to `String`, snafu does that for you, and it only converts if it's needed during an actual error situation.

sources/logdog/src/error.rs
```diff
@@ -124,6 +124,15 @@ pub(crate) enum Error {
...
     ModeMissing {},
 
+    #[snafu(display("Error parsing glob pattern '{}': {}", pattern, source))]
+    ParseGlobPattern {
+        pattern: String,
+        source: glob::PatternError,
+    },
+
+    #[snafu(display("Glob pattern missing"))]
+    PatternMissing {},
```
tjkirch, line 134: It'd be helpful to give a pointer to the file we read that had the missing pattern.

sources/logdog/src/log_request.rs
```diff
@@ -91,19 +94,31 @@ where
...
+            mode,
+            filename: "",
+            instructions: conf.next().context(error::PatternMissing)?,
+        }
+    } else {
+        // get the first and second token: i.e. mode and output filename, put the remainder of the
+        // log request into the instructions field (or default to an empty string).
+        let mut iter = request.splitn(3, ' ');
+        LogRequest {
+            mode: iter.next().context(error::ModeMissing)?,
```
```suggestion
        // Get the second token (output filename) and put the remainder of the
        // log request into the instructions field (or default to an empty string).
        let mut iter = conf.splitn(2, ' ');
        LogRequest {
            mode,
```
Since we already have that information...

srgothi92:
addressed @tjk's comments.

sources/logdog/src/error.rs
srgothi92: I am not sure if I understood your comment, there is no file. This error will occur when someone gives "glob", basically forgot the pattern part in conf.

webern: Yeah these 'conf' files are pulled in statically at compile time.

sources/logdog/src/log_request.rs
webern: ...or just copy the directory itself rather than walking it, right?

sources/logdog/src/error.rs
tjkirch: Ah, then most users would be using what we distribute, so they couldn't change it.  Maybe we should just expand this error message, then.  If we mess up testing, and users only see "Glob pattern missing", it seems like it could be a little confusing.  How about something like ``The logdog configuration has a "glob" line with no glob instructions.  Please check `sources/logdog/conf` and/or file an issue with Bottlerocket.``?

sources/logdog/src/log_request.rs
tjkirch: Copying would always involve walking the structure, even if we farm the job out to another crate, but if we farm it out we run the risk of copying files-in-directories differently than we do files, and we already use walkdir and it's so simple...

sources/logdog/src/error.rs
srgothi92: I feel the message is too verbose. We are creating the conf file, so ideally we would never see this error in actual run. 
Still if it occurs I feel message "The logdog configuration has a "glob" line with no glob instructions" should be good enough ?

sources/logdog/src/error.rs
```diff
@@ -124,6 +124,15 @@ pub(crate) enum Error {
     #[snafu(display("Empty command."))]
     ModeMissing {},
 
+    #[snafu(display("Error parsing glob pattern '{}': {}", pattern, source))]
+    ParseGlobPattern {
+        pattern: String,
+        source: glob::PatternError,
+    },
+
+    #[snafu(display("The logdog configuration has a "glob" line with no glob instructions."))]
```
tjkirch:
```suggestion
    #[snafu(display("The logdog configuration has a 'glob' line with no glob instructions."))]
```

sources/logdog/src/log_request.rs
```diff
@@ -91,19 +94,31 @@ where
...
+        // for glob request format is "glob <pattern>"
+        LogRequest {
+            mode,
+            filename: "",
+            instructions: conf.next().context(error::PatternMissing)?,
+        }
+    } else {
+        // Get the second token (output filename) and put the remainder of the
+        // log request into the instructions field (or default to an empty string).
+        let mut iter = request.splitn(3, ' ');
```
```suggestion
        let mut iter = request.splitn(2, ' ');
```
Already split out the mode.

tjkirch (changes requested):
Would you please confirm you've re-tested what's in the PR description?

sources/logdog/src/log_request.rs
srgothi92: > Would you please confirm you've re-tested what's in the PR description?

I did test it before the last push, there was only minor change after that, sorry that was a miss. I will re-test.

srgothi92: re-tested everything, CI is also passing now.

webern (approved):

sources/logdog/src/log_request.rs
```diff
@@ -9,13 +9,16 @@
 //! these provide the list of log requests that `logdog` will run.
```
tjkirch, line 9: I can't leave a suggestion on the appropriate lines since they didn't change, but we should update the documentation since it's no longer accurate with the new mode.  I suggest this change:

```diff
--- a/sources/logdog/src/log_request.rs
+++ b/sources/logdog/src/log_request.rs
@@ -33,11 +33,12 @@ pub(crate) fn log_requests() -> Vec<&'static str> {
         .collect()
 }
 
-/// A logdog `LogRequest` is in the format `mode filename instructions`. `mode` specifies what type
-/// of command it is, e.g. `exec ` for a command or `http` for an HTTP get request. `filename` is
-/// the name of the output file. `instructions` is any additional information needed.  For example,
-/// an `exec` request'ss instructions will include the program and program arguments. An `http`
-/// request's instructions will be the URL.
+/// A logdog `LogRequest` represents a line from the config file.  It starts with a "mode" that
+/// specifies what type of request it is, e.g. `exec ` for a command or `http` for an HTTP get
+/// request. Some modes then require a `filename` that determines where the data will be saved in
+/// the output tarball.  The final field is `instructions` which is any additional information
+/// needed by the given mode.  For example, an `exec` requests' instructions will include the
+/// program and program arguments. An `http` request's instructions will be the URL.
 ///
 /// # Examples
 ///
@@ -61,11 +62,18 @@ pub(crate) fn log_requests() -> Vec<&'static str> {
 /// ```text
 /// file some-conf /etc/some/conf
 /// ```
+///
+/// This request will copy files with a known prefix into the tarball; this can be useful for dated
+/// log files, for example.
+///
+/// ```text
+/// glob /var/log/my-app.log*
+/// ```
 #[derive(Debug, Clone)]
 struct LogRequest<'a> {
-    /// The log request mode. For example `exec`, `http`, or `file`.
+    /// The log request mode. For example `exec`, `http`, `file`, or `glob`.
     mode: &'a str,
-    /// The filename that the logs will be written to.
+    /// The filename that the logs will be written to, if appropriate for the mode.
     filename: &'a str,
     /// Any additional instructions or commands needed to fulfill the log request. For example, with
     /// `exec` this will be a program invocation like `echo hello world`. For an `http` request this
```

tjkirch (approved):

sources/logdog/src/log_request.rs
srgothi92: Thanks for the complete write-up. 
I see that you already approved, still fixed in `srgothi92 force-pushed the srgothi92:develop branch from 97a8a4e to 6e258a8 5 minutes ago` :P

PR #1516, opened 2021-04-22
release: add new udev rules for mounting CD-ROMs

### LINKED ISSUE #1427: `media-cdrom.mount` fails if the device exists but no media is present
**Image I'm using:**
vmware-dev

A system with a CD-ROM drive attached _but with no media present_ shouldn't cause any systemd units to fail.  The `media-cdrom.mount` unit should only attempt to mount if there is an ISO present.

**What actually happened:**
The `media-cdrom.mount` unit fails attempting to mount an ISO (that doesn't exist).
```
localhost systemd[1]: Mounting CD-ROM mount (/media/cdrom)...
localhost mount[3778]: mount: /media/cdrom: no medium found on /dev/sr0
```

* Attach a CD-ROM device
* Don't include any media in said CD-ROM

etungsten:
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/1427

**Description of changes:**
```
Date:   Wed Apr 14 17:03:22 2021 -0700

    release: add new udev rules for mounting cdroms
    
    Adds new udev rules for activating the media-cdrom.mount unit whenever
    we detect media in the CD drive. Removes `WantedBy=` from the mount unit
    as it's no longer needed.
```

**Testing done:**
Build OVA and created VM in vSphere.

Booting the VM with a CD-Drive connected with datastore stored on a CD-ROM works as expected. The CD-ROM gets mounted correctly, `early-boot-config` is able to read user-data from it.:
boot-with-cdrom

If I remove the CD-ROM from the CD-drive afterwards, the `media-cdrom.mount` goes inactive as does `dev-cdrom.device`:
remove-cd-after

Next, if I boot the VM with no CD-ROM in the drive. The host comes up fine with both `media-cdrom.mount` and `dev-cdrom.device` inactive. If I insert a CD-ROM afterwards, they become active as expected and the CD-ROM is mounted successfully at `/media/cdrom`.

**Terms of contribution:**

### DIFF
--- a/packages/release/media-cdrom.mount
+++ b/packages/release/media-cdrom.mount
@@ -3,7 +3,7 @@ Description=CD-ROM mount (/media/cdrom)
 # Only run this unit if /dev/cdrom exists and is tracked via systemd. (systemd
 # ships with a udev rule to tag and symlink the first suspected cdrom device to
 # /dev/cdrom)
-Requires=dev-cdrom.device
+BindsTo=dev-cdrom.device
 After=dev-cdrom.device
 DefaultDependencies=no
 Conflicts=umount.target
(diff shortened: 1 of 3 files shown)


bcressey (approved):

tjkirch (approved):

zmrow (approved):
:tada: 

Nice work!!

PR #1521, opened 2021-04-23
pubsys: validate-repo without tokio runtime

webern:
**Issue number:**

**Description of changes:**

We were seeing `pubsys validate-repo` hang and we believe it was because of multiple tokio runtimes. Although it's not clear why `tokio::spawn` was working before, re-implementing this to use `std::thread::spawn` instead has fixed the problem.

**Testing done:**

Created an Infra.toml like this:

```toml
[repo.2020-07-07]
metadata_base_url = "https://updates.bottlerocket.aws/2020-07-07/"
targets_url = "https://updates.bottlerocket.aws/targets/"
root_role_url = "https://cache.bottlerocket.aws/root.json"
root_role_sha512 = "90393204232a1ad6b0a45528b1f7df1a3e37493b1e05b1c149f081849a292c8dafb4ea5f7ee17bcc664e35f66e37e4cfa4aae9de7a2a28aa31ae6ac3d9bea4d5"
```

And this a few times:

```sh
cargo make validate-repo \
    -e PUBLISH_REPO=2020-07-07 \
    -e BUILDSYS_VARIANT=aws-k8s-1.15 \
    -e BUILDSYS_ARCH=x86_64 \
    -e PUBLISH_INFRA_CONFIG_PATH="${INFRA_TOML}"
```

I also edited a target in a repo and verified that the program errors out when the repo is bad:

```
Failed to validate repository: Failed to download and write target 'migrate_v1.0.6_metricdog-init.lz4': Hash mismatch for https://somewhere.amazonaws.com/somewhere/targets/17a0a6e4f95471734b77091212ccabf564d3312062fc98c995002df2fbf7c6db.migrate_v1.0.6_metricdog-init.lz4: calculated 2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae, expected 17a0a6e4f95471734b77091212ccabf564d3312062fc98c995002df2fbf7c6db
```
**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/repo/validate_repo/mod.rs
+++ b/tools/pubsys/src/repo/validate_repo/mod.rs
@@ -3,13 +3,13 @@
 
 use crate::repo::{error as repo_error, repo_urls};
 use crate::Args;
-use futures::future::join_all;
 use log::{info, trace};
 use pubsys_config::InfraConfig;
 use snafu::{OptionExt, ResultExt};
 use std::fs::File;
 use std::io;
 use std::path::PathBuf;
+use std::thread::spawn;
 use structopt::StructOpt;
 use tough::{Repository, RepositoryLoader};
 use url::Url;
@@ -39,7 +39,7 @@ pub(crate) struct ValidateRepoArgs {
 }
 
 /// Retrieves listed targets and attempts to download them for validation purposes
-async fn retrieve_targets(repo: &Repository) -> Result<(), Error> {
+fn retrieve_targets(repo: &Repository) -> Result<(), Error> {
     let targets = &repo.targets().signed.targets;
 
     let mut tasks = Vec::new();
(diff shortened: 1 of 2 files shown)


etungsten (approved):

tools/pubsys/src/repo/validate_repo/mod.rs
```diff
@@ -54,22 +53,35 @@ async fn retrieve_targets(repo: &Repository) -> Result<(), Error> {
                 target: target.to_string(),
             })?;
         info!("Downloading target: {}", target);
-        tasks.push(tokio::spawn(async move {
+        // TODO - limit threads https://github.com/bottlerocket-os/bottlerocket/issues/1522
+        tasks.push(std::thread::spawn(move || {
```
tjkirch: nit: `use std::thread;` is more standard.

bcressey (approved):

tools/pubsys/src/repo/validate_repo/mod.rs
```diff
@@ -54,22 +53,35 @@ async fn retrieve_targets(repo: &Repository) -> Result<(), Error> {
...
+    let mut results = Vec::new();
+    for task in tasks {
+        let result = task.join().map_err(|e| error::Error::Join {
+            // the join function is returning an error type that does not implement error or display
+            inner: format!("{:?}", e),
+        })?;
+        results.push(result);
+    }
+
+    // check all results and return the first error we see
```
tjkirch, line 76: Is it worthwhile to wait for every thread, pushing errors into the `results` list, and only checking them down there, if we error on the first one?  Seems like we could `?` in the upper loop and drop the remaining (useless) threads on the floor, if we don't want to see all errors.  (Yes, this was an existing issue...)

tjkirch (approved):
LGTM.  ~~I'd like to see a test confirming that bad cases still fail as we expect, though, ideally.~~ :heavy_check_mark:

tools/pubsys/src/repo/validate_repo/mod.rs
webern: We could, but it seems improper to bail out without making a reasonable effort to join all threads.

tjkirch: That's true, we'd have to return the unjoined threads with the error if we wanted to allow for callers who don't exit immediately.

tools/pubsys/src/repo/validate_repo/mod.rs
webern: done

webern: done (now in the right place)

PR #1526, opened 2021-04-27
add support for kernel 5.10

### LINKED ISSUE #1349: update default sysctls for 5.10 kernel
@sam-aws gave us a heads-up that some of the sysctl defaults that are currently set in the kernel will need to be added by userspace for 5.10, if we want to keep them.

```
# Increasing to account for skb structure growth since the 3.4.x kernel series
net.ipv4.tcp_wmem = 4096 20480 4194304

# Bumped the default TTL to 255 (maximum)
net.ipv4.ip_default_ttl = 255
```

### LINKED ISSUE #1454: build: support kmod kits for kernel 5.10
**What I'd like:**

Currently our build assumes that all the variants all use the same version of the linux kernel. As part of kernel 5.10 support, we need to make sure that the Makefile.toml archive and build-kmod kits targets do the right thing based on which kernel a variant is using.

This is related to #1361

bcressey:
**Issue number:**
Fixes #1349 
Fixes #1454 

**Description of changes:**
Rename the kernel package to kernel-5.4, and add kernel-5.10. Move kernel selection to variant definitions, and adjust kmod-kit builds to support different kernels.

Add sysctl settings to `release` to keep defaults the same for both kernels.

Add new classes and permissions to the SELinux policy.

**Testing done:**
The kernel spec files have only [minor differences](https://gist.github.com/bcressey/6e39351ae5795e02ff2f16ac9db8a487). The Lustre patch is no longer needed, and a few extra files are required in the devel archive.

For the 5.10 kernel, I confirmed that kmod-kits worked for both targets using artifacts built on both hosts. I built ZFS, copied modules to running hosts, and confirmed that they would load. For x86_64, I also installed Falco using helm, and confirmed that the kernel module loaded.

For `aws-k8s-1.19` (kernel 5.4) and `aws-k8s-1.20` (kernel 5.10) I confirmed that the sysctl values were set as expected:
```
$ sysctl net.ipv4.tcp_wmem net.ipv4.ip_default_ttl
net.ipv4.tcp_wmem = 4096	20480	4194304
net.ipv4.ip_default_ttl = 255
```

I also confirmed that the SELinux policy loaded correctly on 5.10, and with an informational message on kernel 5.4.

**aws-k8s-1.20**
```
[    0.381563] SELinux:  policy capability network_peer_controls=1
[    0.382039] SELinux:  policy capability open_perms=1
[    0.382418] SELinux:  policy capability extended_socket_class=1
[    0.382868] SELinux:  policy capability always_check_network=0
[    0.383313] SELinux:  policy capability cgroup_seclabel=1
[    0.383725] SELinux:  policy capability nnp_nosuid_transition=1
[    0.384176] SELinux:  policy capability genfs_seclabel_symlinks=1
```

**aws-k8s-1.19**
```
[    1.376085] SELinux:  policy capability network_peer_controls=1
[    1.376727] SELinux:  policy capability open_perms=1
[    1.377245] SELinux:  policy capability extended_socket_class=1
[    1.377847] SELinux:  policy capability always_check_network=0
[    1.378454] SELinux:  policy capability cgroup_seclabel=1
[    1.379015] SELinux:  policy capability nnp_nosuid_transition=1
[    1.379617] SELinux:  unknown policy capability 6
```

kmod-kits were correctly linked in by `cargo make repo`:
```
❯ ls -l build/repos/default/bottlerocket-1.0.8-*/targets/*kmod-kit*
lrwxrwxrwx. 1 fedora fedora 110 Apr 27 00:08 build/repos/default/bottlerocket-1.0.8-413a3778/targets/344af7281c2daa3ceec9b798a20edb323e8ddddd75bd319efe5fda04baf1e6d4.aws-k8s-1.19-aarch64-kmod-kit-v1.0.8.tar.xz -> /home/fedora/bottlerocket/build/images/aarch64-aws-k8s-1.19/latest/aws-k8s-1.19-aarch64-kmod-kit-v1.0.8.tar.xz

❯ ls -l build/repos/default/bottlerocket-1.0.8-*/targets/*kmod-kit*
lrwxrwxrwx. 1 fedora fedora 108 Apr 27 00:18 build/repos/default/bottlerocket-1.0.8-413a3778/targets/ed7d014aefe7143565daabcfb277e1df01eb2a7d45214dbbc6ad22aa704fe969.aws-k8s-1.19-x86_64-kmod-kit-v1.0.8.tar.xz -> /home/fedora/bottlerocket/build/images/x86_64-aws-k8s-1.19/latest/aws-k8s-1.19-x86_64-kmod-kit-v1.0.8.tar.xz
```

**Terms of contribution:**


bcressey:
All the builds for arm64 failed, so I'm trying something different to see if I can placate CI.

tools/buildsys/src/builder.rs
```diff
@@ -390,6 +394,15 @@ fn getenv(var: &str) -> Result<String> {
     env::var(var).context(error::Environment { var })
 }
 
+/// Map a Linux architecture into the corresponding Docker architecture.
```
webern: An enum with `fn goarch(&self) -> &'static str` might be nice.

packages/kernel-5.10/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[package.metadata.build-package]
+package-name = "kernel-5.10"
+
+[lib]
+path = "pkg.rs"
+
+[[package.metadata.build-package.external-files]]
+# Use latest-srpm-url.sh to get this.
+url = "https://cdn.amazonlinux.com/blobstore/fa04b98fc067a4943beac60d0c2971e2fbef1a29faed4bac1c4096abe4ad4c12/kernel-5.10.29-27.126.amzn2.src.rpm"
+sha512 = "47341f4a1c13ba7e5ea72bad13fe689eefd22cc7547aea08a08fe47238b4a3fe1659786a406b84a1d1508143be20d9be2fae6fe3e7a6924bc85043bf61d4bfce"
```
webern, line 17: This seems to be missing the `[dependencies]` and `[build-dependencies]`. I would say all (external) dependencies should be represented even if some are commented out.

bcressey: The kernel doesn't have any dependencies within the project, since it's a freestanding artifact that doesn't pull in any libraries.

I'll remove the `filesystem` dependency from the kernel.spec & Cargo.toml to help clarify this.

Dockerfile
```diff
@@ -8,7 +8,12 @@
 # filesystem at /host.
```
tjkirch, line 8: For lack of a better place to comment - we should also update the `packages/kernel/` link near the end of the top-level README.md.  I guess I'd point at 5.4 for now, since it's the kernel for our default variant?

packages/kernel-5.10/latest-srpm-url.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+docker run --rm amazonlinux:2 sh -c 'amazon-linux-extras enable kernel-5.10 >/dev/null && yum install -q -y yum-utils && yumdownloader -q --source --urls kernel | grep ^http'
```
tjkirch, line 2: I think we might want to switch the 5.4 script to use the `kernel-5.4` extra, to be sure we're correct even if `kernel-ng` switches to 5.10.

tools/rpm2kmodkit
```diff
@@ -0,0 +1,46 @@
...
+KMOD_KIT_FULL="bottlerocket-${VARIANT}-${ARCH}-${VERSION_ID}-${BUILD_ID}-kmod-kit"
+
+EXTRACT_DIR="$(mktemp -d)"
+KIT_DIR="$(mktemp -d)"
+mkdir -p "${OUTPUT_DIR}" "${KIT_DIR}/${KMOD_KIT}"
+
+# Extract any RPMs and find the kernel development archive.
+pushd "${EXTRACT_DIR}" >/dev/null
+find "${ARCHIVE_DIR}" -type f -name '*.rpm' \
+  -exec rpm2cpio {} \; | cpio -idmu --quiet
```
tjkirch: Do we need the `-u`?  Are there cases where we have to replace files from prior RPMs?  That seems like a problem waiting to happen, and it could be difficult to debug if file contents mysteriously change from what you expect.  (I know it was there before, so call it a nit, but we're adding more RPMs...)

Makefile.toml
```diff
@@ -114,7 +114,7 @@ PUBLISH_AMI_NAME_DEFAULT = "${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH
 
 # The name of the kmod kit archive, used to ease building out-of-tree kernel modules.
 BUILDSYS_KMOD_KIT = "${BUILDSYS_VARIANT}-${BUILDSYS_ARCH}-kmod-kit-v${BUILDSYS_VERSION_IMAGE}"
-BUILDSYS_KMOD_KIT_PATH="${BUILDSYS_ARCHIVES_DIR}/${BUILDSYS_KMOD_KIT}.tar.xz"
+BUILDSYS_KMOD_KIT_PATH="${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_KMOD_KIT}.tar.xz"
```
tjkirch, line 117: ...should we move the OVA into OUTPUT and get rid of /archives?

packages/kernel-5.10/Cargo.toml
webern: Oh I see.

webern (approved):

Makefile.toml
bcressey: I'm planning to do this on the way to fixing #1528.

tjkirch (approved):

arnaldo2792 (approved):
:boot:

webern (approved):
🎸

PR #1529, opened 2021-04-28
add open-vm-tools and dependencies

### LINKED ISSUE #1510: VMWare: Run `open-vm-tools` to report guest information to vSphere
## Problem
While testing for the upcoming VMWare Kubernetes variant (#1451), we noticed that Bottlerocket VMs do not report their DNS name or IP addresses to vSphere.  This is because we don't currently package or run VMWare tools in our images.

When attempting to join these images to a cluster previously provisioned via CAPV, Bottlerocket nodes were able to join the cluster but could never have pods scheduled on them because the `node.cloudprovider.kubernetes.io/uninitialized` taint was never removed by the cloud-provider.  Upon further investigation we discovered that the vSphere cloud controller manager pod (`
kube-system vsphere-cloud-controller-manager`) wasn't able to initialize our nodes.  This is because it appears to query vCenter for VM data (that doesn't exist because we don't report it).
```
I0421 20:33:10.905534       1 node_controller.go:325] Initializing node 198.19.31.255 with cloud provider
E0421 20:33:10.921307       1 datacenter.go:107] Unable to find VM by DNS Name. VM DNS Name: 198.19.31.255
E0421 20:33:10.921331       1 search.go:181] Error while looking for vm=198.19.31.255(byName) in vc=<redacted> and datacenter=<redacted>: No VM found
I0421 20:33:10.936035       1 nodemanager.go:123] Discovered VM using IP address
E0421 20:33:10.940196       1 node_controller.go:370] failed to initialize node 198.19.31.255 at cloudprovider: failed to get instance ID from cloud provider: VM Guest hostname is empty
```

For testing #1451 , we were able to run an [open-vm-tools](https://hub.docker.com/r/immutablecode/open-vm-tools) host container with `superpowered=true`.  This allowed the DNS name and IP's to show up in vSphere.

In testing, simply running the one shot command `/usr/bin/vmware-toolbox-cmd info update network` with this container [as explained here](https://www.virtuallyghetto.com/2019/04/how-to-immediately-refresh-the-network-guestinfo-using-vmware-tools.html) didn't to the trick and info was never updated.  The container needed to run continuously to satisfy the polling interval explained in the previously mentioned article.

### Run `open-vm-tools` as a host container
This is what we did during testing and it works just fine.  The disadvantage is we would probably want to maintain our own container for it and this is an ongoing maintenance burden.  Another disadvantage is that the container would need to be pulled at runtime and any issue with that pull results in a non-operable k8s VM.

### Write our own tool to publish this information
We could write our own Rust tool to respond to the polling and also to publish this information to vSphere.  No crate exists to do this (though we could look into extending `vmw_backdoor`), so we would need to reference existing code or... ?  Doing this will _not_ be quick and our expertise lies elsewhere, so it's probably not the best option.

### Package `open-vm-tools` and run it as a host process via systemd
This option might serve us better.  Rather than maintaining another host container we can run the tools as a daemon via systemd.  Having a package means we don't need to build/publish yet another host container, don't need to worry about migrations and the necessary tooling around them.

bcressey:
**Issue number:**
Fixes #1510

**Description of changes:**
Add `open-vm-tools` and its dependencies: glib2, libffi, libtirpc.

Make a minor fix to the `libmount` package in `util-linux` to correctly express development dependencies.

Adjust `libdbus` build to work correctly if `libselinux-devel` is present in the SDK.

Update to a newer SDK that contains `glib-genmarshal` and `genrpc`. 

**Testing done:**
Built the `vmware-dev` image. Verified that the DNS name and IP addresses are visible in vCenter, and that VMWare Tools reports as running.

**Terms of contribution:**

### DIFF
--- a/packages/libglib/libglib.spec
+++ b/packages/libglib/libglib.spec
@@ -1,0 +1,70 @@
+Name: %{_cross_os}libglib
+Version: 2.68.1
+Release: 1%{?dist}
+Summary: The GLib libraries
+License: LGPL-2.1-or-later
+URL: https://www.gtk.org/
+Source0: https://download.gnome.org/sources/glib/2.68/glib-%{version}.tar.xz
+BuildRequires: meson
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: %{_cross_os}libffi-devel
+BuildRequires: %{_cross_os}libmount-devel
+BuildRequires: %{_cross_os}libpcre-devel
+BuildRequires: %{_cross_os}libselinux-devel
+BuildRequires: %{_cross_os}libz-devel
+Requires: %{_cross_os}libffi
+Requires: %{_cross_os}libmount
+Requires: %{_cross_os}libpcre
+Requires: %{_cross_os}libselinux
+Requires: %{_cross_os}libz
+
+%description
+%{summary}.
+
+%package devel
+Summary: Files for development using the GLib libraries
+Requires: %{name}
+Requires: %{_cross_os}libffi-devel
+
+%description devel
+%{summary}.
+
+%prep
+%autosetup -n glib-%{version} -p1
+
+%build
+CONFIGURE_OPTS=(
+ -Dlibmount=enabled
+ -Dselinux=enabled
+
+ -Dlibelf=disabled
+ -Dnls=disabled
+
+ -Dman=false
+ -Dtests=false
+)
+
+%cross_meson "${CONFIGURE_OPTS[@]}"
+%cross_meson_build
+
+%install
+%cross_meson_install
+
+%files
+%{_cross_attribution_file}
+%{_cross_libdir}/*.so.*
+%exclude %{_cross_datadir}
+
+%files devel
+%{_cross_bindir}/*
+%{_cross_libdir}/*.so
+%dir %{_cross_libdir}/glib-2.0
+%dir %{_cross_libdir}/glib-2.0/include
+%{_cross_libdir}/glib-2.0/include/glibconfig.h
+%dir %{_cross_includedir}/gio-unix-2.0
+%dir %{_cross_includedir}/glib-2.0
+%{_cross_includedir}/gio-unix-2.0/*
+%{_cross_includedir}/glib-2.0/*
+%{_cross_pkgconfigdir}/*.pc
+
+%changelog
--- a/packages/open-vm-tools/open-vm-tools.spec
+++ b/packages/open-vm-tools/open-vm-tools.spec
@@ -1,0 +1,119 @@
+%global buildver 17337674
+
+Name: %{_cross_os}open-vm-tools
+Version: 11.2.5
+Release: 1%{?dist}
+Summary: Tools for VMware
+License: LGPL-2.1-or-later
+URL: https://github.com/vmware/open-vm-tools
+Source0: https://github.com/vmware/open-vm-tools/releases/download/stable-%{version}/open-vm-tools-%{version}-%{buildver}.tar.gz
+Source1: vmtoolsd.service
+Source2: tools.conf
+Source3: open-vm-tools-tmpfiles.conf
+Patch0001: 0001-no_cflags_werror.patch
+Patch0002: 0002-dont-force-cppflags.patch
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: %{_cross_os}libglib-devel
+BuildRequires: %{_cross_os}libtirpc-devel
+BuildRequires: %{_cross_os}libxcrypt-devel
+Requires: %{_cross_os}libglib
+Requires: %{_cross_os}libtirpc
+Requires: %{_cross_os}libxcrypt
+
+%description
+%{summary}.
+
+%package devel
+Summary: Files for development using the tools for VMware
+Requires: %{name}
+
+%description devel
+%{summary}.
+
+%prep
+%autosetup -n open-vm-tools-%{version}-%{buildver} -p1
+
+%build
+autoreconf -fi
+%cross_configure \
+  --disable-deploypkg \
+  --disable-docs \
+  --disable-libappmonitor \
+  --disable-multimon \
+  --disable-resolutionkms \
+  --disable-servicediscovery \
+  --disable-tests \
+  --disable-vgauth \
+  --with-tirpc \
+  --with-udev-rules-dir=%{_cross_udevrulesdir} \
+  --without-dnet \
+  --without-gtk2 \
+  --without-gtk3 \
+  --without-gtkmm \
+  --without-gtkmm3 \
+  --without-icu \
+  --without-kernel-modules \
+  --without-pam \
+  --without-ssl \
+  --without-x \
+  --without-xerces \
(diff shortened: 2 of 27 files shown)


variants/Cargo.lock
```diff
@@ -559,6 +571,13 @@ dependencies = [
  "glibc",
 ]
 
+[[package]]
+name = "libz"
```
tjkirch, line 583: It seems odd that this shows up in the glib commit..?  I think this represents the addition of the libz package itself, not the dependency, so I'm confused.

bcressey:
(fix merge conflict with kernel 5.10 update)

webern (approved):

packages/libglib/libglib.spec
```diff
@@ -0,0 +1,66 @@
...
+Requires: %{_cross_os}libffi-devel
+
+%description devel
+%{summary}.
+
+%prep
+%autosetup -n glib-%{version} -p1
+
+%build
+CONFIGURE_OPTS=(
```
tjkirch, line 36: I wonder if it'd be worth having `-Dselinux=true` (rather than the default of auto) so it'd fail if we change something about our SELinux setup and don't realize the impact on glib2, rather than having it build without support.  Maybe also `-Dlibmount=<whether you want support or not, I don't know>`

packages/open-vm-tools/vmtoolsd.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=VMWare Tools service
+Documentation=https://github.com/vmware/open-vm-tools
+ConditionVirtualization=vmware
```
tjkirch, line 4: Would you please test that this isn't true on other platforms?  We've had some issues with platform detection in systemd :)

packages/open-vm-tools/open-vm-tools.spec
```diff
@@ -0,0 +1,106 @@
+%global buildver 17337674
+
+Name: %{_cross_os}open-vm-tools
+Version: 11.2.5
+Release: 1%{?dist}
+Summary: Tools for VMWare
```
tjkirch:
```suggestion
Summary: Tools for VMware
```
(and in summary below, and in service unit)

packages/open-vm-tools/open-vm-tools.spec
```diff
@@ -0,0 +1,106 @@
...
+
+%description devel
+%{summary}.
+
+%prep
+%autosetup -n open-vm-tools-%{version}-%{buildver} -p1
+
+%build
+autoreconf -fi
+%cross_configure \
```
tjkirch, line 38: nit: seems like `--disable-multimon` would also be pretty safe for our use case, given you have `--without-x` :)

bcressey:
Added configuration for vmtoolsd, so that restarting the VM from vCenter works as expected.

packages/open-vm-tools/open-vm-tools.spec
```diff
@@ -0,0 +1,117 @@
...
+%{_cross_factorydir}%{_cross_sysconfdir}/vmware-tools/tools.conf
+%{_cross_tmpfilesdir}/open-vm-tools.conf
+
+%{_cross_libdir}/*.so.*
+%dir %{_cross_libdir}/open-vm-tools
+%{_cross_libdir}/open-vm-tools/*
+%dir %{_cross_datadir}/open-vm-tools
+%{_cross_datadir}/open-vm-tools/*
+
+%exclude %{_bindir}/vm-support
```
tjkirch: nit: maybe exclude the whole dir, in case this happens again?  and put this down next to /sbin, with a helpful comment including an ASCII representation of skull-and-bones?  ;)

packages/open-vm-tools/tools.conf
```diff
@@ -0,0 +1,5 @@
+[powerops]
+poweron-script=/usr/bin/true
```
zmrow, line 2: I'm assuming if we set these to `/usr/bin/true` the corresponding actions in the webUI for shutdown, poweroff, etc. won't do anything.

I think for resume/suspend it makes sense not to do anything, but what about `poweron` and `poweroff`?

variants/Cargo.lock
bcressey: It wasn't added as part of the zlib commit, so it needed to come in at some point. I thought this was the most logical place to include it - the first time a package depended on it.

packages/open-vm-tools/tools.conf
bcressey: I verified that the actions work.

The default scripts are to run through a series of user or admin-defined commands before carrying out the power event. Since we don't support defining new ones, running `true` causes the event to be carried out immediately.

zmrow: Ah, gotcha.  I misunderstood that.

zmrow (approved):
:1st_place_medal:

bcressey:
Fixes per review feedback

packages/open-vm-tools/vmtoolsd.service
bcressey: Sure, although we'd never add this package to a non-vmware variant.

`aws-dev` with the package added, running under KVM:
```
bash-5.0# systemctl status vmtoolsd
● vmtoolsd.service - VMware Tools service
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/vmtoolsd.service; enabled; vendor preset: enabled)
     Active: inactive (dead)
  Condition: start condition failed at Fri 2021-04-30 13:54:15 UTC; 31s ago
             └─ ConditionVirtualization=vmware was not met
       Docs: https://github.com/vmware/open-vm-tools

Apr 30 13:54:15 ip-10-0-60-75.us-west-2.compute.internal systemd[1]: Condition check resulted in VMware Tools service being skipped.
```

`systemctl status` shows the overall state as "running" (green).

bcressey: Sure, although we'd never add this package to a non-vmware variant.

`aws-dev` with the package added, running under KVM:
```
bash-5.0# systemctl status vmtoolsd
● vmtoolsd.service - VMware Tools service
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/vmtoolsd.service; enabled; vendor preset: enabled)
     Active: inactive (dead)
  Condition: start condition failed at Fri 2021-04-30 13:54:15 UTC; 31s ago
             └─ ConditionVirtualization=vmware was not met
       Docs: https://github.com/vmware/open-vm-tools

Apr 30 13:54:15 ip-10-0-60-75.us-west-2.compute.internal systemd[1]: Condition check resulted in VMware Tools service being skipped.
```

`systemctl status` shows the overall state as "running" (green).

bcressey:
TODO: add `vmtoolsd` to the list of affected services.

tjkirch (approved):

bcressey:
(rebase with no changes to these commits)

tjkirch (approved):

zmrow (approved):
:gear:

Issue #1095, opened 2020-08-31
Tracking: transient CI failures


tjkirch:
This is a place to track CI failures that were not caused by code issues but by a transient issue in the CI runner.

Known issues:
* `failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: unexpected status code https://registry-1.docker.io/v2/docker/dockerfile/manifests/1.1.3-experimental: 500 Server Error`
* `failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: rpc error: code = Unknown desc = failed to build LLB: mount already exists with name`

webern:
`unexpected EOF`

https://github.com/bottlerocket-os/bottlerocket/pull/1006/checks?check_run_id=1058664646

etungsten:
> `unexpected EOF`
> 
> https://github.com/bottlerocket-os/bottlerocket/pull/1006/checks?check_run_id=1058664646

That's not the actual error. The actual error is grub failing to build due to the build container dying(?):
```
error: failed to run custom build command for `grub v0.1.0 (/home/ec2-user/_work/bottlerocket/bottlerocket/packages/grub)`
...
Error: No such container: buildsys-pkg-grub-aarch64-2fe95a9c2d67
Error: No such image: buildsys-pkg-grub-aarch64-2fe95a9c2d67
```

tjkirch:
> That's not the actual error

Heheh... sadly, neither is "no such container" or "no such image", which just mean that the grub build failed and so there was no output, and "failed to run custom build command" just means we need to look for a real error above.  Sadly, it looks like the check logs were lost when they were rerun.

etungsten:
CI runner ran into a random segfault during unit tests:

```
 + /home/ec2-user/_work/bottlerocket/bottlerocket/tools/docker-go --module-path /home/ec2-user/_work/bottlerocket/bottlerocket/sources/host-ctr --sdk-image bottlerocket/sdk-x86_64:v0.13.0-x86_64 --go-mod-cache /home/ec2-user/_work/bottlerocket/bottlerocket/.gomodcache --command 'cd cmd/host-ctr; go test -v'
# internal/cpu
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0xc048111f09 pc=0x440e50]
```

Full log:
<details>

```
[cargo-make] INFO - Running Task: unit-tests
[...]

PR #1545, opened 2021-04-30
retry builds on unexpected EOF error

bcressey:
**Issue number:**
#1095

**Description of changes:**
This is a rare failure that we see in CI. Retry it so the entire run is not marked as failing.

**Testing done:**
Modified a package locally to emit "unexpected EOF" and exit non-zero. Observed that the build was retried ten times before buildsys finally gave up.

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/builder.rs
+++ b/tools/buildsys/src/builder.rs
@@ -38,6 +38,9 @@ static DOCKER_BUILD_FRONTEND_ERROR: &str = concat!(
     r#"frontend grpc server closed unexpectedly"#
 );
 
+/* We also see sporadic CI failures with only this error message. */
+static UNEXPECTED_EOF_ERROR: &str = "unexpected EOF";
+
 static DOCKER_BUILD_MAX_ATTEMPTS: NonZeroU16 = nonzero!(10u16);
 
 pub(crate) struct PackageBuilder;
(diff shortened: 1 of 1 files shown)


zmrow (approved):
:partying_face:

arnaldo2792 (approved):
:+1:

tools/buildsys/src/builder.rs
```diff
@@ -38,6 +38,9 @@ static DOCKER_BUILD_FRONTEND_ERROR: &str = concat!(
     r#"frontend grpc server closed unexpectedly"#
 );
 
+/* We also see sporadic CI failures with only this error message. */
+static UNEXPECTED_EOF_ERROR: &str = "unexpected EOF";
```
tjkirch, line 42: I can live with this given the pain it's caused, but this seems like too common a phrase for me to be totally comfortable claiming that any build with this output can be retried.  We could be ignoring real errors, and if they're not consistent, we might miss them entirely.

One option to be a little more careful would be to modify buildsys to check for the error messages taking an entire line of output, not just appearing anywhere in the output.

tjkirch (approved):

webern (approved):

tools/buildsys/src/builder.rs
tjkirch: I did an alternate solution that lets us be a little more careful about error message positioning: https://github.com/bottlerocket-os/bottlerocket/pull/1557

bcressey:
@tjkirch carried this over the line in #1557.

PR #1547, opened 2021-04-30
aws-k8s: get cluster dns cidr from eks

### LINKED ISSUE #1197: DNS Cluster IP determination inconsistent with EKS AL2 AMIs
**Image I'm using:**
Latest BottleRocket EKS-optimized image (ami-01d975726ace73c6d in us-west-2)

The DNS IP that BottleRocket determines should be consistent with the IP address that the EKS-optimized AL2 determines. 

**What actually happened:**
The DNS IP is not consistent. The logic in the AL2 AMI here -> https://github.com/awslabs/amazon-eks-ami/commit/88648213079b4c0f70fe69c479724519327d7ae1 resolves: 
```bash
DNS_CLUSTER_IP=${SERVICE_IPV4_CIDR%.*}.10
```

The BottleRocket image resolves:
```rust
    const DEFAULT_10_RANGE_DNS_CLUSTER_IP: &str = "172.20.0.10";
    let dns = if mac_cidr_blocks.starts_with("10.") {
        DEFAULT_10_RANGE_DNS_CLUSTER_IP
    } else {
        DEFAULT_DNS_CLUSTER_IP
    } 
```

Run EKS-optimized AMIs on AL2 and BottleRocket with default settings on the same EKS cluster.

webern:
## Issue number:

Closes #1197

## Description of changes:

We previously assumed that an EKS cluster DNS IP was either 10.100.0.10 or 172.20.0.10. This may be incorrect when the cluster has a custom serviceIPv4CIDR setting.

Now we get the serviceIPv4CIDR from EKS describe-cluster so that we can calculate the correct cluster DNS IP in the presence of serviceIPv4CIDR.

Notes 
- this requires rusoto v0.46 because the serviceIPv4CIDR is fairly new to the API 
- rusoto v0.46 relies on tokio v1
- We needed to switch from `reqwest::blocking` to `reqwest` async for the IMDS calls because we need a runtime for rusoto.

#### Reproducing the issue before fixing it:

After reading [this](https://unofficial-kubernetes.readthedocs.io/en/latest/concepts/services-networking/dns-pod-service/)

- I created a cluster with a custom CIDR: `10.31.32.0/24`.
- I ran a production 1.18 AMI and verified that it had the wrong Cluster DNS: `10.100.0.10`.
- I ran a pod like this (there was some nslookup bug with busybox so I used ubuntu and installed dnsutils on the healthy pod. I couldn't apt-get install anything on the unhealthy pod which further proves DNS was broken. For the unhealthy pod I used busybox):
  - `kubectl run -i -t ubuntu --image=ubuntu --restart=Never --command=true -- bash`
- Inside the pod I ran `nslookup kubernetes.default`, which returned

```
/ # nslookup kubernetes.default
;; connection timed out; no servers could be reached
```

In a healthy, non-custom-CIDR cluster, the above produces:

```
root@ubuntu:/# nslookup kubernetes.default
Server:   10.100.0.10
Address:  10.100.0.10#53

Name: kubernetes.default.svc.cluster.local
Address: 10.100.0.1
```

Thus we have proven that Bottlerocket does not work correctly with a custom `serviceIPv4CIDR`.

#### Testing this PR

**Custom CIDR Cluster**

- I ran a node in a cluster with `serviceIPv4CIDR: '10.31.32.0/24'`
- I logged in and checked the Cluster DNS IP
  - `apiclient -u /settings?keys=settings.kubernetes.cluster-dns-ip`
  - `{"kubernetes":{"cluster-dns-ip":"10.31.32.10"}}`
- I ran a pod:
  - `kubectl run -i -t ubuntu --image=ubuntu --restart=Never --command=true -- bash`
  - `nslookup kubernetes.default`

The result shows that cluster DNS resolution is working:

```text
Server:   10.31.32.10
Address:  10.31.32.10#53

Name: kubernetes.default.svc.cluster.local
Address: 10.31.32.1
```

**Non-Custom CIDR Cluster**

- I ran a node in a cluster with no `serviceIPv4CIDR` configuration.
- I logged in and checked the Cluster DNS IP
  - `apiclient -u /settings?keys=settings.kubernetes.cluster-dns-ip`
  - `{"kubernetes":{"cluster-dns-ip":"10.100.0.10"}}`
- I ran a pod:
  - `kubectl run -i -t ubuntu --image=ubuntu --restart=Never --command=true -- bash`
  - `nslookup kubernetes.default`

The result shows that cluster DNS resolution is working:

```text
Server:   10.100.0.10
Address:  10.100.0.10#53

Name: kubernetes.default.svc.cluster.local
Address: 10.100.0.1
```

**Without Describe Cluster Permissions**

- I ran a node in a cluster with no `serviceIPv4CIDR` configuration and with DescribeCluster removed from the Node's instance role.
- I logged in and checked some tracing in the journal.
  - Saw this: `Unable to determine CIDR from EKS, falling back to default cluster DNS IP: Error describing cluster: Request ID: Some("9753370c-9443-455c-91b4-a182491bcf1e") Body: {"message":"User: arn:aws:sts::xxx:assumed-role/ [...] is not authorized to perform: eks:DescribeCluster on resource: [..."}`
  - `apiclient -u /settings?keys=settings.kubernetes.cluster-dns-ip`
  - `{"kubernetes":{"cluster-dns-ip":"10.100.0.10"}}`
- I ran a pod:
  - `kubectl run -i -t ubuntu --image=ubuntu --restart=Never --command=true -- bash`
  - `nslookup kubernetes.default`

The result shows that cluster DNS resolution is working despite failure to call the EKS API and describe-cluster:

```text
Server:   10.100.0.10
Address:  10.100.0.10#53

Name: kubernetes.default.svc.cluster.local
Address: 10.100.0.1
```

## Terms of contribution:

### DIFF
--- a/sources/api/pluto/src/api.rs
+++ b/sources/api/pluto/src/api.rs
@@ -1,0 +1,92 @@
+pub(super) use inner::{get_aws_k8s_info, Error};
+
+/// The result type for the [`api`] module.
+pub(super) type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Default)]
+pub(crate) struct AwsK8sInfo {
+    pub(crate) region: String,
+    pub(crate) cluster_name: String,
+}
+
+/// This code is the 'actual' implementation compiled when the `sources` workspace is being compiled
+/// for `aws-k8s-*` variants.
+// TODO - find a better way https://github.com/bottlerocket-os/bottlerocket/issues/1260
+#[cfg(aws_k8s_variant)]
+mod inner {
+    use super::*;
+    use snafu::{OptionExt, ResultExt, Snafu};
+
+    // FIXME Get these from configuration in the future
+    const DEFAULT_API_SOCKET: &str = "/run/api.sock";
+    const SETTINGS_URI: &str = "/settings";
+
+    #[derive(Debug, Snafu)]
+    pub(crate) enum Error {
+        #[snafu(display("Error calling Bottlerocket API: {}", source))]
+        ApiClient {
+            source: apiclient::Error,
+            uri: String,
+        },
+
+        #[snafu(display("The '{}' setting is missing", setting))]
+        Missing { setting: String },
+
+        #[snafu(display("Unable to deserialize Bottlerocket settings: {}", source))]
+        SettingsJson { source: serde_json::Error },
+    }
+
+    /// Gets the Bottlerocket settings from the API and deserializes them into a struct.
+    async fn get_settings() -> Result<model::Settings> {
+        let (_status, response_body) =
+            apiclient::raw_request(DEFAULT_API_SOCKET, SETTINGS_URI, "GET", None)
+                .await
+                .context(ApiClient { uri: SETTINGS_URI })?;
+
+        serde_json::from_str(&response_body).context(SettingsJson)
+    }
+
+    /// Gets the info that we need to know about the EKS cluster from the Bottlerocket API.
+    pub(crate) async fn get_aws_k8s_info() -> Result<AwsK8sInfo> {
+        let settings = get_settings().await?;
+        Ok(AwsK8sInfo {
+            region: settings
+                .aws
+                .context(Missing { setting: "aws" })?
+                .region
+                .context(Missing { setting: "region" })?
+                .into(),
+            cluster_name: settings
+                .kubernetes
+                .context(Missing {
+                    setting: "kubernetes",
+                })?
+                .cluster_name
+                .context(Missing {
+                    setting: "cluster-name",
+                })?
+                .into(),
+        })
+    }
+}
+
+/// This dummy code is compiled when the `sources` workspace is being compiled for non `aws-k8s-*`
+/// variants.
+// TODO - find a better way https://github.com/bottlerocket-os/bottlerocket/issues/1260
+#[cfg(not(aws_k8s_variant))]
+mod inner {
+    use super::*;
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    pub(crate) enum Error {
+        #[snafu(display(
+            "The get_aws_k8s_info function is only compatible with aws-k8s variants"
+        ))]
+        WrongVariant,
+    }
+
+    pub(crate) async fn get_aws_k8s_info() -> Result<AwsK8sInfo> {
+        WrongVariant.fail()
+    }
+}
--- a/sources/api/pluto/src/eks.rs
+++ b/sources/api/pluto/src/eks.rs
@@ -1,0 +1,48 @@
+use rusoto_core::region::ParseRegionError;
+use rusoto_core::{Region, RusotoError};
+use rusoto_eks::{DescribeClusterError, Eks, EksClient};
+use snafu::{OptionExt, ResultExt, Snafu};
+use std::str::FromStr;
+
+#[derive(Debug, Snafu)]
+pub(super) enum Error {
+    #[snafu(display("Error describing cluster: {}", source))]
+    DescribeCluster {
+        source: RusotoError<DescribeClusterError>,
+    },
+
+    #[snafu(display("Missing field '{}' EKS response", field))]
+    Missing { field: &'static str },
+
+    #[snafu(display("Unable to parse '{}' as a region: {}", region, source))]
+    RegionParse {
+        region: String,
+        source: ParseRegionError,
+    },
+}
+
+type Result<T> = std::result::Result<T, Error>;
+
+/// Returns the cluster's [serviceIPv4CIDR] DNS IP by calling the EKS API.
+/// (https://docs.aws.amazon.com/eks/latest/APIReference/API_KubernetesNetworkConfigRequest.html)
+pub(super) async fn get_cluster_cidr(region: &str, cluster: &str) -> Result<String> {
+    let parsed_region = Region::from_str(region).context(RegionParse { region })?;
+    let client = EksClient::new(parsed_region);
+    let describe_cluster = rusoto_eks::DescribeClusterRequest {
+        name: cluster.to_owned(),
+    };
+    client
+        .describe_cluster(describe_cluster)
+        .await
+        .context(DescribeCluster {})?
+        .cluster
+        .context(Missing { field: "cluster" })?
+        .kubernetes_network_config
+        .context(Missing {
+            field: "kubernetes_network_config",
+        })?
+        .service_ipv_4_cidr
+        .context(Missing {
+            field: "service_ipv_4_cidr",
+        })
+}
--- a/sources/api/pluto/src/main.rs
+++ b/sources/api/pluto/src/main.rs
@@ -5,21 +5,43 @@
 
 pluto is called by sundog to generate settings required by Kubernetes.
 This is done dynamically because we require access to dynamic networking
-setup information.
+and cluster setup information.
 
-It makes calls to IMDS to get meta data:
+It uses IMDS to get information such as:
 
-- Cluster DNS
+- Instance Type
 - Node IP
+
+It uses EKS to get information such as:
+
+- Service IPV4 CIDR
+
+It uses the Bottlerocket API to get information such as:
+
+- Kubernetes Cluster Name
+- AWS Region
+
+# Interface
+
+Pluto takes the name of the setting that it is to generate as its first
+argument.
+It returns the generated setting to stdout as a JSON document.
+Any other output is returned to stderr.
+
+Pluto returns a special exit code of 2 to inform `sundog` that a setting should be skipped. For
+example, if `max-pods` cannot be generated, we want `sundog` to skip it without failing since a
+reasonable default is available.
 */
 
-use reqwest::blocking::Client;
+mod api;
+mod eks;
+
+use reqwest::Client;
+use snafu::{ensure, OptionExt, ResultExt};
 use std::fs::File;
 use std::io::{BufRead, BufReader};
 use std::string::String;
 use std::{env, process};
-
-use snafu::{OptionExt, ResultExt};
 
 // This is the default DNS unless our CIDR block begins with "10."
 const DEFAULT_DNS_CLUSTER_IP: &str = "10.100.0.10";
@@ -40,6 +62,7 @@ const IMDS_INSTANCE_TYPE_ENDPOINT: &str =
 const ENI_MAX_PODS_PATH: &str = "/usr/share/eks/eni-max-pods";
 
 mod error {
+    use crate::eks;
     use snafu::Snafu;
 
     // Taken from sundog.
@@ -55,6 +78,9 @@ mod error {
     #[derive(Debug, Snafu)]
     #[snafu(visibility = "pub(super)")]
     pub(super) enum PlutoError {
+        #[snafu(display("Unable to parse CIDR '{}': {}", cidr, reason))]
+        CidrParse { cidr: String, reason: String },
+
         #[snafu(display("Error {}ing '{}': {}", method, uri, source))]
         ImdsRequest {
             method: String,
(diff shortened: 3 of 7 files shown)


webern:
> webern force-pushed the webern:dns branch from ad82670 to 99317fb

Add Bottlerocket API call to get the AWS region before working on the rebase to pull in #1551.

webern:
> webern force-pushed the webern:dns branch from 99317fb to b29a1cb

Rebase.

webern:
> webern force-pushed the webern:dns branch from b29a1cb to d6cfe90

Fix an oops (region URI was wrong)

webern:
Retest checks out.

sources/api/pluto/src/api.rs
```diff
@@ -0,0 +1,97 @@
...
+        apiclient::raw_request(DEFAULT_API_SOCKET, CLUSTER_NAME_URI, "GET", None)
+            .await
+            .context(ApiClient {
+                uri: CLUSTER_NAME_URI,
+            })?;
+    let parsed_response: serde_json::Value =
+        serde_json::from_str(&raw_response).context(ResponseJsonParse)?;
+
+    Ok(parsed_response
+        .as_object()
```
tjkirch: I think we could simplify by turning `get_cluster_name` and `get_region` into `get_settings` that returns a `model::Settings`; it'd be simpler and more reliable to use the result (one `.context` rather than one per layer, fewer error types, fewer settings calls/URIs, model-checked data) and we need both of the results at the same time anyway.  `schnauzer/src/lib.rs` has an example you could copy, though it has an extra `code.is_success` check that isn't needed because `raw_request` does that for you.  (We should move that to apiclient lib, schnauzer really has no business owning it, but that's a separate issue...)

sources/api/pluto/src/eks.rs
```diff
@@ -0,0 +1,50 @@
+use rusoto_core::region::ParseRegionError;
+use rusoto_core::{Region, RusotoError};
+use rusoto_eks::{DescribeClusterError, Eks as RusotoEks, EksClient};
```
tjkirch: nit: Is renaming the trait needed?  I'm not seeing a conflict, and we're not using it directly.

sources/api/pluto/src/main.rs
```diff
@@ -249,9 +339,25 @@ fn run() -> Result<()> {
...
-    if let Err(e) = run() {
+#[tokio::main]
+async fn main() {
+    if let Err(e) = run().await {
         eprintln!("{}", e);
         process::exit(1);
     }
 }
+
+#[test]
```
tjkirch, line 347: I think unless these have `#[cfg(test)]` they'd be built into production binaries; that's why there's typically a `#[cfg(test)] mod test {}` around them.

sources/api/pluto/src/eks.rs
```diff
@@ -0,0 +1,50 @@
...
+
+#[derive(Debug, Snafu)]
+pub(super) enum Error {
+    #[snafu(display("Error describing cluster: {}", source))]
+    DescribeCluster {
+        source: RusotoError<DescribeClusterError>,
+    },
+
+    #[snafu(display("Cluster object is missing from EKS response"))]
+    MissingCluster {},
```
tjkirch: Just a nitpick idea: a `Missing { missing: String, from: String }` might be a simpler way to represent these similar errors.

sources/api/pluto/src/main.rs
```diff
@@ -55,6 +80,12 @@ mod error {
     #[derive(Debug, Snafu)]
     #[snafu(visibility = "pub(super)")]
     pub(super) enum PlutoError {
+        #[snafu(display("Unable to parse CIDR '{}'", cidr))]
+        CidrParse { cidr: String },
```
tjkirch: nit: maybe a `reason: String` so you could mock a source, like `reason: "not 4 components"` in the case below?

sources/api/pluto/src/main.rs
```diff
@@ -5,22 +5,47 @@
...
+reasonable default is available.
 */
 
-use reqwest::blocking::Client;
+mod api;
+mod eks;
+
+use api::{get_cluster_name, get_region};
+use eks::get_cluster_cidr;
+use error::PlutoError;
```
tjkirch: nit: this one's a bit out of place because it's rescoping something defined below it in this file, not somewhere else like the others - maybe put it back before the `type Result`?

sources/api/pluto/src/main.rs
```diff
@@ -157,9 +192,62 @@ fn get_max_pods(client: &Client, session_token: &str) -> Result<String> {
     error::NoInstanceTypeMaxPods { instance_type }.fail()
 }
 
-fn get_cluster_dns_ip(client: &Client, session_token: &str) -> Result<String> {
+/// Returns the cluster's DNS IPV4 address. First it attempts to call EKS describe-cluster to find
+/// the `serviceIPv4CIDR`. If that works, it returns the first `*.10` address. If the EKS call is
```
tjkirch: I'm not sure I understand the "first `*.10` address" part of this comment - is it "10.*"?  If so, that only seems to apply if we fail to get the data from EKS and we're down on line 261, right?

sources/api/pluto/src/main.rs
```diff
@@ -157,9 +192,62 @@ fn get_max_pods(client: &Client, session_token: &str) -> Result<String> {
...
+/// not successful, it falls back to using IMDS MAC CIDR blocks to return one of two default
+/// addresses.
+async fn get_cluster_dns_ip(client: &Client, session_token: &str) -> Result<String> {
+    let region = get_region().await.context(error::Region)?;
+    let cluster_name = get_cluster_name().await.context(error::ClusterName {})?;
```
tjkirch: If we fail to get one of these, should we continue on to the default path instead of failing entirely?  It seems like that might be friendlier toward misconfiguration for the majority of cases that won't be using custom CIDR.

sources/api/pluto/src/main.rs
```diff
@@ -5,22 +5,47 @@
...
+example, if `max-pods` cannot be generated, we want `sundog` to skip it without failing since a
+reasonable default is available.
 */
 
-use reqwest::blocking::Client;
+mod api;
+mod eks;
+
+use api::{get_cluster_name, get_region};
+use eks::get_cluster_cidr;
```
tjkirch:
```suggestion
```
nit: I think it'd be clearer to use the whole `eks::get_cluster_cidr()` below, to make it clearer that it's not a local function, since there a few similarly named functions here.

sources/api/pluto/src/main.rs
```diff
@@ -157,9 +192,62 @@ fn get_max_pods(client: &Client, session_token: &str) -> Result<String> {
...
+        Err(e) => {
+            eprintln!("Unable to get CIDR from EKS, using default DNS IP: {}", e);
+            return None;
+        }
+    };
+    match get_dns_from_cidr(&cidr) {
+        Ok(dns_ip) => Some(dns_ip),
+        Err(e) => {
+            eprintln!("Unable to parse CIDR from EKS, using default DNS IP: {}", e);
+            None
```
tjkirch: nit: since the message and error handling are the same in the two steps of this function, it seems like it might be simpler combined:
```rust
async fn get_dns_from_eks(region: &str, cluster_name: &str) -> Option<String> { 
    get_cluster_cidr(region, cluster_name).await
        .and_then(|cidr| get_dns_from_cidr(cidr))
        .map_err(|e| eprintln!("Unable to parse CIDR from EKS, using default DNS IP: {}", e))
        .ok()
}
```

sources/api/pluto/src/main.rs
webern: Functions with `#[test]` imply `#[cfg(test)]` and are not included in non-test builds: https://stackoverflow.com/a/55995786/2779792

My personal style guide is to use `#[test]` unless some additional symbols (e.g. static values or helper functions) need to be defined.

webern:
> webern force-pushed the webern:dns branch from d6cfe90 to 0745e06 

I think this addresses all of @tjkirch 's suggestion _except_ for using the Model struct, which I will address separately.

webern:
> webern force-pushed the webern:dns branch from 0745e06 to 7325837

Use the API via strongly-typed settings instead of poking into a serde::json object. This requires conditional compilation to preserve the usability of the `sources` workspace.

sources/api/pluto/src/main.rs
```diff
@@ -121,22 +144,25 @@ mod error {
 
 use error::PlutoError;
 
-type Result<T> = std::result::Result<T, PlutoError>;
+pub(crate) type Result<T> = std::result::Result<T, PlutoError>;
```
webern: That visibility change is surely a mistake.

sources/api/pluto/src/main.rs
```diff
@@ -157,9 +183,68 @@ fn get_max_pods(client: &Client, session_token: &str) -> Result<String> {
...
+}
+
+/// Gets the Service IPV4 CIDR setting from EKS and parses it to calculate the cluster DNS IP.
+/// Prints the error and returns `None` if anything goes wrong.
+async fn get_dns_from_eks() -> Option<String> {
+    let k8s_info = match api::get_aws_k8s_info().await {
+        Ok(value) => value,
+        Err(e) => {
+            eprintln!(
+                "Unable to get region from Bottlerocket API, using default DNS IP: {}",
```
webern: Forgot to update this error message.

sources/api/pluto/src/main.rs
```diff
@@ -157,9 +183,68 @@ fn get_max_pods(client: &Client, session_token: &str) -> Result<String> {
...
+
+    // we were unable to obtain or parse the cidr range from eks, fallback to one of two default
+    // values based on the cidr range of our primary network interface
+    get_cluster_dns_from_imds_mac(client, session_token).await
+}
+
+/// Gets the Service IPV4 CIDR setting from EKS and parses it to calculate the cluster DNS IP.
+/// Prints the error and returns `None` if anything goes wrong.
+async fn get_dns_from_eks() -> Option<String> {
+    let k8s_info = match api::get_aws_k8s_info().await {
```
webern: Maybe `let aws_k8s_info = ` for consistency.

webern:
Retest checks out with 7325837. CI failure is a flake but not re-running because there are some small things to push.

webern:
> webern force-pushed the webern:dns branch from 7325837 to 38fa191

Fix some nits.

sources/api/pluto/src/api.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /// Gets the Bottlerocket settings from the API and deserializes them into a struct.
+    async fn get_settings() -> Result<model::Settings> {
+        let (_status, response_body) =
+            apiclient::raw_request(DEFAULT_API_SOCKET, SETTINGS_URI, "GET", None)
+                .await
+                .context(ApiClient { uri: SETTINGS_URI })?;
+
+        //ensure!(status.is_success(), ApiStatusCode);
```
tjkirch:
```suggestion
```

sources/api/pluto/src/api.rs
```diff
@@ -0,0 +1,101 @@
...
+        let settings: model::Settings =
+            serde_json::from_str(&response_body).context(SettingsJson)?;
+
+        Ok(settings)
```
tjkirch:
```suggestion
        serde_json::from_str(&response_body).context(SettingsJson)
```

sources/api/pluto/src/api.rs
```diff
@@ -0,0 +1,101 @@
...
+        ApiClient {
+            source: apiclient::Error,
+            uri: String,
+        },
+
+        #[snafu(display("AWS settings are missing"))]
+        AwsSettings {},
+
+        #[snafu(display("The 'aws.region' setting is missing"))]
+        AwsRegionMissing,
```
tjkirch: nit: a similar `Missing { missing: String }` could simplify this; the conditional modules are unfortunate and minimizing them seems helpful.  Could also pull the non-`fn` bits out to share..?

tjkirch (approved):

webern:
> webern force-pushed the webern:dns branch from 38fa191 to a821017

Additional cleanups.

etungsten (approved):

sources/api/pluto/src/api.rs
```diff
@@ -0,0 +1,92 @@
...
+            region: settings
+                .aws
+                .context(Missing { setting: "aws" })?
+                .region
+                .context(Missing { setting: "region" })?
+                .into(),
+            cluster_name: settings
+                .kubernetes
+                .context(Missing {
+                    setting: "kuberenetes",
```
tjkirch:
```suggestion
                    setting: "kubernetes",
```

webern:
> webern force-pushed the webern:dns branch from a821017 to 55ada33 

Spelling fix.

tjkirch (approved):

PR #1549, opened 2021-04-30
refactor package and variant dependencies

bcressey:
**Issue number:**
N/A

**Description of changes:**
This shifts some commonly used packages into the set installed by "release". `find`, `ps`, and `conntrack` will now be available on all variants, instead of some on `aws-k8s-*` and others on `aws-ecs-1`.

"host-ctr" and "containerd" are also promoted to the "release" package set. They end up installed regardless because of our host containers feature, and moving the dependency allows us to start building the "os" package sooner.

Most of the dependencies of Go packages have been eliminated, since those aren't needed at build time. The notable exception is the "runc" dependency on "libseccomp".

**Testing done:**
Built all variants locally.

Confirmed that `aws-k8s-1.18` and `aws-ecs-1` worked as expected:
* host containers came up
* orchestrated containers came up
* `find`, `ps`, and `conntrack` were present
* k8s build had the CNI binaries
* ecs build had Docker

**Terms of contribution:**

### DIFF
--- a/packages/cni-plugins/Cargo.toml
+++ b/packages/cni-plugins/Cargo.toml
@@ -18,4 +18,5 @@ glibc = { path = "../glibc" }
 
 # RPM Requires
 [dependencies]
-iptables = { path = "../iptables" }
+# `iptables` is only needed at runtime, and is pulled in by `release`.
+# iptables = { path = "../iptables" }
--- a/packages/cni/Cargo.toml
+++ b/packages/cni/Cargo.toml
@@ -18,4 +18,5 @@ glibc = { path = "../glibc" }
 
 # RPM Requires
 [dependencies]
-iptables = { path = "../iptables" }
+# `iptables` is only needed at runtime, and is pulled in by `release`.
+# iptables = { path = "../iptables" }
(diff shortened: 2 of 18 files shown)


tjkirch (approved):

jpculp (approved):

packages/docker-engine/docker-engine.spec
```diff
@@ -29,11 +29,8 @@ BuildRequires: git
 BuildRequires: %{_cross_os}glibc-devel
 BuildRequires: %{_cross_os}libseccomp-devel
 BuildRequires: %{_cross_os}systemd-devel
-Requires: %{_cross_os}containerd
```
webern: I chose this line to ask a general question. Are all of these `Requires:` removals actually removing something that isn't required? In this example, isn't containerd actually required for Docker to work?

If that's the case (i.e. if some of the `Requires:` are _actually_ required), would it be better to leave them in the spec, comment them out in the Cargo.toml, and note in the Cargo.toml that they are commented out for dependency graph optimization?

If you are trying to compose a new variant, you would get an error message from the build (with true spec `Requires:` statements) rather than a missing dependency at runtime (i.e. if your new variant's dependency graph happened to not build all the required packages).

bcressey: I've reworked the commits to do as you suggest - keep the `Requires:` while commenting out the dependency.

bcressey:
The force-push above adds back the `Requires`.

bcressey:
The force-push above fixes the conflicts by rebasing, and is probably less useful than just looking at the full diff again.

webern (approved):

tjkirch (approved):

PR #1550, opened 2021-04-30
Fixed a misleading typo for setting Labels/Taints

ellistarn:
**Issue number:**
N/A

**Description of changes:**
Labels and taints do not work if the keys aren't wrapped in strings. My nodes never connect to the cluster. There may be other places in this codebase where this issue exists.

**Testing done:**
N/A

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ You should [specify them in user data](#using-user-data).
 * `settings.kubernetes.cluster-certificate`: This is the base64-encoded certificate authority of the cluster.
 * `settings.kubernetes.api-server`: This is the cluster's Kubernetes API endpoint.
 
-The following settings can be optionally set to customize the node labels and taints.
+The following settings can be optionally set to customize the node labels and taints. Remember to quote keys (since they often contain ".") and to quote all values.
 * `settings.kubernetes.node-labels`: [Labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) in the form of key, value pairs added when registering the node in the cluster.
 * `settings.kubernetes.node-taints`: [Taints](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/) in the form of key, value and effect entries added when registering the node in the cluster.
   * Example user data for setting up labels and taints:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -299,11 +299,11 @@ The following settings can be optionally set to customize the node labels and ta
...
     [settings.kubernetes.node-labels]
-    label1 = "foo"
-    label2 = "bar"
+    "label1" = "foo"
+    "label2" = "bar"
     [settings.kubernetes.node-taints]
-    dedicated = "experimental:PreferNoSchedule"
-    special = "true:NoSchedule"
+    "dedicated" = "experimental:PreferNoSchedule"
+    "special" = "true:NoSchedule"
```
bcressey, line 306: In some of the other settings descriptions, we have a note like this:
```
Remember to quote <X> (since they often contain ".") and to quote all values.
```

It might be useful to add that to the end of the `node-labels` and `node-taints` items, just to clarify why the quotes show up in the example.

ellistarn: WDYT about keeping quote in all documentation and recommend users use them in all cases. You probably need to support the unquoted variant for backwards compat, but maybe best to leave that undocumented as undesired behavior.

ellistarn: i.e., removing the need for any reminder of adding the quotes and simply show them in all examples.

tjkirch: Always using quotes would go against the spirit of the TOML standard, where quotes are for special cases; we try to keep things simple for the majority of cases that don't need quoting.

ellistarn: I'm all about following TOML standards, but the problem is that 99% of kubernetes labels and taints are domains, and include special characters, so users (like myself) will trip over this.

tjkirch: I think we agree that quotes are good in this case :)  I thought you were saying to quote keys for all TOML examples in the rest of the documentation, too.  We use TOML in many places, and very little of it requires quoting.

ellistarn: We're on the same page :). Any other spots I might've missed?

tjkirch: I don't think so!  Adding the comment above would help for consistency with other areas where we need quoting, but other than that I think we're good.

ellistarn: image

tjkirch (approved):
Thank you!

etungsten (approved):

bcressey (approved):

bcressey:
(@tjkirch to merge this)

PR #1576, opened 2021-05-14
early-boot-config: Simplify conditional compilation and make `local_file` available to all platforms

### LINKED ISSUE #1457: `early-boot-config`:  Use local file in sequence with other user data retrieval methods
Currently, using a local file (`/etc/early-boot-config/user-data`) for user data is supported in `aws-dev`.  It is an implementer of the `PlatformDataProvider` trait.  In `main()`, we decide to use either local file _or_ another provider because we expect a `PlatformDataProvider` trait object.

We should be able to use local file _and_ another provider's methods.  In order to do this, I think we should move local file support to a shared function that returns `Result<Option<SettingsJson>>` so it can be used in sequence with other methods.

zmrow:
**Issue number:**
Fixes #1457 

**Description of changes:**
This change simplifies the conditional compilation by re-exporting the appropriate `PlatformDataProvider` as the common name `Platform`.  This allows us to remove all conditional bits from `main` by using `Platform` when retrieving user data.

It also makes `local_file` a common function that all platforms can use, rather than a "platform-like" module in itself.  Currently, we only make use of it in the `aws-dev` variant.
``` 
Previously, `local_file` was an implementor of the
`PlatformDataProvider` trait.  This was a bit awkward, considering local
file is _not_ a platform.  It also meant it was a bit awkward to use
inside another provider.  This change makes `local_file` a function that
all platforms can use and makes use of it inside the `aws` module for
`aws-dev` variants.
```

```
This change uses `cfg` directives inside of `provider.rs` to compile the
appropriate `PlatformDataProvider` for the current variant.  It
re-exports the data provider as `Platform`, meaning that all conditional
compilation details are removed from `main.rs` and it can simply use
`Platform` instead of a specific platform trait implementor.
```

**Testing done:**
* Boot an `aws-dev` variant with and without a local file to ensure local file is read if it exists
* Boot an `aws-k8s-1.19` variant to ensure user data is read properly (and local file isn't attempted).
* Boot a `vmware-k8s-1.20` variant to ensure nothing broke there.

**Terms of contribution:**

### DIFF
--- a/sources/api/early-boot-config/src/provider.rs
+++ b/sources/api/early-boot-config/src/provider.rs
@@ -3,14 +3,18 @@
 use crate::settings::SettingsJson;
 use async_trait::async_trait;
 
-#[cfg(any(bottlerocket_platform = "aws", bottlerocket_platform = "aws-dev"))]
-pub(crate) mod aws;
-
 #[cfg(bottlerocket_platform = "aws-dev")]
-pub(crate) mod local_file;
+mod local_file;
+
+#[cfg(any(bottlerocket_platform = "aws", bottlerocket_platform = "aws-dev"))]
+mod aws;
+#[cfg(any(bottlerocket_platform = "aws", bottlerocket_platform = "aws-dev"))]
+pub(crate) use aws::AwsDataProvider as Platform;
 
 #[cfg(bottlerocket_platform = "vmware")]
-pub(crate) mod vmware;
+mod vmware;
+#[cfg(bottlerocket_platform = "vmware")]
+pub(crate) use vmware::VmwareDataProvider as Platform;
 
 /// Support for new platforms can be added by implementing this trait.
 #[async_trait]
--- a/sources/api/early-boot-config/src/provider/aws.rs
+++ b/sources/api/early-boot-config/src/provider/aws.rs
@@ -9,6 +9,9 @@ use snafu::{OptionExt, ResultExt};
 use std::fs;
 use std::path::Path;
 
+#[cfg(bottlerocket_platform = "aws-dev")]
+use crate::provider::local_file::{local_file_user_data, USER_DATA_FILE};
+
 /// Unit struct for AWS so we can implement the PlatformDataProvider trait.
 pub(crate) struct AwsDataProvider;
 
@@ -82,7 +85,16 @@ impl PlatformDataProvider for AwsDataProvider {
 
         let mut client = ImdsClient::new().await.context(error::ImdsClient)?;
 
-        // Instance identity doc first, so the user has a chance to override
+        // Attempt to read from local file first on the `aws-dev` variant
+        #[cfg(bottlerocket_platform = "aws-dev")]
+        {
+            match local_file_user_data()? {
+                None => warn!("No user data found via local file: {}", USER_DATA_FILE),
+                Some(s) => output.push(s),
+            }
+        }
+
+        // Instance identity doc next, so the user has a chance to override
         match Self::identity_document(&mut client).await? {
             None => warn!("No instance identity document found."),
             Some(s) => output.push(s),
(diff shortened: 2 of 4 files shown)


sources/api/early-boot-config/src/provider.rs
```diff
@@ -2,14 +2,18 @@
 
 use crate::settings::SettingsJson;
 
-#[cfg(any(bottlerocket_platform = "aws", bottlerocket_platform = "aws-dev"))]
-pub(crate) mod aws;
-
 #[cfg(bottlerocket_platform = "aws-dev")]
-pub(crate) mod local_file;
+mod local_file;
```
tjkirch, line 7: I'm confused as to why local_file still has the cfg directive, since the PR says "It also makes local_file a common function that all platforms can use" - wouldn't we remove all conditionals on it, and let the data providers / platforms decide whether to call the function?

sources/api/early-boot-config/src/provider/aws.rs
```diff
@@ -186,7 +189,16 @@ impl PlatformDataProvider for AwsDataProvider {
 
         let session_token = Self::fetch_imds_session_token(&client)?;
 
-        // Instance identity doc first, so the user has a chance to override
+        // Attempt to read from local file first on the `aws-dev` variant
+        #[cfg(bottlerocket_platform = "aws-dev")]
+        {
+            match local_file_user_data()? {
+                Some(s) => output.push(s),
+                None => warn!("No user data found via local file: {}", USER_DATA_FILE),
```
tjkirch, line 92: I think not having a user data file is the normal case - is it worth a warning-level message?

tjkirch (approved):

sources/api/early-boot-config/src/provider/aws.rs
zmrow: I think for variants that have the ability to use a local file it's worth having the message so users know that nothing was found there, similar to what we do for other sources.

sources/api/early-boot-config/src/provider.rs
zmrow: We could for sure do that, but when compiling variants that don't use the code in the `local_file` module, we're going to get "unused" warnings like:
```
warning: constant is never used: `USER_DATA_FILE`
...
warning: function is never used: `local_file_user_data`
```

zmrow: I'm not really sure how we get around it other than hiding the whole module behind a `cfg` directive like I've done here... perhaps someone knows a trick I don't know?

sources/api/early-boot-config/src/provider/aws.rs
etungsten: Can we have it as a info level message? I guess that would mean the message won't go to the console?

etungsten (approved):

sources/api/early-boot-config/src/provider.rs
webern: You could just silence the warning with `#[allow(dead_code)]` on the function that is sometimes unused. (need to look up the correct string there)

zmrow: Hm - I suppose we could do that but I like it a bit less.  Are those really the only two options here? :thinking:

zmrow:
^ Rebases on `develop` to fix the merge conflicts.

:( Sorry the rebase wasn't pushed separately.

sources/api/early-boot-config/src/provider.rs
zmrow: [This (very good) article](https://bitshifter.github.io/2020/05/07/conditional-compilation-in-rust/) suggests the pattern I'm using: putting `cfg` directives on module imports to keep the actual code clean and not ignore any valid warnings.

IMHO let's keep the code as is unless we come up with something better.

webern (approved):

sources/api/early-boot-config/src/provider/aws.rs
zmrow: I'll keep it as `warn!` for now to stay consistent with the rest of the program and ensure it outputs to the console as expected.

Issue #1385, opened 2021-03-11
host-containers fails when any expected field is missing


samuelkarp:
**Image I'm using:**
```
bash-5.0# cat /etc/os-release
NAME=Bottlerocket
ID=bottlerocket
PRETTY_NAME="Bottlerocket OS 1.0.6"
VARIANT_ID=aws-ecs-1
VERSION_ID=1.0.6
BUILD_ID=ee56d1fc
```

**What I expected to happen:**
When I reboot a working Bottlerocket instance in a non-degraded state, it comes back up in non-degraded state.

**What actually happened:**
Bottlerocket booted in a degraded state where the `settings-applier.service` unit failed, causing `configured.target` to fail and any other units depending on it to fail.

**How to reproduce the problem:**

1. Make a typo when adding a host-container setting such that you have an unintended container entry like "admiin"
2. Ensure that the "admiin" container is missing at least one required field like "source" or "superpowered"
3. Ensure that the "admiin" container is disabled
4. Reboot

**Possible approaches to fix this:**

1. The settings API should enable us to delete mistaken host-container entries
2. The model should enforce that all required fields are actually present when creating host-container settings
3. The `host-containers` binary should tolerate missing fields, especially when a given host-container entry is not enabled

webern:
> it comes back up in non-degraded state.

I think that's a typo.

tjkirch:
I think we probably want to take all 3 approaches, to some degree.

For (1) we should be able to delete settings from a map, yeah.  We need to figure out what happens when that's done in "bad" situations, like trying to delete just _one_ of the settings needed for a host container, i.e. this issue ;)

For (2) we should probably have some ability to link or verify settings together somehow, though today's serialization/deserialization treats structs as just levels, not groupings, so it'll need some thought.

For (3) we should consider where our services are missing some needed tolerance for errors.  host-containers could treat missing "enabled" as false, for example, but should probably still fail if the container definition is missing a source.  ("fail" is up for debate; it could continue with other containers, or stop; it could fail the systemd service, or not; etc.)

bcressey:
We should probably have reasonable defaults for host containers - "enabled: false" and "superpowered: false" both seem sensible.

etungsten:
> I think we probably want to take all 3 approaches, to some degree.
> 
[...]

PR #1580, opened 2021-05-18
host-containers: default "enabled" and "superpowered" to false

etungsten:
**Issue number:**
Partially addresses https://github.com/bottlerocket-os/bottlerocket/issues/1385

**Description of changes:**
```
Date:   Mon May 17 16:37:52 2021 -0700

    host-containers: default "enabled" and "superpowered" to false
    
    Defaults "enabled" and "superpowered" to false when not defined.

```

**Testing done:**
Launched instance with the following userdata.
`self-terminator` doesn't have either `enabled` or `superpowered` specified
```
enabled = true

[settings.host-containers.control]
enabled = true

[settings.host-containers.self-terminator]
source="snip.dkr.ecr.us-west-2.amazonaws.com/self-terminator:latest"
```

The host was able to come up fine and the settings are as expected:
```
   "host-containers":{
      ...,
      "self-terminator":{
         "source":"snip.dkr.ecr.us-west-2.amazonaws.com/self-terminator:latest"
      }
   },
```

Then checking the host-container's environment variables, both `enabled` and `superpowered` are defaulted to false.
```
bash-5.0# cat /etc/host-containers/self-terminator.env 
CTR_SUPERPOWERED=false
CTR_SOURCE=snip.dkr.ecr.us-west-2.amazonaws.com/self-terminator:latest

# Just for reference; service is enabled or disabled by host-containers service
# CTR_ENABLED=false

```

**Terms of contribution:**

### DIFF
--- a/sources/api/host-containers/src/main.rs
+++ b/sources/api/host-containers/src/main.rs
@@ -363,14 +363,8 @@ where
         name,
         field: "source",
     })?;
-    let enabled = image_details.enabled.context(error::MissingField {
-        name,
-        field: "enabled",
-    })?;
-    let superpowered = image_details.superpowered.context(error::MissingField {
-        name,
-        field: "superpowered",
-    })?;
+    let enabled = image_details.enabled.unwrap_or(false);
+    let superpowered = image_details.superpowered.unwrap_or(false);
 
     info!("Host container '{}' is enabled: {}, superpowered: {}, with source: {}",
           name, enabled, superpowered, source);


webern (approved):

bcressey (approved):

sources/api/host-containers/src/main.rs
```diff
@@ -363,14 +363,8 @@ where
...
     })?;
-    let enabled = image_details.enabled.context(error::MissingField {
-        name,
-        field: "enabled",
-    })?;
-    let superpowered = image_details.superpowered.context(error::MissingField {
-        name,
-        field: "superpowered",
-    })?;
+    let enabled = image_details.enabled.unwrap_or(false);
```
tjkirch, line 366: We talked a bit offline - it seems like we might be able to use serde default= annotations on the fields in the model, instead of doing it here in code.  That would mean that the defaults (false) show up when you query settings, rather than leaving users in suspense and having to check docs.

tjkirch: After some epic testing, it seems the serde default is applying too strongly and overwriting things from our defaults.toml or even user data.  We can figure that out separately someday!

tjkirch (approved):

PR #1589, opened 2021-05-20
kubelet: add setting for configuring containerLogMaxFiles and MaxSize

samjo-nyang:
**Description of changes:**
Add containerLogMaxFiles and containerLogMaxSize to kubelet configuration.
FYI: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

**Testing done:**
This patch is currently use for our infrastructure.

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.16/kubelet-config
+++ b/packages/kubernetes-1.16/kubelet-config
@@ -84,3 +84,9 @@ tlsCipherSuites:
 - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
 maxPods: {{default 110 settings.kubernetes.max-pods}}
 staticPodPath: "/etc/kubernetes/static-pods/"
+{{~#if settings.kubernetes.container-log-max-size includeZero=true}}
+containerLogMaxSize: {{settings.kubernetes.container-log-max-size}}
+{{~/if}}
+{{~#if settings.kubernetes.container-log-max-files includeZero=true}}
+containerLogMaxFiles: {{settings.kubernetes.container-log-max-files}}
+{{~/if}}
--- a/README.md
+++ b/README.md
@@ -346,6 +346,8 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.event-burst`: The maximum size of a burst of event creations.
 * `settings.kubernetes.kube-api-qps`: The QPS to use while talking with kubernetes apiserver.
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
+* `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
+* `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
 
 You can also optionally specify static pods for your node with the following settings.
 Static pods can be particularly useful when running in standalone mode.
--- a/Release.toml
+++ b/Release.toml
@@ -48,3 +48,6 @@ version = "1.1.1"
     "migrate_v1.1.0_kubelet-kube-api-qps-kube-api-burst.lz4",
 ]
 "(1.1.0, 1.1.1)" = []
+"(1.1.1, 1.1.2)" = [
+    "migrate_v1.1.2_kubelet-container-log.lz4",
+]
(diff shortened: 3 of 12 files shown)


jpculp:
@samjo-nyang, thank you for your contribution! It looks great! In order to facilitate proper upgrade/downgrade between Bottlerocket versions, changes to settings like this one require something known as a migration. Migrations remove new settings when a user downgrades to a previous version of Bottlerocket. I encourage you to take a look at [this recent PR](https://github.com/bottlerocket-os/bottlerocket/pull/1494) that adds a similar setting. The [second commit](https://github.com/bottlerocket-os/bottlerocket/pull/1494/commits/104122a47c8608d98b02d668dfb431a919292008) in particular provides the migration. Of course, if you'd rather, we can take care of this for you. Just let us know 🚀

samjo-nyang:
@jahkeup Thanks for your kind feedback! I added the migration files to v1.1.2; is this correct?

jpculp:
@samjo-nyang, code wise this all looks right to me 😃!  I might recommend cleaning up the commits a little bit by either adding more detail to the first commit or removing the word config. It would also be great if you could combine the last 2 to a single migration commit, but we can do that on our end as well. If you wanted to keep the commit messages consistent with our previous kublet settings, you can use the previous PR I mentioned as a reference or search the git log for commits with the prefixes `kubelet:` or `migrations:`. Thank you again for your contribution!

samjo-nyang:
@jpculp I just rebase & reword the commits

jpculp (approved):
Thanks @samjo-nyang! 🚀

bcressey (approved):

packages/kubernetes-1.16/kubelet-config
```diff
@@ -84,3 +84,5 @@ tlsCipherSuites:
 - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
 maxPods: {{default 110 settings.kubernetes.max-pods}}
 staticPodPath: "/etc/kubernetes/static-pods/"
+containerLogMaxSize: {{default "10Mi" settings.kubernetes.container-log-max-size}}
```
webern: Thank you for submitting this PR. I think we should utilize kubelet's default instead of hard-coding what is currently kubelet's default. That way, if Kubernetes changes these values in the future, we would not inadvertently be hard-coding an old default value.

This suggestion would obviously be repeated for all of the k8s variants.

```suggestion
{{~#if settings.kubernetes.container-log-max-size  includeZero=true}}
containerLogMaxSize: {{settings.kubernetes.container-log-max-size}}
{{~/if}}
```

packages/kubernetes-1.16/kubelet-config
```diff
@@ -84,3 +84,5 @@ tlsCipherSuites:
...
+containerLogMaxFiles: {{default 5 settings.kubernetes.container-log-max-files}}
```
webern: Same here:

```suggestion
{{~#if settings.kubernetes.container-log-max-files  includeZero=true}}
containerLogMaxFiles: {{settings.kubernetes.container-log-max-files}}
{{~/if}}
```

webern (changes requested):

webern:
@samjo-nyang, thank you! Did you have a chance to run it with these changes? If not, one of us will check it today and get this merged.

samjo-nyang:
I had no time to run with these exact changes; I'm glad if the bottlerocket team test this pr. Thanks

webern:
Thanks again @samjo-nyang 
I tested the latest push, described below, and it's good.
Merging 🚀 🚀 🚀 

#### Testing of aa6d1aea

I ran an instance without these settings and observed that they were not in the kubelet config.

I logged in and ran:

```sh
apiclient set settings.kubernetes.container-log-max-size=100M
```

And saw that this was added to the kubelet config:

```yaml
containerLogMaxSize: 100M
```

I ran this:

```sh
apiclient set settings.kubernetes.container-log-max-files=13
```

And saw that this was added to the kubelet config

```yaml
containerLogMaxFiles: 13
```

I ran a pod, it worked.

Next, I ran an instance with this in its userdata

```
container-log-max-size = "101M"
container-log-max-files = 33
```

Ran a pod, it worked. Logged in and checked the config and saw:

```yaml
containerLogMaxSize: 101M
containerLogMaxFiles: 33
```

webern (approved):

PR #1598, opened 2021-05-25
imdsclient: return options for all IMDS fetch functions

jpculp:
**Issue number:**

N/A

**Description of changes:**

This commit changes the return type for the IMDS fetch functions to
options returning None when a 404 is encountered. This makes it easier
for the caller to decide what to do when a 404 occurs and is closer to
the behavior before imdsclient was written.

`fetch_identity_document` and its associated `IdentityDocument` struct
were removed in favor of separate `fetch_region` and
`fetch_instance_type` functions. Instance-type is now fetched via IMDS
meta-data as opposed to the identity document. We are returning to the
original behavior as it is possible that the identity document can be
absent in certain situations.

Changes were also made to shibaken, pluto, and early-boot-config to
accommodate the options detailed above.

**Testing done:**

- Built `aws-k8s-1.19` ami and launched instance.

**Other testing:**
- [x] `aws-ecs-1` with no user data set.
- [x] `aws-dev` with no user data set. _(tested by @arnaldo2792)_
- [x] `vmware-k8s-1.20` with user data from guestinfo interface.

**Terms of contribution:**

### DIFF
--- a/sources/api/shibaken/src/main.rs
+++ b/sources/api/shibaken/src/main.rs
@@ -43,10 +43,12 @@ impl UserData {
 async fn fetch_public_keys_from_imds() -> Result<Vec<String>> {
     info!("Connecting to IMDS");
     let mut client = ImdsClient::new().await.context(error::ImdsClient)?;
-    client
+    let public_keys = client
         .fetch_public_ssh_keys()
         .await
-        .context(error::ImdsClient)
+        .context(error::ImdsClient)?
+        .unwrap_or_else(Vec::new);
+    Ok(public_keys)
 }
 
 /// Store the args we receive on the command line.
--- a/sources/imdsclient/src/lib.rs
+++ b/sources/imdsclient/src/lib.rs
@@ -11,8 +11,8 @@ This method is useful for specifying things like a pinned date for the IMDS sche
 use http::StatusCode;
 use log::{debug, info, trace, warn};
 use reqwest::Client;
-use serde::{Deserialize, Serialize};
-use snafu::{ensure, ResultExt};
+use serde_json::Value;
+use snafu::{ensure, OptionExt, ResultExt};
 use std::time::Duration;
 use tokio::time;
 
@@ -28,25 +28,6 @@ pub struct ImdsClient {
     client: Client,
     imds_base_uri: String,
     session_token: String,
-}
-
-/// This is the return type when querying for the IMDS identity document, which contains information
-/// such as region and instance_type. We only include the fields that we are using in Bottlerocket.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
-pub struct IdentityDocument {
-    region: String,
-    instance_type: String,
-}
-
-impl IdentityDocument {
-    pub fn region(&self) -> &str {
-        self.region.as_str()
-    }
-
-    pub fn instance_type(&self) -> &str {
-        self.instance_type.as_str()
-    }
 }
 
 impl ImdsClient {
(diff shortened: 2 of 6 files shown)


zmrow:
Could we also boot variants without user data to check that behavior as well?

sources/imdsclient/src/lib.rs
```diff
@@ -30,25 +30,6 @@ pub struct ImdsClient {
     session_token: String,
 }
 
-/// This is the return type when querying for the IMDS identity document, which contains information
```
zmrow, line 33: Could you elaborate on why this struct was removed?  I don't see anything about it in the description.

I do like that a user doesn't have to call an extra method to get the fields though.  Was that the rationale?

webern: - Both caller sites are using one-or-the-other of the fields.
- instance-identity-document can be absent under certain circumstances.
- Before the refactor instance type was coming from meta-data and not instance-identity-document.

Adding this all up, it seemed better to go back to the behavior that was present before the refactor. The other alternative is to return an `Option<InstanceIdentityDocument>` which has `Option<String>` fields which seems slightly less convenient and could fail to get the instance_type in cases that we previously didn't fail.  🤷 

So we get_region from metadata (which is where it was coming from before) and kept get_instance_type as a call into instance-identity-document.

zmrow: Sounds good to me!  Thanks for that.

jpculp: I'm going to condense what @webern said above into the commit message as well as test additional variants without user-data. This is a little light on detail and testing (hence posted as draft) in its current form, but I wanted to make it available for eyes sooner rather than later.

sources/api/shibaken/src/main.rs
```diff
@@ -43,10 +43,14 @@ impl UserData {
...
-    client
+    match client
         .fetch_public_ssh_keys()
         .await
-        .context(error::ImdsClient)
+        .context(error::ImdsClient)?
+    {
+        Some(public_keys) => Ok(public_keys),
+        None => Ok(Vec::new()),
+    }
```
webern: This should work and do the same thing (without the `match`):

```suggestion
        .context(error::ImdsClient)?
        .unwrap_or_else(|| Vec::new());
```

jpculp:
- Small tweak in `shibaken` based on @webern's feedback.
- Added more details to commit message and PR description.

sources/api/shibaken/src/main.rs
jpculp: 👍, but clippy preferred `.unwrap_or_else(Vec::new)`

jpculp:
Fixed typo in commit message.

webern (approved):

arnaldo2792:
I tested the change and it works:

```
[  OK  ] Started docker engine.
[  OK  ] Reached target Multi-User System.
[  OK  ] Finished Isolates multi-user.target.
Bottlerocket OS 1.1.1
Kernel 5.10.29 on an x86_64 (ttyS0)
ip-172-31-61-1 login:
```

sources/imdsclient/src/lib.rs
```diff
@@ -65,57 +46,74 @@ impl ImdsClient {
...
-            serde_json::from_slice(&response).context(error::Serde)?;
-        Ok(identity_document)
+        let response = match self.fetch_bytes(target).await? {
+            Some(response) => response,
+            None => Vec::new(),
+        };
+        let identity_document: Value = serde_json::from_slice(&response).context(error::Serde)?;
+        let region = identity_document
+            .get("region")
+            .and_then(|value| value.as_str())
```
jpculp, line 63: This seems extraneous, but without it the region is returned in quotes, causing the following error when running `pluto cluster-dns-ip`:
```
Unable to parse CIDR from EKS, using default DNS IP: Unable to parse '"us-west-2"' as a region: Not a valid AWS region: "us-west-2"
```

zmrow: I'm a little confused.  You're saying that you need to do the `value.as_str()` bit ?

sources/imdsclient/src/lib.rs
```diff
@@ -65,57 +46,74 @@ impl ImdsClient {
...
+    /// Returns the region described in the identity document.
+    pub async fn fetch_region(&mut self) -> Result<Option<String>> {
         let target = "dynamic/instance-identity/document";
-        let response = self.fetch_bytes(target).await?;
-        let identity_document: IdentityDocument =
-            serde_json::from_slice(&response).context(error::Serde)?;
-        Ok(identity_document)
+        let response = match self.fetch_bytes(target).await? {
+            Some(response) => response,
+            None => Vec::new(),
```
zmrow: Instead of setting `response` to an empty Vec here and then continuing to operate on it, we should just return from the function since we know we didn't find anything.

sources/imdsclient/src/lib.rs
```diff
@@ -65,57 +46,74 @@ impl ImdsClient {
...
     /// Returns the list of network interface mac addresses.
-    pub async fn fetch_mac_addresses(&mut self) -> Result<Vec<String>> {
+    pub async fn fetch_mac_addresses(&mut self) -> Result<Option<Vec<String>>> {
         let macs_target = "meta-data/network/interfaces/macs";
-        let macs = self.fetch_string(&macs_target).await?;
-        Ok(macs.split('\n').map(|s| s.to_string()).collect())
+        let macs = self
+            .fetch_string(&macs_target)
+            .await?
+            .map(|macs| macs.split('\n').map(|s| s.to_string()).collect());
```
zmrow: Nit: We could use `.lines()` here instead of `.split('\n')`.

sources/imdsclient/src/lib.rs
```diff
@@ -65,57 +46,74 @@ impl ImdsClient {
...
         let mac_cidr_blocks_target = format!(
             "meta-data/network/interfaces/macs/{}/vpc-ipv4-cidr-blocks",
             mac
         );
-        let cidr_blocks = self.fetch_string(&mac_cidr_blocks_target).await?;
-        Ok(cidr_blocks.split('\n').map(|s| s.to_string()).collect())
+        let cidr_blocks = self
+            .fetch_string(&mac_cidr_blocks_target)
+            .await?
+            .map(|cidr_blocks| cidr_blocks.split('\n').map(|s| s.to_string()).collect());
```
zmrow: Same here: `lines()`

sources/imdsclient/src/lib.rs
```diff
@@ -132,7 +130,10 @@ impl ImdsClient {
                 &public_key_targets.len()
             );
 
-            let public_key_text = self.fetch_string(&target).await?;
+            let public_key_text = match self.fetch_string(&target).await? {
```
zmrow: Any reason we can't use `.await?.context(error::KeyNotFound...)?` here?

sources/imdsclient/src/lib.rs
jpculp: Right. I was just commenting on the fact that it looked a little redundant, but it's necessary.

jpculp:
- Addressed @zmrow's feedback.
- Retested by running by launching a new build of `aws-k8s-1.19` and running through `pluto`.

jpculp:
Rebased and re-tested `aws-k8s-1.19`, `aws-ecs-1`, and `vmware-k8s-1.20` to verify everything still works with https://github.com/bottlerocket-os/bottlerocket/pull/1576.

sources/imdsclient/src/lib.rs
```diff
@@ -65,57 +46,74 @@ impl ImdsClient {
...
         let target = "dynamic/instance-identity/document";
-        let response = self.fetch_bytes(target).await?;
-        let identity_document: IdentityDocument =
-            serde_json::from_slice(&response).context(error::Serde)?;
-        Ok(identity_document)
+        let response = match self.fetch_bytes(target).await? {
+            Some(response) => response,
+            None => return Ok(None),
+        };
+        let identity_document: Value = serde_json::from_slice(&response).context(error::Serde)?;
```
zmrow, line 60: Even though it was decided not to return an `IdentityDocument` from any of these functions, it might be clearer and more straightforward to keep the `IdentityDocument` struct for internal use for de-serialization purposes.  That way we are dealing with concrete types we defined and don't have to mess around with `Value`'s and string conversions, etc.

It would look like:

```
let identity_document: IdentityDocument = serde_json::from_slice(&response).context(error::Serde)?;

Ok(identity_document.region())
```

sources/imdsclient/src/lib.rs
zmrow: We could eliminate this chain if we deserialize directly into an `IdentityDocument` - see my comment above.

webern: `value.as_str()` returns an `Option<&str>` and the following `.map` converts this to an owned `Option<String>`.

webern: I'm not opposed to using the IdentityDocument as an implementation detail, but it seems more trouble than it's worth given that this is all tested and working as it.

zmrow (approved):

Issue #1209, opened 2020-11-12
provide settings for managing ephemeral disks


bcressey:
**What I'd like:**
I'd like the ability to opt-in to "managed" ephemeral disks, where the OS handles partitioning, formatting, and mounting the devices. I'd also like to be able to use this storage for the images pulled by the container runtime.

**Any alternatives you've considered:**
It's possible to handle some of this via orchestrated containers, since ephemerals now show up in `/dev/disk/ephemeral`, but that requires a daemonset or daemon task to handle something that could be done through `cloud-init` elsewhere.

However, it's not possible to use this approach to configure storage for container images, since it has to be ready before containers run.

bcressey:
Here's one potential approach, where we would extend `ghostdog` to optionally manage devices that show up in `/dev/disk/ephemeral`.

```
[settings.ephemeral-disks]
# Whether ephemeral disks will be partitioned and formatted.
managed = true 

# Whether ephemeral disks will be mounted.
mounted = true

# Which filesystem to use for formatting.
filesystem-format = [ "ext4" | "xfs" ]
```

On an EC2 instance with two ephemerals, this would create the following paths:
```
/local/ephemeral-disks/by-index/0 -> ../by-uuid/0000-....
/local/ephemeral-disks/by-index/1 -> ../by-uuid/1111-....
/local/ephemeral-disks/by-uuid/0000-....
/local/ephemeral-disks/by-uuid/1111-....
```

The index path would be a symlink to the UUID path. `ghostdog` would create the partition and filesystem if they didn't exist, and otherwise leave the device alone.

It would maintain a local mapping of the device index to the random UUID; that means some level of locking to ensure the index was only assigned once, but otherwise the paths wouldn't be predictable enough to be used by other applications. I'd thought about encoding the index in the partition label or UUID somehow, but we want to support the case where a disk is moved between systems, where the index might already be in use. The mapping could just be the symlink in the `by-index` directory. 

This assumes that all the ephemeral disks are interchangeable - same size, same device type. When we run into the case of heterogenous disks, we could extend the model with `settings.ephemeral-disks.overrides` to allow for matching by properties ("size > 20G") and selectively changing the other properties - whether to mount it, what format to use, etc.

Originally I'd thought to extend the proposal in #1208 to allow different host paths to be specified, and different security options to be applied to the directory, but I'm not sure it's worth it because of how it complicates the next feature.

The other piece of this is configuring the container runtime storage.

```
[...]

PR #1601, opened 2021-06-03
host-containers: allow mount propagations from privileged containers

arnaldo2792:
**Issue number:**
#1209 

**Description of changes:**

```
e8936fc5 host-containers: allow mount propagations from privileged containers
```

This commit adds support to propagate mount points created in bootstrap and superpowered containers, across mount peer groups.

The root filesystem of bootstrap and superpowered containers is setup with the `rshared` configuration to allow mounts propagations across peer groups. All mount points attached to the containers are configured as `rprivate` (except for the `mnt` mount). This prevents bootstrap and superpowered containers from remounting directories in the host's root filesystem.

The `/.bottlerocket/rootfs/mnt` mount point was added to bootstrap and superpowered containers. This mount point is a bind mount that points to `/mnt` in the host, which itself is a bind mount of `/local/mnt`. This is required to let users create mount points underneath `/mnt`. This mount point is setup with the `rshared` configuration to allow propagations across peer groups. This is the only mount point from which propagations are allowed across peer groups.

With this change, bootstrap containers now have access to all the devices in the host. Also, they now have the `CAP_SYS_ADMIN` capability to let users manage ephemeral disks. The logic to build the container specs was refactored to provide a better understanding of what options are set for the containers' spec.

**Testing done:**
- Admin and control containers are still working as expected and can make calls using the `apiclient`
- Non-superpowered containers don't have access to `/dev`, `/.bottlerocket/rootfs/mnt` or the host's root filesystem
- I launched a `c5d.2xlarge` instance, which has one ephemeral disk attached to it. I created a bootstrap container with the following settings:

```toml
[settings.bootstrap-containers.ephemeral]
mode = "once"
source = "<SOURCE>"
essential = false
```

Where `setup-ephemeral-disk` is defined as

```Dockerfile
FROM alpine
RUN apk add e2fsprogs bash parted
ADD script ./
RUN chmod +x ./script
ENTRYPOINT ["sh", "script"]
```
And with `script` as:

```bash
#!/usr/bin/env bash
set -ex

DISK=/.bottlerocket/rootfs/dev/nvme2n1
PERSISTENT_DIR=/.bottlerocket/bootstrap-containers/current
PARTITIONS_CREATED=/.bottlerocket/bootstrap-containers/current/created
BASE_MOUNT_POINT=/.bottlerocket/rootfs/mnt

if [ ! -f $PARTITIONS_CREATED ]; then
  parted -s $DISK mklabel gpt 1>/dev/null
  parted -s $DISK mkpart primary ext4 0% 50% 1>/dev/null
  parted -s $DISK mkpart primary ext4 50% 100% 1>/dev/null
  mkfs.ext4 -F ${DISK}p1
  mkfs.ext4 -F ${DISK}p2
  touch $PARTITIONS_CREATED
fi

mkdir -p $BASE_MOUNT_POINT/part1
mkdir -p $BASE_MOUNT_POINT/part2

mount ${DISK}p1 $BASE_MOUNT_POINT/part1
mount ${DISK}p2 $BASE_MOUNT_POINT/part2
```

I confirmed that the partitions were mounted and propagated:

```bash
# From the admin container
[ec2-user@ip-172-31-5-108 ~]$ lsblk
# ...
├─nvme2n1p1  259:14   0  93.1G  0 part /.bottlerocket/rootfs/mnt/part1
└─nvme2n1p2  259:15   0  93.1G  0 part /.bottlerocket/rootfs/mnt/part2

# From the host
bash-5.0# lsblk
# ...
|-nvme2n1p1  259:14   0  93.1G  0 part /local/mnt/part1
|-nvme2n1p2  259:15   0  93.1G  0 part /local/mnt/part2
```

From the admin container I ran:

```bash
[ec2-user@ip-172-31-12-31 ~]$ sudo su
bash-4.2# umount /.bottlerocket/rootfs/mnt/part1
bash-4.2# mount /dev/nvme2n1p1 /.bottlerocket/rootfs/etc/
bash-4.2# ls /.bottlerocket/rootfs/etc/
lost+found
``` 

And confirmed that the mount point didn't propagate:
```bash
# From the host
bash-5.0# lsblk
# ...
nvme2n1      259:0    0 186.3G  0 disk
|-nvme2n1p1  259:14   0  93.1G  0 part
|-nvme2n1p2  259:15   0  93.1G  0 part /local/mnt/part2

bash-5.0# ls /etc/
bootstrap-containers  ...
```

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -528,7 +528,10 @@ Bootstrap containers are host containers that can be used to "bootstrap" the hos
 
 Bootstrap containers are very similar to normal host containers; they come with persistent storage and with optional user data.
 Unlike normal host containers, bootstrap containers can't be treated as `superpowered` containers.
-However, these containers have access to the underlying root filesystem on `/.bottlerocket/rootfs`.
+However, bootstrap containers do have additional permissions that normal host containers do not have.
+Bootstrap containers have access to the underlying root filesystem on `/.bottlerocket/rootfs` as well as to all the devices in the host, and they are set up with the `CAP_SYS_ADMIN` capability.
+This allows bootstrap containers to create files, directories, and mounts that are visible to the host.
+
 Bootstrap containers are set up to run after the systemd `configured.target` unit is active.
 The containers' systemd unit depends on this target (and not on any of the bootstrap containers' peers) which means that bootstrap containers will not execute in a deterministic order
 The boot process will "wait" for as long as the bootstrap containers run.
@@ -557,6 +560,11 @@ source = "MY-CONTAINER-URI"
 mode = "once"
 essential = true
 ```
+
+##### Mount propagations in bootstrap and superpowered containers
+Both bootstrap and superpowered host containers are configured with the `/.bottlerocket/rootfs/mnt` bind mount that points to `/mnt` in the host, which itself is a bind mount of `/local/mnt`.
+This bind mount is set up with shared propagations, so any new mount point created underneath `/.bottlerocket/rootfs/mnt` in any bootstrap or superpowered host container will propagate across mount namespaces.
+You can use this feature to configure ephemeral disks attached to your hosts that you may want to use on your workloads.
 
 #### Platform-specific settings
 
--- a/packages/release/mnt.mount
+++ b/packages/release/mnt.mount
@@ -1,0 +1,16 @@
+[Unit]
+Description=Mnt Directory (/mnt)
+DefaultDependencies=no
+Conflicts=umount.target
+Before=local-fs.target umount.target
+Wants=prepare-local.service
+After=prepare-local.service
+
+[Mount]
+What=/local/mnt
+Where=/mnt
+Type=none
+Options=rbind,rshared
+
+[Install]
+WantedBy=local-fs.target
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -272,38 +272,30 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 		// Set the destination name for the container persistent storage location
 		persistentDir := cType.PersistentDir()
 
-		// Set up the container spec. See `withSuperpowered` for conditional options
-		// set when configured as superpowered.
-		ctrOpts := containerd.WithNewSpec(
+		specOpts := []oci.SpecOpts{
 			oci.WithImageConfig(img),
 			oci.WithHostNamespace(runtimespec.NetworkNamespace),
 			oci.WithHostHostsFile,
 			oci.WithHostResolvconf,
-			// Mount in the API socket for the Bottlerocket API server, and the API
-			// client used to interact with it
-			oci.WithMounts([]runtimespec.Mount{
-				{
-					Options:     []string{"bind", "rw"},
-					Destination: "/run/api.sock",
-					Source:      "/run/api.sock",
-				},
-				// Mount in the apiclient to make API calls to the Bottlerocket API server
-				{
-					Options:     []string{"bind", "ro"},
-					Destination: "/usr/local/bin/apiclient",
-					Source:      "/usr/bin/apiclient",
-				}}),
 			// Pass proxy environment variables to this container
 			withProxyEnv(),
-			// Mount in the persistent storage location for this container
-			withPersistentStorage(containerName, persistentDir),
-			// Mount the rootfs with an SELinux label that makes it writable
+			// Add a default set of mounts regardless of the container type
+			withDefaultMounts(containerName, persistentDir),
+			// Mount the container's rootfs with an SELinux label that makes it writable
 			withMountLabel("system_u:object_r:secret_t:s0"),
-			// Include conditional options for superpowered containers.
-			withSuperpowered(superpowered),
-			// Mount the rootfs if superpowered or bootstrap
-			withRootFilesystemMounts(superpowered || cType == bootstrap),
-		)
+		}
+
+		// Select the set of specOpts based on the container type
+		switch {
+		case superpowered:
+			specOpts = append(specOpts, withSuperpowered())
+		case cType == bootstrap:
+			specOpts = append(specOpts, withBootstrap())
+		default:
+			specOpts = append(specOpts, withDefault())
+		}
+
+		ctrOpts := containerd.WithNewSpec(specOpts...)
 
 		// Create the container.
 		container, err = client.NewContainer(
@@ -609,6 +601,48 @@ func newContainerdClient(ctx context.Context, containerdSocket string, namespace
 	return client, nil
 }
 
+// withSuperpowered adds container options to grant administrative privileges
+func withSuperpowered() oci.SpecOpts {
+	return oci.Compose(
+		withPrivilegedMounts(),
+		withRootFsShared(),
+		oci.WithHostNamespace(runtimespec.PIDNamespace),
+		oci.WithParentCgroupDevices,
+		oci.WithPrivileged,
+		oci.WithNewPrivileges,
+		oci.WithSelinuxLabel("system_u:system_r:super_t:s0"),
+		oci.WithAllDevicesAllowed,
+	)
+}
+
+// withBootstrap adds container options to grant read-write access to the underlying
+// root filesystem, as well as to manage the devices attached to the
+// host
+func withBootstrap() oci.SpecOpts {
+	return oci.Compose(
+		withPrivilegedMounts(),
+		withRootFsShared(),
+		oci.WithSelinuxLabel("system_u:system_r:control_t:s0"),
+		// Bootstrap containers don't require all "privileges", we only add the
+		// `CAP_SYS_ADMIN` capability. `WithDefaultProfile` will create the proper
+		// seccomp profile based on the container's capabilities.
+		oci.WithAddedCapabilities([]string{"CAP_SYS_ADMIN"}),
+		seccomp.WithDefaultProfile(),
+		oci.WithAllDevicesAllowed,
+	)
+}
+
+// withDefault adds container options for non-privileged containers
+func withDefault() oci.SpecOpts {
+	return oci.Compose(
+		oci.WithSelinuxLabel("system_u:system_r:control_t:s0"),
+		// Non-privileged containers only have access to a subset of the devices
+		oci.WithDefaultUnixDevices,
+		// No additional capabilities required for non-privileged containers
+		seccomp.WithDefaultProfile(),
+	)
+}
+
 // withMountLabel configures the mount with the provided SELinux label.
 func withMountLabel(label string) oci.SpecOpts {
 	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -493,7 +493,8 @@ Bootstrap containers are host containers that can be used to "bootstrap" the hos
 
 Bootstrap containers are very similar to normal host containers; they come with persistent storage and with optional user data.
 Unlike normal host containers, bootstrap containers can't be treated as `superpowered` containers.
-However, these containers have access to the underlying root filesystem on `/.bottlerocket/rootfs`.
+However, these containers have access to the underlying root filesystem on `/.bottlerocket/rootfs` as well as to all the devices in the host, and they are setup with the `CAP_SYS_ADMIN` capability.
```
samuelkarp:
```suggestion
However, bootstrap containers do have additional permissions that normal host containers do not have.
Bootstrap containers have access to the underlying root filesystem on `/.bottlerocket/rootfs`, access to all the devices in the host, and have the `CAP_SYS_ADMIN` capability.
This allows bootstrap containers to create files, directories, and mounts that are visible to the host.
```

README.md
```diff
@@ -523,6 +524,11 @@ mode = "once"
 essential = true
 ```
 
+##### Mount propagations in bootstrap and superpowered containers
+Both bootstrap and superpowered host containers are configured with a bind mount that points to `/mnt` in the host, which itself is a bind mount of `/local/mnt`.
+This bind mount is setup with shared propagations, so any new mount point created underneath `/mnt` in any bootstrap or superpowered host container will propagate across mount namespaces.
```
samuelkarp: nit:
```suggestion
This bind mount is set up with shared propagations, so any new mount point created underneath `/mnt` in any bootstrap or superpowered host container will propagate across mount namespaces.
```

README.md
```diff
@@ -523,6 +524,11 @@ mode = "once"
...
+You can use this feature to configure ephimeral disks attached to your hosts, that you may want to use on your workloads.
```
samuelkarp: nit:
```suggestion
You can use this feature to configure ephemeral disks attached to your hosts that you may want to use on your workloads.
```

README.md
```diff
@@ -523,6 +524,11 @@ mode = "once"
 essential = true
 ```
 
+##### Mount propagations in bootstrap and superpowered containers
+Both bootstrap and superpowered host containers are configured with a bind mount that points to `/mnt` in the host, which itself is a bind mount of `/local/mnt`.
```
samuelkarp: I think it makes more sense for the bind mount to be mounted at `/.bottlerocket/rootfs/mnt` inside the container.  That way all of the host-related configuration conceptually happens inside the same hierarchy.

packages/release/mnt.mount
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Mnt Directory (/mnt)
+DefaultDependencies=no
+Conflicts=umount.target
+Before=local-fs.target umount.target
```
samuelkarp, line 5: Why is `umount.target` included in the `Before` line?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -279,30 +280,20 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
...
 			// Include conditional options for superpowered containers.
 			withSuperpowered(superpowered),
-			// Mount the rootfs if superpowered or bootstrap
-			withRootFilesystemMounts(superpowered || cType == bootstrap),
+			// Mount privileged containers's rootfs with shared propagations
+			withRootFsShared(privileged),
+			// Set the privileges for this container
+			withPrivileges(superpowered, cType == bootstrap),
+			// Set what devices are allowed for this container
+			withDevices(privileged),
```
samuelkarp: I think this is starting to get unwieldy and the logic for "what options are set on the spec" is mixed between this function and all these helper functional options.

I think a somewhat better structure would separate the two and make the code easier to reason about:

```go
specOpts := []oci.SpecOpts{
	// basic spec opts shared between all containers
	oci.WithImageConfig(img),
	oci.WithHostNamespace(runtimespec.NetworkNamespace),
	oci.WithHostHostsFile,
	oci.WithHostResolvconf,
	withProxyEnv(),
	withDefaultMounts(),
	withMountLabel("system_u:object_r:secret_t:s0"),
}
switch {
case superpowered:
	specOpts = append(specOpts, withSuperpowered())
case cType == bootstrap:
	specOpts = append(specOpts, withBootstrap())
default:
	specOpts = append(specOpts, withDefault())
}
``` 

```go
func withSuperpowered() oci.SpecOpts {
	return oci.Compose(
		oci.WithSelinuxLable(...),
		withPrivilegedMounts(),
		withRootFsShared(),
		withPrivileged(),
		withDevices(),
}

func withBootstrap() oci.SpecOpts {
	return oci.Compose(
		withPrivilegedMounts(),
		withRootFsShared(),
		withCapSysAdmin(),
		withDevices(),
	)
}

func withDefault() oci.SpecOpts {
	return oci.Compose(
		// if there's anything that even has to go here...
	)
}
```

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -626,74 +617,165 @@ func withSuperpowered(superpowered bool) oci.SpecOpts {
...
 
-// withRootFileSystemMounts adds container options to mount the root filesystem
-func withRootFilesystemMounts(mountRootFilesystem bool) oci.SpecOpts {
-	// if mountRootFilesystem, return a no-op SpecOpts function
-	if !mountRootFilesystem {
-		return func(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
-			return nil
-		}
+// withMounts configures all the mounts for the container. If the contianer is privileged,
+// the following mounts from the underlying host are included:
```
samuelkarp: As above, I think this is better separated into multiple functional options that each have no logic (or a minimal amount of logic) inside.  One that handles default mounts and another that deals with privileged mounts.

packages/release/mnt.mount
arnaldo2792: That's how we mount other directories like `/var` and `/opt`. According to the systemd's documentation:

> All mount units acquire automatic Before= and Conflicts= on umount.target in order to be stopped during shutdown.

So I guess those two lines aren't actually needed, and I think we added them to clarify what dependencies the unit has.

arnaldo2792:
Forced push includes:

- Fixes in main README
- Refactor how container specs are created as suggested in feedback
- Mount `/.bottlerocket/rootfs/mnt` instead of `/mnt` inside privileged containers

arnaldo2792:
Forced push to fix commit message

README.md
```diff
@@ -493,7 +493,10 @@ Bootstrap containers are host containers that can be used to "bootstrap" the hos
 
 Bootstrap containers are very similar to normal host containers; they come with persistent storage and with optional user data.
 Unlike normal host containers, bootstrap containers can't be treated as `superpowered` containers.
-However, these containers have access to the underlying root filesystem on `/.bottlerocket/rootfs`.
+However, bootstrap containers do have additional permissions that normal host containers do not have.
+Bootstrap containers have access to the underlying root filesystem on `/.bottlerocket/rootfs` as well as to all the devices in the host, and they are setup with the `CAP_SYS_ADMIN` capability.
```
samuelkarp: nit:
```suggestion
Bootstrap containers have access to the underlying root filesystem on `/.bottlerocket/rootfs` as well as to all the devices in the host, and they are set up with the `CAP_SYS_ADMIN` capability.
```

README.md
```diff
@@ -523,6 +526,11 @@ mode = "once"
 essential = true
 ```
 
+##### Mount propagations in bootstrap and superpowered containers
+Both bootstrap and superpowered host containers are configured with the `/.bottlerocket/rootfs/mnt` bind mount that points to `/mnt` in the host, which itself is a bind mount of `/local/mnt`.
+This bind mount is set up with shared propagations, so any new mount point created underneath `/.bottlerocket/rootfs/mnt` in any bootstrap or superpowered host container will propagate across mount namespaces.
+You can use this feature to configure ephemeral disks attached to your hosts, that you may want to use on your workloads.
```
samuelkarp: nit:
```suggestion
You can use this feature to configure ephemeral disks attached to your hosts that you may want to use on your workloads.
```

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -641,59 +615,147 @@ func withSuperpowered(superpowered bool) oci.SpecOpts {
...
-		Options:     []string{"rbind", "rw"},
-		Destination: fmt.Sprintf("/.bottlerocket/%s/%s", persistentDir, containerID),
-		Source:      fmt.Sprintf("/local/%s/%s", persistentDir, containerID),
-	}}
+// withDefaultMounts adds the mount configurations required in all container types
+func withDefaultMounts(containerID string, persistentDir string) oci.SpecOpts {
+	var mounts = []runtimespec.Mount{
+		// Local persistent storage for the container
+		{
+			Options:     []string{"rbind", "rw"},
```
samuelkarp, line 662: Do these mounts need to include `slave` or `rslave` for superpowered/boostrap containers as well?

samuelkarp (commented):
The reorganized code is significantly easier to read; thanks for doing that.

sources/host-ctr/cmd/host-ctr/main.go
arnaldo2792: ~~I don't think we should set up those bindings as `slave`s since there aren't any bindings in the containers' persistent storage directory that we want to propagate from the host to the containers' mount namespace.~~

~~IMO, we should only propagate mounts down when we want to share them consistently across containers. We shouldn't touch the containers' persistent storage directory since doing so could impact how the users are using that storage in their own containers.~~

After an offline chat with @bcressey and some more testing, we decided to keep all mounts as `rprivate`, except for `/.bottlerocket/rootfs/mnt` for two reasons:

- Mounts will behave the same as they did prior to this change
- `rprivate` prevents undesired propagations from mount points other than `/.bottlerocket/rootfs/mnt`

arnaldo2792:
Forced push includes fixes for feedback in the main README

arnaldo2792:
Forced push includes:

- All mounts, except for `/.bottlerocket/rootfs/mnt`, are `rprivate` to prevent propagations from them
- Removed `/dev` from mounts since the host container already has access to it from `/.bottlerocket/rootfs/dev`

packages/release/mnt.mount
```diff
@@ -0,0 +1,16 @@
...
+Wants=prepare-local.service
+After=prepare-local.service
+
+[Mount]
+What=/local/mnt
+Where=/mnt
+Type=none
+Options=rbind,rshared
```
bcressey, line 13: This will change the effective default label of `/mnt` from `external_t` to `local_t`.

When you mount a newly formatted filesystem, what label does its mount point end up with?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -641,59 +615,174 @@ func withSuperpowered(superpowered bool) oci.SpecOpts {
...
+
+	for _, option := range mount.Options {
+		hasPropagation = re.FindString(option) != ""
+
+		if hasPropagation {
+			break
+		}
+	}
+
+	return hasPropagation
```
samuelkarp: I'd prefer to avoid the use of a regular expression here.  There are only six values we need to test for.

```suggestion
	// Propagations can be shared, rshared, private, rprivate, slave, rslave
	for _, option := range mount.Options {
		switch option {
		case "shared", "rshared", "private", "rprivate", "slave", "rslave":
			return true
		}
	}

	return false
```

packages/release/mnt.mount
arnaldo2792: New mount points are labeled with the `system_u:object_r:unlabeled_t:s0` label:

```
[ec2-user@<> ~]$ ls -Z /.bottlerocket/rootfs/mnt/
drwxr-xr-x. root root system_u:object_r:unlabeled_t:s0 part1
drwxr-xr-x. root root system_u:object_r:unlabeled_t:s0 part2
```

arnaldo2792:
Forced push includes refactor in `hasPropagations` method as suggested by @samuelkarp

bcressey (approved):

samuelkarp (approved):

arnaldo2792:
Forced push due to conflicts in rebase

arnaldo2792:
Forced push includes file missing in previous commit to fix rebase conflicts

bcressey (approved):

samuelkarp (approved):

PR #1602, opened 2021-06-03
release: mark ipv6 lease as optional for eth0

### LINKED ISSUE #1351: Booting EKS image in a VPC w/o IPv6 support takes additional ~30s
**Image I'm using:**

`/aws/service/bottlerocket/aws-k8s-1.18/x86_64/latest/image_id` in us-west-2, presently:

- ami-09030b5be1adb0be1
- bottlerocket-aws-k8s-1.18-x86_64-v1.0.5-f306b954

I expected the kubelet to be ready in (from initial boot) somewhere in the 10-20s range

**What actually happened:**

It ended up being somewhere in the 50-60s range

Launch the above ami in a subnet that does not support IPv6 (eg no IPv6 CIDR block and no "auto-assign IPv6 address" turned on). Then run `systemd-analyze blame` and you'll see something like:

```
bash-5.0# systemd-analyze blame
30.087s wicked.service                                                             
 8.077s kubelet.service                                                            
 7.079s systemd-tmpfiles-setup.service                                             
 2.497s containerd.service                                                         
 2.476s host-containerd.service                                                    
 1.167s settings-applier.service                                                   
 1.092s sundog.service
[snip]
```

Whereas if you enable IPv6/auto-assign, the above will look more like:

```
bash-5.0# systemd-analyze blame
3.778s wicked.service                                                             
1.589s kubelet.service                                                            
 554ms prepare-local.service                                                      
 454ms systemd-udev-settle.service                                                
 395ms dev-dm\x2d0.device                                                         
 370ms settings-applier.service                                                   
 336ms systemd-tmpfiles-setup.service                                             
 309ms containerd.service
[snip]
```

etungsten:
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/1351

**Description of changes:**
```
uthor: Erikson Tung <etung@amazon.com>
Date:   Thu Jun 3 12:47:44 2021 -0700

    release: mark ipv6 lease as optional for eth0

    This marks the DHCPv6 as optional in the network interface configuration
    file for eth0.

    When wicked reads the network interface configuration, it knows not to
    wait for an ipv6 address lease. Wicked no longer waits the full 30
    seconds for an ipv6 address lease when there is no ipv6 support.


```

**Testing done:**
Launched 10 instances in an subnet without ipv6 address assignment, all of them get initialized much quicker than before. Wicked no longer waits the full 30 seconds for an ipv6 address assignment.

```
bash-5.0# systemd-analyze blame
7.300s activate-multi-user.service                                                
7.036s kubelet.service                                                            
2.148s systemd-tmpfiles-setup.service                                             
1.838s wicked.service              
...

```

```
bash-5.0# wicked ifstatus eth0
...
eth0            up
      link:     #2, state up, mtu 9001
      type:     ethernet, hwaddr 02:64:19:14:01:f3
      config:   wicked:xml:/etc/wicked/ifconfig/eth0.xml
      leases:   ipv4 dhcp granted
      leases:   ipv6 dhcp requesting
      addr:     ipv4 192.168.25.30/19 [dhcp]
      route:    ipv4 default via 192.168.0.1 proto dhcp
wicked: Exit with status: 0

```

Then I enabled IPv6 address assignment for a subnet and launched an instance into it, wicked comes up fine, the host gets both an ipv4 address and ipv6 address without problem:
```
bash-5.0# systemd-analyze blame
8.869s activate-multi-user.service                                                
8.639s kubelet.service                                                            
7.965s systemd-tmpfiles-setup.service                                             
3.523s wicked.service                 
```

```
bash-5.0# wicked ifstatus eth0
eth0            up
      link:     #2, state up, mtu 9001
      type:     ethernet, hwaddr 02:f8:6d:24:d5:d1
      config:   wicked:xml:/etc/wicked/ifconfig/eth0.xml
      leases:   ipv4 dhcp granted
      leases:   ipv6 dhcp granted
      addr:     ipv4 192.168.18.82/19 [dhcp]
      addr:     ipv6 2600:1f14:326:5500:8a3a:6b96:f3ae:a87e/64 [dhcp]
      route:    ipv4 default via 192.168.0.1 proto dhcp
      route:    ipv6 default via fe80::cc:7cff:feec:c0f1 metric 1024 proto ra
wicked: Exit with status: 0

```

**Terms of contribution:**

### DIFF
--- a/packages/release/eth0.xml
+++ b/packages/release/eth0.xml
@@ -21,5 +21,8 @@
   <ipv6:dhcp>
     <enabled>true</enabled>
     <defer-timeout>1</defer-timeout>
+    <flags>
+      <optional />
+    </flags>
   </ipv6:dhcp>
 </interface>


jpculp (approved):
I wonder if 1 second is slightly too aggressive and if something 2-5 might be a safer choice.

etungsten:
> I wonder if 1 second is slightly too aggressive and if something 2-5 might be a safer choice.

If the wicked client doesn't wait, the dhcp clients would still attempt to lease addresses in the background. Timing out doesn't actually cause `wicked.service` to fail or anything.

Setting it higher would just result in more time spent waiting for dhcpv6 lease when wicked decides to wait. I guess the concern would be that services that depend on `wicked.service` might start up when the host still doesn't have an IP assigned. It's not clear to me what impact that would have. 

It's also not clear to me how much time is appropriate to wait for an IP under suboptimal network conditions. It's unfortunate that we can't separately specify the timeout for dhcpv4 and dhcpv6. I'll go ahead and set the timeout to 5 seconds then. (5000 milliseconds sound like more than enough time for dhcpv4 to get an address).

etungsten:
Push above changes the timeout value from 1 seconds to 5 seconds. Updated PR description to reflect changes.

webern (approved):
nice fix

samuelkarp (approved):

zmrow (approved):

etungsten:
Push above changes the approach. @bcressey found a better way of making wicked skip waiting on an ipv6 address lease.
Instead of changing the wicked `wait_for_interfaces` timeout, we now mark dhcpv6 as optional in the eth0 interface config.

Updated PR description to reflect changes.
Tested things and things work as expected, testing section in the PR description has more details.

webern (approved):

bcressey (approved):

zmrow (approved):

PR #1610, opened 2021-06-14
imdsclient: update readme to reflect new helper functions

jpculp:
**Issue number:**

N/A

**Description of changes:**

This is an update to documentation as `fetch_dynamic` and `fetch_metadata` were previously replaced with more targeted functions.

**Testing done:**

Built `imdsclient` crate.

**Terms of contribution:**

### DIFF
--- a/sources/imdsclient/README.md
+++ b/sources/imdsclient/README.md
@@ -2,11 +2,19 @@
 
 Current version: 0.1.0
 
-The imdsclient library provides high-level methods to interact with the AWS Instance Metadata Service.
-The high-level methods provided are [`fetch_dynamic`], [`fetch_metadata`], and [`fetch_userdata`].
+`imdsclient` provides high-level methods to interact with the AWS Instance Metadata Service (IMDS).
 
-For more control, and to query IMDS without high-level wrappers, there is also a [`fetch_imds`] method.
-This method is useful for specifying things like a pinned date for the IMDS schema version.
+The library uses IMDSv2 (session-oriented) requests over a pinned schema to guarantee compatibility.
+Session tokens are fetched automatically and refreshed if the request receives a `401` response.
+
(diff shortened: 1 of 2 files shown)


sources/imdsclient/README.md
```diff
@@ -3,9 +3,19 @@
...
+- `fetch_instance_type`
+- `fetch_local_ipv4_address`
+- `fetch_mac_addresses`
+- `fetch_public_ssh_keys`
+- `fetch_region`
+- `fetch_userdata`
+
+For custom IMDS end targets, you can use `fetch_bytes` and `fetch_string`.
+
+To query IMDS without high-level wrappers, there is also a `fetch_imds` method.
```
zmrow: Nit: Just from reading this, I'm not sure when I would use `fetch_bytes/string` vs. `fetch_imds`.  Can we clarify a bit?

jpculp:
Updated for further clarity.

zmrow (approved):
:ice_cream:

webern (commented):
I actually think it would be better to remove the enumeration of high level function names since it will always be in danger of being incorrect.

Also the blurbs about fetch_imds, fetch_bytes and fetch_string are incorrect. These are private functions and as-such are implementation details that shouldn't go in the readme.

jpculp:
Hm. I forgot that we switched the last 3 to private, but now that I think about it I wonder if that was really the right thing to do.

jpculp:
Took a different approach to the documentation.

webern (approved):
I like this version much better!

zmrow (approved):

jpculp:
Changed ``` to ` so cargo doesn't try to test the example uri.

zmrow (approved):

webern (approved):

PR #1615, opened 2021-06-16
Docker update

arnaldo2792:
**Issue number:**
N / A

**Description of changes:**

```
b03d2cf2 packages: update docker-proxy
0ba73bff packages: update docker-engine
badba994 packages: update docker-cli
```

Update docker to 20.10.4. In [docker's deprecation page](https://docs.docker.com/engine/deprecated/) they list a few feature that were removed in the 20.10 series. I think none of those affected the ECS variant's behavior since the conformance testing is passing. 

**Testing done:**

In x86_64/aarch64:
- [x] Run container in aws-dev variant
- [x] Run task in aws-ecs variant

In x86_64:
- [x] Run conformance testing for ecs variant

**Terms of contribution:**

### DIFF
--- a/packages/docker-cli/Cargo.toml
+++ b/packages/docker-cli/Cargo.toml
@@ -9,8 +9,8 @@ build = "build.rs"
 path = "pkg.rs"
 
 [[package.metadata.build-package.external-files]]
-url = "https://github.com/docker/cli/archive/v19.03.15/cli-19.03.15.tar.gz"
-sha512 = "163f67a11b1d976eb91c84e59eb1fabf3f26b360ac726bb0bedb9a1532ef807b665f7461f52f6b236cb8af955af042466070800cd4d09b7a6b90092daf718b41"
+url = "https://github.com/docker/cli/archive/v20.10.4/cli-20.10.4.tar.gz"
+sha512 = "861f69657ac3eede228983b7d845ce98c81f4b0aa601aab37024d3f21cf1ca73a182d33bdde8fb9ad89e4954c3903dc4ec2b81fcf7364941a7c38a80ea410e34"
 
 [build-dependencies]
 glibc = { path = "../glibc" }
--- a/packages/docker-engine/docker-engine.spec
+++ b/packages/docker-engine/docker-engine.spec
@@ -3,9 +3,9 @@
 %global goorg github.com/docker
 %global goimport %{goorg}/docker
 
-%global gover 19.03.15
+%global gover 20.10.4
 %global rpmver %{gover}
-%global gitrev 420b1d36250f9cfdc561f086f25a213ecb669b6f
+%global gitrev 363e9a88a11be517d9e8c65c998ff56f774eb4dc
 
 %global source_date_epoch 1363394400
 
@@ -44,7 +44,7 @@ Requires: %{_cross_os}procps
 
 %build
 %cross_go_configure %{goimport}
-BUILDTAGS="autogen journald selinux seccomp"
+BUILDTAGS="journald selinux seccomp"
 BUILDTAGS+=" exclude_graphdriver_btrfs"
 BUILDTAGS+=" exclude_graphdriver_devicemapper"
 BUILDTAGS+=" exclude_graphdriver_vfs"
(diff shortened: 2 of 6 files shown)


packages/docker-engine/docker-engine.spec
```diff
@@ -54,8 +54,7 @@ export VERSION=%{gover}
 export GITCOMMIT=%{gitrev}
 export BUILDTIME=$(date -u -d "@%{source_date_epoch}" --rfc-3339 ns 2> /dev/null | sed -e 's/ /T/')
 export PLATFORM="Docker Engine - Community"
-chmod +x ./hack/make/.go-autogen
-./hack/make/.go-autogen
+source ./hack/make/.go-autogen
```
samuelkarp, line 57: Why is this sourced instead of being run?

packages/docker-cli/Cargo.toml
```diff
@@ -9,8 +9,8 @@ build = "build.rs"
 path = "pkg.rs"
 
 [[package.metadata.build-package.external-files]]
-url = "https://github.com/docker/cli/archive/v19.03.15/cli-19.03.15.tar.gz"
-sha512 = "163f67a11b1d976eb91c84e59eb1fabf3f26b360ac726bb0bedb9a1532ef807b665f7461f52f6b236cb8af955af042466070800cd4d09b7a6b90092daf718b41"
+url = "https://github.com/docker/cli/archive/refs/tags/v20.10.4.tar.gz"
```
etungsten: Just curious, is there a difference between getting the source files from `https://github.com/docker/cli/archive/v"${VERSION}"/cli-"${VERSION}".tar.gz` and `https://github.com/docker/cli/archive/refs/tags/v"${VERSION}".tar.gz`?
I'm guessing the former is downloading from a release tag, and the latter is downloading from a tag?

packages/docker-engine/docker-engine.spec
arnaldo2792: I followed what the [`binary`](https://github.com/moby/moby/blob/2773f81aa5e9e34733675a7aa7e4219391caccb0/hack/make/.binary#L15) script does in the moby repo.

packages/docker-cli/Cargo.toml
arnaldo2792: I had a typo when I tried to download from archive instead of ref/tag, and I thought there wasn't any 20.10.4 version available through the archive. I was using 20.10.04 instead of 20.10.4, which is what I use now.

etungsten (approved):

packages/docker-engine/docker-engine.spec
samuelkarp: `./hack/make/.go-autogen` [changed significantly](https://github.com/moby/moby/commit/675b414f56063826457cbc8f0c165af67198bfa8#diff-28a29ae1b36a23b6fd072d1ab3bd970e743b200728fe8a2598ea14277193025b) between 19.03.15 and 20.10.4 and the result is that building Docker this way will lose a bunch of the version information we expect.

In addition to the `source` change that you made, we'll also need to update the `go build` line to include `${LDFLAGS}`.

samuelkarp (changes requested):

packages/docker-engine/docker-engine.spec
arnaldo2792: I updated the `go build` command to include the `LDFLAGS`. I can see the commit and version in the docker logs:

```
Jun 17 19:41:00  dockerd[1651]: time="2021-06-17T19:41:00.748861055Z" level=info msg="Docker daemon" commit=363e9a88a11be517d9e8c65c998ff56f774eb4dc graphdriver(s)=overlay2 version=20.10.4
```

The log line uses the dockerversion paramters passed in `LDFLAGS`:

https://github.com/moby/moby/blob/b9ad7b96bd86e5f632e5fb4f36f98dcc145014fc/daemon/daemon.go#L1141

samuelkarp: Can you post the output of `docker version` too?

arnaldo2792: The output of `docker version`:

```
bash-5.0# docker version
Client: Docker Engine - Community
 Version:           20.10.4
 API version:       1.41
 Go version:        go1.16.3
 Git commit:        d3cb89ee53287cf9214ad5fbc9a111190f607c72
 Built:             Tue Apr 18 14:29:00 2017
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.4
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.16.3
  Git commit:       363e9a88a11be517d9e8c65c998ff56f774eb4dc
  Built:            Sat Mar 16 00:40:00 2013
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4+bottlerocket
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93+bottlerocket
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:

```

samuelkarp (approved):

tjkirch:
Closing in favor of #1625, which is the same commits rebased onto develop.  That let the CI runs pass.

Issue #1605, opened 2021-06-07
VMware: Provide a `cargo make` target for OVA upload/template


zmrow:
**Feature**
OVAs are already automatically built for VMware variants.  It would be nice to provide a friendly `cargo make` target to upload OVAs to one or more vSphere datacenters, optionally making them into VM templates.

**`cargo make` interface**
We will provide two new make targets `cargo make upload-ova` and `cargo make vmware-template`.  Providing the `upload-ova` target means that one can upload an OVA without making it a template, which can also be useful for some users.  The `vmware-template` target will upload the OVA automatically so a user doesn't need to run multiple targets.

**Under the hood**
We will build `govc` into the SDK container, which a user will already have if they have built any images.  `pubsys` (called via `cargo make`) will orchestrate `docker run` commands to call `govc`.  `govc` configuration will be provided via environment variables in the `docker run` call.  Configuration will be provided for 1 or more datacenters in `Infra.toml` and a new file `~/.config/pubsys/vsphere-credentials.toml`.

At runtime, `pubsys` will iterate through all the datacenters provided in `Infra.toml`.  A user can upload to a subset of datacenters using an environment variable.

Users can provide configuration for multiple datacenters in their own TOML blocks (see below).  Any "common" configuration can be provided for keys that are the same for all datacenters.  **_All configuration may be overridden at runtime via `GOVC_` environment variables_**.  

Credentials will be provided in `~/.config/pubsys/vsphere-credentials.toml`.  At runtime, `pubsys` will look in this file for credentials for the current datacenter.  If they do not exist, it will look for the environment variables `GOVC_USERNAME` and `GOVC_PASSWORD`.

**Infra.toml**
```
[vmware]
datacenters = ["foo", "bar", "baz"]

# Optional common configuration.
# This configuration allow values to be set in a single place if they are common in 
# multiple datacenters.  They can be overridden in the datacenter's block
[vmware.common]
network = "a_network"

[vmware.datacenter.foo]
vsphere_url = "https://a_datacenter.com" # GOVC_URL
datacenter = "a_datacenter" # GOVC_DATACENTER
datastore = "a_datastore" # GOVC_DATASTORE
network = "another_network" # GOVC_NETWORK
folder = "a_folder" # GOVC_FOLDER
resource_pool = "/a/resource/pool" # GOVC_RESOURCE_POOL

[vmware.datacenter.bar]
...
```

**vsphere-credentials.toml**
```
[datacenter.foo]
[...]

PR #1622, opened 2021-06-17
Add support for uploading OVAs to VMware

zmrow:
**Issue number:**
Implements #1605 

**Description of changes:**

It's probably easiest to read through the PR by commit.
```
pubsys-config: Add VMware support
     
    This adds everything needed to support VMware specific configuration,
    including credentials.
```
```
pubsys: Add VMware 'upload-ova' subcommand
    
    This adds the subcommand `upload-ova` to `pubsys`.  This subcommand is
    meant to be used via `cargo make` and allow a user to upload an OVA to
    one or more VMware datacenters, optionally marking the VM as a template.
    Under the hood, it accomplishes this by running `govc` in the SDK
    container via Docker invocation, similar to how `buildsys` accomplishes
    its task.
    
    Like the rest of `pubsys`, it gets much of its configuration from
    `Infra.toml`.  The configuration details all of the values needed to
    communicate with a vSphere instance, and allows for users to specify
    "common" configuration that multiple datacenters share.  Credentials for
    different datacenters can be specified in a file located at
    `~/.config/pubsys/vsphere-credentials.toml`.  Because we use `govc`
    under the hood, at runtime we check for GOVC_* environment variables and
    allow them to override VMware datacenter config in `Infra.toml` or
    `vsphere-credentials.toml`
    
    The subcommand allows a user to override the name of the resulting VM,
    import spec, and specify a subset of datacenters than what's specified
    in config.
```
```
Makefile: Add `upload-ova` and `vmware-template` targets
    
    This change adds the ability to upload a built OVA to one or more VMware
    software defined datacenters.  The `vmware-template` target uploads the
    OVA and marks it as a VM template using the import spec.  Multiple
    environment variables allow a user to override the VM name, list of
    datacenters to upload to, and the import spec used to upload.
```
```
PUBLISHING: Split relevent AWS/VMware details into their own guides.
    
    This change splits `PUBLISHING.md` into multiple files to account for
    the differences in publishing Bottlerocket on AWS and VMware.  The
    `PUBLISHING-AWS.md` guide covers publishing and granting access to AMIs,
    and details around SSM parameters.  The `PUBLISHING-VMWARE.md` guide
    covers the configuration details and process for uploading an OVA to one
    or more VMware SDDC's.  The main `PUBLISHING.md` still covers all
    general information around configuration, repos, etc.
```

**Testing done:**

I performed the test steps with varying configuration setups:
*   _only_ environment variables (no datacenter-specific details in `Infra.toml` or creds file)
*  _only_ datacenter config/credentials file (no environment variables)
* datacenter config/creds file with values missing but populated in the environment
* environment variables with values missing but populated in the files
* all config populated in the `[vmware.common]` section
* various values populated in `common`, `Infra.toml` and environment to ensure the values cascade as expected

Testing steps:
* Upload an OVA
* Upload an OVA as template
* Override VM name, import_spec, and datacenters.  (I don't have multiple datacenters to push to so I populated the same details for multiple datacenters and ensured that pubsys attempts to upload to them both, failing the second time because a VM already exists)
* Attempt to re-upload the same OVA, fails as expected with a nice message:
```
[INFO] Uploading OVA to datacenter 'main' with name 'bottlerocket-vmware-k8s-1.20-x86_64-v1.1.1-5bf37a18-dirty'                                                                      
Failed to upload OVA: Docker invocation failed: /usr/bin/govc: The name 'bottlerocket-vmware-k8s-1.20-x86_64-v1.1.1-5bf37a18-dirty' already exists
```

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -64,6 +64,15 @@ REPO_METADATA_EXPIRING_WITHIN = "3 days"
 # You can use ALLOW_CLOBBER=true with the `ssm` task to make it overwrite existing values.
 # (This is not required with `promote-ssm` because the intent of promotion is overwriting.)
 
+# This can be overridden to provide a custom import spec for a VMware OVA.
+# Using configuration from Infra.toml, we substitute the correct value for
+# network, and whether or not to mark a VM as a template
+VMWARE_IMPORT_SPEC_PATH = "${BUILDSYS_ROOT_DIR}/tools/pubsys/support/vmware/import_spec.template"
+
+# You can set VMWARE_DATACENTERS to override the list of datacenters from
+# Infra.toml for VMware commands; it's a comma-separated list like
+# "datacenter1,datacenter2"
+
 # Disallow pulling directly Upstream URLs when lookaside cache results in MISSes as a fallback.
 # To use the upstream source as fallback, override this on the command line and set it to 'true'
 BUILDSYS_UPSTREAM_SOURCE_FALLBACK = "false"
@@ -119,6 +128,9 @@ BUILDSYS_KMOD_KIT_PATH = "${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_KMOD_KIT}"
 BUILDSYS_OVA = "${BUILDSYS_NAME_VARIANT}-v${BUILDSYS_VERSION_IMAGE}.ova"
 BUILDSYS_OVA_PATH = "${BUILDSYS_OUTPUT_DIR}/latest/${BUILDSYS_OVA}"
 BUILDSYS_OVF_TEMPLATE = "${BUILDSYS_ROOT_DIR}/variants/${BUILDSYS_VARIANT}/template.ovf"
+
+# The default name of uploaded OVAs; override by setting VMWARE_VM_NAME
+VMWARE_VM_NAME_DEFAULT = "${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH}-v${BUILDSYS_VERSION_IMAGE}-${BUILDSYS_VERSION_BUILD}"
 
 [tasks.setup]
 script = [
@@ -936,6 +948,56 @@ pubsys \
    ${PUBLISH_REGIONS:+--regions "${PUBLISH_REGIONS}"}
 '''
 ]
+
+[tasks._upload-ova-base]
+# Rather than depend on "build", which currently rebuilds images each run, we
+# depend on publish-tools and check for the image files below to save time.
+# This does mean that `cargo make` must be run before 
+# `cargo make _upload-ova-base`.
+dependencies = ["setup-build", "publish-tools"]
+script_runner = "bash"
+script = [
+'''
+set -e
+
+export PATH="${BUILDSYS_TOOLS_DIR}/bin:${PATH}"
+
+ova_path="${BUILDSYS_OUTPUT_DIR}/${BUILDSYS_NAME_FULL}.ova"
+if [ ! -s "${ova_path}" ]; then
+   echo "An OVA doesn't exist for the current version/commit - ${BUILDSYS_VERSION_FULL} - please run 'cargo make'" >&2
+   exit 1
+fi
+
+vm_name="${VMWARE_VM_NAME:-${VMWARE_VM_NAME_DEFAULT}}"
+
+pubsys \
+   --infra-config-path "${PUBLISH_INFRA_CONFIG_PATH}" \
+   \
+   upload-ova \
+   \
+   --ova "${ova_path}" \
+   --spec "${VMWARE_IMPORT_SPEC_PATH}" \
+   --name "${vm_name}" \
+   \
+   ${MARK_OVA_AS_TEMPLATE:+--mark-as-template} \
+   \
+   ${VMWARE_DATACENTERS:+--datacenters "${VMWARE_DATACENTERS}"}
+'''
+]
+
+# This task runs `_upload-ova-base` which will upload the OVA and *not* mark it
+# as a template
+[tasks.upload-ova]
+script_runner = "bash"
+extend = "_upload-ova-base"
+
+# This task runs `_upload-ova-base` with the environment variable
+# `MARK_OVA_AS_TEMPLATE` set, which will upload the OVA *and* mark it as a
+# template
+[tasks.vmware-template]
+script_runner = "bash"
+env = { "MARK_OVA_AS_TEMPLATE" = "true" }
+extend = "_upload-ova-base"
 
 [tasks.clean]
 dependencies = [
--- a/PUBLISHING-VMWARE.md
+++ b/PUBLISHING-VMWARE.md
@@ -1,0 +1,75 @@
+# Publishing a Bottlerocket OVA on VMware
+
+This guide will walk through some VMware specific details around making your OVA available as a VM or VM template in one or more software defined datacenters.
+
+### Configuration details
+
+As mentioned in the [PUBLISHING](PUBLISHING.md) guide, the process uses a configuration file called `Infra.toml`.
+For VMware, you can specify details about your various vSphere instances and datacenters in `Infra.toml`, as well as configuration that may be common between datacenters.
+
+It's important to note that we use [`govc`](https://github.com/vmware/govmomi/tree/master/govc) under the hood for interactions with vSphere, so at runtime **all datacenter configuration in `Infra.toml` is overridden by `GOVC_` environment variables.**
+`govc` is run in a container, so you do not need to install it on your machine.
+We first check for environment variables, then use `Infra.toml` for datacenter specific configuration, and finally common configuration.
+The following `GOVC_` environment variables are supported:
+* `GOVC_URL`
+* `GOVC_DATACENTER`
+* `GOVC_DATASTORE`
+* `GOVC_NETWORK`
+* `GOVC_FOLDER`
+* `GOVC_RESOURCE_POOL`
+* `GOVC_USERNAME`
+* `GOVC_PASSWORD`
+
+Credentials for your various datacenters may be stored at `~/.config/pubsys/vsphere-credentials.toml`.
+The format of the file is below; each datacenter gets its own `[datacenter.NAME]` block, where `NAME` corresponds to a datacenter name in `Infra.toml`
+Similar to other datacenter configuration, at runtime we first check for the environment variables `GOVC_USERNAME` and `GOVC_PASSWORD` and use one or both of them if they are set.
+
+```toml
+[datacenter.foo]
+username = "username"
+password = "password"
+
+[datacenter.bar]
+username = "bar"
+password = "baz"
+```
+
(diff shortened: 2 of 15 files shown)


Makefile.toml
```diff
@@ -926,6 +938,50 @@ pubsys \
 '''
 ]
 
+[tasks._upload-ova-base]
+# Rather than depend on "build", which currently rebuilds images each run, we
+# depend on publish-tools and check for the image files below to save time.
+# This does mean that `cargo make` must be run before `cargo make ami`.
```
webern: ?

```suggestion
# This does mean that `cargo make` must be run before `cargo make _upload-ova-base`.
```

Makefile.toml
```diff
@@ -64,6 +64,15 @@ REPO_METADATA_EXPIRING_WITHIN = "3 days"
...
 # (This is not required with `promote-ssm` because the intent of promotion is overwriting.)
 
+# This can be overridden to provide a custom import spec for a VMware OVA.
+# Using configuration from Infra.toml, we substitute the correct value for
+# network, and whether or not to mark a VM as a template
+VMWARE_IMPORT_SPEC_PATH = "${BUILDSYS_ROOT_DIR}/tools/pubsys/support/vmware/import_spec.template"
+
+# You can set VMWARE_DATACENTERS to override the list of datacenters from
+# Infra.toml for VMware commands; it's a comma-separated list like
+# "datacenter1,datacenter2"
```
webern, line 74: Is this comment in the right place?

Makefile.toml
```diff
@@ -926,6 +938,50 @@ pubsys \
...
+'''
+]
+
+[tasks.upload-ova]
+script_runner = "bash"
+extend = "_upload-ova-base"
+
+[tasks.vmware-template]
+script_runner = "bash"
+env = { "MARK_OVA_AS_TEMPLATE" = "true" }
```
webern, line 999: I think this option should be explained with a comment above both upload-ova and vmware-template. It took me a minute to understand what was going on.

tools/pubsys/src/vmware/govc.rs
```diff
@@ -0,0 +1,178 @@
...
+            &format!(
+                "type=bind,source={},target={},readonly",
+                ova_host_path.display(),
+                ova_container_path
+            ),
+        ];
+
+        // govc import.ova -options=/path/to/spec -name bottlerocket_vm_name /path/to/ova
+        let govc_cmd = &[
+            Self::GOVC,
```
webern, line 80: I think it's too strict to require that govc be located at `/usr/bin/govc`. I'm thinking you should just call `govc` and it needs to be in the `PATH`. That could be augmented with a parameter that allows you to specify the path to govc.

Regardless, I think the most important thing is that the error message be super clear that govc is a required binary and that we didn't find it where we expected it.

I was actually thinking that a check should be performed first so that a custom error message can be printed when govc isn't being found. As it is right now I think you might get something generic that could be confusing.

tools/pubsys/src/vmware/govc.rs
```diff
@@ -0,0 +1,178 @@
...
+        .context(error::CommandStart)?;
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    if output.status.success() {
+        trace!("{}", stdout);
+        return Ok(output);
+    } else {
+        trace!("{}", stdout);
+        error::Docker { output: stdout }.fail()
+    }
```
webern:
```suggestion
    let stdout = String::from_utf8_lossy(&output.stdout);
    trace!("{}", stdout);
    if output.status.success() {
        return Ok(output);
    } else {
        error::Docker { output: stdout }.fail()
    }
```

tools/pubsys/src/vmware/upload_ova/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+        !upload_datacenters.is_empty(),
+        error::MissingConfig {
+            missing: "vmware.datacenters"
+        }
+    );
+
+    // Retrieve credentials from GOVC_ environment variables
+    let creds_env = DatacenterCredsBuilder::from_env();
+    // Retrieve credentials from file. The `home` crate is used to construct the VMWARE_CREDS_PATH,
+    // and it's possible (however unlikely) that it is unable to determine the user;s home folder.
```
webern:
```suggestion
    // and it's possible (however unlikely) that it is unable to determine the user's home folder.
```

webern (commented):
saving partial review

tools/pubsys/src/vmware/govc.rs
zmrow: Good point - I think we should just assume that `govc` is in the path.  The error message we'll get is pretty clear (I got it initially when I was testing this with a container that didn't have `govc` :) )

Makefile.toml
zmrow: Will do

Makefile.toml
zmrow: Indeed - It closely mimics the `AWS_REGIONS` comment a few lines up.

tools/pubsys/src/vmware/upload_ova/mod.rs
zmrow: That pesky little finger just couldn't reach over one more key... :)

PUBLISHING-VMWARE.md
```diff
@@ -0,0 +1,65 @@
+# Publishing a Bottlerocket OVA on VMware
+
+This guide will walk through some VMware specific details around making your OVA available as a VM or VM template in one or more software defined datacenters.
+
+### Configuration details
+
+As mentioned in the [PUBLISHING](PUBLISHING.md) guide, the process uses a configuration file called `Infra.toml`.
+For VMware, you can specify details about your various vSphere instances and datacenters in `Infra.toml`, as well as configuration that may be common between datacenters.
+
+It's important to note that we use [`govc`](https://github.com/vmware/govmomi/tree/master/govc) under the hood for interactions with vSphere, so at runtime **all datacenter configuration in `Infra.toml` is overridden by `GOVC_` environment variables.**
```
webern, line 10: This should probably also explain that it needs to be installed. Reading this one might think that you are using it as a library.

tools/pubsys-config/src/vmware.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub datacenter: String,
+    pub datastore: String,
+    pub network: String,
+    pub folder: String,
+    pub resource_pool: String,
+}
+
+/// Attempts to create a `Datacenter` from a `DatacenterBuilder` by ensuring that each
+/// field contains a value.
+impl TryFrom<DatacenterBuilder> for Datacenter {
```
webern: I think it might be more canonical for DatacenterBuilder (and CredsBuilder) to have:

```rust
    fn build(self) -> Result<Datacenter> {
        todo!();
    }
```

PUBLISHING-VMWARE.md
zmrow: We don't need to install it on the host since we're running it in the SDK container.  But it's probably a good note to add anyway.

tools/pubsys-config/src/vmware.rs
zmrow: I can see that.  Interested to hear some additional opinions.  It's not really a traditional builder per se, which is one of the reasons I leaned away from that.

zmrow:
^ addresses @webern 's comments!

PUBLISHING-VMWARE.md
webern: Oh, right.

tools/pubsys/src/vmware/govc.rs
webern: I had forgotten that we were running it via the SDK. But the change to expect it in PATH still makes sense.

tools/pubsys-config/src/vmware.rs
etungsten: Looking at the call sites, I do think it's slightly confusing to implement the `try_from` trait for this purpose. It does seem like a very unique use-case.

I wonder if it's maybe simpler to just keep it to a single struct-type and just have a single `new` method that tries to create the config from multiple sources.

```
impl Datacenter {
    pub fn new(vmware_config: Option<..>, vmware_common: Option<...>) -> Result<Self> {
        // Try getting the config from environmental variables first
        todo!();
        // Try getting the config from infra.toml vmware config
        todo!();
        // Try getting the config from infra.toml vmware common config
        todo!();
        // If any config fields are still missing, return error MissingConfig
        todo!();
        
        Ok(
            Self {
                ...
            }
        )
}
```

etungsten (approved):

tools/pubsys-config/src/vmware.rs
zmrow: :)  This config piece is all a bit tricky and I spent a ton of time working through different ways to do it.  

What you suggested was my original idea but gathering the env/common multiple times (for multiple datacenters) seems a bit inefficient.  I suppose `Datacenter` could have a `new()` method that has arguments for env/common?  And we would call it for each datacenter in the config? I suppose in that case we'd end up using mostly the same logic as the `TryFrom` in order to get the final `Datacenter` since all the fields need to be present.

If most people are against the `TryFrom` implementation, I'm leaning towards the `build()` method as a solution.

zmrow:
^ Removes the `TryFrom` implementation in favor of a `build()` method on the `*Builder` structs.

webern (approved):

zmrow:
^ Above push is a rebase on `develop`.

Now the the SDK changes are shipped, I executed all of the testing again and everything looks good.  (Fighting through the intermittent VMC issues today)

Pulling out of draft...

etungsten (approved):

webern (approved):

PUBLISHING-VMWARE.md
arnaldo2792: This line makes me think that whatever `GOVC_` environment variable I set will be available in the SDK container. But in reality what is happening is that we reference a set of `GOVC_*` environment variables, and make them available in the container. I think we should specify what are the actual `GOVC_` environment variables we are referencing, and also the correspondent mapping in `Infra.toml`, since there is at least one configuration that doesn't match the `GOVC_` convention:

```toml
vsphere_url = "https://vcenter.1234.vmwarevmc.com" # GOVC_URL
```

tools/pubsys/src/vmware/govc.rs
```diff
@@ -0,0 +1,177 @@
...
+        .stderr_to_stdout()
+        .stdout_capture()
+        .unchecked()
+        .run()
+        .context(error::CommandStart)?;
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    trace!("{}", stdout);
+    if output.status.success() {
+        return Ok(output);
```
arnaldo2792: Nit:

```suggestion
        Ok(output)
```

zmrow:
^ Addresses @arnaldo2792 's comments :)

arnaldo2792 (approved):
:boat:

Issue #1628, opened 2021-06-23
Unexpected connection refused when running fluent-bit on 1.20 cluster


gthao313:
**Image I'm using:**
ami-00e7ac29da78d893c

**What I expected to happen:**
Successfully connect to port and fluent-bit starts to work
Example:
```
sh-4.2# while curl localhost:2020; do echo pass; done
{"fluent-bit":{"version":"1.7.9","edition":"Community","flags":["FLB_HAVE_PARSER","FLB_HAVE_RECORD_ACCESSOR","FLB_HAVE_STREAM_PROCESSOR","FLB_HAVE_TLS","FLB_HAVE_OPENSSL","FLB_HAVE_AWS","FLB_HAVE_SIGNV4","FLB_HAVE_SQLDB","FLB_HAVE_METRICS","FLB_HAVE_HTTP_SERVER","FLB_HAVE_SYSTEMD","FLB_HAVE_FORK","FLB_HAVE_TIMESPEC_GET","FLB_HAVE_GMTOFF","FLB_HAVE_UNIX_SOCKET","FLB_HAVE_PROXY_GO","FLB_HAVE_JEMALLOC","FLB_HAVE_LIBBACKTRACE","FLB_HAVE_REGEX","FLB_HAVE_UTF8_ENCODER","FLB_HAVE_LUAJIT","FLB_HAVE_C_TLS","FLB_HAVE_ACCEPT4","FLB_HAVE_INOTIFY"]}}pass
```

**What actually happened:**
```
Failed to connect to localhost port 2020: Connection refused
```

**How to reproduce the problem:**
1. Setting up a 1.20 cluster in eu-central-1
2. Launching bottlerocket instance
3. ```kubectl run --image public.ecr.aws/aws-observability/aws-for-fluent-bit:2.16.0 --command -- sh```
4. ```kubectl exec -ti sh -- sh```
5. ```/fluent-bit/bin/fluent-bit -H -i cpu -o null &```
6. ```while curl localhost:2020; do echo pass; done```

snstanton:
Do you have any idea what might be going on?  This is blocking our cluster rollout on 1.20.  Any suggestions for a workaround would be great.  Otherwise, we're looking at having to roll back the cluster to 1.19.

jhaynes:
We don't have a root cause yet but our on-call @gthao313 is working on figuring out the cause and a work around.

snstanton:
[...]

PR #1638, opened 2021-06-29
kubelet: add setting for configuring cpuManagerPolicy

gthao313:
**Issue number:**
Part of #1628

**Description of changes:**
Fluent-bit does not work with `cpuPolicyManager: static`, and it does work with `cpuPolicyManager: none`. We expose `kubernetes.cpu-manager-policy` as a setting, so customers can specify `cpuManagerPolicy` to `none` and not be blocked by this issue while running fluent bit.

- [x] expose `kubernetes.cpu-manager-policy` as a setting
- [x] expose `kubernetes.cpu-manager-reconcile-policy` as a setting
- [x] add migration for cpu-manager-manager-policy - deal with the issue that changing the CPU manager policy caused issues on upgrade/downgrade ( from none to static and from static to none)
- [x] add migration for cpu-manager-reconcile-period

#### Solution about deal with the issue that changing the CPU manager policy caused issues on upgrade/downgrade
error
```
2028 cpu_manager.go:203] [cpumanager] could not initialize checkpoint manager: could not restore state from checkpoint: configured policy "static" differs from state checkpoint policy "none", please drain this node and delete the CPU manager checkpoint file "/var/lib/kubelet/cpu_manager_state" before restarting Kubelet, please drain node and remove policy state file
```

The reason why upgrade/downgrade of CPU manager policy migration is configured policy "static" differs from state checkpoint policy "none". Therefore, I use migration tool to delete cpu-manager-policy checkpoint for every migration process. No orchestrated workload would be running when the migrator is running, so it should be safe to delete that file before kubelet starts.

**Testing done:**

- [x] Test behavioral effect of new setting `cpu-manager-policy`
- [x] Test behavioral effect of new setting `cpu-manager-reconcile-policy`
- [x] Test migration process - `cpu-manager-policy`
- [x] Test migration process - `cpu-manager-reconcile-policy`

#### `cpu-manager-policy`
launching instance with the AMI which contains new setting and check if `cpu-manager-policy` has been configured expectantly. Meanwhile,  configure it as `none` and test if fluent-bit pod work as expected.

Step1 check if `cpu-manager-policy` has been configured expectantly:

Test with `cpu-manager-policy= "none"`
```
1726 policy_none.go:43] [cpumanager] none policy: Start
```
Test with default ("none")
```
1737 cpu_manager.go:193] [cpumanager] starting with none policy
```
Test with `cpu-manager-policy= "static"`
```
1723 cpu_manager.go:193] [cpumanager] starting with static policy
```

Step2 configure it as `none` and test if fluent-bit pod work as expected:
1. kubectl run --image public.ecr.aws/aws-observability/aws-for-fluent-bit:2.16.0 --command -- sh
2. kubectl exec -ti sh -- sh
3. /fluent-bit/bin/fluent-bit -H -i cpu -o null 
4. while curl localhost:2020; do echo pass; done

result (web server works perfectly ):
```
COMMAND   PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
fluent-bi   1 root   19u  IPv4  20229      0t0  TCP *:24224 (LISTEN)
fluent-bi  36 root   30u  IPv4  19367      0t0  TCP *:2020 (LISTEN)
```

#### Migration test:
upgrade
Step1: Upgrade to v1.1.3
```
bash-5.0# updog check-update -a --json
[
  {
    "variant": "aws-k8s-1.20",
    "arch": "x86_64",
    "version": "1.1.3",
    "max_version": "1.1.3",
.....

```
cat var/lib/kubelet/cpu_manager_state
{"policyName":"static","defaultCpuSet":"0-3","checksum":611748604}

```
bash-5.0# updog update -i 1.1.3 -r -n
Starting update to 1.1.3
```

Step2: check if cpu-manager-policy works as expected after upgrading
```
cat var/lib/kubelet/cpu_manager_state
{"policyName":"none","defaultCpuSet":"","checksum":1353318690}
```

```
journalctl - u kubelet
1701 policy_none.go:43] [cpumanager] none policy: Start
```

Step3: Specify new setting `cpu-manage-policy`through control container
```
apiclient set -j '{"kubernetes": {"cpu-manager-policy": "none"}}'
```
```
bash-5.0# cat /var/lib/bottlerocket/datastore/current/live/settings/kubernetes/cpu-manager-policy
"none"
```

downgrade

Step1: Check migration binary
```
ls -al /var/lib/bottlerocket-migrations
total 492
drwx------.  2 root root   4096 Jun 29 21:51 .
drwxr-xr-x. 14 root root   4096 Jun 29 22:22 ..
-rw-r--r--.  1 root root 488744 Jun 29 22:21 0b168f448fed442fc3be61059256d4c12a423aa6afba5ec825596522dd015d41.migrate_v1.1.3_cpu-manager-policy.lz4

```

Step2: Downgrade to previous verison
```
signpost rollback-to-inactive
reboot
```

Step3: check if cpu-manager-policy works as expected after downgrading
```
cat var/lib/kubelet/cpu_manager_state
{"policyName":"static","defaultCpuSet":"0-3","checksum":611748604}
```

```
journalctl - u kubelet
1700 cpu_manager.go:193] [cpumanager] starting with static policy
```

Step4: Check if `cpu-manager-policy` have been removed
```
cat /var/lib/bottlerocket/datastore/current/live/settings/kubernetes/cpu-manager-policy
cat: /var/lib/bottlerocket/datastore/current/live/settings/kubernetes/cpu-manager-policy: No such file or directory
```

#### `cpu-manager-reconcile-period`
launching instance with the AMI which contains new setting and check if `cpu-manager-reconcile-period` has been configured expectantly.

Step1 check if `cpu-manager-reconcile-period` has been configured expectantly:

Test with `cpu-manager-reconcile-period = 100s`
```
cpu_manager.go:194] [cpumanager] reconciling every 100s
```
Test with `cpu-manager-reconcile-period = 10m`
```
1730 cpu_manager.go:194] [cpumanager] reconciling every 10m0s
```
Test with `cpu-manager-reconcile-period = 1h`
```
1736 cpu_manager.go:194] [cpumanager] reconciling every 1h0m0s
```
Test with `cpu-manager-reconcile-period = 1h0m20s`
```
cpu_manager.go:194] [cpumanager] reconciling every 1h0m20s
```
Test with `cpu-manager-reconcile-period = 10m20s`
```
1726 cpu_manager.go:194] [cpumanager] reconciling every 10m20s
```
Test with `cpu-manager-reconcile-period = 1h10m2s`
```
1727 cpu_manager.go:194] [cpumanager] reconciling every 1h10m2s
```
#### Migration test:
upgrade
Step1: Upgrade to v1.1.3
```
bash-5.0# updog check-update -a --json
[
  {
    "variant": "aws-k8s-1.20",
    "arch": "x86_64",
    "version": "1.1.3",
    "max_version": "1.1.3",
.....

bash-5.0# updog update -i 1.1.3 -r -n
Starting update to 1.1.3
```
Step2: Specify new setting `cpu-manager-reconcile-period`through control container
```
apiclient set -j '{"kubernetes": {"cpu-manager-reconcile-period": "10m"}}'
apiclient set -j '{"kubernetes": {"cpu-manager-policy": "none"}}'
```

```
bash-5.0# cat /var/lib/bottlerocket/datastore/current/live/settings/kubernetes/cpu-manager-reconcile-period
"10m"
cat /var/lib/bottlerocket/datastore/current/live/settings/kubernetes/cpu-manager-policy
"none"
```

```
journalctl -u kubelet
cpu_manager.go:194] [cpumanager] start with none policy
cpu_manager.go:194] [cpumanager] reconciling every 10m0s
```

downgrade
Step1: Check migration binary
```
ls -al /var/lib/bottlerocket-migrations
0b168f448fed442fc3be61059256d4c12a423aa6afba5ec825596522dd015d41.migrate_v1.1.3_cpu-manager-reconcile-period.lz4
0b168f448fed442fc3be61059256d4c12a423aa6afba5ec825596522dd015d41.migrate_v1.1.3_cpu-manager-policy.lz4

```
Step2: Downgrade to previous verison
```
signpost rollback-to-inactive
reboot
```

Step3: Check if `cpu-manager-policy` have been removed
```
cat /var/lib/bottlerocket/datastore/current/live/settings/kubernetes/cpu-manager-reconcile-period
cat /var/lib/bottlerocket/datastore/current/live/settings/kubernetes/cpu-manager-reconcile-period: No such file or directory
cat /var/lib/bottlerocket/datastore/current/live/settings/kubernetes/cpu-manager-policy
cat /var/lib/bottlerocket/datastore/current/live/settings/kubernetes/cpu-manager-policy: No such file or directory
```

```
journalctl -u kubelet
cpu_manager.go:194] [cpumanager] start with static policy
1702 cpu_manager.go:194] [cpumanager] reconciling every 10s
```

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,6 +359,7 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
 * `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`. If you want to allow pods with certain resource characteristics to be granted increased CPU affinity and exclusivity on the node, you can set this setting to `static`. You should reboot if you change this setting after startup - try `apiclient reboot`.
+* `settings.kubernetes.cpu-manager-reconcile-period`: Specifies the CPU manager reconcile period, which controls how often updated CPU assignments are written to cgroupfs. The value is a duration like `30s` for 30 seconds or `1h5m` for 1 hour and 5 minutes.
 
 You can also optionally specify static pods for your node with the following settings.
 Static pods can be particularly useful when running in standalone mode.
--- a/README.md
+++ b/README.md
@@ -358,6 +358,7 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
+* `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`. If you want to allow pods with certain resource characteristics to be granted increased CPU affinity and exclusivity on the node, you can set this setting to `static`. You should reboot if you change this setting after startup - try `apiclient reboot`.
 
 You can also optionally specify static pods for your node with the following settings.
 Static pods can be particularly useful when running in standalone mode.
--- a/packages/kubernetes-1.16/kubelet-config
+++ b/packages/kubernetes-1.16/kubelet-config
@@ -75,6 +75,9 @@ systemReserved:
   {{~/each}}
 {{~/if}}
 cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
+{{~#if settings.kubernetes.cpu-manager-reconcile-period}}
+cpuManagerReconcilePeriod: {{settings.kubernetes.cpu-manager-reconcile-period}}
+{{~/if}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 readOnlyPort: 0
--- a/packages/kubernetes-1.16/kubelet-config
+++ b/packages/kubernetes-1.16/kubelet-config
@@ -74,6 +74,7 @@ systemReserved:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
+cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 readOnlyPort: 0
--- a/packages/kubernetes-1.20/kubelet-config
+++ b/packages/kubernetes-1.20/kubelet-config
@@ -75,6 +75,9 @@ systemReserved:
   {{~/each}}
 {{~/if}}
 cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
+{{~#if settings.kubernetes.cpu-manager-reconcile-period}}
+cpuManagerReconcilePeriod: {{settings.kubernetes.cpu-manager-reconcile-period}}
+{{~/if}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 readOnlyPort: 0
--- a/packages/kubernetes-1.20/kubelet-config
+++ b/packages/kubernetes-1.20/kubelet-config
@@ -74,7 +74,7 @@ systemReserved:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
-cpuManagerPolicy: "static"
+cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 readOnlyPort: 0
--- a/packages/kubernetes-1.21/kubelet-config
+++ b/packages/kubernetes-1.21/kubelet-config
@@ -75,6 +75,9 @@ systemReserved:
   {{~/each}}
 {{~/if}}
 cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
+{{~#if settings.kubernetes.cpu-manager-reconcile-period}}
+cpuManagerReconcilePeriod: {{settings.kubernetes.cpu-manager-reconcile-period}}
+{{~/if}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 readOnlyPort: 0
--- a/packages/kubernetes-1.21/kubelet-config
+++ b/packages/kubernetes-1.21/kubelet-config
@@ -74,7 +74,7 @@ systemReserved:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
-cpuManagerPolicy: "static"
+cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 readOnlyPort: 0
--- a/sources/api/migration/migration-helpers/src/error.rs
+++ b/sources/api/migration/migration-helpers/src/error.rs
@@ -1,6 +1,7 @@
 //! Contains the Error and Result types used by the migration helper functions and migrations.
 
 use snafu::Snafu;
+use std::path::PathBuf;
 
 /// Error contains the errors that can happen in the migration helper functions and in migrations.
 #[derive(Debug, Snafu)]
@@ -98,6 +99,12 @@ pub enum Error {
         setting: String,
         metadata: String,
         data: Vec<serde_json::Value>,
+    },
+
+    #[snafu(display("Failed to delete file '{}': '{}'", path.display(), source))]
+    RemoveFile {
+        path: PathBuf,
+        source: std::io::Error,
     },
 }
 
--- a/sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
+++ b/sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
@@ -1,0 +1,54 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::{error, migrate, Migration, MigrationData, Result};
+use snafu::ResultExt;
+use std::fs;
+use std::io;
+use std::process;
+
+const CPU_MANAGER_POLICY_CHECKPOINT: &str = "/var/lib/kubelet/cpu_manager_state";
+
+/// forward - We always remove the state file on boot, therefore we don't need to explicitly
+/// remove the file during forward migration.
+/// backward - We remove cpu manager policy checkpoint value on downgrade, since older versions did not
+/// clean up this state file on boot.
+pub struct CpuManagerPolicyCleaner;
+
+impl Migration for CpuManagerPolicyCleaner {
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        println!("CpuManagerPolicyCleaner has no work to do on upgrade.");
+        Ok(input)
+    }
+
+    fn backward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        // removing existing cpu_manager_policy_state file
+        println!(
+            "Deleting existing cpu manager policy checkpoint: '{}'",
+            CPU_MANAGER_POLICY_CHECKPOINT
+        );
+        if let Err(e) = fs::remove_file(CPU_MANAGER_POLICY_CHECKPOINT) {
+            if e.kind() != io::ErrorKind::NotFound {
+                return Err(e).context(error::RemoveFile {
+                    path: CPU_MANAGER_POLICY_CHECKPOINT,
+                });
+            } else {
+                println!("NotFound: '{}'", CPU_MANAGER_POLICY_CHECKPOINT)
+            }
+        }
+        Ok(input)
+    }
+}
+/// We changed the default for CPU manager policy and need to handle kubelet's state file.
+fn run() -> Result<()> {
+    migrate(CpuManagerPolicyCleaner)
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -113,7 +113,7 @@ use std::collections::HashMap;
 use std::net::Ipv4Addr;
 
 use crate::modeled_types::{
-    BootstrapContainerMode, DNSDomain, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue,
+    BootstrapContainerMode, CpuManagerPolicy, DNSDomain, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue,
     FriendlyVersion, Identifier, KubernetesAuthenticationMode, KubernetesBootstrapToken,
     KubernetesCloudProvider, KubernetesClusterName, KubernetesEvictionHardKey, KubernetesLabelKey,
     KubernetesLabelValue, KubernetesQuantityValue, KubernetesReservedResourceKey,
@@ -158,6 +158,7 @@ struct KubernetesSettings {
     kube_api_burst: i32,
     container_log_max_size: KubernetesQuantityValue,
     container_log_max_files: i32,
+    cpu_manager_policy: CpuManagerPolicy,
 
     // Settings where we generate a value based on the runtime environment.  The user can specify a
     // value to override the generated one, but typically would not.
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -115,7 +115,7 @@ use std::net::Ipv4Addr;
 use crate::modeled_types::{
     BootstrapContainerMode, CpuManagerPolicy, DNSDomain, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue,
     FriendlyVersion, Identifier, KubernetesAuthenticationMode, KubernetesBootstrapToken,
-    KubernetesCloudProvider, KubernetesClusterName, KubernetesEvictionHardKey, KubernetesLabelKey,
+    KubernetesCloudProvider, KubernetesClusterName, KubernetesDurationValue, KubernetesEvictionHardKey, KubernetesLabelKey,
     KubernetesLabelValue, KubernetesQuantityValue, KubernetesReservedResourceKey,
     KubernetesTaintValue, KubernetesThresholdValue, Lockdown, SingleLineString, SysctlKey, Url,
     ValidBase64,
@@ -159,6 +159,7 @@ struct KubernetesSettings {
     container_log_max_size: KubernetesQuantityValue,
     container_log_max_files: i32,
     cpu_manager_policy: CpuManagerPolicy,
+    cpu_manager_reconcile_period: KubernetesDurationValue,
 
     // Settings where we generate a value based on the runtime environment.  The user can specify a
     // value to override the generated one, but typically would not.
--- a/sources/models/src/modeled_types/kubernetes.rs
+++ b/sources/models/src/modeled_types/kubernetes.rs
@@ -817,3 +817,76 @@ mod test_cpu_manager_policy {
         }
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesDurationValue represents a string that contains a valid Kubernetes duration value.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesDurationValue {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref KUBERNETES_DURATION_VALUE: Regex = Regex::new(
+        r"^(([0-9]+\.)?[0-9]+h)?(([0-9]+\.)?[0-9]+m)?(([0-9]+\.)?[0-9]+s)?(([0-9]+\.)?[0-9]+ms)?$"
+    )
+    .unwrap();
+}
+
+impl TryFrom<&str> for KubernetesDurationValue {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        ensure!(
+            !input.is_empty(),
+            error::InvalidKubernetesDurationValue { input }
+        );
+        ensure!(
+            KUBERNETES_DURATION_VALUE.is_match(input),
+            error::InvalidKubernetesDurationValue { input }
+        );
+        Ok(KubernetesDurationValue {
+            inner: input.to_string(),
+        })
+    }
+}
+
+string_impls_for!(KubernetesDurationValue, "KubernetesDurationValue");
+
+#[cfg(test)]
+mod test_kubernetes_duration_value {
+    use super::KubernetesDurationValue;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn good_tokens() {
+        for ok in &[
+            "9ms",
+            "99s",
+            "20m",
+            "1h",
+            "1h2m3s10ms",
+            "4m5s10ms",
+            "2h3s10ms",
+            "1.5h3.5m",
+        ] {
+            KubernetesDurationValue::try_from(*ok).unwrap();
+        }
+    }
+
+    #[test]
+    fn bad_names() {
+        for err in &[
+            "",
+            "100",
+            "...3ms",
+            "1..5s",
+            "ten second",
+            "1m2h",
+            "9ns",
+            &"a".repeat(23),
+        ] {
+            KubernetesDurationValue::try_from(*err).unwrap_err();
+        }
+    }
+}
--- a/sources/models/src/modeled_types/kubernetes.rs
+++ b/sources/models/src/modeled_types/kubernetes.rs
@@ -768,3 +768,52 @@ mod test_kubernetes_cloud_provider {
         }
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// CpuManagerPolicy represents a string that contains a valid cpu management policy. Default: none
+/// https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct CpuManagerPolicy {
+    inner: String,
+}
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum ValidCpuManagerPolicy {
+    Static,
+    None,
+}
+
+impl TryFrom<&str> for CpuManagerPolicy {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        serde_plain::from_str::<ValidCpuManagerPolicy>(&input).context(
+            error::InvalidCpuManagerPolicy { input })?;
+        Ok(CpuManagerPolicy {
+            inner: input.to_string(),
+        })
+    }
+}
+string_impls_for!(CpuManagerPolicy, "CpuManagerPolicy");
+
+#[cfg(test)]
+mod test_cpu_manager_policy {
+    use super::CpuManagerPolicy;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn good_cpu_manager_policy() {
+        for ok in &["static", "none"] {
+            CpuManagerPolicy::try_from(*ok).unwrap();
+        }
+    }
+
+    #[test]
+    fn bad_cpu_manager_policy() {
+        for err in &["", "bad", "100", &"a".repeat(64)] {
+            CpuManagerPolicy::try_from(*err).unwrap_err();
+        }
+    }
+}
--- a/sources/models/src/modeled_types/mod.rs
+++ b/sources/models/src/modeled_types/mod.rs
@@ -74,6 +74,12 @@ pub mod error {
             input: String,
             source: std::num::ParseFloatError,
         },
+
+        #[snafu(display("Invalid Cpu Manager policy '{}'", input))]
+        InvalidCpuManagerPolicy {
+            input: String,
+            source: serde_plain::Error,
+        },
     }
 }
 
--- a/sources/models/src/modeled_types/mod.rs
+++ b/sources/models/src/modeled_types/mod.rs
@@ -80,6 +80,9 @@ pub mod error {
             input: String,
             source: serde_plain::Error,
         },
+
+        #[snafu(display("Invalid Kubernetes duration value '{}'", input))]
+        InvalidKubernetesDurationValue { input: String },
     }
 }
 
--- a/Release.toml
+++ b/Release.toml
@@ -54,3 +54,6 @@ version = "1.1.2"
     "migrate_v1.1.2_admin-container-v0-7-1.lz4",
     "migrate_v1.1.2_control-container-v0-5-1.lz4",
 ]
+"(1.1.2, 1.1.3)" = [
+    "migrate_v1.1.3_kubelet-cpu-manager-state.lz4",
+]
--- a/Release.toml
+++ b/Release.toml
@@ -56,4 +56,5 @@ version = "1.1.2"
 ]
 "(1.1.2, 1.1.3)" = [
     "migrate_v1.1.3_kubelet-cpu-manager-state.lz4",
+    "migrate_v1.1.3_kubelet-cpu-manager.lz4",
 ]
--- a/packages/kubernetes-1.16/kubernetes-tmpfiles.conf
+++ b/packages/kubernetes-1.16/kubernetes-tmpfiles.conf
@@ -2,3 +2,4 @@ d /etc/kubernetes/static-pods - - - -
 L /etc/kubernetes/manifests - - - - static-pods
 d /var/lib/kubelet/providers/secrets-store - - - -
 L /etc/kubernetes/secrets-store-csi-providers  - - - - /var/lib/kubelet/providers/secrets-store
+r! /var/lib/kubelet/cpu_manager_state
--- a/packages/kubernetes-1.17/kubelet-config
+++ b/packages/kubernetes-1.17/kubelet-config
@@ -75,6 +75,9 @@ systemReserved:
   {{~/each}}
 {{~/if}}
 cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
+{{~#if settings.kubernetes.cpu-manager-reconcile-period}}
+cpuManagerReconcilePeriod: {{settings.kubernetes.cpu-manager-reconcile-period}}
+{{~/if}}
 resolvConf: "/etc/resolv.conf"
 hairpinMode: hairpin-veth
 readOnlyPort: 0
--- a/packages/kubernetes-1.17/kubelet-config
+++ b/packages/kubernetes-1.17/kubelet-config
@@ -74,6 +74,7 @@ systemReserved:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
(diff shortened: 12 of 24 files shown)


webern (approved):

sources/models/src/modeled_types/mod.rs
```diff
@@ -74,6 +74,12 @@ pub mod error {
             input: String,
             source: std::num::ParseFloatError,
         },
+
+        #[snafu(display("Invalid Cpu Manager policy '{}'", input))]
+        InvalideCpuManagerPolicy {
```
bcressey:
```suggestion
        InvalidCpuManagerPolicy {
```

README.md
```diff
@@ -358,6 +358,7 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
+* `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `static`.
```
bcressey: I'd like to default to "none" here given known incompatibilities with fluent-bit as it's deployed today. We can always change the default for future Kubernetes variants.

packages/kubernetes-1.20/kubelet-config
```diff
@@ -74,7 +74,7 @@ systemReserved:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
-cpuManagerPolicy: "static"
+cpuManagerPolicy: {{default "static" settings.kubernetes.cpu-manager-policy}}
```
bcressey: When we originally rolled this out, we reverted it because changing the CPU manager policy caused issues on upgrade.

There's some local state that will need to be fixed up either as part of the migration, or as a helper command to check for a difference between the on-disk value and the configured value. The latter might make this setting possible to change on a running system, if still not entirely safe.

bcressey (commented):
I'd like to add a setting for `cpu-manager-reconcile-period` while we're at it.

samjo-nyang:
Could you backport this option to k8s 1.18 and 1.19? I think the migration process will not make any problems if you set "none" as default value.

gthao313:
Push above 

- [x] add a new setting `cpu-manager-reconcile-period` and it migration.
- [x] add a custom migration for `cpu-manger-policy` to deal with the issue that changing the CPU manager policy caused issues on upgrade/downgrade
- [x] Default  `cpu-manger-policy` to `none`.
- [x] Support `cpu-manger-policy` on 1.16, 1.17, 1.18 and 1.19 variant.

`cpu-manger-policy` migration logic: The reason why upgrade/downgrade of CPU manager policy migration is configured policy "static" differs from state checkpoint policy "none". Therefore, I use migration tool to delete `cpu-manager-policy checkpoint` during migration process. No orchestrated workload would be running when the migrator is running, so it should be safe to delete that file before kubelet starts.
```
2028 cpu_manager.go:203] [cpumanager] could not initialize checkpoint manager: could not restore state from checkpoint: configured policy "static" differs from state checkpoint policy "none", please drain this node and delete the CPU manager checkpoint file "/var/lib/kubelet/cpu_manager_state" before restarting Kubelet, please drain node and remove policy state file

```

etungsten (approved):

packages/kubernetes-1.20/kubelet-config
```diff
@@ -74,7 +74,10 @@ systemReserved:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
-cpuManagerPolicy: "static"
+cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
```
webern, line 77: Are we intentionally changing the default to "none"? I thought we would expose the setting but otherwise leave the default unchanged?

```suggestion
cpuManagerPolicy: {{default "static" settings.kubernetes.cpu-manager-policy}}
```

packages/kubernetes-1.21/kubelet-config
```diff
@@ -74,7 +74,10 @@ systemReserved:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
-cpuManagerPolicy: "static"
+cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
```
webern, line 77: Same question as above.

```suggestion
cpuManagerPolicy: {{default "static" settings.kubernetes.cpu-manager-policy}}
```

sources/api/migration/migrations/v1.1.3/cpu-manager-policy/src/main.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        // removing existing cpu_manager_policy_state file
+        if Path::new(CPU_MANAGER_POLICY_CHECKPOINT).exists() {
+            println!(
+                "Deleting existing cpu manager policy checkpoint: '{}'",
+                CPU_MANAGER_POLICY_CHECKPOINT
+            );
+            fs::remove_file(CPU_MANAGER_POLICY_CHECKPOINT).context(error::RemoveFile {
+                path: CPU_MANAGER_POLICY_CHECKPOINT,
+            })?;
```
webern: I don't think the program should exit with error if the file does not exist. I think the right way to do this is to check the std::io::ErrorKind for NotFound and maybe println! if the ErrorKind is NotFound, otherwise error out. (Same for the `backward` function below as well)

https://doc.rust-lang.org/std/fs/fn.remove_file.html
https://doc.rust-lang.org/std/io/enum.ErrorKind.html

packages/kubernetes-1.20/kubelet-config
gthao313: We currently default it to `none` here given known incompatibilities with fluent-bit as it's deployed today. We can change the default  to `static` in future once fluent-bit bug has been fixed.

sources/api/migration/migrations/v1.1.3/cpu-manager-policy/src/main.rs
gthao313: Sure. I'll update that! thanks

gthao313:
Push above improving error handle for `fileremove` according to Matt's idea.

packages/kubernetes-1.20/kubelet-config
webern: We heard from @bcressey out-of-band that he _does_ want to change the default back to `none` so I am resolving this issue and approving.

packages/kubernetes-1.21/kubelet-config
webern: We heard from @bcressey out-of-band that he _does_ want to change the default back to `none` so I am resolving this issue and approving.

webern (approved):

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -768,3 +768,102 @@ mod test_kubernetes_cloud_provider {
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct CpuManagerReconcilePeriod {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref CPU_MANAGER_RECONCILE_PERIOD: Regex = Regex::new(
+        r"^(([0-9.]+)(h))?(([0-9.]+)(m))?(([0-9.]+)(s))?$"
+    ).unwrap();
+}
```
bcressey: I'd call this type `KubernetesDurationValue` since there are several other settings that could use this same validation logic.

sources/api/migration/migrations/v1.1.3/cpu-manager-policy/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use migration_helpers::{error, migrate, Migration, MigrationData, Result};
+use snafu::ResultExt;
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::process;
+
+const SETTING: &str = "settings.kubernetes.cpu-manager-policy";
+const CPU_MANAGER_POLICY_CHECKPOINT: &str = "var/lib/kubelet/cpu_manager_state";
```
bcressey:
```suggestion
const CPU_MANAGER_POLICY_CHECKPOINT: &str = "/var/lib/kubelet/cpu_manager_state";
```

sources/api/migration/migrations/v1.1.3/cpu-manager-policy/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+        Ok(input)
+    }
+
+    fn backward(&mut self, mut input: MigrationData) -> Result<MigrationData> {
+        // removing cpu-manager-policy setting, which old version doesn't know about it.
+        if let Some(data) = input.data.remove(SETTING) {
+            println!("Removed {}, which was set to '{}'", SETTING, data);
+        } else {
+            println!("Found no {} to remove", SETTING);
+        }
```
bcressey: I'd prefer to handle removing the setting together with `cpu-manager-reconcile-period`, and leave this migration purely for the unusual checkpoint fix.

Maybe name them:
* kubelet-cpu-manager
* kubelet-cpu-manager-state

sources/api/migration/migrations/v1.1.3/cpu-manager-policy/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+use migration_helpers::{error, migrate, Migration, MigrationData, Result};
+use snafu::ResultExt;
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::process;
+
+const SETTING: &str = "settings.kubernetes.cpu-manager-policy";
+const CPU_MANAGER_POLICY_CHECKPOINT: &str = "var/lib/kubelet/cpu_manager_state";
+/// We remove cpu manager policy checkpoint value if checkpoint value is differenet from upgrade version
```
bcressey:
```suggestion
/// We remove cpu manager policy checkpoint value if checkpoint value is different from upgrade version
```

bcressey (commented):
For the CPU state migration, it might be better to just always remove the state file on boot, when we're sure no pods are running.

I'd recommend trying it without the migration, and adding this line to all the `kubernetes-tmpfiles.conf` snippets instead:
```
r! /var/lib/kubelet/cpu_manager_state
```

That still doesn't make the setting safe to switch on a running system, but it at least means a reboot will always fix it.

bcressey:
> I'd recommend trying it without the migration, and adding this line to all the `kubernetes-tmpfiles.conf` snippets instead:

Oops - we'll still need the migration to fix up the downgrade path when going back to a version that didn't have the `tmpfiles` line, so we should do both.

gthao313:
Push above 

- [x] rename variables like  `KubernetesDurationValue `
- [x] improve `cpu-manager-policy` unusual checkpoint to  **For upgrade, add r! /var/lib/kuberlet/cpu_manager_state  and for downgrade we should still use migration backward**
- [x] typo improvement

sources/models/src/lib.rs
```diff
@@ -113,12 +113,11 @@ use std::collections::HashMap;
 use std::net::Ipv4Addr;
 
 use crate::modeled_types::{
-    BootstrapContainerMode, DNSDomain, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue,
-    FriendlyVersion, Identifier, KubernetesAuthenticationMode, KubernetesBootstrapToken,
-    KubernetesCloudProvider, KubernetesClusterName, KubernetesEvictionHardKey, KubernetesLabelKey,
-    KubernetesLabelValue, KubernetesQuantityValue, KubernetesReservedResourceKey,
-    KubernetesTaintValue, KubernetesThresholdValue, Lockdown, SingleLineString, SysctlKey, Url,
-    ValidBase64,
+    CpuManagerPolicy, BootstrapContainerMode, DNSDomain, ECSAgentLogLevel,
```
bcressey: nit on ordering:
```suggestion
    BootstrapContainerMode, CpuManagerPolicy, DNSDomain, ECSAgentLogLevel,
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -768,3 +768,102 @@ mod test_kubernetes_cloud_provider {
...
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesDurationValue represents a string that contains a valid kubernetes duration value.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesDurationValue {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref CPU_MANAGER_RECONCILE_PERIOD: Regex = Regex::new(
```
```suggestion
    pub(crate) static ref KUBERNETES_DURATION_VALUE: Regex = Regex::new(
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -768,3 +768,102 @@ mod test_kubernetes_cloud_provider {
...
+    fn bad_cpu_manager_policy() {
+        for err in &["", "bad", "100", &"a".repeat(64)] {
+            CpuManagerPolicy::try_from(*err).unwrap_err();
+        }
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesDurationValue represents a string that contains a valid kubernetes duration value.
```
```suggestion
/// KubernetesDurationValue represents a string that contains a valid Kubernetes duration value.
```

sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+use migration_helpers::{error, migrate, Migration, MigrationData, Result};
+use snafu::ResultExt;
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::process;
+
+const SETTING: &str = "settings.kubernetes.cpu-manager-policy";
+const CPU_MANAGER_POLICY_CHECKPOINT: &str = "/var/lib/kubelet/cpu_manager_state";
+/// We remove cpu manager policy checkpoint value if checkpoint value is different from upgrade version
```
etungsten: I think we're removing this checkpoint file always. Not only when it's different.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -768,3 +768,102 @@ mod test_kubernetes_cloud_provider {
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesDurationValue {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref CPU_MANAGER_RECONCILE_PERIOD: Regex = Regex::new(
+        r"^(([0-9.]+)(h))?(([0-9.]+)(m))?(([0-9.]+)(s))?$"
```
bcressey: Do we need all the inner capture groups? We should also support `ms`. Not sure if we should allow cases like "....3h" either. Does kubelet have a regex we could use instead?

```suggestion
        r"^([0-9.]+h)?([0-9.]+m)?([0-9.]+s)?([0-9.]+ms)?$"
```

sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+/// or downgrade version cpu-manager-policy setting value.
+pub struct CpuManagerPolicyCleaner;
+
+impl Migration for CpuManagerPolicyCleaner {
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        println!("CpuManagerPolicyCleaner forward has no work to do on upgrade.");
```
etungsten: I think it's best to explain why it doesn't have work to do on upgrade. We're not changing the value of a setting or anything. It's just that we're adding a behavior where we always remove the `cpu_manager_state` checkpoint file, therefore we don't need to explicitly remove the file during forward migration.

etungsten (approved):

sources/models/src/modeled_types/mod.rs
```diff
@@ -74,6 +74,15 @@ pub mod error {
...
             source: std::num::ParseFloatError,
         },
+
+        #[snafu(display("Invalid Cpu Manager policy '{}'", input))]
+        InvalidCpuManagerPolicy {
+            input: String,
+            source: serde_plain::Error,
+        },
+
+        #[snafu(display("Invalid Cpu Manager reconcile period '{}'", input))]
```
bcressey: We might want to reuse `InvalidPlainValue` for this case - we don't want to hardcode "Cpu Manager" here.

webern (approved):

sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+use snafu::ResultExt;
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::process;
+
+const SETTING: &str = "settings.kubernetes.cpu-manager-policy";
+const CPU_MANAGER_POLICY_CHECKPOINT: &str = "/var/lib/kubelet/cpu_manager_state";
+/// We remove cpu manager policy checkpoint value if checkpoint value is different from upgrade version
+/// or downgrade version cpu-manager-policy setting value.
```
bcressey:
```suggestion
/// We remove cpu manager policy checkpoint value on downgrade, since older versions did not
/// clean up this state file on boot.
```

sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
```diff
@@ -0,0 +1,59 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::{error, migrate, Migration, MigrationData, Result};
+use snafu::ResultExt;
+use std::fs;
+use std::io;
+use std::path::Path;
+use std::process;
+
+const SETTING: &str = "settings.kubernetes.cpu-manager-policy";
```
bcressey: not used

sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
bcressey:
```suggestion
        println!("CpuManagerPolicyCleaner has no work to do on upgrade.");
```

sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+                        println! ("NotFound: '{}'", CPU_MANAGER_POLICY_CHECKPOINT)
+                    }
+                }
+            }
+        }
+
+        Ok(input)
+    }
+}
+/// We added a new settings for configuring kubelet, `settings.kubernetes.cpu-manager-policy`
```
bcressey:
```suggestion
/// We changed the default for CPU manager policy and need to handle kubelet's state file. 
```

README.md
```diff
@@ -358,6 +358,8 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
+* `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`.
```
tjkirch: We should probably document that Kubernetes strongly recommends a full reboot if you change this setting after launch.  Our kubernetes service and kubelet-config configuration-file will notice the update and be rewritten, but kubelet isn't designed to handle that change while the system is running, and things could start to go wrong when it notices the state file's policy doesn't match the config file's policy.

README.md
```diff
@@ -358,6 +358,8 @@ The following settings are optional and allow you to further configure your clus
...
+* `settings.kubernetes.cpu-manager-reconcile-period`: Specifies the CPU manager reconcile period.
```
tjkirch: I think it'd help to give a brief explanation of why the user would care about changing these settings; they're pretty opaque, even in the k8s documentation.

sources/api/migration/migration-helpers/src/error.rs
```diff
@@ -99,6 +99,12 @@ pub enum Error {
         metadata: String,
         data: Vec<serde_json::Value>,
     },
+
+    #[snafu(display("Failed to delete file '{}': '{}'", path, source))]
+    RemoveFile {
+        path: String,
```
tjkirch:
```suggestion
    #[snafu(display("Failed to delete file '{}': {}", path.display(), source))]
    RemoveFile {
        path: PathBuf,
```

sources/models/src/modeled_types/kubernetes.rs
gthao313: I don't think kubelet have a regex that we can use since I did not find anything on kubelet source code. I dot not think we should allow cases like `10s10h` in bottlerocket. Although `10s10h` is accepted in kubelet and it will be converted to `10h10s`, `10s10h` is not a meaningful format for me and we should keep a meaningful input standard .

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -817,3 +817,53 @@ mod test_cpu_manager_policy {
...
+
+/// KubernetesDurationValue represents a string that contains a valid Kubernetes duration value.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesDurationValue {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref KUBERNETES_DURATION_VALUE: Regex = Regex::new(
+        r"^([0-9.]+h)?([0-9.]+m)?([0-9.]+s)?([0-9.]+ms)?$"
```
gthao313: New changes:

1. simplify regex like wiping off `( )` 
2. support `ms`
2.  set standard that we do not accept formats like  `10s10h`, `10s1m` ....

gthao313 (commented):
Push above improve `KubernetesDurationValue` regex and improve comments/variable name

packages/kubernetes-1.16/kubelet-config
```diff
@@ -74,6 +74,10 @@ systemReserved:
   {{@key}}: "{{this}}"
   {{~/each}}
 {{~/if}}
+cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
```
tjkirch, line 77: Would you please confirm how you tested this on 1.16 - 1.19?  We weren't listing the setting before, so there's potential for kubelet to behave differently now that we're giving it an explicit value, even if we think it's the same.

sources/models/src/modeled_types/kubernetes.rs
tjkirch: I believe this would still allow empty input; we should probably have a `ensure!(!foo.is_empty(), error::bla)` check in the `try_from` implementation.

tjkirch: As Ben mentioned, it would also allow values like `...3ms` which we wouldn't want.  I think we can handle that by fixing the decimal checks, maybe something like `r"^(([0-9]+\.)?[0-9]+h)?(([0-9]+\.)?[0-9]+m)?(([0-9]+\.)?[0-9]+s)?(([0-9]+\.)?[0-9]+ms)?$"`

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -768,3 +768,102 @@ mod test_kubernetes_cloud_provider {
...
+    #[test]
+    fn good_tokens() {
+        for ok in &["9ms", "99s", "20m", "1h", "1h2m3s10ms", "4m5s10ms", "2h3s10ms"] {
+            KubernetesDurationValue::try_from(*ok).unwrap();
+        }
+    }
+
+    #[test]
+    fn bad_names() {
+        for err in &["100", "ten second", "1m2h", "9ns", &"a".repeat(23)] {
```
tjkirch: We can also check empty input, and cases like `...3ms` and `1..5s`.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -768,3 +768,102 @@ mod test_kubernetes_cloud_provider {
...
+string_impls_for!(KubernetesDurationValue, "KubernetesDurationValue");
+
+#[cfg(test)]
+mod test_kubernetes_duration_value {
+    use super::KubernetesDurationValue;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn good_tokens() {
+        for ok in &["9ms", "99s", "20m", "1h", "1h2m3s10ms", "4m5s10ms", "2h3s10ms"] {
```
tjkirch: We can also check decimals like `1.5h3.5m` since they'd be allowed.

sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl Migration for CpuManagerPolicyCleaner {
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        println!("CpuManagerPolicyCleaner has no work to do on upgrade.");
+        Ok(input)
+    }
+
+    fn backward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        // removing existing cpu_manager_policy_state file
+        if Path::new(CPU_MANAGER_POLICY_CHECKPOINT).exists() {
```
tjkirch: I don't think this check is helping because we already check for NotFound and can indicate that it didn't exist.  I think it'd be better to remove this because then we'd get output from the migration rather than silence.

sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+            println!(
+                "Deleting existing cpu manager policy checkpoint: '{}'",
+                CPU_MANAGER_POLICY_CHECKPOINT
+            );
+            match fs::remove_file(CPU_MANAGER_POLICY_CHECKPOINT) {
+                Ok(()) => {}
+                Err(e) => {
```
tjkirch:
```suggestion
            if let Err(e) = fs::remove_file(CPU_MANAGER_POLICY_CHECKPOINT) {
```

sources/api/migration/migrations/v1.1.3/kubelet-cpu-manager-state/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+                    if e.kind() != io::ErrorKind::NotFound {
+                        return Err(e).context(error::RemoveFile {
+                            path: CPU_MANAGER_POLICY_CHECKPOINT,
+                        });
+                    } else {
+                        println! ("NotFound: '{}'", CPU_MANAGER_POLICY_CHECKPOINT)
```
tjkirch: Would you please run `cargo fmt` and `cargo clippy` and confirm it's clean?

README.md
gthao313: That sounds good. Thanks

packages/kubernetes-1.16/kubelet-config
gthao313: Yeah, sure thing.

gthao313:
Push above:

1. Improve code according to Tom's comments
2. Improve Kubernetes duration regex and unit test.
3. Done configuring setting tests and migration test for kubernetes 1.16 to 1.19 variant. All tests have been passed.

README.md
```diff
@@ -358,6 +358,11 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
+* `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`. If you want to allow
+pods with certain resource characteristics to be granted increased CPU affinity and exclusivity on the node, you can set this setting to `static`. We recommend
+a full reboot if you change this setting after launch.
+* `settings.kubernetes.cpu-manager-reconcile-period`: Specifies the CPU manager reconcile period, which can configure CPU Manager Reconcile Kubelet CPU memory
```
```suggestion
* `settings.kubernetes.cpu-manager-reconcile-period`: Specifies the CPU manager reconcile period, which controls how often updated CPU assignments are written to cgroupfs. 
```

bcressey (approved):

README.md
```diff
@@ -358,6 +358,11 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
+* `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`. If you want to allow
```
webern: nit: We usually wrap at the sentence break. This seems like kind of a wide wrapping.

webern (approved):

gthao313:
push above improve README.

README.md
```diff
@@ -358,6 +358,8 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
+* `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`. If you want to allow pods with certain resource characteristics to be granted increased CPU affinity and exclusivity on the node, you can set this setting to `static`. We recommend a full reboot if you change this setting after launch.
```
```suggestion
* `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`. If you want to allow pods with certain resource characteristics to be granted increased CPU affinity and exclusivity on the node, you can set this setting to `static`. You should reboot if you change this setting after startup - try `apiclient reboot`.
```

README.md
```diff
@@ -358,6 +358,8 @@ The following settings are optional and allow you to further configure your clus
...
+* `settings.kubernetes.cpu-manager-reconcile-period`: Specifies the CPU manager reconcile period, which controls how often updated CPU assignments are written to cgroupfs.
```
```suggestion
* `settings.kubernetes.cpu-manager-reconcile-period`: Specifies the CPU manager reconcile period, which controls how often updated CPU assignments are written to cgroupfs.  The value is a duration like `30s` for 30 seconds or `1h5m` for 1 hour and 5 minutes.
```

tjkirch (approved):

packages/kubernetes-1.16/kubelet-config
gthao313: ### Test for 1.16 -1.19 variants
 check if `cpu-manager-policy` has been configured expectantly and test if fluent-bit pod work as expected:

Test with `cpu-manager-policy= "none"`
```
1726 policy_none.go:43] [cpumanager] none policy: Start
```
Test with default ("none")
```
1737 cpu_manager.go:193] [cpumanager] starting with none policy
```
Test with `cpu-manager-policy= "static"`
```
1723 cpu_manager.go:193] [cpumanager] starting with static policy
```

configure it as `none` and test if fluent-bit pod work as expected:
1. kubectl run --image public.ecr.aws/aws-observability/aws-for-fluent-bit:2.16.0 --command -- sh
2. kubectl exec -ti sh -- sh
3. /fluent-bit/bin/fluent-bit -H -i cpu -o null 
4. while curl localhost:2020; do echo pass; done

Migration test - check if `cpu-manager-policy` still be started after upgrade/downgrade even unusual checkpoint value (same as test for 1.20).

gthao313:
Push above improve README according to Tom's comments

packages/kubernetes-1.16/kubelet-config
tjkirch: We also confirmed offline that regardless of whether you have an explicit "none" setting or the default, you do see both messages, `none policy: Start` and `starting with none policy` - I wanted to make sure we weren't taking a different code path.

tjkirch (approved):

webern (approved):

PR #1646, opened 2021-07-02
[NETDOG] Fixes #1645 : make dns_search a Vec to avoid comma separated search domains in resolv.conf file.

### LINKED ISSUE #1645: resolv.conf search invalid
**Image I'm using:**
eks-1.20-1.1.2 in AWS

I expect the search domains in resolv.conf to be space separated.

**What actually happened:**
The search domains are comma separated which makes them invalid.

In the code that parses the lease file from wicked, each value is being edited to remove spaces.
https://github.com/bottlerocket-os/bottlerocket/blob/036ced8e6a67d7c52b63c35d5193e5fc2e70e777/sources/api/netdog/src/main.rs#L264
The problem is that if the value in LeaseInfo.dns_search is itself a string, it will not be parsed as array by envy but will still be edited to have commas.
For instance, the DNSSEARCH entry in the lease file has valuie
```
DNSSEARCH='domain1 domain2'
```
Before Envy, netdog will make it
```
DNSSEARCH='domain1,domain2'
```

This will result in a resolv.conf file having
```
search domain1,domain2
```
which is invalid
Add DHCP option set in the vpc used by a bottlerocket ec2 instance

hypnoce:
**Issue number:**
1645

**Description of changes:**
In netdog, make dns_search a Vec

**Testing done:**
Yes locally. The resolv.conf file is now valid.

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/main.rs
+++ b/sources/api/netdog/src/main.rs
@@ -146,7 +146,7 @@ struct LeaseInfo {
     #[serde(rename = "dnsdomain")]
     dns_domain: Option<String>,
     #[serde(rename = "dnssearch")]
-    dns_search: Option<String>,
+    dns_search: Option<Vec<String>>,
 }
 
 /// Informs the user about proper usage of the program and exits.
@@ -275,11 +275,11 @@ where
 }
 
 /// Write resolver configuration for libc.
-fn write_resolv_conf(dns_servers: &[&IpAddr], dns_search: &Option<String>) -> Result<()> {
+fn write_resolv_conf(dns_servers: &[&IpAddr], dns_search: &Option<Vec<String>>) -> Result<()> {
     let mut output = String::new();
 
     if let Some(s) = dns_search {
-        writeln!(output, "search {}", s).context(error::ResolvConfBuildFailed)?;
+        writeln!(output, "search {}", s.join(" ")).context(error::ResolvConfBuildFailed)?;
     }
 
     for n in dns_servers {


tjkirch (approved):
I tested and confirmed the results; this does fix the `search` line for multiple domains separated with spaces, so I'm happy.

There's a related (non-blocking) issue to consider that I haven't quite figured out yet.  The VPC console is purposely vague about the format of the domain name (search) field because different operating systems treat it differently; it's more specific about the domain name server line, in saying that they must be comma-separated.  For someone not familiar with resolv.conf, I think it'd be reasonable to enter the domain name values with comma separation as well, to fit the pattern.  However, that doesn't work as-is; we wind up with no search line in resolv.conf.  In fact, wicked doesn't even write the value to the file it hands netdog, so it wouldn't be as easy a fix.  I'm not sure if there's a nice way around that, but maybe someone will have an idea :)

zmrow (approved):
:tada: 

Thanks for the PR!

bcressey (approved):

tjkirch:
> we wind up with no search line in resolv.conf. In fact, wicked doesn't even write the value to the file it hands netdog

wicked doesn't show the value in its lease XML file, either.  Also confirmed that with ` , ` (space, comma, space) between values it gives the same result, just in case parsing was being picky about values within each domain string.  This was a minor concern and we have a working answer for the main use case in this PR, so I'll merge now. :)

hypnoce:
It was almost the 1000th PR 
image
I'm sad :)

PR #1654, opened 2021-07-14
api: add support for custom CA certificates via settings

### LINKED ISSUE #1277: Support For Custom Certificates
As a bottlerocket customer I would like the ability to install custom certificates within BottleRocket in order to access a private docker repository and pull down container images.  One potential solutions might include injecting the certificate as part of Userdata like what can be done with Amazon Linux optimized EKS images.

arnaldo2792:
**Issue number:**
#1277 

**Description of changes:**

```
684705a1 api: add support for custom CA certificates via settings
```

I'm still working on:

- [x] Migration
- [x] README documentation for new API

**Testing done:**

- [x] Pull down host container images
- [x] Send custom CA certificate through user data and verified the store was updated
- [x] Run pod/task/container in ecs-1, aws-k8s-1.19, aws-k8s-1.20
- [x] Run pod in vmware-1.20
- [x] Test container registry with custom CA

With bootstrap  container:

```Dockerfile
FROM alpine
RUN apk add openssl
ADD ./generate-certificate ./csr.conf ./config.yml ./
ENTRYPOINT ["sh", "generate-certificate"]
```

`generate-certificate` as:

```sh
#!/bin/sh

set -xe

PERSISTENT_DIR=/.bottlerocket/rootfs/mnt/certificates

mkdir -p $PERSISTENT_DIR

# Mount certificates as a tmpfs filesystem
mount -t tmpfs none $PERSISTENT_DIR
cp config.yml $PERSISTENT_DIR

# Generate 3 certificates
for i in $(seq 3); do
  # Generate CA Certificate and Key
  openssl genrsa -out $PERSISTENT_DIR/ca-$i.key 2048
  openssl req -x509 -new -nodes -key $PERSISTENT_DIR/ca-$i.key \
    -subj "/CN=bottlerocket/C=US/L=WASHINGTON" -days 1825 -out $PERSISTENT_DIR/ca-$i.crt
done

# Randomly select the CA certificate
TARGET=ca-$(( $RANDOM % 3 + 1 ))
ln -s $PERSISTENT_DIR/$TARGET.crt $PERSISTENT_DIR/ca.crt
ln -s $PERSISTENT_DIR/$TARGET.key $PERSISTENT_DIR/ca.key

# Generate SSL/TLS Certificates
openssl genrsa -out $PERSISTENT_DIR/server.key 2048
openssl req -new -key $PERSISTENT_DIR/server.key -out $PERSISTENT_DIR/server.csr -config csr.conf
openssl x509 -req -in $PERSISTENT_DIR/server.csr -CA $PERSISTENT_DIR/ca.crt -CAkey $PERSISTENT_DIR/ca.key \
  -CAcreateserial -out $PERSISTENT_DIR/server.crt -days 10000 -extensions req_ext \
  -extfile csr.conf

# Save all certificates as a bundle
BUNDLE=$(cat $PERSISTENT_DIR/ca-1.crt $PERSISTENT_DIR/ca-2.crt $PERSISTENT_DIR/ca-3.crt | base64 -w0)
apiclient set pki.local-registry.data=$BUNDLE \
  pki.local-registry.trusted=true
```

Deployed local registry with pod spec (and the equivalent of an ECS task):

```yml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: ca-certificates
spec:
  serviceName: ca-certificates
  replicas: 1
  selector:
    matchLabels:
      app: ca-certificates
  template:
    metadata:
      labels:
        app: ca-certificates
    spec:
      volumes:
        - name: certificates
          hostPath:
            path: /mnt/certificates
            type: Directory
        - name: config
          hostPath:
            path: /mnt/certificates/config.yml
      containers:
        - name: registry
          image: docker.io/library/registry:2
          ports:
            - containerPort: 5000
              hostPort: 5000
          volumeMounts:
            - mountPath: /certificates
              name: certificates
              readOnly: true
            - mountPath: /etc/docker/registry/config.yml
              name: config
              readOnly: true
```

Test that I get a 404 instead of an error related to a cerficate:

```sh
bash-5.0# ctr -a /run/host-containerd/containerd.sock image pull localhost:5000/test:latest
INFO[0000] trying next host - response was http.StatusNotFound  host="localhost:5000"
ctr: failed to resolve reference "localhost:5000/test:latest": localhost:5000/test:latest: not found
# Set the bundle as distrusted
bash-5.0# apiclient set pki.local-registry.trusted=false
# Get error `certificate signed by unknown authority`
bash-5.0# ctr -a /run/host-containerd/containerd.sock image pull localhost:5000/test:latest
INFO[0000] trying next host                              error="failed to do request: Head \"https://localhost:5000/v2/test/manifests/latest\": x509: certificate signed by unknown authority" host="localhost:5000"
ctr: failed to resolve reference "localhost:5000/test:latest": failed to do request: Head "https://localhost:5000/v2/test/manifests/latest": x509: certificate signed by unknown authority
```
**Terms of contribution:**


arnaldo2792:
Forced push includes migration

arnaldo2792:
Forced push includes:

- Documentation
- Minor updates  in  README
- Fix in the service triggered when `pki` settings change

README.md
```diff
@@ -465,6 +465,33 @@ Here are the metrics settings:
...
+* `settings.pki.<bundle-name>.trusted`: Whether the certificates in the bundle are trusted; defaults to `false` when not provided
+
+Here's an example of adding a bundle of self-signed certificates, using API calls:
+
+```sh
+apiclient set \
+  pki.my-trusted-bundle.data="W3N..." \
+  pki.my-trusted-bundle.trusted=true  \
+  pki.dont-trust-these.data="N3W..."  \
+  pki.dont-trust-there.trusted=false
```
bcressey: We don't provide very many `apiclient set` examples elsewhere. Are we leading with this approach because we expect the size of certificate bundles to exceed what's practical to pass through user-data? That might be worth a comment if so.

README.md
```diff
@@ -516,7 +543,7 @@ There are a few important caveats to understand about host containers:
 * If you set `superpowered` to true, they'll essentially have root access to the host.
 
 Because of these caveats, host containers are only intended for special use cases.
-We use it for the control container because it needs to be available early to give you access to the OS, and we use it for the admin container because it needs high levels of privilege and because you need it to debug when orchestration isn't working.
+We use them for the control container because it needs to be available early to give you access to the OS, and for the admin container because it needs high levels of privilege and because you need it to debug when orchestration isn't working.
```
bcressey, line 555: Seems like a good edit but might be better as a separate commit.

README.md
```diff
@@ -535,7 +562,7 @@ Bootstrap containers have access to the underlying root filesystem on `/.bottler
 This allows bootstrap containers to create files, directories, and mounts that are visible to the host.
 
 Bootstrap containers are set up to run after the systemd `configured.target` unit is active.
-The containers' systemd unit depends on this target (and not on any of the bootstrap containers' peers) which means that bootstrap containers will not execute in a deterministic order
+The containers' systemd unit depends on this target (and not on any of the bootstrap containers' peers) which means that bootstrap containers will not execute in a deterministic order.
```
bcressey, line 574: I'd roll this into the other "fixup docs" commit.

sources/Cargo.lock
```diff
@@ -1703,9 +1796,9 @@ checksum = "490cc448043f947bae3cbee9c203358d62dbee0db12107a74be5c30ccfd09771"
 
 [[package]]
 name = "memchr"
-version = "2.4.0"
+version = "2.3.4"
```
bcressey, line 1816: Lots of downgrades like this. Is this a rebase gone awry, or do the new dependencies force us into older versions?

sources/api/certdog/README.md
```diff
@@ -0,0 +1,11 @@
+# certdog
+
+Current version: 0.1.0
+
+  certdog is a tool to manage the trusted certificates store. It adds/removes
```
bcressey:
```suggestion
  certdog is a tool to manage the trusted certificates store. It adds and removes
```

sources/api/migration/migrations/v1.1.4/add-custom-certificates/src/main.rs
```diff
@@ -0,0 +1,20 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added the setting `pki`
```
bcressey: nit:
```suggestion
/// We added settings and services for `pki`
```

sources/models/src/aws-dev/mod.rs
```diff
@@ -5,7 +5,7 @@ use std::collections::HashMap;
 use crate::modeled_types::Identifier;
 use crate::{
     AwsSettings, BootstrapContainer, HostContainer, KernelSettings, MetricsSettings,
-    NetworkSettings, NtpSettings, UpdatesSettings,
+    NetworkSettings, NtpSettings, PEMCertificate, UpdatesSettings,
```
bcressey: If NTP is "Ntp" and AWS is "Aws" then PEM should probably be "Pem":
```suggestion
    NetworkSettings, NtpSettings, PemCertificate, UpdatesSettings,
```

sources/models/src/modeled_types/mod.rs
```diff
@@ -83,6 +84,16 @@ pub mod error {
 
         #[snafu(display("Invalid Kubernetes duration value '{}'", input))]
         InvalidKubernetesDurationValue { input: String },
+
+        #[snafu(display("Invalid x509 certificate: {}", source))]
+        InvalidX509Certificate {
+            source: x509_parser::nom::Err<x509_parser::error::X509Error>,
```
bcressey: Can this be simplified to `x509_parser::error::X509Error`?

sources/models/src/modeled_types/shared.rs
```diff
@@ -618,3 +619,65 @@ mod test_valid_container_mode {
...
+
+    #[test]
+    fn valid_pem_certificate() {
+        assert!(PEMCertificateString::try_from(TEST_PEM).is_ok());
+        assert!(PEMCertificateString::try_from("").is_ok());
+    }
+
+    #[test]
+    fn invalid_pem_certificate() {
+        assert!(PEMCertificateString::try_from(
```
bcressey: Can we also test that it doesn't parse input without the BEGIN and END markers?

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+use std::io::{BufRead, Seek};
+use std::path::Path;
+use std::process;
+use std::str::FromStr;
+use x509_parser;
+
+use model::modeled_types::Identifier;
+
+// FIXME Get from configuration in the future
+const DEFAULT_API_SOCKET: &str = "/run/api.sock";
```
bcressey, line 27: $1 to whoever who actually fixes this FIXME. 😢 

(obviously not a blocker here)

bcressey (commented):
Notes for everything except `certdog` which I need to review in more detail.

sources/api/certdog/src/main.rs
tjkirch: Or just removes the FIXMEs; I'm not sure the idea makes sense any more.

sources/Cargo.lock
webern: Maybe copy from develop and overwrite the Cargo.lock, then build.

sources/api/certdog/src/main.rs
webern: A constants crate would at least consolidate the FIXME to one place, no?

arnaldo2792: I was planning to start the new crate in this PR, provided that I don't have to actually fix all the FIXMEs.

sources/models/src/modeled_types/mod.rs
arnaldo2792: I remember I tried, but I can double check.

sources/models/src/modeled_types/shared.rs
arnaldo2792: Not passing the markers will result in an empty iterator returned by `x509_parser::pem::Pem::iter_from_buffer`, so I think I can modify the code to do something like:

```rust
 let mut empty = true;
 for (_, pem) in x509_parser::pem::Pem::iter_from_buffer(&decoded_bytes).enumerate() {
            // Parse buffer into a PEM object, then to a x509 certificate
            let pem = pem.context(error::InvalidPEM)?;
            pem.parse_x509().context(error::InvalidX509Certificate)?;
            // Or skip the if and always set empty = false
            if empty {
                   empty = false;
            }
        }

  assert!(!empty)
```

README.md
arnaldo2792: I didn't have any intention to lead users to use `apiclient`, but you made a great point and I'll document what could happen if users send bundles through user data.

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+
+/// Stores user-supplied global arguments
+#[derive(Debug)]
+struct Args {
+    log_level: LevelFilter,
+    socket_path: String,
+    trusted_store: String,
+    source_bundle: String,
+    comment: bool,
+}
```
bcressey: I'd prefer using `argh` again here, as we did with `prairiedog`.

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    Ok(final_args)
+}
+
+/// Query the API for the certificate bundles, returns a tuple with trusted
+/// and distrusted PEM certificates
+async fn get_certificate_bundles<P>(
+    socket_path: P,
+) -> Result<(Vec<x509_parser::pem::Pem>, Vec<x509_parser::pem::Pem>)>
```
bcressey: Maybe something like this instead, with a structured type?

```suggestion
) -> Result<CertBundle>

...

struct CertBundle {
    trusted_certs: Vec<x509_parser::pem::Pem>,
    distrusted_certs: Vec<x509_parser::pem::Pem>,
}

impl TryFrom<HashMap<Identifier, model::PEMCertificate>> for CertBundle {};
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+where
+    P: AsRef<Path>,
+{
+    debug!("Querying the API for settings");
+
+    let method = "GET";
+    let uri = API_SETTINGS_URI;
+    let (_code, response_body) = apiclient::raw_request(&socket_path, uri, method, None)
+        .await
+        .context(error::APIRequest { method, uri })?;
```
bcressey, line 73: Can we query for just the settings we need, instead of retrieving all of them?

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+    add_comment: bool,
+) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    let source_bundle = source_bundle.as_ref();
+    let trusted_store = trusted_store.as_ref();
+
+    // The default bundle includes the certificates shipped with the OS
+    let default_bundle = fs::File::open(source_bundle).context(error::ReadFile {
```
bcressey: Let's have a more specific read error, to match the more specific write error.
```suggestion
    let default_bundle = fs::File::open(source_bundle).context(error::ReadSourceBundle {
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+    trusted_bundle.append(&mut trusted);
+    // Remove any distrusted certificate
+    trusted_bundle.retain(|pem| !distrusted.contains(pem));
+
+    // Write a PEM formatted bundle from trusted certificates
+    fs::write(
+        &trusted_store,
+        pems_to_string(&trusted_bundle, add_comment)?,
+    )
+    .context(error::UpdateTrustStore)?;
```
bcressey:
```suggestion
    .context(error::UpdateTrustedStore)?;
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+                })?
+            }
+
+            "-b" | "--source-bundle" => {
+                final_args.source_bundle = iter.next().context(error::Usage {
+                    message: "Did not give argument to --source-bundle",
+                })?
+            }
+
+            "-c" | "--comment" => final_args.comment = true,
```
bcressey: Why not always add comments?

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+            return Ok(Some(organizational_unit_str.to_string()));
+        }
+    } else if let Some(organization) = subject.iter_organization().next() {
+        if let Ok(organization_str) = organization.as_str() {
+            return Ok(Some(organization_str.to_string()));
+        }
+    }
+
+    Ok(None)
+}
```
bcressey: Is this what's required to mimic the default comments in our CA certificates bundle? It seems like a lot of work to produce output that might not be satisfactory in all cases.

```suggestion
fn comment_for_pem(pem: &x509_parser::pem::Pem) -> Result<Option<String>> {
    let cert = pem.parse_x509().context(error::ParseX509Certificate)?;
    let subject = cert.tbs_certificate.subject;
    let comment = subject                                                                                                                              
        .iter_common_name()                                                                                                                            
        .chain(subject.iter_organizational_unit())                                                                                                     
        .chain(subject.iter_organization())                                                                                                            
        .next();                                                                                                                                       
    Ok(comment                                                                                                                                         
        .and_then(|c| c.as_str().ok())                                                                                                                 
        .and_then(|c| Some(c.to_string())))
}
```

sources/Cargo.lock
arnaldo2792: I did regenerate the Cargo.lock file, no bad rebase, as Ben said, the new dependencies forced  older versions :cry:

sources/api/certdog/src/main.rs
arnaldo2792: Comments aren't necessary for the final bundle, but I can make them part of it by default.

sources/api/certdog/src/main.rs
arnaldo2792: I have to read the code in `apiclient` to check if it allows to retrieve specific keys

sources/api/certdog/src/main.rs
arnaldo2792: Yes, it is to mimic that. I haven't nail the logic that the `trust` binary uses, but this is close to what the binary produces.

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+    // Build a Settings struct from the response string
+    debug!("Deserializing response");
+    let settings: model::Settings =
+        serde_json::from_str(&response_body).context(error::ResponseJson { method, uri })?;
+
+    split_bundles(settings.pki.unwrap_or_default())
+}
+
+/// Returns a tuple with two lists, for trusted and distrusted certificates
+fn split_bundles(
```
webern, line 84: This is a pure function (nice!) and should have unit tests.

(I meant to bundle this with a review but hit the wrong button)

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+            distrusted.append(&mut pems);
+        }
+    }
+
+    Ok((trusted, distrusted))
+}
+
+/// Updates the trusted certificates store, removing the distrusted certificates
+/// from the final bundle
+fn update_trusted_store<P>(
```
webern, line 120: Using `TempFile::TempDir` and a checked-in bundle, I believe this could also be unit tested.

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+
+    for pem in pems {
+        writeln!(out, "{}", pem_to_string(pem, add_comment)?).context(error::WritePEMString)?;
+    }
+
+    Ok(out)
+}
+
+/// Transforms a PEM object into a PEM formatted string
+fn pem_to_string(pem: &x509_parser::pem::Pem, add_comment: bool) -> Result<String> {
```
webern: This and `pems_to_string` also seem test-able, but would be covered by `update_trusted_store`.

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+fn usage() {
+    let program_name = env::args().next().unwrap_or_else(|| "program".to_string());
+    eprintln!(
+        r"Usage: {} [ ARGUMENTS... ]
+
+    Global arguments:
+        [ --socket-path PATH ]
+        [ --source-bundle PATH]
+        [ --trusted-store PATH]
+        [ --comment]
```
webern: nit

```suggestion
        [ --source-bundle PATH ]
        [ --trusted-store PATH ]
        [ --comment ]
```

sources/Cargo.toml
```diff
@@ -37,6 +38,7 @@ members = [
     "api/migration/migrations/v1.1.2/control-container-v0-5-1",
     "api/migration/migrations/v1.1.3/kubelet-cpu-manager-state",
     "api/migration/migrations/v1.1.3/kubelet-cpu-manager",
+    "api/migration/migrations/v1.1.4/add-custom-certificates",
```
webern: Looks like the version here doesn't match `Release.toml`.

```suggestion
    "api/migration/migrations/v1.2.0/add-custom-certificates",
```

sources/models/fixtures/test-pem
```diff
@@ -0,0 +1 @@
...YjNEUUVCQVFVQUE0SUJEd0F3Z2dFS0FvSUJBUUMxK0pOTG5ZSHgKcnhqMDkybEFKcDhkcEZUWG01aUduY2RUbm4wcHptTi8wcjFoV1FjS0lPSWNVMHFHZTdYdFJFTHFBT0FuRm92agpLWkRFeUoweHlPU2dZODRRejFXV21Lakh1bDBmV0RydEVpRTVNN1JJcFVwZVY2cHo4TnJUaDlBYkhGdVFhTDJKClpNQjh1UVpxVkVIRWRZdEx6RWh4Z01PTTFLY2l1anI3M1lROUxJS2FQN2lubDFNeGFneHNWOHI2MmRvVkd4NlMKbE9pd1l1UHZnZDZ2VmpkOVo5dVFOSjRxVHcwVVYwTGxidW12cDdjSEx4N1NBM1lOeXpWQVhtMW9hSmJoSnlqTQpSZmRBcGQwd3ZHNE9UV3dneXduNDdiT3NKQStwSTA2NGZTRHhKSGlsbXRlQklUZld4aUNZNnZBb3NNZFRybUVsClBBcWM1bnptVkZCN0FnTUJBQUdqVXpCUk1CMEdBMVVkRGdRV0JCUWZRSWFodXBURTFHWnhQYnJ4WEt0aFYrc08KL3pBZkJnTlZIU01FR0RBV2dCUWZRSWFodXBURTFHWnhQYnJ4WEt0aFYrc08vekFQQmdOVkhSTUJBZjhFQlRBRApBUUgvTUEwR0NTcUdTSWIzRFFFQkN3VUFBNElCQVFBa3NQazZYTExuQnVUWCtMbWFicitFc3JzTkNkQWIrZDl5CnN0aXNkQ2pwa21LQ3Nlc0o5NlFnaTA2LzFNODh1L3ArR2RwK3hITFBJSlVMckdqMDFqMXVybUVicXdpZnhVbWsKZkpGem1tNGVOelFaaldYMkJuRzlQOHh5R09TVmllR29URWVtdzJQTU9YaHVFRXNKOXJDUEl3c3ZYTEc4TkFvMgpzYzhsWjV5TlZ2ZGh3cHpnWS95bzNqTSsyVnl2eCthZ01zYURrN2JoYzkyRnN2Ynk2dmt2cHgvNkx2S0hhb0poCm5sRnFMMXFmRGRmYzlPNXdWanlEcit0NjVQZVcvSTdmQnl6WUxLcjA4SDNyQk80UEV4dGlSa1BwSXBrRFVjZ1AKN3kwekhnenlXOXR2TVZQV3V0Q0wyd1BNY29IaXVya0ZzaDhEbUM3c1MrTDJTSkxtcGwyaAotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
```
webern: Is `fixtures` a canonical name for a test data directory? In migrator I used `tests/data` because `tests` is the Rust built-in directory for 'integration' tests. I wonder how many variations of directory names we have for test input data.

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+    #[snafu(visibility = "pub(super)")]
+    pub(super) enum Error {
+        #[snafu(display("Error sending {} to {}: {}", method, uri, source))]
+        APIRequest {
+            method: String,
+            uri: String,
+            source: apiclient::Error,
+        },
+
+        #[snafu(display("Unable to decode base64 from certificate '{}': '{}'", name, source))]
```
webern: nit

```suggestion
        #[snafu(display("Unable to decode base64 from certificate '{}': {}", name, source))]
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+        ParsePEM {
+            source: x509_parser::error::PEMError,
+        },
+
+        #[snafu(display("Failed to parse cert: {}", source))]
+        ParseX509Certificate {
+            source: x509_parser::nom::Err<x509_parser::error::X509Error>,
+        },
+
+        #[snafu(display("Error while reading file {}: '{}'", path.display(), source))]
```
webern: nit

```suggestion
        #[snafu(display("Error while reading file '{}': {}", path.display(), source))]
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+        ReadFile { path: PathBuf, source: io::Error },
+
+        #[snafu(display(
+            "Error deserializing response as JSON from {} to {}: {}",
```
webern: nit:

(or maybe it could be better than my edit, but slightly awkward as is)

```suggestion
            "Error deserializing '{}' response '{}': {}",
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,401 @@
...
+            source: serde_json::Error,
+        },
+
+        #[snafu(display("{}", message))]
+        Usage { message: String },
+
+        #[snafu(display("Failed to update trust store: {}", source))]
+        UpdateTrustStore { source: io::Error },
+
+        #[snafu(display("failed to write to pem string: {}", source))]
```
webern: nit

```suggestion
        #[snafu(display("Failed to write to pem string: {}", source))]
```

sources/models/shared-defaults/defaults.toml
```diff
@@ -116,3 +116,12 @@ restart-commands = ["/usr/bin/bootstrap-containers create-containers"]
 
 [metadata.settings.bootstrap-containers]
 affected-services = ["bootstrap-containers"]
+
+# Certdog
```
webern, line 132: Are there services that need to be restarted when the bundle changes? Or does everything read the bundle 'just in time'?

sources/models/src/modeled_types/mod.rs
```diff
@@ -9,6 +9,7 @@
 pub mod error {
     use regex::Regex;
     use snafu::Snafu;
+    use x509_parser;
```
webern: nit, I think this is unused because the symbols are fully-qualified below?

```suggestion
```

arnaldo2792:
Forced push includes:

- Updated README
- Moved nits in README unrelated to CA certificates, to their own commit
- Use `argh`  instead of custom arguments parser
- Fixed nit in certdog's documentation
- Always use certificate comments
- Added `CertBundle` struct instead of tuple of `Vec`s
- Changed error `ReadFile` to `ReadSourceBundle`
- Refactor `comment_for_pem` as suggested

sources/models/shared-defaults/defaults.toml
arnaldo2792: I think everything access to the bundle at runtime. In my test, I didn't have to restart `containerd` when I updated the bundle.

sources/api/certdog/src/main.rs
arnaldo2792: As far as I can tell,  no, [`raw_request`](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/apiclient/src/lib.rs
) doesn't allow to filter specific keys.

arnaldo2792:
Forced push includes:
- Fix migration version  from (1.1.3, 1.2.0) to (1.1.4, 1.2.0)
- Rename migration to 1.2.0
- Unit tests in certdog for `split_bundles` and  `update_trusted_store`
- Various nits in error messages
- Move test data from `fixtures` to `tests/data`
- Simplified `comment_for_pem`
- Validation to check if the bundle's data doesn't contain any valid PEM certificate
- Additional unit tests for `PemCertificateString`

sources/api/certdog/src/main.rs
tjkirch: It's not a part of the function, it's the URI you send to.  There are parameters like [keys](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/openapi.yaml#L21) and [prefix](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/openapi.yaml#L32) that let you select what you want to get back.

bcressey (commented):
The docs fix commit is showing up as empty.

sources/api/certdog/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1"
+simplelog = "0.10"
+snafu = "0.6"
+tokio = { version = "1", default-features = false, features = ["macros", "rt-multi-thread"] }
+x509-parser = "0.9.2"
+tempfile = "3.2.0"
+
+[build-dependencies]
+cargo-readme = "3.1"
```
webern: Test-only dependencies can go in `dev-dependencies` and they are not compiled during non-test builds.

```suggestion
x509-parser = "0.9.2"

[build-dependencies]
cargo-readme = "3.1"

[dev-dependencies]
tempfile = "3.2.0"
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+        ParsePEM {
+            source: x509_parser::error::PEMError,
+        },
+
+        #[snafu(display("Failed to parse cert: {}", source))]
+        ParseX509Certificate {
+            source: x509_parser::nom::Err<x509_parser::error::X509Error>,
+        },
+
+        #[snafu(display("Error while reading bundle from file {}: '{}'", path.display(), source))]
```
webern: Nit: we always have the 'thing' in single quotes and the error at the end without quotes.

```suggestion
        #[snafu(display("Error while reading bundle from file '{}': {}", path.display(), source))]
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+        ReadSourceBundle { path: PathBuf, source: io::Error },
+
+        #[snafu(display("Error deserializing response from '{}': '{}'", uri, source))]
```
webern: Nit: we always have the 'thing' in single quotes and the error at the end without quotes.

```suggestion
        #[snafu(display("Error deserializing response from '{}': {}", uri, source))]
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+        ResponseJson {
+            uri: String,
+            source: serde_json::Error,
+        },
+
+        #[snafu(display("Failed to update trust store: {}", source))]
+        UpdateTrustedStore { source: io::Error },
+
+        #[snafu(display("Failed to write to pem string: {}", source))]
+        WritePEMString { source: std::fmt::Error },
```
webern: Nit: per Rust style and consistent with elsewhere, Rust acronyms are given as a CamelCased word.

```suggestion
        WritePemString { source: std::fmt::Error },
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+    #[snafu(visibility = "pub(super)")]
+    pub(super) enum Error {
+        #[snafu(display("Error sending {} to {}: {}", method, uri, source))]
+        APIRequest {
+            method: String,
+            uri: String,
+            source: apiclient::Error,
+        },
+
+        #[snafu(display("Unable to decode base64 from certificate '{}': '{}'", name, source))]
```
webern: Nit: we always have the 'thing' in single quotes and the error at the end without quotes.

```suggestion
        #[snafu(display("Unable to decode base64 from certificate '{}': {}", name, source))]
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+        WritePEMString { source: std::fmt::Error },
+    }
+}
+
+type Result<T> = std::result::Result<T, error::Error>;
+
+#[cfg(test)]
+mod test_certdog {
+    use model;
+    use model::modeled_types::{Identifier, PemCertificateString};
```
webern, line 301: Unless there's a name issue I don't know about, probably cleaner to import `PemCertificate` and drop the import of `model`.

```suggestion
    use model::modeled_types::{Identifier, PemCertificate, PemCertificateString};
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+    use std::fs::File;
+
+    static TEST_PEM: &str = include_str!("../../../models/tests/data/test-pem");
+
+    #[test]
+    fn split_bundles() {
+        let mut bundle = HashMap::new();
+        bundle.insert(
+            Identifier::try_from("trusted").unwrap(),
+            model::PemCertificate {
```
webern, line 314:
```suggestion
            PemCertificate {
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+                data: Some(PemCertificateString::try_from(TEST_PEM).unwrap()),
+                trusted: Some(true),
+            },
+        );
+        bundle.insert(
+            Identifier::try_from("distrusted").unwrap(),
+            model::PemCertificate {
```
webern, line 321:
```suggestion
            PemCertificate {
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+                data: Some(PemCertificateString::try_from(TEST_PEM).unwrap()),
+                trusted: Some(false),
+            },
+        );
+        bundle.insert(
+            Identifier::try_from("distrusted-without-flag").unwrap(),
+            model::PemCertificate {
```
webern, line 328:
```suggestion
            PemCertificate {
```

sources/api/certdog/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+        let (_, pem) =
+            x509_parser::pem::parse_x509_pem(&base64::decode(TEST_PEM.as_bytes()).unwrap())
+                .unwrap();
+        let mut trusted_certs: Vec<x509_parser::pem::Pem> = Vec::new();
+        trusted_certs.push(pem);
+        let certs_bundle = super::CertBundle {
+            trusted_certs,
+            distrusted_certs: Vec::new(),
+        };
+        assert!(super::update_trusted_store(certs_bundle, &trusted_store, &source_bundle).is_ok());
```
webern: It's common to add `use super::*` at the top of `mod test_certdog`

```suggestion
        assert!(update_trusted_store(certs_bundle, &trusted_store, &source_bundle).is_ok());
```

webern (approved):
Thank you for adding unit tests! 🥼 

Just non-blocking cleanup nits here. Nice job.

sources/api/certdog/src/main.rs
arnaldo2792: PemCertificate isn't a modeled type, it is a model, and in other crates, we don't include `model` for some reason, we  always do `model::TheModel`

sources/api/certdog/src/main.rs
arnaldo2792: Same as above, we don't include `model` in other crates

sources/api/certdog/src/main.rs
arnaldo2792: Same

sources/api/certdog/src/main.rs
arnaldo2792: Same

arnaldo2792:
Forced push includes:

- Fixed empty commit
- Changed migrations to 1.1.5 instead of 1.2.0
- Cargo.toml includes `dev-dependencies` for certdog
- Nits in error names and messages

sources/api/certdog/src/main.rs
webern: OK. 🤷

webern (approved):

arnaldo2792:
Forced push moves migrations back to 1.2.0

sources/api/certdog/src/main.rs
arnaldo2792: Per our offline discussion, we will work on an improved helper to serialize/deserialize specific keys rather than the entire settings store.

bcressey (approved):

Issue #1663, opened 2021-07-21
API server: "write once, read many" functionality for settings


zmrow:
**What I'd like:**
I'd like a setting type that can only be set **once**, but read freely.

This would come in handy for things like `settings.network.hostname`, where we wouldn't want users to be able to change it after it is set, which can cause issues with kubelet.

Related to #1608

webern:
We should consider also that a hostname may be writeable without consequence in some variants but in other variants we might want it to be 'write-once'.

stmcginnis:
@cbgbt another fyi for settings...

bcressey:
I've contemplated a lockdown-style switch to suppress settings changes until next boot, which might help if the desire is for a protective barrier against unintentional changes. But otherwise this is probably best addressed through documentation if there's significant operational risk.

PR #1664, opened 2021-07-21
Add a new setting for configuring hostname

### LINKED ISSUE #1608: Allow hostname to be set via user-data
**What I'd like:**
When using bottlerocket with CAPV the `hostname` resolves as the IP of the node.  It is unknown if this causes any real issues, but it has the unfortunate side-effect of naming the kubernetes nodes based on this IP instead of the CAPV set hostname.  This creates an inconsistency when working with CAPV based clusters between bottlerocket and other OSs as well as the node names not matching the CAPV objects that are created in the cluster.

It would be great if the hostname could be set as part of the user-data or some other bootstrapping process to avoid this inconsistency.  

zmrow:
**Issue number:**
Fixes #1608 

**Description of changes:**
Reading through this PR by commit will probably be easiest.

> **IMPORTANT NOTE**
> As mentioned in #1608, kubelet uses the hostname when registering itself and for certificate/authorization purposes.  Given that fact, changing the hostname after the host's initial boot (and by extension, kubelet's initial registration) may cause issues with the kubelet.  We don't currently have a method in the API server for "write once, read many" situations but considered the setting useful enough to move forward.  #1663 tracks thoughts on the "write once, read many" API server functionality.

Previously, hostname was written directly to disk by `netdog` called by `wicked` early in boot.  At the point in the boot sequence when `wicked` calls `netdog`, the API server isn't up yet.  Rather than rework the boot sequence (which would be tricky as this point in boot) and add the ability to netdog to talk to the API, we opted to add a subcommand to `netdog` as a settings generator for the new `settings.network.hostname` setting.  The setting is used in a template file, which is an env file for a new systemd unit that calls `netdog set-hostname`.  This command is what will write the hostname to disk at `/proc/sys/kernel/hostname`.

This means that at boot time, if a user has the `hostname` setting in their user data, we will use it, otherwise we will set the hostname using largely the same logic as was in the program previously.  (The only change is that if the DNS reverse lookup fails, we use the IP in the format "ip-x-x-x" as the hostname). 

`netdog` was reworked to use `argh` for argument parsing since our custom logic was starting to get unwieldy with the addition of more subcommands with their own arguments.

**Testing done:**
* `/etc/resolv.conf` is populated in all cases
* Migrations work correctly and the setting and its metadata are correctly populated on upgrade, and nonexistent on downgrade
* I tested that you can set this setting via `apiclient` at runtime.  `kubelet` logs some errors.  I can't speak to what will happen to the kubelet or if it will continue to function or for how long.

No `settings.network.hostname` in user data:
* Boot `aws-k8s-1.19`: it starts, gets a correct hostname, and connects to the cluster just fine and runs a pod
* Boot `vmware-k8s-1.20`: it starts, gets the string `ip-x-x-x-x` as its hostname, connects to the cluster, runs a pod

`settings.network.hostname="br-test-1"` in user data:
* Boot `aws-k8s-1.10`: it boots, gets the hostname from user data, but (correctly) doesn't ever become ready in the cluster.
* Boot `vmware-k8s-1.20`: it boots, gets hostname from user data, and shows up neatly in the cluster, runs a pod.
```
> kubectl get nodes
NAME                STATUS   ROLES                  AGE     VERSION
br-test-1           Ready    <none>                 6h19m   v1.20.8
mrowicki-qs-xx6l2   Ready    control-plane,master   6h41m   v1.20.8
```

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,6 +416,13 @@ These settings can be changed at any time.
 
 #### Network settings
 
+* `settings.network.hostname`: The desired hostname of the system.
+  **Important note for all Kubernetes variants:** Changing this setting at runtime (not via user data) can cause issues with kubelet registration, as hostname is closely tied to the identity of the system for both registration and certificates/authorization purposes.
+
+Most users don't need to change this setting as the following defaults work for the majority of use cases.
+If this setting isn't set we attempt to use DNS reverse lookup for the hostname.
+If the lookup is unsuccessful, the IP of the node is used in the format `ip-X-X-X-X`.
+
 ##### Proxy settings
 
 These settings will configure the proxying behavior of the following services:
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -14,6 +14,7 @@ Source99: release-tmpfiles.conf
 
 Source200: motd.template
 Source201: proxy-env
+Source202: hostname-env
 
 Source1000: eth0.xml
 Source1001: multi-user.target
@@ -21,6 +22,7 @@ Source1002: configured.target
 Source1003: preconfigured.target
 Source1004: activate-configured.service
 Source1005: activate-multi-user.service
+Source1011: set-hostname.service
 
 # Mounts for writable local storage.
 Source1006: var.mount
@@ -108,7 +110,7 @@ EOF
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 \
   %{S:1001} %{S:1002} %{S:1003} %{S:1004} %{S:1005} \
-  %{S:1006} %{S:1007} %{S:1008} %{S:1009} %{S:1010} \
+  %{S:1006} %{S:1007} %{S:1008} %{S:1009} %{S:1010} %{S:1011} \
   %{S:1015} %{S:1040} %{S:1041} %{S:1060} %{S:1061} %{S:1062} \
   %{buildroot}%{_cross_unitdir}
 
@@ -129,6 +131,7 @@ install -p -m 0644 ${LICENSEPATH}.mount %{buildroot}%{_cross_unitdir}
 install -d %{buildroot}%{_cross_templatedir}
 install -p -m 0644 %{S:200} %{buildroot}%{_cross_templatedir}/motd
 install -p -m 0644 %{S:201} %{buildroot}%{_cross_templatedir}/proxy-env
+install -p -m 0644 %{S:202} %{buildroot}%{_cross_templatedir}/hostname-env
 
 install -d %{buildroot}%{_cross_udevrulesdir}
 install -p -m 0644 %{S:1016} %{buildroot}%{_cross_udevrulesdir}/61-mount-cdrom.rules
@@ -163,9 +166,11 @@ ln -s %{_cross_unitdir}/preconfigured.target %{buildroot}%{_cross_unitdir}/defau
 %{_cross_unitdir}/*-kernels.mount
 %{_cross_unitdir}/*-licenses.mount
 %{_cross_unitdir}/var-lib-bottlerocket.mount
+%{_cross_unitdir}/set-hostname.service
 %dir %{_cross_templatedir}
 %{_cross_templatedir}/motd
 %{_cross_templatedir}/proxy-env
+%{_cross_templatedir}/hostname-env
 %{_cross_udevrulesdir}/61-mount-cdrom.rules
 
 %changelog
--- a/sources/api/netdog/README.md
+++ b/sources/api/netdog/README.md
@@ -5,7 +5,8 @@ Current version: 0.1.0
 ## Introduction
 
 netdog is a small helper program for wicked, to apply network settings received from DHCP.  It
-generates `/etc/resolv.conf`, generates and sets the hostname, and persists the current IP to file.
+generates `/etc/resolv.conf`, generates and sets the hostname, and persists the current IP to a
+file.
 
 It contains two subcommands meant for use as settings generators:
 * `node-ip`: returns the node's current IP address in JSON format
--- a/sources/api/netdog/README.md
+++ b/sources/api/netdog/README.md
@@ -12,6 +12,8 @@ It contains two subcommands meant for use as settings generators:
 * `generate-hostname`: returns the node's hostname in JSON format (it is the resolved IP or the IP
   in format "ip-x-x-x-x" if resolving fails)
 
+The subcommand `set-hostname` sets the hostname for the system.
+
 ## Colophon
 
 This text was generated using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
--- a/sources/api/netdog/README.md
+++ b/sources/api/netdog/README.md
@@ -4,11 +4,13 @@ Current version: 0.1.0
 
 ## Introduction
 
-netdog is a small helper program for wicked, to apply network settings received from DHCP.  It also
-contains a subcommand `node-ip` that returns the node's current IP address in JSON format; this
-subcommand is intended for use as a settings generator.
+netdog is a small helper program for wicked, to apply network settings received from DHCP.  It
+generates `/etc/resolv.conf`, generates and sets the hostname, and persists the current IP to file.
 
-It generates `/etc/resolv.conf`, sets the hostname, and persists the current IP to file.
+It contains two subcommands meant for use as settings generators:
+* `node-ip`: returns the node's current IP address in JSON format
+* `generate-hostname`: returns the node's hostname in JSON format (it is the resolved IP or the IP
+  in format "ip-x-x-x-x" if resolving fails)
 
 ## Colophon
 
--- a/sources/api/netdog/src/main.rs
+++ b/sources/api/netdog/src/main.rs
@@ -17,6 +17,10 @@ It generates `/etc/resolv.conf`, sets the hostname, and persists the current IP 
 
 #![deny(rust_2018_idioms)]
 
+#[macro_use]
+extern crate serde_plain;
+
+use argh::FromArgs;
 use dns_lookup::lookup_addr;
 use envy;
 use ipnet::IpNet;
@@ -27,12 +31,12 @@ use regex::Regex;
 use serde::Deserialize;
 use snafu::ResultExt;
 use std::collections::BTreeSet;
-use std::fmt::{self, Write};
+use std::fmt::Write;
 use std::fs::{self, File};
 use std::io::{BufRead, BufReader};
 use std::net::IpAddr;
 use std::path::{Path, PathBuf};
-use std::{env, process};
+use std::process;
 
 static RESOLV_CONF: &str = "/etc/resolv.conf";
 static KERNEL_HOSTNAME: &str = "/proc/sys/kernel/hostname";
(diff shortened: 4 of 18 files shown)


sources/api/netdog/src/main.rs
```diff
@@ -44,204 +51,133 @@ lazy_static! {
...
+
+    #[argh(positional)]
+    /// lease info data file
+    data_file: PathBuf,
+
+    #[argh(positional)]
+    // wicked might add `info` to the call to this program.  We don't do anything with it but must
+    // be able to parse the option to avoid failing
+    /// ignored
+    info: Option<String>,
```
bcressey: Based on the implementation in `ni_system_updater_generic_install_call`, and specifically `ni_system_updater_generic_leaseinfo_create`, we'll always get called with this "info" parameter, which is meant to identify the format of the lease data file.

sources/api/netdog/src/main.rs
```diff
@@ -259,7 +268,37 @@ fn node_ip() -> Result<()> {
...
+
+/// Attempt to resolve assigned IP address, if unsuccessful use "ip-X-X-X-X" where X's are the
+/// octets of the IP.  For example, IP address 1.2.3.4 becomes the hostname "ip-1-2-3-4".  No dots
+/// in the hostname (hopefully) avoids any confusion for programs that may read this value.
+///
+/// The result is returned as JSON. (intended for use as a settings generator)
+fn generate_hostname() -> Result<()> {
+    let ip_string =
+        fs::read_to_string(CURRENT_IP).context(error::CurrentIpReadFailed { path: CURRENT_IP })?;
+    let ip = IpAddr::from_str(&ip_string).context(error::IpFromString { ip: &ip_string })?;
```
bcressey, line 274: It'd be good to add this deserialization check to `node_ip()` also.

sources/api/netdog/src/main.rs
```diff
@@ -259,7 +268,37 @@ fn node_ip() -> Result<()> {
...
+    let hostname = match lookup_addr(&ip) {
+        Ok(hostname) => hostname,
+        Err(e) => {
+            eprintln!("Reverse DNS lookup failed: {}", e);
+            format!("ip-{}", ip_string.replace(".", "-"))
+        }
+    };
```
bcressey: Have you tested that this error path works?

It's surprising to me that it does since the `lookup_addr(&ip)` is effectively what we were doing in `update_hostname()`, and that seems to be what set the system hostname to the leased IPv4 address on VMware variants.

If it does work, then why?

If it doesn't work, I'd expect that the fix is to see whether the returned address matches the address we passed in. E.g. if `1.2.3.4` goes in and `1.2.3.4` comes out, we didn't actually resolve anything.

sources/api/netdog/src/main.rs
```diff
@@ -44,206 +48,119 @@ lazy_static! {
...
-            interface_type: interface_type.unwrap_or_else(|| usage()),
-            interface_family: interface_family.unwrap_or_else(|| usage()),
-            data_file: data_file.unwrap_or_else(|| usage()),
-        }),
-    ))
+    #[argh(positional)]
+    // wicked might add `info` to the call to this program.  We don't do anything with it but must
+    // be able to parse the option to avoid failing
+    /// ignored
+    info: Option<String>,
```
bcressey: It does not appear that these arguments are supplied during the `remove` call, based on `ni_system_updater_generic_remove_call`.

I'd expect "remove" to be broken by this change - because `data_file` became a required positional argument - so along with fixing it, it'd be good to confirm that you can recreate that failure.

sources/api/netdog/src/main.rs
```diff
@@ -170,6 +171,15 @@ struct NodeIpArgs {}
 /// Generate hostname from DNS reverse lookup or use current IP
 struct GenerateHostnameArgs {}
 
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "set-hostname")]
+/// Write hostname to disk
```
bcressey: nit: this doesn't really write the hostname to disk, it sets the hostname for the system (by writing to a special file). Commit message should be clarified also.

sources/api/netdog/src/main.rs
```diff
@@ -170,6 +171,15 @@ struct NodeIpArgs {}
...
+struct SetHostnameArgs {
+    #[argh(option)]
+    /// hostname for the system
+    hostname: String,
```
bcressey: I'd prefer this as a position argument.

packages/os/os.spec
```diff
@@ -532,4 +533,6 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 %{_cross_unitdir}/bootstrap-containers@.service
 %{_cross_tmpfilesdir}/bootstrap-containers.conf
 
+%{_cross_unitdir}/set-hostname.service
```
bcressey: This should either go in the `netdog` subpackage, or (preferably) in the `release` package with the template it depends on.

packages/release/release.spec
```diff
@@ -14,6 +14,7 @@ Source99: release-tmpfiles.conf
 
 Source200: motd.template
 Source201: proxy-env
+Source202: hostname.template
```
bcressey: I'd prefer the source template and installed template be called `hostname-env` for consistency with `proxy-env`.

sources/models/src/lib.rs
```diff
@@ -204,6 +204,7 @@ struct HostContainer {
 // Network settings. These settings will affect host service components' network behavior
 #[model]
 struct NetworkSettings {
+    hostname: SingleLineString,
```
bcressey: Need to add validation here.

README.md
```diff
@@ -417,6 +417,14 @@ These settings can be changed at any time.
 
 #### Network settings
 
+* 'settings.network.hostname': The desired hostname of the system.
+  **Important note for all Kubernetes variants:** This setting should only be set via user data!
+  Changing this setting at runtime can cause issues with kubelet as hostname is closely tied to the identity of the system for both registration and certificates/authorization purposes.
+  For AWS variants, changing this setting will cause issues with kubelet registration.
```
bcressey: Some notes:
* The warning text could be interpreted as saying that the setting should be set via user data, when really it's a specialized setting and you shouldn't configure it unless you understand the risks.
* I expect that changing this setting could also cause issues with the VMware variants.
* I don't think it strictly has to be set through user data; kubelet in standalone mode is probably fine, for example. It's only when it registers and the cloud provider code kicks in that the results are unpredictable (but probably bad).

README.md
```diff
@@ -417,6 +417,14 @@ These settings can be changed at any time.
...
+
+If this setting isn't set we attempt to use DNS reverse lookup for the hostname.
+If the lookup is unsuccessful, the IP of the node is used in the format `ip-X-X-X-X`.
+
```
bcressey, line 425: Maybe a note about how these defaults will be fine for most use cases.

sources/api/netdog/README.md
```diff
@@ -4,11 +4,13 @@ Current version: 0.1.0
...
-contains a subcommand `node-ip` that returns the node's current IP address in JSON format; this
-subcommand is intended for use as a settings generator.
+netdog is a small helper program for wicked, to apply network settings received from DHCP.  It
+generates `/etc/resolv.conf`, generates and sets the hostname, and persists the current IP to file.
 
-It generates `/etc/resolv.conf`, sets the hostname, and persists the current IP to file.
+It contains two subcommands meant for use as settings generators:
+* `node-ip`: returns the node's current IP address in JSON format
+* `generate-hostname`: returns the node's hostname in JSON format (it is the resolved IP or the IP
+  in format "ip-x-x-x-x" if resolving fails)
```
bcressey, line 14: We should also document `set-hostname`.

sources/api/netdog/src/main.rs
```diff
@@ -44,204 +51,133 @@ lazy_static! {
...
-
-    // The `node-ip` subcommand doesn't require any arguments
-    if sub_command == SubCommand::NodeIp {
-        return Ok((sub_command, None));
-    };
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "remove")]
+// `wicked` may call `remove` with the below args and failing to parse them can cause an error in
+// `wicked`.
+/// Not implemented
```
bcressey: It is implemented at least as far as having a function that prints that it's not implemented - so I'd remove this comment.

zmrow:
^ Addresses @bcressey 's comments!

zmrow:
^ Rebases on `develop`

sources/models/src/lib.rs
zmrow: Added a new type `ValidLinuxHostname` for validation for this field.

sources/api/netdog/src/main.rs
zmrow: Since we previously required these args this was already broken. :fearful: 

With the understanding of the `wicked` code you mentioned, I removed those arguments.  Testing it is going to be really complicated, but I feel pretty comfortable with the change given our understanding of the `wicked` source.

sources/api/netdog/src/main.rs
zmrow: You're right - with the previous code the node got the IP address as it's hostname.

I've updated the code return the `ip-x-x-x-x` string in the event the `lookup_addr()` call returns the same IP as we passed in.

sources/api/netdog/src/main.rs
zmrow: I'll open a separate PR for this.

sources/api/netdog/src/main.rs
```diff
@@ -1,11 +1,15 @@
 /*!
 # Introduction
 
-netdog is a small helper program for wicked, to apply network settings received from DHCP.  It also
-contains a subcommand `node-ip` that returns the node's current IP address in JSON format; this
-subcommand is intended for use as a settings generator.
+netdog is a small helper program for wicked, to apply network settings received from DHCP.  It
+generates `/etc/resolv.conf`, generates and sets the hostname, and persists the current IP to file.
```
bcressey: nit:
```suggestion
generates `/etc/resolv.conf`, generates and sets the hostname, and persists the current IP to a file.
```

sources/models/src/modeled_types/shared.rs
```diff
@@ -130,6 +130,102 @@ mod test_single_line_string {
...
+        );
+
+        // Each segment must be from 1-63 chars long
+        ensure!(
+            input.split(".").all(|x| x.len() >= 1 && x.len() <= 63),
+            error::InvalidLinuxHostname {
+                input,
+                msg: "segment is less than 1 or greater than 63 chars"
+            }
+        );
```
bcressey: It also seems like segments shouldn't start with a hyphen, even if that's not specified in the man page.

bcressey (approved):

etungsten (approved):
LGTM. ~~If we know changing the hostname during runtime is harmful for k8s variants (might be variant specific). We should probably think of a way to disallow that.~~

edit: sorry just remembered it's already covered in an follow up issue

sources/models/src/modeled_types/shared.rs
zmrow: Good point - I'll fix that.

etungsten (approved):

bcressey (approved):

etungsten:
Sorry for pointing this out after the fact. The next release version should be v1.2.0 so the migrations path and Release.toml needs to be updated from v1.1.5 to v1.2.0.

diranged:
I'm a bit confused - I thought that EKS forced us to always use the `ip-xx-xx-xx-xx` pattern, is that not true? Hostnames need to be unique too I believe, so is this something that is passed through a templating engine?

zmrow:
Hi @diranged  - as mentioned in the setting documentation in [the README](https://github.com/bottlerocket-os/bottlerocket/blob/develop/README.md#network-settings), most users shouldn't need to touch this setting as the defaults make sure that the hostname is set the same as it always has been.

In VMware, this setting can be useful to set the hostname the same as the VM name so I expect it will get the most use there.

If you're primarily running in EC2 - you shouldn't ever need to touch this setting. :)

PR #1680, opened 2021-07-30
netdog: Add IP validation to `node_ip()`

zmrow:
**Issue number:**
Requested in #1664 

**Description of changes:**
```
Previously the `node-ip` subcommand, which runs the `node_ip()`
function, would read the current IP from file and return it directly as
JSON.  This change validates what we read from the file is an actual
proper IP address.
```

**Testing done:**
* Run an `aws-k8s-1.19` node, it joins the cluster and runs a pod just fine.
* Manually change `/var/lib/netdog/current_ip` to an invalid IP `1.2.3`, run `netdog node-ip` by hand and (correctly) receive an error.
```
bash-5.0# netdog node-ip
Invalid IP address '1.2.3': invalid IP address syntax
```

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/main.rs
+++ b/sources/api/netdog/src/main.rs
@@ -257,10 +257,13 @@ fn remove(args: RemoveArgs) -> Result<()> {
 
 /// Return the current IP address as JSON (intended for use as a settings generator)
 fn node_ip() -> Result<()> {
-    let ip =
+    let ip_string =
         fs::read_to_string(CURRENT_IP).context(error::CurrentIpReadFailed { path: CURRENT_IP })?;
+    // Validate that we read a proper IP address
+    let _ = IpAddr::from_str(&ip_string).context(error::IpFromString { ip: &ip_string })?;
+
     // sundog expects JSON-serialized output
-    Ok(print_json(ip)?)
+    Ok(print_json(ip_string)?)
 }
 
 /// Attempt to resolve assigned IP address, if unsuccessful use "ip-X-X-X-X" where X's are the


webern (approved):

sources/api/netdog/src/main.rs
```diff
@@ -257,10 +257,13 @@ fn remove(args: RemoveArgs) -> Result<()> {
 
 /// Return the current IP address as JSON (intended for use as a settings generator)
 fn node_ip() -> Result<()> {
-    let ip =
+    let ip_string =
         fs::read_to_string(CURRENT_IP).context(error::CurrentIpReadFailed { path: CURRENT_IP })?;
+    // Validate that we read a proper IP address
+    let _valid_ip = IpAddr::from_str(&ip_string).context(error::IpFromString { ip: &ip_string })?;
```
arnaldo2792: Do you need to store it? I've done things like `my_method.context(error::SomeError)?` without having to store the value of `my_method` and without `cargo fmt` complaining about it

zmrow: I don't have to store it, but I thought it made it a bit clearer to the reader that we explicitly aren't using the IP type that gets returned.

bcressey: `let _ = ...` seems a little more idiomatic for that case

webern: > `let _ = ...` seems a little more idiomatic for that case

I've seen it both ways, but I agree `let _ = ` indicates "this returns something that we don't care about" clearly.

zmrow:
^ changed `_valid_ip` to `_`

sources/api/netdog/src/main.rs
zmrow: I changed it to `let _ =`

bcressey (approved):

PR #1684, opened 2021-08-02
Use runtime.slice and system.slice cgroup settings in k8s variants

### LINKED ISSUE #1681: Use runtime.slice and system.slice cgroup settings in k8s variants
Per the discussion, investigate and possibly implement the recommended cgroup settings "runtime.slice" and "system.slice".

### Discussed in https://github.com/bottlerocket-os/bottlerocket/discussions/1679

<div type='discussions-op-text'>

<sup>Originally posted by **cyrus-mc** July 29, 2021</sup>
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-allocatable.md#recommended-cgroups-setup

As highlighted above there should be a system.slice/cgroup and a podruntime.slice/cgroup that correspond to systemReserved and kubeReserved settings. Looking at the setup of Bottlerocket is appears there is just system.slice under which kubelet and runtime plus all system components run.

Was there a reason for this design decision? As it doesn't map nicely to the configuration settings k8s provides. </div>

<sup>Originally posted by **webern** July 29, 2021</sup>

Thanks for bringing this to our attention, and for the pointer to the proposal. Your link also led us to find this https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#kube-reserved which suggests "runtime.slice" and "system.slice" and points to the design proposal. We'll review our current slice setup and make sure it's aligned with kubelet functionality, unless it's something you'd like to contribute.

cyrus-mc:
**Issue number:**

#1681

**Description of changes:**

```
Follow documented best practices (https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-allocatable.md#recommended-cgroups-setup) by running runtime components (containerd, kubelet) within their own cgroup. Separate from system components.

This allows setting of kubeReserved and systemReserved to be enforced.
```

**Terms of contribution:**

### DIFF
--- a/packages/containerd/containerd.service
+++ b/packages/containerd/containerd.service
@@ -1,10 +1,11 @@
 [Unit]
 Description=containerd container runtime
 Documentation=https://containerd.io
-After=network-online.target configured.target
-Wants=network-online.target configured.target
+After=network-online.target configured.target runtime.slice
+Wants=network-online.target configured.target runtime.slice
 
 [Service]
+Slice=runtime.slice
 EnvironmentFile=/etc/network/proxy.env
 ExecStart=/usr/bin/containerd
 Type=notify
--- a/packages/containerd/containerd.service
+++ b/packages/containerd/containerd.service
@@ -1,8 +1,8 @@
 [Unit]
 Description=containerd container runtime
 Documentation=https://containerd.io
-After=network-online.target configured.target runtime.slice
-Wants=network-online.target configured.target runtime.slice
+After=network-online.target configured.target
+Wants=network-online.target configured.target
 
 [Service]
 Slice=runtime.slice
--- a/packages/kubernetes-1.17/kubelet-config
+++ b/packages/kubernetes-1.17/kubelet-config
@@ -74,6 +74,7 @@ systemReserved:
   {{~#each settings.kubernetes.system-reserved}}
   {{@key}}: "{{this}}"
   {{~/each}}
+systemReservedCgroup: "/system"
 {{~/if}}
 cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
 {{~#if settings.kubernetes.cpu-manager-reconcile-period}}
--- a/packages/kubernetes-1.17/kubelet-config
+++ b/packages/kubernetes-1.17/kubelet-config
@@ -68,6 +68,7 @@ kubeReserved:
   {{~/if}}
   {{~/if}}
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
+kubeReservedCgroup: "/runtime"
 {{~#if settings.kubernetes.system-reserved}}
 systemReserved:
   {{~#each settings.kubernetes.system-reserved}}
--- a/packages/release/release-tmpfiles.conf
+++ b/packages/release/release-tmpfiles.conf
@@ -2,4 +2,6 @@ C /etc/hosts - - - -
 C /etc/nsswitch.conf - - - -
 C /etc/wicked/ifconfig/eth0.xml - - - -
 d /var/log/kdump 0700 root root -
+d /sys/fs/cgroup/cpuset/runtime.slice 0755 root root -
+d /sys/fs/cgroup/hugetlb/runtime.slice 0755 root root -
 T /var/log/kdump - - - - security.selinux=system_u:object_r:secret_t:s0
--- a/packages/release/release-tmpfiles.conf
+++ b/packages/release/release-tmpfiles.conf
@@ -4,4 +4,6 @@ C /etc/wicked/ifconfig/eth0.xml - - - -
 d /var/log/kdump 0700 root root -
 d /sys/fs/cgroup/cpuset/runtime.slice 0755 root root -
 d /sys/fs/cgroup/hugetlb/runtime.slice 0755 root root -
+d /sys/fs/cgroup/cpuset/system.slice 0755 root root -
+d /sys/fs/cgroup/hugetlb/system.slice 0755 root root -
 T /var/log/kdump - - - - security.selinux=system_u:object_r:secret_t:s0
--- a/packages/kubernetes-1.17/kubelet.service
+++ b/packages/kubernetes-1.17/kubelet.service
@@ -6,6 +6,7 @@ Wants=configured.target
 BindsTo=containerd.service
 
 [Service]
+Slice=runtime.slice
 Type=notify
 EnvironmentFile=/etc/network/proxy.env
 EnvironmentFile=/etc/kubernetes/kubelet/env
--- a/packages/kubernetes-1.18/kubelet-config
+++ b/packages/kubernetes-1.18/kubelet-config
@@ -74,6 +74,7 @@ systemReserved:
   {{~#each settings.kubernetes.system-reserved}}
   {{@key}}: "{{this}}"
   {{~/each}}
+systemReservedCgroup: "/system"
 {{~/if}}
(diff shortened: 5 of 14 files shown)


cyrus-mc:
The one outstanding "problem" is that the slice/group directory is not created under every controller. For example under `cpuset` and `hugetlb` the `runtime.slice` directory is not created as it is under other controllers such as `cpu`.

When using `enforceNodeAllocatable` kubelet configuration setting (which doesn't appear to be supported as of yet in Bottlerocket) kubelet will fail to start with error

```
Failed to start ContainerManager Failed to enforce Kube Reserved Cgroup Limits on "/runtime": ["runtime"] cgroup does not exist
```

This is due to the fact that it doesn't find the cgroup under `cpuset` or `hugetlb`. I believe this to be a systemd configuration setting/problem but as of yet haven't found out how to correct.

packages/containerd/containerd.service
```diff
@@ -1,10 +1,11 @@
 [Unit]
 Description=containerd container runtime
 Documentation=https://containerd.io
-After=network-online.target configured.target
-Wants=network-online.target configured.target
+After=network-online.target configured.target runtime.slice
```
cyrus-mc: I guess I could also use WantedBy=slices.target to have slice always active after boot.

cyrus-mc (commented):
https://www.freedesktop.org/software/systemd/man/systemd.special.html

```
slices.target
A special target unit that sets up all slice units (see systemd.slice(5) for details) that shall always be active after boot. By default the generic system.slice slice unit as well as the root slice unit -.slice are pulled in and ordered before this unit (see below).

Adding slice units to slices.target is generally not necessary. Instead, when some unit that uses Slice= is started, the specified slice will be started automatically. Adding WantedBy=slices.target lines to the [Install] section should only be done for units that need to be always active. In that case care needs to be taken to avoid creating a loop through the automatic dependencies on "parent" slices.
```

So I removed the slice from After and Wants.

packages/kubernetes-1.17/kubelet-config
```diff
@@ -68,6 +68,7 @@ kubeReserved:
   {{~/if}}
   {{~/if}}
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
+kubeReservedCgroup: "/runtime"
```
bcressey, line 71: We'd want to conditionally render this based on some other setting, otherwise we'll immediately begin enforcing kube-reserved values instead of leaving them as "soft" limits for kubelet to track.

cyrus-mc: No that isn't that case. This just sets the cgroup. Enforcement is only done if kube-reserved is set in enforceNodeAllocatable setting.

bcressey: Ah, you're right. The docs are ambiguous but [the implementation](https://github.com/kubernetes/kubernetes/blob/9ff3b7e744b34c099c1405d9add192adbef0b6b1/pkg/kubelet/cm/node_container_manager_linux.go#L113) is pretty clear.

bcressey: In that case, why not also set `systemReservedCgroup: "/system"`?

cyrus-mc: Good point. I will add that setting as well.

cyrus-mc:
> The one outstanding "problem" is that the slice/group directory is not created under every controller. For example under `cpuset` and `hugetlb` the `runtime.slice` directory is not created as it is under other controllers such as `cpu`.
> 
> When using `enforceNodeAllocatable` kubelet configuration setting (which doesn't appear to be supported as of yet in Bottlerocket) kubelet will fail to start with error
> 
> ```
> Failed to start ContainerManager Failed to enforce Kube Reserved Cgroup Limits on "/runtime": ["runtime"] cgroup does not exist
> ```
> 
> This is due to the fact that it doesn't find the cgroup under `cpuset` or `hugetlb`. I believe this to be a systemd configuration setting/problem but as of yet haven't found out how to correct.

@bcressey This is the only outstanding issue I have. Currently I don't believe Bottlerocket allows setting (nor sets) the `enforceNodeAllocatable` setting. But once that is exposed and if `kube-reserved` is set the kubelet will fail for the reason stated above.

In my AMI for FedoraCoreOS I created a runtime.service that simply created the directories and that runtime.slice required. But I have to assume there is a more correct way to do that.

bcressey (approved):

bcressey:
> @bcressey This is the only outstanding issue I have. Currently I don't believe Bottlerocket allows setting (nor sets) the `enforceNodeAllocatable` setting. But once that is exposed and if `kube-reserved` is set the kubelet will fail for the reason stated above.
> 
> In my AMI for FedoraCoreOS I created a runtime.service that simply created the directories and that runtime.slice required. But I have to assume there is a more correct way to do that.

systemd doesn't manage `cpuset` or `hugetlb` under cgroup v1, per their docs on [controller support](https://github.com/systemd/systemd/blob/main/docs/CGROUP_DELEGATION.md#controller-support).

`mkdir` works to create the hierarchy but nothing would actually add processes from systemd's `runtime.slice` to `cgroup.procs` for those controllers. However, that might not matter since it doesn't appear that `kube-reserved` or `system-reserved` try to manage the corresponding resources.

ghost:
> > @bcressey This is the only outstanding issue I have. Currently I don't believe Bottlerocket allows setting (nor sets) the `enforceNodeAllocatable` setting. But once that is exposed and if `kube-reserved` is set the kubelet will fail for the reason stated above.
> > In my AMI for FedoraCoreOS I created a runtime.service that simply created the directories and that runtime.slice required. But I have to assume there is a more correct way to do that.
> 
> systemd doesn't manage `cpuset` or `hugetlb` under cgroup v1, per their docs on [controller support](https://github.com/systemd/systemd/blob/main/docs/CGROUP_DELEGATION.md#controller-support).
> 
> `mkdir` works to create the hierarchy but nothing would actually add processes from systemd's `runtime.slice` to `cgroup.procs` for those controllers. However, that might not matter since it doesn't appear that `kube-reserved` or `system-reserved` try to manage the corresponding resources.

https://github.com/kubernetes/kubernetes/blob/9ff3b7e744b34c099c1405d9add192adbef0b6b1/pkg/kubelet/cm/cgroup_manager_linux.go#L282

Checks for the existence of the controllers `whitelisted`. I will add to release package creation of directories to satisfy cgroup hierarchy.

packages/release/release-tmpfiles.conf
```diff
@@ -2,4 +2,6 @@ C /etc/hosts - - - -
 C /etc/nsswitch.conf - - - -
 C /etc/wicked/ifconfig/eth0.xml - - - -
 d /var/log/kdump 0700 root root -
+d /sys/fs/cgroup/cpuset/runtime.slice 0755 root root -
```
matthewce-dat, line 5: @bcressey create directories via tmpfiles so that kubeReserved can be enforced without throwing error.

cyrus-mc, line 5: @bcressey created cgroup hierarchy via tmpfiles.

bcressey (approved):

cyrus-mc:
@samuelkarp Would like to get this reviewed so can be included in next release.

samuelkarp (approved):
There's a small concern with the effects this change will have on the `aws-ecs-1` variant, since Amazon ECS does not expect containerd to be in a separate cgroup.  However, because none of the cgroup settings (other than the hierarchy) are changed, the only effect should be in passive hierarchy-related behavioral differences (for example, CPU share weights under contention).  I think it's reasonable to take this change, and if we end up deciding that we should treat ECS differently later we can always do so.

LGTM

PR #1695, opened 2021-08-04
logdog: include `/var/log/kdump` in the logdog tarball

arnaldo2792:
**Issue number:**
N / A

**Description of changes:**

```
51cb5257 logdog: include `/var/log/kdump` in the logdog taball
```

This commit adds `/var/log/kdump` so that logdog collects crash kernel dumps when they exist.

```
61402201 logdog: change default output directory for tarball
```

This commit changes the default output directory for the tarball created by logdog, from  `/tmp` to `/var/log/support`.

**Testing done:**

In aws-dev, which has kdump support enabled: I verified that the files generated by `prairiedog` were collected by `logdog`:

```sh
tar --list -f /.bottlerocket/rootfs/var/log/support/bottlerocket-logs.tar.gz | grep var:
bottlerocket-logs/var
bottlerocket-logs/var/log
bottlerocket-logs/var/log/kdump
bottlerocket-logs/var/log/kdump/dmesg.log
bottlerocket-logs/var/log/kdump/vmcore.dump
bottlerocket-logs/var/log/kdump/prairiedog.log
``` 

In aws-ecs-1, which doesn't have kdump support enabled, I verified that `logdog` still works:

```sh
tar --list -f /.bottlerocket/rootfs/var/log/support/bottlerocket-logs.tar.gz
bottlerocket-logs/
bottlerocket-logs/logdog.errors
bottlerocket-logs/containerd-config
bottlerocket-logs/containerd-config-host
bottlerocket-logs/df
bottlerocket-logs/df-inodes
bottlerocket-logs/dmesg
bottlerocket-logs/iptables-filter
bottlerocket-logs/iptables-nat
bottlerocket-logs/journalctl-boots
bottlerocket-logs/journalctl.errors
bottlerocket-logs/journalctl.log
bottlerocket-logs/proc-mounts
bottlerocket-logs/settings.json
bottlerocket-logs/signpost
bottlerocket-logs/wicked
bottlerocket-logs/os-release
bottlerocket-logs/docker-info
bottlerocket-logs/docker-daemon.json
bottlerocket-logs/ecs-config.json
bottlerocket-logs/ecs-tasks

```

**Terms of contribution:**

### DIFF
--- a/sources/logdog/conf/logdog.common.conf
+++ b/sources/logdog/conf/logdog.common.conf
@@ -14,3 +14,4 @@ exec settings.json apiclient --method GET --uri /
 exec signpost signpost status
 exec wicked wicked show all
 file os-release /etc/os-release
+glob /var/log/kdump/*
--- a/README.md
+++ b/README.md
@@ -642,14 +642,14 @@ sudo sheltie
 logdog
 ```
 
-This will write an archive of the logs to `/tmp/bottlerocket-logs.tar.gz`.
+This will write an archive of the logs to `/var/log/support/bottlerocket-logs.tar.gz`.
 You can use SSH to retrieve the file.
 Once you have exited from the Bottlerocket host, run a command like:
 
 ```bash
 ssh -i YOUR_KEY_FILE \
     ec2-user@YOUR_HOST \
-    "cat /.bottlerocket/rootfs/tmp/bottlerocket-logs.tar.gz" > bottlerocket-logs.tar.gz
+    "cat /.bottlerocket/rootfs/var/log/support/bottlerocket-logs.tar.gz" > bottlerocket-logs.tar.gz
 ```
 
 For a list of what is collected, see the logdog [command list](sources/logdog/src/log_request.rs).
--- a/packages/release/release-tmpfiles.conf
+++ b/packages/release/release-tmpfiles.conf
@@ -3,3 +3,5 @@ C /etc/nsswitch.conf - - - -
 C /etc/wicked/ifconfig/eth0.xml - - - -
 d /var/log/kdump 0700 root root -
 T /var/log/kdump - - - - security.selinux=system_u:object_r:secret_t:s0
+d /var/log/support 0755 root root -
+T /var/log/support - - - - security.selinux=system_u:object_r:secret_t:s0
--- a/sources/logdog/README.md
+++ b/sources/logdog/README.md
@@ -10,7 +10,7 @@ into a tarball for easy export.
 Usage example:
 ```rust
 $ logdog
-logs are at: /tmp/bottlerocket-logs.tar.gz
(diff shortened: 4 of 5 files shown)


webern (approved):
👍

sources/logdog/conf/logdog.common.conf
```diff
@@ -14,3 +14,4 @@ exec settings.json apiclient --method GET --uri /
 exec signpost signpost status
 exec wicked wicked show all
 file os-release /etc/os-release
+glob /var/log/kdump/*.*
```
tjkirch: Do we believe there will always be a `.` in the name?  Are we trying to prevent collection of something that doesn't have a `.`?

tjkirch (approved):

sources/logdog/conf/logdog.common.conf
arnaldo2792: > Do we believe there will always be a . in the name? 

Yes, we name the files in `prairiedog`

>   Are we trying to prevent collection of something that doesn't have a .?

No, I think I can use `*`, I'll test.

bcressey:
This is an existing problem but the kdump use case exacerbates it - we write the archive to /tmp which is a memory-backed tmpfs.

It'd be better to write it somewhere under `/var`, preferably in a directory labeled `secret_t`.

arnaldo2792:
Forced pushes include:

- Move logdog's output from `tmp` to `/var/log/support` with proper permissions
- Fix documentation to reflect the change
- Sorry for the spam, I forgot the `Sign Off` line, and had to amend the commits

bcressey (approved):

tjkirch (approved):

webern:
> This is an existing problem but the kdump use case exacerbates it - we write the archive to /tmp which is a memory-backed tmpfs.
> 
> It'd be better to write it somewhere under `/var`, preferably in a directory labeled `secret_t`.

It's still accessible from 'outside' sudo sheltie under `/.bottlerocket`? Just trying to make sure I can still use SSH to get the tarball from my local machine via admin container.

arnaldo2792:
>  It's still accessible from 'outside' sudo sheltie under /.bottlerocket? Just trying to make sure I can still use SSH to get the tarball from my local machine via admin container.

Yes:

```fish
agarrcia@ ~> ssh ec2-user@HOST "cat /.bottlerocket/rootfs/var/log/support/bottlerocket-logs.tar.gz" > bottlerocket-logs.tar.gz
agarrcia@ ~> file bottlerocket-logs.tar.gz
bottlerocket-logs.tar.gz: gzip compressed data, original size modulo 2^32 224256
```

webern (approved):
Nice.

nit: one of the commit messages has a spelling error `taball`.

arnaldo2792:
Forced push fixes commits' messages

PR #1701, opened 2021-08-07
update grub to 2.06 and implement hybrid boot support for x86_64

### LINKED ISSUE #1697: update grub to 2.06
We should update to the latest GRUB from Amazon Linux, which in turn is based on newer Fedora builds.

This includes fixes for console handling in EC2, especially for metal instances, as well as improvements in the way Secure Boot is managed.

bcressey:
**Issue number:**
Fixes #1697

**Description of changes:**
Update grub to 2.06, from Amazon Linux by way of Fedora. Build both EFI and BIOS images for GRUB.

Adjust default grub configurations to take advantage of the improved console support on EFI and BIOS, and to disable module loading for security reasons.

Implement hybrid support for x86_64 images, so they work with either BIOS or EFI firmware.

**Testing done:**
Verified that the GPT priority functionality works correctly by following these steps for x86_64 (BIOS, EFI) and aarch64 (EFI):
* launch a test 1.2.0 image
* update to a test 1.3.0 image
* verify that 1.3.0 comes up, reboot, and confirm it's still 1.3.0
* downgrade to a test 1.2.0 image
* verify that 1.2.0 comes up, reboot, and confirm it's still 1.2.0
* update to a defective 1.4.0 image
* verify that the boot fails, reboot, and confirm that it's 1.2.0

Verified that the grub console output was visible on these platforms:
* EC2 virtualized, for x86_64 (BIOS, EFI) and aarch64 (EFI)
* EC2 metal, for x86_64 (BIOS) and aarch64 (EFI)
* VMware guest, for x86_64 (BIOS)

Output was not visible for the VMware guest under EFI, although the guest booted OK. I don't consider this a blocker since we default to BIOS on that platform.

**Terms of contribution:**


bcressey:
Canceled the workflow since I still need to upload the grub source rpm to the lookaside cache.

packages/grub/efi.cfg
```diff
@@ -1,3 +1,5 @@
+set no_modules=y
+terminal_output console
```
arnaldo2792, line 2: Is it possible to have both EFI and BIOS output to `serial`, or are `serial` and `console` the same thing?

packages/grub/grub.spec
```diff
@@ -74,12 +78,42 @@ Summary: Tools for the bootloader with support for Linux and more
...
+tar -xof ../%{gnulib_fixes}.tar.gz; rm ../%{gnulib_fixes}.tar.gz
+mv %{gnulib_fixes} gnulib
+pushd gnulib
+patch -p1 < ../../gnulib-amzn2-cflags.patch
+rm ../../gnulib-amzn2-cflags.patch
+popd
 cp unicode/COPYING COPYING.unicode
+rm -f configure
+
+# Apply upstream and local patches.
```
arnaldo2792, line 105: Is it required to use git to apply the patches?

packages/grub/0001-setup-Add-root-device-argument-to-grub-setup.patch
```diff
@@ -1,4 +1,4 @@
-From 9ddb865df4ea2fb79edaf6c3a8b0122796800014 Mon Sep 17 00:00:00 2001
+From fa856e9d6cce8cac8f55bb00392725961c264b8f Mon Sep 17 00:00:00 2001
```
arnaldo2792, line 1: Nit: should this patch include a Sign-off comment since you did some refactor?

packages/grub/grub.spec
bcressey: It's not required, but it's convenient and leaves the tree in a "clean" state with all patches applied as commits.

That's helpful for rebasing patches since you can run just the prep phase (`rpmbuild -bp <spec>`) and then work with the Git repo.

packages/grub/0001-setup-Add-root-device-argument-to-grub-setup.patch
bcressey: Sure, I can add that.

packages/grub/grub.spec
```diff
@@ -77,9 +77,37 @@ Summary: Tools for the bootloader with support for Linux and more
 %description tools
 %{summary}.
 
+%global gnulib_fixes gnulib-fixes-0e9febb5e
```
tjkirch: This lives in the srpm, and I'm assuming the hash will change over time; can we have this variable up next to the Version/Source so it's more obvious that we need to at least check it?

tools/rpm2img
```diff
@@ -60,36 +60,36 @@ VERITY_HASH_BLOCK_SIZE=4096
...
+BIOS_TYPECODE="ef02"
 
-if [[ "${ARCH}" == "x86_64" ]]; then
-  FIRM_NAME="BIOS-BOOT"
-  BOTTLEROCKET_FIRM_TYPECODE="ef02"
-else
-  FIRM_NAME="EFI_BOOT"
-  BOTTLEROCKET_FIRM_TYPECODE="${EFI_SYSTEM_TYPECODE}"
-fi
+# We'll eventually want to support an active/passive scheme for EFI partitions,
```
tjkirch: It took me a bit of offline discussion to understand what's going on with the EFI partitions, and it boiled down to one key point that might be worth putting into a comment in addition to this explanation - we'll be using the grub stage 2 binary from the EFI partition in the _first_ partition set regardless of the _target_ partition set.  It'll then use the counters to continue booting into the target partition set.  I was confused because of the single BIOS boot partition versus the two EFI partitions, because in all(?) other cases we only use artifacts from the target set.

tjkirch (approved):
Nice work.

packages/grub/efi.cfg
bcressey: I asked BenH about this and will try to summarize his answer. 😁 

EFI and BIOS both support "console" and "serial" output.

For EFI, `console` is an API that's used to interact with the underlying hardware. We could try to bypass it by using `serial`, but that's not a great idea because EFI "owns" the device at this point.

For BIOS, `console` involves using BIOS interrupt calls for output. This can work but is inconsistent; for example, we don't get output on EC2 metal instances.

Using `serial` for BIOS lets us avoid the inconsistency on the platforms we support today, but might be the wrong choice on other platforms in the future if we run into something that needs `gfxterm` instead.

Using `console` for EFI should always be the correct choice.

arnaldo2792 (approved):
:motor_boat:

tools/rpm2img
bcressey: I've expanded the commentary and changed the variable names to avoid confusion with the A/B partition sets.

tjkirch (approved):

PR #1709, opened 2021-08-11
Add constants crate

### LINKED ISSUE #123: Stop hardcoding API constants across several programs
Currently we have a few constants sprinkled around early-boot-config/sundog/tbs, etc.  There needs to be a configuration crate or package where we can centralize these.

This package would contain things like:
* datastore path
* api address/socket
* ...

arnaldo2792:
**Issue number:**
#123

**Description of changes:**

```
7c3cdbe5 sources: use shared constants crate in crates
5b17ca46 sources: add constants crate
56e97663 sources: remove unused struct from apiserver mod
```

This change adds  the `constants` crate and  updates the following crates to use it:
- apiclient
- bootstrap-containers
- certdog
- corndog
- early-boot-config
- ecs-settings-applier
- host-containers
- pluto
- schnauzer
- service-dog
- settings-commiter
- static-pods
- storewolf
- sundog
- thar-be-settings
- thar-be-updates

As part of this PR, I included a small nit to removed an unused struct imported in `sources/api/apiserver/src/server/mod.rs`.

**Testing done:**

**aws-k8s-1.21, aws-k8s-1.19, aws-ecs-1**:
- I ran an nginx pod/task and called `curl http://localhost` from within the pod/task
- `systemctl status` didn't show any failures
- I let the workloads run for +1 hour and `journalctl -p notice` didn't show any alarming message

**Terms of contribution:**

### DIFF
--- a/sources/api/bootstrap-containers/src/main.rs
+++ b/sources/api/bootstrap-containers/src/main.rs
@@ -87,20 +87,14 @@ use std::fs;
 use std::path::Path;
 use std::process::{self, Command};
 use std::str::FromStr;
+use constants;
 
 use model::modeled_types::{BootstrapContainerMode, Identifier};
-
-// FIXME Get from configuration in the future
-const DEFAULT_API_SOCKET: &str = "/run/api.sock";
-const API_SETTINGS_URI: &str = "/settings";
 
 const ENV_FILE_DIR: &str = "/etc/bootstrap-containers";
 const DROPIN_FILE_DIR: &str = "/etc/systemd/system";
 const PERSISTENT_STORAGE_DIR: &str = "/local/bootstrap-containers";
 const DROP_IN_FILENAME: &str = "overrides.conf";
-
-const SYSTEMCTL_BIN: &str = "/bin/systemctl";
-const HOST_CTR_BIN: &str = "/bin/host-ctr";
 
 /// Stores user-supplied global arguments
 #[derive(Debug)]
@@ -113,7 +107,7 @@ impl Default for Args {
     fn default() -> Self {
         Self {
             log_level: LevelFilter::Info,
-            socket_path: DEFAULT_API_SOCKET.to_string(),
+            socket_path: constants::API_SOCKET.to_string(),
         }
     }
 }
@@ -150,7 +144,7 @@ fn usage() {
         --mode MODE
 
     Socket path defaults to {}",
-        program_name, DEFAULT_API_SOCKET,
+        program_name, constants::API_SOCKET,
     );
 }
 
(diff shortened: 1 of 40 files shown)


arnaldo2792:
Forced push includes:

- Rename `bottlerocket-constants` to `constants`

webern:
> * Rename `bottlerocket-constants` to `constants`

I was going to suggest that change. 😎

sources/api/bootstrap-containers/src/main.rs
```diff
@@ -87,21 +87,15 @@ use std::fs;
 use std::path::Path;
 use std::process::{self, Command};
 use std::str::FromStr;
+use constants;
```
webern, line 90: I would probably import the constant names instead of qualifying them everywhere. (Same in all places...)

```suggestion
use constants::{API_SOCKET, HOST_CTR_BIN, SYSTEMCTL_BIN};
```

tjkirch: I think this is a case where the opposite is justifiable - qualified constants make it clear that they're not crate-specific, where unqualified ones look like "standard" constants you'd expect in the same file.

webern: > qualified constants make it clear that they're not crate-specific, where unqualified ones look like "standard" constants you'd expect in the same file.

I guess I don't have the expectation that constants are in the same file. To me they're like all other symbols. That is, I have to jump to definition to know where they are. Not a big deal though.

sources/constants/README.md
```diff
@@ -0,0 +1,9 @@
+# constants
+
+Current version: 0.1.0
+
+  This crate contains constants shared accross multiple bottlerocket's crates
```
zmrow: Across is misspelled :)

I think this should probably read `shared across multiple Bottlerocket crates`

zmrow (commented):
Just one nit - thanks for doing this!

arnaldo2792:
Forced  push includes fix in crate's doc string

sources/api/servicedog/src/main.rs
```diff
@@ -24,16 +24,11 @@ use std::env;
 use std::ffi::OsStr;
 use std::process::{self, Command};
 use std::str::FromStr;
+use constants;
```
bcressey, line 27: $1.00 if we delete `servicedog` since we haven't used it for anything in over a year 😀

bcressey (approved):

sources/api/servicedog/src/main.rs
arnaldo2792: You will owe me $2 dollars :money_mouth_face:

webern (approved):

PR #1710, opened 2021-08-11
kubelet, pluto: support IPv6

### LINKED ISSUE #1696: aws-k8s variants: support IPv6
**What I'd like:**
IPv6 support for Bottlerocket aws-k8s-* nodes.

Once EKS fully supports IPv6 (being tracked in https://github.com/aws/containers-roadmap/issues/835), Bottlerocket (specifically, `pluto`) needs to be able to determine whether the cluster's IP address family is IPv6 or IPv4 and dynamically set `settings.kubernetes.node-ip` and `settings.kubernetes.cluster-dns-ip` accordingly.

N/A

etungsten:
**Issue number:**
Closes https://github.com/bottlerocket-os/bottlerocket/issues/1696

**Description of changes:**
```
Date:   Mon Sep 13 15:59:59 2021 -0700

    kubernetes-1.21: patch kubelet to support ipv6 with AWS
    
    This adds a patch to the in-tree AWS cloud provider code to make kubelet
    append IPv6 addresses to the list of NodeAddresses.

```

```
Date:   Mon Aug 9 12:27:23 2021 -0700

    pluto: support IPv6
    
    This adds support for retrieving IPv6 node IP address for IPv6 EKS
    clusters


```

```
Date:   Fri Aug 13 11:00:47 2021 -0700

    release: enable IPv6 forwarding

```

The testing below has been done with IPv6 changes in aws-vpc-cni that has yet to be formally released: https://github.com/aws/amazon-vpc-cni-k8s/pull/1587

**Testing done:**

**Testing for IPv6 EKS Clusters:**
Created an AMI off of the built images and launched an EC2 instance in my IPv6 EKS cluster.
The node is able to get an IPv6 internal IP address:
```
$ kubectl get nodes -o wide
NAME                                          STATUS   ROLES    AGE   VERSION   INTERNAL-IP                              EXTERNAL-IP      OS-IMAGE                KERNEL-VERSION   CONTAINER-RUNTIME
ip-192-168-3-236.us-west-2.compute.internal   Ready    <none>   26m   v1.21.3   2600:1f13:8ff:6600:6a76:135f:ee92:3395   54.188.180.229   Bottlerocket OS 1.2.0   5.10.50          containerd://1.4.8+bottlerocket
```

Can run pods without problem. The pods get IPv6 addresses assigned without problem. I can ping external ipv4 and ipv6 addresses without problem.

**Testing for IPv4 EKS Clusters:**
Launched node with AMI in IPv4 cluster.
Node becomes Ready with the correct `node-ip` and `cluster-dns-ip`:
```
$ kubectl get nodes -o wide
NAME                                           STATUS   ROLES    AGE   VERSION   INTERNAL-IP      EXTERNAL-IP     OS-IMAGE                KERNEL-VERSION   CONTAINER-RUNTIME
ip-192-168-1-232.us-west-2.compute.internal    Ready    <none>   28s   v1.21.3   192.168.1.232    54.190.54.101   Bottlerocket OS 1.2.0   5.10.50          containerd://1.4.8+bottlerocket
```
Can run pods without problem.

Ran conformance tests with an IPv6 cluster and all tests passed.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,7 +368,7 @@ Static pods can be particularly useful when running in standalone mode.
 For Kubernetes variants in AWS and VMware, the following are set for you automatically, but you can override them if you know what you're doing!
 In AWS, [pluto](sources/api/) sets these based on runtime instance information.
 In VMware, Bottlerocket uses [netdog](sources/api/) (for `node-ip`) or relies on [default values](sources/models/src/vmware-k8s-1.21/defaults.d/).
-* `settings.kubernetes.node-ip`: The IPv4 address of this node.
+* `settings.kubernetes.node-ip`: The IP address of this node.
 * `settings.kubernetes.pod-infra-container-image`: The URI of the "pause" container.
 * `settings.kubernetes.kube-reserved`: Resources reserved for node components.
   * Bottlerocket provides default values for the resources by [schnauzer](sources/api/):
@@ -378,7 +378,7 @@ In VMware, Bottlerocket uses [netdog](sources/api/) (for `node-ip`) or relies on
 
 For Kubernetes variants in AWS, the following settings are set for you automatically by [pluto](sources/api/).
 * `settings.kubernetes.max-pods`: The maximum number of pods that can be scheduled on this node (limited by number of available IPv4 addresses)
-* `settings.kubernetes.cluster-dns-ip`: Derived from the EKS IPV4 Service CIDR or the CIDR block of the primary network interface.
+* `settings.kubernetes.cluster-dns-ip`: Derived from the EKS Service IP CIDR or the CIDR block of the primary network interface.
 
 #### Amazon ECS settings
 
--- a/packages/release/release-sysctl.conf
+++ b/packages/release/release-sysctl.conf
@@ -28,6 +28,12 @@ net.ipv4.ip_default_ttl = 255
 # Enable IPv4 forwarding for container networking.
 net.ipv4.conf.all.forwarding = 1
 
+# Enable IPv6 forwarding for container networking.
+net.ipv6.conf.all.forwarding = 1
+
+# Accept router advertisement (RA) packets even if IPv6 forwarding is enabled on eth0
+net.ipv6.conf.eth0.accept_ra = 2
+
 # This is generally considered a safe ephemeral port range
 net.ipv4.ip_local_port_range = 32768 60999
 
--- a/sources/api/pluto/README.md
+++ b/sources/api/pluto/README.md
@@ -15,7 +15,7 @@ It uses IMDS to get information such as:
 
 It uses EKS to get information such as:
 
-- Service IPV4 CIDR
+- Service IP CIDR
 
 It uses the Bottlerocket API to get information such as:
 
--- a/sources/api/pluto/src/main.rs
+++ b/sources/api/pluto/src/main.rs
@@ -14,7 +14,7 @@ It uses IMDS to get information such as:
 
 It uses EKS to get information such as:
 
-- Service IPV4 CIDR
+- Service IP CIDR
 
 It uses the Bottlerocket API to get information such as:
 
@@ -40,6 +40,8 @@ use imdsclient::ImdsClient;
 use snafu::{ensure, OptionExt, ResultExt};
 use std::fs::File;
 use std::io::{BufRead, BufReader};
+use std::net::IpAddr;
+use std::str::FromStr;
 use std::string::String;
 use std::{env, process};
 
@@ -51,14 +53,31 @@ const DEFAULT_10_RANGE_DNS_CLUSTER_IP: &str = "172.20.0.10";
 const ENI_MAX_PODS_PATH: &str = "/usr/share/eks/eni-max-pods";
 
 mod error {
+    use crate::api;
     use crate::eks;
     use snafu::Snafu;
+    use std::net::AddrParseError;
 
     #[derive(Debug, Snafu)]
     #[snafu(visibility = "pub(super)")]
     pub(super) enum PlutoError {
+        #[snafu(display(
+            "Unable to retrieve cluster name and AWS region from Bottlerocket API: {}",
+            source
+        ))]
+        AwsInfo { source: api::Error },
+
+        #[snafu(display("Missing field '{}' in EKS network config response", field))]
+        MissingNetworkConfig { field: &'static str },
+
         #[snafu(display("Unable to parse CIDR '{}': {}", cidr, reason))]
         CidrParse { cidr: String, reason: String },
+
+        #[snafu(display("Unable to parse IP '{}': {}", ip, source))]
+        BadIp { ip: String, source: AddrParseError },
+
+        #[snafu(display("No IP address found for this host"))]
+        NoIp,
 
         #[snafu(display("IMDS request failed: {}", source))]
         ImdsRequest { source: imdsclient::Error },
@@ -144,42 +163,31 @@ async fn get_max_pods(client: &mut ImdsClient) -> Result<String> {
     error::NoInstanceTypeMaxPods { instance_type }.fail()
 }
 
-/// Returns the cluster's DNS IPV4 address. First it attempts to call EKS describe-cluster to find
-/// the `serviceIPv4CIDR`. If that works, it returns the expected cluster DNS IP address which is
-/// obtained by substituting `10` for the last octet. If the EKS call is not successful, it falls
-/// back to using IMDS MAC CIDR blocks to return one of two default addresses.
+/// Returns the cluster's DNS address.
+///
+/// For IPv4 clusters, first it attempts to call EKS describe-cluster to find the `serviceIpv4Cidr`.
+/// If that works, it returns the expected cluster DNS IP address which is obtained by substituting
+/// `10` for the last octet. If the EKS call is not successful, it falls back to using IMDS MAC CIDR
+/// blocks to return one of two default addresses.
 async fn get_cluster_dns_ip(client: &mut ImdsClient) -> Result<String> {
-    // try calling eks describe-cluster to figure out the dns cluster ip
-    if let Some(dns_ip) = get_dns_from_eks().await {
-        // we were able to calculate the dns ip from the cidr range we received from eks
-        return Ok(dns_ip);
-    }
-
-    // we were unable to obtain or parse the cidr range from eks, fallback to one of two default
-    // values based on the cidr range of our primary network interface
-    get_cluster_dns_from_imds_mac(client).await
-}
-
-/// Gets the Service IPV4 CIDR setting from EKS and parses it to calculate the cluster DNS IP.
-/// Prints the error and returns `None` if anything goes wrong.
-async fn get_dns_from_eks() -> Option<String> {
-    let aws_k8s_info = match api::get_aws_k8s_info().await {
-        Ok(value) => value,
-        Err(e) => {
-            eprintln!(
-                "Unable to get region and cluster name from Bottlerocket API, using default DNS IP: {}",
-                e
-            );
-            return None;
+    // Retrieve the kubernetes network configuration for the EKS cluster
+    if let Ok(aws_k8s_info) = api::get_aws_k8s_info().await.context(error::AwsInfo) {
+        if let Ok(config) =
+            eks::get_cluster_network_config(&aws_k8s_info.region, &aws_k8s_info.cluster_name)
+                .await
+                .context(error::EksError)
+        {
+            // Derive cluster-dns-ip from the service IPv4 CIDR
+            if let Some(ipv4_cidr) = config.service_ipv_4_cidr {
+                if let Ok(dns_ip) = get_dns_from_ipv4_cidr(&ipv4_cidr) {
+                    return Ok(dns_ip);
+                }
+            }
         }
-    };
-
-    eks::get_cluster_cidr(&aws_k8s_info.region, &aws_k8s_info.cluster_name)
-        .await
-        .context(error::EksError)
-        .and_then(|cidr| get_dns_from_cidr(&cidr))
-        .map_err(|e| eprintln!("Unable to parse CIDR from EKS, using default DNS IP: {}", e))
-        .ok()
+    }
+    // If we were unable to obtain or parse the cidr range from EKS, fallback to one of two default
+    // values based on the IPv4 cidr range of our primary network interface
+    get_ipv4_cluster_dns_ip_from_imds_mac(client).await
 }
 
 /// Replicates [this] logic from the EKS AMI:
@@ -188,7 +196,7 @@ async fn get_dns_from_eks() -> Option<String> {
 /// DNS_CLUSTER_IP=${SERVICE_IPV4_CIDR%.*}.10
 /// ```
 /// [this]: https://github.com/awslabs/amazon-eks-ami/blob/732b6b2/files/bootstrap.sh#L335
-fn get_dns_from_cidr(cidr: &str) -> Result<String> {
+fn get_dns_from_ipv4_cidr(cidr: &str) -> Result<String> {
     let mut split: Vec<&str> = cidr.split('.').collect();
     ensure!(
         split.len() == 4,
@@ -203,7 +211,7 @@ fn get_dns_from_cidr(cidr: &str) -> Result<String> {
 
 /// Gets gets the the first VPC IPV4 CIDR block from IMDS. If it starts with `10`, returns
 /// `10.100.0.10`, otherwise returns `172.20.0.10`
-async fn get_cluster_dns_from_imds_mac(client: &mut ImdsClient) -> Result<String> {
+async fn get_ipv4_cluster_dns_ip_from_imds_mac(client: &mut ImdsClient) -> Result<String> {
     // Take the first (primary) MAC address. Others may exist from attached ENIs.
     let mac = client
         .fetch_mac_addresses()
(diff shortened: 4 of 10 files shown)


sources/api/pluto/src/main.rs
```diff
@@ -144,42 +162,37 @@ async fn get_max_pods(client: &mut ImdsClient) -> Result<String> {
...
+async fn get_cluster_dns_ip(
+    client: &mut ImdsClient,
+    network_config: KubernetesNetworkConfigResponse,
+) -> Result<String> {
+    if network_config.ip_family == Some("ipv6".to_string()) {
+        // For IPv6 clusters, `::a` is equivalent to `.10` for IPv4
+        return Ok(
+            DEFAULT_SERVICE_IPV6_CIDR.split('/').collect::<Vec<&str>>()[0].to_string() + "a",
+        );
     }
```
webern: I don't like the way EKS-specifics are bubbling up into `main.rs`. I think I missed an opportunity initially to have the EKS function be `eks::get_cluster_dns_ip` instead of `eks::get_cluster_cidr`. If the function were `eks::get_cluster_dns_ip` then all of your changes could have been in `eks.rs`, right?

Can we do this now? Make the function `eks::get_dns_ip`. If IPv6, return the currently hard-coded address. If custom IPV4 CIDR is indicated, return the correct address for that CIDR. Otherwise return None.

You might need more than just the IP. In that case, in eks.rs:

```rust
pub(crate) enum IpFamily {
    V4,
    V6,
}

pub(crate) EksInfo {
    ip_family: IpFamily,
    cluster_dns_ip: Option<String> // An actual IP Address type might be nice here instead of String
}

pub(crate) async fn get_eks_info(region: &str, cluster: &str) -> Result<EksInfo> {

}
```

Ultimately I'm hoping that we can try to separate the interpretation of Rusoto types from the rest of the program by keeping them hidden in eks.rs.

sources/api/pluto/src/main.rs
```diff
@@ -144,42 +162,37 @@ async fn get_max_pods(client: &mut ImdsClient) -> Result<String> {
...
+/// (The `::a` is equivalent to `.10` for IPv4)
+/// FIXME Update this once EKS changes their approach for determining the IPv6 service CIDR
+async fn get_cluster_dns_ip(
+    client: &mut ImdsClient,
+    network_config: KubernetesNetworkConfigResponse,
+) -> Result<String> {
+    if network_config.ip_family == Some("ipv6".to_string()) {
+        // For IPv6 clusters, `::a` is equivalent to `.10` for IPv4
+        return Ok(
+            DEFAULT_SERVICE_IPV6_CIDR.split('/').collect::<Vec<&str>>()[0].to_string() + "a",
```
webern: Since this constant is only used here. Why not store the actual string you need instead of manipulating it here?

etungsten: I wanted to show how the cluster DNS IP is derived from the service IPv6 CIDR. (i.e. setting the last hextet to `a`). This value is the way it is because EKS has chosen a fixed ULA for the default service IPv6 CIDR.

etungsten: Also, we'd need this manipulation down the road once EKS decides to not hardcode the service IPv6 CIDR and just generate a unique one per cluster (which is likely).

sources/api/pluto/src/main.rs
etungsten: Good idea. I'll try it out.

etungsten: Hmm, can you elaborate a bit more on why it's not good to have pluto's main.rs import `rusoto_eks`/EKS specific stuff?

`pluto` is already EKS specific (we only build `pluto` for aws-k8s variants) so I don't think it hurts to have it have EKS references in its code.

I tried the suggestion out a bit and started realizing I'm basically reconstructing the `rusoto_eks::KubernetesNetworkConfigResponse` struct in eks.rs just so main.rs wouldn't have to import something from `rusoto_eks` and it seems like a waste if I can just re-use that struct definition directly.

etungsten: I still think the `eks.rs` and `main.rs` separation make sense since we only make rusoto EKS API calls in `eks.rs`. But I don't think it's necessary to separate all EKS references in `main.rs` to `eks.rs` since that would almost be everything in `main.rs`.

sources/api/pluto/src/main.rs
etungsten: EKS changed approach to generating a random ULA for each cluster instead of hardcoding it. So now I'm using `serviceIpv6Cidr` directly

sources/api/pluto/src/main.rs
webern: > basically reconstructing the rusoto_eks::KubernetesNetworkConfigResponse struct in eks.rs just so main.rs wouldn't have to import something from rusoto_eks and it seems like a waste if I can just re-use that struct definition directly.

Basically all of pluto is an implementation detail, so I wouldn't block on this. But my thinking for pushing code into eks.rs and creating our own `EksInfo` struct instead of using `rusoto_eks::KubernetesNetworkConfigResponse` is as follows:
- main.rs has a lot going on and it's honestly pretty hard to understand what's happening (it was before your changes and got worse). If you can a offload some code and put it elsewhere, and call it with something like "get the dns cluster ip if possible", then you offload some of the cognitive burdon in main.rs.
- Creating a small type doesn't cost anything, and if it returns exactly the data needed by a caller in exactly the way the caller needs it, then I think that's better than borrowing a (potentially bloated) type from a library.
- Changing from Rusoto to the new SDK would be isolated to eks.rs.
- `KubernetesNetworkConfigResponse` is a ridiculously long type name and I hate seeing it everywhere :)

> I don't think it's necessary to separate all EKS references in main.rs to eks.rs

I think of eks.rs more as rusoto encapsulation than EKS encapsulation.

Anyway, in summary, it's not a big deal. The main feedback is that main.rs is very dense and getting denser.

etungsten:
Push above addresses @webern 's comments.

- Removed rusoto related imports from `main.rs`
- Created new type alias for `rusoto_eks::KubernetesNetworkConfigResponse` for sake of brevity
- Fixed unit tests

sources/api/pluto/src/main.rs
etungsten: > main.rs has a lot going on and it's honestly pretty hard to understand what's happening (it was before your changes and got worse). If you can a offload some code and put it elsewhere, and call it with something like "get the dns cluster ip if possible", then you offload some of the cognitive burdon in main.rs.

I think the issue is that all the logic here is tightly coupled so if we want to move `get_cluster_dns_ip` and `get_node_ip` to eks.rs then all the helpers/unit-tests/constants in main.rs will also move alongside, leaving main.rs with just the `run` and `main` function. So someone trying to understand what's going on `pluto` essentially would have to look at the functions, whether they be in `main.rs` or `eks.rs`. If they just want to understand at a high level of what's going on, I think the README at the top of main.rs and `run()` explains things pretty well.
And I don't particularly think the logic is that hard to understand with the docstrings and inline comments.


> Creating a small type doesn't cost anything, and if it returns exactly the data needed by a caller in exactly the way the caller needs it, then I think that's better than borrowing a (potentially bloated) type from a library.

Currently `KubernetesNetworkConfigResponse` contains exactly what we need, no more no less: `ipfamily`, `serviceipv4cidr`, `serviceipv6cidr`. If the type adds a bunch of stuff we don't need or if we need more information from other parts of the cluster struct down the road, we can always revisit and just create our own type.

> Changing from Rusoto to the new SDK would be isolated to eks.rs.
> KubernetesNetworkConfigResponse is a ridiculously long type name and I hate seeing it everywhere :)

I agree! I went ahead and aliased KubernetesNetworkConfigResponse as `ClusterNetworkInfo` in eks.rs, and thus getting rid of the rusoto import from main.rs and the long type name.

etungsten:
Currently investigating an issue where `cri-container` isn't able to pick up any IPv6 addresses assigned by the vpc cni plugin for pods. There seems to be a disconnect somewhere. The changes here are still valid and necessary. We might need to follow up with a patch to cri-containerd to fix this issue.

etungsten:
Push above changes the approach to determining EKS cluster IP family. Instead of making a call to EKS API to determine the cluster's IP family, we check `cluster-dns-ip`'s IP family and if it's an IPv4 we assign the instance's IPv4 IP as `node-ip`; If it's an IPv6 address, we assign the instance's IPv6 address to be the `node-ip`.

When users want to give their nodes an IPv6 node IP address for an IPv6 EKS Cluster, they need to set the cluster's DNS IP to  `settings.kubernetes.cluster-dns-ip` via  instance userdata.

This is a temporary measure until the EKS API provides ways to let us automatically determine the IP family and service IPv6 CIDR.

etungsten:
Push above fixes README documentation for `settings.kubernetes.node-ip` and `settings.kubernetes.cluster-dns-ip`.

etungsten:
Push above adds the patch to kubelet for supporting ipv6 node addresses in AWS.
Tested everything and they work as expected.

README.md
```diff
@@ -368,17 +368,18 @@ Static pods can be particularly useful when running in standalone mode.
...
     * `cpu`: in millicores from the total number of vCPUs available on the instance.
     * `memory`: in mebibytes from the max num of pods on the instance. `memory_to_reserve = max_num_pods * 11 + 255`.
     * `ephemeral-storage`: defaults to `1Gi`.
 
-For Kubernetes variants in AWS, the following settings are set for you automatically by [pluto](sources/api/).
+For Kubernetes variants in AWS, the following settings are set for you automatically by [pluto](sources/api/) (small caveat for `cluster-dns-ip`, see note below).
 * `settings.kubernetes.max-pods`: The maximum number of pods that can be scheduled on this node (limited by number of available IPv4 addresses)
-* `settings.kubernetes.cluster-dns-ip`: Derived from the EKS IPV4 Service CIDR or the CIDR block of the primary network interface.
+* `settings.kubernetes.cluster-dns-ip`: Derived from the EKS Service IP CIDR or the CIDR block of the primary network interface.
+  * **Note** For IPv6 clusters, you would have to temporarily manually set this to the cluster's DNS IP's IPv6 address.
```
bcressey: Even for settings where we can determine the value automatically, it'd be better for scalability and instance launch time to have it specified through settings.

I'd like to see the automated path supported too, but we should avoid writing this in a way that suggests that manual configuration is undesirable.

packages/release/release-sysctl.conf
```diff
@@ -28,6 +28,10 @@ net.ipv4.ip_default_ttl = 255
 # Enable IPv4 forwarding for container networking.
 net.ipv4.conf.all.forwarding = 1
 
+# Enable IPv6 forwarding for container networking.
+net.ipv6.conf.all.forwarding = 1
+net.ipv6.conf.eth0.forwarding = 0
```
bcressey: Most of the sources I can find say that only the global (`all`) setting has any effect for IPv6 forwarding.

I'd rather omit the `eth0` setting if it's not doing anything.

sources/imdsclient/src/lib.rs
```diff
@@ -103,6 +103,27 @@ impl ImdsClient {
         self.fetch_string(&node_ip_target).await
     }
 
+    /// Gets the IPV6 addresses from instance metadata.
+    pub async fn fetch_ipv6_addresses(&mut self) -> Result<Option<Vec<String>>> {
+        // Get the mac address for the primary network interface
```
bcressey: The interface here is a bit awkward since we assume only the first (primary) interface is interesting, even though the name implies that all IPv6s will be returned to the caller.

It might be better to name it `fetch_primary_ipv6_address` and then do both of the `first` filtering operations. That way we're also centralizing the logic of figuring out which IPv6 address is the "main" one, which doesn't seem all that well documented and therefore might not be implemented the same way by every caller.

sources/api/pluto/src/main.rs
```diff
@@ -269,9 +309,23 @@ async fn run() -> Result<()> {
     let setting_name = parse_args(env::args());
     let mut client = ImdsClient::new().await.context(error::ImdsClient)?;
 
+    // Retrieve the kubernetes network configuration for the EKS cluster
+    let aws_k8s_info = api::get_aws_k8s_info().await.context(error::AwsInfo)?;
+    let cluster_network_config =
+        eks::get_cluster_network_config(&aws_k8s_info.region, &aws_k8s_info.cluster_name)
+            .await
+            .context(error::EksError)?;
```
bcressey: This results in us always making a `DescribeCluster` request for each `pluto` invocation, even if we don't need the response to determine `cluster-dns-ip` or `node-ip`.

There are a few drawbacks to that: it's slow and fallible; it creates unnecessary load on EKS; and the node might not have the role needed to describe the cluster.

It seems like we should first calculate `cluster-dns-ip`, if we don't have it in settings, and then either we're done (if we were asked to generate it) or we can pass the final value into `get_node_ip` (which would also simplify its interface).

Ideally we'd have a way to ensure that we generate `cluster-dns-ip` first before we attempt to generate `node-ip`, so we'd only need to describe the cluster once - not sure that's feasible with the tools we have right now, and it's not a blocker.

bcressey (changes requested):

packages/release/release-sysctl.conf
etungsten: The `eth0` setting is necessary otherwise `aws-k8s-cni` starts failing readiness probes
```
  Warning  Unhealthy  6m15s                  kubelet            Readiness probe failed: {"level":"info","ts":"2021-09-21T18:30:54.192Z","caller":"/usr/local/go/src/runtime/proc.go:203","msg":"timeout: failed to connect service \":50051\" within 1s"}
  Normal   Killing    5m55s                  kubelet            Container aws-node failed liveness probe, will be restarted
```

From asking around, it seems like the reason why we do this is because if forwarding is enabled then by default router advertisement (RA) packets are not accepted on `eth0`. This makes it so that there will be no routes to the default gateway and `aws-node` will not be able to reach the K8s apiserver. If we want to keep forwarding enabled, then we need to explicitly set `accept_ra` to `2`  so RA packets are accepted even if forwarding is enabled. (see http://strugglers.net/~andy/blog/2011/09/04/linux-ipv6-router-advertisements-and-forwarding/)

etungsten: We can do either of the following:
Just disable IPv6 forwarding on `eth0`:
```
# Enable IPv6 forwarding for container networking.
net.ipv6.conf.all.forwarding = 1
net.ipv6.conf.eth0.forwarding = 0
```

Accept RA packets on eth0 even if IPv6 forwarding is enabled on `eth0` :
```
# Enable IPv6 forwarding for container networking.
net.ipv6.conf.all.forwarding = 1
```

etungsten:
Pushes above addresses @bcressey 's comments. Tested the changes and they still work as expected.

packages/release/release-sysctl.conf
bcressey: Nice! I definitely prefer this approach; it's easier to reason about the interaction between two different settings.

sources/api/pluto/README.md
```diff
@@ -15,7 +15,7 @@ It uses IMDS to get information such as:
 
 It uses EKS to get information such as:
 
-- Service IPV4 CIDR
+- Service IPv4 CIDR
```
bcressey: Did we lose the updated version of `README.md`? It doesn't match `main.rs` comments now.

sources/imdsclient/src/lib.rs
```diff
@@ -103,6 +103,27 @@ impl ImdsClient {
...
+            .clone();
+
+        // Get the IPv6 addresses associated with the primary network interface
+        let ipv6_address_target = format!("meta-data/network/interfaces/macs/{}/ipv6s", mac);
+
+        let ipv6_addresses: Option<Vec<String>> = self
+            .fetch_string(&ipv6_address_target)
+            .await?
+            .map(|ipv6_addresses| ipv6_addresses.lines().map(|s| s.to_string()).collect());
+        Ok(ipv6_addresses.map_or(None, |ips| ips.first().cloned()))
```
bcressey: It'd be better to take only the first item from the `lines` iterator and convert that to a `String`, rather than converting everything and then throwing the rest away.

sources/api/pluto/src/main.rs
```diff
@@ -51,15 +55,32 @@ const DEFAULT_10_RANGE_DNS_CLUSTER_IP: &str = "172.20.0.10";
...
     #[derive(Debug, Snafu)]
     #[snafu(visibility = "pub(super)")]
     pub(super) enum PlutoError {
+        #[snafu(display(
+            "Unable to retrieve cluster name and AWS region from Bottlerocket API: {}",
+            source
+        ))]
+        AwsInfo { source: api::Error },
+
+        #[snafu(display("Missing field '{}' in EKS kubernetes network config response", field))]
```
bcressey: nit:
```suggestion
        #[snafu(display("Missing field '{}' in EKS network config response", field))]
```

etungsten:
Push above addresses comments by @bcressey . Tested and things still work as expected.

sources/api/pluto/src/main.rs
```diff
@@ -144,42 +163,31 @@ async fn get_max_pods(client: &mut ImdsClient) -> Result<String> {
...
+    let aws_k8s_info = api::get_aws_k8s_info().await.context(error::AwsInfo)?;
+    let cluster_network_config =
+        eks::get_cluster_network_config(&aws_k8s_info.region, &aws_k8s_info.cluster_name)
+            .await
+            .context(error::EksError)?;
+    // Derive cluster-dns-ip from the service IPv4 CIDR
+    if let Some(ipv4_cidr) = cluster_network_config.service_ipv_4_cidr {
+        if let Ok(dns_ip) = get_dns_from_ipv4_cidr(&ipv4_cidr) {
+            return Ok(dns_ip);
         }
```
webern: The suggestion may not render well, but it looks like an error with EKS call would prevent fallback. I think you can use `if let Ok(cluster_network_config) {}` to ensure fallback.

(I could be wrong about this interpretation)

```suggestion
if let Ok(cluster_network_config) =
    eks::get_cluster_network_config(&aws_k8s_info.region, &aws_k8s_info.cluster_name)
        .await
        .context(error::EksError) {
    // Derive cluster-dns-ip from the service IPv4 CIDR
    if let Some(ipv4_cidr) = cluster_network_config.service_ipv_4_cidr {
        if let Ok(dns_ip) = get_dns_from_ipv4_cidr(&ipv4_cidr) {
            return Ok(dns_ip);
        }
    }
}
```

etungsten: Nice catch!

etungsten:
Push above addresses Matt's comment https://github.com/bottlerocket-os/bottlerocket/pull/1710#discussion_r714206965

sources/api/pluto/src/main.rs
```diff
@@ -144,42 +163,33 @@ async fn get_max_pods(client: &mut ImdsClient) -> Result<String> {
...
-    let aws_k8s_info = match api::get_aws_k8s_info().await {
-        Ok(value) => value,
-        Err(e) => {
-            eprintln!(
-                "Unable to get region and cluster name from Bottlerocket API, using default DNS IP: {}",
-                e
-            );
-            return None;
+    // Retrieve the kubernetes network configuration for the EKS cluster
+    let aws_k8s_info = api::get_aws_k8s_info().await.context(error::AwsInfo)?;
```
etungsten: Need to fix this too to allow fallthrough

etungsten:
Push above fixes the fallthrough for determining the cluster dns ip.

webern (approved):

etungsten:
Ran conformance tests with an IPv6 cluster and all tests passed.

bcressey (approved):

PR #1713, opened 2021-08-18
Add an entry to `/etc/hosts` for the current hostname

### LINKED ISSUE #1700: Populate `/etc/hosts` with an entry for `settings.network.hostname`
Currently [we populate](https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/release/hosts)`/etc/hosts` with entries for `127.0.0.1` and `::1` that alias to `localhost`.

We need to populate it with an entry to alias `127.0.0.1` to the hostname.  

One example:  For other distros (Ubuntu) CAPI populates this entry in `/etc/hosts` and `kube-vip` appears to use `/etc/hosts` to resolve the hostname.

zmrow:
**Issue number:**
Fixes #1700

**Description of changes:**
```
    schnauzer: Add a new template helper: `add_unresolvable_hostname`
    
    This change adds a new template helper `add_unresolvable_hostname` that
    will attempt to resolve the hostname given to it in the template (from
    `settings.network.hostname`).  If the hostname is unresolvable or
    resolves to localhost (127.0.0.1 or ::1), the helper adds an alias to
    the hostname for both the IPv4 and IPv6 addresses
```
```
    Add a hostname entry to `/etc/hosts` if hostname is unresolvable
    
    Previously, `/etc/hosts` was a static file (in the `release` package)
    that we copied into the image when building Bottlerocket.  In order to
    have an entry in the file corresponding to hostname, a new "hosts"
    service was added to the affected services for
    `settings.network.hostname`.  This new "hosts" service has a templated
    configuration file that uses `settings.network.hostname` to populate the
    entry in `/etc/hosts` if the current hostname is unresolvable in DNS.
    The templated configuration file uses the previously added
    `add_unresolvable_hostname` template renderer helper.
```
```
    migrations: Add migrations for `/etc/hosts` "service"
    
    This change includes the 2 migrations necessary to make `/etc/hosts` an
    "affected service" of `settings.network.hostname`: one to add the
    service and configuration file for the hosts service, and another to
    add the hosts service to `settings.network.hostname`.
```

**Testing done:**
Boot an `aws-k8s-1.19` host and observe proper behavior, run a pod, etc.  `/etc/hosts` looks like:
```
bash-5.0# cat /etc/hosts 
```
Upgrade the same host to a version including these changes, observe proper boot and behavior.  `/etc/hosts` looks the same because the hostname resolves in EC2:
```
bash-5.0# cat /etc/hosts 
127.0.0.1 localhost localhost.localdomain localhost4 localhost4.localdomain4l

```

I followed the same procedure with a VMware variant.  `/etc/hosts` before the upgrade:
```
bash-5.0# cat /etc/hosts 
```
`/etc/hosts` after the upgrade (`hosts-test` is the hostname of my VM):
```
bash-5.0# cat /etc/hosts 
127.0.0.1 hosts-test
::1 hosts-test
```

**Terms of contribution:**

### DIFF
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -6,7 +6,6 @@ Release: 0%{?dist}
 Summary: Bottlerocket release
 License: Apache-2.0 OR MIT
 
-Source10: hosts
 Source11: nsswitch.conf
 Source97: release-sysctl.conf
 Source98: release-systemd-system.conf
@@ -15,6 +14,7 @@ Source99: release-tmpfiles.conf
 Source200: motd.template
 Source201: proxy-env
 Source202: hostname-env
+Source203: hosts.template
 
 Source1000: eth0.xml
 Source1001: multi-user.target
@@ -91,7 +91,7 @@ Requires: %{_cross_os}wicked
 
 %install
 install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
-install -p -m 0644 %{S:10} %{S:11} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
+install -p -m 0644 %{S:11} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
 
 install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/wicked/ifconfig
 install -p -m 0644 %{S:1000} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/wicked/ifconfig
@@ -135,6 +135,7 @@ install -d %{buildroot}%{_cross_templatedir}
 install -p -m 0644 %{S:200} %{buildroot}%{_cross_templatedir}/motd
 install -p -m 0644 %{S:201} %{buildroot}%{_cross_templatedir}/proxy-env
 install -p -m 0644 %{S:202} %{buildroot}%{_cross_templatedir}/hostname-env
+install -p -m 0644 %{S:203} %{buildroot}%{_cross_templatedir}/hosts
 
 install -d %{buildroot}%{_cross_udevrulesdir}
 install -p -m 0644 %{S:1016} %{buildroot}%{_cross_udevrulesdir}/61-mount-cdrom.rules
@@ -142,7 +143,6 @@ install -p -m 0644 %{S:1016} %{buildroot}%{_cross_udevrulesdir}/61-mount-cdrom.r
 ln -s %{_cross_unitdir}/preconfigured.target %{buildroot}%{_cross_unitdir}/default.target
 
 %files
-%{_cross_factorydir}%{_cross_sysconfdir}/hosts
 %{_cross_factorydir}%{_cross_sysconfdir}/nsswitch.conf
 %{_cross_factorydir}%{_cross_sysconfdir}/wicked/ifconfig/eth0.xml
 %{_cross_sysctldir}/80-release.conf
@@ -175,6 +175,7 @@ ln -s %{_cross_unitdir}/preconfigured.target %{buildroot}%{_cross_unitdir}/defau
 %{_cross_templatedir}/motd
 %{_cross_templatedir}/proxy-env
 %{_cross_templatedir}/hostname-env
+%{_cross_templatedir}/hosts
 %{_cross_udevrulesdir}/61-mount-cdrom.rules
 
 %changelog
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -24,8 +24,9 @@ members = [
     "api/migration/migration-helpers",
     "api/shibaken",
 
-    # "api/migration/migrations/vX.Y.Z/...
-    # (all migrations currently archived; replace this line with new ones)
+    # "api/migration/migrations/vX.Y.Z/..."
+    "api/migration/migrations/v1.2.1/etc-hosts-service",
+    "api/migration/migrations/v1.2.1/hostname-affects-etc-hosts",
 
     "bottlerocket-release",
 
--- a/sources/api/schnauzer/src/helpers.rs
+++ b/sources/api/schnauzer/src/helpers.rs
@@ -2,6 +2,7 @@
 // be registered with the Handlebars library to assist in manipulating
 // text at render time.
 
+use dns_lookup::lookup_host;
 use handlebars::{Context, Handlebars, Helper, Output, RenderContext, RenderError};
 use lazy_static::lazy_static;
 use num_cpus;
@@ -10,6 +11,7 @@ use snafu::{OptionExt, ResultExt};
 use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::convert::TryFrom;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use url::Url;
 
 lazy_static! {
@@ -97,6 +99,9 @@ const KUBE_RESERVE_2_CORES: f32 = KUBE_RESERVE_1_CORE + 10.0;
 const KUBE_RESERVE_3_CORES: f32 = KUBE_RESERVE_2_CORES + 5.0;
 const KUBE_RESERVE_4_CORES: f32 = KUBE_RESERVE_3_CORES + 5.0;
 const KUBE_RESERVE_ADDITIONAL: f32 = 2.5;
+
+const IPV4_LOCALHOST: IpAddr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
+const IPV6_LOCALHOST: IpAddr = IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1));
 
 /// Potential errors during helper execution
 mod error {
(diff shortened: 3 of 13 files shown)


bcressey:
I'd prefer that we scope this more narrowly so that we only add the hosts entry if our configured hostname isn't resolvable through DNS.

That's really the only case where we know it's needed and helpful.

zmrow:
^ Addresses @bcressey 's concern and adds a new template renderer helper that will only add the entry to `/etc/hosts` if the hostname isn't resolvable or resolves to localhost.

zmrow:
^ Splits the template helper into its own commit

sources/api/schnauzer/src/helpers.rs
```diff
@@ -935,8 +1022,9 @@ fn pause_registry<S: AsRef<str>>(region: S) -> String {
 }
 
 /// Calculates and returns the amount of CPU to reserve
-fn kube_cpu_helper(num_cores: usize) -> Result<String, TemplateHelperError>{
-    let num_cores = u16::try_from(num_cores).context(error::ConvertUsizeToU16 { number: num_cores })?;
+fn kube_cpu_helper(num_cores: usize) -> Result<String, TemplateHelperError> {
```
zmrow, line 1014: This was a `rustfmt` change

sources/api/schnauzer/src/helpers.rs
```diff
@@ -949,7 +1037,11 @@ fn kube_cpu_helper(num_cores: usize) -> Result<String, TemplateHelperError>{
             KUBE_RESERVE_4_CORES + ((num_cores - 4.0) * KUBE_RESERVE_ADDITIONAL)
         }
     };
-    Ok(format!("{}{}", cpu_to_reserve.floor().to_string(), millicores_unit))
+    Ok(format!(
```
zmrow, line 1029: Same here - thanks `rustfmt`

zmrow:
^ rebased on `develop`

sources/Cargo.toml
```diff
@@ -24,8 +24,8 @@ members = [
     "api/migration/migration-helpers",
     "api/shibaken",
 
-    # "api/migration/migrations/vX.Y.Z/...
-    # (all migrations currently archived; replace this line with new ones)
```
webern, line 28: Maybe leave some comment here identifying this as the migration section?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1607,3 +1699,50 @@ mod test_kube_cpu_helper {
...
+        );
+
+        registry.render_template(tmpl, data)
+    }
+
+    #[test]
+    fn resolves_to_localhost_renders_entries() {
+        let result = setup_and_render_template(
+            "{{add_unresolvable_hostname name}}",
+            &json!({"name": "localhost"}),
```
webern, line 1718: I guess this test should always work on every system?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1607,3 +1699,50 @@ mod test_kube_cpu_helper {
...
+            "127.0.0.1 localhost
+::1 localhost"
+        )
+    }
+
+    #[test]
+    fn resolvable_hostname_renders_nothing() {
+        let result = setup_and_render_template(
+            "{{add_unresolvable_hostname name}}",
+            &json!({"name": "amazon.com"}),
```
webern, line 1732: Why is this unresolvable by DNS? Will it always be unresolvable? Would something completely invalid be better? `not*a*valid*hostname`?

sources/api/schnauzer/src/lib.rs
```diff
@@ -131,6 +131,10 @@ pub fn build_template_registry() -> Result<handlebars::Handlebars<'static>> {
         "kube_reserve_memory",
         Box::new(helpers::kube_reserve_memory),
     );
+    template_registry.register_helper(
+        "add_unresolvable_hostname",
```
webern, line 136: Ugh, we now have several each of kabob case and snake case. Maybe we should add a comment that specifies what the preference is and that we messed it up.

webern (approved):
Don't we need to add something to README for the new settings. Possibly a warning in k8s docs about kubelet not liking a hostname change?

sources/api/schnauzer/src/helpers.rs
zmrow: The helper only renders if the name is _unresolvable_ or resolves to `127.0.0.1 or ::1`.  I realize that `amazon.com` isn't a realistic hostname, but it should resolve mostly anywhere and that's the behavior I was after.

sources/api/schnauzer/src/helpers.rs
zmrow: I believe it should?  I'm willing to be told otherwise though.

sources/api/schnauzer/src/lib.rs
zmrow: :( I didn't even see the kabob cased ones.

I can add a comment requesting snake case since that's what we started with and the majority of them are snake case.

sources/Cargo.toml
zmrow: I'll add back the `# "api/migration/migrations/vX.Y.Z/...` comment since that's what was previously there.

zmrow:
^ Added the comments @webern asked for

jpculp (approved):

sources/api/schnauzer/src/lib.rs
```diff
@@ -118,6 +118,7 @@ pub fn build_template_registry() -> Result<handlebars::Handlebars<'static>> {
     // but isn't provided in the data given to the renderer
     template_registry.set_strict_mode(true);
 
+    // Prefer snake case for heper names (we accidentally created a few with kabob case)
```
webern:
```suggestion
    // Prefer snake case for helper names (we accidentally created a few with kabob case)
```

webern (approved):

sources/api/schnauzer/src/lib.rs
zmrow: Oops - good catch.  Fixed.

zmrow:
^ Fixed my misspelling. :)

packages/release/release.spec
```diff
@@ -135,14 +135,14 @@ install -d %{buildroot}%{_cross_templatedir}
 install -p -m 0644 %{S:200} %{buildroot}%{_cross_templatedir}/motd
 install -p -m 0644 %{S:201} %{buildroot}%{_cross_templatedir}/proxy-env
 install -p -m 0644 %{S:202} %{buildroot}%{_cross_templatedir}/hostname-env
+install -p -m 0644 %{S:10} %{buildroot}%{_cross_templatedir}/hosts
```
bcressey: nit: this should be in the 200 range like other templates

sources/api/schnauzer/src/helpers.rs
```diff
@@ -98,6 +101,9 @@ const KUBE_RESERVE_3_CORES: f32 = KUBE_RESERVE_2_CORES + 5.0;
 const KUBE_RESERVE_4_CORES: f32 = KUBE_RESERVE_3_CORES + 5.0;
 const KUBE_RESERVE_ADDITIONAL: f32 = 2.5;
 
+const IPV4_LOCALHOST: &str = "127.0.0.1";
+const IPV6_LOCALHOST: &str = "::1";
```
bcressey: You can make these `IpAddr` constants to avoid doing the conversion at runtime:
```suggestion
const IPV4_LOCALHOST: IpAddr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
const IPV6_LOCALHOST: IpAddr = IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1));
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -869,6 +881,81 @@ pub fn kube_reserve_cpu(
...
+            trace!("DNS hostname lookup failed: {},", e);
+            false
+        }
+    };
+
+    // Only write an entry to the template if the hostname is unresolvable
+    if !hostname_resolveable {
+        let ipv4_entry = format!("{} {}", IPV4_LOCALHOST, hostname_str);
+        let ipv6_entry = format!("{} {}", IPV6_LOCALHOST, hostname_str);
+        let entries = format!("{}\n{}", ipv4_entry, ipv6_entry);
```
bcressey, line 939: It might be better to ensure that a newline is written at the end of the file, rather than relying on handlebars - unless this results in an extra newline.

```suggestion
        let entries = format!("{}\n{}\n", ipv4_entry, ipv6_entry);
```

zmrow: I'd prefer not to add an extra newline since handlebars seems to add one:
```
bash-5.0# cat -E /etc/hosts 
127.0.0.1 localhost localhost.localdomain localhost4 localhost4.localdomain4$
::1 localhost localhost.localdomain localhost6 localhost6.localdomain6$
127.0.0.1 mrow-9-1$
::1 mrow-9-1$
```

zmrow: Adding the extra newline in the template helper adds an additional line in the rendered template:
```
bash-5.0# cat -E /etc/hosts 
127.0.0.1 localhost localhost.localdomain localhost4 localhost4.localdomain4$
::1 localhost localhost.localdomain localhost6 localhost6.localdomain6$
127.0.0.1 mrow-9-1-newline$
::1 mrow-9-1-newline$
$
```

zmrow:
^ Addressed @bcressey 's comments

sources/api/schnauzer/src/helpers.rs
```diff
@@ -149,6 +154,12 @@ mod error {
             template: String,
         },
 
+        #[snafu(display("Invalid IP address '{}': {}", ip, source))]
+        IpFromString {
+            ip: String,
+            source: std::net::AddrParseError,
+        },
+
```
bcressey: I don't think this error is used any more.

zmrow:
^ removed the unused error variant

bcressey (approved):

Issue #1674, opened 2021-07-28
Update to runc 1.0+


tjkirch:
The 1.0.0 and 1.0.1 release notes indicate changes that we should carefully review and test.

bcressey:
Looks like we'll want 1.0.2 if possible for the cgroup v1 fix: https://github.com/opencontainers/runc/issues/3149.

PR #1716, opened 2021-08-19
update containerd to 1.5.5, runc to 1.0.1

### LINKED ISSUE #1670: Update to containerd 1.5.x series

bcressey:
**Issue number:**
Fixes #1670 
#1674

**Description of changes:**
Update containerd to 1.5.5. The log reduction patch is no longer needed, and the volume relabeling patch is replaced by what's likely to be the upstream fix.

Update runc to 1.0.1. There's a 1.0.2 release that's coming soon with some fixes we want, but this is still a useful upgrade over `1.0.0-rc95`.

**Testing done:**
Verified that a `sonobuoy` run passes for a 1.21 cluster. Confirmed that tasks launch for ECS. No unusual new errors from runc, containerd, or kubelet in the logs.

Verified that pods still had the larger number of file descriptors.

Verified that SELinux labeling was still correct.

**Terms of contribution:**


packages/runc/Cargo.toml
```diff
@@ -9,8 +9,9 @@ build = "build.rs"
 path = "pkg.rs"
 
 [[package.metadata.build-package.external-files]]
-url = "https://github.com/opencontainers/runc/archive/b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7/runc-b9ee9c6314599f1b4a7f497e1f1f856fe433d3b7.tar.gz"
-sha512 = "ebc67bbc825aa6d7a51e9504d3b4f41794861c06a18288395c035c1b6929e9fb1bae0dcf1b14d9120b651f4a63fdcd6ce8e575a2e84bf819d0d0d1d1cb4e2bae"
+url = "https://github.com/opencontainers/runc/releases/download/v1.0.1/runc.tar.xz"
+path = "runc-v1.0.1.tar.xz"
```
tjkirch, line 13: Can we use the same `#/desired-name.ext` syntax as in the `Source` line of the spec file, rather than needing a `path` override here?

tjkirch (approved):

packages/runc/Cargo.toml
bcressey: We'd have to update [extract_file_name](https://github.com/bottlerocket-os/bottlerocket/blob/b9c8747e7abf4f5dd02e217571ed8d26f3f3db72/tools/buildsys/src/cache.rs#L101) to prefer a fragment if present.

I'm not sure how I feel about that. It would let us remove the "path" setting and align `Cargo.toml` with the spec file URI, which seems good. But it's also more cryptic.

jpculp (approved):

PR #1723, opened 2021-08-24
infrasys: create TUF infra in AWS using `cargo make create-infra`

aashnasheth:
**Issue number:**
N/A 

**Description of changes:**

- Infrasys is currently comprised of 2 subcommands: `create-infra` and `check-infra-lock`. This PR is the implementation of `create-infra`.
- The main logic for creating the TUF infrastructure is in `main.rs`, which uses helper methods from other modules (`keys.rs` [creating keys], `s3.rs` [creating s3 bucket, adding bucket policy, uploading objects], `shared.rs` [helper methods used in all files], `errors.rs` [a single error file shared by all files], and `root.rs` [creating root, adding keys to it, signing it]).
- `main.rs` reads from Infra.toml using pubsys-config (thus edits to a few files under pubsys-config and pubsys were required) and outputs Infra.lock in yaml (we used yaml to serialize the config because we were unable to serialize enums with structs within them in toml). 
- NOTE: a warning is generated because create-infra-lock has some unused variables (because it's unimplemented)

**Testing done:**
- Added a unit test for for struct -> toml -> struct -> yaml -> struct conversion (main.rs), prefix format checking (s3.rs), and bucket policy insertion (s3.rs)
- Created several test Infra.toml files with various combinations of configurations and ensured that errors were thrown / the resources showed up in the correct region, with the correct settings, in my account
   - Some cases are: available keys only, available keys + keys to be created, create keys only, s3 configs missing, alias missing, signing keys/root keys missing, empty prefix, etc. (and more in-progress!)
- Example of a test where we're creating a root + signing keys 
    <details>
    <summary>Infra.toml</summary>
    <p>
    
    ```toml
    # Check for:
    # key_id is new key in signing keys
    # available_keys is updated
    # s3 info set
    
    [repo.default] 
       file_hosting_config_name = "TUF-Repo-S3-Buck"
       signing_keys = { kms = { key_alias="signing-key-test", regions=["us-east-2"] } }
       root_keys = { kms = { key_alias="root-key-test", regions=["us-east-1"] } }
       root_key_threshold = 1
       pub_key_threshold = 1
       
    [aws] 
        [aws.s3.TUF-Repo-S3-Buck]
        region = "us-west-2"
        vpc_endpoint_id = "vpc-12345" 
        s3_prefix = "/my-bottlerocket-remix" 
    ```
    
    </p>
    </details>
    
     <details>
    <summary>cargo make create-infra output</summary>
    <p>
    
    ```bash
    [cargo-make] INFO - Running Task: check-infra-lock
    Successfully in check_infra_method!
    [cargo-make] INFO - Running Task: create-infra
    20:06:00 [INFO] Parsing Infra.toml...
    20:06:00 [INFO] Creating S3 bucket...
    20:06:01 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:06:21 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:06:42 [INFO] Creating KMS Keys...
    20:06:43 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:07:04 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:07:24 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:07:45 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:08:05 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:08:25 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:08:46 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:09:06 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:09:26 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:09:47 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:10:08 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:10:29 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:10:49 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:11:10 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:11:30 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:11:50 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:12:11 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:12:31 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:12:52 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:13:12 [INFO] Waiting for stack resources to be ready, current status is 'CREATE_IN_PROGRESS'...
    20:13:33 [INFO] Creating and signing root.json...
    f5b8391e577e7abd0354963177ecfc2b2aff6c6aa62b59df72c37a57020acb83
    fe6321b3c61d4f0cc7864e283e4dd007056d8e9ae1dd367166720225efa16681
    20:13:36 [INFO] Uploading root.json to S3 bucket...
    20:13:37 [INFO] Writing Infra.lock...
    20:13:37 [INFO] Complete!
    [cargo-make] INFO - Build Done in 460.56 seconds.
    ```
    
    </p>
    </details>

    <details>
    <summary>Infra.lock</summary>
    <p>
    
    ```yaml
    ---
    repo:
      default:
        root_role_url: "http://tuf-repo-s3-buck-tufrepobucket-1xrapc0rkw7yx.s3-website-us-west-2.amazonaws.com/my-bottlerocket-remix/root.json"
        root_role_sha512: 2b2e33ff34c91c81ee06c0ad42a75d95dd5cb38e0b8eb04015147e6ce709eb086b5e3e03bf77249dab589e4e911cc89270a702c6e4a57f8b864717242ed41de8
        signing_keys:
          kms:
            key_id: "arn:aws:kms:us-east-2:965384286431:key/6928e220-d424-4818-aa1f-f4175e078269"
            available_keys:
              "arn:aws:kms:us-east-2:965384286431:key/6928e220-d424-4818-aa1f-f4175e078269": us-east-2
            key_alias: signing-key-test
            regions:
              - us-east-2
            key_stack_arns:
              "arn:aws:kms:us-east-2:965384286431:key/6928e220-d424-4818-aa1f-f4175e078269": "arn:aws:cloudformation:us-east-2:965384286431:stack/TUF-KMS-signing-key-test/cd84bad0-0516-11ec-9916-0a27ba133414"
        root_keys:
          kms:
            key_id: ~
            available_keys:
              "arn:aws:kms:us-east-1:965384286431:key/184567d9-cffd-4885-acff-0fd2754c504a": us-east-1
            key_alias: root-key-test
            regions:
              - us-east-1
            key_stack_arns:
              "arn:aws:kms:us-east-1:965384286431:key/184567d9-cffd-4885-acff-0fd2754c504a": "arn:aws:cloudformation:us-east-1:965384286431:stack/TUF-KMS-root-key-test/47a07ed0-0517-11ec-bc80-0af9e889672b"
        metadata_base_url: "http://tuf-repo-s3-buck-tufrepobucket-1xrapc0rkw7yx.s3-website-us-west-2.amazonaws.com/my-bottlerocket-remix/metadata/"
        targets_url: "http://tuf-repo-s3-buck-tufrepobucket-1xrapc0rkw7yx.s3-website-us-west-2.amazonaws.com/my-bottlerocket-remix/targets/"
        file_hosting_config_name: TUF-Repo-S3-Buck
        root_key_threshold: 1
        pub_key_threshold: 1
    aws:
      regions: []
      role: ~
      profile: ~
      region: {}
      ssm_prefix: ~
      s3:
        TUF-Repo-S3-Buck:
          region: us-west-2
          s3_prefix: /my-bottlerocket-remix
          vpc_endpoint_id: vpc-12345
          stack_arn: "arn:aws:cloudformation:us-west-2:965384286431:stack/TUF-Repo-S3-Buck/b450b410-0516-11ec-a7c0-061175597a0d"
          bucket_name: tuf-repo-s3-buck-tufrepobucket-1xrapc0rkw7yx
    vmware: ~ 
    ```
    
    </p>
    </details>

    <details>
    <summary>root.json</summary>
    <p>
    
    ```json
   {
    "signed": {
      "_type": "root",
      "spec_version": "1.0.0",
      "consistent_snapshot": true,
      "version": 1,
      "expires": "2022-08-23T20:13:33Z",
      "keys": {
        "fe6321b3c61d4f0cc7864e283e4dd007056d8e9ae1dd367166720225efa16681": {
          "keytype": "rsa",
          "keyval": {
            "public": "-----BEGIN PUBLIC KEY-----\nMIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEAw+Kz32mPNDLtM4Qn+G3X\n+hf/l9h3KCFOGtBpeNxQN9NLtExA3e1QLGS6yQQNVa7ELPIvWep9hcoraxP2kv5X\nLhMdzy0cmFVFWwtlt3/Srd9paihW+cP53ODnr2/KeXUlLAco13s26dMzvxOZmwAg\nxhDzj+846n592WRWGcuGO80kWSFgfipufwld32jbUZoC9OqwvIBVn1Qk3BHm4GHI\ntG/VG65CRdHXtZ2FSC4KgeVjJons3mqGjcgXwN2dVuWmmCJgrTKvUru706KctxF3\nnvFXTMmg0MosM4woArXCHUb82lugFZeLic+egSLupivNceblubVMU2xxF3SE98hv\nQxmaWBI2sQ+4uBFvh1KkKVwAHeSfbrrU+oNYsilVXcHc+hXiyhneMxkOkt21l0iP\nP8XnOyzGLz6dEtc7aMqGlIDGPdVkQgLl/MSnRPpb641uFGENKUo8MWGCXz33G7wB\n1wtwZXJFVMdTv2juhG0Umft8lp/USuxfO7eQOe+cCmpXAgMBAAE=\n-----END PUBLIC KEY-----\n"
          },
          "scheme": "rsassa-pss-sha256"
        },
        "f5b8391e577e7abd0354963177ecfc2b2aff6c6aa62b59df72c37a57020acb83": {
          "keytype": "rsa",
          "keyval": {
            "public": "-----BEGIN PUBLIC KEY-----\nMIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEAnkREp096HUOSmwgeM/K4\n4Zzky0h3rmSBXdJ+wn4q/fFjKYKkVKwxu3H2aa/GueJAb1q/SqVBHism5JNZl46E\ngTczchClMvPDQsshKH9IQ0t4bl8quaH04+zpnzWoYMQHptIv56yTTirVsgg7NBkx\noi8pXXC+fZdH1ed+qKUmr5hwz1wlZrWVwY3SCYLv7QYi/4sL8mRw8tzx7IDd4+5d\nZBXC2Qb9loQ6zjFIoxSubz5Fl1zuGk+cAG/TdOzangr0poGFwu9Xm0mh2RByuAbX\nJ61SZWRaAb9AwkTy7C2Du6hZ0ZwSfYoKcGWfnylVYLrreAxOPi3X6odNKgWt12Yq\na95+lv0n/830dg36zZ9I0eVMQZgIupM4zAUlv3QvBA83QaQnrQV5JSCYawcftah6\nbQ1BaZlAnJi9rGwZ7sGj7deJtz+BqXwSOPA3cdm9q2rinxXlLyllLlQsWcvbvVlx\nBKgi2K5gI9ZxM0T9qBvo7jy0m1GDhXve9+BrdbXmJ351AgMBAAE=\n-----END PUBLIC KEY-----\n"
          },
          "scheme": "rsassa-pss-sha256"
        }
      },
      "roles": {
        "root": {
          "keyids": [
            "fe6321b3c61d4f0cc7864e283e4dd007056d8e9ae1dd367166720225efa16681"
          ],
          "threshold": 1
        },
        "snapshot": {
          "keyids": [
            "f5b8391e577e7abd0354963177ecfc2b2aff6c6aa62b59df72c37a57020acb83"
          ],
          "threshold": 1
        },
        "targets": {
          "keyids": [
            "f5b8391e577e7abd0354963177ecfc2b2aff6c6aa62b59df72c37a57020acb83"
          ],
          "threshold": 1
        },
        "timestamp": {
          "keyids": [
            "f5b8391e577e7abd0354963177ecfc2b2aff6c6aa62b59df72c37a57020acb83"
          ],
          "threshold": 1
        }
      }
    },
    "signatures": [
      {
        "keyid": "fe6321b3c61d4f0cc7864e283e4dd007056d8e9ae1dd367166720225efa16681",
        "sig": "813107cdf34a3181bcd6eb0f83cdce350670e26bf381f535a3e09e33d638d50c18a203f91f1c1ad2966efd689579d5c433db6093811f13d6a3a1d964ebd6839e8edf67c9b968591cd2b5a845cdb360e634cfb72328773010e0646fc1a75738cbddf008f5981e713f7364aec2b987f3125b5764317e4786843307973d6635daa4c7ea2b2a83db77663ccf6288d82f8695bd648376b2105a891d64da91af5a3f63544d4369c1a8da23f36d1039eabb2ea13abc7f6549d281c00a96db6daf32eed51180c966b0ab4251d47a1dffd57cc504b69de0c482d92d909040078e20d7ad59a27f9627aab780e3bcf3e2f5e150d5d72107d8b1fffbc7081b602588dc2608313af97d4b8d9207222c86fe9b626dd76d65ebd2c4278afe0272bb5fd84d9047356463f1463c62a813fe72641f857b970c131db9ddddc14c6d0b67aac4735a5da763ccf5450b6455f55fd4c2b6823ec19c223987a37feeac9047b2b5cd8f38c467fb6c69ed10ad78b7a02d5cca4ab4774ac58ca61bf8c2b4c5706858ef8d45f0e3"
      }
    ]
  }

    ```
    
    </p>
    </details>

     Resources show up correctly in my account (root.json populated in s3 bucket, all public access off for s3 bucket, correct bucket policy, kms keys present)
- Completed E2E testing (cargo make, cargo make repo, cargo make ami, pulling updates)

**Draft Review:**
Here are some initial reviews that etungsten left and were addressed: https://github.com/aashnasheth/bottlerocket/pull/9 

**Next PR:**
- Implementation of `check-infra-lock` (which will check the existence of the lock file and advise users to clean up resources if needed)
- S3 sync in MakeFile
- A way to clean up any created stacks if infrasys fails unexpectedly in the middle of execution 

**Terms of contribution:**


tools/infrasys/src/keys.rs
```diff
@@ -0,0 +1,130 @@
...
+    check_signing_key_config(signing_key_config)?;
+    match signing_key_config {
+        SigningKeyConfig::file { .. } => (),
+        SigningKeyConfig::kms { config, .. } => {
+            config
+                .as_mut()
+                .context(error::MissingConfig {
+                    missing: "config field for a kms key",
+                })?
+                .create_kms_keys()
```
aashnasheth, line 24: Decided to make create_kms_keys a trait of the config so users aren't surprised when config fields are changed. We also considered passing a mutable reference to config to create_kms_keys(), but thought this way was more clear.

tools/infrasys/src/root.rs
```diff
@@ -0,0 +1,196 @@
...
+            for (keyid, region) in available_keys.iter() {
+                tuftool!(
+                region,
+                "root add-key '{}' aws-kms:///'{}' --role snapshot --role targets --role timestamp",
+                filepath,
+                keyid
+                );
+            }
+
+            // Set key_id using a publication key (if one is not already provided)
```
aashnasheth, line 157: We decided to set key_id in this method as it's the only one that differentiates roles (we only want key_id to be set for publication keys, not root keys)

aashnasheth:
Pushes above add some extra tests and make minor changes to existing ones:

- Adds bucket policy insertion test and prefix test
- Adds support for empty prefixes

tools/infrasys/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+structopt = { version = "0.3", default-features = false }  
+clap = "2.33"  
+url = "2.2.2"  
+async-trait = "0.1.51"  
+shell-words = "1.0.0" 
+sha2 = "0.9"
+serde_yaml = "0.8.17"
+simplelog = "0.10.0"
+serde_json = "1.0.66"
+assert-json-diff = "2.0.1"
```
aashnasheth, line 30: Note from etungsten: add [dev-dependencies]

aashnasheth:
Push above addresses warnings raised by `cargo clippy`

tools/infrasys/src/error.rs
```diff
@@ -0,0 +1,152 @@
...
+        what,
+        resource_name
+    ))]
+    ParseResponse { what: String, resource_name: String },
+
+    #[snafu(display("Failed to get bucket policy for bucket {}: {}", bucket_name, source))]
+    GetPolicy {
+        bucket_name: String,
+        source: rusoto_core::RusotoError<rusoto_s3::GetBucketPolicyError>,
+    },
```
etungsten: This seems unused. Can we remove this?

tools/infrasys/src/keys.rs
```diff
@@ -0,0 +1,130 @@
...
+        SigningKeyConfig::ssm { .. } => (),
+    }
+    Ok(())
+}
+
+pub fn check_signing_key_config(signing_key_config: &SigningKeyConfig) -> Result<()> {
+    match signing_key_config {
+        SigningKeyConfig::file { .. } => (),
+        SigningKeyConfig::kms { config, .. } => {
+            let unwrapped_config = config.as_ref().context(error::MissingConfig {
```
etungsten: We can just rebind the variable to be `config`.
```suggestion
            let config = config.as_ref().context(error::MissingConfig {
```

tools/infrasys/src/keys.rs
```diff
@@ -0,0 +1,130 @@
...
+/// Output: Populates KMSKeyConfig with information about resources created
+/// available-keys starts as a map of pre-existing keyids:regions and will end as a map of pre-existing and generated keyids:regions
+/// key-stack-arns starts empty and will end as a map of keyids:stackarn if new keys are created
+#[async_trait]
+impl KMSKeyConfigExt for KMSKeyConfig {
+    async fn create_kms_keys(&mut self) -> Result<()> {
+        // Generating new keys (if regions is non-empty)
+        for region in self.regions.iter() {
+            // Unwrapping key_alias here, because if regions is defined, key_alias must be
+            let stack_name = format!("TUF-KMS-{}", self.key_alias.as_ref().unwrap());
```
etungsten: I know technically we would never panic due to the `unwrap` here since we're checking earlier, but we should still avoid unwrapping since we can just easily substitute this with an `error::KeyConfig`.
```suggestion
            let key_alias =  self.key_alias.as_ref().context(error::KeyConfig { ... })?;
            let stack_name = format!("TUF-KMS-{}", key_alias);
```

Also there's an unwrap a few lines below as well.

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,261 @@
...
+#[derive(Debug, StructOpt)]
+#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+struct CreateInfraArgs {
+    #[structopt(long)]
+    root_role_path: PathBuf,
+}
+
+#[derive(Debug, StructOpt)]
+enum SubCommand {
+    CheckInfraLock,
```
etungsten: I think we should probably leave this sub-command out of this commit if we're making this commit specifically for `CreateInfra`. If we decide we want to make this PR cover both functionality then we can just add this in a second commit.

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,261 @@
...
+        // Step 1: Create S3 Bucket
+        info!("Creating S3 bucket...");
+        let (s3_stack_arn, bucket_name, bucket_url) =
+            s3::create_s3_bucket(s3_region, s3_stack_name).await?;
+        // Set output variables
+        s3_info.stack_arn = Some(s3_stack_arn);
+        s3_info.bucket_name = Some(bucket_name.clone());
+
+        // Step 2: Add Bucket Policy to newly created bucket
+        // TODO: test adding bucket policy without the /*
```
etungsten: Can we remove this `TODO` now that we have a unit test covering that case?

tools/infrasys/src/root.rs
```diff
@@ -0,0 +1,196 @@
...
+        ensure!(status.success(), error::TuftoolResult {
+            command: arg_str,
+            code: status.code().map(|i| i.to_string()).unwrap_or_else(|| "<unknown>".to_string())
+        });
+    }
+}
+
+pub fn check_root(root_role_path: &Path) -> Result<()> {
+    if root_role_path.is_file() {
+        warn!("Please delete file at {}", root_role_path.display());
```
etungsten: Probably want to elaborate a bit more.
```suggestion
        warn!("Cowardly refusing to overwrite the existing root.json at {}. Please manually delete it and run again.", root_role_path.display());
```

tools/infrasys/src/s3.rs
```diff
@@ -0,0 +1,365 @@
...
+        formatted[..formatted.len() - 2].to_string()
+    } else {
+        formatted
+    }
+}
+
+/// Creates a *private* S3 Bucket using a CloudFormation template
+/// Input: The region in which the bucket will be created and the name of the bucket
+/// Output: The stack_arn of the bucket created (will be added as a field to Infra.lock)
+pub async fn create_s3_bucket(region: &str, stack_name: &str) -> Result<(String, String, String)> {
```
etungsten, line 39: Should probably add a comment to clarify the tuple we're returning. Another option is to define another struct and use that as the return type.

tools/infrasys/src/shared.rs
```diff
@@ -0,0 +1,80 @@
...
+        .stacks
+        .context(error::ParseResponse {
+            what: "stacks",
+            resource_name: stack_name,
+        })?[0]
+        .clone();
+
+    // Checking that keys have been created so we can return updated outputs
+    let mut status = stack_outputs.stack_status;
+    while status != "CREATE_COMPLETE" {
```
etungsten, line 54: We should probably add a condition for timing out. Track number of attempts and time out if it exceeds 30 minutes or something. Don't want to potentially wait forever if something is stuck.
We should also error out if the stack status indicates failure.

tools/pubsys-config/src/lib.rs
```diff
@@ -49,10 +49,31 @@ impl InfraConfig {
             Ok(Self::default())
         }
     }
+
+    /// Deserializes an InfraConfig.lock to yaml format from a given path
+    pub fn from_path_yaml<P>(path: P) -> Result<Self>
```
etungsten: Should just be `Infra.lock` and not `InfraConfig.lock` I think?

Also do we really need this to be a method? We're only calling this in a unit test right now. 
For `CheckInfraConfig` subcommand we would need it I guess.

Also `yaml` is mostly an implementation detail. We're just deserializing the lock file from path.
```suggestion
    /// Deserializes an InfraConfig from a Infra.lock file at a given path
    pub fn from_lock_path<P>(path: P) -> Result<Self>
```

tools/pubsys-config/src/lib.rs
```diff
@@ -174,6 +220,12 @@ mod error {
             source: toml::de::Error,
         },
 
+        #[snafu(display("Invalid config file at '{}': {}", path.display(), source))]
+        InvalidYaml {
```
etungsten:
```suggestion
        #[snafu(display("Invalid Infra.lock file at '{}': {}", path.display(), source))]
        InvalidLock {
```

etungsten (commented):
I think we talked about this at some point and now is a good time to do it:
In addition to just setting up the Infrastructure, we should also test out whether we're able to use said infrastructure to actually update a live Bottlerocket host with a custom TUF repository. We should include that in the testing description since that's the ultimate goal.

aashnasheth:
Push above addresses etungsten's comments (and includes a small fix for default prefixes)

aashnasheth:
Push above makes a small change to bucket policy tests

aashnasheth:
Push above adds a new commit that checks if a lock file exists or not (signaling users to clean up resources and re-run infrasys accordingly)

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+
+#[derive(Debug, StructOpt)]
+#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+struct Args {
+    #[structopt(global = true, long, default_value = "INFO")]
+    log_level: LevelFilter,
+
+    // Path to Infra.toml  (NOTE: must be specified before subcommand)
+    #[structopt(long, parse(from_os_str))]
+    infra_config_path: PathBuf,
```
webern, line 32: We've seen positional arguments cause problems in the past. I wonder if it is better to use a regular argument such as `--infra-path`. My thinking is that if we wanted to add a subcommand that does not require Infra.toml, we would have a hard time maintaining compatibility with this positional global.

```suggestion
    #[structopt(global = true, long)]
    infra_path: PathBuf,
```

tools/infrasys/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+license = "Apache-2.0 OR MIT"
+authors = ["Aashna Sheth <shetaash@amazon.com>"]
+edition = "2018"
+publish = false
+
+[dependencies]
+hex = "0.4.0"
+rusoto_core = { version = "0.47", default-features = false, features = ["rustls"] }
+rusoto_cloudformation = { version = "0.47", default-features = false, features = ["rustls"] }
+tokio = { version = "1", default-features = false, features = ["macros", "rt-multi-thread"] }
```
webern: We've been locking tokio to an "LTS" version everywhere.

```suggestion
tokio = { version = "~1.8", default-features = false, features = ["macros", "rt-multi-thread"] } # LTS
```

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+#[derive(Debug, StructOpt)]
+#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+struct CreateInfraArgs {
+    #[structopt(long)]
+    root_role_path: PathBuf,
+}
+
+#[derive(Debug, StructOpt)]
+enum SubCommand {
+    CheckInfraLock,
```
webern: These doc comments become help comments when you run `infrasys --help`

```suggestion
    /// Ensures that the Infra.lock file does not exist. Returns an error if it exists.
    CheckInfraLock,
```

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+    CreateInfra(CreateInfraArgs),
```
webern, line 49: These doc comments become help comments when you run `infrasys --help`

```suggestion
    /// Creates infrastructure specified in the Infra.toml file.
    CreateInfra(CreateInfraArgs),
```

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    #[structopt(subcommand)]
+    subcommand: SubCommand,
+}
+
+#[derive(Debug, StructOpt)]
+#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+struct CreateInfraArgs {
+    #[structopt(long)]
+    root_role_path: PathBuf,
```
webern, line 43: These doc comments become the help comments that the program prints:

```suggestion
    /// Path to the root.json file.
    #[structopt(long)]
    root_role_path: PathBuf,
```

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+        SubCommand::CreateInfra(ref run_task_args) => {
+            let rt = Runtime::new().context(error::Runtime)?;
+            rt.block_on(async {
+                create_infra(&args.infra_config_path, &run_task_args.root_role_path).await
+            })
+        }
+    }
+}
+
+async fn check_infra_lock(toml_path: &Path) -> Result<()> {
```
webern: I don't think this function is actually async.

```suggestion
fn check_infra_lock(toml_path: &Path) -> Result<()> {
```

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+    // Parse and store the args passed to the program
+    let args = Args::from_args();
+
+    SimpleLogger::init(args.log_level, LogConfig::default()).context(error::Logger)?;
+
+    match args.subcommand {
+        SubCommand::CheckInfraLock => {
+            let rt = Runtime::new().context(error::Runtime)?;
+            rt.block_on(async { check_infra_lock(&args.infra_config_path).await })
+        }
```
webern: See`check_infra_lock` comment. Doesn't look like this function needs to be async.

```suggestion
        SubCommand::CheckInfraLock => check_infra_lock(&args.infra_config_path),
```

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+    info!("Parsing Infra.toml...");
+    let mut infra_config = InfraConfig::from_path(toml_path).context(error::Config)?;
+    let repos = infra_config
+        .repo
+        .as_mut()
+        .context(error::MissingConfig { missing: "repo" })?;
+
+    for (repo_name, repo_config) in repos.iter_mut() {
+        // Step 0: Grabbing key variables (and throwing errors if not specified)
+        let s3_stack_name =
```
webern: With very long loops or procedures I prefer to see the "guts" extracted to smaller functions.
It would be nice to see the loop reduced to something like:

```rust
    for (repo_name, repo_config) in repos.iter_mut() {
        // Validate repo_config and unwrap required optional data.
        let repo_info = ValidRepoInfo::new(&repo_config)?;
       
        // Validate the key configurations and root file.
        keys::check_signing_key_config(&repo_info.signing_keys)?;
        keys::check_signing_key_config(&repo_info.root_keys)?;
        root::check_root(root_role_path)?;

        // Create the repo.       
        create_repo_infrastructure(repo_name, repo_info).await?;
        update_root_and_sign_root(/*...*/).await?;

        // Upload root.json.
        info!("Uploading root.json to S3 bucket...");
        s3::upload_file(s3_region, &bucket_name, &prefix, root_role_path).await?;

        // Update infra_config with output parameters if not already set
        if repo_config.metadata_base_url.is_none() {
            repo_config.metadata_base_url = Some(
                Url::parse(format!("{}{}/metadata/", &bucket_url, prefix).as_str())
                    .context(error::ParseUrl { input: &bucket_url })?,
            );
        }
        if repo_config.targets_url.is_none() {
            repo_config.targets_url = Some(
                Url::parse(format!("{}{}/targets/", &bucket_url, prefix).as_str())
                    .context(error::ParseUrl { input: &bucket_url })?,
            );
        }
        if repo_config.root_role_url.is_none() {
            repo_config.root_role_url = Some(
                Url::parse(format!("{}{}/root.json", &bucket_url, prefix).as_str())
                    .context(error::ParseUrl { input: &bucket_url })?,
            );

            let root_role_data = fs::read_to_string(&root_role_path).context(error::FileRead {
                path: root_role_path,
            })?;
            let mut d = Sha512::new();
            d.update(&root_role_data);
            let digest = hex::encode(d.finalize());
            repo_config.root_role_sha512 = Some(digest);
        }
    }
```

Where `ValidRepoInfo` is something like: 

```rust
struct ValidRepoInfo {
    repo_name: String,
    s3_stack_name: String,
    s3_region: String,
    vpc_id: String,
    prefix: String,
    signing_keys: Vec<SigningKeyConfig>,
    root_keys: Vec<SigningKeyConfig>,
}
```

In general, I would recommend removing the enumeration of "steps" in comments (each "step" by the way probably represents a good function opportunity). When someone adds or removes something from the procedure they either remember, or forget, to re-number the steps. Either way it makes the comments harder to maintain.

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+            lock_path.display()
+        );
+        error::FileExists { path: lock_path }
+    });
+    Ok(())
+}
+
+/// Automates setting up infrastructure for a custom TUF repo
+async fn create_infra(toml_path: &Path, root_role_path: &Path) -> Result<()> {
+    info!("Parsing Infra.toml...");
```
webern, line 94: Although we have the subcommand for manually ensuring that Infra.lock does not exist, do you think we should add that check here to make sure we don't do bad things™ by accident?

```rust
// TODO - support updating existing infra
check_infra_lock(/*..*/)?;
```

webern (commented):
I only made it through main.rs today. It looks good. Most of what I have are little things not impacting correctness/design.

aashnasheth:
Push above adds further pubsys integration for InfraConfig::from_path_or_default usage (not just InfraConfig::from_path)

tools/infrasys/src/error.rs
```diff
@@ -0,0 +1,160 @@
+use snafu::Snafu;
+use std::io;
+use std::path::PathBuf;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility = "pub(super)")]
+pub enum Error {
```
zmrow, line 7: Nit (I'm sure other there are other opinions here): when using a single error module like this, IMHO the errors should be in alphabetical order.  I think it makes it easy to find what you're looking for, add new variants, etc.

tools/infrasys/src/error.rs
```diff
@@ -0,0 +1,160 @@
...
+    #[snafu(display("Error reading Infra.toml: {}", source))]
+    Config { source: pubsys_config::Error },
+
+    #[snafu(display("Failed to start tuftool: {}", source))]
+    TuftoolSpawn { source: io::Error },
+
+    #[snafu(display("Infra.toml is missing {}", missing))]
+    MissingConfig { missing: String },
+
+    #[snafu(display("Error splitting shell command - {} - input: {}", source, command))]
```
zmrow: The `source` could be fairly long, so in order to avoid losing the context if the `input`, perhaps you could format it like:
```
("Error splitting shell command '{}': {}", command, source)
```

tools/infrasys/src/error.rs
```diff
@@ -0,0 +1,160 @@
...
+        region: String,
+        source: rusoto_core::RusotoError<rusoto_cloudformation::CreateStackError>,
+    },
+
+    #[snafu(display(
+        "Recieved CREATE_FAILED status for CFN stack {} in {}",
+        stack_name,
+        region
+    ))]
+    CreateStackStatus { stack_name: String, region: String },
```
zmrow: Since the status is `CREATE_FAILED`, this error variant should probably be named as such: `CreateStackFailure`?

tools/infrasys/src/error.rs
```diff
@@ -0,0 +1,160 @@
...
+    ))]
+    KeyConfig { missing: String },
+
+    #[snafu(display(
+        "Failed to create new keys or access pre-existing keys in available_keys list."
+    ))]
+    KeyCreation,
+
+    #[snafu(display("Invalid path {} for {}", path.display(), thing))]
+    Path { path: PathBuf, thing: String },
```
zmrow: I think I'd name this `InvalidPath`

tools/infrasys/src/error.rs
```diff
@@ -0,0 +1,160 @@
...
+    InvalidYaml { source: serde_yaml::Error },
+
+    #[snafu(display("Failed to convert {} to yaml: {}", what, source))]
+    InvalidJson {
+        what: String,
+        source: serde_json::Error,
+    },
+
+    #[snafu(display("Failed to get parent of path: {}", path.display()))]
+    Parent { path: PathBuf },
```
zmrow: At least in practice - this seems to crossover with the `Path` error variant above.

tools/infrasys/src/error.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    #[snafu(display("Logger setup error: {}", source))]
+    Logger { source: log::SetLoggerError },
+
+    #[snafu(display("Returned {}: tuftool {}", code, command))]
```
zmrow: I'd reword this as
```
("tuftool {} returned {}", command, code)
```

Also, is there any more context we can give the user in this case?

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,268 @@
...
+            let rt = Runtime::new().context(error::Runtime)?;
+            rt.block_on(async {
+                create_infra(&args.infra_config_path, &run_task_args.root_role_path).await
+            })
+        }
+    }
+}
+
+async fn check_infra_lock(toml_path: &Path) -> Result<()> {
+    let lock_path = toml_path
```
zmrow: It feels odd to me to have a function that's checking for the existence of a file, but the argument is a different file.  Perhaps we can determine `lock_path` beforehand, and pass it to this function.

zmrow (commented):
Publishing what I have for now.

tools/infrasys/src/error.rs
webern: I didn't notice that we were using tuftool. I had hoped we could use `tough` as a library instead of shelling out to the CLI.

tools/infrasys/src/error.rs
aashnasheth: I'm thinking that if there's an issue with the parent() method in specific, might be nice to know where the error is coming from instead of a general "invalid path"?
For some reason if rust's parent() method breaks, we would be able to catch it here

tools/infrasys/src/main.rs
aashnasheth: I think infrasys would always need some sort of input file (which we might want to keep as Infra.toml?) that specifies what infra should be created (or what infra users already have). Currently pubsys uses the non-global infra_config_path, so for consistency we can keep it and change both when needed?

tools/infrasys/src/main.rs
aashnasheth: Agreed! I think we should remove that subcommand entirely and just have it as a method in the file. It seems we are checking Infra.lock existence via code in pubsys anyway. Maybe later if we need the subcommand, we can add it back (doesn't seem like we need it now, if I'm not mistaken).

tools/infrasys/src/error.rs
aashnasheth: Ah yes, we would need to move a few things from tuftool --> tough, which might take a little longer than the length of my internship :(
But adding this to the next steps. We can also refactor pubsys tuftool use while we're at it!

tools/infrasys/src/main.rs
webern: We discussed this and the current name and positional nature of the argument matches pubsys. Resolving.

tools/infrasys/src/error.rs
webern: I think having a separate error variant for this makes sense. The path itself isn't necessarily invalid, it's just that it doesn't have a parent (e.g. it might be `/`).

tools/infrasys/src/error.rs
webern: As we discussed, using `tough` instead of `tuftool` will have to be out of scope for now since we don't have the cycles to make `tough` do everything we need.

aashnasheth: > I'd reword this as
> 
> ```
> ("tuftool {} returned {}", command, code)
> ```
> 
> Also, is there any more context we can give the user in this case?

Because we don't know where tuftool failed, I think the only thing we can provide them is the error code? Doesn't look like ExitStatus outputs much else: https://doc.rust-lang.org/std/process/struct.ExitStatus.html 
But I might be missing a way to grab more info -- can definitely add it in if there's a way

Makefile.toml
```diff
@@ -511,6 +524,38 @@ cargo install \
...
+script = [
+'''
+set -e
+
+export PATH="${BUILDSYS_TOOLS_DIR}/bin:${PATH}"
+
+infrasys \
+   --infra-config-path "${PUBLISH_INFRA_CONFIG_PATH}" \
+   \
+   check-infra-lock
```
webern: As we discussed elsewhere, we can use bash to do this check and remove the somewhat odd subcommand from infrasys (at least for now). Infrasys can also check for the existence of this file before proceeding with `create-infra` and error-out if it finds the file (as an additional safety net).

Makefile.toml
```diff
@@ -511,6 +524,38 @@ cargo install \
 '''
 ]
 
+[tasks.check-infra-lock]
+dependencies = ["infra-tools"]
```
webern: As we have been discussing elsewhere, if we remove the check-infra-lock subcommand then this task will not actually have any dependencies.

```suggestion
```

Makefile.toml
```diff
@@ -511,6 +524,38 @@ cargo install \
...
+
+infrasys \
+   --infra-config-path "${PUBLISH_INFRA_CONFIG_PATH}" \
+   \
+   check-infra-lock 
+'''
+]
+
+[tasks.create-infra]
+dependencies = ["check-infra-lock"]
```
webern: As we have been discussing elsewhere, if we remove the check-infra-lock subcommand then this task is the one that depends on `setup`.

Does `setup` ensure the installation of `tuftool`?

```suggestion
dependencies = ["check-infra-lock", "setup"]
```

tools/infrasys/Cargo.toml
webern: This doesn't seem to be changed yet. We want the tokio version to be `~1.8`. And we have been putting the comment `# LTS` at the end.

tools/pubsys/src/vmware/upload_ova/mod.rs
```diff
@@ -45,13 +45,24 @@ pub(crate) struct UploadArgs {
...
+    info!("Checking for infra config at path: {}", lock_path.display());
+    let infra_config = if lock_path.is_file() {
+        InfraConfig::from_lock_path(lock_path).context(error::InfraConfig)?
+    } else {
+        info!(
+            "Checking for infra config at path: {}",
+            args.infra_config_path.display()
+        );
+        InfraConfig::from_path_or_default(&args.infra_config_path).context(error::InfraConfig)?
+    };
```
webern: These seems to duplicate some code in `validate_repo/mod.rs`. Can a function be extracted? If the error type makes this difficult I can probably help solve that.

tools/pubsys/src/repo/validate_repo/mod.rs
```diff
@@ -122,12 +122,24 @@ fn validate_repo(
...
+    let infra_config = if lock_path.is_file() {
+        info!("Using infra lock from path: {}", lock_path.display());
+        InfraConfig::from_lock_path(lock_path).context(repo_error::Config)?
+    } else {
+        info!(
+            "Using infra config from path: {}",
+            args.infra_config_path.display()
+        );
+        InfraConfig::from_path(&args.infra_config_path).context(repo_error::Config)?
+    };
```
webern: These seems to duplicate some code in `upload_ova/mod.rs`. Can a function be extracted? If the error type makes this difficult I can probably help solve that.

tools/pubsys/src/repo/refresh_repo/mod.rs
```diff
@@ -128,12 +128,24 @@ fn refresh_repo(
...
+    let infra_config = if lock_path.is_file() {
+        info!("Using infra lock from path: {}", lock_path.display());
+        InfraConfig::from_lock_path(lock_path).context(repo_error::Config)?
+    } else {
+        info!(
+            "Using infra config from path: {}",
+            args.infra_config_path.display()
+        );
+        InfraConfig::from_path(&args.infra_config_path).context(repo_error::Config)?
+    };
```
webern: duplicate code, see `validate_repo/mod.rs` and `upload_ova/mod.rs`.

tools/pubsys/src/repo/check_expirations/mod.rs
```diff
@@ -130,12 +130,24 @@ fn check_expirations(
...
+    let infra_config = if lock_path.is_file() {
+        info!("Using infra lock from path: {}", lock_path.display());
+        InfraConfig::from_lock_path(lock_path).context(repo_error::Config)?
+    } else {
+        info!(
+            "Using infra config from path: {}",
+            args.infra_config_path.display()
+        );
+        InfraConfig::from_path(&args.infra_config_path).context(repo_error::Config)?
+    };
```
webern: More of this duplicate code...

tools/pubsys/src/repo.rs
```diff
@@ -424,12 +426,24 @@ pub(crate) fn run(args: &Args, repo_args: &RepoArgs) -> Result<()> {
...
+    info!("Checking for infra config at path: {}", lock_path.display());
+    let infra_config = if lock_path.is_file() {
+        InfraConfig::from_lock_path(lock_path).context(error::Config)?
+    } else {
+        info!(
+            "Checking for infra config at path: {}",
+            args.infra_config_path.display()
+        );
+        InfraConfig::from_path_or_default(&args.infra_config_path).context(error::Config)?
+    };
```
webern: More of the duplicate code...

tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -56,11 +56,24 @@ pub(crate) struct SsmArgs {
...
+    let infra_config = if lock_path.is_file() {
+        info!("Using infra lock from path: {}", lock_path.display());
+        InfraConfig::from_lock_path(lock_path).context(error::Config)?
+    } else {
+        info!(
+            "Using infra config from path: {}",
+            args.infra_config_path.display()
+        );
+        InfraConfig::from_path(&args.infra_config_path).context(error::Config)?
+    };
```
webern: More of that duplicate code...

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -85,12 +85,24 @@ pub(crate) async fn run(args: &Args, publish_args: &PublishArgs) -> Result<()> {
...
+    info!("Checking for infra config at path: {}", lock_path.display());
+    let infra_config = if lock_path.is_file() {
+        InfraConfig::from_lock_path(lock_path).context(error::Config)?
+    } else {
+        info!(
+            "Checking for infra config at path: {}",
+            args.infra_config_path.display()
+        );
+        InfraConfig::from_path_or_default(&args.infra_config_path).context(error::Config)?
+    };
```
webern: duplicate code

tools/pubsys/src/aws/promote_ssm/mod.rs
```diff
@@ -52,11 +52,25 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
...
+        info!("Using infra lock from path: {}", lock_path.display());
+        InfraConfig::from_lock_path(lock_path).context(error::Config)?
+    } else {
+        info!(
+            "Using infra config from path: {}",
+            args.infra_config_path.display()
+        );
+        InfraConfig::from_path(&args.infra_config_path).context(error::Config)?
+    };
+
```
webern: duplicate code

tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -91,12 +91,24 @@ pub(crate) async fn run(args: &Args, ami_args: &AmiArgs) -> Result<()> {
...
+    info!("Checking for infra config at path: {}", lock_path.display());
+    let infra_config = if lock_path.is_file() {
+        InfraConfig::from_lock_path(lock_path).context(error::Config)?
+    } else {
+        info!(
+            "Checking for infra config at path: {}",
+            args.infra_config_path.display()
+        );
+        InfraConfig::from_path_or_default(&args.infra_config_path).context(error::Config)?
+    };
```
webern: duplicate code

tools/infrasys/src/main.rs
aashnasheth: We wanted to make sure that users weren't able to change where the lock file lived (should always live next to Infra.toml). If we pass it in via commandline, we would do this path splicing in the Makefile. We were thinking it would be easier to do via code inside main.rs?

tools/pubsys-setup/src/main.rs
```diff
@@ -177,15 +176,25 @@ fn run() -> Result<()> {
...
+        let infra_config = if lock_path.is_file() {
+            info!("Found infra lock at path: {}", lock_path.display());
+            InfraConfig::from_lock_path(lock_path).context(error::Config)?
+        } else {
+            info!(
+                "Found infra config at path: {}",
+                args.infra_config_path.display()
+            );
+            InfraConfig::from_path(&args.infra_config_path).context(error::Config)?
+        };
```
webern: This looks almost the same as the duplicate code but slightly different. I can't tell if this is actually duplicative.

tools/pubsys-config/src/lib.rs
```diff
@@ -49,10 +49,31 @@ impl InfraConfig {
             Ok(Self::default())
         }
     }
+
+    /// Deserializes an InfraConfig from a Infra.lock file at a given path
+    pub fn from_lock_path<P>(path: P) -> Result<Self>
```
webern: What makes this function different from `from_path`?

This struct's `impl` might be the best place to de-duplicate the code the loads either `Infra.lock` or `Infra.toml`.

webern (changes requested):
I still have a lot left to get through!
But the main theme of this pass is that there is some duplicate code involving the loading of either `Infra.toml` or `Infra.lock`.

It's great when files in a PR have more red than green and there is an opportunity to make that happen here in all the files that need to load either `toml` or `lock`.

aashnasheth:
Push above addresses initial webern and zmrow comments

aashnasheth:
Push above addresses webern code duplication concerns with pubsys-integration code

tools/pubsys-config/src/lib.rs
aashnasheth: Good idea, added another method `from_path_or_lock` that we call from other files. 
`from_lock_path` is different because it uses yaml to deserialize instead of toml (toml serialization didn't work, so we had to serialize using yaml)

Some other notes:
- Still kept the `from_lock_path` method separate for clarity. And I think there might be cases where we just want to read the lock path (like when we just want to read the toml path), so it might be nice to keep

- In some cases, we wanted a default to be created if a toml didn't exist either (previously using `from_path_or_default`), so I added a default flag -- does that look okay?

aashnasheth:
Push above removes `check-infra-lock` subcommand and calls it as a method at the beginning of the `run()` method instead

Makefile.toml
aashnasheth: Made it dependent on infra-tools which is dependent on setup (added tuftool as a dependency for infra-tools!)

Makefile.toml
aashnasheth: Just added it as an extra method, so we don't have to pass in an extra parameter in bash. It will always run before anything in infrasys runs

aashnasheth:
Push above adds KMSClient code to pubsys-integration commit

tools/infrasys/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+log = "0.4.14"  
+structopt = { version = "0.3", default-features = false }  
+clap = "2.33"  
+url = "2.2.2"  
+async-trait = "0.1.51"  
+shell-words = "1.0.0" 
+sha2 = "0.9"
+serde_yaml = "0.8.17"
+simplelog = "0.10.0"
+serde_json = "1.0.66"
```
webern: Sort me! :)

```suggestion
async-trait = "0.1.51"  
clap = "2.33"  
hex = "0.4.0"
log = "0.4.14"  
pubsys-config = { path = "../pubsys-config/" }  
rusoto_cloudformation = { version = "0.47", default-features = false, features = ["rustls"] }
rusoto_core = { version = "0.47", default-features = false, features = ["rustls"] }
rusoto_s3 = { version = "0.47", default-features = false, features = ["rustls"] }
serde_json = "1.0.66"
serde_yaml = "0.8.17"
sha2 = "0.9"
shell-words = "1.0.0" 
simplelog = "0.10.0"
snafu = "0.6"  
structopt = { version = "0.3", default-features = false }  
tokio = { version = "~1.8", default-features = false, features = ["macros", "rt-multi-thread"] } # LTS
toml = "0.5"
url = "2.2.2"  
```

tools/infrasys/src/error.rs
```diff
@@ -0,0 +1,165 @@
...
+    #[snafu(display("File already exists at '{}'", path.display()))]
+    FileExists { path: PathBuf },
+
+    #[snafu(display("Failed to open file at '{}': {}", path.display(), source))]
+    FileOpen { path: PathBuf, source: io::Error },
+
+    #[snafu(display("Failed to read file at '{}': {}", path.display(), source))]
+    FileRead { path: PathBuf, source: io::Error },
+
+    #[snafu(display("Failed to write file in '{}': {}", path.display(), source))]
```
webern: Is this path a directory or the actual file path that failed to write"?

```suggestion
    #[snafu(display("Failed to write file at '{}': {}", path.display(), source))]
```

aashnasheth:
Push above adds webern's code to clean up tools/pubsys-setup/src/main.rs diff

tools/pubsys-config/Cargo.toml
```diff
@@ -16,3 +16,4 @@ serde = { version = "1.0", features = ["derive"]  }
 snafu = "0.6"
 toml = "0.5"
 url = { version = "2.1.0", features = ["serde"] }
+serde_yaml = "0.8.17"
```
webern: Let's keep these sorted alphabetically.

tools/pubsys/src/repo.rs
```diff
@@ -388,20 +391,38 @@ fn is_file_not_found_error(e: &tough::error::Error) -> bool {
...
-            key_id: key_id.clone(),
-            client: None,
+            key_id: key_id
+                .clone()
+                .context(error::MissingConfig { missing: "key_id" })?,
+            client: if let Some(region) = config
+                .as_ref()
+                .context(error::MissingConfig {
+                    missing: "KMS config",
+                })?
```
webern: It looks like we've added configuration that is not-optional, which is a breaking change for existing files that do not have these configuration items. Am I understanding that correctly? We would want to make sure existing Infra.toml files do not become invalid.

tools/infrasys/src/shared.rs
```diff
@@ -0,0 +1,91 @@
...
+    Publication,
+}
+
+/// Retrieve a BUILDSYS_* variable that we expect to be set in the environment
+pub fn getenv(var: &str) -> Result<String> {
+    env::var(var).context(error::Environment { var })
+}
+
+/// Generates a parameter type object used to specify parameters in CloudFormation templates
+pub fn get_parameter(key: String, val: String) -> Parameter {
```
webern: Nit: maybe call it `create_parameter` or `make_parameter`, or `new_parameter`.
 
```suggestion
pub fn create_parameter(key: String, val: String) -> Parameter {
```

tools/infrasys/src/shared.rs
```diff
@@ -0,0 +1,91 @@
...
+        .context(error::ParseResponse {
+            what: "stacks",
+            resource_name: stack_name,
+        })?[0]
+        .clone();
+
+    // Checking that keys have been created so we can return updated outputs
+    let mut status = stack_outputs.stack_status;
+    // Max wait is 30 mins (90s*20s = 1800s = 30mins)
+    let mut max_attempts: u32 = 90;
```
webern: 30m seems a little excessive, doesn't it?

```suggestion
    // Max wait is 10 mins (30 attempts * 20s = 600s = 10mins)
    let mut max_attempts: u32 = 30;
```

tools/infrasys/src/s3.rs
```diff
@@ -0,0 +1,366 @@
...
+pub async fn add_bucket_policy(
+    region: &str,
+    bucket_name: &str,
+    prefix: &str,
+    vpcid: &str,
+) -> Result<()> {
+    // Get old policy
+    let s3_client =
+        S3Client::new(Region::from_str(region).context(error::ParseRegion { what: region })?);
+    let mut current_bp: serde_json::Value = match s3_client
```
webern: Nit: once you amend it, calling it 'current' doesn't quite work. And abbreviating with `bp` obscures what it represents. (I think the same change would be nice in the tests.)

```suggestion
    let mut policy: serde_json::Value = match s3_client
```

tools/infrasys/src/root.rs
```diff
@@ -0,0 +1,195 @@
...
+            .env("AWS_REGION", $region)
+            .status()
+            .context(error::TuftoolSpawn)?;
+
+        ensure!(status.success(), error::TuftoolResult {
+            command: arg_str,
+            code: status.code().map(|i| i.to_string()).unwrap_or_else(|| "<unknown>".to_string())
+        });
+    }
+}
```
webern, line 31: We should try to dedup this if possible. Can the existing macro be used if it is moved to pubsys-config and exported as part of that lib?

i.e. move this and use it?
https://github.com/bottlerocket-os/bottlerocket/blob/develop/tools/pubsys-setup/src/main.rs#L51...L71

tools/infrasys/src/root.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// Adds KMSKeys to root.json given root or publication type
+/// Input: available-keys (keys to sign with), role (root or publication), threshold for role, filepath for root.JSON,
+/// mutable key_id
+/// Output: in-place edit of root.json and key_id with a valid publication key
+/// (If key-id is populated, it will not change. Otherwise, it will be populated with a key-id of an available key)
+fn add_keys_kms(
+    available_keys: &HashMap<String, String>,
+    role: &KeyRole,
+    threshold: &str,
```
webern: Get this parsed as part of `ValidRepoInfo` and pass it everywhere as `NonZeroUsize`. See comment at `ValidRepoInfo` as well.

```suggestion
    threshold: NonZeroUsize,
```

tools/infrasys/src/main.rs
```diff
@@ -0,0 +1,340 @@
...
+    info!("Complete!");
+    Ok(())
+}
+
+struct ValidRepoInfo<'a> {
+    bucket_name: &'a mut Option<String>,
+    metadata_base_url: &'a mut Option<Url>,
+    prefix: String,
+    pub_key_threshold: &'a String,
+    root_key_threshold: &'a String,
```
webern: Parse these in the `new` function to a usize so you don't have to parse it later. Change the function signatures that take threshold as a &str to take it as a `NonZeroUsize`.

```suggestion
    pub_key_threshold: NonZeroUsize,
    root_key_threshold: NonZeroUsize,
```

tools/infrasys/src/keys.rs
```diff
@@ -0,0 +1,139 @@
...
+            ) {
+                (true, true, None) => error::KeyConfig {
+                    missing: "an available_key or region/key_alias",
+                }
+                .fail()?,
+                (_, false, None) => error::KeyConfig {
+                    missing: "key_alias",
+                }
+                .fail()?,
+                (_, true, Some(..)) => error::KeyConfig { missing: "region" }.fail()?,
```
webern: Can you add comments to each of these cases explaining why they are invalid?

tools/infrasys/src/keys.rs
```diff
@@ -0,0 +1,139 @@
...
+                _ => (),
+            };
+        }
+        SigningKeyConfig::ssm { .. } => (),
+    }
+    Ok(())
+}
+
+/// Must create a trait because can't directly implement a method for an struct in an external crate like KMSKeyConfig (which lives in pubsys-config/lib.rs)
```
webern: Wrap me! :)

tools/infrasys/src/keys.rs
```diff
@@ -0,0 +1,139 @@
...
+#[async_trait]
+trait KMSKeyConfigExt {
+    async fn create_kms_keys(&mut self) -> Result<()>;
+}
+
+/// Creates new KMS keys using cloudformation in regions specified
+/// Input Conditions: Alias+Region or AvailableKeys must be specified
+/// Output: Populates KMSKeyConfig with information about resources created
+/// available-keys starts as a map of pre-existing keyids:regions and will end as a map of pre-existing and generated keyids:regions
```
webern: Wrap me! :)

webern (changes requested):
It's getting close, Nice work!

tools/infrasys/src/shared.rs
etungsten: I think a 30 minutes timeout is appropriate considering some AWS resources take a long time to stand up in a CFN stack. The goal is to not allow it to run beyond what's reasonable when something gets stuck in CloudFormation. 10 minutes would be too short if we decide to add other resources to stand up in Infrasys (VPCs, etc).

tools/pubsys/src/repo.rs
etungsten: ~This is making it optional. If `available_keys` is empty/ non-existent in the config (which is what Infra.toml looks like now), we just use the default region (like before).~
You're right that it shouldn't error if KMS config was unspecified. For some reason during testing, it wasn't erroring out when Infra.toml didn't have that config section specified.

tools/infrasys/src/shared.rs
aashnasheth: KMS keys have been taking quite a while to stand up :sweat_smile:

tools/pubsys/src/repo.rs
etungsten: Interestingly, when `config` is not specified, since it's a flattened struct, it'll always be `Some`:
From infra.toml:
```
[repo.default]
....
signing_keys = { kms = { key_id = "arn:aws:kms:us-west-2:blah" } }

```

results in:
```
[pubsys/src/repo.rs:395] signing_key_config = kms {
    key_id: Some(
        "arn:aws:kms:us-west-2:blah",
    ),
    config: Some(
        KMSKeyConfig {
            available_keys: {},
            key_alias: None,
            regions: [],
            key_stack_arns: {},
        },
    ),
}

```

Should still not rely on this behavior though.

tools/infrasys/src/main.rs
aashnasheth: I use thresholds most commonly as strings and only once as a number to check it is within bounds, does it still make sense to parse beforehand? (I'd have to convert back to a string in the other places I'm using the threshold)

tools/infrasys/src/shared.rs
webern: Ok, ~~(sadly) resolving~~.

webern: Although it might be nice to fix the formatting a unit nits.

tools/infrasys/src/root.rs
webern: Turns out that there is a subtle difference in these two macros that makes de-duplicating arguably more trouble than it's worth. Resolving.

aashnasheth:
Push addresses webern comments

etungsten (approved):

webern (approved):
🎉

Issue #1581, opened 2021-05-18
CloudFormation signaling


gabegorelick:
**What I'd like:**
It would be nice if there was an easy way to call CloudFormation's [`SignalResource`](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_SignalResource.html) when booting a Bottlerocket instance. This is typically considered a best practice when creating an ASG in CloudFormation so that it can roll back to an earlier LaunchTemplate or LaunchConfig if the instances don't come online.

See, for example, the ECS CloudFormation reference architecture, which uses the `cfn-signal` CLI: https://github.com/aws-samples/ecs-refarch-cloudformation/blob/a257e226b33bd9d2a721e5afd9d7e8b66dbacfdc/infrastructure/ecs-cluster.yaml#L87

In Bottlerocket's case, a typical boot issue I've encountered is passing malformed user data. In such a case, Bottlerocket's `early-boot-config.service` will fail. But if you don't signal CloudFormation, CloudFormation will still consider the deploy a success, potentially leaving you with no working instances.

**Any alternatives you've considered:**

Running `cfn-signal` in a bootstrap container would probably work. But it's not clear to me that bootstrap containers run late enough in the boot sequence to verify that all services are up.

mello7tre:
Problem is even bigger if you use ASG CreationPolicy:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html

Having no way to signal CF (usually you did this using cfn-signal), you need to either remove policy or set MinSuccessfulInstancesPercent to zero.

mello7tre:
Just one implementation idea:

- create a rust program for signaling CF (need to wait that rust sdk implement instance role credential, awslabs/smithy-rs/issues/466)
  -  program should be configurable by Env variables
- create a new settings section, like:
```
[settings.aws.cloudformation]
"signal" = true/false
"stack-name" = ""
"logical-resource-id" = ""
```
- Create a service, `cf-signal.service`:
```
[Unit]
Description=Send signal to CloudFormation Stack
Wants=network-online.target
After=multi-user.target

[Service]
Type=simple
RemainAfterExit=true
EnvironmentFile=/etc/cf-signal.env
ExecStart=/bin/sh -c "STATUS=$(/usr/bin/systemctl --wait is-system-running) /usr/bin/cf-signal"

[Install]
WantedBy=multi-user.target
```
[...]

PR #1728, opened 2021-09-01
CloudFormation signal program (issue #1581)

mello7tre:
**Issue number:**
#1581

**Description of changes:**
Created a new rust program, cfsignal to send signal to CloudFormation
Stack.
Program is a sort of cfn-signal
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html
but as cfn-signal need python cannot be used by bottlerocket.

cfsignal read configuration from a cfsignal.toml file configured reading
user-data, so it depends on settings-applier.service.
It cannot send a signal for a failure happening before
settings-applier.service and network-online.target are started.

It is able to send a failure signal for any other service starting from
(included):
activate-multi-user.service

It use systemctl action is-system-running with --wait option.
This way we can know if any service, after systemd boot process
finished, is in a failure status.

***Detail***
I have created a new setting named `cloudformation`.
It support the keys:
```
signal: boolean
stack_name: string
logical_resource_id: string
```

If signal is false (default) program do nothing.
Other way it excute `systemctl --wait is-system-running` to find out the system state.
Then it send the relative signal to the the _stack_name_ stack.

**Testing done:**
Created and updated and AutoScalingGroup configured to wait signaling by created instances.
Tested that instance send SUCCESS signal if all service started successfully.
```
bash-5.0# systemctl status cfsignal
● cfsignal.service - Send signal to CloudFormation Stack
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/cfsignal.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Thu 2022-02-24 03:20:40 UTC; 10min ago
    Process: 1578 ExecStart=/usr/bin/cfsignal (code=exited, status=0/SUCCESS)
   Main PID: 1578 (code=exited, status=0/SUCCESS)

Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] System status is: running [0]
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] Connecting to IMDS
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] Received meta-data/instance-id
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] Received dynamic/instance-identity/document
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] Region: "us-west-2" - InstanceID: "i-051f097c469cd6a7f" - Signal: "SUCCESS"
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal systemd[1]: cfsignal.service: Succeeded.
```

Tested that instance send FAILURE signal if a service after `activate-multi-user.service` fail.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,6 +419,13 @@ These settings can be changed at any time.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
 * `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
 
+#### CloudFormation signal helper settings
+
+For AWS variants, these settings allow you to set up CloudFormation signaling to indicate whether Bottlerocket hosts running in EC2 have been successfully created or updated:
+* `settings.cloudformation.should-signal`: Whether to check status and send signal. Defaults to `false`. If set to `true`, both `stack-name` and `logical-resource-id` need to be specified.
+* `settings.cloudformation.stack-name`: Name of the CloudFormation Stack to signal.
+* `settings.cloudformation.logical-resource-id`: The logical ID of the AutoScalingGroup resource that you want to signal.
+
 #### OCI Hooks settings
 
 Bottlerocket allows you to opt-in to use additional [OCI hooks](https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle) for your orchestrated containers.
--- a/Release.toml
+++ b/Release.toml
@@ -101,3 +101,6 @@ version = "1.6.1"
     "migrate_v1.6.0_public-control-container-v0-5-5.lz4",
 ]
 "(1.6.0, 1.6.1)" = []
+"(1.6.1, 1.6.2)" = [
+    "migrate_v1.6.2_add-cfsignal.lz4",
+]
--- a/packages/os/cfsignal.service
+++ b/packages/os/cfsignal.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Send signal to CloudFormation Stack
+After=network-online.target settings-applier.service
+Requires=settings-applier.service
+Wants=network-online.target
+# We only want to run once, at first boot. This file is created by cfsignal
+# after a successful run.
+ConditionPathExists=!/var/lib/bottlerocket/cfsignal.ran
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/cfsignal
+
+[Install]
+WantedBy=preconfigured.target
--- a/sources/cfsignal/Cargo.toml
+++ b/sources/cfsignal/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "cfsignal"
+version = "0.1.0"
+license = "Apache-2.0 OR MIT"
+edition = "2018"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+log = "0.4"
+serde = { version = "1.0", features = ["derive"] }
+simplelog = "0.11"
+snafu = { version = "0.7" }
+toml = "0.5.1"
+tokio = { version = "~1.14", default-features = false, features = ["macros", "rt-multi-thread"] }
+rusoto_core = { version = "0.47.0", default-features = false, features = ["rustls"] }
+rusoto_cloudformation = { version = "0.47.0", default-features = false, features = ["rustls"] }
+imdsclient = { path = "../imdsclient", version = "0.1.0" }
+hyper = "=0.14.2"
+
+[build-dependencies]
+cargo-readme = "3.1"
--- a/sources/cfsignal/README.md
+++ b/sources/cfsignal/README.md
@@ -1,0 +1,22 @@
+# cfsignal
+
+Current version: 0.1.0
+
+### Introduction
+
+Cfsignal is similar to [cfn-signal](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html).
+
+When creating an Auto Scaling Group, CloudFormation can be configured to wait for the expected number of signals from instances before considering the ASG successfully created. See [CreationPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html) and [UpdatePolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html) for more details.
+
+Cfsignal uses `systemctl is-system-running` to determine whether the boot has succeeded or failed, and sends the corresponding signal to the CloudFormation stack.
+
+### Configuration
+
+Configuration is read from a TOML file, which is generated from Bottlerocket settings:
+* `should_signal`: Whether to check system status and send signal.
+* `stack_name`: Name of the CFN stack to signal.
+* `logical_resource_id`: The logical ID of the AutoScalingGroup resource that you want to signal.
+
+## Colophon
+
+This text was generated from `README.tpl` using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
--- a/sources/cfsignal/README.tpl
+++ b/sources/cfsignal/README.tpl
@@ -1,0 +1,9 @@
+# {{crate}}
+
+Current version: {{version}}
+
+{{readme}}
+
+## Colophon
+
+This text was generated from `README.tpl` using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
--- a/sources/cfsignal/src/cloudformation.rs
+++ b/sources/cfsignal/src/cloudformation.rs
@@ -1,0 +1,61 @@
+use crate::error::{self, Result};
+use imdsclient::ImdsClient;
+use log::info;
+use rusoto_cloudformation::{CloudFormation, CloudFormationClient, SignalResourceInput};
+use rusoto_core::Region;
+use snafu::{OptionExt, ResultExt};
+use std::str::FromStr;
+
+/// Signals Cloudformation stack resource
+pub async fn signal_resource(
+    stack_name: String,
+    logical_resource_id: String,
+    status: String,
+) -> Result<()> {
+    info!("Connecting to IMDS");
+    let mut client = ImdsClient::new();
+    let instance_id = get_instance_id(&mut client).await?;
+    let region = get_region(&mut client).await?;
+
+    info!(
+        "Region: {:?} - InstanceID: {:?} - Signal: {:?}",
+        region, instance_id, status
+    );
+
+    let client = CloudFormationClient::new(
+        Region::from_str(&region).context(error::RegionParseSnafu { region })?,
+    );
+    let signal_resource_input = SignalResourceInput {
+        stack_name,
+        logical_resource_id,
+        status,
+        unique_id: instance_id,
+    };
+
+    client
+        .signal_resource(signal_resource_input)
+        .await
+        .context(error::SignalResourceSnafu)?;
+
+    Ok(())
+}
+
+/// Returns the instanceId
+async fn get_instance_id(client: &mut ImdsClient) -> Result<String> {
+    client
+        .fetch_instance_id()
+        .await
+        .context(error::ImdsRequestSnafu)?
+        .context(error::ImdsNoneSnafu {
+            what: "instance-id",
+        })
+}
+
+/// Returns the region
+async fn get_region(client: &mut ImdsClient) -> Result<String> {
+    client
+        .fetch_region()
+        .await
+        .context(error::ImdsRequestSnafu)?
+        .context(error::ImdsNoneSnafu { what: "region" })
+}
--- a/sources/cfsignal/src/config.rs
+++ b/sources/cfsignal/src/config.rs
@@ -1,0 +1,20 @@
+use crate::error::{self, Result};
+use serde::Deserialize;
+use snafu::ResultExt;
+use std::fs;
+use std::path::Path;
+
+#[derive(Debug, Deserialize)]
+pub(crate) struct Config {
+    pub(crate) should_signal: bool,
+    pub(crate) stack_name: String,
+    pub(crate) logical_resource_id: String,
+}
+
+impl Config {
+    pub(crate) fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let path = path.as_ref();
+        let s = fs::read_to_string(path).context(error::ConfigReadSnafu { path })?;
+        toml::from_str(&s).context(error::ConfigParseSnafu { path })
+    }
+}
--- a/sources/cfsignal/src/error.rs
+++ b/sources/cfsignal/src/error.rs
@@ -1,0 +1,46 @@
+//! Provides the list of errors for `cfsignal`.
+
+use snafu::Snafu;
+use std::path::PathBuf;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub(super)))]
+pub enum Error {
+    #[snafu(display("Command '{}' with args '{:?}' failed: {}", command, args, source))]
+    Command {
+        command: String,
+        args: Vec<String>,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("Failed to parse config file {}: {}", path.display(), source))]
+    ConfigParse {
+        path: PathBuf,
+        source: toml::de::Error,
+    },
+
+    #[snafu(display("Failed to read config file {}: {}", path.display(), source))]
+    ConfigRead {
+        path: PathBuf,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("IMDS request failed: {}", source))]
+    ImdsRequest { source: imdsclient::Error },
+
+    #[snafu(display("IMDS request failed: No '{}' found", what))]
+    ImdsNone { what: String },
+
+    #[snafu(display("SignalResource request failed: {}", source))]
+    SignalResource {
+        source: rusoto_core::RusotoError<rusoto_cloudformation::SignalResourceError>,
+    },
+
+    #[snafu(display("Unable to parse '{}' as a region: {}", region, source))]
+    RegionParse {
+        region: String,
+        source: rusoto_core::region::ParseRegionError,
+    },
+}
--- a/sources/cfsignal/src/main.rs
+++ b/sources/cfsignal/src/main.rs
@@ -1,0 +1,92 @@
+#![deny(rust_2018_idioms, unused_imports)]
+
+/*!
+## Introduction
+
+Cfsignal is similar to [cfn-signal](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html).
+
+When creating an Auto Scaling Group, CloudFormation can be configured to wait for the expected number of signals from instances before considering the ASG successfully created. See [CreationPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html) and [UpdatePolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html) for more details.
+
+Cfsignal uses `systemctl is-system-running` to determine whether the boot has succeeded or failed, and sends the corresponding signal to the CloudFormation stack.
+
+## Configuration
+
+Configuration is read from a TOML file, which is generated from Bottlerocket settings:
+* `should_signal`: Whether to check system status and send signal.
+* `stack_name`: Name of the CFN stack to signal.
+* `logical_resource_id`: The logical ID of the AutoScalingGroup resource that you want to signal.
+*/
+
+#![deny(rust_2018_idioms)]
+
+mod cloudformation;
+mod config;
+mod error;
+mod system_check;
+
+use crate::config::Config;
+use crate::error::Result;
+use crate::system_check::SystemCheck;
+use cloudformation::signal_resource;
+use log::LevelFilter;
+use log::{error, info, warn};
+use simplelog::{Config as LogConfig, SimpleLogger};
+use std::fs;
+use std::process;
+
+// We only want to run cfsignal once, at first boot.  Our systemd unit file has a
+// ConditionPathExists that will prevent it from running again if this file exists.
+// We create it after running successfully.
+const MARKER_FILE: &str = "/var/lib/bottlerocket/cfsignal.ran";
+const CFSIGNAL_TOML: &str = "/etc/cfsignal.toml";
+
+/// pub(crate) for testing.
+async fn run() -> Result<()> {
+    SimpleLogger::init(LevelFilter::Info, LogConfig::default())
+        .expect("unable to configure logger");
+
+    // load the cfsignal config file
+    let config = Config::from_file(CFSIGNAL_TOML)?;
+
+    let mut signal_status = "FAILURE";
+    let system_check = Box::new(SystemCheck {});
+    let system_status = system_check.system_running()?;
+    info!(
+        "System status is: {} [{}]",
+        system_status.status, system_status.exit_code
+    );
+
+    // run only if the opt-in flag is set
+    if config.should_signal {
+        if system_status.is_healthy {
+            signal_status = "SUCCESS";
+        }
+
+        if let Err(err) = signal_resource(
+            config.stack_name,
+            config.logical_resource_id,
+            signal_status.to_owned(),
+        )
+        .await
+        {
+            error!("Error while sending signal: {}", err);
+        }
+    }
+
+    fs::write(MARKER_FILE, "").unwrap_or_else(|e| {
+        warn!(
+            "Failed to create marker file '{}', may unexpectedly run again: '{}'",
+            MARKER_FILE, e
+        )
+    });
+
+    Ok(())
+}
+
+#[tokio::main]
+async fn main() {
+    if let Err(e) = run().await {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/cfsignal/src/system_check.rs
+++ b/sources/cfsignal/src/system_check.rs
@@ -1,0 +1,62 @@
+use crate::error::{self, Result};
+use log::trace;
+use snafu::ResultExt;
+use std::process::Command;
+
+const SYSTEMCTL: &str = "/usr/bin/systemctl";
+
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub(crate) struct SystemHealth {
+    /// Whether or not the system is healthy.
+    pub(crate) is_healthy: bool,
+    /// The system status description.
+    pub(crate) status: String,
+    /// The system status code.
+    pub(crate) exit_code: i32,
+}
+
+pub(crate) struct SystemCheck {}
+
+impl SystemCheck {
+    pub fn system_running(&self) -> Result<SystemHealth> {
+        let mut result = is_system_running()?;
+        // remove trailing newline
+        result.stdout.pop();
+        Ok(SystemHealth {
+            is_healthy: result.is_success(),
+            status: result.stdout,
+            exit_code: result.exit,
+        })
+    }
+}
+
+struct Outcome {
+    exit: i32,
+    stdout: String,
+}
+
+impl Outcome {
+    fn is_success(&self) -> bool {
+        self.exit == 0
+    }
+}
+
+fn systemctl(args: &[&str]) -> Result<Outcome> {
+    trace!("calling systemctl with '{:?}'", args);
+    let output = Command::new(SYSTEMCTL)
+        .args(args)
+        .output()
+        .with_context(|_| error::CommandSnafu {
+            command: SYSTEMCTL,
+            args: args.iter().map(|&s| s.to_owned()).collect::<Vec<String>>(),
+        })?;
+    Ok(Outcome {
+        exit: output.status.code().unwrap_or(-1),
+        stdout: String::from_utf8_lossy(output.stdout.as_slice()).into(),
+    })
+}
+
+fn is_system_running() -> Result<Outcome> {
+    let outcome = systemctl(&["--wait", "is-system-running"])?;
+    Ok(outcome)
+}
--- a/sources/imdsclient/src/lib.rs
+++ b/sources/imdsclient/src/lib.rs
@@ -157,6 +157,12 @@ impl ImdsClient {
         self.fetch_string(&instance_type_target).await
     }
 
+    /// Gets the instance-id from instance metadata.
+    pub async fn fetch_instance_id(&mut self) -> Result<Option<String>> {
+        let instance_type_target = "meta-data/instance-id";
+        self.fetch_string(&instance_type_target).await
+    }
+
     /// Returns a list of public ssh keys skipping any keys that do not start with 'ssh'.
     pub async fn fetch_public_ssh_keys(&mut self) -> Result<Option<Vec<String>>> {
         info!("Fetching list of available public keys from IMDS");
--- a/sources/models/shared-defaults/cf-signal.toml
+++ b/sources/models/shared-defaults/cf-signal.toml
@@ -1,0 +1,15 @@
+[settings.cloudformation]
+should-signal = false
+stack-name = ""
+logical-resource-id = ""
+
+[services.cfsignal]
+configuration-files = ["cfsignal-toml"]
+restart-commands = ["/bin/systemctl try-restart cfsignal.service"]
+
+[configuration-files.cfsignal-toml]
+path = "/etc/cfsignal.toml"
+template-path = "/usr/share/templates/cfsignal-toml"
+
+[metadata.settings.cloudformation]
+affected-services = ["cfsignal"]
--- a/sources/models/src/aws-k8s-1.21/mod.rs
+++ b/sources/models/src/aws-k8s-1.21/mod.rs
@@ -4,9 +4,9 @@ use std::collections::HashMap;
 
 use crate::modeled_types::Identifier;
 use crate::{
-    AwsSettings, BootstrapContainer, HostContainer, KernelSettings, KubernetesSettings,
-    MetricsSettings, NetworkSettings, NtpSettings, OciHooks, PemCertificate, RegistrySettings,
-    UpdatesSettings,
+    AwsSettings, BootstrapContainer, CloudFormationSettings, HostContainer, KernelSettings,
+    KubernetesSettings, MetricsSettings, NetworkSettings, NtpSettings, OciHooks, PemCertificate,
+    RegistrySettings, UpdatesSettings,
 };
 
 // Note: we have to use 'rename' here because the top-level Settings structure is the only one
@@ -26,4 +26,5 @@ struct Settings {
     pki: HashMap<Identifier, PemCertificate>,
     container_registry: RegistrySettings,
     oci_hooks: OciHooks,
+    cloudformation: CloudFormationSettings,
 }
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -270,6 +270,14 @@ struct MetricsSettings {
     service_checks: Vec<String>,
 }
 
+// CloudFormation settings
+#[model]
+struct CloudFormationSettings {
+    should_signal: bool,
+    stack_name: SingleLineString,
+    logical_resource_id: SingleLineString,
+}
+
 ///// Internal services
 
 // Note: Top-level objects that get returned from the API should have a "rename" attribute
--- a/packages/os/Cargo.toml
+++ b/packages/os/Cargo.toml
@@ -22,7 +22,8 @@ source-groups = [
     "imdsclient",
     "retry-read",
     "shimpei",
-    "driverdog"
+    "driverdog",
+    "cfsignal",
 ]
 
 [lib]
--- a/packages/os/cfsignal-toml
+++ b/packages/os/cfsignal-toml
@@ -1,0 +1,3 @@
+should_signal = {{settings.cloudformation.should-signal}}
+stack_name = "{{settings.cloudformation.stack-name}}"
+logical_resource_id = "{{settings.cloudformation.logical-resource-id}}"
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -26,6 +26,7 @@ Source5: updog-toml
 Source6: metricdog-toml
 Source7: host-ctr-toml
 Source8: oci-default-hooks-json
+Source9: cfsignal-toml
 
 # 1xx sources: systemd units
 Source100: apiserver.service
@@ -42,6 +43,7 @@ Source113: send-boot-success.service
 Source114: bootstrap-containers@.service
 Source115: link-kernel-modules.service
 Source116: load-kernel-modules.service
+Source117: cfsignal.service
 
 # 2xx sources: tmpfilesd configs
 Source200: migration-tmpfiles.conf
@@ -88,6 +90,7 @@ Requires: %{_cross_os}static-pods
 
 %if %{_is_aws_variant}
 Requires: %{_cross_os}shibaken
+Requires: %{_cross_os}cfsignal
 %endif
 
 %if "%{_cross_variant}" == "aws-ecs-1"
@@ -249,6 +252,11 @@ Summary: Manages user-defined K8S static pods
 Summary: Setting generator for populating admin container user-data from IMDS.
 %description -n %{_cross_os}shibaken
 %{summary}.
+
+%package -n %{_cross_os}cfsignal
+Summary: Bottlerocket CloudFormation Stack signaler
+%description -n %{_cross_os}cfsignal
+%{summary}.
 %endif
 
 %package -n %{_cross_os}shimpei
@@ -336,6 +344,7 @@ echo "** Output from non-static builds:"
 %endif
 %if %{_is_aws_variant}
     -p shibaken \
+    -p cfsignal \
 %endif
 %if %{_is_k8s_variant}
 %if %{_is_aws_variant}
@@ -372,6 +381,7 @@ for p in \
 %endif
 %if %{_is_aws_variant}
   shibaken \
+  cfsignal \
 %endif
 %if %{_is_k8s_variant}
 %if %{_is_aws_variant}
@@ -438,6 +448,11 @@ install -p -m 0644 \
   %{S:115} %{S:116} \
 %endif
   %{buildroot}%{_cross_unitdir}
+
+%if %{_is_aws_variant}
+install -p -m 0644 %{S:9} %{buildroot}%{_cross_templatedir}
+install -p -m 0644 %{S:117} %{buildroot}%{_cross_unitdir}
+%endif
 
 install -d %{buildroot}%{_cross_tmpfilesdir}
 install -p -m 0644 %{S:200} %{buildroot}%{_cross_tmpfilesdir}/migration.conf
@@ -554,6 +569,12 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 %if %{_is_aws_variant}
 %files -n %{_cross_os}shibaken
 %{_cross_bindir}/shibaken
+
+%files -n %{_cross_os}cfsignal
+%{_cross_bindir}/cfsignal
+%dir %{_cross_templatedir}
+%{_cross_templatedir}/cfsignal-toml
+%{_cross_unitdir}/cfsignal.service
 %endif
 
 %if %{_is_k8s_variant}
--- a/packages/selinux-policy/fs.cil
+++ b/packages/selinux-policy/fs.cil
@@ -44,6 +44,7 @@
 (filecon "/.*/usr/bin/early-boot-config" file api_exec)
 (filecon "/.*/usr/bin/migrator" file api_exec)
 (filecon "/.*/usr/bin/storewolf" file api_exec)
+(filecon "/.*/usr/bin/cfsignal" file api_exec)
 (filecon "/.*/usr/bin/dbus-broker.*" file bus_exec)
 (filecon "/.*/usr/sbin/chronyd" file clock_exec)
 (filecon "/.*/usr/sbin/wicked.*" file network_exec)
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -216,6 +216,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "add-cfsignal"
+version = "0.1.0"
+dependencies = [
+ "migration-helpers",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -647,6 +654,23 @@ name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
+
+[[package]]
+name = "cfsignal"
+version = "0.1.0"
+dependencies = [
+ "cargo-readme",
+ "hyper",
+ "imdsclient",
+ "log",
+ "rusoto_cloudformation",
+ "rusoto_core",
+ "serde",
+ "simplelog",
+ "snafu",
+ "tokio",
+ "toml",
+]
 
 [[package]]
 name = "chrono"
@@ -2656,6 +2680,20 @@ dependencies = [
  "untrusted",
  "web-sys",
  "winapi",
+]
+
+[[package]]
+name = "rusoto_cloudformation"
+version = "0.47.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e00db4cfcfc14725c720d881443f2c17607bd80aa20fecd1382a5936cc2db05d"
+dependencies = [
+ "async-trait",
+ "bytes",
+ "futures",
+ "rusoto_core",
+ "serde_urlencoded",
+ "xml-rs",
 ]
 
 [[package]]
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -40,6 +40,7 @@ members = [
     "api/migration/migrations/v1.6.0/aws-control-container-v0-5-5",
     "api/migration/migrations/v1.6.0/public-admin-container-v0-7-4",
     "api/migration/migrations/v1.6.0/public-control-container-v0-5-5",
+    "api/migration/migrations/v1.6.2/add-cfsignal",
 
     "bottlerocket-release",
 
@@ -54,6 +55,8 @@ members = [
     "prairiedog",
 
     "metricdog",
+
+    "cfsignal",
 
     "logdog",
 
--- a/sources/api/migration/migrations/v1.6.2/add-cfsignal/Cargo.toml
+++ b/sources/api/migration/migrations/v1.6.2/add-cfsignal/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "add-cfsignal"
+version = "0.1.0"
+license = "Apache-2.0 OR MIT"
+edition = "2018"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers", version = "0.1.0"}
--- a/sources/api/migration/migrations/v1.6.2/add-cfsignal/src/main.rs
+++ b/sources/api/migration/migrations/v1.6.2/add-cfsignal/src/main.rs
@@ -1,0 +1,25 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a set of settings for configuring service cfsignal.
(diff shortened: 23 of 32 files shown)


webern:
Thank you for the PR @mello7tre! We are looking at it 👀 .

packages/os/cfsignal.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Send signal to CloudFormation Stack
+After=network-online.target settings-applier.service
+Requires=settings-applier.service
+Wants=network-online.target
+
+[Service]
+Type=simple
```
samuelkarp, line 11:
```suggestion
Type=oneshot
```

I think `oneshot` makes more sense for this service, even though we don't yet have any services that depend on it.

sources/cfsignal/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "cfsignal"
+version = "0.0.1"
+authors = ["Mello7tre <mello+github@ankot.org>"]
```
samuelkarp: While you're certainly welcome to list yourself as the author and provide your email address, we don't require you to do so.  If you prefer, you can omit the `authors` field entirely.

sources/cfsignal/README.tpl
```diff
@@ -0,0 +1,9 @@
+# {{crate}}
```
samuelkarp, line 1: It looks like the actual `README.md` is missing.  Can you add that to your commit as well?

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,78 @@
+#![deny(rust_2018_idioms, unused_imports)]
+
+/*!
+# Introduction
+
+CFSignal test
```
samuelkarp: It would be much appreciated if you wrote a bit about the cfsignal program here, as this would be the content that would end up in the `README.md`.

sources/models/shared-defaults/cf-signal.toml
```diff
@@ -0,0 +1,14 @@
+# CF Signal
+[settings.cloudformation]
+signal = false
+stack-name = "null"
+logical-resource-id = "null"
```
samuelkarp: These should probably be unset or `""` rather than `"null"`.

sources/models/src/lib.rs
```diff
@@ -242,6 +242,14 @@ struct MetricsSettings {
     service_checks: Vec<String>,
 }
 
+// CloudFormation settings
+#[model]
+struct CloudFormationSettings {
+    signal: bool,
```
samuelkarp: I'm wondering if we should call this `should_send` instead of `signal`.

The reason I'm thinking about that is that it might be useful in the future to have a setting that forces cfsignal to explicitly send `FAILURE` even if all the relevant systemd units were successful.  If, for example, there were a bootstrap container that was expected to execute some form of setup job and that failed, the bootstrap container could then indicate its failure all the way back out to CloudFormation.  In that scenario, it might be useful to have a `signal` setting with the values of `SUCCESS`/`FAILURE` and a `should_send` boolean setting that controls whether cfsignal sends anything.

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+#![deny(rust_2018_idioms)]
+
+mod cloudformation;
+mod config;
+mod error;
+mod service_check;
+
+use crate::config::Config;
+use crate::error::Result;
+//use crate::metricdog::Metricdog;
```
samuelkarp: It looks like these commented-out lines can be removed.

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// pub(crate) for testing.
+pub(crate) fn main_inner(service_check: Box<dyn ServiceCheck>) -> Result<()> {
+    // load the metricdog config file
+    let config = Config::from_file("/etc/cfsignal.toml")?;
+    /* let config = match &arguments.config {
+        None => Config::new()?,
+        Some(filepath) => Config::from_file(filepath)?,
+    };
+    */
```
samuelkarp:
```suggestion
    // load the cfsignal config file
    let config = Config::from_file("/etc/cfsignal.toml")?;
```

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+        Some(exit_code) => {
+            error!("System is degraded: {}", exit_code);
+        }
+    }
+
+    if let Err(err) = SignalResource::signal_resource(
+        config.stack_name,
+        config.logical_resource_id,
+        signal_status.to_owned(),
+        //signal_status,
```
samuelkarp:
```suggestion
        signal_status.to_owned(),
```

sources/cfsignal/src/service_check.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn system_running(&self) -> Result<ServiceHealth> {
+        if is_system_running()? {
+            return Ok(ServiceHealth {
+                is_healthy: true,
+                exit_code: None,
+            });
+        }
+        Ok(ServiceHealth {
+            is_healthy: false,
+            exit_code: Some(Ok(1)?),
```
samuelkarp: (I am not a Rust expert) Does this need `Ok` or can it just be `Some(1)` since the type is `Option<i32>`?

sources/cfsignal/src/service_check.rs
```diff
@@ -0,0 +1,64 @@
...
+}
+
+pub(crate) trait ServiceCheck {
+    fn system_running(&self) -> Result<ServiceHealth>;
+}
+
+pub(crate) struct SystemdCheck {}
+
+impl ServiceCheck for SystemdCheck {
+    fn system_running(&self) -> Result<ServiceHealth> {
```
samuelkarp: I am not a Rust expert, but since this always returns a `ServiceHealth` it might not need to be wrapped in a `Result`.

sources/cfsignal/src/service_check.rs
```diff
@@ -0,0 +1,64 @@
...
+            command: "systemctl",
+            args: args.iter().map(|&s| s.to_owned()).collect::<Vec<String>>(),
+        })?;
+    Ok(Outcome {
+        exit: output.status.code().unwrap_or(-1),
+        stdout: String::from_utf8_lossy(output.stdout.as_slice()).into(),
+    })
+}
+
+fn is_system_running() -> Result<bool> {
```
samuelkarp: Rather than returning `Result<bool>` here, it might be useful to return the exit code (`i32`?) so that it can be logged further up.  Using the `bool` here means that the `exit_code` field in `ServiceCheck` is always either `None` or `Some(Ok(1)?)`.

sources/cfsignal/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[dependencies]
+log = "0.4"
+serde = { version = "1.0.100", features = ["derive"] }
+simplelog = "0.10"
+snafu = { version = "0.6" }
+toml = "0.5.1"
+tokio = { version = "1", features = ["full"] }
+rusoto_core = { version = "0.47.0", default-features = false, features = ["rustls"] }
+rusoto_cloudformation = { version = "0.47.0", default-features = false, features = ["rustls"] }
+ec2_instance_metadata = "0.3"
```
samuelkarp: We have a existing [`imdsclient` crate](https://github.com/bottlerocket-os/bottlerocket/tree/develop/sources/imdsclient) which handles the differences between IMDSv1 and IMDSv2.  It would be better to use that crate versus taking a dependency on `ec2_instance_metadata`.

That crate doesn't yet have a method to grab the instance ID, but [this method](https://github.com/bottlerocket-os/bottlerocket/blob/44a10648318fbb187cc9f7dda6806b254cc77b97/sources/imdsclient/src/lib.rs#L106-L110) could be copied and the target changed to `meta-data/instance-id`.

samuelkarp (commented):
@mello7tre thank you so much for putting this together!  I took a look and had a few suggestions for changes.  Please let me know if you have any questions about the suggestions.  (And if you don't have the time to look at those suggestions, we can keep this open until you do or we can take a look at making those changes for you.)

packages/os/cfsignal.service
mello7tre: from systemd.service man page:
* oneshot
```
Behavior of oneshot is similar to simple; however, the service manager will consider the unit started
after the main process exits
```
* simple
```
If set to simple.... , the service manager will consider the unit started
immediately after the main service process has been forked off
```
i cannot wait that process exit, because i am calling systemctl with `--wait` option inside the program itself, i had not tested it but i think that boot process could remain stuck.

sources/cfsignal/src/service_check.rs
mello7tre: i am not rust expert too, as i wrote in main issue comment https://github.com/bottlerocket-os/bottlerocket/issues/1581#issuecomment-909325897, i am not rust programmer (this is first time that i put my hands on rust code)

sources/cfsignal/Cargo.toml
mello7tre: good to know, i will take a look and make the changes

mello7tre:
thanks @samuelkarp for the comments.
Give me some time and i will try to make the changes.
Mine only concern is about changing systemd.service Type, i have explained it better above.

packages/os/cfsignal.service
samuelkarp: > i cannot wait that process exit, because i am calling systemctl with --wait option inside the program itself, i had not tested it but i think that boot process could remain stuck.

Got it, thanks for the explanation!

sources/cfsignal/src/service_check.rs
webern: The `Ok(1)?` is synonymous with `1`, so it can be `exit_code: Some(1),`

packages/os/cfsignal.service
```diff
@@ -0,0 +1,13 @@
...
+RemainAfterExit=true
+ExecStart=/bin/cfsignal
+
+[Install]
+RequiredBy=preconfigured.target
```
arnaldo2792: This service should run almost at the very end of the boot sequence, right? If so, it should be `WantedBy=multi-user.target`. I think the weak dependency (`WantedBy`) should be sufficient since I don't think the `multi-user.target` should fail if this service fails.

Also could you please add `RefuseManualStart=true` and `RefuseManualStop=true`, I think it makes sense to set them  up, since as  you described `simple` is  a `oneshot`,  and we try to use these settings in all our `oneshot` services.

mello7tre: no, the service must be run as soon as network is available and bottlerocket write the configuration files [settings-applier.service] after having read the user-data (we need them for stack-name and logical-resource-id).
This way we can intercept more boot fail causes and signal in advance cloudformation stack that ASG creation failed (without need to wait for the timeout)

Cfsignal invoke `systemctl is-system-running` command using `--wait` option, this is the key to know if boot is successful or not.

Looking at other services unit files i see that when `RefuseManualStart` is used there is a comment like:
```
# unexpected state
```
i do not think that reinvoking cfsignal could cause a problem.

But you are right, the weak dependency `WantedBy` could be sufficient. Will change it.

bcressey: It needs to run early in the boot sequence, so that if we fail to transition to `configured.target` or `multi-user.target`, it can detect that failure and send the appropriate signal.

bcressey: `RefuseManualStart=true` and `RefuseManualStop=true` shouldn't be needed with the change to `WantedBy` that came in after your comment 😁

sources/cfsignal/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+log = "0.4"
+serde = { version = "1.0.100", features = ["derive"] }
+simplelog = "0.10"
+snafu = { version = "0.6" }
+toml = "0.5.1"
+tokio = { version = "1", features = ["full"] }
```
bcressey: This should be pinned to 1.8, the current LTS release for `tokio`, and should turn off any features that aren't needed:
```suggestion
tokio = { version = "~1.8", default-features = false, features = ["macros", "rt-multi-thread"] }  # LTS
```

sources/models/shared-defaults/cf-signal.toml
```diff
@@ -0,0 +1,14 @@
+# CF Signal
+[settings.cloudformation]
+should-signal = false
+stack-name = ""
+logical-resource-id = ""
+
+[services.cfsignal]
+configuration-files = ["cfsignal-toml"]
+restart-commands = []
```
bcressey: In the absence of a restart command, changes to settings won't cause the running service to be restarted.

There's only a limited window of time when this could matter, but someone might want to use a bootstrap container to configure settings, and it'd be nice if it worked.

sources/models/shared-defaults/cf-signal.toml
```diff
@@ -0,0 +1,14 @@
+# CF Signal
```
bcressey: You can omit this comment since it doesn't add much context beyond the file name.

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,80 @@
+#![deny(rust_2018_idioms, unused_imports)]
+
+/*!
+## Introduction
+
+Cfsignal is a sort of cfn-signal [https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html].
+
+Cloudformation AutoscalingGroup can be configured to wait, during creation, for the proper number of signal by his instances before considere the ASG successfully created.
+https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html
+https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html
```
bcressey:
```suggestion
When creating an Auto Scaling Group, CloudFormation can be configured to wait for the expected number of signals from instances before considering the ASG successfully created. See [CreationPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html) and [UpdatePolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html) for more details.
```

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,80 @@
+#![deny(rust_2018_idioms, unused_imports)]
+
+/*!
+## Introduction
+
+Cfsignal is a sort of cfn-signal [https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html].
```
bcressey:
```suggestion
Cfsignal is similar to [cfn-signal] (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html).
```

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+
+Cloudformation AutoscalingGroup can be configured to wait, during creation, for the proper number of signal by his instances before considere the ASG successfully created.
+https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html
+https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html
+
+Cfsingal use `systemctl` command `is-system-running` to find out the AutoscalingGroup instances boot proccess status, and send the relative signal to the Cloudformation stack.
```
bcressey:
```suggestion
Cfsignal uses `systemctl is-system-running` to determine whether the boot has succeeded or failed, and sends the corresponding signal to the CloudFormation stack.
```

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+
+## configuration
+
+Cfsignal is configured using the following settings:
+* `settings.cloudformation.should-signal`: Check status and send signal. Defaults to `false`.
+* `settings.cloudformation.stack-name`: Name of the stack to which send signal.
```
bcressey:
```suggestion
* `settings.cloudformation.stack-name`: Name of the stack to signal.
```

sources/cfsignal/src/config.rs
```diff
@@ -0,0 +1,21 @@
...
+    pub(crate) stack_name: String,
+    pub(crate) logical_resource_id: String,
+}
+
+impl Config {
+    pub(crate) fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let path = path.as_ref();
+        let s = fs::read_to_string(path).context(error::ConfigRead { path })?;
+        let config: Config = toml::from_str(&s).context(error::ConfigParse { path })?;
+        Ok(config)
```
bcressey: Should be able to simplify this to just:
```suggestion
        toml::from_str(&s).context(error::ConfigParse { path })
```

sources/cfsignal/src/error.rs
```diff
@@ -0,0 +1,37 @@
+//! Provides the list of errors for `metricdog`.
```
bcressey:
```suggestion
//! Provides the list of errors for `cfsignal`.
```

sources/cfsignal/src/error.rs
```diff
@@ -0,0 +1,37 @@
...
+        path: PathBuf,
+        source: toml::de::Error,
+    },
+
+    #[snafu(display("Failed to read config file {}: {}", path.display(), source))]
+    ConfigRead {
+        path: PathBuf,
+        source: std::io::Error,
+    },
+    #[snafu(display("IMDS request failed: {}", source))]
```
bcressey: nit: empty line between variants:
```suggestion
    },

    #[snafu(display("IMDS request failed: {}", source))]
```

sources/cfsignal/src/cloudformation.rs
```diff
@@ -0,0 +1,65 @@
...
+use rusoto_core::Region;
+use snafu::OptionExt;
+use snafu::ResultExt;
+use std::str::FromStr;
+
+pub(crate) struct SignalResource {}
+
+impl SignalResource {
+    #[tokio::main]
+    pub(crate) async fn signal_resource(
```
bcressey: Since we don't do anything with the `SignalResource` struct, it'd be better to just use a function - then we can call `signal_resource(...)` instead of `SignalResource::signal_resource(...)`.

It's also a little odd to bury `#[tokio::main]` in a module like this - ideally it'd be on the `main` function.

sources/cfsignal/src/cloudformation.rs
```diff
@@ -0,0 +1,65 @@
+use crate::error::{self, Result};
+use imdsclient::ImdsClient;
+use log::info;
+use rusoto_cloudformation::{
+    CloudFormation, CloudFormationClient, SignalResourceInput,
+};
+use rusoto_core::Region;
+use snafu::OptionExt;
+use snafu::ResultExt;
```
bcressey:
```suggestion
use snafu::{OptionExt, ResultExt};
```

One of these commands should catch or fix stuff like this:
```
cargo check -p cfsignal
cargo clippy -p cfsignal -- --no-deps
cargo fmt -p cfsignal
```

sources/cfsignal/src/cloudformation.rs
```diff
@@ -0,0 +1,65 @@
...
+        };
+
+        match client.signal_resource(signal_resource_input).await {
+            Ok(_output) => (),
+            Err(error) => {
+                println!("Error: {:?}", error);
+            }
+        }
+
+        Ok(())
```
bcressey: This should be collapsed down to just:
```
client.signal_resource(signal_resource_input).await.context(error::SignalResource)
```

If we want to just print the error and exit zero, we can do that in `main` - our standard boilerplate for this is:
```
#[tokio::main]
async fn main() {
    if let Err(e) = run().await {
        eprintln!("{}", e);
        // could make this `0` if we really want to ignore errors
        process::exit(1); 
    }
}
```

sources/cfsignal/src/system_check.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub(crate) struct SystemHealth {
+    /// Whether or not the system is healthy.
+    pub(crate) is_healthy: bool,
+    /// The system status description.
+    pub(crate) status: String,
+    /// The system status code.
+    pub(crate) exit_code: i32,
+}
```
bcressey, line 16: @webern do you have thoughts on making it easier to reuse this functionality from `metricdog`?

It seems suboptimal to have it almost but not quite the same - especially since the added abstraction here isn't supported by unit tests.

mello7tre:
just pushed the requested changes.
I need more time to work on:
https://github.com/bottlerocket-os/bottlerocket/pull/1728#discussion_r712547250
https://github.com/bottlerocket-os/bottlerocket/pull/1728#discussion_r712553087

will came later...

sources/cfsignal/src/cloudformation.rs
mello7tre: just for info, none of the above catch that issue.
(but using `clippy` i found other suggestions, for other part of code, to adopt)

sources/cfsignal/src/system_check.rs
webern: > do you have thoughts on making it easier to reuse this functionality from metricdog?

I think the `SystemHealth` abstraction could probably be eliminated from this PR, or at least renamed in a way that makes it clear it's not the same thing as `SystemHealth` in `metricdog`.

We could have a library for querying systemd, I guess, but it seems like the shape of the queries are different enough that it would be an aggregating library rather than a code re-use library.

mello7tre: but the name are not already different ?
metricdog use: `ServiceHealth`
cfsignal use: `SystemHealth`

or you are saying something different ?

webern: Sorry, I didn't realize the name was already different.

jfrconley:
Is there any movement on this issue? Waiting to adopt Bottlerocket until I can be sure that my asg will wait for instances to register with ECS

jpculp:
All, we're sorry this has been left open for quite some time. From what I can tell it seems to be on the right track. @mello7tre, would you be able to re-base, squash the commits, and request re-review?

mello7tre:
Rebased as requested.

etungsten:
Hi @mello7tre, sorry about the delay. I'm gonna do my best and help you get this over the finish line. I took a look over the existing comments and resolved the ones that have been addressed.

The only comment that still needs addressing is https://github.com/bottlerocket-os/bottlerocket/pull/1728#discussion_r712553087.

Another thing is we want to make sure `cfsignal` only runs on first boot as opposed to on every boot. We can achieve this by making `cfsignal` create an sentinel file at some path and the service unit can conditionally run based on the file's presence. The `early-boot-config` service is an example that does this:
https://github.com/bottlerocket-os/bottlerocket/blob/9749af9973e4494768c2fac563172c753b6bc011/packages/os/early-boot-config.service#L9-L11
And the sentinel file is created like so:
https://github.com/bottlerocket-os/bottlerocket/blob/9749af9973e4494768c2fac563172c753b6bc011/sources/api/early-boot-config/src/main.rs#L144-L149
Let me know what you think and how I can help.

And finally, since we're introducing new settings, we need to create a simple migration binary to migrate our settings datastore between OS versions. You can read more about migrations [here](https://github.com/bottlerocket-os/bottlerocket/tree/develop/sources/api/migration) and how to write them[ here](https://github.com/bottlerocket-os/bottlerocket/tree/9749af9973e4494768c2fac563172c753b6bc011/sources/api/migration#how-to-write-migrations). The closest example would be the `add-network` migration here:
https://github.com/bottlerocket-os/bottlerocket/blob/9749af9973e4494768c2fac563172c753b6bc011/sources/api/migration/migrations/archived/v1.0.5/add-network-settings/src/main.rs#L7-L15 
Of course, we can take care of this for you if you'd prefer not to dabble with it. Just let us know. Thanks again for your contribution!

mello7tre:
Hi @etungsten, just pushed the requested changes.
Have a look if all is right.

sources/cfsignal/src/cloudformation.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    let client = CloudFormationClient::new(Region::from_str(&region).unwrap());
+    let signal_resource_input = SignalResourceInput {
+        stack_name,
+        logical_resource_id,
+        status,
+        unique_id: instance_id,
+    };
+
+    client.signal_resource(signal_resource_input).await.context(error::SignalResource);
```
etungsten: Need to handle the error. This error eventually floats up to main and gets printed out
```suggestion
    client.signal_resource(signal_resource_input).await.context(error::SignalResource)?;
```

sources/cfsignal/src/cloudformation.rs
```diff
@@ -0,0 +1,55 @@
...
+
+pub async fn signal_resource(
+    stack_name: String,
+    logical_resource_id: String,
+    status: String,
+) -> Result<()> {
+    info!("Connecting to IMDS");
+    let mut client = ImdsClient::new();
+    let instance_id = get_instance_id(&mut client).await.unwrap();
+    let region = get_region(&mut client).await.unwrap();
```
etungsten: `unwrap` would cause the program to panic when errors occur, let's try avoiding that.

We can just float these errors up to the caller if they do occur.
```suggestion
    let instance_id = get_instance_id(&mut client).await?;
    let region = get_region(&mut client).await?;
```

sources/cfsignal/src/cloudformation.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    info!(
+        "Region: {:?} - InstanceID: {:?} - Signal: {:?}",
+        region, instance_id, status
+    );
+
+    let client = CloudFormationClient::new(Region::from_str(&region).unwrap());
```
etungsten: We need a new error variant added to `cfsignal/src/error.rs`:
```
    #[snafu(display("Unable to parse '{}' as a region: {}", region, source))]
    RegionParse {
        region: String,
        source: ParseRegionError,
    },
```
Then
```suggestion
    let client = CloudFormationClient::new(Region::from_str(&region).context(error::RegionParse { region })?);
```

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+use crate::error::Result;
+use crate::system_check::SystemCheck;
+use cloudformation::signal_resource;
+use log::LevelFilter;
+use log::{error, warn, info};
+use simplelog::{Config as LogConfig, SimpleLogger};
+use std::fs;
+use std::process;
+
+// We only want to run early-boot-config once, at first boot.  Our systemd unit file has a
```
etungsten:
```suggestion
// We only want to run cfsignal once, at first boot.  Our systemd unit file has a
```

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+    let system_check = Box::new(SystemCheck {});
+    let system_status = system_check.system_running()?;
+    info!(
+        "System status is: {} [{}]",
+        system_status.status, system_status.exit_code
+    );
+
+    // run only if the opt-in flag is set
+    if config.should_signal {
+
```
etungsten: Since we only do anything with the system check status if `should_signal` is set, we can have the check inside the if block.
```suggestion
    // run only if the opt-in flag is set
    if config.should_signal {
        let system_check = Box::new(SystemCheck {});
        let system_status = system_check.system_running()?;
        info!(
            "System status is: {} [{}]",
            system_status.status, system_status.exit_code
        );
```

packages/os/cfsignal.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Send signal to CloudFormation Stack
+After=network-online.target settings-applier.service
+Requires=settings-applier.service
+Wants=network-online.target
+# We only want to run once, at first boot.  This file is created by early-boot-config
```
etungsten:
```suggestion
# We only want to run once, at first boot.  This file is created by cfsignal
```

sources/api/migration/migrations/archived/v1.2.0/cfsignal/src/main.rs
```diff
@@ -0,0 +1,23 @@
...
+use migration_helpers::common_migrations::AddSettingsMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a set of settings for configuring service cfsignal.
+/// Remove the whole `settings.cloudformation` prefix if we downgrade.
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec![
+        "settings.cloudformation",
+    ]))
```
etungsten: There are a few more prefixes that gets added.

```suggestion
    migrate(AddPrefixesMigration(vec![
        "settings.cloudformation",
        "services.cfsignal",
        "configuration-files.cfsignal-toml",
    ]))
```

sources/api/migration/migrations/archived/v1.2.0/cfsignal/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "cfsignal"
```
etungsten: This migration crate should go into a separate directory.
We're currently on v1.6.0. We should rebase onto the latest develop so we can see the new migrations.
Since this feature would be introduced in the next minor version, the path of this migration crate should be:
`sources/api/migration/migrations/v1.7.0/cfsignal/`

I also recommend renaming this crate to `add-cfsignal` to not get mixed up with the actual `cfsignal` crate.

This migration also needs to be listed in `Release.toml` and `sources/Cargo.toml` along with the other migrations. Here's an example of what the two changes look like for any singular migration: https://github.com/bottlerocket-os/bottlerocket/pull/1898/commits/082e231c70f2d36a64b3edf6be25e47395e87da6

etungsten (changes requested):
Looking good. We need to do a rebase to fix some minor conflicts and fix the migration crate path.

sources/cfsignal/src/main.rs
mello7tre: Could be useful to know which would be the system status even without signaling it to CloudFormation, maybe to try it out in a `dry mode` or to debug.

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+/*!
+## Introduction
+
+Cfsignal is similar to [cfn-signal] (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html).
+
+When creating an Auto Scaling Group, CloudFormation can be configured to wait for the expected number of signals from instances before considering the ASG successfully created. See [CreationPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html) and [UpdatePolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html) for more details.
+
+Cfsignal uses `systemctl is-system-running` to determine whether the boot has succeeded or failed, and sends the corresponding signal to the CloudFormation stack.
+
+## configuration
```
etungsten:
```suggestion
## Configuration
```

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+
+Cfsignal is configured using the following settings:
+* `settings.cloudformation.should-signal`: Check status and send signal. Defaults to `false`.
+* `settings.cloudformation.stack-name`: Name of the stack to signal.
+* `settings.cloudformation.logical-resource-id`: The logical ID of the AutoScalingGroup resource that you want to signal.
```
etungsten: It would be nice if these settings also get documented in the main Bottlerocket README's settings section here between "Amazon ECS Settings"  and "OCI Hooks settings":
https://github.com/bottlerocket-os/bottlerocket/blob/9749af9973e4494768c2fac563172c753b6bc011/README.md?plain=1#L420-L422

Something like:
```
#### CloudFormation signal helper settings

For AWS variants, these settings allows you to set up CloudFormation signalling to indicate whether Bottlerocket hosts running in EC2 have been successfully created or updated:
* `settings.cloudformation.should-signal`: Whether to check system status and send signal. Defaults to `false`.
* `settings.cloudformation.stack-name`: Name of the stack to signal.
* `settings.cloudformation.logical-resource-id`: The logical ID of the AutoScalingGroup resource that you want to signal.
```

Then the README here in `cfsignal` can reference the settings via a link and instead describe the toml file settings:
```suggestion
Configuration is read from a TOML file, which is generated from Bottlerocket settings:
* `should_signal`: Whether to check system status and send signal.
* `stack_name`: Name of the CFN stack to signal.
* `logical_resource_id`: The logical ID of the AutoScalingGroup resource that you want to signal.
```

Feel free to play around with the wording a bit.

mello7tre:
applied requested changes
rebased on aeec3aa
squashed commits

mello7tre:
i made a little mess rebasing, so i fixed it and force pushed the right "version"

Release.toml
```diff
@@ -1,4 +1,4 @@
-version = "1.6.0"
+version = "1.7.0"
```
etungsten: Let's not bump this version tag for now. We bump this when we make a new release.
```suggestion
version = "1.6.0"
```

sources/cfsignal/README.md
```diff
@@ -0,0 +1,24 @@
...
+When creating an Auto Scaling Group, CloudFormation can be configured to wait for the expected number of signals from instances before considering the ASG successfully created. See [CreationPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-creationpolicy.html) and [UpdatePolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html) for more details.
+
+Cfsignal uses `systemctl is-system-running` to determine whether the boot has succeeded or failed, and sends the corresponding signal to the CloudFormation stack.
+
+### configuration
+
+Configuration is read from a TOML file, which is generated from Bottlerocket settings:
+`should_send`: Whether to check system status and send signal.
+`stack_name`: Name of the CloudFormation Stack to signal.
+`logical_resource_id`: The logical ID of the AutoScalingGroup resource that you want to signal.
```
etungsten: These edits should go into the top of `cfsignal/src/main.rs` since that is where this README.md is generated from.
After triggering the `build.rs` script via a `cargo check` or build, this README file gets generated. So both places should have the same edits.

sources/cfsignal/src/system_check.rs
```diff
@@ -0,0 +1,60 @@
...
+
+struct Outcome {
+    exit: i32,
+    stdout: String,
+}
+
+impl Outcome {
+    fn is_exit_true(&self) -> bool {
+        self.exit == 0
+    }
```
etungsten: nit: I'm wondering if this is more accurate.
```suggestion
    fn is_success(&self) -> bool {
        self.exit == 0
    }
```

sources/cfsignal/src/error.rs
```diff
@@ -0,0 +1,47 @@
...
+    ConfigRead {
+        path: PathBuf,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("IMDS request failed: {}", source))]
+    ImdsRequest { source: imdsclient::Error },
+
+    #[snafu(display("IMDS client failed: {}", source))]
+    ImdsClient { source: imdsclient::Error },
```
etungsten: Is this error variant being used? I couldn't find any reference to it. Can we remove it?

etungsten (changes requested):
Nice! I'm gonna start asking others to take a look.

https://github.com/bottlerocket-os/bottlerocket/pull/1728/files#r803845979 still needs a quick fix.

sources/api/migration/migrations/v1.7.0/add-cfsignal/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "cfsignal"
```
etungsten: The migration crate name needs to be updated
```suggestion
name = "add-cfsignal"
```

etungsten (dismissed):

mello7tre:
Hi @etungsten, i think that to speed up the development is better if, from now on, you directly make all relevant changes.
So feel free to change it as you prefer, i just would like that this feature will be included in a future release.

Regards, Alberto.

etungsten:
Push above addresses comments. Needed to label `cfsignal` with `api_exec` so it can write the sentinel file to a persistent storage location like `/var/lib/bottlerocket`.
Made sure cfsignal only runs on first boot. On subsequent boots, the cfsignal service unit never starts:
```
Feb 11 21:13:21 systemd[1]: Condition check resulted in Send signal to CloudFormation Stack being skipped.
```

README.md
```diff
@@ -419,6 +419,13 @@ These settings can be changed at any time.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
 * `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
 
+#### CloudFormation signal helper settings
+
+For AWS variants, these settings allows you to set up CloudFormation signalling to indicate whether Bottlerocket hosts running in EC2 have been successfully created or updated:
+* `settings.cloudformation.should-signal`: Whether to check status and send signal. Defaults to `false`.
+* `settings.cloudformation.stack-name`: Name of the CloudFormation Stack to signal.
+* `settings.cloudformation.logical-resource-id`: The logical ID of the AutoScalingGroup resource that you want to signal.
```
webern, line 427: Are both of these required? Will the instance will fail if `should-signal` is `true` and either of these is missing? Maybe document the behavior/dependency if there is one.

sources/cfsignal/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[dependencies]
+log = "0.4"
+serde = { version = "1.0", features = ["derive"] }
+simplelog = "0.11"
+snafu = { version = "0.7" }
+toml = "0.5.1"
+tokio = { version = "~1.14", default-features = false, features = ["macros", "rt-multi-thread"] }
+rusoto_core = { version = "0.47.0", default-features = false, features = ["rustls"] }
+rusoto_cloudformation = { version = "0.47.0", default-features = false, features = ["rustls"] }
```
webern, line 18: Should we switch to the AWS SDK now instead of later?

sources/cfsignal/src/system_check.rs
```diff
@@ -0,0 +1,60 @@
...
+        result.stdout.pop();
+        Ok(SystemHealth {
+            is_healthy: result.is_success(),
+            status: result.stdout,
+            exit_code: result.exit,
+        })
+    }
+}
+
+struct Outcome {
```
webern, line 33: These abstractions are perhaps not as useful here as they are in metricdog (which inspired them). They are harmless though so I don't see a need to eliminate these structs and functions.

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+    let mut signal_status = "FAILURE";
+    let system_check = Box::new(SystemCheck {});
+    let system_status = system_check.system_running()?;
+    info!(
+        "System status is: {} [{}]",
+        system_status.status, system_status.exit_code
+    );
+
+    // run only if the opt-in flag is set
+    if config.should_signal {
```
webern, line 60: Tiny nit (unless you feel that the marker file needs to be written even when should_signal is false):

```suggestion
    // run only if the opt-in flag is set
    if !config.should_signal {
        return Ok((())
    }
```

webern (approved):
:guitar:

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,94 @@
+#![deny(rust_2018_idioms, unused_imports)]
+
+/*!
+## Introduction
+
+Cfsignal is similar to [cfn-signal] (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html).
```
arnaldo2792: nit: extra space

```suggestion
Cfsignal is similar to [cfn-signal](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-signal.html).
```

sources/cfsignal/Cargo.toml
etungsten: I would rather do the switch all at once instead of having to build crates for two different SDKs.

sources/cfsignal/src/main.rs
etungsten: I think we just want to run this only once regardless of what `should_signal` is initially. For the use-case of signalling cloudformation stacks, the expectation is that this setting is set in userdata in the ASG's launch template. So it should only take effect once. If it wasn't set on first boot, we should just never waste time running this service.

README.md
mello7tre: default value for both of them is an empty string (`sources/models/shared-defaults/cf-signal.toml`), so you will get an error like (this is for python):
```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid length for parameter StackName, value: 0, valid min length: 1
```

sources/cfsignal/Cargo.toml
mello7tre: https://github.com/bottlerocket-os/bottlerocket/issues/1581#issuecomment-909325897

sources/cfsignal/src/main.rs
mello7tre: totally agree with @etungsten

cbgbt (approved):

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+// We create it after running successfully.
+const MARKER_FILE: &str = "/var/lib/bottlerocket/cfsignal.ran";
+
+/// pub(crate) for testing.
+async fn run() -> Result<()> {
+    SimpleLogger::init(LevelFilter::Info, LogConfig::default())
+        .expect("unable to configure logger");
+
+    // load the cfsignal config file
+    let config = Config::from_file("/etc/cfsignal.toml")?;
```
arnaldo2792: Nit: this should be a constant

sources/cfsignal/src/system_check.rs
```diff
@@ -0,0 +1,60 @@
...
+
+impl Outcome {
+    fn is_success(&self) -> bool {
+        self.exit == 0
+    }
+}
+
+fn systemctl(args: &[&str]) -> Result<Outcome> {
+    trace!("calling systemctl with '{:?}'", args);
+    let output = Command::new("systemctl")
```
arnaldo2792: Nit: `systemctl` should be a constant with the full path to the binary

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+    );
+
+    // run only if the opt-in flag is set
+    if config.should_signal {
+        match system_status.is_healthy {
+            true => {
+                signal_status = "SUCCESS";
+            }
+            false => (),
+        }
```
arnaldo2792: `signal_status` already defaults to `FAILURE`, you can do this instead, and save yourself the extra branch

```suggestion
        if systemd_status.is_healthy {
            signal_status = "SUCCESS";
        }
```

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+        )
+        .await
+        {
+            error!("Error while sending signal: {}", err);
+        }
+    }
+
+    fs::write(MARKER_FILE, "").unwrap_or_else(|e| {
+        warn!(
+            "Failed to create marker file {}, may unexpectedly run again: {}",
```
arnaldo2792: Nit: I think we use `''` around logged values like file names.

sources/imdsclient/src/lib.rs
```diff
@@ -157,6 +157,12 @@ impl ImdsClient {
         self.fetch_string(&instance_type_target).await
     }
 
+    /// Gets the instance-id from instance metadata.
+    pub async fn fetch_instance_id(&mut self) -> Result<Option<String>> {
+        let instance_type_target = "meta-data/instance-id";
```
arnaldo2792, line 162: Nit: Another potential candidate for a constant

sources/models/src/aws-k8s-1.21/mod.rs
```diff
@@ -4,9 +4,9 @@ use std::collections::HashMap;
 
 use crate::modeled_types::Identifier;
 use crate::{
-    AwsSettings, BootstrapContainer, HostContainer, KernelSettings, KubernetesSettings,
```
arnaldo2792, line 7: I'm putting this here, since there isn't another place. Shouldn't you add this setting to NVIDIA variants as well?

sources/cfsignal/src/main.rs
arnaldo2792, line 60: I think the call to `system_check.system_running()` should happen inside this block, that way, the service exits earlier if it doesn't have to send a signal.

sources/cfsignal/src/cloudformation.rs
```diff
@@ -0,0 +1,60 @@
+use crate::error::{self, Result};
+use imdsclient::ImdsClient;
+use log::info;
+use rusoto_cloudformation::{CloudFormation, CloudFormationClient, SignalResourceInput};
+use rusoto_core::Region;
+use snafu::{OptionExt, ResultExt};
+use std::str::FromStr;
+
```
arnaldo2792, line 8: Nit: Missing comment, to be consistent with the other two functions

packages/os/cfsignal.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Send signal to CloudFormation Stack
+After=network-online.target settings-applier.service
+Requires=settings-applier.service
+Wants=network-online.target
```
arnaldo2792, line 5: Why doesn't the unit declare a stronger dependency on `network-online.target`?

packages/os/cfsignal.service
```diff
@@ -0,0 +1,15 @@
...
+# We only want to run once, at first boot.  This file is created by cfsignal
```
arnaldo2792: Nit: extra space:

```suggestion
# We only want to run once, at first boot. This file is created by cfsignal
```

packages/os/cfsignal.service
```diff
@@ -0,0 +1,15 @@
...
+# after a successful run.
+ConditionPathExists=!/var/lib/bottlerocket/cfsignal.ran
+
+[Service]
+Type=simple
+ExecStart=/bin/cfsignal
```
arnaldo2792: Nit: in other units we use the full path:

```suggestion
ExecStart=/usr/bin/cfsignal
```

sources/cfsignal/src/error.rs
```diff
@@ -0,0 +1,46 @@
...
+#[snafu(visibility(pub(super)))]
+pub enum Error {
+    #[snafu(display("Command '{}' with args '{:?}' failed: {}", command, args, source))]
+    Command {
+        command: String,
+        args: Vec<String>,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("Failed to parse config file {}: {}", path.display(), source))]
```
arnaldo2792, line 18: Nit: Use `' '` around templated values that aren't error messages

sources/cfsignal/src/main.rs
```diff
@@ -0,0 +1,94 @@
...
+use crate::error::Result;
+use crate::system_check::SystemCheck;
+use cloudformation::signal_resource;
+use log::LevelFilter;
+use log::{error, info, warn};
+use simplelog::{Config as LogConfig, SimpleLogger};
+use std::fs;
+use std::process;
+
+// We only want to run cfsignal once, at first boot.  Our systemd unit file has a
```
arnaldo2792, line 37: Nit: extra space

```suggestion
// We only want to run cfsignal once, at first boot. Our systemd unit file has a
```

sources/models/shared-defaults/cf-signal.toml
```diff
@@ -0,0 +1,13 @@
+[settings.cloudformation]
+should-signal = false
+stack-name = ""
+logical-resource-id = ""
+
+[services.cfsignal]
+configuration-files = ["cfsignal-toml"]
+restart-commands = ["/bin/systemctl try-restart cfsignal.service"]
```
arnaldo2792, line 8: Do you need a restart command, given that the service should run only once? I might be wrong but in other services that require a reload after a setting change in the API like the docker daemon we have the `restart-command` to actually trigger a realod. But for this service, since there isn't any `affected-services` setting, do you need it at all? Could it be empty?

sources/imdsclient/src/lib.rs
etungsten: Looking at how the other `fetch...` functions are structured in this library, I would like to stick to the `imdsclient` convention here and just keep it the same.

sources/cfsignal/src/main.rs
etungsten: See reasoning here: https://github.com/bottlerocket-os/bottlerocket/pull/1728#discussion_r803858324

packages/os/cfsignal.service
etungsten: Since there's no need to. `Wants=` + `After=` is sufficient to express "start this service after network is up". This is the same across for all our other network-dependent services.

sources/models/shared-defaults/cf-signal.toml
etungsten: See https://github.com/bottlerocket-os/bottlerocket/pull/1728#discussion_r712282307

etungsten: I'll add the `metadata.settings.cloudformation` `affected-services`.

etungsten:
Push above and below addresses @arnaldo2792 's comments.

Tested changes and it works as expected:
```
bash-5.0# systemctl status cfsignal
● cfsignal.service - Send signal to CloudFormation Stack
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/cfsignal.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Thu 2022-02-24 03:20:40 UTC; 10min ago
    Process: 1578 ExecStart=/usr/bin/cfsignal (code=exited, status=0/SUCCESS)
   Main PID: 1578 (code=exited, status=0/SUCCESS)

Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] System status is: running [0]
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] Connecting to IMDS
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] Received meta-data/instance-id
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] Received dynamic/instance-identity/document
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal cfsignal[1578]: 03:20:40 [INFO] Region: "us-west-2" - InstanceID: "i-051f097c469cd6a7f" - Signal: "SUCCESS"
Feb 24 03:20:40 ip-192-168-16-4.us-west-2.compute.internal systemd[1]: cfsignal.service: Succeeded.
```

arnaldo2792 (approved):
:duck:

cbgbt (approved):

zmrow (approved):
:ice_cream: 

Thanks!

README.md
```diff
@@ -419,6 +419,13 @@ These settings can be changed at any time.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
 * `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
 
+#### CloudFormation signal helper settings
+
+For AWS variants, these settings allows you to set up CloudFormation signalling to indicate whether Bottlerocket hosts running in EC2 have been successfully created or updated:
+* `settings.cloudformation.should-signal`: Whether to check status and send signal. Defaults to `false`. If set to `true`, both `stack-name` and `logical-resource-id` needs to be specified.
```
bcressey: nit:
```suggestion
For AWS variants, these settings allow you to set up CloudFormation signaling to indicate whether Bottlerocket hosts running in EC2 have been successfully created or updated:
* `settings.cloudformation.should-signal`: Whether to check status and send signal. Defaults to `false`. If set to `true`, both `stack-name` and `logical-resource-id` need to be specified.
```

sources/models/src/aws-k8s-1.21/defaults.d/60-cf-signal.toml
```diff
@@ -0,0 +1 @@
+../../../shared-defaults/cf-signal.toml
```
bcressey: nit: "60" overlaps with the lockdown config. We could make this "25" rather than "60", to group it with `20-aws-host-containers.toml`, or else make it "80" to start a new group of settings.

bcressey (approved):
Nice! This looks ready to merge.

@mello7tre, I know this has been a long time coming! I really appreciate the contribution as well as your extraordinary patience. We're planning to ship it in the next release.

mello7tre:
> Nice! This looks ready to merge.
> 
> @mello7tre, I know this has been a long time coming! I really appreciate the contribution as well as your extraordinary patience. We're planning to ship it in the next release.

Thanks, i am glad to have given a little contribution to the project.

etungsten:

etungsten:

bcressey (approved):

etungsten:
Push above moves the migration to the `v1.6.1` to `v1.6.2` migration chain since we're planning this for v1.6.2 release.

etungsten:
Tested migration again and it still works as expected

Issue #1651, opened 2021-07-09
GitLab Runner - SELinux Issue


bdwyertech:
**Image I'm using:**
Latest v1.1.2 on EKS 1.20
Also tried going back to EKS 1.19 back to v1.0.x

**What I expected to happen:**
GitLab Runner should be able to create pods in my cluster.

**What actually happened:**
GitLab Runner was not able to create pods within my cluster.

**How to reproduce the problem:**
Deploy the gitlab/gitlab-runner helm chart to test.

It seems like there is an SELinux issue going on here preventing the pod from pulling some of its containers.  I think it means its attempting to mount an unlabeled_t tmpfs underneath /local (of type local_t).  
```
AVC avc:  denied  { associate } for  pid=3531 comm="containerd" name="bin" dev="nvme1n1p1" ino=398203 scontext=system_u:object_r:unlabeled_t:s0 tcontext=system_u:object_r:local_t:s0 tclass=filesystem permissive=0
```

```shell
Jul 08 23:15:20 ip-10-12-34-56.ec2.internal containerd[3531]: time="2021-07-08T23:15:20.688955107Z" level=info msg="RunPodSandbox for &PodSandboxMetadata{Name:runner-byxzzoom-project-9754-concurrent-0z4zg6,Uid:b7593111-1d5f-4b49-8f6d-a6fbd99caf83,Namespace:default,Attempt:0,} returns sandbox id \"f9e509c805a42f3407d4d01af96df4b40cf24308d61c1d1b2bdc2b1fde070ad6\""
Jul 08 23:15:20 ip-10-12-34-56.ec2.internal containerd[3531]: time="2021-07-08T23:15:20.693249544Z" level=info msg="PullImage \"registry.gitlab.com/gitlab-org/gitlab-runner/gitlab-runner-helper:x86_64-c1edb478\""
Jul 08 23:15:25 ip-10-12-34-56.ec2.internal audit[3531]: AVC avc:  denied  { associate } for  pid=3531 comm="containerd" name="bin" dev="nvme1n1p1" ino=398203 scontext=system_u:object_r:unlabeled_t:s0 tcontext=system_u:object_r:local_t:s0 tclass=filesystem permissive=0
Jul 08 23:15:25 ip-10-12-34-56.ec2.internal systemd[1]: local-var-lib-containerd-tmpmounts-containerd\x2dmount075234526.mount: Succeeded.
[...]

Issue #1712, opened 2021-08-18
enable MCS isolation in the SELinux policy


bcressey:
Currently we allocate random MCS pairs to unprivileged containers, thanks to support in `containerd` and the update to 1.4 in #1336. However, our SELinux policy lacks the constraints required to enforce an isolation boundary.

That means that a process running with the level `s0:c1,c2` can read and write files with the level `s0:c3,c4`, even though those files are nominally owned by a process in a different container.

This will add an additional boundary between containers that supplements the use of namespaces for isolation.

Issue #1617, opened 2021-06-16
Provide settings to change containerd's persistent storage


arnaldo2792:
**What I'd like:**

Originally posted in #1209

> I'd also like to be able to use this storage (ephemeral disks) for the images pulled by the container runtime.

Containerd will store container images in the persistent data storage which is setup with the `Root` property in `/etc/containerd/config.toml`. We need to provide a mechanism to allow users to use ephemeral disks as containerd's persistent data storage.

**Any alternatives you've considered:**
With #1601, users can set up and mount the ephemeral disks attached to the host. However, they can't mount the disks directly to where `Root` points (`/var/lib/containerd`) since mounts will only propagate on `/mnt` in the host. Another problem is that any new mounted filesystem will be labeled with `unlabeled_t` rather than with `local_t`, so we need to figure out what's the best way to handle the initial relabel in those filesystems

arnaldo2792:
It is possible to set up ephemeral disks as the storage for container images, as described in #1991

PR #1729, opened 2021-09-02
fix various SELinux policy issues

bcressey:
**Issue number:**
#1617, #1651, #1712

**Description of changes:**
This includes some SELinux policy fixes, as well as some refactoring to pave the way for MCS isolation. More details are available in the commit messages, but at a high level, here's what's changing.

A new `proc_t` object type is added to fix an AVC denial that processes could encounter when writing to `/proc`. This didn't appear to cause any real trouble; as far as I can tell, this wasn't surfaced to programs as an error.

The `unlabeled_t` and `external_t` object types are removed, and aliases are added for `local_t`. The distinction between different types of local files was never very useful and caused problems when creating new filesystems via bootstrap containers, as discussed in #1617. This should also help with #1651.

While cleaning up the `local_t` mount options that are no longer needed, I fixed an oversight with the `/usr/src/kernels` mount, which is not intended to be writable by unprivileged containers.

Finally there are the changes that lay the groundwork for MCS isolation. Adding a distinct type for container files - `data_t` vs. `local_t` - enables us to assert that `data_t` must always follow the isolation rules. Running privileged and host containers with the full MCS range clarifies the policy goal that they are allowed to bypass isolation.

Specifying the default range transition fixes two problems, both caused by using the source context. Privileged containers would create files with the wrong level - either too weak (`s0`) or too strong (`s0-s0:c0.c1023`). Unprivileged containers would add their MCS pairs to object types that aren't meant to be isolated, such as `any_t`.

**Testing done:**
Verified that nodes came up and sonobuoy passed with no SELinux denials. Previously, nodes would sometimes log a denial for `httpd` because of the "associate" issue that's now resolved.

Confirmed that the directories under `/var/lib/{host,}containerd` had the expected labels.

Verified that only `container_t` and `data_t` types use MCS pairs:
```
# find -context '*:s0:c*,c*' ! -context '*:data_t:*' ! -context '*:container_t:*'
<no output>
```

Verified that only `control_t` and `super_t` types use the full MCS range:
```
# find -context '*:s0-s0:c*.c*' ! -context '*:control_t:*' ! -context '*:super_t:*'
<no output>
```

**Terms of contribution:**

### DIFF
--- a/packages/selinux-policy/fs.cil
+++ b/packages/selinux-policy/fs.cil
@@ -97,7 +97,8 @@
 (filecon "/run/.*" any ())
 
 ; Label external filesystem mounts.
-(filecon "/mnt" any external)
+(filecon "/mnt" any local)
 (filecon "/mnt/.*" any ())
-(filecon "/media" any external)
+(filecon "/media" any local)
+(filecon "/media/cdrom" any local)
 (filecon "/media/.*" any ())
--- a/packages/selinux-policy/fs.cil
+++ b/packages/selinux-policy/fs.cil
@@ -26,7 +26,7 @@
 (genfscon debugfs / any)
 (genfscon kvmfs / any)
 (genfscon nsfs / any)
-(genfscon proc / any)
+(genfscon proc / proc)
 (genfscon pstore / any)
 (genfscon ramfs / any)
 (genfscon rootfs / any)
@@ -81,7 +81,7 @@
 (filecon "/var/lib/netdog/.*" any lease)
 
 ; Label kernel filesystem mounts.
-(filecon "/proc" any any)
+(filecon "/proc" any proc)
 (filecon "/proc/.*" any ())
 (filecon "/sys" any any)
 (filecon "/sys/.*" any ())
--- a/packages/release/prepare-local.service
+++ b/packages/release/prepare-local.service
@@ -10,7 +10,6 @@ After=dev-disk-by\x2dpartlabel-BOTTLEROCKET\x2dDATA.device
 Type=oneshot
 Environment=BOTTLEROCKET_DATA=/dev/disk/by-partlabel/BOTTLEROCKET-DATA
 Environment=LOCAL_DIR=/local
-Environment=CONTEXT="system_u:object_r:local_t:s0"
 
 # To "grow" the partition, we delete it and recreate it at the larger size, then
 # write it back to the device. udevd observes the write via inotify, and tells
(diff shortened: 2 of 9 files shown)


packages/selinux-policy/fs.cil
```diff
@@ -97,7 +97,8 @@
 (filecon "/run/.*" any ())
 
 ; Label external filesystem mounts.
-(filecon "/mnt" any external)
+(filecon "/mnt" any local)
 (filecon "/mnt/.*" any ())
-(filecon "/media" any external)
+(filecon "/media" any local)
+(filecon "/media/cdrom" any local)
```
tjkirch, line 103: What's special about cdrom?

tjkirch (approved):

packages/selinux-policy/fs.cil
bcressey: We create it in the `filesystem` package, and if we don't label it then it can end up with an unexpected SELinux label if we build the image on a host with SELinux enabled.

That happens since we build inside a container, and files end up with container labels in the absence of an override. Our policy has type aliases for the Fedora object types, so they're valid at runtime.

This affected the images I build, and I found it while writing a test to assert that none of the container object labels were present outside `/local`.

Every other file and directory is already covered by a labeling rule.

tjkirch: I'm curious whether changing `(filecon "/media/.*" any ())` to `local` instead of `()` would prevent issues with any devices like cdrom that we may add, without having to think about this bit of policy in the future.  I'm not sure what other effects that might have, though, so I'm happy with whatever you like.

samuelkarp:
> Privileged containers would create files with the wrong level - either too weak (s0) or too strong (s0-s0:c0.c1023). Unprivileged containers would add their MCS pairs to object types that aren't meant to be isolated, such as any_t.

Are these labels persisted through updates?  If an existing Bottlerocket instance has files with these labels from old containers and is updated to a new OS image that includes these policy changes, do those old files need to be relabeled?

bcressey:
I've confirmed that this fixes #1651.

bcressey:
> Are these labels persisted through updates? If an existing Bottlerocket instance has files with these labels from old containers and is updated to a new OS image that includes these policy changes, do those old files need to be relabeled?

Yes, the labels can be persisted - for example, the VPC CNI plugin writes these files:
```
# ls -1Z /run/aws-node/ipam.json /var/log/aws-routed-eni/ipamd.log
system_u:object_r:any_t:s0:c210,c729 /run/aws-node/ipam.json
system_u:object_r:local_t:s0:c210,c729 /var/log/aws-routed-eni/ipamd.log
```

Adding the new `data_t` type will let us say that only `data_t` objects should have MCS isolation enforced. After an upgrade, any files with `local_t:s0:cX,cY` labels will be treated as though they were just `local_t`. On downgrade, any files with a persisted `data_t` label will be treated as `unlabeled_t`, since that label didn't exist in older policies. In both cases, reads and writes will be allowed for any subject.

So I don't think we need any explicit relabel operations. It's something I'd like to avoid if possible since it's rather expensive.

arnaldo2792 (approved):
:boot: :boat:

samuelkarp (approved):

Issue #1569, opened 2021-05-12
Unexpected AVC Denial when running pip install during a docker build on Docker-in-Docker


backjo:
Apologies if this is expected behavior - I'm fairly new to working with SELinux. 

**Platform I'm running on:**
BottleRocket 1.1.0 - EKS 1.19
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubernetes.io/psp: eks.privileged
  name: test-pod
spec:
  containers:
  - image: docker:dind
    name: dind
    securityContext:
      privileged: true
      seLinuxOptions:
        level: s0
        role: system_r
        type: super_t
        user: system_u
```

```
FROM ubuntu:20.04
RUN apt update
RUN apt install python3-pip -y
RUN pip install 'ansible-lint<5'
```

**What I expected to happen:**
When running a docker build of the above Dockerfile on a privileged pod using DinD (Docker in Docker), I'd expect the build to complete successfully.

**What actually happened:**
The docker build failed with permission denied errors.

build error:
```
 error: [Errno 13] Permission denied: '/usr/local/lib/python3.8/dist-packages/ansible_base-2.10.9.egg-info/SOURCES.txt'
```

AVC Denials:
```
[130009.198765] audit: type=1400 audit(1620841175.265:16780): avc:  denied  { relabelfrom } for  pid=3815087 comm="python3" name="SOURCES.txt" dev="overlay" ino=2008528 scontext=system_u:system_r:control_t:s0 tcontext=system_u:object_r:local_t:s0 tclass=file permissive=0
[130014.396610] audit: type=1400 audit(1620841180.465:16781): avc:  denied  { relabelfrom } for  pid=3815920 comm="python3" name="SOURCES.txt" dev="overlay" ino=556241 scontext=system_u:system_r:control_t:s0 tcontext=system_u:object_r:local_t:s0 tclass=file permissive=0
[130018.499288] audit: type=1400 audit(1620841184.565:16782): avc:  denied  { relabelfrom } for  pid=3816115 comm="python3" name="SOURCES.txt" dev="overlay" ino=558269 scontext=system_u:system_r:control_t:s0 tcontext=system_u:object_r:local_t:s0 tclass=file permissive=0
```

Dockerd seems to be running with control_t
```bash 
// From the admin container
ps -e
system_u:system_r:control_t:s0  3795910 ?        00:00:07 dockerd
```

**How to reproduce the problem:**
[...]

PR #1733, opened 2021-09-10
add MCS constraints to the SELinux policy

### LINKED ISSUE #1712: enable MCS isolation in the SELinux policy

bcressey:
**Issue number:**
Fixes #1712
#1569

**Description of changes:**
This adds `mlsconstrain` statements to the SELinux policy so that the MCS category pairs assigned to container processes and files are used for access decisions.

The CIL documentation on [mlsconstrain](https://github.com/SELinuxProject/selinux/blob/master/secilc/docs/cil_constraint_statements.md#mlsconstrain) is helpful to understand what's happening, but briefly:
* `dom` is the dominance test described in the comments
* `h1` is the source "high" context (typically a subject / process)
* `h2` is the target "high" context (either an object / file, or another subject / process)

The "low" context (`l1`, `l2`) should be just `s0` for everything (with no categories), and is not useful for making the access decision.

Some special cases, since it's hard to write useful comments about what's *not* included:
* `(files (load))` has an exception for "subject" files (like `/proc/<pid>/<file>`), so `ps` works for everyone.
* `(files (describe))` does not have a constraint, so `ls` works for everyone.
* `(files (relabel))` and `(processes (transform))` are "relabel" actions; they're not constrained because they're currently reserved for trusted (read: "not container") processes.

**Testing done:**
Tested these scenarios.

| Action | Privileged container | Unprivileged container |
| --- | --- | --- |
| Create files in an `emptyDir` volume | ✅ | ✅ |
| Create files in a `hostPath` volume | ✅ | ✅ |
| Create files in an Amazon EBS volume | ✅ | ✅ |
| Create files in an Amazon EFS volume | ✅ | ✅ |
| Create files in a Docker image volume | ✅ | ✅ |
| Signal processes with a different level | ✅ | ❌ |
| List files in a directory with a different level | ✅ | ❌ |
| Create files in a directory with a different level | ✅ | ❌ |
| Relabel files with the same level | ✅ | ✅ |
| Relabel files with a different level | ❌ | ❌ |
| Relabel processes with the same level | ✅ | ✅ |
| Relabel processes with a different level | ❌ | ❌ |
| List files in host's `/etc` | ✅ | ❌ |
| Create files in host's `/etc` | ❌ | ❌ |
| List files for host's datastore | ✅ | ❌ |
| Create files in host's datastore | ❌  | ❌ |

Verified that `podman` and `pip` can run inside a container without unexpected denials.

**Terms of contribution:**

### DIFF
--- a/packages/selinux-policy/files.cil
+++ b/packages/selinux-policy/files.cil
@@ -9,8 +9,6 @@
 (classmapping files relabel relabel_blk_file)
 (classmapping files relabel relabel_sock_file)
 (classmapping files relabel relabel_fifo_file)
-(classmapping files relabel relabel_filesystem)
-(classmapping files relabel relabel_kernel_service)
 
 ; Permission group for mounts.
 (classmapping files mount mount_file)
@@ -81,8 +79,6 @@
 (classpermission relabel_blk_file)
 (classpermission relabel_sock_file)
 (classpermission relabel_fifo_file)
-(classpermission relabel_filesystem)
-(classpermission relabel_kernel_service)
 (classpermissionset relabel_file (
   file (relabelfrom relabelto)))
 (classpermissionset relabel_dir (
@@ -97,10 +93,6 @@
   sock_file (relabelfrom relabelto)))
 (classpermissionset relabel_fifo_file (
   fifo_file (relabelfrom relabelto)))
-(classpermissionset relabel_filesystem (
-  filesystem (relabelfrom relabelto)))
-(classpermissionset relabel_kernel_service (
-  kernel_service (create_files_as)))
 
 ; Sets of permissions for mounts.
 (classpermission mount_file)
--- a/packages/selinux-policy/mcs.cil
+++ b/packages/selinux-policy/mcs.cil
@@ -1,0 +1,107 @@
+; SELinux dominance rules are somewhat involved but the usage below
+; is straightforward. A quick summary of the relevant details:
+;
+; * s0:c0,c1 (source) dominates s0:c1 (target) since it has two
+;   categories, one of which is c1
+; * s0:c0.c4 (source) dominates s0:c1 (target) since it has a range of
+;   categories, from c0 to c4, which includes c1
+; * s0:c0,c1 (source) neither dominates or is dominated by s0:c1,c2
+;   (target) since only one of the two categories is shared
+; * s0:c0,c1 (source) likewise does not dominate s0:c2,c3 (target)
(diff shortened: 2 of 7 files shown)


tjkirch (approved):

bcressey:
Added a new rule to prevent an untrusted process from altering its own range.

tjkirch (approved):

bcressey:
Modified the rules so that untrusted processes cannot adjust their own labels, even if the resulting value is the same. This aligns with the rules for relabeling files.

tjkirch (approved):

bcressey:
Fixed a typo spotted by @arnaldo2792.

arnaldo2792 (approved):
:tada:

samuelkarp (approved):

bcressey:
Added support for "no-op" relabeling actions where the same label is reapplied. Updated local tests to verify that relabeling actions that do not change the label are allowed, while those that change the label are denied.

tjkirch (approved):

peimanja:
any update on this one?

bcressey:
It should be merged soon, just waiting on one more review since I reworked it a bit since the reviewers last saw it.

Let me know if there's a test case that you're interested in that you'd like me to try?

peimanja:
> Let me know if there's a test case that you're interested in that you'd like me to try?

Thanks @bcressey! Basically we run github actions on self-hosted runners on Bottlerocket with dind runners for building images.

Workaround you've mentioned in https://github.com/bottlerocket-os/bottlerocket/issues/1569#issuecomment-840101836 works fine but would be nice to get a cleaner way of doing that.

samuelkarp (approved):
LGTM

bcressey:
> Workaround you've mentioned in [#1569 (comment)](https://github.com/bottlerocket-os/bottlerocket/issues/1569#issuecomment-840101836) works fine but would be nice to get a cleaner way of doing that.

I've been using a version of the pod spec from that issue, with `privileged: true` and without `seLinuxOptions`, and it works fine after these changes.

PR #1757, opened 2021-09-20
Update hash for v3 root.json

arnaldo2792:
**Description of changes:**

We released a new version of the `root.json` file

**Terms of contribution:**


etungsten (approved):

tjkirch (approved):
(Note: we're not changing the root.json in the cache bucket until this is approved, as we merge it, so that the public documentation is incorrect for the shortest possible period.)

zmrow (approved):
:mag:

webern (approved):

PR #1762, opened 2021-09-28
Add region flag to aws-iam-authenticator command

WilboMo:
**Issue number:**
N/A

**Description of changes:**
Adds `--region` flag to `aws-iam-authenticator` command to reflect AL2 EKS changes to support additional regions. 

**Testing done:**
Created EKS Clusters for the following variants: 
* aws-k8s-1.17
* aws-k8s-1.18
* aws-k8s-1.19
* aws-k8s-1.20
* aws-k8s-1.21

Confirmed updates to `kubelet-kubeconfig` allowed Bottlerocket nodes for each cluster to join successfully and deploy pods. 

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.17/kubelet-kubeconfig
+++ b/packages/kubernetes-1.17/kubelet-kubeconfig
@@ -26,6 +26,10 @@ users:
       - token
       - "-i"
       - "{{settings.kubernetes.cluster-name}}"
+      {{#if settings.aws.region}}
+      - "--region"
+      - "{{settings.aws.region}}"
+      {{/if}}
 {{/if}}
 {{/if}}
 {{#if (eq settings.kubernetes.authentication-mode "tls")}}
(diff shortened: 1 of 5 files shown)


packages/kubernetes-1.17/kubelet-kubeconfig
```diff
@@ -26,6 +26,8 @@ users:
       - token
       - "-i"
       - "{{settings.kubernetes.cluster-name}}"
+      - --region
+      - "{{settings.aws.region}}"
```
bcressey: It's possible to set the auth mode to "aws" for platforms other than AWS, and although it wouldn't work today, it might in the future.

So it'd be good to guard this with a check that the rendered setting is defined:
```suggestion
{{#if settings.aws.region}}
      - --region
      - "{{settings.aws.region}}"
{{/if}}
```

zmrow:
LGTM once @bcressey 's suggestion is fixed.

packages/kubernetes-1.17/kubelet-kubeconfig
WilboMo: Sounds good, thanks for the feedback!

webern (approved):

zmrow (approved):
:jigsaw:

bcressey (approved):

PR #1612, opened 2021-06-15
Add aws-k8s-1.21, vmware-k8s-1.21 variants

etungsten:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket/issues/1611

**Description of changes:**

For reference, the addition of aws-k8s-1.20 was in https://github.com/bottlerocket-os/bottlerocket/pull/1437.

Version of `kubelet` packaged: v1.21.2

```
Author: Erikson Tung <etung@amazon.com>
Date:   Tue Jun 15 11:31:57 2021 -0700

    Add aws-k8s-1.21 variant
    
    Adds aws-k8s-1.21 variant, relinks symlinks in models

```
```
Author: Erikson Tung <etung@amazon.com>
Date:   Tue Jun 15 11:55:19 2021 -0700

    Add vmware-k8s-1.21 variant
    
    Adds vmware-k8s-1.21 variant, relinks symlinks in models

```

```
Author: Erikson Tung <etung@amazon.com>
Date:   Tue Jun 15 11:30:21 2021 -0700

    packages: add kubernetes-1.21
    
    * No longer need '0001-always-set-relevant-variables-for-cross-compiling.patch'
    * Third party package 'github.com/miekg/dns' version bump included some
    changes in its license file so the hash changed. The license is still
    'BSD-3-Clause'.
```

`0001-always-set-relevant-variables-for-cross-compiling.patch` is no longer needed as a result of the following changes in 1.21:
* https://github.com/kubernetes/kubernetes/commit/fd01467c0c3211f7014e7bf3c6fa62ef87d65240 
* https://github.com/kubernetes/kubernetes/commit/f19f22b43edf498444817a0052b71d6084e68db9
We export the `KUBE_*_CC` variables directly.

**Testing done:**
- [x] Pods run OK

- [x]  Conformance testing with 1.21 control plane and x86_64 aws-k8s-1.21 nodes

- [x]  Conformance testing with 1.21 control plane and aarch64 aws-k8s-1.21 nodes

- [x]  Conformance testing with 1.21 control plane and x86_64 vmware-k8s-1.21 nodes

- [x] Confirm the EBS CSI driver still works.

- [x] Confirm the vSphere CSI driver still works - (Block volumes only since file volumes are not yet supported on AWS VMC)

- [x] Check kubelet logs and system logs in general, compared them with the aws-k8s-1.20 variant kubelet logs:

There's a new kubelet warning during start up:
```
Flag --network-plugin has been deprecated, will be removed along with dockershim.
```
[...]

PR #1763, opened 2021-09-29
update glibc to 2.34

### LINKED ISSUE #1675: Update libglib to 2.69.0+
Before 2.69.0, libglib used a bundled copy of libpcre.  We had a dependency on libpcre2, but it [only supports libpcre (v1, i.e. 8.x)](https://gitlab.gnome.org/GNOME/glib/-/issues/1085) so it was building and using its own bundled version.

In 2.69.0, they removed the bundled libpcre and instead use a meson wrapdb build of libpcre.

To update, it seems like we'll need to either (1) package libpcre v1, or (2) make meson wrapdb builds work in buildsys, i.e. fetching its expected artifacts early or otherwise making it believe it doesn't need network access.

### LINKED ISSUE #1717: update to glibc 2.34
[glibc 2.34 is out!](https://lwn.net/Articles/864920/)

We'll also want the fix for [CVE-2021-38604](https://sourceware.org/git/?p=glibc.git;a=commit;h=b805aebd42364fe696e417808a700fdb9800c9e8).

bcressey:
**Issue number:**
Fixes #1717
Fixes #1675

**Description of changes:**
Update glibc to 2.34. Change the patching strategy to include patches added to the maintenance branch after the official 2.34 release, which aligns with Fedora's current approach.

Fix a cross-compilation bug for kubelet that caused us to link with target libraries for host programs when the architectures were the same. This was exposed by missing glibc symbols following the update, but was a latent regression from when we switched to the new upstream method of specifying the toolchain in #1612. It does not apply to older builds of kubelet, which used a local patch instead.

Add the upstream patch for chrony's seccomp filter so it runs with the newer glibc.

Update glib2 to 2.70.0, which includes a fix for the newer glibc. Tweak the cross-compiling file for meson to avoid attempts to execute target files, which failed because of missing glibc symbols.

**Testing done:**
Verified that `chronyd` starts running and stays up for multiple days on both architectures. Using the developer image, confirmed that `chronyc` can issue commands like "dump" without causing a seccomp violation.

Tested `aws-k8s-1.21` for both architectures and verified that `sonobuoy` passed. Confirmed that tasks could run on an `aws-ecs-1` build.

Tested `vmware-k8s-1.21` with a cluster created by EKS Anywhere. Verified that `sonobuoy` passed.

**Terms of contribution:**


etungsten (dismissed):

etungsten:
vmware builds are failing due to `libglib` package needed by `open-vm-tools`

packages/glibc/glibc.spec
```diff
@@ -1,14 +1,47 @@
...
 Summary: The GNU libc libraries
 License: LGPL-2.1-or-later AND (LGPL-2.1-or-later WITH GCC-exception-2.0) AND GPL-2.0-or-later AND (GPL-2.0-or-later WITH GCC-exception-2.0) AND BSD-3-Clause AND ISC
 URL: http://www.gnu.org/software/glibc/
 Source0: https://ftp.gnu.org/gnu/glibc/glibc-%{version}.tar.xz
 Source1: glibc-tmpfiles.conf
-Patch1: glibc-cs-path.patch
-Patch2: glibc-c-utf8-locale.patch
-Patch1001: 1001-move-ldconfig-cache-to-ephemeral-storage.patch
+
+# Upstream patches from 2.34 release branch
```
tjkirch:
```suggestion
# Upstream patches from 2.34 release branch:
# ```
# git checkout origin/release/2.34/master
# git format-patch glibc-2.34..
# ```
```

bcressey:
Back to draft to fix `open-vm-tools`.

tjkirch (commented):
LGTM modulo the vmware build issue.

packages/libglib/libglib.spec
```diff
@@ -1,10 +1,14 @@
...
 Release: 1%{?dist}
 Summary: The GLib libraries
-License: LGPL-2.1-or-later
+# glib2 is LGPL-2.1-only
+# pcre is BSD-3-Clause
+License: LGPL-2.1-only AND BSD-3-Clause
 URL: https://www.gtk.org/
-Source0: https://download.gnome.org/sources/glib/2.68/glib-%{version}.tar.xz
+Source0: https://download.gnome.org/sources/glib/2.70/glib-%{version}.tar.xz
+Source1: https://ftp.pcre.org/pub/pcre/pcre-8.37.tar.bz2
```
tjkirch, line 11:
```suggestion
# Note: the pcre version is specified in the glib archive in subprojects/libpcre.wrap
Source1: https://ftp.pcre.org/pub/pcre/pcre-8.37.tar.bz2
```

tjkirch (approved):
I left a couple suggestions to help future packagers keep these up to date, but overall it looks good to me.

arnaldo2792 (approved):
:motor_boat:

etungsten (approved):

zmrow (approved):
:kangaroo:

PR #1764, opened 2021-09-29
docs: fix link to issue labels

andrewhsu:
Fixed link to issue label indicating a good place to start for contributors.

Would also be good if a maintainer for this repo could apply the `status/helpwelcome` label to a few issues (if there are suitable issues for this label) so the list is not empty.


etungsten (approved):
Thanks for the fix and suggestion! We'll start combing through our issues list and label them as appropriate.

zmrow (approved):
:+1: 
Thanks!

webern (approved):

PR #1766, opened 2021-10-01
docs: Mention static pods in the security guidance around api access

zmrow:
**Description of changes:**
```
We recommend against providing access to the API socket from containers
because of the effects it can have on system configuration and security.
This change specifically calls out the ability to define static pods as
an action that could be taken with API access, and the effects of doing
so.
```

**Terms of contribution:**

### DIFF
--- a/SECURITY_GUIDANCE.md
+++ b/SECURITY_GUIDANCE.md
@@ -71,6 +71,8 @@ It is labeled `api_socket_t`, so only processes with privileged SELinux labels c
 Write access to this socket will grant full control over system configuration.
 This includes the ability to define an arbitrary source for a host container, and to run that container with "superpowers" that bypass other restrictions.
 These "superpowers" are described [below](#limit-use-of-host-containers).
+For Kubernetes variants, it also includes the ability to define and run static pods.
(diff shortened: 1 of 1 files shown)


etungsten (approved):

SECURITY_GUIDANCE.md
```diff
@@ -71,6 +71,7 @@ It is labeled `api_socket_t`, so only processes with privileged SELinux labels c
 Write access to this socket will grant full control over system configuration.
 This includes the ability to define an arbitrary source for a host container, and to run that container with "superpowers" that bypass other restrictions.
 These "superpowers" are described [below](#limit-use-of-host-containers).
+It also includes the ability to define and run static pods, which are managed directly by `kubelet` and are not subject to admission controllers that enforce security policies for the Kubernetes cluster.
```
bcressey:
```suggestion

For Kubernetes variants, it also includes the ability to define and run static pods. These are managed directly by `kubelet` and are not subject to admission controllers that enforce security policies for the cluster.
```

bcressey (commented):
s/affects/effects/g in the commit message

zmrow:
^ Fixed @bcressey 's suggestions

bcressey (approved):

PR #1773, opened 2021-10-05
Add *.zip to packages/.gitignore

jpculp:
**Description of changes:**

This ignores files like `packages/libglib/pcre_8.37-2_patch.zip`.

Also removes an individual `.gitignore` for `ncurses-*.tgz` as the **ncurses** package is now `packages/ncurses/ncurses-6.2.tar.gz`, which is caught by the `*.tar.*` in the repo-level `.gitignore`.

**Testing done:**

Confirmed that there were no untracked ZIPs or TGZs after build.

**Terms of contribution:**

### DIFF
--- a/packages/.gitignore
+++ b/packages/.gitignore
@@ -1,2 +1,3 @@
 *.patch.bz2
 *.src.rpm
+*.zip
--- a/packages/ncurses/.gitignore
+++ b/packages/ncurses/.gitignore
@@ -1,1 +1,0 @@
-ncurses-*.tgz


etungsten (approved):

tjkirch (approved):
I don't think we'd ever check in any zipped patch files; if they're large enough for it, we'd probably put them in the lookaside cache, like the aws-sdk-go dep update patch in kubernetes-1.17.

bcressey (approved):

webern (approved):

PR #1781, opened 2021-10-18
Support '*' in Bottlerocket setting keys

### LINKED ISSUE #1780: Wildcard Container Registry Mirrors
**Image I'm using:**
`bottlerocket-aws-k8s-1.21-x86_64-v1.2.1-02e4faf1`

Bottlerocket starts cleanly and joins an EKS cluster. Is also able to pull unqualified images from either https://registry-0.acme.com or https://registry-1.acme.com.

**What actually happened:**
Bottlerocket fails to join the EKS cluster

Configure Bottlerocket's user data with the following toml:
```toml
[settings.container-registry.mirrors]
"*" = ["https://registry-0.acme.com","https://registry-1.acme.com"]
```

This configuration should be supported by containerd as per https://github.com/containerd/containerd/blob/main/docs/cri/registry.md#configure-registry-endpoint

etungsten:
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/1780

**Description of changes:**
```
Date:   Mon Oct 18 12:13:19 2021 -0700

    datastore: adds '*' to the list of valid key characters
    
    Adds '*' to the list of valid key characters.
    Adds unit tests to make sure paths components are correctly encoded and
    decoded.

```

**Testing done:**
Adds additional unit tests with `*` in the key segments and path components for the datastore and they all pass.

On a bottlerocket host, I was able to set a setting with a key of `*` and see that it gets encoded correctly in the datastore.
Bottlerocket was also able to configuration for containerd with a settings key containing `*`.

Userdata:
```
[settings.container-registry.mirrors]
"*" = ["https://registry-0.acme.com","https://registry-1.acme.com"]
```

On the host (note that `%2A` is `*` URL encoded):
```
bash-5.0# cat /var/lib/bottlerocket/datastore/current/live/settings/container-registry/mirrors/%2A 
["https://registry-0.acme.com","https://registry-1.acme.com"]

```

```
...
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."*"]
endpoint = ["https://registry-0.acme.com", "https://registry-1.acme.com"]

```

**Terms of contribution:**

### DIFF
--- a/sources/api/datastore/src/filesystem.rs
+++ b/sources/api/datastore/src/filesystem.rs
@@ -603,6 +603,24 @@ mod test {
     }
 
     #[test]
+    fn wonky_data_path() {
+        let f = FilesystemDataStore::new("/base");
+        let key = Key::new(KeyType::Data, "a.b.c.*.d").unwrap();
+
+        let tx = "test transaction";
+        let pending = f
+            .data_path(&key, &Committed::Pending { tx: tx.into() })
+            .unwrap();
+        assert_eq!(
+            pending.into_os_string(),
+            "/base/pending/test%20transaction/a/b/c/%2A/d"
+        );
+
+        let live = f.data_path(&key, &Committed::Live).unwrap();
+        assert_eq!(live.into_os_string(), "/base/live/a/b/c/%2A/d");
+    }
+
+    #[test]
     fn metadata_path() {
         let f = FilesystemDataStore::new("/base");
         let data_key = Key::new(KeyType::Data, "a.b.c").unwrap();
@@ -629,6 +647,7 @@ mod test {
         assert_eq!(encode_path_component("a.b"), "a%2Eb");
         assert_eq!(encode_path_component("a/b"), "a%2Fb");
         assert_eq!(encode_path_component("a b%c<d>e"), "a%20b%25c%3Cd%3Ee");
+        assert_eq!(encode_path_component("a.b.*"), "a%2Eb%2E%2A");
     }
 
     #[test]
@@ -640,6 +659,7 @@ mod test {
             decode_path_component("a%20b%25c%3Cd%3Ee", "").unwrap(),
             "a b%c<d>e"
         );
+        assert_eq!(decode_path_component("a%2Eb%2E%2A", "").unwrap(), "a.b.*");
 
         // Invalid UTF-8
         decode_path_component("%C3%28", "").unwrap_err();
--- a/sources/api/datastore/src/key.rs
+++ b/sources/api/datastore/src/key.rs
@@ -233,7 +233,7 @@ impl Key {
     /// separate from quoting; if a character isn't valid, it isn't valid quoted, either.
     fn valid_character(c: char) -> bool {
         match c {
-            'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '-' | '/' => true,
+            'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '-' | '/' | '*' => true,
             _ => false,
         }
     }
@@ -455,7 +455,7 @@ mod test {
 
     #[test]
     fn key_with_special_chars_ok() {
-        data_and_meta!(|t| assert!(Key::new(t, "a-b_c").is_ok()));
+        data_and_meta!(|t| assert!(Key::new(t, "a-b_c*").is_ok()));
     }
 
     #[test]


etungsten:
Dropping the cargo fmt commit for now. I'll create a separate PR for it.

Repeated my testing, and things still work as expected.

jpculp (approved):

arnaldo2792 (approved):

zmrow (approved):

bcressey:
🤔 

LGTM on the surface, but I'm talking this over with @tjkirch to see if there are subtle implications to allowing this.

bcressey:
I'm mostly concerned with cases where people might try to set `host-containers.*.enabled = false` or `kernel.sysctl."net.ipv4.conf.*.rp_filter" = "2"` i.e. where the surface syntax has an "obvious" meaning that is totally different from what the implementation does.

Can you audit the settings where we allow user-defined keys and make sure the types are suitably constrained in the character set they allow?

tjkirch:
> I'm mostly concerned with cases where people might try to set `host-containers.*.enabled = false` or `kernel.sysctl."net.ipv4.conf.*.rp_filter" = "2"` i.e. where the surface syntax has an "obvious" meaning that is totally different from what the implementation does.
> 
> Can you audit the settings where we allow user-defined keys and make sure the types are suitably constrained in the character set they allow?

I think this concern applies anywhere we have a Map in the model.  Today that's host/bootstrap container names, certificate identifiers, sysctl names, registry mirrors (the intended case for *), ECS instance attributes, and for k8s: static pods, node labels/taints, eviction-hard, and kube/system reserved.  They all use modeled types that were intended to keep us safe in cases like this, except for registry mirrors (the setting this PR is about) which allows any SingleLineString as key.

I'm worried more about future expansion, new settings where people don't realize this is an issue and use String/SingleLineString or a new modeled type that isn't careful enough about `*` and suddenly there's wildcard behavior somewhere, or at minimum, some potential confusion about whether `*` in a BR key implies that it applies to a whole subset of settings rather than just one.

I think we should reconsider how we could support the desired behavior here.  One option is to remodel registry mirrors so that instead of a Map of registry-name=value, we have a struct with name=foo and value=bar, or something.  Then the `*` would only appear in a value, which is safer.  We could add the new style, deprecate the current style, notify users in changelogs, and remove it a few releases later, migrating any existing settings to the new.  (Maybe even apply the same migration to multiple releases, so that users inspecting their settings see that their settings were changed to a new format; another form of notification.)  Thoughts?

etungsten:
> host-containers.*.enabled = false

TOML doesn't allow bare `*` keys, so they must be quoted. The `host-containers` model is a map of `Identifier` to the host-container definition struct type. `Identifier` does not allow `*` as a value.

> kernel.sysctl."net.ipv4.conf.*.rp_filter" = "2"

Similarly, `SysctlKey` does not allow `*` characters.

> I'm worried more about future expansion, new settings where people don't realize this is an issue and use String/SingleLineString or a new modeled type that isn't careful enough about * and suddenly there's wildcard behavior somewhere.

With the way we URL encode non-alphanumeric characters in setting keys, that should never happen, I think? At least within the context of the settings datastore. After decoding the settings key for purpose of rendering a configuration file (like in this case), I think it's very clear that the person wants the settings key as is, if the modeled type intentionally allows `*`.

> or at minimum, some potential confusion about whether * in a BR key implies that it applies to a whole subset of settings rather than just one.

I think as long as we're very deliberate and strict about what we allow and disallow via the modeled type, the apiserver will fail to apply the setting to prevent any confusing behaviors around `*` as a settings key.

> I think we should reconsider how we could support the desired behavior here. One option is to remodel registry mirrors so that instead of a Map of registry-name=value, we have a struct with name=foo and value=bar, or something. 

Would the proposed structure look like this in user-data?:
```
[settings.container-registry]
[[mirrors]]
host = "*"
endpoints = ["hello"]

[[mirrors]]
host = "example"
endpoints = [ "blah", "blahblah"]
```

I think the way `container-registry.mirrors` is modelled right now is pretty straightforward and follows [Containerd's Configure Registry Endpoint](https://github.com/containerd/containerd/blob/main/docs/cri/registry.md#configure-registry-endpoint) configuration (also in toml)  in structure. Users familiar with how to configure for containerd can translate their existing configuration elsewhere to Bottlerocket pretty easily. I guess that's maybe a super minor thing.

That being said, I'm OK with going forward with restructuring `container-registry.mirrors` to completely remove `*` from ever becoming a settings key.

tjkirch:
> as long as we're very deliberate and strict

That's the concern in a nutshell - it's hard to guarantee every modeled type will be perfect, every future change to the data store and API will be perfect, etc.  It's safer to avoid the situation entirely, if there's a comparable replacement.

> Would the proposed structure look like this in user-data?

Yeah, something like that, with only known key names.  I think it's equivalent in functionality and comparable in friendliness, and I don't think people expect to directly port settings from other OSes because of our different structure.  It'd be the first time (?) we restructure and deprecate a model setting, but if we need to allow wildcards (we're sure?) then it could be worth it.

etungsten:
Taking PR into draft since we're reconsidering approach.

etungsten:
Alternative approach is in https://github.com/bottlerocket-os/bottlerocket/pull/1791. If that merges, I'll close this PR.

PR #1783, opened 2021-10-21
pluto: don't error on missing settings

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
Date:   Thu Oct 21 10:48:44 2021 -0700

    pluto: don't error on missing settings
    
    This makes it so that pluto does not error early if some Bottlerocket
    settings are missing to allow fallbacks later when determining node-ip
    and cluster-dns-ip.

```

**Testing done:**
Was able to launch a Bottlerocket instance with `kubernetes.standalone-mode` and the host comes up fine and kubelet is running fine with a node-ip:
```
bash-5.0# systemctl status kubelet
● kubelet.service - Kubelet
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/kubelet.service; enabled; vendor preset: enabled)
    Drop-In: /etc/systemd/system/kubelet.service.d
             └─exec-start.conf
     Active: active (running) since Thu 2021-10-21 18:09:10 UTC; 28s ago
...
     CGroup: /runtime.slice/kubelet.service
             └─1581 /usr/bin/kubelet --cloud-provider  --config /etc/kubernetes/kubelet/config --container-runtime=remote --container-runtime-endpoint=unix:///run/dockershim.sock --containerd=/run/dockershim.sock --network-plugin cni --root-dir /var/lib/kubelet --cert-dir /var/lib/kubelet/pki --node-ip 192.168.26.35 --node-labels  --register-with-taints  --pod-infra-container-image 602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause-amd64:3.1
```
```
bash-5.0# cat /etc/kubernetes/kubelet/config 
---
kind: KubeletConfiguration
address: 127.0.0.1
authentication:
  anonymous:
    enabled: true
  webhook:
    enabled: false
authorization:
  mode: AlwaysAllow
```
Repeated this test for combinations of `standalone-mode = true` and other kubernetes settings being set. The results are the same.

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/api.rs
+++ b/sources/api/pluto/src/api.rs
@@ -5,8 +5,8 @@ use std::net::IpAddr;
 pub(super) type Result<T> = std::result::Result<T, Error>;
 
 pub(crate) struct AwsK8sInfo {
-    pub(crate) region: String,
-    pub(crate) cluster_name: String,
+    pub(crate) region: Option<String>,
+    pub(crate) cluster_name: Option<String>,
     pub(crate) cluster_dns_ip: Option<IpAddr>,
 }
 
@@ -16,8 +16,8 @@ pub(crate) struct AwsK8sInfo {
 #[cfg(aws_k8s_variant)]
 mod inner {
     use super::*;
-    use snafu::{OptionExt, ResultExt, Snafu};
     use constants;
+    use snafu::{ResultExt, Snafu};
 
     #[derive(Debug, Snafu)]
     pub(crate) enum Error {
@@ -26,9 +26,6 @@ mod inner {
             source: apiclient::Error,
             uri: String,
         },
-
-        #[snafu(display("The '{}' setting is missing", setting))]
-        Missing { setting: String },
 
         #[snafu(display("Unable to deserialize Bottlerocket settings: {}", source))]
         SettingsJson { source: serde_json::Error },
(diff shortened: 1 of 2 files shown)


sources/api/pluto/src/api.rs
```diff
@@ -48,23 +45,19 @@ mod inner {
...
-                .context(Missing { setting: "region" })?
-                .into(),
-            cluster_name: kubernetes
-                .cluster_name
-                .context(Missing {
-                    setting: "cluster-name",
-                })?
-                .into(),
-            cluster_dns_ip: kubernetes.cluster_dns_ip,
+            region: settings.aws.map(|a| a.region).flatten().map(|s| s.into()),
```
tjkirch: nit: I think `.and_then(|a| a.region)` does the same as `.map(|a| a.region).flatten()`?

sources/api/pluto/src/api.rs
```diff
@@ -48,23 +45,19 @@ mod inner {
...
+            cluster_name: settings
+                .kubernetes
+                .as_ref()
+                .map(|k| k.cluster_name.as_ref())
```
tjkirch: nit: interesting; `KubernetesClusterName` has `Into<String>` via `string_impls_for`; do we need to take a reference and then clone?  Just doing `into()` in the final map doesn't work like the others?

tjkirch (approved):

sources/api/pluto/src/api.rs
etungsten: I take a reference of `cluster_name` because it's an option and it can't be `Copy`ed. I'll clone it directly instead.

sources/api/pluto/src/api.rs
etungsten: Good catch.

etungsten:

webern (approved):

tjkirch (approved):

bcressey (approved):

PR #1791, opened 2021-10-28
model: `container-registry.mirrors` different settings representation

### LINKED ISSUE #1780: Wildcard Container Registry Mirrors

etungsten:
**Issue number:**
Fixes #1780

**Description of changes:**
```
Date:   Wed Oct 27 19:02:01 2021 -0700

    model: change `container-registry.mirrors` type
    
    Changes the model for `container-registry.mirrors` from a `HashMap` to a vector of
    structs.
    Adds a custom deserializer to deserialize from either a TOML table or a
    TOML sequence.

```

```
Date:   Thu Oct 28 12:03:20 2021 -0700

    migrations: 'container-registry.mirrors' model type migration
```

**Testing done:**
Unit test passes.
On a live host, I can specify registry mirrors in both ways. Either as a map of registry to endpoints or an array of tables with `registry` and `endpoints` as object fields.

With the following userdata:
```
[[settings.container-registry.mirrors]]
registry = "*"
endpoint = ["https://registry-0.acme.com","https://registry-1.acme.com"]

[[settings.container-registry.mirrors]]
registry = "example"
endpoint = [ "hello", "hellohello"]
```

On K8s variants, the containerd config registry mirrors configuration gets rendered correctly:
```
bash-5.0# cat /var/lib/bottlerocket/datastore/current/live/settings/container-registry/mirrors 
[{"endpoint":["https://registry-0.acme.com","https://registry-1.acme.com"],"registry":"*"},{"endpoint":["hello","hellohello"],"registry":"example"}]

...
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."*"]
endpoint = ["https://registry-0.acme.com", "https://registry-1.acme.com"]
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."example"]
endpoint = ["hello", "hellohello"]

```

On ECS variants, the docker daemon json file generated correctly for `docker.io` registry mirrors and docker runs fine:
userdata:
```
[[settings.container-registry.mirrors]]
registry = "docker.io"
endpoint = ["https://registry-0.acme.com","https://registry-1.acme.com"]

[[settings.container-registry.mirrors]]
registry = "example"
endpoint = [ "hello", "hellohello"]
```
docker `daemon.json`:
```
{
  "log-driver": "journald",
  "live-restore": true,
  "max-concurrent-downloads": 10,
  "storage-driver": "overlay2",
  "exec-opts": ["native.cgroupdriver=cgroupfs"],
  "data-root": "/var/lib/docker",
  "selinux-enabled": true,
  "default-ulimits": { "nofile": { "Name": "nofile", "Soft": 1024, "Hard": 4096 } }
      ,
  "registry-mirrors": ["https://registry-0.acme.com", "https://registry-1.acme.com"]
}
```

host-containers services registry mirrors config also gets generated correctly:
```
bash-5.0# cat /etc/host-containers/host-ctr.toml 
[mirrors."a"]
endpoints = ["aaaa", "bbb"]
[mirrors."b"]
endpoints = ["cccc", "dddd"]
[mirrors."docker.io"]
endpoints = ["https://registry-0.acme.com", "https://registry-1.acme.com"]
```

Test migration upgrades and downgrades and the `container-registry.mirrors` setting gets migrated successfully:
With the following userdata:
```
[settings.container-registry.mirrors]
"docker.io" = ["https://registry-0.acme.com","https://registry-1.acme.com"]
"a" = ["aaaa","bbb"]
"b" = ["cccc","dddd"]
```
On current version Bottlerocket:
```
bash-5.0# cat /var/lib/bottlerocket/datastore/current/live/settings/container-registry/mirrors/
a            b            docker%2Eio  
bash-5.0# cat /var/lib/bottlerocket/datastore/current/live/settings/container-registry/mirrors/a 
["aaaa","bbb"]
bash-5.0#cat /var/lib/bottlerocket/datastore/current/live/settings/container-registry/mirrors/docker%2Eio 
["https://registry-0.acme.com","https://registry-1.acme.com"]
bash-5.0# updog update -r -n
Starting update to 1.4.0
Cannot schedule shutdown without logind support, proceeding with immediate shutdown.
Update applied: aws-k8s-1.21 1.4.0
```
Checking after upgrade:
```
bash-5.0# cat /var/lib/bottlerocket/datastore/current/live/settings/container-registry/mirrors 
[{"endpoint":["cccc","dddd"],"registry":"b"},{"endpoint":["aaaa","bbb"],"registry":"a"},{"endpoint":["https://registry-0.acme.com",bash-5.0# signpost rollback-to-inactivey":"docker.io"}]
```
Checking containerd config:
```
....

[plugins."io.containerd.grpc.v1.cri".registry.mirrors."a"]
endpoint = ["aaaa", "bbb"]
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."b"]
endpoint = ["cccc", "dddd"]
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
endpoint = ["https://registry-0.acme.com", "https://registry-1.acme.com"]


```
After downgrading back:
```
bash-5.0# cat /var/lib/bottlerocket/datastore/current/live/settings/container-registry/mirrors/docker%2Eio 
["https://registry-0.acme.com","https://registry-1.acme.com"]
```

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,13 +411,18 @@ These settings can be changed at any time.
 #### Container image registry settings
 
 The following setting is optional and allows you to configure image registry mirrors and pull-through caches for your containers.
-* `settings.container-registry.mirrors`: A mapping of container image registry to a list of image registry URL endpoints.  When pulling an image from a registry, the container runtime will try the endpoints one by one and use the first working one.
+* `settings.container-registry.mirrors`: An array of container image registry mirror settings.  Each element specifies the registry and the endpoints for said registry.
+When pulling an image from a registry, the container runtime will try the endpoints one by one and use the first working one.
   (Docker and containerd will still try the default registry URL if the mirrors fail.)
   * Example user data for setting up image registry mirrors:
   ```
-  [settings.container-registry.mirrors]
-  "docker.io" = ["https://<my-docker-hub-mirror-host>"]
-  "gcr.io" = ["https://<my-gcr-mirror-host>","http://<my-gcr-mirror-host-2>"]
+  [[settings.container-registry.mirrors]]
+  registry = "*"
+  endpoint = ["https://<example-mirror>","https://<example-mirror-2>"]
+
+  [[settings.container-registry.mirrors]]
+  registry = "docker.io"
+  endpoint = [ "https://<my-docker-hub-mirror-host>", "https://<my-docker-hub-mirror-host-2>"]
   ```
   If you use a Bottlerocket variant that uses Docker as the container runtime, like `aws-ecs-1`, you should be aware that Docker only supports pull-through caches for images from Docker Hub (docker.io).  Mirrors for other registries are ignored in this case.
 
--- a/Release.toml
+++ b/Release.toml
@@ -74,3 +74,6 @@ version = "1.3.0"
     "migrate_v1.3.0_hostname-affects-etc-hosts.lz4",
     "migrate_v1.3.0_control-container-v0-5-2.lz4",
 ]
+"(1.3.0, 1.4.0)" = [
+    "migrate_v1.4.0_registry-mirror-representation.lz4",
+]
--- a/packages/containerd/containerd-config-toml_k8s
+++ b/packages/containerd/containerd-config-toml_k8s
@@ -32,7 +32,7 @@ conf_dir = "/etc/cni/net.d"
 
 {{#if settings.container-registry.mirrors}}
 {{#each settings.container-registry.mirrors}}
-[plugins."io.containerd.grpc.v1.cri".registry.mirrors."{{@key}}"]
-endpoint = [{{join_array ", " this }}]
+[plugins."io.containerd.grpc.v1.cri".registry.mirrors."{{registry}}"]
+endpoint = [{{join_array ", " endpoint }}]
 {{/each}}
 {{/if}}
--- a/packages/os/host-ctr-toml
+++ b/packages/os/host-ctr-toml
@@ -1,6 +1,6 @@
 {{#if settings.container-registry.mirrors}}
 {{#each settings.container-registry.mirrors}}
-[mirrors."{{@key}}"]
-endpoints = [{{join_array ", " this }}]
+[mirrors."{{registry}}"]
+endpoints = [{{join_array ", " endpoint }}]
 {{/each}}
 {{/if}}
--- a/sources/api/migration/migrations/v1.4.0/registry-mirror-representation/src/main.rs
+++ b/sources/api/migration/migrations/v1.4.0/registry-mirror-representation/src/main.rs
@@ -1,0 +1,133 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::{migrate, Migration, MigrationData, Result};
+use serde_json::{Map, Value};
+use std::collections::HashMap;
+use std::process;
+
+const MIRRORS_SETTING_NAME: &'static str = "settings.container-registry.mirrors";
+const DATASTORE_KEY_SEPARATOR: char = '.';
+
+/// This migration changes the model type of `settings.container-registry.mirrors` from `HashMap<SingleLineString, Vec<Url>>`
+/// to `Vec<RegistryMirrors>` on upgrade and vice-versa on downgrades.
+pub struct ChangeRegistryMirrorsType;
+
+// Snapshot of the `datastore::Key::valid_character` method in Bottlerocket version 1.3.0
+//
+// Determines whether a character is acceptable within a segment of a key name.  This is
+// separate from quoting; if a character isn't valid, it isn't valid quoted, either.
+fn valid_character(c: char) -> bool {
+    match c {
+        'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '-' | '/' => true,
+        _ => false,
+    }
+}
+
+impl Migration for ChangeRegistryMirrorsType {
+    /// Newer versions store `settings.container-registry.mirrors` as `Vec<RegistryMirrors>`.
+    /// Need to convert from `HashMap<SingleLineString, Vec<Url>>`.
+    fn forward(&mut self, mut input: MigrationData) -> Result<MigrationData> {
+        let mirrors: HashMap<_, _> = input
+            .data
+            .iter()
+            .filter(|&(k, _)| k.starts_with(format!("{}.", MIRRORS_SETTING_NAME).as_str()))
+            .map(|(k, v)| (k.to_owned(), v.to_owned()))
+            .collect();
+        let mut new_mirrors = Vec::new();
+        for (setting, endpoint) in mirrors {
+            // Get the registry name from the settings name. Trim any quotes the settings name might have.
+            let registry = setting
+                .strip_prefix(&format!("{}.", MIRRORS_SETTING_NAME))
+                .unwrap_or_default()
+                .trim_matches('"');
+            let mut registry_mirrors = Map::new();
+            registry_mirrors.insert("registry".to_string(), Value::String(registry.to_string()));
+            registry_mirrors.insert("endpoint".to_string(), endpoint.to_owned());
+            new_mirrors.push(Value::Object(registry_mirrors));
+            if let Some(data) = input.data.remove(&setting) {
+                println!("Removed setting '{}', which was set to '{}'", setting, data);
+            }
+        }
+        let data = Value::Array(new_mirrors);
+        println!(
+            "Creating new setting '{}', which is set to '{}'",
+            MIRRORS_SETTING_NAME, &data
+        );
+        input.data.insert(MIRRORS_SETTING_NAME.to_string(), data);
+        Ok(input)
+    }
+
+    /// Older versions store `settings.container-registry.mirrors` as `HashMap<SingleLineString, Vec<Url>>`.
+    /// Need to convert from `Vec<RegistryMirrors>`.
+    fn backward(&mut self, mut input: MigrationData) -> Result<MigrationData> {
+        if let Some(data) = input.data.get_mut(MIRRORS_SETTING_NAME).cloned() {
+            match data {
+                Value::Array(arr) => {
+                    if let Some(data) = input.data.remove(MIRRORS_SETTING_NAME) {
+                        println!(
+                            "Removed setting '{}', which was set to '{}'",
+                            MIRRORS_SETTING_NAME, data
+                        );
+                    }
+                    for obj in arr {
+                        if let Some(obj) = obj.as_object() {
+                            if let (Some(registry), Some(endpoint)) = (
+                                obj.get("registry").and_then(|s| s.as_str()),
+                                obj.get("endpoint"),
+                            ) {
+                                // Ensure the registry contains valid datastore key characters.
+                                // If we encounter any invalid key characters, we skip writing out
+                                // the setting key to prevent breakage of the datastore.
+                                if registry
+                                    .chars()
+                                    .all(|c| valid_character(c) || c == DATASTORE_KEY_SEPARATOR)
+                                {
+                                    let setting_name =
+                                        format!(r#"{}."{}""#, MIRRORS_SETTING_NAME, registry);
+                                    println!(
+                                        "Creating new setting '{}', which is set to '{}'",
+                                        setting_name, &endpoint
+                                    );
+                                    input.data.insert(setting_name, endpoint.to_owned());
+                                } else {
+                                    eprintln!(
+                                        "Container registry '{}' contains invalid datastore key character(s). Skipping to prevent datastore breakage...",
+                                        registry
+                                    );
+                                }
+                            }
+                        } else {
+                            println!(
+                                "'{}' contains non-JSON Object value: '{}'.",
+                                MIRRORS_SETTING_NAME, obj
+                            );
+                        }
+                    }
+                }
+                _ => {
+                    println!(
+                        "'{}' is not a JSON Array value: '{}'.",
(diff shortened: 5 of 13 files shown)


sources/api/migration/migrations/v1.3.1/registry-mirror-representation/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn forward(&mut self, mut input: MigrationData) -> Result<MigrationData> {
+        let mirrors: HashMap<_, _> = input
+            .data
+            .iter()
+            .filter(|&(k, _)| k.starts_with(format!("{}.", MIRRORS_SETTING_NAME).as_str()))
+            .map(|(k, v)| (k.to_owned(), v.to_owned()))
+            .collect();
+        let mut new_mirrors = Vec::new();
+        for (setting, endpoint) in mirrors {
+            println!("{}", setting);
```
etungsten: leftover debug print statement.

sources/models/src/lib.rs
```diff
@@ -179,10 +181,83 @@ struct ECSSettings {
     enable_spot_instance_draining: bool,
 }
 
+#[model]
+struct RegistryMirrors {
```
etungsten: Should probably be singular

README.md
```diff
@@ -411,13 +411,18 @@ These settings can be changed at any time.
...
+* `settings.container-registry.mirrors`: An array of container image registry mirror settings.  Each element specifies the registry and the endpoints for said registry.
+When pulling an image from a registry, the container runtime will try the endpoints one by one and use the first working one.
   (Docker and containerd will still try the default registry URL if the mirrors fail.)
   * Example user data for setting up image registry mirrors:
   ```
-  [settings.container-registry.mirrors]
-  "docker.io" = ["https://<my-docker-hub-mirror-host>"]
-  "gcr.io" = ["https://<my-gcr-mirror-host>","http://<my-gcr-mirror-host-2>"]
+  [[settings.container-registry.mirrors]]
+  registry = "*"
```
tjkirch, line 420: The `*` worked in testing, right?

I assume if you have a `*` entry and then a specific entry, the runtime tries all endpoints from both lists if the specific one matches?

Release.toml
```diff
@@ -74,3 +74,6 @@ version = "1.3.0"
     "migrate_v1.3.0_hostname-affects-etc-hosts.lz4",
     "migrate_v1.3.0_control-container-v0-5-2.lz4",
 ]
+"(1.3.0, 1.3.1)" = [
```
tjkirch:
```suggestion
"(1.3.0, 1.4.0)" = [
```
(and associated changes)

It seems we've settled on minor version bumps unless it's a security release, which wouldn't include this change.

packages/containerd/containerd-config-toml_k8s
```diff
@@ -32,7 +32,7 @@ conf_dir = "/etc/cni/net.d"
```
tjkirch, line 32: Does `packages/docker-engine/daemon-json` need to change as well?  It's trying to `join_array ", " settings.container-registry.mirrors.[docker.io]` but if it doesn't know about the endpoint key it seems like it'd fail.  Either way, can we show it in the testing?

packages/os/host-ctr-toml
```diff
@@ -1,6 +1,6 @@
 {{#if settings.container-registry.mirrors}}
```
tjkirch, line 1: Would you mind showing this file in the testing too?

sources/api/migration/migrations/v1.3.1/registry-mirror-representation/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+use migration_helpers::{migrate, Migration, MigrationData, Result};
+use serde_json::{Map, Value};
+use std::collections::HashMap;
+use std::process;
+
+const MIRRORS_SETTING_NAME: &'static str = "settings.container-registry.mirrors";
+
+/// This migration changes the model type of `settings.container-registry.mirrors` from `HashMap<SingleLineString, Vec<Url>>`
+/// to `Vec<RegistryMirrors>` on upgrade and vice-versa on downgrades
+/// .
```
tjkirch:
```suggestion
/// to `Vec<RegistryMirrors>` on upgrade and vice-versa on downgrade.
```

sources/models/src/lib.rs
```diff
@@ -179,10 +181,83 @@ struct ECSSettings {
...
 #[model]
 struct RegistrySettings {
-    mirrors: HashMap<SingleLineString, Vec<Url>>,
+    #[serde(deserialize_with = "deserialize_mirrors")]
+    mirrors: Vec<RegistryMirrors>,
+}
+
+// We want to allow two possible representations of registry mirrors. Either as a `Vec` of `RegistryMirrors`
+// or a `HashMap` of registry to endpoints.
+fn deserialize_mirrors<'de, D>(deserializer: D) -> Result<Option<Vec<RegistryMirrors>>, D::Error>
```
tjkirch: I think it's useful to keep `models/src/lib.rs` quick to reference and understand, with only basic types, and using modules for more complex or shared types - can we move this to a module?

sources/models/src/lib.rs
```diff
@@ -179,10 +181,83 @@ struct ECSSettings {
...
+        fn expecting(&self, formatter: &mut Formatter) -> std::fmt::Result {
+            formatter.write_str("TOML array or TOML table")
+        }
+
+        fn visit_seq<A>(self, seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: SeqAccess<'de>,
+        {
+            Ok(Some(Deserialize::deserialize(
+                serde::de::value::SeqAccessDeserializer::new(seq),
```
tjkirch: nit: put this in the `use` like the others?

tjkirch (commented):
Nice, this seems like a friendlier solution.

README.md
etungsten: Yes `*` worked in testing.
I believe so. Containerd would just iterate down the list of endpoints until it manages to pull the image.

packages/containerd/containerd-config-toml_k8s
etungsten: I forgot to test ECS variants :(

sources/models/src/de.rs
```diff
@@ -0,0 +1,73 @@
+use crate::RegistryMirror;
+use serde::de::value::SeqAccessDeserializer;
+use serde::de::{MapAccess, SeqAccess, Visitor};
+use serde::{Deserialize, Deserializer};
+use std::fmt::Formatter;
+
+// We want to allow two possible representations of registry mirrors. Either as a `Vec` of `RegistryMirror`
+// or a `HashMap` of registry to endpoints.
```
tjkirch:
```suggestion
// Our standard representation of registry mirrors is a `Vec` of `RegistryMirror`; for backward compatibility, we also allow a `HashMap` of registry to endpoints.
```
(I wasn't sure where to break the line, but it's probably useful to make clear that we don't encourage the old format.)

tjkirch (commented):
LGTM assuming ECS testing is OK; couple small items remain above.

etungsten:
Pushes above addresses @tjkirch 's comments. Adds more testing to the PR description.

packages/os/host-ctr-toml
etungsten: Updated PR description with how `host-ctr.toml` renders for host-containers services.

etungsten:
Push above removes `snafu` as a dependency from the migration.

tjkirch (dismissed):

jpculp (dismissed):

zmrow (dismissed):
:nut_and_bolt: 

Wow nice work!

sources/api/migration/migrations/v1.4.0/registry-mirror-representation/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+                                obj.get("registry").and_then(|s| s.as_str()),
+                                obj.get("endpoint"),
+                            ) {
+                                let setting_name = format!("{}.{}", MIRRORS_SETTING_NAME, registry);
+                                println!(
+                                    "Creating new setting '{}', which is set to '{}'",
+                                    setting_name, &endpoint
+                                );
+                                input.data.insert(setting_name, endpoint.to_owned());
+                            }
```
bcressey: Seems like we also need to check that the `SingleLineString` registry fields are valid keys under the old datastore rules, where `*` wasn't allowed, or else we'll either write an invalid datastore or break during the backward migration.

etungsten:
Sorry about the repeat force pushes. Please take a look at the backward and forward migrations for the new changes.

Pushes above addresses @bcressey 's comment. I also realized cases where the migration would fail if the registry contained `.` in the name or if they were quoted so I had to adjust the forward migration to fix it. Tested the migrations fully and they work as expected.

Updated the PR description with updated migration testing info.

bcressey (approved):

sources/api/migration/migrations/v1.4.0/registry-mirror-representation/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+            .iter()
+            .filter(|&(k, _)| k.starts_with(format!("{}.", MIRRORS_SETTING_NAME).as_str()))
+            .map(|(k, v)| (k.to_owned(), v.to_owned()))
+            .collect();
+        let mut new_mirrors = Vec::new();
+        for (setting, endpoint) in mirrors {
+            // Need to split the full setting name to get the registry name.
+            // We get the index of the third '.' and split the string around there and leave out the '.'.
+            // Trim quotation marks around the registry name.
+            let registry = setting
```
tjkirch, line 39: I think this block can use `strip_prefix` instead to simplify?  (Probably still need to trim quotes, otherwise migrating forward and backward repeatedly would add quotes forever, I think.)

sources/api/migration/migrations/v1.4.0/registry-mirror-representation/src/main.rs
```diff
@@ -0,0 +1,126 @@
+#![deny(rust_2018_idioms)]
+
+use datastore::{Key, KEY_SEPARATOR};
```
tjkirch: We should avoid linking to datastore (or much of anything API-related) in migrations - the code needs to run before and after a version upgrade, during which time the data store code can change, and it makes thinking about the change confusing; what am I linked against?  when am I running?  etc.

I can't think of a better option than copying the valid_character and KEY_SEPARATOR code from datastore into this migration for now.  We do need to know any keys that would be invalid on downgrade so the downgrade doesn't fail; it's better to throw away those settings than keep the user on a new version that's breaking them in other ways.

sources/api/migration/migrations/v1.4.0/registry-mirror-representation/src/main.rs
```diff
@@ -0,0 +1,126 @@
...
+                        if let Some(obj) = obj.as_object() {
+                            if let (Some(registry), Some(endpoint)) = (
+                                obj.get("registry").and_then(|s| s.as_str()),
+                                obj.get("endpoint"),
+                            ) {
+                                // Ensure the registry contains valid datastore key characters.
+                                // If we encounter any invalid key characters, we skip writing out
+                                // the setting key to prevent breakage of the datastore.
+                                for c in registry.chars() {
+                                    if !Key::valid_character(c) && c != KEY_SEPARATOR {
```
tjkirch: Rather than the `for` and needing a label, I think you can use something like `registry.chars().all(|c| valid_character(c) || c == KEY_SEPARATOR)`?

etungsten:
push above addresses @tjkirch 's comments. I retested the migrations and they worked as expected. `registry` set to `*` gets dropped on downgrade:
```
...
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."b"]
endpoint = ["cccc", "dddd"]
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."*"]
endpoint = ["aaaa", "bbb"]
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
endpoint = ["https://registry-0.acme.com", "https://registry-1.acme.com"]
bash-5.0# signpost rollback-to-inactive
bash-5.0# reboot
...
```
On next boot:

```
...
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."b"]
endpoint = ["cccc", "dddd"]
[plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
endpoint = ["https://registry-0.acme.com", "https://registry-1.acme.com"]

```

etungsten:
Push above removes an outdated comment and a redundant `to_string()` call

tjkirch (approved):

jpculp (approved):

bcressey (approved):

zmrow (approved):
:socks:

PR #1801, opened 2021-11-02
imdsclient: add retry to fetch_token

jpculp:
**Description of changes:**

This adds retry logic if `imdsclient` fails to fetch a session token.

Also includes small changes to the `fetch_imds` retry logic, such as moving
the sleep after the ensure `imds_attempts <= max_imds_attempts` and bumping
the wait time from 100ms to 1s.

**Testing done:**

- Built `aws-k8s-1.20` ami and launched instance.

**Terms of contribution:**

### DIFF
--- a/sources/imdsclient/src/lib.rs
+++ b/sources/imdsclient/src/lib.rs
@@ -225,10 +225,10 @@ impl ImdsClient {
         let max_attempts: u8 = 3;
         loop {
             attempt += 1;
+            ensure!(attempt <= max_attempts, error::FailedFetchIMDS { attempt });
             if attempt > 1 {
-                time::sleep(Duration::from_millis(100)).await;
+                time::sleep(Duration::from_secs(1)).await;
             }
-            ensure!(attempt <= max_attempts, error::FailedFetch { attempt });
             let response = self
                 .client
                 .get(&uri)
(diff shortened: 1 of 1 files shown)


sources/imdsclient/src/lib.rs
```diff
@@ -351,23 +354,39 @@ fn build_public_key_targets(public_key_list: &str) -> Vec<String> {
...
+    let max_token_attempts: u8 = 2;
+    loop {
+        token_attempts += 1;
+        ensure!(
+            token_attempts <= max_token_attempts,
+            error::FailedFetchToken { token_attempts }
+        );
+        if token_attempts > 1 {
+            time::sleep(Duration::from_secs(5)).await;
+        }
```
bcressey: I'd like to align the number of retries and the length of the delay with `fetch_imds`  - the requests are going to the same service and reaching the failure state in either function will likely result in a boot failure.

zmrow:
Once @bcressey 's concern is handled, lgtm

sources/imdsclient/src/lib.rs
jpculp: The retries in `fetch_imds` are mostly to trigger token refreshes in the even the token is expired as opposed to IMDS unavailability (hence a shorter sleep). Each client initialization and `fetch_imds` attempt is going to fetch a new token. If IMDS is unreachable for whatever reason, the failure would occur during the `fetch_token`. Since we are adding retries to the token fetch now it might make sense to drop the max_imds_attempts from 3 to 2, but I wouldn't recommend aligning the sleep duration.

jpculp:
- Replaced variables like `imds_attempts` with `attempt`.
- Increased the `fetch_token` max attempts from 2 to 3. Now both `fetch_token` and `fetch_imds` have 3 attempts.
- Changed wording for `fetch_token` error message to be more descriptive.

sources/imdsclient/src/lib.rs
jpculp: Instead of dropping `fetch_imds` from 3 to 2, I bumped `fetch_token` from 2 to 3.

etungsten (approved):

webern (approved):

bcressey (approved):

PR #1809, opened 2021-11-09
add fixes to improve boot speed

bcressey:
**Issue number:**
N/A

**Description of changes:**
This is a collection of fixes to improve boot speed and time to a usable node - at least by 5 seconds, at most by 8 seconds.

Building `kubelet` with the "dockerless" tag saves 5 seconds during service startup, as otherwise cadvisor tries for five seconds to connect to the Docker daemon before printing an error.

The fix for the defer timeout in the wicked DHCPv6 client saves 1 second for around half of launches, in cases where the timer fires a little early and would otherwise trigger another 1 second wait.

Using an overlayfs for the CNI plugin directory saves a variable amount of time by avoiding a potentially slow copy to an unwritten EBS volume in the critical path. `systemd-tmpfiles-setup` previously took 900 milliseconds or more in most cases, and now takes 100 milliseconds or less, with most of the remaining time spent populating the SELinux modules in `/var/lib/selinux`.

Building support for the PS/2 controller, keyboard, and mouse as modules saves around 400 milliseconds during boot under KVM, as otherwise device mapper waits for the configuration to finish before mounting the root filesystem. They are still loaded later, after the root filesystem is mounted, but at that point we can do more work in parallel.

Disabling RAID auto-detect avoids another potential device wait and reduces `printk` messages. Writing to the console device at 115200 bits per second speeds up those operations by 12x. Console logging continues to be a drag on overall boot speed. We can turn it off altogether to gain at least 2 seconds, but only at a severe cost to debugging capabilities if anything goes wrong. Using the higher device speed obviously helps, but its impact is spread across all threads that might draw the short straw after triggering a `printk` call, and is difficult to quantify.

Removing the `udevadm settle` dependency doesn't yield a measurable improvement in boot speed, but does stop `systemd` from blaming `wicked` for slowing everything down.

I've kept the two commits that added debug output for `systemd-tmpfiles` and the wicked clients, since these were instrumental in identifying the underlying issues and confirming the fixes. These logs are all sent to the journal rather than the console, so they don't compete with existing output or slow down the boot.

**Testing done:**
For the kernel change: verified that the keyboard and mouse modules were still loaded on x86_64 nodes.

For the changes to kubelet and the CNI plugins directory: verified that sonobuoy runs passed for these versions, and that no Docker related error messages were logged to the journal.

For the "activate" targets: confirmed that these were no longer blamed by `systemd-analyze blame`, and that bootstrap containers still worked as expected.

For the wicked changes: confirmed that the DHCP6 client would defer after the first timeout, whether the timer fired slightly before or slightly after one second elapsed. On instances with DHCP6 enabled, the lease was successfully acquired.

For the udev settle change: used a hacked up local build where wicked was set up to manage "eth1" rather than "eth0", and verified that wicked would still configure the device if I renamed it into existence during the wait.

For the serial console changes: verified that console logs were present for AWS variants across a range of instance types - c1.xlarge, t2.large, m3.2xlarge, c3.large, c4.large, c5.large, c6g.large - and for VMware variants running on ESXi 7.0. Note that we're already using 115200 for GRUB as of #1701, so this setting has previously been validated on a smaller set of instance types.

**Terms of contribution:**

### DIFF
--- a/packages/release/opt-cni-bin.mount
+++ b/packages/release/opt-cni-bin.mount
@@ -1,0 +1,14 @@
+[Unit]
+Description=CNI Plugin Directory (/opt/cni/bin)
+Conflicts=umount.target
+RequiresMountsFor=/var
+Before=local-fs.target umount.target
+
+[Mount]
+What=overlay
+Where=/opt/cni/bin
+Type=overlay
+Options=noatime,nosuid,nodev,lowerdir=/usr/libexec/cni/bin,upperdir=/var/lib/cni-plugins/.overlay/upper,workdir=/var/lib/cni-plugins/.overlay/work,context=system_u:object_r:local_t:s0
+
+[Install]
+WantedBy=preconfigured.target
--- a/packages/cni-plugins/cni-plugins.spec
+++ b/packages/cni-plugins/cni-plugins.spec
@@ -32,8 +32,8 @@ for d in $(find plugins -mindepth 2 -maxdepth 2 -type d ! -name windows) ; do
 done
 
 %install
-install -d %{buildroot}%{_cross_factorydir}/opt/cni/bin
-install -p -m 0755 bin/* %{buildroot}%{_cross_factorydir}/opt/cni/bin
+install -d %{buildroot}%{_cross_libexecdir}/cni/bin
+install -p -m 0755 bin/* %{buildroot}%{_cross_libexecdir}/cni/bin
 
 %cross_scan_attribution go-vendor vendor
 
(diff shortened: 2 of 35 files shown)


packages/release/opt-cni-bin.mount
```diff
@@ -0,0 +1,14 @@
...
+Description=CNI Plugin Directory (/opt/cni/bin)
+Conflicts=umount.target
+RequiresMountsFor=/opt
+Before=local-fs.target umount.target
+
+[Mount]
+What=overlay
+Where=/opt/cni/bin
+Type=overlay
+Options=noatime,nosuid,nodev,lowerdir=/usr/libexec/cni/bin,upperdir=/opt/cni/upper,workdir=/opt/cni/work,context=system_u:object_r:local_t:s0
```
samuelkarp: Would it be worth adding another directory here to contain the upperdir and workdir to hide them/make it more obvious that they're an implementation detail of the overlay mount?  Something like `/opt/cni/.overlay/upper` and `/opt/cni/.overlay/work`?

samuelkarp (approved):
One super minor suggestion, but otherwise LGTM!

tjkirch (approved):

bcressey:
Rebase; fix the serial console speed commit to account for the removed `aws-k8s-1.17` variant.

packages/release/opt-cni-bin.mount
bcressey: I opted to use `/var/lib/cni-plugins` for the overlay directories, partly to match the treatment of `/var/lib/kernel-devel` (which I also adjusted along these lines), and partly to guard against cases where pods might be mounting in `/opt/cni` and get confused by the new directories.

bcressey:
Adjust overlay directory handling per @samuelkarp

tjkirch (approved):

arnaldo2792 (approved):

samuelkarp (approved):
LGTM

Issue #1743, opened 2021-09-13
Update x509-parser / der-parser


tjkirch:
x509-parser updated its dependency on der-parser to 5.1.2, but we lock to 5.1.0 because of breaking changes in 5.1.1.  (See #1687.)  We need to investigate whether/how we can update x509-parser and/or der-parser.

Issue #1620, opened 2021-06-17
Update gptman dependency


tjkirch:
There have been some big updates to gptman.  It's a critical dependency for us in a few places, so we should carefully evaluate the changes.

Issue #1471, opened 2021-04-08
Build does not work with hyper versions >= 0.14.3


webern:
We build most Bottlerocket binaries with `RUSTFLAGS=-Cprefer-dynamic` so that we can dynamically link to the Rust std library and reduce the size of our binaries.

Hyper 0.14.3 adds this line to its Cargo.toml to emit a c interface in a dynamic library:
https://github.com/hyperium/hyper/blob/48d4594930da4e227039cfa254411b85c98b63c5/Cargo.toml#L23

This, in conjunction with `reqwest` and `prefer-dynamic` causes cargo/rustc to try to link using the emitted hyper c library and the linker fails with missing symbols. The problem is described precisely here: https://github.com/rust-lang/rust/issues/82151#issuecomment-779368017

As it stands right now, this means that the highest version of hyper we can use with Bottlerocket is 0.14.2.

webern:
I want to link this to the underlying issue, which is this: https://github.com/rust-lang/rust/issues/82151

And specifically this comment explains it well:
https://github.com/rust-lang/rust/issues/82151#issuecomment-779368017

webern:
Trying again because the backlink was not created
https://github.com/rust-lang/rust/issues/82151

webern:
Note, this PR attempts to fix the underlying issue, but appears to be stuck https://github.com/rust-lang/rust/pull/88101

webern:
Good news!
This RFC was has been created, accepted and implemented in nightly: https://rust-lang.github.io/rfcs/3180-cargo-cli-crate-type.html

@seanmonstar reports that this will allow the removal of cdylib from Hyper's Cargo.toml and allow builds to instead opt-in to the creation of the cdylib. Once that change to Hyper's Cargo.toml is released, we should be able to update.

webern:
It will be ready soon! https://github.com/hyperium/hyper/commit/1c6637060e36654ddb2fdfccb0d146c7ad527476

webern:
Fixed by #2019

Issue #1776, opened 2021-10-13
build: check clippy and cargo fmt during github actions checks


webern:
**What I'd like:**

During GitHub actions CI, the build should fail (preferably early) if the code is not fmt'ed with cargo fmt and if it has cargo clippy warnings.

We probably wouldn't want this to happen as part of the RPM build process. Instead this would probably be more like what we do now with `cargo test`. The downside is that a mismatch between the version of cargo in the build runners and a developer's local version of cargo could cause a mismatch in clippy lints (less likely that a fmt mismatch would occur).

**Any alternatives you've considered:**

- Doing this with the SDK
- Doing this with "local" cargo/rust
- Don't check lint/fmt in CI

webern:
The simplest solution might be to add these under the existing target `cargo make test`. (Note that `cargo make test` only runs `sources` tests currently. We should probably expand the `cargo make test` target to encompass both `sources` and `tools`)

jpmcb:
Closing since this was accomplished earlier this year ✅ 

https://github.com/bottlerocket-os/bottlerocket/blob/4885e9a8502031d51f1215b51a775f738dc948ef/.github/workflows/build.yml#L118-L120

PR #1810, opened 2021-11-09
Update Rust dependencies

cbgbt:
**Description of changes:**

```
tools: update tough to 0.12
tools: cargo update

sources: update nix to 0.23
sources: update tough to 0.12
sources: update to actix-web-4.0.0-beta.10
sources: cargo update
```
Not updated: [gptman](https://github.com/bottlerocket-os/bottlerocket/issues/1620), [hyper](https://github.com/bottlerocket-os/bottlerocket/issues/1471) (in `sources/`), [x509-parser](https://github.com/bottlerocket-os/bottlerocket/issues/1743)
webpki (must match the version used by reqwest)

**Testing done:**

- [x] Unit tests pass.
- [x] Built AWS k8s 1.21 for aarch64 and ECS AMIs for x86_64. Confirmed `systemctl status` `running`, dmesg and journal OK. Ran a pod/task OK. Made API changes OK.
- [x] Ensured that handlebars templating changes don't cause semantic changes to rendered templates (thanks @tjkirch)
- [x] Confirmed that `cargo make repo` and `cargo make ssm` still work OK.
- [x] Test `apiclient exec` with the new `nix` change

**Terms of contribution:**

### DIFF
--- a/sources/api/migration/migrator/src/main.rs
+++ b/sources/api/migration/migrator/src/main.rs
@@ -32,6 +32,7 @@ use semver::Version;
 use simplelog::{Config as LogConfig, SimpleLogger};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::HashSet;
+use std::convert::TryInto;
 use std::env;
 use std::fs::{self, File};
 use std::os::unix::fs::symlink;
@@ -243,14 +244,24 @@ where
 
     for migration in migrations {
         let migration = migration.as_ref();
+        let migration = migration
+            .try_into()
+            .context(error::TargetName { target: migration })?;
+
         // get the migration from the repo
         let lz4_bytes = repository
-            .read_target(migration)
-            .context(error::LoadMigration { migration })?
-            .context(error::MigrationNotFound { migration })?;
+            .read_target(&migration)
+            .context(error::LoadMigration {
+                migration: migration.raw(),
+            })?
+            .context(error::MigrationNotFound {
+                migration: migration.raw(),
+            })?;
 
         // Add an LZ4 decoder so the bytes will be deflated on read
-        let mut reader = lz4::Decoder::new(lz4_bytes).context(error::Lz4Decode { migration })?;
+        let mut reader = lz4::Decoder::new(lz4_bytes).context(error::Lz4Decode {
+            migration: migration.raw(),
+        })?;
 
         // Create a sealed command with pentacle, so we can run the verified bytes from memory
         let mut command =
@@ -480,9 +491,10 @@ where
 
 fn load_manifest(repository: &tough::Repository) -> Result<Manifest> {
     let target = "manifest.json";
+    let target = target.try_into().context(error::TargetName { target })?;
     Manifest::from_json(
         repository
-            .read_target(target)
+            .read_target(&target)
             .context(error::ManifestLoad)?
             .context(error::ManifestNotFound)?,
     )
--- a/sources/deny.toml
+++ b/sources/deny.toml
@@ -52,8 +52,8 @@ skip = [
     # older version used by hyper 0.14.2
     { name = "socket2", version = "0.3.19" },
 
-    # older version used by tough 0.11.2
-    { name = "serde_plain", version = "0.3.0" },
+    # older version used by webpki-roots 0.21.1
+    { name = "pem", version = "0.8.3" },
 ]
 
 skip-tree = [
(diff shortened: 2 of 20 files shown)


sources/api/migration/migrator/src/main.rs
```diff
@@ -479,10 +490,10 @@ where
 }
 
 fn load_manifest(repository: &tough::Repository) -> Result<Manifest> {
-    let target = "manifest.json";
+    let target = "manifest.json".try_into().unwrap();
```
webern: Did you forget to use an error variant here?

sources/deny.toml
```diff
@@ -52,8 +52,11 @@ skip = [
     # older version used by hyper 0.14.2
     { name = "socket2", version = "0.3.19" },
 
-    # older version used by tough 0.11.2
-    { name = "serde_plain", version = "0.3.0" },
+    # older version used by webpki-roots 0.21.1
```
webern, line 55: I looked at the lock file and I don't see a change to the webpki-roots, which is a good thing because of https://github.com/bottlerocket-os/bottlerocket/blob/5a083d917e0c1101bec5a8668f5a5a8048239454/sources/webpki-roots-shim/Cargo.toml#L3..L11

sources/updater/updog/src/main.rs
```diff
@@ -583,10 +590,10 @@ fn main_inner() -> Result<()> {
 }
 
 fn load_manifest(repository: &tough::Repository) -> Result<Manifest> {
-    let target = "manifest.json";
+    let target = "manifest.json".try_into().unwrap();
```
webern: Another error variant is needed here.

Also I might suggest using the type name so a reader can see what type is involved.

```rust
let target = TargetName::new("manifest.json").context(error::InvalidTargetName{ blah blah }?;
```

tools/pubsys/src/repo.rs
```diff
@@ -352,10 +352,11 @@ where
     match repo_load_result {
         // If we load it successfully, build an editor and manifest from it.
         Ok(repo) => {
+            let target = "manifest.json".try_into().unwrap();
```
webern: Same as other locations.

BTW, if we get clippy going in Bottlerocket CI https://github.com/bottlerocket-os/bottlerocket/issues/1776 then we can add this to make sure we don't panic anywhere:

```rust
#![deny(
    clippy::expect_used,
    clippy::get_unwrap,
    clippy::panic,
    clippy::panic_in_result_fn,
    clippy::panicking_unwrap,
    clippy::unwrap_in_result,
    clippy::unwrap_used
)]
```

sources/api/migration/migrator/src/main.rs
cbgbt: Because `"manifest.json"` is a constant, we know it can't possibly return an error here, so we just went with `unwrap()`. Should apply to any other addition of unwrap in this PR.

sources/deny.toml
cbgbt: Yeah. In this case, one of our other deps (I believe `actix-web`) moved forward to use a newer version of `pem`. Our cargo-deny config flags differences in dependency versions, but we can't change webpki-roots due to what you've linked.

tjkirch (commented):
LGTM if the remaining testing passes.

tjkirch (approved):

tools/pubsys/src/repo.rs
tjkirch: I wonder if tough could have const checks for static strings, so we could construct things like this without runtime error checking...

sources/api/migration/migrator/src/main.rs
cbgbt: I ended up adding `context()` calls to these constant unwraps, which should make clippy lints more meaningful, which I think is worth it.

cbgbt:
The previous two force pushes:
* Added the `context()` calls where previously `unwrap()` was being used
* Rebased atop the latest code in the `develop` branch.

Based on test failures, looks like I need to modify the `nix` change to also update for `apiclient exec`, which was added in the rebase. I'll get a fix ready.

cbgbt:
I've pushed just a change to move `apiclient` to `nix 0.23`. I tested `apiclient exec` with the changes.

tjkirch (approved):
LGTM assuming checks pass!

webern (approved):

jpculp (approved):

PR #1826, opened 2021-11-15
add support for custom image sizes

### LINKED ISSUE #1821: Alternate image layout in rpm2img
**What I'd like:**
As part of #1799, we will add support for NVIDIA GPUS. However, additional libraries provided by the `.run` script are required for all GPU workloads to work. The space required for the libraries is around ~432M, which causes build errors because the image size isn't sufficient. Since we don't want to increase the size of the image for all variants, we need to change `rpm2img` so that we can have alternate image layouts, depending on the variant being built.

N / A

bcressey:
**Issue number:**
Fixes #1821

**Description of changes:**
Refactor partition layout code into a helper library, and add support for custom OS and data image sizes in the variant specification.

Fix a bug that affected the last partition on each disk, where it was one sector larger than intended. (The corresponding filesystems were already the correct size.)

**Testing done:**
- [x] verify that the partition offsets are the same (modulo fix) for the default layout
- [x] verify that an undersized hash partition is rejected by the new check
- [x] verify that in-place upgrade works for the default layout, since this touches the GPT priority bits
- [x] verify that in-place upgrade works for custom layouts with a larger OS image

Confirmed that the "old" and "new" layouts only differ by one sector in the last partition.

**Old layout**
```
Sector size (logical): 512 bytes
Disk identifier (GUID): 309D0119-AA57-4863-B5DE-37833FEE0E97
Partition table holds up to 128 entries
Main partition table begins at sector 2 and ends at sector 33
First usable sector is 34, last usable sector is 4194270
Partitions will be aligned on 2048-sector boundaries
Total free space is 4028 sectors (2.0 MiB)

Number  Start (sector)    End (sector)  Size       Code  Name
   1            2048           10239   4.0 MiB     EF02  BIOS-BOOT
   2           10240           20479   5.0 MiB     EF00  EFI-SYSTEM
   3           20480          102399   40.0 MiB    FFFF  BOTTLEROCKET-BOOT-A
   4          102400         1986559   920.0 MiB   FFFF  BOTTLEROCKET-ROOT-A
   5         1986560         2007039   10.0 MiB    FFFF  BOTTLEROCKET-HASH-A
   6         2007040         2058239   25.0 MiB    FFFF  BOTTLEROCKET-RESERV...
   7         2058240         2068479   5.0 MiB     FFFF  EFI-BACKUP
   8         2068480         2150399   40.0 MiB    FFFF  BOTTLEROCKET-BOOT-B
   9         2150400         4034559   920.0 MiB   FFFF  BOTTLEROCKET-ROOT-B
  10         4034560         4055039   10.0 MiB    FFFF  BOTTLEROCKET-HASH-B
  11         4055040         4106239   25.0 MiB    FFFF  BOTTLEROCKET-RESERV...
  12         4106240         4192256   42.0 MiB    FFFF  BOTTLEROCKET-PRIVATE
```

**New layout**
```
Sector size (logical): 512 bytes
Disk identifier (GUID): 1754185D-33D2-4008-A1DA-FC84B274EFFB
Partition table holds up to 128 entries
Main partition table begins at sector 2 and ends at sector 33
First usable sector is 34, last usable sector is 4194270
Partitions will be aligned on 2048-sector boundaries
Total free space is 4029 sectors (2.0 MiB)

Number  Start (sector)    End (sector)  Size       Code  Name
   1            2048           10239   4.0 MiB     EF02  BIOS-BOOT
   2           10240           20479   5.0 MiB     EF00  EFI-SYSTEM
   3           20480          102399   40.0 MiB    FFFF  BOTTLEROCKET-BOOT-A
   4          102400         1986559   920.0 MiB   FFFF  BOTTLEROCKET-ROOT-A
   5         1986560         2007039   10.0 MiB    FFFF  BOTTLEROCKET-HASH-A
   6         2007040         2058239   25.0 MiB    FFFF  BOTTLEROCKET-RESERV...
   7         2058240         2068479   5.0 MiB     FFFF  EFI-BACKUP
   8         2068480         2150399   40.0 MiB    FFFF  BOTTLEROCKET-BOOT-B
   9         2150400         4034559   920.0 MiB   FFFF  BOTTLEROCKET-ROOT-B
  10         4034560         4055039   10.0 MiB    FFFF  BOTTLEROCKET-HASH-B
  11         4055040         4106239   25.0 MiB    FFFF  BOTTLEROCKET-RESERV...
  12         4106240         4192255   42.0 MiB    FFFF  BOTTLEROCKET-PRIVATE
```

Confirmed that an undersized hash partition is rejected if I reduce `HASH_SCALE_FACTOR` to `2`.
```
  #12 18.32 verity content is larger than partition (4M)
  #12 ERROR: executor failed running [/bin/sh -c /host/tools/rpm2img ...
```

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/manifest.rs
+++ b/tools/buildsys/src/manifest.rs
@@ -66,11 +66,30 @@ releases-url = "https://www.example.com/releases"
 included-packages = ["release"]
 ```
 
-`image-format` is the desired format of the built image.
+`image-format` is the desired format for the built images.
 This can be `raw` (the default), `vmdk`, or `qcow2`.
 ```
 [package.metadata.build-variant]
 image-format = "vmdk"
+```
+
+`image-layout` is the desired layout for the built images.
+
+`os-image-size-gib` is the desired size of the "os" disk image in GiB.
+The specified size will be automatically divided into two banks, where each
+bank contains the set of partitions needed for in-place upgrades. Roughly 40%
+will be available for each root filesystem partition, with the rest allocated
+to other essential system partitions.
+
+`data-image-size-gib` is the desired size of the "data" disk image in GiB.
+The full size will be used for the single data partition, except for the 2 MiB
+overhead for the GPT labels and partition alignment. The data partition will be
+automatically resized to fill the disk on boot, so it is usually not necessary
+to increase this value.
+```
+[package.metadata.build-variant.image-layout]
+os-image-size-gib = 2
+data-image-size-gib = 1
 ```
 
 `supported-arches` is the list of architectures the variant is able to run on.
@@ -100,6 +119,9 @@ use std::collections::HashSet;
 use std::fmt;
 use std::fs;
 use std::path::{Path, PathBuf};
+
+static DEFAULT_OS_IMAGE_SIZE_GIB: u32 = 2;
+static DEFAULT_DATA_IMAGE_SIZE_GIB: u32 = 1;
 
 /// The nested structures here are somewhat complex, but they make it trivial
 /// to deserialize the structure we expect to find in the manifest.
@@ -146,6 +168,11 @@ impl ManifestInfo {
     /// Convenience method to return the image format override, if any.
     pub(crate) fn image_format(&self) -> Option<&ImageFormat> {
         self.build_variant().and_then(|b| b.image_format.as_ref())
+    }
+
+    /// Convenience method to return the image layout, if specified.
+    pub(crate) fn image_layout(&self) -> Option<&ImageLayout> {
+        self.build_variant().and_then(|b| b.image_layout.as_ref())
     }
 
     /// Convenience method to return the supported architectures for this variant.
@@ -204,6 +231,7 @@ pub(crate) struct BuildPackage {
 pub(crate) struct BuildVariant {
     pub(crate) included_packages: Option<Vec<String>>,
     pub(crate) image_format: Option<ImageFormat>,
+    pub(crate) image_layout: Option<ImageLayout>,
     pub(crate) supported_arches: Option<HashSet<SupportedArch>>,
     pub(crate) kernel_parameters: Option<Vec<String>>,
 }
@@ -214,6 +242,34 @@ pub(crate) enum ImageFormat {
     Qcow2,
     Raw,
     Vmdk,
+}
+
+#[derive(Deserialize, Debug, Copy, Clone)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) struct ImageLayout {
+    #[serde(default = "ImageLayout::default_os_image_size_gib")]
+    pub(crate) os_image_size_gib: u32,
+    #[serde(default = "ImageLayout::default_data_image_size_gib")]
+    pub(crate) data_image_size_gib: u32,
+}
+
+impl ImageLayout {
+    fn default_os_image_size_gib() -> u32 {
+        DEFAULT_OS_IMAGE_SIZE_GIB
+    }
+
+    fn default_data_image_size_gib() -> u32 {
+        DEFAULT_DATA_IMAGE_SIZE_GIB
+    }
+}
+
+impl Default for ImageLayout {
+    fn default() -> Self {
+        Self {
+            os_image_size_gib: Self::default_os_image_size_gib(),
+            data_image_size_gib: Self::default_data_image_size_gib(),
+        }
+    }
 }
 
 #[derive(Deserialize, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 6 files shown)


tools/rpm2img
```diff
@@ -56,58 +61,44 @@ VERITY_HASH_ALGORITHM=sha256
...
+  # before the next partition starts. Specify the end point in sectors so we
+  # can subtract a sector to fix the off-by-one error that comes from adding
+  # start and size together. (1 MiB contains 2048 512-byte sectors.)
+  part_start=${partoff["${part}"]}
+  part_end=$((${part_start} + ${partsize["${part}"]}))
+  part_end=$((${part_end} * 2048 - 1))
+
+  partargs+=(-n 0:${part_start}M:${part_end})
+  partargs+=(-c 0:${partlabel["${part}"]})
+  partargs+=(-t 0:${parttype["${part}"]})
```
jpculp:
```suggestion
  partargs+=(-n 0:"${part_start}M:${part_end}")
  partargs+=(-c 0:"${partlabel["${part}"]}")
  partargs+=(-t 0:"${parttype["${part}"]}")
```
nit.

jpculp (commented):
This seems OK to me pending testing.

tools/buildsys/src/manifest.rs
```diff
@@ -216,6 +244,33 @@ pub(crate) enum ImageFormat {
...
+
+    fn default_data_image_size_gib() -> u32 {
+        DEFAULT_DATA_IMAGE_SIZE_GIB
+    }
+}
+
+impl Default for ImageLayout {
+    fn default() -> Self {
+        Self {
+            os_image_size_gib: Self::default_os_image_size_gib(),
```
zmrow, line 269: Out of curiosity - why not point these defaults directly at the constants?

zmrow (approved):
Whew nice work!  :waffle:

tools/rpm2img
tjkirch: (if you do quote, there's no reason to leave the other constant text out of the quotes, it's more confusing for new people to understand why you're introducing quotes for only some text)

tjkirch (commented):
I like it, it's just clever enough to get the flexibility we need without the logic being too hard to read.  The math involved, though, is tough and non-obvious - I think it'd help to have an ASCII/draw.io diagram of an example partition layout with an explanation of MB vs. MiB vs. GiB vs. sectors where appropriate.  The old `sgdisk` call at least had the partitions laid out next to each other, which helped a bit, even though it was hiding some secrets.

LGTM if upgrade/downgrade testing comes out OK.

tools/buildsys/src/manifest.rs
bcressey: AFAIK that's not supported by `serde` per https://github.com/serde-rs/serde/issues/368.

tools/rpm2img
bcressey: I'd prefer to skip the quotes here for consistency with the other uses of the dictionaries. I avoided quoting those expressions because the string keys also need to be in double-quotes, which introduces a fair amount of escaping overhead for relatively little gain.

tjkirch (approved):

tools/rpm2img
bcressey: @jpculp explained that this fixes a `shellcheck` warning, which makes sense to me.

bcressey:
Fix or selectively disable shellcheck warnings.

tools/rpm2img
```diff
@@ -3,6 +3,10 @@
 set -eu -o pipefail
 shopt -qs failglob
 
+# import the partition helper functions
+# shellcheck disable=SC1090 # ignore warning to avoid hardcoding path
```
tjkirch:
```suggestion
# shellcheck source=partyplanner
```
If you tell it where the file is you may get better tips.

tools/rpm2img
```diff
@@ -56,58 +62,45 @@ VERITY_HASH_ALGORITHM=sha256
...
+  # Boot partition attributes:
+  #  48 = gptprio priority bit
+  #  56 = gptprio successful bit
+  case "${part}" in
+    BOOT-A) partargs+=(-A 0:"set":48 -A 0:"set":56) ;;
+    BOOT-B) partargs+=(-A 0:"clear":48 -A 0:"clear":56) ;;
+  esac
+done
+
+# shellcheck disable=SC2086 # intentional word splitting
```
tjkirch: Would `"${partargs[@]}"` result in the same thing as `${partargs[*]}` here and pacify shellcheck?  (It's not a useful shellcheck warning here, but can be in general, because using the splitting behavior of `*` can result in weirdness if you change `IFS`.)

tjkirch (approved):

tjkirch (approved):

jpculp (approved):

zmrow (approved):
:kick_scooter:

PR #1841, opened 2021-11-24
imdsclient: better retries with tokio retry and timeout

jpculp:
**Description of changes:**

To mitigate things like closed connections, IMDSv2 token and data
fetches will continue to retry until success or until the timeout has
been reached.

The default timeout is 300s to match the ifup timeout in wicked, but can
be set manually using `.with_timeout` when creating a new ImdsClient.

Also moves the token fetch to just before the IMDS data fetch so
that building a new ImdsClient no longer needs an await.

`refresh_token` has been replaced by `clear_token` and `write_token`.

**Testing done:**
In addition to the unit testing...
- Built `aws-k8s-1.21` ami and launched instance.

**Terms of contribution:**

### DIFF
--- a/sources/api/early-boot-config/src/provider/aws.rs
+++ b/sources/api/early-boot-config/src/provider/aws.rs
@@ -85,7 +85,7 @@ impl PlatformDataProvider for AwsDataProvider {
     ) -> std::result::Result<Vec<SettingsJson>, Box<dyn std::error::Error>> {
         let mut output = Vec::new();
 
-        let mut client = ImdsClient::new().await.context(error::ImdsClient)?;
+        let mut client = ImdsClient::new();
 
         // Attempt to read from local file first on the `aws-dev` variant
         #[cfg(bottlerocket_platform = "aws-dev")]
--- a/sources/imdsclient/README.md
+++ b/sources/imdsclient/README.md
@@ -6,6 +6,9 @@ Current version: 0.1.0
 
 The library uses IMDSv2 (session-oriented) requests over a pinned schema to guarantee compatibility.
 Session tokens are fetched automatically and refreshed if the request receives a `401` response.
+If an IMDS token fetch or query fails, the library will continue to retry with a fibonacci backoff
+strategy until it is successful or times out. The default timeout is 300s to match the ifup timeout
+set in wicked.service, but can configured using `.with_timeout` during client creation.
 
 Each public method is explicitly targeted and return either bytes or a `String`.
 
--- a/sources/imdsclient/src/lib.rs
+++ b/sources/imdsclient/src/lib.rs
@@ -3,6 +3,9 @@
 
 The library uses IMDSv2 (session-oriented) requests over a pinned schema to guarantee compatibility.
 Session tokens are fetched automatically and refreshed if the request receives a `401` response.
+If an IMDS token fetch or query fails, the library will continue to retry with a fibonacci backoff
+strategy until it is successful or times out. The default timeout is 300s to match the ifup timeout
+set in wicked.service, but can configured using `.with_timeout` during client creation.
 
 Each public method is explicitly targeted and return either bytes or a `String`.
 
@@ -16,26 +19,38 @@ The result is returned as a `String` _(ex. m5.large)_.
 
 #![deny(rust_2018_idioms)]
 
+use std::sync::RwLock;
+
 use http::StatusCode;
 use log::{debug, info, trace, warn};
 use reqwest::Client;
 use serde_json::Value;
 use snafu::{ensure, OptionExt, ResultExt};
-use std::time::Duration;
-use tokio::time;
+use tokio::time::{timeout, Duration};
+use tokio_retry::{strategy::FibonacciBackoff, Retry};
 
 const BASE_URI: &str = "http://169.254.169.254";
 const PINNED_SCHEMA: &str = "2021-01-03";
 
-// Currently only able to get fetch session tokens from `latest`
+// Currently only able to get fetch session tokens from `latest`.
 const SESSION_TARGET: &str = "latest/api/token";
+
+// Retry timeout tied to wicked.service ifup timeout.
+const RETRY_TIMEOUT_SECS: u64 = 300;
+
+fn retry_strategy() -> impl Iterator<Item = Duration> {
+    // Retry attempts at 0.25s, 0.5s, 1s, 1.75s, 3s, 5s, 8.25s, 13.5s, 22s and then every 10s after.
+    FibonacciBackoff::from_millis(250).max_delay(Duration::from_secs(10))
+}
 
 /// A client for making IMDSv2 queries.
 /// It obtains a session token when it is first instantiated and is reused between helper functions.
 pub struct ImdsClient {
     client: Client,
     imds_base_uri: String,
-    session_token: String,
+    retry_timeout: Duration,
+    // The token is reader-writer locked to prevent reads while it's being refreshed in retry logic.
+    session_token: RwLock<String>,
 }
 
 impl ImdsClient {
@@ -45,12 +60,21 @@ impl ImdsClient {
 
     async fn new_impl(imds_base_uri: String) -> Result<Self> {
         let client = Client::new();
-        let session_token = fetch_token(&client, &imds_base_uri).await?;
+        let retry_timeout = Duration::from_secs(RETRY_TIMEOUT_SECS);
+        let session_token =
+            RwLock::new(fetch_token(&client, &imds_base_uri, &retry_timeout).await?);
         Ok(Self {
             client,
             imds_base_uri,
+            retry_timeout,
             session_token,
         })
+    }
+
+    /// Overrides the default timeout when building your own ImdsClient.
+    pub fn with_timeout(mut self, timeout: Duration) -> Self {
+        self.retry_timeout = timeout;
+        self
     }
 
     /// Gets `user-data` from IMDS. The user-data may be either a UTF-8 string or compressed bytes.
@@ -114,7 +138,7 @@ impl ImdsClient {
             .context(error::MacAddresses)?
             .clone();
 
-        // Get the IPv6 addresses associated with the primary network interface
+        // Get the IPv6 addresses associated with the primary network interface.
         let ipv6_address_target = format!("meta-data/network/interfaces/macs/{}/ipv6s", mac);
 
         let ipv6_address = self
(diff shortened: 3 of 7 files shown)


sources/imdsclient/src/lib.rs
```diff
@@ -21,20 +21,27 @@ use log::{debug, info, trace, warn};
...
+use tokio_retry::{strategy::FibonacciBackoff, Retry};
 
 const BASE_URI: &str = "http://169.254.169.254";
 const PINNED_SCHEMA: &str = "2021-01-03";
 
 // Currently only able to get fetch session tokens from `latest`
 const SESSION_TARGET: &str = "latest/api/token";
 
+fn retry_token_strategy() -> impl Iterator<Item = Duration> {
+    FibonacciBackoff::from_millis(1000)
```
tjkirch:
```suggestion
    FibonacciBackoff::from_millis(500)
```
1s seems kind of high for a local service that's in the boot path.

sources/imdsclient/src/lib.rs
```diff
@@ -21,20 +21,27 @@ use log::{debug, info, trace, warn};
...
+        .factor(3)
```
tjkirch: Why 3?  I don't think the docs or code around factor are very clear, but this seems to me to sort of arbitrarily multiply the result by 3..?  So if you have `FibonacciBackoff::from_millis(1000)`, you'd actually start by waiting 3s, then 3s, 6s, 9s, etc.  Why not 1s, 1s, 2s, 3s, etc.?  I guess this is similar to my suggestion to move down to 500ms; I don't see a reason to wait very long (in the beginning) for a local service that's in the boot path.

sources/imdsclient/src/lib.rs
```diff
@@ -627,14 +638,17 @@ mod test {
     async fn fetch_token_timeout() {
         let server = Server::run();
         let base_uri = format!("http://{}", server.addr());
+        let retry_token_timeout = Duration::from_secs(10);
```
tjkirch: Am I correct in understanding that this test expects to wait this full timeout value, confirming that it does fail after that time?  If so, this seems like quite a long time to block a unit test process.  Can we do 1s?

sources/imdsclient/src/lib.rs
```diff
@@ -45,10 +52,12 @@ impl ImdsClient {
 
     async fn new_impl(imds_base_uri: String) -> Result<Self> {
         let client = Client::new();
-        let session_token = fetch_token(&client, &imds_base_uri).await?;
+        let retry_token_timeout = Duration::from_secs(300);
```
tjkirch: This seems like a long time to wait, and quite a jump from what seemed to be 15 seconds before - is this right?  I would tend to think a host is unhealthy and replace it long before 5 minutes of attempting to contact IMDS.

sources/imdsclient/src/lib.rs
```diff
@@ -21,20 +21,27 @@ use log::{debug, info, trace, warn};
 use reqwest::Client;
```
tjkirch, line 26: I think it'd be worth having a doc comment or something explaining the various timeout values involved here.  I see 60 seconds for the retry strategy, 300 seconds for the retry token timeout (which I think means we'd run the retry process ~5 times?), and 1 second between non-token IMDS request attempts.  It would be helpful to document where those numbers come from and how they interact.  (They all seem high to me.)

jpculp:
- Reduced initial token retry duration from 1000ms to 500ms.
- Reduced factor from 3 to 1.
- Added comments to clarify where timings are coming from.

jpculp:
- Fixed incorrect test. _(Correct test timeout was reduced from 10s in the initial commit to 2s)_

sources/imdsclient/src/lib.rs
jpculp: I think I was being overly cautious as to not spam IMDS. I dropped to 500ms and went with a factor of 1 with a max delay of 10 to keep things simple.

sources/imdsclient/src/lib.rs
jpculp: I dropped it to 2s since the other tests take about 2s on my machine.

sources/imdsclient/src/lib.rs
jpculp: This is tied to our timeout in [wicked.service](https://github.com/bottlerocket-os/bottlerocket/commit/6793cdf1241716fdefc43560ed9ba8136d7cb2f8#diff-6b590f776f0f2cce22e6d222e1455d8f218e3492cd569654b8eaa918dd8906ce), but I've added a comment for clarity.

sources/imdsclient/src/lib.rs
```diff
@@ -21,20 +21,28 @@ use log::{debug, info, trace, warn};
...
 const BASE_URI: &str = "http://169.254.169.254";
 const PINNED_SCHEMA: &str = "2021-01-03";
 
 // Currently only able to get fetch session tokens from `latest`
 const SESSION_TARGET: &str = "latest/api/token";
 
+fn retry_token_strategy() -> impl Iterator<Item = Duration> {
+    // Retries at 0.5s, 1s, 1.5s, 2.5s, 4s, 6.5s, 10.5s, 17s, and then every 10s after.
+    FibonacciBackoff::from_millis(500)
+        .factor(1)
```
tjkirch:
```suggestion
```
I think specifying the default here just raises questions, and I'd argue that the `factor` method doesn't really make much sense anyway, so they're not super useful questions to raise...

sources/imdsclient/src/lib.rs
tjkirch: There are a couple comments in the code now, but it seems to me that the retries and delay are an important part of usage of the library, and since they're not all configurable and are tied to a somewhat arbitrary timeout in another program, I think a comment of higher prominence could be helpful - right in the module documentation perhaps?

sources/imdsclient/src/lib.rs
```diff
@@ -21,20 +21,28 @@ use log::{debug, info, trace, warn};
...
+use tokio_retry::{strategy::FibonacciBackoff, Retry};
 
 const BASE_URI: &str = "http://169.254.169.254";
 const PINNED_SCHEMA: &str = "2021-01-03";
 
 // Currently only able to get fetch session tokens from `latest`
 const SESSION_TARGET: &str = "latest/api/token";
 
+fn retry_token_strategy() -> impl Iterator<Item = Duration> {
+    // Retries at 0.5s, 1s, 1.5s, 2.5s, 4s, 6.5s, 10.5s, 17s, and then every 10s after.
```
tjkirch: Is this right?  I would have guessed 0.5s, 1s, 2s, 3.5s, etc.  (It might also help to clarify that these are absolute times and not relative to the previous delay, since I was confused how we could have 10.5s and 17s delays with a 10s max.)

tjkirch (approved):

sources/imdsclient/src/lib.rs
jpculp: Looks like my math was initially incorrect. I've verified the fixed comment with some logged messages.

sources/imdsclient/src/lib.rs
jpculp: I was a tad worried if the comments were too far away from the code the timings could potentially be updated without making the appropriate changes to the documentation. I could add blurb in the module documentation that describes the behavior, but doesn't called out the exact timings if that helps.

webern: > I could add blurb in the module documentation that describes the behavior

That seems reasonable to me. I think it is unusual to retry something for several minutes, so a bit of a prominent documentation of this makes sense.

webern (approved):

jpculp:
In this latest push rather than just using tokio retry and timeout on the token fetch, they are used for fetching IMDS data as well. This introduces things like RwLock to allow for the token to be refreshed within the timeout/retry logic.

In a second commit I've also moved the token fetch out of ImdsClient::new() so that it no longer requires an await, but we can move that to a separate PR if folks are concerned it is out of scope.

sources/imdsclient/src/lib.rs
```diff
@@ -245,7 +244,13 @@ impl ImdsClient {
...
             self.retry_timeout,
             Retry::spawn(retry_strategy(), || async {
-                let session_token = self.read_token().await?;
+                let session_token = match self.read_token().await? {
+                    Some(session_token) => session_token,
+                    None => {
+                        self.write_token().await?;
+                        error::EmptyToken.fail()?
+                    }
+                };
```
jpculp: Since this triggers a retry we are adding about 500ms to the first call of each new ImdsClient, but it would not impact subsequent calls. I'm not sure this is noticeable, but if it's a concern we can change the backoff strategy to start with 250ms or maybe add a nested read before error.

cbgbt: Maybe `write_token` can return a clone of the fetched token, that way if it succeeds here we don't need to trigger a retry?

Github seems to have replicated this comment for reasons I don't understand, sorry :sweat_smile:

sources/imdsclient/src/lib.rs
```diff
@@ -221,92 +241,109 @@ impl ImdsClient {
...
-    async fn refresh_token(&mut self) -> Result<()> {
-        self.session_token = fetch_token(&self.client, &self.imds_base_uri).await?;
-        Ok(())
+    /// Helper to read session token within the ImdsClient
+    async fn read_token(&self) -> Result<Option<String>> {
+        let session_token = match self
+            .session_token
+            .read()
+            .map_err(|_| error::Error::FailedReadToken {})?
+            .clone()
```
cbgbt, line 357: nit: It is a bit unusual to return a `clone()` of a value rather than a reference. It might be worth leaving a comment that the idea is to allow the `RwLock` to be released as soon as possible, as this detail might not be so obvious as time passes.

jpculp:
- Adjusted logic to avoid triggering a retry on first run.
- Added comment about clone in `read_token`.
- Start backoff with 250ms for even quicker retries.

sources/imdsclient/README.md
```diff
@@ -6,6 +6,9 @@ Current version: 0.1.0
 
 The library uses IMDSv2 (session-oriented) requests over a pinned schema to guarantee compatibility.
 Session tokens are fetched automatically and refreshed if the request receives a `401` response.
+If an IMDS token or query fails to be fetched, the library will continue to retry with a fibonacci
```
zmrow: Nit: Might be a little smoother to say "If an IMDS token fetch or query fails, the library ..."

sources/imdsclient/src/lib.rs
```diff
@@ -16,43 +19,60 @@ The result is returned as a `String` _(ex. m5.large)_.
...
 
 impl ImdsClient {
-    pub async fn new() -> Result<Self> {
-        Self::new_impl(BASE_URI.to_string()).await
+    pub fn new() -> Result<Self> {
+        Self::new_impl(BASE_URI.to_string())
     }
 
-    async fn new_impl(imds_base_uri: String) -> Result<Self> {
+    fn new_impl(imds_base_uri: String) -> Result<Self> {
```
zmrow: We're not allowing the user to override `imds_base_uri` anywhere, so this method probably doesn't need to take it as an argument.

sources/imdsclient/src/lib.rs
```diff
@@ -16,43 +19,60 @@ The result is returned as a `String` _(ex. m5.large)_.
...
         let client = Client::new();
-        let session_token = fetch_token(&client, &imds_base_uri).await?;
+        // Retry token timeout tied to wicked.service ifup timeout
+        let retry_timeout = Duration::from_secs(300);
```
zmrow: Nit: I'd probably make the default timeout a constant so it's not buried in the code down here.

sources/imdsclient/src/lib.rs
```diff
@@ -16,43 +19,60 @@ The result is returned as a `String` _(ex. m5.large)_.
...
     imds_base_uri: String,
-    session_token: String,
+    retry_timeout: Duration,
+    session_token: RwLock<Option<String>>,
 }
 
 impl ImdsClient {
-    pub async fn new() -> Result<Self> {
-        Self::new_impl(BASE_URI.to_string()).await
+    pub fn new() -> Result<Self> {
```
zmrow: This doesn't need to return a `Result` anymore as there's nothing in the method that can fail; it can return `Self`.

sources/imdsclient/src/lib.rs
```diff
@@ -16,43 +19,60 @@ The result is returned as a `String` _(ex. m5.large)_.
...
+}
+
 /// A client for making IMDSv2 queries.
 /// It obtains a session token when it is first instantiated and is reused between helper functions.
 pub struct ImdsClient {
     client: Client,
     imds_base_uri: String,
-    session_token: String,
+    retry_timeout: Duration,
+    session_token: RwLock<Option<String>>,
```
zmrow, line 52: Nice - `RwLock` is a good idea here!  Could you please add a comment explaining why it's necessary?

sources/imdsclient/src/lib.rs
```diff
@@ -400,19 +442,29 @@ mod error {
     #[derive(Debug, Snafu)]
     #[snafu(visibility = "pub(super)")]
 
+    // snafu doesn't yet support the lifetimes used by std::sync::PoisonError
     pub enum Error {
         #[snafu(display("Response '{}' from '{}': {}", get_status_code(source), uri, source))]
         BadResponse { uri: String, source: reqwest::Error },
 
+        #[snafu(display("Fetched IMDSv2 session token"))]
```
zmrow: Should this say something about the token being empty?  Or am I reading this wrong?

sources/imdsclient/src/lib.rs
```diff
@@ -221,92 +241,114 @@ impl ImdsClient {
...
+    /// Fetches a new session token and writes it to the current ImdsClient.
+    async fn write_token(&self) -> Result<String> {
+        let session_token =
+            match fetch_token(&self.client, &self.imds_base_uri, &self.retry_timeout).await? {
+                Some(session_token) => {
+                    *self
+                        .session_token
+                        .write()
+                        .map_err(|_| error::Error::FailedWriteToken {})? =
+                        Some(session_token.clone());
```
cbgbt: nit: The prefix `*` notation in Rust for dereferencing can be a bit difficult to parse for readers. In this case, the way that `RwLock` provides you write access is by returning an `RwLockWriteGuard<String>` which protects our value. `RwLockWriteGuard<T>` gives you access to the internal value by implementing `Deref`.

It might read more cleanly to call the `deref()` method over using the `*` prefix.


```suggestion
                    self
                        .session_token
                        .write()
                        .map_err(|_| error::Error::FailedWriteToken {})?
                        .deref() =
                        Some(session_token.clone());
```

sources/imdsclient/src/lib.rs
```diff
@@ -422,6 +474,9 @@ mod error {
         #[snafu(display("Response was not UTF-8: {}", source))]
         NonUtf8Response { source: std::string::FromUtf8Error },
 
+        #[snafu(display("IMDSv2 session token was refreshed."))]
```
zmrow: This message is a little confusing to me.  The error is `UnauthorizedTokenRefreshed`, but the message says the token was refreshed.  Maybe I'm missing some context here.

sources/imdsclient/src/lib.rs
```diff
@@ -400,19 +442,29 @@ mod error {
...
+
         #[snafu(display("IMDS fetch failed after {} attempts", attempt))]
         FailedFetchIMDS { attempt: u8 },
 
-        #[snafu(display("Failed to fetch IMDSv2 session token after {} attempts", attempt))]
-        FailedFetchToken { attempt: u8 },
+        #[snafu(display("Failed to fetch IMDSv2 session token"))]
+        FailedFetchToken,
+
+        #[snafu(display("Failed to read token within ImdsClient"))]
```
zmrow, line 469: Nit: Since these are "internal" implementation detail errors, let's make the message a bit clearer.  Something like "Failed to read token from RwLock..."? (I'm not sold on that wording in particular)

sources/api/early-boot-config/src/provider/aws.rs
```diff
@@ -85,7 +85,7 @@ impl PlatformDataProvider for AwsDataProvider {
     ) -> std::result::Result<Vec<SettingsJson>, Box<dyn std::error::Error>> {
         let mut output = Vec::new();
 
-        let mut client = ImdsClient::new().await.context(error::ImdsClient)?;
+        let mut client = ImdsClient::new().context(error::ImdsClient)?;
```
zmrow: Since the `new()` method can't really fail now, I suggested below that it return a `Self` rather than a `Result`.  Once we do that we can remove the `.context(...)` here! :)

webern (approved):

sources/api/early-boot-config/src/provider/aws.rs
jpculp: True!

sources/imdsclient/src/lib.rs
jpculp: new_impl is just for unit testing so that we can point the client to the server httptest creates.

jpculp:
- Replaced `Result<Self>` with `Self` in `ImdsClient::new()` (and added **Default** for `clippy`).
- Made the 300s `retry_timeout` a constant (**RETRY_TIMEOUT_SECS**).
- Renamed **UnauthorizedTokenRefreshed** to **TokenRefreshed** to be less confusing.
- Removed outdated comment about fetching the token on instantiation.
- Other small fixes for increased clarity.

sources/imdsclient/src/lib.rs
jpculp: My original line of thinking was the token was refreshed because the query returned a 401, but I went ahead and renamed this to **TokenRefreshed**.

sources/imdsclient/src/lib.rs
jpculp: This was the token being fetched because it was empty, but actually this error isn't used anymore so I've removed it in this latest push.

webern (approved):

cbgbt (approved):

sources/imdsclient/src/lib.rs
cbgbt: After talking with @jpculp, we concluded that this won't work! It's because I misunderstood how `Deref` works. See [this SA answer](https://stackoverflow.com/a/62506306):

> If x has type &T or &mut T, then *x has type T.
>
> What if x has some other type S? If S implements Deref or DerefMut, then Rust will call .deref() or .deref_mut() on x, and dereference that.

sources/imdsclient/src/lib.rs
```diff
@@ -221,92 +246,110 @@ impl ImdsClient {
...
+                    StatusCode::NOT_FOUND => Ok(None),
 
-                    return Ok(Some(response_body));
-                }
+                    // IMDS returns 401 if the session token is expired or invalid.
+                    StatusCode::UNAUTHORIZED => {
+                        warn!("Session token is invalid or expired");
+                        self.write_token().await?;
+                        error::TokenRefreshed.fail()
+                    }
```
bcressey: This feels a bit like using exceptions for control flow, but it seems clear enough once it clicks that returning an error here is what prompts Retry to try again.

The logic might be clearer if written like this:
```suggestion
                    // IMDS returns 401 if the session token is expired or invalid.
                    StatusCode::UNAUTHORIZED => {
                        warn!("Session token is invalid or expired");
                        self.clear_token(); // swaps the Option back to None
                        error::TokenInvalid.fail()
                    }
```

Otherwise, we might successfully refresh the token, and then immediately return an error showing that we failed to refresh the token.

This also avoids some conceptual duplication with the token refresh that happens at the top of each iteration.

jpculp:
- Implemented @bcressey's suggestion to clear the token on 401/Unauthorized.
- Fixed some comment tweaks I missed in the previous push.

bcressey (approved):

zmrow (approved):

cbgbt (approved):

PR #1861, opened 2021-12-09
Change when ldconfig runs and what files it looks for

arnaldo2792:
**Issue number:**
N / A

**Description of changes:**
Related to #1799 , that PR is quite big. I'm taking some of the commits in that PR as their individual PR. The commit in this PR is variant independent, since it changes a package used among all variants.

```
507d5eae systemd: delay ldconfig.service
```

There are packages that create `ldconfig.d/*.conf` files using a tmpfilesd file to cache additional libraries, thus `ldconfig.service` has to run after the tmpfilesd daemon creates the files.

```
8b20ba9b Include ld.so.conf.d/* to build ld cache
```
This commit adds a configuration file for ldconfig to read additional configurations from /etc/ld.so.conf, which will be created at runtime by the tmpfilesd daemon.

**Testing done:**
- Build a Bottlerocket aws-dev image
- Confirmed that `ldconfig.service` waits for `systemd-tmpfiles-setup.service`
- Confirmed that `ldconfig` looks for libraries in directories specified in `/etc/ld.so.conf.d/*.conf` files

**Terms of contribution:**

### DIFF
--- a/packages/glibc/glibc-tmpfiles.conf
+++ b/packages/glibc/glibc-tmpfiles.conf
@@ -1,4 +1,5 @@
-f /etc/ld.so.conf 0644 root root -
+C /etc/ld.so.conf 0644 root root -
 f /etc/ld.so.cache 0644 root root -
+d /etc/ld.so.conf.d/ 0644 root root -
 d /run/cache/ldconfig 0700 root root -
 f /run/cache/ldconfig/aux-cache 0600 root root -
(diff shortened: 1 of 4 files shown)


etungsten (approved):

packages/release/ldconfig-service.conf
```diff
@@ -0,0 +1,4 @@
+[Unit]
+# Run `ldconfig.service` after `tmpfilesd`, since `ldconfig` will attempt to
+# read files from `/etc/ld.so.conf.d`, which are created by `tmpfilesd`
+After=systemd-tmpfiles-setup.service
```
bcressey: Let's put this in `glibc` as well; I think that's more clear since it's the use of `ld.so.conf` via tmpfiles that creates this ordering requirement.

arnaldo2792:
Forced push includes:

- Move drop-in file to `glibc` instead of `release`

arnaldo2792:
Forced push squashes both commits. In previous versions, the commits modified different packages, and now they modify the same package.

bcressey (approved):

etungsten (approved):

PR #1868, opened 2021-12-13
Add shimpei to inject runc prestart hooks

arnaldo2792:
**Issue number:**
N / A

**Description of changes:**
Related to #1799 , that PR is quite big. I'm taking some of the commits in that PR as their individual PR.

```
 sources: add shimpei

 shimpei is an OCI shim wrapper around `oci-add-hooks`, whose sole
 porpuse is to call `oci-add-hooks` with additional parameters that can't
 be provided by containerd.
```

```
packages: add oci-add-hooks
```

`oci-add-hooks` is built as part of the release package, but it isn't included in any variant. Instead, packages that depend on it will include it in their spec file, as follows:

Requires: %{_cross_os}oci-add-hooks

**Testing done:**
In the new variant created in #1799:

I created a new containerd configuration that calls `shimpei` instead of runc:

```toml
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia.options]
SystemdCgroup = true
BinaryName = "nvidia-oci" # this is a symlink to `shimpei`, used to figure out the `hooks.json` file 
```

The prestart hook in #1799, sets up GPUs and mounts helper binaries like `nvidia-smi`. In my test, I called `nvidia-smi` in a GPU-enabled orchestrated container:

```
kubectl exec -it nvidia-device-plugin-daemonset-4q2zf -n kube-system -- /bin/bash
root@nvidia-device-plugin-daemonset-4q2zf:/# nvidia-smi
Mon Dec 13 20:33:53 2021
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 470.57.02    Driver Version: 470.57.02    CUDA Version: N/A      |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla V100-SXM2...  Off  | 00000000:00:1E.0 Off |                    0 |
| N/A   29C    P0    24W / 300W |      0MiB / 16160MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```

**Terms of contribution:**

### DIFF
--- a/packages/release/Cargo.toml
+++ b/packages/release/Cargo.toml
@@ -36,6 +36,11 @@ libgcc = { path = "../libgcc" }
 libstd-rust = { path = "../libstd-rust" }
 makedumpfile = { path = "../../packages/makedumpfile" }
 os = { path = "../os" }
+# We don't include `oci-add-hooks` in all variants, and its only consumer is
+# `shimpei` in the "os" package. However, we want to start the "os" package
+# build ASAP since it takes the most time. Since `oci-add-hooks` builds quickly,
+# it doesn't affect other variants very much even when it's not used.
+oci-add-hooks = { path = "../oci-add-hooks" }
 policycoreutils = { path = "../policycoreutils" }
 procps = { path = "../procps" }
 selinux-policy = { path = "../selinux-policy" }
--- a/sources/shimpei/src/main.rs
+++ b/sources/shimpei/src/main.rs
@@ -1,0 +1,110 @@
+/*!
+  shimpei is an OCI compatible shim wrapper around `oci-add-hooks`. Its sole purpose is
+  to call `oci-add-hooks` with the additional `--hook-config-path` and `--runtime-path`
+  parameters that can't be provided by containerd.
+*/
+
+#![deny(rust_2018_idioms)]
+
+#[macro_use]
+extern crate log;
+
+use nix;
+use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
+use snafu::{OptionExt, ResultExt};
+use std::env;
+use std::ffi::CString;
+use std::path::Path;
+use std::process;
+
+/// Path to runc binary
+const RUNC_BIN_PATH: &str = "/usr/bin/runc";
+
+/// Path to hooks definitions
+const HOOK_CONFIG_BASE_PATH: &str = "/usr/share/oci-add-hooks";
+
+/// Path to oci-add-hooks
+const OCI_ADD_HOOKS: &str = "/usr/bin/oci-add-hooks";
+
+fn run() -> Result<()> {
+    setup_logger()?;
+    let mut args = env::args();
+    let prefix = args.next().context(error::MissingArg { what: "name" })?;
+    let hook_path = Path::new(HOOK_CONFIG_BASE_PATH).join(format!("{}-hook.json", prefix));
+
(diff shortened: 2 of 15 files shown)


packages/release/Cargo.toml
```diff
@@ -36,6 +36,7 @@ libgcc = { path = "../libgcc" }
 libstd-rust = { path = "../libstd-rust" }
 makedumpfile = { path = "../../packages/makedumpfile" }
 os = { path = "../os" }
+oci-add-hooks = { path = "../oci-add-hooks" }
```
bcressey, line 43: nit: this might be worth an explanation -

```suggestion
# We don't include `oci-add-hooks` in all variants, and its only consumer is
# `shimpei` in the "os" package. However, we want to start the "os" package
# build ASAP since it takes the most time. Since `oci-add-hooks` builds quickly,
# it doesn't affect other variants very much even when it's not used.
oci-add-hooks = { path = "../oci-add-hooks" }
```

Also open to push back from others, since the obvious counter-argument is that if it's quick to build, it doesn't matter too much and it'd be better to add the dependency to "os" where it belongs.

sources/shimpei/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+        Logger { source: log::SetLoggerError },
+
+        #[snafu(display("Invalid log level '{}'", log_level))]
+        LogLevel {
+            log_level: String,
+            source: log::ParseLevelError,
+        },
+
+        #[snafu(display("Invalid string"))]
+        InvalidString { source: std::ffi::NulError },
```
bcressey: nit:
```suggestion
        #[snafu(display("Couldn't create CString from '{}': {}", input, source))]
        InvalidString { source: std::ffi::NulError, input: &str },
```

sources/shimpei/src/main.rs
```diff
@@ -0,0 +1,97 @@
...
+
+        #[snafu(display("Failed to execute syscall `exec`: {}", source))]
+        Execv { source: nix::Error },
```
bcressey: nit:
```suggestion
        #[snafu(display("Failed to exec '{}': {}", program, source))]
        Execv { source: nix::Error, program: &str },
```

bcressey (approved):

bcressey:
I expect this will build successfully once the files are uploaded to the lookaside cache.

packages/release/Cargo.toml
arnaldo2792: I think we can be consistent with what we do with other packages, like `makedumpfile`, and keep the build in release, not in os.

arnaldo2792:
Forced push includes:

- Add comment explaining why `oci-add-hooks` is in `release/Cargo.toml`
- Nits in `shimpei`'s messages

I also updated the lookaside cache

bcressey (approved):

etungsten (approved):

PR #1869, opened 2021-12-16
update SDK and build log4j2 hotpatch

bcressey:
**Issue number:**
N/A

**Description of changes:**
Update SDK to the new release, which includes maven, and use it to build the [log4j2 hotpatch](https://github.com/corretto/hotpatch-for-apache-log4j2/).

**Testing done:**
Built some variants on both architectures, for both architectures.

Additional tests:
- [x] smoke tested a task for `aws-ecs-1`
- [x] sonobuoy passed for `aws-k8s-1.21`
- [x] @cbgbt verified that the hotpatch jar works

**Terms of contribution:**

### DIFF
--- a/packages/log4j2-hotpatch/log4j2-hotpatch.spec
+++ b/packages/log4j2-hotpatch/log4j2-hotpatch.spec
@@ -1,0 +1,32 @@
+%global debug_package %{nil}
+
+%global project hotpatch-for-apache-log4j2
+
+Name: %{_cross_os}log4j2-hotpatch
+Version: 1.3.0
+Release: 1%{?dist}
+Summary: Tool for hot patching log4j2 vulnerabilities
+License: Apache-2.0
+URL: https://github.com/corretto/%{project}
+Source0: https://github.com/corretto/%{project}/archive/%{version}/%{version}.tar.gz#/%{project}-%{version}.tar.gz
+
+%description
+%{summary}.
+
+%prep
+%autosetup -n %{project}-%{version}
+
+%build
+xmvn --offline clean package
+
+%install
+install -d %{buildroot}%{_cross_datadir}/hotdog
+install -p -m 0644 target/Log4jHotPatch.jar %{buildroot}%{_cross_datadir}/hotdog
+
+%files
+%license LICENSE
+%{_cross_attribution_file}
+%dir %{_cross_datadir}/hotdog
+%{_cross_datadir}/hotdog/Log4jHotPatch.jar
+
+%changelog
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -26,7 +26,7 @@ BUILDSYS_NAME = "bottlerocket"
 # "Bottlerocket Remix by ${CORP}" or "${CORP}'s Bottlerocket Remix"
 BUILDSYS_PRETTY_NAME = "Bottlerocket OS"
 # SDK version used for building
-BUILDSYS_SDK_VERSION="v0.23.1"
+BUILDSYS_SDK_VERSION="v0.24.0"
 # Site for fetching the SDK
 BUILDSYS_REGISTRY="public.ecr.aws/bottlerocket"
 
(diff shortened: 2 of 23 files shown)


cbgbt:
LGTM, pending testing.

jpculp:
Looks good!

jpculp (approved):

packages/log4j2-hotpatch/log4j2-hotpatch.spec
```diff
@@ -0,0 +1,36 @@
...
+
+%global project hotpatch-for-apache-log4j2
+
+Name: %{_cross_os}log4j2-hotpatch
+Version: 1.3.0
+Release: 1%{?dist}
+Summary: Tool for hot patching log4j2 vulnerabilities
+License: Apache-2.0
+URL: https://github.com/corretto/%{project}
+Source0: https://github.com/corretto/%{project}/archive/%{version}/%{version}.tar.gz#/%{project}-%{version}.tar.gz
```
arnaldo2792, line 11: Nit: left over comment at the end? Or is this intentional?

arnaldo2792 (approved):

packages/log4j2-hotpatch/log4j2-hotpatch.spec
bcressey: It's intentional; it's a URL fragment that `rpm` will parse as the file name for use in the `%setup` macro.

jpculp (approved):

packages/log4j2-hotpatch/0001-Log-to-stderr-instead-of-stdout.patch
```diff
@@ -0,0 +1,35 @@
+From 61921bad328e3b12305fd91162c30e9a54bedfbf Mon Sep 17 00:00:00 2001
+From: Stewart Smith <trawets@amazon.com>
+Date: Thu, 16 Dec 2021 07:54:47 -0800
+Subject: [PATCH] Log to stderr instead of stdout
```
samuelkarp: Should we also suppress the log output entirely?  Otherwise it will be visible in container logs, and if those are being parsed it could be problematic.

(I think we can suppress the log output entirely with another JVM option to the hotpatch in hotdog.)

cbgbt (approved):

packages/log4j2-hotpatch/0001-Log-to-stderr-instead-of-stdout.patch
```diff
@@ -0,0 +1,35 @@
...
++      System.err.println(message);
+     }
+   }
+ 
+@@ -260,7 +260,7 @@ public class Log4jHotPatch {
+         log("Patching all JVMs!");
+       }
+     } else if (args.length == 1 && ("-h".equals(args[0]) || "-help".equals(args[0]) || "--help".equals(args[0]))) {
+-      System.out.println("usage: Log4jHotPatch [<pid> [<pid> ..]]");
++      System.err.println("usage: Log4jHotPatch [<pid> [<pid> ..]]");
```
samuelkarp: Is it common for `-help` output to be on stderr?

samuelkarp (approved):
(hit the wrong button, approving)

packages/log4j2-hotpatch/0001-Log-to-stderr-instead-of-stdout.patch
bcressey: It's not. Not sure if there's a good reason for this - @stewartsmith ?

webern (approved):

packages/log4j2-hotpatch/0001-Log-to-stderr-instead-of-stdout.patch
bcressey: Dropping this patch in favor of suppressing output in hotdog.

samuelkarp: Corresponding hotdog PR: https://github.com/bottlerocket-os/hotdog/pull/1

samuelkarp (approved):

PR #1870, opened 2021-12-16
add support for unified image layout

bcressey:
**Issue number:**
N/A

**Description of changes:**
Add support for a "unified" image layout, where the OS and data partitions reside on the same disk.

In order to resize the data partition when it's on the same disk as the root filesystem, switch from our homegrown `growpart` tool to `systemd-repart`, which knows how to tell the kernel about changes to individual partition entries.

Refactor the unit dependencies for setting up `/local` to take advantage of the new repart functionality, which lets us an ordinary mount unit for the filesystem.

**Testing done:**

- [x] "unified" and "split" layout AMI registration works
- [x] "unified" and "split" layout OVA creation works
- [x] `/local` is correctly resized to fill the disk for both "unified" and "split"
- [x] `systemd-repart` and `growpart` resize the disk to the same end sector
- [x] overlayfs and bind mount directories are created and labeled

**Terms of contribution:**


Makefile.toml
```diff
@@ -438,57 +444,139 @@ cleanup() {
...
-   echo "No OVF template or VMDK images, skipping OVA build"
-   exit 0
+# Short circuit if no OVF template exists.
+if [ ! -s "${BUILDSYS_OVF_TEMPLATE}" ] ; then
+   if [ ! -s "${root_vmdk_path}" ]; then
+      # If no VMDK exists either, there's nothing to do.
+      echo "No OVF template or VMDK images, skipping OVA build"
+      exit 0
+   else
+      # Warn the user if a VMDK exists but an OVF template does not.  Assume we do not
```
arnaldo2792, line 475: Nit: I don't know if this is intended, but there is an extra space here
```suggestion
      # Warn the user if a VMDK exists but an OVF template does not. Assume we do not
```

Makefile.toml
```diff
@@ -438,57 +444,139 @@ cleanup() {
...
    exit 1
 fi
 
+is_split="no"
+if [ -s "${data_vmdk_path}" ] ; then
+   is_split="yes"
+fi
+
+root_image_size_bytes="$(measure_image "${root_vmdk_path}")"
+root_image_size_gib="$((root_image_size_bytes / 1024 / 1024 / 1024))"
```
arnaldo2792: Is it possible for `1024 * 1024 * 1024` and `1024 / 1024 / 1024` to be constants?

jpculp (approved):

packages/release/release.spec
```diff
@@ -117,9 +123,9 @@ EOF
 
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 \
-  %{S:1001} %{S:1002} %{S:1003} %{S:1004} %{S:1005} \
-  %{S:1006} %{S:1007} %{S:1008} %{S:1009} %{S:1010} %{S:1011} %{S:1012} \
-  %{S:1015} %{S:1040} %{S:1041} %{S:1060} %{S:1061} %{S:1062} %{S:1080} \
+  %{S:1001} %{S:1002} %{S:1003} %{S:1004} %{S:1005} %{S:1006} %{S:1007} \
+  %{S:1008} %{S:1009} %{S:1010} %{S:1011} %{S:1012} %{S:1013} %{S:1015} \
+  %{S:1040} %{S:1041} %{S:1042} %{S:1060} %{S:1061} %{S:1062} %{S:1080} \
```
cbgbt, line 128: This hurts to review :cold_sweat:

cbgbt (approved):

packages/release/release.spec
```diff
@@ -101,6 +104,9 @@ install -p -m 0644 %{S:11} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
 install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/wicked/ifconfig
 install -p -m 0644 %{S:1000} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/wicked/ifconfig
 
+install -d %{buildroot}%{_cross_libdir}/repart.d
+install -p -m 0644 %{S:96} %{buildroot}%{_cross_libdir}/repart.d/80-local.conf
```
arnaldo2792, line 108: Why is this a drop-in, and not part of the unit itself?

bcressey: `systemd-repart` looks for its config files in `/usr/lib/repart.d`, like how`systemd-tmpfiles` looks in `/usr/lib/tmpfiles.d`.

Makefile.toml
bcressey: I tried to preserve the existing "two spaces after a period" convention in this section, even though I belong to the "one space after a period" tribe.

Makefile.toml
```diff
@@ -794,22 +885,59 @@ set -e
...
+
+# We will only have a data image if the variant uses the "split" format.
+is_split="no"
+if [ -s "${datalz4}" ] ; then
+   lz4 -df "${datalz4}" "${data_image}"
+   is_split="yes"
+fi
+
+root_image_size_gib="$(($(stat -c %s "${root_image}") / 1024 / 1024 / 1024))"
+if [ "${is_split}" == "yes" ] ; then
```
webern, line 915: It feels like we are at the limit of what should be done with scriptlets inside of a Makefile.toml!

bcressey:
Added `bytes_in_gib` constant to make the calculations more readable.

arnaldo2792 (approved):
:boat:

webern (approved):

Makefile.toml
bcressey: Consider the gauntlet thrown down!

(But yes - we definitely need a different approach.)

zmrow (approved):
🥇

PR #1875, opened 2021-12-17
Add missing shimpei configuration

cbgbt:
**Description of changes:**
Settings generators for shimpei were missing for Kubernetes 1.18 and 1.19 variants. This adds them.

**Terms of contribution:**

### DIFF
--- a/sources/models/src/aws-k8s-1.19/defaults.d/70-oci-hooks.toml
+++ b/sources/models/src/aws-k8s-1.19/defaults.d/70-oci-hooks.toml
@@ -1,0 +1,1 @@
+../../../shared-defaults/oci-hooks.toml


bcressey (approved):

sources/models/src/aws-k8s-1.19/defaults.d/70-oci-hooks.toml
```diff
@@ -0,0 +1 @@
+../../../shared-defaults/oci-hooks.toml
```
samuelkarp, line 1: ~This just looks like 1.19, are we also missing for 1.18?~

1.18 is just a symlink to 1.19

samuelkarp (approved):

webern (approved):

PR #1883, opened 2021-12-22
host-ctr: add '/etc/bottlerocket-release' to host containers

jpculp:
**Description of changes:**

This grants host containers context to the underlying OS by mounting
Bottlerocket's `/etc/os-release` as `/etc/bottlerocket-release`.

**Testing done:**
- Built `aws-k8s-1.21` instance.
- Ran `cat /etc/bottlerocket-release` in both the control and admin host containers.

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -695,6 +695,12 @@ func withDefaultMounts(containerID string, persistentDir string) oci.SpecOpts {
 			Source:      "cgroup",
 			Options:     []string{"ro", "nosuid", "noexec", "nodev"},
 		},
+		// Bottlerocket release information for the container
+		{
+			Options:     []string{"bind", "ro"},
+			Destination: fmt.Sprintf("/etc/bottlerocket-release"),
+			Source:      fmt.Sprintf("/etc/os-release"),
+		},
 	}
 
 	// The `current` dir was added for easier referencing in Dockerfiles and scripts.


webern (approved):

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -695,6 +695,12 @@ func withDefaultMounts(containerID string, persistentDir string) oci.SpecOpts {
 			Source:      "cgroup",
 			Options:     []string{"ro", "nosuid", "noexec", "nodev"},
 		},
+		// Bottlerocket release information for the container
```
arnaldo2792, line 698: Is this required if it can be accessed from `/.bottlerocket/rootfs/etc/release`?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -695,6 +695,12 @@ func withDefaultMounts(containerID string, persistentDir string) oci.SpecOpts {
...
+		{
+			Options:     []string{"rbind", "r"},
+			Destination: fmt.Sprintf("/etc/bottlerocket-release"),
+			Source:      fmt.Sprintf("/etc/os-release"),
+		},
```
arnaldo2792: Is this required if it can be accessed from `/.bottlerocket/rootfs/etc/release`? Is this mount just for better developer ergonomics?

I couldn't find the `r` option [here](https://www.man7.org/linux/man-pages/man8/mount.8.html), I think the options should be:

```suggestion
		// Bottlerocket release information for the container
		{
			Destination: fmt.Sprintf("/etc/bottlerocket-release"),
			Source:      fmt.Sprintf("/etc/os-release"),
			Options:     []string{"bind", "ro"}
		},
```

`bind`: you don't need `rbind`, since nothing underneath `/etc/os-release` is mounted
`ro`: set the mount as `read-only`

zmrow: Agree with @arnaldo2792  - unless we have a need/ask for better ergonomics around this file we probably don't need to add this since it's already available! :)

jpculp: We only mount up `/.bottlerocket/rootfs/` on superpowered containers, but this file would be potentially useful on less privileged containers as well. Mounting to `/etc/bottlerocket-release` is also a bit more user friendly.

I was a little silly with the mounting options though, so I will definitely make that change!

jpculp:
Fixed mount options.

bcressey (approved):

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -695,6 +695,12 @@ func withDefaultMounts(containerID string, persistentDir string) oci.SpecOpts {
 			Source:      "cgroup",
 			Options:     []string{"ro", "nosuid", "noexec", "nodev"},
 		},
+		// Bottlerocket release information for the container
+		{
+			Options:     []string{"bind", "ro"},
+			Destination: fmt.Sprintf("/etc/bottlerocket-release"),
```
arnaldo2792, line 701: nit: I do like the simplicity of just having `/etc` instead of `/.bottlerocket/bottlerocket-release`, but I still think we should have the prefix just to be consistent with the container's persistent storage mount: `/.bottlerocket/<host-containers>/<bootstrap-containers>/`

arnaldo2792 (approved):

sources/host-ctr/cmd/host-ctr/main.go
bcressey: I went back and forth on this as well. As a general rule, we try to be hands-off with the container's root filesystem, but we've bent that rule to put things in the "expected" places - `apiclient` in `/usr/local/bin`, and the API socket in `/run`.

Having the release file in `/etc` seems like a reasonable exception by that standard. And if we're touching `/etc` in any case, making it a symlink to `/.bottlerocket/<blah>/etc` just seems overly complicated.

zmrow (approved):
:octopus:

webern (approved):

PR #1893, opened 2022-01-06
grub: Use `grub-mkimage` from build step rather than SDK version

zmrow:
**Description of changes:**
```
This changes the grub build process to use `grub-mkimage` from the build
step in the spec, rather than use `grub2-mkimage` from the SDK.  The
reason for this is the SDK's version is a Fedora build which may
not match our grub images or modules.
```

**Testing done:**
* Build and boot an x86/aarch AWS variant successfully
* Build and boot an x86 VMware variant successfully

**Terms of contribution:**

### DIFF
--- a/packages/grub/grub.spec
+++ b/packages/grub/grub.spec
@@ -178,7 +178,7 @@ MODS="configfile echo ext2 gptprio linux normal part_gpt reboot sleep zstd"
 pushd bios-build
 %make_install
 mkdir -p %{buildroot}%{biosdir}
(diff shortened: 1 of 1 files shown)


bcressey (approved):

webern (approved):

zmrow:
^ Removed the `-v` from the call for EFI

bcressey (approved):

cbgbt (approved):

PR #1895, opened 2022-01-10
Add `metal-dev` variant

zmrow:
**Description of changes:**
This series of changes adds initial support for running Bottlerocket on bare metal.  A new variant `metal-dev` takes advantage of these capabilities and is similar in scope to the other `-dev` variants.

The first commit adds additional config to our `kernel-5.10` package.  The config is still relatively narrow, given the narrow scope of hardware I have to test on (see testing section).  This config will change and grow over time.
```
    kernel 5.10: Add config to support additional hardware
    
    This commit adds new config to kernel 5.10 for variants meant to run on bare
    metal.  It adds support for SATA drives, Intel networking hardware (1G/10G), as
    well as additional suport for UEFI video.  This config will continue to grow
    and morph over time as we support additional hardware, etc
```

This commit changes the location that `early-boot-config` looks for user data away from the "legacy" location in `/etc`.  The previous location was primarily for development purposes early in the projects' life.  Putting it in `/var/lib/bottlerocket` means it will live on the `PRIVATE` partition, and be relatively easy for provisioning systems to mount/write to.  It also means that it will be read restricted by SELinux policy.
```
    early-boot-config: Change location of local file user data
    
    This changes the location where early-boot-config looks for user data on
    the local filesystem to `/var/lib/bottlerocket/user-data.toml`
```
```
    early-boot-config: Add metal provider
    
    This change adds an additional provider in `early-boot-config` that is
    meant for variants that run on bare metal.  It looks for user data at a
    specific location on the filesystem
```
```
    variants: Add metal-dev variant
    
    This change adds an additional variant `metal-dev`, meant for
    development work on bare metal.  It includes similar defaults as the
    other `-dev` variants, and includes Docker for running containers, as
    well as the `login` package for a local prompt
```

**Testing done:**
*Hardware description*
The below testing was performed on 2 [Supermicro SYS-E200-8D](https://www.supermicro.com/en/products/system/Mini-ITX/SYS-E200-8D.cfm) machines.  The machines are both equipped with Intel 1G/10G NICs, an onboard 250G NVME drive, and a 1TB SATA drive.  The machines have a dedicated IPMI interface, and I made use of that functionality during all testing (SOL, chassis power control, etc).  The machines have the ability to boot via UEFI and BIOS.

*Provisioning*
Bottlerocket images were built with the "unified" layout, and images were written directly to the desired disk.  At provisioning time, partition 12 (`BOTTLEROCKET-PRIVATE`) was mounted, and user data was written there at `user-data.toml`.  

*Hardware configuration*
The following configurations were booted and able to run containers successfully via Docker.  I get full output from grub and the kernel on the local console and via serial console.
* BIOS: Boot from SATA  (BIOS / NVME isn't supported by my hardware)
* UEFI: Boot from NVME
* UEFI: Boot from SATA

**Not tested:**
* Intel 10G: (I don't have the ability to test 10G here, but do see that the hardware is recognized and modules/drivers are loaded properly)
* OS/Data split:  I used "unified" images throughout my testing and did not test the ability to put the OS / data on different disks.  I do expect it to work however, given that our systemd units looks for a partition label.
* Multiple NICs: Our current network setup uses the first interface `eth0`, and I didn't test other network configurations.  I _did_ mistakenly plug a cable into the second 1G port and because that port doesn't come up as `eth0`, networking was broken.

**Additional testing:**
I built and successfully ran the following variants to ensure that the kernel changes didn't affect their functionality.  Nodes joined the cluster properly and were able to run pods.
* `aws-k8s-1.21` (EC2 virtualized and metal)
* `vmware-k8s-1.21`

**Terms of contribution:**


variants/metal-dev/Cargo.toml
```diff
@@ -0,0 +1,58 @@
...
+
+[package.metadata.build-variant.image-layout]
+partition-plan = "unified"
+data-image-size-gib = 5
+
+[package.metadata.build-variant]
+image-format = "raw"
+supported-arches = ["x86_64"]
+kernel-parameters = [
+    "console=ttyS1,115200n8",
```
zmrow: Considering that this will be different for most hardware configurations, we will need to come up with a more "dynamic" way to pass these parameters to the kernel command line at runtime.  For now since this image must be built by the developer, we will leave this up to them to change at build time.

variants/README.md
```diff
@@ -88,6 +88,12 @@ If user data exists at both places, settings read from guestinfo will override i
 
 This variant is compatible with Kubernetes 1.21, 1.22, and 1.23 clusters.
 
+### metal-dev: Metal development build
+
+The [metal-dev](metal-dev/Cargo.toml) variant has useful packages for local development of the OS, and is intended to run bare metal.
```
jpculp:
```suggestion
The [metal-dev](metal-dev/Cargo.toml) variant has useful packages for local development of the OS and is intended to run on bare-metal.
```
nit.

variants/README.md
```diff
@@ -88,6 +88,12 @@ If user data exists at both places, settings read from guestinfo will override i
...
+It includes tools for troubleshooting as well as Docker for running containers.
+User data will be read from a file located at `/var/lib/bottlerocket/user-data.toml`.
```
jpculp:
```suggestion
User data will be read from `/var/lib/bottlerocket/user-data.toml`.
```

jpculp (approved):

webern (approved):
Cool, looks great.
Maybe the dev variants could share their defaults.d/50-*-dev.toml settings?

variants/metal-dev/Cargo.toml
arnaldo2792: I don't think we can change the kernel command line at runtime if we support secure boot. ~~I might be wrong :thinking: .~~

I do recall that depending on how we support secure boot, we won't be able to update the kernel command line at runtime:

https://www.linuxjournal.com/content/take-control-your-pc-uefi-secure-boot

sources/models/src/metal-dev/defaults.d/60-lockdown-none.toml
```diff
@@ -0,0 +1 @@
+../../../shared-defaults/lockdown-none.toml
```
arnaldo2792, line 1: Nit: Maybe we should try to update the `*-dev` variants to use the `lockdown` mode, since we already use that configuration whenever we release a new variant.

variants/README.md
arnaldo2792: Nit: Maybe some hints on how to add that file?

arnaldo2792 (approved):

variants/metal-dev/Cargo.toml
```diff
@@ -0,0 +1,58 @@
...
+version = "0.1.0"
+edition = "2018"
+publish = false
+build = "build.rs"
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[package.metadata.build-variant.image-layout]
+partition-plan = "unified"
+data-image-size-gib = 5
```
bcressey: Why `5` instead of the default `1`?

etungsten (approved):

sources/models/src/metal-dev/defaults.d/60-lockdown-none.toml
zmrow: For production variants absolutely - do we want it for development variants?

zmrow:
^ Fixed up the wording as requested by @jpculp .

Also removed `data-image-size-gib` from the `image-layout` section.

jpculp (approved):

zmrow:
^ Swaps the order of `console=` kernel command line arguments so the serial port gets both kernel _and_ systemd output.

I also added an additional commit to add the `metal-dev` variant to our GH checks.

sources/api/early-boot-config/src/provider.rs
```diff
@@ -16,6 +16,11 @@ mod vmware;
 #[cfg(bottlerocket_platform = "vmware")]
 pub(crate) use vmware::VmwareDataProvider as Platform;
 
+#[cfg(bottlerocket_platform = "metal")]
+mod metal;
+#[cfg(bottlerocket_platform = "metal")]
+pub(crate) use metal::MetalDataProvider as Platform;
+
```
bcressey, line 23: I'd like to see us drop the `aws-dev` special case rules and just treat local file in `/var/lib/bottlerocket` as a valid source for all platforms.

That will streamline the platforms we support to just the three (`aws`, `vmware`, `metal`) rather than "three and a little more."

bcressey (approved):

variants/metal-dev/Cargo.toml
zmrow: That's outdated - it's setup to use the default `1` now. :)

sources/api/early-boot-config/src/provider.rs
zmrow: Opened a new issue for this request - I'll take care of it!

PR #1903, opened 2022-01-13
Make local file a valid source for all platforms

### LINKED ISSUE #1902: `early-boot-config`: Make local file a valid source for all platforms
I'd like to see us drop the `aws-dev` special case rules and just treat local file in `/var/lib/bottlerocket` as a valid source for all platforms.

That will streamline the platforms we support to just the three (`aws`, `vmware`, `metal`) rather than "three and a little more."

_Originally posted by @bcressey in https://github.com/bottlerocket-os/bottlerocket/pull/1895#discussion_r783448693_

zmrow:
**Issue number:**
Fixes #1902 

**Description of changes:**
```
early-boot-config: Make local file a valid source for all platforms

This change adds local file as a valid user data source for all
platforms we support.  Previously local file was only supported on
`aws-dev` and `metal-dev` variants.  A nice side affect of this change
is that it simplifies the platforms we support and streamlines the
conditional compilation a bit more.
```
```
variants: Document user data sources

This commit centralizes information regarding the various sources from
which variants read user data.  It also explains the order in which
sources are read, and how identical settings are handled.
```

**Testing done:**
*metal-dev*
Built an image with a user data file and observed the image boot and ingest user data.  Built another image without the file and the system booted successfully, logging that no user data was present at the desired location.

*aws-k8s-1.21*
*vmware-k8s-1.21*
Build an image with a local file user data containing:
```
settings.motd = "HELLO FROM FILE"
settings.host-containers.admin.enabled = false
```
At runtime I supplied user data to effectively "override" and enable the admin container: `settings.host-containers.admin.enabled = true`.  For both variants I observed via `early-boot-config` logs and the `apiclient` commands below, that the user data was correctly ingested via local file.  The settings ingested from user data via IMDS/guestinfo at runtime correctly overwrote the settings provided via the file.
```
bash-5.0# apiclient -u /settings?keys=settings.motd
{"motd":"HELLO FROM FILE"}

bash-5.0# apiclient -u /settings?keys=settings.host-containers.admin.enabled
{"host-containers":{"admin":{"enabled":true}}}
```
I conducted the same experiment except supplied user data at runtime that _would not_ overwrite the settings supplied via file and observed the correct behavior.
```
bash-5.0# apiclient -u /settings?keys=settings.motd
{"motd":"HELLO FROM FILE"}

$ apiclient -u /settings?keys=settings.host-containers.admin.enabled
{"host-containers":{"admin":{"enabled":false}}}
```

For both variants, I also built images without a local user data file and observed both correctly check for user data at the file location, find nothing, and move on to the other sources.

**Terms of contribution:**

### DIFF
--- a/sources/api/early-boot-config/src/provider/aws.rs
+++ b/sources/api/early-boot-config/src/provider/aws.rs
@@ -9,7 +9,6 @@ use snafu::{OptionExt, ResultExt};
 use std::fs;
 use std::path::Path;
 
-#[cfg(bottlerocket_platform = "aws-dev")]
 use crate::provider::local_file::{local_file_user_data, USER_DATA_FILE};
 
 /// Unit struct for AWS so we can implement the PlatformDataProvider trait.
@@ -87,25 +86,22 @@ impl PlatformDataProvider for AwsDataProvider {
 
         let mut client = ImdsClient::new();
 
-        // Attempt to read from local file first on the `aws-dev` variant
-        #[cfg(bottlerocket_platform = "aws-dev")]
-        {
-            match local_file_user_data()? {
-                None => warn!("No user data found via local file: {}", USER_DATA_FILE),
-                Some(s) => output.push(s),
-            }
+        // Attempt to read from local file first
+        match local_file_user_data()? {
+            Some(s) => output.push(s),
+            None => warn!("No user data found via local file: {}", USER_DATA_FILE),
         }
 
         // Instance identity doc next, so the user has a chance to override
         match Self::identity_document(&mut client).await? {
-            None => warn!("No instance identity document found."),
             Some(s) => output.push(s),
+            None => warn!("No instance identity document found."),
         }
 
         // Optional user-specified configuration / overrides
         match Self::user_data(&mut client).await? {
-            None => warn!("No user data found."),
             Some(s) => output.push(s),
+            None => warn!("No user data found."),
         }
 
         Ok(output)
--- a/sources/api/early-boot-config/src/provider/metal.rs
+++ b/sources/api/early-boot-config/src/provider/metal.rs
@@ -16,8 +16,8 @@ impl PlatformDataProvider for MetalDataProvider {
         let mut output = Vec::new();
 
         match local_file_user_data()? {
-            None => warn!("No user data found via local file: {}", USER_DATA_FILE),
             Some(s) => output.push(s),
+            None => warn!("No user data found via local file: {}", USER_DATA_FILE),
         }
 
         Ok(output)
--- a/sources/api/early-boot-config/src/provider/vmware.rs
+++ b/sources/api/early-boot-config/src/provider/vmware.rs
@@ -15,6 +15,8 @@ use std::iter::FromIterator;
 use std::path::Path;
 use std::str;
 
+use crate::provider::local_file::{local_file_user_data, USER_DATA_FILE};
+
 pub(crate) struct VmwareDataProvider;
 
 impl VmwareDataProvider {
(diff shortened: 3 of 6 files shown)


sources/api/early-boot-config/src/provider/aws.rs
```diff
@@ -87,25 +86,22 @@ impl PlatformDataProvider for AwsDataProvider {
...
+        match local_file_user_data()? {
+            Some(s) => output.push(s),
+            None => warn!("No user data found via local file: {}", USER_DATA_FILE),
         }
 
         // Instance identity doc next, so the user has a chance to override
         match Self::identity_document(&mut client).await? {
-            None => warn!("No instance identity document found."),
             Some(s) => output.push(s),
+            None => warn!("No instance identity document found."),
```
zmrow, line 98: Here I moved the `None` below the `Some` to make these `match`es uniform with the other providers.

sources/api/early-boot-config/src/provider/metal.rs
```diff
@@ -16,8 +16,8 @@ impl PlatformDataProvider for MetalDataProvider {
         let mut output = Vec::new();
 
         match local_file_user_data()? {
-            None => warn!("No user data found via local file: {}", USER_DATA_FILE),
             Some(s) => output.push(s),
+            None => warn!("No user data found via local file: {}", USER_DATA_FILE),
```
zmrow, line 20: I did the same here.

etungsten (approved):
:icecream:

sources/api/early-boot-config/src/provider/vmware.rs
```diff
@@ -243,6 +245,12 @@ impl PlatformDataProvider for VmwareDataProvider {
     ) -> std::result::Result<Vec<SettingsJson>, Box<dyn std::error::Error>> {
         let mut output = Vec::new();
 
+        // Attempt to read from local file first
+        match local_file_user_data()? {
+            Some(s) => output.push(s),
+            None => warn!("No user data found via local file: {}", USER_DATA_FILE),
+        }
+
         // Look at the CD-ROM for user data first, and then...
```
bcressey: nit: no longer first

variants/README.md
```diff
@@ -22,6 +22,15 @@ A variant is essentially a list of packages to install, plus a model that define
...
 Information about API settings for variants can be found in the [models](../sources/models/) documentation.
 
+### User data
+Bottlerocket variants ingest TOML-formatted [user data](../README.md#using-user-data) from various sources.
+User data is read from sources in order; the most recently read user data source overrides previously read sources.
+All variants first attempt to read user data from `/var/lib/bottlerocket/user-data.toml`.
+AWS variants then retrieve user data from IMDS.
+VMware variants will attempt to read user data from a mounted CD-ROM (from a file named "user-data" or from an OVF file), and then from VMware's guestinfo interface.
+
+For example, in a VMware variant, settings read from the guestinfo interface override settings from CD-ROM, and settings from CD-ROM override settings from file.
```
bcressey:
```suggestion
Bottlerocket variants ingest TOML-formatted [user data](../README.md#using-user-data) from various sources in a predefined order.
All variants first attempt to read user data from `/var/lib/bottlerocket/user-data.toml`.
AWS variants then retrieve user data from IMDS.
VMware variants will attempt to read user data from a mounted CD-ROM (from a file named "user-data" or from an OVF file), and then from VMware's guestinfo interface.

If a setting is defined in more than one source, the value in later sources will override earlier values.
For example, in a VMware variant, settings read from the guestinfo interface will override settings from CD-ROM, and settings from CD-ROM will override settings from the file.
```

bcressey (commented):
🍰

zmrow:
^ Wording fixes per @bcressey (thanks!)

bcressey (approved):

PR #1904, opened 2022-01-14
variants: Add metal-k8s-1.21 variant

zmrow:
**Description of changes:**
Following along the series of changes in the `metal-dev` PR (#1895), this PR adds a variant meant for running Bottlerocket as a worker node in Kubernetes clusters on bare metal.  It is similar in scope to the other Kubernetes variants, and includes sane defaults for running off AWS.
```
This change adds an additional variant `metal-k8s-1.21`, which includes
necessary Kubernetes packages and settings for running Bottlerocket on
metal in a Kubernetes cluster.
```
```
VMware and metal variants both use host containers pulled from public
ECR.  This change creates a single file that both variants can use.
```
```
VMware and metal variants use the same logdog config, so this change
renames the config to `k8s.conf` and fixes up the symlinks accordingly.
```

**Testing done:**
Similar to #1895 , all testing was performed on 2 [Supermicro SYS-E200-8D](https://www.supermicro.com/en/products/system/Mini-ITX/SYS-E200-8D.cfm) machines.  Testing was performed using the following configurations:
* BIOS: Boot from SATA (BIOS / NVME isn't supported by my hardware)
* UEFI: Boot from NVME
* UEFI: Boot from SATA

Tests:
* Ran kubelet in standalone mode and confirmed it was able to run a redis pod via static pods. (I don't currently have a Kubernetes cluster at home to join)
* All host containers worked swimmingly given the correct user data
  * Admin container / sheltie works as normal
  * Used SSM to start a session on the control container and all is well
* Bootstrap containers are happy
* Inspected journal logs, nothing alarming there (though kubelet would _really_ like a CNI plugin :) )

* Successfully built and ran the vmware-k8s-1.21 variant to ensure I didn't break anything there. :)

**Terms of contribution:**

T license.


sources/models/shared-defaults/kubernetes-metal.toml
```diff
@@ -0,0 +1,19 @@
+[settings.kubernetes]
+cluster-domain = "cluster.local"
+standalone-mode = false
+authentication-mode = "tls"
+pod-infra-container-image = "k8s.gcr.io/pause:3.2"
```
bcressey: We should switch these over to the ECR public images:
```suggestion
pod-infra-container-image = "public.ecr.aws/eks-distro/kubernetes/pause:3.5"
```

sources/models/shared-defaults/metal-host-containers.toml
```diff
@@ -0,0 +1,14 @@
...
+# either of those options.
+[settings.host-containers.admin]
+enabled = false
+superpowered = true
+source = "public.ecr.aws/bottlerocket/bottlerocket-admin:v0.7.2"
+
+[settings.host-containers.control]
+enabled = false
+superpowered = false
+source = "public.ecr.aws/bottlerocket/bottlerocket-control:v0.5.2"
```
bcressey: Need to make sure these are the latest so we don't end up with different versions for migrations to consider:

```suggestion
[settings.host-containers.admin]
enabled = false
superpowered = true
source = "public.ecr.aws/bottlerocket/bottlerocket-admin:v0.7.3"

[settings.host-containers.control]
enabled = false
superpowered = false
source = "public.ecr.aws/bottlerocket/bottlerocket-control:v0.5.4"
```

Alternatively we might want to rename the file "public-host-containers.toml" so we don't multiply the number of references to the same images.

variants/metal-k8s-1.21/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+publish = false
+build = "build.rs"
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[package.metadata.build-variant.image-layout]
+partition-plan = "unified"
+
+[package.metadata.build-variant]
+image-format = "raw"
```
bcressey, line 17: Let's also expand the OS image size to 4 GiB (two banks of 2 GiB) - since I expect to need some extra space for firmware, etc.

sources/logdog/conf/metal-k8s.conf
```diff
@@ -0,0 +1 @@
+exec kube-status systemctl status kube* -l --no-pager
```
webern: Is this different from the other k8s variants? (edit: the underlying question is I was wondering if there's a shared file or if we should make one)

webern (approved):

sources/logdog/conf/metal-k8s.conf
zmrow: The `aws` file differs, but we could probably unify the `vmware` and `metal` files since they're the same, and both are "off-AWS" variants.

zmrow:
^ Added 2 additional commits: the first unifies the public host containers into one file in models, the second unifies the "off-AWS" config into a single file in logdog.

etungsten (approved):

bcressey (approved):

PR #1910, opened 2022-01-15
models: add missing serde attributes to `registrysettings.mirrors` field

### LINKED ISSUE #1909: Passing an empty TOML table for `container-registry` causes boot failure
**Image I'm using:**
Bottlerocker v1.5.2

When I pass an empty TOML table for `container-registry` settings, the host should boot up fine.
For example
```
[settings.container-registry]
```

**What actually happened:**

The host fails to start-up expecting `mirrors` to be specified, even though it should be optional:
```
[    4.831386] early-boot-config[1425]: Error PATCHing '/settings?tx=bottlerocket-launch': Status 400 when PATCHing /settings?tx=bottlerocket-launch: Json deserialize error: missing field `mirrors` at line 1 column 24
```

Specify an empty TOML table for `container-registry` settings:
```
[settings.container-registry]
```

etungsten:
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/1909

**Description of changes:**
```
Date:   Fri Jan 14 17:01:00 2022 -0800

    models: add missing serde attributes to `registrysettings.mirrors` field
    
    Because we're using a custom deserializer for `mirrors`, we need to also
    add the serde attribute for populating the field with a default value if
    mirrors does not exist in the registry settings table
```

**Testing done:**
Launching an instance with an empty `container-registry` table:
```
[settings.container-registry]
```
And the instance comes up fine. The datastore also doesn't have any thing created for the empty table as expected.

**Terms of contribution:**

### DIFF
--- a/sources/models/src/de.rs
+++ b/sources/models/src/de.rs
@@ -63,6 +63,12 @@ mod mirrors_tests {
     }
 
     #[test]
+    fn registry_mirrors_none_representation() {
+        let registry_settings = toml::from_str::<RegistrySettings>("").unwrap();
+        assert!(registry_settings.mirrors.is_none());
+    }
+
+    #[test]
     fn representation_equal() {
         assert_eq!(
             toml::from_str::<RegistrySettings>(TEST_MIRRORS_TABLE).unwrap(),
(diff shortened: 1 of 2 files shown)


webern (approved):

arnaldo2792 (approved):

zmrow (approved):
:fondue:

PR #1916, opened 2022-01-19
packages: add NVIDIA k8s device plugin

arnaldo2792:
**Issue number:**
N / A

**Description of changes:**
```
packages: add NVIDIA k8s device plugin

The NVIDIA k8s device plugin exposes to the kubelet the NVIDIA GPUs in
the host. By providing the k8s device plugin in the base image, the
startup time for a GPU workload decreases, since the kubelet doesn't
have to fetch yet another system container.
```

**Testing done:**
I launched a nodegroup without the NVIDIA k8s device plugin daemonset:

```
eksctl create nodegroup --install-nvidia-plugin=false --config-file=../k8s-1.21.yamlh
```

I was able to launch a pod in the new host and run `nvidia-smi` within the host

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Start NVIDIA kubernetes device plugin
+RefuseManualStart=true
+RefuseManualStop=true
+After=kubelet.service
+BindsTo=kubelet.service
+
+[Service]
+ExecStart=/usr/bin/nvidia-device-plugin
+Type=simple
+TimeoutSec=0
+RestartSec=2
+Restart=always
+StandardError=journal+console
+
+[Install]
+WantedBy=multi-user.target
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
@@ -1,0 +1,43 @@
+%global goproject github.com/NVIDIA
+%global gorepo k8s-device-plugin
+%global goimport %{goproject}/%{gorepo}
+
+%global gover 0.10.0
+%global rpmver %{gover}
+
+Name: %{_cross_os}nvidia-k8s-device-plugin
+Version: %{rpmver}
+Release: 1%{?dist}
+Summary: Kubernetes device plugin for NVIDIA GPUs
+License: Apache-2.0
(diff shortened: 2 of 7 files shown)


arnaldo2792:
Forced push adds `releases-url`

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Start NVIDIA k8s device plugin
+RefustManualStart=true
```
bcressey:
```suggestion
RefuseManualStart=true
```

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
```diff
@@ -0,0 +1,46 @@
...
+%global gorepo k8s-device-plugin
+%global goimport %{goproject}/%{gorepo}
+
+%global gover 0.10.0
+%global rpmver %{gover}
+
+Name: %{_cross_os}nvidia-k8s-device-plugin
+Version: %{rpmver}
+Release: 1%{?dist}
+Summary: K8s device plugin for NVIDIA GPUs
```
bcressey: super-nit:
```suggestion
Summary: Kubernetes device plugin for NVIDIA GPUs
```

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
```diff
@@ -0,0 +1,46 @@
...
+%cross_go_setup %{gorepo}-%{gover} %{goproject} %{goimport}
+# We don't set `-Wl,-z,now`, because the binary uses lazy loading
+# to load the NVIDIA libraries in the host
+%global set_env \
+%cross_go_configure %{goimport} \\\
+export CGO_LDFLAGS="-Wl,-z,relro" \\\
+%{nil}
+
+%build
+%set_env
```
bcressey: nit: this is fine but I'd probably do it like this just so `%build` looks more like the other Go packages -

```suggestion
%build
%cross_go_configure %{goimport}
# We don't set `-Wl,-z,now`, because the binary uses lazy loading
# to load the NVIDIA libraries in the host
export CGO_LDFLAGS="-Wl,-z,relro"
```

arnaldo2792:
Forced pushes address:

- Typo in systemd unit
- Summary description in k8s-device-plugin spec
- Changed compilation commands in spec file to be more like other go packages

bcressey (approved):

jpculp (approved):

PR #1928, opened 2022-01-26
kmod-5.10-nvidia: add remaining libraries

### LINKED ISSUE #1822: Add additional NVIDIA libraries to aws-k8s-1.21-nvidia variant
**What I'd like:**
Due to size constraints, we didn't add the libraries required for all GPU workloads to work in the aws-k8s-1.21-nvidia variant. Once  #1821 is implemented, we should add the additional libraries to the variant.

N / A

arnaldo2792:
**Issue number:**
Closes #1822

**Description of changes:**
```
kmod-5.10-nvidia: add remaining libraries
kmod-5.10-nvidia: add releases url
```

The NVIDIA sources provide user-space libraries that will be mounted into the containers, depending on the set of driver capabilities configured for the workload.

**Testing done:**

I ran a daemonset in a `p3.2xlarge` instance with the following image definition:

```Dockerfile
FROM nvidia/cuda:11.4.3-devel-ubuntu20.04 as cuda-samples

RUN apt update
RUN apt install git build-essential -y
RUN git clone https://github.com/NVIDIA/cuda-samples.git

# Compute samples
RUN mkdir -p /samples
RUN cd /cuda-samples/Samples/0_Introduction/vectorAdd && make -j && [ -f vectorAdd ] && cp vectorAdd /samples/
RUN cd /cuda-samples/Samples/1_Utilities/bandwidthTest && make -j && [ -f bandwidthTest ] && cp bandwidthTest /samples/
RUN cd /cuda-samples/Samples/1_Utilities/deviceQuery && make -j && [ -f deviceQuery ] && cp deviceQuery /samples/
RUN cd /cuda-samples/Samples/1_Utilities/topologyQuery && make -j && [ -f topologyQuery ] && cp topologyQuery /samples/

FROM alpine as builder
RUN apk update \
  && apk add --update git

FROM builder as benchmarks
RUN git clone https://github.com/tensorflow/benchmarks.git \
  && cd benchmarks \
  && git checkout cnn_tf_v1.15_compatible

FROM tensorflow/tensorflow:1.15.2-gpu
ENV SAMPLES="vectorAdd bandwidthTest deviceQuery topologyQuery"
COPY ./entrypoint.sh /
COPY --from=benchmarks /benchmarks /opt/benchmarks
COPY --from=cuda-samples /samples/* /usr/bin/
RUN chmod +x ./entrypoint.sh && mkdir -p /opt
ENTRYPOINT ["sh", "-c", "/entrypoint.sh"]
```

Entrypoint:

```sh
#! /usr/bin/env bash

# Cuda samples:

for sample in $SAMPLES; do
  $sample
done

# GPU benchmark:
python3 /opt/benchmarks/scripts/tf_cnn_benchmarks/tf_cnn_benchmarks.py \
  --data_name=imagenet                                                 \
  --model=resnet50                                                     \
  --num_batches=100                                                    \
  --batch_size=4                                                       \
  --num_gpus=1                                                         \
  --gpu_memory_frac_for_testing=0.2
```

The containers ran successfully.

TODO:

- [x] Same test for aarch64

**Terms of contribution:**

### DIFF
--- a/packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
+++ b/packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
@@ -1,4 +1,6 @@
-%global nvidia_tesla_470_version 470.82.01
+%global tesla_470 470.82.01
+%global tesla_470_libdir %{_cross_libdir}/nvidia/tesla/%{tesla_470}
+%global tesla_470_bindir %{_cross_libexecdir}/nvidia/tesla/bin/%{tesla_470}
 %global spdx_id %(bottlerocket-license-tool -l %{_builddir}/Licenses.toml spdx-id nvidia)
 %global license_file %(bottlerocket-license-tool -l %{_builddir}/Licenses.toml path nvidia -p ./licenses)
 
@@ -12,8 +14,8 @@ License: Apache-2.0 OR MIT
 URL: http://www.nvidia.com/
 
 # NVIDIA .run scripts from 0 to 199
-Source0: https://us.download.nvidia.com/tesla/%{nvidia_tesla_470_version}/NVIDIA-Linux-x86_64-%{nvidia_tesla_470_version}.run
-Source1: https://us.download.nvidia.com/tesla/%{nvidia_tesla_470_version}/NVIDIA-Linux-aarch64-%{nvidia_tesla_470_version}.run
+Source0: https://us.download.nvidia.com/tesla/%{tesla_470}/NVIDIA-Linux-x86_64-%{tesla_470}.run
+Source1: https://us.download.nvidia.com/tesla/%{tesla_470}/NVIDIA-Linux-aarch64-%{tesla_470}.run
 
 # Common NVIDIA conf files from 200 to 299
 Source200: nvidia-tmpfiles.conf.in
@@ -33,7 +35,7 @@ BuildRequires: %{_cross_os}kernel-5.10-archive
 
 %package tesla-470
 Summary: NVIDIA 470 Tesla driver
-Version: %{nvidia_tesla_470_version}
+Version: %{tesla_470}
 License: %{spdx_id}
 Requires: %{name}
 
@@ -43,13 +45,13 @@ Requires: %{name}
 %prep
 # Extract nvidia sources with `-x`, otherwise the script will try to install
 # the driver in the current run
-sh %{_sourcedir}/NVIDIA-Linux-%{_cross_arch}-%{nvidia_tesla_470_version}.run -x
+sh %{_sourcedir}/NVIDIA-Linux-%{_cross_arch}-%{tesla_470}.run -x
 
 %global kernel_sources %{_builddir}/kernel-devel
 tar -xf %{_cross_datadir}/bottlerocket/kernel-devel.tar.xz
 
 %build
-pushd NVIDIA-Linux-%{_cross_arch}-%{nvidia_tesla_470_version}/kernel
+pushd NVIDIA-Linux-%{_cross_arch}-%{tesla_470}/kernel
 
 # This recipe was based in the NVIDIA yum/dnf specs:
 # https://github.com/NVIDIA/yum-packaging-precompiled-kmod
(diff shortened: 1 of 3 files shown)


packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -136,20 +136,115 @@ install kernel/nvidia-peermem/nvidia-peermem.o %{buildroot}%{_cross_datadir}/nvi
...
 install -m 755 nvidia-smi %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}
 install -m 755 nvidia-debugdump %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}
 install -m 755 nvidia-cuda-mps-control %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}
 install -m 755 nvidia-cuda-mps-server %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}
+
+# We only install the libraries required by all the DRIVER_CAPABILITIES, described here:
+# https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/user-guide.html#driver-capabilities
+
+# Utility libs
+install -m755 libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
```
bcressey: Consider defining something like:
```
%global nvidia_tesla_470_dir %{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
```

And then using it to cut down on some of the repetition.

Might also be good to shorten `nvidia_tesla_470_version` to just `tesla_470`.

jpculp:
```suggestion
install -m 755 libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
```

nit. You might want to add a space to be consistent with the preexisting code, but it seems to work either way.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -136,20 +136,115 @@ install kernel/nvidia-peermem/nvidia-peermem.o %{buildroot}%{_cross_datadir}/nvi
...
+ln -s libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/libnvidia-ml.so.1
```
bcressey: What is this `libnvidia-ml.so.1` symlink for? Does it match the library's embedded soname?

```
readelf -a <lib>|grep SONAME
```

jpculp (approved):

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -136,20 +136,115 @@ install kernel/nvidia-peermem/nvidia-peermem.o %{buildroot}%{_cross_datadir}/nvi
 install kernel/nvidia-drm.mod.o %{buildroot}/%{_cross_datadir}/nvidia/tesla/%{nvidia_tesla_470_version}/module-objects.d
 install kernel/nvidia-drm.o %{buildroot}/%{_cross_datadir}/nvidia/tesla/%{nvidia_tesla_470_version}/module-objects.d
 
+# Binaries
 install -m 755 nvidia-smi %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}
 install -m 755 nvidia-debugdump %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}
 install -m 755 nvidia-cuda-mps-control %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}
 install -m 755 nvidia-cuda-mps-server %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}
+
+# We only install the libraries required by all the DRIVER_CAPABILITIES, described here:
```
bcressey: How do we map "utility" to the actual library list?
* libnvidia-ml.so
* libnvidia-cfg.so

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -136,20 +136,115 @@ install kernel/nvidia-peermem/nvidia-peermem.o %{buildroot}%{_cross_datadir}/nvi
...
+# https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/user-guide.html#driver-capabilities
+
+# Utility libs
+install -m755 libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
+ln -s libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/libnvidia-ml.so.1
+
+install -m755 libnvidia-cfg.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
+ln -s libnvidia-cfg.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/libnvidia-cfg.so
+ln -s libnvidia-cfg.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/libnvidia-cfg.so.1
```
bcressey: Why create a `libnvidia-cfg.so` symlink but not `libnvidia-ml.so`?

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -136,20 +136,115 @@ install kernel/nvidia-peermem/nvidia-peermem.o %{buildroot}%{_cross_datadir}/nvi
...
+# Binaries
 install -m 755 nvidia-ngx-updater %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}
-%endif
 
-# TODO: add remaining libraries
-# misc
-# Add libnvidia-ml.so for testing purposes
-install -m755 libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
+# Compute libs
+install -m755 libnvidia-compiler.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
```
bcressey: nit: it's a style thing, but I'd personally do these arch conditionals in the same sections of the spec as the other libraries in the category - compute, etc.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -177,8 +272,7 @@ popd
 %{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}/nvidia-smi
 
 # Libraries
-%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/libnvidia-ml.so.1
-%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/libnvidia-ml.so.%{nvidia_tesla_470_version}
+%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/*.so*
```
bcressey: This suggestion is part style, but mostly about future-proofing.

One concern I have with picking and choosing libraries in the `%install` section is that if new libraries appear in a driver update, we might not notice and then we'd ship without them.

An alternative that might help would be to bulk install all libraries, and create all the .so and .so.1 symlinks (if those are truly needed). Then down here in the `%files` section, group them by "compute", "graphics" etc, and exclude everything that we don't intend to ship.

That way, when we update the driver, the build will fail if a new library shows up, and we'll have to add it to the right section or exclude it as well.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
arnaldo2792: Yes, it matches the `SONAME` in the library. `libnvidia-container` will try to resolve the SONAME. When I was adding the libraries, and testing my workloads, I saw that `libnvidia-container` was complaining about some missing symlinks. So whenever I saw a failure, I ran `readelf -a | grep SONAME` in a library and created the symlinks for it. As I said, I'll confirm that the extra symlinks (.so) aren't required.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
arnaldo2792: Let me double check if it isn't needed. I think I created this link because I was following what arch does.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
arnaldo2792: :smile: I had it as you describe it, but I thought that people wouldn't like all the `if`s in the spec. I'll change it back.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
arnaldo2792: `libnvidia-container` categorizes the libraries in the groups I used,  you can see it here:

https://github.com/NVIDIA/libnvidia-container/blob/d48f9b0d505fca0aff7c88cee790f9c56aa1b851/src/nvc_info.c#L61

arnaldo2792:
Forced push includes:

- Install all libraries, and explicitly include and exclude the libraries in the `%files` section
- Short global variable `nvidia_tesla_470_version`
- Create only the required symlinks, based on the output of `readelf -a <lib> | grep SONAME`

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -90,66 +92,90 @@ install -d %{buildroot}%{_cross_libdir}/modules-load.d
...
-# Add libnvidia-ml.so for testing purposes
-install -m755 libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
-
-ln -s libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/libnvidia-ml.so.1
+# We install all the libraries, and filter them out in the 'files' section, so we can catch
+# when new libraries are added
+install -m 755 *.so* %{buildroot}/%{tesla_470_libdir}/
+
+# Add required symlinks
+ln -s libnvidia-ml.so.%{tesla_470} %{buildroot}%{tesla_470_libdir}/libnvidia-ml.so.1
```
bcressey: Would something like this work to automate creating the correct links?

```
for lib in $(find . -maxdepth 1 -type f -name 'lib*.so.*' -printf '%P\n') ; do
  soname="$(%{_cross_target}-readelf -d "${lib}" | awk '/SONAME/{print $5}'|tr -d '[]')"
  [ -n "${soname}" ] || continue
  [ "${lib}" == "${soname}" ] && continue
  ln -s "${lib}" "${soname}"
done
```

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -166,50 +192,135 @@ popd
...
+%exclude %{_cross_datadir}/nvidia/tesla/%{tesla_470}/module-objects.d/nvidia-drm.mod.o
+%exclude %{_cross_datadir}/nvidia/tesla/%{tesla_470}/module-objects.d/nvidia-drm.o
+%exclude %{_cross_libexecdir}/nvidia/tesla/bin/%{tesla_470}/nvidia-cuda-mps-control
+%exclude %{_cross_libexecdir}/nvidia/tesla/bin/%{tesla_470}/nvidia-cuda-mps-server
 %if "%{_cross_arch}" == "x86_64"
-%exclude %{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}/nvidia-ngx-updater
+%exclude %{_cross_libexecdir}/nvidia/tesla/bin/%{tesla_470}/nvidia-ngx-updater
 %endif
+
+# None of these libraries are required
```
bcressey: Let's expand the comment a bit to say why they're not required.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
bcressey: Might be good to include this link in the comment as well!

etungsten (approved):
I tested both aarch64 and x86_64 builds with benchmarks and samples and they pass.

arnaldo2792:
Forced push includes:

- Automated symlink creation
- Explanation on why some libraries are excluded

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -90,66 +92,77 @@ install -d %{buildroot}%{_cross_libdir}/modules-load.d
...
-install -m755 libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
-
-ln -s libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/libnvidia-ml.so.1
+# We install all the libraries, and filter them out in the 'files' section, so we can catch
+# when new libraries are added
+install -m 755 *.so* %{buildroot}/%{tesla_470_libdir}/
+
+# These libraries have the same SONAME as other required libraries, we don't
+# create a symlink for these to prevent collisions
+EXCLUDED_LIBS="libEGL.so.1.1.0"
```
arnaldo2792: `libEGL.so.1.1.0` and `libEGL.so.<version>` have the same SONAME, but they are different (I compared their checksums). I didn't find any reference about which library should be included. These are 32-bit compatible libraries, used in OpenGL programs. Ubuntu and Arch only include the 64-bit library called `libEGL_nvidia.so`, which is already included here. I decided to keep `libEGL.so.<version>`, since <version> matches the driver version.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -90,66 +92,77 @@ install -d %{buildroot}%{_cross_libdir}/modules-load.d
...
+for lib in $(find . -maxdepth 1 -type f -name 'lib*.so.*' -printf '%p\n'); do
+  # remove `./` from lib
+  lib="${lib//.\/}"
```
bcressey: Can we use `%P` instead?

```suggestion
for lib in $(find . -maxdepth 1 -type f -name 'lib*.so.*' -printf '%P\n'); do
```

arnaldo2792:
Forced push includes:

- Remove compat32 libs

jpculp (approved):

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -166,50 +184,128 @@ popd
...
-%exclude %{_cross_libexecdir}/nvidia/tesla/bin/%{nvidia_tesla_470_version}/nvidia-ngx-updater
+%exclude %{_cross_libexecdir}/nvidia/tesla/bin/%{tesla_470}/nvidia-ngx-updater
 %endif
+
+# None of these libraries are required by libnvidia-container, so they
+# won't be used by a containerized workload
+%exclude %{tesla_470_libdir}/libGLX.so.0
+%exclude %{tesla_470_libdir}/libGLdispatch.so.0
+%exclude %{tesla_470_libdir}/libOpenCL.so.1.0.0
+%exclude %{tesla_470_libdir}/libOpenGL.so.0
```
bcressey: With the exception of `libOpenCL.so.1.0.0` these are part of the interface provided by `libglvnd`, so it seems reasonable on the surface to expect containers to include that library. If we're going to remove the others, we should remove these too (except OpenCL).

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -90,66 +92,82 @@ install -d %{buildroot}%{_cross_libdir}/modules-load.d
...
+install -m 755 nvidia-ngx-updater %{buildroot}%{_cross_libexecdir}/nvidia/tesla/bin/%{tesla_470}
 %endif
 
-# TODO: add remaining libraries
-# misc
-# Add libnvidia-ml.so for testing purposes
-install -m755 libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}
-
-ln -s libnvidia-ml.so.%{nvidia_tesla_470_version} %{buildroot}%{_cross_libdir}/nvidia/tesla/%{nvidia_tesla_470_version}/libnvidia-ml.so.1
+# We don't install 32-bit compat libraries
```
bcressey: Not sure that the "32-bit" part is correct.

It seems like these were deprecated in favor of [libglvnd](https://github.com/NVIDIA/libglvnd) which provides GLdispatch, GL, EGL, GLX, GLESv1, GLESv2, and OpenGL.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -166,50 +184,128 @@ popd
...
+%{tesla_470_libdir}/libnvoptix.so.%{tesla_470}
+%{tesla_470_libdir}/libnvoptix.so.1
+%if "%{_cross_arch}" == "x86_64"
+%{tesla_470_libdir}/libnvidia-ifr.so.%{tesla_470}
+%{tesla_470_libdir}/libnvidia-ifr.so.1
+%endif
+
+# Graphics GLVND libs
+%{tesla_470_libdir}/libnvidia-glvkspirv.so.%{tesla_470}
+%{tesla_470_libdir}/libnvidia-cbl.so.%{tesla_470}
```
bcressey, line 267: These both seem to be Vulkan related - are you sure they're part of GLVND?

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -166,50 +184,128 @@ popd
...
+%exclude %{tesla_470_libdir}/libGLX.so.0
+%exclude %{tesla_470_libdir}/libGLdispatch.so.0
+%exclude %{tesla_470_libdir}/libOpenCL.so.1.0.0
+%exclude %{tesla_470_libdir}/libOpenGL.so.0
+%exclude %{tesla_470_libdir}/libglxserver_nvidia.so.470.82.01
+%exclude %{tesla_470_libdir}/libnvidia-allocator.so.470.82.01
+%exclude %{tesla_470_libdir}/libnvidia-egl-wayland.so.1.1.7
+%exclude %{tesla_470_libdir}/libnvidia-gtk2.so.470.82.01
+%exclude %{tesla_470_libdir}/libnvidia-gtk3.so.470.82.01
+%exclude %{tesla_470_libdir}/libnvidia-nvvm.so.4.0.0
```
bcressey: I think we want to keep both `allocator` and `nvvm` based on the [descriptions](https://download.nvidia.com/XFree86/Linux-x86_64/470.82.01/README/installedcomponents.html):
> Various libraries that are used internally by other driver components. These include ... /usr/lib/libnvidia-allocator.so.470.82.01.
> ...
> The NVVM Compiler library (/usr/lib/libnvidia-nvvm.so.4.0.0) is loaded by the CUDA driver to do JIT link-time-optimization.

Seems OK to exclude the desktop ones - libglxserver (for Xorg), libnvidia-egl-wayland, libnvidia-gtk2, libnvidia-gtk3.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -166,50 +184,128 @@ popd
...
+%exclude %{tesla_470_libdir}/libnvidia-opencl.so.470.82.01
+%exclude %{tesla_470_libdir}/libnvidia-vulkan-producer.so.470.82.01
```
bcressey: Do we need to include OpenCL, or is it expected to come from the CUDA base container?

This Vulkan library may need to be included because we have other Vulkan related components.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
```diff
@@ -166,50 +184,128 @@ popd
...
+%exclude %{tesla_470_libdir}/nvidia_drv.so
+%exclude %{tesla_470_libdir}/libOpenCL.so.1
+%exclude %{tesla_470_libdir}/libnvidia-egl-wayland.so.1
```
bcressey: Except (possibly) for OpenCL, these look OK to exclude - `nvidia_drv` is for the X server, and `libnvidia-egl-wayland` is Wayland related.

bcressey:
Still trying to come up with better advice for what to include, since our current method seems pretty high touch and error prone.

I'd like to err on the side of including everything and letting `libnvidia-container` sort it out, with the possible exception of the Gtk and Wayland stuff that we know is excluded.

It seems like the only problem with that plan is what to do about the libEGL.so.1 symlink, which should point to one of the two libraries with that SONAME, and perhaps one of them should be excluded.

bcressey:
> It seems like the only problem with that plan is what to do about the libEGL.so.1 symlink, which should point to one of the two libraries with that SONAME, and perhaps one of them should be excluded.

Let's point `libEGL.so.1` to `libEGL.so.1.1.0` since that seems to be how the other libglvnd libraries are treated. We can still include both. On a running instance afterwards, you can check to see whether `ldconfig --print-cache` agrees with that resolution.

If we could get the output from libnvidia-container when it's checking the compat libraries for inclusion, that might help determine whether either or both of them is OK. Or check how the other driver container images handle this.

packages/kmod-5.10-nvidia/kmod-5.10-nvidia.spec
arnaldo2792: These libraries are part of the `graphics_libs_glvnd` list in libnvidia-container:

https://github.com/NVIDIA/libnvidia-container/blob/d40877e0f914ac782c8b6b4f7507e4c84ca04733/src/nvc_info.c#L111

And they are included when the `GLVND` flag is set:

https://github.com/NVIDIA/libnvidia-container/blob/d40877e0f914ac782c8b6b4f7507e4c84ca04733/src/nvc_info.c#L385

arnaldo2792:
> If we could get the output from libnvidia-container when it's checking the compat libraries for inclusion, that might help determine whether either or both of them is OK. Or check how the other driver container images handle this.

`libnvidia-container` will complain when a library is missing, with a message like this, visible in the journal:

```
missing <compat32> library: <library>
```

With these changes, I didn't see any complains:

```
bash-5.0# uname -a
Linux ip-192-168-74-162.us-west-2.compute.internal 5.10.93 #1 SMP Wed Jan 26 19:56:51 UTC 2022 x86_64 GNU/Linux
bash-5.0# journalctl | grep missing
Feb 03 23:12:00 ip-192-168-74-162.us-west-2.compute.internal kernel: nvidia: module verification failed: signature and/or required key missing - tainting kernel
```

```
bash-5.0# uname -a
Linux ip-192-168-78-160.us-west-2.compute.internal 5.10.93 #1 SMP Wed Jan 26 19:54:26 UTC 2022 aarch64 GNU/Linux
bash-5.0# journalctl | grep missing
Feb 03 23:12:20 ip-192-168-78-160.us-west-2.compute.internal kernel: nvidia: module verification failed: signature and/or required key missing - tainting kernel
```

bcressey (approved):

PR #1949, opened 2022-02-09
netdog: add attribute to ignore compiler dead-code warning

etungsten:
**Issue number:**
N/A

**Description of changes:**
Compiling netdog with rust v1.58.1 gives a compiler warning about dead_code in netdog.

```
Date:   Wed Feb 9 15:11:25 2022 -0800

    netdog: add attribute to ignore compiler dead-code warning
    
    We want to store fields from a DHCP lease, even if we don't necessary
    need to use all the fields right now.

```

**Testing done:**
Before the change, when we compile:
```
warning: field is never read: `dns_domain`
  --> api/netdog/src/main.rs:57:5
   |
57 |     dns_domain: Option<String>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `netdog` (bin "netdog") generated 1 warning
```
After the change:
No more compiler warnings in netdog.

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/main.rs
+++ b/sources/api/netdog/src/main.rs
@@ -48,6 +48,7 @@ lazy_static! {
 
 /// Stores fields extracted from a DHCP lease.
 #[derive(Debug, Deserialize)]
+#[allow(dead_code)]
 struct LeaseInfo {
     #[serde(rename = "ipaddr")]
     ip_address: IpNet,


webern (approved):

zmrow (approved):
I think we probably still want to keep this field - lgtm!

:safety_vest:

PR #1953, opened 2022-02-14
kernel: TCMU support

cvlc:
**Issue number:**
N/A (See [AmazonLinux](https://github.com/amazonlinux/amazon-linux-2022/issues/88) issue)

**Description of changes:**
Added the TCMU (LinuxIO user subsystem) feature support as a module to the Bottlerocket kernel. TCMU is a feature of the Linux kernel that exposes block devices (eg. virtual/physical disks) to user space via LinuxIO (LIO) - for storage fabric solutions, eg. Ceph iSCSI Gateway and Ondat. This enables those solutions to run on Bottlerocket.

**Testing done:**
Standard `cargo` build.

**Terms of contribution:**


bcressey:
Looks reasonable to me. Just to confirm - the userspace solutions you have in mind would all be deployed via containers?

Is there a getting started guide focused on Kubernetes deployments for either of the solutions you mentioned? Ideally we'd add some smoke testing to our test clusters for this functionality.

cvlc:
Whoops, apologies for the close-reopen! 

Yes, indeed, it is aimed at containers:

There are a few options for testing
- https://docs.ondat.io/docs/install/ is the install guide for Ondat using a `kubectl` plugin
- https://github.com/storageos/init is an init container that runs some basic checks for TCMU module availability
- https://docs.ceph.com/en/latest/rbd/iscsi-overview/ covers the Ceph feature, though it's a little more involved, I think that Red Hat has an operator and that Rook supports this configuration.

arnaldo2792:
I checked how the init container works, and it will attempt to load various kernel modules. I think the simple smoke test will be to attempt to do the same (load the kernel modules), and a more complete test could be to attempt to run the [ondat-operator](https://github.com/ondat/charts/tree/main/charts/ondat-operator) in Bottlerocket, which under the hood uses the referenced init container. Anyhow, looks good! Thanks @cvlc!

arnaldo2792 (approved):

zmrow (approved):
Thanks! :tada:

bcressey (approved):

PR #1955, opened 2022-02-15
models, containerd, ecs-agent, host-ctr: support registry credentials

### LINKED ISSUE #1227: Configure image registry credentials in containerd
**What I'd like:**

A way to configure image registry credentials in containerd, for pulling images from private registries (or in the case of Dockerhub, with authentication to avoid rate limiting).

This is a common alternative to defining image pull secrets on a per pod basis, and is [recommended by the Kubernetes docs](https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry) if you want particular credentials to apply to all pods on a node.

Having this exposed as a setting would be ideal. Containerd seems to support configuration of registry authentication via toml file as per [this pr](https://github.com/containerd/cri/issues/835). As a first pass, could we expose a setting that passes through to this section of the containerd configuration?

1. Image pull secrets per pod. Inherently difficult and fiddly without automation on the side of creating manifests.
2. A daemonset that mounts host paths to create the required configuration file. Not sure if this will work, it seems as if we might need to restart containerd.
3. Adding configuration via the admin container. Seems like it should be possible, but would be difficult to orchestrate.
4. Pre-pulling images. If the image is already present on the node and the `ImagePullPolicy` is set correctly, the node will just use the pre-pulled image. We could likely orchestrate pulling all essential images shortly after startup, but again, it would be difficult to do.

etungsten:
**Issue number:**
Resolves https://github.com/bottlerocket-os/bottlerocket/issues/1227

**Description of changes:**
```
Date:   Mon Feb 14 17:03:36 2022 -0800

    host-ctr: configure resolver's registry auth based on registry config
    
    host-ctr will set up the custom resolver's authorizer based on the
    registry credentials information stored in host-ctr.toml.
```
```
Date:   Mon Feb 14 16:20:38 2022 -0800

    host-ctr: move registryHosts to the registry pkg
```
```
Date:   Mon Feb 14 17:26:10 2022 -0800

    models, containerd, ecs-agent: add 'container-registry.credentials'
    
    This adds support for configuring registry auth credentials when pulling
    images. The setting is used to render containerd, ecs-agent
    configuration to enable authenticated pulls with the configured auth
    information.
```

**Testing done:**
I created a private docker image registry, "docker.io/etungsten/nginx", that's just a clone of the nginx image repository.

Registry credentials is provided via apiclient in the control container like so:
```
apiclient set --json '{"container-registry": {
      "credentials": [
        {
          "registry": "docker.io",
          "auth": "myauthstring"
        }
      ]
    }}'
```

**For ECS variants:**
The affected-services all restart as expected.

The rendered ECS configuration file:
```
bash-5.0# cat /etc/ecs/ecs.config
ECS_LOGFILE=/var/log/ecs/ecs-agent.log
ECS_LOGLEVEL="info"
ECS_ENGINE_AUTH_TYPE=dockercfg
ECS_ENGINE_AUTH_DATA='{"https://index.docker.io/v1/":{"email": ".","auth": "myauthstring"}}'
```

Ran an ECS tasks that runs the nginx image from my private registry. The task runs successfully.
Checking ECS agent logs. The image gets pulled successfully using the creds I provided
```
Feb 15 02:42:43 amazon-ecs-agent[3412]: level=info time=2022-02-15T02:42:43Z msg="Task engine [arn:aws:ecs:us-west-2:blep:task/566afd42-3dac-44d4-89cd-23ec62c07baa]: finished pulling image etungsten/nginx:latest for container test-ecs in 5.15496234s" module=docker_task_engine.go
Feb 15 02:42:43 amazon-ecs-agent[3412]: level=info time=2022-02-15T02:42:43Z msg="Handling container change event" taskARN="arn:aws:ecs:us-west-2:blep:task/566afd42-3dac-44d4-89cd-23ec62c07baa" container="test-ecs" runtimeID="" status="PULLED"
...
[ec2-user@ip-172-31-27-24 ~]$ curl localhost:80
<!DOCTYPE html>
<html>
<head>
...
```

**For K8s variants:**
The affected-services all restart as expected.

The rendered containerd configuration:
```
...
[plugins."io.containerd.grpc.v1.cri".registry.configs."registry-1.docker.io".auth]
auth = "myauthstring"

```
I'm able to run a K8s pod that has a nginx container that pulls from my private registry, the manifest:
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - name: nginx
    image: docker.io/etungsten/nginx:latest
    imagePullPolicy: Always
```

containerd logs shows it gets pulled successfully:
```
Feb 15 03:44:43 containerd[24471]: time="2022-02-15T03:44:43.020414307Z" level=info msg="PullImage \"docker.io/etungsten/nginx:latest\" returns image reference \"sha256:c316d5a335a5cf324b0dc83b3da82d7608724769f6454f6d9a621f3ec2534a5a\""
Feb 15 03:44:43 containerd[24471]: time="2022-02-15T03:44:43.022183856Z" level=info msg="CreateContainer within sandbox \"2f690146d2f24782bfcd86a18ccea5e385a103f922d6983a765bba2bbf72fcc4\" for container &ContainerMetadata{Name:nginx,Attempt:0,}"

```

The pod runs successfully:
```
$ kubectl get pods -A
NAMESPACE     NAME                       READY   STATUS    RESTARTS   AGE
default       nginx                      1/1     Running   0          3m40s
...
```

**For host-containers:**
The rendered host-ctr configuration:
```
bash-5.0# cat /etc/host-containers/host-ctr.toml
[creds."registry-1.docker.io"]
auth = "myauthstring"

```

I specified an nginx host-container via user-data:
```
[settings.host-containers.nginx]
source = "docker.io/etungsten/nginx:latest"
enabled = true
```

After configuring my registry credentials via apiclient, the host-container is able to pull the image and start:
```
bash-5.0# journalctl --no-hostname -u host-containers@nginx
Feb 15 02:58:34 systemd[1]: Started Host container: nginx.
Feb 15 02:58:36 host-ctr[3850]: time="2022-02-15T02:58:36Z" level=info msg="pulled image successfully" img="docker.io/etungsten/nginx:latest"
Feb 15 02:58:36 host-ctr[3850]: time="2022-02-15T02:58:36Z" level=info msg="unpacking image..." img="docker.io/etungsten/nginx:latest"
...
```

To test selinux changes, I tried accessing the container runtime configuration files from an unprivileged container and it fails as expected:
```
root@ubuntu:/test-rc# ls -lZ
total 68
....
drwxrwxrwt. 3 root root system_u:object_r:any_t:s0      60 Mar  1 01:15 cni
drwxrwxrwt. 2 root root system_u:object_r:secret_t:s0   60 Mar  1 01:15 containerd
...
drwxr-x---. 2 root root system_u:object_r:secret_t:s0  120 Mar  1 01:15 host-containers
....
drwxr-xr-x. 3 root root system_u:object_r:etc_t:s0     140 Mar  1 01:15 wicked
root@ubuntu:/test-rc# cd containerd/
bash: cd: containerd/: Permission denied
root@ubuntu:/test-rc# ls containerd/
ls: cannot open directory 'containerd/': Permission denied
```

Checked `dmesg` on the host and I see AVC denials as expected:
```
[  365.061339] audit: type=1400 audit(1646097682.561:5): avc:  denied  { search } for  pid=5436 comm="bash" name="/" dev="tmpfs" ino=1 scontext=system_u:system_r:container_t:s0:c327,c498 tcontext=system_u:object_r:secret_t:s0 tclass=dir permissive=0
[  365.078879] audit: type=1400 audit(1646097682.561:6): avc:  denied  { search } for  pid=5436 comm="bash" name="/" dev="tmpfs" ino=1 scontext=system_u:system_r:container_t:s0:c327,c498 tcontext=system_u:object_r:secret_t:s0 tclass=dir permissive=0
[  367.830809] audit: type=1400 audit(1646097685.329:7): avc:  denied  { read } for  pid=5573 comm="ls" name="/" dev="tmpfs" ino=1 scontext=system_u:system_r:container_t:s0:c327,c498 tcontext=system_u:object_r:secret_t:s0 tclass=dir permissive=0
```

Tested migration and they worked as expected. On downgrade the new credential settings gets deleted as expected and the OS boots up fine.

**Terms of contribution:**


stevehipwell:
@etungsten how much more work do you think is required to get this production ready? I'm just about to look at the steps to make Bottlerocket a production ready option for our clusters and #1227 is the final real hurdle.

Our current pattern for AL2 is to persist a read only set of Docker credentials to _/var/lib/kubelet/config.json_ as part of the bootstrap, these credentials are rotated in the LT which will be applied to new instances. Our nodes are usually replaced within 2 weeks and always updated within 4 weeks. We'd like to follow the same pattern with Bottlerocket including the full node replacement (we're not going to use the built in OS update).

etungsten:
Hi @stevehipwell , thanks for reaching out. Unfortunately I can't give a firm date on when this feature will be in a formal release. There are still discussions to be had on these changes. Feel free to subscribe to notifications as I update this PR. I'll also try to be as detailed as I can in the PR description to give a sense of what the progress is. In general, once this PR gets merged, it would most likely be in the next minor release after that.

etungsten:
Push above sets up mount points for directories for where we write out configuration files that might contain secrets so we can set up selinux labels for them

sources/host-ctr/cmd/host-ctr/registry.go
```diff
@@ -1,18 +1,37 @@
...
+	runtime "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
 )
 
 // Mirror contains the config related to the registry mirror
 type Mirror struct {
-	Endpoints []string
+	Endpoints []string `toml:"endpoints,omitempty"`
+}
+
+// Credential contain registry credential
```
webern:
```suggestion
// Credential contains a registry credential
```

sources/models/shared-defaults/docker-services.toml
```diff
@@ -6,6 +6,6 @@ configuration-files = ["docker-daemon-config", "proxy-env"]
 path = "/etc/docker/daemon.json"
 template-path = "/usr/share/templates/docker-daemon-json"
 
-# Image registries
-[metadata.settings.container-registry]
+# Image registry mirrors
+[metadata.settings.container-registry.mirrors]
```
webern, line 15: Is there a model change? Or are you making the restart condition more specific?

webern (commented):
Is there a migration needed?

sources/models/shared-defaults/docker-services.toml
etungsten: I'm making the restart condition more specific. `docker` is only impacted by `container-registry.mirrors` and not `container-registry.credentials`

etungsten:
Push above fixes the mount unit so that the SElinux labels actually apply to the mount directory.

Tested things and they all still work as expected.

Also tried accessing the container runtime configuration files from an unprivileged container and it fails as expected:
```
root@ubuntu:/test-rc# ls -lZ
total 68
....
drwxrwxrwt. 3 root root system_u:object_r:any_t:s0      60 Mar  1 01:15 cni
drwxrwxrwt. 2 root root system_u:object_r:secret_t:s0   60 Mar  1 01:15 containerd
...
drwxr-x---. 2 root root system_u:object_r:secret_t:s0  120 Mar  1 01:15 host-containers
....
drwxr-xr-x. 3 root root system_u:object_r:etc_t:s0     140 Mar  1 01:15 wicked
root@ubuntu:/test-rc# cd containerd/
bash: cd: containerd/: Permission denied
root@ubuntu:/test-rc# ls containerd/
ls: cannot open directory 'containerd/': Permission denied
```

Checked `dmesg` on the host and I see AVC denials as expected:
```
[  365.061339] audit: type=1400 audit(1646097682.561:5): avc:  denied  { search } for  pid=5436 comm="bash" name="/" dev="tmpfs" ino=1 scontext=system_u:system_r:container_t:s0:c327,c498 tcontext=system_u:object_r:secret_t:s0 tclass=dir permissive=0
[  365.078879] audit: type=1400 audit(1646097682.561:6): avc:  denied  { search } for  pid=5436 comm="bash" name="/" dev="tmpfs" ino=1 scontext=system_u:system_r:container_t:s0:c327,c498 tcontext=system_u:object_r:secret_t:s0 tclass=dir permissive=0
[  367.830809] audit: type=1400 audit(1646097685.329:7): avc:  denied  { read } for  pid=5573 comm="ls" name="/" dev="tmpfs" ino=1 scontext=system_u:system_r:container_t:s0:c327,c498 tcontext=system_u:object_r:secret_t:s0 tclass=dir permissive=0
```

etungsten:
Push above adds settings descriptions to the README and the migrations for the new setting.

etungsten:
Push above fixes a typo caught by @webern

etungsten:
Push above just updates a link to be more accurate in the README

etungsten:
Push above fixes the key name for `identitytoken` in `hostr-ctr.toml` config file.

Release.toml
```diff
@@ -100,3 +100,7 @@ version = "1.6.0"
     "migrate_v1.6.0_public-admin-container-v0-7-4.lz4",
     "migrate_v1.6.0_public-control-container-v0-5-5.lz4",
 ]
+"(1.6.0, 1.7.0)" = [
```
webern:
```suggestion
"(1.6.1, 1.7.0)" = [
```

webern (approved):

packages/containerd/containerd-config-toml_k8s
```diff
@@ -37,3 +37,25 @@ conf_dir = "/etc/cni/net.d"
...
 {{/each}}
 {{/if}}
+
+{{#if settings.container-registry.credentials}}
+{{#each settings.container-registry.credentials}}
+{{#if (eq registry "docker.io" )}}
+[plugins."io.containerd.grpc.v1.cri".registry.configs."registry-1.docker.io".auth]
+{{else}}
+[plugins."io.containerd.grpc.v1.cri".registry.configs."{{registry}}".auth]
+{{/if}}
```
bcressey, line 47: There's a potential edge case if someone specifies `docker.io` and `registry-1.docker.io`, where we'd end up with two entries. Not sure if that's worth handling though, especially if it doesn't cause an error in the client. I'd be OK with undefined or non-deterministic behavior.

packages/host-ctr/etc-host\x2dcontainers.mount
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Host containers Configuration Directory (/etc/host-containers)
+DefaultDependencies=no
```
bcressey: nit: an alternative to the funky file name required by systemd would be what we do in `release.spec`. File would be named "etc-host-containers.mount.in" and we'd do this in the spec file:
```
MOUNT=$(systemd-escape --path /etc/host-containers)
install -p -m 0644 ${S:100} %{buildroot}%{_cross_unitdir}/${MOUNT}.mount
```

packages/host-ctr/host-ctr.spec
```diff
@@ -29,20 +34,21 @@ install -d %{buildroot}%{_cross_bindir}
...
 install -p -m 0644 %{S:12} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/host-containerd/config.toml
 
-%cross_scan_attribution go-vendor vendor
+%cross_scan_attribution --clarify %{S:1000} go-vendor vendor
 
 %files
 %{_cross_attribution_vendor_dir}
 %{_cross_bindir}/host-ctr
 %{_cross_unitdir}/host-containerd.service
+"%{_cross_unitdir}/etc-host\x2dcontainers.mount"
```
bcressey: nit: since quoting these entries is so unusual:
```suggestion
%{_cross_unitdir}/*.mount
```

sources/models/shared-defaults/docker-services.toml
bcressey, line 15: To check my understanding:
* existing ecs variants will continue to have "affected-services" metadata for the parent setting ("container-registry")
* however, it won't be used after an upgrade because we'll add metadata to the child settings ("mirrors", "credentials"
* on downgrade, the parent setting will be used again, because we never removed it

That sounds fine. The wrinkle I see is if we launch from a newer version that never had metadata for the parent setting, and then downgrade to an older version, we'll be left without "affected-services" for changes to mirrors.

It might be better to leave these defaults unchanged, with a comment that we're keeping them for backwards compatibility, and just add the more specific ones. That would also give us "affected-services" for credentials for the `*-dev` variants using Docker, which don't seem to have them right now.

(A related question is whether we should support non-ECS Docker use of these configured credentials for `*-dev` variants, by writing `/root/.docker/config.json` or something. I'd be happy to backlog that though.)

```suggestion
# Image registries. Retained for backwards compatibility, but superseded by the
# more specific metadata for mirrors and credentials.
affected-services = ["docker", "host-containers", "bootstrap-containers"]

# Image registry mirrors
[metadata.settings.container-registry.mirrors]
affected-services = ["docker", "host-containers", "bootstrap-containers"]

# Image registry credentials
[metadata.settings.container-registry.mirrors]
affected-services = ["host-containers", "bootstrap-containers"]
```

sources/models/src/aws-ecs-1/defaults.d/51-aws-ecs-1.toml
```diff
@@ -22,3 +22,11 @@ service-checks = ["apiserver", "chronyd", "containerd", "host-containerd", "dock
 # Network
 [metadata.settings.network]
 affected-services = ["containerd", "docker", "ecs", "host-containerd", "host-containers"]
```
bcressey: I can't comment on the `51-docker-services.toml` rename, but I'd prefer to keep that as `51-docker-services.toml`, since that's the link name that other models use.

sources/api/migration/migrations/v1.7.0/container-registry-credentials-metadata/src/main.rs
```diff
@@ -0,0 +1,24 @@
...
+use std::process;
+
+/// We added a new setting and `affected-services` metadata for `container-registry.credentials`
+/// This is only for the ECS variant where we only want to restart the ECS agent and not the docker daemon when credentials settings change.
+fn run() -> Result<()> {
+    migrate(AddMetadataMigration(&[SettingMetadata {
+        metadata: &["affected-services"],
+        setting: "settings.container-registry.credentials",
+    }]))
+}
```
bcressey: Aren't we also adding `affected-services` metadata for `container-registry.mirrors`? Maybe that gets swept up by existing migrations. You could try downgrading below `1.2.0` to make sure nothing strange happens.

packages/ecs-agent/ecs.config
```diff
@@ -1,2 +1,18 @@
...
+    "https://index.docker.io/v1/":
+    {{~else~}}
+    "{{registry}}":
+    {{~/if~}}
+    {"email": "."
+        {{~#if auth~}},"auth": "{{{auth}}}"{{/if}}
+        {{~#if username~}},"username": "{{{username}}}"{{/if}}
+        {{~#if password~}},"password": "{{{password}}}"}{{/if}}
+    {{~/each~}}}}
+{{/if}}
```
bcressey: Can we add single-quotes around this so it looks more like a shell variable, or does the ecs agent not like that?

sources/host-ctr/cmd/host-ctr/registry.go
```diff
@@ -25,3 +44,94 @@ func NewRegistryConfig(registryConfigFile string) (*RegistryConfig, error) {
...
+			authConfig.Auth = registryConfig.Credentials[defaultHost].Auth
+			authConfig.IdentityToken = registryConfig.Credentials[defaultHost].IdentityToken
+			authOpts = append(authOpts, docker.WithAuthClient(&http.Client{
+				Transport: newTransport(),
+			}))
+			authOpts = append(authOpts, docker.WithAuthCreds(func(host string) (string, string, error) {
+				return server.ParseAuth(&authConfig, host)
+			}))
+		}
+		authorizer = docker.NewDockerAuthorizer(authOpts...)
```
bcressey: Is it OK to share an instance of `authorizer` like this?  The CRI implementation builds one per registry endpoint.

I assume so because the previous implementation did that when passed an authorizer override, although I'm not sure we'd ever end up with more than one in the case where it was used (ecr-public).

It's a bit awkward to do the work to construct the authorizer and then just throw it away if we were given an override.

etungsten:
Push above rebases onto latest develop.

packages/containerd/containerd-config-toml_k8s
etungsten: Good call out. I tested this and containerd does complain about duplicate entries and then refuse to start:
```
Mar 04 00:27:32 containerd[124031]: containerd: failed to load TOML: /etc/containerd/config.toml: (37, 2): duplicated tables
Mar 04 00:27:32 systemd[1]: containerd.service: Main process exited, code=exited, status=1/FAILURE
Mar 04 00:27:32 systemd[1]: containerd.service: Failed with result 'exit-code'.
Mar 04 00:27:32 systemd[1]: Failed to start containerd container runtime.
```
But I kinda see setting both `docker.io` and `registry-1.docker.io` akin to setting the same registry twice which would also make containerd fail.

We do not enforce uniqueness of the registries in the `container-registry.credentials` settings model. We would essentially need to wrap `Vec<RegistryCredential>` into its own settings model type that would special case `docker.io` and enforce uniqueness of the `registry` field. Lemme know what you think!

etungsten: Alternatively, I can remove this conditional and we can just make users specify `registry-1.docker.io` when they want to set up credentials for dockerhub. And for ECS variants, `https://index.docker.io/v1/`.

sources/api/migration/migrations/v1.7.0/container-registry-credentials-metadata/src/main.rs
etungsten: Nice catch, I'll add that.

The existing migrations do not wipe away the metadata for the parent setting. Back when `container-registry` was first introduced, we didn't think it was necessary to delete new settings metadata on downgrade since the settings wouldn't exist in the downgraded version. We've added a new `AddMetadataMigration` helper since then to help with that.

etungsten:
[This push](https://github.com/bottlerocket-os/bottlerocket/compare/4e7ee3511b1358d779ec79440cad20dfd33c6813..b3659def70693f94554297b28999512189ef21d4) addresses majority of @bcressey 's comments.

The push above fixes formatting.

Tested things and they still work as described in the PR.

etungsten:
Push above moves the migrations to `"(1.6.1, 1.6.2)" `  now that we're planning this for `v1.6.2`

bcressey (approved):

README.md
```diff
@@ -446,6 +446,30 @@ When pulling an image from a registry, the container runtime will try the endpoi
...
+        },
+        {
+          "registry": "docker.io",
+          "auth": "example_base64_encoded_auth_string"
+        }
+      ]
+    }
+  }'
+  ```
+In additional to the container runtime daemons, these credential settings will also apply to [host-container](#host-containers-settings) and [bootstrap-container](#bootstrap-containers-settings) image pulls as well.
```
arnaldo2792: This might be a typo? `In additional`, or `In addition`?

packages/containerd/containerd-config-toml_k8s
bcressey: I'm good with your current approach, after further thought and based on our discussion out of band.

Without the conditional, on `aws-ecs-1` to override registry credentials for Docker Hub, you'd need to specify both so that `host-ctr` and `ecs-agent` both get the form they expect (and ignore the other form). That's awkward and confusing enough that it's worth the extra bit implementation effort to paper over it.

arnaldo2792 (approved):
The commit message that starts with this:

```
ecs-agent, host-ctr, containerd: add mount units for configuration di
```

Seems truncated

etungsten:
> The commit message that starts with this:
> 
> ```
> ecs-agent, host-ctr, containerd: add mount units for configuration di
> ```
> 
> Seems truncated

The title is truncated because it was too long, but the message is still there.

etungsten:
Push above fixes typo found by @arnaldo2792

etungsten:
Tested migration again and it still works as expected

etungsten:
Push above rebases to fix conflicts in `Release.toml` and in `sources/Cargo.toml`

arnaldo2792 (approved):

PR #1983, opened 2022-03-04
update systemd to 250.4

### LINKED ISSUE #1671: Update systemd
In the [last package update](https://github.com/bottlerocket-os/bottlerocket/pull/1619) I was unable to update systemd beyond 247.x because 248.2+ have [this change](https://github.com/systemd/systemd/commit/c2c331056a7c331a5478124b3cd6a34c9f539839) which breaks dbus-broker startup unless you disable private mount namespaces for the service unit.  systemd 249.2 still has this change, so we'll have to figure out how we want to update; 247.x won't be updated forever.

### LINKED ISSUE #1815: Update libseccomp to 2.53+
There's a conflict with systemd we'll have to figure out; this is what happens when trying to build systemd 247.10 after updating to libseccomp 2.53:

```
  #18 13.63 In file included from /x86_64-bottlerocket-linux-gnu/sys-root/usr/include/seccomp.h:821,                                                                        
  #18 13.63                  from ../src/shared/seccomp-util.c:6:                                                                                                           
  #18 13.63 ../src/shared/seccomp-util.c: In function 'seccomp_restrict_sxid':                                                                                              
  #18 13.63 ../src/shared/seccomp-util.c:2077:25: error: '__NR_openat2' undeclared (first use in this function); did you mean '__openat_2'?                                 
  #18 13.63  2077 |                         SCMP_SYS(openat2),                                                                                                              
  #18 13.63       |                         ^~~~~~~~
```

bcressey:
**Issue number:**
Fixes #1671, fixes #1815

**Description of changes:**
Update to the most recent stable release of systemd. Rebase patches to account for upstream changes since 247, and specify the expected set of build options.

Also update libseccomp, since that was held back by the older version of systemd.

**Testing done:**

Note that I didn't see the issue with dbus-broker startup that was observed in #1671, but I haven't bisected to find the actual fix.

Stable machine IDs confirmed on:
- [x] EC2 / Nitro
- [x] EC2 / Xen
- [x] VMware

Workload tests passed:
- [x] aws-k8s-1.21 / x86_64
- [x] aws-ecs-1 / aarch64
- [x] vmware-k8s-1.21 / x86_64

**Terms of contribution:**


zmrow (approved):
:dancers:

cbgbt (approved):

packages/systemd/systemd.spec
```diff
@@ -118,11 +118,13 @@ CONFIGURE_OPTS=(
  -Drepart=true
  -Dcoredump=false
  -Dpstore=true
+ -Doomd=false
```
arnaldo2792, line 121: [systemd-ood](https://www.man7.org/linux/man-pages/man8/systemd-oomd.service.8.html) seems like a nice features I wonder if we could use it to set some limits in the units we maintain when we move to cgroups v2 :thinking:

packages/systemd/systemd.spec
```diff
@@ -242,6 +268,7 @@ install -p -m 0644 %{S:4} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/i
 %{_cross_bindir}/systemd-cat
 %{_cross_bindir}/systemd-cgls
 %{_cross_bindir}/systemd-cgtop
+%{_cross_bindir}/systemd-creds
```
arnaldo2792, line 271: Are we planning to [pass credentials](https://man7.org/linux/man-pages//man1/systemd-creds.1.html) to systemd units? Otherwise, should this be excluded with `%exclude` or installed in other subpackage?

packages/systemd/systemd.spec
bcressey: At one point systemd-oomd required swap, which we don't support today. It's also not obvious how useful it is with kubelet in the picture.

packages/systemd/systemd.spec
bcressey: I'm interested in this as a potential future direction, and wanted it enabled to facilitate that. Similar story to `systemd-repart` which we ended up using.

arnaldo2792 (approved):

bcressey:
systemd-stable 250.4 is out, so I'm moving this back to draft while I review the changes there.

bcressey:
(rebase only)

bcressey:
(switch to systemd 250.4, add libseccomp 2.5.3 update)

arnaldo2792 (approved):
:tada:

etungsten (approved):

jpculp (approved):

Issue #2015, opened 2022-03-22
update webpki


webern:
**What I'd like:**

It looks like our `webpki-roots-shim` is incompatible with the next major version of webpki roots (0.22):

```
$ cargo build
   Compiling webpki-roots v0.22.2 (/home/mjb/repos/bottlerocket/sources/webpki-roots-shim)
error[E0432]: unresolved import `webpki::trust_anchor_util`
  --> webpki-roots-shim/src/lib.rs:10:13
   |
10 | use webpki::trust_anchor_util::cert_der_as_trust_anchor;
   |             ^^^^^^^^^^^^^^^^^ could not find `trust_anchor_util` in `webpki`

error[E0423]: expected function, tuple struct or tuple variant, found type alias `TLSServerTrustAnchors`
  --> webpki-roots-shim/src/lib.rs:17:9
   |
17 |         TLSServerTrustAnchors(&ROOTS)
   |         ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: can't use a type alias as a constructor

warning: use of deprecated type alias `webpki::TLSServerTrustAnchors`: use TlsServerTrustAnchors
  --> webpki-roots-shim/src/lib.rs:11:14
   |
11 | use webpki::{TLSServerTrustAnchors, TrustAnchor};
   |              ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default

warning: use of deprecated type alias `webpki::TLSServerTrustAnchors`: use TlsServerTrustAnchors
  --> webpki-roots-shim/src/lib.rs:14:38
   |
14 |     pub static ref TLS_SERVER_ROOTS: TLSServerTrustAnchors<'static> = {
   |                                      ^^^^^^^^^^^^^^^^^^^^^

Some errors have detailed explanations: E0423, E0432.
For more information about an error, try `rustc --explain E0423`.
warning: `webpki-roots` (lib) generated 2 warnings
error: could not compile `webpki-roots` due to 2 previous errors; 2 warnings emitted
FAIL: 101
```

**Any alternatives you've considered:**

PR #2049, opened 2022-03-31
Dependabot updates to rust dependencies 

jpculp:
**Description of changes:**

This is a collection of rust dependency updates brought on by dependabot.

**Testing done:**

- Built and launched aws-k8s-1.21 instance.
- Verified the instance connected to the cluster.
- Enabled and interacted with the admin container via `apiclient exec`.

**Terms of contribution:**

### DIFF
--- a/sources/webpki-roots-shim/src/lib.rs
+++ b/sources/webpki-roots-shim/src/lib.rs
@@ -29,7 +29,13 @@ lazy_static::lazy_static! {
 fn tls_server_roots_pem() -> Vec<Pem> {
     match std::fs::read(&*CERT_PATH) {
         Ok(data) => {
-            let mut v = pem::parse_many(&data);
+            let mut v = match pem::parse_many(&data) {
+                Ok(v) => v,
+                Err(err) => {
+                    error!("failed to parse {}: {}", CERT_PATH.display(), err);
+                    Vec::new()
+                }
+            };
             v.shrink_to_fit();
             v
         }
--- a/tools/infrasys/src/main.rs
+++ b/tools/infrasys/src/main.rs
@@ -15,7 +15,7 @@ use std::collections::HashMap;
 use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::{fs, process};
-use structopt::StructOpt;
+use structopt::{clap, StructOpt};
 use tokio::runtime::Runtime;
 use url::Url;
 
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -2374,17 +2374,6 @@ dependencies = [
 
 [[package]]
 name = "pem"
-version = "0.8.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fd56cbd21fea48d0c440b41cd69c589faacade08c992d9a54e471b79d0fd13eb"
-dependencies = [
- "base64",
- "once_cell",
- "regex",
-]
-
-[[package]]
-name = "pem"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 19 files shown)


sources/webpki-roots-shim/src/lib.rs
```diff
@@ -29,7 +29,13 @@ lazy_static::lazy_static! {
 fn tls_server_roots_pem() -> Vec<Pem> {
     match std::fs::read(&*CERT_PATH) {
         Ok(data) => {
-            let mut v = pem::parse_many(&data);
+            let mut v = match pem::parse_many(&data) {
+                Ok(v) => v,
+                Err(err) => {
+                    error!("failed to parse {}: {}", CERT_PATH.display(), err);
+                    Vec::new()
```
cbgbt, line 36: We probably should modify this function to return a `Result` if we can, I'd prefer it fail rather than disregard invalid certs.

tools/infrasys/src/main.rs
```diff
@@ -15,7 +15,7 @@ use std::collections::HashMap;
 use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::{fs, process};
-use structopt::StructOpt;
```
cbgbt, line 18: Does it not build if we don't `use clap`?

We should probably remove `structopt` in favor of just `clap` going forward, though it could be done as a separate task. I haven't investigated what a migration looks like, but the feature set we're using should have something synonymous in `clap`.

Check out the structop docs here: https://docs.rs/structopt/latest/structopt/#maintenance

jpculp: From what I can tell we would want to switch to `clap` sooner rather than later. Because of the version bump, our tools don't build unless we prefix every `clap` with `structopt::`, or we add it to the use declaration.

jpculp:
Rebased on develop.

sources/webpki-roots-shim/src/lib.rs
jpculp: At the moment we're already disregarding invalid certs further down in the `tls_server_roots_pem` function, so I'm a bit reluctant to change behavior in a dependency update commit. Should I open up an issue to re-evaluate how this handles errors?

jpculp:
Bump `actix-web` and `actix-web-actors`.

Note: `actix-web-actors` 4.0.0 also works. 4.1.0 adds support for `actix` 0.13 which I don't think we're using yet.

webern:
Looks like this closes #2015?

jpculp:
- Bump `clap` to 3.1.8.
- Drop `actix-web-actors` to 4.0.0.

jpculp:
Fixed version in commit message.

sources/webpki-roots-shim/src/lib.rs
webern: > reluctant to change behavior in a dependency update commit

I think I agree with your assessment. Previously `parse_many` was swallowing errors, so to remain consistent we should swallow errors here. Otherwise, somebody could have a new Bottlerocket error that they didn't have before with an invalid cert.

Previous error swallowing:

https://github.com/jcreekmore/pem-rs/blob/e869f244552321015e0173a3c71dbb45a43cb981/src/lib.rs#L324

tools/infrasys/src/main.rs
webern: > it could be done as a separate task

Yes, I think replacing StructOpt with `clap` should be done separately from regular library updates.

> Does it not build if we don't `use clap`

Seems like this change is fine... it makes it compile!

webern:
> Looks like this closes #2015?

No, ignore me. The webpki version is not changing and neither is the webpki-roots version.

webern (approved):
This looks good. Nothing is changing about webpki-roots. We need to update webpki and make sure that webpki-roots-shim points to the updated version, but we should do that separately from these more routine updates.

cbgbt (approved):

Issue #2052, opened 2022-04-01
Go Modules in packages


samuelkarp:
In order to provide core functionality for the operating system, Bottlerocket includes a number of third-party software packages.  Bottlerocket builds these software packages from source code, and stores the source code in a “lookaside” cache.  The lookaside cache helps ensure that builds can be reproducible and reduces the load from Bottlerocket builds on upstream source repositories.  Bottlerocket’s build system (buildsys) expects sources to be declared in an individual package’s `Cargo.toml` `package.metadata.external-files` key, then (by default) will download the specified source code from the lookaside cache (and can optionally fall-back to the upstream source during package development).  Buildsys then runs each package build in an environment with network access disabled, preventing build scripts from retrieving additional source code not declared in the `Cargo.toml` file and downloaded from the lookaside cache.

A small (but growing) number of the third-party software packages included in Bottlerocket are written in the Go programming language.  These include components critical to running containers such as runc, containerd, Docker (in some variants), and the orchestrator agents (Kubelet and ECS agent).

In the early days of the Go ecosystem, “vendoring” was common.  Vendoring is the practice of a software project copying the source code of its dependencies into its own source tree.  Vendoring became popular in the Go ecosystem as the Go tooling (formerly) provided little other support for managing dependencies; copying was a reliable way to ensure the correct versions of dependencies were available when compiling a project.  In more recent versions of Go, native dependency management with “modules” was added where the Go tool is responsible for retrieving the necessary dependency sources at build time.  Modern Go programs use Go modules and declare their dependencies in `go.mod` and `go.sum` files; some maintain “legacy” vendor directories but the ecosystem is gradually moving away from vendoring.

Most of the Go programs included in Bottlerocket (runc, containerd, Docker, Kubelet, ECS agent) currently employ the vendoring practice upstream.  Consequently, the process to ingest sources for these programs is fairly simple: copy the upstream source tarball vended by the project into Bottlerocket’s lookaside cache and build from there.  However, there are upstream dependencies which are more modern and thus do *not* employ the vendoring process.  In this case we need to solve the question: how to properly retrieve and store those dependencies such that they can be used during a build?

**Option A**
[...]

PR #2053, opened 2022-04-01
buildsys: add feature to vendor go modules

samuelkarp:
**Issue number:**
https://github.com/bottlerocket-os/bottlerocket/issues/2052

**Description of changes:**
This pull request implements Option C in https://github.com/bottlerocket-os/bottlerocket/issues/2052: changes to buildsys that enable automatic Go module downloads ahead of the rpmbuild invocation.

**Testing done:**
```
cargo make -e PACKAGE=oci-add-hooks build-package
cargo make -e PACKAGE=hotdog build-package
```

```
cargo make -e BUILDSYS_VARIANT=aws-ecs-1
```

Launch an `aws-ecs-1` AMI, verify that oci-add-hooks are built properly and have dependency licenses in `/usr/share/licenses`

**Terms of contribution:**


etungsten (approved):
This is awesome! Thanks!!

samuelkarp:
> [samuelkarp](https://github.com/samuelkarp) force-pushed the buildsys-go-modules branch from 65bd339 to a7b439a

Added `.gitignore` files to the packages for the `src/` directory that gets created during build.

packages/hotdog/hotdog.spec
```diff
@@ -10,26 +10,12 @@
...
-%global goselinux 1.10.0
-
-%global libcap 1.2.62
-
 Name: %{_cross_os}hotdog
 Version: 1.0.1
 Release: 1%{?dist}
 Summary: Tool with OCI hooks to run the Log4j Hot Patch in containers
 License: Apache-2.0
 URL: https://github.com/awslabs/oci-add-hooks
```
bcressey, line 18: pre-existing nit: we have the wrong URL here

packages/hotdog/Cargo.toml
```diff
@@ -12,23 +12,10 @@ path = "pkg.rs"
...
-sha512 = "88c2591416aaea346bf438021c5f5095212ab6eec765c4b9ba149bd229cdc92a6baddc11f4aef7cfa7f0a5df6eb453e5670689679fd831848860eb28aca034a2"
-
-[[package.metadata.build-package.external-files]]
-path = "libcap-v1.2.62.tar.gz"
-url = "https://git.kernel.org/pub/scm/libs/libcap/libcap.git/snapshot/libcap-cap/v1.2.62.tar.gz"
-sha512 = "d533563e4a2b4b9f910730286180d160c00e436e7f6970e124a67991ab2dcdeca4a6b07cb230d812cc6b4771935accc49ba0ed0465a581a571f0ddb4752284f7"
+[[package.metadata.build-package.go-mods]]
+input = "hotdog-3f2ca92.tar.gz"
+mod-dir = "hotdog-3f2ca9275fae8db87409c3a0999aa2c8a4bd44d1"
+output-dir = "src"
```
bcressey, line 18: I'd tend to prefer coming up with some standard contract and automating that in a way that's easy for spec files to consume, so there's less risk of each spec doing things in slightly different ways, which tends to obscure packaging mistakes.

Specifically, if instead of an output dir we created a tarball, then we could reference that archive in the spec file in the usual way - listing it as a `Source`, and using the `%setup` or `%autosetup` macro. The latter would be valuable for any patches we'd carry. There's also a modest performance benefit from not having lots of small files to copy into the build context.

Then we could specify the behavior into the `external-files` declaration like:
```
url = "https://github.com/bottlerocket-os/hotdog/archive/3f2ca9275fae8db87409c3a0999aa2c8a4bd44d1/hotdog-3f2ca92.tar.gz"
path = "override-upstream.tar.gz"
sha512 = "ebf8f2f8f2c7b12443eae5eccdf0d37724637ac972874b9a88eb8d21d88cbe661b0b53680015ebb8c06e17bde708acf90f39fac82002a4c72266ecdac87c3fb5"
bundle-modules = [ "go" ]
bundle-path = "override-upstream-bundled.tar.gz"
```

(Not entirely sure about what to name the output file but there's probably an automatic convention like `x.tar.*` -> `x-bundled.tar.*` that we could use.)

What do you think?

packages/oci-add-hooks/oci-add-hooks.spec
```diff
@@ -4,48 +4,36 @@
...
-tar -C GOPATH/src/github.com/joeshaw/json-lossless -xzf %{SOURCE2} --strip 1
-cp GOPATH/src/github.com/joeshaw/json-lossless/LICENSE LICENSE.json-lossless
+%setup -T -c
+cp -r /home/builder/src/%{gorepo}-%{gitrev}/* .
 
 %build
-%cross_go_configure %{goimport}
-# We use `GO111MODULE=off` to force golang to look for the dependencies in the GOPATH
-GO111MODULE=off go build -v -x -buildmode=pie -ldflags="${GOLDFLAGS}" -o oci-add-hooks
+%set_cross_go_flags
```
bcressey, line 25: Awesome! It'll be really nice to get rid of `cross_go_setup` and `cross_go_configure` at some point.

tools/buildsys/src/main.rs
```diff
@@ -155,6 +161,10 @@ fn build_package() -> Result<()> {
         }
     }
 
+    if let Some(mods) = manifest.go_mods() {
+        GoMod::vendor(&root_dir, &manifest_dir, &mods).context(error::GoModSnafu)?;
+    }
```
bcressey, line 166: Need to emit `cargo:rerun-if-changed` directives for the output files, or tarball if we switch to that.

tools/buildsys/src/gomod/error.rs
```diff
@@ -0,0 +1,34 @@
...
+    CommandStart { source: std::io::Error },
+
+    #[snafu(display("Failed to execute command: 'docker {}'", args))]
+    DockerExecution { args: String },
+
+    #[snafu(display("input is required"))]
+    InputFile,
+
+    #[snafu(display("input must be a file and end with .tar.gz"))]
+    InputFileBad,
```
bcressey, line 16: nit: we don't seem to check for the "ends with .tar.gz" part (and might not want to, since other compression formats might be used)

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,160 @@
...
+        root_dir: &PathBuf,
+        package_dir: &PathBuf,
+        gomods: &[manifest::GoModule],
+    ) -> Result<Self> {
+        for g in gomods {
+            let input_path_arg = g.input.as_ref().context(error::InputFileSnafu)?;
+            let input_path = package_dir.join(input_path_arg);
+            if !input_path.is_file() {
+                return Err(error::Error::InputFileBad);
+            }
```
bcressey, line 40: I was wondering about the lack of the "Snafu" extension - should use `ensure!` instead. It's also conventional to provide the actual path as context, so it's easier to reason about the error.

```suggestion
            ensure!(input_path.is_file(), error::InputFileBadSnafu { path: input_path });
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,160 @@
...
+            let gosumdb = go_env("GOSUMDB").unwrap_or("".to_string());
+            let goprivate = go_env("GOPRIVATE").unwrap_or("".to_string());
+
+            let args = DockerGoArgs {
+                module_path: package_dir.clone(),
+                sdk_image: sdk,
+                go_mod_cache: root_dir.join(".gomodcache".to_string()),
+                command: format!(
+                    "mkdir -p {outdir}
+                    tar zxf {input} -C {outdir} &&
```
bcressey, line 62: You can just have `tar` figure out what kind of decompression to use:

```suggestion
                    tar xf {input} -C {outdir} &&
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,160 @@
...
+                    moddir = mod_dir.to_string_lossy(),
+                    goproxy = goproxy,
+                    gosumdb = gosumdb,
+                    goprivate = goprivate,
+                ),
+            };
+            match docker_go(root_dir, &args) {
+                Ok(_) => continue,
+                Err(e) => return Err(e),
+            }
```
bcressey, line 79: This can be written more concisely as:
```suggestion
            let _ = docker_go(root_dir, &args)?;
```

(Might be able to get rid of the `let _ =` part also.)

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,160 @@
...
+use std::process::Output;
+
+pub(crate) struct GoMod;
+
+impl GoMod {
+    pub(crate) fn vendor(
+        root_dir: &PathBuf,
+        package_dir: &PathBuf,
+        gomods: &[manifest::GoModule],
+    ) -> Result<Self> {
```
bcressey, line 34: Looks like you can just return nothing in the happy case:
```suggestion
    ) -> Result<()> {
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,160 @@
...
+
+/// Run `docker-go` with the specified arguments.
+fn docker_go(root_dir: &PathBuf, dg_args: &DockerGoArgs) -> Result<Output> {
+    let args = vec![
+        "--module-path",
+        dg_args
+            .module_path
+            .to_str()
+            .context(error::InputFileSnafu)
+            .unwrap(),
```
bcressey, line 101: Hmm why do we still need to `unwrap()` here?

Generally that would mean that we would add another `.context(error::WhateverSnafu)` to the chain.

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    println!("{}", &stdout);
+    return if output.status.success() {
+        Ok(output)
+    } else {
+        Err(error::Error::DockerExecution {
+            args: "".to_string(),
+        })
+    };
```
bcressey, line 130:
```suggestion
    ensure!(output.status.success(), error::DockerExecution { args });
    Ok(output)
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,160 @@
...
+        Ok(v) => v,
+        Err(_) => return None,
+    };
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    println!("{}", &stdout);
+    return if output.status.success() {
+        Some(stdout.to_string())
+    } else {
+        None
+    };
```
bcressey, line 151: Does this work?

```suggestion
    output.status.success().then(|| stdout.to_string())
```

samuelkarp:
@bcressey Thanks for the detailed review here!  I no longer work for Amazon and am not currently in a position to dedicate my own personal time or resources to moving this forward.  I'd be happy for you or for someone else on the team to pick up my contribution and finish it.  Please feel free to close this once a replacement PR is open.

packages/hotdog/hotdog.spec
jpmcb: Updated in https://github.com/bottlerocket-os/bottlerocket/pull/2326 👀

Issue #1788, opened 2021-10-27
Support Warm Pool for EC2 Auto Scaling!


samjo-nyang:
**What I'd like:**
It is good to Bottlerocket support warm pool.

For now, kubelets that in warm pools instances are started so they are registered in the control plane. And then, the instances are stopped so the nodes are falling into "NotReady" state.

Therefore, I think the Bottlerocket instances should (optionally) read their current status, and postpone the kubelet up. (such as https://github.com/kubernetes/kops/pull/11216)
**Any alternatives you've considered:** (none)

arnaldo2792:
Hi @samjo-nyang thanks for sharing this! It is an interesting feature, we will check how to support this down the road!

samjo-nyang:
For your information, it's kind of cheating... but it works.
```diff
 cmd/kubelet/app/server.go | 47 +++++++++++++++++++++++++++++++++++++++
 1 file changed, 47 insertions(+)

diff --git a/cmd/kubelet/app/server.go b/cmd/kubelet/app/server.go
index ff7e05feec1..4ea90d8ec39 100644
--- a/cmd/kubelet/app/server.go
+++ b/cmd/kubelet/app/server.go
@@ -22,6 +22,10 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	awsautoscaling "github.com/aws/aws-sdk-go/service/autoscaling"
 	"math"
 	"net"
 	"net/http"
@@ -474,7 +478,50 @@ func makeEventRecorder(kubeDeps *kubelet.Dependencies, nodeName types.NodeName)
 	}
 }
 
+func isInWarmEC2() bool {
+	sess, err := session.NewSession()
+	if err != nil {
+		return false
+	}
+	svc := ec2metadata.New(sess)
+	if !svc.Available() {
+		return false
+	}
+	doc, err := svc.GetInstanceIdentityDocument()
+	if err != nil {
+		return false
+	}
+	instanceId := doc.InstanceID
+	region := doc.Region
+
+	asgClient := awsautoscaling.New(sess, &aws.Config{Region: aws.String(region)})
+	resp, err := asgClient.DescribeAutoScalingInstances(&awsautoscaling.DescribeAutoScalingInstancesInput{
[...]

PR #2063, opened 2022-04-11
Support for ECS Agent parameters ImagePullBehavior and WarmPoolsSupport

mello7tre:
**Issue number:**
None.
I have looked in the backlog and found no task regarding new ecs-agent parameters, so i directly opened the PR.
Maybe WarmPoolsSupport can be used for #1788 (for the ECS part).

**Description of changes:**
Added support for ECS Agent parameters `ImagePullBehavior` and `WarmPoolsSupport`.
(relative to variables ECS_IMAGE_PULL_BEHAVIOR and ECS_WARM_POOLS_CHECK)

Can be configured using the keys:
`settings.ecs.image-pull-behavior`
`settings.ecs.warm-pools-check`

image-pull-behavior can have the values: default, always, once, prefer-cached.

**Testing done:**
executed `cargo make unit-tests`.
Launched Bottlerocket instance using builded ami with user-data:
```
[settings.ecs]
image-pull-behavior = "once"
warm-pools-check = true
```
verified that the content of file `/etc/ecs/ecs.config.json` included the new settings:
```
{
  "InstanceAttributes": {
    "bottlerocket.variant": "aws-ecs-1"
  },
  "PrivilegedDisabled": true,
  "AvailableLoggingDrivers": [
    "json-file",
    "awslogs",
    "none"
  ],
  "WarmPoolsSupport": true,
  "ImagePullBehavior": 2,
  "TaskIAMRoleEnabled": true,
  "TaskIAMRoleEnabledForNetworkHost": true,
  "SELinuxCapable": true,
  "OverrideAWSLogsExecutionRole": true,
  "TaskENIEnabled": true
}
```
Note: `ImagePullBehavior` is a `iota`.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,13 +432,17 @@ These settings can be changed at any time.
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
 * `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
-
+* `settings.ecs.image-pull-behavior`: The behavior used to customize the [pull image process](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html#ecs-agent-availparam) for your container instances.
+  Supported values are `default`, `always`, `once`, `prefer-cached`, and the default is `default`.
 #### CloudFormation signal helper settings
 
 For AWS variants, these settings allow you to set up CloudFormation signaling to indicate whether Bottlerocket hosts running in EC2 have been successfully created or updated:
 * `settings.cloudformation.should-signal`: Whether to check status and send signal. Defaults to `false`. If set to `true`, both `stack-name` and `logical-resource-id` need to be specified.
 * `settings.cloudformation.stack-name`: Name of the CloudFormation Stack to signal.
 * `settings.cloudformation.logical-resource-id`: The logical ID of the AutoScalingGroup resource that you want to signal.
+
+#### Auto Scaling group settings
+* `settings.autoscaling.should-wait`: Whether to wait for the instance to reach the `InService` state before the orchestrator agent joins the cluster. Defaults to `false`. Set this to `true` only if the instance is part of an Auto Scaling group, or will be attached to one later.
 
 #### OCI Hooks settings
 
--- a/sources/api/ecs-settings-applier/src/ecs.rs
+++ b/sources/api/ecs-settings-applier/src/ecs.rs
@@ -9,9 +9,11 @@ embedded lists.  The structure and names of fields in the document can be found
 */
 use constants;
 use log::debug;
+use model::modeled_types::ECSImagePullBehavior;
 use serde::Serialize;
 use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
 use snafu::{OptionExt, ResultExt};
+use std::convert::TryFrom;
 use std::fs;
 use std::path::Path;
 use std::{env, process};
@@ -36,6 +38,12 @@ struct ECSConfig {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     spot_instance_draining_enabled: Option<bool>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    warm_pools_support: Option<bool>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    image_pull_behavior: Option<u8>,
 
     #[serde(rename = "TaskIAMRoleEnabled")]
     task_iam_role_enabled: bool,
@@ -70,8 +78,8 @@ async fn run() -> Result<()> {
     let args = parse_args(env::args());
     SimpleLogger::init(LevelFilter::Info, LogConfig::default()).context(error::LoggerSnafu)?;
 
-    // Get all settings values for config file templates
-    debug!("Requesting settings values");
+    // Get all ecs and autoscaling settings values for config file templates
+    debug!("Requesting ecs and autoscaling settings values");
     let settings = schnauzer::get_settings(&args.socket_path)
         .await
         .context(error::SettingsSnafu)?;
@@ -79,19 +87,32 @@ async fn run() -> Result<()> {
     debug!("settings = {:#?}", settings.settings);
     let ecs = settings
         .settings
-        .and_then(|s| s.ecs)
+        .as_ref()
+        .and_then(|s| s.ecs.as_ref())
+        .context(error::ModelSnafu)?;
+
+    let autoscaling = settings
+        .settings
+        .as_ref()
+        .and_then(|s| s.autoscaling.as_ref())
         .context(error::ModelSnafu)?;
 
     let mut config = ECSConfig {
-        cluster: ecs.cluster,
+        cluster: ecs.cluster.clone(),
         privileged_disabled: ecs.allow_privileged_containers.map(|s| !s),
         available_logging_drivers: ecs
             .logging_drivers
+            .clone()
             .unwrap_or_default()
             .iter()
             .map(|s| s.to_string())
             .collect(),
         spot_instance_draining_enabled: ecs.enable_spot_instance_draining,
+        warm_pools_support: autoscaling.should_wait,
+        image_pull_behavior: ecs
+            .image_pull_behavior
+            .as_ref()
+            .map(|b| ECSImagePullBehavior::try_from(b.as_ref()).unwrap() as u8),
 
         // Task role support is always enabled
         task_iam_role_enabled: true,
@@ -115,7 +136,7 @@ async fn run() -> Result<()> {
             .instance_attributes
             .insert(VARIANT_ATTRIBUTE_NAME.to_string(), os.variant_id);
     }
-    if let Some(attributes) = ecs.instance_attributes {
+    if let Some(attributes) = &ecs.instance_attributes {
         for (key, value) in attributes {
             config
                 .instance_attributes
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -141,13 +141,13 @@ use std::net::IpAddr;
 use crate::de::{deserialize_mirrors, deserialize_node_taints};
 use crate::modeled_types::{
     BootConfigKey, BootConfigValue, BootstrapContainerMode, CpuManagerPolicy, DNSDomain,
-    ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue, FriendlyVersion, Identifier,
-    KubernetesAuthenticationMode, KubernetesBootstrapToken, KubernetesCloudProvider,
-    KubernetesClusterName, KubernetesDurationValue, KubernetesEvictionHardKey, KubernetesLabelKey,
-    KubernetesLabelValue, KubernetesQuantityValue, KubernetesReservedResourceKey,
-    KubernetesTaintValue, KubernetesThresholdValue, Lockdown, PemCertificateString,
-    SingleLineString, SysctlKey, TopologyManagerPolicy, TopologyManagerScope, Url, ValidBase64,
-    ValidLinuxHostname,
+    ECSAgentImagePullBehavior, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue,
+    FriendlyVersion, Identifier, KubernetesAuthenticationMode, KubernetesBootstrapToken,
+    KubernetesCloudProvider, KubernetesClusterName, KubernetesDurationValue,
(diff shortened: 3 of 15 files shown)


zmrow:
Hi @mello7tre !  Appreciate the contribution.  We're looking at this - we'd love to figure out a good way to accomplish this for both ECS/EKS in a more generic way.

bcressey:
Can you help me better understand your use case? Since the ECS agent just waits until the instance reaches the `Ready` state, what's actually happening behind the scenes to advance the state?

In #1788 @samjo-nyang proposed a method for pulling container images and then marking the instance as `Ready`. If you're using a similar host container approach, I'm not entirely clear on the value of that vs. just pulling the images in a bootstrap container before we even try to start `ecs-agent`.

mello7tre:
Sure, the problem arise if you use a CapacityProvider [CP], start an instance out of the AutoscalingGroup [ASG] that the CP is using and later attach it to the ASG (this is what happen for warm pools).

I have opend a ticket on AWS Support too, but then i found out that they have already taken in account that case and solved it with the code triggered by the param `ECS_WARM_POOLS_CHECK`.

In detail:
Seem that when a new container instance is registered to an ECS cluster using a CP there is a process that just one time check if that instance belongs to the CP ASG, if not, no further check are executed, so when later we attach the container instance to the CP ASG, it's not taken in account by the CP.

The effect is that even if the container instance is registered in the cluster, services configured to use the capacity providers are unable to start on that instance.

So, the `ECS_WARM_POOLS_CHECK`, do not check if the instance is in a `Ready` state, but wait until the instance is part of an AutoscalingGroup before registering it with the ECS Cluster.

README.md
```diff
@@ -425,7 +425,12 @@ These settings can be changed at any time.
...
 * `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
-
+* `settings.ecs.image-pull-behavior`: The behavior used to customize the [pull image process](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html#ecs-agent-availparam) for your container instances.
+  Supported values are `default`, `always`, `once`, `prefer-cached`, and the default is `default`.
+* `settings.ecs.warm-pools-check`: When true, the Amazon ECS agent polls the instance metadata to determine if the instance is ready to be included in the Amazon ECS cluster.
+  Set this variable to `true` only if you have an Auto Scaling group and want a warm pool associated with it.
+  ECS Agent wait until the Container Instance is associated to an AutoScaling Group before registering it with the Cluster.
+  Can be used anytime an ECS Container Instance is started outside of an AutoScaling Group and attached later.
 #### CloudFormation signal helper settings
```
bcressey: If you can change this to be a bit more generic, we think we can use the same setting (but a different implementation) to handle the warm pool check for kubelet also.

```suggestion
#### Auto Scaling group settings
* `settings.autoscaling.should-wait`: Whether to wait for the instance to reach the `InService` state before the orchestrator agent joins the cluster. Defaults to `false`. Set this to `true` only if the instance is part of an Auto Scaling group, or will be attached to one later.
```

The ecs-agent implementation works by polling IMDS for this `InService` state - https://github.com/aws/amazon-ecs-agent/commit/3909ba914699715902e9789791088b73e8b754ae.

I've suggested `settings.autoscaling.should-wait` for symmetry with `settings.cloudformation.should-signal` - but could be talked into something more verbose. An important point that I got from your use case is that the behavior isn't really specific to the ASG warm pools feature - it also applies when preparing instances outside of the ASG altogether.

sources/models/src/modeled_types/ecs.rs
```diff
@@ -228,3 +228,57 @@ mod test_ecs_agent_log_level {
...
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        serde_plain::from_str::<ECSImagePullBehavior>(&input).context(error::InvalidPlainValueSnafu {
+            field: "ecs.image_pull_behavior",
+        })?;
+        Ok(ECSAgentImagePullBehavior {
+            inner: input.to_string(),
+            inner_code: (serde_plain::from_str::<ECSImagePullBehavior>(&input).unwrap() as u8).to_string(),
+        })
```
bcressey: This avoids the `unwrap()` which nominally can panic, and the second conversion attempt:
```suggestion
        let behavior = serde_plain::from_str::<ECSImagePullBehavior>(&input).context(error::InvalidPlainValueSnafu {
            field: "ecs.image_pull_behavior",
        })?;
        Ok(ECSAgentImagePullBehavior {
            inner: input.to_string(),
            inner_code: (behavior as u8).to_string(),
        })
```

sources/api/ecs-settings-applier/src/ecs.rs
```diff
@@ -87,6 +93,8 @@ async fn run() -> Result<()> {
             .map(|s| s.to_string())
             .collect(),
         spot_instance_draining_enabled: ecs.enable_spot_instance_draining,
+        warm_pools_support: ecs.warm_pools_check,
+        image_pull_behavior: Some(ecs.image_pull_behavior.unwrap().inner_code.parse::<u8>().unwrap()),
```
bcressey: With the `TryInto` suggestion I made elsewhere, this would become something like:
```suggestion
        image_pull_behavior: ecs.image_pull_behavior.try_into() as u8,
```

sources/models/src/modeled_types/ecs.rs
```diff
@@ -228,3 +228,57 @@ mod test_ecs_agent_log_level {
...
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// ECSAgentImagePullBehavior represents a string that contains a valid ECS Image Pull Behavior for the ECS agent.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ECSAgentImagePullBehavior {
+    inner: String,
+    pub inner_code: String,
+}
```
bcressey: My concern with this representation is that it implies we'll end up serializing both the string and the code to the datastore, but the `string_impls_for!` macro only handles a single `inner` field in its serialize impl (and we only want it to).

Instead I'd suggest adding a `TryFrom<&str>` impl for `ECSImagePullBehavior`, which should give you a `TryInto` impl that you can use to convert the string into an enum value, which you can then cast to `u8`.

mello7tre: Sorry but i do not understand.
`sources/api/ecs-settings-applier/src/ecs.rs` read input, that is a string, from `ecs.image_pull_behavior`.
`ecs.image_pull_behavior` contains an object like:
```
        Ok(ECSAgentImagePullBehavior {
            inner: input.to_string(),
        })
```
How can i "have access" to the impl TryFrom/TryTo for `ECSImagePullBehavior`, that is an enum, in `sources/api/ecs-settings-applier/src/ecs.rs` ?
If i understand correctly, the enum values are only used to force validation.
Problem is that, in this case, i need to write in ecs json config not the input value but the enum one (casted as u8).

As i am a newby rust coder, maybe i am missing something obvious.

bcressey: It's not obvious or very common, but we do make use of the modeled types in other programs. Here's the gist of [what I had in mind](https://gist.github.com/bcressey/0d52bff9ec7ec23d68f9556cf47b5a58).

Essentially:
* add `TryFrom<&str>` impl to `ECSImagePullBehavior` (and reuse it in `ECSAgentImagePullBehavior`)
* make `ECSImagePullBehavior` pub, since we want its `TryFrom` impl outside the crate
* add a dependency on the `models` crate to `ecs-settings-applier`
* use the `TryFrom` impls to turn the string into the enum, and then cast as a u8

(The `unwrap()` is a bit of a smell but we know we haven't changed the value and that it was verified during deserialization, and it's awkward to go from an Option to a Result to an Option in a function chain.)

mello7tre: Thanks for the clear explanation and for the code.
I just made a push that include it and your other suggestion too.
If you want to do changes or write it down in a better way feel free to do it.

sources/models/shared-defaults/defaults.toml
```diff
@@ -149,3 +149,8 @@ restart-commands = ["/usr/bin/certdog"]
 
 [metadata.settings.pki]
 affected-services = ["pki"]
+
+# AutoScaling
+
+[settings.autoscaling]
+should-wait = false
```
bcressey: This would add the default to all models, even VMware and bare metal that run outside AWS.

You can put it in `ecs.toml` and `kubernetes-aws.toml` instead to cover all variants. But I'd prefer to limit it to `ecs.toml` and only add `AutoScalingSettings` to the two ECS variant models for now - otherwise we'll expose a setting on the Kubernetes variants that isn't actually implemented.

sources/models/src/lib.rs
```diff
@@ -214,6 +214,8 @@ struct ECSSettings {
     logging_drivers: Vec<SingleLineString>,
     loglevel: ECSAgentLogLevel,
     enable_spot_instance_draining: bool,
+    warm_pools_check: bool,
```
bcressey: Can remove this.

README.md
```diff
@@ -431,14 +431,22 @@ These settings can be changed at any time.
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
 * `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
-
+* `settings.ecs.image-pull-behavior`: The behavior used to customize the [pull image process](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html#ecs-agent-availparam) for your container instances.
+  Supported values are `default`, `always`, `once`, `prefer-cached`, and the default is `default`.
+* `settings.ecs.warm-pools-check`: When true, the Amazon ECS agent polls the instance metadata to determine if the instance is ready to be included in the Amazon ECS cluster.
```
bcressey: Can remove the `warm-pools-check` documentation here since it's superseded by the `should-wait` setting.

bcressey (commented):
Thanks! This looks really close, just a few clean-up suggestions. Let me know if you want us to handle them.

zmrow (approved):
LGTM!  Thanks!

:wave:

bcressey:
Can you squash all the commits together into a single commit? I'd also still prefer to back out the changes to `aws-k8s-*` and `aws-dev` since the functionality isn't supported on those variants yet, but we can do that in a follow-up PR if needed.

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -21,3 +21,6 @@ service-checks = ["apiserver", "chronyd", "containerd", "host-containerd", "kube
 
 [metadata.settings.network]
 affected-services = ["containerd", "kubernetes", "host-containerd", "host-containers"]
+
+[settings.autoscaling]
+should-wait = false
```
bcressey: Sorry to be a pain - can you back this specific change out as well? It's tied to the Kubernetes variant models.

bcressey (commented):
Nice!

sources/models/shared-defaults/kubernetes-aws.toml
mello7tre: no prob, i was in doubt if keep it or no, but as you specifically mentioned it before, i leaved it in.

bcressey:
CI failed for this reason:
```
error: the lock file /home/ec2-user/_work/bottlerocket/bottlerocket/sources/Cargo.lock needs to be updated but --locked was passed to prevent this
```

Looks like you'll need to add the modified `Cargo.lock` to the commit. It shouldn't be explicitly updated, or it will pull in a lot of other newer dependencies outside the scope of this PR. We just want the `Cargo.lock` with minor changes that you'd get from a clean checkout of your branch and running a sequence of commands like this:
```
cd sources
export VARIANT=aws-ecs-1
cargo test
```

If you can add that and squash together the commits again, I'll kick off the CI run.

mello7tre:
running `cargo tree` inside sources dir does the trick

bcressey:
Now the `cargo fmt` check is failing. You can run this to fix:
```
cargo fmt --manifest-path sources/Cargo.toml --message-format short --all
```

bcressey (approved):
LGTM, thanks!

mello7tre:
sorry for the troubles, next time (if any), i will remember to include Cargo.lock and to run format command.

zmrow (approved):
Thanks!

mello7tre:
let me know if i need to rebase it on latest commits to resolve conflicts....

bcressey:
@mello7tre yes, please rebase this if you don't mind, and then we can try to get it merged before more conflicts appear.

mello7tre:
rebased

bcressey (approved):

bcressey:
Thank you!

PR #2066, opened 2022-04-12
Enable network config generation via `netdog`

zmrow:
**Description of changes:**
Until now, metal variants were using the same hard-coded `eth0.xml` Wicked configuration file as other variants.  This set of commits adds the initial support for user-supplied network interface configuration.  Additional options and documentation will come after some additional use and feedback - this PR enables the feature.

```
    netdog: Add ability to generate network configuration
    
    This commit adds a subcommand `generate-net-config` to `netdog`.  It
    uses a user-provided TOML file `net.toml` to generate the proper wicked
    configuration files for the network interfaces in the system.  This file
    lives on the BOTTLEROCKET-PRIVATE partition.
    
    If no file is found, `netdog` looks at the kernel command line
    (/proc/cmdline) for the `netdog.default-interface` prefix and configures
    the interface listed there.  A single default interface may be defined
    on the kernel command line with the format:
    `netdog.default-interface=<interface name>:option1,option2`.
    "<interface name>" is the name of the interface, and valid options are
    "dhcp4" and "dhcp6".  A "?" may be added to the option to signify that
    the lease for the protocol is optional and the system chouldn't wait for
    it.  If the prefix does not exist on the kernel command line, no network
    configuration is generated.
    
    At configuration generation time, `netdog` writes the primary interface
    to `/var/lib/netdog/primary_interface`.  Primary interface may be set in
    `net.toml`; if it is absent in the file the first interface in the list
    is assumed to be the primary.  If the interface is passed via kernel
    command line, it is assumed to be the default.
    
    Previously, the `netdog install` command, which is called by wicked for
    configured interfaces, matched on the name of the interface (eth0)
    before it took any action.  The `install` command writes the
    resolv.conf, and the current_ip file.  For AWS/VMware we are explicitly
    only configuring `eth0`, and for Metal variants we will not be able to
    determine the configured interface names.  This commit changes the
    behavior.  Rather than filtering on interface name in order to write
    `resolv.conf` and the current IP file, we only write the files if the
    interface currently being configured is the primary interface (read from
    the previously written `primary_interface` file).
    
    Note: Previously, our network configuration for `eth0` included a few
    options, `arp-verify` and `arp-notify`.  These options were
    (theoretically) meant to speed up boot time in EC2 since we know our IP
    won't be a duplcate and therefore we don't need to ARP for it.  (That's
    not how networking in EC2 works)  In our testing, `arp-very/notify` made
    no significantly measureable difference in the time it took to get from
    "lease request" to "committed lease".  From this commit on, these
    options aren't part of our generated config for any variant.

```
```
    Use generated network configuration by default
    
    This commit removes the hard-coded `eth0.xml` network configuration file
    in favor of using `netdog` to generate the network configuration for all
    variants.  It adds a new systemd unit file to run `netdog
    generate-net-config` early in boot, before the network is up.  To
    generate the network configuration for AWS/VMware variants, we pass
    `netdog.default-interface=eth0:dhcp4,dhcp6?` which `netdog` interprets
    and generates network config similar to the hardcoded file previously
    used for these variants.
    
    For metal variants, we decided to use systemd-udevd's predictable device
    naming so we can count on network devices being named identically every
    boot.  We currently pass `net.ifnames=0` on the kernel command line
    which disables predictable naming, which is fine for AWS and VMware
    variants as hardware is controlled and instances typically initially
    come up with a single interface in the same PCIe location.  In order to
    continue using `net.ifnames=0` for AWS/VMware, we move this parameter
    out of the default kernel command line to the KERNEL_PARAMETERS section
    of each variants `Cargo.toml`.
    
    Aside: We previously passed `biosdevname=0` on the kernel command line
    as well.  `biosdevname` is a udev helper utility written by Dell for
    consistent device naming based on SMBIOS info.  We don't currently
    package the helper or include the udev rule that uses it, so we have
    removed the `biosdevname` parameter entirely.

```

**Testing done:**
Boot the following variants and ensure they come up correctly, network config is written properly, and they join the cluster, run pods, etc.  Also ensure that the generated network configuration matches what we currently use today, minus the arp-verify/notify bits (see netdog commit message above).  Ensure `primary_interface` is written.

```
bash-5.1# wicked show all
...
eth0            up
      link:     #2, state up, mtu 1500
      type:     ethernet, hwaddr 00:50:56:85:03:31
      config:   wicked:xml:/etc/wicked/ifconfig/eth0.xml
      leases:   ipv4 dhcp granted
      leases:   ipv6 dhcp requesting
      addr:     ipv4 198.19.129.141/15 [dhcp]
      route:    ipv4 default via 198.18.0.1 proto dhc
...

bash-5.1# more /etc/wicked/ifconfig/eth0.xml 
<interface><name>eth0</name><control><mode>boot</mode><link-detection><require-link></require-link></link-detection></control><ipv4:dhcp><enabled>
true</enabled></ipv4:dhcp><ipv6:dhcp><enabled>true</enabled><defer-timeout>1</defer-timeout><flags><optional></optional></flags></ipv6:dhcp></inte
rface>
```

- [x] aws-dev
- [x] aws-ecs-1
- [x] aws-k8s-1.19
- [x] aws-k8s-1.20
- [x] aws-k8s-1.21
- [x] aws-k8s-1.22
- [x] aws-k8s-1.21-nvidia
- [x] aws-k8s-1.22-nvidia
- [x] vmware-k8s-1.20
- [x] vmware-k8s-1.21
- [x] vmware-k8s-1.22

For metal (tested on on my SuperMicro SYS-E200-8D): 
* ensure  no config gets written in the absence of `net.toml`/kernel arg
* write a `net.toml` and test all 4 interfaces work
* ensure that the "primary" interface is chosen correctly via config or in the absence of the `primary` key the first interface is chosen
* ensure that `current_ip` is written with the IP of the primary interface
* ensure that all interfaces configured in `net.toml` get a configuration file written, and if they do not have an active link that wicked waits for them
* ensure active links get an IP.

- [x] metal-dev
- [x] metal-k8s-1.21
- [x] metal-k8s-1.22

**Terms of contribution:**


zmrow:
^ Push above:
* fixes a few typos
* Removes the last remaining references to `eth0.xml` from the `release` package
* Ensures that the `/etc/wicked/ifconfig` directory is created
* Changes the behavior of `netdog install` a little bit - rather than matching on network interface name (which, for metal, we can't know at the time of invocation), we only create the resolv.conf and current_ip files if they don't exist.  This feels like the proper behavior since the `netdog` command is only called for configured interfaces (eth0 for AWS/VMware) and (for now) it shouldn't matter which interface is chosen to decide the resolv.conf or current IP.

packages/os/network-config.service
```diff
@@ -0,0 +1,20 @@
...
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/netdog generate-net-config
+RemainAfterExit=true
+StandardError=journal+console
+
+[Install]
+WantedBy=multi-user.target
```
arnaldo2792: I don't think this directive is required, if you already made this unit `RequiredBy=preconfigured.target`

sources/api/netdog/src/main.rs
```diff
@@ -60,12 +71,6 @@ struct LeaseInfo {
     dns_search: Option<Vec<String>>,
 }
 
-#[derive(Debug, PartialEq, Deserialize)]
-#[serde(rename_all = "kebab-case")]
-enum InterfaceName {
```
zmrow, line 65: This was the only thing I was a little leery of.  However, considering that wicked only calls `install` for interfaces that we are explicitly configuring it seems like this guard is unnecessary.

packages/os/network-config.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Bottlerocket network configuration system
+Requires=var-lib-bottlerocket.mount
+After=local-fs.target
+Before=wickedd.service network-pre.target
+Wants=network-pre.target
```
arnaldo2792: Did you add this so that we can have the `network-pre.target` available for other units, and use it as convergence point? Or do you expect something being initialized by the target? If the later, then I'd just remove it since there is nothing initialized by `network-pre.target`.

sources/api/netdog/README.md
```diff
@@ -14,6 +14,10 @@ It contains two subcommands meant for use as settings generators:
 
 The subcommand `set-hostname` sets the hostname for the system.
 
+The subcommand `generate-net-config` generates the network interface configuration for the host.
+If a `net.toml` exists, it is used to generate the configuration, otherwise a default configuration
```
arnaldo2792: Nit:

```suggestion
If a `net.toml` file exists in `/var/lib/bottlerocket`, it is used to generate the configuration, otherwise a default configuration
```

packages/os/network-config.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Bottlerocket network configuration system
+Requires=var-lib-bottlerocket.mount
```
arnaldo2792: Nit: Maybe a note here saying that the mount should be there because the service will attempt to read from it? Also in other units we use this when we need a mount to be present:

```suggestion
RequiresMountsFor=/var/lib/bottlerocket
```

sources/api/netdog/build.rs
```diff
@@ -4,8 +4,31 @@ use std::env;
...
+        } else if variant.starts_with("metal") {
+            println!("cargo:rustc-cfg=bottlerocket_platform=\"metal\"");
+        } else {
+            eprintln!(
+            "For local builds, you must set the 'VARIANT' environment variable so we know which data \
+            provider to build. Valid values are the directories in models/src/variants/, for \
+            example 'aws-ecs-1'."
+            );
+            process::exit(1);
+        }
```
arnaldo2792: I had to do something similar somewhere else (not yet in the code), what do you think about?

```suggestion
let platform = variant.split('-').collect::<Vec<&str>>()[0].to_string();
println!("cargo:rustc-cfg=bottlerocket_platform=\"{}\"", platform);
```

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,280 @@
...
+                enabled: o.enabled,
+                route_priority: o.route_metric,
+                _f: (),
+            },
+        }
+    }
+}
+
+// The default configuration for AWS based variants
+#[cfg(any(bottlerocket_platform = "aws", bottlerocket_platform = "vmware"))]
```
arnaldo2792: Nit: the comment and the code aren't consistent. Should the `vmware` flag be used below, with the `metal` platform?

packages/os/network-config.service
zmrow: I may have misinterpreted what I [read here](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/), but it seemed like I should have it.

>Services that want to be run before the network is configured should place Before=network-pre.target and also set Wants=network-pre.target to pull it in. This way, unless there's actually a service that needs to be ordered before the network is up the target is not pulled in, hence avoiding any unnecessary synchronization point.

arnaldo2792:
For the lack of a better place to add this comment:

Nit: I believe the commit message in `netdog: Add ability to generate network configuration` has a grammatical error, or else I need to learn this :smile:  :

```
For AWS/VMware we are explicitly only configure `eth0`
```

sources/api/netdog/src/main.rs
```diff
@@ -10,13 +10,23 @@ It contains two subcommands meant for use as settings generators:
...
 
 #[macro_use]
 extern crate serde_plain;
 
+mod net_config;
+use net_config::NetConfig;
+
+mod wicked;
+use wicked::WickedInterface;
+
```
webern: nit: is there a reason to keep these use statements separate from all the others? rust fmt will just sort the lists of mods and uses if they are aggregated.

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,280 @@
...
+            ipv4: None,
+            ipv4_dhcp: wicked_dhcp4,
+            ipv6_dhcp: wicked_dhcp6,
+        }
+    }
+
+    /// Serialize the interface's configuration file
+    // Consume `self` to enforce that changes aren't made to the interface type after it has been
+    // written to file
+    pub(crate) fn into_config_file(self) -> Result<()> {
```
webern: `into`, generally implies that a converted type will be returned. Also the function becomes test-able if it takes the path and extension as input.

```suggestion
    pub(crate) fn write_config_file(self, config_dir: P, extention: S) -> Result<()> {
```

sources/api/netdog/src/wicked.rs
webern: Can you reduce the conditionally-compiled code even further to a set of constants?

```rust
#[cfg(any(bottlerocket_platform = "aws", bottlerocket_platform = "vmware"))]
const DEFAULT_WICKED_IPV4: Option<WickedIpv4> = Some(WickedIpv4{ /* ... */ });

#[cfg(bottlerocket_platform = "metal")]
const DEFAULT_WICKED_IPV4: Option<WickedIpv4> = Some(WickedIpv4{ /* ... */ });

impl Default for WickedInterface {
    fn default() -> Self {
WickedInterface {
            name: DEFAULT_INTERFACE.to_string(),
            control: WickedControl::default(),
            ipv4: DEFAULT_WICKED_IPV4,
           /* etc */
        }
    }
}
```

You may also be able to emit a useful error message in the case that someone is trying to define a new variant. (This is pseudo code for the cfg part...)

```rust
#[cfg(none_of(bottlerocket_platform = "aws", bottlerocket_platform = "vmware", bottlerocket_platform = "metal"))]
compile_error!("Unexpected bottlerocket_platform. Default Wicked settings must be defined.");
```

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    #[test]
+    #[cfg(bottlerocket_platform = "metal")]
+    fn default() {
+        let expected: &str = include_str!("../test_data/wicked/default.xml");
+
+        let generated = serde_xml_rs::to_string(&WickedInterface::default()).unwrap();
+        assert_eq!(expected.trim(), generated)
+    }
```
webern: If you did what I suggested above and conditionally define the constants, then you could conditionally define the test file's path as a constant here in the tests mod. That way there would only be one test for checking the default `WickedInterface`.

sources/api/netdog/src/net_config.rs
```diff
@@ -0,0 +1,40 @@
...
+// into an enum with variants for each version, i.e.
+// enum NetConfig {
+//     V1(NetInterfaceV1)
+//     V2(NetInterfaceV2)
+// }
+#[derive(Debug, Deserialize)]
+pub(crate) struct NetConfig {
+    // This attribute is needed to support versions, but we aren't using it yet
+    #[allow(dead_code)]
+    pub(crate) version: u8,
```
webern, line 28: Maybe you should check this value somewhere and raise an error if it is not `1`? That way, when a version `2` exists, you would provide a useful error message if someone tried to deserialize it, e.g. after a downgrade.

sources/api/netdog/src/main.rs
```diff
@@ -280,6 +294,38 @@ fn generate_hostname() -> Result<()> {
     Ok(print_json(hostname)?)
 }
 
+/// Generate configuration for network interfaces.
+fn generate_net_config() -> Result<()> {
```
webern, line 311: I think this whole thing becomes test-able if you take the constants as input.

```suggestion
fn generate_net_config(net_config: P1, wicked_config_dir: P2, wicked_ext: S) -> Result<()> {
```

sources/api/netdog/build.rs
zmrow: It's definitely less code, and I like the fact that we don't have to specify all the variants, but I could see a few issues.
* We don't get a nice error message if someone misspells the variant - we get a more cryptic compiler message
* I think it's reasonable to assume that a variant name wouldn't be zero-length (covering the index issue), but we get the same issue as above where the build will happen with a misspelled `cargo:rustc-cfg=bottlerocket_platform="awsk8s121"` and we'd get a less than friendly error.

zmrow:
^ The above fixes a few typos and adds a `rename` to `route-metric` in `net_config`.

zmrow:
Flipping this out of draft since I feel pretty comfortable with the functionality.

sources/api/netdog/build.rs
arnaldo2792: - For 1), in the current implementation I can do `aws-not-existing` and get a compiler error
- For 2), in the current implementation, `awsk8s121` will cause the same error as 1)

So the question is how much validation do we want to do while parsing the variant's name to provide a friendly error?, or else we have to trade off some errors for others when the user passes an invalid variant.

zmrow:
^ fixed the test data :facepalm:

etungsten (approved):

sources/api/netdog/src/main.rs
zmrow: I'm not sure it's necessary to have a unit test for this function.  Other than the filesystem operations (which I don't really see as needing to be tested), this function basically drives the other module functions.  The module functionality is tested in the `wicked` module!

sources/api/netdog/src/wicked.rs
zmrow: I see that the `ipv4/6_dhcp` defaults could probably be factored out as well.  That would help a bunch.

sources/api/netdog/src/main.rs
webern: One could always make that argument! I'm seeing a fair amount of logic here "if the file exists do x", "if interfaces is set and dhcp4 is not set raise an error).

sources/api/netdog/src/net_config.rs
zmrow: Good point - I'll do this in `main` after the deserialize for the short term.  I've already made good headway into the code that will handle versions which will come as a fast-follow to this PR.

zmrow:
^ The above push:
* Corrects the systemd unit to `RequireMountsFor`
* Simplify the `build.rs` logic in favor of a `compile_error!()` if the platform is not known
* Reduce the conditionally compiled code into constants and implement `Default` for a few more types to simplify the `Default` implementation for `WickedInterface`
* Change `into_config_file()` to `write_config_file()`
* As a result of the above changes, simplify the test code in `wicked.rs` using conditionally compiled defaults

zmrow:
^ Adds a quick net config version check.  This check will be removed when we handle versioning (already in progress).

I believe at this point I've addressed all of the comments minus @webern 's comment about testing the `generate_net_config()` function.  I do agree there's some logic that doesn't get tested there - some of it will be removed when we start handling versions.

sources/api/netdog/build.rs
zmrow: I implemented your suggestion, and opted to add a `compile_error!()` after a `cfg` directive - see `wicked.rs`

webern (approved):

packages/os/network-config.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Bottlerocket network configuration system
```
bcressey: nit:
```suggestion
Description=Generate network configuration
```

packages/os/network-config.service
```diff
@@ -0,0 +1,20 @@
...
+# The service attempts to read from this mount
+RequiresMountsFor=/var/lib/bottlerocket
+After=local-fs.target
+Before=wickedd.service network-pre.target
+Wants=network-pre.target
```
bcressey: Since `wicked` comes after `network-pre.target`, we can just use the target as the synchronization point:
```suggestion
After=local-fs.target
Before=network-pre.target
```

packages/os/network-config.service
```diff
@@ -0,0 +1,20 @@
...
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/netdog generate-net-config
+RemainAfterExit=true
+StandardError=journal+console
+
+[Install]
+RequiredBy=preconfigured.target
```
bcressey:
```suggestion
[Install]
RequiredBy=network-pre.target
```

packages/os/os.spec
```diff
@@ -45,6 +45,7 @@ Source114: bootstrap-containers@.service
 Source115: link-kernel-modules.service
 Source116: load-kernel-modules.service
 Source117: cfsignal.service
+Source118: network-config.service
```
bcressey: nit: to disambiguate from a service that configures network devices:
```suggestion
Source118: generate-network-config.service
```

packages/wicked/wicked.spec
```diff
@@ -91,6 +91,7 @@ sed -i 's|^runpath_var=LD_RUN_PATH|runpath_var=DIE_RPATH_DIE|g' libtool
 # install custom configuration
 rm -rf %{buildroot}%{_cross_sysconfdir}/wicked
 install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/wicked
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/wicked/ifconfig
```
bcressey: We shouldn't need to create or package this directory; the contents of `/usr/share/factory` aren't needed for all `tmpfiles` directives, just the ones that need to copy content into place.

variants/aws-dev/Cargo.toml
```diff
@@ -12,7 +12,8 @@ kernel-parameters = [
     "console=tty0",
     "console=ttyS0,115200n8",
     # Only reserve if there are at least 2GB
-    "crashkernel=2G-:256M"
+    "crashkernel=2G-:256M",
+    "net.ifnames=0"
```
bcressey: nit: this would be a good time to add the trailing comma to all list elements, since that's more friendly for future diffs
```suggestion
    "crashkernel=2G-:256M",
    "net.ifnames=0",
```

sources/api/netdog/build.rs
```diff
@@ -6,6 +6,16 @@ use std::io::Write;
...
 fn main() {
+    // The code below emits `cfg` operators to conditionally compile this program based on the
+    // current variant.
+    // TODO: Replace this approach when the build system supports ideas like "variant
+    // tags": https://github.com/bottlerocket-os/bottlerocket/issues/1260
+    println!("cargo:rerun-if-env-changed=VARIANT");
+    if let Ok(variant) = env::var("VARIANT") {
+        let platform = variant.split('-').collect::<Vec<&str>>()[0].to_string();
+        println!("cargo:rustc-cfg=bottlerocket_platform=\"{}\"", platform);
+    }
```
bcressey: It's worth considering whether we want to allow control over netdog behavior via kernel command line parameters, since it's already influenced by that to a great extent. Then the default netdog configuration would live in one place (variant kernel parameters) rather than partially there and partially hardcoded at build time.

For example, we could have a parameter like:
```
netdog.default-interface=eth0:dhcp4,dhcp6?
```

To indicate eth0 as interface, with dhcp4 and optional dhcp6.

If we care about the arp verify & notify stuff - we might not, we should measure! - then we could add another sigil like `+` to the protocol to indicate  that the lease IP is trusted to be unique.

(As one benefit: this would facilitate experimentation with setups like DHCP6 only instances in EC2.)

sources/api/netdog/src/main.rs
```diff
@@ -10,23 +10,32 @@ It contains two subcommands meant for use as settings generators:
 * `generate-hostname`: returns the node's hostname in JSON format. If the lookup is unsuccessful, the IP of the node is used.
 
 The subcommand `set-hostname` sets the hostname for the system.
+
+The subcommand `generate-net-config` generates the network interface configuration for the host.
+If a `net.toml` file exists in `/var/lib/bottlerocket`, it is used to generate the configuration,
+otherwise a default configuration is written for interface `eth0` (AWS/VMware variants) or `eno1`
```
bcressey: If we end up using the kernel command line, we can just skip generating any config if don't have the parameter or `net.toml` - that could be nice for variants that don't care about networking for some reason.

sources/api/netdog/src/main.rs
```diff
@@ -220,19 +230,23 @@ fn write_current_ip(ip: &IpAddr) -> Result<()> {
...
+                // Randomize name server order, for libc implementations like musl that send
+                // queries to the first N servers.
+                let mut dns_servers: Vec<_> = info.dns_servers.iter().collect();
+                dns_servers.shuffle(&mut thread_rng());
+                write_resolv_conf(&dns_servers, &info.dns_search)?;
+            }
+
+            if !Path::new(CURRENT_IP).exists() {
+                write_current_ip(&info.ip_address.addr())?;
+            }
```
bcressey: I'm a bit skeptical but OK with the "skip if exists" logic for `resolv.conf` but it seems like a serious problem for the current IP, which we may later use to generate the node IP that kubelet registers with.

If we're DHCP'ing on multiple interfaces then we might acquire leases in any order, and if only one of those addresses is "correct" then that will be quite racy.

I expect we need a concept like "primary interface", which we either add as an explicit boolean in `net.toml` or implicitly select from the first element in the list. Then we can write that interface into `/var/lib/netdog/primary_interface` and update `current_ip` if the lease we're handling is for that interface.

(We could use the same logic to strengthen our contract around where we source `resolv.conf` from.)

sources/api/netdog/src/main.rs
```diff
@@ -280,6 +294,47 @@ fn generate_hostname() -> Result<()> {
...
+            .context(error::InterfaceConfigWriteSnafu);
+    }
+
+    let net_config_str = fs::read_to_string(NET_CONFIG)
+        .context(error::NetConfigReadFailedSnafu { path: NET_CONFIG })?;
+    let net_config: NetConfig =
+        toml::from_str(&net_config_str).context(error::NetConfigParseSnafu { path: NET_CONFIG })?;
+
+    // TODO Get rid of this version check and validate version during deserialization.  See TODO in
+    // `net_config.rs`
```
bcressey: nit: I think one `TODO` for the feature is probably enough

sources/api/netdog/src/net_config.rs
```diff
@@ -0,0 +1,39 @@
+//! The net_config module contains the strucures needed to deserialize a `net.toml` file.
+//!
+//! These structures are the user-facing options for configuring 1 or more network interfaces.
```
bcressey: nit:
```suggestion
//! These structures are the user-facing options for configuring one or more network interfaces.
```

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,277 @@
+//! The wicked module contains the data structures and functions needed to create network interface
+//! configuration files for wicked.
+//!
+//! The structures in this module are meant to be created from the user-facing structures in the
+//! `net_config` module, hence the `From` implementations.  `Default` implementations for
+//! WickedInterface exist here as well, which provide sane defaults in the absence of a desired
```
bcressey: nit:
```suggestion
//! WickedInterface exist here as well, which provide sensible defaults in the absence of a desired
```

sources/api/netdog/src/net_config.rs
```diff
@@ -0,0 +1,39 @@
...
+// into an enum with variants for each version, i.e.
+// enum NetConfig {
+//     V1(NetInterfaceV1)
+//     V2(NetInterfaceV2)
+// }
+#[derive(Debug, Deserialize)]
+pub(crate) struct NetConfig {
+    pub(crate) version: u8,
+    #[serde(flatten)]
+    pub(crate) interfaces: HashMap<String, NetInterface>,
```
bcressey: We should have some kind of validation on interface names - at the very least a `SingleLineString`, but ideally something that matches the kernel's rules.

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,277 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    mode: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    link_detection: Option<String>,
+    // `serde_xml_rs` has a known issue with serializing nested structures, where it will insert
+    // additional tag with the structure name.  It has since been fixed but not released officially
+    // yet.  This struct member works around that issue.
+    // https://github.com/RReverser/serde-xml-rs/issues/126
+    // The workaround:
+    // https://stackoverflow.com/questions/70124048/how-to-create-xml-from-struct-in-rust
```
bcressey: Maybe a `TODO` tag here since we're somewhat likely to update the crate without remembering that we can remove the workaround. Or we could add a unit test to force the workaround to be removed when the behavior changes.

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,277 @@
...
+        }
+    }
+}
+
+impl Default for WickedDhcp6 {
+    fn default() -> Self {
+        WickedDhcp6 {
+            enabled: Some(true),
+            defer_timeout: Some(1),
+            flags: Some("<optional />".to_string()),
```
bcressey: It seems weird for this "optional" parameter to sneak in here; it's pretty EC2 specific, and an ugly hack at that.

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,277 @@
...
+            _f: (),
+        }
+    }
+}
+
+impl From<Dhcp4> for WickedDhcp4 {
```
bcressey: Might be good to prefix or suffix this `Dhcp4` struct, e.g. `Dhcp4Config`.

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,277 @@
...
+const DEFAULT_WICKED_IPV4: Option<WickedIpv4> = Some(WickedIpv4 {
+    arp_verify: Some(false),
+    arp_notify: Some(false),
+    _f: (),
+});
+
+#[cfg(bottlerocket_platform = "metal")]
+const DEFAULT_INTERFACE: &str = "eno1";
+#[cfg(bottlerocket_platform = "metal")]
+const DEFAULT_WICKED_IPV4: Option<WickedIpv4> = None;
```
bcressey: Even if the `netdog.default-interface` thing doesn't happen, it'd be worth pulling these defaults out of this module for a cleaner separation of concerns.

Right now we're sort of combining "platform defaults" with "struct defaults" in a way that I find confusing.

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,277 @@
...
+            ipv4: None,
+            ipv4_dhcp: wicked_dhcp4,
+            ipv6_dhcp: wicked_dhcp6,
+        }
+    }
+
+    /// Serialize the interface's configuration file
+    // Consume `self` to enforce that changes aren't made to the interface type after it has been
+    // written to file
+    pub(crate) fn write_config_file(self) -> Result<()> {
```
bcressey: I don't feel strongly about this but it seems like an unusual interface.

If we took a reference to self, then it might be easier to unit test that we can round trip back to an equivalent struct.

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,277 @@
...
+        let mut cfg_path = Path::new(WICKED_CONFIG_DIR).join(&self.name);
+        cfg_path.set_extension(WICKED_FILE_EXT);
+
+        // TODO: pretty print these files.  `serde_xml_rs` doesn't support pretty printing;
+        // `quick_xml` does, however we require a few features that haven't been released yet to
+        // properly serialize the above data structures:
+        // https://github.com/tafia/quick-xml/issues/340#issuecomment-981093602
```
bcressey, line 216: Or we could just decide not to bother with pretty-printing. 😀

packages/os/network-config.service
zmrow: That makes sense - I think we still need the `Wants=network-pre.target` though, based what I [read here](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/).

> Services that want to be run before the network is configured should place Before=network-pre.target and also set Wants=network-pre.target to pull it in. This way, unless there's actually a service that needs to be ordered before the network is up the target is not pulled in, hence avoiding any unnecessary synchronization point.

zmrow:
^ This force push was a pretty good sized change, hence requesting all reviewers again.

* Remove all conditionally compiled code - This is the biggest change here.  Per @bcressey 's request, I've moved all default interface definition to a kernel command line argument.  All AWS/VMware variants now set their default interface in the KERNEL_PARAMETERS.  `netdog` parses this value in the absence of a `net.toml`.  See the commit message for all the nitty gritty.
* Add interface name validation according to the rules defined in the kernel
* Add the ability for a user to specify the primary interface via a key in the `net.toml`.  In the absence of the key, the first value in the map is chosen.  In order to preserve the order of the interfaces (it's a map), we are now using the `toml` crate feature `preserve-order` which uses an `IndexMap` under the hood; we also use `IndexMap` for the interfaces.
* Generate `resolve.conf` and `current_ip` only for the primary interface
* Add default interface definitions to all AWS/VMware variants

zmrow:
I'll spin back through the variants again to make sure they're all tested, but my smoke testing of `metal-dev`, `aws-dev`, and `aws-k8s-1.21` look great.

sources/api/netdog/src/net_config.rs
zmrow: I've moved this to the `from_path()` method here.  It will end up being in the versioning code to be out after this PR.

packages/os/generate-network-config.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Generate network configuration
+# The service attempts to read from this mount
+RequiresMountsFor=/var/lib/bottlerocket
+After=local-fs.target
+Before=network-pre.target
+Wants=network-pre.target
```
bcressey: These are redundant with the `RequiredBy` below, and should be omitted.

packages/os/generate-network-config.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Generate network configuration
+# The service attempts to read from this mount
+RequiresMountsFor=/var/lib/bottlerocket
+After=local-fs.target
```
bcressey: This is part of the default dependencies on `sysinit.target` and doesn't need to be specified.

sources/api/netdog/README.md
```diff
@@ -14,6 +14,16 @@ It contains two subcommands meant for use as settings generators:
...
 The subcommand `set-hostname` sets the hostname for the system.
 
+The subcommand `generate-net-config` generates the network interface configuration for the host. If
+a `net.toml` file exists in `/var/lib/bottlerocket`, it is used to generate the configuration. If
+`net.toml` doesn't exist, the kernel command line `/proc/cmdline` is checked for the prefix
+`netdog.default-interface`.  If an interface is defined with that prefix, it is used to generate an
+interface configuration.  A single default interface may be defined on the kernel command line with
+the format: `netdog.default-interface=<interface name>:option1,option2`.  "<interface name>" is the
+name of the interface, and valid options are "dhcp4" and "dhcp6".  A "?" may be added to the option
+to signify that the lease for the protocol is optional and the system chouldn't wait for it.  A
```
bcressey:
```suggestion
to signify that the lease for the protocol is optional and the system shouldn't wait for it.  A
```

sources/api/netdog/README.md
```diff
@@ -14,6 +14,16 @@ It contains two subcommands meant for use as settings generators:
 
 The subcommand `set-hostname` sets the hostname for the system.
 
+The subcommand `generate-net-config` generates the network interface configuration for the host. If
+a `net.toml` file exists in `/var/lib/bottlerocket`, it is used to generate the configuration. If
+`net.toml` doesn't exist, the kernel command line `/proc/cmdline` is checked for the prefix
+`netdog.default-interface`.  If an interface is defined with that prefix, it is used to generate an
+interface configuration.  A single default interface may be defined on the kernel command line with
+the format: `netdog.default-interface=<interface name>:option1,option2`.  "<interface name>" is the
```
bcressey: "\<interface name\>" is not rendering in the markdown view.

sources/api/netdog/src/interface_name.rs
```diff
@@ -0,0 +1,195 @@
+//! The interface_name module contains the definition of a valid network interface name and the
+//! deserializtion code to support creation of the structure from string.
+//!
+//! A valid network interface name is defined by the criteria in the linux hernel:
```
bcressey:
```suggestion
//! A valid network interface name is defined by the criteria in the Linux kernel:
```

sources/api/netdog/src/interface_name.rs
```diff
@@ -0,0 +1,195 @@
+//! The interface_name module contains the definition of a valid network interface name and the
+//! deserializtion code to support creation of the structure from string.
```
bcressey:
```suggestion
//! deserialization code to support creation of the structure from string.
```

sources/api/netdog/src/interface_name.rs
```diff
@@ -0,0 +1,195 @@
...
+//!
+//! A valid network interface name is defined by the criteria in the linux hernel:
+//! https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/core/dev.c#n988
```
bcressey: Can this be a permalink of some kind? You can use `elixir.bootlin.com` links to a particular kernel version.

sources/api/netdog/src/net_config.rs
```diff
@@ -0,0 +1,197 @@
+//! The net_config module contains the strucures needed to deserialize a `net.toml` file.
+//!
+//! These structures are the user-facing options for configuring one or more network interfaces.
+use crate::interface_name::{deserialize_interfaces_map, InterfaceName};
+use indexmap::IndexMap;
+use serde::Deserialize;
+use snafu::{ensure, OptionExt, ResultExt};
+use std::{fs, path::Path};
+
+pub(crate) static NET_CONFIG: &str = "/var/lib/bottlerocket/net.toml";
```
bcressey:
```suggestion
pub(crate) static DEFAULT_NET_CONFIG_FILE: &str = "/var/lib/bottlerocket/net.toml";
```

I also think it'd make more sense to define this in `main.rs` where it's used, rather than making it part of the module's interface to specify a default.

sources/api/netdog/src/main.rs
```diff
@@ -220,19 +238,24 @@ fn write_current_ip(ip: &IpAddr) -> Result<()> {
...
-            let mut dns_servers: Vec<_> = info.dns_servers.iter().collect();
-            dns_servers.shuffle(&mut thread_rng());
-            write_resolv_conf(&dns_servers, &info.dns_search)?;
-            write_current_ip(&info.ip_address.addr())?;
+    // Wicked doesn't mangle interface names, but let's be defensive.
+    let install_interface = args.interface_name.trim().to_lowercase();
+    let primary_interface =
+        fs::read_to_string(PRIMARY_INTERFACE).context(error::PrimaryInterfaceReadSnafu {
+            path: PRIMARY_INTERFACE,
+        })?;
```
bcressey: Two nits:
* should we also apply `trim().to_lowercase()` to the string we read from the file?
* could return early if it's not the primary interface, so we don't need the conditional in the match statement

sources/api/netdog/src/main.rs
```diff
@@ -280,6 +303,93 @@ fn generate_hostname() -> Result<()> {
...
 }
 
+/// Generate configuration for network interfaces.
+fn generate_net_config() -> Result<()> {
+    if !Path::exists(Path::new(NET_CONFIG)) {
+        match fetch_default_interface(KERNEL_CMDLINE)? {
+            Some(interface_str) => {
+                let default_interface = WickedInterface::from_str(&interface_str)
+                    .context(error::InvalidInterfaceCmdlineSnafu)?;
+                write_primary_interface(default_interface.name())?;
```
bcressey: Consider adding a `NetConfig::from_command_line` associated function to streamline this to something like:
```
let net_config = if Path::exists(Path::new(NET_CONFIG)) {
    NetConfig::from_path(NET_CONFIG).context(...)?
} else {
    NetConfig::from_command_line(KERNEL_CMDLINE).context(...)?
}

if net_config.is_none() {
    return Ok(());
}

// then use same logic regardless of config source
```

The reason I prefer this is because it'd be easy to "forget" to apply a critical action, such as writing the primary interface marker, in one branch or the other, and consolidating them means that only one section of code needs to be changed.

EDIT: it looks like the lack of symmetry comes from not exposing the "optional" field on Dhcp4Config and Dhcp6Config, which we define as "go ahead and try to acquire a lease but do it in the background after 1 second."

We could just go ahead and allow that to simplify the code - letting us work with NetConfig and NetInterfaces until we're ready to turn those into WickedInterfaces.

sources/api/netdog/src/main.rs
```diff
@@ -280,6 +303,93 @@ fn generate_hostname() -> Result<()> {
...
+    write_primary_interface(primary_interface)?;
+
+    for (name, config) in net_config.interfaces {
+        if config.dhcp4.is_none() && config.dhcp6.is_none() {
+            return error::InvalidNetConfigSnafu {
+                interface: name.to_string(),
+                reason: "must configure dhcp4 or dhcp6, or both",
+            }
+            .fail();
+        }
```
bcressey: This check is present in the kernel command line case, although it's in a very different form - we don't allow bare interface names like `eno1`, and some options are required, so the parse will fail if either `dhcp4` or `dhcp6` aren't specified.

I find this much clearer, which is part of why I'm advocating for unifying behind `NetConfig` + `NetInterface` structs at this layer.

sources/api/netdog/src/net_config.rs
```diff
@@ -0,0 +1,197 @@
...
+//     V1(NetInterfaceV1)
+//     V2(NetInterfaceV2)
+// }
+#[derive(Debug, Deserialize)]
+pub(crate) struct NetConfig {
+    pub(crate) version: u8,
+    // Use an IndexMap to preserve the order of the devices defined in the net.toml.  The TOML
+    // library supports this through a feature making use of IndexMap.  Order is important because
+    // we use the first device in the list as the primary device if the `primary` key isn't set for
+    // any of the devices.
```
bcressey, line 32: It seems equally reasonable to pick the primary interface as the first interface in key order, and use a BTreeMap instead of an IndexMap. That'd get rid of the additional dependency, but this is fine also.

sources/api/netdog/src/interface_name.rs
```diff
@@ -0,0 +1,195 @@
...
+    Deserializer, Serialize, Serializer,
+};
+use snafu::ensure;
+use std::{convert::TryFrom, fmt, ops::Deref};
+use toml::Value;
+
+/// InterfaceName can only be created from a string that contains a valid network interface name.
+/// Validation is handled in the `TryFrom` implementation below.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub(crate) struct InterfaceName {
```
bcressey: It looks like you can get rid of the entire `deserialize_interfaces_map` function by using the `TryFrom<&str>` impl for deserialization.

```suggestion
#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
#[serde(try_from = "&str")]
pub(crate) struct InterfaceName {
```

I tried this with a local checkout of your branch and all the tests passed, after removing the function and the field deserializer in `NetConfig`.

sources/api/netdog/src/net_config.rs
```diff
@@ -0,0 +1,197 @@
...
+                reason: "invalid version"
+            }
+        );
+
+        ensure!(
+            !net_config.interfaces.is_empty(),
+            error::InvalidNetConfigSnafu {
+                reason: "no interfaces defined"
+            }
+        );
```
bcressey: We could potentially allow a NetConfig with no interfaces. I'm not sure exactly what the use case is, but it seems to match up with the kernel command line case - where we might not have an interface defined there either. Might be better to let the calling code decide what to do, like we do with our priority ladder for user data.

The main complication is the `primary_interface()` method, but that could return an `Option<String>` instead.

sources/api/netdog/src/net_config.rs
```diff
@@ -0,0 +1,197 @@
...
+        let mut maybe_primary = self
+            .interfaces
+            .iter()
+            .filter(|(_, v)| v.primary == Some(true));
+
+        let primary_interface = if let Some(interface) = maybe_primary.next() {
+            ensure!(
+                maybe_primary.next().is_none(),
+                error::InvalidNetConfigSnafu {
+                    reason: "multple primary interfaces defined",
```
bcressey:
```suggestion
                    reason: "multiple primary interfaces defined",
```

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,504 @@
...
+//! configuration files for wicked.
+//!
+//! The structures in this module are meant to be created from the user-facing structures in the
+//! `net_config` module or from a string, hence the `From` and `FromStr` implementations.  `Default` implementations for
+//! WickedInterface exist here as well, which provide sensible defaults in the absence of a desired
+//! network configuration.
+//!
+//! It's also important to note that there are network configuration options here that are not
+//! exposed to the user via `NetConfig`.  These options exist only to support the current sane
+//! defaults in place for the AWS platform.
```
bcressey: Is this still the case, now that `arp-notify` and `arp-verify` are gone?

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,504 @@
...
+use crate::{
+    interface_name::InterfaceName,
+    net_config::{Dhcp4Config, NetInterface},
+};
+use serde::Serialize;
+use snafu::{ensure, OptionExt, ResultExt};
+use std::{collections::HashSet, convert::TryInto, fs, ops::Deref, path::Path, str::FromStr};
```
bcressey: nit: I'm used to seeing each `use std::X` import on separate lines, as though each were a different crate.

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,504 @@
...
+#[derive(Default, Clone, Debug, Serialize, PartialEq)]
+struct AddrConfFlags {
+    optional: (),
+    #[serde(flatten, skip)]
+    _f: (),
+}
+
+// We assume that all configured interfaces are wanted at boot and will require a link to
+// be considered configured
+impl Default for WickedControl {
```
bcressey: This file might be easier to follow if the `impls` are near the structs they belong to.

sources/api/netdog/src/wicked.rs
```diff
@@ -0,0 +1,504 @@
...
+
+/// Allow a simple interface definition to be parsed from a string.  The expected input string
+/// looks like: `interface-name:option1,option2`.  The colon is required.  Acceptable options are
+/// "dhcp4", and "dhcp6".  For both options an additional sigil, "?", may be provided to signify
+/// that the protocol is optional.  "Optional" in this context means that we will not wait for a
+/// lease in order to consider the interface operational.
+///
+/// An full and sensible example could look like: `eno1:dhcp4,dhcp6`.  This would create an
+/// interface configuration for the interface named `eno1`, enable both dhcp4 and dhcp6, and
+/// consider a dhcp6 lease optional.
```
bcressey: Wouldn't DHCP6 be required in the case of `eno1:dhcp4,dhcp6`?

sources/api/netdog/src/main.rs
```diff
@@ -280,6 +303,93 @@ fn generate_hostname() -> Result<()> {
...
+        .primary_interface()
+        .context(error::GetPrimaryInterfaceSnafu)?;
+    write_primary_interface(primary_interface)?;
+
+    for (name, config) in net_config.interfaces {
+        if config.dhcp4.is_none() && config.dhcp6.is_none() {
+            return error::InvalidNetConfigSnafu {
+                interface: name.to_string(),
+                reason: "must configure dhcp4 or dhcp6, or both",
+            }
```
bcressey: If we don't specify either DHCP4 or DHCP6 in the wicked interface XML file, what does wicked do? Bring it up with just local addresses? Leave it down? Error out?

webern (approved):

sources/api/netdog/src/net_config.rs
zmrow: I thought about that as well.  I think it could start to get a little confusing in the case of interface names like `enp5s0`.  I like the idea of eliminating the dependency, but feel choosing the first in line is pretty clear.

sources/api/netdog/src/interface_name.rs
zmrow: You're right!  Thanks!

sources/api/netdog/src/wicked.rs
zmrow: Yep- this was a typo. Fixed

sources/api/netdog/src/wicked.rs
zmrow: You're right - removed that comment

zmrow:
Addresses all of @bcressey 's comments! 
* Centralizes all parsing (from file and command line) into the `net_config` module, which in turn simplifies the calling code in `main`.
* Enforces the same guards on valid network configuration whether it is read from file or command line, and does so in the parsing code rather than in main

sources/api/netdog/src/main.rs
```diff
@@ -10,23 +10,38 @@ It contains two subcommands meant for use as settings generators:
...
 
 The subcommand `set-hostname` sets the hostname for the system.
+
+The subcommand `generate-net-config` generates the network interface configuration for the host. If
+a `net.toml` file exists in `/var/lib/bottlerocket`, it is used to generate the configuration. If
+`net.toml` doesn't exist, the kernel command line `/proc/cmdline` is checked for the prefix
+`netdog.default-interface`.  If an interface is defined with that prefix, it is used to generate an
+interface configuration.  A single default interface may be defined on the kernel command line with
+the format: `netdog.default-interface=interface name:option1,option2`.  "interface name" is the
+name of the interface, and valid options are "dhcp4" and "dhcp6".  A "?" may be added to the option
```
bcressey: nit: to avoid ambiguity about whether spaces are allowed:
```suggestion
the format: `netdog.default-interface=interface-name:option1,option2`.  "interface-name" is the
name of the interface, and valid options are "dhcp4" and "dhcp6".  A "?" may be added to the option
```

sources/api/netdog/src/main.rs
```diff
@@ -364,6 +438,27 @@ mod error {
             output: String,
             source: serde_json::error::Error,
         },
+
+        #[snafu(display("No valid network interface config found"))]
+        InvalidNetConfig,
```
bcressey: Is this unused now?

sources/api/netdog/src/main.rs
```diff
@@ -364,6 +438,27 @@ mod error {
...
+        },
+
+        #[snafu(display("Failed to write network interface configuration: {}", source))]
+        InterfaceConfigWrite { source: wicked::Error },
+
+        #[snafu(display("Failed to write primary interface to '{}': {}", path.display(), source))]
+        PrimaryInterfaceWrite { path: PathBuf, source: io::Error },
+
+        #[snafu(display("Failed to read primary interface to '{}': {}", path.display(), source))]
+        PrimaryInterfaceRead { path: PathBuf, source: io::Error },
```
bcressey, line 455: nit: these could potentially be consolidated into errors like:
```
#[snafu(display("Failed to write {} to '{}': {}", what, path.display(), source))]
InterfaceWrite {}

#[snafu(display("Failed to read {} from '{}': {}", what, path.display(),source))]
InterfaceRead {}
```

bcressey (approved):
🥅 🐶

sources/api/netdog/src/main.rs
zmrow: Good catch - thought I got all those!

sources/api/netdog/src/main.rs
zmrow: The sources are different: `io::Error` vs `wicked::Error` :)  

In the case of `InterfaceConfigWrite` where we're writing the wicked config, we won't know the path at this layer.  The error from the wicked module will include the path.

zmrow:
^ Fixed the last couple nits!

etungsten (approved):

Issue #1105, opened 2020-09-03
Variant Aware Migrations


webern:
**What I'd like:**

#1100 adds a setting that is only relevant to the ECS variant. Looking at the code we realize this should be OK since a migration that adds a setting is actually a no-op, and removing a setting is also a no-op if the setting does not exist.

However, this raises an issue. The migration will be downloaded and executed for all variants despite being only relevant to the ECS variant.

It seems like a feature could be added to the build system and/or update system to only obtain/run migrations that are intended for the current variant (and arch?). 

**Any alternatives you've considered:**

The first thing that comes to mind is increasing the richness of the data in Release.toml and pubsys such that migrations somehow specified per-variant and only added to the manifest when relevant. Also unsure at the moment if the manifest data format could still support multiple variants (as it currently can despite the fact that we are deploying one variant per manifest currently).

Another, much less appealing possibility would be to make the update system (updog/migrator) aware of the variant-ness of migrations and skip them when not applicable.

bcressey:
For a while we dealt with this by conditionally compiling migrations based on the variant being built. Now conditional compilation of that form has been eradicated, and won't be coming back.

My hope is we'll get to settings extensions which bundle their own migration logic before this becomes relevant again. If not, we'll need a migration helper that detects the variant at runtime, e.g. by parsing `/etc/os-release`.

PR #2076, opened 2022-04-15
Restart container runtimes when certificates store changes

### LINKED ISSUE #2021: Container clients don't use the updated certificates store
**Image I'm using:**
aws-k8s-1.21 1.6.2, but this may affect versions >=1.2.0

I expect the container client (docker or containerd-cri) to use the updated certificates store, even if the store is updated after the clients start.

**What actually happened:**
The container clients cache the certificates store, and don't reload it if the store changed. 

- From a host-container, update the certificates store with the custom CA certificate
- Launch an orchestrated container, with an image that references a self-hosted registry that uses a certificate signed by the CA certificate used to update the store
- The container client will fail to pull down the image, even though the local store was updated. The container clients will succeed to pull down the images, after they are restarted through `systemctl restart <service>`.

A work around for this problem is to use a bootstrap container to update the store, since they will run before the container runtime services start.

arnaldo2792:
**Issue number:**
Fixes #2021 

**Description of changes:**

```
ca-certificates: restart container runtimes when certificates store changes

The containerd and docker daemons cache the certificates store, so any
updates to the store will be ignored by the daemons. With this, the
daemons will be restarted whenever the certificates store is updated.
```

**Testing done:**
In aws-ecs-1, aws-k8s-1.21:
- I tried to pull down an image from a local registry, which uses custom CA certificates
- I validated the pull failed before the CA certificates store was updated
- I validated the pull succeeded after the CA certificates store was updated

ECS:

```bash
# Pull image from local registry, with pki.local-registry.trusted=false, which failed as expected
bash-5.0# docker pull <ip>.us-west-2.compute.amazonaws.com:5000/alpine
Using default tag: latest
Error response from daemon: Get "https://<ip>.us-west-2.compute.amazonaws.com:5000/v2/": x509: certificate signed by unknown authority

# Update certificates store to trust custom CA certificate
bash-5.0# apiclient set pki.local-registry.trusted=true
# Successfully pull down image
bash-5.0# docker pull <ip>.us-west-2.compute.amazonaws.com:5000/alpine
Using default tag: latest
latest: Pulling from alpine
df9b9388f04a: Pull complete
Digest: sha256:a777c9c66ba177ccfea23f2a216ff6721e78a662cd17019488c417135299cd89
Status: Downloaded newer image for <ip>.us-west-2.compute.amazonaws.com:5000/alpine:latest
<ip>.us-west-2.compute.amazonaws.com:5000/alpine:latest
```

k8s:
I deployed a pod that used an image in my local registry, since k8s variants don't have docker. After I updated the certificates store (like in the ECS test), the image was successfully pulled:

```
  Normal   Pulling    43s (x3 over 79s)  kubelet            Pulling image "<ip>.us-west-2.compute.amazonaws.com:5000/alpine:latest"
  Warning  Failed     43s (x3 over 79s)  kubelet            Failed to pull image "<ip>.us-west-2.compute.amazonaws.com:5000/alpine:latest": rpc error: code = Unknown desc = failed to pull and unpack image "<ip>.us-west-2.compute.amazonaws.com:5000/alpine:latest": failed to resolve reference "<ip>.us-west-2.compute.amazonaws.com:5000/alpine:latest": failed to do request: Head "https://<ip>.us-west-2.compute.amazonaws.com:5000/v2/alpine/manifests/latest": x509: certificate signed by unknown authority
  Warning  Failed     43s (x3 over 79s)  kubelet            Error: ErrImagePull
  Normal   BackOff    30s (x3 over 78s)  kubelet            Back-off pulling image "<ip>.us-west-2.compute.amazonaws.com:5000/alpine:latest"
  Warning  Failed     30s (x3 over 78s)  kubelet            Error: ImagePullBackOff
  Normal   Pulling    20s                kubelet            Pulling image "<ip>.us-west-2.compute.amazonaws.com:5000/alpine:latest"
  Normal   Pulled     20s                kubelet            Successfully pulled image "<ip>.us-west-2.compute.amazonaws.com:5000/alpine:latest" in 73.145474ms
  Normal   Created    20s                kubelet            Created container image
  Normal   Started    20s                kubelet            Started container image

```

- [x] aws-k8s-1.22 upgrade/downgrade
- [x] aws-k8s-1.19 upgrade/downgrade
- [x] aws-k8s-1.22-nvidia upgrade/downgrade
- [x] aws-ecs-1 upgrade/downgrade
- [x] aws-ecs-1-nvidia upgrade/downgrade
- [x] vmware-k8s-1.22 upgrade/downgrade

**Terms of contribution:**

### DIFF
--- a/sources/api/migration/migrations/v1.8.0/pki-affected-services/build.rs
+++ b/sources/api/migration/migrations/v1.8.0/pki-affected-services/build.rs
@@ -1,0 +1,6 @@
+use bottlerocket_variant::Variant;
+
+fn main() {
+    let variant = Variant::from_env().unwrap();
+    variant.emit_cfgs();
+}
--- a/sources/api/migration/migrations/v1.8.0/pki-affected-services/src/main.rs
+++ b/sources/api/migration/migrations/v1.8.0/pki-affected-services/src/main.rs
@@ -1,0 +1,35 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::{
+    MetadataListReplacement, ReplaceMetadataListsMigration,
+};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We updated the 'affected-services' list metadata for 'settings.pki' to include
+/// containerd or docker on upgrade, and to remove them on downgrade depending on the
+/// running variant.
+fn run() -> Result<()> {
+    migrate(ReplaceMetadataListsMigration(vec![
+        MetadataListReplacement {
+            setting: "settings.pki",
+            metadata: "affected-services",
+            old_vals: &["pki"],
+            new_vals: if cfg!(variant_runtime = "k8s") {
+                &["pki", "containerd"]
+            } else {
+                &["pki", "docker"]
+            },
+        },
+    ]))
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/Release.toml
+++ b/Release.toml
@@ -123,4 +123,5 @@ version = "1.7.2"
     "migrate_v1.8.0_etc-hosts.lz4",
     "migrate_v1.8.0_etc-hosts-metadata.lz4",
     "migrate_v1.8.0_cluster-dns-ip-list.lz4",
+    "migrate_v1.8.0_pki-affected-services.lz4"
 ]
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -2537,6 +2537,14 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
+name = "pki-affected-services"
+version = "0.1.0"
+dependencies = [
+ "bottlerocket-variant",
+ "migration-helpers",
+]
+
+[[package]]
 name = "pluto"
 version = "0.1.0"
 dependencies = [
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -56,6 +56,7 @@ members = [
     "api/migration/migrations/v1.8.0/etc-hosts",
     "api/migration/migrations/v1.8.0/etc-hosts-metadata",
     "api/migration/migrations/v1.8.0/cluster-dns-ip-list",
+    "api/migration/migrations/v1.8.0/pki-affected-services",
 
     "bottlerocket-release",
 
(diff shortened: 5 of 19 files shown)


webern (approved):

zmrow (approved):
:headphones:

sources/models/shared-defaults/containerd-pki.toml
```diff
@@ -0,0 +1,2 @@
+[metadata.settings.pki]
+affected-services = ["pki", "containerd"]
```
bcressey: We need a migration for this (`ReplaceMetadataListsMigration`) since we're changing the list of affected services. However, the problem is that the new value depends on the variant, and migrations previously haven't been variant-specific.

You might be able to make this work with conditional compilation based on the variant, since the migration binaries are part of the same artifacts as the update images - meaning the right migration binary would end up in the right repo.

It'd need to be thoroughly tested though, since it's new territory.

sources/models/src/aws-k8s-1.19/defaults.d/53-containerd-pki.toml
```diff
@@ -0,0 +1 @@
+../../../shared-defaults/containerd-pki.toml
```
bcressey: Nit: I'd prefer "containerd-cri" over "containerd" for clarity in cases where we are doing something that's CRI specific.

bcressey (changes requested):

bcressey:
It'd be worth rebasing this to exercise more of the functionality after the `pem` crate update in #2049.

arnaldo2792:
(Rebase to pull down migrations directory for next release)

arnaldo2792:
(Rebase to pull down latest changes in `build.rs` files)

arnaldo2792:
Forced push includes migration and configuration files for newly added variants

arnaldo2792:
(Rebased to pull down fixes in aws-ecs-1-nvidia)

arnaldo2792:
(Fix since I was using `runtime`, and the new buildsys library emits `variant_runtime`)

sources/api/migration/migrations/v1.8.0/pki-affected-services/build.rs
```diff
@@ -0,0 +1,6 @@
+use bottlerocket_variant::Variant;
+
+fn main() {
+    let variant = Variant::from_env().unwrap();
+    variant.emit_cfgs();
```
webern, line 5: :rocket:

sources/models/shared-defaults/containerd-pki.toml
webern: > migrations previously haven't been variant-specific.

For the sake of bookkeeping, here is the issue for that #1105

webern (approved):

sources/api/migration/migrations/v1.8.0/pki-affected-services/src/main.rs
```diff
@@ -0,0 +1,34 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::{
+    MetadataListReplacement, ReplaceMetadataListsMigration,
+};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We updated the 'affected-services' list metadata for 'settings.network.hostname' to include the
+/// hosts "service" on upgrade, and to remove it on downgrade.
```
etungsten: This comment is not accurate.

etungsten (approved):

arnaldo2792:
(Forced push to fix comment in migration)

bcressey (approved):

arnaldo2792:
(Forced push to fix merge conflicts with `Release.toml` and `sources/Cargo.toml`)

Issue #1074, opened 2020-08-25
[ECS] GPU support


samuelkarp:
This issue is tracking adding support for GPU devices to the ECS variant of Bottlerocket.

This will require adding the Nvidia kernel module and userspace drivers.

kniec:
Hey, super interested in that issue as it related to a lot what the HPC community would want BottleRocket to use for. As I wrote in the (renamed) EKS issue #769: I would like the container to provide the userland driver and not do a bind-mount for that. 
From my HPC PoV the extreme benefit of bottlerocket in this scenario would be that a customer can identify a reproducible OS host system (hash or tag) and together with the container sha256 provide a reproducible execution environment that does not leave room for interpretation. While the tuple would allow the host to add a driver to the container, I would like to keep that seperate.

arnaldo2792:
Hey @kniec, for our kubernetes GPU support we went ahead with the general approach of mounting the userland driver in containers, since as described in the issue you linked, we didn't want customers to rethink how they have to build their container images. However, I do think that your use case could be supported in Bottlerocket, but we need to design how we actually want to do it. I'll bring your use case up to the team and discuss if we will support it in future releases.

PR #2098, opened 2022-04-23
Add GPUSupportEnabled to ECS configurations

### LINKED ISSUE #2086: [ECS] Conditionally set GPUSupportEnabled in ecs-settings-applier
We need to conditionally set the value of the `GPUSupportEnabled` flag in `ecs-settings-applier` at compilation time, depending on what type of image is being built (accelerated vs general purpose).

This is to support #1074

arnaldo2792:
**Issue number:**
Closes #2086 

**Description of changes:**

In preparation for #1074, this adds the `GPUSupportEnabled` flag as `false` for the existing ECS variant.

```
Now the value of the `GPUSupportEnabled` flag will be set at
compile-time depending on the variant being built.
```

**Testing done:**

I checked that the `ecs.config.json` file was created with the expected flag set as `false` for the general purpose variant ECS variant:

```
[root@admin]# cat /.bottlerocket/rootfs/etc/ecs/ecs.config.json  | jq
{
  "Cluster": "bottlerocket",
  "InstanceAttributes": {
    "bottlerocket.variant": "aws-ecs-1"
  },
  "PrivilegedDisabled": true,
  "AvailableLoggingDrivers": [
    "json-file",
    "awslogs",
    "none"
  ],
  "TaskIAMRoleEnabled": true,
  "TaskIAMRoleEnabledForNetworkHost": true,
  "SELinuxCapable": true,
  "OverrideAWSLogsExecutionRole": true,
  "TaskENIEnabled": true,
  "GPUSupportEnabled": false
}

```

**Terms of contribution:**

### DIFF
--- a/sources/api/ecs-settings-applier/build.rs
+++ b/sources/api/ecs-settings-applier/build.rs
@@ -9,6 +9,17 @@ fn main() {
     println!("cargo:rerun-if-env-changed=VARIANT");
     if let Ok(variant) = env::var("VARIANT") {
         println!("cargo:rustc-cfg=variant=\"{}\"", variant);
+        let parts = variant.split('-').collect::<Vec<&str>>();
+        println!(
+            "cargo:rustc-cfg=variant_family=\"{}\"",
+            parts[0..2].join("-")
+        );
+        let variant_type = if parts.len() > 3 {
+            parts[3]
+        } else {
+            "general_purpose"
+        };
+        println!("cargo:rustc-cfg=variant_type=\"{}\"", variant_type);
     }
 
     // Check for environment variable "SKIP_README". If it is set,
--- a/sources/api/ecs-settings-applier/src/ecs.rs
+++ b/sources/api/ecs-settings-applier/src/ecs.rs
@@ -10,6 +10,7 @@ embedded lists.  The structure and names of fields in the document can be found
 use constants;
 use log::debug;
 use serde::Serialize;
+use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
 use snafu::{OptionExt, ResultExt};
 use std::fs;
 use std::path::Path;
@@ -64,6 +65,7 @@ pub(crate) async fn main() -> () {
 
 async fn run() -> Result<()> {
     let args = parse_args(env::args());
+    SimpleLogger::init(LevelFilter::Info, LogConfig::default()).context(error::LoggerSnafu)?;
 
     // Get all settings values for config file templates
     debug!("Requesting settings values");
@@ -188,6 +190,11 @@ mod error {
         #[snafu(display("Failed to read settings: {}", source))]
         Settings {
             source: schnauzer::Error,
+        },
+
+        #[snafu(display("Logger setup error: {}", source))]
+        Logger {
+            source: log::SetLoggerError,
         },
 
         Model,
--- a/sources/api/ecs-settings-applier/src/ecs.rs
+++ b/sources/api/ecs-settings-applier/src/ecs.rs
@@ -51,6 +51,9 @@ struct ECSConfig {
 
     #[serde(rename = "TaskENIEnabled")]
     task_eni_enabled: bool,
+
+    #[serde(rename = "GPUSupportEnabled")]
+    gpu_support_enabled: bool,
 }
 
 // Returning a Result from main makes it print a Debug representation of the error, but with Snafu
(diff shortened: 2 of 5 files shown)


sources/api/ecs-settings-applier/src/ecs.rs
```diff
@@ -101,6 +109,8 @@ async fn run() -> Result<()> {
 
         // awsvpc mode is always available
         task_eni_enabled: true,
+
+        gpu_support_enabled: GPU_SUPPORT_ENABLED,
```
etungsten: nit: I wonder if we can make the condition more clear by using the [`cfg!`](https://doc.rust-lang.org/std/macro.cfg.html) macro here:
```suggestion
        gpu_support_enabled: !(cfg!(variant_type = "general_purpose")),
```

then we can ditch the intermediary const bool.

sources/api/ecs-settings-applier/src/ecs.rs
```diff
@@ -64,6 +71,7 @@ pub(crate) async fn main() -> () {
 
 async fn run() -> Result<()> {
     let args = parse_args(env::args());
+    SimpleLogger::init(LevelFilter::Info, LogConfig::default()).context(error::LoggerSnafu)?;
```
etungsten, line 71: I guess logging was never working prior to this change?

nit: Can we move these changes to a separate commit?

arnaldo2792: That's correct, nothing was logged out. I can put this in a different commit

sources/api/ecs-settings-applier/src/ecs.rs
arnaldo2792: I can give this a spin :+1:

arnaldo2792:
Last two forced pushes address:

- Remove constant value and use `!cfg` instead
- Separate logging change into its own commit

arnaldo2792:
(Forced push to rebase develop)

sources/api/ecs-settings-applier/build.rs
```diff
@@ -9,6 +9,17 @@ fn main() {
...
+        println!(
+            "cargo:rustc-cfg=variant_family=\"{}\"",
+            parts[0..2].join("-")
+        );
+        let variant_type = if parts.len() > 3 {
+            parts[3]
+        } else {
+            "general_purpose"
+        };
+        println!("cargo:rustc-cfg=variant_type=\"{}\"", variant_type);
```
webern, line 22: It doesn't have to be part of this PR, but we should generalize this logic. Possibly buildsys could have a lib.rs that we use for things like this.

etungsten (approved):

sources/api/ecs-settings-applier/build.rs
```diff
@@ -9,6 +9,17 @@ fn main() {
...
+        let parts = variant.split('-').collect::<Vec<&str>>();
+        println!(
+            "cargo:rustc-cfg=variant_family=\"{}\"",
+            parts[0..2].join("-")
+        );
+        let variant_type = if parts.len() > 3 {
+            parts[3]
+        } else {
+            "general_purpose"
+        };
```
bcressey, line 21: Why not check for NVIDIA specifically? Is it because we'd also set `GPUSupportEnabled` for AMD and Inferentia?

bcressey (approved):

sources/api/ecs-settings-applier/build.rs
arnaldo2792: I didn't see support for AMD in the ECS agent, but to enable inferentia support, a different flag has to be (`InferentiaSupportEnabled`) set and we will set it based on the variant's name.

bcressey: It might be better to pin it to NVIDIA if `GPUSupportEnabled` is synonymous with that hardware.

arnaldo2792: But we still need to know what "variant type" is built, since we could have `aws-k8s-*-inferentia` so that in `ecs-settings-applier` we do:

```rust
{
    GPUSupportEnabled : cfg!(variant_type = "nvidia"), 
    InferentiaSupportEnabled: cfg!(variant_type = "inferentia")
}
```

arnaldo2792:
Forced push changes the condition to set `GPUSupportEnabled`

etungsten (approved):

bcressey (approved):

webern (approved):

PR #2100, opened 2022-04-23
Add ecs-gpu-init helper program

### LINKED ISSUE #2088: [ECS] Add ecs-gpu-init to create nvidia-gpu-info.json
`ecs-gpu-init` will create the file `/var/lib/ecs/gpu/nvidia-gpu-info.json`, which is used by the ECS agent to configure GPUs for ECS tasks.

This is to support #1074

arnaldo2792:
**Issue number:**
Closes #2088 

**Description of changes:**

```
`ecs-gpu-init` queries the NVIDIA GPU devices available in the host, and
writes down the `nvidia-gpu-info.json` file used by the ECS agent to
configure GPUs for orchestrated tasks.
```

**Testing done:**
In [this branch](https://github.com/arnaldo2792/bottlerocket/tree/nvidia-ecs-variant), I checked that the configuration file was generated and that the ECS agent used it to configure a task that required GPUs:

```bash
[root@admin]# cat /.bottlerocket/rootfs/var/lib/ecs/gpu/nvidia-gpu-info.json | jq
{
  "DriverVersion": "470.82.01",
  "GPUIDs": [
    "<GPU-ID>"
  ]
}

```

**Terms of contribution:**


Makefile.toml
```diff
@@ -281,7 +281,7 @@ cargo test \
 
 # host-ctr unit tests (go)
 ${BUILDSYS_TOOLS_DIR}/docker-go \
-  --module-path ${BUILDSYS_SOURCES_DIR}/host-ctr \
```
etungsten: I know that `ecs-gpu-init` currently doesn't have any unit tests, but we can still add it here in case that it does in the future. Not sure how much overhead this would incur though? Maybe it's not worth it?

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// getGPUDeviceIDs returns the UUIDs of all the NVIDIA GPUs
+func getGPUDeviceIDs() ([]string, error) {
+	count, ret := nvml.DeviceGetCount()
+	if ret != nvml.SUCCESS {
+		return nil, errors.New("Failed to get device count")
+	}
+	var gpuIDs []string
+	var errors []error
```
etungsten: nit: 
```suggestion
	var (
		gpuIDs []string
		errors []error
        )
```

etungsten (approved):

Makefile.toml
arnaldo2792: I tested it, and I didn't see any slowdown in the execution so I'll keep it.

arnaldo2792:
(Forced push fixes comments above)

packages/ecs-gpu-init/ecs-gpu-init-tmpfiles.conf
```diff
@@ -0,0 +1 @@
+d /var/lib/ecs/gpu 0700 root root
```
bcressey: In general, for anything that touches persistent storage, I prefer the combination of:
```
R /var/lib/ecs/gpu - - - -
C /var/lib/ecs/gpu - - - -
Z /var/lib/ecs/gpu 0700 root root -
```

In other words:
* remove the directory if we don't strictly need the persistent state
* create the directory
* relabel the directory according to the active SELinux policy rules

This is somewhat more future-proof since it means that when we downgrade into an older version, we know it won't be confused by newer file formats or unexpected SELinux labels.

packages/ecs-gpu-init/ecs-gpu-init.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Initialize ECS agent GPUs config
```
bcressey: nit: maybe simplify a bit:
```suggestion
Description=Initialize ECS GPU config
```

packages/ecs-gpu-init/ecs-gpu-init.service
```diff
@@ -0,0 +1,13 @@
...
+Documentation=https://aws.amazon.com/documentation/ecs/
```
bcressey: This is probably not the best documentation URL; I'd either omit it or point it into our own repo.

packages/ecs-gpu-init/ecs-gpu-init.service
```diff
@@ -0,0 +1,13 @@
...
+Before=ecs.service
+After=configured.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ecs-gpu-init
+
+[Install]
+WantedBy=multi-user.target
+RequiredBy=ecs.service
```
bcressey: Is there any reason not to run this in `preconfigured.target` with the usual safeguards for oneshots?

tools/docker-go
```diff
@@ -75,16 +77,24 @@ for i in http_proxy https_proxy no_proxy HTTP_PROXY HTTPS_PROXY NO_PROXY ; do
...
-  --security-opt label:disable \
-  ${DOCKER_RUN_ARGS} \
-  -v "${GO_MOD_CACHE}":/tmp/go/pkg/mod \
-  -v "${GO_MODULE_PATH}":/usr/src/host-ctr \
-  -w /usr/src/host-ctr \
-  "${SDK_IMAGE}" \
-    bash -c "${COMMAND}"
+IFS=','
+for dir in ${GO_MODULES_PATHS}; do
+  package=$(basename $dir)
```
bcressey: nit: I prefer:
```suggestion
  package=${dir##*/}
```

tools/docker-go
```diff
@@ -75,16 +77,24 @@ for i in http_proxy https_proxy no_proxy HTTP_PROXY HTTPS_PROXY NO_PROXY ; do
...
+  cmd=$(echo ${COMMAND} | sed -e "s|MODULE|${package}|")
```
bcressey: I'd prefer duplication in `Makefile.toml` to this sort of magical substitution. This approach complicates reuse (c.f. #2053) since the command to run is no longer what's literally run.

One path forward might be to wrap the call to this script with a function in `Makefile.toml`.

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+const (
+	// NvidiaGPUInfoFilePath is the file path where GPUs and driver info are saved
+	NvidiaGPUInfoFilePath = "/var/lib/ecs/gpu/nvidia-gpu-info.json"
+)
+
+func main() {
+	if err := generateGPUInfoConfig(); err != nil {
+		panic(err)
```
bcressey: Is panic really the right function here?

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	gpuIDs, err := getGPUDeviceIDs()
+	if err != nil {
+		return err
+	}
+
+	if err := writeGPUInfo(version, gpuIDs); err != nil {
+		return err
+	}
+	return nil
```
bcressey: Can this be simplified? It seems like there'd be a more idiomatic Go construct, e.g.:
```suggestion
	return writeGPUInfo(version, gpuIDs)
```

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
```diff
@@ -0,0 +1,102 @@
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+)
```
bcressey: How large is this 100 LOC Go binary on disk?

bcressey (changes requested):

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
arnaldo2792: I used `hotdog` as my reference, and [we do use panic](https://github.com/bottlerocket-os/hotdog/blob/aa4ab06fff7ce33094856cc8e7df42481c6f13ef/cmd/hotdog-hotpatch/main.go#L62) there

webern: > Is panic really the right function here?

I suppose you could print the error and exit 1 instead of spewing a stack trace that points to main.

tools/docker-go
webern: > magical substitution

I tend to agree. I didn't understand `MODULE` when I saw it earlier in Makefile.toml. I was asking "is that really the directory name?"

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
arnaldo2792: Around 2M:

```bash
bash-5.1# ls -lh ecs-gpu-init
-rwxr-xr-x. 1 root root 2.1M Apr 27 18:07 ecs-gpu-init
```

arnaldo2792:
Forced push includes:
- Update tmpfilesd configuration file to remove, create and relabel the target directory
- Small refactor in `ecs-gpu-init` to be more idiomatic
- Remove for-loop in `docker-go` and passed that logic to `Makefile.toml` with wrapper functions to run `go fetch`, `go fmt` and unit tests in go packages

Makefile.toml
```diff
@@ -259,11 +260,19 @@ chmod -R o+r ${CARGO_HOME}
 dependencies = ["fetch-sdk"]
 script = [
 '''
-${BUILDSYS_TOOLS_DIR}/docker-go \
-  --module-path ${BUILDSYS_SOURCES_DIR}/host-ctr \
-  --sdk-image ${BUILDSYS_SDK_IMAGE} \
-  --go-mod-cache ${GO_MOD_CACHE} \
-  --command "go list -mod=readonly ./... >/dev/null && go mod vendor"
+function go_fetch
+{
```
bcressey: nit: elsewhere in this file we use the `go_fetch() {...}` form

Makefile.toml
```diff
@@ -259,11 +260,19 @@ chmod -R o+r ${CARGO_HOME}
...
+  local module=$1
```
bcressey: I prefer this form for expected variables: put the `local` declaration on its own line, so it doesn't mask errors, and then do the assignment and error if no argument was passed in that position.

```suggestion
  local module
  module="${1:?}"
```

Makefile.toml
```diff
@@ -259,11 +260,19 @@ chmod -R o+r ${CARGO_HOME}
...
+  ${BUILDSYS_TOOLS_DIR}/docker-go \
+    --module-path ${BUILDSYS_SOURCES_DIR}/$module \
```
bcressey: pedantic style nit:
```suggestion
    --module-path "${BUILDSYS_SOURCES_DIR}/${module}" \
```

I'd recommend just always using braces and quotes for variables to avoid needing to think about when they're optional or might be required.

packages/ecs-gpu-init/ecs-gpu-init.service
```diff
@@ -0,0 +1,12 @@
...
+Before=ecs.service
+After=configured.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/ecs-gpu-init
+
+[Install]
+WantedBy=preconfigured.target
+RequiredBy=ecs.service
```
bcressey: Hmm, why `After=configured.target` if we also have `WantedBy=preconfigured.target`? I'd expect us to finish creating the file either in preconfigured or in configured.

arnaldo2792:
Forced push includes:
- Use `go_fmt() {...}` instead of `function go_fmt {...}`
- Change `ecs-gpu-init.service` dependencies; now it is `RequiredBy=preconfigured.target` and will be executed`After=load-kerne-module.service`
- Add `RemainAfterExit` and `StandardError`, as we do in other `oneshot` systemd units.

etungsten (approved):

Makefile.toml
```diff
@@ -259,11 +260,20 @@ chmod -R o+r ${CARGO_HOME}
 dependencies = ["fetch-sdk"]
 script = [
 '''
-${BUILDSYS_TOOLS_DIR}/docker-go \
-  --module-path ${BUILDSYS_SOURCES_DIR}/host-ctr \
-  --sdk-image ${BUILDSYS_SDK_IMAGE} \
-  --go-mod-cache ${GO_MOD_CACHE} \
-  --command "go list -mod=readonly ./... >/dev/null && go mod vendor"
+go_fetch()
+{
```
bcressey: nit:
```suggestion
go_fetch() {
```

Makefile.toml
```diff
@@ -259,11 +260,20 @@ chmod -R o+r ${CARGO_HOME}
...
+  module="${1:?}"
+  ${BUILDSYS_TOOLS_DIR}/docker-go \
+    --module-path "${BUILDSYS_SOURCES_DIR}/${module}" \
+    --sdk-image ${BUILDSYS_SDK_IMAGE} \
+    --go-mod-cache ${GO_MOD_CACHE} \
+    --command "go list -mod=readonly ./... >/dev/null && go mod vendor"
+}
+
+for m in ${GO_MODULES}; do
+  go_fetch $m
```
bcressey: nit:
```suggestion
  go_fetch ${m}
```

(same for other places where we're using the $m or $module variables)

packages/ecs-gpu-init/ecs-gpu-init.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Initialize ECS GPU config
+# This has to be executed after the kernel modules are loaded
+# otherwise the userspace component of the driver will fail to
+# query the /dev devices
+After=load-kernel-modules.service
```
bcressey, line 6:
```suggestion
# This has to be executed after the kernel modules are loaded
# otherwise the userspace component of the driver will fail to
# query the /dev devices
After=load-kernel-modules.service
Requires=load-kernel-modules.service
```

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
```diff
@@ -0,0 +1,98 @@
...
+const (
+	// NvidiaGPUInfoFilePath is the file path where GPUs and driver info are saved
+	NvidiaGPUInfoFilePath = "/var/lib/ecs/gpu/nvidia-gpu-info.json"
+)
+
+func main() {
+	if err := generateGPUInfoConfig(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
```
bcressey: nit: should we print to stderr?

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
```diff
@@ -0,0 +1,98 @@
...
+// GenerateGPUInfoConfig generates the configuration used by the ECS agent
+func generateGPUInfoConfig() error {
+	if ret := nvml.Init(); ret != nvml.SUCCESS {
+		return errors.New("Failed to initialize NVML")
+	}
+	defer nvml.Shutdown()
+	version, ret := nvml.SystemGetDriverVersion()
+	if ret != nvml.SUCCESS {
+		return errors.New("Failed to get version")
+	}
```
bcressey: nit: if we logged the value of `ret` then that might help us reason about an error message

packages/ecs-gpu-init/ecs-gpu-init.service
```diff
@@ -0,0 +1,20 @@
...
+# Block manual interactions with this service. It doesn't
+# make sense to regenerate the GPU config file if the ECS
+# agent won't read it when it changes
```
bcressey, line 10: If we expect the file to change at runtime (do we?) we could arrange to restart the ecs-agent service after it changes. If we don't then the comment is a bit confusing.

bcressey (approved):

packages/ecs-gpu-init/ecs-gpu-init.service
arnaldo2792: We don't expect to change the file at runtime, but I thought it will be good to prevent users from attempting to change it. Can you elaborate on what part is confusing? :smile:

arnaldo2792:
Forced pushed includes:
- Add `ret` to error message
- Follow established convention for functions in `Makefile.toml`
- Add `Requires=load-kernel-modules.service` to `ecs-gpu-init.service`

packages/ecs-gpu-init/ecs-gpu-init.service
bcressey: The comment reads like a `TODO` to me, like we're saying there's a behavior change in the ECS agent that we want but we haven't addressed that yet.

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+const (
+	// NvidiaGPUInfoFilePath is the file path where GPUs and driver info are saved
+	NvidiaGPUInfoFilePath = "/var/lib/ecs/gpu/nvidia-gpu-info.json"
+)
+
+func main() {
+	if err := generateGPUInfoConfig(); err != nil {
+		fmt.Fprintf(os.Stderr, "Got error %v", err)
```
bcressey, line 19: nit: here you can probably just print the error as-is, with no prefix

sources/ecs-gpu-init/cmd/ecs-gpu-init/main.go
```diff
@@ -0,0 +1,97 @@
...
+		os.Exit(1)
+	}
+}
+
+// GenerateGPUInfoConfig generates the configuration used by the ECS agent
+func generateGPUInfoConfig() error {
+	if ret := nvml.Init(); ret != nvml.SUCCESS {
+		return fmt.Errorf("Failed to initialize NVML, got ret %v", ret)
```
bcressey, line 27: nit: usual approach would be something like:
```suggestion
		return fmt.Errorf("Failed to initialize NVML: %v", ret)
```

(to avoid the "got ret" phrase which is superfluous and unconventional)

bcressey (approved):
LGTM. Just a couple more nits that you can feel free to ignore.

PR #2107, opened 2022-04-27
add c7g support

lizthegrey:
**Description of changes:**
see https://github.com/awslabs/amazon-eks-ami/pull/903

c7g metadata is not supplied via the central service since it's in preview; it needs to be hardcoded, so supply the same values as for c6g.

**Terms of contribution:**

### DIFF
--- a/packages/os/eni-max-pods
+++ b/packages/os/eni-max-pods
@@ -147,6 +147,15 @@ c6i.8xlarge 234
 c6i.large 29
 c6i.metal 737
 c6i.xlarge 58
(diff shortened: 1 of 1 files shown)


jpculp (approved):
nit. It might be a good idea to remove or reword the `This file was generated at 2022-02-15T18:47:49Z` on L14 since this diverges from the one that was auto-generated, but otherwise looks good! 🚀

webern (approved):

etungsten (approved):

Issue #2045, opened 2022-03-31
Populate /etc/hosts with custom DNS mapping entries


johnsonjnpan:
**What I'd like:**
Add custom static DNS mapping entries for `/etc/hosts` through settings.

Our project requires us to support resolving certain DNS domain(s) to IP(s) even when the host is disconnected from the internet.
I wonder if BottleRocket can provide some way to set/add such DNS mappings reliably (can sustain a reboot).

E.g. By running `apiclient set settings.dns.mappings=["10.0.0.2 somehost.com"]`  to set it into `/etc/hosts`

**Any alternatives you've considered:**
* We tried adding some DNS mappings inside `/etc/hosts` through admin container. However, we found those changes are lost after we reboot the host. (Due to /etc mapped as tmpfs mount)

jpculp:
Hi @johnsonjnpan! Thanks for bringing this to our attention. We'll take a look at what our options are.

cbgbt:
As an update, I'm currently working on integration of this feature.

cbgbt:
This feature has been merged, and should be released with Bottlerocket 1.8.0.

cbgbt:
This is resolved in the latest Bottlerocket release, [v1.8.0](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.8.0).

[Check out the documentation](https://github.com/bottlerocket-os/bottlerocket#network-settings) for more details on how to use the new setting, `settings.network.hosts`. Let us know if you have any questions!

Issue #921, opened 2020-05-01
Adding local DNS cache


danmx:
**What I'd like:**

So see a local DNS cache (e.g. dnsmasq) baked into the OS image to improve DNS resolution performance.

**Any alternatives you've considered:**

Adding it to my fork of bottlerocket but it seems like a such generic issue it could be addressed in the source.

zmrow:
Hi @danmx! 

Could you provide more details about your setup? Would you be wanting the DNS cache for the host or for the containers?

danmx:
My team is looking into setting DNS cache for both OS and containers.
The setup should look like this: dnsmasq (running locally) -> NodeLocal (running locally) -> CoreDNS (running remotely)

chancez:
You can run a local DNS cache in a container and configure your containers to use it pretty easily. One example for running this type of setup on Kubernetes is https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/.

WilboMo:
Hello! I wanted to check in and see if having DNS cache in containers was sufficient for you or if having DNS cache on the host ended up being a requirement?

danmx:
In the end I moved to solutions like Istio's DNS sidecar or Cilium's DNS Proxy

WilboMo:
Thanks for the update, I'm glad to hear you found a solution that works for you. We'll keep this issue open for consideration in future releases.

ami-descope:
Does 1.28 release added something of DNS or resolver running on the same port 53?

i am running local-node-dns on the cluster and 1.27 was fine but with  1.28  i get

` Listen: listen tcp 0.0.0.0:53: bind: address already in use`

bcressey:
`systemd-resolved` was added to variants using `systemd-networkd`, starting with `aws-k8s-1.28`.

PR #2129, opened 2022-05-04
Add support for static DNS mappings

cbgbt:
**Issue number:** #2045 

**Description of changes:**
* [x] Implement models and settings changes
* [x] Write migrations for new settings

```
This allows users to populate /etc/hosts with static DNS entries via
settings and the Bottlerocket API. Bottlerocket doesn't currently
perform resolver caching, so there is no need to restart networked
services after rendering a new /etc/hosts file.
```

According to `man hosts:`

>Before the advent of DNS, the host table was the only way of resolving hostnames on the fledgling Internet.

Today, this file is still used to configure a static lookup table for hostnames on Linux. The hosts file is a text file that associates IP addresses (v4 or v6) with hostnames, one line per IP address.

```
IP_address canonical_hostname [aliases...]
```

e.g.

```
10.0.0.1 myhost.example.com myhost
```

Would ensure that `myhost.example.com` and `myhost` both resolve to `10.0.0.1`. The man page explicitly calls out the first element as being the `canonical_hostname`, but there are no technical requirements that differentiate this to aliases — each element must:

* Contain only alphanumeric characters, minus signs (“-”) and periods (“.”)
* Begin with an alphabetic character
* End with an alphanumeric character

The modeled settings enforce these requirements.

Per #921, we don't support DNS caching on Bottlerocket, so we shouldn't need to restart any additional services when we modify static DNS mappings.

**Testing done:**
* [x] Test on aws-ecs-1
* [x] Test on aws-k8s-1.22
* [x] Test migrations

Example of using the feature:
```
bash-5.1# cat /etc/hosts


bash-5.1# apiclient set -j '{"settings": {"network": {"hosts": [["10.0.0.1", ["test.example.com"]], ["127.0.0.1", ["test.bottlerocket.aws"]], ["10.0.0.1", ["test2.example.com"]]]}}}'
bash-5.1# apiclient get settings.network.hosts
{
  "settings": {
    "network": {
      "hosts": [
        [
          "10.0.0.1",
          [
            "test.example.com"
          ]
        ],
        [
          "127.0.0.1",
          [
            "test.bottlerocket.aws"
          ]
        ],
        [
          "10.0.0.1",
          [
            "test2.example.com"
          ]
        ]
      ]
    }
  }
}
bash-5.1# cat /etc/hosts
127.0.0.1 localhost localhost.localdomain localhost4 localhost4.localdomain4 test.bottlerocket.aws

10.0.0.1 test.example.com test2.example.com
```

**Terms of contribution:**


zmrow:
LGTM so far!  Once the migrations are out I will give it another look-see

cbgbt:
* Added README documentation
* Use `BTreeMap` to store host entries to preserve ordering

cbgbt:
I'm going to make a few changes to the `/etc/hosts` template, documentation, and settings name based on some verbal feedback from the team. Should have revision up shortly :smile:

cbgbt:
A few changes:
* Setting name changed to `settings.network.hosts`.
* Modified `/etc/hosts` such that it can't contain duplicate line entries for host addresses

cbgbt:
I keep ending up having strange conflicts with develop. Will fix the git tree.

cbgbt:
Fixed, as well as fixed some code clarity issues.

README.md
```diff
@@ -502,11 +502,27 @@ In addition to the container runtime daemons, these credential settings will als
 
 * `settings.network.hostname`: The desired hostname of the system.
   **Important note for all Kubernetes variants:** Changing this setting at runtime (not via user data) can cause issues with kubelet registration, as hostname is closely tied to the identity of the system for both registration and certificates/authorization purposes.
+* `settings.network.hosts`: A mapping of IP addresses to domain names which should resolve to those IP addresses.
+   This setting results in modifications to the `/etc/hosts` file  for Bottlerocket.
+   Note that this setting does not typically impact name resolution for containers, which usually rely on [orchestrator-specific](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HostEntry.html) [mechanisms](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/) for configuring static resolution.
```
webern: Just a suggestion...

```suggestion
   Note that this setting does not typically impact name resolution for containers, which usually rely on orchestrator-specific mechanisms for configuring static resolution.
   (See [ECS](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HostEntry.html) and [Kubernetes](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/) documentation for those mechanisms.)
```

README.md
```diff
@@ -502,11 +502,27 @@ In addition to the container runtime daemons, these credential settings will als
...
+   ```
+   Entries which point to the loopback address are merged as aliases with the lowest precedence on the existing loopback line.
+
+
 
 Most users don't need to change this setting as the following defaults work for the majority of use cases.
 If this setting isn't set we attempt to use DNS reverse lookup for the hostname.
 If the lookup is unsuccessful, the IP of the node is used.
-
 ##### Proxy settings
```
webern: nit: newline crimes were committed

```
Entries which point to the loopback address are merged as aliases with the lowest precedence on the existing loopback line.

Most users don't need to change this setting as the following defaults work for the majority of use cases.
If this setting isn't set we attempt to use DNS reverse lookup for the hostname.
If the lookup is unsuccessful, the IP of the node is used.

##### Proxy settings
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -969,77 +969,252 @@ pub fn kube_reserve_cpu(
...
     out: &mut dyn Output,
 ) -> Result<(), RenderError> {
     // To give context to our errors, get the template name, if available.
-    trace!("Starting base64_decode helper");
+    trace!("Starting localhost_aliases helper");
     let template_name = template_name(renderctx);
     trace!("Template name: {}", &template_name);
 
-    // Check number of parameters, must be exactly one
+    // Check number of parameters, must be exactly two (IP version, hostname, hosts overrides)
```
webern:
```suggestion
    // Check number of parameters, must be exactly three (IP version, hostname, hosts overrides)
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -969,77 +969,252 @@ pub fn kube_reserve_cpu(
...
-            msg: "Found no params after confirming there is one param",
+            msg: "Missing params after confirming that they exist.",
+        })?;
+    trace!("IP version value from template: {}", ip_version_value);
+
+    let hostname_value = helper
+        .param(1)
+        .map(|v| v.value())
+        .context(error::InternalSnafu {
+            msg: "Missing params after confirming that they exist.",
```
webern: Maybe since we use this same message all over we could add ParamUnwrapSnafu variant with this message baked in and eliminate the strings from the code.

```
            msg: "Missing params after confirming that they exist.",
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -969,77 +969,252 @@ pub fn kube_reserve_cpu(
...
+
+        // If any static mappings in `settings.network.hosts` are for localhost, add them as well.
+        hosts
+            .iter()
+            .find(|(ip_address, _)| {
+                if let Ok(ip_address) = ip_address.parse::<IpAddr>() {
+                    ip_address == localhost_comparator
+                } else {
+                    false
+                }
```
webern: Golfable if desired

```suggestion
                ip_address.parse::<IpAddr>().map(|ip_address|  ip_address == localhost_comparator).unwrap_or_default()
```

sources/models/src/lib.rs
```diff
@@ -270,6 +270,8 @@ struct HostContainer {
 #[model]
 struct NetworkSettings {
     hostname: ValidLinuxHostname,
+    // Ordering matters in /etc/hosts if multiple domains point to the same IP, so we use BTreeMap to preserve order.
+    hosts: BTreeMap<IpAddr, Vec<ValidLinuxHostname>>,
     https_proxy: Url,
```
cbgbt: I've been doing some extra reading about `BTreeMap` and my assumption that it preserves iteration order based on insertion is false: BTreeMap preserves iteration order based on key order. The common solution in the Rust community is to use [IndexMap](https://crates.io/crates/indexmap) for a map which preserves insertion order.


This also means that I would need to assert that every serializer and deserializer which touches settings objects respects ordering. Fortunately for `serde_json` and `toml` this can be enabled by adding a feature flag, but we have a few points e.g. in `migrations` where helpers assume that a `Map` can be safely represented via a HashMap.


I'm looking into solving this either by building in order preservation to settings (which could be extremely difficult to continuously enforce in the Bottlerocket repo, even with great testing in the core settings path) or by changing our model here to use an ordered data structure, although this adds some uncomfortable ergonomic questions over how to use the feature.

cbgbt:
I reworked the static DNS feature to use a `Vec<(IpAddr, Vec<ValidLinuxHostname>)>` as its internal representation, then added a helper method to merge IP address lines together (though this doesn't dedupe hostnames).

While it does make it a bit less ergonomic to specify mappings due to the use of lists to represent pairs, it safely models the behavior of `/etc/hosts`.

cbgbt:
Force-pushed to rebase atop develop.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -969,77 +983,199 @@ pub fn kube_reserve_cpu(
...
+        })?;
+    trace!("Hostname string from template: {}", hostname);
+
+    let mut results: Vec<String> = vec![];
 
-        out.write(&entries).context(error::TemplateWriteSnafu {
+    // If our hostname isn't resolveable, add it to the alias list.
+    if hostname.len() > 0 && !hostname_resolveable(hostname) {
+        results.push(hostname.to_owned());
+    }
```
bcressey: I think there's an edge case here where the hostname isn't initially resolvable, but should resolve to a non-localhost IP provided in `settings.network.hosts`, after `/etc/hosts` is populated.

In that case, we'd incorrectly write `/etc/hosts` with the hostname as an alias for localhost.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -969,77 +983,199 @@ pub fn kube_reserve_cpu(
...
+            value: hosts_value.to_owned(),
+            template: template_name.to_owned(),
+        })?;
+        trace!("Hosts from template: {:?}", hosts);
+
+        hosts
+            .iter_merged()
+            .filter(|(ip_address, _)| {
+                // Cast aside anything that doesn't parse as an IP Address, or any references to localhost.
+                *ip_address != IPV4_LOCALHOST && *ip_address != IPV6_LOCALHOST
```
bcressey: Where are we casting aside anything that doesn't parse as an IP address?

sources/api/schnauzer/src/helpers.rs
cbgbt: Nice catch! I'll fix that and see if I can't add a unit test to cover that edge case.

sources/api/schnauzer/src/helpers.rs
cbgbt: Whoops, holdover comment. This used to be much more "stringly typed" but I ended up routing all of the parsing through the `models` -- much cleaner now.

cbgbt:
Force-pushed to address comments by @bcressey.

Added a unit test for the edge case. Additional testing of that change: https://gist.github.com/cbgbt/d0e7789fe3a9a12e2b59e59b6f2f0278

bcressey:
LGTM once `cargo fmt` is happy.

cbgbt:
Force-push to fix `cargo-fmt` issue.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -969,77 +983,202 @@ pub fn kube_reserve_cpu(
...
     out: &mut dyn Output,
 ) -> Result<(), RenderError> {
     // To give context to our errors, get the template name, if available.
-    trace!("Starting base64_decode helper");
+    trace!("Starting localhost_aliases helper");
     let template_name = template_name(renderctx);
     trace!("Template name: {}", &template_name);
 
-    // Check number of parameters, must be exactly one
+    // Check number of parameters, must be exactly two (IP version, hostname, hosts overrides)
```
arnaldo2792: nit: should the number be three?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -969,77 +983,202 @@ pub fn kube_reserve_cpu(
...
+    out: &mut dyn Output,
+) -> Result<(), RenderError> {
+    // To give context to our errors, get the template name, if available.
+    trace!("Starting etc_hosts_entries helper");
+    let template_name = template_name(renderctx);
+    trace!("Template name: {}", &template_name);
+
+    // Check number of parameters, must be exactly two (IP version, hostname, hosts overrides)
+    trace!("Number of params: {}", helper.params().len());
+    check_param_count(helper, template_name, 1)?;
```
arnaldo2792: nit: another inconsistent count?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1128,6 +1267,41 @@ fn kube_cpu_helper(num_cores: usize) -> Result<String, TemplateHelperError> {
     ))
 }
 
+/// Returns whether or not a hostname resolves to a non-loopback IP address.
+///
+/// If `configured_hosts` is Some, the hostname will be considered resolvable if it is listed as an alias for any given IP address.
```
arnaldo2792: nit:

```suggestion
/// If `configured_hosts` is set, the hostname will be considered resolvable if it is listed as an alias for any given IP address.
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1854,35 +2028,109 @@ mod test_add_unresolvable_hostname {
...
+            &json!({"hostname": "localhost"}),
+        )
+        .unwrap();
+        assert_eq!(result, "localhost")
+    }
+
+    #[test]
+    fn hostname_resolves_to_static_mapping() {
+        // Given a configured hostname that does not resolve in DNS
+        // and an /etc/hosts configuration which points that contains that hostname as an alias to an IP address,
```
arnaldo2792: Nit: I think there are extra words here? Should the line be:
```suggestion
        // and an /etc/hosts configuration that contains that hostname as an alias to an IP address,
```

sources/models/src/modeled_types/shared.rs
```diff
@@ -230,6 +231,149 @@ mod test_valid_linux_hostname {
...
+}
+
+#[cfg(test)]
+mod test_etc_hosts_entries {
+    use super::{EtcHostsEntries, ValidLinuxHostname};
+    use std::net::IpAddr;
+
+    #[test]
+    fn test_valid_etc_hosts_entries() {
+        assert!(serde_json::from_str::<EtcHostsEntries>(
```
arnaldo2792, line 266: why can't all these tests be part of the same assert? Same question similar tests below.

README.md
```diff
@@ -508,9 +508,43 @@ In addition to the container runtime daemons, these credential settings will als
...
+    ["10.0.0.0", ["test.example.com", "test1.example.com"]],
+    ["10.1.1.1", ["test2.example.com"]],
+    ["10.1.1.0", ["test3.example.com"]],
+   ]
+   ```
+   Would result in `/etc/hosts` entries like so:
+   ```
+   10.0.0.0 test.example.com test1.example.com test3.example.com
+   10.1.1.1 test2.example.com
+   ```
```
fgutmann: Hi I'm trying to validate if the current logic will work for our use-case.

1. Why in the example test3.example.com (`10.1.1.0`) is merged with the `10.0.0.0` line and thus effectively rewritten to `10.0.0.0`?

2. Is it possible with the current logic to have one hostname to resolve to multiple IPs? The input syntax seems to support it. Just want to make sure no de-duplication will happen in this case. We will require a final `/etc/hosts` file that looks something like this:

```
10.0.2.252 test1.example.com
10.0.2.236 test1.example.com
10.0.3.205 test1.example.com
```

cbgbt: Hi, thanks for taking a look at this PR.

1. This is a documentation mistake, I meant for `10.1.1.0` to be a repeat of `10.0.0.0`. I'll fix that.
2. Yes, you can have one hostname resolve to multiple IPs. De-duplication will only occur for IPs, not hostnames. If you submit something like the following, you should end up with `/etc/hosts` entries like the ones you've given:

```
hosts = [
    ["10.0.2.252", ["test1.example.com"]],
    ["10.0.2.236", ["test1.example.com"]],
    ["10.0.3.205", ["test1.example.com]]
]
```

sources/models/src/modeled_types/shared.rs
cbgbt: I've tried to logically separate different checks into multiple assert lines here, so if one fails it's easier to reason about what behavior the check was testing for.

cbgbt:
Force push to address comments from @arnaldo2792

cbgbt:
Additional push to address README issue pointed out by @fgutmann. Thanks!

arnaldo2792 (approved):

sources/api/schnauzer/src/helpers.rs
```diff
@@ -969,77 +983,202 @@ pub fn kube_reserve_cpu(
...
+        .map(|v| v.value())
+        .context(error::InternalSnafu {
+            msg: "Missing params after confirming that they exist.",
+        })?;
+    trace!("Hosts value from template: {}", hosts_value);
+
+    if hosts_value.is_null() {
+        // If hosts aren't set, just exit.
+        Ok(())
+    } else {
```
webern: It's a stylistic preference, but I would return here and unindent the "else" code.

```suggestion
    if hosts_value.is_null() {
        // If hosts aren't set, just exit.
       return  Ok(())
    }
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -969,77 +983,202 @@ pub fn kube_reserve_cpu(
...
+                    value: hosts_value.to_owned(),
+                    template: template_name.to_owned(),
+                },
+            )
+        })
+        .transpose()?;
+    trace!("Hosts from template: {:?}", hosts);
+
+    // If our hostname isn't resolveable, add it to the alias list.
+    if hostname.len() > 0 && !hostname_resolveable(hostname, hosts.as_ref()) {
```
webern, line 1078: I was trying to think of a way to remove this impurity from this otherwise testable (and complex) function. Alas, I don't see a way to inject this network dependency.

Edit, well... as usual I did think of a way:

`localhost_aliases` could be a thin wrapper for `localhost_aliases_impl`.

`localhost_aliases_impl` takes a functor matching the signature of `hostname_resolvable`, then `localhost_aliases` becomes:

```rust
pub fn localhost_aliases(
    helper: &Helper<'_, '_>,
    h: &Handlebars,
    c: &Context,
    renderctx: &mut RenderContext<'_, '_>,
    out: &mut dyn Output,
) -> Result<(), RenderError> {
    localhost_aliases_impl(helper, h, c, renderctx, out, &hostname_resolvabe)
}
```

Then I think you could write tests for `localhost_aliases_impl` that takes a mock of the `hostname_resolvable` function.

sources/models/src/modeled_types/shared.rs
```diff
@@ -230,6 +231,149 @@ mod test_valid_linux_hostname {
...
+        let mut merged = indexmap::IndexMap::with_capacity(self.0.len());
+
+        for (ip_address, aliases) in &self.0 {
+            merged
+                .entry(ip_address.clone())
+                .or_insert(vec![])
+                .append(&mut (aliases.clone()));
+        }
+
+        merged.into_iter()
```
webern, line 255: :exploding_head: 

Nice rust.

webern (approved):
I'm cursed with the ability to always see a way to unit test. Otherwise this looks great.

sources/api/schnauzer/src/helpers.rs
cbgbt: I love function purity. Thanks for the tip, I'll implement this.

README.md
```diff
@@ -508,9 +508,43 @@ In addition to the container runtime daemons, these credential settings will als
...
   **Important note for all Kubernetes variants:** Changing this setting at runtime (not via user data) can cause issues with kubelet registration, as hostname is closely tied to the identity of the system for both registration and certificates/authorization purposes.
 
-Most users don't need to change this setting as the following defaults work for the majority of use cases.
-If this setting isn't set we attempt to use DNS reverse lookup for the hostname.
-If the lookup is unsuccessful, the IP of the node is used.
+  Most users don't need to change this setting as the following defaults work for the majority of use cases.
+  If this setting isn't set we attempt to use DNS reverse lookup for the hostname.
+  If the lookup is unsuccessful, the IP of the node is used.
+
+* `settings.network.hosts`: A mapping of IP addresses to domain names which should resolve to those IP addresses.
```
etungsten, line 515: Should we explain what the default is when this setting is not set? Or maybe it's obvious enough, and it would be too verbose.

sources/api/schnauzer/src/helpers.rs
cbgbt: I toyed around with implementing something like this, but it ended up making it more challenging to understand and didn't necessarily improve testability, since there are a few tests here already which just use more apparently resolve-able or un-resolve-able hostnames.

cbgbt:
Force push to make some stylistic and comment changes suggested by @webern

webern (approved):

etungsten (approved):

cbgbt:
Force push to rebase.

README.md
cbgbt: I think it might be obvious enough to omit. I noticed that we don't always include the default value, for settings as well.

PR #2133, opened 2022-05-11
kernel: add support for MegaRAID SAS

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
    kernel-5.10: add kernel config to enable support for RAID
    
    This adds kernel configs to enable support for MegaRAID SAS RAID
    controllers.

```

**Testing done:**
I was able to boot Bottlerocket from RAID0 on a baremetal server.

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.10/config-bottlerocket
+++ b/packages/kernel-5.10/config-bottlerocket
@@ -123,3 +123,6 @@ CONFIG_BOOT_CONFIG=y
 
 # Enables support for checkpoint/restore
 CONFIG_CHECKPOINT_RESTORE=y
+
+# Enables support for LSI Logic's SAS based RAID controllers
+CONFIG_MEGARAID_SAS=y
--- a/packages/kernel-5.4/config-bottlerocket
+++ b/packages/kernel-5.4/config-bottlerocket
@@ -85,3 +85,6 @@ CONFIG_MOUSE_PS2=m
 
 # Enables support for checkpoint/restore
 CONFIG_CHECKPOINT_RESTORE=y
+
+# Enables support for LSI Logic's SAS based RAID controllers
+CONFIG_MEGARAID_SAS=y


bcressey:
I'd tend to also want to change 5.4 as well, even though 5.4 is on its way out, just to minimize the delta.

Can you add a diff of the final config for x86_64 and aarch64? Just to give a complete picture of what's changed.

etungsten:
> Can you add a diff of the final config for x86_64 and aarch64? Just to give a complete picture of what's changed.

```bash
$ git diff --no-index aarch64_without aarch64_with_new_conf
diff --git a/aarch64_without b/aarch64_with_new_conf
index fa8ddb4a..49173034 100644
--- a/aarch64_without
+++ b/aarch64_with_new_conf
@@ -2251,7 +2251,7 @@ CONFIG_SCSI_MVUMI=m
 # CONFIG_SCSI_ESAS2R is not set
 # CONFIG_MEGARAID_NEWGEN is not set
 # CONFIG_MEGARAID_LEGACY is not set
-CONFIG_MEGARAID_SAS=m
+CONFIG_MEGARAID_SAS=y
 CONFIG_SCSI_MPT3SAS=m
 CONFIG_SCSI_MPT2SAS_MAX_SGE=128
 CONFIG_SCSI_MPT3SAS_MAX_SGE=128


$ git diff --no-index x86_without x86_with_new_conf
diff --git a/x86_without b/x86_with_new_conf
index fa7b6b1b..b39e1785 100644
--- a/x86_without
+++ b/x86_with_new_conf
@@ -2081,7 +2081,7 @@ CONFIG_ISCSI_BOOT_SYSFS=m
 # CONFIG_SCSI_ESAS2R is not set
 # CONFIG_MEGARAID_NEWGEN is not set
 # CONFIG_MEGARAID_LEGACY is not set
-# CONFIG_MEGARAID_SAS is not set
+CONFIG_MEGARAID_SAS=y
 CONFIG_SCSI_MPT3SAS=m
 CONFIG_SCSI_MPT2SAS_MAX_SGE=128
 CONFIG_SCSI_MPT3SAS_MAX_SGE=128
```

Interesting how `CONFIG_SCSI=y` isn't part of the diff. I guess it's `=y` by default?

etungsten:
Push above removes `CONFIG_SCSI=y` since it's already `=y` by default in the kernel config.

zmrow:
nit: Could we change the commit message to say something like `Add Megaraid SAS modules`?  There are other RAID controllers and modules so saying we added support for RAID isn't pedantically accurate. :)

bcressey (approved):

jpculp (approved):

foersleo (approved):
With enabling more hardware through including drivers like this one we might want to put all of these into a separate file like `config-bottlerocket-hardware` that lives co-located to the current config and then have that additional config added to the `merge-config` step in the [spec file](https://github.com/bottlerocket-os/bottlerocket/blob/6a82bc56b203ab27c78d5d043c3b4f4d5f1a68c8/packages/kernel-5.10/kernel-5.10.spec#L91-L96).

That would enable us in the future to easily make this configurable for potential separate variants with and without that hardware support (similar to the conditional addition of config-microcode on x86).

But we can also move these things around at a later point in a separate PR to not keep this one open for too long. If that makes sense to folks I can take care of that separate PR. Let me know what you think.

zmrow:
> But we can also move these things around at a later point in a separate PR to not keep this one open for too long. If that makes sense to folks I can take care of that separate PR. Let me know what you think.

Sounds like a perfect idea!

zmrow (approved):

markusboehme (approved):

Issue #1419, opened 2021-03-25
cargo make test


gregdek:
@webern wrote in April 2022:

We have created a system for testing Bottlerocket images over at https://github.com/bottlerocket-os/bottlerocket-test-system/
We want to integrate that system here. The ultimate goal is to run `cargo make test` after `cargo make build`, `cargo make ami`, and `cargo make repo` (and I guess we need something for pushing the repo), end then the `bottlerocket-test-system` will run integration tests (e.g. sonobuoy for k8s, upgrade/downgrade, etc).

- [x] #2147
- [x] #2148
- [x] #2149
- [x] #2150
- [x] #2151
- [x] #2152

@gregdek wrote in March 2021: Currently we use an internal testing system for validating Bottlerocket instances. We would like to run more publicly, and need to design a suitable replacement.  This is a large effort and may spawn several related issues.

ecpullen:
All listed testing workflows are now supported. Closing this issue!

Issue #1261, opened 2021-01-05
Variant Crossgrades


webern:
To upgrade from one version of Kubernetes to the next requires using a different variant of Bottlerocket. We should add first-class support for changing an instance's variant through the update system.

There are several issues to consider, here are some of them:
- How should an instance know that it is eligible for crossgrade?
- How should instances be prevented from crossgrading to an incompatible variant?
- How should the information of available crossgrades be presented (e.g. via updog and the settings api)?
- How will we migrate the datastore if the model is different between the two variants?

Issue #1260 discusses whether variants should be more composable and might have some bearing on variant crossgrades, but it is not clear whether these two issues are interdependent.

webern:
There are no thumbs-ups on this after a few years and I don't think it's something we are planning to do. I'm going to close this as "wontfix" to keep our backlog more meaningful.

webern:
I'm re-opening this due to a request that came in from the brupop project.

https://github.com/bottlerocket-os/bottlerocket-update-operator/issues/465

webern:
Actually after re-reading, it's not clear if the feature is being requested. I'll keep this closed for now and let the inquirer guide us.

PR #2134, opened 2022-05-12
buildsys: variant and readme standardization

webern:
**Issue number:**

Related to #1260
Also will be used for #1419
Also see https://github.com/bottlerocket-os/bottlerocket/pull/2098#discussion_r859041469

**Description of changes:**

```
generate readme

We had readme generating code pasted into most of our build.rs file.
This has now been moved to a public function in a library so that the
code is not duplicated.
```

```
standardize variant concepts

We had several crates that interpreted the VARIANT env variable in
different ways in build.rs. This standardizes the treatment of the
VARIANT variable by centralizing it in a library.
```

**Ain't nobody got time for that**

- First look at `tools/buildsys/src/public/readme.rs`
- Then look at, e.g. `sources/shimpei/build.rs`: simple, we've just refactored away the duplicative readme generation code.
- Next look at `tools/buildsys/src/public/variant.rs`
- Then look at, e.g. `sources/api/pluto/build.rs`: we have factored out the emitting of `cfg` variables in the build.rs

That's basically it. Most of the crates are copy-paste nothing interesting. But these crates are the ones that actually have `cfg` usage:
- `sources/api/early-boot-config/build.rs`
- `sources/api/ecs-settings-applier/build.rs`
- `sources/api/pluto/build.rs`
- `sources/api/static-pods/build.rs`
- `sources/api/storewolf/build.rs`
- `sources/logdog/build.rs`
- `sources/models/build.rs`

**Testing done:**

- [x] I changed a `README.tpl` file and observed that `README.md` was changed.
- [x] I changed a rust doc comment and observed that `README.md` was changed.
- [x] I verified that SKIP_README skips readme generation.
- [x] run an aws-k8s sonobuoy 'quick'
- [x] run an ecs task
- [x] run a vmware sonobuoy quick

**Terms of contribution:**

### DIFF
--- a/sources/api/bootstrap-containers/build.rs
+++ b/sources/api/bootstrap-containers/build.rs
@@ -1,32 +1,3 @@
-// Automatically generate README.md from rustdoc.
-
-use std::env;
-use std::fs::File;
-use std::io::Write;
-use std::path::PathBuf;
-
 fn main() {
-    // Check for environment variable "SKIP_README". If it is set,
-    // skip README generation
-    if env::var_os("SKIP_README").is_some() {
-        return;
-    }
-
-    let mut source = File::open("src/main.rs").unwrap();
-    let mut template = File::open("README.tpl").unwrap();
-
-    let content = cargo_readme::generate_readme(
-        &PathBuf::from("."), // root
-        &mut source,         // source
-        Some(&mut template), // template
-        // The "add x" arguments don't apply when using a template.
-        true,  // add title
-        false, // add badges
-        false, // add license
-        true,  // indent headings
-    )
-    .unwrap();
-
-    let mut readme = File::create("README.md").unwrap();
-    readme.write_all(content.as_bytes()).unwrap();
+    generate_readme::from_main().unwrap();
 }
--- a/sources/api/certdog/build.rs
+++ b/sources/api/certdog/build.rs
@@ -1,32 +1,3 @@
-// Automatically generate README.md from rustdoc.
-
-use std::env;
-use std::fs::File;
-use std::io::Write;
-use std::path::PathBuf;
-
 fn main() {
-    // Check for environment variable "SKIP_README". If it is set,
-    // skip README generation
-    if env::var_os("SKIP_README").is_some() {
-        return;
-    }
-
-    let mut source = File::open("src/main.rs").unwrap();
-    let mut template = File::open("README.tpl").unwrap();
-
-    let content = cargo_readme::generate_readme(
-        &PathBuf::from("."), // root
-        &mut source,         // source
-        Some(&mut template), // template
-        // The "add x" arguments don't apply when using a template.
-        true,  // add title
-        false, // add badges
-        false, // add license
-        true,  // indent headings
-    )
-    .unwrap();
-
-    let mut readme = File::create("README.md").unwrap();
-    readme.write_all(content.as_bytes()).unwrap();
+    generate_readme::from_main().unwrap();
 }
--- a/sources/api/ecs-settings-applier/build.rs
+++ b/sources/api/ecs-settings-applier/build.rs
@@ -1,9 +1,6 @@
 // Automatically generate README.md from rustdoc.
 
 use std::env;
-use std::fs::File;
-use std::io::Write;
-use std::path::PathBuf;
 
 fn main() {
     println!("cargo:rerun-if-env-changed=VARIANT");
@@ -22,27 +19,5 @@ fn main() {
         println!("cargo:rustc-cfg=variant_type=\"{}\"", variant_type);
     }
 
-    // Check for environment variable "SKIP_README". If it is set,
-    // skip README generation
-    if env::var_os("SKIP_README").is_some() {
-        return;
-    }
-
-    let mut source = File::open("src/ecs.rs").unwrap();
-    let mut template = File::open("README.tpl").unwrap();
-
-    let content = cargo_readme::generate_readme(
-        &PathBuf::from("."), // root
-        &mut source,         // source
-        Some(&mut template), // template
-        // The "add x" arguments don't apply when using a template.
-        true,  // add title
-        false, // add badges
-        false, // add license
-        true,  // indent headings
-    )
-    .unwrap();
-
-    let mut readme = File::create("README.md").unwrap();
-    readme.write_all(content.as_bytes()).unwrap();
+    generate_readme::from_file("src/ecs.rs").unwrap();
 }
--- a/sources/api/ecs-settings-applier/build.rs
+++ b/sources/api/ecs-settings-applier/build.rs
@@ -1,23 +1,7 @@
-// Automatically generate README.md from rustdoc.
-
-use std::env;
+use bottlerocket_variant::Variant;
 
 fn main() {
-    println!("cargo:rerun-if-env-changed=VARIANT");
-    if let Ok(variant) = env::var("VARIANT") {
-        println!("cargo:rustc-cfg=variant=\"{}\"", variant);
-        let parts = variant.split('-').collect::<Vec<&str>>();
-        println!(
-            "cargo:rustc-cfg=variant_family=\"{}\"",
-            parts[0..2].join("-")
-        );
-        let variant_type = if parts.len() > 3 {
-            parts[3]
-        } else {
-            "general_purpose"
-        };
-        println!("cargo:rustc-cfg=variant_type=\"{}\"", variant_type);
-    }
-
+    let variant = Variant::from_env().unwrap();
+    variant.emit_cfgs();
     generate_readme::from_file("src/ecs.rs").unwrap();
 }
--- a/sources/bottlerocket-variant/src/lib.rs
+++ b/sources/bottlerocket-variant/src/lib.rs
@@ -1,0 +1,442 @@
+/*!
+This library provides a structure for representing a Bottlerocket variant as well as functionality
+useful in build scripts and other tooling that is variant-aware.
+*/
+
+use error::Error;
+use serde::de::Error as SerdeError;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use snafu::{ensure, OptionExt, ResultExt};
+use std::borrow::Borrow;
+use std::convert::TryFrom;
+use std::fmt::{Display, Formatter};
+use std::ops::Deref;
+use std::str::FromStr;
+
+/// The name of the environment variable that tells us the current variant. Variant-sensitive crates
+/// will need to be rebuilt if this changes. `Makefile.toml` emits the variant string in the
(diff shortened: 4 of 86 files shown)


webern:
^ fix a mistake in pluto's build.rs and improve a doc comment

webern:
^ fix a couple of mistakes

webern:
^ rebase with fixed conflicts

webern:
^ improve some doc comments

webern:
:cricket: :cricket: :cricket: 

@zmrow @cbgbt can someone get the ball rolling on this PR? I'll try to do VMWare testing today.

webern:
^ rebase

sources/api/ecs-settings-applier/build.rs
```diff
@@ -1,26 +1,15 @@
...
-        );
-        let variant_type = if parts.len() > 3 {
-            parts[3]
-        } else {
-            "general_purpose"
-        };
-        println!("cargo:rustc-cfg=variant_type=\"{}\"", variant_type);
-    }
+    // This build has variant-specific logic.
+    let variant = Variant::from_env().unwrap();
```
zmrow, line 4: This `unwrap()` made me a little sad initially, but on further thought I think it might be fine since it's during the build and should print something intelligible.

zmrow (approved):
lgtm

:nut_and_bolt:

sources/api/ecs-settings-applier/build.rs
webern: > This unwrap() made me a little sad initially, but on further thought I think it might be fine since it's during the build and should print something intelligible.

Yeah I think it's normal to panic in build.rs. I suppose technically we could `eprintln!` and then `std::process::exit(1)` but we weren't doing that before.

Dockerfile
```diff
@@ -114,6 +114,7 @@ RUN --mount=source=.cargo,target=/home/builder/.cargo \
     --mount=type=cache,target=/home/builder/rpmbuild/BUILD/sources/models/src/variant,from=variantcache,source=/variantcache \
     --mount=type=cache,target=/home/builder/rpmbuild/BUILD/sources/logdog/conf/current,from=variantcache,source=/variantcache \
     --mount=source=sources,target=/home/builder/rpmbuild/BUILD/sources \
+    --mount=source=tools,target=/home/builder/rpmbuild/BUILD/tools \
```
bcressey: This crosses a line that might exist only in my mind, but is still worth considering:
* `sources` for everything that might ship in an OS image
* `tools` for the disjoint set of everything needed to build OS images

This has some symmetry with the host vs. target distinction that cross-compilation introduces.

Things that could be independent crates (`parse-datetime`) that are useful to both are placed in sources until they graduate to their new home.

I'd prefer to follow that same pattern by adding a `generate-readme` crate to `sources` and then using that everywhere, including in `tools`.

(Eventually, in pursuit of faster build times, I could see pushing everything in `tools` into the SDK instead, at which point we'd need the shared dependencies to become external crates.)

sources/api/bootstrap-containers/build.rs
```diff
@@ -1,32 +1,7 @@
...
-        true,  // add title
-        false, // add badges
-        false, // add license
-        true,  // indent headings
-    )
-    .unwrap();
-
-    let mut readme = File::create("README.md").unwrap();
-    readme.write_all(content.as_bytes()).unwrap();
+    generate_readme(ReadmeSource::Main).unwrap()
```
bcressey: I don't mind this but it might be just as intuitive to have functions like:
```
generate_readme_main()
generate_readme_lib()
```

or
```
generate_readme::from_main()
generate_readme::from_lib()
```

(if it were a separate crate and we didn't want to have the import)

sources/api/certdog/build.rs
```diff
@@ -1,32 +1,7 @@
 // Automatically generate README.md from rustdoc.
```
bcressey: We could potentially clean up these comments as well.

tools/buildsys/src/public/variant.rs
```diff
@@ -0,0 +1,412 @@
...
+use std::convert::TryFrom;
+use std::fmt::{Display, Formatter};
+use std::ops::Deref;
+use std::str::FromStr;
+
+/// The name of the environment variable that tells us the current variant. Variant-sensitive crates
+/// will need to be rebuilt if this changes. `Makefile.toml` emits the variant string in the
+/// `BUILDSYS_VARIANT` environment variable. This is then passed to crate builds by the `Dockerfile`
+/// as `VARIANT`.
+pub const VARIANT_ENV: &str = "VARIANT";
```
bcressey: This could also be a separate crate (`bottlerocket-variant`) and we could import it in `builder.rs` to ensure we set the expected `VARIANT_ENV` value.

For that matter it might be useful to export all of these other fields into the environment, so that spec files don't have to mimic the parsing (as `os.spec` does).

tools/buildsys/src/public/variant.rs
```diff
@@ -0,0 +1,412 @@
...
+
+/// The default `variant_version`. If the third position of a variant string tuple does not exist,
+/// then the `variant_version` is `"undefined"`.
+pub const DEFAULT_VARIANT_VERSION: &str = "undefined";
+
+/// The default `variant_type`. If the fourth position of a variant string tuple does not exist,
+/// then the variant is considered to be `"general_purpose"`.
+pub const DEFAULT_VARIANT_TYPE: &str = "general_purpose";
```
bcressey: Can we omit these default values and just leverage the fact that the fields are `Option<String>`s and leave them set to `None`?

Otherwise we could default to more innocuous values, like `0` for the version and `"none"` for the type.

("general purpose" is a bit of an EC2-ism.)

tools/buildsys/src/public/variant.rs
```diff
@@ -0,0 +1,412 @@
...
+///
+/// // In a `build.rs` file, you may want to emit cfgs that you can use for conditional compilation.
+/// variant.emit_cfgs();
+/// ```
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct Variant {
+    variant: String,
+    platform: String,
+    runtime: String,
+    family: String,
```
bcressey: I don't love `family` but I don't have a better suggestion. I think this ends up being a pretty important concept for us elsewhere, e.g. #1261.

I'm fine with this for now but we might want to revisit this name (and "variant" for that matter) before we use it more widely.

tools/buildsys/src/public/variant.rs
```diff
@@ -0,0 +1,412 @@
...
+    pub fn version(&self) -> &str {
+        self.version.as_deref().unwrap_or(DEFAULT_VARIANT_VERSION)
+    }
+
+    /// The variant's type. This is the optional fourth value in the variant string tuple. For
+    /// example for `aws-k8s-1.21-nvidia` the `variant_type` is `nvidia`. If the `variant_type` does
+    /// not exist, [`DEFAULT_VARIANT_TYPE`] is returned.
+    pub fn variant_type(&self) -> &str {
+        self.variant_type.as_deref().unwrap_or(DEFAULT_VARIANT_TYPE)
+    }
```
bcressey: Maybe bring back `flavor` for this concept? Naming is hard, but I'd like to get away from terminology with a counterpart in genetics.

bcressey (commented):
Thanks! Overall this looks like a nice win in terms of eliminating a lot copy/paste code and bringing some consistency to how we interact with variants programmatically.

tools/buildsys/src/public/variant.rs
```diff
@@ -0,0 +1,412 @@
+/// Provides a structure for representing a Bottlerocket variant. Provides Functionality useful in
+/// build scripts and other tooling that are variant-aware.
```
arnaldo2792: nit: should the two phrases be joined with `and`?

tools/buildsys/src/public/variant.rs
```diff
@@ -0,0 +1,412 @@
...
+        let variant_type = parts.next().map(|s| s.to_string());
+        if let Some(value) = variant_type.as_ref() {
+            ensure!(
+                !value.is_empty(),
+                error::VariantPartEmptySnafu {
+                    part_name: "variant_type",
+                    variant: variant.clone()
+                }
+            );
+        }
```
arnaldo2792: Could you save yourself these two checks by adding a regex with the valid format of the variant? If `aws-k8s-` is given, the regex will catch the problem and the `is_empty()` check can be avoided.

Dockerfile
webern: `generate_readme` can be put into `sources` no problem.

The challenge with `Variant` is that we want to use it in both `tools` and `sources`.

Option 1:

If we put it into `sources` then we can reference it from `tools`. We have references already from `tools` -> `sources` already, so this is precedented https://github.com/bottlerocket-os/bottlerocket/blob/a7816b921cbb734a33e66d9e8fc95584fdb9c281/tools/pubsys/Cargo.toml#L46

Option 2:

We could start a new workspace at the top level, e.g. `shared`, for types that are going to be used by both `tools` and `sources` which would make crates used by both tools and sources visually apparent.

webern: P.S. in re-reading I see you already hinted at Option 2. So probably what we do here is put the new symbols in `sources` in their own crate then reference it from `testsys` in `tools`. This removes the copy of `tools` into the build container.

I should have done this initially, but I was enamored with the idea of the `buildsys` crate having to do with all things build related.

tools/buildsys/src/public/variant.rs
webern: I would prefer not to use regex just to check if the tuple position is empty. It looks like more code duplication than it actually is because 90% of it is error instantiation.

webern:
^ I believe I have addressed all of the comments.

sources/bottlerocket-variant/src/lib.rs
```diff
@@ -0,0 +1,442 @@
...
+///
+/// For example, here are some valid variant strings:
+/// - aws-ecs-1
+/// - vmware-k8s-1.18
+/// - metal-dev
+/// - aws-k8s-1.21-nvidia
+///
+/// The `platform` and `runtime` values are required. `variant_version` and `variant_flavor` values
+/// are optional and will default to `"undefined"` and `"general_purpose"` respectively.
+///
```
bcressey, line 65:
```suggestion
/// The `platform` and `runtime` values are required. `variant_version` and `variant_flavor` values
/// are optional and will default to `"0"` and `"none"` respectively.
///
```

bcressey (approved):

arnaldo2792 (approved):

webern:
I reran testing for aws-k8s, vmware-k8s and aws-ecs

PR #2142, opened 2022-05-18
`netdog`: Deny unknown fields

zmrow:
**Description of changes:**
```
    netdog: Deny unknown fields in interface/dhcp4/dhcp6
    
    Prior to this change, any unknown fields in the `net.toml` would be
    ignored rather than failing to parse.  Failing fast and printing an
    error means we understand immediately if the problem is a config format
    problem or if other debug is necessary.
```
```
    netdog: Extend `net.toml` unit test to include all supported configs
    
    This change extends the test file to include all the supported
    configurations possible with the added `optional` flag.
```

**Testing done:**
Unit tests all pass.

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/test_data/net_config/invalid_dhcp6_options.toml
+++ b/sources/api/netdog/test_data/net_config/invalid_dhcp6_options.toml
@@ -1,0 +1,9 @@
+version = 1
+
+[eno1]
+dhcp4 = true
+
+[eno2.dhcp6]
+optional = true
+# `route-metric` is not a valid `dhcp6` option
+route-metric = 100
--- a/sources/api/netdog/src/net_config.rs
+++ b/sources/api/netdog/src/net_config.rs
@@ -39,6 +39,7 @@ pub(crate) struct NetConfig {
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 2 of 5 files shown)


etungsten (approved):

webern (approved):

sources/api/netdog/test_data/net_config/invalid_dhcp6_options.toml
```diff
@@ -0,0 +1,8 @@
+version = 1
+
+[eno1]
+dhcp4 = true
+
+[eno2.dhcp6]
+optional = true
+route-metric = 100
```
bcressey: A comment in each of these saying why they're invalid might be helpful. It's kind of subtle!

bcressey (approved):

sources/api/netdog/test_data/net_config/invalid_dhcp6_options.toml
zmrow: Added!

zmrow:
^ added comments in the test files to clarify why they're invalid (thanks @bcressey )

bcressey (approved):

PR #2144, opened 2022-05-18
 Move primary interface sysctl config to netdog

zmrow:
**Description of changes:**
```
    systemd: rename var-run-tmpfiles to systemd-tmpfiles
    
    This renames var-run-tmpfiles to systemd-tmpfiles to better match
    convention in our other packages
```
```
    systemd: Add /etc/sysctl.d to tmpfiles
    
    This adds a line to `systemd-tmpfiles` to ensure that the
    `/etc/sysctl.d` directory is created at boot.
```
```
    Move primary interface sysctl config to netdog
    
    eth0 is not the default interface for all variants any longer.  This
    change removes the hardcoded `eth0` specific sysctls from default
    settings.  `netdog` is now responsible for setting these default sysctls
    for whatever the primary interface is for the platform.
    
    In order to do this, a new subcommand `prepare-primary-interface` is
    added to `netdog` and run via a oneshot service.  The command reads the
    primary interace from file (previously generated by `netdog`) and writes
    the sysctls for that interface to file in `/etc/sysctl.d`.  It then
    executes `systemd-sysctl`, which sets these sysctls.  Since all of this
    happens early in boot, a user is able to override these via API settings
    as usual.
```

**Testing done:**
Boot `aws-k8s-1.19`, `aws-k8s-1.22`, and `metal-dev` and ensure the sysctl file is written and sysctls are set:
```
net.ipv4.conf.eth0.rp_filter = 2

bash-5.1# sysctl net.ipv6.conf.eth0.accept_ra

bash-5.1# sysctl net.ipv4.conf.eth0.rp_filter
net.ipv4.conf.eth0.rp_filter = 2
```

**Terms of contribution:**


etungsten (approved):

sources/models/shared-defaults/eth0-sysctls.toml
```diff
@@ -0,0 +1,5 @@
+[settings.kernel.sysctl]
+# Accept router advertisement (RA) packets even if IPv6 forwarding is enabled on eth0
+"net.ipv6.conf.eth0.accept_ra" = "2"
+# Enable loose mode for reverse path filter
+"net.ipv4.conf.eth0.rp_filter" = "2"
```
bcressey: No migration is needed because these defaults will be applied to existing hosts on upgrade? It's worth testing that specifically.

sources/models/src/aws-dev/defaults.d/15-eth0-sysctls.toml
```diff
@@ -0,0 +1 @@
+../../../shared-defaults/eth0-sysctls.toml
```
bcressey: It might be better to have a platform defaults file - `aws.toml`, `vmware.toml`, `metal.toml` - to collect things like this.

bcressey (commented):
What takes care of applying the equivalent of this change to the primary interface on metal hosts? It feels kind of rough to expect the admin to discover they also need to add these sysctl settings.

Since we have the concept of primary interface in `netdog` now, I'd prefer letting `netdog` own this via a new oneshot service that runs at the right point in the boot sequence: after `/proc/sys/net/ipv4/conf/<primary>` exists; before settings are applied so that sysctl settings can still be used to override the defaults.

packages/systemd/systemd-tmpfiles.conf
```diff
@@ -0,0 +1 @@
+d /etc/sysctl.d 0700 root root -
```
bcressey, line 6: It might be better to rename `var-run-tmpfiles.conf` to `systemd-tmpfiles.conf` - to better match our convention in other packages - and then add this line to it.

sources/api/netdog/src/main.rs
```diff
@@ -58,6 +61,8 @@ static CURRENT_IP: &str = "/var/lib/netdog/current_ip";
 static KERNEL_CMDLINE: &str = "/proc/cmdline";
 static PRIMARY_INTERFACE: &str = "/var/lib/netdog/primary_interface";
 static DEFAULT_NET_CONFIG_FILE: &str = "/var/lib/bottlerocket/net.toml";
+static PRIMARY_SYSCTL_CONF: &str = "/etc/sysctl.d/10-primary_interface.conf";
```
bcressey: Ideally this would sort after `/usr/lib/sysctl.d/50-default.conf` which is where the upstream defaults come from, and after `/usr/lib/sysctl.d/80-release.conf` which is where our defaults come from.

Otherwise the behavior might change when `systemd-sysctl` is run with just this snippet, vs. with all snippets.

```suggestion
static PRIMARY_SYSCTL_CONF: &str = "/etc/sysctl.d/90-primary_interface.conf";
```

sources/api/netdog/src/main.rs
```diff
@@ -371,6 +382,50 @@ fn set_hostname(args: SetHostnameArgs) -> Result<()> {
...
+    S: AsRef<str>,
+    P: AsRef<Path>,
+{
+    let interface = interface.as_ref();
+    let path = path.as_ref();
+    // TODO if we accumulate more of these we should have a better way to create than format!()
+    // Accept router advertisement (RA) packets even if IPv6 forwarding is enabled on interface
+    let ipv6_accept_ra = format!("net.ipv6.conf.{}.accept_ra = 2", interface);
+    // Enable loose mode for reverse path filter
+    let ipv4_rp_filter = format!("net.ipv4.conf.{}.rp_filter = 2", interface);
```
bcressey: Given the potentially dire consequences of `systemd-sysctl` failing to return successfully, we might want to allow failure here:
```suggestion
    // Accept router advertisement (RA) packets even if IPv6 forwarding is enabled on interface
    let ipv6_accept_ra = format!("-net.ipv6.conf.{}.accept_ra = 2", interface);
    // Enable loose mode for reverse path filter
    let ipv4_rp_filter = format!("-net.ipv4.conf.{}.rp_filter = 2", interface);
```

zmrow:
^ Fixes per @bcressey 's suggestions

bcressey (approved):

cbgbt (approved):

PR #2146, opened 2022-05-19
netdog: Require `enabled` key in DHCP configuration

zmrow:
**Description of changes:**
```
The change makes the `enabled` option required when setting DHCP
options.  Previously the `net.toml` would be parsed and wicked
configuration would be generated just fine, but the interface wouldn't
ever come up because DHCP 4/6 wasn't enabled.  This causes a string of
failures in later services that depend on network being up.  This change
ensures the config requires the `enabled` option be set.

Leaving the `enabled` option out entirely was considered, as one could
assume the user wants DHCP to be enabled if additional options were
being set.  However, we didn't want to assume this for users, as there
could possibly be a use case for setting options, but leaving DHCP
disabled for an interface, perhaps to be used later.
```

**Testing done:**
* Boot a `metal-dev` image with `enabled` missing from the dhcp4/6 options and observed netdog properly fail to parse this
```
version = 1
[eno1]
dhcp4 = true  <-- this is valid syntax for simply turning on dhcp4

[eno2.dhcp4]  <--Missing the `enabled` key
optional = true
```
* Ensure `metal-dev` properly boots with properly formatted config, i.e:
```
version = 1
[eno1]
dhcp4 = true

[eno2.dhcp4]
enabled = true
optional = true
```

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/net_config.rs
+++ b/sources/api/netdog/src/net_config.rs
@@ -57,7 +57,7 @@ pub(crate) enum Dhcp4Config {
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub(crate) struct Dhcp4Options {
-    pub(crate) enabled: Option<bool>,
+    pub(crate) enabled: bool,
     pub(crate) optional: Option<bool>,
     #[serde(rename = "route-metric")]
     pub(crate) route_metric: Option<u32>,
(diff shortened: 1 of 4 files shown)


etungsten (approved):

bcressey:
Regarding this test case:
```
version = 1
[eno1]
dhcp4 = true

[eno2]
enabled = true
optional = true
```

Is that `eno2` configuration really valid?

zmrow:
>Regarding this test case:

>version = 1
>[eno1]
>dhcp4 = true

>[eno2]
>enabled = true
>optional = true

>Is that eno2 configuration really valid?

Sorry - copy/paste error.  I fixed it in the description.

bcressey (approved):

Issue #2132, opened 2022-05-10
Able to set --cluster-dns


mzupan:
**What I'd like:**

Being able to set --cluster-dns flag for kubelet. This allows the kubelet to set the dns search to something other then coredns like node-local-dns then fail back to coredns

gthao313:
Hi @mzupan  Thanks for opening this issue. Bottlerocket team will research on this and reach back to you soon. Thanks.

mzupan:
@gthao313 thanks how it was used with just a linux ami in eks was like

```
 --cluster-dns=169.254.20.10,172.20.0.10
```

The first IP being the IP node-local sets on each node and the 2nd being coredns.

That set any pod running on that node to have a resolv.conf that looked like

```
❯ k exec -it debug-6ff4966bc5-dn6tq -- cat /etc/resolv.conf
nameserver 169.254.20.10
nameserver 172.20.0.10
search default.svc.cluster.local svc.cluster.local cluster.local ec2.internal
options ndots:5
```

bcressey:
Does `settings.kubernetes.cluster-dns-ip` not work for this? We generate a value for the setting automatically, but if you specify a value that should override the default.

mzupan:
@bcressey that doesn't seem to allow you to specify multiple values like `--cluster-dns` in the kubelet does. Really there for if node-local somehow fails it uses coredns as a failback

bcressey:
Now that you point it out, it's obvious that it can be a list of IPs in the [kubelet config template](https://github.com/bottlerocket-os/bottlerocket/blob/6a82bc56b203ab27c78d5d043c3b4f4d5f1a68c8/packages/kubernetes-1.22/kubelet-config#L31), but it never crossed my mind to make it a list in our API. 😄 

@etungsten seems like we need some of the serde magic here to migrate `cluster-dns-ip` into a list shape on update.

cbgbt:
This is resolved in the latest Bottlerocket release, [v1.8.0](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.8.0) -- `settings.kubernetes.cluster-dns-ip` now supports being set to an array of IP addresses. Check [the documentation](https://github.com/bottlerocket-os/bottlerocket#kubernetes-settings) for an example on setting this.

Let us know if you have any questions!

PR #2140, opened 2022-05-18
Add aws-k8s-1.23, vmware-k8s-1.23, aws-k8s-1.23-nvidia, metal-k8s-1.23 variants

etungsten:
**Issue number:**
Resolves https://github.com/bottlerocket-os/bottlerocket/issues/2139

**Description of changes:**

For reference, the addition of k8s-1.22 variants was in https://github.com/bottlerocket-os/bottlerocket/pull/1962
```
    k8s-1.23: enable `CSIMigration*` feature gates in kubelet
    
    This enables the CSIMigration features gates that enables migration from
    in-tree CSI plugins to solution-specific CSI drivers.
    
    EKS 1.23 is planning to enable CSIMigrationAWS by default.
    
    vSphere container storage plugin migration has been in beta since 1.19
    and the in-tree plugin does not support the newest features in the
    vSphere CSI driver.
    
    Upstream kubernetes is eventually going to deprecate in-tree CSI
    plugins.
```
```
    metal-k8s: limit supported arch to x86_64
```
```
    Add metal-k8s-1.23 variant
    
    This change adds an additional variant `metal-k8s-1.23`, which includes
    necessary Kubernetes packages and settings for running Bottlerocket on
    metal in a Kubernetes v1.23 cluster.
```
```

    Add vmware-k8s-1.23 variant
    
    Adds vmware-k8s-1.23 variant, relinks symlinks in models
    vmware-k8s-1.23 supports boot config settings in models.

    Rename symlink for oci-hooks in vmware-k8s-1.22

```
```
    Add aws-k8s-1.23-nvidia variant
    
    Adds aws-k8s-1.23-nvidia variant, create symlinks for previous k8s
    version nvidia variant in models.
    
    aws-k8s-1.23-nvidia supports boot config settings while other nvidia
    variants do not.
```
```
    Add aws-k8s-1.23 variant
    
    Adds aws-k8s-1.23 variant, relinks symlinks in models.
    aws-k8s-1.23 supports settings.boot while older aws-k8s-* variants do not.
```
```
    containerd,k8s: add new containerd-config templates, update defaults
    
    For K8s 1.23, we're having kubelet use `/run/containerd/containerd.sock`
    for the container runtime socket instead of dockershim.sock.
    
    For compatiblity purposes, K8s 1.23 is maintaining a symlink at
    /run/dockershim.sock. For existing older K8s variants, we're keeping
    dockershim.sock as the main socket for now to avoid breakage until we
[...]

PR #2176, opened 2022-06-03
Allow `cluster-dns-ip` to be set to a list of IP Addresses

cbgbt:
**Issue number:** #2132 

**Description of changes:**
This allows `settings.kubernetes.cluster-dns-ip` to be set to a list. The value may also remain as a string, for backwards compatibility. The model will store the data using the type submitted, so existing code expecting string types will continue to work.

If a list is present, downwards migrations will truncate the list, using the first element as the configured value.

**Testing done:**
* [x] Ensure that `/etc/kubernetes/kubelet/config` contains `clusterDNS` settings, regardless of if the setting is a string or a list
* [x] Push the new migrations to a TUF repo and ensure they work as expected
* [x] Unit tests
* [x] Test on all modified k8s versions (1.19, 1.20, 1.21, 1.22)
* [x] Launch a pod with multiple DNS IPs configured and assert that `/etc/resolv.conf` contains them

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.20/kubelet-config
+++ b/packages/kubernetes-1.20/kubelet-config
@@ -29,7 +29,11 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 {{#if settings.kubernetes.cluster-dns-ip}}
 clusterDNS:
+{{#each settings.kubernetes.cluster-dns-ip}}
+- {{this}}
+{{else}}
 - {{settings.kubernetes.cluster-dns-ip}}
+{{/each}}
 {{/if}}
 {{#if settings.kubernetes.eviction-hard}}
 evictionHard:
--- a/packages/kubernetes-1.22/kubelet-config
+++ b/packages/kubernetes-1.22/kubelet-config
@@ -29,7 +29,11 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 {{#if settings.kubernetes.cluster-dns-ip}}
 clusterDNS:
+{{#each settings.kubernetes.cluster-dns-ip}}
+- {{this}}
+{{else}}
 - {{settings.kubernetes.cluster-dns-ip}}
+{{/each}}
 {{/if}}
 {{#if settings.kubernetes.eviction-hard}}
 evictionHard:
--- a/sources/api/migration/migration-helpers/src/lib.rs
+++ b/sources/api/migration/migration-helpers/src/lib.rs
@@ -60,7 +60,7 @@ pub type Metadata = HashMap<String, Value>;
 /// MigrationData holds all data that can be migrated in a migration, and serves as the input and
 /// output format of migrations.  A serde Value type is used to hold the arbitrary data of each
 /// key because we can't represent types when they could change in the migration.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct MigrationData {
     /// Mapping of data key names to their arbitrary values.
     pub data: HashMap<String, Value>,
--- a/sources/api/migration/migrations/v1.8.0/cluster-dns-ip-list/src/main.rs
+++ b/sources/api/migration/migrations/v1.8.0/cluster-dns-ip-list/src/main.rs
@@ -1,0 +1,204 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::{migrate, Migration, MigrationData, Result};
+use std::process;
+
+const CLUSTER_DNS_IP_KEY: &str = "settings.kubernetes.cluster-dns-ip";
+
+/// We changed `settings.kubernetes.cluster-dns-ip` to support being either a string or a list of strings.
+fn run() -> Result<()> {
+    migrate(ClusterDNSIPListMigration)
+}
+
+struct ClusterDNSIPListMigration;
+
+impl Migration for ClusterDNSIPListMigration {
+    /// New versions allow the older string values to be present, so we don't need to do anything.
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        println!("ClusterDNSIPListMigration has no work to do on upgrade.");
+        Ok(input)
+    }
+
+    /// Older versions don't know about list-style settings, so we need to create a scalar setting using the first value.
+    fn backward(&mut self, mut input: MigrationData) -> Result<MigrationData> {
+        let maybe_prior_value = input.data.get(CLUSTER_DNS_IP_KEY);
+
+        // If the current value is a string, don't touch it.
+        if let Some(prior_value) = maybe_prior_value {
+            if prior_value.is_string() {
+                println!(
+                    "{} is already a string value ('{}'), and does not require migration.",
+                    CLUSTER_DNS_IP_KEY, prior_value
+                );
+                return Ok(input);
+            }
+        }
+
+        // If the current value is an array and the first element is a string, that element becomes the new value.
+        // Any other cases result in clearing the value.
+        let new_value = maybe_prior_value
+            .and_then(|dns_ip_value| {
+                println!(
+                    "Found existing value for '{}': '{}'",
+                    CLUSTER_DNS_IP_KEY, dns_ip_value
+                );
+                dns_ip_value.as_array()
+            })
+            .and_then(|ip_array| ip_array.iter().next())
+            .map(|ip_value| ip_value.clone());
+
+        match new_value {
+            Some(ip_value) if ip_value.is_string() => {
+                input
+                    .data
+                    .insert(CLUSTER_DNS_IP_KEY.to_string(), ip_value.clone());
+                println!(
+                    "Replaced prior value for '{}' with '{}'",
+                    CLUSTER_DNS_IP_KEY, ip_value
+                );
+            }
+            _ => {
+                println!(
+                    "Prior value for '{}' was not recognized. Removing it.",
+                    CLUSTER_DNS_IP_KEY
+                );
+                input.data.remove(CLUSTER_DNS_IP_KEY);
+            }
+        };
+
+        Ok(input)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use std::collections::HashMap;
+
+    #[test]
+    fn test_downgrade_string() {
+        let input = MigrationData {
+            data: serde_json::from_str(r#"{"settings.kubernetes.cluster-dns-ip": "10.0.0.1"}"#)
+                .unwrap(),
+            metadata: HashMap::new(),
+        };
+        let expected = MigrationData {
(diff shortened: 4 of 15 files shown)


cbgbt:
Fixed formatting. I always forget to run check-fmt :face_exhaling:

webern (approved):

packages/kubernetes-1.22/kubelet-config
```diff
@@ -29,7 +29,11 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 {{#if settings.kubernetes.cluster-dns-ip}}
 clusterDNS:
+{{#each settings.kubernetes.cluster-dns-ip}}
+- {{this}}
+{{else}}
 - {{settings.kubernetes.cluster-dns-ip}}
+{{/each}}
```
bcressey, line 36: Need to make sure this ends up in 1.23 as well, if #2140 lands first.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -994,3 +995,121 @@ mod test_topology_manager_policy {
...
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesClusterDNSIp represents the --cluster-dns settings for kubelet.
+///
+/// The default is determined via information from IMDS in AWS.
+/// This model allows the value to be either a list of IPs, or a single IP string
+/// for backwards compatibility.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum KubernetesClusterDNSIp {
```
bcressey: Maybe this instead?
```suggestion
pub enum KubernetesClusterDnsIp {
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -994,3 +995,121 @@ mod test_topology_manager_policy {
         }
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// KubernetesClusterDNSIp represents the --cluster-dns settings for kubelet.
+///
+/// The default is determined via information from IMDS in AWS.
```
bcressey: Do we need to mention where the default comes from? Feels like a different concern.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -994,3 +995,121 @@ mod test_topology_manager_policy {
...
+            KubernetesClusterDNSIp::Scalar(IpAddr::from_str("::1").unwrap())
+        );
+    }
+
+    #[test]
+    fn test_parse_cluster_dns_ip_from_list() {
+        assert_eq!(
+            serde_json::from_str::<KubernetesClusterDNSIp>(r#"[]"#).unwrap(),
+            KubernetesClusterDNSIp::Vector(vec![])
+        );
```
bcressey: If we don't allow an empty string, should we allow an empty list?

I am OK either way as long as this doesn't cause trouble with `kubelet` when rendering the template:
```
[settings.kubernetes]
cluster-dns-ip = []
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -994,3 +995,121 @@ mod test_topology_manager_policy {
...
+        match self {
+            Self::Scalar(inner) => vec![inner],
+            Self::Vector(inner) => inner,
+        }
+        .into_iter()
+    }
+
+    pub fn first(&self) -> Option<&IpAddr> {
+        self.iter().next()
+    }
```
bcressey: Why add a special `first` method vs. making callers write `iter().next()` ? It seems like giving them the iterator and letting them use any of the trait methods is more powerful.

sources/api/migration/migration-helpers/src/lib.rs
```diff
@@ -60,7 +60,7 @@ pub type Metadata = HashMap<String, Value>;
 /// MigrationData holds all data that can be migrated in a migration, and serves as the input and
 /// output format of migrations.  A serde Value type is used to hold the arbitrary data of each
 /// key because we can't represent types when they could change in the migration.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
bcressey, line 63: We're adding `PartialEq` and `Eq` to facilitate unit testing? That is fine, just checking my understanding.

sources/api/migration/migrations/v1.8.0/cluster-dns-ip-list/src/main.rs
```diff
@@ -0,0 +1,204 @@
...
+        let new_value = maybe_prior_value
+            .and_then(|dns_ip_value| {
+                println!(
+                    "Found existing value for '{}': '{}'",
+                    CLUSTER_DNS_IP_KEY, dns_ip_value
+                );
+                dns_ip_value.as_array()
+            })
+            .and_then(|ip_array| ip_array.iter().next())
+            .map(|ip_value| ip_value.clone());
```
bcressey, line 48: Checking my understanding here also -

If we have a value and it's the empty list, we'll have a `None` here, and remove the setting from the datastore. The previous template had a guard for whether the setting was defined, so this is safe - we'll render a config file without that section.

packages/kubernetes-1.20/kubelet-config
```diff
@@ -29,7 +29,11 @@ authorization:
 clusterDomain: {{settings.kubernetes.cluster-domain}}
 {{#if settings.kubernetes.cluster-dns-ip}}
 clusterDNS:
+{{#each settings.kubernetes.cluster-dns-ip}}
+- {{this}}
+{{else}}
 - {{settings.kubernetes.cluster-dns-ip}}
+{{/each}}
```
bcressey, line 36: I'm a bit surprised that this works - `#each` iterates over a list and `else` triggers if it's defined but not actually a list?

sources/models/src/modeled_types/kubernetes.rs
cbgbt: The guard in the `kubelet` configs protects the config from an empty list, as empty lists [are considered "falsey" by handlebars](https://github.com/sunng87/handlebars-rust/blob/master/src/json/value.rs#L148).

I've tested empirically and confirmed that an empty list results in `clusterDNS` being absent from the `kubelet` config.

I think an empty string is "less correct" model-wise than an empty list. I'd like the consistency of disallowing both, but I'm not sure it's worth writing the custom deserializer since there's [not an easy way to enforce the constraint in serde](https://github.com/serde-rs/serde/issues/642) otherwise.

sources/api/migration/migration-helpers/src/lib.rs
cbgbt: That's right.

sources/api/migration/migrations/v1.8.0/cluster-dns-ip-list/src/main.rs
cbgbt: Yes, that's right.

packages/kubernetes-1.20/kubelet-config
cbgbt: I was pleased to discover this behavior, because otherwise we would need a custom helper which would get a bit messy. But yes, I looked [through the handlebars source](https://github.com/sunng87/handlebars-rust/blob/master/src/helpers/helper_each.rs#L143) and noticed that `#each` will render the `inverse()` statement in the template (which is the `else`) if the given value is not a list or object.

This is [consistent with the javascript implementation](https://handlebarsjs.com/playground.html#format=1&currentExample=%7B%22template%22%3A%22%7B%7B%23each%20listInput%7D%7D%5Cnlist%20%7B%7Bthis%7D%7D%5Cn%7B%7Belse%7D%7D%5Cnother%20%7B%7BlistInput%7D%7D%5Cn%7B%7B%2Feach%7D%7D%5Cn%5Cn%7B%7B%23each%20strInput%7D%7D%5Cnlist%20%7B%7Bthis%7D%7D%5Cn%7B%7Belse%7D%7D%5Cnother%20%7B%7BstrInput%7D%7D%5Cn%7B%7B%2Feach%7D%7D%22%2C%22partials%22%3A%5B%5D%2C%22input%22%3A%22%7B%5Cn%20%20listInput%3A%20%5B%5C%22test1%5C%22%2C%20%5C%22test2%5C%22%5D%2C%5Cn%20%20strInput%3A%20%5C%22test3%5C%22%2C%5Cn%7D%5Cn%22%2C%22output%22%3A%22list%20test1%5Cnlist%20test2%5Cn%5Cnother%20test3%5Cn%22%2C%22preparationScript%22%3A%22%22%2C%22handlebarsVersion%22%3A%224.7.7%22%7D) as well.

cbgbt:
Force push to address feedback from @bcressey.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -994,3 +995,119 @@ mod test_topology_manager_policy {
...
+    pub fn into_iter(self) -> impl Iterator<Item = IpAddr> {
+        match self {
+            Self::Scalar(inner) => vec![inner],
+            Self::Vector(inner) => inner,
+        }
+        .into_iter()
+    }
+}
+
+#[cfg(test)]
```
webern, line 1029: Nice tests.

Edit: somehow my other comment on the migration tests was lost. Basically, awesome that we have migration tests! 🚀

sources/api/pluto/src/main.rs
```diff
@@ -257,7 +257,11 @@ async fn get_node_ip(client: &mut ImdsClient) -> Result<String> {
     // Retrieve the user specified cluster DNS IP if it's specified, otherwise use the pluto-generated
     // cluster DNS IP
     let aws_k8s_info = api::get_aws_k8s_info().await.context(error::AwsInfoSnafu)?;
-    let cluster_dns_ip = if let Some(ip) = aws_k8s_info.cluster_dns_ip {
+    let configured_cluster_dns_ip = aws_k8s_info
+        .cluster_dns_ip
+        .and_then(|cluster_ip| cluster_ip.iter().next().cloned());
```
webern, line 262: Maybe check clippy on this.

webern (approved):

cbgbt:
Rebase on develop. Fix two clippy warnings in pluto.

sources/api/pluto/src/main.rs
cbgbt: It didn't have any tips here, but it did suggest some others in pluto, so I pushed some small changes.

etungsten (approved):

bcressey (approved):

zmrow (commented):
LGTM!

:mechanic:

Issue #834, opened 2020-03-09
Documentation: local dev/test workflow


jhaynes:
**What I'd like:**

When building a bottlerocket image, I get some disk images before converting them into an AMI. I'd like to be able to test them locally (e.g. via `qemu`or VMDK images) with full connectivity.

mdaniel:
There is already [a comment](https://github.com/bottlerocket-os/bottlerocket/blob/v0.3.1/packages/kernel/config-bottlerocket#L14) saying "for local testing with qemu," so even just adding a quick doc of how the existing developers run would be an improvement

The discussion in #842 seems relevant, also (and is what I tried first, but I had the same `dm-mapper` outcome when using VBox and I don't know enough qemu flags to know how to boot it up)

PR #2194, opened 2022-06-08
Qemu wrapper

markusboehme:
**Issue number:** #834 

**Description of changes:**

```
tools: Add start-local-vm script

Starts a virtual machine running a locally built Bottlerocket image via
QEMU and KVM. This is meant to ease development and experimentation for
situations that don't call for integration into a Kubernetes cluster or
other amenities provided in a cloud VM.

This does the minimal amount of work to meaningfully interact with the
launched VM: It configures the serial console for direct login available
in the -dev variants and forwards host TCP port 2222 to VM TCP port 22
so login via SSH works if both the network is configured and the admin
container enabled.

Users can inject files into the private partition of a Bottlerocket
image before it is launched to simulate the presence of user data and
other configuration. For example, "--inject-file ipv4-net.toml:net.toml"
adds the file "ipv4-net.toml" to the private partition as "net.toml".

Potentially helpful future work includes running images where host and
guest architecture differ, made possible by QEMU's TCG, as well as
generating and automatically injecting bare-bones variants of "net.toml"
and "user-data.toml" files to kick-start exploration.
```

**Testing done:**

Booted various VMs and configurations on both a c5.metal (x86_64) and c6g.metal (aarch64) EC2 instance. Testing is most fruitful with the metal-dev variant, since e.g. aws-dev seems to require the presence of IMDS.

**Further thoughts:**

* Do we want to lift the architecture restriction on the metal-dev variant? It works well on virtualized Arm and is pretty convenient for local testing since it doesn't have any dependencies on external infrastructure.
* Where might be a good place to put a pointer to this script and a usage example? I've been thinking of a `QUICKSTART-QEMU.md` file, possibly linked from `BUILDING.md`'s "Use your image" section.

**Terms of contribution:**


tools/start-local-vm
```diff
@@ -0,0 +1,242 @@
...
+                        a valid QEMU memory specifier (default is ${vm_mem})
+    --vm-cpus           number of CPUs to spawn for VM (default is ${vm_cpus})
+    --force-extract     force recreation of the extracted Bottlerocket image,
+                        e.g. to incorporate new --inject-file options
+    --inject-file       adds a local file to the private partition of the
+                        Bottlerocket image before launching the virtual machine
+                        (may be given multiple times)
+    --help              shows this usage text
+
+The virtual machine's port 22 (SSH) will be exposed via the local port 2222,
```
webern: Should the SSH host port mapping be an argument?

webern (commented):
This is cool.

tools/start-local-vm
```diff
@@ -0,0 +1,242 @@
...
+    # causing udev to use the bus location when naming the NIC (enp0s16). Since
+    # QEMU does not support PCI hot plug via ACPI on Arm, turn it off for the
+    # emulated x86_64 chipset only to achieve parity.
+    if [[ ${arch} = x86_64 ]]; then
+        qemu_args+=( -global PIIX4_PM.acpi-root-pci-hotplug=off )
+    fi
+
+    if [[ ${arch} = aarch64 ]]; then
+        qemu_args+=( -machine virt )
+        qemu_args+=( -bios /usr/share/edk2/aarch64/QEMU_EFI.silent.fd )
```
arnaldo2792, line 253: Is this provided by all distros? Or is distro specific, and the file is installed in different locations depending on the distro?

tools/start-local-vm
```diff
@@ -0,0 +1,242 @@
...
+                        host architecture ($(uname -m)); may be omitted if the
+                        BUILDSYS_ARCH environment variable is set)
+    --variant           Bottlerocket variant to run (may be omitted if the
+                        BUILDSYS_VARIANT environment variable is set)
+    --vm-memory         amount of memory to assign to the VM; VM_MEMORY must be
+                        a valid QEMU memory specifier (default is ${vm_mem})
+    --vm-cpus           number of CPUs to spawn for VM (default is ${vm_cpus})
+    --force-extract     force recreation of the extracted Bottlerocket image,
+                        e.g. to incorporate new --inject-file options
+    --inject-file       adds a local file to the private partition of the
```
arnaldo2792, line 49: I really liked this :D! One potential addition to this might be to let the users attach their own NICs. I know that I would use that since I run a bunch of VMs in metal variants to test OS specifics, with an ENI attached to them which allows me to use SSM to connect to the VMs instead of SSH.

tools/start-local-vm
```diff
@@ -0,0 +1,242 @@
...
+                        Bottlerocket image before launching the virtual machine
+                        (may be given multiple times)
+    --help              shows this usage text
+
+The virtual machine's port 22 (SSH) will be exposed via the local port 2222,
+i.e. if the Bottlerocket admin container has been enabled via user-data, it can
```
arnaldo2792: For the lack of a better place to ask, what happens to the control container? Does it just fail?

tools/start-local-vm
markusboehme: In case the host's port 2222 is already taken, this would be helpful, yes. It won't do anything to be able to run multiple VMs off the script since there's other work needed for that (e.g. switching to a separate layer for disk writes so VMs don't share their images).

I think a separate `--host-port-forwards` option defaulting to `tcp::2222:22` is most generic and avoids tying this option to SSH too much while still keeping the base case simple.

tools/start-local-vm
markusboehme: Totally decided by a distro's packagers. On current Fedora, this is where the `edk2-arm` package (pulled in by `qemu`) puts it. On an Ubuntu machine it might be provided by `ovmf` and only pulled in as a recommended package. For a start, I wanted to go with the fixed location and see whether there's actually any breakage.

tools/start-local-vm
markusboehme: Passing through a secondary ENI on an EC2 instance? I haven't tried that myself yet, but it's definitely one way to extend this and improve the iteration time!

tools/start-local-vm
markusboehme: I don't know! :-) The control container isn't active in the metal variants, and with, for example, aws-dev early-boot-config fails for lack of access to the IMDS (which likely has knock-on effects on the control container). Here's where the extension idea of ENI pass-through comes into play.

tools/start-local-vm
```diff
@@ -0,0 +1,242 @@
...
+    else
+        bail "${usage}"
+    fi
+}
+
+parse_args() {
+    while [[ $# -gt 0 ]]; do
+        case $1 in
+            -h|--help)
+                show_usage ;;
```
bcressey: nit: it is atypical to return non-zero when help is explicitly requested

tools/start-local-vm
```diff
@@ -0,0 +1,242 @@
...
+                extra_files[${local_file}]=${image_file}
+                ;;
+            *)
+                show_usage "unknown option '$1'" ;;
+        esac
+        shift
+    done
+
+    [[ -n ${arch} ]] || show_usage 'Architecture needs to be set via either --arch or BUILDSYS_ARCH.'
+    [[ -n ${variant} ]] || show_usage 'Variant needs to be set via either --variant or BUILDSYS_VARIANT.'
```
bcressey: Regarding the `BUILDSYS_` variables - is the goal to eventually integrate this as a task in `Makefile.toml`? Or just as a convenience for developers who export these variables into their environment?

tools/start-local-vm
```diff
@@ -0,0 +1,242 @@
...
+
+remove_old_images() {
+    # Any of the latest images can serve as our touchstone. Older images can go.
+    declare -r any_current_image=${current_images[0]}
+
+    for extracted_image in "${repo_root}/build/images/${arch}-${variant}"/*.img; do
+        if [[ ${extracted_image} -ot ${any_current_image} ]]; then
+            rm "${extracted_image}"
+        fi
+    done
```
bcressey, line 144: This is a nice touch.

tools/start-local-vm
```diff
@@ -0,0 +1,242 @@
...
+
+inject_files() {
+    if [[ ${#extra_files[@]} -eq 0 ]]; then
+        return 0
+    fi
+
+    local loopback_device
+    if ! loopback_device=$(sudo losetup --find --show --partscan "${boot_image}"); then
+        bail 'Failed to create the loopback device needed to splice in user data.'
+    fi
```
bcressey: Since we're only modifying the private partition, it might be simpler and faster to:
* copy local files to a temporary directory
* create a new ext4 filesystem from the directory
* write that filesystem to the offset in the boot image

That would avoid the need for privileged operations, and the general pain of dealing with `losetup`.

bcressey (approved):
Cool!

bcressey:
> Do we want to lift the architecture restriction on the metal-dev variant? It works well on virtualized Arm and is pretty convenient for local testing since it doesn't have any dependencies on external infrastructure.

+1

> Where might be a good place to put a pointer to this script and a usage example? I've been thinking of a QUICKSTART-QEMU.md file, possibly linked from BUILDING.md's "Use your image" section.

That makes sense to me. We also need to document the Fedora / Ubuntu package requirements.

It might be an interesting experiment to see if we could install QEMU in the SDK and invoke it via `docker run` instead of needing it to be installed by developers. If that worked it could pave the way for `Makefile.toml` integration (`cargo make qemu-run ...`).

zmrow (approved):
Neat! :camera_flash:

foersleo (approved):

markusboehme:
Ah, rebasing wasn't a great idea. Now the comparison between revisions is tedious a best. The GitHub workflow is new to me--I assume we don't rebase our forks when having open PRs on them?

tools/start-local-vm
markusboehme: Converted to `--host-port-forwards` with the default as described.

tools/start-local-vm
markusboehme: I hadn't thought about integration into cargo-make yet but it is something to consider (perhaps after the script learns how to auto-generate some configuration such as `net.toml`). My original motivation for those indeed was convenience in case one has those exported anyway.

tools/start-local-vm
markusboehme: I wasn't really happy with the loopback device and the mount either. Regenerating the partition didn't occur to me, but I really like the idea. Losing its original content is no big deal for the development workflow. (As an alternative, @zmrow  pointed me to the great libguestfs/guestfish that I didn't know until then, but aside from the performance aspect of launching a VM each time that bothered me a little, I was unable to determine the private partition using its label.) I implemented your proposal in the version I just pushed.

markusboehme:
@bcressey:

> It might be an interesting experiment to see if we could install QEMU in the SDK and invoke it via `docker run` instead of needing it to be installed by developers. If that worked it could pave the way for `Makefile.toml` integration (`cargo make qemu-run ...`).

It's a fun idea. I don't see why it shouldn't work. Comes with the benefit of only having one environment to adapt to, too. I'll play with this.

bcressey:
> Ah, rebasing wasn't a great idea. Now the comparison between revisions is tedious a best. The GitHub workflow is new to me--I assume we don't rebase our forks when having open PRs on them?

If you don't need to rebase, it's generally easier to avoid it.

The convention we've adopted in cases where it's necessary:
1) rebase the branch with no changes to the review code
2) force push and add a "rebase, no changes" comment to the PR
3) make changes to the review code
4) force push and add a summary comment to the PR

You have to wait 2-3 minutes between the two force pushes, otherwise GitHub will coalesce them into a single timeline note and the force push diff link is useless.

It isn't the end of the world if something goes wrong - your "oops, rebase" note is fine and people can just re-review the full changes.

bcressey (approved):

foersleo (approved):

PR #2205, opened 2022-06-09
Documentation: Add and update documentation for metal variants

zmrow:
**Description of changes:**
```
    Add HARDWARE.md
    
    This adds a new file HARDWARE.md meant to document the hardware that
    Bottlerocket is currently tested on, not necessarily all of the hardware
    that can currently boot Bottlerocket.
```
```
    Add PROVISIONING-METAL.md
    
    This adds a new document that describes the files and process needed to
    provision Bottlerocket on a bare metal machine.
```
```
    README.md: Update for metal
    
    Updates README.md to add some additional information for metal variants.
```

**Terms of contribution:**

### DIFF
--- a/PROVISIONING-METAL.md
+++ b/PROVISIONING-METAL.md
@@ -1,0 +1,188 @@
+# Provisioning Bottlerocket on metal
+
+This guide will describe what is needed to properly provision Bottlerocket on bare metal.
+Provisioning Bottlerocket on metal is different than provisioning other general-purpose distros.
+Since Bottlerocket has a `dm-verity`-checked boot and root partition, and is immutable at runtime, a user cannot provision an image and directly write configuration files.
+Bottlerocket requires a few files to be generated and written to disk at provisioning time in order to boot properly; these files are described below.
+
+For more information about the hardware that Bottlerocket for bare metal is currently tested on, see [SUPPORTED-HARDWARE](SUPPORTED-HARDWARE.md).
+
+## High level provisioning steps
+
+The high level steps to provision Bottlerocket images for bare metal to your host are below.
+Most provisioning systems provide methods to achieve the following:
+
+* Decompress (`unlz4`) and write the Bottlerocket image to the desired disk
+* Mount the `BOTTLEROCKET-PRIVATE` partition (partition 12)
+* Write the below files to the mounted partition (these files are further described below):
+  * (Required) [`user-data.toml`](#user-data)
+  * (Required) [`net.toml`](#network-interface-configuration)
+  * (Optional, recommended) [`bootconfig.data`](#boot-configuration)
+* Reboot
+
+### Fetch the Bottlerocket image for bare metal
+
+The Bottlerocket image for bare metal is signed and uploaded alongside the rest of the Bottlerocket release artifacts.
+
+You first need the Bottlerocket root role, which is used by `tuftool` to verify the image.
+The following will download and verify the root role itself:
+
+```
+curl -O "https://cache.bottlerocket.aws/root.json"
+sha512sum -c <<<"e9b1ea5f9b4f95c9b55edada4238bf00b12845aa98bdd2d3edb63ff82a03ada19444546337ec6d6806cbf329027cf49f7fde31f54d551c5e02acbed7efe75785  root.json"
+```
+
+Next, set your desired version and variant, and use `tuftool` to download the image:
+To install `tuftool` you'll need to install Rust (via [rustup](https://rustup.rs/) or the official site), and then you can run `cargo install tuftool`.
+```
+ARCH="x86_64"
+VERSION="1.8.0"
+VARIANT="metal-k8s-1.23"
+COMMIT="a6233c22"
+IMAGE="bottlerocket-${VARIANT}-${ARCH}-${VERSION}-${COMMIT}.img.lz4"
+OUTDIR="${VARIANT}-${VERSION}"
+
+tuftool download "${OUTDIR}" --target-name "${IMAGE}" \
+   --root ./root.json \
+   --metadata-url "https://updates.bottlerocket.aws/2020-07-07/${VARIANT}/x86_64/" \
+   --targets-url "https://updates.bottlerocket.aws/targets/"
+```
+
+### User data
+
+Bottlerocket for bare metal expects a TOML-formatted file named `user-data.toml` that contains user data settings.
+Acceptable settings can be found in the [settings docs](https://github.com/bottlerocket-os/bottlerocket#settings).
+
+If you're just getting started and want to provision a host without connecting to a Kubernetes cluster, you can use the following example user data which will start `kubelet` in standalone mode.
+
+```toml
+[settings.kubernetes]
+standalone-mode = true
+```
+
+For remote access to your running Bottlerocket hosts, you will need to add user data to enable host containers.
+The Bottlerocket images for bare metal don't enable any host containers by default.
+You can use our [admin](https://github.com/bottlerocket-os/bottlerocket-admin-container) and/or [control](https://github.com/bottlerocket-os/bottlerocket-control-container) containers, but they need to be configured first.
+Full configuration details are covered in the [admin container documentation](https://github.com/bottlerocket-os/bottlerocket-admin-container#authenticating-with-the-admin-container) and the [control container documentation](https://github.com/bottlerocket-os/bottlerocket-control-container#connecting-to-aws-systems-manager-ssm).
+
+### Network interface configuration
+
+Bottlerocket for bare metal provides the means to configure the physical network interfaces in the system via TOML-formatted file `net.toml`.
+For now, simple DHCP4 and DHCP6 configuration is supported with plans to support additional configuration in the future.
+
+`net.toml` is read at boot time and generates the proper configuration files in the correct format for each interface described; no default configuration is provided.
+If no network configuration is provided, boot-time services like host containers, `containerd`, and `kubelet` will fail to start.
+When these services fail, your machine will not connect to any cluster and will be unreachable via host containers.
+
+#### `net.toml` structure
+
+The configuration file must be valid TOML and have the filename `net.toml`.
+The first and required top level key in the file is `version`, currently only `1` is supported.
+The rest of the file is a map of interface name to supported settings.
+Interface names are expected to be correct as per `udevd` naming, no interface naming or matching is supported.
+(See the note below regarding `udevd` interface naming.)
+
+#### Supported interface settings
+
+* `primary` (boolean): Use this interface as the primary network interface. `kubelet` will use this interface's IP when joining the cluster.  If none of the interfaces has `primary` set, the first interface in the file is used as the primary interface.
+* `dhcp4` (boolean or map): Turns on DHCP4 for the interface.  If additional DHCP4 configuration is required, the following settings are supported and may be provided as a map with the following keys:
+  * `enabled` (boolean, required): Enables DHCP4.
+  * `route-metric` (integer): Prioritizes routes by setting values for preferred interfaces.
+  * `optional` (boolean): the system will request a lease using this protocol, but will not wait for a valid lease to consider this interface configured.
+* `dhcp6` (boolean or map): Turns on DHCP6 for the interface.  If additional DHCP6 configuration is required, the following settings are supported and may be provided as a map with the following keys:
+  * `enabled` (boolean, required): Enables DHCP6.
+  * `optional` (boolean): the system will request a lease using this protocol, but will not wait for a valid lease to consider this interface configured.
+
+Example `net.toml` with comments:
+```toml
+version = 1
+
+# "eno1" is the interface name
+[eno1]
+# Users may turn on dhcp4 and dhcp6 via boolean
+dhcp4 = true
+dhcp6 = true
+primary = true
+
+# "eno2" is the second interface in this example
+[eno2.dhcp4]
+# `enabled` is a boolean and is a required key when
+# setting up DHCP this way
+enabled = true
+# Route metric may be supplied for ipv4
+route-metric = 200
+
+[eno2.dhcp6]
+enabled = true
+optional = true
+```
+
+**An additional note on network device names**
+
+Interface name policies are [specified in this file](https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/release/80-release.link#L6); with name precedence in the following order: onboard, slot, path.
+Typically on-board devices are named `eno*`, hot-plug devices are named `ens*`, and if neither of those names are able to be generated, the “path” name is given, i.e `enp*s*f*`.
+
+### Boot Configuration
+
+Bottlerocket for bare metal uses a feature of the Linux kernel called [Boot Configuration](https://www.kernel.org/doc/html/latest/admin-guide/bootconfig.html), which allows a user to pass additional arguments to the kernel command line at runtime.
+An immediate use of this feature for most users is setting `console` settings so boot messages can be seen on the appropriate consoles.
+
+In order to make use of this feature, an initrd is created with the desired settings encoded inside it.
+The initrd is empty save for the encoded boot config data.
+To create the initrd, you must first create a configuration file containing key value pairs for the settings you would like to pass to kernel / init.
+Full syntax is described in the [Boot Config documentation](https://www.kernel.org/doc/html/latest/admin-guide/bootconfig.html#config-file-syntax), but a simple example is provided below that shows the format of console settings as well as an example `systemd` parameter.
+
+The two acceptable prefixes to settings are `kernel` and `init`.
+Settings prefixed with `kernel` are added to the beginning of the kernel command line.
+Settings prefixed with `init` are added to the kernel command line after the `--`, but before any existing init parameters.
+
+In the example below, two console devices are set up, and `systemd`'s log level is set to `debug`.
+
+Example Boot Configuration:
+```
+kernel {
+    console = tty0, "ttyS1,115200n8"
+}
+init {
+    systemd.log_level = debug
+}
+```
+
+The Bottlerocket SDK provides the `bootconfig` CLI tool, which is used to create a Boot Configuration initrd.
+To create the Boot Configuration initrd, create a config file named `bootconfig-input` containing your desired key/value pair kernel and init arguments.
+
+Then run the following (you will need Docker installed):
+```
+ARCH=$(uname -m)
+SDK_VERSION="v0.26.0"
+SDK_IMAGE="public.ecr.aws/bottlerocket/bottlerocket-sdk-${ARCH}:${SDK_VERSION}"
+
+touch $(pwd)/bootconfig.data
+
+docker run --rm \
+   --network=none \
+   --user "$(id -u):$(id -g)" \
+   --security-opt label:disable \
+   -v $(pwd)/bootconfig-input:/tmp/bootconfig-input \
+   -v $(pwd)/bootconfig.data:/tmp/bootconfig.data \
+   "${SDK_IMAGE}" \
+   bootconfig -a /tmp/bootconfig-input /tmp/bootconfig.data
+```
+
+The above command will create the properly named initrd `bootconfig.data` in your current directory.
+This is the file you will write to disk during provisioning.
+
+You can list a `bootconfig.data`'s contents, which also validates its format, by running:
+```
+ARCH=$(uname -m)
+SDK_VERSION="v0.26.0"
+SDK_IMAGE="public.ecr.aws/bottlerocket/bottlerocket-sdk-${ARCH}:${SDK_VERSION}"
+
+docker run --rm \
+   --network=none \
+   --user "$(id -u):$(id -g)" \
+   --security-opt label:disable \
+   -v $(pwd)/bootconfig.data:/tmp/bootconfig.data \
+   "${SDK_IMAGE}" \
+   bootconfig -l /tmp/bootconfig.data
+```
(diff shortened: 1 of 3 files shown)


PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+The following will download and verify the root role itself:
+
+```
+curl -O "https://cache.bottlerocket.aws/root.json"
+sha512sum -c <<<"e9b1ea5f9b4f95c9b55edada4238bf00b12845aa98bdd2d3edb63ff82a03ada19444546337ec6d6806cbf329027cf49f7fde31f54d551c5e02acbed7efe75785  root.json"
+```
+
+Next, set your desired version and variant, and download the image:
+```
+VERSION="v1.8.0"
```
etungsten: There should be no `v` here.

HARDWARE.md
```diff
@@ -0,0 +1,13 @@
+# Bottlerocket test hardware
```
webern: I feel like this could be consolidated with the PROVISIONING-METAL doc, and I wonder if that one should be called QUICKSTART-METAL.

zmrow: IMHO, the hardware felt worthy of it's own doc rather than buried in other metal documentation.

Re: QUICKSTART-METAL, I opted not to create a "quickstart" type doc since provisioning bare metal machines isn't really prescriptive like we can be with AWS variants; "run this command to start an instance", etc.

PROVISIONING-METAL.md
zmrow: Fixed!

zmrow:
^ addresses @etungsten 's comments

zmrow:
* Finished hardware table
* Moved out of draft

HARDWARE.md
webern: > IMHO, the hardware felt worthy of it's own doc rather than buried in other metal documentation.
>
> Re: QUICKSTART-METAL, I opted not to create a "quickstart" type doc since provisioning bare metal machines isn't really prescriptive like we can be with AWS variants; "run this command to start an instance", etc.

Makes sense. I'm still wondering if HARDWARE.md and PROVISIONING-METAL.md should be consolidated instead of being separate documents. Maybe METAL.md?

webern (approved):

HARDWARE.md
```diff
@@ -0,0 +1,13 @@
...
+
+Bottlerocket has been tested on and known to work with the hardware below.
+
+Bottlerocket's kernel configuration isn't as extensive as some other distros, so it's possible that it is missing drivers for your specific hardware.
+If so, please [submit an issue](https://github.com/bottlerocket-os/bottlerocket/issues/new?assignees=&labels=&template=feature.md) and we'll work on integrating the proper kernel modules for your hardware!
+
+| Server model | BIOS/UEFI | CPU _| Network Card | Disk | RAID/Storage controller |
```
etungsten: The values for theses two columns are flipped in the rich diff

README.md
```diff
@@ -410,7 +418,7 @@ Static pods can be particularly useful when running in standalone mode.
 
 For Kubernetes variants in AWS and VMware, the following are set for you automatically, but you can override them if you know what you're doing!
 In AWS, [pluto](sources/api/) sets these based on runtime instance information.
-In VMware, Bottlerocket uses [netdog](sources/api/) (for `node-ip`) or relies on [default values](sources/models/src/vmware-k8s-1.23/defaults.d).
+In VMware and on bare metal, Bottlerocket uses [netdog](sources/api/) (for `node-ip`) or relies on [default values](sources/models/src/vmware-k8s-1.23/defaults.d).
```
etungsten: The default values link points to vmware k8s variant defaults and it's not necessarily the same as the metal variant defaults.

README.md
```diff
@@ -4,7 +4,7 @@ Welcome to Bottlerocket!
 
 Bottlerocket is a free and open-source Linux-based operating system meant for hosting containers.
 
-If you’re ready to jump right in, read one of our setup guides for running Bottlerocket in [Amazon EKS](QUICKSTART-EKS.md), [Amazon ECS](QUICKSTART-ECS.md), or [VMware](QUICKSTART-VMWARE.md).
+If you’re ready to jump right in, read one of our setup guides for running Bottlerocket in [Amazon EKS](QUICKSTART-EKS.md), [Amazon ECS](QUICKSTART-ECS.md), [VMware](QUICKSTART-VMWARE.md), or provisioning on [bare metal](PROVISIONING-METAL).
```
```suggestion
If you’re ready to jump right in, read one of our setup guides for running Bottlerocket in [Amazon EKS](QUICKSTART-EKS.md), [Amazon ECS](QUICKSTART-ECS.md), [VMware](QUICKSTART-VMWARE.md), or provisioning on [bare metal](PROVISIONING-METAL.md).
```

README.md
```diff
@@ -95,6 +101,8 @@ These guides describe:
 * how to set up a cluster with the orchestrator, so your Bottlerocket instance can run containers
 * how to launch a Bottlerocket instance in EC2 or VMware
 
+To see how to provision Bottlerocket on bare metal, see [PROVISIONING-METAL](PROVISIONING-METAL).
```
```suggestion
To see how to provision Bottlerocket on bare metal, see [PROVISIONING-METAL](PROVISIONING-METAL.md).
```

etungsten (changes requested):

HARDWARE.md
bcressey:
```suggestion
| Server model | BIOS/UEFI | CPU | Network Card | Disk | RAID/Storage controller |
```

HARDWARE.md
bcressey: "test hardware" doesn't sound correct to me. Also I think the file should be called `SUPPORTED-HARDWARE.md`.

Seems like we want two tiers:
* official support - hardware that the project or partners have access to and qualify builds on prior to release
* community support - hardware that's been reported to work

We might also want to consider using the wiki feature for the other category, if we want to streamline the ability to mark new platforms for the "community support" level.

@stockholmux what do you think?

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
+# Povisioning Bottlerocket on metal
```
bcressey:
```suggestion
# Provisioning Bottlerocket on metal
```

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+
+This guide will describe what is needed to properly provision Bottlerocket on bare metal.
+Provisioning Bottlerocket on metal is different than provisioning other general-purpose distros.
+Since Bottlerocket has a `dm-verity`-checked boot and root partition, and is immutable at runtime, we cannot simply provision an image and directly write configuration files.
```
bcressey: :boromir-meme:

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+Bottlerocket requires a few files to be generated and written to disk at provisioning time in order to boot properly; these files are described below.
+
+> For more information about the hardware that Bottlerocket is currently tested on, see [Hardware](HARDWARE.md).
```
bcressey: I am not sure about the stylistic merits of using a quote like this as a sort of footnote. Maybe just remove the quote marker.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+The high level steps to provision Bottlerocket to your host are below.
+Most provisioning systems provide methods to achieve the following:
+
+* Decompress (`unlz4`) and write the Bottlerocket image to the desired disk
+* Mount the `BOTTLEROCKET-PRIVATE` partition (partition 12)
+* Write the below files to the mounted partition:
+  * (Required) `user-data.toml`
+  * (Required) `net.toml`
+  * (Optional, recommended) `bootconfig.data`
+* Reboot
```
bcressey: This might be better at the top, before prerequisites, so the reader has the big picture before getting the details.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+
+> For more information about the hardware that Bottlerocket is currently tested on, see [Hardware](HARDWARE.md).
+
+## Prerequisites
+
+* Bottlerocket metal image
+  * [`tuftool`](https://github.com/awslabs/tough/blob/develop/tuftool/README.md) is required to fetch the image
+* [User data TOML file](#user-data): `user-data.toml`
+* [Network configuration TOML file](#network-interface-configuration): `net.toml`
+* (Optional) [Boot configuration file](#boot-configuration): `bootconfig.data`
```
bcressey: "Prerequisites" might not be the right word; that tells me I need these things before continuing, when the purpose of the guide is to tell me how to get them.

The high-level steps below might work as a replacement for this section.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Fetch the Bottlerocket metal image
+
+The Bottlerocket metal image is signed and uploaded alongside the rest of the Bottlerocket release artifacts.
+
+You first need the Bottlerocket root role, which is used by `tuftool` to verify the image.
```
bcressey, line 27: Let's tell them how to get `tuftool`.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+```
+curl -O "https://cache.bottlerocket.aws/root.json"
+sha512sum -c <<<"e9b1ea5f9b4f95c9b55edada4238bf00b12845aa98bdd2d3edb63ff82a03ada19444546337ec6d6806cbf329027cf49f7fde31f54d551c5e02acbed7efe75785  root.json"
+```
+
+Next, set your desired version and variant, and download the image:
+```
+VERSION="1.8.0"
+VARIANT="metal-k8s-1.23"
+SHA="a6233c22"
```
bcressey: I would call this `COMMIT` or `GIT` or something more descriptive.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+Acceptable settings can be found in the [settings docs](https://github.com/bottlerocket-os/bottlerocket#settings).
+
+If you're just getting started and want to provision a host without connecting to a kubernetes cluster, you can use the following example user data which will start kubelet in standalone mode.
+
+```toml
+[settings.kubernetes]
+standalone-mode = true
+```
+
+It's important to note for remote access to your running Bottlerocket hosts, you will need to add user data to enable host containers.
```
bcressey:
```suggestion
For remote access to your running Bottlerocket hosts, you will need to add user data to enable host containers.
```

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+The Bottlerocket metal images don't enable any host containers by default.
```
bcressey: As a general editorial style note, I'd recommend coming up with 2-3 phrases and using them consistently:
* "Bottlerocket images for bare metal"
* "Bottlerocket for bare metal"

Currently it reads kind of sloppy and imprecise because we switch between talking about "Bottlerocket" and various permutations of "Bottlerocket for bare metal", and it's not clear when the reader is learning something that's generally applicable to all of Bottlerocket, versus something that's specific to bare metal.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+You can use our [admin](https://github.com/bottlerocket-os/bottlerocket-admin-container) and/or [control](https://github.com/bottlerocket-os/bottlerocket-control-container) containers, but they need to be configured first.
+Full configuration details are covered in the [admin container documentation](https://github.com/bottlerocket-os/bottlerocket-admin-container#authenticating-with-the-admin-container) and the [control container documentation](https://github.com/bottlerocket-os/bottlerocket-control-container#connecting-to-aws-systems-manager-ssm).
+
+### Network interface configuration
+
+Bottlerocket provides the means to configure the physical network interfaces in the system via TOML-formatted file `net.toml`.
+Bottlerocket will read this file at boot time and generate the proper configuration files in the correct format for each interface described; no default configuration is provided.
```
bcressey: Let's mention the implications of "no default configuration is provided".

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+For now, we support simple DHCP4 and DHCP6 configuration and plan to support additional configuration in the future.
+
+#### `net.toml` structure
+
+The configuration file must be valid TOML and have the filename `net.toml`.
+The first and required top level key in the file is `version`, currently only `1` is supported.
+The rest of the file is a map of interface name to supported settings.
+Interface names are expected to be correct as per `udevd` naming, no interface naming or matching is supported.
```
bcressey, line 82: Add a link to read more about the naming convention used.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+
+#### Supported interface settings
+
+* `primary` (boolean): Uses the interface as the primary network interface, kubelet will use this interface's IP to register.  If none of the interfaces has `primary` set, we choose the first interface in the file.
```
bcressey:
```suggestion
* `primary` (boolean): Use this interface as the primary network interface. `kubelet` will use this interface's IP when joining the cluster.  If none of the interfaces has `primary` set, we choose the first interface in the file.
```

Another stylistic note - @stockholmux might be trying to move away from the "we" language especially when talking about implementation mechanics.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+[eno2.dhcp6]
+enabled = true
+optional = true
+```
+
+> **An additional note on network device names**
+> 
+> Bottlerocket’s interface name policies are [specified in this file](https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/release/80-release.link#L6); with name precedence in the following order: onboard, slot, path.
+> What this means is that udev finds and generates the names for the device, and precedence is given in that order.
+> Typically on-board devices are named `eno*`, hot-plug devices are named `ens*`, and if neither of those names are able to be generated, the “path” name is given, i.e `enp*s*f*`.
```
bcressey: This probably eliminates the need for the earlier `udevd` reference. Also, I'd get rid of the quote marker, and maybe condense the material if it feels too detailed.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Boot Configuration
+
+Bottlerocket uses a neat feature of the Linux kernel called [Boot Configuration](https://www.kernel.org/doc/html/latest/admin-guide/bootconfig.html), which allows a user to pass additional arguments to the kernel command line at runtime.
```
bcressey:
```suggestion
Bottlerocket uses a feature of the Linux kernel called [Boot Configuration](https://www.kernel.org/doc/html/latest/admin-guide/bootconfig.html), which allows a user to pass additional arguments to the kernel command line at runtime.
```

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+The most immediate use of this feature for most users is setting `console` settings so boot messages can be seen on the appropriate consoles.
+
+In order to make use of this feature, one must create an initrd with the desired settings encoded inside it.
```
bcressey: :boromir-meme:
> One does not simply create an initrd with the desired settings encoded inside it.

(should avoid phrasing like this)

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+(Bottlerocket technically doesn't use an initrd, so the file is empty save for the encoded boot config data)
```
bcressey: This is confusing - we say to create an initrd and then say that Bottlerocket doesn't need it.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+To create the initrd, you must first create a configuration file containing key value pairs for the settings you would like to pass to kernel / init.
+Full syntax is described in the [Boot Config documentation](https://www.kernel.org/doc/html/latest/admin-guide/bootconfig.html#config-file-syntax), but a simple example is provided below that shows the format of console settings as well as an example `systemd` parameter.
```
bcressey, line 133:
```suggestion
The full syntax is described in the [Boot Config documentation](https://www.kernel.org/doc/html/latest/admin-guide/bootconfig.html#config-file-syntax), but a simple example is provided below that shows the format of console settings as well as an example `systemd` parameter.
```

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+
+The two acceptable prefixes to settings are `kernel` and `init`.
+Settings prefixed with `kernel` are added to the beginning of the kernel command line.
+Settings prefixed with `init` are added to the kernel command line after the `--`, but before any existing init parameters.
+
+In the example below, two console devices are set up, and `systemd`'s log level is set to `debug`.
+It is important to note that the kernel will output to both console devices, but `systemd` outputs to the last device in line (`ttyS1,115200n8` in the example below).
```
bcressey: Maybe add this note below, although it feels a bit discursive.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,169 @@
...
+}
+```
+
+Bottlerocket provides a command to create the Boot Configuration initrd from your config file, but there are a few prerequisites.
+You will need the [Bottlerocket code repo](https://github.com/bottlerocket-os/bottlerocket), Rust (via [rustup](https://rustup.rs/) or the official site), Docker (20.10.10 or later), and [cargo-make](https://sagiegurari.github.io/cargo-make/) (`cargo install cargo-make`).
+To build the Boot Configuration initrd, create a file called `bootconfig-input` in the root of the Bottlerocket repo containing your desired key/value pair kernel and init arguments.
+Then run:
+```
+cargo make boot-config
+```
```
bcressey: Is there a `docker run` command using the SDK we could provide instead? Otherwise the requirements are quite involved.

README.md
bcressey: I would move the provisioning note to its own sentence - it's much harder to jump into.

Something like: "If you're interested in running Bottlerocket on bare metal servers, please refer to the provisioning guide to get started."

PROVISIONING-METAL.md
zmrow: I'll remove the "simply"

HARDWARE.md
stockholmux: @bcressey I'd like to avoid introduce anything new with the concept of 'community' involved. It creates an artificial line.

I would suggest 'confirmed support' and 'reported support'.

I also like markdown files in the repo because there is a consistent contribution and approval model. For example, if there was some small change that enabled XYZ hardware to work, one PR would handle the change and the update to the support. That and we _want_ folks to contribute even if it's just modifying some markdown.

PROVISIONING-METAL.md
stockholmux: Thanks @bcressey! I do have opinions in this - back in my previous organization [I wrote an explainer on the use of 'we'](https://gist.github.com/stockholmux/52b9b224ad806acd0dd389fe3e79daa0). 

In the case of above you're use 'we' to refer to the logic in the software, which is actually a separate issue: software anthropomorphism. It's usually easy to fix - anthropomorphism is unneeded and switch to the explicit name: 

> If none of the interfaces has `primary` set, we choose the first interface in the file.

becomes

> If none of the interfaces has `primary` set, `kubelet` will use the first interface in the file.

zmrow:
^ Address all comments

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,187 @@
...
+}
+init {
+    systemd.log_level = debug
+}
+```
+
+The Bottlerocket SDK provides the `bootconfig` CLI tool, which is used to create a Boot Configuration initrd.
+To create the Boot Configuration initrd, create a config file named `bootconfig-input` containing your desired key/value pair kernel and init arguments.
+
+Then run the following (you will need Docker installed):
```
etungsten, line 154: Should we also mention the related `cargo make` tasks as another option in addition to the `docker run` commands?

etungsten (approved):

zmrow:
^ Fix the HP model number in the supported hardware table

zmrow:
^ I made a mistake with the HP model number, it's correct now.

PROVISIONING-METAL.md
```diff
@@ -0,0 +1,187 @@
...
+sha512sum -c <<<"e9b1ea5f9b4f95c9b55edada4238bf00b12845aa98bdd2d3edb63ff82a03ada19444546337ec6d6806cbf329027cf49f7fde31f54d551c5e02acbed7efe75785  root.json"
+```
+
+Next, set your desired version and variant, and use `tuftool` to download the image:
+To install `tuftool` you'll need to install Rust (via [rustup](https://rustup.rs/) or the official site), and then you can run `cargo install tuftool`.
+```
+VERSION="1.8.0"
+VARIANT="metal-k8s-1.23"
+COMMIT="a6233c22"
+IMAGE="bottlerocket-${VARIANT}-x86_64-${VERSION}-${COMMIT}.img.lz4"
```
bcressey: nit: maybe add an `ARCH` variable as a bit of future-proofing?

README.md
```diff
@@ -70,6 +71,12 @@ We also have variants that are designed to be Kubernetes worker nodes in VMware:
 - `vmware-k8s-1.22`
 - `vmware-k8s-1.23`
 
+The following variants are designed to be Kubernetes worker nodes on bare metal:
+
+- `metal-k8s-1.21`
+- `metal-k8s-1.22`
+- `metal-k8s-1.23`
+
 The following variants are no longer supported.:
```
bcressey: pre-existing issue:
```suggestion
The following variants are no longer supported:
```

SUPPORTED-HARDWARE.md
```diff
@@ -0,0 +1,25 @@
+# Supported hardware for Bottlerocket on bare metal
+
+This document captures the hardware Bottlerocket for bare metal is tested on and known to work with.
+It also contains hardware the community has reported to be functional.
```
bcressey: nit:
```suggestion
It also contains hardware that contributors have reported to be functional.
```

bcressey (approved):

zmrow:
^ Addresses the last few nits

PR #2219, opened 2022-06-15
kubelet: add image GC threshold settings

### LINKED ISSUE #2065: Support for kubelet config options: imageGCHighThresholdPercent, imageGCLowThresholdPercent
**What I'd like:**

We were running into an issue on our development EKS cluster when we tried to update multiple deployments trough Argo CD on the same EKS node. The node was hitting the Condition `DiskPressure` for one to two minutes on each parallel deployment, before it reverted back to reporting `NoDiskPressure`. We tried to track the issue down and could identify the issue with the following timeline:

* New images (read: > 20 images at approx the same time) are pushed trough CI/CD to different ECR repos
* Argo CD image updater detects these new images and changes the deployment image tags in each deployment
* Kubernetes tries to schedule some of these pods on a specific node
* Node is currently sitting at around 82-84% of free disk space
* Kubelet tries to download the new images (each of them has approx. 150-200 megabyte) 
* The Kubelet garbage collector kicks in and tries to delete images, since it hit the 85% (default) threshold. It will take a few minutes to delete all of them 
* New images are downloaded in parallel, tipping the free disk space > 90% and the hard eviction threshold is met
* Node changes state into `DiskPressure` not scheduling any new nodes
* Kubelet garbage collector finishes deletion of images (after a few minutes)
* Node changes back into `NoDiskPressure`

This garbage collect interval could be changed via the kubelet configuration arguments. Unfortunately currently this is not supported by Bottlerocket. The following options should be able to configure via bootstrap arguments:
```
imageGCHighThresholdPercent: xx
imageGCLowThresholdPercent: xx
```

AWS itself allows this option on their EKS optimised instances: https://aws.amazon.com/premiumsupport/knowledge-center/eks-worker-nodes-image-cache/

We had to decrease the hard eviction limits from the defaults to lower values in order to not run into this condition:
```
"nodefs.available" = "5%"
"imagefs.available" = "10%"
```

This is not really optimal, since it makes the whole node more likely to get close to the available disk space.

### LINKED ISSUE #2216: Add imageGC settings to all k8s variants’ models
Add imageGCHighThresholdPercent and imageGCLowThresholdPercent to all k8s variants’ models

Required for: #2065

### LINKED ISSUE #2217: Add migrations required for imageGC settings to all k8s variants’ models
Add migrations required for the imageGCHighThresholdPercent and imageGCLowThresholdPercent settings to all k8s variants’ models

Required for: #2065 
Blocked by: #2216

mchaker:
**Issue number:**
Closes #2216 
Closes #2217 
Closes #2065

**Description of changes:**
Adds two kubelet settings: `imageGCHighThresholdPercent` and `imageGCLowThresholdPercent`.
Also adds validation for those two settings.

From the kubernetes repo (https://pkg.go.dev/k8s.io/kubernetes/pkg/kubelet/apis/config#KubeletConfiguration):

```go
// imageGCHighThresholdPercent is the percent of disk usage after which
// image garbage collection is always run. The percent is calculated as
// this field value out of 100.

// imageGCLowThresholdPercent is the percent of disk usage before which
// image garbage collection is never run. Lowest disk usage to garbage
// collect to. The percent is calculated as this field value out of 100.
```

~~**Some Q&A:**~~

~~1. _Why validate twice: at the apiserver level, and the template rendering level?_~~
    ~~1. There are two kinds of invalidity for the settings this PR adds: independent invalidity and _interdependent invalidity_. Independent invalidity would be a value outside of the defined 0-100 threshold. _Interdependent invalidity_ is when the `High` threshold percent is lower than the `Low` threshold percent, and is only calculable by comparing (or knowing) both settings at the same time.~~
    ~~1. The independent invalidity can and should be checked as soon as possible and stop the user from entering invalid values. That kind of validation can and is done at the apiserver level.~~
    ~~1. The _interdependent_ invalidity is only able to be checked when the template is about to be rendered -- when _both_ setting values are known (since the user can provide only one of the two new settings in an apiserver request, and these settings also have default kubelet values to initially compare against).~~
~~1. _Why does your template code contain `~` characters at each end of the helper function call?_~~
    ~~1. That is a [handlebars](https://docs.rs/handlebars/latest/handlebars/index.html) feature that removes whitespace (including newlines) before or after that `~` character.~~
    ~~1. In the case of this PR, the helper function `image_gc_threshold_percent` actually handles the rendering of necessary whitespace after writing out the setting values (specifically: the newlines after each setting), so the template rendering engine/handlebars should remove the existing newline it sees at the end of the helper function call (after the `}}` at the end of the line). Therefore, the `~` feature of handlebars was used.~~

**Testing done:**
- [x] Upgrade/downgrade for variant: aws-k8s-1.22
- ~[ ] Upgrade/downgrade for variant: metal 1.22~
- [x] Rendering and setting updates for variant: vmware 1.22
- [x] Manual testing of valid and invalid values, in varying order, on a bottlerocket aws-k8s-1.22 node in a k8s 1.22 cluster

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,8 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.topology-manager-policy`: Specifies the topology manager policy. Possible values are `none`, `restricted`, `best-effort`, and `single-numa-node`. Defaults to `none`.
 * `settings.kubernetes.topology-manager-scope`: Specifies the topology manager scope. Possible values are `container` and `pod`. Defaults to `container`. If you want to group all containers in a pod to a common set of NUMA nodes, you can set this setting to `pod`.
 * `settings.kubernetes.pod-pids-limit`: The maximum number of processes per pod.
+* `settings.kubernetes.image-gc-high-threshold-percent`: The percent of disk usage after which image garbage collection is always run.
+* `settings.kubernetes.image-gc-low-threshold-percent`: The percent of disk usage before which image garbage collection is never run.
 * `settings.kubernetes.provider-id`: This sets the unique ID of the instance that an external provider (i.e. cloudprovider) can use to identify a specific node.
 
 You can also optionally specify static pods for your node with the following settings.
--- a/packages/kubernetes-1.19/kubelet-config
+++ b/packages/kubernetes-1.19/kubelet-config
@@ -93,6 +93,12 @@ topologyManagerPolicy: {{settings.kubernetes.topology-manager-policy}}
 {{#if settings.kubernetes.pod-pids-limit includeZero=true}}
 podPidsLimit: {{settings.kubernetes.pod-pids-limit}}
 {{/if}}
+{{#if settings.kubernetes.image-gc-high-threshold-percent includeZero=true}}
+imageGCHighThresholdPercent: {{settings.kubernetes.image-gc-high-threshold-percent}}
+{{/if}}
+{{#if settings.kubernetes.image-gc-low-threshold-percent includeZero=true}}
+imageGCLowThresholdPercent: {{settings.kubernetes.image-gc-low-threshold-percent}}
+{{/if}}
 {{#if settings.kubernetes.provider-id}}
 providerID: {{settings.kubernetes.provider-id}}
 {{/if}}
--- a/sources/api/migration/migrations/v1.9.0/image-gc-thresholds/Cargo.toml
+++ b/sources/api/migration/migrations/v1.9.0/image-gc-thresholds/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "image-gc-thresholds"
+version = "0.1.0"
+edition = "2018"
+authors = ["Mahdi Chaker <mmchaker@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers", version = "0.1.0"}
--- a/sources/models/src/modeled_types/kubernetes.rs
+++ b/sources/models/src/modeled_types/kubernetes.rs
@@ -11,6 +11,10 @@ use std::fmt;
 use std::net::IpAddr;
 use std::ops::Deref;
 
+// Declare constant values usable by any type
+const IMAGE_GC_THRESHOLD_MAX: i32 = 100;
+const IMAGE_GC_THRESHOLD_MIN: i32 = 0;
+
 /// KubernetesName represents a string that contains a valid Kubernetes resource name.  It stores
 /// the original string and makes it accessible through standard traits.
 // https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
@@ -997,6 +1001,145 @@ mod test_topology_manager_policy {
         for err in &["", "bad", "100", &"a".repeat(64)] {
             TopologyManagerPolicy::try_from(*err).unwrap_err();
         }
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// imageGCHighThresholdPercent is the percent of disk usage after which image
+/// garbage collection is always run. The percent is calculated by dividing this
+/// field value by 100, so this field must be between 0 and 100, inclusive. When
+/// specified, the value must be greater than imageGCLowThresholdPercent.
+/// Default: 85
+/// https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ImageGCHighThresholdPercent {
+    inner: String,
+}
+
+impl TryFrom<&str> for ImageGCHighThresholdPercent {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        let parsed_input: i32 = input
+            .parse::<i32>()
+            .context(error::ParseIntSnafu { input })?;
+        ensure!(
+            !input.is_empty(),
+            error::InvalidImageGCHighThresholdPercentSnafu {
+                input,
+                msg: "must not be empty",
+            }
+        );
+        ensure!(
+            (IMAGE_GC_THRESHOLD_MIN..=IMAGE_GC_THRESHOLD_MAX).contains(&parsed_input),
+            error::InvalidImageGCHighThresholdPercentSnafu {
+                input,
+                msg: "must be between 0 and 100 (inclusive)"
+            }
+        );
+
+        Ok(ImageGCHighThresholdPercent {
+            inner: input.to_owned(),
+        })
+    }
+}
+string_impls_for!(ImageGCHighThresholdPercent, "ImageGCHighThresholdPercent");
+
+#[cfg(test)]
+mod test_image_gc_high_threshold_percent {
+    use super::ImageGCHighThresholdPercent;
+    use std::convert::TryFrom;
+
+    // test 1: good values should succeed
+    #[test]
+    fn image_gc_high_threshold_percent_between_0_and_100_inclusive() {
+        for ok in &["0", "1", "99", "100"] {
+            ImageGCHighThresholdPercent::try_from(*ok).unwrap();
+        }
+    }
+
+    // test 2: values too low should return Errors
+    #[test]
+    fn image_gc_high_threshold_percent_less_than_0_fails() {
+        ImageGCHighThresholdPercent::try_from("-1").unwrap_err();
+    }
+
+    // test 3: values too high should return Errors
+    #[test]
+    fn image_gc_high_threshold_percent_greater_than_100_fails() {
+        ImageGCHighThresholdPercent::try_from("101").unwrap_err();
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// imageGCLowThresholdPercent is the percent of disk usage before which image
+/// garbage collection is never run. Lowest disk usage to garbage collect to.
+/// The percent is calculated by dividing this field value by 100, so the field
+/// value must be between 0 and 100, inclusive. When specified, the value must
+/// be less than imageGCHighThresholdPercent.
+/// Default: 80
+/// https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ImageGCLowThresholdPercent {
+    inner: String,
+}
+
+impl TryFrom<&str> for ImageGCLowThresholdPercent {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        let parsed_input: i32 = input
+            .parse::<i32>()
+            .context(error::ParseIntSnafu { input })?;
+        ensure!(
+            !input.is_empty(),
+            error::InvalidImageGCLowThresholdPercentSnafu {
+                input,
+                msg: "must not be empty",
+            }
+        );
+        ensure!(
+            (IMAGE_GC_THRESHOLD_MIN..=IMAGE_GC_THRESHOLD_MAX).contains(&parsed_input),
+            error::InvalidImageGCLowThresholdPercentSnafu {
+                input,
+                msg: "must be between 0 and 100 (inclusive)"
+            }
+        );
+
+        Ok(ImageGCLowThresholdPercent {
+            inner: input.to_owned(),
+        })
+    }
+}
+string_impls_for!(ImageGCLowThresholdPercent, "ImageGCLowThresholdPercent");
+
+#[cfg(test)]
+mod test_image_gc_low_threshold_percent {
+    use super::ImageGCLowThresholdPercent;
+    use std::convert::TryFrom;
+
+    // test 1: good values should succeed
+    #[test]
+    fn image_gc_low_threshold_percent_between_0_and_100_inclusive() {
+        for ok in &["0", "1", "99", "100"] {
+            ImageGCLowThresholdPercent::try_from(*ok).unwrap();
+        }
+    }
+
+    // test 2: values too low should return Errors
+    #[test]
+    fn image_gc_low_threshold_percent_less_than_0_fails() {
+        ImageGCLowThresholdPercent::try_from("-1").unwrap_err();
+    }
+
+    // test 3: values too high should return Errors
+    #[test]
+    fn image_gc_low_threshold_percent_greater_than_100_fails() {
+        ImageGCLowThresholdPercent::try_from("101").unwrap_err();
     }
 }
 
--- a/Release.toml
+++ b/Release.toml
@@ -134,4 +134,5 @@ version = "1.8.0"
     "migrate_v1.9.0_ntp-affected-services.lz4",
     "migrate_v1.9.0_shibaken-admin-userdata-semantics.lz4",
     "migrate_v1.9.0_shibaken-send-metrics.lz4",
+    "migrate_v1.9.0_image-gc-thresholds.lz4",
 ]
--- a/packages/kubernetes-1.20/kubelet-config
+++ b/packages/kubernetes-1.20/kubelet-config
@@ -93,6 +93,12 @@ topologyManagerPolicy: {{settings.kubernetes.topology-manager-policy}}
 {{#if settings.kubernetes.pod-pids-limit includeZero=true}}
 podPidsLimit: {{settings.kubernetes.pod-pids-limit}}
 {{/if}}
+{{#if settings.kubernetes.image-gc-high-threshold-percent includeZero=true}}
+imageGCHighThresholdPercent: {{settings.kubernetes.image-gc-high-threshold-percent}}
+{{/if}}
+{{#if settings.kubernetes.image-gc-low-threshold-percent includeZero=true}}
+imageGCLowThresholdPercent: {{settings.kubernetes.image-gc-low-threshold-percent}}
+{{/if}}
 {{#if settings.kubernetes.provider-id}}
 providerID: {{settings.kubernetes.provider-id}}
 {{/if}}
--- a/packages/kubernetes-1.21/kubelet-config
+++ b/packages/kubernetes-1.21/kubelet-config
@@ -93,6 +93,12 @@ topologyManagerPolicy: {{settings.kubernetes.topology-manager-policy}}
 {{#if settings.kubernetes.pod-pids-limit includeZero=true}}
 podPidsLimit: {{settings.kubernetes.pod-pids-limit}}
 {{/if}}
+{{#if settings.kubernetes.image-gc-high-threshold-percent includeZero=true}}
+imageGCHighThresholdPercent: {{settings.kubernetes.image-gc-high-threshold-percent}}
+{{/if}}
+{{#if settings.kubernetes.image-gc-low-threshold-percent includeZero=true}}
+imageGCLowThresholdPercent: {{settings.kubernetes.image-gc-low-threshold-percent}}
+{{/if}}
 {{#if settings.kubernetes.provider-id}}
 providerID: {{settings.kubernetes.provider-id}}
 {{/if}}
--- a/packages/kubernetes-1.22/kubelet-config
+++ b/packages/kubernetes-1.22/kubelet-config
@@ -93,6 +93,12 @@ topologyManagerPolicy: {{settings.kubernetes.topology-manager-policy}}
 {{#if settings.kubernetes.pod-pids-limit includeZero=true}}
 podPidsLimit: {{settings.kubernetes.pod-pids-limit}}
 {{/if}}
+{{#if settings.kubernetes.image-gc-high-threshold-percent includeZero=true}}
+imageGCHighThresholdPercent: {{settings.kubernetes.image-gc-high-threshold-percent}}
+{{/if}}
+{{#if settings.kubernetes.image-gc-low-threshold-percent includeZero=true}}
+imageGCLowThresholdPercent: {{settings.kubernetes.image-gc-low-threshold-percent}}
+{{/if}}
 {{#if settings.kubernetes.provider-id}}
 providerID: {{settings.kubernetes.provider-id}}
 {{/if}}
--- a/packages/kubernetes-1.23/kubelet-config
+++ b/packages/kubernetes-1.23/kubelet-config
@@ -93,6 +93,12 @@ topologyManagerPolicy: {{settings.kubernetes.topology-manager-policy}}
 {{#if settings.kubernetes.pod-pids-limit includeZero=true}}
 podPidsLimit: {{settings.kubernetes.pod-pids-limit}}
 {{/if}}
+{{#if settings.kubernetes.image-gc-high-threshold-percent includeZero=true}}
+imageGCHighThresholdPercent: {{settings.kubernetes.image-gc-high-threshold-percent}}
+{{/if}}
+{{#if settings.kubernetes.image-gc-low-threshold-percent includeZero=true}}
+imageGCLowThresholdPercent: {{settings.kubernetes.image-gc-low-threshold-percent}}
+{{/if}}
 {{#if settings.kubernetes.provider-id}}
 providerID: {{settings.kubernetes.provider-id}}
 {{/if}}
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -1744,6 +1744,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "image-gc-thresholds"
+version = "0.1.0"
+dependencies = [
+ "migration-helpers",
+]
+
+[[package]]
 name = "imdsclient"
 version = "0.1.0"
 dependencies = [
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -64,6 +64,7 @@ members = [
     "api/migration/migrations/v1.9.0/ntp-affected-services",
     "api/migration/migrations/v1.9.0/shibaken-admin-userdata-semantics",
     "api/migration/migrations/v1.9.0/shibaken-send-metrics",
+    "api/migration/migrations/v1.9.0/image-gc-thresholds",
 
     "bottlerocket-release",
 
--- a/sources/api/migration/migrations/v1.9.0/image-gc-thresholds/src/main.rs
+++ b/sources/api/migration/migrations/v1.9.0/image-gc-thresholds/src/main.rs
@@ -1,0 +1,26 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddSettingsMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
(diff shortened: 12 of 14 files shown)


packages/kubernetes-1.19/kubelet-config
```diff
@@ -93,6 +93,12 @@ topologyManagerPolicy: {{settings.kubernetes.topology-manager-policy}}
 {{#if settings.kubernetes.pod-pids-limit includeZero=true}}
 podPidsLimit: {{settings.kubernetes.pod-pids-limit}}
 {{/if}}
+{{#if settings.kubernetes.image-gc-high-threshold-percent includeZero=true}}
+imageGCHighThresholdPercent: {{default "85" settings.kubernetes.image-gc-high-threshold-percent}}
+{{/if}}
+{{#if settings.kubernetes.image-gc-low-threshold-percent includeZero=true}}
+imageGCLowThresholdPercent: {{default "80" settings.kubernetes.image-gc-low-threshold-percent}}
+{{/if}}
```
bcressey: We need to enforce the [validation rules](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/) somehow:
>  imageGCHighThresholdPercent is the percent of disk usage after which image garbage collection is always run. The percent is calculated by dividing this field value by 100, so this field must be between 0 and 100, inclusive. When specified, the value must be greater than imageGCLowThresholdPercent. 

> imageGCLowThresholdPercent is the percent of disk usage before which image garbage collection is never run. The percent is calculated by dividing this field value by 100, so the field value must be between 0 and 100, inclusive. When specified, the value must be less than imageGCHighThresholdPercent. 

In the API, we should reject values outside of `0 <= n <= 100`, instead of accepting any `i32`.

Here in the template we need a way to ensure that we don't render an illegal configuration, which will mean comparing the values for the two settings.

sources/api/migration/migrations/v1.9.0/image-gc-thresholds/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2018"
+authors = ["Mahdi Chaker <mmchaker@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers", version = "0.1.0"}
```
mchaker, line 12: I'm not sure if any values here should be customized. 
E.g.: should 2018 be 2022?

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1,6 +1,7 @@
 use lazy_static::lazy_static;
 use regex::Regex;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use x509_parser::nom::ParseTo;
```
mchaker: This was unintentional. Removing.

mchaker:
_Latest force push: consolidate branch history. Summary of changes below._

# apiserver and schnauzer: Add validation for new kubelet settings `imageGCHighThresholdPercent` and `imageGCLowThresholdPercent`.

## Description

- apiserver: add new setting types to the model with basic apiserver-level validation
- schnauzer: add template rules and more in-depth validation for new settings

### What settings?

in the `kubelet-config` template:

- `imageGCHighThresholdPercent` and `imageGCLowThresholdPercent`

## How and why

Add value validation to _both_ the apiserver and template (via schnauzer), since these settings are _interdependently validated_. Meaning, since the `High` Image GC Threshold Percent must be _greater than_ the `Low` Image GC Threshold Percent, both settings must be checked against each other. That is not possible at the apiserver level, or at least is much easier to do in the template rendering level (once all the values from the datastore are ready to be written).

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1002,6 +1006,183 @@ mod test_topology_manager_policy {
...
+        ensure!(
+            !input.is_empty(),
+            error::InvalidImageGCHighThresholdPercentSnafu {
+                input,
+                msg: "must not be empty",
+            }
+        );
+        // apiserver validation 2: reject values that are too low to be a percentage
+        ensure!(
+            i32::ge(&input.parse::<i32>().unwrap(), &IMAGE_GC_THRESHOLD_MIN),
```
mchaker: Would this be clearer as 
```rust
input.parse::<i32>().unwrap() > IMAGE_GC_THRESHOLD_MIN
```
?

cbgbt: Yes, but:

You should avoid using `unwrap()` or `expect()` in almost all cases except for tests, instead transforming an error in parsing into an error with the parent function. Ultimately this would look like;

```rust
let parsed_input: i32 = input.parse()
    .context(error::InvalidImageGCHighThresholdPercentSnafu { input, msg: "Could not parse input as integer." })?;

ensure!(
    parsed_input > Image_GC_THRESHOLD_MIN,
    // etc
    )
```

Basically you should never panic, instead deciding if you want to handle an error in some way, or bubble it up as a Result to the caller.

packages/kubernetes-1.19/kubelet-config
```diff
@@ -93,6 +93,7 @@ topologyManagerPolicy: {{settings.kubernetes.topology-manager-policy}}
 {{#if settings.kubernetes.pod-pids-limit includeZero=true}}
 podPidsLimit: {{settings.kubernetes.pod-pids-limit}}
 {{/if}}
+{{~image_gc_threshold_percent settings.kubernetes.image-gc-high-threshold-percent settings.kubernetes.image-gc-low-threshold-percent~}}
```
bcressey: It's a bit weird and inconsistent to move the entire generation into a helper. One downside is that it's no longer obvious which settings are supported - the reader could guess that `image_gc_threshold_percent` sets the corresponding settings, but the only way to know for sure is to dig into that code.

We also don't necessarily need to include "percent" in the settings name, if it's documented and defined as taking a percentage value.

```suggestion
{{#if image_gc_threshold low settings.kubernetes.image-gc-high settings.kubernetes.image-gc-low-threshold}}
imageGCLowThresholdPercent: {{image_gc_threshold low settings.kubernetes.image-gc-high settings.kubernetes.image-gc-low-threshold}}
{{/if}}
{{#if image_gc_threshold high settings.kubernetes.image-gc-high settings.kubernetes.image-gc-low-threshold}}
imageGCHighThresholdPercent: {{image_gc_threshold high settings.kubernetes.image-gc-high settings.kubernetes.image-gc-low-threshold}}
{{/if}}
```

sources/models/src/modeled_types/kubernetes.rs
mchaker: Update: followed this suggestion (thank you!) and added a new error type for `std::num::ParseIntError`s

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1002,6 +1006,177 @@ mod test_topology_manager_policy {
...
+/// imageGCHighThresholdPercent is the percent of disk usage after which image
+/// garbage collection is always run. The percent is calculated by dividing this
+/// field value by 100, so this field must be between 0 and 100, inclusive. When
+/// specified, the value must be greater than imageGCLowThresholdPercent.
+/// Default: 85
+/// https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ImageGCHighThresholdPercent {
+    // All settings are brought in as Strings.
```
bcressey: This isn't accurate. Models certainly can hold non-string values; `max_pods` is an example.

This particular approach is used to add validation rules on top of a basic type, such as `String`. Historically we added this to enforce constraints on strings, like `SingleLineString`.

It looks like we've used `inner: String` for non-string values elsewhere, but that was probably a mistake. This probably should be `inner: i32` but then we'd need an `i32_impls_for!` macro to make the struct behave like an `i32` after validation.

I won't insist on doing that as part of this PR, but it might be good to backlog an issue for extending this validation framework to types besides `String`.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1002,6 +1006,177 @@ mod test_topology_manager_policy {
...
+            // Fun fact: (IMAGE_GC_THRESHOLD_MIN..=IMAGE_GC_THRESHOLD_MAX) is a range of values in Rust
+            // and you can check if a value is contained in a given range!
+            (IMAGE_GC_THRESHOLD_MIN..=IMAGE_GC_THRESHOLD_MAX).contains(&parsed_input),
+            error::InvalidImageGCHighThresholdPercentSnafu {
+                input,
+                msg: "must be between 0 and 100 (inclusive)"
+            }
+        );
+
+        // When everything goes well, store the input value into the custom type
```
```suggestion
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1002,6 +1006,177 @@ mod test_topology_manager_policy {
...
+                msg: "must not be empty",
+            }
+        );
+        // apiserver validation 2: reject values that are out of bounds
+        let parsed_input: i32 = input
+            .parse::<i32>()
+            .context(error::ParseIntSnafu { input })?;
+        ensure!(
+            // Fun fact: (IMAGE_GC_THRESHOLD_MIN..=IMAGE_GC_THRESHOLD_MAX) is a range of values in Rust
+            // and you can check if a value is contained in a given range!
```
bcressey: These sorts of comments are not ideal. You can assume the reader knows Rust and is familiar with how language constructs work.

```suggestion
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1002,6 +1006,177 @@ mod test_topology_manager_policy {
...
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        // apiserver validation 1: reject empty values
+        ensure!(
+            !input.is_empty(),
+            error::InvalidImageGCHighThresholdPercentSnafu {
+                input,
+                msg: "must not be empty",
+            }
+        );
+        // apiserver validation 2: reject values that are out of bounds
```
bcressey: Both of these comments seem like the "restate what the code is doing" variety. Usually these aren't necessary.

```suggestion
        ensure!(
            !input.is_empty(),
            error::InvalidImageGCHighThresholdPercentSnafu {
                input,
                msg: "must not be empty",
            }
        );
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1002,6 +1006,177 @@ mod test_topology_manager_policy {
...
+
+        // When everything goes well, store the input value into the custom type
+        Ok(ImageGCHighThresholdPercent {
+            inner: input.to_owned(),
+        })
+    }
+}
+// The following macro is necessary for Kubernetes Settings custom types.
+// I learned this the hard way when trying to make ImageGCHighThresholdPercent's
+// inner value an i32 instead.
```
```suggestion
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -106,6 +106,9 @@ const KUBE_RESERVE_ADDITIONAL: f32 = 2.5;
 const IPV4_LOCALHOST: IpAddr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
 const IPV6_LOCALHOST: IpAddr = IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1));
 
+const IMAGE_GC_HIGH_THRESHOLD_DEFAULT: &str = "85";
+const IMAGE_GC_LOW_THRESHOLD_DEFAULT: &str = "80";
```
bcressey: If we conditionally render the block in kubelet's config, we can avoid hardcoding the current default values, which might not stay the same over time.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1170,6 +1185,179 @@ pub fn etc_hosts_entries(
...
+    // To give context to our errors, get the template name (e.g. what file we are rendering), if available.
+    trace!("Starting image_gc_threshold_percent helper");
+    let template_name = template_name(renderctx);
+    trace!("Template name: {}", &template_name);
+
+    // Check number of parameters, must be exactly two (high and low GC threshold percentages)
+    trace!("Number of params: {}", helper.params().len());
+    check_param_count(helper, template_name, 2)?;
+
+    // ----- Get the GC Threshold Percentage values out of the template.
```
bcressey: Is this meant to be a "start of section" comment? It's probably not worth trying to convey that with a few leading `-`s.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1170,6 +1185,179 @@ pub fn etc_hosts_entries(
...
+    // Get the GC High Threshold Percent from the template
+    let image_gc_high_threshold_percent_param = get_param(helper, 0)?;
+    // Check if the GC High Threshold Percent was not provided by the user (e.g.
+    // when bottlerocket first boots, the setting is not explicitly given, but
```
bcressey:
```suggestion
    // when Bottlerocket first boots, the setting is not explicitly given, but
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1170,6 +1185,179 @@ pub fn etc_hosts_entries(
...
+                image_gc_low_threshold_percent,
+                image_gc_high_threshold_percent,
+            );
+            return Err(RenderError::from(
+                error::TemplateHelperError::InvalidTemplateValueCustomMessage {
+                    message: "rule is (imageGCLowThresholdPercent < imageGCHighThresholdPercent) but got (imageGCLowThresholdPercent >= imageGCHighThresholdPercent)",
+                    template: template_name.to_owned(),
+                },
+            ));
+        }
```
bcressey: We need to deal with these inputs somehow without erroring out, possibly by omitting one or the other in the rendered config. It's too late to break kubelet startup - we've already accepted these values as valid settings.

What does `kubelet` do if you pass it a "high" value that's lower than its default "low" value?

bcressey (changes requested):

sources/api/schnauzer/src/helpers.rs
mchaker: What `kubelet` does if it encounters an invalid value pair ("high" < "low"): append an error to a slice of errors with a message explaining the rule that was violated: https://github.com/kubernetes/kubernetes/blob/b2ed6ca64f55452cdb4063d62920dddccd4344f7/pkg/kubelet/apis/config/validation/validation.go#L82-L83

etungsten: My suggestion would be to combine both `imageGCHighThresholdPercent` and `imageGCLowThresholdPercent` under a single modeled type `ImageGCThresholdPercent` and write a [custom deserialize function](https://serde.rs/field-attrs.html#deserialize_with) for it. In the custom function, you can then check during deserialization to make sure the configuration is valid, e.g ("high" < "low"), and return an error if it's not before the setting is accepted.

See https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/src/de.rs for examples. You would then under `struct KubernetesSettings` do something like:
```
#[model]
struct ImageGCThresholdPercent {
      #[serde(
          alias = "high",
          default,
          skip_serializing_if = "Option::is_none"
      )]
      image_gc_high_threshold_percent: ImageGCHighThresholdPercent,
      #[serde(
          alias = "low",
          default,
          skip_serializing_if = "Option::is_none"
      )]
}

struct KubernetesSettings {
    ...
    #[serde(
        default,
        skip_serializing_if = "Option::is_none",
        deserialize_with = "deserialize_image_gc_threshold_percent" // or maybe call it validate_image_gc_threshold_percent
    )]
    image_gc_threshold_percent: ImageGCThresholdPercent,
```

mchaker: Thank you so much @etungsten ! I didn't know you could use decorators that way -- very interesting. I'll do that!

mchaker:
_Latest force push: remove setting validation from template rendering layer. Remove helper function in template rendering layer. Remove interdependent validation entirely, but keep independent validation. Clarify kubelet config template (leave setting names and defaults in the template, instead of in Rust code)._

sources/api/schnauzer/src/helpers.rs
mchaker: Update: After discussions, I have decided to go with a much simpler approach: removing interdependent validation entirely, and allowing kubelet to perform interdependent validation (and output any resulting errors).

packages/kubernetes-1.19/kubelet-config
mchaker: I have simplified the template and removed validation from the template rendering layer (as well as the helper function entirely).

sources/models/src/modeled_types/kubernetes.rs
mchaker: I have opened issue #2257 and my updated changes will still use `inner: String` (until #2257 is implemented).

sources/models/src/modeled_types/kubernetes.rs
mchaker: Agreed and removed.

mchaker:
_Latest force push: remove remaining unused constants._

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1002,6 +1006,163 @@ mod test_topology_manager_policy {
...
+            ImageGCHighThresholdPercent::try_from(*ok).unwrap();
+        }
+    }
+
+    // test 2: values too low should return Errors
+    #[test]
+    fn image_gc_high_threshold_percent_less_than_0_fails() {
+        for err in &["-1"] {
+            ImageGCHighThresholdPercent::try_from(*err).unwrap_err();
+        }
```
arnaldo2792: A `for` might be over-killing for a single test case, same for similar test cases

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1002,6 +1006,163 @@ mod test_topology_manager_policy {
...
+/// Default: 85
+/// https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ImageGCHighThresholdPercent {
+    // Important note: custom types currently need to have `string_impls_for!()`
+    // after their TryFrom<&str> implementation. (this requires that the inner
+    // type of a custom Kubernetes settings type is a String).
+    // Ideally, `string_impls_for!()` would be accompanied by `i32_impls_for!()`
+    // and macros for other `inner` types.
```
arnaldo2792: I don't know if this comment is relevant since there is a GH issue for the problem described. Maybe just add the link for the issue here and remove it when the problem is fixed?

arnaldo2792 (changes requested):

sources/models/src/modeled_types/kubernetes.rs
mchaker: Oops, meant to remove this in my latest change. Thanks!

packages/kubernetes-1.19/kubelet-config
```diff
@@ -93,6 +93,12 @@ topologyManagerPolicy: {{settings.kubernetes.topology-manager-policy}}
 {{#if settings.kubernetes.pod-pids-limit includeZero=true}}
```
arnaldo2792, line 93: For the lack of a better place to post this comment: please group your changes in your commits, I think you could have just one since all the changes you made work as a unit. But I'm happy with two commits as well as long as the grouping makes sense.

mchaker: Good point - I'll squash this to 1 commit since it does work as one unit.

packages/kubernetes-1.19/kubelet-config
```diff
@@ -93,6 +93,12 @@ topologyManagerPolicy: {{settings.kubernetes.topology-manager-policy}}
...
 podPidsLimit: {{settings.kubernetes.pod-pids-limit}}
 {{/if}}
+{{#if settings.kubernetes.image-gc-high-threshold-percent includeZero=true}}
+imageGCHighThresholdPercent: {{default 85 settings.kubernetes.image-gc-high-threshold-percent}}
+{{/if}}
+{{#if settings.kubernetes.image-gc-low-threshold-percent includeZero=true}}
+imageGCLowThresholdPercent: {{default 80 settings.kubernetes.image-gc-low-threshold-percent}}
+{{/if}}
```
bcressey: Do we need to use the default helper if we're testing that the value is defined?

I can see the value in showing the default values in this case, so it's fine to use the helper and drop the "if" checks - just pick one or the other.

mchaker: I'll choose keeping the `{{#if}}` blocks to stick to convention and clarity -- also removing hardcoded default values (in case kubelet changes from 85/80 high/low)

mchaker:
_Latest force push: clean up overkill tests (remove unnecessary `for` loop), remove unnecessary comments, squashed commits to 1 commit, and chose to keep the `{{#if}}` path of settings validation instead of also using the `default` helper function_

mchaker:
_Latest force push: fix commit message_

bcressey (approved):

bcressey:
Code change LGTM.

For writing commit messages, I tend to roughly follow this guide with some minor changes:
https://cbea.ms/git-commit/

Minor changes:
* I don't capitalize the subject
* I start with a "subsystem: blah" tag unless the change is very cross-cutting

So for this commit I would say:
```
kubelet: add image GC threshold settings
```

And combine your current subject and body into a new body wrapped at 70 characters.

mchaker:
> Code change LGTM.
> 
> For writing commit messages, I tend to roughly follow this guide with some minor changes: https://cbea.ms/git-commit/
> 
> Minor changes:
> 
> * I don't capitalize the subject
> * I start with a "subsystem: blah" tag unless the change is very cross-cutting
> 
> So for this commit I would say:
> 
> ```
> kubelet: add image GC threshold settings
> ```
> 
> And combine your current subject and body into a new body wrapped at 70 characters.

Thank you @bcressey ! I've clarified the commit message and PR title.

arnaldo2792 (approved):

sources/api/migration/migrations/v1.9.0/image-gc-thresholds/Cargo.toml
mchaker: Update: 2018 should stay 2018 -- it's a Rust thing (ecosystem version)

bcressey (approved):

sources/api/schnauzer/src/helpers.rs
```diff
@@ -160,6 +160,12 @@ mod error {
             template: String,
         },
 
+        #[snafu(display("Invalid template value, {} in template {}", message, template))]
+        InvalidTemplateValueCustomMessage {
+            message: &'static str,
+            template: String,
+        },
+
```
etungsten: Is this enum variant being used anywhere? If it's not being used, we should remove it.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -244,6 +250,12 @@ mod error {
             target: String,
         },
 
+        #[snafu(display("Failed to parse {} to integer: {}", what, source))]
+        ParseInt {
+            what: String,
+            source: std::num::ParseIntError,
+        },
+
```
etungsten: Same as above. I don't see this being used in schnauzer anywhere.

etungsten (changes requested):
Can you please update the PR description to reflect the latest state of the PR? Some parts of it is out of date.

sources/api/schnauzer/src/helpers.rs
mchaker: Correct, this one is not used -- it is a remnant of a previous approach to the PR (pre-latest-refactor).
Removing..

sources/api/schnauzer/src/helpers.rs
mchaker: ~~Good point - this snafu is not used directly in schnauzer, but is used in `kubernetes.rs`, [L1027](https://github.com/bottlerocket-os/bottlerocket/pull/2219/files#diff-bba9e4e3e46203be2b12f22e0d654ebd270f0b478dd34f40c31d7aa695620f2fR1027) and [L1097](https://github.com/bottlerocket-os/bottlerocket/pull/2219/files#diff-bba9e4e3e46203be2b12f22e0d654ebd270f0b478dd34f40c31d7aa695620f2fR1097).~~

~~Keeping this one.~~

EDIT: It looks like this is actually _not_ useful in schnauzer, but [_is_ useful in `mod.rs`](https://github.com/bottlerocket-os/bottlerocket/pull/2219/files#diff-2907309a5a3d881c962136148bd4340ec869adfa84eae13208d965370543caf4R139).

Removing this definition of the `ParseInt` snafu.

mchaker:
_Last 3 git force-pushes: added `README` documentation for the new kubernetes settings and updated `Release.toml` to include the latest migration._

README.md
```diff
@@ -15,6 +15,7 @@ Bottlerocket-specific additions focus on reliable updates and on the API.
 Instead of making configuration changes manually, you can change settings with an API call, and these changes are automatically migrated through updates.
 
 Some notable features include:
+
```
mchaker: the linter in my editor (vscode) added all these newlines any time I saved the file -- if this is benign, I'll leave them in.

etungsten: Please leave them out of this PR and make a separate commit and PR it separately.

mchaker: Ok thanks, will do

mchaker:
_Latest force push: remove linter formatting-on-save unrelated changes_

mchaker:
_Latest force push: Fixed `README.md` getting over-tuned by my linter. Also recovered my PR from having too many commits squashed into it accidentally._

etungsten (approved):
:ice_cream:

mchaker:
_Latest force push: had to rebase my branch/PR due to merge conflicts with develop/HEAD_

arnaldo2792 (approved):

README.md
```diff
@@ -411,6 +411,8 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.topology-manager-policy`: Specifies the topology manager policy. Possible values are `none`, `restricted`, `best-effort`, and `single-numa-node`. Defaults to `none`.
 * `settings.kubernetes.topology-manager-scope`: Specifies the topology manager scope. Possible values are `container` and `pod`. Defaults to `container`. If you want to group all containers in a pod to a common set of NUMA nodes, you can set this setting to `pod`.
 * `settings.kubernetes.pod-pids-limit`: The maximum number of processes per pod.
+* `settings.kubernetes.image-gc-high-threshold-percent`: The disk utilization percentage where `kubelet` begins garbage collection over the stored container images. In short, this is the "disk is too full" mark.
```
bcressey: This doesn't sound correct; my reading of the documentation is that the garbage collector will run less frequently - not always, but not never - if disk usage percentage falls between high and low.

I would also avoid phrases like "In short". Maybe just strike the second sentence from each since you're introducing an abstract concept - whether or not the disk is too full or empty enough - to what would otherwise be a precise description of how the setting controls the kubelet garbage collector.

mchaker: You are correct. 

EDIT: Will use the following, as per the kubelet config documentation:

```plain
settings.kubernetes.image-gc-high-threshold-percent: the percent of disk usage after which image garbage collection is always run.

settings.kubernetes.image-gc-low-threshold-percent: the percent of disk usage before which image garbage collection is never run.
```

mchaker:
_Latest force push: make README documentation accurate to `kubelet` behavior as per its documentation_

mchaker:
_Latest force push: merge in the latest from `develop`_

PR #2220, opened 2022-06-15
schnauzer: add support for non-string JSON data types

jpculp:
**Description of changes:**

This adds support for using schnauzer for settings other than strings, like conditionally toggling a setting that expects a boolean. Since this isn't particularly common and we want to ensure backwards compatibility, if a type isn't specified it will default to string.

**Example use case:**

If you wanted to enable/disable metricdog barking conditionally by region, rather than a hard-coded value, you could configure your model as such:
```
[metadata.settings.metrics.should-bark]
setting-generator = "schnauzer settings.metrics.should-bark --type bool"
template = "{{ should-bark settings.aws.region }}"
```
Assuming you write a helper in schnauzer for barking to return `"true"` or `"false"`, Bottlerocket would previously fail to apply this to the API as the model is expecting a bool instead of a string. With the proposed changes, schnauzer would instead return a `true` or `false`. The difference is in the quotes!

**Testing done:**

- Unit testing passed.
- Built `aws-k8s-1.21` and checked ec2 console for errors.
- Built `aws-k8s-1.21` with a bool setting and verified it rendered correctly in API.

**Terms of contribution:**

### DIFF
--- a/sources/api/schnauzer/src/main.rs
+++ b/sources/api/schnauzer/src/main.rs
@@ -10,6 +10,10 @@ The setting we're generating is expected to have a metadata key already set: "te
 For example, if we're generating "settings.x" and we have template "foo-{{ settings.bar }}", we look up the value of "settings.bar" in the API.
 If the returned value is "baz", our generated value will be "foo-baz".
 
+When dealing with settings that should return a non-string value (ex. booleans), just specify the JSON data type by adding `--type <json-data-type>` to the setting-generator line.
+The resulting setting values will still be returned as stdout to sundog, but will not be wrapped in quotes.
+For example, "schnauzer settings.foo.bar --type bool" could return false and the value in the API would be a proper boolean.
+
 (The name "schnauzer" comes from the fact that Schnauzers are search and rescue dogs (similar to this search and replace task) and because they have mustaches.)
 */
 
@@ -53,6 +57,9 @@ mod error {
             output: String,
             source: serde_json::error::Error,
         },
+
+        #[snafu(display("Setting type '{}' not supported.", setting_type))]
+        SettingWrongType { setting_type: String },
 
         #[snafu(display("Missing metadata {} for key: {}", meta, key))]
         MissingMetadata { meta: String, key: String },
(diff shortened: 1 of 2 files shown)


sources/api/schnauzer/src/main.rs
```diff
@@ -161,8 +219,7 @@ async fn run() -> Result<()> {
 
     // sundog expects JSON-serialized output so that many types can be represented, allowing the
     // API model to use more accurate types.
-    let output = serde_json::to_string(&setting)
-        .context(error::SerializeOutputSnafu { output: &setting })?;
+    let output = parse_output(setting_type, setting)?;
```
bcressey, line 222: This approach doesn't seem correct to me.

The point of `schnauzer` is to render a template, which by definition means we want a string to come out. When given a string value, `serde_json` should serialize it as a string.

I don't really want the dynamic power that's implied by letting `schnauzer` serialize arbitrary templates into arbitrary JSON objects. It's much better for the API model and the generator to cooperate and share a static type definition.

In this case I'd recommend simply not using `schnauzer` at all if its string output is a concern. This is consistent with other generators which emit different types depending on the target setting.

```
bash-5.1# pluto max-pods
58

bash-5.1# netdog node-ip
"10.0.95.193"

bash-5.1# prairiedog generate-boot-settings
{}

bash-5.1# metricdog should-bark
false
```

bcressey:
Can you provide an example of the proposed setting generator and template that would use this functionality?

jpculp:
Added example to the description.

sources/api/schnauzer/src/main.rs
jpculp: I might be misunderstanding the concern. If we take pluto for example, `max-pods` returns unquoted stdout which the API interprets as a number, but `cluster-dns-ip` is quoted stdout which the API interprets as a string. At the end of the day, the proposed change is really just adding the ability to return a string that is not wrapped in quotes.

bcressey:
> Assuming you write a helper in schnauzer for barking to return "true" or "false", Bottlerocket would previously fail to apply this to the API as the model is expecting a bool instead of a string.

I think it would be better to handle this logic in `metricdog` rather than a `schnauzer` helper. Unfortunately that means that `metricdog` then needs to talk to the API which it hasn't needed until now.

This might be nice:
```
[metadata.settings.metrics.send-metrics]
setting-generator = "metricdog should-bark --region {{ settings.aws.region }}"
```

In other words, teach `sundog` to do the template expansion trick that `schnauzer` does, possibly using `schnauzer` itself on the `setting-generator` value, to avoid propagating the API dependency down into programs that we want to keep more self-contained (like `updog` and `metricdog`).

PR #2223, opened 2022-06-17
adjust permissions for /boot and System.map

bcressey:
**Issue number:**
N/A

**Description of changes:**
`System.map` likely isn't used by anything on modern systems, and the copy in `/boot` is even less likely to be used since for most of Bottlerocket's history (until 9e4c2cc055c7eea16550af5ec38cf29f0949050c) we didn't even mount that filesystem. Since `/boot` is space-constrained, especially on aarch64, we should just get rid of it.

Make `/boot` and the remaining copy of `System.map` readable only by UID 0, per convention.

**Testing done:**
Confirmed the new permissions on a running system.

```
bash-5.0# ls -latr /boot
total 32997
-rwx------.  1 root root 39059968 Jun 17 20:41 vmlinuz
-rw-------.  1 root root   165786 Jun 17 20:41 config
drwx------.  2 root root     1024 Jun 17 20:44 efi
drwxr-xr-x. 19 root root     4096 Jun 17 20:44 ..
drwx------.  2 root root    12288 Jun 17 20:44 lost+found
drwx------.  2 root root     1024 Jun 17 20:44 grub
drwxr-xr-x.  5 root root     1024 Jun 17 20:44 .

bash-5.0# ls -latr /lib/modules/5.10.118/build/System.map 
-rw-------. 1 root root 5303557 Jun 17 20:39 /lib/modules/5.10.118/build/System.map
```

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.10/kernel-5.10.spec
+++ b/packages/kernel-5.10/kernel-5.10.spec
@@ -118,7 +118,6 @@ make -s\\\
 install -d %{buildroot}/boot
 install -T -m 0755 arch/%{_cross_karch}/boot/%{_cross_kimage} %{buildroot}/boot/vmlinuz
 install -m 0644 .config %{buildroot}/boot/config
-install -m 0644 System.map %{buildroot}/boot/System.map
 
 find %{buildroot}%{_cross_prefix} \
    \( -name .install -o -name .check -o \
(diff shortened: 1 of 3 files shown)


arnaldo2792 (approved):

foersleo (approved):
Looks good to me.

As far as i can tell the failed test should not be related to this change (grub build failing for one variant during configure phase due to unexpected EOF).

markusboehme (approved):

PR #2226, opened 2022-06-21
Add kernel 5.15 sources

### LINKED ISSUE #2008: Add support for Linux Kernel 5.15 LTS
Please add support for Linux Kernel 5.15 LTS to support newer hardware and to include the new features included in the newer kernel.

arnaldo2792:
**Issue number:**
Closes #2008

**Description of changes:**

```
variants: update kernel for dev variants
packages: add kernel-5.15 sources
```

TODO:

- [x] Update lookaside cache

**Testing done:**

- Launched `aws-dev` VM, and it booted successfully:

```
bash-5.1# uname -a
Linux ip-172-31-61-1.us-west-2.compute.internal 5.15.43 #1 SMP Tue Jun 21 22:24:09 UTC 2022 x86_64 GNU/Linux
```

- Tested the `kmod-kit` to compile an out-of-tree kernel module:

```bash
> make driver -j
warning: the compiler differs from the one used to build the kernel
  The kernel was built by: x86_64-bottlerocket-linux-gnu-gcc (Buildroot 2021.02.3) 10.3.0
  You are using:           x86_64-bottlerocket-linux-musl-gcc (Buildroot 2021.02.3) 10.3.0
Skipping BTF generation for /tmp/falco/build/driver/src/falco.ko due to unavailability of vmlinux
Built target driver
> [ -f driver/falco.ko ] && echo "Exists!"
Exists!
```

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.15/config-bottlerocket
+++ b/packages/kernel-5.15/config-bottlerocket
@@ -1,0 +1,129 @@
+# Because Bottlerocket does not have an initramfs, modules required to mount
+# the root filesystem must be set to y.
+
+# The root filesystem is ext4
+CONFIG_EXT4_FS=y
+
+# NVMe support
+CONFIG_BLK_DEV_NVME=y
+CONFIG_NVME_CORE=y
+
+# Xen blkfront for Xen-based EC2 platforms
+CONFIG_XEN_BLKDEV_FRONTEND=y
+
+# virtio for local testing with QEMU
+CONFIG_VIRTIO=y
+CONFIG_VIRTIO_BLK=y
+CONFIG_VIRTIO_PCI=y
+
+# dm-verity and enabling it on the kernel command line
+CONFIG_BLK_DEV_DM=y
+CONFIG_DAX=y
+CONFIG_DM_INIT=y
+CONFIG_DM_VERITY=y
+
+# TCMU/LIO
+CONFIG_TCM_USER2=m
+
+# EFI
+CONFIG_EFI=y
+CONFIG_EFI_STUB=y
+CONFIG_EFI_MIXED=y
+
+# EFI video
+CONFIG_FB=y
+CONFIG_FB_EFI=y
+CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER=y
+
+# yama LSM for ptrace restrictions
+CONFIG_SECURITY_YAMA=y
+
+# Do not allow SELinux to be disabled at boot.
+CONFIG_SECURITY_SELINUX_BOOTPARAM=n
+
+# Do not allow SELinux to be disabled at runtime.
+CONFIG_SECURITY_SELINUX_DISABLE=n
+
+# Do not allow SELinux to use `enforcing=0` behavior.
+CONFIG_SECURITY_SELINUX_DEVELOP=n
+
+# Check the protection applied by the kernel for mmap and mprotect,
+# rather than the protection requested by userspace.
+CONFIG_SECURITY_SELINUX_CHECKREQPROT_VALUE=0
+
+# Enable support for the kernel lockdown security module.
+CONFIG_SECURITY_LOCKDOWN_LSM=y
+
+# Enable lockdown early so that if the option is present on the
+# kernel command line, it can be enforced.
+CONFIG_SECURITY_LOCKDOWN_LSM_EARLY=y
+
+# Enable zstd compression for squashfs.
+CONFIG_SQUASHFS_ZSTD=y
+
+# enable /proc/config.gz
+CONFIG_IKCONFIG=y
+CONFIG_IKCONFIG_PROC=y
+
+# kernel headers at /sys/kernel/kheaders.tar.xz
+CONFIG_IKHEADERS=y
+
+# BTF debug info at /sys/kernel/btf/vmlinux
+CONFIG_DEBUG_INFO_BTF=y
+
+# We don't want to extend the kernel command line with any upstream defaults;
+# Bottlerocket uses a fairly custom setup that needs tight control over it.
+CONFIG_CMDLINE_EXTEND=n
+
+# Enable ZSTD kernel image compression
+CONFIG_HAVE_KERNEL_ZSTD=y
+CONFIG_KERNEL_ZSTD=y
+CONFIG_ZSTD_COMPRESS=y
+CONFIG_ZSTD_DECOMPRESS=y
+CONFIG_DECOMPRESS_ZSTD=y
+
+# Load i8042 controller, keyboard, and mouse as modules, to avoid waiting for
+# them before mounting the root device.
+CONFIG_SERIO_I8042=m
+CONFIG_KEYBOARD_ATKBD=m
+CONFIG_MOUSE_PS2=m
+
+# Add support for IPMI drivers
+CONFIG_IPMI_HANDLER=m
+
+# Add support for bootconfig
+CONFIG_BOOT_CONFIG=y
+
+# Enables support for checkpoint/restore
+CONFIG_CHECKPOINT_RESTORE=y
+
+# Disable unused filesystems.
+CONFIG_AFS_FS=n
+CONFIG_CRAMFS=n
+CONFIG_ECRYPT_FS=n
+CONFIG_EXT2_FS=n
+CONFIG_EXT3_FS=n
+CONFIG_EXT4_USE_FOR_EXT2=y
+CONFIG_GFS2_FS=n
+CONFIG_HFS_FS=n
+CONFIG_HFSPLUS_FS=n
+CONFIG_JFS_FS=n
+CONFIG_JFFS2_FS=n
+CONFIG_NFS_V2=n
+CONFIG_NILFS2_FS=n
+CONFIG_NTFS_FS=n
+CONFIG_ROMFS_FS=n
+CONFIG_UFS_FS=n
+CONFIG_ZONEFS_FS=n
+CONFIG_NTFS3_FS=n
+
+# Disable unused network protocols.
+CONFIG_AF_RXRPC=n
+CONFIG_ATM=n
+CONFIG_CAN=n
+CONFIG_HSR=n
+CONFIG_IP_DCCP=n
+CONFIG_L2TP=n
+CONFIG_RDS=n
+CONFIG_RFKILL=n
+CONFIG_TIPC=n
--- a/packages/selinux-policy/class.cil
+++ b/packages/selinux-policy/class.cil
@@ -124,6 +124,7 @@
 (classcommon qipcrtr_socket socket)
 (classcommon smc_socket socket)
 (classcommon xdp_socket socket)
+(classcommon mctp_socket socket)
 
 ; Add permissions specific to some socket classes.
 (class socket ())
@@ -186,6 +187,7 @@
 (class qipcrtr_socket ())
 (class smc_socket ())
 (class xdp_socket ())
+(class mctp_socket ())
 
 ; Add permissions for various network classes.
 (class node (recvfrom sendto))
@@ -259,7 +261,7 @@
   phonet_socket ieee802154_socket caif_socket alg_socket
   nfc_socket vsock_socket kcm_socket qipcrtr_socket smc_socket
   infiniband_pkey infiniband_endport bpf xdp_socket
-  perf_event lockdown))
+  mctp_socket perf_event lockdown))
 
 ; Add permissions for SELinux-aware applications.
 ; This includes systemd and dbus-broker.
--- a/packages/selinux-policy/class.cil
+++ b/packages/selinux-policy/class.cil
@@ -50,6 +50,7 @@
 (classcommon blk_file file)
 (classcommon sock_file file)
 (classcommon fifo_file file)
+(classcommon anon_inode file)
 
 ; Add permissions specific to some file classes.
 (class file (execute_no_trans entrypoint))
@@ -60,6 +61,7 @@
 (class blk_file ())
 (class sock_file ())
 (class fifo_file ())
+(class anon_inode ())
 
 ; Add permissions shared by all socket classes.
 (common socket (
@@ -261,7 +263,7 @@
   phonet_socket ieee802154_socket caif_socket alg_socket
   nfc_socket vsock_socket kcm_socket qipcrtr_socket smc_socket
   infiniband_pkey infiniband_endport bpf xdp_socket
-  mctp_socket perf_event lockdown))
+  mctp_socket perf_event lockdown anon_inode))
 
 ; Add permissions for SELinux-aware applications.
 ; This includes systemd and dbus-broker.
--- a/packages/selinux-policy/files.cil
+++ b/packages/selinux-policy/files.cil
@@ -9,6 +9,7 @@
 (classmapping files relabel relabel_blk_file)
 (classmapping files relabel relabel_sock_file)
 (classmapping files relabel relabel_fifo_file)
+(classmapping files relabel relabel_anon_inode)
 
 ; Permission group for mounts.
 (classmapping files mount mount_file)
@@ -19,6 +20,7 @@
 (classmapping files mount mount_sock_file)
 (classmapping files mount mount_fifo_file)
 (classmapping files mount mount_filesystem)
+(classmapping files mount mount_anon_inode)
 
 ; Permission group for relaxing security constraints on files.
 (classmapping files relax relax_file)
@@ -28,6 +30,7 @@
 (classmapping files relax relax_blk_file)
 (classmapping files relax relax_sock_file)
 (classmapping files relax relax_fifo_file)
+(classmapping files relax relax_anon_inode)
 
 ; Permission group for using files as entry points.
 (classmapping files enter enter_file)
@@ -41,6 +44,7 @@
 (classmapping files describe describe_sock_file)
 (classmapping files describe describe_fifo_file)
 (classmapping files describe describe_filesystem)
+(classmapping files describe describe_anon_inode)
 
 ; Permission group for reading and executing files.
 (classmapping files load load_file)
@@ -52,6 +56,7 @@
 (classmapping files load load_fifo_file)
 (classmapping files load load_filesystem)
 (classmapping files load load_fd)
+(classmapping files load load_anon_inode)
 
 ; Permission group for blocking access to files.
 (classmapping files block block_file)
@@ -61,6 +66,7 @@
 (classmapping files block block_blk_file)
 (classmapping files block block_sock_file)
 (classmapping files block block_fifo_file)
+(classmapping files block block_anon_inode)
 
 ; Permission group for mutating files.
 (classmapping files mutate mutate_file)
@@ -70,6 +76,7 @@
 (classmapping files mutate mutate_blk_file)
 (classmapping files mutate mutate_sock_file)
 (classmapping files mutate mutate_fifo_file)
+(classmapping files mutate mutate_anon_inode)
 
 ; Sets of permissions for relabeling file objects.
 (classpermission relabel_file)
@@ -79,6 +86,7 @@
 (classpermission relabel_blk_file)
 (classpermission relabel_sock_file)
 (classpermission relabel_fifo_file)
+(classpermission relabel_anon_inode)
 (classpermissionset relabel_file (
   file (relabelfrom relabelto)))
 (classpermissionset relabel_dir (
@@ -93,6 +101,8 @@
   sock_file (relabelfrom relabelto)))
 (classpermissionset relabel_fifo_file (
   fifo_file (relabelfrom relabelto)))
+(classpermissionset relabel_anon_inode(
+  anon_inode (relabelfrom relabelto)))
 
 ; Sets of permissions for mounts.
 (classpermission mount_file)
@@ -103,6 +113,7 @@
 (classpermission mount_sock_file)
 (classpermission mount_fifo_file)
 (classpermission mount_filesystem)
+(classpermission mount_anon_inode)
 (classpermissionset mount_file (
   file (mounton quotaon)))
 (classpermissionset mount_dir (
@@ -119,6 +130,8 @@
   fifo_file (mounton quotaon)))
 (classpermissionset mount_filesystem (
   filesystem (mount quotamod remount unmount)))
+(classpermissionset mount_anon_inode (
+  anon_inode (mounton quotaon)))
 
 ; Sets of permissions that relax security constraints for file objects.
 (classpermission relax_file)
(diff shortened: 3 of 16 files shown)


zmrow (approved):
I can't speak for the nvidia side of things - but everything else looks good.

arnaldo2792:
Forced push removed the NVIDIA driver for the 5.15 kernel, since the 470 version didn't compile aarch64 with this new kernel.

packages/kernel-5.15/1002-Makefile-expose-tools-builds-for-external-modules.patch
```diff
@@ -0,0 +1,61 @@
+From f39069aa728e8e7038ba0568f81b97298e47cd0f Mon Sep 17 00:00:00 2001
+From: Arnaldo Garcia Rincon <agarrcia@amazon.com>
+Date: Tue, 21 Jun 2022 16:42:09 +0000
+Subject: [PATCH 1002/1002] Makefile: expose tools/ builds for external modules
```
foersleo: As far as I can tell this is a clean revert of the upstream patch mentioned. We should name it as what it is and provide the reasoning on why we are reverting this one (as you have done in the commit message).

I will have to take a look at kmod kit to see if this is the right way to do this.

markusboehme: Reverting this (https://lore.kernel.org/all/20210512065201.35268-1-masahiroy@kernel.org/) looks like a decent approach to me. The commit moved the definition of the `tools` target into a block guarded by `KBUILD_EXTMOD` (https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/Makefile?h=v5.15.43#n1117), only set when Kbuild is invoked for an out-of-tree build. I, too, would prefer a proper revert commit the way `git revert` would generate it, though.

An alternative would be to (cross-compile if needed and) package the required tools in the kmod kit, but this right here seems simpler.

markusboehme:
(Note to myself: Comparing to the 5.10 kernel build, we could drop the two additional patches to support zstd since they were originally taken from upstream 5.13.)

Since I did not see any mention of config changes, have you checked the resulting Bottlerocket kernel configs against the ones for 5.10? Going from upstream 5.10 to upstream 5.15 naturally changes the available config options (new features, drivers, etc.), but there's also been work on the Amazon Linux kernel configs we base our configs on. I've only skimmed the resulting differences and mostly found welcome changes (e.g. `CONFIG_DEVMEM=n` that we talked about on other occasion), changes we're indifferent about (e.g. disabling support for Bluetooth, dropping the NTFS driver), and changes we might consider reverting at a later point (e.g. building support for the CAN bus). Some are not immediately clear (e.g. disabling support for EFI variables in sysfs in favor of efivarfs [probably the right time for that], dropping support for file-based encryption [probably fine to add if there's actually any users])... On the plus side, I did not see any obvious blockers.

packages/kernel-5.15/1002-Makefile-expose-tools-builds-for-external-modules.patch
arnaldo2792: The alternative approach that I thought of was to copy the targets inside the `KBUILD_EXTMOD` block instead of reverting the commit just to make our lives easier in the future in case our revert conflicts with updates in the Makefile.

Regarding this:

> An alternative would be to (cross-compile if needed and) package the required tools in the kmod kit, but this right here seems simpler.

There is a comment in the other patch that explains why we didn't include the compiled binaries in the archive:

>Unlike other distributions, we cannot include these programs in our kernel-devel archive, because we rely on cross-compilation: these are "host" programs and may not match the architecture of the target

arnaldo2792:
Regarding the diff between the 5.10 and 5.15 configs, I'll add a comment in this PR comparing the two configurations :+1:

packages/kernel-5.15/config-bottlerocket
```diff
@@ -0,0 +1,136 @@
+# Because Bottlerocket does not have an initramfs, modules required to mount
+# the root filesystem must be set to y.
```
bcressey, line 2: Please keep "rebasing" this on 5.10 kernel config changes until this merges, and then please nag others to keep this config current.

bcressey (commented):
Can you share the SELinux related output from early boot? It'll look like this:

```
[    4.150596] SELinux:  policy capability network_peer_controls=1
[    4.155730] SELinux:  policy capability open_perms=1
[    4.160314] SELinux:  policy capability extended_socket_class=1
[    4.165506] SELinux:  policy capability always_check_network=0
[    4.170627] SELinux:  policy capability cgroup_seclabel=1
[    4.175380] SELinux:  policy capability nnp_nosuid_transition=1
[    4.180373] SELinux:  policy capability genfs_seclabel_symlinks=1
[    4.185486] SELinux:  policy capability ioctl_skip_cloexec=0
[    4.233927] systemd[1]: Successfully loaded SELinux policy in 238.564ms.
```

There are likely to be new classes or actions that need rules, if we don't want them to be denied by default.

packages/kernel-5.15/config-bottlerocket
arnaldo2792: There are two new classes that need to be added:

```
[    0.956969] SELinux:  Class mctp_socket not defined in policy.
[    0.957602] SELinux:  Class anon_inode not defined in policy.
[    0.958165] SELinux: the above unknown classes and permissions will be denied
[    0.960123] SELinux:  policy capability network_peer_controls=1
[    0.960731] SELinux:  policy capability open_perms=1
[    0.961217] SELinux:  policy capability extended_socket_class=1
[    0.961812] SELinux:  policy capability always_check_network=0
[    0.962377] SELinux:  policy capability cgroup_seclabel=1
[    0.962907] SELinux:  policy capability nnp_nosuid_transition=1
[    0.963484] SELinux:  policy capability genfs_seclabel_symlinks=1
[    0.964114] SELinux:  policy capability ioctl_skip_cloexec=0
```

arnaldo2792:
( forced push includes the proper revert patch for the commit required to compile out-of-tree kernel modules)

arnaldo2792:
Forced push includes:

- Fix in SELinux policy to include missing classes
- Add configs from other kernels

arnaldo2792:
@markusboehme here is the script that I used to generate a "diff":

```fish
#! /usr/bin/env fish

set --local old aws-ecs-1-x86_64-kmod-kit-v1.8.0/kernel-devel/.config
set --local new aws-dev-x86_64-kmod-kit-v1.8.0/kernel-devel/.config

for c in (sed -e 's|#.*||g' -e 's|#||g' $old  | rg -v '^$')
  # First check if setting exists in new configs
  set --local config_name (echo $c | awk ' { split($1,c,"="); print(c[1]) } ')
  if ! rg -q $config_name $new
    echo "Config isn't set in 5.15: $config_name"
  else if ! rg -q $c $new # check if the values are different
    set --local new_config (rg $config_name $new)
    echo "Value in 5.10: $c, Value in 5.15: $new_config"
  end
end

for c in (sed -e 's|#.*||g' -e 's|#||g' $new  | rg -v '^$')
  set --local config_name (echo $c | awk ' { split($1,c,"="); print(c[1]) } ')
  # Check new configs
  if ! rg -q $config_name $old
    echo "Config is new in 5.15: $c"
  end
end
```

```
Config isn't set in 5.15: CONFIG_PRINTK_NMI
Config isn't set in 5.15: CONFIG_SLUB_MEMCG_SYSFS_ON
Config isn't set in 5.15: CONFIG_X86_RESERVE_LOW
Config isn't set in 5.15: CONFIG_OPROFILE
Config isn't set in 5.15: CONFIG_HAVE_OPROFILE
Config isn't set in 5.15: CONFIG_OPROFILE_NMI_TIMER
Config isn't set in 5.15: CONFIG_UNUSED_SYMBOLS
Config isn't set in 5.15: CONFIG_BLK_SCSI_REQUEST
Config isn't set in 5.15: CONFIG_BLK_CMDLINE_PARSER
Config isn't set in 5.15: CONFIG_NEED_MULTIPLE_NODES
Config isn't set in 5.15: CONFIG_BOUNCE
Config isn't set in 5.15: CONFIG_DAMON_PADDR
Config isn't set in 5.15: CONFIG_DAMON_RECLAIM
Config isn't set in 5.15: CONFIG_NF_LOG_COMMON
Config isn't set in 5.15: CONFIG_NF_LOG_NETDEV
Config isn't set in 5.15: CONFIG_NF_LOG_BRIDGE
Config isn't set in 5.15: CONFIG_HAVE_NET_DSA
Config isn't set in 5.15: CONFIG_XEN_PCIDEV_FRONTEND
Config isn't set in 5.15: CONFIG_HAVE_IDE
Value in 5.10: CONFIG_DM_ZONED=m, Value in 5.15: # CONFIG_DM_ZONED is not set
Config isn't set in 5.15: CONFIG_INPUT_POLLDEV
Value in 5.10: CONFIG_DEVMEM=y, Value in 5.15: # CONFIG_DEVMEM is not set
Config isn't set in 5.15: CONFIG_RAW_DRIVER
Config isn't set in 5.15: CONFIG_MAX_RAW_DEVS
Config isn't set in 5.15: CONFIG_DRM_KMS_FB_HELPER
Config isn't set in 5.15: CONFIG_DRM_TTM_DMA_PAGE_POOL
Config isn't set in 5.15: CONFIG_SWIOTLB_XEN
Value in 5.10: CONFIG_LUSTREFSX_LIBCFS=m, Value in 5.15: # CONFIG_LUSTREFSX_LIBCFS is not set
Value in 5.10: CONFIG_LUSTREFSX_LNET=m, Value in 5.15: # CONFIG_LUSTREFSX_LNET is not set
Config isn't set in 5.15: CONFIG_LUSTREFSX_FS
Config isn't set in 5.15: CONFIG_DCDBAS
Config isn't set in 5.15: CONFIG_DELL_RBU
Config isn't set in 5.15: CONFIG_CLKDEV_LOOKUP
Config isn't set in 5.15: CONFIG_MANDATORY_FILE_LOCKING
Config isn't set in 5.15: CONFIG_CIFS_WEAK_PW_HASH
Config isn't set in 5.15: CONFIG_CRYPTO_GLUE_HELPER_X86
Config isn't set in 5.15: CONFIG_CRYPTO_RMD128
Config isn't set in 5.15: CONFIG_CRYPTO_RMD256
Config isn't set in 5.15: CONFIG_CRYPTO_RMD320
Config isn't set in 5.15: CONFIG_CRYPTO_TGR192
Config isn't set in 5.15: CONFIG_CRYPTO_SALSA20
Config isn't set in 5.15: CONFIG_RING_BUFFER_ALLOW_SWAP
Value in 5.10: CONFIG_STRICT_DEVMEM=y, Value in 5.15: # CONFIG_STRICT_DEVMEM is not set
Config is new in 5.15: CONFIG_AS_IS_GNU=y
Config is new in 5.15: CONFIG_AS_VERSION=23502
Config is new in 5.15: CONFIG_LD_IS_BFD=y
Config is new in 5.15: CONFIG_CC_HAS_NO_PROFILE_FN_ATTR=y
Config is new in 5.15: CONFIG_CGROUP_MISC=y
Config is new in 5.15: CONFIG_HAVE_ARCH_USERFAULTFD_MINOR=y
Config is new in 5.15: CONFIG_ARCH_NR_GPIO=1024
Config is new in 5.15: CONFIG_XEN_PVHVM_GUEST=y
Config is new in 5.15: CONFIG_PERF_EVENTS_AMD_UNCORE=y
Config is new in 5.15: CONFIG_ACPI_VIOT=y
Config is new in 5.15: CONFIG_ACPI_PRMT=y
Config is new in 5.15: CONFIG_HAVE_KVM_PM_NOTIFIER=y
Config is new in 5.15: CONFIG_ARCH_WANTS_NO_INSTR=y
Config is new in 5.15: CONFIG_ARCH_SUPPORTS_LTO_CLANG=y
Config is new in 5.15: CONFIG_ARCH_SUPPORTS_LTO_CLANG_THIN=y
Config is new in 5.15: CONFIG_LTO_NONE=y
Config is new in 5.15: CONFIG_HAVE_CONTEXT_TRACKING_OFFSTACK=y
Config is new in 5.15: CONFIG_HAVE_MOVE_PUD=y
Config is new in 5.15: CONFIG_HAVE_IRQ_EXIT_ON_IRQ_STACK=y
Config is new in 5.15: CONFIG_HAVE_SOFTIRQ_ON_OWN_STACK=y
Config is new in 5.15: CONFIG_HAVE_ARCH_RANDOMIZE_KSTACK_OFFSET=y
Config is new in 5.15: CONFIG_HAVE_PREEMPT_DYNAMIC=y
Config is new in 5.15: CONFIG_ARCH_HAS_ELFCORE_COMPAT=y
Config is new in 5.15: CONFIG_ARCH_HAS_PARANOID_L1D_FLUSH=y
Config is new in 5.15: CONFIG_MODULE_COMPRESS_NONE=y
Config is new in 5.15: CONFIG_MODPROBE_PATH="/sbin/modprobe"
Config is new in 5.15: CONFIG_BLK_DEV_BSG_COMMON=y
Config is new in 5.15: CONFIG_BLOCK_HOLDER_DEPRECATED=y
Config is new in 5.15: CONFIG_ARCH_HAS_ZONE_DMA_SET=y
Config is new in 5.15: CONFIG_SECRETMEM=y
Config is new in 5.15: CONFIG_AF_UNIX_OOB=y
Config is new in 5.15: CONFIG_NF_LOG_SYSLOG=m
Config is new in 5.15: CONFIG_NETFILTER_XTABLES_COMPAT=y
Config is new in 5.15: CONFIG_PCPU_DEV_REFCNT=y
Config is new in 5.15: CONFIG_SOCK_RX_QUEUE_MAPPING=y
Config is new in 5.15: CONFIG_NET_SELFTESTS=m
Config is new in 5.15: CONFIG_AUXILIARY_BUS=y
Config is new in 5.15: CONFIG_SYSFB=y
Config is new in 5.15: CONFIG_ZRAM_DEF_COMP_LZORLE=y
Config is new in 5.15: CONFIG_ZRAM_DEF_COMP="lzo-rle"
Config is new in 5.15: CONFIG_SCSI_COMMON=y
Config is new in 5.15: CONFIG_BNXT_SRIOV=y
Config is new in 5.15: CONFIG_BNXT_FLOWER_OFFLOAD=y
Config is new in 5.15: CONFIG_BNXT_HWMON=y
Config is new in 5.15: CONFIG_MLX5_CORE=m
Config is new in 5.15: CONFIG_MLX5_CORE_EN=y
Config is new in 5.15: CONFIG_MLX5_EN_ARFS=y
Config is new in 5.15: CONFIG_MLX5_EN_RXNFC=y
Config is new in 5.15: CONFIG_MLX5_MPFS=y
Config is new in 5.15: CONFIG_MLX5_ESWITCH=y
Config is new in 5.15: CONFIG_MLX5_BRIDGE=y
Config is new in 5.15: CONFIG_MLX5_CLS_ACT=y
Config is new in 5.15: CONFIG_MLX5_TC_SAMPLE=y
Config is new in 5.15: CONFIG_MLX5_CORE_EN_DCB=y
Config is new in 5.15: CONFIG_MLX5_SW_STEERING=y
Config is new in 5.15: CONFIG_MLXFW=m
Config is new in 5.15: CONFIG_SWPHY=y
Config is new in 5.15: CONFIG_FWNODE_MDIO=m
Config is new in 5.15: CONFIG_ACPI_MDIO=m
Config is new in 5.15: CONFIG_PTP_1588_CLOCK_OPTIONAL=m
Config is new in 5.15: CONFIG_DRM_I915_REQUEST_TIMEOUT=20000
Config is new in 5.15: CONFIG_MLX5_INFINIBAND=m
Config is new in 5.15: CONFIG_VFIO_PCI_CORE=m
Config is new in 5.15: CONFIG_ARCH_HAS_RESTRICTED_VIRTIO_MEMORY_ACCESS=y
Config is new in 5.15: CONFIG_VIRTIO_PCI_LIB=y
Config is new in 5.15: CONFIG_SURFACE_PLATFORMS=y
Config is new in 5.15: CONFIG_IOMMU_IO_PGTABLE=y
Config is new in 5.15: CONFIG_IOMMU_DEFAULT_DMA_LAZY=y
Config is new in 5.15: CONFIG_VIRTIO_IOMMU=m
Config is new in 5.15: CONFIG_NETFS_SUPPORT=m
Config is new in 5.15: CONFIG_NTFS3_FS=m
Config is new in 5.15: CONFIG_HUGETLB_PAGE_FREE_VMEMMAP=y
Config is new in 5.15: CONFIG_PSTORE_DEFAULT_KMSG_BYTES=10240
Config is new in 5.15: CONFIG_NFS_V4_2_SSC_HELPER=y
Config is new in 5.15: CONFIG_SMBFS_COMMON=m
Config is new in 5.15: CONFIG_MODULE_SIG_KEY_TYPE_RSA=y
Config is new in 5.15: CONFIG_ASN1_ENCODER=m
Config is new in 5.15: CONFIG_PAHOLE_HAS_SPLIT_BTF=y
Config is new in 5.15: CONFIG_DEBUG_INFO_BTF_MODULES=y
Config is new in 5.15: CONFIG_HAVE_ARCH_KFENCE=y
Config is new in 5.15: CONFIG_HAVE_DYNAMIC_FTRACE_WITH_ARGS=y
Config is new in 5.15: CONFIG_HAVE_OBJTOOL_MCOUNT=y
Config is new in 5.15: CONFIG_DYNAMIC_FTRACE_WITH_ARGS=y
Config is new in 5.15: CONFIG_FTRACE_MCOUNT_USE_CC=y
Config is new in 5.15: CONFIG_ARCH_USE_MEMTEST=y
```

markusboehme:
Thanks for listing the config changes! I like to use [diffconfig](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/scripts/diffconfig?h=v5.18) for these purposes. Perhaps you'll find it helpful, too. Either way, I don't have any doubts about unwelcome config changes anymore.

markusboehme (approved):
The SELinux policy changes look reasonable to me by comparison to similar object classes. I am unsure about the significance of the position of the items in the `classorder` and `sidorder` definitions. I'll have to defer to someone with more SELinux expertise on that until I catch up.

bcressey:
```
Config is new in 5.15: CONFIG_NTFS3_FS=m
```

I'd very much prefer to *not* take this change. Everything else looks OK.

packages/kernel-5.15/config-bottlerocket
```diff
@@ -0,0 +1,145 @@
...
+CONFIG_NETDEVICES=y
+
+# Intel network support
+CONFIG_IGB=m
+CONFIG_IGBVF=m
+CONFIG_NET_VENDOR_INTEL=m
+CONFIG_IGB_HWMON=y
+CONFIG_E1000=m
+CONFIG_E1000e=m
+CONFIG_E1000e_hwts=y
```
bcressey: Needs to be "rebased" on the changes in #2264.

packages/selinux-policy/class.cil
```diff
@@ -259,7 +263,7 @@
   phonet_socket ieee802154_socket caif_socket alg_socket
   nfc_socket vsock_socket kcm_socket qipcrtr_socket smc_socket
   infiniband_pkey infiniband_endport bpf xdp_socket
-  perf_event lockdown))
+  mctp_socket perf_event anon_inode lockdown))
```
bcressey: This doesn't match the kernel's class order in `security/selinux/include/classmap.h`:
```suggestion
  mctp_socket perf_event lockdown anon_inode))
```

It'd be good to fix that.

packages/selinux-policy/sid.cil
```diff
@@ -33,9 +34,9 @@
 ; The order of ISIDs must match the kernel's order, for now.
 (sidorder (
    kernel security unlabeled fs file file_labels init any_socket port
-   netif netmsg node igmp_packet icmp_socket tcp_socket sysctl_modprobe
-   sysctl sysctl_fs sysctl_kernel sysctl_net sysctl_net_unix sysctl_vm
-   sysctl_dev kmod policy scmp_packet devnull))
+   netif netmsg node igmp_packet icmp_socket tcp_socket mctp_socket 
+   sysctl_modprobe sysctl sysctl_fs sysctl_kernel sysctl_net sysctl_net_unix
+   sysctl_vm sysctl_dev kmod policy scmp_packet devnull))
```
bcressey: This doesn't look correct.

There's no new initial SID for `mctp_socket` in `security/selinux/include/initial_sid_to_string.h`, and it would cause trouble if there were because the order would be wrong on older kernels.

You should revert all changes to `sid.cil`.

packages/selinux-policy/files.cil
```diff
@@ -9,6 +9,7 @@
 (classmapping files relabel relabel_blk_file)
 (classmapping files relabel relabel_sock_file)
 (classmapping files relabel relabel_fifo_file)
+(classmapping files relabel relabel_anon_inode)
```
bcressey, line 12: This treatment of `anon_inode` looks OK.

In the absence of the transition rules, it looks like the `anon_inode` objects will continue to use the subject's label, and be covered by this blanket rule:
```
(allow all_s self (files (mutate)))
```

To actually write policy for this, we'd need a named transition rule and additional rules like:
```
(typetransition all_s all_s anon_inode "[userfaultfd]" userfaultfd_t)
(allow trusted_s userfaultfd_t (files (mutate)))
(neverallow untrusted_s userfaultfd_t (files (mutate)))
```

In other words - force all userfaultfd descriptors to be created with a particular label, and then restricting mutations to just the subjects that we want to be able to use them.

Before we did that, we'd probably want to default to turning off userfaultfd for unprivileged users via `vm.unprivileged_userfaultfd = 0`, and then understand why that wasn't sufficient.

arnaldo2792:
Forced push includes:

- Revert changes in `sid.cil`
- Split kernel configuration as in #2264
- Include configuration from #2269 and #2255
- Set `CONFIG_NTFS3_FS` to `n`

packages/kernel-5.15/config-bottlerocket
```diff
@@ -0,0 +1,129 @@
...
+CONFIG_HFSPLUS_FS=n
+CONFIG_JFS_FS=n
+CONFIG_JFFS2_FS=n
+CONFIG_NFS_V2=n
+CONFIG_NILFS2_FS=n
+CONFIG_NTFS_FS=n
+CONFIG_ROMFS_FS=n
+CONFIG_UFS_FS=n
+CONFIG_ZONEFS_FS=n
+CONFIG_NTFS3_FS=n
```
bcressey, line 118: nit: could put this in sorted order

bcressey (approved):

foersleo (approved):

Issue #1968, opened 2022-02-22
replace rusoto with aws rust sdk


webern:
Issue #1960 might be blocked by our use of rusoto in pubsys. See my comment in that issue for the specifics. Regardless, we should move to the AWS SDK.

kdaula:
Moving to 1.10.0 for addressing Pubsys

markusboehme:
One more thing that this migration will be helpful for: Running Pubsys on EC2 instances with IMDS v1 disabled. Unfortunately, Rusoto doesn't support the two-step process (fetch token, query IMDS) of IMDS v2.

PR #2227, opened 2022-06-21
sundog, pluto: run settings generators with network proxy environment vars

### LINKED ISSUE #2197: sundog does not respect proxy settings
Sundog and the generators that is calls should respect HTTP proxy settings. For example, Pluto will try to call the EKS API and should use the HTTP proxy settings.

etungsten:
**Issue number:**
Fixes https://github.com/bottlerocket-os/bottlerocket/issues/2197

**Description of changes:**
```
    sundog: run settings generators with network proxy environment vars
    
    We make sure settings generators are kicked-off with user-specified
    network proxy environment variables.
```
```
Date:   Tue Jun 21 18:16:38 2022 -0700

    pluto: respect `https_proxy` and `no_proxy` when making EKS API reqs
    
    Both `rusoto` and `aws-sdk-rust` do not respect `https_proxy` and
    `no_proxy` env vars. We have to handle it ourselves by creating a custom
    hyper http 'connector' that's configured with the right proxying
    configuration.

```

**Testing done:**
Builds and runs fine. 
Launched instance with aws-k8s-1.23 AMI with user-data that includes the following:
```
[settings.network]
https-proxy = "my-proxy:9898"
```
Checked proxy server logs and found all traffic being tunneled including the EKS API calls for `DescribeCluster`
```
CONNECT   Jun 22 08:00:00 [3794]: Connect (file descriptor 7): ec2-35-88-231-112.us-west-2.compute.amazonaws.com [35.88.231.112]   
CONNECT   Jun 22 08:00:00 [3794]: Request (file descriptor 7): CONNECT eks.us-west-2.amazonaws.com:443 HTTP/1.1                    
INFO      Jun 22 08:00:00 [3794]: No upstream proxy for eks.us-west-2.amazonaws.com                                                
INFO      Jun 22 08:00:00 [3794]: opensock: opening connection to eks.us-west-2.amazonaws.com:443                                  
INFO      Jun 22 08:00:00 [3794]: opensock: getaddrinfo returned for eks.us-west-2.amazonaws.com:443                               
CONNECT   Jun 22 08:00:00 [3794]: Established connection to host "eks.us-west-2.amazonaws.com" using file descriptor 8.            
INFO      Jun 22 08:00:00 [3794]: Not sending client headers to remote machine                                                     
INFO      Jun 22 08:00:00 [3794]: Closed connection between local client (fd:7) and remote client (fd:8) 
```

Checking CloudTrail, I can see that the DescribeCluster API call source IP is my proxy server
```
...
    "eventTime": "2022-06-22T08:00:00Z",
    "eventSource": "eks.amazonaws.com",
    "eventName": "DescribeCluster",
    "awsRegion": "us-west-2",
    "sourceIPAddress": "my-proxy-IP",
    "userAgent": "rusoto/0.48.0 rust/1.61.0 linux",
...
```

If I set `no-proxy` for the EKS API endpoint for a new instance launch like so:
```
[settings.network]
https-proxy = "my-proxy:9898"
no-proxy = ["eks.us-west-2.amazonaws.com"]
```

Then I can see that EKS API calls do not get proxied and the source IP is my actual instance:
```
...
    "eventTime": "2022-06-22T08:10:04Z",
    "eventSource": "eks.amazonaws.com",
    "eventName": "DescribeCluster",
    "awsRegion": "us-west-2",
    "sourceIPAddress": "34.211.227.47",
    "userAgent": "rusoto/0.48.0 rust/1.61.0 linux",
...
```

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/eks.rs
+++ b/sources/api/pluto/src/eks.rs
@@ -1,7 +1,13 @@
+use hyper::http::uri::InvalidUri;
+use hyper::Uri;
+use hyper_proxy::{Proxy, ProxyConnector};
+use hyper_rustls::HttpsConnectorBuilder;
+use rusoto_core::credential::ChainProvider;
 use rusoto_core::region::ParseRegionError;
 use rusoto_core::{Region, RusotoError};
 use rusoto_eks::{DescribeClusterError, Eks, EksClient, KubernetesNetworkConfigResponse};
 use snafu::{OptionExt, ResultExt, Snafu};
+use std::env;
 use std::str::FromStr;
 
 pub(crate) type ClusterNetworkConfig = KubernetesNetworkConfigResponse;
@@ -21,6 +27,12 @@ pub(super) enum Error {
         region: String,
         source: ParseRegionError,
     },
+
+    #[snafu(display("Unable to parse '{}' as URI: {}", input, source))]
+    UriParse { input: String, source: InvalidUri },
+
+    #[snafu(display("Failed to create proxy creator: {}", source))]
+    ProxyConnector { source: std::io::Error },
 }
 
 type Result<T> = std::result::Result<T, Error>;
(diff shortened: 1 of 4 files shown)


etungsten:
So turns out there's a bit of complication with rusoto not respecting proxy environment variables automatically. The gist of it is that I'll have to create a custom `rusoto_core::HttpClient` with the right proxy config in pluto for the `EksClient`. See https://github.com/rusoto/rusoto/issues/1172  

Once we've switch over to aws-rust-sdk wholesale, we can just remove the custom http client from pluto.
See https://github.com/bottlerocket-os/bottlerocket/issues/1968

etungsten:
Apparently this is still an issue for `aws-rust-sdk` as well, so switching over won't help pluto either. See https://github.com/awslabs/aws-sdk-rust/issues/169

webern (commented):
Wow, I'm disappointed that the AWS Rust SDK does not support proxy env vars. Rusoto I could understand because it may be using an old version of hyper, but I would have expected the AWS Rust SDK to be using the latest hyper under-the-hood.

zmrow (approved):
:ship: 

Nice work!

jpculp (approved):
It looks like we're getting multiple versions of a couple crates due to things depending of `hyper-rustls`, but since `aws-rust-sdk` also uses **0.22.1**, it will get sorted out when we transition.

sources/api/sundog/src/main.rs
```diff
@@ -221,6 +224,54 @@ where
...
+// Builds the proxy environment variables to pass to settings generators
+async fn build_proxy_env<P>(socket_path: P) -> Result<HashMap<String, String>>
+where
+    P: AsRef<Path>,
+{
+    // Retrieve network proxy-related settings.
+    let settings: serde_json::Value =
+        schnauzer::get_json(&socket_path, "/settings", Some(("prefix", "network")))
+            .await
+            .context(error::RetrieveSettingsSnafu)?;
```
bcressey: nit: using `schnauzer` is inconsistent with the other API accesses in this code. I'd prefer all one way or the other, but I'm OK with a backlog item for converting over the rest if that's what we want the idiom to be going forward.

sources/api/sundog/src/main.rs
```diff
@@ -221,6 +224,54 @@ where
...
+    let mut proxy_envs = HashMap::new();
+    if let Some(network_settings) = settings.get("network") {
```
bcressey: nit: I'd check for `None` and do an early return to avoid excessive indents and the related cognitive burden of keeping track of nested conditionals.

sources/api/sundog/src/main.rs
```diff
@@ -221,6 +224,54 @@ where
...
+        if let Some(https_proxy) = network_settings.get("https-proxy").and_then(|s| s.as_str()) {
+            proxy_envs.insert("https_proxy".to_string(), https_proxy.to_string());
+            proxy_envs.insert("HTTPS_PROXY".to_string(), https_proxy.to_string());
+        }
+        if let Some(no_proxy) = network_settings.get("no-proxy").and_then(|v| v.as_array()) {
```
bcressey: nit: we could do an early return here since we don't care about `NO_PROXY` if we don't have either proxy settings.

sources/api/sundog/src/main.rs
```diff
@@ -221,6 +224,54 @@ where
...
+                .map(|s| s.as_str().unwrap_or_default().to_string())
+                .filter(|s| !s.is_empty())
+                .collect();
+            no_proxy.push("localhost".to_string());
+            no_proxy.push("127.0.0.1".to_string());
+            // We potentially need to also no-proxy some K8s related domains for K8s variants
+            let settings: serde_json::Value =
+                schnauzer::get_json(socket_path, "/settings", Some(("prefix", "kubernetes")))
+                    .await
+                    .context(error::RetrieveSettingsSnafu)?;
```
bcressey: I think we need to `NO_PROXY` these even if we don't have an explicit `settings.network.no-proxy` setting. Do we also need to remove duplicate values from the array, in case these were already set?

(I find an aggressive approach to outdenting and early returns helps to catch logic bugs like this, if it is a bug, or to focus the comment on why the logic actually depends on the parent condition if it's not obvious.)

sources/api/pluto/src/eks.rs
```diff
@@ -32,15 +44,77 @@ pub(super) async fn get_cluster_network_config(
...
+        .and_then(|s| s.ok());
+
+    let client = if let Some(https_proxy) = https_proxy {
+        // Determines whether a request of a given scheme, host and port should be proxied according to `https_proxy` and `no_proxy`.
+        let intercept = move |scheme: Option<&str>, host: Option<&str>, _port| {
+            if let Some(host) = host {
+                if let Some(no_proxy) = &no_proxy {
+                    if scheme != Some("http") && scheme != Some("https") {
+                        return false;
+                    }
```
bcressey: Do we want to use `HTTPS_PROXY` for the `http` scheme?

sources/api/sundog/src/main.rs
etungsten: Nice catch!

> Do we also need to remove duplicate values from the array, in case these were already set?

I believe duplicate entries are OK from what we've tested before with `host-ctr` (aws go sdk).We also don't prevent duplicate values in the rendered `proxy.env` `EnvironmentFile` we set for other systemd units.

etungsten:

Tested changes they still work as described in the PR description

jpculp (approved):

bcressey (approved):

rpkelly (approved):

PR #2239, opened 2022-06-24
apply strict mount options and enforce execution rules

### LINKED ISSUE #2234: apply strict mount options where possible
**What I'd like:**
Most of Bottlerocket's default mounts have `nosuid`, `nodev`, and `noexec` applied, but a few do not:
* `/lib/modules`
* `/media/cdrom`
* `/usr/share/licenses`
* `/usr/src/kernels`
* `/var/lib/bottlerocket`
* `/boot`

These are all either read-only or otherwise restricted to privileged users, but consistency is better.

Continue with the current mount options. They're not insecure as such, but they could cause problems in the future if an existing mount unit is copied to a new one without careful review of the options.

bcressey:
**Issue number:**
Fixes #2234

**Description of changes:**
Overall the goal is to apply as many of the `nosuid`, `nodev`, and `noexec` options as we can to each mount.

In systemd, patch the default mount options for `/dev/shm` and `/run` to add `noexec`. This isn't compatible with everything, but nothing in Bottlerocket expects to execute binaries on these filesystems. Also patch the sysusers snippet for `root` to set a "valid" shell: `/sbin/nologin`. I also reworked the patch to mount `/etc` with the expected SELinux label to avoid having a patch that modifies another patch.

Elsewhere, add the desired options to most mounts. Nothing here is very interesting except to note that `/var/lib/bottlerocket` was historically mounted `exec` because we used to write migration binaries there, way back before 1.0. Now they have their own path on `/var/lib/bottlerocket-migrations`. We don't actually run the binaries directly, though; we use [pentacle](https://docs.rs/pentacle/latest/pentacle/) to read them into an anonymous memory-backed file, so they end up running out of tmpfs.

The main weakness of "noexec" is that we end up with `/local` (and subsequently `/var`, ...) mounted `exec`, because otherwise we couldn't run certain containers. Rather than settle for that, I added SELinux policy rules to prevent most host programs from executing binaries on mutable storage locations.

SELinux policy changes can be tricky to read. The change in `files.cil` is fairly mechanical: splitting out permissions from the `load` permission set into a new `execute` permission set. To understand the change in `rules.cil`, pretend that the previous version had these two rules:
```
; All subjects can execute anything that's public.
(allow all_s public (files (execute)))

; Privileged subjects can execute anything at all.
(allow privileged_s global (files (execute)))
```

The new rules replace those two older ones with more fine-grained rules, culminating in the new `neverallow` rule that asserts what most host binaries can execute. Note that `neverallow` rules are more of a compile-time lint; they give you compile errors if the policy breaks the rules.

There's some interesting behavior for SELinux and overlayfs mounts that I've tried to document, and a link in the corresponding commit for more background reading. I've also added labeling for CNI plugin processes, to treat them as unprivileged containers rather than extensions of the container runtime.

**Testing done:**
I tested quite a few different scenarios and everything looks OK.
* `/dev/shm`, `/tmp`, and `/run` have the `noexec` option
* root has `/sbin/nologin` as its shell
* sonobuoy passes for `aws-k8s-1.22` on x86_64
* `aws-dev` and `aws-ecs-1` can run containers correctly
* `findmnt` shows the expected mount options for everything
* `vmware-k8s-1.21` and the Cilium CNI work
* `aws-k8s-1.22-nvidia` can link and load the driver on aarch64
* migrations work properly

😅 

**Terms of contribution:**


packages/selinux-policy/rules.cil
```diff
@@ -122,8 +132,36 @@
...
+; also run other system binaries.
+(allow api_t any_t (files (execute)))
+(allow api_t immutable_o (files (execute)))
+
+; Subjects that must run verified code can execute immutable objects, since
+; those are all protected by dm-verity.
+(allow verified_s immutable_o (files (execute)))
+
+; Subjects that must run verified code cannot execute mutable objects.
+(neverallow verified_s mutable_o (files (execute)))
```
arnaldo2792, line 164: I assume this prevents `runtime_t` from executing literally anything (a binary in a mount), and instead the runtimes are only allowed to execute files with the `immutable_o` label?

bcressey: The `verified_s` subjects exclude `mount_t`, `api_t`, and `runtime_t` since those types of processes need to be able to execute from mutable storage.

etungsten (approved):

foersleo (approved):

Issue #2218, opened 2022-06-15
Symlinking /var/lib/kubelet/pods (or /var/lib/kubelet) breaks the AWS EFS CSI Driver...


diranged:
**Image I'm using:**
Bottlerocket 1.8.0 (either AMD64 or ARM64)

**Background**
We recently had a conversation with @bcressey on Slack (https://awsdevelopers.slack.com/archives/CV5GQHN2Y/p1654555993450939) about mounting the "local disk" from EC2 instances such that Kubernetes would use them for container storage. The suggested answer at https://github.com/bottlerocket-os/bottlerocket/discussions/1991 was then modified based on some suggestions at https://gist.github.com/bcressey/2d3e139dbf01a1e1562f5ba49118258b, and we thought we had a working solution. 

**What went wrong?**

This morning at 2AM an engineering team escalated an issue to us - their Pods were failing to start up with errors like this:
```
Jun 15 08:35:45 ip-100-64-5-48.us-west-2.compute.internal containerd[10986]: time="2022-06-15T08:35:45.445512286Z" level=error msg="CreateContainer within sandbox \"7a431deadde10db2eb406da89caf5f77722f0841fd7c7e3497f7e70746575145\" for &ContainerMetadata{Name:taskmanager,Attempt:0,} failed" error="failed to generate container \"0c57cabdc198b1a68ed412a7f34dbe68b9004974d15f9146aa2b947c2f8d9e43\" spec: failed to generate spec: relabel \"/var/lib/kubelet/pods/e252e874-5b4e-4870-9fa5-633ff8eaa871/volumes/kubernetes.io~csi/pvc-be58bf32-bfaa-4518-9da0-3f9e41f85b21/mount\" with \"system_u:object_r:data_t:s0:c577,c1017\" failed: lsetxattr /var/lib/kubelet/pods/e252e874-5b4e-4870-9fa5-633ff8eaa871/volumes/kubernetes.io~csi/pvc-be58bf32-bfaa-4518-9da0-3f9e41f85b21/mount/55c0ed2511f6c836524254929bb885bc: operation not supported"
```

We discovered that the pods worked fine on hosts where our bootstrap container had not detected any ephemeral disks (`c6i`, `r6i`, etc..) - but they would fail 100% of the time when launching on a host where our bootstrap container had found local disks (`c6id`, etc).

**Root cause?**

[...]

PR #2240, opened 2022-06-24
improve support for container storage mounts

bcressey:
**Issue number:**
#2218 

**Description of changes:**
```
release: set DefaultDependencies=no
```
The first commit is a clean-up commit to end up with consistent dependencies for all mount units.

```
release: adjust mount propagation for /local
release: split up /local preparation
release: mask storage on /local
```
The second set of commits is all related to setting the "private" option for `/local` and dealing with the fallout. `/var` and `/opt` join `/mnt` in getting the "rshared" option; this is what ultimately allows for `/var` to have new mounts for container storage to propagate toward it.

Previously all mounts under `/var` were also listed under `/local`, which could lead to a lot of duplicate mount entries for Kubernetes and Docker volumes. When allowing container storage mounts, `/proc/self/mountinfo` could potentially end up with four entries for each mount, if a `/mnt` directory was bound to a `/var` directory: `/mnt/blah`, `/var/blah`, `/local/mnt/blah`, `/local/var/blah`.

Hence `/local` is made private so it is not part of the peer group. Since `/local` is private, the contents of `/local/var/lib/containerd` might not match `/var/lib/containerd`, if a separate container storage filesystem was mounted. To avoid this inconsistency, each of the `/local` directories are masked with an empty, read-only directory.

This is potentially a breaking change for pods that expect `/local/var` to match `/var`. However, I could not find any public examples of pod specs using `/local/var` in GitHub search, and it seems unusual enough that there's a decent chance that nobody is depending on this behavior.

`/local` setup is split into three parts - setup for `/var`, `/opt`, and `/mnt` - so that it's easier to reason about when setup is "done" for each directory, and when it's OK to mask the contents by mounting an empty directory over it.

```
create mount points for container storage
kubelet: defer creation of subdirectories
host-ctr: propagate container storage mounts
```

The final set of commits is what actually enables bootstrap containers to mount custom volumes over the container storage directories. A directory is created for each container runtime, after first removing any symlink that may have been created by a bootstrap container, since otherwise it might point to a missing directory. Then `host-ctr` mounts any of these directories it finds into bootstrap containers with the "rbind,rshared" flags so mounts will propagate back to the host.

**Testing done:**
Verified that I could use a bootstrap container to create custom storage mounts and bind mount them over `/var/lib/containerd`, `/var/lib/docker`, and `/var/lib/kubelet`. In the `kubelet` case, all CSI drivers I tested continued to work: EFS, EBS, and Secrets Store. 

**Terms of contribution:**


packages/release/mask-local-mnt.service
```diff
@@ -0,0 +1,21 @@
...
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+
+# Mask `/local/mnt` to avoid confusion, since it will have most of the contents
+# of `/mnt` but not any of the mounts.
+ExecStart=/usr/bin/mount --bind --options nosuid,nodev,noexec,private /srv /local/mnt
+ExecStop=/usr/bin/umount /local/mnt
```
arnaldo2792, line 22: If this is a `oneshot`, do you still need the `ExecStop` directive?

packages/kubernetes-1.19/make-kubelet-dirs.conf
```diff
@@ -0,0 +1,5 @@
+[Service]
+# Create the directories for symlinks in /etc
```
arnaldo2792: Both in the commit message and in the systemd services: the directories are created in `/var` not in `/etc`

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -763,6 +763,30 @@ func withPrivilegedMounts() oci.SpecOpts {
...
 
+// withStorageMounts adds options to share container storage mounts
+func withStorageMounts() oci.SpecOpts {
+	var mounts []runtimespec.Mount
+
+	storageDirs := []string{
+		"/var/lib/containerd",
+		"/var/lib/docker",
+		"/var/lib/kubelet",
+	}
```
arnaldo2792, line 776: Should this be a constant, declared close to where it used?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -763,6 +763,30 @@ func withPrivilegedMounts() oci.SpecOpts {
...
+			mounts = append(mounts, runtimespec.Mount{
+				Options:     []string{"rbind", "rshared"},
+				Destination: fmt.Sprintf("/.bottlerocket/rootfs/%s", storageDir),
+				Source:      storageDir,
+				Type:        "bind",
+			})
+		}
+	}
+
+	return oci.WithMounts(mounts)
```
arnaldo2792, line 791: nit: consistency with the other `with*Mounts` functions in the file:

```suggestion
	return withMounts(mounts)
```

packages/release/mask-local-mnt.service
bcressey: This guards against the case where `/mnt` is unmounted.

It should never happen in practice, but if it did, then if `/mnt` were mounted again via the `mnt.mount` unit, it might bind-mount the `/srv` mount over `/mnt`, instead of bind-mounting the underlying directory `/local/mnt`.

packages/kubernetes-1.19/make-kubelet-dirs.conf
bcressey: Would this help clarify? The symlinks are still created in `/etc`, and these are the directories they point to.

```suggestion
# Create the backing directories for symlinks in /etc
```

sources/host-ctr/cmd/host-ctr/main.go
bcressey: Can you give me an example of what you mean?

It feels pretty close to where it's used already but I'm guessing from your comment that it might be idiomatic to declare it as part of the `for` statement somehow?

sources/host-ctr/cmd/host-ctr/main.go
bcressey: It felt unnecessary to call that function since all the mounts are known to have propagation options specified.

I could add a comment instead?

packages/release/mask-local-mnt.service
```diff
@@ -0,0 +1,21 @@
...
+Before=umount.target
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+
+# Mask `/local/mnt` to avoid confusion, since it will have most of the contents
+# of `/mnt` but not any of the mounts.
+ExecStart=/usr/bin/mount --bind --options nosuid,nodev,noexec,private /srv /local/mnt
```
etungsten, line 17: TIL `/srv`, somewhat unexpected way of using it.

etungsten (approved):

sources/host-ctr/cmd/host-ctr/main.go
arnaldo2792:
```golang
const storageDirs := []string{
                "/var/lib/containerd",
		"/var/lib/docker",
		"/var/lib/kubelet",
}

func withStorageMounts() oci.SpecOpts {
    var mounts []runtimespec.Mount

     for _, storageDir := range storageDirs {
          // ...
     }
}
```

sources/host-ctr/cmd/host-ctr/main.go
arnaldo2792: Yeah, a comment will work :+1:

packages/kubernetes-1.19/make-kubelet-dirs.conf
arnaldo2792: Yeah, I think the suggest comment is clearer :+1:

packages/release/mask-local-mnt.service
bcressey: It's definitely odd. I figured relying on an empty `/srv` as an implementation detail was a less grievous sin against [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html) than inventing a top-level `/empty` directory to serve the same purpose.

I'm open to other suggestions if this is too weird.

sources/host-ctr/cmd/host-ctr/main.go
bcressey: Go doesn't support `const` arrays or slices, so I don't think this alternative approach can work.

arnaldo2792 (approved):

bcressey:
Force push with no changes other than a fix for the merge conflict in `local.mount`.

jpculp (approved):

PR #2241, opened 2022-06-24
kernel-5.10: enable mellanox modules

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
    kernel-5.10: enable mellanox modules
    
    This adds the kernel configs necessary for Mellanox NICs.
```

**Testing done:**
Built metal-k8s-1.23 without problem. Provisioned onto a HP DL160 with Mellanox Connect X-5 ethernet cards and the host booted up with network interfaces correctly configured.

Kernel configuration changes:
x86_64:
```
diff --git a/x86_64 without b/x86_64 with
index a0cfca8..adc3268 100644
--- a/x86_64 without    
+++ b/x86_64 with       
@@ -1677,7 +1677,7 @@
  CONFIG_MPLS_IPTUNNEL=m
  CONFIG_NET_NSH=m
  CONFIG_HSR=m
- # CONFIG_NET_SWITCHDEV is not set
+ CONFIG_NET_SWITCHDEV=y
  CONFIG_NET_L3_MASTER_DEV=y
  # CONFIG_QRTR is not set
  # CONFIG_NET_NCSI is not set
@@ -2411,7 +2411,22 @@
  # CONFIG_IGC is not set
  # CONFIG_JME is not set
  # CONFIG_NET_VENDOR_MARVELL is not set
- # CONFIG_NET_VENDOR_MELLANOX is not set
+ CONFIG_NET_VENDOR_MELLANOX=y
+ # CONFIG_MLX4_EN is not set
+ CONFIG_MLX5_CORE=m
+ # CONFIG_MLX5_FPGA is not set
+ CONFIG_MLX5_CORE_EN=y
+ CONFIG_MLX5_EN_ARFS=y
+ CONFIG_MLX5_EN_RXNFC=y
+ CONFIG_MLX5_MPFS=y
+ CONFIG_MLX5_ESWITCH=y
+ CONFIG_MLX5_CLS_ACT=y
+ CONFIG_MLX5_CORE_EN_DCB=y
+ # CONFIG_MLX5_CORE_IPOIB is not set
+ # CONFIG_MLX5_IPSEC is not set
+ CONFIG_MLX5_SW_STEERING=y
+ # CONFIG_MLXSW_CORE is not set
+ CONFIG_MLXFW=m
  # CONFIG_NET_VENDOR_MICREL is not set
  # CONFIG_NET_VENDOR_MICROCHIP is not set
  # CONFIG_NET_VENDOR_MICROSEMI is not set
@@ -4026,6 +4041,7 @@
  # CONFIG_INFINIBAND_MTHCA is not set
  # CONFIG_INFINIBAND_EFA is not set
  # CONFIG_MLX4_INFINIBAND is not set
+ CONFIG_MLX5_INFINIBAND=m
  # CONFIG_INFINIBAND_OCRDMA is not set
  # CONFIG_INFINIBAND_VMWARE_PVRDMA is not set
  # CONFIG_INFINIBAND_USNIC is not set
```

aarch64:
None, presumably it's already enabled from upstream kernel package for aarch64.

**Terms of contribution:**


packages/kernel-5.10/config-bottlerocket
```diff
@@ -40,6 +40,21 @@ CONFIG_IXGBE_DCB=y
 CONFIG_IXGBE_HWMON=y
 CONFIG_IXGBEVF=m
 
+# InfiniBand protocol support
+CONFIG_INFINIBAND=m
+CONFIG_INFINIBAND_USER_VERBS=y
```
bcressey: Looking at kernel 5.10 sources, this doesn't seem to be defined anymore.

I'd recommend looking at the build output for the 5.10 kernel build and checking for warnings and the like - there's usually output to show that you tried to define something that isn't used.

```suggestion
```

packages/kernel-5.10/config-bottlerocket
```diff
@@ -40,6 +40,21 @@ CONFIG_IXGBE_DCB=y
...
+
+# Mellanox network support
+CONFIG_MLXFW=m
+CONFIG_MLX5_CORE=m
+CONFIG_MLX5_INFINIBAND=m
+CONFIG_PCI_HYPERV_INTERFACE=m
```
bcressey: Why add this?

```suggestion
```

packages/kernel-5.10/config-bottlerocket
```diff
@@ -40,6 +40,21 @@ CONFIG_IXGBE_DCB=y
...
+CONFIG_NET_VENDOR_MELLANOX=y
+CONFIG_MLX5_CORE_EN=y
+CONFIG_MLX5_EN_IPSEC=y
+CONFIG_MLX5_TLS=y
+CONFIG_MLX5_EN_TLS=y
```
bcressey: These don't end up in the final config:
```suggestion
```

packages/kernel-5.10/config-bottlerocket
```diff
@@ -40,6 +40,21 @@ CONFIG_IXGBE_DCB=y
 CONFIG_IXGBE_HWMON=y
 CONFIG_IXGBEVF=m
 
+# InfiniBand protocol support
+CONFIG_INFINIBAND=m
+CONFIG_INFINIBAND_USER_VERBS=y
+
+# Mellanox network support
+CONFIG_MLXFW=m
```
bcressey, line 44: To sync with the arm64 config, I think we potentially want to set these:
```
CONFIG_NET_SWITCHDEV=y
CONFIG_MLX5_BRIDGE=n
```

packages/kernel-5.10/config-bottlerocket
etungsten: `pci_hyperv_intf` is used by `mlx5_core`

https://cateee.net/lkddb/web-lkddb/MLX5_CORE.html lists `( CONFIG_PCI_HYPERV_INTERFACE|| ! CONFIG_PCI_HYPERV_INTERFACE )` as a dependency.

etungsten: Though I guess the condition implies it could also be `=n`? I'm not sure.

zmrow (approved):
Looks good pending a test!

:mage:

bcressey (approved):

packages/kernel-5.10/config-bottlerocket
```diff
@@ -40,6 +40,15 @@ CONFIG_IXGBE_DCB=y
...
 CONFIG_IXGBEVF=m
 
+# Mellanox network support
+CONFIG_MLXFW=m
+CONFIG_MLX5_CORE=m
+CONFIG_MLX5_INFINIBAND=m
+CONFIG_NET_VENDOR_MELLANOX=y
+CONFIG_MLX5_CORE_EN=y
+CONFIG_NET_SWITCHDEV=y
+CONFIG_MLX5_BRIDGE=n
```
foersleo: This option would not be interesting for 5.10, as it was introduced first in 5.14.

packages/kernel-5.10/config-bottlerocket
```diff
@@ -40,6 +40,15 @@ CONFIG_IXGBE_DCB=y
 CONFIG_IXGBE_HWMON=y
 CONFIG_IXGBEVF=m
 
+# Mellanox network support
+CONFIG_MLXFW=m
+CONFIG_MLX5_CORE=m
+CONFIG_MLX5_INFINIBAND=m
```
foersleo, line 46: I am not sure if Infiniband is actually needed. If I read the Connect X-4 page at nvidia [1] correctly the Cards mentioned in the Overview do not support Infiniband. Connect X-[5,6,7] seem to be the infiniband offerings by Mellanox [2].

[1] https://www.nvidia.com/en-us/networking/ethernet/connectx-4-lx/
[2] https://www.nvidia.com/en-us/networking/infiniband-adapters/

etungsten: I think I was mistaken in assuming it was Connect X-4. I was just basing it off of a HP DL160 spec file I found but apparently DL160 has the option for Connect X-5 which is what the machine needing the driver has (see [here](https://www.hpe.com/psnow/doc/c04154440.pdf?jumpid=in_lit-psnow-getpdf)). When the machine booted with ubuntu, we saw MLX5 drivers get loaded.

(I fixed the PR description)

packages/kernel-5.10/config-bottlerocket
etungsten: Does that mean it's safe to assume it defaults to `=n` and we wouldn't need to specify it here?

foersleo: In a way, yes.

The option does not exist in kernels prior to 5.14. It will be ignored as there is no code that switches anything off or on, no matter what we choose here. It will probably produce something like an unknown option log line somewhere, that is all.

No occurrence of `MLX5_BRIDGE` in the kernel for 5.4.124:

```
HEAD is now at 4f3fee72a74c Linux 5.10.124
[foersleo@devbox linux]$ grep -nr MLX5_BRIDGE *
[foersleo@devbox linux]$ 
```

On kernels after 5.14 however, it will default to `=y`:

```
config MLX5_BRIDGE
	bool
	depends on MLX5_ESWITCH && BRIDGE
	default y
	help
	  mlx5 ConnectX offloads support for Ethernet Bridging (BRIDGE).
	  Enable adding representors of mlx5 uplink and VF ports to Bridge and
	  offloading rules for traffic between such ports. Supports VLANs (trunk and
	  access modes).
```

There is some handy webpages, that can show some info on which kernel versions have certain options, without having to go through the versions oneself. So far I have not seen any problems with regards to reliability of the data (probably automagically created). For example https://cateee.net/lkddb/web-lkddb/MLX5_BRIDGE.html

etungsten:
Push above removes `CONFIG_MLX5_BRIDGE=n` as it's not an existing option for the 5.10 kernel.

markusboehme (approved):

PR #2242, opened 2022-06-24
schnauzer: add support for container registries in China

jpculp:
**Description of changes:**

This adds logic to support regions in China by adding a partition map
for helper scripts to make adjustments where necessary. The ECR map was
also updated with registries for `cn-north-1` and `cn-northwest-1`.

**Testing done:**

Built `aws-k8s-1.21` variant, launched in each partition, and verified host containers and pause containers were pulled successfully from the correct sources.

**Terms of contribution:**

### DIFF
--- a/sources/api/schnauzer/src/helpers.rs
+++ b/sources/api/schnauzer/src/helpers.rs
@@ -27,6 +27,8 @@ lazy_static! {
         m.insert("ap-southeast-2", "328549459982");
         m.insert("ap-southeast-3", "386774335080");
         m.insert("ca-central-1", "328549459982");
+        m.insert("cn-north-1", "183470599484");
+        m.insert("cn-northwest-1", "183901325759");
         m.insert("eu-central-1", "328549459982");
         m.insert("eu-north-1", "328549459982");
         m.insert("eu-south-1", "586180183710");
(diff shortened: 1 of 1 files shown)


sources/api/schnauzer/src/helpers.rs
```diff
@@ -90,6 +92,44 @@ lazy_static! {
 const PAUSE_FALLBACK_REGISTRY: &str = "602401143452";
 const PAUSE_FALLBACK_REGION: &str = "us-east-1";
 
+lazy_static! {
+    /// A map to tell us which partition from for a given region.
+    static ref PARTITION_MAP: HashMap<&'static str, &'static str> = {
+        let mut m = HashMap::new();
```
etungsten: Non-blocking comment: I was looking at `aws-sdk-rust` to see if we can get this information from their region types but it seems like the answer is no. So I guess we'll have to live with this. I really don't like the proliferation of these kinds of static hash maps that should be sourcing the truth from elsewhere.

etungsten (approved):

sources/api/schnauzer/src/helpers.rs
jpculp: I agree. I did add a method of fetching the partition from IMDS in https://github.com/bottlerocket-os/bottlerocket/pull/2224, but based on prior discussions the consensus was this might be more performant, especially if multiple settings need partitional context.

rpkelly: Another non-blocking comment: I would also prefer not to see a large hashmap like this, and would prefer that it only list the exceptions (in this case, the aws-cn and aws-us-gov regions) rather than need to be updated each time a new region was built.

rpkelly (approved):

jpculp:
- Rather than mapping out every region, just map out the non-standard regions.

etungsten (approved):

rpkelly (approved):

bcressey (approved):

Issue #1255, opened 2020-12-29
Please make it available in cn-north-1 and cn-northwest-1


terryares:
Hi team,

Do you have a plan about when bottlerocket is supported in China region?

jhaynes:
Thanks for asking about the China region. We're excited to bring Bottlerocket into the other AWS partitions, but I can't commit to a timeline right now.

walkley:
Hey, any updates to make Bottlerocket available in AWS China regions?

zmrow:
This is still on our radar, but we don't have any dates to share at this time!

jpculp:
@arestuzi & @walkley, I'm happy to announce that with the release of [**1.9.0**](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.9.0), AMIs are available in `cn-north-1` and `cn-northwest-1`!

kdaula:
Closing issue now that Bottlerocket AMIs are available in cn-north-1 and cn-northwest-1

PR #2247, opened 2022-06-27
Generate `settings.metrics.send-metrics` based on AWS partition for AWS variants

cbgbt:
**Issue number:** Related to #1255

**Description of changes:**
This set of changes modifies `settings.metrics.send-metrics` to use a setting-generator for AWS variants. Specifically, we enable metrics-sending by default on AWS variants if the host is in the `aws` or `aws-us-gov` partitions, according to the IMDS.

This information is retrieved by `shibaken`, which required some changes to the interface, as well as the admin user-data generation which uses that interface.

**Testing done:**
* [x] Checked that admin container user-data is still generated successfully
* [x] Checked that `send-metrics` successfully uses shibaken in the `aws` partition
* [x] Checked that `send-metrics` successfully uses shibaken in the `aws-us-gov` partition
* [x] Tested migrations with a test TUF repo

**Terms of contribution:**

### DIFF
--- a/sources/api/migration/migration-helpers/src/common_migrations.rs
+++ b/sources/api/migration/migration-helpers/src/common_migrations.rs
@@ -1145,3 +1145,227 @@ mod test_replace_metadata_list {
         .unwrap_err();
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// We use this migration when we need to replace a single metadata value;
+/// for example, what a release changes the `setting-generator` associated with a setting.
+// String is the only type we use today, and handling multiple value types is more complicated than
+// we need at the moment.  Allowing &[serde_json::Value] seems nice, but it would allow arbitrary
+// data transformations that the API model would then fail to load.
+
+#[derive(Debug, Clone)]
+pub struct MetadataReplacement {
+    pub setting: &'static str,
+    pub metadata: &'static str,
+    pub old_val: &'static str,
+    pub new_val: &'static str,
+}
+
+impl MetadataReplacement {
+    /// Executes the metadata replacement on given datastore.
+    ///
+    /// State which prevents the replacement from being performed results in messages to stdout.
+    /// Returns whether or not the migration performed changes.
+    fn perform_replacement(&self, input: &mut MigrationData) -> bool {
+        input
+            .metadata
+            .get_mut(self.setting)
+            .or_else(|| {
+                println!("Found no setting '{}'", self.setting);
+                None
+            })
+            .and_then(|found_metadata| {
+                let metadata_value = found_metadata.get_mut(self.metadata);
+                if metadata_value.is_none() {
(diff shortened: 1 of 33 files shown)


jpculp:
It might be a good idea to split `metrics.toml` into three: `metrics-aws.toml` for aws-specific defaults, `metrics-public.toml` for non-aws defaults, and keep `metrics.toml` for shared values.

sources/models/src/aws-dev/defaults.d/31-send-metrics-aws.toml
```diff
@@ -0,0 +1 @@
+../../../shared-defaults/send-metrics-aws.toml
```
jpculp, line 1: nit. Do we want the filename to be this specific? If we decide to regionalize other metrics aspects we would need to create yet another toml.

sources/models/src/metal-k8s-1.22/defaults.d/31-send-metrics.toml
```diff
@@ -0,0 +1 @@
+../../../shared-defaults/send-metrics-global.toml
```
jpculp, line 1: nit. We use 'public' elsewhere, but 'global' is also accurate.

jpculp (approved):

sources/api/shibaken/src/admin_userdata.rs
```diff
@@ -0,0 +1,71 @@
+/// This module contains utilities for populating userdata for the admin-container with user SSH keys from IMDS.
+use clap::Parser;
```
bcressey: Elsewhere we've tried to use `argh` for binary size reasons - can we do that here? `structopt` is also OK since it's currently in `updog` and `metricdog`.

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -1145,3 +1145,227 @@ mod test_replace_metadata_list {
         .unwrap_err();
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// We use this migration when we need to replace a single metadata value;
+/// for example, what a release changes the `setting-generator` associated with a setting.
```
bcressey, line 1152: Very nice!

etungsten (approved):
LGMT once existing comments are addressed.

sources/models/src/aws-dev/defaults.d/31-send-metrics-aws.toml
cbgbt: I think for now I'll leave the filename, since other metrics metadata is currently shared, but we can pretty easily change it if we do an expansion.

cbgbt:
Addresses comment by @bcressey  and rebases.

Moving from `clap` to `argh` reduced the aarch64 binary size from 4.1MB to 3.7MB. Not bad!

bcressey (approved):
Nice!

Issue #2211, opened 2022-06-13
Split metal and non-metal kernel config


zmrow:
As time goes on and we continue to add kernel modules to support new and different hardware, we don't want our non-metal variants to carry all of the extra modules.  We should split out the metal-specific config.  

There are a few considerations here however:
* We don't want a separate kernel package to maintain
* We need to ensure that the kernel gets re-built for metal/non-metal as required (will need to use the `variant-sensitive` key in `Cargo.toml`)

webern:
> will need to use the variant-sensitive key in Cargo.toml

We should consider some additional logic to ensure that the kernel is only rebuilt when switching between metal and other versions, and not rebuilt when switching between eg aws-ecs and aws-k8s

foersleo:
Yes, I think there is some steps to this here. I am thinking of something like the following:

- [x] Separate the config but have the same build with all options (Done with https://github.com/bottlerocket-os/bottlerocket/pull/2264)
- [x] Separate the builds to only have the metal specific configs build into the metal variants
- [ ] build time improvements *

(*) For the build time improvements there is multiple levels to that I think:
1. Re-use builds as much as possible (as mentioned by Matt above) - This will probably be some work on the capabilities of the build system
2. What parts of one build might we be able to re-use through something like ccache or similar. - This will need some experimentation on how such caching tools work with the kernel and changing configs.

foersleo:
I have been looking into how to split the builds.

Doing this the basic way of just adding some logic in the `kernel.spec` that determines when to merge the metal config and adding the `variant-sensitive` flag to the kernel package is straight forward and works in some testing I have done. However it also incurs some extra package rebuilds when switching variants within the `metal` platform or outside the `metal` platform.

So probably we want to think about adding some feature to specify a specific part of the variant (platform, runtime, family, ...) and a specific value for that part we want to have rebuilds of a specific package for. I have done some poking around, but I probably need to get more familiar with `buildsys` to understand how to implement that.

foersleo:
The functional parts of splitting the configs and building the kernel with the variant appropriate config are done. As discussed in https://github.com/bottlerocket-os/bottlerocket/pull/2279 the impact of re-building the kernel with each variant is less than expected as the variant specific build of the `os` package takes similar amounts of time as the kernel build.

I will either way have a look on how we might be able to recoup some of the additional time spent in building with this split.

bcressey:
[...]

PR #2264, opened 2022-07-05
kernel: config: split bare metal specific config

foersleo:
```
Split out bare metal specific config options. All these are enabling
hardware support, mostly for specific storage or networking hardware
that is not available in settings where the non-metal variants are
used. For now split out the metal specific config options to later
enable only building those drivers for the metal variants and keep
the non-metal variants lean.

This moves the settings originally introduced in commits

    b2e64488 Merge pull request #2243 from zmrow/broadcom_bnxt
    6ab5fb62 kernel-5.10: enable mellanox modules
    3be82c94 kernel 5.10: Add `bnxt` module for Broadcom 10/25Gb network
    58d15ce0 kernel-5.10: add support for Microsemi PQI
    a69bcfa3 kernel: adjust option for megaraid sas
    d0d2c68e kernel-5.10: enable support for broadcom ethernet cards
    31d50628 kernel: add support for MegaRAID SAS
    928a481b kernel 5.10: Add config to support additional hardware

Signed-off-by: Leonard Foerster <foersleo@amazon.com>
```

**Issue number:** https://github.com/bottlerocket-os/bottlerocket/issues/2211

**Description of changes:**

This is a non-functional change moving all the bare metal specific configs (mostly drivers for certain hardware) into its own kernel config fragment. The new fragment gets merged into the config so we do not lose these settings.

Right now this is not making a difference between variants for the build - all of them still get all the config options.

**Testing done:**

Built variants for `x86_64` and `aarch64` with and without the patch. The diff for the generated kernel config files is empty as expected.

**Terms of contribution:**


packages/kernel-5.10/config-bottlerocket
```diff
@@ -77,11 +30,6 @@ CONFIG_EFI=y
 CONFIG_EFI_STUB=y
 CONFIG_EFI_MIXED=y
 
-# EFI video
-CONFIG_FB=y
-CONFIG_FB_EFI=y
-CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER=y
```
bcressey: I'd like to keep the EFI related variables in the common config, since they apply to platforms like EC2 and ESXi when using EFI boot.

packages/kernel-5.10/config-bottlerocket
```diff
@@ -8,47 +8,6 @@ CONFIG_EXT4_FS=y
...
-# SATA support
-CONFIG_BLK_DEV_SD=y
-CONFIG_SATA_AHCI=y
-CONFIG_ATA=y
-CONFIG_ATA_PIIX=y
-
-# Network support
-CONFIG_ETHERNET=y
-CONFIG_NET_CORE=y
-CONFIG_NETDEVICES=y
```
bcressey, line 20: Are these actually necessary to define? They all seem pretty essential. It might be better to drop them.

foersleo: Right, we can drop these. They are selected anyways.

packages/kernel-5.10/config-bottlerocket
foersleo: That makes sense. These were added when initially adding more hardware support, but probably were not just hardware enablement.

foersleo:
And there I fat fingered the updates for this PR.

So first update has the actual changes as requested by @bcressey, but I had some weird extra state due to screwed up rebases.
Doing that properly again fixed this. Sorry for the noise.

markusboehme (approved):

bcressey (approved):

PR #2270, opened 2022-07-10
fix ntp service restart when settings change

bcressey:
**Issue number:**
N/A

**Description of changes:**
Since the service is named "ntp" rather than "chronyd", the restart commands were not run when NTP settings changed.

**Testing done:**
Confirmed that both newly launched and upgraded hosts restart the NTP service after settings change.

**Terms of contribution:**

### DIFF
--- a/Release.toml
+++ b/Release.toml
@@ -130,3 +130,6 @@ version = "1.8.0"
     "migrate_v1.8.0_public-admin-container-v0-9-0.lz4",
     "migrate_v1.8.0_public-control-container-v0-6-1.lz4",
 ]
+"(1.8.0, 1.9.0)" = [
+    "migrate_v1.9.0_ntp-affected-services.lz4",
+]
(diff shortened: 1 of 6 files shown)


mchaker (approved):

markusboehme (approved):
This PR was a helpful exercise for me. Leaving this here for someone to correct me if my understanding is wrong: The mix-up happened because Bottlerocket's API and model just talk of an NTP service in the abstract, whereas currently this is implemented via Chrony (`chrony` systemd service). The API and thar-be-settings only care about the generic service name. The fact it's Chrony is hidden in the model and considered an implementation detail.

etungsten (approved):

bcressey:
> The API and thar-be-settings only care about the generic service name. The fact it's Chrony is hidden in the model and considered an implementation detail.

That's right!

PR #2276, opened 2022-07-14
defer squashfs mounts

bcressey:
**Issue number:**
N/A

**Description of changes:**
Mounting `/usr/share/licenses` and `/usr/src/kernels` can add up to one second of delay before reaching `local-fs.target`.

It's not necessary to read software licenses or build out-of-tree kernel modules at this stage in the boot process. Moving the mounts after `local-fs.target` doesn't eliminate the delay, but does allow the system to do other useful work in parallel, like acquiring DHCP leases and applying settings.

**Testing done:**
I verified that `link-kernel-modules.service` and `load-kernel-modules` run correctly on the `aws-k8s-1.22-nvidia` variant.

It's difficult to quantify the actual benefit of this since it can vary widely, and depends on other factors like the amount of console logging and whether the wrong kernel thread gets "stuck" flushing to the console when these messages are printed:
```
[    4.630105] loop: module loaded
[    4.644181] squashfs: version 4.0 (2009/01/31) Phillip Lougher
```

Instead I'll just point to the actual mount unit times:
```
 275ms var-lib-kernel\x2ddevel-.overlay-lower.mount
 274ms x86_64\x2dbottlerocket\x2dlinux\x2dgnu-sys\x2droot-usr-share-licenses.mount
 146ms sys-kernel-tracing.mount
 145ms tmp.mount
 140ms etc-cni.mount
 140ms sys-kernel-debug.mount
 140ms dev-mqueue.mount
 139ms dev-hugepages.mount
  24ms opt-cni-bin.mount
  22ms var-lib-bottlerocket.mount
  21ms etc-containerd.mount
  20ms x86_64\x2dbottlerocket\x2dlinux\x2dgnu-sys\x2droot-usr-lib-modules.mount
  18ms etc-host\x2dcontainers.mount
  17ms sys-kernel-config.mount
  17ms sys-fs-fuse-connections.mount
  14ms local.mount
  11ms mnt.mount
   9ms opt.mount
   8ms var.mount
   5ms x86_64\x2dbottlerocket\x2dlinux\x2dgnu-sys\x2droot-usr-src-kernels.mount
```

These are consistently the two slowest mounts, and intuitively it is preferable to deal with slow mounts later in boot if possible, to avoid unlucky sequences where the system stalls for 300 ms before reaching `local-fs.target`.

**Terms of contribution:**

### DIFF
--- a/packages/os/link-kernel-modules.service.in
+++ b/packages/os/link-kernel-modules.service.in
@@ -1,5 +1,6 @@
 [Unit]
 Description=Link additional kernel modules
+RequiresMountsFor=PREFIX/lib/modules PREFIX/src/kernels
 # Rerunning this service after the system is fully loaded will override
 # the already linked kernel modules. This doesn't affect the running system,
 # since kernel modules are linked early in the boot sequence, but we still
--- a/packages/os/load-kernel-modules.service.in
+++ b/packages/os/load-kernel-modules.service.in
@@ -1,5 +1,6 @@
 [Unit]
 Description=Load additional kernel modules
+RequiresMountsFor=PREFIX/lib/modules PREFIX/src/kernels
 After=link-kernel-modules.service
 Requires=link-kernel-modules.service
 # Disable manual restarts to prevent loading kernel modules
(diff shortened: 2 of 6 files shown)


zmrow (approved):
Neat! :camera_flash:

arnaldo2792 (approved):

Issue #2228, opened 2022-06-22
Deprecate aws-k8s-1.19 variant by August 2022


cbgbt:
Kubernetes 1.19 went out of upstream support in August, 2021. AWS EKS is maintaining support for 1.19 until [August 1st, 2022](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#kubernetes-release-calendar). Bottlerocket will deprecate support of our `aws-k8s-1.19` variant at that time.

While you’ll still be able to launch `aws-k8s-1.19` nodes beyond August 1st, 2022, we’ll no longer provide updated `aws-k8s-1.19` images, and the Bottlerocket repositories will no longer receive new updates for nodes running `aws-k8s-1.19`.

We recommend you replace existing Bottlerocket `aws-k8s-1.19` nodes with [a newer variant](https://github.com/bottlerocket-os/bottlerocket/tree/develop/variants#variants) such as `aws-k8s-1.22` - ideally the latest your Kubernetes cluster supports.

For instructions on how to migrate your nodes to newer supported versions, please consult the [EKS documentation for updating clusters](https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html) or reach out to us through your AWS support channels for additional questions.

PR #2277, opened 2022-07-14
use "quiet" kernel parameter for aws and vmware

bcressey:
**Issue number:**
N/A

**Description of changes:**
Use the "quiet" parameter for platforms where the hardware is known and the kernel output is not generally needed on the console in order to diagnose issues. This includes the "aws" and "vmware" variants.

The "metal" variants continue to use the more verbose default output so that any hardware without matching driver support can be detected.

Force systemd to always print the status of units, since this can be invaluable data when troubleshooting failures caused by dependency ordering problems.

Remove the printk settings from the default sysctls, since these will be applied by `systemd` very early and override the "quiet" behavior. The default values in `/proc/sys/kernel/printk` are controlled by the kernel config, and will be "7 4 1 7" without the "quiet" parameter, and "4 4 1 7" when it is present.

The main motivation for this is to improve boot performance and make it more consistent. printk logging is expensive, and can slow down many code paths in the kernel.

**Testing done:**
Verified that we still get reasonable systemd output in the "success" and "failure" cases:
https://gist.github.com/bcressey/1b8889c772cef9d88ce37b38f6051b0a

Overall this change saves at least two seconds when booting an m5ad.large instance in EC2. I ignored the results from the first boot to rule out the impact of populating the EBS volume with snapshot blocks.

Measurements were done with `systemctl show --all | grep UserspaceTimestampMonotonic`, which is what `systemd-analyze` uses as the zero point for its own measurements.

Before:
```
UserspaceTimestampMonotonic=1815492
UserspaceTimestampMonotonic=1711044
UserspaceTimestampMonotonic=1806789
```

After:
```
UserspaceTimestampMonotonic=598559
UserspaceTimestampMonotonic=594318
UserspaceTimestampMonotonic=599811
```

**Terms of contribution:**

### DIFF
--- a/variants/aws-k8s-1.19/Cargo.toml
+++ b/variants/aws-k8s-1.19/Cargo.toml
@@ -15,6 +15,7 @@ kernel-parameters = [
     "console=ttyS0,115200n8",
     "net.ifnames=0",
     "netdog.default-interface=eth0:dhcp4,dhcp6?",
+    "quiet",
 ]
 included-packages = [
     "aws-iam-authenticator",
--- a/packages/release/release-sysctl.conf
+++ b/packages/release/release-sysctl.conf
@@ -1,7 +1,4 @@
 ## AL2 defaults ##
-# Maximize console logging level for kernel printk messages
-kernel.printk = 8 4 1 7
-
 # Wait 10 seconds and then reboot
 kernel.panic = 10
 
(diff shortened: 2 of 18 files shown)


zmrow (approved):
:night_with_stars:

foersleo (approved):

markusboehme (approved):
> Overall this change saves at least two seconds when booting an m5ad.large instance in EC2.

This is an impressive improvement, even more so for such a conceptually simple change! Nice!

variants/aws-k8s-1.19/Cargo.toml
```diff
@@ -15,6 +15,7 @@ kernel-parameters = [
     "console=ttyS0,115200n8",
     "net.ifnames=0",
     "netdog.default-interface=eth0:dhcp4,dhcp6?",
+    "quiet",
```
arnaldo2792, line 18: NIT: can we skip this change? The variant will be deprecated by the next month:

#2228

arnaldo2792 (approved):

PR #2278, opened 2022-07-15
avoid unnecessary systemd reloads

bcressey:
**Issue number:**
N/A

**Description of changes:**
`systemctl daemon-reload` can be expensive, ranging from 200-500 ms.

On an `aws-k8s` variant, we issue five reloads during the boot:
1. when enabling or disabling the "admin" host container
2. when enabling or disabling the "control" host container
3. when running restart commands for `kubelet`
4. when switching to `configured.target`
5. when switching to `multi-user.target`

These changes eliminate the first three reloads, for a savings of roughly 600ms.

**Testing done:**
Verified that the "admin" and "control" containers are correctly enabled or disabled during boot, and that no `systemctl daemon-reload` runs.

Verified that host containers are correctly enabled or disabled after boot, and that `systemctl daemon-reload` runs, and that `systemctl show host-containers@foo` reflects the expected state.

Verified that `systemctl daemon-reload` is not run when configuring `kubelet` during boot, and that it does run afterwards, and that changes to command-line parameters are picked up.

Verified that the restart command for `kubelet` is updated on upgrade, and restored on downgrade.

**Terms of contribution:**

### DIFF
--- a/Release.toml
+++ b/Release.toml
@@ -135,4 +135,5 @@ version = "1.8.0"
     "migrate_v1.9.0_shibaken-admin-userdata-semantics.lz4",
     "migrate_v1.9.0_shibaken-send-metrics.lz4",
     "migrate_v1.9.0_image-gc-thresholds.lz4",
+    "migrate_v1.9.0_kubelet-no-daemon-reload.lz4",
 ]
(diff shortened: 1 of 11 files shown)


bcressey:
Force push above applies `cargo fmt`.

zmrow (approved):
:kangaroo:

etungsten (approved):

PR #2280, opened 2022-07-15
docs: Add quickstart guide for local QEMU/KVM VMs

markusboehme:
**Issue number:** #834

**Description of changes:** Commit e2b58d60 ("tools: Add start-local-vm script") added a wrapper script to simplify running a locally built Bottlerocket image in a local VM using QEMU and KVM. Provide a quickstart guide analogous to the existing ones for EKS, ECS, and VMware to help make use of it.

The guide currently assumes anyone following along to be running Fedora. Changes to the guide for other distros will be minimal, if they are needed at all. However, changes in start-local-vm are likely to be required. Leave this to a dedicated later change.

**Testing done:** Put steps together while building a Bottlerocket image on a new c6g.metal instance and launching a local VM with it afterwards.

**Terms of contribution:**

### DIFF
--- a/QUICKSTART-LOCAL.md
+++ b/QUICKSTART-LOCAL.md
@@ -1,0 +1,169 @@
+# Testing Bottlerocket in a local virtual machine
+
+This quickstart will walk through launching a Bottlerocket VM guest on a local machine using QEMU and KVM.
+The VM will not join an ECS or Kubernetes cluster.
+This way of running Bottlerocket is therefore best used for testing purposes when developing Bottlerocket components that do not need to integrate with any orchestrators or to just get a feel for what a Bottlerocket node looks from the inside.
+
+
+## Prerequisites
+
+We assume you are following along on a machine running Fedora.
+If you are using a cloud VM, ensure you can use hardware-assisted virtualization.
+For example, on Amazon EC2 this requires the use of a .metal instance type.
+
+You need a clone of the main Bottlerocket repository and a build of the metal-dev variant.
+Please refer to [`BUILDING.md`](https://github.com/bottlerocket-os/bottlerocket/blob/develop/BUILDING.md) for instructions on how to build a Bottlerocket image and ensure you pass `-e BUILDSYS_VARIANT=metal-dev` to `cargo make`.
+
+The use of QEMU requires extra packages which you may install using this dnf invocation:
+
+```
+sudo dnf install qemu
+```
+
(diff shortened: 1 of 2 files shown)


QUICKSTART-LOCAL.md
```diff
@@ -0,0 +1,169 @@
...
+```
+PUBKEY_FILE="${HOME}/.ssh/id_rsa.pub"
+PUBKEY=$(< "${PUBKEY_FILE}")
+ADMIN_USER_DATA="$(echo '{"ssh": {"authorized-keys": ["'"${PUBKEY}"'"]}}' | base64 -w 0)"
+
+cat <<EOF >>user-data.toml
+[settings.host-containers.admin]
+enabled = true
+user-data = "${ADMIN_USER_DATA}"
+source = "public.ecr.aws/bottlerocket/bottlerocket-admin:v0.9.0"
```
markusboehme, line 76: This is a bit awkward because it needs to be kept up to date. However, it seems the defaults defined in `sources/models/shared-defaults/public-host-containers.toml` are replaced entirely with the `user-data.toml` placed directly on the file system, i.e. if the `source` is left off here, thar-be-settings will notice and error out on boot.

1. Any ideas for forcing a merge?
2. The public admin and control containers don't have a `latest` tag which would be helpful here. Are there downsides to introducing one and referring to it instead?

zmrow: Of course we would like to keep this reasonably up to date, but for the purposes of this guide using that container version is not going to break at any point in the future.

The wider ramifications of creating a `latest` tag are probably not worth it for the purpose of setting up the host containers in a quickstart.

zmrow (commented):
Nice work!

QUICKSTART-LOCAL.md
markusboehme: > The wider ramifications of creating a `latest` tag are probably not worth it for the purpose of setting up the host containers in a quickstart.

Fully agree with that on effort alone, and I couldn't think of any other use: It makes sense for a Bottlerocket release to define those explicitly to avoid surprise updates and have everything be tested in combination. As a container newbie I couldn't help but wondering out loud whether there are any downsides, though. :-)

stockholmux: For what it's worth, this should probably be merged into the documentation/website repo (once that is available). There we can interpolate the latest version, so if we update the repo, it updates the docs.

zmrow (approved):
:cocktail:

stockholmux (approved):
LGTM

Issue #2214, opened 2022-06-13
Support versioning in `net.toml`


zmrow:
As we require additional network config options, we'll need to bump the version in `net.toml`.  The code needs to support deserialization of different versioned files.

PR #2281, opened 2022-07-15
netdog: Support versioning in `net.toml`

zmrow:
**Issue number:**
Implements #2214 

**Description of changes:**
```
This change adds the ability to handle additional versions of `net.toml`
as the project continues to add features to network configuration.  A
top level `NetConfig` type has been removed in favor of a few traits.
Each new version of config is expected to implement these traits, the
most important of which converts the network configuration into
`WickedInterface` structs which are suitable for serializing directly to
file.

As part of this change, the `net_config` module has been lightly
refactored and split into submodules to make the pieces easier to use
among new versions.
```

**Testing done:**
* Build and boot an `aws-k8s-1.21 variant` and ensure the network config is correct
* Build and boot a `metal-dev` variant and ensure the network config is correct via a given `net.toml`

**Terms of contribution:**


sources/api/netdog/src/interface_name.rs
```diff
@@ -10,16 +10,16 @@ use std::ops::Deref;
 
 /// InterfaceName can only be created from a string that contains a valid network interface name.
 /// Validation is handled in the `TryFrom` implementation below.
-#[derive(Debug, Eq, PartialEq, Hash, Deserialize)]
-#[serde(try_from = "&str")]
+#[derive(Clone, Debug, Eq, PartialEq, Hash, Deserialize)]
+#[serde(try_from = "String")]
```
zmrow, line 14: Since we now deserialize the majority of the network config to a `toml::Value` first (to handle the versions), this is stored as a `String` inside of the `toml::Value`, hence the change from `&str` to `String`.

sources/api/netdog/src/net_config/v1.rs
```diff
@@ -0,0 +1,274 @@
...
+            error::InvalidNetConfigSnafu {
+                reason: "multiple primary interfaces defined, expected 1"
+            }
+        );
+
+        Ok(())
+    }
+}
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
```
zmrow, line 100: Nothing has really changed about the below code.   This code was moved here into `v1.rs` since we use the `NetConfigV1` struct for the simple configuration we get from the kernel command line.

zmrow:
The above force push:
* changes the return type of `deserialize_config` from `impl Interfaces` to `Box<dyn Interfaces>` because we'll be returning different `Interface` implementors (netconfigv1, v2, etc).  Rust considers them different types which means we need to use a `Box` rather than `impl trait`.
* Adds a small `validate_config()` function to simplify the `match` statement when deserializing config
* Implements the `Validate` and `Interfaces` traits for `Box<T>`

sources/api/netdog/src/net_config/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    Ok(Some(net_config))
+}
+
+/// Deserialize the network config, using the version key to determine which config struct to
+/// deserialize into
+fn deserialize_config(config_str: &str) -> Result<Box<dyn Interfaces>> {
+    #[derive(Debug, Deserialize)]
+    struct ConfigToml {
+        version: u8,
```
etungsten, line 85: Is version as a field mandatory? Can we assume it's the latest version of net_config if it's not specified?

etungsten (approved):

jpculp (approved):

arnaldo2792 (approved):

sources/api/netdog/src/net_config/mod.rs
zmrow: Version is mandatory currently, to keep things explicit.

Some considerations:
* In order to know what "latest" is at any given time we'd need to maintain a constant or something that would we would need to remember to change.
* Without the version field, we'd need to attempt to deserialize into a struct with a version field first, and if that failed for some reason we'd attempt to deserialize into "latest"?

In this case I think I'd rather be explicit and avoid any guessing. :)

Issue #2282, opened 2022-07-16
Documentation: README needs linting


mchaker:
**What I'd like:**

The `README.md` file contains a mix of list item characters (both `*` and `-`) and some paragraphs don't have newlines between them.

I would like the `README.md` file to go through a linter.

PR #2283, opened 2022-07-16
Docs: README Markdown linting

mchaker:
**Issue number:**

#2282 

**Description of changes:**

Standardize list item characters to `*` and add newlines
between paragraphs. Automatic linting from within VS Code.

**Testing done:**

eyeballs only 👀 

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Bottlerocket-specific additions focus on reliable updates and on the API.
 Instead of making configuration changes manually, you can change settings with an API call, and these changes are automatically migrated through updates.
 
 Some notable features include:
+
 * [API access](#api) for configuring your system, with secure out-of-band [access methods](#exploration) when you need them.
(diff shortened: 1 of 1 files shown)


jpculp (approved):
Nice. It might be worth iterating through some of the other README's that aren't auto-generated as well. (see `sources/api` and `sources/updater`)

stmcginnis (approved):
This will be good if we decide to enable some sort of markdown lint checking.

zmrow (approved):

mchaker:
I did some digging and the linter is markdownlint, using Microsoft Docs Authoring Pack rules in VScode. 

Adding the rules to a pipeline/github action should be possible.

mchaker:
I found what I think are the rules used by the Microsoft Docs Authoring Pack VScode extension (applied via `markdownlint`): https://github.com/microsoft/vscode-docs-authoring/tree/main/packages/docs-linting/markdownlint-custom-rules

PR #2285, opened 2022-07-18
schnauzer: add support for update repositories in China

jpculp:
**Description of changes:**

This adds helpers that settings generators can use to set the `updates.metadata-base-url` and `updates.targets-base-url` values to regionalized repositories. For now, this only includes China regions `cn-north-1` and `cn-northwest-1`, while current regions will continue to use the public updates repository, updates.bottlerocket.aws.

**Testing done:**

Built and launched Bottlerocket and verified the expected `updates` settings in:

- [x] us-west-2
- [x] cn-north-1
- [x] vmware (thanks, @mchaker!)

**Terms of contribution:**

### DIFF
--- a/sources/api/schnauzer/src/helpers.rs
+++ b/sources/api/schnauzer/src/helpers.rs
@@ -93,6 +93,18 @@ const PAUSE_FALLBACK_REGISTRY: &str = "602401143452";
 const PAUSE_FALLBACK_REGION: &str = "us-east-1";
 
 lazy_static! {
+    /// A map to tell us which endpoint to pull updates from for a given region.
+    static ref TUF_ENDPOINT_MAP: HashMap<&'static str, &'static str> = {
+        let mut m = HashMap::new();
+        m.insert("cn-north-1", "bottlerocket-updates-cn-north-1.s3.dualstack");
+        m.insert("cn-northwest-1", "bottlerocket-updates-cn-northwest-1.s3.dualstack");
+        m
+    };
+}
+
+const TUF_PUBLIC_REPOSITORY: &str = "https://updates.bottlerocket.aws";
+
+lazy_static! {
     /// A map to tell us the partition for a given non-standard region.
     static ref ALT_PARTITION_MAP: HashMap<&'static str, &'static str> = {
         let mut m = HashMap::new();
@@ -131,17 +143,17 @@ mod error {
     #[derive(Debug, Snafu)]
     #[snafu(visibility(pub(super)))]
     pub(super) enum TemplateHelperError {
+        #[snafu(display("Expected an AWS region, got '{}' in template {}", value, template))]
+        AwsRegion {
+            value: handlebars::JsonValue,
+            template: String,
+        },
+
         #[snafu(display(
             "Expected ecr helper to be called with either 'registry' or 'region', got '{}'",
             value,
         ))]
         EcrParam { value: String },
-
-        #[snafu(display("Expected an AWS region, got '{}' in template {}", value, template))]
-        EcrRegion {
-            value: handlebars::JsonValue,
-            template: String,
-        },
 
         #[snafu(display(
             "Incorrect number of params provided to helper '{}' in template '{}' - {} expected, {} received",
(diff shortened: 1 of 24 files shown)


rpkelly (approved):

jpculp:
Added `updates.targets-base-url` migration.

jpculp:
Rebased.

sources/models/shared-defaults/aws-tuf.toml
```diff
@@ -0,0 +1,7 @@
+[metadata.settings.updates.targets-base-url]
+setting-generator = "schnauzer settings.updates.targets-base-url"
+template = "{{ tuf-prefix settings.aws.region }}/targets/"
+
+[metadata.settings.updates.metadata-base-url]
+setting-generator = "schnauzer settings.updates.metadata-base-url"
+template = "{{ tuf-prefix settings.aws.region }}{{ metadata-prefix settings.aws.region }}2020-07-07/{{ os.variant_id }}/{{ os.arch }}/"
```
bcressey: There's what looks like an inconsistency here between `tuf-prefix` - doesn't include trailing slash? - and `metadata-prefix` - which does?

Subjectively, "without trailing slash" seems cleaner since then the final structure is more obvious in the template.

jpculp:
Added symlinks for `k8s-1.23` variants.

etungsten (approved):

sources/models/shared-defaults/aws-tuf.toml
jpculp: I wasn't quite sure how handlebars would react with `metadata-prefix` being an empty string, but it turns out it's fine. The weirdness here is our regional repositories will have the metadata in a `metadata` directory that isn't present in the current public repository.

jpculp:
- Moved the trailing `/` out of `metadata-prefix` and added it to the template instead.
- Built and launched an `aws-k8s-1.23` variant in `us-west-2` to verify there isn't a space or something in the rendered setting.

etungsten (approved):

sources/api/schnauzer/src/helpers.rs
```diff
@@ -737,6 +749,98 @@ pub fn pause_prefix(
...
+        value: aws_region.to_owned(),
+        template: template_name,
+    })?;
+
+    // construct the prefix
+    let metadata_location = if TUF_BUCKET_MAP.contains_key(aws_region) {
+        "/metadata"
+    } else {
+        ""
+    };
```
bcressey: nit: IMO the template would be slightly more clear if we omit the leading `/` here:
```suggestion
    // construct the prefix
    let metadata_location = if TUF_BUCKET_MAP.contains_key(aws_region) {
        "metadata"
    } else {
        ""
    };
```

So then the template would be:
```
{{ tuf-prefix settings.aws.region }}/{{ metadata-prefix settings.aws.region }}/2020-07-07/{{ os.variant_id }}/{{ os.arch }}/
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1266,6 +1370,24 @@ fn pause_registry<S: AsRef<str>>(region: S) -> String {
...
+        Some(bucket) => (region.as_ref(), *bucket),
+    };
+    let partition = match ALT_PARTITION_MAP.borrow().get(region) {
+        None => (STANDARD_PARTITION),
+        Some(partition) => *partition,
+    };
+    match partition {
+        "aws-cn" => format!("https://{}.s3.{}.amazonaws.com.cn/latest", bucket, region),
+        _ => format!("https://{}.s3.{}.amazonaws.com/latest", bucket, region),
+    }
```
bcressey: Per discussion, we should try to use the dual stack endpoints here for forward compatibility with IPv6-only subnets.

sources/api/schnauzer/src/helpers.rs
jpculp: The problem there would be regions pointing to the current public repository would render as: `https://updates.bottlerocket.aws//2020-07-07/aws-k8s-1.23/x86_64/`.

jpculp:
Tweaked the code a bit to support virtual hosted-style dual-stack s3 endpoints for forward compatibility with IPv6-only subnets.

sources/api/schnauzer/src/helpers.rs
jpculp: Rather than just specifying the bucket, I appended `.s3.dualstack` to the strings directly _(and verified it still works for cn-north-1 and cn-northwest-1)_. In the event that we add a region that does not support dual stack endpoints we can just do `<bucket_id>.s3`. We could get fancier with the map, but I wanted to keep things simple.

etungsten (approved):

rpkelly (approved):

bcressey (approved):

Issue #2135, opened 2022-05-13
pubsys: cargo make ami should "just work"


webern:
**What I'd like:**

I have a aws profile with a default region, but when I use `cargo make ami` I get `Failed to build AMI: Infra.toml is missing aws.regions`. For the simple workflow `cargo make build && cargo make ami`, pubsys should be savvy and use my default region without needing me to create this file:

```toml
[aws]
regions = ["us-west-2"]
```

ecpullen:
You can use `-e PUBLISH_REGIONS="us-west-2"` as a workaround.

webern:
I didn't realize that, but I wonder if it would be possible to use my default region in the absence of that environment variable argument.

mchaker:
Closing this issue as per https://github.com/bottlerocket-os/bottlerocket/pull/2287#issuecomment-1197039705

PR #2287, opened 2022-07-18
pubsys: use default AWS region if none are given

mchaker:
**Issue number:**

#2135 

**Description of changes:**

Currently, pubsys returns an error (`Failed to build AMI: Infra.toml is missing aws.regions`) if the user does not either specify a target AWS region(s) as a command line parameter, or provide pubsys with an `Infra.toml` configuration file containing AWS region(s).

This change allows pubsys to acquire the user's default AWS region (if no regions are specified otherwise) for publishing AMIs (`cargo make ... ami`).

**Testing done:**

- [x] manual testing
  - specified only a default AWS region via `aws configure set region us-west-2` (no command-line parameters or `Infra.toml` passed to `cargo make ... ami`)
  - confirmed that pubsys picked my default AWS region when I attempted to upload an AMI via `cargo make ... ami`

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/ami/mod.rs
+++ b/tools/pubsys/src/aws/ami/mod.rs
@@ -100,10 +100,13 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
     let aws = infra_config.aws.unwrap_or_else(|| Default::default());
 
     // If the user gave an override list of regions, use that, otherwise use what's in the config.
+    // If there is nothing in the config, use the user's default AWS region.
     let mut regions = if !ami_args.regions.is_empty() {
         ami_args.regions.clone()
-    } else {
+    } else if !aws.regions.is_empty() {
         aws.regions.clone().into()
+    } else {
+        vec![Region::default().name().to_string()]
     }
     .into_iter()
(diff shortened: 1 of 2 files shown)


zmrow:
I think we need to be very careful about assuming a region, especially when it comes to publishing AMIs.  The danger here is that we publish an AMI to an unintended region by mistake.

The reason we specifically use an `Infra.toml` or an argument rather than as assumption is to make publishing to a region a very explicit action and avoid any unintended publishes.  I'd vote to keep the code as is, and encourage users to use an `Infra.toml`.

mchaker:
@zmrow Understandable. Would the following be an acceptable option?:

Asking the user to **interactively** confirm the default region (or confirm the region selection from any source) either by default or via a command-line flag (`--confirm-region`)?

Or, is pubsys supposed to function as if called non-interactively?

zmrow:
pubsys shouldn't need to be interactive.  The config file and/or existing arguments give the user enough configurability and is about as explicit as is reasonable. :)

mchaker:
This change ultimately was not merged in order to maintain a predictable user experience: having the default AWS region selected automatically on an AMI push action could be a surprise for a user and lead to unexpected consequences.

PR #2290, opened 2022-07-21
kernel: config: clean up '=n' settings

foersleo:
```
Disabling a kernel config setting is usually done in a config file in
the form of

\# CONFIG_<CONFIG_NAME> is not set

instead of

CONFIG_<CONFIG_NAME>=n

While both today should be equivalent historically there was problems
with specifying '=n' still making that setting defined and having the
opposite effect of what is intended by setting '=n'. Nowadays either way
will result in a final config that will only have the `is not set`
variant for all deactivated options, so the diff of the resulting
artifact before and after this patch will be empty.

As the `is not used` form is the commonly used form, lets adhere to it.
As a bonus it reduces the noise when comparing configs that are manually
merged using the merge-config script.

Replaced all the '=n' settings by the common form through command:

for file in packages/kernel*/config*;
do
    sed -E -i "s/^(CONFIG_[A-Z0-9_]*)=n/# \1 is not set/g" $file;
done

Signed-off-by: Leonard Foerster <foersleo@amazon.com>
```

**Issue number:** https://github.com/bottlerocket-os/bottlerocket-kernel-kit/issues/72

**Description of changes:**

Clean up an un-common way of disabling kernel config options. See commit message above for more detail.

**Testing done:**

Checked resulting config that will be installed as part of the kernel rpm before and after this patch. The diff is empty.

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.10/config-bottlerocket
+++ b/packages/kernel-5.10/config-bottlerocket
@@ -39,13 +39,13 @@ CONFIG_FRAMEBUFFER_CONSOLE_DEFERRED_TAKEOVER=y
 CONFIG_SECURITY_YAMA=y
 
 # Do not allow SELinux to be disabled at boot.
-CONFIG_SECURITY_SELINUX_BOOTPARAM=n
+# CONFIG_SECURITY_SELINUX_BOOTPARAM is not set
 
 # Do not allow SELinux to be disabled at runtime.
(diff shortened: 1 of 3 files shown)


zmrow (approved):
:jack_o_lantern:

markusboehme (approved):

jpculp (approved):

bcressey (approved):

PR #2298, opened 2022-07-25
modeled_types: Use FromStr instead of TryFrom

### LINKED ISSUE #1061: modeled_types: Use FromStr instead of TryFrom<&str>
[`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) and [`TryFrom<&str>`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) are functionally equivalent traits, but the former has existed since Rust 1.0.0 and has much wider support across the ecosystem; `serde_plain` is a key example of this.

mchaker:
**Issue number:**

Closes #1061 

**Description of changes:**

In order to clean up the code, this change replaces TryFrom<&str>
with FromStr. This change has been made for all `modeled_types`.

**Testing done:**

- [x] performed a complete build of `aws-dev` with no errors
- [x] performed a complete build of `metal-dev` with no errors
- [x] performed a complete build of `vmware-dev` with no errors

**Terms of contribution:**

### DIFF
--- a/sources/api/apiserver/src/server/controller.rs
+++ b/sources/api/apiserver/src/server/controller.rs
@@ -563,8 +563,9 @@ mod test {
     use datastore::memory::MemoryDataStore;
     use datastore::{Committed, DataStore, Key, KeyType};
     use maplit::{hashmap, hashset};
+    use model::modeled_types::SingleLineString;
     use model::{ConfigurationFile, Service};
-    use std::convert::TryInto;
+    use std::str::FromStr;
 
     #[test]
     fn get_settings_works() {
@@ -579,7 +580,10 @@ mod test {
 
         // Retrieve with helper
         let settings = get_settings(&ds, &Committed::Live).unwrap();
-        assert_eq!(settings.motd, Some("json string".try_into().unwrap()));
+        assert_eq!(
+            settings.motd,
+            Some(("json string").parse::<String>().unwrap())
+        );
     }
 
     #[test]
@@ -597,13 +601,19 @@ mod test {
         let settings = get_settings_prefix(&ds, "settings.", &Committed::Live)
             .unwrap() // Result Ok
             .unwrap(); // got Some result
-        assert_eq!(settings.motd, Some("json string".try_into().unwrap()));
+        assert_eq!(
+            settings.motd,
+            Some(("json string").parse::<String>().unwrap())
+        );
 
         // Retrieve with more specific prefix OK
         let settings = get_settings_prefix(&ds, "settings.mot", &Committed::Live)
             .unwrap() // Result Ok
             .unwrap(); // got Some result
-        assert_eq!(settings.motd, Some("json string".try_into().unwrap()));
+        assert_eq!(
+            settings.motd,
+            Some(("json string").parse::<String>().unwrap())
+        );
 
         // No match should return None; the "view" layer of the API, in mod.rs, turns this into an
         // empty object if desired.
@@ -636,7 +646,10 @@ mod test {
         // Retrieve with helper
         let settings =
             get_settings_keys(&ds, &hashset!("settings.motd"), &Committed::Live).unwrap();
-        assert_eq!(settings.motd, Some("json string 1".try_into().unwrap()));
+        assert_eq!(
+            settings.motd,
+            Some(("json string 1").parse::<String>().unwrap())
+        );
         assert_eq!(settings.ntp, None);
     }
 
@@ -663,7 +676,7 @@ mod test {
         assert_eq!(
             services,
             hashmap!("foo".to_string() => Service {
-                configuration_files: vec!["file1".try_into().unwrap()],
+                configuration_files: vec![("file1").parse::<SingleLineString>().unwrap()],
                 restart_commands: vec!["echo hi".to_string()]
             })
         );
@@ -694,7 +707,7 @@ mod test {
         assert_eq!(
             services,
             hashmap!("foo".to_string() => Service {
-                configuration_files: vec!["file1".try_into().unwrap()],
+                configuration_files: vec![("file1").parse::<SingleLineString>().unwrap()],
                 restart_commands: vec!["echo hi".to_string()]
             })
         );
@@ -735,8 +748,8 @@ mod test {
         assert_eq!(
             configuration_files,
             hashmap!("foo".to_string() => ConfigurationFile {
-                path: "file".try_into().unwrap(),
-                template_path: "template".try_into().unwrap(),
+                path: ("file").parse::<SingleLineString>().unwrap(),
+                template_path: ("template").parse::<SingleLineString>().unwrap(),
             })
         );
 
@@ -754,7 +767,7 @@ mod test {
     #[test]
     fn set_settings_works() {
         let mut settings = Settings::default();
-        settings.motd = Some("tz".try_into().unwrap());
+        settings.motd = Some(("tz").parse::<String>().unwrap());
 
         // Set with helper
         let mut ds = MemoryDataStore::new();
@@ -832,7 +845,10 @@ mod test {
 
         // Confirm pending
         let settings = get_settings(&ds, &pending).unwrap();
-        assert_eq!(settings.motd, Some("json string".try_into().unwrap()));
+        assert_eq!(
+            settings.motd,
+            Some(("json string").parse::<String>().unwrap())
+        );
         // No live settings yet
         get_settings(&ds, &Committed::Live).unwrap_err();
 
@@ -843,6 +859,9 @@ mod test {
         get_settings(&ds, &pending).unwrap_err();
         // Confirm live
         let settings = get_settings(&ds, &Committed::Live).unwrap();
-        assert_eq!(settings.motd, Some("json string".try_into().unwrap()));
+        assert_eq!(
+            settings.motd,
+            Some(("json string").parse::<String>().unwrap())
+        );
     }
 }
--- a/sources/api/bootstrap-containers/src/main.rs
+++ b/sources/api/bootstrap-containers/src/main.rs
@@ -80,7 +80,6 @@ use datastore::{serialize_scalar, Key, KeyType};
 use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::HashMap;
-use std::convert::TryFrom;
 use std::env;
 use std::ffi::OsStr;
 use std::fmt::Write;
@@ -243,7 +242,8 @@ fn parse_mark_bootstrap_args(args: Vec<String>) -> Result<Subcommand> {
     Ok(Subcommand::MarkBootstrap(MarkBootstrapArgs {
         container_id: container_id,
         // Fail if 'mode' is invalid
-        mode: BootstrapContainerMode::try_from(mode.to_string())
+        mode: mode
+            .parse::<BootstrapContainerMode>()
             .context(error::BootstrapContainerModeSnafu)?,
     }))
 }
--- a/sources/api/ecs-settings-applier/src/ecs.rs
+++ b/sources/api/ecs-settings-applier/src/ecs.rs
@@ -16,6 +16,7 @@ use snafu::{OptionExt, ResultExt};
 use std::convert::TryFrom;
 use std::fs;
 use std::path::Path;
+use std::str::FromStr;
 use std::{env, process};
 
 const DEFAULT_ECS_CONFIG_PATH: &str = "/etc/ecs/ecs.config.json";
@@ -112,7 +113,7 @@ async fn run() -> Result<()> {
         image_pull_behavior: ecs
             .image_pull_behavior
             .as_ref()
-            .map(|b| ECSImagePullBehavior::try_from(b.as_ref()).unwrap() as u8),
+            .map(|b| b.parse::<ECSImagePullBehavior>().unwrap() as u8),
 
         // Task role support is always enabled
         task_iam_role_enabled: true,
--- a/sources/api/prairiedog/src/bootconfig.rs
+++ b/sources/api/prairiedog/src/bootconfig.rs
@@ -5,8 +5,8 @@ use model::modeled_types::{BootConfigKey, BootConfigValue};
 use model::BootSettings;
 use snafu::{ensure, ResultExt};
 use std::collections::HashMap;
-use std::convert::TryInto;
 use std::path::Path;
+use std::str::FromStr;
 use tokio::io;
 
 // Boot config related consts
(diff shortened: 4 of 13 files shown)


mchaker:
_Force push: fix `try_into` calls that escaped local testing_

mchaker:
_Force push: fix more `try_into` calls that escaped local testing_

mchaker:
_Force push: clear up rustfmt errors_

zmrow:
I'm a little curious about this motivation behind this change!

Also, assuming this all came up via compiler errors, but one nice thing about implementing `TryFrom` means you can use the reciprocal `TryInto`.  Are you sure we're not calling `try_into()` in places that would require the `TryFrom` implementation?

mchaker:
> I'm a little curious about this motivation behind this change!

Thanks for your interest in this change! The two people who initially recommended this change are no longer on the team, so I will provide what I believe is the reason for this change. 

I believe the reasons for this change are discussed in the following posts: 

* [What is the difference between the FromStr and TryFrom traits?](https://stackoverflow.com/questions/67385956/what-is-the-difference-between-the-fromstr-and-tryfromstring-traits)
* [FromStr vs TryFrom(&str)](https://www.reddit.com/r/rust/comments/vzgvjv/fromstr_vs_tryfromstr/)

Essentially, it is more idiomatic to convert from strings using `FromStr` because doing so allows you to use the _default string parsing syntax_ (`string_variable.parse()`). In fact, I'm in the middle of changing all of the `from_str()` calls in this PR to `string_variable.parse()` calls. :)

> Also, assuming this all came up via compiler errors, but one nice thing about implementing `TryFrom` means you can use the reciprocal `TryInto`. Are you sure we're not calling `try_into()` in places that would require the `TryFrom` implementation?

I only replaced `TryFrom<&str>` calls that I was sure were conversions from strings, and left everything else untouched (e.g. conversions from custom types such as `FriendlyVersion`)

The `TryFrom` implementation that the macro creates for custom types in `modeled_types` is for the `String` type ([permalink](https://github.com/bottlerocket-os/bottlerocket/blob/cd483742d04e8d6a0d740d02215323d4aee5d724/sources/models/src/modeled_types/mod.rs#L157)). That `TryFrom<String>` implementation actually just used to call `TryFrom<&str>` under the hood, and each `modeled_types` custom type implemented the `TryFrom<&str>` trait.

Now, each `modeled_types` custom type implements `FromStr`, which allows us to call `.parse()` from within the macro instead of `TryFrom<&str>`'s `try_from()`.

mchaker:
_Force push: changed all `::from_str()` calls to `.parse::<Type>()` calls._

mchaker:
_Force push: typo_

arnaldo2792 (approved):

sources/api/apiserver/src/server/controller.rs
```diff
@@ -579,7 +580,10 @@ mod test {
 
         // Retrieve with helper
         let settings = get_settings(&ds, &Committed::Live).unwrap();
-        assert_eq!(settings.motd, Some("json string".try_into().unwrap()));
+        assert_eq!(
+            settings.motd,
+            Some(("json string").parse::<String>().unwrap())
+        );
```
cbgbt, line 586: Is the turbofish `::<String>` necessary here? I'm curious why the old code was able to deduce the type, but the `parse` invocation may be struggling to.

In general, we tend to let the type inference do its job where it can, and specify types otherwise.

sources/models/src/modeled_types/mod.rs
```diff
@@ -158,7 +158,7 @@ macro_rules! string_impls_for {
             type Error = $crate::modeled_types::error::Error;
 
             fn try_from(input: String) -> Result<Self, Self::Error> {
-                Self::try_from(input.as_ref())
+                input.parse::<Self>()
```
cbgbt, line 161: We should update the module-level docstring, which provides some information about `TryFrom<&str>` and how this macro behaves.

sources/api/apiserver/src/server/controller.rs
mchaker: There is one file where the types can/do not get automatically inferred:
`sources/models/src/de.rs`

attempting to use `.parse()` in e.g. `node_taints_list_representation()` results in:

```
type annotations needed
required because of the requirements on the impl of `std::cmp::PartialEq<std::vec::Vec<_>>` for `std::vec::Vec<modeled_types::kubernetes::KubernetesTaintValue>`rustc[E0283](https://doc.rust-lang.org/error-index.html#E0283)
mod.rs(40, 32): Actual error occurred here
mod.rs(225, 9): multiple `impl`s satisfying `modeled_types::kubernetes::KubernetesTaintValue: std::cmp::PartialEq<_>` found
de.rs(80, 43): consider specifying the type argument in the method call: `::<F>`
type annotations needed
cannot infer type for type parameter `T`rustc[E0282](https://doc.rust-lang.org/error-index.html#E0282)
macros.rs(50, 36): Actual error occurred here
de.rs(80, 43): consider specifying the type argument in the method call: `::<F>`
```

etungsten: What about the other locations? Like Sean said, I think we should let type inference do its job if possible.

mchaker: I'm fixing the other locations, no worries.

Would it be best to leave the ambiguous type inference I mentioned in my previous comment as a turbofish, and let all the other locations use `.parse()` without turbofishing?

etungsten: Yes, I'm ok with that.

sources/models/src/modeled_types/mod.rs
mchaker: How does this sound? (Just changed the `TryFrom<&str>` to `FromStr` -- the rest looks accurate?)

```rust
// The pattern in this module is to make a struct and implement FromStr with code that does
// necessary checks and returns the struct.  Other traits that treat the struct like a string can
// be implemented for you with the string_impls_for macro.
```

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -27,10 +28,10 @@ lazy_static! {
...
 }
 
-impl TryFrom<&str> for KubernetesName {
-    type Error = error::Error;
+impl FromStr for KubernetesName {
+    type Err = error::Error;
 
-    fn try_from(input: &str) -> Result<Self, Self::Error> {
+    fn from_str(input: &str) -> Result<Self, Self::Err> {
         ensure!(
```
rpkelly, line 35: I could be misunderstanding here, but, since we have declared `KubernetesName` as public, is it reasonable to remove `TryFrom`? Could we not just keep it, while adding `FromStr` as well? This same question is valid below as well with `KubernetesLabelKey`

sources/api/bootstrap-containers/src/main.rs
```diff
@@ -243,7 +242,8 @@ fn parse_mark_bootstrap_args(args: Vec<String>) -> Result<Subcommand> {
     Ok(Subcommand::MarkBootstrap(MarkBootstrapArgs {
         container_id: container_id,
         // Fail if 'mode' is invalid
-        mode: BootstrapContainerMode::try_from(mode.to_string())
+        mode: mode
+            .parse::<BootstrapContainerMode>()
             .context(error::BootstrapContainerModeSnafu)?,
```
rpkelly, line 247: This stands out as not being unit tested. Can we add a unit test to this to make sure that we aren't changing behavior?

sources/api/ecs-settings-applier/src/ecs.rs
```diff
@@ -112,7 +113,7 @@ async fn run() -> Result<()> {
         image_pull_behavior: ecs
             .image_pull_behavior
             .as_ref()
-            .map(|b| ECSImagePullBehavior::try_from(b.as_ref()).unwrap() as u8),
+            .map(|b| b.parse::<ECSImagePullBehavior>().unwrap() as u8),
```
rpkelly, line 116: This function looks difficult to unit test. Do you know if it's covered by an integration test of some sort?

sources/api/storewolf/src/main.rs
```diff
@@ -186,10 +185,12 @@ fn parse_metadata_toml(md_toml_val: toml::Value) -> Result<Vec<model::Metadata>>
...
-                let md =
-                    SingleLineString::try_from(md_key).context(error::SingleLineStringSnafu)?;
-                let key =
-                    SingleLineString::try_from(data_key).context(error::SingleLineStringSnafu)?;
+                let md = md_key
+                    .parse::<SingleLineString>()
+                    .context(error::SingleLineStringSnafu)?;
+                let key = data_key
+                    .parse::<SingleLineString>()
+                    .context(error::SingleLineStringSnafu)?;
```
rpkelly, line 193: It looks like we should be able to unit test this function to make sure this hasn't changed behavior

sources/models/src/modeled_types/shared.rs
```diff
@@ -26,10 +26,10 @@ pub struct ValidBase64 {
 }
```
rpkelly, line 26: This has the same question as in `kubernetes.rs` should we leave `TryFrom` and add `FromStr` instead of completely replacing it?

sources/models/src/modeled_types/kubernetes.rs
zmrow: It feels a bit odd and misleading to have both.  Is there any reason to remove `TryFrom` here other than consistency?

arnaldo2792: I would be more concerned if this was a public library used in multiple places since this change will actually cause pain to the users and require refactoring. But, since we are the only consumers of this code, and the painful refactor was already done, it should be fine to only keep `FromStr`.

Downstream forks of Bottlerocket may have to do some refactor as well if they rely on what `TryFrom` provides, but that is also true for whatever new feature we introduce in Bottlerocket, specially in the API model.

sources/api/ecs-settings-applier/src/ecs.rs
mchaker: I think the test I will execute is to actually run Bottlerocket and use a feature that utilizes `ECSImagePullBehavior`.

sources/api/bootstrap-containers/src/main.rs
mchaker: If you mean for `.parse::<BootstrapContainerMode>()`, I think the test is located here: https://github.com/bottlerocket-os/bottlerocket/blob/cf7a49a705883bf5377f3a6de746aa8642a0a2ca/sources/models/src/modeled_types/shared.rs#L966-L981

sources/api/storewolf/src/main.rs
mchaker: If you mean `.parse::<SingleLineString>()`, I think the unit test is located here: https://github.com/bottlerocket-os/bottlerocket/blob/cf7a49a705883bf5377f3a6de746aa8642a0a2ca/sources/models/src/modeled_types/shared.rs#L101-L129

sources/models/src/modeled_types/shared.rs
mchaker: I'll leave both `TryFrom` and `FromStr` for the following mentioned `pub struct`s:

* `sources/models/src/modeled_types/kubernetes.rs`:
    * `KubernetesName`
    * `KubernetesLabelKey`
* `sources/models/src/modeled_types/shared.rs`
    * `ValidBase64`

Is that correct?

stmcginnis: If possible, it would be good to have some comments on why/when to use one or the other. That way when someone (me) works on this months from now there isn't any confusion about the correct one to use.

sources/api/ecs-settings-applier/src/ecs.rs
mchaker: Upon further investigation, `ECSImagePullBehavior` is not directly used -- `ECSAgentImagePullBehavior` is what is used. I say that because that is what is used in `ECSSettings`: 

https://github.com/bottlerocket-os/bottlerocket/blob/fa341b494872a4228a43de9c0ee95bc80f8553ba/sources/models/src/lib.rs#L238

That `ECSAgentImagePullBehavior` code is tested here:

https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/src/modeled_types/ecs.rs#L275-L293

sources/models/src/modeled_types/shared.rs
mchaker: After further consideration, I agree with @arnaldo2792 in that only `FromStr` should be available:
https://github.com/bottlerocket-os/bottlerocket/pull/2298#discussion_r967393727

etungsten (approved):

sources/api/bootstrap-containers/src/main.rs
rpkelly: I mean `fn parse_mark_bootstrap_args`

sources/api/storewolf/src/main.rs
rpkelly: I mean `fn parse_metadata_toml`. The tests above make sure that `parse` works, but, is it exactly equivalent to the previously used `try_from` call?

sources/api/bootstrap-containers/src/main.rs
rpkelly: We agreed to open an issue for then entire main.rs file since it's missing tests completely.

sources/api/storewolf/src/main.rs
rpkelly: We agreed to open an issue for then entire main.rs file since it's missing tests completely.

rpkelly (approved):

sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -5,8 +5,8 @@ use model::modeled_types::{BootConfigKey, BootConfigValue};
 use model::BootSettings;
 use snafu::{ensure, ResultExt};
 use std::collections::HashMap;
-use std::convert::TryInto;
 use std::path::Path;
+use std::str::FromStr;
```
bcressey, line 9: This gets flagged as an unused import.

sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -142,7 +142,8 @@ fn parse_value(input: &str) -> Result<BootConfigValue> {
     } else {
         input
     };
-    s.try_into().context(error::ParseBootConfigValueSnafu)
+    s.parse::<BootConfigValue>()
+        .context(error::ParseBootConfigValueSnafu)
```
bcressey, line 146: Requiring explicit types in places where they weren't required before doesn't really show this change in the best light.

It'd be better to use type inference where possible, e.g.
```
    s.parse::<>()
        .context(error::ParseBootConfigValueSnafu)
```

bcressey (commented):
For testing, I would like to see test results that exercise the serialization and deserialization paths for all of the types where `TryFrom<&str>` became `FromStr`.

PR #2300, opened 2022-07-26
Update sources to use AWS SDK Rust

rpkelly:
**Issue number:**
[1968](https://github.com/bottlerocket-os/bottlerocket/issues/1968)

**Description of changes:**
Updates pluto and cfsignal in sources to use AWS SDK Rust.

**Testing done:**
Unit testing and built image locally.

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/eks.rs
+++ b/sources/api/pluto/src/eks.rs
@@ -1,14 +1,11 @@
+use aws_sdk_eks::model::KubernetesNetworkConfigResponse;
+use aws_types::region::Region;
 use hyper::http::uri::InvalidUri;
 use hyper::Uri;
 use hyper_proxy::{Proxy, ProxyConnector};
 use hyper_rustls::HttpsConnectorBuilder;
-use rusoto_core::credential::ChainProvider;
-use rusoto_core::region::ParseRegionError;
-use rusoto_core::{Region, RusotoError};
-use rusoto_eks::{DescribeClusterError, Eks, EksClient, KubernetesNetworkConfigResponse};
 use snafu::{OptionExt, ResultExt, Snafu};
 use std::env;
-use std::str::FromStr;
 
 pub(crate) type ClusterNetworkConfig = KubernetesNetworkConfigResponse;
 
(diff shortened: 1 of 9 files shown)


etungsten:
I've verified that the sundog/pluto proxy testing from https://github.com/bottlerocket-os/bottlerocket/pull/2227 still works as expected.

sources/cfsignal/src/cloudformation.rs
```diff
@@ -1,10 +1,10 @@
+use std::str::FromStr;
+
```
etungsten, line 2: nit: extra line

sources/api/pluto/src/eks.rs
```diff
@@ -1,33 +1,24 @@
...
-use std::str::FromStr;
 
 pub(crate) type ClusterNetworkConfig = KubernetesNetworkConfigResponse;
 
 #[derive(Debug, Snafu)]
 pub(super) enum Error {
     #[snafu(display("Error describing cluster: {}", source))]
     DescribeCluster {
-        source: RusotoError<DescribeClusterError>,
+        source: aws_sdk_eks::types::SdkError<aws_sdk_eks::error::DescribeClusterError>,
```
etungsten, line 16: nit: I recommend some use declarations to make this path shorter. I think it'll help with readability a bit. This comment applies to other long import paths elsewhere as well.

etungsten (approved):

jpculp (approved):

sources/api/pluto/src/eks.rs
etungsten: on second thought, I think being explicit here is probably better

zmrow (approved):
:fondue:

sources/api/pluto/src/eks.rs
```diff
@@ -57,6 +46,11 @@ pub(super) async fn get_cluster_network_config(
         .find(|env_var| *env_var != Err(env::VarError::NotPresent))
         .and_then(|s| s.ok());
 
+    let config = aws_config::from_env()
+        .region(Region::new(region.to_owned()))
```
etungsten, line 50: I'm a little weirded out by the fact that `aws_sdk_rust` doesn't attempt to do validation on the input string. We also don't do any validation on `settings.aws.regions`. I'm assuming this just takes whatever is given here and generates API endpoints. Just something to keep in mind.

bcressey (approved):

sources/api/pluto/src/eks.rs
rpkelly: We could probably make some guesses surrounding what a region name looked like, but, I feel like we open ourselves to being broken every time a new region is released by doing so.

PR #2308, opened 2022-07-27
gitignore: add VScode workspaces

mchaker:
**Issue number:**

N/A

**Description of changes:**

VScode user workspaces should not be tracked.

I work from multiple client machines and want my workspace to persist/sync between them.

**Testing done:**

visual confirmation

**Terms of contribution:**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 /Licenses.toml
 /licenses
 *.run
+*.code-workspace


mchaker:
Update: alternatively, I have decided to store personal VScode workspace files outside of the repository. This removes the need for adding a new item to the `.gitignore`. 

This also opens up the possibility for including specific workspaces as demo/quickstart/onboarding workspaces (to use with GitHub Codespaces 😄 )

stmcginnis:
Alternatively, it might be worth considering adding something like this so PRs like this don't keep popping up from time to time: https://github.com/vmware-tanzu/community-edition/pull/2896/files

PR #2311, opened 2022-07-28
netdog: Move to `quick-xml` for XML serialization

zmrow:
**Description of changes:**

```
`serde_xml_rs` has a few bugs that we were able to work around, but the
inability to serialize a `Vec` is something we can't work around.  This
change moves the netdog code to `quick-xml` for serialization.
```

This change also allows us to get rid of the `_f: ()` struct member that worked around a `serde_xml_rs` bug.  However, `quick-xml` is slightly more wordy in that you need to supply ` $unflatten` attributes for anything that is a simple type and needs to have it's own tag; otherwise it gets made an attribute of the tag itself.

**Testing done:**
* All unit tests still pass
* Built and ran an aws-k8s-1.21 node successfully.

**Terms of contribution:**

### DIFF
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -2493,10 +2493,10 @@ dependencies = [
  "indexmap",
  "ipnet",
  "lazy_static",
+ "quick-xml",
  "rand",
  "regex",
  "serde",
- "serde-xml-rs",
  "serde_json",
  "serde_plain",
  "snafu",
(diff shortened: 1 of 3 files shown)


zmrow:
^ Rebases on `develop` and resolves merge conflicts

etungsten (approved):
:hamburger:

stmcginnis (approved):
Tests are happy and the code changes look good to me. Not super familiar with the code and any side effects changes might have though, so will leave to someone else to merge.

PR #2322, opened 2022-08-05
Autoscaling Warm Pool Support for K8S Variants

parth3sh:
**Issue number:**
#1788 

**Description of changes:**
**Goal: To support EC2 Autoscaling “warm pools” in K8S Bottlerocket variants.**

- Add autoscaling.should-wait setting to K8S variants. Handle migrations for setting in case of downgrade.
- Create "hachiko", a helper program that retrieves the autoscaling.should-wait setting and autoscaling lifecycle state of host. If this setting is true, program waits until lifecycle state is "InService".
- autoscaling.should-wait setting is written to a config file (according to hachiko-toml template) during instance boot and is later parsed by the helper program.
- Create hachiko.service to execute helper program prior to the running of Kubelet in order to prevent warm pool instances from running Kubelet and falling into the "NotReady" state.

**Testing done:**

- Program is installed properly.
- Service runs as expected and at the right point during host configuration.
- Waiting program tested locally and on a launched bottlerocket instance for ability to correctly parse config file and retrieve lifecycle state of the host.
- Config file with autoscaling.should-wait setting value is correctly generated and parsed. 

**Testing required:**
- Confirm autoscaling.should-wait is migrated properly by confirming it is added to k8s variants upon upgrade to version 10.0.0 and removed upon downgrade.

**Terms of contribution:**

### DIFF
--- a/packages/os/hachiko.service
+++ b/packages/os/hachiko.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Autoscaling warm pools support
+After=network-online.target settings-applier.service
+Before=kubelet.service
+Requires=settings-applier.service network-online.target
+# We only want to run once, at first boot. This file is created by hachiko
+# after a successful run.
+ConditionPathExists=!/var/lib/bottlerocket/hachiko.ran
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/hachiko
+RemainAfterExit=true
+
+[Install]
+WantedBy=multi-user.target
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -25,6 +25,7 @@ Source6: metricdog-toml
 Source7: host-ctr-toml
 Source8: oci-default-hooks-json
 Source9: cfsignal-toml
+Source10: hachiko-toml
 
 # 1xx sources: systemd units
 Source100: apiserver.service
@@ -44,6 +45,7 @@ Source116: load-kernel-modules.service.in
 Source117: cfsignal.service
 Source118: generate-network-config.service
 Source119: prepare-primary-interface.service
+Source120: hachiko.service
 
 # 2xx sources: tmpfilesd configs
 Source200: migration-tmpfiles.conf
@@ -91,6 +93,7 @@ Requires: %{_cross_os}static-pods
 %if %{with aws_platform}
 Requires: %{_cross_os}shibaken
 Requires: %{_cross_os}cfsignal
+Requires: %{_cross_os}hachiko
 %endif
 
 %if %{with ecs_runtime}
@@ -247,6 +250,11 @@ Summary: Setting generator for populating admin container user-data from IMDS.
 Summary: Bottlerocket CloudFormation Stack signaler
 %description -n %{_cross_os}cfsignal
 %{summary}.
+
+%package -n %{_cross_os}hachiko
+Summary: Autoscaling warmpool support
+%description -n %{_cross_os}hachiko
+%{summary}.
 %endif
 
 %package -n %{_cross_os}shimpei
@@ -328,7 +336,7 @@ echo "** Output from non-static builds:"
     -p certdog \
     -p shimpei \
     %{?with_ecs_runtime: -p ecs-settings-applier} \
-    %{?with_aws_platform: -p shibaken -p cfsignal} \
+    %{?with_aws_platform: -p shibaken -p cfsignal -p hachiko} \
     %{?with_aws_k8s_family: -p pluto} \
     %{?with_k8s_runtime: -p static-pods} \
     %{?with_nvidia_flavor: -p driverdog} \
@@ -354,7 +362,7 @@ for p in \
   ghostdog bootstrap-containers \
   shimpei \
   %{?with_ecs_runtime: ecs-settings-applier} \
-  %{?with_aws_platform: shibaken cfsignal} \
+  %{?with_aws_platform: shibaken cfsignal hachiko} \
   %{?with_aws_k8s_family: pluto} \
   %{?with_k8s_runtime: static-pods} \
   %{?with_nvidia_flavor: driverdog} \
@@ -416,7 +424,9 @@ install -p -m 0644 \
 
 %if %{with aws_platform}
 install -p -m 0644 %{S:9} %{buildroot}%{_cross_templatedir}
+install -p -m 0644 %{S:10} %{buildroot}%{_cross_templatedir}
 install -p -m 0644 %{S:117} %{buildroot}%{_cross_unitdir}
+install -p -m 0644 %{S:120} %{buildroot}%{_cross_unitdir}
 %endif
 
 install -d %{buildroot}%{_cross_tmpfilesdir}
@@ -536,6 +546,12 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 %dir %{_cross_templatedir}
 %{_cross_templatedir}/cfsignal-toml
 %{_cross_unitdir}/cfsignal.service
+
+%files -n %{_cross_os}hachiko
+%{_cross_bindir}/hachiko
+%dir %{_cross_templatedir}
+%{_cross_templatedir}/hachiko-toml
+%{_cross_unitdir}/hachiko.service
 %endif
 
 %if %{with nvidia_flavor}
--- a/packages/selinux-policy/fs.cil
+++ b/packages/selinux-policy/fs.cil
@@ -47,6 +47,7 @@
 (filecon "/.*/usr/bin/storewolf" file api_exec)
 (filecon "/.*/usr/bin/cfsignal" file api_exec)
 (filecon "/.*/usr/bin/thar-be-settings" file api_exec)
+(filecon "/.*/usr/bin/hachiko" file api_exec)
 (filecon "/.*/usr/bin/dbus-broker.*" file bus_exec)
 (filecon "/.*/usr/sbin/chronyd" file clock_exec)
 (filecon "/.*/usr/sbin/wicked.*" file network_exec)
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -27,6 +27,7 @@ members = [
     # "api/migration/migrations/vX.Y.Z/..."
     "api/migration/migrations/v1.10.0/dns-settings",
     "api/migration/migrations/v1.10.0/dns-settings-metadata",
+    "api/migration/migrations/v1.10.0/add-eks-autoscaling",
 
     "bottlerocket-release",
 
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -61,7 +61,9 @@ members = [
 
     "constants",
 
-    "shimpei"
+    "shimpei",
+
+    "hachiko"
 ]
 
 [profile.release]
--- a/sources/api/migration/migrations/v1.10.0/add-eks-autoscaling/src/main.rs
+++ b/sources/api/migration/migrations/v1.10.0/add-eks-autoscaling/src/main.rs
@@ -1,0 +1,24 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting prefix for configuring autoscaling in k8s variants.
+/// Remove the whole `settings.autoscaling` prefix if we downgrade.
+fn run() -> Result<()> {
+    if cfg!(variant_runtime = "k8s") {
+        migrate(AddPrefixesMigration(vec!["settings.autoscaling"]));
+    }
+    Ok(())
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/hachiko/src/config.rs
+++ b/sources/hachiko/src/config.rs
@@ -1,0 +1,58 @@
+use crate::error::{self, Result};
+use serde::Deserialize;
+use snafu::ResultExt;
+use std::fs;
+use std::path::Path;
+
+#[derive(Debug, Deserialize)]
+pub(crate) struct Config {
+    pub(crate) should_wait: bool,
+    pub(crate) marker_path: String,
+}
+
+impl Config {
+    /// Parses configuration file at passed in path and returns struct containing settings value.
+    pub(crate) fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let path = path.as_ref();
+        let string_parse = fs::read_to_string(path).context(error::ConfigReadSnafu { path })?;
+        let config: Config =
+            toml::from_str(&string_parse).context(error::ConfigParseSnafu { path })?;
+        Ok(config)
+    }
+}
+
+#[cfg(test)]
+mod config_test {
+    use crate::config::Config;
+    use tempfile::TempDir;
+
+    // Example config file where user sets autoscaling.should-wait to false.
+    const FALSE_SHOULD_WAIT: &str = r#"
+    should_wait = false
+    marker_path = "/var/lib/bottlerocket/hachiko.ran"
+    "#;
+
+    // Example config file where user sets autoscaling.should-wait to true.
+    const TRUE_SHOULD_WAIT: &str = r#"
+    should_wait = true
+    marker_path = "/var/lib/bottlerocket/hachiko.ran"
+    "#;
+
+    #[test]
+    fn false_should_wait_test() {
+        let dir = TempDir::new().unwrap();
+        let path = dir.path().join("config.toml");
+        std::fs::write(&path, FALSE_SHOULD_WAIT).unwrap();
+        let config = Config::from_file(&path).unwrap();
+        assert!(!config.should_wait);
+    }
+
+    #[test]
+    fn true_should_wait_test() {
+        let dir = TempDir::new().unwrap();
+        let path = dir.path().join("config.toml");
+        std::fs::write(&path, TRUE_SHOULD_WAIT).unwrap();
+        let config = Config::from_file(&path).unwrap();
+        assert!(config.should_wait);
+    }
+}
--- a/sources/hachiko/src/main.rs
+++ b/sources/hachiko/src/main.rs
@@ -1,0 +1,78 @@
+mod config;
+mod error;
+use crate::config::Config;
+use crate::error::Result;
+use imdsclient::ImdsClient;
+use log::LevelFilter;
+use log::{info, warn};
+use simplelog::{Config as LogConfig, SimpleLogger};
+use snafu::{OptionExt, ResultExt};
+use std::fs;
+use std::process;
+use tokio::time::{sleep, Duration};
+
+// Path to config file containing host's autoscaling.should-wait setting
+const CONFIG_PATH: &str = "/etc/hachiko.toml";
+
+/// Uses an imdsclient function to fetch the lifecyclestate of the host
+async fn get_lifecycle_state(client: &mut ImdsClient) -> Result<String> {
+    client
+        .fetch_lifecycle_state()
+        .await
+        .context(error::ImdsRequestSnafu)?
+        .context(error::ImdsNoneSnafu {
+            what: "instance-id",
+        })
+}
+
+/// Continuously fetches lifecycle state of the host.
+/// Returns from function when host is "InService"
+async fn wait_until_inservice() -> Result<()> {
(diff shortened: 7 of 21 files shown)


sources/Cargo.toml
```diff
@@ -68,6 +68,7 @@ members = [
     "api/migration/migrations/v1.9.0/kernel-modules-setting",
     "api/migration/migrations/v1.9.0/kernel-modules-setting-metadata",
     "api/migration/migrations/v1.9.0/kubelet-no-daemon-reload",
+    "api/migration/migrations/v1.10.0/add-eks-autoscaling",
     "api/migration/migrations/v1.9.0/updates-targets-base-url",
```
cbgbt: This migration should be ordered below the ones for `1.9.0`.

It also seems like this may be included in the wrong commit.

sources/ascwp/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[dependencies]
+log = "0.4"
+serde = { version = "1.0", features = ["derive"] }
+simplelog = "0.12"
+snafu = { version = "0.7" }
+toml = "0.5.1"
+tokio = { version = "~1.14", default-features = false, features = ["macros", "rt-multi-thread"] }
+rusoto_core = { version = "0.48.0", default-features = false, features = ["rustls"] }
+rusoto_cloudformation = { version = "0.48.0", default-features = false, features = ["rustls"] }
```
cbgbt: These dependencies can likely be removed.

sources/ascwp/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+imdsclient = { path = "../imdsclient", version = "0.1.0" }
+hyper = "0.14.2"
+async-std = "1.12.0"
+
+[dev-dependencies]
+httptest = "0.15"
```
cbgbt: Seems like this dependency can also be removed.

sources/ascwp/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+serde = { version = "1.0", features = ["derive"] }
+simplelog = "0.12"
+snafu = { version = "0.7" }
+toml = "0.5.1"
+tokio = { version = "~1.14", default-features = false, features = ["macros", "rt-multi-thread"] }
+rusoto_core = { version = "0.48.0", default-features = false, features = ["rustls"] }
+rusoto_cloudformation = { version = "0.48.0", default-features = false, features = ["rustls"] }
+imdsclient = { path = "../imdsclient", version = "0.1.0" }
+hyper = "0.14.2"
+async-std = "1.12.0"
```
cbgbt: These also seem to not be used.

sources/ascwp/src/config.rs
```diff
@@ -0,0 +1,60 @@
...
+const DEFAULT_CONFIG_PATH: &str = "/testConfig.toml";
+
+#[derive(Debug, Deserialize)]
+pub(crate) struct Config {
+    pub(crate) should_wait: bool,
+}
+
+impl Config {
+    //uses default config path
+    pub(crate) fn new() -> Result<Self> {
```
cbgbt: For functions, can you use docstring comments? They begin with `/// `. These comments generally use capitalized sentences with punctuation. (Same should be true for other functions in the PR)

sources/ascwp/src/config.rs
```diff
@@ -0,0 +1,60 @@
...
+        Self::from_file(PathBuf::from(DEFAULT_CONFIG_PATH))
+    }
+
+    //uses path passed into function
+    pub(crate) fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let path = path.as_ref();
+        let s = fs::read_to_string(path).context(error::ConfigReadSnafu { path })?;
```
cbgbt: I would avoid single character variable names.

sources/Cargo.toml
parth3sh: I have moved that migration below the 1.9.0 migrations and shifted the change to ["Handle migrations for eks-autoscaling.should-wait"](https://github.com/bottlerocket-os/bottlerocket/pull/2322/commits/c173ed58cc568c2664e44db6fd36520b347aa5be)

sources/ascwp/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+use log::{error, info, warn};
+use std::time::Duration;
+use async_std::task;
+
+
+
+//Path to config file containing host's eks_autoscaling.should-wait setting
+const CONFIG_PATH: &str = "/etc/ascwp.toml";
+//Marker file that is created after first run. We only want this program to run once.
+const MARKER_FILE: &str = "/var/lib/bottlerocket/ascwp.ran";
```
cbgbt: I think it would be good to make this configurable via `ascwp.toml`

sources/ascwp/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+use imdsclient::ImdsClient;
+use snafu::{OptionExt, ResultExt};
+use std::fs;
+use log::{error, info, warn};
+use std::time::Duration;
+use async_std::task;
+
+
+
+//Path to config file containing host's eks_autoscaling.should-wait setting
```
cbgbt: For non docstring comments as well, I'd add a space between `//` and the start of the comment.

sources/ascwp/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+    client
+        .fetch_lifecycle_state()
+        .await
+        .context(error::ImdsRequestSnafu)?
+        .context(error::ImdsNoneSnafu {
+            what: "instance-id",
+        })
+}
+
+//Continously fetches lifecycle state of the host. Returns when host is "InService"
```
cbgbt: nit: `Continuously`

sources/ascwp/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+async fn wait_until_inservice() -> Result<()> {
+    let mut client = ImdsClient::new();
+    let mut lifecycle_state = get_lifecycle_state(&mut client).await?;
+    while lifecycle_state.ne("InService") {
+        lifecycle_state = get_lifecycle_state(&mut client).await?;
+        task::sleep(Duration::from_secs(5)).await;
```
cbgbt: I believe we typically use `tokio`'s [sleep implementation](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html). You may be able to remove an additional dependency that way.

Separately: The poll interval should be hoisted to a `CONSTANT`, or perhaps even moved to the config file (with a default value).

sources/ascwp/src/config.rs
```diff
@@ -0,0 +1,60 @@
+use crate::error::{self, Result};
+use serde::Deserialize;
+use snafu::ResultExt;
+use std::fs;
+use std::path::{Path, PathBuf};
+const DEFAULT_CONFIG_PATH: &str = "/testConfig.toml";
```
cbgbt: It looks like we set a separate default below. Should we remove this one? Or centralize the default to this file?

sources/imdsclient/src/lib.rs
```diff
@@ -1,19 +1,14 @@
 /*!
```
cbgbt: It looks like some whitespace may have been accidentally removed in this docstring comment.

sources/models/src/aws-k8s-1.22/mod.rs
```diff
@@ -27,4 +27,5 @@ struct Settings {
     container_registry: RegistrySettings,
     oci_hooks: OciHooks,
     cloudformation: CloudFormationSettings,
+    eks_autoscaling: AutoScalingSettings,
```
cbgbt: I would just call this setting `autoscaling` so that it is consistent with the ECS setting.

packages/os/os.spec
```diff
@@ -423,13 +434,13 @@ install -d %{buildroot}%{_cross_datadir}/updog
 install -p -m 0644 %{_cross_repo_root_json} %{buildroot}%{_cross_datadir}/updog
 
 install -d %{buildroot}%{_cross_templatedir}
-install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{buildroot}%{_cross_templatedir}
+install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{S:9} %{S:10} %{buildroot}%{_cross_templatedir}
```
cbgbt: It looks like we're adding a reference here to sources 9 and 10, but only 10 is new in this PR. Is that right?

sources/api/migration/migrations/v1.10.0/add-eks-autoscaling/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting prefix for configuring autoscaling in k8s variants.
+/// Remove the whole `settings.eks_autoscaling` prefix if we downgrade.
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec!["settings.autoscaling"]))
+}
```
cbgbt: We'll want this migration to only run on EKS variants. You *should* be able to do something similar to [this](https://github.com/bottlerocket-os/bottlerocket/blob/2a6c5e92352134bb2c1e17c5317d1446b97ad7f5/sources/api/migration/migrations/v1.8.0/pki-affected-services/src/main.rs#L18). You'll want to test the migration on k8s and ECS variants, though.

sources/api/migration/migrations/v1.10.0/add-eks-autoscaling/src/main.rs
```diff
@@ -0,0 +1,21 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting prefix for configuring autoscaling in k8s variants.
+/// Remove the whole `settings.eks_autoscaling` prefix if we downgrade.
```
cbgbt: nit:

```suggestion
/// Remove the whole `settings.autoscaling` prefix if we downgrade.
```

sources/hachiko/src/config.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    #[test]
+    fn true_should_wait_test() {
+        let dir = TempDir::new().unwrap();
+        let path = dir.path().join("config.toml");
+        std::fs::write(&path, true_should_wait).unwrap();
+        let config = Config::from_file(&path).unwrap();
+        assert!(config.should_wait);
+    }
+}
```
cbgbt: The whitespacing on this file is unusual in a few places, and it's missing a [newline at the end of the file](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

I think running `cargo fmt` in the `hachiko` directory should fix this.

sources/hachiko/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+use crate::error::Result;
+use imdsclient::ImdsClient;
+use snafu::{OptionExt, ResultExt};
+use std::fs;
+use log::{error, trace, info, warn};
+use log::LevelFilter;
+use simplelog::{Config as LogConfig, SimpleLogger};
+use tokio::time::{sleep, Duration};
+
+/// Path to config file containing host's autoscaling.should-wait setting
```
cbgbt: I'm actually not sure if docstrings work for constants. You can check by running `cargo doc` and peeking at the output. I tend to just use regular comments.

```suggestion
// Path to config file containing host's autoscaling.should-wait setting
```

sources/hachiko/src/main.rs
```diff
@@ -56,11 +56,22 @@ async fn get_config() -> Result<Config> {
 async fn main() {
     SimpleLogger::init(LevelFilter::Info, LogConfig::default())
     .expect("unable to configure logger");
-    let should_wait_value = get_config().await.unwrap().should_wait;
+    let config_parse = get_config().await.unwrap();
```
gthao313, line 59: `unwrap` leaves us with a less meaningful output than explicit handling, so it would be better to use `context`  which is intended to be a human-readable, user-facing explanation for the error that has occurred.

gthao313: `unwrap` leaves us with a less meaningful output than explicit handling, so it would be better to use `context`  which is intended to be a human-readable, user-facing explanation for the error that has occurred.

sources/hachiko/src/main.rs
```diff
@@ -56,11 +56,22 @@ async fn get_config() -> Result<Config> {
...
+    let should_wait_value = config_parse.should_wait;
+    let marker_file_path = config_parse.marker_path;
     info!(
         "autoscaling.should-wait value is {}", should_wait_value
     );
     if should_wait_value{
         wait_until_inservice().await.unwrap();
```
gthao313: Same as above `unwrap`

packages/os/hachiko.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description= Waits for host to be in-service before running kubelet if autoscaling.should_wait is true.
+After=network-online.target settings-applier.service
+Before=kubelet.service
+Requires=settings-applier.service network-online.target
+# We only want to run once, at first boot. This file is created by hachiko
+# after a successful run.
+ConditionPathExists=!/var/lib/bottlerocket/hachiko.ran
+
+[Service]
```
zmrow, line 10: I think we probably want `RemainAfterExit=true` for this oneshot service so systemd doesn't show it as "deactivating" or "dead" after it runs.

See the service unit for `early-boot-config` for example.

packages/os/os.spec
```diff
@@ -546,6 +558,7 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 %{_cross_unitdir}/metricdog.timer
 %{_cross_unitdir}/send-boot-success.service
 
+
```
zmrow: Is this a spurious extra space?

zmrow (commented):
Just a few nits.

It sounds like you're already testing the migration; once that's done please add it to the PR description!

sources/hachiko/src/main.rs
```diff
@@ -0,0 +1,71 @@
...
+async fn get_config() -> Result<Config> {
+    Config::from_file(CONFIG_PATH)
+}
+
+/// If autoscaling.should-wait is true, calls waiting function.
+/// Writes marker file to prevent service from unexpectedly re-running.
+#[tokio::main]
+async fn main() {
+    SimpleLogger::init(LevelFilter::Info, LogConfig::default())
+    .expect("unable to configure logger");
```
zmrow: Let's use the `.context(snafu_error_here)` to provide context rather than using `expect()`

sources/hachiko/src/config.rs
```diff
@@ -0,0 +1,58 @@
...
+    use tempfile::TempDir;
+
+    // Example config file where user sets autoscaling.should-wait to false.
+    const false_should_wait: &str = r#"
+    should_wait = false
+    marker_path = "/var/lib/bottlerocket/hachiko.ran"
+    "#;
+
+    // Example config file where user sets autoscaling.should-wait to true.
+    const true_should_wait: &str = r#"
```
gthao313: `Const` variable needs upper case
```suggestion
    const TRUE_SHOULD_WAIT: &str = r#"
```

sources/hachiko/src/config.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+#[cfg(test)]
+mod configTest {
+    use crate::config::Config;
+    use tempfile::TempDir;
+
+    // Example config file where user sets autoscaling.should-wait to false.
+    const false_should_wait: &str = r#"
```
arnaldo2792: Same as below, the constant name should be all `UPPER_CASE`

packages/os/hachiko.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description= Waits for host to be in-service before running kubelet if autoscaling.should_wait is true.
```
arnaldo2792: Descriptions in systemd units are "especial". This is what systemd will print to the journal when it attempts to start a unit, so a simpler, shorter, message is preferred. Look at other systemd units for inspiration.

packages/os/hachiko.service
```diff
@@ -0,0 +1,16 @@
...
+# after a successful run.
+ConditionPathExists=!/var/lib/bottlerocket/hachiko.ran
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/hachiko
+RemainAfterExit=true
+
+[Install]
+WantedBy=kubelet.service
```
arnaldo2792: I looked at all the other services in Bottlerocket and none of them use `*.service` here. Thus, I think a target should be used, and for this case `multi-user.target` should be adequate.

packages/os/os.spec
```diff
@@ -232,6 +236,11 @@ Summary: Settings generator for ECS
 Summary: Dynamic setting generator for kubernetes
 %description -n %{_cross_os}pluto
 %{summary}.
+
+%package -n %{_cross_os}hachiko
+Summary: Autoscaling warmpool support
+%description -n %{_cross_os}hachiko
+%{summary}.
 %endif
```
arnaldo2792: The conditionals to add packages depending on the variant being built, change, you need to rebase `develop` to bring those changes.

packages/os/os.spec
```diff
@@ -423,13 +434,13 @@ install -d %{buildroot}%{_cross_datadir}/updog
...
 install -d %{buildroot}%{_cross_templatedir}
-install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{buildroot}%{_cross_templatedir}
+install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{S:10} %{buildroot}%{_cross_templatedir}
 
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 \
   %{S:100} %{S:101} %{S:102} %{S:103} %{S:105} \
   %{S:106} %{S:107} %{S:110} %{S:111} %{S:112} \
-  %{S:113} %{S:114} %{S:118} %{S:119} \
+  %{S:113} %{S:114} %{S:118} %{S:119} %{S:120} \
```
arnaldo2792: The `hachiko.service` should only be installed for `aws` variants, this is why the build is failing for VMware and Metal variants, since this service in packages conditionally a few lines below.

packages/selinux-policy/fs.cil
```diff
@@ -47,6 +47,7 @@
 (filecon "/.*/usr/bin/storewolf" file api_exec)
 (filecon "/.*/usr/bin/cfsignal" file api_exec)
 (filecon "/.*/usr/bin/thar-be-settings" file api_exec)
+(filecon "/.*/usr/bin/hachiko" file api_exec)
```
arnaldo2792, line 50: @bcressey there are a few programs now that are `api_exec` just to let them write to the persistent partition. Do you think it will be worth having a less privileged label that grants this permission for programs like `cfsignal` or `hachiko`?

sources/hachiko/src/main.rs
```diff
@@ -0,0 +1,58 @@
+mod config;
```
arnaldo2792, line 1: This program is modified in two commits, I think it will be best to squash those two commits since in commit 6d3efecc57f83c2b9a72e63bdb8c20bf4f038ee3, the program uses `.expect` to handle errors, whereas in commit a0764b084f31601f7e05eb66f88a076b54d68a1e Snafu contexts are used.

sources/hachiko/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+/// Continuously fetches lifecycle state of the host.
+/// Returns from function when host is "InService"
+async fn wait_until_inservice() -> Result<()> {
+    let mut client = ImdsClient::new();
+    let mut lifecycle_state = get_lifecycle_state(&mut client).await?;
+    info!("Lifecycle state is {}", lifecycle_state);
+    while lifecycle_state.ne("InService") {
+        lifecycle_state = get_lifecycle_state(&mut client).await?;
+        info!("Lifecycle state is {} ... waiting.", lifecycle_state);
+        sleep(Duration::from_secs(5)).await;
```
arnaldo2792, line 37: Shouldn't this `await` include `?`?

packages/selinux-policy/fs.cil
```diff
@@ -47,6 +47,7 @@
 (filecon "/.*/usr/bin/storewolf" file api_exec)
```
arnaldo2792, line 47: For the lack of a better place to leave this comment. This commit message isn't properly formatted.

arnaldo2792 (changes requested):

sources/hachiko/src/main.rs
gthao313: I don't think so : ). standard syntax of `tokio::time::sleep` is  `sleep(Duration).await;`. https://docs.rs/tokio/latest/tokio/time/fn.sleep.html

gthao313:
replace by https://github.com/bottlerocket-os/bottlerocket/pull/2562

PR #2328, opened 2022-08-10
kernel: Disable drivers for USB-attached network interfaces

markusboehme:
**Issue number:** n/a

**Description of changes:** There is no need to support USB-attached network interfaces in the cloud, and such devices are equally unlikely to be encountered in servers. Make sure not to build any drivers for them as they would just be dead weight.

We had those drivers for the 5.4 kernel variants (last `aws-k8s-1.19`, now gone) already and picked them up for the 5.10 kernel variants (any variants except `*-dev`) with release 1.9.0. This reverses course for the 5.10 kernel and ensures future upstream changes won't bring those drivers into the 5.15 kernel.

**Testing done:**

* [x] Builds and boots
* [x] Config differences in final builds are only those expected

The full list of resulting config changes can be found in [this Gist](https://gist.github.com/markusboehme/ba3478def8a1408dd36104df9112b183).

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.10/config-bottlerocket
+++ b/packages/kernel-5.10/config-bottlerocket
@@ -130,3 +130,6 @@ CONFIG_EXT4_USE_FOR_EXT2=y
 # CONFIG_RDS is not set
 # CONFIG_RFKILL is not set
 # CONFIG_TIPC is not set
+
+# Disable USB-attached network interfaces, unused in the cloud and on server-grade hardware.
+# CONFIG_USB_NET_DRIVERS is not set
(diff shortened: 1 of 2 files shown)


stmcginnis:
Do we still want these for metal and homelab types of deployments? They may be needed for those types of hardware setups.

cc @mchaker

mchaker:
In the case of a homelab variant, a separate kernel config would be necessary -- there would be a lot of hardware to support that shouldn't weigh down the official metal variant kernel config.

bcressey (approved):

foersleo (approved):

Issue #2204, opened 2022-06-09
Support static addressing via `net.toml`


zmrow:
- Depends on #2214 

**What I'd like:**
I would like the ability to set up a static IP for a network interface.

This means we will need to be able to set the following at the least

- IPv4/6 IP
- gateway
- nameservers
- destination IP

rickymulder:
Is there an opportunity to backport to the `1.9.x` branch?
I can file a new bug if needed, but to explain the short version is:

Attempting to spin up a version `1.21` eks cluster using bottlerocket without the ENI config using tunneling mode.  When creating pods on the host network, they inherit the hosts `/etc/resolv.conf` and therefore cannot do cluster dns resolution.  The routes already exist as when attempting to dig @ the `kube-dns` service IP resolution completes, but not all applications provide options to override the hosts `/etc/resolv.conf` or set custom NS.  Theoretically setting the hosts nameservers to the VPC DNS resolver with a backup as the `kube-dns` service IP would resolve that, at least for fully qualified DNS lookups.

#2353 would solve this specific use case.

Another option would be to expose the kubelet's `--resolv-conf` flag per [official docs](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/), though that would also require a manner to write said file.
> If you don't want this, or if you want a different DNS config for pods, you can use the kubelet's --resolv-conf flag. Set this flag to "" to prevent Pods from inheriting DNS. Set it to a valid file path to specify a file other than /etc/resolv.conf for DNS inheritance.

EDIT: The route to the service network does _not_ exist.  I'm barking up the wrong tree at the moment it would seem.

zmrow:
Hi @rickymulder !  We won't be backporting the static addressing feature, but when it's ready we'll release new versions of Bottlerocket that include the feature.

Your example seems to reference DNS settings, which should be released at the same time as the static addressing feature!

> EDIT: The route to the service network does not exist. I'm barking up the wrong tree at the moment it would seem.

^ Does this comment mean that the DNS settings _will not_ solve your issue?

PR #2330, opened 2022-08-10
netdog: refactor to prepare for upcoming additions

zmrow:
**Issue number:**
Related to #2204 

**Description of changes:**
This refactor is in preparation for the upcoming additions related to static addressing.

Almost no code was changed minus the few clippy lints that are in the last commit.  The majority of the code was moved around as is.
```
    netdog: Extract lease parsing functions to module
```
```
    netdog: refactor cli code into modules
    
    Since `main.rs` was starting to get large and unwieldy, and additional
    code is expected to be added, this commit refactors all the CLI code
    into separate files grouped under a `cli` module.
```
```
    netdog: implement the suggested clippy lints
```

**Testing done:**
* Unit tests continue to pass
* Build and run an `aws-k8s-1.21` image successfully (`current_ip`, `resolv.conf`, `primary_interface`, and `/etc/sysctl.d/90-primary_interface.conf` all get properly written)
```
-net.ipv6.conf.eth0.accept_ra = 2
```
* Build and run a `metal-dev` image successfully (`net.toml` correctly parsed and all above files written)
* Manually run `netdog`'s settings generator commands and observe proper output

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/cli/generate_hostname.rs
+++ b/sources/api/netdog/src/cli/generate_hostname.rs
@@ -1,0 +1,32 @@
+use super::{error, print_json, Result};
+use crate::CURRENT_IP;
+use argh::FromArgs;
+use dns_lookup::lookup_addr;
+use snafu::ResultExt;
+use std::fs;
+use std::net::IpAddr;
+use std::str::FromStr;
+
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "generate-hostname")]
+/// Generate hostname from DNS reverse lookup or use current IP
+pub(crate) struct GenerateHostnameArgs {}
+
+/// Attempt to resolve assigned IP address, if unsuccessful use the IP as the hostname.
+///
+/// The result is returned as JSON. (intended for use as a settings generator)
+pub(crate) fn run() -> Result<()> {
+    let ip_string = fs::read_to_string(CURRENT_IP)
+        .context(error::CurrentIpReadFailedSnafu { path: CURRENT_IP })?;
+    let ip = IpAddr::from_str(&ip_string).context(error::IpFromStringSnafu { ip: &ip_string })?;
+    let hostname = match lookup_addr(&ip) {
+        Ok(hostname) => hostname,
+        Err(e) => {
+            eprintln!("Reverse DNS lookup failed: {}", e);
+            ip_string
+        }
+    };
+
+    // sundog expects JSON-serialized output
+    print_json(hostname)
+}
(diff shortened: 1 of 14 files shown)


sources/api/netdog/src/net_config/mod.rs
```diff
@@ -29,6 +29,7 @@ pub(crate) trait Interfaces {
 
     /// Converts the network config into a list of `WickedInterface` structs, suitable for writing
     /// to file
+    #[allow(clippy::wrong_self_convention)]
     fn into_wicked_interfaces(&self) -> Vec<WickedInterface>;
```
zmrow: In the original PR for this code, we agreed on _not_ consuming `self`.  I don't want to change the interface as I think it's pretty clear so I'm ignoring the lint.

etungsten (approved):

zmrow:
The above force push moves the `parse_lease_info()` function into an associated function of `LeaseInfo`; it makes more sense this way. :)

etungsten (approved):

sources/api/netdog/src/net_config/mod.rs
cbgbt: I think `as_$THING` is the idiomatic naming for this sort of conversion.

cbgbt (approved):

sources/api/netdog/src/cli/generate_hostname.rs
```diff
@@ -0,0 +1,32 @@
+use super::{error, print_json, Result};
+use crate::CURRENT_IP;
```
bcressey, line 2: What do you think about moving the constants into some other module, like the parent?

bcressey (approved):
Nice!

sources/api/netdog/src/cli/generate_hostname.rs
zmrow: I spent more time than I'd like to admit playing and thinking about that...  I ended up keeping them where they were since they get used in the CLI _and_ other modules so I didn't really like them in `cli/mod.rs` and I also didn't really like the idea of making a top-level `lib.rs` just for constants.

Totally willing to entertain this though - perhaps there's an elegant way I just couldn't come up with! :smile:

sources/api/netdog/src/net_config/mod.rs
zmrow: ^ Ahhh, yeah I like that

zmrow:
^ Adds a commit changing the `into_wicked_interfaces()` function signature into `as_wicked_interfaces()`... much better.

cbgbt (approved):

PR #2335, opened 2022-08-11
Increase inotify default limits

### LINKED ISSUE #1525: Change default inotify settings
**What I'd like:**

In #371, there was discussion about adopting the learning from  [From 30 to 230 docker containers per host](http://sven.stormbind.net/blog/posts/docker_from_30_to_230/) . As part of this discussion, it looks like the inotify limits were discussed, but the default settings seemed to have been interpreted as what the settings should have been changed to ( see discrepancy in the settings [here](https://github.com/bottlerocket-os/bottlerocket/issues/371#issuecomment-538772527) compared to the referenced article). Tuning these settings to the recommendations in the article: 
```
fs.inotify.max_user_instances = 4096
fs.inotify.max_user_watches = 32768
```
will allow for more containers to run per node, assuming they're hitting the inotify limits already. 
Building a custom image with these settings tweaked. This will always be something that heavy users of inotify will have to do, but tweaking the defaults will probably provide a better developer experience for your average operator.

stmcginnis:
**Issue number:**

Closes #1525 

**Description of changes:**

We have had several reports from users that the inotify limits
(fs.inotify.max_user_instances and fs.inotify.max_user_watches) are too
low for their workloads, causing them to get errors when deploying pods.

The user data settings can be used to raise (or lower) these defaults if
an end user needs to fine tune their settings.

Related Amazon Linux changes:
- https://github.com/awslabs/amazon-eks-ami/pull/589
- https://github.com/awslabs/amazon-eks-ami/pull/614

**Testing done:**

Made changes and built image to make sure there were no errors.

Published AMI and spun up EKS cluster. Connected to console, went to admin container, and used
sheltie to verify values returned are what is expected for these settings.

**Terms of contribution:**

### DIFF
--- a/packages/release/release-sysctl.conf
+++ b/packages/release/release-sysctl.conf
@@ -58,3 +58,7 @@ kernel.unprivileged_bpf_disabled = 1
 
 # Turn on BPF JIT hardening, if the JIT is enabled.
 net.core.bpf_jit_harden = 2
+
+# Increase inotify limits to allow for a greater number of containers
+fs.inotify.max_user_instances = 8192
+fs.inotify.max_user_watches = 524288


stmcginnis:
@markusboehme and @foersleo - would be great if you could weigh in on this change too. Thanks!

markusboehme (approved):
I don't see any downsides to these changes. I have to say I was surprised by the high value for the `vm.max_map_count` sysctl, though. I wouldn't have thought of a process needing that many VMAs, but e.g. OpenSearch recommends configuring a minimum of 256k. Aside from the typical memory-mapped files, memory allocators will use anonymous maps to request memory from the kernel. To satisfy my curiosity, it'll be interesting to see which of these are the reason for OpenSearch's recommendation.

Not blocking approval on it, but as a minor improvement I'd prefer to see the changes split into two commits. The bump in inotify resource limits is independent of the limit on the number of VMAs per process.

stmcginnis:
Thanks @markusboehme! I've dropped the `vm.max_map_count` from this PR and will propose a separate one to handle that. For our immediate needs, the inotify settings are probably the more important ones.

jpmcb (commented):
This looks good to me and seems to be what `kubernetes/kops` and other container focused distros set their limits to:

https://github.com/kubernetes/kops/blob/f442cc2d0a163b501c28c335ff12234f16bcb38a/nodeup/pkg/model/sysctls.go#L114-L118

Although, more generally, I am curious, why these numbers? Simply because they are sane middle grounds; not too high, not too low?

stmcginnis:
> Although, more generally, I am curious, why these numbers? Simply because they are sane middle grounds; not too high, not too low?

Part middle ground so it just works for most end users without needing to make adjustments. Part just because we want to match what Amazon Linux uses as a default so the user experience is consistent if they switch between distros.

foersleo (approved):

PR #2337, opened 2022-08-12
fix clippy warnings for tools and run lints for pull requests

bcressey:
**Issue number:**

#1776 (partial fix)

**Description of changes:**
Fix or silence all clippy warnings for `tools`, and add a step to the pull request workflow so they stay fixed.

**Testing done:**
Verified that builds completed via `buildsys`.

Ran the following `pubsys` related tasks:
* ami
* repo
* validate-repo
* check-repo-expirations
* refresh-repo
* ssm
* promote-ssm
* upload-ova
* vmware-template

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -366,6 +366,27 @@ fi
 '''
 ]
 
+[tasks.check-lints]
+script = [
+'''
+rc=0
+
+# For rust first-party source code
+if ! cargo clippy \
+  --manifest-path ${BUILDSYS_TOOLS_DIR}/Cargo.toml \
+  --locked -- -D warnings --no-deps; then
+  rc=1
+fi
+
+# TODO: fix clippy lints in "sources" and check that too
+
+if [ "${rc}" -ne 0 ]; then
+  echo "Found lint warnings. First-party source code is checked with clippy." >&2
+  exit $rc
+fi
+'''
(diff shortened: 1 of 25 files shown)


bcressey:
Let me run `rustup` and fix some more lints. 😆

Makefile.toml
```diff
@@ -366,6 +366,25 @@ fi
...
 ]
 
+[tasks.check-lints]
+script = [
+'''
+rc=0
+
+# For rust first-party source code
+if ! cargo clippy \
+  --manifest-path ${BUILDSYS_TOOLS_DIR}/Cargo.toml \
```
etungsten, line 376: I guess the comment isn't exactly accurate yet. Are we planning to do this to first-party code in `/sources` as a follow up?

etungsten (approved):

Makefile.toml
bcressey: Yes, I'd like first-party code to be covered as well, but it's more than I had the time to tackle today. I was working on an unrelated update to buildsys and got annoyed by the existing `clippy` warnings. 😀

zmrow (approved):
:safety_vest:

cbgbt (approved):
:rocket:

PR #2339, opened 2022-08-12
better variant support in buildsys

### LINKED ISSUE #2211: Split metal and non-metal kernel config

bcressey:
**Issue number:**

Closes #2211

**Description of changes:**
This builds on the variant standardization work done in #2134 in a few ways.

Packages can now specify different levels of variant sensitivity; for example, they may only need to be rebuilt when the platform changes, rather than when any component changes.

Packages can make use of a new set of variant-related macros, such as `%{_cross_variant_platform}`, along with the existing `%{_cross_variant}` macro.

Packages can also take advantage of automatically generated [build conditionals](https://rpm-software-management.github.io/rpm/manual/conditionalbuilds.html) to handle conditional logic in the idiomatic way for RPM spec files. This is similar in spirit to `cfg` attributes emitted for Rust builds.

**Testing done:**
Built `aws`, `vmware`, and `metal` variants.

As expected, the kernel package emitted this `cargo` directive:
```
cargo:rerun-if-env-changed=BUILDSYS_VARIANT_PLATFORM
```

Confirmed that the correct files were included in `os`:
* `static-pods` for `*-k8s-*`
* `cfsignal` and `shibaken` for `aws-*`
* `pluto` for `aws-k8s-*`
* `ecs-settings-applier` for `*-ecs-*`
* `driverdog` for `*-nvidia`

**Terms of contribution:**


bcressey:
Rebase with no changes to confirm that the `clippy` lints pass.

packages/kernel-5.10/kernel-5.10.spec
```diff
@@ -96,9 +97,7 @@ scripts/kconfig/merge_config.sh \
   ../config-microcode \
 %endif
   %{SOURCE100} \
-%if %{_is_metal_variant}
-  %{SOURCE101}
-%endif
+  %{_sourcedir}/config-bottlerocket-%{_cross_variant_platform}
```
foersleo, line 100: Not quite sure if that would make sense, but could we also put the `_cross_variant_platform` substitution up into a single `Source101` label and just use that here?

As I said I am not sure which version is preferable. The variant as is in this commit has the advantage, that it clearly lists which possible variants we have available overall. But apart from that we would not necessarily need to list them as a `SourceXXX` if I am not mistaken (Comparing with `config-microcode` which comes from a separate package).

tools/buildsys/src/manifest.rs
```diff
@@ -44,13 +44,29 @@ package-name = "better.name"
...
+using the variant to affect its build process.
+
 ```
 variant-sensitive = true
 ```
 
+Some packages might only be sensitive to certain components of the variant
+tuple, such as the platform, runtime, or family. The `variant-sensitive` field
+can take also take a string to indicate the source of the sensitivity.
```
foersleo: nit: remove extra `take`

foersleo (commented):
The failing tests are due to a clippy warning for testsys. Should be gone with a rebase including https://github.com/bottlerocket-os/bottlerocket/commit/58651147f489dc91585d6175db56cd1374761aa2

I was thinking about possible ways to shave off unnecessary rebuilds by adding even finer grained control, like define sets of platforms (e.g. `[metal], [aws, vmware]`) and only rebuilding at transitions between these sets. But that is probably adding a lot of unnecessary complexity for rather small gain in build times, while also reducing flexibility. Any thoughts on that?

stmcginnis:
> I was thinking about possible ways to shave off unnecessary rebuilds by adding even finer grained control, like define sets of platforms (e.g. [metal], [aws, vmware]) and only rebuilding at transitions between these sets.

I like that idea. I think it might be worth trying to see how much time that could save.

bcressey:
> I was thinking about possible ways to shave off unnecessary rebuilds by adding even finer grained control, like define sets of platforms (e.g. `[metal], [aws, vmware]`) and only rebuilding at transitions between these sets. But that is probably adding a lot of unnecessary complexity for rather small gain in build times, while also reducing flexibility. Any thoughts on that?

This would be pretty trivial to do with serde, I think, just a matter of deserializing to a list of platforms (etc) rather than a string.

One my of back pocket changes is disabling i8042 altogether for the aws platform, since it keeps showing up as a potential 500+ ms stall at boot, and it isn't used for anything. It might be needed for the VMware console though.

Lustre is in the opposite category, a driver that's aws-only that could usefully be disabled for other platforms. Xen might be another candidate.

So I see specializing the kernel config for these use cases as somewhat desirable or inevitable, which means the build time savings would disappear. But again, the added complexity isn't very much.

zmrow (approved):
This all makes sense and I like the conciseness of it all!

packages/kernel-5.10/kernel-5.10.spec
bcressey: For RPM packaging in general it's bad practice to conditionally include sources in a spec file, because then you might get a different SRPM based on the architecture of the Koji build host that builds it from the SCM checkout, or you might distribute an SRPM that only works for one of the distribution's supported architectures.

It's kind of irrelevant in the context of this project, where the source files are always available and SRPMs aren't even built, but I've worked with RPM packaging long enough to have an ingrained preference for doing it this way.

foersleo (approved):
Thanks for clarifying those questions. Splitting the config more and fitting it closer onto each use case makes sense to me.

Dockerfile
```diff
@@ -93,7 +93,11 @@ RUN rpmdev-setuptree \
    && echo "%_cross_repo_root_json %{_builddir}/root.json" >> .rpmmacros \
    && echo "%_topdir /home/builder/rpmbuild" >> .rpmmacros \
    && rm ${ARCH} shared rust cargo \
-   && mv *.spec rpmbuild/SPECS \
+   && echo "%bcond_without $(V=${VARIANT_PLATFORM,,}; echo ${V//-/_})_platform" > .bconds \
+   && echo "%bcond_without $(V=${VARIANT_RUNTIME,,}; echo ${V//-/_})_runtime" >> .bconds \
+   && echo "%bcond_without $(V=${VARIANT_FAMILY,,}; echo ${V//-/_})_family" >> .bconds \
+   && echo "%bcond_without $(V=${VARIANT_FLAVOR:-no}; V=${V,,}; echo ${V//-/_})_flavor" >> .bconds \
```
arnaldo2792, line 99: I saw that this will end up as `with_nvidia_flavor` for NVIDIA variants, and one of its uses is to include `driverdog`. I wonder if we should be proactive and instead of doing `with_nvidia_flavor` we could have something more generic for future hardware integrations (AMD, Inferentia, etc).

arnaldo2792 (approved):

bcressey:
above force push takes away the redundant "take"

Dockerfile
bcressey: I'd prefer to wait until we have some other flavors to figure out what this should look like. GPU support is a likely source of "flavor" but we could have others, like a FIPS flavor, that don't have a hardware component.

arnaldo2792 (approved):

tools/buildsys/Cargo.toml
```diff
@@ -9,6 +9,7 @@ publish = false
 exclude = ["README.md"]
 
 [dependencies]
+bottlerocket-variant = { version = "0.1", path = "../../sources/bottlerocket-variant" }
```
webern, line 12: yay. but also weird. do we have tools crates depending on sources crates and sources crates depending on tools crates? should we consider having a `common` directory someday?

PR #2353, opened 2022-08-16
Add configurable DNS settings

zmrow:
**Issue number:**
Related to #2204 

**Description of changes:**
It's probably easiest to review this commit by commit, rather than as a whole.  All the pieces are pretty small.
```
    netdog: Add dns module

    This commit adds a new `dns` module to centralize the logic of gathering
    DNS information from an API-generated configuration file, and merging
    any missing settings from a given DHCP lease.  This module also handles
    writing the `/etc/resolv.conf` using the DNS information
```
```
    netdog: Update `install` subcommand to use `dns` module

    This change updates the `install` subcommand to make use of the
    newly-added `dns` module.  The most important change here is that the
    code will use DNS settings from config if they exist, supplementing any
    missing settings with settings from DHCP lease.  The code also takes
    into account DHCP6 lease renewals when rewriting the `resolv.conf`.
    When multiple lease files exist, the code uses the DHCP4 lease to gather
    DHCP lease data
```
```
    netdog: Add new `write-resolv-conf` subcommand

    This commit adds a new `write-resolv-conf` subcommand, which is meant to
    be used as a restart command for DNS API settings.  It will write the
    `resolv.conf` using DNS settings if they exist, supplementing any
    missing settings with settings from the DHCP lease of the primary
    interface
```
```
    settings: Add DNS configurability via `settings.dns`

    This commit adds the ability to configure DNS name servers and search
    suffixes via a new settings prefix `settings.dns`.  These DNS settings
    populate a configuration file which gets used by `netdog` to write the
    `resolv.conf`.  `netdog write-resolv-conf` is triggered via restart
    command.  If DNS settings exist, they are used to write the
    `resolv.conf`, otherwise the DHCP lease for the primary interface is
    used
```
```
    migrations: Add migrations for `settings.dns`

    This adds the migrations needed for `settings.dns`, one for the setting,
    its service and configuration file, and one for the setting metadata
```
```
    README: Add DNS settings documentation
```

**Testing done:**
* Boot an `aws-k8s-1.22` instance and confirm `resolv.conf` is still written correctly *with no* DNS settings
* Boot a `metal-dev` image with _and_ without user data and ensure `resolv.conf` is written properly with the correct settings
* Manually change DNS settings via `apiclient` and observe `resolv.conf` written with desired settings, using DHCP if only one of the settings is set.
```
$ cat /etc/resolv.conf 
search us-west-2.compute.internal
nameserver 192.168.0.2

$ apiclient set -j '{"dns": {"name-servers": ["1.2.3.4", "2.3.4.5"]}}'

$ cat /etc/resolv.conf                                                                                       
search us-west-2.compute.internal
nameserver 1.2.3.4
nameserver 2.3.4.5

$ apiclient set -j '{"dns": {"search-list": ["foo.bar", "baz.foo"]}}'

$ cat /etc/resolv.conf 
search foo.bar baz.foo
nameserver 1.2.3.4
nameserver 2.3.4.5
```
* Test migrations work properly

**Terms of contribution:**


zmrow:
^ Rebases on develop and fix the merge conflict in `Release.toml`

zmrow:
^ Fixed a clippy lint I originally missed

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// Create a DnsSettings from TOML config file
+    pub(crate) fn from_config() -> Result<Option<Self>> {
+        Self::from_config_impl(DNS_CONFIG)
+    }
+
+    fn from_config_impl<P>(path: P) -> Result<Option<Self>>
+    where
+        P: AsRef<Path>,
+    {
+        // Ensure we don't attempt to load a nonexistent or empty config file At boot time, the
```
arnaldo2792: Nit: capital `A` in `At boot time`

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,259 @@
...
+        // We can assume if the file is empty that the settings don't exist.
+        let path = path.as_ref();
+        let config_exists = if Path::exists(path) {
+            let file_len = fs::metadata(path)
+                .context(error::DnsConfMetaSnafu { path })?
+                .len();
+            file_len != 0
+        } else {
+            false
+        };
```
arnaldo2792, line 70: Nit: How about this:

```rust
let config_exists = Path::exists(path) && fs::metadata(path)
    .context(error::DnsConfMetaSnafu { path })?
    .len() != 0
```

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,259 @@
...
+
+        if config_exists {
+            let config_str =
+                fs::read_to_string(path).context(error::DnsConfReadFailedSnafu { path })?;
+            let dns_config =
+                toml::from_str(&config_str).context(error::DnsConfParseSnafu { path })?;
+
+            Ok(Some(dns_config))
+        } else {
+            eprintln!("No DNS configuration exists in {}", DNS_CONFIG);
```
arnaldo2792, line 80: Could you please elaborate why is the message send to the `stdout`? A missing `DNS_CONFIG` file doesn't represent an error, right?

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,259 @@
...
+            writeln!(output, "search {}", s.join(" "))
+                .context(error::ResolvConfBuildFailedSnafu)?;
+        }
+
+        for n in &self.nameservers {
+            writeln!(output, "nameserver {}", n).context(error::ResolvConfBuildFailedSnafu)?;
+        }
+
+        fs::write(path, output).context(error::ResolvConfWriteFailedSnafu { path })?;
+        Ok(())
```
arnaldo2792: I think you should be able to just do:

```rust
fs::write(path, output).context(error::ResolvConfWriteFailedSnafu { path })?
```

(Or maybe without the `?`)

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,259 @@
...
+        let format2 =
+            "search us-west-2.compute.internal\nnameserver 1.2.3.4\nnameserver 192.168.0.2\n";
+
+        settings.write_resolv_conf_impl(&fake_file).unwrap();
+        let resolv_conf = std::fs::read_to_string(&fake_file).unwrap();
+        let is_format1 = resolv_conf == format1;
+        let is_format2 = resolv_conf == format2;
+
+        // The resulting file must be either format 1 or 2
+        assert_ne!(is_format1, is_format2)
```
arnaldo2792: Nit:

```
assert!(resolve_conf == format1 || resolve_conf == format2)
```

sources/api/netdog/test_data/dns/leaseinfo.eth0.dhcp.ipv4.multiple-dns
```diff
@@ -0,0 +1,20 @@
+# The following links are the code that formats the DNSSERVERS portion of the file
+# https://github.com/openSUSE/wicked/blob/1c2c048d9ca64cf08d2223a15cfa2a06442cb99d/src/leaseinfo.c#L411
+# https://github.com/openSUSE/wicked/blob/1c2c048d9ca64cf08d2223a15cfa2a06442cb99d/src/leaseinfo.c#L120
```
arnaldo2792, line 3: Nit: shouldn't this comments be in the struct definition of `LeaseInfo`?

packages/release/netdog.template
```diff
@@ -0,0 +1,6 @@
+{{#if settings.dns.name-servers}}
```
arnaldo2792, line 1: Nit: For the lack of a better place to leave this comment:

The subject for this commit doesn't include a "subsystem", in other commits `netdog` was used.

Release.toml
```diff
@@ -140,5 +140,8 @@ version = "1.9.1"
     "migrate_v1.9.0_kubelet-no-daemon-reload.lz4",
```
arnaldo2792: Nit: for the lack of a better place for this comment

The subject for this commit is missing a "subsystem", in other commits `netdog` was used.

sources/api/migration/migrations/v1.10.0/dns-settings/src/main.rs
```diff
@@ -0,0 +1,25 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added new settings under `settings.kernel.modules` for configuring
```
arnaldo2792: The comment doesn't match what the migration is changing, it should be something like `settings.dns.*`

README.md
```diff
@@ -627,6 +627,17 @@ In addition to the container runtime daemons, these credential settings will als
    10.1.1.1 test2.example.com
    ```
 
+The following allows for custom DNS settings, which are used to generate the `/etc/resolv.conf`.
+If neither DNS setting is populated, the system will use the DHCP lease of the primary interface to gather these setings.
+See the `resolv.conf` [man page](https://man7.org/linux/man-pages/man5/resolv.conf.5.html) for more detail.
+* `settings.dns.name-servers` (array of IP address strings): The desired name server IP address(es).
+* `settings.dns.search-list` (array of quoted domain strings): The desired domain search path(s).
```
arnaldo2792: You could include the "type" as part of the descriptions?

```markdown
`settings.dns.name-servers`: An array of IP addresses that represent the desired name server
```

arnaldo2792 (changes requested):

sources/api/netdog/src/dns.rs
zmrow: I mention this is the (rather lengthy) comment above this code but the tl;dr is that on first boot this file won't exist (which is fine) and I don't want to return an error when attempting to run `fs::metadata()` against it.

I initially wrote some code running `fs::metadata()` and matching on the error that was returned, but I thought that code was less readable and clear than this.  I'm totally willing to debate that and change it though. :)

sources/api/netdog/src/dns.rs
zmrow: It doesn't represent an error, but it's helpful to log something in the case a user doesn't supply DNS settings _and_ isn't using DHCP.

sources/api/netdog/src/dns.rs
zmrow: Good catch - thanks!

zmrow:
^ Addresses @arnaldo2792 's comments

zmrow:
^ Rebases on `develop` and fixes the `sources/Cargo.toml` merge conflict

zmrow:
^ Changes the type of `DnsSettings::name-servers` to `Option` since a user may not supply it... (It was an oversight, `search-list` was already `Option`)

sources/models/shared-defaults/defaults.toml
```diff
@@ -155,3 +155,15 @@ affected-services = ["bootstrap-containers"]
...
+
+# DNS
+[metadata.settings.dns]
+affected-services = ["dns"]
+
+[services.dns]
+configuration-files = ["netdog"]
+restart-commands = ["netdog write-resolv-conf"]
+
+[configuration-files.netdog]
```
etungsten: nit: to follow prior convention
```suggestion
[configuration-files.netdog-toml]
```

Need to make corresponding change in migration as well

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,258 @@
...
+            Ok(None)
+        }
+    }
+
+    /// Create a DnsSettings from LeaseInfo struct
+    pub(crate) fn from_lease_info(info: &LeaseInfo) -> Self {
+        // Randomize name server order, for libc implementations like musl that send
+        // queries to the first N servers.
+        let mut dns_servers: Vec<IpAddr> = info.dns_servers.clone().into_iter().collect();
+        dns_servers.shuffle(&mut thread_rng());
```
etungsten: Is this randomization something we have to do, otherwise the resolver won't work? Or does the libc resolver library do the randomization?

etungsten (approved):

sources/api/netdog/src/dns.rs
zmrow: This randomization is something that was in the previous code, I kept it.

zmrow:
^ Addresses @etungsten 's coments

sources/api/migration/migrations/v1.10.0/dns-settings/src/main.rs
```diff
@@ -0,0 +1,24 @@
...
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added new settings under `settings.dns` for configuring /etc/resolv.conf
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec![
+        "settings.dns",
+        "services.dns",
+        "configuration-files.netdog",
```
etungsten: Now that we've change the name to `netdog-toml`, although this will still work, I think it'd be better to be more specific.
```suggestion
        "configuration-files.netdog-toml",
```

zmrow: :)  Thanks for pointing that out - I was going back through to make sure I hadn't missed anything and didn't catch that yet.

zmrow:
^ Fixes the migration filename (thanks @etungsten!)

etungsten (approved):

arnaldo2792 (approved):

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,258 @@
...
+            false
+        };
+
+        if config_exists {
+            let config_str =
+                fs::read_to_string(path).context(error::DnsConfReadFailedSnafu { path })?;
+            let dns_config =
+                toml::from_str(&config_str).context(error::DnsConfParseSnafu { path })?;
+
+            Ok(Some(dns_config))
```
bcressey: In the `from_lease_info()` path, if we're given a lease info file with no relevant fields, we return a `DnsSettings` struct with both optional fields set to `None`.

Would it make sense to do that here as well, and always return a `Result<Self>` instead?

sources/api/netdog/src/cli/install.rs
```diff
@@ -57,35 +55,26 @@ pub(crate) fn run(args: InstallArgs) -> Result<()> {
...
+            let dns_settings = if let Some(settings) =
+                DnsSettings::from_config().context(error::DnsFromConfSnafu)?
+            {
+                settings
+            } else {
+                DnsSettings::from_lease_info(&info)
+            };
+            dns_settings
+                .write_resolv_conf()
+                .context(error::ResolvConfWriteFailedSnafu)?;
```
bcressey: This relates to `from_config()` returning a valid  `DnsSettings` struct - potentially we want to divide this up rather than making it all API settings or all lease settings:
* if we have name-servers in the config, use those, otherwise use the lease
* if we have a search-list in the config, use it, otherwise use the lease

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -0,0 +1,49 @@
...
+            settings
+        } else {
+            let lease_path = primary_interface_lease()?;
+            DnsSettings::from_lease_file(lease_path).context(error::DnsFromLeaseSnafu)?
+        };
+    dns_settings
+        .write_resolv_conf()
+        .context(error::ResolvConfWriteFailedSnafu)?;
+    Ok(())
+}
```
bcressey: Can the merging of the DNS config settings and the lease settings be a helper function? Especially if becomes slightly more complicated than using one source or the other.

We're doing almost but not quite exactly the same thing as in the DHCP helper entry point, which seems like it could lead to bugs.

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    // If both ipv4 and ipv6 leases exist, use the ipv4 lease for DNS settings
+    let ipv4_exists = Path::exists(&ipv4);
+    let ipv6_exists = Path::exists(&ipv6);
+    match (ipv4_exists, ipv6_exists) {
+        (true, true) => Ok(ipv4),
+        (true, false) => Ok(ipv4),
+        (false, true) => Ok(ipv6),
+        (false, false) => error::MissingLeaseSnafu { path: LEASE_FOLDER }.fail(),
+    }
```
bcressey: We entirely ignore the DHCP6 lease coming from the "install" path - it seems like we'd want the behavior to be consistent, otherwise we'll lose the IPv6 version of the settings the next time the lease is renewed and `resolv.conf` is rewritten.

zmrow:
^ Addresses @bcressey 's comments.  The highlights:
* `DnsSettings` public interface is much smaller, preferring creation from config and merging in settings from DHCP lease as necessary.  This reduces the paths possible, addressing the possibility for bug-ridden multiple paths to the DNS settings.
* DNS settings from DHCP lease are used if the DNS API settings aren't set
* The `install` DHCP helper now correctly uses the proper DHCP path when gathering DNS settings.  This means we won't inadvertently lose settings on DHCP lease renewal, and ensures the `write-resolv-conf` and `install` paths are doing the same thing.

sources/api/netdog/src/cli/install.rs
```diff
@@ -52,37 +50,23 @@ pub(crate) fn run(args: InstallArgs) -> Result<()> {
...
-                    path: &args.data_file,
-                })?;
-            // Randomize name server order, for libc implementations like musl that send
-            // queries to the first N servers.
-            let mut dns_servers: Vec<_> = info.dns_servers.iter().collect();
-            dns_servers.shuffle(&mut thread_rng());
-            write_resolv_conf(&dns_servers, &info.dns_search)?;
-            write_current_ip(&info.ip_address.addr())?;
-        }
-        _ => eprintln!("Unhandled 'install' command: {:?}", &args),
```
zmrow, line 67: The compiler was yelling at me to remove this line... I kept the `match` syntax here as it felt the most clean.  I also didn't want to throw a grenade into the current arg parsing since failing to parse these args means the helpers fail.

zmrow:
^ Above push avoids returning an error if the DHCP lease doesn't exist.  The lease won't exist in the case a user is not using DHCP (static addressing).

It also makes the DHCP lease optional when constructing a `DnsSettings`.

zmrow:
^ Rebases on develop and fixes the merge conflict with `Release.toml`

sources/api/netdog/src/cli/install.rs
```diff
@@ -52,37 +50,26 @@ pub(crate) fn run(args: InstallArgs) -> Result<()> {
...
     match (&args.interface_type, &args.interface_family) {
-        (InterfaceType::Dhcp, InterfaceFamily::Ipv4) => {
-            let info =
-                LeaseInfo::from_lease(&args.data_file).context(error::LeaseParseFailedSnafu {
-                    path: &args.data_file,
+        (InterfaceType::Dhcp, InterfaceFamily::Ipv4 | InterfaceFamily::Ipv6) => {
+            // A lease should exist when using DHCP
+            let primary_lease_path =
+                lease_path(&primary_interface).context(error::MissingLeaseSnafu {
+                    interface: primary_interface,
```
bcressey, line 57: I'd still like to use `args.data_file` in some way, if only to check that it matches `primary_lease_path` with an early return if it doesn't.

Otherwise we potentially end up rewriting `resolv.conf` whenever any DHCP lease is renewed, instead of just the one for the primary interface.

Also, depending on the order in which leases are obtained, we might not have a lease for the primary interface yet early in the boot, if other interfaces are also using DHCP.

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,253 @@
...
+    /// Merge missing DNS settings into `self` using DHCP lease
+    fn merge_lease(&mut self, lease: &LeaseInfo) {
+        if self.nameservers.is_none() {
+            // Randomize name server order, for libc implementations like musl that send
+            // queries to the first N servers.
+            let mut dns_servers: Vec<IpAddr> = lease.dns_servers.clone().into_iter().collect();
+            dns_servers.shuffle(&mut thread_rng());
+
+            self.nameservers = Some(BTreeSet::from_iter(dns_servers.into_iter()));
+        }
```
bcressey: I'd prefer to shuffle the nameservers after the merge, just before writing `resolv.conf` - to guarantee that they're always shuffled. Here we don't shuffle if the nameservers are set via API.

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,253 @@
...
+            .nameservers
+            .is_some());
+        assert!(DnsSettings::from_config_impl(&config)
+            .unwrap()
+            .search
+            .is_some())
+    }
+
+    #[test]
+    fn empty_config() {
```
bcressey, line 170: Maybe add a test for `missing_config` too, where the file doesn't exist.

sources/api/netdog/src/main.rs
```diff
@@ -47,6 +52,7 @@ static PRIMARY_INTERFACE: &str = "/var/lib/netdog/primary_interface";
 static DEFAULT_NET_CONFIG_FILE: &str = "/var/lib/bottlerocket/net.toml";
 static PRIMARY_SYSCTL_CONF: &str = "/etc/sysctl.d/90-primary_interface.conf";
 static SYSTEMD_SYSCTL: &str = "/usr/lib/systemd/systemd-sysctl";
+static LEASE_FOLDER: &str = "/var/run/wicked";
```
bcressey: nit: 🪟 -> 🐧 
```suggestion
static LEASE_DIR: &str = "/var/run/wicked";
```

sources/api/netdog/src/cli/install.rs
zmrow: Good point about the timing!

Also, there's a check just before the `match` statement that returns if the install interface isn't the primary interface

zmrow:

I also simplified a few tests - I was calling `DnsSettings::from_config_impl()` multiple times when it was unnecessary...

sources/api/netdog/src/main.rs
```diff
@@ -47,6 +52,7 @@ static PRIMARY_INTERFACE: &str = "/var/lib/netdog/primary_interface";
 static DEFAULT_NET_CONFIG_FILE: &str = "/var/lib/bottlerocket/net.toml";
 static PRIMARY_SYSCTL_CONF: &str = "/etc/sysctl.d/90-primary_interface.conf";
 static SYSTEMD_SYSCTL: &str = "/usr/lib/systemd/systemd-sysctl";
+static LEASE_DIR: &str = "/run/wicked";
```
zmrow, line 55: I previously had `/var/run` which is a symlink to `/run`... better to just use the actual directory as that's what `wicked` uses when calling `install`.

bcressey (approved):

rickymulder:
Is there plans to merge this into `1.9.x`?  Attempting to spin up an EKS 1.21 cluster with bottlerocket and my host-network pods can't do cluster dns resolution without calling out the `kube-dns` service ip.

sources/api/netdog/src/dns.rs
```diff
@@ -0,0 +1,255 @@
...
+            "search us-west-2.compute.internal\nnameserver 192.168.0.2\nnameserver 1.2.3.4\n";
+        let format2 =
+            "search us-west-2.compute.internal\nnameserver 1.2.3.4\nnameserver 192.168.0.2\n";
+
+        // The resulting file must be either format 1 or 2
+        let resolv_conf = std::fs::read_to_string(&fake_file).unwrap();
+        assert_ne!(resolv_conf == format1, resolv_conf == format2)
+    }
+
+    #[test]
```
webern, line 239: Nice tests!

PR #2368, opened 2022-08-23
tools: add diff-kernel-config to identify kernel config changes

markusboehme:
**Issue number:** n/a

**Description of changes:**

```
tools: add diff-kernel-config to identify kernel config changes

With multiple kernel versions being supported for two architectures in
cloud and metal flavors it can be tedious and error-prone to get the
full picture on the effects changes have on the resulting kernel
configurations. The new diff-kernel-config tool helps automate this
task.

diff-kernel-config compares kernel configurations before and after a
series of commits to the Bottlerocket repository. For this, it runs
several full kernel builds that cover all possible combinations of
kernel version, architecture, and cloud/metal flavor that could be
shipped in any of the official variants. Combinations not used in any
variant are skipped.

diff-kernel-config produces kernel config diffs broken down by
combination in its output directory, alongside a comprehensive report of
all changes and a summary that can aid the code review process.
```

**Testing done:**

* [x] ShellCheck is happy
* [x] Used various development versions of this for my previous pull requests
* [x] Sample run with [the commit disabling the USB network drivers](https://github.com/bottlerocket-os/bottlerocket/commit/781e6b4018537473720849d6bc34f369e2cf8e29), output below:

```plain
$ time ./tools/diff-kernel-config -a fork/fix/disable-usb-net-drivers -b fork/fix/disable-usb-net-drivers^ -k 5.10 -k 5.15 -o configs
[...]

config-aarch64-5.10-aws-k8s-1.23-diff:   42 removed,   0 added,   1 changed
config-aarch64-5.15-aws-dev-diff:         0 removed,   0 added,   0 changed
config-aarch64-5.15-metal-dev-diff:       0 removed,   0 added,   0 changed
config-x86_64-5.10-aws-k8s-1.23-diff:    35 removed,   0 added,   1 changed
config-x86_64-5.10-metal-k8s-1.23-diff:  35 removed,   0 added,   1 changed
config-x86_64-5.15-aws-dev-diff:          0 removed,   0 added,   0 changed
config-x86_64-5.15-metal-dev-diff:        0 removed,   0 added,   0 changed

A full report has been placed in 'configs/diff-report'

real    64m21.385s
user    3m31.946s
sys     1m30.428s
```

**Terms of contribution:**

### DIFF
--- a/tools/diff-kernel-config
+++ b/tools/diff-kernel-config
@@ -1,0 +1,260 @@
+#!/usr/bin/env bash
+
+#
+# Common error handling
+#
+
+exit_trap_cmds=()
+
+on_exit() {
+    exit_trap_cmds+=( "$1" )
+}
+
+run_exit_trap_cmds() {
+    for cmd in "${exit_trap_cmds[@]}"; do
+        eval "${cmd}"
+    done
+}
+
+trap run_exit_trap_cmds EXIT
+
+bail() {
+    if [[ $# -gt 0 ]]; then
+        >&2 echo "Error: $*"
+    fi
+    exit 1
+}
+
+usage() {
+    cat <<EOF
+Usage: $0 -b GITREV_BEFORE -a GITREV_AFTER -o OUTPUT_DIR [-k KERNEL_VERSION] [-h]
+Compare kernel configurations before and after a series of commits.
+
+    -a, --after         new Git revision to compare from
+    -b, --before        baseline Git revision to compare against
+    -k, --kernel        kernel versions to compare configs for, may be given
+                        multiple times (optional, defaults to all kernels)
+    -o, --output-dir    path to the output directory; must not exist yet
+    -h, --help          show this help text
+
+Example invocation:
+
+    This compares the config changes for kernels 5.10 and 5.15 before and
+    after the most recent commit has been applied:
+
+        $0 -b HEAD^ -a HEAD -k 5.10 -k 5.15 -o configs
+
+Notes:
+
+    This compares the config changes for all combinations of aarch64/x86_64,
+    cloud/metal, and kernel versions. Combinations without a corresponding
+    Bottlerocket variant are skipped. Since this involves numerous full kernel
+    builds the comparison will take some time. Consider the working tree this
+    is invoked on busy while the script is running.
+
+EOF
+}
+
+usage_error() {
+    >&2 usage
+    bail "$1"
+}
+
+
+#
+# Parse arguments
+#
+
+kernel_versions=()
+
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        -a|--after)
+            shift; gitrev_after_arg=$1 ;;
+        -b|--before)
+            shift; gitrev_before_arg=$1 ;;
+        -k|--kernel)
+            shift; kernel_versions+=( "$1" ) ;;
+        -o|--output-dir)
+            shift; output_dir=$1 ;;
+        -h|--help)
+            usage; exit 0 ;;
+        *)
+            usage_error "Invalid option '$1'" ;;
+    esac
+    shift
+done
+
+if [[ ${#kernel_versions[@]} -eq 0 ]]; then
+    kernel_versions=( 5.10 5.15 )
+else
+    for kver in "${kernel_versions[@]}"; do
+        case ${kver} in
+            5.10) continue ;;
+            5.15) continue ;;
+            *)    bail "Unknown kernel version '${kver}'" ;;
+        esac
+    done
+fi
+readonly kernel_versions
+
+[[ -n ${output_dir} ]] || usage_error 'require -o|--output-dir'
+[[ -e ${output_dir} ]] && bail "Output directory '${output_dir}' exists already, not touching it"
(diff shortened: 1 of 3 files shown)


tools/diff-kernel-config
```diff
@@ -0,0 +1,240 @@
...
+                    1) kernel_rpm=${kernel_rpms[0]} ;;
+                    *) bail "More than one kernel RPM found for ${debug_id}" ;;
+                esac
+
+
+                #
+                # Extract kernel config
+                #
+
+                mkdir -p "${output_dir}" || bail "Failed to create output directory '${output_dir}'"
```
foersleo: I believe this is something we should do before we start any build. Would be a shame if we spend the time of a build to find out that we can not create the output directory for some reason.

foersleo (commented):
This is awesome. Currently giving it a spin for some of my hardware enablement patches.

tools/diff-kernel-config
markusboehme: Thankfully a single build is comparatively cheap. Agree, though, for obvious mistakes (e.g. no write permissions) we should fail early. Changed in the force push.

foersleo (approved):
Reiterating that this is super helpful, especially with bigger config changes to check what else gets pulled in when switching kernel options.

I have used it a few times the past couple of days to straighten out my work on the hardware enablement.

arnaldo2792 (approved):
How can we let users about the existence of this script? Should we mention it in the main README or probably in the `QUICKSTART-LOCAL` doc?

markusboehme:
> How can we let users about the existence of this script? Should we mention it in the main README or probably in the `QUICKSTART-LOCAL` doc?

Users wouldn't be interested in this unless they spin their own variant and modify the kernel packages. I wouldn't expect it in either the top-level `README` (script is too narrow and specific) or `QUICKSTART-LOCAL` (script is not related to running local VMs). It might find a mention in a dedicated `README` in the kernel packages, but perhaps it's also fine to not mention it at all for now and do that when a number of contributions without config diff stats come in. What do you think?

tools/diff-kernel-config
```diff
@@ -0,0 +1,241 @@
...
+#
+# Post-process the collected pairs of "before" and "after" configs (generate diffs, a report, a summary)
+#
+
+# Get the helpful diffconfig script from the kernel source tree. We package it
+# in the kernel-archive RPM from where it can be extracted. Here we extract the
+# latest version of the script, but any kernel version and arch will do.
+latest_kver=$(printf '%s\n' "${kernel_versions[@]}" | sort -V | tail -n1)
+latest_archive_rpms=( ./build/rpms/bottlerocket-aarch64-kernel-"${latest_kver}"-archive-*.rpm )
+diffconfig=$(mktemp --suffix -bottlerocket-diffconfig)
```
bcressey, line 229: nit: would be good to add cleanup handler (e.g. `trap 'cleanup' EXIT`) to ensure the temp file doesn't leak

tools/diff-kernel-config
```diff
@@ -0,0 +1,241 @@
...
+rpm2cpio "${latest_archive_rpms[0]}" \
+    | cpio --quiet --extract --to-stdout \
+    | tar --xz --extract --to-stdout kernel-devel/scripts/diffconfig >"${diffconfig}"
```
bcressey, line 233: There is a lot to love in this script but this may be my favorite part.

tools/diff-kernel-config
```diff
@@ -0,0 +1,241 @@
...
+        /^+/   { added   += 1 }
+        / -> / { changed += 1 }
+        END    { printf \"${config_base}:\t%3d removed, %3d added, %3d changed\n\", removed, added, changed }
+    " "${config_diff}"
+done | sort -V
+echo
+
+# Generate combined report of changes
+head -v -n 999999 "${output_dir}"/*-diff >"${output_dir}"/diff-report
+echo "A full report has been placed in '${output_dir}/diff-report'"
```
bcressey, line 260: If the report itself is the only reason to keep the directory around, it might be a nicer interface to use a temporary directory for the output, and just copy the report out as a permanent, gitignore'd file at the end.

That would trade off the need to keep track of output directories that don't yet exist with the need to occasionally prune older report files.

bcressey (approved):
This is great!

tools/diff-kernel-config
markusboehme: Indeed. I originally planned on adding it for the happy path and littering /tmp for the unlikely failure path, but eventually even forgot about the former. To repent, I fixed both in the current revision. :smile:

tools/diff-kernel-config
markusboehme: Personally, I haven't really made up my mind yet on this matter. I've found use for both the combined report and the individual files. Let's see how folks will use it in the next PRs and decide one way or the other? I agree on simplifying the interface should it turn out the combined report is all that's getting used.

markusboehme:
The force push fixed leaking a temporary file and now comes with readme files in the kernel packages that point towards this script.

jpculp (approved):

Issue #2359, opened 2022-08-19
Prairiedog and the kernel may disagree about parsing boot config keys without values


markusboehme:
**Image I'm using:** Build of metal-dev variant fresh off develop

**What I expected to happen:**

When using the example user data from the boot settings section of the `README.md`, ...

```
[settings.boot.kernel-parameters]
"console" = [
  "tty0",
  "ttyS1,115200n8",
]
"crashkernel" = [
  "2G-:256M",
]
"slub_debug" = [
  "options,slabs",
]
"usbcore.quirks" = [
  "0781:5580:bk",
  "0a5c:5834:gij",
]

[settings.boot.init-parameters]
"log_level" = ["debug"]
"splash" = []
```

...I expected those settings to always reflect what the kernel and systemd would see on boot.

**What actually happened:**

Prairiedog may generate the following boot config...:

```
bash-5.1# cat /var/lib/bottlerocket/bootconfig.data 
kernel.slub_debug = "options,slabs"
kernel.crashkernel = "2G-:256M"
kernel.usbcore.quirks = "0781:5580:bk", "0a5c:5834:gij"
kernel.console = "tty0", "ttyS1,115200n8"
init.splash = 
init.log_level = "debug"
```

...which on next boot is read by the kernel as:

```
bash-5.1# cat /proc/bootconfig 
kernel.slub_debug = "options,slabs"
kernel.crashkernel = "2G-:256M"
kernel.usbcore.quirks = "0781:5580:bk", "0a5c:5834:gij"
kernel.console = "tty0", "ttyS1,115200n8"
init.splash = 'init.log_level = "debug"'
```

Note that the `log_level` key got pulled in as the value for the actually value-less `splash` key.

**How to reproduce the problem:**

Put the boot settings fragment from above into your user data. There seems to be a random component to it, and it may take a few (clean) boots to hit.

markusboehme:
As for the semi-random occurrence of the issue: The value-less key must not be the last line in the configured boot config. The order in which prairiedog writes out these keys is determined by Rust's `HashMap.iter()` method, which is documented to produce an "arbitrary" order which in my tests mostly, but not always, happened to be one that put `splash` last.

Clearly there is disagreement about how to represent value-less keys in the boot config, and in case of doubt the kernel's implementation wins. Prairiedog could put an empty pair of quotes after the value-less key. I'll check whether there's other ways to represent such keys, though (IIRC `init { splash }` would be one).

[...]

PR #2375, opened 2022-08-24
Automated reconciliation of boot settings

### LINKED ISSUE #2155: support automated reconciliation of kernel command line settings
**What I'd like:**
I'd like a way to provide kernel or init command line parameters via user data or bootstrap container, and have the node automatically reboot into a state where they're applied.

This can be optional, and off by default.

An example use case would be enabling cgroup v2:
```
[settings.boot]
reboot-to-reconcile = true

[settings.boot.init-parameters]
"systemd.unified_cgroup_hierarchy" = "1"
```

Handling the reboot in a bootstrap container, but this is potentially awkward if I have multiple bootstrap containers, some of which apply settings and one of which issues the reboot, since there is no explicit ordering among them.

markusboehme:
**Issue number:**

Closes #2155 

**Description of changes:**

This PR implements what has been proposed in #2155, the addition of a new `settings.boot.reboot-to-reconcile` setting to automatically reboot Bottlerocket if either the kernel or systemd command line were changed during boot. At a high level, prairiedog now compares any newly written bootconfig initrd against the one already present and flags the need for a reboot. Before entering the `configured` systemd target, a new service unit then triggers the reboot if needed.

To reviewers, I recommend going commit by commit. The commits are fairly detailed and hopefully tell a story if followed in order.

**Testing done:**

* [x] No boot settings specified: No automatic reboot; host can be rebooted
* [x] Boot settings specified with `reboot-to-reconcile` unset: No automatic reboot; host can be rebooted and comes up with specified settings
* [x] Boot settings specified with `reboot-to-reconcile` set to false: No automatic reboot; host can be rebooted and comes up with specified settings
* [x] Boot settings specified with `reboot-to-reconcile` set to true: Automatic reboot; host comes up with specified settings
* [x] Boot settings specified with `reboot-to-reconcile` set to true and bootstrap containers: Automatic reboot after bootstrap containers have exited; host comes up with specified settings
* [x] Tested forward and backward migration from 1.9.2 to a faux 1.10.0 with this PR on builds of the `aws-k8s-1.23` variant, both with and without `reboot-to-reconcile` set

**Notes:**

* ~While this implements a new setting, I don't think a migration is needed as the setting is persisted and restored by prairiedog in the bootconfig initrd. Older versions of prairiedog will just ignore the setting.~ Update: A migration is needed as pointed out by @etungsten: While prairiedog may ignore the unknown setting, it will bite when trying to deserialize the data store. The current version comes with a migration.
* If you're taking this on a test ride yourself, please see #2359: If you configure a boot setting that doesn't require a value, please specify the empty string for now.

**Terms of contribution:**

### DIFF
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -44,6 +44,7 @@ Source116: load-kernel-modules.service.in
 Source117: cfsignal.service
 Source118: generate-network-config.service
 Source119: prepare-primary-interface.service
+Source120: reboot-if-required.service
 
 # 2xx sources: tmpfilesd configs
 Source200: migration-tmpfiles.conf
@@ -402,7 +403,7 @@ install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 \
   %{S:100} %{S:101} %{S:102} %{S:103} %{S:105} \
   %{S:106} %{S:107} %{S:110} %{S:111} %{S:112} \
-  %{S:113} %{S:114} %{S:118} %{S:119} \
+  %{S:113} %{S:114} %{S:118} %{S:119} %{S:120} \
   %{buildroot}%{_cross_unitdir}
 
 %if %{with nvidia_flavor}
@@ -563,6 +564,7 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 
 %files -n %{_cross_os}prairiedog
 %{_cross_bindir}/prairiedog
+%{_cross_unitdir}/reboot-if-required.service
 
 %files -n %{_cross_os}certdog
 %{_cross_bindir}/certdog
--- a/packages/os/reboot-if-required.service
+++ b/packages/os/reboot-if-required.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Reboot system if needed for configuration changes to take effect
+# Block manual interactions for now to prevent unplanned reboots.
+RefuseManualStart=true
+RefuseManualStop=true
+After=configured.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/prairiedog reboot-if-required
+StandardError=journal+console
--- a/sources/api/prairiedog/src/bootconfig.rs
+++ b/sources/api/prairiedog/src/bootconfig.rs
@@ -16,6 +16,11 @@ const DEFAULT_BOOTCONFIG_STR: &str = r#"
     kernel = ""
     init = ""
 "#;
+const DEFAULT_BOOT_SETTINGS: BootSettings = BootSettings {
+    reboot_to_reconcile: None,
+    kernel_parameters: None,
+    init_parameters: None,
+};
 
 fn append_boot_config_value_list(values: &[BootConfigValue], output: &mut String) {
     for (i, v) in values.iter().enumerate() {
@@ -255,11 +260,63 @@ fn boot_config_to_boot_settings_json(bootconfig_str: &str) -> Result<String> {
     serde_json::to_string(&boot_settings).context(error::OutputJsonSnafu)
 }
 
+/// Decides whether the host should be rebooted to have its boot settings take effect
+pub(crate) async fn is_reboot_required<P>(socket_path: P) -> Result<bool>
+where
+    P: AsRef<Path>,
+{
+    let old_boot_settings = match read_proc_bootconfig().await? {
+        Some(proc_bootconfig) => parse_boot_config_to_boot_settings(&proc_bootconfig)?,
+        None => DEFAULT_BOOT_SETTINGS,
+    };
+
+    let new_boot_settings = get_boot_config_settings(socket_path)
+        .await?
+        .unwrap_or(DEFAULT_BOOT_SETTINGS);
+
+    let reboot_required = if new_boot_settings.reboot_to_reconcile.unwrap_or(false) {
+        boot_settings_change_requires_reboot(&old_boot_settings, &new_boot_settings)
+    } else {
+        false
+    };
+
+    Ok(reboot_required)
+}
+
+/// Check whether `model::BootSettings` changed in a way to warrant a reboot
+fn boot_settings_change_requires_reboot(
+    old_boot_settings: &BootSettings,
+    new_boot_settings: &BootSettings,
+) -> bool {
+    fn parameters_changed_materially(
+        old_params: &Option<HashMap<BootConfigKey, Vec<BootConfigValue>>>,
+        new_params: &Option<HashMap<BootConfigKey, Vec<BootConfigValue>>>,
+    ) -> bool {
+        // Consider a missing hash map equal to an empty one: There is no configuration in either case.
+        match (old_params, new_params) {
+            (None, None) => false,
+            (None, Some(new)) => !new.is_empty(),
+            (Some(old), None) => !old.is_empty(),
+            (Some(old), Some(new)) => old != new,
+        }
+    }
+
+    // Only reboot for changes actually requiring a reboot. Changing a Bottlerocket setting
+    // like boot.reboot-to-reconcile does not qualify as a reason to reboot.
+    parameters_changed_materially(
+        &old_boot_settings.kernel_parameters,
+        &new_boot_settings.kernel_parameters,
+    ) || parameters_changed_materially(
+        &old_boot_settings.init_parameters,
+        &new_boot_settings.init_parameters,
+    )
+}
+
 #[cfg(test)]
 mod boot_settings_tests {
     use crate::bootconfig::{
-        boot_config_to_boot_settings_json, serialize_boot_settings_to_boot_config,
-        DEFAULT_BOOTCONFIG_STR,
+        boot_config_to_boot_settings_json, boot_settings_change_requires_reboot,
+        serialize_boot_settings_to_boot_config, DEFAULT_BOOTCONFIG_STR,
     };
     use maplit::hashmap;
     use model::modeled_types::{BootConfigKey, BootConfigValue};
@@ -462,5 +519,75 @@ mod boot_settings_tests {
             )
             .unwrap()
         );
+    }
+
+    #[test]
+    fn test_unchanged_boot_settings_require_no_reboot() {
+        let a = BootSettings {
+            reboot_to_reconcile: None,
+            kernel_parameters: None,
+            init_parameters: to_boot_settings_params(hashmap! {
+                "systemd.log_level" => vec!["debug"],
+            }),
+        };
+        let b = BootSettings {
+            reboot_to_reconcile: None,
+            kernel_parameters: None,
+            init_parameters: to_boot_settings_params(hashmap! {
+                "systemd.log_level" => vec!["debug"],
+            }),
+        };
+        assert!(!boot_settings_change_requires_reboot(&a, &b));
+    }
+
+    #[test]
+    fn test_changed_boot_settings_require_a_reboot() {
+        let a = BootSettings {
+            reboot_to_reconcile: None,
+            kernel_parameters: None,
+            init_parameters: to_boot_settings_params(hashmap! {
+                "systemd.log_level" => vec!["debug"],
+            }),
+        };
+        let b = BootSettings {
+            reboot_to_reconcile: None,
+            kernel_parameters: to_boot_settings_params(hashmap! {
+                "debug" => vec![""],
+            }),
+            init_parameters: to_boot_settings_params(hashmap! {
+                "systemd.log_level" => vec!["debug"],
+            }),
+        };
+        assert!(boot_settings_change_requires_reboot(&a, &b));
+    }
+
+    #[test]
+    fn test_missing_boot_settings_require_no_reboot() {
+        let a = BootSettings {
+            reboot_to_reconcile: None,
+            kernel_parameters: None,
+            init_parameters: to_boot_settings_params(hashmap! {}),
+        };
+        let b = BootSettings {
+            reboot_to_reconcile: None,
+            kernel_parameters: to_boot_settings_params(hashmap! {}),
+            init_parameters: None,
+        };
+        assert!(!boot_settings_change_requires_reboot(&a, &b));
+    }
+
+    #[test]
+    fn test_changed_bottlerocket_boot_settings_require_no_reboot() {
+        let a = BootSettings {
+            reboot_to_reconcile: None,
+            kernel_parameters: None,
+            init_parameters: None,
+        };
+        let b = BootSettings {
+            reboot_to_reconcile: Some(true),
+            kernel_parameters: None,
+            init_parameters: None,
+        };
+        assert!(!boot_settings_change_requires_reboot(&a, &b));
     }
 }
--- a/sources/api/prairiedog/src/bootconfig.rs
+++ b/sources/api/prairiedog/src/bootconfig.rs
@@ -227,6 +227,7 @@ fn parse_boot_config_to_boot_settings(bootconfig: &str) -> Result<BootSettings> 
     }
 
     Ok(BootSettings {
+        reboot_to_reconcile: None,
         kernel_parameters: if kernel_params.is_empty() {
             None
         } else {
@@ -264,6 +265,7 @@ mod boot_settings_tests {
     #[test]
     fn boot_settings_to_string() {
         let boot_settings = BootSettings {
+            reboot_to_reconcile: None,
             kernel_parameters: Some(
                 hashmap! {
                     "console" => vec!["ttyS1,115200n8", "tty0"],
@@ -320,6 +322,7 @@ mod boot_settings_tests {
     #[test]
     fn none_boot_settings_to_string() {
         let boot_settings = BootSettings {
+            reboot_to_reconcile: None,
             kernel_parameters: None,
             init_parameters: None,
         };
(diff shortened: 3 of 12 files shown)


etungsten:
> I don't think a migration is needed as the setting is persisted and restored by prairiedog in the bootconfig initrd. Older versions of prairiedog will just ignore the setting.

We would still need a migration as we're introducing a new setting. If this new setting is not properly removed from the settings data-store on downgrade then the host will fail to boot because the settings model will not anticipate this new setting when apiserver tries to de-serialize the existing settings data-store.

markusboehme:
>> I don't think a migration is needed as the setting is persisted and restored by prairiedog in the bootconfig initrd. Older versions of prairiedog will just ignore the setting.
>
> We would still need a migration as we're introducing a new setting. If this new setting is not properly removed from the settings data-store on downgrade then the host will fail to boot because the settings model will not anticipate this new setting when apiserver tries to de-serialize the existing settings data-store.

I see, thanks! I was under the impression all serialization/deserialization for `BootSettings` would be handled by prairiedog, but clearly the settings are persisted in the data store as well.

While anything under and including `settings.boot` would be scrubbed by the migration that introduced that setting as a prefix in v1.8.0, this doesn't take care of any rollbacks to v1.8.0, v1.9.0, etc. I will add a new `AddSettingsMigration` for the `settings.boot.reboot-to-reconcile` key.

packages/os/bootstrap-containers@.service
```diff
@@ -27,3 +28,6 @@ ExecStartPost=/usr/bin/bootstrap-containers mark-bootstrap \
     --mode '${CTR_MODE}'
 RemainAfterExit=true
 StandardError=journal+console
+
+[Install]
+RequiredBy=reboot-to-reconcile.service
```
arnaldo2792: This will be overridden by a drop-in configuration:

https://github.com/bottlerocket-os/bottlerocket/blob/037d0e2798a7e8d4a12d0fb35d3c266db3d74f25/sources/api/bootstrap-containers/src/main.rs#L376

Regardless of ^, bootstrap containers shouldn't be `RequiredBy=reboot-to-reconcile.service`. Non-essential bootstrap containers don't block the boot sequence when they fail. By making all bootstrap containers `RequiredBy=reboot-to-reconcile.service`, failed non-essential bootstrap containers will indirectly block the boot sequence since `reboot-to-reconcile.service` is `RequiredBy=configured.target`.

bcressey: The right way to accomplish this might be to order `reboot-to-reconcile.service` via `activate-multi-user.service`:
```
diff --git a/packages/release/activate-multi-user.service b/packages/release/activate-multi-user.service
index 964301e9..bdb6f475 100644
--- a/packages/release/activate-multi-user.service
+++ b/packages/release/activate-multi-user.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Isolates multi-user.target
-After=configured.target
-Requires=configured.target
+After=configured.target reboot-to-reconcile.service
+Requires=configured.target reboot-to-reconcile.service
 
 [Service]
 Type=oneshot
```

These "activate" units run in a sort of weird limbo state - they're pulled in by the target, but only run after that target has been reached. We haven't had to do anything besides the target switch in these states before, but conceptually this is when we want the reconciliation reboot to happen.

packages/os/reboot-to-reconcile.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+Description=Reboot system if needed to reconcile boot settings
+# The marker file is put here by prairiedog after regenerating the initrd.
+ConditionPathExists=/run/reboot-required
+# Block manual interactions for now to prevent unplanned reboots.
+RefuseManualStart=true
+RefuseManualStop=true
+Before=configured.target
+Requires=settings-applier.service
+After=settings-applier.service
```
bcressey: nit: I expect these `Requires` and `After` lines are redundant and can be omitted.

Being in the `configured.target` phase implies that we've already done everything in the `preconfigured.target` phase, and for simplicity I'd prefer to avoid specifying unit dependencies across phases if possible.

packages/os/reboot-to-reconcile.service
```diff
@@ -0,0 +1,21 @@
...
+
+[Service]
+# Service type "notify" ensures the boot doesn't proceed past a required
+# reboot since the notification is never sent.
+Type=notify
```
bcressey: Clever!

bcressey (commented):
I don't quite understand the need to persist the `reboot-to-reconcile` flag within the actual bootconfig data.

Let me back up a step. As a rule I tend to find marker files distasteful, and one of my stronger architectural preferences is to favor API queries over ad-hoc files, since those files tend to create their own mini-API surfaces that aren't consistent.

(This isn't entirely binding - we use markers for `early-boot-config` and `cfnsignal`, and those can make sense if the alternative would be creating weird marker-file-like settings, like `settings.marker-files.early-boot-config-ran = true`, which I think would be taking the above principle too far.)

Two patterns that tend to avoid some of the downsides of applying this architectural preference:
* if a program would be overly complicated by querying the API, or if the API isn't always available to query, it might be better to render a config file for it based on settings and just have it load its own config (see: updog, netdog).
* if a program is writing a marker file so some other, easier to use utility can avoid querying the API, it might be better to add a sub-command to that program and wrap the execution of the utility instead (`prairiedog` sort of does this in spirit already by wrapping the `mount` command for the boot partition)

Applying that second pattern would lead me to expect a command like `prairiedog reboot-if-required` where we'd look at the bootconfig file we'd previously rendered, compare that with the kernel settings, and query the API to see if we need to reboot after all, and then run `systemctl reboot` ourselves. (I'm not saying this is right or better, just sharing where I'm coming from.)

Encoding the reconcile boolean in the bootconfig data could make sense if we want to optimize away that final API query - after all, we've already seen all the settings before and can record that fact somewhere - but then I'm not sure why we don't just write the marker file instead, if we find a delta after creating the config, or remove it if there's no longer a delta.

It's really that nested sequence of state that's not clear to me - why we have API settings and bootconfig and the marker file, and if we can do away with one or both of the latter bits of state.

packages/os/bootstrap-containers@.service
markusboehme: I don't think the drop-in created by bootstrap-containers is going to override the `RequiredBy` in the template, it should rather add an extra dependency. `systemd-analyze dot` seems to confirm this. Nonetheless, you're entirely right I missed to consider non-essential bootstrap containers. It goes to show the way the ordering was achieved here is rather clumsy.

I like Ben's suggestion. It takes care of the required distinction of essential/non-essential bootstrap containers and also appears semantically right: The host should reboot once it has been configured.

markusboehme:
> I don't quite understand the need to persist the `reboot-to-reconcile` flag within the actual bootconfig data.

Originally, this is mainly a result of misunderstanding the serialization of `BootSettings`, i.e. assuming they would *only* be persisted by prairiedog. Erikson clarified this in his earlier comment. There should be nothing in the way of dropping that setting from the bootconfig data again. When invoked by sundog, `prairiedog generate-boot-settings` will have to query the API for the current data and merge the kernel and init settings read from the bootconfig data into it. I will try this next--I expect it will get rid of quite a bit of extra code.

> Let me back up a step. As a rule I tend to find marker files distasteful, and one of my stronger architectural preferences is to favor API queries over ad-hoc files, since those files tend to create their own mini-API surfaces that aren't consistent. [...]
>
> It's really that nested sequence of state that's not clear to me - why we have API settings and bootconfig and the marker file, and if we can do away with one or both of the latter bits of state.

Thanks for taking the time to share your thoughts on the overall design! I think a part of the confusion stems from trying to see deeper reasons in having `reboot-to-reconcile` in the bootconfig datat when I thought of it as just a necessity. :blush:  Not doing that anymore will likely make things clearer.

I had considered, but not implemented, the idea of a new prairiedog command. Decoupling the decision to reboot and actually rebooting via a marker file was motivated by the idea a bootstrap container might find other reasons to request a reboot before the host enters its normal operational state. There is no such request for now, so perhaps it makes sense to drop this and go with the specific solution that keeps all logic inside prairiedog.

bcressey:
> Decoupling the decision to reboot and actually rebooting via a marker file was motivated by the idea a bootstrap container might find other reasons to request a reboot before the host enters its normal operational state. There is no such request for now, so perhaps it makes sense to drop this and go with the specific solution that keeps all logic inside prairiedog.

I was thinking that an alternative using helper commands might be a oneshot unit with multiple potential restart commands, e.g.:
```
ExecStart=/usr/bin/prairiedog reboot-if-required
ExecStart=/usr/bin/netdog reboot-if-required
ExecStart=/usr/bin/updog reboot-if-required
```

Where each program gets to encapsulate its own state checks and decision-making.

But your use case - allowing bootstrap containers to "schedule" a reboot after all other bootstrap containers have run, by writing a marker file in a well-known place - makes sense too, and wouldn't be possible with the sequenced helper commands.

I'm happy to keep the `/run/reboot-required` approach.

packages/os/bootstrap-containers@.service
bcressey: I forgot to mention that you'd also need to move `reboot-to-reconcile.service` to be `After=configured.target` instead of before, so it doesn't get started earlier than expected in the `configured.target` phase.

markusboehme:
@bcressey I see what you mean now. This definitely looks and feels more as if made from one piece, and doesn't preclude bootstrap containers from signalling a need to reboot either: There likewise could be a `/usr/bin/bootstrap-containers reboot-if-required`, checking e.g. a marker file that is local to a particular bootstrap container. This even would provide more flexibility (e.g. accountability, or honoring the request only for certain containers). I'm thus moving away from the global marker file in the next iteration of the PR.

bcressey:
> There likewise could be a `/usr/bin/bootstrap-containers reboot-if-required`, checking e.g. a marker file that is local to a particular bootstrap container.

Nice! I really like this idea.

markusboehme:
The force push changes the approach as discussed above:
* The reboot now happens after reaching the `configured.target`, sidestepping the need to explicitly consider bootstrap containers and whether they're essential or not.
* The initrd boot config now only contains the kernel and init command line arguments. The new `reboot-to-reconcile` setting is only persisted in the data store.
* `prairiedog` grew a new `reboot-if-required` command to compare the initrd boot config against the boot config the host booted with, issuing a reboot if that is required and `settings.boot.reboot-to-reconcile` is set to `true`. Other first-party components can likewise introduce `reboot-if-required` commands of their own and hook into the existing `reboot-if-required.service`.
* The `reboot-if-required` setting now comes with the migration needed for a rollback and has been tested.

There's merge conflicts to sort out in `Cargo.toml` and `Release.toml` needs merging as well. I refrained from rebasing for now to allow for easy diffing against the older revision. Will rebase to unblock the test process once folks have signaled they're in agreement with the current approach.

sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -249,6 +260,49 @@ fn boot_config_to_boot_settings_json(bootconfig_str: &str) -> Result<String> {
...
+        .unwrap_or(DEFAULT_BOOT_SETTINGS);
+
+    let mut reboot_required = false;
+
+    if new_boot_settings.reboot_to_reconcile.unwrap_or(false) {
+        /// Determines whether two optional hash maps changed. Treats empty and missing hash maps as equal.
+        fn has_changed_materially(
+            old_params: &Option<HashMap<BootConfigKey, Vec<BootConfigValue>>>,
+            new_params: &Option<HashMap<BootConfigKey, Vec<BootConfigValue>>>,
+        ) -> bool {
```
bcressey: It might be worth splitting this out into a helper function that takes two `BootSettings` and returns a `bool`, so that it can be unit-tested.

sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -249,6 +260,49 @@ fn boot_config_to_boot_settings_json(bootconfig_str: &str) -> Result<String> {
...
+            match (old_params, new_params) {
+                (None, None) => false,
+                (None, Some(new)) => !new.is_empty(),
+                (Some(old), None) => !old.is_empty(),
+                (Some(old), Some(new)) => old != new,
+            }
```
bcressey: The two middle cases were what triggered my "oh, I hadn't thought of that" reaction and prompted me to look for unit tests. 😀

sources/api/prairiedog/src/main.rs
```diff
@@ -277,6 +283,20 @@ fn load_crash_kernel() -> Result<()> {
     Ok(())
 }
 
+async fn reboot_if_required<P>(socket_path: P) -> Result<()>
+where
+    P: AsRef<Path>,
+{
+    if is_reboot_required(socket_path).await? {
+        info!("Boot settings changed and require a reboot to take effect. Initiating reboot...");
+        command("/usr/bin/systemctl", &["reboot"])?;
```
bcressey, line 294: Did we lose the property from the previous revision where the system would never continue starting other units?

The manual for `systemctl reboot` says it's mostly equivalent to:
```
systemctl start reboot.target --job-mode=replace-irreversibly --no-block
```

It might be good to somehow confirm that `activate-multi-user.target` never gets a chance to run. I'd guess that since it has `DefaultDependencies=yes` (implied) then it will conflict with `shutdown.target` and never actually start because the stop job will run first.

bcressey (approved):
Looks good!

packages/os/os.spec
```diff
@@ -48,6 +48,7 @@ Source116: load-kernel-modules.service.in
 Source117: cfsignal.service
```
arnaldo2792, line 44: Nit in the commit message: there is an extra `implement`

packages/os/reboot-if-required.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
```
arnaldo2792, line 1: I think we still want this unit to be a `oneshot` and configured with `RemainAfterExit`, right? According to the docs, since the unit has `ExecStart` it won't be considered `oneshot` by default, and the behavior of a `simple` service doesn't quite fit for `reboot-if-required`:

> the service manager will consider the unit started immediately after the main service process has been forked off.
>...
> the service manager will immediately proceed starting follow-up units, right after creating the main service process, and before executing the service's binary. Note that this means systemctl start command lines for simple services will report success even if the service's binary cannot be invoked successfully

arnaldo2792 (commented):
This is great!

sources/api/prairiedog/src/bootconfig.rs
markusboehme: Excellent point! I've done so now, and cleaned up a couple existing tests in the process (in a separate commit, of course).

sources/api/prairiedog/src/main.rs
markusboehme: We did, for the reason that Arnaldo noticed below: I accidentally changed the systemd service type to one that wouldn't wait for anything but a successful `fork`. Once `systemctl reboot` returns, from looking at the code (and later on logs as well) the system would be working on the shutdown already. I realize this is all a bit subtle, though, so I also added an explicit loop that'll wait for systemd to send prairiedog a `SIGTERM` and adds a second line of defense together with the `Type=oneshot`.

packages/os/os.spec
markusboehme: Fixed.

packages/os/reboot-if-required.service
markusboehme: That's a good catch, thanks! Previously, there was a brief window of time in which the boot would proceed, which practically speaking probably wouldn't be problematic, but still is super undesirable. I set `Type=oneshot` explicitly now, and the debug logs confirm the mechanism is working as intended:

```
reboot-if-required.service: starting held back, waiting for: configured.target
activate-multi-user.service: starting held back, waiting for: reboot-if-required.service
reboot-if-required.service: About to execute /usr/bin/prairiedog reboot-if-required
reboot-if-required.service: Changed dead -> start
[   13.305648] prairiedog[1163]: 19:35:00 [INFO] Boot settings changed and require a reboot to take effect. Initiating reboot...
Unit reboot.target has alias ctrl-alt-del.target.
reboot.target: Trying to enqueue job reboot.target/start/replace-irreversibly
shutdown.target: Installed new job shutdown.target/start as 472
reboot-if-required.service: Job 425 reboot-if-required.service/start finished, result=canceled
reboot-if-required.service: Installed new job reboot-if-required.service/stop as 510
activate-multi-user.service: Job 424 activate-multi-user.service/start finished, result=canceled
activate-multi-user.service: Installed new job activate-multi-user.service/stop as 452
reboot.target: Installed new job reboot.target/start as 426
```

markusboehme:
The latest force push contains these changes:

* `reboot-if-required.service` now is `Type=oneshot` as it should have been, waiting for prairiedog to issue the reboot
* Extracted a function to actually compare the boot settings for reboot-worthy changes, now coming with dedicated unit tests

I'll rebase onto `develop` and resolve merge conflicts once there's no need to diff against the previous revisions anymore. This will also unblock the automated testing for this PR.

bcressey (approved):
🎉

arnaldo2792 (approved):

markusboehme:
Thanks for the reviews! I force-pushed now to resolve the conflicts in `Cargo.toml` and `Release.toml`, unblocking the automated tests.

PR #2377, opened 2022-08-24
Add k8s credential provider support

stmcginnis:
**Issue number:**

Closes #1702

**Description of changes:**

This adds the capability to use Kubernetes image credential providers to
retrieve credentials to use when pulling images for container creation.

Initially we will only support the ecr-credential-provider, but things
are set up so we may add more providers in future updates.

---

Yet to do:

- [X] Include `ecr-credential-provider` binary
- [X] Add new config settings
- [X] Perform full testing and verification
- [X] Documentation on using this feature (basic docs, more detailed docs TBD)

---

**Testing done:**

- [X] For `ecr-credential-provider` binary: Added RPM files, built and deployed instance, verified the binary was in the expected location on the filesystem.
- [X] Addition of new settings: Build, deployed, set values by using `apiclient set`. Verified settings saved and available using `apiclient get` and viewing config settings written to filesystem.
- [X] `any-enabled` template helper: Addition of unit tests to validate various input. Created small test app to execute templating code and verify expected output.
- [x] Application of settings:
  - [X] default cred provider config file written.
  - [X] Setting values recreates file and restarts kubelet with correct parameters.
  - [X] Sonobuoy quick tests pass
  - [X] Validate use of ECR plugin for registry validation
  - [X] Tested with Kubernetes 1.20, 1.21, 1.22, and 1.23
  - [X] Tested with separate `aws-config` and `aws-credentials` as well as combined using only `aws-config`

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #2378, opened 2022-08-24
buildsys: extend `external-files` to vendor go modules

### LINKED ISSUE #2052: Go Modules in packages

jpmcb:
### Issue number:

Closes #2052
Closes #2053

### Description of changes:

```
buildsys: extend external-files to vendor go modules

Using the `bundle-*` keys on
`packag.metadata.build-package.external-files`, buildsys can vendor Go
dependencies into a new archive (leaving the prestine upstream archive
intact). Both archives may then be used to build the package in an
isolated environment.

toosl/docker-go now also retrieves the `GOSUMDB` environment variable
from the host.

Co-authored-by: John McBride <jpmmcb@amazon.com>
Co-authored-by: Samuel Karp <skarp@amazon.com>

---

hotdog: use new buildsys go module support

Co-authored-by: John McBride <jpmmcb@amazon.com>
Co-authored-by: Samuel Karp <skarp@amazon.com>

---

oci-add-hooks: use new buildsys go module support

Co-authored-by: John McBride <jpmmcb@amazon.com>
Co-authored-by: Samuel Karp <skarp@amazon.com>
```

This patch carries many of the changes from #2053 but adds recommendations from subsequent code reviews.

The `external-files` interface has been extended to now take several optional `bundle-* modifiers that enable it to consume an upstream archive, unpack it, vendor it's go module dependencies, and create a new archive with those vendored go dependencies.

Further, this enables a possible future paradigm of vendoring _many_ modules from a single `external-files`. With this PR, only "go" is supported in `bundled-modules`.

**Testing done:**

Builds 👍🏼 
```
cargo make -e PACKAGE=oci-add-hooks build-package
cargo make -e PACKAGE=hotdog build-package
```

and launching an ecs instance looks 👍🏼 
```
cargo make -e BUILDSYS_VARIANT=aws-ecs-1
```

**Terms of contribution:**

### DIFF
--- a/packages/hotdog/hotdog.spec
+++ b/packages/hotdog/hotdog.spec
@@ -10,27 +10,14 @@
 %global gitrev b85b75576adbbd7e133b54d71ebc11a28acf40db
 %global shortrev %(c=%{gitrev}; echo ${c:0:7})
 
-%global gosysrev 8c9f86f7a55f5af45a9080f6ed3ac915382d369c
-%global gosysrevshort %(c=%{gosysrev}; echo ${c:0:7})
-
-%global runtimespec 1.0.2
-
-%global goselinux 1.10.1
-
-%global libcap 1.2.63
-
 Name: %{_cross_os}hotdog
 Version: 1.0.1
 Release: 1%{?dist}
 Summary: Tool with OCI hooks to run the Log4j Hot Patch in containers
 License: Apache-2.0
 URL: https://github.com/bottlerocket-os/hotdog
-Source0: https://%{goimport}/archive/%{gorev}/%{gorepo}-%{shortrev}.tar.gz
-Source1: https://github.com/opencontainers/runtime-spec/archive/v%{runtimespec}/runtime-spec-%{runtimespec}.tar.gz
-Source2: https://github.com/golang/sys/archive/%{gosysrev}/sys-%{gosysrevshort}.tar.gz
-Source3: https://github.com/opencontainers/selinux/archive/refs/tags/v%{goselinux}.tar.gz#/go-selinux-v%{goselinux}.tar.gz
-Source4: https://git.kernel.org/pub/scm/libs/libcap/libcap.git/snapshot/libcap-cap/v%{libcap}.tar.gz#/libcap-v%{libcap}.tar.gz
-
+Source0: %{gorepo}-%{shortrev}.tar.gz
+Source1: bundled-%{gorepo}-%{shortrev}.tar.gz
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{_cross_os}log4j2-hotpatch
 
@@ -38,39 +25,19 @@ Requires: %{_cross_os}log4j2-hotpatch
 %{summary}.
 
 %prep
-%autosetup -Sgit -n %{gorepo}-%{gitrev} -p1
-%cross_go_setup %{gorepo}-%{gitrev} %{goproject} %{goimport}
-
-# We need to manage these third-party dependencies because the hotdog
-# "release" that we use doesn't include the `vendor` directory, unlike our other
-# go third party dependencies
-mkdir -p GOPATH/src/github.com/opencontainers/runtime-spec
-tar -C GOPATH/src/github.com/opencontainers/runtime-spec -xzf %{SOURCE1} --strip 1
-cp GOPATH/src/github.com/opencontainers/runtime-spec/LICENSE LICENSE.runtime-spec
-
-mkdir -p GOPATH/src/golang.org/x/sys
-tar -C GOPATH/src/golang.org/x/sys -xzf %{SOURCE2} --strip 1
-cp GOPATH/src/golang.org/x/sys/LICENSE LICENSE.golang-sys
-
-mkdir -p GOPATH/src/github.com/opencontainers/selinux
-tar -C GOPATH/src/github.com/opencontainers/selinux -xzf %{SOURCE3} --strip 1
-cp GOPATH/src/github.com/opencontainers/selinux/LICENSE LICENSE.go-selinux
-
-mkdir -p GOPATH/src/kernel.org/pub/linux/libs/security/libcap
-tar -C GOPATH/src/kernel.org/pub/linux/libs/security/libcap -xzf %{SOURCE4} --strip 2
-cp GOPATH/src/kernel.org/pub/linux/libs/security/libcap/License LICENSE.libcap
+%setup -n %{gorepo}-%{gitrev}
+%setup -T -D -n %{gorepo}-%{gitrev} -b 1
 
 %build
-%cross_go_configure %{goimport}
+%set_cross_go_flags
 
 # Set CGO_ENABLED=0 to statically link hotdog-hotpath, since it runs inside containers that
 # may not have the glibc version used to compile it
-# Set `GO111MODULE=off` to force golang to look for the dependencies in the GOPATH
-CGO_ENABLED=0 GO111MODULE=off go build -installsuffix cgo -a -ldflags "-s" -o hotdog-hotpatch ./cmd/hotdog-hotpatch
+CGO_ENABLED=0 go build ${GOFLAGS} -installsuffix cgo -a -ldflags "-s" -o hotdog-hotpatch ./cmd/hotdog-hotpatch
 
 # The oci hooks commands can be compiled as we usually compile golang packages
 for cmd in hotdog-cc-hook hotdog-poststart-hook; do
-  GO111MODULE=off go build -buildmode=pie -ldflags "${GOLDFLAGS}" -o $cmd ./cmd/$cmd
+  go build ${GOFLAGS} -buildmode=pie -ldflags "${GOLDFLAGS}" -o $cmd ./cmd/$cmd
 done
 
 %install
@@ -83,9 +50,12 @@ for cmd in hotdog-cc-hook hotdog-poststart-hook; do
   install -p -m 0755 $cmd %{buildroot}%{_cross_libexecdir}/hotdog
 done
 
+%cross_scan_attribution go-vendor vendor
+
 %files
-%license LICENSE LICENSE.runtime-spec LICENSE.golang-sys LICENSE.go-selinux LICENSE.libcap
+%license LICENSE
 %{_cross_attribution_file}
+%{_cross_attribution_vendor_dir}
 %dir %{_cross_libexecdir}/hotdog
 %dir %{_cross_datadir}/hotdog
 %{_cross_libexecdir}/hotdog/hotdog-cc-hook
--- a/tools/buildsys/src/gomod.rs
+++ b/tools/buildsys/src/gomod.rs
@@ -1,0 +1,199 @@
+/*!
+Packages using the Go programming language may have upstream tar archives that
+include only the source code of the project, but not the source code of any
+dependencies. The Go programming language promotes the use of "modules" for
+dependencies. Projects adopting modules will provide `go.mod` and `go.sum` files.
+
+This Rust module extends the functionality of `packages.metadata.build-package.external-files`
+and provides the ability to retrieve and validate dependencies
+declared using Go modules given a tar archive containing a `go.mod` and `go.sum`.
+
+The location where dependencies are retrieved from are controlled by the
+standard environment variables employed by the Go tool: `GOPROXY`, `GOSUMDB`, and
+`GOPRIVATE`. These variables are automatically retrieved from the host environment
+when the docker-go script is invoked.
+
+ */
+
+pub(crate) mod error;
+use error::Result;
+
+use super::manifest;
+use duct::cmd;
+use snafu::{ensure, OptionExt, ResultExt};
+use std::io::Write;
+use std::os::unix::fs::PermissionsExt;
+use std::path::{Path, PathBuf};
+use std::{env, fs};
+
+pub(crate) struct GoMod;
+
+const GO_MOD_DOCKER_SCRIPT_NAME: &str = "docker-go-script.sh";
+
+// The following bash template script is intended to be run within a container
+// using the docker-go tool found in this codebase under `tools/docker-go`.
+//
+// This script inspects the top level directory found in the package upstream
+// archive and uses that as the default Go module path if no explicit module
+// path was provided. It will then untar the archive, vendor the Go
+// dependencies, create a new archive using the {module-path}/vendor directory
+// and name it the output path provided. If no output path was given, it
+// defaults to "bundled-{package-file-name}". Finally, it cleans up by removing
+// the untar'd source code. The upstream archive remains intact and both tar
+// files can then be used during packaging.
+//
+// This script exists as an in memory template string literal and is populated
+// into a temporary file in the package directory itself to enable buildsys to
+// be as portable as possible and have no dependecy on runtime paths. Since
+// buildsys is executed from the context of many different package directories,
+// managing a temporary file via this Rust module prevents having to aquire the
+// path of some static script file on the host system.
+const GO_MOD_SCRIPT_TMPL: &str = r###"#!/bin/bash
+
+set -e
+
+toplevel=$(tar tf __LOCAL_FILE_NAME__ | head -1)
+if [ -z __MOD_DIR__ ] ; then
+    targetdir="${toplevel}"
+else
+    targetdir="__MOD_DIR__"
+fi
+
+tar xf __LOCAL_FILE_NAME__
+
+pushd "${targetdir}"
+    go list -mod=readonly ./... >/dev/null && go mod vendor
+popd
+
+tar czf __OUTPUT__ "${targetdir}"/vendor
+rm -rf "${targetdir}"
+"###;
+
+impl GoMod {
+    pub(crate) fn vendor(
+        root_dir: &Path,
+        package_dir: &Path,
+        external_file: &manifest::ExternalFile,
+    ) -> Result<()> {
+        let url_file_name = extract_file_name(&external_file.url)?;
+        let local_file_name = &external_file.path.as_ref().unwrap_or(&url_file_name);
+        ensure!(
+            local_file_name.components().count() == 1,
+            error::InputFileSnafu
+        );
+
+        let full_path = package_dir.join(local_file_name);
+        ensure!(
+            full_path.is_file(),
+            error::InputFileBadSnafu { path: full_path }
+        );
+
+        // If a module directory was not provided, set as an empty path.
+        // By default, without a provided module directory, tar will be passed
+        // the first directory found in the archives as the top level Go module
+        let default_empty_path = PathBuf::from("");
+        let mod_dir = external_file
+            .bundle_root_path
+            .as_ref()
+            .unwrap_or(&default_empty_path);
+
+        // Use a default "bundle-{name-of-file}" if no output path was provided
+        let default_output_path =
+            PathBuf::from(format!("bundled-{}", local_file_name.to_string_lossy()));
+        let output_path_arg = external_file
+            .bundle_output_path
+            .as_ref()
+            .unwrap_or(&default_output_path);
+        println!(
+            "cargo:rerun-if-changed={}",
+            output_path_arg.to_string_lossy()
+        );
+
+        // Our SDK and toolchain are picked by the external `cargo make` invocation.
+        let sdk = env::var("BUILDSYS_SDK_IMAGE").context(error::EnvironmentSnafu {
+            var: "BUILDSYS_SDK_IMAGE",
+        })?;
+
+        let args = DockerGoArgs {
+            module_path: package_dir,
+            sdk_image: sdk,
+            go_mod_cache: &root_dir.join(".gomodcache"),
+            command: format!("./{}", GO_MOD_DOCKER_SCRIPT_NAME),
+        };
+
+        // Create and/or write the temporary script file to the package directory
+        // using the script template string and placeholder variables
+        let script_contents = GO_MOD_SCRIPT_TMPL
+            .replace("__LOCAL_FILE_NAME__", &local_file_name.to_string_lossy())
+            .replace("__MOD_DIR__", &mod_dir.to_string_lossy())
+            .replace("__OUTPUT__", &default_output_path.to_string_lossy());
+        let script_path = format!(
+            "{}/{}",
+            package_dir.to_string_lossy(),
+            GO_MOD_DOCKER_SCRIPT_NAME
+        );
+        {
+            let mut script_file = fs::File::create(&script_path).unwrap();
+            fs::set_permissions(&script_path, fs::Permissions::from_mode(0o777)).unwrap();
+            script_file.write_all(script_contents.as_bytes()).unwrap();
+        }
+
+        let res = docker_go(root_dir, &args);
+        fs::remove_file(script_path).unwrap();
+        res
+    }
+}
+
+fn extract_file_name(url: &str) -> Result<PathBuf> {
+    let parsed = reqwest::Url::parse(url).context(error::InputUrlSnafu { url })?;
+    let name = parsed
+        .path_segments()
+        .context(error::InputFileBadSnafu { path: url })?
+        .last()
+        .context(error::InputFileBadSnafu { path: url })?;
+    Ok(name.into())
+}
+
+struct DockerGoArgs<'a> {
+    module_path: &'a Path,
+    sdk_image: String,
+    go_mod_cache: &'a Path,
+    command: String,
+}
+
+/// Run `docker-go` with the specified arguments.
+fn docker_go(root_dir: &Path, dg_args: &DockerGoArgs) -> Result<()> {
+    let args = vec![
+        "--module-path",
+        dg_args
+            .module_path
+            .to_str()
+            .context(error::InputFileSnafu)?,
+        "--sdk-image",
+        &dg_args.sdk_image,
+        "--go-mod-cache",
+        dg_args
+            .go_mod_cache
+            .to_str()
+            .context(error::InputFileSnafu)?,
+        "--command",
+        &dg_args.command,
+    ];
+    let arg_string = args.join(" ");
+    let program = root_dir.join("tools/docker-go");
+    println!("program: {}", program.to_string_lossy());
+    let output = cmd(program, args)
+        .stderr_to_stdout()
+        .stdout_capture()
+        .unchecked()
+        .run()
+        .context(error::CommandStartSnafu)?;
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    println!("{}", &stdout);
+    ensure!(
+        output.status.success(),
+        error::DockerExecutionSnafu { args: arg_string }
+    );
+    Ok(())
+}
--- a/tools/buildsys/src/gomod/error.rs
+++ b/tools/buildsys/src/gomod/error.rs
@@ -1,0 +1,33 @@
+use std::path::PathBuf;
+
+use snafu::Snafu;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub(super)))]
+pub(crate) enum Error {
+    #[snafu(display("Failed to start command: {}", source))]
+    CommandStart { source: std::io::Error },
+
+    #[snafu(display("Failed to execute docker-go script. 'args: {}'", args))]
+    DockerExecution { args: String },
+
+    #[snafu(display("Input url is required"))]
+    InputFile,
+
+    #[snafu(display("Input file {} must be a file", path.display()))]
+    InputFileBad { path: PathBuf },
+
+    #[snafu(display("Bad file url '{}': {}", url, source))]
+    InputUrl {
+        url: String,
+        source: url::ParseError,
+    },
+
+    #[snafu(display("Missing environment variable '{}'", var))]
+    Environment {
+        var: String,
+        source: std::env::VarError,
+    },
+}
+
+pub(super) type Result<T> = std::result::Result<T, Error>;
--- a/tools/buildsys/src/main.rs
+++ b/tools/buildsys/src/main.rs
@@ -10,10 +10,13 @@ The implementation is closely tied to the top-level Dockerfile.
 */
 mod builder;
 mod cache;
+mod gomod;
 mod manifest;
 mod project;
 mod spec;
 
+use crate::gomod::GoMod;
+use crate::manifest::BundleModule;
 use builder::{PackageBuilder, VariantBuilder};
 use cache::LookasideCache;
 use manifest::{ManifestInfo, SupportedArch};
@@ -41,6 +44,10 @@ mod error {
 
         ExternalFileFetch {
             source: super::cache::error::Error,
+        },
+
+        GoMod {
+            source: super::gomod::error::Error,
         },
 
         ProjectCrawl {
@@ -143,6 +150,19 @@ fn build_package() -> Result<()> {
 
     if let Some(files) = manifest.external_files() {
         LookasideCache::fetch(files).context(error::ExternalFileFetchSnafu)?;
+        for f in files {
+            if f.bundle_modules.is_none() {
+                continue;
+            }
+
+            for b in f.bundle_modules.as_ref().unwrap() {
+                match b {
+                    BundleModule::Go => {
+                        GoMod::vendor(&root_dir, &manifest_dir, f).context(error::GoModSnafu)?
+                    }
+                }
+            }
+        }
     }
 
     if let Some(groups) = manifest.source_groups() {
--- a/tools/buildsys/src/manifest.rs
+++ b/tools/buildsys/src/manifest.rs
@@ -33,6 +33,44 @@ url = "https://bar"
 sha512 = "123456"
 ```
 
+The `bundle-*` keys on `external-files` are a group of optional modifiers
+and are used to untar an upstream external file archive, vendor any dependent
+code, and produce an additional archive with those dependencies.
+Only `bundle-modules` is required when bundling an archive's dependences.
+
+`bundle-modules` is a list of module "paradigms" the external-file should
+be vendored through. For example, if a project contains a `go.mod` and `go.sum`
+file, adding "go" to the list will vendor the dependencies through go modules.
+Currently, only "go" is supported.
+
+`bundle-root-path` is an optional argument that provides the filepath
+within the archive that contains the module. By default, the first top level
+directory in the archive is used. So, for example, given a Go project that has
+the necessary `go.mod` and `go.sum` files in the archive located at the
+filepath `a/b/c`, this `bundle-root-path` value should be "a/b/c". Or, given an
+archive with a single directory that contains a Go project that has `go.mod`
+and `go.sum` files located in that top level directory, this option may be
+omitted since the single top-level directory will authomatically be used.
+
+`bundle-output-path` is an optional argument that provides the desired path of
+the output archive. By default, this will use the name of the existing archive,
+but pre-pended with "bundled-". For example, if "my-unique-archive-name.tar.gz"
+is entered as the value for `bundle-output-path`, then the output directory
+will be named `my-unique-archive-name.tar.gz`. Or, by default, given the name
+of some upstream archive is "my-package.tar.gz", the output archive would be
+named `bundled-my-package.tar.gz`. This output path may then be referenced
+within an RPM spec or when creating a package in order to access the vendored
+upstream dependencies during build time.
+```
+[[package.metadata.build-package.external-files]]
+path = "foo"
+url = "https://foo"
+sha512 = "abcdef"
+bundle-modules = [ "go" ]
+bundle-root-path = "path/to/module"
+bundle-output-path = "path/to/output.tar.gz"
+```
+
 `package-name` lets you override the package name in Cargo.toml; this is useful
 if you have a package with "." in its name, for example, which Cargo doesn't
 allow.  This means the directory name and spec file name can use your preferred
@@ -347,11 +385,20 @@ impl fmt::Display for GrubFeature {
 }
 
 #[derive(Deserialize, Debug)]
+#[serde(rename_all = "lowercase")]
+pub(crate) enum BundleModule {
+    Go,
+}
+
+#[derive(Deserialize, Debug)]
 #[serde(rename_all = "kebab-case")]
 pub(crate) struct ExternalFile {
     pub(crate) path: Option<PathBuf>,
     pub(crate) sha512: String,
     pub(crate) url: String,
+    pub(crate) bundle_modules: Option<Vec<BundleModule>>,
+    pub(crate) bundle_root_path: Option<PathBuf>,
+    pub(crate) bundle_output_path: Option<PathBuf>,
 }
 
 impl fmt::Display for SupportedArch {
--- a/tools/docker-go
+++ b/tools/docker-go
@@ -56,7 +56,7 @@ parse_args "${@}"
 
 # Pass through relevant Go variables, from the config or environment.
 go_env=( )
-for i in GOPROXY GONOPROXY GOPRIVATE ; do
+for i in GOPROXY GONOPROXY GOPRIVATE GOSUMDB ; do
   if command -v go >/dev/null 2>&1 ; then
     govar="$(go env ${i})"
     if [ -n "${govar}" ] ; then
--- a/packages/hotdog/Cargo.toml
+++ b/packages/hotdog/Cargo.toml
@@ -11,24 +11,7 @@ path = "pkg.rs"
 [[package.metadata.build-package.external-files]]
 url = "https://github.com/bottlerocket-os/hotdog/archive/b85b75576adbbd7e133b54d71ebc11a28acf40db/hotdog-b85b755.tar.gz"
 sha512 = "9b2d5cb0e25d774d11dd6eb577e07af85f36fcd6e816b9df88d7ca1da273695f15ce6831026d28e68355512a07d0ac673b5ce9771d969f1c5ca4f14bc631deb8"
-
-[[package.metadata.build-package.external-files]]
-url = "https://github.com/opencontainers/runtime-spec/archive/v1.0.2/runtime-spec-1.0.2.tar.gz"
-sha512 = "96676b702d02409d33a5c81886c4db4bf45283c628933c6f0f4c2ed0d7cc44fafe95249151d7dc2d1fc5225944d172cdb45fc2f2f5f9bb87531e93421933b664"
-
-[[package.metadata.build-package.external-files]]
-url = "https://github.com/golang/sys/archive/8c9f86f7a55f5af45a9080f6ed3ac915382d369c/sys-8c9f86f.tar.gz"
-sha512 = "054dfba40563e2537dffcd8464debf44bcb71d5790603f21a324ec940d7018d8161d50d535471f5ffeada9647ef390d599d567dd0515eeaa677103c60e502b40"
-
-[[package.metadata.build-package.external-files]]
-path = "go-selinux-v1.10.1.tar.gz"
-url = "https://github.com/opencontainers/selinux/archive/refs/tags/v1.10.1.tar.gz"
-sha512 = "f79af22c28ac14b3ca93c1c60fc6f986ec6b22c0f2d77fc4192b95e0a497798acfe9be48f4e162e0a31c7e4c3c78470bdee6faab1c06f8688b51a0c20331b77a"
-
-[[package.metadata.build-package.external-files]]
-path = "libcap-v1.2.63.tar.gz"
-url = "https://git.kernel.org/pub/scm/libs/libcap/libcap.git/snapshot/libcap-cap/v1.2.63.tar.gz"
-sha512 = "f88f85ce5849c0c6a519c3f269a5b221a66b98d30abfa15a69a760ba85d84b5cf89f41ad1912fc0643a5533b44f0b2d09597229b61f4d08473ad6124e40bc864"
+bundle-modules = [ "go" ]
 
 [build-dependencies]
 glibc = { path = "../glibc" }
--- a/packages/oci-add-hooks/Cargo.toml
+++ b/packages/oci-add-hooks/Cargo.toml
@@ -11,14 +11,7 @@ path = "pkg.rs"
 [[package.metadata.build-package.external-files]]
 url = "https://github.com/awslabs/oci-add-hooks/archive/ef29fe312d2e1858d5eb28ab0abe0cbee298a165/oci-add-hooks-ef29fe3.tar.gz"
 sha512 = "018b561f838172e768a70acdeb2c27939f931391ced019a23c5193eee6b8970bc02a3e5fa05917010ca2064d1876649ef139d7657700c42a3ddd6e2c174f27dc"
-
-[[package.metadata.build-package.external-files]]
-url = "https://github.com/bitly/go-simplejson/archive/v0.5.0/go-simplejson-0.5.0.tar.gz"
-sha512 = "39c0d85d6ee06a8a795c1e344f0669f5ae8371d1122f09a1b13e5ff7629dd7faf633f9fcb449e19aadab9ad3e42e93143205781a822a29f27758872cf7e09e18"
-
-[[package.metadata.build-package.external-files]]
-url = "https://github.com/joeshaw/json-lossless/archive/e0cd1ca6349bf167e33d44f28c14c728a277205f/json-lossless-e0cd1ca.tar.gz"
-sha512 = "b9eb6170f662a396370ae1e170d89e15efc0a96fee6046fbd749c7a65f09f808e08bc2cf91962db65fd86a2aac4dddf428412b568fe1d03a77a7de22ad0690aa"
+bundle-modules = [ "go" ]
 
 [build-dependencies]
 glibc = { path = "../glibc" }
--- a/packages/oci-add-hooks/oci-add-hooks.spec
+++ b/packages/oci-add-hooks/oci-add-hooks.spec
@@ -4,9 +4,6 @@
 
 %global gitrev ef29fe312d2e1858d5eb28ab0abe0cbee298a165
 %global shortrev %(c=%{gitrev}; echo ${c:0:7})
-%global gosimplejson 0.5.0
-%global jsonlosslessrev e0cd1ca6349bf167e33d44f28c14c728a277205f
-%global jsonlosslessshort %(c=%{jsonlosslessrev}; echo ${c:0:7})
 
 Name: %{_cross_os}oci-add-hooks
 Version: 1.0.0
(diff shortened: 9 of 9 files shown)


jpmcb:
~Whoops! A few files need `rustfmt` - please hold!~

Force pushed to address failing GitHub actions needing `rustfmt`-ing

jpmcb:
Force pushed to address `cargo make check-lints` and `cargo make check-fmt`

jpmcb:
Force pushed to address not having to unwrap here:

```diff
- dg_args
-         .go_mod_cache 
-         .to_str()
-         .context(error::InputFileSnafu)
-         .unwrap()
+ dg_args
+        .go_mod_cache
+        .to_str()
+        .context(error::InputFileSnafu)?
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,176 @@
...
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    println!("{}", &stdout);
+    output.status.success().then(|| stdout.to_string())
+}
+
+/// Retrieve a BUILDSYS_* variable that we expect to be set in the environment,
+/// and ensure that we track it for changes, since it will directly affect the
+/// output.
+fn getenv(var: &str) -> Result<String> {
+    println!("cargo:rerun-if-env-changed={}", var);
```
arnaldo2792: I think we are not supposed to do this anymore and instead use `bottlerocket-variant`, and probably this function? https://github.com/bottlerocket-os/bottlerocket/blob/037d0e2798a7e8d4a12d0fb35d3c266db3d74f25/sources/bottlerocket-variant/src/lib.rs#L178

I'll let others comment on this since I'm not an expert about when the library linked should be used.

tools/buildsys/src/gomod/error.rs
```diff
@@ -0,0 +1,42 @@
...
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub(super)))]
+pub(crate) enum Error {
+    #[snafu(display("Failed to start command: {}", source))]
+    CommandStart { source: std::io::Error },
+
+    #[snafu(display("Failed to execute command: 'docker {:?}'", args))]
+    DockerExecution { args: String },
+
+    #[snafu(display("input url is required"))]
```
arnaldo2792: Nit: `s/input/Input`, same for error below

tools/buildsys/src/gomod/error.rs
```diff
@@ -0,0 +1,42 @@
...
+    },
+
+    #[snafu(display("mod-dir is required"))]
+    ModDir,
+
+    #[snafu(display("output-dir is required"))]
+    OutputDir,
+
+    #[snafu(display("output_dir must not exist or be a directory"))]
+    OutputDirBad,
```
arnaldo2792: This error isn't used, you could remove it.

tools/buildsys/src/manifest.rs
```diff
@@ -33,6 +33,35 @@ url = "https://bar"
 sha512 = "123456"
 ```
 
+The `bundle-*` keys on `external-files` are a group of optional modifiers
+and are used to untar an upstream external file achive,
+vendor any dependent code, and bundle it back up to be consumed.
+All three options are required when bundling an archieves dependences.
```
arnaldo2792: Nit: missing `'` in `archives`?

tools/buildsys/src/manifest.rs
```diff
@@ -33,6 +33,35 @@ url = "https://bar"
 sha512 = "123456"
 ```
 
+The `bundle-*` keys on `external-files` are a group of optional modifiers
+and are used to untar an upstream external file achive,
```
stmcginnis: s/achive/archive/

tools/buildsys/src/manifest.rs
stmcginnis: s/archieves dependences/archive's dependencies/

stmcginnis (commented):
Since it looks like you will need to make other updates, added a couple small nits to the list.

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,176 @@
...
+            module_path: package_dir,
+            sdk_image: sdk,
+            go_mod_cache: &root_dir.join(".gomodcache"),
+            command: format!(
+                "tar xf {input} &&
+                pushd {moddir} &&
+                export GOPROXY={goproxy} &&
+                export GOSUMDB={gosumdb} &&
+                export GOPRIVATE={goprivate} &&
+                go list -mod=readonly ./... >/dev/null && go mod vendor &&
```
stmcginnis: This `go mod vendor` seems to be causing problems for me. For projects that just use `go.mod`, this causes a `vendor` directory to be created and everything to be downloaded (again) into that new directory. That seems to be the cause of these errors:

```
  #19 0.532 go: inconsistent vendoring in /home/builder/rpmbuild/BUILD/cloud-provider-aws-1.22.3/GOPATH/src/github.com/kubernetes/cloud-provider-aws:
  #19 0.532     github.com/aws/aws-sdk-go@v1.43.32: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
  #19 0.532     github.com/golang/mock@v1.4.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
  #19 0.532     github.com/google/go-cmp@v0.5.5: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
  #19 0.532     github.com/stretchr/testify@v1.7.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
...
```

Can we add a check somewhere that first checks if there is a `vendor` directory, and based on that decide whether to run `go mod vendor` or not?

stmcginnis: Maybe this isn't the actual source of the failure. I tried removing this locally, but I still get the error. So my error may be coming from elsewhere, or it may be left over cached files in my environment. If it's something left behind, I haven't been able to figure out yet where that would be or how to clean it up though.

But somewhere a `go mod vendor` is happening and seems to be the source of my problems.

tools/buildsys/src/gomod.rs
jpmcb: Good question: This was a chunk of code from Sam's initial PR and I assumed would force a rerun if any of the go mod control env variables changed. If there's a better way to do this, let me know!

tools/buildsys/src/gomod/error.rs
jpmcb: Good catch - thanks!

tools/buildsys/src/gomod.rs
jpmcb: > Can we add a check somewhere that first checks if there is a vendor directory, and based on that decide whether to run go mod vendor or not?

If there is already an existing `vendor` directory, we shouldn't be calling out to `go mod` again. This would indicate that the upstream project has declared their dependencies, has them part of their source control, and the project is ready to be build via `go build`. I _believe_ this is already the current paradigm as evident by packages like `runc`: 

https://github.com/bottlerocket-os/bottlerocket/blob/211e3db67a7c0ac758c6eb135213285513c92f82/packages/runc/runc.spec#L34-L38

and their vendoring inside the project

https://github.com/opencontainers/runc/tree/main/vendor

jpmcb:
Force pushed to address nits / spelling mistakes ✏️

tools/buildsys/src/gomod.rs
stmcginnis: > If there is already an existing vendor directory, we shouldn't be calling out to go mod again

No, my point was if there ISN'T an existing `vendor` directory, then don't try to run `go mod vendor` as that's not the type of dependency management that project uses. But based on our offline conversation, sounds like this may still be wanted to be a way to "cache" those dependencies locally to allow things to be built in a network isolated container.

So maybe this is just a red herring and there's some other issue with the package I'm working on. Still trying to sort that out.

jpmcb: > may still be wanted to be a way to "cache" those dependencies locally to allow things to be built in a network isolated container.

Right, this will enable packages being build in our network isolated `buildsys` container to not have to gather the dependencies manually (like hotdog _was_ doing) but rather, build a "bundled" tarball with the vendored dependencies to then be consumed in the isolated environment

packages/hotdog/hotdog.spec
```diff
@@ -10,67 +10,32 @@
...
 Summary: Tool with OCI hooks to run the Log4j Hot Patch in containers
 License: Apache-2.0
 URL: https://github.com/bottlerocket-os/hotdog
-Source0: https://%{goimport}/archive/%{gorev}/%{gorepo}-%{shortrev}.tar.gz
-Source1: https://github.com/opencontainers/runtime-spec/archive/v%{runtimespec}/runtime-spec-%{runtimespec}.tar.gz
-Source2: https://github.com/golang/sys/archive/%{gosysrev}/sys-%{gosysrevshort}.tar.gz
-Source3: https://github.com/opencontainers/selinux/archive/refs/tags/v%{goselinux}.tar.gz#/go-selinux-v%{goselinux}.tar.gz
-Source4: https://git.kernel.org/pub/scm/libs/libcap/libcap.git/snapshot/libcap-cap/v%{libcap}.tar.gz#/libcap-v%{libcap}.tar.gz
-
+Source0: %{gorepo}-%{shortrev}-bundled.tar.gz
```
bcressey: I'd like to have two source files in this case:
* Source0 - pristine upstream archive
* Source1 - just the vendored modules

tools/buildsys/src/manifest.rs
```diff
@@ -33,6 +33,35 @@ url = "https://bar"
...
+
+`bundle-modules` is a list of module "paradigms" the external-file should
+be vendored through. For example, if a project contains a go.mod and go.sum file,
+"go" may be entered to vendor the dependencies.
+Currently, only "go" is supported.
+
+`bundle-path` is the filepath _within_ the archive that contains the module.
+So, for example, given a go project that has the necessary go.mod and go.sum
+files within the archive located at the filepath a/b/c,
+this `bundle-path` value should be `a/b/c`.
```
bcressey: Can we default this to the first sub-directory in the archive, assuming the root level is otherwise empty?

tools/buildsys/src/manifest.rs
```diff
@@ -33,6 +33,35 @@ url = "https://bar"
...
+
+`bundle-output-path` is the path of the output tar archive.
+This name may be referenced within a spec, when creating a package, etc.
+For example, `my-package-bundled.tar.gz` may be entered as an output path
+indicating the package is "bundled" and has been archived.
```
bcressey: A default value using a convention where `-bundled` is appended to the last part of the external filename before `.tar.*` would be nice.

tools/buildsys/src/manifest.rs
```diff
@@ -352,6 +381,9 @@ pub(crate) struct ExternalFile {
     pub(crate) path: Option<PathBuf>,
     pub(crate) sha512: String,
     pub(crate) url: String,
+    pub(crate) bundle_modules: Option<Vec<String>>,
```
bcressey: We can enforce stronger typing here using an enum, like `SupportedArch` and `GrubFeature` do.

(Then use that enum with a match statement in `main.rs` to ensure that all known values are handled.)

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,176 @@
...
+            goproxy.pop();
+        }
+        let mut gosumdb = go_env("GOSUMDB").unwrap_or_else(|| "".to_string());
+        if gosumdb.ends_with('\n') {
+            gosumdb.pop();
+        }
+        let mut goprivate = go_env("GOPRIVATE").unwrap_or_else(|| "".to_string());
+        if goprivate.ends_with('\n') {
+            goprivate.pop();
+        }
```
bcressey: nit: maybe factor this into a helper function that takes a list of possible environment variables

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,176 @@
...
+        .unchecked()
+        .run()
+    {
+        Ok(v) => v,
+        Err(_) => return None,
+    };
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    println!("{}", &stdout);
+    output.status.success().then(|| stdout.to_string())
+}
```
bcressey: We don't want to assume that a `go` install exists on the host or works as expected.

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,176 @@
...
+
+/// Retrieve a BUILDSYS_* variable that we expect to be set in the environment,
+/// and ensure that we track it for changes, since it will directly affect the
+/// output.
+fn getenv(var: &str) -> Result<String> {
+    println!("cargo:rerun-if-env-changed={}", var);
+    env::var(var).context(error::EnvironmentSnafu { var })
+}
```
bcressey: `BUILDSYS_SDK_IMAGE` will be tracked by the usage of `getenv` in `builder.rs`, so we don't need it here.

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,176 @@
...
+        ensure!(path_arg.components().count() == 1, error::InputFileSnafu);
+
+        let path = package_dir.join(path_arg);
+        ensure!(path.is_file(), error::InputFileBadSnafu { path });
+        let mod_dir = gomod.bundle_path.as_ref().context(error::ModDirSnafu)?;
+        let output_path_arg = gomod
+            .bundle_output_path
+            .as_ref()
+            .context(error::OutputDirSnafu)?;
+        let output_path = package_dir.join(output_path_arg);
```
bcressey: Somewhere, maybe right here, we need to emit `rerun-if-changed` for the output bundle archive.

tools/buildsys/src/gomod.rs
jpmcb: Gotcha - what would good indicators that buildsys shouldn't proceed with vendoring the go module on the host?

My first thought is to check:
- is the `go` tool in the path?
- is the output of `go version` greater than 1.16?

If either of the answer to those is "no", we fail and exit early.

packages/hotdog/hotdog.spec
jpmcb: > I'd like to have two source files in this case:

Hmmm, what is the benefit of having a separate untouched upstream archive alongside it's vendored modules?

In this PRs current state, the archive is downloaded, unpacked, modified via `go mod vendor` (which populates it's dependencies into a `vendor` directory) and then tar'd back up into a file named by `bundle-output-path`. This gives a simple, **single file** interface to then use in the build spec. It also emulates what a Go project _would_ look like if it had vendored it's modules into the source control.

But, I suppose with separate archives for the upstream source and bundled modules, we could have greater control over what makes it into end result. I.e., if we needed to patch an upstream's dependency at build time, we'd want to do that out of band from the actual upstream source code which would be painful if it's all already packaged into one archive.  

I guess this is really about which piece of the package building process should we combine the vendored modules and the upstream source: in the lookaside cahce's archive? After `buildsys` has downloaded the archive into a single tar file that gets passed to the spec? Within the spec itself during build time?

Seems like for the most control, we'd want the last option, at buildtime?

tools/buildsys/src/gomod.rs
bcressey: Hmm, is this redundant with the checks in `docker-go`? That script already has the logic to use `go` if we have it, otherwise read the environment variable directly.

packages/hotdog/hotdog.spec
bcressey: > Hmmm, what is the benefit of having a separate untouched upstream archive alongside it's vendored modules?

It feels surprising to have `Cargo.toml` list one file with its hash, and have the spec file switch over to using a different file with an unknown hash without any warning. If you don't know about the magic that `buildsys` is doing, you might think the hash isn't actually being used, which would suggest a security problem.

By having the bundle archive just cover the delta - e.g. the vendor directory - you don't run the risk of something happening to one of the source files that isn't expected to change, which would then be difficult to track down.

> Seems like for the most control, we'd want the last option, at buildtime?

This seems right to me, even if it means a second `%setup` invocation to extract the second archive.

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        // Return early if the output path exists and is a file, assuming it's already been built
+        if output_path.exists() && output_path.is_file() {
+            return Ok(());
+        }
```
bcressey: The discussion on single vs separate archives made me think of a potential failure mode here, which is that if the input external file changes in some way but keeps the same filename, we need to detect that and recreate the dependency bundle, rather than reusing the one we have.

If this optimization doesn't save very much time, it might be simpler to just always recreating the bundle file. I'm assuming that the `go mod vendor` step is pretty quick if all the modules are already downloaded into the cache.

stmcginnis: Does adding `rerun-if-changed` help here? Like rust's `Cargo.lock` file, the `go.mod` and `go.sum` files "lock" dependencies to a specific version. So there shouldn't be any change in what `go mod vendor` would download unless there are changes to those files.

At least I'm pretty sure. I've never mixed `go.mod` module management with vendoring in a project before, but this is the way I understand it to work.

jpmcb: > it might be simpler to just always recreating the bundle file ... the go mod vendor step is pretty quick if all the modules are already downloaded into the cache.

I think recreating the dependency bundle is the way to go for now and should be almost instant given the modcahce, but I'll also do some experimenting to see if it slows down compared to my testing yesterday.

bcressey: > Does adding `rerun-if-changed` help here? Like rust's Cargo.lock file, the go.mod and go.sum files "lock" dependencies to a specific version.

It sort of helps, in the sense that any change to the file would trigger a rebuild - but if we're in `buildsys` at all then that means a (re)build has been triggered, and we don't know exactly why, just that `cargo` decided it was necessary. Since we're doing our own caching here, we'd need to detect whether the original input file changed - something else might've changed instead - and invalidate the bundled archive if it did.

That could be done by also saving a hash file of the original file somewhere, but then we have to manage that file too. Or we could prepend the hash of the original file to the bundled archive filename, so we could always check it, but then we'd want to strip off that prefix before passing into the package build environment since it wouldn't be predictable.

Not insurmountable problems but if the added benefit of the secondary cache is minor, it might not be worth the complexity.

stmcginnis:
I've found some issues with using `go mod vendor` for pulling in dependencies for projects that do not using vendoring for dependency management. Not sure if there's a better way for us to handle it than current plans, but thought I should add my findings here so others are aware and we can possibly think about a better approach.

## Vendoring approach

I found these issues as I'm trying to build the `ecr-credential-provider` binary out of https://github.com/kubernetes/cloud-provider-aws. If I try to build it isolated (airgapped, offline, etc), I first run `go mod vendor` to create and populate a vendor directory. But then when I try to build the project, it appears the vendor directory is too deep or something, so I get this error:

```shell
$ make ecr-credential-provider
GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GOPROXY=direct go build \
	-trimpath \
	-ldflags="-w -s -X k8s.io/component-base/version.gitVersion=v1.22.3-dirty" \
	-o=ecr-credential-provider \
	cmd/ecr-credential-provider/*.go
make: /bin/bash: Argument list too long
make: *** [Makefile:38: ecr-credential-provider] Error 127
```

## `go mod download` Approach

This led me to explore a more module-centric solution to capturing all deps to be able to build offline.

I first tried to capture dependencies into a separate directory so I could use that to copy into the offline build.

```shell
GOCACHE="$(mktemp -d)"
GOPATH="$(mktemp -d)"

GOPATH="${GOPATH}" go mod download
cp -rp "${GOPATH}"/pkg/mod/cache/download/* "${GOCACHE}/"

docker run --rm --network none \
    -v "${GOCACHE}":/gocache \
    -v ${PWD}:/build -w /build \
    -e GOCACHE=/cache \
    -e GOPROXY=file:///gocache golang:1.18 \
    make ecr-credential-provider
```

This had a couple issues. For some reason, more often than not the `go mod download` would exit with failure being unable to find one of the dependencies:

```shell
go: github.com/matttproud/golang_protobuf_extensions@v1.0.2-0.20181231171920-c182affec369 requires
	golang.org/x/sync@v0.0.0-20181221193216-37e7f081c4d4: invalid version: unknown revision 37e7f081c4d4
```

If I reran the command a few times, eventually it would succeed and get everything downloaded.

But the second issue then was that `go mod download` doesn't go deep. The downloaded dependencies wouldn't actually have everything needed to build the project, so running in an offline environment would error out stating there were things missing:

```
...
cmd/ecr-credential-provider/main.go:32:2: github.com/aws/aws-sdk-go@v1.43.32: reading file:///gocache/github.com/aws/aws-sdk-go/@v/v1.43.32.zip: no such file or directory
cmd/ecr-credential-provider/main.go:34:2: k8s.io/apimachinery@v0.22.8: reading file:///gocache/k8s.io/apimachinery/@v/v0.22.8.zip: no such file or directory
cmd/ecr-credential-provider/plugin.go:28:2: k8s.io/apimachinery@v0.22.8: reading file:///gocache/k8s.io/apimachinery/@v/v0.22.8.zip: no such file or directory
...
```

## Double Build Approach

I was able to get everything downloaded by first building the project while still connected:

```
GOPATH="${GOPATH}" GOPROXY="file:///${GOCACHE}" make ecr-credential-provider
```

then building again isolated:

```
rm ecr-credential-provider

docker run --rm --network none \
    -v "${GOCACHE}":/gocache \
    -v ${PWD}:/build -w /build \
    -e GOCACHE=/cache \
    -e GOPROXY=file:///gocache golang:1.18 \
    make ecr-credential-provider
```

It is a small project that doesn't take long to build, but that seems a little redundant. If we did end up having this codified in the way buildsys handles go projects, this definitely would not be a good thing to do for all go dependencies.

## "Thorough" Approach

What ended up working the best was to "force" `go mod download` into actually following all dependencies. This is what I ended up with that would consistently work:

```
GOCACHE="$(mktemp -d)"
GOPATH="$(mktemp -d)"

go mod graph | \
    awk '{print $2}' | \
    sort -u | \
    GOPATH="${GOPATH}" xargs go mod download
cp -rp "${GOPATH}"/pkg/mod/cache/download/* "${GOCACHE}/"

docker run --rm --network none \
    -v "${GOCACHE}":/gocache \
    -v ${PWD}:/build -w /build \
    -e GOCACHE=/cache \
    -e GOPROXY=file:///gocache golang:1.18 \
    make ecr-credential-provider
```

The downside is the `go mod graph` call does take some time since it is actually thorough. I think we could mitigate that by having a local (non-isolated) GOCACHE that buildsys could use while downloading the dependencies. Then we shouldn't need to download the same packages repeatedly.

jpmcb:
Added another commit to address latest review comments:

Now, when `buildsys` vendors the go dependencies, it places them into a separate archive. The pristine upstream archive can then be used in conjunction with the dependency archive to create the package.

Adds sane defaults for several of the bundling options.

---

I will plan to squash this into a co-authored commit upon approval, but this was a bit more of a change that seemed to warrant it's own intermittent commit

packages/hotdog/hotdog.spec
jpmcb: > I'd like to have two source files in this case

👍🏼 Two archives are now used as `Source0` (the upstream pristine archive) and `Source1` which has the vendored dependencies

tools/buildsys/src/manifest.rs
jpmcb: 👍🏼 Now defaults to the first top level directory in the archive if `bundle-path` is not provided

tools/buildsys/src/manifest.rs
jpmcb: 👍🏼 I punted on this _abit_ since I was having trouble with Rust string manipulation. It now defaults to `bundled-{name-of-tar}`. So, for example, in the `hotdog` package, the default would be `bundled-hotdog-b85b755.tar.gz` if `bundled-output-path` is not provided.

tools/buildsys/src/manifest.rs
jpmcb: 👍🏼  Added enum for this struct field

tools/buildsys/src/gomod.rs
jpmcb: 👍🏼 This code block was redundant since `docker-go` gets these variables from the environment so this block was removed.

tools/buildsys/src/gomod.rs
jpmcb: 👍🏼 Removed this block since `docker-go` is doing the same thing and gets the necessary go env vars

tools/buildsys/src/gomod.rs
jpmcb: 👍🏼 Removed this function and get `BUILDSYS_SDK_IMAGE` directly. Omitted the `rerun-if-env-changed` call

tools/buildsys/src/gomod.rs
jpmcb: 👍🏼 Done!

tools/buildsys/src/gomod.rs
jpmcb: Ended up going with rebuilding the vendor directory and re-archiving. It didn't have an impact on performance since the local modcache should already be populated with the dependencies. It also seemed overly complicated to track the hashes of the dependencies and force a rebuild if any of those change.

But, like Sean said, I'd be very very surprised if those dependencies _did_ change out of band from the upstream releasing a new version of their software: the `go.mod` and `go.sum` act as locks on what dependencies are needed and shouldn't be different build to build.

jpmcb:
@stmcginnis - I think something is wrong with the cloud-provider-aws `v1.22.3` tag and they have an out of date `go.mod` file.

---

I am able to pull down the tar with
```
❯ wget https://github.com/kubernetes/cloud-provider-aws/archive/refs/tags/v1.22.3.tar.gz
```

and unpack it with:
```
❯ tar -zxvf v1.22.3.tar.gz
```

Then, going into that directory, I can vendor the code. But when I attempt to manually run a `go build` I get the following:
```
cd cloud-provider-aws-1.22.3
go mod vendor
GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=arm64 GOPROXY=https://proxy.golang.org,direct go build -x \
        -trimpath \
        -ldflags="-w -s -X k8s.io/component-base/version.gitVersion=" \
        -o=ecr-credential-provider \
        cmd/ecr-credential-provider/*.go
WORK=/tmp/go-build4216558415
go: updates to go.mod needed, disabled by -mod=vendor
        (Go version in go.mod is at least 1.14 and vendor directory exists.)
        to update it:
        go mod tidy
```

which is super weird. It's asking me to run a `go mod tidy`? I'd expect that to "just work" and there not need to be an update to the module dependencies. Switching over to the actual repository, I checked out their `v1.22.3` tag and see the same problem with the command above.

So, running `go mod tidy` on their git repo _does_ make it dirty and show the dependencies are out of date:
```
diff --git a/go.mod b/go.mod
index 27a8a4a..b19147b 100644
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
        github.com/aws/aws-sdk-go v1.43.32
        github.com/golang/mock v1.4.4
        github.com/google/go-cmp v0.5.5
+       github.com/spf13/pflag v1.0.5
        github.com/stretchr/testify v1.7.0
        gopkg.in/gcfg.v1 v1.2.0
        k8s.io/api v0.22.8
@@ -66,7 +67,6 @@ require (
        github.com/prometheus/common v0.26.0 // indirect
        github.com/prometheus/procfs v0.6.0 // indirect
        github.com/spf13/cobra v1.1.3 // indirect
-       github.com/spf13/pflag v1.0.5 // indirect
        github.com/stretchr/objx v0.2.0 // indirect
        go.etcd.io/etcd/api/v3 v3.5.0 // indirect
        go.etcd.io/etcd/client/pkg/v3 v3.5.0 // indirect
@@ -106,7 +106,6 @@ require (
        gopkg.in/yaml.v2 v2.4.0 // indirect
        gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
        k8s.io/apiserver v0.22.8 // indirect
-       k8s.io/controller-manager v0.22.8 // indirect
        k8s.io/gengo v0.0.0-20201214224949-b6c5ce23f027 // indirect
        k8s.io/kube-openapi v0.0.0-20211109043538-20434351676c // indirect
        sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.30 // indirect
```

The `pflag` one is especially interesting since it _does_ seems they use it directly in their controller:

```
❯ rg pflag
...

pkg/controllers/options/tagging_controller.go
18:     "github.com/spf13/pflag"
29:func (o *TaggingControllerOptions) AddFlags(fs *pflag.FlagSet) {
```

But checking out their `master` branch, cleaning everything up, and running through the sequence again works fine without any problem in their `go.mod`:

```
❯ git checkout master

❯ rm ecr-credential-provider && rm -rf vendor

❯ git reset --hard origin/master
HEAD is now at c72f89b Merge pull request #471 from wangamel/master

❯ GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=arm64 GOPROXY=https://proxy.golang.org,direct go build -x \
        -trimpath \
        -ldflags="-w -s -X k8s.io/component-base/version.gitVersion=" \
        -o=ecr-credential-provider \
        cmd/ecr-credential-provider/*.go

❯ ll ecr-credential-provider
-rwxr-xr-x. 1 fedora fedora 14M Aug 26 22:53 ecr-credential-provider
```

---

My presumption is their `v1.22.3` tag had an out of date go.mod which causes `go mod vendor` to get the wrong dependencies which then prevents a build.

I also wonder if `make` was eating all of that error output as an argument list?

Can we ask them to cut a new, up to date tag / tar with the right go.mod?

stmcginnis:
> My presumption is their v1.22.3 tag had an out of date go.mod which causes go mod vendor to get the wrong dependencies which then prevents a build.

Go would update the go.mod file, so I had assumed that would resolve any issues before vendoring. But I think you're right there. I found there is a newer v1.24.1. tag and the dependency issues have since been cleaned up (with one minor change still stale).

That seems to be enough that it fixed whatever was happening though:

```
$ git branch
* (HEAD detached at v1.24.1)
  master
$ go mod tidy
$ git status
HEAD detached at v1.24.1
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   go.mod

no changes added to commit (use "git add" and/or "git commit -a")
$ PAGER= git diff
diff --git a/go.mod b/go.mod
index 76e3379..8c6d8ca 100644
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/google/go-cmp v0.5.8
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.2
+	golang.org/x/time v0.0.0-20220210224613-90d013bbcef8
 	gopkg.in/gcfg.v1 v1.2.3
 	k8s.io/api v0.24.2
 	k8s.io/apimachinery v0.24.2
@@ -94,7 +95,6 @@ require (
 	golang.org/x/sys v0.0.0-20220209214540-3681064d5158 // indirect
 	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211 // indirect
 	golang.org/x/text v0.3.7 // indirect
-	golang.org/x/time v0.0.0-20220210224613-90d013bbcef8 // indirect
 	golang.org/x/tools v0.1.10-0.20220218145154-897bd77cd717 // indirect
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
$ go mod vendor 
$ docker run --rm --network none -v ${PWD}:/build -w /build golang:1.18 make ecr-credential-provider
GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=arm64 GOPROXY=https://proxy.golang.org,direct go build \
	-trimpath \
	-ldflags="-w -s -X k8s.io/component-base/version.gitVersion=v1.24.1-dirty" \
	-o=ecr-credential-provider \
	cmd/ecr-credential-provider/*.go
$ ls ecr-credential-provider
ecr-credential-provider
```

Glad this is at least captured here as I think we may run into other weirdness like this as we add other go packages. But looks like I can at least proceed for now.

And this change looks good to me. Again, just wanted to capture some weirdness so there's some record of it.

jpmcb:
> Glad this is at least captured here as I think we may run into other weirdness like this as we add other go packages. But looks like I can at least proceed for now.

@stmcginnis Awesome!!! 👏🏼 glad that works!! And yes, glad it's captured here

stmcginnis (approved):
Tested this out with the work I'm doing. I'm still working through other issues, but I've gotten past the part covered by this and everything appears to be working well.

packages/hotdog/hotdog.spec
```diff
@@ -10,67 +10,44 @@
...
-mkdir -p GOPATH/src/github.com/opencontainers/runtime-spec
-tar -C GOPATH/src/github.com/opencontainers/runtime-spec -xzf %{SOURCE1} --strip 1
-cp GOPATH/src/github.com/opencontainers/runtime-spec/LICENSE LICENSE.runtime-spec
 
-mkdir -p GOPATH/src/golang.org/x/sys
-tar -C GOPATH/src/golang.org/x/sys -xzf %{SOURCE2} --strip 1
-cp GOPATH/src/golang.org/x/sys/LICENSE LICENSE.golang-sys
+# Expand the upstream source code tarball
+# -n: use the named directory "hotdog-{long-sha}"
+%setup -n %{gorepo}-%{gitrev}
```
bcressey, line 28: This level of detail is helpful in the ecs-agent spec, which has so much going on that it'd be inscrutable without the comments.

Here I think it'd be simpler as a two-liner:
```suggestion
%setup -n %{gorepo}-%{gitrev}
%setup -T -D -n %{gorepo}-%{gitrev} -b 1
```

Otherwise it's a bit like trying to teach the language through comments.

tools/buildsys/src/manifest.rs
```diff
@@ -33,6 +33,46 @@ url = "https://bar"
...
+Only `bundle-modules` is required when bundling an archive's dependences.
+
+`bundle-modules` is a list of module "paradigms" the external-file should
+be vendored through. For example, if a project contains a go.mod and go.sum file,
+"go" may be entered to vendor the dependencies.
+Currently, only "go" is supported.
+
+`bundle-path` is an optional argument that provides the filepath
+within the archive that contains the module.
+By default, the first top level directory in the archieve is used.
```
bcressey:
```suggestion
By default, the first top level directory in the archive is used.
```

tools/buildsys/src/manifest.rs
```diff
@@ -33,6 +33,46 @@ url = "https://bar"
...
+So, for example, given a go project that has the necessary go.mod and go.sum
+files within the archive located at the filepath a/b/c,
+this `bundle-path` value should be `a/b/c`.
+Or, given a archieve with a single directory that contains a go project
+that has go.mod and go.sum files, this option may be omitted since the single
+top level directory will be used.
```
bcressey: nit: various formatting suggestions
```suggestion
So, for example, given a Go project that has the necessary `go.mod` and `go.sum`
files within the archive located at the filepath `a/b/c`,
this `bundle-path` value should be "a/b/c".
Or, given an archive with a single directory that contains a Go project
that has `go.mod` and `go.sum` files, this option may be omitted since the single
top-level directory will be used.
```

tools/buildsys/src/manifest.rs
```diff
@@ -33,6 +33,46 @@ url = "https://bar"
...
+
+`bundle-output-path` is an optional argument that provides
+the desired path of the output archive.
+By default, this will use the name of the existing archive,
+but pre-pended with "bundled-".
+This path may be referenced within a spec, when creating a package, etc.
+For example, `my-package-bundled.tar.gz` may be entered as an output path
+indicating the package is "bundled" and has been archived.
```
bcressey: I don't quite follow the second part of this sentence.

It might be better to reorder the sentences somewhat and talk about usage in the spec file at the end.

tools/buildsys/src/manifest.rs
```diff
@@ -33,6 +33,46 @@ url = "https://bar"
...
+Or, by default, given the name of the upstream archive is "my-package.tar.gz",
+the output archive would be named "bundled-my-package.tar.gz"
+```
+[[package.metadata.build-package.external-files]]
+path = "foo"
+url = "https://foo"
+sha512 = "abcdef"
+bundle-modules = [ "go" ]
+bundle-path = "path/to/module/"
```
bcressey: nit: unless the trailing slash is significant (which it shouldn't be) I'd omit it so that people don't feel like they always have to include it.

Maybe the field should be `bundle-control-path` or something, to better distinguish it from `bundle-output-path`?

tools/docker-go
```diff
@@ -56,7 +56,7 @@ parse_args "${@}"
 
 # Pass through relevant Go variables, from the config or environment.
 go_env=( )
-for i in GOPROXY GONOPROXY GOPRIVATE ; do
+for i in GOPROXY GONOPROXY GOPRIVATE GOSUMDB ; do
```
bcressey, line 59: Nice, this is much simpler.

tools/buildsys/src/main.rs
```diff
@@ -143,6 +150,20 @@ fn build_package() -> Result<()> {
 
     if let Some(files) = manifest.external_files() {
         LookasideCache::fetch(files).context(error::ExternalFileFetchSnafu)?;
+        for f in files {
+            if f.bundle_modules.is_none() {
+                continue;
+            }
+
+            for b in f.bundle_modules.as_ref().unwrap() {
+                println!("cargo:rerun-if-changed={:?}", f.bundle_output_path);
```
bcressey: This doesn't do the right thing if the output path is left unspecified, right? Since we haven't calculated the default value yet. But I think it's redundant with the logic in `gomod.rs`.

tools/buildsys/src/gomod/error.rs
```diff
@@ -0,0 +1,33 @@
...
+use snafu::Snafu;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub(super)))]
+pub(crate) enum Error {
+    #[snafu(display("Failed to start command: {}", source))]
+    CommandStart { source: std::io::Error },
+
+    #[snafu(display("Failed to execute command: 'docker {:?}'", args))]
+    DockerExecution { args: String },
```
bcressey: `String` implements `Display`, so you don't have to use the `Debug` impl:
```suggestion
    #[snafu(display("Failed to execute command: 'docker {}'", args))]
    DockerExecution { args: String },
```

tools/buildsys/src/gomod/error.rs
```diff
@@ -0,0 +1,33 @@
...
+
+    #[snafu(display("Input url is required"))]
+    InputFile,
+
+    #[snafu(display("Input file {:?} must be a file", path))]
+    InputFileBad { path: PathBuf },
```
bcressey:
```suggestion
    #[snafu(display("Input file {} must be a file", path.display()))]
    InputFileBad { path: PathBuf },
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,151 @@
+/*!
+Packages using the Go programming language may have upstream tar archives that
+include only the source code of the project, but not the source code of any
+dependencies. The Go programming language promotes the use of "modules" for
+dependencies and projects adopting modules will provide go.mod and go.sum
```
bcressey: nit:
```suggestion
dependencies and projects adopting modules will provide `go.mod` and `go.sum`
```

(same nit below)

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,151 @@
...
+files.
+
+This module extends the functionality of `packages.metadata.build-package.external-files`
+and provides the ability to retrieve and validate dependencies
+declared using Go modules given a tar archive containing a go.mod and go.sum.
+
+The location where dependencies are retrieved from are controlled by the
+standard environment variables employed by the Go tool: GOPROXY, GOSUMDB, and
+GOPRIVATE and are automatically retrieved from the host environment
+when docker-go is envoked.
```
bcressey:
```suggestion
The location where dependencies are retrieved from are controlled by the
standard environment variables employed by the Go tool: `GOPROXY`, `GOSUMDB`, and
`GOPRIVATE`. These variables are automatically retrieved from the host environment
when docker-go is invoked.
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,151 @@
...
+pub(crate) struct GoMod;
+
+impl GoMod {
+    pub(crate) fn vendor(
+        root_dir: &Path,
+        package_dir: &Path,
+        gomod: &manifest::ExternalFile,
+    ) -> Result<()> {
+        let url_file_name = extract_file_name(&gomod.url)?;
+        let path_arg = &gomod.path.as_ref().unwrap_or(&url_file_name);
```
bcressey: nit:
```suggestion
        let local_file_name = &gomod.path.as_ref().unwrap_or(&url_file_name);
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,151 @@
...
+use std::path::{Path, PathBuf};
+use std::process::Output;
+
+pub(crate) struct GoMod;
+
+impl GoMod {
+    pub(crate) fn vendor(
+        root_dir: &Path,
+        package_dir: &Path,
+        gomod: &manifest::ExternalFile,
```
bcressey: Maybe this for clarity?
```suggestion
        external_file: &manifest::ExternalFile,
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,151 @@
...
+        let url_file_name = extract_file_name(&gomod.url)?;
+        let path_arg = &gomod.path.as_ref().unwrap_or(&url_file_name);
+        ensure!(path_arg.components().count() == 1, error::InputFileSnafu);
+
+        let path = package_dir.join(path_arg);
+        ensure!(path.is_file(), error::InputFileBadSnafu { path });
+
+        // If a module directory was not provided, set as an empty path.
+        // By default, without a provided module directory, tar will be passed
+        // the first directory found in the archieve as the top level Go module
```
bcressey:
```suggestion
        // the first directory found in the archive as the top level Go module
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,151 @@
...
+        let default_empty_path = PathBuf::from("");
+        let mod_dir = gomod.bundle_path.as_ref().unwrap_or(&default_empty_path);
```
bcressey:
```suggestion
        let mod_dir = gomod.bundle_path.as_ref().unwrap_or_else(|| "".to_string());
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,151 @@
...
+                // If no output path was given, it defaults to "bundled-{tar-file-name}".
+                // Finally, it cleans up by removing the untar'd source code.
+                // The upstream archive remains and both tar files can be used during packaging.
+                "toplevel=$(tar tf {input} | head -1)
+                tar xf {input} &&
+                pushd $(if [ -z {moddir} ]; then echo $toplevel; else echo {moddir}; fi) &&
+                go list -mod=readonly ./... >/dev/null && go mod vendor &&
+                popd &&
+                tar czf {output} $(if [ -z {moddir} ]; then echo $toplevel; else echo {moddir}; fi)/vendor && 
+                rm -rf $(if [ -z {moddir} ]; then echo $toplevel; else echo {moddir}; fi)",
```
bcressey: This might be more readable as a constant toward the top of the program, using a multi-line raw string literal.

It might also be worth defining a function (like `go_fmt` or `go_fetch` in `Makefile.toml` so it's more clear how the parameters we pass into the format string line up with the parts of the execution that vary.

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,151 @@
...
+                // if no explicit module path was provided.
+                // It will then untar the archive, vendor the Go dependencies,
+                // create a new archieve using the {module-path}/vendor directory
+                // and name it the output path provided.
+                // If no output path was given, it defaults to "bundled-{tar-file-name}".
+                // Finally, it cleans up by removing the untar'd source code.
+                // The upstream archive remains and both tar files can be used during packaging.
+                "toplevel=$(tar tf {input} | head -1)
+                tar xf {input} &&
+                pushd $(if [ -z {moddir} ]; then echo $toplevel; else echo {moddir}; fi) &&
```
bcressey: I'd recommend not doing the Dockerfile-ish technique of writing `&&` everywhere if it's not necessary.

This would be more readable as:
```
if [ -n "${moddir}" ] ; then
  targetdir="${moddir}"
else
  targetdir="${toplevel}"
fi

pushd "${targetdir}"
# do the vendoring
popd

if [ ! -d "${targetdir}/vendor ] ;
  echo "no vendor directory found" >&2
  exit 1
fi

tar czf ... "${targetdir}/vendor"

# clean up - why?
rm -rf "${targetdir}"
```

bcressey (commented):
Nice work! Lots of nits but I don't see anything major missing.

packages/hotdog/hotdog.spec
jpmcb: > Otherwise it's a bit like trying to teach the language through comments.

Yeah, super fair. I wasn't sure what level of detail to go into with what was happening here (and was looking at a bunch of the other specs to see what was happening) but yes, a simple 2 liner makes sense!

tools/buildsys/src/manifest.rs
jpmcb: Hmmmm good point. Naming things is hard. What about `bundle-root-path` to indicate it's the root of the bundle where the module's dependency tracking occurs?

tools/buildsys/src/main.rs
jpmcb: Ah yes, correct; it is redundant - this was a leftover from the first commit.

Removing it.

tools/buildsys/src/gomod.rs
jpmcb: I had tried this originally but got hung up on the typing:

```
error[E0308]: mismatched types
  --> buildsys/src/gomod.rs:48:76
   |
48 |         let mod_dir = external_file.bundle_path.as_ref().unwrap_or_else(|| "".to_string()); // &default_empty_path);
   |                                                                            ^^^^^^^^^^^^^^ expected `&PathBuf`, found struct `std::string::String`
```

My understanding is that since `external_file_path` is a `PathBuf` and we are attempting to unwrap it, we would also need to return a pathbuf if it fails, not just an empty string? That's what lead me down the path of declaring an empty pathbuf and using it's borrowed value as the default if none was provided.

jpmcb:
Added _another_ commit which reworks how the bundling script is used by `docker-go`:

Now, instead of attempting to feed a long script delimited by `&&` or `\`, the script lives as a `const` string that gets added as a temporary file to the package directory. This file is then accessible by the container and easily run with `bash -c ./docker-go-script.sh`. Further, the bundling script now expects several environment variables which are loaded through a new `--additional-docker-args` option in the `docker-go` tool; this enables us to use the `-e` arguments for `docker` to load up environment variables into the container.

Further documentation in comments in code.

tools/docker-go
```diff
@@ -18,6 +18,9 @@ Required:
     --sdk-image                 Name of the SDK image to use
     --go-mod-cache              The Go module cache path to mount into the container
     --command                   The command to run in the SDK container
+
+Optional:
+	--additional-docker-args    Additional free-form arguments to provide to "docker run" (i.e. mounts, env-vars, etc.)
```
stmcginnis: Indentation is a little off here. Minor thing if you need to do another update.

jpmcb: Hmmm weird, nvim inserted a tab here instead of 4 spaces. But yes, good catch! Will update with other feedback

tools/buildsys/src/manifest.rs
```diff
@@ -33,6 +33,49 @@ url = "https://bar"
 sha512 = "123456"
 ```
 
+The `bundle-*` keys on `external-files` are a group of optional modifiers
+and are used to untar an upstream external file archive,
+vendor any dependent code,
+and produce an additional archive with those dependencies.
+Only `bundle-modules` is required when bundling an archive's dependences.
```
bcressey: nit: the line-wrapping seems pretty short or inconsistent here, not that it was very consistent in this file before. I'd say either 80 or 100 are acceptable values. It's more noticeable down by the `bundle-output-path` comments.

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,213 @@
...
+if [ -z "${MOD_DIR}" ] ; then
+    targetdir="${toplevel}"
+else
+    targetdir="${MOD_DIR}"
+fi
+
+tar xf "${LOCAL_FILE_NAME}"
+
+pushd "${targetdir}"
+	go list -mod=readonly ./... >/dev/null && go mod vendor
```
bcressey: nit: a tab snuck in here
```suggestion
    go list -mod=readonly ./... >/dev/null && go mod vendor
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,213 @@
...
+        match docker_go(root_dir, &args) {
+            Ok(_) => {
+                fs::remove_file(script_path).unwrap();
+                Ok(())
+            }
+            Err(e) => {
+                fs::remove_file(script_path).unwrap();
+                Err(e)
+            }
+        }
```
bcressey: Hmm, a more idiomatic way to do this might be:
```suggestion
        let res = docker_go(root_dir, &args);
        fs::remove_file(script_path)?;
        res
```

tools/buildsys/src/gomod.rs
```diff
@@ -0,0 +1,213 @@
...
+            module_path: package_dir,
+            sdk_image: sdk,
+            go_mod_cache: &root_dir.join(".gomodcache"),
+            command: format!("./{}", GO_MOD_DOCKER_SCRIPT_NAME),
+            extra_docker_args: vec![
+                // Sets container environment variables needed for the docker-go-script
+                format!("-e LOCAL_FILE_NAME={}", local_file_name.to_string_lossy()),
+                format!("-e MOD_DIR={}", mod_dir.to_string_lossy()),
+                format!("-e OUTPUT={}", output_path_arg.to_string_lossy()),
+            ],
```
bcressey: I don't completely love the new "extra docker args" functionality since then it's harder to reason about what the tool will or won't do.

Rather than use variables in the script, it might be worth using placeholders like `__LOCAL_FILE_NAME__` and then just replacing them with these values before writing out the file.

bcressey (commented):
Looks good. In terms of the commits:
* I'd "unsquash" the first one and have two separate commits for the `hotdog` and `oci-add-hooks` changes.
* There's at least one typo in the second commit message
* You can squash all the `buildsys` commits together as the first patch in the series (probably you were planning to do this).

tools/buildsys/src/gomod.rs
jpmcb: Good call - this was likely an implementation artifact from exploring how a script could be used with the `docker-go` tool. I think using a template string is less complex in the end and if someone wanted to look at the script afterwards (or while in process) they could do see those values populated

jpmcb:
Force pushed to flesh out the commit structure per @bcressey's recommendations and address a few nits!

Removed changes to `tools/docker-go` which added the `--additional-docker-args` and instead, the go module script uses a few placeholder variables that are populated before the temporary file is put in the package directory.

bcressey (approved):
Yay!

stmcginnis (approved):
Tested with https://github.com/bottlerocket-os/bottlerocket/pull/2377 and it looks like everything works as expected.

Issue #2246, opened 2022-06-27
Metal: Expand server hardware compatibility


zmrow:
For our metal variants, we should greatly expand our hardware compatibility.  Ideally, we would support all main classes of server hardware.

foersleo:
First need to split out the metal specific parts so we keep track of what went in for metal specifically (Issue https://github.com/bottlerocket-os/bottlerocket/issues/2211).

foersleo:
I have been working towards this and checked what we are currently missing, mostly in the storage and networking side.

We are currently mostly missing support for some 10G+ networking equipment by various vendors. (https://github.com/bottlerocket-os/bottlerocket/issues/2361 is referring to one of the missing device series we are missing).

There is some features that are supported by multiple devices where we can have a discussion if those are necessary for Bottlerocket use cases and enable them for all supporting devices (i.e. DCB).

foersleo:
There is some additional planning work that needs to happen here.
We so far have worked towards enabling what I'd call modern common server hardware  (standard RAID controllers and 10G+ Ethernet NICs). However, it is not quite clear if that is the actual boundaries where we want to stop at for the bare metal variant.

Some of the questions that need to be discussed:

- [ ] Where are the boundaries of expanding HW compatibility?
- [ ] What do people need in that regard? What does the community want?
- [ ] What level of support and/or maintainership can we offer for what kind of HW?
- [ ] How can we standardize testing so that we can reason about compatibility for each version?
- [ ] How can we load share testing/verification for different platforms within the community? Who can contribute?

PR #2379, opened 2022-08-25
metal: hw enablement of 10G+ networking equipment

### LINKED ISSUE #2361: metal: Support the Intel X710 NIC
We should support the Intel X710 NIC, which I believe should be covered by the `i40e` driver, i.e `CONFIG_I40E`

foersleo:
**Issue number:** https://github.com/bottlerocket-os/bottlerocket/issues/2361 https://github.com/bottlerocket-os/bottlerocket/issues/2246

Closes #2361

**Description of changes:**

Enable 10G+ networking equipment for the metal variants for most vendors. While there is some cards that provide quite some advanced network features we are aiming for a meaningful minimum here, by enabling mostly the base Ethernet capabilities for these devices.

For advanced features we will have a separate discussion on if we need these out of the box or if those are valuable at all in the Bottlerocket use case. Features that I have identified and will discuss with the team: Infiniband support, DCB, Fibre Channel, SCSI/ISCSI (target) capabilities, offloading capabilities for some of the hardware (crypto offloading, flow processing).

As it is easier to add those features in later than to add them now and disable them later I have aimed for as few special features as possible.

For any Options resulting from enabling a specific device family I have explicitly set those to the desired setting.

The following options are not set explicitly as those are device families that the kernel has disabled by default:

```
# Option available due to CONFIG_NET_VENDOR_AMD being set
AMD8111_ETH n
PCNET32 n

# Options available due to CONFIG_NET_VENDOR_CHELSIO being set
CHELSIO_T1 n
CHELSIO_T3 n
```

Let's discuss if we we want to set those explicitly or if we are fine with these devices being disabled implicitly. Given that all these drivers are for older hardware it is reasonable to assume that they will not change their default down the line.

In addition there is option `AMD_XGBE_HAVE_ECC` which is set conditionally based on the architecture being `X86` when `AMD_XGBE` is set. As we currently do not have separate config fragments for `aarch64` and `x86_64` we can not explicitly set it without causing trouble.

**Testing done:**

Built some versions and compared resulting kernel configurations with the tooling currently under review in https://github.com/bottlerocket-os/bottlerocket/pull/2368 

```
config-aarch64-5.10-aws-k8s-1.23-diff:	  0 removed,   0 added,   0 changed
config-aarch64-5.15-aws-dev-diff:	  0 removed,   0 added,   0 changed
config-aarch64-5.15-metal-dev-diff:	  0 removed,  30 added,  11 changed
config-x86_64-5.10-aws-k8s-1.23-diff:	  0 removed,   0 added,   0 changed
config-x86_64-5.10-metal-k8s-1.23-diff:	  0 removed,  32 added,  11 changed
config-x86_64-5.15-aws-dev-diff:	  0 removed,   0 added,   0 changed
config-x86_64-5.15-metal-dev-diff:	  0 removed,  32 added,  11 changed
```

The full resulting config diff-report can be found in https://gist.github.com/foersleo/918342143826d4291c6d84bcc4e18194

**Terms of contribution:**


zmrow (approved):
:tada: :taco:

packages/kernel-5.10/config-bottlerocket-metal
```diff
@@ -7,32 +7,67 @@ CONFIG_SATA_AHCI=y
...
+CONFIG_NET_VENDOR_EMULEX=y
+CONFIG_BE2NET=m
+CONFIG_BE2NET_BE2=y
+CONFIG_BE2NET_BE3=y
+CONFIG_BE2NET_HWMON=y
+CONFIG_BE2NET_LANCER=y
+CONFIG_BE2NET_SKYHAWK=y
+
+# Huawei network support
+CONFIG_NET_VENDOR_HUAWEI
```
markusboehme: This is just mentioning the config option, not giving any value (i.e. `=y`). Same in the config for kernel-5.15.

packages/kernel-5.15/config-bottlerocket-metal
```diff
@@ -42,6 +77,23 @@ CONFIG_NET_VENDOR_MELLANOX=y
...
+CONFIG_MYRI10GE_DCA=y
+
+# Pensando network support
+CONFIG_NET_VENDOR_PENSANDO=y
+CONFIG_IONIC=m
+
+# Solarflare network support
+CONFIG_NET_VENDOR_SOLARFLARE=y
+CONFIG_SFC=m
+CONFIG_SRIOV=y
```
markusboehme: `CONFIG_SFC_SRIOV`?

packages/kernel-5.10/config-bottlerocket-metal
```diff
@@ -7,32 +7,67 @@ CONFIG_SATA_AHCI=y
...
 CONFIG_NET_VENDOR_BROADCOM=y
 CONFIG_TIGON3_HWMON=y
 CONFIG_TIGON3=m
+CONFIG_BNX2X=m
+CONFIG_BNX2X_SRIOV=y
 CONFIG_BNXT=m
 
+# Chelsio network support
+CONFIG_NET_VENDOR_CHELSIO=y
+CONFIG_CHELSIO_T4=m
```
markusboehme, line 25: In the full config diff, I see the Chelsio T1 and T3 drivers are deselected/they're not explicitly selected here. I'm not familiar with Chelsio's product catalogue and haven't looked deeper yet, but those also seem to be 10Gbit/s Ethernet NICs. Why are we not selecting them as well?

packages/kernel-5.10/config-bottlerocket-metal
foersleo: Ouch, this is embarrassing... I thought I had worked out all these kinks. Fixed in the force push below.

packages/kernel-5.15/config-bottlerocket-metal
foersleo: Ouch, this is similarly embarrassing... I thought I had worked out all these kinks. Fixed in the force push below.

packages/kernel-5.10/config-bottlerocket-metal
foersleo: So as far as I understand the Chelsio T1 and T3 devices are mid 2000s vintage and have been discontinued for quite a while. All later models (T4, T5, T6) are serviced by the T4 variant of the driver. While those older T1/T3 cards have 10G support I left them out for that reason.

I am happy to take these if we find a use case for them or as a precaution to be complete here.

What we might want to use as a guide here: RHEL seems to not include these, while openSUSE does. Happy to have a discussion on this.

foersleo:
Force push fixed the two bugs I had missed before sending this. Currently running the `diff-kernel-config` script again. Will share the output of that as soon as it finishes.

packages/kernel-5.10/config-bottlerocket-metal
markusboehme: Thanks for explaining, this makes sense. I agree on leaving them out and would be very surprised to still find them in any data center then.

foersleo:
Update kernel-config diff in the gist linked above and the statistics after the last change.

markusboehme (approved):

bcressey (approved):
🎸

PR #2384, opened 2022-08-29
makefile: add top level "check" target

### LINKED ISSUE #2381: Makefile: provide a unified `check` target that performs all `check-*` tasks
### What I'd like:

Currently, as a developer of Bottlerocket, checking formatting and linters requires both `cargo make check-fmt` and `cargo make check-lints` which is a bit annoying and easy to miss one or the other.

To facilitate easier development and have fewer GitHub action failures, we should provide a `cargo make check` target that performs _all_ dev related checks.

### Any alternatives you've considered:

- Alias my own local script that performs the check tasks and use that for a pre-commit hook

Would love feedback on this idea or if others would find it useful!

jpmcb:
### Issue number:

Closes #2381

### Description of changes:

```
"cargo make check" will check the cargo version, invoke the unit tests
(which in turn calls rustc thereby creating necessary build file links for
the default variant), checks formatting, and checks the linters.

```

Provides a unified, top level "check" target to the makefile that can be used in a `.git/hooks/pre-commit` script to ensure review and patch readiness.

Incorporates only the `check-*` targets that are relevant for development work (unit tests, formatting, linting, etc.)

### Testing done:

Ran `cargo make check` and 👍🏼
Ran `cargo make check` in a new git clone of the repo and 👍🏼 

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -330,6 +330,10 @@ done
 '''
 ]
 
+# A top level target for devs to ensure review and patch readiness
+[tasks.check]
+dependencies = ["check-cargo-version", "unit-tests", "check-fmt", "check-lints"]
+
 [tasks.check-fmt]
 script = [
 '''


stmcginnis:
I get an error running this:

```
Error writing files: failed to resolve mod `variant`: /home/smcginnis/src/bottlerocket-os/bottlerocket/sources/models/src/variant.rs does not exist
```

Peek 2022-08-29 14-13

I believe `variant.rs` is a generated file, but I'm not sure where that's done.

stmcginnis:
Thanks to @etungsten we poked around a bit to see what's going on.

On a fresh clone where a build hasn't been done, there are a couple symlinks that are not created yet. Under `sources/models/src/variant`, after a build you end up with a `current` and a `mod.rs` symlink. These are created in the `build.rs` file, so that doesn't get created until an actual build is done.

I was able to workaround this by adding:

```
 # For rust first-party source code
+if [ ! -f "${BUILDSYS_SOURCES_DIR}/models/src/variant/mod.rs" ]; then
+    pushd "${BUILDSYS_SOURCES_DIR}/models/src/variant"
+    ln -s ../variant_mod.rs mod.rs
+    popd
+fi
+if [ ! -f "${BUILDSYS_SOURCES_DIR}/models/src/variant/current" ]; then
+    pushd "${BUILDSYS_SOURCES_DIR}/models/src/variant"
+    ln -s ../aws-k8s-1.23 current
+    popd
+fi
```

So I think we either need to detect if the link is missing an create it, or detect if it is missing and error out with some sort of message like "Build must be performed before running this target".

jpmcb:
Force pushed to address @stmcginnis comments regarding failure modes if bottlerocket hasn't been built yet

Makefile.toml
```diff
@@ -330,6 +330,28 @@ done
...
+    popd > /dev/null
+fi
+
+if [ ! -d "${BUILDSYS_SOURCES_DIR}/models/src/variant/current" ]; then
+    pushd "${BUILDSYS_SOURCES_DIR}/models/src/variant" > /dev/null
+    ln -s ../aws-k8s-1.23 current
+    popd > /dev/null
+fi
+'''
+]
```
bcressey: This makes me uncomfortable, on account of [all](https://github.com/bottlerocket-os/bottlerocket/blob/e354d2cc7c907f717d8851aede27cd1f3991c001/Dockerfile#L36) [the](https://github.com/bottlerocket-os/bottlerocket/blob/e354d2cc7c907f717d8851aede27cd1f3991c001/Dockerfile#L114) [logic](https://github.com/bottlerocket-os/bottlerocket/blob/e354d2cc7c907f717d8851aede27cd1f3991c001/sources/models/build.rs#L54) involved in getting the right thing to happen. It'd be easy to break builds outside the container without realizing it, in a subtle way like reusing the model from a previous variant's build.

I'd recommend doing what `unit-tests` does instead and just export the expected variable in each of the tasks:
```
export VARIANT="${BUILDSYS_VARIANT}"
```

etungsten: The symlinks are created as part of `models`'s build script. Cargo build scripts are only invoked upon compilation of the crate. Since `cargo fmt`, unlike `cargo test`, doesn't invoke `rustc` to compile or build anything, the `build.rs` script won't get invoked.

bcressey: 😿 

Maybe add `unit-tests` into the list before `check-fmt`, so the links get created first?

Ideally we'd be able to run clippy and `check-lints` could come before `check-fmt`, but that's probably too much to bite off right now.

Even if a workaround is required, I'd really prefer not to duplicate this sort of link logic outside the build script for `models` - the behavior is already very complicated.

jpmcb: > Maybe add unit-tests into the list before check-fmt, so the links get created first?

I think that's the right thing here given the complexity of how those files are to be populated. Additionally, running the tests as part of a top level "check" isn't a bad idea anyways

jpmcb:
Force pushed to address Ben's concerns around complexity in creating links out of band from the build or the rust compiler. Now, `cargo make check` invokes the unit tests which accomplishes the same thing.

etungsten (approved):

zmrow (approved):
:seedling:

jpculp (approved):

bcressey (approved):

PR #2393, opened 2022-08-31
testsys: Add declarative testing through `Test.toml`

ecpullen:
**Issue number:**

Closes # N/A

**Description of changes:**

Enables configuring testing through `Test.toml`. There are now 5 levels of configuration.

```
# The test configuration is broken up into different levels depending on the variant.
# Each configuration's value will be selected in the following order:
# Arguments from the command line
# Environment variables from the command line
# '{variant}' configuration in `Test.toml`
# '{family}-{flavor}' configuration in `Test.toml`
# '{family}' configuration in `Test.toml`
# '{platform}-{flavor}' configuration in `Test.toml`
# '{platform}' configuration in `Test.toml`
# TestSys default values
```

**Testing done:**

Testing workflow over many different configurations.
* Create a sample config file
* Run `cargo make test`
* Use `kubectl describe test <TEST-NAME>` and `kubectl describe resource <RESOURCE-NAME>` to see the `configuration`
* Verify that the values in the `configuration` match the values in the config file.

I tested this with a variety of different configurations making sure that each field was selected properly for each possible field. I then tested that the configuration values were placed in the correct spots in the crds.

I also tested `Test.toml.example` with the redacted fields filled in to make sure that the documentation was correct. 

**Terms of contribution:**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /.gomodcache
 /html
 /Infra.toml
+/Test.toml
 /testsys.kubeconfig
 /*.pem
 /keys
--- a/TESTING.md
+++ b/TESTING.md
@@ -118,6 +118,25 @@ The Bottlerocket components are found in the `testsys-bottlerocket-aws` Kubernet
 
 Now that you have the testsys cluster set up, it's time to run a Bottlerocket integration test!
 
+### Configuration
+
+There are many arguments that can be configured via environment variables with `cargo make`; however, it is possible to create a configuration file instead.
+Check out the [example config file](tools/testsys/Test.toml.example) for a sample `Test.toml` file.
+
+For example, the instance type can be specified based on variant requirements:
+
+```yaml
+[aws-k8s]
+# Set the default instance type for all `aws-k8s` variants
+instance-type = "m5.xlarge"
+
+[aws-k8s-nvidia]
+# Override the instance type for `nvidia` `aws-k8s` variants
+instance-type = "g5g.2xlarge"
+```
+
+Since `aws-k8s-nvidia` is a `<FAMILY>-<FLAVOR>` level configuration it will take precedence over `aws-k8s` which is `<FAMILY>` level configuration.
+
 ### Variants
 
 Different Bottlerocket variants require different implementations in the test system.
--- a/tools/testsys-config/src/lib.rs
+++ b/tools/testsys-config/src/lib.rs
@@ -1,0 +1,387 @@
+use bottlerocket_variant::Variant;
+pub use error::Error;
+use handlebars::Handlebars;
+use log::warn;
+use maplit::btreemap;
+use model::SecretName;
+use serde::{Deserialize, Serialize};
+use snafu::ResultExt;
+use std::collections::{BTreeMap, HashMap};
+use std::fs;
+use std::path::Path;
+pub type Result<T> = std::result::Result<T, error::Error>;
+
+/// Configuration needed to run tests
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "kebab-case")]
+pub struct TestConfig {
+    /// High level configuration for TestSys
+    pub test: Option<Test>,
+
+    #[serde(flatten, serialize_with = "toml::ser::tables_last")]
+    /// Configuration for testing variants
+    pub configs: HashMap<String, GenericConfig>,
+}
+
+impl TestConfig {
+    /// Deserializes a TestConfig from a given path
+    pub fn from_path<P>(path: P) -> Result<Self>
+    where
+        P: AsRef<Path>,
+    {
+        let path = path.as_ref();
+        let test_config_str = fs::read_to_string(path).context(error::FileSnafu { path })?;
+        toml::from_str(&test_config_str).context(error::InvalidTomlSnafu { path })
+    }
+
+    /// Deserializes a TestConfig from a given path, if it exists, otherwise builds a default
+    /// config
+    pub fn from_path_or_default<P>(path: P) -> Result<Self>
+    where
+        P: AsRef<Path>,
+    {
+        if path.as_ref().exists() {
+            Self::from_path(path)
+        } else {
+            warn!(
+                "No test config was found at '{}'. Using the default config.",
+                path.as_ref().display()
+            );
+            Ok(Self::default())
+        }
+    }
+
+    /// Create a single config for the `variant` and `arch` from this test configuration by
+    /// determining a list of tables that contain information relevant to the arch, variant
+    /// combination. Then, the tables are reduced to a single config by selecting values from the
+    /// table based on the order of precedence. If `starting_config` is provided it will be used as
+    /// the config with the highest precedence.
+    pub fn reduced_config<S>(
+        &self,
+        variant: &Variant,
+        arch: S,
+        starting_config: Option<GenericVariantConfig>,
+    ) -> GenericVariantConfig
+    where
+        S: Into<String>,
+    {
+        let arch = arch.into();
+        // Starting with a list of keys ordered by precedence, return a single config with values
+        // selected by the order of the list.
+        config_keys(variant)
+            // Convert the vec of keys in to an iterator of keys.
+            .into_iter()
+            // Convert the iterator of keys to and iterator of Configs. If the key does not have a
+            // configuration in the config file, remove it from the iterator.
+            .filter_map(|key| self.configs.get(&key).cloned())
+            // Take the iterator of configurations and extract the arch specific config and the
+            // non-arch specific config for each config. Then, convert them into a single iterator.
+            .flat_map(|config| vec![config.for_arch(&arch), config.config])
+            // Take the iterator of configurations and merge them into a single config by populating
+            // each field with the first value that is not `None` while following the list of
+            // precedence.
+            .fold(
+                starting_config.unwrap_or_default(),
+                GenericVariantConfig::merge,
+            )
+    }
+}
+
+/// High level configurations for a test
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub struct Test {
+    /// The name of the repo in `Infra.toml` that should be used for testing
+    pub repo: Option<String>,
+
+    #[serde(flatten)]
+    /// The URI of TestSys images
+    pub testsys_images: TestsysImages,
+
+    /// A registry containing all TestSys images
+    pub testsys_image_registry: Option<String>,
+}
+
+#[derive(Debug, Default)]
+pub struct AwsK8sVariantConfig {
+    /// The names of all clusters this variant should be tested over. This is particularly useful
+    /// for testing Bottlerocket on ipv4 and ipv6 clusters.
+    pub cluster_names: Vec<String>,
+    /// The instance type that instances should be launched with
+    pub instance_type: Option<String>,
+    /// The secrets needed by the agents
+    pub secrets: BTreeMap<String, SecretName>,
+    /// The role that should be assumed for this particular variant
+    pub assume_role: Option<String>,
+    /// The kubernetes conformance image that should be used for this variant
+    pub kube_conformance_image: Option<String>,
+    /// The e2e repo containing sonobuoy images
+    pub e2e_repo_registry: Option<String>,
+}
+
+#[derive(Debug, Default)]
+pub struct AwsEcsVariantConfig {
+    /// The names of all clusters this variant should be tested over
+    pub cluster_names: Vec<String>,
+    /// The instance type that instances should be launched with
+    pub instance_type: Option<String>,
+    /// The secrets needed by the agents
+    pub secrets: BTreeMap<String, SecretName>,
+    /// The role that should be assumed for this particular variant
+    pub assume_role: Option<String>,
+}
+
+/// Create a vec of relevant keys for this variant ordered from most specific to least specific.
+fn config_keys(variant: &Variant) -> Vec<String> {
+    let (family_flavor, platform_flavor) = variant
+        .variant_flavor()
+        .map(|flavor| {
+            (
+                format!("{}-{}", variant.family(), flavor),
+                format!("{}-{}", variant.platform(), flavor),
+            )
+        })
+        .unwrap_or_default();
+
+    // The keys used to describe configuration (most specific -> least specific)
+    vec![
+        variant.to_string(),
+        family_flavor,
+        variant.family().to_string(),
+        platform_flavor,
+        variant.platform().to_string(),
+    ]
+}
+
+/// All configurations for a specific config level, i.e `<PLATFORM>-<FLAVOR>`
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct GenericConfig {
+    #[serde(default)]
+    aarch64: GenericVariantConfig,
+    #[serde(default)]
+    x86_64: GenericVariantConfig,
+    #[serde(default, flatten)]
+    config: GenericVariantConfig,
+}
+
+impl GenericConfig {
+    /// Get the configuration for a specific arch.
+    pub fn for_arch<S>(&self, arch: S) -> GenericVariantConfig
+    where
+        S: Into<String>,
+    {
+        match arch.into().as_str() {
+            "x86_64" => self.x86_64.clone(),
+            "aarch64" => self.aarch64.clone(),
+            _ => Default::default(),
+        }
+    }
+}
+
+/// The configuration for a specific config level (<PLATFORM>-<FLAVOR>). This may or may not be arch
+/// specific depending on it's location in `GenericConfig`.
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
+pub struct GenericVariantConfig {
+    /// The names of all clusters this variant should be tested over. This is particularly useful
+    /// for testing Bottlerocket on ipv4 and ipv6 clusters.
+    #[serde(default)]
+    pub cluster_names: Vec<String>,
+    /// The instance type that instances should be launched with
+    pub instance_type: Option<String>,
+    /// The secrets needed by the agents
+    #[serde(default)]
+    pub secrets: BTreeMap<String, SecretName>,
+    /// The role that should be assumed for this particular variant
+    pub agent_role: Option<String>,
+    /// The custom images used for conformance testing
+    pub conformance_image: Option<String>,
+    /// The custom registry used for conformance testing
+    pub conformance_registry: Option<String>,
+}
+
+impl GenericVariantConfig {
+    /// Overwrite the unset values of `self` with the set values of `other`
+    fn merge(self, other: Self) -> Self {
+        let cluster_names = if self.cluster_names.is_empty() {
+            other.cluster_names
+        } else {
+            self.cluster_names
+        };
+
+        let secrets = if self.secrets.is_empty() {
+            other.secrets
+        } else {
+            self.secrets
+        };
+
+        Self {
+            cluster_names,
+            instance_type: self.instance_type.or(other.instance_type),
+            secrets,
+            agent_role: self.agent_role.or(other.agent_role),
+            conformance_image: self.conformance_image.or(other.conformance_image),
+            conformance_registry: self.conformance_registry.or(other.conformance_registry),
+        }
+    }
+}
+
+impl From<GenericVariantConfig> for AwsK8sVariantConfig {
+    fn from(val: GenericVariantConfig) -> Self {
+        Self {
+            cluster_names: val.cluster_names,
+            instance_type: val.instance_type,
+            secrets: val.secrets,
+            assume_role: val.agent_role,
+            kube_conformance_image: val.conformance_image,
+            e2e_repo_registry: val.conformance_registry,
+        }
+    }
+}
+
+impl From<GenericVariantConfig> for AwsEcsVariantConfig {
+    fn from(val: GenericVariantConfig) -> Self {
+        Self {
+            cluster_names: val.cluster_names,
+            instance_type: val.instance_type,
+            secrets: val.secrets,
+            assume_role: val.agent_role,
+        }
+    }
+}
+
+/// Fill in the templated cluster name with `arch` and `variant`.
+pub fn rendered_cluster_name<S1, S2>(cluster_name: String, arch: S1, variant: S2) -> Result<String>
+where
+    S1: Into<String>,
+    S2: Into<String>,
+{
+    let mut cluster_template = Handlebars::new();
+    cluster_template.register_template_string("cluster_name", cluster_name)?;
+    Ok(cluster_template.render(
+        "cluster_name",
+        &btreemap! {"arch".to_string() => arch.into(), "variant".to_string() => variant.into()},
+    )?)
+}
+
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
+pub struct TestsysImages {
+    pub eks_resource_agent_image: Option<String>,
+    pub ecs_resource_agent_image: Option<String>,
+    pub ec2_resource_agent_image: Option<String>,
+    pub sonobuoy_test_agent_image: Option<String>,
+    pub ecs_test_agent_image: Option<String>,
+    pub migration_test_agent_image: Option<String>,
+    pub testsys_agent_pull_secret: Option<String>,
+}
+
+const AGENT_VERSION: &str = "v0.0.2";
+
+impl TestsysImages {
+    /// Create an images config for a specific registry.
+    pub fn new<S>(registry: S) -> Self
+    where
+        S: Into<String>,
+    {
+        let registry = registry.into();
+        Self {
+            eks_resource_agent_image: Some(format!(
+                "{}/eks-resource-agent:{AGENT_VERSION}",
+                registry
+            )),
+            ecs_resource_agent_image: Some(format!(
+                "{}/ecs-resource-agent:{AGENT_VERSION}",
+                registry
+            )),
+            ec2_resource_agent_image: Some(format!(
+                "{}/ec2-resource-agent:{AGENT_VERSION}",
+                registry
+            )),
+            sonobuoy_test_agent_image: Some(format!(
+                "{}/sonobuoy-test-agent:{AGENT_VERSION}",
+                registry
+            )),
+            ecs_test_agent_image: Some(format!("{}/ecs-test-agent:{AGENT_VERSION}", registry)),
+            migration_test_agent_image: Some(format!(
+                "{}/migration-test-agent:{AGENT_VERSION}",
+                registry
+            )),
+            testsys_agent_pull_secret: None,
+        }
+    }
+
+    pub fn merge(self, other: Self) -> Self {
+        Self {
+            eks_resource_agent_image: self
+                .eks_resource_agent_image
+                .or(other.eks_resource_agent_image),
+            ecs_resource_agent_image: self
+                .ecs_resource_agent_image
+                .or(other.ecs_resource_agent_image),
+            ec2_resource_agent_image: self
+                .ec2_resource_agent_image
+                .or(other.ec2_resource_agent_image),
+            sonobuoy_test_agent_image: self
+                .sonobuoy_test_agent_image
+                .or(other.sonobuoy_test_agent_image),
+            ecs_test_agent_image: self.ecs_test_agent_image.or(other.ecs_test_agent_image),
+            migration_test_agent_image: self
+                .migration_test_agent_image
+                .or(other.migration_test_agent_image),
+            testsys_agent_pull_secret: self
+                .testsys_agent_pull_secret
+                .or(other.testsys_agent_pull_secret),
+        }
+    }
+
+    pub fn public_images() -> Self {
+        Self::new("public.ecr.aws/bottlerocket-test-system")
+    }
+}
+
+mod error {
+    use snafu::Snafu;
+    use std::io;
+    use std::path::PathBuf;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub enum Error {
+        #[snafu(display("Failed to read '{}': {}", path.display(), source))]
+        File { path: PathBuf, source: io::Error },
+
+        #[snafu(display("Invalid config file at '{}': {}", path.display(), source))]
+        InvalidToml {
+            path: PathBuf,
+            source: toml::de::Error,
+        },
+
+        #[snafu(display("Invalid lock file at '{}': {}", path.display(), source))]
+        InvalidLock {
+            path: PathBuf,
+            source: serde_yaml::Error,
+        },
+
+        #[snafu(display("Missing config: {}", what))]
+        MissingConfig { what: String },
+
+        #[snafu(display("Failed to get parent of path: {}", path.display()))]
+        Parent { path: PathBuf },
+
+        #[snafu(
+            context(false),
+            display("Failed to create template for cluster name: {}", source)
+        )]
+        TemplateError { source: handlebars::TemplateError },
+
+        #[snafu(
+            context(false),
+            display("Failed to render templated cluster name: {}", source)
+        )]
+        RenderError { source: handlebars::RenderError },
+    }
+}
--- a/tools/testsys/Test.toml.example
+++ b/tools/testsys/Test.toml.example
@@ -1,0 +1,119 @@
+# This is an example testing configuration for TestSys, the tool that is used to validate
+# Bottlerocket builds.
+
+# This section contains configuration details for all testing
+[test]
+
+# The repo from `Infra.toml` that should be used for Bottlerocket update images. It may be useful to
+# create a repo in `Infra.toml` that contains the infrastructure needed for testing
+repo = "default"
+
+# The registry containing alternate TestSys agent images
+testsys-images-registry = "public.ecr.aws/bottlerocket-test-system"
+
+# The URI for the EKS resource agent that should be used. An individual agent's provided URI will be
+# used even if `testsys-images-registry` is present.
+eks-resource-agent-image = "public.ecr.aws/bottlerocket-test-system/eks_resource_agent:v0.0.2"
+
+# Test Configurations
+#
+# Testing requirements tend to differ by variant and architecture. This configuration file provides
+# the ability to set values that apply generally to a broad group of similar variants, and to
+# override those values at a more granular level. For example, you can set a value for all `aws-k8s`
+# variants, then override that value for 'aws-k8s-nvidia' variants, and further override the value
+# for 'aws-k8s-nvidia'.aarch64 builds.
+#
+# The mechanism for resolving configuration values has the following order of precedence:
+#
+# '<VARIANT>'.ARCH
+# '<VARIANT>'
+# '<FAMILY>-<FLAVOR>'.ARCH
+# '<FAMILY>-<FLAVOR>'
+# '<FAMILY>'.ARCH
+# '<FAMILY>'
+# '<PLATFORM>-<FLAVOR>'.ARCH
+# '<PLATFORM>-<FLAVOR>'
+# '<PLATFORM>'.ARCH
+# '<PLATFORM>'
+#
+# For concrete example, given a variant such as `aws-k8s-1.23-nvidia` with the architecture
+# `x86_64`, configurations will have the following order of precedence:
+# ['aws-k8s-1.23-nvidia'.x86_64]
+# ['aws-k8s-1.23-nvidia']
+# ['aws-k8s-nvidia'.x86_64]
+# ['aws-k8s-nvidia']
+# ['aws-k8s'.x86_64]
+# ['aws-k8s']
+# ['aws-nvidia'.x86_64]
+# ['aws-nvidia']
+# ['aws'.x86_64]
+# ['aws']
+#
+# Note: values passed by command line argument will take precedence over those passed by environment
+# variable, and both take precedence over values set by `Test.toml`.
+
+# Example Configurations
+
+# Configuration for all variants with the `aws` platform.
+[aws]
+agent-role = "<AWS-TESTING-ROLE>"
+
+# Configuration for all nvidia AWS variants on x86_64 (platform-flavor level configuration)
+[aws-nvidia.x86_64]
+instance-type = "p3.2xlarge"
+
+# Configuration for all nvidia AWS variants on aarch64 (platform-flavor level configuration)
+[aws-nvidia.aarch64]
+instance-type = "g5g.2xlarge"
+
+# Configuration for all `aws-k8s` variants testing (family level configuration).
+[aws-k8s]
+# A single role can be assumed by agents to test all `aws-k8s` variants in a separate
+# testing account.
+agent-role = "arn:aws:iam::<ACCOUNT-ID>:role/<ROLE-NAME>"
+
+# The cluster name templating can be defined for all `aws-k8s` variants. To test on ipv4 and ipv6
+# clusters, the following templates could be used. Note: TestSys does not currently support creating
+# ipv6 clusters, so the ipv6 cluster must already exist.
+cluster-names = ["{{arch}}-{{variant}}", "{{arch}}-{{variant}}-ipv6"]
+
+# A custom conformance registry may be needed for testing if image pull reliability is a concern.
+conformance-registry = "<ACCOUNT>.dkr.ecr.cn-north-1.amazonaws.com.cn"
+
+# If testing using a kind cluster, AWS credentials need to be passed as a K8s secret.
+secrets = {"awsCreds" = "myAwsCredentials"}
+
+# Configuration for all nvidia AWS variants on x86_64 (family-flavor level configuration)
+[aws-ecs-nvidia.x86_64]
+instance-type = "p3.2xlarge"
+
+# Configuration for all nvidia AWS variants on aarch64 (family-flavor level configuration)
+[aws-ecs-nvidia.aarch64]
+instance-type = "g5g.2xlarge"
+
+# Configuration for only the `aws-k8s-1.21` variant (variant level configuration).
+["aws-k8s-1.21".aarch64]
+conformance-image = "<KUBERNETES-CONFORMANCE-IMAGE-URI>"
+
+# Configurable values:
+#
+# cluster-names: 
+#     All clusters the variant should be tested over. Cluster naming supports templated strings, and
+#     both `arch` and `variant` are provided as variables (`{{arch}}-{{variant}}`).
+#
+# instance-type:
+#     The instance type that should be used for testing.
+#
+# secrets:
+#     A map containing all secrets needed for resource creation and testing.
+#
+# agent-role:
+#     The role that should be assumed by each test and resource agent.
+#
+# conformance-image: (K8s only)
+#     Specify a custom image for conformance testing. For `aws-k8s` variants this will be used as a
+#     custom Kubernetes conformance image for Sonobuoy.
+#
+# conformance-registry: (K8s only)
+#     Specify a custom registry for conformance testing images.
+#     For `aws-k8s` variants this will be used as the Sonobuoy e2e registry.
--- a/tools/testsys/src/aws_resources.rs
+++ b/tools/testsys/src/aws_resources.rs
@@ -1,4 +1,4 @@
-use crate::run::{TestType, TestsysImages};
+use crate::run::TestType;
 use anyhow::{anyhow, Context, Result};
 use bottlerocket_types::agent_config::{
     ClusterType, CreationPolicy, Ec2Config, EcsClusterConfig, EcsTestConfig, EksClusterConfig,
@@ -10,6 +10,7 @@ use aws_sdk_ec2::Region;
 use bottlerocket_variant::Variant;
 use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
 use k8s_openapi::serde_json::Value;
+use log::debug;
 use maplit::btreemap;
 use model::constants::NAMESPACE;
 use model::{
@@ -17,17 +18,16 @@ use model::{
     TestSpec,
 };
 use std::collections::BTreeMap;
+use testsys_config::{
+    rendered_cluster_name, AwsEcsVariantConfig, AwsK8sVariantConfig, TestsysImages,
+};
 
 pub(crate) struct AwsK8s {
     pub(crate) arch: String,
     pub(crate) variant: String,
     pub(crate) region: String,
-    pub(crate) assume_role: Option<String>,
-    pub(crate) instance_type: Option<String>,
     pub(crate) ami: String,
-    pub(crate) secrets: Option<BTreeMap<String, SecretName>>,
-    pub(crate) kube_conformance_image: Option<String>,
-    pub(crate) target_cluster_name: Option<String>,
+    pub(crate) config: AwsK8sVariantConfig,
     pub(crate) tuf_repo: Option<TufRepoConfig>,
     pub(crate) starting_version: Option<String>,
     pub(crate) migrate_starting_commit: Option<String>,
@@ -43,30 +43,57 @@ impl AwsK8s {
         test: TestType,
         testsys_images: &TestsysImages,
     ) -> Result<Vec<Crd>> {
-        match test {
-            TestType::Conformance => {
-                self.sonobuoy_test_crds(testsys_images, SonobuoyMode::CertifiedConformance)
-            }
-            TestType::Quick => self.sonobuoy_test_crds(testsys_images, SonobuoyMode::Quick),
-            TestType::Migration => self.migration_test_crds(testsys_images).await,
+        let mut crds = Vec::new();
+        let target_cluster_names = if self.config.cluster_names.is_empty() {
+            debug!("No cluster names were provided using default name");
+            vec![self.default_cluster_name()]
+        } else {
+            self.config.cluster_names.clone()
+        };
+        for template_cluster_name in target_cluster_names {
+            let cluster_name = &rendered_cluster_name(
+                template_cluster_name,
+                self.kube_arch(),
+                self.kube_variant(),
+            )?;
+            crds.append(&mut match &test {
+                TestType::Conformance => self.sonobuoy_test_crds(
+                    testsys_images,
+                    SonobuoyMode::CertifiedConformance,
+                    cluster_name,
+                )?,
+                TestType::Quick => {
+                    self.sonobuoy_test_crds(testsys_images, SonobuoyMode::Quick, cluster_name)?
+                }
+                TestType::Migration => {
+                    self.migration_test_crds(cluster_name, testsys_images)
+                        .await?
+                }
+            })
         }
+        Ok(crds)
     }
 
     fn sonobuoy_test_crds(
         &self,
         testsys_images: &TestsysImages,
         sonobuoy_mode: SonobuoyMode,
+        cluster_name: &str,
     ) -> Result<Vec<Crd>> {
         let crds = vec![
-            self.eks_crd(testsys_images)?,
-            self.ec2_crd(testsys_images, None)?,
-            self.sonobuoy_crd("-test", sonobuoy_mode, None, testsys_images)?,
+            self.eks_crd(cluster_name, testsys_images)?,
+            self.ec2_crd(cluster_name, testsys_images, None)?,
+            self.sonobuoy_crd("-test", cluster_name, sonobuoy_mode, None, testsys_images)?,
         ];
         Ok(crds)
     }
 
     /// Creates `Test` crds for migration testing.
-    async fn migration_test_crds(&self, testsys_images: &TestsysImages) -> Result<Vec<Crd>> {
+    async fn migration_test_crds(
+        &self,
+        cluster_name: &str,
+        testsys_images: &TestsysImages,
+    ) -> Result<Vec<Crd>> {
         let ami = self
             .starting_image_id
             .as_ref()
@@ -81,15 +108,25 @@ impl AwsK8s {
                 .await?,
             )
             .to_string();
-        let eks = self.eks_crd(testsys_images)?;
-        let ec2 = self.ec2_crd(testsys_images, Some(ami))?;
+        let eks = self.eks_crd(cluster_name, testsys_images)?;
+        let ec2 = self.ec2_crd(cluster_name, testsys_images, Some(ami))?;
+        let instance_provider = ec2
+            .name()
+            .expect("The EC2 instance provider crd is missing a name.");
         let mut depends_on = Vec::new();
         // Start with a `quick` test to make sure instances launched properly
-        let initial = self.sonobuoy_crd("-1-initial", SonobuoyMode::Quick, None, testsys_images)?;
+        let initial = self.sonobuoy_crd(
+            "-1-initial",
+            cluster_name,
+            SonobuoyMode::Quick,
+            None,
+            testsys_images,
+        )?;
         depends_on.push(initial.name().context("Crd missing name")?);
         // Migrate instances to the target version
         let start_migrate = self.migration_crd(
-            format!("{}-2-migrate", self.cluster_name()),
+            format!("{}-2-migrate", cluster_name),
+            instance_provider.clone(),
             MigrationVersion::Migrated,
             Some(depends_on.clone()),
             testsys_images,
@@ -98,6 +135,7 @@ impl AwsK8s {
         depends_on.push(start_migrate.name().context("Crd missing name")?);
         let migrated = self.sonobuoy_crd(
             "-3-migrated",
+            cluster_name,
             SonobuoyMode::Quick,
             Some(depends_on.clone()),
             testsys_images,
@@ -105,7 +143,8 @@ impl AwsK8s {
         // Migrate instances to the starting version
         depends_on.push(migrated.name().context("Crd missing name")?);
         let end_migrate = self.migration_crd(
-            format!("{}-4-migrate", self.cluster_name()),
+            format!("{}-4-migrate", cluster_name),
+            instance_provider,
             MigrationVersion::Starting,
             Some(depends_on.clone()),
             testsys_images,
@@ -114,6 +153,7 @@ impl AwsK8s {
         depends_on.push(end_migrate.name().context("Crd missing name")?);
         let last = self.sonobuoy_crd(
             "-5-final",
+            cluster_name,
             SonobuoyMode::Quick,
             Some(depends_on.clone()),
             testsys_images,
@@ -146,13 +186,11 @@ impl AwsK8s {
     }
 
     /// Bottlerocket cluster naming convention.
-    fn cluster_name(&self) -> String {
-        self.target_cluster_name
-            .clone()
-            .unwrap_or_else(|| format!("{}-{}", self.kube_arch(), self.kube_variant()))
+    fn default_cluster_name(&self) -> String {
+        format!("{}-{}", self.kube_arch(), self.kube_variant())
     }
 
-    fn eks_crd(&self, testsys_images: &TestsysImages) -> Result<Crd> {
+    fn eks_crd(&self, cluster_name: &str, testsys_images: &TestsysImages) -> Result<Crd> {
         let cluster_version = K8sVersion::parse(
             Variant::new(&self.variant)
                 .context("The provided variant cannot be interpreted.")?
@@ -160,10 +198,9 @@ impl AwsK8s {
                 .context("aws-k8s variant is missing k8s version")?,
         )
         .map_err(|e| anyhow!(e))?;
-        let cluster_name = self.cluster_name();
         let eks_crd = Resource {
             metadata: ObjectMeta {
-                name: Some(cluster_name.clone()),
+                name: Some(cluster_name.to_string()),
                 namespace: Some(NAMESPACE.into()),
                 labels: Some(self.labels()),
                 ..Default::default()
@@ -173,23 +210,26 @@ impl AwsK8s {
                 conflicts_with: None,
                 agent: Agent {
                     name: "eks-provider".to_string(),
-                    image: testsys_images.eks_resource.clone(),
-                    pull_secret: testsys_images.secret.clone(),
+                    image: testsys_images
+                        .eks_resource_agent_image
+                        .to_owned()
+                        .expect("Missing default image for EKS resource agent"),
+                    pull_secret: testsys_images.testsys_agent_pull_secret.clone(),
                     keep_running: false,
                     timeout: None,
                     configuration: Some(
                         EksClusterConfig {
-                            cluster_name,
+                            cluster_name: cluster_name.to_string(),
                             creation_policy: Some(CreationPolicy::IfNotExists),
                             region: Some(self.region.clone()),
                             zones: None,
                             version: Some(cluster_version),
-                            assume_role: self.assume_role.clone(),
+                            assume_role: self.config.assume_role.clone(),
                         }
                         .into_map()
                         .context("Unable to convert eks config to map")?,
                     ),
-                    secrets: self.secrets.clone(),
+                    secrets: Some(self.config.secrets.clone()),
                     capabilities: None,
                 },
                 destruction_policy: DestructionPolicy::Never,
@@ -199,12 +239,16 @@ impl AwsK8s {
         Ok(Crd::Resource(eks_crd))
     }
 
-    fn ec2_crd(&self, testsys_images: &TestsysImages, override_ami: Option<String>) -> Result<Crd> {
-        let cluster_name = self.cluster_name();
+    fn ec2_crd(
+        &self,
+        cluster_name: &str,
+        testsys_images: &TestsysImages,
+        override_ami: Option<String>,
+    ) -> Result<Crd> {
         let mut ec2_config = Ec2Config {
             node_ami: override_ami.unwrap_or_else(|| self.ami.clone()),
             instance_count: Some(2),
-            instance_type: self.instance_type.clone(),
+            instance_type: self.config.instance_type.clone(),
             cluster_name: format!("${{{}.clusterName}}", cluster_name),
             region: format!("${{{}.region}}", cluster_name),
             instance_profile_arn: format!("${{{}.iamInstanceProfileArn}}", cluster_name),
@@ -214,7 +258,7 @@ impl AwsK8s {
             certificate: Some(format!("${{{}.certificate}}", cluster_name)),
             cluster_dns_ip: Some(format!("${{{}.clusterDnsIp}}", cluster_name)),
             security_groups: vec![],
-            assume_role: self.assume_role.clone(),
+            assume_role: self.config.assume_role.clone(),
         }
         .into_map()
         .context("Unable to create ec2 config")?;
@@ -233,16 +277,19 @@ impl AwsK8s {
                 ..Default::default()
             },
             spec: ResourceSpec {
-                depends_on: Some(vec![cluster_name]),
+                depends_on: Some(vec![cluster_name.to_string()]),
                 conflicts_with: None,
                 agent: Agent {
                     name: "ec2-provider".to_string(),
-                    image: testsys_images.ec2_resource.clone(),
-                    pull_secret: testsys_images.secret.clone(),
+                    image: testsys_images
+                        .ec2_resource_agent_image
+                        .to_owned()
+                        .expect("Missing default image for EC2 resource agent"),
+                    pull_secret: testsys_images.testsys_agent_pull_secret.clone(),
                     keep_running: false,
                     timeout: None,
                     configuration: Some(ec2_config),
-                    secrets: self.secrets.clone(),
+                    secrets: Some(self.config.secrets.clone()),
                     capabilities: None,
                 },
                 destruction_policy: DestructionPolicy::OnDeletion,
@@ -255,11 +302,11 @@ impl AwsK8s {
     fn sonobuoy_crd(
         &self,
         test_name_suffix: &str,
+        cluster_name: &str,
         sonobuoy_mode: SonobuoyMode,
         depends_on: Option<Vec<String>>,
         testsys_images: &TestsysImages,
     ) -> Result<Crd> {
-        let cluster_name = self.cluster_name();
         let ec2_resource_name = format!("{}-instances", cluster_name);
         let test_name = format!("{}{}", cluster_name, test_name_suffix);
         let sonobuoy = Test {
@@ -275,8 +322,11 @@ impl AwsK8s {
                 retries: Some(5),
                 agent: Agent {
                     name: "sonobuoy-test-agent".to_string(),
-                    image: testsys_images.sonobuoy_test.clone(),
-                    pull_secret: testsys_images.secret.clone(),
+                    image: testsys_images
+                        .sonobuoy_test_agent_image
+                        .to_owned()
+                        .expect("Missing default image for Sonobuoy test agent"),
+                    pull_secret: testsys_images.testsys_agent_pull_secret.clone(),
                     keep_running: true,
                     timeout: None,
                     configuration: Some(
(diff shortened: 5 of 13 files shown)


tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+# The repo from `Infra.toml` that should be used for testing. It may be
+# useful to create a repo in `Infra.toml` that contains the infrastructure
+# needed for testing
+repo = "default"
+
+# Configuration for all `aws-k8s` variants testing.
+[aws-k8s]
+# A single assume role can be used to test all `aws-k8s` variants in a separate
+# testing account.
+assume_role = "arn:aws:iam::<account_id>:role/<role_name>
```
bcressey: Does the use of this role follow the same nested approach as `Infra.toml`? Meaning:
* first we find environment creds or a profile
* then we assume the global role (if any)
* then we assume the regional role (if any)
* then we assume the test role here (if any)

(nit: I'd also prefer calling the field `role` to align with `Infra.toml`.)

Also, if we don't need to support different AWS accounts per variant under test, we could potentially add a `test_role` to the global and regional levels in `Infra.toml` instead, just to consolidate the AWS specifics.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+
+# To test sonobuoy certified conformance in other AWS partitions, sonobuoy e2e
+# images need to be copied in a registry.
+e2e_repo_registry = "<account>.dkr.ecr.cn-north-1.amazonaws.com.cn"
+
+# If testing using a kind cluster, aws credentials need to be passed as a k8s secret.
+secrets = {"awsCreds" = "myAwsCredentials"}
```
bcressey: What is the interaction between this secret and the role credentials we're meant to assume? Does the secret get automatically created and refreshed?

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+/// Configuration needed to run tests
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct TestConfig {
+    // High level configuration for TestSys
+    pub test: Option<Test>,
+}
+
+impl TestConfig {
+    /// Deserializes an TestConfig from a given path
```
arnaldo2792: Nit:

```suggestion
    /// Deserializes a TestConfig from a given path
```

Same for all occurrences

tools/testsys/src/aws_resources.rs
```diff
@@ -286,7 +319,30 @@ impl AwsK8s {
                             mode: sonobuoy_mode,
                             kubernetes_version: None,
                             kube_conformance_image: self.kube_conformance_image.clone(),
-                            e2e_repo_config_base64: None,
+                            e2e_repo_config_base64: self.e2e_repo_registry.as_ref().cloned().map(
+                                |e2e_registry| {
+                                    base64::encode(format!(
+                                        r#"buildImageRegistry: {}
+dockerGluster: {}
```
arnaldo2792: Is this intended, or should it be `dockerCluster`?

tools/testsys-config/src/lib.rs
```diff
@@ -93,6 +102,42 @@ pub struct Test {
     pub repo: Option<String>,
 }
 
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
+pub struct AwsK8sConfig {
+    /// The rle that should be assumed by all `aws-k8s` variants
```
arnaldo2792:
```suggestion
    /// The role that should be assumed by all `aws-k8s` variants
```

tools/testsys-config/src/lib.rs
```diff
@@ -93,6 +102,42 @@ pub struct Test {
...
+
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
+pub struct AwsK8sVariantConfig {
+    /// The names of all clusters this variant should be tested over. This is particularly useful
+    /// for testing Bottlerocket on ipv4 and ipv6 clusters.
+    #[serde(default)]
+    pub cluster_names: Vec<String>,
+    /// The instance type that instance types should be launched with
```
arnaldo2792: Should this be?

```suggestion
    /// The instance type that instances should be launched with
```

tools/testsys/src/run.rs
```diff
@@ -44,6 +44,10 @@ pub(crate) struct Run {
     #[clap(long, env = "AMI_INPUT")]
     ami_input: String,
 
+    /// The repo containing Sonobuoy e2e images
+    #[clap(long, env = "TESTSYS_E2E_REPO_REGISTRY")]
```
arnaldo2792: Should this be namespaced for k8s/sonobuoy? i. e. `TESTSYS_<HERE>_E2E_REPO_REGISTRY`

tools/testsys/src/run.rs
```diff
@@ -171,16 +169,70 @@ impl Run {
...
+                        arch_var_config
+                            .map(|arch_var| arch_var.cluster_names.clone())
+                            .unwrap_or_default()
+                    })
+                    .into_iter()
+                    .map(Some)
+                    .collect();
+                if target_cluster_names.len() == 0 {
+                    target_cluster_names.push(None);
+                }
```
arnaldo2792: I feel this is a little hacky, and instead the code should reflect the intent of using defaults when no cluster names were passed. With this it is less obvious that `None` actually means "use default values".

tools/testsys/src/aws_resources.rs
```diff
@@ -286,7 +319,30 @@ impl AwsK8s {
...
+                                        e2e_registry,
+                                        e2e_registry,
+                                        e2e_registry,
+                                        e2e_registry,
+                                        e2e_registry,
+                                        e2e_registry,
+                                        e2e_registry
+                                    ))
+                                },
+                            ),
```
arnaldo2792: You could do this instead:

```suggestion
dockerLibraryRegistry: {e2e_registry}
e2eRegistry: {e2e_registry}
e2eVolumeRegistry: {e2e_registry}
gcRegistry: {e2e_registry}
gcEtcdRegistry: {e2e_registry}
promoterE2eRegistry: {e2e_registry}
sigStorageRegistry: {e2e_registry}"#,
                                        e2e_registry,
```

tools/testsys-config/src/lib.rs
```diff
@@ -106,7 +108,7 @@ pub struct Test {
 #[serde(deny_unknown_fields)]
 #[serde(rename_all = "kebab-case")]
 pub struct AwsK8sConfig {
-    /// The rle that should be assumed by all `aws-k8s` variants
+    /// The role that should be assumed by all `aws-k8s` variants
```
arnaldo2792: This shouldn't be in this commit, since the commit is related to `aws-ecs`.

tools/testsys-config/src/lib.rs
```diff
@@ -138,6 +140,36 @@ pub struct AwsK8sVariantConfig {
     pub e2e_repo_registry: Option<String>,
 }
 
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
+pub struct AwsEcsConfig {
+    /// The role tat should be assumed by all `aws-k8s` variants
```
arnaldo2792: Nit

```suggestion
    /// The role that should be assumed by all `aws-ecs` variants
```

tools/testsys-config/src/lib.rs
```diff
@@ -138,6 +140,36 @@ pub struct AwsK8sVariantConfig {
...
+    pub assume_role: Option<String>,
+    /// A set of configurations for specific `aws-k8s` variants
```
arnaldo2792: `s/aws-k8s/aws-ecs` in this commit?

tools/testsys/src/run.rs
```diff
@@ -146,7 +146,7 @@ impl Run {
             .get(
                 &self
                     .repo
-                    .unwrap_or(test.repo.unwrap_or("default".to_string())),
+                    .unwrap_or_else(|| test.repo.unwrap_or_else(|| "default".to_string())),
```
arnaldo2792: If this are cleaning changes, I'd prefer them to be in the commit that introduce the change (i. e. `target_cluster_names.len() == 0` was introduced in a different commit)

tools/testsys/src/run.rs
```diff
@@ -250,15 +250,42 @@ impl Run {
                 debug!("Variant is in 'aws-ecs' family");
                 let bottlerocket_ami = ami(&self.ami_input, &region)?;
                 debug!("Using ami '{}'", bottlerocket_ami);
-                let secrets = Some(self.secret.into_iter().collect());
+                let aws_ecs = test_config.aws_ecs.as_ref();
+                let arch_var_config = aws_ecs.and_then(|aws_config| {
+                    aws_config.variant.get(&format!(
+                        "{}-{}",
+                        self.arch,
+                        self.variant.replace('.', "")
```
arnaldo2792: A note on why `.` is removed from i. e. `aws-k8s-1.21`?

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+
+# Configuration for all `aws-k8s` variants testing.
+[aws-k8s]
+# A single assume role can be used to test all `aws-k8s` variants in a separate
+# testing account.
+assume_role = "arn:aws:iam::<account_id>:role/<role_name>
+
+# To test sonobuoy certified conformance in other AWS partitions, sonobuoy e2e
+# images need to be copied in a registry.
+e2e_repo_registry = "<account>.dkr.ecr.cn-north-1.amazonaws.com.cn"
```
arnaldo2792: Maybe you can add a comment for an example of other regions?

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+
+# If testing using a kind cluster, aws credentials need to be passed as a k8s secret.
```
arnaldo2792: `s/aws/AWS`

Also an example of the format for the AWS credentials?

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+secrets = {"awsCreds" = "myAwsCredentials"}
+
+# Variant based configuration can be set using [aws-k8s.variant.<arch>-<variant>] with the '.'
+# removed from the variant.
+[aws-k8s.variant.x86_64-aws-k8s-121]
+
+# Testing may need to be tested on different clusters to test certain behavior. A test will be run
+# on each cluster specified.
```
arnaldo2792: I think this should be sufficient? 

```suggestion
# A test will run on each cluster specified in this list
```

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+# set here. This is useful while testing NVIDIA variants, since they require instance types with 
+# support for NVIDIA GPUs.
+instance_type = "p3.2xlarge"
+
+# If testing using a kind cluster, aws credentials need to be passed as a k8s 
+# secret.
+secrets = {"awsCreds" = "myAwsCredentials"}
+
+# A custom kube conformance image can be used for sonobuoy conformance testing.
+kube_conformance_image = "kube_conformance_image_uri,
```
arnaldo2792: Nit:

```suggestion
kube_conformance_image = "kube_conformance_image_uri"
```

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+repo = "default"
+
+# Configuration for all `aws-k8s` variants testing.
+[aws-k8s]
+# A single assume role can be used to test all `aws-k8s` variants in a separate
+# testing account.
+assume_role = "arn:aws:iam::<account_id>:role/<role_name>
+
+# To test sonobuoy certified conformance in other AWS partitions, sonobuoy e2e
+# images need to be copied in a registry.
```
arnaldo2792:
```suggestion
images need to be available in a registry in the targeted region
```

Same for other occurrences.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
+# This is an example testing configuration for testsys, the tool that
```
arnaldo2792: For the lack of a better place for this comment, don't we need documentation about this file and how it will be used by `cargo make test`?

arnaldo2792 (changes requested):

tools/testsys/src/aws_resources.rs
ecpullen: It is intended.

tools/testsys/Test.toml.example
bcressey: This feels problematic in that we might need to expand this along a regional matrix, and this structure isn't really set up to support that, unlike `Infra.toml`.

It might be better in `Infra.toml` instead.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+[aws-ecs]
+# A single assume role can be used to test all `aws-ecs` variants in a separate
+# testing account.
+assume_role = "arn:aws:iam::<account_id>:role/<role_name>
+
+# If testing using a kind cluster, aws credentials need to be passed as a k8s secret.
+secrets = {"awsCreds" = "myAwsCredentials"}
+
+# Variant based configuration can be set using [aws-ecs.variant.<arch>-<variant>].
+[aws-ecs.variant.x86_64-aws-ecs-1]
```
bcressey: This is unsatisfying because we've already named part of the variant (the family) so effectively this parses as:
```
{{ family }}.variant.{{ arch }}{{ variant }}
```

Which is just kind of a jumble.

I wonder if it'd be better to support composition and some sort of defined precedence order, e.g.
```
# platform, base level
# we're using hardware-like identifiers which are platform-specific
[aws.x86_64]
instance_type = ..."

# ... can be overridden by "family" specifier
[aws-k8s.x86_64]
instance_type = ..."

# ... can be overridden by the fully qualified variant name
[aws-k8s-1.21-nvidia.x86_64]
instance_type = ..."

# might be nice to have, some sort of "platform plus flavor" specifier
[aws-nvidia.x86_64]
instance_type = "..."

[aws-nvidia.aarch64]
instance_type = "..."
```

To wind that back, I'd say the minimum is:
* a platform level override, e.g. to set a default instance type for any `aws` variant
* a fully-qualified variant override, to override the instance type on a per-variant basis

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+
+# Testing may need to be tested on different clusters to test certain behavior. A test will be run
+# on each cluster specified.
+cluster_names = ["myCluster", "anotherTestCluster"]
```
bcressey: A challenge here is that there's not really a place for a platform- or family-level default, because we're naming specific clusters which are unlikely to work for all variants.

I'd recommend leaning into an opinionated, convention-over-configuration approach and supporting templates so that only exceptions need to be documented:
```suggestion
[aws]
cluster_names = ["{{ variant }}-{{ arch }}"]

[aws-k8s]
cluster_names = ["{{ variant }}-{{ arch }}", "{{ variant }}-{{ arch }}-ipv6"]
```

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+
+# The default instance type to be used is m5.large for x86_64 and m6g.large for aarch64, but can be 
+# set here. This is useful while testing NVIDIA variants, since they require instance types with 
+# support for NVIDIA GPUs.
+instance_type = "p3.2xlarge"
```
bcressey: With the suggested composition model, we wouldn't need hard-coded defaults.

tools/testsys/Test.toml.example
ecpullen: The assume role here is the role that is assumed by each testsys agent not by the testsys cli, but the nested approach is implemented in the individual agents.

tools/testsys/Test.toml.example
ecpullen: These are secrets that live in the k8s cluster. If the k8s cluster is a kind cluster, the nodes will not have aws credentials. This secret provides a way for those agents to gain aws credentials for using the aws sdk. The secret is applied to the cluster using `testsys add secret map ...`.

ecpullen:
^ Contains a large refactor of changes. `Test.toml` schema has been significantly changed to allow a more concise config.

tools/testsys/Test.toml.example
ecpullen: I opted to wait on this change until after I reworked the rest of the pr.

TESTING.md
```diff
@@ -118,6 +118,11 @@ The Bottlerocket components are found in the `testsys-bottlerocket-aws` Kubernet
 
 Now that you have the testsys cluster set up, it's time to run a Bottlerocket integration test!
 
+### Configuration
+
+There are many arguments that can be configured via environment variables with `cargo make`; however, it is possible to create a configuration file instead.
+Check out the [example config file](tools/testsys/Test.toml.example) for a sample `Test.toml` file.
```
arnaldo2792, line 124: In other places where a configuration file is used, there is documentation that explains a few fields of the configuration file, and what happens when they are set, `Test.toml` should have something similar, and you should provide snippets explaining what will happen when used.

arnaldo2792 (commented):
There is some clean up going on, I'll post this comment so that it isn't deleted with the force push.

TESTING.md
ecpullen: Do we need more documentation than the example config has?

arnaldo2792: I won't argue about this since it seems like we already link to example configuration files to explain the different sections they have. We could consider not doing so, and do what other open source projects do: provide a document to explain the purpose of each section.

ecpullen:
^ Fixes some rebasing issues and code cleanup.

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+    let mut keys = vec![variant.to_string()];
+    if let Some(flavor) = variant.variant_flavor() {
+        keys.push(format!("{}-{}", variant.family(), flavor));
+    }
+    keys.push(variant.family().to_string());
+    if let Some(flavor) = variant.variant_flavor() {
+        keys.push(format!("{}-{}", variant.platform(), flavor));
+    }
+    keys.push(variant.platform().to_string());
+    keys
```
arnaldo2792:
```suggestion
    let mut family_flavor: String = "".to_string();
    let mut platform_flavor: String = "".to_string();

    if let Some(flavor) = variant.variant_flavor {
        family_flavor = format!("{}-{}", variant.family(), flavor);
        platform_flavor = format!("{}-{}", variant.platform(), flavor);
    }

    let mut keys = vec![variant.to_string(),
      family_flavor,
      variant.family().to_string(),
      platform_flavor,
      variant.platform().to_string()
    ];

    keys.retain(|s| !s.is_empty());
    keys
```

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+            secrets: if self.secrets.is_empty() {
+                other.secrets
+            } else {
+                self.secrets
+            },
+            agent_role: self.agent_role.or(other.agent_role),
+            kube_conformance_image: self.kube_conformance_image.or(other.kube_conformance_image),
+            e2e_repo_registry: self.e2e_repo_registry.or(other.e2e_repo_registry),
+        }
+    }
```
arnaldo2792: I've seen in other places in the repo that it is preferred to assign values to variables, and then create the struct at the end of the function, instead of having inline ifs to choose the value to be set.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+repo = "default"
+# The registry containing alternate TestSys agent images
+testsys-images-registry = "public.ecr.aws/bottlerocket-test-system"
+
+# The uri for the EKS resource agent that should be used.
+# An individual agent's provided URI will be used even if `testsys-images-registry` is present.
+eks-resource-agent-image = "public.ecr.aws/bottlerocket-test-system/eks_resource_agent:v0.0.2"
+
+# The test configuration is broken up into different levels depending on the variant.
+# Each configuration's value will be selected in the following order:
```
arnaldo2792: Probably "selected" isn't the right term here, probably you should word this like "the configurations have the following precedence". I still think it is worth having an example snippet of configuration in the markdown file that explains what values will be used from each section to help users understand the precedence when two settings are provided at different levels.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+# This section contains configuration details for all testing
+[test]
+# The repo from `Infra.toml` that should be used for testing. It may be
+# useful to create a repo in `Infra.toml` that contains the infrastructure
+# needed for testing
+repo = "default"
+# The registry containing alternate TestSys agent images
+testsys-images-registry = "public.ecr.aws/bottlerocket-test-system"
+
+# The uri for the EKS resource agent that should be used.
```
arnaldo2792: Nit: `s/uri/URI`

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+# Configuration for only the `aws-k8s-1.21` variant (variant level configuration).
+["aws-k8s-1.21"]
+x86_64 = {}
+aarch64 = {kube-conformance-image = "<kube-conformance-image-uri>"}
+
+# Configurable fields:
+# cluster-names: 
+#     For `aws-k8s` variants testing will be completed over all clusters listed.
+#     For `aws-ecs` variants, only the first cluster name will be used.
+#     Cluster naming supports handlebars for `arch` and `variant` (`{{arch}}-{{variant}}`).
```
arnaldo2792:
```suggestion
#     Cluster naming supports templated strings, and both `arch` and `variant` are provided as variables (`{{arch}}-{{variant}}`).
```

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+    #[serde(default)]
+    aarch64: GenericVariantConfig,
+    #[serde(default)]
+    x86_64: GenericVariantConfig,
+    #[serde(default, flatten)]
+    config: GenericVariantConfig,
+}
+
+impl GenericConfig {
+    pub fn config<S1>(&self, arch: S1) -> GenericVariantConfig
```
arnaldo2792: I would use a different name, elsewhere in the PR I was confused since this method, attribute in the struct and the variable have the same name `vec![config.config(&arch), config.config]`. Probably `for_arch()` will be more idiomatic?

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,399 @@
...
+        Ok(test_config_path
+            .as_ref()
+            .parent()
+            .context(error::ParentSnafu {
+                path: test_config_path.as_ref(),
+            })?
+            .join("Test.lock"))
+    }
+
+    pub fn reduced_config<S1>(
```
arnaldo2792: Nit: add comments to at least public methods. There are comments in other sources of the repo that explain what each method does so it will be nice to keep the same bar.

tools/testsys/src/aws_resources.rs
```diff
@@ -81,15 +106,23 @@ impl AwsK8s {
                 .await?,
             )
             .to_string();
-        let eks = self.eks_crd(testsys_images)?;
-        let ec2 = self.ec2_crd(testsys_images, Some(ami))?;
+        let eks = self.eks_crd(cluster_name.clone(), testsys_images)?;
```
arnaldo2792: Do you need to clones `cluster_name` twice? It seems like in the method where `cluster_name` is used, the string is cloned again. Wouldn't it be sufficient to pass a reference here and clone down the road?

tools/testsys/src/aws_resources.rs
```diff
@@ -660,6 +730,22 @@ impl Migration for AwsEcs {
...
+//     /// The AMI id of the starting version for migration testing
+//     pub(crate) starting_image_id: Option<String>,
+//     /// The short commit SHA of the starting version
+//     pub(crate) migrate_starting_commit: Option<String>,
+//     /// The target version for Bottlerocket migrations
+//     pub(crate) migrate_to_version: Option<String>,
+//     /// Additional capabilities that need to be enabled on the agent's pods
+//     pub(crate) capabilities: Option<Vec<String>>,
+// }
+
```
arnaldo2792: Is this required? or can it be removed?

tools/testsys/src/run.rs
```diff
@@ -266,55 +309,47 @@ pub(crate) struct Image {
...
     )]
-    pub(crate) eks_resource: String,
+    pub(crate) eks_resource: Option<String>,
 
     /// ECS resource agent URI. If not provided the latest released resource agent will be used.
     #[clap(
         long = "ecs-resource-agent-image",
-        env = "TESTSYS_ECS_RESOURCE_AGENT_IMAGE",
-        default_value = "public.ecr.aws/bottlerocket-test-system/ecs-resource-agent:v0.0.2"
+        env = "TESTSYS_ECS_RESOURCE_AGENT_IMAGE"
```
arnaldo2792, line 326: It seems like default values were deleted.

tools/testsys/src/aws_resources.rs
```diff
@@ -173,8 +206,12 @@ impl AwsK8s {
                 conflicts_with: None,
                 agent: Agent {
                     name: "eks-provider".to_string(),
-                    image: testsys_images.eks_resource.clone(),
-                    pull_secret: testsys_images.secret.clone(),
+                    image: testsys_images
+                        .eks_resource_agent_image
+                        .as_ref()
+                        .cloned()
+                        .unwrap(),
```
arnaldo2792: I think you could add a comment here explaining why this `unwrap` is safe. It took me a while to understand that the resulting `testsys_images` object will have a default value for `eks_resource_agent_image` regardless of the configurations provided by `Test.toml`.

Please do the same for any `unwrap` you left around that is safe to keep.

tools/testsys/src/run.rs
```diff
@@ -266,55 +309,47 @@ pub(crate) struct Image {
     pub(crate) id: String,
 }
 
+/// This is a CLI parsable version of `testsys_config::TestsysImages`
 #[derive(Debug, Parser)]
 pub(crate) struct TestsysImages {
```
arnaldo2792, line 315: `CliConfig` has a 1 to 1 mapping with a different struct name (`GenerictVariantConfig`), would it make sense to do the same for this struct and either rename this or the counterpart in `testsys_config`?

arnaldo2792 (changes requested):

tools/testsys/src/aws_resources.rs
ecpullen: That was work from another branch. Removing.

tools/testsys/src/run.rs
ecpullen: That is correct and expected. The images can be overwritten in the config file.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+
+# Configuration for only the `aws-k8s-1.21` variant (variant level configuration).
+["aws-k8s-1.21"]
+x86_64 = {}
+aarch64 = {kube-conformance-image = "<kube-conformance-image-uri>"}
+
+# Configurable fields:
+# cluster-names: 
+#     For `aws-k8s` variants testing will be completed over all clusters listed.
+#     For `aws-ecs` variants, only the first cluster name will be used.
```
bcressey: Why the different behavior? I don't think it's useful today but I suppose there could be cluster-level differences that we'd want to test in the future, and it seems harmless to allow it

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+instance-type = "p3.2xlarge"
+
+# Configuration for all nvidia AWS variants on aarch64 (family-flavor level configuration)
+[aws-ecs-nvidia.aarch64]
+instance-type = "g5g.2xlarge"
+
+# Configuration for only the `aws-k8s-1.21` variant (variant level configuration).
+["aws-k8s-1.21"]
+x86_64 = {}
+aarch64 = {kube-conformance-image = "<kube-conformance-image-uri>"}
```
bcressey: nit: is the empty table needed?

I would also keep the same style for all the override tables, e.g. 
```suggestion
["aws-k8s-1.21".aarch64]
kube-conformance-image = "<kube-conformance-image-uri>"
```

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+
+# Configurable fields:
+# cluster-names:
```
bcressey, line 100: Can you include an example above of setting cluster names - maybe for `aws-k8s` with and without `-ipv6`?

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+# The test configuration is broken up into different levels depending on the variant.
+# Each configuration's value will be selected in the following order:
+# Arguments from the command line
+# Environment variables from the command line
+# '{variant}' configuration in `Test.toml`
+# '{family}-{flavor}' configuration in `Test.toml`
+# '{family}' configuration in `Test.toml`
+# '{platform}-{flavor}' configuration in `Test.toml`
+# '{platform}' configuration in `Test.toml`
+# TestSys default values
```
bcressey: It might be easier to read the example if you provide each of them in priority order (from `variant` to `platform`) with a comment above each explaining which is which.

I'd also recommend picking a convention and sticking to it throughout, and specifically avoiding `{}` because of the overlap with handlebars.

Maybe `<USER-SPECIFIED-SHOUTING-KEBAB-CASE>` - there are some inconsistencies with `-` vs `_` (e.g. `aws-testing-role` but `account_id`).

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+
+# Configuration for all AWS variants (platform level configuration).
+[aws]
+agent-role = "<aws-testing-role>"
+
+# Configuration for all nvidia AWS variants on x86_64 (platform-flavor level configuration)
+[aws-nvidia.x86_64]
+instance-type = "p3.2xlarge"
```
bcressey, line 63: You had mentioned maybe allowing a form like this:
```
[aws-nvidia]
instance-type = "p3.2xlarge"
```

Did you change your mind?

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+instance-type = "g5g.2xlarge"
+
+# Configuration for all `aws-k8s` variants testing (family level configuration).
+[aws-k8s]
+# A single role can be assumed by agents to test all `aws-k8s` variants in a separate
+# testing account.
+agent-role = "arn:aws:iam::<account_id>:role/<role_name>"
+
+# To test sonobuoy certified conformance in other AWS partitions, sonobuoy e2e
+# images need to be available in a registry in the targeted region (cn-north-1, us-gov-west-1).
```
bcressey: nit: this is not technically true if pulling images from ECR Public in GovCloud, and if reliability is not a concern then perhaps not for China either.

It might be better to document it as a way to set a custom conformance image, without reference to specific AWS partitions.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,78 @@
...
+#     For `aws-ecs` variants, only the first cluster name will be used.
+#     Cluster naming supports handlebars for `arch` and `variant` (`{{arch}}-{{variant}}`).
+# secrets:
+#     A map containing all secrets needed for resource creation and testing.
+# agent-role:
+#     The role that should be assumed by each test and resource agent.
+# kube-conformance-image: (K8s only)
+#     Specify a custom Kubernetes conformance image for E2E sonobuoy testing.
+# e2e-repo-registry: (K8s only)
+#     Configure the registry used for Sonobuoy E2E testing images.
```
bcressey: Should we make these fields more general, e.g. `conformance-image` and `conformance-registry`?

tools/testsys/Test.toml.example
ecpullen: We do allow that form; however we typically use different instance types based on the architecture.

ecpullen:
^ Address comments
* Rework `config_keys` vec creation to be more idiomatic
* Added comment to `config_keys` function
* Removed commented `MigrationsConfig`
* `merge` assign some variables before struct creation
* 'uri' -> 'URI'
* Removed unnecessary x86_64 table from 'Test.toml.example'
* 'selected' -> 'precedence'
* Provided configuration precedence for `aws-k8s-1.23-nvidia` variant on `x86_64`
* Provided a sample `cluster-names` for ipv6 and ipv4 testing with a disclaimer that ipv6 cluster creation is not yet supported.
* Changed `e2e-repo-registry` to `conformance-registry` to make it more general.
* Changed `kube-conformance-image` to `conformance-image` to make it more general.
* Reword cluster-names templated string comment.
* Comments for `unwrap()`'s
* Convert `cluster_name` to be `&str` to reduce cloning
* Added `debug` log if no cluster templates were provided
* Fixed inconsistencies with in user values


Still need to support multiple clusters with ecs.

ecpullen:
^ Enables ecs testing on multiple clusters and fixes an error in `TestsysImages::new()`
Tested with `cargo make -e BUILDSYS_VARIANT=aws-ecs-1 test` with the following configuration.
```yaml
[test]
repo="default"

[aws-ecs]
cluster-names = ["{{arch}}-{{variant}}", "{{arch}}-{{variant}}-dup"]
```

ecpullen:
^ Resolve merge conflict.

ecpullen:
^ Add small example config snippet to `TESTING.md` for instance type configuration.

ecpullen:
^ Update `Test.toml.example`'s `aws-k8s-1.23-nvidia` hierarchy example to also show what configuration level is being represented.

ecpullen:
^ Fixed 2 outdated comments related to testing on multiple ECS clusters.

stmcginnis (approved):
Looks fine to me, but would be good to get feedback from Ben and Arnaldo since they had previous comments.

Can you resolve the conversations that were addressed so it's clear what is done and what might still be an open question?

tools/testsys/src/run.rs
ecpullen: I don't think it quite works out, because the images available from the config can be split into 2 separate sections, controller and agents. I may add the controller image to the config file in a later pr.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,97 @@
...
+# Configuration for all nvidia AWS variants on aarch64 (family-flavor level configuration)
+[aws-ecs-nvidia.aarch64]
+instance-type = "g5g.2xlarge"
+
+# Configuration for only the `aws-k8s-1.21` variant (variant level configuration).
+["aws-k8s-1.21".aarch64]
+conformance-image = "<KUBERNETES-CONFORMANCE-IMAGE-URI>"
+
+# Configurable fields:
+# cluster-names:
```
bcressey: Should `instance-type` also be listed here? It might also make sense to put this before the "Configuration for all AWS variants" section, so that people see the possible values before the sample configurations.

tools/testsys/src/run.rs
```diff
@@ -101,16 +83,65 @@ pub(crate) struct Run {
...
-            None
-        } else {
-            Some(self.secret.into_iter().collect())
-        };
+
+        // If a lock file exists, use that, otherwise use Test.toml or default
+        let test_config = TestConfig::from_path_or_lock(&self.test_config_path, true)
+            .context("Unable to read test config")?;
+
+        let test = test_config.test.as_ref().cloned().unwrap_or_default();
```
bcressey: nit: I'd prefer a more specific variable name, like
```suggestion
        let test_opts = test_config.test.as_ref().cloned().unwrap_or_default();
```

tools/testsys/src/run.rs
```diff
@@ -150,21 +186,31 @@ impl Run {
...
+        let images = vec![
+            Some(self.agent_images.into()),
+            Some(test.testsys_images),
+            test.testsys_image_registry
+                .map(testsys_config::TestsysImages::new),
+            Some(testsys_config::TestsysImages::public_images()),
+        ]
+        .into_iter()
+        .flatten()
+        .fold(Default::default(), testsys_config::TestsysImages::merge);
```
bcressey: The difference between `testsys_config` (a module) and `test_config` (a structure) is pretty key to understanding this, but I had them conflated in my mind for a while.

Maybe something like this to get rid of the ambiguity:
```
let (cfg_test_opts, cfg_variant_overrides) = TestConfig::from_path_or_lock(&self.test_config_path, true)
            .context("Unable to read test config")?.split();
```

Where `split` takes `Self` and destructures it into a tuple, so we don't need to keep a `TestConfig` around.

tools/testsys/src/aws_resources.rs
```diff
@@ -173,23 +208,29 @@ impl AwsK8s {
...
                     name: "eks-provider".to_string(),
-                    image: testsys_images.eks_resource.clone(),
-                    pull_secret: testsys_images.secret.clone(),
+                    image: testsys_images
+                        .eks_resource_agent_image
+                        .as_ref()
+                        .cloned()
+                        // This is a safe unwrap because `testsys_images` has a default value set
+                        // for `eks_resource_agent`.
+                        .unwrap(),
```
bcressey: nit: this is the same info but more useful to a developer that sees the theoretical panic:
```suggestion
                        .expect("missing default image for eks resource agent"),
```

tools/testsys/src/aws_resources.rs
```diff
@@ -390,25 +452,54 @@ impl AwsEcs {
         test: TestType,
         testsys_images: &TestsysImages,
```
bcressey, line 440: Not for this PR - it would help to split out `AwsEcs` and `AwsK8s` into separate modules. This one is getting pretty long.

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,416 @@
...
+
+    /// Deserializes a TestConfig from a Test.lock file at a given path
+    pub fn from_lock_path<P>(path: P) -> Result<Self>
+    where
+        P: AsRef<Path>,
+    {
+        let path = path.as_ref();
+        let test_config_str = fs::read_to_string(path).context(error::FileSnafu { path })?;
+        serde_yaml::from_str(&test_config_str).context(error::InvalidLockSnafu { path })
+    }
```
bcressey: Do we need to support `Test.lock`?

IIRC for `Infra.toml` the idea was that `Infra.lock` would hold the concrete resource IDs as opposed to abstract values. If we don't have a similar need here, I'd just leave it out for now.

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,416 @@
...
+    pub cluster_names: Vec<String>,
+    /// The instance type that instances should be launched with
+    pub instance_type: Option<String>,
+    /// The secrets needed by the agents
+    pub secrets: BTreeMap<String, SecretName>,
+    /// The role that should be assumed for this particular variant
+    pub assume_role: Option<String>,
+}
+
+/// Create a vec if relevant keys for this variant ordered from most specific to least specific.
```
bcressey: Nice, this is very lightweight for the power it offers.

```suggestion
/// Create a vec of relevant keys for this variant ordered from most specific to least specific.
```

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,416 @@
...
+            other.cluster_names
+        } else {
+            self.cluster_names
+        };
+
+        let secrets = if self.secrets.is_empty() {
+            other.secrets
+        } else {
+            self.secrets
+        };
```
bcressey, line 218: It is sort of tempting to have `cluster_names` and `secrets` be `Option`, and to have empty lists deserialize into `None`, just so all of the merge logic is the same. But I'm not sure it's worth it 😀

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,416 @@
...
+    pub ecs_test_agent_image: Option<String>,
+    pub migration_test_agent_image: Option<String>,
+    pub testsys_agent_pull_secret: Option<String>,
+}
+
+const AGENT_VERSION: &str = "v0.0.2";
+
+impl TestsysImages {
+    /// Create an images config for a specific registry.
+    pub fn new<S1>(registry: S1) -> Self
```
bcressey: nit:
```suggestion
    pub fn new<S>(registry: S) -> Self
```

tools/testsys-config/src/lib.rs
ecpullen: Ok. I just tried to replicate `Infra.toml`. I can take it out.

tools/testsys/src/run.rs
ecpullen: I like that `reduced_config` is a function of `TestConfig` splitting would remove that feature since the other configs are just a `HashMap`.

ecpullen:
^
* Adds `instance-types` to example config file
* Fixed typos in lib.rs
* Remove lock support for `Test.toml`
* Replace `unwrap`s with `expect`

ecpullen:
^ Change generic `S1` to `S`.

ecpullen:
^ Fix `cargo fmt`

bcressey (approved):

ecpullen:
^ rebase to develop to fix merge conflict

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+    /// for testing Bottlerocket on ipv4 and ipv6 clusters.
+    pub cluster_names: Vec<String>,
+    /// The instance type that instances should be launched with
+    pub instance_type: Option<String>,
+    /// The secrets needed by the agents
+    pub secrets: BTreeMap<String, SecretName>,
+    /// The role that should be assumed for this particular variant
+    pub assume_role: Option<String>,
+    /// The sonobuoy conformance image that should be used for this variant
+    pub kube_conformance_image: Option<String>,
```
etungsten: nit: to clarify a bit more
```suggestion
    /// The kubernetes conformance image that should be used for this variant
    pub kube_conformance_image: Option<String>,
```

tools/testsys/src/aws_resources.rs
```diff
@@ -43,30 +43,57 @@ impl AwsK8s {
...
+            )?;
+            crds.append(&mut match &test {
+                TestType::Conformance => self.sonobuoy_test_crds(
+                    testsys_images,
+                    SonobuoyMode::CertifiedConformance,
+                    cluster_name,
+                )?,
+                TestType::Quick => {
+                    self.sonobuoy_test_crds(testsys_images, SonobuoyMode::Quick, cluster_name)?
+                }
```
etungsten, line 67: Seeing this makes me wonder how expanding this list would look like. For example, if I want to run a different sonobuoy test mode like `conformance-lite`, do I add to the `TestType` enum? But `TestType` enum needs to be generic across all tests including ECS testing where `conformance-lite` doesn't make sense.

tools/testsys/src/aws_resources.rs
```diff
@@ -730,13 +832,17 @@ trait Migration {
...
                 depends_on,
                 retries: None,
                 agent: Agent {
                     name: "migration-test-agent".to_string(),
-                    image: testsys_images.migration_test.to_string(),
-                    pull_secret: testsys_images.secret.clone(),
+                    image: testsys_images
+                        .migration_test_agent_image
+                        .as_ref()
+                        .cloned()
```
etungsten: This can probably be simplified to just a `cloned()` or `to_owned()`

tools/testsys/src/aws_resources.rs
```diff
@@ -497,21 +591,25 @@ impl AwsEcs {
                 conflicts_with: None,
                 agent: Agent {
                     name: "ecs-provider".to_string(),
-                    image: testsys_images.ecs_resource.clone(),
-                    pull_secret: testsys_images.secret.clone(),
+                    image: testsys_images
+                        .ecs_resource_agent_image
+                        .as_ref()
+                        .cloned()
```
etungsten: Is the `as_ref()` is necessary?
```suggestion
                        .cloned()
```

tools/testsys/src/aws_resources.rs
```diff
@@ -549,16 +651,20 @@ impl AwsEcs {
...
+                depends_on: Some(vec![cluster_name.to_string()]),
                 conflicts_with: None,
                 agent: Agent {
                     name: "ec2-provider".to_string(),
-                    image: testsys_images.ec2_resource.clone(),
-                    pull_secret: testsys_images.secret.clone(),
+                    image: testsys_images
+                        .ec2_resource_agent_image
+                        .as_ref()
+                        .cloned()
```
etungsten: Same here maybe?
```suggestion
                        .cloned()
```

tools/testsys/src/run.rs
```diff
@@ -101,16 +83,65 @@ pub(crate) struct Run {
...
+    conformance_image: Option<String>,
+
+    /// The role that should be assumed by the agents
+    #[clap(long, env = "TESTSYS_ASSUME_ROLE")]
+    assume_role: Option<String>,
+
+    /// Specify the instance type that should be used. This is only applicable for aws-* variants.
+    /// It can be omitted for non-aws variants and can be omitted to use default instance types.
+    #[clap(long, env = "TESTSYS_INSTANCE_TYPE")]
+    instance_type: Option<String>,
```
etungsten, line 113: This needs to be updated once https://github.com/bottlerocket-os/bottlerocket-test-system/pull/602 merges I think

etungsten (commented):
[testsys.kubeconfig.eksctl.lock](https://github.com/bottlerocket-os/bottlerocket/pull/2393/files#diff-ea6827e4ef448d2a22b7c301c146c7419eb1a07c55a187c9f03cff1ff372fcd9) shouldn't be part of the PR.

Can you also give more details on how the testing was conducted, e.g how `test.toml` was set up and what the outputs look like?

webern:
Does `testsys.kubeconfig.eksctl.lock` need to be deleted and .gitignored?

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,99 @@
...
+testsys-images-registry = "public.ecr.aws/bottlerocket-test-system"
+
+# The URI for the EKS resource agent that should be used.
+# An individual agent's provided URI will be used even if `testsys-images-registry` is present.
+eks-resource-agent-image = "public.ecr.aws/bottlerocket-test-system/eks_resource_agent:v0.0.2"
+
+# The test configuration is broken up into different levels depending on the variant.
+# Configurations will have the following precedence:
+# Arguments from the command line
+# Environment variables from the command line
```
webern:
```suggestion
# Arguments passed at the command line will take precedence over environment variables.
# Configurations will have the following resolution precedence:
```

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,99 @@
...
+# '<VARIANT>' configuration in `Test.toml`
+# '<FAMILY>-<FLAVOR>' configuration in `Test.toml`
+# '<FAMILY>' configuration in `Test.toml`
+# '<PLATFORM>-<FLAVOR>' configuration in `Test.toml`
```
webern: Possible dumb question - How do you distinguish between a variant and `<PLATFORM>-<FLAVOR>`. For example, if my platform is `p` and my flavor is `f`, then `p-f` could be a platform-flavor or it could be a variant name. Note, I haven't read the code yet. I'm just thinking about a case where I have a platform `foo` and a flavor `bar`, e.g. in the variant `foo-x-1.1-bar`, but I also have a variant named `foo-bar`.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,99 @@
...
+# '<PLATFORM>' configuration in `Test.toml`
+# TestSys default values
+
+# For the `aws-k8s-1.23-nvidia` variant on `x86_64`, configurations will have the following precedence:
```
webern: In the previous section you described variant, family, platform etc, without mentioning ARCH. Now we have ARCH without having introduced the concept. I'm left unsure how ARCH figures into the precedence scheme.

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,99 @@
...
+# Arguments from the command line
```
webern: I don't quite understand why it says this here. In the previous section you talked about arguments from the command line taking precedence over environment variables. In this section we have the sentence fragment:

```text
# Arguments from the command line
```

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,99 @@
...
+# testing account.
+agent-role = "arn:aws:iam::<ACCOUNT-ID>:role/<ROLE-NAME>"
+
+# The cluster name templating can be defined for all `aws-k8s` variants.
+# To test on ipv4 and ipv6 clusters, the following templates could be used.
+# Note: TestSys does not currently support creating ipv6 clusters, so the ipv6 cluster must already exist.
+cluster-names = ["{{arch}}-{{variant}}", "{{arch}}-{{variant}}-ipv6"]
+
+# A custom conformance registry may be needed for testing if image pull reliability is a concern.
+conformance-registry = "<ACCOUNT>.dkr.ecr.cn-north-1.amazonaws.com.cn"
```
webern, line 81: Does every key in these "configuration" sections correspond to a testsys command line argument? If so, could that be more explicitly described in the overview at the beginning?

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,99 @@
...
+
+# Configuration for all nvidia AWS variants on aarch64 (family-flavor level configuration)
+[aws-ecs-nvidia.aarch64]
+instance-type = "g5g.2xlarge"
+
+# Configuration for only the `aws-k8s-1.21` variant (variant level configuration).
+["aws-k8s-1.21".aarch64]
+conformance-image = "<KUBERNETES-CONFORMANCE-IMAGE-URI>"
+
+# Configurable fields:
```
webern: What is a "field"?

webern (changes requested):
I think if you land go to Test.toml.example to try and figure out what Test.toml is for and how it is used, you might be left confused. I've left a few comments that hopefully show what I mean.

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+        P: AsRef<Path>,
+    {
+        let path = path.as_ref();
+        let test_config_str = fs::read_to_string(path).context(error::FileSnafu { path })?;
+        toml::from_str(&test_config_str).context(error::InvalidTomlSnafu { path })
+    }
+
+    /// Deserializes a TestConfig from a given path, if it exists, otherwise builds a default
+    /// config
+    pub fn from_path_or_default<P>(path: P) -> Result<Self>
```
webern, line 39: It might be helpful for a user to know (via stdout) whether they are getting their file loaded or getting a default. For example if I have a typo in the path that I pass `/home/me/bottlerocket-teesting/Test.toml`, I could be left scratching my head as to why my configuration file is not working.

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+            Self::from_path(path)
+        } else {
+            Ok(Self::default())
+        }
+    }
+
+    /// Create a single config for the `variant` and `arch` from this config. `starting_config` is
+    /// used as the config with the highest precedence. If no starting config is available,
+    /// `Default::default()` can be used.
+    pub fn reduced_config<S>(
```
webern, line 59: `resolve_config`?

Can you better explain the purpose of the `starting_config` argument?

"Create a single config [...] from this config" is unclear.

Maybe it should be an `Option<GenericVariantConfig>`. It's not clear how to call the function if "no starting_config is available", but maybe what you're saying is that it is an optional argument.

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+    Ok(cluster_template.render(
+        "cluster_name",
+        &btreemap! {"arch".to_string() => arch.into(), "variant".to_string() => variant.into()},
+    )?)
+}
+
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
+pub struct TestsysImages {
```
webern, line 272: We're hard-coding a lot of things for our "supported" variants. Perhaps this is fine and to be dealt with later when we work on out-of-tree builds.

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+        family_flavor,
+        variant.family().to_string(),
+        platform_flavor,
+        variant.platform().to_string(),
+    ]
+}
+
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct GenericConfig {
```
webern, line 159: Explain

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+            "aarch64" => self.aarch64.clone(),
+            _ => Default::default(),
+        }
+    }
+}
+
+#[derive(Debug, Default, Deserialize, Serialize, PartialEq, Eq, Clone)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
+pub struct GenericVariantConfig {
```
webern, line 187: Explain

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "kebab-case")]
+pub struct TestsysImages {
+    pub eks_resource_agent_image: Option<String>,
+    pub ecs_resource_agent_image: Option<String>,
+    pub ec2_resource_agent_image: Option<String>,
+    pub sonobuoy_test_agent_image: Option<String>,
+    pub ecs_test_agent_image: Option<String>,
+    pub migration_test_agent_image: Option<String>,
+    pub testsys_agent_pull_secret: Option<String>,
```
webern, line 279: Can I have one agent container that is published in a different place? Or would I need to publish all of these containers to the same container registry?

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+    ) -> GenericVariantConfig
+    where
+        S: Into<String>,
+    {
+        let arch = arch.into();
+        config_keys(variant)
+            .into_iter()
+            .filter_map(|key| self.configs.get(&key).cloned())
+            .flat_map(|config| vec![config.for_arch(&arch), config.config])
+            .fold(starting_config, GenericVariantConfig::merge)
```
webern: Since is the core of the matching algorithm, perhaps a description of how it works could go here in case one's (ahem) functional programming in Rust skills are lacking.

webern (changes requested):

tools/testsys/src/aws_resources.rs
```diff
@@ -285,14 +336,31 @@ impl AwsK8s {
...
+                            kube_conformance_image: self.config.kube_conformance_image.clone(),
+                            e2e_repo_config_base64: self
+                                .config
+                                .e2e_repo_registry
+                                .as_ref()
+                                .cloned()
+                                .map(|e2e_registry| {
+                                    base64::encode(format!(
+                                        r#"buildImageRegistry: {e2e_registry}
+dockerGluster: {e2e_registry}
```
webern, line 343: Is `dockerGluster` correct?

tools/testsys/src/aws_resources.rs
```diff
@@ -10,24 +10,24 @@ use aws_sdk_ec2::Region;
...
 use model::{
     Agent, Configuration, Crd, DestructionPolicy, Resource, ResourceSpec, SecretName, Test,
     TestSpec,
 };
 use std::collections::BTreeMap;
+use testsys_config::{
+    rendered_cluster_name, AwsEcsVariantConfig, AwsK8sVariantConfig, TestsysImages,
+};
 
 pub(crate) struct AwsK8s {
```
webern, line 25: I'll leave this file's large diff alone, but I'm flagging it as another place that we will have to think about extensibility for out-of-tree variants.

tools/testsys/src/run.rs
ecpullen: It does not. We pin to a released version of testsys.

tools/testsys/Test.toml.example
ecpullen: If that happens then it will treat the `foo-bar` table as being variant level configuration. That is an edge case that I don't think we need to document.

tools/testsys-config/src/lib.rs
ecpullen: This issue the same loading behavior as `Infra.toml` if we are going to change one we should probably change both.

tools/testsys-config/src/lib.rs
ecpullen: We decided that the only acceptable fields in a config are fields that we know about and implement.

tools/testsys-config/src/lib.rs
ecpullen: They can be in different ones. There is a note about it in `Test.toml.example`.

tools/testsys/src/aws_resources.rs
ecpullen: Yes.

tools/testsys/src/aws_resources.rs
ecpullen: We will add the `conformance-lite` test type and then report that it is unavailable for the other variants if a user tries to use them.

tools/testsys/Test.toml.example
ecpullen: That is a rule with come cli's but that is not guaranteed here, for instance `testsys` can only use a single cluster, but multiple clusters can be described in `Test.toml`.

tools/testsys-config/src/lib.rs
ecpullen: The technical term for the process is `reduce` so I'm not sure we want to change that. I have made `starting_config` an option.

ecpullen:
^ 
* Added `testsys.kubeconfig.eksctl.lock` to `.gitignore`
* Change comment for `kube_conformance_image`
* Convert `.as_ref().cloned()` to `to_owned()`
*  Reworded precedence comment in `Test.toml.example`
* Added a note describing architecture configuration in `Test.toml.example`
* Removed `arguments from command line` from precedence lists and added a not about them instead.
*  Changed `Configurable fields` to `Configurable values`
* Added doc comment for `reduced_config` function
* Added comments documenting the transformations occurring in `reduced_config`
* Added doc comment for `GenericConfig` and `GenericVariantConfig`

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,103 @@
+# This is an example testing configuration for TestSys, the tool that
+# is used to validate Bottlerocket builds.
+
+# This section contains configuration details for all testing
+[test]
+# The repo from `Infra.toml` that should be used for testing. It may be
+# useful to create a repo in `Infra.toml` that contains the infrastructure
+# needed for testing
+repo = "default"
+# The registry containing alternate TestSys agent images
```
webern, line 11: Is "alternate" correct in this sentence?

```suggestion
# The registry containing TestSys agent images
```

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,103 @@
...
+eks-resource-agent-image = "public.ecr.aws/bottlerocket-test-system/eks_resource_agent:v0.0.2"
+
+# Arguments passed at the command line will take precedence over environment variables and any
+# values specified in the configuration file.
+# Configurations will have the following resolution precedence:
+# '<VARIANT>' configuration in `Test.toml`
+# '<FAMILY>-<FLAVOR>' configuration in `Test.toml`
+# '<FAMILY>' configuration in `Test.toml`
+# '<PLATFORM>-<FLAVOR>' configuration in `Test.toml`
+# '<PLATFORM>' configuration in `Test.toml`
```
webern: I'm not sure if it should be demonstrated in this fashion, but do I understand this correctly?

(Also, is " configuration in `Test.toml`" actually required after each of these? If not, i.e. if it's obvious, then it would be less cluttered without that repeated suffix)

```suggestion
# '<VARIANT>'.ARCH
# '<VARIANT>'
# '<FAMILY>-<FLAVOR>'.ARCH
# '<FAMILY>-<FLAVOR>'
# '<FAMILY>'.ARCH
# '<FAMILY>'
# '<PLATFORM>-<FLAVOR>'.ARCH
# '<PLATFORM>-<FLAVOR>'
# '<PLATFORM>'.ARCH
# '<PLATFORM>'
```

tools/testsys-config/src/lib.rs
webern: Hmmm, it should probably warn me there as well. I thought I remembered seeing a warning about a default Infra.toml. That might be because we generate one earlier in the build process when one does not exist.

.gitignore
```diff
@@ -6,7 +6,9 @@
 /.gomodcache
 /html
 /Infra.toml
+/Test.toml
 /testsys.kubeconfig
+/testsys.kubeconfig.eksctl.lock
```
etungsten: I don't think we should add this here. How is this lock file generated? From my experience, it's only when I create a new cluster with `eksctl`. This lock file only shows up under pretty specific use-cases (i.e. using eksctl to create the testsys cluster in the root dir of this repo).

edit: I believe this is a bug in eksctl that eventually will be fixed: https://github.com/weaveworks/eksctl/issues/5698/. In any case, I don't think an eksctl-specific lock file should be added here

webern: >  I don't think an eksctl-specific lock file should be added here

I've never really understood the aversion to putting things in .gitignore.

TESTING.md
```diff
@@ -118,6 +118,24 @@ The Bottlerocket components are found in the `testsys-bottlerocket-aws` Kubernet
...
```
webern, line 124: I stumbled on this a little bit because I was thinking about a configuration file for `cargo make` (like an `.env` file). I think it could be said a bit more directly. Here's an attempt:

```suggestion
Tests can be configured either by passing environment variables to `cargo make`, or by creating a `Test.toml` configuration file.
Check out the [example config file](tools/testsys/Test.toml.example) for a sample `Test.toml` file.
```

TESTING.md
```diff
@@ -118,6 +118,24 @@ The Bottlerocket components are found in the `testsys-bottlerocket-aws` Kubernet
...
+
+To support different instance types on regular `aws-k8s` variants and `nvidia` `aws-k8s` variants, the following configuration could be used.
```
webern: Sorry to be so nit picky with the documentation... something about the word "regular" doesn't sit right. Would it work to let the example speak for itself?

(Also, this PR confirms my instinct that we want a newline to precede code blocks https://github.com/bottlerocket-os/bottlerocket/pull/2283/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L490)

```suggestion
For example, we can specify different instance types based on variant requirements:

```

tools/testsys-config/src/lib.rs
```diff
@@ -0,0 +1,382 @@
...
+    pub eks_resource_agent_image: Option<String>,
+    pub ecs_resource_agent_image: Option<String>,
+    pub ec2_resource_agent_image: Option<String>,
+    pub sonobuoy_test_agent_image: Option<String>,
+    pub ecs_test_agent_image: Option<String>,
+    pub migration_test_agent_image: Option<String>,
+    pub testsys_agent_pull_secret: Option<String>,
+}
+
+const AGENT_VERSION: &str = "v0.0.2";
```
webern, line 282: We shouldn't block on this, but I'm noticing that the version string `0.0.2` exists in a few different places. I wonder if we can consolidate it later.

Edit: I think this is related https://github.com/bottlerocket-os/bottlerocket-test-system/issues/542

tools/testsys/Test.toml.example
```diff
@@ -0,0 +1,103 @@
+# This is an example testing configuration for TestSys, the tool that
```
webern: I tried to see if I could make any improvements, and it was easier to do in an editor than in the PR. Some things I tried:
- The wrapping seemed off, so I set a ruler at 100 and re-wrapped it.
- I re-wrote some things because it felt like some context was missing. 
- I added some newlines and removed some repeated text to let it breathe
- It felt like architecture was missing from the explanation of precedence (as noted in another comment), so I tried incorporating it.

I don't want to block this important PR, but did want to take a stab improvement so spent some time on it.

```toml
# This is an example testing configuration for TestSys, the tool that is used to validate
# Bottlerocket builds.

# This section contains configuration details for all testing
[test]

# The repo from `Infra.toml` that should be used for Bottlerocket update images. It may be useful to
# create a repo in `Infra.toml` that contains the infrastructure needed for testing
repo = "default"

# The registry containing alternate TestSys agent images
testsys-images-registry = "public.ecr.aws/bottlerocket-test-system"

# The URI for the EKS resource agent that should be used. An individual agent's provided URI will be
# used even if `testsys-images-registry` is present.
eks-resource-agent-image = "public.ecr.aws/bottlerocket-test-system/eks_resource_agent:v0.0.2"

# Test Configurations
#
# Testing requirements tend to differ by variant and architecture. This configuration file provides
# the ability to set values that apply generally to a broad group of similar variants, and to
# override those values at a more granular level. For example, you can set a value for all `aws-k8s`
# variants, then override that value for 'aws-k8s-nvidia' variants, and further override the value
# for 'aws-k8s-nvidia'.aarch64 builds.
#
# The mechanism for resolving configuration values has the following order of precedence:
#
# '<VARIANT>'.ARCH
# '<VARIANT>'
# '<FAMILY>-<FLAVOR>'.ARCH
# '<FAMILY>-<FLAVOR>'
# '<FAMILY>'.ARCH
# '<FAMILY>'
# '<PLATFORM>-<FLAVOR>'.ARCH
# '<PLATFORM>-<FLAVOR>'
# '<PLATFORM>'.ARCH
# '<PLATFORM>'
#
# For concrete example, given a variant such as `aws-k8s-1.23-nvidia` with the architecture
# `x86_64`, configurations will have the following order of precedence:
# ['aws-k8s-1.23-nvidia'.x86_64]
# ['aws-k8s-1.23-nvidia']
# ['aws-k8s-nvidia'.x86_64]
# ['aws-k8s-nvidia']
# ['aws-k8s'.x86_64]
# ['aws-k8s']
# ['aws-nvidia'.x86_64]
# ['aws-nvidia']
# ['aws'.x86_64]
# ['aws']
#
# Note: values passed by command line argument will take precedence over those passed by environment
# variable, and both take precedence over values set by `Test.toml`.

# Example Configurations

# Configuration for all variants with the `aws` platform.
[aws]
agent-role = "<AWS-TESTING-ROLE>"

# Configuration for all nvidia AWS variants on x86_64 (platform-flavor level configuration)
[aws-nvidia.x86_64]
instance-type = "p3.2xlarge"

# Configuration for all nvidia AWS variants on aarch64 (platform-flavor level configuration)
[aws-nvidia.aarch64]
instance-type = "g5g.2xlarge"

# Configuration for all `aws-k8s` variants testing (family level configuration).
[aws-k8s]
# A single role can be assumed by agents to test all `aws-k8s` variants in a separate
# testing account.
agent-role = "arn:aws:iam::<ACCOUNT-ID>:role/<ROLE-NAME>"

# The cluster name templating can be defined for all `aws-k8s` variants. To test on ipv4 and ipv6
# clusters, the following templates could be used. Note: TestSys does not currently support creating
# ipv6 clusters, so the ipv6 cluster must already exist.
cluster-names = ["{{arch}}-{{variant}}", "{{arch}}-{{variant}}-ipv6"]

# A custom conformance registry may be needed for testing if image pull reliability is a concern.
conformance-registry = "<ACCOUNT>.dkr.ecr.cn-north-1.amazonaws.com.cn"

# If testing using a kind cluster, AWS credentials need to be passed as a K8s secret.
secrets = {"awsCreds" = "myAwsCredentials"}

# Configuration for all nvidia AWS variants on x86_64 (family-flavor level configuration)
[aws-ecs-nvidia.x86_64]
instance-type = "p3.2xlarge"

# Configuration for all nvidia AWS variants on aarch64 (family-flavor level configuration)
[aws-ecs-nvidia.aarch64]
instance-type = "g5g.2xlarge"

# Configuration for only the `aws-k8s-1.21` variant (variant level configuration).
["aws-k8s-1.21".aarch64]
conformance-image = "<KUBERNETES-CONFORMANCE-IMAGE-URI>"

# Configurable values:
#
# cluster-names: 
#     All clusters the variant should be tested over. Cluster naming supports templated strings, and
#     both `arch` and `variant` are provided as variables (`{{arch}}-{{variant}}`).
#
# instance-type:
#     The instance type that should be used for testing.
#
# secrets:
#     A map containing all secrets needed for resource creation and testing.
#
# agent-role:
#     The role that should be assumed by each test and resource agent.
#
# conformance-image: (K8s only)
#     Specify a custom image for conformance testing. For `aws-k8s` variants this will be used as a
#     custom Kubernetes conformance image for Sonobuoy.
#
# conformance-registry: (K8s only)
#     Specify a custom registry for conformance testing images.
#     For `aws-k8s` variants this will be used as the Sonobuoy e2e registry.
```

webern (commented):
Ok, I've totally finished reviewing this. Great work!

TESTING.md
arnaldo2792: Nit on the nit: use "you" instead of "we".

TESTING.md
```diff
@@ -118,6 +118,24 @@ The Bottlerocket components are found in the `testsys-bottlerocket-aws` Kubernet
...
+[aws-k8s]
+# Set the default instance type for all `aws-k8s` variants
+instance-type = "m5.xlarge"
+
+[aws-k8s-nvidia]
+# Override the instance type for `nvidia` `aws-k8s` variants
+instance-type = "g5g.2xlarge"
+```
+
+Since `aws-k8s-nvidia` which is a `<FAMILY>-<FLAVOR>` level configuration it will take precedence over `aws-k8s` which is `<FAMILY>` level configuration.
```
arnaldo2792:
```suggestion
Since `aws-k8s-nvidia` is a `<FAMILY>-<FLAVOR>` level configuration it will take precedence over `aws-k8s` which is `<FAMILY>` level configuration.
```

.gitignore
arnaldo2792: For the lack of a better place to put this, `testsys.kubeconfig.eksctl.lock` was added in the first commit but deleted in the second commit. Please clean up the file so that it doesn't show up in either commit.

tools/testsys-config/src/lib.rs
arnaldo2792: I think there is a warning, as well as another warning when the default "development" keys are used to sign the repos.

tools/testsys/Test.toml.example
arnaldo2792: It is odd to me to document in the example configuration, rather than in the configuration file. I have not seen this convention in some of the open source projects I use, so probably we should consider stop documenting in example configuration files, and write down proper documentation where it should be. (Not for this PR)


Regarding the formatting in the toml file, would it make sense to use a linting tool and add that linting to the PR checks, and fix whatever we have to in other example configuration files? (Not in this PR)

tools/testsys/src/aws_resources.rs
arnaldo2792: Isn't the `conformance-lite` test type in sonobuoy just a NGINX pod? If so, we could have the same `conformance-lite` testing for ECS and support both orchestrators.

tools/testsys/src/aws_resources.rs
```diff
@@ -81,15 +108,23 @@ impl AwsK8s {
                 .await?,
             )
             .to_string();
-        let eks = self.eks_crd(testsys_images)?;
-        let ec2 = self.ec2_crd(testsys_images, Some(ami))?;
+        let eks = self.eks_crd(cluster_name, testsys_images)?;
+        let ec2 = self.ec2_crd(cluster_name, testsys_images, Some(ami))?;
+        let instance_provider = ec2.name().unwrap();
```
arnaldo2792: In a previous forced push you replaced `unwrap` with `expect`, but you missed two, I left a comment in the other.

tools/testsys/src/aws_resources.rs
```diff
@@ -426,29 +502,41 @@ impl AwsEcs {
                 .await?,
             )
             .to_string();
-        let ecs = self.ecs_crd(testsys_images)?;
-        let ec2 = self.ec2_crd(testsys_images, Some(ami))?;
+        let ecs = self.ecs_crd(cluster_name, testsys_images)?;
+        let ec2 = self.ec2_crd(cluster_name, testsys_images, Some(ami))?;
+        let instance_provider = ec2.name().unwrap();
```
arnaldo2792: Here is the other `unwrap` to be replaced with `expect`.

arnaldo2792 (changes requested):

tools/testsys/src/aws_resources.rs
etungsten: You're thinking of the quick test mode. `conformance-lite` is a subset of tests from the certified K8s conformance tests suite that is faster to run.

tools/testsys/Test.toml.example
webern: I doubt if a TOML linter would introspect the formatting of comments, which is where I had issues. I agree that maybe some of this could/maybe-should go in documentation instead of the example config. However it might actually be easier to find in the configuration. I know documentation discoverability is an area we want to improve in the project.

ecpullen:
^ 
* Remove `eksctl.lock` from history
* Updated `TESTING.md` with comments
* Added warning for use of default `Test.toml`
* Used the proposed `Test.toml.example`
* Replaced`unwrap()` with `expect()`

webern (approved):
:star: :rocket: :tada: :star:

arnaldo2792 (approved):

.gitignore
etungsten: Please remove `/testsys.kubeconfig.eksctl.lock` from `.gitignore`. It's not specific to Bottlerocket or any of its build/test frameworks. I see this akin to adding other developer workflow specific things like IDE configuration to the project's `.gitignore`.

For developers who don't want `eksctl.lock` files to show up in any of their git trees, they can use their personal global/local `.gitignore`. See https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreexcludesFile

etungsten (changes requested):

etungsten (approved):

PR #2403, opened 2022-09-07
buildsys: Use bundle name setting for output

### LINKED ISSUE #2402: buildsys: external files `bundle-output-path` not being honored
**Platform I'm building on:**

Any

Package build `external-files` has an option to provide `bundle-output-path` for naming the tarball that gets generated by bundling the source package with all of its vendored in dependencies when packaging a Go project. This setting is being ignored, resulting in build failures when the specified name is different than the defaulted name.

**What actually happened:**

Working on packaging a binary from `cloud-provider-aws`, the release tarball is named `v1.24.1.tar.gz`. Since this isn't very specific, I tried to rename the bundled output to include the project name for extra clarity. The build would fail though because it was still looking for the default name of `bundled-v1.24.1.tar.gz`.

Attempt to rename a bundled package by giving it a name other than its default.

stmcginnis:
**Issue number:**

Closes #2402 

**Description of changes:**

The `bundled-output-path` setting for external files had a small error when processing the output bundle for Go project dependencies where it would always use the default value, even when a different value was provided. This fixes the name of the variable used so it uses the provided name instead. This already handles setting this to the default value if the package maintainer has not provided an overridden value.

**Testing done:**

Made local change in #2377 and made sure it would try to use the specified name instead of the default.

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/gomod.rs
+++ b/tools/buildsys/src/gomod.rs
@@ -126,7 +126,7 @@ impl GoMod {
         let script_contents = GO_MOD_SCRIPT_TMPL
             .replace("__LOCAL_FILE_NAME__", &local_file_name.to_string_lossy())
             .replace("__MOD_DIR__", &mod_dir.to_string_lossy())
-            .replace("__OUTPUT__", &default_output_path.to_string_lossy());
+            .replace("__OUTPUT__", &output_path_arg.to_string_lossy());
         let script_path = format!(
             "{}/{}",
             package_dir.to_string_lossy(),


jpmcb (approved):
Excellent! Tested this out on the `hotdog` package and gave the following `Cargo.toml` change:

```
bundle-output-path: "woof-woof.tar.gz"
```

and I see it now correctly renames it in the package's working directory:

```
❯ ls -la
total 1204
drwxr-xr-x.  3 fedora fedora    4096 Sep  7 21:56 .
drwxr-xr-x. 95 fedora fedora    4096 Sep  6 22:27 ..
-rw-r--r--.  1 fedora fedora     209 Aug  5 21:51 build.rs
-rw-r--r--.  1 fedora fedora     624 Sep  7 21:53 Cargo.toml
-rw-r--r--.  1 fedora fedora   17270 Sep  7 21:56 hotdog-b85b755.tar.gz
-rw-r--r--.  1 fedora fedora    2042 Sep  6 16:24 hotdog.spec
-rw-r--r--.  1 fedora fedora      12 Aug  5 21:51 pkg.rs
drwxr-xr-x.  3 fedora fedora    4096 Aug 23 19:25 target
-rw-r--r--.  1 fedora fedora 1180409 Sep  7 21:56 woof-woof.tar.gz
```

🐕

bcressey (approved):

PR #2414, opened 2022-09-12
infrasys: update to AWS SDK Rust

jpculp:
**Issue number:**

#1968

**Description of changes:**

Replaces **rusoto** with **aws-sdk-for-rust** in **infrasys**.

**Testing done:**

_Coming soon!_

**Terms of contribution:**

### DIFF
--- a/tools/Cargo.lock
+++ b/tools/Cargo.lock
@@ -149,7 +149,7 @@ dependencies = [
  "http",
  "hyper",
  "ring",
- "time 0.3.14",
+ "time 0.3.15",
  "tokio",
  "tower",
  "tracing",
@@ -186,6 +186,30 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "tracing",
+]
+
+[[package]]
+name = "aws-sdk-cloudformation"
+version = "0.18.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cafaf0b9199f52cd69ef07c1d15fc7a57bf3ff53a8b0885cf708110fa49f6450"
+dependencies = [
+ "aws-endpoint",
+ "aws-http",
+ "aws-sig-auth",
+ "aws-smithy-async",
+ "aws-smithy-client",
+ "aws-smithy-http",
+ "aws-smithy-http-tower",
+ "aws-smithy-query",
+ "aws-smithy-types",
+ "aws-smithy-xml",
+ "aws-types",
+ "bytes",
+ "fastrand",
+ "http",
+ "tokio-stream",
+ "tower",
 ]
 
 [[package]]
@@ -255,6 +279,34 @@ dependencies = [
  "http",
  "tokio-stream",
  "tower",
+]
+
+[[package]]
+name = "aws-sdk-s3"
+version = "0.18.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "323b9107094fc396a0116326b577af48d9cfb26ec7c09588584ec82cee057b81"
+dependencies = [
+ "aws-endpoint",
+ "aws-http",
+ "aws-sig-auth",
+ "aws-sigv4",
+ "aws-smithy-async",
+ "aws-smithy-checksums",
+ "aws-smithy-client",
+ "aws-smithy-eventstream",
+ "aws-smithy-http",
+ "aws-smithy-http-tower",
+ "aws-smithy-types",
+ "aws-smithy-xml",
+ "aws-types",
+ "bytes",
+ "bytes-utils",
+ "http",
+ "http-body",
+ "tokio-stream",
+ "tower",
+ "tracing",
 ]
 
 [[package]]
@@ -331,6 +383,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0ee4bf20136757fd9f606bb4adafe6d19fb02bc48033a8d4f205f21d56fa783a"
 dependencies = [
  "aws-sigv4",
+ "aws-smithy-eventstream",
  "aws-smithy-http",
  "aws-types",
  "http",
@@ -343,7 +396,9 @@ version = "0.48.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b99b21b3aceaf224cccd693b353e1f38af4ede8c5fc618b97dd458bb63238efc"
 dependencies = [
+ "aws-smithy-eventstream",
  "aws-smithy-http",
+ "bytes",
  "form_urlencoded",
  "hex",
  "http",
@@ -351,7 +406,7 @@ dependencies = [
  "percent-encoding",
  "regex",
  "ring",
- "time 0.3.14",
+ "time 0.3.15",
  "tracing",
 ]
 
@@ -365,6 +420,27 @@ dependencies = [
  "pin-project-lite",
  "tokio",
  "tokio-stream",
+]
+
+[[package]]
+name = "aws-smithy-checksums"
+version = "0.48.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d6face4c12b335ba734a4416c15d5eeb0af88aa61182a84ff50db62bfa261183"
+dependencies = [
+ "aws-smithy-http",
+ "aws-smithy-types",
+ "bytes",
+ "crc32c",
+ "crc32fast",
+ "hex",
+ "http",
+ "http-body",
+ "md-5",
+ "pin-project-lite",
+ "sha1",
+ "sha2",
+ "tracing",
 ]
 
 [[package]]
@@ -391,11 +467,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-smithy-eventstream"
+version = "0.48.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b959c2c1752c2afbd863953046c06f7ee592f68d64719b7bab3193ac3b0fa77"
+dependencies = [
+ "aws-smithy-types",
+ "bytes",
+ "crc32fast",
+]
+
+[[package]]
 name = "aws-smithy-http"
 version = "0.48.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "23861d0b53a1369eab1e8d48c8bb3492eb3def1c2f2222dfb1bad58dd03914a5"
 dependencies = [
+ "aws-smithy-eventstream",
  "aws-smithy-types",
  "bytes",
  "bytes-utils",
@@ -454,7 +542,7 @@ dependencies = [
  "itoa",
  "num-integer",
  "ryu",
- "time 0.3.14",
+ "time 0.3.15",
 ]
 
 [[package]]
@@ -511,15 +599,6 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "block-buffer"
-version = "0.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4152116fd6e9dadb291ae18fc1ec3575ed6d84c29642d97890f4b4a3417297e4"
-dependencies = [
- "generic-array",
-]
-
-[[package]]
-name = "block-buffer"
 version = "0.10.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "69cce20737498f97b993470a6e536b8523f0af7892a4f928cceb1ac5e52ebe7e"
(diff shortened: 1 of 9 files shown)


jpculp:
Opening draft on @rpkelly's behalf. We'll want to merge around the same time that we transition **pubsys**.

tools/infrasys/src/main.rs
```diff
@@ -62,7 +63,17 @@ fn run() -> Result<()> {
     // Parse and store the args passed to the program
     let args = Args::from_args();
 
-    SimpleLogger::init(args.log_level, LogConfig::default()).context(error::LoggerSnafu)?;
+    match args.log_level {
+        // Set log level for AWS SDK to error to reduce verbosity.
+        LevelFilter::Info => Builder::new()
+            .filter_level(args.log_level)
+            .filter(Some("aws_config"), LevelFilter::Error)
+            .filter(Some("aws_smithy"), LevelFilter::Error)
```
bcressey: In `pubsys` we set this to `LevelFilter::Warn` - should we do that here for consistency?

tools/infrasys/src/root.rs
```diff
@@ -48,22 +60,26 @@ pub fn create_root(root_role_path: &Path) -> Result<()> {
         thing: "root role",
     })?;
     fs::create_dir_all(role_dir).context(error::MkdirSnafu { path: role_dir })?;
+
+    let default_region = thread::spawn(move || {
+        let runtime = tokio::runtime::Runtime::new().context(error::RuntimeCreationSnafu)?;
+        runtime.block_on(async_get_region())
+    })
+    .join()
+    .map_err(|_| error::Error::ThreadJoin {})??;
```
bcressey: We should do this the same way as `pubsys`, with the tokio runtime that drops out of scope before interacting with `tough`.

tools/infrasys/src/root.rs
```diff
@@ -108,11 +124,16 @@ fn add_keys_kms(
             num_keys: (*available_keys).len(),
         }
     );
-
+    let default_region = thread::spawn(move || {
+        let runtime = tokio::runtime::Runtime::new().context(error::RuntimeCreationSnafu)?;
+        runtime.block_on(async_get_region())
+    })
+    .join()
+    .map_err(|_| error::Error::ThreadJoin {})??;
```
bcressey: Can we abstract this into `get_region` that's called from `add_keys` and `create_root`? Or alternatively, handle it once before calling either function, since otherwise there's potential for this to blow up on refactoring if another tokio runtime has been started.

tools/infrasys/src/s3.rs
```diff
@@ -90,21 +89,24 @@ pub async fn create_s3_bucket(region: &str, stack_name: &str) -> Result<(String,
...
     region: &str,
     bucket_name: &str,
     prefix: &str,
     vpcid: &str,
 ) -> Result<()> {
     // Get old policy
-    let s3_client =
-        S3Client::new(Region::from_str(region).context(error::ParseRegionSnafu { what: region })?);
+    let config = aws_config::from_env()
+        .region(Region::new(region.to_owned()))
```
bcressey: Maybe turn this into a `Region` or even an `SdkConfig` when `ValidRepoInfo` is created, so we don't need to do the conversion each time.

tools/Cargo.lock
```diff
@@ -46,9 +46,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.58"
```
ecpullen: Do we want the updated packages in a separate commit?

tools/infrasys/src/s3.rs
```diff
@@ -90,21 +89,24 @@ pub async fn create_s3_bucket(region: &str, stack_name: &str) -> Result<(String,
...
+        .await;
+    let s3_client = S3Client::new(&config);
     let mut policy: serde_json::Value = match s3_client
-        .get_bucket_policy(GetBucketPolicyRequest {
-            bucket: bucket_name.to_string(),
-            expected_bucket_owner: None,
-        })
+        .get_bucket_policy()
+        .bucket(bucket_name.to_string())
+        .set_expected_bucket_owner(None)
```
ecpullen: Isn't this `None` by default?

tools/infrasys/src/s3.rs
```diff
@@ -176,8 +179,11 @@ pub async fn upload_file(
     prefix: &str,
     file_path: &Path,
 ) -> Result<()> {
-    let s3_client =
-        S3Client::new(Region::from_str(region).context(error::ParseRegionSnafu { what: region })?);
+    let config = aws_config::from_env()
```
ecpullen, line 187: Is it possible/desirable to pass the `aws_config` instead of recreating it every time?

tools/infrasys/src/shared.rs
```diff
@@ -66,10 +68,9 @@ pub async fn get_stack_outputs(
         );
         thread::sleep(time::Duration::from_secs(20));
         stack_outputs = cfn_client
-            .describe_stacks(DescribeStacksInput {
-                stack_name: Some(stack_name.to_string()),
-                ..Default::default()
-            })
+            .describe_stacks()
+            .stack_name(stack_name.to_string())
```
ecpullen: This and various other places shouldn't need `to_string()` as the input is `Into<String>`.

tools/infrasys/src/shared.rs
```diff
@@ -78,7 +79,10 @@ pub async fn get_stack_outputs(
                 resource_name: stack_name,
             })?[0]
```
ecpullen: Is `[0]` safe?

tools/infrasys/src/main.rs
rpkelly: Fixed

tools/infrasys/src/root.rs
rpkelly: fixed

tools/infrasys/src/root.rs
rpkelly: fixed

tools/infrasys/src/s3.rs
rpkelly: Added Region, since the string will still be needed in some places. Not sure about SDKconfig, but, I'll have that conversation in more depth in @ecpullen's comment

tools/infrasys/src/shared.rs
rpkelly: The description of the method makes some claims that it would be, but, there's no reason not to handle the case. Fixing.

tools/infrasys/src/s3.rs
rpkelly: I'm hesitant to add the config, mostly because the most obvious place to add it keeps track of lifetimes, while the aws sdk wants a static lifetime for everything. It makes things a bit awkward, as you can see from the Region changes.

tools/infrasys/Cargo.toml
```diff
@@ -12,9 +12,10 @@ clap = "3.1"
 hex = "0.4.0"
 log = "0.4.14"
 pubsys-config = { path = "../pubsys-config/", version = "0.1.0" }
-rusoto_cloudformation = { version = "0.48.0", default-features = false, features = ["rustls"] }
-rusoto_core = { version = "0.48.0", default-features = false, features = ["rustls"] }
-rusoto_s3 = { version = "0.48.0", default-features = false, features = ["rustls"] }
+aws-config = "0.46.0"
+aws-types = "0.46.0"
+aws-sdk-cloudformation = "0.16.0"
+aws-sdk-s3 = "0.16.0"
```
jpculp:
```suggestion
aws-config = "0.48.0"
aws-types = "0.48.0"
aws-sdk-cloudformation = "0.18.0"
aws-sdk-s3 = "0.18.0"
```

jpculp (commented):
With this change, we should be able to remove L66-67 in [`tools/deny.toml`](https://github.com/bottlerocket-os/bottlerocket/blob/ea9634f82d97b0b7ea8aa6174e946ad8663e0646/tools/deny.toml#L66).

tools/infrasys/src/error.rs
```diff
@@ -149,17 +156,27 @@ pub enum Error {
...
     #[snafu(display("'tuftool {}' returned {}", command, code))]
     TuftoolResult { command: String, code: String },
 
     #[snafu(display("Failed to start tuftool: {}", source))]
     TuftoolSpawn { source: io::Error },
+
+    #[snafu(display("Unable to create tokio runtime: {}", source))]
+    RuntimeCreation {
+        source: std::io::Error,
+        backtrace: Backtrace,
```
jpculp: In [pubsys](https://github.com/bottlerocket-os/bottlerocket/pull/2415) we ended up removing backtrace from RuntimeCreation.

jpculp: Or maybe drop this in favor of `Runtime` from L162-L163.

tools/infrasys/src/error.rs
```diff
@@ -14,7 +14,9 @@ pub enum Error {
     CreateStack {
         stack_name: String,
         region: String,
-        source: rusoto_core::RusotoError<rusoto_cloudformation::CreateStackError>,
+        source: aws_sdk_cloudformation::types::SdkError<
```
zmrow: Nit: I might import the various `*::types` crates  so we don't have to have the fully qualified imports in every single error variant

tools/infrasys/src/keys.rs
```diff
@@ -89,9 +87,13 @@ impl KMSKeyConfigExt for KMSKeyConfig {
...
             );
-            let cfn_client = CloudFormationClient::new(
-                Region::from_str(region).context(error::ParseRegionSnafu { what: region })?,
-            );
+
+            let config = aws_config::from_env()
+                .region(Region::new(region.to_owned()))
+                .load()
+                .await;
+            let cfn_client = aws_sdk_cloudformation::Client::new(&config);
```
zmrow: Nit: I'd also import `aws_sdk_cloudformation::Client as CloudFormationClient` or something so we don't need to have the full path here.

tools/infrasys/src/shared.rs
```diff
@@ -32,23 +32,27 @@ pub async fn get_stack_outputs(
     cfn_client: &CloudFormationClient,
     stack_name: &str,
     region: &str,
-) -> Result<Vec<rusoto_cloudformation::Output>> {
+) -> Result<Vec<aws_sdk_cloudformation::model::Output>> {
```
zmrow: Nit: Same here re: imports

jpculp: What we probably want is something like

```use aws_sdk_cloudformation::output::DescribeStacksOutput;```

and then have

```Result<Vec<DescribeStacksOutput>>```

Comment (rpkelly) [reply tools/infrasys/src/shared.rs]: imported outputs at the top. Outputs is actually the type we want to return here.

Comment (jpculp) [reply tools/infrasys/src/shared.rs]: Ah, I see it changes from `DescribeStacksOutput` when you get to the `.stacks`.

`Output` seems very non-descriptive though and could conflict with something else down the road. Would it be better to do `use aws_sdk_cloudformation::model::{Output as CfnOutput, Parameter}`?

File: tools/infrasys/src/error.rs [hunk 15]
```diff
@@ -149,17 +153,24 @@ pub enum Error {
...
+    ThreadJoin,
+
     #[snafu(display("'tuftool {}' returned {}", command, code))]
     TuftoolResult { command: String, code: String },
 
     #[snafu(display("Failed to start tuftool: {}", source))]
     TuftoolSpawn { source: io::Error },
+
+    #[snafu(display("Unable to create tokio runtime: {}", source))]
+    RuntimeCreation { source: std::io::Error },
```
bcressey: Do we still need the new `ThreadJoin` and `RuntimeCreation` errors?

bcressey (approved):

tools/infrasys/src/keys.rs
```diff
@@ -34,7 +33,7 @@ pub fn check_signing_key_config(signing_key_config: &SigningKeyConfig) -> Result
         SigningKeyConfig::file { .. } => (),
         SigningKeyConfig::kms { config, .. } => {
             let config = config.as_ref().context(error::MissingConfigSnafu {
-                missing: "config field for a kms key",
+                missing: "config field for a kms keys",
```
jpmcb: This sounds weird: should it be "config field for kms keys" since it's now plural?

jpmcb (approved):

tools/infrasys/src/keys.rs
rpkelly: Fixed

tools/infrasys/src/error.rs
rpkelly: Fixed.

bcressey (approved):

PR #2303, opened 2022-07-27
Update rust dependencies for v1.9.0 release

etungsten:
**Issue number:**
Resolves https://github.com/bottlerocket-os/bottlerocket/issues/2208

**Description of changes:**
```
    sources: update rust dependencies
    
    Ran 'cargo update' in sources workspace

```
```
    tools: update rust dependencies
    
    Ran 'cargo update' to update rust dependencies.
```

**Testing done:**
Built `aws-k8s-1.23` image and AMI
Host comes up fine, joins cluster fine. Runs pods fine.
Tried changing settings and checked `apiserver` logs, everything works as expected.

`cargo make check-repo-expirations` works fine
```
[ec2-user@ip-172-31-32-162 bottlerocket]$ cargo make -e PUBLISH_REPO=etung -e BUILDSYS_VARIANT=metal-k8s-1.23 -e BUILDSYS_ARCH=x86_64 
check-repo-expirations                                                                        
[cargo-make] INFO - cargo make 0.35.1                                                         
[cargo-make] INFO - Build File: Makefile.toml                                                 
[cargo-make] INFO - Task: check-repo-expirations                                              
[cargo-make] INFO - Profile: development                                                      
[cargo-make] INFO - Running Task: setup                                                       
[cargo-make] INFO - Running Task: fetch-sources                                               
[cargo-make] INFO - Running Task: tuftool                                                     
[cargo-make] INFO - Running Task: publish-setup-tools                                
[cargo-make] INFO - Running Task: publish-setup                                   
01:52:21 [INFO] Found infra config at path: /home/ec2-user/bottlerocket/Infra.toml            
[cargo-make] INFO - Running Task: setup                                                       
[cargo-make] INFO - Running Task: fetch-sources                                               
[cargo-make] INFO - Running Task: publish-tools                                                           
[cargo-make] INFO - Running Task: check-repo-expirations                          
01:52:23 [INFO] Found infra config at path: /home/ec2-user/bottlerocket/Infra.toml                        
01:52:23 [INFO] Loaded TUF repo:        https://blah.cloudfront.net/etung/metal-k8s-1.23/x86_64                
01:52:23 [INFO] Root expiration:        2022-09-07 22:08:13 UTC                               
01:52:23 [INFO] Snapshot expiration:    2022-08-17 01:38:54.595712907 UTC                     
01:52:23 [INFO] Targets expiration:     2022-08-17 01:38:54.595712907 UTC                     
[...]

PR #2415, opened 2022-09-12
pubsys: update to AWS SDK Rust

jpculp:
**Issue number:**

#1968

**Description of changes:**

Replaces **rusoto** with **aws-sdk-for-rust** in **pubsys**.

**Testing done:**
- [x] publish ami to us-west-2
- [x] publish multiple amis with chained assumed roles (2 different accounts, 3 regions)
- [x] grant ami access
- [x] revoke ami access
- [x] publish ssm parameters
- [x] promote ssm parameters

**TUF testing done:**
_Thanks, @etungsten!_
- [x] use KMS key when creating TUF repo
- [x] check TUF repo expirations
- [x] validate TUF repo
- [x] refresh TUF repo

**Terms of contribution:**

### DIFF
--- a/tools/deny.toml
+++ b/tools/deny.toml
@@ -58,14 +58,17 @@ multiple-versions = "deny"
 wildcards = "deny"
 
 skip-tree = [
-    # temporarily using a different version of snafu
-    { name = "parse-datetime", version = "0.1.0" },
-
     # rusoto_signature uses an older version of sha2
     { name = "rusoto_signature" },
 
-    # reqwest uses an older rustls-pemfile
-    { name = "reqwest", version = "0.11.10" },
+    # argh_derive pulls in an older version of heck
+    { name = "argh_derive", version = "0.1.8" },
+
+    # structopt pulls in an older version of clap
+    { name = "structopt", version = "0.3.26" },
+
+    # aws-smithy-client uses an older hyper-rustls
+    { name = "aws-smithy-client", version = "0.46.0" },
 ]
 
 [sources]
--- a/tools/pubsys/src/aws/ami/mod.rs
+++ b/tools/pubsys/src/aws/ami/mod.rs
@@ -6,19 +6,22 @@ mod snapshot;
 pub(crate) mod wait;
 
 use crate::aws::publish_ami::{get_snapshots, modify_image, modify_snapshots};
-use crate::aws::{client::build_client, parse_arch, region_from_string};
+use crate::aws::{client::build_client_config, parse_arch, region_from_string};
 use crate::Args;
+use aws_sdk_ebs::Client as EbsClient;
+use aws_sdk_ec2::error::CopyImageError;
+use aws_sdk_ec2::model::{ArchitectureValues, OperationType};
+use aws_sdk_ec2::output::CopyImageOutput;
+use aws_sdk_ec2::types::SdkError;
+use aws_sdk_ec2::{Client as Ec2Client, Region};
+use aws_sdk_sts::error::GetCallerIdentityError;
+use aws_sdk_sts::output::GetCallerIdentityOutput;
+use aws_sdk_sts::Client as StsClient;
 use futures::future::{join, lazy, ready, FutureExt};
 use futures::stream::{self, StreamExt};
 use log::{error, info, trace, warn};
-use pubsys_config::{AwsConfig, InfraConfig};
+use pubsys_config::{AwsConfig as PubsysAwsConfig, InfraConfig};
 use register::{get_ami_id, register_image, RegisteredIds};
-use rusoto_core::{Region, RusotoError};
-use rusoto_ebs::EbsClient;
-use rusoto_ec2::{CopyImageError, CopyImageRequest, CopyImageResult, Ec2, Ec2Client};
-use rusoto_sts::{
-    GetCallerIdentityError, GetCallerIdentityRequest, GetCallerIdentityResponse, Sts, StsClient,
-};
 use serde::{Deserialize, Serialize};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::{HashMap, HashSet};
@@ -41,15 +44,15 @@ pub(crate) struct AmiArgs {
 
     /// Desired root volume size in gibibytes
     #[structopt(long)]
-    root_volume_size: Option<i64>,
+    root_volume_size: Option<i32>,
 
     /// Desired data volume size in gibibytes
     #[structopt(long)]
-    data_volume_size: Option<i64>,
+    data_volume_size: Option<i32>,
 
     /// The architecture of the machine image
     #[structopt(short = "a", long, parse(try_from_str = parse_arch))]
-    arch: String,
+    arch: ArchitectureValues,
 
     /// The desired AMI name
     #[structopt(short = "n", long)]
@@ -106,8 +109,8 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
         aws.regions.clone().into()
     }
     .into_iter()
-    .map(|name| region_from_string(&name, &aws).context(error::ParseRegionSnafu))
-    .collect::<Result<Vec<Region>>>()?;
+    .map(|name| region_from_string(&name))
+    .collect::<Vec<Region>>();
 
     ensure!(
         !regions.is_empty(),
@@ -120,46 +123,37 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
     let base_region = regions.remove(0);
 
     // Build EBS client for snapshot management, and EC2 client for registration
-    let base_ebs_client = build_client::<EbsClient>(&base_region, &base_region, &aws).context(
-        error::ClientSnafu {
-            client_type: "EBS",
-            region: base_region.name(),
-        },
-    )?;
-    let base_ec2_client = build_client::<Ec2Client>(&base_region, &base_region, &aws).context(
-        error::ClientSnafu {
-            client_type: "EC2",
-            region: base_region.name(),
-        },
-    )?;
+    let client_config = build_client_config(&base_region, &base_region, &aws).await;
+
+    let base_ebs_client = EbsClient::new(&client_config);
+
+    let base_ec2_client = Ec2Client::new(&client_config);
 
     // Check if the AMI already exists, in which case we can use the existing ID, otherwise we
     // register a new one.
     let maybe_id = get_ami_id(
         &ami_args.name,
         &ami_args.arch,
-        base_region.name(),
+        &base_region,
         &base_ec2_client,
     )
     .await
     .context(error::GetAmiIdSnafu {
         name: &ami_args.name,
-        arch: &ami_args.arch,
-        region: base_region.name(),
+        arch: ami_args.arch.as_ref(),
+        region: base_region.as_ref(),
     })?;
 
     let (ids_of_image, already_registered) = if let Some(found_id) = maybe_id {
         warn!(
             "Found '{}' already registered in {}: {}",
-            ami_args.name,
-            base_region.name(),
-            found_id
+            ami_args.name, base_region, found_id
         );
         let snapshot_ids = get_snapshots(&found_id, &base_region, &base_ec2_client)
             .await
             .context(error::GetSnapshotsSnafu {
                 image_id: &found_id,
-                region: base_region.name(),
+                region: base_region.as_ref(),
             })?;
         let found_ids = RegisteredIds {
             image_id: found_id,
@@ -167,29 +161,22 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
         };
         (found_ids, true)
     } else {
-        let new_ids = register_image(
-            ami_args,
-            base_region.name(),
-            base_ebs_client,
-            &base_ec2_client,
-        )
-        .await
-        .context(error::RegisterImageSnafu {
-            name: &ami_args.name,
-            arch: &ami_args.arch,
-            region: base_region.name(),
-        })?;
+        let new_ids = register_image(ami_args, &base_region, base_ebs_client, &base_ec2_client)
+            .await
+            .context(error::RegisterImageSnafu {
+                name: &ami_args.name,
+                arch: ami_args.arch.as_ref(),
+                region: base_region.as_ref(),
+            })?;
         info!(
             "Registered AMI '{}' in {}: {}",
-            ami_args.name,
-            base_region.name(),
-            new_ids.image_id
+            ami_args.name, base_region, new_ids.image_id
         );
         (new_ids, false)
     };
 
     amis.insert(
-        base_region.name().to_string(),
+        base_region.as_ref().to_string(),
         Image::new(&ids_of_image.image_id, &ami_args.name),
     );
 
@@ -211,7 +198,7 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
     .await
     .context(error::WaitAmiSnafu {
         id: &ids_of_image.image_id,
-        region: base_region.name(),
+        region: base_region.as_ref(),
     })?;
 
     // For every other region, initiate copy-image calls.
@@ -223,18 +210,14 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
 
     // Get the account ID used in the base region; we don't need to grant to it so we can remove it
     // from the list.
-    let base_sts_client = build_client::<StsClient>(&base_region, &base_region, &aws).context(
-        error::ClientSnafu {
-            client_type: "STS",
-            region: base_region.name(),
+    let client_config = build_client_config(&base_region, &base_region, &aws).await;
+    let base_sts_client = StsClient::new(&client_config);
+
+    let response = base_sts_client.get_caller_identity().send().await.context(
+        error::GetCallerIdentitySnafu {
+            region: base_region.as_ref(),
         },
     )?;
-    let response = base_sts_client
-        .get_caller_identity(GetCallerIdentityRequest {})
-        .await
-        .context(error::GetCallerIdentitySnafu {
-            region: base_region.name(),
-        })?;
     let base_account_id = response.account.context(error::MissingInResponseSnafu {
         request_type: "GetCallerIdentity",
         missing: "account",
@@ -249,7 +232,7 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
         modify_snapshots(
             Some(account_id_vec.clone()),
             None,
-            "add",
+            &OperationType::Add,
             &ids_of_image.snapshot_ids,
             &base_ec2_client,
             &base_region,
@@ -257,21 +240,20 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
         .await
         .context(error::GrantAccessSnafu {
             thing: "snapshots",
-            region: base_region.name(),
+            region: base_region.as_ref(),
         })?;
 
         modify_image(
             Some(account_id_vec.clone()),
             None,
-            "add",
+            &OperationType::Add,
             &ids_of_image.image_id,
             &base_ec2_client,
-            &base_region,
         )
         .await
-        .context(error::GrantAccessSnafu {
+        .context(error::GrantImageAccessSnafu {
             thing: "image",
-            region: base_region.name(),
+            region: base_region.as_ref(),
         })?;
     }
 
@@ -279,11 +261,8 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
     // clients because they're used in a future and need to live until the future is resolved.
     let mut ec2_clients = HashMap::with_capacity(regions.len());
     for region in regions.iter() {
-        let ec2_client =
-            build_client::<Ec2Client>(region, &base_region, &aws).context(error::ClientSnafu {
-                client_type: "EC2",
-                region: region.name(),
-            })?;
+        let client_config = build_client_config(region, &base_region, &aws).await;
+        let ec2_client = Ec2Client::new(&client_config);
         ec2_clients.insert(region.clone(), ec2_client);
     }
 
@@ -292,7 +271,7 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
     let mut get_requests = Vec::with_capacity(regions.len());
     for region in regions.iter() {
         let ec2_client = &ec2_clients[region];
-        let get_request = get_ami_id(&ami_args.name, &ami_args.arch, region.name(), ec2_client);
+        let get_request = get_ami_id(&ami_args.name, &ami_args.arch, region, ec2_client);
         let info_future = ready(region.clone());
         get_requests.push(join(info_future, get_request));
     }
@@ -305,41 +284,33 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
     for (region, get_response) in get_responses {
         let get_response = get_response.context(error::GetAmiIdSnafu {
             name: &ami_args.name,
-            arch: &ami_args.arch,
-            region: region.name(),
+            arch: ami_args.arch.as_ref(),
+            region: region.as_ref(),
         })?;
         if let Some(id) = get_response {
             info!(
                 "Found '{}' already registered in {}: {}",
-                ami_args.name,
-                region.name(),
-                id
+                ami_args.name, region, id
             );
-            amis.insert(region.name().to_string(), Image::new(&id, &ami_args.name));
+            amis.insert(region.as_ref().to_string(), Image::new(&id, &ami_args.name));
             continue;
         }
 
         let ec2_client = &ec2_clients[&region];
-        let copy_request = CopyImageRequest {
-            description: ami_args.description.clone(),
-            name: ami_args.name.clone(),
-            source_image_id: ids_of_image.image_id.clone(),
-            source_region: base_region.name().to_string(),
-            ..Default::default()
-        };
-        let copy_future = ec2_client.copy_image(copy_request);
+        let base_region = base_region.to_owned();
+        let copy_future = ec2_client
+            .copy_image()
+            .set_description(ami_args.description.clone())
+            .set_name(Some(ami_args.name.clone()))
+            .set_source_image_id(Some(ids_of_image.image_id.clone()))
+            .set_source_region(Some(base_region.as_ref().to_string()))
+            .send();
 
-        let base_region_name = base_region.name();
         // Store the region so we can output it to the user
         let region_future = ready(region.clone());
         // Let the user know the copy is starting, when this future goes to run
-        let message_future = lazy(move |_| {
-            info!(
-                "Starting copy from {} to {}",
-                base_region_name,
-                region.name()
-            )
-        });
+        let message_future =
+            lazy(move |_| info!("Starting copy from {} to {}", base_region, region));
         copy_requests.push(message_future.then(|_| join(region_future, copy_future)));
     }
 
@@ -357,7 +328,7 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
     // Run through the stream and collect results into a list.
     let copy_responses: Vec<(
         Region,
-        std::result::Result<CopyImageResult, RusotoError<CopyImageError>>,
+        std::result::Result<CopyImageOutput, SdkError<CopyImageError>>,
     )> = request_stream.collect().await;
 
     // Report on successes and errors; don't fail immediately if we see an error so we can report
@@ -369,26 +340,23 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
                 if let Some(image_id) = success.image_id {
                     info!(
                         "Registered AMI '{}' in {}: {}",
-                        ami_args.name,
-                        region.name(),
-                        image_id,
+                        ami_args.name, region, image_id,
                     );
                     amis.insert(
-                        region.name().to_string(),
+                        region.as_ref().to_string(),
                         Image::new(&image_id, &ami_args.name),
                     );
                 } else {
                     saw_error = true;
                     error!(
                         "Registered AMI '{}' in {} but didn't receive an AMI ID!",
-                        ami_args.name,
-                        region.name(),
+                        ami_args.name, region,
                     );
                 }
             }
             Err(e) => {
                 saw_error = true;
-                error!("Copy to {} failed: {}", region.name(), e);
+                error!("Copy to {} failed: {}", region, e);
             }
         }
     }
@@ -420,7 +388,7 @@ impl Image {
 async fn get_account_ids(
     regions: &[Region],
     base_region: &Region,
-    aws: &AwsConfig,
+    pubsys_aws_config: &PubsysAwsConfig,
 ) -> Result<HashSet<String>> {
     let mut grant_accounts = HashSet::new();
 
@@ -428,18 +396,15 @@ async fn get_account_ids(
     // live until the future is resolved.
     let mut sts_clients = HashMap::with_capacity(regions.len());
     for region in regions.iter() {
-        let sts_client =
-            build_client::<StsClient>(region, base_region, aws).context(error::ClientSnafu {
-                client_type: "STS",
-                region: region.name(),
-            })?;
+        let client_config = build_client_config(region, base_region, pubsys_aws_config).await;
+        let sts_client = StsClient::new(&client_config);
         sts_clients.insert(region.clone(), sts_client);
     }
 
     let mut requests = Vec::with_capacity(regions.len());
     for region in regions.iter() {
         let sts_client = &sts_clients[region];
-        let response_future = sts_client.get_caller_identity(GetCallerIdentityRequest {});
+        let response_future = sts_client.get_caller_identity().send();
 
         // Store the region so we can include it in any errors
         let region_future = ready(region.clone());
@@ -450,12 +415,12 @@ async fn get_account_ids(
     // Run through the stream and collect results into a list.
     let responses: Vec<(
         Region,
-        std::result::Result<GetCallerIdentityResponse, RusotoError<GetCallerIdentityError>>,
+        std::result::Result<GetCallerIdentityOutput, SdkError<GetCallerIdentityError>>,
     )> = request_stream.collect().await;
 
     for (region, response) in responses {
         let response = response.context(error::GetCallerIdentitySnafu {
-            region: region.name(),
+            region: region.as_ref(),
         })?;
         let account_id = response.account.context(error::MissingInResponseSnafu {
             request_type: "GetCallerIdentity",
@@ -469,9 +434,10 @@ async fn get_account_ids(
 }
 
 mod error {
-    use crate::aws::{self, ami, publish_ami};
-    use rusoto_core::RusotoError;
-    use rusoto_sts::GetCallerIdentityError;
+    use crate::aws::{ami, publish_ami};
+    use aws_sdk_ec2::error::ModifyImageAttributeError;
+    use aws_sdk_ec2::types::SdkError;
+    use aws_sdk_sts::error::GetCallerIdentityError;
     use snafu::Snafu;
     use std::path::PathBuf;
 
@@ -481,17 +447,8 @@ mod error {
         #[snafu(display("Some AMIs failed to copy, see above"))]
         AmiCopy,
 
-        #[snafu(display("Error creating {} client in {}: {}", client_type, region, source))]
-        Client {
-            client_type: String,
-            region: String,
-            source: aws::client::Error,
-        },
-
         #[snafu(display("Error reading config: {}", source))]
-        Config {
-            source: pubsys_config::Error,
-        },
+        Config { source: pubsys_config::Error },
 
         #[snafu(display("Failed to create file '{}': {}", path.display(), source))]
         FileCreate {
@@ -510,7 +467,7 @@ mod error {
         #[snafu(display("Error getting account ID in {}: {}", region, source))]
         GetCallerIdentity {
             region: String,
-            source: RusotoError<GetCallerIdentityError>,
+            source: SdkError<GetCallerIdentityError>,
         },
 
         #[snafu(display(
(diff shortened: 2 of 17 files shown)


tools/pubsys/src/aws/client.rs
```diff
@@ -1,161 +1,164 @@
...
-/// Box<ProvideAwsCredentials>.
-struct CredentialsProvider(Box<dyn ProvideAwsCredentials + Send + Sync + 'static>);
-#[async_trait]
-impl ProvideAwsCredentials for CredentialsProvider {
-    async fn credentials(&self) -> std::result::Result<AwsCredentials, CredentialsError> {
-        self.0.credentials().await
+    if maybe_regional_role.is_some() {
+        build_chained_config(&region, sts_region, assume_roles.clone(), &aws.profile)
+            .await
+            .expect("?????")
```
jpculp: This needs a proper expect message or an alternative method of unwrapping the result.

tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -119,77 +118,58 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
     // We register in this base region first, then copy from there to any other regions.
     let base_region = regions.remove(0);
 
+    let arch = ami_args.arch.as_str().to_owned();
```
etungsten: nit: I would drop the `.to_owned()` here and do that for the places that need it.
```suggestion
    let arch = ami_args.arch.as_str();
```

Then when you're passing the `arch` to the error enum variants, you can just for example do:
```
...
    .context(error::GetAmiIdSnafu {
        name: &ami_args.name,
        arch,
        region: &base_region,
    })?;
```

Same comment applies elsewhere where you do similar things with `let arch`.

tools/pubsys/src/aws/client.rs
jpculp: Fixed this in my latest push.

jpculp:
Fixed the result unwrapping issue while also improving how and where the profile gets loaded.

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,151 @@
...
-    where
-        P: ProvideAwsCredentials + Send + Sync + 'static,
-        D: DispatchSignedRequest + Send + Sync + 'static,
-    {
-        Self::new_with(request_dispatcher, credentials_provider, region)
-    }
-}
+    // Add custom endpoint if specified in aws.region.REGION.endpoint.
+    let maybe_regional_endpoint = aws.region.get(&region).and_then(|r| r.endpoint.clone());
+    config = if let Some(endpoint) = &maybe_regional_endpoint {
```
zmrow: You probably don't need the additional `config =` assignment here or the `else` (unless I'm missing something).

Since `config` is mutable, you could do:
```
if let Some(...) = ... {
    config.endpoint_resolver...
}
```

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,151 @@
...
-    )?;
-    Ok(T::new_with(
-        rusoto_core::HttpClient::new().context(error::HttpClientSnafu)?,
-        provider,
-        region.clone(),
-    ))
-}
+    // If the user specified a profile, use that, otherwise use the default
+    // credentials mechanisms.
+    config = if let Some(profile) = &aws.profile {
```
zmrow: Same here.

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,151 @@
...
-/// Box<ProvideAwsCredentials>.
-struct CredentialsProvider(Box<dyn ProvideAwsCredentials + Send + Sync + 'static>);
-#[async_trait]
-impl ProvideAwsCredentials for CredentialsProvider {
-    async fn credentials(&self) -> std::result::Result<AwsCredentials, CredentialsError> {
-        self.0.credentials().await
-    }
+    // Load a chained credential config if role is specified in aws.region.REGION.role.
+    let maybe_regional_role = aws.region.get(&region).and_then(|r| r.role.clone());
+    let config = if maybe_regional_role.is_some() {
```
zmrow: I'm confused about this `let config`... are we replacing the `config` we just built?

stmcginnis (commented):
This looks good to me - with the caveat that I am not very familiar with this code. But updated code looks logical, and maybe most importantly, all tests are passing.

zmrow (commented):
As an aside, if you're not already doing so it'd probably be good to run down the entire set of pubsys commands in your own account, making sure that the grant/revoke and SSM params all work as expected.  Those are super important.

tools/pubsys/src/aws/client.rs
zmrow: If `config.endpoint_resolver` isn't a "setter" I suppose my suggestion won't work.

tools/pubsys/src/aws/client.rs
jpculp: Correct, but I can rename it to make it more clear.

tools/pubsys/src/aws/client.rs
zmrow: Ah - according to the docs it consumes `self` and overrides the endpoint resolver. https://docs.rs/aws-config/latest/aws_config/struct.ConfigLoader.html#method.endpoint_resolver

jpculp:
- Fixed missing region from chained configuration.
- Removed the extraneous `else`'s.

tools/pubsys/src/repo.rs
```diff
@@ -753,6 +763,13 @@ mod error {
             source: tough::error::Error,
         },
 
+        /// The library failed to instantiate 'tokio Runtime'.
+        #[snafu(display("Unable to create tokio runtime: {}", source))]
+        RuntimeCreation {
+            source: std::io::Error,
+            backtrace: Backtrace,
```
etungsten: Why do we need a backtrace here? I also think you need to enable the `backtraces` feature flag for the snafu crate for this to have any effect? https://docs.rs/snafu/0.6.3/snafu/guide/feature_flags/index.html#backtraces

tools/pubsys/src/main.rs
```diff
@@ -39,8 +40,17 @@ fn run() -> Result<()> {
...
     let args = Args::from_args();
 
-    // SimpleLogger will send errors to stderr and anything less to stdout.
-    SimpleLogger::init(args.log_level, LogConfig::default()).context(error::LoggerSnafu)?;
+    match args.log_level {
+        // Set log level for AWS SDK to error to reduce verbosity.
+        LevelFilter::Info => Builder::new()
+            .filter_level(args.log_level)
+            .filter(Some("aws_config"), LevelFilter::Error)
+            .filter(Some("aws_smithy"), LevelFilter::Error)
```
etungsten: What does filtering at `Warning` look like for the `aws-sdk-rust` crates? Is it just as noisy?

tools/pubsys/src/main.rs
```diff
@@ -39,8 +40,17 @@ fn run() -> Result<()> {
...
+            .init(),
+
+        // Set the supplied log level across the whole crate.
+        _ => Builder::new().filter_level(args.log_level).init(),
```
etungsten, line 53: If I don't explicitly set the logging level for `pubsys`, I think I'd still want to skip out on all the extraneous `info` level log from the `aws-rust-sdk` crates.

tools/pubsys/src/aws/ssm/ssm.rs
```diff
@@ -97,7 +94,7 @@ where
         ensure!(
             total_count == expected_len,
             error::MissingInResponseSnafu {
-                region: region.name(),
+                region: &region,
```
etungsten: nit: can we do this?:

```suggestion
                region,
```

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -503,31 +494,30 @@ pub(crate) async fn modify_regional_images(
...
     use std::path::PathBuf;
 
     #[derive(Debug, Snafu)]
     #[snafu(visibility(pub(super)))]
     pub(crate) enum Error {
-        #[snafu(display("Error creating {} client in {}: {}", client_type, region, source))]
-        Client {
-            client_type: String,
+        #[snafu(display("Error creating client config in {}: {}", region, source))]
+        ClientConfig {
```
etungsten: We removed the `client_type` field from this error. When something does fail with the client config generation, is it clear which type of client (e.g. EC2 client) we're generating the config for? Or does it not matter?

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -150,21 +147,21 @@ pub(crate) async fn run(args: &Args, publish_args: &PublishArgs) -> Result<()> {
...
-    let wait_responses: Vec<((Region, String), std::result::Result<(), wait::Error>)> =
+    let wait_responses: Vec<((String, String), std::result::Result<(), wait::Error>)> =
         request_stream.collect().await;
 
     // Make sure waits succeeded and AMIs are available.
     for ((region, image_id), wait_response) in wait_responses {
         wait_response.context(error::WaitAmiSnafu {
             id: &image_id,
-            region: region.name(),
+            region: &region,
```
etungsten: nit: can we do this?
```suggestion
            region,
```

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -120,28 +119,26 @@ pub(crate) async fn run(args: &Args, publish_args: &PublishArgs) -> Result<()> {
         }
     );
 
-    // Parse region names, adding endpoints from InfraConfig if specified
+    // Parse regions, adding endpoints from InfraConfig if specified
```
etungsten: Are we still parsing the region string? 
```suggestion
    // Adding endpoints from InfraConfig if specified
```

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,150 @@
...
+                .secret_access_key()
+                .context(error::CredentialsMissingSnafu {
+                    credentials_key: "secret_access_key",
+                    role_arn: &assume_role,
+                })?,
+            sts_credentials.session_token().map(|s| s.to_string()),
+            Some(
+                SystemTime::try_from(sts_credentials.expiration.unwrap())
+                    .context(error::DateTimeToSystemTimeSnafu)?,
+            ),
```
etungsten: Is that a safe `unwrap`? What if it's `None`? Why can't we use the `sts_credentials.expiration` as is?

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,150 @@
...
-    where
-        P: ProvideAwsCredentials + Send + Sync + 'static,
-        D: DispatchSignedRequest + Send + Sync + 'static,
-    {
-        Self::new_with(request_dispatcher, credentials_provider, region)
+    // Add custom endpoint if specified in aws.region.REGION.endpoint.
+    let maybe_regional_endpoint = aws.region.get(&region).and_then(|r| r.endpoint.clone());
+    if let Some(endpoint) = &maybe_regional_endpoint {
+        config =
+            config.endpoint_resolver(Endpoint::immutable(endpoint.parse().expect("valid URI")));
```
etungsten: Let's avoid panicking if at all possible.
```suggestion
            config.endpoint_resolver(Endpoint::immutable(endpoint.parse().context(...)?));
```

etungsten (changes requested):
Can you please also test `cargo make check-repo-expirations`, `cargo make validate-repo`, and `cargo make refresh-repo`?

You can find examples in the PR descriptions here:
https://github.com/bottlerocket-os/bottlerocket/pull/2303
https://github.com/bottlerocket-os/bottlerocket/pull/1575

tools/pubsys/src/aws/ami/mod.rs
jpculp: I was running into borrow issues earlier, but it looks like they aren't there now.

tools/pubsys/src/repo.rs
jpculp: I only had it there to stay consistent with our implementation in [`tough-kms`](https://github.com/awslabs/tough/blob/develop/tough-kms/src/error.rs#L19), but removing it doesn't cause any issues that I'm aware of.

tools/pubsys/src/main.rs
jpculp: I set to filter at `Error` since that's the level that was set in `bottlerocket-test-system`, but `Warn` could probably make sense here.

tools/pubsys/src/main.rs
jpculp: Since `pubsys` will get a default value of `INFO` for `args.log_level`, I think this works as is. The filtering will apply even if the level isn't explicitly set, and if one were passed we would want that intentional log level applied to everything.

tools/pubsys/src/aws/publish_ami/mod.rs
jpculp: It doesn't matter, but I'll throw it back in to be closer to the original messaging.

jpculp:
- Removed extraneous clones and references.
- Set default log level for `aws_config` and `aws_smithy` to `Warn`.
- Improved error handling and logging.

etungsten:
Tested check-repo-expirations:
```
$ cargo make  -e BUILDSYS_VARIANT=aws-k8s-1.23 -e BUILDSYS_ARCH=x86_64 check-repo-expirations
[cargo-make] INFO - cargo make 0.35.15
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: check-repo-expirations
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Running Task: setup
[cargo-make] INFO - Running Task: fetch-sources
[cargo-make] INFO - Running Task: tuftool
[cargo-make] INFO - Running Task: publish-setup-tools
[cargo-make] INFO - Running Task: publish-setup
19:01:24 [INFO] Found infra config at path: /home/ec2-user/bottlerocket/Infra.toml
[cargo-make] INFO - Running Task: setup
[cargo-make] INFO - Running Task: fetch-sources
[cargo-make] INFO - Running Task: publish-tools
[cargo-make] INFO - Running Task: check-repo-expirations
[2022-09-19T19:01:26Z INFO  pubsys_config] Found infra config at path: /home/ec2-user/bottlerocket/Infra.toml
[2022-09-19T19:01:26Z INFO  pubsys::repo::check_expirations] Loaded TUF repo:   file:///home/ec2-user/bottlerocket/build/repos/default/bottlerocket-1.9.2-deea371c/aws-k8s-1.23/x86_64
[2022-09-19T19:01:26Z INFO  pubsys::repo::check_expirations] Root expiration:   2023-09-18 18:58:32 UTC
[2022-09-19T19:01:26Z INFO  pubsys::repo::check_expirations] Snapshot expiration:       2022-10-03 19:01:12.156262570 UTC
[2022-09-19T19:01:26Z INFO  pubsys::repo::check_expirations] Targets expiration:        2022-10-03 19:01:12.156262570 UTC
[2022-09-19T19:01:26Z INFO  pubsys::repo::check_expirations] Timestamp expiration:      2022-09-26 19:01:12.156262570 UTC
[2022-09-19T19:01:26Z INFO  pubsys::repo::check_expirations] Looking for metadata expirations happening from now to 2022-09-22 19:01:26.973914183 UTC
[cargo-make] INFO - Build Done in 4.72 seconds.
```

Tested `validate-repo`:
```
$ cargo make  -e BUILDSYS_VARIANT=aws-k8s-1.23 -e BUILDSYS_ARCH=x86_64 validate-repo
[cargo-make] INFO - cargo make 0.35.15
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: validate-repo
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Running Task: setup
[cargo-make] INFO - Running Task: fetch-sources
[cargo-make] INFO - Running Task: tuftool
[cargo-make] INFO - Running Task: publish-setup-tools
[cargo-make] INFO - Running Task: publish-setup
19:01:45 [INFO] Found infra config at path: /home/ec2-user/bottlerocket/Infra.toml
[cargo-make] INFO - Running Task: setup
[cargo-make] INFO - Running Task: fetch-sources
[cargo-make] INFO - Running Task: publish-tools
[cargo-make] INFO - Running Task: validate-repo
[2022-09-19T19:01:47Z INFO  pubsys_config] Found infra config at path: /home/ec2-user/bottlerocket/Infra.toml
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Loaded TUF repo: file:///home/ec2-user/bottlerocket/build/repos/default/bottlerocket-1.9.2-deea371c/aws-k8s-1.23/x86_64
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: migrate_v1.10.0_dns-settings.lz4
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: bottlerocket-aws-k8s-1.23-x86_64-v1.9.2.img.lz4
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: manifest.json
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: bottlerocket-aws-k8s-1.23-x86_64-1.9.2-deea371c-root.ext4.lz4
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: bottlerocket-aws-k8s-1.23-x86_64-1.9.2-deea371c-boot.ext4.lz4
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: bottlerocket-aws-k8s-1.23-x86_64-1.9.2-deea371c.img.lz4
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: aws-k8s-1.23-x86_64-kmod-kit-v1.9.2.tar.xz
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: bottlerocket-aws-k8s-1.23-x86_64-1.9.2-deea371c-data.img.lz4
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: bottlerocket-aws-k8s-1.23-x86_64-v1.9.2-data.img.lz4
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: bottlerocket-aws-k8s-1.23-x86_64-1.9.2-deea371c-root.verity.lz4
[2022-09-19T19:01:47Z INFO  pubsys::repo::validate_repo] Downloading target: migrate_v1.10.0_dns-settings-metadata.lz4
[cargo-make] INFO - Build Done in 4.25 seconds.

```

Tested `refresh-repo` with KMS key:
```
$ cargo make  -e BUILDSYS_VARIANT=aws-k8s-1.23 -e BUILDSYS_ARCH=x86_64 refresh-repo        
[cargo-make] INFO - cargo make 0.35.15
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: refresh-repo
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Running Task: setup
[cargo-make] INFO - Running Task: fetch-sources
[cargo-make] INFO - Running Task: tuftool
[cargo-make] INFO - Running Task: publish-setup-tools
[cargo-make] INFO - Running Task: publish-setup
19:05:28 [INFO] Found infra config at path: /home/ec2-user/bottlerocket/Infra.toml
[cargo-make] INFO - Running Task: publish-tools
[cargo-make] INFO - Running Task: refresh-repo
[2022-09-19T19:05:28Z INFO  pubsys_config] Found infra config at path: /home/ec2-user/bottlerocket/Infra.toml
[2022-09-19T19:05:28Z INFO  pubsys::repo::refresh_repo] Using repo expiration policy from path: /home/ec2-user/bottlerocket/tools/pubsys/policies/repo-expiration/2w-2w-1w.toml
[2022-09-19T19:05:28Z INFO  pubsys::repo::refresh_repo] Loaded TUF repo: file:///home/ec2-user/bottlerocket/build/repos/default/bottlerocket-1.9.2-deea371c-bak/aws-k8s-1.23/x86_64
[2022-09-19T19:05:28Z INFO  pubsys::repo] Setting non-root metadata expiration times:
        snapshot:  2022-10-03 19:05:28.522312124 UTC
        targets:   2022-10-03 19:05:28.522312124 UTC
        timestamp: 2022-09-26 19:05:28.522312124 UTC
[2022-09-19T19:05:28Z INFO  pubsys::repo::refresh_repo] Writing repo metadata to: /home/ec2-user/bottlerocket/build/repos/default/bottlerocket-1.9.2-deea371c/aws-k8s-1.23/x86_64
[cargo-make] INFO - Build Done in 2.40 seconds.
```

etungsten (approved):

tools/deny.toml
```diff
@@ -58,14 +58,17 @@ multiple-versions = "deny"
 wildcards = "deny"
 
 skip-tree = [
-    # temporarily using a different version of snafu
-    { name = "parse-datetime", version = "0.1.0" },
-
     # rusoto_signature uses an older version of sha2
     { name = "rusoto_signature" },
```
bcressey, line 62: This is still needed until infrasys is updated?

tools/pubsys/src/aws/ami/wait.rs
```diff
@@ -11,8 +11,8 @@ use std::time::Duration;
 /// `success_required` checks in a row.
 pub(crate) async fn wait_for_ami(
     id: &str,
-    region: &Region,
-    sts_region: &Region,
+    region: &str,
+    sts_region: &str,
```
bcressey: Why pass `&str` around as the type instead of converting to an `aws_types::region::Region` as early as possible and using that?

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,165 @@
...
+pub(crate) async fn build_client_config<S1, S2>(
+    region: S1,
+    sts_region: S2,
+    aws: &AwsConfig,
+) -> Result<SdkConfig>
+where
+    S1: AsRef<str>,
+    S2: AsRef<str>,
+{
+    let mut config = aws_config::from_env();
```
bcressey: Why construct a new config here instead of using the `aws` one that's passed in?

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,165 @@
...
-    where
-        P: ProvideAwsCredentials + Send + Sync + 'static,
-        D: DispatchSignedRequest + Send + Sync + 'static,
-    {
-        Self::new_with(request_dispatcher, credentials_provider, region)
-    }
-}
+pub(crate) async fn build_client_config<S1, S2>(
+    region: S1,
+    sts_region: S2,
```
bcressey: Why can't these be `Region` types instead?

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,165 @@
...
+    aws: &AwsConfig,
```
bcressey: This should have a more descriptive name:
```suggestion
    pubsys_aws_config: &PubsysAwsConfig,
```

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,165 @@
-use async_trait::async_trait;
+use aws_config::profile::ProfileFileCredentialsProvider;
+
+use aws_config::{ConfigLoader, SdkConfig};
+use aws_sdk_ec2::{Endpoint, Region};
+use aws_sdk_sts::model::Credentials as StsCredentials;
+use aws_sdk_sts::{Client as StsClient, Credentials as SdkCredentials};
 use pubsys_config::AwsConfig;
```
bcressey: I found the client builder code very confusing and the ambiguous name of this type was a big part of why.

I would recommend either renaming this type to something like `AwsConfigOverride` or importing it differently:
```suggestion
use pubsys_config::AwsConfig as PubsysAwsConfig;
```

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,165 @@
...
-impl ProvideAwsCredentials for CredentialsProvider {
-    async fn credentials(&self) -> std::result::Result<AwsCredentials, CredentialsError> {
-        self.0.credentials().await
-    }
+    // Load a chained credential config if role is specified in aws.region.REGION.role.
+    let maybe_regional_role = aws.region.get(&region).and_then(|r| r.role.clone());
+    let config = if maybe_regional_role.is_some() {
+        let assume_roles = aws.role.iter().chain(maybe_regional_role.iter()).cloned();
+        let base_config = config.region(Region::new(sts_region)).load().await;
+        build_chained_config(&region, assume_roles.clone(), base_config).await?
```
bcressey: I may be mis-reading this but it looks like we won't build the chained config unless we have a region-specific role. If there's a top level role in `aws.role` we also need to assume that using the profile credentials, even if there's no regional role.

I.e. this `Infra.toml` would work as expected:
```
[aws]
# If specified, we assume this role before making any API calls.
role = "arn:aws:iam::012345678901:role/assume-global"

[aws.region.us-west-2]
# If specified, we assume this role before making any API calls in this region.
# (This is assumed after the "global" aws.role, if that is also specified.)
role = "arn:aws:iam::012345678901:role/assume-regional"
```

But this one would not:
```
[aws]
# If specified, we assume this role before making any API calls.
role = "arn:aws:iam::012345678901:role/assume-global"
```

We really ought to have unit tests for these if that's possible.

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,165 @@
...
+            PROVIDER_NAME,
         );
-        provider = CredentialsProvider(Box::new(
-            AutoRefreshingProvider::new(expiring_provider).context(error::ProviderSnafu)?,
-        ));
+        sts_config = aws_config::from_env()
+            .credentials_provider(credentials.clone())
+            .region(Region::new(region.clone()))
+            .load()
+            .await;
```
bcressey: Does this end up passing the custom endpoints all the way down the chain? I would guess not because we start with a fresh `aws_config::from_env()` each time.

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,165 @@
...
+                    role_arn: &assume_role,
+                })?,
+            sts_credentials.session_token().map(|s| s.to_string()),
+            Some(
+                sts_credentials
+                    .expiration
+                    .context(error::DateTimeExpirationSnafu)?
+                    .try_into()
+                    .context(error::DateTimeToSystemTimeSnafu)?,
+            ),
```
bcressey: It seems like if we somehow got `sts_credentials` with no expiration, we should instead pass `None` into this field of `SdkCredentials`, rather than blowing up.

tools/pubsys/src/aws/mod.rs
```diff
@@ -10,24 +8,11 @@ pub(crate) mod promote_ssm;
...
-}
-
 /// Parses the given string as an architecture, mapping values to the ones used in EC2.
-pub(crate) fn parse_arch(input: &str) -> Result<String> {
+pub(crate) fn parse_arch(input: &str) -> Result<ArchitectureValues> {
     match input {
-        "x86_64" | "amd64" => Ok("x86_64".to_string()),
-        "arm64" | "aarch64" => Ok("arm64".to_string()),
+        "x86_64" | "amd64" => Ok(ArchitectureValues::X8664),
+        "arm64" | "aarch64" => Ok(ArchitectureValues::Arm64),
```
bcressey, line 21: 😬 

Not your fault but the upstream enum and variant names are really unfortunate.

tools/pubsys/src/aws/promote_ssm/mod.rs
```diff
@@ -65,10 +65,7 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
         promote_args.regions.clone()
     } else {
         aws.regions.clone().into()
-    }
-    .into_iter()
-    .map(|name| region_from_string(&name, &aws).context(error::ParseRegionSnafu))
-    .collect::<Result<Vec<Region>>>()?;
```
bcressey: I don't love dropping the conversion into the more accurate types here at the edges where arguments are parsed. Even if the conversion is trivial and infallible, I'd still prefer to pass around `Region` and `&Region` rather than strings.

tools/pubsys/src/aws/promote_ssm/mod.rs
```diff
@@ -78,13 +75,17 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
     );
     let base_region = &regions[0];
 
+    let arch = promote_args.arch.as_str();
```
bcressey: Why convert this back into a string from the more accurate type? That doesn't seem correct.

tools/pubsys/src/aws/ami/wait.rs
```diff
@@ -71,13 +71,18 @@ pub(crate) async fn wait_for_ami(
...
-                                .iter()
-                                .any(|e| e == found_state),
+                            ![
+                                ImageState::Invalid,
+                                ImageState::Deregistered,
+                                ImageState::Failed,
+                                ImageState::Error
+                            ]
+                            .iter()
+                            .any(|e| e == &found_state),
```
bcressey: A `match` on the known failure states would be more idiomatic.

tools/pubsys/src/repo.rs
```diff
@@ -425,17 +425,33 @@ fn get_signing_key_source(signing_key_config: &SigningKeyConfig) -> Result<Box<d
...
+        // We are cloning this so that we can send it across a thread boundary
+        let region = region.to_owned();
+        // We need to spin up a new thread to deal with the async nature of the
+        // AWS SDK Rust
+        thread::spawn(move || {
+            let runtime = tokio::runtime::Runtime::new().context(error::RuntimeCreationSnafu)?;
+            Ok(Some(runtime.block_on(async_get_client(&region))))
+        })
+        .join()
+        .map_err(|_| error::Error::ThreadJoin {})?
```
bcressey: I'd prefer to handle this the same way that `main.rs` handles conditionally creating a Tokio runtime around certain subcommands.

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -348,9 +345,9 @@ pub(crate) async fn modify_snapshots(
 pub(crate) async fn modify_regional_snapshots(
     user_ids: Option<Vec<String>>,
     group_names: Option<Vec<String>>,
-    operation: &str,
-    snapshots: &HashMap<Region, Vec<String>>,
-    clients: &HashMap<Region, Ec2Client>,
+    operation: OperationType,
```
bcressey: Can this just take a reference instead, so that callers don't have to clone it?
```suggestion
    operation: &OperationType,
```

(Same note for the other functions that now take an `OperationType`.)

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -456,27 +447,30 @@ pub(crate) async fn modify_regional_images(
...
+        std::result::Result<ModifyImageAttributeOutput, SdkError<ModifyImageAttributeError>>,
+    )> = request_stream.collect().await;
 
     // Count up successes and failures so we can give a clear total in the final error message.
     let mut error_count = 0u16;
     let mut success_count = 0u16;
     for ((region, image_id), modify_image_response) in responses {
         match modify_image_response {
-            Ok(()) => {
+            Ok(_) => {
```
bcressey, line 472: Is there any useful data in the response that's being discarded here? As a hypothetical, if it contained the `image_id`, that would let us simplify the tuple type, or ensure that it matched the expected ID.

bcressey (changes requested):

tools/deny.toml
jpculp: Correct.

tools/pubsys/src/aws/ami/register.rs
```diff
@@ -74,17 +72,17 @@ async fn _register_image(
...
-    };
+    let root_bdm = BlockDeviceMapping::builder()
+        .set_device_name(Some(ROOT_DEVICE_NAME.to_string()))
+        .set_ebs(Some(
+            EbsBlockDevice::builder()
+                .set_delete_on_termination(Some(true))
+                .set_snapshot_id(Some(root_snapshot.clone()))
+                .set_volume_type(Some(VolumeType::from(VOLUME_TYPE)))
+                .set_volume_size(ami_args.root_volume_size)
+                .build(),
```
ecpullen, line 85: The sdk allows not including `Some` if you set the values without `set_`. (This happens in a lot of places so might not be in the scope of this pr)
```suggestion
        .device_name(ROOT_DEVICE_NAME.to_string())
        .ebs(EbsBlockDevice::builder()
                .delete_on_termination(true)
                .snapshot_id(root_snapshot.clone())
                .volume_type(VolumeType::from(VOLUME_TYPE))
                .set_volume_size(ami_args.root_volume_size)
                .build(),
```

jpculp: This definitely works, but my understanding was the setters are more idiomatic for the SDK.

tools/pubsys/src/aws/promote_ssm/mod.rs
jpculp: In the SSM modules, `BuildContext` is building a template of **str**-strings, so this was just to simplify code. The AMI modules are the only ones that take advantage of the stronger type. That said, it is probably a good idea to switch to the string at the point that you need it to be one.

tools/pubsys/src/aws/client.rs
rpkelly: This actually might be really important if the passed in aws config uses a different profile than the default somehow.

jpculp: This aws_config is the SDK's aws_config. Hopefully will be clearer after the rename.

tools/pubsys/src/aws/publish_ami/mod.rs
jpculp: The response is a [ModifyImageAttributeOutput](https://docs.rs/aws-sdk-ec2/latest/aws_sdk_ec2/output/struct.ModifyImageAttributeOutput.html) type, but I couldn't find a good way to make use of it directly for a success message.

jpculp:
- Brought back stronger-typed regions and `region_from_string()`.
- Renamed `aws: AwsConfig` to `pubsys_aws_config: PubsysAwsConfig` to reduce confusion.
- Kept `arch` as a stronger-type throughout.
- Replaced **ImageState** `ensure` statement with `match`.
- Replaced `thread::spawn` with `tokio::runtime::Runtime::new()`.

tools/pubsys/src/aws/mod.rs
```diff
@@ -10,24 +9,16 @@ pub(crate) mod promote_ssm;
...
-    Ok(match maybe_endpoint {
-        Some(endpoint) => Region::Custom {
-            name: name.to_string(),
-            endpoint,
-        },
-        None => name.parse().context(error::ParseRegionSnafu { name })?,
-    })
+/// Builds a Region from the given region name.
+fn region_from_string(name: &str) -> Result<Region> {
+    Ok(Region::new(name.to_owned()))
```
bcressey: If this is infallible now we should probably just return a `Region`. However, I like it this way for the purposes of the SDK migration, to minimize the diff. Maybe add a `FIXME` to rework later.

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,169 @@
...
+        .region
+        .get(region.as_ref())
+        .and_then(|r| r.endpoint.clone());
+    if let Some(endpoint) = &maybe_regional_endpoint {
+        config = config.endpoint_resolver(Endpoint::immutable(
+            endpoint
+                .parse()
+                .context(error::UriParseSnafu { input: endpoint })?,
+        ));
+    }
```
bcressey: The SDK comment for `endpoint_resolver` says:
```
/// This method will override the endpoint resolver used for **all** AWS services. This mainly                                               
/// exists to set a static endpoint for tools like `LocalStack`. For live traffic, AWS services                                              
/// require the service-specific endpoint resolver they load by default.    
```

Do we actually use that endpoint functionality for anything today? If not, we should just retract the option in `PubsysAwsConfig`.

To handle it correctly, we'd need service-specific endpoint fields in `Infra.toml`, and SDK clients would need to be built to use those endpoints. That would happen in the calling code rather than here.

tools/pubsys/src/aws/client.rs
```diff
@@ -1,159 +1,169 @@
...
-    fn new_with<P, D>(request_dispatcher: D, credentials_provider: P, region: Region) -> Self
-    where
-        P: ProvideAwsCredentials + Send + Sync + 'static,
-        D: DispatchSignedRequest + Send + Sync + 'static,
-    {
-        Self::new_with(request_dispatcher, credentials_provider, region)
-    }
-}
+use aws_config::{ConfigLoader, SdkConfig};
+use aws_sdk_ec2::{Endpoint, Region};
```
bcressey: nit: maybe we should use `aws_sdk_sts::Region` here, since that's the only service we end up talking to.

tools/pubsys/src/aws/ami/register.rs
bcressey: Where did you see the perspective on idiomatic use? Not disagreeing, and it's not a blocker here in any case, but for me the code that @ecpullen wrote seems preferable since it's fewer characters and we don't have to wrap everything in an `Option`.

tools/pubsys/src/aws/client.rs
jpculp: I looked into it, but I wasn't able to find anywhere we are using it today.

tools/pubsys/src/aws/mod.rs
jpculp: This is a pretty quick change. Rather than a `FIXME`, should I just make it the final change once we think this PR hits all the edge cases?

jpculp:
- Changed credentials logic will run even if only a single role is specified.
- Credential expirations can now be None.
- Removed endpoints.

bcressey (approved):

etungsten (approved):

zmrow (approved):
:safety_vest:

jpculp:
- Rewrote client.rs to return providers instead of credentials.
- Removed returning unneccessary result-types (and their associated error contexts).

jpculp:
Suppressed noisy `[2022-09-22T20:01:26Z INFO  tracing::span] assume_role;` log.

ecpullen (approved):

etungsten (approved):

tools/pubsys/src/aws/ami/register.rs
jpculp: It was just an impression we got from the SDK team that the non-setter way might eventually be deprecated.

zmrow (approved):
:microscope:

bcressey (approved):

ecpullen (approved):

PR #2428, opened 2022-09-15
update to SDK v0.27.0

### LINKED ISSUE #2175: Many packages require newer build tools in the SDK
**What I'd like:** To update dbus-broker to v31 or greater.

dbus-broker requires a newer build of meson to be included in our SDK. Currently, the SDK [uses F35 as a base](https://github.com/bottlerocket-os/bottlerocket-sdk/blob/c726347768074d39d499c8e5b235a174756c1438/Dockerfile#L1), which doesn't include a recent enough build, so we would need to update that.

Note that when we do update dbus-broker, we also need to change our systemd unit to account for the following changes:
* https://github.com/bus1/dbus-broker/commit/28af5ac996c94922fc49660e0efa955f57575187
* https://github.com/bus1/dbus-broker/commit/fa19773b27775ffa81c248de1ce4b8febb526adf

**Any alternatives you've considered:** N/A.

### LINKED ISSUE #2418: update glibc to 2.36
glibc 2.36 is available: https://sourceware.org/pipermail/libc-alpha/2022-August/141193.html

We're also bumping glibc to 2.36 as part of the bottlerocket-sdk update: https://github.com/bottlerocket-os/bottlerocket-sdk/pull/82

### LINKED ISSUE #2419: update binutils to 2.38
`bottlerocket-sdk` is moving to `binutils` v2.38 in https://github.com/bottlerocket-os/bottlerocket-sdk/pull/82. We should match that version when the new SDK comes out.

bcressey:
**Issue number:**
Fixes #2419
Fixes #2418
Fixes #2175

**Description of changes:**
Update to the latest SDK, from https://github.com/bottlerocket-os/bottlerocket-sdk/pull/82.

This adds two workarounds for new `rpm` behavior in Fedora 36, and syncs glibc and binutils with the versions found in the SDK. The version of Rust used for CI builds is also updated to match 1.63.0 in the SDK. The newer meson lets us update `dbus-broker` to v32, and requires some minor changes to `iputils` to drop unknown options.

The only controversial change may be the addition of `-Bsymbolic-functions` to the default linker flags. There's some light background reading [here](https://maskray.me/blog/2021-05-16-elf-interposition-and-bsymbolic) and [here](https://maskray.me/blog/2021-05-09-fno-semantic-interposition). The quick summary is that I don't really expect problems since we've been using `-fno-semantic-interposition` on the compiler side since 237946dc43e04b94f2aae429eaae16c71a520263, so it's hopefully just free performance at runtime from simplifying the dynamic linker's job. 🤞🏻 

**Testing done:**
Ran the quick tests for `aws-*` using testsys. Everything passed except for `aws-k8s-1.20` on `aarch64`, which needs a custom conformance image, and `aws-k8s-1.23`, which needs a [newer eksctl](https://github.com/bottlerocket-os/bottlerocket-test-system/issues/573).

Verified that `aws-k8s-1.22-nvidia` still works. Ran some basic smoke tests for `vmware-dev` and `vmware-k8s-1.22`.

**Terms of contribution:**


.github/workflows/build.yml
```diff
@@ -104,7 +104,7 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
-      - run: rustup default 1.61.0
+      - run: rustup default 1.63.0
       - run: cargo install --version 0.35.12 cargo-make
```
jpculp: Should we bump `cargo-make` as well?

Dockerfile
```diff
@@ -166,6 +166,7 @@ RUN --mount=target=/host \
         -x '*-debugsource-*.rpm' \
         --no-database \
         /host/build/rpms \
+    && echo '%_dbpath %{_sharedstatedir}/rpm' >> /etc/rpm/macros \
```
arnaldo2792, line 169: For the lack of a better place to ask a question about something mentioned in the commit message (and for my own understanding):

Why should `/usr`  be a symlink to the sysroot directory?

@bcressey kindly replied offline:

> If it is anything other than a symlink, then packages from Bottlerocket could clash with packages from Fedora at build time.
>By keeping the two roots separate until the last minute, we avoid some conflicts and chances of pollution. The symlink lets us have normal paths at runtime.

macros/shared
```diff
@@ -15,8 +15,8 @@
   '-fexceptions', '-fstack-clash-protection', '-fno-semantic-interposition' \
   %{nil}}
 %_cross_cxxflags %_cross_cflags
-%_cross_ldflags -Wl,-z,relro -Wl,-z,now
-%_cross_c_link_args '-Wl,-z,relro', '-Wl,-z,now'
+%_cross_ldflags -Wl,-z,relro -Wl,-z,now -Wl,-Bsymbolic-functions
```
arnaldo2792, line 18: I searched for the flag and it seems like it is used to improve library load times? Do you have a link to a blogpost or something like what you shared in other commits (the links to the fedora docs), that explains why this is a good addition to the list of flags? The man pages aren't that descriptive, and I'm not a `gcc/binutils` expert so I don't fully understand the impact of this or its benefits :sweat_smile: :

> tells the linker to bind references to global function symbols to the definition inside the shared library

packages/glibc/glibc.spec
```diff
@@ -13,39 +13,17 @@ Source3: ldconfig-service.conf
...
-Patch0020: 0020-hppa-Fix-bind-now-audit-BZ-28857.patch
-Patch0021: 0021-Fix-elf-tst-audit2-on-hppa.patch
-Patch0022: 0022-hppa-Fix-swapcontext.patch
-Patch0023: 0023-hppa-Revise-gettext-trampoline-design.patch
-Patch0024: 0024-hppa-Fix-warnings-from-_dl_lookup_address.patch
-Patch0025: 0025-nptl-Fix-cleanups-for-stack-grows-up-BZ-28899.patch
-Patch0026: 0026-io-Add-fsync-call-in-tst-stat.patch
-Patch0027: 0027-nss-Do-not-mention-NSS-test-modules-in-gnu-lib-names.patch
-Patch0028: 0028-nss-Protect-against-errno-changes-in-function-lookup.patch
+Patch001: 0001-stdlib-Suppress-gcc-diagnostic-that-char8_t-is-a-key.patch
```
arnaldo2792: There are 39 patches (or at least that's the last number in the patches' names), but only 6 are applied, is there a reason why the remaining patches are included but not applied? Also there are two patches that modify `NOTES` and `NEWS`, are they required even if they don't make any functional change?

tools/testsys/src/aws_resources.rs
```diff
@@ -391,11 +391,9 @@ impl AwsEcs {
         testsys_images: &TestsysImages,
     ) -> Result<Vec<Crd>> {
         match test {
-            TestType::Conformance => {
-                return Err(anyhow!(
-                    "Conformance testing for ECS variants is not supported."
-                ))
-            }
+            TestType::Conformance => Err(anyhow!(
```
arnaldo2792: Doesn't seem like this change is related to the other commits in the series :stuck_out_tongue:, should it be part of a different PR?

bcressey: It's related to the update to Rust 1.63.0 in the actions workflow, but it looks like @stmcginnis is fixing it in #2443. I'll rebase and drop the patch when that's merged.

packages/glibc/glibc.spec
bcressey: Whoops - that's my mistake. There were only six patches when I first added glibc 2.36, then the number grew 😀 I added the patches but forgot to update the spec.

As for the patches modifying `NOTES` and `NEWS` - the patch series is more maintainable if it's just "all patches to 2.36 since release" rather than cherry-picking a subset.

macros/shared
bcressey: I shared these links in the description:
> There's some light background reading [here](https://maskray.me/blog/2021-05-16-elf-interposition-and-bsymbolic) and [here](https://maskray.me/blog/2021-05-09-fno-semantic-interposition).

bcressey:
Force push above:
* applies all the patches in `glibc.spec`
* updates `cargo-make` as well
* drops the clippy fix from 23255a829f96b6464ebf116b06d9f724ee1d17be

bcressey:
Force push for a rebase to pull in the lint fix.

jpculp (commented):
LGTM 🚀

bcressey:
Force push applies `go fmt` from Go 1.19.

packages/iputils/iputils.spec
```diff
@@ -39,8 +38,6 @@ CONFIGURE_OPTS=(
  -DBUILD_NINFOD=false
  -DBUILD_RARPD=false
  -DBUILD_RDISC=false
- -DBUILD_TFTPD=false
- -DBUILD_TRACEROUTE6=false
```
markusboehme, line 43: I was curious why these cause trouble now, since they are/should be defined by iputils and the iputils build didn't change. All those options were already gone in the iputils build referenced here, but it wasn't until Meson 0.60 that passing unkonwn configure options causes build failures rather than warnings ([source](https://mesonbuild.com/Release-notes-for-0-60-0.html#unknown-options-are-now-always-fatal)).

packages/binutils/binutils.spec
```diff
@@ -31,7 +31,6 @@ Requires: %{name}
   --disable-gdb \
   --with-system-zlib \
   --without-gnu-as \
-  --disable-plugins \
```
markusboehme, line 34: I was unaware of libdep before this change, but it seems the libdep plugin only was introduced with binutils 2.36. It is not included in the current/to be replaced version of binutils. With this, I don't understand the intention of the change/commit message. Is the libdep plugin built independently of the choice to `--enable-plugins`/`--disable-plugins`?

It also seems the libdep plugin is useful when linking against static libraries, since transitive dependencies can be pulled in without having to explicitly specify them. This doesn't seem too useful on a Bottlerocket host. Is this just to more closely mirror the version of binutils there is in the SDK?

bcressey: I agree that my commit message could be more clear on what motivated the change.

The upstream configure script no longer has an option to disable building plugins, likely because libdep was added in a way that causes it to always be built.

I could try to patch it out but opted to just go with it. I don't think it will be useful in this project, but it also seems harmless to have around.

bcressey: At least IIRC the configure option is gone. But either way it has no effect on libdep being built.

markusboehme: Agree it's no big deal to ship libdep. Thanks for clarifying your reasoning!

jpculp (approved):

stmcginnis (approved):
Tests are happy and everything looks fine to me.

zmrow (approved):
:icecream:

arnaldo2792 (approved):

markusboehme (approved):

PR #2436, opened 2022-09-19
k8s: increase the kube-api-server QPS from 5/10 to 10/20

tzneal:
**Issue number:**

N/A

**Description of changes:**

This changes the default setting for EKS v1.22+ where API Priority & Fairness is available and there is a specific queue for kubelet health.

**Testing done:**

`cargo make unit-tests` and 

- Built a custom K8s v1.23 Bottlerocket AMI
- Launched a node using that AMI
- Verified that the config changes were applied by logging in via SSM
```
bash-5.1# cat /etc/kubernetes/kubelet/config  | grep kubeAPI
kubeAPIQPS: 10
kubeAPIBurst: 20
```

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.22/kubelet-config
+++ b/packages/kubernetes-1.22/kubelet-config
@@ -56,12 +56,8 @@ eventRecordQPS: {{settings.kubernetes.event-qps}}
 {{#if settings.kubernetes.event-burst includeZero=true}}
 eventBurst: {{settings.kubernetes.event-burst}}
 {{/if}}
-{{#if settings.kubernetes.kube-api-qps includeZero=true}}
-kubeAPIQPS: {{settings.kubernetes.kube-api-qps}}
-{{/if}}
-{{#if settings.kubernetes.kube-api-burst includeZero=true}}
-kubeAPIBurst: {{settings.kubernetes.kube-api-burst}}
-{{/if}}
+kubeAPIQPS: {{default 10 settings.kubernetes.kube-api-qps}}
+kubeAPIBurst: {{default 20 settings.kubernetes.kube-api-burst}}
 kubeReserved:
   cpu: "{{kube_reserve_cpu settings.kubernetes.kube-reserved.cpu}}"
   {{#if settings.kubernetes.kube-reserved.memory}}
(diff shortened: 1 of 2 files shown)


tzneal:
In testing increasing from 5 QPS to 10 QPS effectively halves the overall time for 100 pods to become ready on a single node from 67.5 to 32.8 seconds while also halving the time for the first pod to become ready from to 22 seconds to 10 second.
image
For larger scale-up events (in this case 3k pods and 30 nodes) the benefits above 10/20 QPS settings decrease, though a large benefit still remains if the values are increased from the default 5/10 QPS.
image

Any change in the kubelet QPS settings is scaled by the number of nodes in each cluster. This poses a risk that we can overwhelm the kube-api-server by increasing the amount of traffic received from kubelet running on each node. The primary mitigation to this risk is that for K8s v1.20+ the [API Priority & Fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/) feature is enabled in EKS. This added to Kubernetes the concept of priority levels where requests in different priority levels cannot starve each other. In v1.22+ specifically there is a separate queue for K8s health checks which will prevent increased load from pod churn from interfering with Kubelet health reporting. The increase here is small while realizing most of the benefit, which further decreases negative impact risks.

etungsten (approved):
> This changes the default setting for EKS v1.22+

These kubelet configs are also used in non-EKS K8s variants as well, e.g. `metal-k8s-*`, `vmware-k8s-*`. 

I think increasing the defaults still makes sense. I also agree with the assessment that the benefits here outweighs any potential downsides.

jpmcb (approved):
Very nice! 🏃🏼

zmrow (approved):
Thanks!

bcressey (approved):
I really appreciate the focus on metrics and data here. Looks like a nice win.

tzneal:
> I really appreciate the focus on metrics and data here. Looks like a nice win.

Thanks! I don't have permissions on this repo to merge, so feel free to merge whenever.

arnaldo2792:
I'm merging since all tests passed (thanks @tzneal!)

PR #2439, opened 2022-09-19
testsys: Support environment kubecontext

### LINKED ISSUE #2426: testsys: Support other kubeconfig locations
**What I'd like:**

When using a kubeconfig not in `testsys.kubeconfig`, `cargo make test` will error. Instead it should check to see if the provided kubeconfig file path exists and if not, use the default kubeconfig.

ecpullen:
**Issue number:**

Closes #2426 

**Description of changes:**

Enables testsys to use an environments kubeconfig if the kubeconfig path doesn't exist.

**Testing done:**

Tested with kubeconfig in `testsys.kubeconfig` and with `kind create cluster`.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -127,8 +127,8 @@ AMI_DATA_FILE_SUFFIX = "amis.json"
 #    5: a final `quick` test on the downgraded instances
 # TESTSYS_STARTING_IMAGE_ID can be used to provide the correct starting image for migration tests.
 TESTSYS_TEST = "quick"
-# The path to the testsys cluster's kubeconfig file. This is used for all testsys calls.
-TESTSYS_KUBECONFIG_PATH = "${BUILDSYS_ROOT_DIR}/testsys.kubeconfig"
+# The default path to the testsys cluster's kubeconfig file. This is used for all testsys calls.
+CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH = "${BUILDSYS_ROOT_DIR}/testsys.kubeconfig"
 # The last released version of bottlerocket.
 TESTSYS_STARTING_VERSION = { script = ["git tag --list --sort=version:refname 'v*' | tail -1"] }
 # The commit for the last release of bottlerocket.
@@ -183,6 +183,23 @@ VMWARE_VM_NAME_DEFAULT = "${BUILDSYS_NAME}-${BUILDSYS_VARIANT}-${BUILDSYS_ARCH}-
 BOOT_CONFIG_INPUT = "${BUILDSYS_ROOT_DIR}/bootconfig-input"
 # Boot Configuration initrd
 BOOT_CONFIG = "${BUILDSYS_ROOT_DIR}/bootconfig.data"
+
+# Determines the kubeconfig that should be used by testsys. If no kubeconfig was provided and the
+# default kubeconfig location does not exist, use the users default kubeconfig.
+CARGO_MAKE_TESTSYS_KUBECONFIG_ARG = {script = [
+'''
+if [ -n "${TESTSYS_KUBECONFIG}" ]; then
+   # If the user provides a kubeconfig path it should be used.
+   echo "--kubeconfig ${TESTSYS_KUBECONFIG}"
+elif [ -f "${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}" ]; then
+   # If the default kubeconfig exists it should be used.
+   echo "--kubeconfig ${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}"
+fi
+'''
+]}
+
+# Args that will be passed into all testsys invocations.
+CARGO_MAKE_TESTSYS_ARGS = "${CARGO_MAKE_TESTSYS_KUBECONFIG_ARG}"
 
 [tasks.setup]
 script = [
@@ -1507,7 +1524,7 @@ script = [
     '''
     set -eu
     export PATH="${BUILDSYS_TOOLS_DIR}/bin:${PATH}"
-    testsys --kubeconfig ${TESTSYS_KUBECONFIG_PATH} install
+    testsys ${CARGO_MAKE_TESTSYS_ARGS} install
     '''
 ]
 
@@ -1526,7 +1543,7 @@ script = [
     export PATH="${BUILDSYS_TOOLS_DIR}/bin:${PATH}"
     # The ami that is selected from `amis.json` is determined by `TESTSYS_REGION` if set; otherwise,
     # it is the first region listed in `Infra.toml` (for aws variants).
-    testsys --kubeconfig ${TESTSYS_KUBECONFIG_PATH} run ${TESTSYS_TEST} \
+    testsys ${CARGO_MAKE_TESTSYS_ARGS} run ${TESTSYS_TEST} \
       --ami-input "${ami_input}" \
       ${TESTSYS_AWS_SECRET_NAME:+--secret ${TESTSYS_AWS_SECRET_NAME}} \
     '''
@@ -1539,7 +1556,7 @@ script = [
     '''
     set -eu
     export PATH="${BUILDSYS_TOOLS_DIR}/bin:${PATH}"
-    testsys --kubeconfig ${TESTSYS_KUBECONFIG_PATH} delete
+    testsys ${CARGO_MAKE_TESTSYS_ARGS} delete
     '''
 ]
 
@@ -1550,7 +1567,7 @@ script = [
    '''
    set -eu
    export PATH="${BUILDSYS_TOOLS_DIR}/bin:${PATH}"
-   testsys --kubeconfig ${TESTSYS_KUBECONFIG_PATH} uninstall
+   testsys ${CARGO_MAKE_TESTSYS_ARGS} uninstall
    '''
 ]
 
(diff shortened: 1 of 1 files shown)


tools/testsys/src/main.rs
```diff
@@ -34,16 +34,23 @@ struct TestsysArgs {
...
     #[clap(subcommand)]
     command: Command,
 }
 
 impl TestsysArgs {
     async fn run(self) -> Result<()> {
-        let client = match self.kubeconfig {
+        let kubeconfig_path = self
+            .kubeconfig
+            .filter(|kubeconfig| !self.default || kubeconfig.exists());
```
arnaldo2792: Following our offline discussion:

I didn't read the code properly. Now it makes senses but I have a different complain, even though `-d` is always set, it will be ignored if `TESTSYS_KUBECONFIG_PATH` exists, that makes me feel like the CLI isn't respecting the instructions set with the `testsys -d` command, which clearly says use the default.

If your intention is to always use `TESTSYS_KUBECONFIG_PATH` regardless of the flags, then I think what I suggested fits better: don't add a flag, and just express the precedence.

```
// pseudo-code
if path.is_some() && path.exists() {
    TestManager::from_file(&path)
} else {
     TestManager::new();
}
```

With something similar to ^, you respect the instruction given in the command:

```bash
# Use env variable
export TESTSYS_KUBECONFIG_PATH=here; testsys logs --test

# Use default
testsys logs --test
```

It is easier to document as well, otherwise the documentation will be like "if ENV is set, it will be used even if `-d` is passed" (I don't like when CLIs aren't clear with their flags :sweat_smile:)

ecpullen: The only problem with that is it may not do what the user expects.
Assume my kubeconfig is `my-kubeconfig.yaml`.
If I ran `testsys --kubeconfig my_kubeconfig.yaml status` (instead of  `testsys --kubeconfig my-kubeconfig.yaml status`), it would use my default context which I clearly didn't want since I specified a kubeconfig.

arnaldo2792: Not quite, `TESTSYS_KUBECONFIG_PATH` is mapped to the `kubeconfig` option, so `self.kubeconfig` will be "something" and technically these two will be interchangeable:

```bash
# This won't use the default
testsys --kubeconfig <path-to-config> <blah>
# Neither will this
export TESTSYS_KUBECONFIG_PATH=<path-to-config> testsys <blah>
```

And only this will use `default`, without any flags passed:
```bash
# `kubeconfig` isn't set, so use the default kubeconfig
testsys <blah>
```

And ^ still mimics what other CLIs do: use `$HOME/.kube/config` unless `--kubeconfig` (or in your case `TESTSYS_KUBECONFIG_PATH`) is set.

ecpullen: How would that work with `cargo make` having a default kubeconfig location set? Then cargo make would have to check to see if `TESTSYS_KUBECONFIG_PATH` exists. If it exists pass it into the command if not ignore it. Is that the implementation you would prefer? I prefer the command handling the file path logic in the CLI instead of Makefile scripts.

bcressey: I agree with @arnaldo2792 - I wouldn't expect to have to pass a flag to use the default. I also would say it's more robust to have `testsys` get a known-good set of arguments, and to error out if told to use a kubeconfig that doesn't exist, since otherwise that could mask a misconfiguration.

Since `cargo make` is invoking `testsys` and sort of owns deciding how to invoke based on the repo state, it makes sense for it to test whether `${BUILDSYS_ROOT_DIR}/testsys.kubeconfig` exists, and if it doesn't, just not pass that argument.

I might like to generalize it a bit and populate `CARGO_MAKE_TESTSYS_ARGS`, and then just always pass that field into `testsys` invocations - similar to how `CARGO_MAKE_CARGO_ARGS` works.

(You can also add `TESTSYS_ARGS` at the same time so there's a variable that's ready to go for overrides that won't change the default args - similar to `CARGO_BUILD_ARGS`.)

ecpullen:
^ Updated to remove `-d` from testsys commands.

The kube context is now determined by in order:
User provided kubeconfig path `TESTSYS_KUBECONFIG_PATH`
Default kubeconfig path `testsys.kubeconfig` (if it exists it is used)
User kubecontext

Tested with a kind cluster context, eks context, setting the path with env variables and using the default path.

Makefile.toml
```diff
@@ -127,13 +127,30 @@ AMI_DATA_FILE_SUFFIX = "amis.json"
...
+   '''
+      if [[ -n ${TESTSYS_KUBECONFIG_PATH} ]]; then
+         # If the user provides a kubeconfig path it should be used.
+         echo "--kubeconfig ${TESTSYS_KUBECONFIG_PATH}"
+      elif [[ -f "${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}" ]]; then
+         # If the default kubeconfig exists it should be used.
+         echo "--kubeconfig ${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}"
+      fi
+   '''
+]}
```
bcressey: Is there a need for both `TESTSYS_KUBECONFIG_PATH` and `CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH` ?

Couldn't we just keep the old default value and then they could override the kubeconfig by setting that one variable via `-e TESTSYS_KUBECONFIG_PATH`?

If you tried that and it didn't work, it might help to move this into the `env.development` block. I believe dependencies between env variables was a feature added in a very recent `cargo-make` release, but I don't quite want to rely on that yet.

ecpullen: If we only use `TESTSYS_KUBECONFIG_PATH` there is no way to determine if the user was trying to override the path or if it was the default path. I'll move it to `env.development`.

ecpullen:
^ Move `CARGO_MAKE_TESTSYS_KUBECONFIG_ARG` to `env.development`

Makefile.toml
stmcginnis: Do we need to know that? It seems like it would be nice if this worked like some of our other overrides (e.g. `-e BUILDSYS_VARIANT`) where if not specified it just uses the default, otherwise the user can override the variable with their custom location.

ecpullen: The main issue is, if the user supplies a kubeconfig, but there is a typo and it doesn't exist, it would default to using an unplanned kubeconfig which should not happen.

Makefile.toml
```diff
@@ -184,6 +184,23 @@ BOOT_CONFIG_INPUT = "${BUILDSYS_ROOT_DIR}/bootconfig-input"
 # Boot Configuration initrd
 BOOT_CONFIG = "${BUILDSYS_ROOT_DIR}/bootconfig.data"
 
+# Determines the kubeconfig that should be used by testsys. If no kubeconfig was provided and the
+# default kubeconfig location does not exist, use the users default kubeconfig.
+CARGO_MAKE_TESTSYS_KUBECONFIG_ARG = {script = [
+   '''
```
bcressey: nit: align these quotes and the following script to the left margin

Makefile.toml
```diff
@@ -184,6 +184,23 @@ BOOT_CONFIG_INPUT = "${BUILDSYS_ROOT_DIR}/bootconfig-input"
...
+      if [[ -n ${TESTSYS_KUBECONFIG_PATH} ]]; then
```
bcressey: style nits: quote all shell variables and avoid `[[` if `[` will do:
```suggestion
      if [ -n "${TESTSYS_KUBECONFIG_PATH}" ]; then
```

Makefile.toml
```diff
@@ -184,6 +184,23 @@ BOOT_CONFIG_INPUT = "${BUILDSYS_ROOT_DIR}/bootconfig-input"
...
+         # If the user provides a kubeconfig path it should be used.
+         echo "--kubeconfig ${TESTSYS_KUBECONFIG_PATH}"
+      elif [[ -f "${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}" ]]; then
```
bcressey:
```suggestion
      elif [ -f "${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}" ]; then
```

Makefile.toml
```diff
@@ -184,6 +184,23 @@ BOOT_CONFIG_INPUT = "${BUILDSYS_ROOT_DIR}/bootconfig-input"
 # Boot Configuration initrd
 BOOT_CONFIG = "${BUILDSYS_ROOT_DIR}/bootconfig.data"
 
+# Determines the kubeconfig that should be used by testsys. If no kubeconfig was provided and the
+# default kubeconfig location does not exist, use the users default kubeconfig.
+CARGO_MAKE_TESTSYS_KUBECONFIG_ARG = {script = [
+   '''
+      if [[ -n ${TESTSYS_KUBECONFIG_PATH} ]]; then
+         # If the user provides a kubeconfig path it should be used.
+         echo "--kubeconfig ${TESTSYS_KUBECONFIG_PATH}"
```
bcressey: Since we expect it to be overridden, `TESTSYS_KUBECONFIG` might be a better name for the variable in that it's shorter and not too ambiguous.

bcressey (approved):

ecpullen:
^ Force push to address style nits and change `TESTSYS_KUBECONFIG_PATH` to `TESTSYS_KUBECONFIG`.

webern (approved):
Are the `CARGO_MAKE_` prefixes necessary? They make the variables wordy but maybe don't add value? Maybe there is a convention already in place that you are following...

ecpullen:
> Are the `CARGO_MAKE_` prefixes necessary? They make the variables wordy but maybe don't add value? Maybe there is a convention already in place that you are following...

This was recommended as those env variables are determined from user input for cargo make to use. They probably shouldn't be overridden by a user either.

bcressey (approved):

arnaldo2792 (approved):

PR #2445, opened 2022-09-20
Support static addressing via `net.toml`

### LINKED ISSUE #2204: Support static addressing via `net.toml`

zmrow:
**Issue number:**
Closes #2204 

**Description of changes:**
*It's probably best to review this PR by commit to follow the chain of changes!*

Adds support for setting ipv4/6 static addresses and simple routes in `net.toml`.  Using the DNS API settings added in #2353 , a user can configure a complete and working simple static addressing setup.

In order to add this support, a new version of network config has been created that adds a few additional settings: `static4`, `static6`, and `route`.  These settings are deserialized in  `net_config/v2` into the structures defined in `net_config/static_addressing`.  They are then converted to the associated structures defined in `wicked/static_addressing`.  The `Wicked*` structures are then serialized to file.

An additional test macro has been defined in `net_config` containing unit tests for the static addressing validation and deserialization.  The e2e tests defined in `wicked` have also been updated to include various configurations.  The `wicked` e2e test now iterates through the network configuration versions, validating both of them.

The `wicked` helper `netdog install` has been updated to handle static addressses, reading from the correct lease file and ensuring that the current IP is always written.  In the case multiple IPs exist, the first IP in the `addresses` list is uses as the primary.  If ipv4 and ipv6 addresses are set, the first ipv4 addresses is used as the primary.

An example valid net config from my home setup:
```
version = 2

[eno1.static4]
addresses = ["192.168.86.200/24"]

[[eno1.route]]
to = "default"
via "192.168.86.1"
```
along with the corresponding DNS API settings (`user-data.toml`):
```
[settings.dns]
name-servers = ["192.168.86.1"]
```

**Testing done:**
* All new and old unit tests pass.
* Manually validate (using `aws-dev` VM) that _all_ of the config setups defined in `test_data/wicked/net_config.toml` are properly parsed and addresses are properly set.  I did this by writing the resulting XML config directly to `/etc/wicked/ifconfig/eth0` and restarting the network.  Since my network is simple, the kernel chokes on trying to reach some of the defined routes and therefore doesn't set them, but at least we know the configuration is properly parsed.
* Boot `aws-k8s-1.21` and ensure the simple network defined in the kernel cmdline continues to work.
* Boot `metal-dev` with a few different static IP configurations, DNS settings, and a default route that would work on my network.  Successfully reaches the internet and run containers.

**Terms of contribution:**


zmrow:
^ Add an additional comment regarding the link scope when a route's gateway (`via`) is omitted.

sources/api/netdog/src/wicked/dhcp.rs
```diff
@@ -0,0 +1,119 @@
+use crate::net_config::{Dhcp4ConfigV1, Dhcp4OptionsV1, Dhcp6ConfigV1, Dhcp6OptionsV1};
```
zmrow, line 1: Nothing in this file is new code, all of it was moved out of `wicked.rs`

sources/api/netdog/src/net_config/static_address.rs
```diff
@@ -0,0 +1,64 @@
...
+
+// Allows the user to pass the string "default" or a valid ip address prefix.  We can't use an
+// untagged enum for this (#[serde(untagged)]) because "default" directly maps to one of the
+// variants.  Serde will only allow the "untagged" attribute if neither variant directly matches.
+impl TryFrom<String> for RouteTo {
+    type Error = error::Error;
+
+    fn try_from(input: String) -> Result<Self> {
+        Ok(match input.as_str() {
+            "default" | "Default" => RouteTo::DefaultRoute,
```
bcressey: nit: maybe convert the string to lowercase prior to the match, if we want to allow all variations like `DEFAULT`

sources/api/netdog/src/net_config/v1.rs
```diff
@@ -5,7 +5,7 @@ use super::{error, Dhcp4ConfigV1, Dhcp6ConfigV1, Error, Interfaces, Result, Vali
 use crate::{
     interface_name::InterfaceName,
     net_config::{Dhcp4OptionsV1, Dhcp6OptionsV1},
-    wicked::{WickedControl, WickedDhcp4, WickedDhcp6, WickedInterface},
+    wicked::{WickedDhcp4, WickedDhcp6, WickedInterface},
```
bcressey, line 8: nit: typo in commit message, "constructure", plus a subject/verb agreement issue ("provide")

sources/api/netdog/src/net_config/v2.rs
```diff
@@ -0,0 +1,145 @@
...
+        let primary_count = self
+            .interfaces
+            .values()
+            .filter(|v| v.primary == Some(true))
+            .count();
+        ensure!(
+            primary_count <= 1,
+            error::InvalidNetConfigSnafu {
+                reason: "multiple primary interfaces defined, expected 1"
+            }
```
bcressey, line 131: Do we expect `1` or `0 or 1`?

sources/api/netdog/src/lease.rs
```diff
@@ -21,6 +21,9 @@ lazy_static! {
 #[derive(Debug, Deserialize)]
 #[allow(dead_code)]
 pub(crate) struct LeaseInfo {
+    // When multiple IP addresses exist for an interface, the second address's key will be
```
bcressey: nit: typo in commit, "fucntionality"

sources/api/netdog/src/lease.rs
```diff
@@ -21,6 +21,9 @@ lazy_static! {
...
+    // `IPADDR_1`, `IPADDR_2`, and so on.  Parsing the lease in this way means we will always pick
+    // up the first configured IP address.
```
bcressey: To check my understanding - we pick up the first configured IP because we ignore the `IPADDR_N` fields, since there's no "rename" deserialization specified?

sources/api/netdog/src/cli/install.rs
```diff
@@ -1,12 +1,12 @@
 use super::{error, InterfaceFamily, InterfaceType, Result};
 use crate::dns::DnsSettings;
-use crate::lease::{lease_path, LeaseInfo};
+use crate::lease::{dhcp_lease_path, static_lease_path, LeaseInfo};
```
bcressey, line 3: nit: commit says "we use similar log" where I think "logic" is meant

sources/api/netdog/src/cli/install.rs
```diff
@@ -50,36 +50,60 @@ pub(crate) fn run(args: InstallArgs) -> Result<()> {
     }
 
     match (&args.interface_type, &args.interface_family) {
-        (InterfaceType::Dhcp, InterfaceFamily::Ipv4 | InterfaceFamily::Ipv6) => {
+        (interface_type @ InterfaceType::Dhcp, InterfaceFamily::Ipv4 | InterfaceFamily::Ipv6) => {
```
bcressey: Is this equivalent? It would remove the need for the second match arm.
```
(interface_type @ (InterfaceType::Dhcp | InterfaceType::Static), InterfaceFamily::Ipv4 | InterfaceFamily::Ipv6)
```

PROVISIONING-METAL.md
```diff
@@ -92,9 +92,22 @@ Interface names are expected to be correct as per `udevd` naming, no interface n
...
 
+As of version `2` static addressing with simple routes is supported via the below settings.
+Please keep in mind that when using static addresses, DNS information must be supplied to the system via user data: [`settings.dns`](https://github.com/bottlerocket-os/bottlerocket#network-settings).
+* `static4` (map): IPv4 static address settings.
+  * `addresses` (list of quoted IPv4 address including prefix): The desired IPv4 IP addresses, including prefix i.e. `["192.168.14.2/24"]`.  The first IP in the list will be used as the primary IP which `kubelet` will use when joining the cluster.  If ipv4 and ipv6 static addresses exist, the first ipv4 address is used.
+* `static6` (map): IPv6 static address settings.
+  * `addresses` (list of quoted IPv6 address including prefix): The desired IPv6 IP addresses, including prefix i.e. `["2001:dead:beef::2/64"]`.  The first IP in the list will be used as the primary IP which `kubelet` will use when joining the cluster.  If ipv4 and ipv6 static addresses exist, the first ipv4 address is used.
+
```
bcressey, line 101: nit: for consistency, use "IPv4" and "IPv6" throughout, rather than also the lowercase forms

bcressey (approved):
The code changes look great to me. I'd like to try this out under QEMU just to see what it looks like on a running system, but otherwise I don't see any blockers. Nice work.

sources/api/netdog/src/lease.rs
zmrow: Yes, that's correct!  We would need a `rename` or an additional struct field such as `ipaddr_1`, etc.

sources/api/netdog/src/cli/install.rs
zmrow: In the "static" match arm we pull the static lease instead of the DHCP lease, otherwise the actions are the same.  

I suppose I could move the `InterfaceFamily` enum somewhere and use it more globally, and then use that enum as an argument to a "get lease" function

sources/api/netdog/src/net_config/v2.rs
zmrow: We expect 0 or 1.  In the case a user does not tell us which interface is primary, we document that we choose the first interface in the file.

sources/api/netdog/src/lease.rs
zmrow: I'll clarify the comment to make it a bit clearer

sources/api/netdog/src/cli/install.rs
zmrow: Yes, you're right - I'll do that!

zmrow:
^ Addresses all of @bcressey 's comments!  Thanks!

PROVISIONING-METAL.md
```diff
@@ -92,9 +92,22 @@ Interface names are expected to be correct as per `udevd` naming, no interface n
...
+* `static4` (map): IPv4 static address settings.
+  * `addresses` (list of quoted IPv4 address including prefix): The desired IPv4 IP addresses, including prefix i.e. `["192.168.14.2/24"]`.  The first IP in the list will be used as the primary IP which `kubelet` will use when joining the cluster.  If IPv4 and IPv6 static addresses exist, the first IPv4 address is used.
+* `static6` (map): IPv6 static address settings.
+  * `addresses` (list of quoted IPv6 address including prefix): The desired IPv6 IP addresses, including prefix i.e. `["2001:dead:beef::2/64"]`.  The first IP in the list will be used as the primary IP which `kubelet` will use when joining the cluster.  If IPv4 and IPv6 static addresses exist, the first IPv4 address is used.
+
+* `route` (map): Static route; multiple routes can be added. (cannot be used in conjuction with DHCP)
+  * `to` (`"default"` or IP address with prefix, required): Destination address.
+  * `from` (IP address): Source IP address.
+  * `via` (IP address): Gateway IP address.  If no gateway is provided, a scope of `link` is assumed.
+  * `route-metric` (integer): Relative route priority.
```
foersleo, line 106: nit: I do not know if we need to be that specific but this should probably be `unsigned integer`.

foersleo (approved):

PROVISIONING-METAL.md
zmrow: Hm.  I don't think we need to get that specific, but we'll definitely change this if we get reports of folks having issues with it!

Issue #2420, opened 2022-09-14
 Update go dependencies for 1.10.0


etungsten:
This would include dependencies for packages like host-ctr and hotdog. A previous example of this is https://github.com/bottlerocket-os/bottlerocket/pull/2299

PR #2451, opened 2022-09-23
Ecs gpu init go 1.19

jpmcb:
**Issue number:**

Related: https://github.com/bottlerocket-os/bottlerocket/issues/2420

**Description of changes:**

Upgrades the go build directive to [1.19 which is the latest version in the SDK](https://github.com/bottlerocket-os/bottlerocket-sdk/pull/82#discussion_r978830565). This can be upgraded out of band from what version of the SDK is being used in bottlerocket since older versions of go can still build modules with newer build directives (as long as there aren't breaking changes in go):

```
cd sources/ecs-gpu-init && go1.18.6 build cmd/
```

**Testing done:**

Built the ECS nvidia variant ami, used it with a p2.xlarge instance (1 GPU) and see that the GPU info was generated correctly:

```
bash-5.1# cat /var/lib/ecs/gpu/nvidia-gpu-info.json
{"DriverVersion":"470.82.01","GPUIDs":["<GPU-ID>"]}
```

**Terms of contribution:**


jpmcb:
Whoops! This had a commit from a different branch I was on: should be fixed up now with the force push

PR #2454, opened 2022-09-26
pubsys: switch back to simplelog

jpculp:
**Description of changes:**

Since `env_logger` can only log output all to stderr or to stdout, I switched back to `simplelog` to maintain the mixed output `pubsys` had before #2415. Using **CombinedLogger**, we can spin up two **SimpleLogger**s with separate levels and filters as needed.

Also sorted `Cargo.toml`'s dependencies as it was getting a little messy.

**Testing done:**

Published the same AMI several times to verify a successful and unsuccessful push.
Also tried publishing a bad endpoint, and the output matched what we had with `env_logger`.

**Terms of contribution:**


jpculp:
Removed `env_logger` from `Cargo.toml`.

zmrow (approved):
:genie:

bcressey (approved):

PR #2437, opened 2022-09-19
Add k8s 1.24 variants

arnaldo2792:
**Issue number:**

Closes #2374

**Description of changes:**

```
sources: add symlink for metal-k8s-1.23 variant
sources: add symlink for vmware-k8s-1.23 variant
sources: add symlink for aws-k8s-1.23 variant
sources,variants: add metal-k8s-1.24 variant
sources,variants: add vmware-k8s-1.24 variant
sources,variants: add aws-k8s-1.24 variant
packages: add kubernetes 1.24
```

As part of this PR, the k8s 1.23 variants directories in `sources/model/src/*` were replaced with a symlink to the new k8s-1.24 variant for each platform. 

Starting with k8s 1.24, all command-line parameters to manage CNI plugins with the kubelet [were removed](https://v1-24.docs.kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/), as part of the efforts to clean up any remaining work related to the `dockershim`.

**Testing done:**

- [X] Test aws variant: the nodes joined a 1.23/1.24 cluster , and pods were successfully deployed to them.

```bash
[ec2-user@admin]$ apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "e08be563",
    "pretty_name": "Bottlerocket OS 1.10.0 (aws-k8s-1.24)",
    "variant_id": "aws-k8s-1.24",
    "version_id": "1.10.0"
  }
}
[ec2-user@admin]$ uname -a
Linux <ip>.us-west-2.compute.internal 5.15.54 #1 SMP Mon Sep 19 18:51:14 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

```

- [x] Test vmware variant (Note: tested in a 1.23 cluster)
- [x] Test metal variant (Note: tested in a 1.23 cluster)
- [x] Build 1.23 variants 
- [x] Test ipv6 clusters 

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #2460, opened 2022-09-26
kubelet: Add setting to change log level

### LINKED ISSUE #1931: Make kubelet log verbosity configurable
Environment: EKS v1.20

**What I'd like:** 
A configuration option to adjust the log verbosity of the kubelet in all applicable kubernetes versions [(via the option `--v=<log level>`)](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).

We have noticed differences in log verbosity in EKS clusters that is differentiated only by the use of Bottlerocket, which currently has no method to adjust kubelet log verbosity.

We have considered attempting to modify the `kubelet-exec-start-conf` file during node startup however this does not feel like the best course of action as kubelet configuration options should be supported by default.

stmcginnis:
**Issue number:**

Closes #1931 

**Description of changes:**

This adds a new `kubernetes.log-level` setting to allow configuration of the kubelet logging verbosity. With this, an end user can turn up verbosity if they are trying to troubleshoot something, or turn down verbosity if they don't need as many log messages.

**Testing done:**

Created EKS cluster using development build.
Checked `kubelet` command line arguments from `systemctl status kubelet` and verified no `-v` setting.
Ran `apiclient set -j '{"kubernetes": {"log-level": 0}}'`
Ran `apiclient get settings.kubernetes` and verified correct value for `log-level`.
Checked `systemctl status kubelet` and verified `-v 0` was one of the arguments used to launch the process.
Ran `journalctl -u kubelet -f` and watched output to see if any additional log messages were being emitted.
Ran `apiclient set -j '{"kubernetes": {"log-level": 9}}'` and checked setting with `apiclient get`.
Checked `systemctl status kubelet` and verified `-v 9` was one of the arguments used to launch the process.
Ran `journalctl -u kubelet -f` and watched additional recurring log messages showing all activity.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,6 +447,8 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.image-gc-high-threshold-percent`: The percent of disk usage after which image garbage collection is always run.
 * `settings.kubernetes.image-gc-low-threshold-percent`: The percent of disk usage before which image garbage collection is never run.
 * `settings.kubernetes.provider-id`: This sets the unique ID of the instance that an external provider (i.e. cloudprovider) can use to identify a specific node.
+* `settings.kubernetes.log-level`: Adjust the logging verbosity of the `kubelet` process.
+  The default log level is 2, with higher numbers enabling more verbose logging.
 
 You can also optionally specify static pods for your node with the following settings.
 Static pods can be particularly useful when running in standalone mode.
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -28,6 +28,7 @@ members = [
     "api/migration/migrations/v1.10.0/dns-settings",
     "api/migration/migrations/v1.10.0/dns-settings-metadata",
     "api/migration/migrations/v1.10.0/reboot-to-reconcile-setting",
+    "api/migration/migrations/v1.10.0/kubelet-log-level",
 
     "bottlerocket-release",
 
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -216,6 +216,7 @@ struct KubernetesSettings {
     image_gc_high_threshold_percent: ImageGCHighThresholdPercent,
     image_gc_low_threshold_percent: ImageGCLowThresholdPercent,
     provider_id: Url,
+    kubelet_log_level: u8,
 
     // Settings where we generate a value based on the runtime environment.  The user can specify a
     // value to override the generated one, but typically would not.
--- a/Release.toml
+++ b/Release.toml
@@ -146,4 +146,5 @@ version = "1.10.0"
     "migrate_v1.10.0_dns-settings.lz4",
     "migrate_v1.10.0_dns-settings-metadata.lz4",
     "migrate_v1.10.0_reboot-to-reconcile-setting.lz4",
+    "migrate_v1.10.0_kubelet-log-level.lz4",
 ]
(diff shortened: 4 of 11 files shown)


README.md
```diff
@@ -447,6 +447,8 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.image-gc-high-threshold-percent`: The percent of disk usage after which image garbage collection is always run.
 * `settings.kubernetes.image-gc-low-threshold-percent`: The percent of disk usage before which image garbage collection is never run.
 * `settings.kubernetes.provider-id`: This sets the unique ID of the instance that an external provider (i.e. cloudprovider) can use to identify a specific node.
+* `settings.kubernetes.kubelet-log-level`: Adjust the logging verbosity of the `kubelet` process.
+  The default log level is 2, with higher numbers enabling more verbose logging.
```
bcressey: Hindsight being what it is, `settings.kubernetes` would have been better named `settings.kubelet`.

The naming dates from very early in the project when it wasn't clear whether we'd want to ship control plane components some day; now it seems pretty certain that we won't.

With that in mind, I would prefer to treat `settings.kubernetes` and `settings.kubelet` as aliases - which some day they might be - and just call this `settings.kubernetes.log-level`.

sources/models/src/lib.rs
```diff
@@ -216,6 +216,7 @@ struct KubernetesSettings {
     image_gc_high_threshold_percent: ImageGCHighThresholdPercent,
     image_gc_low_threshold_percent: ImageGCLowThresholdPercent,
     provider_id: Url,
+    kubelet_log_level: i32,
```
bcressey: Is `i32` the right type? Does `kubelet` handle negative or very large values, or does it expect it to be capped in some way?

bcressey (commented):
Nice!

README.md
stmcginnis: I was originally thinking we may want something like `settings.kubernetes.log-level.kubelet`, `settings.kubernetes.log-level.foo`, etc., thinking there might be other aspects of kubernetes as a whole that we might want to expose separate settings for. But I don't think that fits very cleanly either.

I will update to just be `log-level` and we can decide in the future if that is applicable to anything other than kubelet. 👍

sources/models/src/lib.rs
stmcginnis: Perhaps this was laziness, but I used `i32` just based on other settings we had. In reality, the number should only be between 0 and 9. I wonder if this would be worth adding a custom type to help enforce? I don't believe it would cause any problems if a user set it to something like -12, but it would help make it clear what the actual valid values should be.

jpmcb (approved):

sources/Cargo.toml
```diff
@@ -28,6 +28,7 @@ members = [
     "api/migration/migrations/v1.10.0/dns-settings",
     "api/migration/migrations/v1.10.0/dns-settings-metadata",
     "api/migration/migrations/v1.10.0/reboot-to-reconcile-setting",
+    "api/migration/migrations/v1.10.0/kubelet-log-level",
```
markusboehme, line 31: The migration also needs to be added to `Release.toml`, so that `migrator` will know when to run the migration.

markusboehme (commented):
There is a dependency between this PR and #2437. Whichever merges last will have to update the kubelet config template for k8s 1.24.

sources/Cargo.toml
stmcginnis: Thanks, I missed that file!

markusboehme (approved):
:writing_hand:

bcressey (approved):

PR #2461, opened 2022-09-27
CI: extract checks from build actions

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

We currently have one workflow defined for our GitHub Actions. This workflow uses a matrix to run tests for every variant defined. In addition to performing the actual build for these variants, this also runs unit tests, checks formatting, and checks lint issues. Since these checks do not change based on the variant, this results in a lot of duplicated testing and adds a significant delay to how long it takes to get results.

This change extracts these common checks into a separate workflow so they are only run once. This also has the benefit of being able to run in parallel with some of the variant builds and may allow for faster feedback if there is something like a formatting issue.

**Testing done:**

Will be validated via GitHub Action runs.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -115,9 +115,6 @@ jobs:
             { path = "NVIDIA", license-url = "https://www.nvidia.com/en-us/drivers/nvidia-license/" }
           ]
           EOF
-      - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} unit-tests
-      - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} check-fmt
-      - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} check-lints
       - run: |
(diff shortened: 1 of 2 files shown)


ecpullen (approved):
LGTM

stmcginnis:
Going to mark as draft while I do some more investigation. It was pointed out our use of `build.rs` may cause different unit test behavior based on variant. In this case, the new job is using our default variant.

I still think there's value in running this separately to get faster feedback, but need to look closer to make sure we don't miss anything.

webern:
Actually... there are some differences in the unit tests per variant. This is because we have conditional compilation taking place. Take a look at ECS settings applier for an example of unit tests that would not run for other variants.

stmcginnis:
Will have to think about whether there is a good way to get fast feedback. Since the current checks are run first in the existing job definition, they actually do provide faster feedback for formatting issues than the time it takes to run the complete job, so now that I understand this a little better I don't see this change adding as much value as I thought.

Will close and push up a new PR if there ends up being anything useful to do separately from the main build checks.

PR #2467, opened 2022-09-30
gitignore: add my/ scratch pad directory to ignore list

markusboehme:
**Issue number:** n/a

**Description of changes:** Checked out repositories tend to accumulate files related to personal workflows. Add a `my/` directory in the repo root to the git ignore list so there's a dedicated area for those files while keeping "git status" output clean.

Examples of files I keep there `my/net.toml`, `my/kernel-update-1.10-diff`, `my/k8s-migration-user-data.toml`, ... The name may be a bit on the cute side, but arguably has some benefits:

1. It's short to type
2. It's easy to remember
3. It's leading to copy-pastable commands that show the reader where to substitute their own files

**Testing done:** Moved all my valuables to `my/` and rejoiced in clean `git status` output.

**Terms of contribution:**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 /Licenses.toml
 /licenses
 *.run
+/my/


stmcginnis (approved):
Interesting idea. This works for me.

.gitignore
```diff
@@ -13,3 +13,4 @@
 /Licenses.toml
 /licenses
 *.run
+/my/
```
bcressey, line 16: Do you also want this ignored by `.dockerignore`? That controls what is and isn't copied into package and image builds as part of the build context.

jpmcb (commented):
> Checked out repositories tend to accumulate files related to personal workflows. Add a my/ directory in the repo root to the git ignore list so there's a dedicated area for those files while keeping "git status" output clean.

I'm _generally_ not a fan of adding non-project specific things to a `.gitignore` since it won't ever scale to the vastly different ways a community might work on a project.

For example, I don't think we should start adding ignores for workspace specific editor settings like `.vscode`, `init.vim`, `workspace.xml`, etc. etc.

Instead, I usually advocate people populate a global `.gitignore` under `~./.gitignore` and then use it via `git config --global core.excludesfile ~/.gitignore`.

stmcginnis:
I've typically done what @jpmcb suggests, even going so far as adding instructions to the top of `.gitignore` in other projects directing people how to set up those system-wide settings.

If having a `./my` directory is a pattern we want to follow as a project, I don't think there is much harm in it. But we may want to consider keeping things minimal and directing folks to set up their environments individually for what makes sense for them.

.gitignore
markusboehme: I was unaware of this (TIL--thanks!). Yes, it should be excluded there as well.

markusboehme:
I see your point, John, and agree it'll be better for contributors to set up their own personal ignore rules for files and habits specific to their workflows. I was unaware of the ability to specify ignore files outside the repo, so I will do just that. I also like Sean's suggestion of including instructions for doing so in the `.gitignore`.

Thanks for the discussion, this was helpful! I'll close this PR and send another one for incorporating Sean's suggestion.  :+1:

PR #2473, opened 2022-10-03
specify mode for tmpfs mounts

bcressey:
**Issue number:**

N/A

**Description of changes:**
Specify the mode for all tmpfs mounts. In practice, the mode didn't really matter because the SELinux label would enforce the expected permissions, but the "wide open" default mode for tmpfs mounts could trigger an alert from scanning tools.

**Testing done:**
Verified that `aws-ecs-1` and `aws-k8s-1.22` quick tests passed:
```
 NAME                            TYPE       STATE       PASSED   SKIPPED   FAILED
 aarch64-aws-ecs-1               Resource   completed
 aarch64-aws-ecs-1-instances     Resource   completed
 aarch64-aws-ecs-1-test          Test       passed      1        0         0
 aarch64-aws-k8s-122             Resource   completed
 aarch64-aws-k8s-122-instances   Resource   completed
 aarch64-aws-k8s-122-test        Test       passed      1        6441	   0
```

Confirmed that each mount had the expected permissions.

**Terms of contribution:**

### DIFF
--- a/packages/release/etc-cni.mount
+++ b/packages/release/etc-cni.mount
@@ -8,7 +8,7 @@ Before=local-fs.target umount.target
 What=tmpfs
 Where=/etc/cni
 Type=tmpfs
-Options=nosuid,nodev,noexec,noatime
+Options=nosuid,nodev,noexec,noatime,mode=0755
 
 [Install]
 WantedBy=preconfigured.target
--- a/packages/containerd/etc-containerd.mount
+++ b/packages/containerd/etc-containerd.mount
@@ -10,7 +10,7 @@ Wants=selinux-policy-files.service
 What=tmpfs
 Where=/etc/containerd
 Type=tmpfs
-Options=nosuid,nodev,noexec,noatime,context=system_u:object_r:secret_t:s0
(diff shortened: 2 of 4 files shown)


packages/release/etc-cni.mount
```diff
@@ -8,7 +8,7 @@ Before=local-fs.target umount.target
 What=tmpfs
 Where=/etc/cni
 Type=tmpfs
-Options=nosuid,nodev,noexec,noatime
+Options=nosuid,nodev,noexec,noatime,mode=0755
```
stmcginnis, line 11: Do we need a comment anywhere explaining why this one needs `0755` while everything else gets set to `0750`?

stmcginnis (approved):
Just a question mostly out of curiosity, but looks good to me. Matches what I've confirmed for settings with the pending credential provider work.

arnaldo2792 (approved):

packages/release/etc-cni.mount
bcressey: `/etc/cni` is used by CNI plugins, where we don't necessarily know for sure what user containers in the pod will run with. The other directories should only be accessed by host processes.

PR #2482, opened 2022-10-05
Clean up settings information in the README

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This cleans up our README doc with a few main changes:

- Settings entries are sorted within their sections to help make it easier to find specific items
- Example code blocks are tagged with their type to allow syntax highlighting
- Minor format consistencies.

**Testing done:**

Visual inspection.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ Once the instance is started, you can start a session:
 If you prefer a command-line tool, you can start a session with a recent [AWS CLI](https://aws.amazon.com/cli/) and the [session-manager-plugin](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html).
 Then you'd be able to start a session using only your instance ID, like this:
 
-```
+```txt
 aws ssm start-session --target INSTANCE_ID
 ```
 
@@ -172,7 +172,7 @@ With the [default control container](https://github.com/bottlerocket-os/bottlero
 To do even more, read the next section about the [admin container](#admin-container).
 You can access the admin container from the control container like this:
 
-```
+```sh
 enter-admin-container
 ```
 
@@ -185,32 +185,32 @@ Outside of AWS, you can [pass in your own SSH keys](https://github.com/bottleroc
 
 To enable the container, you can change the setting in user data when starting Bottlerocket, for example EC2 instance user data:
 
-```
+```toml
 [settings.host-containers.admin]
 enabled = true
 ```
 
 If Bottlerocket is already running, you can enable the admin container from the default [control container](#control-container) like this:
 
-```
+```sh
 enable-admin-container
 ```
 
 Or you can start an interactive session immediately like this:
 
-```
+```sh
 enter-admin-container
 ```
 
 If you're using a custom control container, or want to make the API calls directly, you can enable the admin container like this instead:
 
-```
+```txt
 apiclient set host-containers.admin.enabled=true
 ```
 
 Once you've enabled the admin container, you can either access it through SSH or execute commands from the control container like this:
 
-```
+```txt
 apiclient exec admin bash
 ```
 
@@ -245,7 +245,7 @@ apiclient knows how to handle those update APIs for you, and you can run it from
 
 To see what updates are available:
 
-```
+```txt
 apiclient update check
 ```
 
@@ -254,20 +254,20 @@ The `available_updates` field will show the full list of available versions, inc
 
 To apply the latest update:
 
-```
+```txt
 apiclient update apply
 ```
 
 The next time you reboot, you'll start up in the new version, and system configuration will be automatically [migrated](sources/api/migration/).
 To reboot right away:
 
-```
+```txt
 apiclient reboot
 ```
 
 If you're confident about updating, the `apiclient update apply` command has `--check` and `--reboot` flags to combine the above actions, so you can accomplish all of the above steps like this:
 
-```
+```txt
 apiclient update apply --check --reboot
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -545,7 +555,7 @@ When pulling an image from a registry, the container runtime will try the endpoi
   endpoint = [ "https://<my-docker-hub-mirror-host>", "https://<my-docker-hub-mirror-host-2>"]
   ```
 
-  If you use a Bottlerocket variant that uses Docker as the container runtime, like `aws-ecs-1`, you should be aware that Docker only supports pull-through caches for images from Docker Hub (docker.io).  Mirrors for other registries are ignored in this case.
+  If you use a Bottlerocket variant that uses Docker as the container runtime, like `aws-ecs-1`, you should be aware that Docker only supports pull-through caches for images from Docker Hub (docker.io). Mirrors for other registries are ignored in this case.
```
zmrow, line 562: I think there might be some extra spaces at the beginning of this line.

README.md
```diff
@@ -554,7 +564,8 @@ The following setting is optional and allows you to configure image registry cre
 * `settings.container-registry.credentials`: An array of container images registry credential settings. Each element specifies the registry and the credential information for said registry.
 The credential fields map to [containerd's registry credential fields](https://github.com/containerd/containerd/blob/v1.6.0/docs/cri/registry.md#configure-registry-credentials), which in turn map to the fields in `.docker/config.json`.
 It is recommended to programmatically set these settings via `apiclient` through the Bottlerocket control container and/or custom host-containers.
-  * An example `apiclient` call to set registry credentials for `gcr.io` and `docker.io` looks like this:
+
+  An example `apiclient` call to set registry credentials for `gcr.io` and `docker.io` looks like this:
```
zmrow, line 572: Same here

stmcginnis: This is definitely intentional. We want the example to be text under the bullet item and not outside of it so that it stays connected to the setting that it is giving an example for.

README.md
stmcginnis: I believe this was intended in the original so that it is text underneath the bullet item and not a separate paragraph after the bullet list.

README.md
zmrow: Ah I see that now - disregard!

README.md
zmrow: ^ You're right! Carry on. :)

zmrow (approved):

README.md
```diff
@@ -181,32 +181,32 @@ Outside of AWS, you can [pass in your own SSH keys](https://github.com/bottleroc
...
-```
+```sh
 enter-admin-container
 ```
 
 If you're using a custom control container, or want to make the API calls directly, you can enable the admin container like this instead:
 
-```
+```sh
 apiclient set host-containers.admin.enabled=true
```
jpmcb, line 208: I'm personally not a fan of adding syntax highlighting for these kind of short one offs (mostly because GitHub resolves individual _words_ with highlights, not whole commands). For example, here `set` gets highlighted since it is a keyword in bash:

Screen Shot 2022-10-07 at 10 23 16 AM

But other commands with this format don't get this treatment since they don't contain keywords:

Screen Shot 2022-10-07 at 10 27 06 AM

So, it's a bit awkward and creates it's own inconsistencies. Seems the two problem words are `set` and `exec`.

README.md
```diff
@@ -181,32 +181,32 @@ Outside of AWS, you can [pass in your own SSH keys](https://github.com/bottleroc
...
 ```
 
 Once you've enabled the admin container, you can either access it through SSH or execute commands from the control container like this:
 
-```
+```sh
 apiclient exec admin bash
```
jpmcb, line 214: Same here. `exec` is highlighted since it's a keyword:

Screen Shot 2022-10-07 at 10 23 23 AM

README.md
```diff
@@ -293,20 +293,20 @@ Here we'll describe the settings you can configure on your Bottlerocket instance
...
+```sh
 apiclient get settings
 ```
 
 This will return all of the current settings in JSON format.
 For example, here's an abbreviated response:
 
-```
+```json
 {"motd":"...", {"kubernetes": ...}}
```
jpmcb: This one is also awkward: the three `...` aren't valid JSON so they get highlighted as an error. Maybe it just needs to be wrapped in a string like the `motd` key is?

Screen Shot 2022-10-07 at 10 28 13 AM

README.md
stmcginnis: I'll admit, I hesitated on that. In a lot of cases, the `sh` type really doesn't add a lot of value. I'll change those to `txt` since we know we don't want or need syntax highlighting.

README.md
stmcginnis: I actually think this may be OK. It kind of highlights "we're giving you a JSON example here, but it's obvious this is just a snippet and not a real valid JSON structure". But I could go either way if there's a strong opinion against that.

jpmcb: Sounds good! I don't have crazy strong opinions on this, so whatever you think looks best!

etungsten (approved):

README.md
webern: That seems fine to me as I'm used to seeing keywords highlighted in shell snippets.

README.md
webern: I agree that highlighted errors are not ideal.  `txt` or put the quotes if `json`, either way.

webern (approved):

stmcginnis:
Updated to address comments and rebase on latest `develop`. Will wait and give folks a chance to review again to make sure everyone is cool with the updates. :)

jpmcb (approved):

etungsten (approved):

PR #2493, opened 2022-10-14
Add fallback region to early-boot-config

jpculp:
**Description of changes:**

Replace `ImdsMissingRegion` panic with a fallback value in certain development scenarios without IMDS.

Also addresses some clippy warnings for general housekeeping.

**Testing done:**

- [x] Build `aws-k8s` variant and launch in an environment without IMDS.
- [x] Build `aws-k8s` variant and launch in EC2.

**Terms of contribution:**

### DIFF
--- a/sources/api/early-boot-config/src/provider/aws.rs
+++ b/sources/api/early-boot-config/src/provider/aws.rs
@@ -16,6 +16,7 @@ pub(crate) struct AwsDataProvider;
 
 impl AwsDataProvider {
     const IDENTITY_DOCUMENT_FILE: &'static str = "/etc/early-boot-config/identity-document";
+    const FALLBACK_REGION: &str = "us-east-1";
 
     /// Fetches user data, which is expected to be in TOML form and contain a `[settings]` section,
     /// returning a SettingsJson representing the inside of that section.
@@ -65,7 +66,7 @@ impl AwsDataProvider {
                 .fetch_region()
                 .await
                 .context(error::ImdsRequestSnafu)?
-                .context(error::ImdsMissingRegionSnafu)?
+                .unwrap_or_else(|| Self::FALLBACK_REGION.to_owned())
         };
         trace!(
             "Retrieved region from instance identity document: {}",
@@ -135,9 +136,6 @@ mod error {
 
         #[snafu(display("Unable to read input file '{}': {}", path.display(), source))]
         InputFileRead { path: PathBuf, source: io::Error },
-
-        #[snafu(display("IMDS request failed: missing region"))]
-        ImdsMissingRegion {},
 
         #[snafu(display("IMDS request failed: {}", source))]
         ImdsRequest { source: imdsclient::Error },
--- a/sources/api/early-boot-config/src/provider/aws.rs
+++ b/sources/api/early-boot-config/src/provider/aws.rs
@@ -15,7 +15,7 @@ use crate::provider::local_file::{local_file_user_data, USER_DATA_FILE};
 pub(crate) struct AwsDataProvider;
 
 impl AwsDataProvider {
-    const IDENTITY_DOCUMENT_FILE: &'static str = "/etc/early-boot-config/identity-document";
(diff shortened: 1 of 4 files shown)


webern (approved):

bcressey (approved):

markusboehme (approved):

sources/api/early-boot-config/src/provider/local_file.rs
```diff
@@ -5,7 +5,7 @@ use crate::compression::expand_file_maybe;
 use snafu::ResultExt;
 use std::path::Path;
 
-pub(crate) const USER_DATA_FILE: &'static str = "/var/lib/bottlerocket/user-data.toml";
+pub(crate) const USER_DATA_FILE: &str = "/var/lib/bottlerocket/user-data.toml";
```
zmrow, line 8: Out of curiosity, was removing `static` something clippy asked for?  If so, we should probably do the same to `IDENTITY_DOCUMENT_FILE`.

jpculp: clippy flagged `USER_DATA_FILE` as a redundant static lifetime, but not `IDENTITY_DOCUMENT_FILE`. That said, I didn't get any errors when I removed it from `IDENTITY_DOCUMENT_FILE` just now, so we can go ahead and do that for consistency.

jpculp:
Removed `'static` from `IDENTITY_DOCUMENT_FILE` as well.

zmrow (approved):
:feather:

sources/api/early-boot-config/src/provider/aws.rs
```diff
@@ -15,7 +15,8 @@ use crate::provider::local_file::{local_file_user_data, USER_DATA_FILE};
 pub(crate) struct AwsDataProvider;
 
 impl AwsDataProvider {
-    const IDENTITY_DOCUMENT_FILE: &'static str = "/etc/early-boot-config/identity-document";
+    const IDENTITY_DOCUMENT_FILE: &str = "/etc/early-boot-config/identity-document";
```
stmcginnis, line 18: I'm getting local errors now due to this:

```
  --> api/early-boot-config/src/provider/aws.rs:18:35
   |
18 |     const IDENTITY_DOCUMENT_FILE: &str = "/etc/early-boot-config/identity-document";
   |                                   ^ expected named lifetime parameter
```

Is this a mismatch of rust toolchain versions where it's an error for me without, but an error for you with?

jpculp: Actually the only redundant lifetime was on `USER_DATA_FILE`, but I was asked to remove it from `IDENTITY_DOCUMENT_FILE` for consistency. Which rust toolchain are you currently using? Do you get the same error for `USER_DATA_FILE` or just `IDENTITY_DOCUMENT_FILE`?

stmcginnis: I did get it for both constants.

```bash
$ cargo --version
cargo 1.63.0
```

stmcginnis: I saw this running `cargo make check` on a local change that I had rebased on to the latest from `develop`. Although I got the failure locally, I did push the change up to my PR, and the GH Actions that ran against the change all passed. Those show they were running `1.64.0`.

I will try updating to 1.64.0 as well and see if I get better results. Will report back here if that's the case so we can point anyone else that runs in to this here.

jpculp: It looks like rust ≤ **1.63** wants `'static` on `IDENTITY_DOCUMENT_FILE` and  the newly added `FALLBACK_REGION`, with `USER_DATA_FILE` remaining `&str`. Since the Bottlerocket SDK uses **1.64** this isn't a super big problem, but I'll post a fix so we can increase our toolchain compatibility.

jpculp: Opened a PR (#2507) for the compatibility fix .

PR #2494, opened 2022-10-14
settings: Add container-runtime settings

### LINKED ISSUE #2292: allow setting max_container_log_line_size for containerd
**What I'd like:**

It will be really nice, when we will be able to set max_container_log_line_size value for containerd.

### LINKED ISSUE #2466: Allow rootless containers to bind privileged ports and ping
**What I'd like:**
Add config options to allow containers to bind privileged ports (<1024) and open icmp echo sockets without being root.  (This the defaults with dockerd)
It can be enabled within containerd config (/etc/containerd/config.toml).
For kubernetes:
```
[plugins."io.containerd.grpc.v1.cri"]
enable_unprivileged_ports = true
enable_unprivileged_icmp = true
```

This can be worked around with kubernetes (>1.22) with a SecurityContext on pods:
```
securityContext:
  sysctls:
    - name: net.ipv4.ip_unprivileged_port_start
      value: "0"
    - name: net.ipv4.ping_group_range
      value: "0 2147483647"
```

It can also be configured within a bootstrap container but I'm not sure how to restart containerd from there.

stmcginnis:
**Issue number:**

Closes #2292
Closes: #2466

**Description of changes:**

This adds `settings.container-runtime` settings. The affect how containerd is configured.

`max-container-log-line-size` controls how long a log line can be from a container before containerd breaks it into multiple separate lines.

`max-concurrent-downloads` controls how many concurrent downloads will be done in parallel to download an image.

`settings.container-runtime.enable-unprivileged-icmp`: Allow unprivileged containers to open ICMP echo sockets.

`settings.container-runtime.enable-unprivileged-ports`: Allow unprivileged containers to bind to ports < 1024.

**Testing done:**

Created EKS cluster with custom build, checked setting and retrieving settings via `apiclient get settings.container-runtime`. Verified containerd service config would get updated and service restarted.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -580,6 +580,27 @@ It is recommended to programmatically set these settings via `apiclient` through
 
 In addition to the container runtime daemons, these credential settings will also apply to [host-container](#host-containers-settings) and [bootstrap-container](#bootstrap-containers-settings) image pulls as well.
 
+#### Container runtime settings
+
+Some behavior of the container runtime (currently `containerd`) can be modified with the following settings:
+
+* `settings.container-runtime.enable-unprivileged-icmp`: Allow unprivileged containers to open ICMP echo sockets.
+* `settings.container-runtime.enable-unprivileged-ports`: Allow unprivileged containers to bind to ports < 1024.
+* `settings.container-runtime.max-concurrent-downloads`: Restricts the number of concurrent layer downloads for each image.
+* `settings.container-runtime.max-container-log-line-size`: Controls how long container log messages can be.
+  If the log output is longer than the limit, the log message will be broken into multiple lines.
+
+Example container runtime settings:
+
+```toml
+[settings.container-runtime]
+# Set log line length to unlimited
+max-container-log-line-size = -1
+max-concurrent-downloads = 4
+enable-unprivileged-icmp = true
+enable-unprivileged-ports = true
+```
+
 #### Updates settings
 
 * `settings.updates.metadata-base-url`: The common portion of all URIs used to download update metadata.
--- a/packages/containerd/containerd-config-toml_k8s
+++ b/packages/containerd/containerd-config-toml_k8s
@@ -16,6 +16,18 @@ address = "/run/dockershim.sock"
 enable_selinux = true
 # Pause container image is specified here, shares the same image as kubelet's pod-infra-container-image
 sandbox_image = "{{settings.kubernetes.pod-infra-container-image}}"
+{{#if settings.container-runtime.max-container-log-line-size}}
+max_container_log_line_size = {{settings.container-runtime.max-container-log-line-size}}
+{{/if}}
+{{#if settings.container-runtime.max-concurrent-downloads}}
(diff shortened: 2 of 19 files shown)


webern (approved):
LGTM pending testing (and passing builds). The only thought I had is whether we want to constrain the values to those that containerd accepts (e.g. -1 - 16384 for max_container_log_line_size, questionable whether 0 is allowable), but that seems like more trouble than its worth right now.

stmcginnis:
> The only thought I had is whether we want to constrain the values to those that containerd accepts

Yeah, good call. This has come up a couple times now. From what I understand, we don't have a good mechanism to handle extra validation for numeric values yet, so I do think we should get to that as a (hopefully soon) follow on.

README.md
```diff
@@ -511,6 +511,27 @@ These settings can be changed at any time.
...
+* `settings.container-runtime.enable-unprivileged-ports`: Allow unprivileged containers to bind to ports < 1024.
+* `settings.container-runtime.max-concurrent-downloads`: Restricts the number of concurrent downloads for each image.
+* `settings.container-runtime.max-container-log-line-size`: Controls how long container log messages can be.
+  If the log output is longer than the limit, the log message will be broken into multiple lines.
+
+Example container runtime settings:
+
+```toml
+[settings.container-runtime]
+# Set log line length to unlimited
```
bcressey: nit: I would tend to put this a little further down, near `oci-hooks` and `container-registry` settings.

README.md
```diff
@@ -511,6 +511,27 @@ These settings can be changed at any time.
 * `settings.ecs.image-pull-behavior`: The behavior used to customize the [pull image process](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html#ecs-agent-availparam) for your container instances.
   Supported values are `default`, `always`, `once`, `prefer-cached`, and the default is `default`.
 
+#### Container runtime settings
+
+Some behavior of the container runtime (currently `containerd`) can be modified with the following settings:
+
+* `settings.container-runtime.enable-unprivileged-icmp`: Allow unprivileged containers to open ICMP echo sockets.
+* `settings.container-runtime.enable-unprivileged-ports`: Allow unprivileged containers to bind to ports < 1024.
+* `settings.container-runtime.max-concurrent-downloads`: Restricts the number of concurrent downloads for each image.
```
bcressey: For my own knowledge - concurrent downloads means "how many layers for a given image are downloaded at once", rather than "how many images are downloaded at once"?

I assume this does not actually allow `containerd` to download the same image multiple times, so it might be worth clarifying that it's about layers and manifests somehow (if that's true).

bcressey (approved):

README.md
stmcginnis: Moved.

README.md
stmcginnis: Tweaked.

bcressey (approved):

Issue #2501, opened 2022-10-17
add support for Secure Boot


bcressey:
**What I'd like:**
On platforms where Secure Boot is supported, I'd like strong cryptographic verification of all components from the bootloader to the running kernel.

**Any alternatives you've considered:**
None.

bcressey:
This has been in-flight for a while [on a branch in my personal fork](https://github.com/bcressey/bottlerocket/commits/secureboot-shenanigans) but is now nearing the point where the full picture is visible and pieces can start to be merged. There's also some work related to bottlerocket-os/bottlerocket#2486 since I'd like to support the resigning workflow described there as a prerequisite.

Tasks related to resigning images:
- [x] Refactor image and symlink handling, to simplify bundling up artifacts from the image build step.
- [x] Move OVA creation into the image build step. 
- [ ] ~~Implement `import-images` and `export-images` tasks to store and retrieve artifacts from the repo.~~
- [ ] ~~Implement `resign-images` task.~~

Tasks related to Secure Boot:
- [x] Update GRUB to fix Boothole 3 vulnerabilities.
- [x] Package shim for initial bootloader verification.
- [x] Add SBAT info for GRUB.
- [x] Add scripts to create necessary keys and certificates for Secure Boot signing.
- [x] Add modules to GRUB to enable signed config files.
- [x] Add feature flag to enable Secure Boot for variants.
- [x] Sign bootloader, kernel, GRUB config when Secure Boot is enabled.
- [x] Publish AMIs and OVAs with required Secure Boot metadata.

bcressey:
I dropped the task related to a separate firmware build step, since @markusboehme has a patch that allows embedding the GPG public key into a section in the grub binary, which allows for "just in time" key modification without the need to rebuild `grub` from source.

markusboehme:
> I dropped the task related to a separate firmware build step, since @markusboehme has a patch that allows embedding the GPG public key into a section in the grub binary, which allows for "just in time" key modification without the need to rebuild grub from source.

For the record, these patches are available at https://github.com/markusboehme/bottlerocket/tree/feature/grub-pubkey-section

bcressey:
@yeazelm is working on the image resigning tasks in bottlerocket-os/bottlerocket#2486 and bottlerocket-os/twoliter#176.

bcressey:
Resolving now that bottlerocket-os/bottlerocket#3097 is merged! 🎉

Issue #2486, opened 2022-10-12
Tooling to Alter a Built Bottlerocket Image and Re-sign it with DM Verity


webern:
Consider the case when a developer is working on a change to Bottlerocket that requires a migration. In order to test the migration, you need to build TWO versions of Bottlerocket. One to migrate FROM and the one that you are working on (the migration destination).

You cannot use a production/published build of Bottlerocket as the FROM version because it is built with the production-keyed `root.json` file. You have to rebuild all of Bottlerocket just to produce an image with your own `root.json`.

Possible Solution: Create a tool that can "open" or "mount" the production Bottlerocket image, make certain changes to it (for example replacing the `root.json`) then remake the DM Verity signature.

This almost certainly has other uses beyond just replacing the `root.json` file. Supporting secure boot might be another case where this would be useful.

bcressey:
I have an ad hoc, informally-specified, bug-ridden, slow implementation of half of this already done.

bcressey:
Implemented in https://github.com/bottlerocket-os/twoliter/pull/231.

PR #2505, opened 2022-10-20
prepare foundations for Secure Boot and image resigning

bcressey:
**Issue number:**

#2486, #2501

**Description of changes:**
The commits here divide into three rough groups, though they're all related to the larger goals of being able to resign images and to add support for Secure Boot. 

**First up** is adding support to `buildsys` to track symlinks and subdirectories, and then moving the bulk of symlink creation into the `build-variant` task. Images and symlinks are now placed into a versioned directory, rather than all together in a single directory per variant/arch combination. The ultimate goal here is to be able to cleanly archive and later unpack the full contents of a specific build, without worrying about excluding artifacts from other builds or recreating the (surprisingly fiddly and prolific) symlinks.

There's still "latest" but it's now a symlink to the most recent versioned directory (which contains the symlinks alongside the artifacts) rather than a directory (which contains symlinks to artifacts in the parent directory). Scripts that rely on paths to symlinks, as well as tasks in `Makefile.toml`, should still work. However, any process that relies on moving images into the "expected" directory, such as the GovCloud AMI registration workflow, will break and will need to be updated.

The **second set** of changes turns the manifest parser into a crate, and makes the variant manifest the new source of truth for published image sizes. `buildsys` and `pubsys` are then updated to consume this data. This allows the `build-variant` task to generate OVAs directly, rather than relying on follow-up tasks in `Makefile.toml`, which again is in pursuit of the goal for a single task that's responsible for generating all artifacts and symlinks.

What's lost here is the ability to override AMI sizes via environment variables at the `cargo make ami` step, or to do the same for OVAs during the `cargo make` step. I didn't feel this was worth the implementation effort, or the cost of duplicate logic, since for OVAs to support this it would be necessary for both the `build` and `resign` tasks to track the environment variables and the desired image layout, just to override a property that can now be set in the variant manifest.

**Next**, the `grub-features` property in the variant manifest is upgraded to the more general `image-features` property to store feature flags. Some of these will be GRUB-related (`grub-set-private-var`, `grub-signed-config`) and some will pertain to the image as a whole (`uefi-secure-boot`). Features are also passed into package builds so that they can be used for build conditionals.

**Finally** there are a couple of minor fixes. The `repobuild` stage in the `Dockerfile` is streamlined and the parts of it that are specific to `kmod-kit` and `migrations` are moved into those stages to better separate concerns. `aws-dev` and `vmware-dev` receive the ability to configure boot settings, and `vmware-dev` moves to the newer "unified" layout. Both are done under the general theory that `dev` variants are a laboratory for experimenting with new features and functionality, where backwards compatibility can be cheerfully ignored.

**Testing done:**
Built lots of OVAs and AMIs. Verified that the images and symlinks were created correctly and that related tasks (`ami`, `repo`, `upload-ova`) still found the expected inputs and produced the expected results in terms of published disk sizes.

Confirmed that the boot settings feature works on `aws-dev` and `vmware-dev`.

Here's a [visual aid](https://gist.github.com/bcressey/454c39ad93d8f438ea8acfd515402528) for how the artifacts and symlinks are created, before and after this change.

**Terms of contribution:**


tools/rpm2img
```diff
@@ -38,16 +38,39 @@ case "${PARTITION_PLAN}" in
...
 esac
 
+# Store output artifacts in a versioned directory.
+OUTPUT_DIR="${OUTPUT_DIR}/${VERSION_ID}-${BUILD_ID}"
 mkdir -p "${OUTPUT_DIR}"
 
 FILENAME_PREFIX="${IMAGE_NAME}-${VARIANT}-${ARCH}-${VERSION_ID}-${BUILD_ID}"
+SYMLINK_PREFIX="${IMAGE_NAME}-${VARIANT}-${ARCH}"
+VERSIONED_SYMLINK_PREFIX="${IMAGE_NAME}-${VARIANT}-${ARCH}-${VERSION_ID}"
+FRIENDLY_VERSIONED_SYMLINK_PREFIX="${IMAGE_NAME}-${VARIANT}-${ARCH}-v${VERSION_ID}"
```
zmrow, line 78: The "v" :cry:

zmrow (approved):
Whew!  Nice job!

:mag:

tools/buildsys/src/builder.rs
```diff
@@ -372,19 +385,20 @@ fn create_build_dir(kind: &BuildType, name: &str, arch: &str) -> Result<PathBuf>
...
-                .map(|s| !s.ends_with(MARKER_EXTENSION))
-                .unwrap_or(false)
+        entry.path().is_dir()
+            || (entry.file_type().is_file()
+                && entry
+                    .file_name()
+                    .to_str()
+                    .map(|s| !s.ends_with(MARKER_EXTENSION))
+                    .unwrap_or(false))
+            || (entry.file_type().is_symlink())
```
webern: This is getting hard to read


 elsewhere it looks like you have an `is_marker` function,  so maybe something like:

```suggestion
entry.path().is_dir() || is_marker(entry) || entry.file_type().is_symlink()
```

Or if not, you can always store a few booleans along the way:

```rust
let is_dir = entry.path().is_dir();
let is_file = entry.file_type().is_file()
let is_marker = is_file && entry
                                .file_name()
                                .to_str()
                                .map(|s| !s.ends_with(MARKER_EXTENSION))
                                .unwrap_or(false);
is_symlink = entry.file_type().is_symlink();
```

Now the logic is easy to read
```rust
is_dir || is_marker || is_symlink
```

webern (commented):
This is pretty big and I'm probably only 20% of the way through it. I wanted to save my work, sorry for what will be multiple reviews:

nits

- This commit comment `buildsys: refactor manifest parser into crate` misled me a little since what you are doing is making a `buildsys` library crate that exposes a `manifest` module.
- I wonder if "hint" is the best name for `publish_image_size_hint_gib`. Maybe `publish_image_minimum_size_gib` or `publish_image_desired_size_gib`.

tools/buildsys/src/manifest.rs
```diff
@@ -167,100 +175,111 @@ kernel-parameters = [
...
 ]
 
-`grub-features` is a list of supported grub features.
-This list allows us to conditionally use or exclude certain grub features in specific variants.
-The only supported value at this time is `set-private-var`.
-This value means that the grub config for the current variant includes the command to find the
-BOTTLEROCKET_PRIVATE partition and set the appropriate `$private` variable for the grub to
-consume.
-Adding this value to `grub-features` enables the use of Boot Config.
+`image-features` is a map of image feature flags, which can be enabled or disabled. This allows us
```
webern, line 178: In buildsys you have added a loop to add these as --build-args when calling docker build. However in `Dockerfile` I think you need to say `ARG GRUB_SET_PRIVATE_VAR`. In other words, these are not exactly free-form, there is an enumerable list of valid strings/keys right? Should they be constrained by enum or otherwise?

Also I don't quite understand how it goes from being `grub-set-private-var` to `GRUB_SET_PRIVATE_VAR`, I didn't see that conversion to screaming snake case.

Edit: I figured it out later.

tools/buildsys/src/manifest.rs
```diff
@@ -167,100 +175,111 @@ kernel-parameters = [
...
-    pub(crate) fn kernel_parameters(&self) -> Option<&Vec<String>> {
+    pub fn kernel_parameters(&self) -> Option<&Vec<String>> {
         self.build_variant()
             .and_then(|b| b.kernel_parameters.as_ref())
     }
 
-    /// Convenience method to return the GRUB features for this variant.
-    pub(crate) fn grub_features(&self) -> Option<&Vec<GrubFeature>> {
-        self.build_variant().and_then(|b| b.grub_features.as_ref())
+    /// Convenience method to return the enabled image features for this variant.
```
webern, line 273: Are image features only ever going to be boolean flags? Should you allow for the future possibility than an image feature can have a string or numeric value?

tools/buildsys/src/builder.rs
```diff
@@ -108,6 +109,12 @@ impl PackageBuilder {
             arch = arch,
         );
 
+        if let Some(image_features) = image_features {
+            for image_feature in image_features.iter() {
+                args.build_arg(format!("{}", image_feature), "1");
```
webern, line 114: Elsewhere I have comment on two things:
- Should the allowable image features be constrained?
- Should we allow for future image features that are not flags, i.e. where they have a string or numeric value.

Even if you keep image features as a list of unconstrained booleans, I think you need to double check that the value given is "true" since someone could write `grub-set-private-var = false` in their toml.

Dockerfile
```diff
@@ -66,6 +66,7 @@ ARG VARIANT_RUNTIME
 ARG VARIANT_FAMILY
 ARG VARIANT_FLAVOR
 ARG REPO
+ARG GRUB_SET_PRIVATE_VAR
```
webern, line 69: Elsewhere I have noted that these are effectively an enum of allowable build args.

Makefile.toml
```diff
@@ -584,159 +578,12 @@ export BUILDSYS_VARIANT_FLAVOR="${BUILDSYS_VARIANT_FLAVOR}"
 # because we build host tools with cargo as well, like buildsys and pubsys.
 export CARGO_TARGET_DIR=${BUILDSYS_ROOT_DIR}/variants/target/${BUILDSYS_ARCH}
 
+rm -rf "${BUILDSYS_OUTPUT_DIR}/latest"
 cargo build \
   ${CARGO_BUILD_ARGS} \
   ${CARGO_MAKE_CARGO_ARGS} \
   --manifest-path variants/${BUILDSYS_VARIANT}/Cargo.toml
-'''
```
webern, line 591: I love seeing code removed from the Makefile! :tada:

tools/buildsys/src/manifest.rs
```diff
@@ -405,34 +476,44 @@ impl SupportedArch {
...
+        }
+    }
 }
 
-impl fmt::Display for GrubFeature {
+impl fmt::Display for ImageFeature {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            GrubFeature::SetPrivateVar => write!(f, "GRUB_SET_PRIVATE_VAR"),
+            ImageFeature::GrubSetPrivateVar => write!(f, "GRUB_SET_PRIVATE_VAR"),
```
webern, line 488: Ok, this is where the uppercase transformation happens.

tools/buildsys/src/manifest.rs
```diff
@@ -405,34 +476,44 @@ impl SupportedArch {
 }
 
 #[derive(Deserialize, Debug, PartialEq, Eq, Hash)]
-#[serde(rename_all = "kebab-case")]
-pub(crate) enum GrubFeature {
-    SetPrivateVar,
+#[serde(try_from = "String")]
+pub enum ImageFeature {
+    GrubSetPrivateVar,
```
webern, line 472: Oh, Ok, I missed this. So the image features *are* being constrained during the parsing of the Cargo.toml

tools/buildsys/src/manifest.rs
```diff
@@ -405,34 +476,44 @@ impl SupportedArch {
...
+}
+
+impl TryFrom<String> for ImageFeature {
+    type Error = Error;
+    fn try_from(s: String) -> Result<Self> {
+        match s.as_str() {
+            "grub-set-private-var" => Ok(ImageFeature::GrubSetPrivateVar),
```
webern, line 479: If interested, serde plain can do this for you. [serde_plain::derive_fromstr_from_deserialize](https://docs.rs/serde_plain/latest/serde_plain/macro.derive_fromstr_from_deserialize.html)

tools/buildsys/src/manifest.rs
```diff
@@ -141,13 +141,21 @@ overhead for the GPT labels and partition alignment. The data partition will be
 automatically resized to fill the disk on boot, so it is usually not necessary
 to increase this value.
 
+`publish-image-size-hint-gib` is the desired size of the published image in GiB.
```
webern, line 144: I think I commented on it elsewere:

```suggestion
`publish-image-desired-size-gib` is the desired size of the published image in GiB.
```

webern (commented):
I need to go through and self resolve or delete some of these things as I was figuring stuff out as I went along. I think the main, potentially useful feedback is that image-features are always assumed to be boolean (might we want other value types in the future) and I think maybe you assume that the presence of the key in the toml file means `true` (although I might have filled a filter).

tools/rpm2img
```diff
@@ -38,16 +38,39 @@ case "${PARTITION_PLAN}" in
...
 
-OS_IMAGE_BASENAME="${FILENAME_PREFIX}"
-DATA_IMAGE_BASENAME="${FILENAME_PREFIX}-data"
+OS_IMAGE_NAME="${FILENAME_PREFIX}"
+OS_IMAGE_SYMLINK="${SYMLINK_PREFIX}"
```
jpculp, line 81: Should we exclude [SC2034](https://www.shellcheck.net/wiki/SC2034) for all of `rpm2img`?

tools/buildsys/src/manifest.rs
```diff
@@ -344,42 +355,93 @@ pub enum ImageFormat {
     Vmdk,
 }
 
+#[derive(Deserialize, Debug, Copy, Clone)]
+#[serde(try_from = "i32")]
+pub struct ImageSize(i32);
```
jpculp: Is this i32 because of `pubsys` or could this be u16 instead?

tools/rpm2img
```diff
@@ -38,6 +42,28 @@ case "${PARTITION_PLAN}" in
...
 
+# Fail fast if the OVF template doesn't exist, or doesn't match the layout.
+if [ "${OUTPUT_FMT}" == "vmdk" ] ; then
+  if [ ! -s "${OVF_TEMPLATE}" ] ; then
+    echo "required OVF template not found: ${OVF_TEMPLATE}" >&2
+    exit 1
+  fi
+
+  if [ "${PARTITION_PLAN}" == "split" ] ; then
+    if ! grep -Fq '{{DATA_DISK}}' ${OVF_TEMPLATE} ; then
```
jpculp: `${OVF_TEMPLATE}` should probably be quoted (also on L60).

tools/rpm2img
```diff
@@ -304,9 +307,8 @@ veritysetup verify "${ROOT_IMAGE}" "${VERITY_IMAGE}" "${VERITY_ROOT_HASH}"
 dd if="${VERITY_IMAGE}" of="${OS_IMAGE}" conv=notrunc bs=1M seek="${partoff[HASH-A]}"
 
 # write GRUB config
```
jpculp, line 310: nit. Would this look better with a blank line where `GRUB_FEATURES_ARRAY` was?

tools/buildsys/src/manifest.rs
bcressey: The reason I liked the "hint" phrasing is because it suggests that it might be ignored under some conditions. "desired" sounds like maybe the build should fail if the desire can't be satisfied.

tools/buildsys/src/manifest.rs
bcressey: Correct, it's `i32` because of `pubsys`, to avoid dealing with fallible numeric conversions. I'll see if a `u16` might work too without much boilerplate.

Makefile.toml
bcressey: This is also my favorite part of this PR.

tools/buildsys/src/builder.rs
bcressey: The `image_features()` helper method has a filter where only "true" values are returned.

I briefly considered feature flags that aren't booleans, and figured that we could cross that bridge down the road when the use case was better understood.

tools/buildsys/src/builder.rs
```diff
@@ -411,34 +419,74 @@ where
...
+        }
+        std::fs::remove_file(&path).context(error::FileRemoveSnafu { path })?;
+        let mut parent = path.parent();
+        while let Some(p) = parent {
+            if p == top || dirs.contains(p) {
+                break;
+            }
+            dirs.insert(p.into());
+            parent = p.parent()
+        }
```
arnaldo2792: Will it be possible to clean up the parent directories here as well? Or, why do they have to be deleted after the marker and output files are deleted?

tools/buildsys/src/manifest.rs
bcressey: Perhaps I'm not clear how to use it, but it seems to lead to relatively more boilerplate.

Without the `#[serde(try_from = "String")]` container attribute, deserialization fails with an error like this, because TOML keys are always strings:
```
invalid type: string "grub-set-private-var", expected enum ImageFeature for key ...
```

With the container attribute, the type is first deserialized into a string and then converted to the enum.

`derive_fromstr_from_deserialize` gives me a `FromStr` method, but I don't see a way to get `serde` to use that method without implementing a custom deserialize method.

(I also turned up `serde_with::DeserializeFromStr`, but we don't use `serde_with` in the project today and I didn't feel like it was worth bringing in a new dependency.)

tools/rpm2img
bcressey: Yes, that seems right. `symlink_image` is just too "clever". I feel a bit bad about that, but the alternative was much more verbose.

We should also add `shellcheck` to the CI lints for `tools/rpm2*` so you don't always have to run it for me. 😀

tools/buildsys/src/builder.rs
bcressey: It would be possible but not as efficient. For `N` output files, we'd end up checking whether the parent directory was empty `N` times, rather than once.

tools/buildsys/src/manifest.rs
webern: not worth the hassle i guess, sorry for sending you down that road.

webern (approved):
My feedback was rather minor and I don't want to block this. Nice work.

arnaldo2792 (approved):

tools/rpm2img
jpculp: We could implement something like this (https://github.com/bottlerocket-os/bottlerocket/pull/2532). 🔍

bcressey:
First force push applies the suggested fixes from @webern and @jpculp - thanks!

Second force push squashes in the "fixup" commit that I missed.

webern (approved):

jpculp (approved):

PR #2506, opened 2022-10-20
trivial: format imports and fix comment

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

I noticed while doing some other work that a couple imports ended up being inserted between a comment and the import it was commenting on. This moves that comment back to where it was supposed to be and runs format on the file to organize those imports.

**Testing done:**

Ran `cargo make` to make sure there were no errors or warnings.

**Terms of contribution:**

### DIFF
--- a/sources/models/src/modeled_types/shared.rs
+++ b/sources/models/src/modeled_types/shared.rs
@@ -1,10 +1,9 @@
+use super::error;
 use lazy_static::lazy_static;
 use regex::Regex;
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
-// Just need serde's Error in scope to get its trait methods
-use super::error;
 use semver::Version;
 use serde::de::Error as _;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use snafu::{ensure, ResultExt};
 use std::borrow::Borrow;
 use std::convert::TryFrom;


sources/models/src/modeled_types/shared.rs
```diff
@@ -1,9 +1,9 @@
+use super::error;
 use lazy_static::lazy_static;
 use regex::Regex;
+use semver::Version;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 // Just need serde's Error in scope to get its trait methods
```
webern: Needing to import a trait is standard in Rust, renaming it to not conflict with another error also seems standard. I would just delete the comment so that the list of imports doesn't require further manual intervention.

```suggestion
```

webern (approved):

sources/models/src/modeled_types/shared.rs
stmcginnis: Yeah, fair enough. I'll just drop it.

jpculp (approved):

arnaldo2792:
I don't understand why the `check-fmt` command we run in the github actions missed this formatting problem, is it broken or is there something special about the crate this file belongs that prevents the formatting tool to do the right thing?

zmrow (approved):
:old_key:

stmcginnis:
> I don't understand why the check-fmt command we run in the github actions missed this formatting problem

This could be another case where I was running a different version than what is run in the gate.

I think the bigger thing that this PR now addresses is what Matt recommended of just getting rid of the comment that was in there. It really wasn't necessary, so while maybe some of the reordering wasn't strictly necessary, getting rid of the comment that was now in the wrong place is the bigger benefit.

Normally I would wait for a response, but since this is pretty trivial, has the approvals, and has become more work than I originally intended it to be, I'll just go ahead with the merge. If the response above does not fully address concerns, please let me know and I'm happy to follow up with anything to improve it. ;)

PR #2507, opened 2022-10-20
early-boot-config: set constant lifetimes for compatibility

jpculp:
**Description of changes:**

The `IDENTITY_DOCUMENT_FILE` and `FALLBACK_REGION` constants require a named lifetime parameter when building with Rust toolchains older than **1.64.0**.

This fixes an edge-case regression introduced in #2493.

**Testing done:**
- `cargo +1.63 check`
- `cargo +1.61 check`

**Terms of contribution:**

### DIFF
--- a/sources/api/early-boot-config/src/provider/aws.rs
+++ b/sources/api/early-boot-config/src/provider/aws.rs
@@ -15,8 +15,8 @@ use crate::provider::local_file::{local_file_user_data, USER_DATA_FILE};
 pub(crate) struct AwsDataProvider;
 
 impl AwsDataProvider {
-    const IDENTITY_DOCUMENT_FILE: &str = "/etc/early-boot-config/identity-document";
-    const FALLBACK_REGION: &str = "us-east-1";
+    const IDENTITY_DOCUMENT_FILE: &'static str = "/etc/early-boot-config/identity-document";
+    const FALLBACK_REGION: &'static str = "us-east-1";
 
     /// Fetches user data, which is expected to be in TOML form and contain a `[settings]` section,
     /// returning a SettingsJson representing the inside of that section.


stmcginnis (approved):
Tested with 1.63:

```bash
$ cargo make clean
$ rustup default 1.63
$ cargo --version
cargo 1.63.0 (fd9c4297c 2022-07-01
$ cargo make check
...
    Checking pubsys v0.1.0 (/home/stmcg/src/bottlerocket-os/bottlerocket/tools/pubsys)
    Finished dev [unoptimized + debuginfo] target(s) in 2m 25s
[cargo-make] INFO - Build Done in 271.91 seconds
```

Updated to 1.64 and ran again:

```bash
$ rustup default 1.63
$ cargo --version
$ cargo --version
cargo 1.64.0 (387270bc7 2022-09-16
$ cargo make check
...
help: remove these parentheses
     |
1388 -         None => (return TUF_PUBLIC_REPOSITORY.to_string()),
1388 +         None => return TUF_PUBLIC_REPOSITORY.to_string(),
     |

warning: unnecessary parentheses around match arm expression
    --> /home/stmcg/src/bottlerocket-os/bottlerocket/sources/api/schnauzer/src/helpers.rs:1392:17
     |
1392 |         None => (STANDARD_PARTITION),
     |                 ^                  ^
     |
help: remove these parentheses
     |
1392 -         None => (STANDARD_PARTITION),
1392 +         None => STANDARD_PARTITION,
     |

warning: 4 warnings emitted
...
    Checking testsys v0.1.0 (/home/stmcg/src/bottlerocket-os/bottlerocket/tools/testsys)
    Checking pubsys v0.1.0 (/home/stmcg/src/bottlerocket-os/bottlerocket/tools/pubsys)
    Finished dev [unoptimized + debuginfo] target(s) in 2m 19s
[cargo-make] INFO - Build Done in 298.48 seconds.
```

So some minor warnings in there that we should probably address at some point, but this change fixes the issue it is trying to address.

bcressey:
On the one hand I don't like gratuitous breakage, but I'm also not excited about trying to preserve compatibility with any release of Rust older than the one that CI is pinned to.

To put it another way, if the bug was to fix an issue where 1.60.0 couldn't compile the project, I don't think we'd bother.

It's only because 1.63.0 is also relatively recent that it feels like a regression.

jpculp:
> So some minor warnings in there that we should probably address at some point, but this change fixes the issue it is trying to address.

With each new version of the rust toolchain we'll get a new batch of `clippy` suggestions. My approach has been each time I work in a crate that has some warnings I'll address them in a second housekeeping commit like in #2493.

jpculp:
> On the one hand I don't like gratuitous breakage, but I'm also not excited about trying to preserve compatibility with any release of Rust older than the one that CI is pinned to.
> 
> To put it another way, if the bug was to fix an issue where 1.60.0 couldn't compile the project, I don't think we'd bother.
> 
> It's only because 1.63.0 is also relatively recent that it feels like a regression.

Right. I think currently everything in Bottlerocket builds with **1.61+**, I would hate for _this_ to be the thing that forces **1.64**, especially since, like you said, **1.63** is relatively recent.

bcressey (approved):
I'm fine with this since it's a minor change, but I'd also be fine with requiring >= 1.64.0.

webern (approved):

zmrow (approved):

PR #2508, opened 2022-10-20
housekeeping: Cancel GH Action on new push

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

When a PR is proposed to the develop branch, a set of GitHub Action jobs are kicked off to test the various configuration we need to cover. With the current configuration of our workflow, if the developer notices an issue and pushes updates to the source branch it will kick off a new set of jobs for that updates code, but the jobs running for the previous revision will keep running.

These jobs take a lot of time and resources to run. Once a new change is pushed to the source branch, there is no need to continue testing the older version of the code as it is no longer relevant.

This updates our workflow definition to cancel and currently running jobs for the same source (`github.ref`) to avoid unnecessary runner load that may slow down getting results for current PRs.

**Testing done:**

Testing will be done by GitHub Action runs. If any issues are found we can quickly revert the change.

Ref:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idenv

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,10 @@ on:
       # Sample config files and OpenAPI docs
       - '**.yaml'
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: [self-hosted, linux, x64]


jpculp:
You might be able to test this prior to merge by committing a temporary commit or amending the current commit (before the current actions finish running). If you do a temporary commit you should drop into draft just to be safe.

stmcginnis:
I thought with the way we have things configured, when I've tried that before it wouldn't take effect until the workflow update was actually merged. I'll give it a shot though - great idea!

jpculp (approved):
Looks like it works to me!

bcressey (approved):
The GitHub mobile app doesn't show the correct status, but that's a fairly minor concern.

It shows "20 failed, 42 passed" rather "31 passed", with an overall verdict of "failed". This reflects the status that 11 of the previous run's checks and all 31 of the current run's checks passed.

jpculp:
I'm also seeing the strange status on the GH mobile app (Android/1.85.0).

webern (approved):

PR #2357, opened 2022-08-17
migrations: archive old migrations before v1.9.1

etungsten:
**Issue number:**
N/A

**Description of changes:**
For reference, the previous archiving: https://github.com/bottlerocket-os/bottlerocket/pull/1699. This makes the same change.
```
See 2ce22f1 for the last time we did this.

Same explanation as before:
These migrations have already been built and released with Bottlerocket
releases and don't need to be built again.  The build system and os package
currently build everything in the migrations directory, which can take a
significant amount of time with unneeded migrations.

Their source is still useful as a reference, and the files are small, so it's
worthwhile keeping them in-tree rather than having to search git history.

```

**Testing done:**
Before this change, time to build `os`, `release` packages and `aws-k8s-1.23` variant: `Finished dev [optimized] target(s) in 3m 25s`
After this change, time to build `os`, `release` packages and `aws-k8s-1.23` variant: `Finished dev [optimized] target(s) in 1m 58s`

Before this change, the contents of the migrations rpm shows:
```
[ec2-user@ip-172-31-32-162 rpms]$ rpm -qlp bottlerocket-x86_64-migrations-0.0-0.aarch64.rpm         
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations                                                                        
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.3.0_control-container-v0-5-2         
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.3.0_etc-hosts-service    
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.3.0_hostname-affects-etc-hosts
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.4.0_registry-mirror-representation
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.4.2_admin-container-v0-7-3
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.4.2_control-container-v0-5-3
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.5.0_oci-hooks-setting
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.5.0_oci-hooks-setting-metadata
[...]

PR #2518, opened 2022-10-25
Archive 1.10.0 migrations

arnaldo2792:
**Description of changes:**
Regardless of the next release version, archiving migrations improves builds' speed. This is similar to what was done in #2357 and #1699.

**Testing done:**
Builds are successful after archiving the migrations.

**Terms of contribution:**


webern:
I can think of reasons we might want to keep the last N versions around. Was there a discussion and decision that we will archive migrations right away after a release?

arnaldo2792:
> Was there a discussion and decision that we will archive migrations right away after a release?

Yes, but wasn't registered anywhere :crying_cat_face: and it happened too long ago so I don't remember most of the discussion. But we can start a thread now:

I do remember that the argument was that we end up spending time (multiplied by all the variants) building artifacts that already exist in the tuf repos. This also affects the github workflows execution time.

stmcginnis:
I remember the discussion, but I don't remember where. The key realization for me was that archiving the migrations doesn't mean someone can't upgrade from one of those older versions. The migrations are still available, they just aren't rebuilt every time.

stmcginnis (approved):
lgtm

etungsten:
> I can think of reasons we might want to keep the last N versions around

Can you elaborate on some of them?

webern:
> > I can think of reasons we might want to keep the last N versions around
> 
> Can you elaborate on some of them?

I was mostly thinking about development work in which you want to build a repo with a few earlier versions to check migration paths, but I guess you would build from tags for that anyway, so nevermind.

etungsten (dismissed):

webern (changes requested):
Well, I hadn't considered this eventuality, but it looks like we want to fix a bug in a 1.10.x migration: #2523 

I'm marking this as "Request Changes" to pause merging for now.

arnaldo2792:
@stmcginnis created #2523 to fix the migration, I'll remove that migration this PR.

arnaldo2792:
(forced push adds back the 1.10.1 migrations, but I forgot to removed them from archive and add them to `sources/Cargo.toml`)

arnaldo2792:
(forced push adds back v1.10.1 migrations to `sources/api/migration/migrations` and to `sources/Cargo.toml` / `sources/Cargo.lock`)

arnaldo2792:
(forced push is a rebase onto develop, to make sure I have the fixes in https://github.com/bottlerocket-os/bottlerocket/pull/2523)

webern (approved):

etungsten (approved):

Issue #2293, opened 2022-07-22
`net.toml`: Support configuration using MAC instead of device name


zmrow:
Currently the only way to configure a network device in `net.toml` is by device name:
```
[eno1]
...
```

It would be nice to be able to use MAC address as the key since many organizations already have the information and may not know how the device will be named.
```
[ab:cd:ef:gh:ij]
...
```

bcressey:
It feels oddly specific to know the MAC address but not know or care what network device is the "primary" one for the purposes of getting DHCP leases, setting the hostname, or registering the node with the Kubernetes cluster.

What's the larger architectural context here?

Cajga:
This initiative seems to be interesting. Let me add our use case:
EKS-A bare metal requires to define the MAC address at first place so, the built-in DHCP server will reply on the DHCP request from the new HW. The MAC is normally written on the HW/Server with a sticker.
In the next step, one needs to define the net.toml in a template file which will work for the node. But as hinted above, no one knows the Linux interface name for a new HW before it booted a Linux... So, what we do now is that we boot a Ubuntu Live and check the interface name so we can create the net.toml file in the EKS-A bare metal config.

Personally, I am not sure if using the MAC here is the best solution (as one normally is able to change a MAC of a given interface through network config) but who knows :)

PR #2519, opened 2022-10-25
wicked: allow interface configuration via Ethernet permanent address

markusboehme:
**Issue number:** #2293 

**Description of changes:** Add support for identifying Ethernet interfaces to configure via their permanent/hardware/MAC address to wicked. For example, allow the following:

```xml
<interface>
    <name namespace="ethernet">
        <permanent-address>52:54:00:12:34:56</permanent-address>
    </name>
    <ipv4:dhcp>
        <enabled>true</enabled>
    </ipv4:dhcp>
</interface>
```

to configure the Ethernet interface with MAC address 52:54:00:12:34:56 via DHCP.

Note that:

1. These are downstream patches to wicked, eventually also to be proposed to the upstream project.
2. These only prepare being able to configure interfaces via other means than interface name. Components like netdog will still need to be adapted.

**Testing done:** In a local VM, I successfully configured interfaces both via interface name and MAC address (at first by modifying `/etc/wicked/ifconfig/` and restarting wicked components, later by just having netdog write the above XML snippet on boot).

**Terms of contribution:**

### DIFF
--- a/packages/wicked/1005-client-validate-ethernet-namespace-node.patch
+++ b/packages/wicked/1005-client-validate-ethernet-namespace-node.patch
@@ -1,0 +1,109 @@
+From 63bd7bb45b45ef7a44afb4190a7f04bf83e841f5 Mon Sep 17 00:00:00 2001
+From: Markus Boehme <markubo@amazon.com>
+Date: Tue, 25 Oct 2022 18:05:20 +0000
+Subject: [PATCH] client: validate ethernet namespace node
+
+Validate the ethernet namespace node in config files to identify
+interfaces via other means than their name, here via their hardware
+address. This restores a previously existing feature (see
+samples/wicked/eth0-identify.xml) that stopped working due to extra
+validation added at a later point.
+
+Fixes: ad7b47d0a ("client: Check for config priorities at reading stage")
+Signed-off-by: Markus Boehme <markubo@amazon.com>
+---
+ client/read-config.c | 56 ++++++++++++++++++++++++++++++++++++++++----
+ 1 file changed, 51 insertions(+), 5 deletions(-)
+
+diff --git a/client/read-config.c b/client/read-config.c
+index 231d656..8965848 100644
+--- a/client/read-config.c
++++ b/client/read-config.c
+@@ -327,19 +327,23 @@ __ifconfig_read_get_ifname(xml_node_t *ifnode, unsigned int *ifindex)
+ 	const char *namespace;
+ 	char *ifname = NULL;
+ 
+-	/* Check for   <name> node */
++	/* Check for <name> node */
+ 	nnode = xml_node_get_child(ifnode, "name");
+-	if (!nnode || ni_string_empty(nnode->cdata)) {
++	if (!nnode) {
+ 		ni_debug_ifconfig("cannot get interface name - "
+-			"config has no valid <name> node");
++			"config has no <name> node");
+ 		goto error;
+ 	}
+ 
+-	ifname = nnode->cdata;
+-
+ 	/* Resolve a namespace if specified */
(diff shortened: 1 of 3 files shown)


packages/wicked/1005-client-validate-ethernet-namespace-node.patch
```diff
@@ -0,0 +1,107 @@
...
+ 		unsigned int value;
+ 		char name_buf[IF_NAMESIZE+1];
+ 
++		if (ni_string_empty(nnode->cdata)) {
++			ni_debug_ifconfig("config is missing an interface name in <name> node");
++			goto error;
++		}
++		ifname = nnode->cdata;
++
+ 		if (ni_parse_uint(ifname, &value, 10) < 0) {
```
bcressey, line 61: Since the interface "name" for the `ifindex` namespace is actually an integer, would it be more consistent with the other two branches to use the MAC address as the "name" for the `ethernet` namespace?

It looks like using the `permanent-address` field is suggested by the original docs, so what you have makes sense to me.

packages/wicked/1005-client-validate-ethernet-namespace-node.patch
```diff
@@ -0,0 +1,107 @@
...
+ 	}
++	else if (ni_string_eq(namespace, "ethernet")) {
++		xml_node_t *addrnode = NULL;
++		ni_string_array_t all_ifnames = NI_STRING_ARRAY_INIT;
++		unsigned int i;
++
++		addrnode = xml_node_get_child(nnode, "permanent-address");
++		if (!addrnode || ni_string_empty(addrnode->cdata)) {
++			ni_debug_ifconfig("cannot get interface name - "
++				"config has no valid <ethernet> node");
```
bcressey: nit: isn't it a `<name>` node still, rather than an `<ethernet>` node?

packages/wicked/1005-client-validate-ethernet-namespace-node.patch
```diff
@@ -0,0 +1,107 @@
...
++			goto error;
++		}
++
++		if (ni_scandir("/sys/class/net", NULL, &all_ifnames) <= 0)
++			goto error;
```
bcressey: nit: add a debug message indicating the read failure?

bcressey (approved):

foersleo (approved):

arnaldo2792 (approved):

zmrow:
I like @bcressey 's suggestions for changes - other than that this looks great!

zmrow (approved):

packages/wicked/1005-client-validate-ethernet-namespace-node.patch
markusboehme: I think I know what you mean, but am not 100% sure, so to clarify: Is what you're suggesting to store the MAC address in `ifname` when handling the `ethernet` namespace, similar to how the interface index is stored in `ifname` when handling the `ifindex` namespace?

If so, no, `ifname` needs to be the actual interface name in both cases. The diff is hiding the part of the code handling the `ifindex` namespace where the index temporarily stored in `ifname` is used to look up the actual interface name (edit: https://github.com/openSUSE/wicked/blob/a5762e9f372b9f8da3be4a7803f0a6d81c6dda6a/client/read-config.c#L356,L364). Unfortunately, the variable reuse doesn't help to make this clear either.

markusboehme:
In the force push:

* Improving debug message for invalid `<name>` node in `ethernet` namespace in config
* Adding extra debug logging for failure to enumerate network interfaces

zmrow (approved):
:taco: 

Thanks!

bcressey (approved):

PR #2524, opened 2022-10-26
testsys: Enable multiple tests on a cluster

### LINKED ISSUE #2531: testsys: Enable multiple tests on a single cluster
**What I'd like:**

The TestSys framework provides the ability to mark resources as conflicting. Resources will not start their creation if any conflicting resources exist. `testsys` should use this feature to allow 2 instance providers to use the same cluster provider without the instances being launched at the same time.

https://github.com/bottlerocket-os/bottlerocket-test-system/commit/cc49c94d18c45565f443d5b890ad39f808739ebe

ecpullen:
**Issue number:**

Closes #2531 

**Description of changes:**

TestSys uses labels on the crd's to determine which EC2 resources rely on a specific EKS cluster. If a new instance provider requires a cluster that is already created, TestSys marks the new instance provider as conflicting with all other ones which prevents the new instances from being created until the other instance providers have been deleted.

**Testing done:**

Tested with
```bash
cargo make TESTSYS_TEST=migration test
cargo make test
```

The migration test ran first and the quick test started once the migration test was finished.

**Terms of contribution:**

### DIFF
--- a/tools/testsys/src/aws_resources.rs
+++ b/tools/testsys/src/aws_resources.rs
@@ -12,7 +12,9 @@ use k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta;
 use k8s_openapi::serde_json::Value;
 use log::debug;
 use maplit::btreemap;
+use model::clients::{AllowNotFound, CrdClient};
 use model::constants::NAMESPACE;
+use model::test_manager::{SelectionParams, TestManager};
 use model::{
     Agent, Configuration, Crd, DestructionPolicy, Resource, ResourceSpec, SecretName, Test,
     TestSpec,
@@ -40,6 +42,7 @@ impl AwsK8s {
     /// Create the necessary test and resource crds for the specified test type.
     pub(crate) async fn create_crds(
         &self,
+        client: &TestManager,
         test: TestType,
         testsys_images: &TestsysImages,
     ) -> Result<Vec<Crd>> {
@@ -56,18 +59,56 @@ impl AwsK8s {
                 self.kube_arch(),
                 self.kube_variant(),
             )?;
+            // Check for existing cluster crd
+            let cluster_exists = client
+                .resource_client()
+                .get(&cluster_name)
+                .await
+                .allow_not_found(|_| ())?
+                .is_some();
+            if !cluster_exists {
+                debug!("Cluster crd does not exist");
+                crds.push(self.eks_crd(cluster_name, testsys_images, &test)?)
+            }
+            // Check for conflicting resources (ones that use the same cluster)
+            let conflicting_resources: Vec<String> = if cluster_exists {
+                client
+                    .list(&SelectionParams::Label(format!(
+                        "testsys/cluster={}, testsys/type=instances",
+                        cluster_name
+                    )))
+                    .await?
+                    .into_iter()
+                    // Retrieve the name from each resource
+                    .filter_map(|crd| crd.name())
+                    .collect()
+            } else {
+                Default::default()
+            };
+
             crds.append(&mut match &test {
                 TestType::Conformance => self.sonobuoy_test_crds(
+                    &test,
                     testsys_images,
                     SonobuoyMode::CertifiedConformance,
                     cluster_name,
+                    &conflicting_resources,
                 )?,
-                TestType::Quick => {
-                    self.sonobuoy_test_crds(testsys_images, SonobuoyMode::Quick, cluster_name)?
-                }
+                TestType::Quick => self.sonobuoy_test_crds(
+                    &test,
+                    testsys_images,
+                    SonobuoyMode::Quick,
+                    cluster_name,
+                    &conflicting_resources,
+                )?,
                 TestType::Migration => {
-                    self.migration_test_crds(cluster_name, testsys_images)
-                        .await?
+                    self.migration_test_crds(
+                        cluster_name,
+                        &test,
+                        &conflicting_resources,
+                        testsys_images,
+                    )
+                    .await?
                 }
             })
         }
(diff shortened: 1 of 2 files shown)


tools/testsys/src/aws_resources.rs
```diff
@@ -76,14 +117,30 @@ impl AwsK8s {
...
         testsys_images: &TestsysImages,
         sonobuoy_mode: SonobuoyMode,
         cluster_name: &str,
+        conflicting_resources: &[String],
     ) -> Result<Vec<Crd>> {
         let crds = vec![
-            self.eks_crd(cluster_name, testsys_images)?,
-            self.ec2_crd(cluster_name, testsys_images, None)?,
-            self.sonobuoy_crd("-test", cluster_name, sonobuoy_mode, None, testsys_images)?,
+            self.ec2_crd(
```
webern, line 127: this was already here but the name of the function `sonobuoy_test_crds`, conflates `sonobuoy` with `eks` (or at least `ec2`).

tools/testsys/src/aws_resources.rs
```diff
@@ -170,10 +238,22 @@ impl AwsK8s {
...
+    where
+        S1: Into<String>,
+        S2: Into<String>,
+    {
         btreemap! {
             "testsys/arch".to_string() => self.arch.to_string(),
             "testsys/variant".to_string() => self.variant.to_string(),
+            "testsys/cluster".to_string() => cluster_name.into(),
+            "testsys/type".to_string() => testsys_type.into(),
+            "testsys/test".to_string() => test_type.to_string(),
```
webern, line 256: It occurs to me, relating to this https://github.com/bottlerocket-os/bottlerocket-test-system/issues/628, that we use `testsys` here as if we "own" that word, but in the afore mentioned issue we are contemplating making `testsys` the namespace of the more general testing system. I guess it's fine and reflects the journey we have taken...

webern (approved):
Nice.

I'm becoming concerned that `aws_resources.rs` is a huge bunch of code that might not be as organized or as self-documenting as we'd like. That being said, I don't want to slow you down at this point.

etungsten (approved):
I'm OK with the code changes. 

~One thing I found confusing at first is I wasn't sure what problem are we solving and what behavior are we changing. It would be nice if you can describe in the commit message/PR description what the behavior was before this change and what the behavior is now with the change.~

Edit: I now see the added linked issue. Thanks!

tools/testsys/src/aws_resources.rs
```diff
@@ -56,18 +59,56 @@ impl AwsK8s {
                 self.kube_arch(),
                 self.kube_variant(),
             )?;
+            // check for existing cluster crd
```
arnaldo2792: Nit: You have two comments that state with lowercase, and one with uppercase. For consistency, default to uppercase.

tools/testsys/src/aws_resources.rs
```diff
@@ -108,16 +167,22 @@ impl AwsK8s {
...
-            .expect("The EC2 instance provider crd is missing a name.");
+        let ec2 = self.ec2_crd(
+            cluster_name,
+            "migration",
+            test_type,
+            conflicting_resources,
+            testsys_images,
+            Some(ami),
+        )?;
+        let instance_provider = ec2.name().unwrap();
```
arnaldo2792: I think `expect` should be used here to handle the theoretical non-existing error.

tools/testsys/src/aws_resources.rs
```diff
@@ -56,18 +59,56 @@ impl AwsK8s {
...
+                .get(&cluster_name)
+                .await
+                .allow_not_found(|_| ())?
+                .is_some();
+            if !cluster_exists {
+                debug!("Cluster crd does not exist");
+                crds.push(self.eks_crd(cluster_name, testsys_images, &test)?)
+            }
+            // check for conflicting resources (ones that use the same cluster)
+            let conflicting_resources: Vec<String> = if cluster_exists {
```
arnaldo2792, line 74: I saw that `conflicting_resources` is created here, but then passed down to `sonobuoy_test_crds` -> `ec2_crd` -> `Resource { conflicts_with }`. There could be a way to increase cohesion and pass down the reference only once, but it will require some refactoring, obviously not in this PR.

For this concern, the rusteceans in the team could weight in, is passing down dependencies like this until the method where they are used OK in rust?

arnaldo2792 (approved):

tools/testsys/src/aws_resources.rs
ecpullen: The best I could come up with was passing `&[String]` instead of `Vec<String>`, maybe others have a better idea?

ecpullen:
^ Rebase, fixed the unwrap and 2 comments

tools/testsys/src/aws_resources.rs
```diff
@@ -290,28 +383,31 @@ impl AwsK8s {
...
         test_name_suffix: &str,
         cluster_name: &str,
+        test_type: &TestType,
+        ec2_resource_name_suffix: &str,
         sonobuoy_mode: SonobuoyMode,
         depends_on: Option<Vec<String>>,
         testsys_images: &TestsysImages,
     ) -> Result<Crd> {
-        let ec2_resource_name = format!("{}-instances", cluster_name);
+        let ec2_resource_name = format!("{}-instances-{}", cluster_name, ec2_resource_name_suffix);
```
bcressey, line 404: I feel like `clippy` is on to something here. 😀 

It seems like there's a type that could be factored out, e.g.
```
struct CrdInput {
    cluster_name: ...,
    test_type: ...,
    suffix: ...,
    testsys_images: ...,
}
```

With options for the fields that aren't common, or else a constructor that takes a base `CrdInput` and turns it into a `SonobuoyCrdInput`.

This is not a blocker but the module feels overdue for refactoring, and that might be better done before the next round of features.

tools/testsys/src/aws_resources.rs
bcressey: My instinct here would be to create a new type, like the `CrdInput` type I mentioned, and make `conflicting_resources` a method on that type, assuming it has all the fields it needs to handle the query.

That way, you can call the method where it's ultimately used, rather than at the outset where you have to pass it down into the call chain.

bcressey:
What is the behavior of two invocations of the same test type? E.g.
```
cargo make TESTSYS_TEST=migration test
cargo make TESTSYS_TEST=migration test
```

Are they de-duplicated, or does the second one wait for the first to complete and then run, clobbering its results, or something else?

ecpullen:
> What is the behavior of two invocations of the same test type? E.g.
> 
> ```
> cargo make TESTSYS_TEST=migration test
> cargo make TESTSYS_TEST=migration test
> ```
> 
> Are they de-duplicated, or does the second one wait for the first to complete and then run, clobbering its results, or something else?

The first test will run, the second call will error out because the migration crds already exist in the cluster (from the first run).
Once the build id is appended to the crds, creating tests from different builds will be possible. We could also append a uid to prevent name collisions.

Issue #2257, opened 2022-06-30
Unable to make custom Kubernetes Settings types with `inner` types other than `String`


mchaker:
**Image I'm using:**

Any k8s variant

**What I expected to happen:**

Ability to use `inner: i32` or any other type besides `inner: String` in the Kubernetes Settings model (https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/src/modeled_types/kubernetes.rs).

**What actually happened:**

When trying to use any `inner` type besides `String`, compile errors result from mismatched types between what `string_impls_for!()` expects, and having `inner: i32`.

`string_impls_for!()` is required due to the following error that appears when not using `string_impls_for!()` on a custom Kubernetes Settings type:

```plain
the trait bound `modeled_types::kubernetes::ImageGCHighThresholdPercent: _::_serde::Serialize` is not satisfied
required because of the requirements on the impl of `_::_serde::Serialize` for `std::option::Option<modeled_types::kubernetes::ImageGCHighThresholdPercent>`rustc_[E0277](https://doc.rust-lang.org/error-index.html#E0277)_ (https://doc.rust-lang.org/error-index.html#E0277)
mod.rs(1899, 12): required by a bound in `_::_serde::ser::SerializeStruct::serialize_field`
the trait bound `modeled_types::kubernetes::ImageGCHighThresholdPercent: _::_serde::Deserialize<'_>` is not satisfied
required because of the requirements on the impl of `_::_serde::Deserialize<'_>` for `std::option::Option<modeled_types::kubernetes::ImageGCHighThresholdPercent>`
```

**How to reproduce the problem:**

Construct a new Kubernetes Settings type with an `inner` type other than `String` and try to build Bottlerocket.

webern:
I would be interested in working on this. The key issue is that we support returning a `&str` of `inner` and the only way to do that is to store a `String` internally. So we would need to either forego some of the "stringyness" of these types, or we would need for them to have a way to store both the `String` representation *and* the actual desired inner type.

Edit:

Basically think of it like this: if you're holding an `i64` and you are required to implement this:

```rust
fn as_str(&self) -> &str {
   // uh oh, I can't
}
```

The initial goal was for these types to behave like `String`s in all contexts. So we can either forego the &str, or we can make some machinery that creates and stores the `String` representation next to the actual type.

mchaker:
> I would be interested in working on this.

[...]

Issue #2463, opened 2022-09-28
Add validation for `settings.kubernetes.log-level` value


stmcginnis:
**What I'd like:**

The `log-level` setting controls logging verbosity for kubelet. The only real valid values for this setting are 0 through 8, but since we can currently only use an unsigned integer type we aren't able to validate settings are within an expected range.

If/when we are able to implement a custom type (see https://github.com/bottlerocket-os/bottlerocket/issues/2257) for numeric settings, we should take advantage of that for this setting to make sure we restrict things to the valid range.

**Any alternatives you've considered:**

Alternative is to keep it as is with an unsigned value. These "work", just are not as correct.

webern:
This is pseudo-blocked by #2257, I can take this one after 2257.

webern:
Closing because kubelet accepts values out of this range: see https://github.com/bottlerocket-os/bottlerocket/pull/2595#issuecomment-1331445809

PR #2527, opened 2022-10-26
Add additional ECS API configurations

### LINKED ISSUE #2483: Enable more ECS features
**What I'd like:**

We should investigate the work required to enable these features in the ECS agent:

```
ECS_TASK_METADATA_RPS_LIMIT
ECS_CONTAINER_STOP_TIMEOUT
ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION
ECS_RESERVED_MEMORY
```

arnaldo2792:
**Issue number:**

Closes #2483

**Description of changes:**

With this, the following configurations will be supported through API settings:

- ECS_CONTAINER_STOP_TIMEOUT
- ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION
- ECS_TASK_METADATA_RPS_LIMIT
- ECS_RESERVED_MEMORY

For `ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION` and `ECS_CONTAINER_STOP_TIMEOUT`, I added the new `ECSDurationValue` struct used to represent a `time.Duration` string. I decided to use a new struct instead of refactoring `KubernetesDurationValue` to keep both duration values decoupled from one another, in case the type of any of them changes in future releases. The two aforementioned configurations are rendered in the `/etc/ecs/ecs.config` file, which holds environment variables passed to the ECS agent's daemon. This is because parsing valid `time.Duration` strings in golang requires calls to `time.ParseDuration` which isn't called by the `json.Unmarshall` function used by the ECS agent to parse the JSON formatted configuration at `/etc/ecs/ecs.config.json`. Even though it is possible to deserialize `time.Duration` values from a JSON string, I decided against this since the values in the JSON file must be of type [`int64`] which isn't as user friendly as the duration strings `1m`, `1s`, `1h`, etc.

`ECS_TASK_METADATA_RPS_LIMIT` contains a comma-separated string with two values used to set up the throttling configurations of the ECS agent's metadata service. The two values are assigned to [two fields] of the struct type used to represent the agent's configuration. These two values are independent from one another, meaning that a) the agent doesn't perform [coupled validations] on the values, and b) if any of them is missing or is zero, the agent will use [default values]. Since both values must be [integers], they can be deserialized without any helper functions (unlike `time.Duration`). As a result, I decided to keep these two values as separate settings in the API (`metadata-service-rps` and `metadata-service-burst`) instead of one single configuration, and render them in `/etc/ecs/ecs.config.json` . The [official AWS documentation] for the agent's configurations clearly states what each value in the tuple represents, so it should be relatively simple for users to migrate from the comma-separated configuration. No validation is required to verify that the new configurations are positive integers since the ECS agent won't fail to start, also a [useful message] is printed.

`ECS_RESERVED_MEMORY` represents the amount of memory (in MB) that the agent won't use for the allocated tasks. Since this configuration is of type `unit16` and it can be serialized without any helper functions, I decided to render it in `/etc/ecs/ecs.config.json`.

TODO:
- [x] Add documentation for the new ECS settings.

[`int64`]: https://pkg.go.dev/time#Duration
[two fields]: https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/config.go#L515
[coupled validations]: https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/config.go#L360
[default values]: https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/config.go#L195
[official AWS documentation]: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html
[integers]: https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/config.go#L360
[datastore serializer]: https://github.com/bottlerocket-os/bottlerocket/blob/2a202f877fefbbfc54e689b164460ced6d66dacc/sources/api/datastore/src/serialization/pairs.rs#L172
[useful message]: https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/config.go#L361

**Testing done:**

- In my own 1.11.0 variant, I verified I could set the new API settings, and that the ECS agent picked them up:
  - `task-cleanup-wait`: I confirmed that the container resources were cleaned up after the specified time
  -  `container-stop-timeout`: I created a container with a process that ignored `SIGTERM` signals; the agent sent `SIGKILL` after the specified time
  - `metadata-service-rps`, `metadata-service-burst`: I used `-1` as the values for the settings and confirmed that the ECS agent logged a warning
  - `reserved-memory`: I confirmed the ECS agent logged the [remaining memory]
- I did upgrade/downgrade testing and confirmed the host came back up

[remaining memory]: https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/api/ecsclient/client.go#L274

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -533,6 +533,8 @@ These settings can be changed at any time.
 * `settings.ecs.allow-privileged-containers`: Whether launching privileged containers is allowed on the container instance.
   If this value is set to false, privileged containers are not permitted.
   Bottlerocket sets this value to false by default.
+* `settings.ecs.container-stop-timeout`: Time to wait for the task's containers to stop on their own before they are forcefully stopped.
+Valid time units include `s`, `m`, and `h`, e.g. `1h`, `1m1s`.
 * `settings.ecs.enable-spot-instance-draining`: If the instance receives a spot termination notice, the agent will set the instance's state to `DRAINING`, so the workload can be moved gracefully before the instance is removed. Defaults to `false`.
 * `settings.ecs.image-pull-behavior`: The behavior used to customize the [pull image process](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html#ecs-agent-availparam) for your container instances.
   Supported values are `default`, `always`, `once`, `prefer-cached`, and the default is `default`.
@@ -541,6 +543,13 @@ These settings can be changed at any time.
   Bottlerocket enables the `json-file`, `awslogs`, and `none` drivers by default.
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.metadata-service-rps`: The steady state rate limit of the throttling configurations set for the task metadata service.
+* `settings.ecs.metadata-service-burst`: The burst rate limit of the throttling configurations set for the task metadata service.
+* `settings.ecs.reserved-memory`: The amount of memory, in MiB, reserved for critical system processes.
+* `settings.ecs.task-cleanup-wait`: Time to wait before the task's containers are removed after they are stopped.
+Valid time units are `s`, `m`, and `h`, e.g. `1h`, `1m1s`.
+
+  **Note**: `metadata-service-rps` and `metadata-service-burst` directly map to the values set by the `ECS_TASK_METADATA_RPS_LIMIT` environment variable.
 
 #### CloudFormation signal helper settings
 
--- a/packages/ecs-agent/ecs.config
+++ b/packages/ecs-agent/ecs.config
@@ -16,3 +16,9 @@ ECS_ENGINE_AUTH_DATA='{
         {{~#if password~}},"password": "{{{password}}}"}{{/if}}
     {{~/each~}}}}'
 {{/if}}
+{{#if settings.ecs.container-stop-timeout}}
+ECS_CONTAINER_STOP_TIMEOUT="{{settings.ecs.container-stop-timeout}}"
+{{/if}}
+{{#if settings.ecs.task-cleanup-wait}}
+ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION="{{settings.ecs.task-cleanup-wait}}"
+{{/if}}
--- a/sources/api/migration/migrations/v1.11.0/ecs-additional-configurations/src/main.rs
+++ b/sources/api/migration/migrations/v1.11.0/ecs-additional-configurations/src/main.rs
@@ -1,0 +1,26 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddSettingsMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added additional configurations for the ECS agent
+fn run() -> Result<()> {
+    migrate(AddSettingsMigration(&[
+        "settings.ecs.container-stop-timeout",
+        "settings.ecs.task-cleanup-wait",
+        "settings.ecs.metadata-service-rps",
+        "settings.ecs.metadata-service-burst",
+        "settings.ecs.reserved-memory",
+    ]))
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -147,7 +147,7 @@ use crate::de::{deserialize_mirrors, deserialize_node_taints};
 use crate::modeled_types::{
     BootConfigKey, BootConfigValue, BootstrapContainerMode, CpuManagerPolicy, CredentialProvider,
     DNSDomain, ECSAgentImagePullBehavior, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue,
-    EtcHostsEntries, FriendlyVersion, Identifier, ImageGCHighThresholdPercent,
+    ECSDurationValue, EtcHostsEntries, FriendlyVersion, Identifier, ImageGCHighThresholdPercent,
     ImageGCLowThresholdPercent, KmodKey, KubernetesAuthenticationMode, KubernetesBootstrapToken,
     KubernetesCloudProvider, KubernetesClusterDnsIp, KubernetesClusterName,
     KubernetesDurationValue, KubernetesEvictionHardKey, KubernetesLabelKey, KubernetesLabelValue,
@@ -230,6 +230,11 @@ struct ECSSettings {
     loglevel: ECSAgentLogLevel,
     enable_spot_instance_draining: bool,
     image_pull_behavior: ECSAgentImagePullBehavior,
+    container_stop_timeout: ECSDurationValue,
+    task_cleanup_wait: ECSDurationValue,
+    metadata_service_rps: i64,
+    metadata_service_burst: i64,
+    reserved_memory: u16,
 }
 
 #[model]
--- a/sources/models/src/modeled_types/ecs.rs
+++ b/sources/models/src/modeled_types/ecs.rs
@@ -291,3 +291,69 @@ mod test_ecs_agent_image_pull_behavior {
         }
     }
 }
+
+/// ECSDurationValue represents a string that contains a valid ECS duration value
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ECSDurationValue {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref ECS_DURATION_VALUE: Regex =
+        Regex::new(r"^(([0-9]+\.)?[0-9]+h)?(([0-9]+\.)?[0-9]+m)?(([0-9]+\.)?[0-9]+s)?(([0-9]+\.)?[0-9]+ms)?(([0-9]+\.)?[0-9]+(u|µ)s)?(([0-9]+\.)?[0-9]+ns)?$").unwrap();
+}
+
+impl TryFrom<&str> for ECSDurationValue {
+    type Error = error::Error;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        ensure!(
+            !input.is_empty() && ECS_DURATION_VALUE.is_match(input),
+            error::InvalidECSDurationValueSnafu { input }
+        );
+        Ok(ECSDurationValue {
+            inner: input.to_string(),
+        })
+    }
+}
+
+string_impls_for!(ECSDurationValue, "ECSDurationValue");
+
+#[cfg(test)]
+mod test_ecs_duration_value {
+    use super::ECSDurationValue;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn valid_values() {
+        for ok in &[
+            "99s",
+            "20m",
+            "1h",
+            "1h2m3s",
+            "4m5s",
+            "2h3s",
+            "1.5h3.5m",
+            "1ms1us1ns",
+            "1s1µs1ns",
+        ] {
+            ECSDurationValue::try_from(*ok).unwrap();
+        }
+    }
+
+    #[test]
+    fn invalid_values() {
+        for err in &[
+            "",
+            "100",
+            "...3ms",
+            "1..5s",
+            "ten second",
+            "1m2h",
+            "1y2w",
+            &"a".repeat(23),
+        ] {
+            ECSDurationValue::try_from(*err).unwrap_err();
+        }
+    }
+}
--- a/Release.toml
+++ b/Release.toml
@@ -163,4 +163,5 @@ version = "1.10.1"
     "migrate_v1.11.0_credential-providers.lz4",
     "migrate_v1.11.0_kubelet-tls-config.lz4",
     "migrate_v1.11.0_kubelet-new-config-files.lz4",
+    "migrate_v1.11.0_ecs-additional-configurations.lz4",
 ]
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -1368,6 +1368,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "ecs-additional-configurations"
+version = "0.1.0"
+dependencies = [
+ "migration-helpers",
+]
+
+[[package]]
 name = "ecs-settings-applier"
 version = "0.1.0"
 dependencies = [
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -33,6 +33,7 @@ members = [
     "api/migration/migrations/v1.11.0/credential-providers",
     "api/migration/migrations/v1.11.0/kubelet-tls-config",
     "api/migration/migrations/v1.11.0/kubelet-new-config-files",
+    "api/migration/migrations/v1.11.0/ecs-additional-configurations",
 
     "bottlerocket-release",
 
--- a/sources/api/ecs-settings-applier/src/ecs.rs
+++ b/sources/api/ecs-settings-applier/src/ecs.rs
@@ -62,6 +62,21 @@ struct ECSConfig {
 
     #[serde(rename = "GPUSupportEnabled")]
     gpu_support_enabled: bool,
+
+    #[serde(
+        rename = "TaskMetadataSteadyStateRate",
+        skip_serializing_if = "Option::is_none"
+    )]
+    metadata_service_rps: Option<i64>,
+
+    #[serde(
+        rename = "TaskMetadataBurstRate",
+        skip_serializing_if = "Option::is_none"
+    )]
+    metadata_service_burst: Option<i64>,
+
+    #[serde(rename = "ReservedMemory", skip_serializing_if = "Option::is_none")]
+    reserved_memory: Option<u16>,
 }
 
 // Returning a Result from main makes it print a Debug representation of the error, but with Snafu
@@ -129,6 +144,9 @@ async fn run() -> Result<()> {
         task_eni_enabled: true,
 
         gpu_support_enabled: cfg!(variant_flavor = "nvidia"),
+        reserved_memory: ecs.reserved_memory,
+        metadata_service_rps: ecs.metadata_service_rps,
+        metadata_service_burst: ecs.metadata_service_burst,
         ..Default::default()
     };
     if let Some(os) = settings.os {
--- a/sources/api/migration/migrations/v1.11.0/ecs-additional-configurations/Cargo.toml
+++ b/sources/api/migration/migrations/v1.11.0/ecs-additional-configurations/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "ecs-additional-configurations"
+version = "0.1.0"
+edition = "2018"
+authors = ["Arnaldo Garcia Rincon <agarrcia@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers", version = "0.1.0"}
--- a/sources/models/src/modeled_types/mod.rs
+++ b/sources/models/src/modeled_types/mod.rs
@@ -141,6 +141,9 @@ pub mod error {
         #[snafu(display("Invalid imageGCLowThresholdPercent '{}': {}", input, msg))]
         InvalidImageGCLowThresholdPercent { input: String, msg: String },
 
+        #[snafu(display("Invalid ECS duration value '{}'", input))]
+        InvalidECSDurationValue { input: String },
+
         #[snafu(display("Could not parse '{}' as an integer", input))]
         ParseInt {
             input: String,


sources/models/src/modeled_types/ecs.rs
```diff
@@ -291,3 +291,71 @@ mod test_ecs_agent_image_pull_behavior {
...
+
+/// ECSDurationValue represents a string that contains a valid ECS duration value
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ECSDurationValue {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref ECS_DURATION_VALUE: Regex = Regex::new(
+        r"^(([0-9]+\.)?[0-9]+h)?(([0-9]+\.)?[0-9]+m)?(([0-9]+\.)?[0-9]+s)?(([0-9]+\.)?[0-9]+ms)?$"
```
etungsten: The ECS documentation for `ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION` reads
> (Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", and "h".)

Should we try and cover "ns", "us/µs" as well?

webern (approved):

sources/models/src/modeled_types/ecs.rs
arnaldo2792: Actually, the ECS agent checks that both duration values are at least 1s: [container stop timeout] and [task cleanup]

[container stop timeout]: https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/parse.go#L61
[task cleanup]: https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/config.go#L340

So I think we can keep seconds as the minimum unit supported and even drop `ms`

arnaldo2792:
Forced push includes:

- Add documentation for the new settings
- Remove `ms` as a supported unit for ECS duration types, since the minimal value supported in the ECS agent is `1s`.

README.md
```diff
@@ -519,6 +521,11 @@ These settings can be changed at any time.
   Bottlerocket enables the `json-file`, `awslogs`, and `none` drivers by default.
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.metadata-service-steady-state-rate`: The steady state rate limit of the throttling configurations set for the task metadata service
+* `settings.ecs.metadata-service-burst-rate`: The burst rate limit of the throttling configurations set for the task metadata service
+* `settings.ecs.reserved-memory`: The amount of memory, in MB, reserved for critical system processes
+* `settings.ecs.task-cleanup-wait-duration`: Time to wait before the tasks resources are removed after they are stopped
+Valid time units are `s`, `m`, and `h`, i.e: `1h`, `1m1s`
```
etungsten: Missing punctuation. It becomes one line when rendered in markdown.
```suggestion
* `settings.ecs.task-cleanup-wait-duration`: Time to wait before the tasks resources are removed after they are stopped.
Valid time units are `s`, `m`, and `h`, i.e: `1h`, `1m1s`
```

arnaldo2792:
(Forced push to add missing `.` in documentation)

README.md
```diff
@@ -519,6 +521,11 @@ These settings can be changed at any time.
   Bottlerocket enables the `json-file`, `awslogs`, and `none` drivers by default.
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.metadata-service-steady-state-rate`: The steady state rate limit of the throttling configurations set for the task metadata service
+* `settings.ecs.metadata-service-burst-rate`: The burst rate limit of the throttling configurations set for the task metadata service
+* `settings.ecs.reserved-memory`: The amount of memory, in MB, reserved for critical system processes
+* `settings.ecs.task-cleanup-wait-duration`: Time to wait before the tasks resources are removed after they are stopped.
+Valid time units are `s`, `m`, and `h`, i.e: `1h`, `1m1s`
```
etungsten: Sorry, should have caught this in my earlier review. 
nit: `i.e.` should be `e.g.` in this context since you're listing examples and not rephrasing or elaborating the previous clause.

See https://www.merriam-webster.com/words-at-play/ie-vs-eg-abbreviation-meaning-usage-difference

sources/models/src/modeled_types/ecs.rs
```diff
@@ -291,3 +291,61 @@ mod test_ecs_agent_image_pull_behavior {
...
+
+/// ECSDurationValue represents a string that contains a valid ECS duration value
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ECSDurationValue {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref ECS_DURATION_VALUE: Regex =
+        Regex::new(r"^(([0-9]+\.)?[0-9]+h)?(([0-9]+\.)?[0-9]+m)?(([0-9]+\.)?[0-9]+s)?$").unwrap();
```
jpmcb: Curious how this regex was derived? If we're going off [the Go `time.ParseDuration` API](https://pkg.go.dev/time#ParseDuration) (where this duration string will be consumed):

> A duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".

This doesn't support negative time durations (like `-1h` with the leading sign) which probably makes sense for this use case but would be a valid duration to be consumed by the Go serialization. I think we should support it just in case customers are doing something with negative time duration (and if the ECS agent supports it)

---

[The Go source code](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/time/format.go;l=1523) does offer a regex we could reference:
```
[-+]?([0-9]*(\.[0-9]*)?[a-z]+)+
```
although this regex is more permissive of the unit types. Instead, [they reference a static in memory unit slice and valid units that way.](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/time/format.go;l=1506-1515)

jpmcb (commented):
Overall looks great - built an image off this patch and see the rendered environment variables under the ECS config file and in the json file

sources/models/src/modeled_types/ecs.rs
arnaldo2792: > . I think we should support it just in case customers are doing something with negative time duration (and if the ECS agent supports it)

Even thought the deserialization of negative values is supported in the golang API, the minimum duration value supported by the ECS agent is `1s`, hence the regex. 

> I think we should support it just in case customers are doing something with negative time duration (and if the ECS agent supports it)

I thought on supporting the same regex provided by the golang APIs, but decided against that because it seems odd to let a user set `-1h` in the API system, and then have the ECS agent ignore the value due to internal validations. I think the interface contract is between the API and the ECS agent, not the API with golang primitives, thus the API system should only allow what the ECS agent considers "valid" values, instead of what the golang APIs allow.

arnaldo2792:
(forced push replaces "i. e.:" with "e.g.")

etungsten (approved):

sources/models/src/modeled_types/ecs.rs
jpmcb: That makes total sense - thanks for the deep dive!! 👏

sources/models/src/lib.rs
```diff
@@ -237,6 +237,11 @@ struct ECSSettings {
     loglevel: ECSAgentLogLevel,
     enable_spot_instance_draining: bool,
     image_pull_behavior: ECSAgentImagePullBehavior,
+    container_stop_timeout: ECSDurationValue,
+    task_cleanup_wait_duration: ECSDurationValue,
+    metadata_service_steady_state_rate: i64,
+    metadata_service_burst_rate: i64,
```
bcressey: > Even though both values must be positive integers, I used i64 instead of u64 for the configurations because the [datastore serializer](https://github.com/bottlerocket-os/bottlerocket/blob/2a202f877fefbbfc54e689b164460ced6d66dacc/sources/api/datastore/src/serialization/pairs.rs#L172) doesn't support u64. I decided against of having a new PositiveIntegerValue custom type to perform validations at the API level for two reasons: a) I don't think we should have (at least not for this PR) a custom type to perform validations for positive integers, instead we should figure out if there is a way to support u64 serialization and b) the ECS agent will provide a [useful message](https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/config.go#L361) if any of the provided values isn't a positive integer.

I am pretty unhappy with this rationale and frankly just tired of accepting all the recent justifications along these lines. (See also #2463, #2257.)

Especially for the datastore serializer, `i64` seems like it would have the same problems as `u64` would for JSON. Is it actually problematic or does `serde_json` do the right thing in both cases?

I am actually fine with accepting negative values here if `ecs-agent` also tolerates them, which it appears to. What numeric type does it use? It's clearly a signed type. Can we just use that?

(Note that if it's actually an `i64`, then using a `u64` here would be an error that could expose the agent to invalid input.)

sources/models/src/modeled_types/ecs.rs
bcressey: The PR overview says this regex is the "effectively" the same as `KubernetesDurationValue`, but it doesn't accept `ms` values.

I've been leaning away from aggressive validation at the API level because the failure mode (bricked instance) is much worse than the consequence of letting the agent handle its own validation, and possibly failing to start or falling back to defaults.

Here I think the goal should be constraining the API settings values to the same range that the agent recognizes as legal values, even if they are not "accepted" values.

The API should filter out obvious jank like mismatched types, but going further than that imposes a steep usability cost that only makes sense if there are security concerns or something along those lines to motivate the extra validation.

README.md
```diff
@@ -519,6 +521,11 @@ These settings can be changed at any time.
   Bottlerocket enables the `json-file`, `awslogs`, and `none` drivers by default.
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.metadata-service-steady-state-rate`: The steady state rate limit of the throttling configurations set for the task metadata service
+* `settings.ecs.metadata-service-burst-rate`: The burst rate limit of the throttling configurations set for the task metadata service
```
bcressey: I don't feel very strongly about this, but if we are going to invent new settings, I'd prefer to name them in a way that's consistent across Bottlerocket.

For kubelet we have similar fields:
```
settings.kubernetes.event-burst
settings.kubernetes.event-qps
settings.kubernetes.kube-api-burst
settings.kubernetes.kube-api-qps
```

Here the equivalent might be settings like:
```
settings.ecs.metadata-service-rps
settings.ecs.metadata-service-burst
```

bcressey:
Here is some additional design guidance around API settings to provide more context.

The input value should match the output value to the extent possible. Meaning, don't take strings as input for numeric output, and don't take large unsigned integers for small signed integers. This is confusing for users and creates potential security risks.

Some types, like strings, need additional constraints (e.g. `SingleLineString`) to avoid creating problems when rendered in a template. This is a purely host level concern, meant to guard against rendering garbage that breaks the file structure, or having the value for one setting used as a backdoor to apply a completely different setting.

Beyond that, validation has two goals which are in tension.

If the agent (`kubelet`, `ecs-agent`) dies with a fatal error when given input of the correct type that does not match some other constraint - e.g. a string that does not match a regex, or a value that's outside a specific range - then ideally the API should reject that. Otherwise, a typo in an automated command applied fleet-wide could take down a large number of nodes at once. This is not an unbreakable rule - people could just not make typos - but it's reasonable to expect the API to have your back.

The drawback with this approach is that on first boot, nodes will fail to come up at all if given this kind of input. That is not a great experience. However, *at scale*, my bias is toward prioritizing the health of nodes that are in service over the usability of nodes that have never been in service.

On the other hand, if the agent does not die with an error when presented with input that doesn't match the constraint, then the preferred approach can be reversed. This is especially true if the agent logs a message saying that the setting is being ignored. The automated fleet-wide command will succeed, and may appear to give the desired results ("all nodes running with the new setting!"), but the logs will tell the tale.

The advantage of this approach is that the first boot experience is better. The node will come up as a functioning system, even if it is not configured exactly as expected. There's no risk to nodes in service, so it's OK to prefer usability over exactitude.

Globally, I'd also prefer some measure of consistency - e.g. using the same `KubernetesDurationValue` for two settings, one of which causes a fatal error and warrants the extra safeguards, and the other of which is ignored. This is related to another aspect of usability - whether the user can learn to successfully predict the behavior of the API - which hopefully makes for fewer frustrations and surprises over time.

With that in mind, this is why I was unhappy with the rationale you provided:
* We shouldn't use `u64` instead of `i64` just to enforce a positive number constraint, since that changes the input type and may lead to confusion or security risks.
* If we add a setting, we need to also add the correct validation to the setting to protect fleets at scale. This isn't supposed to be left for a follow-up PR. If there's an operational safety risk, it should be addressed when the setting is first added.
* However, if no additional validation is actually required for the agent to remain healthy, we shouldn't feel obligated to go beyond the minimum of matching up types.

sources/models/src/lib.rs
jpmcb: > I am actually fine with accepting negative values here if ecs-agent also tolerates them, which it appears to. What numeric type does it use? It's clearly a signed type. Can we just use that?

When the ECS agent loads configurations, it'll throw a warning if it finds negative values for either `TaskMetadataSteadyStateRate` or `TaskMetadataBurstRate`. Instead, [it'll use the defaults.](https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/config.go#L360-L364)

So yes, negative numbers are "valid" values in the sense that they can be deserialized into the Go struct, but the program logic will reject them.

> (Note that if it's actually an i64, then using a u64 here would be an error that could expose the agent to invalid input.)

Not sure if this is relevant, but [the ecs-agent defines both these values as a Go `int`](https://github.com/aws/amazon-ecs-agent/blob/5fb8e801e9d11630c3470abb778291913ebb9d7f/agent/config/types.go#L271-L275) which is signed and should be 64 bits wide on a x64 system (and 32 on a x32 system). So yes, a `u64` integer could overflow a signed `int` providing an invalid value

webern: > I am pretty unhappy with this rationale and frankly just tired of accepting all the recent justifications along these lines. (See also https://github.com/bottlerocket-os/bottlerocket/issues/2463, https://github.com/bottlerocket-os/bottlerocket/issues/2257.)

I think I can fix #2257, I added a comment there.

arnaldo2792: > Especially for the datastore serializer, i64 seems like it would have the same problems as u64 would for JSON. Is it actually problematic or does serde_json do the right thing in both cases?

The serializer only complains about `u64`:

```bash
bash-5.1# apiclient set ecs.metadata-service-burst-rate=1
Failed to change settings: Failed PATCH request to '/settings?tx=apiclient-set-cFwvZoTEUPHub2iq': Status 500 when PATCHing /settings?tx=apiclient-set-cFwvZoTEUPHub2iq: Error serializing Settings: 'u64' not allowed by Serializer
```

`i64` does work fine.

arnaldo2792:
Forced push includes:

- Fix supported values for duration type
- Update API settings names as suggested

arnaldo2792:
(Forced push to fix merge conflicts in `sources/Cargo.toml` and `Release.toml`)

README.md
```diff
@@ -511,6 +511,8 @@ These settings can be changed at any time.
 * `settings.ecs.allow-privileged-containers`: Whether launching privileged containers is allowed on the container instance.
   If this value is set to false, privileged containers are not permitted.
   Bottlerocket sets this value to false by default.
+* `settings.ecs.container-stop-timeout`: Time to wait for the task's container to stop on their own before they are forcefully stopped.
+Valid time units are `ns`, `µs`, `ms`, `s`, `m`, and `h`, e.g. `1h`, `1m1s`
```
bcressey: nit: if there are values that are too small for the `ecs-agent` to bother with, we could omit them:
```suggestion
Valid time units include `s`, `m`, and `h`, e.g. `1h`, `1m1s`
```

README.md
```diff
@@ -519,6 +521,11 @@ These settings can be changed at any time.
   Bottlerocket enables the `json-file`, `awslogs`, and `none` drivers by default.
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.metadata-service-rps`: The steady state rate limit of the throttling configurations set for the task metadata service
+* `settings.ecs.metadata-service-burst`: The burst rate limit of the throttling configurations set for the task metadata service
+* `settings.ecs.reserved-memory`: The amount of memory, in MB, reserved for critical system processes
```
bcressey: nit:
```suggestion
* `settings.ecs.reserved-memory`: The amount of memory, in MiB, reserved for critical system processes
```

packages/ecs-agent/ecs.config
```diff
@@ -16,3 +16,9 @@ ECS_ENGINE_AUTH_DATA='{
         {{~#if password~}},"password": "{{{password}}}"}{{/if}}
     {{~/each~}}}}'
 {{/if}}
+{{#if settings.ecs.container-stop-timeout}}
+ECS_CONTAINER_STOP_TIMEOUT="{{settings.ecs.container-stop-timeout}}"
+{{/if}}
+{{#if settings.ecs.task-cleanup-wait-duration}}
+ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION="{{settings.ecs.task-cleanup-wait-duration}}"
```
bcressey: nit: I think `task-stop-timeout` and `task-cleanup-wait` would be more consistent, naming-wise.

(dropping "duration" from `task-cleanup-wait` seems valuable since it's sort of implied that it's a Duration)

README.md
```diff
@@ -519,6 +521,11 @@ These settings can be changed at any time.
...
+* `settings.ecs.task-cleanup-wait-duration`: Time to wait before the tasks resources are removed after they are stopped.
```
bcressey: Do task resources include something other than containers?

README.md
```diff
@@ -511,6 +511,8 @@ These settings can be changed at any time.
 * `settings.ecs.allow-privileged-containers`: Whether launching privileged containers is allowed on the container instance.
   If this value is set to false, privileged containers are not permitted.
   Bottlerocket sets this value to false by default.
+* `settings.ecs.container-stop-timeout`: Time to wait for the task's container to stop on their own before they are forcefully stopped.
```
bcressey: nit:
```suggestion
* `settings.ecs.container-stop-timeout`: Time to wait for the task's containers to stop on their own before they are forcefully stopped.
```

README.md
```
ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION
Time to wait from when a task is stopped until the Docker container is removed.

ECS_CONTAINER_STOP_TIMEOUT
Time to wait from when a task is stopped before its containers are forcefully stopped
if they do not exit normally on their own.
```

What happens if task-cleanup-wait is shorter than container-stop-timeout ? Or does the cleanup timer not start running until the last container stops? The docs imply that both timeouts run concurrently, starting from the instant the task is stopped.

README.md
bcressey: It's really only possible to reserve 64 GiB of memory? On the one hand that seems like a lot, but on the other hand there are EC2 instance types with quite a lot more memory.

sources/api/migration/migrations/v1.11.0/ecs-additional-configurations/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added additional configurations for the ECS agent:
+///
+/// settings.ecs.container-stop-timeout
+/// settings.ecs.task-cleanup-wait-duration
+/// settings.ecs.metadata-service-rps
+/// settings.ecs.metadata-service-burst
+/// settings.ecs.reserved-memory
```
bcressey: nit: seems like the list in this comment just duplicates the code
```suggestion
/// We added additional configurations for the ECS agent.
```

sources/models/src/modeled_types/ecs.rs
```diff
@@ -291,3 +291,69 @@ mod test_ecs_agent_image_pull_behavior {
...
+
+/// ECSDurationValue represents a string that contains a valid ECS duration value
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct ECSDurationValue {
+    inner: String,
+}
+
+lazy_static! {
+    pub(crate) static ref ECS_DURATION_VALUE: Regex =
+        Regex::new(r"^(([0-9]+\.)?[0-9]+h)?(([0-9]+\.)?[0-9]+m)?(([0-9]+\.)?[0-9]+s)?(([0-9]+\.)?[0-9]+ms)?(([0-9]+\.)?[0-9]+(u|µ){1}s)?(([0-9]+\.)?[0-9]+ns)?$").unwrap();
```
bcressey: nit: regex101 called the `{1}` repetition "useless", and suggested using a non-capturing group for `(u|µ)` instead:
```suggestion
        Regex::new(r"^(([0-9]+\.)?[0-9]+h)?(([0-9]+\.)?[0-9]+m)?(([0-9]+\.)?[0-9]+s)?(([0-9]+\.)?[0-9]+ms)?(([0-9]+\.)?[0-9]+(?:u|µ)s)?(([0-9]+\.)?[0-9]+ns)?$").unwrap();
```

README.md
arnaldo2792: I'll change `resources` to `containers`, to prevent confusions.

arnaldo2792: I addressed this offline, but here are my findings:

`ContainerStopTimeout` is used when the ECS agent signals docker that it should stop a container. After this call, the ECS agent updates the state of the container to `ContainerStopped`. Once all the task's containers are `ContainerStopped` , the task transitions to `TaskStopped`. There is an infinite loop that checks if the task is in `TaskStopped`, if it is, the agent will begin the cleanup after `TaskCleanupDuration` have passed.

Thus, even if `ContainerStopTimeout` is lower than `TaskCleanupDuration` , no race conditions will occur since the cleanup will only happen after all the containers have stopped.

README.md
arnaldo2792: Well, the only thing that is setting "a limit" is the type of the field:

```golang
// ReservedMemory specifies Reduction, in MiB, of the memory capacity of the instance
// that is reported to Amazon ECS. Used by Amazon ECS when placing tasks on container instances.
// This doesn't reserve memory usage on the instance
ReservedMemory uint16
```

So yes, only 64Gib even in hosts with hundreds of Gib.

packages/ecs-agent/ecs.config
arnaldo2792: I think `task-stop-timeout` could be inaccurate, since this value is used when stopping a container in a task (and there could be plenty of them) not when stopping a task.

arnaldo2792:
Forced push includes:

- Rename `task-cleanup-wait-duration` to `task-cleanup-wait`
- Updated regular expression to validate duration values
- Fixed comments in documentation

arnaldo2792:
(Forced push to fix merge coonflicts)

README.md
```diff
@@ -541,6 +543,11 @@ These settings can be changed at any time.
   Bottlerocket enables the `json-file`, `awslogs`, and `none` drivers by default.
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.metadata-service-rps`: The steady state rate limit of the throttling configurations set for the task metadata service
+* `settings.ecs.metadata-service-burst`: The burst rate limit of the throttling configurations set for the task metadata service
```
jpmcb: For the `metadata-service-rps` and `metadata-service-burst` settings, I think we should add a small snippet of documentation commenting on the correlation between these two settings in the Bottlerocket API and the comma separated string for `ECS_TASK_METADATA_RPS_LIMIT` in the agent.

Even something like the following would suffice:

> "these two settings will be combined to generate the `ECS_TASK_METADATA_RPS_LIMIT` ECS agent setting"

I _don't_ think it'll be immediate clear to users that these two values are related to the `ECS_TASK_METADATA_RPS_LIMIT` agent setting. [The ECS agent documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html) _is_ clear what this setting does, but a quick scan of setting names doesn't mention anything about "burst" so I see this as a likely place of confusion for users without abit of extra documentation.

README.md
```diff
@@ -541,6 +543,11 @@ These settings can be changed at any time.
...
+* `settings.ecs.reserved-memory`: The amount of memory, in MiB, reserved for critical system processes
+* `settings.ecs.task-cleanup-wait`: Time to wait before the tasks containers are removed after they are stopped.
```
jpmcb: The name of this setting is confusing: which "cleanup" setting does it directly correlate to? Looking at the ECS agent documentation, there is `ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION`, `ECS_IMAGE_CLEANUP_INTERVAL`, `ECS_IMAGE_MINIMUM_CLEANUP_AGE`, etc.

Reading deeper into the ECS agent docs, it becomes clear this is correlated to the `ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION` setting.

It would make sense to me to rename this setting to
`settings.ecs.engine-task-cleanup-wait-duration` so there's an obvious 1:1 relation with it's ECS agent setting.

Again, this is in the spirit of being as clear to users as possible (and therefore provide a better UX)

README.md
jpmcb: I'm also wondering about the names here: both the names `metadata-service-rps` and `metadata-service-burst` don't strongly correlate to `ECS_TASK_METADATA_RPS_LIMIT`

What if we renamed
- `metadata-service-rps` to `task-metadata-rps-limit`
- `metadata-service-burst` to `task-metadata-burst-limit`

so it's clear these two relate to limits on each?

jpmcb:
On re-reading other comments before mine (https://github.com/bottlerocket-os/bottlerocket/pull/2527#discussion_r1013419040) looks like maybe I'm missing something when it comes to naming our API settings.

I don't feel strongly on this and I'm happy to go with whatever prior art there is on naming bottlerocket API endpoints. But I would like to call out the potential confusion users might have when attempting to correlate settings in software we consume (and then expose in settings via our API). In the past, I've preferred to keep these kind of things as close as possible to a 1:1 ratio to avoid confusion.

README.md
arnaldo2792: There is a clear distinction between `ECS_IMAGE_CLEANUP_INTERVAL`/`ECS_IMAGE_MINIMUM_CLEANUP_AGE` and `ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION`, the latter is used for tasks (ECS_TASK) whereas the others are used for images (ECS_IMAGE). The name of the new setting (`task-cleanup-wait`) is part of the ECS setting name (`TASK_CLEANUP_WAIT`), which is unique among the ECS agent's settings.

README.md
arnaldo2792: In https://github.com/bottlerocket-os/bottlerocket/pull/2527#discussion_r1007677221, it was decided to drop the `rate` part of the setting (which was similar to the suggestion you make with `limit`) to be consistent with other existing APIs.

I do think that I can make this setting a bit more "clear" by adding the prefix `task-`:

- `task-metadata-service-rps`
- `task-metadata-service-burst`

But if folks don't feel strongly about this, I can keep what I have :sweat_smile:

arnaldo2792:
Forced push include a note to clarify how the API settings map to the ECS provided settings, and a few missing dots in the documentation for the new settings.

README.md
```diff
@@ -541,6 +543,13 @@ These settings can be changed at any time.
   Bottlerocket enables the `json-file`, `awslogs`, and `none` drivers by default.
 * `settings.ecs.loglevel`: The level of verbosity for the ECS agent's logs.
   Supported values are `debug`, `info`, `warn`, `error`, and `crit`, and the default is `info`.
+* `settings.ecs.metadata-service-rps`: The steady state rate limit of the throttling configurations set for the task metadata service.
+* `settings.ecs.metadata-service-burst`: The burst rate limit of the throttling configurations set for the task metadata service.
+* `settings.ecs.reserved-memory`: The amount of memory, in MiB, reserved for critical system processes.
+* `settings.ecs.task-cleanup-wait`: Time to wait before the tasks containers are removed after they are stopped.
```
```suggestion
* `settings.ecs.task-cleanup-wait`: Time to wait before the task's containers are removed after they are stopped.
```

README.md
bcressey: I would prefer `task-metadata-service-*`, but I don't feel too strongly about it.

bcressey (approved):

arnaldo2792:
(forced push to fix documentation's typo)

etungsten (approved):

PR #2532, opened 2022-10-27
Add ShellCheck to `check-lints` (for tools)

jpculp:
**Description of changes:**

During the `check-lints` stage of the build, **ShellCheck** will analyze any bash scripts under the `tools` directory and raise an error for any findings with a severity of `warning` or above.

For more targeted linting, `check-lints` was also split into `check-clippy` and `check-shell` which can be called individually.

Also addressed several warnings to avoid breaking the build.

**Testing done:**

Ran with and without the `tools: fix ShellCheck warnings` commit.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -332,7 +332,12 @@ done
 
 # A top level target for devs to ensure review and patch readiness
 [tasks.check]
-dependencies = ["check-cargo-version", "unit-tests", "check-fmt", "check-lints"]
+dependencies = [
+   "check-cargo-version",
+   "unit-tests",
+   "check-fmt",
+   "check-lints",
+   ]
 
 [tasks.check-fmt]
 script = [
@@ -382,6 +387,12 @@ fi
 ]
 
 [tasks.check-lints]
+dependencies = [
+   "check-clippy",
+   "check-shell",
+]
+
+[tasks.check-clippy]
 script = [
 '''
 rc=0
@@ -397,6 +408,44 @@ fi
 
 if [ "${rc}" -ne 0 ]; then
   echo "Found lint warnings. First-party source code is checked with clippy." >&2
+  exit $rc
+fi
+'''
+]
+
+[tasks.check-shell]
+script = [
+'''
+rc=0
+
+# For bash first-party shell code
+if ! docker run --rm \
+  --network=none \
+  --user "$(id -u):$(id -g)" \
+  --security-opt label:disable \
+  -v "${BUILDSYS_TOOLS_DIR}":/tmp/tools \
+  "${BUILDSYS_SDK_IMAGE}" \
+  bash -c \
+    'flagged_scripts=0 && \
+    cd /tmp/tools && \
+    for shell_script in $(grep -l --directories=skip --regexp="^#\!.*bash$" * ); do
+      if ! shellcheck \
+        --external-sources \
+        --source-path=SCRIPTDIR \
+        --format=gcc \
+        --severity=warning \
+        "${shell_script}"; then
+        ((++flagged_scripts))
+      fi
+    done && \
+    if [ "${flagged_scripts}" -ne 0 ]; then
+      exit 1
+    fi'; then
+  rc=1
+fi
+
+if [ "${rc}" -ne 0 ]; then
+  echo "Found lint warnings. First-party shell code is checked with ShellCheck." >&2
   exit $rc
 fi
 '''
--- a/tools/docker-go
+++ b/tools/docker-go
@@ -30,6 +30,7 @@ required_arg() {
    fi
 }
 
+# shellcheck disable=SC2124  # TODO: improve command interface (#2534)
 parse_args() {
   while [ ${#} -gt 0 ] ; do
     case "${1}" in
--- a/tools/start-local-vm
+++ b/tools/start-local-vm
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# shellcheck disable=SC2054  # Arrays are formatted for passing args to other tools
 
 shopt -s nullglob
 
@@ -14,12 +15,16 @@ current_images=()
 boot_image=
 data_image=
 
-readonly repo_root=$(git rev-parse --show-toplevel)
-
 bail() {
     >&2 echo "$@"
     exit 1
 }
+
+if ! git_toplevel=$(git rev-parse --show-toplevel); then
+    bail "Failed to get the root of the repo."
+else
+    readonly repo_root="${git_toplevel}"
+fi
 
 show_usage() {
     echo "\
@@ -116,8 +121,9 @@ parse_args() {
     [[ -n ${arch} ]] || usage_error 'Architecture needs to be set via either --arch or BUILDSYS_ARCH.'
     [[ -n ${variant} ]] || usage_error 'Variant needs to be set via either --variant or BUILDSYS_VARIANT.'
 
-    local host_arch=$(uname -m)
-    [[ ${arch} = ${host_arch} ]] || bail "Architecture needs to match host architecture (${host_arch}) for hardware virtualization."
+    declare -l host_arch
+    host_arch=$(uname -m)
+    [[ ${arch} == "${host_arch}" ]] || bail "Architecture needs to match host architecture (${host_arch}) for hardware virtualization."
 
     for path in "${!extra_files[@]}"; do
         [[ -e ${path} ]] || bail "Cannot find local file '${path}' to inject."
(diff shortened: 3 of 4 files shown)


Makefile.toml
```diff
@@ -395,8 +395,28 @@ fi
 
 # TODO: fix clippy lints in "sources" and check that too
 
+# For bash first-party source code
```
stmcginnis: Can we add this under a `tasks.check-shell` or something? Then we could potentially clean up our GitHub Action jobs to only trigger the Rust linting on `*.rs` file changes and ShellCheck on shell file changes.

Then of course also add this new task to `tasks.check` so the full suite can be easily run.

tools/start-local-vm
```diff
@@ -116,8 +121,9 @@ parse_args() {
     [[ -n ${arch} ]] || usage_error 'Architecture needs to be set via either --arch or BUILDSYS_ARCH.'
     [[ -n ${variant} ]] || usage_error 'Variant needs to be set via either --variant or BUILDSYS_VARIANT.'
 
+    # shellcheck disable=SC2155  # This is safe even with a masked return code
```
bcressey: nit: hypothetically, if `uname` failed, the return code would be masked by the local assignment. IMO it's better practice to just always declare locals first before any assignment.

Makefile.toml
```diff
@@ -395,8 +395,28 @@ fi
...
+  bash -c \
+    'cd /tmp/tools && \
+    for shell_script in $(grep -l --directories=skip --regexp="^#!.*bash$" * ); do \
+      shellcheck \
+        --external-sources \
+        --source-path=SCRIPTDIR \
+        --format=gcc \
+        --severity=warning \
+        "${shell_script}"
+    done'; then
```
bcressey: Isn't the return code from the `for` loop the result of the last `shellcheck` invocation?

I'd prefer writing this to look more like `tasks.check-fmt` or `tasks.unit-tests`, where each script is passed to a function that invokes `shellcheck` via `docker run` on just the one script argument.

markusboehme: A downside of the "container per script file" approach is speed: Setting up and tearing down a container with the SDK image takes about 2 seconds on my development machine. The exit codes of the individual shellcheck invocations will need to be properly aggregated of course, but is there another reason why you'd prefer to only check one script per container?

tools/start-local-vm
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# shellcheck disable=SC2054  # Arrays are formatted for passing args to other tools
```
markusboehme, line 2: That shellcheck linter probably assumes the script is doing something dumb while assembling the QEMU invocation, where QEMU just needs commas in individual arguments. We could also quote those, but I think disabling the warning is the better choice for readability. :+1:

tools/start-local-vm
```diff
@@ -14,13 +15,17 @@ current_images=()
...
 bail() {
     >&2 echo "$@"
     exit 1
 }
 
+if ! git_toplevel=$(git rev-parse --show-toplevel); then
+ bail "Failed to get the root of the repo."
+else
+  readonly repo_root="${git_toplevel}"
+fi
```
markusboehme: Nit picking: The indentation of the two branches differs (one vs two spaces), and also differs from the rest of the script (four spaces). I recognize there's some diversity in styles among the tools scripts already, but unless/until somebody proposes to normalize it all (and keep it normalized), I'd prefer to at least keep the consistency within individual scripts.

tools/docker-go
```diff
@@ -37,7 +37,7 @@ parse_args() {
         --module-path ) shift; GO_MODULE_PATH="${1}" ;;
         --sdk-image ) shift; SDK_IMAGE="${1}" ;;
         --go-mod-cache ) shift; GO_MOD_CACHE="${1}" ;;
-        --command ) shift; COMMAND="${@:1}" ;;
+        --command ) shift; COMMAND="${*:1}" ;;
```
markusboehme: This makes the warning go away and retains current behavior, but the warning hints at an actual problem: The only safe way to use `docker-go` is to pass a single argument to `--command`. Anything else would lead to unexpected behavior, for example `docker-go --command foo bar baz` is indistinguishable from `docker-go --command foo 'bar baz'` and `docker-go --command foo --module-path bar` will both pass `--module-path bar` to `foo`, and actually modify the script's `GO_MODULE_PATH` variable.

Is it feasible to change the script's interface without too many coordinated changes? A safer interface could interpret anything after `--` as the command to execute, and keep all following arguments in an array (`COMMAND=( "$@" )`, later: `"${COMMAND[@]}"`) to retain fidelity of the originally passed-in arguments.

If that's too much for this PR or the script can't be changed for other reasons, disabling the warning could at least serve as a sign calling attention to the unexpected behavior.

Makefile.toml
bcressey: No, consistency with the other approaches was the only reason. If it's causing performance issues, I agree it's worth the cost to avoid it.

tools/docker-go
jpculp: That is a very good point. I wanted to keep the changes in this PR to the bare minimum to avoid introducing too many changes at once, but I went ahead and created issue #2534 so that we can follow-up on this.

tools/start-local-vm
jpculp: I am not sure how I missed this, but I'm definitely in agreement.

jpculp:
- Added improved logic for counting the scripts that contain warnings.
- Expanded `check-lints` into `check-clippy` and `check-shell` so that linting can be more targeted in the future.
- Fixed indentation inconsistency.
- Declare and assign `arch` separately.
- Replaced fix in `docker-go` with a TODO to address in https://github.com/bottlerocket-os/twoliter/issues/492.

jpculp:
Fixed a typo in one of the commit messages.

tools/docker-go
markusboehme: Makes sense, it's best to keep the PR focused. Thanks!

markusboehme (approved):

stmcginnis (approved):

Makefile.toml
```diff
@@ -402,6 +413,44 @@ fi
...
+if ! docker run --rm \
+  --network=none \
+  --user "$(id -u):$(id -g)" \
+  --security-opt label:disable \
+  -v "${BUILDSYS_TOOLS_DIR}":/tmp/tools \
+  "${BUILDSYS_SDK_IMAGE}" \
+  bash -c \
+    'flagged_scripts=0 && \
+    cd /tmp/tools && \
+    for shell_script in $(grep -l --directories=skip --regexp="^#!.*bash$" * ); do
```
bcressey:
```suggestion
    for shell_script in $(grep -l --directories=skip --regexp='^#!.*bash$' * ); do
```

Makefile.toml
```diff
@@ -402,6 +413,44 @@ fi
...
+        --source-path=SCRIPTDIR \
+        --format=gcc \
+        --severity=warning \
+        "${shell_script}"; then
+        ((++flagged_scripts))
+      fi
+    done && \
+    if [ ${flagged_scripts} -ne 0 ]; then
+      exit 1
+    fi'; then
```
bcressey: nit: since I am in the "quote and brace everything" school

```suggestion
    if [ "${flagged_scripts}" -ne 0 ]; then
      exit 1
    fi'; then
```

Makefile.toml
```diff
@@ -402,6 +413,44 @@ fi
...
+  rc=1
+fi
+
+if [ ${rc} -ne 0 ]; then
+  echo "Found lint warnings. First-party shell code is checked with ShellCheck." >&2
+  exit $rc
+fi
```
bcressey:
```suggestion
if [ "${rc}" -ne 0 ]; then
  echo "Found lint warnings. First-party shell code is checked with ShellCheck." >&2
  exit "${rc}"
fi
```

bcressey (approved):
Some minor nits, but nothing blocking.

Makefile.toml
jpculp: Normally I would use single quotes, but currently the whole block is wrapped in single quotes since it's being passed through `bash -c`.

Makefile.toml
jpculp: I tend to do "quote and brace everything" UNLESS the variable value is an integer or a variable name is single character, but I'll go ahead and make this change since I have to make another revision already.

Makefile.toml
jpculp: After further investigation, it seems in some environments the `!` can cause issues if you copy/paste into a terminal. I'll go ahead and escape it.

jpculp:
- Escaped `!` in `grep` regex.
- Added quotes to be more consistent with `check-clippy`.

etungsten (approved):

jpculp:
I rebased locally and ran `cargo make check-lints` and there haven't been any new warnings.

bcressey (approved):

PR #2536, opened 2022-10-29
feature: Enable configuration of Kubelet TLS certs

### LINKED ISSUE #2481: Configure kubelet parameters via BR API
**What I'd like:**

Our team needs to change some of the kubelet parameters (in our case we change `-v --client-ca-file, --tls-cert-file, --tls-key-file`). We would like to have these parameters exposed to the API, ideally all of the [flags](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/) can be changed.

To unblock our usecase we have created privileged host container that changes KubeletConfig file on the filesystem.

stmcginnis:
**Issue number:**

Closes #2481 

**Description of changes:**

This enables the ability to provide a TLS public and private key to be used by the kubelet process for HTTPS communication.
It adds `settings.kubernetes.server-certificate` and `settings.kubernetes.server-key`.

This corresponds to the `--tls-cert-file` and `--tls-key-file` arguments (or the `tlsCertFile` and `tlsPrivateKeyFile` config settings).

**Testing done:**

Generated an x509 public and private pair.
Used sample eksctl yaml and added the two new settings with the base64 contents of the generated certs.
Deployed cluster and verified creation successful.
Checked contents of generated `/etc/kubernetes/kubelet/config` file were correct.
`cat`'d contents of the two certificate files written out and made sure they matched my local copies of them.
Checked `systemctl status kubelet` and verified service was running and happy.
Checked `journalctl -u kubelet` to check if there were any errors.
Checked output from `kubectl get pods -A` and `kubectl get nodes` to make sure cluster was functional.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,6 +476,8 @@ The following settings are optional and allow you to further configure your clus
   ephemeral-storage= "1Gi"
   ```
 
+* `settings.kubernetes.server-certificate`: The base64 encoded content of an x509 certificate for the Kubelet web server, which is used for retrieving logs and executing commands.
+* `settings.kubernetes.server-key`: The base64 encoded content of an x509 private key for the Kubelet web server.
 * `settings.kubernetes.topology-manager-policy`: Specifies the topology manager policy. Possible values are `none`, `restricted`, `best-effort`, and `single-numa-node`. Defaults to `none`.
 * `settings.kubernetes.topology-manager-scope`: Specifies the topology manager scope. Possible values are `container` and `pod`. Defaults to `container`. If you want to group all containers in a pod to a common set of NUMA nodes, you can set this setting to `pod`.
 
--- a/Release.toml
+++ b/Release.toml
@@ -161,4 +161,6 @@ version = "1.10.1"
     "migrate_v1.11.0_aws-creds.lz4",
     "migrate_v1.11.0_aws-creds-metadata.lz4",
     "migrate_v1.11.0_credential-providers.lz4",
+    "migrate_v1.11.0_kubelet-tls-config.lz4",
+    "migrate_v1.11.0_kubelet-new-config-files.lz4",
 ]
--- a/packages/kubernetes-1.21/kubelet-config
+++ b/packages/kubernetes-1.21/kubelet-config
@@ -113,7 +113,12 @@ featureGates:
   CSIMigration: false
 protectKernelDefaults: true
 serializeImagePulls: false
+{{#if (and (default "" settings.kubernetes.server-certificate) (default "" settings.kubernetes.server-key))}}
+tlsCertFile: "/etc/kubernetes/pki/kubelet-server.crt"
+tlsPrivateKeyFile: "/etc/kubernetes/pki/kubelet-server.key"
+{{else}}
 serverTLSBootstrap: {{settings.kubernetes.server-tls-bootstrap}}
+{{/if}}
 tlsCipherSuites:
 - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
 volumePluginDir: "/var/lib/kubelet/plugins/volume/exec"
--- a/packages/kubernetes-1.21/kubernetes-1.21.spec
+++ b/packages/kubernetes-1.21/kubernetes-1.21.spec
@@ -35,6 +35,9 @@ Source7: kubelet-bootstrap-kubeconfig
 Source8: kubernetes-tmpfiles.conf
 Source9: kubelet-sysctl.conf
 Source10: prepare-var-lib-kubelet.service
+Source11: kubelet-server-crt
+Source12: kubelet-server-key
+Source13: etc-kubernetes-pki.mount
 
 # ExecStartPre drop-ins
 Source20: prestart-pull-pause-ctr.conf
@@ -90,7 +93,7 @@ install -d %{buildroot}%{_cross_bindir}
 install -p -m 0755 ${output}/kubelet %{buildroot}%{_cross_bindir}
 
 install -d %{buildroot}%{_cross_unitdir}
-install -p -m 0644 %{S:1} %{S:10} %{buildroot}%{_cross_unitdir}
+install -p -m 0644 %{S:1} %{S:10} %{S:13} %{buildroot}%{_cross_unitdir}
 
 install -d %{buildroot}%{_cross_unitdir}/kubelet.service.d
 install -p -m 0644 %{S:20} %{S:21} %{S:22} %{buildroot}%{_cross_unitdir}/kubelet.service.d
@@ -102,6 +105,8 @@ install -m 0644 %{S:4} %{buildroot}%{_cross_templatedir}/kubelet-kubeconfig
 install -m 0644 %{S:5} %{buildroot}%{_cross_templatedir}/kubernetes-ca-crt
 install -m 0644 %{S:6} %{buildroot}%{_cross_templatedir}/kubelet-exec-start-conf
 install -m 0644 %{S:7} %{buildroot}%{_cross_templatedir}/kubelet-bootstrap-kubeconfig
+install -m 0644 %{S:11} %{buildroot}%{_cross_templatedir}/kubelet-server-crt
+install -m 0644 %{S:12} %{buildroot}%{_cross_templatedir}/kubelet-server-key
 
 install -d %{buildroot}%{_cross_tmpfilesdir}
 install -p -m 0644 %{S:8} %{buildroot}%{_cross_tmpfilesdir}/kubernetes.conf
@@ -123,6 +128,7 @@ ln -rs \
 %{_cross_bindir}/kubelet
 %{_cross_unitdir}/kubelet.service
 %{_cross_unitdir}/prepare-var-lib-kubelet.service
+%{_cross_unitdir}/etc-kubernetes-pki.mount
 %dir %{_cross_unitdir}/kubelet.service.d
 %{_cross_unitdir}/kubelet.service.d/prestart-pull-pause-ctr.conf
 %{_cross_unitdir}/kubelet.service.d/make-kubelet-dirs.conf
@@ -134,6 +140,8 @@ ln -rs \
 %{_cross_templatedir}/kubelet-bootstrap-kubeconfig
 %{_cross_templatedir}/kubelet-exec-start-conf
 %{_cross_templatedir}/kubernetes-ca-crt
+%{_cross_templatedir}/kubelet-server-crt
+%{_cross_templatedir}/kubelet-server-key
 %{_cross_tmpfilesdir}/kubernetes.conf
 %{_cross_sysctldir}/90-kubelet.conf
 %dir %{_cross_libexecdir}/kubernetes
--- a/packages/kubernetes-1.23/kubernetes-1.23.spec
+++ b/packages/kubernetes-1.23/kubernetes-1.23.spec
@@ -35,6 +35,9 @@ Source7: kubelet-bootstrap-kubeconfig
 Source8: kubernetes-tmpfiles.conf
 Source9: kubelet-sysctl.conf
 Source10: prepare-var-lib-kubelet.service
+Source11: kubelet-server-crt
+Source12: kubelet-server-key
+Source13: etc-kubernetes-pki.mount
 
 # ExecStartPre drop-ins
 Source20: prestart-pull-pause-ctr.conf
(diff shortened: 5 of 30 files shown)


Release.toml
```diff
@@ -156,3 +156,6 @@ version = "1.10.1"
     "migrate_v1.10.1_container-runtime.lz4",
     "migrate_v1.10.1_container-runtime-metadata.lz4"
 ]
+"(1.10.1, 1.11.0)" = [
+    "migrate_v1.11.0_kubelet-tls-config.lz4",
+]
```
etungsten: Missing newline at the end of this file.

packages/kubernetes-1.24/kubelet-config
```diff
@@ -118,6 +118,10 @@ serializeImagePulls: false
 serverTLSBootstrap: {{settings.kubernetes.server-tls-bootstrap}}
 tlsCipherSuites:
 - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+{{#if (and settings.kubernetes.tls-certificate settings.kubernetes.tls-private-certificate)}}
+tlsCertFile: "/etc/kubernetes/pki/tls-cert.pub"
+tlsPrivateKeyFile: "/etc/kubernetes/pki/tls-cert"
+{{/if}}
```
etungsten: Does it make sense to evaluate these settings individually rather than together at once? 

`kubelet` doesn't seem to care if one or the other is missing but here we're enforcing both settings to be set for anything to be written to the kubelet config. Maybe we just render the configuration if the setting exist even if won't do anything without the other being specified?

sources/api/migration/migrations/v1.11.0/kubelet-tls-config/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2018"
+authors = ["Sean McGinnis <stmcg@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers", version = "0.1.0" }
```
etungsten: Missing newline at the end of file.

etungsten (changes requested):
We're missing a migration for the newly added settings. We need to migrate with the `AddSettingsMigration` helper.

packages/kubernetes-1.24/kubelet-config
stmcginnis: It might not error, but it will not use the provided files unless BOTH are provided. So it doesn't do us any good to add one without the other. You can't use the public key without the private one, and the private one will not be used to generate a public key, so it doesn't provide any benefit on its own.

There's not a lot of descriptive documentation for this, but the comments in kubelet do say this (emphasis mine):

> If tlsCertFile **AND** tlsPrivateKeyFile are not provided, a self-signed certificate and key are generated

packages/kubernetes-1.24/kubernetes-tls-crt
```diff
@@ -0,0 +1,3 @@
+{{~#if settings.kubernetes.tls-certificate~}}
+{{base64_decode settings.kubernetes.tls-certificate}}
+{{~/if~}}
```
etungsten: non-blocking: I think the conditional is somewhat redundant since this file will not be rendered unless the setting got set. Same for tls-private-crt

etungsten (approved):
:boat:

packages/kubernetes-1.24/kubernetes-tls-crt
etungsten: Ah nevermind, the `configuration-file` metadata applies for the entire `settings.kubernetes`

README.md
```diff
@@ -456,6 +456,8 @@ The following settings are optional and allow you to further configure your clus
   ephemeral-storage= "1Gi"
   ```
 
+* `settings.kubernetes.tls-certificate`: The base64 encoded content of an x509 certificate for Kubelet HTTPS communication.
+* `settings.kubernetes.tls-private-certificate`: The base64 encoded content of an x509 private key corresponding to the `tls-certificate` public key.
```
bcressey: Let's try to maintain a really clear distinction between certificate (always public) and key (always private).
```suggestion
* `settings.kubernetes.tls-private-key`: The base64 encoded content of an x509 private key corresponding to the `tls-certificate` certificate.
```

Some of this is coming out of the Secure Boot experience, where "key" is frequently used to mean "one or more public x.509 certificates", in contrast to its conventional usage.

packages/kubernetes-1.21/kubelet-config
```diff
@@ -116,6 +116,10 @@ serializeImagePulls: false
 serverTLSBootstrap: {{settings.kubernetes.server-tls-bootstrap}}
 tlsCipherSuites:
 - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+{{#if (and settings.kubernetes.tls-certificate settings.kubernetes.tls-private-certificate)}}
+tlsCertFile: "/etc/kubernetes/pki/tls-cert.pub"
+tlsPrivateKeyFile: "/etc/kubernetes/pki/tls-cert"
+{{/if}}
```
bcressey: Need to make `/etc/kubernetes/pki` a tmpfs mount and give it a `secret_t` SELinux label, since we could end up writing secrets here (like your `root-.aws.mount` elsewhere).

packages/kubernetes-1.21/kubelet-config
```diff
@@ -116,6 +116,10 @@ serializeImagePulls: false
 serverTLSBootstrap: {{settings.kubernetes.server-tls-bootstrap}}
```
bcressey, line 120: We may want to only do `serverTLSBootstrap` if we don't have the cert + private key, even if it's set to "true", because otherwise the bootstrap process will either do nothing (?) or else bootstrap a `/var/lib/kubelet/pki/kubelet-server-current.pem` certificate that (confusingly) isn't used.

sources/api/migration/migrations/v1.11.0/kubelet-tls-config-files/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+            "kubelet-env",
+            "kubelet-config",
+            "kubelet-kubeconfig",
+            "kubelet-bootstrap-kubeconfig",
+            "kubelet-exec-start-conf",
+            "kubernetes-ca-crt",
+            "proxy-env",
+            "kubernetes-tls-crt",
+            "kubernetes-tls-private-crt",
+        ],
```
bcressey: Just a reminder that we'll somehow need to sync up this list with #2377, which adds "credential-provider-config-yaml" as well, or else ensure that the migrations run in a specific order.

(Maybe rename this to a shared migration called `kubelet-new-config-files`.)

packages/kubernetes-1.23/kubernetes-1.23.spec
```diff
@@ -99,6 +101,8 @@ install -m 0644 %{S:4} %{buildroot}%{_cross_templatedir}/kubelet-kubeconfig
 install -m 0644 %{S:5} %{buildroot}%{_cross_templatedir}/kubernetes-ca-crt
 install -m 0644 %{S:6} %{buildroot}%{_cross_templatedir}/kubelet-exec-start-conf
 install -m 0644 %{S:7} %{buildroot}%{_cross_templatedir}/kubelet-bootstrap-kubeconfig
+install -m 0644 %{S:11} %{buildroot}%{_cross_templatedir}/kubernetes-tls-crt
+install -m 0644 %{S:12} %{buildroot}%{_cross_templatedir}/kubernetes-tls-private-crt
```
bcressey: You might also consider rendering both the certificate and the key to a single templated file, e.g. `/etc/kubernetes/pki/kubelet-server-current.pem`, the way that the combined client / server files are today under `/var/lib/kubelet/pki`.

packages/kubernetes-1.21/kubelet-config
bcressey: My only misgiving about giving this the `secret_t` label is whether pods might expect to be able to mount in `/etc/kubernetes/pki/ca.crt`, or the directory, for some reason. With the new label, only privileged pods could read the file. Maybe GitHub search would tell us if that's likely to be a problem.

stmcginnis: We have other files that are already written here. Would that cause upgrade issues if we now switched that to a tmpfs mount?

stmcginnis: I think (🤞) we might be safe: https://github.com/search?q=%2Fetc%2Fkubernetes%2Fpki%2Fca.crt+filename%3ADockerfile&type=Code&ref=advsearch&l=&l=

packages/kubernetes-1.21/kubelet-config
```diff
@@ -113,7 +113,12 @@ featureGates:
   CSIMigration: false
 protectKernelDefaults: true
 serializeImagePulls: false
+{{#if (and (default "" settings.kubernetes.tls-certificate) (default "" settings.kubernetes.tls-private-key))}}
```
stmcginnis: I'm not too happy about this, but without the `default`s in there the rendering of the file blows up with a handlebars error:

```
helper: Couldn't read parameter x
```

If anyone knows of a better way, please let me know.

packages/kubernetes-1.21/kubelet-config
```diff
@@ -113,7 +113,12 @@ featureGates:
...
+tlsCertFile: "/etc/kubernetes/pki/tls-cert.pub"
+tlsPrivateKeyFile: "/etc/kubernetes/pki/tls-cert"
```
bcressey: I'd prefer to name these files something that better indicates their purpose:
```suggestion
tlsCertFile: "/etc/kubernetes/pki/kubelet-server.key"
tlsPrivateKeyFile: "/etc/kubernetes/pki/kubelet-server.crt"
```

packages/kubernetes-1.21/kubernetes-1.21.spec
```diff
@@ -35,6 +35,9 @@ Source7: kubelet-bootstrap-kubeconfig
 Source8: kubernetes-tmpfiles.conf
 Source9: kubelet-sysctl.conf
 Source10: prepare-var-lib-kubelet.service
+Source11: kubernetes-tls-crt
+Source12: kubernetes-tls-private-key
```
bcressey: In line with the output filename suggestions:
```suggestion
Source11: kubelet-server-crt
Source12: kubelet-server-key
```

README.md
```diff
@@ -456,6 +456,8 @@ The following settings are optional and allow you to further configure your clus
   ephemeral-storage= "1Gi"
   ```
 
+* `settings.kubernetes.tls-certificate`: The base64 encoded content of an x509 certificate for Kubelet HTTPS communication.
+* `settings.kubernetes.tls-private-key`: The base64 encoded content of an x509 private key corresponding to the `tls-certificate` public certificate.
```
bcressey: I've said in the past that `settings.kubernetes` is misnamed, and should really be `settings.kubelet`.

If we pretend for a moment that the renaming has already happened, and that the "path" to this setting implicitly mentions `kubelet`, then I would prefer that these settings better match their function:
```suggestion
* `settings.kubernetes.server-certificate`: The base64 encoded content of an x509 certificate for the Kubelet web server, which is used for retrieving logs and executing commands.
* `settings.kubernetes.server-key`: The base64 encoded content of an x509 private key for the Kubelet web server.
```

stmcginnis:
Rebased on `develop` to resolve merge conflicts.

packages/kubernetes-1.21/kubelet-config
bcressey: You could also do nested `unless`, e.g.
```suggestion
{{#unless settings.kubernetes.tls-certificate}}
{{#unless settings.kubernetes.tls-private-key}}
serverTLSBootstrap: {{settings.kubernetes.server-tls-bootstrap}}
{{/unless}}
{{/unless}}
```

I wouldn't call it "better", just "different" - and reminiscent of C `ifdef` nesting, which is probably why it comes to mind.

stmcginnis: We would still need the `if` conditional to determine whether to write out of the `tlsCert` and `tlsPrivateKeyFile` though, right? So the double `unless` would work to determine if we need to write out `serverTLSBootstrap`, but we also need to know when to do the inverse.

sources/models/src/lib.rs
```diff
@@ -207,6 +207,8 @@ struct KubernetesSettings {
     image_gc_low_threshold_percent: ImageGCLowThresholdPercent,
     provider_id: Url,
     log_level: u8,
+    tls_certificate: ValidBase64,
+    tls_private_key: ValidBase64,
```
bcressey: This should be `server_certificate` and `server_key` correct?

stmcginnis: Yep, I missed that rename. Fixed in update. Also rebased and addressed merged conflicts.

bcressey (approved):

etungsten (approved):

PR #2537, opened 2022-10-31
Allow bootstrap containers to manage swap

### LINKED ISSUE #1911: Support Swap
**What I'd like:**
Support enabling swap via Linux commands like fallocate, mkswap, and swapon. Using swap allows the container to write excess memory requirements to disk when the container has exhausted all the RAM that is available to it.

Have submitted an issue on aws-cdk to enable swap through aws-cdk: https://github.com/aws/aws-cdk/issues/18460

stefansundin:
**Issue number:**

~~Closes~~ Provides a workaround for #1911

I don't really want to close #1911 because I think configuring swap should be possible without a bootstrap container. It should be as simple as configuring one or more `[[swap]]` section with a location and size. Doing that would probably require packaging `mkswap` and `swapon` though, and that is currently not something I know how to do.

**Description of changes:**

Allows bootstrap containers to manage swap by allowing the `swapon` and `swapoff` syscalls.

**Testing done:**

I built my own ami with this change and tested that a bootstrap container is able to set up swap. I created [a docker image that just runs whatever you give it as user data](https://github.com/stefansundin/bottlerocket-bootstrap-exec-user-data), to make it easy to change the script.

This is my script:

```shell
#!/bin/bash -ex
if [[ ! -f /.bottlerocket/rootfs/var/swapfile ]]; then
  dd if=/dev/zero of=/.bottlerocket/rootfs/var/swapfile bs=1M count=300
  chmod 600 /.bottlerocket/rootfs/var/swapfile
  mkswap /.bottlerocket/rootfs/var/swapfile
fi
swapon /.bottlerocket/rootfs/var/swapfile
```

In the OS configuration I run it with:

```toml
[settings.bootstrap-containers.swap]
source = "public.ecr.aws/stefansundin/bottlerocket-bootstrap-exec-user-data:latest"
mode = "always"
essential = false
user-data = "IyEvYmluL2Jhc2ggLWV4CmlmIFtbICEgLWYgLy5ib3R0bGVyb2NrZXQvcm9vdGZzL3Zhci9zd2FwZmlsZSBdXTsgdGhlbgogIGRkIGlmPS9kZXYvemVybyBvZj0vLmJvdHRsZXJvY2tldC9yb290ZnMvdmFyL3N3YXBmaWxlIGJzPTFNIGNvdW50PTMwMAogIGNobW9kIDYwMCAvLmJvdHRsZXJvY2tldC9yb290ZnMvdmFyL3N3YXBmaWxlCiAgbWtzd2FwIC8uYm90dGxlcm9ja2V0L3Jvb3Rmcy92YXIvc3dhcGZpbGUKZmkKc3dhcG9uIC8uYm90dGxlcm9ja2V0L3Jvb3Rmcy92YXIvc3dhcGZpbGU="
```

On the instance I can see that it works:

```
bash-5.1# free -m
               total        used        free      shared  buff/cache   available
Mem:             410         234           5           0         170         165
Swap:            299          30         269
```

Would be great to allow for swap. It is currently the only thing blocking me from utilizing Bottlerocket to a greater extent. Thanks!

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -642,6 +642,7 @@ func withBootstrap() oci.SpecOpts {
 		// container's capabilities.
 		seccomp.WithDefaultProfile(),
 		oci.WithAllDevicesAllowed,
+		withSwapManagement,
 	)
 }
 
@@ -837,6 +838,18 @@ func withRootFsShared() oci.SpecOpts {
 		}
 		return nil
 	}
+}
+
+// withSwapManagement allows the swapon and swapoff syscalls
+func withSwapManagement(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
+	if s.Linux != nil && s.Linux.Seccomp != nil && s.Linux.Seccomp.Syscalls != nil {
+		s.Linux.Seccomp.Syscalls = append(s.Linux.Seccomp.Syscalls, runtimespec.LinuxSyscall{
+			Names:  []string{"swapon", "swapoff"},
+			Action: runtimespec.ActAllow,
+			Args:   []runtimespec.LinuxSeccompArg{},
+		})
+	}
+	return nil
 }
 
 // withProxyEnv reads proxy environment variables and returns a spec option for passing said proxy environment variables


sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -642,6 +642,15 @@ func withBootstrap() oci.SpecOpts {
...
 		oci.WithAllDevicesAllowed,
+		func(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
+			// Allow bootstrap containers to manage swap
+			s.Linux.Seccomp.Syscalls = append(s.Linux.Seccomp.Syscalls, runtimespec.LinuxSyscall{
+				Names:  []string{"swapon", "swapoff"},
+				Action: runtimespec.ActAllow,
+				Args:   []runtimespec.LinuxSeccompArg{},
+			})
+			return nil
+		},
```
etungsten: Can you please encapsulate this in a separate function like we do for `withRootFsShared`? https://github.com/bottlerocket-os/bottlerocket/blob/1be9ec869cc825a48da0de9e89905a4e3791e123/sources/host-ctr/cmd/host-ctr/main.go#L841-L849

etungsten (commented):
Thanks for the PR! 
I'll let others comment on if it's OK to enable these syscalls in the bootstrap container.

sources/host-ctr/cmd/host-ctr/main.go
stefansundin: Done, including testing that it still works.

bcressey (approved):

bcressey:
Thanks for implementing this!

stefansundin:
@bcressey Is there a reason why we need a function that returns the function used?

I tried this and it seems to work just as well:


```diff
$ git diff sources/host-ctr/cmd/host-ctr/main.go
diff --git a/sources/host-ctr/cmd/host-ctr/main.go b/sources/host-ctr/cmd/host-ctr/main.go
index 935aba07..977330d8 100644
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -642,7 +642,7 @@ func withBootstrap() oci.SpecOpts {
                // container's capabilities.
                seccomp.WithDefaultProfile(),
                oci.WithAllDevicesAllowed,
-               withSwapManagement(),
+               withSwapManagement,
        )
 }
 
@@ -841,17 +841,15 @@ func withRootFsShared() oci.SpecOpts {
 }
 
 // withSwapManagement allows the swapon and swapoff syscalls
-func withSwapManagement() oci.SpecOpts {
-       return func(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
-               if s.Linux != nil && s.Linux.Seccomp != nil && s.Linux.Seccomp.Syscalls != nil {
-                       s.Linux.Seccomp.Syscalls = append(s.Linux.Seccomp.Syscalls, runtimespec.LinuxSyscall{
-                               Names:  []string{"swapon", "swapoff"},
-                               Action: runtimespec.ActAllow,
-                               Args:   []runtimespec.LinuxSeccompArg{},
-                       })
-               }
-               return nil
+func withSwapManagement(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
+       if s.Linux != nil && s.Linux.Seccomp != nil && s.Linux.Seccomp.Syscalls != nil {
+               s.Linux.Seccomp.Syscalls = append(s.Linux.Seccomp.Syscalls, runtimespec.LinuxSyscall{
+                       Names:  []string{"swapon", "swapoff"},
+                       Action: runtimespec.ActAllow,
+                       Args:   []runtimespec.LinuxSeccompArg{},
+               })
        }
+       return nil
 }
```

And I also think early-returning when checking for `nil` might also make it a little bit better?

```go
// withSwapManagement allows the swapon and swapoff syscalls
func withSwapManagement(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
	if s.Linux == nil || s.Linux.Seccomp == nil || s.Linux.Seccomp.Syscalls == nil {
		return nil
	}
	s.Linux.Seccomp.Syscalls = append(s.Linux.Seccomp.Syscalls, runtimespec.LinuxSyscall{
		Names:  []string{"swapon", "swapoff"},
		Action: runtimespec.ActAllow,
		Args:   []runtimespec.LinuxSeccompArg{},
	})
	return nil
}
```

Anyway, I won't change it unless you tell me to. :)

arnaldo2792:
I did some research to understand why the containerd's default seccomp filter doesn't include the `swapon` / `swapoff` syscalls, just to be sure we weren't doing the wrong thing. Probably they will answer later in [this discussion].

Regardless of the outcome of the linked discussion, in an offline chat I had with @dims I learned that by default the kubelet will fail to start *if* it wasn't started with `--fail-swap-on=false`, which is clearly the case in Bottlerocket's implementation. For details refer to [this PR].

There are [ongoing efforts] to better support swap memory in the kubelet, and there is some work that [already landed], but said work has to be picked up by the CRI implementations (aka, containerd) to close the gap. I think once they close that gap, we should follow @stefansundin suggestion of having an API to allocate swap memory *and* support the new settings that allow the kubelet to use the allocated memory (@bcressey / @etungsten what do you all think?).

All that said, @stefansundin *if* you are using an ECS image you should be able to use the swap memory without impacting the ECS agent. But that isn't the case for the kubernetes images, since as mentioned before, the kubelet will refuse to start.

[this discussion]: https://github.com/containerd/containerd/discussions/7608
[this PR]: https://github.com/kubernetes/kubernetes/pull/47181
[ongoing efforts]: https://github.com/kubernetes/enhancements/issues/2400
[already landed]: https://github.com/kubernetes/kubernetes/issues/53533#issuecomment-875703679

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -839,6 +840,20 @@ func withRootFsShared() oci.SpecOpts {
 	}
 }
 
+// withSwapManagement allows the swapon and swapoff syscalls
```
arnaldo2792, line 843: I joined a little late to the party:

I think for this function, it should be fine to skip the closure and just use the implementation you suggested in your other comment. The reason being that there aren't any parameters used by the internal function to further customize the configuration created.

This pattern (not using a closure) is something that we already use, a few lines above we call `oci.WithAllDevicesAllowed` which is just a function that doesn't return a closure.

One more thing, could you please squash your commits :sweat_smile: ? 

(Lastly, thanks for implementing this! Kudos for digging into the implementation details!)

stefansundin: 👍 I removed the closure.

> One more thing, could you please squash your commits 😅 ?

I can do it if you definitely want me to, but first I just want to make sure that you know that GitHub can squash the commits when you merge? Just select <kbd>Squash and merge</kbd> in the dropdown next to the merge button.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits

That should give you full control over the commit message and what the final commit looks like.

If that for some reason isn't sufficient then please let me know and I'll be happy to squash it. 👍

etungsten (approved):

sources/host-ctr/cmd/host-ctr/main.go
arnaldo2792: We are aware of the feature :sweat_smile: , but we favor contributors squashing/fixing up their commits instead of using the tool provided by GitHub. If you are OK with us squashing the commits for you, we can do it!

stefansundin: I squashed the commits. 👍

bcressey (approved):

arnaldo2792 (approved):

Issue #1702, opened 2021-08-09
kubelet: add credential provider feature-gate


samjo-nyang:
**What I'd like:**
Allow arbitrary kubelet feature-gates settings to the kubelet configuration file, such as
```
featureGates:
  RotateKubeletServerCertificate: true
  CSIMigration: false
{{#if settings.kubernetes.feature-gates}}
{{#each settings.kubernetes.feature-gates as |v k|}}  {{k}}: {{v}}
{{/each}}
{{/if}}
```

**Any alternatives you've considered:** (nothing)

FYI: I'd like to try https://kubernetes.io/docs/tasks/kubelet-credential-provider/kubelet-credential-provider/ , but it requires to enable feature-gates on kubelet. I have internal patch, but there are no equivalent features on the upstream.

samuelkarp:
Hi @samjo-nyang, thanks for opening this issue.  Is the Kubelet credential provider the only feature gate you're looking to enable or are there others that you're interested in as well?

samjo-nyang:
currently, i am only interested in the credential provider feature

samuelkarp:
Hey @samjo-nyang, thanks for clarifying!  We've discussed this and at this time we're not planning to add arbitrary feature-gates; Bottlerocket's validated settings API is one of the mechanisms we believe helps with both security and predictability of configuration at scale.  However, we are happy to consider adding individual feature gates such as the credential provider feature.  I've re-titled this issue to reflect the request for the credential provider feature gate.

Can you tell us a bit more about how you'd like to use that feature?  Are you interested in adopting a credential provider for a registry other than Amazon ECR?

samjo-nyang:
Yes, I have a private oci image registry and I want to implement a custom credential provider for the registry.

stmcginnis:
Hey @samjo-nyang - As part of some other changes I am working on to enable credential providers, that will bring in the ability to enable the `KubeletCredentialProviders` feature gate.

I'm wondering if that change is enough to address this and close out this issue? Or is there a use case beyond that that you would still need addressed? There are some concerns about exposing this arbitrary set of feature gates, but we can discuss more if there is something you need to do that is prevented by not having this.

Thanks!

samjo-nyang:
Hi, @stmcginnis 
For now, it is enough to enable (or give ability to enable) `KubeletCredentialProviders`.
Thanks!

stmcginnis:
[...]

PR #2541, opened 2022-11-01
schnauzer: Add any_enabled template helper

stmcginnis:
**Issue number:**

Related https://github.com/bottlerocket-os/bottlerocket/issues/1702

**Description of changes:**

This adds a new template helper called `any_enabled` that can be used in our templates to check if anything in a collection contains something with the property of `enabled` set to true.

The helper looks for an array, then checks that the objects within the array has an "enabled" property. If they do and if any of them are set to "true" the condition evaluates to be "truey" for conditional blocks within the template.

If the passed in value is not an array, does not have an "enabled" property, or has no elements with that property set to true it will evaluate to falsey.

**Testing done:**

See https://github.com/bottlerocket-os/bottlerocket/pull/2377 for complete testing details.

**Terms of contribution:**

### DIFF
--- a/sources/api/schnauzer/src/helpers.rs
+++ b/sources/api/schnauzer/src/helpers.rs
@@ -3,8 +3,11 @@
 // text at render time.
 
 use dns_lookup::lookup_host;
-use handlebars::{Context, Handlebars, Helper, Output, RenderContext, RenderError};
+use handlebars::{
+    handlebars_helper, Context, Handlebars, Helper, Output, RenderContext, RenderError,
+};
 use lazy_static::lazy_static;
+use serde::Deserialize;
 use serde_json::value::Value;
 use snafu::{OptionExt, ResultExt};
 use std::borrow::Borrow;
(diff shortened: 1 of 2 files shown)


sources/api/schnauzer/src/helpers.rs
```diff
@@ -2362,3 +2397,75 @@ mod test_etc_hosts_helpers {
...
+    fn test_any_enabled_with_empty_elements() {
+        let result = setup_and_render_template(TEMPLATE, &json!([])).unwrap();
+        let expected = "disabled";
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn test_any_enabled_with_bool_flag_as_string() {
+        // Helper is only expected to work with boolean values
+        let result = setup_and_render_template(TEMPLATE, &json!([{"enabled": "true"}])).unwrap();
```
webern, line 2491: Seems important for testing to verify that fields other than `enabled` have no effect.

```suggestion
        let result = setup_and_render_template(TEMPLATE, &json!([{"foo": [], "enabled": "true"}])).unwrap();
```

webern (approved):

sources/api/schnauzer/src/helpers.rs
stmcginnis: This is being done on line 2429. Since this (line 2451) tests a negative case I don't know if it would add much value here. Thoughts?

webern: Sorry I missed that. This is good.

etungsten:
I think depending on the resolution of https://github.com/bottlerocket-os/bottlerocket/pull/2377#discussion_r1009912214, we might need to change this from interpreting a list to a map.

stmcginnis:
> I think depending on the resolution of https://github.com/bottlerocket-os/bottlerocket/pull/2377#discussion_r1009912214, we might need to change this from interpreting a list to a map.

Hmm, this may be an issue. Not sure how that would be handled using handlebar syntax. Maybe another helper that would convert the items in the hashmap into a list that would then be passed in to this helper?

stmcginnis:
> we might need to change this from interpreting a list to a map.

Added ability to also use a map as input with unit test coverage for various input.

etungsten (approved):
:ice_cream:

PR #2542, opened 2022-11-01
packages: Add AWS config

stmcginnis:
**Issue number:**

Related #1702

**Description of changes:**

Adds configuration settings for controlling AWS credentials and configuration. This enables setting the `~/.aws/config` and `~/.aws/credentials` file contents so anything using an aws client can be configured to use something other than the default instance role.

This adds the AwsSettings to the non-AWS k8s variants in preparation of their use for configuring credential providers.

**Testing done:**

See https://github.com/bottlerocket-os/bottlerocket/pull/2377 for complete testing details.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -951,9 +951,25 @@ They can be overridden for testing purposes in [the same way as other settings](
 
 ##### AWS-specific settings
 
-AWS-specific settings are automatically set based on calls to the Instance MetaData Service (IMDS).
+* `settings.aws.config`: The base64 encoded content to use for AWS configuration (e.g. `base64 -w0 ~/.aws/config`).
+* `settings.aws.credentials`: The base64 encoded content to use for AWS credentials (e.g. `base64 -w0 ~/.aws/credentials`).
+* `settings.aws.profile`: The profile name to use from the provided `config` and `credentials` settings.
+
+  For example:
+
+  ```toml
+  [settings.aws]
+  profile = "myprofile"
+  ```
+
+  **Note**: If `settings.aws.profile` is not set, the setting will fallback to the "default" profile.
+
+  **Note:** The `config`, `credentials`, and `profile` are optional and do not need to be set when using an Instance Profile when running on an AWS instance.
 
 * `settings.aws.region`: This is set to the AWS region in which the instance is running, for example `us-west-2`.
+
+  The `region` setting is automatically inferred based on calls to the Instance MetaData Service (IMDS) when running within AWS.
+  It does not need to be explicitly set unless you have a reason to override this default value.
 
 ### Logs
 
--- a/sources/api/migration/migrations/v1.11.0/aws-creds/src/main.rs
+++ b/sources/api/migration/migrations/v1.11.0/aws-creds/src/main.rs
@@ -1,0 +1,30 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::{AddPrefixesMigration, AddSettingsMigration};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added new settings for AWS credential configuration.
+fn run() -> Result<()> {
+    if cfg!(variant_platform = "aws") {
+        migrate(AddSettingsMigration(&[
+            "settings.aws.config",
+            "settings.aws.credentials",
+            "settings.aws.profile",
+        ]))
+    } else {
+        // Non-AWS variants did not have any AWS setting until this point,
+        // so need to completely clean up on downgrade.
+        migrate(AddPrefixesMigration(vec!["settings.aws"]))
+    }
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -340,6 +340,9 @@ struct BootSettings {
 #[model]
 struct AwsSettings {
     region: SingleLineString,
+    config: ValidBase64,
+    credentials: ValidBase64,
+    profile: SingleLineString,
 }
 
 // Metrics settings
--- a/Release.toml
+++ b/Release.toml
@@ -156,3 +156,8 @@ version = "1.10.1"
     "migrate_v1.10.1_container-runtime.lz4",
     "migrate_v1.10.1_container-runtime-metadata.lz4"
 ]
+"(1.10.1, 1.11.0)" = [
+    "migrate_v1.11.0_aws-config-settings.lz4",
+    "migrate_v1.11.0_aws-creds.lz4",
+    "migrate_v1.11.0_aws-creds-metadata.lz4",
+]
--- a/packages/filesystem/filesystem.spec
+++ b/packages/filesystem/filesystem.spec
@@ -28,9 +28,10 @@ mkdir -p %{buildroot}%{_cross_datadir}
 mkdir -p %{buildroot}%{_cross_infodir}
 mkdir -p %{buildroot}%{_cross_mandir}
 mkdir -p %{buildroot}%{_cross_localstatedir}
-mkdir -p %{buildroot}/{boot,dev,proc,root,run,sys,tmp}
+mkdir -p %{buildroot}/{boot,dev,proc,run,sys,tmp}
 mkdir -p %{buildroot}/{home,local,media,mnt,opt,srv}
 mkdir -p %{buildroot}/media/cdrom
+mkdir -p %{buildroot}/root/.aws
 
 ln -s .%{_cross_prefix} %{buildroot}%{_prefix}
 ln -s .%{_cross_bindir} %{buildroot}/bin
--- a/packages/release/aws-config
+++ b/packages/release/aws-config
@@ -1,0 +1,3 @@
+{{~#if settings.aws.config~}}
+{{base64_decode settings.aws.config}}
+{{~/if~}}
--- a/packages/release/aws-credentials
+++ b/packages/release/aws-credentials
@@ -1,0 +1,3 @@
+{{~#if settings.aws.credentials~}}
+{{base64_decode settings.aws.credentials}}
+{{~/if~}}
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -19,6 +19,8 @@ Source202: hostname-env
 Source203: hosts.template
 Source204: modprobe-conf.template
 Source205: netdog.template
+Source206: aws-config
+Source207: aws-credentials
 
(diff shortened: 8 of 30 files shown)


sources/models/src/lib.rs
```diff
@@ -340,6 +340,12 @@ struct BootSettings {
 #[model]
 struct AwsSettings {
     region: SingleLineString,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    config: Option<ValidBase64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    credentials: Option<ValidBase64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    profile: Option<SingleLineString>,
```
etungsten: The `skip_serializing_if` attribute and `Option` wrapping for the fields should be taken care of by `#[model]`. We only need to add them if we're adding any additional attribute that `#[model]` doesn't cover. See

https://github.com/bottlerocket-os/bottlerocket/blob/b3c35caa5808a15325c961e63bcc39bf163d8a3a/sources/models/model-derive/src/lib.rs#L143-L168

sources/api/migration/migrations/v1.11.0/aws-creds/src/main.rs
```diff
@@ -0,0 +1,24 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddSettingsMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting for AWS credential configuration.
```
etungsten: nit: 
```suggestion
/// We added new settings for AWS credential configuration.
```

sources/api/migration/migrations/v1.11.0/kubelet-aws-config/src/main.rs
```diff
@@ -0,0 +1,29 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting for configuring kubelet's image credential
+/// provider plugins. Initially this is only to support ecr-credential-provider,
```
etungsten: This comment seems inaccurate.

sources/api/migration/migrations/v1.11.0/kubelet-aws-config/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+/// but others may be added as needed.
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec![
+        "settings.aws.credentials",
+        "settings.aws.config",
+        "settings.aws.profile",
```
etungsten: These three settings should be covered by the `aws-creds` `AddSettings` migration, so they don't need to be listed here again

etungsten (changes requested):
I can't comment too much on changes to the rpm spec files and the new mount unit. But they look fine to me.

sources/models/src/lib.rs
stmcginnis: Great! I just copied what I picked up somewhere else, so we may have a chance for some other clean up after this. Removed from this instance.

sources/api/migration/migrations/v1.11.0/kubelet-aws-config/src/main.rs
stmcginnis: How so? It does seem a little too specific. Since it's being added for ecr-credential-provider support, but with the hope that it will be used elsewhere, there probably isn't a need to capture that driver in a comment though. I'll drop it.

sources/models/src/lib.rs
```diff
@@ -340,6 +340,9 @@ struct BootSettings {
 #[model]
 struct AwsSettings {
     region: SingleLineString,
+    config: Option<ValidBase64>,
+    credentials: Option<ValidBase64>,
+    profile: Option<SingleLineString>,
```
etungsten: We can leave out the `Option` as well.

etungsten (approved):

stmcginnis:
Updated to address this comment from the original uber PR: https://github.com/bottlerocket-os/bottlerocket/pull/2377/files#r1010049045

README.md
```diff
@@ -951,10 +951,26 @@ They can be overridden for testing purposes in [the same way as other settings](
...
+  profile = "myprofile"
+  ```
+
+  **Note**: If `settings.aws.profile` is not set, it will fallback to using the "default" profile.
+
+  **Note:** The `config`, `credentials`, and `profile` are optional and do not need to be set when using an Instance Profile when running on an AWS instance.
 
 * `settings.aws.region`: This is set to the AWS region in which the instance is running, for example `us-west-2`.
 
+  The `region` setting is automatically inferred based on calls to the Instance MetaData Service (IMDS).
```
arnaldo2792: This is only true in AWS variants, right? In VMWare / metal variants, IMDS isn't used to automatically fetch the region. Should this be specified in the note?

sources/api/migration/migrations/v1.11.0/aws-creds/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+            "settings.aws.profile",
+        ]))
+    } else {
+        // Non-AWS variants did not have the region setting until this point,
+        // so need to be included here to remove on downgrade.
+        migrate(AddSettingsMigration(&[
+            "settings.aws.config",
+            "settings.aws.credentials",
+            "settings.aws.profile",
+            "settings.aws.region",
```
arnaldo2792: I'm confused, I didn't see any update to the other platforms' models that added `AwsSettings` as in [AWS variants] (I also looked at the original PR but couldn't find anything there either). Is that going to happen in a different PR?

[AWS variants]: https://github.com/bottlerocket-os/bottlerocket/blob/ea2b374ea67ad634d20a18f83d3d4dfdb57bdcb1/sources/models/src/aws-k8s-1.24-nvidia/mod.rs#L25

stmcginnis: `AwsSettings` already was there, but it would only be present for `aws-*` variants. 

See Ben's comment here: https://github.com/bottlerocket-os/bottlerocket/pull/2377/files#r1010049045

README.md
stmcginnis: Added some more wording.

sources/api/migration/migrations/v1.11.0/aws-creds/src/main.rs
stmcginnis: Those changes were being done in a later commit when they were actually being used for these other non-AWS variants. But to make it more clear as far as the migration is concerned I have now pulled those additions into this PR. So they will be available to the other variants, but a later PR will actually wire them in to be used when the other settings are configured.

README.md
```diff
@@ -951,10 +951,26 @@ They can be overridden for testing purposes in [the same way as other settings](
 
 ##### AWS-specific settings
 
-AWS-specific settings are automatically set based on calls to the Instance MetaData Service (IMDS).
+* `settings.aws.config`: The base64 encoded content to use for AWS configuration (e.g. `base64 ~/.aws/config`).
+* `settings.aws.credentials`: The base64 encoded content to use for AWS credentials (e.g. `base64 ~/.aws/credentials`).
```
bcressey: perhaps this, to avoid line breaks in the output:
```suggestion
* `settings.aws.config`: The base64 encoded content to use for AWS configuration (e.g. `base64 -w0 ~/.aws/config`).
* `settings.aws.credentials`: The base64 encoded content to use for AWS credentials (e.g. `base64 -w0 ~/.aws/credentials`).
```

README.md
```diff
@@ -951,10 +951,26 @@ They can be overridden for testing purposes in [the same way as other settings](
...
+* `settings.aws.profile`: The profile name to use from the provided `config` and `credentials` settings.
+
+  For example:
+
+  ```toml
+  [settings.aws]
+  profile = "myprofile"
+  ```
+
+  **Note**: If `settings.aws.profile` is not set, it will fallback to using the "default" profile.
```
bcressey: What is "it" that will fall back?

sources/api/migration/migrations/v1.11.0/aws-creds/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::process;
+
+/// We added new settings for AWS credential configuration.
+fn run() -> Result<()> {
+    if cfg!(variant_platform = "aws") {
+        migrate(AddSettingsMigration(&[
+            "settings.aws.config",
+            "settings.aws.credentials",
+            "settings.aws.profile",
+        ]))
```
bcressey, line 14: This cfg attribute needs a build script, like `pki-affected-services/build.rs`, in order for this to ever be set.

sources/api/migration/migrations/v1.11.0/aws-creds/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+    } else {
+        // Non-AWS variants did not have the region setting until this point,
+        // so need to be included here to remove on downgrade.
+        migrate(AddSettingsMigration(&[
+            "settings.aws.config",
+            "settings.aws.credentials",
+            "settings.aws.profile",
+            "settings.aws.region",
+        ]))
+    }
```
bcressey: @etungsten do you know if this branch needs to be an `AddPrefixesMigration` across all of `settings.aws` in order to clean up the "aws" directory in the datastore?

(It might not hurt to do that just in case.)

sources/api/migration/migrations/v1.11.0/kubelet-aws-config/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "kubelet-aws-config"
```
bcressey: nit: since this is not specific to kubelet
```suggestion
name = "aws-config"
```

sources/api/migration/migrations/v1.11.0/aws-creds/src/main.rs
stmcginnis: I was actually wondering about that for the non-AWS case.

sources/api/migration/migrations/v1.11.0/kubelet-aws-config/Cargo.toml
stmcginnis: That conflicts with another package named `aws-config`, so I will change this to `aws-config-settings` to be more descriptive.

sources/api/migration/migrations/v1.11.0/aws-creds/src/main.rs
etungsten: > in order to clean up the "aws" directory in the datastore?

It shouldn't leave behind a directory if there are no settings values living under that directory. 

> (It might not hurt to do that just in case.)

I agree, let's change it to `AddPrefixMigration` for non-AWS variants since it fits the usecase.

bcressey (approved):

PR #2548, opened 2022-11-03
fix package rebuilds for feature flags

### LINKED ISSUE #2546: gratuitous package rebuilds after changing the variant manifest
**Platform I'm building on:**
Linux

When I modify `Cargo.toml` for a variant, any packages that make use of image feature flags should be rebuilt. Packages that aren't sensitive to variant changes should not be rebuilt.

**What actually happened:**
All the packages got rebuilt, as a result of the change in https://github.com/bottlerocket-os/bottlerocket/commit/eb44f13a787. 😿 

1. From a clean checkout, perform a clean build of `aws-dev`. All packages will be built.
2. Modify `variants/aws-dev/Cargo.toml` in some way.
3. Perform a second build of `aws-dev`. All packages will be rebuilt.

bcressey:
**Issue number:**

Closes #2546

**Description of changes:**
```
buildsys: fix package rebuilds for feature flags

The previous approach for detecting changes to image feature flags
that should trigger a rebuild was incorrect. It would cause packages
to be rebuilt whenever the variant manifest that was the target at
the time of last rebuild changed.

This would happen regardless of whether image feature flags changed,
whether the same flags were set for the current variant manifest, or
whether the flags were even used by the package.

Fixing this requires two prerequisites.

First, the current image feature flags must be set in the environment
prior to invoking `cargo build`, or it will be too late to emit the
`cargo:rerun-if-env-changed` directives, since the environment that
`cargo` runs in is not influenced by changes from the build script.

Second, each package must declare which image feature flags it cares
about, if any, so that the correct cargo directive can be omitted.
The environment does not carry this information, since the absence of
a particular environment variable can also mean that a package needs
to be rebuilt to disable a previously-enabled feature.

With this functionality in place, each package build can emit the
directives needed for correct change detection. For robustness, only
the feature flags tracked by the package are passed through to the
build, meaning the feature will always be treated as "disabled" by
bconds in the spec file.
```

**Testing done:**
Verified that modifying the variant manifest no longer triggers a rebuild of all packages originally built for that variant.

On my Secure Boot branch, verified that switching back and forth between variants that have different feature flags trigger the expected rebuilds of `shim` and `grub`.

With `export CARGO_LOG=cargo::core::compiler::fingerprint=trace` set, I see messages like this when the switch causes a rebuild:
```
[2022-11-03T17:42:42Z INFO  cargo::core::compiler::fingerprint]     err: env var `BUILDSYS_VARIANT_IMAGE_FEATURE_UEFI_SECURE_BOOT` changed: previously Some("1"), now None
```

**Terms of contribution:**

### DIFF
--- a/tools/buildsys/src/bin/bottlerocket-variant/main.rs
+++ b/tools/buildsys/src/bin/bottlerocket-variant/main.rs
@@ -1,5 +1,7 @@
 use bottlerocket_variant::Variant;
+use buildsys::manifest::ManifestInfo;
 use snafu::ResultExt;
+use std::path::PathBuf;
 use std::{env, process};
 
 // Returning a Result from main makes it print a Debug representation of the error, but with Snafu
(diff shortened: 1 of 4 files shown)


stmcginnis (approved):
Code changes look fine to me, and CI tests are happy.

webern (approved):

jpmcb (commented):
👏🏼  very nice!

jpmcb (approved):

PR #2553, opened 2022-11-04
Enable IAM Roles Anywhere with the k8s  `ecr-credential-provider` plugin

### LINKED ISSUE #2310: Kubelet Credential Provider Support
**What I'd like:**
Allow credential-provider settings to be passed into a credential helper. In this case I would like to use [IAM Roles Anywhere](https://docs.aws.amazon.com/rolesanywhere/latest/userguide/credential-helper.html) with [ecr-credential-provider](https://github.com/kubernetes/cloud-provider-aws/tree/master/cmd/ecr-credential-provider). For this particular case it would require an api setting 
```
settings.eks.ecr-credential-provider-iam-roles-anywhere
enabled = true
profile-arn = "..."
role-arn = "..."
trust-anchor-arn = "..."
```

These values would get piped down into a [configuration file](https://kubernetes.io/docs/tasks/kubelet-credential-provider/kubelet-credential-provider/#configure-a-kubelet-credential-provider) for kubelet credential provider.

If this setting is enabled the feature gate in #1702 could just be enabled for the kubelet, as those are needed for [kubelet-credential-provider](https://kubernetes.io/docs/tasks/kubelet-credential-provider/kubelet-credential-provider/#before-you-begin).

stmcginnis:
**Issue number:**

Closes #2310 

**Description of changes:**

This adds a new package to place the `aws_signing_helper` binary in the /usr/bin PATH to enable its use for k8s credential provider support of IAM Roles Anywhere.

Documentation also added to give an example of how to configure the AWS client settings to use this.

**Testing done:**

Deployed node and verified `aws_signing_helper` is present and executable in the `/usr/bin` directory.
Additional testing TBD with the EKS-A team to validate functionality.

**Terms of contribution:**

### DIFF
--- a/packages/aws-signing-helper/Cargo.toml
+++ b/packages/aws-signing-helper/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "aws-signing-helper"
+version = "0.1.0"
+edition = "2018"
+publish = false
+build = "build.rs"
+
+[lib]
+path = "pkg.rs"
+
+[package.metadata.build-package]
+releases-url = "https://github.com/aws/rolesanywhere-credential-helper/releases"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/aws/rolesanywhere-credential-helper/archive/v1.0.2/rolesanywhere-credential-helper-v1.0.2.tar.gz"
+sha512 = "b364bf8f73f33e7ac1db6a1153880ffa2e4af52a3d8f7b224cc5a9a2e545432a9bf408191048a406fdb995a034b997cdeeb752b4dfcea8288be5baeca8e69b9a"
+bundle-modules = [ "go" ]
+
+[build-dependencies]
+glibc = { path = "../glibc" }
--- a/packages/aws-signing-helper/aws-signing-helper.spec
+++ b/packages/aws-signing-helper/aws-signing-helper.spec
@@ -1,0 +1,46 @@
+%global goproject github.com/aws
+%global gorepo rolesanywhere-credential-helper
+%global goimport %{goproject}/%{gorepo}
+
+%global gover 1.0.2
+%global rpmver %{gover}
+
+%global _dwz_low_mem_die_limit 0
+
+Name: %{_cross_os}aws-signing-helper
+Version: %{rpmver}
+Release: 1%{?dist}
+Summary: AWS signing helper for IAM Roles Anywhere support
+License: Apache-2.0
+URL: https://github.com/aws/rolesanywhere-credential-helper
+
+Source: rolesanywhere-credential-helper-v%{gover}.tar.gz
+Source1: bundled-rolesanywhere-credential-helper-v%{gover}.tar.gz
+
+BuildRequires: %{_cross_os}glibc-devel
+
+%description
+%{summary}.
+
+%prep
+%setup -n %{gorepo}-%{gover} -q
+%setup -T -D -n %{gorepo}-%{gover} -b 1 -q
+
+%build
+%set_cross_go_flags
+
+go build ${GOFLAGS} -buildmode=pie -ldflags "-X 'main.Version=${gover}' ${GOLDFLAGS}" -o aws-signing-helper cmd/aws_signing_helper/main.go
+
+%install
+install -d %{buildroot}%{_cross_bindir}
+install -p -m 0755 aws-signing-helper %{buildroot}%{_cross_bindir}/aws_signing_helper
+ln -sf aws_signing_helper %{buildroot}%{_cross_bindir}/aws-signing-helper
+
+%cross_scan_attribution go-vendor vendor
+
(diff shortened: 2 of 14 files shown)


packages/aws-signing-helper/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+releases-url = "https://github.com/aws/rolesanywhere-credential-helper/releases"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/aws/rolesanywhere-credential-helper/archive/b193b4e041c18896203b77724fa0577cf73949fa/rolesanywhere-credential-helper-b193b4e041c18896203b77724fa0577cf73949fa.tar.gz"
+sha512 = "90b13b9d88d1c351859f89e1d53a36caa1b02ae2475e3fba059c1e151e62dea4105b08ed7a9d79077caa9dece48ae557c1a3aef94924f05b299a9d7e355ee1e7"
+bundle-modules = [ "go" ]
+
+[build-dependencies]
+glibc = { path = "../glibc" }
+
```
bcressey: nit: extra line

packages/aws-signing-helper/aws-signing-helper.spec
```diff
@@ -0,0 +1,47 @@
...
+%setup -T -D -n %{gorepo}-%{gitrev} -b 1 -q
+
+%build
+%set_cross_go_flags
+
+go build ${GOFLAGS} -buildmode=pie -ldflags "-X 'main.Version=${gover}' ${GOLDFLAGS}" -o aws-signing-helper cmd/aws_signing_helper/main.go
+
+%install
+install -d %{buildroot}%{_cross_bindir}
+install -p -m 0755 aws-signing-helper %{buildroot}%{_cross_bindir}/aws_signing_helper
```
bcressey, line 36: Should we add a symlink for `aws-signing-helper`, or is the snake-case version the only one people are likely to use?

packages/aws-signing-helper/pkg.rs
```diff
@@ -0,0 +1,2 @@
+// not used
+
```
bcressey: nit: extra newline here

bcressey (approved):

packages/aws-signing-helper/aws-signing-helper.spec
stmcginnis: Good call - I've certainly made that mistake before, so having both `_` and `-` names will help make sure the customer doesn't accidentally use the wrong one.

packages/aws-signing-helper/aws-signing-helper.spec
```diff
@@ -0,0 +1,49 @@
+%global goproject github.com/aws
+%global gorepo rolesanywhere-credential-helper
+%global goimport %{goproject}/%{gorepo}
+
+%global gover 1.0.2
+%global rpmver %{gover}
+
+%global _dwz_low_mem_die_limit 0
+
+%global gitrev b193b4e041c18896203b77724fa0577cf73949fa
```
webern: This is what needs to be replaced by a version tag, right?

webern (approved):

webern (approved):
Looks good, but can't be merged until the source is pushed to the lookaside cache and the version tag is obtained.

webern (approved):

packages/aws-signing-helper/aws-signing-helper.spec
stmcginnis: Here and in `Cargo.toml` will need to be updated to the official release tag. Then based on that, the source tar.gz needs to be added to our lookaside cache. Then we should be good to go.

webern (approved):

jpmcb (approved):
Looks good!! 👏🏼 
```
❯ sha512sum rolesanywhere-credential-helper-v1.0.2.tar.gz
b364bf8f73f33e7ac1db6a1153880ffa2e4af52a3d8f7b224cc5a9a2e545432a9bf408191048a406fdb995a034b997cdeeb752b4dfcea8288be5baeca8e69b9a  rolesanywhere-credential-helper-v1.0.2.tar.gz
```

packages/aws-signing-helper/aws-signing-helper.spec
```diff
@@ -0,0 +1,49 @@
...
+%global shortrev %(c=%{gitrev}; echo ${c:0:7})
```
bcressey: nit: neither of these variables are used, but we can clean them up next time there's an update

bcressey (approved):

packages/aws-signing-helper/aws-signing-helper.spec
stmcginnis: I'll just take care of that now. Should be a quick update.

I'll merge after jobs pass unless anyone let's me know they would like to review

PR #2557, opened 2022-11-07
migrations: add migration for 'configuration-files.kubelet-server*'

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
    migrations: add migration for 'configuration-files.kubelet-server*'
```

**Testing done:**
Performed upgrade downgrade testing.
On v1.11.0, I verified `configuration-files.kubelet-server-*` are in the datastore:
```bash
bash-5.1# apiclient get /os                                                                       
{
  "arch": "x86_64",
  "build_id": "ba64a2b1-dirty",
  "pretty_name": "Bottlerocket OS 1.11.0 (aws-k8s-1.23)",
  "variant_id": "aws-k8s-1.23",
  "version_id": "1.11.0"
}
bash-5.1# ls -al /var/lib/bottlerocket/datastore/v1.11.0/live/configuration-files/kubelet-server-*
/var/lib/bottlerocket/datastore/v1.11.0/live/configuration-files/kubelet-server-crt:
total 16
drwxr-xr-x.  2 root root 4096 Nov  7 19:04 .
drwxr-xr-x. 26 root root 4096 Nov  7 19:04 ..
-rw-r--r--.  1 root root   40 Nov  7 19:04 path
-rw-r--r--.  1 root root   41 Nov  7 19:04 template-path

/var/lib/bottlerocket/datastore/v1.11.0/live/configuration-files/kubelet-server-key:
total 16
drwxr-xr-x.  2 root root 4096 Nov  7 19:04 .
drwxr-xr-x. 26 root root 4096 Nov  7 19:04 ..
-rw-r--r--.  1 root root   40 Nov  7 19:04 path
-rw-r--r--.  1 root root   41 Nov  7 19:04 template-path
bash-5.1# 
```

Then I downgraded back to v1.10.1, host comes back up fine and I verified that the migration successfully removed those paths on downgrade:
```bash
bash-5.1# apiclient get /os
{
  "arch": "x86_64",
  "build_id": "5d27ae74",
  "pretty_name": "Bottlerocket OS 1.10.1 (aws-k8s-1.23)",
  "variant_id": "aws-k8s-1.23",
  "version_id": "1.10.1"
}
bash-5.1# ls -al /var/lib/bottlerocket/datastore/current/live/configuration-files/kubelet-server-*
ls: cannot access '/var/lib/bottlerocket/datastore/current/live/configuration-files/kubelet-server-*': No such file or directory
```

**Terms of contribution:**

### DIFF
--- a/Release.toml
+++ b/Release.toml
@@ -162,6 +162,7 @@ version = "1.10.1"
     "migrate_v1.11.0_aws-creds-metadata.lz4",
     "migrate_v1.11.0_credential-providers.lz4",
     "migrate_v1.11.0_kubelet-tls-config.lz4",
+    "migrate_v1.11.0_kubelet-tls-files.lz4",
     "migrate_v1.11.0_kubelet-new-config-files.lz4",
     "migrate_v1.11.0_ecs-additional-configurations.lz4",
 ]
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -2037,6 +2037,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "kubelet-tls-files"
+version = "0.1.0"
+dependencies = [
+ "migration-helpers",
+]
+
+[[package]]
 name = "language-tags"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -32,6 +32,7 @@ members = [
     "api/migration/migrations/v1.11.0/aws-config-settings",
     "api/migration/migrations/v1.11.0/credential-providers",
     "api/migration/migrations/v1.11.0/kubelet-tls-config",
+    "api/migration/migrations/v1.11.0/kubelet-tls-files",
     "api/migration/migrations/v1.11.0/kubelet-new-config-files",
     "api/migration/migrations/v1.11.0/ecs-additional-configurations",
 
--- a/sources/api/migration/migrations/v1.11.0/kubelet-tls-files/Cargo.toml
+++ b/sources/api/migration/migrations/v1.11.0/kubelet-tls-files/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "kubelet-tls-files"
+version = "0.1.0"
+authors = ["Erikson Tung <etung@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+# Don't rebuild crate just because of changes to README.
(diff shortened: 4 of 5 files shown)


stmcginnis (commented):
Isn't this covered by https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/migration/migrations/v1.11.0/kubelet-new-config-files/src/main.rs

etungsten:
> Isn't this covered by https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/migration/migrations/v1.11.0/kubelet-new-config-files/src/main.rs

That migration replaces the new values in this list:
https://github.com/bottlerocket-os/bottlerocket/blob/01d3a631f099b6c12098e18284d5c54cbcad0b71/sources/models/shared-defaults/kubernetes-services.toml#L1-L13

Each entry is a reference to a `configuration-files` API object. Since we added new `configuration-files`, we need to migrate them like we do for settings on downgrade. Otherwise they linger in the Bottlerocket datastore and `thar-be-settings` will choke on an unknown `configuration-files`.

stmcginnis:
Can we rename and get them added to https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/migration/migrations/v1.11.0/credential-providers/src/main.rs? Ben had asked some of the other migrations be combined so we don't have as many migrations being built.

etungsten:
> Can we rename and get them added to https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/migration/migrations/v1.11.0/credential-providers/src/main.rs? Ben had asked some of the other migrations be combined so we don't have as many migrations being built.

Since the migrations are for separate features, I think it'd be best to keep them separates in cases where we might want to revert/modify one or the other.

stmcginnis:
Sounds good. I'd just like to get @bcressey's 👀  on this since he had made the previous combining request.

bcressey:
> Since the migrations are for separate features, I think it'd be best to keep them separates in cases where we might want to revert/modify one or the other.

 > Sounds good. I'd just like to get @bcressey's 👀 on this since he had made the previous combining request.

I am fine with splitting them up. [kubelet-new-config-files](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/migration/migrations/v1.11.0/kubelet-new-config-files/src/main.rs) is special because we don't want to touch the same list twice during an upgrade, unless we're very careful or very confident about the order in which the migrations run.

stmcginnis (approved):

bcressey (approved):

PR #2558, opened 2022-11-07
Allow bootstrap containers to manage network configuration

jpculp:
**Description of changes:**

Gives bootstrap containers the ability to write files to `/var/lib/netdog`. This can serve as an alternative/overriding location to store a `net.toml`.

Also cleans up some `clippy` warnings.

**Testing done:**

Bootstrap container was able to configure network by outputting a `/var/lib/netdog/net.toml` file.

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/cli/generate_net_config.rs
+++ b/sources/api/netdog/src/cli/generate_net_config.rs
@@ -1,5 +1,8 @@
 use super::{error, Result};
-use crate::{net_config, DEFAULT_NET_CONFIG_FILE, KERNEL_CMDLINE, PRIMARY_INTERFACE};
+use crate::{
+    net_config, DEFAULT_NET_CONFIG_FILE, KERNEL_CMDLINE, OVERRIDE_NET_CONFIG_FILE,
+    PRIMARY_INTERFACE,
+};
 use argh::FromArgs;
 use snafu::{OptionExt, ResultExt};
 use std::{fs, path::Path};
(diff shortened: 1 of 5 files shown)


sources/api/netdog/src/cli/generate_net_config.rs
```diff
@@ -15,6 +17,10 @@ pub(crate) fn run() -> Result<()> {
         net_config::from_path(DEFAULT_NET_CONFIG_FILE).context(error::NetConfigParseSnafu {
             path: DEFAULT_NET_CONFIG_FILE,
         })?
+    } else if Path::exists(Path::new(ALT_NET_CONFIG_FILE)) {
+        net_config::from_path(ALT_NET_CONFIG_FILE).context(error::NetConfigParseSnafu {
+            path: ALT_NET_CONFIG_FILE,
+        })?
```
bcressey: The downside of this ordering is that it becomes impossible for a bootstrap container to override a copy of `net.toml` provided at provisioning time, or through other means.

That would create a feature gap between `metal` variants, which generally couldn't override via bootstrap container, and `aws` and `vmware` variants, which could.

It might be better to reverse the order and adjust the naming:
* `OVERRIDE_NET_CONFIG_FILE` -> `/var/lib/netdog/net.toml`
* `DEFAULT_NET_CONFIG_FILE` -> `/var/lib/bottlerocket/net.toml`

sources/api/netdog/src/wicked/mod.rs
```diff
@@ -185,9 +185,9 @@ mod tests {
                 assert_eq!(
                     expected.trim(),
                     generated,
-                    "failed test for net config version: '{}', interface: '{}'",
+                    "failed test for net config version: '{}', interface: '{:?}'",
                     version,
-                    interface.name.to_string()
+                    interface.name
```
bcressey: It's weird that clippy would suggest switching from `Display` to `Debug` format here. What was the actual warning?

sources/api/netdog/src/cli/generate_net_config.rs
zmrow: +1 to that - great idea.

sources/api/netdog/src/wicked/mod.rs
jpculp: `clippy` was upset that `interface.name` should already implement DISPLAY without being converted to string, and suggested changing it to `*interface.name` instead. That suggestion ends up triggering another error as the size can't be known at compilation time. I figured `{:?}` was slightly less ugly than adding a `clippy` exception, but I could go either way.

jpculp:
Implemented @bcressey's suggestion and switched to a `clippy` exception for the `interface.name` warning.

bcressey (approved):

zmrow (approved):
:video_game:

Issue #1501, opened 2021-04-19
models: make enums better


webern:
**What I'd like:**

In #1388 we added some enum types to the model. And, in general, settings can often be enumerations. However the `string_impls_for!` macro led us to express these enum types as structs with an inner string.

For example, this struct https://github.com/bottlerocket-os/bottlerocket/blob/0912ca54432088a12e054b293325409dd5962ede/sources/models/src/modeled_types/kubernetes.rs#L460 exists along with this enum https://github.com/bottlerocket-os/bottlerocket/blob/0912ca54432088a12e054b293325409dd5962ede/sources/models/src/modeled_types/kubernetes.rs#L466

It would be better if the enum were the only type, and the struct with `inner: String` were not required. This would involve implementing all the right impls for a `pub enum`. This could be done with a new macro, `string_impls_for_enum!`, that does the same thing as `string_impls_for!`, but for enums instead of structs with an inner string.

This would allow new types to be expressed as enums when an enum is the most natural representation of the setting.

webern:
This is blocked by #2257 
Once the blocker is done I will make our enums less wonky.

bcressey:
Resolving - as I understand it, this is now complete.

PR #2559, opened 2022-11-07
models: create scalar derive macro

webern:
**Issue number:**

Progress on #2257

**Description of changes:**

A macro named `string_impls_for` has caused us to create all validated model types like this:

```rust
struct MyModelType {
    inner: String
}

string_impls_for!(MyModelType);
```

Although it was never strictly necessary to use the macro, it has lead to some questionable patterns where we create a struct with an inner `String` even when the type we are modeling is, for example, a number or an enum.

In this PR I have written a couple of traits and a derive macro that will replace `string_impls_for`. When the macro encounters a struct like the one above (with `inner: String`), it produces basically the same code as `string_impls_for`. However when the macro encounters a struct with an inner type of anything that's not a string, it "does the right thing" and implements similar dereferencing conveniences for the inner type.

In order to limit the blast radius of this PR, only a single setting in the ECS variant is altered to use the new macro. The rest of the settings in all of the various models are left untouched for now.

Later the `string_impls_for` macro will be deleted once the new `Scalar` macro has replaced `string_impls_for` in all of the existing settings.

An additional logic branch will be added to the derive macro in a subsequent PR to gracefully handle enums, which will allow me to solve #1501

#2463 can be fixed after this PR.

**Testing done:**

Testing was done on this sha f07fc3bd

Only `aws-ecs-1` is affected:
- [x] Ran an `aws-ecs-1` instance and a simple task.
- [x] I ran `apiclient set settings.ecs.instance-attributes.good-key=good-value` and it worked (the setting was added)
- [x] I ran:
```
apiclient set settings.ecs.instance-attributes.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa=good-value-2
```
and got the expected error:
```
Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: Error during deserialization: Unable to deserialize into ECSAttributeKey: ECS attribute key given invalid input: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
- [x] I started an instance with this userdata:

```
[settings.ecs]
cluster = "x86-64-aws-ecs-1"
instance-attributes.good-key="good-value"
```

It worked fine and when I checked the settings I found good-key=good-value

- [x] I started an instance with this userdata:

```toml
[settings.ecs]
cluster = "x86-64-aws-ecs-1"
instance-attributes.bad-key-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa="good-value"
```

It failed to during the boot sequence as expected:

```
[  OK  ] Reached target Network is Online.
         Starting Bottlerocket userdata configuration system...
[    8.551491] early-boot-config[1461]: Error PATCHing '/settings?tx=bottlerocket-launch': Status 400 when PATCHing /settings?tx=bottlerocket-launch: Json deserialize error: Unable to deserialize into ECSAttributeKey: ECS attribute key given invalid input: bad-key-aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa at line 1 column 280
[DEPEND] Dependency failed for Send signal to CloudFormation Stack.
```

**Terms of contribution:**


webern:
This is ready for review

sources/models/scalar-derive/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+// The `Scalar` derive macro has made it so that we can use `Pizza` as if it were a `String`,
+// but we know that the value has been validated.
+
+let pizza = Pizza::new("pepperoni").unwrap();
+// `pizza` behaves like a string!
+assert!("pepperoni" == pizza);
+
+let err = Pizza::new("pineapple");
+// no that's gross!
```
zmrow, line 68: :rofl: 

I'm firmly in the "pineapple == good" camp.... so I'll have to respectfully disagree here :)

sources/models/scalar-derive/README.md
```diff
@@ -0,0 +1,175 @@
...
+impl std::ops::Deref for WeirdType {
+    type Target = str;
+    fn deref(&self) -> &str {
+        self.as_ref()
+    }
+}
+
+// We also need `Borrow` to treat it like a string with `as_ref_str`.
+impl core::borrow::Borrow<str> for WeirdType {
+        fn borrow(&self) -> &str {
```
zmrow: The formatting here is a bit off.

sources/models/src/modeled_types/mod.rs
```diff
@@ -150,6 +151,20 @@ pub mod error {
             source: std::num::ParseIntError,
         },
     }
+
+    /// Creates a `ValidationError` with a consistent message for strings with regex validations
+    /// where the regex is too big to display to the user.
+    pub(crate) fn big_pattern_error<S1, S2>(thing: S1, input: S2) -> ValidationError
```
zmrow, line 157: Would you expect these functions to be written for every single scalar we'd be validating?

webern: No. But there are some very frequent patterns such as "regex validation error" and "big regex validation error". I anticipate figuring out some additional code re-use for the regex validations since most have the same validation code.

webern:
^ fix formatting error found by @zmrow 

https://github.com/bottlerocket-os/bottlerocket/compare/805b33ec79c047a1984e21204a3700d7bc5e58c8..2c8d6fe769c7ef7ab899fc0521f53c351992d3e5

webern:
^ rebase - large diff

https://github.com/bottlerocket-os/bottlerocket/compare/2c8d6fe769c7ef7ab899fc0521f53c351992d3e5..4629448b147c02add8bdacb406ad179938a1d23d

zmrow (approved):
:money_with_wings:

sources/models/scalar-derive/src/lib.rs
```diff
@@ -0,0 +1,493 @@
+/*!
+
+A convenience macro for Bottlerocket model types.
+
+## Description
+
+The `Scalar` macro can be used for for strings or numbers that need to be validated in the
```
cbgbt: nit:
```suggestion
The `Scalar` macro can be used for strings or numbers that need to be validated in the
```

sources/models/scalar/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+use serde::Serialize;
+use std::error::Error;
+use std::fmt::{Display, Formatter};
+
+/// Because the `Scalar` trait has the same name as the `Scalar` derive macro, we place it and the
+/// `Validate` traits into a module named `traits`.
+pub mod traits {
+    use super::*;
+
+    /// The `Scalar` trait is used to wrap some inner, validated, value. It is intended for use  in
```
cbgbt, line 18: Really excellent docstrings on this, thanks!

sources/models/scalar-derive/src/lib.rs
```diff
@@ -0,0 +1,493 @@
...
+    let struct_info = StructInfo::new(&derive_input, settings);
+
+    // Write impls.
+    let mut ast2 = TokenStream2::new();
+    struct_info.write_impls(&mut ast2);
+    ast2.into_token_stream().into()
+}
+
+/// Store args given by the user inside `#[scalar(as_ref_str=false, ...)]` using `darling` (it's
+/// kind of like `clap` for derive macro arguments).
```
cbgbt, line 197: I hadn't seen `darling` before. Cool.

sources/models/scalar-derive/src/lib.rs
```diff
@@ -0,0 +1,493 @@
...
+    /// Whether or not we should treat the inner reference type as `&str`.
+    as_ref_str: bool,
+}
+
+impl StructInfo {
+    fn new(derive_input: &DeriveInput, settings: RawSettings) -> Self {
+        let scalar = derive_input.ident.to_string();
+
+        let (inner_field, inner_type) = match derive_input.data.clone() {
+            Data::Struct(s) => find_inner_field(s, settings.inner.as_ref().map(|s| s.as_str())),
```
cbgbt, line 231: While reading this I wondered whether or not it is right to allow a struct with more than one field. Is that really a Scalar?

I think though that e.g. a URI which could be split to store the scheme, domain, and path under separate properties would still find this useful. Probably best not to dwell.

sources/models/scalar-derive/src/lib.rs
```diff
@@ -0,0 +1,493 @@
...
+/// This function finds the "inner" field of the struct. The most common example is:
+///
+/// ```no_run
+/// struct SomeStruct {
+///     inner: String
+/// }
+/// ```
+///
+/// In the above example, this function would return the field ("inner", "String")
+fn find_inner_field(data_struct: DataStruct, field_name: Option<&str>) -> (String, String) {
```
cbgbt, line 474: It might be nice to write some test cases to validate "correct usage." I'm not sure that there's a good way to validate errors in proc macros?

sources/models/scalar-derive/src/lib.rs
webern: > While reading this I wondered whether or not it is right to allow a struct with more than one field. Is that really a Scalar?

I can think of cases where you might want more than one field. For example, you might have some static constants in the struct or something that had to be parsed and you wanted to "cache" the results of the parse during construction. They might be a little contrived, but it seemed easy enough to just allow it.

I would say, yes it's still a scalar if the other fields are internal implementation details.

sources/models/scalar-derive/src/lib.rs
webern: > It might be nice to write some test cases to validate "correct usage." I'm not sure that there's a good way to validate errors in proc macros?

The doc tests are compiling and running. Do you think that's good enough or should I add some more tests? These can be added in doc tests, a `tests` directory, or an `examples` directory. They can't be in the same crate like unit tests are.

Testing failure cases would be pretty difficult. I'm not sure you can mark a test as "shouldn't compile". The good news is that a panic in this code can only ever break a build, it can't affect Bottlerocket runtime.

webern:
^ fix repeated word typo found by @cbgbt 

https://github.com/bottlerocket-os/bottlerocket/compare/4629448b147c02add8bdacb406ad179938a1d23d..1eec4364d6225b8eb340f9d85ff348d72034fa9a

sources/models/scalar-derive/src/lib.rs
cbgbt: It might be good to have a simple examples directory, or even just doc tests over this that assert that the expected inner value is selected correctly in the additional supported cases (multiple properties of a struct, struct with unnamed properties.)

sources/models/scalar-derive/src/lib.rs
```diff
@@ -0,0 +1,514 @@
...
+        )
+    }
+
+    fn write_impls(&self, stream: &mut TokenStream2) {
+        // We need to store our information in local variables for the quote macro.
+        let scalar = format_ident!("{}", &self.scalar);
+        let inner_type = format_ident!("{}", &self.inner_type);
+        let inner_ref_type = self.inner_ref_type();
+
+        // Create the trait Scalar trait implementation, which is different based on whether the
```
cbgbt: just a nit
```suggestion
        // Create the Scalar trait implementation, which is different based on whether the
```

cbgbt (approved):
Nice work! 🚀

webern:
^ fix a typo and undo unnecessary Cargo.toml and Cargo.lock changes

https://github.com/bottlerocket-os/bottlerocket/compare/a75f62e5b42b5ea799cd84eae93242cb402d64cb..aef2d88658993e012e2c2b121330ee97bf9196d4

sources/models/scalar-derive/README.md
```diff
@@ -0,0 +1,175 @@
...
+use scalar::ValidationError;
+use scalar_derive::Scalar;
+
+// We create a struct with an inner type in a field named `inner`. We derive `Scalar`.
+#[derive(Debug, PartialEq, Scalar)]
+struct Pizza {
+    inner: String
+}
+
+// We must implement the `Validate` trait for out type.
```
arnaldo2792, line 48: I know this is closed, but there is a typo here.

sources/models/scalar-derive/README.md
```diff
@@ -0,0 +1,175 @@
...
+#[scalar(as_ref_str = true, inner = "some_field")]
+struct MyWrapper {
+    some_field: WeirdType,
+}
+
+impl Validate for MyWrapper {
+    fn validate<T: Into<WeirdType>>(input: T) -> Result<MyWrapper, ValidationError> {
+        Ok(Self{ some_field: input.into() })
+    }
+}
```
arnaldo2792, line 169: Some asserts would have been nice like in the examples above.

sources/models/scalar/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// ```
+    ///
+    pub trait Scalar {
+        /// The inner type who's value is being wrapped by the `Scalar` object. For example,
+        /// `String`, `u32`, etc.
+        ///
+        /// The traits bounds on the `Inner` type exist because the `Scalar` derive macro expects
+        /// them to be there. Placing the trait bounds here makes error messages easier to interpret
+        /// when using the `Scalar` derive macro. The macro also requires `Deserialize<'de>`, but it
+        /// it is unhelpful to add it to the trait bounds here because of the lifetime parameter.
```
arnaldo2792, line 62: There is an extra `it` here

PR #2562, opened 2022-11-08
Add Autoscaling Warm Pool Support for K8S Variants

### LINKED ISSUE #1788: Support Warm Pool for EC2 Auto Scaling!

gthao313:
**Issue number:**

Closes #1788

**Description of changes:**
Refactor PR #2322 

updated: 

- Add warm pool help program to shibaken (they both need fetch IMDS so not need for a new binary)
- Boot autoscaling warm pool part on shibaken to become a service

**Testing done:**

- [x] Unit test
- [x] Test new setting `autoscaling.should-wait=false` and `autoscaling.should-wait=false`
- [x] Migration test new setting `autoscaling.should-wait` 
- [x] Integration test on warm pool 

Integration test deatails
1. create autoscaling group
2. create warm pool on autoscaling group
3. check if the instances on warm pool don't join the eks cluster
4. set a one time autoscaling scale up
5. monitor it those instances on warm pool have been added to autoscaling group and successfully join to eks cluster

Screen Shot 2022-11-08 at 3 20 24 PM

Screen Shot 2022-11-08 at 3 20 33 PM

Screen Shot 2022-11-08 at 3 22 27 PM

Screen Shot 2022-11-08 at 3 24 13 PM

Screen Shot 2022-11-08 at 3 24 27 PM

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -579,6 +579,12 @@ For AWS variants, these settings allow you to set up CloudFormation signaling to
 #### Auto Scaling group settings
 
 * `settings.autoscaling.should-wait`: Whether to wait for the instance to reach the `InService` state before the orchestrator agent joins the cluster. Defaults to `false`. Set this to `true` only if the instance is part of an Auto Scaling group, or will be attached to one later.
+  For example:
+
+  ```toml
+  [settings.autoscaling]
+  should-wait = true
+  ```
 
 #### OCI Hooks settings
 
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -25,6 +25,7 @@ Source6: metricdog-toml
 Source7: host-ctr-toml
 Source8: oci-default-hooks-json
 Source9: cfsignal-toml
+Source10: warm-pool-wait-toml
 
 # 1xx sources: systemd units
 Source100: apiserver.service
@@ -44,6 +45,7 @@ Source116: load-kernel-modules.service.in
 Source117: cfsignal.service
 Source118: generate-network-config.service
 Source119: reboot-if-required.service
+Source120: warm-pool-wait.service
 
 # 2xx sources: tmpfilesd configs
 Source200: migration-tmpfiles.conf
@@ -239,7 +241,7 @@ Summary: Manages user-defined K8S static pods
 
 %if %{with aws_platform}
 %package -n %{_cross_os}shibaken
-Summary: Setting generator for populating admin container user-data from IMDS.
+Summary: Run tasks reliant on IMDS
 %description -n %{_cross_os}shibaken
 %{summary}.
 
@@ -415,6 +417,10 @@ install -p -m 0644 \
 %endif
 
 %if %{with aws_platform}
+%if %{with aws_k8s_family}
+install -p -m 0644 %{S:10} %{buildroot}%{_cross_templatedir}
+install -p -m 0644 %{S:120} %{buildroot}%{_cross_unitdir}
+%endif
 install -p -m 0644 %{S:9} %{buildroot}%{_cross_templatedir}
 install -p -m 0644 %{S:117} %{buildroot}%{_cross_unitdir}
 %endif
@@ -529,6 +535,11 @@ install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-stor
 %if %{with aws_platform}
 %files -n %{_cross_os}shibaken
 %{_cross_bindir}/shibaken
+%dir %{_cross_templatedir}
+%if %{with aws_k8s_family}
+%{_cross_templatedir}/warm-pool-wait-toml
+%{_cross_unitdir}/warm-pool-wait.service
+%endif
 
 %files -n %{_cross_os}cfsignal
 %{_cross_bindir}/cfsignal
--- a/packages/os/warm-pool-wait.service
+++ b/packages/os/warm-pool-wait.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Wait for the instance to be in service
+After=configured.target
+Wants=configured.target
+Before=kubelet.service
+# We only want to run once, at first boot. This file is created by shibaken
+# after a successful run.
+ConditionPathExists=!/var/lib/bottlerocket/warm-pool-wait.ran
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/shibaken warm-pool-wait
+RemainAfterExit=true
+
+[Install]
+WantedBy=multi-user.target
+RequiredBy=kubelet.service
--- a/packages/selinux-policy/fs.cil
+++ b/packages/selinux-policy/fs.cil
@@ -55,6 +55,7 @@
 (filecon "/.*/usr/bin/docker.*" file runtime_exec)
 (filecon "/.*/usr/bin/host-ctr" file runtime_exec)
 (filecon "/.*/usr/sbin/runc" file runtime_exec)
+(filecon "/.*/usr/bin/shibaken" file api_exec)
 
 ; Label local storage mounts.
 (filecon "/local" any local)
--- a/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting/src/main.rs
+++ b/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting/src/main.rs
@@ -1,0 +1,29 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting prefix for configuring autoscaling in k8s variants.
+/// Remove the whole `settings.autoscaling` prefix if we downgrade.
+fn run() -> Result<()> {
+    if cfg!(variant_family = "aws-k8s") {
+        migrate(AddPrefixesMigration(vec![
+            "settings.autoscaling",
+            "services.autoscaling-warm-pool",
+            "configuration-files.warm-pool-wait-toml",
+        ]));
+    };
+
+    Ok(())
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/api/shibaken/README.md
+++ b/sources/api/shibaken/README.md
@@ -11,6 +11,7 @@ shibaken is used to fetch data from the instance metadata service (IMDS) in AWS.
 shibaken can:
 * Fetch and populate the admin container's user-data with authorized ssh keys from the IMDS.
 * Perform boolean queries about the AWS partition in which the host is located.
+* Wait in a warm pool until the instance is marked as InService before starting the orchestrator.
 
 (The name "shibaken" comes from the fact that Shiba are small, but agile, hunting dogs.)
 
--- a/sources/api/shibaken/src/error.rs
+++ b/sources/api/shibaken/src/error.rs
@@ -1,3 +1,4 @@
+use crate::warmpool::error::WarmPoolCheckError;
 use snafu::Snafu;
 
 #[derive(Debug, Snafu)]
@@ -27,6 +28,9 @@ pub(crate) enum Error {
 
     #[snafu(display("Error serializing to JSON: {}", source))]
     SerializeJson { source: serde_json::error::Error },
+
+    #[snafu(display("Failed to check autoscaling warm pool: {}", source))]
+    WarmPoolCheckFailed { source: WarmPoolCheckError },
 }
 
 pub(crate) type Result<T> = std::result::Result<T, Error>;
--- a/sources/api/shibaken/src/main.rs
+++ b/sources/api/shibaken/src/main.rs
@@ -8,6 +8,7 @@ shibaken is used to fetch data from the instance metadata service (IMDS) in AWS.
 shibaken can:
 * Fetch and populate the admin container's user-data with authorized ssh keys from the IMDS.
 * Perform boolean queries about the AWS partition in which the host is located.
+* Wait in a warm pool until the instance is marked as InService before starting the orchestrator.
 
 (The name "shibaken" comes from the fact that Shiba are small, but agile, hunting dogs.)
 */
@@ -24,6 +25,7 @@ use crate::error::Result;
 mod admin_userdata;
 pub(crate) mod error;
 mod partition;
+mod warmpool;
 
 /// Returns information gathered from the AWS instance metadata service (IMDS).
 #[derive(FromArgs, Debug)]
@@ -45,6 +47,9 @@ enum Commands {
     /// Fetch and return whether or not this host is in the given partition.
     /// Accepts multiple partitions, returning `true` if the host is in any of the given partitions.
     IsPartition(partition::IsPartition),
+
+    /// Poll lifecycle state and wait until instance to be marked as InService
+    WarmPoolWait(warmpool::autoscaling_warm_pool::WarmPoolWait),
 }
 
 async fn run() -> Result<()> {
@@ -67,6 +72,10 @@ async fn run() -> Result<()> {
             generate_admin_userdata.run().await
         }
         Commands::IsPartition(is_partition) => is_partition.run().await,
+        Commands::WarmPoolWait(warm_pool_wait) => warm_pool_wait
+            .run()
+            .await
+            .context(error::WarmPoolCheckFailedSnafu),
     }
 }
 
--- a/sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
+++ b/sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
@@ -1,0 +1,122 @@
+/// This module contains utilities for waiting until the instance on warm pool to be
+/// marked as InService.
+use crate::warmpool::error::{self, Result};
+use argh::FromArgs;
+use imdsclient::ImdsClient;
+use serde::Deserialize;
+use snafu::{OptionExt, ResultExt};
+use std::fs;
+use std::path::Path;
+use tokio::time::{sleep, Duration};
+
+// Path to config file containing host's autoscaling.should-wait setting
+const CONFIG_PATH: &str = "/etc/warm-pool-wait.toml";
+const FETCH_LIFECYCLE_INTERVAL_SECS: Duration = Duration::from_secs(5);
+
+#[derive(FromArgs, Debug)]
+#[argh(subcommand, name = "warm-pool-wait")]
+/// Wait until the instance reaches the `InService` state
+pub(crate) struct WarmPoolWait {}
+
+impl WarmPoolWait {
+    pub(crate) async fn run(self) -> Result<()> {
+        let config_parse = get_config().await?;
+
+        let should_wait_value = config_parse.should_wait;
+        let marker_file_path = config_parse.marker_path;
+
+        println!("autoscaling.should-wait value is {}", should_wait_value);
+        if should_wait_value {
+            wait_until_inservice().await?;
+        }
+
+        fs::write(&marker_file_path, "").unwrap_or_else(|e| {
+            log::warn!("Failed to create marker file '{}', warm-pool-wait service may unexpectedly run again: '{}'",
+            &marker_file_path, e);
+        });
+        println!("Marker file path is {}", marker_file_path);
+
+        Ok(())
+    }
+}
+
+/// Continuously fetches lifecycle state of the host.
+/// Returns from function when host is "InService"
+async fn wait_until_inservice() -> Result<()> {
+    let mut client = ImdsClient::new();
+
+    loop {
+        let lifecycle_state = client
+            .fetch_autoscaling_lifecycle_state()
+            .await
+            .context(error::ImdsRequestSnafu)?;
+
+        if let Some(lifecycle_state) = lifecycle_state {
+            if lifecycle_state == "InService" {
+                println!("Lifecycle state is {} ... exiting.", lifecycle_state);
+                return Ok(());
+            }
+        }
+
+        sleep(FETCH_LIFECYCLE_INTERVAL_SECS).await;
+        continue;
+    }
+}
+
+/// Parses config file for autoscaling.should-wait setting.
+async fn get_config() -> Result<Config> {
+    Config::from_file(CONFIG_PATH)
+}
+
+#[derive(Debug, Deserialize)]
+struct Config {
+    should_wait: bool,
+    marker_path: String,
+}
+
+impl Config {
+    /// Parses configuration file at passed in path and returns struct containing settings value.
+    fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let path = path.as_ref();
+        let string_parse = fs::read_to_string(path).context(error::ConfigReadSnafu { path })?;
+        let config: Config =
+            toml::from_str(&string_parse).context(error::ConfigParseSnafu { path })?;
+        Ok(config)
+    }
+}
+
+#[cfg(test)]
+mod config_test {
+    use super::Config;
+    use tempfile::TempDir;
+
+    // Example config file where user sets autoscaling.should-wait to false.
+    const FALSE_SHOULD_WAIT: &str = r#"
+    should_wait = false
+    marker_path = "/var/lib/bottlerocket/warm-pool-wait.ran"
+    "#;
+
+    // Example config file where user sets autoscaling.should-wait to true.
+    const TRUE_SHOULD_WAIT: &str = r#"
+    should_wait = true
+    marker_path = "/var/lib/bottlerocket/warm-pool-wait.ran"
+    "#;
+
+    #[test]
+    fn false_should_wait_test() {
+        let dir = TempDir::new().unwrap();
+        let path = dir.path().join("config.toml");
+        std::fs::write(&path, FALSE_SHOULD_WAIT).unwrap();
+        let config = Config::from_file(&path).unwrap();
+        assert!(!config.should_wait);
+    }
+
+    #[test]
+    fn true_should_wait_test() {
+        let dir = TempDir::new().unwrap();
+        let path = dir.path().join("config.toml");
+        std::fs::write(&path, TRUE_SHOULD_WAIT).unwrap();
+        let config = Config::from_file(&path).unwrap();
+        assert!(config.should_wait);
+    }
+}
--- a/sources/imdsclient/src/lib.rs
+++ b/sources/imdsclient/src/lib.rs
@@ -168,6 +168,12 @@ impl ImdsClient {
         self.fetch_string(&instance_type_target).await
     }
 
+    /// Get lifecycle state from instance metadata.
+    pub async fn fetch_autoscaling_lifecycle_state(&mut self) -> Result<Option<String>> {
+        let instance_type_target = "meta-data/autoscaling/target-lifecycle-state";
+        self.fetch_string(&instance_type_target).await
+    }
+
     /// Returns a list of public ssh keys skipping any keys that do not start with 'ssh'.
     pub async fn fetch_public_ssh_keys(&mut self) -> Result<Option<Vec<String>>> {
         info!("Fetching list of available public keys from IMDS");
--- a/sources/models/shared-defaults/aws-autoscaling.toml
+++ b/sources/models/shared-defaults/aws-autoscaling.toml
@@ -1,0 +1,3 @@
+# Autoscaling warm pool support
+[settings.autoscaling]
+should-wait = false
--- a/sources/models/shared-defaults/kubernetes-aws.toml
+++ b/sources/models/shared-defaults/kubernetes-aws.toml
@@ -21,3 +21,14 @@ service-checks = ["apiserver", "chronyd", "containerd", "host-containerd", "kube
 
 [metadata.settings.network]
 affected-services = ["containerd", "kubernetes", "host-containerd", "host-containers"]
+
+[services.autoscaling-warm-pool]
+configuration-files = ["warm-pool-wait-toml"]
+restart-commands = []
+
+[configuration-files.warm-pool-wait-toml]
+path = "/etc/warm-pool-wait.toml"
+template-path = "/usr/share/templates/warm-pool-wait-toml"
+
+[metadata.settings.autoscaling]
+affected-services = ["autoscaling-warm-pool"]
--- a/Release.toml
+++ b/Release.toml
@@ -173,4 +173,6 @@ version = "1.12.0"
 "(1.11.0, 1.11.1)" = []
 "(1.11.1, 1.12.0)" = [
     "migrate_v1.12.0_k8s-private-pki-path.lz4",
+    "migrate_v1.12.0_add-k8s-autoscaling-warm-pool-setting.lz4",
+    "migrate_v1.12.0_add-k8s-autoscaling-warm-pool-setting-metadata.lz4",
 ]
--- a/packages/os/warm-pool-wait-toml
+++ b/packages/os/warm-pool-wait-toml
@@ -1,0 +1,2 @@
+should_wait = {{settings.autoscaling.should-wait}}
+marker_path = "/var/lib/bottlerocket/warm-pool-wait.ran"
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -3295,7 +3295,9 @@ dependencies = [
  "serde_json",
  "simplelog",
  "snafu",
+ "tempfile",
  "tokio",
+ "toml",
 ]
 
 [[package]]
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -217,6 +217,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "add-k8s-autoscaling-warm-pool-setting"
+version = "0.1.0"
+dependencies = [
+ "bottlerocket-variant",
+ "migration-helpers",
+]
+
+[[package]]
+name = "add-k8s-autoscaling-warm-pool-setting-metadata"
+version = "0.1.0"
+dependencies = [
+ "bottlerocket-variant",
+ "migration-helpers",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -27,6 +27,8 @@ members = [
     # "api/migration/migrations/vX.Y.Z/..."
     # (all previous migrations archived; add new ones after this line)
     "api/migration/migrations/v1.12.0/k8s-private-pki-path",
+    "api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting",
+    "api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting-metadata",
 
     "bottlerocket-release",
 
--- a/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting-metadata/Cargo.toml
+++ b/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting-metadata/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "add-k8s-autoscaling-warm-pool-setting-metadata"
+version = "0.1.0"
+authors = ["Tianhao Geng <tianhg@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2018"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers", version = "0.1.0"}
+
+[build-dependencies]
+bottlerocket-variant = { version = "0.1", path = "../../../../../bottlerocket-variant" }
--- a/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting-metadata/build.rs
+++ b/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting-metadata/build.rs
@@ -1,0 +1,6 @@
+use bottlerocket_variant::Variant;
+
+fn main() {
+    let variant = Variant::from_env().unwrap();
+    variant.emit_cfgs();
+}
--- a/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting-metadata/src/main.rs
+++ b/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting-metadata/src/main.rs
@@ -1,0 +1,27 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::common_migrations::{AddMetadataMigration, SettingMetadata};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting and `affected-services` metadata for `settings.autoscaling`
+fn run() -> Result<()> {
+    if cfg!(variant_family = "aws-k8s") {
+        migrate(AddMetadataMigration(&[SettingMetadata {
+            metadata: &["affected-services"],
+            setting: "settings.autoscaling",
+        }]));
+    };
+
+    Ok(())
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting/Cargo.toml
+++ b/sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "add-k8s-autoscaling-warm-pool-setting"
+version = "0.1.0"
+authors = ["Tianhao Geng <tianhg@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2018"
+publish = false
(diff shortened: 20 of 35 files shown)


packages/os/autoscaling-warm-pool.service
```diff
@@ -0,0 +1,17 @@
...
+Requires=settings-applier.service
+Wanted=network-online.target
+Before=kubelet.service
+# We only want to run once, at first boot. This file is created by shibaken
+# after a successful run.
+ConditionPathExists=!/var/lib/bottlerocket/shibaken.ran
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/shibaken boot-shibaken
```
jpculp: I'm not sure that we want to have `autoscaling-warm-pool` as the name of the service running `shibaken`'s `boot-shibaken` service mode. There should be a parent `shibaken` service instead, or the ExecStart should be something like `/usr/bin/shibaken is_warmpool_healthy`.

sources/imdsclient/src/lib.rs
```diff
@@ -168,6 +168,12 @@ impl ImdsClient {
         self.fetch_string(&instance_type_target).await
     }
 
+    /// Get lifecycle state from instance metadata.
+    pub async fn fetch_lifecycle_state(&mut self) -> Result<Option<String>> {
```
jpculp: Are there other lifecycles this could conflict with? Should this be `fetch_autoscaling_lifecycle_state` instead?

packages/os/autoscaling-warm-pool.service
```diff
@@ -0,0 +1,17 @@
...
+RemainAfterExit=true
+
+[Install]
+WantedBy=multi-user.target
```
jpculp:
```suggestion
WantedBy=multi-user.target

```
A couple files are missing the end-of-file newline.

sources/api/shibaken/src/boot_shibaken.rs
```diff
@@ -0,0 +1,45 @@
+/// This module contains mechanism for support Bottlerocket refraining from joining a Kubernetes cluster
```
jpculp: This seems too targeted and would make it awkward to expand functionality in the future. I would recommend making boot shibaken a generic subcommand within `main.rs` that just iterates through a list of self-contained non-settings-generator subcommands. All the warmpool stuff would be better contained in `IsAutoscalingWarmPoolHealthy`.

jpculp (dismissed):

packages/os/autoscaling-warm-pool.service
bcressey: I'd expect something like:
* unit name - `warm-pool-wait.service`
* unit description - "Wait for the instance to be in service"
* `shibaken` subcommand - `shibaken warm-pool-wait`

packages/os/autoscaling-warm-pool.service
bcressey: To be clear, we just want one newline at the end of the file, not two newlines (which is what shows in @jpculp's suggestion.) This may be a configuration option in your text editor that needs to be set.

packages/os/os.spec
```diff
@@ -404,6 +406,7 @@ install -p -m 0644 \
   %{S:100} %{S:101} %{S:102} %{S:103} %{S:105} \
   %{S:106} %{S:107} %{S:110} %{S:111} %{S:112} \
   %{S:113} %{S:114} %{S:118} %{S:119} %{S:120} \
+  %{S:121} \
```
bcressey: This needs to be conditional:
```suggestion
```

packages/os/os.spec
```diff
@@ -417,7 +420,9 @@ install -p -m 0644 \
 
 %if %{with aws_platform}
 install -p -m 0644 %{S:9} %{buildroot}%{_cross_templatedir}
+install -p -m 0644 %{S:10} %{buildroot}%{_cross_templatedir}
 install -p -m 0644 %{S:117} %{buildroot}%{_cross_unitdir}
+install -p -m 0644 %{S:121} %{buildroot}%{_cross_unitdir}
 %endif
```
bcressey: I suppose I can get behind this implementation for `aws-ecs-1` as well, although it's somewhat redundant there since there's an agent-level configuration in that variant.

packages/os/os.spec
```diff
@@ -25,6 +25,7 @@ Source6: metricdog-toml
 Source7: host-ctr-toml
 Source8: oci-default-hooks-json
 Source9: cfsignal-toml
+Source10: autoscaling-warm-pool-k8s-toml
```
bcressey: This doesn't really have anything to do with Kubernetes, so the config file shouldn't have `k8s` in the name.

packages/os/autoscaling-warm-pool.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Run tasks reliant on IMDS
+After=network-online.target settings-applier.service
+Requires=settings-applier.service
+Wanted=network-online.target
+Before=kubelet.service
```
bcressey: `kubelet.service` needs to take the dependency on the "wait for in service" service, not the other way around - this service provides more general functionality.

packages/os/autoscaling-warm-pool-k8s-toml
```diff
@@ -0,0 +1,2 @@
+should_wait = {{settings.autoscaling.should-wait}}
+marker_path = "/var/lib/bottlerocket/shibaken.ran"
```
bcressey:
```suggestion
marker_path = "/var/lib/bottlerocket/warm-pool-wait.ran"
```

packages/os/autoscaling-warm-pool.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Run tasks reliant on IMDS
+After=network-online.target settings-applier.service
+Requires=settings-applier.service
+Wanted=network-online.target
```
bcressey: These dependencies can just be:
```suggestion
After=configured.target
Wants=configured.target
```

packages/runc/Cargo.lock
```diff
@@ -0,0 +1,22 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 3
+
+[[package]]
+name = "glibc"
+version = "0.1.0"
+
```
bcressey: This file should be removed; you end up with these lock files after running `cargo make` in a package directory, like `packages/runc`, rather than at the root of the repo.

sources/api/shibaken/README.md
```diff
@@ -11,6 +11,8 @@ shibaken is used to fetch data from the instance metadata service (IMDS) in AWS.
 shibaken can:
 * Fetch and populate the admin container's user-data with authorized ssh keys from the IMDS.
 * Perform boolean queries about the AWS partition in which the host is located.
+* Support Bottlerocket refraining from joining a Kubernetes cluster if it is warming up in an autoscaling
```
bcressey: This needs to be phrased in terms of a general feature - "wait until the instance is marked as `InService` before continuing" or something.

sources/api/shibaken/src/boot_shibaken.rs
```diff
@@ -0,0 +1,45 @@
...
+
+        let should_wait_value = config_parse.should_wait;
+        let marker_file_path = config_parse.marker_path;
+
+        info!("autoscaling.should-wait value is {}", should_wait_value);
+        if should_wait_value {
+            let a = warmpool::autoscaling_warm_pool::IsAutoscalingWarmPoolHealthy {};
+            a.run()
+                .await
+                .context(error::AutoscalingWarmPoolFailedSnafu)?;
```
bcressey: The naming of this is awkward; it's not really a check about whether the ASG warm pool is healthy, it's about whether the instance itself is ready to enter service.

sources/api/shibaken/src/main.rs
```diff
@@ -45,6 +49,13 @@ enum Commands {
     /// Fetch and return whether or not this host is in the given partition.
     /// Accepts multiple partitions, returning `true` if the host is in any of the given partitions.
     IsPartition(partition::IsPartition),
+
+    /// Fetch to IMDS to check if the instance on autoscaling warm pool is InService.
+    IsAutoscalingWarmPoolHealthy(warmpool::autoscaling_warm_pool::IsAutoscalingWarmPoolHealthy),
+
+    /// Support Bottlerocket refraining from joining a Kubernetes cluster if it's warming up in an autoscaling
+    /// warm pool until it is signaled to join the cluster.
+    BootShibaken(boot_shibaken::BootShibaken),
```
bcressey: Why two commands for this? I would just do it all in a single command. If there's a need to have a one-off check for status, you could take an argument, e.g.:
* `shibaken in-service-status`
* `shibaken in-service-status --wait`

But if you only need the "wait" behavior, `shibaken warm-pool-wait` is more succinct.

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,60 @@
...
+    let mut client = ImdsClient::new();
+    let mut lifecycle_state = get_lifecycle_state(&mut client).await?;
+
+    info!("Lifecycle state is {}", lifecycle_state);
+
+    while lifecycle_state.ne("InService") {
+        lifecycle_state = get_lifecycle_state(&mut client).await?;
+        info!("Lifecycle state is {} ... waiting.", lifecycle_state);
+        sleep(Duration::from_secs(FETCH_LIFECYCLE_INTERVAL)).await;
+    }
```
bcressey: This imposes an overhead of a five second wait even if the instance is already in service.

sources/api/shibaken/src/warmpool/config.rs
```diff
@@ -0,0 +1,58 @@
...
+impl Config {
+    /// Parses configuration file at passed in path and returns struct containing settings value.
+    pub(crate) fn from_file<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let path = path.as_ref();
+        let string_parse = fs::read_to_string(path).context(error::ConfigReadSnafu { path })?;
+        let config: Config =
+            toml::from_str(&string_parse).context(error::ConfigParseSnafu { path })?;
+        Ok(config)
+    }
+}
```
bcressey: I suppose it's a matter of taste but this is a pretty small module - I would personally just include the struct and tests in the same module where it's used.

sources/api/shibaken/src/warmpool/test_config.toml
```diff
@@ -0,0 +1,2 @@
+should_wait = true
+marker_path = "/var/lib/bottlerocket/shibaken.ran"
```
bcressey: Why does this file need to be checked in if the unit tests are writing a temporary copy of these configs?

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -21,3 +21,11 @@ service-checks = ["apiserver", "chronyd", "containerd", "host-containerd", "kube
 
 [metadata.settings.network]
 affected-services = ["containerd", "kubernetes", "host-containerd", "host-containers"]
+
+# AutoScaling Warm Pool
+[settings.autoscaling]
+should-wait = false
```
bcressey: Can we make sure all `aws` variants have `AutoscalingSettings` in their model, and just move this to the common `aws` settings instead?

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -21,3 +21,11 @@ service-checks = ["apiserver", "chronyd", "containerd", "host-containerd", "kube
...
+
+[configuration-files.shibaken-toml]
+path = "/etc/autoscaling-warm-pool-k8s.toml"
+template-path = "/usr/share/templates/autoscaling-warm-pool-k8s-toml"
```
bcressey: You need a migration to add & remove configuration files.

sources/models/shared-defaults/kubernetes-aws.toml
gthao313: >Can we make sure all aws variants have AutoscalingSettings in their model

Sure

>and just move this to the common aws settings instead

Do you mean do like this `settings.aws.autoscaling.should-wait`. The reason why we do `settings.autoscaling.should-wait` is that one of our customers contributed to Bottlerocket to support “warm pools” for ECS clusters. They used the setting `settings.autoscaling.should-wait` for ECS, so we think we should use the same one as ECS.

sources/api/shibaken/src/warmpool/test_config.toml
gthao313: Yeah. You're correct! I probably accidentally added this to the commit. thanks

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
gthao313: Thanks! I'll use `loop {}` instead of `while`.

packages/os/autoscaling-warm-pool.service
gthao313: I think It should be 
```
After=network-online.target configured.target
Wants=network-online.target configured.target
```

`warm-pool-wait.service` needs wait `network` to be ready and then be able to fetch instance.

sources/api/shibaken/src/main.rs
```diff
@@ -45,6 +47,9 @@ enum Commands {
     /// Fetch and return whether or not this host is in the given partition.
     /// Accepts multiple partitions, returning `true` if the host is in any of the given partitions.
     IsPartition(partition::IsPartition),
+
+    /// Fetch to IMDS and wait until instance to be marked as InService.
```
jpculp: It might be better to say something like "Poll lifecycle state and wait until instance to be marked as `InService`".

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+use log::{info, warn};
+use serde::Deserialize;
+use snafu::{OptionExt, ResultExt};
+use std::fs;
+use std::path::Path;
+use tokio::time::{sleep, Duration};
+
+// Path to config file containing host's autoscaling.should-wait setting
+const CONFIG_PATH: &str = "/etc/warm-pool-wait.toml";
+const FETCH_LIFECYCLE_INTERVAL: u64 = 5;
```
jpculp: It would be nice to rename this as `FETCH_LIFECYCLE_INTERVAL_SECS` to give the `5` more context. Another option would be to make the constant a `Duration` type.

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,132 @@
+/// This module contains utilities for waiting until the instance on warm pool to be
+/// marked as InService.
+use crate::warmpool::error::{self, Result};
+use argh::FromArgs;
+use imdsclient::ImdsClient;
+use log::{info, warn};
```
jpculp: One thing to be aware of, since `shibaken` was originally only a settings generator all of it's logs go to STDERR. With `WarmPoolWait` getting called outside of `sundog`, you may want to have your info messages redirect to STDOUT.

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -28,4 +28,4 @@ should-wait = false
 
 [configuration-files.warm-pool-wait-toml]
 path = "/etc/warm-pool-wait.toml"
-template-path = "/usr/share/templates/warm-pool-wait-toml"
```
jpculp: nit. should this be in the previous commit?

packages/os/autoscaling-warm-pool.service
arnaldo2792: `configured.target` depends on `preconfigured.target` which requires many services that depend on`network-online.target`, so you can safely do what @bcressey suggested :+1:

packages/selinux-policy/fs.cil
```diff
@@ -55,6 +55,7 @@
 (filecon "/.*/usr/bin/docker.*" file runtime_exec)
```
arnaldo2792, line 55: Nit:

If the imdsclient change is completely independent from the shibaken change, it could be in its own commit. Otherwise, you can format the commit message as follows:

```
shibaken, imdsclient: query instance lifecycle for warmpool support
```

Or something along those lines.

sources/api/shibaken/src/main.rs
```diff
@@ -45,19 +50,33 @@ enum Commands {
...
+                .build(),
+            TerminalMode::Stderr,
+            ColorChoice::Auto,
+        ),
+        // Simple logger for subcommands not called by sundog
+        SimpleLogger::new(
+            args.log_level,
+            ConfigBuilder::new()
+                .add_filter_allow_str("warmpool")
+                .build(),
```
arnaldo2792: Do you need this filter? Or will `SimpleLogger` take the filters from other loggers in the resultant logger?

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+use std::path::Path;
+use tokio::time::{sleep, Duration};
+
+// Path to config file containing host's autoscaling.should-wait setting
+const CONFIG_PATH: &str = "/etc/warm-pool-wait.toml";
+const FETCH_LIFECYCLE_INTERVAL_SECS: Duration = Duration::from_secs(5);
+
+#[derive(FromArgs, Debug)]
+#[argh(subcommand, name = "warm-pool-wait")]
+/// Fetch to IMDS and wait until instance is marked as InService.
```
arnaldo2792: Nit:

I would say that `Fetch to IMDS` is irrelevant here:

```suggestion
Wait until the instance reaches the `InService` state
```

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+            wait_until_inservice().await?;
+        }
+
+        info!("Marker file path is {}", marker_file_path);
+        fs::write(&marker_file_path, "").unwrap_or_else(|e| {
+            warn!(
+                "Failed to create marker file '{}', shibaken service may unexpectedly run again: '{}'",
+                &marker_file_path, e
+            )
+        });
```
arnaldo2792: It could be confusing while debugging if this is a message logged in the service logs could be confusing since the systemd service doesn't  mention `shibaken`.

packages/os/autoscaling-warm-pool.service
gthao313: We discuss offline, and we think we still need add `network-online.target` here.

sources/api/shibaken/src/main.rs
```diff
@@ -45,19 +50,33 @@ enum Commands {
...
 
-    // TerminalMode::Stderr will send all logs to stderr, as sundog only expects the json output of
-    // the setting on stdout.
-    TermLogger::init(
-        args.log_level,
-        LogConfig::default(),
-        TerminalMode::Stderr,
-        ColorChoice::Auto,
-    )
+    CombinedLogger::init(vec![
```
jpculp: Nice!

sources/models/shared-defaults/kubernetes-aws.toml
arnaldo2792: It seems like we missed "nesting" the settings under `aws` when `autoscaling.should-wait` was introduced in https://github.com/bottlerocket-os/bottlerocket/pull/2063. I agree with moving `autoscaling.should-wait` under `aws`, to make it clear that this API only works for AWS variants. Having it out of `aws` could make users think that the API works for VMware or metal variants.

sources/api/shibaken/src/error.rs
```diff
@@ -27,6 +28,9 @@ pub(crate) enum Error {
 
     #[snafu(display("Error serializing to JSON: {}", source))]
     SerializeJson { source: serde_json::error::Error },
+
+    #[snafu(display("Failed to check autoscaling warm pool: {}", source))]
+    AutoscalingWarmPoolFailed { source: AutoscalingWarmPoolError },
```
arnaldo2792: Nit:

```suggestion
    WarmPoolCheckFailed { source: WarmPoolCheckError },
```

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    loop {
+        let lifecycle_state = get_lifecycle_state(&mut client).await?;
+
+        if lifecycle_state.eq("InService") {
+            info!("Lifecycle state is {} ... exiting.", lifecycle_state);
+            break;
+        } else {
+            info!("Lifecycle state is {} ... waiting.", lifecycle_state);
+            sleep(FETCH_LIFECYCLE_INTERVAL_SECS).await;
```
arnaldo2792: Doesn't this return a `Future` that must be consumed like `await?` as in the line above?

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+const CONFIG_PATH: &str = "/etc/warm-pool-wait.toml";
+const FETCH_LIFECYCLE_INTERVAL_SECS: Duration = Duration::from_secs(5);
+
+#[derive(FromArgs, Debug)]
+#[argh(subcommand, name = "warm-pool-wait")]
+/// Fetch to IMDS and wait until instance is marked as InService.
+pub(crate) struct WarmPoolWait {}
+
+impl WarmPoolWait {
+    pub(crate) async fn run(self) -> Result<()> {
```
arnaldo2792, line 22: Nit:

You can make the config path and the sleep interval parameters of the command.

sources/api/migration/migrations/v1.12.0/add-eks-autoscaling/src/main.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting prefix for configuring autoscaling in k8s variants.
+/// Remove the whole `settings.autoscaling` prefix if we downgrade.
+fn run() -> Result<()> {
+    if cfg!(variant_runtime = "k8s") {
+        migrate(AddPrefixesMigration(vec![
```
arnaldo2792: This will change once the setting is migrated under `aws`

packages/kubernetes-1.21/kubelet.service
```diff
@@ -1,8 +1,9 @@
 [Unit]
 Description=Kubelet
 Documentation=https://github.com/kubernetes/kubernetes
-After=containerd.service configured.target
+After=containerd.service configured.target warm-pool-wait.service
```
arnaldo2792: Nit:

```suggestion
After=configured.target containerd.service warm-pool-wait.service
```

packages/os/os.spec
arnaldo2792: I think we should just use the ECS feature for ECS variants. There will be an additional IMDS call in the agent even if the instance is in `InService` which could increase "readyness" time.

sources/api/shibaken/src/main.rs
gthao313: Shibaken was set up to send all logs to STDERR, including non-errors since it's called by sundog. Warm pool code isn't called by sundog, so we need log the normal way of STDOUT and STDERR.

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
gthao313: Not need here due to `the ? operator can only be applied to values that implement Try`. I believe it's safe here and we have [example](https://github.com/bottlerocket-os/bottlerocket/blob/67e5abfb0e1575fb6bcb7a710cdb3d422ef12283/sources/api/apiclient/src/update.rs#L252)

gthao313:
Push above address comments.

major changes:
- systemd: only build warm pool service on aws k8s family.

Extra test on ECS variant 

Launched Bottlerocket instance using builded ami with user-data:

```
[settings.autoscaling]
should-wait = true
```

verified that the `WarmPoolsSupport` value of file /etc/ecs/ecs.config.json  is `true`
```
{"Cluster":"bottlerocket",
 "InstanceAttributes":{"bottlerocket.variant":"aws-ecs-1"},
 "PrivilegedDisabled":true,"AvailableLoggingDrivers":["json-file","awslogs","none"],
 "WarmPoolsSupport":true,
 "ImagePullBehavior":0,
 "TaskIAMRoleEnabled":true,
 "TaskIAMRoleEnabledForNetworkHost":true,
 "SELinuxCapable":true,
 "OverrideAWSLogsExecutionRole":true,
 "TaskENIEnabled":true,
 "GPUSupportEnabled":false }
```

sources/api/shibaken/src/main.rs
cbgbt: I wonder if just using `println!` in the warm pool code would be less brittle than relying on the logging filters? In the future, we'll need to remember to update these strings if we add or change subcommands.

EDIT: To clarify, logging and diagnostics pretty typically get sent to stderr in Linux programs regardless, but user-directed output or results usually end up in stdout.

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+    Ok(())
+}
+
+/// Parses config file for aws.autoscaling.should-wait setting.
+pub(crate) async fn get_config() -> Result<Config> {
+    Config::from_file(CONFIG_PATH)
+}
+
+#[derive(Debug, Deserialize)]
+pub(crate) struct Config {
```
cbgbt: nit: These seem like they should be private.

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+/// Uses an imdsclient function to fetch the lifecyclestate of the host
+async fn get_lifecycle_state(client: &mut ImdsClient) -> Result<String> {
+    client
+        .fetch_autoscaling_lifecycle_state()
+        .await
+        .context(error::ImdsRequestSnafu)?
+        .context(error::ImdsNoneSnafu {
+            what: "instance-id",
```
cbgbt: Should this be `what: "target-lifecycle-state"`?

cbgbt:
I updated [my comment about logging filters and `stdout`, `stderr`.](https://github.com/bottlerocket-os/bottlerocket/pull/2562#discussion_r1063066294)

gthao313:
push above fix readme typo

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,129 @@
...
+        println!("autoscaling.should-wait value is {}", should_wait_value);
+        if should_wait_value {
+            wait_until_inservice().await?;
+        }
+
+        println!("Marker file path is {}", marker_file_path);
+        fs::write(&marker_file_path, "").unwrap_or_else(|e| {
+            println!("Failed to create marker file '{}', warm-pool-wait service may unexpectedly run again: '{}'",
+            &marker_file_path, e);
+        });
```
cbgbt: nit: I'd change the inner `println!` back to `warn!` so that it appears on `stderr`, and move the first `println!` to after this statement so that it only prints if the marker file is successfully written.

cbgbt (approved):

README.md
```diff
@@ -578,7 +578,13 @@ For AWS variants, these settings allow you to set up CloudFormation signaling to
 
 #### Auto Scaling group settings
 
-* `settings.autoscaling.should-wait`: Whether to wait for the instance to reach the `InService` state before the orchestrator agent joins the cluster. Defaults to `false`. Set this to `true` only if the instance is part of an Auto Scaling group, or will be attached to one later.
+* `settings.autoscaling.should-wait`: Whether to wait for the instance to reach the `InService` state before the orchestrator agent joins the cluster.Defaults to `false`. Set this to `true` only if the instance is part of an Auto Scaling group, or will be attached to one later.
```
```suggestion
* `settings.autoscaling.should-wait`: Whether to wait for the instance to reach the `InService` state before the orchestrator agent joins the cluster. Defaults to `false`. Set this to `true` only if the instance is part of an Auto Scaling group, or will be attached to one later.
```

packages/kubernetes-1.21/kubelet.service
```diff
@@ -1,8 +1,9 @@
 [Unit]
 Description=Kubelet
 Documentation=https://github.com/kubernetes/kubernetes
-After=containerd.service configured.target
+After=configured.target containerd.service warm-pool-wait.service
 Wants=configured.target
+Requires=warm-pool-wait.service
```
bcressey: One drawback of this is that `kubelet.service` will have these non-existent dependencies on other platforms (vmware, metal). It might be better to handle this in `warm-pool-wait.service` to avoid that and to keep it self-contained.

That would also avoid a mid-air collision with #2699 which is adding another `kubelet.service` file.

EDIT: I actually gave you the opposite advice [in this comment](https://github.com/bottlerocket-os/bottlerocket/pull/2562#discussion_r1020406485) so let me think about that a little more.

Right now `warm-pool-wait.service` is specific to `aws-k8s-*` variants. The most precise way to add these dependencies would be to conditionally include symlinks like these in `os.spec`:
```
%if %{with aws_platform}
%files -n %{_cross_os}shibaken
%{_cross_bindir}/shibaken

# include warm-pool-wait for `aws-*`
%{_cross_templatedir}/warm-pool-wait-toml
%{_cross_unitdir}/warm-pool-wait.service
%endif

%if %{with aws_k8s_family}
# add a "requires" symlink to the kubelet service
%dir %{_cross_unitdir}/kubelet.service.requires
%{_cross_unitdir}/kubelet.service.requires/warm-pool-wait.service
%endif

%if %{with aws_nomad_family}
# add a "requires" symlink to the hypothetical nomad service
%dir %{_cross_unitdir}/nomad.service.requires
%{_cross_unitdir}/nomad.service.requires/warm-pool-wait.service
%endif
```

However, that's a lot of fuss to avoid a condition that systemd handles just fine (ignoring non-existent services). I'd be fine if you want to ignore my previous suggestion (sorry!) and put the `Before` + `RequiredBy` dependencies in `warm-pool-wait.service` instead.

If it ever becomes more general (not specific to `aws-k8s-*` we can look at a different approach, but the simpler method seems better for now.

sources/models/src/aws-ecs-1-nvidia/defaults.d/80-aws-autoscaling.toml
```diff
@@ -0,0 +1 @@
+../../../shared-defaults/aws-autoscaling.toml
```
bcressey: These links would be better in the 20-29 range, along with `20-aws-host-containers.toml` and `25-cf-signal.toml`. Maybe `26-aws-autoscaling`.

packages/os/warm-pool-wait.service
```diff
@@ -0,0 +1,15 @@
...
+# after a successful run.
+ConditionPathExists=!/var/lib/bottlerocket/warm-pool-wait.ran
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/shibaken warm-pool-wait
+RemainAfterExit=true
+
+[Install]
+WantedBy=multi-user.target
```
bcressey: If you take the above suggestion and don't touch `kubelet.service`:

```suggestion
[Unit]
Description=Wait for the instance to be in service
After=configured.target
Wants=configured.target
Before=kubelet.service
# We only want to run once, at first boot. This file is created by shibaken
# after a successful run.
ConditionPathExists=!/var/lib/bottlerocket/warm-pool-wait.ran

[Service]
Type=oneshot
ExecStart=/usr/bin/shibaken warm-pool-wait
RemainAfterExit=true

[Install]
WantedBy=multi-user.target
RequiredBy=kubelet.service
```

(You still don't need to depend on `network-online.target`, it's implied by `configured.target`.)

bcressey (commented):
LGTM. Some nice-to-have suggestions, but no blockers.

packages/os/warm-pool-wait.service
gthao313: I think we may still need `network-online.target`.  I consistently have this error when we don't depend on `network-online.target` here. I think we mean to make  the `configured.target` depend on `network-online.target`, but it seems the systemd doesn't "wait" for the entire graph of dependencies to start a service. 

```
× warm-pool-wait.service - Wait for the instance to be in service
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/warm-pool-wait.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Tue 2023-01-10 02:43:21 UTC; 4min 33s ago
    Process: 1276 ExecStart=/usr/bin/shibaken warm-pool-wait (code=exited, status=1/FAILURE)
   Main PID: 1276 (code=exited, status=1/FAILURE)

Jan 10 02:43:21 ip-192-168-7-249.us-west-2.compute.internal systemd[1]: Starting Wait for the instance to be in service...
Jan 10 02:43:21 ip-192-168-7-249.us-west-2.compute.internal shibaken[1276]: 02:43:21 [INFO] shibaken started
Jan 10 02:43:21 ip-192-168-7-249.us-west-2.compute.internal shibaken[1276]: autoscaling.should-wait value is true
Jan 10 02:43:21 ip-192-168-7-249.us-west-2.compute.internal shibaken[1276]: Failed to check autoscaling warm pool: IMDS request failed: No 'target-lifecycle-state' found
Jan 10 02:43:21 ip-192-168-7-249.us-west-2.compute.internal systemd[1]: warm-pool-wait.service: Main process exited, code=exited, status=1/FAILURE
Jan 10 02:43:21 ip-192-168-7-249.us-west-2.compute.internal systemd[1]: warm-pool-wait.service: Failed with result 'exit-code'.
Jan 10 02:43:21 ip-192-168-7-249.us-west-2.compute.internal systemd[1]: Failed to start Wait for the instance to be in service.
```

bcressey: Hmm, the dependencies of `early-boot-config.service` alone should be enough to push `preconfigured.target` (and thus `configured.target`) beyond `network-online.target`. I am 99.99% confident here; if I'm wrong then we have bigger issues to fix. 😀 

I don't know why the additional dependency would make a difference here, but my hunch is that the correct fix would be to retry until `target-lifecycle-state` doesn't return a 404, and only then check the returned value.

Otherwise it could be that the dependency is somehow introducing enough of a delay that we "win" a race with a behind-the-scenes update to IMDS data served to the instance.

sources/models/shared-defaults/aws-autoscaling.toml
```diff
@@ -0,0 +1,3 @@
+# Autoscaling warm pool support
+[settings.autoscaling]
```
webern, line 2: The name of this file is no longer correct, right? The setting was moved out of `aws` so all platforms want this default now?

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,129 @@
...
+    let mut client = ImdsClient::new();
+
+    loop {
+        let lifecycle_state = get_lifecycle_state(&mut client).await?;
+
+        if lifecycle_state.eq("InService") {
+            println!("Lifecycle state is {} ... exiting.", lifecycle_state);
+            break;
+        } else {
+            println!("Lifecycle state is {} ... waiting.", lifecycle_state);
```
webern: Couldn't this loop go on for a long time? Do we want to print so many of these? Should we print one every minute instead of 20 times per minute?

sources/api/shibaken/src/main.rs
```diff
@@ -8,6 +8,7 @@ shibaken is used to fetch data from the instance metadata service (IMDS) in AWS.
 shibaken can:
 * Fetch and populate the admin container's user-data with authorized ssh keys from the IMDS.
 * Perform boolean queries about the AWS partition in which the host is located.
+* Wait until the instance on warm pool to be marked as InService before joining a Kubernetes cluster
```
webern: Future proofing

```suggestion
* Wait in a warm pool until the instance is marked as InService before starting the orchestrator
```

sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+use std::process;
+
+/// We added a new setting prefix for configuring autoscaling in k8s variants.
+/// Remove the whole `settings.autoscaling` prefix if we downgrade.
+fn run() -> Result<()> {
+    if cfg!(variant_runtime = "k8s") {
+        migrate(AddPrefixesMigration(vec![
+            "settings.autoscaling",
+            "configuration-files.warm-pool-wait-toml",
+            "warm-pool-wait.service",
```
webern: This doesn't seem right.

Does `"configuration-files.warm-pool-wait-toml"` require a metadata migration?
This doesn't look like what I'm used to seeing in a prefix migration `"warm-pool-wait.service"`

Unfortunately I don't know the right answers here.

sources/models/shared-defaults/aws-autoscaling.toml
bcressey: All `aws` variants should have this, but it's `settings.autoscaling`, not `settings.aws.autoscaling`. `vmware` and `metal` variants won't have this.

sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool/src/main.rs
arnaldo2792: FWIW, there are archived migrations where we do something similar:

https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/migration/migrations/archived/v1.6.2/add-cfsignal/src/main.rs#L13

arnaldo2792 (commented):
LGTM, provided that the systemd service is updated accordingly with @bcressey suggestion regarding the dependency on `network.target`

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
gthao313: This loop goes on for a very short time.

gthao313:
Push above address comments.

major changes:
- autoscaling-should-wait setting: add "service” to link the settings and configuration-files.
- warm-pool-wait.service: warm-pool-wait.service only depends on `configured.target` and kubelet.service relies on it.
- shibaken(autoscaling warm pool): add logic on warm pool helper program that retry until target-lifecycle-state doesn't return a 404. 

tests:
- migration test on EKS and ECS both.

sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting/src/main.rs
```diff
@@ -0,0 +1,25 @@
...
+
+/// We added a new setting prefix for configuring autoscaling in k8s variants.
+/// Remove the whole `settings.autoscaling` prefix if we downgrade.
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec![
+        "settings.autoscaling",
+        "services.autoscaling-warm-pool",
+        "configuration-files.warm-pool-wait-toml",
+    ]))
+}
```
bcressey: This needs to be conditionalized on the `aws-k8s` family, otherwise it will run on downgrade for `aws-ecs-1` and remove `settings.autoscaling` (which has been in that variant since 1.8.0.

`migrations/archived/v1.8.0/pki-affected-services` has an example of doing this; there are two parts, one in `build.rs` to set the variables, and another in `main.rs` to consume them.

It shouldn't hurt to do the same to the `metadata` migration but it's not strictly necessary since `aws-ecs-1` doesn't have that metadata.

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,147 @@
...
+            .fetch_autoscaling_lifecycle_state()
+            .await
+            .context(error::ImdsRequestSnafu)?
+        {
+            Some(target_lifecycle_state) => return Ok(target_lifecycle_state),
+            None => {
+                sleep(FETCH_IMDS_SLEEP_DURATION).await;
+                attempts += 1;
+                continue;
+            }
```
bcressey: I'd prefer to have this function return a `Result<Option<String>>` and have the caller handle the waiting and retries - otherwise we're stacking up a lot of retry logic. There's already some retrying going on in `imdsclient`.

The caller can handle checking whether it's `None` (from a 404) or the desired string.

sources/api/shibaken/src/warmpool/autoscaling_warm_pool.rs
```diff
@@ -0,0 +1,146 @@
...
+async fn get_lifecycle_state(client: &mut ImdsClient) -> Result<Option<String>> {
+    client
+        .fetch_autoscaling_lifecycle_state()
+        .await
+        .context(error::ImdsRequestSnafu)
+}
+
+/// Continuously fetches lifecycle state of the host.
+/// Returns from function when host is "InService"
+async fn wait_until_inservice() -> Result<()> {
```
bcressey, line 45: This still seems more complicated than what you need. If we're willing to wait forever for the state to reach "InService", it seems like we should also be willing to wait forever for IMDS not to return 404.

Accordingly, what I had in mind was:
```
/// Continuously fetches lifecycle state of the host.
/// Returns from function when host is "InService"
async fn wait_until_inservice() -> Result<()> {
    let mut client = ImdsClient::new();

    loop {
        let lifecycle_state = client
            .fetch_autoscaling_lifecycle_state()
            .await
            .context(error::ImdsRequestSnafu)?;
        if let Some(lifecycle_state) = lifecycle_state {
            if lifecycle_state == "InService" {
                println!("Lifecycle state is {} ... exiting.", lifecycle_state);
                return Ok(());
            }
        }
        sleep(FETCH_LIFECYCLE_INTERVAL_SECS).await;
        continue;
    }
}
```

Here's the [git diff](https://gist.github.com/bcressey/2964cfaa2af04bbc8f5b02fb93570de0) output from my edits if that's useful.

bcressey (approved):

cbgbt (approved):

PR #2566, opened 2022-11-09
pluto: limit EKS API call timeout

### LINKED ISSUE #2370: Limit `pluto`'s EKS API request retries/timeout when describing the cluster
**What I'd like:**
`pluto` to not block settings generation for 30 minutes when retrying requests to the EKS API when retrieving the cluster's network configuration.

Currently, a Bottlerocket host running in a network-isolated environment will hang during boot if `settings.kubernetes.cluster-dns-ip` is not already set in user-data on first boot. `sundog` will kick-off `pluto` to try and generate that setting. If there is no network access to the EKS API, `pluto` (via AWS Rust SDK) will retry requests for 30 minutes before giving up and falling back to a default value. See https://github.com/bottlerocket-os/bottlerocket/blob/037d0e2798a7e8d4a12d0fb35d3c266db3d74f25/sources/api/pluto/src/main.rs#L166-L172.

What we want is to limit the retry timeout somehow via https://docs.rs/aws-sdk-eks/latest/aws_sdk_eks/struct.RetryConfig.html.

Ideally we will match the timeout for our IMDS request retries, which is currently around 5 minutes.

N/A

etungsten:
**Issue number:**

Resolves https://github.com/bottlerocket-os/bottlerocket/issues/2370

**Description of changes:**
```
    pluto: limit EKS API call timeout to 5 minutes
    
    We don't want pluto to block settings generation for too long during
    boot. This limits the retry duration for the EKS describe cluster call
    to 5 minutes which matches the IMDS retry duration of 5 minutes
    elsewhere. If the API call times out, pluto will fallback to using
    default values for the cluster-dns-ip.

```

**Testing done:**
Built aws-k8s-1.23 and launched instance in a network-restricted environment.
Verified that pluto timeouts after 5 minutes when trying to describe cluster and returns a default value based on the CIDR of the instance's primary network interface.
```
bash-5.1# time pluto cluster-dns-ip
"10.100.0.10"

real    5m0.011s
user    0m0.000s
sys     0m0.009s

```

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/eks.rs
+++ b/sources/api/pluto/src/eks.rs
@@ -6,6 +6,10 @@ use hyper_proxy::{Proxy, ProxyConnector};
 use hyper_rustls::HttpsConnectorBuilder;
 use snafu::{OptionExt, ResultExt, Snafu};
 use std::env;
+use std::time::Duration;
+
+// Limit the timeout for the EKS describe cluster API call to 5 minutes
+const EKS_DESCRIBE_CLUSTER_TIMEOUT: Duration = Duration::from_secs(300);
 
 pub(crate) type ClusterNetworkConfig = KubernetesNetworkConfigResponse;
 
@@ -15,6 +19,9 @@ pub(super) enum Error {
     DescribeCluster {
         source: aws_sdk_eks::types::SdkError<aws_sdk_eks::error::DescribeClusterError>,
     },
+
+    #[snafu(display("Timed-out waiting for EKS Describe Cluster API response: {}", source))]
+    DescribeClusterTimeout { source: tokio::time::error::Elapsed },
 
     #[snafu(display("Missing field '{}' EKS response", field))]
     Missing { field: &'static str },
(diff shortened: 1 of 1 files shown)


webern (commented):
This looks good, but wouldn't we want it to be much shorter?

Does the situation actually resolve sometimes after, say 30 seconds or 60 seconds? Is there a benefit to waiting another 4+ minutes?

etungsten:
> This looks good, but wouldn't we want it to be much shorter?
> 
> Does the situation actually resolve sometimes after, say 30 seconds or 60 seconds? Is there a benefit to waiting another 4+ minutes?

That's a good question. I based the duration on `imdsclient`'s retry timeout, which in turn is based on wicked's interface-up timeout value. 
https://github.com/bottlerocket-os/bottlerocket/blob/f81dfac3bac980f723c84a260090fe67fc4856ce/sources/imdsclient/src/lib.rs#L38-L39
It's not clear to me why it needs to be the same as wicked ifup's time out duration since sundog doesn't run until `network.target` is reached. 

In any case, whatever duration we decide here should probably be propagated to the duration in `imdsclient` since they're all part of sundog's call-stack during settings generation.

bcressey (approved):

zmrow (approved):
:fuelpump:

webern (approved):
Approving. Five minutes seems to long to me, but it is definitely better than forever.

PR #2569, opened 2022-11-10
Update kernels to v5.10.157 and v5.15.79

foersleo:
**Issue number:** n/a

Closes #

**Description of changes:** This rebases the Bottlerocket kernels in preparation for the 1.10 release. The new kernels are based on 5.10.149 and 5.15.75, respectively.

**Testing done:**

## Initial update that uncovered a lot of changes in the config:

Report from `tools/diff-kernel-config`:

```
config-aarch64-5.10-aws-k8s-1.23-diff:    1 removed,   4 added,   5 changed
config-aarch64-5.15-aws-dev-diff:       353 removed,  17 added, 104 changed
config-aarch64-5.15-metal-dev-diff:     354 removed,  17 added, 102 changed
config-x86_64-5.10-aws-k8s-1.23-diff:     0 removed,   3 added,   2 changed
config-x86_64-5.10-metal-k8s-1.23-diff:   0 removed,   3 added,   2 changed
config-x86_64-5.15-aws-dev-diff:        261 removed,  14 added, 116 changed
config-x86_64-5.15-metal-dev-diff:      262 removed,  14 added, 114 changed
```

The full diff can be found [in this gist](https://gist.github.com/foersleo/e317aafac7b931afb7f9c429af2a1ae5).

## Current config diff with changes and analysis as detailed in comments below

Report from `tools/diff-kernel-config`:

```
config-aarch64-5.10-aws-k8s-1.23-diff:    1 removed,   4 added,   5 changed
config-aarch64-5.15-aws-dev-diff:       339 removed,  17 added, 103 changed
config-aarch64-5.15-metal-dev-diff:     338 removed,  28 added, 102 changed
config-x86_64-5.10-aws-k8s-1.23-diff:     0 removed,   3 added,   2 changed
config-x86_64-5.10-metal-k8s-1.23-diff:   0 removed,   3 added,   2 changed
config-x86_64-5.15-aws-dev-diff:        247 removed,  14 added, 115 changed
config-x86_64-5.15-metal-dev-diff:      235 removed,  14 added, 117 changed
```

The full diff can be found [in this gist](https://gist.github.com/foersleo/b19194123efb5837546f8954a785918f)

**Terms of contribution:**


foersleo:
Finished reviewing the config differences in our upstream for x86 (just going by the differences in `config-x86_64` from the srpms).

A lot of cleanup is happening which is good for us (e.g. deactivation of the media subsystem providing drivers for webcams and video codecs).

There is also a group of removal of filesystems and partitioning schemes. Most of which should be fine, however `squashfs` is removed as well, which we rely on at least for providing kernel headers.

There is some additional subsystems that get jiggled around (removal of `kgdb`, changing the panic behavior in most cases of hangs and lockups) which need further evaluation on if we need them.

Will work with AL (our upstream) to discuss the remaining points and getting access to the commit history to get some further context.

foersleo:
So this took quite some time to go through.

The force push rebased this branch on the current develop and added some fixups to the configs that I deemed necessary. Fine below discussion of the changes that were made to the config.

For the 5.10 kernel I think we are mostly good with the exception of a back ported driver that was added by Amazon Linux. As that driver code is quite a bit of code I went ahead and deactivated the driver so it does not get build in commit `kernel-5.10: disable SCSI_MPI3MR module` of this series. The driver was not shipped with 5.10 before and is not shipped with 5.15 either, where it is part of the upstream tree. So keeping it out of our kernels as long as there is no ask for it.

For the 5.15 kernel the story was a lot more elaborate as it has lots of changes:

```
config-aarch64-5.15-aws-dev-diff:       353 removed,  17 added, 104 changed
config-aarch64-5.15-metal-dev-diff:     354 removed,  17 added, 102 changed
config-x86_64-5.15-aws-dev-diff:        261 removed,  14 added, 116 changed
config-x86_64-5.15-metal-dev-diff:      262 removed,  14 added, 114 changed
```
Unfortunately the context for some of the changes was hard to come by or not easily available at all. I have tried to group all the options by their connection and those smaller groups into bigger topics to make reviewing easier. I then went ahead and sliced these groups again as follows.

### Changes that I deem fine to go in:

These are mostly cleanup or removal of drivers for hardware that is old or not in the target range for Bottlerocket at the moment. A full CSV of options, their sub group and which variant the are applicable to can be found [in this gist (summary_fine.csv)](https://gist.github.com/foersleo/b917f38fe80c5d093ce7c3677ccf97c1). (For easier viewing I suggest to download the list and view with `cat summary_fine.csv | column -t -s, | less -S` or similar).

The groups in there are:

* **removal of non-supported ARM products:** we used to carry lots of drivers for specific ARM platforms we do not target -> fine to remove
* **removal of the `MEDIA` subsystem:** This includes lots of drivers for different webcams and similar. We do not need that in bottlerocket
* **removal of debug and testing facilities:** This includes specialized modules to create dummy devices or debug the DMA subsystem as well as KGDB. None of that seems to be valuable in the generic variants of Bottlerocket
* **removal of unused drivers and features:** This includes lots of drivers for mostly outdated (and sometimes specialized) hardware like FPGAs, out-of-production NICs, etc. as well as partitioning schemes and older protocol implementations
* **general config options:** is somewhat weirdly named, but includes settign the kernel up to panic in lockup situations as well as changes to config options that result in upstream changes
* **added features:** includes some smaller added features, like support for core scheduling on x86 and additional crypto/compression algorithm implementations
* **hardening:** configures the kernel for signature validation on loading a kernel on kexec, increasing the size of the memory area not mmap-able by userspace, restricts the readability of kernel dmesg and restrict the setid family of syscalls
* **cleanup:** includes cleanups like removal of unused initrd compression algorithms, cleanup of some unusable netfilter modules and moving some minor options from builtin to module or vice versa
* **interfaces to CPU features:** This removes some legacy interfaces to certain CPU features as well as some interfaces for `cpuid` and `msr`through `/dev`.

Overall this section has the bulk of the changes we needed to review:

```
config-aarch64-5.15-aws-dev-diff:       316 removed,  16 added, 72 changed
config-aarch64-5.15-metal-dev-diff:     316 removed,  16 added, 72 changed
config-x86_64-5.15-aws-dev-diff:        203 removed,  11 added, 77 changed
config-x86_64-5.15-metal-dev-diff:      203 removed,  11 added, 77 changed
```

### Changes that I think are fine as they are, but would like to have a discussion on:

A full CSV of options, their sub group and which variant the are applicable to can be found [in this gist(summary_possibly_fine.csv](https://gist.github.com/foersleo/63dc27c84dc0bc553e38db66dfa845f5). (For easier viewing I suggest to download the list and view with `cat summary_possibly_fine.csv | column -t -s, | less -S` or similar).

* **Glitched removal of PS2 options:** Amazon Linux seems to have removed most options regarding PS2 drivers, however some that were not build before seem to have snuck back in due to defaults. Those can possible be removed at some point.
* **Aligning all AL kernels to the same 100Hz tick frequency:** This one makes also some part of the changes in 5.10 kernels. I do not expect a huge impact for general cases. I was not able to find reasoning for this other than aligning all kernels.
* **Removal of ACL feature for btrfs:** This seems to be a glitch of removing btrfs completely and re-enabling it again overlooking this option. I am not sure if this is something we would use in Bottlerocket.
* **Lower the maximum number of supported CPUs:** Set the maximum supported number of CPUs to 512. I guess this is ok at the moment, given the target audience of Bottlerocket. This has an impact on how some of the handling of per cpu data structures is done. I do not expect this to be a problem.

This section has a small section of the changes we needed to review:

```
config-aarch64-5.15-aws-dev-diff:       0 removed,  1 added, 16 changed
config-aarch64-5.15-metal-dev-diff:     0 removed,  1 added, 16 changed
config-x86_64-5.15-aws-dev-diff:        1 removed,  3 added, 20 changed
config-x86_64-5.15-metal-dev-diff:      1 removed,  3 added, 20 changed
```

### Changes that I am not sure about and would like to have a discussion on:

A full CSV of options, their sub group and which variant the are applicable to can be found [in this gist (summary_unsure.csv)](https://gist.github.com/foersleo/5a41ab3d5f4f5edb95f792b0992d8eb9). (For easier viewing I suggest to download the list and view with `cat summary_unsure.csv | column -t -s, | less -S` or similar).

* **removal of the precision time protocol** PTP is used to do high precision synchronization between multiple machines in a cluster. I guess we are fine removing it, but we I do not know if any of our users relies on it. 
* **removal of hardware watchdog timers** Drivers supporting some hardware watchdog timers. This is most likely not relevant for Bottlerocket and would only make sense in bare metal use cases.
* **Disabling Kernel Samepage merging:** This one I feel might be something that could come in handy for saving memory on nodes running multiple containers with the same image. However, I am not sure if we really need this and want to keep it enabled.
* **Disable 842 compression algorithm:** I do not think we use this compression algorithm today, so we should be fine to remove it.
* **remove support for EDD** BIOS enhanced disk drive calls might help determine the boot disk. I do not think we need to add this one. If we would need this it should only be applicable to bare metal use cases.
* **Fibre channel support** We have had SCSI fibre channel support disabled for most of the applicable network devices. This would remove the whole fibre channel support from Bottlerocket. I think we should be fine for now and can re-enable if someone needs it.
* **iSCSI target mode** This really depends if we see Bottlerocket as a viable distro for iSCSI targets (storage server side). I would assume we are good removing this, but would like a second opinion.
* **iSCSI initiator side** I could see a Bottlerocket system to be the user of iSCSI storage, but I am not sure.


This section has the following stats:

```
config-aarch64-5.15-aws-dev-diff:       21 removed,  0 added, 13 changed
config-aarch64-5.15-metal-dev-diff:     22 removed,  0 added, 11 changed
config-x86_64-5.15-aws-dev-diff:        24 removed,  0 added, 15 changed
config-x86_64-5.15-metal-dev-diff:      25 removed,  0 added, 13 changed
```


### Changes that I am sure about and the series contains adjustments:

A full CSV of options, their sub group and which variant the are applicable to can be found [in this gist (summary_sure.csv)](https://gist.github.com/foersleo/d864a9b259cdc3239cf38344e1a5cf11). (For easier viewing I suggest to download the list and view with `cat summary_sure.csv | column -t -s, | less -S` or similar).

* **removal of LSI SAS controllers:** We want to retain support for these SAS controllers. I have hence re-enabled the driver with commit `kernel-5.15: metal: Enable mpt3sas driver` in this series. We do not need to add the mpt2sas driver explicitly as that option is an alias for mpt3sas for compatibility reasons.
* **removal of QLogic NIC support:** We want to ship the drivers for 10G+ NICs by that vendor in their base configuration as we have done for other network interface vendors on metal variants. Commit `kernel-5.15: metal: adjust settings for QLogic NICs` takes care of this.
* **removal of squashfs support:** We rely on squashfs to provide our kernel headers in Bottlerocket. Re-enabled with `kernel-5.15: Re-enable squashfs support`

This section has the following stats (before my patches):

```
config-aarch64-5.15-aws-dev-diff:       16 removed,  0 added, 3 changed
config-aarch64-5.15-metal-dev-diff:     16 removed,  0 added, 3 changed
config-x86_64-5.15-aws-dev-diff:        33 removed,  0 added, 4 changed
config-x86_64-5.15-metal-dev-diff:      33 removed,  0 added, 4 changed
```

foersleo:
There is already newer kernels available. After spending so much time on the analysis I did not want to complicate things further. I am currently preparing a follow up which will update to the latest kernels available from the repos on top of this one. As far as I have seen that one will have a lot less change in the config and should be easier and quicker to review.

packages/kernel-5.15/config-bottlerocket-metal
```diff
@@ -94,11 +94,27 @@ CONFIG_SFC_SRIOV=y
...
+# CONFIG_QLCNIC_DCB is not set
+# CONFIG_QLCNIC_HWMON is not set
+# CONFIG_NETXEN_NIC is not set
+
 # Cisco UCS HBA support
 CONFIG_FCOE_FNIC=m
 CONFIG_SCSI_SNIC=m
 
 # LSI Logic's SAS based RAID controllers
+CONFIG_SCSI_MPT3SAS=m
```
bcressey: Seems like we'd need this to be `y` to support booting from a disk behind one of these controllers, even if that's a change from how it was previously.

bcressey (commented):
#### "fine"
> removal of unused drivers and features

re: `VIRTIO_CONSOLE`, `HVC_DRIVER` -  We don't publish PV AMIs so dropping HVC seems fine. Would we want to keep virtio console around for QEMU?

re: `HW_RANDOM_VIRTIO` - I believe my QEMU runner scripts make use of via these arguments:
```
  -object rng-random,filename=/dev/urandom,id=rng0 \
  -device virtio-rng-pci,rng=rng0,max-bytes=1024,period=1000 \
```

If the virtio interfaces are the "modern" devices for use with QEMU, it seems like we should keep them.

> hardening

re: `KEXEC_SIG` - can you make sure the new kexec signature verification options don't break kdump support?

I am not sure if there's a potential performance impact from the `IMA_DEFAULT_HASH` change. Rather than digging into that, it might be better to backlog an issue to see if we should disable `IMA` entirely. It doesn't seem like we're doing anything useful with it now, in the absence of a TPM and an IMA policy, and if we're hashing files unnecessarily that could slow down boot.

re: `SAFESETID` - I'm fine with enabling this, but we should backlog an issue to see about creating a policy for it.

***
#### "possibly fine"
> removal of PS2 options

If we don't need these at all, even on the x86 metal side, it might be good to get rid of them in a separate commit. I changed some of the i8042 related options to modules to get a slight boot speed increase - they were causing a delay before mounting the root filesystem - and I've observed a similar delay after systemd is started, though it's not consistent.

> Aligning all AL kernels to the same 100Hz tick frequency

Not an expert on this, but it makes sense to align them.

> Removal of ACL feature for btrfs

I'd tend to want to bring this back to avoid withdrawing functionality someone might be using, though I doubt there's much btrfs usage today. It's not supported in the EBS CSI driver, but the vSphere CSI driver appears to support it.

> Lower the maximum number of supported CPUs

+1

***
#### "unsure"
> removal of the precision time protocol 
> removal of hardware watchdog timers
> Disable 842 compression algorithm
> Fibre channel support

+1

> Kernel same-page merging

As I understand it, this wouldn't really help with containers currently since all processes in the containers would need to cooperate by marking memory regions as mergeable. Binaries and shared libraries in common would already be shared by the page cache.

> Remove support for EDD

This can be useful for installers (e.g. Anaconda) that need to determine the correct BIOS boot disk from a number of otherwise identical, inconsistently named devices. I agree that we don't need it in Bottlerocket.

> iSCSI target mode
> iSCSI initiator side

We've had a few requests for the initiator side (e.g. #2570), so we should keep that. It could be useful even in AWS (for "Amazon FSx for NetApp ONTAP").

I could imagine the target side being useful as well for deploying something like a Ceph / Rook iSCSI storage solution into a cluster.

foersleo:
> #### "fine"
> [...]
> re: `VIRTIO_CONSOLE`, `HVC_DRIVER` - We don't publish PV AMIs so dropping HVC seems fine. Would we want to keep virtio console around for QEMU?
> 
> re: `HW_RANDOM_VIRTIO` - I believe my QEMU runner scripts make use of via these arguments
> [...]

This makes sense. I was looking at this only from the host and not from the guest perspective. Re-reading it we would want this for the qemu case. The `HVC_DRIVER` one is selected by `VIRTIO_CONSOLE`, so we will get that as by-catch.

> > hardening
> 
> re: `KEXEC_SIG` - can you make sure the new kexec signature verification options don't break kdump support?

I got a qemu testhost to kdump and write out the `vmcore.dump`, `dmesg.log`, and `prairiedof.log` to `/var/log/kdump` successfully by injecting a crash through `echo c > /proc/sysrq-trigger`. Our documentation states kdump is currently only enabled on vmware for which I do not have a test setup yet. Do we expect this to work differently there? Is our documentation current in that regard?

> I am not sure if there's a potential performance impact from the `IMA_DEFAULT_HASH` change. Rather than digging into that, it might be better to backlog an issue to see if we should disable `IMA` entirely. It doesn't seem like we're doing anything useful with it now, in the absence of a TPM and an IMA policy, and if we're hashing files unnecessarily that could slow down boot.

Right, looks like we are currently not benefiting from it. However, with it not being enabled through the kernel cmdline we should be fine without touching it further right now. Opened https://github.com/bottlerocket-os/bottlerocket/issues/2707 for tracking further research into this at a later time.

> re: `SAFESETID` - I'm fine with enabling this, but we should backlog an issue to see about creating a policy for it.

Agreed, opened backlog issue: https://github.com/bottlerocket-os/bottlerocket/issues/2708

> #### "possibly fine"
> 
> > removal of PS2 options
> 
> If we don't need these at all, even on the x86 metal side, it might be good to get rid of them in a separate commit. I changed some of the i8042 related options to modules to get a slight boot speed increase - they were causing a delay before mounting the root filesystem - and I've observed a similar delay after systemd is started, though it's not consistent.

Looking into this further I get why we see those now. We are explicitly setting `CONFIG_MOUSE_PS2` in [config_bottlerocket:L93](https://github.com/bottlerocket-os/bottlerocket/blob/e44527ce646e12dcdf31e6a158cbc1785857000a/packages/kernel-5.15/config-bottlerocket#L93). The AL linux config used to set that as well and have all the sub-options for specific models disabled. With those now gone from the AL config they get set to their default. Will disable them to keep the status quo.

> [...]
> > Removal of ACL feature for btrfs
> 
> I'd tend to want to bring this back to avoid withdrawing functionality someone might be using, though I doubt there's much btrfs usage today. It's not supported in the EBS CSI driver, but the vSphere CSI driver appears to support it.

Makes sense. Will re-enable it to keep the status quo.

> [...]
> #### "unsure"
> [...]
> > Kernel same-page merging
> 
> As I understand it, this wouldn't really help with containers currently since all processes in the containers would need to cooperate by marking memory regions as mergeable. Binaries and shared libraries in common would already be shared by the page cache.

That makes sense. So leaving it as changed by the ingestion from AL.

>[...] 
> > iSCSI target mode
> > iSCSI initiator side
> 
> We've had a few requests for the initiator side (e.g. #2570), so we should keep that. It could be useful even in AWS (for "Amazon FSx for NetApp ONTAP").
> 
> I could imagine the target side being useful as well for deploying something like a Ceph / Rook iSCSI storage solution into a cluster.

Let me put in changes to keep the status quo on these and leave further work for the linked request issues.

foersleo:
Updated the PR with ben's latest comments and rebased at the same time, which led to some additional delta in the comparison. 

All the relevant changes are in commits `kernl-5.15: metal: Enable mpt3sas driver` and onwards. The other commits have not been touched.

The rebase to the last versions published by Amazon Linux does not have any interesting config changes (a couple of new options introduced in linux upstream that should not be of concern to us). Is there anything against piggy backing that change on top of this PR?

foersleo:
Config diff for the latest changes over the previous state of the PR:

```
config-aarch64-5.15-aws-dev-diff:         1 removed,   1 added,  10 changed    
config-aarch64-5.15-metal-dev-diff:       1 removed,   5 added,  15 changed    
config-x86_64-5.15-aws-dev-diff:          1 removed,   0 added,  11 changed
config-x86_64-5.15-metal-dev-diff:        1 removed,   4 added,  16 changed
```

Full list can be found [in this gist](https://gist.github.com/foersleo/f484a8db394cb94d9a2d1c390358bf2a)

foersleo:
Added the additional kernel update to latest state on top of the last state after discussing with @markusboehme .

Changes on top what was presented previously:

```
config-aarch64-5.10-aws-k8s-1.23-diff:    0 removed,   3 added,   0 changed
config-aarch64-5.15-aws-dev-diff:         0 removed,   1 added,   0 changed
config-aarch64-5.15-metal-dev-diff:       0 removed,   1 added,   0 changed
config-x86_64-5.10-aws-k8s-1.23-diff:     0 removed,   2 added,   0 changed
config-x86_64-5.10-metal-k8s-1.23-diff:   0 removed,   2 added,   0 changed
config-x86_64-5.15-aws-dev-diff:          0 removed,   0 added,   0 changed
config-x86_64-5.15-metal-dev-diff:        0 removed,   0 added,   0 changed
```

Full diff-report in [this gist](https://gist.github.com/foersleo/211f3ef26a307d05265867a4675dfbdf)

Nothing interesting in here other than inclusions as part of upstream changes.

foersleo:
Forgot to upload the new kernel srpms to the lookaside cache. Currently fixing and will re-trigger the checks once done.

packages/kernel-5.15/config-bottlerocket
```diff
@@ -91,6 +109,21 @@ CONFIG_MODULE_COMPRESS_XZ=y
 CONFIG_SERIO_I8042=m
 CONFIG_KEYBOARD_ATKBD=m
 CONFIG_MOUSE_PS2=m
```
markusboehme, line 111: As I understand it, this config option originally was `=y` in the Amazon Linux kernel and now has been unset. @bcressey changed this to a module to speed up the boot when this was `=y`. We could potentially drop the code entirely now.

packages/kernel-5.15/config-bottlerocket-metal
```diff
@@ -94,12 +94,37 @@ CONFIG_SFC_SRIOV=y
...
 CONFIG_SCSI_SMARTPQI=y
+
+# provide minimal iscsi via tcp support for initiator and target mode
+# initiator side
+CONFIG_ISCSI_TCP=m
+CONFIG_ISCSI_BOOT_SYSFS=m
+CONFIG_SCSI_ISCSI_ATTRS=m
+# target side
+CONFIG_ISCSI_TARGET=m
+# CONFIG_INFINIBAND_ISERT is not set
```
markusboehme, line 130: iSCSI support originally was available to all variants, now only to the metal variants. If we decide it e.g. to support "Amazon FSx for NetApp ONTAP" as Ben mentioned, this should move into the generic config fragment.

markusboehme (commented):
Filtering through those must have been a huge pile of work. Thanks for your effort in grouping, slicing and dicing the configs that made the review so much less of a burden than it otherwise would have been!

I left some comments on the diff, a couple in this very comment. Overall the change looks good to me. Before approving I'd like to have the kdump question settled. Also I'm currently leaning towards following up with another patch to move iSCSI support to the common config fragment.

>> re: KEXEC_SIG - can you make sure the new kexec signature verification options don't break kdump support?
>
> I got a qemu testhost to kdump and write out the `vmcore.dump`, `dmesg.log`, and `prairiedof.log` to `/var/log/kdump` successfully by injecting a crash through `echo c > /proc/sysrq-trigger`. Our documentation states kdump is currently only enabled on vmware for which I do not have a test setup yet. Do we expect this to work differently there? Is our documentation current in that regard?

I cannot speak to the question regarding this being exclusive to VMware, but want to point out that as I understand it signature verification is enforced on aarch64 but not on x86_64. If this is not exclusive to the VMware variants, we should make sure to test both architectures. (I'm also wondering why the config differs between the architectures here.)

> Changes that I am not sure about and would like to have a discussion on:
>
> removal of the precision time protocol PTP is used to do high precision synchronization between multiple machines in a cluster. I guess we are fine removing it, but we I do not know if any of our users relies on it.

While PTP can be used to sync time between host and guests in virtualized environments, I don't think there should be any users this change would inconvenience. NTP would interfere with PTP, and there's no Bottlerocket setting to turn off NTP at the moment.

bcressey:
> I got a qemu testhost to kdump and write out the vmcore.dump, dmesg.log, and prairiedof.log to /var/log/kdump successfully by injecting a crash through echo c > /proc/sysrq-trigger. Our documentation states kdump is currently only enabled on vmware for which I do not have a test setup yet. Do we expect this to work differently there? Is our documentation current in that regard?

If it works for QEMU it should also work in VMware; the code path isn't platform-specific.

bcressey:
> Filtering through those must have been a huge pile of work. Thanks for your effort in grouping, slicing and dicing the configs that made the review so much less of a burden than it otherwise would have been!

+1 - very nicely organized, @foersleo. I definitely appreciate the effort you put in to make this approachable to others, and it shows the depth and thoroughness of your own review.

bcressey (approved):
Nothing else beyond what @markusboehme tagged, so this LGTM once he's happy.

markusboehme:
>>> re: KEXEC_SIG - can you make sure the new kexec signature verification options don't break kdump support?
>>
>> I got a qemu testhost to kdump and write out the vmcore.dump, dmesg.log, and prairiedof.log to /var/log/kdump successfully by injecting a crash through echo c > /proc/sysrq-trigger. Our documentation states kdump is currently only enabled on vmware for which I do not have a test setup yet. Do we expect this to work differently there? Is our documentation current in that regard?
>
> I cannot speak to the question regarding this being exclusive to VMware, but want to point out that as I understand it signature verification is enforced on aarch64 but not on x86_64. If this is not exclusive to the VMware variants, we should make sure to test both architectures. (I'm also wondering why the config differs between the architectures here.)

I just tested this on x86_64 and aarch64, and got crash dumps in both cases, so we're all good here.

markusboehme (approved):

markusboehme:
Merging as agreed with Leo to get the kernel update in. I'll follow up with the iSCSI changes.

PR #2573, opened 2022-11-12
host-ctr: Unmask `/sys/firmware` from host containers

jpculp:
**Description of changes:**

This provides extra insight into the hardware of the underlying Bottlerocket host, such as the number of CPU sockets on aarch64 variants.

_NOTE: List of masked paths come's from containerd's [`oci/spec.go`](https://github.com/containerd/containerd/blob/6906b57c721f9114377ceb069662b196876915c0/oci/spec.go#L164)._

**Testing done:**

Built and launched an aarch64 k8s variant on a `c6g.large` and ran `lscpu` from the control container.

```
Architecture:           aarch64
  CPU op-mode(s):       32-bit, 64-bit
  Byte Order:           Little Endian
CPU(s):                 2
  On-line CPU(s) list:  0,1
Vendor ID:              ARM
  Model name:           Neoverse-N1
    Model:              1
    Thread(s) per core: 1
    Core(s) per socket: 2
    Socket(s):          1
    Stepping:           r3p1
    BogoMIPS:           243.75
    Flags:              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid
                         asimdrdm lrcpc dcpop asimddp ssbs
```

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -288,6 +288,20 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 			oci.WithHostNamespace(runtimespec.NetworkNamespace),
 			oci.WithHostHostsFile,
 			oci.WithHostResolvconf,
+			// Unmask `/sys/firmware` by passing an alternate list of masked paths
+			// List is based on the DefaultUnixSpec's MaskedPaths for Linux
+			// (https://github.com/containerd/containerd/blob/e9af808/oci/spec.go#L164)
+			oci.WithMaskedPaths([]string{
+				"/proc/acpi",
+				"/proc/asound",
+				"/proc/kcore",
+				"/proc/keys",
+				"/proc/latency_stats",
+				"/proc/timer_list",
+				"/proc/timer_stats",
+				"/proc/sched_debug",
+				"/proc/scsi",
+			}),
 			// Pass proxy environment variables to this container
 			withProxyEnv(),
 			// Add a default set of mounts regardless of the container type


sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -288,6 +288,18 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 			oci.WithHostNamespace(runtimespec.NetworkNamespace),
 			oci.WithHostHostsFile,
 			oci.WithHostResolvconf,
+			// Unmask `/sys/firmware` by passing an alternate list of masked paths
+			oci.WithMaskedPaths([]string{
+				"/proc/acpi",
+				"/proc/asound",
```
yeazelm, line 296: Can you link to where you got this list? I'd like to know how we make sure it remains up-to-date or at least ensure we all think its correct.

jpculp: It's part of [`oci/spec.go`](https://github.com/containerd/containerd/blob/6906b57c721f9114377ceb069662b196876915c0/oci/spec.go#L164). I would like to implement something a little better where we fetch the `MaskedPaths` from the spec generated by `populateDefaultUnixSpec` and then just drop `/sys/firmware` from the list, but that may have to be implemented in a follow-up PR.

yeazelm: That works for me.

bcressey: I'd recommend putting the link in a comment above the code, at a minimum.

jpculp:
Added the reference for masked paths into the code.

webern (approved):

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -288,6 +288,20 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 			oci.WithHostNamespace(runtimespec.NetworkNamespace),
 			oci.WithHostHostsFile,
 			oci.WithHostResolvconf,
+			// Unmask `/sys/firmware` by passing an alternate list of masked paths
+			// List is based on the DefaultUnixSpec's MaskedPaths for Linux
+			// (https://github.com/containerd/containerd/blob/main/oci/spec.go)
```
bcressey: nit: a permalink to the exact `MaskedPaths` list would be preferable, in case that list gets refactored out of the file

bcressey (approved):

rpkelly (approved):

jpculp:
Pinned the comment to the commit the masked paths were based on.

bcressey (approved):

PR #2575, opened 2022-11-12
Fix minor formatting issues

stefansundin:
**Issue number:**

N/A

**Description of changes:**

I have been experimenting with Bottlerocket recently and I've been a bit annoyed that when saving some files, my editor often fixes up some whitespaces unrelated to my change, like whitespace at the end of a line or missing newline at the end of files.

I also noticed that some documentation was out of date, and I had `.DS_Store` files showing up in my working tree. So I decided to take a pass to find and fix as many of these small problems as possible in one go and try to get it merged upstream.

Please let me know if there's any change you don't agree with and I can revert those changes. Thanks!

**Testing done:**

I've rendered some of the documentation pages and verified that they still look good. I found some code formatting were using the wrong tag which should now be fixed.

**Terms of contribution:**


tools/partyplanner
```diff
@@ -58,7 +58,7 @@ BIOS_MIB="4" # one per disk
 OVERHEAD_MIB="$((GPT_MIB * 2 + BIOS_MIB))"
 
 # The 'recommended' size for the EFI partition is 100MB but our EFI images are
-# under 1MB, so this will suffice for now. It would be possible to increase the
+# under 2MB, so this will suffice for now. It would be possible to increase the
```
stefansundin, line 61: I have tried to build an image using a 1 MB EFI partition but that failed. Using a 2 MB partition worked fine.

.gitignore
```diff
@@ -14,3 +14,5 @@
 /Licenses.toml
 /licenses
 *.run
+*.DS_Store
```
stmcginnis: These should really be part of your global gitignore configuration. I don't think we should be adding OS and IDE specific things here.

See `core.excludesfile` in https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

stmcginnis (approved):
Awesome, a lot of great clean up here. This would get us pretty close to being able to add the mdlint job that we use in some of the other repos.

If you do an update, please address the `.gitignore` comment. Otherwise I don't think that's too critical, so while I'd prefer not to have it, I don't think it's a big deal if we include it.

.gitignore
stefansundin: I see where you are coming from on this but in general I don't agree that omitting these files from the gitignore is worth it. While these files are pretty dumb and it would be nice if there was simpler way to never have them show up, the fact is that some day someone will accidentally commit one of these files to this repository. Since everything here is thoroughly code-reviewed I don't think the file will actually end up being merged to the upstream branch. Having them in gitignore will save a little bit of time and avoid the need for a couple of code review comments and some back-and-forths. The trade-off is ~20 bytes of data in the gitignore to ensure some junk files that no one ever wants to see in this repository will ever end up in it.

Anyway, that's simply my pragmatic view of why I usually put these files in my gitignore. It's just my opinion though, so I've already amended the commit to oblige what is probably already established as consensus in this project. :-)

stmcginnis: Thanks @stefansundin - yeah, I agree there is no "one right way" here and the way you had it certainly works. Thanks a lot for your work on this PR and updating this!

sources/api/certdog/README.md
```diff
@@ -8,4 +8,4 @@ Current version: 0.1.0
 
 ## Colophon
 
-This text was generated using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
\ No newline at end of file
+This text was generated using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
```
etungsten: I'm wondering what this change does? It seems like it makes no difference in the rich diff. The reason I ask is because all the READMEs that have this are auto-generated from docstrings from our source code, so I worry that this change won't last when we touch the source in any way.

etungsten (commented):
Thanks for the fixes

BUILDING.md
```diff
@@ -68,28 +68,34 @@ Docker's [post-installation steps for Linux](https://docs.docker.com/install/lin
 
 To build an image, run:
 
-```
+```shell
 cargo make
 ```
 
-This will build an image for the default variant, `aws-k8s-1.21`.
+This will build an image for the default variant (a recent `aws-k8s-*`, see [Makefile.toml](Makefile.toml)).
```
jpmcb: Should we just point to the `BUILDSYS_VARIANT` line here?

https://github.com/bottlerocket-os/bottlerocket/blob/ac833d8191a235bd85c185bb30c22647ea9d7013/Makefile.toml#L21

BUILDING.md
```diff
@@ -227,11 +233,11 @@ tuftool download "${OUTDIR}" --target-name ${VARIANT}-${ARCH}-kmod-kit-${VERSION
 ### Using the kmod kit
 
 To use the kmod kit, extract it, and update your PATH to use its toolchain:
-```bash
+```shell
 tar xf "${VARIANT}-${ARCH}-kmod-kit-${VERSION}.tar.xz"
 
 export CROSS_COMPILE="${ARCH}-bottlerocket-linux-musl-"
-export KERNELDIR="${PWD}/${VARIANT}-${ARCH}-kmod-kit-${VERSION}/kernel-devel
+export KERNELDIR="${PWD}/${VARIANT}-${ARCH}-kmod-kit-${VERSION}/kernel-devel"
```
jpmcb, line 240: Good catch!

TESTING.md
```diff
@@ -269,7 +269,7 @@ WORKING_BRANCH="develop"
 git checkout "${WORKING_BRANCH}"
 ```
 
-Next, build Bottlerocket images and repos and sync TUF repos. 
+Next, build Bottlerocket images and repos and sync TUF repos.
```
jpmcb, line 278: I wonder if there's an opportunity here to have a linter that will fail if we have hanging whitespace. Is that something the MDlinter does @stmcginnis ?

jpmcb (commented):
Awesome!! 👏🏼 Thanks for the hard work on this one - huge value add!

sources/api/certdog/README.md
jpmcb: I'm wondering the same thing: I'm not seeing a change in the GitHub UI and it doesn't appear to be adding a newline at the end?

TESTING.md
stmcginnis: Yep, this is one of the things that would catch.

I wouldn't want it added to this PR since there's already so much, but I think these changes would get us most of the way there to be able to enable markdown linting like in https://github.com/bottlerocket-os/bottlerocket-test-system/pull/514

Then we can make sure these kinds of things don't slip back in.

sources/api/certdog/README.md
stefansundin: Yes, it is the newline at EOF. It was missing in many of these files. It is strange the GitHub is not rendering the symbol that it renders in some views.

It's automatically added when I run `cargo readme > README.md`. I see that there's a `generate-readme` package that is supposed to automatically generate these but I don't know if there's an easy way to invoke this. I suppose it happens at build time. I just built this and there were no changes to the README files.

Here's an alternative way to view the diff: https://github.com/bottlerocket-os/bottlerocket/commit/d5487d348f6bbf1e8c2ea973d0f23441d37d88fc.diff

```diff

diff --git a/sources/api/certdog/README.md b/sources/api/certdog/README.md
index 8769328a9c..e87b8418e0 100644
--- a/sources/api/certdog/README.md
+++ b/sources/api/certdog/README.md
@@ -8,4 +8,4 @@ Current version: 0.1.0
 
 ## Colophon
 
-This text was generated using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
\ No newline at end of file
+This text was generated using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
```

TESTING.md
```diff
@@ -105,7 +105,7 @@ cargo make testsys add secret map  \
 ### Conveniences
 
 All testsys commands can be run using cargo make to eliminate the chance of 2 different versions of
-testsys bing used. 
+testsys bing used.
```
bcressey: nit: while you're making a change 😀 
```suggestion
testsys being used.
```

sources/api/corndog/README.md
```diff
@@ -9,4 +9,4 @@ It sets kernel-related settings, for example:
 
 ## Colophon
 
-This text was generated from `README.tpl` using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
\ No newline at end of file
+This text was generated from `README.tpl` using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
```
bcressey, line 12: Most of these edits end up being not stable across a run where `README.md` is recreated - if I check out your branch and do something like this, the newlines are lost:
```
cd sources
find -name README.md -delete
export VARIANT=aws-dev
cargo build
```

To make them permanent, you'd need to change the `generate_readme` helper to write the trailing newline, since this is what each crate's build script invokes. The offending code is here: https://github.com/bottlerocket-os/bottlerocket/blob/b530f30850d58ad84a8b0ded2e08fa91d554c7a8/sources/generate-readme/src/lib.rs#L87

bcressey (commented):
Nice, thank you for doing this!

sources/api/corndog/README.md
stefansundin: This seems to be a problem in cargo-readme where running `cargo readme > README.md` can produce a different result than when using `cargo_readme::generate_readme()` 😞 

There's a call to `template.trim_end_matches("\n")` here: https://github.com/livioribeiro/cargo-readme/blob/11ce95834c5ef52922174a22cd91e9d1868cf3d4/src/readme/template.rs#L54

I suspect that when `cargo readme` is called a newline is added at the end at some other part of the code.

Anyway, I made a change to `generate_readme` to append a newline. Hopefully that is good enough.

It might make sense to try to fix this in cargo-readme, but it looks like there's not much development there anymore. Let me know if you want me to revert these newlines.

TESTING.md
stefansundin: Done 👍

PUBLISHING-AWS.md
```diff
@@ -136,7 +136,8 @@ This will copy the fully versioned parameter from earlier, something like:
 /your/prefix/here/aws-k8s-1.21/x86_64/1.0.1-dafe3b16/image_id
 ```
 
-...to a simpler parameter name:
+... to a simpler parameter name:
```
webern: The space seems less idiomatic here to me.

```suggestion
...to a simpler parameter name:
```

sources/api/apiclient/README.md
```diff
@@ -153,30 +153,30 @@ To see verbose response data, including the HTTP status code, use `-v` or `--ver
...
 
-```
+```shell
 apiclient raw -u /settings
 ```
 
 This will return all of the current settings in JSON format.
 For example, here's an abbreviated response:
-```
+```json
```
webern: Since this one is pseudo-json we get a syntax highlight error from it.

```suggestion
```text
```

sources/api/apiclient/README.md
```diff
@@ -204,4 +204,4 @@ HTTP method, and any request body data.
 
 ## Colophon
 
-This text was generated from `README.tpl` using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/lib.rs`.
\ No newline at end of file
+This text was generated from `README.tpl` using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/lib.rs`.
```
webern: This would need to be changed in the `README.tpl` file if it's a change we care about. I can't tell from this GitHub diff rendering what was changed.

sources/generate-readme/src/lib.rs
```diff
@@ -84,6 +84,9 @@ where
     )
     .map_err(|e| error::ReadmeGenerateSnafu { error: e }.build())?;
 
+    // Make sure the end of the file has a newline
+    content = content + "\n";
```
webern: Should we double-check before we add a newline? Maybe it's not necessary if we expect cargo readme to always trim it...

```suggestion
    if content.chars().last().unwrap_or_default() != '\n' {
        content = content + "\n";
    }
```

BUILDING.md
webern: > Should we just point to the BUILDSYS_VARIANT line here?

I think that would be brittle, but we could augment the sentence, something like:

```suggestion
This will build an image for the default variant (a recent `aws-k8s-*`, see the `BUILDSYS_VARIANT` variable in [Makefile.toml](Makefile.toml) to find the current default variant).
```

webern (approved):
Than you for this PR.

I was writing notes on how we generate our readmes before I saw your conversation with @bcressey and your change to the generate_readme function. Thank you for figuring that out.

If you haven't recently, can go into sources and `export VARIANT=aws-dev && cargo build` and see that there is no git diff after building?

BUILDING.md
stefansundin: Made this change. 👍

PUBLISHING-AWS.md
stefansundin: 👍

sources/api/apiclient/README.md
stefansundin: 👍

sources/generate-readme/src/lib.rs
stefansundin: Sure 👍

stefansundin:
Thanks for the review @webern.

> If you haven't recently, can go into sources and `export VARIANT=aws-dev && cargo build` and see that there is no git diff after building?

I tried this but I got an error from one package that I'm not running Linux. I assume that when I run the actual build inside of Docker it won't propagate these kinds of things to my git worktree, right?

I can run this on a Linux machine later today.

webern:
> I tried this but I got an error from one package that I'm not running Linux. I assume that when I run the actual build inside of Docker it won't propagate these kinds of things to my git worktree, right?

Yeah, there are linux things in there and I found it very hard to cross compile from a mac when I tried. This seems to work though. First I purturbed a generated readme file and committed that change. Then I compiled with this:

webern:
> I tried this but I got an error from one package that I'm not running Linux. I assume that when I run the actual build inside of Docker it won't propagate these kinds of things to my git worktree, right?

Yeah, there are Linux things in there and I found it very hard to cross compile from a mac when I tried. This seems to work though. First I perturbed a generated readme file and committed that change. Then I compiled with this (from the root of the Bottlerocket git tree):

```shell
docker run \
    --mount "type=bind,src=$(pwd)/sources,dst=/sources" \
    --workdir="/sources" \
    --name build-bottlerocket-sources \
    --rm \
    -e VARIANT=aws-k8s-1.23 \
    --user root \
    public.ecr.aws/bottlerocket/bottlerocket-sdk-x86_64:v0.28.0 \
    cargo build
```

This did show that the README was out-of-sync

```text
$ git status
On branch delete-this-branch
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   sources/models/README.md
```

stefansundin:
Thanks @webern. I just ran that and no files changed. 🎉

zmrow (approved):
:volcano: 

Thanks!

jpmcb (approved):
👏🏼

stmcginnis:
Shoot, there's a merge conflict now. Would you mind rebasing on the latest from `develop`? Then we can try to wrap this up quickly so we don't run into any more conflicts.

stefansundin:
@stmcginnis Done!

jpmcb (approved):
Rebase looks good!

stmcginnis:
Let's get this in! Would be good to have feedback from a few of the folks that had commented earlier, but since this is a holiday week in the US it's probably not worth waiting in the hopes of getting that. Easy enough to follow up with any new PRs if there is anything in these changes that are of concern.

PR #2595, opened 2022-11-19
modeled-types: constrain kubelet log level

### LINKED ISSUE #2463: Add validation for `settings.kubernetes.log-level` value

webern:
**Issue number:**

Closes #2463

**Description of changes:**

Background: we have had challenges validating non-string values such as numeric ranges. See:
- https://github.com/bottlerocket-os/bottlerocket/pull/2527#discussion_r1007653722
- https://github.com/bottlerocket-os/bottlerocket/pull/2527#issuecomment-1295126294
- #2257 
- https://github.com/bottlerocket-os/bottlerocket/pull/2460#discussion_r981182921
- Enabled by #2559

**Testing done:**

- Ran a k8s variant with `log-level=255`. Boot failed with `[    9.093943] early-boot-config[1489]: Error PATCHing '/settings?tx=bottlerocket-launch': Status 400 when PATCHing /settings?tx=bottlerocket-launch: Json deserialize error: Unable to deserialize into KubeletLogLevel: KubeletLogLevel value '255' is invalid, should be 0-8 at line 1 column 62`
- Ran a k8s variant with `log-level=8`. Succeeded. Checked `/etc/systemd/system/kubelet.service.d/exec-start.conf` and saw `-v 8`.
- Ran a k8s variant with `log-level=0`. Succeeded. Checked /etc/systemd/system/kubelet.service.d/exec-start.conf and saw `-v 0`. 

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,7 +475,8 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.kube-api-qps`: The QPS to use while talking with kubernetes apiserver.
 * `settings.kubernetes.log-level`: Adjust the logging verbosity of the `kubelet` process.
-  The default log level is 2, with higher numbers enabling more verbose logging.
+  Valid values are 0-8 with higher numbers enabling more verbose logging.
+  The default log level is 2.
 * `settings.kubernetes.pod-pids-limit`: The maximum number of processes per pod.
 * `settings.kubernetes.provider-id`: This sets the unique ID of the instance that an external provider (i.e. cloudprovider) can use to identify a specific node.
 * `settings.kubernetes.registry-burst`: The maximum size of bursty pulls.
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -148,12 +148,13 @@ use crate::modeled_types::{
     BootConfigKey, BootConfigValue, BootstrapContainerMode, CpuManagerPolicy, CredentialProvider,
     DNSDomain, ECSAgentImagePullBehavior, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue,
     ECSDurationValue, EtcHostsEntries, FriendlyVersion, Identifier, ImageGCHighThresholdPercent,
-    ImageGCLowThresholdPercent, KmodKey, KubernetesAuthenticationMode, KubernetesBootstrapToken,
-    KubernetesCloudProvider, KubernetesClusterDnsIp, KubernetesClusterName,
-    KubernetesDurationValue, KubernetesEvictionHardKey, KubernetesLabelKey, KubernetesLabelValue,
-    KubernetesQuantityValue, KubernetesReservedResourceKey, KubernetesTaintValue,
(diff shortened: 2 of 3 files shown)


stmcginnis:
Looks good, but the question has come up whether we should be enforcing this during boot or not.

The good thing about boot time enforcement is we are ensuring things are correct. This makes it very precise, which I really like.

The downside is if there is a configuration typo the node will fail to boot, and it's not always clear to an end user why that is happening. If we allow something like 255 through, then either kublet will just ignore it or it will fail. I don't know its exact behavior, but I think if it's implemented in a robust enough way, it would just treat something like that as `$MAX_LOG_LEVEL` and just carry on.

But if not, and kubelet fails to start, then at least the system is running and the user can poke around in the logs and get an error message from kubelet giving a little clearer reason why something is misconfigured.

webern:
> If we allow something like 255 through, then either kublet will just ignore it or it will fail.

I tried it, and I think the kubelet clamps the value to 8 (basically I could tell the logging was verbose with the value of 255).

I agree it's questionable whether we want to do this or not. If we think this isn't a good idea, no problem, I can close this PR and the associated issue. I wasn't sure if the issue was created because somebody definitely thought we should constrain this value.

etungsten:
Can you check if a migration is necessary or not? I suspect we need a migration to migrate the value stored in the settings datastore. I would try doing a downgrade first to verify.

stmcginnis:
> Can you check if a migration is necessary or not? I suspect we need a migration to migrate the value stored in the settings datastore. I would try doing a downgrade first to verify.

The value is the same data type, so the concern I think would be what would happen if the user had set the value to something outside of the valid range, then tried to upgrade to a version that enforced these constraints.

webern:
> Can you check if a migration is necessary or not? I suspect we need a migration to migrate the value stored in the settings datastore. I would try doing a downgrade first to verify.

Migrations are not necessary for the changes I am making to the model (including this one). These changes are transparent from the standpoint of userdata and the api. These are implementation details only.

I suspect we will be closing this PR and issue though since values outside of the constrained range are not a problem for the kubelet.

stmcginnis:
@bcressey can we get your thoughts on this? I could see either way, but I'm starting to lean towards allowing a node to boot with configuration warnings versus boot failure as a better user experience.

bcressey:
> @bcressey can we get your thoughts on this? I could see either way, but I'm starting to lean towards allowing a node to boot with configuration warnings versus boot failure as a better user experience.

If `kubelet` tolerates values outside of a specific range, either by emitting a warning or by clamping them to a valid number, then I'm happy to let it see the larger value.

I wrote more detailed thoughts here - https://github.com/bottlerocket-os/bottlerocket/pull/2527#issuecomment-1295126294 - and this sounds like the third case: match the types but don't do additional validation.

webern:
Closing since kubelet accepts values outside of this range.

PR #2604, opened 2022-11-22
Remove use of io/ioutil in go code

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

The `io/ioutil` package was [deprecated in Go 1.16](https://go.dev/doc/go1.16#ioutil). Since we now use go 1.19 it is safe to migrate away from the deprecated modules so there are no issues once they are actually removed.

This updates the few instances we had where we were importing and using `io/ioutil`.

**Testing done:**

Built code.

**Terms of contribution:**


jpmcb (approved):
👍🏼 Thanks Sean!

bcressey (approved):

etungsten (approved):

PR #2607, opened 2022-11-22
docs: Add example of mounting API socket

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This adds a small example to the API docs to show how a pod would need to be defined in order to access the API socket. We mention it in a couple places in the docs, but there are no concrete examples of how this is accomplished.

**Testing done:**

Viewed rendered output of markdown changes.

**Terms of contribution:**

### DIFF
--- a/sources/api/README.md
+++ b/sources/api/README.md
@@ -28,6 +28,26 @@ It's available in Bottlerocket, whether you're accessing it through a control ch
 
 Rust code can use the `apiclient` library to make requests to the Unix-domain socket of the [apiserver](#apiserver).
 
+The API socket can be mounted and accessed from a privileged container.
+An example pod specification to expose API access would be similar to:
+
+```yaml
+containers:
+- name: my-api-access
+  image: my-api-access
+  volumeMounts:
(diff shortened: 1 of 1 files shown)


zmrow (approved):

sources/api/README.md
```diff
@@ -28,6 +28,23 @@ It's available in Bottlerocket, whether you're accessing it through a control ch
 
 Rust code can use the `apiclient` library to make requests to the Unix-domain socket of the [apiserver](#apiserver).
 
+The API socket can be mounted and accessed from a privileged container.
+An example pod specification to expose API access would be similar to:
```
bcressey, line 32: It would be good to link to the security guidance on this topic - https://github.com/bottlerocket-os/bottlerocket/blob/develop/SECURITY_GUIDANCE.md#restrict-access-to-the-host-api-socket

stmcginnis: Good call - I've added a note to the section pointing to the security guidance doc.

bcressey (approved):

PR #2588, opened 2022-11-17
Third-party package updates after 1.11 release

etungsten:
**Issue number:**

N/A

**Description of changes:**
We're updating third party packages earlier in the release cycle to so they're sufficiently exercised by our regular development testing up to the next release.

```
5dea53cc packages: update makedumpfile
9a2358e4 packages: update libxcrypt
b80d70b5 packages: update libnftnl
823ebc9a packages: update liblzma
779256b6 packages: update libglib
2f0ceb9e packages: update libffi
ec2fda3f packages: update libelf
a2fc1915 packages: update ecs-agent
fedfc7ac packages: update docker-cli
53692f6c packages: update docker-engine
7abd7397 packages: update containerd
627053f4 packages: update conntrack-tools
250c90e0 packages: update ca-certificates
```

Regenerated patch that don't cleanly apply for `ecs-agent` 

I left out the following package updates in this PR:

- binutils 2.38 -> 2.39. `bottlerocket-sdk` has to move to 2.39 first.
- ncurses 6.2 -> 6.3. See https://github.com/bottlerocket-os/bottlerocket/issues/1813
- cni/cni-plugins 0.8.1 -> 1.1.2. See https://github.com/bottlerocket-os/bottlerocket/issues/1745
- systemd 250.4 -> 251/252. See https://github.com/bottlerocket-os/bottlerocket/issues/2410
- wicked 0.6.68 -> 0.6.70. See comment: https://github.com/bottlerocket-os/bottlerocket/pull/2588#issuecomment-1319342336 Created an issue here: https://github.com/bottlerocket-os/bottlerocket/issues/2591

**Testing done:**
Built aws-ecs-1 and aws-k8s-1.23 without problems.

Testing:

- [x] aws-k8s-1.24 can run pods fine
- [x] aws-k8s-1.24-nvidia can run sample K8s GPU workloads
- [x] aws-ecs-1 can run an simple ECS nginx task
- [x] aws-ecs-1-nvidia can run simple GPU workloads
- [x] vmware-k8s, created cluster with eks-a with OVA built from changes and can run workloads fine
- [x] metal-k8s, created cluster with eks-a with image built from changes and can run workloads fine

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #2608, opened 2022-11-22
containerd: drop unused `fs.RootPath` patch from source control

jpmcb:
**Issue number:**

Closes: n/a

**Description of changes:**

```
This patch was removed from the containerd rpm spec in 7115cde4b81505d87aa26b3ea050711f99e5a31a
and was merged in upstream containerd in https://github.com/containerd/containerd/commit/9f316fa4b5d4089d8ccc75f02e9e348c8b94544c.

This patch removes the unused patch from our source control and doesn't
have any functional change.

```

**Testing done:**

Able to `cargo make` with no problem.

**Terms of contribution:**


stmcginnis (approved):
We currently use containerd 1.6.8. Confirmed the change is present in that release:

https://github.com/containerd/containerd/blob/v1.6.8/pkg/cri/opts/container.go#L121

Also verified we are not pulling in that patch in the containerd.spec file, so it looks like it was an oversight that the patch was removed from the package build but the file was left.

etungsten (approved):
Nice catch. I'll rebase https://github.com/bottlerocket-os/bottlerocket/pull/2588 once this is merged. Or I can cherrypick this commit as part of it. Not sure which is the cleanest, so I'll let you decide!

jpmcb:
@etungsten - I usually like rebase / force pushing, but it's up to you! I'll go ahead and merge this!!

PR #2610, opened 2022-11-22
Go: implements golangci-lint makefile target

jpmcb:
**Issue number:**

Closes - n/a

Related to reviewing: https://github.com/bottlerocket-os/bottlerocket/pull/2604

**Description of changes:**

```
Go: implements golangci-lint makefile target & GitHub actions workflow

Now, when `cargo make check-lints` is run, golangci-lint will run on
all go modules via docker with a `.golangci.yaml` config file.

Adds workflow for Go linting when Go files change.

Avoids running Go lints in our GitHub actions build.

Fixes the following Go linting errors:
- Unnecessary `fmt.Sprintf`
- Fix `Https` var name with `HTTPS`
- Fix un-`goimports`-ed test file

```

**Testing done:**

Able to `cargo make` and run golangci-lint with `cargo make check-golangci-lint` or `cargo make check-lints`.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,7 +133,9 @@ jobs:
           EOF
       - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} unit-tests
       - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} check-fmt
-      - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} check-lints
+      # Avoid running Go lint check via `cargo make check-lints` since there's a separate golangci-lint workflow
+      - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} check-clippy
+      - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} check-shell
       - run: |
           cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} \
             -e BUILDSYS_ARCH=${{ matrix.arch }} \
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -349,7 +349,7 @@ dependencies = [
    "unit-tests",
    "check-fmt",
    "check-lints",
-   ]
+]
 
 [tasks.check-fmt]
 script = [
@@ -402,6 +402,7 @@ fi
 dependencies = [
    "check-clippy",
    "check-shell",
+   "check-golangci-lint",
 ]
 
 [tasks.check-clippy]
@@ -461,6 +462,25 @@ if [ "${rc}" -ne 0 ]; then
   exit $rc
 fi
 '''
+]
+
+[tasks.check-golangci-lint]
+script = [
+'''
+top_path=$(pwd)
+config_path="${top_path}/.golangci.yaml"
+
+for m in ${GO_MODULES}; do
+    cd "sources/${m}"
+    mod_name=$(pwd)
+    docker run --rm \
+        -v "${mod_name}":/"${mod_name}" \
+        -v "${config_path}":/"${config_path}" \
+        -w /"${mod_name}" \
+        golangci/golangci-lint \
+        golangci-lint run --config "${config_path}"
+    cd "${top_path}"
+done'''
 ]
 
 [tasks.build-tools]
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,0 +1,29 @@
+name: golangci-lint
+on:
+  pull_request:
+    branches: [develop]
+    # Only run this workflow if Go files have been modified
+    paths:
+      - '**.go'
+      - '*/go.mod'
+      - '*/go.sum'
+
+jobs:
+  golangci-lint:
+    name: lint
+    runs-on: [self-hosted, linux, x64]
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+      - uses: actions/checkout@v3
+      - name: lint-host-ctr
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: latest
+          working-directory: sources/host-ctr
+      - name: lint-ecs-gpu-init
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: latest
+          working-directory: sources/ecs-gpu-init
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,17 @@
+linters:
+  enable:
+    - errcheck
+    - goimports
+    - ineffassign
+    - misspell
+    - revive
+    - staticcheck
+    - unconvert
+    - unused
+    - vet
+
+run:
+  timeout: 3m
+  skip-dirs:
+    - vendor
+    - .gomodcache
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -475,7 +475,7 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 
 	// Return error if container exists with non-zero status
 	if code != 0 {
-		return errors.New(fmt.Sprintf("Container %s exited with non-zero status", containerID))
+		return fmt.Errorf("Container %s exited with non-zero status", containerID)
 	}
 
 	return nil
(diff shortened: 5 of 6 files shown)


stmcginnis:
Awesome, I was thinking the same thing!

Do we want to add a `.golangci.yaml` so we can make our settings explicit.

Makefile.toml
```diff
@@ -463,6 +463,18 @@ fi
 '''
 ]
 
+[tasks.check-golangci-lint]
```
etungsten, line 467: If we add this as a dependency of `check-lints` then we wouldn't need the separate workflow since `cargo make check-lints` is covered in the `build` workflow.

stmcginnis: But then we also can't only have it run when *.go files have changed.

jpmcb: I opted out of adding this as a dependency for `check-lints` since it will break everyone's build who don't already have Go and/or `golangci-lint` installed.

But, if we want to make that change and enforce these checks on every build, I'm all for it!

jpmcb:
FYI - successful run of the new workflow based off this PR: https://github.com/bottlerocket-os/bottlerocket/actions/runs/3527323950/jobs/5916257315

Makefile.toml
etungsten: > I opted out of adding this as a dependency for check-lints since it will break everyone's build who don't already have Go and/or golangci-lint installed.

We can try doing it in a docker container like we do for `cargo make check-shell` with the bottlerocket-sdk which is a part of `check-lint`.
https://github.com/bottlerocket-os/bottlerocket/blob/bb92641a238c2a7897de968b9c0d3adf3070e180/Makefile.toml#L433-L443

> But then we also can't only have it run when *.go files have changed.

Does the lint checker take a long time? I think if it's very fast then I don't think it's a big issue. We do shellcheck for every PR right now as well.

jpmcb: Sure - If the prior-art is to do the lints on each run/build, I think that's fair and yes, golangci-lint is very fast.

marking this as draft to go back and iterate!

stmcginnis: > Does the lint checker take a long time? I think if it's very fast then I don't think it's a big issue. We do shellcheck for every PR right now as well.

It doesn't, but I would like to see that shellcheck and other checks broken out in our Actions so they are not run on every run. They don't individually take long, but right now we are running all of them for every variant for every PR. granted it's not a huge savings, but added up over time it would be a lot more efficient if we only ran checks when needed and only for things that have changed since the last time they were checked.

webern (commented):
I think I'm saying the same thing as @etungsten, but I'd like to be able to run things like this locally with `cargo make lint`.

Edit: we should do this with the SDK and include the rust linting in the target as well.

stmcginnis:
Maybe add another target so we can easily run all checks locally, but have separate targets that allow us to set up GitHub Actions to only run certain tests when needed?

jpmcb:
Force pushed:

- Added a basic `.golangci-lint.yaml` config file
- Utilize the `golangci-lint` docker image to lint the target Go modules
- Removed the `.github/workflows/golangci-lint.yaml` workflow. Instead, added the go lint target as part of the `check-lints` target per suggestions. Maybe a future step out of band of this PR is breaking our lints / checks into seperate github workflows to better parallelize these checks

stmcginnis:
>  Maybe a future step out of band of this PR is breaking our lints / checks into seperate github workflows to better parallelize these checks

This is really the only one we can break out. I attempted to break out the others before, but since we do conditional compilation for the different variants that ended up being a trickier one than I originally thought.

Since there is no difference with variants with the go code, I still strongly believe this should be its own target with its own workflow.

Makefile.toml
webern: > It doesn't, but I would like to see that shellcheck and other checks broken out in our Actions so they are not run on every run.

I wouldn't want to see this breakout take place without at least quantifying the amount of time being spent of the checks. If it's, e.g. 1% of the actions run, then I would prefer to run them for every variant/arch/build

Makefile.toml
```diff
@@ -402,6 +402,7 @@ fi
 dependencies = [
    "check-clippy",
    "check-shell",
+   "check-golangci-lint",
```
webern, line 405: :+1:

webern (approved):

Makefile.toml
```diff
@@ -463,6 +464,26 @@ fi
 '''
 ]
 
+[tasks.check-golangci-lint]
+script = [
+'''
+top_path=$(pwd)
+config_path=$top_path/.golangci.yaml
```
zmrow: Nit: to be more consistent with the variable expansion in the rest of the file:
```
config_path="${top_path}/.golangci.yaml"
```
There's a few uses below as well. :)

jpmcb:
> Since there is no difference with variants with the go code, I still strongly believe this should be its own target with its own workflow.

@stmcginnis - that is super fair and I believe you're right, it doesn't make sense to run a linter on each run of a build in github actions if there are no changes and it's not variant dependent.

However, I don't see a way forward to have `check-golangci-lint` as a dependency in `check-lints` which is then executed on each variant in our build workflow: https://github.com/bottlerocket-os/bottlerocket/blob/f825677d128e0931b57efe08dfc0424968e055fc/.github/workflows/build.yml#L136

We could break that target up in the github workflow to call all linters _except_ for the Go one but it seems both @etungsten and @webern are not in favor of that and would prefer an uber lint target?

stmcginnis:
> We could break that target up in the github workflow to call all linters except for the Go one but it seems both @etungsten and @webern are not in favor of that and would prefer an uber lint target?

Yeah, I think it would make more sense to have a target someone could run locally that would run all linters, but then the two separate targets that the workflows would use to run golangci-lint by itself and the rest as part of the build.

But if the majority vote is for the redundancy, so be it. I will not give my approval on the PR, but not going to try to block it and go against the majority opinion.

Makefile.toml
jpmcb: True - these paths should be wrapped to avoid unintentional escape

---

One note however: we have vastly inconsistent escape patterns in our Makefile shell code. So I just went with what was easiest.

This uses the `{}` to escape the path:
https://github.com/bottlerocket-os/bottlerocket/blob/f825677d128e0931b57efe08dfc0424968e055fc/Makefile.toml#L414

This only wraps the statement in a string:
 https://github.com/bottlerocket-os/bottlerocket/blob/f825677d128e0931b57efe08dfc0424968e055fc/Makefile.toml#L336

This wraps in single quotes `''` (although there's no risk of unescaped path causing problems here):
https://github.com/bottlerocket-os/bottlerocket/blob/f825677d128e0931b57efe08dfc0424968e055fc/Makefile.toml#L527

We should consider a way to shell check the Makefile targets. Although, that may be weird / challenging since it's only a toml file

stmcginnis: +1 for the `config_path="${top_path}/.golangci.yaml"` syntax. If we were able to figure out a way to run shellcheck against this code, I think that is what the tool enforced by default too.

etungsten:
> Yeah, I think it would make more sense to have a target someone could run locally that would run all linters, but then the two separate targets that the workflows would use to run golangci-lint by itself and the rest as part of the build.

I'm OK with that. We can preserve `check-lints` to cover all linting invocation. For github actions we can granualize and call other Makefile targets for our workflows that can specific to the variant being built.

jpmcb:
> For github actions we can granualize and call other Makefile targets for our workflows that can specific to the variant being built.

Good call - I've made that change and re-introduced the `golangci-lint` GitHub action workflow to target Go files

etungsten (approved):

.github/workflows/build.yml
```diff
@@ -133,7 +133,10 @@ jobs:
           EOF
       - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} unit-tests
       - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} check-fmt
-      - run: cargo make -e BUILDSYS_VARIANT=${{ matrix.variant }} check-lints
+      # Only run clippy and shell linters (avoids also running Go lint checkers)
+      - run: |
```
jpculp: Can this be two separate runs instead?

jpmcb:
Force pushed to address GitHub actions formatting and each lint being it's own step 📌

.github/workflows/build.yml
stmcginnis: `check-shell` is another one that's not variant-dependent, so it would be nice to break that out into a different workflow that only runs once on a PR. But that's good for a follow up.

stmcginnis (approved):

etungsten (approved):

PR #2612, opened 2022-11-23
Changelog: small fix to Kubernetes EKS-D versions

mjsterckx:
**Description of changes:**

To avoid confusion for people who may not be familiar with EKS-D, it would be better to state the actual underlying Kubernetes versions in the changelog.

**Terms of contribution:**


etungsten (approved):

stmcginnis (approved):
Looks good to me.

webern (approved):
Thanks for fixing this. It seemed wrong to me even when I wrote it, but I saw this precedent  and asked around about it. I agree it makes more sense to use the kube version, not the distro version.

https://github.com/bottlerocket-os/bottlerocket/blob/16e65f609e6ddb6d1fe317d12599e6c15d8b04d0/CHANGELOG.md?plain=1#L98

PR #2596, opened 2022-11-19
Add bonding and vlan tagging support to Bottlerocket

yeazelm:
**Issue number:**
Closes bottlerocket-os/bottlerocket#2369

**Description of changes:**
This will add bonding and vlan tagging support to bottlerocket. There is a lot of code here mostly because we needed to bump the configuration version to 3 and refactor a bunch of logic since there are now different types (kinds) of devices which have different validation logic, testing concerns, etc. This should allow us to add more virtual network devices in the future without so much refactoring. It also moves some of the device-specific code into a new `devices` module for easier understanding of what is a `net_config` concern and what is a "network device" concern. 

**Testing done:**
I've validated this on hardware and qemu. (More testing will come as the PR progresses) The tests all pass as well right now. I also ran clippy to cover those bases as well.

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #2622, opened 2022-12-01
Support configuration using MAC instead of device name

### LINKED ISSUE #2293: `net.toml`: Support configuration using MAC instead of device name

zmrow:
**Issue number:**
Closes #2293 

**Description of changes:**
This PR adds the ability for a user to use MAC address (something like `f8:75:a4:d5:32:64`) instead of the interface name (`eno`) when configuring interfaces.  The commits messages are detailed and a summary provided below.

The PR adds a new type `InterfaceId` which is an enum holding either a name  or MAC address.  The new `MacAddress` type validates the string given is a valid MAC address.  The `InterfaceId` type is used as the key in the main map of ID > interface config.  Keeping the `InterfaceName` type is significant, as it will still be used as the key for bonds/vlans once #2596 is merged, since configuring those things using MAC isn't possible/desirable.

In Bottlerocket, the concept of the "primary" interface is important.  We use this interface's IP when registering with kubelet, to determine DNS settings when using DHCP, and also set some default sysctls on this interface.  However, at the time we are generating network config, `udev` may still be asynchronously renaming interfaces so there isn't a reliable or safe way to map a MAC to an interface name.  Later in the boot process (after interfaces have been setup by wicked and have leases), when interface name is necessary we can map a MAC address to a name using sysfs (`/sys/class/net/*`).  When using a MAC as the identifier for the primary interface, we write a new file `primary_mac_address` to disk instead of `primary_interface`.  We use these files later to determine the primary interface name.  Under the hood if a MAC is used, it is correlated with sysfs to determine the interface name.

One significant change this causes is that we are now setting default sysctls in the wicked "install" helper, rather than using a separate service to do this.  Previously the service ran during `network-pre`, at which point we cant' correlate a MAC to a name.  Setting the sysctls in the helper provides a useful synchronization point where we know that our interfaces have a lease and a more stable name.  This helper is called when setting up new interfaces (static and DHCP) and when DHCP leases are renewed.

A simple example of using MAC, also described in the documentation:
```
["0e:b3:69:44:b6:33"]
dhcp4 = true
```

**Testing done:**
For each of the tests below, I made sure `/etc/resolv.conf` was properly written, current_ip existed, and either `primary_interface` or `primary_mac_address` had been written.  Also verified the sysctls were set properly for the primary interface.
* Boot an `aws-k8s-1.22` node without any customization - it joins the cluster and runs pods just fine.
* Boot an `aws-k8s-1.22` node into an IPv6-enabled EKS cluster and ensure the node receives an ipv6 address (validating the IPv6 sysctl `accept_ra` is still working)
* Boot an `metal-dev` variant using MAC address to set up the primary interface for DHCP.  Also tested using MAC address with a statically configured setup.
```
version = 3
["3c:ec:ef:79:ae:ac"]
dhcp4 = true
```

**Terms of contribution:**


PROVISIONING-METAL.md
```diff
@@ -77,8 +77,8 @@ When these services fail, your machine will not connect to any cluster and will
 #### `net.toml` structure
 
 The configuration file must be valid TOML and have the filename `net.toml`.
-The first and required top level key in the file is `version`; the latest is version `2`.
-The rest of the file is a map of interface name to supported settings.
+The first and required top level key in the file is `version`; the latest is version `3`.
```
zmrow, line 79: This will all drastically change when I rebase on the bonding/vlan changes.

Just looking to get eyes on wording, etc.

sources/api/netdog/src/interface_id.rs
```diff
@@ -0,0 +1,377 @@
+//! The interface_id module contains the definition of a valid network interface name and MAC
```
zmrow, line 1: Not sure why git decided this was a new file, rather than a giant update and rename from `interface_name`. :shrug:

Basically everything is new down to `InterfaceName`, and then the only changes beyond that are the additional unit tests.

sources/api/netdog/src/net_config/v3.rs
```diff
@@ -0,0 +1,129 @@
+use super::v2::NetInterfaceV2;
```
zmrow: This will all drastically change when I rebase on the bonding/vlan changes.  I only added an additional version so I could add unit tests and actually use the new types.

sources/api/netdog/src/net_config/v3.rs
```diff
@@ -0,0 +1,129 @@
...
+
+            wicked_interfaces.push(interface);
+        }
+
+        wicked_interfaces
+    }
+}
+
+impl Validate for NetConfigV3 {
+    fn validate(&self) -> Result<()> {
```
zmrow, line 63: One thing that crossed my mind about validating a config is that currently, a user could configure an interface using a MAC, but also mistakenly configure the same interface using the name.  We wouldn't be able to fully validate that is the case unless we were 100% sure udev was finished renaming things, which we can't really do at this point in the boot process (or ever really).

sources/api/netdog/src/wicked/mod.rs
```diff
@@ -37,6 +37,46 @@ pub(crate) struct WickedInterface {
...
+            InterfaceId::MacAddress(mac) => Self {
+                namespace: Some("ethernet".to_string()),
+                name_body: None,
+                permanent_address: Some(mac),
+            },
+        }
+    }
+}
+
+impl Display for WickedName {
```
zmrow, line 124: This `impl Display` is only necessary for config file naming.  Wicked is specific about it's config file names, so if using MAC, we need to remove the colons from the file name.

zmrow:
^ Rebased on develop and fixup changes related to the bonding/vlan PR

PROVISIONING-METAL.md
```diff
@@ -144,10 +145,14 @@ Vlan tagging is configured as a new virtual network device stacked on another de
...
   * `kind = "vlan"`: This setting is required to specify a vlan device.
-  * `device` (string for device): Defines the device the vlan should be configured on. If VLAN tagging is required, this device should recieve all IP address configuration instead of the underlying device.
+  * `device` (string for device name, not MAC address): Defines the device the vlan should be configured on. If VLAN tagging is required, this device should recieve all IP address configuration instead of the underlying device.
   * `id` (integer): Number between 0 and 4096 specifying the vlan tag on the device
 
 Example `net.toml` version `3` with comments:
+=======
+
+Example `net.toml` with comments:
+>>>>>>> 3b2de953 ([PROVISIONING-METAL] Add documentation for using MAC addresses)
```
webern: oops

PROVISIONING-METAL.md
```diff
@@ -194,6 +199,18 @@ to = "10.10.10.0/24"
 from = "192.168.14.5"
 via = "192.168.14.25"
 
+# Interfaces may be configured using their MAC address rather than the interface name.
+# The MAC address must be quoted.
+["0e:b3:69:44:b6:33"]
+dhcp4 = true
+
+["3e:03:69:49:e6:31".static4]
```
webern, line 202: So you can't refer to this interface in a bonding or vlan configuration because it has no name?

yeazelm: As far as we could tell, wicked simply won't support it. But we might have missed something, a deep dive into wicked's code could confirm if this is true though.

PROVISIONING-METAL.md
yeazelm: We probably need to remove line 71 since we now support much more than DHCP4 and DHCP6.

sources/api/netdog/src/net_config/test_macros/bonding.rs
```diff
@@ -75,6 +75,20 @@ macro_rules! bonding_tests {
                 let rendered = render_config_template(bad);
                 assert!(deserialize_config(&rendered).is_err())
             }
+
+            #[test]
+            fn mac_as_identifier() {
+                let bad = net_config().join("mac_as_identifier.toml");
```
yeazelm, line 81: It occurs to me that we don't do a good job separating bad from good configs in that directory by name. One must look at the test name to know if it should or shouldn't be valid. If someone just looked through and saw these files, they might mistake them for valid. There really isn't any explanation why they are bad.

sources/api/netdog/src/net_config/v3.rs
yeazelm: This is true, but quite a stretch to have someone get to that point. Our primary assumption is a user with either know interface name or mac address, not both. In this case it would require them mixing and matching, it might be worth calling out this is one of our assumptions with mac address specifications? Unfortunately wicked doesn't help us by only supporting interface names for vlans and bonds.

sources/api/netdog/src/net_config/v3.rs
```diff
@@ -70,21 +79,30 @@ impl Validate for NetConfigV3 {
         }
 
         for (name, device) in &self.net_devices {
+            // Bonds / vlans cannot be configured via MAC address
```
yeazelm: I'd like to specifically call out they cannot be configured via MAC address due to wicked not supporting this use. There may be a future where a different network backed could support this.
```suggestion
            // Bonds / vlans cannot be configured in wicked via MAC address
```

sources/api/netdog/src/net_config/v3.rs
```diff
@@ -70,21 +79,30 @@ impl Validate for NetConfigV3 {
...
             // Bonds create the interfaces automatically, specifying those interfaces would cause a
             // collision so this emits an error for any that are found
             if let NetworkDeviceV1::BondDevice(config) = device {
                 for interface in &config.interfaces {
-                    // This checks if the insert already found one, which would be a failure
-                    if !interface_names.insert(interface) {
-                        return error::InvalidNetConfigSnafu {
+                    ensure!(
+                        !interface_names.contains(interface),
+                        error::InvalidNetConfigSnafu {
```
yeazelm: I don't think this works, now you aren't adding the interface to the HashSet as you find it. So if someone added the same interface to two different bonds, this logic would miss it.

sources/api/netdog/src/wicked/mod.rs
```diff
@@ -86,6 +86,46 @@ pub(crate) struct WickedInterface {
...
+                permanent_address: None,
+            },
+            InterfaceId::MacAddress(mac) => Self {
+                namespace: Some("ethernet".to_string()),
+                name_body: None,
+                permanent_address: Some(mac),
+            },
+        }
+    }
+}
```
yeazelm: Nice! Its probably worth a comment for those not certain what the resulting xml must look like for wicked. I would have a hard time understanding what this is doing unless I had looked very carefully at wicked xml examples.

sources/api/netdog/test_data/wicked/net_config.toml
```diff
@@ -189,4 +189,14 @@ miimon-frequency-ms = 100
 miimon-updelay-ms = 1000
 miimon-downdelay-ms = 1000
 
+["f8:74:a4:d5:32:64"]
+dhcp4 = true
```
yeazelm, line 193: We support `-` as well correct? If so should we have tests for it?

yeazelm (commented):
Great work!

sources/api/netdog/src/net_config/v3.rs
zmrow: Ah good catch - I read right over that!

markusboehme:
I have yet to look at the code, just wondering from the change description:

> One significant change this causes is that we are now setting default sysctls in the wicked DHCP helper, rather than using a separate service to do this. Previously the service ran during network-pre, at which point we cant' correlate a MAC to a name. Setting the sysctls in the DHCP helper provides a useful synchronization point where we know that our interfaces have a lease and a more stable name.

Does this mean sysctls won't be applied for a statically configured interface that is identified by its MAC address?

zmrow:
@markusboehme 
> Does this mean sysctls won't be applied for a statically configured interface that is identified by its MAC address?

The helper mentioned is actually an "install" helper called by `wicked` anytime it brings up an interface, static or not.  So to answer your question, the sysctls will be set for both dhcp and statically configured interfaces.  (I'll update that wording in the change description as well.)

PROVISIONING-METAL.md
zmrow: Confirmed: wicked doesn't support this.

sources/api/netdog/src/net_config/test_macros/bonding.rs
zmrow: I suppose that's true.  Probably something we could take care of separately if we're concerned about it.

zmrow:
^ Addresses all the comments and ensures MACs can be handled upper/lowercase, and dash/colon separated.

PROVISIONING-METAL.md
```diff
@@ -107,7 +106,8 @@ Please keep in mind that when using static addresses, DNS information must be su
   * `via` (IP address): Gateway IP address.  If no gateway is provided, a scope of `link` is assumed.
   * `route-metric` (integer): Relative route priority.
 
-Version `3` adds in support for bonding and vlan tagging.
+Version `3` adds the ability to use a MAC address (colon or dash separated) as the identifier for an interface; this support is limited to interface configuration *only* and may not be used in conjunction with bonds or vlans.
+It also adds support for bonding and vlan tagging.
```
yeazelm: nit: You refer to bonds and vlans for the first time in a negative light. Ideally we would introduce bonds and vlans and MAC address features, then tell people about them about any specific problems they might have with these features. I don't know if its better to flip it or not, but possibly we would do: 
```suggestion
Version `3` adds support for bonding, vlan tagging, and the ability to use a MAC address (colon or dash separated) as the identifier for an interface.
MAC address identification is limited to interface configuration *only* and may not be used in conjunction with bonds or vlans.
```
I'm fine being overruled too.

yeazelm (approved):

zmrow:
^ Uses @yeazelm 's docs suggestion - thanks!

sources/api/netdog/src/cli/generate_net_config.rs
```diff
@@ -40,7 +43,10 @@ pub(crate) fn run() -> Result<()> {
     let primary_interface = net_config
         .primary_interface()
         .context(error::GetPrimaryInterfaceSnafu)?;
-    write_primary_interface(primary_interface)?;
+    // Remove existing primary interface files since the current primary may have changed or may
+    // now be using a MAC address (via an override net.toml/reboot)
+    rm_old_primary_interface()?;
```
bcressey: nit: I'd prefer `remove` over `rm` since it's not a shell script 😀 
```suggestion
    remove_old_primary_interface()?;
```

sources/api/netdog/src/cli/install.rs
```diff
@@ -54,14 +51,57 @@ pub(crate) fn run(args: InstallArgs) -> Result<()> {
...
+                let systemd_sysctl_result = Command::new(SYSTEMD_SYSCTL)
+                    .arg(PRIMARY_SYSCTL_CONF)
+                    .output()
+                    .context(error::SystemdSysctlExecutionSnafu)?;
+                ensure!(
+                    systemd_sysctl_result.status.success(),
+                    error::FailedSystemdSysctlSnafu {
+                        stderr: String::from_utf8_lossy(&systemd_sysctl_result.stderr)
+                    }
+                );
```
bcressey: If for whatever reason we are able to write the sysctl file, but the `systemd-sysctl` command fails, we'll never retry applying the settings on a second DHCP lease event.

It might be better to separate these steps:
* write the sysctl.conf if it doesn't exist
* run the systemd-sysctl command if a specific marker file doesn't exist, then write the marker file

sources/api/netdog/src/cli/mod.rs
```diff
@@ -47,9 +48,50 @@ where
...
+
+    for lease in leases {
+        let lease_info =
+            LeaseInfo::from_lease(lease.path()).context(error::LeaseParseFailedSnafu)?;
+
+        if let Some(lease_mac) = lease_info.mac_address {
+            if primary_mac == clean(lease_mac) {
+                return Ok(clean(lease_info.interface_name));
+            }
+        }
```
bcressey: I have some misgivings about this:
* it seems like we're reading lease files while being actively called as a helper by wicked, and therefore we might potentially read files in an inconsistent state 
* overall there's a fair amount of parsing required and we have to do it on every lease event, which can happen frequently for IPv6
* we end up reading and parsing the same lease file twice, since it's read again by `fetch_lease`

Can we iterate over `/sys/class/net/*/address` instead and return the name of the directory where the address file matches?

zmrow:
^ Addresses @bcressey 's feedback!

zmrow:
^ Removes the commit parsing extra fields from the lease.  We're not using those fields now that we're using sysfs to determine interface name

yeazelm (approved):

sources/api/netdog/src/main.rs
```diff
@@ -49,11 +49,14 @@ static KERNEL_HOSTNAME: &str = "/proc/sys/kernel/hostname";
 static CURRENT_IP: &str = "/var/lib/netdog/current_ip";
 static KERNEL_CMDLINE: &str = "/proc/cmdline";
 static PRIMARY_INTERFACE: &str = "/var/lib/netdog/primary_interface";
+static PRIMARY_MAC_ADDRESS: &str = "/var/lib/netdog/primary_mac_address";
 static DEFAULT_NET_CONFIG_FILE: &str = "/var/lib/bottlerocket/net.toml";
 static OVERRIDE_NET_CONFIG_FILE: &str = "/var/lib/netdog/net.toml";
 static PRIMARY_SYSCTL_CONF: &str = "/etc/sysctl.d/90-primary_interface.conf";
+static SYSCTL_MARKER_FILE: &str = "/var/lib/netdog/primary_sysctls_set";
```
bcressey: This is a persistent path, which means the sysctls will be skipped after a reboot since the marker file will already exist.

bcressey:
Can we drop the patch to add the MAC address to the wicked lease, if we're no longer using it?

sources/api/netdog/src/main.rs
markusboehme: Related: What undesirable outcome are the checks before rewriting and resetting the sysctls preventing? The previous code didn't do either explicitly (and I may be missing how that was achieved implicitly).

sources/api/netdog/src/main.rs
```diff
@@ -84,7 +84,6 @@ fn run() -> cli::Result<()> {
         SubCommand::GenerateHostname(_) => cli::generate_hostname::run()?,
         SubCommand::GenerateNetConfig(_) => cli::generate_net_config::run()?,
         SubCommand::SetHostname(args) => cli::set_hostname::run(args)?,
-        SubCommand::PreparePrimaryInterface(_) => cli::prepare_primary_interface::run()?,
```
markusboehme, line 87: Adding here since GitHub doesn't let me too far away from the diff: The comment at the top and netdog's readme still mention the `prepare-primary-interface` command.

markusboehme (commented):
Just some nit picks from my side. The generation of the wicked stanzas taught me some more Serde (`$unflatten`). :+1: 

I'm ready to approve once the reboot concern identified by Ben is addressed.

markusboehme:
> Can we drop the patch to add the MAC address to the wicked lease, if we're no longer using it?

Related, the commit message of ` [netdog] Handle using MAC address as primary interface identifier ` also still mentions using the lease file.

sources/api/netdog/src/main.rs
zmrow: Good eyes - I'll fix that up!

zmrow:
> > Can we drop the patch to add the MAC address to the wicked lease, if we're no longer using it?
> 
> Related, the commit message of `[netdog] Handle using MAC address as primary interface identifier` also still mentions using the lease file.

Thanks!  I'll fix that up

sources/api/netdog/src/main.rs
zmrow: @markusboehme  Previously the sysctl's were set with a system service that only ran once every boot.  The sysctl code is now in the "install" helper that wicked runs when setting up an interface as well as every time a DHCP lease is renewed.  Since this could happen often (especially for dhcp6), it doesn't really make sense to be setting and resetting the sysctls that often.  It probably doesn't cause harm to do it, but why waste the cycles? :shrug:

zmrow:
> Can we drop the patch to add the MAC address to the wicked lease, if we're no longer using it?

@bcressey : https://github.com/bottlerocket-os/bottlerocket/pull/2726

zmrow:
^ Addresses @bcressey and @markusboehme 's comments.  Thanks!

sources/api/netdog/src/main.rs
zmrow: I've fixed this by writing the marker file to `/run/netdog`

bcressey (approved):
🥅

markusboehme (approved):
:rocket:

PR #2637, opened 2022-12-06
host-ctr: build masked paths list dynamically

jpculp:
**Description of changes:**

Replaces the hardcoded masked paths list (introduced in https://github.com/bottlerocket-os/bottlerocket/pull/2573) with a generated one.

**Testing done:**

Built and launched an aarch64 k8s variant on a `c6g.large` and ran `lscpu` from the control container.

```
Architecture:           aarch64
  CPU op-mode(s):       32-bit, 64-bit
  Byte Order:           Little Endian
CPU(s):                 2
  On-line CPU(s) list:  0,1
Vendor ID:              ARM
  Model name:           Neoverse-N1
    Model:              1
    Thread(s) per core: 1
    Core(s) per socket: 2
    Socket(s):          1
    Stepping:           r3p1
    BogoMIPS:           243.75
    Flags:              fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs
```

`/.bottlerocket/rootfs/run/host-containerd/runc/default/control/state.json`:
```
    "mask_paths": [
      "/proc/acpi",
      "/proc/asound",
      "/proc/kcore",
      "/proc/keys",
      "/proc/latency_stats",
      "/proc/timer_list",
      "/proc/timer_stats",
      "/proc/sched_debug",
      "/proc/scsi"
    ],
```

`/.bottlerocket/rootfs/run/host-containerd/runc/default/admin/state.json`:
```
    "mask_paths": null,
```

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -228,6 +228,16 @@ func (ct containerType) Prefix() string {
 	return ""
 }
 
+// SliceContains returns true if a slice contains a string
+func SliceContains(s []string, v string) bool {
+	for _, n := range s {
+		if n == v {
+			return true
+		}
+	}
+	return false
+}
+
 func runCtr(containerdSocket string, namespace string, containerID string, source string, superpowered bool, registryConfigPath string, cType containerType, useCachedImage bool) error {
 	// Check if the containerType provided is valid
 	if !cType.IsValid() {
@@ -301,20 +311,9 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 			oci.WithHostNamespace(runtimespec.NetworkNamespace),
 			oci.WithHostHostsFile,
 			oci.WithHostResolvconf,
-			// Unmask `/sys/firmware` by passing an alternate list of masked paths
-			// List is based on the DefaultUnixSpec's MaskedPaths for Linux
-			// (https://github.com/containerd/containerd/blob/e9af808/oci/spec.go#L164)
-			oci.WithMaskedPaths([]string{
-				"/proc/acpi",
-				"/proc/asound",
-				"/proc/kcore",
-				"/proc/keys",
-				"/proc/latency_stats",
-				"/proc/timer_list",
-				"/proc/timer_stats",
-				"/proc/sched_debug",
-				"/proc/scsi",
-			}),
+			// Unmask `/sys/firmware` to provide extra insight into the hardware of the
+			// underlying host, such as the number of CPU sockets on aarch64 variants
+			withUnmaskedPaths([]string{"/sys/firmware"}),
 			// Pass proxy environment variables to this container
 			withProxyEnv(),
 			// Add a default set of mounts regardless of the container type
@@ -1083,5 +1082,23 @@ func withDynamicResolver(ctx context.Context, ref string, registryConfig *Regist
 	default:
 		// For all other registries
 		return defaultResolver
+	}
+}
+
+// withUnmaskedPaths sets an alternate list of masked paths, less the paths provided
+func withUnmaskedPaths(unmaskPaths []string) oci.SpecOpts {
+	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
+		if s.Linux != nil && s.Linux.MaskedPaths != nil {
+			var maskedPaths []string
+			for _, path := range s.Linux.MaskedPaths {
+				if SliceContains(unmaskPaths, path) {
+					continue
+				}
+				maskedPaths = append(maskedPaths, path)
+			}
+			// Replace the default Linux.MaskedPaths with our own
+			s.Linux.MaskedPaths = maskedPaths
+		}
+		return nil
 	}
 }


jpculp:
Just tweaking some wording for clarity.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1085,3 +1074,25 @@ func withDynamicResolver(ctx context.Context, ref string, registryConfig *Regist
...
 	}
 }
+
+// withMaskedPaths sets an alternate list of masked paths
+func withMaskedPaths() oci.SpecOpts {
+	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
+		if s.Linux != nil && s.Linux.MaskedPaths != nil {
+			// Paths we want to unmask
+			unmaskPaths := []string{
+				"/sys/firmware",
```
yeazelm: Can we add a comment on why we want to unmask this? Just something to remind us why we went through the work.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
+	"k8s.io/utils/strings/slices"
```
arnaldo2792: How much will this increase the size of the binary? I'd prefer if we don't include unnecessary packages if that means pulling down the entire source and compile it. (We can always copy what `slices` do in your function). @jpmcb, will the compiler strip out the functions that aren't called from the package when `host-ctr` is built?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -301,20 +302,8 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
...
-				"/proc/kcore",
-				"/proc/keys",
-				"/proc/latency_stats",
-				"/proc/timer_list",
-				"/proc/timer_stats",
-				"/proc/sched_debug",
-				"/proc/scsi",
-			}),
+			// Pass an alternate list of masked paths
+			withMaskedPaths(),
```
arnaldo2792: :tada: nice!

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1085,3 +1074,25 @@ func withDynamicResolver(ctx context.Context, ref string, registryConfig *Regist
 		return defaultResolver
 	}
 }
+
+// withMaskedPaths sets an alternate list of masked paths
+func withMaskedPaths() oci.SpecOpts {
```
arnaldo2792: I think this function should let the caller pass what paths to unmask, that way the function is more usable in other places if we need to unmask depending on different conditions.


```suggestion
func withMaskedPaths(unmaskedPaths []string) oci.SpecOpts {
```

sources/host-ctr/cmd/host-ctr/main.go
jpculp: I don't believe it is increasing the binary size at all since it was already being pulled as an indirect dependency.

sources/host-ctr/cmd/host-ctr/main.go
jpculp: I was originally going to, but I thought we might want to be purposefully inflexible given the nature of masked paths.

sources/host-ctr/cmd/host-ctr/main.go
jpmcb: > since it was already being pulled as an indirect dependency.

^ this. Go binaries are statically linked so if there is a dependency that then pull in other dependencies, we get all of those as well. Since we already pull `k8s.io/utils` as an indirect dependency through `k8s.io/cri-api`, there shouldn't be any change to our the binary size and it's deps. I'm not sure what the Go compiler is doing under the hood for optimizations, but if there are functions/apis we aren't calling into, it shouldn't also bring those along. The `k8s.io/utils` [library is pretty thin too,](https://github.com/kubernetes/utils/blob/master/go.mod) so I'm not worried about it bringing in dependencies we don't want (like different versions of the k8s client-go or some containerd library).

This would be worrisome if we were bringing in a _new_ dependency but since it was already a `indirect` dependency, it should be ok.

Edit: see my comment below. If we don't need anything special from `k8s.io/utils`, we shouldn't use it.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1085,3 +1074,25 @@ func withDynamicResolver(ctx context.Context, ref string, registryConfig *Regist
...
+	return func(_ context.Context, _ oci.Client, _ *containers.Container, s *runtimespec.Spec) error {
+		if s.Linux != nil && s.Linux.MaskedPaths != nil {
+			// Paths we want to unmask
+			unmaskPaths := []string{
+				"/sys/firmware",
+			}
+			var maskedPaths []string
+			for _, path := range s.Linux.MaskedPaths {
+				if slices.Contains(unmaskPaths, path) {
```
jpmcb: If all we need `k8s.io/utils` for is a string slice checker, I'm in favor of dropping that dependency and writing it ourselves:

```diff
-                if slices.Contains(unmaskPaths, path) {
-                         continue
-                 }
+                for _, unmaskPath := range unmaskPaths {
+                    if unmaskPath == path {
+                        continue
+                    }
+                }
```

Because [a little copying is better than a little dependency.](https://youtu.be/PAAkCSZUG1c?t=569)

jpculp: I'm not opposed to this, but I guess I'm wondering if we should just skip the `[]string` altogether since we only have one path, and switch back if that is ever not the case (which it may never be).

jpmcb: I think handling a slice of length 1 is fine. And it makes the code abit more extensible too. I also like @arnaldo2792's suggestion of making the function signature:
```go
func withMaskedPaths(unmaskedPaths []string) oci.SpecOpts {
```
so then we can call this function like:
```go
WithMaskedPaths([]string{"/sys/firmware"})
```
and the function _body itself_ doesn't have any hard coded assumptions. Any changes in the future would just be changes to the arguments of a function, not the function itself.

But, I'm getting philosophic again, so it's up to you on how you'd like to handle the actual implementation, I just think removing `k8s.io/utils` is wise.

jpculp:
- Replaced direct dependency on `k8s.io/utils/strings/slices` with a similar function.
- Replaced `withMaskedPaths` to `withUnmaskedPaths`, which now takes a slice of paths.
- Added in-code comment as to why we want to unmask `/sys/firmware`.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -228,6 +228,16 @@ func (ct containerType) Prefix() string {
 	return ""
 }
 
+// SliceContains returns true if a slice contains a string
+func SliceContains(s []string, v string) bool {
```
jpmcb, line 232: nit: Do we anticipate exporting `SliceContains` out of this package (i.e., will other Go programs import this and use it)? If not, this should be `sliceContains` - it probably makes the most sense to keep it private to this package

jpmcb (approved):
😷 👍🏼

sources/host-ctr/cmd/host-ctr/main.go
jpculp: Not yet, but seemed useful enough that something else in the directory could potentially use it later on.

jpmcb: Fair enough! 👏🏼

arnaldo2792 (approved):

yeazelm (approved):

PR #2639, opened 2022-12-06
kubelet: Restrict access to TLS private key

### LINKED ISSUE #2638: Unable to run kube-bench job in Bottlerocket
**Image I'm using:**
"build_id": "104f8e0f",
"pretty_name": "Bottlerocket OS 1.11.1 (aws-k8s-1.23)",
"variant_id": "aws-k8s-1.23",

I am trying to run the kube-bench jobs to validate the CIS benchmarks for the EKS Cluster and expect the job to run

**What actually happened:**

The kube-bench job unable to access the `ca.crt` file in the `/etc/kubernetes/pki/ca.crt` location . 

1. Launch an EKS cluster with latest Bottlerocket Image and run the kube-bench jobs from https://github.com/aquasecurity/kube-bench
2. Run `kubectl logs -f kube-bench-*` 
3. Observe the error

stmcginnis:
**Issue number:**

Closes #2638 

**Description of changes:**

This moves the `tlsPrivateKeyFile` to be under a `private` subdirectory. This allows us to keep the public cert files readable for non-system processes that expect to be able to read them while limiting access to the more sensitive private key.

**Testing done:**

Built and deployed cluster.
Ran `kubectl apply -f job.yaml` with contents of the [kube-bench sample job](https://raw.githubusercontent.com/aquasecurity/kube-bench/main/job.yaml) with the exception of commenting out the mount of `/srv/kubernetes`.
Verified job completed successfully and no errors in the `kube-bench` logs:

```txt
...
== Summary total ==
14 checks PASS
0 checks FAIL
37 checks WARN
0 checks INFO
```

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.21/etc-kubernetes-pki.mount
+++ b/packages/kubernetes-1.21/etc-kubernetes-pki.mount
@@ -1,5 +1,5 @@
 [Unit]
-Description=Kubernetes PKI directory (/etc/kubernetes/pki)
+Description=Kubernetes PKI private directory (/etc/kubernetes/pki/private)
 DefaultDependencies=no
 Conflicts=umount.target
 Before=local-fs.target umount.target
@@ -8,7 +8,7 @@ Wants=selinux-policy-files.service
 
 [Mount]
 What=tmpfs
-Where=/etc/kubernetes/pki
+Where=/etc/kubernetes/pki/private
 Type=tmpfs
 Options=nosuid,nodev,noexec,noatime,context=system_u:object_r:secret_t:s0,mode=0700
 
--- a/packages/kubernetes-1.22/etc-kubernetes-pki.mount
+++ b/packages/kubernetes-1.22/etc-kubernetes-pki.mount
@@ -1,5 +1,5 @@
 [Unit]
-Description=Kubernetes PKI directory (/etc/kubernetes/pki)
+Description=Kubernetes PKI private directory (/etc/kubernetes/pki/private)
 DefaultDependencies=no
 Conflicts=umount.target
 Before=local-fs.target umount.target
@@ -8,7 +8,7 @@ Wants=selinux-policy-files.service
 
 [Mount]
 What=tmpfs
-Where=/etc/kubernetes/pki
+Where=/etc/kubernetes/pki/private
 Type=tmpfs
 Options=nosuid,nodev,noexec,noatime,context=system_u:object_r:secret_t:s0,mode=0700
 
--- a/sources/api/migration/migrations/v1.12.0/k8s-private-pki-path/src/main.rs
+++ b/sources/api/migration/migrations/v1.12.0/k8s-private-pki-path/src/main.rs
@@ -1,0 +1,81 @@
+#![deny(rust_2018_idioms)]
+
+use migration_helpers::{migrate, Migration, MigrationData, Result};
+use std::process;
+
+const SETTING: &'static str = "configuration-files.kubelet-server-key.path";
+const OLD_VALUE: &'static str = "/etc/kubernetes/pki/kubelet-server.key";
+const NEW_VALUE: &'static str = "/etc/kubernetes/pki/private/kubelet-server.key";
+
+/// We moved the render output location for the kubelet PKI private key to be in a restricted
+/// subdirectory. We need to update this output path in the stored configuration so updated nodes
+/// pick up the change.
+fn run() -> Result<()> {
+    migrate(KubeletServerKey {})
+}
+
+pub struct KubeletServerKey {}
+
+impl KubeletServerKey {
+    fn migrate(&mut self, mut input: MigrationData, action: &'static str) -> Result<MigrationData> {
+        let old_value;
+        let new_value;
+        if action == "upgrade" {
+            old_value = OLD_VALUE;
+            new_value = NEW_VALUE;
+        } else {
+            // Downgrade: everything old is new again
+            old_value = NEW_VALUE;
+            new_value = OLD_VALUE;
+        }
+
+        if let Some(data) = input.data.get_mut(SETTING) {
+            match data {
+                serde_json::Value::String(current_value) => {
+                    if current_value == old_value {
+                        *data = new_value.into();
+                        println!(
+                            "Changed '{}' from {:?} to {:?} on {}",
+                            SETTING, old_value, new_value, action
+                        );
+                    } else {
+                        println!(
+                            "'{}' is already set to {:?}, leaving alone",
+                            SETTING, new_value
+                        );
+                    }
+                }
+                _ => {
+                    println!(
+                        "'{}' is set to non-string value '{}'; KubeletServerKey only handles strings",
+                        SETTING, data
+                    );
+                }
+            }
+        } else {
+            println!("Found no setting '{}'", SETTING);
+        }
+
+        Ok(input)
+    }
+}
+
+impl Migration for KubeletServerKey {
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        self.migrate(input, "upgrade")
+    }
+
+    fn backward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        self.migrate(input, "downgrade")
+    }
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/models/shared-defaults/kubernetes-services.toml
+++ b/sources/models/shared-defaults/kubernetes-services.toml
@@ -40,7 +40,7 @@ path = "/etc/kubernetes/pki/kubelet-server.crt"
 template-path = "/usr/share/templates/kubelet-server-crt"
 
 [configuration-files.kubelet-server-key]
-path = "/etc/kubernetes/pki/kubelet-server.key"
+path = "/etc/kubernetes/pki/private/kubelet-server.key"
 template-path = "/usr/share/templates/kubelet-server-key"
 
 [configuration-files.kubelet-exec-start-conf]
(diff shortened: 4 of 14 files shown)


packages/kubernetes-1.22/etc-kubernetes-pki.mount
```diff
@@ -8,9 +8,9 @@ Wants=selinux-policy-files.service
...
 What=tmpfs
-Where=/etc/kubernetes/pki
+Where=/etc/kubernetes/pki/private
 Type=tmpfs
 Options=nosuid,nodev,noexec,noatime,context=system_u:object_r:secret_t:s0,mode=0700
 
 [Install]
-WantedBy=preconfigured.target
\ No newline at end of file
+WantedBy=preconfigured.target
```
jpmcb, line 16: Do the added newlines in these `.mount` files matter?

jpmcb (approved):
Nice! 👏🏼

packages/kubernetes-1.22/etc-kubernetes-pki.mount
stmcginnis: They shouldn't functionally matter. They were added here as there was some guidance to configured IDE settings to always have a newline at the end of files when editing. I've gotten a few review comments when my updates were missing that. So including them in this change as a trivial update.

jpmcb: Nice! Makes sense! Wasn't sure about `.mount` files

webern (dismissed):

sources/models/shared-defaults/kubernetes-services.toml
```diff
@@ -40,7 +40,7 @@ path = "/etc/kubernetes/pki/kubelet-server.crt"
 template-path = "/usr/share/templates/kubelet-server-crt"
 
 [configuration-files.kubelet-server-key]
-path = "/etc/kubernetes/pki/kubelet-server.key"
+path = "/etc/kubernetes/pki/private/kubelet-server.key"
```
arnaldo2792, line 43: @webern / @etungsten  isn't a migration required for this?

etungsten: Yes, a migration is necessary for this since we changed a default value. Existing Bottlerocket hosts won't get this new value after an update.

stmcginnis: Doesn't the templated file just get rewritten to the new location? No new files are being added. Do we have a past example if how that migration would need to work? I've only done them when new files were being added.

etungsten: This concerns the `configuration-files.kubelet-server-key.path` value in the settings datastore. These defaults are only applied upon first-boot after the settings datastore has been created. When we change the default and we want all hosts to start rendering a configuration file to a different `path` to fix some issue, we need to ensure existing host out there also use the correct `path` value here after upgrading. So we need a migration to check on upgrade that old value in the datastore, and if it is and we know it's incorrect, then we migrate it to the new value. 

We actually don't have an example migration of this kind, so you have to implement `Migration` trait. I can help you with that. Here's example migration you can reference: https://github.com/bottlerocket-os/bottlerocket/blob/3cbe1d843f3abcdc87efefd3679c7ece0378693e/sources/api/migration/migrations/archived/v1.1.0/shared-containerd-configs/src/main.rs

Instead of `template-path`, we're replacing the `path`.

stmcginnis: Added a migration. Please take a look and confirm that is what's needed when you get a chance. Thanks!

sources/api/migration/migrations/v1.12.0/k8s-private-pki-path/src/main.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn migrate(&mut self, mut input: MigrationData, action: &'static str) -> Result<MigrationData> {
+        let old_value;
+        let new_value;
+        if action == "upgrade" {
+            old_value = OLD_VALUE;
+            new_value = NEW_VALUE;
+        } else {
+            // Downgrade: everything old is new again
+            old_value = NEW_VALUE;
+            new_value = OLD_VALUE;
```
etungsten, line 29: If we know the old value is bad and shouldn't be used, do we still want to replace the good value with the old value on downgrade? Maybe there's a case for not doing anything on downgrade.

stmcginnis: If we're downgrading then the older version (v1.11.1) is going to expect the file to be in the old location. Or do I have a fundamental misunderstanding of what a downgrade is?

etungsten: Ah you're right. I missed that part. Kubelet will be looking for the file at that exact path. Ignore me!

etungsten (approved):
Looks good to me!

Small nit: I think it's better to separate out a commit just for the migration for maintenance and debugging but it's not a big deal.

stmcginnis:
>  I think it's better to separate out a commit just for the migration for maintenance and debugging but it's not a big deal.

Makes sense, I've separated it out into a second commit.

README.md
```diff
@@ -449,8 +449,8 @@ The following settings are optional and allow you to further configure your clus
   [default]
   region = us-west-2
   credential_process = aws_signing_helper credential-process \
-     --certificate /var/lib/kubelet/pki/kubelet-client-current.pem \
-     --private-key /var/lib/kubelet/pki/kubelet-client-current.pem \
+     --certificate /etc/kubernetes/pki/kubelet-server.crt \
+     --private-key /etc/kubernetes/pki/private/kubelet-server.key \
```
bcressey: Do we want to make this change? AFAIK `/var/lib/kubelet/pki/kubelet-client-current.pem` is the correct path for the default clusters created by EKS-A.

packages/kubernetes-1.21/etc-kubernetes-pki.mount
```diff
@@ -1,5 +1,5 @@
 [Unit]
-Description=Kubernetes PKI directory (/etc/kubernetes/pki)
+Description=Kubernetes PKI directory (/etc/kubernetes/pki/private)
```
bcressey:
```suggestion
Description=Kubernetes PKI private directory (/etc/kubernetes/pki/private)
```

bcressey (commented):
Changes LGTM.

What were all the warnings from `kube-bench`? That seems like a lot.

IIRC we should run the EKS-specific `job-eks.yaml` for `aws-k8s-*` to avoid some warnings about TLS certs and keys that aren't used.

Not sure if the same is true for `vmware-k8s-*` or `metal-k8s-*`, might be good to check.

README.md
stmcginnis: Yeah, that's probably better to keep that. Reverted the change.

packages/kubernetes-1.21/etc-kubernetes-pki.mount
stmcginnis: Done.

stmcginnis (commented):
> What were all the warnings from kube-bench? That seems like a lot.
> 
> IIRC we should run the EKS-specific job-eks.yaml for aws-k8s-* to avoid some warnings about TLS certs and keys that aren't used.

I didn't realize there was an EKS-specific option. I will run that and verify there are no errors. I was only looking to make sure there are no certificate related failures, so at least that part appears to be fine after this change. Will report back if there are any issues running the EKS tests.

stmcginnis:
Full output of running `jobs-eks.yaml`:

```txt
[INFO] 3 Worker Node Security Configuration
[INFO] 3.1 Worker Node Configuration Files
[PASS] 3.1.1 Ensure that the kubeconfig file permissions are set to 644 or more restrictive (Manual)
[PASS] 3.1.2 Ensure that the kubelet kubeconfig file ownership is set to root:root (Manual)
[PASS] 3.1.3 Ensure that the kubelet configuration file has permissions set to 644 or more restrictive (Manual)
[PASS] 3.1.4 Ensure that the kubelet configuration file ownership is set to root:root (Manual)
[INFO] 3.2 Kubelet
[PASS] 3.2.1 Ensure that the --anonymous-auth argument is set to false (Automated)
[PASS] 3.2.2 Ensure that the --authorization-mode argument is not set to AlwaysAllow (Automated)
[PASS] 3.2.3 Ensure that the --client-ca-file argument is set as appropriate (Manual)
[PASS] 3.2.4 Ensure that the --read-only-port argument is set to 0 (Manual)
[PASS] 3.2.5 Ensure that the --streaming-connection-idle-timeout argument is not set to 0 (Manual)
[PASS] 3.2.6 Ensure that the --protect-kernel-defaults argument is set to true (Automated)
[PASS] 3.2.7 Ensure that the --make-iptables-util-chains argument is set to true (Automated) 
[PASS] 3.2.8 Ensure that the --hostname-override argument is not set (Manual)
[WARN] 3.2.9 Ensure that the --eventRecordQPS argument is set to 0 or a level which ensures appropriate event capture (Automated)
[PASS] 3.2.10 Ensure that the --rotate-certificates argument is not set to false (Manual)
[PASS] 3.2.11 Ensure that the RotateKubeletServerCertificate argument is set to true (Manual)
[INFO] 3.3 Container Optimized OS
[WARN] 3.3.1 Prefer using Container-Optimized OS when possible (Manual)

== Remediations node ==
3.2.9 If using a Kubelet config file, edit the file to set eventRecordQPS: to an appropriate level.
If using command line arguments, edit the kubelet service file
/etc/systemd/system/kubelet.service.d/10-kubeadm.conf on each worker node and
set the below parameter in KUBELET_SYSTEM_PODS_ARGS variable.
Based on your system, restart the kubelet service. For example:
systemctl daemon-reload
systemctl restart kubelet.service

3.3.1 audit test did not run: No tests defined

== Summary node ==
14 checks PASS
0 checks FAIL
2 checks WARN
0 checks INFO

== Summary total ==
14 checks PASS
0 checks FAIL
2 checks WARN
0 checks INFO
```

I may investigate why this line shows up as a `WARN`, but that is not related to this change:

```txt
[WARN] 3.3.1 Prefer using Container-Optimized OS when possible (Manual)
```

stmcginnis:
It appears the test blindly warns on the container optimized OS regardless of the nodes being used: https://github.com/aquasecurity/kube-bench/blob/07e01cf38c54583cb21f281b97a67b0b05c8dc76/cfg/eks-1.1.0/node.yaml#L328

So very safe to ignore that. Would be nice if there was a consistent way to test and show that condition though.

bcressey (approved):

PR #2641, opened 2022-12-07
datastore: format with rustfmt

webern:
**Issue number:**

N/A

**Description of changes:**

I'm not sure why `datastore` was considered exempt from code formatting, but can I please fix it?

**Testing done:**

This is just a `cargo fmt`

**Terms of contribution:**

### DIFF
--- a/sources/api/datastore/src/serialization/mod.rs
+++ b/sources/api/datastore/src/serialization/mod.rs
@@ -34,7 +34,6 @@ fn bad_key<T>(typename: &str) -> Result<T> {
     error::BadMapKeySnafu { typename }.fail()
 }
 
-#[rustfmt::skip]
 impl ser::Serializer for &MapKeySerializer {
     type Ok = String;
     type Error = Error;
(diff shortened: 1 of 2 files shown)


stmcginnis (approved):
Nice!

sources/api/datastore/src/serialization/mod.rs
```diff
@@ -34,7 +34,6 @@ fn bad_key<T>(typename: &str) -> Result<T> {
     error::BadMapKeySnafu { typename }.fail()
 }
 
-#[rustfmt::skip]
```
jpmcb, line 37: 👏👏👏👏👏👏

jpmcb (approved):

etungsten:
We don't `cargo fmt` the serialization module because we call the same macro for a lot of the types we handle the same way and we considered it easier to read for them to bundled together with most of them written out as one line so all of them can be viewed together. I think it does make sense and `cargo fmt` doesn't grant much readability improvement here (similar to when we're using `maplit` to initialize HashMap, rustfmt always end up making it worse). 

Can we keep the `rustfmt::skip` and just format the function signatures and the `serialization_` functions where we don't call a macro and keep the other `bad_type` and `compound_type` serialization to one line, bundled together?

webern:
> Can we keep the rustfmt::skip and just format the function signatures and the serialization_ functions where we don't call a macro and keep the other bad_type and compound_type serialization to one line, bundled together?

No, I disagree with this logic entirely. It defeats the purpose of consistent code formatting provided by `rustfmt`. On one hand I get the original reasoning. These functions are simple so we want them on one line. On the other hand, these are the only special one-line functions in the project. The point of code formatting is for every function to be formatted the same way.

etungsten:
> The point of code formatting is for every function to be formatted the same way.

I'm not sure if I agree with this statement. But in any case I don't feel too strongly about this and the end result is still readable just across a larger file. 

I think the original motivator for this was, `serialize_str` (the actual accepted type) kept getting `rustfmt`'d into the middle of all the other `bad_types` which made it super hard to pick out, but maybe `rustfmt` is now smart enough not to do that.

etungsten (approved):

PR #2645, opened 2022-12-08
actions-workflow: switch to using 32-core github-hosted actions runners

etungsten:
I'll follow up this PR with another PR to add caching to our main `develop` branch so PR branches can use the cache to speed up workflows.

**Issue number:**
N/A

**Description of changes:**
```
    actions-workflow: switch to using 32-core github-hosted actions runners
    
    This switches our build workflow from using self-hosted runners to using
    32-core GitHub-hosted ubuntu runners. Currently they provide comparable
    performance and allow us to more easily use caching.
```

**Testing done:**
In this PR's Actions workflow. 

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,9 @@ concurrency:
 
 jobs:
   build:
-    runs-on: [self-hosted, linux, x64]
+    runs-on:
+      group: bottlerocket
+      labels: bottlerocket_ubuntu-latest_32-core
     continue-on-error: ${{ matrix.supported }}
     strategy:
       matrix:
@@ -120,7 +122,7 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
-      - run: rustup default 1.64.0
+      - run: rustup default 1.64.0 && rustup component add rustfmt && rustup component add clippy
       - run: cargo install --version 0.36.0 cargo-make
       - if: contains(matrix.variant, 'nvidia')
         run: |


jpmcb (approved):
Screen Shot 2022-12-08 at 1 45 17 PM

👏🏼 👏🏼 👏🏼

bcressey (approved):

.github/workflows/build.yml
```diff
@@ -120,7 +122,7 @@ jobs:
       fail-fast: false
     steps:
       - uses: actions/checkout@v3
-      - run: rustup default 1.64.0
+      - run: rustup default 1.64.0 && rustup component add rustfmt && rustup component add clippy
```
stmcginnis, line 125: I wonder if it would be better to switch to an action action for this: https://github.com/marketplace/actions/rust-toolchain

stmcginnis (approved):
Execution time is on par with our current self-hosted runners:

image

.github/workflows/build.yml
jpmcb: ^ that's a great idea. Then we can just get the image (and have that image cache on our runners) and not have to spend time getting all the stuff. Looks like it's `components` input supports adding `rustfmt`, `clippy`, etc.

etungsten: It seems like the `actions-rs` project has been unmaintained since Nov 2020. See https://github.com/actions-rs/toolchain/issues/216. I don't really feel comfortable using it now.

The alternative seems to be https://github.com/dtolnay/rust-toolchain, but that's meant for one-line use and not what we're doing with `cargo-make` and `cargo-sweep`. I guess I'm also not sure what the difference/benefit would be in switching to a dedicated rust-toolchain action since the default runner already has `rustup` and allows for customization of the toolchain we want to use.

stmcginnis: Yeah, fair enough. Sometimes the actions provide the benefit of speed due to well tested integrated caching mechanisms, but I don't think this is a big issue for us.

jpmcb: Yeah, def not a blocker. Good eye on the maintenance there. Like Sean said, usually a speed thing since we can get the benefit of having a cached image on the runner instead of having to rebuild the toolchain every time. But likely an marginal speed upgrade.

PR #2647, opened 2022-12-09
Improve reliability of `settings.network.hostname` generator

### LINKED ISSUE #2585: Hostname resolution
**What I'd like:**
Have more robust mechanism when BR sets the hostname of the instance. Currently we have:

>Most users don't need to change this setting as the following defaults work for the majority of use cases. If this setting isn't set we attempt to use DNS reverse lookup for the hostname. If the lookup is unsuccessful, the IP of the node is used. 

If the reverse lookup fails then the IP is used as hostname, maybe we can also check EC2 (DescribeInstance or maybe IMDS `http://169.254.169.254/latest/meta-data/hostname`)?

I would also want to know are there any retries to the initial DNS reverse lookup? 

The reason why I open this issue is because we had a EKS cluster running on BR, and one our etcd members was not starting up, problem was that the issued cert (issued to the hostname) was not matching the current hostname of the machine (which was IP address, because the reverse DNS failed). We were thinking about adding a bootstrap container that will set the hostname by doing (DescribeInstance or checking with IMDS).

### LINKED ISSUE #2600: `sundog` should log stderr even if the generator process returns 0
`sundog` [doesn't log anything](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/sundog/src/main.rs#L350) in the case a setting generator returns 0.

This means that setting generators don't have any way to log errors/warnings to a user, even in the case they can continue and finish properly.  

A good example of this is the generator for hostname: [`netdog generate-hostname`](https://github.com/bottlerocket-os/bottlerocket/blob/b530f30850d58ad84a8b0ded2e08fa91d554c7a8/sources/api/netdog/src/cli/generate_hostname.rs#L25).  The generator logs a message if the reverse DNS request for hostname fails, using the IP instead.  The failure message never gets logged to the journal.

cbgbt:
**Issue number:**

Closes #2585
Closes #2600

**Description of Changes:**
Alongside the title, this also introduces improved settings-generator logging from sundog.

```

    netdog: try harder to determine the hostname
    
    Failure scenarios when resolving a host's hostname are common enough to
    cause issues. This adds retries to our DNS queries for the current
    hostname.
    
    On AWS variants, we also attempt to query IMDS as a fallback mechanism.

---

    imdsclient: add method for fetching hostname

---

    sundog: always log settings generators' stderr

```

Just as a note, adding `tokio` and `imdsclient` moves netdog's binary size from `1.4M` -> `1.6M`.

**Testing done:**
I blocked DNS resolution in my Bottlerocket VPC at the network level, then ran `netdog generate-hostname` on the `aws-ecs-1` variant, noting that hostname generation succeeded despite DNS resolution failing.

Note that the first line is output to stderr, and the second to stdout.

```
bash-5.1# netdog generate-hostname
Reverse DNS lookup failed: failed to lookup address information: Temporary failure in name resolution
"ip-192-168-19-221.us-west-2.compute.internal"
```

I also tested sundog logging by attempting to boot with partial DNS blocking and checking the systemd journal:

```
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Sundog started
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Retrieving setting generators
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Retrieving settings values
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] shibaken started
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Received meta-data/services/partition
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] shibaken started
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Connecting to IMDS
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Fetching list of available public keys from IMDS
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Received meta-data/public-keys
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Generating targets to fetch text of available public keys
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Fetching public key (1/1)
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Received meta-data/public-keys/0/openssh-key
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Generating user-data
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Encoding user-data
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Setting generator command 'shibaken' stderr: 23:38:50 [INFO] Outputting base64-encoded user-data
Dec 09 23:38:50 localhost sundog[400]: 23:38:50 [INFO] Sending settings values to the API
Dec 09 23:38:50 localhost systemd[1]: Finished User-specified setting generators.
```

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/cli/generate_hostname.rs
+++ b/sources/api/netdog/src/cli/generate_hostname.rs
@@ -3,9 +3,21 @@ use crate::CURRENT_IP;
 use argh::FromArgs;
 use dns_lookup::lookup_addr;
 use snafu::ResultExt;
+use tokio::time::Duration;
+use tokio_retry::{
+    strategy::{jitter, FibonacciBackoff},
+    Retry,
+};
+
 use std::fs;
 use std::net::IpAddr;
 use std::str::FromStr;
+
+// Maximum number of retries for querying DNS for the hostname.
+const DNS_QUERY_MAX_RETRIES: usize = 3;
+// Starting retry interval for fibonacci backoff strategy.
+const DNS_QUERY_FIBONACCI_RETRY_INTERVAL: u64 = 50;
+// getnameinfo() takes about 5 seconds to time out - we add our own brief delay between retries.
 
 #[derive(FromArgs, PartialEq, Debug)]
 #[argh(subcommand, name = "generate-hostname")]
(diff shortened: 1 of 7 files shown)


sources/api/sundog/src/main.rs
```diff
@@ -350,7 +350,14 @@ where
         // Match on the generator's exit code. This code lays the foundation
         // for handling alternative exit codes from generators.
         match result.status.code() {
-            Some(0) => {}
+            Some(0) => {
+                if !result.stderr.is_empty() {
+                    let cmd_stderr = String::from_utf8_lossy(&result.stderr);
+                    for line in cmd_stderr.lines() {
+                        info!("{}: {}", command, line);
```
zmrow: Nit: perhaps the log could be a little bit more verbose so we know exactly what we're looking at in the logs. It could get a little confusing in the context of the whole journal.
```
info!("Setting generator command '{}' stderr: {}", command, line);
```

cbgbt:

zmrow (approved):
:face_in_clouds:

sources/api/netdog/src/cli/generate_hostname.rs
```diff
@@ -15,18 +29,42 @@ pub(crate) struct GenerateHostnameArgs {}
...
+        .await
+        .or_else(|e| {
             eprintln!("Reverse DNS lookup failed: {}", e);
-            ip_string
+            Err(e)
+        })
+        .ok();
+
+    // On AWS variants, we fallback to inspecting IMDS.
+    #[cfg(variant_platform = "aws")]
```
webern: This gives me pause. Elsewhere we have gone out of our way to segregate conditionally-compiled code from unconditionally-compiled code: https://github.com/bottlerocket-os/bottlerocket/blob/d20ea8efe911214ea635da81c2e8fed9c9703c77/sources/api/early-boot-config/src/provider.rs#L8..L21

I'm not sure that makes sense here, but I'm thinking about a case where we would want to query differently for a different provider  (vmware? or a different cloud, for example), and ending up with a string of sections of code with compiler guards.

Perhaps a compromise would be something like this:
(maybe it's too much trouble, but it's an idea)

```rust
    let hostname: Option<String> = Retry::spawn(retry_strategy(), || async { lookup_addr(&ip) })
        .await
        .or_else(|e| {
            eprintln!("Reverse DNS lookup failed: {}", e);
            ip_string
            Err(e)
        })
        .ok()
        .or_else(|| query_provider_hostname().await)
```

elsewhere:

```rust
async fn query_provider_hostname() -> Option<String> {
    #[cfg(variant_platform = "aws")]
    query_imds_hostname().await

    #[cfg(not(variant_platform = "aws"))]
    None
}

#[cfg(variant_platform = "aws")]
/// On AWS variants, we fallback to inspecting IMDS.
async fn query_imds_hostname() -> Option<String> {
    todo!()
}
```

webern (approved):

gthao313 (approved):
:tada:

sources/api/netdog/src/cli/generate_hostname.rs
cbgbt: I had similar worries but didn't give it enough thought to land at this strategy. I like it, I'll do this.

cbgbt:
* Sequestered conditionally compiled code into its own module per a recommendation from @webern
* Reduced the number of DNS lookups to 3 after noting that it takes ~5 seconds for glibc to timeout these requests

cbgbt:
^ One more change to add a missing docstring.

webern (approved):

cbgbt:
I forgot to run clippy 🤦

gthao313 (approved):

PR #2649, opened 2022-12-09
actions-workflow: add caching to develop and PR build workflows

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
    actions-workflow: add caching to build workflow for PRs
    
    This adds caching for `tools/`, `cargo-make`, `cargo-sweep`, and
    first-party crate dependencies. This should allow PRs to use the cache
    created by the 'cache' workflow on the develop branch and maintain their
    own separate branch caches.
```
```

    actions-workflow: add new cache workflow for develop
    
    This adds a caching workflow for the develop branch so PRs against
    develop can use the cache for 'tools/' and crate dependencies to save
    time when running through the build workflow.
```
```

    Makefile: simplify build-tools task dep
    
    We don't have a hard dependency on the bottlerocket-sdk for building
    buildsys. Removed `fetch-sdk` from `build-tools` dependencies and filled
    them in places where `build-tools` is an dependency.
    
    This makes it faster for us to build and cache buildsys for CI purposes

```

**Testing done:**
Tested in a separate repository. To see what cache is being used, take note of the cache key.

## First `cache` run on the default (develop) branch, expected cache miss and creates cache:
<details>

`cargo-make`, `cargo-sweep` and other 3rd party rust tools
```
Run actions/cache@v3
  with:
    path: ~/.cargo
  
    key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64
  env:
    OS_ARCH: x86_64
Cache not found for input keys: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64
...
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstdmt -cf cache.tzst --exclude cache.tzst -P -C  --files-from manifest.txt
Cache Size: ~364 MB (381604335 B)
Cache saved successfully
Cache saved with key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64
```
First-party crate dependency
```
Run actions/cache@v3
  with:
    path: .cargo
  
    key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-086c0ca2db45943dfabf3c82ff10899cbfe3560e6842327bf758d69f13e18d65
  env:
    OS_ARCH: x86_64
Cache not found for input keys: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-086c0ca2db45943dfabf3c82ff10899cbfe3560e6842327bf758d69f13e18d65
...
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstdmt -cf cache.tzst --exclude cache.tzst -P -C  --files-from manifest.txt
Cache Size: ~409 MB (429074235 B)
Cache saved successfully
Cache saved with key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-086c0ca2db45943dfabf3c82ff10899cbfe3560e6842327bf758d69f13e18d65
```
`tools/` build artifacts etc
```
Run actions/cache@v3
  with:
    path: tools/bin
  tools/.crates.toml
  tools/.crates2.json
  tools/target
  
    key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-951366615b3f3e5ca8be9c8279f05caf260bfe5321442cf109f622b96f902b27
  env:
    OS_ARCH: x86_64
Cache not found for input keys: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-951366615b3f3e5ca8be9c8279f05caf260bfe5321442cf109f622b96f902b27
...
..
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstdmt -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/ --files-from manifest.txt
Cache Size: ~443 MB (464083304 B)
Cache saved successfully
Cache saved with key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-951366615b3f3e5ca8be9c8279f05caf260bfe5321442cf109f622b96f902b27
```

**Total time: 11m 30s**
</details>

## `build` workflow on a PR against the default (develop) branch, gets cache-hit on the cache created by the `cache` workflow on initial run as expected. Creates a cache specific to the PR `build` workflow as expected:

<details>

`cargo-make`, `cargo-sweep` and other 3rd party rust tools
```
Run actions/cache@v3
Received 272629760 of 381604335 (71.4%), 259.7 MBs/sec
Received 381604335 of 381604335 (100.0%), 205.8 MBs/sec
Cache Size: ~364 MB (381604335 B)
/usr/bin/tar --use-compress-program unzstd -xf /home/runner/work/_temp/1632a57b-55a0-43cf-99b2-90a28b57d40d/cache.tzst -P -C ...
Cache restored successfully
Cache restored from key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64
...
...
 Post job cleanup.
Cache hit occurred on the primary key 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64, not saving cache.
```

First-party crate dependency
```
Run actions/cache@v3
Received 226492416 of 429074235 (52.8%), 215.8 MBs/sec
Received 429074235 of 429074235 (100.0%), 196.9 MBs/sec
Cache Size: ~409 MB (429074235 B)
/usr/bin/tar --use-compress-program unzstd -xf /home/runner/work/_temp/8704c94c-43d1-408d-84dc-57db97421e72/cache.tzst -P -C ...
Cache restored successfully
Cache restored from key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-086c0ca2db45943dfabf3c82ff10899cbfe3560e6842327bf758d69f13e18d65
...
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstdmt -cf cache.tzst --exclude cache.tzst -P -C ... --files-from manifest.txt
Cache Size: ~409 MB (429151951 B)
Cache saved successfully
Cache saved with key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-086c0ca2db45943dfabf3c82ff10899cbfe3560e6842327bf758d69f13e18d65-8157c59e0b7bc35821d0fa73af60a658fcf07b607c9bfbdc924539841d3114d7-cache-test-kick
```

`tools/` build artifacts etc
```
Run actions/cache@v3
Received 226492416 of 464083304 (48.8%), 214.9 MBs/sec
Received 464083304 of 464083304 (100.0%), 191.1 MBs/sec
Cache Size: ~443 MB (464083304 B)
/usr/bin/tar --use-compress-program unzstd -xf /home/runner/work/_temp/4d51eea0-8913-4c0a-9cef-7f7763a53fa7/cache.tzst -P -C ...
Cache restored successfully
Cache restored from key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-951366615b3f3e5ca8be9c8279f05caf260bfe5321442cf109f622b96f902b27
...
...
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstdmt -cf cache.tzst --exclude cache.tzst -P -C ... --files-from manifest.txt
Cache Size: ~782 MB (820347345 B)
Cache saved successfully
Cache saved with key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-951366615b3f3e5ca8be9c8279f05caf260bfe5321442cf109f622b96f902b27-8157c59e0b7bc35821d0fa73af60a658fcf07b607c9bfbdc924539841d3114d7-cache-test-kick
```
**Total runtime: 19 min 54 secs**

</details>

## Second `build` workflow run on the same PR against the default (develop) branch, gets cache-hit on the PR-specific cache as expected:

<details>

```
Run actions/cache@v3
Received 268435456 of 381604335 (70.3%), 255.5 MBs/sec
Received 381604335 of 381604335 (100.0%), 226.9 MBs/sec
Cache Size: ~364 MB (381604335 B)
/usr/bin/tar --use-compress-program unzstd -xf /home/runner/work/_temp/07055670-d4a7-4c76-9229-f5a71e88afa4/cache.tzst -P -C ...
Cache restored successfully
Cache restored from key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64
```
```
Run actions/cache@v3
Received 209715200 of 429151951 (48.9%), 199.8 MBs/sec
Received 429151951 of 429151951 (100.0%), 194.1 MBs/sec
Cache Size: ~409 MB (429151951 B)
/usr/bin/tar --use-compress-program unzstd -xf /home/runner/work/_temp/85651a97-2674-4ae0-90c7-7117fe0dcce5/cache.tzst -P -C...
Cache restored successfully
Cache restored from key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-086c0ca2db45943dfabf3c82ff10899cbfe3560e6842327bf758d69f13e18d65-8157c59e0b7bc35821d0fa73af60a658fcf07b607c9bfbdc924539841d3114d7-cache-test-kick
```
```

Run actions/cache@v3
Received 234881024 of 820347345 (28.6%), 223.8 MBs/sec
Received 473956352 of 820347345 (57.8%), 225.8 MBs/sec
Received 713031680 of 820347345 (86.9%), 226.4 MBs/sec
Received 820347345 of 820347345 (100.0%), 196.4 MBs/sec
Cache Size: ~782 MB (820347345 B)
/usr/bin/tar --use-compress-program unzstd -xf /home/runner/work/_temp/d05265df-8537-4f91-be86-d17e8455f321/cache.tzst -P -C ...
Cache restored successfully
Cache restored from key: 62035382743f7a544e93cbb11afa03977cb023b80037dd93cdda145a65605d4a-Linux-x86_64-951366615b3f3e5ca8be9c8279f05caf260bfe5321442cf109f622b96f902b27-8157c59e0b7bc35821d0fa73af60a658fcf07b607c9bfbdc924539841d3114d7-cache-test-kick
```
**Total runtime: 19m 59s**
</details>

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,9 +121,40 @@ jobs:
             fetch-upstream: "true"
       fail-fast: false
     steps:
+      - run: |
+          echo "OS_ARCH=`uname -m`" >> $GITHUB_ENV
       - uses: actions/checkout@v3
+      # Cache `cargo-make`, `cargo-cache`, `cargo-sweep`
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}
+      # Cache first-party rust code crate dependencies
+      - uses: actions/cache@v3
+        with:
+          path: |
+            .cargo
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}-${{ hashFiles('.github/workflows/build.yml') }}
+          restore-keys: |
+            ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}
+      # Cache 'tools/' dependencies and build artifacts
+      - uses: actions/cache@v3
+        with:
+          path: |
+            tools/bin
+            tools/.crates.toml
+            tools/.crates2.json
+            tools/target
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('tools/Cargo.lock') }}-${{ hashFiles('.github/workflows/build.yml') }}
+          restore-keys: |
+            ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('tools/Cargo.lock') }}
       - run: rustup default 1.64.0 && rustup component add rustfmt && rustup component add clippy
       - run: cargo install --version 0.36.0 cargo-make
+      - run: cargo install --version 0.6.2 cargo-sweep
+      - run: |
+          cargo sweep -i -r tools/
+          cargo sweep -t 7 -r tools/
       - if: contains(matrix.variant, 'nvidia')
         run: |
           cat <<-EOF > Licenses.toml
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -1,0 +1,58 @@
+# This workflow caches crate dependencies and build artifacts for tools (except 'test-tools' since we don't use them in build workflows).
+# The cache is only usable by workflows started from pull requests against the develop branch.
+name: CacheDepsAndTools
+on:
+  push:
+    branches: [develop]
+    paths:
+      - '.github/**'
+      - 'sources/Cargo.lock'
+      - 'tools/buildsys/**'
+      - 'tools/pubsys*/**'
+      - '!tools/pubsys/policies/**'
+      - '!tools/pubsys/**.example'
+      - '!tools/pubsys/**.template'
+      - 'tools/Cargo.lock'
+jobs:
+  cache:
+    runs-on:
+      group: bottlerocket
+      labels: bottlerocket_ubuntu-latest_16-core
+    continue-on-error: true
+    steps:
+      - run: |
+          echo "OS_ARCH=`uname -m`" >> $GITHUB_ENV
+      - uses: actions/checkout@v3
+      # Cache `cargo-make`, `cargo-cache`, `cargo-sweep`
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}
+      # Cache first-party code dependencies
+      - uses: actions/cache@v3
+        with:
+          path: |
+            .cargo
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}
+      # Cache 'tools/' dependencies and build artifacts
+      - uses: actions/cache@v3
+        with:
+          path: |
+            tools/bin
+            tools/.crates.toml
+            tools/.crates2.json
+            tools/target
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('tools/Cargo.lock') }}
+      - run: rustup default 1.64.0
+      - run: cargo install --locked --version 0.36.0 cargo-make
+      - run: cargo install --locked --version 0.8.3 --no-default-features --features ci-autoclean cargo-cache
+      - run: cargo install --locked --version 0.6.2 cargo-sweep
+      - run: |
+          cargo sweep -i -r tools/
+          cargo sweep -t 7 -r tools/
+      - run: cargo make publish-setup-tools
+      - run: cargo make publish-tools
+      - run: cargo make build-tools
+      # This cleans the cargo cache in ~/.cargo
+      - run: cargo-cache
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -484,7 +484,7 @@ done'''
 ]
 
 [tasks.build-tools]
-dependencies = ["fetch"]
+dependencies = ["setup", "fetch-sources"]
 script = [
 '''
 cargo install \
@@ -616,7 +616,7 @@ docker run --rm \
 
 # Builds a package including its build-time and runtime dependency packages.
 [tasks.build-package]
-dependencies = ["check-cargo-version", "build-tools", "publish-setup", "fetch-licenses"]
+dependencies = ["check-cargo-version", "fetch-sdk", "build-tools", "publish-setup", "fetch-licenses"]
 script_runner = "bash"
 script = [
 '''
@@ -649,7 +649,7 @@ cargo build \
 ]
 
 [tasks.build-variant]
-dependencies = ["build-tools", "publish-setup"]
+dependencies = ["fetch-sdk", "build-tools", "publish-setup"]
 script = [
 '''
 export PATH="${BUILDSYS_TOOLS_DIR}/bin:${PATH}"


.github/workflows/cache.yml
```diff
@@ -0,0 +1,58 @@
+# This workflow caches crate dependencies and build artifacts for tools (except 'test-tools' since we don't use them in build workflows).
+# The cache is only usable by workflows started from pull requests against the develop branch.
+name: CacheDepsAndTools
```
jpmcb, line 3: I'm not following why we need an entirely separate workflow for creating caches? Isn't this redundant since the build workflow (which runs on PRs against the develop branch and covers the same paths as this workflow) has it's own steps for creating / accessing caches?

.github/workflows/cache.yml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+      - run: |
+          echo "OS_ARCH=`uname -m`" >> $GITHUB_ENV
+      - uses: actions/checkout@v3
+      # Cache `cargo-make`, `cargo-cache`, `cargo-sweep`
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}
```
jpmcb, line 31: `.github/workflows/cache.yaml` shouldn't be used as the exclusive hash file. Changes to this file will cause caches to be rebuilt that don't have any association to this file. (a.k.a, files in `.github` don't correlate to cargo dependencies)

Doesn't it make more sense to use some `Cargo.lock` as a hash file since that is what will change as dependencies change?

I'd imagine this will frequently result in a incorrect hits since `.github/workflows/cache.yaml` won't change or unnecessary cache rebuilds if we make arbitrary changes to files in `.github`

jpmcb: > Cache `cargo-make`, `cargo-cache`, `cargo-sweep`

Or is the idea that this _should_ result in a frequent hit (and we don't want to change this file) since `cargo-make`, `cargo-cache`, and `cargo-sweep` are tools that are directly associated with this workflow .. ? I'm not understanding something here 🤔

.github/workflows/cache.yml
```diff
@@ -0,0 +1,58 @@
...
+      # Cache first-party code dependencies
+      - uses: actions/cache@v3
+        with:
+          path: |
+            .cargo
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}
```
jpmcb, line 37: > hashFiles('.github/workflows/cache.yml')

If we need to touchup `.github/lworkflows/cache.yml` _without_ changing our first party code dependencies, then the cache here will need to be rebuilt since the first part of the key has changed?

.github/workflows/build.yml
```diff
@@ -121,9 +121,42 @@ jobs:
...
+        with:
+          path: |
+            ~/.cargo
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}
+      # Cache first-party rust code crate dependencies
+      - uses: actions/cache@v3
+        with:
+          path: |
+            .cargo
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}-${{ hashFiles('.github/workflows/build.yml') }}-${{ github.head_ref }}
```
jpmcb:
```
${{ github.head_ref }}
```

as part of the hash key seems redundent. If the action runs, and creates a cache with key:
```
{{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}
```
 it'll create a cache general use cache that is associated with:
- that runner's os
- that runner's arch
- the hash from the first party code dependencies

If we need to run that action again, it'll hit that same cache (since rerunning it, non of those 3 variables should have changed).

I believe simplifiying the keys will also cause fewer caches to be created in general since we won't have to worry about arbitrary files or variable (like in those files in `.github`). Instead, general use caches will be generated that can be used across many different runs or many different PRs.

.github/workflows/build.yml
```diff
@@ -121,9 +121,42 @@ jobs:
...
+          restore-keys: |
+            ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}-${{ hashFiles('.github/workflows/build.yml') }}
```
jpmcb: Similar to the above, to simplify, why not just make the access key:
```
{{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}
```
so it can be used across many different PRs or many different runs?

jpmcb (commented):
A few stream of consciousness comments here but generally, I'm confused why we need an entirely separate cache building workflow that triggers off files in the `tools` directory. What benefit does that have over letting the main build workflow run and build / access caches it creates (since we run `build` on almost all PRs)?

It also seems like we wouldn't want to rebuild or access caches based on files in the `.github/` directory. Simple yaml changes there could trigger all caches to be recreated.

I think generalizing the caches as much as possible results in fewer caches being created and more usage across many different runs or different PRs.

.github/workflows/cache.yml
etungsten: Different PR build workflows can't share caches. Merged PR's cache can't be re-used by other PRs even if it goes to develop. However, you can use caches from the default branch (develop). This is a major point of confusion with github action caches. https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

.github/workflows/cache.yml
etungsten: We should invalidate the cache when the cache workflow changes. e.g. maybe we no longer want to cache something or we add something to be cached.

.github/workflows/cache.yml
etungsten: As for why we generate a cache for develop, we want the initial build workflow by a fresh PR be able to use cached tools so it doesn't have to suffer the 8-10 min penalty in building it and creating it's own fresh cache.

.github/workflows/cache.yml
etungsten: Yes, see https://github.com/bottlerocket-os/bottlerocket/pull/2649#discussion_r1044859121. The develop cache is important because all PRs are able to use it, so if something changes about that cache workflow, we want to refresh the develop cache from the cache workflow.

.github/workflows/build.yml
etungsten: You can't share caches across PRs. Each PR build workflow cache is specific to that PR only. The reason why we append the `github.head_ref` is so that the PR may make changes to the `Lockfiles` and we want to cache the changes for **only** that PR workflow. 

If we don't append `github.head_ref`, then it won't create a cache for the PR since that cache key already exists and on subsequent workflow execution it'll keep restoring the cache from the cache workflow on develop when we want to use the cached changed dependency/whatever based on that PR's changes.

See https://github.com/actions/cache#cache-scopes
>  If the provided key matches an existing cache, a new cache is not created and if the provided key doesn't match an existing cache, a new cache is automatically created provided the job completes successfully.

.github/workflows/build.yml
etungsten: See https://github.com/bottlerocket-os/bottlerocket/pull/2649#discussion_r1044863100

.github/workflows/build.yml
etungsten: > I believe simplifiying the keys will also cause fewer caches to be created

There is no limit in the number of caches we can have at a time. By default caches that aren't used over 10 days or something gets purged. We do have a limit on total cached size, around 10 GB.

.github/workflows/cache.yml
etungsten: Also see https://github.com/actions/cache/issues/843#issuecomment-1179771256

> You will get a cache miss on a different branch if the cache save happened on a different branch **with the same key, unless the save happens in the main/master branch**

Also https://github.com/actions/cache#cache-scopes

.github/workflows/cache.yml
jpmcb: That's fair! I guess we're not losing much by adding that as a cache key!

.github/workflows/build.yml
jpmcb: > You can't share caches across PRs.

Yikes, that's new to me! I always thought github action caches shared across all branches regardless of the base branch. Thanks for clarifying!

> If we don't append github.head_ref, then it won't create a cache for the PR since that cache key already exists and on subsequent workflow execution it'll keep restoring the cache from the cache workflow on develop when we want to use the cached changed dependency/whatever based on that PR's change

I thought that the `sources/Cargo.lock` file the workflow is using is based on the PR's branch and the cache action would use that as the key. So wouldn't the hash for that lockfile be unique to that PR's branch and cause a new cache to be created?

Example:
- `develop` has `source/Cargo.lock` with hash `abc123`
- Some branch `my-feature` makes dependency changes (which change the `sources/Cargo.lock` file). So _now_, on that branch, the `sources/Cargo.lock` file has a hash of `xyz789`
- When a PR is opened from `my-feature` to `develop`, since we first use `action/checkout`, the [`GITHUB_REF` points to the PR merge branch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) and [`GITHUB_SHA` point to the last commit on that merge branch ](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request). So for all purposes, the file being inspected under `sources/Cargo.lock` will have the hash `xyz789` in the workflow. This creates a different key and will generate a new cache for this branch.
- Subsequent runs of workflows under this PR will have scoped access to that cache (unless `sources/Cargo.lock` changes again with some additional push.)

What am i missing here?

.github/workflows/build.yml
jpmcb: Resolving this convo and focusing on above thread ^

.github/workflows/cache.yml
jpmcb: > However, you can use caches from the default branch (develop).

I see, that's a great idea! I think the key piece here I was missing was:

```yaml
on:
  push:
    branches: [develop]
```

Which runs (and creates caches) when new commits are pushed to develop. These then can be accessed by subsequent PRs 👏🏼

.github/workflows/build.yml
etungsten: Right, that scenario is the desired behavior. 

> So wouldn't the hash for that lockfile be unique to that PR's branch and cause a new cache to be created?

I guess that's true, if the PR changes the lockfiles, it's guaranteed to cache-miss on the initial build workflow and have to generate their own. That's also the reason why I split it into 3 cache actions so some parts can still use the cache even if only `sources/Cargo.lock` is changed.

I'll remove the `github_ref` from the key.

jpmcb: Thanks for talking through this with me and helping me understand what's happening! The GitHub cache actions are for sure weird and confusing to get right. I know @stmcginnis also has good experience with them and can weigh in here as well!

etungsten:
Push above removes `${{ github.head_ref }}` from the cache key since it's redundant in the presence of `${{ hashFiles('.github/workflows/build.yml') }}`.

jpmcb (approved):

stmcginnis (approved):
I thought you could share across PRs too, but it's been awhile now since I've set workflow caching and it's entirely possible things have changed or I misunderstood from the start. ;)

This looks good to me and should improve some speed. Nice work!

.github/workflows/cache.yml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          path: |
+            tools/bin
+            tools/.crates.toml
+            tools/.crates2.json
+            tools/target
+          key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('tools/Cargo.lock') }}
+      - run: rustup default 1.64.0
+      - run: cargo install --locked --version 0.36.0 cargo-make
+      - run: cargo install --locked --version 0.8.3 --no-default-features --features ci-autoclean cargo-cache
```
jpculp, line 49: I wonder if we really need to pin the versions for `ci-autoclean`, `cargo-cache`, and `cargo-sweep`. I worry that we'll either forget to bump them, or worse bump them in some places but not in others. I guess the same could be said for `cargo-make` as well.

etungsten: I'm more afraid of breaking changes bring introduced. We've been burned by that with `cargo-make` more than once. This freezes with versions of the tools we know that will work.

jpculp: I think `cargo-make` has a higher chance of that occurring since it is dealing with the build itself. At least with `cargo-cache` I've never really ran into a regression. It seems like something occurring in these housekeeping packages is less likely, and would be easier to tackle if something did occur.

stmcginnis: I've been burned by new breaking changes too, so I'm usually more comfortable explicitly pinning like this and making sure upgrades are a conscious decision.

If we want to try to keep up to date, dependabot can be configured to propose GitHub Action updates for these too.

jpculp: Oh, I really like the idea of dependabot doing it. That way we can get a heads up of when we do get a breaking change, but it wouldn't break our PRs across the board. It would also make sure all workflows are getting the same bumps.

jpmcb: > I'm more afraid of breaking changes bring introduced.

> I've been burned by new breaking changes too

+1 to that - Also in favor of pinning versions of these tools. And yes! I also like the idea of getting dependabot to do it. However, speaking of, _where in our project do we declare those things as dependencies?_ The only place I see it [defined as a dependency is in our `BUILDING.md` file.](https://github.com/bottlerocket-os/bottlerocket/blob/develop/BUILDING.md#rust) In order for depndabot to manage it, we'd have to codify it in some `Cargo.toml` file or something.

stmcginnis: Dependabot understands these things separately from Rust or Go code: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

We would just need to add that configuration file. But we might have dependabot completely disabled on this main repo right now, so we would probably need someone with admin rights to turn that one once we have the config file in place.

stmcginnis: Figured I might as well just throw that up there: https://github.com/bottlerocket-os/bottlerocket/pull/2655

etungsten: We disabled dependabot PRs for the entire repository since it was introducing way too much PR noise for rust crate bumps. We know to do that before every release so there was limited-to-no value with the dependabot PRs. We still have security notification enabled which we do rely on from time to time.

`cargo-make` introduced breaking changes between patch version before so limiting the notifications to just minor version bumps doesn't really help and just introduces PR noise. And honestly, since these `cargo-*` tools are extremely limited in scope (as with the workflow itself), it's very hard for any kind of security vulnerability with them to affect us in any meaningful way (e.g. we don't use the build artifacts etc). At worse, `cargo-make` somehow would give us a false positive or caches fill up because `cargo-sweep` doesn't work (but we know it does right now)

jpmcb: The above PR looks good to me, however, we should note that it will only keep _Actions_ we `use` up to date, not the cargo tools we install. For example, if there's an upgrade to `actions/cache`, it'll open a PR to upgrade that action.

> When you enable Dependabot version updates for GitHub Actions, Dependabot will help ensure that references to actions in a repository's workflow.yml file are kept up to date.

[Source](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#about-dependabot-version-updates-for-actions)

We'd still need to codify our dependency of `cargo-make`, etc. we are pinned to to have dependabot upgrade it for us.

Makefile.toml
```diff
@@ -484,7 +484,7 @@ done'''
 ]
 
 [tasks.build-tools]
-dependencies = ["fetch"]
+dependencies = ["setup", "fetch-sources"]
```
webern, line 487: just checking, is "fetch-vendored" worth caching as well? Or maybe that is happening already.

webern (approved):

webern (approved):

Makefile.toml
etungsten: I was thinking about caching `.gomodcache` but it really only provides about a 2-3 second benefit right now with caching it and would require us to have a separate cache step that is scoped to just the host-ctr and ecs-init `go.sum` file changes.

.github/workflows/cache.yml
jpculp: > We disabled dependabot PRs for the entire repository since it was introducing way too much PR noise for rust crate bumps.

Right, but if the scope is limited to the couple crates of actions it shouldn't be too bad. We definitely don't want to turn dependabot on for the whole repo again.

PR #2650, opened 2022-12-09
[wicked] Include MAC address in the lease files

zmrow:
**Issue number:**
Related to/enables #2293

**Description of changes:**
Adds a patch to `wicked` which includes the MAC address of the interface to the DHCP and static lease files.

Adding the MAC address to these lease files enables us to correlate a MAC with an interface name.

**Testing done:**
* Boot an `aws-k8s-1.22` variant: joins the cluster and runs pods just fine.  Verify the MAC shows up in the DHCP lease file:
```
 bash-5.1# cat /run/wicked/leaseinfo.eth0.dhcp.ipv4 
MACADDR='0a:17:17:fb:71:bd'
INTERFACE='eth0'
...
```
* Boot a `metal-dev` variant with static address settings and verify the mac shows up in the static lease file
```
bash-5.1# cat /run/wicked/leaseinfo.eno1.static.ipv4
MACADDR='3c:ec:ef:79:ae:ac'
INTERFACE='eno1'
```

**Terms of contribution:**

### DIFF
--- a/packages/wicked/1007-include-mac-address-in-leaseinfo.patch
+++ b/packages/wicked/1007-include-mac-address-in-leaseinfo.patch
@@ -1,0 +1,31 @@
+diff --git a/src/leaseinfo.c b/src/leaseinfo.c
+index c47241f5..86265b45 100644
+--- a/src/leaseinfo.c
++++ b/src/leaseinfo.c
+@@ -692,7 +692,26 @@ __ni_leaseinfo_dump(FILE *out, const ni_addrconf_lease_t *lease,
+ {
+ 	unsigned int i;
+ 	char *key = NULL;
++	ni_netconfig_t *nc;
++	ni_netdev_t *dev;
+ 
++	/* Find the network device using the name, and print the MAC in the
++	 * lease file */
++	if (!(nc = ni_global_state_handle(0))) {
++		ni_warn("Unable to query global network interface state");
++		goto write_lease;
++	}
++
++	if (!(dev = ni_netdev_by_name(nc, ifname))) {
++		ni_warn("Unable to find interface with name %s", ifname);
++		goto write_lease;
++	}
++
++	__ni_leaseinfo_print_string(out, prefix, "MACADDR",
++			ni_link_address_print(&dev->link.hwaddr),
++			"", 0);
++
++write_lease:
+ 	__ni_leaseinfo_print_string(out, prefix, "INTERFACE", ifname, "", 0);
+ 
+ 	/* wicked specific vars */
--- a/packages/wicked/wicked.spec
+++ b/packages/wicked/wicked.spec
@@ -46,6 +46,7 @@ Patch1003: 1003-ship-mkconst-and-schema-sources-for-runtime-use.patch
 Patch1004: 1004-adjust-safeguard-for-dhcp6-defer-timeout.patch
 Patch1005: 1005-client-validate-ethernet-namespace-node.patch
 Patch1006: 1006-server-discover-hardware-address-of-unconfigured-int.patch
+Patch1007: 1007-include-mac-address-in-leaseinfo.patch
 
 BuildRequires: %{_cross_os}glibc-devel
 BuildRequires: %{_cross_os}libdbus-devel


packages/wicked/1007-include-mac-address-in-leaseinfo.patch
```diff
@@ -0,0 +1,23 @@
+diff --git a/src/leaseinfo.c b/src/leaseinfo.c
+index c47241f5..95f7fb0e 100644
+--- a/src/leaseinfo.c
++++ b/src/leaseinfo.c
+@@ -692,6 +692,18 @@ __ni_leaseinfo_dump(FILE *out, const ni_addrconf_lease_t *lease,
+ {
+ 	unsigned int i;
+ 	char *key = NULL;
++	ni_netconfig_t *nc = ni_global_state_handle(0);
```
zmrow: There is prior art for querying a device in this manner in [update.c](https://github.com/openSUSE/wicked/blob/5e50cdbb5d74490a9d728c0d0be537985bb65b38/src/update.c#L503-L511) which is actually where `leaseinfo_dump()` is called.

packages/wicked/1007-include-mac-address-in-leaseinfo.patch
```diff
@@ -0,0 +1,23 @@
...
++	ni_netdev_t *dev;
```
webern, line 10:
```suggestion
+	ni_netdev_t *dev = NULL;
```

packages/wicked/1007-include-mac-address-in-leaseinfo.patch
bcressey: It might be better to check the result of the `ni_global_state_handle()` call for errors - I see that's done elsewhere in the code. If you're certain there's no way it can fail here then it's OK.

packages/wicked/1007-include-mac-address-in-leaseinfo.patch
```diff
@@ -0,0 +1,23 @@
...
++
++	/* Find the network device using the name, and print the MAC in then
++	 * lease file */
++	if (!(dev = ni_netdev_by_name(nc, ifname))) {
++		ni_error("Unable to find interface with name %s", ifname);
++		return;
++	}
```
bcressey: Rather than erroring out, maybe emit a warning message and skip writing the `MACADDR` field.

packages/wicked/1007-include-mac-address-in-leaseinfo.patch
cbgbt: Seems like following the same logic of printing a warning and continuing without printing `MACADDR` is probably the safest bet if you get a NULL back.

zmrow:

cbgbt (approved):

packages/wicked/1007-include-mac-address-in-leaseinfo.patch
```diff
@@ -0,0 +1,28 @@
+diff --git a/src/leaseinfo.c b/src/leaseinfo.c
+index c47241f5..8a827ef7 100644
+--- a/src/leaseinfo.c
++++ b/src/leaseinfo.c
+@@ -692,6 +692,23 @@ __ni_leaseinfo_dump(FILE *out, const ni_addrconf_lease_t *lease,
+ {
+ 	unsigned int i;
+ 	char *key = NULL;
++	ni_netconfig_t *nc = ni_global_state_handle(0);
```
bcressey: This assignment is clobbered by the second call, so you should remove it:
```suggestion
+	ni_netconfig_t *nc;
```

packages/wicked/1007-include-mac-address-in-leaseinfo.patch
```diff
@@ -0,0 +1,28 @@
...
++			__ni_leaseinfo_print_string(out, prefix, "MACADDR",
++					ni_link_address_print(&dev->link.hwaddr),
++					"", 0);
++		} else {
++			ni_warn("Unable to find interface with name %s", ifname);
++		}
++	} else {
++		ni_warn("Unable to query global network interface state");
++
++	}
```
bcressey: The preferred way to write this would be something like:
```
        if !(nc = ni_global_state_handle(0)) {
                ni_warn("Unable to query global network interface state");
                goto write_lease;
        }

        if !(dev = ni_netdev_by_name(nc, ifname)) {
                ni_warn("Unable to find interface with name %s", ifname);
                goto write_lease;
        }

        __ni_leaseinfo_print_string(out, prefix, "MACADDR",
                        ni_link_address_print(&dev->link.hwaddr),
                        "", 0);

write_lease:
        __ni_leaseinfo_print_string(out, prefix, "INTERFACE", ifname, "", 0);
```

zmrow: :)  Thanks for the help!

zmrow:
^ Removed the nested logic

zmrow:
^ Fix the clobbered variable

packages/wicked/1007-include-mac-address-in-leaseinfo.patch
cbgbt: Probably borrowed from other code that calls `ni_global_state_handle()`, there seems to be similar clobbering elsewhere in wicked.

cbgbt (approved):

bcressey (approved):

webern (approved):
Didn't dive into the APIs but LGTM

foersleo (approved):
Looks good to me. Do we have any plan to contribute this back to wicked upstream at some point?

zmrow:
@foersleo I could open the PR, not sure if this "feature" is something anyone else wants! :)

foersleo:
@zmrow I was mostly wondering what our model there would be. In general I think we would want to minimize our downstream patch set size for each package as it reduces maintenance burden for us, might help someone else, and get some additional kind of vetting from someone that is potentially more intimately familiar with the code base.

I obviously do not think we should block our progress on this discussion, but it possibly is a worthwhile discussion to have on a more general level.

zmrow:
Also double checked the compiler output for this package - no warnings/issues related to this change.  (There were a few warnings in distant files not touched here)

zmrow:
> we would want to minimize our downstream patch set size for each package as it reduces maintenance burden for us, might help someone else, and get some additional kind of vetting from someone that is potentially more intimately familiar with the code base.

Fully agree there - I can do that

PR #2653, opened 2022-12-12
testsys: Add `testsys-image-tag` to  `Test.toml`

ecpullen:
**Issue number:**

N/A

**Description of changes:**

Allow users to specify the tag that should be used for testsys image pulls when using an alternate `testsys-image-registry`. This is especially useful for `bottlerocket-test-system` when testing new images.

**Testing done:**

I ran `cargo make test` on custom testing images with `latest` tag and verified that the images used by the agent were the ones with the correct tag.

**Terms of contribution:**

### DIFF
--- a/tools/testsys/Test.toml.example
+++ b/tools/testsys/Test.toml.example
@@ -11,6 +11,9 @@ repo = "default"
 # The registry containing alternate TestSys agent images
 testsys-images-registry = "public.ecr.aws/bottlerocket-test-system"
 
+# The tag that should be used with `testsys-images-registry` for image pulls
+testsys-images-registry = "latest"
+
 # The URI for the EKS resource agent that should be used. An individual agent's provided URI will be
 # used even if `testsys-images-registry` is present.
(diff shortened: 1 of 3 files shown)


tools/testsys/Test.toml.example
```diff
@@ -11,6 +11,9 @@ repo = "default"
 # The registry containing alternate TestSys agent images
 testsys-images-registry = "public.ecr.aws/bottlerocket-test-system"
 
+# The tag that should be used with `testsys-images-registry` for image pulls
+testsys-images-registry = "latest"
```
etungsten, line 15: Do we push a "latest" tag for the TestSys images? I don't think we do for the other container images we publish under Bottlerocket, should we follow that rule here?

ecpullen: We do not push to the latest tag. We only push the testsys version. Due to image cacheing in k8s clusters, latest should be used when debugging new images. This feature allows someone testing new agents to specify a tag and a registry instead of explicitly listing each agent.

etungsten (approved):

webern (approved):

jpculp (approved):

PR #2655, opened 2022-12-12
Add Action dependabot config

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This adds a dependabot configuration to regularly check for updates to our GitHub Actions.

**Testing done:**

Will see once in place and change/update/remove as needed based on how that goes.

**Terms of contribution:**

### DIFF
--- a/dependabot.yaml
+++ b/dependabot.yaml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every month
+      interval: "monthly"


webern (approved):

jpmcb (approved):

dependabot.yaml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every month
+      interval: "monthly"
```
jpculp, line 8: I wish GH supported something between "weekly" and "monthly".

jpculp (approved):

dependabot.yaml
stmcginnis: Unfortunately the [only options](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval) are "daily", "weekly", and "monthly". I think this ends up fine though. Based on experience with past projects that have used this, there really aren't too frequent of updates.

stmcginnis:
> Runs once each month. This is on the first day of the month.

Will make a note to check on things in January and see if additional repo configuration is needed to actually enable this. Worst case, it's just an extra file in the repo and we can remove it.

PR #2666, opened 2022-12-15
actions-workflow: install build deps before running build workflow

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
    actions-workflow: install build deps before running build workflow
    
    Ensure that build dependencies are installed on the runner before the
    build workflow.

```

**Testing done:**
To be seen in the build workflow

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,8 +121,10 @@ jobs:
             fetch-upstream: "true"
       fail-fast: false
     steps:
-      - run: |
+      - name: Preflight step to set up the runner
+        run: |
           echo "OS_ARCH=`uname -m`" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):

.github/workflows/build.yml
```diff
@@ -123,6 +123,7 @@ jobs:
     steps:
       - run: |
           echo "OS_ARCH=`uname -m`" >> $GITHUB_ENV
+          sudo apt -y install lz4
```
jpculp: Rather than `lz4`, I think you can get away with `liblz4-tool`. Along those lines, I think you should just go ahead and add all the packages we detail in our [BUILDING.md](https://github.com/bottlerocket-os/bottlerocket/blob/develop/BUILDING.md). If some of those packages aren't necessary then they should be removed from that doc.

jpculp (approved):

PR #2678, opened 2022-12-20
actions-workflow: user github runners for golang lint workflow

etungsten:
**Issue number:**
N/A

**Description of changes:**
We no longer have self-hosted runners.
Use 8-core github-hosted runners for the golang-lint check in actions workflow.

**Testing done:**
See Actions for this PR

**Terms of contribution:**

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -2,16 +2,19 @@ name: golangci-lint
 on:
   pull_request:
     branches: [develop]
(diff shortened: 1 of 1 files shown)


jpmcb (approved):
👏🏼 Nice!!

stmcginnis (approved):
PR description says 16-core, so I was going to ask if we could actually run this on smaller instances. Looks like 8-core, so that's actually good.

PR #2679, opened 2022-12-21
actions-workflow: use free github runners for golang lint workflow

etungsten:
**Issue number:**
N/A

**Description of changes:**
Instead of using 8-core github runners, which costs a bit of money. The default ubuntu runner should do the job just fine.

**Testing done:**
See Actions workflow

**Terms of contribution:**

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -12,9 +12,7 @@ on:
 jobs:
   golangci-lint:
     name: lint
-    runs-on:
-      group: bottlerocket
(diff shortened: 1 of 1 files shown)


jpculp (approved):

stmcginnis (approved):
Nice!

jpmcb (approved):

webern (commented):
I don't understand why we are running lints, or anything, in CI with tools that are outside the SDK. We should strive to have commands that can be run locally, with the Makefile, that are absolutely identical to those that run in CI.

PR #2680, opened 2022-12-21
pubsys: Do not allow private AMIs to be published to SSM parameters in the public namespace

cbgbt:
**Description of changes:**
This change by default will disallow publishing SSM parameters to the public `/aws/` SSM namespace if they refer to private AMIs.

**Testing done:**

* Here's a sample run where I tried to set some of my own private AMIs to the `/aws/service/doggierocket` namespace. ( :dog: )
```
03:58:05 [INFO] Using AMI data from path: ../../build/images/aarch64-aws-k8s-1.21/bottlerocket-aws-k8s-1.21-aarch64-1.10.0-867fbbda-amis.json
03:58:05 [INFO] Parsing SSM parameter templates from policies/ssm/defaults.toml
03:58:05 [INFO] Ensuring that only public images are published to public parameters.
03:58:05 [ERROR] Attempted to set parameter '/aws/service/doggierocket/aws-k8s-1.21/arm64/1.10.0/image_version' in us-west-2 to '1.10.0', based on AMI ami-0bc2a2c72b2ce0a28. That AMI is not marked public!
03:58:05 [ERROR] Attempted to set parameter '/aws/service/doggierocket/aws-k8s-1.21/arm64/1.10.0/image_id' in us-west-2 to 'ami-0bc2a2c72b2ce0a28', based on AMI ami-0bc2a2c72b2ce0a28. That AMI is not marked public!
Failed to update SSM: Cowardly refusing to publish private image to public namespace without ALLOW_PRIVATE_IMAGES

[ec2-user@ip-10-0-0-27 pubsys]$ echo $?
1
```

* Using `--allow-private-images` passes this check (but then fails because we are trying to update parameters that don't exist)

Originally I sought to use Traits and fakes to provide additional automated testing here, but it was pretty difficult to do it in a way that didn't require more significant refactoring.

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/ami/public.rs
+++ b/tools/pubsys/src/aws/ami/public.rs
@@ -1,0 +1,64 @@
+use aws_sdk_ec2::Client as Ec2Client;
+use snafu::{ensure, OptionExt, ResultExt};
+
+/// Returns whether or not the given AMI ID refers to a public AMI.
+pub(crate) async fn ami_is_public(
+    ec2_client: &Ec2Client,
+    region: &str,
+    ami_id: &str,
+) -> Result<bool> {
+    let ec2_response = ec2_client
+        .describe_images()
+        .image_ids(ami_id.to_string())
+        .send()
+        .await
+        .context(error::DescribeImagesSnafu {
+            ami_id: ami_id.to_string(),
+            region: region.to_string(),
+        })?;
+
+    let returned_images = ec2_response.images().unwrap_or_default();
+
+    ensure!(
+        returned_images.len() <= 1,
+        error::TooManyImagesSnafu {
+            ami_id: ami_id.to_string(),
+            region: region.to_string(),
+        }
+    );
+
+    Ok(returned_images
+        .first()
+        .context(error::NoSuchImageSnafu {
+            ami_id: ami_id.to_string(),
+            region: region.to_string(),
+        })?
+        .public()
+        .unwrap_or(false))
+}
+
+mod error {
+    use aws_sdk_ec2::error::DescribeImagesError;
+    use aws_sdk_ec2::types::SdkError;
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub(crate) enum Error {
+        #[snafu(display("Error describing AMI {} in {}: {}", ami_id, region, source))]
+        DescribeImages {
+            ami_id: String,
+            region: String,
+            #[snafu(source(from(SdkError<DescribeImagesError>, Box::new)))]
+            source: Box<SdkError<DescribeImagesError>>,
+        },
+
+        #[snafu(display("AMI {} not found in {}", ami_id, region))]
+        NoSuchImage { ami_id: String, region: String },
+
+        #[snafu(display("Multiples AMIs with ID {} found in  {}", ami_id, region))]
+        TooManyImages { ami_id: String, region: String },
+    }
+}
+pub(crate) use error::Error;
+type Result<T> = std::result::Result<T, error::Error>;
--- a/tools/pubsys/src/aws/ssm/mod.rs
+++ b/tools/pubsys/src/aws/ssm/mod.rs
@@ -5,18 +5,28 @@
 pub(crate) mod ssm;
 pub(crate) mod template;
 
-use crate::aws::{ami::Image, client::build_client_config, parse_arch, region_from_string};
+use self::template::RenderedParameter;
+use crate::aws::{
+    ami::public::ami_is_public, ami::Image, client::build_client_config, parse_arch,
+    region_from_string,
+};
 use crate::Args;
-use aws_sdk_ec2::model::ArchitectureValues;
+use aws_config::SdkConfig;
+use aws_sdk_ec2::{model::ArchitectureValues, Client as Ec2Client};
 use aws_sdk_ssm::{Client as SsmClient, Region};
-use log::{info, trace};
+use futures::stream::{StreamExt, TryStreamExt};
+use governor::{prelude::*, Quota, RateLimiter};
+use log::{error, info, trace};
+use nonzero_ext::nonzero;
 use pubsys_config::InfraConfig;
 use serde::Serialize;
 use snafu::{ensure, OptionExt, ResultExt};
-use std::collections::{HashMap, HashSet};
-use std::fs::File;
 use std::iter::FromIterator;
 use std::path::PathBuf;
+use std::{
+    collections::{HashMap, HashSet},
+    fs::File,
+};
 use structopt::{clap, StructOpt};
 
 /// Sets SSM parameters based on current build information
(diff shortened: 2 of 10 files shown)


tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -156,8 +210,56 @@ pub(crate) async fn run(args: &Args, ssm_args: &SsmArgs) -> Result<()> {
     Ok(())
 }
 
+// Determines the number of AMIs that can be checked for public permissions in parallel.
+const CONCURRENT_AMI_CHECKS: usize = 16;
```
stmcginnis: Would it be better to set this at runtime based on the running host's number of cores?

stmcginnis (approved):
Looks good to me once the build failure is addressed. Great to have this!

tools/pubsys/src/aws/ssm/mod.rs
cbgbt: Since this is network-bound, I picked an arbitrarily "low" number of concurrent checks. I've found that you often hit throttle limits trying to hit AWS using all of your cores if your host is too beefy. This can be rectified with more careful throttling avoidance, but I opted for a simpler route here.

I should probably clarify why the number was chosen in a comment.

stmcginnis: Interesting, I was actually worried this number was too high. I was thinking something like `Min(cpu_cores, 16)`, but even wondering if that should be more like `8` than `16`.

cbgbt: `8` is probably a safer bet!

cbgbt:
* Fix clippy warning on large size variation in Enum variants
* Lower concurrent AMI query to a safer value to avoid throttling

yeazelm (approved):
I'm still learning Rust for concurrency things, so I didn't review those specifics, but otherwise it looks good to me.

tools/pubsys/src/aws/ami/public.rs
```diff
@@ -0,0 +1,67 @@
...
+        .await
+        .context(error::DescribeImagesSnafu {
+            ami_id: ami_id.to_string(),
+            region: region.to_string(),
+        })?;
+
+    let all_returned_images = described_image.images().context(error::NoSuchImageSnafu {
+        ami_id: ami_id.to_string(),
+        region: region.to_string(),
+    })?;
```
etungsten: nit: `NoSuchImage` isn't accurate when `images()` returns `None`.

If `images` is `None`, it means the request has failed and not that there is no images in the list. If there were no images, the list would just be `Some(empty)`. It's kind of a weird paradigm with the AWS Rust SDK due to how Rust error handling works since if the request fails the error gets caught earlier with the `Result<_, SdkError>`. Theoretically this could never be `None` if the call returns `Ok`.

tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -112,21 +123,64 @@ pub(crate) async fn run(args: &Args, ssm_args: &SsmArgs) -> Result<()> {
...
     let new_parameters =
-        template::render_parameters(template_parameters, amis, ssm_prefix, &build_context)
+        template::render_parameters(template_parameters, &amis, ssm_prefix, &build_context)
             .context(error::RenderTemplatesSnafu)?;
     trace!("Generated templated parameters: {:#?}", new_parameters);
 
+    // Generate AWS Clients to use for the updates.
+    let mut param_update_ops: Vec<SsmParamUpdateOp> = Vec::new();
+    let mut aws_sdk_configs: HashMap<Region, SdkConfig> = HashMap::new();
+    let mut ssm_clients = HashMap::new();
```
etungsten: This was `    let mut ssm_clients = HashMap::with_capacity(amis.len());` before. Any harm in still pre-reserving memory for the number the AMIs we know we're gonna need to publish parameters for? Same with the other collections, do we know ahead of time the needed capacity?

tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -156,8 +210,57 @@ pub(crate) async fn run(args: &Args, ssm_args: &SsmArgs) -> Result<()> {
...
+            );
+        }
+
+        is_public
+    };
+
+    // Concurrently check our input parameter updates...
+    let results: Vec<Result<bool>> = futures::stream::iter(public_namespace_updates)
+        .then(|update| async move { Ok(check_ami_public(update)) })
+        .try_buffer_unordered(CONCURRENT_AMI_CHECKS)
```
etungsten: Since there's no wait involved, wouldn't we still potentially hit some API call rate limit and get throttled or get errors? The rate at which we reach out to EC2 API doesn't change much even if we try 8 at a time (provided that the calls return promptly)?

tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -356,4 +483,6 @@ mod error {
     }
 }
 pub(crate) use error::Error;
+
+use self::template::RenderedParameter;
```
etungsten: nit:  This should probably near the top with the other `use` declarations

tools/pubsys/src/aws/ssm/template.rs
```diff
@@ -70,13 +70,31 @@ pub(crate) fn get_parameters(
     Ok(template_parameters)
 }
 
+/// A value which stores rendered SSM parameters alongside metadata used to render their templates
+#[derive(Debug, Eq, PartialEq, Hash, Clone)]
+pub(crate) struct RenderedParameter {
+    pub(crate) ami: Image,
```
etungsten, line 76: Can we just store the AMI-ID, or do we need everything else about the Image as well?

etungsten (commented):
Nice! Just have a few non-blocking questions and small suggestions.

tools/pubsys/src/aws/ami/public.rs
etungsten: My suggestion here is to just `unwrap_or_default`

tools/pubsys/src/aws/ssm/template.rs
cbgbt: We do currently only use the AMI ID, but I added the full AMI here so that the image could be used in the future while logging to provide context to any errors. Could be useful in cases where we want to understand why a template was rendered in a given way.

I'd be okay with removing it since we don't currently do that, but it seemed like a good idea to me.

tools/pubsys/src/aws/ssm/mod.rs
cbgbt: Good point. I looked into it, and rate-limiting a stream was actually pretty easy with the `governor` crate, so I went ahead and implemented it.

cbgbt:
Addressed @etungsten's feedback
* Cleaned up `Option` handling when calling EC2 DescribeImages
* Added proper rate limiting to DescribeImages calls via `governor` crate
* Pre-allocate memory for data structures with known sizes
* linting fixes

etungsten (dismissed):

cbgbt:
* Re-added `BSD-2-Clause` to `deny.toml`. This license is allowed, it was just previously not used.

etungsten (approved):

PR #2683, opened 2022-12-22
testsys: refactor and allow overrides in shared 'migration_crd' generation

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
    testsys: refactor and allow aws region override in 'migration_crd'
    
    Migration testing for VMware variants require a region for SSM to
    dispatch upgrade/downgrade RunCommands to managed instances in vSphere.
    
    Unlike ECS or EKS, TestSys cannot assume the cluster resource CRD
    includes a region for vmware-k8s testing. Therefore we pass 'us-west-2'
    as the override region when creating the migration test CRD.
```
```
    testsys: add param to 'migration_crd' to specify instance id field name
    
    The TestSys EC2 resource CRD stores the instance IDs of created
    instances in a field named "ids", but the vSphere VM resource CRD stores
    the instance IDs of the managed instances in a field named
    "instance_ids".
```

Before these changes, migration testing on vSphere will fail because the templated fields for migration test don't exist in the VM resource CRD:
```
[2022-12-21T21:17:05Z ERROR migration_test_agent] client error: Unable to resolve config templates: An error occured while resolving the config: No field 'region' in created resource

```

**Testing done:**

The migration tests successfully kick-off and finish:
```
Every 2.0s: testsys --kubeconfig /home/etung/1.12.0-release/testsys_kubeconfig.yaml status                                                                                   Thu Dec 22 00:09:47 2022

 NAME                                                      TYPE                         STATE                         PASSED                     SKIPPED                     FAILED
 x86-64-vmware-k8s-124                                     Resource                     completed
 x86-64-vmware-k8s-124-1-initial                           Test                         pass                          1                          6972                        0
 x86-64-vmware-k8s-124-2-migrate                           Test                         pass                          2                          0                           0
 x86-64-vmware-k8s-124-3-migrated                          Test                         pass                          1                          6972                        0
 x86-64-vmware-k8s-124-4-migrate                           Test                         pass                          2                          0                           0
 x86-64-vmware-k8s-124-5-final                             Test                         pass                          1                          6972                        0
 x86-64-vmware-k8s-124-test                                Test                         running                       2                          0                           0
 x86-64-vmware-k8s-124-vms-conformance                     Resource                     completed

```

**Terms of contribution:**

### DIFF
--- a/tools/testsys/src/migration.rs
+++ b/tools/testsys/src/migration.rs
@@ -6,9 +6,12 @@ use model::Test;
 use snafu::OptionExt;
 
 /// Create a CRD for migrating Bottlerocket instances using SSM commands.
+/// `aws_region_override` allows the region that's normally derived from the cluster resource CRD to be overridden
+/// `instance_id_field_name` specifies the VM/Instance resource CRD field name for retrieving the instances IDs of the created instances
 pub(crate) fn migration_crd(
     migration_input: MigrationInput,
     aws_region_override: Option<String>,
+    instance_id_field_name: &str,
 ) -> Result<Test> {
     let cluster_resource_name = migration_input
         .cluster_crd_name
@@ -45,7 +48,7 @@ pub(crate) fn migration_crd(
     // Construct the migration CRD.
     let mut migration_config = MigrationConfig::builder();
 
-    // Use the specified aws-region for the migration test.
+    // Use the specified AWS region for the migration test.
     // If no region is specified, derive the appropriate region based on the region of the
     // cluster resource CRD (assuming it's an ECS or EKS cluster).
     if let Some(aws_region) = aws_region_override {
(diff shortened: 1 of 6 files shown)


tools/testsys/src/migration.rs
```diff
@@ -0,0 +1,95 @@
+use crate::crds::{MigrationDirection, MigrationInput};
+use crate::error::{self, Result};
+use bottlerocket_types::agent_config::MigrationConfig;
+use maplit::btreemap;
+use model::Test;
+use snafu::OptionExt;
+
+/// Create a CRD for migrating Bottlerocket instances using SSM commands.
```
ecpullen, line 8: I'd prefer to keep this in `aws_resources.rs` to signify that this migration is performed using aws tools.

ecpullen (approved):

tools/testsys/src/migration.rs
etungsten: I thought about this a little bit more and I think it makes sense to separate out the migration CRD construction from `aws_resources` since the migration CRD is for a test CRD and the test itself is not limited to just run on EKS/ECS. Sure it may use AWS tools but we also use AWS tools in VMware VMs and cluster resource as well.

webern (approved):

Issue #1703, opened 2021-08-09
containerd: set process_rlimit_no_file_soft/hard


samjo-nyang:
**What I'd like:**
I'd like to set the maximum number of open file descriptors on containers dynamically, such as set
```
[plugins."io.containerd.grpc.v1.cri"]
process_rlimit_no_file_soft = 1234
process_rlimit_no_file_hard = 12345
```
on https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/containerd/containerd-config-toml_k8s

(from https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/containerd/1002-cri-set-default-RLIMIT_NOFILE.patch)

**Any alternatives you've considered:** (noting)

samuelkarp:
Hi @samjo-nyang, thanks for opening this issue.  Can you tell us a bit more about the challenges (if any) you are experiencing with the hard-coded rlimit values?  More information would help us to prioritize this feature request.

samjo-nyang:
1. `ProcessRLimitNoFileSoft: 65536` is not enough for some microservices.
2. I have an unknown issue for akka clusters on the bottlerocket nodes. I contacted to the bottlerocket team, and I've got the following answer: `Akka uses select() so potentially it's failing because of Bottlerocket's higher default soft limit for open file descriptors. Can the customer try building a custom bottlerocket variant with "65536" here set to "1024"` / Of course, I can build the custom image, but I think that it is convenient if I have this feature.

rverma-dev:
hi @samuelkarp even we are facing same issues with our microservices. It would be nice if we can push it further to 1048576 as mentioned here https://github.com/bottlerocket-os/bottlerocket/blob/084fc3b30ea7662029976ce8205fbadd6dadd91f/packages/containerd/1001-cri-set-default-RLIMIT_NOFILE.patch#L46. But couldn't find any doc on how to put this.

zmrow:
Thanks!  We'll take look at if / how we can make this more configurable.

zmrow:
There's an [open Kubernetes issue](https://github.com/kubernetes/kubernetes/issues/3595) requesting the ability to make these limits configurable.  At first blush, making the limits configurable via Kubernetes (per pod or something) may be a bit more desirable than setting the limits for the entire host.

nairb774:
[...]

PR #2687, opened 2022-12-29
datastore: allow simple enums as map keys

### LINKED ISSUE #2686: datastore: allow simple enums as map keys
Change the datastore so that an enum whose variants have no data, and which can serialized and deserialized as a scalar string can be used as a map key.

Background:

In #2597 I provided a solution so that "simple" enums (those with no data-bearing variants) can be treated like string values in the datastore. (See also #1501 and #2559)

In the implementation to #1703, we want to take this further and use a "simple" enum as a map key value.

webern:
**Issue number:**

Closes #2686

**Description of changes:**

This allows us to use an enum as a map key in the model, which is something we want to do in the upcoming implementation for #1703

**Testing done:**

Created this enum

```rust
#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize, Scalar)]
#[serde(rename_all = "snake_case")]
pub enum RandomWords {
    Gloom,
    TheftClash,
}
```

Added it to the EcsSettings struct

```rust
// ECS settings.
#[model]
struct ECSSettings {
    /* ... */
    stuff: HashMap<RandomWords, String>,
}
```

Ran an `aws-ecs-1` variant:

Tried to set a bad map key, api rejected it as expected:

```
apiclient set ecs.stuff.bad_key=value
Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: Error during deserialization: unknown variant `bad_key`, expected one of `begin`, `bet`, `crack`, `elect`, `fix`, `gloom`, `item`, `lump`, `new`, `news`, `strap`, `theft_clash`, `time`, `tip`
```

Set a couple of good key/vals:

```
apiclient set ecs.stuff.theft_clash=value_1
apiclient set ecs.stuff.gloom=value_2
```

Get the map:

```
apiclient get settings.ecs.stuff
{
  "settings": {
    "ecs": {
      "stuff": {
        "gloom": "value_2",
        "theft_clash": "value_1"
      }
    }
  }
}
```

**Terms of contribution:**

### DIFF
--- a/sources/api/datastore/src/serialization/mod.rs
+++ b/sources/api/datastore/src/serialization/mod.rs
@@ -121,13 +121,14 @@ impl ser::Serializer for &MapKeySerializer {
         bad_key("unit_struct")
     }
 
+    /// A simple enum can be used as if it were a string, so we allow these to serve as map keys.
     fn serialize_unit_variant(
         self,
         _name: &'static str,
         _variant_index: u32,
-        _variant: &'static str,
+        variant: &'static str,
     ) -> Result<String> {
-        bad_key("unit_variant")
+        self.serialize_str(variant)
     }
 
     fn serialize_newtype_struct<T: ?Sized>(self, _name: &'static str, _value: &T) -> Result<String>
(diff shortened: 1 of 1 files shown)


yeazelm (approved):
LGTM

stmcginnis (approved):
Looks good!

PR #2706, opened 2023-01-05
build: clean up kernel command line

bcressey:
**Issue number:**
N/A

**Description of changes:**
This removes two `dm-verity` options that aren't actually supported - `dm_verity.max_bios=-1` and `dm_verity.dev_wait=1` - which came from the original, out-of-tree Chrome OS implementation.

The more significant change is to put the `dm-mod.create` options into a GRUB variable. This removes some extraneous whitespace, and makes it easier to implement the planned "resign" workflow, which will need to replace these options in `grub.cfg` after modifying the root image and recalculating the dm-verity parameters.

**Testing done:**

Before:
```
bash-5.1# cat /proc/cmdline
BOOT_IMAGE=(hd0,gpt3)/vmlinuz console=tty0 console=ttyS0,115200n8 crashkernel=2G-:256M net.ifnames=0 netdog.default-interface=eth0:dhcp4,dhcp6? quiet bootconfig root=/dev/dm-0 rootwait ro raid=noautodetect random.trust_cpu=on selinux=1 enforcing=1 dm_verity.max_bios=-1 dm_verity.dev_wait=1 "dm-mod.create=root,,,ro,0 1359912 verity 1 PARTUUID=cfd4dcbd-27df-416e-b3af-38c3036990d4/PARTNROFF=1 PARTUUID=cfd4dcbd-27df-416e-b3af-38c3036990d4/PARTNROFF=2        4096 4096 169989 1 sha256 74941a0af03bfc742d8cb12f72ddecded61551a75b02265ba73991827ee71236 1accea696d3d89909e3e3dff7d49074ff55df591c26fc209cc8fecbeb0ddf662        2 restart_on_corruption ignore_zero_blocks" -- systemd.log_target=journal-or-kmsg systemd.log_color=0 systemd.show_status=true
```

After:
```
bash-5.1# cat /proc/cmdline
BOOT_IMAGE=(hd0,gpt3)/vmlinuz console=tty0 console=ttyS0,115200n8 crashkernel=2G-:256M net.ifnames=0 netdog.default-interface=eth0:dhcp4,dhcp6? quiet bootconfig root=/dev/dm-0 rootwait ro raid=noautodetect random.trust_cpu=on selinux=1 enforcing "dm-mod.create=root,,,ro,0 1359912 verity 1 PARTUUID=aa25f9a0-f230-4fc4-8963-05a23cb28154/PARTNROFF=1 PARTUUID=aa25f9a0-f230-4fc4-8963-05a23cb28154/PARTNROFF=2 4096 4096 169989 1 sha256 4b9468038dcf4ced397f82dfd7100870fe02c447311833bb34951d15a5327b45 4c9c6892c8711f45ef13a283a81446f8c661642bcd437dbb1c80e6b145bbf160 2 restart_on_corruption ignore_zero_blocks" -- systemd.log_target=journal-or-kmsg systemd.log_color=0 systemd.show_status=true
```

Images still boot and the root filesystem is set up as expected.

**Terms of contribution:**

### DIFF
--- a/tools/rpm2img
+++ b/tools/rpm2img
@@ -307,11 +307,31 @@ VERITY_SALT="$(grep '^Salt:' <<<"${veritysetup_output}" | awk '{ print $NF }')"
 veritysetup verify "${ROOT_IMAGE}" "${VERITY_IMAGE}" "${VERITY_ROOT_HASH}"
 dd if="${VERITY_IMAGE}" of="${OS_IMAGE}" conv=notrunc bs=1M seek="${partoff[HASH-A]}"
 
+declare -a DM_VERITY_ROOT
+DM_VERITY_ROOT=(
+  "root,,,ro,0"
+  "${VERITY_DATA_512B_BLOCKS}"
+  "verity"
+  "${VERITY_VERSION}"
+  "PARTUUID=\$boot_uuid/PARTNROFF=1"
+  "PARTUUID=\$boot_uuid/PARTNROFF=2"
+  "${VERITY_DATA_BLOCK_SIZE}"
+  "${VERITY_HASH_BLOCK_SIZE}"
+  "${VERITY_DATA_4K_BLOCKS}"
+  "1"
+  "${VERITY_HASH_ALGORITHM}"
+  "${VERITY_ROOT_HASH}"
+  "${VERITY_SALT}"
+  "2"
+  "restart_on_corruption"
+  "ignore_zero_blocks"
+)
+
 # write GRUB config
 # If GRUB_SET_PRIVATE_VAR is set, include the parameters that support Boot Config
 if [ "${GRUB_SET_PRIVATE_VAR}" == "yes" ] ; then
    BOOTCONFIG='bootconfig'
-   INITRD='initrd ($private)/bootconfig.data'
+   INITRD="initrd (\$private)/bootconfig.data"
 else
    BOOTCONFIG=""
    INITRD=""
@@ -320,6 +340,7 @@ fi
 cat <<EOF > "${BOOT_MOUNT}/grub/grub.cfg"
 set default="0"
 set timeout="0"
+set dm_verity_root="${DM_VERITY_ROOT[@]}"
 
 menuentry "${PRETTY_NAME} ${VERSION_ID}" {
    linux (\$root)/vmlinuz \\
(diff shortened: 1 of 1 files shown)


tools/rpm2img
```diff
@@ -320,20 +340,21 @@ fi
 cat <<EOF > "${BOOT_MOUNT}/grub/grub.cfg"
 set default="0"
 set timeout="0"
+set dm_verity_root="${DM_VERITY_ROOT[@]}"
```
bcressey, line 343: Any time a new command is used in the GRUB config, it's worth extra scrutiny since GRUB doesn't get updated by in-place updates. It'd be easy to rely on new functionality that would create problems on upgrade for older images.

Here I'm confident it's safe because we've used `set` commands forever, and never added a new GRUB module for that functionality.

tools/rpm2img
```diff
@@ -320,20 +340,21 @@ fi
...
 
 menuentry "${PRETTY_NAME} ${VERSION_ID}" {
    linux (\$root)/vmlinuz \\
        ${KERNEL_PARAMETERS} \\
        ${BOOTCONFIG} \\
        root=/dev/dm-0 rootwait ro \\
        raid=noautodetect \\
-       random.trust_cpu=on selinux=1 enforcing=1 \\
-       dm_verity.max_bios=-1 dm_verity.dev_wait=1 \\
```
yeazelm, line 331: FWIW they *just* added [dm-mod.waitfor](https://www.kernel.org/doc/html/latest/admin-guide/device-mapper/dm-init.html) into the kernel so at some point in the future we may want to add this back in for the root partition once we are on 6.2+. This is especially useful for slower block devices that use probed asynchronously.

yeazelm (approved):
I like the readability improvements!

tools/rpm2img
```diff
@@ -320,20 +340,21 @@ fi
...
-       dm-mod.create="root,,,ro,0 $VERITY_DATA_512B_BLOCKS verity $VERITY_VERSION PARTUUID=\$boot_uuid/PARTNROFF=1 PARTUUID=\$boot_uuid/PARTNROFF=2 \\
-       $VERITY_DATA_BLOCK_SIZE $VERITY_HASH_BLOCK_SIZE $VERITY_DATA_4K_BLOCKS 1 $VERITY_HASH_ALGORITHM $VERITY_ROOT_HASH $VERITY_SALT \\
-       2 restart_on_corruption ignore_zero_blocks" \\
+       random.trust_cpu=on \\
+       selinux=1 enforcing \\
```
markusboehme: From my reading of the code, `enforcing` without any value won't actually be recognized as a valid option. On closer look, the current kernels don't even have the respective config options for these two command line options enabled (`CONFIG_SECURITY_SELINUX_BOOTPARAM` for `selinux` and `CONFIG_SECURITY_SELINUX_DEVELOP` for `enforcing`). This is fine in so far as both of them fall back to secure (enabled, enforcing) defaults.

It would need some digging to find out when the config options were dropped. Should they be brought back? Being unable to disable SELinux via boot-time options might not fundamentally change the security posture, but adds an extra hurdle.

zmrow (approved):
LGTM! Nice work

tools/rpm2img
bcressey: The kernels are configured without these options by design - SELinux isn't ever meant to be disabled on Bottlerocket.

I've retained `selinux=1 enforcing` on the off-chance that there's some application out there that tries to detect SELinux state by checking `/proc/cmdline`. That's pretty broken, but the correct approach - linking to libselinux - might not be appealing for various reasons. It serves as a bit of a hint - "here be SELinux". I'm not sure that hint is good for much other than reporting somewhere - "yup, this node probably has SELinux turned on".

So I've occasionally been tempted to drop them. It's a bit ugly that they get passed down to `init` as parameters, since the kernel doesn't recognize them. The safest way to do that would be to shift them into kernel parameters in all the existing variant definitions, and then omit them in newer variants going forward.

webern (approved):

tools/rpm2img
markusboehme: Thanks for clarifying! I figured something like this could also have been the reason. Should you add the `=1` back onto `enforcing`, as right now it's probably not serving any purpose and wouldn't even be recognized if the kernel config option was enabled? Approving in any case, as the original doubts have been settled.

markusboehme (approved):

tools/rpm2img
bcressey: > Should you add the =1 back onto enforcing, as right now it's probably not serving any purpose and wouldn't even be recognized if the kernel config option was enabled?

Oops - yes!

stmcginnis (approved):
Looks fine to me.

Issue #2449, opened 2022-09-23
Switch to `systemd-networkd` as a replacement for `wicked`


zmrow:
**What I'd like:**
I'd like to move to `systemd-networkd` to streamline dependencies and provide a more familiar networking stack for users.

Bottlerocket [switched to `wicked`](https://github.com/bottlerocket-os/bottlerocket/pull/150) in the early stages of the project.  A few things have changed since then and some aspects of that PR aren't applicable anymore.

**Any alternatives you've considered:**
Continue using wicked and keep the status quo

zmrow:
Our main goal for this change is to make the switch as invisible as possible to the user.  Ideally, no one should even notice the switch happened.  This means there can be no functionality loss.  It also means we will continue to support all current network config versions.

While the change to `systemd-networkd` opens the door for more networking-related features, we won't be adding any new features as part of this switch.  We will make the switch to `systemd-networkd` first, and add features later.

Tasks list (subject to update and change):

- [x] Add `systemd-networkd` to the tree.
  - [x] Add `networkd` as a subpackage of `systemd`
  - [x] Add a build flag to enable the use of `systemd-networkd` in any variant.
  - [x] Evaluate `networkd`-related packages from the above PR to determine if they are still necessary
- [x] Implement `networkd` config file generation in netdog 
- [x] Implement the "optional lease" behavior `wicked` provides.  (interface config for `systemd-networkd-wait-online.service`?) 
~~- [ ] Implement a helper to run on interface initialization and DHCP lease renewal (DBUS listener?)  This rewrites `current_ip` and `resolv.conf` (WIP: @yeazelm )~~
~~- [ ] Handle DNS settings - read from networkd lease (WIP: @yeazelm)~~
- [x] Integrate systemd-resolved
- [x] Handle sysctls for primary interface (if necessary)
- [x] Evaluate any new SELinux rules that might be necessary
[...]

PR #2709, opened 2023-01-06
variants: move wicked out of release

yeazelm:
**Issue number:** [ 2449](https://github.com/bottlerocket-os/bottlerocket/issues/2449)

Closes #

**Description of changes:** 
We need to move wicked out of release so that we can toggle what network backend a variant uses while working on networkd enablement. This makes it possible to remove wicked from one variant without affecting others or have a variant just for networkd testing. This commit also makes the variant Cargo.toml ordering consistent with all variants for clarity and consistency.

The diff is lousy because of the shifting of the files. The only changes should be the addition of wicked to packages included and wicked for dependencies. The rest is commenting and sorting to get them all consistent. 

**Testing done:**
Built all the variants. This should not actually affect the resulting images at all.

**Terms of contribution:**

### DIFF
--- a/packages/release/Cargo.toml
+++ b/packages/release/Cargo.toml
@@ -44,4 +44,3 @@ procps = { path = "../procps" }
 selinux-policy = { path = "../selinux-policy" }
 systemd = { path = "../systemd" }
 util-linux = { path = "../util-linux" }
-wicked = { path = "../wicked" }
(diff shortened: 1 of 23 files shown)


yeazelm:
The builds caught some I missed. I'll update and also break out the reformatting out from the wicked change.

PR #2714, opened 2023-01-07
expose efi variables to privileged host containers

bcressey:
**Issue number:**

#2501 

**Description of changes:**
Build `systemd` with the "efi" option set, so that `systemd` will mount `efivarfs` on boot under `/sys/firmware/efi/efivars`.

For privileged host containers, mount `/sys/firmware` with the "rbind" option so that child mounts such as `efivarfs` are also propagated.

**Testing done:**
Confirmed that `/sys/firmware/efi/efivars` shows up in the admin container, and that it can be read from and written to by commands like `mokutil`.

```
[bottlerocket@admin]$ mokutil --sb-state
SecureBoot disabled
```

**Terms of contribution:**


zmrow (approved):
:safety_vest:

stmcginnis (approved):

jpculp (approved):

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -781,6 +781,12 @@ func withPrivilegedMounts() oci.SpecOpts {
 			Source:      "/usr/src/kernels",
 			Type:        "bind",
 		},
+		{
```
arnaldo2792, line 784: Will it make sense to make this mount read-only?

bcressey: Per [efivarfs docs](https://docs.kernel.org/filesystems/efivarfs.html):
> Due to the presence of numerous firmware bugs where removing non-standard UEFI variables causes the system firmware to fail to POST, efivarfs files that are not well-known standardized variables are created as immutable files. This doesn’t prevent removal - “chattr -i” will work - but it does prevent this kind of failure from being accomplished accidentally.

For the most part they're marked as immutable already, and anyone running in the admin container will also be able to remount the filesystem read-write.

`mokutil` uses some of these variables to pass update requests to `shim`, which then verifies physical presence (or at least console access) as well as knowledge of a password before actually updating the variable. So there's at least one use case for having a read-write filesystem, even though I don't expect it to be used very often.

arnaldo2792 (approved):

PR #2716, opened 2023-01-09
sundog: generate dynamic settings in parallel

cbgbt:
**Issue number:**
n/a

**Description of changes:**
I was looking at `systemd-analyze plot` for bottlerocket and noted that we are quite bottlenecked by `sundog.service` and occassionally `settings-applier.service`.

Taking a closer peek at those, we sequentially perform settings generation and service re-starting in both cases, which are IO-bound. We don't otherwise enforce ordering on these tasks, so they should be embarrassingly parallel.

This change focuses only on `sundog` for now.

```
sundog: generate dynamic settings in parallel

This also updates nix to the latest version, which is required to use
the latest rayon.
```

**Testing done:**
Initial testing suggests that `sundog` completes on the order of a few hundred milliseconds, instead of a few thousand:

**Before:**
before

**After:**
after

* Confirmed that generated settings are present after making the change
* Confirmed that `thar-be-settings` is still restarting all services. Despite not making  changes to this code, it seems like the systemd service is quite variable in its timing.

**Terms of contribution:**

### DIFF
--- a/sources/api/sundog/src/main.rs
+++ b/sources/api/sundog/src/main.rs
@@ -12,6 +12,7 @@ The output is collected and sent to a known Bottlerocket API server endpoint.
 #[macro_use]
 extern crate log;
 
+use rayon::prelude::*;
 use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::{HashMap, HashSet};
@@ -296,8 +297,6 @@ async fn get_dynamic_settings<P>(
 where
     P: AsRef<Path>,
 {
-    let mut settings = HashMap::new();
-
     // Build the list of settings to query from the datastore to see if they
     // are currently populated.
     // `generators` keys are setting names in the proper dotted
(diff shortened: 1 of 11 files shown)


sources/api/sundog/src/main.rs
```diff
@@ -309,123 +308,133 @@ where
...
-        //  implementations I can find on crates.io. We can roll our own at some point if this
-        //  becomes a bottleneck
-        if populated_settings
-            .iter()
-            .any(|k| k.starts_with_segments(setting.segments()))
-        {
-            debug!("Setting '{}' is already populated, skipping", setting);
-            continue;
-        }
+    // for (setting_str, generator) in generators {
```
arnaldo2792: nit: extra comment should be removed

arnaldo2792 (approved):

cbgbt:
Remove extraneous comment.

zmrow (approved):
Nice! :mushroom:

stmcginnis (approved):
Nice!

etungsten (approved):

webern (commented):
I guess you feel that there is no hidden magic about the order in which the settings are currently generated? (I don't know this program.)

I think there is an opportunity here to create about 4-6 smaller functions that are called in sequence. This would have the benefit of the code remaining at the same indentation level and might improve the diff. But more importantly it would be easier to read the "big picture" and understand what is being done in parallel and what the iters are etc.

cbgbt:
I wasn't pleased with how much variability I saw in the data that I gathered while checking to see if this improved things, so I set out to gather more data by increasing my sample size.

So far as I can tell, there's not strong evidence that this improves boot performance on average (and may, in fact, make it worse). After digging more into why `sundog` takes so much time on boot, it's largely due to latent requests to IMDS taking place in shibaken. Possibly this can be improved by allowing for tighter retried requests, but as is I'm not necessarily thrilled about merging this change anymore.

Going to convert this back to a draft while I gather more data.

PR #2699, opened 2023-01-05
Add k8s 1.25 variants

stmcginnis:
**Issue number:**

Closes #2698

**Description of changes:**

This adds a k8s-1.25 package.
Existing 1.24 models are moved to be 1.25, with symlinks created for 1.24 models to point to the new locations.
Finally, new variants are defined for aws-k8s-1.25, aws-k8s-1.25-nvidia, metal-k8s-1.25, and vmware-k8s-1.25.

**Testing done:**

To be done:

- [x] Update package with final EKS-D 1.25 package source
- [x] Test aws-k8s-1.25 x86_64 joining 1.25 cluster and ran full conformance test
```
12:04:16       e2e   global   complete   passed   Passed:362, Failed:  0                                                                 
12:04:16 Sonobuoy has completed. Use `sonobuoy retrieve` to get results. 
```
- [x] Test aws-k8s-1.25 ARM joining 1.25 cluster and deploying pods
- [x] Test aws-k8s-1.25-nvidia x86-64 joining 1.25 cluster and ran full conformance test
```
      "progress": {
        "name": "e2e",
        "node": "global",
        "timestamp": "2023-01-13T21:45:38.815114657Z",
        "msg": "Test Suite completed",
        "total": 0,
        "completed": 362
      }
```
- [x] Test aws-k8s-1.25-nvidia ARM joining 1.25 cluster and ran full conformance test
- [x] Test metal-k8s-1.25 joining 1.24 cluster and ran full conformance test
```
Plugin: e2e
Status: passed
Total: 6973
Passed: 356
Failed: 0
Skipped: 6617
```
- [x] Test vmware-k8s-1.25 joining 1.24 cluster and ran full conformance test
```
      "progress": {
        "name": "e2e",
        "node": "global",
        "timestamp": "2023-01-13T23:27:27.38052641Z",
        "msg": "Test Suite completed",
        "total": 356,
        "completed": 356
      }
```
- [x] Check kubelet logs and system logs in general, compared them with the aws-k8s-1.24 variant kubelet logs.

<details>

Note that there are a few flags that are being deprecated. We are planning migrating to `--cloud-provider external` for 1.26 and beyond.

```bash
bash-5.1# journalctl --no-hostname -u  kubelet
Jan 13 18:23:49 systemd[1]: Starting Kubelet...
[...]

PR #2721, opened 2023-01-10
Correct mount file name for k8s PKI private files

### LINKED ISSUE #2720: /etc/kubernetes/pki/private not mounted
**Image I'm using:**
aws-k8s-1.22

`/etc/kubernetes/pki/private` should be mounted with an SELinux context option to override MAC permissions.

**What actually happened:**
`dmesg` reports a failure:
```
[    1.716988] systemd[1]: etc-kubernetes-pki.mount: Where= setting doesn't match unit name. Refusing.
[    1.724101] systemd[1]: etc-kubernetes-pki.mount: Cannot add dependency job, ignoring: Unit etc-kubernetes-pki.mount has a bad unit file setting.
```

Launch an instance with the `aws-k8s-1.22` image.

stmcginnis:
**Issue number:**

Closes #2720 

**Description of changes:**

The tempfs mount for the PKI directory used by Kubernetes was moved from `/etc/kubernetes/pki` to /etc/kubernetes/pki/private` to separate out the public and private certificate files. When this was done, the mount unit file should have also been renamed to match this new mount path. This was overlooked, causing errors during boot.

This change renames the mount file to match the actual path.

**Testing done:**

Built and deployed a 1.22 cluster. Logged in to joined node and checked for any errors:

```bash
bash-5.1# dmesg | grep pki
bash-5.1# dmesg | grep mount
[    0.473685] EXT4-fs (dm-0): mounted filesystem without journal. Opts: (null)
[    0.498266] devtmpfs: mounted
[    3.453494] systemd[1]: Starting Remount Root and Kernel File Systems...
[    4.504398] EXT4-fs (nvme1n1p1): mounted filesystem with ordered data mode. Opts: (null)
[    4.564735] EXT4-fs (nvme0n1p12): mounted filesystem with ordered data mode. Opts:
[    5.499234] EXT4-fs (nvme0n1p3): mounted filesystem without journal. Opts: (null)
bash-5.1# mount  | grep pki
tmpfs on /etc/kubernetes/pki/private type tmpfs (rw,nosuid,nodev,noexec,noatime,context=system_u:object_r:secret_t:s0,mode=700)
```

**Terms of contribution:**


bcressey (approved):
Thanks! I see you also fixed this in #2699 - nice!

arnaldo2792 (approved):

PR #2723, opened 2023-01-10
testsys: Use `Test.toml` for controller install

### LINKED ISSUE #2571: testsys: Use `Test.toml` for controller image uri in `testsys install`
**What I'd like:**

Currently `testsys` has a default image location for the controller, but won't use the registry provided in `Test.toml`. Instead, `Test.toml` should be used in `testsys install`.

ecpullen:
**Issue number:**

Closes #2571 

**Description of changes:**

This pr allows setting the testsys controller uri with `Test.toml`.

**Testing done:**

Tested `cargo make setup-test` with a custom uri in `Test.toml` and the default uri.

**Terms of contribution:**

### DIFF
--- a/tools/testsys/src/install.rs
+++ b/tools/testsys/src/install.rs
@@ -1,40 +1,57 @@
 use crate::error::Result;
+use crate::run::TestsysImages;
 use clap::Parser;
 use log::{info, trace};
 use model::test_manager::{ImageConfig, TestManager};
+use std::path::PathBuf;
+use testsys_config::TestConfig;
 
 /// The install subcommand is responsible for putting all of the necessary components for testsys in
 /// a k8s cluster.
 #[derive(Debug, Parser)]
 pub(crate) struct Install {
-    /// Controller image pull secret. This is the name of a Kubernetes secret that will be used to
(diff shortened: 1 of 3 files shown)


tools/testsys/src/install.rs
```diff
@@ -1,40 +1,57 @@
...
+            }),
+            Some(testsys_config::TestsysImages::public_images()),
+        ]
+        .into_iter()
+        .flatten()
+        .fold(Default::default(), testsys_config::TestsysImages::merge);
+
+        let controller_uri = images
+            .controller_image
+            .expect("The default controller image is missing.");
```
stmcginnis, line 42: Is "default" the right phrase here? If we're reading it from the config file, and we've removed the CLI arguments to provide it, I wouldn't think of anything as a default. Maybe better to include the expected controller image value in the output?

ecpullen: `default` is the correct word in this case, the default image location is coded in `testsys-config`, so if this panics, it means the default value wasn't set, but that's impossible (reason for `expect` and not an error).

stmcginnis (approved):

webern (approved):

PR #2726, opened 2023-01-10
Revert "[wicked] Include MAC address in the lease files"

zmrow:
**Issue number:**
N/A

**Description of changes:**
This removes the patch to `wicked` that adds the MAC address to the lease files.  We aren't going to make use of this patch, preferring the use of sysfs instead.

**Testing done:**
Build and boot an `aws-k8s-1.22` and verify the MAC is no longer in the lease file.

**Terms of contribution:**


bcressey (approved):
It could be useful to say why we're doing the revert in the commit message, just for posterity's sake.

yeazelm (approved):

webern (approved):

zmrow:
> It could be useful to say why we're doing the revert in the commit message, just for posterity's sake.

I updated the commit message :)

bcressey:
> I updated the commit message :)

the line lengths are a bit off but the content looks fine

markusboehme (approved):

zmrow:
> > I updated the commit message :)
> 
> the line lengths are a bit off but the content looks fine

@bcressey  I think GitHub shortens the commit to the short SHA which makes the line length look weird.  `git log` shows the commit like so:
```
commit bd4a6cd4ef0d0c0af894f961c4b854ca2487b59f (HEAD -> revert-wicked-mac-patch, origin/revert-wicked-mac-patch)
Date:   Tue Jan 10 16:03:30 2023 +0000

    Revert "[wicked] Include MAC address in the lease files"
    
    This reverts commit 71675f00f4b63969b0d3aee440fd7c06832c0638 since we
    are not using the MAC address fields from the lease, preferring the use
    of sysfs.
```

bcressey:
Ah, carry on then!

PR #2730, opened 2023-01-11
update to SDK v0.29.0

bcressey:
**Issue number:**
N/A

**Description of changes:**
Update SDK to v0.29.0. Includes changes from https://github.com/bottlerocket-os/bottlerocket-sdk/pull/87.

The major change in this update is to GCC 11, which initially miscompiled `dbus-broker` when built with `-fstrict-aliasing` (the default at the `-O2` optimization level). I added a patch to address that, and also elevated aliasing warnings into build errors to try to catch similar errors in the future. This didn't actually flag the `dbus-broker` problem, but still feels like a useful change.

There was also an update to `pahole` via the `dwarves` package in Fedora 37 which caused problems with the older kernel 5.10 and 5.15 builds; that's been fixed by #2569.

**Testing done:**
Built variants for both target architectures on both host architectures. Verified aws, vmware, and metal builds were reasonably clean - no alarming warnings across all the build output in `variants/target`, and no strange behavior at runtime.

**Terms of contribution:**


bcressey:
Putting this out in draft until I can publish SDK images for the more recent changes in https://github.com/bottlerocket-os/bottlerocket-sdk/pull/87, and do some additional testing.

jpmcb (approved):

jpculp (approved):

PR #2732, opened 2023-01-11
early-boot-config: downgrade warning about inability to iopl to debug

### LINKED ISSUE #2727: early-boot-config fails with iopl error on vmware-dev variant
**Image I'm using:** vmware-dev

**What I expected to happen:** image boots, doesn't emit scary warnings

**What actually happened:** early-boot-config logs the following warning (repeatedly):

```
Unable to access guestinfo via privileged mode, using unprivileged: vmware backdoor error: iopl failed: Function not implemented (errno: 38)
```

**How to reproduce the problem:** Try to launch the current vmware-dev variant after the merge of #2569

markusboehme:
**Issue number:** #2727

Closes #2727

**Description of changes:**

The VMware provider of early-boot-config communicates with the hypervisor via the VMware backdoor, a couple ports in the x86 port IO address space. By default Linux does not allow IO port access to user space processes, and so the used crate optionally invokes the `iopl` system call to change the IO privilege level of the process. The crate calls this "privileged" access.

The kernel update in 22a6a58db6c8 ("kernel-5.15: update to 5.15.79") changed the configuration to disable the `iopl` system call. This causes the VMware provider to emit a warning about the inability to use the VMware backdoor via "privileged" access. The provider falls back to "unprivileged" access, i.e. proceeds to use the backdoor without changing the IO privilege level first.

This fallback works just as well since both VMware and KVM (in its VMware guest data emulation) special-case the two IO ports used for the VMware backdoor. A process trying to access these will only enter into the hypervisor's fault handler which proceeds to serve the requests; the hypervisor does not inject a fault into the guest. Hence, the prior call to `iopl` is redundant, and "unprivileged" access works just as well.

Only downgrade the warning about the inability to invoke `iopl` to debug level for now. After some more experience with the behavior on variants with the 5.15 kernel we can then eliminate the "privileged" access for the variants with the 5.10 kernel as well.

**Testing done:**

Booted a vmware-dev build on VMware with and without this change. Without this change, I see repeated messages of the form:

```
[WARN] Unable to access guestinfo via privileged mode, using unprivileged: vmware backdoor error: iopl failed: Function not implemented (errno: 38)
```

They disappear with the change.

**Terms of contribution:**

### DIFF
--- a/sources/api/early-boot-config/src/provider/vmware.rs
+++ b/sources/api/early-boot-config/src/provider/vmware.rs
@@ -213,13 +213,14 @@ impl VmwareDataProvider {
     /// Request a key's value from guestinfo
     fn backdoor_get_bytes(key: &str) -> Result<Option<Vec<u8>>> {
         // Probe and access the VMware backdoor.  `kernel lockdown(7)` may block "privileged"
-        // mode because of its use of `iopl()`.  According the the bug report below, in theory
-        // "privileged" mode is more reliable than "unprivileged" but both provide access to
-        // the same data so we fall back to "unprivileged" if the first call fails.
-        // https://github.com/lucab/vmw_backdoor-rs/issues/6
+        // mode because of its use of `iopl()`; the 5.15 kernels have it disabled regardless
+        // of lockdown mode. If this fails, fall back to "unprivileged" access without first
+        // requesting access to the relevant IO ports. KVM and VMware both have them special-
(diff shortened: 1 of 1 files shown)


bcressey (approved):

zmrow (approved):
:vulcan_salute:

arnaldo2792 (approved):

PR #2737, opened 2023-01-17
testsys: Add support for a `tests` directory

ecpullen:
**Issue number:**

Closes #

**Description of changes:**

```bash
dea965a8 testsys: Add userdata files to `tests`
b4228725 testsys: Enable test-type dependencies
2fd6b870 testsys: Allow configs in `test` table
874bd083 testsys: Eksctl config file support
1bb95e14 testsys: Support custom tests from `tests` dir
74295e5a testsys: kubeconfig and test config in tests dir
08411e14 testsys: Add `tests` to .gitignore
```

`dea965a8 testsys: Add userdata files to 'tests'`
Allow users to add files to the testsys directory either in `tests/test_type`, `tests/shared`, `tests/shared/userdata`

`b4228725 testsys: Enable test-type dependencies`
Allows a test type to provide configuration beyond an existing test type.

```toml
# Test.toml

[tests.configuration.foo]
# Tell TestSys that this is just another configuration of quick testing and the quick testing crds should be used.
test-type = "quick"
# Use `foo.toml` for userdata for this test type.
userdata = "foo"
```

`2fd6b870 testsys: Allow configs in `test` table`
Before, configurations had to be in a variant based table. Now configurations can also be present in the top level table.

`874bd083 testsys: Eksctl config file support`
Add support for eksctl config files.

```yaml
# `tests/shared/cluster-config/basic-cluster.yaml`

apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig
metadata:
  name: "{{kube-arch}}-{{kube-variant}}-test"
  version: "{{version}}"
  region: "{{region}}"
kubernetesNetworkConfig:
  ipFamily: IPv4
vpc:
  clusterEndpoints:
    publicAccess: true
    privateAccess: true
iam:
  withOIDC: true
addons:
  - name: vpc-cni
    version: latest
  - name: coredns
    version: latest
  - name: kube-proxy
    version: latest
managedNodeGroups:
  - name: br-mng-{{kube-arch}}-{{kube-variant}}
    availabilityZones: []
    minSize: 0
    desiredCapacity: 0
    maxSize: 1
```

```toml
# Test.toml

[aws-k8s]
# TestSys will see that `basic-cluster.yaml` exists in `tests/shared/cluster-config` and use that in the crd. Since `existing-cluster.yaml` doesn't exist, it will handle `existing-cluster` without a cluster config.
cluster-names = ["basic-cluster", "existing-cluster"]
```

`1bb95e14 testsys: Support custom tests from `tests` dir`
A custom tests yaml configuration can be added to `tests/<TEST_TYPE>/test.yaml` (or a few other options)

`74295e5a testsys: kubeconfig and test config in tests dir`

Add support to `Makefile.toml` for `Test.toml` and `testsys.kubeconfig` in the `tests` directory _or_ in the top level directory.

`08411e14 testsys: Add `tests` to .gitignore`

Add the `tests` directory to the projects `.gitignore`.

**Testing done:**

Tested various configurations of userdata, cluster config and custom test locations and verified that the correct file was used. Tested various nested configurations in `Test.toml` and verified that they were resolved properly.

**Terms of contribution:**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 /Licenses.toml
 /licenses
 *.run
+/tests
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -123,6 +123,7 @@ CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH = "${BUILDSYS_ROOT_DIR}/testsys.kubec
 TESTSYS_STARTING_VERSION = { script = ["git tag --list --sort=version:refname 'v*' | tail -1"] }
 # The commit for the last release of bottlerocket.
 TESTSYS_STARTING_COMMIT = { script = ["git describe --tag ${TESTSYS_STARTING_VERSION} --always --exclude '*' || echo 00000000"] }
+TESTSYS_TESTS_DIR = "${BUILDSYS_ROOT_DIR}/tests"
 TESTSYS_TEST_CONFIG_PATH = "${BUILDSYS_ROOT_DIR}/Test.toml"
 
 [env.development]
@@ -183,10 +184,17 @@ BOOT_CONFIG = "${BUILDSYS_ROOT_DIR}/bootconfig.data"
 # default kubeconfig location does not exist, use the users default kubeconfig.
 CARGO_MAKE_TESTSYS_KUBECONFIG_ARG = {script = [
 '''
+if ! [ -n "${TESTSYS_KUBECONFIG}" ] && [ -s "${TESTSYS_TESTS_DIR}/testsys.kubeconfig" ] && [ -s "${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}" ];then
+   echo "No kubeconfig was specified and a kubeconfig was found in 2 possible locations: '${TESTSYS_TESTS_DIR}/testsys.kubeconfig' and '${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}'"
+   exit 1
+fi
 if [ -n "${TESTSYS_KUBECONFIG}" ]; then
    # If the user provides a kubeconfig path it should be used.
    echo "--kubeconfig ${TESTSYS_KUBECONFIG}"
-elif [ -f "${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}" ]; then
+elif [ -s "${TESTSYS_TESTS_DIR}/testsys.kubeconfig" ]; then
+   # If the kubeconfig is in the TESTSYS_TESTS_DIR it should be used.
+   echo "--kubeconfig ${TESTSYS_TESTS_DIR}/testsys.kubeconfig"
+elif [ -s "${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}" ]; then
    # If the default kubeconfig exists it should be used.
    echo "--kubeconfig ${CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH}"
 fi
@@ -195,6 +203,24 @@ fi
 
 # Args that will be passed into all testsys invocations.
 CARGO_MAKE_TESTSYS_ARGS = "${CARGO_MAKE_TESTSYS_KUBECONFIG_ARG}"
+
+TESTSYS_TEST_CONFIG_PATH = { script = [
+'''
+if [ -s "${TESTSYS_TEST_CONFIG_PATH}" ] && [ -s "${TESTSYS_TESTS_DIR}/Test.toml" ];then
+   echo "There can only be 1 config file. 2 config files were found: '${TESTSYS_TEST_CONFIG_PATH}' and '${TESTSYS_TESTS_DIR}/Test.toml'"
+   exit 1
+fi
+if [ -s "${TESTSYS_TEST_CONFIG_PATH}" ]; then
+   # If the config path exists
+   echo "${TESTSYS_TEST_CONFIG_PATH}"
+elif [ -s "${TESTSYS_TESTS_DIR}/Test.toml" ]; then
+   # If the test config is in the TESTSYS_TESTS_DIR it should be used.
+   echo "${TESTSYS_TESTS_DIR}/Test.toml"
+else
+   echo "${TESTSYS_TEST_CONFIG_PATH}"
+fi
+'''
+] }
 
 [tasks.setup]
 script = [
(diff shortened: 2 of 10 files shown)


webern (commented):
Can you improve the title of the PR. PR titles are important when sifting through changes for the changelog, etc.

tools/testsys/src/crds.rs
```diff
@@ -166,6 +168,42 @@ impl<'a> CrdInput<'a> {
...
+        let test_type = &self.test_type.to_string();
+        // List all acceptable paths to the custom crd to allow users some freedom in the way
+        // `tests` is organized.
+        let acceptable_paths = vec![
+            // Check for <TEST-TYPE>.yaml in the top level directory
+            PathBuf::new().join(test_type).with_extension("yaml"),
+            // Check for <TESTSYS_FOLDER>/<TEST-TYPE>/<TEST-TYPE>.yaml
+            self.tests_directory
+                .join(test_type)
+                .join(test_type)
```
stmcginnis, line 239: We don't want `.join(test_type)` twice, do we?

ecpullen: We do. This allows putting the test yaml at `<TESTSYS_FOLDER>/<TEST-TYPE>/<TEST-TYPE>.yaml`

stmcginnis: Ah, sorry, I see it now.

stmcginnis (approved):
Finally had a chance to read through some more. Everything looks good to me!

webern (approved):

.gitignore
```diff
@@ -14,3 +14,4 @@
 /Licenses.toml
 /licenses
 *.run
+/tests
```
bcressey, line 17: If there's any chance of larger files ending up here, it might be worth adding to `.dockerignore` as well.

Makefile.toml
```diff
@@ -123,6 +123,7 @@ CARGO_MAKE_DEFAULT_TESTSYS_KUBECONFIG_PATH = "${BUILDSYS_ROOT_DIR}/testsys.kubec
 TESTSYS_STARTING_VERSION = { script = ["git tag --list --sort=version:refname 'v*' | tail -1"] }
 # The commit for the last release of bottlerocket.
 TESTSYS_STARTING_COMMIT = { script = ["git describe --tag ${TESTSYS_STARTING_VERSION} --always --exclude '*' || echo 00000000"] }
+TESTSYS_FOLDER = "${BUILDSYS_ROOT_DIR}/tests"
```
bcressey: "folder" is a bit of a Windows-ism:
```suggestion
TESTSYS_TESTS_DIR = "${BUILDSYS_ROOT_DIR}/tests"
```

Makefile.toml
```diff
@@ -196,6 +200,20 @@ fi
...
+'''
+if [ -f "${TESTSYS_TEST_CONFIG_PATH}" ]; then
+   # If the config path exists
+   echo "${TESTSYS_TEST_CONFIG_PATH}"
+elif [ -f "${TESTSYS_FOLDER}/Test.toml" ]; then
+   # If the test config is in the TESTSYS_FOLDER it should be used.
+   echo "${TESTSYS_FOLDER}/Test.toml"
+else
+   echo "${TESTSYS_TEST_CONFIG_PATH}"
+fi
```
bcressey: Is this the right order of precedence? `Test.toml` from the root rather than the `tests` sub-directory? It's not consistent with `testsys.kubeconfig`.

Either way, I'd consider adding an assert that we don't have two `Test.toml` or `testsys.kubeconfig` files, one in the root and one in `tests`, since that's pretty likely to be an accident that will lead to unexpected behavior.

as a nit: `[ -s "${FILE}" ]` is often preferable to `-f`, because `-s` will also assert that the file is non-zero sized, which tends to correlate with usefulness.

tools/testsys/src/crds.rs
```diff
@@ -159,13 +215,110 @@ impl<'a> CrdInput<'a> {
...
+                .with_extension("yaml"),
+            // Check for <TESTSYS_FOLDER>/<TEST-TYPE>/test.yaml
+            self.tests_directory.join(test_type).join("test.yaml"),
+            // Check for <TESTSYS_FOLDER>/<TEST-TYPE>/crd.yaml
+            self.tests_directory.join(test_type).join("crd.yaml"),
+            // Check for <TESTSYS_FOLDER>/<TEST-TYPE>/crds.yaml
+            self.tests_directory.join(test_type).join("crds.yaml"),
+            // Check for <TESTSYS_FOLDER>/shared/<TEST-TYPE>.yaml
```
bcressey: Do we need to support all three filenames? I'd prefer to be a bit more opinionated - `crd.yaml` only. `test.yaml` is a bit likely to be confused with `Test.toml`.

tools/testsys/src/crds.rs
```diff
@@ -159,13 +215,110 @@ impl<'a> CrdInput<'a> {
...
+    fn cluster_config_file_path(&self, cluster_name: &str) -> Option<PathBuf> {
+        let test_type = &self.test_type.to_string();
+        // List all acceptable paths to the custom crd to allow users some freedom in the way
+        // `tests` is organized.
+        let acceptable_paths = vec![
+            // Check for <TESTSYS_FOLDER>/<TEST-TYPE>/<CLUSTER-NAME>.yaml
+            self.tests_directory
+                .join(test_type)
+                .join(cluster_name)
+                .with_extension("yaml"),
```
bcressey, line 270: There's some potential ambiguity here, e.g. if my cluster is named "quick" then we'd resolve the same file for both the cluster config and the CRD template.

What do you think about enforcing a convention that encodes the file type in the name, like `<CLUSTER-NAME>.cluster.yaml` vs `<CLUSTER-NAME>.test.yaml`?

I'm a bit sad that there's no orthogonal case to `<TESTSYS_FOLDER>/<TEST-TYPE>/crd.yaml` here; could we also support `<TESTSYS_FOLDER>/<TEST-TYPE>/cluster.yaml`, or does that not make sense for some reason?

ecpullen: `<TESTSYS_FOLDER>/<TEST-TYPE>/cluster.yaml` doesn't work because we support running a test on multiple clusters.

Makefile.toml
ecpullen: The reason for the order is to support a user having a single `TESTSYS_TESTS_DIR` but multiple `TEST.toml`s

tools/testsys/src/crds.rs
ecpullen: The name of the cluster config does not have the be the name of the cluster. If you wanted a cluster named quick, you could name the cluster config `quick-cluster.yaml` to avoid the collision.

bcressey (approved):

Issue #2656, opened 2022-12-13
Create a way to skip SELinux relabelling a volume


rpkelly:
**What I'd like:**
An option or series of options to skip relabeling all files in an attached volume.

**Any alternatives you've considered:**
Marking a volume `readonly` already this, but for obvious reasons is not always a workable solution.

bcressey:
With the policy change in #2738, it is now possible to skip relabeling by mounting the filesystem with this option:
```
context=system_u:object_r:local_t:s0
```

(Using an object type other than `local_t` is not permitted.)

For CSI drivers, there is an additional hurdle. The `mount` command is frequently built to be SELinux-aware, and will query libselinux to see if it reports that SELinux is enabled. If not, it will silently discard SELinux-related options like `context`. libselinux in turn will check for a mounted selinuxfs and an `/etc/selinux/config` with the `SELINUX=` option set. These conditions are chosen so that most containers will report that SELinux is disabled unless the container specifically opts in.

So the first step is to add volumes for the CSI driver to mount the SELinux-related paths from the host:
```
volumes:
# other CSI driver volumes
- hostPath:
    path: /var/lib/kubelet
    type: Directory
  name: kubelet-dir
...
# SELinux specific volumes
- hostPath:
    path: /sys/fs/selinux
    type: Directory
  name: selinuxfs
- hostPath:
    path: /etc/selinux/config
    type: FileOrCreate
  name: selinux-config
```

Then the node agent needs to place these mounts in the expected location:
```
volumeMounts:
# other CSI driver volume mounts
- mountPath: /var/lib/kubelet
  mountPropagation: Bidirectional
  name: kubelet-dir
...
# SELinux specific volume mounts
- mountPath: /sys/fs/selinux
  name: selinuxfs
- mountPath: /etc/selinux/config
  name: selinux-config
```

After that, the mount option can be specified in the storage class for the CSI driver:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
volumeBindingMode: WaitForFirstConsumer
parameters:
  fsType: ext4
  type: gp3
mountOptions:
  - context="system_u:object_r:local_t:s0"

Issue #2556, opened 2022-11-07
Kubelet volume relabel fails for Dynatrace OneAgent CSI volumes


baichinger:
**Image I'm using:**
Bottlerocket OS 1.10.1 (aws-k8s-1.23)

**What I expected to happen:**
Dynatrace OneAgent CSI volumes work on Bottlerocket.

**What actually happened:**
1. Dynatrace OneAgent CSI driver successfully mounts the volume to the pod:
    ```
    {"level":"info","ts":"2022-11-07T08:39:07.861Z","logger":"csi-driver","msg":"GRPC call","method":"NodePublishVolume","volume-id":"csi-1f0a5b5faf0aa280913d7a16ab0023393ec52aa563886dc7eebfdef711c13ca4"}
    {"level":"info","ts":"2022-11-07T08:39:07.861Z","logger":"csi-driver","msg":"publishing volume","csiMode":"app","target":"/var/lib/kubelet/pods/174ca84d-4ed5-4c9b-8f09-ce397f5c07a3/volumes/kubernetes.io~csi/oneagent-bin/mount","fstype":"","readonly":false,"volumeID":"csi-1f0a5b5faf0aa280913d7a16ab0023393ec52aa563886dc7eebfdef711c13ca4","attributes":{"csi.storage.k8s.io/ephemeral":"true","csi.storage.k8s.io/pod.name":"pause","csi.storage.k8s.io/pod.namespace":"default","csi.storage.k8s.io/pod.uid":"174ca84d-4ed5-4c9b-8f09-ce397f5c07a3","csi.storage.k8s.io/serviceAccount.name":"default","dynakube":"pup86408-application-monitoring","mode":"app"},"mountflags":[]}
    ```
2. Kubelet tries to relabel the files and fails:
    ```
    Error: failed to generate container "d3408d8e5fbc55df1769265072ade6c67c6a7176bafdae267f4fff904fdb7844" spec: failed to generate spec: relabel "/var/lib/kubelet/pods/174ca84d-4ed5-4c9b-8f09-ce397f5c07a3/volumes/kubernetes.io~csi/oneagent-bin/mount" with "system_u:object_r:data_t:s0:c65,c249" failed: lsetxattr /var/lib/kubelet/pods/174ca84d-4ed5-4c9b-8f09-ce397f5c07a3/volumes/kubernetes.io~csi/oneagent-bin/mount/agent/bin: permission denied
    ```
[...]

PR #2738, opened 2023-01-17
widen access to filesystem relabel in SELinux policy

bcressey:
**Issue number:**
#2556, #2656

**Description of changes:**
Adjust the default label for filesystems using extended attributes to store SELinux labels from `any_t` to `local_t`, partly for consistency with the rest of the policy, and partly because I'm more comfortable allowing a "no-op" relabel from `local_t` to `local_t` than allowing an actual label change.

Split out filesystem relabel permissions into a separate permission set, so they can be granted independently from the systems management permissions.

Add a rule to allow containers to use filesystem mount options to "relabel" from `local_t` to `local_t`. A filesystem relabeled with the `context=` option won't be subsequently relabeled by `containerd`.

This can be useful in cases like #2556, where relabeling files isn't permitted by the SELinux policy, and in cases like #2656, where relabeling files may take an excessively long time.

**Testing done:**
Verified that `aws-k8s-1.24` and `aws-ecs-1` can boot and run containers with no AVC denials.

Confirmed that the EBS CSI driver could create `ext4` and `xfs` volumes where the context is specified as a mount option in the storage class, like this:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
provisioner: ebs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
mountOptions:
  - context=system_u:object_r:local_t:s0
parameters:
  fsType: ext4
```

Verified that other context options, such as `context=system_u:object_r:any_t:s0`, are blocked by the SELinux policy.

**Terms of contribution:**

### DIFF
--- a/packages/selinux-policy/files.cil
+++ b/packages/selinux-policy/files.cil
@@ -345,3 +345,7 @@
     execute ioctl getattr map open read execmod
     relabelfrom relabelto mounton quotaon
     watch watch_mount watch_reads watch_sb watch_with_perm))))
+
+; Permission group for filesystems.
+(classmap filesystems (relabel))
+(classmapping filesystems relabel (filesystem (relabelfrom relabelto)))
--- a/packages/selinux-policy/fs.cil
+++ b/packages/selinux-policy/fs.cil
@@ -1,7 +1,7 @@
 ; Label inodes by using xattrs for supported filesystems.
-(fsuse xattr ext4 any)
-(fsuse xattr overlay any)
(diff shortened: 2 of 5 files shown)


markusboehme (approved):

stmcginnis (approved):
Can't speak for the security aspects, but as far as comments and structure, everything looks good to me.

foersleo (approved):

PR #2741, opened 2023-01-18
Enable `systemd-networkd` development via build flag

zmrow:
**Issue number:**
Related to: #2449 

**Description of changes:**
This initial set of commits enables development for the project's transition to `systemd-networkd`, allowing developers to conditionally include `systemd-networkd` (and exclude `wicked`) in any variant.

A new subpackage `networkd` has been added to the `systemd` package and contains all of the relevant files, excluding them from the broader `systemd` build.  At least for now, it appears we can continue to avoid `polkit` so those files continue to be excluded.  

A new build flag `systemd-networkd` has been added, which can be tacked on to the `image-features` of a variant to disable wicked and enable `systemd-networkd`.  Under the hood, the flag conditionally builds `systemd-networkd` in the `os` package.  Sample variant config snippet:
```
grub-set-private-var = true
systemd-networkd = true
```

**Testing done:**
* Build and boot an `aws-dev` and `aws-k8s-1.22` variant without the build flag to ensure it works as expected and only `wicked` artifacts are included
* Build and locally boot an `aws-dev` variant with the build flag enabled to ensure the `os` package gets rebuilt and `systemd-networkd` is added instead of `wicked`.  I added an early dependency on `console-getty.service` to the `preconfigured.target` since I knew early services would be broken but I still wanted to be able to poke around and verify the above.
:tada: 
```
bash-5.1# networkctl 
IDX LINK TYPE     OPERATIONAL SETUP    
  1 lo   loopback carrier     unmanaged
  2 eth0 ether    off         unmanaged

2 links listed.

bash-5.1# find / -name "*wicked*"
bash-5.1# 
```

**Terms of contribution:**

### DIFF
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -121,6 +121,11 @@ Summary: Bottlerocket userdata configuration system
 
 %package -n %{_cross_os}netdog
 Summary: Bottlerocket network configuration helper
+%if %{with systemd_networkd}
+Requires: %{_cross_os}systemd-networkd
+%else
+Requires: %{_cross_os}wicked
+%endif
 %description -n %{_cross_os}netdog
 %{summary}.
 
--- a/packages/systemd/systemd.spec
+++ b/packages/systemd/systemd.spec
@@ -95,6 +95,12 @@ Requires: %{name}
 %description devel
 %{summary}.
 
+%package networkd
+Summary: Files for networkd
+
+%description networkd
+%{summary}.
+
 %prep
 %autosetup -n systemd-stable-%{version} -p1
 
@@ -139,7 +145,7 @@ CONFIGURE_OPTS=(
  -Dsysext=false
  -Duserdb=false
  -Dhomed=false
- -Dnetworkd=false
+ -Dnetworkd=true
  -Dtimedated=false
  -Dtimesyncd=false
  -Dremote=false
@@ -321,6 +327,12 @@ install -p -m 0644 %{S:4} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/i
 %{_cross_libdir}/sysusers.d/*
 %{_cross_libdir}/systemd/*
 %{_cross_libdir}/udev/*
+%exclude %{_cross_libdir}/sysusers.d/systemd-network.conf
+%exclude %{_cross_libdir}/systemd/systemd-networkd
+%exclude %{_cross_libdir}/systemd/systemd-networkd-wait-online
+%exclude %{_cross_libdir}/systemd/system/systemd-networkd.service
+%exclude %{_cross_libdir}/systemd/system/systemd-networkd-wait-online.service
+%exclude %{_cross_libdir}/systemd/system/systemd-networkd.socket
 
 %{_cross_tmpfilesdir}/*
 %exclude %{_cross_tmpfilesdir}/x11.conf
@@ -332,6 +344,8 @@ install -p -m 0644 %{S:4} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/i
 
 %{_cross_datadir}/dbus-1/*
 %exclude %{_cross_datadir}/polkit-1
+%exclude %{_cross_datadir}/dbus-1/system-services/org.freedesktop.network1.service
+%exclude %{_cross_datadir}/dbus-1/system.d/org.freedesktop.network1.conf
 
 %dir %{_cross_factorydir}
 %{_cross_factorydir}%{_cross_sysconfdir}/issue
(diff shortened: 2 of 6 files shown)


packages/systemd/9012-adjust-default-networkd-dbus-policy.patch
```diff
@@ -0,0 +1,13 @@
...
++++ b/src/network/org.freedesktop.network1.conf
+@@ -20,7 +20,7 @@
+         </policy>
+
+         <policy context="default">
+-                <allow send_destination="org.freedesktop.network1"/>
++                <deny send_destination="org.freedesktop.network1"/>
+                 <allow receive_sender="org.freedesktop.network1"/>
+         </policy>
+
```
bcressey: Why do this?

That's sort of a multi-layered question. Patch files added to the project should always contain a header to work out the reasoning behind the change. There are some historical patches that don't follow that rule, but it's been consistently applied for a while now.

But I'm also not sure why this is necessary - is there a security issue that you're trying to mitigate? If so, how do we know this change is working as intended? What worked before that doesn't work now?

packages/systemd/systemd.spec
```diff
@@ -321,6 +331,12 @@ install -p -m 0644 %{S:4} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/i
 %{_cross_libdir}/sysusers.d/*
 %{_cross_libdir}/systemd/*
 %{_cross_libdir}/udev/*
+%exclude %{_cross_libdir}/sysusers.d/systemd-network.conf
+%exclude %{_cross_libdir}/systemd/systemd-networkd
+%exclude %{_cross_libdir}/systemd/systemd-networkd-wait-online
+%exclude %{_cross_libdir}/systemd/system/systemd-networkd.service
+%exclude %{_cross_libdir}/systemd/system/systemd-networkd-wait-online.service
+%exclude %{_cross_libdir}/systemd/system/systemd-networkd.socket
```
arnaldo2792, line 335: I don't think these are necessary. AFAIU, since these files will only be included in the subpackage they will never be part of the main package and thus don't have to be excluded

packages/os/os.spec
```diff
@@ -121,6 +121,11 @@ Summary: Bottlerocket userdata configuration system
 
 %package -n %{_cross_os}netdog
 Summary: Bottlerocket network configuration helper
+%if %{with systemd_networkd}
+Requires: %{_cross_os}systemd-networkd
+%else
+Requires: %{_cross_os}wicked
+%endif
```
arnaldo2792, line 128: Why the move from `release` to `os`? Whenever there is a "dog" here that has a dependency on a third-party-package, `os/Cargo.toml` is updated with a comment as well as the spec, but the package is still a dependency in `release.spec`. See `kexec-tools` and `prairiedog` for reference

zmrow: Perhaps I'm misunderstanding, but I think the difference here is that I need to explicitly include the required networking backend and exclude the other.  If I leave `wicked` as a `Requires` in `release.spec` it will always be pulled in.

packages/systemd/systemd.spec
zmrow: If I don't explicitly exclude these here, they get brought in with the `systemd` package via the `%{_cross_libdir}/{sysusers.d,systemd}/*` wildcards directly above.  This means without the `systemd-networkd` build flag set,  both `wicked` and `systemd-networkd` are brought in and they tend to disagree about networking :)  

I wanted to ensure these files are only brought in if the `systemd-networkd` subpackage is included.  If there's a better way to do that, let me know!

packages/os/os.spec
arnaldo2792: You can still use `%{with systemd_networkd}` from the `release.spec` file (unless somebody in the team disagrees).

zmrow: > You can still use `%{with systemd_networkd}` from the `release.spec` file (unless somebody in the team disagrees).

You're right, I could do that.  I was trying to avoid proliferating additional places the macro was used  that we would need to clean up if/when `wicked` is removed from the tree.

bcressey: I would rather add the condition to `os.spec` - which is already heavily conditionalized - rather than making `release.spec` variant-specific.

packages/systemd/9012-adjust-default-networkd-dbus-policy.patch
zmrow: Good call out.  I suppose I jumped the gun a bit by including this patch.  I fully expect to need to examine this DBUS policy and further lock it down to avoid unintended or malicious changes to a hosts networking config and stay inline with Bottlerocket's overall security posture.

That being said, I meant for this PR to enable development only and would rather not derail on this topic.  I've removed the patch from this PR and added an item to the [task list](https://github.com/bottlerocket-os/bottlerocket/issues/2449#issuecomment-1334535852) in #2449 to review and implement a proper security policy for this.

packages/os/os.spec
arnaldo2792: Sounds good. There is still a missing comment in `os/Cargo.toml` referencing the new "dependency" :+1:

bcressey (approved):

yeazelm (approved):
LGTM!

zmrow:
^ Adds the missing comment requested by @arnaldo2792

arnaldo2792 (approved):
:tada:

PR #2744, opened 2023-01-21
Parse netconfig

yeazelm:
**Description of changes:**

This is a change to introduce better ways to test `net.toml` besides booting an image. This adds a `cargo make netdog-parse` target which allows testing the provided `NET_CONFIG_INPUT` file. It runs it in a container so that the path can be used that currently exists in the code. We could also shift to passing an argument to netdog that takes in the file instead as well. 

**Testing done:**
I tested several different options to see how it works:
Working file:
```bash
$  VARIANT=metal-dev cargo run --manifest-path sources/api/netdog/Cargo.toml -- validate-net-config -f sources/api/netdog/test_data/net_config/net_config.toml

eno2 found as primary interface
Found eno1
Found eno2
Found eno3
Found eno4
Found eno5
Found eno6
Found eno7
Found eno8
Found eno9
Found eno10
net.toml file provided successfully parsed!
```
or a bad test:
```bash
$ VARIANT=metal-dev cargo run --manifest-path sources/api/netdog/Cargo.toml -- validate-net-config -f sources/api/netdog/test_data/net_config/basic/bad_version.toml

Unable to read/parse network config from 'sources/api/netdog/test_data/net_config/basic/bad_version.toml': Invalid network configuration: Unknown network config version: 50
```

**Terms of contribution:**


stmcginnis (approved):
Awesome, I could see this being very useful!

Makefile.toml
```diff
@@ -614,6 +614,46 @@ docker run --rm \
...
+  cargo install \
+    ${CARGO_MAKE_CARGO_ARGS} \
+    --path sources/api/netdog \
+    --root tools \
+    --force \
+    --quiet
+  '''
+]
+
+[tasks.netdog-parse]
```
webern: I'm wondering if this belongs in the Makefile.toml. Since it's a developer workflow anyway it could be done locally `cargo run --package net-dog -- parse --file blah`

Is it in the Makefile so that users (i.e. not just developers) can use it? If so, is there documentation missing?

yeazelm: Yes, there is documentation missing, I actually was writing that this morning since I realized I forgot that. I mostly wanted to see if using a docker container this way was acceptable or if we wanted to plumb through providing the file directly and let it run natively like you suggest. We have similar workflows for bootconfig which made me think it was desirable to put this in the Makefile.

webern: > which made me think it was desirable to put this in the Makefile.

You may be right, let's see if anyone else chimes in. I can see this being useful to metal users. And if so, then having it in the Makefile simplifies things for them somewhat.

arnaldo2792: > Since it's a developer workflow anyway it could be done locally cargo run --package net-dog -- parse --file blah

@webern, are you saying that creating a `net.toml` configuration file is a developer workflow? If that's the case, I disagree, I think it is both a developer and user workflow, since as @yeazelm said, a user shouldn't have to boot an instance in order to check if the configuration is valid, e.g. I don't want to wait until I boot my instance to know that I have a typo.

sources/api/netdog/src/cli/check_net_config.rs
```diff
@@ -0,0 +1,38 @@
...
+pub(crate) struct CheckNetConfigArgs {}
+
+/// Check that the configuration for network interfaces parses
+pub(crate) fn run() -> Result<()> {
+    // `maybe_net_config` could be `None` if no interfaces were defined
+    let net_config = match check_net_config() {
+        Ok(Some(net_config)) => net_config,
+        Ok(None) => {
+            eprintln!("No network interfaces were configured");
+            return Ok(());
```
arnaldo2792: Shouldn't the process exit non-zero if the configuration is invalid? i.e. shouldn't this return an actual error?

arnaldo2792 (approved):

sources/api/netdog/src/cli/check_net_config.rs
yeazelm: Thats a good question. For `generate-net-config` its considered desirable to exit 0 on this case. Technically it did parse correctly, but it would not result in useful network configuration (which my update to the docs says is technically right). I guess we could envision this being added to a workflow to validate the configuration which would mean we should be very precise about exit codes.

Makefile.toml
zmrow: I would agree that it's both a developer and a user workflow.  

A Bottlerocket developer will have cargo/rust/cargo-make and other build tools available, so the current workflow is fine and provides a nice easy-access way to check a net.toml.

A Bottlerocket user/admin may not have all of the above tools installed, so it would be nice if this were available in the SDK so they wouldn't need to install anything but Docker.  However, putting this in the SDK container comes with it's own set of considerations.

That being said, I'm ok with starting here for now.

PROVISIONING-METAL.md
```diff
@@ -265,6 +265,38 @@ Older networking configuration versions (such as `1` or `2`) are supported in ne
 | Version 2                     | [v1.10.0](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.10.0) |
 | Version 3                     | Unreleased                                                                      |
 
+#### Validation network configuration
```
zmrow: s/Validation/Validate

PROVISIONING-METAL.md
```diff
@@ -265,6 +265,38 @@ Older networking configuration versions (such as `1` or `2`) are supported in ne
...
+
+There is a helper `make` target `netdog-parse` to help validate that network configuration files parse correctly.
```
zmrow: nit, a bit wordy: "There is a `make` target `netdog-parse` to validate network configuration..."

PROVISIONING-METAL.md
```diff
@@ -265,6 +265,38 @@ Older networking configuration versions (such as `1` or `2`) are supported in ne
...
+This target is intended to provide a way to ensure the provided file parses before attempting a boot.
+It won't guarantee that the generated configuration will work in a particular context, but is aimed at validation of format and structure.
```
zmrow: nit, we could probably combine these sentences:  
"This target is intended to validate the format and structure of the file, it will not guarantee the generated configuration will work in a particular context."

PROVISIONING-METAL.md
```diff
@@ -265,6 +265,38 @@ Older networking configuration versions (such as `1` or `2`) are supported in ne
...
+The command can be run by passing the absolute path to a `net.toml` file:
+```bash
+cargo make -e VARIANT=metal-dev -e NET_CONFIG_INPUT=/home/yeazelm/git/bottlerocket/sources/api/netdog/test_data/net_config/net_config.toml netdog-parse
```
zmrow: Is `VARIANT` necessary here?

Makefile.toml
```diff
@@ -614,6 +614,46 @@ docker run --rm \
...
+script = [
+  '''
+  set -euo pipefail
+
+  if [ ! -s "${NET_CONFIG_INPUT}" ]; then
+     echo "No network config file exists, please create one at ${NET_CONFIG_INPUT}"
+     exit 1
+  fi
+
+  docker run --rm \
```
zmrow: What do we gain by running this in a container if the tool is installed?  In the case of failure, it makes the resulting error message a bit weird by referencing `/var/lib/bottlerocket/net.toml` which probably won't exist on a user's system.

Makefile.toml
zmrow: I might also name this target more generically: `validate-net-config` or something.  `netdog-parse` is fairly specific and non-intuitive unless you know the system really well.

sources/api/netdog/src/cli/check_net_config.rs
```diff
@@ -0,0 +1,38 @@
+use super::{check_net_config, error, Result};
+use argh::FromArgs;
+use snafu::OptionExt;
+
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "check-net-config")]
```
zmrow: nit: We can get into semantics here, but I'd rather see this named `verify-net-config` or `validate-net-config`.

sources/api/netdog/src/cli/check_net_config.rs
```diff
@@ -0,0 +1,38 @@
...
+/// Check the network configuration
+pub(crate) struct CheckNetConfigArgs {}
```
zmrow: In my mind, it makes sense to take a path to a network config as an argument here.

Doing so would eliminate the shared logic between this command and `generate-net-config`, which makes sense because here we just need to validate a file, and don't care about it's location.

sources/api/netdog/src/cli/mod.rs
```diff
@@ -47,6 +54,25 @@ where
     Ok(())
 }
 
+/// Search for the network configuration file and return the configuration if it parses successfully,
+/// otherwise return an error
+fn check_net_config() -> Result<Option<Box<dyn Interfaces>>> {
```
zmrow: As previously mentioned, I don't think we need or want to share this logic.  This logic is specific to a running Bottlerocket system because of the assumed paths.

sources/api/netdog/src/main.rs
```diff
@@ -21,10 +21,15 @@ name of the interface, and valid options are "dhcp4" and "dhcp6".  A "?" may be
...
+The subcommand `check-net-config` follows the same logic as `generate-net-config` to find the
+network interface configuration for the host. Rather than generate the configuration, it prints very
+basic information about the found configuration or the error found while attempting to parse the
+file. This command is intended to be run by a user to test network configuration files.
+
 The subcommand `write-resolv-conf` writes the resolv.conf, favoring DNS API settings and
 supplementing any missing settings with DNS settings from the primary interface's DHCP lease.  It
 is meant to be used as a restart command for DNS API settings.
-*/
+ */
```
zmrow: nit: extra space here

PROVISIONING-METAL.md
yeazelm: Yes. This is a change from the IMDS where we conditionally compile things in for `aws` variants so its now required.

PROVISIONING-METAL.md
yeazelm: Whoops, I'll correct the typo, I had intended to say "Validating" but "Validate" works as well.

PROVISIONING-METAL.md
yeazelm: yeah, I added things in and didn't remove, I'll rephrase.

PROVISIONING-METAL.md
yeazelm: Yeah, works for me.

Makefile.toml
yeazelm: It helps to avoid adding in logic just for taking in a file via an argument and instead uses the same lookup logic that `generate-net-config` uses. This all came from when I was going to just call `generate-net-config` and shuffle things in and out of the container to make things *just work* but now that we are creating a separate command for parse-only, maybe it makes sense to just call it with an argument instead.

Makefile.toml
yeazelm: That makes sense, I like `validate-net-config` more than mine.

sources/api/netdog/src/cli/check_net_config.rs
yeazelm: Yeah, I'm fine with that. I've chosen 3 different names because I couldn't decide which one I liked the most. Validate does seem to be the wording we are settling on so I'll make the code consistent.

sources/api/netdog/src/cli/check_net_config.rs
yeazelm: That works for me. I hadn't really leaned into "just for users and developers" and was trying to keep it more consistent with the rest of the usage in `netdog` but if folks are ok with us sharing use cases, I think its a much cleaner interface for the user.

Makefile.toml
cbgbt: A `cargo install` option might be a nice alternative to including this in the SDK -- you'd be able to avoid pulling down the container image if you hadn't already.

webern (approved):

sources/api/netdog/src/main.rs
yeazelm: For some reason my IDE keeps putting this back if I don't watch it. Very strange. I'll fix nonetheless.

Makefile.toml
yeazelm: I pulled this out of the PR. It sounds like we probably need to provide something different for folks that don't have the repo pulled and rust set up. Nonetheless, the new doc line works on different architectures much better, so one should be able to use their local machine (like a Mac) to run this command which I think might actually be an improvement until we solve the tools problem more broadly.

sources/api/netdog/src/cli/check_net_config.rs
yeazelm: In the case that this branch is hit, the following would be printed:
```bash
No network interfaces were configured
```
with a zero exit. The file *did* parse, but it probably doesn't provide meaningful network configuration. I think this is still correct behavior since its just validating that it parses, not that it will work for a particular setup.

Makefile.toml
yeazelm: I've removed this from the PR, we will solve this problem a different way.

PROVISIONING-METAL.md
```diff
@@ -265,6 +265,35 @@ Older networking configuration versions (such as `1` or `2`) are supported in ne
 | Version 2                     | [v1.10.0](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.10.0) |
 | Version 3                     | [v1.12.0](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.12.0) |
 
+#### Validate network configuration
+
+There is a `make` target `netdog-parse` to validate that network configuration files parse correctly.
```
webern: no there isn't

yeazelm: Whoops, I updated the command, not the docs.

PROVISIONING-METAL.md
```diff
@@ -265,6 +265,35 @@ Older networking configuration versions (such as `1` or `2`) are supported in ne
 | Version 2                     | [v1.10.0](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.10.0) |
 | Version 3                     | [v1.12.0](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.12.0) |
 
+#### Validate network configuration
+
+`netdog` has a command `validate-net-config` to validate that network configuration files parse correctly.
+This command is intended to validate the format and structure of the file, it will not guarantee the generated configuration will work in a particular context.
+The command can be run by passing the path to a `net.toml` file:
+```bash
```
webern: nit: throughout this file we are separating code blocks from text with a newline.

```suggestion
The command can be run by passing the path to a `net.toml` file:

```bash
```

webern (approved):

webern (approved):

PR #2745, opened 2023-01-23
fix clippy warnings for sources

### LINKED ISSUE #1776: build: check clippy and cargo fmt during github actions checks

bcressey:
**Issue number:**

Fixes #1776 (finally)

**Description of changes:**
Fix all clippy warnings for the `sources` tree, and add coverage to the CI lint checks.

**Testing done:**
`cargo make check-lints` passes for three platforms:
* `aws-k8s-1.24`
* `vmware-k8s-1.24`
* `metal-k8s-1.24`

**Terms of contribution:**

### DIFF
--- a/sources/api/datastore/src/key.rs
+++ b/sources/api/datastore/src/key.rs
@@ -79,9 +79,9 @@ impl Key {
     where
         S: AsRef<str>,
     {
-        let name = Self::encode_name_segments(&segments)?;
+        let name = Self::encode_name_segments(segments)?;
 
-        Self::check_key(key_type, &name, &segments)?;
+        Self::check_key(key_type, &name, segments)?;
 
         Ok(Self {
             name,
@@ -208,7 +208,7 @@ impl Key {
         match key_type {
             KeyType::Data => {
                 ensure!(
-                    segments.len() >= 1,
+                    !segments.is_empty(),
                     error::InvalidKeySnafu {
                         name,
                         msg: "data keys must have at least one segment",
@@ -232,10 +232,7 @@ impl Key {
     /// Determines whether a character is acceptable within a segment of a key name.  This is
     /// separate from quoting; if a character isn't valid, it isn't valid quoted, either.
     fn valid_character(c: char) -> bool {
-        match c {
-            'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '-' | '/' => true,
-            _ => false,
-        }
+        matches!(c, 'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '-' | '/')
     }
 
     /// Given a key name, returns a list of its name segments, separated by KEY_SEPARATOR.
--- a/sources/api/migration/migration-helpers/src/common_migrations.rs
+++ b/sources/api/migration/migration-helpers/src/common_migrations.rs
@@ -1,4 +1,4 @@
-use crate::{error, Metadata, Migration, MigrationData, Result};
+use crate::{error, Migration, MigrationData, Result};
 use serde::Serialize;
 use snafu::{OptionExt, ResultExt};
 use std::collections::HashMap;
@@ -632,17 +632,14 @@ impl ReplaceTemplateMigration {
         incoming_template: &str,
         input: &mut MigrationData,
     ) -> Result<()> {
-        if let Some(template) = &self.get_setting_template(&input) {
+        if let Some(template) = &self.get_setting_template(input) {
             if template == outgoing_template {
                 println!(
                     "Changing template of '{}' from '{}' to '{}'",
                     self.setting, outgoing_template, incoming_template
                 );
                 // Update the setting's template
-                let metadata = input
-                    .metadata
-                    .entry(self.setting.to_string())
-                    .or_insert(Metadata::new());
+                let metadata = input.metadata.entry(self.setting.to_string()).or_default();
                 metadata.insert(
                     "template".to_string(),
                     serde_json::Value::String(incoming_template.to_string()),
@@ -703,6 +700,7 @@ impl Migration for ReplaceTemplateMigration {
             self.update_template_and_data(
                 // Clone the input string; we need to give the function mutable access to
                 // the structure that contains the string, so we can't pass a reference into the structure.
+                #[allow(clippy::unnecessary_to_owned)]
                 &data.to_owned(),
                 self.old_template,
                 self.new_template,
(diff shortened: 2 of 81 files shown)


stmcginnis (approved):
A lot of files touched, but overall not too bad considering no linting was being performed before.

I tried looking through for any nuanced behavioral changes, but nothing stood out. Looks good to me.

sources/updater/update_metadata/src/lib.rs
```diff
@@ -369,23 +366,21 @@ impl Update {
...
-                    }
-                    let time_per_seed = wave_duration / num_seeds_allocated_to_wave;
-                    // Derive the target time position within the wave given the host's seed.
-                    let target_time = start_time + (time_per_seed * (seed as i32));
-                    // If the current time is past the target time position in the wave, the update is
-                    // marked available
-                    return time >= target_time;
+            if let Some((start_time, Some(end_time), start_seed, end_seed)) = bound {
+                // This host is not part of last wave
+                // Determine the duration of this host's wave
```
etungsten, line 371: I think this might be a case of the previous being better for readability? But this is also fine.

sources/cfsignal/src/cloudformation.rs
```diff
@@ -33,7 +33,7 @@ pub async fn signal_resource(
         .logical_resource_id(logical_resource_id)
         .status(
             aws_sdk_cloudformation::model::ResourceSignalStatus::from_str(&status)
-                .context(error::ParseStatusSnafu)?,
+                .expect("infallible"),
```
webern, line 36: I don't like it, but OK. I should read up on infallible errors.

zmrow (approved):
Thanks!

:gem:

etungsten (approved):

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -703,6 +700,7 @@ impl Migration for ReplaceTemplateMigration {
             self.update_template_and_data(
                 // Clone the input string; we need to give the function mutable access to
                 // the structure that contains the string, so we can't pass a reference into the structure.
+                #[allow(clippy::unnecessary_to_owned)]
                 &data.to_owned(),
```
webern, line 704: This doesn't work?

```suggestion
                &data,
```

webern (approved):
Nice, we probably want to merge this asap.

sources/api/thar-be-updates/src/status.rs
```diff
@@ -100,11 +100,9 @@ pub fn get_update_status(_lockfile: &File) -> Result<UpdateStatus> {
...
         path: UPDATE_STATUS_FILE,
     })?;
-    Ok(
-        serde_json::from_reader(status_file).context(error::StatusParseSnafu {
-            path: UPDATE_STATUS_FILE,
-        })?,
-    )
+    serde_json::from_reader(status_file).context(error::StatusParseSnafu {
+        path: UPDATE_STATUS_FILE,
+    })
```
yeazelm, line 105: How is the `Result` returned without the `Ok()`?

yeazelm (approved):
LGTM but I will admit I'm unsure of the `Box`ing of those errors.

sources/api/thar-be-updates/src/status.rs
webern: `.context` converts the `Result<JsonError` to `Result<OurError>` which is being returned since we are not using `?` on it.

yeazelm: Thanks!

sources/cfsignal/src/cloudformation.rs
bcressey: This avoids a clippy warning about unreachable code in the error type, behind the `Snafu` proc macro. I couldn't find a way to retain the `ParseStatus` enum variant and get rid of the warning.

sources/api/migration/migration-helpers/src/common_migrations.rs
bcressey: That's what clippy suggests, but it leads to this compiler error:
```
error[E0502]: cannot borrow `input` as mutable because it is also borrowed as immutable
   --> api/migration/migration-helpers/src/common_migrations.rs:706:17
    |
690 |         if let Some(input_value) = input.data.get(self.setting) {
    |                                    ---------------------------- immutable borrow occurs here
...
700 |             self.update_template_and_data(
    |                  ------------------------ immutable borrow later used by call
...
706 |                 &mut input,
    |                 ^^^^^^^^^^ mutable borrow occurs here
```

sources/api/migration/migrations/v1.12.0/add-k8s-autoscaling-warm-pool-setting-metadata/src/main.rs
```diff
@@ -10,7 +10,7 @@ fn run() -> Result<()> {
         migrate(AddMetadataMigration(&[SettingMetadata {
             metadata: &["affected-services"],
             setting: "settings.autoscaling",
-        }]));
+        }]))?;
     };
```
cbgbt, line 14: I hadn't considered that stuff like this isn't currently caught by CI. Awesome.

sources/api/datastore/src/key.rs
```diff
@@ -232,10 +232,7 @@ impl Key {
     /// Determines whether a character is acceptable within a segment of a key name.  This is
     /// separate from quoting; if a character isn't valid, it isn't valid quoted, either.
     fn valid_character(c: char) -> bool {
-        match c {
-            'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '-' | '/' => true,
-            _ => false,
-        }
+        matches!(c, 'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '-' | '/')
```
cbgbt, line 235: TIL about `matches!`

cbgbt (approved):
Nice! 🚀

sources/api/datastore/src/key.rs
bcressey: Me too!

PR #2750, opened 2023-01-25
netdog: Enable conditional compilation based on network backend

zmrow:
**Issue number:**
Related to #2449 

**Description of changes:**
This set of temporary changes enables conditional compilation of `netdog` based on network backend.  "Temporary" in this case means that these changes (similar to those in #2741 ) will be removed once `systemd-networkd` development has finished.

A new function `emit_net_backend_cfgs()` has been added to `netdog`'s `build.rs`  that triggers off the `SYSTEMD_NETWORKD` environment variable that has been added to the build environment.  The crate can now use the following to conditionally compile code:
```
#[cfg(net_backend = "systemd-networkd")]
{ do stuff }
```

**Testing done:**
Added the following blurb to `netdog generate-net-config` to be conditionally compiled:
```
#[cfg(net_backend = "systemd-networkd")]
{
    eprintln!("I'm on systemd-networkd!!!")
}
```

* Build and boot an aws-k8s-1.22 image that runs properly and connects to the cluster - do not observe the above lines in the journal
* Build and locally boot an aws-dev variant. I added an early dependency on console-getty.service to the preconfigured.target since I knew early services would be broken but I still wanted to be able to poke around and verify the above. With the `systemd-networkd` build flag, I observed the above message in the journal.  Without the flag, the message didn't show up.
```
Starting Network Configuration...
[    2.081577] netdog[901]: I'm on systemd-networkd!!!
```

**Terms of contribution:**


sources/bottlerocket-variant/src/lib.rs
```diff
@@ -370,6 +381,20 @@ impl PartialEq<Variant> for &str {
...
+/// TODO: Remove this function once systemd-networkd integration development is finished
+pub fn emit_net_backend_cfgs() {
+    let net_backend = if Some("1".to_string()) == std::env::var(NET_BACKEND_OVERRIDE_ENV).ok() {
+        NET_BACKEND_OVERRIDE
+    } else {
+        DEFAULT_NET_BACKEND
+    };
+    println!("cargo:rerun-if-env-changed={}", NET_BACKEND_OVERRIDE_ENV);
+    println!("cargo:rustc-cfg=net_backend=\"{}\"", net_backend);
+}
```
bcressey: I'd rather do this in a netdog-specific build script. It's not something most other crates will care about, while quite a few benefit from structured parsing of the variant tuple.

bcressey (commented):
I'm on the fence about merging this in the current state, since there's no real functionality tied to it, and it seems like an easy patch to maintain on a feature branch. If you can scope it down to `netdog` then it's probably OK.

sources/bottlerocket-variant/src/lib.rs
webern: > I'd rather do this in a netdog-specific build script. It's not something most other crates will care about, while quite a few benefit from structured parsing of the variant tuple.

Not sure why I didn't think that. I think when we discussed it, @zmrow, it was only netdog that needed this?

zmrow: That's fair.  My reasoning for putting it in the variant crate was "just in case we need it outside of netdog".  I'm fine with scoping it to netdog for now.

zmrow:
^ Scoped these changes to `netdog` per @bcressey

sources/api/netdog/build.rs
```diff
@@ -14,6 +25,19 @@ fn main() {
...
     generate_readme::from_main().unwrap();
 }
+
+/// Emit `cfg` values that can be used for conditional compilation based on the network backend.
+/// TODO: Remove this function once systemd-networkd integration development is finished
+fn emit_net_backend_cfgs() {
+    let net_backend = if Some("1".to_string()) == std::env::var(NET_BACKEND_OVERRIDE_ENV).ok() {
+        NET_BACKEND_OVERRIDE
+    } else {
+        DEFAULT_NET_BACKEND
```
webern, line 39: very much a nit: our other `cfg!` compile guards use lowercase, kebab-case strings, e.g. `aws-k8s` etc. So these screaming snake cases might look a little out of place at their usage sites.

webern (approved):

sources/api/netdog/build.rs
zmrow: The actual cfgs emitted will be: `wicked` and `systemd-networkd`.  These screaming snake case constants are set to those static strings. :)

webern: ah i see :+1:

bcressey (approved):

PR #2755, opened 2023-01-28
containerd: remove unnecessary patch

arnaldo2792:
**Description of changes:**

This patch wasn't deleted as part of the update in 47b629cd

**Testing done:**
- Executed a task in ECS
- Executed a pod in k8s 1.23
- Build runs successfully

**Terms of contribution:**


stmcginnis (approved):
Confirmed - no longer needed and no explicit reference to the patch in https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/containerd/containerd.spec

webern (approved):

PR #2757, opened 2023-01-29
Update docker run `--security-opt` command format

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

During our build process we have several places where we use the Bottlerocket SDK to perform the build. In these `docker run` commands we pass `--security-opt label:disable` to avoid labeling a large number of files.

It appears this syntax may have changed since we started using it, and the [correct syntax is now `label=disable`](https://docs.docker.com/engine/reference/run/#security-configuration). Based on reports in the Docker repo, the way we had it of `label:disable` [may not actually work](https://github.com/docker/for-linux/issues/1000).

While there are other issues to be addressed, the older format generates an error when trying to build with finch, while the newer format - [though not recognized](https://github.com/runfinch/finch/issues/192) - only generates a warning message.

**Testing done:**

Performed full build, made sure everything completed and no errors.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -441,7 +441,7 @@ rc=0
 if ! docker run --rm \
   --network=none \
   --user "$(id -u):$(id -g)" \
-  --security-opt label:disable \
+  --security-opt="label=disable" \
   -v "${BUILDSYS_TOOLS_DIR}":/tmp/tools \
   "${BUILDSYS_SDK_IMAGE}" \
   bash -c \
@@ -592,7 +592,7 @@ fi
 docker run --rm \
    --network=none \
    --user "$(id -u):$(id -g)" \
-   --security-opt label:disable \
+   --security-opt="label=disable" \
    -v "${BOOT_CONFIG_INPUT}":/tmp/bootconfig-input \
    -v "${boot_config}":/tmp/bootconfig.data \
    "${BUILDSYS_SDK_IMAGE}" \
@@ -613,7 +613,7 @@ script = [
 docker run --rm \
    --network=none \
    --user "$(id -u):$(id -g)" \
-   --security-opt label:disable \
+   --security-opt="label=disable" \
    -v "${BOOT_CONFIG}":/tmp/bootconfig.data \
    "${BUILDSYS_SDK_IMAGE}" \
    bootconfig -l /tmp/bootconfig.data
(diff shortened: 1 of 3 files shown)


webern:
Do you think there's any chance it is related to this? bottlerocket-os/twoliter#173

stmcginnis:
> Do you think there's any chance it is related to this? bottlerocket-os/twoliter#173

Tangentially related, only in that if you are not using Docker Desktop and are instead trying to use finch, having that argument the way it is now will completely blow up.

I'm not really sure about the issue with Docker not really enforcing it if it is not formatted correctly. It is possible that could be related to the failure, but I have a feeling we're going to find some other sharp edges if we try to support building on macOS.

cbgbt:
Is there significant overlap in Docker support for the new semantics vs the old? Do we need an "MSDV" (Minimum Supported Docker Version) advertised in the build instructions for Bottlerocket? Seems like a possible footgun for folks using the old docker client.

stmcginnis:
> Is there significant overlap in Docker support for the new semantics vs the old? Do we need an "MSDV" (Minimum Supported Docker Version) advertised in the build instructions for Bottlerocket? Seems like a possible footgun for folks using the old docker client.

Good question. From what I could find, the "new" syntax I am using here has been the correct syntax since _at least_ 2017.

https://github.com/docker/docker-ce/commit/5a52b7e76e910ff340667cd939f8c5b2215a2590#diff-817dee6c24c28e3521d4bcd856594f7af2dce8e1712f45451fbe9e90e39f2ea8R649

So my assumption would be that this PR is correcting the format to what it was always supposed to be. It used to handle parsing the non-official format we have been using, but that has never been the documented format for using this arg. Or if it was, it was so long ago that the risk of someone still using a client old enough where this could be an issue should be close to 0.

jpculp:
We should verify if this is still true, but in our [BUILDING.md](https://github.com/bottlerocket-os/bottlerocket/blob/develop/BUILDING.md) we recommend _"Docker 20.10.10 or later"_.

As far as if `label:disable` still works, I believe that it does, but Docker-like tools such as podman absolutely require the new syntax (at least last time I tinkered with it).

jpculp (approved):

cbgbt (approved):

PR #2772, opened 2023-02-01
bootstrap-containers: extend bootstrap-containers capabilities

### LINKED ISSUE #2652: Increase bootstrap containers permissions
**What I'd like:**

- Execute actions like modprobe or service startup inside a bootstrap container

I tried a lot of forms to do a modprobe. Executing sheltie inside a bootstrap container with:
- modprobe XXXX
- systemctl start modprobe@XXXX.service
- Modifying filesystem (impossible because it's Read-Only)

But the output is always "Operation Not Permitted". I guess it must be some SELinux restriction.

Works well inside a host container, but host container doesn't have "essential" param. So if the actions fails, the node boots 
without problems. In my case, I want to activate ip_vs_sed to use it with ipvs, so if the module activation fails, I will not be able to use ipvs and I will never be able to handle this err.

arnaldo2792:
**Issue number:**

Closes #2652

**Description of changes:**

This adds `CAP_SYS_MODULE` and `CAP_CHROOT` to bootstrap containers

**Testing done:**

- I created a container image with this definition:

```Dockerfile
FROM fedora
ENTRYPOINT ["chroot", "/.bottlerocket/rootfs", "modprobe", "vfio_pci"]
```

I confirmed the kernel module was loaded:

```bash
bash-5.1# lsmod | grep vfio
vfio_pci               77824  0
vfio_virqfd            16384  1 vfio_pci
vfio_iommu_type1       36864  0
vfio                   45056  2 vfio_iommu_type1,vfio_pci
irqbypass              16384  1 vfio_pci

```

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -660,9 +660,12 @@ func withBootstrap() oci.SpecOpts {
 		withRootFsShared(),
 		oci.WithSelinuxLabel("system_u:system_r:control_t:s0-s0:c0.c1023"),
 		// Bootstrap containers don't require all capabilities. We only add
-		// `CAP_SYS_ADMIN` for mounting filesystems, and `CAP_NET_ADMIN` for
-		// managing iptables rules.
-		oci.WithAddedCapabilities([]string{"CAP_SYS_ADMIN", "CAP_NET_ADMIN"}),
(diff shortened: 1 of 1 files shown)


arnaldo2792:
(forced push to fix typo in commit message)

cbgbt (approved):

bcressey (approved):

webern (approved):

PR #2774, opened 2023-02-02
kubelet.service: change the restart policy to always

### LINKED ISSUE #2512: kubelet service doesn't restart on failure
We are running Bottlerocket managed nodes in EKS, and last week got a node failure where kubelet.service stopped with the following log. That node was running version 1.9.2 on k8s 1.22.

```
Oct 21 05:10:31 ip-10-1-22-8.eu-west-1.compute.internal kubelet[520747]: W1021 05:10:31.864779  520747 clientconn.go:1326] [core] grpc: addrConn.createTransport failed to connect to {/run/dockershim.sock /run/dockershim.sock <nil> 0 <nil>}. Err: connection error: desc = "transport: failed to write client preface: write unix @->/run/dockershim.sock: write: broken pipe".Reconnecting...
```

After this failure the kubectl node status became "NotReady" and the NodeGroup status "Unknown", which leads to a state where nothing happens. The EC2 doesn't get terminated since the kubelet status is ot part of the health check, and the kubelet service is not restarted automatically so the node is not reachable.

The containers running on the node continued running, so nothing triggered outside of the node status.

After logging into the node and manually starting the kubelet service again everything came back, but later the kubelet service failed again. After that we terminated the node and a new one was created with version 1.10.1, it's been stable since that.

**What I'd like:**

Add a restart condition to the kubelet service.

AWS adding a health monitor for EKS nodes which take the kubelet health into consideration, but that would lead to a complete node termination which might not be required.

gthao313:
**Issue number:**

Closes #2512

**Description of changes:**
```
kubelet.service: change the restart policy to always
```
currently we only specify kubelet to be restarted when it on failure. However, the situation resources starvation evicts kubelet isn't recognized as failure, so kubelet won't be restarted on this situation. Therefore, setting kubelet restart policy to `always` can solve this issue.

**Testing done:**
- [x] created resource starvation issue to evict kubelet (exit code = 0)
        - launch small instance  and add heavy computable pods to the cluster
- [x] verify if kubelet has been restarted

been

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.21/kubelet.service
+++ b/packages/kubernetes-1.21/kubelet.service
@@ -14,7 +14,7 @@ ExecStartPre=/sbin/iptables -P FORWARD ACCEPT
 # Must be overridden by a drop-in file or `kubelet` won't start
 ExecStart=/usr/bin/false
 
-Restart=on-failure
+Restart=always
 RestartForceExitStatus=SIGPIPE
 RestartSec=5
 Delegate=yes
--- a/packages/kubernetes-1.22/kubelet.service
+++ b/packages/kubernetes-1.22/kubelet.service
@@ -14,7 +14,7 @@ ExecStartPre=/sbin/iptables -P FORWARD ACCEPT
 # Must be overridden by a drop-in file or `kubelet` won't start
 ExecStart=/usr/bin/false
 
-Restart=on-failure
+Restart=always
 RestartForceExitStatus=SIGPIPE
 RestartSec=5
 Delegate=yes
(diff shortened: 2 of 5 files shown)


zmrow (approved):

webern (approved):
Makes sense to me. @etungsten do you foresee any issues with this change?

etungsten (commented):
Looks fine to me as long as testing results are as expected.

stmcginnis (approved):

etungsten (approved):

PR #2793, opened 2023-02-09
Makefile: changed Rust lint clippy to the one included in SDK

### LINKED ISSUE #2704: Use the SDK for Linting
If we use the local installs of Go and Cargo/Rust to lint the code, we risk breaking every time the compiler updates and we risk having conflicting lints between GitHub actions and developers' local builds. This is especially true with clippy which aggressively adds lints with each update.

We should use the SDK toolchains to perform these lints so that we control the time at which lints change (i.e. when we update the SDK).

>      This is why we should be linting with the SDK tool chains instead of locally installed tool chains. Linting rules should only change when we update the SDK.

_Originally posted by @webern in https://github.com/bottlerocket-os/bottlerocket/issues/2701#issuecomment-1372344522_

mjsterckx:
**Issue number:**

Closes #2704 

**Description of changes:**

Changed the `clippy` used in the `check-clippy` target for Rust linting to the one included in the `BUILDSYS_SDK` to avoid conflicting lints between local builds and GitHub actions. Blocked until https://github.com/bottlerocket-os/bottlerocket-sdk/pull/89 is merged and included in an SDK update.

**Testing done:**

Ran `cargo make check-clippy` with a local `cargo` install which resulted in a few warnings/errors. Ran it again with the SDK version of `clippy` which finished with no problems and exit code 0.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -439,14 +439,21 @@ rc=0
 export VARIANT="${BUILDSYS_VARIANT}"
 
 # For rust first-party source code
-if ! cargo clippy \
-  --manifest-path ${BUILDSYS_TOOLS_DIR}/Cargo.toml \
+if ! docker run --rm \
+   -v "$(pwd)":/tmp \
+   "${BUILDSYS_SDK_IMAGE}" \
+   cargo clippy \
+  --manifest-path /tmp/tools/Cargo.toml \
   --locked -- -D warnings --no-deps; then
   rc=1
 fi
 
-if ! cargo clippy \
-  --manifest-path ${BUILDSYS_SOURCES_DIR}/Cargo.toml \
+if ! docker run --rm \
+   -v "$(pwd)":/tmp \
+   -e VARIANT \
+   "${BUILDSYS_SDK_IMAGE}" \
+   cargo clippy \
+  --manifest-path /tmp/sources/Cargo.toml \
   --locked -- -D warnings --no-deps; then
   rc=1
 fi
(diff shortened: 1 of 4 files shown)


jpculp:
I'm not sure how often it would actually get used, but do we want to add an override based on an environment variable to use the local clippy? It might be nice for developers that want to ensure this workflow will pass with a newer toolchain than the SDK provides.

mjsterckx:
I personally don't think it would get used enough to warrant an override. Developers who want to check their syntax with a newer toolchain can just run `cargo clippy`. @bcressey @webern any thoughts?

webern:
> I personally don't think it would get used enough to warrant an override. Developers who want to check their syntax with a newer toolchain can just run `cargo clippy`. @bcressey @webern any thoughts?

I tend to agree. I think it's easy enough to go into the `sources` or `tools` directory and run `cargo clippy` by hand rather than using an override. (That's what I would probably do...)

bcressey (approved):

bcressey:
Doesn't have to happen in this PR, but do we also want to give `cargo fmt` the same treatment, in case the behavior changes across versions?

mjsterckx:
> Doesn't have to happen in this PR, but do we also want to give `cargo fmt` the same treatment, in case the behavior changes across versions?

I'm not sure how frequently the behavior changes, but I like the idea of staying consistent. It should be an easy addition, so I can create an issue for that and work on it in a different PR.

jpculp (approved):

PR #2799, opened 2023-02-11
testsys: Provide advanced config control

ecpullen:
**Issue number:**

N/A

**Description of changes:**

When developing TestSys, the simplicity of `cargo make test` calls can make it difficult to force certain behavior to occur. It is also possible that the default values are not the best for a user. This pr introduces `DeveloperConfig` for additional control.

`DeveloperConfig` is intended to contain advanced configuration values that most users shouldn't need to care about. There is currently a 1:1 mapping for `GenericVariantConfig : Command Line Env Variables` that I'd like to preserve which is why I created a new type.

Currently there are 4 configuration values:
* `cluster-destruction-policy` allows users to control the destruction policy of cluster crds. The defaults currently used were carefully selected to improve the ux, but some users may want to alter this behavior
* `bottlerocket-destruction-policy` allows users to control the destruction policy of Bottlerocket crds.
* `keep-tests-running` is used to set the `keep_running` field on tests. With `keep_running: true`, the test's pods are kept in a running state until the test is deleted, as a result, the default behavior has been set to `false`, but can be enabled if a user wishes to.
* `image-account-id` is used to set an alternate account as the one starting AMIs are stored in. This is useful for storing AMIs in a separate account from the one testing is occurring in. Without `image-account-id`, a user needs to find the ami id from the account and set `TESTSYS_STARTING_IMAGE_ID`. With this new variable, all manual steps are removed.

**Testing done:**

Tested each new value and verified that the expected values were added to the CRDs.

**Terms of contribution:**

### DIFF
--- a/tools/testsys-config/src/lib.rs
+++ b/tools/testsys-config/src/lib.rs
@@ -4,7 +4,7 @@ use handlebars::Handlebars;
 use log::{debug, trace, warn};
 use maplit::btreemap;
 use model::constants::TESTSYS_VERSION;
-use model::SecretName;
+use model::{DestructionPolicy, SecretName};
 use serde::{Deserialize, Serialize};
 use snafu::ResultExt;
 use std::collections::{BTreeMap, HashMap};
@@ -261,6 +261,8 @@ pub struct GenericVariantConfig {
     pub control_plane_endpoint: Option<String>,
     /// The path to userdata that should be used for Bottlerocket launch
     pub userdata: Option<String>,
+    #[serde(default)]
+    pub dev: DeveloperConfig,
 }
 
 impl GenericVariantConfig {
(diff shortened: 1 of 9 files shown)


stmcginnis (changes requested):
Can we get a small section added to `TESTING.md` that documents this?

ecpullen:
> Can we get a small section added to `TESTING.md` that documents this?

The `dev` configurations were purposely left out of `TESTING.md` to keep it from getting too long and keep it simple. I prefer to keep these configuration values documented in the code base instead to prevent accidental usage.

etungsten (approved):

stmcginnis (approved):
I would prefer to see some documentation somewhere other than the code, but makes sense. Everything else looks good to me.

ecpullen:
Added a note in `Test.toml.example` for the new fields.

stmcginnis (approved):
Awesome, thanks @ecpullen!

PR #2786, opened 2023-02-08
Removes k8s.gcr.io in favor of registry.k8s.io

jpmcb:
**Issue number:**

N/a - `k8s.gcr.io` is being deprecated in favor of `registry.k8s.io`.
See: https://kubernetes.io/blog/2023/02/06/k8s-gcr-io-freeze-announcement/

**Description of changes:**

Replaces registry for pause image. Image is present:

```
❯ crane ls registry.k8s.io/pause
0.8.0
1.0
2.0
3.0
3.1
3.2
3.3
3.4.1
3.5
3.6
3.7
3.8
3.9
go
latest
sha256-7031c1b283388d2c2e09b57badb803c05ebed362dc88d84b480cc47f72a21097.sig
sha256-9001185023633d17a2f98ff69b6ff2615b8ea02a825adffa40422f51dfdcde9d.sig
test
test2
```

**Testing done:**

What testing should we do here since this only affects defaults for vmware k8s?

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #2606, opened 2022-11-22
Archive migrations before v1.12.0

mjsterckx:
**Description of changes:**

Archiving migrations improves build speed. Following the example of https://github.com/bottlerocket-os/bottlerocket/pull/2518.

**Testing done:**

Before this change, the contents of the migrations rpm shows:
```
ubuntu@ip-172-31-65-200:~/dev/bottlerocket/build/rpms$ rpm -qlp bottlerocket-x86_64-migrations-0.0-0.x86_64.rpm
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.10.1_container-runtime
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.10.1_container-runtime-metadata
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_aws-admin-container-v0-9-3
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_aws-config-settings
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_aws-control-container-v0-6-4
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_aws-creds
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_aws-creds-metadata
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_credential-providers
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_ecs-additional-configurations
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_kubelet-new-config-files
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_kubelet-tls-config
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_kubelet-tls-files
/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/migrations/migrate_v1.11.0_public-admin-container-v0-9-3
[...]

PR #2800, opened 2023-02-13
Archive `v1.12.0` migrations in preparation for `v1.13.0` migrations

jpmcb:
**Issue number:**

Closes - N/a

Similar to https://github.com/bottlerocket-os/bottlerocket/pull/2606 and https://github.com/bottlerocket-os/bottlerocket/pull/2518

**Description of changes:**

Archives v1.12.0 migrations in preperation for v1.13.0 migrations (like https://github.com/bottlerocket-os/bottlerocket/pull/2786)

**Testing done:**

Successfully built locally

**Terms of contribution:**


jpmcb:
Marking as draft - we might want to keep some of the `n-1`, `n-2`, etc. migrations around in case we need to make a "pivot" migration repo that skips versions. Will need to open this up for more discussion before this merges.

webern (changes requested):
I think we should keep migrations around through a few versions in case we need to build development repos that skip versions. During a development workflow recently I wanted to do upgrade downgrade between v1.10.1 and v1.12.0 and needed to build the intervening v1.11.0 tag because migrations had been quickly archived. If the migrations stuck around a little longer, the intervening build would not have been necessary.

etungsten:
> I think we should keep migrations around through a few versions in case we need to build development repos that skip versions. During a development workflow recently I wanted to do upgrade downgrade between v1.10.1 and v1.12.0 and needed to build the intervening v1.11.0 tag because migrations had been quickly archived. If the migrations stuck around a little longer, the intervening build would not have been necessary.

I think in those cases, we can do it in a development branch that reverts this change. Archiving migrations does provide a pretty nice build time improvement and also keeps the number of builds artifacts that we need to sync to TUF repositories down.

jpmcb:
> I think we should keep migrations around through a few versions in case we need to build development repos that skip versions. 

👍🏼 

Closing.

PR #2806, opened 2023-02-15
prairiedog: allow '=' in bootconfig values

bcressey:
**Issue number:**
N/A

**Description of changes:**
'=' is allowed in values, so constrain the split accordingly.

This fixes inputs like a `dm-mod.create` parameter that refers to the target block device with the `PARTUUID=.../PARTNROFF=#` syntax.

**Testing done:**
I found this while trying to use a corrupted `dm-mod.create` parameter in `settings.boot`, which caused a boot failure when `prairiedog` attempted to parse `/proc/bootconfig`.

I've confirmed that this fix solves that problem and that the new unit test passes.

**Terms of contribution:**


etungsten (approved):

arnaldo2792 (approved):

markusboehme (approved):
Nice catch!

PR #2811, opened 2023-02-16
add Nutanix AHV platform variant

tuxtof:
**Description of changes:**

implement new variants to support Nutanix AHV platform
update early-boot-config to support AHV userdata/metadata default provisioning mechanism

**Testing done:**

- unit test OK
- building of each variants OK
- standalone k8s OK
- EKS-A integration OK

**Terms of contribution:**


variants/nutanix-k8s-1.21/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+    "console=tty0",
+    # Only reserve if there are at least 2GB
+    "crashkernel=2G-:256M",
+    "net.ifnames=0",
+    "netdog.default-interface=eth0:dhcp4,dhcp6?",
+]
+included-packages = [
+    "cni",
+    "cni-plugins",
+    "kernel-5.15",
```
stmcginnis: Are these kernel version changes intentional for 1.21 through 1.23? Or a copy/paste issue. The kernel config updates above make me think it was an accident and keeping with the other variants version should be fine?

stmcginnis (commented):
Thanks for this PR @tuxtof, it looks like you have been able to figure out a lot of the internals of how these things are put together. Nice work!

I wonder if we can avoid the need for new variants and go a slightly different direction. A lot of the changes here are relatively minor in the end. Perhaps we can make these changes to the existing `metal-*` variants. One of the hopes for that variant was it would end up being a sort of catchall for different platforms. It does not directly translate to a virtualized host, but I think we can get pretty close. And if this is done as part of the metal variants, we could be more confident that issues will be caught in the existing release process.

The kernel module changes could be done there without any obvious harm. Enabling something like `CONFIG_SCSI_VIRTIO` wouldn’t be needed on a normal bare metal host, but it also shouldn’t cause any problems.

For the kernel-parameter settings, I think you could [use boot config](https://github.com/bottlerocket-os/bottlerocket/blob/develop/PROVISIONING-METAL.md#boot-configuration) to get what we need.

The early boot config for metal could be expanded too. There is some file handling with the cdrom path in the [VMware early boot config handling](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/early-boot-config/src/provider/vmware.rs#L25). That could be made more generic so all variants could use it. I could see it being useful to handle the full config drive paths and [json format](https://docs.openstack.org/nova/latest/user/metadata.html#openstack-format-metadata) for other uses too. Then if someone would like to explore running on something like OpenStack, there would at least be one existing piece there.

The only other main gotcha I can see is the qcow2 format. Would it possible to take the existing metal variant build output and use qemu-img to convert it (e.g. `qemu-img convert -f raw -O qcow2 image.img image.qcow2`)?

tuxtof:
Hello @stmcginnis 
Happy to learn that I have correctly follow the internals .

Concerning the variant decision, if you are OK we would prefer stay with a dedicated variant and this for multiple reasons, this first PR is only the beginning:
- we will soon bring a new different metatada system and we would like to integrate it in the early boot config stage
- we are thinking to integrate our Nutanix Guest Tools inside the image
- we plan also to check how we can provide prerequisites dependencies for our CSI drivers
- native qcow2 support is a bonus
- we want to be propose our customers to directly install images coming from your tuftool process
- and probably more to come ...

variants/nutanix-k8s-1.21/Cargo.toml
tuxtof: I will fix this one , copy/paste error

tuxtof: done

stmcginnis:
Had some offline discussions with Nutanix - capturing the results of that here.

For the time being, it has been decided not to add a new variant for Nutanix. The work here was excellent, but it doesn't align with current support requirements and plans. That may change in the future, so it's possible we will reopen this and proceed. But for now, closing this PR and will revisit when necessary.

Thanks for a great work! This should be a very good reference for anyone else looking at creating their own variants.

PR #2832, opened 2023-02-24
systemd: Add patch to split networkd tmpfiles into a separate file

zmrow:
**Issue number:**
Related to #2741 and #2449 

**Description of changes:**
It appears I missed some files when I made the `networkd` subpackage in `systemd`.  These tmpfiles should only be included if the `networkd` subpackage is included.  [An upstream patch](https://github.com/systemd/systemd-stable/commit/eb1446f8f155e2253c9cc7068b29dcdbb6d68aa6) moves these entries into a separate file for situations such as this, so we include that patch.

The offending journal entries that tipped us off to this mistake looked like:
```
Feb 24 18:36:39 localhost systemd-tmpfiles[1562]: /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/tmpfiles.d
/systemd.conf:18: Failed to resolve user 'systemd-network': No such process
```
```
    systemd: Add patch to split networkd tmpfiles into a separate file
    
    This commit adds an upstream patch that moves systemd-networkd tmpfiles
    entries into a separate file.  This makes it simpler to include/exclude
    these tmpfiles depending on the inclusion of the networkd subpackage.
    
    The systemd.spec has been updated to exclude this new file, while the
    networkd subpackage explicitly includes it
```

**Testing done:**
* Build and boot an `aws-k8s-1.24` variant and ensure no messages referring to `systemd-network`'s tmpfiles show up in the journal.
```
bash-5.1# journalctl | grep -i systemd-network                                                                                                   
bash-5.1# journalctl | grep -i networkd       
bash-5.1#
```
* Build and boot an `aws-dev` variant with the `systemd-networkd` image-feature enabled (and add a dependency on `console-getty.service` in `preconfigured.target` since I knew early boot services will fail).  Observe the `systemd-networkd` tmpfiles being created properly and no other networkd-related issues in the journal:
```
bash-5.1# ls -lah /run/systemd/netif/
total 4.0K
drwxr-xr-x.  5 systemd-network systemd-network 120 Feb 24 21:32 .
drwxr-xr-x. 20 root            root            540 Feb 24 21:32 ..
drwxr-xr-x.  2 systemd-network systemd-network  40 Feb 24 21:32 leases
drwxr-xr-x.  2 systemd-network systemd-network  80 Feb 24 21:32 links
drwxr-xr-x.  2 systemd-network systemd-network  40 Feb 24 21:32 lldp
-rw-r--r--.  1 systemd-network systemd-network 135 Feb 24 21:32 state

bash-5.1# journalctl | grep -i systemd-network
Feb 24 21:32:34 localhost systemd-sysusers[315]: Creating group 'systemd-network' with GID 981.
Feb 24 21:32:34 localhost systemd-sysusers[315]: Creating user 'systemd-network' (systemd Network Management) with UID 981 and GID 981.
Feb 24 21:32:35 localhost systemd-tmpfiles[894]: Reading config file "/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/tmpfiles.d/systemd-network.conf"…
...
```

**Terms of contribution:**

### DIFF
--- a/packages/systemd/9012-tmpfiles-Split-networkd-entries-into-a-separate-file.patch
+++ b/packages/systemd/9012-tmpfiles-Split-networkd-entries-into-a-separate-file.patch
@@ -1,0 +1,67 @@
+From abdd268ab3c16c606a1578e5d40d5847a1d99523 Mon Sep 17 00:00:00 2001
+From: Daan De Meyer <daan.j.demeyer@gmail.com>
+Date: Wed, 27 Apr 2022 10:25:22 +0100
+Subject: [PATCH] tmpfiles: Split networkd entries into a separate file
+
+Many distributions ship systemd-networkd as a separate file so we
+need to be able to ship the tmpfiles networkd entries as part of
+that separate networkd package. Let's split the networkd entries
+into a separate file to make that possible.
+---
+ tmpfiles.d/meson.build          |  1 +
+ tmpfiles.d/systemd-network.conf | 13 +++++++++++++
+ tmpfiles.d/systemd.conf.in      |  6 ------
+ 3 files changed, 14 insertions(+), 6 deletions(-)
+ create mode 100644 tmpfiles.d/systemd-network.conf
+
+diff --git a/tmpfiles.d/meson.build b/tmpfiles.d/meson.build
+index b8d3919025..7c2604cfe1 100644
+--- a/tmpfiles.d/meson.build
++++ b/tmpfiles.d/meson.build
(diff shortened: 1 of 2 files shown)


stmcginnis (approved):
Looks good to me!

arnaldo2792 (approved):

yeazelm (approved):

bcressey:
I'm OK with a patch for this but we could also just include `sysusers.d/systemd-network.conf` in the main package so that the user gets created.

Issue #2669, opened 2022-12-16
Enable out-of-tree Bottlerocket variants


cbgbt:
**What I’d Like:**
A supported mechanism for creating custom Bottlerocket variants that live within their own git trees. These variants should support some mechanism for defining a dependency on Bottlerocket’s core in a way that allows the custom variant to evolve independently and consume Bottlerocket-core updates safely.

We've been calling the project to enable this "Out of Tree Builds" and shortening that to OOTB, so if you see references to either, they are a reference to this issue.

**Background:**
Currently a set of variants are maintained in the Bottlerocket tree, but any additional variants must be created by forking the repo and maintaining a series of patches which add the special variant. Some risks and challenges of this include:

* The custom variant repo must frequently rebase a substantial volume of code to stay atop Bottlerocket changes. These rebases are unpredictable and likely non-trivial in many cases.
* The integration points of custom variants are not well-defined. For example, changes to the semantic meaning of settings could impact out-of-tree variants in ways that are difficult to discern during rebase.
* The forked repo will include a significant amount of code/documentation/information which is not relevant to the specific changes needed for the custom variant.

**Edit:** We have started a new GitHub repository for a tool that will drive these builds. Track https://github.com/bottlerocket-os/twoliter/issues/1 for the design.

grosser:
similarly: if we need 1 custom package we also have to fork atm, so ideally the variant workflow would address this issue too
for example having a "build" repo that uses bottlerocket as a git-submodule and all the customization stays outside

jfbette:
Coming back to this functionnality. On my side it is a must-have capability and seems complex for now to make custom Bottlerock variants.

On my side the identified customization needed could be listed as following : 
- add/remove packages
- remove some functionnalities that are not really core system but more linked to AWS ecosystem.
- ability switch to other components doing same functionnality. E.g. switch from grub to direct EFI boot (this is an example, I know the impact on update process).
- specific compilation options (kernel, ...) or apply specific patches
- customize trademarks

For exemple I tried to create a new variant but even in the name of the variant their are some hard coded reference/tests to the name itself. For example, in early-boot-config, it tests if it is a metal variant and then apply some configuration. Being a metal should not be related to the name of the variant but rather a configuration variants

webern:
[...]

PR #2834, opened 2023-02-27
docs: add design doc for settings extensions

cbgbt:
**Issue number:**
#2669

**Description of changes:**
This adds a design document detailing the changes that are being made to Bottlerockets' settings system in order to support out-of-tree variant builds. The user interface for settings via `apiclient` should remain the same, but this outlines how developers will interact with the settings system for in-tree and out-of-tree variants.

This document will be moved under `sources/apiserver` as documentation after it is implemented.

**Terms of contribution:**

### DIFF
--- a/docs/design/settings/extensions/modular-settings-extensions.md
+++ b/docs/design/settings/extensions/modular-settings-extensions.md
@@ -1,0 +1,252 @@
+# Out-of-Tree Builds and Modular Settings Extensions
+## Overview
+The settings API is a fundamental aspect of Bottlerocket as a Linux distribution.
+Therefore, as a design requirement for out-of-tree Bottlerocket builds, variants must be able to extend Bottlerocket’s settings system with custom settings.
+Rather than build a separate system for handling out-of-tree settings, Bottlerocket will move all settings to a modular system -- each setting will have a module installed on the host at runtime.
+These modules will be called "Settings Extensions".
+Extensions will allow the same ergonomics and features of settings in current Bottlerocket, including the ability to migrate data between versions, render data into service configuration files, and interact with system service restarts
+
+This document provides an overview of the overhauled settings system, including details about how settings data is stored and communicated between Bottlerocket components, and how settings extensions are developed.
+Changes related to how settings extensions are packaged via RPM will be discussed in a separate document.
+
+## Requirements
+In an out-of-tree builds world, settings extensions from multiple different sources (Bottlerocket's core, settings defined by other variants, or your own settings) must all work harmoniously in a built system.
+This makes it crucial that all interactions with settings resources adhere to versioned APIs, and that dependencies between settings extensions must be modeled in the build system in a similar way to dependencies on dynamically linked libraries.
+Behaviorally, settings changes in other repositories should not cause breaking changes to other variant source trees, since these would constitute an API change and be vended as a new version.
+Most of the more specific requirements of the system fall out of these principles:
+
+* The settings API must support extensions through a series of stable, versioned, documented APIs that can be implemented by binaries and configurations maintained out-of-tree.
+* Settings behavior must be determined dynamically at runtime based on the set of installed settings extensions.
+* All autonomous interactions with the settings API must explicitly state the settings extension being interacted with, and at what version.
+* Migrations between settings representations as we know them today must cease to exist -- settings extensions must independently expose settings at supported interface versions and implement data migrations between them.
+* Settings extension interfaces, services, and configurations must all be represented in the build system, with the ability to express dependencies on these artifacts.
+
+## Settings Extensions
+This document provides an overview on the following:
+* Storage
+* Validation
+* Retrieval
+* Model migrations
+* System configuration via templating
+* Defaults generation
+* Extensions development with the Settings SDK
+
+### Setting Representation
+Today, Bottlerocket's `apiserver` utilizes models to perform strict validation of settings inputs, which are then serialized to JSON and stored in Bottlerocket's datastore.
+The models are defined using Rust structures and are statically compiled into the operating system, using Rust's type system and compiler to subsequently check interactions with these modelled objects.
+
+In principle, this will remain true -- Bottlerocket will continue to own settings objects which it stores as JSON, and changes will be validated against models on update.
+A primary difference here is that the settings objects will be opaque to Bottlerocket's core.
+Bottlerocket will understand that settings are JSON objects and be able to make modifications requested when, for example, calling `apiclient set`; however, the changes are then submitted to the settings extensions installed on the system in order to validate the proposed changes.
+
+### Extension Binaries
+All settings in a Bottlerocket host will be described by their own settings extension, which are defined by a binary packaged onto the system and symlinked into a common directory, as well as a configuration file placed in a corresponding `settings.d`-style directory.
+Extension binaries must respond to command line arguments following a specified protocol spoken by the Bottlerocket `apiserver`.
+The protocol is versioned, so all commands to an extension binary begin with the protocol version (in this document, all commands fall under the initial `proto1` protocol version -- so-named to avoid conflation with settings versions, which will be explained later).
+Command line arguments are used to signify the "request" being made of the extension.
+Exit codes will be used to signal the status of the request, with formatted output being delivered on `stdout`, and logs delivered on `stderr`.
+
+As an example, suppose a user wants to set the "message of the day" for their instance:
+```bash
+$ apiclient set motd="Hello, Bottlerocket!"
+```
+
+The apiserver would load the current value of `motd` and then create a new object based on the inputted string.
+It would then find the extension responsible for `motd` and make a request to determine whether or not the change should proceed:
+
+```bash
+$ motd proto1 set --setting-version v1 --value '"Hello, Bottlerocket!"'  # Values are passed as JSON, hence the quoting
+```
+
+### Extension Naming & Setting Ownership
+Every top-level setting in Bottlerocket's API must be owned by a singular extension, and that extension will share a name with the setting.
+This name will also be shared by the extension's configuration file and binary components.
+This means that whenever the system must interact with a setting (be it via user action on the API, or rendering templates to be used as configuration files), Bottlerocket can identify the setting extension to invoke in order to satisfy that interaction.
+
+As an example, the following suggests the filesystem layout for a settings extension for the `settings.network` settings:
+
+```
+usr
+└── lib
+    └── bottlerocket
+        └── settings
+            ├── config.d
+            │   └── network.toml
+            └── extensions.d
+                └── network -> /usr/bin/network-extension
+```
+
+#### Transactional Writes and Cross-Validation
+Bottlerocket today is missing a mechanism for cross-validation of settings.
+As an example, suppose we want to represent a range of integers in settings by allowing a `setting.min-value` and `setting.max-value` to be set -- it should be possible for settings extensions to set both simultaneously, while also verifying that the state from setting both is valid (`setting.min-value <= setting.max-value`).
+
+Let us cast this requirement in a different light: imagine an OS administrator wishes to restrict what values can be written to an existing setting -- for example, they may wish to only allow settings values to `network.hosts` so long as they include as a subset the entries mandated by their organization.
+To allow this, settings extensions will be capable of registering themselves as a *validator* for any other setting on the host, including settings that they do not own.
+When handling write requests for multiple settings in a single transaction, the `apiserver` will first gather the resulting writes that would occur under each settings extension involved in the writes.
+Then the `apiserver` will provide the provisional settings state to each settings extension which has been registered as a validator of any of the target settings.
+Any one settings extension has the power to halt the transaction by returning a non-zero exit-code on validation.
+
+Extensions must specify the setting which triggers the validation in their configuration file:
+
+```toml
+# auditor.toml -- setting extension config file for `settings.auditor`
+[extension.validates]
+network = "v1"
+```
+
+It is technically valid for an extension to specify that it validates a setting which does not exist; however, such a specification can never be triggered.
+In the case that the setting exists but the specified version does not, all validations will result in a failure.
+Settings extensions should model their dependency requirements (including validations) in RPM so to avoid cases where the resulting system is missing settings extensions with the appropriate version.
+
+#### Datastore Layout
+Once the `apiserver` has a target set of data to write for a given `set` request and it has been validated by all required validators, it must persist the new settings to the datastore.
+Much like the current datastore, stored settings are stored to the filesystem via a particular pattern.
+
+```bash
+datastore/
+└── motd
+    ├── v1
+    │   └── motd.json
+    └── v2
+        └── motd.json
+```
+Any data written to directories created during a `set` transaction is moved into the datastore.
+Despite only one version of the data being provided, the `apiserver` will populate all versions supported by the settings extension at this time by requesting that the settings extension perform all necessary migrations.
+Failure to populate a version listed as "supported" will result in a failed transaction.
+See the next section about migrations for more details.
+
+### Extension Versioning and Migrations
+Bottlerocket's settings data model changes frequently: sometimes to support new features, or sometimes to correct mistaken model shapes.
+In order to support changes to the settings model, settings must be capable of exposing versioning information about the settings extension, as well as providing faculties for migrating settings data to new versions.
+Our versioning scheme will take heavy influence from Kubernetes' scheme for [versioning CustomResourceDefinitions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/).
+Settings extension versioning information is surfaced through the extension config file:
+
+```toml
+# hostname.toml -- setting extension config file for settings.hostname
+[extension]
+supported-versions = [
+     "v1",
+     "v2"
+]
+default-version = "v2"
+```
+
+Any setting can be written or read at any of the `supported-versions` listed by the extension.
+Unversioned requests to read or write this setting are assumed to be the `default-version` and should only take place when a user makes a request using the CLI -- all automated process *should* refer to an explicit version.
+During a `set` transaction, the `apiserver` will consider the provided version to be the `canonical-version`, and will ask the settings extension to perform migrations from the `canonical` version to all supported versions using the `migrate` command.
+These migrations always begin at the `canonical` version, and settings extensions are free to perform the migrations however they wish; however, Bottlerocket's Settings SDK will provide utilities to help settings developers support a series of linear setting upgrades and downgrades.
+Read on for more information about the Settings SDK.
+
+All migrations are performed and stored at setting-write-time because Bottlerocket rollbacks may result in running an OS with a settings extension which does not support the version at which a given setting is stored.
+If Bottlerocket boots and finds that a setting is stored at a value which is not supported by its installed setting extension, it will allow that version to persist unless a new value is `set` by the `apiserver`, at which point it is removed -- this is to attempt to prevent data loss in the case that the image can eventually update to a new image which supports the stored setting version.
+
+### Configuration Templating
+Once Bottlerocket settings are stored in the datastore, they must somehow then be used to influence the behavior of the system.
+Fetching settings from the Bottlerocket API/datastore is currently done via two mechanisms:
+* Fetching data directly from the API using `apiclient`
+* Rendering config files (or just strings) using `schnauzer`
+
+For open-source tools, Bottlerocket prefers the rendered config file approach, as it doesn't require patching upstream code in order to consume the Bottlerocket API.
+First-party tools have no such constraints, and so today there exists  tools that take either approach.
+
+Because settings are discovered at runtime on the system, and indeed may no longer be handled by models written in Rust, it will no longer be feasible to create an `apiclient` with the same strong typing as exists today.
+It will instead be preferred that system services, including first-party services, all route their fetching of API settings through their own configuration files, which will be rendered by Bottlerocket whenever the related settings are changed.
+
+Today, configuration templates are rendered by `schnauzer` using [handlebars](https://handlebarsjs.com/) templates.
+`schnauzer` fetches the entirety of the settings tree, defines some useful helper functions, and then uses these as context for rendering the template.
+With the introduction of settings extensions, this will change such that templates specify `(name, version)` tuples for all extensions needed to render the template.
+These extension specifications will be given within the template as "front matter" -- a TOML block specified at the head of the document and delineated with `+++` as such:
+
+```toml
++++
+[required-extensions]
+hostname = v1
+kubernetes = v1
++++
+# Below here the template is written as usual
+```
+
+This will allow Bottlerocket to find the owning extensions for the data needed to render the template.
+It also tightly binds the template to a given implementation of the requisite settings, making it much more difficult to accidentally break "downstream" templates with "upstream" changes.
+
+`schnauzer` can also continue to be invoked as a standalone binary without the need for a templated configuration file.
+In those circumstances, you are still required to list your extension dependencies and their version, but these can be done as command line arguments:
+
+```bash
+$ schnauzer --required-extension 'hostname=v1' '{{ network.hostname }}'
+```
+
+#### Extending Handlebars Helpers
+Handlebars has a few utilities for applying simple processes to data in order to format it appropriately for your document; however, you often need to implement your own "helpers" for more complex text formatting.
+This is currently done in Bottlerocket by adding helper implementations to the `schnauzer` tool directly.
+Bottlerocket settings extensions will expose versioned handlebar helpers:
+
+```bash
+# hostname.toml setting extension config file
+[templating.v1] # template helpers are versioned
+helpers = [
+    "localhost_aliases"
+]
+
+
+$ ./hostname proto1 template-helper --version v1 --helper localhost_aliases 'arg1' 'arg2'
+#  rendered output returned as a string on stdout
+```
+
+Within the handlerbars template, the helper name will be prepended with `$EXTENSION_NAME.` to avoid collisions with other settings extensions.
+As an example, to use the `localhost_aliases` helper above from a template, your template may look something like this:
+
+```
+{{ hostname.localhost_aliases(arg1, arg2) }}
+```
+
+### Settings Generation
+Some settings have default values which cannot be known statically and must be computed after the system is running.
+Bottlerocket settings currently have the ability to compute a default value via something called a "settings generator".
+During boot, settings extensions are given a chance to populate settings with default values that are then persisted, effectively determining the setting value for the lifetime of that instance.
+
+Settings generation often depends on the presence of other settings.
+It's also possible for settings to have interdependencies on generation order, for example:
+
+* `foo.bar` depends on `baz.bot`
+* `boz.blop` depends on `foo.bar`
+
+In order to resolve this, settings generators are allowed to require that they be provided whatever data is currently applied to dependent settings.
+The generators are "re-entrant" in that they return a status dictating whether or not they are finished generating, or if they need more data to proceed.
+Bottlerocket will iteratively invoke generators until all generation has completed, or in cases where it seems likely that a deadlock has occurred.
+
+```toml
+# setting.toml requires the network and kubernetes settings before it can emit its own
+[generation.required]
+network = v1
+kubernetes = v2
+```
+
+## Settings Extension Development and the Settings SDK
+Settings extensions can be written in any programming language, so long as a symlink to a binary speaking the "extension protocol" lands in the right spot on the filesystem.
+That said, many settings extensions need to tackle the same challenges: How do we represent and implement settings versions? Migrations? Validation? Template utilities? Rather than forcing all extensions to implement this in isolation, we will create a shared library, called the "Settings SDK", which will be published to crates.io.
+
+The goal of the settings SDK is to provide standard functions and macros for interfacing with the settings extension protocol.
+While our model here provides quite a lot of power to settings extension developers, the SDK would allow us to exert some pressure in the form of opinions on simple ways to manage data or perform migrations.
+
+As an example, the SDK will provide a Rust trait which can be implemented for each model in order to adequately implement all settings extensions methods on a command line interface.
+Once those traits are implemented, you might implement this CLI as such:
+
+```rust
+fn main() -> Result<()> {
+    bottlerocket_settings_sdk::SettingsExtension::with_models(vec![
+        BottlerocketSetting::<v1::MotdV1>::model(),
+        BottlerocketSetting::<v2::MotdV2>::model(),
+    ])
+    .run_extension()
+    .context("Settings extension encountered an error.")
+}
+```
+
+## Asked and Anticipated Questions
+### Why not use some existing schema or modeling language to define the shape of settings?
+Existing modeling languages are powerful, but even in our existing Bottlerocket settings we have validations which we have implemented as functions in the Rust code (or otherwise wished we had a simple way to do so.) Rather than strip power from settings extensions by dictating that they use one of these modeling languages, we leave it entirely up to the binary to decide how best to validate inputs.
+In many cases, this may mean that the binaries carry with them their own implementation of a popular schema language, like JSONschema.
+
+### How will the `apiserver` handle concurrent requests in the face of transactional writes?
+The `apiserver` will use a [filesystem-based](https://man7.org/linux/man-pages/man2/flock.2.html) reader-writer lock to ensure that the datastore is only accessed by concurrent readers or a single writer. Writes which are sent to the `apiserver` while the file is lock are placed in a queue.
--- a/docs/design/settings/extensions/settings-extension-example.md
+++ b/docs/design/settings/extensions/settings-extension-example.md
@@ -1,0 +1,159 @@
+# Settings Extensions Example
+This document walks through a typical use-case for Bottlerocket's settings extension system.
+Technical details of the design of the settings extension system are described elsewhere.
+
+## Background
+Suppose you are creating a Bottlerocket variant which adds some new daemon software that you've created.
+The daemon is configured via a YAML configuration file with a web address.
+When the daemon starts, it parses its configuration file for the target address and then begins periodically querying that address for updates.
+
+You'd like the following behavior to be accomplished:
+* You can configure the daemon using the Bottlerocket settings API, the value will be stored in `settings.mydaemon.query-url`.
+* Whenever the daemon is reconfigured using the Bottlerocket API, it is automatically restarted with the new configuration.
+
+In order to do this, we need to create the following resources in our Bottlerocket variant:
+* A setting extension crate called `mydaemon-settings` to add the new value `settings.mydaemon.query-url` to the API.
+* A `template` which will be rendered using the setting value and used as a configuration for the daemon.
+* A `service` file, which tells Bottlerocket how to handle the lifecycle of our daemon process as its settings change.
+
+## Creating the Settings Extension
+The easiest way to create an extension is to create a new Rust package in your project which utilizes the `bottlerocket-settings-sdk` crate.
+
+```
+├── packages
+├── sources
+│   └── mydaemon-settings
+│       ├── Cargo.toml
+│       └── src
+│           └── main.rs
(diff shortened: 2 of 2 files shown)


cbgbt:
I think we should discuss how we'd like to accept pull requests for designs as a project. e.g. I'm partial to how the Rust project organizes RFCs in a [separate repository](https://github.com/rust-lang/rfcs).

Since we don't yet have such a process, I'd like to keep my design docs in the Bottlerocket repository for now. They can be moved later if we determine a better place for them.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,314 @@
...
+            ├── config.d
+            │   └── network.toml
+            └── extensions.d
+                └── network -> /usr/bin/network-extension
+```
+
+### Setting Storage
+Today, Bottlerocket's `apiserver` utilizes models to perform strict validation of settings inputs, which are then serialized to JSON and stored in Bottlerocket's datastore.
+
+Writes to Bottlerocket settings is going to change in many ways under our new paradigm.
```
webern: nit

```suggestion
Writing to Bottlerocket settings is going to change in many ways under our new paradigm.
```

or

Writes to Bottlerocket settings are going to change in many ways under our new paradigm.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,314 @@
...
+Because settings are discovered at runtime on the system, and indeed may no longer be handled by models written in Rust, it will no longer be feasible to create an `apiclient` with the same strong typing as exists today.
+It will instead be preferred that system services, including first-party services, all route their fetching of API settings through their own configuration files, which will be rendered by Bottlerocket whenever the related settings are changed.
+
+Today, configuration templates are rendered by `schnauzer` using [handlebars](https://handlebarsjs.com/) templates.
+`schnauzer` fetches the entirety of the settings tree, defines some useful helper functions, and then uses these as context for rendering the template.
+With the introduction of settings extensions, this will change such that templates specify `(name, version)` tuples for all extensions needed to render the template.
+These extension specifications will be given within the template as "front matter" -- a TOML block specified at the head of the document and delineated with `---` as such:
+
+```toml
+---
```
stockholmux: I would suggest delimiting front mater with something other than triple dash.

Triple plus (`+++`) is my preference as [Hugo, Forestry, and Zola use it. Some other software uses `---toml`](https://github.com/denoland/deno_std/issues/3094#issuecomment-1400018219) but that is not symmetrical, the ending being triple dash.. yuck. 

Triple dash is likely to confuse people as that alone is widely used for yaml.

docs/design/settings/extensions/settings-extension-example.md
```diff
@@ -0,0 +1,105 @@
...
+Suppose you are creating a Bottlerocket variant which adds some new daemon software that you've created.
+The daemon is configured via a YAML configuration file with a web address.
+When the daemon starts, it parses its configuration file for the target address and then begins periodically querying that address for updates.
+
+You'd like the following behavior to be accomplished:
+* You can configure the daemon using the Bottlerocket settings API, the value will be stored in `settings.mydaemon.query-url`.
+* Whenever the daemon is reconfigured using the Bottlerocket API, it is automatically restarted with the new configuration.
+
+In order to do this, we need to create the following resources in our Bottlerocket variant:
+* A setting extension called `mydaemon` to add the new value `settings.mydaemon.query-url` to the API.
```
webern: Should it be called mydaemon-settings to distinguish it from the mydaemon daemon binary?

```suggestion
* A setting extension called `mydaemon-settings` to add the new value `settings.mydaemon.query-url` to the API.
```

docs/design/settings/extensions/settings-extension-example.md
```diff
@@ -0,0 +1,105 @@
...
+The configuration template includes "frontmatter" which informs the settings system to use settings values owned by the `mydaemon` extension, then the rest of the template is used to render the configuration file.
+
+## Service File
+The settings system needs to know how to restart `mydaemon` so that it can parse changes to its config file after settings have been changed.
+In order to do that,you will create a `mydaemon-service.toml` file.
+This file will be referred to via a symlink on the Bottlerocket filesystem which instructs the settings system to restart the `mydaemon-service` when changes to `settings.mydaemon` take place:
+
+```
+├── packages
+│   └── hello-agent
```
webern:
```suggestion
│   └── mydaemon
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,314 @@
...
+As an example, suppose we want to represent a range of integers in settings by allowing a `setting.min-value` and `setting.max-value` to be set -- it should be possible for settings extensions to set both simultaneously, while also verifying that the state from setting both is valid (`setting.min_value <= setting.max_value`).
+
+Let us cast this requirement in a different light: imagine an OS administrator wishes to restrict what values can be written to an existing setting -- for example, they may wish to only allow settings values to `network.hosts` so long as they include as a subset the entries mandated by their organization.
+To allow this, settings extensions will be capable of register themselves as a *validator* for any other setting on the host, including settings that they do not own.
+When handling write requests for multiple settings in a single transaction, the `apiserver` will first gather the resulting writes that would occur under each settings extension involved in the writes.
+Afterwards, the `apiserver` will provide the provisional settings state to each settings extension which has been registered as a validator of any of the target settings.
```
webern:
```suggestion
Then the `apiserver` will provide the provisional settings state to each settings extension which has been registered as a validator of any of the target settings.
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,314 @@
...
+Any one settings extension has the power to halt the transaction by returning a non-zero exit-code on validation.
+
+Validations must specify the setting which triggers the validation, as well as a list of templates which will be rendered and provided to the validator.
+
+```toml
+# auditor.toml -- setting extension config file for `settings.auditor`
+[[extension.validations]]
+validates = "network.hostname"
```
webern:
```suggestion
validates = [ "network.hostname" ]
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,314 @@
...
+with-templates = [
+    "/templates/auditor-extension/validator-template.toml"
```
webern: I'm not sure I understand how a template is involved in the validation.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,314 @@
...
+supported-versions = [
+     "v1",
+     "v2"
+]
+default-version = "v2"
+```
+
+Any setting can be written or read at any of the `supported-versions` listed by the extension.
+Unversioned requests to read or write this setting are assumed to be the `default-version`.
+During a `set` transaction, the `apiserver` will consider the provided version to be the `canonical-version`, and will ask the settings extension to perform migrations from the `canonical` version to all supported versions using the `migrate` command.
```
webern, line 136: Should `migrate` be rolled into `set`? i.e. should `set` also do the "migrations" since there is no case when `set` would *not* be followed by `migrate`?

docs/design/settings/extensions/modular-settings-extensions.md
webern: I would prefer not to use frontmatter since I'm concerned it will confuse syntax highlighters, linters, etc. Case in point, see the red highlights in this GitHub render. It rather see a separate file with a naming convention, theconfig.meta.toml or theconfig.frontmatter.toml... something like that.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,314 @@
...
+It also tightly binds the template to a given implementation of the requisite settings, making it much more difficult to accidentally break "downstream" templates with "usptream" changes.
+
+`schnauzer` can also continue to be invoked as a setting generator without the need for a templated configuration file.
+In those circumstances, you are still required to list your extension dependencies and their version, but these can be done as command line arguments:
+
+```bash
+$ schnauzer --required-extension 'hostname=v1' '{{ network.hostname }}'
+```
+
+As an optimization, templates which render to `systemd` units will invoke a restart command to systemd
```
webern: Some might not be running yet during the boot sequence.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,314 @@
...
+Handlebars has a few utilities for applying simple processes to data in order to format it appropriately for your document; however, you often need to implement your own "helpers" for more complex text formatting.
+This is currently done in Bottlerocket by adding helper implementations to the `schnauzer` tool directly.
+Bottlerocket settings extensions will expose versioned handlebar helpers:
+
+```bash
+# hostname.toml setting extension config file
+[templating.v1] # template helpers are versioned
+helpers = [
+    "localhost_aliases"
+]
```
webern, line 189: Something seems off here. It looks like there would be multiple versions of the schema by which you can specify which helpers exist.

Maybe this? I guess I'm not super clear on the versioning of helpers.

```suggestion
[templating.helpers] # template helpers are versioned
localhost_aliases = "v1"
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,314 @@
...
+
+```
+{{ hostname.localhost_aliases(arg1, arg2) }}
+```
+
+### Settings Generation
+Bottlerocket settings currently have the ability to compute a default value via something called a "settings generator".
+During boot, settings extensions are given a chance to populate settings with default values that are then persisted, effectively determining the setting value for the lifetime of that instance.
+
+Settings generation often depends on the presence of other settings; therefore the invocation protocol will allow settings generation to accept rendered templates as input:
```
webern: rendered templates instead of other settings? wouldn't be more convenient for me to say I need `aws.region` than to find a template which contains that value?

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: This is by design.

Each existing version of a setting would have its own helpers that it provides. During template rendering, we request a settings at a given version, we need to ensure that the template helpers at that version continue to exist and provide the same API.

In practice, this likely means that you'd specify more or less the same helpers at all versions, but this gives us a valve to move to new versions.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: This design made `aws.region` more opaque from the apiserver's perspective, so it wouldn't be possible for the server to resolve that. It's one of the reasons why we'd need to move system services to use template files. Basically anything that would request a setting would do so via a template.

cbgbt:
After prototyping and a few virtual discussions about this design, we've opted to move to something a bit simpler, opting instead to be more JSON-based at the expense of extension flexibility regarding storage and formatting.

I'll move this to draft until I have the time to update the document to reflect the current direction. For now I'm focusing on fleshing out the prototype a bit more to make sure I catch any other lurking gotchas.

cbgbt:
After spending time [prototyping the Settings SDK](https://github.com/cbgbt/bottlerocket-settings-sdk/tree/0bee8537a08cb015fbaa7ebd769fe98d0d72aaba), considering some of the feedback here, and thinking about how to simplify the settings design, I've significantly reworked this doc to account for my revised thinking.

Brief Changelog:
* Added more explicit requirements and exposition to help better-set the stage for the purpose of these changes.
* Removed RPM-packaging specifications to be discussed in greater depth in a separate document. This is a complex problem, and this doc is already a bit too long.
* Reworked the interfaces for settings extensions to strictly involve passing JSON messages.
* These JSON changes now have the serialized settings data owned entirely by the OS, even if the objects are opaque to it.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+All settings in a Bottlerocket host will be described by their own settings extension, which are defined by a binary packaged onto the system and symlinked into a common directory, as well as a configuration file placed in a corresponding `settings.d`-style directory.
+Extension binaries must respond to command line arguments following a specified protocol spoken by the Bottlerocket `apiserver`.
+The protocol is versioned, so all commands to an extension binary begin with the protocol version (in this document, all commands fall under the initial `proto1` protocol version -- so-named to avoid conflation with settings versions, which will be explained later).
+Command line arguments are used to signify the "request" being made of the extension.
+Exit codes will be used to signal the status of the request, with formatted output being delivered on `stdout`, and logs delivered on `stderr`.
+
+As an example, suppose a user wants to set the "message of the day" for their instance:
+```bash
+$ apiclient set motd="Hello, Bottlerocket!"
+```
```
bcressey, line 52: How will the type inference work? Same sort of approach as [massage_set_input](https://github.com/bottlerocket-os/bottlerocket/blob/d5acb9e7b33121f0bab1f3102b2b8c295d750ed0/sources/api/apiclient/src/main.rs#L588) today?

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+
+Let us cast this requirement in a different light: imagine an OS administrator wishes to restrict what values can be written to an existing setting -- for example, they may wish to only allow settings values to `network.hosts` so long as they include as a subset the entries mandated by their organization.
+To allow this, settings extensions will be capable of registering themselves as a *validator* for any other setting on the host, including settings that they do not own.
+When handling write requests for multiple settings in a single transaction, the `apiserver` will first gather the resulting writes that would occur under each settings extension involved in the writes.
+Then the `apiserver` will provide the provisional settings state to each settings extension which has been registered as a validator of any of the target settings.
+Any one settings extension has the power to halt the transaction by returning a non-zero exit-code on validation.
+
+Validations must specify the setting which triggers the validation in their configuration file:
```
bcressey: nit: is this reading correct?
```suggestion
Extensions must specify the setting which triggers the validation in their configuration file:
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+
+```
+usr
+└── lib
+    └── bottlerocket
+        └── settings
+            ├── config.d
+            │   └── network.toml
+            └── extensions.d
+                └── network -> /usr/bin/network-extension
```
bcressey, line 76: Is the 1:1 match between config file and symlink name enforced, or just a convention?

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+#### Transactional Writes and Cross-Validation
+Bottlerocket today is missing a mechanism for cross-validation of settings.
+As an example, suppose we want to represent a range of integers in settings by allowing a `setting.min-value` and `setting.max-value` to be set -- it should be possible for settings extensions to set both simultaneously, while also verifying that the state from setting both is valid (`setting.min_value <= setting.max_value`).
```
bcressey: Does the switch from `min-value` to `min_value` signify anything?

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+When handling write requests for multiple settings in a single transaction, the `apiserver` will first gather the resulting writes that would occur under each settings extension involved in the writes.
+Then the `apiserver` will provide the provisional settings state to each settings extension which has been registered as a validator of any of the target settings.
+Any one settings extension has the power to halt the transaction by returning a non-zero exit-code on validation.
+
+Validations must specify the setting which triggers the validation in their configuration file:
+
+```toml
+# auditor.toml -- setting extension config file for `settings.auditor`
+[extension.validates]
+network = "v1"
```
bcressey, line 94: Are extensions allowed to validate top-level settings that don't exist? E.g.,
```
[extension.validates]
not-found = "v1"
```

What about versions that the extension doesn't support?

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+[extension]
+supported-versions = [
+     "v1",
+     "v2"
+]
+default-version = "v2"
+```
+
+Any setting can be written or read at any of the `supported-versions` listed by the extension.
+Unversioned requests to read or write this setting are assumed to be the `default-version` and should only take place when a user makes a request using the CLI -- all automated process *should* refer to an explicit version.
```
bcressey, line 135: Unversioned requests will also be used when handling user-data, right?

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+During a `set` transaction, the `apiserver` will consider the provided version to be the `canonical-version`, and will ask the settings extension to perform migrations from the `canonical` version to all supported versions using the `migrate` command.
+These migrations always begin at the `canonical` version, and settings extensions are free to perform the migrations however they wish; however, Bottlerocket's Settings SDK will provide utilities to help settings developers support a series of linear setting upgrades and downgrades.
+Read on for more information about the Settings SDK
```
bcressey:
```suggestion
Read on for more information about the Settings SDK.
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+
+All migrations are performed and stored at setting-write-time because Bottlerocket rollbacks may result in running an OS with a settings extension which does not support the version at which a given setting is stored.
+If Bottlerocket boots and finds that a setting is stored at a value which is not supported by it's installed setting extension, it will allow that version to persist unless a new value is `set` by the `apiserver`, at which point it is removed -- this is to attempt to prevent dataloss in the case that the image can eventually update to a new image which supports the stored setting version.
```
bcressey:
```suggestion
If Bottlerocket boots and finds that a setting is stored at a value which is not supported by its installed setting extension, it will allow that version to persist unless a new value is `set` by the `apiserver`, at which point it is removed -- this is to attempt to prevent data loss in the case that the image can eventually update to a new image which supports the stored setting version.
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+`schnauzer` fetches the entirety of the settings tree, defines some useful helper functions, and then uses these as context for rendering the template.
+With the introduction of settings extensions, this will change such that templates specify `(name, version)` tuples for all extensions needed to render the template.
+These extension specifications will be given within the template as "front matter" -- a TOML block specified at the head of the document and delineated with `+++` as such:
+
+```toml
++++
+[required-extensions]
+hostname = v1
+kubernetes = v1
++++
```
bcressey, line 165: I still have some doubts about the "front matter" approach, but I'm willing to defer judgment until I see the implementation 😀 

I'm still somewhat partial to the filesystem-oriented approach, e.g.:
```
usr/
└── share
    └── templates
        └── etc
            └── kubelet
                ├── kubelet.config
                └── kubelet.config.required.d
                    ├── kubernetes
                    └── network
```

Since this is a bit more consistent with some of the other ways dependencies are modeled, at the cost of giving up tight binding to the template.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+# Below here the template is written as usual
+```
+
+This will allow Bottlerocket to find the owning extensions for the data needed to render the template.
+It also tightly binds the template to a given implementation of the requisite settings, making it much more difficult to accidentally break "downstream" templates with "usptream" changes.
```
bcressey:
```suggestion
It also tightly binds the template to a given implementation of the requisite settings, making it much more difficult to accidentally break "downstream" templates with "upstream" changes.
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+```toml
+# setting.toml requires the network and kubernetes settings before it can emit its own
+[generation.required]
+network = v1
+kubernetes = v2
+```
+
+## Settings Extension Development and the Settings SDK
+Settings extensions can be written in any programming language, so long as a symlink to a binary speaking the "extension protocol" lands in the right spot on the filesystem.
+That said, many settings extensions need to tackle the same challenges: How do we represent and implement settings versions? Migrations? Validation? Template utilities? Rather than forcing all extensions to implement this in isolation, we will create a shared library, called the "Settings SDK", which will be built in Bottlerocket's monorepo and published to crates.io.
```
bcressey: Let's not name this or say what repo it will be built from in this doc; it's not clear that we'd put this crate in the core repo vs. having another dedicated project for it.

I would like to avoid having a wide variety of different artifacts published from a single repo, since that complicates CI/CD workflows and the releases page.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Appendices
+### A. Assisted EC2 Userdata Generation and Validation
+Bottlerocket user issues are often-enough caused by minor misuse of specifying Bottlerocket settings via ec2 Userdata.
+The feedback loop here is relatively slow: you must launch a new Bottlerocket instance and determine whether or not it booted into the appropriate state in order to determine whether or not a given userdata document is correct.
+Indeed, even Bottlerocket core developers wish for tooling to better generate and validate EC2 userdata without the need to interact with the EC2 APIs.
+
+This feature is possible even in a world where settings are defined by extensions which are discovered dynamically.
+During a variant build, we can modify sodapop to extract the settings extension binaries and construct a Docker container which can use the information exposed by those binaries in order to validate userdata, or render a userdata template.
```
bcressey:
```suggestion
During a variant build, we can modify twoliter to extract the settings extension binaries and construct a Docker container which can use the information exposed by those binaries in order to validate userdata, or render a userdata template.
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+## Asked and Anticipated Questions
+### Why not use some existing schema or modeling language to define the shape of settings?
+Existing modeling languages are powerful, but even in our existing Bottlerocket settings we have validations which we have implemented as functions in the Rust code (or otherwise wished we had a simple way to do so.) Rather than strip power from settings extensions by dictating that they use one of these modeling languages, we leave it entirely up to the binary to decide how best to validate inputs.
+In many cases, this may mean that the binaries carry with them their own implementation of a popular schema language, like JSONschema.
+
+### How will the `apiserver` handle concurrent requests in the face of transactional writes?
+The `apiserver` will use a [filesystem-based](https://man7.org/linux/man-pages/man2/flock.2.html) reader-writer lock to ensure that the datastore is only accessed by concurrent readers or a single writer. Writes which are sent to the `apiserver` while the file is lock are placed in a queue.
+
+## Appendices
+### A. Assisted EC2 Userdata Generation and Validation
```
bcressey: Would it be better to tackle this in a follow-up design? I agree that it's a potential benefit, but it seems orthogonal to the main concepts here.

docs/design/settings/extensions/settings-extension-example.md
```diff
@@ -0,0 +1,112 @@
...
+bottlerocket-settings-sdk = 1.0
+...
+```
+
+Implementing the `bottlerocket_settings_sdk::Extension` trait provides a straightforward path to tell Bottlerocket how to validate and store the desired `query-url` setting.
+
+```rust
+#[derive(Serialize, Deserialize, Debug, Default, PartialEq, Clone)]
+struct MyDaemonSettingV1 {
+    query-url: Url,
```
bcressey:
```suggestion
    query_url: Url,
```

docs/design/settings/extensions/settings-extension-example.md
```diff
@@ -0,0 +1,112 @@
...
+mydaemon: v1
+---
+query-url: {{ mydaemon.query-url }}
+```
+The configuration template includes "frontmatter" which informs the settings system to use settings values owned by the `mydaemon` extension, then the rest of the template is used to render the configuration file.
+
+## Service File
+The settings system needs to know how to restart `mydaemon` so that it can parse changes to its config file after settings have been changed.
+In order to do that,you will create a `mydaemon-service.toml` file.
+This file will be referred to via a symlink on the Bottlerocket filesystem which instructs the settings system to restart the `mydaemon-service` when changes to `settings.mydaemon` take place:
```
bcressey: I find it hard to follow these files without a picture of the final install layout - the examples here tell me how to arrange the source files, but not what the destination layout should be, so I can't visualize the resulting spec file.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: Hmm, I hadn't considered this. I think we could resolve this via the same algorithm with the same caveats. It's worth considering if we can come up with a cleaner solution here, but I don't think failing to come up with one introduces a change in behavior.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: It's enforced: This is how the system locates the configuration file for the given extension.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: Whoops! That's a mistake.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: Good questions. I'll update the doc, but the answer to both is no.

At the spec/packaging level, you'd need to specify a dependency on settings that are required. We can add a build-time check to ensure that validated versions are supported.

At runtime either of these will result in an error occurring and the setting failing to be set.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: Yes!

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: Totally fair. It's included here to capture some verbal discussion of this document that occurred and I wanted to capture it as a possible feature of the new system, but it can be delegated to another doc.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+
+$ ./hostname proto1 template-helper --version v1 --helper localhost_aliases 'arg1' 'arg2'
+#  rendered output returned as a string on stdout
+```
+
+Within the handlerbars template, the helper name will be prepended with `$EXTENSION_NAME.` to avoid collisions with other settings extensions.
+As an example, to use the `localhost_aliases` helper above from a template, your template may look something like this:
+
+```
+{{ hostname.localhost_aliases(arg1, arg2) }}
```
zmrow, line 200: Perhaps I missed it, but how are the helpers registered with handlebars?

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+```
+
+### Settings Generation
+Bottlerocket settings currently have the ability to compute a default value via something called a "settings generator".
```
zmrow, line 205: I wouldn't say that computing a default value is settings generator's main purpose.  Their main purpose is to calculate settings at runtime for things that can't be given a static default.

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+The protocol is versioned, so all commands to an extension binary begin with the protocol version (in this document, all commands fall under the initial `proto1` protocol version -- so-named to avoid conflation with settings versions, which will be explained later).
+Command line arguments are used to signify the "request" being made of the extension.
+Exit codes will be used to signal the status of the request, with formatted output being delivered on `stdout`, and logs delivered on `stderr`.
+
+As an example, suppose a user wants to set the "message of the day" for their instance:
+```bash
+$ apiclient set motd="Hello, Bottlerocket!"
+```
+
+The apiserver would load the current value of `motd` and then create a new object based on the inputted string.
```
webern, line 54: I may have already asked this, but why does it need to load the current value before proceeding with the rest of this?

docs/design/settings/extensions/modular-settings-extensions.md
webern, line 94: Would this also be possible (i.e. validating only a subsection of a top-level setting?)

```toml
# auditor.toml -- setting extension config file for `settings.auditor`
[extension.validates]
network.foo.bar = "v1"
```

docs/design/settings/extensions/modular-settings-extensions.md
webern: My take was something like this, where the metadata file is found by naming convention:

```
usr/
└── share
    └── templates
        └── etc
            └── kubelet
                ├── kubelet.config
                └── kubelet.config.info.toml
```

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+hostname = v1
+kubernetes = v1
++++
+# Below here the template is written as usual
+```
+
+This will allow Bottlerocket to find the owning extensions for the data needed to render the template.
+It also tightly binds the template to a given implementation of the requisite settings, making it much more difficult to accidentally break "downstream" templates with "usptream" changes.
+
+`schnauzer` can also continue to be invoked as a setting generator without the need for a templated configuration file.
```
webern: Is this considered a temporary arrangement?

docs/design/settings/extensions/modular-settings-extensions.md
```diff
@@ -0,0 +1,256 @@
...
+
+The apiserver would load the current value of `motd` and then create a new object based on the inputted string.
+It would then find the extension responsible for `motd` and make a request to determine whether or not the change should proceed:
+
+```bash
+$ motd proto1 set --setting-version v1 --value '"Hello, Bottlerocket!"'  # Values are passed as JSON, hence the quoting
+```
+
+### Extension Naming & Setting Ownership
+Every top-level setting in Bottlerocket's API must be owned by a singular extension, and that extension will share a name with the setting.
```
webern, line 62: I think I might want to write a single extension that handles multiple settings. Will that be possible?

docs/design/settings/extensions/settings-extension-example.md
```diff
@@ -0,0 +1,112 @@
...
+## Configuration File Template
+In your variant repository, you will need to create an RPM package which installs `mydaemon` into your custom Bottlerocket variant.
+In this package, you will also want to include a templated configuration file, which is what the settings system interacts with:
+
+```
+├── packages
+│   └── mydaemon
+│       ├── build.rs
+│       ├── Cargo.toml
+│       ├── mydaemon.spec
```
webern, line 80: There's enough boilerplate including main.rs etc that I'm wondering if in the future sodapop or the internal build system could generate, at least the package.

webern (approved):
Nice work.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: I've changed my mind about this. I think the first case is still valid. If you attempt to validate a setting which isn't on the system, then that will silently succeed.

This could happen in cases where a vendor creates a settings extension that you are interested in, and it takes on many responsibilities but you don't particularly care for the validation functionality.

A version not being present will cause validation to fail. In this case, the author clearly wanted a validation to occur against settings which are present on the system but it cannot succeed.

Settings should have a `Requires` line in their RPM package specifying their related settings to prevent these issues at build time.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: I'll open another PR discussing particulars about this design.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: The algorithm is something like:

* The template rendering system will use the frontmatter section of the template to determine which settings extensions it needs to use to render the document.
* The template system uses the API to determine what handlebar helpers are provided by the settings extensions required.
* For each helper discovered, the template renderer will register a helper where the implementation results in an API call to execute the settings extension's method for calling that helper.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: You may be remembering an earlier form of this design where `get` requests would be routed to the extension to respond to. The idea here is that the API server can handle any arbitrary `get` request by referring to the stored JSON document.

Since the API server is also completely capable of handling arbitrary patches to JSON objects, it performs sets by changing the values requested by the user, and then just asking the settings extensions to validate them for correctness.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: With this design it's not possible. The setting extension itself would have to determine whether or not relevant subtrees of other settings are changed.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: I think this is actually just wrong, settings generators in concept go away. What may be relevant is that you could invoke schnauzer to arbitrarily render a view of Bottlerocket settings without needing to invoke the API.

docs/design/settings/extensions/modular-settings-extensions.md
cbgbt: Can you given an example of what you'd want here? It's not possible with this design, though you could probably get pretty close to what you want by sharing code via shared rust crates and taking advantage of cross-validation.

cbgbt:
^ Addresses feedback from reviewers.

docs/design/settings/extensions/settings-extension-example.md
cbgbt: I've given an example of how files will be arranged on the filesystem -- let me know if that seems ok to you.

docs/design/settings/extensions/settings-extension-example.md
```diff
@@ -0,0 +1,156 @@
...
+```
+sys-root
+└── usr
+    ├── libexec
+    │   ├── my-setting-binary
+    │   └── settings
+    │       └── my-setting -> ../my-setting-binary
+    └── share
+        └── settings
+            └── my-setting.toml
```
bcressey: nit: It'd help to stick with the `mydaemon` example to show how it all fits together.

```suggestion
sys-root
└── usr
    ├── libexec
    │   ├── mydaemon-settings
    │   └── settings
    │       └── mydaemon-> ../mydaemon-settings
    └── share
        └── settings
            └── mydaemon.toml
```

docs/design/settings/extensions/settings-extension-example.md
```diff
@@ -0,0 +1,156 @@
...
+        └── templates
+            ├── motd.template
+            ├── motd.template.affected-services
+            │       └── motd.service -> /etc/systemd/system/motd.service
+            └── motd.template.rendered-to
+                    └── motd.conf
+```
+
+`motd.template.rendered-to/motd.conf` contains instructions similar to `systemd-tmpfiles.d` explaining locations to which the rendered template will be placed.
+The format of this file dictates the resulting file location and mode of the resulting file:
```
bcressey: Clever, I missed the implications of this before. I like reusing part of the `tmpfiles.d` syntax for this. It might be good to also support setting UID / GID in addition to mode.
```
/etc/mydaemon/mydaemon.json - - -
/etc/mydaemon.env 0755 root root
```

bcressey (approved):
Nice work.

cbgbt:
^ Addressed feedback from @bcressey.

stockholmux:
@cbgbt: I know this is long-closed and merged, but re-reading a few months on, I have some follow up questions:
- would existing settings be implemented in this way as well (e.g. `settings.kubernetes.*`, `settings.network.*`, etc.)?
- is there any changes anticipated from the existing settings?

PR #2839, opened 2023-02-28
host-ctr: PID namespace tracking via bind-mounted manifest file

### LINKED ISSUE #2796: `host-ctr`: launched process namespace tracking via json file
### What I'd like:

I want to be able to read a file on the bottlerocket system that lists the PID namespaces of current host-containers. This way, via my own admin container or host container, I can monitor the currently running host containers and audit kernel events solely based on those PID namespaces.

**Proposal**: When host containers are launched via `host-ctr`, we could update a well known json file on the host with the PID namespaces of running host-containers.

This file could be placed in a well known location:

```
/var/lib/bottlerocket/host-container/container-namespaces.json
```

[similar to how `application.json` is populated in host-ctr for the RPM inventory.](https://github.com/bottlerocket-os/bottlerocket/blob/756f33c8f72bd8a78d9e8f41fb76a53eecc3df3d/sources/host-ctr/cmd/host-ctr/main.go#L731-L736)

This can be made available to running host-containers as a mount and also added as a link to a top level location:

```
/.bottlerocket/container-namespaces.json
```

This way, the file is easily accessible to _both_ host-containers with the mount and as system level packages down the road.

When containers stop, we would also want to update the json file to prevent any "dangling" PID namespaces that no longer have running host-containers.

Something else to keep in mind: when Bottlerocket goes through a reboot, containers are stopped and restarted. This "host container PID namespaces" file should likely _not_ be kept in persistent storage since it is directly tied to the running state of bottlerocket, not necessarily the state of the _image_.

### Any alternatives you've considered:

N/a - would love feedback from the community on this proposal!

I could also use some guidance on how we would anticipate these files to land on the system. Does the above proposed filepaths make sense?

etungsten:
**Issue number:**

Resolves https://github.com/bottlerocket-os/bottlerocket/issues/2796

**Description of changes:**

```
    host-ctr: add host containers information bind mount
    
    Superpowered host-containers now has a bind mount containing all
    host-container information (e.g. PID namespaces).

```
```

    host-ctr: container task PID NS tracking
    
    Create a manifest file that stores the PID namespace of all
    non-superpowered host-containers.

```

I was debating whether to have `host-ctr` sync the PID namespace "manifest" with the all of `host-containerd` daemon's container tasks every time it's invoked or just have `host-ctr` only update the "manifest" for the container task it's responsible for. I eventually settled for the latter since for me it was simpler to reason about in terms of concurrency and it adheres to the model of how `host-ctr` works so far, e.g. the `host-ctr` process for the `admin` container only deals with the `admin` container and does not interfere or gather information about other containers.

I don't feel too strongly about this, so if people want to have `host-ctr` sync for every container running in `host-containerd` then I can change it pretty easily.

**Testing done:**
From a fresh host's admin container, the control container is running.
You can check the list of host-containers PID namespaces via `/.bottlerocket/host-containers/info/pid-namespaces.json` in a superpowered host-container (e.g. admin):
```
[root@admin]# cat /.bottlerocket/host-containers/info/pid-namespaces.json                                                                        
{
  "pid-namespaces": [
    4026532532
  ]
[root@admin]# lsns | grep 4026532532                                                                                                             
4026532532 pid         6  1591 root /usr/bin/amazon-ssm-agent
                                                                 
```
After enabling a nginx host-container, I see the manifest gets populated with the nginx PID namespace inode number:
```
[root@admin]# apiclient set --json '{"host-containers": {"nginx":{"enabled":true, "source":"docker.io/library/nginx:latest"}}}' 
[root@admin]# cat /.bottlerocket/host-containers/info/pid-namespaces.json                                                       
{
  "pid-namespaces": [
    4026532532,
    4026532677
  ]
}
[root@admin]# lsns | grep 4026532677                                                                                            
4026532677 pid         3  7038 root nginx: master process nginx -g daemon off;
```
Then I disabled the nginx host container and observed the manifest gets updated accordingly:
```
[root@admin]# apiclient set --json '{"host-containers": {"nginx":{"enabled":false, "source":"docker.io/library/nginx:latest"}}}'
[root@admin]# cat /.bottlerocket/host-containers/info/pid-namespaces.json                                                       
{
  "pid-namespaces": [
    4026532532
  ]
}
```

Now I try the nginx again, but this time making it super-powered:
```
[root@admin]# apiclient set --json '{"host-containers": {"nginx":{"enabled":true, "superpowered":true, "source":"docker.io/library/nginx:latest"}}}'
[root@admin]# cat /.bottlerocket/host-containers/info/pid-namespaces.json                                                                           
{
  "pid-namespaces": [
    4026532532
  ]
}
```
It's PID NS is not tracked since it's shared with the root PID namespace. This is expected.
Nginx is confirmed running:
```
[root@admin]# curl localhost:80
<!DOCTYPE html>
<html>
<head>
<style>
html { color-scheme: light dark; }
body { width: 35em; margin: 0 auto;
font-family: Tahoma, Verdana, Arial, sans-serif; }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>

</body>
</html>
```
Cleaning it up:
```
[root@admin]# apiclient set --json '{"host-containers": {"nginx":{"enabled":false, "superpowered":true, "source":"docker.io/library/nginx:latest"}}}'
[root@admin]# curl localhost:80                                                                                                                      
curl: (7) Failed to connect to localhost port 80 after 0 ms: Connection refused
[root@admin]# cat /.bottlerocket/host-containers/info/pid-namespaces.json                                                                            
{
  "pid-namespaces": [
    4026532532
  ]
}
```

**Terms of contribution:**

### DIFF
--- a/packages/os/host-containers-tmpfiles.conf
+++ b/packages/os/host-containers-tmpfiles.conf
@@ -1,3 +1,4 @@
 d /etc/host-containers 0750 root root -
+d /etc/host-ctr-namespaces 0750 root root -
 d /local/host-containers 0700 root root -
 T /local/host-containers - - - - security.selinux=system_u:object_r:secret_t:s0
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -840,6 +840,12 @@ func withPrivilegedMounts() oci.SpecOpts {
 			Source:      "/mnt",
 			Type:        "bind",
 		},
+		{
+			Options:     []string{"rbind", "ro"},
+			Destination: "/.bottlerocket/host-ctr-namespaces",
+			Source:      "/etc/host-ctr-namespaces",
+			Type:        "bind",
+		},
 	})
 }
 
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	pidns "host-ctr/pidnamespaces"
 	"io"
 	"math/rand"
 	"os"
@@ -31,6 +32,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
+	"golang.org/x/sys/unix"
 )
 
 // Expecting to match ECR image names of the form:
@@ -432,6 +434,21 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 		log.G(ctx).Info("successfully started container task")
 	}
 
+	// Track the task's PID namespace in a manifest so monitoring agents can access it
+	taskPidNs, pidNsErr := pidns.GetTaskPidNs(task.Pid())
+	if taskPidNs == -1 {
+		if pidNsErr == unix.EPERM {
+			log.G(ctx).Info("container task shares root PID namespace, not tracking in PID NS manifest")
+		} else {
+			log.G(ctx).WithError(pidNsErr).Error("failed to get task process PID NS")
+		}
+	} else {
+		err = pidns.AddPidNsToManifest(taskPidNs)
+		if err != nil {
+			log.G(ctx).WithError(err).Error("failed to add task PID NS to manifest")
+		}
+	}
+
 	// Block until an OS signal (e.g. SIGTERM, SIGINT) is received or the
 	// container task finishes and exits on its own.
 
@@ -488,6 +505,14 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 	// Return error if container exists with non-zero status
 	if code != 0 {
 		return fmt.Errorf("Container %s exited with non-zero status", containerID)
+	}
+
+	// Remove the task's PID NS from the tracking manifest since the task has finished
+	if taskPidNs != -1 {
+		err = pidns.RemovePidNsFromManifest(taskPidNs)
+		if err != nil {
+			log.G(ctx).WithError(err).Error("failed to remove task PID NS from manifest")
+		}
 	}
 
 	return nil
@@ -564,6 +589,17 @@ func cleanUp(containerdSocket string, namespace string, containerID string) erro
 				return err
 			}
 			log.G(ctx).WithField("container-id", containerID).Info("killed existing container task")
+
+			// Remove the task's PID NS from the tracking manifest
+			taskPidNs, err := pidns.GetTaskPidNs(task.Pid())
+			if err != nil {
+				log.G(ctx).WithError(err).Error("failed to get task process PID NS")
+			} else if taskPidNs != -1 {
+				err = pidns.RemovePidNsFromManifest(taskPidNs)
+				if err != nil {
+					log.G(ctx).WithError(err).Error("failed to remove task PID NS from manifest")
+				}
+			}
 		}
 		if err := container.Delete(ctx, containerd.WithSnapshotCleanup); err != nil {
 			log.G(ctx).WithField("container-id", containerID).WithError(err).Error("failed to delete existing container")
--- a/sources/host-ctr/go.mod
+++ b/sources/host-ctr/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-sdk-go v1.44.176
 	github.com/awslabs/amazon-ecr-containerd-resolver v0.0.0-20221221205310-1903c4ed45d1
 	github.com/containerd/containerd v1.6.15
+	github.com/gofrs/flock v0.8.1
 	github.com/opencontainers/runtime-spec v1.0.3-0.20220909204839-494a5a6aca78
 	github.com/pelletier/go-toml v1.9.5
 	github.com/pkg/errors v0.9.1
--- a/sources/host-ctr/pidnamespaces/pid_namespaces.go
+++ b/sources/host-ctr/pidnamespaces/pid_namespaces.go
@@ -1,0 +1,166 @@
+package pidnamespaces
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/gofrs/flock"
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	lockfile     string = "/run/lock/host-ctr-pid-ns.lock"
+	manifestPath string = "/etc/host-ctr-namespaces/pid-namespaces.json"
+)
+
+// HostCtrPidNs represents the list of PID namespaces
+type HostCtrPidNs struct {
+	PidNamespaces []int64 `json:"pid-namespaces,omitempty"`
+}
+
+// GetTaskPidNs returns the PID NS associated with the container task PID or error on failure
+func GetTaskPidNs(pid uint32) (int64, error) {
+	pidNsFile, err := os.Open(fmt.Sprintf("/proc/%d/ns/pid", pid))
+	if err != nil {
+		return -1, fmt.Errorf("failed to open %d PID namespace file", pid)
+	}
+	pidNsFd := int(pidNsFile.Fd())
+
+	// Get file descriptor that refers to a parent namespace, ioctl() request '_IO(NSIO, 0x2)'
+	parentPidNsFd, err := unix.IoctlRetInt(pidNsFd, unix.NS_GET_PARENT)
+	if parentPidNsFd == -1 && err == unix.EPERM {
+		// If we can't access the parent PID namespace, this means the current container task is sharing the root PID
+		// namespace. We don't want to track the root PID namespace in the manifest, so we return -1.
+		return -1, err
+	} else if err != nil {
+		return -1, errors.Wrap(err, fmt.Sprintf("ioctl failed to 'NS_GET_PARENTNS' for '%s'", pidNsFile.Name()))
+	}
+
+	// The PID namespace is represented by the inode number of the task's PID namespace file
+	var sb unix.Stat_t
+	if err := unix.Fstat(pidNsFd, &sb); err != nil {
+		return -1, errors.Wrap(err, "failed to stat task PID namespace fd")
+	}
+
+	return int64(sb.Ino), nil
+}
+
+// AddPidNsToManifest takes a given PID's namespace and adds it to the internal manifest
+func AddPidNsToManifest(pidNs int64) (err error) {
+	fileLock := flock.New(lockfile)
+	lockCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	ok, err := fileLock.TryLockContext(lockCtx, 500*time.Millisecond)
+	if err != nil && !ok {
+		return errors.Wrap(err, "acquiring lock for writing to PID NS manifest failed with error")
+	}
+	defer func() {
+		if err = fileLock.Unlock(); err != nil {
+			err = errors.Wrap(err, "failed to release lock for writing to PID NS manifest")
+		}
+	}()
+
+	// Add task PID NS to manifest
+	hostCtrPidNs, err := readManifest(manifestPath)
+	if err != nil {
+		return err
+	}
+	for _, ele := range hostCtrPidNs.PidNamespaces {
+		// If it already exists in the manifest list for whatever reason, we can just return without doing anything.
+		if ele == pidNs {
+			return nil
+		}
+	}
+	hostCtrPidNs.PidNamespaces = append(hostCtrPidNs.PidNamespaces, pidNs)
+	if err = writeManifest(&hostCtrPidNs, manifestPath); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// RemovePidNsFromManifest takes a PID's namespace and removes it from the internal manifest
+func RemovePidNsFromManifest(pidNs int64) (err error) {
+	fileLock := flock.New(lockfile)
+	lockCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	ok, err := fileLock.TryLockContext(lockCtx, 500*time.Millisecond)
+	if err != nil && !ok {
+		return errors.Wrap(err, "acquiring lock for writing to PID NS manifest failed with error")
+	}
+	defer func() {
+		if err = fileLock.Unlock(); err != nil {
+			err = errors.Wrap(err, "failed to release lock for writing to PID NS manifest")
+		}
+	}()
+
+	// Remove task PID NS from manifest
+	hostCtrPidNs, err := readManifest(manifestPath)
+	if err != nil {
+		return err
+	}
+
+	for i, ele := range hostCtrPidNs.PidNamespaces {
+		// Remove PID NS from list
+		if ele == pidNs {
+			hostCtrPidNs.PidNamespaces = append(hostCtrPidNs.PidNamespaces[:i], hostCtrPidNs.PidNamespaces[i+1:]...)
+		}
+	}
+	if err = writeManifest(&hostCtrPidNs, manifestPath); err != nil {
+		return err
+	}
+	return nil
+}
+
+func readManifest(manifestPath string) (HostCtrPidNs, error) {
+	manifest, err := os.OpenFile(manifestPath, os.O_RDWR|os.O_CREATE, 0644)
+	if err != nil {
+		return HostCtrPidNs{}, errors.Wrap(err, "failed to open PID NS manifest file")
+	}
+	defer manifest.Close()
+
+	manifestStat, err := manifest.Stat()
+	if err != nil {
+		return HostCtrPidNs{}, errors.Wrap(err, "failed to stat PID NS manifest file")
+	}
+	raw := make([]byte, manifestStat.Size())
+	if _, err = manifest.Read(raw); err != nil {
+		return HostCtrPidNs{}, errors.Wrap(err, "failed to read PID NS manifest file")
+	}
+	hostCtrPidNsList := HostCtrPidNs{}
+	if len(raw) != 0 {
+		err = json.Unmarshal(raw, &hostCtrPidNsList)
+		if err != nil {
+			return HostCtrPidNs{}, errors.Wrap(err, "failed to unmarshal PID NS manifest")
+		}
+	}
+	return hostCtrPidNsList, nil
+}
+
+// writeManifest writes the list of host-container PID NS to the open manifest file
+func writeManifest(hostCtrPidNsList *HostCtrPidNs, manifestPath string) error {
+	newData, err := json.Marshal(*hostCtrPidNsList)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal PID NS list to JSON")
+	}
+	tempFile, err := os.CreateTemp(filepath.Dir(manifestPath), "temp-pidns-*.json")
+	if err != nil {
+		return errors.Wrap(err, "failed to create temporary PID NS file")
+	}
+	defer tempFile.Close()
+
+	if _, err = tempFile.Write(newData); err != nil {
+		return errors.Wrap(err, "failed to write to temporary PID NS file")
+	}
+
+	// Persist the temporary PID NS file to the PID NS manifest path to ensure atomicity
+	if err = os.Rename(tempFile.Name(), manifestPath); err != nil {
+		return errors.Wrap(err, "failed to persist temporary PID NS file to manifest path")
+	}
+	return nil
+}
--- a/sources/host-ctr/go.sum
+++ b/sources/host-ctr/go.sum
@@ -437,6 +437,8 @@ github.com/godbus/dbus/v5 v5.0.4/go.mod h1:xhWf0FNVPg57R7Z0UbKHbJfkEywrmjJnf7w5x
 github.com/godbus/dbus/v5 v5.0.6/go.mod h1:xhWf0FNVPg57R7Z0UbKHbJfkEywrmjJnf7w5xrFpKfA=
 github.com/godbus/dbus/v5 v5.1.0 h1:4KLkAxT3aOY8Li4FRJe/KvhoNFFxo0m6fNuFUO8QJUk=
 github.com/godbus/dbus/v5 v5.1.0/go.mod h1:xhWf0FNVPg57R7Z0UbKHbJfkEywrmjJnf7w5xrFpKfA=
+github.com/gofrs/flock v0.8.1 h1:+gYjHKf32LDeiEEFhQaotPbLuUXjY5ZqxKgXy7n59aw=
+github.com/gofrs/flock v0.8.1/go.mod h1:F1TvTiK9OcQqauNUHlbJvyl9Qa1QvF/gOUDKA14jxHU=
 github.com/gogo/googleapis v1.2.0/go.mod h1:Njal3psf3qN6dwBtQfUmBZh2ybovJ0tlu3o/AC7HYjU=
 github.com/gogo/googleapis v1.4.0/go.mod h1:5YRNX2z1oM5gXdAkurHa942MDgEJyk02w4OecKY87+c=
 github.com/gogo/googleapis v1.4.1 h1:1Yx4Myt7BxzvUr5ldGSbwYiZG6t9wGBZ+8/fX3Wvtq0=


etungsten:
Push above tries to make the linter happy.

etungsten:
Push above changes where the PID NS update happens when we call `host-ctr clean-up` to after the task is confirm killed/deleted

sources/host-ctr/go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-sdk-go v1.44.176
 	github.com/awslabs/amazon-ecr-containerd-resolver v0.0.0-20221221205310-1903c4ed45d1
 	github.com/containerd/containerd v1.6.15
+	github.com/gofrs/flock v0.8.1
```
jpmcb, line 9: `gofrs/flock` is abit notorious in the Go community since there hasn't been an accepted PR in years. Thankfully, it has almost no dependencies, so I'm not worried about using it from a security standpoint, it's just alittle stale.

The Go community has also considered exposing the internal file lock mechanism that Go itself uses: https://go.googlesource.com/proposal/+/master/design/33974-add-public-lockedfile-pkg.md but who knows when that'll happen.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+
+	"github.com/containerd/containerd"
+	"github.com/gofrs/flock"
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	// Get file descriptor that refers to a parent namespace, ioctl() request '_IO(NSIO, 0x2)'
+	nsGetParentNs        = 0xb702
```
arnaldo2792: You could use `unix.NS_GET_PARENT`

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+	return int64(sb.Ino), nil
+}
+
+// HostCtrPidNs represents the list of PID namespaces
+type HostCtrPidNs struct {
+	PidNamespaces []int64 `json:"pid-namespaces,omitempty"`
+}
+
+func addPidNsToManifest(pidNs int64) error {
+	fileLock := flock.New(lockfile)
```
arnaldo2792: Why not locking the manifest file itself?

jpmcb: Hmmm I'm not sure you're using the file lock correctly:

The [`Flock` struct](https://github.com/gofrs/flock/blob/6f010d1acea74a32f2f2066bfe324c08bbee30e3/flock.go#L29-L35) is a simple wrapper around a read write mutex:
```go
type Flock struct {
	path string
	m    sync.RWMutex
	fh   *os.File
	l    bool
	r    bool
}
```

And each subsequent call to `getTaskPidNs` and `addPidNsToManifest`creates a new `flock`:
```go
fileLock := flock.New(lockfile)
```

Which then is going to create a new RW mutex. From reading the `flock` code, it doesn't seem that those mutexes are mutually exclusive based on the _file path name_. A.k.a, creating a new `flock` with a the same filepath as an existing `flock` _will_ create a new RW mutex and won't prevent concurrent operations if one `flock` is used in one thread and the other flock is used in another.

Then, when you attempt to `TryLockContext`:
```go
locked, err := fileLock.TryLockContext(lockCtx, 500*time.Millisecond)
```
I think that will always succeed since the RW mutex has no other holds on it.

I believe the expectation from `flock` is that a single `flock` be used across many goroutines/threads to gain/prevent access to read and/or write a file. In that case, we _shouldn't_ be creating new locks inside threads, we should hoist that logic to then pass those `flock` mutexes down into the individual goroutines.

Maybe it makes sense to hoist it into the top level `main` function and pass it down into the dependent functions.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -834,6 +834,12 @@ func withPrivilegedMounts() oci.SpecOpts {
 			Source:      "/mnt",
 			Type:        "bind",
 		},
+		{
```
arnaldo2792, line 843: nit: typo in commit message `s/host-containers now has/host-containers now have`

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: Good point. I've always used a separate file for locking because of how we did it for the BR updates API flock. In that case we had to use a separate lock file because of other concerns I can't remember at the moment.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+		return errors.Wrap(err, "failed to open PID NS manifest file")
+	}
+	defer manifest.Close()
+
+	manifestStat, err := manifest.Stat()
+	if err != nil {
+		return errors.Wrap(err, "failed to stat PID NS manifest file")
+	}
+	raw := make([]byte, manifestStat.Size())
+	if _, err = manifest.Read(raw); err != nil {
```
jpmcb: Typically, instead of allocating memory ourselves, would be to call `os.ReadFile` which returns a byte slice anyways.

```go
func os.ReadFile(name string) ([]byte, error)
```

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return -1, fmt.Errorf("failed to open %d PID namespace file", pid)
+	}
+	pidNsFd := int(pidNsFile.Fd())
+
+	_, err = unix.IoctlGetInt(pidNsFd, nsGetParentNs)
+	if err == syscall.EPERM {
+		// If we can't access the parent PID namespace, this means the current container task is sharing the root PID
+		// namespace. We don't want to track the root PID namespace in the manifest, so we return -1.
+		return -1, nil
```
jpmcb: > the current container task is sharing the root PID namespace. We don't want to track the root PID namespace

I'd like to learn more about this - my intuition is that if the container is using the root PID namespace, that would qualify as a "host" task that we would want to enable third party software to be able to observe.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -432,6 +432,17 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 		log.G(ctx).Info("successfully started container task")
 	}
 
+	// Track the task's PID namespace in a manifest so monitoring agents can access it
+	taskPidNs, pidNsErr := getTaskPidNs(&task)
+	if pidNsErr != nil {
```
jpmcb: I'm fine with this, but looks like typically just call these `err` in the previous code although this is more clear.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: Actually, thinking about it a bit more. Is there any harm in maintaining a separate file for the lock? I don't see a compelling reason to lock on the file I'm trying to R/W from/to, I also have to create the path to the file and I'm not sure how ok it is for `flock` to be the one creating the file. In general, I feel safer keeping a different lock file, but that might be imprudent.

jpmcb: See my comment below - `flock` is just a thin wrapper around a read / write mutex. I don't believe the file lock is  actually locking on files themselves. If all we care about is 1 static file, maybe we remove some of the cruft at just use and provision a read / write mutex ourselves.

etungsten: The `flock` package is using UNIX flock https://github.com/gofrs/flock/blob/6f010d1acea74a32f2f2066bfe324c08bbee30e3/flock_unix.go#L57.

> Which then is going to create a new RW mutex. From reading the flock code, it doesn't seem that those mutexes are mutually exclusive based on the file path name. A.k.a, creating a new flock with a the same filepath as an existing flock will create a new RW mutex and won't prevent concurrent operations if one flock is used in one thread and the other flock is used in another.

If we're going by POSIX `flock` (https://linux.die.net/man/2/flock) then all that matters is the file descriptor for the lockfile stays consistent across calls when obtaining/releasing locks. Which it seems to be doing https://github.com/gofrs/flock/blob/6f010d1acea74a32f2f2066bfe324c08bbee30e3/flock_unix.go#L97.

Basically the `sync.RWMutex` mutex is not used for the actual `flock` but rather for making the `flock` syscalls.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: If you include the root PID namespace, then that will include basically all processes on the host and I thought that one of the requirements was not to monitor user workloads from orchestrators.

sources/host-ctr/cmd/host-ctr/main.go
etungsten: Reason why I keep it separate here is because I need to check later to make sure the previous `getTaskPidNs` call succeeded before I remove the PID NS from the manifest file after the task has exited:
https://github.com/bottlerocket-os/bottlerocket/blob/2401a7679181f9479f97252fc79957a549770ad4/sources/host-ctr/cmd/host-ctr/main.go#L504-L510

sources/host-ctr/go.mod
etungsten: Yeah, I saw the proposal. But luckily the package is extremely light-weight and straightforward. If we were to implement flocks ourselves, then we'd probably end up basically the same syscalls. So I'm not overly concerned about the dependency not having more recent commits.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: `os.ReadFile` opens the file and reads everything then closes the file.
I don't want to `syscall.Open` the file again while I already have an open file descriptor from the `os.OpenFile` before this with custom flags and modes, I just want to do everything with the single `manifest` open file descriptor.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: Huh - maybe I've been using `flock` wrong this whole time 😂 

Sounds good - I didn't dig that deep. Makes sense why my way works too: the `flock`s have those mutexes for shared locking and being able to call `TryLock` on the same `flock` across many different threads/goroutines.

So this works 👏🏼

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: Yeah I was wondering if that was the strategy you were going for since you already call `Stat` on the file descriptor 👏🏼 that works! Maybe a comment noting why we need the open file descriptor?

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: Ahhhh, yeah that makes sense 🤔

Would a way to subvert this be to start a host-container that assumes PID 0?

sources/host-ctr/cmd/host-ctr/main.go
jpmcb: Ahh, I missed that part 👏🏼 good thinking

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: Edit: I don't think there's harm in managing a seperate lock file. And yes, good call on not letting flock attempt to create a file. I think that makes sense.

arnaldo2792: Other than preventing concurrent writes from other processes while host-ctr is actually writing to the file, and not using yet another file in the `tmpfs` `/etc` filesystem, no. I just asked to understand if there was a deeper reason to create a separate file for the lock.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+	} else if err != nil {
+		return -1, errors.Wrap(err, fmt.Sprintf("ioctl failed to 'NS_GET_PARENTNS' for '%s'", pidNsFile.Name()))
+	}
+
+	// The PID namespace is represented by the inode number of the task's PID namespace file
+	var sb syscall.Stat_t
+	if err := syscall.Fstat(pidNsFd, &sb); err != nil {
```
arnaldo2792: I've seen that `syscall` is ~strikedthrough~, @jpmcb is the package on the deprecation path? 

Regardless, lets be consistent and use `unix.Fstat` (and `unix.*` wherever possible?)

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+
+	locked, err := fileLock.TryLockContext(lockCtx, 500*time.Millisecond)
+	if err != nil {
+		return errors.Wrap(err, "acquiring lock for writing to PID NS manifest failed with error")
+	}
+	if !locked {
+		return errors.New("failed to acquire lock for writing to PID NS manifest")
+	}
+	defer func() {
+		_ = fileLock.Unlock()
```
arnaldo2792: Should you handle the error here, and print a message? If not, then why not just this `defer fileLock.Unlock()`?

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+	fileLock := flock.New(lockfile)
+	lockCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	locked, err := fileLock.TryLockContext(lockCtx, 500*time.Millisecond)
+	if err != nil {
+		return errors.Wrap(err, "acquiring lock for writing to PID NS manifest failed with error")
+	}
+	if !locked {
+		return errors.New("failed to acquire lock for writing to PID NS manifest")
```
arnaldo2792: Shouldn't you retry a few times before bailing out? Or will we rely on systemd as the retry mechanism?

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		for _, ele := range tmpPidNsList.PidNamespaces {
+			// If it already exists in the manifest list for whatever reason, we can just return without doing anything.
+			if ele == pidNs {
+				return nil
+			}
+		}
+	}
+	hostCtrPidNsList.PidNamespaces = append(tmpPidNsList.PidNamespaces, pidNs)
+	newData, err := json.MarshalIndent(hostCtrPidNsList, "", "  ")
```
arnaldo2792: There isn't much value on writing the file with indentation. For debugging purposes, we can always do `cat <blah> | jq`.

Also this entire block could be its own function `writeNewData`, that recieves the open `File`, and the data.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+		return errors.New("failed to acquire lock for writing to pid-ns manifest")
+	}
+	defer func() {
+		_ = fileLock.Unlock()
+	}()
+
+	// Remove task PID NS from manifest
+	if err := os.MkdirAll(filepath.Dir(manifestPath), 0666); err != nil {
+		return errors.Wrap(err, "failed to create PID NS manifest directory")
+	}
```
arnaldo2792: You should create the manifest path with tmpfiles.d configurations, as we do with other `/etc` directories.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: Yes. [The standard library `syscall` is deprecated](https://pkg.go.dev/syscall):

> Deprecated: this package is locked down. Callers should use the corresponding package in the `golang.org/x/sys` repository instead.

So, instead,[ we'd want something like](https://pkg.go.dev/golang.org/x/sys@v0.5.0/unix#Fstat)

```go
import (
    golang.org/x/sys/unix
)

if err := unix.Fstat(pidNsFd, &sb); err != nil {
   // ...
}
```

jpmcb: Also note - although `syscall` has been deprecated for years (I think it was officially deprecated and frozen in Go 1.4), we'll see it used all over the place. For example, `gofrs/flock` uses it. And for most cases, it's fine and is much 1:1 in `/x/sys`: so i wouldn't call this a blocker if for some reason we can't use `/x/sys`

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: +1 - we should surface the error that this may raise from the nested `syscall.Flock` this attempts to make

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: > There isn't much value on writing the file with indentation.

I disagree, but I can see it going either way.

It's nice to save a few kilobytes of space in the file that is primarily intended for machines to read. But if a human is trying to do some quick debugging (and maybe they don't have `jq`), trying to read unformatted json is super not fun.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: `TryLockContext` blocks and retries to obtain the lock on a set interval (in this case every 0.5 seconds) until the context times out (set to 10 seconds). If it can't obtain the lock before then it errors.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: I originally had `defer fileLock.Unlock()` and the linter complained about throwing away the error. I'll try returning it.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+
+const (
+	// Get file descriptor that refers to a parent namespace, ioctl() request '_IO(NSIO, 0x2)'
+	nsGetParentNs        = 0xb702
+	lockfile      string = "/run/lock/host-ctr-pid-ns.lock"
+	manifestPath  string = "/etc/host-containers/info/pid-namespaces.json"
+)
+
+func getTaskPidNs(task *containerd.Task) (int64, error) {
+	pid := (*task).Pid()
```
jpmcb: Instead of having to do a confusing type assertion here (and to prevent having to copy the entire `task` interface into this function), can we simplify this by changing the signature to:

```go
func getPidNs(pid uint32) (int64, error)
```

Then, the call to this function would be:

```go
taskPidNs, err := getPidNs(task.Pid())
```

This also has the added benefit of dropping `github.com/containerd/containerd` as a dependency to this file (and therefore, further generalizes this in the weird future that we aren't using containerd)

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
...
+	var sb syscall.Stat_t
+	if err := syscall.Fstat(pidNsFd, &sb); err != nil {
+		return -1, errors.Wrap(err, "failed to stat task PID namespace fd")
+	}
+
+	return int64(sb.Ino), nil
+}
+
+// HostCtrPidNs represents the list of PID namespaces
+type HostCtrPidNs struct {
```
jpmcb: Nit: typically I've seen package public type / struct definitions towards the top, above private package functions

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: > Would a way to subvert this be to start a host-container that assumes PID 0?

I'm not sure if that's possible. There's no way to force set a PID of a process.

https://en.wikipedia.org/wiki/Process_identifier
> Process IDs are usually allocated on a sequential basis, beginning at 0 and rising to a maximum value which varies from system to system. Once this limit is reached, allocation restarts at 300 and again increases.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,184 @@
+package main
```
jpmcb: I think this makes sense as it's own package.

The API I'd imagine this package could provide could maybe be:
```go
// the pidnamespaces package provides functions for
package pidnamespaces

// GetPidNs returns the namespace for a give PID or error on failure
func GetPidNs(pid uint32) (int64, error)

// AddPidNsToManifest takes a given PID's namespace and adds it to the internal manifest
func AddPidNsToManifest(pidNs int64) error

// RemovePidNsFromManifest takes a PID's namespace and removes it from the internal manifest
func RemovePidNsFromManifest(pidNs int64) error
```

Generally, we should be mindful of what we pollute into the `main` package. The above abstraction gives us the ability in the future to possibly tease some things out further in the future (like an interface that can store PIDs and namespaces, or dynamically set the manifest path, etc.)

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: > Maybe a comment noting why we need the open file descriptor?

Not sure what I'd say to clarify what's happening since I believe the flow of what's happening to the `fd` seems pretty straightforward.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: Isn't checking that `locked == true` redundant? I'm not seeing a case in the flock code where an error is returned that is not nil _and_ locked does not equal true.

Or in other words, I think this can generally be refactored to be:

```go
ok, err := fileLock.TryLockContext(lockCtx, 500*time.Millisecond)
if err != nil && !ok {
	return errors.Wrap(err, "acquiring lock for writing to PID NS manifest failed with error")
}
```

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: > Other than preventing concurrent writes from other processes while host-ctr is actually writing to the file

Nothing prevents an outside process to write to the file even if it was `flock`ed by `host-ctr`.

From the man page:
>  flock() places advisory locks only; given suitable permissions on a file, a process is free to ignore the use of flock() and perform I/O on the file.

The `flock`ing is really only used for concurrency between `host-ctr` processes.

I elected to use a separate file for the filelock since I believe it's simpler (when thinking about which `host-ctr` is gonna flock first/create the manifest file) and the only downside being needing to allocate an extra file descriptor/inode.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
```diff
@@ -0,0 +1,172 @@
...
+		return errors.Wrap(err, "failed to stat PID NS manifest file")
+	}
+	raw := make([]byte, manifestStat.Size())
+	if _, err = manifest.Read(raw); err != nil {
+		return errors.Wrap(err, "failed to read PID NS manifest file")
+	}
+	var (
+		tmpPidNsList     HostCtrPidNs
+		hostCtrPidNsList HostCtrPidNs
+	)
```
jpmcb: Nit: typically, I've seen the community prefer the `:=` operator. Something like:

```go
tmpPidNsList := &HostCtrPidNs{}
hostCtrPidNsList := &HostCtrPidNs{}
```

This is a bit more clear that you're allocating empty `HostCtrPidNs`

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: Looked alittle at this, and it seems that returning errors out of `defer` isn't super simple or elegant.

Maybe the easiest thing for us to do is just to call `fileLock.Unlock()` ourselves and capture the error.

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
etungsten: I chose to remove the indents for now since it's mostly meant for machines and there's not much going on here anyways to make it hard to read. We can always revisit this when we decide to add more to it.

etungsten:
https://github.com/bottlerocket-os/bottlerocket/compare/e3c4eb2094477e918a3e29c34202bb67be05149b..e995e8e79cc7986c773d0787aa885401d1f1833e
https://github.com/bottlerocket-os/bottlerocket/compare/e995e8e79cc7986c773d0787aa885401d1f1833e..7fb85bb9a580fe870db59921cef1c0496ab68f4f
https://github.com/bottlerocket-os/bottlerocket/compare/7fb85bb9a580fe870db59921cef1c0496ab68f4f..8ba8a62b8f2d6aaa020432399e34538ea71f37ee
Pushes above addresses @jpmcb and @arnaldo2792 comments.

- Broke out pid_namespaces.go to it's own local package.
- Add tmpfile.d conf for creating `/etc/host-containers/info`
- and other misc

confirmed that things still work as expected:
```
[root@admin]# cat /.bottlerocket/host-containers/info/pid-namespaces.json 
{"pid-namespaces":[4026532532,4026532536]}
[root@admin]# apiclient set --json '{"host-containers": {"nginx":{"enabled":false, "source":"docker.io/library/nginx:latest"}}}'
[root@admin]# cat /.bottlerocket/host-containers/info/pid-namespaces.json                                                       
{"pid-namespaces":[4026532536]}                                                 
[root@admin]# apiclient set --json '{"host-containers": {"nginx":{"enabled":true, "source":"docker.io/library/nginx:latest"}}}'  
[root@admin]# cat /.bottlerocket/host-containers/info/pid-namespaces.json                                                        
{"pid-namespaces":[4026532536,4026532532]}
```

etungsten:
Push above removes a redundant `os.MkdirAll`

sources/host-ctr/cmd/host-ctr/pid_namespaces.go
jpmcb: 👏🏼  sounds good!

sources/host-ctr/pidnamespaces/pid_namespaces.go
```diff
@@ -0,0 +1,165 @@
...
+			if ele == pidNs {
+				return nil
+			}
+		}
+	}
+	hostCtrPidNsList.PidNamespaces = append(tmpPidNsList.PidNamespaces, pidNs)
+	if err = writeManifest(hostCtrPidNsList, manifest); err != nil {
+		return err
+	}
+	if err = fileLock.Unlock(); err != nil {
```
jpmcb: Hmm on second thought - will this keep the file lock if we return earl on an error? I guess the benefit of having the `defer` is that it will always call that 😢 how often do we expect a `Syscall` to fail deep in the flock code ...? Maybe go back to throwing away the error if that's the easiest thing and prevents a deadlock

etungsten: Yeah, that's what I was thinking as well. I can return an error from a deferred function using named error vars, so I'll just do that again.

etungsten: Also to answer 
>  how often do we expect a Syscall to fail deep in the flock code

I do not expect this to ever error if the previous call to obtain the lock has succeeded. Of all the possible errors that can be returned by `flock(2)`, only 1 of them can happen during unlock; it only errors if the underlying file descriptor becomes invalid (e.g. lock file deleted or something)

etungsten:
Push above reverts back to deferring unlocking the lockfile.

sources/host-ctr/pidnamespaces/pid_namespaces.go
jpmcb: > using named error vars

Excellent solution!

```go
defer func() {
	if err = fileLock.Unlock(); err != nil {
		err = errors.Wrap(err, "failed to release lock for writing to PID NS manifest")
	}
}()
```

arnaldo2792 (approved):

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -804,6 +835,12 @@ func withPrivilegedMounts() oci.SpecOpts {
 			Source:      "/mnt",
 			Type:        "bind",
 		},
+		{
+			Options:     []string{"rbind", "ro"},
+			Destination: "/.bottlerocket/host-containers/info/",
+			Source:      "/etc/host-containers/info",
```
bcressey: This path is also used for the host container's user data and persistent storage, so a host container named `info` would have a bad time.

packages/os/host-containers-tmpfiles.conf
```diff
@@ -1,3 +1,4 @@
 d /etc/host-containers 0750 root root -
+d /etc/host-containers/info 0640 root root -
```
bcressey: It is better to create directories with the execute bit, otherwise they don't work properly unless the current user can override DAC:
```suggestion
d /etc/host-containers/info 0750 root root -
```

bcressey: nit: might also be better to use a new directory, like `host-namespaces` for this data. That would help with the collision I mentioned below for a container named "info".

sources/host-ctr/pidnamespaces/pid_namespaces.go
```diff
@@ -0,0 +1,169 @@
...
+// GetTaskPidNs returns the PID NS associated with the container task PID or error on failure
+func GetTaskPidNs(pid uint32) (int64, error) {
+	pidNsFile, err := os.Open(fmt.Sprintf("/proc/%d/ns/pid", pid))
+	if err != nil {
+		return -1, fmt.Errorf("failed to open %d PID namespace file", pid)
+	}
+	pidNsFd := int(pidNsFile.Fd())
+
+	// Get file descriptor that refers to a parent namespace, ioctl() request '_IO(NSIO, 0x2)'
+	_, err = unix.IoctlGetInt(pidNsFd, unix.NS_GET_PARENT)
```
bcressey: From looking at [pid_namespaces.go](https://man7.org/tlpi/code/online/dist/namespaces/pid_namespaces.go.html) it seems like the parent pid namespace fd will be the return value from this ioctl, and it's the combination of `EPERM` and `-1` that tells us we reached the root PID namespace.

So I don't think we want to discard the return value; then we can't differentiate this case from a general `EPERM`.

sources/host-ctr/pidnamespaces/pid_namespaces.go
```diff
@@ -0,0 +1,169 @@
...
+		return errors.Wrap(err, "acquiring lock for writing to PID NS manifest failed with error")
+	}
+	defer func() {
+		if err = fileLock.Unlock(); err != nil {
+			err = errors.Wrap(err, "failed to release lock for writing to PID NS manifest")
+		}
+	}()
+
+	// Add task PID NS to manifest
+	manifest, err := os.OpenFile(manifestPath, os.O_RDWR|os.O_CREATE, 0664)
```
bcressey:
```suggestion
	manifest, err := os.OpenFile(manifestPath, os.O_RDWR|os.O_CREATE, 0644)
```

sources/host-ctr/pidnamespaces/pid_namespaces.go
```diff
@@ -0,0 +1,169 @@
...
+	defer manifest.Close()
+
+	manifestStat, err := manifest.Stat()
+	if err != nil {
+		return errors.Wrap(err, "failed to stat PID NS manifest file")
+	}
+	raw := make([]byte, manifestStat.Size())
+	if _, err = manifest.Read(raw); err != nil {
+		return errors.Wrap(err, "failed to read PID NS manifest file")
+	}
```
bcressey: This has a TOCTOU issue since the file might grow between the `stat` and `read` calls.

Is there not a helper that reads all bytes from a file into a `[]byte` without the stat call?

sources/host-ctr/pidnamespaces/pid_namespaces.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := manifest.Truncate(0); err != nil {
+		return errors.Wrap(err, "failed to truncate PID NS manifest")
+	}
+	if _, err := manifest.Seek(0, 0); err != nil {
+		return errors.Wrap(err, "failed to reset PID NS manifest I/O offset")
+	}
+	if _, err = manifest.Write(newData); err != nil {
+		return errors.Wrap(err, "failed to write to PID NS manifest")
+	}
```
bcressey: Ideally we'd write a temporary file into the same directory, then copy it over top of the target file.

That way, an application using inotify would see either the old file or the new file. Here I think it might see a zero-byte file or else a partially-written one.

sources/host-ctr/pidnamespaces/pid_namespaces.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to stat PID NS manifest file")
+	}
+	raw := make([]byte, manifestStat.Size())
+	if _, err = manifest.Read(raw); err != nil {
+		return errors.Wrap(err, "failed to read PID NS manifest file")
+	}
+	hostCtrPidNsList := &HostCtrPidNs{}
+	if len(raw) != 0 {
+		err = json.Unmarshal(raw, &hostCtrPidNsList)
```
bcressey: Maybe factor out a helper that just turns a file path (`manifestPath`) into a `HostCtrPidNs{}`.

sources/host-ctr/pidnamespaces/pid_namespaces.go
```diff
@@ -0,0 +1,169 @@
...
+		if err != nil {
+			return errors.Wrap(err, "failed to unmarshal PID NS manifest")
+		}
+		for i, ele := range hostCtrPidNsList.PidNamespaces {
+			// Remove PID NS from list
+			if ele == pidNs {
+				hostCtrPidNsList.PidNamespaces = append(hostCtrPidNsList.PidNamespaces[:i], hostCtrPidNsList.PidNamespaces[i+1:]...)
+			}
+		}
```
bcressey: Is it possible to deserialize into something like a HashSet instead of an array? Then we wouldn't have to iterate across the entire list each time.

sources/host-ctr/pidnamespaces/pid_namespaces.go
etungsten: @jpmcb called out the same thing in https://github.com/bottlerocket-os/bottlerocket/pull/2839#discussion_r1120911348. My reasoning was that I already had an open fd I can use without having to call `ReadFile` which opens the file again for reading. There is no utility in `os.File` that lets you read all the bytes. You can only call `os.ReadFile` for that (which actually does the same thing https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/os/file.go;l=674; they stat then make the buffer).

In terms of TOCTOU, it is still a concern if there can only be one `host-ctr` process opening/touching this file?

sources/host-ctr/pidnamespaces/pid_namespaces.go
etungsten: `golang` std library is pretty bare in terms of collections. So I'd have to make my own set, something like `map[int]bool{}` with it's own interface. I was thinking the size of this would be small enough that iterating over the list of host-containers isn't too bad.

etungsten:

etungsten:

- refactored out a `readManifest` helper
- changed manifest path from `/etc/host-containers/info/pid-namespaces.json` to `/etc/host-ctr-namespaces/pid-namespaces.json`
- write to tempfile then persist it for atomic replacement of the manifest file.


Tested changes and it still works as expected.

sources/host-ctr/pidnamespaces/pid_namespaces.go
etungsten: After the refactor, I ended up using `os.ReadFile` but it doesn't avoid the `stat` and `read`.

sources/host-ctr/pidnamespaces/pid_namespaces.go
jpmcb: It wouldn't need it's own interface necessarily
```go
piHashMap := make(map[int]bool)
```
or if it was the primary data structure in `hostCtrPidNs`:
```go
type HostCtrPidNs struct {
	PidNamespaces map[int64]bool `json:"pid-namespaces,omitempty"`
}
```

Then we could populate it like:
```go
// add PID to map
hostCtrPidNs.PidNamespaces[pid] = true

// remove PID from map
delete(hostCtrPidNs.PidNamespaces, pid)
```
instead of using append and having to use wonky slice mapping:
```go

for i, ele := range hostCtrPidNs.PidNamespaces {
	// Remove PID NS from list
	if ele == pidNs {
		hostCtrPidNs.PidNamespaces = append(hostCtrPidNs.PidNamespaces[:i], hostCtrPidNs.PidNamespaces[i+1:]...)
	}
}
```
accessing is also nice:
```go
if _, ok := hostCtrPidNs.PidNamespaces[pid]; ok {
    // yes, pid is in map, do stuff with it here
}
```

If we did drop the map into the `hostCtrPidNs`, the tricky part would be figuring out how to deserialize it elegantly into json without abunch of `true` values with the pid keys

etungsten:
Push above attempts to clean-up logic differentiating container task sharing root pid ns vs other errors.

sources/host-ctr/pidnamespaces/pid_namespaces.go
etungsten: There apparently is no easy way to drop map values during JSON marshaling/unmarshaling in Go without having to write my own custom marshal-er. I don't think the benefits of using a set for a relatively small size scale of host-containers is worth the additional complexity and maintenance of implementing my own marshaler. 

I'd essentially have to copy map keys to a list before marshaling and rebuild the map when unmarshaling from a list. The benefit is the `O(1)` insert vs `O(n)` insert which on the scale a few hundred host-containers (not sure people run that many host-containers) seems to matter not much and we don't really gain anything because the cost of copying to/from a list is `O(n)` still (happens every time we want to insert/remove).

jpmcb: > If we did drop the map into the hostCtrPidNs, the tricky part would be figuring out how to deserialize it elegantly into json without abunch of true values with the pid keys

Hmmm thinking further on this (and chatting with Erikson), i don't know if there's a great way to marshal a struct into json as a json array without having a custom marshaler:

```go
type PidHashMap map[int64]bool

func (phm PidHashMap) MarshalJSON() ([]byte, error) {
	var out []int64
	for k := range phm {
		out = append(out, k)
	}
	return json.Marshal(out)
}
```

Reference: https://go.dev/play/p/MvPImVWHIck

Note that we would have to allocate a slice anyways in order to output only a json array without the key/values as json objects

sources/host-ctr/pidnamespaces/pid_namespaces.go
etungsten: `IoctlGetInt` and `IoctlRetInt` both return `0` if there's an error.

https://cs.opensource.google/go/x/sys/+/refs/tags/v0.5.0:unix/ioctl_linux.go;l=15
https://go-review.googlesource.com/c/sys/+/192780


Not sure how we can differentiate. I think I might need to just `syscall` directly without the wrapper and deal with the return value myself.

etungsten:
Closing this PR for the time being now that we've determine a path forward for supporting monitoring agents. Will revisit if we truly need `host-ctr` to inform host-containers of what other host containers are running.

Issue #2843, opened 2023-02-28
Support and use cgroup v2


markusboehme:
**What I'd like:** I'd like Bottlerocket to use cgroup v2/the unified cgroup hierarchy. The general ecosystem is moving into this direction:

* cgroup v2 has been considered stable for some time in the kernel already.
* Kubernetes 1.25 declared support for cgroup v2 as GA with release 1.25.
* systemd changed its default to cgroup v2.
* Features such as Pressure Stall Information metrics from the kernel are available on a per-cgroup level with cgroup v2.

cgroup v2 support should be available for all variants of Bottlerocket per opt-in. Newly released variants should make cgroup v2 the default (with the chance to opt-out).

**Any alternatives you've considered:** I don't see any alternatives for the long term. cgroup v1 will receive less attention and testing as time goes by, and likely will be phased out of projects Bottlerocket relies on.

markusboehme:
Announced the upcoming change in the GitHub discussions: https://github.com/bottlerocket-os/bottlerocket/discussions/2874

markusboehme:
Relevant PRs:

* https://github.com/bottlerocket-os/bottlerocket-admin-container/pull/76
* #2802
* #2845 
* #2864 
* #2875

PR #2845, opened 2023-02-28
buildsys, systemd: add unified-cgroup-hierarchy image feature

markusboehme:
**Issue number:**

#2843 

**Description of changes:**

Add a new image feature toggle `unified-cgroup-hierarchy`. The image feature allows variants to select either cgroup v1 or cgroup v2 (the "unified cgroup hierarchy") as the default for systemd to set up during host boot.

The image feature changes how systemd is configured at build time. This results in slightly differing builds of systemd depending on variants' selected image features. Users can still override the default choice via a command line argument (`systemd.unified_cgroup_hierarchy=[01]`) supplied via the user data and boot config mechanisms.

Alternatively, the default cgroup version to use could be selected by baking a command line argument into the image. However, this would present a liability since GRUB and its configuration are not being updated during image updates, and some hosts would be stuck with it forever. Whether a user could override this setting via their own command line arguments would come down to parsing logic handles duplicate settings and seems brittle, especially when trying to provide guarantees "forever".

**Testing done:**

* [x] Modified `variants/metal-dev/Cargo.toml` to carry `unified-cgroup-hierarchy = true` in the `package.metadata.build-variant.image-features]` section
* [x] Built the `metal-dev` variant
* [x] Confirmed the image booted with a cgroup v2 hierarchy by checking `mount | grep cgroup` output and only finding a single `cgroup2` mount
* [x] Rebuilt the `metal-dev` variant to confirm the systemd build is cached if the image feature flag doesn't change
* [x] Built the (unchanged) `aws-dev` variant to confirm systemd and dependent packages get rebuilt if the image feature flag changes

**Reviewer notes:**

This does not yet make any variant actually default to cgroup v2, but only provides the means to do so. This is still pending version 0.10.0 of the admin container getting deployed, which carries changes to make it work on cgroup v2 hosts.

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,6 +68,7 @@ ARG VARIANT_FLAVOR
 ARG REPO
 ARG GRUB_SET_PRIVATE_VAR
 ARG SYSTEMD_NETWORKD
+ARG UNIFIED_CGROUP_HIERARCHY
 ENV SYSTEMD_NETWORKD=${SYSTEMD_NETWORKD}
 ENV VARIANT=${VARIANT}
 WORKDIR /home/builder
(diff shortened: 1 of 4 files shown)


bcressey:
> However, this would present a liability since GRUB and its configuration are not being updated during image updates, and some hosts would be stuck with it forever.

Pedantic note: it's only the minimal config baked into GRUB that isn't updated, since GRUB itself is never updated. The main GRUB config on the `/boot` partition does get updated; otherwise the dm-verity hash for the new root filesystem would be incorrect.

> Whether a user could override this setting via their own command line arguments would come down to parsing logic handles duplicate settings and seems brittle, especially when trying to provide guarantees "forever".

I previously believed, incorrectly, that boot config could override some of the stock parameters, but you're right that it almost always doesn't work that way, and it's probably best that it doesn't. So while I was surprised to see an image feature flag used here, I agree that it's the correct way to handle it to ensure that nodes can be downgraded into cgroup v1 if needed.

bcressey (approved):

arnaldo2792 (approved):

foersleo (approved):

markusboehme:
>> However, this would present a liability since GRUB and its configuration are not being updated during image updates, and some hosts would be stuck with it forever.
>
> Pedantic note: it's only the minimal config baked into GRUB that isn't updated, since GRUB itself is never updated. The main GRUB config on the /boot partition does get updated; otherwise the dm-verity hash for the new root filesystem would be incorrect.

Thanks for pointing this out! It indeed is the quite obvious counterexample. I updated the commit message accordingly.

PR #2848, opened 2023-03-02
Makefile: use SDK for Rust formatting

### LINKED ISSUE #2846: Use the SDK for `cargo fmt`
Follow-up to https://github.com/bottlerocket-os/bottlerocket/issues/2704 and https://github.com/bottlerocket-os/bottlerocket/pull/2793.

Aside from using the SDK for `cargo clippy`, it would be good to use it for `cargo fmt` as well. This avoids conflicts between differing versions locally and in the Github actions.

mjsterckx:
**Issue number:**

Closes #2846

**Description of changes:**

Changed the `rustfmt` used in the `check-fmt` target for Rust formatting to the one included in the BUILDSYS_SDK to avoid conflicting lints between local builds and GitHub actions.

Testing done:

Ran cargo make check-fmt locally.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -470,7 +470,6 @@ if ! docker run --rm \
    -e CARGO_HOME="/tmp/.cargo" \
    -v "${CARGO_HOME}":/tmp/.cargo \
    -v "${BUILDSYS_ROOT_DIR}/sources":/tmp/sources \
-   -v "${BUILDSYS_ROOT_DIR}/tools":/tmp/tools \
    -e VARIANT \
    "${BUILDSYS_SDK_IMAGE}" \
    cargo clippy \
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -401,16 +401,29 @@ for m in ${GO_MODULES}; do
 done
 
 # For rust first-party source code
-if ! cargo fmt \
-  --manifest-path ${BUILDSYS_SOURCES_DIR}/Cargo.toml \
+if ! docker run --rm \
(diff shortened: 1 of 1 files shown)


Makefile.toml
```diff
@@ -401,16 +401,30 @@ for m in ${GO_MODULES}; do
...
 
 # For rust first-party source code
-if ! cargo fmt \
-  --manifest-path ${BUILDSYS_SOURCES_DIR}/Cargo.toml \
+if ! docker run --rm \
+   -u $(id -u):$(id -g) \
+   -e CARGO_HOME="/tmp/.cargo" \
+   -v "${CARGO_HOME}":/tmp/.cargo \
+   -v "${BUILDSYS_ROOT_DIR}/sources":/tmp/sources \
+   -v "${BUILDSYS_ROOT_DIR}/tools":/tmp/tools \
```
jpculp:
```suggestion
```
Do we need `tools` mounted if we're just checking `sources`? I suspect it can be removed and the reverse is true for the next block. We might also want to make the same change to the `clippy` piece as well (in a separate commit).

mjsterckx: We need `sources` mounted in both blocks but `tools` is indeed only necessary when the `tools` manifest is checked. I'll go ahead and change that.

jpculp: Ah. It looks like a couple of `tools` are leveraging `parse-datetime` from `sources`.

jpculp (approved):

jpmcb (approved):

PR #2829, opened 2023-02-24
packages: Add ethtool to release

yeazelm:
**Issue number:**

Closes # 2794

**Description of changes:**
This adds ethtool to all variants.

**Testing done:**
Validated it works on a k8s variant in AWS:
```bash
[root@admin]# sudo sheltie
bash-5.1# ethtool eth0
Settings for eth0:
        Current message level: 0x000000f3 (243)
                               drv probe ifdown ifup rx_err tx_err
        Link detected: yes
bash-5.1# ethtool --version
ethtool version 6.2
```
Tested the use case from https://github.com/bottlerocket-os/bottlerocket/issues/2794
``` bash 
bash-5.1# ethtool -S eth0
NIC statistics:
     tx_timeout: 0
     suspend: 0
     resume: 0
     wd_expired: 0
     interface_up: 1
     interface_down: 0
     admin_q_pause: 0
     reset_fail: 0
     bw_in_allowance_exceeded: 17
     bw_out_allowance_exceeded: 0
     pps_allowance_exceeded: 0
     conntrack_allowance_exceeded: 0
     linklocal_allowance_exceeded: 0
     conntrack_allowance_available: 76950
     queue_0_tx_cnt: 11387
     queue_0_tx_bytes: 1587195
     queue_0_tx_queue_stop: 0
     queue_0_tx_queue_wakeup: 0
     queue_0_tx_dma_mapping_err: 0
     queue_0_tx_linearize: 0
     queue_0_tx_linearize_failed: 0
     queue_0_tx_napi_comp: 17282
     queue_0_tx_tx_poll: 17593
     queue_0_tx_doorbells: 11222
     queue_0_tx_prepare_ctx_err: 0
     queue_0_tx_bad_req_id: 0
     queue_0_tx_llq_buffer_copy: 2029
     queue_0_tx_missed_tx: 0
     queue_0_tx_unmask_interrupt: 17282
     queue_0_tx_xsk_need_wakeup_set: 0
     queue_0_tx_xsk_wakeup_request: 0
     queue_0_rx_cnt: 124148
     queue_0_rx_bytes: 225658754
     queue_0_rx_rx_copybreak_pkt: 5130
     queue_0_rx_csum_good: 120904
     queue_0_rx_refil_partial: 0
     queue_0_rx_csum_bad: 0
     queue_0_rx_page_alloc_fail: 0
     queue_0_rx_skb_alloc_fail: 0
     queue_0_rx_dma_mapping_err: 0
[...]

PR #2850, opened 2023-03-03
vmware: Disable UDP offload for primary interface

zmrow:
**Issue number:**
N/A
Depends on #2829 

**Description of changes:**
We've had reports of Bottlerocket in VMware having networking issues that we've traced back to the combination of our 5.10 kernel and the `vmxnet3` driver.  This driver enables UDP offload which has caused conflicts with NSX-T in certain environments.

This PR adds a systemd unit to VMware variants only.  The systemd unit disables UDP offload via `ethtool` commands for the default primary interface in VMware: `eth0`.

The unit _does_ reference `eth0`.  Given that `eth0` is the default primary interface in VMware, and changing it isn't trivial (especially with EKS-A), this felt pretty safe and should cover the 98% use case.

If a user does opt to do their own networking shenanigans, and runs into an issue where UDP offload is a problem, they are able to use a bootstrap container to run this same set of commands for their preferred interface.

**Testing done:**
* Build an `aws-k8s-1.24` image and ensure the unit does not exist and is not run
```
bash-5.1# systemctl --type=service | grep -i disable-udp
bash-5.1#
```
* Build a `vmware-k8s-1.24` image and ensure the unit runs and the settings are applied for the interface. 
```
bash-5.1# systemctl --type=service | grep -i disable-udp
  disable-udp-offload.service            loaded active     exited          Disables UDP offload

bash-5.1# ethtool -k eth0        
Features for eth0:
...
tx-udp_tnl-segmentation: off
tx-udp_tnl-csum-segmentation: off
...
```

Without the systemd unit - the settings show as "on":
```
bash-5.1# ethtool -k eth0 | grep -i tx-udp
tx-udp_tnl-segmentation: on
tx-udp_tnl-csum-segmentation: on
```

**Terms of contribution:**

### DIFF
--- a/packages/os/disable-udp-offload.service
+++ b/packages/os/disable-udp-offload.service
@@ -1,0 +1,20 @@
+[Unit]
+Description=Disables UDP offload
+After=network-online.target
+# Block manual interactions with this service
+RefuseManualStart=true
+RefuseManualStop=true
+# This unit disables UDP offload for the default primary interface (eth0) in
+# VMware.  This avoids an issue with the VXLAN / Geneve tunnels used by Cilium,
+# which affects the vmxnet3 driver in recent kernels.
+ConditionVirtualization=vmware
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/ethtool -K eth0 tx-udp_tnl-segmentation off
+ExecStart=/usr/sbin/ethtool -K eth0 tx-udp_tnl-csum-segmentation off
+RemainAfterExit=true
+StandardError=journal+console
+
+[Install]
+RequiredBy=preconfigured.target
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -46,6 +46,7 @@ Source117: cfsignal.service
 Source118: generate-network-config.service
 Source119: reboot-if-required.service
 Source120: warm-pool-wait.service
+Source121: disable-udp-offload.service
 
 # 2xx sources: tmpfilesd configs
 Source200: migration-tmpfiles.conf
@@ -442,6 +443,10 @@ install -d %{buildroot}%{_cross_udevrulesdir}
 install -p -m 0644 %{S:300} %{buildroot}%{_cross_udevrulesdir}/80-ephemeral-storage.rules
 install -p -m 0644 %{S:301} %{buildroot}%{_cross_udevrulesdir}/81-ebs-volumes.rules
 
+%if %{with vmware_platform}
+install -p -m 0644 %{S:121} %{buildroot}%{_cross_unitdir}
+%endif
+
 %cross_scan_attribution --clarify %{_builddir}/sources/clarify.toml \
     cargo --offline --locked %{_builddir}/sources/Cargo.toml
 
@@ -465,6 +470,10 @@ install -p -m 0644 %{S:301} %{buildroot}%{_cross_udevrulesdir}/81-ebs-volumes.ru
 %{_cross_bindir}/netdog
 %{_cross_tmpfilesdir}/netdog.conf
 %{_cross_unitdir}/generate-network-config.service
+%if %{with vmware_platform}
+%{_cross_unitdir}/disable-udp-offload.service
+%endif
+
 
 %files -n %{_cross_os}corndog
 %{_cross_bindir}/corndog


yeazelm (approved):

stmcginnis (approved):

bcressey:
> In VMware, Bottlerocket's 5.10 kernel and the vmxnet3 driver can conflict with NSX-T in certain environments.  Disabling UDP offload works around the problem.

Couple notes:
* 5.15 kernel is also affected
* NSX / NSX-T reportedly is not involved

Please revise the PR description and commit message accordingly.

packages/os/disable-udp-offload.service
```diff
@@ -0,0 +1,22 @@
+# This unit disables UDP offload for the default primary interface (eth0) in
+# VMware.  This avoids an issue where newer kernels (5.10+) and the vmxnet3
+# driver conflict with NSX-T.
```
bcressey:
```suggestion
# VMware.  This avoids an issue with the VXLAN / Geneve tunnels used by
# Cilium, which affects the vmxnet3 driver in recent kernels.
```

packages/os/disable-udp-offload.service
```diff
@@ -0,0 +1,22 @@
...
+# bootstrap container if they desired.
+Before=bootstrap-containers@.service
+# Block manual interactions with this service
+RefuseManualStart=true
+RefuseManualStop=true
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/ethtool -K eth0 tx-udp_tnl-segmentation off
+ExecStart=/usr/sbin/ethtool -K eth0 tx-udp_tnl-csum-segmentation off
```
bcressey, line 15: Can you verify that `ethtool -k eth0` on an existing image has "on" values for both these settings?

packages/os/disable-udp-offload.service
```diff
@@ -0,0 +1,22 @@
+# This unit disables UDP offload for the default primary interface (eth0) in
+# VMware.  This avoids an issue where newer kernels (5.10+) and the vmxnet3
+# driver conflict with NSX-T.
+[Unit]
+Description=Disables UDP offload
+After=network-online.target
+# Run before bootstrap containers so a user could reverse the change via
+# bootstrap container if they desired.
+Before=bootstrap-containers@.service
```
bcressey: Is this syntax correct? You're using it to mean "before all units instantiated from this template" but I'm surprised that would be valid.

In any case, running before bootstrap containers is implied by `preconfigured.target`, so this isn't necessary and should be omitted.

packages/os/os.spec
```diff
@@ -591,4 +596,8 @@ install -p -m 0644 %{S:301} %{buildroot}%{_cross_udevrulesdir}/81-ebs-volumes.ru
 %{_cross_unitdir}/bootstrap-containers@.service
 %{_cross_tmpfilesdir}/bootstrap-containers.conf
 
+%if %{with vmware_platform}
+%{_cross_unitdir}/disable-udp-offload.service
+%endif
```
bcressey: This places the file in the `bootstrap-containers` subpackage, which doesn't seem correct - it's unrelated to bootstrap containers.

I don't think there's an amazing spot for it in the existing subpackages. Maybe netdog?

packages/os/disable-udp-offload.service
```diff
@@ -0,0 +1,22 @@
...
+# Block manual interactions with this service
+RefuseManualStart=true
+RefuseManualStop=true
```
bcressey, line 6:
```suggestion
# Block manual interactions with this service
RefuseManualStart=true
RefuseManualStop=true

# (move the comment from the top about VMware specifics here)
ConditionVirtualization=vmware
```

packages/os/disable-udp-offload.service
zmrow: Ah - you're absolutely right - thanks!

packages/os/disable-udp-offload.service
zmrow: Oh nice - good find!

packages/os/os.spec
zmrow: :facepalm:   I overlooked that - I'll get it up in the netdog subpackage.

packages/os/disable-udp-offload.service
zmrow: Yep - I booted a VMware image without the systemd unit and verified the settings show as "on".  Updated the PR description as well:
```
bash-5.1# ethtool -k eth0 | grep -i tx-udp
tx-udp_tnl-segmentation: on
tx-udp_tnl-csum-segmentation: on
```

zmrow:

bcressey (approved):

PR #2876, opened 2023-03-10
kubernetes-1.26: fix cloud provider condition

gthao313:
**Issue number:**

Closes #

**Description of changes:**
```
Author: Tianhao Geng <tianhg@amazon.com>
Date:   Fri Mar 10 10:09:08 2023 +0000

    kubernetes-1.26: fix cloud provider condition

    Bottlerocket settings model for cloud_provider converts the empty string to
    "\"\"", so we should improve the if condition to eq to it.
```

**Testing done:**

- [x] - `aws-*` conformance test
- [x] - `metal-*` conformance test
- [x] - `vmware-*` conformance test

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.26/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.26/kubelet-exec-start-conf
@@ -2,7 +2,7 @@
 ExecStart=
 ExecStart=/usr/bin/kubelet \
 {{#unless settings.kubernetes.standalone-mode}}
-{{#if (eq settings.kubernetes.cloud-provider "")}}
+{{#if (eq settings.kubernetes.cloud-provider "\"\"")}}
     --cloud-provider "" \
 {{else}}
     --cloud-provider "external" \


packages/kubernetes-1.26/kubelet-exec-start-conf
```diff
@@ -2,10 +2,12 @@
 ExecStart=
 ExecStart=/usr/bin/kubelet \
 {{#unless settings.kubernetes.standalone-mode}}
-{{#if (eq settings.kubernetes.cloud-provider "")}}
-    --cloud-provider "" \
-{{else}}
+{{#if (eq settings.kubernetes.cloud-provider "external")}}
```
stmcginnis: Is there ever a case where the value of `settings.kubernetes.cloud-provider` should be used directly and not set to `"external"` in the generated output? If not, it would be simpler to do:

```txt
{{#if (ne settings.kubernetes.cloud-provider "")}}
    --cloud-provider "external" \
{{else}}
    --cloud-provider "" \
{{/if}}
```

See test here:
https://gist.github.com/stmcginnis/326a239daef49a84916bb96c85bba80c

jpmcb (commented):
Overall, makes sense - I'm just abit confused what the context on this is. Also curious to hear what failures you were seein with this.

Looks like in other Kubernetes packages, we were using a `default "external"` block to always get that. Which is similar to what Sean suggested.
https://github.com/bottlerocket-os/bottlerocket/blob/24e757428ac943548cea58237a03d9cf5260854d/packages/kubernetes-1.25/kubelet-exec-start-conf#L3-L12

Can we just do something similar to what we do in the Kubernetes 1.25 package?

gthao313:
> Can we just do something similar to what we do in the Kubernetes 1.25 package?

@jpmcb  We can't. Ideally we should migrate it to `cloud_provider=external` but metal-* needs empty string as value here so wen can't directly set this to `external` for all. If we still use `{{default "external" settings.kubernetes.cloud-provider}} \` here, for `aws-*`  variants, they default the value to [`aws`](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/shared-defaults/kubernetes-aws.toml#L6) which will break the change here.

packages/kubernetes-1.26/kubelet-exec-start-conf
etungsten: In that case, I believe you can just do this:
```
{{#if settings.kubernetes.cloud-provider }}
  --cloud-provider "external" \
{{else}}
   --cloud-provider "" \
{{/if}}
```

gthao313: > I tried this way and failed on test. I was thinking if this will be true when we set `settings.kubernetes.cloud-provider` to "". 

I think it works as well. 
```
use handlebars::Handlebars;
use std::collections::BTreeMap;

fn main() {
        test_handlebars("");
        test_handlebars("aws");
        test_handlebars("external");
        test_handlebars("foo");
}

fn test_handlebars(cloud_provider: &str) {
    let mut handlebars = Handlebars::new();

    // register the template. The template string will be verified and compiled.
    let source = "
    {{#if cloud-provider }}
    external
    {{else}}
    ''
    {{/if}}";
    assert!(handlebars.register_template_string("t1", source).is_ok());

    let mut data = BTreeMap::new();
    data.insert("cloud-provider".to_string(), cloud_provider.to_string());
    
    println!("cloud-provider: '{}': {}", cloud_provider, handlebars.render("t1", &data).unwrap());
}

```
```
cloud-provider: '':
    ''

cloud-provider: 'aws':
    external

cloud-provider: 'external':
    external

cloud-provider: 'foo':
    external
```

gthao313:
Push above improve the condition

settings model for `cloud_provider` converts the empty string to `\"\"`, so we should improve it to equal to `"\"\""`.
```
// Kubelet expects the empty string to be double-quoted when be passed to `--cloud-provider`
let cloud_provider = if input.is_empty() { "\"\"" } else { input };
```



Test:
launched metal cluster
```
✅ Tinkerbell Provider setup is valid
✅ Validate certificate for registry mirror
✅ Validate authentication for git provider
✅ Create preflight validations pass
Creating new bootstrap cluster
Provider specific pre-capi-install-setup on bootstrap cluster
Installing cluster-api providers on bootstrap cluster
Provider specific post-setup
Creating new workload cluster
Installing networking on workload cluster
Creating EKS-A namespace
Installing cluster-api providers on workload cluster
Installing EKS-A secrets on workload cluster
Installing resources on management cluster
Moving cluster management from bootstrap to workload cluster
Installing EKS-A custom components (CRD and controller) on workload cluster
Installing EKS-D components on workload cluster
Creating EKS-A CRDs instances on workload cluster
Installing GitOps Toolkit on workload cluster
GitOps field not specified, bootstrap flux skipped
Writing cluster config file
Deleting bootstrap cluster
🎉 Cluster created!
```

etungsten (approved):

stmcginnis (approved):

PR #2881, opened 2023-03-10
Add CIS Agent for automated CIS compliance checking

### LINKED ISSUE #2731: Add CIS benchmark agent for easy compliance reporting
**What I'd like:**

Bottlerocket has gone through [CIS benchmarking](https://github.com/bottlerocket-os/bottlerocket/issues/1297), but it was a manual process to collect all necessary data. Compliance can also vary based on configuration.

To make this easier to check for future certification refresh, and so users are able to easily get the status of their unique deployments, it would be useful to have an agent as part of Bottlerocket that could be triggered to perform a series of check and report the results.

Bottlerocket exposes some actions via the API such as `apiclient reboot`. It could be useful to have something like `apiclient cis-benchmark` or `cis-report` that would execute these checks and provide output that could be used to evaluate the compliance. There isn't a standard for this reporting format, but it may be useful to adhere to [localtoast](https://github.com/google/localtoast) or [OpenSCAP](http://www.open-scap.org/) formats for the sake of some consistency.

We can keep this a manual thing, but that is very error prone and very labor intensive to do on a regular basis.

stmcginnis:
**Issue number:**

Closes #2731

**Description of changes:**

This adds tooling to be able to run [Bottlerocket CIS compliance checks](https://www.cisecurity.org/benchmark/bottlerocket).

This is made up of three parts:

- [x] Add command line that can check the various system settings and report the results
- [x] Wire up `apiserver` to be able to call this CLI and get its results
- [x] Wire up `apiclient` as the user interface for someone to get compliance status from a running system

**Testing done:**

- Built and deployed image, verified level 1 and level 2 checks were performed correctly with the `bloodhound` CLI.

Removed from this PR, but just for the sake of showing test coverage, also did:

- Tested `apiserver` endpoint by using `apiclient raw -m GET -u /cis-report` and `apiclient raw -m GET -u /cis-report?level=2&format=json`.
- Test `apiclient` interface.
   - Ran `apiclient` and verified help output contained information about new subcommand.
   - Ran `apiclient cisreport` and verified default level 1 text report was output.
   - Ran `apiclient cisreport -l 2` and verified level 2 report generated.
   - Ran `apiclient cisreport -f json` and verified json report generated. Ran through `jq` to view pretty-print view.
   - Ran `apiclient cisreport -f invalid` and verified error message returned.

**Terms of contribution:**


stmcginnis:
This is already quite large, so I've broken each checker into separate commits to try to make it a little easier to review.

~We may want to consider merging this one first to get the CLI on the system, then have a separate PR for the API plumbing just to keep things simple. For now, marking as draft until we decide on an approach.~

API wiring was relatively small, so just included it here.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+  -f, --format <FORMAT>     Format of the output [default: text] [possible values: text, json]
+  -l, --level <LEVEL>       The CIS benchmark compliance level to check [default: 1]
+  -o, --output <OUTPUT>     Write output to a file at given path [default: stdout]
+  -h, --help                Print help
+  ```
+*/
+
+use bloodhound::output::{JsonReportWriter, ReportWriter, TextReportWriter};
+use bloodhound::results::{CheckStatus, CheckerMetadata, CheckerResult, ReportResults};
+use clap::{Parser, ValueEnum};
```
arnaldo2792: I think we use `argh` for CLIs that we include in the OS, and `clap` for more complex CLIs that aren't part of the OS. Probably somebody else in the team can chime in here.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+        .filter_map(|file| file.ok())
+        .collect();
+    let mut result = HashMap::with_capacity(entries.len());
+
+    for entry in entries {
+        if let Ok(file_metadata) = fs::metadata(entry.path()) {
+            // Filter out any subdirectories
+            if file_metadata.is_dir() {
+                continue;
+            }
```
arnaldo2792, line 75: I feel doubtful about this check.

If we, the core distro developers, own the structure of this directory, we know we won't put directories in the checkers patch. However, OOTB users may do what they please and organize their checks differently e.g., in directories; then this check will ignore their checks without even providing a heads-up saying “FYI, ignoring unexpected directory”.

Thus, I think we either traverse directories that are placed here, or we print a message saying that the directories will be ignored, and document in OOTB that for this feature, `bloodhound` expects actual binaries.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+            if file_metadata.permissions().mode() & 0o111 == 0 {
+                continue;
+            }
+
+            // It's an executable file, make sure it implements our expected checker interface
+            let metadata;
+            if let Ok(output) = Command::new(entry.path()).arg("metadata").output() {
+                metadata = String::from_utf8_lossy(&output.stdout).to_string();
+            } else {
+                eprintln!(
```
arnaldo2792, line 87: Another candidate to be more “consistent” with other CLI-like tools in the base image. We use Snafu to generate errors. Other folks in the team can chime in here as well.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+/// to find the executable files that appear to be bloodhound checkers.
+/// It will parse the checker metadata to provide a mapping between a filesystem
+/// path and a specific check, including only those checkers that are within the
+/// request compliance level.
+fn find_checkers(check_dir: &PathBuf, level: u8) -> HashMap<String, CheckerMetadata> {
+    let entries: Vec<DirEntry> = fs::read_dir(check_dir)
+        .unwrap()
+        .filter_map(|file| file.ok())
+        .collect();
+    let mut result = HashMap::with_capacity(entries.len());
```
arnaldo2792: I think you can skip the initialization of the capacity? If there are many entries in the directory but they aren't for the provided level, the memory will be assigned and never used.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+    }
+
+    result
+}
+
+/// Processes the output results from calling an individual checker. Results of
+/// parsing the checker output are added to the overall execution report.
+fn process_checker_results(output: Output, data: CheckerMetadata, report: &mut ReportResults) {
```
arnaldo2792, line 113: nit: Why not handling this in `add_result`? You will save yourself the extra `mut` reference.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+        Some(ref path) => File::create(path).map(|f| Box::new(f) as Box<dyn Write>),
+        None => Ok(Box::new(stdout())),
+    }
+}
+
+fn main() {
+    let args = Arguments::parse();
+
+    // Discover all checkers in checks directory
+    let check_dir = args.check_dir.unwrap_or_else(|| PathBuf::from("/usr/libexec/bloodhound-checks"));
```
arnaldo2792: `/usr/libexec/bloodhound-checks` should be a const

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+                data,
+                CheckerResult {
+                    status: CheckStatus::SKIP,
+                    error: msg,
+                },
+            );
+        }
+    }
+
+    // Wite appropriate output results report
```
arnaldo2792: nit: Typo

arnaldo2792 (commented):
I'm still going through it but here are a few things that caught my eye.

sources/bloodhound/src/main.rs
stmcginnis: This is very much intentional. We don't want someone to accidentally point `--checks` at something like `/` by mistake and have it recurse through the entire filesystem trying to run every binary it finds. The design calls out that for someone to provide checks, it needs to be a binary in a specific directory.

If OOTB users want to organize their checkers in a different structure for whatever reason, they could easily provide symlinks for which ones they want to include.

We also don't want to emit warning output on this, because at least in some cases the output is something that could be ingested into some other parsing. Emitting warnings for something like this could cause problems doing that.

It is also intentional that subdirectories are ignored. Some future work being looked at would rely on this for handling different scenarios, so it could be common even in Bottlerocket that there would be a directory that should just be silently ignored.

sources/bloodhound/src/main.rs
stmcginnis: Reallocation of memory can be more expensive than having a few unused slots. But in our case we know this will be a small number anyway, so I can drop this.

sources/bloodhound/src/main.rs
stmcginnis: I'm not bubbling up an error here though. The tool needs to print it a failure message, so `eprintln!()` can't really be replaced by something like Snafu.

Also trying to limit the number of dependencies as much as possible so if someone wants to use bloodhound outside of Bottlerocket, it does drag a lot of dependencies along with it too.

sources/bloodhound/src/main.rs
stmcginnis: That wouldn't make sense to pass an `Output` object through to the results since that is a different domain of control and results could be provided from something other than through a `Command.output()` result, especially if I am able to add better unit test coverage. We need to know whether to extract the stdout or stderr.

sources/bloodhound/src/main.rs
stmcginnis: I don't think `argh` adds much value, so switching to plain old arg parsing. ;)

sources/bloodhound/src/main.rs
stmcginnis: Fixed!

stmcginnis:
Improved report information by adding a metadata file to be included in the checkers directory. This gives a little more flexibility if we want to have multiple types of compliance checks by just pointing bloodhound at a different checks directory where it can discover metadata about the checks being performed.

sources/bloodhound/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0"
+authors = ["Sean McGinnis <stmcg@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+chrono = "0.4"
```
jpculp:
```suggestion
chrono = { version = "0.4", default-features = false, features = ["std", "clock"] }
```
You can avoid pulling in the old version of `time` by only calling the `chrono` features you need.

jpculp (dismissed):

sources/bloodhound/Cargo.toml
stmcginnis: Tried it out and looks like all we really need is `clock`. Nice clean up of transient dependencies - great suggestion!

stmcginnis:
Force pushed a refactoring of the code to make it better organized.

stmcginnis:
And a couple more force-pushes to get rid of commits that are already in `develop` but GitHub somehow thinks they are new. :/

sources/api/apiclient/README.md
```diff
@@ -192,8 +192,8 @@ For example, if you want the name "FOO", you can `PATCH` to `/settings?tx=FOO` a
 ## apiclient library
 
 The apiclient library provides high-level methods to interact with the Bottlerocket API.  See
-the documentation for submodules [`apply`], [`exec`], [`get`], [`reboot`], [`set`], and
-[`update`] for high-level helpers.
+the documentation for submodules [`apply`], [`cisreport`], [`exec`], [`get`], [`reboot`], [`set`],
+[`update`], for high-level helpers.
```
bcressey: This is ultimately something we could address later, but I wonder if something like `apiclient report` would feel less specialized for a top-level sub-command:
```
apiclient report cis --level 1 --format json
apiclient report cis-k8s --level 2 --format text

# some other, future kinds of reports
apiclient report boot-metrics --format json

# other ideas for reports
# these might be better as `history` sub-commands
# since they introspect the API instead of examining the host
apiclient history changed-settings
apiclient history executed-commands
```

`report` could also be `benchmark` but that invites confusion with performance-related benchmarks instead of compliance-related ones.

sources/api/apiclient/src/cisreport.rs
```diff
@@ -0,0 +1,63 @@
+use snafu::ResultExt;
+use std::path::Path;
+
+/// Requests a CIS complaince report through the API.
```
bcressey: nit:
```suggestion
/// Requests a CIS compliance report through the API.
```

sources/api/apiserver/src/server/mod.rs
```diff
@@ -543,6 +545,39 @@ async fn reboot() -> Result<HttpResponse> {
     Ok(HttpResponse::NoContent().finish())
 }
 
+/// Triggers a bloodhound check for compliance and returns the results. Defaults
+/// to level 1 text report unless specified otherwise.
+async fn get_cis_benchmark(query: web::Query<HashMap<String, String>>) -> Result<HttpResponse> {
+    info!("Running CIS benchmark");
+
+    let mut cmd = AsyncCommand::new("/usr/bin/bloodhound");
```
bcressey: For robustness, I tend to want to model this along the same lines as the apiclient -> apiserver -> thar-be-updates -> updog flow.

I have misgivings about mixing tokio and actix; actix uses tokio under the hood but the [api exec](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/api-exec.md#resource-management) docs warn that the async approach is very different. This is why `apiserver` did not previously depend directly on tokio.

Among other things, this doesn't throttle or lock concurrent requests so someone could drive load on the host by repeatedly generating reports. It also assumes all the helper binaries are robust for overlapping execution. I am wary of trying to add locks at the apiserver level; IIRC we attempted that with updates and ended up with `thar-be-updates` as the simpler solution.

Splitting up the request and the response might help, e.g.
```
# submit a request for a report type, format specified by parameters
GET /actions/report

# retrieve the results
GET /reports/{{ report_id }}
```

sources/bloodhound/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Checks are performed and their results are provided in an overall report.
+The checker report can be written to a file, or viewed from stdout.
+By default the report is provided in a human readable text format, but can also
+be generated as JSON to make it easy to consume programmatically for integrating
+into further compliance automation.
+
+## Usage
+
+Bloodhound is ultimately intended to be used throught he Bottlerocket `apiclient`
```
bcressey: nit:
```suggestion
Bloodhound is ultimately intended to be used through the Bottlerocket `apiclient`
```

sources/bloodhound/src/args.rs
```diff
@@ -0,0 +1,115 @@
...
+pub struct Arguments {
+    /// Path to the directory containing checker binaries [default: {DEFAULT_CHECK_PATH}]
+    pub check_dir: Option<PathBuf>,
+    /// Format of the output
+    pub format: Format,
+    /// The CIS benchmark compliance level to check
+    pub level: u8,
+    /// Write output to a file at given path [default: stdout]
+    pub output: Option<PathBuf>,
+}
```
bcressey: I would really prefer that we use `argh` for this, for consistency with the newer CLI tools in the host.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+Checks are performed and their results are provided in an overall report.
+The checker report can be written to a file, or viewed from stdout.
+By default the report is provided in a human readable text format, but can also
+be generated as JSON to make it easy to consume programmatically for integrating
+into further compliance automation.
+
+# Usage
+
+Bloodhound is ultimately intended to be used throught he Bottlerocket `apiclient`
+interface. It can be executed from sheltie with the following options:
```
bcressey: I would prefer not to mention `sheltie` in this context; it's not really part of the host.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+
+```txt
+Command line arguments for the bloodhound program
+
+Usage: bloodhound [OPTIONS]
+
+Options:
```
bcressey: We should come up with some way of auto-generating this material if we want to include it in `README.md`. Otherwise it's very likely to drift out of sync.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,1280 @@
...
+const SYSTEMCTL_CMD: &str = "/usr/bin/systemctl";
+const MODPROBE_CMD: &str = "/bin/modprobe";
+const SESTATUS_CMD: &str = "/usr/bin/sestatus";
+const APICLIENT_CMD: &str = "/usr/bin/apiclient";
+const IPTABLES_CMD: &str = "/usr/sbin/iptables";
+const IP6TABLES_CMD: &str = "/usr/sbin/ip6tables";
+
+// =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<=
+
+pub struct BR01010101Checker {}
```
bcressey, line 21: It's more a matter of where the boilerplate goes rather than eliminating it altogether, but I would be tempted to add an enum to capture all these individual structs:
```
pub enum BRChecker {
  BR01010101(BR01010101Checker),
  ....
}
```

Then you can derive `Deserialize` for the enum and use `serde_plain` in `main()` to turn the program name into an enum variant to arrive at the struct. That avoids replicating the somewhat cryptic struct outside of this file. It might be easier if all the `ManualCheckers` were unique structs for consistency.

You could also potentially find a crate or some other solution to iterate over all variants in the enum and list them out, then call that function in `build.rs` to generate a file that can be used in the rpm package to create the expected symlinks.

Potentially the `metadata.json` could also be generated in the same way, from a method on the enum that's written to a file by the build script. That would tie it more closely to these checks, since right now it's pretty decoupled.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,1280 @@
...
+
+const PROC_MODULES_FILE: &str = "/proc/modules";
+const PROC_CMDLINE_FILE: &str = "/proc/cmdline";
+const LOCKDOWN_FILE: &str = "/sys/kernel/security/lockdown";
+const JOURNALD_CONF_FILE: &str = "/usr/lib/systemd/journald.conf.d/journald.conf";
+const SYSCTL_CMD: &str = "/usr/sbin/sysctl";
+const SYSTEMCTL_CMD: &str = "/usr/bin/systemctl";
+const MODPROBE_CMD: &str = "/bin/modprobe";
+const SESTATUS_CMD: &str = "/usr/bin/sestatus";
+const APICLIENT_CMD: &str = "/usr/bin/apiclient";
```
bcressey: This is kind of a scary request flow, calling `apiclient` from inside an `apiserver` request. It is not quite what the benchmark says, but we could also check `/etc/chrony.conf` since we're running inside the host.

sources/bloodhound/src/results.rs
```diff
@@ -0,0 +1,182 @@
...
+/// Used to help serialize output into simpler JSON structure.
+#[derive(Debug, Serialize)]
+pub struct IndividualResult {
+    #[serde(flatten)]
+    pub metadata: CheckerMetadata,
+    #[serde(flatten)]
+    pub result: CheckerResult,
+}
+
+/// ReportResults are the overall complaince checking containing the results of
```
bcressey:
```suggestion
/// ReportResults are the overall compliance checking containing the results of
```

sources/bloodhound/src/results.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub level: u8,
+    pub total: usize,
+    pub passed: usize,
+    pub skipped: usize,
+    pub failed: usize,
+    pub status: CheckStatus,
+    pub timestamp: String,
+    #[serde(flatten)]
+    pub metadata: ReportMetadata,
+    pub results: BTreeMap<String, IndividualResult>,
```
bcressey, line 140: Do all these fields need to be `pub`? It seems like you provide methods for creating and modifying fields in a controlled way.

Generally I read `pub` as a signal that other modules or crates will need to directly manipulate these fields - so it's better to use it only where needed.

sources/bloodhound/src/output.rs
```diff
@@ -0,0 +1,58 @@
...
+
+impl ReportWriter for TextReportWriter {
+    /// Writes a text formatted report to the provided output destination.
+    fn write(&self, report: &ReportResults, output: &mut dyn Write) -> Result<(), Error> {
+        if let Some(name) = &report.metadata.name {
+            writeln!(output, "Benchmark name:  {}", name)?;
+        }
+        if let Some(version) = &report.metadata.version {
+            writeln!(output, "Version:         {}", version)?;
+        }
```
bcressey: It might be more maintainable to use the fill/alignment options rather than manually aligning via spaces.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+            let metadata;
+            if let Ok(output) = Command::new(entry.path()).arg("metadata").output() {
+                metadata = String::from_utf8_lossy(&output.stdout).to_string();
+            } else {
+                eprintln!(
+                    "{:?} does not appear to be a checker executable",
+                    entry.path()
+                );
+                continue;
+            }
```
bcressey, line 92: It seems like we should only treat it as a checker executable if:
* the command returns non-zero status
* the command produces output on stdout
* the output deserializes into the expected structure

This check would only fail if the program itself can't be executed for some reason.

It might be better to factor all the logic into its own helper function that returns `Option<CheckerMetadata>`.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    let check_output = String::from_utf8_lossy(&output.stderr).to_string();
+    report.add_result(
+        data,
+        CheckerResult {
+            status: CheckStatus::SKIP,
+            error: check_output,
+        },
+    );
```
bcressey, line 129: Why not also expect checkers to emit fully-formed output on stderr? It's a bit odd that the checker handles the `PASS` and `FAIL` cases, and provides the error for the `SKIP` case but not the status.

That would simplify the logic here:
* if success, output is on stdout
* otherwise, output is on stderr
* deserialize the chosen output from json and add to report

sources/bloodhound/src/lib.rs
```diff
@@ -0,0 +1,130 @@
+use std::{ffi::OsStr, fs::File, io::BufRead, io::BufReader, process::Command};
+
+pub mod args;
+pub mod output;
+pub mod results;
+
+/// Reads a file and checks if the given `search_str` is present in its contents.
+pub fn look_for_string_in_file(path: &str, search_str: &str) -> Option<bool> {
+    if let Ok(modules) = File::open(path) {
```
bcressey: It is unusual to swallow the potential I/O error here, and the potential execution error in `look_for_string_in_output`, which is part of why this function reads a bit "off" - it doesn't benefit from `?` or more condensed syntax like:
```
let reader = BufReader::new(File::open(path)?;
reader.lines().any(|line| line == search_str)
```

It's also prone to misuse or confusion. A check may interpret a `None` result as "I didn't find this problematic string in the config file" and report success, when a skip result might be more appropriate ("I couldn't run this check because the file didn't exist").

I would recommend returning `Result<Option<bool>>` and just dealing with the three cases (error / some / none) in the caller.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,1280 @@
...
+
+        result
+    }
+
+    fn metadata(&self) -> results::CheckerMetadata {
+        results::CheckerMetadata {
+            title: "Ensure dm-verity is configured".to_string(),
+            id: "1.3.1".to_string(),
+            level: 1,
+            name: "br01030100".to_string(),
```
bcressey, line 77: Would it be worth naming these checks something like `br_1_3_1` rather than using zeroes as placeholders?

You could also use a proc macro to derive a method that would turn the struct name into the ID and the name.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,1280 @@
...
+        // Trying to avoid bringing in regex for now
+        let to_match = &vec![
+            ("SELinux status: ", " enabled"),
+            ("Loaded policy name: ", " fortified"),
+            ("Current mode: ", " enforcing"),
+            ("Mode from config file: ", " enforcing"),
+            ("Policy MLS status: ", " enabled"),
+            ("Policy deny_unknown status: ", " denied"),
+            ("Memory protection checking: ", " actual (secure)"),
+        ];
```
bcressey: You could potentially have two versions of your `check_for` helper functions, one that takes a single line that must be in the output, and one that takes multiple lines, and have the single line one call into the multi-line function.

This would also be useful in the dm-verity checker, and in the `sysctl` checker, since you can query multiple sysctls at once:
```
❯ sysctl net.ipv4.conf.all.send_redirects net.ipv4.conf.default.send_redirects
net.ipv4.conf.all.send_redirects = 0
net.ipv4.conf.default.send_redirects = 1
```

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,1280 @@
...
+
+                if to_match.len() == matched {
+                    result.status = results::CheckStatus::PASS;
+                } else {
+                    result.error = "Unable to find expected SELinux values".to_string();
+                    result.status = results::CheckStatus::FAIL;
+                }
+            }
+        } else {
+            result.error = "unable to verify selinx settings".to_string();
```
bcressey:
```suggestion
            result.error = "unable to verify SELinux settings".to_string();
```

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,1280 @@
...
+        if let Some(found) = look_for_string_in_file(JOURNALD_CONF_FILE, "Storage=persistent") {
+            if !found {
+                result.error = "journald is not configured".to_string();
+                result.status = results::CheckStatus::FAIL;
+            } else {
+                result.status = results::CheckStatus::PASS;
+            }
+        } else {
+            result.error = "unable to verify journald settings".to_string();
+        }
```
bcressey: This same sort of pattern is repeated so often that it might be worth turning into a macro.
```
check_file_contains!(path, strings_to_match, unable_to_check_error, did_not_find_error)

check_file_contains!(JOURNALD_CONF_FILE, "Storage=persistent", "unable to verify journald settings", "journald is not configured")
```

Then most of these checks would read like asserts in unit tests.

sources/api/apiserver/src/server/mod.rs
stmcginnis: > I have misgivings about mixing tokio and actix

This should be OK. They actually recommend using `tokio` in the docs: https://actix.rs/docs/server/#multi-threading

sources/bloodhound/src/main.rs
stmcginnis: Line 106 handles the second and third bullet items. I can move this into a separate helper function, but I wanted to have it here since it is a small amount of logic and easier to read through and see what's happening.

sources/bloodhound/src/main.rs
stmcginnis: The checker contract is either it is able to perform its check and report the results (stdout) or something failed running the checker (stderr). What would be the fully-formed output you would expect from stderr?

sources/bloodhound/src/results.rs
stmcginnis: Yes, this is needed for the serialization to the output. This is all used to collect the overall report results, in a very specific way, with only some helper functions below to eliminate some extra typing.

stmcginnis:
Removed `apiclient` and `apiserver` commits for now to keep this smaller and to work through some of the suggestions there.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
stmcginnis: The name needs the zero padding to allow for easy human expected sorting of the results. But a macro may help with some of this boilerplate. I'll look into that.

stmcginnis: I couldn't think of a good way to improve on this with a macro, but happy to be told otherwise.

The problem I see is with needing the title, ID #, level, and prefix (e.g. `br`), having a macro that would take the necessary input wouldn't be any more concise than just having it stated here, and there would be the overhead of the macro expansion and any manipulation of the ID and prefix to emit the desired field values.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -1,7 +1,11 @@
+// #[macro_use]
```
stmcginnis: Clean up staged, waiting for any other changes before pushing up.

sources/bloodhound/src/output.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+
+pub struct JsonReportWriter {}
+
+impl ReportWriter for JsonReportWriter {
+    /// Writes a json formatted report to the provided output destination.
+    fn write(&self, report: &ReportResults, output: &mut dyn Write) -> Result<(), Error> {
+        let json = serde_json::to_string(&report).unwrap_or_default();
+        writeln!(output, "{}", json)
```
arnaldo2792, line 56: Shouldn't this function fail if the `report` fails to deserialize?

packages/os/Cargo.toml
```diff
@@ -21,6 +21,7 @@ source-groups = [
     "shimpei",
     "driverdog",
     "cfsignal",
+    "bloodhound",
```
arnaldo2792, line 24: nit: I think this commit can be squashed with the commit that introduced `bloodhound`, since the new helper was added to all variants.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
arnaldo2792: nit: and probably a comment to describe what each checker validates (like in the commit message).

sources/bloodhound/src/lib.rs
```diff
@@ -13,7 +13,57 @@ pub fn look_for_string_in_file(path: &str, search_str: &str) -> Result<bool, io:
...
+pub fn look_for_strings_in_file(path: &str, search_strs: &[&str]) -> Result<bool, io::Error> {
+    let mut matched = 0;
+
+    let reader = BufReader::new(File::open(path)?);
+    for line in reader.lines() {
+        let content = line.unwrap_or_default();
+        for search_str in search_strs {
+            if content.contains(search_str) {
+                matched += 1;
+            }
```
arnaldo2792, line 26: You could return earlier if an expected string isn't found.

sources/bloodhound/src/lib.rs
```diff
@@ -13,7 +13,57 @@ pub fn look_for_string_in_file(path: &str, search_str: &str) -> Result<bool, io:
...
+
+        if let Ok(found) = look_for_strings_in_file($path, $strings_to_match) {
+            if found {
+                result.status = results::CheckStatus::PASS;
+            } else {
+                result.error = $unable_to_find_error.to_string();
+                result.status = results::CheckStatus::FAIL;
+            }
+        } else {
+            result.error = $unable_to_check_error.to_string();
```
arnaldo2792, line 56: Isn't `result.status = results::CheckStatus::Fail` on this branch?

sources/bloodhound/src/lib.rs
```diff
@@ -13,7 +13,57 @@ pub fn look_for_string_in_file(path: &str, search_str: &str) -> Result<bool, io:
...
+/// If all `strings_to_match` are found in `path`, the `CheckerResult` returned will have a `CheckStatus::PASS`.
+///
+/// If one or more of `strings_to_match` are not found in `path`, then even if some are found the `CheckerResult`
+/// returned will have a `CheckStatus::FAIL` and the `error` field will contain `unable_to_find_error` as its content.
+///
+/// If `path` cannot be read or there is some other error checking the content of the file, returned status will be
+/// `CheckStatus::SKIP` indicating a manual check will need to be performed and the `error` field will contain the
+/// `unable_to_check_error` value.
+#[macro_export]
+macro_rules! check_file_contains {
```
arnaldo2792, line 44: I wonder if this is a macro is a bit overkill here, what's the benefit of having this function as a macro instead of just a plain function?

sources/bloodhound/src/lib.rs
```diff
@@ -86,6 +86,66 @@ where
...
+///
+/// If all `strings_to_match` are found in the output, the `CheckerResult` returned will have a `CheckStatus::PASS`.
+///
+/// If one or more of `strings_to_match` are not found in the output, then even if some are found the `CheckerResult`
+/// returned will have a `CheckStatus::FAIL` and the `error` field will contain `unable_to_find_error` as its content.
+///
+/// If execution fails or the output cannot be read, returned status will be `CheckStatus::SKIP` indicating a manual
+/// check will need to be performed and the `error` field will contain the `unable_to_check_error` value.
+#[macro_export]
+macro_rules! check_output_contains {
```
arnaldo2792, line 111: Same as elsewhere, what's the benefit of a macro over a function?

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -180,3 +180,29 @@ impl results::Checker for BR01040300Checker {
...
+
+// =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<=
+
+pub struct BR01040400Checker {}
+
+impl results::Checker for BR01040400Checker {
+    fn execute(&self) -> results::CheckerResult {
+        check_output_contains!(
+            SYSCTL_CMD,
+            ["user.max_user_namespaces"],
```
arnaldo2792, line 169: hmm, is this a check in the CIS benchmark that must pass? It will be fun to revisit this once this feature moves to stable:

https://kubernetes.io/blog/2022/10/03/userns-alpha/

sources/bloodhound/src/lib.rs
```diff
@@ -13,7 +13,7 @@ pub fn look_for_string_in_file(path: &str, search_str: &str) -> Result<bool, io:
         .any(|line| line.unwrap_or_default().contains(search_str)))
 }
 
-/// Reads a file and checks if the given `search_str` is present in its contents.
+/// Reads a file and checks if the given `search_strs` are present in its contents.
```
arnaldo2792, line 16: nit: seems like the updates in this file were included in the wrong commit.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -672,3 +672,82 @@ impl results::Checker for BR03040101Checker {
...
+
+            if first_found && second_found {
+                result.status = results::CheckStatus::PASS;
+            } else {
+                result.error = "Unable to find expected iptables INPUT values".to_string();
+                result.status = results::CheckStatus::FAIL;
+                return result;
+            }
+        } else {
+            result.error = "unable to verify iptables INPUT settings".to_string();
```
arnaldo2792, line 689: Isn't a `results::CheckStatus::Fail` missing?

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -784,3 +784,81 @@ impl results::Checker for BR03040201Checker {
...
```
arnaldo2792, line 796: Probably worth looking at the `else` branch of all the checks, seems like a few are missing update the result to `results::CheckStatus::Fail`.

packages/os/Cargo.toml
stmcginnis: I'm keeping it separate to have very specific steps that can be reviewed.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
stmcginnis: > a comment to describe what each checker validates

I believe the metadata content covers that, right? I wouldn't want to duplicate much more from the spec document.

sources/bloodhound/src/lib.rs
stmcginnis: I'm not seeing a good way to exit early here. We need to go through all lines to see if the strings are present. And we need to check each string on each line because the one line could contain multiple substrings that we are trying to match.

If you have something in mind that I'm missing, can you add some pseudo code to describe it?

sources/bloodhound/src/lib.rs
stmcginnis: I think either a macro or a function could work here. @bcressey had asked for a macro, so maybe he has some opinions on using one vs the other. But I do think it makes the usage nice and concise.

sources/bloodhound/src/lib.rs
stmcginnis: No, in this branch we got a result back that we were not able to perform the check, not that the check was done and the expected value was missing. So since we had an error performing the check, we don't know whether it is right or not and would not want to report it as failing.

Basically, the expectation is:

`PASS`: we checked if the state matches the spec and it does
`FAIL`: we checked if the state matches the spec and it does not
`SKIP`: we were unable to verify the state and user must perform manual verification to determine if system is compliant.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
stmcginnis: Yeah, I think basically it means that folks requiring level 2 compliance will not be able to take advantage of user namespaces. Along with a few other things. Level 2 compliance is definitely more restrictive, so I would guess folks will not use it unless they are required.

sources/bloodhound/src/lib.rs
stmcginnis: This was a typo fix, and below was an evolution of the function based on the needs of this particular check.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
stmcginnis: No, like previous comment, in this case we were unable to verify, so it stays with the default of `SKIP`.

sources/bloodhound/src/output.rs
stmcginnis: Yeah, probably better to error out here if something goes really wrong.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
stmcginnis: I'm noticing now I never responded to the original enum comment. Sorry, I thought I had done that.

I spent some time trying different things, but I couldn't get that to work. It might be that I'm just not strong enough with Rust still, but I don't think the checker structs can be used that way? I would get this error when trying to deserialize them from an enum:

```
"invalid type: unit variant, expected newtype variant"
```

There was also a lot of boilerplate code needed to implement `PartialEq` for each checker. I couldn't get it happy with adding `eq()` to the `Checker` trait, and the default `derive` wouldn't work.

If you have some quick POC code I can look at, or an example of a similar implementation being done elsewhere, please do share. I could also put up what I've tried in a separate branch or in a gist if that helps to see if I'm doing something wrong here.

sources/bloodhound/src/output.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn write(&self, report: &ReportResults, output: &mut dyn Write) -> Result<(), Error>;
+}
+
+pub struct TextReportWriter {}
+
+impl ReportWriter for TextReportWriter {
+    /// Writes a text formatted report to the provided output destination.
+    fn write(&self, report: &ReportResults, output: &mut dyn Write) -> Result<(), Error> {
+        if let Some(name) = &report.metadata.name {
+            writeln!(output, "{:17}{}", "Benchmark name:", name)?;
```
yeazelm, line 17: nit: I'm not sure if this is possible, but could you move the padding count out to a constant that could be bumped once?

sources/bloodhound/src/results.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub name: String,
+    pub id: String,
+    pub title: String,
+    pub level: u8,
+}
+
+impl Checker for ManualChecker {
+    fn execute(&self) -> CheckerResult {
+        CheckerResult {
+            error: "Manual check, see benchmark for audit details.".to_string(),
```
arnaldo2792, line 102: The `ManualChecker` message is confusing. Does `Manual` mean that the user has to manually verify ID in the benchmark doc? Or where are the audit details?

sources/bloodhound/src/lib.rs
```diff
@@ -1,3 +1,123 @@
...
+/// Reads a file and checks if the given `search_str` is present in its contents.
+pub fn look_for_string_in_file(path: &str, search_str: &str) -> Result<bool, io::Error> {
+    let reader = BufReader::new(File::open(path)?);
+    Ok(reader
+        .lines()
+        .any(|line| line.unwrap_or_default().contains(search_str)))
+}
+
+/// Executes a command and checks if the give `search_str` is in the output.
+pub fn look_for_string_in_output<I, S>(cmd: &str, args: I, search_str: &str) -> Option<bool>
```
arnaldo2792, line 64: Wouldn't it be simpler to execute the command in `execute`, and let each checker decide how to treat non-zero exit codes (I'm assuming there will be commands that exit non-zero, but that's expected for some checkers)?. Otherwise, in here the function considers non-zero exit codes as “couldn't find the string” but there might be cases where the string will be found on the stderr.

Also, elsewhere there is a function to check for many strings, instead of just one, I think you can get rid of this function and just use the other with the one string passed to it.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -371,3 +371,43 @@ impl results::Checker for BR03010100Checker {
...
+pub struct BR03020100Checker {}
+
+impl results::Checker for BR03020100Checker {
+    fn execute(&self) -> results::CheckerResult {
+        let settings = vec![
+            "net.ipv4.conf.all.accept_source_route",
+            "net.ipv4.conf.default.accept_source_route",
+            "net.ipv6.conf.all.accept_source_route",
+            "net.ipv6.conf.default.accept_source_route",
+        ];
```
arnaldo2792: Can you avoid the `vec!` macro here? (and in similar places). It seems like it is possible to do so since there are other places in this PR where a `[]` with a single element was passed to `check_output_contains!`

sources/bloodhound/src/lib.rs
```diff
@@ -0,0 +1,333 @@
...
+        } else {
+            None
+        }
+    } else {
+        None
+    }
+}
+
+/// Executes a command and checks if the given `search_strs` are in the output.
+pub fn look_for_strings_in_output<I, S>(cmd: &str, args: I, search_strs: &[&str]) -> Option<bool>
```
arnaldo2792, line 74: Why do callers to of this function have to give ownership of `args`? It seems like the function could receive a reference to `args`.

stmcginnis: It could - can you explain why we would want to avoid passing ownership of the args here though?

sources/bloodhound/src/lib.rs
stmcginnis: This is really a convenience to library consumers to have a simple interface to call. I've updated the function though to just delegate to the `_strings_` function.

As far as non-zero exit codes, that shouldn't be a need for other checkers. The commands should be expected to run, it's the `Option<bool>` that is of interest. So if a check needs to make sure a string value does _not_ exist, then it can check that bool result as appropriate.

If a check does end up needing some odd handling (which I haven't seen yet), then it can duplicate this code as a worst case scenario. But I definitely would not want to repeat this code in every `execute` block if it's just repetition.

sources/bloodhound/src/results.rs
stmcginnis: `Manual` is a term used in the CIS benchmarks, so it should be well known to someone interested in benchmark compliance. It basically means "the compliance state of this item cannot be automatically determined and you will need to perform manual verification auditing to evaluate it yourself." The audit details are in the benchmark and we do not want to duplicate any of that here.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
stmcginnis: Done!

sources/bloodhound/README.md
```diff
@@ -0,0 +1,25 @@
+# bloodhound
+
+Current version: 0.1.0
+
+## Introduction
+
+Bloodhound is a command line orchestrator for running a set of compliance
+checks. This can be used to run CIS benchmark compliance, though can be extended
```
bcressey: nit:
```suggestion
checks. This can be used to run CIS benchmark compliance, though it can be extended
```

sources/bloodhound/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+    use tempfile::NamedTempFile;
+
+    use super::*;
+
+    #[test]
+    fn test_string_in_file_found() {
+        let mut test_file = NamedTempFile::new().unwrap();
+        writeln!(test_file, "udf 139264 0 - Live 0xffffffffc05e1000").unwrap();
+        writeln!(test_file, "crc_itu_t 16384 1 udf, Live 0xffffffffc05dc000").unwrap();
+        writeln!(test_file, "configfs 57344 1 - Live 0xffffffffc0320000").unwrap();
```
bcressey: Nice tests!

nit: could do something like this to line up the file contents:
```suggestion
        writeln!(test_file, concat!(
            "udf 139264 0 - Live 0xffffffffc05e1000\n",
            "crc_itu_t 16384 1 udf, Live 0xffffffffc05dc000\n",
            "configfs 57344 1 - Live 0xffffffffc0320000\n")
        )).unwrap();
```

sources/bloodhound/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+
+        let found = look_for_string_in_file(
+            test_file
+                .into_temp_path()
+                .canonicalize()
+                .unwrap()
+                .to_str()
+                .unwrap(),
```
bcressey: nit: you may be tired of macro suggestions but:
```
let found = look_for_string_in_file(temp_file_path!(test_file), " udf,").unwrap();
```

sources/bloodhound/src/lib.rs
arnaldo2792: If I consume this library, I can call this function with my parameters. But, in my use case I may want to use `args` after I call this function, but since I gave ownership of `args` I can't. What I would end up doing is `args.clone()` and pass that to keep my original `args` and don't lose ownership.

I guess TL;DR will be: don't take ownership if you don't absolutely need it.

sources/bloodhound/src/args.rs
```diff
@@ -0,0 +1,51 @@
...
+        from_str_fn(str_to_format),
+        short = 'f'
+    )]
+    pub format: Format,
+    /// the CIS benchmark compliance level to check
+    #[argh(option, default = "1", short = 'l')]
+    pub level: u8,
+    /// write output to a file at given path [default: stdout]
+    #[argh(option, default = "\"\".to_string()", short = 'o')]
+    pub output: String,
```
bcressey: Can `output` be an `Option<String>` with no default? Then we could use the more idiomatic check (`Some` or `None`) in main's `get_output()`, rather than `output.is_empty()`.

sources/bloodhound/src/lib.rs
stmcginnis: There are no cases I have seen where a consumer would need to still own those args. But unfortunately, due to `Command::args()`'s signature, it wouldn't do us much good anyway. We would need to copy those values over to a new Iterator. That would be worse that requiring a potential future library consumer to have to redeclare command arguments for a second use.

sources/bloodhound/src/main.rs
```diff
@@ -0,0 +1,199 @@
...
+        } else {
+            // Something failed in execution, mark as skipped with message
+            let msg = format!("Error executing {} checker.", data.name);
+            report.add_result(
+                data,
+                CheckerResult {
+                    status: CheckStatus::SKIP,
+                    error: msg,
+                },
+            );
```
bcressey, line 172: Part of what I had in my mind with my earlier feedback on `process_checker_results()` came from this spot. "SKIP" sounds like a result that we might want the checker to be able to emit. For example, a particular check might not apply on a given architecture or kernel version, and it might be valuable to distinguish that case from "bloodhound couldn't run the checker at all".

So we might want an `ERROR` or `WARN` check status that only `bloodhound` sets, in these cases:
* the checker can't be executed
* the checker executes, but returns an error
* the checker executes successfully, but its output can't be deserialized into the right shape

Using `CheckStatus::SKIP` for three different purposes - error handling and reporting in `bloodhound`, and also for "executable not found, check skipped" successful results from checkers, and also for "something went wrong and no checks ran at all" - makes the overall error handling approach hard for me to follow.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,935 @@
+use bloodhound::results;
```
bcressey:
```suggestion
use bloodhound::results::{self, Checker, CheckStatus, CheckerMetadata};
```

That would cut out some boilerplate but still let you use `result::Mode` or whatever in places where the type might be ambiguous.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,935 @@
...
+// =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<=
+
+pub struct BR01010101Checker {}
+
+impl results::Checker for BR01010101Checker {
+    fn execute(&self) -> results::CheckerResult {
+        let mut result = results::CheckerResult {
+            error: String::new(),
+            status: results::CheckStatus::SKIP,
+        };
```
bcressey: If you derive or implement `Default` for `CheckerResult`, you can do:
```
let mut result = CheckerResult::default();
```

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,935 @@
...
+            mode: results::Mode::Automatic,
+        }
+    }
+}
+
+/// Parses the loaded modules to see if the udf module is already loaded.
+fn check_udf_loaded(result: &mut results::CheckerResult) {
+    if result.status == results::CheckStatus::FAIL {
+        return;
+    }
```
bcressey: nit: it would be more consistent with other checkers to do this in the execute method for `BR01010101Checker`

bcressey (approved):
Apologies for the protracted wait on my follow-up review. Everything looks good to me; my only non-nitpick is the overloaded uses `CheckStatus::SKIP`, but I don't want to block on that.

sources/bloodhound/README.md
stmcginnis: Done

sources/bloodhound/src/args.rs
stmcginnis: This does open the chance that someone could call `bloodhound -o ""`, but since we don't expect this to be run directly, that should be fine.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
stmcginnis: Will move back, was just trying to make it more granular.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
stmcginnis: Good call!

sources/bloodhound/src/lib.rs
stmcginnis: I like it!

sources/bloodhound/src/main.rs
stmcginnis: What would the user do differently in these cases? We know internally that there are different cases that can lead to the current `SKIP` status, but that's not something the end user would know or care about. From their perspective, either bloodhound can tell them if the system is compliant or not, or they need to do some manual verification to make that determination on their own.

The idea of something being only applicable on a given architecture or kernel version is interesting. We don't currently have something like that, but if something were to be introduced in the future, I could see it being handled in a few different ways:

- Have the compilation take in to account the architecture it is being compiled for and include/exclude appropriately
- Default a non-applicable platform/check combination to `PASS` since to the end user, there is no failure and there is nothing to manually check
- Introduce another test state if there is a need to report this non-applicable state

I think the second one is probably the UX we would want, but until we have one of those cases I don't think we have the full context to make that decision right now. So I would rather defer until there is a need before trying to proactively handle it.

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -0,0 +1,905 @@
...
+                result.status = CheckStatus::FAIL;
+            } else {
+                result.status = CheckStatus::PASS;
+            }
+        } else {
+            result.error = "unable to parse modprobe output to check if udf is enabled".to_string();
+        }
+
+        result
+    }
```
bcressey: Is this equivalent?

```suggestion
    fn execute(&self) -> CheckerResult {
        let module_result = check_file_contains!(
            PROC_MODULES_FILE,
            &[" udf,"],
            "unable to parse modules to check for udf",
            "udf is currently loaded"
        );
        if module_result.status != CheckStatus::PASS {
            return module_result;
        }

        check_output_contains!(
            MODPROBE_CMD,
            ["-n", "-v", "udf"],
            &["install /bin/true"],
            "unable to parse modprobe output to check if udf is enabled",
            "modprobe for udf is not disabled"
        )
    }
```

bcressey (approved):

sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
stmcginnis: Yep, much clearer. Update coming.

sources/bloodhound/src/main.rs
bcressey: >What would the user do differently in these cases?

My main concern is really that someone will read "SKIP" and decide that the check was skipped because it didn't apply, not because they actually need to dig in and do additional work. But the error messages should help clarify the distinction between these two cases.

yeazelm (approved):

sources/bloodhound/src/output.rs
stmcginnis: There isn't a great way to just drop in a constant in the formatting string template. I would need to do something like:

```rust
writeln!(output, "{:pad$}{}", "Benchmark name:", name, pad=PADDING)
```

That seems less readable to me, especially for something that probably isn't going to ever change.

I can make that change if you think it would be better, but my personal inclination would be to leave it as is. Just let me know your preference.

yeazelm: If you don't think it will ever change then I'm fine to leave it. If we do ever find it changing that might be the time to switch to the `pad=PADDING` method.

stmcginnis: Definitely - sounds like a good plan!

PR #2896, opened 2023-03-15
Add pigz for parallel decompression by containerd

### LINKED ISSUE #2886: containerd with pigz
Hi, I work on improving image pull performance in bottlerocket.

containerd has the ability to speed up decompression with pigz ( https://github.com/containerd/containerd/pull/2640 ).
Is it already enabled in containerd in bottlerocket?
If not, is it possible to enable it?

ohkinozomu:
**Issue number:**

Closes #2886

**Description of changes:**

Let containerd use pigz (or more precisely, `unpigz`).
pigz doesn't have a LICENSE file (https://github.com/madler/pigz/issues/89) and I decided to use `README`...

If there is no pigz, and the log level of containerd is set to debug, the following log will be output at the time of pull.
```
Mar 14 09:30:08 10.0.2.15 containerd[1175]: time="2023-03-14T09:30:08.609765577Z" level=debug msg="unpigz not found, falling back to go gzip" error="exec: \"unpigz\": executable file not found in $PATH"
```
I have confirmed that with this change that log disappears (i.e. containerd uses unpigz).

Here is a quick benchmark.

machine spec:
Google Compute Engine c3-highcpu-22 Ubuntu 22.04.2 (enable nested virtualization)

Command to execute:
```bash
ctr image pull docker.io/library/mysql:8.0.32
```

without pigz(commit b1ac8ef6922e776ee410b6e3020ec8133c51a959)
```
6.139457575s
7.236172524s
7.730825494s
```

with pigz(this commit)
```
5.842488681s
6.319545383s
5.777997267s
```

**Testing done:** yes

**Terms of contribution:**

### DIFF
--- a/packages/pigz/Cargo.toml
+++ b/packages/pigz/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "pigz"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "build.rs"
+
+[lib]
+path = "pkg.rs"
+
+[[package.metadata.build-package.external-files]]
+url = "https://zlib.net/pigz/pigz-2.7.tar.gz"
+sha512 = "9f9f61de4a0307fc057dc4e31a98bd8d706d9e709ecde0be02a871534fddf6a1fe1321158aa72708603aaaece43f83d2423b127f7689b6219b23aea4f989e8f5"
+
+# RPM BuildRequires
+[build-dependencies]
+glibc = { path = "../glibc" }
+libz = { path = "../libz" }
+
+# RPM Requires
+[dependencies]
+# None
--- a/packages/pigz/pigz.spec
+++ b/packages/pigz/pigz.spec
@@ -1,0 +1,35 @@
+Name: %{_cross_os}pigz
+Version: 2.7
+Release: 1%{?dist}
+Summary: pigz is a parallel implementation of gzip which utilizes multiple cores
+License: Zlib AND Apache-2.0
+URL: http://www.zlib.net/pigz
+Source0: https://zlib.net/pigz/pigz-%{version}.tar.gz
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: %{_cross_os}libz-devel
+
+%description
+%{summary}.
+
+%prep
+%autosetup -n pigz-%{version} -p1
+
+%global set_env \
+%set_cross_build_flags \\\
+export CC=%{_cross_target}-gcc \\\
+%{nil}
+
+%build
+%set_env
+%make_build CC="${CC}" CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}"
+
+%install
+install -d %{buildroot}%{_cross_bindir}
+install -p -m 0755 unpigz %{buildroot}%{_cross_bindir}
+
+%files
+%license README zopfli/COPYING
+%{_cross_bindir}/unpigz
+%{_cross_attribution_file}
+
+%changelog
--- a/packages/README.md
+++ b/packages/README.md
@@ -74,7 +74,7 @@ Our sample package has the following manifest.
 [package]
 name = "libwoof"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
 publish = false
 build = "build.rs"
 
--- a/packages/containerd/Cargo.toml
+++ b/packages/containerd/Cargo.toml
@@ -22,3 +22,4 @@ glibc = { path = "../glibc" }
 # RPM Requires
 [dependencies]
 runc = { path = "../runc" }
+pigz = { path = "../pigz" }
--- a/packages/containerd/containerd.spec
+++ b/packages/containerd/containerd.spec
@@ -38,6 +38,7 @@ Source1000: clarify.toml
 BuildRequires: git
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{_cross_os}runc
+Requires: %{_cross_os}pigz
 
 %description
 %{summary}.
--- a/packages/pigz/build.rs
+++ b/packages/pigz/build.rs
@@ -1,0 +1,9 @@
+use std::process::{exit, Command};
+
+fn main() -> Result<(), std::io::Error> {
+    let ret = Command::new("buildsys").arg("build-package").status()?;
+    if !ret.success() {
+        exit(1);
+    }
+    Ok(())
+}
--- a/packages/pigz/pkg.rs
+++ b/packages/pigz/pkg.rs
@@ -1,0 +1,1 @@
+// not used
--- a/variants/Cargo.lock
+++ b/variants/Cargo.lock
@@ -268,6 +268,7 @@ name = "containerd"
 version = "0.1.0"
 dependencies = [
  "glibc",
+ "pigz",
  "runc",
 ]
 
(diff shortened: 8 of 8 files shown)


packages/pigz/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[lib]
+path = "pkg.rs"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/madler/pigz/archive/refs/tags/v2.7.tar.gz"
+sha512 = "6a617c701db3f29b83854bf40ef29554585f992928468d94ce15fee513b6b532618569fbdd8b0c8efe1bb60d4831a23f56e199bb17ffd8465241873dda506c28"
+
+# RPM BuildRequires
+[build-dependencies]
+# None
```
bcressey: This should require "glibc" and "libz" here and in the spec (with the `%{_cross_os}` prefix in the spec).

The fact that it doesn't means we're not linking against the correct zlib dependency.

Setting the `CC` variable in the spec should result in at least some build failures until / unless these dependencies are updated.

packages/pigz/pigz.spec
```diff
@@ -0,0 +1,24 @@
+Name: %{_cross_os}pigz
+Version: 2.7
+Release: 1%{?dist}
+Summary: pigz is a parallel implementation of gzip which utilizes multiple cores
+License: zlib
```
bcressey:
```suggestion
License: Zlib AND Apache-2.0
```

This is the SPDX short name for the license. It looks like we also need Apache 2.0 to cover the zopfli sources that are included in the build.

packages/pigz/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "build.rs"
+
+[lib]
+path = "pkg.rs"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/madler/pigz/archive/refs/tags/v2.7.tar.gz"
```
bcressey: I'd prefer to use the "official" source releases since those have a published hash and GPG signature:
```suggestion
url = "https://zlib.net/pigz/pigz-2.7.tar.gz"
```

packages/pigz/pigz.spec
```diff
@@ -0,0 +1,24 @@
...
+%description
+%{summary}.
+
+%prep
+%autosetup -n pigz-%{version} -p1
+
+%build
+make
+install -d %{buildroot}%{_cross_bindir}
+install -p -m 0755 unpigz %{buildroot}%{_cross_bindir}
```
bcressey, line 28: These should go in an `%install` section.

packages/pigz/pigz.spec
```diff
@@ -0,0 +1,24 @@
...
+Source0: v%{version}.tar.gz
+
+%description
+%{summary}.
+
+%prep
+%autosetup -n pigz-%{version} -p1
+
+%build
+make
```
bcressey: The `libzstd` package has an example of overriding `CC` for cross-compiling a Makefile driven build.

This isn't going to cross-compile correctly as-is; you'd end up with an x86_64 binary when building for aarch64.

packages/pigz/pigz.spec
```diff
@@ -0,0 +1,24 @@
...
+install -d %{buildroot}%{_cross_bindir}
+install -p -m 0755 unpigz %{buildroot}%{_cross_bindir}
+
+%files
+%license README
```
bcressey:
```suggestion
%license README zopfli/COPYING
```

packages/release/release.spec
```diff
@@ -100,6 +100,7 @@ Requires: %{_cross_os}iptables
 Requires: %{_cross_os}kexec-tools
 Requires: %{_cross_os}makedumpfile
 Requires: %{_cross_os}os
+Requires: %{_cross_os}pigz
```
bcressey: Since `containerd` has this as a soft dependency, it might be better to add it to that package instead.

bcressey (commented):
Thanks so much for picking this up!

Overall looks good, but needs a few fixes for some subtle cross-compiling and software licensing issues.

ohkinozomu:
Thanks for your review, I have fixed.

packages/pigz/pigz.spec
```diff
@@ -0,0 +1,39 @@
+%global debug_package %{nil}
+
+Name: %{_cross_os}pigz
+Version: 2.7
+Release: 1%{?dist}
+Summary: pigz is a parallel implementation of gzip which utilizes multiple cores
+License: Zlib AND Apache-2.0
+URL: http://www.zlib.net/pigz
+Source0: pigz-%{version}.tar.gz
```
stmcginnis: If you need to do an update, please set to:

```
Source0: https://zlib.net/pigz/pigz-%{version}.tar.gz
```

stmcginnis (commented):
Thanks for working on this!

packages/pigz/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+sha512 = "9f9f61de4a0307fc057dc4e31a98bd8d706d9e709ecde0be02a871534fddf6a1fe1321158aa72708603aaaece43f83d2423b127f7689b6219b23aea4f989e8f5"
+
+# RPM BuildRequires
+[build-dependencies]
+glibc = { path = "../glibc" }
+libz = { path = "../libz" }
+
+# RPM Requires
+[dependencies]
+# None
```
cbgbt: nit: Would you mind adding a newline to the end of files that are missing them?

packages/pigz/pigz.spec
```diff
@@ -0,0 +1,39 @@
+%global debug_package %{nil}
+
```
bcressey: This shouldn't be needed - there's another tweak required for the build to get the right compiler and linker flags passed.

cbgbt (commented):
Thanks so much for working on this! It looks good to me, just a minor nit.

arnaldo2792 (approved):

packages/pigz/pigz.spec
```diff
@@ -0,0 +1,39 @@
...
+%global set_env \
+%set_cross_build_flags \\\
+export CC=%{_cross_target}-gcc \\\
+export PREFIX=%{_cross_prefix} \\\
+export DESTDIR=%{buildroot}%{_cross_rootdir} \\\
+%{nil}
+
+%build
+%set_env
+%make_build
```
bcressey: The way `CC`, `CFLAGS` and `LDFLAGS` are defined in the Makefile means that they need to be overridden like this, and can't just be set in the environment:
```suggestion
%make_build CC="${CC}" CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}"
```

This will fix the missing debug symbols error also.

packages/pigz/pigz.spec
```diff
@@ -0,0 +1,39 @@
...
+%{summary}.
+
+%prep
+%autosetup -n pigz-%{version} -p1
+
+%global set_env \
+%set_cross_build_flags \\\
+export CC=%{_cross_target}-gcc \\\
+export PREFIX=%{_cross_prefix} \\\
+export DESTDIR=%{buildroot}%{_cross_rootdir} \\\
```
bcressey: nit: don't need these since the pigz Makefile doesn't use them

```suggestion
```

ohkinozomu:
Thanks to all reviewers!
Fix completed.

bcressey (approved):

bcressey:
Nice work. Thanks for doing this, @ohkinozomu!

stmcginnis (approved):
All comments addressed - looks great!

PR #2904, opened 2023-03-17
Fix aws.profile rendering in cred-provider template

### LINKED ISSUE #2885: `aws.profile` not respected in `ecr-credential-provider` template.
**Image I'm using:**

Bottlerocket OS 1.12.0 (aws-k8s-1.22)

Using settings like the following should result in the `AWS_PROFILE` environment variable being set to `kubelet` within the `/etc/kubernetes/kubelet/credential-provider-config.yaml` file on the host. Additionally the desired role isn't being used for pulling images.

```toml
[settings.aws]
config = "W3Byb2ZpbGUga3ViZWxldF0KY3JlZGVudGlhbF9zb3VyY2UgPSBFYzJJbnN0YW5jZU1ldGFkYXRhCnJvbGVfYXJuID0gYXJuOmF3czppYW06OjEyMzQ1Njc4OTAxMjpyb2xlL015RXh0cmFSb2xlCg=="
profile = "kubelet"
region = "us-east-1"

[settings.kubernetes.credential-providers.ecr-credential-provider]
enabled = true
image-patterns = ["*.dkr.ecr.*.amazonaws.com"]
```

The expanded `settings.aws.config` is as follows:

```
[profile kubelet]
credential_source = Ec2InstanceMetadata
role_arn = arn:aws:iam::123456789012:role/MyExtraRole
```

**What actually happened:**

The credential helper file is missing the needed environment variable. This is the output as seen with the above configuration:

```
[root@admin]# cat /.bottlerocket/rootfs/etc/kubernetes/kubelet/credential-provider-config.yaml
apiVersion: kubelet.config.k8s.io/v1alpha1
kind: CredentialProviderConfig
providers:
  - name: ecr-credential-provider
    matchImages:
      - "*.dkr.ecr.*.amazonaws.com"
    defaultCacheDuration: "12h"
    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
    env:
      - name: HOME
        value: /root
```

Use the EKS BottlerocketOS image with the additional settings shown above to observe the incorrect configuration generation. Obviously the role referenced in the `kubelet` profile should exist with sufficient permissions to allow the instance role to assume it, and be useful for container image pulling. Given this looks to, at a minimum, just be a templating issue the whole setup might not be entirely necessary.

stmcginnis:
**Issue number:**

Closes #2885

**Description of changes:**

The ecr-credential-provider can use a profile for determining which credential settings to use out of `~/.aws`. This wasn't getting rendered in the configuration template because it was inside an `{{#each}}` block that would change the handlebars context so it would look for settings values relative to the setting being iterated.

This fixes that by explicitly referencing the `aws.profile` setting from the root, ignoring iteration context.

Another issue was that setting `aws.profile` after setting the credential provider settings would not trigger re-rendering of the credential provider configuration file. This adds an "affected-services" value for the setting to tell it to restart kubelet, which triggers the rendering of the template.

A migration was needed because of this affected-services metadata change.

**Testing done:**

Set credential provider data:

```
# apiclient apply << EOF
> [settings.kubernetes.credential-providers.ecr-credential-provider]
> enabled = true
> # (optional - defaults to "12h")
> cache-duration = "30m"
> image-patterns = [
>   # One or more URL paths to match an image prefix. Supports globbing of subdomains.
>   "*.dkr.ecr.us-east-2.amazonaws.com",
>   "*.dkr.ecr.us-west-2.amazonaws.com"
> ]
> EOF
```

Verified the config file rendered with the default `aws.profile` value:

```yaml
# cat /etc/kubernetes/kubelet/credential-provider-config.yaml
kind: CredentialProviderConfig
providers:
  - name: ecr-credential-provider
    matchImages:
      - "*.dkr.ecr.us-east-2.amazonaws.com"
      - "*.dkr.ecr.us-west-2.amazonaws.com"
    defaultCacheDuration: "30m"
    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
    env:
      - name: HOME
        value: /root
      - name: AWS_PROFILE
        value: default
```

Updated the profile setting:

```
# apiclient set -j '{"aws": { "profile": "foo"}}'
```

Verified the config file was re-rendered with the correct information:

```yaml
# cat /etc/kubernetes/kubelet/credential-provider-config.yaml
kind: CredentialProviderConfig
providers:
  - name: ecr-credential-provider
    matchImages:
      - "*.dkr.ecr.us-east-2.amazonaws.com"
      - "*.dkr.ecr.us-west-2.amazonaws.com"
    defaultCacheDuration: "30m"
    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
    env:
      - name: HOME
        value: /root
      - name: AWS_PROFILE
        value: foo
```

**Terms of contribution:**

### DIFF
--- a/Release.toml
+++ b/Release.toml
@@ -189,3 +189,6 @@ version = "1.13.0"
     "migrate_v1.13.0_aws-control-container-v0-7-1.lz4",
     "migrate_v1.13.0_public-control-container-v0-7-1.lz4",
 ]
+"(1.13.0, 1.13.1)" = [
+    "migrate_v1.13.1_aws-profile-cred-provider.lz4",
+]
--- a/packages/kubernetes-1.22/credential-provider-config-yaml
+++ b/packages/kubernetes-1.22/credential-provider-config-yaml
@@ -15,9 +15,9 @@ providers:
     env:
       - name: HOME
         value: /root
-{{#if settings.aws.profile}}
+{{#if @root.settings.aws.profile}}
       - name: AWS_PROFILE
-        value: {{settings.aws.profile}}
+        value: {{@root.settings.aws.profile}}
 {{/if}}
 {{/if}}
 {{/if}}
--- a/packages/kubernetes-1.23/credential-provider-config-yaml
+++ b/packages/kubernetes-1.23/credential-provider-config-yaml
@@ -15,9 +15,9 @@ providers:
     env:
       - name: HOME
         value: /root
-{{#if settings.aws.profile}}
+{{#if @root.settings.aws.profile}}
       - name: AWS_PROFILE
-        value: {{settings.aws.profile}}
+        value: {{@root.settings.aws.profile}}
 {{/if}}
 {{/if}}
 {{/if}}
--- a/packages/kubernetes-1.24/credential-provider-config-yaml
+++ b/packages/kubernetes-1.24/credential-provider-config-yaml
@@ -15,9 +15,9 @@ providers:
     env:
       - name: HOME
         value: /root
-{{#if settings.aws.profile}}
+{{#if @root.settings.aws.profile}}
       - name: AWS_PROFILE
-        value: {{settings.aws.profile}}
+        value: {{@root.settings.aws.profile}}
 {{/if}}
 {{/if}}
 {{/if}}
--- a/packages/kubernetes-1.25/credential-provider-config-yaml
+++ b/packages/kubernetes-1.25/credential-provider-config-yaml
@@ -15,9 +15,9 @@ providers:
     env:
       - name: HOME
         value: /root
-{{#if settings.aws.profile}}
+{{#if @root.settings.aws.profile}}
       - name: AWS_PROFILE
-        value: {{settings.aws.profile}}
+        value: {{@root.settings.aws.profile}}
 {{/if}}
 {{/if}}
 {{/if}}
--- a/packages/kubernetes-1.26/credential-provider-config-yaml
+++ b/packages/kubernetes-1.26/credential-provider-config-yaml
@@ -15,9 +15,9 @@ providers:
     env:
       - name: HOME
         value: /root
-{{#if settings.aws.profile}}
+{{#if @root.settings.aws.profile}}
       - name: AWS_PROFILE
-        value: {{settings.aws.profile}}
+        value: {{@root.settings.aws.profile}}
 {{/if}}
 {{/if}}
 {{/if}}
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -562,6 +562,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-profile-cred-provider"
+version = "0.1.0"
+dependencies = [
+ "migration-helpers",
+]
+
+[[package]]
 name = "aws-sdk-cloudformation"
 version = "0.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -40,6 +40,7 @@ members = [
     "api/migration/migrations/v1.13.0/public-admin-container-v0-10-0",
     "api/migration/migrations/v1.13.0/aws-control-container-v0-7-1",
     "api/migration/migrations/v1.13.0/public-control-container-v0-7-1",
+    "api/migration/migrations/v1.13.1/aws-profile-cred-provider",
 
     "bottlerocket-release",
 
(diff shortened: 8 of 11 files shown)


sources/models/shared-defaults/aws-creds.toml
```diff
@@ -11,6 +11,9 @@ restart-commands = []
 [metadata.settings.aws]
 affected-services = ["aws"]
 
+[metadata.settings.aws.profile]
+affected-services = ["kubernetes"]
```
bcressey: I'd expect that this needs to go in a k8s-specific model file, since `aws-ecs-1` will include this but will not have a "kubernetes" service in the model.

stmcginnis: This is only pulled in to k8s specific models:

```sh
$ find ./ -name *aws-creds.toml
./sources/models/shared-defaults/aws-creds.toml
./sources/models/src/aws-k8s-1.26-nvidia/defaults.d/40-aws-creds.toml
./sources/models/src/aws-k8s-1.22-nvidia/defaults.d/40-aws-creds.toml
./sources/models/src/aws-k8s-1.26/defaults.d/40-aws-creds.toml
./sources/models/src/vmware-k8s-1.22/defaults.d/40-aws-creds.toml
./sources/models/src/metal-k8s-1.22/defaults.d/40-aws-creds.toml
./sources/models/src/vmware-k8s-1.26/defaults.d/40-aws-creds.toml
./sources/models/src/aws-k8s-1.22/defaults.d/40-aws-creds.toml
./sources/models/src/metal-k8s-1.26/defaults.d/40-aws-creds.toml
```

Release.toml
```diff
@@ -189,3 +189,6 @@ version = "1.13.0"
     "migrate_v1.13.0_aws-control-container-v0-7-1.lz4",
     "migrate_v1.13.0_public-control-container-v0-7-1.lz4",
 ]
+"(1.13.0, 1.14.0)" = [
```
stmcginnis: This should probably be `1.13.0, 1.13.1` in case we want to backport and do a stable release for this. I'll update it to make cherry-picking to the `1.13.x` branch easier. Then if we do a point release it is there. If not, easy enough to change back to 1.14.0 on the `develop` branch.

sources/models/shared-defaults/aws-creds.toml
bcressey: That's a separate bug then since it means the ECS variant will accept these settings in the API but treat them as a no-op - not rendering any files. Which I suppose is OK since nothing uses those settings, but is still surprising.

Maybe rename this file so it's clear that it's k8s-specific? Then linking it into the ECS models in the future might be flagged by a reviewer.

stmcginnis: Could it be possible we may need to use AWS credentials for ECS or other platforms in the future though?

The only k8s-specific part about this is the service restart. Would it be possible to have:

```toml
[metadata.settings.aws.profile]
affected-services = ["kubernetes"]
```

In another model that would only be pulled in to k8s variants, leaving this open to being able to be used in non-k8s variants for other use cases?

bcressey: Yes, if we put this in one of the k8s templates then that should be fine.

jpmcb (approved):

bcressey (approved):

packages/kubernetes-1.22/credential-provider-config-yaml
```diff
@@ -15,9 +15,9 @@ providers:
     env:
       - name: HOME
         value: /root
-{{#if settings.aws.profile}}
+{{#if @root.settings.aws.profile}}
```
webern, line 18: Out of curiosity, why was the `@root.` addition needed?

webern (approved):

packages/kubernetes-1.22/credential-provider-config-yaml
stmcginnis: Yeah, that was the tricky, key part of this.

Because this is inside an [{{#each}}](https://github.com/bottlerocket-os/bottlerocket/blob/f825b8186faabe4283a06ead96b2bfd4123bd0da/packages/kubernetes-1.24/credential-provider-config-yaml#L5) block, that means handlebars iterates through "each" of the `settings.kubernetes.credential-providers`  values in that collection.

What's not immediately obvious is that it also means you are within the context of those iterated values. So that's why on the next line you can do:

```
{{#if this.enabled}}
```

instead of some sort of indexing operation into the collection. But that's also where the subtle gotcha is.

It meant that later on with the line:

```
{{#if settings.aws.profile}}
```

it was actually looking at `settings.kubernetes.credential-providers[x].settings.aws.profile`. Since we're inside the `each` block, we actually need to break out of that local context to get to the real `settings.aws.profile`.

Handlebars allows you to do this in two different ways - either relationally (`../../aws.profile`) or by explicitly going from the anchor like what has been done here. This looks a lot cleaner, and probably safer, so I went with the `@root` route.

PR #2907, opened 2023-03-17
packages: add keyutils

cbgbt:
**Issue number:**
n/a

**Description of changes:**
```
keyutils is a mechanism used by the kernel to reach into userspace. Typically this is used to
fetch security keys, but it is also a mechanism that the kernel uses to resolve domain names.
```

This change is required when mounting some distributed filesystems. The Linux kernel needs the ability to resolve domain names. To do this, it uses an "upcall" procedure to interact with dns resolution in userspace. That "upcall" procedure uses the Linux userspace keyring system -- specifically `/sbin/request-key`.

Of note:
I've removed keyring configuration options from `/etc/request-key.conf` other than the DNS resolver. Our userspace utilities don't need them, and we haven't otherwise fielded requests about the missing utilities for the kernel.

**Testing done:**
* [x] Launched an instance and noted that the contents of `/etc/request-key.conf` are correct
* [x] Noted the presence of keyutils shared libraries in /usr/lib64
* [x] Tested mounting an CIFS/AD filesystem
* [x] Test new config which only includes `dns_resolver`

**Terms of contribution:**

### DIFF
--- a/packages/keyutils/keyutils.spec
+++ b/packages/keyutils/keyutils.spec
@@ -1,0 +1,78 @@
+Name: %{_cross_os}keyutils
+Version: 1.6.1
+Release: 1%{?dist}
+Summary: Linux key management utilities
+License: GPL-2.0-or-later AND GPL-2.1-or-later
+Url: http://people.redhat.com/~dhowells/keyutils/
+Source0: http://people.redhat.com/~dhowells/keyutils/keyutils-%{version}.tar.bz2
+Source1: keyutils-tmpfiles.conf
+Source2: request-key.conf
+
+BuildRequires: %{_cross_os}glibc-devel
+
+%description
+%{summary}.
+
+%package devel
+Summary: Development package for building Linux key management utilities
+
+%description devel
+%{summary}.
+
+%prep
+%setup -n keyutils-%{version} -q
+
+%global keyutilsmake \
+%set_cross_build_flags \\\
+export CC=%{_cross_target}-gcc ; \
+%make_build \\\
+  NO_ARLIB=1 \\\
+  ETCDIR=%{_cross_sysconfdir} \\\
+  LIBDIR=%{_cross_libdir} \\\
+  USRLIBDIR=%{_cross_libdir} \\\
+  BINDIR=%{_cross_bindir} \\\
+  SBINDIR=%{_cross_sbindir} \\\
+  MANDIR=%{_cross_mandir} \\\
+  INCLUDEDIR=%{_cross_includedir} \\\
+  SHAREDIR=%{_cross_datadir}/keyutils \\\
+  RELEASE=.%{release} \\\
+  NO_GLIBC_KEYERR=1 \\\
+  CC="${CC}" \\\
+  CFLAGS="${CFLAGS}" \\\
+  LDFLAGS="${LDFLAGS}" \\\
+  DESTDIR=%{buildroot} \\\
+%{nil}
+
+
+%build
+%keyutilsmake
+
+%install
+%keyutilsmake install
+
+install -d %{buildroot}%{_cross_tmpfilesdir}
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
+
+install -p -m 0644 %{S:1} %{buildroot}%{_cross_tmpfilesdir}/keyutils.conf
+install -p -m 0644 %{S:2} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/request-key.conf
+
+%files
+%{_cross_attribution_file}
+%license LICENCE.GPL LICENCE.LGPL
+%{_cross_tmpfilesdir}/keyutils.conf
+%{_cross_bindir}/keyctl
+%{_cross_sbindir}/key.dns_resolver
+%{_cross_sbindir}/request-key
+%{_cross_datadir}/keyutils
+%{_cross_libdir}/libkeyutils.so.*
+%{_cross_factorydir}%{_cross_sysconfdir}/request-key.conf
+
+%exclude %{_cross_mandir}
+%exclude %{_cross_sysconfdir}/request-key.conf
+
+%files devel
+%{_cross_libdir}/libkeyutils.so
+%{_cross_includedir}/keyutils.h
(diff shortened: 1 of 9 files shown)


stmcginnis (approved):
This looks good to me, assuming the testing results are good.

packages/keyutils/keyutils.spec
```diff
@@ -0,0 +1,85 @@
+Name: %{_cross_os}keyutils
+Version: 1.6.1
+Release: 1%{?dist}
+Summary: Linux Key Management Utilities
+License: GPLv2+ and LGPLv2+
```
bcressey: We use SPDX identifiers and expressions in the license field, so this should be:
```suggestion
License: GPL-2.0-or-later AND LGPL-2.1-or-later
```

packages/keyutils/keyutils.spec
```diff
@@ -0,0 +1,85 @@
+Name: %{_cross_os}keyutils
+Version: 1.6.1
+Release: 1%{?dist}
+Summary: Linux Key Management Utilities
```
bcressey: nit:
```suggestion
Summary: Linux key management utilities
```

packages/keyutils/keyutils.spec
```diff
@@ -0,0 +1,85 @@
...
+
+%description devel
+%{summary}.
+
+%prep
+%setup -n keyutils-%{version} -q
+
+%global set_env \
+%set_cross_build_flags \\\
+export CC=%{_cross_target}-gcc \\\
```
bcressey: nit: maybe merge this into the `%keyutilsmake` macro

packages/keyutils/keyutils.spec
```diff
@@ -0,0 +1,85 @@
...
+
+
+%build
+%set_env
+%keyutilsmake \
+  RELEASE=.%{release} \
+  NO_GLIBC_KEYERR=1 \
+  CC="${CC}" \
+  CFLAGS="${CFLAGS}" \
+  LDFLAGS="${LDFLAGS}"
```
bcressey: nit: you could probably move all this into `%keyutilsmake` also

Depending on how the Makefile is set up, changes to some of these variables - which happens implicitly when omitting them when they were previously defined - can cause `make install` to actually rebuild the binaries, potentially in an incorrect way if the compiler changes.

Here I don't think it matters, but it makes me uncomfortable to see the flags changing across `%build` and `%install` because I've been burned by it before.

packages/keyutils/keyutils.spec
```diff
@@ -0,0 +1,85 @@
...
+%{_cross_attribution_file}
+%license LICENCE.GPL LICENCE.LGPL
+%{_cross_tmpfilesdir}/keyutils.conf
+%{_cross_bindir}/keyctl
+%{_cross_sbindir}/key.dns_resolver
+%{_cross_sbindir}/request-key
+%{_cross_datadir}/keyutils
+%{_cross_libdir}/libkeyutils.so.*
+
+%dir %{_cross_factorydir}
```
bcressey: You can omit this; I only use `%dir` to claim ownership of a directory when it's clearly tied to this package, usually because it's a subdirectory named for the software. Otherwise you end up with multiple packages owning the same directory, which isn't ideal.

```suggestion
```

packages/keyutils/request-key.conf
```diff
@@ -0,0 +1,27 @@
...
+################################################################################
+
+#OP	TYPE	DESCRIPTION	CALLOUT INFO	PROGRAM ARG1 ARG2 ARG3 ...
+#======	=======	===============	===============	===============================
+create  dns_resolver *		*               /sbin/key.dns_resolver %k
+create	user	debug:*		negate		/bin/keyctl negate %k 30 %S
+create  user    debug:*         rejected        /bin/keyctl reject %k 30 %c %S
+create  user    debug:*         expired         /bin/keyctl reject %k 30 %c %S
+create  user    debug:*         revoked         /bin/keyctl reject %k 30 %c %S
+create	user	debug:loop:*	*		|/bin/cat
```
bcressey: I am tempted to omit all of the documentation comments and just include the `create dns_resolver` and `negate` entries - the non-user debug ones.

cbgbt:
Force pushed to address feedback from @bcressey

packages/keyutils/keyutils.spec
```diff
@@ -0,0 +1,78 @@
...
+%prep
+%setup -n keyutils-%{version} -q
+
+%global keyutilsmake \
+%set_cross_build_flags \\\
+export CC=%{_cross_target}-gcc ; \
+%make_build \\\
+  NO_ARLIB=1 \\\
+  ETCDIR=%{_cross_sysconfdir} \\\
+  LIBDIR=%{_cross_libdir} \\\
```
bcressey, line 31: nit: it's very unlikely that this will ever matter, but setting `PREFIX` will put the right paths in `libkeyutils.pc` (which we don't and probably won't use):
```suggestion
  PREFIX=%{_cross_prefix} \\\
  LIBDIR=%{_cross_libdir} \\\
```

bcressey (approved):
LGTM ☘️

arnaldo2792 (approved):

zmrow (approved):

packages/keyutils/keyutils.spec
cbgbt: I added the PREFIX here and compared `libkeyutils.pc` before and after the change, but it seems like the paths are already correct?

Before:

```
libdir=/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib
includedir=/x86_64-bottlerocket-linux-gnu/sys-root/usr/include

Name: libkeyutils
Description: keyutils library
Version: 1.6.1
Cflags: -I${includedir}
Libs: -L${libdir} -lkeyutils
```

After:

```
libdir=/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib
includedir=/x86_64-bottlerocket-linux-gnu/sys-root/usr/include

Name: libkeyutils
Description: keyutils library
Version: 1.6.1
Cflags: -I${includedir}
Libs: -L${libdir} -lkeyutils
```

PR #2921, opened 2023-03-21
hardening: Enable setting templated file permissions

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This adds the ability to specify the permission mode of files created from a template. This enables the ability to have file-by-file control for sensitive files to enforce more restrictive permissions.

Note that Bottlerocket is not a multiuser system and this is just extra precaution to change the permissions on these files. It does happen from time to time that security tools meant for more general purpose Linux distributions are run against Bottlerocket and report issues based on the default file permissions, so this is partly a cosmetic change to prevent these tools from report false errors.

**Testing done:**

Built and deployed EKS cluster. Verified that the file permissions on the modified template files were set to the correct permissions and that the system operated as expected.

**Terms of contribution:**

### DIFF
--- a/sources/api/thar-be-settings/src/config.rs
+++ b/sources/api/thar-be-settings/src/config.rs
@@ -162,8 +162,7 @@ impl RenderedConfigFile {
         let mut binding = OpenOptions::new();
         let options = binding.write(true).create(true).mode(DEFAULT_FILE_MODE);
 
-        // See if this file has a config setting for a specific mode, but don't error if we are
-        // not able to honor that request.
+        // See if this file has a config setting for a specific mode
         if let Some(mode) = &self.mode {
             let mode_int =
                 u32::from_str_radix(mode.as_str(), 8).context(error::TemplateModeSnafu {
--- a/sources/api/thar-be-settings/src/config.rs
+++ b/sources/api/thar-be-settings/src/config.rs
@@ -3,11 +3,14 @@ use crate::{error, Result};
 use itertools::join;
 use snafu::{ensure, ResultExt};
 use std::collections::HashSet;
-use std::fs;
+use std::fs::{self, OpenOptions};
+use std::io::prelude::*;
+use std::os::unix::fs::OpenOptionsExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
 const SYSTEMCTL_DAEMON_RELOAD: &str = "systemctl daemon-reload";
+const DEFAULT_FILE_MODE: u32 = 0o644;
 
 /// Query the API for ConfigurationFile data
 #[allow(clippy::implicit_hasher)]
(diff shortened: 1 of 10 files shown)


sources/api/thar-be-settings/src/config.rs
```diff
@@ -148,10 +159,33 @@ impl RenderedConfigFile {
...
+        // not able to honor that request.
+        if let Some(mode) = &self.mode {
+            if let Ok(mode_int) = u32::from_str_radix(mode.as_str(), 8) {
+                options.mode(mode_int);
+            } else {
+                warn!("Unable to parse mode setting: {}", mode);
+            }
+        } else {
+            warn!("Unable to parse mode string setting: {:?}", self.mode);
+        }
```
bcressey: Why not error if the intended mode can't be set? That seems like a potentially unpleasant surprise.

It might make sense to fall back to the most restricted mode, `0o000`, in this case.

bcressey (commented):
Looks good to me overall, just one suggestion.

sources/api/thar-be-settings/src/config.rs
stmcginnis: I didn't think it likely this could ever fail. I would expect other issues in the boot sequence to have a problem by the time we got to rendering templates if there was some sort of disk IO issue.

But also was concerned about failing the boot process just due to this. I believe if a failure bubbles up at this point, it would actually cause the node to fail to boot rather than just causing a kubelet issue. Since we are mostly protected by the other Bottlerocket security issues - and this feels kind of like wearing a belt with suspenders - I thought the most user friendly thing would be just to log it.

But I'm fine being told I'm making some bad assumptions! :) I guess along the lines of "I don't think it likely this could ever fail"... if so, having this error here wouldn't make that much of a difference and could mean something is very wrong anyway.

stmcginnis:
Rebased to address `Release.toml` conflict due to 1.13.3.

yeazelm (approved):

sources/api/thar-be-settings/src/config.rs
```diff
@@ -3,11 +3,14 @@ use crate::{error, Result};
...
 use std::collections::HashSet;
-use std::fs;
+use std::fs::{self, OpenOptions};
+use std::io::prelude::*;
+use std::os::unix::fs::OpenOptionsExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
 const SYSTEMCTL_DAEMON_RELOAD: &str = "systemctl daemon-reload";
+const DEFAULT_FILE_MODE: u32 = 0o644;
```
jpmcb, line 13: This will be readable by all groups? I'm curious the reasoning behind this permission mode as the default? My gut instinct would be to have a `0600` (which it looks like you've set elsewhere) to prevent other users from reading. Granted, this is weird and nuanced since Bottlerocket isn't a multi-user system anyways but curious to hear your thoughts.

jpmcb (approved):

sources/api/thar-be-settings/src/config.rs
stmcginnis: That is the existing default mode, so I didn't want to change any behavior by adding the ability to set it to something other than that. I agree it ultimately doesn't make much difference, but wanted to minimize any possible side effects.

stmcginnis:
Force push to resolve merge conflict.

sources/api/thar-be-settings/src/config.rs
```diff
@@ -148,10 +159,32 @@ impl RenderedConfigFile {
...
         };
 
-        fs::write(&self.path, self.rendered.as_bytes()).context(error::TemplateWriteSnafu {
-            path: &self.path,
-            pathtype: "file",
-        })
+        let mut binding = OpenOptions::new();
+        let options = binding.write(true).create(true).mode(DEFAULT_FILE_MODE);
+
+        // See if this file has a config setting for a specific mode, but don't error if we are
```
bcressey: nit: this comment is out-of-date now that we error out

bcressey (approved):

sources/api/thar-be-settings/src/config.rs
stmcginnis: Cleaned up.

Issue #2883, opened 2023-03-11
Why are `image-gc-high-threshold-percent` and `image-gc-low-threshold-percent` strings?


jonathan-innis:
**Image I'm using:**

`bottlerocket-aws-k8s-1.23-x86_64-v1.12.0-6ef1139f`

**What I expected to happen:**

The [kubeletConfiguration] defines these values as int32s in the spec so it's odd to me that these values would be expected to be strings in the current bottlerocket setting configuration. I imagine (due to breaking changes) this would be hard to fix now, but it's surprising behavior for someone who is coming from configuration the `kubelet-config.json` on something like an AL2 image.

**What actually happened:**

Bottlerocket fails to start if it's passed in the values as integers rather than strings

**How to reproduce the problem:**

[settings.kubernetes]
image-gc-high-threshold-percent = 85
image-gc-low-threshold-percent = 80

yeazelm:
Thanks for submitting this issue @jonathan-innis! I have been digging a bit around on this and it looks like this is mostly due to Bottlerocket originally only supporting strings but we have added [support for scalar values](https://github.com/bottlerocket-os/bottlerocket/pull/2559) so it should be possible to convert these over.

> I imagine (due to breaking changes) this would be hard to fix now

We have been able to do some of these types of changes in the past, so it's not unreasonable to change this.

jonathan-innis:
Thanks @yeazelm. Yeah, it seems like a weird opinionated stance to take. The current state certainly isn't a blocker; however, I do think that it should be properly documented somewhere in the BR repo (maybe in the README under the settings). This seems like a recurring issue that I've seen a few other people hit and I know that there is [already an issue opened in this repo](https://github.com/bottlerocket-os/bottlerocket/issues/2324) asking for this to be documented.

webern:
I have opened a PR that attempts to remedy this situation. It allows these to be expressed as numbers and adds documentation to the README.

> a weird opinionated stance to take

It wasn't intentional. There were some implementation details in the code that caused us to do the wrong thing with these settings when they were added.

I looked through all of the settings and couldn't find any others that should be numbers instead of strings.

Issue #2742, opened 2023-01-19
kubelet >=v1.25: expose setting to enable the use of `RuntimeDefault` as the default seccomp profile for all workloads


etungsten:
**What I'd like:**
A new `settings.kubernetes` setting for controlling whether `--seccomp-default` is passed to kubelet or not.

In K8s 1.25 `SeccompDefault` got promoted to beta which allows users to use a default seccomp profile for all their orchestrated container workloads.

See https://kubernetes.io/docs/tutorials/security/seccomp/#enable-the-use-of-runtimedefault-as-the-default-seccomp-profile-for-all-workloads:
> If enabled, the kubelet will use the RuntimeDefault seccomp profile by default, which is defined by the container runtime, instead of using the Unconfined (seccomp disabled) mode. The default profiles aim to provide a strong set of security defaults while preserving the functionality of the workload.

webern:
I haven't fully wrapped my head around all of this, but I hope we don't run into an issue with the: https://github.com/containerd/containerd/blob/d769f03592f11e6c0f06bdeeb64527c28f0ee984/pkg/cri/server/container_create_linux.go#L132

etungsten:
I don't think there's any conflicts with the containerd logic when picking the runtime spec. With the flag enabled, Kubelet will use the default seccomp profile as defined by `cri/containerd`.

zshen-figma:
Hello, do we have an estimate on when this will be implemented?

cartrius-a:
Hi @zshen-figma, I've gone ahead and opened a PR (#3334) to add this to Bottlerocket.

PR #2697, opened 2023-01-03
containerd: add OCI default spec and settings

mchaker:
**Issue number:**

Closes #1703 

**Description of changes:**

This PR adds support for setting default OCI spec values. The default OCI spec stored at `/etc/containerd/cri-base.json` is applied to **all** containers run by containerd, unless otherwise configured (e.g. by Kubernetes).

The OCI spec can be found in detail at https://github.com/opencontainers/runtime-spec/blob/main/config.md .

This PR also adds some default settings that we want Bottlerocket to ship with: namely, a specific set of process capabilities (`sources/models/shared-defaults/oci-capabilities.toml`) and resource limits (`sources/models/shared-defaults/oci-resource-limits.toml`).

# **Testing done:**

## OCI Defaults Test Plan
​
#### Baseline
​
- [x] Baseline with Production AMI v1.11 (i.e. `ami-09d516ec16c94c0e0`)
  - [x] Get the OCI spec for a sample nginx workload container launched via kubernetes (save it for comparison)
  - [x] Run a pod that opens as many files as it can, as well as printing the container capabilities
    - [x] How many files were opened: 65533 due to [our containerd patch](https://github.com/bottlerocket-os/bottlerocket/blob/d2a03715618aa52b0b0d8c07e7f6194080abee44/packages/containerd/1001-cri-set-default-RLIMIT_NOFILE.patch).
    - [x] Bounding capabilities: {CAP_SETFCAP, CAP_NET_RAW, CAP_KILL, CAP_CHOWN, CAP_SETUID, CAP_FSETID, CAP_FOWNER, CAP_AUDIT_WRITE, CAP_MKNOD, CAP_SYS_CHROOT, CAP_DAC_OVERRIDE, CAP_SETPCAP, CAP_NET_BIND_SERVICE, CAP_SETGID} 
​
#### Test case: feature enabled and configured with default settings
​
Make sure there is no behavior change when not configuring the new settings.
​
- [x] Run modified AMI (ami-0f58fc68a762109e8, 7ad3110d) **without** setting `oci-defaults.capabilities`, `oci-defaults.resource-limits` in userdata
  - [x] Get the OCI spec for a nginx workload container: it matched v1.11
  - [x] Run a pod that opens as many files as it can, as well as printing the container capabilities
    - [x] How many files were opened: 65533 (matched v1.11)
[...]

PR #2597, opened 2022-11-21
provide a better way to model enums in settings

webern:
**Issue number:**

Provides a solution for #1501 
Follows the work of #2559

It doesn't close the issue because we have plenty of additional enums to convert after this PR.

**Description of changes:**

See `sources/models/src/modeled_types/ecs.rs` to understand why. This PR:

- Makes a small change to the datastore to support enum values.
- Augments the `Scalar` derive macro to generate some basic string conveniences for enum types.
- Eliminates the unnecessary `SomeStruct{ inner: String }` wrappers for the two enums found in `aws-ecs-1`.
- Doesn't convert modeled enums in other variants yet to limit blast radius.

# Testing done:

## Baseline with Good Values

I ran the latest production ECS ami, `ami-08cf4a2f9255f1cfb`

with this userdata:

```toml
[settings.ecs]
image-pull-behavior="once"
loglevel="warn"
```

Things I checked:

- `cat /etc/os-release` -> `VERSION="1.11.0 (aws-ecs-1)"` and `BUILD_ID=b530f308`
- ✅ `cat /etc/ecs/ecs.config` had `ECS_LOGLEVEL="warn"`
- ✅ `cat /etc/ecs/ecs.config.json` had `"ImagePullBehavior":2`
- ✅ `apiclient get settings.ecs.loglevel` -> `warn`
- ✅ `apiclient get settings.ecs.image-pull-behavior` -> `once`
- ✅ `cat /var/lib/bottlerocket/datastore/current/live/settings/ecs/loglevel` -> `"warn"`
- ✅ `cat /var/lib/bottlerocket/datastore/current/live/settings/ecs/image-pull-behavior` -> `"once"`
- ✅ `apiclient set settings.ecs.loglevel=badvalue` -> Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: Error deserializing scalar value: Unable to deserialize into ECSAgentLogLevel: Invalid input for field ecs.loglevel: unknown variant `badvalue`, expected one of `debug`, `info`, `warn`, `error`, `crit`
- ✅ `apiclient set settings.ecs.image-pull-behavior=badvalue` -> Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: Error deserializing scalar value: Unable to deserialize into ECSAgentImagePullBehavior: Invalid input for field ecs.image_pull_behavior: unknown variant `badvalue`, expected one of `default`, `always`, `once`, `prefer-cached`
[...]

PR #2930, opened 2023-03-22
Additional kubelet configuration settings

### LINKED ISSUE #1445: Pass all arguments related to soft eviction to the kubelet
_This is awesome - we've built an AMI with this patch and its working wonderfully. In a followup PR, we'd like to see the `evictionSoft` settings exposed, as well as the `eviction-max-pod-grace-period` setting. Overall though, this is great!_

_Originally posted by @diranged in https://github.com/bottlerocket-os/bottlerocket/pull/1388#discussion_r606458121_

What I'd like:
Support customer to pass all arguments related to soft eviction to Kubelet in Bottlerocket. https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/#soft-eviction-thresholds

### LINKED ISSUE #1792: Support for Graceful Node Shutdown in the Kubelet Configuration
**What I'd like:**
Graceful Node Shutdown is enabled by default in Kubernetes 1.21, but it requires that you configure ShutdownGracePeriod and ShutdownGracePeriodCriticalPods within the Kubelet configuration.

It'd be excellent if we could get support for configuring it via user-data settings.

Relevant KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2000-graceful-node-shutdown
Docs: https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown

### LINKED ISSUE #2877: Option to set cpu-cfs-quota in Kubelet Argument to avoid CPU throttling
**What I'd like:**
I would like to get the option to setup [cpuCFSQuota](https://v1-24.docs.kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#:~:text=exit.%20Default%3A%20false-,cpuCFSQuota,-bool) in kubelet argument, as this is not supported as of now

**Why:**
We want to make this option false to avoid unnecessary CPU Throttling in in EKS when used with bottlerocket. For reference [#1026](https://github.com/zalando-incubator/kubernetes-on-aws/issues/1026) this works with aws amazon linux

We have considered removing CPU Limit from the pod constraint but that's only the workaround and we want to go back to CPU Limit in all containers.

I walked around the documentation and didn't find any flag for cfs-quota but Let me know in case there is already an option that i might have missed Or any other alternatives which I can use to achieve this

### LINKED ISSUE #2977: Add in support for the --memory-manager-policy flag?
**What I'd like:**

https://github.com/bottlerocket-os/bottlerocket/pull/1659 helps us use the topology manager - but it's real power comes with the 
 [memory manager](https://kubernetes.io/docs/tasks/administer-cluster/memory-manager/) flags. Can we have these added in so that we have parity with the `--cpu-manager-policy` flags?

stmcginnis:
### Issue number:

Closes: #2877 
Closes: #1792 
Closes: #1445
Closes: #2977

### Description of changes:

This adds new settings that can be used for soft eviction, graceful shutdown, CPU quota enforcement, and memory manager policy.

The new settings are:

**Soft eviction**

- `settings.kubernetes.eviction-soft`
- `settings.kubernetes.eviction-soft-grace-period`
- `settings.kubernetes.eviction-max-pod-grace-period`

`eviction-soft` is the same as the existing `eviction-hard` but allows a grace period before pods are evicted. The `eviction-soft-grace-period` allows setting this grace period for each of the support eviction keys. `eviction-max-grace-period` is used to set an overall maximum for the eviction grace period.

**Graceful shutdown**

- `settings.kubernetes.shutdown-grace-period`
- `settings.kubernetes.shutdown-grace-period-for-critical-pods`

These allow setting a grace period to wait on node shut down to allow pods to exit. Of that grace period, a portion can be dedicated to making sure critical pods are shut down.

**CPU quota enforcement**

- `settings.kubernetes.cpu-cfs-quota-enforced`

This sets a flag telling kubelet whether or not to enforce quota for containers that specify CPU limits. The default behavior is to enforce these quotas.

For additional reference on the kubelet setting values, see [the node eviction documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/) and the [`KubeletConfiguration` docs](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration).

**Memory manager**

- `settings.kubernetes.memory-manager-policy`
- `settings.kubernetes.memory-manager-reserved-memory`

Allows configuring kubelet settings for the [memory manager](https://kubernetes.io/docs/tasks/administer-cluster/memory-manager/).

### Testing done:

Built and deployed a cluster. Configured these settings with:

```sh
apiclient apply << EOF
"memory.available" = "15%"

[settings.kubernetes.eviction-soft]
"memory.available" = "12%"

[settings.kubernetes.eviction-soft-grace-period]
"memory.available" = "30s"

[settings.kubernetes]
cpu-cfs-quota-enforced = false
shutdown-grace-period = "30s"
shutdown-grace-period-for-critical-pods = "10s"
eviction-max-pod-grace-period = 40
memory-manager-policy = "Static"

[settings.kubernetes.memory-manager-reserved-memory.0]
enabled = true
memory = "674Mi"
EOF
```

Verified values were accepted. The performed:

```sh
cat /etc/kubernetes/kubelet/config
```

Verified the config file had the expected configuration keys and values. Then ran:

```sh
systemctl status kubelet
journalctl -u kubelet
```

And made sure there were no errors or issues with service operation.

### Terms of contribution:

### DIFF
--- a/README.md
+++ b/README.md
@@ -526,6 +526,34 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.kube-api-qps`: The QPS to use while talking with kubernetes apiserver.
 * `settings.kubernetes.log-level`: Adjust the logging verbosity of the `kubelet` process.
   The default log level is 2, with higher numbers enabling more verbose logging.
+* `settings.kubernetes.memory-manager-policy`: The memory management policy to use: `None` (default) or `Static`.
+  Note, when using the `Static` policy you should also set `settings.kubernetes.memory-manager-reserved-memory` values.
+* `settings.kubernetes.memory-manager-reserved-memory`: Used to set the total amount of reserved memory for a node.
+  These settings are used to configure memory manager policy when `settings.kubernetes.memory-manager-policy` is set to `Static`.
+
+  `memory-manager-reserved-memory` is set per NUMA node. For example:
+
+  ```toml
+  [settings.kubernetes]
+  "memory-manager-policy" = "Static"
+
+  [settings.kubernetes.memory-manager-reserved-memory.0]
+  # Reserve a single 1GiB huge page along with 674MiB of memory
+  "enabled" = true
+  "memory" = "674Mi"
+  "hugepages-1Gi" = "1Gi"
+
+  [settings.kubernetes.memory-manager-reserved-memory.1]
+  # Reserve 1,074 2MiB huge pages
+  "enabled" = true
+  "hugepages-2Mi" = "2148Mi"
+  ```
+
+  **Warning:** `memory-manager-reserved-memory` settings are an advanced configuration and requires a clear understanding of what you are setting.
+  Misconfiguration of reserved memory settings may cause the Kubernetes `kubelet` process to fail.
+  It can be very difficult to recover from configuration errors.
+  Use the memory reservation information from `kubectl describe node` and make sure you understand the Kubernetes documentation related to the [memory manager](https://kubernetes.io/docs/tasks/administer-cluster/memory-manager/) and how to [reserve compute resources for system daemons](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/).
+
 * `settings.kubernetes.pod-pids-limit`: The maximum number of processes per pod.
 * `settings.kubernetes.provider-id`: This sets the unique ID of the instance that an external provider (i.e. cloudprovider) can use to identify a specific node.
 * `settings.kubernetes.registry-burst`: The maximum size of bursty pulls.
--- a/README.md
+++ b/README.md
@@ -518,6 +518,8 @@ If you downgrade from v1.14.0 to an earlier version, and you have these values s
 * `settings.kubernetes.registry-burst`: The maximum size of bursty pulls.
 * `settings.kubernetes.registry-qps`: The registry pull QPS.
 * `settings.kubernetes.server-tls-bootstrap`: Enables or disables server certificate bootstrap. When enabled, the kubelet will request a certificate from the certificates.k8s.io API. This requires an approver to approve the certificate signing requests (CSR). Defaults to `true`.
+* `settings.kubernetes.shutdown-grace-period`: Delay the node should wait for pod termination before shutdown. Default is `0s`.
+* `settings.kubernetes.shutdown-grace-period-for-critical-pods`: The portion of the shutdown delay that should be dedicated to critical pod shutdown. Default is `0s`.
 * `settings.kubernetes.standalone-mode`: Whether to run the kubelet in standalone mode, without connecting to an API server. Defaults to `false`.
 * `settings.kubernetes.system-reserved`: Resources reserved for system components.
 
--- a/README.md
+++ b/README.md
@@ -417,6 +417,7 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.cluster-domain`: The DNS domain for this cluster, allowing all Kubernetes-run containers to search this domain before the host's search domains. Defaults to `cluster.local`.
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
+* `settings.kubernetes.cpu-cfs-quota-enforced`: Whether CPU CFS quotas are enforced. Defaults to `true`.
 * `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`. If you want to allow pods with certain resource characteristics to be granted increased CPU affinity and exclusivity on the node, you can set this setting to `static`. You should reboot if you change this setting after startup - try `apiclient reboot`.
 * `settings.kubernetes.cpu-manager-policy-options`: Policy options to apply when `cpu-manager-policy` is set to `static`. Currently `full-pcpus-only` is the only option.
 
--- a/README.md
+++ b/README.md
@@ -477,37 +477,50 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.event-burst`: The maximum size of a burst of event creations.
 * `settings.kubernetes.event-qps`: The maximum event creations per second.
 * `settings.kubernetes.eviction-hard`: The signals and thresholds that trigger pod eviction.
+* `settings.kubernetes.eviction-max-pod-grace-period`: Maximum grace period, in seconds, to wait for pod termination before soft eviction. Default is `0`.
+* `settings.kubernetes.eviction-soft`: The signals and thresholds that trigger pod eviction with a provided grace period.
+* `settings.kubernetes.eviction-soft-grace-period`: Delay for each signal to wait for pod termination before eviction.
+
   Remember to quote signals (since they all contain ".") and to quote all values.
 
-  Example user data for setting up eviction hard:
+  Example user data for setting up eviction values:
 
   ```toml
   [settings.kubernetes.eviction-hard]
   "memory.available" = "15%"
+
+  [settings.kubernetes.eviction-soft]
+  "memory.available" = "12%"
+
+  [settings.kubernetes.eviction-soft-grace-period]
+  "memory.available" = "30s"
+
+  [settings.kubernetes]
+  "eviction-max-pod-grace-period" = 40
   ```
 
 * `settings.kubernetes.image-gc-high-threshold-percent`: The percent of disk usage after which image garbage collection is always run, expressed as an integer from 0-100 inclusive.
 * `settings.kubernetes.image-gc-low-threshold-percent`: The percent of disk usage before which image garbage collection is never run, expressed as an integer from 0-100 inclusive.
 
-Since v1.14.0 `image-gc-high-threshold-percent` and `image-gc-low-threshold-percent` can be represented as numbers.
-For example:
+  Since v1.14.0 `image-gc-high-threshold-percent` and `image-gc-low-threshold-percent` can be represented as numbers.
+  For example:
 
-```toml
-[settings.kubernetes]
-image-gc-high-threshold-percent = 85
-image-gc-low-threshold-percent = 80
-```
+  ```toml
+  [settings.kubernetes]
+  image-gc-high-threshold-percent = 85
+  image-gc-low-threshold-percent = 80
+  ```
 
-For backward compatibility, both string and numeric representations are accepted since v1.14.0.
-Prior to v1.14.0 these needed to be represented as strings, for example:
+  For backward compatibility, both string and numeric representations are accepted since v1.14.0.
+  Prior to v1.14.0 these needed to be represented as strings, for example:
 
-```toml
-[settings.kubernetes]
-image-gc-high-threshold-percent = "85"
-image-gc-low-threshold-percent = "80"
-```
+  ```toml
+  [settings.kubernetes]
+  image-gc-high-threshold-percent = "85"
+  image-gc-low-threshold-percent = "80"
+  ```
 
-If you downgrade from v1.14.0 to an earlier version, and you have these values set as numbers, they will be converted to strings on downgrade.
+  If you downgrade from v1.14.0 to an earlier version, and you have these values set as numbers, they will be converted to strings on downgrade.
 
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.kube-api-qps`: The QPS to use while talking with kubernetes apiserver.
--- a/Release.toml
+++ b/Release.toml
@@ -204,5 +204,6 @@ version = "1.14.0"
 "(1.13.5, 1.14.0)" = [
     "migrate_v1.14.0_kubernetes-gc-percent-type-change.lz4",
     "migrate_v1.14.0_kubelet-config-settings.lz4",
+    "migrate_v1.14.0_kubelet-prefix-config-settings.lz4",
     "migrate_v1.14.0_k8s-services-mode.lz4",
 ]
--- a/sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
+++ b/sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
@@ -8,6 +8,9 @@ fn run() -> Result<()> {
         "settings.kubernetes.cpu-cfs-quota-enforced",
         "settings.kubernetes.shutdown-grace-period",
         "settings.kubernetes.shutdown-grace-period-for-critical-pods",
+        "settings.kubernetes.eviction-soft",
+        "settings.kubernetes.eviction-soft-grace-period",
+        "settings.kubernetes.eviction-max-pod-grace-period",
     ]))
 }
 
--- a/sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
+++ b/sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
@@ -5,6 +5,7 @@ use std::process;
 fn run() -> Result<()> {
     migrate(AddSettingsMigration(&[
         "settings.kubernetes.cpu-manager-policy-options",
+        "settings.kubernetes.cpu-cfs-quota-enforced",
     ]))
 }
 
--- a/sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
+++ b/sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
@@ -6,6 +6,8 @@ fn run() -> Result<()> {
     migrate(AddSettingsMigration(&[
         "settings.kubernetes.cpu-manager-policy-options",
         "settings.kubernetes.cpu-cfs-quota-enforced",
+        "settings.kubernetes.shutdown-grace-period",
+        "settings.kubernetes.shutdown-grace-period-for-critical-pods",
     ]))
 }
 
--- a/sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
+++ b/sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
@@ -11,6 +11,7 @@ fn run() -> Result<()> {
         "settings.kubernetes.eviction-soft",
         "settings.kubernetes.eviction-soft-grace-period",
         "settings.kubernetes.eviction-max-pod-grace-period",
+        "settings.kubernetes.memory-manager-policy",
     ]))
 }
 
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -242,6 +242,7 @@ struct KubernetesSettings {
     kube_api_burst: i32,
     container_log_max_size: KubernetesQuantityValue,
     container_log_max_files: i32,
+    cpu_cfs_quota_enforced: bool,
     cpu_manager_policy: CpuManagerPolicy,
     cpu_manager_reconcile_period: KubernetesDurationValue,
     cpu_manager_policy_options: Vec<KubernetesCPUManagerPolicyOption>,
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -256,6 +256,8 @@ struct KubernetesSettings {
     credential_providers: HashMap<Identifier, CredentialProvider>,
     server_certificate: ValidBase64,
     server_key: ValidBase64,
+    shutdown_grace_period: KubernetesDurationValue,
+    shutdown_grace_period_for_critical_pods: KubernetesDurationValue,
 
     // Settings where we generate a value based on the runtime environment.  The user can specify a
     // value to override the generated one, but typically would not.
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -174,6 +174,8 @@ mod de;
 
 use modeled_types::KubernetesCPUManagerPolicyOption;
 use modeled_types::KubernetesEvictionKey;
+use modeled_types::KubernetesMemoryManagerPolicy;
+use modeled_types::KubernetesMemoryReservation;
 use modeled_types::NonNegativeInteger;
 pub use variant::*;
 
@@ -263,6 +265,8 @@ struct KubernetesSettings {
     server_key: ValidBase64,
     shutdown_grace_period: KubernetesDurationValue,
     shutdown_grace_period_for_critical_pods: KubernetesDurationValue,
+    memory_manager_reserved_memory: HashMap<Identifier, KubernetesMemoryReservation>,
+    memory_manager_policy: KubernetesMemoryManagerPolicy,
 
     // Settings where we generate a value based on the runtime environment.  The user can specify a
     // value to override the generated one, but typically would not.
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -173,6 +173,7 @@ mod variant;
 mod de;
 
 use modeled_types::KubernetesCPUManagerPolicyOption;
+use modeled_types::KubernetesEvictionKey;
 pub use variant::*;
 
 // Types used to communicate between client and server for 'apiclient exec'.
@@ -193,11 +194,11 @@ use crate::modeled_types::{
     DNSDomain, ECSAgentImagePullBehavior, ECSAgentLogLevel, ECSAttributeKey, ECSAttributeValue,
     ECSDurationValue, EtcHostsEntries, FriendlyVersion, Identifier, IntegerPercent, KmodKey,
     KubernetesAuthenticationMode, KubernetesBootstrapToken, KubernetesCloudProvider,
-    KubernetesClusterDnsIp, KubernetesClusterName, KubernetesDurationValue,
-    KubernetesEvictionHardKey, KubernetesLabelKey, KubernetesLabelValue, KubernetesQuantityValue,
-    KubernetesReservedResourceKey, KubernetesTaintValue, KubernetesThresholdValue, Lockdown,
-    OciDefaultsCapability, OciDefaultsResourceLimitType, PemCertificateString, SingleLineString,
-    SysctlKey, TopologyManagerPolicy, TopologyManagerScope, Url, ValidBase64, ValidLinuxHostname,
+    KubernetesClusterDnsIp, KubernetesClusterName, KubernetesDurationValue, KubernetesLabelKey,
+    KubernetesLabelValue, KubernetesQuantityValue, KubernetesReservedResourceKey,
+    KubernetesTaintValue, KubernetesThresholdValue, Lockdown, OciDefaultsCapability,
+    OciDefaultsResourceLimitType, PemCertificateString, SingleLineString, SysctlKey,
+    TopologyManagerPolicy, TopologyManagerScope, Url, ValidBase64, ValidLinuxHostname,
 };
 
 // Kubernetes static pod manifest settings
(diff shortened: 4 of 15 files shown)


sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -485,27 +485,27 @@ enum EvictionSignal {
...
             field: "Eviction Hard key",
         })?;
-        Ok(KubernetesEvictionHardKey {
+        Ok(KubernetesEvictionKey {
             inner: input.to_string(),
         })
     }
 }
-string_impls_for!(KubernetesEvictionHardKey, "KubernetesEvictionHardKey");
+string_impls_for!(KubernetesEvictionKey, "KubernetesEvictionKey");
```
webern: Can you convert these to use the new derive macro since they are being touched? You will no longer need a separate Enum and Struct, you can just have the enum.

See #2597, `sources/models/src/modeled_types/ecs.rs`

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -460,11 +460,11 @@ mod test_kubernetes_bootstrap_token {
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
-/// KubernetesEvictionHardKey represents a string that contains a valid Kubernetes eviction hard key.
+/// KubernetesEvictionKey represents a string that contains a valid Kubernetes eviction hard key.
```
gthao313: Nit: should remove "hard" from the description as well?
```suggestion
/// KubernetesEvictionKey represents a string that contains a valid Kubernetes eviction key.
```

sources/models/src/lib.rs
```diff
@@ -228,10 +228,14 @@ struct KubernetesSettings {
     authentication_mode: KubernetesAuthenticationMode,
     bootstrap_token: KubernetesBootstrapToken,
     standalone_mode: bool,
-    eviction_hard: HashMap<KubernetesEvictionHardKey, KubernetesThresholdValue>,
+    eviction_hard: HashMap<KubernetesEvictionKey, KubernetesThresholdValue>,
+    eviction_soft: HashMap<KubernetesEvictionKey, KubernetesThresholdValue>,
+    eviction_soft_grace_period: HashMap<KubernetesEvictionKey, KubernetesDurationValue>,
+    eviction_max_grace_period: i32,
```
etungsten: Do we need to constrain this value in anyway? Do we care about passing nonsensical values like `-100` to kubelet? We can use a modeled type to validate now that it supports scalar values.

stmcginnis: I think the only constraint is that it is a non-negative value.

Do you have an example of how to set that constraint?

etungsten: There is no existing setting that uses this yet but it is supported:
https://github.com/bottlerocket-os/bottlerocket/tree/develop/sources/models/scalar-derive#validating-a-number

gthao313: Maybe we can use regex to set the constraint? when I implemented eviction hard values, I used regex to limi the input value like this https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/src/modeled_types/kubernetes.rs#L538-L552

etungsten: We want to avoid using strings for value types that are supposed to be integers. Users have already complained about this inconsistency: https://github.com/bottlerocket-os/bottlerocket/issues/2883

stmcginnis: There are quite a few `i32` values in the kubernetes settings. I think nearly all of them have at least the constraint of needing to be non-negative. If we're going to enforce these constraints, we should probably do it for all of them.

Do we want to do that here, or would that be better kept it's only PR?

etungsten: Let's do it for this now. Need to start somewhere. The other ones were added before `scalar-derive` was a thing. That's on my plate to resolve.

stmcginnis: Error output isn't super great, but it does tell what the issue is:

```sh
[ssm-user@control]$ apiclient set --json '{"kubernetes": {"eviction-max-grace-period": -5}}'
Unable to deserialize input JSON into model: Unable to deserialize into PositiveInteger: number must be great than or equal to 0
[ssm-user@control]$ apiclient set --json '{"kubernetes": {"eviction-max-grace-period": 5}}'
[ssm-user@control]$ apiclient set --json '{"kubernetes": {"eviction-max-grace-period": 2147483647}}'
[ssm-user@control]$ apiclient set --json '{"kubernetes": {"eviction-max-grace-period": 2147483648}}'
Unable to deserialize input JSON into model: invalid value: integer `2147483648`, expected i32
```

Release.toml
```diff
@@ -192,4 +192,10 @@ version = "1.14.0"
 "(1.13.0, 1.13.1)" = [
     "migrate_v1.13.1_aws-profile-cred-provider.lz4",
 ]
+<<<<<<< HEAD
 "(1.13.1, 1.14.0)" = []
+=======
+"(1.13.1, 1.14.0)" = [
+    "migrate_v1.14.0_kubelet_config_settings.lz4",
+]
+>>>>>>> 2283ee47 (kubelet: Add `disable-cpu-cfs-enforcement` setting)
```
webern: oops

webern (approved):

sources/models/src/modeled_types/shared.rs
```diff
@@ -1130,3 +1133,44 @@ mod test_valid_kmod_key {
...
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// Input value that needs to be a positive value, but should not be greater
+/// than an i32::MAX.
+#[derive(Debug, PartialEq, Scalar)]
+pub struct PositiveInteger {
+    inner: i32,
```
webern, line 1143: You might be able to make this work for `T: PartionOrd + SomethingAboutTheParse`, but maybe we don't have any `i16`, `i64`, or whatever.

stmcginnis: I blame VS Code.

Release.toml
stmcginnis: I blame VS Code.

sources/models/src/modeled_types/shared.rs
stmcginnis: I thought about making it more flexible, but the problem with that is then the upper bounds could be too high.

There are some settings that take `i64` at least. But if I understand right, and please correct me if I'm wrong, if we made it more flexible it would make it possible to set an `i64` value for a field that eventually has to go into an `i32`. So I was thinking it would be safer to just be more verbose and have `PositiveInteger64`, `PositiveInteger16`, etc., as needed for whatever fields we want to contrain.

webern: I would have to try it out to see if I could make it work with type `T`. No problem, it can be changed to generic later and the change would be an implementation detail.

jpculp:
Should https://github.com/bottlerocket-os/bottlerocket/issues/2742 be part of this as well?

stmcginnis:
> Should https://github.com/bottlerocket-os/bottlerocket/issues/2742 be part of this as well?

My initial reaction was yes, but after reading through that some more, I think it might be safer to keep that one separate. It has some special casing since it's only applicable to 1.25 and later.

I think if this one merges first, that one could just get added to the same migration if it ends up being in the same release cycle. But I think it would be better to keep that separate in its own PR.

kahirokunn:
I am very happy with this PR as I just needed a shutdown-grace-period !

stmcginnis:
Rebased to resolve merge conflicts.

stmcginnis:
Added memory manager policy and settings.

README.md
```diff
@@ -456,50 +456,91 @@ The following settings are optional and allow you to further configure your clus
 
   See the [Roles Anywhere documentation](https://docs.aws.amazon.com/rolesanywhere/latest/userguide/credential-helper.html) for more details on the `aws_signing_helper` arguments.
 
+* `settings.kubernetes.disable-cpu-cfs-enforcement`: Whether to disable enforcing quotas for containers that specify CPU limits.
 * `settings.kubernetes.event-burst`: The maximum size of a burst of event creations.
 * `settings.kubernetes.event-qps`: The maximum event creations per second.
 * `settings.kubernetes.eviction-hard`: The signals and thresholds that trigger pod eviction.
-  Remember to quote signals (since they all contain ".") and to quote all values.
+* `settings.kubernetes.eviction-max-grace-period`: Maximum grace period, in seconds, to wait for pod termination before soft eviction. Default is `0`.
```
bcressey: Why not follow kubelet's name for the setting?

```suggestion
* `settings.kubernetes.eviction-max-pod-grace-period`: Maximum grace period, in seconds, to wait for pod termination before soft eviction. Default is `0`.
```

README.md
```diff
@@ -456,50 +456,91 @@ The following settings are optional and allow you to further configure your clus
 
   See the [Roles Anywhere documentation](https://docs.aws.amazon.com/rolesanywhere/latest/userguide/credential-helper.html) for more details on the `aws_signing_helper` arguments.
 
+* `settings.kubernetes.disable-cpu-cfs-enforcement`: Whether to disable enforcing quotas for containers that specify CPU limits.
```
bcressey: Typically we've preferred the positive form of this for settings, e.g. `enable` rather than `disable`.

It might be better to use a minor variation on the setting name, for clarity and discoverability. The goal would be to preserve the search phrase "cpu-cfs-quota" which aligns with the kubelet config:
```suggestion
* `settings.kubernetes.cpu-cfs-quota-enforced`: Whether CPU CFS quotas are enforced. Defaults to `true`.
```

README.md
```diff
@@ -456,50 +456,91 @@ The following settings are optional and allow you to further configure your clus
...
+
+  [settings.kubernetes.eviction-soft-grace-period]
+  "memory.available" = "30s"
+
+  [settings.kubernetes]
+  "eviction-max-grace-period" = 40
   ```
 
+  Remember to quote signals (since they all contain ".") and to quote all values.
+
```
bcressey: I find this sentence confusing now, since it's further away from the context of what signals are.

README.md
```diff
@@ -456,50 +456,91 @@ The following settings are optional and allow you to further configure your clus
...
+  ```
 
-If you downgrade from v1.14.0 to an earlier version, and you have these values set as numbers, they will be converted to strings on downgrade.
+  If you downgrade from v1.14.0 to an earlier version, and you have these values set as numbers, they will be converted to strings on downgrade.
 
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.kube-api-qps`: The QPS to use while talking with kubernetes apiserver.
 * `settings.kubernetes.log-level`: Adjust the logging verbosity of the `kubelet` process.
   The default log level is 2, with higher numbers enabling more verbose logging.
+* `settings.kubernetes.memory-manager-policy`: The memory management policy to use: `None` (default) or `Static`.
```
bcressey, line 529: Let's move the documentation down closer to `reserved-memory` to group related settings together.

For alignment with `cpu-manager-policy`, I'd also prefer to document "none" and "static" as the accepted values, even if we accept and emit the upper-cased form in code + templates.

README.md
```diff
@@ -456,50 +456,91 @@ The following settings are optional and allow you to further configure your clus
...
+  Misconfiguration of reserved memory settings may cause the Kubernetes `kubelet` process to fail.
+  It can be very difficult to recover from configuration errors.
+  Use the memory reservation information from `kubectl describe node` and make sure you understand the Kubernetes documentation related to the [memory manager](https://kubernetes.io/docs/tasks/administer-cluster/memory-manager/) and how to [reserve compute resources for system daemons](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/).
+
 * `settings.kubernetes.server-tls-bootstrap`: Enables or disables server certificate bootstrap. When enabled, the kubelet will request a certificate from the certificates.k8s.io API. This requires an approver to approve the certificate signing requests (CSR). Defaults to `true`.
+* `settings.kubernetes.shutdown-grace-period`: Delay the node should wait for pod termination before shutdown. Default is `0s`.
+* `settings.kubernetes.shutdown-critical-pod-grace-period`: The portion of the shutdown delay that should be dedicated to critical pod shutdown. Default is `0s`.
```
bcressey: Maybe one of these instead? The goal would be to avoid breaking up the search phrase "shutdown grace period".

```suggestion
* `settings.kubernetes.shutdown-grace-period-for-critical-pods`: The portion of the shutdown delay that should be dedicated to critical pod shutdown. Default is `0s`.
```

```suggestion
* `settings.kubernetes.critical-pod-shutdown-grace-period`: The portion of the shutdown delay that should be dedicated to critical pod shutdown. Default is `0s`.
```

(I have a slight preference for the first.)

README.md
```diff
@@ -456,50 +456,91 @@ The following settings are optional and allow you to further configure your clus
...
+  "memory-manager-policy" = "Static"
+
+  [settings.kubernetes.reserved-memory.0]
+  "enabled" = true
+  "memory" = "674Mi"
+  "hugepages-1Gi" = "1Gi"
+
+  [settings.kubernetes.reserved-memory.1]
+  "enabled" = true
+  "hugepages-2Mi" = "2148Mi"
```
bcressey: Am I reading this correctly?
* first example - 1 GiB memory reserved in the form of a single 1 GiB huge page, in addition to the 674 MiB reservation
* second example - 2148 MiB memory reserved in the form of 1,074 2 MiB huge pages

If not, it might be good to add comments inline on what these settings achieve.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1324,3 +1307,98 @@ pub struct CredentialProvider {
...
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+enum MemoryReservationResources {
+    #[serde(rename = "memory")]
+    Memory,
+    #[serde(rename = "hugepages-2Mi")]
+    HugePages2Mi,
+    #[serde(rename = "hugepages-1Gi")]
+    HugePages1Gi,
```
bcressey, line 1356: Can you confirm that `hugepages-2Mi` and `hugepages-1Gi` are not mutually exclusive? I was originally under the impression that they might be, but from the docs I've drawn the impression that these are treated as three distinct types of memory resource limits.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1324,3 +1307,98 @@ pub struct CredentialProvider {
...
+/// KubernetesMemoryReservationKey represents a string that contains a valid Kubernetes memory
+/// resource reservation key.
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct KubernetesMemoryReservationKey {
+    inner: String,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize)]
+enum MemoryReservationResources {
```
bcressey: Why do we need the "inner" pattern here but we were able to refactor `KubernetesEvictionHardKey` /  `EvictionSignal` into the new `KubernetesEvictionKey` enum?

sources/models/src/modeled_types/shared.rs
```diff
@@ -1130,3 +1133,44 @@ mod test_valid_kmod_key {
...
+        assert!(PositiveInteger::try_from(0).is_ok());
+        assert!(PositiveInteger::try_from(i32::MAX).is_ok());
+        assert!(PositiveInteger::try_from(42).is_ok());
+    }
+
+    #[test]
+    fn invalid_kmod_key() {
+        assert!(PositiveInteger::try_from(i32::MIN).is_err());
+        assert!(PositiveInteger::try_from(-1).is_err());
+    }
```
bcressey:
```suggestion
    #[test]
    fn valid_positive_integer() {
        assert!(PositiveInteger::try_from(0).is_ok());
        assert!(PositiveInteger::try_from(i32::MAX).is_ok());
        assert!(PositiveInteger::try_from(42).is_ok());
    }

    #[test]
    fn invalid_positive_integer() {
        assert!(PositiveInteger::try_from(i32::MIN).is_err());
        assert!(PositiveInteger::try_from(-1).is_err());
    }
```

sources/models/src/modeled_types/kubernetes.rs
webern: You are right, the struct can go away. See better patterns here:
- https://github.com/bottlerocket-os/bottlerocket/pull/2597
- https://github.com/bottlerocket-os/bottlerocket/pull/2697

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1324,3 +1307,98 @@ pub struct CredentialProvider {
...
+    #[serde(rename = "memory")]
+    Memory,
+    #[serde(rename = "hugepages-2Mi")]
+    HugePages2Mi,
+    #[serde(rename = "hugepages-1Gi")]
+    HugePages1Gi,
+}
```
webern:
```suggestion
/// KubernetesMemoryReservationKey represents a string that contains a valid Kubernetes memory
/// resource reservation key.
#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Scalar)]
enum KubernetesMemoryReservationKey {
    #[serde(rename = "memory")]
    Memory,
    #[serde(rename = "hugepages-2Mi")]
    HugePages2Mi,
    #[serde(rename = "hugepages-1Gi")]
    HugePages1Gi,
}
```

Again for better rendering:

```rust
/// KubernetesMemoryReservationKey represents a string that contains a valid Kubernetes memory
/// resource reservation key.
#[derive(Debug, Clone, Eq, PartialEq, Hash, Deserialize, Scalar)]
enum KubernetesMemoryReservationKey {
    #[serde(rename = "memory")]
    Memory,
    #[serde(rename = "hugepages-2Mi")]
    HugePages2Mi,
    #[serde(rename = "hugepages-1Gi")]
    HugePages1Gi,
}
```

README.md
stmcginnis: I double checked, but while `cpu-manager-policy` takes `none` and `static` (must be lower case), `memory-manager-policy` takes `None` and `Static` (must be title cased). So unfortunately there's an inconsistency, but it's an upstream inconsistency. Kubelet will fail to start if either of the settings are not cased correctly.

stmcginnis: > Let's move the documentation down closer to reserved-memory to group related settings together.

All current settings are in alphabetical order. I'd kind of hate to break that for one setting. What if I just add a "see `reserved-memory` for static memory manager configuration"?

README.md
stmcginnis: I had a comment on this in the code, but I'm not seeing it now. I must have accidentally dropped it in a rebase. 

The problem that led me to make it `disable-*` is due to handlebars. Having `cpu-cfs-quota-enforced = false` is a problem since the default is that it is enforced. So needing to set it to false makes the templating a challenge with it being evaluated as "falsey".

I'll take another look, but I don't think this worked:

```
{{#if (eq settings.kubernetes.cpu-cfg-quota-enforced false)}}
enforced = false
{{/if}}
```

I'll take another shot and see if I can find a better way.

stmcginnis: Figured it out - brain got stuck on the wrong pattern. :]

README.md
stmcginnis: > What if I just add...

Oh, I did that with the next line. Let me know if that's not enough. For now I will also add a note to `reserved-memory` to reference back to this setting.

sources/models/src/modeled_types/kubernetes.rs
stmcginnis: The docs aren't great here, but these are treated as distinct types of memory resource limits. The user would need to determine what is available by using `kubectl describe node` to see what is reported as the available values:

```
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource                    Requests    Limits
  --------                    --------    ------
  cpu                         325m (16%)  0 (0%)
  memory                      140Mi (2%)  340Mi (5%)
  ephemeral-storage           0 (0%)      0 (0%)
  hugepages-1Gi               0 (0%)      0 (0%)
  hugepages-2Mi               0 (0%)      0 (0%)
```

These would vary based on the type of hardware and/or the instance type.

README.md
stmcginnis: One other idea if you don't think the cross referencing is enough. It might actually be nice to prepend `memory-manager-reserved-memory`. That would allow us to keep alphabetical sorting while keep these related settings together. It would also make it clear that the `memory-manager-reserved-memory` settings are related to the `memory-manager-policy`.

Just let me know what you think.

stmcginnis:
Rebased to address conflict in `Release.toml` due to 1.13.3 and 1.13.4.

sources/models/src/modeled_types/kubernetes.rs
stmcginnis: @foersleo or @markusboehme - can you tell if anything else is needed to [support using hugepages](https://www.kernel.org/doc/Documentation/vm/hugetlbpage.txt)?

From what I've read, this would require passing in kernel parameters on boot, but I'm not clear on the full implications of that and how things are handled with Bottlerocket.

markusboehme: You're right, some kernel configuration is required. At some point the kernel needs to be told how many huge pages of which size to set aside. This does not necessarily have to happen at boot, it can also be done at any later point in time via procfs or sysfs. However, a huge page is a long stretch of physically contiguous memory, and physical memory will get fragmented over the course of normal system operation. This means the chance of successfully allocating huge pages decreases over time. Hence, it's best to do so early.

In Bottlerocket, the above translates to setting the kernel command line via the user data mechanism or setting 
corresponding values in sysfs. For example, to set aside 1024x 2 MiB huge pages at boot, one might use

```toml
[settings.boot]
reboot-to-reconcile = true

[settings.boot.kernel]
hugepages = ["1024"]
```

To not pay the price of a reboot (modifying the kernel command line happens after first boot), an alternative would be to set a sysctl (this happens later in the boot, though, after Bottlerocket's corndog is running):

```toml
[settings.kernel.sysctl]
"vm.nr_hugepages" = "1024"
```

Or with a bootstrap container writing `/sys/kernel/mm/hugepages/hugepages-*/nr_hugepages`, here with control over which size huge pages to set aside.

Some implications/things to watch out for:

* Huge pages set aside this way cannot be used for other, regular allocations. Corollaries: 1. If no pods use them, they're dead memory. 2. They add to memory pressure. Workloads can run out of memory earlier if they're not appropriately sized ("requests" in pod spec).
* Huge pages cannot be swapped out. Not a problem for Kubernetes, since swap is not supported anyway.
* Requesting too many huge pages at boot time (i.e. close to or more than the system memory) prevents the system from booting.

zmrow (approved):
LGTM!

README.md
bcressey: > It might actually be nice to prepend memory-manager-reserved-memory. That would allow us to keep alphabetical sorting while keep these related settings together.

Yes, please - I like this idea.

bcressey: > I double checked, but while cpu-manager-policy takes none and static (must be lower case), memory-manager-policy takes None and Static (must be title cased). So unfortunately there's an inconsistency, but it's an upstream inconsistency. Kubelet will fail to start if either of the settings are not cased correctly.

We should be able to use [serde aliases](https://serde.rs/field-attrs.html#alias) to paper over the inconsistency:
```
 #[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize, Scalar)]
 pub enum KubernetesMemoryManagerPolicy {
+    #[serde(alias = "static")]
     Static,
+    #[serde(alias = "none")]
     None,
 }
```

And perhaps give the CPU manager policy struct the same treatment. Then they will accept both forms (`"Static"` and `"static"`) when deserializing, but the expected form will be used for all serialization.

yeazelm (approved):

README.md
stmcginnis: `ValidCpuManagerPolicy` has `#[serde(rename_all = "lowercase")]` - that would be functionally equivalent, right?

sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
```diff
@@ -5,6 +5,14 @@ use std::process;
...
     migrate(AddSettingsMigration(&[
         "settings.kubernetes.cpu-manager-policy-options",
+        "settings.kubernetes.cpu-cfs-quota-enforced",
+        "settings.kubernetes.shutdown-grace-period",
+        "settings.kubernetes.shutdown-grace-period-for-critical-pods",
+        "settings.kubernetes.eviction-soft",
+        "settings.kubernetes.eviction-soft-grace-period",
+        "settings.kubernetes.eviction-max-pod-grace-period",
+        "settings.kubernetes.memory-manager-policy",
+        "settings.kubernetes.memory-manager-reserved-memory",
```
bcressey: This one needs to go in a different `AddPrefixes` migration since we allow user-defined keys underneath it.

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1358,3 +1339,75 @@ mod test_kubernetes_cpu_manager_policy_option {
...
+
+#[cfg(test)]
+mod test_kubernetes_memory_manager_policy {
+    use super::KubernetesMemoryManagerPolicy;
+    use std::convert::TryFrom;
+
+    #[test]
+    fn good_policy_key() {
+        for ok in &["Static", "None"] {
```
```suggestion
        for ok in &["Static", "static", "None", "none"] {
```

sources/models/src/lib.rs
```diff
@@ -255,6 +263,10 @@ struct KubernetesSettings {
     credential_providers: HashMap<Identifier, CredentialProvider>,
     server_certificate: ValidBase64,
     server_key: ValidBase64,
+    shutdown_grace_period: KubernetesDurationValue,
+    shutdown_grace_period_for_critical_pods: KubernetesDurationValue,
+    memory_manager_reserved_memory: HashMap<Identifier, KubernetesMemoryReservation>,
```
bcressey, line 268: Is `Identifier` the best type for the key or should it be another `PositiveInteger`?

sources/models/src/modeled_types/shared.rs
```diff
@@ -1130,3 +1133,44 @@ mod test_valid_kmod_key {
         assert!(KmodKey::try_from(vec!["z"; KMOD_KEY_LENGTH + 1].join("")).is_err());
     }
 }
+
+// =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+/// Input value that needs to be a positive value, but should not be greater
+/// than an i32::MAX.
+#[derive(Debug, PartialEq, Scalar)]
+pub struct PositiveInteger {
```
bcressey: Should this be named `NonNegativeInteger` since zero is allowed and is neither positive or negative?

I realize that's a bit of a math nitpick but I could see someone assuming that the PositiveInteger type would exclude zero without looking too closely at the implementation.

README.md
bcressey: > `ValidCpuManagerPolicy` has `#[serde(rename_all = "lowercase")]` - that would be functionally equivalent, right?

That doesn't quite do the same thing - it means serde will expect the lowercase form when deserializing, and serialize into the lowercase form. Which is correct for the CPU manager policy, but not the memory manager policy.

What we want is to always serialize / deserialize into the expected form, but *also* to accept certain aliases when deserializing. So we'd need to add `Static` and `None` aliases on `ValidCpuManagerPolicy` variants (it already accepts / expects "static" and "none").

stmcginnis: Done!

sources/models/src/lib.rs
stmcginnis: It does end up just being a numeric identifier. But I don't think we can enforce that here since integer values can't be used as HashMap keys. It is an "Identifier" to kubelet, so it's also probably better to have kubelet ignore an invalid setting than to fail boot if someone has something strange set up in their user data.

sources/models/src/modeled_types/kubernetes.rs
stmcginnis: Done

sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
stmcginnis: Added/moved.

sources/models/src/modeled_types/shared.rs
stmcginnis: Renamed to appease all mathematical purists. :D

bcressey (approved):

yeazelm (approved):

sanjeevpandey19:
Hi @stmcginnis, First of all, thanks for this PR. Just out of curiosity, wanted to know if this is going to be part of the 1.14.0 release.

stmcginnis:
> Hi @stmcginnis, First of all, thanks for this PR. Just out of curiosity, wanted to know if this is going to be part of the 1.14.0 release.

Hi @sanjeevpandey19! Yes, hoping to have this part of the 1.14.0 release. Everything is subject to change based on what is found and testing and other issues of course, but hoping to have this in a 1.14.0 release in the very near future. ;)

PR #2934, opened 2023-03-23
Enable environment variables for Kubernetes credential providers

### LINKED ISSUE #2602: Support env in `CredentialProviderConfig`
**What I'd like:**
Hi, I'd like to add some custom environment variables to the credential provider config, such as
```toml
[settings.kubernetes.credential-providers.samjo-credential-provider]
enabled = true
cache-duration = "30m"
image-patterns = ["samjo.private.registry"]

[settings.kubernetes.credential-providers.samjo-credential-provider.env]
my-env = "1"
```
and it transforms into the following.
```yaml
apiVersion: kubelet.config.k8s.io/v1alpha1
kind: CredentialProviderConfig
providers:
  - name: samjo-credential-provider
    matchImages:
      - "samjo.private.registry"
    defaultCacheDuration: "30m"
    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
    env:
      - name: MY_ENV
        value: "1"
```

Also, I have a question of the following line - how does this work?
https://github.com/bottlerocket-os/bottlerocket/blame/develop/packages/kubernetes-1.24/credential-provider-config-yaml#L18

Since the line is in the "#each" context on line 5, should it be `../settings.aws.profile`?

**Any alternatives you've considered:** (nothing)

stmcginnis:
**Issue number:**

Closes #2602 

**Description of changes:**

This adds the capability to provide environment variables that can be passed through to the Kubernetes credential provider configuration.

This makes it possible to provide user data settings like:

```toml
[settings.kubernetes.credential-providers.my-provider.env-variables]
"GOMAXPROCS" = "2"
```

and have that show up in the credential provider configuration file as:

```yaml
providers:
  - name: my-provider
    matchImages:
      - "*.dkr.ecr.us-east-2.amazonaws.com"
    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
    env:
      - name: GOMAXPROCS
        value: 2
```

**Testing done:**

Deployed node and verified launched successfully.

Ran:

```sh
apiclient apply <<EOL
  [settings.kubernetes.credential-providers.ecr-credential-provider]
  enabled = true
  # (optional - defaults to "12h")
  cache-duration = "30m"
  image-patterns = [
    # One or more URL paths to match an image prefix. Supports globbing of subdomains.
    "*.dkr.ecr.us-east-2.amazonaws.com",
    "*.dkr.ecr.us-west-2.amazonaws.com"
  ]
EOL
```

Checked `kubelet` status and cat'd `/etc/kubernetes/kubelet/credential-provider-config.yaml to make sure it reflected normal (pre-this change) configuration.

Ran:

```sh
apiclient apply <<EOL
  [settings.kubernetes.credential-providers.ecr-credential-provider.env-variables]
  # The following are not used with ecr-credential-provider, but are provided for illustration
  "KEY" = "abc123xyz"
  "GOMAXPROCS" = "2"
EOL
```

Checked `kubelet` status and cat'd `/etc/kubernetes/kubelet/credential-provider-config.yaml to see that it now contained the expected `env:` entries.

On new clean nodes, tried specifying all settings at once, and only current settings with adding environment variables later to make sure there was proper handling for template generation in all cases.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,11 @@ The following settings are optional and allow you to further configure your clus
     "*.dkr.ecr.us-east-2.amazonaws.com",
     "*.dkr.ecr.us-west-2.amazonaws.com"
   ]
+
+  [settings.kubernetes.credential-providers.ecr-credential-provider.environment]
+  # The following are not used with ecr-credential-provider, but are provided for illustration
+  "KEY" = "abc123xyz"
+  "GOMAXPROCS" = "2"
   ```
 
   **Note:** `ecr-credential-provider` is currently the only supported provider.
--- a/packages/kubernetes-1.22/credential-provider-config-yaml
+++ b/packages/kubernetes-1.22/credential-provider-config-yaml
@@ -11,13 +11,21 @@ providers:
 {{/each}}
     defaultCacheDuration: "{{default "12h" this.cache-duration}}"
     apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
-{{#if (eq @key "ecr-credential-provider")}}
+{{#if (or (eq @key "ecr-credential-provider") this.environment)}}
     env:
+{{#if this.environment}}
+{{#each this.environment}}
+      - name: {{@key}}
+        value: '{{this}}'
+{{/each}}
+{{/if}}
+{{#if (eq @key "ecr-credential-provider")}}
       - name: HOME
-        value: /root
+        value: '/root'
 {{#if @root.settings.aws.profile}}
       - name: AWS_PROFILE
-        value: {{@root.settings.aws.profile}}
+        value: '{{@root.settings.aws.profile}}'
+{{/if}}
 {{/if}}
 {{/if}}
 {{/if}}
--- a/sources/models/src/modeled_types/kubernetes.rs
+++ b/sources/models/src/modeled_types/kubernetes.rs
@@ -7,6 +7,7 @@ use serde::de::Error as _;
 use serde_json::Value;
 use snafu::{ensure, ResultExt};
 use std::borrow::Borrow;
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::fmt;
 use std::fmt::{Display, Formatter};
@@ -1315,12 +1316,15 @@ mod test_cluster_dns_ip {
     }
 }
 
+type EnvVarMap = HashMap<SingleLineString, SingleLineString>;
+
 /// CredentialProvider contains the settings for a credential provider for use
 /// in CredentialProviderConfig.
-#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct CredentialProvider {
     enabled: bool,
     image_patterns: Vec<SingleLineString>,
     cache_duration: Option<KubernetesDurationValue>,
+    environment: Option<EnvVarMap>,
 }
--- a/packages/kubernetes-1.23/credential-provider-config-yaml
+++ b/packages/kubernetes-1.23/credential-provider-config-yaml
@@ -11,13 +11,21 @@ providers:
 {{/each}}
     defaultCacheDuration: "{{default "12h" this.cache-duration}}"
     apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
-{{#if (eq @key "ecr-credential-provider")}}
+{{#if (or (eq @key "ecr-credential-provider") this.environment)}}
     env:
+{{#if this.environment}}
+{{#each this.environment}}
+      - name: {{@key}}
+        value: '{{this}}'
+{{/each}}
+{{/if}}
+{{#if (eq @key "ecr-credential-provider")}}
       - name: HOME
-        value: /root
+        value: '/root'
 {{#if @root.settings.aws.profile}}
       - name: AWS_PROFILE
-        value: {{@root.settings.aws.profile}}
+        value: '{{@root.settings.aws.profile}}'
+{{/if}}
 {{/if}}
 {{/if}}
 {{/if}}
--- a/packages/kubernetes-1.24/credential-provider-config-yaml
+++ b/packages/kubernetes-1.24/credential-provider-config-yaml
@@ -11,13 +11,21 @@ providers:
 {{/each}}
     defaultCacheDuration: "{{default "12h" this.cache-duration}}"
     apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
-{{#if (eq @key "ecr-credential-provider")}}
+{{#if (or (eq @key "ecr-credential-provider") this.environment)}}
     env:
+{{#if this.environment}}
+{{#each this.environment}}
+      - name: {{@key}}
+        value: '{{this}}'
+{{/each}}
+{{/if}}
+{{#if (eq @key "ecr-credential-provider")}}
       - name: HOME
-        value: /root
+        value: '/root'
 {{#if @root.settings.aws.profile}}
       - name: AWS_PROFILE
-        value: {{@root.settings.aws.profile}}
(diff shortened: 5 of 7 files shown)


packages/kubernetes-1.22/credential-provider-config-yaml
```diff
@@ -11,8 +11,15 @@ providers:
 {{/each}}
     defaultCacheDuration: "{{default "12h" this.cache-duration}}"
     apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
-{{#if (eq @key "ecr-credential-provider")}}
+{{#if (or (eq @key "ecr-credential-provider") this.env-variables)}}
```
webern: :sweat_smile:

sources/models/src/modeled_types/kubernetes.rs
```diff
@@ -1265,12 +1266,15 @@ mod test_cluster_dns_ip {
...
 /// CredentialProvider contains the settings for a credential provider for use
 /// in CredentialProviderConfig.
-#[derive(Debug, Clone, Eq, PartialEq, Hash, Serialize, Deserialize)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct CredentialProvider {
     enabled: bool,
     image_patterns: Vec<SingleLineString>,
     cache_duration: Option<KubernetesDurationValue>,
+    env_variables: Option<EnvVarMap>,
```
webern: Probably doesn't matter here, but these will be randomly ordered when iterated for the config file.

webern (approved):

sources/models/src/modeled_types/kubernetes.rs
stmcginnis: Yep, it shouldn't matter what order these come in as in the final rendering of the template. They all just end up being environment variables in the credential process.

I briefly considered using a `BTreeMap`, but it ultimately doesn't matter.

packages/kubernetes-1.22/credential-provider-config-yaml
stmcginnis: Right?!

gthao313 (approved):
LGTM!

README.md
```diff
@@ -434,6 +434,11 @@ The following settings are optional and allow you to further configure your clus
     "*.dkr.ecr.us-east-2.amazonaws.com",
     "*.dkr.ecr.us-west-2.amazonaws.com"
   ]
+
+  [settings.kubernetes.credential-providers.ecr-credential-provider.env-variables]
+  # The following are not used with ecr-credential-provider, but are provided for illustration
+  "KEY" = "abc123xyz"
```
bcressey: For consistency with the overall API "style", where settings tend to use fully expanded words rather than abbreviations, I'd suggest either `environment` or `environment-variables`. Either is OK but I have a slight preference for `environment`.

packages/kubernetes-1.22/credential-provider-config-yaml
```diff
@@ -11,8 +11,15 @@ providers:
...
     env:
+{{#if this.env-variables}}
+{{#each this.env-variables}}
+      - name: {{@key}}
+        value: {{this}}
+{{/each}}
+{{/if}}
+{{#if (eq @key "ecr-credential-provider")}}
       - name: HOME
```
bcressey, line 23: If `HOME` and `AWS_PROFILE` appear twice in this list - which can happen because they're set for `ecr-credential-provider` in the template, and also set in settings - does `kubelet` accept the YAML as well-formed or does that cause issues?

packages/kubernetes-1.22/credential-provider-config-yaml
```diff
@@ -11,8 +11,15 @@ providers:
 {{/each}}
     defaultCacheDuration: "{{default "12h" this.cache-duration}}"
     apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
-{{#if (eq @key "ecr-credential-provider")}}
+{{#if (or (eq @key "ecr-credential-provider") this.env-variables)}}
     env:
+{{#if this.env-variables}}
+{{#each this.env-variables}}
+      - name: {{@key}}
+        value: {{this}}
```
bcressey: There are values accepted by `SingleLineString`, such as `"`, that would result in an invalid YAML file. We could either filter those out at the settings level, or else defensively quote them here.

```suggestion
        value: '{{this}}'
```


Rendering `settings.aws.profile` likely needs the same fix.

packages/kubernetes-1.22/credential-provider-config-yaml
stmcginnis: I was thinking about that possibility too, though now I can't remember if I actually got around to testing out behavior. I will double check and report back on that to be safe. I suspect the environment variable just ends up with whatever is parsed last, but will make sure that really is the case.

stmcginnis:
```sh
$ apiclient set --json '{"aws": {"profile": "ecr"}}'
```

```sh
$ apiclient apply <<EOF
>   [settings.kubernetes.credential-providers.ecr-credential-provider]
>   enabled = true
>   # (optional - defaults to "12h")
>   cache-duration = "30m"
>   image-patterns = [
>     # One or more URL paths to match an image prefix. Supports globbing of subdomains.
>     "*.dkr.ecr.us-east-2.amazonaws.com",
>     "*.dkr.ecr.us-west-2.amazonaws.com"
>   ]
> 
>   [settings.kubernetes.credential-providers.ecr-credential-provider.environment]
>   # The following are not used with ecr-credential-provider, but are provided for illustration
>   "KEY" = "abc123xyz"
>   "GOMAXPROCS" = "2"
>   AWS_PROFILE = "whodis"
>   HOME = "/usr/libexec"
> EOF
```

```json
$ apiclient get settings.kubernetes.credential-providers
{
  "settings": {
    "kubernetes": {
      "credential-providers": {
        "ecr-credential-provider": {
          "cache-duration": "30m",
          "enabled": true,
          "environment": {
            "AWS_PROFILE": "whodis",
            "GOMAXPROCS": "2",
            "HOME": "/usr/libexec",
            "KEY": "abc123xyz"
          },
          "image-patterns": [
            "*.dkr.ecr.us-east-2.amazonaws.com",
            "*.dkr.ecr.us-west-2.amazonaws.com"
          ]
        }
      }
    }
  }
}
```

```yaml
kind: CredentialProviderConfig
providers:
  - name: ecr-credential-provider
    matchImages:
      - "*.dkr.ecr.us-east-2.amazonaws.com"
      - "*.dkr.ecr.us-west-2.amazonaws.com"
    defaultCacheDuration: "30m"
    apiVersion: credentialprovider.kubelet.k8s.io/v1alpha1
    env:
      - name: AWS_PROFILE
        value: 'whodis'
      - name: GOMAXPROCS
        value: '2'
      - name: HOME
        value: '/usr/libexec'
      - name: KEY
        value: 'abc123xyz'
      - name: HOME
        value: /root
      - name: AWS_PROFILE
        value: ecr
```

```text
# systemctl status kubelet
● kubelet.service - Kubelet
     Loaded: loaded (/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/kubelet.service; enabled; vendor preset: enabled)
    Drop-In: /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/kubelet.service.d
             └─dockershim-symlink.conf
             /etc/systemd/system/kubelet.service.d
             └─exec-start.conf
             /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/kubelet.service.d
             └─load-ipvs-modules.conf, make-kubelet-dirs.conf, prestart-pull-pause-ctr.conf
     Active: active (running) since Wed 2023-04-05 14:42:28 UTC; 2min 4s ago
```

Kubelet service is happy, template is rendered as expected, and no errors.

Duplicate environment variables mean one or the other will be the last one set, but that's a user configuration issue. The service itself does not have an issue with the extra configuration.

bcressey (approved):

Issue #2324, opened 2022-08-08
Bottlerocket settings: Add data type / example for all configuration settings in Bottlerocket documentation (README)


RobertNorthard:
**What I'd like:**

Not all Bottlerocket settings in the readme provide examples. For example, for the following settings image-gc-low-threshold-percent, image-gc-high-threshold-percent it is not clear if the expected value is a string or integer, which can cause confusion during Bottlerocket configuration and result in time spent debugging. Please can this be made clearer for all settings values?

**Any alternatives you've considered:**
N/A

zmrow:
Thanks @RobertNorthard !  This is a good idea - we'll look into making that happen.

stmcginnis:
I think the original concern has been addressed with the odd data types for some values with #2939, but it would be nice to have some clearer documentation around settings and their expected values.

@stockholmux do you think this would be good for the project-website?

stockholmux:
@stmcginnis That's fine although API documentation didn't make it into the first round of documentation. It will probably remain on the README for a while yet.

webern:
I believe the current status is: "scalar values are strings unless otherwise noted" and the k8s-gc-threshold settings are the only ones otherwise noted.

stmcginnis:
I think this is now addressed with all the work done in https://bottlerocket.dev/en/os/1.16.x/api/settings/ so closing this issue. If there is anything further missing there, please file an issue in the https://github.com/bottlerocket-os/bottlerocket-project-website repo.

PR #2939, opened 2023-03-24
allow k8s gc percent thresholds to be set as numbers

### LINKED ISSUE #2883: Why are `image-gc-high-threshold-percent` and `image-gc-low-threshold-percent` strings?

webern:
**Issue number:**

Closes #2883
Related to, but not comprehensive for #2324

**Description of changes:**

Make this intuitive userdata work:

```toml
[settings.kubernetes]
image-gc-high-threshold-percent = 85
image-gc-low-threshold-percent = 80
```

Prior to this change, the values had to be quoted as strings.

This change is backward compatible so that users relying on these being strings will not be broken by the change.

**Testing done:**

- [x] Booted with `settings.kubernetes.image-gc-high-threshold-percent = 85` and `settings.kubernetes.image-gc-low-threshold-percent = 80`
  - [x] Checked datastore
  - [x] Checked config files
  - [x] Ran a pod
- [x] Changed value `settings.kubernetes.image-gc-high-threshold-percent = "83"` with the API
  - [x] Checked datastore - it is still stored as a number (turns out `apiclient set` only sets scalar strings, so I also used apiclient to change the setting with JSON, and when I did I was able to convert it to a string)
  - [x] Checked config files - they look the same, but with the new value
- [x] Downgraded to v1.13.0
  - [x] Checked datastore - both values are now strings
  - [x] Checked config files - values look the same
  - [x] Ran a pod
- [x] Upgraded back up to v1.14.0
  - [x] Checked datastore - both values are still strings
  - [x] Checked config files - values look good
  - [x] Ran a pod
- [x] Booted a v1.13.0 AMI with `settings.kubernetes.image-gc-high-threshold-percent = "87"` and `settings.kubernetes.image-gc-low-threshold-percent = "79"`
  - [x] Checked datastore - these are strings
  - [x] Checked config file - correct
- [x] Upgraded to v1.14.0
  - [x] Checked datastore - these are still strings
  - [x] Checked config file, looks the same
  - [x] Ran a pod
- [x] Downgraded to v1.13.0
  - [x] Checked datastore - these are strings
  - [x] Checked config file - correct
  - [x] Ran a pod

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -468,8 +468,29 @@ The following settings are optional and allow you to further configure your clus
   "memory.available" = "15%"
   ```
 
-* `settings.kubernetes.image-gc-high-threshold-percent`: The percent of disk usage after which image garbage collection is always run.
-* `settings.kubernetes.image-gc-low-threshold-percent`: The percent of disk usage before which image garbage collection is never run.
+* `settings.kubernetes.image-gc-high-threshold-percent`: The percent of disk usage after which image garbage collection is always run, expressed as an integer from 0-100 inclusive.
+* `settings.kubernetes.image-gc-low-threshold-percent`: The percent of disk usage before which image garbage collection is never run, expressed as an integer from 0-100 inclusive.
+
+Since v1.14.0 `image-gc-high-threshold-percent` and `image-gc-low-threshold-percent` can be represented as numbers.
+For example:
+
+```toml
+[settings.kubernetes]
+image-gc-high-threshold-percent = 85
+image-gc-low-threshold-percent = 80
+```
+
+For backward compatibility, both string and numeric representations are accepted since v1.14.0.
+Prior to v1.14.0 these needed to be represented as strings, for example:
+
+```toml
+[settings.kubernetes]
+image-gc-high-threshold-percent = "85"
+image-gc-low-threshold-percent = "80"
+```
+
+If you downgrade from v1.14.0 to an earlier version, and you have these values set as numbers, they will be converted to strings on downgrade.
+
 * `settings.kubernetes.kube-api-burst`: The burst to allow while talking with kubernetes.
 * `settings.kubernetes.kube-api-qps`: The QPS to use while talking with kubernetes apiserver.
 * `settings.kubernetes.log-level`: Adjust the logging verbosity of the `kubelet` process.
--- a/sources/api/migration/migrations/v1.14.0/kubernetes-gc-percent-type-change/src/main.rs
+++ b/sources/api/migration/migrations/v1.14.0/kubernetes-gc-percent-type-change/src/main.rs
@@ -1,0 +1,54 @@
+use migration_helpers::{migrate, Migration, MigrationData, Result};
+use serde_json::Value;
+use std::process;
+
+const GC_HIGH_SETTING: &str = "settings.kubernetes.image-gc-high-threshold-percent";
+const GC_LOW_SETTING: &str = "settings.kubernetes.image-gc-low-threshold-percent";
+
+/// We changed these settings so that they can be specified as numbers. Previously they could only
+/// be specified as strings, which was confusing since they are numeric. On upgrade we don't need
+/// to do anything because a valid string representation will still be accepted. On downgrade, we
+/// need to check if the values are represented as numbers, and if so, convert them to strings.
+pub struct ChangeK8sGcPercentType;
+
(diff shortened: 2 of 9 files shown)


stmcginnis (approved):
Test coverage is good, and everything looks right to me.

README.md
```diff
@@ -468,8 +468,32 @@ The following settings are optional and allow you to further configure your clus
...
+Prior to 1.14.0 these needed to be represented as strings, for example:
+
+```toml
+[settings.kubernetes]
+image-gc-high-threshold-percent = "85"
+image-gc-low-threshold-percent = "80"
+```
+
+If you downgrade from v1.14.0 to an earlier version, and you have these values set as numbers,
+they will be converted to strings on downgrade.
```
```suggestion
If you downgrade from v1.14.0 to an earlier version, and you have these values set as numbers, they will be converted to strings on downgrade.
```

README.md
```diff
@@ -468,8 +468,32 @@ The following settings are optional and allow you to further configure your clus
   "memory.available" = "15%"
   ```
 
-* `settings.kubernetes.image-gc-high-threshold-percent`: The percent of disk usage after which image garbage collection is always run.
-* `settings.kubernetes.image-gc-low-threshold-percent`: The percent of disk usage before which image garbage collection is never run.
+* `settings.kubernetes.image-gc-high-threshold-percent`: The percent of disk usage after which image garbage collection is always run,
+  expressed as an integer from 0-100 inclusive.
+* `settings.kubernetes.image-gc-low-threshold-percent`: The percent of disk usage before which image garbage collection is never run,
+  expressed as an integer from 0-100 inclusive.
```
etungsten: No need for line-breaks to fit terminal size since we're rendering in markdown.

```suggestion
* `settings.kubernetes.image-gc-high-threshold-percent`: The percent of disk usage after which image garbage collection is always run, expressed as an integer from 0-100 inclusive.
* `settings.kubernetes.image-gc-low-threshold-percent`: The percent of disk usage before which image garbage collection is never run, expressed as an integer from 0-100 inclusive.
```

sources/api/migration/migrations/v1.14.0/kubernetes-gc-percent-type-change/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+            convert_to_string(v);
+        }
+        if let Some(v) = input.data.get_mut(GC_LOW_SETTING) {
+            convert_to_string(v);
+        }
+        Ok(input)
+    }
+}
+
+/// We added `affected-services` metadata for `aws.profile`
```
etungsten: Need to fix comment.

etungsten (approved):

README.md
```diff
@@ -468,8 +468,32 @@ The following settings are optional and allow you to further configure your clus
...
+For example:
+
+```toml
+[settings.kubernetes]
+image-gc-high-threshold-percent = 85
+image-gc-low-threshold-percent = 80
+```
+
+For backward compatibility, both string and numeric representations are accepted since v1.14.0.
+Prior to 1.14.0 these needed to be represented as strings, for example:
```
zmrow: Nit: missing the "v" before 1.14.0

zmrow (approved):

webern:
https://github.com/bottlerocket-os/bottlerocket/compare/f2bcfef6d41996ee9d5ca0113417e54c1dce2294..574022b05d4fe2c5ebb6aa6181cd64c79a386385

Fix suggestions

webern:
https://github.com/bottlerocket-os/bottlerocket/compare/574022b05d4fe2c5ebb6aa6181cd64c79a386385..e3cfc79c9a125c64c1d8bbb24780b96ca7bdbbca

Rebase

PR #2945, opened 2023-03-29
netdog: Module housekeeping

zmrow:
**Issue number:**
N/A

**Description of changes:**
This PR performs some minor module housekeeping in  `netdog` to make the code structure and import paths cleaner, as well as prepare for upcoming changes related to `systemd-networkd`.  

DHCP, static addressing, and bond configuration-related structs currently live in the `net_config` module, but are re-exported and used inside the `wicked` module.  It made more sense to me to have them live in their own standalone module if multiple modules (`net_config` and `wicked`) are going to make use of them; this will also be the case in the upcoming `networkd` related code.  This change moves DHCP and static config structures to a module named `addressing`, and bond structures to a module name `bonding`

The final small change is to rename the `net_config/devices/bonding` module to `net_config/devices/bond`.  This is purely cosmetic, but makes the `bond` module uniform with the other device-related modules with names that are nouns: `vlan`, `interface`.  In the context of a "device", `bond` also made more sent to me than `bonding`. :man_shrugging: 

**Testing done:**
* Build and boot an `aws-k8s-1.24` instance as a sanity check to make sure there were no runtime snafus

**Terms of contribution:**


yeazelm (approved):
LGTM

etungsten (approved):

PR #2956, opened 2023-03-29
aws-iam-authenticator: Upgrade to v0.6.8

jpmcb:
**Issue number:**

N/a

**Description of changes:**

```
Fixes building on the tip of `develop` when using GOPROXY="direct".
v0.6.2 of aws-iam-authenticator has a Go code dependency on kubernetes
1.22. There are privated GitHub repositories that can not be pulled down
through the "direct" method.

Signed-off-by: John McBride <jpmmcb@amazon.com>
```

Example of the broken build using `GOPROXY="direct"`:
```
❯ GOPROXY=direct cargo make -e PACKAGE=aws-iam-authenticator -e BUILDSYS_VARIANT=aws-k8s-1.22 build-package
...

   Compiling aws-iam-authenticator v0.1.0 (/home/ubuntu/workspace/bottlerocket-os/bottlerocket/packages/aws-iam-authenticator)
error: failed to run custom build command for `aws-iam-authenticator v0.1.0 (/home/ubuntu/workspace/bottlerocket-os/bottlerocket/packages/aws-iam-authenticator)`
 
Caused by:
  process didn't exit successfully: `/home/ubuntu/workspace/bottlerocket-os/bottlerocket/variants/target/x86_64/debug/build/aws-iam-authenticator-d52bf5b23e1e1e35/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=Cargo.toml
  cargo:rerun-if-changed=bundled-aws-iam-authenticator-0.6.2.tar.gz
  program: /home/ubuntu/workspace/bottlerocket-os/bottlerocket/tools/docker-go
  direct
 
 
  sum.golang.org
  /usr/src/module/aws-iam-authenticator-0.6.2 /usr/src/module
  go: k8s.io/kubernetes@v1.22.0 requires
        github.com/storageos/go-api@v2.2.0+incompatible: reading github.com/storageos/go-api/go.mod at revision v2.2.0: git ls-remote -q origin in /tmp/go/pkg/mod/cache/vcs/f4474392d641b836e8296e236f2861d83ba42188d7ff117e30acb6b06774a32d: exit status 128:
        fatal: could not read Username for 'https://github.com': terminal prompts disabled
  Confirm the import path was entered correctly.
  If this is a private repository, see https://golang.org/doc/faq#git_https for additional information.
  go: downloading k8s.io/apimachinery v0.23.0-alpha.0
  go: downloading k8s.io/client-go v0.22.0
[...]

PR #2957, opened 2023-03-29
GitHub actions: Use `GOPROXY=direct` during builds

jpmcb:
**Issue number:**

N/a

**Description of changes:**

Forces Go during builds to vendor code from direct upstream sources.
This is a useful and early signal to the team around problems with upstream
code repos.

Also note: this may make the builds for GitHub runners slightly longer.
The upstream Go proxy server has a CDN / cacheing / etc. which makes it
rather efficient. However, when Go gets code "directly", it's falling back
to using git to clone repositories (which can be slow for large codebases).
This seems like an acceptable tradeoff since we have very few Go code
packages and we get an early signal on problematic upstream repos.

**Testing done:**

N/a - check run in PR.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,6 +22,14 @@ concurrency:
   group: ${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  # When Go packages are built, buildsys will vendor in dependent Go code for
+  # that package and bundle it up in a tarball. This env variable is consumed
+  # and used to configure Go to directly download code from its upstream source.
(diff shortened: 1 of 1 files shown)


jpculp (approved):

jpmcb:
Note: the AWS variants will fail until we get #2956 in

jpmcb:
Force pushed to rebase and capture the changes that were merged from #2956 - should now see all actions (including AWS variants) passing.

rpkelly (approved):

jpculp (approved):

Issue #2617, opened 2022-11-30
SSM Parameter pointing at missing AMI?


nairb774:
**Image I'm using:**

Trying to use the Kubernetes 1.22 x86_64 AMI but the SSM parameter seems to point at a missing image.

**What I expected to happen:**

The SSM parameter ` /aws/service/bottlerocket/aws-k8s-1.22/x86_64/latest/image_version` would point to a version (`1.11.1-104f8e0f` right now), which would allo looking up an SSM parameter (`/aws/service/bottlerocket/aws-k8s-1.22/x86_64/1.11.1-104f8e0f/image_id`) which would point to an existing AMI that can be deployed.

**What actually happened:**

The last SSM parameter references an AMI that doesn't seem to exist (or is not yet public?).

**How to reproduce the problem:**

In the us-east-1 region:

```

$ aws ssm get-parameter --name /aws/service/bottlerocket/aws-k8s-1.22/x86_64/latest/image_version
{
    "Parameter": {
        "Name": "/aws/service/bottlerocket/aws-k8s-1.22/x86_64/latest/image_version",
        "Type": "String",
        "Value": "1.11.1-104f8e0f",
        "Version": 12,
        "LastModifiedDate": "2022-11-30T12:48:37.203000-08:00",
        "ARN": "arn:aws:ssm:us-east-1::parameter/aws/service/bottlerocket/aws-k8s-1.22/x86_64/latest/image_version",
        "DataType": "text"
    }
}

$ aws ssm get-parameter --name /aws/service/bottlerocket/aws-k8s-1.22/x86_64/1.11.1-104f8e0f/image_id
{
    "Parameter": {
        "Name": "/aws/service/bottlerocket/aws-k8s-1.22/x86_64/1.11.1-104f8e0f/image_id",
        "Type": "String",
        "Value": "ami-0776799044a14839c",
        "Version": 1,
        "LastModifiedDate": "2022-11-30T12:37:43.323000-08:00",
        "ARN": "arn:aws:ssm:us-east-1::parameter/aws/service/bottlerocket/aws-k8s-1.22/x86_64/1.11.1-104f8e0f/image_id",
        "DataType": "text"
    }
}

$ aws ec2 describe-images --filters "Name=image-id,Values=ami-0776799044a14839c"
{
    "Images": []
}

```

nairb774:
Looks like the 1.21 images are impacted as well:

```
[...]

PR #2969, opened 2023-03-30
pubsys: added SSM parameter validation

mjsterckx:
Inspired by bottlerocket-os/bottlerocket#2617
Related to bottlerocket-os/twoliter#493

**Description of changes:**

Added a `validate_ssm` mod to the `pubsys` crate. This mod adds a subcommand `validate-ssm` with the following signature:
```
pubsys-validate-ssm 0.1.0
Validates SSM parameters and AMIs

USAGE:
    pubsys --infra-config-path <infra-config-path> validate-ssm [OPTIONS] --validation-config-path <validation-config-path>

FLAGS:

OPTIONS:
        --validation-config-path <validation-config-path>    File holding the validation configuration
        --write-results-path <write-results-path>            
        --log-level <log-level>
```

- `validation-config-path` is the path to the file containing the validation configuration. This file should look like this:
```
{
    "validation_regions": [ "us-west-2", "us-east-1" ],
    "expected_parameter_lists": [ "./canary/ssm/ami_lists/1.11.0.json", "./canary/ssm/ami_lists/1.11.1.json", "./canary/ssm/ami_lists/1.12.0.json", "./canary/ssm/ami_lists/latest.json" ]
}
```

Each expected_parameter_list should have the following structure:
```
{
  "us-west-2": {
    "ami-12345678": {
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/arm64/1.12.0-6ef1139f/image_id": "ami-12345678",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/arm64/1.12.0-6ef1139f/image_version": "1.12.0-abcdefgh",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/arm64/1.12.0/image_id": "ami-12345678",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/arm64/1.12.0/image_version": "1.12.0-abcdefgh"
    },
    "ami-082d59d8979d777a6": {
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/x86_64/1.12.0-6ef1139f/image_id": "ami-87654321",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/x86_64/1.12.0-6ef1139f/image_version": "1.12.0-hgfedcba",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/x86_64/1.12.0/image_id": "ami-87654321",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/x86_64/1.12.0/image_version": "1.12.0-hgfedcba"
    },`
    ...
  },
  "us-east-1": {
    ...
  }
}
```

- `write-results-path` is the path to the file where the validation results will be written. The file will look like this:
```
[
  {
    "name": "/aws/service/bottlerocket/aws-k8s-1.22-nvidia/arm64/latest/image_id",
    "expected_value": "ami-12345678",
    "actual_value": "ami-23456789",
    "region": "eu-west-2",
    "ami_id": "ami-12345678",
    "status": "Incorrect"
  },
  {
    "name": "/aws/service/bottlerocket/aws-k8s-1.25-nvidia/arm64/latest/image_id",
    "expected_value": "ami-34567890",
    "actual_value": "ami-45678901",
    "region": "eu-central-1",
    "ami_id": "ami-34567890",
    "status": "Incorrect"
  },
  ...
]
```

- `write-results-filter` is a vec of potential statuses, which limits the validation results written to the above file. If the vec contains `Correct` and `Incorrect`, then only the validation results with those statuses will be written to the file and `Missing` and `Unexpected` validation results will not.

The command outputs a tabled summary of the validation results. This table will look like this:
```
+----------------+---------+-----------+---------+------------+------------+
| String         | correct | incorrect | missing | unexpected | accessible |
+----------------+---------+-----------+---------+------------+------------+
| ap-southeast-2 | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| eu-west-1      | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| eu-north-1     | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| us-east-1      | 2232    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| us-west-2      | 2232    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| ca-central-1   | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| us-east-2      | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| eu-west-3      | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| ap-south-1     | 2222    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| ap-northeast-3 | 1820    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| ap-southeast-1 | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| eu-west-2      | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| eu-central-1   | 2232    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| ap-northeast-2 | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| sa-east-1      | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| us-west-1      | 2212    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
| ap-northeast-1 | 2232    | 40        | 0       | 200        | true       |
+----------------+---------+-----------+---------+------------+------------+
```

The meaning of the different columns is this:
- `correct`: the expected value of the parameter is equal to the retrieved value
- `incorrect`: the expected value of the parameter is different from the retrieved value
- `missing`: the parameter was expected in that region but not retrieved
- `unexpected`: the retrieved parameter was not expected in that region
- `accessible`: SSM parameters were successfully retrieved from that region. If an invalid region was given, this would say `false` and all other columns in that row would show `-1`

**Testing done:**

- Unit tests
- Manually retrieved public SSM parameters and used them as input for the subcommand. Before the 1.13.0 release, all parameters showed as `correct`. After the 1.13.0 and 1.13.1 releases, each monitored region has 40 incorrect parameters (which is `latest`, because I didn't update my local copies of the parameters) and 200 `unexpected` parameters (which are the 1.13.0 and 1.13.1 parameters).

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/Cargo.toml
+++ b/tools/pubsys/Cargo.toml
@@ -41,6 +41,7 @@ serde_json = "1"
 simplelog = "0.12"
 snafu = "0.7"
 structopt = { version = "0.3", default-features = false }
+tabled = "0.10"
 tempfile = "3"
 tinytemplate = "1"
 tokio = { version = "1", features = ["full"] }  # LTS
--- a/tools/pubsys/src/aws/ssm/ssm.rs
+++ b/tools/pubsys/src/aws/ssm/ssm.rs
@@ -7,8 +7,8 @@ use aws_sdk_ssm::output::{GetParametersOutput, PutParameterOutput};
 use aws_sdk_ssm::types::SdkError;
 use aws_sdk_ssm::{Client as SsmClient, Region};
 use futures::future::{join, ready};
-use futures::stream::{self, StreamExt};
-use log::{debug, error, trace, warn};
+use futures::stream::{self, FuturesUnordered, StreamExt};
+use log::{debug, error, info, trace, warn};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::{HashMap, HashSet};
 use std::time::Duration;
@@ -132,6 +132,88 @@ where
         );
     }
 
+    Ok(parameters)
+}
+
+/// Fetches all SSM parameters under a given prefix using the given clients
+pub(crate) async fn get_parameters_by_prefix<'a>(
+    clients: &'a HashMap<Region, SsmClient>,
+    ssm_prefix: &str,
+) -> HashMap<&'a Region, Result<SsmParameters>> {
+    // Build requests for parameters; we have to request with a regional client so we split them by
+    // region
+    let mut requests = Vec::with_capacity(clients.len());
+    for region in clients.keys() {
+        trace!("Requesting parameters in {}", region);
+        let ssm_client: &SsmClient = &clients[region];
+        let get_future = get_parameters_by_prefix_in_region(region, ssm_client, ssm_prefix);
+
+        requests.push(join(ready(region), get_future));
+    }
+
+    // Send requests in parallel and wait for responses, collecting results into a list.
+    requests
+        .into_iter()
+        .collect::<FuturesUnordered<_>>()
+        .collect()
+        .await
+}
+
+/// Fetches all SSM parameters under a given prefix in a single region
+pub(crate) async fn get_parameters_by_prefix_in_region(
+    region: &Region,
+    client: &SsmClient,
+    ssm_prefix: &str,
+) -> Result<SsmParameters> {
+    info!("Retrieving SSM parameters in {}", region.to_string());
+    let mut parameters = HashMap::new();
+
+    // Send the request
+    let mut get_future = client
+        .get_parameters_by_path()
+        .path(ssm_prefix)
+        .recursive(true)
+        .into_paginator()
+        .send();
+
+    // Iterate over the retrieved parameters
+    while let Some(page) = get_future.next().await {
+        let retrieved_parameters = page
+            .context(error::GetParametersByPathSnafu {
+                path: ssm_prefix,
+                region: region.to_string(),
+            })?
+            .parameters()
+            .unwrap_or_default()
+            .to_owned();
+        for parameter in retrieved_parameters {
+            // Insert a new key-value pair into the map, with the key containing region and parameter name
+            // and the value containing the parameter value
+            parameters.insert(
+                SsmKey::new(
+                    region.to_owned(),
+                    parameter
+                        .name()
+                        .ok_or(error::Error::MissingField {
+                            region: region.to_string(),
+                            field: "name".to_string(),
+                        })?
+                        .to_owned(),
+                ),
+                parameter
+                    .value()
+                    .ok_or(error::Error::MissingField {
+                        region: region.to_string(),
+                        field: "value".to_string(),
+                    })?
+                    .to_owned(),
+            );
+        }
+    }
+    info!(
+        "SSM parameters in {} have been retrieved",
+        region.to_string()
+    );
     Ok(parameters)
 }
 
@@ -324,8 +406,8 @@ pub(crate) async fn validate_parameters(
     Ok(())
 }
 
-mod error {
-    use aws_sdk_ssm::error::GetParametersError;
+pub(crate) mod error {
+    use aws_sdk_ssm::error::{GetParametersByPathError, GetParametersError};
     use aws_sdk_ssm::types::SdkError;
     use snafu::Snafu;
     use std::error::Error as _;
@@ -334,12 +416,27 @@ mod error {
     #[derive(Debug, Snafu)]
     #[snafu(visibility(pub(super)))]
     #[allow(clippy::large_enum_variant)]
-    pub(crate) enum Error {
+    pub enum Error {
         #[snafu(display("Failed to fetch SSM parameters in {}: {}", region, source.source().map(|x| x.to_string()).unwrap_or("unknown".to_string())))]
         GetParameters {
             region: String,
             source: SdkError<GetParametersError>,
         },
+
+        #[snafu(display(
+            "Failed to fetch SSM parameters by path {} in {}: {}",
+            path,
+            region,
+            source
+        ))]
+        GetParametersByPath {
+            path: String,
+            region: String,
+            source: SdkError<GetParametersByPathError>,
+        },
+
+        #[snafu(display("Missing field in parameter in {}: {}", region, field))]
+        MissingField { region: String, field: String },
 
         #[snafu(display("Response to {} was missing {}", request_type, missing))]
         MissingInResponse {
@@ -369,4 +466,4 @@ mod error {
     }
 }
 pub(crate) use error::Error;
-type Result<T> = std::result::Result<T, error::Error>;
+pub(crate) type Result<T> = std::result::Result<T, error::Error>;
--- a/tools/pubsys/src/aws/validate_ssm/mod.rs
+++ b/tools/pubsys/src/aws/validate_ssm/mod.rs
@@ -1,0 +1,757 @@
+//! The validate_ssm module owns the 'validate-ssm' subcommand and controls the process of
+//! validating SSM parameters and AMIs
+
+pub mod results;
+
+use self::results::{SsmValidationResult, SsmValidationResultStatus, SsmValidationResults};
+use super::ssm::ssm::get_parameters_by_prefix;
+use super::ssm::{SsmKey, SsmParameters};
+use crate::aws::client::build_client_config;
+use crate::Args;
+use aws_sdk_ssm::{Client as SsmClient, Region};
+use log::{info, trace};
+use pubsys_config::InfraConfig;
+use serde::Deserialize;
+use snafu::ResultExt;
+use std::collections::{HashMap, HashSet};
+use std::fs::File;
+use std::path::PathBuf;
+use structopt::{clap, StructOpt};
+
+/// Validates SSM parameters and AMIs
+#[derive(Debug, StructOpt)]
+#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+pub struct ValidateSsmArgs {
+    /// File holding the validation configuration
+    #[structopt(long, parse(from_os_str))]
+    validation_config_path: PathBuf,
+
+    /// Optional path where the validation results should be written
+    #[structopt(long, parse(from_os_str))]
+    write_results_path: Option<PathBuf>,
+
+    #[structopt(long, requires = "write-results-path")]
+    /// Optional filter to only write validation results with these statuses to the above path
+    /// Available statuses are: `Correct`, `Incorrect`, `Missing`, `Unexpected`
+    write_results_filter: Option<Vec<SsmValidationResultStatus>>,
+
+    /// If this flag is added, print the results summary table as JSON instead of a
+    /// plaintext table
+    #[structopt(long)]
+    json: bool,
+}
+
+/// Structure of the validation configuration file
+#[derive(Debug, Deserialize)]
+pub(crate) struct ValidationConfig {
+    /// Vec of paths to JSON files containing expected metadata (image ids and SSM parameters)
+    expected_metadata_lists: Vec<PathBuf>,
+
+    /// Vec of regions where the parameters should be validated
+    validation_regions: Vec<String>,
+}
+
+/// A structure that allows us to store a parameter value along with the AMI ID it refers to. In
+/// some cases, then AMI ID *is* the parameter value and both fields will hold the AMI ID. In other
+/// cases the parameter value is not the AMI ID, but we need to remember which AMI ID it refers to.
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) struct SsmValue {
+    /// The value of the SSM parameter
+    pub(crate) value: String,
+
+    /// The ID of the AMI the parameter is associated with, used for validation result reporting
+    pub(crate) ami_id: String,
+}
+
+/// Performs SSM parameter validation and returns the `SsmValidationResults` object
+pub async fn validate(
+    args: &Args,
+    validate_ssm_args: &ValidateSsmArgs,
+) -> Result<SsmValidationResults> {
+    info!("Parsing Infra.toml file");
+
+    // If a lock file exists, use that, otherwise use Infra.toml
+    let infra_config = InfraConfig::from_path_or_lock(&args.infra_config_path, false)
+        .context(error::ConfigSnafu)?;
+
+    let aws = infra_config.aws.clone().unwrap_or_default();
+
+    trace!("Parsed infra config: {:#?}", infra_config);
+
+    // Read the validation config file and parse it into the `ValidationConfig` struct
+    let validation_config_file = File::open(&validate_ssm_args.validation_config_path).context(
+        error::ReadValidationConfigSnafu {
+            path: validate_ssm_args.validation_config_path.clone(),
+        },
+    )?;
+    let validation_config: ValidationConfig = serde_json::from_reader(validation_config_file)
+        .context(error::ParseValidationConfigSnafu)?;
+
+    let ssm_prefix = aws.ssm_prefix.as_deref().unwrap_or("");
+
+    // Parse the parameter lists found in the validation config
+    info!("Parsing expected parameter lists");
+    let expected_parameters = parse_parameter_lists(
+        validation_config.expected_metadata_lists,
+        &validation_config.validation_regions,
+    )
+    .await?;
+
+    info!("Parsed expected parameter lists");
+
+    // Create a Vec of Regions based on the region names in the validation config
+    let validation_regions: Vec<Region> = validation_config
+        .validation_regions
+        .iter()
+        .map(|s| Region::new(s.clone()))
+        .collect();
+
+    // Create a HashMap of SsmClients, one for each region where validation should happen
+    let base_region = &validation_regions[0];
+    let mut ssm_clients = HashMap::with_capacity(validation_regions.len());
+
+    for region in &validation_regions {
+        let client_config = build_client_config(region, base_region, &aws).await;
+        let ssm_client = SsmClient::new(&client_config);
+        ssm_clients.insert(region.clone(), ssm_client);
+    }
+
+    // Retrieve the SSM parameters using the SsmClients
+    info!("Retrieving SSM parameters");
+    let parameters = get_parameters_by_prefix(&ssm_clients, ssm_prefix).await;
+
+    // Validate the retrieved SSM parameters per region
+    info!("Validating SSM parameters");
+    let results: HashMap<Region, crate::aws::ssm::ssm::Result<HashSet<SsmValidationResult>>> =
+        parameters
+            .into_iter()
+            .map(|(region, region_result)| {
+                (
+                    region.clone(),
+                    region_result.map(|result| {
+                        validate_parameters_in_region(
+                            expected_parameters.get(region).unwrap_or(&HashMap::new()),
+                            &result,
+                        )
+                    }),
+                )
+            })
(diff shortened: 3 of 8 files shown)


mjsterckx:
^ clippy fixes

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
+//! The validate_ssm module owns the 'validate-ssm' subcommand and controls the process of validating
+//! SSM parameters and AMIs
+
+pub mod results;
+
+use self::results::{SsmValidationResult, SsmValidationResultStatus, SsmValidationResults};
+
+use super::ssm::ssm::get_parameters_by_prefix;
```
webern: Group these and let rust fmt sort them.

```suggestion
use self::results::{SsmValidationResult, SsmValidationResultStatus, SsmValidationResults};
use super::ssm::ssm::get_parameters_by_prefix;
```

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+
+/// Validates SSM parameters and AMIs
+#[derive(Debug, StructOpt)]
+#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+pub struct ValidateSsmArgs {
+    /// File holding the validation configuration
+    #[structopt(long, parse(from_os_str))]
+    validation_config_path: PathBuf,
+
+    // Optional path where the validation results should be written
```
webern: Actually important so that structopt will add this information to the help messages.

```suggestion
    /// Optional path where the validation results should be written
```

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+    #[structopt(long, parse(from_os_str))]
+    write_results_path: Option<PathBuf>,
+
+    #[structopt(long)]
+    // Optional filter to only write validation results with these statuses to the above path
```
webern: Actually important so that structopt will add this information to the help messages.

Also, **list the available options** in this comment.

```suggestion
    /// Optional filter to only write validation results with these statuses to the above path
```

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+#[derive(Debug, Deserialize)]
+pub(crate) struct ValidationConfig {
+    // Vec of paths to JSON files containing expected parameters
+    expected_parameter_lists: Vec<PathBuf>,
+
+    // Vec of regions where the parameters should be validated
+    validation_regions: Vec<String>,
+}
+
+/// Structure of an SSM parameter value
```
webern: When we have to do something strange, it's a good idea to explain in detail.

```suggestion
/// A structure that allows us to store a parameter value along with the AMI ID it refers to. In
/// some cases, then AMI ID *is* the parameter value and both fields will hold the AMI ID. In other
/// cases the parameter value is not the AMI ID, but we need to remember which AMI ID it refers to.
```

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+
+    trace!("Parsed infra config: {:#?}", infra_config);
+
+    // Read the validation config file and parse it into the `ValidationConfig` struct
+    let validation_config_file = File::open(&validate_ssm_args.validation_config_path)
+        .context(error::ReadValidationConfigSnafu)?;
+    let validation_config: ValidationConfig = serde_json::from_reader(validation_config_file)
+        .context(error::ParseValidationConfigSnafu)?;
+
+    let ssm_prefix = aws.ssm_prefix.as_deref().unwrap_or("");
```
webern, line 90: Will the rest of the code work properly if ssm_prefix is `""`? If not, then raise an error.

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+    println!("{}", results);
+    Ok(())
+}
+
+mod error {
+    use std::path::PathBuf;
+
+    use snafu::Snafu;
+
+    use crate::aws::ssm::ssm;
```
webern:
```suggestion
    use std::path::PathBuf;
    use snafu::Snafu;
    use crate::aws::ssm::ssm;
```

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub enum Error {
+        #[snafu(display("Error reading config: {}", source))]
+        Config { source: pubsys_config::Error },
+
+        #[snafu(display("Error reading validation config: {}", source))]
+        ReadValidationConfig { source: std::io::Error },
```
webern: Show the path if you have it. std::io::Error doesn't include the path.

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+        #[snafu(display("Found no parameters in source version {}", version))]
+        EmptySource { version: String },
+
+        #[snafu(display("Failed to fetch parameters from SSM: {}", source))]
+        FetchSsm { source: ssm::error::Error },
+
+        #[snafu(display("Infra.toml is missing {}", missing))]
+        MissingConfig { missing: String },
+
+        #[snafu(display("Failed to validate SSM parameters: {}", missing))]
```
webern, line 327: Sorry this comment is in a weird place. Do we already have a retry policy in place? I think there is something automatic/builtin to the SDK for retries.

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+        SerializeValidationResults { source: serde_json::Error },
+
+        #[snafu(display("Failed to write validation results to {:?}: {}", path, source))]
+        WriteValidationResults {
+            path: PathBuf,
+            source: std::io::Error,
+        },
+    }
+}
+pub(crate) use error::Error;
```
webern:
```suggestion
}

pub(crate) use error::Error;
```

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+type Result<T> = std::result::Result<T, error::Error>;
+
+#[cfg(test)]
+mod test {
+    use std::collections::{HashMap, HashSet};
+
+    use crate::aws::{
```
webern:
```suggestion
    use std::collections::{HashMap, HashSet};
    use crate::aws::{
```

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+        ssm::{SsmKey, SsmParameters},
+        validate_ssm::{results::SsmValidationResult, validate_parameters_in_region, SsmValue},
+    };
+    use aws_sdk_ssm::Region;
+
+    // These tests assert that the parameters can be validated correctly.
+
+    // Tests validation of parameters where the expected value is equal to the actual value
+    #[test]
+    fn validate_parameters_all_correct() {
```
webern, line 375: Nice tests!

tools/pubsys/src/aws/validate_ssm/results.rs
```diff
@@ -0,0 +1,681 @@
...
+    pub(crate) ami_id: Option<String>,
+
+    // The validation status of the parameter
+    pub(crate) status: SsmValidationResultStatus,
+}
+
+fn display_option(option: &Option<String>) -> &str {
+    match option {
+        Some(option) => option,
+        None => "N/A",
```
webern, line 83: Might be cleaner in the display?

```suggestion
        None => "",
```

tools/pubsys/src/aws/validate_ssm/results.rs
```diff
@@ -0,0 +1,681 @@
...
+    }
+}
+
+impl Serialize for SsmValidationResult {
```
webern: Why is this custom serializer necessary?

tools/pubsys/src/aws/validate_ssm/results.rs
```diff
@@ -0,0 +1,681 @@
...
+
+impl SsmValidationResult {
+    pub(crate) fn new(
+        name: String,
+        expected_value: Option<String>,
+        actual_value: Option<String>,
+        region: Region,
+        ami_id: Option<String>,
+    ) -> SsmValidationResult {
+        // Determine the validation status based on equality, presence, and absence of expected and actual parameter values
```
webern: Put a ruler at 100 and wrap comments at it.

tools/pubsys/src/aws/validate_ssm/results.rs
```diff
@@ -0,0 +1,681 @@
...
+                .map(|(region, results)| (region.to_string(), results))
+                .collect::<Vec<(String, &SsmValidationRegionSummary)>>(),
+        )
+        .to_string();
+        write!(f, "{}", table)
+    }
+}
+
+impl SsmValidationResults {
+    pub fn new(results: HashMap<Region, Result<HashSet<SsmValidationResult>>>) -> Self {
```
webern, line 198: Seems weird. Shouldn't the Result be unwrapped somewhere else?

```suggestion
    pub fn new(results: HashMap<Region, HashSet<SsmValidationResult>>) -> Self {
```

tools/pubsys/src/aws/ssm/ssm.rs
```diff
@@ -135,6 +135,84 @@ where
...
+                            field: "name".to_string(),
+                        })?
+                        .to_owned(),
+                ),
+                parameter
+                    .value()
+                    .ok_or(error::Error::MissingField {
+                        region: region.to_string(),
+                        field: "value".to_string(),
+                    })?
```
webern, line 208: Is it possible that a "null" value is something we could encounter and would want to report up through the results instead of erroring out here?

tools/pubsys/src/aws/validate_ssm/results.rs
mjsterckx: We keep the Result there to indicate the difference between "couldn't retrieve the SSM parameters in that region for some reason" and "some of the retrieved parameters were wrong/missing/unexpected".

tools/pubsys/src/aws/validate_ssm/results.rs
mjsterckx: Because SsmValidationResult contains a Region field, which is not serializable normally.

tools/pubsys/src/aws/validate_ssm/mod.rs
cbgbt: I actually organize my `use` statements the same way that @mjsterckx has done here. I think it's a holdover from developing in Python, where you have separate blocks for importing from:
* standard library
* third party packages
* your own package

I guess I've never looked around to see if that's how other folks do it.

tools/pubsys/src/aws/validate_ssm/results.rs
cbgbt: Check out [`#[serde(serialize_with)]`](https://serde.rs/field-attrs.html#serialize_with)

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+
+    #[structopt(long)]
+    // Optional filter to only write validation results with these statuses to the above path
+    write_results_filter: Option<Vec<SsmValidationResultStatus>>,
+}
+
+/// Structure of the validation configuration file
+#[derive(Debug, Deserialize)]
+pub(crate) struct ValidationConfig {
+    // Vec of paths to JSON files containing expected parameters
```
cbgbt: It's not as important as the structopt comments @webern pointed out, but it's good form to use docstring comments on attributes as well, since they would be added to any generated rustdocs.

tools/pubsys/src/aws/validate_ssm/mod.rs
mjsterckx: It will work properly. This is the same logic that was already present in `promote_ssm`.

tools/pubsys/src/aws/ssm/ssm.rs
mjsterckx: Based on the information in https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html, I don't think it's possible to have a parameter without the field "value", so this error will essentially never happen. If the "value" field contains a "null" value (which I assume means empty string), it'll be propagated in the results as usual.

tools/pubsys/src/aws/validate_ssm/results.rs
mjsterckx: I feel like this is a matter of opinion. An empty string could suggest that the empty string was actually retrieved as value from SSM, which would make the parameter "Incorrect" instead of "Missing"/"Unexpected".

tools/pubsys/src/aws/validate_ssm/mod.rs
mjsterckx: The SDK will retry retryable failures twice by default. I don't think any additional retry policies are necessary on pubsys' end.

tools/pubsys/src/aws/ssm/ssm.rs
```diff
@@ -135,6 +135,84 @@ where
...
+
+    // Iterate over the retrieved parameters
+    while let Some(page) = get_future.next().await {
+        for parameter in page
+            .context(error::GetParametersByPathSnafu {
+                path: ssm_prefix,
+                region: region.to_string(),
+            })?
+            .parameters()
+            .unwrap_or_default()
```
gthao313: NIT: I think this can be easier to read.
```suggestion
       let retrieved_parameters = page.context(error::GetParametersByPathSnafu {
                path: ssm_prefix,region: region.to_string(),
             })?
             .parameters()
             .unwrap_or_default();
        for parameter in retrieved_parameters {}
```

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+
+/// Parse the lists of parameters whose paths are in `parameter_lists`. Only parse the parameters
+/// in the regions present in `validation_regions`. Return a HashMap of Region mapped to a HashMap of
+/// the parameters in that region, with each parameter being a mapping of `SsmKey` to `SsmValue`.
+pub(crate) async fn parse_parameter_lists(
+    parameter_lists: Vec<PathBuf>,
+    validation_regions: &[String],
+) -> Result<HashMap<Region, HashMap<SsmKey, SsmValue>>> {
+    let mut parameter_map: HashMap<Region, HashMap<SsmKey, SsmValue>> = HashMap::new();
+    for list in parameter_lists {
```
gthao313: NIT: I would prefer to give it to a more expressive variable name. 
```suggestion
    for parameter_list in parameter_lists {
```

tools/pubsys/src/aws/ssm/ssm.rs
```diff
@@ -135,6 +135,84 @@ where
...
+                    .value()
+                    .ok_or(error::Error::MissingField {
+                        region: region.to_string(),
+                        field: "value".to_string(),
+                    })?
+                    .to_owned(),
+            );
+        }
+    }
+    info!("SSM parameters in {} retrieved", region.to_string());
```
gthao313:
```suggestion
    info!("SSM parameters in {} have been retrieved", region.to_string());
```

tools/pubsys/Cargo.toml
```diff
@@ -41,6 +41,7 @@ serde_json = "1"
 simplelog = "0.12"
```
gthao313, line 41: The code and tests look good to me! awesome!! 

I think it might be better to change the commit message to 
```
pubsys: added SSM parameter validation

"a short paragraph to describe what changes you have and those
changes will help to build canary something like that"
```

mjsterckx:
^ Addressed all comments and suggested changes

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -0,0 +1,757 @@
...
+    use snafu::Snafu;
+    use std::path::PathBuf;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub enum Error {
+        #[snafu(display("Error reading config: {}", source))]
+        Config { source: pubsys_config::Error },
+
+        #[snafu(display("Error reading validation config at path {:?}: {}", path, source))]
```
webern: Paths are weird because they aren't guaranteed to be UTF-8 encoded, here's how you do it:

```suggestion
        #[snafu(display("Error reading validation config at path {}: {}", path.display(), source))]
```

tools/pubsys/src/aws/validate_ssm/mod.rs
webern: > The SDK will retry retryable failures twice by default. I don't think any additional retry policies are necessary on pubsys' end.

Ok, I wonder if we want to increase that number of retries. Any guesses @cbgbt or @etungsten as to how many times we should retry on SSM calls?

webern (approved):
I updated the description to tag a couple of related issues.

I left one new comment, to use `path.display()`.

cbgbt:
@webern @mjsterckx 

> Any guesses as to how many times we should retry on SSM calls?

I usually set up retries using random jitter and exponential backoff (to avoid a [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem)) for some reasonable amount of time based on the use-case, rather than some static retry count. Maybe 30 seconds is within reason here?

We don't have great discipline with this in pubsys today. I think the functions to set parameters do their own retries, but most `GET`s rely on the SDK retry logic.

cbgbt (approved):

mjsterckx:
> I usually set up retries using random jitter and exponential backoff (to avoid a [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem)) for some reasonable amount of time based on the use-case, rather than some static retry count. Maybe 30 seconds is within reason here?

It seems like that logic should be added to all `get_parameters` functions and not just the ones added in this PR. Would that be acceptable to move to a later PR?

cbgbt (approved):

PR #2976, opened 2023-04-03
iptables: fix check for rule existence in ip6tables v1.8.9

### LINKED ISSUE #2975: ip6tables bug causing malfunctioning of host ports redirection on Kubernetes
**Image I'm using:**

BOTTLEROCKET_x86_64 1.13.1-32e9bb46 AMI on EKS managed node group.

Pod host ports working.

**What actually happened:**

Only the first host port of the pod is working.

On an EKS IPv6 cluster with a Bottlerocket node group, deploy a pod with several host ports. Only the first host port will work.

After tracking down the issue to missing iptables rules, I ended up instrumenting the portmap CNI plugin binary with logs in this function https://github.com/containernetworking/plugins/blob/8813bfea7b706ee02acfa9687369c7ce0983bf02/pkg/utils/iptables.go#L125:

```
func InsertUnique(ipt *iptables.IPTables, table, chain string, prepend bool, rule []string) error {
	log.Printf("inserting: %#v in %s\n", rule, chain)
	exists, err := ipt.Exists(table, chain, rule...)
	if err != nil {
		return err
	}
	if exists {
		log.Println("already exists")
		return nil
	}
	log.Println("does not exist")

	if prepend {
		log.Println("inserting")
		return ipt.Insert(table, chain, 1, rule...)
	} else {
		log.Println("appending")
		return ipt.Append(table, chain, rule...)
	}
}
```

I was surprised to see that the rules for the first host port were added correctly whereas the rules for the remaining host ports were said to already exist. Manually issuing the ip6tables commands seems to confirm the issue. There’s an open ip6tables bug report describing the problem: https://bugzilla.netfilter.org/show_bug.cgi?id=1667.

markusboehme:
**Issue number:**

Closes #2975 

**Description of changes:**

iptables v1.8.9 breaks the rule existence check in `ip6tables`. Fix this downstream until the issue has been patched in the upstream project.

I proposed this patch upstream: https://marc.info/?l=netfilter-devel&m=168055689214144

**Testing done:**

* [x] Locally: Reproducer from [upstream bug report](https://bugzilla.netfilter.org/show_bug.cgi?id=1667) shows expected behavior and reports the missing rule
* [x] Locally: Valgrind doesn't report any out-of-bounds reads anymore
* [x] Bottlerocket: `aws-k8s-1.24` on x86_64 builds and can launch pods
* [x] Bottlerocket: pod can have multiple host ports forwarded to it with IPv4 (confirmed via `iptables -t nat -L` on the host)
* [x] Bottlerocket: pod can have multiple host ports forwarded to it with IPv6 (on an IPv6 cluster; details see below)

Pod spec:

```
apiVersion: v1
kind: Pod
metadata:
  name: test-poly-host-ports
spec:
  restartPolicy: OnFailure
  containers:
    - name: test
      image: amazonlinux:2023
      command: ["sleep", "3600"]
      ports:
        - containerPort: 80
          hostPort: 10080
        - containerPort: 81
          hostPort: 10081
```

On the host (admin container --> `sudo sheltie`):

```
bash-5.1# ip6tables -t nat -L -n
[...]
Chain CNI-DN-b4a7dc28d99c75afb1337 (1 references)
target     prot opt source               destination         
CNI-HOSTPORT-SETMARK  6    --  2a05:d014:a18:a000:8e24::2  ::/0                 tcp dpt:10080
DNAT       6    --  ::/0                 ::/0                 tcp dpt:10080 to:[2a05:d014:a18:a000:8e24::2]:80
CNI-HOSTPORT-SETMARK  6    --  2a05:d014:a18:a000:8e24::2  ::/0                 tcp dpt:10081
DNAT       6    --  ::/0                 ::/0                 tcp dpt:10081 to:[2a05:d014:a18:a000:8e24::2]:81

Chain CNI-HOSTPORT-DNAT (2 references)
target     prot opt source               destination         
CNI-DN-b4a7dc28d99c75afb1337  6    --  ::/0                 ::/0                 /* dnat name: "aws-cni" id: "77c07952c638f66094ba7b92262990f5e9c4e680a038d91a81f1162d4beef863" */ multiport dports 10080,10081
[...]
```

**Terms of contribution:**

### DIFF
--- a/packages/iptables/1002-ip6tables-Fix-checking-existence-of-rule.patch
+++ b/packages/iptables/1002-ip6tables-Fix-checking-existence-of-rule.patch
@@ -1,0 +1,31 @@
+From ba75342ff3e01605258810eb7f5683d8e326ffd8 Mon Sep 17 00:00:00 2001
+From: Markus Boehme <markubo@amazon.com>
+Date: Mon, 3 Apr 2023 22:20:23 +0200
+Subject: [PATCH] ip6tables: Fix checking existence of rule
+
+Pass the proper entry size when creating a match mask for checking the
+existence of a rule. Failing to do so causes wrong results.
+
+Reported-by: Jonathan Caicedo <jonathan@jcaicedo.com>
+Fixes: eb2546a846776 ("xshared: Share make_delete_mask() between ip{,6}tables")
+Signed-off-by: Markus Boehme <markubo@amazon.com>
+---
+ iptables/ip6tables.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/iptables/ip6tables.c b/iptables/ip6tables.c
+index 345af451..9afc32c1 100644
+--- a/iptables/ip6tables.c
++++ b/iptables/ip6tables.c
+@@ -331,7 +331,7 @@ check_entry(const xt_chainlabel chain, struct ip6t_entry *fw,
+ 	int ret = 1;
+ 	unsigned char *mask;
+ 
+-	mask = make_delete_mask(matches, target, sizeof(fw));
++	mask = make_delete_mask(matches, target, sizeof(*fw));
+ 	for (i = 0; i < nsaddrs; i++) {
+ 		fw->ipv6.src = saddrs[i];
+ 		fw->ipv6.smsk = smasks[i];
+-- 
+2.25.1
+
--- a/packages/iptables/iptables.spec
+++ b/packages/iptables/iptables.spec
@@ -17,6 +17,7 @@ Requires: %{_cross_os}libnftnl
 Requires: %{_cross_os}libnetfilter_conntrack
 
 Patch1001: 1001-extensions-NAT-Fix-for-Werror-format-security.patch
+Patch1002: 1002-ip6tables-Fix-checking-existence-of-rule.patch
 
 %description
 %{summary}.


etungsten (approved):
LGTM if IPv6 testing goes well

stmcginnis (approved):
LGTM

jpmcb (approved):

bcressey (approved):

markusboehme:
I successfully tested the change on an IPv6 EKS cluster and updated the PR description with the steps I took.

yann-soubeyrand:
Hello,

I don’t know if this is the right place to ask this question, but do you know when new AMI will be released and available on EKS node groups and if the current AMI will be retracted in the interim?

markusboehme:
@yann-soubeyrand: There is no plan to retract the AMIs for Bottlerocket v1.13.x or their SSM parameters at this point as this may break other users.

The fix unfortunately missed the window for the just released v1.13.2. It will for sure be included in the next official release, be that another v1.13.x point release or v1.14.0. There is no fixed date for such a release yet, but we are looking into providing a v1.13.3. Please expect more information on that early next week.

For the time being, you may avoid the issue by creating a custom launch template referencing an AMI for Bottlerocket v1.12.0. The [documentation for AWS EKS Managed Node Groups](https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html) contains more information on this. You can determine suitable AMI IDs by querying SSM parameters, e.g. `aws ssm get-parameter --region eu-central-1 --name '/aws/service/bottlerocket/aws-k8s-1.25/x86_64/1.12.0/image_id' --query Parameter.Value --output text` (substituting the desired Bottlerocket variant, and `arm64` for `x86_64` if you're using Graviton-/Arm-based instances).

markusboehme:
@yann-soubeyrand A new point release of Bottlerocket (1.13.3) is in the works, [including this fix](https://github.com/bottlerocket-os/bottlerocket/commit/a12c40e1058730299c875f5406daaea968c49efc).

yann-soubeyrand:
Thanks @markusboehme!

PR #2979, opened 2023-04-05
tools/diff-kernel-config: enhance functionality and output formats

foersleo:
**Issue number:** - 

**Description of changes:**

Make the `diff-config-tool` a bit more useful by:

* shifting its perspective on config changes from a kernel version anchored view to a variant anchored view. This has the advantage that we do not need to hard-code a translation of variants for specific kernel versions. In addition this comes in handy when doing major kernel version updates. In those cases we will usually change the kernel used in `-dev` variants, allowing for easy comparison of configs between major kernel versions.
* adding a more structured output format for config changes. Presenting changes in tabular form makes comparison between multiple variants easier to see which config changes are shared between variants and architectures and which are distinct to specific configurations.

_Review hint: The diff for the first patch of the series seems bigger than the interesting parts are. This is an unfortunate side effect of changing indentation on some loop nesting._

**Testing done:**

I am using the new tooling for the major kernel update to kernel 6.1. It helps a lot to get the reports without extra manual steps for major kernel updates and getting the tabular data provides a good starting point for working through the massive amount of config changes in a structured way.

The summary for an example invocation looking at dev variants looks familiar:
```
./tools/diff-kernel-config -a HEAD -b develop -v aws-dev -v metal-dev -v vmware-dev -o config-diff-test
[...]
config-aarch64-aws-dev-diff:     92 removed, 226 added,   2 changed
config-aarch64-metal-dev-diff:   92 removed, 227 added,   2 changed
config-x86_64-aws-dev-diff:      81 removed, 231 added,   3 changed
config-x86_64-metal-dev-diff:    81 removed, 233 added,   3 changed
config-x86_64-vmware-dev-diff:   97 removed, 229 added,   4 changed
```

Excerpt from the newly created `diff_table` can be seen on [gist](https://gist.github.com/foersleo/c27315baaada31945e0d4f7eca9ae50a).

**Terms of contribution:**

### DIFF
--- a/tools/diff-kernel-config
+++ b/tools/diff-kernel-config
@@ -32,8 +32,9 @@ Compare kernel configurations before and after a series of commits.
 
     -a, --after         new Git revision to compare from
     -b, --before        baseline Git revision to compare against
-    -k, --kernel        kernel versions to compare configs for, may be given
-                        multiple times (optional, defaults to all kernels)
+    -v, --variant       variant to pick kernel from to compare configs for, may
+                        be given multiple times (optional, defaults to this list:
+                        'aws-k8s-1.23', 'metal-k8s-1.23', 'aws-dev', 'metal-dev')
     -o, --output-dir    path to the output directory; must not exist yet
     -h, --help          show this help text
 
@@ -66,6 +67,7 @@ usage_error() {
 #
 
 kernel_versions=()
+variants=()
 
 while [[ $# -gt 0 ]]; do
     case $1 in
@@ -73,8 +75,8 @@ while [[ $# -gt 0 ]]; do
             shift; gitrev_after_arg=$1 ;;
         -b|--before)
             shift; gitrev_before_arg=$1 ;;
-        -k|--kernel)
-            shift; kernel_versions+=( "$1" ) ;;
+        -v|--variant)
+            shift; variants+=( "$1" ) ;;
         -o|--output-dir)
             shift; output_dir=$1 ;;
         -h|--help)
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
lgtm

foersleo:
Seems like the shellcheck integration in my editor is broken... Will fix the script accordingly.

foersleo:
Force push fixed the issues called out by shellcheck. CI should hopefully get through happily.

tools/diff-kernel-config
```diff
@@ -266,3 +256,24 @@ echo
...
+    sed -i "s/$/,/" "${output_dir}"/diff_table
+    sed -i "/^config change/ s/$/${col_name}/" "${output_dir}"/diff_table
+
+    mapfile -t diff_lines < "${config_diff}"
+
+    for line in "${diff_lines[@]}"; do
+        sed -i "/^${line}/ s/$/x/" "${output_dir}"/diff_table
+    done
+done
+echo "A tabular report in csv-format has been placed in '${output_dir}/diff-table'"
```
markusboehme, line 279: Nit pick: The actual file generated is named `diff_table` with an underscore (though it would be nice for all files to follow the same convention of either using dashes or underscores).

tools/diff-kernel-config
```diff
@@ -266,3 +256,24 @@ echo
...
+cat "${output_dir}"/*-diff | sort | uniq >> "${output_dir}"/diff_table
+
+for config_diff in "${output_dir}"/config-*-diff; do
+    variant_name=$(echo "${config_diff}" | sed -e "s%^${output_dir}/config-%%" -e "s%-diff$%%")
+    kver_before=$(grep "${variant_name}-before" "${output_dir}/kver_mapping" | cut -d ' ' -f 3)
+    kver_after=$(grep "${variant_name}-after" "${output_dir}/kver_mapping" | cut -d ' ' -f 3)
+    col_name="${variant_name} (${kver_before} -> ${kver_after})"
+
+    sed -i "s/$/,/" "${output_dir}"/diff_table
+    sed -i "/^config change/ s/$/${col_name}/" "${output_dir}"/diff_table
```
markusboehme: Extending the table column by column is clever! I don't think I've seen that before.

markusboehme (approved):
Nice one! Agree it makes more sense to look at things from the angle of variants rather than kernel versions, not least because kernel configs may change for the same version due to platform-specific config fragments.

foersleo:
Force push addressed the inconsistency in naming output files (underscores vs. dashes). Change `diff_table` to `diff-table` to be in line with the other output files.

As there is no real functional change I am going ahead merging this without re-review after checks have passed.

PR #2982, opened 2023-04-05
Remove extraneous parameter property from OpenAPI spec

### LINKED ISSUE #2981: Bottlerocket OpenAPI spec does not validate
**Platform I'm building on:** n/a (ironically, hugo but see below)

I'm attempting to import [Bottlerocket's OpenAPI spec](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/openapi.yaml) into [the documentation site](https://github.com/bottlerocket-os/project-website). Our website builder has native support for OpenAPI and I should be able to just point the template at the spec to generate documentation.

**What actually happened:**

I get the follow error in Hugo:

Screenshot 2023-04-05 at 1 07 00 PM

I ran it through multiple OpenAPI 3.0.x compatible [validators](https://apitools.dev/swagger-parser/online/) and [editors](https://editor.swagger.io/) and get the same error.

It seems that OpenAPI spec for Bottlerocket isn't compliant to the standard.

See above.

stmcginnis:
**Issue number:**

Closes #2981

**Description of changes:**

The `configuration-files` endpoint has two query parameters. These should be listed as two items under the single `parameters` property, but it looks like there was a copy/paste error that added a second `parameters` property. This causes parsing errors for the API spec.

**Testing done:**

Pasted output into OpenAPI editor and verified it was able to be validated against the schema.

**Terms of contribution:**

### DIFF
--- a/sources/api/openapi.yaml
+++ b/sources/api/openapi.yaml
@@ -339,7 +339,6 @@ paths:
           style: form
           explode: false
           required: false
-      parameters:
         - in: query
           name: prefix
           description: "Specific key prefix to query. This parameter will be ignored if 'names' is also supplied"


stockholmux (approved):
Confirmed. This works in Hugo.

markusboehme (approved):

PR #2983, opened 2023-04-06
pubsys: added EC2 image validation

mjsterckx:
Added a `validate-ami` subcommand to pubsys to validate EC2 images, given a config file with regions and paths to files containing image ids.

**Description of changes:**

Added a validate_ami mod to the pubsys crate. This mod adds a subcommand validate-ami with the following signature:
```
pubsys-validate-ami 0.1.0
Validates EC2 images

USAGE:
    pubsys --infra-config-path <infra-config-path> validate-ami [FLAGS] [OPTIONS] --validation-config-path <validation-config-path>

FLAGS:
        --json       If this argument is given, print the validation results summary as a JSON object instead of a
                     plaintext table

OPTIONS:
        --validation-config-path <validation-config-path>    File holding the validation configuration
        --write-results-path <write-results-path>
            Optional path where the validation results should be written

            Optional filter to only write validation results with these statuses to the above path The available
            statuses are: `Correct`, `Incorrect`, `Missing`
        --log-level <log-level>
```
- validation-config-path is the path to the file containing the validation configuration. This file should look like this:
```
{
    "validation_regions": [ "us-west-2", "us-east-1" ],
    "expected_metadata_lists": [ "./canary/ami/ami_lists/1.11.0.json", "./canary/ami/ami_lists/1.11.1.json", "./canary/ami/ami_lists/1.12.0.json", "./canary/ami/ami_lists/latest.json" ]
}
```
Each expected_metadata_list should have the following structure:
```
{
  "us-west-2": {
    "ami-12345678": {
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/arm64/1.12.0-6ef1139f/image_id": "ami-12345678",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/arm64/1.12.0-6ef1139f/image_version": "1.12.0-abcdefgh",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/arm64/1.12.0/image_id": "ami-12345678",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/arm64/1.12.0/image_version": "1.12.0-abcdefgh"
    },
    "ami-082d59d8979d777a6": {
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/x86_64/1.12.0-6ef1139f/image_id": "ami-87654321",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/x86_64/1.12.0-6ef1139f/image_version": "1.12.0-hgfedcba",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/x86_64/1.12.0/image_id": "ami-87654321",
      "/aws/service/bottlerocket/aws-ecs-1-nvidia/x86_64/1.12.0/image_version": "1.12.0-hgfedcba"
    },`
    ...
  },
  "us-east-1": {
    ...
  }
}
```
The parameters inside the image-id objects are irrelevant, but this same file structure is used for SSM validation (https://github.com/bottlerocket-os/bottlerocket/pull/2969) so this allows both validations to use the same input files.
- write-results-path is the path to the file where the validation results will be written. The file will look like this:
```
[
  {
    "image_id": "ami-0f4c91eb881453540",
    "expected_value": {
      "ImageId": "ami-0f4c91eb881453540",
      "Public": true,
      "EnaSupport": true,
      "SriovNetSupport": "simple"
    },
    "actual_value": {
      "ImageId": "ami-0f4c91eb881453540",
      "Public": true,
      "EnaSupport": true,
      "SriovNetSupport": "simple"
    },
    "region": "us-east-1",
    "status": "Correct"
  },
  ...
]
```
- write-results-filter is a vec of potential statuses, which limits the validation results written to the above file. If the vec contains Correct and Incorrect, then only the validation results with those statuses will be written to the file and Missing validation results will not.

The command outputs a tabled summary of the validation results. This table will look like this (unless the `--json` flag is passed, in which case a JSON object will be printed):
```
+----------------+---------+-----------+---------+------------+
| String         | correct | incorrect | missing | accessible |
+----------------+---------+-----------+---------+------------+
| ap-northeast-3 | 448     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| eu-west-2      | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| ap-northeast-1 | 558     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| eu-west-1      | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| ca-central-1   | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| ap-southeast-1 | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| ap-south-1     | 553     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| us-east-1      | 558     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| sa-east-1      | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| us-east-2      | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| us-west-1      | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| eu-west-3      | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| ap-southeast-2 | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| eu-north-1     | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| us-west-2      | 558     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| eu-central-1   | 558     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
| ap-northeast-2 | 548     | 0         | 0       | true       |
+----------------+---------+-----------+---------+------------+
```
The meaning of the different columns is this:

- correct: the expected validated values of the images are equal to the values of the retrieved image
- incorrect: the expected value of the parameter is different from the retrieved value
- missing: the parameter was expected in that region but not retrieved
- accessible: SSM parameters were successfully retrieved from that region. If an invalid region was given, this would say false and all other columns in that row would show -1

The validation will check the following 3 fields with their expected values:
- `Public`: true
- `EnaSupport`: true
- `SriovNetSupport`: `simple`

**Testing done:**

- Unit tests
- Compiled a list of Bottlerocket image ids for each version in each validation region based on the public SSM parameters and used this as input for the command. Results are shown in the table above (each image validated correctly).

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/validate_ami/ami.rs
+++ b/tools/pubsys/src/aws/validate_ami/ami.rs
@@ -1,0 +1,148 @@
+//! The ami module owns the describing of images in EC2.
+
+use aws_sdk_ec2::model::Image;
+use aws_sdk_ec2::{Client as Ec2Client, Region};
+use futures::future::{join, ready};
+use futures::stream::{FuturesUnordered, StreamExt};
+use log::{info, trace};
+use serde::{Deserialize, Serialize};
+use snafu::ResultExt;
+use std::collections::HashMap;
+
+/// Structure of the EC2 image fields that should be validated
+#[derive(Serialize, Deserialize, PartialEq, Eq, Hash, Debug, Clone)]
+#[serde(rename_all = "PascalCase")]
+pub(crate) struct ImageDef {
+    /// The id of the EC2 image
+    pub(crate) image_id: String,
+
+    /// Whether or not the EC2 image is public
+    pub(crate) public: bool,
+
+    /// Whether or not the EC2 image supports Elastic Network Adapter
+    pub(crate) ena_support: bool,
+
+    /// The level of the EC2 image's Single Root I/O Virtualization support
+    pub(crate) sriov_net_support: String,
+}
+
+impl From<Image> for ImageDef {
+    fn from(image: Image) -> Self {
+        Self {
+            image_id: image.image_id().unwrap_or_default().to_string(),
+            public: image.public().unwrap_or_default(),
+            ena_support: image.ena_support().unwrap_or_default(),
+            sriov_net_support: image.sriov_net_support().unwrap_or_default().to_string(),
+        }
+    }
+}
+
+impl ImageDef {
+    // Creates a new ImageDef with a given image_id and expected values for public, ena_support,
+    // and sriov_net_support
+    pub(crate) fn expected(image_id: String) -> Self {
+        Self {
+            image_id,
+            public: true,
+            ena_support: true,
+            sriov_net_support: "simple".to_string(),
+        }
+    }
+}
+
+pub(crate) async fn describe_images<'a>(
+    clients: &'a HashMap<Region, Ec2Client>,
+    image_ids: &HashMap<Region, Vec<String>>,
+) -> HashMap<&'a Region, Result<HashMap<String, ImageDef>>> {
+    // Build requests for images; we have to request with a regional client so we split them by
+    // region
+    let mut requests = Vec::with_capacity(clients.len());
+    for region in clients.keys() {
+        trace!("Requesting images in {}", region);
+        let ec2_client: &Ec2Client = &clients[region];
+        let get_future = describe_images_in_region(
+            region,
+            ec2_client,
+            image_ids
+                .get(region)
+                .map(|i| i.to_owned())
+                .unwrap_or(vec![]),
+        );
+
+        requests.push(join(ready(region), get_future));
+    }
+
+    // Send requests in parallel and wait for responses, collecting results into a list.
+    requests
+        .into_iter()
+        .collect::<FuturesUnordered<_>>()
+        .collect()
+        .await
+}
+
+/// Fetches all images in a single region
+pub(crate) async fn describe_images_in_region(
+    region: &Region,
+    client: &Ec2Client,
+    image_ids: Vec<String>,
+) -> Result<HashMap<String, ImageDef>> {
+    info!("Retrieving images in {}", region.to_string());
+    let mut images = HashMap::new();
+
+    // Send the request
+    let mut get_future = client
+        .describe_images()
+        .include_deprecated(true)
+        .set_image_ids(Some(image_ids))
+        .into_paginator()
+        .send();
+
+    // Iterate over the retrieved images
+    while let Some(page) = get_future.next().await {
+        let retrieved_images = page
+            .context(error::DescribeImagesSnafu {
+                region: region.to_string(),
+            })?
+            .images()
+            .unwrap_or_default()
+            .to_owned();
+        for image in retrieved_images {
+            // Insert a new key-value pair into the map, with the key containing image id
+            // and the value containing the ImageDef object created from the image
+            images.insert(
+                image
+                    .image_id()
+                    .ok_or(error::Error::MissingField {
+                        missing: "image_id".to_string(),
+                    })?
+                    .to_string(),
+                ImageDef::from(image.to_owned()),
+            );
+        }
+    }
+
+    info!("Images in {} have been retrieved", region.to_string());
+    Ok(images)
+}
+
+pub(crate) mod error {
+    use aws_sdk_ec2::error::DescribeImagesError;
+    use aws_sdk_ssm::types::SdkError;
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    #[allow(clippy::large_enum_variant)]
+    pub enum Error {
+        #[snafu(display("Failed to describe images in {}: {}", region, source))]
+        DescribeImages {
+            region: String,
+            source: SdkError<DescribeImagesError>,
+        },
+
+        #[snafu(display("Missing field in image: {}", missing))]
+        MissingField { missing: String },
+    }
+}
+
+pub(crate) type Result<T> = std::result::Result<T, error::Error>;
--- a/tools/pubsys/src/aws/validate_ami/mod.rs
+++ b/tools/pubsys/src/aws/validate_ami/mod.rs
@@ -1,0 +1,550 @@
+//! The validate_ami module owns the 'validate-ami' subcommand and controls the process of validating
+//! EC2 images
+
+pub(crate) mod ami;
+pub(crate) mod results;
+
+use self::ami::ImageDef;
+use self::results::{AmiValidationResult, AmiValidationResultStatus, AmiValidationResults};
+use crate::aws::client::build_client_config;
+use crate::aws::validate_ami::ami::describe_images;
+use crate::Args;
+use aws_sdk_ec2::{Client as AmiClient, Region};
+use log::{info, trace};
+use pubsys_config::InfraConfig;
+use serde::Deserialize;
+use snafu::ResultExt;
+use std::collections::{HashMap, HashSet};
+use std::fs::File;
+use std::path::PathBuf;
+use structopt::{clap, StructOpt};
+
+/// Validates EC2 images
+#[derive(Debug, StructOpt)]
+#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+pub(crate) struct ValidateAmiArgs {
+    /// File holding the validation configuration
+    #[structopt(long, parse(from_os_str))]
+    validation_config_path: PathBuf,
+
+    /// Optional path where the validation results should be written
+    #[structopt(long, parse(from_os_str))]
+    write_results_path: Option<PathBuf>,
+
+    #[structopt(long, requires = "write-results-path")]
+    /// Optional filter to only write validation results with these statuses to the above path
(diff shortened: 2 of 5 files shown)


tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+    /// If this argument is given, print the validation results summary as a JSON object instead
+    /// of a plaintext table
+    json: bool,
+}
+
+/// Structure of the validation configuration file
+#[derive(Debug, Deserialize)]
+pub(crate) struct ValidationConfig {
+    /// Vec of paths to JSON files containing expected metadata (image ids and SSM parameters)
+    expected_metadata_lists: Vec<PathBuf>,
```
webern, line 50: Slightly confused. These are paths to additional configuration files? If so, be specific about whether absolute paths are required. Or, if relative paths are allowed, what are they relative to? i.e. would paths relative to this aggregating config file work?

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,549 @@
...
+    #[structopt(long)]
+    /// If this argument is given, print the validation results summary as a JSON object instead
+    /// of a plaintext table
+    json: bool,
+}
+
+/// Structure of the validation configuration file
+#[derive(Debug, Deserialize)]
+pub(crate) struct ValidationConfig {
+    /// Vec of paths to JSON files containing expected metadata (image ids and SSM parameters)
```
webern, line 48: Is this comment a paste error? Are `metadata`, `image_ids` and `parameters` the right nouns here?

tools/pubsys/src/aws/validate_ami/ami.rs
```diff
@@ -0,0 +1,148 @@
...
+pub(crate) async fn describe_images_in_region(
+    region: &Region,
+    client: &Ec2Client,
+    image_ids: Vec<String>,
+) -> Result<HashMap<String, ImageDef>> {
+    info!("Retrieving images in {}", region.to_string());
+    let mut images = HashMap::new();
+
+    // Send the request
+    let mut get_future = client
```
webern, line 93: This is applicable to the implementation in your previous PR as well. I can't quite tell from reading this if things are happening truly in parallel? Can you tell from your tests how the performance will scale to 30 regions with 100s of AMIs in each?

I think the key question is whether the `send()` function is returning a *unordered* collection of futures.

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,714 @@
...
+use serde::{Deserialize, Serialize};
+use std::collections::{HashMap, HashSet};
+use std::fmt::{self, Display};
+use std::str::FromStr;
+use tabled::{Table, Tabled};
+
+/// Represent the possible status of an EC2 image validation
+#[derive(Debug, Eq, Hash, PartialEq, Serialize, Deserialize)]
+pub(crate) enum AmiValidationResultStatus {
+    /// The expected value was equal to the actual value
```
webern: In the case of AMIs can we be more specific than `value`? In this case do we actually mean:

```rust
/// The AMI was found and is public
```

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,714 @@
...
+}
+
+/// Represents a single EC2 image validation result
+#[derive(Debug, Eq, Hash, PartialEq, Serialize)]
+pub(crate) struct AmiValidationResult {
+    /// The id of the image
+    pub(crate) image_id: String,
+
+    /// The expected value of the image
+    pub(crate) expected_value: ImageDef,
```
webern: This naming seems to be a holdover from SSM? Should this field be called `expected_image_def`? Should the next field be called `actual_image_def`?

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,714 @@
...
+    pub(crate) fn new(
+        image_id: String,
+        expected_value: ImageDef,
+        actual_value: Option<ImageDef>,
+        region: Region,
+    ) -> Self {
+        // Determine the validation status based on equality, presence, and absence of expected and
+        // actual image values
+        let status = match (&expected_value, &actual_value) {
+            (expected_value, Some(actual_value)) if actual_value.eq(expected_value) => {
```
webern: slightly more idiomatic to use the operator

```suggestion
            (expected_value, Some(actual_value)) if actual_value. == expected_value => {
```

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,714 @@
...
+
+impl AmiValidationRegionSummary {
+    fn no_valid_results() -> Self {
+        // When the images in a region couldn't be retrieved, use `-1` to indicate this in the output table
+        // and set `accessible` to `false`
+        AmiValidationRegionSummary {
+            correct: -1,
+            incorrect: -1,
+            missing: -1,
+            accessible: false,
```
webern, line 139: It seems like this boolean might be duplicating the information that -1, -1, -1 represents. Not a big deal though.

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,714 @@
...
+                Region::new("us-west-2"),
+                Ok(HashSet::from([
+                    AmiValidationResult::new(
+                        "test3-image-id".to_string(),
+                        ImageDef::expected("test3-image-id".to_string()),
+                        Some(ImageDef {
+                            image_id: "test3-image-id".to_string(),
+                            public: true,
+                            ena_support: false,
+                            sriov_net_support: "simple".to_string(),
```
webern, line 264: Do we actually want to be including these fields in our configuration files and comparisons? @cbgbt @rpkelly @yeazelm Should we be simply validating that an AMI_ID is public, or should we be including these extra validations (I don't know what ena_support or sriov_net_support are, and I suspect we just get whatever is default for these)

tools/pubsys/src/aws/validate_ami/mod.rs
mjsterckx: This is correct, because the input files are the same as for the SSM parameter validation. The parameters don't necessarily have to be there, but the file should follow the structure of region->image_id->{}.

tools/pubsys/src/aws/validate_ami/results.rs
mjsterckx: This was @cbgbt 's suggestion, because we know for all Bottlerocket AMIs that ena_support should be true and sriov_net_support should be simple.

tools/pubsys/src/aws/validate_ami/results.rs
mjsterckx: It does, but I wanted an extra column that clearly indicates there was a problem with the region itself instead of just vague `-1` cells.

tools/pubsys/src/aws/validate_ami/ami.rs
mjsterckx: The requests are sent completely in parallel. These image requests are fast so the difference isn't really noticeable. For the SSM parameters, before they were parallelized, the total duration was over 3 minutes. After parallelizing, it took 1 minute (which is how long it takes to get the parameters from the region with the highest latency, even if that region is the only one to be queried).

If there were 30 regions with 100s of AMIs, the runtime would be the same as for a single region with the highest latency of those 30.

tools/pubsys/src/aws/validate_ami/mod.rs
mjsterckx: They can be relative paths, but they are relative to the pwd of the caller. I'll add that to the comments.

tools/pubsys/src/aws/validate_ami/results.rs
yeazelm: These are unlikely to be wrong, but a big deal if they are. So its worth ensuring the state of the AMI is not only that it exists, but has the right fields set. I'm a fan of including them if it doesn't affect complexity too much.

mjsterckx:
^ Rebased after https://github.com/bottlerocket-os/bottlerocket/pull/2969 got merged.

tools/pubsys/src/aws/validate_ami/results.rs
webern: Ok, sounds good he knows more than me.

mjsterckx:
Closing this PR to address @bcressey's comments on https://github.com/bottlerocket-os/bottlerocket/pull/2987. Will create a new PR with the change to how the subcommand takes its input.

PR #2986, opened 2023-04-06
kubelet: Enable setting cpuManagerPolicyOptions

### LINKED ISSUE #2342: Support CPUManager policy options
**What I'd like:**

I'm able to set the `static` CPU Manager policy with `settings.kubernetes.cpu-manager-policy`, but I can't set any options for it. I'd like to be able to set the `full-pcpus-only` policy option.

Not aware of any.

stmcginnis:
**Issue number:**

Closes #2342

**Description of changes:**

This adds a new `settings.kubernetes.cpu-manager-policy-options` setting to allow configuring the kubelet cpuManagerPolicyOptions. When the CPU manager policy is set to "static", these options allow affecting the behavior of CPU allocation for pods.

The only supported policy option at this point is [`full-pcpus-only` as it is the only one that is not hidden by an alpha feature gate](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy-options). More options can be supported later as they leave alpha.

This change could have been included with #2930, but that PR is already getting quite large. So at this point it will probably be better dealing with merge conflicts than inflating that further. If one or the other doesn't make it in time for the next release they can also be divided up and updated to have the correct migration for their ultimate release version.

**Testing done:**

Deployed built image and applied these settings:

```toml
[settings.kubernetes]
cpu-manager-policy = "static"
cpu-manager-policy-options = [
  "full-pcpus-only"
]
```

Verified the `/etc/kubernetes/kubelet/config` file rendered as expected:

```yaml
cpuManagerPolicy: static
cpuManagerPolicyOptions:
  full-pcpus-only: "true"
```

Checked `systemctl status kubelet` and made sure the service was healthy and reviewed logs to make sure `kubelet` was able to start up and recognize the new configuration.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,6 +418,18 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
 * `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`. If you want to allow pods with certain resource characteristics to be granted increased CPU affinity and exclusivity on the node, you can set this setting to `static`. You should reboot if you change this setting after startup - try `apiclient reboot`.
+* `settings.kubernetes.cpu-manager-policy-options`: Policy options to apply when `cpu-manager-policy` is set to `static`. Currently `full-pcpus-only` is the only option.
+
+  For example:
+
+  ```toml
+  [settings.kubernetes]
+  cpu-manager-policy = "static"
+  cpu-manager-policy-options = [
+    "full-pcpus-only"
+  ]
+  ```
+
 * `settings.kubernetes.cpu-manager-reconcile-period`: Specifies the CPU manager reconcile period, which controls how often updated CPU assignments are written to cgroupfs. The value is a duration like `30s` for 30 seconds or `1h5m` for 1 hour and 5 minutes.
 * `settings.kubernetes.credential-providers`: Contains a collection of Kubelet image credential provider settings.
   Each name under `credential-providers` is the name of the plugin to configure.
--- a/packages/kubernetes-1.24/kubelet-config
+++ b/packages/kubernetes-1.24/kubelet-config
@@ -80,6 +80,12 @@ cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
 {{#if settings.kubernetes.cpu-manager-reconcile-period}}
 cpuManagerReconcilePeriod: {{settings.kubernetes.cpu-manager-reconcile-period}}
 {{/if}}
+{{#if settings.kubernetes.cpu-manager-policy-options}}
+cpuManagerPolicyOptions:
+{{#each settings.kubernetes.cpu-manager-policy-options}}
+    {{this}}: "true"
+{{/each}}
+{{/if}}
 {{#if settings.kubernetes.topology-manager-scope}}
 topologyManagerScope: {{settings.kubernetes.topology-manager-scope}}
 {{/if}}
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -418,6 +418,7 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.container-log-max-files`: The maximum number of container log files that can be present for a container.
 * `settings.kubernetes.container-log-max-size`: The maximum size of container log file before it is rotated.
 * `settings.kubernetes.cpu-manager-policy`: Specifies the CPU manager policy. Possible values are `static` and `none`. Defaults to `none`. If you want to allow pods with certain resource characteristics to be granted increased CPU affinity and exclusivity on the node, you can set this setting to `static`. You should reboot if you change this setting after startup - try `apiclient reboot`.
+* `settings.kubernetes.cpu-manager-policy-options`: Policy options to apply when `cpu-manager-policy` is set to `static`. Currently `full-pcpus-only` is the only option.
```
bcressey, line 421: Would it make sense to represent this as a list?

```
[settings.kubernetes]
cpu-manager-policy = "static"
cpu-manager-policy-options = [
  "full-pcpus-only",
  "distribute-cpus-across-numa",
  "align-by-socket",
]
```

Versus:
```
[settings.kubernetes]
cpu-manager-policy = "static"

[settings.kubernetes.cpu-manager-policy-options]
full-pcpus-only = true
distribute-cpus-across-numa = true
align-by-socket = true
```

I suppose the drawback is that if we ever needed to add a non-feature-flag option, we'd have to support both with complex serialization logic.

In either case, it'd be good to add an example since this documentation is ambiguous on what the shape should be.

sources/api/migration/migrations/v1.14.0/kubelet-config-settings/src/main.rs
```diff
@@ -0,0 +1,19 @@
+use migration_helpers::{common_migrations::AddSettingsMigration, migrate, Result};
+use std::process;
+
+/// Additional `settings.kubernetes` options for this release.
+fn run() -> Result<()> {
+    migrate(AddSettingsMigration(&[
+        "settings.kubernetes.cpu-manager-policy-option",
```
bcressey:
```suggestion
        "settings.kubernetes.cpu-manager-policy-options",
```

packages/kubernetes-1.24/kubelet-config
```diff
@@ -80,6 +80,12 @@ cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
 {{#if settings.kubernetes.cpu-manager-reconcile-period}}
 cpuManagerReconcilePeriod: {{settings.kubernetes.cpu-manager-reconcile-period}}
 {{/if}}
+{{#if settings.kubernetes.cpu-manager-policy-option}}
+cpuManagerPolicyOptions:
+{{#each settings.kubernetes.cpu-manager-policy-option}}
+    {{@key}}: "{{this}}"
+{{/each}}
+{{/if}}
```
bcressey:
```suggestion
{{#if settings.kubernetes.cpu-manager-policy-options}}
cpuManagerPolicyOptions:
{{#each settings.kubernetes.cpu-manager-policy-options}}
    {{@key}}: "{{this}}"
{{/each}}
{{/if}}
```

README.md
stmcginnis: Great suggestion - that's actually much better since it allows doing something like:

```
apiclient apply <<EOF
  [settings.kubernetes]
  cpu-manager-policy = "static"
  cpu-manager-policy-options = []
EOF
```

to be able to remove any of these flags, rather than needing to have them included but disabled.

Expanded readme to include this as an example.

bcressey (approved):

stmcginnis:
Rebased to resolve merge conflict with `Release.toml` 1.13.3 updates

jpmcb (approved):

gthao313 (approved):

PR #2987, opened 2023-04-06
pubsys: write SSM parameters and AMI permissions to output files

mjsterckx:
**Description of changes:**

Modified the `pubsys ssm` and `pubsys promote-ssm` commands to allow the parameters to be written to a file. Modified the `pubsys ami` and `pubsys publish-ami` commands to have them write AMI launch permissions and publicity info to the output file.

- `pubsys ssm` has one new argument: `ssm_parameter_output`. If set, the rendered parameters will be written to the file at this path with the following structure:
```
{
  "<region-name>": {
    "<parameter-name>": "<parameter-value>",
    ...
  }
}
```

- `pubsys promote-ssm` has one new argument: `ssm_parameter_output`. The newly promoted parameters will be combined with the parameters already present in the file and written to it.

- `pubsys ami` now writes `public` and `launch_permissions` to the output file along with `id` and `name`. The `public` field is set to `false` by default when registering the AMI, but if the AMI ID has already been registered before, then the field is set to the publicity of that AMI. The `launch_permissions` field gets an empty vec by default, but if the AMI is already registered, then the launch permissions for that AMI are retrieved and stored in that field.

- `pubsys publish-ami` updates the `public` and `launch_permissions` fields in the input file if anything has changed after the publishing has finished.

**Testing done:**

- Unit tests
- Ran `cargo make`, `cargo make ami`, `cargo make ssm` with the new argument, and got a file like this:
```
{
  "us-west-2": {
    "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_id": "ami-012345678",
    "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_version": "1.14.0-f782c96d-dirty"
  }
}
```
The AMI file looked like this:
```
{
    "us-west-2": {
        "id": "ami-012345678",
        "name": "bottlerocket-aws-k8s-1.24-x86_64-v1.14.0-f782c96d-dirty",
        "public": false,
        "launch_permissions": [ ]
    }
}
```

- Ran `cargo make promote-ssm` with the above file as input and `keep_original_parameters` false, and got a file like this:
```
{
  "us-west-2": {
    "/msterckx/aws-k8s-1.24/x86_64/1.14.0/image_version": "1.14.0-f782c96d-dirty",
    "/msterckx/aws-k8s-1.24/x86_64/1.14.0/image_id": "ami-012345678",
    "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_id": "ami-012345678",
    "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_version": "1.14.0-f782c96d-dirty"
  }
}
```

- Ran `cargo make ami-public` and the `amis.json` file looked like this afterwards:
```
{
  "us-west-2": {
    "id": "ami-012345678",
    "name": "bottlerocket-aws-k8s-1.24-x86_64-v1.14.0-f782c96d-dirty",
    "public": true,
    "launch_permissions": [
      {
        "group": "all",
        "user_id": null,
        "organization_arn": null,
        "organizational_unit_arn": null
      }
    ]
  }
}
```

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/ami/launch_permissions.rs
+++ b/tools/pubsys/src/aws/ami/launch_permissions.rs
@@ -1,0 +1,98 @@
+use aws_sdk_ec2::{model::LaunchPermission, Client as Ec2Client};
+use serde::{Deserialize, Serialize};
+use snafu::ResultExt;
+
+/// Returns the launch permissions for the given AMI
+pub(crate) async fn get_launch_permissions(
+    ec2_client: &Ec2Client,
+    region: &str,
+    ami_id: &str,
+) -> Result<Vec<LaunchPermissionDef>> {
+    let ec2_response = ec2_client
+        .describe_image_attribute()
+        .image_id(ami_id)
+        .attribute(aws_sdk_ec2::model::ImageAttributeName::LaunchPermission)
+        .send()
+        .await
+        .context(error::DescribeImageAttributeSnafu {
+            ami_id,
+            region: region.to_string(),
+        })?;
+
+    let mut launch_permissions = vec![];
+
+    let responses: Vec<LaunchPermission> =
+        ec2_response.launch_permissions().unwrap_or(&[]).to_vec();
+    for permission in responses {
+        launch_permissions.push(LaunchPermissionDef::try_from(permission)?)
+    }
+    Ok(launch_permissions)
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq, Hash)]
+#[serde(rename_all = "lowercase")]
+pub(crate) enum LaunchPermissionDef {
+    /// The name of the group
+    Group(String),
+
+    /// The Amazon Web Services account ID
+    UserId(String),
+
+    /// The ARN of an organization
+    OrganizationArn(String),
+
+    /// The ARN of an organizational unit
+    OrganizationalUnitArn(String),
+}
+
+impl TryFrom<LaunchPermission> for LaunchPermissionDef {
+    type Error = crate::aws::ami::launch_permissions::Error;
+
+    fn try_from(launch_permission: LaunchPermission) -> std::result::Result<Self, Self::Error> {
+        let LaunchPermission {
+            group,
+            user_id,
+            organization_arn,
+            organizational_unit_arn,
+            ..
+        } = launch_permission.clone();
+        match (group, user_id, organization_arn, organizational_unit_arn) {
+            (Some(group), None, None, None) => {
+                Ok(LaunchPermissionDef::Group(group.as_str().to_string()))
+            }
+            (None, Some(user_id), None, None) => Ok(LaunchPermissionDef::UserId(user_id)),
+            (None, None, Some(organization_arn), None) => {
+                Ok(LaunchPermissionDef::OrganizationArn(organization_arn))
+            }
+            (None, None, None, Some(organizational_unit_arn)) => Ok(
+                LaunchPermissionDef::OrganizationalUnitArn(organizational_unit_arn),
+            ),
+            _ => Err(Error::InvalidLaunchPermission { launch_permission }),
+        }
+    }
+}
+
+mod error {
+    use aws_sdk_ec2::error::DescribeImageAttributeError;
+    use aws_sdk_ec2::model::LaunchPermission;
+    use aws_sdk_ec2::types::SdkError;
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub(crate) enum Error {
+        #[snafu(display("Error describing AMI {} in {}: {}", ami_id, region, source))]
+        DescribeImageAttribute {
+            ami_id: String,
+            region: String,
+            #[snafu(source(from(SdkError<DescribeImageAttributeError>, Box::new)))]
+            source: Box<SdkError<DescribeImageAttributeError>>,
+        },
+
+        #[snafu(display("Invalid launch permission: {:?}", launch_permission))]
+        InvalidLaunchPermission { launch_permission: LaunchPermission },
+    }
+}
+pub(crate) use error::Error;
+
+type Result<T> = std::result::Result<T, error::Error>;
--- a/tools/pubsys/src/aws/ami/mod.rs
+++ b/tools/pubsys/src/aws/ami/mod.rs
@@ -1,11 +1,14 @@
 //! The ami module owns the 'ami' subcommand and controls the process of registering and copying
 //! EC2 AMIs.
 
+pub(crate) mod launch_permissions;
 pub(crate) mod public;
 mod register;
 mod snapshot;
 pub(crate) mod wait;
 
+use crate::aws::ami::launch_permissions::get_launch_permissions;
+use crate::aws::ami::public::ami_is_public;
 use crate::aws::publish_ami::{get_snapshots, modify_image, modify_snapshots, ModifyOptions};
 use crate::aws::{client::build_client_config, parse_arch, region_from_string};
 use crate::Args;
@@ -26,7 +29,6 @@ use register::{get_ami_id, register_image, RegisteredIds};
 use serde::{Deserialize, Serialize};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::{HashMap, HashSet};
-use std::fs::File;
 use std::path::PathBuf;
 use structopt::{clap, StructOpt};
 use wait::wait_for_ami;
@@ -78,10 +80,7 @@ pub(crate) async fn run(args: &Args, ami_args: &AmiArgs) -> Result<()> {
         Ok(amis) => {
             // Write the AMI IDs to file if requested
             if let Some(ref path) = ami_args.ami_output {
-                let file = File::create(path).context(error::FileCreateSnafu { path })?;
-                serde_json::to_writer_pretty(file, &amis)
-                    .context(error::SerializeSnafu { path })?;
-                info!("Wrote AMI data to {}", path.display());
+                write_amis(path, &amis).context(error::WriteAmisSnafu { path })?;
             }
             Ok(())
         }
@@ -141,6 +140,10 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
         region: base_region.as_ref(),
     })?;
 
+    // If the AMI does not exist yet, `public` should be false and `launch_permissions` empty
+    let mut public = false;
+    let mut launch_permissions = vec![];
+
     let (ids_of_image, already_registered) = if let Some(found_id) = maybe_id {
         warn!(
             "Found '{}' already registered in {}: {}",
@@ -153,9 +156,25 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
                 region: base_region.as_ref(),
             })?;
         let found_ids = RegisteredIds {
-            image_id: found_id,
+            image_id: found_id.clone(),
             snapshot_ids,
         };
+
+        public = ami_is_public(&base_ec2_client, base_region.as_ref(), &found_id)
+            .await
+            .context(error::IsAmiPublicSnafu {
+                image_id: found_id.clone(),
+                region: base_region.to_string(),
+            })?;
+
+        launch_permissions =
+            get_launch_permissions(&base_ec2_client, base_region.as_ref(), &found_id)
+                .await
+                .context(error::DescribeImageAttributeSnafu {
+                    image_id: found_id,
+                    region: base_region.to_string(),
+                })?;
+
         (found_ids, true)
     } else {
         let new_ids = register_image(ami_args, &base_region, base_ebs_client, &base_ec2_client)
@@ -174,7 +193,12 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
 
     amis.insert(
         base_region.as_ref().to_string(),
-        Image::new(&ids_of_image.image_id, &ami_args.name),
+        Image::new(
+            &ids_of_image.image_id,
+            &ami_args.name,
+            Some(public),
+            Some(launch_permissions),
+        ),
     );
 
     // If we don't need to copy AMIs, we're done.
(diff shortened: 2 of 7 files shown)


bcressey:
```
{
  "us-west-2": {
    "ami-012345678": {
      "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_id": "ami-012345678",
      "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_version": "1.14.0-f782c96d-dirty"
    }
  }
}
```

Why not always write the file, the way that `pubsys ami` always writes `amis.json`? That would have these benefits:
* `cargo make ssm` and `cargo make promote-ssm` would just work and keep `ssm-parameters.json` up to date
* the input / output / keep original distinctions could go away, since the input and output would be the same
* overall the behavior would be more like `pubsys ami` which reduces developer surprise

I am also not convinced that everything should go under an AMI key like `ami-012345678`. The `image_version` field is unrelated to the AMI, and the `image_id` field duplicates that information.

A flattened map like this seems more intuitive:
```
{
  "us-west-2": {
    "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_id": "ami-012345678",
    "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_version": "1.14.0-f782c96d-dirty",
    "/msterckx/aws-k8s-1.24/x86_64/latest/image_id": "ami-012345678",
    "/msterckx/aws-k8s-1.24/x86_64/latest/image_version": "1.14.0-f782c96d-dirty"
  }
}
```

mjsterckx:
The goal of this code was to generate files that were easily consumed by the SSM validation and EC2 validation code (https://github.com/bottlerocket-os/bottlerocket/pull/2969 and https://github.com/bottlerocket-os/bottlerocket/pull/2983). I should've probably made that more clear in the description.

Doing it this way allows us to create a file for version 1.14.0. and a separate file for latest, which is what we use for the validation. Having the AMI id as a key lets the EC2 validation use the same files to check all of the images in there, without having to parse each parameter to see if it contains an image id.

bcressey:
> The goal of this code was to generate files that were easily consumed by the SSM validation and EC2 validation code (https://github.com/bottlerocket-os/bottlerocket/pull/2969 and https://github.com/bottlerocket-os/bottlerocket/pull/2983).

I don't think it makes sense for the SSM validation code to be organizing parameters under an AMI ID - the AMI ID is irrelevant there. All the SSM validation should care about is whether the current parameter values match the expected values.

For EC2 validation, it kind of does seem like the validator logic should be going through and finding the subset of parameters that refer to an AMI ID. But if you wanted something more robust than a suffix match, you could organize it slightly differently:

```
{
  "us-west-2": {
    "image_ids": {
      "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_id": "ami-012345678",
      "/msterckx/aws-k8s-1.24/x86_64/latest/image_id": "ami-012345678",
    },
    "image_versions": {
      "/msterckx/aws-k8s-1.24/x86_64/1.14.0-f782c96d-dirty/image_version": "1.14.0-f782c96d-dirty",
      "/msterckx/aws-k8s-1.24/x86_64/latest/image_version": "1.14.0-f782c96d-dirty"
    }
  }
}
```

> Doing it this way allows us to create a file for version 1.14.0. and a separate file for latest, which is what we use for the validation.

For pubsys / makefile integration, having the ability to easily work with all published artifacts for a specific build ID (like `1.14.0-f782c96d-dirty`) is valuable. If I make AMIs and publish SSM parameters, I should be able to run `cargo make validate-ssm` and have it validate everything. If I subsequently promote SSM parameters, I'd expect `cargo make validate-ssm` to validate all the SSM parameters I've promoted. I would not expect to need to pass different arguments or point these operations to different JSON files; this should all just be maintained for me in the JSON file that resides in build artifacts.

For other use cases, like an account-wide `ssm-params.json` covering all SSM parameters ever published, a single unified file also seems to make the most sense. It'd be pretty easy to generate that file by concatenating all the files together, overwriting fields that are also present in later files (like "latest" params). And then the code path for doing the validation would be the same as in the local developer case.

mjsterckx:
^ Addressed @bcressey's suggestions and added the writing of AMI publicity info and launch permissions to the `pubsys ami` and `pubsys publish-ami` commands.

cbgbt:
It would be a lot easier to review this if it were split into two commits, one for each feature. Would you mind resubmitting it?

mjsterckx:
^ Split the one commit into two.

tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -372,20 +418,63 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
     Ok(amis)
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq, Hash)]
+pub(crate) struct LaunchPermissionDef {
```
cbgbt: This seems like it should be an enum, something like:

```rust
#[derive(Serialize, Deserialize, etc)]
#[serde(rename_all = "lowercase")]
enum LaunchPermissionDef {
    Group(String),
    UserId(String),
    OrganizationArn(String),
    OrganizationalUnitArn(String),
}
```

I'm pretty sure this still [serializes to JSON the way you want](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8ac38bfb19f7e379aec294d3746e54b6).

tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -372,20 +418,63 @@ async fn _run(args: &Args, ami_args: &AmiArgs) -> Result<HashMap<String, Image>>
...
+}
+
+impl From<LaunchPermission> for LaunchPermissionDef {
+    fn from(launch_permission: LaunchPermission) -> Self {
+        Self {
+            group: launch_permission
+                .group()
+                .map(|group| group.as_str().to_owned()),
+            user_id: launch_permission
+                .user_id()
```
cbgbt: Supposing you do make this an enum, you could rewrite this pretty succinctly with a match:

```rust
let LaunchPermission { group, user_id, organization_arn, organizational_unit_arn } = launch_permission;
    match (group, user_id, organization_arn, organizational_unit_arn) {
        (Some(group), None, None, None) => LaunchPermissionDef::Group(group),
        (None, Some(user_id), None, None) => LaunchPermissionDef::UserId(user_id),
        (None, None, Some(organization_arn), None) => LaunchPermissionDef::OrganizationArn(organization_arn),
        (None, None, None, Some(organizational_unit_arn)) => LaunchPermissionDef::OrganizationalUnitArn(organizational_unit_arn),
        _ => // error!
    };
```


You would have to change this from `impl From` to `impl TryFrom` to handle the error case of all of the elements being `None` though.

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -200,18 +203,32 @@ pub(crate) async fn run(args: &Args, publish_args: &PublishArgs) -> Result<()> {
...
 
+    write_amis(&publish_args.ami_input, &amis)?;
+
+    Ok(())
+}
+
+fn write_amis(ami_output: &PathBuf, amis: &HashMap<Region, Image>) -> Result<()> {
+    let file = File::create(ami_output).context(error::FileSnafu {
+        op: "write amis to file",
+        path: ami_output,
```
cbgbt: This function should probably also be [used in `pubsys ami` command](https://github.com/bottlerocket-os/bottlerocket/pull/2987/commits/2254cb6fadc000b2915d634042c88c9e5f823101#diff-ce04b11594176c7886cc95d134e3d1100b817c0986274d5adde9889b4821c225R81-R89), right?

tools/pubsys/src/aws/promote_ssm/mod.rs
```diff
@@ -216,8 +231,86 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
     Ok(())
 }
 
+/// Read parameters in given file, add newly promoted parameters, and write combined parameters to
+/// the given file
+async fn write_rendered_parameters(
```
cbgbt: I think `append_rendered_parameters` may be more telling, since `write` often carries a connotation of `overwrite`

tools/pubsys/src/aws/promote_ssm/mod.rs
```diff
@@ -216,8 +231,86 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
...
+    ssm_parameters_output: &PathBuf,
+    set_parameters: &HashMap<SsmKey, String>,
+    source_target_map: HashMap<&String, &String>,
+) -> Result<()> {
+    info!(
+        "Writing promoted SSM parameters to {}",
+        ssm_parameters_output.display()
+    );
+    let parsed_parameters = parse_expected_parameters(&ssm_parameters_output.to_owned())
```
cbgbt: Should this be `parse_existing_parameters` or just `parse_parameters`? Seems like the name is a holdover from parsing it in the validation code, but we need the method more generically.

```suggestion
    let parsed_parameters = parse_parameter_file(&ssm_parameters_output.to_owned())
```

tools/pubsys/src/aws/promote_ssm/mod.rs
```diff
@@ -216,8 +231,86 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
...
+        .await
+        .context(error::ParseExistingSsmParametersSnafu {
+            path: ssm_parameters_output,
+        })?;
+
+    let combined_parameters: HashMap<Region, HashMap<SsmKey, String>> =
+        combine_parameters(parsed_parameters, set_parameters, source_target_map);
+
+    serde_json::to_writer_pretty(
```
cbgbt: Should this be a function somewhere? We write out this parameter data structure in more than one place now.

tools/pubsys/src/aws/promote_ssm/mod.rs
```diff
@@ -216,8 +231,86 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
...
+
+    info!(
+        "Wrote promoted SSM parameters to {}",
+        ssm_parameters_output.display()
+    );
+    Ok(())
+}
+
+/// Return a HashMap of Region mapped to a HashMap of SsmKey, String pairs, representing the newly
+/// promoted parameters as well as the original parameters
```
cbgbt: I think the behavior is correct, but it's worth calling out that any parameter collisions take the promoted parameter value.

That case probably deserves a test too.

tools/pubsys/src/aws/ssm/mod.rs
```diff
@@ -213,6 +223,31 @@ pub(crate) async fn run(args: &Args, ssm_args: &SsmArgs) -> Result<()> {
     Ok(())
 }
 
+/// Write rendered parameters to the file at `ssm_parameters_output`
+async fn write_rendered_parameters(
```
cbgbt: Oh, you do have one! Just need to share it.

tools/pubsys/src/aws/promote_ssm/mod.rs
```diff
@@ -216,8 +231,86 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
...
+) -> Result<()> {
+    info!(
+        "Writing promoted SSM parameters to {}",
+        ssm_parameters_output.display()
+    );
+    let parsed_parameters = parse_expected_parameters(&ssm_parameters_output.to_owned())
+        .await
+        .context(error::ParseExistingSsmParametersSnafu {
+            path: ssm_parameters_output,
+        })?;
```
cbgbt, line 254: Am I correct that this will fail if the file is empty or doesn't exist? Is it better to just write the new parameters in that case, rather than refusing to proceed?

webern (approved):

tools/pubsys/src/aws/promote_ssm/mod.rs
mjsterckx: That is correct, but I don't know if we just want to write the new parameters. One reason it could fail is that the syntax in the file is incorrect, in which case I think it's better to alert the user and halt the program than completely overwrite the content of the file.

tools/pubsys/src/aws/ami/launch_permissions.rs
```diff
@@ -0,0 +1,97 @@
...
+    let responses: Vec<LaunchPermission> =
+        ec2_response.launch_permissions().unwrap_or(&[]).to_vec();
+    for permission in responses {
+        launch_permissions.push(LaunchPermissionDef::try_from(permission)?)
+    }
+    Ok(launch_permissions)
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq, Hash)]
+pub(crate) enum LaunchPermissionDef {
```
cbgbt: I think it would be more coherent with our other serialization code if you lowercased these:

```suggestion
#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq, Hash)]
#[serde(rename_all = "lowercase")]
pub(crate) enum LaunchPermissionDef {
```

tools/pubsys/src/aws/ami/launch_permissions.rs
```diff
@@ -0,0 +1,97 @@
...
+    type Error = crate::aws::ami::launch_permissions::Error;
+
+    fn try_from(launch_permission: LaunchPermission) -> std::result::Result<Self, Self::Error> {
+        let LaunchPermission {
+            group,
+            user_id,
+            organization_arn,
+            organizational_unit_arn,
+            ..
+        } = launch_permission.clone();
```
cbgbt, line 58: I am surprised that you had to clone this.

tools/pubsys/src/aws/promote_ssm/mod.rs
cbgbt: We chatted offline and concluded that we could specially handle `Error::ReadExpectedParameterFile`.

cbgbt (commented):
Looks good to me other than the casing.

tools/pubsys/src/aws/ami/launch_permissions.rs
mjsterckx: Wouldn't it make more sense to use snake_case?

cbgbt: If I `git grep` for `rename_all` in Bottlerocket, it seems like we typically use `kebab-case` or `lowercase`

cbgbt (approved):

Issue #2989, opened 2023-04-07
1.13.2: aws-iam-authenticator now finding `/root/.aws/config`


nairb774:
**Image I'm using:**

Upgrading `Bottlerocket OS 1.13.1 (aws-k8s-1.22)` to `Bottlerocket OS 1.13.2 (aws-k8s-1.22)`

**What I expected to happen:**

A successful upgrade.

**What actually happened:**

Nodes failed to successfully authenticate with the cluster.

**How to reproduce the problem:**

I apologize that this might be long, and that the config/goals might feel a little odd.

Start nodes with the following configuration:

```toml
[settings.aws]
config = "W3Byb2ZpbGUgZGVmYXVsdF0KY3JlZGVudGlhbF9zb3VyY2UgPSBFYzJJbnN0YW5jZU1ldGFkYXRhCnJvbGVfYXJuID0gYXJuOmF3czppYW06OjEyMzQ1Njc4OTAxMjpyb2xlL015RXh0cmFSb2xlCg=="
profile = "default"
region = "us-east-1"
```

Where the `settings.aws.config` is something like:
```ini
[profile default]
credential_source = Ec2InstanceMetadata
role_arn = arn:aws:iam::123456789012:role/MyExtraRole
```

The role attached to the instance, and the additional role referenced above have all the correct permissions to allow the assume role to take place. Additionally, the instance role and the extra role have the same set of permissions attached to them (the required [AmazonEKSWorkerNodePolicy](https://us-east-1.console.aws.amazon.com/iam/home#/policies/arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy) and [AmazonEC2ContainerRegistryReadOnly](https://us-east-1.console.aws.amazon.com/iam/home#/policies/arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly)).

When the v1.13.1 instance starts up, the older v0.6.2 version of the `aws-iam-authenticator` would use the instance's role to generate the token needed to authenticate with the control plane. With the 1.13.2 version, which contains the v0.6.8 of the `aws-iam-authenticator`, it will discover the `/root/.aws/config` file and will make use of the default provider to create the token for authenticating the node with the control plane. This change in behavior was due to https://github.com/aws/aws-sdk-go/pull/4519 being included in the `aws-sdk-go` library upgrades. That PR adds a fallback to read the home directory information from `/etc/passwd` if the `HOME` environment variable is not set.

[...]

PR #2990, opened 2023-04-07
aws.profile: Add note to avoid default profile

stmcginnis:
**Issue number:**

Related: #2989

**Description of changes:**

Recent versions of `aws-iam-authenticator`, and perhaps other AWS components, not pick up the default credential settings when `settings.aws.profile` is set to `default`. To help avoid this situation, this adds a note to the settings documentation to avoid using `default` for AWS credentials.

**Testing done:**

👀 

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1118,6 +1118,8 @@ They can be overridden for testing purposes in [the same way as other settings](
   ```
 
   **Note**: If `settings.aws.profile` is not set, the setting will fallback to the "default" profile.
+  In general it is recommended not to include a `[profile default]` section in the `aws.config` contents though.
+  This may have unintended side effects for other AWS services running on the node (e.g. `aws-iam-authenticator`).
 
   **Note:** The `config`, `credentials`, and `profile` are optional and do not need to be set when using an Instance Profile when running on an AWS instance.
 


README.md
```diff
@@ -1118,6 +1118,8 @@ They can be overridden for testing purposes in [the same way as other settings](
   ```
 
   **Note**: If `settings.aws.profile` is not set, the setting will fallback to the "default" profile.
+  It is recommended to use something other than `default` for the `aws.config` profile name.
+  Using `default` may have unintended side effects for other AWS services running on the node, such as `aws-iam-authenticator`.
```
nairb774: In some ways, I'd include a note that having a `[profile default]` defined in the config might cause un-intended consequences. Imagine someone having 2 profiles defined, `default` and `kubelet`, in the config but had this set to `kubelet`, then they would still step on the issue but would still be following the advice.

In other words, I might adjust the wording to address the contents of `settings.aws.config` rather than the value of `settings.aws.profile`.

stmcginnis: Great point, thanks @nairb774! The issue really is more about having `[profile default]` in that file, rather than the `aws.profile` setting. Good call!

I'll get this updated to phrase the issue better shortly.

ecpullen (approved):

jpmcb (approved):

markusboehme (approved):

README.md
```diff
@@ -1118,6 +1118,8 @@ They can be overridden for testing purposes in [the same way as other settings](
   ```
 
   **Note**: If `settings.aws.profile` is not set, the setting will fallback to the "default" profile.
+  It is recommended not to use `[profile default]` in the `aws.config` contents.
```
etungsten: Maybe instead of "use", the verb "set"/"define" would be more accurate? Since this issue happens if you have [profile default] set in the blob even if you don't use it?

Also I find it weird the sentence right above says the profile setting defaults to "default" and in the next sentence we say not to use "default"

stmcginnis: Reworded slightly to hopefully make it more clear and easier to parse.

etungsten (approved):

PR #2992, opened 2023-04-08
Update Rust dependencies

### LINKED ISSUE #2995: v1.14.0 Rust dependency updates
Tracking issue for the v1.14.0 release to update Rust dependencies.

stmcginnis:
**Issue number:**

Closes: #2995

**Description of changes:**

Ran cargo update for `sources` and `tools`, avoided some updates that pull unnecessary extra versions. One change needed in `clarify.toml` due to [clean up of Apache 2.0 license content](https://github.com/dtolnay/unicode-ident/commit/600ff66352f47c929d6ce1c7f1a831d4362cffc0).

**Testing done:**

Deployed `aws-k8s-1.24` cluster. Ran `sonobuoy run --mode=quick`. No errors:

```
12:06:16             e2e                                       global   complete   passed   Passed:  1, Failed:  0, Remaining:  0
12:06:16    systemd-logs   ip-192-168-7-41.us-east-2.compute.internal   complete   passed
```

**Terms of contribution:**

### DIFF
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -45,16 +45,16 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.2.2"
+version = "3.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c83abf9903e1f0ad9973cc4f7b9767fd5a03a583f51a5b7a339e07987cd2724"
+checksum = "0070905b2c4a98d184c4e81025253cb192aa8a73827553f38e9410801ceb35bb"
 dependencies = [
  "actix-codec",
  "actix-rt",
  "actix-service",
  "actix-utils",
  "ahash",
- "base64",
+ "base64 0.21.0",
  "bitflags",
  "bytes",
  "bytestring",
@@ -74,6 +74,8 @@ dependencies = [
  "rand",
  "sha1",
  "smallvec",
+ "tokio",
+ "tokio-util",
  "tracing",
 ]
 
@@ -84,7 +86,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "465a6172cf69b960917811022d8f29bc0b7fa1398bc4f78b3c466673db1213b6"
 dependencies = [
  "quote",
- "syn",
+ "syn 1.0.109",
 ]
 
 [[package]]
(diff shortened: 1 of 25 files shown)


sources/Cargo.lock
```diff
@@ -421,39 +423,40 @@ checksum = "2777730b2039ac0f95f093556e61b6d26cebed5393ca6f152717777cec3a42ed"
...
-version = "0.1.64"
+version = "0.1.68"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1cd7fce9ba8c3c042128ce72d8b2ddbf3a05747efb67ea0313c635e10bda47a2"
+checksum = "b9ccdd8f2a161be9bd5c023df56f1b2a0bd1d83872ae53b71a84a12c9bf6e842"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.13",
```
jpculp, line 459: I think I would hold off on bumping `async-trait` and other things requiring syn 2 for now.

jpculp (commented):
Would it be possible to reduce the number of added crates, particularly in `sources`?

sources/Cargo.lock
stmcginnis: Is the concern having multiple versions of some dependencies?

If so, I would be very curious to see how much overhead that actually brings. It doesn't sit quite right with me to avoid taking bug fixes and improvements for some things just because there are other dependencies that haven't gotten around to upgrading yet.

jpculp: I've found that there is typically a happy medium where while the latest crate brings in extra dependencies, the release prior that might only be a week or two older does not. This can help keep the size down. Of course if the latest version brings in a specific feature we are waiting for, or a security fix, we should absolutely go ahead with it. Regarding `async-trait` and some related crates, I'm surprised to see `syn` 1.x→2.x pulled into patch releases. It is a couple weeks old at this point so we could take it for a spin if we really want.

jpculp (commented):
It looks like for `pluto` we'll have to do something like this:
```rust
let eks_config = aws_sdk_eks::config::Builder::from(&config).http_connector(http_client).build();
aws_sdk_eks::Client::from_conf(eks_config)
```

jpculp (approved):

yeazelm (approved):

PR #3020, opened 2023-04-14
pubsys: added ami validation

mjsterckx:
**Description of changes:**

Added a `validate-ami` subcommand to the `pubsys` tool. This command has the following signature:
```
pubsys-validate-ami 0.1.0
Validates EC2 images

USAGE:
    pubsys --infra-config-path <infra-config-path> validate-ami [FLAGS] [OPTIONS] --expected-amis-path <expected-amis-path>

FLAGS:
        --json       If this argument is given, print the validation results summary as a JSON object instead of a
                     plaintext table

OPTIONS:
        --expected-amis-path <expected-amis-path>           File holding the expected amis
        --write-results-path <write-results-path>           Optional path where the validation results should be written
            Optional filter to only write validation results with these statuses to the above path The available
            statuses are: `Correct`, `Incorrect`, `Missing`
        --log-level <log-level>
```

The function takes a path to a file containing an image object per region. An example file looks like this:
```
{
  "us-west-2": {
    "id": "ami-012345678",
    "name": "bottlerocket-aws-k8s-1.24-x86_64-v1.14.0-abcdef-dirty",
    "public": true,
    "launch_permissions": [
      {
        "group": "all"
      }
    ]
  }
}
```

The function will call the `describe-images` function on each image in the file to ensure that the retrieved id, name, and publicity match the expected values. If `public` is false, then the specific launch_permissions are retrieved using `describe-image-attribute` and included in the comparison. Aside from publicity and launch_permissions, the function will check that `sriov_net_support` is `simple` and `ena_support` is `true`.

The result of the function looks like this:
```
+-----------+---------+-----------+---------+------------+
| String    | correct | incorrect | missing | unreachable |
+-----------+---------+-----------+---------+------------+
| us-west-2 | 0       | 0         | 1       | 0          |
+-----------+---------+-----------+---------+------------+
```
- `correct` indicates that all retrieved values match the expected values
- `incorrect` indicates that at least one value does not match the expected value
- `missing` indicates that the specific image failed to be retrieved
- `unreachable` indicates that the images in this region could not be retrieved because the region could not be reached

`--write-results-path` and `--write-results-filter` allow the user to write the results of the ami validation to a JSON file. The filter determines which statuses will be written to this file. The file looks like this:

```
[
  {
    "id": "ami-012345678",
    "expected_image_def": {
      "id": "ami-012345678",
      "name": "bottlerocket-aws-k8s-1.24-x86_64-v1.14.0-abcdef-dirty",
      "public": true,
      "launch_permissions": null,
      "ena_support": true,
      "sriov_net_support": "simple"
    },
    "actual_image_def": null,
    "region": "us-west-2",
    "status": "Missing"
  }
]
```

Modified the SSM validation results to represent the new `Unreachable` status as well.

**Testing done:**

- Unit tests
- Ran the program on an unregistered ami (results seen above) and on registered public ones, which showed as `correct`.

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/validate_ami/ami.rs
+++ b/tools/pubsys/src/aws/validate_ami/ami.rs
@@ -1,0 +1,224 @@
+//! The ami module owns the describing of images in EC2.
+
+use aws_sdk_ec2::model::Image;
+use aws_sdk_ec2::{Client as Ec2Client, Region};
+use futures::future::{join, ready};
+use futures::stream::{FuturesUnordered, StreamExt};
+use log::{info, trace};
+use serde::{Deserialize, Serialize};
+use snafu::ResultExt;
+use std::collections::HashMap;
+
+use crate::aws::ami::launch_permissions::{get_launch_permissions, LaunchPermissionDef};
+
+/// Wrapper structure for the `ImageDef` struct, used during deserialization
+#[derive(Deserialize)]
+#[serde(untagged)]
+pub(crate) enum ImageData {
+    Image(ImageDef),
+    ImageList(Vec<ImageDef>),
+}
+
+impl ImageData {
+    pub(crate) fn images(&self) -> Vec<ImageDef> {
+        match self {
+            ImageData::Image(image) => vec![image.to_owned()],
+            ImageData::ImageList(images) => images.to_owned(),
+        }
+    }
+}
+
+/// Structure of the EC2 image fields that should be validated
+#[derive(Serialize, Deserialize, PartialEq, Eq, Hash, Debug, Clone)]
+pub(crate) struct ImageDef {
+    /// The ID of the EC2 image
+    pub(crate) id: String,
+
+    /// The name of the EC2 image
+    pub(crate) name: String,
+
+    /// Whether or not the EC2 image is public
+    #[serde(default)]
+    pub(crate) public: bool,
+
+    /// The launch permissions for the EC2 image.
+    pub(crate) launch_permissions: Option<Vec<LaunchPermissionDef>>,
+
+    /// Whether or not the EC2 image supports Elastic Network Adapter
+    #[serde(default = "default_ena_support")]
+    pub(crate) ena_support: bool,
+
+    /// The level of the EC2 image's Single Root I/O Virtualization support
+    #[serde(default = "default_sriov_net_support")]
+    pub(crate) sriov_net_support: String,
+}
+
+fn default_ena_support() -> bool {
+    true
+}
+
+fn default_sriov_net_support() -> String {
+    "simple".to_string()
+}
+
+impl From<(Image, Option<Vec<LaunchPermissionDef>>)> for ImageDef {
+    fn from(args: (Image, Option<Vec<LaunchPermissionDef>>)) -> Self {
+        Self {
+            id: args.0.image_id().unwrap_or_default().to_string(),
+            name: args.0.name().unwrap_or_default().to_string(),
+            public: args.0.public().unwrap_or_default(),
+            launch_permissions: args.1,
+            ena_support: args.0.ena_support().unwrap_or_default(),
+            sriov_net_support: args.0.sriov_net_support().unwrap_or_default().to_string(),
+        }
+    }
+}
+
+/// Fetches all images whose IDs are keys in `expected_images`. The map `expected_image_public` is
+/// used to determine if the launch permissions for the image should be fetched (only if the image is not
+/// public). The return value is a HashMap of Region to a Result, which is `Ok` if the request for
+/// that region was successful and `Err` if not. The Result contains a HashMap of `image_id` to
+/// `ImageDef`.
+pub(crate) async fn describe_images<'a>(
+    clients: &'a HashMap<Region, Ec2Client>,
+    expected_images: &HashMap<Region, Vec<ImageDef>>,
+) -> HashMap<&'a Region, Result<HashMap<String, ImageDef>>> {
+    // Build requests for images; we have to request with a regional client so we split them by
+    // region
+    let mut requests = Vec::with_capacity(clients.len());
+    clients.iter().for_each(|(region, ec2_client)| {
+        trace!("Requesting images in {}", region);
+        let get_future = describe_images_in_region(
+            region,
+            ec2_client,
+            expected_images
+                .get(region)
+                .map(|i| i.to_owned())
+                .unwrap_or_default()
+                .into_iter()
+                .map(|i| (i.id.clone(), i))
+                .collect::<HashMap<String, ImageDef>>(),
+        );
+
+        requests.push(join(ready(region), get_future));
+    });
+
+    // Send requests in parallel and wait for responses, collecting results into a list.
+    requests
+        .into_iter()
+        .collect::<FuturesUnordered<_>>()
+        .collect()
+        .await
+}
+
+/// Fetches the images whose IDs are keys in `expected_images`
+pub(crate) async fn describe_images_in_region(
+    region: &Region,
+    client: &Ec2Client,
+    expected_images: HashMap<String, ImageDef>,
+) -> Result<HashMap<String, ImageDef>> {
+    info!("Retrieving images in {}", region.to_string());
+    let mut images = HashMap::new();
+
+    // Send the request
+    let mut get_future = client
+        .describe_images()
+        .include_deprecated(true)
+        .set_image_ids(Some(Vec::from_iter(
+            expected_images.keys().map(|k| k.to_owned()),
+        )))
+        .into_paginator()
+        .send();
+
+    // Iterate over the retrieved images
+    while let Some(page) = get_future.next().await {
+        let retrieved_images = page
+            .context(error::DescribeImagesSnafu {
+                region: region.to_string(),
+            })?
+            .images()
+            .unwrap_or_default()
+            .to_owned();
+        for image in retrieved_images {
+            // Insert a new key-value pair into the map, with the key containing image ID
+            // and the value containing the ImageDef object created from the image
+            let image_id = image
+                .image_id()
+                .ok_or(error::Error::MissingField {
+                    missing: "image_id".to_string(),
+                })?
+                .to_string();
+            let expected_public = expected_images
+                .get(&image_id)
+                .ok_or(error::Error::MissingExpectedPublic {
+                    missing: image_id.clone(),
+                })?
+                .public;
+            // If the image is not expected to be public, retrieve the launch permissions
+            trace!(
+                "Retrieving launch permissions for {} in {}",
+                image_id,
+                region.as_ref()
+            );
+            let launch_permissions = if !expected_public {
+                Some(
+                    get_launch_permissions(client, region.as_ref(), &image_id)
+                        .await
+                        .context(error::GetLaunchPermissionsSnafu {
+                            region: region.as_ref(),
+                            image_id: image_id.clone(),
+                        })?,
+                )
+            } else {
+                None
+            };
+            let image_def = ImageDef::from((image.to_owned(), launch_permissions));
+            images.insert(image_id, image_def);
+        }
+    }
+
+    info!("Images in {} have been retrieved", region.to_string());
+    Ok(images)
+}
+
+pub(crate) mod error {
+    use aws_sdk_ec2::error::DescribeImagesError;
+    use aws_sdk_ssm::types::SdkError;
+    use aws_smithy_types::error::display::DisplayErrorContext;
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    #[allow(clippy::large_enum_variant)]
+    pub(crate) enum Error {
+        #[snafu(display(
+            "Failed to describe images in {}: {}",
+            region,
+            DisplayErrorContext(source)
+        ))]
+        DescribeImages {
+            region: String,
+            source: SdkError<DescribeImagesError>,
+        },
+
+        #[snafu(display(
+            "Failed to retrieve launch permissions for image {} in region {}: {}",
+            image_id,
+            region,
+            source
+        ))]
+        GetLaunchPermissions {
+            region: String,
+            image_id: String,
+            source: crate::aws::ami::launch_permissions::Error,
+        },
+
+        #[snafu(display("Missing field in image: {}", missing))]
+        MissingField { missing: String },
+
+        #[snafu(display("Missing image ID in expected image publicity map: {}", missing))]
+        MissingExpectedPublic { missing: String },
+    }
+}
+
+pub(crate) type Result<T> = std::result::Result<T, error::Error>;
--- a/tools/pubsys/src/aws/validate_ami/mod.rs
+++ b/tools/pubsys/src/aws/validate_ami/mod.rs
@@ -1,0 +1,851 @@
+//! The validate_ami module owns the 'validate-ami' subcommand and controls the process of validating
+//! EC2 images
+
+pub(crate) mod ami;
+pub(crate) mod results;
+
+use self::ami::{ImageData, ImageDef};
+use self::results::{AmiValidationResult, AmiValidationResultStatus, AmiValidationResults};
+use crate::aws::client::build_client_config;
+use crate::aws::validate_ami::ami::describe_images;
+use crate::Args;
+use aws_sdk_ec2::{Client as AmiClient, Region};
+use log::{error, info, trace};
+use pubsys_config::InfraConfig;
+use snafu::ResultExt;
+use std::collections::{HashMap, HashSet};
+use std::fs::File;
+use std::path::PathBuf;
+use structopt::{clap, StructOpt};
+
+/// Validates EC2 images by calling `describe-images` on all images in the file given by
+/// `expected-amis-path` and ensuring that the returned `public`, `ena-support`,
+/// `sriov-net-support`, and `launch-permissions` fields have the expected values.
+#[derive(Debug, StructOpt)]
+#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+pub(crate) struct ValidateAmiArgs {
+    /// File holding the expected amis
+    #[structopt(long, parse(from_os_str))]
+    expected_amis_path: PathBuf,
+
+    /// Optional path where the validation results should be written
+    #[structopt(long, parse(from_os_str))]
+    write_results_path: Option<PathBuf>,
+
+    #[structopt(long, requires = "write-results-path")]
+    /// Optional filter to only write validation results with these statuses to the above path
+    /// The available statuses are: `Correct`, `Incorrect`, `Missing`.
+    write_results_filter: Option<Vec<AmiValidationResultStatus>>,
+
+    #[structopt(long)]
+    /// If this argument is given, print the validation results summary as a JSON object instead
+    /// of a plaintext table
+    json: bool,
+}
+
+/// Performs EC2 image validation and returns the `AmiValidationResults` object
+pub(crate) async fn validate(
+    args: &Args,
+    validate_ami_args: &ValidateAmiArgs,
+) -> Result<AmiValidationResults> {
+    info!("Parsing Infra.toml file");
+
+    // If a lock file exists, use that, otherwise use Infra.toml
+    let infra_config = InfraConfig::from_path_or_lock(&args.infra_config_path, false)
+        .context(error::ConfigSnafu)?;
+
+    trace!("Parsed infra config: {:#?}", infra_config);
+
+    let aws = infra_config.aws.unwrap_or_default();
+
+    // Parse the expected ami file
+    info!("Parsing expected ami file");
+    let expected_images = parse_expected_amis(&validate_ami_args.expected_amis_path).await?;
+
+    info!("Parsed expected ami file");
+
+    // Create a `HashMap` of `AmiClient`s, one for each region where validation should happen
+    let base_region = &Region::new(
+        aws.regions
+            .get(0)
+            .ok_or(error::Error::EmptyInfraRegions {
+                path: args.infra_config_path.clone(),
+            })?
+            .clone(),
+    );
+    let mut ami_clients = HashMap::with_capacity(expected_images.len());
+
+    for region in expected_images.keys() {
+        let client_config = build_client_config(region, base_region, &aws).await;
+        let ami_client = AmiClient::new(&client_config);
+        ami_clients.insert(region.clone(), ami_client);
+    }
+
+    // Retrieve the EC2 images using the `AmiClient`s
+    info!("Retrieving EC2 images");
+    let images = describe_images(&ami_clients, &expected_images)
+        .await
+        .into_iter()
+        .map(|(region, result)| {
+            (
+                region,
+                result.map_err(|e| {
+                    error!(
+                        "Failed to retrieve images in region {}: {}",
+                        region.to_string(),
+                        e
+                    );
+                    error::Error::UnreachableRegion {
+                        region: region.to_string(),
+                    }
+                }),
+            )
+        })
+        .collect::<HashMap<&Region, Result<_>>>();
+
+    // Validate the retrieved EC2 images per region
+    info!("Validating EC2 images");
+    let results: HashMap<Region, HashSet<AmiValidationResult>> = images
+        .into_iter()
+        .map(|(region, region_result)| {
+            (
+                region.clone(),
+                validate_images_in_region(
+                    &expected_images
+                        .get(region)
+                        .map(|e| e.to_owned())
+                        .unwrap_or_default(),
+                    &region_result,
+                    region,
+                ),
+            )
+        })
+        .collect();
+
+    let validation_results = AmiValidationResults::from_result_map(results);
+
+    // If a path was given, write the results
+    if let Some(write_results_path) = &validate_ami_args.write_results_path {
+        // Filter the results by given status, and if no statuses were given, get all results
+        info!("Writing results to file");
+        let results = if let Some(filter) = &validate_ami_args.write_results_filter {
+            validation_results.get_results_for_status(filter)
+        } else {
+            validation_results.get_all_results()
+        };
+
+        // Write the results as JSON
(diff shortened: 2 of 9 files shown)


tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+    /// File holding the expected amis
+    #[structopt(long, parse(from_os_str))]
+    expected_amis_path: PathBuf,
+
+    /// Optional path where the validation results should be written
+    #[structopt(long, parse(from_os_str))]
+    write_results_path: Option<PathBuf>,
+
+    #[structopt(long, requires = "write-results-path")]
+    /// Optional filter to only write validation results with these statuses to the above path
```
cbgbt, line 36: nit, but the help text looks funky without this.

```suggestion
    /// Optional filter to only write validation results with these statuses to the above path.
```

tools/pubsys/src/aws/validate_ami/ami.rs
```diff
@@ -0,0 +1,195 @@
...
+            // Will be populated later if the image is not expected to be public. If it is expected
+            // to be public, then individual launch permissions are irrelevant
+            launch_permissions: None,
+            ena_support: image.ena_support().unwrap_or_default(),
+            sriov_net_support: image.sriov_net_support().unwrap_or_default().to_string(),
+        }
+    }
+}
+
+pub(crate) async fn describe_images<'a>(
```
cbgbt: Could use a docstring here to explain why we need this.

tools/pubsys/src/aws/validate_ami/ami.rs
```diff
@@ -0,0 +1,195 @@
...
+        .collect()
+        .await
+}
+
+/// Fetches all images in a single region
+pub(crate) async fn describe_images_in_region(
+    region: &Region,
+    client: &Ec2Client,
+    image_ids: Vec<String>,
+    expected_image_public: &HashMap<String, bool>,
```
cbgbt: I'd document what the purpose of this parameter is in the docstring, since it's not obvious from the function signature.

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,1028 @@
...
+        }
+    }
+}
+
+#[derive(Tabled, Serialize)]
+struct AmiValidationRegionSummary {
+    correct: i32,
+    incorrect: i32,
+    missing: i32,
+    accessible: bool,
```
cbgbt: I don't have strong feelings about rewriting this, but I'm realizing that because `accessible` removes meaning from the other values here, I think a more maintainable way to write this would be to:

* Change `AmiValidationRegionSummary` to an enum of something like

```rust
enum AmiValiationRegionSummary {
    Results {
        correct: u64, // etc
    },
    Unreachable,
```

* Modify `display()` to render this by copying it to a private implementation which uses the marker values, or `n/a` or something similar in the table.

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+                            .get(region)
+                            .map(|e| e.to_owned())
+                            .unwrap_or(vec![]),
+                        &result,
+                        region,
+                    )
+                }),
+            )
+        })
+        .collect::<HashMap<Region, ami::Result<HashSet<AmiValidationResult>>>>();
```
cbgbt: Bit of a gnarly turbofish that probably isn't needed, since there's a type specifier on the variable binding.

```suggestion
        .collect();
```

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+
+    let validation_results = AmiValidationResults::new(results);
```
cbgbt: nit, but usually `new` isn't appropriate when you're initializing something that's computation heavy like that. A better name would be `from_result_map` or `evaluate_results()` or something.

```suggestion
    let validation_results = AmiValidationResults::new(results);
```

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+            path: expected_amis_path,
+        })?,
+    )
+    .context(error::ParseExpectedImagesFileSnafu)?;
+
+    // If the value for each Region is a JSON object instead of an array, wrap it in an array
+    let vectored_images = expected_amis
+        .into_iter()
+        .map(|(region, value)| {
+            let image_values = match value.as_array() {
```
cbgbt: Can we write a test for this behavior? As an alternative tip, you can get Serde to parse this for you "for free" via `#[serde(untagged)]` on an enum.

```rust
enum ImageDef {
    Image(Image),
    ImageList(Vec<Image>),
}

impl ImageDef {
    fn images(&self) -> Vec<Image> {
        match self {
            // etc
        }
    }
}
```

cbgbt (commented):
Nice work! Most of these are just nitpicks or tips. The thing that I care the most to see changed is to add a test for the AMI file parsing code. The function is a pure function and the logic inside is complex enough to warrant it.

tools/pubsys/src/aws/validate_ami/ami.rs
```diff
@@ -0,0 +1,195 @@
...
+    clients: &'a HashMap<Region, Ec2Client>,
+    image_ids: &HashMap<Region, Vec<String>>,
+    expected_image_public: &HashMap<String, bool>,
+) -> HashMap<&'a Region, Result<HashMap<String, ImageDef>>> {
+    // Build requests for images; we have to request with a regional client so we split them by
+    // region
+    let mut requests = Vec::with_capacity(clients.len());
+    for region in clients.keys() {
+        trace!("Requesting images in {}", region);
+        let ec2_client: &Ec2Client = &clients[region];
```
ecpullen:
```suggestion
        let ec2_client: &Ec2Client = &clients.get(region).context()...;
```

Indexing with `[]` can panic.

tools/pubsys/src/aws/validate_ami/ami.rs
```diff
@@ -0,0 +1,195 @@
...
+}
+
+pub(crate) async fn describe_images<'a>(
+    clients: &'a HashMap<Region, Ec2Client>,
+    image_ids: &HashMap<Region, Vec<String>>,
+    expected_image_public: &HashMap<String, bool>,
+) -> HashMap<&'a Region, Result<HashMap<String, ImageDef>>> {
+    // Build requests for images; we have to request with a regional client so we split them by
+    // region
+    let mut requests = Vec::with_capacity(clients.len());
```
ecpullen, line 88: Using an iterator for this function could simplify some of the logic below.

tools/pubsys/src/aws/validate_ami/ami.rs
```diff
@@ -0,0 +1,195 @@
...
+                },
+            )?;
+            // If the image is not expected to be public, retrieve the launch permissions
+            trace!(
+                "Retrieving launch permissions for {} in {}",
+                image_id,
+                region.as_ref()
+            );
+            let mut image_def = ImageDef::from(image.to_owned());
+            if !*expected_public {
```
ecpullen: Can this go above 141 and then make `ImageDef::from()` take in the bool so image_def doesn't need to be mut?

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+    args: &Args,
+    validate_ami_args: &ValidateAmiArgs,
+) -> Result<AmiValidationResults> {
+    info!("Parsing Infra.toml file");
+
+    // If a lock file exists, use that, otherwise use Infra.toml
+    let infra_config = InfraConfig::from_path_or_lock(&args.infra_config_path, false)
+        .context(error::ConfigSnafu)?;
+
+    let aws = infra_config.aws.clone().unwrap_or_default();
```
ecpullen: Does this need `.clone()`

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+    trace!("Parsed infra config: {:#?}", infra_config);
+
+    // Parse the expected ami file
+    info!("Parsing expected ami file");
+    let expected_images = parse_expected_amis(&validate_ami_args.expected_amis_path).await?;
+
+    info!("Parsed expected ami file");
+
+    // Create a HashMap of AmiClients, one for each region where validation should happen
+    let base_region = &Region::new(aws.regions[0].clone());
```
ecpullen:
```suggestion
    let base_region = &Region::new(aws.regions.get(0).context()...);
```
`[]` can panic

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+    for region in expected_images.keys() {
+        let client_config = build_client_config(region, base_region, &aws).await;
+        let ami_client = AmiClient::new(&client_config);
+        ami_clients.insert(region.clone(), ami_client);
+    }
+
+    // Create a map of image_id to a bool indicating whether or not the image is public. This map
+    // is needed to determine if launch permissions should be retrieved for the image (which is the
+    // case if it is not public)
+    let expected_image_public = expected_images
```
ecpullen: We are already passing expected_images into the describe_images function. Is it necessary to duplicate the image public?

tools/pubsys/src/aws/validate_ami/ami.rs
```diff
@@ -0,0 +1,195 @@
...
+            sriov_net_support: image.sriov_net_support().unwrap_or_default().to_string(),
+        }
+    }
+}
+
+pub(crate) async fn describe_images<'a>(
+    clients: &'a HashMap<Region, Ec2Client>,
+    image_ids: &HashMap<Region, Vec<String>>,
+    expected_image_public: &HashMap<String, bool>,
+) -> HashMap<&'a Region, Result<HashMap<String, ImageDef>>> {
```
ecpullen, line 85:
```suggestion
) -> Result<HashMap<&'a Region, HashMap<String, ImageDef>>> {
```

Can we extract this Result out?

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+        let filtered_results = validation_results.get_results_for_status(
+            validate_ami_args
+                .write_results_filter
+                .as_ref()
+                .unwrap_or(&vec![
+                    AmiValidationResultStatus::Correct,
+                    AmiValidationResultStatus::Incorrect,
+                    AmiValidationResultStatus::Missing,
+                ]),
+        );
```
ecpullen:
```suggestion
        let results = if let Some(filter) = validate_ami_args.write_results_filter {
            validation_results.get_results_for_status(filter)
        else {
            validation_results.all_results()
       };
```
It's a bit unusual to list all variants of an enum as a default. You could also have `get_results_for_status` take an Option.

```suggestion
        let results = validation_results.get_results_for_status(validate_ami_args.write_results_filter);
```

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+            &filtered_results,
+        )
+        .context(error::SerializeValidationResultsSnafu)?;
+    }
+
+    Ok(validation_results)
+}
+
+/// Validates EC2 images in a single region, based on a Vec (ImageDef) of expected images
+/// and a HashMap (AmiId, ImageDef) of actual retrieved images. Returns a HashSet of
```
ecpullen: Super nit

```suggestion
/// Validates EC2 images in a single region, based on a `Vec<ImageDef>` of expected images
/// and a `HashMap<AmiId, ImageDef>` of actual retrieved images. Returns a `HashSet<AmiValidationResult>` containing ...
```

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+/// AmiValidationResult objects.
+pub(crate) fn validate_images_in_region(
+    expected_images: Vec<ImageDef>,
+    actual_images: &HashMap<AmiId, ImageDef>,
```
ecpullen: Is there a reason 1 is a reference and the other is taking ownership?

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+) -> HashSet<AmiValidationResult> {
+    let mut results = HashSet::new();
+
+    // Validate all expected images, creating an AmiValidationResult object
+    for mut image in expected_images {
+        // If the image is expected to be public, the specific launch permissions are irrelevant
+        // and were not retrieved for the actual image
+        if image.public {
+            image.launch_permissions = None;
+        }
```
ecpullen: Why isn't `image.launch_permissions` set to None when the `ImageDef` is created? If that's not possible can `launch_permissions` be a function?

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+        results.insert(AmiValidationResult::new(
+            image.id.clone(),
+            image.clone(),
```
ecpullen: Do we need to duplicate the `image.id`?

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+) -> Result<HashMap<Region, Vec<ImageDef>>> {
+    // Parse the JSON file as a HashMap of region_name, mapped to a JSON value
+    let expected_amis: HashMap<RegionName, serde_json::Value> = serde_json::from_reader(
+        &File::open(expected_amis_path.clone()).context(error::ReadExpectedImagesFileSnafu {
+            path: expected_amis_path,
+        })?,
+    )
+    .context(error::ParseExpectedImagesFileSnafu)?;
+
+    // If the value for each Region is a JSON object instead of an array, wrap it in an array
```
ecpullen: What would cause the value to be an object instead of an array? This is a file that another pubsys command is writing right?

ecpullen (commented):
Still need to check `validate_ami/results.rs`

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,1028 @@
...
+
+impl Display for AmiValidationResultStatus {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            AmiValidationResultStatus::Correct => write!(f, "Correct"),
+            AmiValidationResultStatus::Incorrect => write!(f, "Incorrect"),
+            AmiValidationResultStatus::Missing => write!(f, "Missing"),
+        }
+    }
+}
```
ecpullen:
```suggestion
derive_display_from_serialize!(AmiValidationResultStatus);
```
That will have the same effect. And expand if new variants of the enum are added.

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,1028 @@
...
+        match s {
+            "Correct" => Ok(Self::Correct),
+            "Incorrect" => Ok(Self::Incorrect),
+            "Missing" => Ok(Self::Missing),
+            filter => Err(Self::Err::InvalidStatusFilter {
+                filter: filter.to_string(),
+            }),
+        }
+    }
+}
```
ecpullen:
```suggestion
derive_fromstr_from_deserialize!(AmiValidationResultStatus)
```

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,1028 @@
...
+
+    /// ImageDef containing actual values for the image
+    pub(crate) actual_image_def: Option<ImageDef>,
+
+    /// The region the image resides in
+    #[serde(serialize_with = "serialize_region")]
+    pub(crate) region: Region,
+
+    /// The validation status of the image
+    pub(crate) status: AmiValidationResultStatus,
```
ecpullen, line 48: This should just be a function since it is described by other fields of this enum.

ecpullen (commented):
+1 to the suggested `AmiValidationRegionSummary` refactor. Otherwise just a few small things.

tools/pubsys/src/aws/validate_ami/results.rs
mjsterckx: It's currently implemented the same way in the ssm-validation so I will address those together in a different PR.

tools/pubsys/src/aws/validate_ami/ami.rs
mjsterckx: We can't, because that Result indicates whether or not the requests in that specific region were successful (i.e. if that region was reachable or not). If we extract it out, we only know whether or not all requests in all regions were successful.

tools/pubsys/src/aws/validate_ami/mod.rs
mjsterckx: This works the same way in the ssm-validation, so I will address those together in a different PR.

tools/pubsys/src/aws/validate_ami/mod.rs
mjsterckx: Correct, this takes a file that is written by `pubsys ami` where each region has a single AMI object. However, I wanted to add functionality where the command can take a file containing an array of AMI objects per region as well.

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+        .into_iter()
+        .map(|(region, region_result)| {
+            (
+                region.clone(),
+                region_result.map(|result| {
+                    validate_images_in_region(
+                        expected_images
+                            .get(region)
+                            .map(|e| e.to_owned())
+                            .unwrap_or(vec![]),
```
etungsten: nit:
```suggestion
                            .unwrap_or_default();,
```

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+    Ok(())
+}
+
+mod error {
+    use snafu::Snafu;
+    use std::path::PathBuf;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub(crate) enum Error {
```
etungsten, line 241: Please check the error enum variants and ensure they're being used. I see a few that aren't being used anywhere. e.g. `ReadValidationConfig`, `ParseValidationConfig`, `MissingField` etc.

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,1028 @@
...
+                filter: filter.to_string(),
+            }),
+        }
+    }
+}
+
+/// Represents a single EC2 image validation result
+#[derive(Debug, Eq, Hash, PartialEq, Serialize)]
+pub(crate) struct AmiValidationResult {
+    /// The id of the image
```
etungsten: nit: proper capitalization

```suggestion
    /// The ID of the image
```

and where ever else this may apply

tools/pubsys/src/aws/validate_ami/results.rs
```diff
@@ -0,0 +1,1028 @@
...
+    pub(crate) id: String,
+
+    /// ImageDef containing expected values for the image
```
etungsten: nit:
```suggestion
    /// `ImageDef` containing expected values for the image
```

and where ever else this may apply.

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+use aws_sdk_ec2::{Client as AmiClient, Region};
+use log::{info, trace};
+use pubsys_config::InfraConfig;
+use snafu::ResultExt;
+use std::collections::{HashMap, HashSet};
+use std::fs::File;
+use std::path::PathBuf;
+use structopt::{clap, StructOpt};
+
+/// Validates EC2 images
```
webern: Describe this in more detail. What's the input, what's the output, etc.

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+                        region,
+                    )
+                }),
+            )
+        })
+        .collect::<HashMap<Region, ami::Result<HashSet<AmiValidationResult>>>>();
+
+    let validation_results = AmiValidationResults::new(results);
+
+    // If a path was given to write the results to, write the results
```
webern: Writing nit

```suggestion
    // If a path was given, write the results
```

tools/pubsys/src/aws/validate_ami/mod.rs
webern: > Please check the error enum variants and ensure they're being used. I see a few that aren't being used anywhere. e.g. ReadValidationConfig, ParseValidationConfig, MissingField etc.

As an aside, the reason you have to check is because the code that Snafu generates "uses" each enum variant so the compiler can't warn you about unused variants.

webern (commented):
I love the testing.

tools/pubsys/src/aws/validate_ami/ami.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    #[allow(clippy::large_enum_variant)]
+    pub(crate) enum Error {
+        #[snafu(display("Failed to describe images in {}: {}", region, source))]
+        DescribeImages {
+            region: String,
+            source: SdkError<DescribeImagesError>,
+        },
```
etungsten: Where-ever we're using a `SdkError` as source we need to do this to get the full error message:

```suggestion
use aws_smithy_types::error::display::DisplayErrorContext;

        #[snafu(display("Failed to describe images in {}: {}", region, DisplayErrorContext(source)))]
        DescribeImages {
            region: String,
            source: SdkError<DescribeImagesError>,
        },
```

See https://docs.rs/aws-smithy-client/latest/aws_smithy_client/enum.SdkError.html
> When logging an error from the SDK, it is recommended that you either wrap the error in [DisplayErrorContext](https://docs.rs/aws-smithy-types/0.55.1/x86_64-unknown-linux-gnu/aws_smithy_types/error/display/struct.DisplayErrorContext.html), use another error reporter library that visits the error’s cause/source chain, or call [Error::source](https://doc.rust-lang.org/nightly/core/error/trait.Error.html#method.source) for more details about the underlying cause.

**Note: This needs to be a tree wide change, but might as well start here.** 

Please create an issue for the other places if you're willing.

tools/pubsys/src/aws/validate_ami/results.rs
cbgbt: I think I suggested this be added as a field in a prior revision, since it's otherwise not really a `Result` being computed, but rather a validation computation being created which is later used to compute the result.

I guess if we renamed this `AmiValidation` with a method `AmiValidation::get_result()` or something, I'd be ok with that.

But aren't these serialized? Do we want to keep the status so that the computation result is specifically kept?

tools/pubsys/src/aws/validate_ami/ami.rs
```diff
@@ -0,0 +1,220 @@
...
+            name: args.0.name().unwrap_or_default().to_string(),
+            public: args.0.public().unwrap_or_default(),
+            launch_permissions: args.1,
+            ena_support: args.0.ena_support().unwrap_or_default(),
+            sriov_net_support: args.0.sriov_net_support().unwrap_or_default().to_string(),
+        }
+    }
+}
+
+/// Fetches all images with IDs in `image_ids`. The map `expected_image_public` is used to
```
cbgbt: It looks like the docstrings for this and `describe_images_in_region` may not have been updated for some subsequent changes to the code. I would fix this before merging.

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -0,0 +1,847 @@
...
+    info!("Retrieving EC2 images");
+    let images = describe_images(&ami_clients, &expected_images)
+        .await
+        .into_iter()
+        .map(|(region, result)| {
+            (
+                region,
+                match result {
+                    Ok(result) => Ok(result),
+                    Err(_) => Err(error::Error::UnreachableRegion {
```
cbgbt: nit/tip (don't feel compelled to change this): I think this is the same as

```rust
result.map_err(|_| error::Error::UnreachableRegion { region: region.to_string() })
```

`match` on `Result` and `Option` when the operation to be carried out is already expressed as a method on those structs tends to result in deep nesting.

tools/pubsys/src/aws/validate_ssm/results.rs
```diff
@@ -120,7 +91,7 @@ struct SsmValidationRegionSummary {
     incorrect: i32,
     missing: i32,
     unexpected: i32,
-    accessible: bool,
+    unreachable: i32,
```
cbgbt: nit: these should probably be unsigned.

tools/pubsys/src/aws/validate_ssm/mod.rs
```diff
@@ -80,52 +80,56 @@ pub async fn validate(
...
+        .await
+        .into_iter()
+        .map(|(region, result)| {
+            (
+                region,
+                match result {
+                    Ok(result) => Ok(result),
+                    Err(_) => Err(error::Error::UnreachableRegion {
+                        region: region.to_string(),
+                    }),
```
cbgbt: Does it make sense for this to hold onto the original error?

I think we just want to make sure the reason for the unavailability can be surfaced. Ensuring that it is logged somewhere at `ERROR` level probably suffices, since it's not like the deeper result information is stored in our serialized format.

cbgbt (commented):
Looks good!

Sorry to pile on more feedback -- this is a substantial change! Just some tidying up.

cbgbt (approved):

etungsten (approved):

PR #3035, opened 2023-04-19
Storewolf: adds feature to create inventory file symlink

### LINKED ISSUE #2967: Bottlerocket inventory file symlink on host
**What I'd like:**

I'd like the file that lives at
```
/usr/share/bottlerocket/application-inventory.json
```
on the bottlerocket host and is surfaced to
```
/var/lib/bottlerocket/inventory/application.json
```

within host containers to _also_ be linked to the same location on the host.

This way, software that typically runs as a host-container (for example, the ssm-agent) may be added as a host level packages (as part of the OS) and not have to modify the location of the file it looks for to consume the Bottlerocket inventory.

This would make things alittle easier for people building bespoke Bottlerocket and dropping software onto the host that consumes the inventory.

jpmcb:
**Issue number:**

Closes https://github.com/bottlerocket-os/bottlerocket/issues/2967

**Description of changes:**

Feature for `storewolf` to create a symlink at a given path and destination:

```bash
/usr/bin/storewolf \
    --data-store-base-path /var/lib/bottlerocket/datastore \
    --inventory-file-symlink-path /var/lib/bottlerocket/inventory/application.json
```

will then create the a symlink to the default file location `/usr/share/bottlerocket/application-inventory.json`. Users may also specify the source location of the inventory file via the optional `--inventory-file-path` flag.

**Testing done:**

Built bottlerocket, launched instance and see symlink created correctly:
```
bash-5.1# ls -la /var/lib/bottlerocket/inventory/
total 8
drwxr-xr-x. 2 root root 4096 Apr 19 16:37 .
drwxr-xr-x. 5 root root 4096 Apr 19 16:37 ..
lrwxrwxrwx. 1 root root   50 Apr 19 16:37 application.json -> /usr/share/bottlerocket/application-inventory.json
```

Contents looks good:
```
bash-5.1# cat /var/lib/bottlerocket/inventory/application.json | head
{
  "Content": [
    {
      "Name": "acpid",
      "Publisher": "Bottlerocket",
      "Version": "1.14.0",
      "Release": "1ecaea34-dirty",
      "InstalledTime": "2023-04-19T16:35:20Z",
      "ApplicationType": "Unspecified",
      "Architecture": "x86_64",
```

And bootlogs look good:
```
bash-5.1# journalctl | grep "storewolf"
Apr 19 16:37:54 localhost storewolf[1537]: 16:37:54 [INFO] Storewolf started
Apr 19 16:37:54 localhost storewolf[1537]: 16:37:54 [INFO] Deleting pending transactions
Apr 19 16:37:54 localhost storewolf[1537]: 16:37:54 [INFO] Populating datastore at: /var/lib/bottlerocket/datastore
Apr 19 16:37:54 localhost storewolf[1537]: 16:37:54 [INFO] Creating datastore at: /var/lib/bottlerocket/datastore/current/live
Apr 19 16:37:54 localhost storewolf[1537]: 16:37:54 [INFO] Datastore populated
Apr 19 16:37:54 localhost storewolf[1537]: 16:37:54 [INFO] Creating inventory file symlink at: /var/lib/bottlerocket/inventory/application.json
Apr 19 16:37:54 localhost storewolf[1537]: 16:37:54 [INFO] Inventory symlink created
```

**Terms of contribution:**

### DIFF
--- a/packages/os/storewolf.service
+++ b/packages/os/storewolf.service
@@ -9,7 +9,9 @@ RefuseManualStop=true
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/storewolf --data-store-base-path /var/lib/bottlerocket/datastore
+ExecStart=/usr/bin/storewolf \
+  --data-store-base-path /var/lib/bottlerocket/datastore \
+  --inventory-file-symlink-path /var/lib/bottlerocket/inventory/application.json
 RemainAfterExit=true
 StandardError=journal+console
 
--- a/sources/api/storewolf/src/main.rs
+++ b/sources/api/storewolf/src/main.rs
@@ -15,6 +15,7 @@ use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
 use std::io;
+use std::os::unix;
 use std::path::Path;
 use std::str::FromStr;
 use std::{env, fs, process};
@@ -23,6 +24,9 @@ use datastore::key::{Key, KeyType};
 use datastore::serialization::{to_pairs, to_pairs_with_prefix};
 use datastore::{self, DataStore, FilesystemDataStore, ScalarError};
 use model::modeled_types::SingleLineString;
+
+// The default path to the RPM inventory json file
+const INVENTORY_PATH: &str = "/usr/share/bottlerocket/application-inventory.json";
 
 mod error {
     use std::io;
@@ -107,6 +111,18 @@ mod error {
 
         #[snafu(display("Keys can't contain newlines: {}", source))]
         SingleLineString { source: ModeledTypesError },
+
+        #[snafu(display("Could not get parent directory from inventory symlink path: {}", path.display()))]
+        SymlinkParentDirPath { path: PathBuf },
+
+        #[snafu(display("The symlink parent dir '{}' could not be created: {}.", path.display(), source))]
+        SymlinkParentDirCreate { path: PathBuf, source: io::Error },
+
+        #[snafu(display("Could not delete old inventory symlink: {}", source))]
+        SymlinkDelete { source: io::Error },
+
+        #[snafu(display("Could not create new inventory symlink: {}", source))]
+        SymlinkCreate { source: io::Error },
     }
 }
 
@@ -385,9 +401,38 @@ fn populate_default_datastore<P: AsRef<Path>>(
     Ok(())
 }
 
+/// Creates a symlink given a source path and a destination path. Creates the
+/// necessary parent directories along the destination path.
+fn create_inventory_symlink<P1, P2>(source: P1, destination: P2) -> Result<()>
+where
+    P1: AsRef<Path>,
+    P2: AsRef<Path>,
+{
+    let parent_dir = destination
+        .as_ref()
+        .parent()
+        .context(error::SymlinkParentDirPathSnafu {
+            path: destination.as_ref(),
+        })?;
+
+    fs::create_dir_all(parent_dir)
+        .context(error::SymlinkParentDirCreateSnafu { path: parent_dir })?;
+
+    // Remove symlink if it already exists
+    if let Err(e) = fs::remove_file(&destination) {
+        if e.kind() != io::ErrorKind::NotFound {
+            return Err(e).context(error::SymlinkDeleteSnafu);
+        }
+    }
+
+    unix::fs::symlink(source, destination).context(error::SymlinkCreateSnafu)
+}
+
 /// Store the args we receive on the command line
 struct Args {
     data_store_base_path: String,
+    inventory_symlink_path: String,
+    inventory_file_path: String,
     log_level: LevelFilter,
     version: Option<Version>,
 }
@@ -398,6 +443,8 @@ fn usage() -> ! {
     eprintln!(
         r"Usage: {}
             --data-store-base-path PATH
+            --inventory-file-symlink-path PATH
+            [ --inventory-file-path PATH (default: /usr/share/bottlerocket/application-inventory.json) ]
             [ --version X.Y.Z ]
             [ --log-level trace|debug|info|warn|error ]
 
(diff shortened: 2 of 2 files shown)


packages/os/storewolf.service
```diff
@@ -9,7 +9,7 @@ RefuseManualStop=true
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/storewolf --data-store-base-path /var/lib/bottlerocket/datastore
+ExecStart=/usr/bin/storewolf --data-store-base-path /var/lib/bottlerocket/datastore --inventory-file-symlink-path /var/lib/bottlerocket/inventory/application.json
```
webern: Does a new symlink on the system need to be deleted on downgrade? @bcressey? Seems harmless, right?

jpmcb:
Needs a rebase and `cargo fmt` now that https://github.com/bottlerocket-os/bottlerocket/pull/2848 has merged - incoming!

packages/os/storewolf.service
bcressey: nit:
```suggestion
ExecStart=/usr/bin/storewolf \
  --data-store-base-path /var/lib/bottlerocket/datastore \
  --inventory-file-symlink-path /var/lib/bottlerocket/inventory/application.json
```

sources/api/storewolf/src/main.rs
```diff
@@ -385,9 +398,25 @@ fn populate_default_datastore<P: AsRef<Path>>(
...
+fn create_inventory_symlink<P: AsRef<Path>>(source: P, destination: P) -> Result<()> {
+    let parent_dir = destination
+        .as_ref()
+        .parent()
+        .context(error::SymlinkParentDirPathSnafu {
+            path: destination.as_ref(),
+        })?;
+    fs::create_dir_all(parent_dir)
+        .context(error::SymlinkParentDirCreateSnafu { path: parent_dir })?;
+    unix::fs::symlink(source, destination).context(error::SymlinkCreateSnafu)
```
bcressey: AFAIK this doesn't replace any existing symlink, so you'd want to remove it first to make sure the correct symlink gets created.

The approach elsewhere in `sources` is to create the link with a temporary name and rename it on top of the target, which would also work.

sources/api/storewolf/src/main.rs
```diff
@@ -385,9 +398,25 @@ fn populate_default_datastore<P: AsRef<Path>>(
     Ok(())
 }
 
+/// Creates a symlink given a source path and a destination path. Creates the
+/// necessary parent directories along the destination path.
+fn create_inventory_symlink<P: AsRef<Path>>(source: P, destination: P) -> Result<()> {
```
zmrow: Nit: Probably not a big deal in this particular case, but `source` and `destination` are both `P: Asref<Path>`.  This means they both must be the same underlying type at function call time; you couldn't pass a `String` and a `&PathBuf` for example.

To be a bit more flexible, you can use the same trait bound, but split into multiple params, which allows the underlying types to be different. (Excuse the formatting)
```
fn create_inventory_symlink<P1, P2>(source: P1, destination: P2) -> Result<()>
    where P1: AsRef<Path>, P2: AsRef<Path>
{
...
}
```

zmrow (approved):
Provided @bcressey / @webern 's feedback is addressed, this looks good.

One non-blocking nit. :)

packages/os/storewolf.service
jpmcb: I'm not sure - my assumption is "probably" since a downgrade would likely carry a different version of the rpm json inventory.  But with the "delete" functionality that will remove an existing symlink if it's there is abit of a moot point.

sources/api/storewolf/src/main.rs
jpmcb: Huh interesting! I was not aware that calling into the params type would assume the same underlying type _for both_. I'll make that change 👍🏼

jpmcb:
Force pushed to address Ben's comments on deleting the symlink (if it's there) before attempting to create it.

jpmcb:
Force pushed to address Zac's comment: now use two different `AsRef` params - tests still look good:
```
bash-5.1# ls -la /var/lib/bottlerocket/inventory/application.json
lrwxrwxrwx. 1 root root 50 Apr 24 15:51 /var/lib/bottlerocket/inventory/application.json -> /usr/share/bottlerocket/application-inventory.json
bash-5.1# journalctl | grep storewolf
Apr 24 15:51:26 localhost storewolf[1581]: 15:51:26 [INFO] Storewolf started
Apr 24 15:51:26 localhost storewolf[1581]: 15:51:26 [INFO] Deleting pending transactions
Apr 24 15:51:26 localhost storewolf[1581]: 15:51:26 [INFO] Populating datastore at: /var/lib/bottlerocket/datastore
Apr 24 15:51:26 localhost storewolf[1581]: 15:51:26 [INFO] Creating datastore at: /var/lib/bottlerocket/datastore/current/live
Apr 24 15:51:26 localhost storewolf[1581]: 15:51:26 [INFO] Datastore populated
Apr 24 15:51:26 localhost storewolf[1581]: 15:51:26 [INFO] Creating inventory file symlink at: /var/lib/bottlerocket/inventory/application.json
Apr 24 15:51:26 localhost storewolf[1581]: 15:51:26 [INFO] Inventory symlink created
```

jpmcb:
Force pushed to add migration and rebase on `develop`

jpmcb:
Force pushed for `cargo make check` shenanigans.

webern (approved):

sources/api/migration/migrations/v1.14.0/storewolf-inventory-symlink/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+                return Err(e).context(error::RemoveFileSnafu {
+                    path: INVENTORY_SYMLINK_PATH,
+                });
+            } else {
+                println!("NotFound: '{}'", INVENTORY_SYMLINK_PATH)
+            }
+        }
+        Ok(input)
+    }
+}
```
bcressey: A migration for this seems like overkill. I'm not opposed to including it but I don't see any real upside.

jpmcb:
Rebased to fix merge conflict in `Release.toml`

bcressey (approved):

jpmcb:
Force pushed to remove migration per @bcressey! Will merge once tests pass again!

PR #3046, opened 2023-04-24
Add kubernetes-1.27 package and variants

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This adds a new package for installing k8s 1.27. It also adds all variants for the *-k8s-* variant.

**Testing done:**

- [x] aarch64 IPv4 conformance tests
- [x] aarch64 IPv6 conformance tests
- [x] aarch64 NVIDIA smoke tests
- [x] x86_64 IPv4 conformance tests
- [x] x86_64 IPv6 conformance tests
- [x] x86_64 NVIDIA smoke tests
- [ ] `metal-k8s-1.27` validation
- [x] `vmware-k8s-1.27` validation

**Terms of contribution:**


sources/models/README.md
```diff
@@ -72,6 +72,16 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
...
 
+### aws-k8s-1.27: Kubernetes 1.27
+
+* [Model](src/aws-k8s-1.26/mod.rs)
+* [Default settings](src/aws-k8s-1.26/defaults.d/)
+
+### aws-k8s-1.27-nvidia: Kubernetes 1.27 NVIDIA
+
+* [Model](src/aws-k8s-1.26-nvidia/mod.rs)
+* [Default settings](src/aws-k8s-1.26-nvidia/defaults.d/)
```
gthao313: should be 1.27.

sources/models/README.md
```diff
@@ -112,6 +122,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/vmware-k8s-1.26/mod.rs)
 * [Default settings](src/vmware-k8s-1.26/defaults.d/)
 
+### vmware-k8s-1.27: VMware Kubernetes 1.27
+
+* [Model](src/vmware-k8s-1.26/mod.rs)
+* [Default settings](src/vmware-k8s-1.26/defaults.d/)
```
gthao313:
```suggestion
* [Model](src/vmware-k8s-1.27/mod.rs)
* [Default settings](src/vmware-k8s-1.27/defaults.d/)
```

sources/models/README.md
```diff
@@ -142,6 +157,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/metal-k8s-1.26/mod.rs)
 * [Default settings](src/metal-k8s-1.26/defaults.d/)
 
+### metal-k8s-1.27: Metal Kubernetes 1.27
+
+* [Model](src/metal-k8s-1.26/mod.rs)
+* [Default settings](src/metal-k8s-1.26/defaults.d/)
```
gthao313:
```suggestion
* [Model](src/metal-k8s-1.27/mod.rs)
* [Default settings](src/metal-k8s-1.27/defaults.d/)
```

sources/models/README.md
stmcginnis: Odd, none of the changes from `lib.rs` ended up here. Ran a cargo build directly from this directory and now it worked. Changes staged and will push up with next update.

Makefile.toml
```diff
@@ -28,9 +28,9 @@ BUILDSYS_PRETTY_NAME = "Bottlerocket OS"
 # SDK name used for building
 BUILDSYS_SDK_NAME="bottlerocket"
 # SDK version used for building
-BUILDSYS_SDK_VERSION="v0.31.0"
+BUILDSYS_SDK_VERSION="thar-be-go-updates"
```
jpmcb: FYI, ready for a rebase and use `v0.32.0` of the SDK  🎈🕺🏼

packages/kubernetes-1.27/prestart-pull-pause-ctr.conf
```diff
@@ -0,0 +1,9 @@
+[Service]
+# Pull the pause container image before starting `kubelet` so `containerd/cri` wouldn't have to
+ExecStartPre=/usr/bin/host-ctr \
+    --containerd-socket=/run/containerd/containerd.sock \
+    --namespace=k8s.io \
+    pull-image \
+    --source=${POD_INFRA_CONTAINER_IMAGE} \
+    --registry-config=/etc/host-containers/host-ctr.toml \
+    --skip-if-image-exists=true
```
bcressey, line 9: Just curious - do we have a path for getting rid of this at some point? It seems weird that `kubelet` can't handle this.

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -5,6 +5,17 @@ authentication-mode = "aws"
...
+[settings.kubernetes.credential-providers.ecr-credential-provider]
+enabled = true
+cache-duration = "12h"
+image-patterns = [
+    "*.dkr.ecr.*.amazonaws.com",
+    "*.dkr.ecr.*.amazonaws.cn",
+    "*.dkr.ecr-fips.*.amazonaws.com",
+    "*.dkr.ecr.us-iso-east-1.c2s.ic.gov",
+    "*.dkr.ecr.us-isob-east-1.sc2s.sgov.gov"
+]
```
bcressey: Do we really want to switch to the credential provider by default for all existing k8s variants? That feels like it's more risk than necessary.

Some alternatives:
* create separate `sources/models/src/aws-k8s-1.27` and `sources/models/src/aws-k8s-1.27-nvidia` trees (as we eventually did for 1.26, so we could add `kubernetes-aws-external-cloud-provider.toml`)
* use the credential provider for aws-k8s-1.26 also

I don't love either but the separate model seems like the safest path.

packages/kubernetes-1.27/kubernetes-1.27.spec
```diff
@@ -0,0 +1,192 @@
...
+cd ../%{cpgorepo}-%{cpgover}
+install -d %{buildroot}%{_cross_libexecdir}/kubernetes/kubelet/plugins
+install -p -m 0755 ecr-credential-provider %{buildroot}%{_cross_libexecdir}/kubernetes/kubelet/plugins/ecr-credential-provider
+mv LICENSE LICENSE.%{cpgorepo}
+
+cd ../%{gorepo}-%{gover}
+%cross_scan_attribution --clarify %{S:1000} go-vendor vendor
+
+cd ../%{cpgorepo}-%{cpgover}
+%cross_scan_attribution --clarify %{S:1000} go-vendor vendor
```
bcressey: `ecs-agent` does this differently so that the `bottlerocket-license-scan` tool only runs once - it would be good to confirm that we end up with all the expected attribution info with this approach.

stmcginnis: It wasn't working quite right, so I've changed tack and separated the `ecr-credential-provider` back out into its own package. So while they are tightly coupled, they come from two different places with two different teams of maintainers and two different roadmaps. So it probably makes sense to have a set of provider packages that mirror how we have a set of kubernetes packages.

packages/kubernetes-1.27/prestart-pull-pause-ctr.conf
stmcginnis: I'm not familiar with it. We should discuss with someone that has more background on the need for it in the first place. If it is not actually needed we should be able to remove it in a follow on patch.

sources/models/shared-defaults/kubernetes-aws.toml
stmcginnis: Separated.

packages/kubernetes-1.27/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+
+[[package.metadata.build-package.external-files]]
+url = "https://distro.eks.amazonaws.com/kubernetes-1-27/releases/1/artifacts/kubernetes/v1.27.1/kubernetes-src.tar.gz"
+sha512 = "bec6b87a1cf974feb041ac684057be42d0a9abd332371bc8a8a7311d7f23e4ecba923dc9f46d4db8a5a2818976ddf7e06baef7b9fd87ff5fda27b6665b956c07"
+
+[[package.metadata.build-package.external-files]]
+url = "https://codeload.github.com/kubernetes/cloud-provider-aws/tar.gz/v1.27.1"
+path = "cloud-provider-aws-1.27.1.tar.gz"
+sha512 = "d7a28f4fb3cb2a1e7ee8d94405e3268608562af0ac509b51c32fcca19353eb68c87b023bd7dae1e84a76d9e856e4951cbc8a2260bab358d1eb492e47caedd29d"
+bundle-modules = [ "go" ]
```
bcressey: This can be dropped now.

bcressey (approved):

packages/kubernetes-1.27/Cargo.toml
stmcginnis: Oops.

yeazelm (approved):
LGTM

PR #3061, opened 2023-04-27
netdog: add primary-interface argument

yeazelm:
**Issue number:** #2449 has a sub task for D-Bus listener, this is part 2 of this change.

Closes #

**Description of changes:**
This command provides a way to get the machine's primary interface without reading the files directly which provides a CLI API to access this value. This will be used in yet to be implemented logic to watch this interface for changes on D-Bus.

**Testing done:**
cargo build, fmt, clippy, and test to ensure everything passes. This command isn't in play unless systemd-networkd is configured. 

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/cli/mod.rs
+++ b/sources/api/netdog/src/cli/mod.rs
@@ -1,6 +1,8 @@
 pub(crate) mod generate_hostname;
 pub(crate) mod generate_net_config;
 pub(crate) mod node_ip;
+#[cfg(net_backend = "systemd-networkd")]
+pub(crate) mod primary_interface;
 pub(crate) mod set_hostname;
 pub(crate) mod write_resolv_conf;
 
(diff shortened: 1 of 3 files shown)


sources/api/netdog/src/cli/mod.rs
```diff
@@ -299,6 +303,9 @@ mod error {
         PrimaryInterfaceAddress {
             source: networkd_status::NetworkDStatusError,
         },
+
+        #[snafu(display("Failed to get primary interface: {}", source))]
+        InterfaceFromString { source: interface_id::Error },
```
zmrow: I don't think this error variant is used anywhere - we can probably remove it.

yeazelm: Good catch, I refactored this error out.

sources/api/netdog/src/cli/primary_interface.rs
```diff
@@ -0,0 +1,13 @@
+use super::{primary_interface_name, print_json, Result};
+use argh::FromArgs;
+
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "primary-interface")]
+/// Return the current IP address
+pub(crate) struct PrimaryInterfaceArgs {}
+
+/// Return the current IP address as JSON (intended for use as a settings generator)
```
bcressey, line 9: nit: this comment seems like a copy/paste artifact, since I don't really see how we could use a settings generator for primary interface.
```suggestion
```

bcressey (approved):

zmrow (approved):
Looks good!

PR #3062, opened 2023-04-27
chore(host-ctr): Update containerd dependency

gazal-k:
**Issue number:**

Related to https://github.com/emicklei/go-restful/issues/489 & https://github.com/containerd/containerd/pull/8271

**Description of changes:**
Updates containerd to the latest 1.6.x release; https://github.com/containerd/containerd/releases/tag/v1.6.20

**Testing done:**
```sh
$ go mod tidy
$ go mod vendor
$ go build ./...
$ ./host-ctr --help
NAME:
   host-ctr - manage host containers
   ...
```

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/go.mod
+++ b/sources/host-ctr/go.mod
@@ -5,7 +5,7 @@ go 1.19
 require (
 	github.com/aws/aws-sdk-go v1.44.208
 	github.com/awslabs/amazon-ecr-containerd-resolver v0.0.0-20230225154133-288898c0f318
-	github.com/containerd/containerd v1.6.19
+	github.com/containerd/containerd v1.6.20
 	github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
 	github.com/pelletier/go-toml v1.9.5
 	github.com/pkg/errors v0.9.1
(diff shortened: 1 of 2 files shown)


sources/host-ctr/go.mod
```diff
@@ -16,16 +16,6 @@ require (
 )
 
 replace (
```
jpmcb: You may want to run `go mod tidy`  to get things cleaned up since we don't need the multiline replace anymore. Go should re-format it for you as:

```
// containerd still uses the v1alpha2 APIs in k8s.io/cri-api.
// These were removed in v0.26 of cri-api resulting in a dependency conflict.
// TODO: Remove this `replace` when containerd uses v1 of the cri-api
replace k8s.io/cri-api => k8s.io/cri-api v0.25.5
```

sources/host-ctr/go.sum
```diff
@@ -359,8 +359,7 @@ github.com/dustin/go-humanize v1.0.0/go.mod h1:HtrtbFcZ19U5GC7JDqmcUSB87Iq5E25Kn
 github.com/elazarl/goproxy v0.0.0-20180725130230-947c36da3153 h1:yUdfgN0XgIJw7foRItutHYUIhlcKzcSf5vDpdhQAKTc=
 github.com/elazarl/goproxy v0.0.0-20180725130230-947c36da3153/go.mod h1:/Zj4wYkgs4iZTTu3o/KG3Itv/qCCa8VVMlb3i9OVuzc=
 github.com/emicklei/go-restful v0.0.0-20170410110728-ff4f55a20633/go.mod h1:otzb+WCGbkyDHkqmQmT5YD2WR4BBwUdeQoFo8l/7tVs=
-github.com/emicklei/go-restful v2.16.0+incompatible h1:rgqiKNjTnFQA6kkhFe16D8epTksy9HQ1MyrbDXSdYhM=
-github.com/emicklei/go-restful v2.16.0+incompatible/go.mod h1:otzb+WCGbkyDHkqmQmT5YD2WR4BBwUdeQoFo8l/7tVs=
+github.com/emicklei/go-restful v2.9.5+incompatible/go.mod h1:otzb+WCGbkyDHkqmQmT5YD2WR4BBwUdeQoFo8l/7tVs=
```
jpmcb, line 362: Hmmm, this line looks like it's adding `github.com/emicklei/go-restful v2.9.5` back into the mix.

And it does look like v2.9.5 of go restful is still being brought in somewhere in containerd:
https://github.com/containerd/containerd/blob/2806fc1057397dbaeefbea0e4e17bddfbd388f38/go.sum#L352-L353

Where and how this is being pulled in on the upstream requires more investigation upstream.

jpmcb (commented):
Thank you so much for this PR!!

sources/host-ctr/go.sum
gazal-k: 😞 it seems to be present even in 1.7.0. https://github.com/containerd/containerd/blob/v1.7.0/go.sum

jpmcb: Hmmm yeah, just noticed that as well: https://github.com/containerd/containerd/issues/8168

We'll want to follow that upstream. But for now, I'll close this and we can keep an eye on what containerd does

jpmcb: @gazal-k - if you want to open a subsequent PR for just the containerd upgrades (without replacing go-restful), I'd be happy to review that

PR #3065, opened 2023-05-01
testsys: Update testsys to v0.0.7

ecpullen:
**Issue number:**

Closes #N/A

**Description of changes:**

Update the TestSys library version to v0.0.7 - https://github.com/bottlerocket-os/bottlerocket-test-system/releases/tag/v0.0.7

**Testing done:**

Tested `cargo make test`.

```bash
Every 2.0s: testsys --kubeconfig /home/ec2-user/bottlerocket/testsys.kubeconfig status                                                                                         Mon May  1 17:59:38 2023

 NAME                                  TYPE                  STATE                             PASSED              FAILED              SKIPPED   BUILD ID              LAST UPDATE
 k-test-3-quick                        Test                  running                                1                   0                 6972   752a994d              2023-05-01T17:59:29Z
 k-test-3                              Resource              completed                                                                           752a994d              2023-04-30T21:02:06Z
 k-test-3-instances-jlro               Resource              completed                                                                           752a994d              2023-05-01T17:58:16Z
```

**Terms of contribution:**

### DIFF
--- a/tools/deny.toml
+++ b/tools/deny.toml
@@ -52,6 +52,13 @@ license-files = [
     { path = "LICENSE", hash = 0x001c7e6c },
 ]
 
+[[licenses.clarify]]
+name = "rustls-webpki"
+expression = "ISC"
+license-files = [
+    { path = "LICENSE", hash = 0x001c7e6c },
+]
+
 [bans]
 # Deny multiple versions or wildcard dependencies.
 multiple-versions = "deny"
@@ -75,6 +82,10 @@ skip-tree = [
     # dependency tree because windows-sys has many sub-crates
     # that differ in major version.
     { name = "windows-sys", version = "=0.42.0" },
+
+    # TestSys uses a newer version of base64 and serde_yaml
+    { name = "testsys-model", version = "=0.0.7" },
+    { name = "bottlerocket-types", version = "=0.0.7" },
 ]
 
 [sources]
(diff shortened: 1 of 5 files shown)


jpmcb (approved):

gthao313 (approved):

tools/deny.toml
```diff
@@ -75,6 +82,10 @@ skip-tree = [
     # dependency tree because windows-sys has many sub-crates
     # that differ in major version.
     { name = "windows-sys", version = "=0.42.0" },
+
+    # TestSys uses a newer version of base64 and serde_yaml
+    { name = "testsys-model", version = "=0.0.7" },
+    { name = "bottlerocket-types", version = "=0.0.7" },
```
bcressey, line 88: Can we `skip-tree` the older versions rather than the newer ones? I find it more useful to consider what's holding us back from upgrading vs. tracking where we're already on new versions.

bcressey (approved):

tools/deny.toml
ecpullen: There are several issues with TestSys 0.0.7 due to our change from openssl to rustls. This includes `base64`, `serde_yaml`, `thiserror`, `json_patch` and possibly others. Is it worth explicitly calling each one out?

bcressey: It seems fair to assume that testsys dependencies will be as current as possible, if we're following the same methodology there, so skipping the tree here makes sense to avoid duplication of effort.

PR #3075, opened 2023-05-04
Add support for ECS exec

### LINKED ISSUE #1649: Add support for ECS Exec
ECS Exec is an important feature of ECS. It allows customers to obtain a shell in the context of a running container (à la `kubectl exec`), without requiring admin access to the host. 

Currently, ECS Exec is supported on the Amazon Linux 2-based ECS optimized AMI and similar variants, but is not supported on Bottlerocket. Given ECS Exec's popularity, it makes sense to add support for it in Bottlerocket as well.

arnaldo2792:
**Issue number:**

Closes #1649 

**Description of changes:**
This includes the changes required to support ECS exec in Bottlerocket.

Under the hood, the ECS agent runs the Amazon SSM agent in each container configured to use EXEC. The ECS agent bind-mounts the statically-linked SSM agent's binaries (`amazon-ssm-agent`, `ssm-agent-worker`, `ssm-session-worker`) onto each container and uses docker to execute them within the container's namespaces.

In the ECS Optimized AMIs, the ECS agent looks at the path `/managed-agents` for the dependencies needed to use ECS exec. This directory is a bind-mount to `/var/lib/ecs/deps/`. The ECS agent checks that the `bin`, `config` and `certs` directories exist to mark the instances as “capable” to run ECS exec. The ECS agent requires the SSM binaries to be in a versioned directory within `bin`, and a CA bundle named `tls-ca-bundle.pem` to be available in `certs`. The agent uses `config` to store configurations created at runtime for each ECS exec session, so it needs to be writable by agent.

In this PR, a patch for the ECS agent updates the path used to check for dependencies from `/managed-agents` to `/usr/libexec/amazon-ecs-agent/managed-agents`. This is because in Bottlerocket the ECS agent runs as a system service instead of a docker container, and `/usr/libexec` is used to store binaries that are used by other programs.

The `config` directory for the SSM sessions is a symbolic link to `/var/ecs/managed-agents/execute-command/config` to allow the ECS agent generate the required configurations at runtime. This directory is created via `tmpfiles.d`. The `certs` directory contains a symbolic link to the CA Bundle of the host. The `bin` directory is a symbolic link to the directory that contains the versioned SSM binaries.

**Testing done:**
- SSM exec tests in the ECS conformance test suite pass.
- I validated the ECS exec sessions don't interfere with the SSM agent in the control host container
- I validated I can exec into a task

```bash
[ssm-user@control]$ apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "c76b9874",
    "pretty_name": "Bottlerocket OS 1.14.0 (aws-ecs-1)",
    "variant_id": "aws-ecs-1",
    "version_id": "1.14.0"
  }
}

#...

~ ❯ aws ecs execute-command --cluster bottlerocket \
    --task <task-arn> --container fedora \
    --interactive --command "cat /etc/os-release" | rg PRETTY_NAME
PRETTY_NAME="Fedora Linux 35 (Container Image)"
```

**Terms of contribution:**

### DIFF
--- a/packages/amazon-ssm-agent/amazon-ssm-agent.spec
+++ b/packages/amazon-ssm-agent/amazon-ssm-agent.spec
@@ -1,0 +1,50 @@
+# Don't generate debug packages because we are compiling without CGO,
+# and the `gc` compiler doesn't append the  the ".note.gnu.build-id" section
+# https://fedoraproject.org/wiki/PackagingDrafts/Go#Build_ID
+%global debug_package %{nil}
+
+%global goproject github.com/aws
+%global gorepo amazon-ssm-agent
+%global goimport %{goproject}/%{gorepo}
+
+Name: %{_cross_os}amazon-ssm-agent
+Version: 3.2.815.0
+Release: 1%{?dist}
+Summary: An agent to enable remote management of EC2 instances
+License: Apache-2.0
+URL: https://github.com/aws/amazon-ssm-agent
+Source0: %{gorepo}-%{version}.tar.gz
+BuildRequires: %{_cross_os}glibc-devel
+
+%description
+%{summary}.
+
+%prep
+%setup -n %{gorepo}-%{version}
+
+%build
+%set_cross_go_flags
+
+# Set CGO_ENABLED=0 to statically link binaries that will be bind-mounted by the ECS agent
+CGO_ENABLED=0 go build ${GOFLAGS} -installsuffix cgo -a -ldflags "-s" -o amazon-ssm-agent \
+  ./core/agent.go ./core/agent_unix.go ./core/agent_parser.go
+CGO_ENABLED=0 go build ${GOFLAGS} -installsuffix cgo -a -ldflags "-s" -o ssm-agent-worker \
+  ./agent/agent.go ./agent/agent_unix.go ./agent/agent_parser.go
+CGO_ENABLED=0 go build ${GOFLAGS} -installsuffix cgo -a -ldflags "-s" -o ssm-session-worker \
+  ./agent/framework/processor/executer/outofproc/sessionworker/main.go
+
+%install
+# Install the SSM agent under 'libexecdir', since it is meant to be used by other programs
+install -d %{buildroot}%{_cross_libexecdir}/amazon-ssm-agent/bin/%{version}
+for b in amazon-ssm-agent ssm-agent-worker ssm-session-worker; do
+  install -D -p -m 0755 ${b} %{buildroot}%{_cross_libexecdir}/amazon-ssm-agent/bin/%{version}
+done
+
+%cross_scan_attribution go-vendor vendor
+
+%files
+%license LICENSE
+%{_cross_attribution_file}
+%{_cross_attribution_vendor_dir}
+%dir %{_cross_libexecdir}/amazon-ssm-agent
+%{_cross_libexecdir}/amazon-ssm-agent
--- a/packages/ecs-agent/0006-execcmd-change-execcmd-directories-for-Bottlerocket.patch
+++ b/packages/ecs-agent/0006-execcmd-change-execcmd-directories-for-Bottlerocket.patch
@@ -1,0 +1,67 @@
+From c9f3e2e695fa0c426c7c9196354c5ac7f138845a Mon Sep 17 00:00:00 2001
+From: Arnaldo Garcia Rincon <agarrcia@amazon.com>
+Date: Wed, 3 May 2023 18:23:40 +0000
+Subject: [PATCH] execcmd: change execcmd directories for Bottlerocket
+
+The ECS agent performs checks on directories that, in normal
+circumstances, are mounted on the ECS agent container.  Since the ECS
+agent runs as a service in Bottlerocket, the paths to the directories
+are different.
+
+Signed-off-by: Arnaldo Garcia Rincon <agarrcia@amazon.com>
+---
+ agent/app/agent_capability_unix.go              | 2 +-
+ agent/engine/execcmd/manager_init_task_linux.go | 4 ++--
+ agent/engine/execcmd/manager_linux.go           | 2 +-
+ 3 files changed, 4 insertions(+), 4 deletions(-)
+
+diff --git a/agent/app/agent_capability_unix.go b/agent/app/agent_capability_unix.go
+index 51b4393..76492c7 100644
+--- a/agent/app/agent_capability_unix.go
++++ b/agent/app/agent_capability_unix.go
+@@ -37,7 +37,7 @@ const (
+ 	SSE41                 = "sse4_1"
+ 	SSE42                 = "sse4_2"
+ 	CpuInfoPath           = "/proc/cpuinfo"
+-	capabilityDepsRootDir = "/managed-agents"
++	capabilityDepsRootDir = "/usr/libexec/amazon-ecs-agent/managed-agents"
+ )
+ 
+ var (
+diff --git a/agent/engine/execcmd/manager_init_task_linux.go b/agent/engine/execcmd/manager_init_task_linux.go
+index 05af158..6117e55 100644
+--- a/agent/engine/execcmd/manager_init_task_linux.go
++++ b/agent/engine/execcmd/manager_init_task_linux.go
+@@ -24,7 +24,7 @@ import (
+ )
+ 
+ const (
+-	ecsAgentExecDepsDir = "/managed-agents/execute-command"
++	ecsAgentExecDepsDir = "/usr/libexec/amazon-ecs-agent/managed-agents/execute-command"
+ 
+ 	// ecsAgentDepsBinDir is the directory where ECS Agent will read versions of SSM agent
+ 	ecsAgentDepsBinDir   = ecsAgentExecDepsDir + "/bin"
+@@ -40,7 +40,7 @@ const (
+ 	ContainerLogDir    = "/var/log/amazon/ssm"
+ 	ECSAgentExecLogDir = "/log/exec"
+ 
+-	HostCertFile            = "/var/lib/ecs/deps/execute-command/certs/tls-ca-bundle.pem"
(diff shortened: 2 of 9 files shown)


arnaldo2792:
Builds will fail until I update the look aside cache

packages/ecs-agent/0006-execcmd-change-execcmd-directories-for-Bottlerocket.patch
```diff
@@ -0,0 +1,67 @@
...
++	ecsAgentExecDepsDir = "/usr/libexec/amazon-ecs-agent/managed-agents/execute-command"
+ 
+ 	// ecsAgentDepsBinDir is the directory where ECS Agent will read versions of SSM agent
+ 	ecsAgentDepsBinDir   = ecsAgentExecDepsDir + "/bin"
+@@ -40,7 +40,7 @@ const (
+ 	ContainerLogDir    = "/var/log/amazon/ssm"
+ 	ECSAgentExecLogDir = "/log/exec"
+ 
+-	HostCertFile            = "/var/lib/ecs/deps/execute-command/certs/tls-ca-bundle.pem"
++	HostCertFile            = "/usr/libexec/amazon-ecs-agent/managed-agents/execute-command/certs/tls-ca-bundle.pem"
```
bcressey, line 49: You are confident this is another bind mount from the host, and not a CA cert bundle that the agent retrieves from elsewhere and manages?

packages/ecs-agent/ecs-agent.spec
```diff
@@ -248,6 +252,27 @@ install -D -p -m 0644 %{S:102} %{buildroot}%{_cross_tmpfilesdir}/ecs.conf
...
+install -d %{buildroot}%{managed_agents}
+
+# Directory for ECS exec artifacts
+%global ecs_exec_dir %{managed_agents}/execute-command
+install -d %{buildroot}%{ecs_exec_dir}
+
+# The ECS agent looks for real versioned directories under bin, symlinks will be
+# ignored. Thus, link the bin directory in the ssm-agent directory which contains
+# the versioned binaries.
+ln -s %{_cross_libexecdir}/amazon-ssm-agent/bin %{buildroot}/%{ecs_exec_dir}/bin
```
bcressey: Can we use a relative symlink here? It is best to avoid full paths in symlinks for packages. If not possible, please add a comment explaining why.

```suggestion
# The ECS agent looks for real versioned directories under bin, symlinks will be
# ignored. Thus, link the bin directory in the ssm-agent directory which contains
# the versioned binaries.
ln -rs %{buildroot}%{_cross_libexecdir}/amazon-ssm-agent/bin %{buildroot}%{ecs_exec_dir}/bin
```

packages/ecs-agent/ecs-agent.spec
```diff
@@ -248,6 +252,27 @@ install -D -p -m 0644 %{S:102} %{buildroot}%{_cross_tmpfilesdir}/ecs.conf
...
 
+# Directory for agents used by the ECS agent, e.g. SSM, Service Connect
+%global managed_agents %{_cross_libexecdir}/amazon-ecs-agent/managed-agents
+install -d %{buildroot}%{managed_agents}
+
+# Directory for ECS exec artifacts
+%global ecs_exec_dir %{managed_agents}/execute-command
+install -d %{buildroot}%{ecs_exec_dir}
+
+# The ECS agent looks for real versioned directories under bin, symlinks will be
```
bcressey, line 263: Does the agent have specific versions it checks for?

packages/ecs-agent/ecs-agent.spec
```diff
@@ -248,6 +252,27 @@ install -D -p -m 0644 %{S:102} %{buildroot}%{_cross_tmpfilesdir}/ecs.conf
...
+ln -s %{_cross_libexecdir}/amazon-ssm-agent/bin %{buildroot}/%{ecs_exec_dir}/bin
+
+# The ECS agent generates and stores configurations for ECS exec sessions inside
+# "config", thus reference it with a symlink to a directory under /var
+ln -s %{_cross_localstatedir}/ecs/managed-agents/execute-command/config %{buildroot}%{ecs_exec_dir}/config
+
+# Use the host's certificates bundle for ECS exec sessions
+install -d %{buildroot}%{ecs_exec_dir}/certs
+ln -s %{_cross_sysconfdir}/pki/tls/certs/ca-bundle.crt %{buildroot}%{ecs_exec_dir}/certs/tls-ca-bundle.pem
+
```
bcressey: nit: relative symlinks here too if possible
```suggestion
# The ECS agent generates and stores configurations for ECS exec sessions inside
# "config", thus reference it with a symlink to a directory under /var
ln -rs %{buildroot}%{_cross_localstatedir}/ecs/managed-agents/execute-command/config %{buildroot}%{ecs_exec_dir}/config

# Use the host's certificates bundle for ECS exec sessions
install -d %{buildroot}%{ecs_exec_dir}/certs
ln -rs %{buildroot}%{_cross_sysconfdir}/pki/tls/certs/ca-bundle.crt %{buildroot}%{ecs_exec_dir}/certs/tls-ca-bundle.pem

```

packages/ecs-agent/ecs-tmpfiles.conf
```diff
@@ -1,2 +1,4 @@
 d /var/lib/ecs/data 0700 root root
 d /var/log/ecs 0755 root root
+d /var/log/ecs/exec 0755 root root -
+d /var/ecs/managed-agents/execute-command/config 0750 root root -
```
bcressey: If we don't need this state preserved across updates, it might be better to clean up the directory. That way, if we're ever downgrading into this version from a newer version, we won't confuse the ecs-agent with data that it's not able to handle.

```suggestion
d /var/log/ecs/exec 0755 root root -
R /var/ecs/managed-agents - - - -
d /var/ecs/managed-agents/execute-command/config 0750 root root -
```

packages/ecs-agent/ecs-agent.spec
```diff
@@ -289,6 +314,8 @@ mv %{vpccni_gorepo}-%{vpccni_gitrev}/vendor go-vendor/%{vpccni_gorepo}
 %{_cross_libexecdir}/amazon-ecs-agent/ecs-eni
 %{_cross_libexecdir}/amazon-ecs-agent/ecs-ipam
 %{_cross_libexecdir}/amazon-ecs-agent/vpc-branch-eni
+%{managed_agents}
+%{ecs_exec_dir}
```
bcressey: I would prefer using the standard macros here:
```suggestion
%{_cross_libexecdir}/amazon-ecs-agent/managed-agents
```

I'd expect rpm to emit a "file listed twice" warning here because it seems like these macros would expand to:
```
%{_cross_libexecdir}/amazon-ecs-agent/managed-agents
%{_cross_libexecdir}/amazon-ecs-agent/managed-agents/execute-command
```

packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -0,0 +1,49 @@
...
+  install -D -p -m 0755 ${b} %{buildroot}%{_cross_libexecdir}/amazon-ssm-agent/bin/%{version}
+done
+
+%cross_scan_attribution go-vendor vendor
+
+%files
+%license LICENSE
+%{_cross_attribution_file}
+%{_cross_attribution_vendor_dir}
+%{_cross_libexecdir}/amazon-ssm-agent/
```
bcressey: nit: I like to mark directories specifically, because file-to-directory and directory-to-file transitions are quite painful in normal RPM usage:
```suggestion
%dir %{_cross_libexecdir}/amazon-ssm-agent
%{_cross_libexecdir}/amazon-ssm-agent
```

packages/ecs-agent/ecs-agent.spec
arnaldo2792: No, it lists all the versions available in the directory, and uses the latest. I guess in AL*, there could be many agents installed at once.

packages/ecs-agent/0006-execcmd-change-execcmd-directories-for-Bottlerocket.patch
arnaldo2792: Yes, this is another bind mount:

```bash
ecs-exec on  ecs-exec [$] ❯ aws ecs execute-command \
    --cluster bottlerocket --task <task-id> --container fedora \
    --interactive --command "cat /proc/self/mountinfo" | rg tls
466 439 0:30 /pki/tls/certs/ca-bundle.crt /ecs-execute-command-384f4d6e-39e3-41ee-9e00-609338445345/certs/amazon-ssm-agent.crt ro,noatime - tmpfs tmpfs rw,context=system_u:object_r:etc_t:s0,mode=755

```

The agent does not manage the file. Even in the ECS anywhere script, the host's CA bundle is copied to this location:

https://github.com/aws/amazon-ecs-init/blob/master/scripts/ecs-anywhere-install.sh#L595

arnaldo2792:
(Forced push addresses comments above)

packages/ecs-agent/0006-execcmd-change-execcmd-directories-for-Bottlerocket.patch
bcressey: Can the patch just point the agent to `/etc/pki/tls/certs/ca-bundle.crt` then? Then we wouldn't need a symlink.

bcressey (approved):

packages/ecs-agent/0006-execcmd-change-execcmd-directories-for-Bottlerocket.patch
arnaldo2792: We could do this, however the patch will be bigger since there are other places in the agent that should be updated to reference this file, e.g.

https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/execcmd/manager_init_task_linux.go#L140

zmrow (approved):
LGTM!

yeazelm (approved):
LGTM! 🚢

PR #3089, opened 2023-05-06
start-local-vm: add "--product-name" argument to support BUILDSYS_NAME

jfbette:
**Issue:**
The script start-local-vm don't find the image when the image is built with a custom BUILDSYS_NAME.

```
# cargo make -e BUILDSYS_VARIANT=metal-dev -e BUILDSYS_NAME=my-bottlerocket
# ./tools/start-local-vm --arch x86_64 --inject-file net.toml --inject-file user-data.toml --variant metal-dev
Boot image not found. Did the last build fail?
```

In the `start-local-vm` the bottlerocket image name prefix is hard coded.

**Description of changes:**
Add an optionnal argument `--product-name` to the script `start-local-vm`

**Testing done:**
Verified the VM launch without  `--product-name` when the image is built without custom BUILDSYS_NAME
```
# cargo make -e BUILDSYS_VARIANT=metal-dev
# ./tools/start-local-vm --arch x86_64 --inject-file net.toml --inject-file user-data.toml --variant metal-dev 

```
--> the VM start

Verified the VM launch with argument `--product-name` supplied when the image is built custom BUILDSYS_NAME
```
# cargo make -e BUILDSYS_VARIANT=metal-dev -e BUILDSYS_NAME=my-bottlerocket
# ./tools/start-local-vm --arch x86_64 --inject-file net.toml --inject-file user-data.toml --variant metal-dev --product-name my-bottlerocket

```
--> the VM start

Terms of contribution:

### DIFF
--- a/tools/start-local-vm
+++ b/tools/start-local-vm
@@ -5,6 +5,7 @@ shopt -s nullglob
 
 arch=${BUILDSYS_ARCH}
 variant=${BUILDSYS_VARIANT}
+product_name=${BUILDSYS_NAME:-bottlerocket}
 host_port_forwards=tcp::2222-:22
 vm_mem=4G
 vm_cpus=4
@@ -41,6 +42,10 @@ Options:
                         BUILDSYS_ARCH environment variable is set)
     --variant           Bottlerocket variant to run (may be omitted if the
                         BUILDSYS_VARIANT environment variable is set)
+    --product-name
+                        product name used for file and directory naming used when
+                        building with the "-e BUILDSYS_NAME" option; may be omitted if the
+                        BUILDSYS_NAME environment variable is set. Otherwise default is bottlerocket if not defined or empty
     --host-port-forwards
                         list of host ports to forward to the VM; HOST_PORT_FWDS
                         must be a valid QEMU port forwarding specifier (default
@@ -92,6 +97,8 @@ parse_args() {
                 shift; arch=$1 ;;
             --variant)
                 shift; variant=$1 ;;
+            --product-name)
+                shift; product_name=$1 ;;
             --host-port-forwards)
                 shift; host_port_forwards=$1 ;;
             --vm-memory)
@@ -141,8 +148,8 @@ extract_image() {
 
 prepare_raw_images() {
     local -r image_dir=build/images/${arch}-${variant}/latest
-    local -r compressed_boot_image=${image_dir}/bottlerocket-${variant}-${arch}.img.lz4
-    local -r compressed_data_image=${image_dir}/bottlerocket-${variant}-${arch}-data.img.lz4
+    local -r compressed_boot_image=${image_dir}/${product_name}-${variant}-${arch}.img.lz4
+    local -r compressed_data_image=${image_dir}/${product_name}-${variant}-${arch}-data.img.lz4
 
     if [[ -e ${compressed_boot_image} ]]; then
         readonly boot_image=${compressed_boot_image%*.lz4}
--- a/tools/start-local-vm
+++ b/tools/start-local-vm
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 # shellcheck disable=SC2054  # Arrays are formatted for passing args to other tools
+
 shopt -s nullglob
 
 arch=${BUILDSYS_ARCH}
 variant=${BUILDSYS_VARIANT}
-product_name=bottlerocket
 host_port_forwards=tcp::2222-:22
 vm_mem=4G
 vm_cpus=4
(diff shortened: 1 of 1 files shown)


jpmcb (commented):
Hi @jfbette - thanks for the first time contribution!! 🎉 

Changes look safe to me - approving workflows to run. Changes overall look good. And thanks for providing testing!

yeazelm (approved):
Looks good! Thanks for the PR!

zmrow:
Thanks for the contribution!  This is a great addition.

I'm thinking `--image-name` or `--image-prefix` might be a bit more accurate in this context.  What do you think?  We're also happy to make the change if you're busy.

stmcginnis:
> Thanks for the contribution! This is a great addition.
> 
> I'm thinking `--image-name` or `--image-prefix` might be a bit more accurate in this context. What do you think? We're also happy to make the change if you're busy.

+1 since this isn't technically a "product" :)

tools/start-local-vm
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 # shellcheck disable=SC2054  # Arrays are formatted for passing args to other tools
-
 shopt -s nullglob
 
 arch=${BUILDSYS_ARCH}
 variant=${BUILDSYS_VARIANT}
+product_name=bottlerocket
```
markusboehme: I suggest initializing with `${BUILDSYS_NAME-bottlerocket}` so in case a buildsys environment variable is already set it takes precedence over the default, similar to how `BUILDSYS_ARCH` and `BUILDSYS_VARIANT` from the environment are considered.

markusboehme:
Good catch, thanks!

In case you push another revision to change the option name, please also modify the commit message slightly (e.g. `start-local-vm: add "--product-name" argument to support BUILDSYS_NAME`). The prefix helps later when skimming the commit log and deciding whether to pay more attention to any individual commit.

tools/start-local-vm
jfbette: Thanks guys for all your comments.

I would rather initialize with `${BUILDSYS_NAME}` because the objective is to allow creating unbranded builds (still compliant with the bottlerocket trademark?) . Moreover it will be consistant with the `Makefile.toml` where their are no constraints on `BUILDSYS_NAME` (no prefix, no suffix, ...).

> > Thanks for the contribution! This is a great addition.
> > I'm thinking `--image-name` or `--image-prefix` might be a bit more accurate in this context. What do you think? We're also happy to make the change if you're busy.
> 
> +1 since this isn't technically a "product" :)

Yes I hesitated also! But in when I saw the comment on `BUILDSYS_NAME` in `Makefile.toml`, I aligned it with the comment ;) : `# Product name used for file and directory naming`.
If I think again that name is to implement the trademark rather a product that an image, no? So I would say stay with that name. We can said that with the script we launch a variant from a product line.

I will push again when all aligned.

markusboehme: > I would rather initialize with `${BUILDSYS_NAME}` because the objective is to allow creating unbranded builds (still compliant with the bottlerocket trademark?) . Moreover it will be consistant with the `Makefile.toml` where their are no constraints on `BUILDSYS_NAME` (no prefix, no suffix, ...).

Sorry for not being clearer. This does not add a mandatory "bottlerocket" suffix, but is a form of parameter expansion that provides a default in case a variable is unset (see [this section of POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02)). For example,

```
$ echo ${BUILDSYS_NAME}

$ echo ${BUILDSYS_NAME-bottlerocket}
bottlerocket
$ BUILDSYS_NAME=foobar
$ echo ${BUILDSYS_NAME-bottlerocket}
foobar
```

Hence, if `BUILDSYS_NAME` is already set in the environment, it will transparently be used as `--product-name`/`--image-prefix` (I don't have a strong opinion on either), similar to how that is happening in the various `cargo make` tasks.

jfbette:
Oh yes got it, read to fast...my bad. I would use use the more safe `${BUILDSYS_NAME:-bottlerocket}` in case it is defined but empty, it could lead to strange name (beginning with a dash)

tools/start-local-vm
jfbette: Oh yes got it, read to fast...my bad. I would use use the more safe ${BUILDSYS_NAME:-bottlerocket} in case it is defined but empty, it could lead to strange name (beginning with a dash)

markusboehme (approved):

stmcginnis (approved):
Looks good, thanks!

Issue #3087, opened 2023-05-05
Kernel panic - not syncing: hung_task: blocked tasks


stbraswell:
Image I'm using: Bottlerocket OS 1.13.5 (aws-k8s-1.24)

What I expected to happen: No kernel panics and node not to reboot

What actually happened:
I upgraded the cluster to Bottlerocket 1.13.5 on Tuesday, it seemed to be running fine until Friday morning.  I noticed pods stuck in containerCreating and even unknown status.  The pods in question were all scheduled on this node.  The node cycled through about 3 restarts, each about 10 minutes apart.  Afterwards we cordoned and drained the node. 

<pre>
[228678.482741] INFO: task containerd-shim:931118 blocked for more than 122 seconds.
[228678.485778]       Not tainted 5.15.102 #1
[228678.487450] "echo 0 > /proc/sys/kernel/hung_task_timeout_secs" disables this message.
[228678.493423] task:containerd-shim state:D stack:    0 pid:931118 ppid:     1 flags:0x00004002
[228678.499631] Call Trace:
[228678.502142]  <TASK>
[228678.504570]  __schedule+0x1fb/0x520
[228678.507451]  schedule+0x5b/0xd0
[228678.510238]  wb_wait_for_completion+0x51/0x80
[228678.513463]  ? do_wait_intr_irq+0xb0/0xb0
[228678.516564]  __writeback_inodes_sb_nr+0xa2/0xd0
[228678.519846]  sync_filesystem+0x2d/0x90
[228678.522849]  ovl_sync_fs+0x59/0x90 [overlay]
[228678.526035]  sync_filesystem+0x77/0x90
[228678.529027]  generic_shutdown_super+0x22/0x110
[228678.532268]  kill_anon_super+0x14/0x30
[228678.535264]  deactivate_locked_super+0x33/0xa0
[228678.538518]  cleanup_mnt+0x131/0x190
[228678.541439]  task_work_run+0x5c/0x90
[228678.544407]  exit_to_user_mode_loop+0x115/0x120
[228678.547648]  exit_to_user_mode_prepare+0x76/0x80
[228678.550957]  syscall_exit_to_user_mode+0x1d/0x40
[228678.554300]  do_syscall_64+0x48/0x90
[228678.557214]  entry_SYSCALL_64_after_hwframe+0x61/0xcb
[228678.560691] RIP: 0033:0x5612de3250ce
[...]

PR #3091, opened 2023-05-08
kernel-5.15: Default disable panic on hung task and lockups

foersleo:
**Issue number:** #3087 

**Description of changes:**

Change the default for panic on hung tasks and lockups from default on to default off.
This is intended to be a discussion starter on if we want to change the default or not.

```
Setting panic as defaults for hung tasks and lockups is probably not the best
idea for the general case. In scenarios with remote storage we may see hangs
as interactions with that storage might be delayed due to network connectivity
issues. In such cases there is use cases that do not benefit from panic and reboot.

Disable the default for those panic options. This does not take away the option to
configure panic on hung tasks or lockups through either sysctl or kernel boot parameters.

As a bonus this aligns both kernels 5.10 and 5.15 on coherent settings for the panic choices.
```

**Testing done:**

Checked we are only changing the appropriate options:

```
config-aarch64-aws-dev-diff:      0 removed,   0 added,   4 changed
config-x86_64-aws-dev-diff:       0 removed,   0 added,   6 changed
```

```
==> configs_panic_disable/config-aarch64-aws-dev-diff <==
 BOOTPARAM_HUNG_TASK_PANIC y -> n
 BOOTPARAM_HUNG_TASK_PANIC_VALUE 1 -> 0
 BOOTPARAM_SOFTLOCKUP_PANIC y -> n
 BOOTPARAM_SOFTLOCKUP_PANIC_VALUE 1 -> 0

==> configs_panic_disable/config-x86_64-aws-dev-diff <==
 BOOTPARAM_HARDLOCKUP_PANIC y -> n
 BOOTPARAM_HARDLOCKUP_PANIC_VALUE 1 -> 0
 BOOTPARAM_HUNG_TASK_PANIC y -> n
 BOOTPARAM_HUNG_TASK_PANIC_VALUE 1 -> 0
 BOOTPARAM_SOFTLOCKUP_PANIC y -> n
 BOOTPARAM_SOFTLOCKUP_PANIC_VALUE 1 -> 0
```

No further testing done as these config options merely change the default value of a variable,
which may or may not be overridden at runtime or system boot through sysctl or kernel boot parameters.

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.15/config-bottlerocket
+++ b/packages/kernel-5.15/config-bottlerocket
@@ -195,3 +195,8 @@ CONFIG_SCSI_ISCSI_ATTRS=m
 # target side
 CONFIG_ISCSI_TARGET=m
 # CONFIG_INFINIBAND_ISERT is not set
+
+# Disable panic on hung and lockup conditions by default
+# CONFIG_BOOTPARAM_SOFTLOCKUP_PANIC is not set
+# CONFIG_BOOTPARAM_HARDLOCKUP_PANIC is not set
+# CONFIG_BOOTPARAM_HUNG_TASK_PANIC is not set


stmcginnis (approved):
Looks good, and I like the consistency with 5.10.

bcressey (approved):

yeazelm (approved):

markusboehme (approved):

PR #3096, opened 2023-05-09
Actions: Only use GOPROXY=direct on nightly build

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This adds a new workflow to run nightly that will use the `GOPROXY=direct` setting to verify that the caching of Go modules is not hiding an issue with upstream module availability changes.

It trims down the normal PR workflow to not perform the additional checks like unit-tests and shellchecks.

With the addition of the nightly workflow to identify these issues, this restores the test coverage of the GOPROXY setting from the `build` workflow that runs on each PR. This should reduce the number of failed jobs that need to be rerun due to throttling of go module fetches that seems to happen very often during normal working hours.

**Testing done:**

Will watch GitHub Action results and adjust as needed.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,64 @@
+# This is basically a duplicate of the main "build" workflow, but uses GOPROXY=direct
+# to try to catch errors close to their introduction due to yanked Go modules. These
+# could otherwise be covered up by caching and not discovered until much later when
+# bypassing the main cache.
+name: Nightly
+on:
+  schedule:
+    # Run once a day at 02:15 UTC. Randomly chosen as a "quiet" time for this to run.
+    # See syntax for format details: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule
+    - cron: '15 2 * * *'
+
+env:
+  # When Go packages are built, buildsys will vendor in dependent Go code for
+  # that package and bundle it up in a tarball. This env variable is consumed
+  # and used to configure Go to directly download code from its upstream source.
+  # This is a useful early signal during GitHub actions to see if there are
+  # upstream Go code problems.
+  GOPROXY: direct
+
+jobs:
+  list-variants:
+    # This needs to be its own job since the build job needs its output before
+    # it can initialize
+    if: github.repository == 'bottlerocket-os/bottlerocket'
+    name: "Determine variants"
+    runs-on: ubuntu-latest
+    outputs:
+      variants: ${{ steps.get-variants.outputs.variants }}
+      aarch-enemies: ${{ steps.get-variants.outputs.aarch-enemies }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/list-variants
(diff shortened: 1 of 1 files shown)


jpmcb (approved):

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,190 @@
...
+  # When Go packages are built, buildsys will vendor in dependent Go code for
+  # that package and bundle it up in a tarball. This env variable is consumed
+  # and used to configure Go to directly download code from its upstream source.
+  # This is a useful early signal during GitHub actions to see if there are
+  # upstream Go code problems.
+  GOPROXY: direct
+
+jobs:
+  build:
+    runs-on:
```
bcressey:
```suggestion
  build:
    if: github.repository == 'bottlerocket-os/bottlerocket'
    runs-on:
```

bcressey (commented):
Rather than do this nightly, can we merge it into `cache.yaml` so it happens on merge to develop? Right now this workflow is a hybrid of `build.yaml` and `cache.yaml` which seems like it could be annoying to maintain and keep in sync.

I'm also wondering if some sort of `BUILDSYS_DOWNLOAD_ONLY` flag would help skip the expensive per-variant builds and just exercise the part we care about; then we could loop through all packages rather than all variants.

stmcginnis:
`cache.yaml` doesn't seem right, and has some other issues with the things being done there.

I'm going to play around with [action composition](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) to see if there is a good way to reduce duplication, but still be able to cover everything we want. That might be awhile though, so if anyone else feels like tackling this, please feel free.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,64 @@
+# This is basically a duplicate of the main "build" workflow, but uses GOPROXY=direct
+# to try to catch errors close to their introduction due to yanked Go modules. These
+# could otherwise be covered up by caching and not discovered until much later when
+# bypassing the main cache.
+name: Nightly
+on:
+  schedule:
+    - cron: '15 2 * * *'
```
etungsten, line 10: A comment describing the schedule might help with parsing this for people unfamiliar with cron schedules.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,64 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/list-variants
+        id: get-variants
+
+  build:
+    needs: list-variants
+    runs-on:
+      group: bottlerocket
+      labels: bottlerocket_ubuntu-latest_32-core
```
etungsten: To save costs, maybe we can just use the free runners for this since result promptness is not as important

stmcginnis: Good call!

etungsten (approved):

jpculp (approved):
This makes sense to me as a distinct action.

bcressey (approved):

PR #3097, opened 2023-05-10
add support for Secure Boot

bcressey:
**Issue number:**

#2501

**Description of changes:**
The first set of commits brings GRUB up-to-date to fix any known flaws in Secure Boot support, adds shim as the initial EFI bootloader, and prepares the shim and grub binaries for modification in the final image build step.

```
packages: update grub
grub: move embedded public key into dedicated section
grub: add SBAT info
grub: add modules for GPG signature verification
tools: add script to replace the public key embedded in GRUB
packages: build shim
shim: embed placeholder for vendor cert
```

The second set of commits introduces the concept of a Secure Boot profile - all of the private keys, certificates, and configuration files necessary for Secure Boot signing - and automates the creation of a local profile. It also adds an image feature flag for Secure Boot support, along with OVF templates for VMware that include placeholders for EFI variables. 

```
build: add scripts to create Secure Boot profiles
build: auto-create a local Secure Boot profile
build: pass Secure Boot profile secrets to builds
build: add Secure Boot OVF templates
build: add image feature flag for Secure Boot
```

The third set of commits extends `buildsys` (mostly `rpm2img`) and `pubsys` to do the required signing and registration for images, OVAs, and AMIs. This is where all of the above actually pays off.

```
build: sign artifacts for Secure Boot
pubsys: register AMIs with Secure Boot support
```

The fourth set of commits is developer-oriented. It enables the feature flag for `*-dev` variants, documents the new tools for generating Secure Boot profiles, and extends `start-local-vm` to support firmware overrides for testing the functionality.

```
start-local-vm: add options to override firmware
docs: document tools to generate Secure Boot Keys
variants: enable Secure Boot for dev builds
```

**Testing done:**
Verified that Secure Boot enabled images worked on:
* AWS (for Nitro-based instance types)
* VMware (in VMC-A)
* metal (under QEMU)

Verified that older instance types could be launched from the same AMIs, and fell back to BIOS boot as expected.

Performed negative testing as well with Secure Boot enabled:
* firmware refuses to boot shim if not signed with a trusted key
* shim refuses to boot grub if not signed with a trusted key
* grub refuses to load grub.cfg if the detached signature is invalid or from an untrusted key
* grub refuses to load the Linux kernel if not signed with a trusted key

**Terms of contribution:**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 /*.pem
 /keys
 /roles
+/sbkeys/**/
 /Licenses.toml
 /licenses
 *.run
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,6 +67,7 @@ ARG VARIANT_FAMILY
 ARG VARIANT_FLAVOR
 ARG REPO
 ARG GRUB_SET_PRIVATE_VAR
+ARG UEFI_SECURE_BOOT
 ARG SYSTEMD_NETWORKD
 ARG UNIFIED_CGROUP_HIERARCHY
 ARG XFS_DATA_PARTITION
@@ -101,6 +102,7 @@ RUN rpmdev-setuptree \
    && echo "%bcond_without $(V=${VARIANT_FAMILY,,}; echo ${V//-/_})_family" >> .bconds \
    && echo "%bcond_without $(V=${VARIANT_FLAVOR:-no}; V=${V,,}; echo ${V//-/_})_flavor" >> .bconds \
    && echo -e -n "${GRUB_SET_PRIVATE_VAR:+%bcond_without grub_set_private_var\n}" >> .bconds \
+   && echo -e -n "${UEFI_SECURE_BOOT:+%bcond_without uefi_secure_boot\n}" >> .bconds \
    && echo -e -n "${SYSTEMD_NETWORKD:+%bcond_without systemd_networkd\n}" >> .bconds \
    && echo -e -n "${UNIFIED_CGROUP_HIERARCHY:+%bcond_without unified_cgroup_hierarchy\n}" >> .bconds \
    && echo -e -n "${XFS_DATA_PARTITION:+%bcond_without xfs_data_partition\n}" >> .bconds \
@@ -200,6 +202,7 @@ ARG DATA_IMAGE_PUBLISH_SIZE_GIB
 ARG KERNEL_PARAMETERS
 ARG GRUB_SET_PRIVATE_VAR
 ARG XFS_DATA_PARTITION
+ARG UEFI_SECURE_BOOT
 ENV VARIANT=${VARIANT} VERSION_ID=${VERSION_ID} BUILD_ID=${BUILD_ID} \
     PRETTY_NAME=${PRETTY_NAME} IMAGE_NAME=${IMAGE_NAME} \
     KERNEL_PARAMETERS=${KERNEL_PARAMETERS}
@@ -219,6 +222,7 @@ RUN --mount=target=/host \
       --ovf-template="/host/variants/${VARIANT}/template.ovf" \
       ${XFS_DATA_PARTITION:+--xfs-data-partition=yes} \
       ${GRUB_SET_PRIVATE_VAR:+--with-grub-set-private-var=yes} \
+      ${UEFI_SECURE_BOOT:+--with-uefi-secure-boot=yes} \
     && echo ${NOCACHE}
 
 # =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^= =^..^=
--- a/Dockerfile
+++ b/Dockerfile
@@ -210,6 +210,19 @@ WORKDIR /root
 
 USER root
 RUN --mount=target=/host \
+    --mount=type=secret,id=PK.crt,target=/root/sbkeys/PK.crt \
+    --mount=type=secret,id=KEK.crt,target=/root/sbkeys/KEK.crt \
+    --mount=type=secret,id=db.crt,target=/root/sbkeys/db.crt \
+    --mount=type=secret,id=vendor.crt,target=/root/sbkeys/vendor.crt \
+    --mount=type=secret,id=shim-sign.key,target=/root/sbkeys/shim-sign.key \
+    --mount=type=secret,id=shim-sign.crt,target=/root/sbkeys/shim-sign.crt \
+    --mount=type=secret,id=code-sign.key,target=/root/sbkeys/code-sign.key \
+    --mount=type=secret,id=code-sign.crt,target=/root/sbkeys/code-sign.crt \
+    --mount=type=secret,id=config-sign.key,target=/root/sbkeys/config-sign.key \
+    --mount=type=secret,id=kms-sign.json,target=/root/.config/aws-kms-pkcs11/config.json \
+    --mount=type=secret,id=aws-access-key-id.env,target=/root/.aws/aws-access-key-id.env \
+    --mount=type=secret,id=aws-secret-access-key.env,target=/root/.aws/aws-secret-access-key.env \
+    --mount=type=secret,id=aws-session-token.env,target=/root/.aws/aws-session-token.env \
     /host/tools/rpm2img \
       --package-dir=/local/rpms \
       --output-dir=/local/output \
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -843,7 +843,7 @@ cargo build \
 ]
 
 [tasks.build-variant]
-dependencies = ["fetch-sdk", "build-tools", "publish-setup"]
+dependencies = ["fetch-sdk", "build-tools", "build-sbkeys", "publish-setup"]
 script = [
 '''
 export PATH="${BUILDSYS_TOOLS_DIR}/bin:${PATH}"
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -1228,6 +1228,7 @@ pubsys \
    "${data_volume_args[@]}" \
    \
    --variant-manifest "${BUILDSYS_ROOT_DIR}/variants/${BUILDSYS_VARIANT}/Cargo.toml" \
+   --uefi-data "${BUILDSYS_SBKEYS_PROFILE_DIR}/efi-vars.aws" \
    --arch "${BUILDSYS_ARCH}" \
    --name "${ami_name}" \
    --description "${PUBLISH_AMI_DESCRIPTION:-${ami_name}}" \
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -10,6 +10,8 @@ BUILDSYS_STATE_DIR = "${BUILDSYS_BUILD_DIR}/state"
 BUILDSYS_IMAGES_DIR = "${BUILDSYS_BUILD_DIR}/images"
 BUILDSYS_TOOLS_DIR = "${BUILDSYS_ROOT_DIR}/tools"
 BUILDSYS_SOURCES_DIR = "${BUILDSYS_ROOT_DIR}/sources"
+BUILDSYS_SBKEYS_DIR = "${BUILDSYS_ROOT_DIR}/sbkeys"
+BUILDSYS_SBKEYS_PROFILE = { script = ['echo "${BUILDSYS_SBKEYS_PROFILE:-local}"'] }
 BUILDSYS_TIMESTAMP = { script = ["date +%s"] }
 BUILDSYS_VERSION_BUILD = { script = ["git describe --always --dirty --exclude '*' || echo 00000000"] }
 # For now, release config path can't be overridden with -e, because it's used
@@ -156,6 +158,9 @@ BUILDSYS_NAME_FRIENDLY = "${BUILDSYS_NAME_VARIANT}-v${BUILDSYS_VERSION_IMAGE}"
 
 # For variant build artifacts.
 BUILDSYS_VARIANT_DIR = "${BUILDSYS_OUTPUT_DIR}/${BUILDSYS_VERSION_FULL}"
+
+# Depends on ${BUILDSYS_SBKEYS_DIR} and ${BUILDSYS_SBKEYS_PROFILE}.
+BUILDSYS_SBKEYS_PROFILE_DIR = "${BUILDSYS_SBKEYS_DIR}/${BUILDSYS_SBKEYS_PROFILE}"
 
 # Path to repo-specific root role.
 PUBLISH_REPO_ROOT_JSON = "${BUILDSYS_ROOT_DIR}/roles/${PUBLISH_REPO}.root.json"
@@ -677,6 +682,52 @@ cargo install \
   --root tools \
   --force \
   --quiet
+'''
+]
+
+[tasks.build-sbkeys]
+dependencies = ["fetch"]
+script_runner = "bash"
+script = [
+'''
+# Check the profile for all files needed for Secure Boot signing.
+profile="${BUILDSYS_SBKEYS_PROFILE_DIR}"
+
+found=0
+# A local profile has signing keys and certificates, while an AWS profile
+# has a config for the aws-kms-pkcs11 helper. Either type is supported.
+if [ -s "${profile}/shim-sign.key" ] && \
+   [ -s "${profile}/shim-sign.crt" ] && \
+   [ -s "${profile}/code-sign.key" ] && \
+   [ -s "${profile}/code-sign.crt" ] ; then
+   let found+=1
+elif [ -s "${profile}/kms-sign.json" ] ; then
+   let found+=1
+fi
+
+expected=1
+for f in {PK,KEK,db,vendor}.crt config-sign.key efi-vars.{json,aws} ; do
+  let expected+=1
+  [ -s "${profile}/${f}" ] && let found+=1
+done
+
+if [ "${found}" -eq "${expected}" ] ; then
+  exit 0
+fi
+
+if [ "${found}" -gt 0 ] ; then
+  echo "Incomplete Secure Boot signing profile found in ${profile}." >&2
+  echo "Missing $(( expected - found )) files." >&2
+  exit 1
+fi
+
+echo "No Secure Boot signing profile found in ${profile}." >&2
+echo "Generating local keys." >&2
+
+mkdir -p "${BUILDSYS_SBKEYS_PROFILE_DIR}"
+${BUILDSYS_SBKEYS_DIR}/generate-local-sbkeys \
+  --sdk-image "${BUILDSYS_SDK_IMAGE}" \
+  --output-dir "${BUILDSYS_SBKEYS_PROFILE_DIR}"
 '''
 ]
 
--- a/PROVISIONING-METAL.md
+++ b/PROVISIONING-METAL.md
@@ -333,3 +333,20 @@ docker run --rm \
    "${SDK_IMAGE}" \
    bootconfig -l /tmp/bootconfig.data
 ```
+
+### Enable Secure Boot
+
+Starting with metal-k8s-1.28, the Bottlerocket images for bare metal support Secure Boot when used on a platform with UEFI firmware.
+UEFI boot mode must be used, rather than legacy BIOS boot mode, and Secure Boot must be enabled.
+The UEFI firmware may provide a Compatibility Support Module (CSM) option to enable legacy BIOS emulation.
+The CSM option must not be enabled.
+These options can be set in the firmware setup menu, which can be accessed during boot by pressing a certain key (such as F2 or F12).
+
+Many Linux distros ship a copy of the [shim](https://github.com/rhboot/shim) bootloader signed by Microsoft with a key that is trusted by default.
+Although Bottlerocket also uses `shim`, its copy is not signed by Microsoft and will not be trusted without additional configuration.
+After installing Bottlerocket, the appropriate vendor certificate can be found on the EFI System Partition (ESP).
+The firmware setup menu should provide an option to import a new vendor certificate by selecting a file on the ESP.
+Either the PEM format (`db.crt`) or DER format (`db.cer`) certificate can be imported, depending on what the firmware supports.
+
+The firmware setup menu should be password-protected to prevent unauthorized changes to the Secure Boot configuration.
+Please refer to the documentation from your hardware vendor for more information on this procedure.
--- a/SECURITY_FEATURES.md
+++ b/SECURITY_FEATURES.md
@@ -13,14 +13,16 @@ We will add and enhance security features over time based on these goals.
 
 ## Overview
 
-| Feature                                                                           | Version |
-| :-------------------------------------------------------------------------------- | :-----: |
-| [Automated security updates](#automated-security-updates)                         |  0.3.0  |
-| [Immutable rootfs backed by dm-verity](#immutable-rootfs-backed-by-dm-verity)     |  0.3.0  |
-| [Stateless tmpfs for /etc](#stateless-tmpfs-for-etc)                              |  0.3.0  |
-| [No shell or interpreters installed](#no-shell-or-interpreters-installed)         |  0.3.0  |
-| [Executables built with hardening flags](#executables-built-with-hardening-flags) |  0.3.0  |
-| [SELinux enabled in enforcing mode](#selinux-enabled-in-enforcing-mode)           |  0.3.0  |
+| Feature                                                                           | Version  |
+| :-------------------------------------------------------------------------------- | :------: |
+| [Automated security updates](#automated-security-updates)                         |  0.3.0   |
+| [Immutable rootfs backed by dm-verity](#immutable-rootfs-backed-by-dm-verity)     |  0.3.0   |
+| [Stateless tmpfs for /etc](#stateless-tmpfs-for-etc)                              |  0.3.0   |
+| [No shell or interpreters installed](#no-shell-or-interpreters-installed)         |  0.3.0   |
+| [Executables built with hardening flags](#executables-built-with-hardening-flags) |  0.3.0   |
+| [SELinux enabled in enforcing mode](#selinux-enabled-in-enforcing-mode)           |  0.3.0   |
+| [Kernel lockdown in integrity mode](#kernel-lockdown-in-integrity-mode)           |  1.1.0   |
+| [Secure Boot enabled](#secure-boot-enabled)                                       |  1.15.0  |
 
 The version listed indicates the first release of Bottlerocket that included the feature.
 Features may evolve or improve over time.
@@ -71,8 +73,8 @@ The kernel is configured to restart if corruption is detected.
 That allows the system to fail closed if the underlying block device is unexpectedly modified and the node is in an unknown state.
 The uncontrolled reboot will disrupt running containers, which can trigger alarms and prompt administrators to investigate.
 
-Although this provides a powerful layer of protection, it is **incomplete**.
-An attacker with full access to the block device could alter both the verity metadata and the contents of the root filesystem.
+Although this provides a powerful layer of protection, it is incomplete unless [Secure Boot is enabled](#secure-boot-enabled).
+Otherwise, an attacker with full access to the block device could alter both the verity metadata and the contents of the root filesystem.
 
 ### Stateless tmpfs for /etc
 
@@ -151,3 +153,30 @@ The policy in Bottlerocket has the following objectives:
 
 The policy is currently aimed at hardening the OS against persistent threats.
 Future enhancements to the policy will focus on mitigating the impact of OS vulnerabilities, and protecting containers from other containers.
+
+### Kernel lockdown in integrity mode
+
+Bottlerocket enables Lockdown in "integrity" mode by default on most variants.
+
+Lockdown is a Linux Security Module (LSM) that blocks certain actions which could compromise the Linux kernel.
+As with SELinux, even processes that run as root with full capabilities are subject to these restrictions.
+
+Certain variants such as `*-nvidia` need to load unsigned kernel modules at runtime.
+This is prohibited by the "integrity" mode, but required for the hardware to work as expected.
+On these variants, Lockdown is set to "none" instead.
+
+### Secure Boot enabled
+
+Bottlerocket enables Secure Boot for all new variants on platforms that support UEFI boot.
+
+The goal is to prevent unsigned, untrusted code from running at any point until containers are started.
+This is achieved by establishing the following chain of trust:
+1) The trusted platform firmware verifies that shim is signed correctly, then loads it.
+2) shim verifies that grub is signed correctly, then loads it.
+3) grub verifies that its grub.cfg is signed correctly, then loads it.
+4) grub verifies that the Linux kernel is signed correctly, then loads it.
+5) The Linux kernel verifies that the [immutable root filesystem](#immutable-rootfs-backed-by-dm-verity) has not been altered.
+
+Secure Boot only applies to platforms using UEFI firmware, and it is only enforced when the feature is enabled in the firmware.
+Therefore, systems using the legacy BIOS boot mode cannot benefit from Secure Boot.
+This includes Xen-based EC2 instance types, and bare metal machines configured to emulate the legacy BIOS boot mode.
--- a/packages/grub/0045-mkimage-pgp-move-single-public-key-into-its-own-sect.patch
+++ b/packages/grub/0045-mkimage-pgp-move-single-public-key-into-its-own-sect.patch
@@ -1,0 +1,261 @@
+From eaa371d26be8f76f9d7ce7e9c4b5d30684a48f11 Mon Sep 17 00:00:00 2001
+From: Markus Boehme <markubo@amazon.com>
+Date: Mon, 24 Oct 2022 15:20:23 +0000
+Subject: [PATCH] mkimage, pgp: move single public key into its own section of
+ EFI image
+
+If mkimage is asked to embed a single public key for signature checking
+into an EFI image, move that key into a new .pubkey section of the PE
+file. Moving the key into its dedicated section allows for easily
+swapping the public key, e.g. using objcopy, without having to rebuild
+the image.
+
+If more than one key is to be embedded, no new section is created and
+all keys are embedded as modules just as before. The PGP signature
+verification will check both of these sources for valid keys.
+
+Signed-off-by: Markus Boehme <markubo@amazon.com>
+---
+ grub-core/commands/pgp.c | 28 ++++++++++++++
+ include/grub/efi/efi.h   |  2 +-
+ util/mkimage.c           | 84 +++++++++++++++++++++++++++-------------
+ 3 files changed, 87 insertions(+), 27 deletions(-)
+
+diff --git a/grub-core/commands/pgp.c b/grub-core/commands/pgp.c
+index b81ac0a..5fa1e8e 100644
+--- a/grub-core/commands/pgp.c
++++ b/grub-core/commands/pgp.c
+@@ -32,6 +32,7 @@
+ #include <grub/kernel.h>
+ #include <grub/extcmd.h>
+ #include <grub/verify.h>
++#include <grub/efi/efi.h>
+ 
+ GRUB_MOD_LICENSE ("GPLv3+");
+ 
+@@ -959,6 +960,33 @@ GRUB_MOD_INIT(pgp)
+     grub_pk_trusted = pk;
+   }
+ 
++#ifdef GRUB_MACHINE_EFI
++  {
++    grub_addr_t pubkey_section;
++    grub_uint32_t pubkey_vsz;
++
++    pubkey_section = grub_efi_section_addr (".pubkey", &pubkey_vsz);
++    if (pubkey_section)
++      {
++	struct grub_file pseudo_file;
++	struct grub_public_key *pk;
++
++	grub_memset (&pseudo_file, 0, sizeof (pseudo_file));
++
++	pseudo_file.fs = &pseudo_fs;
++	pseudo_file.size = pubkey_vsz;
++	pseudo_file.data = (char *) pubkey_section;
++
++	pk = grub_load_public_key (&pseudo_file);
++	if (!pk)
++	  grub_fatal ("error loading key from .pubkey section: %s\n", grub_errmsg);
++
++	pk->next = grub_pk_trusted;
++	grub_pk_trusted = pk;
++      }
++  }
++#endif /* GRUB_MACHINE_EFI */
++
+   if (!val)
+     grub_env_set ("check_signatures", grub_pk_trusted ? "enforce" : "no");
+ 
+diff --git a/include/grub/efi/efi.h b/include/grub/efi/efi.h
+index 5841a2e..d580b6b 100644
+--- a/include/grub/efi/efi.h
++++ b/include/grub/efi/efi.h
+@@ -152,7 +152,7 @@ grub_err_t grub_arch_efi_linux_boot_image(grub_addr_t addr, grub_size_t size,
+ 					  char *args, int nx_enabled);
+ #endif
+ 
+-grub_addr_t grub_efi_section_addr (const char *section, grub_uint32_t *vsz);
++grub_addr_t EXPORT_FUNC(grub_efi_section_addr) (const char *section, grub_uint32_t *vsz);
+ 
+ void grub_efi_mm_init (void);
+ void grub_efi_mm_fini (void);
+diff --git a/util/mkimage.c b/util/mkimage.c
+index 1455c94..01362d1 100644
+--- a/util/mkimage.c
++++ b/util/mkimage.c
+@@ -65,14 +65,14 @@
+ 				    + GRUB_PE32_SIGNATURE_SIZE		\
+ 				    + sizeof (struct grub_pe32_coff_header) \
+ 				    + sizeof (struct grub_pe32_optional_header) \
+-				    + 5 * sizeof (struct grub_pe32_section_table), \
++				    + 6 * sizeof (struct grub_pe32_section_table), \
+ 				    GRUB_PE32_FILE_ALIGNMENT)
+ 
+ #define EFI64_HEADER_SIZE ALIGN_UP (GRUB_PE32_MSDOS_STUB_SIZE		\
+ 				    + GRUB_PE32_SIGNATURE_SIZE		\
+ 				    + sizeof (struct grub_pe32_coff_header) \
+ 				    + sizeof (struct grub_pe64_optional_header) \
+-				    + 5 * sizeof (struct grub_pe32_section_table), \
++				    + 6 * sizeof (struct grub_pe32_section_table), \
+ 				    GRUB_PE32_FILE_ALIGNMENT)
+ 
+ static const struct grub_install_image_target_desc image_targets[] =
+@@ -887,12 +887,13 @@ grub_install_generate_image (const char *dir, const char *prefix,
+   char *kernel_img, *core_img;
+   size_t total_module_size, core_size;
+   size_t memdisk_size = 0, config_size = 0;
+-  size_t prefix_size = 0, dtb_size = 0, sbat_size = 0;
++  size_t prefix_size = 0, dtb_size = 0, pubkey_size = 0, sbat_size = 0;
+   char *kernel_path;
+   size_t offset;
+   struct grub_util_path_list *path_list, *p;
+   size_t decompress_size = 0;
+   struct grub_mkimage_layout layout;
++  int pubkey_section = 0;
+ 
+   if (comp == GRUB_COMPRESSION_AUTO)
+     comp = image_target->default_compression;
+@@ -911,7 +912,10 @@ grub_install_generate_image (const char *dir, const char *prefix,
+   else
+     total_module_size = sizeof (struct grub_module_info32);
+ 
+-  {
++  if (npubkeys == 1 && image_target->id == IMAGE_EFI)
++    pubkey_section = 1;
++
++  if (!pubkey_section) {
+     size_t i;
+     for (i = 0; i < npubkeys; i++)
+       {
+@@ -1048,24 +1052,25 @@ grub_install_generate_image (const char *dir, const char *prefix,
+       offset += mod_size;
+     }
+ 
+-  {
+-    size_t i;
+-    for (i = 0; i < npubkeys; i++)
+-      {
+-	size_t curs;
+-	struct grub_module_header *header;
+-
+-	curs = grub_util_get_image_size (pubkey_paths[i]);
+-
+-	header = (struct grub_module_header *) (kernel_img + offset);
+-	header->type = grub_host_to_target32 (OBJ_TYPE_GPG_PUBKEY);
+-	header->size = grub_host_to_target32 (curs + sizeof (*header));
+-	offset += sizeof (*header);
+-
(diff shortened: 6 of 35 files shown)


69XXX304 (approved):

bcressey:
^ force push fixes lint check failures 😿

packages/shim/shim.spec
```diff
@@ -0,0 +1,59 @@
...
+%global commit 11491619f4336fef41c3519877ba242161763580
+
+Name: %{_cross_os}shim
+Version: %{shimver}
+Release: 1%{?dist}
+Summary: UEFI shim loader
+License: BSD-3-Clause
+URL: https://github.com/rhboot/shim/
+Source0: https://github.com/rhboot/shim/archive/%{shimver}/shim-%{shimver}.tar.gz
+Source1: https://github.com/rhboot/gnu-efi/archive/refs/heads/shim-%{gnuefiver}.tar.gz#/gnu-efi-shim-%{gnuefiver}.tar.gz
```
arnaldo2792, line 21: Doesn't this have to match the `shim` version? FWIW, there is a new version 15.7:

https://github.com/rhboot/gnu-efi/tree/shim-15.7

sbkeys/generate-aws-sbkeys
```diff
@@ -0,0 +1,335 @@
+#!/usr/bin/env bash
```
arnaldo2792, line 1: Shouldn't these scripts be in `tools`?

tools/buildsys/src/builder.rs
```diff
@@ -369,6 +375,37 @@ enum Retry<'a> {
...
+            &s.file_name().to_string_lossy(),
+            &s.path().to_string_lossy(),
+        );
+    }
+
+    for var in &[
+        "AWS_ACCESS_KEY_ID",
+        "AWS_SECRET_ACCESS_KEY",
+        "AWS_SESSION_TOKEN",
+    ] {
```
arnaldo2792, line 413: How does buildsys get the region? Is it read from the `Infra.toml` file?

sbkeys/README.md
```diff
@@ -0,0 +1,91 @@
...
+This is done to minimize costs and to avoid requiring developers to set up infrastructure for key management ahead of time.
+However, because these local profiles offer direct access to private key materials, they are **strongly discouraged** for any kind of production use.
+
+Different profiles can be [specified at build time](#specify-a-profile-at-build-time) to use a custom set of keys.
+
+## Create a profile with local resources
+
+The `generate-local-sbkeys` tool can be used to create a local Secure Boot Keys profile.
+
+It uses `openssl` to generate private keys and certificate authorities (CAs), and `gpg` to create a GPG private key.
```
arnaldo2792, line 25: I wonder if we should mention that `openssl` and `virt-fw-vars` are called in the SDK. Or probably just call out that these tools don't have to be present in the building host and the only dependency is `docker`?

sbkeys/README.md
```diff
@@ -0,0 +1,91 @@
...
+It also uses [virt-fw-vars](https://pypi.org/project/virt-firmware/) to generate EFI variable data for the edk2 variable stores used by QEMU, and for registering Amazon EC2 AMIs.
```
arnaldo2792: > for registering Amazon EC2 AMIs.

Is that accurate, the scripts call `virt-fw-vars` to output the EFI vars, but not to register AMIs (unless I misread the scripts).

bcressey: You understood correctly - I'll see if I can revise the wording. The EFI variable data is subsequently used for QEMU runs, and also used when registering AMIs.

tools/buildsys/src/builder.rs
bcressey: The region is embedded in `kms-sign.json`, so it doesn't need to come in from the environment.

packages/shim/shim.spec
bcressey: `gnuefi` is pulled in as a Git submodule during the shim build, so this needs to match whatever the shim release expects. Currently that's the same as the most recent 15.6 commit in the gnuefi repo.

sbkeys/generate-aws-sbkeys
bcressey: I started with them in `tools`, but I felt like it was easier to understand the `sbkeys` directory by having the scripts colocated with the output and the `README.md` that documents them.

Dockerfile
```diff
@@ -196,13 +198,27 @@ ARG OS_IMAGE_PUBLISH_SIZE_GIB
...
 ARG GRUB_SET_PRIVATE_VAR
+ARG UEFI_SECURE_BOOT
 ENV VARIANT=${VARIANT} VERSION_ID=${VERSION_ID} BUILD_ID=${BUILD_ID} \
     PRETTY_NAME=${PRETTY_NAME} IMAGE_NAME=${IMAGE_NAME} \
     KERNEL_PARAMETERS=${KERNEL_PARAMETERS}
 WORKDIR /root
 
 USER root
 RUN --mount=target=/host \
+    --mount=type=secret,id=PK.crt,target=/root/sbkeys/PK.crt \
```
webern, line 213: These files are always required?

.gitignore
```diff
@@ -11,6 +11,7 @@
 /*.pem
 /keys
 /roles
+/sbkeys/**/
```
webern, line 14: I'm confused. Did you check-in the `sbkeys` README and scripts with a git force? Should you explicitly allow-list them in the `.gitignore`?

sbkeys/generate-aws-sbkeys
```diff
@@ -0,0 +1,335 @@
...
+  token="${2:?}"
+  cn="${3:?}"
+  ca_arn="${4:?}"
+
+  openssl req -new \
+    -key "pkcs11:token=${token}" -keyform engine -engine pkcs11 \
+    -subj "/CN=${cn}/" \
+    -out "${key}.csr"
+    cert_arn="$(\
+    aws acm-pca issue-certificate \
```
webern, line 212: I think the aws CLI is a new dependency. Is this running in the SDK container? I'm also seeing `gencert` and `gpg`, `virt-fw-vars` as new dependencies. I think this must be running in the SDK.

In general, I don't love seeing so much bash and wish more could be done with buildsys or pubsys.

sbkeys/README.md
webern: Does this make it clearer?

```suggestion
It also uses [virt-fw-vars](https://pypi.org/project/virt-firmware/) to generate EFI variable data for the edk2 variable stores used by Amazon EC2 AMIs and QEMU .
```

sbkeys/README.md
```diff
@@ -0,0 +1,91 @@
...
+KEK_CA="arn:aws:acm-pca:us-west-2:999999999999:certificate-authority/22222222-2222-2222-2222-222222222222"
+DB_CA="arn:aws:acm-pca:us-west-2:999999999999:certificate-authority/33333333-3333-3333-3333-333333333333"
+VENDOR_CA="arn:aws:acm-pca:us-west-2:999999999999:certificate-authority/44444444-4444-4444-4444-444444444444"
+
+# AWS KMS asymmetric keys
+SHIM_SIGN_KEY="arn:aws:kms:us-west-2:999999999999:key/aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa"
+CODE_SIGN_KEY="arn:aws:kms:us-west-2:999999999999:key/bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb"
+CONFIG_SIGN_KEY="arn:aws:kms:us-west-2:999999999999:key/cccccccc-cccc-cccc-cccc-cccccccccccc"
+
+./generate-aws-sbkeys \
```
webern, line 67: Should `pubsys` be able to do this instead of a script? If so, wouldn't we want it to be possible to set and forget this by pointing Infra.toml at the right resources?

tools/buildsys/src/manifest.rs
```diff
@@ -214,6 +214,15 @@ line arguments set in the boot configuration.
 [package.metadata.build-variant.image-features]
 unified-cgroup-hierarchy = true
 ```
+
+`uefi-secure-boot` means that the bootloader and kernel are signed. The grub image for the current
+variant will have a public GPG baked in, and will expect the grub config file to have a valid
+detached signature. Published artifacts such as AMIs and OVAs will enforce the signature checks
+when the platform supports it.
+```ignore
```
webern:
```suggestion
when the platform supports it.

```ignore
```

tools/pubsys/src/aws/ami/register.rs
```diff
@@ -117,11 +122,25 @@ async fn _register_image(
...
     }
 
+    let uefi_secure_boot_enabled = variant_manifest
+        .image_features()
+        .iter()
+        .flatten()
+        .any(|f| **f == ImageFeature::UefiSecureBoot);
+
+    let (boot_mode, uefi_data) = if uefi_secure_boot_enabled {
+        (Some("uefi-preferred".into()), Some(uefi_data))
```
webern, line 132: There's no SDK constant for this magic string?

Makefile.toml
```diff
@@ -680,6 +685,52 @@ cargo install \
...
+  echo "Incomplete Secure Boot signing profile found in ${profile}." >&2
+  echo "Missing $(( expected - found )) files." >&2
+  exit 1
+fi
+
+echo "No Secure Boot signing profile found in ${profile}." >&2
+echo "Generating local keys." >&2
+
+mkdir -p "${BUILDSYS_SBKEYS_PROFILE_DIR}"
+${BUILDSYS_SBKEYS_DIR}/generate-local-sbkeys \
```
webern, line 728: So it's not running in the SDK, which means there are a lot of new dependencies of the build host.

Makefile.toml
```diff
@@ -680,6 +685,52 @@ cargo install \
...
+script_runner = "bash"
+script = [
+'''
+# Check the profile for all files needed for Secure Boot signing.
+profile="${BUILDSYS_SBKEYS_PROFILE_DIR}"
+
+found=0
+# A local profile has signing keys and certificates, while an AWS profile
+# has a config for the aws-kms-pkcs11 helper. Either type is supported.
+if [ -s "${profile}/shim-sign.key" ] && \
```
webern, line 699: I would love to see the build tools like buildsys and pubsys do more and bash do less, in general.

variants/vmware-dev/template.ovf
```diff
@@ -1 +1 @@
-../shared/template-unified.ovf
\ No newline at end of file
+../shared/template-unified-secboot.ovf
```
webern, line 1: Do we need to update VMWare usage instructions to account for secure boot enablement?

sbkeys/generate-aws-sbkeys
```diff
@@ -0,0 +1,335 @@
...
+export AWS_DEFAULT_OUTPUT="text"
+export AWS_KMS_PKCS11_DEBUG=1
+export PKCS11_MODULE_PATH="/usr/lib64/pkcs11/aws_kms_pkcs11.so"
+
+# Fetch CA certificates.
+getcacert() {
+  local arn ca
+  arn="${1:?}"
+  ca="${2:?}"
+  aws acm-pca get-certificate-authority-certificate \
```
yeazelm, line 173: Do we cover what permissions are required to run these scripts? I'm guessing there is a minimal set of things an IAM User/Role needs to make all these calls that we probably should call them out or at least link to the upstream docs on permissions? I think its also different permissions needed for generating the keys vs signing builds correct?

packages/grub/grub.spec
```diff
@@ -14,9 +14,10 @@ Release: 1%{?dist}
 Summary: Bootloader with support for Linux and more
 License: GPL-3.0-or-later AND Unicode-DFS-2015
 URL: https://www.gnu.org/software/grub/
-Source0: https://al2022-repos-us-west-2-9761ab97.s3.dualstack.us-west-2.amazonaws.com/blobstore/aa41fdf9982b65a4c4dad5df5b49ba143b1710d60f82688221966f3c790c6c63/grub2-2.06-42.amzn2022.0.1.src.rpm
+Source0: https://cdn.amazonlinux.com/al2023/blobstore/74f9ee6e75b8f89fe91ccda86896243179968a8664ba045bece11dc5aff61f4e/grub2-2.06-61.amzn2023.0.6.src.rpm
```
webern, line 18: This looks fishy. Do we keep this in the lookaside cache or somewhere under the `bottlerocket.aws` domain?

tools/replace-grub-pubkey
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
webern: We should probably differentiate scripts that run in the SDK vs those that run on the host somehow. Maybe comments would suffice.

tools/rpm2img
```diff
@@ -255,11 +266,171 @@ fi
...
+  # Generate the PKCS12 archives for import.
+  openssl pkcs12 \
+    -export \
+    -passout pass: \
+    -inkey "${SBKEYS}/shim-sign.key" \
+    -in "${SBKEYS}/shim-sign.crt" \
+    -certfile "${SBKEYS}/db.crt" \
+    -out "${SBKEYS}/shim-sign.p12"
+
+  openssl pkcs12 \
```
webern: our previous efforts to keep openssl out of the sdk were in vain.

webern (approved):

tools/start-local-vm
```diff
@@ -160,6 +187,32 @@ prepare_raw_images() {
...
+        cp "${original_vars}" "${firmware_vars}"
+    fi
+
+    if [[ ${arch} = aarch64 ]]; then
+        firmware_code=${firmware_code:-/usr/share/edk2/aarch64/QEMU_EFI.silent.fd}
+        original_vars=${firmware_vars:-/usr/share/edk2/aarch64/QEMU_VARS.fd}
+        firmware_vars="$(mktemp)"
+        on_exit "rm '${firmware_vars}'"
+        cat "${original_vars}" /dev/zero \
+            | head -c 64m > "${firmware_vars}"
```
yeazelm, line 222: I'm curious why aarch64 requires different treatment?

tools/rpm2img
```diff
@@ -255,11 +266,171 @@ fi
...
+  output="${2:?}"
+  size="$(stat -c %s "${input}")"
+  rm -f "${output}"
+  # The cert payload is offset by four 4-byte uint32_t values in the header.
+  offset="$((4 * 4))"
+  for n in "${size}" 0 "${offset}" "$(( size + offset ))" ; do
+    printf -v uint32_t '\\x%02x\\x%02x\\x%02x\\x%02x' \
+      $((n & 255)) $((n >> 8 & 255)) $((n >> 16 & 255)) $((n >> 24 & 255))
+    printf "${uint32_t}" >> "${output}"
+  done
```
yeazelm, line 368: 🤯 this is raising the bar in what I've seen folks do with bash! 😄

tools/rpm2img
```diff
@@ -488,6 +692,19 @@ if [ "${OUTPUT_FMT}" == "vmdk" ] ; then
      -e "s/{{DATA_DISK_BYTES}}/${data_disk_bytes}/g" \
      > "${ova_dir}/${ovf}"
 
+  if [ "${UEFI_SECURE_BOOT}" == "yes" ] ; then
```
yeazelm, line 741: It might be good to leave a comment as to why we are doing this in case we ever need to change things around ovf things.

packages/grub/grub.spec
markusboehme: The RPM lists the canonical source URL. A copy of the source artifact lives in the look-aside cache and is used during regular builds. This isn't any different from other third-party packages in that respect. For GRUB, as for the kernel and microcode, Bottlerocket's upstream just happens to be Amazon Linux.

packages/grub/grub.spec
```diff
@@ -177,7 +177,8 @@ sed -e "s,__VERSION__,%{version},g" %{S:3} > sbat.csv
 popd
 
 %install
-MODS="configfile echo ext2 gptprio linux normal part_gpt reboot sleep zstd search"
+MODS=(configfile echo ext2 gptprio linux normal part_gpt reboot sleep zstd search)
+MODS+=(pgp crypto gcry_sha256 gcry_sha512 gcry_dsa gcry_rsa)
```
markusboehme: Include the crypto modules only if the Secure Boot feature flag is set?

packages/grub/grub.spec
```diff
@@ -189,7 +190,7 @@ mkdir -p %{buildroot}%{biosdir}
   -O "i386-pc" \
   -o "%{buildroot}%{biosdir}/core.img" \
   -p "(hd0,gpt2)/boot/grub" \
-  biosdisk serial ${MODS}
+  biosdisk serial ${MODS[@]}
```
markusboehme, line 196: The crypto modules could be factored out since the BIOS build of GRUB won't participate in Secure Boot.

tools/replace-grub-pubkey
markusboehme: This was originally intended as a test script. I didn't mean for it to end up here, though if it works, it works... I didn't leave any "here be dragons" foot guns I'm aware of. :-) But this may explain the absence of help texts, more convenient argument parsing, explicit error handling instead of `set -e`. 

I don't think scripts should need to know from where they are invoked, though. It seems outside of their area of control. In which situations do you think it would help? To delineate which scripts are implementation details and can be easily refactored, and which might be depended on from out-of-tree users?

packages/shim/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "shim"
+version = "0.1.0"
+edition = "2018"
+publish = false
+build = "build.rs"
+
+[lib]
+path = "pkg.rs"
```
markusboehme: A rebase could pick up your changes to get rid of `pkg.rs` and `build.rs`.

sbkeys/generate-aws-sbkeys
```diff
@@ -0,0 +1,335 @@
...
+    -e SHIM_SIGN_KEY="${SHIM_SIGN_KEY}" \
+    -e CODE_SIGN_KEY="${CODE_SIGN_KEY}" \
+    -e CONFIG_SIGN_KEY="${CONFIG_SIGN_KEY}" \
+    -e AWS_KMS_PKCS11_CONF="/tmp/aws-kms-pkcs11-conf" \
+    -e OUTPUT_DIR="/tmp/output" \
+    -w /tmp \
+    "${SDK_IMAGE}" bash /tmp/sbkeys
+else
+  export PK_CA KEK_CA DB_CA VENDOR_CA
+  export CODE_SIGN_KEY CONFIG_SIGN_KEY
```
markusboehme: The `SHIM_SIGN_KEY` is not being exported here.

sbkeys/generate-aws-sbkeys
```diff
@@ -0,0 +1,335 @@
...
+  key="${1:?}"
+  token="${2:?}"
+  cn="${3:?}"
+  ca_arn="${4:?}"
+
+  openssl req -new \
+    -key "pkcs11:token=${token}" -keyform engine -engine pkcs11 \
+    -subj "/CN=${cn}/" \
+    -out "${key}.csr"
+    cert_arn="$(\
```
markusboehme: Super nit: Indentation threw me off for a moment here (i.e. this is not a continuation of the `openssl` invocation).

sbkeys/generate-aws-sbkeys
```diff
@@ -0,0 +1,335 @@
...
+gpg --no-tty --expert --full-generate-key --command-fd 0 <<EOF
+13
+${KEYGRIP}
+e
+q
+0
+Bottlerocket Config Signing Key
+
+
+EOF
```
markusboehme, line 295: It would be helpful to have the input to GPG commented, for example:

```
# 13           Existing key
# ${KEYGRIP}   Which key to edit
# e            Toggle the encrypt capability (presumably we turn it off?)
# q            Finished
# 0            key does not expire
# Bottlero...  Real name
#              Email address
#              Comment
```

sbkeys/generate-aws-sbkeys
```diff
@@ -0,0 +1,335 @@
...
+    -subj "/CN=${cn}/" \
+    -out "${key}.csr"
+    cert_arn="$(\
+    aws acm-pca issue-certificate \
+      --certificate-authority-arn "${ca_arn}" \
+      --template-arn arn:aws:acm-pca:::template/BlankEndEntityCertificate_APICSRPassthrough/V1 \
+      --csr "fileb://${key}.csr" \
+      --api-passthrough "file://codesign.json" \
+      --signing-algorithm "SHA256WITHRSA" \
+      --validity Value=5,Type="YEARS" \
```
markusboehme, line 218: I assume the impact of those certificates expiring is the boot failing. From an operational point of view, will there be a new set of keys per variant? Five years of lifetime seems safe now, as variants would phase out before that (perhaps `aws-ecs-1` could come close). Should this be documented anyway?

Makefile.toml
markusboehme: It's both running on the host and in the SDK. [`generate-local-sbkeys`](https://github.com/bottlerocket-os/bottlerocket/blob/38dcb4c6f018f6af0bd8f635f3b7a1e839744fe4/sbkeys/generate-local-sbkeys) is two scripts in one/"self-extracting"/has a prelude that executes its actually meaty parts inside the SDK if passed `--sdk-image` like is done here. There should not be any extra dependencies for the build host.

tools/rpm2img
```diff
@@ -257,11 +266,171 @@ fi
...
+    -passout pass: \
+    -inkey "${SBKEYS}/code-sign.key" \
+    -in "${SBKEYS}/code-sign.crt" \
+    -certfile "${SBKEYS}/vendor.crt" \
+    -out "${SBKEYS}/code-sign.p12"
+
+  # Import certificates and private key archive.
+  PEDB="/etc/pki/pesign"
+
+  certutil -d "${PEDB}" -A -n db -i "${SBKEYS}/db.crt" -t "CT,C,C"
```
markusboehme: I'm curious, why set the SSL and email trust attributes for CA certs, and why those?

tools/rpm2img
```diff
@@ -257,11 +266,171 @@ fi
 # package has placed the image in /boot/efi/EFI/BOOT.
 mv "${ROOT_MOUNT}/boot/efi"/* "${EFI_MOUNT}"
 
+# For an AWS profile, we expect a config file for the PKCS11
+# helper. Otherwise, there should be a local key and cert.
```
markusboehme: To ensure I'm not missing anything (and it gets documented in the code if I am, to prevent future refactoring), the following blocks could be conditional on the use of Secure Boot, right? For example, shim could live with the all-zeros `vendor_cert` structure already embedded (unlike GRUB where the pubkey section needs to be removed)?

tools/rpm2img
```diff
@@ -257,11 +266,171 @@ fi
...
+grub="${grubs[0]}"
+mokms=(mm*.efi)
+mokm="${mokms[0]}"
+if [ "${UEFI_SECURE_BOOT}" == "yes" ] ; then
+  # Replace the embedded vendor certificate, then sign shim with the db key.
+  objdumpcopy "${shim}" \
+    --update-section ".vendor_cert=${SBKEYS}/vendor.obj"
+  pesign -i "${shim}" -o "${shim}.signed" -s "${SHIM_SIGN_KEY[@]}"
+  mv "${shim}.signed" "${shim}"
+  pesigcheck -i "${shim}" -n 0 -c "${SBKEYS}/db.cer"
```
markusboehme, line 404: As an aside, that's curious argument parsing behavior on pesign/popt. The help text suggests `-n, --no-system-db=INT`, i.e. short flag or long option with value, but a mix is fine, too.

PROVISIONING-METAL.md
```diff
@@ -332,3 +332,18 @@ docker run --rm \
    "${SDK_IMAGE}" \
    bootconfig -l /tmp/bootconfig.data
 ```
+
+### Enable Secure Boot
+
+Starting with metal-k8s-1.28, the Bottlerocket images for bare metal support Secure Boot when used on a platform with UEFI firmware.
+UEFI boot mode must be used, rather than legacy BIOS boot mode, and Secure Boot must be enabled.
```
markusboehme, line 340: Nit: We may also want to mention CSM. I think I've seen various machine's firmware call it differently.

For example, "UEFI boot mode must be used, rather than legacy BIOS boot mode (sometimes referred to as Compatibility Support Module or CSM), and Secure Boot must be enabled."

markusboehme (commented):
I'm sorry it took a while to get to, and get through! I needed to rebuild some context, but am happy to see the Secure Boot feature so polished! :sparkles:

.gitignore
bcressey: This excludes subdirectories of `sbkeys` since that is where the default "local" profile is written. It doesn't affect files in the `sbkeys` directory.

sbkeys/generate-aws-sbkeys
bcressey: Yes, these are all dependencies that exist in the SDK.

tools/pubsys/src/aws/ami/register.rs
bcressey: We'd need a newer AWS SDK for Rust, which I don't want to block on here.

variants/vmware-dev/template.ovf
bcressey: No, the usage is the same - newly imported OVAs will just be configured to use Secure Boot by default.

tools/replace-grub-pubkey
bcressey: @markusboehme should I just drop this test script? I didn't want to just throw it away but I agree it's not really needed at the moment.

tools/rpm2img
bcressey: No, we've always needed `openssl` for the host in the SDK. It's a build dependency for the kernel, which generates temporary private keys to sign its modules as part of the build.

tools/start-local-vm
bcressey: The AAVMF and OVMF firmware implementations expect different inputs.

packages/grub/grub.spec
bcressey: > Include the crypto modules only if the Secure Boot feature flag is set?

I'd originally done this, but I've been trying to eliminate variant-specific package builds where possible, since they're all going to be awkward for out-of-tree builds.

Dockerfile
bcressey: > These files are always required?

Secrets like this aren't always required - if they exist then they'll be mounted in the specified place, but the build proceeds without error if they're not present.

tools/replace-grub-pubkey
markusboehme: I don't have a strong opinion either way. It is helpful when maintaining the respective GRUB patches; perhaps there's other uses. It's not outright dangerous or anything, just not as polished as other scripts in `tools/`.

/edit: Thinking about it, perhaps that makes it a good candidate for moving into `packages/grub` instead.

packages/grub/grub.spec
markusboehme: That's okay with me. I'll need to read up more on out-of-tree to avoid introducing similar foot guns myself.

tools/rpm2img
webern: I have been confused of late between the static library version that we used to build and the "os" version. :+1:

bcressey:
⬆️ force push:
* rebase on develop
* adjust macro usage in `shim.spec` and `grub.spec` based on changes in https://github.com/bottlerocket-os/bottlerocket-sdk/pull/117

I'll follow up with a different push for any functional changes or fixes.

sbkeys/README.md
bcressey: > Should pubsys be able to do this instead of a script? If so, wouldn't we want it to be possible to set and forget this by pointing Infra.toml at the right resources?

Conceptually, this script is somewhat closer to what we were trying to do with `infrasys`, where we create assets that can later be referenced in `Infra.toml`.

But I didn't see a good reason to try to standardize an "aws" shape and a "local" shape in `Infra.toml`, since the two profile types are very different and require different handling in `rpm2img` for signing to actually work.

`rpm2img` eventually does channel them into a mostly-shared code path, which requires correct NSS and GPG configurations so that `pesign` and `gpg` can "just work" later on. Even there, different arguments are required for the actual CLI invocation when signing.

What would help is if there were some sort of virtual smartcard support for local keys generated by OpenSSL. Then we could use the PKCS#11 code path consistently, and `pubsys-setup` or some other helper could have "configure smartcard for signing" as its contract, and `rpm2img` could just rely on a correctly-configured smartcard interface with standard key and token names.

AFAICT we'd have to write our own PKCS#11 provider with a "local" backend for that to work - like `aws-kms-pkcs11` does for "aws" - which was more than I wanted to tackle here. With that provider in place, we'd need something like these scripts to generate the required files. The output would just look more like this:
```
❯ tree aws local
aws
├── config-sign.key
├── kms-sign.json
└── efi-certs.json
local
├── config-sign.key
├── local-sign.json
└── efi-certs.json
```

We could shift some of what the call to `virt-fw-vars` does into an implementation where `pubsys-setup` reads cert information from `efi-certs.json` and creating `efi-vars.json`.

 I'm happy to backlog an issue for it; I could see it being useful for OOTBs for the same reason that it's useful that `tuftool` can abstract a bit over "local keys" vs. "KMS keys".

sbkeys/generate-aws-sbkeys
bcressey: > I assume the impact of those certificates expiring is the boot failing.

I expect it's firmware-specific, but QEMU seems happy with an expired certificate, if a VM is started with `-rtc base=2036-11-11T11:11:11`, and shim says ["UEFI explicitly allows expired certificates"](https://github.com/rhboot/shim/blob/e24681211ce8a71583b5233084bd1290c3c7e872/Cryptlib/Pk/CryptPkcs7Verify.c#L103).

In most cases it seems undesirable to have boot fail just because the system clock is wrong, like if the RTC battery dies and it resets back to the epoch.

> From an operational point of view, will there be a new set of keys per variant? Five years of lifetime seems safe now, as variants would phase out before that (perhaps aws-ecs-1 could come close). Should this be documented anyway?

For our builds, the plan is to create a new generation of signing keys whenever we introduce a newer kernel. Not sure where we'd document that though; we don't document our (annual) TUF root rotation schedule either.

tools/rpm2img
bcressey: IIRC I found those in attributes in a wiki that gave an example of `certutil` and `pesign` usage. As you suggest, it doesn't really make sense to set them, and it doesn't seem to affect `pesign` at all, so I'll drop them.

tools/rpm2img
bcressey: > To ensure I'm not missing anything (and it gets documented in the code if I am, to prevent future refactoring), the following blocks could be conditional on the use of Secure Boot, right?

I'll clean this up and move it all into a Secure Boot conditional. I'd wanted to ensure that the setup parts always got exercised, since we always expect a valid signing profile, but doing it this way means that it'd be easy to miss cases where we try to sign something or copy a file that should only exist when the feature flag is turned on.

> For example, shim could live with the all-zeros vendor_cert structure already embedded (unlike GRUB where the pubkey section needs to be removed)?

`shim` ignores the cert if Secure Boot is not enabled. However, in double-checking this I found a discrepancy in `mokutil` output:

```
# All-zeroes vendor cert specified at shim build time, and never replaced.
$ mokutil --list-enrolled
[key 1]
Invalid X509 certificate

# No vendor cert specified at shim build time.
$ mokutil --list-enrolled
[key 1]
  [SHA-256]
  0000000000000000000000000000000000000000000000000000000000000000
```

Replacing the empty cert with a zero-length cert cleans this up so we always get the second case.

bcressey:
⬆️ force push addresses most of the feedback.

bcressey:
⬆️ force push for whitespace fix

packages/grub/0045-mkimage-pgp-move-single-public-key-into-its-own-sect.patch
```diff
@@ -0,0 +1,261 @@
...
++	pseudo_file.size = pubkey_vsz;
++	pseudo_file.data = (char *) pubkey_section;
++
++	pk = grub_load_public_key (&pseudo_file);
++	if (!pk)
++	  grub_fatal ("error loading key from .pubkey section: %s\n", grub_errmsg);
++
++	pk->next = grub_pk_trusted;
++	grub_pk_trusted = pk;
++      }
```
foersleo, line 63: nit: Indentation seems to be quite a mess in grub code. But I think we would want to at least be consistent within the code we are adding. This section seems to be tabs, while the surrounding added code is spaces.

packages/shim/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "shim"
+version = "0.1.0"
+edition = "2018"
```
foersleo: We probably want to go with 2021 for this, like we do for all other packages since https://github.com/bottlerocket-os/bottlerocket/commit/b670aab01626dcd328468c5fddc54faa1a8edb0b

foersleo (commented):
Looks fine to me, just the two minor nits that are not blocking anything.
I can right now however not fully grasp why the linter gets thrown off on the build jobs.

bcressey:
> I can right now however not fully grasp why the linter gets thrown off on the build jobs.

I noticed those failures just now and I'm equally mystified. I'll try the clippy run locally and see if I can repro.

bcressey:
⬆️ force push to document IAM permissions for `generate-aws-sbkeys`

sbkeys/generate-aws-sbkeys
bcressey: Good call - I've updated `README.md` to cover the permissions needed for generating the profile and for signing builds with the profile.

packages/grub/0045-mkimage-pgp-move-single-public-key-into-its-own-sect.patch
bcressey: Hmm, I think these tabs in `pgp.c` might be correct (per GNU style) since the `if` statement increases the indent to 8 spaces, but `mkimage.c` doesn't seem to match the style.

bcressey: Unless you feel very strongly about this, I'm inclined to just leave the patch to `mkimage.c` as-is. I tried a number of approaches - `indent` runs, vim retabs - and wasn't happy with the results.

bcressey:
⬆️ force push to fix the edition in the `shim` package

bcressey:
⬆️ force push rebases on develop with no other changes. I think I need the changes from #3228 on this branch to appease clippy, which seems to be confused by the newly cached artifacts.

sbkeys/generate-aws-sbkeys
bcressey: If I try to sign with an expired certificate then `pesign` complains:
```
  #14 9.402 Peer's Certificate has expired.
  #14 9.402 Searching db db.cer
  #14 9.402 Searching db db.cer
  #14 9.402 Signature has impossible time constraint: 1688764933 <= 1688763121
  #14 9.402 pesigcheck: "bootx64.efi" is invalid.
```

So the main operational impact should just be that we won't be able to (easily) use the same signing certs for more than five years, which seems fine. There's a balance to strike between not using the same certs forever vs. using few enough that it's reasonable to revoke based on the cert rather than a list of individual hashes.

packages/grub/0045-mkimage-pgp-move-single-public-key-into-its-own-sect.patch
foersleo: I do not feel strongly about it. Let's leave it as is. Checking random files in grub they all had mixed-indent so it does not really matter.

sbkeys/generate-aws-sbkeys
markusboehme: > > I assume the impact of those certificates expiring is the boot failing.
> 
> I expect it's firmware-specific, but QEMU seems happy with an expired certificate, if a VM is started with `-rtc base=2036-11-11T11:11:11`, and shim says ["UEFI explicitly allows expired certificates"](https://github.com/rhboot/shim/blob/e24681211ce8a71583b5233084bd1290c3c7e872/Cryptlib/Pk/CryptPkcs7Verify.c#L103).

That's interesting, thanks for digging it up!



> > From an operational point of view, will there be a new set of keys per variant? Five years of lifetime seems safe now, as variants would phase out before that (perhaps aws-ecs-1 could come close). Should this be documented anyway?
> 
> For our builds, the plan is to create a new generation of signing keys whenever we introduce a newer kernel. Not sure where we'd document that though; we don't document our (annual) TUF root rotation schedule either.

Viewed that way, this makes sense to me. Key rotation is a matter that concerns image providers. The mechanism exists to implement various policies. Bottlerocket doesn't need to prescribe (and document) a particular one.

sbkeys/generate-aws-sbkeys
```diff
@@ -0,0 +1,344 @@
...
+
+PRELUDE_END=$(awk '/=\^\.\.\^=/ { print NR+1; exit 0; }' "${0}")
+SBKEYS_SCRIPT="$(mktemp)"
+AWS_KMS_PKCS11_CONF="$(mktemp)"
+cleanup() {
+  rm -f "${SBKEYS_SCRIPT}" "${AWS_KMS_PKCS11_CONF}"
+}
+trap 'cleanup' EXIT
+tail -n +"${PRELUDE_END}" "${0}" >"${SBKEYS_SCRIPT}"
+chmod +x "${SBKEYS_SCRIPT}"
```
foersleo, line 90: I have to say I wondered about this a bit when reviewing last week. Unpacking part of the script into a temporary script and calling it from the original "outer script" seemed like an interesting trick to me, but I had a hard time understanding *why* we are doing that. Off and on over the weekend I could not really come up with an answer that made me happy.

I discussed this with Markus earlier today and we just came to the conclusion that it probably has something to do with having one script name less to be concerned about while ensuring the correct pairing of "inner" and "outer" script between the two separate variants (local and aws). (?)

Maybe we want to add some good reasoning/commenting around the divider line. Just to ensure that someone coming across this code in 5 years has an easier way to accept that this is an interesting technique aiding the proper paring of two scripts and not for some deeper reason that needs to be deciphered.

foersleo (approved):

packages/grub/0045-mkimage-pgp-move-single-public-key-into-its-own-sect.patch
markusboehme: Indeed, it's GNU style indentation. There are [mixed opinions](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/coding-style.rst?h=v6.4#n12) on it...

markusboehme (approved):
:closed_lock_with_key: :penguin:

packages/shim/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[lib]
+path = "/dev/null"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/rhboot/shim/archive/15.7/shim-15.7.tar.gz"
+sha512 = "95ef9c0125269cfa0263a32e4f343d8ccc8813d71fa918a2f54850781e3a2d6a06a719249be355fdb24c935899e0e11370815501ecde1800bdd974a9a79c5612"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/rhboot/gnu-efi/archive/refs/heads/shim-15.6.tar.gz"
```
stmcginnis, line 16: Unfortunate naming scheme.

tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -50,6 +50,10 @@ pub(crate) struct AmiArgs {
     #[arg(short = 'v', long)]
     variant_manifest: PathBuf,
 
+    /// Path to the UEFI data
+    #[structopt(short = 'e', long)]
```
stmcginnis: s/structopt/arg/

stmcginnis (commented):
Looks great overall. Just one minor thing with using a dependency of `clap` instead of `clap` itself for one of the arguments that I think we should clean up.

sbkeys/generate-aws-sbkeys
bcressey: > I discussed this with Markus earlier today and we just came to the conclusion that it probably has something to do with having one script name less to be concerned about while ensuring the correct pairing of "inner" and "outer" script between the two separate variants (local and aws).

That's essentially it. I was trying to avoid the pattern seen in `tools/docker-go` where we have a somewhat generic "run this in Docker" script that's decoupled from the business logic of what to run. Such a script ends up having to support the superset of arguments for all its callers (tasks in `Makefile.toml`, and `gomod.rs` in `buildsys`), which is confusing to implement and also to use correctly (since `--help` just shows all the supported args).

The scripts could be slightly more opinionated and make the SDK image a required parameter. During development, I found it helpful to both the "host" and "container" execution modes available since the "aws" profile type needed a lot of debugging.

If I liked `docker-go` a bit better, or if I didn't care about preserving both modes of execution, then I'd probably have specialized scripts that called into a shared executor script.

bcressey:
⬆️ force push fixes `clap` usage and adds a comment to the generator scripts about the divider line

yeazelm (approved):

packages/shim/shim.spec
```diff
@@ -0,0 +1,63 @@
...
+rm -rf %{buildroot}%{_datadir}
+
+%files
+%license COPYRIGHT
+%{_cross_attribution_file}
+%dir %{efidir}
+%{efidir}/%{boot_efi_image}
+%{efidir}/%{mokm_efi_image}
+
+%changelog
```
arnaldo2792: nit: we don't use this, right? Otherwise I've been given folks the wrong advice :sweat:

sbkeys/README.md
```diff
@@ -0,0 +1,134 @@
...
+    {
+      "Action": [
+        "acm-pca:GetCertificate",
+        "acm-pca:GetCertificateAuthorityCertificate",
+        "acm-pca:IssueCertificate",
+        "kms:GetPublicKey",
+        "kms:Sign"
+      ],
+      "Effect": "Allow",
+      "Resource": "*"
```
arnaldo2792, line 95: nit: probably in the future we could narrow down the permissions in the policy to be the CAs used for the build. Same in other policies.

SECURITY_FEATURES.md
```diff
@@ -151,3 +153,30 @@ The policy in Bottlerocket has the following objectives:
...
+Lockdown is a Linux Security Module (LSM) that blocks certain actions which could compromise the Linux kernel.
+As with SELinux, even processes that run as root with full capabilities are subject to these restrictions.
+
+Certain variants such as `*-nvidia` need to load unsigned kernel modules at runtime.
+This is prohibited by the "integrity" mode, but required for the hardware to work as expected.
+On these variants, Lockdown is set to "none" instead.
+
+### Secure Boot enabled
+
+Bottlerocket enables Secure Boot for all new variants on platforms that support UEFI boot.
```
arnaldo2792, line 170: nit: I'm OK with this right now, but what would a “new variant” be? Should we instead use the first version that enabled secure boot?

SECURITY_FEATURES.md
```diff
@@ -151,3 +153,30 @@ The policy in Bottlerocket has the following objectives:
...
+
+The goal is to prevent unsigned, untrusted code from running at any point until containers are started.
```
arnaldo2792, line 172: nit: is there any note anywhere about the init ram disk and that it is OK that in the chain of trust we don't verify its signature?

arnaldo2792 (approved):

packages/shim/shim.spec
bcressey: > nit: we don't use this, right? Otherwise I've been given folks the wrong advice 😓

MokManager is installed in case an administrator runs `mokutil` and tries to enroll a new key. Otherwise `shim` ends up in a perpetual boot failure, trying to execute it to process the enrollment.

sbkeys/README.md
bcressey: > nit: probably in the future we could narrow down the permissions in the policy to be the CAs used for the build. Same in other policies.

I'd prefer to leave that as an exercise to the reader, if it's important to them, since they'll need to fill in the correct ARNs.

SECURITY_FEATURES.md
bcressey: > nit: I'm OK with this right now, but what would a “new variant” be? Should we instead use the first version that enabled secure boot?

I'd rather document that in the release notes, as we did for the new Lockdown defaults in [v1.1.0](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.1.0).

Otherwise this content becomes less "evergreen" - if we mention `*-k8s-1.28` variants here and a year from now they've been pruned from the tree, does that mean Secure Boot isn't enabled? Conversely, if we mention a specific Bottlerocket version rather than variants, it could be interpreted as saying that everything supports Secure Boot as of that version.

packages/shim/shim.spec
stmcginnis: I think @arnaldo2792 was referring to the `%changelog` line. I've always used it, but noticed [he had said to remove it elsewhere](https://github.com/bottlerocket-os/bottlerocket/pull/3198#discussion_r1245953496).

bcressey:
⬆️ force push makes a defensive copy of the firmware code on aarch64 also, to fix an error when no code or variables are specified:
```
❯ ./tools/start-local-vm --arch aarch64 --variant metal-dev --inject-file net.toml
...
qemu-system-aarch64: device requires 67108864 bytes, block backend provides 2097152 bytes

❯ ls -latr /usr/share/edk2/aarch64/
total 206404
-rw-r--r--. 1 root root 67108864 Feb 17 21:43 vars-template-pflash.raw
-rw-r--r--. 1 root root   786432 Feb 17 21:43 QEMU_VARS.fd
-rw-r--r--. 1 root root 67108864 Feb 17 21:43 QEMU_EFI-pflash.raw
-rw-r--r--. 1 root root  2097152 Feb 17 21:43 QEMU_EFI.fd
-rw-r--r--. 1 root root 67108864 Feb 17 21:44 QEMU_EFI-silent-pflash.raw
-rw-r--r--. 1 root root  2097152 Feb 17 21:44 QEMU_EFI.silent.fd
-rw-r--r--. 1 root root   806912 Feb 17 21:46 vars-template-pflash.qcow2
-rw-r--r--. 1 root root  2117632 Feb 17 21:46 QEMU_EFI-silent-pflash.qcow2
-rw-r--r--. 1 root root  2117632 Feb 17 21:46 QEMU_EFI-pflash.qcow2
```

I could also change the script to default to `QEMU_EFI-silent-pflash.raw` and `vars-template-pflash.raw`, but this approach seems more consistent with the x86_64 case.

bcressey:
⬆️ force push removes `%changelog` from `shim.spec`

markusboehme:
Recent pushes still look good to me. :+1: Restarted a CI build that failed due to timing out on the network.

bcressey:
⬆️ force push rebases on b038795a, prior to xfs feature merge, so the next rebase diff will better show the conflict resolution.

bcressey:
⬆️ force push rebases on `develop` with the XFS feature conflicts fixed.

yeazelm (approved):
LGTM!

PR #3106, opened 2023-05-11
Dynamically get list of variants for GitHub Actions

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This updates the `build` workflow to determine the set of variants to test against at runtime. It looks at what is under the `variants` directory to determine what to include, rather than needing to hard code this list in the action itself.

This also gives the potential that we can use this in a composable workflow so multiple workflows can get the list of variants without needing to remember to update the list in multiple places whenever a variant is added or removed.

**Testing done:**

Testing GitHub Action results.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,141 +31,36 @@ env:
   GOPROXY: direct
 
 jobs:
+  list-variants:
+    if: github.repository == 'bottlerocket-os/bottlerocket'
+    runs-on: ubuntu-latest
+    outputs:
+      variants: ${{ steps.get-variants.outputs.variants }}
+      aarch-enemies: ${{ steps.get-variants.outputs.aarch-enemies }}
+    steps:
+      - uses: actions/checkout@v3
+      - id: get-variants
+        name: Determine variants
+        run: |
+          cd variants
+          output="variants=$(ls -d */ | cut -d'/' -f 1 | grep -vE '^(shared|target)$' | jq -R -s -c 'split("\n")[:-1]')"
+          echo $output
+          echo $output >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


etungsten (approved):
Nice idea!

.github/workflows/build.yml
```diff
@@ -31,141 +31,36 @@ env:
...
+    outputs:
+      variants: ${{ steps.get-variants.outputs.variants }}
+      aarch-enemies: ${{ steps.get-variants.outputs.aarch-enemies }}
+    steps:
+      - uses: actions/checkout@v3
+      - id: get-variants
+        name: Determine variants
+        run: |
+          cd variants
+          output="variants=$(ls -d */ | cut -d'/' -f 1 | grep -v shared | grep -v target | jq -R -s -c 'split("\n")[:-1]')"
```
jpculp: nit: you can combine the greps with `grep -vE '^(shared|target)$'`.

.github/workflows/build.yml
```diff
@@ -31,141 +31,36 @@ env:
...
+          echo $output
+          echo $output >> $GITHUB_OUTPUT
+          output="aarch-enemies=$(ls -d */ | cut -d'/' -f 1 | grep -e "\-dev" -e "metal" -e "vmware" | jq -R -s -c 'split("\n")[:-1] | [ .[] | {"variant": ., "arch": "aarch64"}]')"
```
jpculp: nit: you can consolidate the grep with `grep -E '(^(metal|vmware)|\-dev$)'` while making sure you're only filtering on prefixes and suffixes.

jpculp (approved):

.github/workflows/build.yml
stmcginnis: I was trying to keep things simple and clear, but I switch to be more concise.

jpculp (approved):

Issue #2855, opened 2023-03-06
Support and use v6.1 LTS kernel


stevehipwell:
**What I'd like:**
I'd like Bottlerocket to support the `v6.1` LTS kernel and start using it for new versions (once it's been tested). This would align well with AL2023 which it looks like will be using `v6.1` by default.

**Any alternatives you've considered:**
n/a

stmcginnis:
Hey @stevehipwell thanks for this issue. I have heard this being discussed, but I don't believe there has been an issue out there to track it happening. Let's use this issue for that. Thanks!

foersleo:
I have the base enablement for kernel 6.1 [on my fork](https://github.com/foersleo/bottlerocket/tree/kernel-6.1).

Still need to do some testing and we will also need to do some work for ensuring the nvidia drivers are working for the nvidia variants, but I am working on this.

foersleo:
The kernel update in itself seems to work fine. I have successfully done some minimal boot testing and had a look at how much the kernel config changes between 5.15 and 6.1, using a process similar to what we do with `tools/diff-kernel-config`:

```
diff_x86_64:     82 removed, 235 added,   3 changed
diff_aarch64:    92 removed, 226 added,   2 changed
```

Open tasks that I will be working on to get this into the tree:

- [x] Review kernel config changes and see if we need to make any adjustments to what is new
- [x] More elaborate testing
- [x] Create `kmod-6.1-nvidia` package to enable platforms with nvidia drivers

foersleo:
PR for adding the nvidia driver for the new kernel: #3203

foersleo:
With the new kernel packages and accompanying nvidia drivers merged we are set to start using these kernels in the next upcoming variants.

The current plan is to have new variants using 6.1 kernel. The first candidates for that would be k8s-1.28 based variants, aimed to be released with 1.15.0 (https://github.com/bottlerocket-os/bottlerocket/issues/3169) tentatively aiming at August/Early September.

stmcginnis:
Looks like everything is addressed now, so closing out. Feel free to open a new issue if there are any other follow on asks. Thanks!

Issue #2947, opened 2023-03-29
eBPF builds fail in images with the 5.10 kernel in aarch64 


arnaldo2792:
**Image I'm using:**
Any image with the 5.10 kernel in aarch64

**What I expected to happen:**
I can compile the eBPF probe regardless of the kernel / architecture combo

**What actually happened:**
The compilation failed for the 5.10 kernel in aarch64

**How to reproduce the problem:**
Attempt to compile the falco eBPF probe in an image with the 5.10 kernel in aarch64. You will see an error similar to this:

```
falco-driver-loader * Trying to compile the eBPF probe (falco_bottlerocket_5.10.165_1_1.13.1-aws.o)
falco-driver-loader Warning: Kernel ABI header at 'tools/arch/arm64/include/asm/insn.h' differs from latest version at 'arch/arm64/include/asm/insn.h'
falco-driver-loader Warning: Kernel ABI header at 'tools/arch/arm64/lib/insn.c' differs from latest version at 'arch/arm64/lib/insn.c'
falco-driver-loader gcc: error: unrecognized command line option '-mbranch-protection=pac-ret+leaf'
falco-driver-loader gcc: error: unrecognized command line option '-fpatchable-function-entry=2'
falco-driver-loader make[2]: *** [scripts/Makefile.build:286: scripts/mod/empty.o] Error 1
falco-driver-loader make[1]: *** [Makefile:1681: modules_prepare] Error 2
falco-driver-loader make: *** [Makefile:38: all] Error 2
falco-driver-loader mv: cannot stat '/usr/src/falco-4.0.0+driver/bpf/probe.o': No such file or directory
falco-driver-loader Unable to load the falco eBPF probe
```

stmcginnis:
@markusboehme - I seem to remember you being involved in investigating something related to this, but I can't find that now. Do you know if this is still an issue? I think I saw something in the #falco channel in the k8s slack that someone stated they were able to build an aarch64 now.

markusboehme:
@stmcginnis Sorry I had missed this until now! The last thing related to eBPF I debugged probably was #2504 which is unrelated. In particular the gcc errors make me believe the wrong compiler is being used to compile the eBPF probe, i.e. not the one from the SDK (or rather not the one from the toolchain produced by Buildroot; the SDK has another gcc as part of its base image).

The reproduction steps in this issue don't provide enough context for me to reproduce yet, so I'll have to read up on this. Assigning to me for now.

[...]

Issue #2685, opened 2022-12-24
Bottlerocket v1.24 is not able to mount fsx for lustre filesystem


wael-sadek:
**Image I'm using:**
v1.24.6-eks-4360b32

Arch: x86_64
Kubernetes: v1.24
aws-fsx-csi-driver:v0.9.0

**What I expected to happen:**
Be able to mount fsx for lustre filesystem.

**What actually happened:**
Both the pod and the fsx csi driver reported the below errors when the pod was trying to mount the filesystem.

```
E1223 21:41:09.905407       1 driver.go:86] GRPC error: rpc error: code = Internal desc = Could not mount "fs-070a48fa26621de87.fsx.eu-central-1.amazonaws.com@tcp:/knkipbmv" at "/var/lib/kubelet/pods/ede9090b-0fc4-4cfc-9398-711fc923601f/volumes/kubernetes.io~csi/pvc-workbench-home-fsx/mount": mount failed: exit status 19
Mounting command: mount
Mounting arguments: -t lustre -o flock fs-070a48fa26621de87.fsx.eu-central-1.amazonaws.com@tcp:/knkipbmv /var/lib/kubelet/pods/ede9090b-0fc4-4cfc-9398-711fc923601f/volumes/kubernetes.io~csi/pvc-workbench-home-fsx/mount
Output: mount.lustre: mount fs-070a48fa26621de87.fsx.eu-central-1.amazonaws.com@tcp:/knkipbmv at /var/lib/kubelet/pods/ede9090b-0fc4-4cfc-9398-711fc923601f/volumes/kubernetes.io~csi/pvc-workbench-home-fsx/mount failed: No such device
Are the lustre modules loaded?
Check /etc/modprobe.conf and /proc/filesystems
```

**How to reproduce the problem:**
Use bottlerocket v1.24 with fsx csi driver and mount an fsx for lustre filesystem.

However bottlerocket v1.23 works fine.

jpmcb:
Thanks for raising this issue - I'm not super familiar with this CSI driver, but I'm wondering if it's related to this issue in the upstream kubernetes SIG?

https://github.com/kubernetes-sigs/aws-fsx-csi-driver/issues/243

The error: 

> Are the lustre modules loaded?

Makes me think that the mount is attempted before the daemonset /modules are ready. Are you able to apply a taint that will wait for the daemonset to be ready?

I can also spin up a quick cluster and give this a try.

wael-sadek:
[...]

PR #3121, opened 2023-05-16
Add kernel-6.1 sources and use it for -dev variants

foersleo:
**Issue number:** https://github.com/bottlerocket-os/bottlerocket/issues/2855

**Description of changes:**

Add the sources for Linux kernel-6.1 and enable it for all the dev variants.

The following todo list does not block review of what is already there (config analysis) and will be added soon.
TODO:

- [x] Adjust SELinux policies for new classes `user_namespace` and `io_uring` (See Testing section below)
- [x] Update lookaside cache
- [x] potentially disable candidate options (See Testing section below)
- [x] Test build of an external kmod using kmod-kit

**Testing done:**

* Launch a `aws-dev` image and see it booting successfully with the correct kernel:
```
bash-5.1# uname -a
Linux ip-10-1-171-244.eu-central-1.compute.internal 6.1.19 bottlerocket-os/bottlerocket#1 SMP PREEMPT_DYNAMIC Mon May 15 16:07:55 UTC 2023 x86_64 GNU/Linux
```

* Check for missing SELinux classes (See todo above):
```
May 16 13:22:07 localhost kernel: SELinux:  Class io_uring not defined in policy.
May 16 13:22:07 localhost kernel: SELinux:  Class user_namespace not defined in policy.
May 16 13:22:07 localhost kernel: SELinux: the above unknown classes and permissions will be denied
May 16 13:22:07 localhost kernel: SELinux:  policy capability network_peer_controls=1
May 16 13:22:07 localhost kernel: SELinux:  policy capability open_perms=1
May 16 13:22:07 localhost kernel: SELinux:  policy capability extended_socket_class=1
May 16 13:22:07 localhost kernel: SELinux:  policy capability always_check_network=0
May 16 13:22:07 localhost kernel: SELinux:  policy capability cgroup_seclabel=1
May 16 13:22:07 localhost kernel: SELinux:  policy capability nnp_nosuid_transition=1
May 16 13:22:07 localhost kernel: SELinux:  policy capability genfs_seclabel_symlinks=1
May 16 13:22:07 localhost kernel: SELinux:  policy capability ioctl_skip_cloexec=0
May 16 13:22:07 localhost systemd[1]: Successfully loaded SELinux policy in 324.808ms.
```

* check for config changes against existing 5.15 kernels on all dev variants:
```
config-aarch64-aws-dev-diff:	 89 removed, 221 added,   4 changed
config-aarch64-metal-dev-diff:	 89 removed, 222 added,   4 changed
config-x86_64-aws-dev-diff:	 82 removed, 232 added,   6 changed
config-x86_64-metal-dev-diff:	 82 removed, 234 added,   6 changed
config-x86_64-vmware-dev-diff:	 98 removed, 230 added,   7 changed
```

As expected with a major kernel update there is lots of changes to the kernel config. I have tried to organize them into buckets for easier consumption and review. All changes that can be mapped to a specific commit and upstream series have been annotated with the corresponding data and a short summary has been added to all options.

The change buckets are (ordered by decreasing interest):

* **Candidates for disabling** (6/468 config changes). These are candidates for explicit disabling. I currently do not really see a good use case for damon in Bottlerocket, so saving the build cycles might be a good idea. Similarly we should probably disable the advanced driver interfaces for `ICE` as we have done for other NIC drivers. Details in [diff_disable.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_disable-csv)
* **Downstream config changes Amazon Linux did** (32/468 config changes). These are mainly due to dropping downstream backports of some drivers and a downstream port of lustreFS as well as some smaller config changes. Non of these should be problematic. Details are in [diff_al.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_al-csv)
* **Changes that are not needed right away but might be interesting** (9/468 config changes). This includes some options around FIPS related proc fs files to advertise certain fips properties, some net vendors that provide modern 10G+ NICs that might be relevant for metal and the audit driver to add DM_INTEGRITY logging to the audit subsystem. Details in [diff_interesting.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_interesting-csv).
* **Unexplainable changes** (5/468 config changes). These are config changes we are seeing between 5.15 and 6.1 for architectures where these options should not have been available or set to the new value all along through automatic selection or architecture dependent invisibility. I have no good explanation for these. Details in [diff_wat.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_wat-csv).
* **Architecture and toolchain specific changes** (72/468 config changes). These are config changes where either a differentiation is made based on the architecture or where the used build toolchain will have an impact on how certain features can be used or have to be configured. Nothing we can or would do about these. Details can be found in [diff_architecture.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_architecture-csv) and [diff_toolchain.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_toolchain-csv).
* **New drivers that are not needed** (85/468 config changes). These are drivers for various hardware that we do not need. A lot of these is for certain embedded applications. Nothing really interesting in this bucket. Details can be found in [diff_new_drivers_not_needed.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_new_drivers_not_needed-csv).
* **New functionality I am confident we do not need** (47/468 config changes). This is a mix of various new functionalities, from debug features to certain subsystems to new counters that can be exposed through sysfs. I think that non of these would be relevant for Bottlerocket use cases. Details can be found in [diff_new_not_needed.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_new_not_needed-csv).
* **General config cleanup** (197/468). This is a mix of various janitorial changes to the kernels config system. This includes removals due to deprecation of features, renaming of config options, changing of visibility through changes in dependencies, refactoring with splitting/merging certain options. Details can be found in [diff_config_cleanup.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_config_cleanup-csv).
* **The uninteresting remainder** (15/468). A bunch of options I could not really match to any of the other buckets, but deemed not really interesting. Details in [diff_uninteresting.csv](https://gist.github.com/foersleo/87a42ae412aed31591e695a1ad7b0fec#file-diff_uninteresting-csv).

**Terms of contribution:**


foersleo:
The three failing checks are expected to fail with me not having uploaded the srpm to the lookaside cache. I am hesitant to do that until we have some more clarity on pushing this out.

**SELinux**

I will have to do some extra dive there. I do not have a lot of experience on getting these together and both new classes seem to be not similar to any other classes to model them after that.

The new classes have been introduced as follows:
* [740b03414b20e7 - selinux: add support for the io_uring access controls](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=740b03414b20e7f1879cd99aae27d8c401bbcbf9) in 5.16. This implements two access controls, `override_creds` to allow impersonating another domain in io_uring operations, and `sqpoll` to restrict which domains can create io_uring polling threads.
* [ed5d44d42c95e8 - selinux: Implement userns_create hook](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=ed5d44d42c95e8a13bb54e614d2269c8740667f9) in 6.1. This implements one new access control, `create` to restrict which domains can create user namespaces.

While poking around I have also found that the `lockdown` selinux implementation has been removed through [f5d0e5e9d72d3a - selinux: remove the SELinux lockdown implementation](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f5d0e5e9d72d3a06018efbfa3adccc0e09a129f9) in 5.16. As far as I can tell we have not done any particular policies for the lockdown selinux implementation, so just removing them should be fine. (Edit: However, given that we still have two kernels that have the lockdown selinux support we probably need to keep it in until we retire 5.10 and 5.15 at some point in the future).

foersleo:
Added the new SELinux classes. I am not quite sure if we would want to further restrict the access/use of these generally. Particularly the user_namespace one might be interesting here.

arnaldo2792:
For orchestrated containers, the SELinux policies permissions are "relaxed", meaning that they are granted almost all permissions since we don't know what users are doing in their workloads. For `io_uring` and `user_namespaces`, I expect we want the same, so you may have to add something like this:

https://github.com/bottlerocket-os/bottlerocket/commit/4e09a949c86d95bb63746f44eadafb164b33a3b4

markusboehme:
I've got to review this in batches, but you made it easy to do so. :-)

* **Candidates for disabling (6/468 config changes)**: agree with those. Some of the DAMON features (LRU list sorting, reclaim) would be interesting if we could benchmark those. No harm done in leaving them out for now, they'd need to be opted into at boot time anyway.
* **Downstream config changes Amazon Linux did (32/468 config changes)**: A few thing that stand out here:
  * `NET_VENDOR_MELLANOX`: The Mellanox drivers are needed. We enabled the MLX5 driver on user request (commit 2e5458b72f89) in addition to the MLX4 driver.
  * `LUSTREFSX_FS`: Same goes for Lustre support. I am surprised Amazon Linux dropped this. We were initially missing it for the 5.15 kernel, same as Amazon Linux, and it was noticed (#2685). [Amazon FSx for Lustre](https://aws.amazon.com/fsx/lustre/) needs this.
  * `FS_ENCRYPTION`: Somewhat on the fence about including this, leaning towards not doing so. Not all filesystems support this, of the mainstream ones only ext4 does. Will offering this tie us to ext4 for the data partition? If used on additional disks (ephemeral disks, EBS volumes), EBS volume encryption or block-level encryption with dm-crypt might be more universal choices. 
* **Changes that are not needed right away but might be interesting (9/468 config changes)**: Looks good to me.

foersleo:
> I've got to review this in batches, but you made it easy to do so. :-)
> 
> [...]
> * **Downstream config changes Amazon Linux did (32/468 config changes)**: A few thing that stand out here:
>   * `NET_VENDOR_MELLANOX`: The Mellanox drivers are needed. We enabled the MLX5 driver on user request (commit 2e5458b72f89) in addition to the MLX4 driver.

This one only has an impact on the VMware variant. For metal and AWS variants we explicitly set the ones we need there. The table shows an x only in the vmware-variant's column. If we need this one for vmware, we can add it back. Given that we did not explicitly enable it there before I would have thought we are good leaving it out.


>   * `LUSTREFSX_FS`: Same goes for Lustre support. I am surprised Amazon Linux dropped this. We were initially missing it for the 5.15 kernel, same as Amazon Linux, and it was noticed (#2685). [Amazon FSx for Lustre](https://aws.amazon.com/fsx/lustre/) needs this.

Good point, will have a discussion with Al team on why they dropped this.

>   * `FS_ENCRYPTION`: Somewhat on the fence about including this, leaning towards not doing so. Not all filesystems support this, of the mainstream ones only ext4 does. Will offering this tie us to ext4 for the data partition? If used on additional disks (ephemeral disks, EBS volumes), EBS volume encryption or block-level encryption with dm-crypt might be more universal choices. 

Working through all these i was under the impression that it would allow to encrypt and not force encryption by default. Re-reading this now I come to the same conclusions as you.

markusboehme:
> > * **Downstream config changes Amazon Linux did (32/468 config changes)**: A few thing that stand out here:
> >   
> >   * `NET_VENDOR_MELLANOX`: The Mellanox drivers are needed. We enabled the MLX5 driver on user request (commit [2e5458b](https://github.com/bottlerocket-os/bottlerocket/commit/2e5458b72f89f25ef709dc53b345bbf21da5d008)) in addition to the MLX4 driver.
> 
> This one only has an impact on the VMware variant. For metal and AWS variants we explicitly set the ones we need there. The table shows an x only in the vmware-variant's column. If we need this one for vmware, we can add it back. Given that we did not explicitly enable it there before I would have thought we are good leaving it out.

Agree, everything is as it should be. I totally missed this only affects the VMware build.


> > * `FS_ENCRYPTION`: Somewhat on the fence about including this, leaning towards not doing so. Not all filesystems support this, of the mainstream ones only ext4 does. Will offering this tie us to ext4 for the data partition? If used on additional disks (ephemeral disks, EBS volumes), EBS volume encryption or block-level encryption with dm-crypt might be more universal choices.
> 
> Working through all these i was under the impression that it would allow to encrypt and not force encryption by default. Re-reading this now I come to the same conclusions as you.

Yes, that's the case: Users can opt into encrypting individual directories if the filesystem supports this. Since ext4 (the current file system for the data partition) supports per-file encryption but e.g. XFS does not, I was worried that exposing this feature might get users stuck with ext4. I since looked into this some more and per-file encryption does not work through overlayfs (yet?). That lessens the concern somewhat, though it may at become relevant again in the future. For situations where the filesystem is under user control, like ephemeral disks or EBS volumes, dm-crypt seemed like a better choice.

**Unexplainable changes (5/468 config changes)**:

* `HAVE_LIVEPATCH` and related changes: Those were downstream Amazon Linux patches. Kernel live patching on Arm is not supported with the upstream kernel. While there have been patches, how to approach it is still being discussed. This will be a topic at this year's LPC as well. I'm surprised to see these missing for the 6.1 kernel, though. The documentation for Amazon Linux 2023 [states it is supported](https://docs.aws.amazon.com/linux/al2023/ug/live-patching.html).
* `MLX_PLATFORM`: :shrug: 
* `ARCH_NR_GPIO` on Arm: For 64-bit Arm this defaulted to 512 using an arch-generic definition on v5.15; only 32-bit Arm knew `ARCH_NR_GPIO`. This [changed in v5.19](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=5028fbad2d579). A value of 0 means it's still defaulting back to the arch-generic definition.

**Architecture and toolchain specific changes (72/468 config changes)**:

* `ARCH_WANTS_THP_SWAP`: `THP_SWAP` looks interesting, and apparently is an improvement on x86_64 as well. We're not currently using swap (IIRC there's an issue wishing for it), but might be something to revisit when that changes.
* `HAVE_STACK_VALIDATION`: Used to be set on Arm due to support for kernel live patching.

**New drivers that are not needed (85/468 config changes)~Architecture and toolchain specific changes (72/468 config changes)~**: Looks good to me!

markusboehme:
**New functionality I am confident we do not need (47/468 config changes):** Looks good to me.

**General config cleanup (197/468):**

* `VIVALDI_FMAP`: Nothing to be done for us here, but that upstream change is questionable. To save 300 bytes in an all-yes build, we now have a new module for a single small function due to us building `KEYBOARD_ATKBD` as a module.
* `XEN_PCI_STUB`: This just made me notice that we can `XEN_BACKEND=n` to cut out a bunch of code to act as a Xen dom0 or driver domain. Let's do this after the merge.

**The uninteresting remainder (15/468):** Looks good to me.

packages/kernel-6.1/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "kernel-6_1"
+version = "0.1.0"
+edition = "2018"
```
markusboehme: Commit b670aab01626 bumped this to 2021 tree-wide.

packages/kernel-6.1/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[package.metadata.build-package]
+variant-sensitive = "platform"
+package-name = "kernel-6.1"
+
+[lib]
+path = "pkg.rs"
+
+[[package.metadata.build-package.external-files]]
+# Use latest-srpm-url.sh to get this.
+url = "https://cdn.amazonlinux.com/al2023/core/guids/c21f02b8b6bcdc9593b87e257d8839676b3e1bd5ac90e4a9cabdceada709cc2a/SRPMS/../../../../blobstore/7da4ac2135147d0ab97afaa82b81cb7408dc5a2fa8f8dd3e785a908585977bd3/kernel-6.1.19-30.43.amzn2023.src.rpm"
```
markusboehme: GRUB's `latest-srpm-url.sh` tries to canonicalize the URL: https://github.com/bottlerocket-os/bottlerocket/blob/0888cb29aa94cef59fa74d88e9e9ec6988b6a3ce/packages/grub/latest-srpm-url.sh

This should avoid the URL changing when there's a repo push without a kernel update.

packages/selinux-policy/class.cil
```diff
@@ -238,6 +238,7 @@
 (class bpf (map_create map_read map_write prog_load prog_run))
 (class perf_event (open cpu kernel tracepoint read write))
 (class lockdown (integrity confidentiality))
+(class io_uring (override_creds sqpoll))
```
markusboehme: This is missing the `cmd` permission introduced in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f4d653dcaa4e4

packages/kernel-6.1/config-bottlerocket-metal
```diff
@@ -0,0 +1,121 @@
+# This file holds all the settings that are specific to hardware enablement
+# we do for the metal variants.
```
markusboehme, line 2: This metal config loses the `SCSI_VIRTIO` change from 29e818d87870

packages/selinux-policy/systems.cil
```diff
@@ -21,5 +21,7 @@
 (classmapping systems use (key (all)))
 (classmapping systems use (lockdown (all)))
 (classmapping systems use (perf_event (all)))
+(classmapping systems use (io_uring (all)))
+(classmapping systems use (user_namespace (all)))
```
markusboehme: Nit: could keep those alphabetically sorted with the others in that section

markusboehme (commented):
Some additional notes:

* I tested kdump still works.
* Would be good to test building an out-of-tree module using the still works.
* I'll have to defer judgement on the semantics of the SELinux changes, alas. Is my interpretation correct that everything is being allowed to use the mentioned io_uring features and create user namespaces (still blocked via sysctl)?

packages/kernel-6.1/1003-initramfs-unlink-INITRAMFS_FORCE-from-CMDLINE_-EXTEN.patch
```diff
@@ -0,0 +1,51 @@
+From 002d1909e18b7ca876edd4680ffcf8b59dea6c1b Mon Sep 17 00:00:00 2001
+From: Ben Cressey <bcressey@amazon.com>
+Date: Tue, 18 Oct 2022 22:24:52 +0000
+Subject: [PATCH 3/3] initramfs: unlink INITRAMFS_FORCE from
```
bcressey: Do we want to bring back the patch numbering? We dropped the numbers for 5.10 and 5.15.

packages/kernel-6.1/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition = "2018"
+publish = false
+build = "build.rs"
+
+[package.metadata.build-package]
+variant-sensitive = "platform"
+package-name = "kernel-6.1"
+
+[lib]
+path = "pkg.rs"
```
bcressey: Assuming #3148 merges first:
```suggestion
build = "../build.rs"

[package.metadata.build-package]
variant-sensitive = "platform"
package-name = "kernel-6.1"

[lib]
path = "/dev/null"
```

(And then remove `pkg.rs`).

packages/kernel-6.1/kernel-6.1.spec
```diff
@@ -0,0 +1,272 @@
...
+# Create a tarball of the same files, for use outside the running system.
+# In theory we could extract these files with `unsquashfs`, but we do not want
+# to require it to be installed on the build host, and it errors out when run
+# inside Docker unless the limit for open files is lowered.
+tar cf kernel-devel.tar src_squashfs/%{version} --transform='s|src_squashfs/%{version}|kernel-devel|'
+xz -T0 kernel-devel.tar
+
+install -D kernel-devel.squashfs %{buildroot}%{_cross_datadir}/bottlerocket/kernel-devel.squashfs
+install -D kernel-devel.tar.xz %{buildroot}%{_cross_datadir}/bottlerocket/kernel-devel.tar.xz
+install -d %{buildroot}%{kernel_sourcedir}
```
bcressey, line 224: Historically I used Falco to test whether this kernel-devel archive worked, but #2947 suggests it may fail for unrelated reasons. The `kmod-*-nvidia` package also depends on the "archive" version working as expected, but I believe @arnaldo2792 mentioned a newer NVIDIA driver might be required.

If it's useful, here's my original [test case for kmod kits](https://gist.github.com/bcressey/969797cd544038311b73a6498e70d241) - of which only the ZFS module might still be relevant.

It expects the kmod kit built by the build system to be in the same directory, and (optionally, commented out) a `kdk.tar` with supplemental files taken from a Linux kernel checkout, in case you run into missing files.

packages/selinux-policy/systems.cil
```diff
@@ -21,5 +21,7 @@
 (classmapping systems use (key (all)))
 (classmapping systems use (lockdown (all)))
 (classmapping systems use (perf_event (all)))
+(classmapping systems use (io_uring (all)))
```
bcressey: `io_uring (cmd)` seems similar to `file (ioctl)` and could also go into the `files (load)` mapping. 

`io_uring (sqpoll)` and `io_uring (override_creds)` are trickier since they involve credentials.

`io_uring (sqpoll)` appears to be "reflexive" in that `selinux_uring_sqpoll()` only ever checks whether the current sid has permission for the current sid.

There are blanket rules in `rules.cil` covering these two "self" cases:
```
(allow all_s self (processes (interact)))
(allow all_s self (processes (transform)))
```

Of the two, `processes (interact)` is probably the best fit, since "transform" is more for SELinux-related label changes.

Since `io_uring (override_creds)` allows one domain to impersonate another, that one should go in the `processes (transform)` map.

packages/selinux-policy/systems.cil
bcressey: I'm sad about this, but I don't know what compatibility-breaking gremlins would be unleashed by restricting user namespace creation to privileged pods and tasks.

packages/kernel-6.1/1003-initramfs-unlink-INITRAMFS_FORCE-from-CMDLINE_-EXTEN.patch
foersleo: I seem to have forgotten passing the `-N` flag when adapting these patches and formatting them again. Will be dropped with the next iteration.

foersleo:
Above force push tended many of the comments. Still need to clean up the `io_uring` selinux policies according to ben's comment.

foersleo:
Latest force-push was to rebase on top of current `develop` branch. No changes to the feature branch.

packages/kernel-6.1/Cargo.toml
foersleo: Fixed in latest updates.

packages/kernel-6.1/Cargo.toml
foersleo: Changed the `latest-srpm-url.sh` to model similar behavior as the grub one, but using al2023 instead of 2022. Updated urls in `Cargo.toml` and `kernel-6.1.spec`

packages/selinux-policy/class.cil
foersleo: Hah, overlooked that one. I probably looked at the file at the revision that introduced the initial permissions for `io_uring`. Fixed in latest update.

packages/kernel-6.1/config-bottlerocket-metal
foersleo: Good catch! Fixed.

packages/selinux-policy/systems.cil
foersleo: I was so focused on retaining the order as it is defined in the kernel I did not see this list is alphabetically sorted. Fixed.

packages/kernel-6.1/1003-initramfs-unlink-INITRAMFS_FORCE-from-CMDLINE_-EXTEN.patch
foersleo: Fixed.

foersleo:
The new change disabling FS_ENCRYPTION had the expected results disabling `FS_ENCRYPTION` and removing the dependent config options:

```
$ cat diff-report 
==> 6.1_fixups/config-aarch64-aws-dev-diff <==
-FS_ENCRYPTION_ALGS y
-FS_ENCRYPTION_INLINE_CRYPT n
 FS_ENCRYPTION y -> n

==> 6.1_fixups/config-aarch64-metal-dev-diff <==
-FS_ENCRYPTION_ALGS y
-FS_ENCRYPTION_INLINE_CRYPT n
 FS_ENCRYPTION y -> n

==> 6.1_fixups/config-x86_64-aws-dev-diff <==
-FS_ENCRYPTION_ALGS y
-FS_ENCRYPTION_INLINE_CRYPT n
 FS_ENCRYPTION y -> n

==> 6.1_fixups/config-x86_64-metal-dev-diff <==
-FS_ENCRYPTION_ALGS y
-FS_ENCRYPTION_INLINE_CRYPT n
 FS_ENCRYPTION y -> n
```

packages/selinux-policy/systems.cil
foersleo: I am trying to wrap my head around these, so if I understand correctly we want to move these out of the `systems use` one and into more fine grained mappings. If I understand correctly we would want yo do something like the following?

> `io_uring (cmd)` seems similar to `file (ioctl)` and could also go into the `files (load)` mapping.

Add something like

```
(classmapping files load (io_uring (cmd)))
```

to either [files.cil - line 60](https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/selinux-policy/files.cil#L60) or a new file concerned with io_uring rules.

> `io_uring (sqpoll)` and `io_uring (override_creds)` are trickier since they involve credentials.
> 
> `io_uring (sqpoll)` appears to be "reflexive" in that `selinux_uring_sqpoll()` only ever checks whether the current sid has permission for the current sid.
> 
> There are blanket rules in `rules.cil` covering these two "self" cases:
> 
> ```
> (allow all_s self (processes (interact)))
> (allow all_s self (processes (transform)))
> ```
> 
> Of the two, `processes (interact)` is probably the best fit, since "transform" is more for SELinux-related label changes.
> 
> Since `io_uring (override_creds)` allows one domain to impersonate another, that one should go in the `processes (transform)` map.


Probably similar, but something like
```
(classmapping processes interact (io_uring (sqpoll)))
(classmapping processes transform (io_uring (override_creds)))
```

to either a suitable place in the current files or a new file for io_uring mappings.

Or am I completely misunderstanding what you are proposing?

packages/selinux-policy/systems.cil
markusboehme: Would right now not be the time if we want to (try to) lock this down? In the default configuration, user namespaces remain turned off on kernels 5.10, 5.15, and 6.1. Only kernel 6.1 make user namespace creation controllable via SELinux. The change in policy could be safely tried for the new variant (and weakened if need be) for users turning on user namespaces.

packages/kernel-6.1/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "kernel-6_1"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
```
markusboehme, line 6: The `build.rs` inside the `kernel-6.1` directory can now be dropped.

packages/selinux-policy/systems.cil
bcressey: Yup, that's what I had in mind - expanding the current maps in `files.cil` and `processes.cil` the way you've indicated here.

I might structure the changes to `files.cil` a bit differently for consistency with the rest of the mappings, e.g.:
```
(classmapping files load load_io_uring)
(classpermission load_io_uring)
(classpermissionset load_io_uring (
  io_uring (cmd))
```

It's more boilerplate than the implicit syntax, but I find it helpful to be able to see at a glance where some of the classes show up. E.g. "filesystem" is only in some of the maps, and "fd" is only in load.

packages/selinux-policy/systems.cil
bcressey: Effectively today we are defaulting to "all" and can't change that default in the SELinux policy without some way to opt into the old behavior.

To do this I think we'd want a setting, like `settings.kernel.allow-user-namespaces`, with two or three values:
* "all" - all processes are allowed to create user namespaces
* "privileged" - privileged processes are allowed to create user namespaces
* "trusted" - trusted processes are allowed to use create namespaces

Then we would conditionally load an SELinux module based on which field was set.

We would then set `user.max_user_namespaces` up from zero by default and rely on the setting to restrict or allow access as needed.

I don't think it has to happen now but it would be a nice way to let people opt into more fine-grained access to user namespaces.

bcressey:
This is probably on your radar, but we should also add the [qlen patch](https://github.com/bottlerocket-os/bottlerocket/commit/ddc075761b315d76f86668a9552bc524591aaf5e) to 6.1.

bcressey:
> **Candidates for disabling**

I'm fine with disabling these.

> **Downstream config changes Amazon Linux did**

No concerns as long as Lustre comes back.

> **Changes that are not needed right away but might be interesting**

The FIPS selftest sounds like something we might want or need. It might be OK if it has a negligible impact on boot time.

> **Unexplainable changes**

+1

> **Architecture and toolchain specific changes**

Should we be setting `PREEMPT_DYNAMIC` on arm64 now that it's supported?

> **New drivers that are not needed**

+1

> **New functionality I am confident we do not need**

`EFI_ZBOOT` sounds interesting if it lets us have a compressed arm64 image.

`MODULE_DECOMPRESS` seems good if it eventually lets us have more flexibility on the compression format we're able to use for our modules. I'm sad we had to roll back the zstd change because older versions of `kmod` in pods didn't support it, and it'd be nice for that limitation to go away.

> **General config cleanup**

It seems like we had `STACK_VALIDATION` before for arm64, and now we don't have its replacement `CONFIG_OBJTOOL`. Is this related to the lack of live-patching support - we're just missing the patches that would be needed to enable it for 6.1?

Do we need the `i915` driver? It seems like there was a lot of config churn from that - `ACPI_WMI`, `DRM_DISPLAY_HELPER`, etc.

> **The uninteresting remainder**

+1

If it's creating delta from the Amazon Linux kernel config, we could potentially drop these items:
```
CONFIG_SERIO_I8042=m
CONFIG_KEYBOARD_ATKBD=m
CONFIG_MOUSE_PS2=m
```

I.e. we don't need or want the modules if they're disabled altogether in recent AL kernels.

packages/selinux-policy/systems.cil
markusboehme: The mechanism you propose is much more polished than what I had in mind for now. Since the 6.1 kernel would only be included in new variants, and the SELinux policy change hence constrained to those, I saw this as a chance to try to this admittedly incompatible change. If there were users opting into user namespaces and using them from unprivileged pods, we'd hear during their upgrade testing. From your response I take it you're against such changes even across variant upgrades, though, which is fine--the proposal was somewhat brisk. :-)

markusboehme:
> `MODULE_DECOMPRESS` seems good if it eventually lets us have more flexibility on the compression format we're able to use for our modules. I'm sad we had to roll back the zstd change because older versions of `kmod` in pods didn't support it, and it'd be nice for that limitation to go away.

We had to roll this back due to old user space not supporting zstd. Unfortunately, even with `MODULE_DECOMPRESS` we need a (even more) recent user space, since it has to explicitly request kernel-side decompression. On top of that the kernel currently only supports xz and gzip here, but unlike widespread user space support this would at least be easily fixable ourselves.

foersleo:
Latest push includes following adjustments:

* remove `build.rs` from `packages/kernel-6.1`
* add `max_dgram_qlen` fix
* adjustments to `io_uring` selinux rules
* remove candidate kconfig options as discussed (see `diff-report` below as expected)

```
==> 6.1_fixups_2/config-aarch64-aws-dev-diff <==
-DAMON_DBGFS y
-DAMON_LRU_SORT y
-DAMON_PADDR y
-DAMON_RECLAIM y
-DAMON_SYSFS y
-DAMON_VADDR y
-PAGE_IDLE_FLAG y
 DAMON y -> n

==> 6.1_fixups_2/config-aarch64-metal-dev-diff <==
-DAMON_DBGFS y
-DAMON_LRU_SORT y
-DAMON_PADDR y
-DAMON_RECLAIM y
-DAMON_SYSFS y
-DAMON_VADDR y
-PAGE_IDLE_FLAG y
 DAMON y -> n
 ICE_SWITCHDEV y -> n

==> 6.1_fixups_2/config-x86_64-aws-dev-diff <==
-DAMON_DBGFS y
-DAMON_LRU_SORT y
-DAMON_PADDR y
-DAMON_RECLAIM y
-DAMON_SYSFS y
-DAMON_VADDR y
-PAGE_IDLE_FLAG y
 DAMON y -> n

==> 6.1_fixups_2/config-x86_64-metal-dev-diff <==
-DAMON_DBGFS y
-DAMON_LRU_SORT y
-DAMON_PADDR y
-DAMON_RECLAIM y
-DAMON_SYSFS y
-DAMON_VADDR y
-PAGE_IDLE_FLAG y
 DAMON y -> n
 ICE_HWTS y -> n
 ICE_SWITCHDEV y -> n
```

packages/kernel-6.1/kernel-6.1.spec
foersleo: So the build of zfs as a trial kmod works as long as I do not want to cross compile. Looking at what is contained in the kmod-kit I think cross-compilation is not supposed to work, but maybe I am missing something?

The dockerfile I have modeled after the one Ben had posted is here:
https://gist.github.com/foersleo/1d62d86a8272e9c0dc6500c92877ecb0

Looking at the kmod-kit with Markus, we think it is not supposed to support cross-compilation. Maybe something we should look into for future versions.

armenr:
Happy to stay completely out of the way and not pollute your PRs, but we're *very* excited for this. We've been waiting for a while to be able to properly use `BBR congestion control` for our workloads in EKS. 👏

If you need real-life customer testers, I volunteer.

packages/kernel-6.1/kernel-6.1.spec
bcressey: > Looking at the kmod-kit with Markus, we think it is not supposed to support cross-compilation. Maybe something we should look into for future versions.

That's right - cross-compilation isn't supported in kmod kits today.

bcressey (approved):

stmcginnis (approved):
Definitely not my area of expertise, but everything looks correct to me and at least no structural issues that I could see.

packages/kernel-6.1/kernel-6.1.spec
foersleo: Created follow-up issue to look into the future of kmod-kit: https://github.com/bottlerocket-os/bottlerocket/issues/3199

packages/selinux-policy/systems.cil
foersleo: Created follow-up issue to track working towards a proper solution on this: https://github.com/bottlerocket-os/bottlerocket/issues/3200

markusboehme (approved):
:sparkles: :penguin:

armenr:
Sorry for the potentially inappropriate post on this PR, but what's a good way to test the dev/6.x builds? 🫠

foersleo:
> Sorry for the potentially inappropriate post on this PR, but what's a good way to test the dev/6.x builds? 🫠

It is absolutely not inappropriate at all. We are extremely happy when people want to test and extend their helping hands in the project overall. Thank you for offering your time.

Unfortunately, we currently do not have a great way to supply you with AMIs for testing. Additionally the `*-dev` variants do not come with an orchestrator, so these are not super helpful if you want to use them in a kubernetes cluster.

I currently have a testing commit for nvidia-variants that aids me in testing the new kernel on those platforms for a subsequent PR enabling the new nvidia kmods. You should be able to do something similar to that to get a variant of your choice working with the 6.1 kernel *for testing*. You can find that commit [on my fork](https://github.com/foersleo/bottlerocket/commit/4c5d8cd817ac81ded69deb994c01946dedcda50d). Instructions for building are in [BUILDING.md](https://github.com/bottlerocket-os/bottlerocket/blob/develop/BUILDING.md).

Feel free to reach out if you have any further questions or need any assistance.

armenr:
Thank you @foersleo !!

PR #3125, opened 2023-05-17
Actions: Modularize common logic for reuse

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This modularizes some of the common setup pieces that we have to make it easier to reuse this functionality in multiple workflows without repeating ourselves. We can avoid needing to make the same updates in multiple places as we add other workflows that need to perform the same steps.

**Testing done:**

Testing changes in a personal repo to validate the concepts. Further testing will be provided by the results of this PR.

**Terms of contribution:**

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -1,0 +1,47 @@
+name: "Node setup"
+description: "Performs setup for caching and other common needs."
+inputs:
+  perform-cache-cleanup:
+    description: "Whether to perform cache cleanup"
+    required: false
+    default: false
+    type: boolean
+runs:
+  using: "composite"
+  steps:
+    - run: |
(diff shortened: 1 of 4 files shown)


.github/actions/setup-node/action.yml
```diff
@@ -0,0 +1,47 @@
+name: "Node setup"
```
jpmcb, line 1: This is really "workflow" setup, no? But I suppose in our space, "node" is abit overloaded. So this is probably fine!

jpmcb (approved):

.github/actions/setup-node/action.yml
stmcginnis: Yeah, I suppose it could be either. In this context, I had considered "node setup" to be configuring the host on which this action is running. I can change it if folks think a different term would be more clear.

etungsten (approved):

PR #3127, opened 2023-05-18
host-ctr: add support for pulling private ECR images from `il-central-1`

etungsten:
**Issue number:**
N/A

**Description of changes:**
```
    host-ctr: add temporary workaround for bypassing checks for parsing ECR ref
    
    In order to support pulling images from ECR repositories from
    il-central-1, we need to generate our own canonical ECR image reference
    since the region is not officially supported by aws-go-sdk yet.
```

**Testing done:**
I am able to pull images from `il-central-1`:
```
$ ./host-ctr -s /run/containerd/containerd.sock pull-image --source 777777777777.dkr.ecr.il-central-1.amazonaws.com/eks/pause:3.2-eksbuild.1
time="2023-05-17T18:13:49-07:00" level=info msg="Image does not exist, proceeding to pull image from source." ref="ecr.aws/arn:aws:ecr:il-central-1:777777777777:repository/eks/pause:3.2-eksbuild.1"
time="2023-05-17T18:13:49-07:00" level=info msg="pulling with Amazon ECR Resolver" ref="ecr.aws/arn:aws:ecr:il-central-1:777777777777:repository/eks/pause:3.2-eksbuild.1"
time="2023-05-17T18:13:54-07:00" level=info msg="pulled image successfully" img="ecr.aws/arn:aws:ecr:il-central-1:777777777777:repository/eks/pause:3.2-eksbuild.1"
time="2023-05-17T18:13:54-07:00" level=info msg="unpacking image..." img="ecr.aws/arn:aws:ecr:il-central-1:777777777777:repository/eks/pause:3.2-eksbuild.1"
time="2023-05-17T18:13:54-07:00" level=info msg="tagging image" img="777777777777.dkr.ecr.il-central-1.amazonaws.com/eks/pause:3.2-eksbuild.1"

$ ./host-ctr -s /run/containerd/containerd.sock pull-image --source 777777777777.dkr.ecr.il-central-1.amazonaws.com/bottlerocket-admin:v0.10.1
time="2023-05-17T18:12:52-07:00" level=info msg="pulling with Amazon ECR Resolver" ref="ecr.aws/arn:aws:ecr:il-central-1:777777777777:repository/bottlerocket-admin:v0.10.1"
time="2023-05-17T18:12:53-07:00" level=info msg="pulled image successfully" img="ecr.aws/arn:aws:ecr:il-central-1:777777777777:repository/bottlerocket-admin:v0.10.1"
time="2023-05-17T18:12:53-07:00" level=info msg="unpacking image..." img="ecr.aws/arn:aws:ecr:il-central-1:777777777777:repository/bottlerocket-admin:v0.10.1"
time="2023-05-17T18:12:53-07:00" level=info msg="tagging image" img="777777777777.dkr.ecr.il-central-1.amazonaws.com/bottlerocket-admin:v0.10.1"
```

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -25,8 +25,10 @@ import (
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/oci"
+	"github.com/containerd/containerd/reference"
 	"github.com/containerd/containerd/remotes/docker"
 	"github.com/containerd/containerd/runtime/v2/runc/options"
+	"github.com/opencontainers/go-digest"
 	runtimespec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -575,16 +577,85 @@ func cleanUp(containerdSocket string, namespace string, containerID string) erro
 	return nil
 }
 
+// parseImageURISpecialRegions mimics the parsing in ecr.ParseImageURI but constructs the canonical ECR references while
+// skipping certain checks. We only do this for special regions that are not yet supported by the aws-go-sdk.
+// Referenced source: https://github.com/awslabs/amazon-ecr-containerd-resolver/blob/a5058cf091f4fc573813a032db37a9820952f1f9/ecr/ref.go#L70-L71
+func parseImageURISpecialRegions(input string) (string, error) {
+	ecrRefPrefixMapping := map[string]string{
+		"il-central-1": "ecr.aws/arn:aws:ecr:il-central-1:",
+	}
+	// Matching on account, region
+	matches := ecrRegex.FindStringSubmatch(input)
+	if len(matches) < 3 {
+		return "", fmt.Errorf("invalid image URI: %s", input)
+	}
+	account := matches[1]
+	region := matches[2]
+
+	// Get the ECR image reference prefix from the AWS region
+	ecrRefPrefix, exist := ecrRefPrefixMapping[region]
+	if !exist {
+		return "", fmt.Errorf("unhandled special region in ECR image repo URL: %s", input)
+	}
+
+	// Need to include the full repository path and the imageID (e.g. /eks/image-name:tag)
+	tokens := strings.SplitN(input, "/", 2)
+	if len(tokens) != 2 {
+		return "", fmt.Errorf("invalid image URI: %s", input)
+	}
+	fullRepoPath := tokens[len(tokens)-1]
+	// Run simple checks on the provided repository.
+	switch {
+	case
+		// Must not be empty
+		fullRepoPath == "",
+		// Must not have a partial/unsupplied label
+		strings.HasSuffix(fullRepoPath, ":"),
+		// Must not have a partial/unsupplied digest specifier
+		strings.HasSuffix(fullRepoPath, "@"):
+		return "", errors.New("incomplete reference provided")
+	}
+
+	// Parse out image reference's to validate.
+	ref, err := reference.Parse("repository/" + fullRepoPath)
+	if err != nil {
+		return "", err
+	}
+	// If the digest is provided, check that it is valid.
+	if ref.Digest() != "" {
+		err := ref.Digest().Validate()
+		// Digest may not be supported by the client despite it passing against
+		// a rudimentary check. The error is different in the passing case, so
+		// that's considered a passing check for unavailable digesters.
+		//
+		// https://github.com/opencontainers/go-digest/blob/ea51bea511f75cfa3ef6098cc253c5c3609b037a/digest.go#L110-L115
+		if err != nil && err != digest.ErrDigestUnsupported {
+			return "", fmt.Errorf("%v: %w", "invalid image URI", err)
+		}
+	}
+
+	return ecrRefPrefix + account + ":repository/" + fullRepoPath, nil
+}
+
 // fetchECRImage does some additional conversions before resolving the image reference and fetches the image.
 func fetchECRImage(ctx context.Context, source string, client *containerd.Client, registryConfigPath string, fetchCachedImageIfExist bool) (containerd.Image, error) {
 	ref := source
+
 	ecrRef, err := ecr.ParseImageURI(ref)
 	if err != nil {
-		log.G(ctx).WithError(err).WithField("source", source).Error("failed to parse ECR reference")
-		return nil, err
+		// The parsing might fail if the AWS region is special, parse again with special handling:
+		log.G(ctx).WithError(err).WithField("source", source).Error("failed to parse ECR reference, trying fallback parsing for special regions")
+		var nestedErr error
+		ref, nestedErr = parseImageURISpecialRegions(ref)
+		if nestedErr != nil {
+			log.G(ctx).WithError(nestedErr).WithField("source", source).Error("fallback parsing for \"special\" ECR region failed")
+			// Return both errors
+			return nil, fmt.Errorf("%v - fallback error: %v", err, nestedErr)
+		}
+	} else {
+		ref = ecrRef.Canonical()
 	}
 
-	ref = ecrRef.Canonical()
 	log.G(ctx).
 		WithField("ref", ref).
 		WithField("source", source).
--- a/sources/host-ctr/cmd/host-ctr/main_test.go
+++ b/sources/host-ctr/cmd/host-ctr/main_test.go
@@ -160,3 +160,39 @@ func TestBadRegistryHosts(t *testing.T) {
 	_, err := f("docker.io")
 	assert.Error(t, err)
 }
+
+func TestEcrParserSpecialRegion(t *testing.T) {
+	test := struct {
+		name           string
+		ecrImgURI      string
+		expectedString string
+	}{
+
+		"Parse ECR repo URL for special-case region",
+		"111111111111.dkr.ecr.il-central-1.amazonaws.com/bottlerocket/container:1.2.3",
+		"ecr.aws/arn:aws:ecr:il-central-1:111111111111:repository/bottlerocket/container:1.2.3",
+	}
+
+	t.Run(test.name, func(t *testing.T) {
+		result, err := parseImageURISpecialRegions(test.ecrImgURI)
+		assert.NoError(t, err)
+		assert.Equal(t, test.expectedString, result)
+	})
+}
+
+func TestEcrParserUnsupportedSpecialRegions(t *testing.T) {
+	test := struct {
+		name           string
+		ecrImgURI      string
+		expectedString string
+	}{
+		"Unsupported special region",
+		"111111111111.dkr.ecr.outer-space.amazonaws.com/bottlerocket/container:1.2.3",
+		"",
+	}
+
+	t.Run(test.name, func(t *testing.T) {
+		_, err := parseImageURISpecialRegions(test.ecrImgURI)
+		assert.Error(t, err)
+	})
+}


sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -13,6 +12,11 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/containerd/containerd/reference"
+	"github.com/opencontainers/go-digest"
+
+	"fmt"
```
stmcginnis: nit: Import grouping got messed up from std, third party, local, but not a critical issue.

stmcginnis (approved):
Some formatting looks a little odd, but golangci-lint isn't complaining about it, so good enough for me!

sources/host-ctr/cmd/host-ctr/main.go
jpmcb: Weird that `goimports` doesn't re-arrange. I think since it got added to it's own block, goimports is treating it seperatly:

```bash
find . -name "*.go" -not -path "*/vendor/*" | xargs goimports -w
```

so moving `fmt` into the top block with the other standard library things and running goimports to get it sorted should fix it.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -575,16 +579,80 @@ func cleanUp(containerdSocket string, namespace string, containerID string) erro
...
+		//
+		// https://github.com/opencontainers/go-digest/blob/ea51bea511f75cfa3ef6098cc253c5c3609b037a/digest.go#L110-L115
+		if err != nil && err != digest.ErrDigestUnsupported {
+			return "", fmt.Errorf("%v: %w", "invalid image URI", err)
+		}
+	}
+
+	// Get the ECR image reference prefix from the AWS region
+	ecrRefPrefix := ecrRefPrefixMapping[region]
+	return ecrRefPrefix + account + ":repository/" + fullRepoPath, nil
```
jpmcb, line 637: The first test case I wrote fails where I expected it to just work:

```
❯ go test ./...
--- FAIL: TestEcrSpecialRegionsParser (0.00s)
    --- FAIL: TestEcrSpecialRegionsParser/Parse_amazon-ecr-resolver_region (0.00s)
        main_test.go:46:
                Error Trace:    /home/ubuntu/workspace/bottlerocket-os/bottlerocket/sources/host-ctr/cmd/host-ctr/main_test.go:46
                Error:          Not equal:
                                expected: "ecr.aws/arn:aws:ecr:us-west-2:111111111111:repository/bottlerocket/container:1.2.3"
                                actual  : "111111111111:repository/bottlerocket/container:1.2.3"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -ecr.aws/arn:aws:ecr:us-west-2:111111111111:repository/bottlerocket/container:1.2.3
                                +111111111111:repository/bottlerocket/container:1.2.3
                Test:           TestEcrSpecialRegionsParser/Parse_amazon-ecr-resolver_region
FAIL
FAIL    host-ctr/cmd/host-ctr   0.016s
FAIL
```

It looks like with a `us-west-2` region, which is bundled by `aws-sdk-go` and should be handled by the `amazon-ecr-resolver`, falls back to using our `ecrRegPrefixMapping` which does not have the `us-west-2` region key and returns an empty string, resulting in `111111111111:repository/bottlerocket/container:1.2.3`.

Is this the expected behavior?

jpmcb (commented):
I won't comment too much on the actual code since that was mostly lifted from `amazon-ecr-resolver`

This is a good opportunity to add some additional tests in `main_test.go`

```go
func TestEcrSpecialRegionsParser(t *testing.T) {
        tests := []struct {
                name           string
                ecrImgURI      string
                expectedErr    error
                expectedString string
        }{
                // The default case for normal regions handled by the aws-go-sdk
                {
                        "Parse amazon-ecr-resolver region",
                        "111111111111.dkr.ecr.us-west-2.amazonaws.com/bottlerocket/container:1.2.3",
                        nil,
                        "ecr.aws/arn:aws:ecr:us-west-2:111111111111:repository/bottlerocket/container:1.2.3",
                },

                // Regions parsed and handled by host-ctr
                {
                        "Parse special case region",
                        "111111111111.dkr.ecr.il-central-1.amazonaws.com/bottlerocket/container:1.2.3",
                        nil,
                        "ecr.aws/arn:aws:ecr:il-central-1:111111111111:repository/bottlerocket/container:1.2.3",
                },

                // Failure and edge cases
                {
                        "Fallback to public ECR on invalid region",
                        "111111111111.dkr.ecr.outer-space.amazonaws.com/bottlerocket/container:1.2.3",
                        nil,
                        "111111111111:repository/bottlerocket/container:1.2.3",
                },
        }

        for _, tc := range tests {
                t.Run(tc.name, func(t *testing.T) {
                        result, err := parseImageURISpecialRegions(tc.ecrImgURI)
                        assert.Equal(t, tc.expectedErr, err)
                        assert.Equal(t, tc.expectedString, result)
                })
        }
}
```

These don't capture all the edge cases, but it will definitely make it easier for other devs on the team to see what's going on with this function.

---

sources/host-ctr/cmd/host-ctr/main.go
jpmcb: We probably need to check the map access here?

```go
	val, ok := ecrRefPrefixMapping[region]
	if !ok {
                return "", fmt.Errorf("%v: %s", "invalid region in internal mapping", region)
	}
```

Otherwise we will always be masking unknown values in the prefix mapping by assuming an empty string.

Here's a little go playground to demonstrate this as well https://go.dev/play/p/W_8XzLuWOwp

jpmcb: This also makes me wonder what the actual expected behavior for completely invalid regions is:

```go
                {
                        "Fallback to public ECR on invalid region",
                        "111111111111.dkr.ecr.outer-space.amazonaws.com/bottlerocket/container:1.2.3",
                        nil,
                        "111111111111:repository/bottlerocket/container:1.2.3",
                },
```

So, for region `outer-space` in this test case, what's the expected result? It looks like the same thing is happening where the value from `ecrRefPrefixMapping` (which doesn't exist) is returning the empty string and is not populating the `ecrRefPrefix` portion of the returned image uri

jpmcb: We can probably disregard the error I'm seeing with this function on `us-west-2` since calling the `amazon-ecr-resolver` _shouldn't_ error out when called on those regions _before_ we make it to the host-ctr's `parseImageURISpecialRegions`

https://github.com/etungsten/bottlerocket/blob/cc6c6530311ec5be4d96bfe72c04688bf6f791de/sources/host-ctr/cmd/host-ctr/main.go#L642-L651

Although that makes me abit nervous since there's a clear failure path where a valid region is returned in error by the ecr-resolver and is not covered by our subsequent function call.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -575,16 +579,80 @@ func cleanUp(containerdSocket string, namespace string, containerID string) erro
...
-		log.G(ctx).WithError(err).WithField("source", source).Error("failed to parse ECR reference")
-		return nil, err
+		var nestedErr error
+		// The parsing might fail if the AWS region is special, parse again with special handling:
+		ref, nestedErr = parseImageURISpecialRegions(ref)
+		if nestedErr != nil {
+			log.G(ctx).WithError(err).WithField("source", source).Error("failed to parse ECR reference")
+			// Return the original error
+			return nil, err
+		}
```
jpmcb: This never does anything with `nestedErr`.

`err` in this context is the error returned out of the ecr-resolver's `ParseImageURI` which we might expect to fail if we need to handle a special region.

It's unfortunate that the ecr resolver doesn't return typed errors (that we could check to _only_ do our own parsing when given a certain error). But we _should_ still handle all errors out of our own parsing function. So, instead, we maybe we return both errors constructed through a new error:

```go
if nestedErr != nil {
    log.G(ctx).WithError(err).WithField("source", source).Error("failed to parse ECR reference")
    log.G(ctx).WithError(nestedErr).WithField("source", source).Error("failed to parse special ECR reference")
    // Return both errors
    return nil, fmt.Errorf("%w: could not parse special regions: %w", err, nestedErr)
}
```

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -575,16 +579,80 @@ func cleanUp(containerdSocket string, namespace string, containerID string) erro
 	return nil
 }
 
+// parseImageURISpecialRegions mimics the parsing in ecr.ParseImageURI but constructs the canonical ECR references while
+// skipping certain checks. We only do this for special regions that are not yet supported by the aws-go-sdk.
+// Referenced source: https://github.com/awslabs/amazon-ecr-containerd-resolver/blob/a5058cf091f4fc573813a032db37a9820952f1f9/ecr/ref.go#L70-L71
+func parseImageURISpecialRegions(input string) (string, error) {
+	ecrRefPrefixMapping := map[string]string{
+		"il-central-1": "ecr.aws/arn:aws:ecr:il-central-1:",
+	}
```
yeazelm, line 586: I think I would prefer we keep these mappings in a file that we read in. This allows us to keep the code that solves this specific problem separate from a region's arrival in the upstream SDK. It would be even nicer if it was a general purpose file that other tools that need endpoints could use. This is our only use we have found but it's conceivable other things would need similar "overrides". We would probably need the format of the file to be more generic: 

```
{
    "endpoints": [
        "il-central-1": {
            "ecrPrefix":  "ecr.aws/arn:aws:ecr:il-central-1:",
            ....
        }
    ]
} 
```
This is purely a nice optimization and I don't think we have to do this now, but it would make this code a little more maintainable since the JSON file would change periodically, but we don't have to _hopefully_ worry about this code block in the future. [This](https://github.com/aws/aws-sdk-go/blob/main/models/endpoints/endpoints.json) is more or less how the rest of the AWS SDK's like to think about regional endpoints.

sources/host-ctr/cmd/host-ctr/main.go
etungsten: I actually had to run `goimports` to placate the linter and this is what the tool gave me.

etungsten:
We've decided to take a different approach to this, so taking this PR back into draft.

sources/host-ctr/cmd/host-ctr/main.go
etungsten: > It looks like with a us-west-2 region, which is bundled by aws-sdk-go and should be handled by the amazon-ecr-resolver, falls back to using our ecrRegPrefixMapping which does not have the us-west-2 region key and returns an empty string, resulting in 111111111111:repository/bottlerocket/container:1.2.3.
Is this the expected behavior?

Yes, that's expected behavior. The fallback only triggers when `ecr.ParseImageURI` fails to parse out the ECR image source.


> So, for region outer-space in this test case, what's the expected result? It looks like the same thing is happening where the value from ecrRefPrefixMapping (which doesn't exist) is returning the empty string and is not populating the ecrRefPrefix portion of the returned image uri

We error since we're limiting the fallback to only `il-central-1` and no other "special" region that isn't yet supported by the `aws-go-sdk`. This is only a temporary workaround for `il-central-1` that is meant to removed when a more stable and long-term fix is available.

> Although that makes me abit nervous since there's a clear failure path where a valid region is returned in error by the ecr-resolver and is not covered by our subsequent function call.

If `ecr.ParseImageURI` errors for whatever reason on a ECR image URL belonging to some "valid" region (valid meaning supported by the `aws-go-sdk`. We will fail regardless in the fallback, so the end result is the same.

sources/host-ctr/cmd/host-ctr/main.go
etungsten: As we've discussed offline, this workaround is temporary and we have no intention of supporting this in the long term. We will attempt to fast track a more stable and long term fix for special non-SDK-supported regions.

etungsten:
Ditching this in favor of https://github.com/bottlerocket-os/bottlerocket/pull/3138

PR #3130, opened 2023-05-18
Fetch and populate partition and domain with `shibaken`

jpculp:
**Description:**

Fetches the partition and domain from IMDS meta-data using the `shibaken` settings generator.

**Testing done:**

- [x] Launch host in `aws`.
- [x] Launch host in `aws-us-gov`.
- [x] Launch host in `aws-cn.
- [ ] Migrate from existing host.

**Terms of contribution:**

### DIFF
--- a/Release.toml
+++ b/Release.toml
@@ -211,4 +211,8 @@ version = "1.15.0"
     "migrate_v1.14.0_aws-control-container-v0-7-2.lz4",
     "migrate_v1.14.0_public-control-container-v0-7-2.lz4",
 ]
-"(1.14.0, 1.15.0)" = []
+"(1.14.0, 1.14.1)" = [
+    "migrate_v1.14.1_aws-partition-setting.lz4",
+    "migrate_v1.14.1_shibaken-aws-partition.lz4",
+]
+"(1.14.1, 1.15.0)" = []
--- a/Release.toml
+++ b/Release.toml
@@ -214,5 +214,7 @@ version = "1.15.0"
 "(1.14.0, 1.14.1)" = [
     "migrate_v1.14.1_aws-partition-setting.lz4",
     "migrate_v1.14.1_shibaken-aws-partition.lz4",
+    "migrate_v1.14.1_aws-domain-setting.lz4",
+    "migrate_v1.14.1_shibaken-aws-domain.lz4",
 ]
 "(1.14.1, 1.15.0)" = []
--- a/sources/api/migration/migrations/v1.14.1/aws-partition-setting/src/main.rs
+++ b/sources/api/migration/migrations/v1.14.1/aws-partition-setting/src/main.rs
@@ -1,0 +1,18 @@
+use migration_helpers::common_migrations::AddSettingsMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting for AWS partition.
+fn run() -> Result<()> {
+    migrate(AddSettingsMigration(&["settings.aws.partition"]))
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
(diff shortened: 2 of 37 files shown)


jpculp:
Moved logic from `early-boot-config` to `shibaken`.

Release.toml
```diff
@@ -1,4 +1,4 @@
-version = "1.15.0"
+version = "1.14.1"
```
stmcginnis: We could either do this, or modify the cherry-pick back to `1.14.x`. I'll make a note to follow up to bump this back to 1.15.0 if we do this here. Only advantage is a cleaner backport.

sources/api/migration/migrations/v1.14.1/aws-partition-setting/src/main.rs
```diff
@@ -0,0 +1,24 @@
+use migration_helpers::common_migrations::{AddPrefixesMigration, AddSettingsMigration};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new setting for AWS partition.
+fn run() -> Result<()> {
+    if cfg!(variant_platform = "aws") {
+        migrate(AddSettingsMigration(&["settings.aws.partition"]))
+    } else {
+        // Non-AWS variants did not have any AWS setting until this point,
```
stmcginnis: That's not actually true anymore. Since v.1.11.0 we've needed `settings.aws` to support AWS IAM Roles Anywhere.

https://github.com/bottlerocket-os/bottlerocket/pull/2542/commits/cedda93d79c849948c063dccedc076d1a05580fe

sources/api/migration/migrations/v1.14.1/shibaken-aws-partition/src/main.rs
```diff
@@ -0,0 +1,26 @@
...
+/// We added a `setting-generator` for `settings.aws.partition` on AWS variants.
+/// This migration will do nothing on upgrade, but will remove the metadata if present on downgrade.
+fn run() -> Result<()> {
+    if cfg!(variant_platform = "aws") {
+        migrate(AddMetadataMigration(&[SettingMetadata {
+            setting: "settings.aws.partition",
+            metadata: &["setting-generator"],
+        }]))
+    } else {
+        migrate(NoOpMigration)
```
stmcginnis, line 14: This should be fine for non-AWS variants though, since there is no concept of a partition when it comes to IAM Roles Anywhere.

sources/api/migration/migrations/v1.14.1/aws-partition-setting/src/main.rs
jpculp: Ah, in that case I'll replace the else with a `NoOpMigration`.

sources/api/migration/migrations/v1.14.1/shibaken-aws-partition/src/main.rs
jpculp: Like the region behavior in `early-boot-config`, partition should fallback to `aws` if IMDS fails to pull one.

jpculp:
- Reverted `Release.toml` version back to `1.15.0`.
- Replaced else `AddPrefixesMigration(vec!["settings.aws"])` with `NoOpMigration`.

sources/api/migration/migrations/v1.14.1/aws-partition-setting/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+use std::process;
+
+/// We added a new setting for AWS partition.
+fn run() -> Result<()> {
+    if cfg!(variant_platform = "aws") {
+        migrate(AddSettingsMigration(&["settings.aws.partition"]))
+    } else {
+        migrate(NoOpMigration)
+    }
+}
```
bcressey: This is not correct since `settings.aws.partition` will exist in all variants since `AwsSettings` is not conditionally compiled.
```suggestion
/// We added a new setting for AWS partition.
fn run() -> Result<()> {
    migrate(AddSettingsMigration(&["settings.aws.partition"]))
}
```

sources/api/shibaken/README.md
```diff
@@ -11,6 +11,7 @@ shibaken is used to fetch data from the instance metadata service (IMDS) in AWS.
 shibaken can:
 * Fetch and populate the admin container's user-data with authorized ssh keys from the IMDS.
 * Perform boolean queries about the AWS partition in which the host is located.
+* Fetch and populate the 'aws.partition' setting.
```
bcressey, line 14: IMO we should go ahead and populate `aws.domain` as well while we're at it.

zmrow (approved):
Provided @bcressey 's feedback is addressed, this looks good.

jpculp:
- Fixed conditionally compiled migration.
- Added commits to support a domain setting.

jpculp:
I think I might need to tweak this a bit. I'll open a new PR when I do.

PR #3134, opened 2023-05-22
netdog: Add initial networkd config structs and associated `Display` macro

zmrow:
**Issue number:**
Related to #2449 

**Description of changes:**
This is the first in a series of changes to `netdog` that add the ability to generate `systemd-networkd` config files.  This PR adds the initial set of structs that represent the two kinds of `systemd-networkd` config files we currently use: `.network` and `.netdev`.  These structs contain the config entries we currently will make use of; it is expected they will grow over time.

These structs must be serialized to file.  Though a serde library exists for INI, `systemd` doesn't strictly follow INI spec as it allows for duplicate sections and duplicate keys within a section.  Rather than hack around the library, I opted to write a proc macro that generates `Display` implementations in the `systemd` unit format for the structs so the user can serialize the structs to string/file.  The proc macro allows for annotation of structs that represent config file sections, as well as fields that represent entries within the sections.  

This allows us to write the following:
```
// Very abbreviated example
#[derive(SystemdUnit)]
struct NetworkConfig {
    match_section: Option<MatchSection>
}

#[derive(SystemdUnitSection)
#[systemd(section = "Match")] // The name of the networkd config section
struct MatchSection {
    #[systemd(entry = "Name")] // The name of the networkd config entry
    name: Option<String>
}
```

which would represent the following section of a .network file:
```
[Match]
Name=...
```

**Testing done:**
Additional integration tests to come as the code is added to build these structs.

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/networkd/config/network.rs
+++ b/sources/api/netdog/src/networkd/config/network.rs
@@ -1,0 +1,123 @@
+use crate::interface_id::{InterfaceName, MacAddress};
+use ipnet::IpNet;
+use std::fmt::Display;
+use std::net::IpAddr;
+use systemd_derive::{SystemdUnit, SystemdUnitSection};
+
+#[derive(Debug, Default, SystemdUnit)]
+pub(crate) struct NetworkConfig {
+    r#match: Option<MatchSection>,
+    link: Option<LinkSection>,
+    network: Option<NetworkSection>,
+    route: Vec<RouteSection>,
+    dhcp4: Option<Dhcp4Section>,
+    dhcp6: Option<Dhcp6Section>,
+}
+
+#[derive(Debug, Default, SystemdUnitSection)]
+#[systemd(section = "Match")]
+struct MatchSection {
+    #[systemd(entry = "Name")]
+    name: Option<InterfaceName>,
+    #[systemd(entry = "PermanentMACAddress")]
+    permanent_mac_address: Vec<MacAddress>,
+}
+
+#[derive(Debug, Default, SystemdUnitSection)]
+#[systemd(section = "Link")]
+struct LinkSection {
+    #[systemd(entry = "RequiredForOnline")]
+    required: Option<bool>,
+    #[systemd(entry = "RequiredFamilyForOnline")]
+    required_family: Option<RequiredFamily>,
+}
+
+#[derive(Debug, Default, SystemdUnitSection)]
+#[systemd(section = "Network")]
+struct NetworkSection {
+    #[systemd(entry = "Address")]
+    addresses: Vec<IpNet>,
+    #[systemd(entry = "Bond")]
+    bond: Option<InterfaceName>,
+    #[systemd(entry = "ConfigureWithoutCarrier")]
+    configure_wo_carrier: Option<bool>,
+    #[systemd(entry = "DHCP")]
+    dhcp: Option<DhcpBool>,
+    #[systemd(entry = "LinkLocalAddressing")]
+    link_local_addressing: Option<DhcpBool>,
+    #[systemd(entry = "PrimarySlave")]
+    primary_bond_worker: Option<bool>,
+    #[systemd(entry = "VLAN")]
+    vlan: Vec<InterfaceName>,
+}
+
+#[derive(Debug, Default, SystemdUnitSection)]
+#[systemd(section = "Route")]
+struct RouteSection {
+    #[systemd(entry = "Destination")]
+    destination: Option<IpNet>,
+    #[systemd(entry = "Gateway")]
+    gateway: Option<IpAddr>,
+    #[systemd(entry = "Metric")]
+    metric: Option<u32>,
+    #[systemd(entry = "PreferredSource")]
+    preferred_source: Option<IpAddr>,
+}
+
+#[derive(Debug, Default, SystemdUnitSection)]
+#[systemd(section = "DHCPv4")]
+struct Dhcp4Section {
+    #[systemd(entry = "RouteMetric")]
+    metric: Option<u32>,
+    #[systemd(entry = "UseDNS")]
+    use_dns: Option<bool>,
+    #[systemd(entry = "UseDomains")]
+    use_domains: Option<bool>,
+}
+
+#[derive(Debug, Default, SystemdUnitSection)]
+#[systemd(section = "DHCPv6")]
+struct Dhcp6Section {
+    #[systemd(entry = "UseDNS")]
+    use_dns: Option<bool>,
+    #[systemd(entry = "UseDomains")]
+    use_domains: Option<bool>,
+}
+
+#[derive(Debug)]
+enum RequiredFamily {
+    Any,
+    Both,
+    Ipv4,
+    Ipv6,
+}
+
+impl Display for RequiredFamily {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            RequiredFamily::Any => write!(f, "any"),
+            RequiredFamily::Both => write!(f, "both"),
+            RequiredFamily::Ipv4 => write!(f, "ipv4"),
+            RequiredFamily::Ipv6 => write!(f, "ipv6"),
+        }
+    }
+}
+
+#[derive(Debug)]
+enum DhcpBool {
+    Ipv4,
+    Ipv6,
+    No,
+    Yes,
+}
+
+impl Display for DhcpBool {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            DhcpBool::Ipv4 => write!(f, "ipv4"),
+            DhcpBool::Ipv6 => write!(f, "ipv6"),
+            DhcpBool::No => write!(f, "no"),
+            DhcpBool::Yes => write!(f, "yes"),
+        }
+    }
+}
--- a/sources/api/netdog/systemd-derive/Cargo.toml
+++ b/sources/api/netdog/systemd-derive/Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "systemd-derive"
+version = "0.1.0"
+authors = ["Zac Mrowicki <mrowicki@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+build = "build.rs"
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[lib]
+path = "src/lib.rs"
+proc-macro = true
+
+[dependencies]
+darling = { version = "0.14", default-features = false }
+proc-macro2 = "1"
+quote = "1"
+syn = { version = "1", default-features = false, features = ["derive"] }
+
+[build-dependencies]
+generate-readme = { path = "../../../generate-readme", version = "0.1" }
--- a/sources/api/netdog/systemd-derive/README.md
+++ b/sources/api/netdog/systemd-derive/README.md
@@ -1,0 +1,129 @@
+# systemd-derive
+
+Current version: 0.1.0
+
+
+A macro to serialize structs to `systemd` unit file format
+
+### Description
+
+The `SystemdUnit` and `SystemdUnitSection` macros can be used to serialize structs representing
+`systemd` unit files.
+
+Under the hood, the macros implement `Display` for structs.  This allows converting the structs to
+a string suitable for writing directly to a file.
+
+The implementation is fairly rigid to the way the "INI-like" structure of `systemd` unit files is
+represented. `systemd` differs from standard INI format in that duplicate sections are allowed,
+along with duplicate keys within a section.  The macros expect there will be a "top-level" struct
+that represents the unit file, with nested structs representing the sections of said file. These
+nested structs representing sections have fields that are the configuration key/value pairs for
+that section.
+
+All struct fields must be either `Option`s or `Vec`s containing an object that implements
+`Display`.  Fields that are `Vec`s will be iterated upon and each value of the `Vec` will be
+serialized.  For structs deriving `SystemdUnit`, fields represent sections and therefore a `Vec`
+field would serialize as a repeated section. Structs deriving `SystemdUnitSection` have fields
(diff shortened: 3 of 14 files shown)


sources/api/netdog/networkd-derive/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[lib]
+path = "src/lib.rs"
+proc-macro = true
+
+[dependencies]
+darling = "0.14"
+proc-macro2 = "1"
+quote = "1"
+syn = {version = "1", default-features = false, features = ["full"]}
```
bcressey: Is it possible to refine the list of `darling` and `syn` features used? These dependencies tend to adversely affect compile times, and it would be nice to minimize that overhead.

sources/api/netdog/networkd-derive/tests/tests.rs
```diff
@@ -0,0 +1,78 @@
...
+struct MatchSection {
+    #[networkd(entry = "Name")]
+    name: Option<String>,
+}
+
+#[derive(Debug, Default, NetworkDSection)]
+#[networkd(section = "Network")]
+struct NetworkSection {
+    #[networkd(entry = "Address")]
+    addresses: Vec<String>,
```
yeazelm: Will we need `Gateway` for static addressing?

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -0,0 +1,81 @@
...
+    dhcp6_section: Option<Dhcp6Section>,
+}
+
+#[derive(Debug, Default, NetworkDSection)]
+#[networkd(section = "Match")]
+struct MatchSection {
+    #[networkd(entry = "Name")]
+    name: Option<String>,
+    #[networkd(entry = "PermanentMACAddress")]
+    permanent_mac_address: Option<String>,
```
yeazelm: I'm guessing we need to catch that only one of these is valid at a time?

sources/api/netdog/networkd-derive/tests/tests.rs
zmrow: This is just a test to ensure repeated entries work.

The `RouteSection` definition in `network.rs` includes `Gateway`. :)

sources/api/netdog/src/networkd/config/network.rs
zmrow: At this level we aren't doing any validation.  We ensure that an interface only has a name/MAC when we deserialize the network config provided by the user.

That being said - an upcoming PR has a constructor for this struct that can only accept one or the other.

stmcginnis (approved):

yeazelm (approved):
I didn't deeply review the proc_macro or darling usage but the Networkd specifics all look right!

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[derive(Debug, Default, NetworkDFile)]
+pub(crate) struct NetworkConfig {
+    match_section: Option<MatchSection>,
+    link_section: Option<LinkSection>,
+    network_section: Option<NetworkSection>,
+    route_sections: Vec<RouteSection>,
+    dhcp4_section: Option<Dhcp4Section>,
+    dhcp6_section: Option<Dhcp6Section>,
+}
```
bcressey: Kind of an aesthetic nit but I find the `_section` suffixes on the fields intrusive:
```suggestion
#[derive(Debug, Default, NetworkDFile)]
pub(crate) struct NetworkConfig {
    match: Option<MatchSection>,
    link: Option<LinkSection>,
    network: Option<NetworkSection>,
    routes: Vec<RouteSection>,
    dhcp4: Option<Dhcp4Section>,
    dhcp6: Option<Dhcp6Section>,
}
```

sources/api/netdog/src/networkd/config/network.rs
bcressey: I would like to see additional validation on these types, beyond just "String".

`PermanentMACAddress`  is also defined as a list of whitespaced-separated addresses, so it might be better to use `Option<Vec<MacAddress>>` here.

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[derive(Debug, Default, NetworkDSection)]
+#[networkd(section = "Link")]
+struct LinkSection {
+    #[networkd(entry = "RequiredForOnline")]
+    required: Option<String>,
+    #[networkd(entry = "RequiredFamilyForOnline")]
+    required_family: Option<String>,
+}
```
bcressey:
```suggestion
struct LinkSection {
    #[networkd(entry = "RequiredForOnline")]
    required: Option<bool>,
    #[networkd(entry = "RequiredFamilyForOnline")]
    required_family: Option<Family>,
}
```

sources/api/netdog/networkd-derive/README.md
```diff
@@ -0,0 +1,133 @@
+# networkd-derive
+
+Current version: 0.1.0
+
+
+A macro to serialize structs to `systemd-networkd` configuration format.
```
bcressey: Can this be recast as a macro to serialize structs into the systemd unit file format?

What is actually `systemd-networkd` specific about it apart from the names? `SystemdUnit` and `SystemdUnitSection` might work as well.

sources/api/netdog/networkd-derive/README.md
```diff
@@ -0,0 +1,133 @@
...
+#[networkd(section = "Route")]
+struct RouteSection {
+    #[networkd(entry = "Destination")]
+    destination: Option<String>,
+}
+```
+
+The following demonstrates instantiating an instance of the above structs and the resulting serialized form from calling `to_string()`.
+
+Note: The following is just an example to demonstrate the macro, it would be a very odd network configuration.
```
bcressey: It seems clear from context that this is just an example.

```suggestion
```

sources/api/netdog/src/networkd/config/network.rs
zmrow: Fair - I waffled on this a bit early on and settled on the `_section` prefixes to be perhaps overly descriptive.  I like them better without as well.

sources/api/netdog/src/networkd/config/network.rs
zmrow: I mention this in the docs for the macro but at this level, we're not doing any validation.  We validate input at the time we deserialize network config (`net.toml`), so at the point we are creating these files we can assume input is valid.

Good catch on the `PermanentMACAddress` bit: we don't currently allow multiple MACs in our config so I didn't account for it here.  I'd probably leave it alone until we allow for multiple MACs in config.

All that being said, if we intend to publish this macro/structs as a standalone crate somewhere I would absolutely agree about the validation of types.  (There would be a bit more work to do to publish this; addition of all the settings entries, etc)  At the current time, however, it fulfills our rigid use case well and avoids all the extra code repetition.

sources/api/netdog/src/networkd/config/network.rs
zmrow: Related to the above discussion - given that we validate input, I decided against strict types here.

If we intend to use strict types, we'll have a large amount of extra types to model each setting as there's not much overlap between settings.  These strict types also need to have some annotations for their representation as a string.  For example, `Family` would need to be something like
```
enum RequiredFamily {
    Ipv4,
    Ipv6,
    Both,
    Any
}
```
The macro would probably need to grow to understand annotations for the string representation of each of the enum variants so we would print `ipv4`, `both`, `any`, etc.

sources/api/netdog/networkd-derive/README.md
zmrow: I suppose that's fine.  I didn't account for any differences that may exist between the two, if any.

sources/api/netdog/src/networkd/config/network.rs
bcressey: > We validate input at the time we deserialize network config (net.toml), so at the point we are creating these files we can assume input is valid.

Strong typing at this layer is not about user input validation; rather, it's about making use of the type system so that the compiler has our back when we go to modify this code.

For example, `networkd` has its own rules around fields like `RequiredFamily` that don't map exactly to what's coming from input, so there's value in ensuring that the `required_family` field only gets appropriate values.

The current approach also doesn't seem aligned with how the `wicked` structs work, which do make use of stronger types.

> The macro would probably need to grow to understand annotations for the string representation of each of the enum variants so we would print ipv4, both, any, etc.

Can't the macro just assume that each field implements `Display` or some other trait, and call that impl?

> All that being said, if we intend to publish this macro/structs as a standalone crate somewhere I would absolutely agree about the validation of types.

I don't think it's a single standalone crate; rather, you have:
* macros to simplify serializing Rust structs as systemd unit files
* structs that correspond specifically to the data shapes that networkd expects

zmrow:
^ Addresses all the comments...  

(Ugh, I rebased onto develop and didn't push before I added my updates... :cry:   Sorry about that)

yeazelm (approved):

sources/api/netdog/systemd-derive/README.md
```diff
@@ -0,0 +1,129 @@
+# systemd-derive
```
arnaldo2792, line 1: I may benefit from this crate. Would it make sense to make it its own lib in `sources` out of `api`? We could name the crate `systemd-lib` and extend it as needed. I already know I want to have a function to receive the file descriptors passed by systemd, and I was planning to have this as a standalone lib that may be used by other “dogs”. We will use this both in EFS utils replacement, and when we decide to implement socket activation for API server.

sources/api/netdog/systemd-derive/README.md
```diff
@@ -0,0 +1,129 @@
...
+Under the hood, the macros implement `Display` for your structs.  This allows converting the
+structs to a string suitable for writing directly to file.
+
+The implementation is fairly rigid to the way we represent the "INI-like" structure of `systemd` unit
+files. `systemd` differs from standard INI format in that duplicate sections are allowed, along with
+duplicate keys within a section.  The macros expect that you will have a "top-level" struct that
+represents the unit file, with nested structs representing the sections of said file.  These nested
+structs representing sections have fields that are the configuration key/value pairs for that
+section. The expectation is that any validation of fields will happen before serialization; no
+validation happens within the macros.
```
bcressey: nit: avoid "we" / "you" language

The final sentence seems weird to me; I wouldn't expect the serializer to do any validation of fields unless there were values that were legal in Rust structs and illegal or unrepresentable in the unit file format.

sources/api/netdog/systemd-derive/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[lib]
+path = "src/lib.rs"
+proc-macro = true
+
+[dependencies]
+darling = {version = "0.14", default-features = false}
+proc-macro2 = "1"
+quote = "1"
+syn = {version = "1", default-features = false, features = ["derive"]}
```
bcressey: nit:
```suggestion
darling = { version = "0.14", default-features = false }
proc-macro2 = "1"
quote = "1"
syn = { version = "1", default-features = false, features = ["derive"] }
```

sources/api/netdog/systemd-derive/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+All struct fields must be either `Option`s or `Vec`s of `String`.  Fields that are `Vec`s will be
+iterated upon and each value of the `Vec` will be serialized.  For structs deriving `SystemdUnit`,
```
bcressey: This limitation is no longer present, right? The type just has to implement `Display` but doesn't need to be a `String`.

sources/api/netdog/systemd-derive/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+A macro to serialize structs to `systemd` unit file format
+
+### Description
+
+The `SystemdUnit` and `SystemdUnitSection` macros can be used to serialize structs representing
+`systemd` unit files.
+
+Under the hood, the macros implement `Display` for your structs.  This allows converting the
+structs to a string suitable for writing directly to file.
```
bcressey:
```suggestion
structs to a string suitable for writing directly to a file.
```

sources/api/netdog/systemd-derive/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+                }
+            }
+
+        });
+    }
+}
+
+impl ToTokens for SystemdUnitSectionField {
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
+        let struct_field_name = self.ident.as_ref().expect("Should always have a name");
```
bcressey, line 249: Is there an attribute that ensures this property?

bcressey (approved):
LGTM. Just a few documentation nits.

sources/api/netdog/systemd-derive/README.md
zmrow: Good catch - that's right!

sources/api/netdog/systemd-derive/src/lib.rs
zmrow: Yes, but indirectly :)

The attribute `supports(struct_named)` on the containing struct `SystemdUnitSection` ensures the macro only supports structs with named fields.  This means it does not support tuple or newtype structs.

`SystemdUnitSectionField` derives `FromField`, meaning it can only be "deserialized" from a struct field, and given the above `struct_named` attribute is guaranteed to have an ident/name.

sources/api/netdog/systemd-derive/README.md
zmrow: I don't mind if the macros move out of `netdog`, but I'd rather not conflate these macros as part of a larger systemd-related "library" crate.

zmrow:
Fixed up @bcressey 's requests

sources/api/netdog/systemd-derive/src/lib.rs
zmrow: Regardless - if the developer attempted to do "the wrong thing" here, it would be a compile-time issue printing out the `expect` message.

bcressey (approved):

PR #3137, opened 2023-05-23
pubsys: fix bug preventing multiple ssm promotions

### LINKED ISSUE #3103: pubsys: Using `promote-ssm` in a tiered promotion structure causes pubsys to panic
When calling `pubsys promote-ssm` using `--ssm-parameter-output` multiple times to perform a tiered rollout of SSM parameter deployments, then SSM parameters should be successfully promoted along those tiers.

**What actually happened:**
```
thread 'main' panicked at 'no entry found for key', pubsys/src/aws/promote_ssm/mod.rs:291:49
```

A few problems exist here:
* The `combine_parameters` file makes an assumption that all existing parameters in the file are being mapped to a target during the promotion, but in cases of tiered promotions, that isn't true.
* We use an Index operation on a HashMap which panics if the content is not accessible.

* Use `pubsys ssm --ssm-parameter-output $PATH` to create a set of SSM parameters
* Use `pubsys promote-ssm --ssm-parameter-output $PATH` to promote those to a new set of SSM parameters
* Use `pubsys promote-ssm --ssm-parameter-output $PATH` to once again promote those to a final set of parameters.

**Existing Workaround:**
You can work around this issue for now by refraining from using `--ssm-parameter-output` with pubsys. You may need to modify `Makefile.toml` if you are interfacing with pubsys via cargo-make.

cbgbt:
**Issue number:**

Closes #3103

**Description of changes:**
The goal of this feature is to keep a file of all published SSM parameters for the current Bottlerocket build.

During promotions, we parse the existing SSM file then add any newly created SSM parameter.

Prior to this change, the code which combines prior state with the newly-created parameters falsely assumed that all constructed promoted parameters would have a 1:1 mapping with a parameter that already exists in the file. This fails in some cases, e.g. when performing tiered promotions where the first tier is largely irrelevant when performing a promotion from the second to the third tier.

The change modifies the code to simply concatenate the existing data with any newly-created parameter, overwriting if the key exists.

**Testing done:**
* [x] unit tests pass
* [x] manual testing of a tiered promotion

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/promote_ssm/mod.rs
+++ b/tools/pubsys/src/aws/promote_ssm/mod.rs
@@ -211,8 +211,7 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
     // write the newly promoted parameters to `ssm_parameter_output` along with the original
     // parameters
     if let Some(ssm_parameter_output) = &promote_args.ssm_parameter_output {
-        append_rendered_parameters(ssm_parameter_output, &set_parameters, source_target_map)
-            .await?;
+        append_rendered_parameters(ssm_parameter_output, &set_parameters).await?;
     }
 
     // SSM set   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
@@ -236,7 +235,6 @@ pub(crate) async fn run(args: &Args, promote_args: &PromoteArgs) -> Result<()> {
 async fn append_rendered_parameters(
     ssm_parameters_output: &PathBuf,
     set_parameters: &HashMap<SsmKey, String>,
-    source_target_map: HashMap<&String, &String>,
 ) -> Result<()> {
     // If the file doesn't exist, assume that there are no existing parameters
     let parsed_parameters = parse_parameters(&ssm_parameters_output.to_owned())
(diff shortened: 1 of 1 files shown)


tools/pubsys/src/aws/promote_ssm/mod.rs
```diff
@@ -388,31 +382,30 @@ type Result<T> = std::result::Result<T, error::Error>;
...
-                )]),
+                SsmKey::new(Region::new("us-west-2"), "test2-parameter-name".to_string()),
+                "test2-parameter-value".to_string(),
+            ),
+            (
+                SsmKey::new(Region::new("us-east-1"), "test3-parameter-name".to_string()),
+                "test3-parameter-value".to_string(),
+            ),
+            (
+                SsmKey::new(
```
zmrow, line 404: This is just an additional key yeah?  I'm not missing more context there?

zmrow (approved):
Looks good; great simplification! :fireworks: 

Provided the manual testing goes well, I'm happy.

tools/pubsys/src/aws/promote_ssm/mod.rs
cbgbt: Correct! The idea was that this key isn't involved in the promotion.

This test would fail on the previous commit.

yeazelm (approved):

stmcginnis (approved):

stmcginnis (approved):
Just some lint issues to resolve.

cbgbt:
^ fixes clippy warnings.

etungsten (approved):

Issue #3059, opened 2023-04-27
Set net.unix.max_dgram_qlen for containers from the system-value...


diranged:
**What I'd like:**

Currently we can set `net.unix.max_dgram_qlen` in the Sysctl settings for the bottlerocket node on startup - however, that value is not propagated to the underlying containers when they start. They start up with `net.unix.max_dgram_qlen: 10` no matter what. I'd like to see an ability for us to tune `net.unix.max_dgram_qlen` for our containers.

One simple approach is to set the `net.unix.max_dgram_qlen` value in https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/containerd/containerd-cri-base-json based on whatever the user supplies to the `settings.kubernetes.sysctl` bottlerocket settings.

Another approach is to have a dedicated setting section for `settings.containerd.sysctl`, and allow us to set our settings there (at our own risk).

**Any alternatives you've considered:**

The only other alternative I've found is to try to use the CalicoCNI to modify settings on startup of containers via the `tuning` plugin .. however the Tigera Operator doesn't yet (https://github.com/projectcalico/calico/issues/5886#issuecomment-1098530412) support that.

bcressey:
It'd be nice to template this and expose it as `settings.oci-defaults.sysctl`, possibly as part of the work on #2862 since the same code paths are involved.

diranged:
@bcressey Is there any chance your team can share a target ETA for this feature? We are struggling internally with the default values for sending UDP/UDS packets.

vyaghras:
> Is there any chance your team can share a target ETA for this feature? We are struggling internally with the default values for sending UDP/UDS packets.

@diranged  We are planning to start on this soon.

bcressey:
@diranged tested an early version of this change and ran into trouble with [runc's validation logic](https://github.com/opencontainers/runc/blob/ba58ee9c3b9550c3e32b94802b0fb29761955290/libcontainer/configs/validate/validator.go#L150).

For this to be useful or safe, what we'd like is for runc to either always apply the sysctls, or skip them whenever they'd touch a host namespace while warning about non-namespaced sysctls, but that's not how runc does it:
* IPC related sysctls are only allowed when the container has a new IPC namespace
* network related sysctls are only allowed when the container has a new network namespace
* hostname / domainname sysctls are only allowed when the container has a new UTS namespace
* other sysctls are not allowed at all and cause an error

[...]

PR #3139, opened 2023-05-23
kernel: increase max_dgram_qlen sysctl value to 512

markusboehme:
**Issue number:** #3059 

**Description of changes:** Increase the kernel's default value for the net.unix.max_dgram_qlen sysctl to 512. This is a change to the kernel rather than a plain sysctl setting since systemd-sysctl only applies settings to the host, while the changed default value in the kernel also applies to every new network namespace.

I will propose this upstream as well. In case of rejection, Bottlerocket can continue to carry this downstream without any expected maintenance effort.

**Testing done:**

With a plain Fedora pod (`test`) on a aws-k8s-1.23 x86_64 node running

...with this change:

```
$ kubectl exec test -- cat /proc/sys/net/unix/max_dgram_qlen
512
```

...without this change:

```
$ kubectl exec test -- cat /proc/sys/net/unix/max_dgram_qlen
10
```

**Terms of contribution:**


yeazelm (approved):

bcressey (approved):

etungsten (approved):

diranged:
@etungsten amazing, thank you!

Issue #3143, opened 2023-05-24
EKS node do not get IPv6 address from wicked-dhcp6


sdomme:
**Image I'm using:** Bottlerocket OS 1.14.0 (aws-k8s-1.24)

**What I expected to happen:** 
In an EKS IPv6 Cluster I expect to get an IPv6 addr on the `eth0` interface (primary ENI) of the nodes, what also happen the most of the time. For example:
```
bash-5.1# wicked show all
lo              device-unconfigured
      link:     #1, state up
      type:     loopback
      addr:     ipv4 127.0.0.1/8
      addr:     ipv6 ::1/128

eth0            up
      link:     #2, state up, mtu 9001
      type:     ethernet, hwaddr 0a:ff:ec:87:a6:e8
      config:   wicked:xml:/etc/wicked/ifconfig/eth0.xml
      leases:   ipv4 dhcp granted
      leases:   ipv6 dhcp granted
      addr:     ipv4 10.33.254.4/24 [dhcp]
      addr:     ipv6 2a05:d014:dad:dc05::9cc/64 [dhcp]
      route:    ipv4 default via 10.33.254.1 [dhcp]
      route:    ipv6 default via fe80::850:32ff:fe76:84 metric 1024 proto ra
```

**What actually happened:**

It happen on multiple EKS Clusters (nodes are provisioned by Karpenter dynamically) that the nodes don't get an IPv6 addr on `eth0` even if the ENI has this IP addr configured (from the AWS console). Tools (in our case cilium) using IPv6 to talk to the kubernetes API run into timeout and break the whole network stack. 

What we could find so far:

IPv6 request is stuck. It says `granted` on a working node.
```
ash-5.1# wicked show eth0
eth0            up
      link:     #2, state up, mtu 9001
      type:     ethernet, hwaddr 02:ca:73:3a:e3:d0
      config:   wicked:xml:/etc/wicked/ifconfig/eth0.xml
      leases:   ipv4 dhcp granted
      leases:   ipv6 dhcp requesting
      addr:     ipv4 10.32.36.132/24 [dhcp]
      route:    ipv4 default via 10.32.36.1 [dhcp]
      route:    ipv6 default via fe80::b5:76ff:fec1:2b40 metric 1024 proto ra
```

Trying to get an IP manually leads to an timeout. 
Gives up after 10 seconds:
```
ash-5.1# wicked test dhcp6 eth0
[...]

Issue #2615, opened 2022-11-29
Network configuration: Provide an `accept-ra` option for IPv6 DHCP


zmrow:
**What I'd like:**
I'd like to be able to configure the ability for an interface to accept Router Advertisements.  This is an important option, especially for ENI's in IPv6 enabled VPC's on AWS.

Ideally this would be a boolean option inside an interface's `dhcp6` section in `net.toml`:
```
[eth0.dhcp6]
accept-ra = true 
```

**Any alternatives you've considered:**
* Setting sysctl's for an interface via settings: This doesn't work because the sysctl's aren't set until _after_ wicked brings up the interface
* Setting the `ipv6.conf.*.accept_ra` to `2` for all interfaces - we don't want to assume this is ok for all interfaces using IPv6 across all platforms

zmrow:
Related to: https://github.com/bottlerocket-os/bottlerocket/discussions/2561

PR #3159, opened 2023-06-01
netdog: set IPv6 accept-ra for primary interface via config rather than sysctl

zmrow:
**Issue number:**
Related to bottlerocket-os/bottlerocket#3143 

*A separate PR will be opened by @markusboehme for the related patches to `wicked`*

**Description of changes:**
```
    netdog: Add wicked IPv6 config structures to support RA
    
    This change adds the necessary structures to wicked in order to support
    setting the `accept_ra` IPv6 option via interface config.
```
```
    netdog: Set default accept-ra setting via config rather than sysctl
    
    Previously, we set the IPv6 accept-ra setting for the primary interface
    via `systemd-sysctl` during the wicked "install" helper.  Setting this
    sysctl via the helper can cause a race condition between the kernel and
    wicked.  The kernel sets a flag indicating a router advertisement has
    been received (`IF_RA_RCVD`), but only after it completes duplicate
    address detection and decides whether to send a router solicitation.  If
    the sysctl isn't set by the time duplicate address detection completes,
    the solicitation doesn't happen and the `IF_RA_RCVD` flag doesn't get
    set.  wicked uses this flag to decide whether or not to kick off the
    state machine that handles DHCP6.
    
    This change adds the accept-ra setting to the interface config if the
    primary interface is set up via kernel command line.  The primary
    interface is supplied via kernel command line for AWS and VMware
    variants.  This change also removes the accept-ra setting from the
    `systemd-sysctl` config.  Allowing wicked to manage this setting
    eliminates any chance of races between sysctl/kernel/wicked.
```
In the future we will support a network config setting that sets this in config for any interface specified.
**Testing done:**
* Launch 6,000 AWS instances (in batches of 2,000) including these fixes and the aforementioned pending `wicked` patches (thanks @markusboehme! ) into an IPv6 EKS cluster.  Observe all instances join the cluster, indicating they all received an IPv6 address successfully. 
* Also launched an `aws-k8s-1.24` instance into a "normal" IPv4 cluster successfully.
* Launch a `vmware-k8s-1.24` VM and observe it join my cluster successfully.
* Observe the proper config being set for `eth0`:
```
<interface><name>eth0</name>...<ipv6><accept-ra>router</accept-ra></ipv6></interface>
```
* Ovserve the proper sysctls being set via `systemd-sysctl` and from wicked
```

bash-5.1# sysctl net.ipv6.conf.eth0.accept_ra
```

**Terms of contribution:**


sources/api/netdog/test_data/wicked/eno1-ra.xml
```diff
@@ -0,0 +1 @@
+<interface><name>eno1</name><control><mode>boot</mode><link-detection><require-link></require-link></link-detection></control><ipv4:dhcp><enabled>true</enabled></ipv4:dhcp><ipv6><accept-ra>router</accept-ra></ipv6></interface>
```
yeazelm, line 1: Is it correct to do this if the interface is not configured for IPv6 otherwise?

sources/api/netdog/src/cli/generate_net_config.rs
```diff
@@ -48,8 +50,16 @@ pub(crate) fn run() -> Result<()> {
...
-    for interface in wicked_interfaces {
+    let mut wicked_interfaces = net_config.as_wicked_interfaces();
+    for interface in &mut wicked_interfaces {
+        // The kernel command line is too limited to fully specify an interface's configuration;
+        // fix some defaults to match legacy behavior.
+        // Note: we only allow 1 interface to be listed via kernel command line, so this will only
+        // be added to a single interface
+        if from_cmd_line {
+            interface.accept_ra();
+        }
```
markusboehme, line 61: I don't understand why this is conditional on the interface being configured via the kernel command line. Before this change, the interface's `accept_ra` sysctl would be set to `2` regardless. Should this instead be conditional on it being the primary interface? (With #2615 later it would also need to check for conflicting explicit configuration, but not yet.)

zmrow: Good question.  We want to make this fully configurable in #2615 .  

At the moment, metal deployments via EKS-A don't support IPv6, so limiting this change to variants that get their config via the command line (AWS/VMware) feels pretty safe from any regression.  AWS/VMware variants configure a single interface (the primary) via command line, so this config will only be applied to that primary interface.

markusboehme (approved):

sources/api/netdog/test_data/wicked/eno1-ra.xml
yeazelm: I think from further looking, this probably won't affect anything so I'm fine with it remaining even if it looks a little odd.

yeazelm (approved):

sources/api/netdog/test_data/wicked/eno1-ra.xml
zmrow: This feels pretty safe given that it only gets added if the interface is configured via kernel command line, which currently only happens in AWS/VMware variants where DHCP6 is optional by default.  At least currently, it would be a bit of a chore to specify an interface via kernel command line with other variants.

zmrow: Ideally, and I mentioned this in my comments with @markusboehme , we will make this more configurable in the future via net config.

stmcginnis (approved):
Changes look good, and coverage appears thorough enough.

sources/api/netdog/src/wicked/mod.rs
```diff
@@ -206,6 +206,12 @@ impl WickedInterface {
         }
     }
 
+    /// Add config to accept IPv6 router advertisements
+    // TODO: expose a network config option for this
+    pub(crate) fn accept_ra(&mut self) {
+        self.ipv6 = Some(WickedIpv6::default())
+    }
```
bcressey, line 213: I would've expected this change, and the changes to `write_config_file` to be in the previous commit.

bcressey (approved):
LGTM overall, but if you have a chance to clean up the commits so that all the necessary modifications to `wicked/mod.rs` happen together, that seems like a more logical sequence.

sources/api/netdog/src/wicked/mod.rs
zmrow: Good catch.... That was how I intended it to be; must have mixed it up somewhere along the way.  Will fix.

zmrow:
^ Fixed up the commits per @bcressey

bcressey (approved):

PR #3180, opened 2023-06-06
open-vm-tools: Modify shutdown for Bottlerocket

yeazelm:
**Description of changes:**

Bottlerocket doesn't have `/bin/sh` which makes the call `system()` fail with an unexpected error code. This results in the guest shutdown and reboot calls via vmtoolsd returning as if the request was successful but the call actually failed completely. This commit replaces the current call to `system()` with `g_spawn_command_line_sync()` which should be a drop in replacement besides removing the dependency on `/bin/sh`. For those reading the original code review, the testing all comes out the same and this code is functionally the same as the original logic.

**Testing done:**
Using the hard power setting works since power is pulled, but the `Shut Down Guest OS` or `Restart Guest OS` don't work without this PR. The same call for `Shut Down Guest OS` is managed with `govc vm.power -s=true $VMNAME`. Before:
```
$ govc vm.power -s=true yeazelm-debug
Shutdown guest VirtualMachine:vm-9463... OK

$ govc vm.info yeazelm-debug
Name:           yeazelm-debug
...
  Guest name:   Other 4.x or later Linux (64-bit)
  Memory:       8192MB
  CPU:          2 vCPU(s)
  Power state:  poweredOn
```
With the patch:
```
$ govc vm.power -s=true yeazelm-fork2
Shutdown guest VirtualMachine:vm-9466... OK

$ govc vm.info yeazelm-fork2
Name:           yeazelm-fork2
...
  Guest name:   Other 4.x or later Linux (64-bit)
  Memory:       8192MB
  CPU:          2 vCPU(s)
  Power state:  poweredOff
```
The Graphical UI buttons also work (including Restart) but is harder to show in text.

**Terms of contribution:**

### DIFF
--- a/packages/open-vm-tools/0003-Update-shutdown-code-to-work-for-Bottlerocket.patch
+++ b/packages/open-vm-tools/0003-Update-shutdown-code-to-work-for-Bottlerocket.patch
@@ -1,0 +1,76 @@
+From f3ae441a5ff9d27a3dd8223d55a7367ae519a0c8 Mon Sep 17 00:00:00 2001
+From: Matthew Yeazel <yeazelm@amazon.com>
+Date: Wed, 7 Jun 2023 13:34:49 -0700
+Subject: [PATCH] Update shutdown code to work for Bottlerocket
+
+Bottlerocket doesn't have /bin/sh which makes the call system() fail
+with an unexpected error code. This results in the guest shutdown and
+reboot calls via vmtoolsd returning as if the request was successful but
+the call actually failed completely. This commit replaces the current
+call to system() with g_spawn_sync() which should be a drop in
+replacement for system().
+
+Signed-off-by: Matthew Yeazel <yeazelm@amazon.com>
+---
+ lib/system/Makefile.am   |  3 +++
+ lib/system/systemLinux.c | 14 +++++++++++++-
+ 2 files changed, 16 insertions(+), 1 deletion(-)
+
+diff --git a/lib/system/Makefile.am b/lib/system/Makefile.am
+index 52b1a1dd..d2613c43 100644
+--- a/lib/system/Makefile.am
++++ b/lib/system/Makefile.am
+@@ -17,5 +17,8 @@
+ 
+ noinst_LTLIBRARIES = libSystem.la
+ 
++libSystem_la_CPPFLAGS =
++libSystem_la_CPPFLAGS += @GLIB2_CPPFLAGS@
++
+ libSystem_la_SOURCES =
+ libSystem_la_SOURCES += systemLinux.c
+diff --git a/lib/system/systemLinux.c b/lib/system/systemLinux.c
+index a688ab25..d24e2c4c 100644
+--- a/lib/system/systemLinux.c
++++ b/lib/system/systemLinux.c
+@@ -54,6 +54,7 @@
+ #include <sys/ioctl.h>
+ #include <net/if.h>
+ #include <sys/ioctl.h>
++#include <glib.h>
+ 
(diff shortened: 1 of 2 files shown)


packages/open-vm-tools/0003-Update-shutdown-code-to-work-for-Bottlerocket.patch
```diff
@@ -0,0 +1,106 @@
...
++  if (reboot) {
++      command = "reboot";
++  } else {
++    command = "poweroff";
++  }
++
++  pid = fork();
++  // pid == -1 means we couldn't fork
++  if (pid == -1) {
++    fprintf(stderr, "Unable to fork pid: %d\n", pid);
```
markusboehme: If this branch is entered, emitting `errno` or `sterror(errno)` would provide more information than `pid` (which we already know is -1).

markusboehme (approved):
The commit message has a typo: `s/getpid/waitpid/`. Looks good otherwise. :ship: 

For the more generic solution, perhaps this could continue using the `shutdown` compatibility shim provided by systemd. Then there would be no need either decide at runtime or carry Bottlerocket-specific code.

packages/open-vm-tools/0003-Update-shutdown-code-to-work-for-Bottlerocket.patch
yeazelm: True, this is a good catch. I'll update!

etungsten (approved):

packages/open-vm-tools/0003-Update-shutdown-code-to-work-for-Bottlerocket.patch
```diff
@@ -0,0 +1,106 @@
...
+
+diff --git a/lib/include/system.h b/lib/include/system.h
+index f1119cef..b82b2a26 100644
+--- a/lib/include/system.h
++++ b/lib/include/system.h
+@@ -37,6 +37,7 @@
+ uint64 System_GetTimeMonotonic(void);
+ uint64 System_Uptime(void);
+ void System_Shutdown(Bool reboot);
++void Bottlerocket_Shutdown(Bool reboot);
```
arnaldo2792: Would it make sense to avoid calling this function `Bottlerocket_Shutdown`, and rework the patch to check if `/bin/sh` exists, and if it does call `system`? With the hope that probably the VMware folks will take this patch upstream? :crossed_fingers:

arnaldo2792 (approved):

packages/open-vm-tools/0003-Update-shutdown-code-to-work-for-Bottlerocket.patch
yeazelm: @bcressey helped me find a better option which is `g_spawn_command_line_sync()` which should be a more upstream-able change.

yeazelm:
I refactored this PR to a different approach, resetting reviews since its fundamentally a different change.

etungsten (approved):

arnaldo2792 (approved):

bcressey (approved):
Nice!

packages/open-vm-tools/0003-Update-shutdown-code-to-work-for-Bottlerocket.patch
markusboehme: @arnaldo2792: If in doubt, you can also invoke `system(NULL)` to determine whether prerequisites are met.

TIL about `g_spawn_command_line_sync`. Nifty! Luckily there's a dependency on GLib already.

PR #3185, opened 2023-06-07
netdog: Add stronger typing to bond/vlan config

zmrow:
**Issue number:**
Related to #3134 

**Description of changes:**
This change adds stronger types to the bond/vlan configuration, and as a side effect simplifies the deserialization.

Previously, bond/vlan config implemented validation of the `kind` field via a custom `Deserialize` implementation, which complemented the derived `Deserialize` implementation via the serde annotation `serde(remote = "Self")`.  This annotation changes the generated deserialize function from trait implementations to inherent methods, which is how we could call into `::deserialze()`.  Adding an additional Deserialize implementation is unnecessary in this case as we can achieve the validation via stronger types.  We now validate the `kind` field using a single variant enum.  (Sidenote: the `kind` field is solely used for deserialization validation)

Previously we stored the VLAN ID as a u16 and did validation through the aforementioned `Deserialize` implementation.  Since VLAN ID will be used in the upcoming networkd-related changes, it made sense to make it a stronger type.  This new `VlanId` type implements `Deserialize` and does the same validation as before (must be between 0-4094).

**Testing done:**
* All unit tests continue to pass - which cover de/serialization both at the net config unit test level and e2e netconfig -> wicked config level.

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/vlan_id.rs
+++ b/sources/api/netdog/src/vlan_id.rs
@@ -1,0 +1,42 @@
+//! The vlan_id module contains the definition of a valid VLAN ID, and the code to support
+//! deserialization of the structure.  A valid VLAN ID must fall between the range of 0-4094.
+use serde::de::Error;
+use serde::{Deserialize, Deserializer};
+use std::fmt::Display;
+use std::ops::Deref;
+
+#[derive(Debug)]
+pub(crate) struct VlanId {
+    inner: u16,
+}
+
+impl<'de> Deserialize<'de> for VlanId {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let id: u16 = Deserialize::deserialize(deserializer)?;
+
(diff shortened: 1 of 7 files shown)


zmrow:
^ Fixed up the error message for VLAN ID to include the invalid ID (and make clippy happy :) )

webern (approved):

sources/api/netdog/src/vlan_id.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+impl<'de> Deserialize<'de> for VlanId {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let id: u16 = Deserialize::deserialize(deserializer)?;
+
+        if id > 4094 {
```
arnaldo2792, line 20: It will be nice to have a named constant for this limit

arnaldo2792 (approved):

yeazelm (approved):

PR #3188, opened 2023-06-08
docs: correct dm-verity text in security features

bcressey:
**Issue number:**
N/A

**Description of changes:**
Partial revert of b7c1ee45071202a87436203c8e44d4fa2eb94420.

"Fail closed" is correct in that the kernel will panic and reboot if corruption is detected.

"Fail close to when the block device is modified" is not correct; in the absence of memory pressure to evict previously read blocks from the cache, the detection could occur after much time has passed.

**Testing done:**
Viewed the rendered doc.

**Terms of contribution:**

### DIFF
--- a/SECURITY_FEATURES.md
+++ b/SECURITY_FEATURES.md
@@ -68,7 +68,8 @@ The root filesystem is marked as read-only and cannot be directly modified by us
 This protects against some container escape vulnerabilities such as [CVE-2019-5736](https://www.openwall.com/lists/oss-security/2019/02/11/2).
 
 The kernel is configured to restart if corruption is detected.
-That allows the system to fail close to when the underlying block device is unexpectedly modified.
+That allows the system to fail closed if the underlying block device is unexpectedly modified and the node is in an unknown state.
+The uncontrolled reboot will disrupt running containers, which can trigger alarms and prompt administrators to investigate.
 
 Although this provides a powerful layer of protection, it is **incomplete**.
 An attacker with full access to the block device could alter both the verity metadata and the contents of the root filesystem.


etungsten (approved):

SECURITY_FEATURES.md
```diff
@@ -68,7 +68,7 @@ The root filesystem is marked as read-only and cannot be directly modified by us
 This protects against some container escape vulnerabilities such as [CVE-2019-5736](https://www.openwall.com/lists/oss-security/2019/02/11/2).
 
 The kernel is configured to restart if corruption is detected.
-That allows the system to fail close to when the underlying block device is unexpectedly modified.
+That allows the system to fail closed if the underlying block device is unexpectedly modified.
```
stmcginnis: I think we need more descriptive text here explaining what this means. I've been peripherally around security for years, and this just read as a typo to me. So a few more words hinting at what this phrase means would be useful.

bcressey: This particular document uses other security-related terms of art - "memory corruption", "container escape", "persistent threat" - without defining them. "Fail closed" feels like a similarly useful term, and comes up in various places around Bottlerocket.

I've tried to elaborate a bit on the value of this design choice; please let me know if that helps to clarify.

webern (approved):

SECURITY_FEATURES.md
stmcginnis: I believe those other terms are more commonly understood, or at least a little easily understood contextually.

The expanded wording does help a bit, so I do think this is an overall improvement from the original text. Thanks!

stmcginnis (approved):

PR #3193, opened 2023-06-12
k8s: update eni-max-pods mapping for latest values

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This pulls in the pulls in instance types to the eni-max-pods mapping file. This is a periodic maintenance task we need to do each release to make sure we have the latest and greatest instance type information.

**Testing done:**

Launched instance and verified mapping values were used.

**Terms of contribution:**

### DIFF
--- a/packages/os/eni-max-pods
+++ b/packages/os/eni-max-pods
@@ -44,6 +44,7 @@ a1.large 29
 a1.medium 8
 a1.metal 234
 a1.xlarge 58
+bmn-sf1.metal 737
 c1.medium 12
 c1.xlarge 58
 c3.2xlarge 58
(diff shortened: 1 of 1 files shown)


etungsten (approved):

bcressey:
There's a slightly newer version here:
https://raw.githubusercontent.com/awslabs/amazon-eks-ami/master/files/eni-max-pods.txt

I've come to this PR a few times and gotten stalled when I attempted to validate the file's contents. I don't know if others have the same problem, but for me it's faster to download two files and compare them, vs. looking in `os.spec` and finding out I need to check out another project, rebuild a CLI, and run it in order to get the output.

bcressey (approved):

PR #3196, opened 2023-06-13
pubsys: Make AMI already registered message more visible

### LINKED ISSUE #2459: `cargo make ami` should scream at me if it didn't create an ami due to the name already existing
Edit: I changed the title to reflect the discussion that happened below. We don't EVER want to overwrite an AMI, but we want the tool to make it extremely apparent when it hasn't written an AMI. Currently the warning is too subtle.

While doing local development I need to test the changes I am making. That is made very easy now by running something like:

```
cargo make && cargo make ami && cargo make setup-test && cargo make test
```

Under normal situations, this compiles local changes, creates and registers an AMI, then runs testsys to get a cluster running using the local changes in the node image to validate.

**What actually happened:**

The problem is during a development cycle the git SHA-based identifier doesn't change. So if you are not paying close attention, the `cargo make ami` from above may print out the following:

```
[2022-09-26T22:20:39Z WARN  pubsys::aws::ami] Found 'bottlerocket-aws-k8s-1.22-x86_64-v1.10.0-8d731056-dirty' already registered in us-east-2: ami-06dd1bdaa7e4f351c
```

But then carry on. This can result in a cluster running without the recent changes, making it confusing why a change is not being reflected in the test deployment.

`cargo make ami` should probably overwrite any existing AMI, or error out. Alternatively, we could use some sort of timestamp based identifier rather than, or in addition to, the `-[sha]-dirty` identifier.

Make local changes, deploy, make more local changes, try to deploy again.

stmcginnis:
**Issue number:**

Closes #2459

**Description of changes:**

Pubsys emits a message when asked to publish an AMI that is already found. This is currently a single line message that blends in with the other output when doing `cargo make ami` and makes it very easy to miss.

Normally this is good, but when actively developing and making local code changes, this can make it more likely that the developer will not notice the message and spend time trying to understand why their code changes are not visible when running new instances using this AMI.

This changes the warning message to be a little more visible so it is much more noticeable when trying to publish an AMI that is already registered to try to avoid these cases.

**Testing done:**

Built and published an AMI. Attempted to publish it again and verified output:

```
16:49:41 [WARN] 
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Found 'bottlerocket-aws-k8s-1.26-x86_64-v1.15.0-f2eb176d-dirty' already registered in us-east-2: ami-0e8c69958b31082ba

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

**Terms of contribution:**

### DIFF
--- a/tools/pubsys/src/aws/ami/mod.rs
+++ b/tools/pubsys/src/aws/ami/mod.rs
@@ -33,6 +33,8 @@ use std::path::PathBuf;
 use structopt::{clap, StructOpt};
 use wait::wait_for_ami;
 
+const WARN_SEPARATOR: &str = "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!";
+
 /// Builds Bottlerocket AMIs using latest build artifacts
 #[derive(Debug, StructOpt)]
 #[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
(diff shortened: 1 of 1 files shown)


webern (approved):
+10000000

zmrow (approved):
:pizza:

PR #3197, opened 2023-06-13
Add dependabot config

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This adds a base dependabot config that does a few things:

1. Enables updates for GitHub Actions
2. Disables updates for most dependencies as we try to maintain those manually
3. Allows updates for dependencies under the `/tools` directory since our tooling doesn't need to be as strict
4. Still allows updates for any dependencies if a security issue is identified
5. Sets an appropriate label to match our label scheme instead of the default dependabot ones

**Testing done:**

Will see how dependabot behaves and tweak as necessary.

**Terms of contribution:**

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,45 @@
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "area/dependencies"
+
+  # We maintain updates for most dependencies. This disables updates other than
+  # security ones.
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "area/dependencies"
+    open-pull-requests-limit: 0
+
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "area/dependencies"
+    open-pull-requests-limit: 0
+
+  # Enable updates for the `tools` dependencies
+  - package-ecosystem: "cargo"
+    directory: "/tools"
+    ignore:
+      # For AWS SDK for Rust, we'll update when we bump tough/coldsnap
+      - dependency-name: "aws-config"
+      - dependency-name: "aws-endpoint"
+      - dependency-name: "aws-http"
+      - dependency-name: "aws-hyper"
+      - dependency-name: "aws-sig*"
+      - dependency-name: "aws-sdk*"
+      - dependency-name: "aws-smithy*"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "area/dependencies"


.github/dependabot.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - "area/dependencies"
+    open-pull-requests-limit: 0
+
+  - package-ecosystem: "gomod"
+    directory: "/tools"
+    schedule:
+      interval: "daily"
+    labels:
+      - "area/dependencies"
+    open-pull-requests-limit: 0
```
bcressey: We don't have any `Go` module dependencies in tools - why add these?

stmcginnis: I guess just as a safeguard in case we do add something, but it can be removed. I suppose chances are pretty small that we will add something that is not rust-based to our tooling code.

cbgbt (approved):

.github/dependabot.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "area/dependencies"
+
+  # We maintain updates for most dependencies. This disables updates other than
```
webern, line 12: How does this disable non-security updates. Seems like a weird API if the default for "no-config" includes non-security updates and the default(?) if there is a config is to include security configs only. I'm just curious.

webern (approved):

.github/dependabot.yaml
stmcginnis: The trick is in setting `open-pull-requests-limit` to 0. The default is a higher number, which allows updates for dependencies to be proposed in new PRs. Setting it to 0 effectively disallows any PRs to be created.

The exception to that limit is for security updates though. So if a security issue needs to be addressed, dependabot will just go ahead with a PR despite whatever the limit number has been set to.

.github/dependabot.yaml
```diff
@@ -0,0 +1,36 @@
...
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "area/dependencies"
+    open-pull-requests-limit: 0
+
+  # Enable updates for the `tools` dependencies
+  - package-ecosystem: "cargo"
+    directory: "/tools"
```
jpculp, line 32:
```suggestion
    directory: "/tools"
    ignore:
      # For AWS SDK for Rust, we'll update when we bump tough/coldsnap
      - dependency-name: "aws-endpoint"
      - dependency-name: "aws-http"
      - dependency-name: "aws-hyper"
      - dependency-name: "aws-sig*"
      - dependency-name: "aws-sdk*"
      - dependency-name: "aws-smithy*"
```

jpculp (changes requested):
Let's ignore bumping AWS SDK for Rust arbitrarily. They also all have to be bumped in the same PR when we do want to bump.

.github/dependabot.yaml
stmcginnis: Yeah, make sense.

jpculp (approved):

PR #3198, opened 2023-06-13
Add Xfs package to Bottlerocket

yeazelm:
**Issue number:** 3036

**Description of changes:**
This adds xfsprogs and it's dependencies to Bottlerocket. There is more work to properly make the variant boundary work but I have a rough set of changes to force the data partition into `xfs`. 

Note: We may hold back the aws-dev variant until we feel we are ready but for the time being, it simply makes the tools available, the data partition still remains `ext4` for the time being.

**Testing done:**
I build my own variant with xfs hard coded into the data partition. It boots, resizes, and you can see xfs data with xfs_* tools:
```
bash-5.1# xfs_info /local
meta-data=/dev/nvme1n1p1         isize=512    agcount=401, agsize=130816 blks
         =                       reflink=1    bigtime=1 inobtcount=1
data     =                       bsize=4096   blocks=52428288, imaxpct=25
         =                       sunit=128    swidth=128 blks
naming   =version 2              bsize=16384  ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=16384, version=2
```

**Terms of contribution:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,6 +69,7 @@ ARG REPO
 ARG GRUB_SET_PRIVATE_VAR
 ARG SYSTEMD_NETWORKD
 ARG UNIFIED_CGROUP_HIERARCHY
+ARG XFS_DATA_PARTITION
 ENV SYSTEMD_NETWORKD=${SYSTEMD_NETWORKD}
 ENV VARIANT=${VARIANT}
 WORKDIR /home/builder
@@ -102,6 +103,7 @@ RUN rpmdev-setuptree \
    && echo -e -n "${GRUB_SET_PRIVATE_VAR:+%bcond_without grub_set_private_var\n}" >> .bconds \
    && echo -e -n "${SYSTEMD_NETWORKD:+%bcond_without systemd_networkd\n}" >> .bconds \
    && echo -e -n "${UNIFIED_CGROUP_HIERARCHY:+%bcond_without unified_cgroup_hierarchy\n}" >> .bconds \
+   && echo -e -n "${XFS_DATA_PARTITION:+%bcond_without xfs_data_partition\n}" >> .bconds \
    && cat .bconds ${PACKAGE}.spec >> rpmbuild/SPECS/${PACKAGE}.spec \
    && find . -maxdepth 1 -not -path '*/\.*' -type f -exec mv {} rpmbuild/SOURCES/ \; \
    && echo ${NOCACHE}
@@ -197,6 +199,7 @@ ARG OS_IMAGE_PUBLISH_SIZE_GIB
 ARG DATA_IMAGE_PUBLISH_SIZE_GIB
 ARG KERNEL_PARAMETERS
 ARG GRUB_SET_PRIVATE_VAR
+ARG XFS_DATA_PARTITION
 ENV VARIANT=${VARIANT} VERSION_ID=${VERSION_ID} BUILD_ID=${BUILD_ID} \
     PRETTY_NAME=${PRETTY_NAME} IMAGE_NAME=${IMAGE_NAME} \
     KERNEL_PARAMETERS=${KERNEL_PARAMETERS}
@@ -214,6 +217,7 @@ RUN --mount=target=/host \
       --data-image-publish-size-gib="${DATA_IMAGE_PUBLISH_SIZE_GIB}" \
       --partition-plan="${PARTITION_PLAN}" \
       --ovf-template="/host/variants/${VARIANT}/template.ovf" \
+      ${XFS_DATA_PARTITION:+--xfs-data-partition=yes} \
       ${GRUB_SET_PRIVATE_VAR:+--with-grub-set-private-var=yes} \
     && echo ${NOCACHE}
 
--- a/packages/libinih/Cargo.toml
+++ b/packages/libinih/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "libinih"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "/dev/null"
+
+[package.metadata.build-package]
+releases-url = "https://github.com/benhoyt/inih/releases"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/benhoyt/inih/archive/refs/tags/r56.tar.gz"
+path = "inih-r56.tar.gz"
+sha512 = "ff3e0910990f73e5b21fddc84737ab346279f201c86c7ad864c6cad9de5bde57c3e0a433b9b8f3585b7d86feaae2ea074185f92891dcadc98c274c1c0745d2d2"
+
+[build-dependencies]
+glibc = { path = "../glibc" }
--- a/packages/libinih/libinih.spec
+++ b/packages/libinih/libinih.spec
@@ -1,0 +1,44 @@
+Name: %{_cross_os}libinih
+Version: 56
+Release: 1%{?dist}
+Summary: Simple INI file parser library
+License: BSD-3-Clause
+URL: https://github.com/benhoyt/inih
+Source0: https://github.com/benhoyt/inih/archive/refs/tags/r%{version}.tar.gz#/inih-r%{version}.tar.gz
+
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: meson
+
+%description
+%{summary}.
+
+%package devel
+Summary: Files for development using the simple INI file parser library
+Requires: %{name}
+
+%description devel
+%{summary}.
+
+%prep
+%autosetup -n inih-r%{version}
+
+%build
+CONFIGURE_OPTS=(
+  -Dwith_INIReader=false
+)
+
+%cross_meson "${CONFIGURE_OPTS[@]}"
+%cross_meson_build
+
+%install
+%cross_meson_install
+
+%files
+%license LICENSE.txt
+%{_cross_attribution_file}
+%{_cross_libdir}/libinih.so.0
+
+%files devel
+%{_cross_pkgconfigdir}/inih.pc
+%{_cross_libdir}/libinih.so
+%{_cross_includedir}/ini.h
--- a/packages/liburcu/liburcu.spec
+++ b/packages/liburcu/liburcu.spec
@@ -1,0 +1,69 @@
+Name: %{_cross_os}liburcu
+Version: 0.14.0
+Release: 1%{?dist}
+Summary: Library for userspace RCU
+License: LGPL-2.1-only AND GPL-2.0-or-later AND MIT
+URL: http://liburcu.org
+Source0: http://lttng.org/files/urcu/userspace-rcu-latest-0.14.tar.bz2
+
+BuildRequires: %{_cross_os}glibc-devel
+
+%description
+%{summary}.
+
+%package devel
+Summary: Files for development using the library for userspace RCU
+Requires: %{name}
+
+%description devel
+%{summary}.
+
+%prep
+%autosetup -n userspace-rcu-%{version}
+
+%build
+%cross_configure --disable-static
+
+# "fix" rpath
+sed -i 's|^hardcode_libdir_flag_spec=.*|hardcode_libdir_flag_spec=""|g' libtool
+sed -i 's|^runpath_var=LD_RUN_PATH|runpath_var=DIE_RPATH_DIE|g' libtool
+
+%make_build
+
+%install
+%make_install
+
+%files
+%license LICENSE gpl-2.0.txt lgpl-relicensing.txt lgpl-2.1.txt
+%{_cross_attribution_file}
+
+%{_cross_libdir}/liburcu.so.8*
+%{_cross_libdir}/liburcu-common.so.8*
+
+%exclude %{_cross_libdir}/liburcu-bp.so.8*
+%exclude %{_cross_libdir}/liburcu-cds.so.8*
+%exclude %{_cross_libdir}/liburcu-mb.so.8*
+%exclude %{_cross_libdir}/liburcu-memb.so.8*
+%exclude %{_cross_libdir}/liburcu-qsbr.so.8*
+%exclude %{_cross_libdir}/liburcu-signal.so.8*
+%exclude %{_cross_docdir}
+
+%files devel
+%{_cross_includedir}/*
+%{_cross_libdir}/liburcu-common.so
+%{_cross_libdir}/liburcu.so
+%{_cross_libdir}/pkgconfig/liburcu.pc
+
+%exclude %{_cross_libdir}/pkgconfig/liburcu-bp.pc
+%exclude %{_cross_libdir}/pkgconfig/liburcu-cds.pc
+%exclude %{_cross_libdir}/pkgconfig/liburcu-mb.pc
+%exclude %{_cross_libdir}/pkgconfig/liburcu-memb.pc
+%exclude %{_cross_libdir}/pkgconfig/liburcu-qsbr.pc
+%exclude %{_cross_libdir}/pkgconfig/liburcu-signal.pc
+%exclude %{_cross_libdir}/liburcu-bp.so
+%exclude %{_cross_libdir}/liburcu-cds.so
+%exclude %{_cross_libdir}/liburcu-common.so
+%exclude %{_cross_libdir}/liburcu-mb.so
+%exclude %{_cross_libdir}/liburcu-memb.so
+%exclude %{_cross_libdir}/liburcu-qsbr.so
+%exclude %{_cross_libdir}/liburcu-signal.so
--- a/packages/release/prepare-local-fs.service
+++ b/packages/release/prepare-local-fs.service
@@ -10,8 +10,9 @@ RefuseManualStop=true
 [Service]
 Type=oneshot
 
+EnvironmentFile=/usr/share/bottlerocket/image-features.env
 # Create the filesystem on the partition, if it doesn't exist.
-ExecStart=/usr/lib/systemd/systemd-makefs ext4 /dev/disk/by-partlabel/BOTTLEROCKET-DATA
+ExecStart=/usr/lib/systemd/systemd-makefs ${DATA_PARTITION_FILESYSTEM} /dev/disk/by-partlabel/BOTTLEROCKET-DATA
 
 # Stop and mask the repart-data-* oneshots in case they're waiting on non-existent data partitions.
 # 'BOTTLEROCKET-DATA' already exists so we can move on.
--- a/packages/xfsprogs/0001-libxfs-do-not-try-to-run-the-crc32selftest.patch
+++ b/packages/xfsprogs/0001-libxfs-do-not-try-to-run-the-crc32selftest.patch
@@ -1,0 +1,44 @@
+From 3a77dfc54271059dcac305384bf6ace34fe1f3d3 Mon Sep 17 00:00:00 2001
+From: "Yann E. MORIN" <yann.morin.1998@free.fr>
+Date: Sun, 18 Dec 2016 15:37:27 +0100
+Subject: [PATCH] libxfs: do not try to run the crc32selftest
+
+Even though the crc32selftest is natively compiled (because it is to be
+executed), it fails in cross-compilation as the host may lack the
+required headers, like uuid/uuid.h (e.g. in a minimal environment).
+
+Moreover, running the crc32selftest natively is completely wrong,
+because it passing on the host does not mean it would still pass n the
+target (because endianness or bitness or alignment differences).
+
+So, just disable running the crc32selftest altogether.
+
+Note that there's a remaining bug-in-hiding, because the crc32 table
+generator is natively built, but with the target CFLAGS.
+
+Signed-off-by: "Yann E. MORIN" <yann.morin.1998@free.fr>
+Signed-off-by: "Fabrice Fontaine" <fontaine.fabrice@gmail.com>
+[Update for 4.18.0: crc32 has been moved from libxfs to libfrog]
+---
+ libfrog/Makefile | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/libfrog/Makefile b/libfrog/Makefile
+index 0110708..675c1cb 100644
+--- a/libfrog/Makefile
++++ b/libfrog/Makefile
+@@ -54,9 +54,9 @@ ifeq ($(HAVE_GETMNTENT),yes)
+ LCFLAGS += -DHAVE_GETMNTENT
+ endif
+ 
+-LDIRT = gen_crc32table crc32table.h crc32selftest
++LDIRT = gen_crc32table crc32table.h
+ 
+-default: crc32selftest ltdepend $(LTLIBRARY)
++default: ltdepend $(LTLIBRARY)
+ 
+ crc32table.h: gen_crc32table.c crc32defs.h
+ 	@echo "    [CC]     gen_crc32table"
+-- 
+2.40.1
+
--- a/packages/xfsprogs/Cargo.toml
+++ b/packages/xfsprogs/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "xfsprogs"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "/dev/null"
+
+[package.metadata.build-package]
+releases-url = "https://mirrors.edge.kernel.org/pub/linux/utils/fs/xfs/xfsprogs/"
+
+[[package.metadata.build-package.external-files]]
+url = "http://kernel.org/pub/linux/utils/fs/xfs/xfsprogs/xfsprogs-6.3.0.tar.xz"
+sha512 = "dbb3e77d0d9cf184a0e647b8231350401a7549a23a0bfd9121cf2a1b48e85f71d98329dff440fc6e984bcecfdcc2a72f0f27c4989560f3c55359f21f3fb434bb"
+
+[build-dependencies]
+glibc = { path = "../glibc" }
+libinih = { path = "../libinih" }
+liburcu = { path = "../liburcu" }
+util-linux = { path = "../util-linux" }
+
+# RPM Requires
+[dependencies]
+# none - will likely change
--- a/packages/xfsprogs/Cargo.toml
+++ b/packages/xfsprogs/Cargo.toml
@@ -23,4 +23,4 @@ util-linux = { path = "../util-linux" }
 
 # RPM Requires
 [dependencies]
-# none - will likely change
+# none
--- a/packages/xfsprogs/xfsprogs.spec
+++ b/packages/xfsprogs/xfsprogs.spec
@@ -1,0 +1,84 @@
+Name: %{_cross_os}xfsprogs
+Version: 6.3.0
+Release: 1%{?dist}
+Summary: Utilities for managing the XFS filesystem
+License: GPL-2.0-only and LGPL-2.1-only
+URL: https://xfs.wiki.kernel.org
+Source0: http://kernel.org/pub/linux/utils/fs/xfs/xfsprogs/xfsprogs-%{version}.tar.xz
+Patch1: 0001-libxfs-do-not-try-to-run-the-crc32selftest.patch
+
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: %{_cross_os}libuuid-devel
+BuildRequires: %{_cross_os}libinih-devel
+BuildRequires: %{_cross_os}liburcu-devel
+BuildRequires: %{_cross_os}libblkid-devel
+
+Requires: %{_cross_os}liburcu
+Requires: %{_cross_os}libinih
+
+%description
+%{summary}.
+
+%package devel
+Summary: XFS filesystem-specific headers
+Requires: %{name}
+
+%description devel
+%{summary}.
+
+%prep
+%autosetup -n xfsprogs-%{version} -p1
+
+%build
+%cross_configure \
+  --enable-blkid=yes \
+  --enable-lto=no \
+  --enable-editline=no \
+  --enable-scrub=no
+
+%make_build
+
+%install
+make DIST_ROOT=%{buildroot} install install-dev \
+  PKG_ROOT_SBIN_DIR=%{_cross_sbindir} PKG_ROOT_LIB_DIR=%{_cross_libdir}
+
+rm -f %{buildroot}/%{_cross_libdir}/*.{la,a}
+
+%files
+%license LICENSES/GPL-2.0 LICENSES/LGPL-2.1
+%{_cross_attribution_file}
+%{_cross_libdir}/*.so.*
+%{_cross_sbindir}/mkfs.xfs
+%{_cross_sbindir}/xfs_copy
+%{_cross_sbindir}/xfs_db
+%{_cross_sbindir}/xfs_estimate
+%{_cross_sbindir}/xfs_fsr
+%{_cross_sbindir}/xfs_growfs
+%{_cross_sbindir}/xfs_io
+%{_cross_sbindir}/xfs_logprint
+%{_cross_sbindir}/xfs_mdrestore
+%{_cross_sbindir}/xfs_quota
+%{_cross_sbindir}/xfs_repair
+%{_cross_sbindir}/xfs_rtcp
+%{_cross_sbindir}/xfs_spaceman
+%{_cross_datadir}/xfsprogs/mkfs/*.conf
+
+# Exclude shell scripts
+%exclude %{_cross_sbindir}/fsck.xfs
+%exclude %{_cross_sbindir}/xfs_admin
+%exclude %{_cross_sbindir}/xfs_bmap
+%exclude %{_cross_sbindir}/xfs_freeze
+%exclude %{_cross_sbindir}/xfs_info
+%exclude %{_cross_sbindir}/xfs_metadump
+%exclude %{_cross_sbindir}/xfs_mkfile
+%exclude %{_cross_sbindir}/xfs_ncheck
+
+%exclude %{_cross_mandir}
+%exclude %{_cross_localedir}
+%exclude %{_cross_docdir}
+
+
+%files devel
+%dir %{_cross_includedir}/xfs
+%{_cross_includedir}/xfs/*.h
+%{_cross_libdir}/*.so
--- a/tools/buildsys/src/manifest.rs
+++ b/tools/buildsys/src/manifest.rs
@@ -214,6 +214,13 @@ line arguments set in the boot configuration.
 [package.metadata.build-variant.image-features]
 unified-cgroup-hierarchy = true
 ```
+
+`xfs-data-partition` changes the filesystem for the data partition from ext4 to xfs. The
+default will remain ext4 and xfs is opt-in.
+
+```ignore
+[package.metadata.build-variant.image-features]
+xfs-data-partition = true
 */
 
 mod error;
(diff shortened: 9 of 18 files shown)


packages/libinih/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+build = "../build.rs"
+
+[lib]
+path = "/dev/null"
+
+[package.metadata.build-package]
+releases-url = "https://github.com/benhoyt/inih/archive/refs/tags/"
+
+[[package.metadata.build-package.external-files]]
+url = "https://github.com/benhoyt/inih/archive/refs/tags/r56.tar.gz"
```
bcressey, line 15:
```suggestion
url = "https://github.com/benhoyt/inih/archive/refs/tags/r56.tar.gz"
path = "inih-r56.tar.gz"
```

packages/libinih/libinih.spec
```diff
@@ -0,0 +1,48 @@
+Name: %{_cross_os}libinih
+Version: 56
+Release: 1%{?dist}
+Summary: Simple INI file parser library
+License: BSD
+URL: https://github.com/benhoyt/inih
+Source0: https://github.com/benhoyt/inih/archive/refs/tags/r%{version}.tar.gz
```
bcressey:
```suggestion
Source0: https://github.com/benhoyt/inih/archive/refs/tags/r%{version}.tar.gz#/inih-%{version}.tar.gz
```

packages/libinih/libinih.spec
```diff
@@ -0,0 +1,48 @@
...
+%{_cross_libdir}/libinih.so.0
+
+%files devel
+%{_cross_pkgconfigdir}/inih.pc
+%{_cross_libdir}/libinih.so
+
+%{_cross_includedir}/ini.h
+
+
+%changelog
```
bcressey: nit:
```suggestion
%files devel
%{_cross_pkgconfigdir}/inih.pc
%{_cross_libdir}/libinih.so
%{_cross_includedir}/ini.h

%changelog
```

packages/libuserspace_rcu/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "libuserspace_rcu"
```
bcressey: I would prefer to call this `liburcu` here and in the spec, since that matches the library name.

packages/libuserspace_rcu/libuserspace_rcu.spec
```diff
@@ -0,0 +1,67 @@
...
+Source0: http://lttng.org/files/urcu/userspace-rcu-latest-0.14.tar.bz2
+
+BuildRequires: %{_cross_os}glibc-devel
+
+%description
+%{summary}.
+
+%package devel
+Summary: Development package for %{name}
+Requires: %{name}
```
bcressey: nit: the other packages follow these conventions for summaries -
```
main -
Summary: Library for <X>

devel -
Summary: Files for development using the library for <X>
```

So I would do this as:
```
Summary: Library for userspace RCU
Summary: Files for development using the library for userspace RCU
```

This is pedantic but is meant to reduce any effort spent on creative writing for these fields. Same rationale applies for `%description`.

packages/libuserspace_rcu/libuserspace_rcu.spec
```diff
@@ -0,0 +1,67 @@
...
+%{_cross_libdir}/liburcu-mb.so.8*
+%{_cross_libdir}/liburcu-memb.so.8*
+%{_cross_libdir}/liburcu.so.8*
+%{_cross_libdir}/liburcu-qsbr.so.8*
+%{_cross_libdir}/liburcu-signal.so.8*
+
+%exclude %{_cross_datadir}/doc/*
+
+%files devel
+%{_cross_includedir}/*
```
bcressey: `%{_cross_includedir}/*` and `%{_cross_includedir}/*.h` below seem redundant.

packages/libuserspace_rcu/libuserspace_rcu.spec
```diff
@@ -0,0 +1,67 @@
...
+%files
+%license LICENSE
+%{_cross_attribution_file}
+	
+%{_cross_libdir}/liburcu-bp.so.8*
+%{_cross_libdir}/liburcu-cds.so.8*
+%{_cross_libdir}/liburcu-common.so.8*
+%{_cross_libdir}/liburcu-mb.so.8*
+%{_cross_libdir}/liburcu-memb.so.8*
+%{_cross_libdir}/liburcu.so.8*
```
bcressey: Can we just include this version of the library, or whichever one xfsprogs uses? I don't want to ship a bunch of different implementations when we only use one.

packages/xfsprogs/0001-libxfs-do-not-try-to-run-the-crc32selftest.patch
```diff
@@ -0,0 +1,44 @@
+From 353e3b5a88849efced5b9faedbf8b64c6a183933 Mon Sep 17 00:00:00 2001
+From: Matthew Yeazel <yeazelm@amazon.com>
+Date: Tue, 13 Jun 2023 20:06:06 +0000
+Subject: [PATCH] [PATCH] libxfs: do not try to run the crc32selftest
```
bcressey: Let's keep the original buildroot patch header for attribution purposes.

packages/xfsprogs/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+build = "../build.rs"
+
+[lib]
+path = "/dev/null"
+
+[package.metadata.build-package]
+releases-url = "https://xfs.wiki.kernel.org"
+
+[[package.metadata.build-package.external-files]]
+url = "http://kernel.org/pub/linux/utils/fs/xfs/xfsprogs/xfsprogs-5.15.0.tar.xz"
```
bcressey: Does the latest version not work? Typically these packages are moved to the latest version as part of the update cycle, so if it doesn't work now then it'll just break for the next maintainer that comes along.

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,86 @@
...
+%prep
+%autosetup -n xfsprogs-%{version} -p1
+
+%build
+%cross_configure \
+  --enable-editline=yes \
+  --enable-blkid=yes \
+  --enable-lto=no \
+  --enable-editline=no \
+  --enable-scrub=no
```
bcressey:
```suggestion
  --enable-blkid=yes \
  --enable-lto=no \
  --enable-editline=no \
  --enable-scrub=no
```

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,86 @@
...
+
+
+%make_build \
+  INCLUDEDIR=%{_cross_includedir} \
+  LIBDIR=%{_cross_libdir}
```
bcressey: Are `INCLUDEDIR` and `LIBDIR` not set to the correct default values?

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,86 @@
...
+
+
+%install
+make DIST_ROOT=%{buildroot} install install-dev \
+	PKG_ROOT_SBIN_DIR=%{_cross_sbindir} PKG_ROOT_LIB_DIR=%{_cross_libdir}
```
bcressey: No hard tabs!

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,86 @@
...
+%files
+%license LICENSES/GPL-2.0 LICENSES/LGPL-2.1
+%{_cross_attribution_file}
+%{_cross_libdir}/*.so.*
+#%dir %{_cross_libdir}/xfsprogs
+#%{_cross_libdir}/xfsprogs/*
+%exclude %{_cross_mandir}
+%exclude %{_cross_datadir}/doc/*
+%exclude %{_cross_datadir}/locale/*
+%{_cross_sbindir}/*
```
bcressey: Can we list out these programs?

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,86 @@
...
+%make_build \
+  INCLUDEDIR=%{_cross_includedir} \
+  LIBDIR=%{_cross_libdir}
+
+
+%install
+make DIST_ROOT=%{buildroot} install install-dev \
+	PKG_ROOT_SBIN_DIR=%{_cross_sbindir} PKG_ROOT_LIB_DIR=%{_cross_libdir}
+
+
```
bcressey: nit: I would clean up some of this double-spacing

packages/xfsprogs/Cargo.toml
yeazelm: I was having difficulties compiling with some strange errors. I've lost them since but will try moving to 6.3.0 and see if they reappear.

packages/libuserspace_rcu/libuserspace_rcu.spec
yeazelm: Makes sense to me! Corrected!

yeazelm: I missed this update for libinih and will update shortly.

packages/libuserspace_rcu/libuserspace_rcu.spec
yeazelm: Fixed

packages/libuserspace_rcu/libuserspace_rcu.spec
yeazelm: I was able to determine xfsprogs only links against `liburcu.so.8` and `liburcu-common.so.8`

packages/xfsprogs/0001-libxfs-do-not-try-to-run-the-crc32selftest.patch
yeazelm: Whoops, I meant to do this and spaced it, thanks for the catch!

packages/xfsprogs/xfsprogs.spec
yeazelm: These were lingering from other failures, they do not add anything.

packages/xfsprogs/xfsprogs.spec
yeazelm: Listed them out instead.

packages/libuserspace_rcu/Cargo.toml
yeazelm: Updated the naming.

packages/xfsprogs/Cargo.toml
yeazelm: I moved back to 6.3.0 and it builds and worked perfectly. I think it might have been other things stabilizing with liburcu so it seems fine now.

packages/xfsprogs/xfsprogs.spec
yeazelm: Not sure how it came in, sorry!

packages/libinih/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "/dev/null"
+
+[package.metadata.build-package]
+releases-url = "https://github.com/benhoyt/inih/archive/refs/tags/"
```
etungsten: This link 404s for me.
```suggestion
releases-url = "https://github.com/benhoyt/inih/releases"
```

packages/liburcu/liburcu.spec
```diff
@@ -0,0 +1,67 @@
+Name: %{_cross_os}liburcu
+Version: 0.14.0
+Release: 1%{?dist}
+Summary: Library for userspace RCU
+License: LGPL-2.1-or-later
+URL: http://liburcu.org
+Source0: http://lttng.org/files/urcu/userspace-rcu-latest-0.14.tar.bz2
```
etungsten: nit:
```suggestion
Version: 0.14
Release: 1%{?dist}
Summary: Library for userspace RCU
License: LGPL-2.1-or-later
URL: http://liburcu.org
Source0: http://lttng.org/files/urcu/userspace-rcu-latest-%{version}.tar.bz2
```

packages/xfsprogs/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "/dev/null"
+
+[package.metadata.build-package]
+releases-url = "https://xfs.wiki.kernel.org"
```
etungsten: Would this be more accurate?
```suggestion
releases-url = "https://mirrors.edge.kernel.org/pub/linux/utils/fs/xfs/xfsprogs/"
```

etungsten (approved):

packages/liburcu/liburcu.spec
yeazelm: The version is nonstandard, the version *is* 0.14.0 but the tarball that is downloaded is named 0.14. So to use `%{version}` during the the `autosetup` step, it needs to have the .0 in it. So either we have to do conversion in the `Source0` or the `autosetup` and I felt like it is "more correct" to change the name from the Source than in the `autosetup`. I can switch it around though if you feel its better this way.

packages/libinih/Cargo.toml
yeazelm: Good call out. I can fix this.

packages/xfsprogs/Cargo.toml
yeazelm: Yeah, that is probably more accurate. I was thinking of this more as the `URL` in the specfile but strictly speaking the releases are found at that URL.

packages/liburcu/liburcu.spec
etungsten: The way you currently have it is fine. I wasn't aware of how the version is used elsewhere.

variants/aws-dev/Cargo.toml
```diff
@@ -36,6 +36,7 @@ included-packages = [
     "strace",
     "tcpdump",
     "chrony-tools",
+    "xfsprogs",
```
stmcginnis: Looks good for now while working through next steps. Just to note, would be good to get this moved to be part of the `os` package once we are ready.

stmcginnis (approved):
lgtm

variants/aws-dev/Cargo.toml
yeazelm: Correct, for the moment, this is just to ensure it gets built at all. The next PR will be adding in the proper support for XFS in variants which will move this to the right place.

packages/liburcu/liburcu.spec
```diff
@@ -0,0 +1,67 @@
...
+
+%exclude %{_cross_libdir}/liburcu-bp.so
+%exclude %{_cross_libdir}/liburcu-cds.so
+%exclude %{_cross_libdir}/liburcu-common.so
+%exclude %{_cross_libdir}/liburcu-mb.so
+%exclude %{_cross_libdir}/liburcu-memb.so
+%exclude %{_cross_libdir}/liburcu-qsbr.so
+%exclude %{_cross_libdir}/liburcu-signal.so
+
+%changelog
```
arnaldo2792: nit: you don't need this line

variants/aws-dev/Cargo.toml
arnaldo2792: I agree we should only include it in the variants that will be released with XFS. Existing variants should remain in ext4.

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,93 @@
+Name: %{_cross_os}xfsprogs
+Version: 6.3.0
+Release: 1%{?dist}
+Summary: Utilities for managing the XFS filesystem
+License: GPL+ and LGPLv2+
```
arnaldo2792: This package includes copies of the GPL-2.0-only and LGPL-2.1-only licenses, this should match those licenses.

packages/liburcu/liburcu.spec
```diff
@@ -0,0 +1,67 @@
+Name: %{_cross_os}liburcu
+Version: 0.14.0
+Release: 1%{?dist}
+Summary: Library for userspace RCU
+License: LGPL-2.1-or-later
```
arnaldo2792: Different items in this package are distributed under different licenses, according to the LICENSE file in the sources:

```
The library part is distributed under LGPLv2.1 or later.
.....
.....
xchg() primitive has been rewritten from scratch starting from atomic_ops 1.2
which has a MIT-style license that is intended to allow use in both free and
proprietary software:
....
....
Library test code is distributed under the GPLv2 license. See gpl-2.0.txt for
...
...
The following build-related macro is under GPLv3 (or later):
```

packages/liburcu/liburcu.spec
```diff
@@ -0,0 +1,67 @@
...
+
+%make_build
+
+%install
+%make_install
+
+find %{buildroot} -type f -name "*.a" -delete
+
+%files
+%license LICENSE
```
arnaldo2792: The sources distribute copies of the licenses that apply to the package, should you use those instead of the `LICENSE`file?

packages/libinih/libinih.spec
```diff
@@ -0,0 +1,46 @@
+Name: %{_cross_os}libinih
+Version: 56
+Release: 1%{?dist}
+Summary: Simple INI file parser library
+License: BSD
```
arnaldo2792: This package is released under the new BSD-3 clause:

```bash
inih-r56 is 📦 v56 ❯ askalono crawl .
./LICENSE.txt
License: BSD-3-Clause (original text)
```

packages/libinih/libinih.spec
```diff
@@ -0,0 +1,46 @@
...
+%license LICENSE.txt
+%{_cross_attribution_file}
+%{_cross_libdir}/libinih.so.0
+
+%files devel
+%{_cross_pkgconfigdir}/inih.pc
+%{_cross_libdir}/libinih.so
+%{_cross_includedir}/ini.h
+
+%changelog
```
arnaldo2792: This line isn't needed

variants/Cargo.lock
```diff
@@ -630,13 +630,6 @@ dependencies = [
  "util-linux",
 ]
 
-[[package]]
-name = "libinih"
```
arnaldo2792: nit: some cleanup in this commit.

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,93 @@
...
+%{_cross_includedir}/xfs/xfs_fs.h
+%{_cross_includedir}/xfs/xfs_fs_compat.h
+%{_cross_includedir}/xfs/xfs_types.h
+%{_cross_includedir}/xfs/xfs_format.h
+%{_cross_includedir}/xfs/xfs_da_format.h
+%{_cross_includedir}/xfs/xfs_log_format.h
+%{_cross_includedir}/xfs/xqm.h
+%{_cross_libdir}/*.so
+
+%changelog
```
arnaldo2792: This line isn't needed.

arnaldo2792 (changes requested):

packages/xfsprogs/xfsprogs.spec
yeazelm: I'll take a look and make sure this is right.

packages/liburcu/liburcu.spec
yeazelm: Same with this, I will take a look and make sure its correct.

variants/Cargo.lock
yeazelm: Yeah, I'll do this cleanup if I can, I've tried twice in rebases to fix it but I'll take another shot at it.

Dockerfile
```diff
@@ -102,6 +103,7 @@ RUN rpmdev-setuptree \
    && echo -e -n "${GRUB_SET_PRIVATE_VAR:+%bcond_without grub_set_private_var\n}" >> .bconds \
    && echo -e -n "${SYSTEMD_NETWORKD:+%bcond_without systemd_networkd\n}" >> .bconds \
    && echo -e -n "${UNIFIED_CGROUP_HIERARCHY:+%bcond_without unified_cgroup_hierarchy\n}" >> .bconds \
+   && echo -e -n "${XFS_DATA_PARTITION:+%bcond_without xfs_data_partition\n}" >> .bconds \
```
yeazelm, line 106: I added this for if we need to do conditional specfile things. I didn't end up adding `xfsprogs` conditionally to the image based upon this but we could if we want. If we don't see the need for this I can remove it as well.

tools/rpm2img
```diff
@@ -408,8 +417,28 @@ mkfs_data() {
...
   offset="${3:?}"
-  mkfs.ext4 -m 0 -d "${DATA_MOUNT}" "${BOTTLEROCKET_DATA}" "${size}"
-  echo "${UNLABELED}" | debugfs -w -f - "${BOTTLEROCKET_DATA}"
+  # Create an XFS filesystem if requested
+  if [ "${XFS_DATA_PARTITION}" == "yes" ] ; then
+    echo "writing XFS filesystem for DATA"
+    # Create a file to write the filesystem to first
+    dd if=/dev/zero of="${BOTTLEROCKET_DATA}" bs=1M count=${size%?}
+    # block size of 4096, directory block size of 16384
+    # enable inotbtcount and bigtime
```
yeazelm: We could also enable reflink if we want. I haven't done any testing to validate if it changes performance at all for Bottlerocket.

packages/liburcu/liburcu.spec
```diff
@@ -0,0 +1,65 @@
...
+
+%prep
+%autosetup -n userspace-rcu-%{version}
+
+%build
+%cross_configure
+
+# "fix" rpath
+sed -i 's|^hardcode_libdir_flag_spec=.*|hardcode_libdir_flag_spec=""|g' libtool
+sed -i 's|^runpath_var=LD_RUN_PATH|runpath_var=DIE_RPATH_DIE|g' libtool
```
arnaldo2792: I'm curious why is this needed. Could you please elaborate on why this replacement is needed? Should we do what fedora does?

```spec
# Reinitialize libtool with the fedora version to remove Rpath
autoreconf -vif -W all,error
%configure --disable-static # You could include --disable-static and save yourself the find command below
```

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,93 @@
...
+  --enable-editline=no \
+  --enable-scrub=no
+
+%make_build
+
+%install
+make DIST_ROOT=%{buildroot} install install-dev \
+  PKG_ROOT_SBIN_DIR=%{_cross_sbindir} PKG_ROOT_LIB_DIR=%{_cross_libdir}
+
+rm -f %{buildroot}/%{_cross_libdir}/*.{la,a}
```
arnaldo2792, line 45: Same as elsewhere, is there a flag you can use to disable compiling static libraries?

Dockerfile
arnaldo2792: Should it be something more generic? Like `DATA_PARTITION_FORMAT=<the-format>`? Wouldn't OOTB benefit from this? Say I'm a weird developer that wants to use btrfs for my data partition in my OOT variant.

packages/release/prepare-local-fs.service
```diff
@@ -10,8 +10,9 @@ RefuseManualStop=true
 [Service]
 Type=oneshot
 
+EnvironmentFile=/usr/lib/data-partition-filesystem
```
arnaldo2792: Instead of an environment variable, what if `prepare-local-fs.service` becomes an `in` file, and in the `ExecStart` directive you use `DATA_PARTITION_FILESYSTEM` to replace it with `sed`?

packages/libinih/libinih.spec
```diff
@@ -0,0 +1,44 @@
...
+Source0: https://github.com/benhoyt/inih/archive/refs/tags/r%{version}.tar.gz#/inih-r%{version}.tar.gz
+
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: meson
+
+%description
+%{summary}.
+
+%package devel
+Summary: Files for development using the Simple INI file parser library
```
bcressey: nit:
```suggestion
Summary: Files for development using the simple INI file parser library
```

packages/liburcu/liburcu.spec
```diff
@@ -0,0 +1,65 @@
...
+%exclude %{_cross_libdir}/liburcu-memb.so.8*
+%exclude %{_cross_libdir}/liburcu-qsbr.so.8*
+%exclude %{_cross_libdir}/liburcu-signal.so.8*
+%exclude %{_cross_datadir}/doc/*
+
+%files devel
+%{_cross_includedir}/*
+%{_cross_libdir}/liburcu-common.so
+%{_cross_libdir}/liburcu.so
+%{_cross_libdir}/pkgconfig/liburcu*.pc
```
bcressey: Do these wildcards include headers and `.pc` files for the other liburcu libraries? It'd be good to exclude those.

packages/release/prepare-local-fs.service
bcressey: nit: should use `/usr/share` for non-executable content, and this could have a more generic name to entice us to extend it in the future with other feature flag env variables:
```suggestion
EnvironmentFile=/usr/share/bottlerocket/image-features.env
```

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,91 @@
...
+
+%files
+%license LICENSES/GPL-2.0 LICENSES/LGPL-2.1
+%{_cross_attribution_file}
+%{_cross_libdir}/*.so.*
+%{_cross_sbindir}/fsck.xfs
+%{_cross_sbindir}/mkfs.xfs
+%{_cross_sbindir}/xfs_admin
```
bcressey: Need to exclude all shell scripts.

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,91 @@
...
+%{_cross_sbindir}/xfs_mkfile
+%{_cross_sbindir}/xfs_ncheck
+%{_cross_sbindir}/xfs_quota
+%{_cross_sbindir}/xfs_repair
+%{_cross_sbindir}/xfs_rtcp
+%{_cross_sbindir}/xfs_spaceman
+%{_cross_datadir}/xfsprogs/mkfs/*.conf
+%exclude %{_cross_mandir}
+%exclude %{_cross_datadir}/doc/*
+%exclude %{_cross_datadir}/locale/*
```
bcressey: nit: if you define `PKG_DATA_DIR ` as `%{_cross_datadir}` then you should be able to exclude the normal paths - `%{_cross_localedir}` and `%{_cross_docdir}`.

packages/xfsprogs/xfsprogs.spec
```diff
@@ -0,0 +1,91 @@
...
+%{_cross_includedir}/xfs/linux.h
+%{_cross_includedir}/xfs/xfs.h
+%{_cross_includedir}/xfs/xfs_arch.h
+%{_cross_includedir}/xfs/xfs_fs.h
+%{_cross_includedir}/xfs/xfs_fs_compat.h
+%{_cross_includedir}/xfs/xfs_types.h
+%{_cross_includedir}/xfs/xfs_format.h
+%{_cross_includedir}/xfs/xfs_da_format.h
+%{_cross_includedir}/xfs/xfs_log_format.h
+%{_cross_includedir}/xfs/xqm.h
```
bcressey: nit: this is fine too --
```suggestion
%dir %{_cross_includedir}/xfs
%{_cross_includedir}/xfs/*.h
```

packages/xfsprogs/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[build-dependencies]
+glibc = { path = "../glibc" }
+libinih = { path = "../libinih" }
+liburcu = { path = "../liburcu" }
+util-linux = { path = "../util-linux" }
+
+# RPM Requires
+[dependencies]
+# none - will likely change
```
bcressey: nit: can just remove this section, or just say "none" in the comment.

Mostly I use these as follows:
* `[build-dependencies]` corresponds to `BuildRequires` - if this isn't correct then the build may fail
* `[dependencies]` corresponds to `Requires` - but if it's already listed as a build dependency, we don't need to list it again

tools/buildsys/src/manifest.rs
```diff
@@ -214,6 +214,12 @@ line arguments set in the boot configuration.
 [package.metadata.build-variant.image-features]
 unified-cgroup-hierarchy = true
 ```
+
+`xfs-data-partition` changes the filesystem for the data partition from ext4 to xfs. The
+default will remain ext4 and xfs is opt-in.
+```ignore
+[package.metadata.build-variant.image-features]
+xfs-data-partition = true
```
bcressey: nit: @webern suggested this in the Secure Boot PR. I think it's for consistency with GitHub markdown.

```suggestion
`xfs-data-partition` changes the filesystem for the data partition from ext4 to xfs. The
default will remain ext4 and xfs is opt-in.

```ignore
xfs-data-partition = true
```

tools/rpm2img
```diff
@@ -408,8 +417,28 @@ mkfs_data() {
...
+    # use an internal log with starting size of 64m
+    # use the mininmal 2 Allocation groups, this still overprovisions when expanded
```
bcressey:
```suggestion
    # use the minimal 2 Allocation groups, this still overprovisions when expanded
```

Dockerfile
bcressey: > I didn't end up adding xfsprogs conditionally to the image based upon this but we could if we want.

It's fine to add the bcond support here even if we aren't using it, just for consistency. I did the same in #3097.

bcressey:
I can't seem to respond to [@arnaldo2792 's question](https://github.com/bottlerocket-os/bottlerocket/pull/3198#discussion_r1256300627) in the normal way:
> Should it be something more generic? Like DATA_PARTITION_FORMAT=<the-format>? Wouldn't OOTB benefit from this? Say I'm a weird developer that wants to use btrfs for my data partition in my OOT variant.

If we really want to do this, we can make `data-image-filesystem` a field in the `image-layout` struct (instead of a feature flag), and default it to "ext4".

Dockerfile
yeazelm: I'm having a hard time replying to the other threads about @arnaldo2792's comment:
>Should it be something more generic? Like DATA_PARTITION_FORMAT=<the-format>? Wouldn't OOTB benefit from this? Say I'm a weird developer that wants to use btrfs for my data partition in my OOT variant.

I did look at trying to make it work this way and discovered what @bcressey called out in his comment about `image-layout`. Since `image-features` are booleans these lend themselves as something like `XFS_DATA_PARTITION` vs `DATA_PARTITION_FILESYSTEM` or similar. As @bcressey mentioned, we probably would need to shift to using either `image-layout` (which I feel is a little awkward) or adding in yet another data structure for this. I actually like the idea of leaving the door open for other filesystems in the future but also feel like it might be over-indexing on something we have no concrete plans to do. I can shift to that approach if we think it's useful for OOTB work ensuring we have this flexibility in the future. Ultimately though, my opinion is that there is a lot of other work involved in adding something like `btrfs` and adding in another filesystem would be the right time to solve this more expansively but I could be overruled if that doesn't align with other features.

packages/liburcu/liburcu.spec
yeazelm: From what I understand, we do this fix in a few other autoconf packages (and it works quite well for this package as well). I believe it was introduced in https://github.com/bottlerocket-os/bottlerocket/pull/92 and is there to ensure the rpaths are sane and makes `check-buildroot`|`check-rpaths` happy.

> Should we do what fedora does?

I didn't play with `autoreconf` much but it starts breaking the build in new ways and still doesn't fix the rpaths. Nonetheless the --disable-static might be useful.

packages/release/prepare-local-fs.service
yeazelm: See @bcressey's comment below but I think I'll convert to his approach where its an `image-features.env` and make this a bit more generic for any other features. This allows them to be a bit more consistent. I think I like the `EnvironmentFile` approach a bit more. Is there a reason you would prefer the `in` file and `sed`?

yeazelm: I debated back and forth on `/usr/lib/` and `/usr/share` and apparently debated myself into the wrong place :) I'll fix the path and I really like the more generic option for these variables. Thanks for the suggestion!

packages/xfsprogs/Cargo.toml
yeazelm: I really assumed I would need to change this but missed that the build dependency would solve this. I'll clean up the comment.

packages/xfsprogs/xfsprogs.spec
yeazelm: I didn't see a specific flag for it for `xfsprogs`. Nonetheless, is the approach Fedora takes.

packages/xfsprogs/xfsprogs.spec
yeazelm: Yeah, that is a good point. I meant to call out that we will likely want to replace functionality for, at minimum, `xfs_admin`, `xfs_info`, and `fsck.xfs` with a something written in Rust. I'll move these to excluded for now and we can have another change to add those non-shell script versions into the package.

packages/liburcu/liburcu.spec
yeazelm: Good catch! I'll exclude.

tools/rpm2img
yeazelm: I noticed that it was enabled even without specifying it (which is fine, it only comes into play if the tool making the changes on the filesystem is asking for it) so I decided to add it explicitly in anyway.
```bash
bash-5.1# xfs_spaceman -p xfs_info -c info /local
meta-data=/dev/nvme1n1p1         isize=512    agcount=41, agsize=130816 blks
         =                       reflink=1    bigtime=1 inobtcount=1 nrext64=0
data     =                       bsize=4096   blocks=5242368, imaxpct=25
         =                       sunit=128    swidth=128 blks
naming   =version 2              bsize=16384  ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=16384, version=2
```

tools/rpm2img
```diff
@@ -408,8 +417,28 @@ mkfs_data() {
   target="${1:?}"
   size="${2:?}"
   offset="${3:?}"
-  mkfs.ext4 -m 0 -d "${DATA_MOUNT}" "${BOTTLEROCKET_DATA}" "${size}"
-  echo "${UNLABELED}" | debugfs -w -f - "${BOTTLEROCKET_DATA}"
+  # Create an XFS filesystem if requested
+  if [ "${XFS_DATA_PARTITION}" == "yes" ] ; then
+    echo "writing XFS filesystem for DATA"
+    # Create a file to write the filesystem to first
+    dd if=/dev/zero of="${BOTTLEROCKET_DATA}" bs=1M count=${size%?}
```
bcressey, line 424: nit: `${size}` is known to be non-empty at this point:

```suggestion
    dd if=/dev/zero of="${BOTTLEROCKET_DATA}" bs=1M count=${size}
```

bcressey (approved):

arnaldo2792 (approved):

tools/rpm2img
yeazelm: This is actually trimming the last character off the `${size}` since `dd` wants a count but the value contains an `M` for `mkfs.ext4`. So you end up with `1024` instead of `1024M`.

Issue #819, opened 2020-03-04
missing `objtool` from kernel-devel


bcressey:
While troubleshooting an issue with an out-of-tree module, I found that it failed with an error like this when building in a Docker container:
> make[1]: *** No rule to make target 'kmod.o'

When building in a host container, the same invocation succeeded.

Docker command:
```
docker run -it --rm --net host \
  --mount type=bind,source=${HOME}/mod,target=/tmp/mod \
  --mount type=bind,source=${HOME}/bottlerocket/build/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/modules,target=/lib/modules \
  --mount type=bind,source=${HOME}/bottlerocket/build/x86_64-bottlerocket-linux-gnu/sys-root/usr/share/bottlerocket/squashfs-root,target=/usr/src/kernels \
  bottlerocket/sdk-x86_64:v0.9.0
```

Build command:
```
make -C /lib/modules/5.4.16/build/ M=`pwd` -d modules V=1
```

I compared the output of both builds. The failing build had references to `objtool`:
```
Trying rule prerequisite 'tools/objtool/objtool'.
Looking for a rule with intermediate file 'tools/objtool/objtool'.
Rejecting impossible rule prerequisite 'tools/objtool/objtool'.
```

Adding the following mount option to the Docker command worked around the problem:
```
docker run -it --rm --net host ... \
  --mount type=bind,source=/usr/bin/true,target=/usr/src/kernels/5.4.16/tools/objtool/objtool
  bottlerocket/sdk-x86_64:v0.9.0
```

This may mean we need to find a way to cross-compile `objtool` for our target and include it in the expected path. We might also need to put it in the SDK.

First it would be good to understand why this isn't needed on inside a host container.

sam-aws:
After staring at Makefiles for too long this looks like it comes down to this snippet in the main Makefile:
```
ifdef CONFIG_STACK_VALIDATION
  has_libelf := $(call try-run,\
                echo "int main() {}" | $(HOSTCC) -xc -o /dev/null $(HOST_LIBELF_LIBS) -,1,0)
  ifeq ($(has_libelf),1)
    objtool_target := tools/objtool FORCE
  else
    SKIP_STACK_VALIDATION := 1
    export SKIP_STACK_VALIDATION
  endif
endif
```
[...]

PR #3203, opened 2023-06-15
Add package kmod-6.1-nvidia

### LINKED ISSUE #3085: Update Nvidia driver to 525
**What I'd like:**
Update bottlerocket to version 525 (or at least make it an option) of the NVIDIA driver.  
Nvidia 525 now includes support for CUDA 12.0+ 

I have tried making a variant in Nvidia 525 to update the driver but I am having too many issues to make it work with EKS.

**Please note that the Firmware gsp installation it is now changed for 525.**
"GSP firmware is now distributed as multiple firmware files: this release has gsp_tu10x.bin and gsp_ad10x.bin replacing gsp.bin from previous releases."
Also the libnvidia-egl-wayland.so.1.1.9 was updated to libnvidia-egl-wayland.so.1.1.10

And there is 3 additional libraries in this version:
libcudadebugger.so.1
libcudadebugger.so.525.xx.xx
libnvidia-api.so.1

foersleo:
**Issue number:**

Closes #3085 

**Description of changes:**

Add a package for the nvidia kmod for the new kernel 6.1. With the new kernel we are also updating to the latest available version of the nvidia driver (525.105.17). In the past we had put in place some infrastructure in the spec files for those modules to enable us to ship more than one version of the driver in parallel to the same machine. We have never used that capability and likely will never use it. So along with enabling the new driver for kernel 6.1 I went ahead and cleaned up most of the overhead we had previously implemented for that multi-version enablement.

**Note to reviewers:** I have tried to split the changes into individual commits, each with limited scope, to simplify review. I strongly suggest to review by commit.

**Testing done:**

For testing I have added one [additional commit](https://github.com/foersleo/bottlerocket/commit/136d587e438bef85a0ec8b7730ccd5cbae215b3e) on top of the branch under review in [testing_nvidia_kmod](https://github.com/foersleo/bottlerocket/tree/testing_nvidia_kmod) to build the latest `aws-k8s-1.27-nvidia` variant with kernel 6.1 and the new kmod instead of the standard 5.15.

Testing for successful GPS firmware loading on various hardware and for GPU being accessible inside a pod:

**p3 - NVIDIA V100**

* GSP not in use
```
bash-5.1# /usr/libexec/nvidia/tesla/bin/nvidia-smi -q | grep "GSP Firmware Version"
    GSP Firmware Version                  : N/A
bash-5.1# uname -a
Linux ip-192-168-92-53.us-west-2.compute.internal 6.1.19 #1 SMP PREEMPT_DYNAMIC Thu Jun 15 13:39:41 UTC 2023 x86_64 GNU/Linux
```
* availability in pod:
```
$ kubectl exec ${pod} -- nvidia-smi
Fri Jun 16 14:00:37 2023       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.105.17   Driver Version: 525.105.17   CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla V100-SXM2...  Off  | 00000000:00:1E.0 Off |                    0 |
| N/A   38C    P0    23W / 300W |      0MiB / 16384MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```

**g4dn - NVIDIA T4**
* GSP in use
```
bash-5.1# /usr/libexec/nvidia/tesla/bin/nvidia-smi -q | grep "GSP Firmware Version"
    GSP Firmware Version                  : 525.105.17
bash-5.1# uname -a
Linux ip-192-168-95-166.us-west-2.compute.internal 6.1.19 #1 SMP PREEMPT_DYNAMIC Thu Jun 15 13:39:41 UTC 2023 x86_64 GNU/Linux
```
* availability in pod:
```
$ kubectl exec ${pod} -- nvidia-smi
Fri Jun 16 14:00:40 2023       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.105.17   Driver Version: 525.105.17   CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla T4            Off  | 00000000:00:1E.0 Off |                    0 |
| N/A   25C    P8     9W /  70W |      2MiB / 15360MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```

**g5g - NVIDIA_T4G**
* GSP used
```
bash-5.1# /usr/libexec/nvidia/tesla/bin/nvidia-smi -q | grep "GSP Firmware Version"
    GSP Firmware Version                  : 525.105.17
bash-5.1# uname -a
Linux ip-192-168-42-201.us-west-2.compute.internal 6.1.19 #1 SMP Thu Jun 15 13:56:16 UTC 2023 aarch64 GNU/Linux
```
* availability in pod
```
$ kubectl exec ${pod} -- nvidia-smi
Fri Jun 16 14:00:34 2023       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 525.105.17   Driver Version: 525.105.17   CUDA Version: 12.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  NVIDIA T4G          Off  | 00000000:00:1F.0 Off |                    0 |
| N/A   41C    P8    11W /  70W |      2MiB / 15360MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```

**Terms of contribution:**


iterion:
I tested this change on k8s 1.24 running two different applications and it worked fine for both. I needed this version to get nvidia driver 525 for Vulkan 1.3. Thanks for working on this.

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
```diff
@@ -102,22 +102,20 @@ install -d %{buildroot}%{tesla_525_libdir}
...
 install -m 0644 nvidia-path.env %{buildroot}%{_cross_factorydir}/nvidia/tesla
-# We need to add `_cross_libdir/tesla_525` to the paths loaded by the ldconfig service
+# We need to add `_cross_libdir` to the paths loaded by the ldconfig service
 # because libnvidia-container uses the `ldcache` file created by the service, to locate and mount the
 # libraries into the containers
-sed -e 's|__LIBDIR__|%{_cross_libdir}|' %{S:303} | sed -e 's|__NVIDIA_VERSION__|%{tesla_525}|' \
-  > nvidia-tesla-%{tesla_525}.conf
-install -m 0644 nvidia-tesla-%{tesla_525}.conf %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/ld.so.conf.d/
+sed -e 's|__LIBDIR__|%{_cross_libdir}|' %{S:303} > nvidia-tesla.conf
+install -m 0644 nvidia-tesla.conf %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/ld.so.conf.d/
```
markusboehme, line 125: The unrelated reuse of the `nvidia-tesla.conf` file name is a bit unfortunate: It's both the source of the systemd-tmpfiles configuration and (referencing a file by the same name, generated here) the configuration for `ld` copied into place by that tmpfiles configuration. I'd retain some hint of Source300 being the tmpfiles configuration in its file name (e.g. `nvidia-tesla-tmpfiles.conf`) and specify the target file name when installing: `install -m 0644 %{S:300} %{buildroot}%{_cross_tmpfilesdir}/nvidia-tesla.conf` to remove any ambiguity.

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
```diff
@@ -5,6 +5,17 @@
...
+# With the split of the firmware binary from firmware/gsp.bin to firmware/gsp_ad10x.bin
+# and firmware/gsp_tu10x.bin the file format changed from executable to relocatable.
+# The __spec_install_post macro will by default try to strip all binary files.
+# Unfortunately the strip used is not compatible with the new file format.
+# The macro seems to have an obscure past and nobody really seems to know how it
+# came to be and what its overall purpose is. The process of disabling stripping
+# seems to be to redefine __spec_install_post to nil. However, as we rely on our
+# own __arch_install_post macro that gets invoced through __spec_install_post to
+# generate the _cross_attribution_file. So redefine to directly invoce __arch_install_post.
+%define __spec_install_post %{__arch_install_post}
```
markusboehme: This way of skipping stripping is a bit heavy-handed/unspecific as it cuts out whatever else was defined in `__spec_install_post`. Right now that's:

```
[root@f78bf2a149ef builder]# rpm --eval '%{__os_install_post}'

     
    /usr/lib/rpm/brp-compress 
    
    /usr/lib/rpm/brp-strip /usr/bin/x86_64-bottlerocket-linux-gnu-strip 
    /usr/lib/rpm/brp-strip-comment-note /usr/bin/x86_64-bottlerocket-linux-gnu-strip /usr/bin/x86_64-bottlerocket-linux-gnu-objdump 
     
    /usr/lib/rpm/redhat/brp-strip-lto /usr/bin/x86_64-bottlerocket-linux-gnu-strip 
    /usr/lib/rpm/brp-strip-static-archive /usr/bin/x86_64-bottlerocket-linux-gnu-strip 
    /usr/lib/rpm/check-rpaths 
    /bin/true 
    /usr/lib/rpm/brp-remove-la-files 
    
    /usr/lib/rpm/redhat/brp-python-bytecompile "" "1" "0" 
     
    /usr/lib/rpm/redhat/brp-python-hardlink 
 


```

If only stripping is problematic, you should be able to make that a no-op via `%define __strip /usr/bin/true`. You could even replace that with a wrapper that only skips stripping the firmware files.

markusboehme (commented):
Thanks for taking the time to clean up the multi-version support parts of the code! Has this ever been used or just seemed like a good idea at the time and turned out to not be needed after all?

I'm fine with having the change to the kernel `Makefile` in this PR. Since the commit message already hints that there's two very different changes in that last commit, ideally you could split it before merging (drive-by nit: `HAVE_OBJTOOL` signals whether an arch supports objtool, `OBJTOOL` is selected depending on availability and other options).

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
foersleo: Great point. I was mulling this around in my head back and forth, but did not come to a good conclusion here. I will adjust.

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
foersleo: It is indeed a drastic measure. Admittedly, I went for the easiest way here. I will play around with the suggestions and see what the overall difference is in the different cases.

foersleo:
> Thanks for taking the time to clean up the multi-version support parts of the code! Has this ever been used or just seemed like a good idea at the time and turned out to not be needed after all?

As far as i understand that is exactly what happened here. In the past it seemed like the right thing to do, but turned out we never used it.

 
> I'm fine with having the change to the kernel `Makefile` in this PR. Since the commit message already hints that there's two very different changes in that last commit, ideally you could split it before merging (drive-by nit: `HAVE_OBJTOOL` signals whether an arch supports objtool, `OBJTOOL` is selected depending on availability and other options).

The split makes sense. Regarding the nit: Is it about the commit message claiming it to be architecture dependent? Given that it is more factors than the architecture I think indexing off of `CONFIG_OBJTOOL` would be the right way to go here.

markusboehme:
> Regarding the nit: Is it about the commit message claiming it to be architecture dependent? Given that it is more factors than the architecture I think indexing off of `CONFIG_OBJTOOL` would be the right way to go here.

Indeed only about arch support in the commit message. Agree that making this conditional on `CONFIG_OBJTOOL` is the right call.

packages/kernel-6.1/1001-Makefile-add-prepare-target-for-external-modules.patch
```diff
@@ -23,29 +23,33 @@ existing solutions.
...
++modules_prepare:
 +	$(Q)$(MAKE) $(build)=scripts/basic
 +	$(Q)$(MAKE) $(build)=scripts/dtc
 +	$(Q)$(MAKE) $(build)=scripts/mod
 +	$(Q)$(MAKE) $(build)=scripts
 +
++ifdef CONFIG_OBJTOOL
++prepare: tools/objtool
++endif
++
```
bcressey, line 47: Does this all still work OK when cross-compiling on an x86_64 host to aarch64 target, and vice versa?

foersleo: I checked it worked for building x86_64 and aarch64 from an x86_64 build host. Will try the other way around as well. Is there concerns for this not working out with regards to ordering or with regards to choosing the right thing to do?

Without this I was not able to build the kmod for aarch64 on my x86_64 machine as it would fail when trying to build objtool (not supported for arm64 on 6.1). Either way in the long run this should converge to work as it did for 5.15/5.10 as AL has in the meantime ported their changes for objtool/livepatch to support arm64. It will land with us with the next kernel update.

I think that making this conditional on objtool support for the target is the cleanest way to go (even if the implementation might look different from what I proposed here).

bcressey: > Is there concerns for this not working out with regards to ordering or with regards to choosing the right thing to do?

I've no concerns about the approach and it seems like it should work fine. It's more that `objtool` and aarch64 have been troublesome in the past (e.g. #819), and in that case the fix had to be cleaned up when we added support for builds on an aarch64 host.

arnaldo2792:
The 535 driver just landed, given that the EOL for the 525 driver is December 2023, should we consider using the newer driver?

https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-535-54-03/index.html

foersleo:
> The 535 driver just landed, given that the EOL for the 525 driver is December 2023, should we consider using the newer driver?
> 
> https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-535-54-03/index.html

Yes, I guess we should. I have working through the comments on this one on my agenda for today/tomorrow. So I will have a look. This will probably not be too much effort.

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
foersleo: So I have built this now once with just disabling the stripping and once with the patch as is. As far as I can tell from diffing the resulting rpms with diffoscope the only differences are what is expected from building the same rpm at two different times:

* timestamp differences
* build time differences
* hash differences because the buildtimes are included in the calculation of those checksums

Either way, I will go with the proposed less intrusive change.

foersleo:
The above force-push addressed all the comments:

* fixing the config file naming collision
* reducing scope of disabling stripping of binaries
* split the unrelated changes for kernel-6.1
* Adding the newest driver

For the last point I added extra commits, it was easiest and hopefully helps with reviewing the delta.

packages/kernel-6.1/1001-Makefile-add-prepare-target-for-external-modules.patch
foersleo: I confirmed this is working as expected building both target architectures from x86_64, as well as aarch64 build-hosts.

foersleo:
Testing with the latest 535 driver:

**p3 - NVIDIA V100**

* GSP not in use

```
bash-5.1# /usr/libexec/nvidia/tesla/bin/nvidia-smi -q | grep "GSP Firmware Version"
    GSP Firmware Version                  : N/A
bash-5.1# uname -a
Linux ip-192-168-52-161.us-west-2.compute.internal 6.1.19 #1 SMP PREEMPT_DYNAMIC Thu Jun 29 14:34:38 UTC 2023 x86_64 GNU/Linux
```

* availability in pod:

```
Thu Jun 29 18:17:09 2023       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.54.03              Driver Version: 535.54.03    CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  Tesla V100-SXM2-16GB           Off | 00000000:00:1E.0 Off |                    0 |
| N/A   35C    P0              23W / 300W |      0MiB / 16384MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                         
+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+
```

**g5g - NVIDIA_T4G**

* GSP used

```
bash-5.1# /usr/libexec/nvidia/tesla/bin/nvidia-smi -q | grep "GSP Firmware Version"
    GSP Firmware Version                  : 535.54.03
bash-5.1# uname -a
Linux ip-192-168-84-224.us-west-2.compute.internal 6.1.19 #1 SMP Thu Jun 29 16:59:09 UTC 2023 aarch64 GNU/Linux
```

* availability in pod:
```
Thu Jun 29 18:17:13 2023       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.54.03              Driver Version: 535.54.03    CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA T4G                     Off | 00000000:00:1F.0 Off |                    0 |
| N/A   37C    P8               9W /  70W |      2MiB / 15360MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                         
+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+
```

**g4dn - NVIDIA T4**

* GSP in use

```
bash-5.1# /usr/libexec/nvidia/tesla/bin/nvidia-smi -q | grep "GSP Firmware Version"
    GSP Firmware Version                  : 535.54.03
bash-5.1# uname -a
Linux ip-192-168-24-122.us-west-2.compute.internal 6.1.19 #1 SMP PREEMPT_DYNAMIC Thu Jun 29 14:34:38 UTC 2023 x86_64 GNU/Linux
```

* availability in pod:
```
Thu Jun 29 18:17:16 2023       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.54.03              Driver Version: 535.54.03    CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  Tesla T4                       Off | 00000000:00:1E.0 Off |                    0 |
| N/A   35C    P8              12W /  70W |      2MiB / 15360MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                         
+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+
```

markusboehme (approved):

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
```diff
@@ -0,0 +1,334 @@
...
+%global tesla_ver %{tesla_major}.%{tesla_minor}.%{tesla_patch}
+%global spdx_id %(bottlerocket-license-tool -l %{_builddir}/Licenses.toml spdx-id nvidia)
+%global license_file %(bottlerocket-license-tool -l %{_builddir}/Licenses.toml path nvidia -p ./licenses)
+
+# With the split of the firmware binary from firmware/gsp.bin to firmware/gsp_ga10x.bin
+# and firmware/gsp_tu10x.bin the file format changed from executable to relocatable.
+# The __spec_install_post macro will by default try to strip all binary files.
+# Unfortunately the strip used is not compatible with the new file format.
+# Redefine strip, so that these firmware binaries do not derail the build.
+%define __strip /usr/bin/true
```
bcressey: nit: [prefer `%global` over `%define`](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_global_preferred_over_define).

Commit message should also be updated to refer to this new approach.

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
```diff
@@ -250,7 +250,8 @@ popd
 %{_cross_libdir}/nvidia/tesla/libOpenCL.so.1.0.0
 %{_cross_libdir}/nvidia/tesla/libOpenCL.so.1
 %if "%{_cross_arch}" == "x86_64"
-%{_cross_libdir}/nvidia/tesla/libnvidia-compiler.so.%{tesla_ver}
+%{_cross_libdir}/nvidia/tesla/libnvidia-pkcs11.so.%{tesla_ver}
+%{_cross_libdir}/nvidia/tesla/libnvidia-pkcs11-openssl3.so.%{tesla_ver}
```
bcressey, line 254: Do these libraries have any runtime dependencies on `libssl` or similar?

bcressey (approved):

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
arnaldo2792: They depend on `libcrypto.so`:

```bash
NVIDIA-Linux-x86_64-535.54.03 ❯ readelf -d libnvidia-pkcs11-openssl3.so.535.54.03 | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libcrypto.so.3]
NVIDIA-Linux-x86_64-535.54.03 ❯ readelf -d libnvidia-pkcs11.so.535.54.03 | rg NEEDED
 0x0000000000000001 (NEEDED)             Shared library: [libcrypto.so.1.1]
```

But, the linker should try to load `libcrypto.so` from the container's file system, right? At least `libnvidia-container` won't try to bind-mount `libcrypto` from the host's file system.

arnaldo2792 (approved):
I pulled down the changes, and built a variant. Everything seems to be working fine! :tada:

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
foersleo: As pointed out by Arnaldo those rely on openssl crypto libs. I am not quite sure what to make of this. Looking again through the archives dropped by nvidia there is no mention of these new libraries being added like they have for other changes in their documentation/change notes. The only hint is an extra licensing blurb in the README regarding using pkcs11 header files from the oasis open organization.

I will see if I can come up with some easy testing to validate what Arnaldo mentioned where the container file system would supply that dependency before merging.

foersleo:
:arrow_up: force push was tending Ben's comment about the redefinition of the `__strip` macro and the commit message.

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
foersleo: Learning something new every day! Thanks.

packages/kmod-6.1-nvidia/kmod-6.1-nvidia.spec
foersleo: Digging around in those libraries, they seem to be almost empty "shells" for wrapping pkcs11 functionality. The only function they implement is pkcs11's `GetFunctionList` function to list available pkcs functions. Other than that it just redirects directly to `libcrypto` functions (one library each for `openssl 1.1`, and `openssl 3` implementation).

I can only speculate what these are for, but my bet would be on nvidia shipping libraries of the same name for other hardware that has offloading capabilities and does the crypto calculations on that hardware, while for tesla cards it merely resorts to the software `libcrypto` implementations. They would need the "shell"-library to have a uniform library interface that makes any example code work no matter what the hardware platform is.

Looking at `libnvidia-container` it works just as Arnaldo had pointed out. The container will have to provide necessary libraries outside of nvidia libraries from the container image, including `libcrypto`.

bcressey (approved):

PR #3204, opened 2023-06-15
metricdog: add boot time related metrics to send-boot-success

etungsten:
**Issue number:**

N/A

**Description of changes:**
```
    metricdog: add boot-time-related metrics to send-boot-success
    
    Refactors a bit to accommodate a new trait for checking host status
    using systemd utilities.
```

**Testing done:**
Checked emitted event from `metricdog` upon first boot and was able to serialize out the following:
```json
{
...
  "sender": "metricdog",
  "version": "1.15.0",
  "variant": "aws-k8s-1.27",
  "arch": "x86_64",
  "region": "us-west-2",
  "seed": 2039,
  "version_lock": "latest",
  "ignore_waves": true,
  "event": "boot_success",
  "is_first_boot": true,
  "preconfigured_time_ms": 12841,
  "configured_time_ms": 16096,
  "network_ready_time_ms": 7966,
  "filesystem_ready_time_ms": 4624
}
```

After a reboot, metricdog sent another `send-boot-success` event with the following parameters:
```json
{
..
  "sender": "metricdog",
  "version": "1.15.0",
  "variant": "aws-k8s-1.27",
  "arch": "x86_64",
  "region": "us-west-2",
  "seed": 2039,
  "version_lock": "latest",
  "ignore_waves": true,
  "event": "boot_success",
  "is_first_boot": false,
  "preconfigured_time_ms": 5090,
  "configured_time_ms": 5780,
  "network_ready_time_ms": 4399,
  "filesystem_ready_time_ms": 2072
}
```

Note how `is_first_boot` is no longer true and it took less time for the host to become ready as expected.

Checked a health-ping event to make sure there's no regression:
```json
{
..
  "sender": "metricdog",
  "version": "1.15.0",
  "variant": "aws-k8s-1.27",
  "arch": "x86_64",
  "region": "us-west-2",
  "seed": 2039,
  "version_lock": "latest",
  "ignore_waves": true,
  "event": "health_ping",
  "is_healthy": true,
  "failed_services": {}
}
```

**Terms of contribution:**


webern (approved):

stmcginnis (approved):

sources/metricdog/src/systemd.rs
```diff
@@ -0,0 +1,270 @@
...
+            EXIT_STATUS_PROPERTY,
+        )
+        .map(|exit_code| exit_code.parse::<i32>().ok())
+        .flatten();
+
+        assert!(got.is_none());
+    }
+}
+
+impl HostCheck for SystemdCheck {
```
arnaldo2792: nit: I'd put this before the `parse_property` function or just before the tests.

sources/metricdog/src/host_check.rs
```diff
@@ -0,0 +1,15 @@
...
+
+pub(crate) trait HostCheck {
+    /// Checks whether this current boot is the first boot of the host.
+    fn is_first_boot(&self) -> Result<bool>;
+
+    /// The time, in milliseconds, it took for the host to reach the 'preconfigured' stage.
+    fn preconfigured_time(&self) -> Result<String>;
+
+    /// The time, in milliseconds, it took for the host to reach the 'configured' stage.
+    fn configured_time(&self) -> Result<String>;
```
arnaldo2792: Do you need one as well for `multi-user.target`? The orchestrators' agents are started in this target, they could take a while to initialize, and we may want to know why.

sources/metricdog/src/main.rs
```diff
@@ -113,7 +124,7 @@ pub(crate) fn main_inner(arguments: Arguments, service_check: Box<dyn ServiceChe
     .context(error::BottlerocketReleaseSnafu)?;
 
     // instantiate the metricdog object
-    let metricdog = Metricdog::from_parts(config, os_release, service_check)?;
+    let metricdog = Metricdog::from_parts(config, os_release, service_check, host_check)?;
```
arnaldo2792, line 126: 4 may already be too many arguments. What about a `MetricDogParams` struct (or a different name if you have a better proposal)?. I've seen this is a common practice in other crates (e.g., `rustls`).

sources/metricdog/src/metricdog.rs
```diff
@@ -105,8 +111,29 @@ impl Metricdog {
 
     /// Sends a notification to the metrics url that boot succeeded.
     pub(crate) fn send_boot_success(&self) -> Result<()> {
-        // timeout of 3 seconds to prevent blocking the completion of mark-boot-success
-        self.send("metricdog", "boot_success", None, Some(3))?;
+        let mut values = HashMap::new();
+        // Note whether this is the first boot or not
+        values.insert(
+            "is_first_boot".to_string(),
```
arnaldo2792: Should this be a constant? Same for the others below.

sources/metricdog/src/systemd.rs
```diff
@@ -0,0 +1,270 @@
...
+/// Parses out the systemd unit's exit status from `systemctl show` output
+fn parse_service_exit_code(service: &str) -> Result<Option<i32>> {
+    // we don't check the command's exit code because systemctl returns non-zero codes for various
+    // non-exceptional execution outcomes.
+    let outcome = command(
+        SYSTEMCTL,
+        &["show", "--property", EXIT_STATUS_PROPERTY, service],
+    )?;
+    Ok(parse_property(&outcome.stdout, EXIT_STATUS_PROPERTY)
+        .and_then(|exit_code| exit_code.parse::<i32>().ok()))
```
arnaldo2792, line 84: Probably a `.context` here or in the function's caller to provide a better error other than `Failed to parse i32` since as it is the error will bubble up to `send_health_ping` without providing much context about exactly where the code failed.

sources/metricdog/src/systemd.rs
```diff
@@ -0,0 +1,270 @@
...
+
+/// Utility function to parse out a systemd unit's property value
+fn parse_property(stdout: &str, property: &str) -> Option<String> {
+    trace!(
+        "parsing stdout from 'systemctl show --property {}':\n{}",
+        property,
+        stdout
+    );
+
+    // we expect the response to be formatted like this: `ExecMainStatus=1\n`
```
arnaldo2792: nit: I tend to avoid `we` or `us` in the comments. You could rewrite it as "The format of the response is expected to be: <>", and same for other comments.

sources/metricdog/src/host_check.rs
etungsten: `send-boot-success.service` runs before `multi-user.target` so I would need to capture this event separately and have a separate subcommand for it. I believe in offline discussions we decided to not track orchestrator agent start times for now since there are too many factors involved with service start times in `multi-user.target`. Instead we'll track parts of the boot that are more consistent and not as heavily influenced by different configuration.

etungsten: We can also move `send-boot-success` to `multi-user.target` and have it run after `kubelet`/`ecs-agent` which means we would no longer send a boot success if the orchestration agent fails. I think that change makes sense but I would like to hear some opinions on that.

arnaldo2792: > Instead we'll track parts of the boot that are more consistent and not as heavily influenced by different configuration.

Ok, I agree with this, thanks for the clarification.

sources/metricdog/src/systemd.rs
etungsten: This section was moved as-is as part of the refactoring of `service_check.rs`. The parse does not error and instead gets passed up as a `None` if it fails. If the exit status is `None`, then metricdog only sends the name of the service that fails and not its exit code.

sources/metricdog/src/metricdog.rs
etungsten: I don't think they have to be constants since the values are only going to be referenced here.

sources/metricdog/src/main.rs
etungsten: I don't think adding a struct definition for the sake of being able to pass a single object to a constructor would enhance clarity here. I'm willing to be convinced otherwise.

etungsten:
Push above addresses @arnaldo2792's comments.

Push below rebases onto develop

sources/metricdog/src/systemd.rs
```diff
@@ -0,0 +1,270 @@
...
+        SYSTEMCTL,
+        &["show", "--property", EXIT_STATUS_PROPERTY, service],
+    )?;
+    Ok(parse_property(&outcome.stdout, EXIT_STATUS_PROPERTY)
+        .and_then(|exit_code| exit_code.parse::<i32>().ok()))
+}
+
+impl HostCheck for SystemdCheck {
+    fn is_first_boot(&self) -> Result<bool> {
+        // Upon first boot, we expect this output to have a single line entry leading with '0 <hash>'
```
arnaldo2792: nit: there are still occurrences of `we` or `us` in the PR.

arnaldo2792 (approved):

sources/metricdog/src/host_check.rs
```diff
@@ -0,0 +1,15 @@
...
+
+    /// The time, in milliseconds, it took for the host network to become ready.
+    fn network_ready_time(&self) -> Result<String>;
```
bcressey: The time it took filesystems to become ready would also be interesting. I've found a couple boot time stalls there, and we end up doing more work in some scenarios (e.g. creating the data filesystem).

```suggestion
    /// The time, in milliseconds, it took for the host network to become ready.
    fn network_ready_time(&self) -> Result<String>;
    
    /// The time, in milliseconds, it took for the filesystems to become ready.
    fn filesystem_ready_time(&self) -> Result<String>;
```

sources/metricdog/src/systemd.rs
```diff
@@ -0,0 +1,270 @@
...
+        let outcome = command(JOURNALCTL, &["--list-boots", "-q"])?;
```
bcressey: nit: I prefer not to mix short and long options. The long ones tend to be more self-documenting in that I don't have to go to a manual or run the command to confirm that it does what I expect.

```suggestion
        let outcome = command(JOURNALCTL, &["--list-boots", "--quiet"])?;
```

sources/metricdog/src/metricdog.rs
```diff
@@ -105,8 +111,29 @@ impl Metricdog {
...
+            self.host_check.preconfigured_time()?,
+        );
+        values.insert(
+            "configured_time".to_string(),
+            self.host_check.configured_time()?,
+        );
+        values.insert(
+            "network_ready_time".to_string(),
+            self.host_check.network_ready_time()?,
+        );
```
bcressey: If we want these values to be stable, to simplify querying the results - would it make sense to have a real struct with known, optional fields? Or an enum with two variants - `BootStatus` and `HealthStatus`?

sources/metricdog/src/host_check.rs
```diff
@@ -0,0 +1,15 @@
+use crate::error::Result;
+
+pub(crate) trait HostCheck {
+    /// Checks whether this current boot is the first boot of the host.
+    fn is_first_boot(&self) -> Result<bool>;
+
+    /// The time, in milliseconds, it took for the host to reach the 'preconfigured' stage.
+    fn preconfigured_time(&self) -> Result<String>;
```
bcressey: nit: in general, time fields without an associated unit are harder to reason about:
```suggestion
    /// The time, in milliseconds, it took for the host to reach the 'preconfigured' stage.
    fn preconfigured_time_millis(&self) -> Result<String>;
```

sources/metricdog/src/systemd.rs
etungsten: Removed all instances of 'we" and "us.

etungsten:
Push above addresses @bcressey's comments.

Adds `filesystem_ready_time_ms` as another dimension that tracks when `local-fs.target` becomes active.
Refactors how the query parameters are constructed to use structs and enums to denote fields more stably.

Unit tests pass.

Generated metrics event:
```
{
...
  "sender": "metricdog",
  "version": "1.15.0",
  "variant": "aws-k8s-1.27",
  "arch": "x86_64",
  "region": "us-west-2",
  "seed": 2039,
  "version_lock": "latest",
  "ignore_waves": true,
  "event": "boot_success",
  "is_first_boot": true,
  "preconfigured_time_ms": 12841,
  "configured_time_ms": 16096,
  "network_ready_time_ms": 7966,
  "filesystem_ready_time_ms": 4624
}
```

Updated PR description with new testing.

sources/metricdog/src/systemd.rs
```diff
@@ -0,0 +1,274 @@
...
+            command: cmd,
+            args: args.iter().map(|&s| s.to_owned()).collect::<Vec<String>>(),
+        })?;
+    Ok(Outcome {
+        exit: output.status.code().unwrap_or(-1),
+        stdout: String::from_utf8_lossy(output.stdout.as_slice()).into(),
+    })
+}
+
+fn is_active(service: &str) -> Result<bool> {
```
zmrow, line 61: Nit: My knee-jerk reaction anytime I see an `&str` as a funtion argument is to suggest to use `AsRef<str>` to make it more friendly to use and robust.  Given these functions are only used in this module... I can see why we wouldn't do that.  It just looked odd to my eyes given most of our other code does so.

stmcginnis (approved):
lgtm

stmcginnis (approved):
Better error handling!

zmrow (approved):

bcressey (approved):

PR #3149, opened 2023-05-26
Update rust dependencies

jpculp:
**Description of changes:**

Ran `cargo update` for `sources` and `tools`, avoided some updates that pull unnecessary extra versions, and adjusted the deny TOMLs as necessary. Also included updates for **AWS SDK for Rust**, `coldsnap`, and `tough`.

**Testing done:**

_Coming soon!_

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #3205, opened 2023-06-16
testsys: Upgrade to v0.0.8

ecpullen:
**Issue number:**

Closes #

**Description of changes:**

**Testing done:**

All `cargo make test` features still work as expected.

**Terms of contribution:**

### DIFF
--- a/tools/deny.toml
+++ b/tools/deny.toml
@@ -84,8 +84,8 @@ skip-tree = [
     { name = "windows-sys", version = "=0.42.0" },
 
     # TestSys uses a newer version of base64 and serde_yaml
-    { name = "testsys-model", version = "=0.0.7" },
-    { name = "bottlerocket-types", version = "=0.0.7" },
+    { name = "testsys-model", version = "=0.0.8" },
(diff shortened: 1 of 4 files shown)


gthao313 (approved):

tools/deny.toml
```diff
@@ -84,8 +84,8 @@ skip-tree = [
     { name = "windows-sys", version = "=0.42.0" },
 
     # TestSys uses a newer version of base64 and serde_yaml
-    { name = "testsys-model", version = "=0.0.7" },
-    { name = "bottlerocket-types", version = "=0.0.7" },
+    { name = "testsys-model", version = "=0.0.8" },
+    { name = "bottlerocket-types", version = "=0.0.8" },
```
jpculp, line 88: Can we skip `base64` and `serde_yaml` individually instead of skipping the entire tree here? I think for `base64` you can just remove the explicit version on L71.

etungsten (approved):

tools/deny.toml
jpculp: I think this might be easier to revisit in #3149 if we want to just merge this for now.

Issue #2862, opened 2023-03-07
support all resource limits in settings.oci-defaults


bcressey:
**What I'd like:**
I'd like to be able to set the default values for all resource limits, not just [max-open-files](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/src/modeled_types/oci_defaults.rs#L91).

**Any alternatives you've considered:**
See past discussion in #1703 - there's just not another good way to adjust these settings right now since they can't be set via pod specs.

vyaghras:
Resolved by [#3206](https://github.com/bottlerocket-os/bottlerocket/pull/3206) pull request.

PR #3206, opened 2023-06-16
containerd: resource-limit settings for oci default

### LINKED ISSUE #2814: Expose additional `oci-defaults.resource-limits` settings
**What I'd like:**

Based on comments in https://github.com/bottlerocket-os/bottlerocket/issues/1240#issuecomment-1140401598 - Bottlerocket users need the ability to adjust more of the `settings.oci-defaults.resource-limits` to handle things like memory settings. All OCI runtime settings [0] should be checked for anything useful to expose in Bottlerocket settings.

[0] https://github.com/opencontainers/runtime-spec/blob/main/config.md

vyaghras:
**Issue number:**

Closes #2814  #2862 

**Description of changes:**
Enable the capability to set the values for following resource limits
MaxAddressSpace,
MaxCoreFileSize,
MaxCpuTime,
MaxDataSize,
MaxFileLocks,
MaxFileSize,
MaxLockedMemory,
MaxMsgqueueSize,
MaxNicePriority,
MaxPendingSignals,
MaxProcesses,
MaxRealtimePriority,
MaxRealtimeTimeout,
MaxResidentSet,
MaxStackSize

**Testing done:**
- [x] Try setting a value for all the settings using api client -  It should show these values in etc/containerd/cri-base.json
- [x] Try setting a "unlimited" as hard-limit for max-locked-memory in the settings using api client -  It should show u64::MAX as hard limit for RLIMIT_MEMLOCK in etc/containerd/cri-base.json and the ulimit -Hl should set to  unlimited. 
- [x] Create a pod on bottlerocket host and check /proc/self/limits - It should list all the settings that has been done
         - Create a sample nginx workload container launched via kubernetes and check /proc/self/limits - This has default value for Max locked memory  as 65536.
         - Change Max locked memory setting using api client to 6536.
         - Create another container that runs the nginx image and check the /proc/self/limits - This has default value for Max locked memory  as 6536.
- [x] Migration testing 
         - Create an ami with 1.14.3 version and create instance using that.
         -  Try to set the max-locked-memory setting-  1.14.3 will not allow to do this setting.
         - Upgrade to 1.15.0 with the image in your custom TUF repo.
         - Try to set the max-locked-memory setting-  1.15.0 will allow to do memlock setting and add this in etc/containerd/cri_base.json.
         - Downgrade back to the older version 1.14.3.
         - Try to set the max-locked-memory setting-  1.14.3 will not allow to do this setting.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -756,12 +756,189 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
 Please see the [`getrlimit` linux manpage](https://man7.org/linux/man-pages/man7/capabilities.7.html) for meanings of `hard-limit` and `soft-limit`.
 
 The full list of resource limits that can be configured in Bottlerocket are:
+<table>
+<tr>
+  <th>Resource limit</th>
+  <th>Setting</th>
+  <th>Default value</th>
+  <th>Unit</th>
+</tr>
+<tr>
+  <td rowspan="2"> <code>RLIMIT_AS</code> </td>
+  <td><code>settings.oci-defaults.resource-limits.max-address-space.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-address-space.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_CORE</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-core-file-size.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-core-file-size.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_CPU</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-cpu-time.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">seconds</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-cpu-time.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_DATA</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-data-size.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-data-size.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_LOCKS</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-file-locks.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">locks</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-file-locks.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_FSIZE</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-file-size.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-file-size.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_MEMLOCK</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-locked-memory.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-locked-memory.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_MSGQUEUE</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-msgqueue-size.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-msgqueue-size.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_NICE</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-nice-priority.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">-</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-nice-priority.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_NOFILE</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-open-files.soft-limit</code></td>
+  <td>65536</td>
+  <td rowspan="2">files</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-open-files.hard-limit</code></td>
+  <td>1048576</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_SIGPENDING</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-pending-signals.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">signals</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-pending-signals.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_NPROC</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-processes.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">processes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-processes.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_RTPRIO</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-realtime-priority.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">-</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-realtime-priority.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_RTTIME</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-realtime-timeout.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">microseconds</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-realtime-timeout.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_RSS</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-resident-set.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-resident-set.hard-limit</code></td>
+  <td>-</td>
+</tr>
+<tr>
+  <td rowspan="2"><code>RLIMIT_STACK</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-stack-size.soft-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-stack-size.hard-limit</code></td>
+  <td>-</td>
+</tr>
+</table>
 
-resource limit | setting | default value
------ | ----- | -----
-`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.hard-limit` | 1048576
-`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.soft-limit` | 65536
-
+Limits can be any integer between 0 to `int64::MAX`. Either `-1` or `"unlimited"` can be used to remove the limit.
+* Specifying the maximum value (`i64::MAX`) for a limit:
+  ```toml
+  [settings.oci-defaults.resource-limits.<rlimit>>]
+  soft-limit = 65536
+  hard-limit = 9223372036854775807
+  ```
+* Removing a limit:
+  ```toml
+  [settings.oci-defaults.resource-limits.<rlimit>>]
+  soft-limit = 65536
+  hard-limit = "unlimited"
+  ```
+  
 #### Container image registry settings
 
 The following setting is optional and allows you to configure image registry mirrors and pull-through caches for your containers.
--- a/Release.toml
+++ b/Release.toml
@@ -221,4 +221,7 @@ version = "1.15.0"
     "migrate_v1.14.3_aws-control-container-v0-7-3.lz4",
     "migrate_v1.14.3_public-control-container-v0-7-3.lz4",
 ]
-"(1.14.3, 1.15.0)" = []
+"(1.14.3, 1.15.0)" = [
+    "migrate_v1.15.0_oci-defaults-resource-setting.lz4",
+    "migrate_v1.15.0_oci-defaults-max-open-files.lz4",
+]
--- a/sources/api/migration/migrations/v1.15.0/oci-defaults-max-open-files/src/main.rs
+++ b/sources/api/migration/migrations/v1.15.0/oci-defaults-max-open-files/src/main.rs
@@ -1,0 +1,68 @@
+use migration_helpers::common_migrations::NoOpMigration;
+use migration_helpers::{migrate, Migration, MigrationData, Result};
+use serde_json::Value;
+use std::process;
+
+const HARD_RESOURCE_LIMIT_SETTING_NAME: &str =
+    "settings.oci-defaults.resource-limits.max-open-files.hard-limit";
+const SOFT_RESOURCE_LIMIT_SETTING_NAME: &str =
+    "settings.oci-defaults.resource-limits.max-open-files.soft-limit";
+
+/// This migration changes the hard and soft limit for rlimit_nofile to u32 from i64 on downgrade.
+/// There is no need of migration on upgrade as u32 will automatically change to i64
+pub struct ChangeMaxOpenFileResourceLimitType;
+
+fn convert_to_u32(value: &mut Value) {
+    if !value.is_i64() {
+        return;
+    }
+    let v: i64 = serde_json::from_value(value.clone()).unwrap();
+    let s = match v {
+        -1 => u32::MAX,
+        v if v > u32::MAX as i64 => u32::MAX,
+        _ => v as u32,
+    };
+
+    *value = Value::Number(s.into());
+}
+
+impl Migration for ChangeMaxOpenFileResourceLimitType {
+    /// On upgrade there is nothing to do (see above).
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        Ok(input)
+    }
+
+    /// On downgrade, if the value is an i64 integer, we need to convert it to a u32.
+    ///
+    /// Note that this potentially causes data loss, if current value of the setting
+    /// is -1 or higher than u_32::MAX we will set it to max possible value i.e. u32::MAX.
+    fn backward(&mut self, mut input: MigrationData) -> Result<MigrationData> {
+        if let Some(v) = input.data.get_mut(HARD_RESOURCE_LIMIT_SETTING_NAME) {
+            convert_to_u32(v);
+        }
+        if let Some(v) = input.data.get_mut(SOFT_RESOURCE_LIMIT_SETTING_NAME) {
+            convert_to_u32(v);
+        }
+        Ok(input)
+    }
+}
+
+fn run() -> Result<()> {
+    if cfg!(variant_runtime = "k8s") {
+        migrate(ChangeMaxOpenFileResourceLimitType)?
+    } else {
+        migrate(NoOpMigration)?;
+    }
+
+    Ok(())
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/api/migration/migrations/v1.15.0/oci-defaults-resource-setting/src/main.rs
+++ b/sources/api/migration/migrations/v1.15.0/oci-defaults-resource-setting/src/main.rs
@@ -1,0 +1,40 @@
+use migration_helpers::common_migrations::{AddPrefixesMigration, NoOpMigration};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added new resource limit settings for configuring the default OCI runtime spec.
+fn run() -> Result<()> {
+    if cfg!(variant_runtime = "k8s") {
+        migrate(AddPrefixesMigration(vec![
+            "settings.oci-defaults.resource-limits.max-address-space",
+            "settings.oci-defaults.resource-limits.max-core-file-size",
+            "settings.oci-defaults.resource-limits.max-cpu-time",
+            "settings.oci-defaults.resource-limits.max-data-size",
+            "settings.oci-defaults.resource-limits.max-file-locks",
+            "settings.oci-defaults.resource-limits.max-file-size",
+            "settings.oci-defaults.resource-limits.max-locked-memory",
+            "settings.oci-defaults.resource-limits.max-msgqueue-size",
+            "settings.oci-defaults.resource-limits.max-nice-priority",
+            "settings.oci-defaults.resource-limits.max-pending-signals",
+            "settings.oci-defaults.resource-limits.max-processes",
+            "settings.oci-defaults.resource-limits.max-realtime-priority",
+            "settings.oci-defaults.resource-limits.max-realtime-timeout",
+            "settings.oci-defaults.resource-limits.max-resident-set",
+            "settings.oci-defaults.resource-limits.max-stack-size",
+        ]))?
+    } else {
+        migrate(NoOpMigration)?;
+    }
+
+    Ok(())
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/api/schnauzer/src/helpers.rs
+++ b/sources/api/schnauzer/src/helpers.rs
@@ -1496,8 +1496,8 @@ fn oci_spec_resource_limits(value: &Value) -> Result<String, RenderError> {
             format!(
                 r#"{{ "type": "{}", "hard": {}, "soft": {} }}"#,
                 rlimit_type.to_linux_string(),
-                values.hard_limit,
-                values.soft_limit,
+                values.get_hard_limit(),
+                values.get_soft_limit(),
             )
         })
         .collect::<Vec<String>>()
@@ -2703,6 +2703,7 @@ mod test_any_enabled {
 mod test_oci_spec {
     use crate::helpers::*;
     use serde_json::json;
+    use OciDefaultsResourceLimitType::*;
 
     #[test]
     fn oci_spec_capabilities_test() {
@@ -2728,13 +2729,63 @@ mod test_oci_spec {
         );
     }
 
+    fn check_all_rlimits(
+        (cap, bottlerocket, hard_limit, soft_limit): (OciDefaultsResourceLimitType, &str, i64, i64),
+    ) {
+        let json = json!({bottlerocket: {"hard-limit": hard_limit, "soft-limit": soft_limit}});
+        let rendered = oci_spec_resource_limits(&json).unwrap();
+        let result = format!(
+            r#"{{ "type": "{}", "hard": {}, "soft": {} }}"#,
+            cap.to_linux_string(),
+            hard_limit,
+            soft_limit,
+        );
+        assert_eq!(rendered, result);
+    }
+
     #[test]
     fn oci_spec_resource_limits_test() {
-        let json = json!({"max-open-files": {"hard-limit": 1, "soft-limit": 2}});
+        let rlimits = [
+            (MaxAddressSpace, "max-address-space", 2, 1),
+            (MaxCoreFileSize, "max-core-file-size", 4, 3),
+            (MaxCpuTime, "max-cpu-time", 5, 4),
+            (MaxDataSize, "max-data-size", 6, 5),
+            (MaxFileLocks, "max-file-locks", 7, 6),
+            (MaxFileSize, "max-file-size", 8, 7),
+            (MaxLockedMemory, "max-locked-memory", 9, 8),
+            (MaxMsgqueueSize, "max-msgqueue-size", 10, 9),
+            (MaxNicePriority, "max-nice-priority", 11, 10),
+            (MaxOpenFiles, "max-open-files", 12, 11),
+            (MaxPendingSignals, "max-pending-signals", 17, 15),
+            (MaxProcesses, "max-processes", 13, 12),
+            (MaxRealtimePriority, "max-realtime-priority", 18, 23),
+            (MaxRealtimeTimeout, "max-realtime-timeout", 14, 13),
+            (MaxResidentSet, "max-resident-set", 15, 14),
+            (MaxStackSize, "max-stack-size", 16, 15),
+        ];
+
+        for rlimit in rlimits {
+            check_all_rlimits(rlimit);
+        }
+    }
+
+    #[test]
+    fn oci_spec_max_locked_memory_as_unlimited_resource_limit_test() {
+        let json = json!({"max-locked-memory": {"hard-limit": "unlimited", "soft-limit": 18}});
         let rendered = oci_spec_resource_limits(&json).unwrap();
         assert_eq!(
             rendered,
-            r#"{ "type": "RLIMIT_NOFILE", "hard": 1, "soft": 2 }"#
+            r#"{ "type": "RLIMIT_MEMLOCK", "hard": 18446744073709551615, "soft": 18 }"#
+        );
+    }
+
+    #[test]
+    fn oci_spec_max_locked_memory_as_minus_one_resource_limit_test() {
+        let json = json!({"max-locked-memory": {"hard-limit": -1, "soft-limit": 18}});
+        let rendered = oci_spec_resource_limits(&json).unwrap();
+        assert_eq!(
+            rendered,
+            r#"{ "type": "RLIMIT_MEMLOCK", "hard": 18446744073709551615, "soft": 18 }"#
         );
     }
 }
--- a/sources/models/src/de.rs
+++ b/sources/models/src/de.rs
@@ -207,3 +207,127 @@ mod mirrors_tests {
         );
     }
 }
+
+// This specifies that any non negative i64 integer, -1, and "unlimited"
+/// are the valid resource-limits. The hard-limit set to "unlimited" or -1
+/// and soft-limit set to "unlimited" or -1 are converted to u64::MAX in
+/// the spec file for the container runtime which ultimately represents
+/// unlimited for that resource
+pub(crate) fn deserialize_limit<'de, D>(deserializer: D) -> Result<i64, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum StringOrInt64 {
+        String(String),
+        Int(i64),
+    }
+
+    match StringOrInt64::deserialize(deserializer)? {
+        StringOrInt64::String(s) => {
+            if s == "unlimited" {
+                Ok(-1)
+            } else {
+                Err(Error::custom(format!(
+                    "Invalid rlimit {}, expected -1 to {} or \"unlimited\"",
+                    s,
+                    i64::MAX
+                )))
+            }
+        }
+        StringOrInt64::Int(i) => {
+            if (-1..=i64::MAX).contains(&i) {
+                Ok(i)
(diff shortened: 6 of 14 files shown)


stmcginnis (commented):
Looks like it's doing what's expected. I ran:

```sh
apiclient apply <<EOF
[settings.oci-defaults.resource-limits.max-address-space]
hard-limit = 2000000000
soft-limit = 1000000000

[settings.oci-defaults.resource-limits.max-core-file-size]
hard-limit = 2000000000
soft-limit = 1000000000

[settings.oci-defaults.resource-limits.max-cpu-time]
hard-limit = 10000
soft-limit = 1000

[settings.oci-defaults.resource-limits.max-data-size]
hard-limit = 1000000000
soft-limit = 100000000

[settings.oci-defaults.resource-limits.max-file-locks]
hard-limit = 2048
soft-limit = 1024

[settings.oci-defaults.resource-limits.max-file-size]
hard-limit = 204800
soft-limit = 102400

[settings.oci-defaults.resource-limits.max-locked-memory]
hard-limit = 1000000000
soft-limit = 100000000

[settings.oci-defaults.resource-limits.max-msgqueue-size]
hard-limit = 1000000001
soft-limit = 1000000000

[settings.oci-defaults.resource-limits.max-nice-priority]
hard-limit = 7
soft-limit = 6

[settings.oci-defaults.resource-limits.max-open-files]
hard-limit = 1048576
soft-limit = 65536

[settings.oci-defaults.resource-limits.max-pending-signals]
hard-limit = 1000
soft-limit = 500

[settings.oci-defaults.resource-limits.max-processes]
hard-limit = 10000
soft-limit = 9999

[settings.oci-defaults.resource-limits.max-realtime-priority]
hard-limit = 1000000000
soft-limit = 100000000

[settings.oci-defaults.resource-limits.max-realtime-timeout]
hard-limit = 1000000000
soft-limit = 100000000

[settings.oci-defaults.resource-limits.max-resident-set]
hard-limit = 1000000000
soft-limit = 100000000

[settings.oci-defaults.resource-limits.max-stack-size]
hard-limit = 1000000000
soft-limit = 100000000
EOF
```

Which results in `/etc/containerd/cri-base.json` containing:

```json
"rlimits": [{ "type": "RLIMIT_SIGPENDING", "hard": 1000, "soft": 500 },
{ "type": "RLIMIT_RTPRIO", "hard": 1000000000, "soft": 100000000 },
{ "type": "RLIMIT_RSS", "hard": 1000000000, "soft": 100000000 },
{ "type": "RLIMIT_NICE", "hard": 7, "soft": 6 },
{ "type": "RLIMIT_CORE", "hard": 2000000000, "soft": 1000000000 },
{ "type": "RLIMIT_NPROC", "hard": 10000, "soft": 9999 },
{ "type": "RLIMIT_AS", "hard": 2000000000, "soft": 1000000000 },
{ "type": "RLIMIT_RTTIME", "hard": 1000000000, "soft": 100000000 },
{ "type": "RLIMIT_NOFILE", "hard": 1048576, "soft": 65536 },
{ "type": "RLIMIT_STACK", "hard": 1000000000, "soft": 100000000 },
{ "type": "RLIMIT_CPU", "hard": 10000, "soft": 1000 }, 
{ "type": "RLIMIT_MEMLOCK", "hard": 1000000000, "soft": 100000000 },
{ "type": "RLIMIT_FSIZE", "hard": 204800, "soft": 102400 },
{ "type": "RLIMIT_MSGQUEUE", "hard": 1000000001, "soft": 1000000000 },
{ "type": "RLIMIT_LOCKS", "hard": 2048, "soft": 1024 },
{ "type": "RLIMIT_DATA", "hard": 1000000000, "soft": 100000000 }]
```

I do think you will need an `AddPrefixesMigration` migration though since these values didn't exist before. Might be good to confirm that with someone else that has a stronger grasp of our migration/datastore handling though.

cbgbt:
> I do think you will need an `AddPrefixesMigration`

That's correct. Consider the following workflow:
* You are running Bottlerocket X
* You upgrade to Y, which adds these new RLIMIT attributes to the model
* You decide to set `settings.oci-defaults.resource-limits.max-resident-set`
* You downgrade Bottlerocket from Y back to X. Version X of Bottlerocket doesn't know what `max-resident-set` is, so it fails to load the model.


An `AddPrefixesMigration` ensures that any new settings prefixes which won't be understood on older versions of Bottlerocket are removed on downgrade. You'd need to specify one of these migrations for each of the new types added (All in the same binary, `AddPrefixesMigration` accepts a list.)

sources/api/schnauzer/src/helpers.rs
```diff
@@ -2737,4 +2737,14 @@ mod test_oci_spec {
             r#"{ "type": "RLIMIT_NOFILE", "hard": 1, "soft": 2 }"#
         );
     }
+
+    #[test]
+    fn oci_spec_mamlock_resource_limits_test() {
```
yeazelm:
```suggestion
    fn oci_spec_memlock_resource_limits_test() {
```

sources/models/src/modeled_types/oci_defaults.rs
```diff
@@ -125,5 +155,11 @@ mod oci_defaults_rlimits {
             "max-open-files",
             "RLIMIT_NOFILE",
         );
+
+        check_rlimit_strings(
+            OciDefaultsResourceLimitType::MaxLockedMemory,
+            "max-locked-memory",
+            "RLIMIT_MEMLOCK",
+        );
```
yeazelm: Is there a reason there is only this one new test and not one for all of the new settings we are adding? What is special about this setting?

vyaghras: There is no strong reason for testing only this one but as this test is to check the mapping so I thought it will be redundant to test for all the settings.

sources/api/migration/migrations/v1.15.0/oci-defaults-resource-setting/src/main.rs
```diff
@@ -0,0 +1,43 @@
+use migration_helpers::common_migrations::{AddPrefixesMigration, NoOpMigration};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added new resource limit settings for configuring the default OCI runtime spec,
+/// `settings.oci-defaults.resource-limits`, which will initially contain only
+/// `settings.oci-defaults.resource-limits.max-open-files`. We have to remove other
+/// settings on downgrade.
```
etungsten: Should update this comment. Can just generically describe that we've added additional resource limits.

etungsten (approved):
nice! :rocket:

stmcginnis:
This can now be rebased on the latest `develop` to pick up a pin of the rust toolchain version used on the GitHub worker nodes. That should allow the checks to pass. This is a temporary workaround until the SDK is bumped to use rust 1.70.0.

stmcginnis (approved):

sources/models/src/modeled_types/oci_defaults.rs
yeazelm: I find it a little odd to have only 2 unit tests for two of the sixteen settings we have here. Either testing that any one comes through correctly as a basic test makes sense or testing that each setting ends up where we want it makes sense, but just these two are hard to reason about. Its a lot of extra boilerplate but having one test for each setting is my preference since it would cover each case.

vyaghras: Sure Its worth to have more checks always.. I will add one test for all 16 settings

yeazelm (approved):
This looks good. I do wonder if we do any validation on these values. For example, setting a soft limit above the hard limit is not possible or setting a value for nice outside of  -20 - 19. I'm not sure what we would do in those cases, but its worth considering if there is anything we would like to do there.

Release.toml
```diff
@@ -212,4 +212,6 @@ version = "1.15.0"
     "migrate_v1.14.0_public-control-container-v0-7-2.lz4",
 ]
 "(1.14.0, 1.14.1)" = []
-"(1.14.1, 1.15.0)" = []
+"(1.14.1, 1.15.0)" = [
+    "migrate_v1.15.0_oci-defaults-resource-setting.lz4",
+]
```
bcressey: after rebase, this now needs to be:
```suggestion
"(1.14.2, 1.15.0)" = [
    "migrate_v1.15.0_oci-defaults-resource-setting.lz4",
]
```

sources/api/migration/migrations/v1.15.0/oci-defaults-resource-setting/src/main.rs
```diff
@@ -0,0 +1,43 @@
+use migration_helpers::common_migrations::{AddPrefixesMigration, NoOpMigration};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added new resource limit settings for configuring the default OCI runtime spec
+/// like max-address-space, max-core-file-size, max-cpu-time, max-data-size, max-file-locks
+/// max-file-size, max-locked-memory, max-msgqueue-size, max-nice-priority, max-pending-signals,
+/// max-processes, max-realtime-priority, max-realtime-timeout, max-resident-set, max-stack-size.
```
bcressey: nit: no need to list all of these prefixes again:
```suggestion
/// We added new resource limit settings for configuring the default OCI runtime spec.
```

sources/models/src/modeled_types/oci_defaults.rs
```diff
@@ -120,10 +150,100 @@ mod oci_defaults_rlimits {
 
     #[test]
     fn linux_rlimit_strings() {
+        check_rlimit_strings(
+            OciDefaultsResourceLimitType::MaxAddressSpace,
+            "max-address-space",
+            "RLIMIT_AS",
+        );
```
bcressey: nit: consider doing something like this:
```
        use OciDefaultsResourceLimitType::*;
        let rlimits = [
            (MaxAddressSpace, "max-address-space", "RLIMIT_AS"),
            (MaxCoreFileSize, "max-core-file-size", "RLIMIT_CORE"),
            ...
        ];

        for rlimit in rlimits {
            check_rlimit_strings(rlimit);
        }
```

With a slight tweak to the signature for `check_rlimit_strings` to take a tuple instead:
```
    fn check_rlimit_strings((cap, bottlerocket, linux): (OciDefaultsResourceLimitType, &str, &str)) { /* ... */ }
```

That way, the implementation matches the larger intent of calling the same function in the same way for each of the enum's variants. If some other function ever needs to be called, it will be obvious that it's an exception since it won't fit in the loop without special casing of some sort.

It also simplifies the reviewer / reader's job - it's just a question of whether the inputs are comprehensive, and not whether the right function is called in the right way at each call site.

sources/models/src/modeled_types/oci_defaults.rs
```diff
@@ -94,15 +94,45 @@ mod oci_defaults_capabilities {
...
+    MaxResidentSet,
+    MaxStackSize,
 }
 
 // We are leaving this implementation open for the easy addition of
 // future resource limits.
 impl OciDefaultsResourceLimitType {
     pub fn to_linux_string(&self) -> &'static str {
         match self {
+            OciDefaultsResourceLimitType::MaxAddressSpace => "RLIMIT_AS",
```
bcressey: nit: you can clean up some of the repetition with a `use` statement:
```suggestion
    pub fn to_linux_string(&self) -> &'static str {
        use OciDefaultsResourceLimitType::*;
        match self {
            MaxAddressSpace => "RLIMIT_AS",
```

README.md
```diff
@@ -749,10 +749,40 @@ Please see the [`getrlimit` linux manpage](https://man7.org/linux/man-pages/man7
...
 The full list of resource limits that can be configured in Bottlerocket are:
 
-resource limit | setting | default value
------ | ----- | -----
-`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.hard-limit` | 1048576
-`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.soft-limit` | 65536
+resource limit | setting | default value | unit
+----- | ----- | ----- | -----
+`RLIMIT_AS` | `settings.oci-defaults.resource-limits.max-address-space.hard-limit` | - | bytes
+`RLIMIT_AS` | `settings.oci-defaults.resource-limits.max-address-space.soft-limit` | - | bytes
```
bcressey: If we use an HTML table instead, we could use `rowspan=2` for the resource limit and unit fields to avoid duplicating those values. Might be worth trying out to see if the result seems more polished.

README.md
```diff
@@ -749,10 +749,40 @@ Please see the [`getrlimit` linux manpage](https://man7.org/linux/man-pages/man7
...
+`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.hard-limit` | 1048576 | files
+`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.soft-limit` | 65536 | files
+`RLIMIT_SIGPENDING` | `settings.oci-defaults.resource-limits.max-pending-signals.hard-limit` | - | signals
+`RLIMIT_SIGPENDING` | `settings.oci-defaults.resource-limits.max-pending-signals.soft-limit` | - | signals
+`RLIMIT_NPROC` | `settings.oci-defaults.resource-limits.max-processes.hard-limit` | - | processes
+`RLIMIT_NPROC` | `settings.oci-defaults.resource-limits.max-processes.soft-limit` | - | processes
+`RLIMIT_RTPRIO` | `settings.oci-defaults.resource-limits.max-realtime-priority.hard-limit` | - | -
+`RLIMIT_RTPRIO` | `settings.oci-defaults.resource-limits.max-realtime-priority.soft-limit` | - | -
+`RLIMIT_RTTIME` | `settings.oci-defaults.resource-limits.max-realtime-timeout.hard-limit` | - | us
+`RLIMIT_RTTIME` | `settings.oci-defaults.resource-limits.max-realtime-timeout.soft-limit` | - | us
```
bcressey: nit: `microseconds` might be more readable than `us`:
```suggestion
`RLIMIT_RTTIME` | `settings.oci-defaults.resource-limits.max-realtime-timeout.hard-limit` | - | microseconds
`RLIMIT_RTTIME` | `settings.oci-defaults.resource-limits.max-realtime-timeout.soft-limit` | - | microseconds
```

bcressey (commented):
It'd be nice to confirm that `-1` (for "infinity" / "unlimited") is accepted everywhere, or else to document what the right value to use is in order to get that behavior:

```
[settings.oci-defaults.resource-limits.max-address-space]
hard-limit = -1
soft-limit = -1
```

I'm not confident that `-1` will deserialize into `u32::MAX`, and I could see someone wanting to raise or remove a limit at runtime that they specified at launch time. That could be a problem since we don't allow settings to be deleted.

README.md
```diff
@@ -755,11 +755,190 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+<table>
+<tr>
+  <th>resource limit</th>
+  <th>setting</th>
+  <th>default value</th>
+  <th>unit</th>
+</tr>
+<tr>
+  <td rowspan="2">RLIMIT_AS</td>
+  <td>settings.oci-defaults.resource-limits.max-address-space.hard-limit</td>
```
arnaldo2792: You are missing " \`\` "  in all the settings. Same for `RLIMIT_<>`.

README.md
```diff
@@ -755,11 +755,190 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
 
 The full list of resource limits that can be configured in Bottlerocket are:
-
-resource limit | setting | default value
------ | ----- | -----
-`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.hard-limit` | 1048576
-`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.soft-limit` | 65536
+<table>
+<tr>
+  <th>resource limit</th>
```
arnaldo2792: nit: capiltalize for each first letter

README.md
```diff
@@ -755,11 +755,190 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
 Please see the [`getrlimit` linux manpage](https://man7.org/linux/man-pages/man7/capabilities.7.html) for meanings of `hard-limit` and `soft-limit`.
 
 The full list of resource limits that can be configured in Bottlerocket are:
-
-resource limit | setting | default value
------ | ----- | -----
-`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.hard-limit` | 1048576
-`RLIMIT_NOFILE` | `settings.oci-defaults.resource-limits.max-open-files.soft-limit` | 65536
+<table>
```
stmcginnis, line 759: It's highly unlikely we will want to render into anything other than HTML, but is there a reason for using HTML tags directly rather than sticking with markdown table syntax?

vyaghras: So that we can use rowspan. This will avoid the repetition of `RLIMIT_NOFILE`  rlimit names.

README.md
```diff
@@ -755,11 +755,190 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+  <th>Resource limit</th>
+  <th>Setting</th>
+  <th>Default value</th>
+  <th>Unit</th>
+</tr>
+<tr>
+  <td rowspan="2"> <code>RLIMIT_AS</code> </td>
+  <td><code>settings.oci-defaults.resource-limits.max-address-space.hard-limit</code></td>
+  <td>-</td>
+  <td>bytes</td>
```
```suggestion
  <td rowspan="2">bytes</td>
```

bcressey (commented):
Code LGTM. One nit on the doc table, and I still have that open question about how to specify "unlimited" via settings, and making sure that works.

vyaghras:
> Code LGTM. One nit on the doc table, and I still have that open question about how to specify "unlimited" via settings, and making sure that works.

Now we can use "unlimited" string to set any rlimit as unlimited.

README.md
```diff
@@ -755,12 +755,181 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-stack-size-limit</code></td>
+  <td>-</td>
+</tr>
+</table>
+
+**Note**:  String `unlimited` can be used to set resource-limit as unlimited as
```
yeazelm:
```suggestion
**Note**:  The string `unlimited` can be used to set resource-limit as unlimited. For example, to set the hard limit to unlimited:
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -2737,4 +2737,151 @@ mod test_oci_spec {
             r#"{ "type": "RLIMIT_NOFILE", "hard": 1, "soft": 2 }"#
         );
     }
+
+    #[test]
+    fn oci_spec_max_address_space_resource_limits_test() {
+        let json = json!({"max-address-space": {"hard-limit": 3, "soft-limit": 4}});
+        let rendered = oci_spec_resource_limits(&json).unwrap();
+        assert_eq!(rendered, r#"{ "type": "RLIMIT_AS", "hard": 3, "soft": 4 }"#);
+    }
```
yeazelm: I'd like to see a test to confirm `unlimited` makes it through since we do some custom deserialization and e2e tests are a nice way to make sure things haven't shifted under us when making changes.

sources/models/src/modeled_types/oci_defaults.rs
```diff
@@ -120,10 +157,155 @@ mod oci_defaults_rlimits {
...
+        ];
+
+        for rlimit in rlimits {
+            check_rlimit_strings(rlimit);
+        }
+    }
+}
+
+/// OciDefaultsResourceLimitValue specifies what all values are
+/// allowed to be set for rlimit in default spec.
```
yeazelm: We should call out `unlimited` and how this special case needs to have something more than a `u32`.

sources/models/src/modeled_types/oci_defaults.rs
```diff
@@ -120,10 +157,155 @@ mod oci_defaults_rlimits {
...
+
+                    Ok(OciDefaultsResourceLimitValue::Value(s))
+                } else {
+                    Err(de::Error::custom(format!(
+                        "Provided integer is out of range: {}",
+                        number
+                    )))
+                }
+            }
+            Value::String(s) => match s.to_lowercase().as_str() {
```
yeazelm: I have mixed feelings on doing `to_lowercase()` here. It's a balance between ease of use and correctness. I'm unfamiliar with if we typically do this elsewhere in settings but we should remain consistent with the rest of them.

sources/models/src/modeled_types/oci_defaults.rs
```diff
@@ -120,10 +157,155 @@ mod oci_defaults_rlimits {
...
+                        de::Error::custom(format!(
+                            "Unable to parse {} as a valid rlimit value: {}",
+                            number, e
+                        ))
+                    })?;
+
+                    Ok(OciDefaultsResourceLimitValue::Value(s))
+                } else {
+                    Err(de::Error::custom(format!(
+                        "Provided integer is out of range: {}",
```
yeazelm: `Value::Number` can be a float too, so `1.0000` is in the range but not an `u64`. I would rather this error be more specific "Provided number is not within the range of 0 - {}: {}" and then put u64::MAX in there to provide the most correct error.

sources/models/src/modeled_types/oci_defaults.rs
```diff
@@ -120,10 +157,155 @@ mod oci_defaults_rlimits {
...
+    where
+        D: Deserializer<'de>,
+    {
+        let json_value = Value::deserialize(deserializer)?;
+
+        return match &json_value {
+            Value::Number(number) => {
+                if number.is_u64() {
+                    let s = number.to_string().parse::<u32>().map_err(|e| {
+                        de::Error::custom(format!(
```
yeazelm: I believe we use `Snafu` for these errors in `modeled_types` so I would like to use that for these errors.

vyaghras: Serde expect errors while deserializing  in form of D::Error, hence we need to return the error in form of D::Error.

yeazelm: These custom errors only in Serde deserializers makes sense.

README.md
```diff
@@ -755,12 +755,195 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+<tr>
+  <td rowspan="2"><code>RLIMIT_STACK</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-stack-size-limit</code></td>
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-stack-size-limit</code></td>
+  <td>-</td>
+</tr>
```
```suggestion
<tr>
  <td rowspan="2"><code>RLIMIT_STACK</code></td>
  <td><code>settings.oci-defaults.resource-limits.max-stack-size.hard-limit</code></td>
  <td>-</td>
  <td rowspan="2">bytes</td>
</tr>
<tr>
  <td><code>settings.oci-defaults.resource-limits.max-stack-size.soft-limit</code></td>
  <td>-</td>
</tr>
```

README.md
```diff
@@ -755,12 +755,195 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+<tr>
+  <td rowspan="2"><code>RLIMIT_NOFILE</code></td>
+  <td><code>settings.oci-defaults.resource-limits.max-open-files.hard-limit</code></td>
+  <td>1048576</td>
+  <td rowspan="2">files</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-open-files.soft-limit</code></td>
+  <td>65536</td>
+</tr>
```
bcressey: nit: having "soft" come before "hard" would sort the numbers numerically, but should be done for all other limits too.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -2737,4 +2737,160 @@ mod test_oci_spec {
...
+
+    #[test]
+    fn oci_spec_max_realtime_timeout_resource_limits_test() {
+        let json = json!({"max-realtime-timeout": {"hard-limit": 15, "soft-limit": 16}});
+        let rendered = oci_spec_resource_limits(&json).unwrap();
+        assert_eq!(
+            rendered,
+            r#"{ "type": "RLIMIT_RTTIME", "hard": 15, "soft": 16 }"#
+        );
+    }
```
bcressey: nit: these could probably all be combined into one test that tests everything.

Typically when there are multiple test cases, I'd expect to see boundary testing - largest positive and smallest negative numbers, hard less than soft, that sort of thing - in addition to the test that exercises the happy path functionality.

I don't think you need more than two cases - one that checks that everything can be rendered, and one that checks for unlimited handling specifically.

sources/models/src/lib.rs
```diff
@@ -516,8 +517,8 @@ struct OciDefaults {
 ///// The hard and soft limit values for an OCI defaults resource limit.
 #[model(add_option = false)]
 struct OciDefaultsResourceLimit {
-    hard_limit: u32,
-    soft_limit: u32,
+    hard_limit: OciDefaultsResourceLimitValue,
+    soft_limit: OciDefaultsResourceLimitValue,
```
bcressey: This will require a data migration for the "max open files" limit - on downgrade, we need to turn `OciDefaultsResourceLimitValue` back into a `u32`, and on upgrade we might need to go the other way.

sources/models/src/modeled_types/oci_defaults.rs
```diff
@@ -120,10 +157,158 @@ mod oci_defaults_rlimits {
...
+    }
+}
+
+/// OciDefaultsResourceLimitValue specifies that any u32 integer and "unlimited"
+/// are the valid resource-limits. The hard-limit="unlimited" or
+/// soft-limit="unlimited" sets u64::MAX in spec file for container
+/// that converts to unlimited in containers for that resource
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub enum OciDefaultsResourceLimitValue {
+    Value(u32),
```
bcressey: I am not confident about this - "max file size" seems to take a non-negative i64 value. You should check whether the same range is accepted for other limits.

We have a `NonNegativeInteger` type already (for i32) and could do the same for a `NonNegativeI64` type if needed.

If it turns out that `NonNegativeI64` is the right range for incoming values, then I would prefer to store everything as an i64, and represent "unlimited" as `-1`.

README.md
```diff
@@ -755,12 +755,195 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-stack-size-limit</code></td>
+  <td>-</td>
+</tr>
+</table>
+
+Any integer between 0 to 4294967295 (2 ^32^ -1) and `unlimited` are valid values to set hard-limit and soft limit for a resource.
```
arnaldo2792: nit: missing `-` in "soft limit"

README.md
```diff
@@ -755,12 +755,195 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+</table>
+
+Any integer between 0 to 4294967295 (2 ^32^ -1) and `unlimited` are valid values to set hard-limit and soft limit for a resource. 
+The string `unlimited` as resource-limit denotes limit as unlimited. Following are the examples to set hard-limit and soft-limit:
+* Setting any valid integer value as limit:
+  ```toml
+  [settings.oci-defaults.resource-limits.<rlimit>>]
+  hard-limit = 65536
+  soft-limit = 65536
+    ```
```
arnaldo2792: nit: There is an extra space here, and Github paints all of this red.

vyaghras:
Allow range of 0- i64::Max as limit and -1 and "unlimited" string to mark any limit as unlimited.

sources/api/migration/migrations/v1.15.0/oci-defaults-max-open-files/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+        match v {
+            -1 => u32::MAX,
+            v if v > u32::MAX as i64 => u32::MAX,
+            _ => v as u32,
+        }
+    } else {
+        return;
+    };
+    *value = Value::Number(s.into());
+}
```
yeazelm: This doesn't cover all cases. You need to either cover all cases that a `Value` can be, or limit the type to the ones you plan to cover (in this case `i64`). I think we are pretty certain this can only be a few things (mostly `i64` or `u32` since we are storing this in our datastore) so at least you could ensure its a `Number` but then you should also cover the case where could be a floating point. This is a little nitpicky but we should endeavor to use the type system when we can to avoid strange issues.

sources/api/migration/migrations/v1.15.0/oci-defaults-max-open-files/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl Migration for ChangeMaxOpenFileResourceLimitType {
+    /// On upgrade there is nothing to do (see above).
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        Ok(input)
```
yeazelm, line 32: If we did an end to end migration that was set as `unlimited`, would we end up with `unlimited` again or `u32::MAX`? This would be pretty surprising if its not getting us back to `unlimited`.

sources/models/src/de.rs
```diff
@@ -207,3 +207,126 @@ mod mirrors_tests {
         );
     }
 }
+
+// This specifies that any non negative i64 integer, -1 and "unlimited"
+/// are the valid resource-limits. The hard-limit="unlimited" or -1 and
+/// soft-limit="unlimited" or -1 sets u64::MAX in spec file for container
+/// that converts to unlimited in containers for that resource
```
yeazelm:
```suggestion
// This specifies that any non negative i64 integer, -1, and "unlimited"
/// are the valid resource-limits. The hard-limit set to "unlimited" or -1 
/// and soft-limit set to "unlimited" or -1 are converted to u64::MAX in 
/// the spec file for the container runtime which ultimately represents 
/// unlimited for that resource
```

The last sentence was a little hard to understand, does this still mean the same thing or did I mis-understand what you are trying to say?

README.md
```diff
@@ -756,12 +756,201 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+  <td>-</td>
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-stack-size.hard-limit</code></td>
+  <td>-</td>
+</tr>
+</table>
+
+Any integer between 0 to 9223372036854775807 (2 ^63^ -1), -1 and `unlimited` are valid values to set hard-limit and soft limit for a resource.
```
```suggestion
Limits can be any integer between 0 to `int64::MAX`. Either `-1` or `"unlimited"` can be used to remove the limit.
```

README.md
```diff
@@ -756,12 +756,201 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-stack-size.hard-limit</code></td>
+  <td>-</td>
+</tr>
+</table>
+
+Any integer between 0 to 9223372036854775807 (2 ^63^ -1), -1 and `unlimited` are valid values to set hard-limit and soft limit for a resource. 
+The string `unlimited` or -1 as resource-limit denotes limit as unlimited. Following are the examples to set hard-limit and soft-limit:
```
```suggestion
```

README.md
```diff
@@ -756,12 +756,201 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+* Setting any valid integer value as limit:
+  ```toml
+  [settings.oci-defaults.resource-limits.<rlimit>>]
+  hard-limit = 65536
+  soft-limit = 65536
+  ```
```
bcressey: This is implied by the next example.
```suggestion
```

README.md
```diff
@@ -756,12 +756,201 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+* Setting max possible integer value as limit:
+  ```toml
+  [settings.oci-defaults.resource-limits.<rlimit>>]
+  hard-limit = 9223372036854775807
+  soft-limit = 65536
+  ```
```
```suggestion
* Specifying the maximum value (`u64::MAX`) for a limit:
  ```toml
  [settings.oci-defaults.resource-limits.<rlimit>]
  soft-limit = 65536
  hard-limit = 9223372036854775807
  ```
```

README.md
```diff
@@ -756,12 +756,201 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+* Setting unlimited value as limit:
+  ```toml
+  [settings.oci-defaults.resource-limits.<rlimit>>]
+  hard-limit = "unlimited"
+  soft-limit = 65536
```
```suggestion
* Removing a limit:
  ```toml
  [settings.oci-defaults.resource-limits.<rlimit>]
  soft-limit = 65536
  hard-limit = "unlimited"
```

README.md
```diff
@@ -756,12 +756,201 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+  ```
+* Setting unlimited value as limit:
+  ```toml
+  [settings.oci-defaults.resource-limits.<rlimit>>]
+  hard-limit = -1
+  soft-limit = 65536
+  ```
```
bcressey: We can support this in code, but don't need to document it. The "unlimited" one is more obvious and should be preferred.

sources/api/migration/migrations/v1.15.0/oci-defaults-max-open-files/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+const HARD_RESOURCE_LIMIT_SETTING_NAME: &str =
+    "settings.oci-defaults.resource-limits.max-open-files.hard-limit";
+const SOFT_RESOURCE_LIMIT_SETTING_NAME: &str =
+    "settings.oci-defaults.resource-limits.max-open-files.soft-limit";
+
+/// This migration changes the hard and soft limit for rlimit_nofile to u32 from i64 on downgrade.
+/// There is no need of migration on upgrade as u32 will automatically change to i64
+pub struct ChangeMaxOpenFileResourceLimitType;
+
+fn convert_to_u_32(value: &mut Value) {
```
bcressey: nit:
```suggestion
fn convert_to_u32(value: &mut Value) {
```

sources/models/src/de.rs
```diff
@@ -207,3 +207,126 @@ mod mirrors_tests {
...
+        Int(i64),
+    }
+
+    match StringOrInt64::deserialize(deserializer)? {
+        StringOrInt64::String(s) => {
+            if s == "unlimited" {
+                Ok(-1)
+            } else {
+                Err(Error::custom(format!(
+                    "Invalid rlimit {}, use integer within the range of -1 - {} or string \" unlimited \" to set rlimit unlimited",
```
bcressey:
```suggestion
                    "Invalid rlimit {}, expected -1 to {} or \"unlimited\"",
```

sources/models/src/de.rs
```diff
@@ -207,3 +207,126 @@ mod mirrors_tests {
...
+                    i64::MAX
+                )))
+            }
+        }
+        StringOrInt64::Int(i) => {
+            if (-1..=i64::MAX).contains(&i) {
+                Ok(i)
+            } else {
+                Err(Error::custom(format!(
+                    "Provided number {} is not within the range of -1 - {}",
```
bcressey: This can use the same error as above.

sources/models/src/lib.rs
```diff
@@ -530,9 +530,28 @@ struct OciDefaults {
...
+    soft_limit: i64,
+}
+
+impl OciDefaultsResourceLimit {
+    pub fn get_hard_limit(self) -> u64 {
+        match self.hard_limit {
+            -1 => u64::MAX,
+            _ => self.hard_limit as u64,
+        }
+    }
```
bcressey: nit: the body could be factored into a private `get_limit` function that takes either `self.hard_limit` or `self.soft_limit` as a parameter

bcressey (commented):
Very nice. Some of the documentation and code can be tightened up to convey the essential information and avoid repetition, but the logic looks solid.

sources/api/migration/migrations/v1.15.0/oci-defaults-max-open-files/src/main.rs
bcressey: It should suffice to handle `value.is_i64()` and otherwise just leave the value alone, right?
```suggestion
fn convert_to_u_32(value: &mut Value) {
    if !value.is_i64() {
        return;
    }
    let v: i64 = serde_json::from_value(value.clone()).unwrap();
    match v {
        -1 => u32::MAX,
        v if v > u32::MAX as i64 => u32::MAX,
        _ => v as u32,
    }
    *value = Value::Number(s.into());
}
```

sources/api/migration/migrations/v1.15.0/oci-defaults-max-open-files/src/main.rs
bcressey: It feels like `u32::MAX` is the best we can do in that case - earlier versions didn't support anything beyond that, or "unlimited". Removing the limit entirely might cause it to be set much lower than expected, and `u32::MAX` open files is still quite a few.

yeazelm: Yeah, I think I'm ok with `u32::MAX` over dropping it. I was considering if we should take `u32::MAX` on the forward case as `unlimited` but that might be less desirable than leaving it as we find it. Either way, I think the delta between `u32::MAX` and `unlimited` actually mattering in the real world is pretty low likelihood.

sources/api/migration/migrations/v1.15.0/oci-defaults-max-open-files/src/main.rs
yeazelm: I agree with @bcressey's suggestion.

sources/models/src/de.rs
vyaghras: Yeah it means same, will change to the one that you suggested.

README.md
```diff
@@ -756,12 +756,189 @@ Each of the `resource-limits` settings below contain two numeric fields: `hard-l
...
+  <td rowspan="2">bytes</td>
+</tr>
+<tr>
+  <td><code>settings.oci-defaults.resource-limits.max-stack-size.hard-limit</code></td>
+  <td>-</td>
+</tr>
+</table>
+
+Limits can be any integer between 0 to `int64::MAX`. Either `-1` or `"unlimited"` can be used to remove the limit.
+* Specifying the maximum value (`u64::MAX`) for a limit:
```
```suggestion
* Specifying the maximum value (`i64::MAX`) for a limit:
```

sources/api/migration/migrations/v1.15.0/oci-defaults-max-open-files/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+impl Migration for ChangeMaxOpenFileResourceLimitType {
+    /// On upgrade there is nothing to do (see above).
+    fn forward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        Ok(input)
+    }
+
+    /// On downgrade, if the value is an i64 integer, we need to convert it to a u32.
+    ///
+    /// Note that this potentially causes data loss, if current value of the setting
+    /// is -1 or higher than u_32::MAX we  will set it to max possible value i.e. u32::MAX.
```
bcressey:
```suggestion
    /// is -1 or higher than u_32::MAX we will set it to max possible value i.e. u32::MAX.
```

sources/models/src/lib.rs
```diff
@@ -530,9 +530,29 @@ struct OciDefaults {
...
+    soft_limit: i64,
+}
+
+impl OciDefaultsResourceLimit {
+    fn get_limit(limit: i64) -> u64 {
+        match limit {
+            -1 => u64::MAX,
+            _ => limit as u64,
+        }
+    }
```
bcressey: nit: I forget where I picked up this convention, maybe Java, but I prefer to put public functions first since they're the external interface for the type, and private functions later since they're implementation details.

bcressey (approved):
LGTM with a few nits.

yeazelm (approved):

vyaghras:
I have tested this change again with all the migrations. The change is working as expected.

PR #3213, opened 2023-06-20
update SDK to 0.33.0

bcressey:
**Issue number:**
N/A

**Description of changes:**
Switch to the new SDK, which includes:
* Rust 1.70.0
* Go 1.20.5
* `xfsprogs`
* re-homed RPM macros

The macro move requires some minor fixes. The most visible change will be that RPMs now have the architecture set correctly to the target's architecture, rather than including that in the package name. RPMs aren't directly user-facing except in the case of `application-inventory.json`, which is now simpler to generate.

I've also cleaned up packages that were (incorrectly) tagged as "noarch". This changes `application-inventory.json` for the affected packages, which move from `"Architecture": "noarch"` to the target's architecture.

**Testing done:**
Built some variants:
* aws-k8s-1.24 on x86_64 host for x86_64 target
* aws-k8s-1.24 on aarch64 host for aarch64 target
* aws-k8s-1.27 on x86_64 host for aarch64 target
* aws-k8s-1.27 on aarch64 host for x86_64 target

I checked `application-inventory.json` on the running instances and confirmed that the prefix was correctly matched and removed, and that the "Architecture" field was set to the target's architecture for all packages.

**Terms of contribution:**


stmcginnis:
Please revert ee7205d92c7f8aaeb4ea8d66105f158c9731fc6c in this PR.

We had to pin the node's rust version temporarily, otherwise the cargo index format was getting a mismatch between the 1.69 format and the newer 1.70 sparse format. Once we move to 1.70 in the SDK, we can remove that pin so the host gets the `latest` version, which will be 1.70 or later that supports the right format.

etungsten (approved):

stmcginnis (approved):
Looks good to me.

PR #3214, opened 2023-06-20
pluto: add more retries for requests out to AWS

etungsten:
**Issue number:**

N/A

**Description of changes:**
```
    pluto: increase retries for requests out to AWS
    
    This adds more resiliency when pluto makes API requests out to EC2 and
    EKS. We choose a large number of retries to force the SDK to keep
    retrying until we timeout on the overall request.
    
    It's now less likely for the request to overall fail due to
    transient connection errors.

```

```
    sundog: enforce settings generator execution timeout
    
    Enforces a 6 minute timeout on settings generator execution so as to not
    hang the boot for too long before erroring out.

```

**Testing done:**
On a host where 25% of all outbound packets get dropped, I was able to call `pluto private-dns-name` where it retried the request 5 times (beyond the default 3 tries set by the SDK).

```
18:54:36 [DEBUG] (4) hyper::client::connect::dns: resolving host="ec2.us-west-2.amazonaws.com"
18:54:39 [DEBUG] (1) aws_smithy_client::retry: attempt 1 failed with Error(TransientError); retrying after 93.530282ms
18:54:39 [DEBUG] (1) aws_smithy_client::retry: retry; kind=Error(TransientError)
18:54:39 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="resolve_endpoint"
18:54:39 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="resolve_endpoint"
18:54:39 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="generate_user_agent"
18:54:39 [DEBUG] (1) aws_smithy_http_tower::map_request: async_map_request; name="retrieve_credentials"
18:54:39 [DEBUG] (1) aws_credential_types::cache::lazy_caching: loaded credentials from cache
18:54:39 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="sigv4_sign_request"
18:54:39 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="recursion_detection"
18:54:39 [DEBUG] (1) tracing::span: dispatch;
18:54:39 [DEBUG] (5) hyper::client::connect::dns: resolving host="ec2.us-west-2.amazonaws.com"
18:54:42 [DEBUG] (1) aws_smithy_client::retry: attempt 2 failed with Error(TransientError); retrying after 381.207122ms
18:54:42 [DEBUG] (1) aws_smithy_client::retry: retry; kind=Error(TransientError)
18:54:43 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="resolve_endpoint"
18:54:43 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="resolve_endpoint"
18:54:43 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="generate_user_agent"
18:54:43 [DEBUG] (1) aws_smithy_http_tower::map_request: async_map_request; name="retrieve_credentials"
18:54:43 [DEBUG] (1) aws_credential_types::cache::lazy_caching: loaded credentials from cache
18:54:43 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="sigv4_sign_request"
18:54:43 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="recursion_detection"
18:54:43 [DEBUG] (1) tracing::span: dispatch;
18:54:43 [DEBUG] (6) hyper::client::connect::dns: resolving host="ec2.us-west-2.amazonaws.com"
18:54:46 [DEBUG] (1) aws_smithy_client::retry: attempt 3 failed with Error(TransientError); retrying after 1.246607675s
18:54:46 [DEBUG] (1) aws_smithy_client::retry: retry; kind=Error(TransientError)
18:54:47 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="resolve_endpoint"
18:54:47 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="resolve_endpoint"
18:54:47 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="generate_user_agent"
18:54:47 [DEBUG] (1) aws_smithy_http_tower::map_request: async_map_request; name="retrieve_credentials"
18:54:47 [DEBUG] (1) aws_credential_types::cache::lazy_caching: loaded credentials from cache
18:54:47 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="sigv4_sign_request"
18:54:47 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="recursion_detection"
18:54:47 [DEBUG] (1) tracing::span: dispatch;
18:54:47 [DEBUG] (5) hyper::client::connect::dns: resolving host="ec2.us-west-2.amazonaws.com"
18:54:50 [DEBUG] (1) aws_smithy_client::retry: attempt 4 failed with Error(TransientError); retrying after 1.733219105s
18:54:50 [DEBUG] (1) aws_smithy_client::retry: retry; kind=Error(TransientError)
18:54:52 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="resolve_endpoint"
18:54:52 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="resolve_endpoint"
18:54:52 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="generate_user_agent"
18:54:52 [DEBUG] (1) aws_smithy_http_tower::map_request: async_map_request; name="retrieve_credentials"
18:54:52 [DEBUG] (1) aws_credential_types::cache::lazy_caching: loaded credentials from cache
18:54:52 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="sigv4_sign_request"
18:54:52 [DEBUG] (1) aws_smithy_http_tower::map_request: map_request; name="recursion_detection"
18:54:52 [DEBUG] (1) tracing::span: dispatch;
18:54:52 [DEBUG] (4) hyper::client::connect::dns: resolving host="ec2.us-west-2.amazonaws.com"
18:54:52 [DEBUG] (1) hyper::client::connect::http: connecting to 52.94.214.26:443
18:54:52 [DEBUG] (1) hyper::client::connect::http: connected to 52.94.214.26:443
18:54:52 [DEBUG] (1) rustls::client::hs: No cached session for DnsName(DnsName(DnsName("ec2.us-west-2.amazonaws.com")))
18:54:52 [DEBUG] (1) rustls::client::hs: Not resuming any session
18:54:52 [DEBUG] (1) rustls::client::hs: ALPN protocol is Some(b"http/1.1")
18:54:52 [DEBUG] (1) rustls::client::hs: Using ciphersuite TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
18:54:52 [DEBUG] (1) rustls::client::tls12: ECDHE curve is ECParameters { curve_type: NamedCurve, named_group: secp256r1 }
18:54:52 [DEBUG] (1) rustls::client::tls12: Server DNS name is DnsName(DnsName(DnsName("ec2.us-west-2.amazonaws.com")))
18:54:52 [DEBUG] (1) rustls::client::tls12: Session saved
18:54:52 [DEBUG] (2) hyper::proto::h1::io: flushed 1994 bytes
18:54:52 [DEBUG] (3) hyper::proto::h1::io: parsed 8 headers
18:54:52 [DEBUG] (3) hyper::proto::h1::conn: incoming body is chunked encoding
18:54:52 [DEBUG] (1) tracing::span: load_response;
18:54:52 [DEBUG] (1) tracing::span: parse_unloaded;
18:54:52 [DEBUG] (1) tracing::span: read_body;
18:54:52 [DEBUG] (3) hyper::proto::h1::decode: incoming chunked header: 0x2000 (8192 bytes)
18:54:52 [DEBUG] (3) hyper::proto::h1::decode: incoming chunked header: 0x10EB (4331 bytes)
18:54:52 [DEBUG] (2) hyper::proto::h1::conn: incoming body completed
18:54:52 [DEBUG] (2) hyper::client::pool: pooling idle connection for ("https", ec2.us-west-2.amazonaws.com)
18:54:52 [DEBUG] (1) tracing::span: parse_loaded;
18:54:52 [DEBUG] (1) aws_smithy_client: send_operation; status="ok"
"i-abcd1234.us-west-2.compute.internal"
18:54:52 [DEBUG] (3) rustls::conn: Sending warning alert CloseNotify

real	0m17.869s
user	0m0.022s
sys	0m0.000s
```

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/aws.rs
+++ b/sources/api/pluto/src/aws.rs
@@ -1,0 +1,47 @@
+use aws_config::default_provider::credentials::DefaultCredentialsChain;
+use aws_config::imds;
+use aws_smithy_types::retry::{RetryConfig, RetryConfigBuilder};
+use aws_types::region::Region;
+use aws_types::SdkConfig;
+use snafu::{ResultExt, Snafu};
+use std::time::Duration;
+
+// Max request retry attempts; Retry many many times and let the caller decide when to terminate
+const MAX_ATTEMPTS: u32 = 100;
+const IMDS_CONNECT_TIMEOUT: Duration = Duration::from_secs(3);
+
+#[derive(Debug, Snafu)]
+pub(super) enum Error {
+    #[snafu(display("Failed to build IMDS client: {}", source))]
+    SdkImds {
+        source: imds::client::error::BuildError,
+    },
+}
+
+type Result<T> = std::result::Result<T, Error>;
+
+async fn sdk_imds_client() -> Result<imds::Client> {
+    imds::Client::builder()
+        .max_attempts(MAX_ATTEMPTS)
+        .connect_timeout(IMDS_CONNECT_TIMEOUT)
+        .build()
+        .await
+        .context(SdkImdsSnafu)
+}
+
+fn sdk_retry_config() -> RetryConfig {
+    RetryConfigBuilder::new().max_attempts(MAX_ATTEMPTS).build()
+}
+
+pub(crate) async fn sdk_config(region: &str) -> Result<SdkConfig> {
+    let provider = DefaultCredentialsChain::builder()
+        .imds_client(sdk_imds_client().await?)
+        .build()
+        .await;
+    Ok(aws_config::from_env()
+        .region(Region::new(region.to_owned()))
+        .credentials_provider(provider)
+        .retry_config(sdk_retry_config())
+        .load()
+        .await)
+}
--- a/sources/api/pluto/src/ec2.rs
+++ b/sources/api/pluto/src/ec2.rs
@@ -1,6 +1,6 @@
-use crate::proxy;
+use crate::aws::sdk_config;
+use crate::{aws, proxy};
 use aws_smithy_types::error::display::DisplayErrorContext;
-use aws_types::region::Region;
 use snafu::{OptionExt, ResultExt, Snafu};
 use std::time::Duration;
 
@@ -27,6 +27,9 @@ pub(super) enum Error {
 
     #[snafu(context(false), display("{}", source))]
     Proxy { source: proxy::Error },
+
+    #[snafu(context(false), display("{}", source))]
+    SdkConfig { source: aws::Error },
 }
 
 type Result<T> = std::result::Result<T, Error>;
@@ -35,10 +38,7 @@ pub(super) async fn get_private_dns_name(region: &str, instance_id: &str) -> Res
     // Respect proxy environment variables when making AWS EC2 API requests
     let (https_proxy, no_proxy) = proxy::fetch_proxy_env();
 
-    let config = aws_config::from_env()
-        .region(Region::new(region.to_owned()))
-        .load()
-        .await;
+    let config = sdk_config(region).await?;
 
     let client = if let Some(https_proxy) = https_proxy {
         let http_client = proxy::setup_http_client(https_proxy, no_proxy)?;
--- a/sources/api/pluto/src/eks.rs
+++ b/sources/api/pluto/src/eks.rs
@@ -1,6 +1,6 @@
-use crate::proxy;
+use crate::aws::sdk_config;
+use crate::{aws, proxy};
 use aws_sdk_eks::model::KubernetesNetworkConfigResponse;
-use aws_types::region::Region;
 use snafu::{OptionExt, ResultExt, Snafu};
 use std::time::Duration;
 
@@ -19,11 +19,14 @@ pub(super) enum Error {
     #[snafu(display("Timed-out waiting for EKS Describe Cluster API response: {}", source))]
     DescribeClusterTimeout { source: tokio::time::error::Elapsed },
 
-    #[snafu(display("Missing field '{}' EKS response", field))]
+    #[snafu(display("Missing field '{}' in EKS response", field))]
     Missing { field: &'static str },
 
     #[snafu(context(false), display("{}", source))]
     Proxy { source: proxy::Error },
+
+    #[snafu(context(false), display("{}", source))]
+    SdkConfig { source: aws::Error },
 }
 
 type Result<T> = std::result::Result<T, Error>;
@@ -37,10 +40,7 @@ pub(super) async fn get_cluster_network_config(
     // Respect proxy environment variables when making AWS EKS API requests
     let (https_proxy, no_proxy) = proxy::fetch_proxy_env();
 
-    let config = aws_config::from_env()
-        .region(Region::new(region.to_owned()))
-        .load()
-        .await;
+    let config = sdk_config(region).await?;
 
     let client = if let Some(https_proxy) = https_proxy {
         let http_client = proxy::setup_http_client(https_proxy, no_proxy)?;
--- a/sources/api/pluto/src/main.rs
+++ b/sources/api/pluto/src/main.rs
@@ -32,6 +32,7 @@ reasonable default is available.
 */
 
 mod api;
+mod aws;
 mod ec2;
 mod eks;
 mod proxy;
--- a/sources/api/sundog/Cargo.toml
+++ b/sources/api/sundog/Cargo.toml
@@ -20,7 +20,7 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 simplelog = "0.12"
 snafu = "0.7"
-tokio = { version = "~1.25", default-features = false, features = ["macros", "rt-multi-thread"] }  # LTS
+tokio = { version = "~1.25", default-features = false, features = ["process", "macros", "rt-multi-thread"] }  # LTS
 
 [build-dependencies]
 generate-readme = { version = "0.1", path = "../../generate-readme" }
--- a/sources/api/sundog/src/main.rs
+++ b/sources/api/sundog/src/main.rs
@@ -17,9 +17,14 @@ use std::env;
 use std::path::Path;
 use std::process;
 use std::str::{self, FromStr};
+use std::time::Duration;
+use tokio::process::Command as AsyncCommand;
 
 use datastore::serialization::to_pairs_with_prefix;
 use datastore::{self, deserialization, Key, KeyType};
+
+// Limit settings generator execution to at most 6 minutes to prevent boot from hanging for too long.
+const SETTINGS_GENERATOR_TIMEOUT: Duration = Duration::from_secs(360);
 
 /// Potential errors during Sundog execution
 mod error {
@@ -36,6 +41,16 @@ mod error {
         CommandFailure {
             program: String,
             source: std::io::Error,
+        },
+
+        #[snafu(display(
+            "Timed-out waiting for settings generator '{}' to finish: {}",
+            generator,
+            source
+        ))]
+        CommandTimeout {
+            generator: String,
+            source: tokio::time::error::Elapsed,
         },
 
         #[snafu(display("Generator command is invalid (empty, etc.) - '{}'", command))]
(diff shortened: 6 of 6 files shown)


webern (approved):
Looks reasonable to me. Is this the same timeout that we were getting by default before?

```
const IMDS_CONNECT_TIMEOUT: Duration = Duration::from_secs(2);
```

etungsten:
> Is this the same timeout that we were getting by default before?

The default timeout for the IMDS clients are set to 1 second. https://github.com/awslabs/smithy-rs/blob/7ed51b21290aba818fcfd7a5501fe7035cde5c24/aws/rust-runtime/aws-config/src/imds/client.rs#L47

stmcginnis (approved):

zmrow (approved):

sources/api/pluto/src/aws.rs
```diff
@@ -0,0 +1,47 @@
...
+use aws_config::imds;
+use aws_smithy_types::retry::{RetryConfig, RetryConfigBuilder};
+use aws_types::region::Region;
+use aws_types::SdkConfig;
+use snafu::{ResultExt, Snafu};
+use std::time::Duration;
+
+// Max request retry attempts
+const MAX_ATTEMPTS: u32 = 10;
+const IMDS_CONNECT_TIMEOUT: Duration = Duration::from_secs(2);
```
bcressey: Can we make this 3 seconds to align with the AWS SDK?

sources/api/pluto/src/aws.rs
```diff
@@ -0,0 +1,47 @@
...
+        .max_attempts(MAX_ATTEMPTS)
+        .connect_timeout(IMDS_CONNECT_TIMEOUT)
+        .build()
+        .await
+        .context(SdkImdsSnafu)
+}
+
+pub(crate) fn sdk_retry_config() -> RetryConfig {
+    RetryConfigBuilder::new().max_attempts(MAX_ATTEMPTS).build()
+}
```
bcressey: Do these need to be `pub(crate)`?
```suggestion
async fn sdk_imds_client() -> Result<imds::Client> {
    imds::Client::builder()
        .max_attempts(MAX_ATTEMPTS)
        .connect_timeout(IMDS_CONNECT_TIMEOUT)
        .build()
        .await
        .context(SdkImdsSnafu)
}

fn sdk_retry_config() -> RetryConfig {
    RetryConfigBuilder::new().max_attempts(MAX_ATTEMPTS).build()
}
```

sources/api/pluto/src/aws.rs
```diff
@@ -0,0 +1,47 @@
+use aws_config::default_provider::credentials::DefaultCredentialsChain;
+use aws_config::imds;
+use aws_smithy_types::retry::{RetryConfig, RetryConfigBuilder};
+use aws_types::region::Region;
+use aws_types::SdkConfig;
+use snafu::{ResultExt, Snafu};
+use std::time::Duration;
+
+// Max request retry attempts
+const MAX_ATTEMPTS: u32 = 10;
```
bcressey: It looks like the random jitter applied to the backoff could cause all requests to be retried very quickly.

I'd like a stronger property, like "we'll retry for up to 3 minutes", without also getting into really long delays (tens of seconds), since then we might spend the last minute or two waiting for a retry that will never happen.

bcressey: It might be better to handle that stronger property - retrying for many minutes - as an overall timeout on the `pluto` execution, so we don't end up stacking lengthy timeouts together for each request that happens within an execution of the settings handle.

So maybe for the SDK client it's more of a "retry forever" strategy but without letting the backoff grow to absurd numbers.

cbgbt: It is kind of unfortunate that you can't set a maximum backoff in this SDK.

To achieve this property are you suggesting we write our own retry loop in pluto which combines with the internal retries to create the behavior we want?

etungsten: > as an overall timeout on the pluto execution, so we don't end up stacking lengthy timeouts together for each request that happens within an execution of the settings handle.

Double checking my understanding; Are we saying to wrap each `sundog` settings generator execution in a `tokio::time::timeout`?

> for the SDK client it's more of a "retry forever" strategy but without letting the backoff grow to absurd numbers.

Max backoff seems to cap out at 20 seconds: https://github.com/awslabs/aws-sdk-rust/blob/v0.15.0/sdk/aws-smithy-client/src/retry.rs#L95

Unfortunately there's no public interface that lets me customize that value: https://github.com/awslabs/aws-sdk-rust/blob/66423e05991ee831696bc32fe3e452694cf0d231/sdk/aws-smithy-client/src/retry.rs#L50-L51

However, we can change the multiplier for the exponential backoff. The default is 1 second and we can lower that if necessary: https://docs.rs/aws-sdk-ec2/latest/aws_sdk_ec2/config/retry/struct.RetryConfig.html#method.with_initial_backoff

In the worst case, the jitter can still cause each subsequent retry to happen almost immediately. So it's awkward to pick out an appropriate `MAX_ATTEMPTS` here. Assuming there's no backoff between retries. 3 minutes of retries would roughly equate to 58 retries with the default 3.1 second connection timeout.

So to summarize, I'll wrap each settings generator call in a 3 minute enforced tokio timeout block. Then I'll increase the max retry attempts for the SDK to ~60 (or higher) to accommodate the "worst" case scenario where backoffs are nonexistent due to jitters. To ensure the backoff doesn't grow to a ridiculous number, we scale the multiplier down to something like 100ms, so the backoff will essentially never grow beyond 100ms * 60 retries =  6 seconds. Does that sounds right?

bcressey: > So to summarize, I'll wrap each settings generator call in a 3 minute enforced tokio timeout block.

Sounds good. Maybe a 5 minute timeout for consistency with the IMDS client timeout.

> Then I'll increase the max retry attempts for the SDK to ~60 (or higher) to accommodate the "worst" case scenario where backoffs are nonexistent due to jitters. To ensure the backoff doesn't grow to a ridiculous number, we scale the multiplier down to something like 100ms, so the backoff will essentially never grow beyond 100ms * 60 retries = 6 seconds. Does that sounds right?

If the backoff is capped at 20 seconds, that might be OK too. Then we can just retry indefinitely within the time allotted by `sundog`.

etungsten: One problem I see with the retry-indefinitely strategy is for when we call `pluto cluster-dns-ip`. If the EKS describe cluster call fails for whatever reason, `pluto` defaults to a fallback value based on IMDS. In this new model,  we would never fallback and hard fail the boot, whereas before, if `pluto` couldn't reach the EKS endpoint, it would at least be able to set a value after the 5 min timeout in pluto.

etungsten:
Pushes above and below addresses @bcressey 's comment.

- Change IMDS connection timeout from 2 seconds to 3 seconds.
- Remove extraneous crate visibility specification for functions that don't need it.
- In pluto, increase max retry attempts to a large number of retries and let sundog terminate the process if it takes too long.
- Enforce settings generator execution timeout in sundog so as to not let settings generator hang the boot indefinitely.

stmcginnis (approved):
Updates look good to me.

sources/api/pluto/src/aws.rs
etungsten: I opted to limit the settings generator execution to 6 mins in sundog and I kept the overall API request timeout in pluto to 5 minutes so as to give pluto a chance to use fallback values if possible.

bcressey (approved):

PR #3219, opened 2023-06-21
apiserver: Add /report endpoint for CIS reporting

stmcginnis:
**Issue number:**

Related #2731 

**Description of changes:**

This adds a new `/report` endpoint. GETs to this endpoint get a list of the available reports which can then be called directly. Initially this is only `/report/cis` that can be used to trigger a bloodhound CIS benchmark report, returning the results. The type argument controls which report to run. Right now there is only the CIS report, but this will be expanded in the future as we add other compliance or benchmark reports.

```
# apiclient -u /report
[{"name":"cis","description":"CIS Bottlerocket Benchmark"}]
```

The bloodhound options for `--level` and `--format` are supported with:

```
# apiclient -u "/report/cis&level=2&format=json"
```

**Testing done:**

```
# apiclient -u /report
[{"name":"cis","description":"CIS Bottlerocket Benchmark"}]

# apiclient -u /report/cis
Failed GET request to '/report': Status 400 when GETing /report: Report type must be specified

# apiclient -u "/report?type=cis-k8s"
Failed GET request to '/report?type=cis-k8s': Status 422 when GETing /report?type=cis-k8s: Report type 'cis-k8s' is not supported

# apiclient raw -u "/report?type=cis&level=2&format=json"

{...}
```

**Terms of contribution:**

### DIFF
--- a/sources/api/apiserver/src/server/mod.rs
+++ b/sources/api/apiserver/src/server/mod.rs
@@ -15,7 +15,7 @@ use error::Result;
 use fs2::FileExt;
 use http::StatusCode;
 use log::info;
-use model::{ConfigurationFiles, Model, Services, Settings};
+use model::{ConfigurationFiles, Model, Report, Services, Settings};
 use nix::unistd::{chown, Gid};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::{HashMap, HashSet};
@@ -27,6 +27,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::sync;
 use thar_be_updates::status::{UpdateStatus, UPDATE_LOCKFILE};
+use tokio::process::Command as AsyncCommand;
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 
@@ -121,6 +122,11 @@ where
             )
             .service(web::scope("/updates").route("/status", web::get().to(get_update_status)))
             .service(web::resource("/exec").route(web::get().to(exec::ws_exec)))
+            .service(
+                web::scope("/report")
+                    .route("", web::get().to(list_reports))
+                    .route("/cis", web::get().to(get_cis_report)),
+            )
     })
     .workers(threads)
     .bind_uds(socket_path.as_ref())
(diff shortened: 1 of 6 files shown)


sources/api/apiserver/src/server/mod.rs
```diff
@@ -543,6 +545,49 @@ async fn reboot() -> Result<HttpResponse> {
     Ok(HttpResponse::NoContent().finish())
 }
 
+/// Triggers report generation for the requested report type and returns the results.
+async fn get_report(query: web::Query<HashMap<String, String>>) -> Result<HttpResponse> {
+    trace!("Generating report");
+
+    match query.get("type") {
+        Some(report_type) => match report_type.as_str() {
+            "cis" => get_cis_report(&query).await,
```
etungsten: How long does `bloodhound` take to generate the CIS report? We want to minimize the time `apiserver` blocks on any API request.

stmcginnis: For the most part, the report generation is just checking the content of a handful of files. So the CIS report is actually pretty quick:

```
real    0m0.079s
user    0m0.043s
sys     0m0.008s
```

We may eventually want to add other types of reports. These will all be exposed through `apiclient` under a common `report` subcommand. If/when we add a type of report that would take more time, the plan is `apiclient` logic will abstract away the details and call an async endpoint under the hood.

So with the current command, `apiclient report cis` would end up calling `GET /report&type=cis`.

If we add something more resource intesive, we would have something like `apiclient report io-sampling` that would end up routing things through something like `POST /actions/report body: {"type": "io-sample"}`, then poll for the results from something like `GET /actions/report-result body {"id": $async_id}`.

Specific details to be figured out if/when we go there, of course. But for now, this is so quick that I figured adding async request/result handling would just add extra overhead on both the client and server ends.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -121,6 +122,7 @@ where
             )
             .service(web::scope("/updates").route("/status", web::get().to(get_update_status)))
             .service(web::resource("/exec").route(web::get().to(exec::ws_exec)))
+            .service(web::scope("/report").route("", web::get().to(get_report)))
```
yeazelm: I had expected to see a `/report/cis?level=2&format=json` or `/report/cis/2` instead of `/report?type=cis&level=2&format=json` and allow different types of reports to be routes within this API.

It's more complex, but then this would allow `/report/list` to list the available reports. I'm not sure if I'm just mistaken on how we have typically designed these APIs in the past to keep it consistent though.

stmcginnis: We could certainly do it that way. The only main downside I see is having a little more duplication of boilerplate code to route to the right processing (maybe?) versus the current argument approach. I could see a `/list` endpoint being useful for the `apiclient` discovery. I'll take a look at changing things.

stmcginnis: One small change - rather than `/report/list` to get the available reports, it's more "restful" to have the GET on `/report` to get its options, then `/report/cis` to get the specific report.

etungsten (approved):

sources/api/apiserver/src/server/mod.rs
```diff
@@ -543,6 +549,46 @@ async fn reboot() -> Result<HttpResponse> {
...
+/// Gets the Bottlerocket CIS benchmark report.
+async fn get_cis_report(query: web::Query<HashMap<String, String>>) -> Result<HttpResponse> {
+    let mut cmd = AsyncCommand::new("/usr/bin/bloodhound");
+
+    // Check for requested level, default is 1
+    if let Some(level) = query.get("level") {
+        cmd.arg("-l").arg(level);
+    }
+
+    // Check for requested format, default is text
```
cbgbt, line 571: Does it make sense for us to specify the defaults to the underlying tool here? That way behavior changes to the tool's defaults don't impact the apiserver.

cbgbt (approved):

sources/api/apiserver/src/server/mod.rs
stmcginnis: I don't see ever changing defaults to `bloodhound`, so I tried to avoid sending unnecessary data here. Do you think it's important to make those defaults explicit here?

If we do update bloodhound, we can update the `apiserver` at the same time. Or is there a situation where we could end up with an older `apiserver` and newer `bloodhound`?

yeazelm (approved):

sources/api/apiserver/src/server/mod.rs
stmcginnis: I'm assuming since you approved with this, this was more of just a question versus a strong opinion. Since this has the approvals I am going to merge it, but please do open an issue (or just ping me) if you feel strongly that anything should change here. We still have some time to tweak things, especially something like this, so happy to follow up.

bcressey:
LGTM!

PR #3220, opened 2023-06-21
netdog: Add networkd devices and builders for config

zmrow:
**Issue number:**
Related to #2449

**Description of changes:**
This PR is best reviewed in commit order! :)

This is the second in a series of changes to `netdog` that add the ability to generate `systemd-networkd` config files.  This PR adds the structs that represent the `systemd-networkd` devices we currently use; bonds, vlans, and interfaces.  In code, these structs represent the latest version of the underlying config structures.

The bulk of this PR is the addition of builders for the network and netdev structs that represent config files.  Using builders has a few advantages.  First, a builder allows for centralization of the translation logic that is required to get from our network config related structs to the `systemd-networkd` structs.  Second, a builder allows us to limit the creation methods that can be called, making creation of these structs safer for developers.  (We use type parameters under the hood.  Putting them on the builder keeps them from proliferating all over the code, another nice bonus)

Lastly, unit tests were added for the builders.  They become _somewhat_ redundant once the full integration tests are in place, but I wanted to ensure that everything stayed sane through this development process.  (Another nice side effect is we can use the reference files for the integration tests as well.)

So a small diagram of the config generation process and where we are:
```
          Next to be implemented
 -------------------------------------           This PR                            MERGED #3134
           NET CONFIG                   -----------------------------------    --------------------
+---------------+                               NETWORKD DEVICES                 NETWORKD CONFIG
|               |
| InterfaceV1   +-----------------+   +-----------------+      Builder
|               |                 |   |                 +----------------+
+---------------+                 +-> |NetworkdInterface|                |
+---------------+                 +-> |                 |                |     +-----------------+
|               |                 |   |                 |                +---> |                 |
| InterfaceV2   +-----------------+   +-----------------+                      |                 |
|               |                                                 +----------> | .network        |
+---------------+                     +-----------------+ Builder |            |                 |
                                      |                 +---------+            |                 |
+---------------+               +---> | NetworkdBond    |              +-----> +-----------------+
|               |               |     |                 +---------+    |
|               +---------------+     |                 | Builder |    |       +-----------------+
| BondV1        |                     +-----------------+         |    |       |                 |
|               |                                                 +----+-----> |                 |
+---------------+                     +-----------------+              |       | .netdev         |
                                      |                 +--------------+       |                 |
+---------------+                     | NetworkdVlan    |                 +->  |                 |
|               |                +--> |                 |   Builder       |    +-----------------+
|               |                |    |                 +-----------------+
| VlanV1        +----------------+    +-----------------+
|               |
+---------------+
```

**Testing done:**
* All old and new unit tests pass!
* Build an `aws-k8s-1.24` image to ensure nothing bled into existing images

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/networkd/config/mod.rs
+++ b/sources/api/netdog/src/networkd/config/mod.rs
@@ -3,10 +3,23 @@
 mod netdev;
 mod network;
 
+use super::Result;
 use netdev::NetDevConfig;
 use network::NetworkConfig;
+
+const NETWORKD_CONFIG_DIR: &str = "/etc/systemd/network";
+const CONFIG_FILE_PREFIX: &str = "10-";
 
 pub(crate) enum NetworkDConfigFile {
     Network(NetworkConfig),
     NetDev(NetDevConfig),
+}
+
+impl NetworkDConfigFile {
+    pub(crate) fn write_config_file(&self) -> Result<()> {
+        match self {
+            NetworkDConfigFile::Network(network) => network.write_config_file(),
+            NetworkDConfigFile::NetDev(netdev) => netdev.write_config_file(),
+        }
+    }
 }
--- a/sources/api/netdog/src/networkd/config/mod.rs
+++ b/sources/api/netdog/src/networkd/config/mod.rs
@@ -5,7 +5,7 @@ mod network;
 
 use super::Result;
 pub(crate) use netdev::{NetDevBuilder, NetDevConfig};
-use network::NetworkConfig;
+pub(crate) use network::{NetworkBuilder, NetworkConfig};
 
 const NETWORKD_CONFIG_DIR: &str = "/etc/systemd/network";
 const CONFIG_FILE_PREFIX: &str = "10-";
@@ -18,7 +18,7 @@ pub(crate) enum NetworkDConfigFile {
 impl NetworkDConfigFile {
     pub(crate) fn write_config_file(&self) -> Result<()> {
         match self {
-            NetworkDConfigFile::Network(network) => network.write_config_file(),
+            NetworkDConfigFile::Network(network) => network.write_config_file(NETWORKD_CONFIG_DIR),
             NetworkDConfigFile::NetDev(netdev) => netdev.write_config_file(NETWORKD_CONFIG_DIR),
         }
     }
--- a/sources/api/netdog/src/networkd/config/mod.rs
+++ b/sources/api/netdog/src/networkd/config/mod.rs
@@ -4,7 +4,7 @@ mod netdev;
 mod network;
 
 use super::Result;
-use netdev::NetDevConfig;
+pub(crate) use netdev::{NetDevBuilder, NetDevConfig};
 use network::NetworkConfig;
 
 const NETWORKD_CONFIG_DIR: &str = "/etc/systemd/network";
@@ -19,7 +19,62 @@ impl NetworkDConfigFile {
     pub(crate) fn write_config_file(&self) -> Result<()> {
         match self {
             NetworkDConfigFile::Network(network) => network.write_config_file(),
-            NetworkDConfigFile::NetDev(netdev) => netdev.write_config_file(),
+            NetworkDConfigFile::NetDev(netdev) => netdev.write_config_file(NETWORKD_CONFIG_DIR),
         }
+    }
+}
+
+// This private module defines some empty traits meant to be used as type parameters for the
+// networkd config builders.  The type parameters limit the methods that can be called on the
+// builders so a user of this code can't inadvertently add configuration options that aren't
+// applicable to a particular device.  For example, a user can't add bond monitoring options to a
+// VLAN config.
+//
+// The following traits and enums are only meant to be used within the config module of this crate;
+// putting them in a private module guarantees this behavior.  See the "sealed trait" pattern here:
+// https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed
+mod private {
+    // The following zero-variant enums represent the device types we currently support.  They
+    // cannot be constructed and exist only as phantom types.
+    pub enum Bond {}
+    pub enum Interface {}
+    pub enum Vlan {}
+    pub enum BondWorker {} // interfaces that are bound to a bond
+
+    // The devices for which we are generating a configuration file.  All device types should
+    // implement this trait.
+    pub trait Device {}
+    impl Device for Bond {}
+    impl Device for Interface {}
+    impl Device for Vlan {}
+    impl Device for BondWorker {}
+
+    // Devices not bound to a bond, i.e. everything EXCEPT BondWorker(s)
+    pub trait NotBonded {}
+    impl NotBonded for Bond {}
+    impl NotBonded for Interface {}
+    impl NotBonded for Vlan {}
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::networkd::devices::{NetworkDBond, NetworkDInterface, NetworkDVlan};
+    use serde::Deserialize;
+    use std::fs;
+    use std::path::{Path, PathBuf};
+
+    pub(super) const BUILDER_DATA: &str = include_str!("../../../test_data/networkd/builder.toml");
+
+    pub(super) fn test_data() -> PathBuf {
+        PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("test_data")
+            .join("networkd")
+    }
+
+    #[derive(Debug, Deserialize)]
+    pub(super) struct TestDevices {
+        pub(super) interface: Vec<NetworkDInterface>,
+        pub(super) bond: Vec<NetworkDBond>,
+        pub(super) vlan: Vec<NetworkDVlan>,
     }
 }
--- a/sources/api/netdog/src/networkd/config/netdev.rs
+++ b/sources/api/netdog/src/networkd/config/netdev.rs
@@ -100,12 +100,14 @@ impl Display for ArpValidate {
 #[derive(Debug)]
 enum ArpAllTargets {
     All,
+    Any,
 }
 
 impl Display for ArpAllTargets {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             ArpAllTargets::All => write!(f, "all"),
+            ArpAllTargets::Any => write!(f, "any"),
         }
     }
 }
--- a/sources/api/netdog/src/networkd/config/netdev.rs
+++ b/sources/api/netdog/src/networkd/config/netdev.rs
@@ -1,7 +1,12 @@
+use super::{CONFIG_FILE_PREFIX, NETWORKD_CONFIG_DIR};
 use crate::interface_id::InterfaceName;
+use crate::networkd::{error, Result};
 use crate::vlan_id::VlanId;
+use snafu::{OptionExt, ResultExt};
 use std::fmt::Display;
+use std::fs;
 use std::net::IpAddr;
+use std::path::{Path, PathBuf};
 use systemd_derive::{SystemdUnit, SystemdUnitSection};
 
 #[derive(Debug, Default, SystemdUnit)]
@@ -109,5 +114,34 @@ impl Display for ArpAllTargets {
             ArpAllTargets::All => write!(f, "all"),
             ArpAllTargets::Any => write!(f, "any"),
         }
+    }
+}
+
+impl NetDevConfig {
+    const FILE_EXT: &str = "netdev";
+
+    /// Write the config to the proper directory with the proper prefix and file extention
+    pub(crate) fn write_config_file(&self) -> Result<()> {
+        let cfg_path = self.config_path()?;
+
+        fs::write(&cfg_path, self.to_string()).context(error::NetworkDConfigWriteSnafu {
+            what: "netdev_config",
+            path: cfg_path,
+        })
+    }
+
+    /// Build the proper prefixed path for the config file
+    fn config_path(&self) -> Result<PathBuf> {
+        let device_name = &self.netdev.as_ref().and_then(|n| n.name.clone()).context(
+            error::ConfigMissingNameSnafu {
+                what: "netdev config".to_string(),
+            },
+        )?;
+
+        let filename = format!("{}{}", CONFIG_FILE_PREFIX, device_name);
+        let mut path = Path::new(NETWORKD_CONFIG_DIR).join(filename);
+        path.set_extension(Self::FILE_EXT);
+
+        Ok(path)
     }
 }
--- a/sources/api/netdog/src/networkd/config/netdev.rs
+++ b/sources/api/netdog/src/networkd/config/netdev.rs
@@ -1,10 +1,13 @@
+use super::private::{Bond, Device, Vlan};
 use super::{CONFIG_FILE_PREFIX, NETWORKD_CONFIG_DIR};
+use crate::bonding::{ArpMonitoringConfigV1, ArpValidateV1, BondModeV1, MiiMonitoringConfigV1};
 use crate::interface_id::InterfaceName;
 use crate::networkd::{error, Result};
 use crate::vlan_id::VlanId;
 use snafu::{OptionExt, ResultExt};
 use std::fmt::Display;
 use std::fs;
+use std::marker::PhantomData;
 use std::net::IpAddr;
 use std::path::{Path, PathBuf};
 use systemd_derive::{SystemdUnit, SystemdUnitSection};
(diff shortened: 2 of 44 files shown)


sources/api/netdog/src/networkd/config/netdev.rs
```diff
@@ -100,12 +108,214 @@ impl Display for ArpValidate {
...
+            BondMonitoringConfigV1::ArpMon(arpmon) => netdev.with_arpmon_config(arpmon),
+        }
+        netdev.build()
+    }
+
+    fn netdev_from_vlan(vlan: NetworkDVlan) -> NetDevConfig {
+        let mut netdev = NetDevBuilder::new_vlan(vlan.name.clone());
+        netdev.with_vlan_id(vlan.id);
+        netdev.build()
+    }
```
zmrow, line 293: This logic will most likely go away once it's integrated into the `NetworkD*` device code.  It exists to enable unit testing now before that code is available.

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -121,3 +138,477 @@ impl Display for DhcpBool {
...
+    }
+
+    fn network_from_bond(bond: NetworkDBond) -> NetworkConfig {
+        let mut network = NetworkBuilder::new_bond(bond.name.clone());
+        network.with_dhcp(bond.dhcp4, bond.dhcp6);
+        bond.static4.map(|s| network.with_static_config(s));
+        bond.static6.map(|s| network.with_static_config(s));
+        bond.routes.map(|r| network.with_routes(r));
+        network.build()
+    }
```
zmrow, line 558: This logic will most likely go away once it's integrated into the `NetworkD*` device code.  It exists to enable unit testing now before that code is available.

sources/api/netdog/src/networkd/config/mod.rs
```diff
@@ -3,10 +3,23 @@
 mod netdev;
 mod network;
 
+use super::Result;
 use netdev::NetDevConfig;
 use network::NetworkConfig;
 
+const NETWORKD_CONFIG_DIR: &str = "/etc/systemd/network";
+const CONFIG_FILE_PREFIX: &str = "10-";
+
```
cbgbt: How is this prefix selected? How do we know it doesn't conflict with other files?

sources/api/netdog/test_data/networkd/network/eno3.network
```diff
@@ -0,0 +1,13 @@
...
+RequiredForOnline=true
+RequiredFamilyForOnline=ipv4
+[Network]
+DHCP=ipv4
+[DHCPv4]
+UseDNS=true
+UseDomains=true
+[DHCPv6]
+UseDNS=true
+UseDomains=true
```
yeazelm: I would expect the configuration to disable this if `dhcp6 = false` in our config? Or are these innocuous to networkd under the hood?

sources/api/netdog/test_data/networkd/network/eno4.network
```diff
@@ -0,0 +1,13 @@
+[Match]
+Name=eno4
+[Link]
+RequiredForOnline=true
+RequiredFamilyForOnline=ipv6
+[Network]
+DHCP=ipv6
+[DHCPv4]
+UseDNS=true
+UseDomains=true
```
yeazelm: Same as above for ipv6, should we set these to false (or omit them) if not asked to configure that side?

sources/api/netdog/test_data/networkd/network/eno7.network
```diff
@@ -0,0 +1,13 @@
...
+RequiredForOnline=true
+RequiredFamilyForOnline=ipv4
+[Network]
+DHCP=yes
+[DHCPv4]
+UseDNS=true
+UseDomains=true
+[DHCPv6]
+UseDNS=true
+UseDomains=true
```
yeazelm, line 13: I don't see it attempting to optionally get DHCPv6 as set according to 
```
[[interface]]
name = "eno7"
dhcp4 = true
[interface.dhcp6]
enabled = true
optional = true
```

yeazelm: Ok, I see what I missed before, its the difference between `DHCP=yes` and `DHCP=ip4` etc.

sources/api/netdog/test_data/networkd/network/eno3.network
zmrow: Good point - I overlooked that.  I'll leave those entries out if DHCP6=false

sources/api/netdog/test_data/networkd/network/eno7.network
zmrow: Other important bits are in the `[Link]` section, and more specifically the `RequiredForOnline` and `RequiredFamilyForOnline` entries.  These should tell `systemd-networkd-wait-online` that we do need this interface to come up, but we only really care about ipv4.

sources/api/netdog/test_data/networkd/builder.toml
```diff
@@ -0,0 +1,214 @@
...
+to = "3001:dead:beef::2/64"
+from = "2001:dead:beef::2"
+via = "2001:beef:beef::1"
+
+# Bonds and vlans
+[[vlan]]
+name = "myvlan"
+id = 42
+dhcp4 = true
+device = "eno1"
```
yeazelm, line 161: I might be missing this, but I didn't see the `.network` file that would tie this device to the vlan?

sources/api/netdog/src/networkd/devices/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+mod interface;
+mod vlan;
+
+pub(crate) use bond::NetworkDBond;
+pub(crate) use interface::NetworkDInterface;
+pub(crate) use vlan::NetworkDVlan;
+
+pub(crate) enum NetworkDDevice {
+    Interface(NetworkDInterface),
+    BondDevice(NetworkDBond),
```
zmrow: Adding a comment here to remind myself to remove the `Device` from `Bond` and `Vlan`... :facepalm:

sources/api/netdog/src/networkd/config/netdev.rs
```diff
@@ -144,4 +147,175 @@ impl NetDevConfig {
...
+    netdev: NetDevConfig,
+    spooky: PhantomData<T>,
+}
+
+impl<T: Device> NetDevBuilder<T> {
+    // The builder stores and directly manipulates the NetDevConfig being created to avoid storing
+    // yet another copy of the data supplied to the builder.  Generally builders have a `build()`
+    // method that returns the type being built; stick with that pattern here.  This also avoids
+    // proliferating the type parameters and provides the opportunity to add any additional
+    // creation logic needed in the future.
```
cbgbt: I'm not sure I follow the comment about storing yet another copy of the data, but I think the rest of this comment may be superfluous.

sources/api/netdog/src/networkd/config/netdev.rs
```diff
@@ -144,4 +147,175 @@ impl NetDevConfig {
...
+    pub(crate) fn build(self) -> NetDevConfig {
+        self.netdev
+    }
+}
+
+impl NetDevBuilder<Bond> {
+    /// Create a new .netdev config for a bond.
+    pub(crate) fn new_bond(name: InterfaceName) -> Self {
```
cbgbt, line 188: naming nit: I might just refer to this as `new`, calling it `new_bond` implied to me an ability to create multiple new bonds at once, but this is instantiating the builder.

sources/api/netdog/src/networkd/config/netdev.rs
```diff
@@ -144,4 +147,175 @@ impl NetDevConfig {
...
+        };
+
+        Self {
+            netdev,
+            spooky: PhantomData,
+        }
+    }
+
+    /// Add bond mode
+    pub(crate) fn with_mode(&mut self, mode: BondModeV1) {
```
cbgbt, line 204: nit: You might be able to avoid maintaining this match statement using `Into`.
```suggestion
    pub(crate) fn with_mode<B: Into<BondMode>>(&mut self, mode: B) {
```

sources/api/netdog/src/networkd/config/mod.rs
```diff
@@ -23,3 +23,67 @@ impl NetworkDConfigFile {
...
+    pub(super) fn test_data() -> PathBuf {
+        PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("test_data")
+            .join("networkd")
+    }
+
+    pub(super) fn builder_data() -> PathBuf {
+        test_data().join("builder.toml")
+    }
```
cbgbt: nit: You can simplify test data loading like this using `include_str!` to just include the text directly into the test binaries.

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -1,14 +1,24 @@
...
 use std::fmt::Display;
 use std::fs;
+use std::marker::PhantomData;
 use std::net::IpAddr;
 use std::path::{Path, PathBuf};
 use systemd_derive::{SystemdUnit, SystemdUnitSection};
 
+lazy_static! {
+    static ref DEFAULT_ROUTE_IPV4: IpNet = "0.0.0.0/0".parse().unwrap();
+    static ref DEFAULT_ROUTE_IPV6: IpNet = "::/0".parse().unwrap();
```
cbgbt, line 19: nit: It's probably safer/better practice to use the proper constructors for these, rather than parsing a string. While we can guarantee these are correct in code review, this would theoretically panic at runtime and can't be checked by the compiler.

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -169,4 +201,414 @@ impl NetworkConfig {
...
+    /// also provides the opportunity to add any additional logic needed in the future.
+    pub(crate) fn build(self) -> NetworkConfig {
+        self.network
+    }
+}
+
+impl NetworkBuilder<Interface> {
+    /// Create a new .network config for an interface; not meant for an interface that will be a
+    /// bond worker.  See `.new_bond_worker()` for that case.
+    pub(crate) fn new_interface<I>(id: I) -> Self
```
cbgbt, line 255: Same nit here as with bonds.

sources/api/netdog/src/networkd/config/netdev.rs
zmrow: Unfortunately, a `NetDevBuilder<T>` can't have multiple `new()` methods (I was a little surprised too...), hence making the constructors a bit more descriptive about what will be returned.

sources/api/netdog/src/networkd/config/mod.rs
zmrow: Netplan uses a similar `10-` prefix, and I didn't see a compelling reason to change prefix.

The reason for the prefix is that systemd will process all of these files in lexicographical order.  We should be the only process putting files here.

sources/api/netdog/src/networkd/config/netdev.rs
zmrow: Sorry I should have been a bit more clear.  I was alluding to the fact that builders sometimes store all of the data for the type it is constructing, and then do all the processing during a `build()` method.  Instead here we are directly manipulating the underlying type rather than store a copy of the data.

sources/api/netdog/test_data/networkd/builder.toml
zmrow: That will come next in the integration tests.  These unit tests just test each single builder in isolation.

Some logic outside the builder, and inside the devices, will gather the map of vlan -> device, and drive the builder to add the VLAN to the network config.

sources/api/netdog/src/networkd/config/netdev.rs
zmrow: I think I'll remove it - it does seem a bit superfluous.

sources/api/netdog/src/networkd/config/netdev.rs
zmrow: I thought about that and decided against it for no reason other than it felt like an odd place to toss in a slightly different interface and another trait to impl, which would contain the `match` statement anyway.  Happy to make the change if you feel it's important.

sources/api/netdog/src/networkd/config/netdev.rs
```diff
@@ -111,3 +116,32 @@ impl Display for ArpAllTargets {
...
+    /// Build the proper prefixed path for the config file
+    fn config_path(&self) -> Result<PathBuf> {
+        let device_name = &self.netdev.as_ref().and_then(|n| n.name.clone()).context(
+            error::ConfigMissingNameSnafu {
+                what: "netdev config".to_string(),
+            },
+        )?;
+
+        let filename = format!("{}{}", CONFIG_FILE_PREFIX, device_name);
+        let mut path = Path::new(NETWORKD_CONFIG_DIR).join(filename);
```
webern: If you pass in `NETWORKD_CONFIG_DIR` then file writing and path generation becomes test-able.

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -123,3 +128,45 @@ impl Display for DhcpBool {
...
+
+        fs::write(&cfg_path, self.to_string()).context(error::NetworkDConfigWriteSnafu {
+            what: "network config",
+            path: cfg_path,
+        })
+    }
+    /// Build the proper prefixed path for the config file
+    fn config_path(&self) -> Result<PathBuf> {
+        let match_section = self
+            .r#match
```
webern, line 178: woah what is this syntax?

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -123,3 +128,45 @@ impl Display for DhcpBool {
...
+            (None, None) => {
+                return error::ConfigMissingNameSnafu {
+                    what: "network_config".to_string(),
+                }
+                .fail();
+            }
+        };
+
+        let filename = format!("{}{}", CONFIG_FILE_PREFIX, device_name);
+        let mut cfg_path = Path::new(NETWORKD_CONFIG_DIR).join(filename);
```
webern: Pass this in and you can test file writing: `NETWORKD_CONFIG_DIR`

sources/api/netdog/src/networkd/config/mod.rs
```diff
@@ -23,3 +23,67 @@ impl NetworkDConfigFile {
         }
     }
 }
+
+// This private module defines some empty traits meant to be used as type parameters for the
```
webern, line 27: nit: doc comments seem fine to me everywhere they are allowed, even on private symbols. They have the benefit of rendering some markdown syntax highlighting.

sources/api/netdog/src/networkd/config/netdev.rs
```diff
@@ -144,4 +147,175 @@ impl NetDevConfig {
 
         Ok(path)
     }
+
+    // The following *mut() methods are private and primarily meant for use by the NetDevBuilder.
```
webern, line 151: nit, another case where I would use doc comments even though it is private. not sure if we have a consistent style on this.

sources/api/netdog/src/networkd/config/netdev.rs
webern: Yeah, it seems overly geared toward the PR, i.e. to convince reviewers that the design is right, which becomes somewhat moot after it is merged.

sources/api/netdog/src/networkd/config/netdev.rs
```diff
@@ -144,4 +147,175 @@ impl NetDevConfig {
...
+impl NetDevBuilder<Bond> {
+    /// Create a new .netdev config for a bond.
+    pub(crate) fn new_bond(name: InterfaceName) -> Self {
+        let netdev = NetDevConfig {
+            netdev: Some(NetDevSection {
+                name: Some(name),
+                kind: Some(NetDevKind::Bond),
+            }),
+            ..Default::default()
+        };
```
webern, line 195: I tried to think of a way you could have a single `new` function and somehow specialize on type, but that is not possible. Overall the idea of specializing on type in order to expose only the correct functions seems consistent with Rust idioms.

sources/api/netdog/src/networkd/config/network.rs
webern: Unless the constructor is a const function then it's unavoidable, right? What you could do is move the constants to the same module that defines IpNet and put these strings directly into the struct bypassing the parsing constructor.

sources/api/netdog/src/networkd/config/network.rs
zmrow: The `r#match`?  Since `match` is a rust keyword you need to "escape" it in order to use it as an identifier.  "match" makes the most sense in this context, so I opted not to rename it.

sources/api/netdog/src/networkd/config/network.rs
zmrow: I think it's unavoidable, even if we use the proper constructors. 

> move the constants to the same module that defines IpNet

IpNet is an external crate, so I don't think that will be possible.

sources/api/netdog/src/networkd/config/netdev.rs
zmrow: > not sure if we have a consistent style on this.

I believe we've typically followed this pattern of doc comments on "public" symbols/functions, and regular comments on "private" symbols and things.

zmrow:
The above pushes address all comments other than those still being discussed:
* Omit spurious DHCP4/6 sections of the protocol is disabled
* Pass a path to `config_path()` to enable testing and add tests
* Remove spurious `*Device` from bond/vlan devices in networkd device enum

zmrow:
^ Uses `include_str!()` for the unit test reference data as suggested by @cbgbt

webern (approved):

yeazelm (approved):
LGTM!

cbgbt (approved):

PR #3221, opened 2023-06-22
testsys: Add `--wait` flag for test runs

ecpullen:
**Issue number:**

Closes #

**Description of changes:**

```
testsys: Add `--wait` flag for test runs
    
    Adds 2 new flags to `testsys run`. If `--wait` is used, TestSys will
    block and monitor the status of all crds created and report the results.
    `--output-path` can be used to write the test results in json form to a
    desired location.
```

**Testing done:**

Ran `cargo make test` with `TESTSYS_WAIT=true` and tried setting `TESTSYS_OUTPUT_PATH` to verify output.

Sample output file:
```json
{
  "k-test-6-quick": [
    {
      "outcome": "pass",
      "numPassed": 1,
      "numFailed": 0,
      "numSkipped": 6972,
      "otherInfo": ""
    }
  ]
}
```

**Terms of contribution:**

### DIFF
--- a/tools/testsys/src/error.rs
+++ b/tools/testsys/src/error.rs
@@ -1,7 +1,9 @@
 use aws_sdk_ec2::error::DescribeImagesError;
 use aws_sdk_ec2::types::SdkError;
 use snafu::Snafu;
+use std::io;
 use std::path::PathBuf;
+use testsys_model::ResourceError;
 
 pub type Result<T> = std::result::Result<T, Error>;
 
@@ -73,6 +75,12 @@ pub enum Error {
     #[snafu(context(false), display("{}", source))]
     PubsysConfig { source: pubsys_config::Error },
 
+    #[snafu(display("Resource '{}' failed to be created: {}", resource_name, error))]
+    ResourceCreation {
+        resource_name: String,
+        error: ResourceError,
+    },
+
     #[snafu(display("Unable to create secret name for '{}': {}", secret_name, source))]
     SecretName {
         secret_name: String,
@@ -96,6 +104,14 @@ pub enum Error {
         source: testsys_model::test_manager::Error,
     },
 
+    #[snafu(display("Test '{}' failed to run: {}", test_name, error))]
+    TestRun { test_name: String, error: String },
+
+    #[snafu(display("{}", source))]
+    TestsysClient {
+        source: testsys_model::clients::Error,
+    },
+
     #[snafu(context(false), display("{}", source))]
     TestsysConfig { source: testsys_config::Error },
 
@@ -117,5 +133,18 @@ pub enum Error {
     #[snafu(display("Error reading config: {}", source))]
     VmwareConfig {
         source: pubsys_config::vmware::Error,
+    },
+
+    #[snafu(display("Timed out waiting for '{}': {}", what, source))]
+    WaitTimeout {
+        what: String,
+        source: tokio::time::error::Elapsed,
+    },
+
+    #[snafu(display("Unable to write {} to '{}': {}", what, path.display(), source))]
+    Write {
+        what: String,
+        path: PathBuf,
+        source: io::Error,
     },
 }
--- a/tools/testsys/src/run.rs
+++ b/tools/testsys/src/run.rs
@@ -5,6 +5,7 @@ use crate::error;
 use crate::error::Result;
 use crate::metal_k8s::MetalK8sCreator;
 use crate::vmware_k8s::VmwareK8sCreator;
+use crate::wait::wait_for_crds;
 use bottlerocket_variant::Variant;
 use clap::Parser;
 use log::{debug, info};
@@ -16,9 +17,10 @@ use pubsys_config::InfraConfig;
 use serde::{Deserialize, Serialize};
 use serde_plain::{derive_display_from_serialize, derive_fromstr_from_deserialize};
 use snafu::{OptionExt, ResultExt};
-use std::fs::read_to_string;
+use std::fs::{self, read_to_string};
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::time::Duration;
 use testsys_config::{GenericVariantConfig, ResourceAgentType, TestConfig};
 use testsys_model::test_manager::TestManager;
 use testsys_model::SecretName;
@@ -114,6 +116,18 @@ pub(crate) struct Run {
     /// The template file that should be used for custom testing.
     #[arg(long = "template-file", short = 'f')]
     custom_crd_template: Option<PathBuf>,
+
+    /// Wait for the test to finish before exiting
+    #[clap(long, short = 'w', env = "TESTSYS_WAIT")]
+    pub wait: bool,
+
+    /// Error if test execution takes longer than `wait_timeout_secs`.
+    #[clap(long, env = "TESTSYS_WAIT_TIMEOUT", default_value = "10800")]
+    pub wait_timeout_secs: u64,
+
+    /// The location to write the test results instead of stdout.
+    #[clap(long, env = "TESTSYS_OUTPUT_PATH", requires = "wait")]
+    output_path: Option<PathBuf>,
 }
 
 /// This is a CLI parsable version of `testsys_config::GenericVariantConfig`.
@@ -435,9 +449,48 @@ impl Run {
         };
 
         debug!("Adding crds to testsys cluster");
-        for crd in crds {
-            let crd = client.create_object(crd).await?;
+        for crd in &crds {
+            let crd = client.create_object(crd.clone()).await?;
             info!("Successfully added '{}'", crd.name().unwrap());
+        }
+
+        if self.wait {
+            info!(
+                "Waiting for test completion. Timing out after {} seconds.",
+                self.wait_timeout_secs
+            );
+            let results = tokio::time::timeout(
+                Duration::from_secs(self.wait_timeout_secs),
+                wait_for_crds(&client, &crds),
+            )
+            .await
+            .context(error::WaitTimeoutSnafu {
+                what: "test execution",
+            })??;
+
+            debug!("Testing completed writing results");
+            if let Some(output) = self.output_path {
+                fs::write(
+                    &output,
+                    serde_json::to_string_pretty(&results).context(error::SerdeJsonSnafu {
+                        what: "Unable to deserialize test results",
+                    })?,
+                )
+                .context(error::WriteSnafu {
+                    what: "JSON formatted test results",
+                    path: output,
+                })?;
+            } else {
+                for (test, res) in results {
+                    println!("{test}");
+                    for result in res {
+                        println!(
+                            "{}: {} Pass {} Fail",
+                            result.outcome, result.num_passed, result.num_failed
+                        );
+                    }
+                }
+            }
         }
 
         Ok(())
--- a/tools/testsys/src/wait.rs
+++ b/tools/testsys/src/wait.rs
@@ -1,0 +1,359 @@
+use crate::error::{self, Result};
+use futures::future::join_all;
+use log::{error, info, warn};
+use snafu::{OptionExt, ResultExt};
+use std::collections::HashMap;
+use std::time::Duration;
+use testsys_model::clients::{AllowNotFound, CrdClient};
+use testsys_model::test_manager::TestManager;
+use testsys_model::{
+    Crd, CrdExt, DestructionPolicy, Outcome, Resource, ResourceError, TaskState, Test, TestResults,
+};
+use tokio_retry::strategy::{jitter, ExponentialBackoff};
+use tokio_retry::Retry;
+
+const WAIT_TIMEOUT: Duration = Duration::from_secs(600);
+const WAIT_CHECK: Duration = Duration::from_secs(10);
+
+const RETRY_ATTEMPTS: usize = 5;
+
+/// Wait until all of the created crds have completed and return a HashMap containing each `Test`'s
+/// results.
+pub(crate) async fn wait_for_crds(
+    client: &TestManager,
+    crds: &Vec<Crd>,
+) -> Result<HashMap<String, Vec<TestResults>>> {
+    // Get list of all dependencies of the created crds
+    let dependencies: Vec<_> = crds
+        .iter()
+        .filter_map(|crd| match crd {
+            Crd::Test(test) => Some(test.spec.resources.to_owned()),
+            Crd::Resource(resource) => resource.spec.depends_on.to_owned(),
+        })
+        .flatten()
+        .collect();
+
+    // Get a list of all conflicting resources with the created crds
+    let conflicting_resource_names: Vec<_> = crds
+        .iter()
+        .filter_map(|crd| match crd {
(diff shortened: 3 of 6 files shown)


tools/testsys/src/run.rs
```diff
@@ -114,6 +115,19 @@ pub(crate) struct Run {
...
+    #[clap(long, short = 'w', env = "TESTSYS_WAIT")]
+    pub wait: bool,
+
+    /// The location to write the test results instead of stdout.
+    #[clap(
+        long,
+        env = "TESTSYS_OUTPUT_PATH",
+        parse(from_os_str),
+        requires = "wait"
+    )]
```
etungsten: Why is the `output-path` flag tied to the `wait` flag?

Can this instead be tied to the `testsys status` command?

tools/testsys/src/wait.rs
```diff
@@ -0,0 +1,285 @@
...
+                ) {
+                    warn!("A conflicting resource '{resource_name}' with `DestructionPolicy:OnTestSuccess' is waiting for a failed test '{}' to be deleted. The test will need to be deleted manually.", failed_test.object_name());
+                }
+            }
+        }
+
+        info!(
+            "Still waiting for resource '{resource_name}' to be deleted from the cluster. Sleeping 60s"
+        );
+        tokio::time::sleep(Duration::from_secs(60)).await;
```
etungsten: 60 seconds between polls seem a bit long to me. Any reasoning behind this?
Can we capture the duration of the wait in a const can reuse it whereever it applies?

tools/testsys/src/run.rs
ecpullen: The result of the tests created by the invocation are output to this output-path. We should also add the same feature to status, but in a separate pr.

tools/testsys/src/wait.rs
ecpullen: 60 seconds was arbitrary, but I didn't want to go too short since it could be polling for hours, and 60 seconds didn't seem too bad. I'll bring it out as a constant anyways.

tools/testsys/src/run.rs
ecpullen: This pr is enabling automation to call `cargo make -e TESTSYS_WAIT=true -e TESTSYS_OUTPUT_PATH=foo.json test` and then immediately use `jq` on `foo.json` to get the results of the test without having to deal with TestSys itself.

tools/testsys/src/wait.rs
webern: I'd vote for something like 10 just so you don't waste time when you're doing a `quick` run, i.e. if this becomes part of ones developer flow.

tools/testsys/src/wait.rs
```diff
@@ -0,0 +1,285 @@
...
+/// Wait until the CRD has completed and determine whether it was successful or not.
+async fn wait_for_crd(client: &TestManager, crd: &Crd) -> Result<TestRunResults> {
+    match crd {
+        Crd::Test(test) => wait_for_test(client, test.object_name()).await,
+        Crd::Resource(resource) => wait_for_resource_creation(client, resource.object_name()).await,
+    }
+}
+
+#[derive(Debug)]
+pub enum TestRunResults {
```
webern: A plural noun implies some sort of collection.

```suggestion
pub enum TestRunResult {
```

webern (commented):
There should be a configurable overall timeout, i.e. we might want the `--wait` invocation to timeout after 3 hours instead of hanging indefinitely.

tools/testsys/src/wait.rs
ecpullen: > There should be a configurable overall timeout, i.e. we might want the --wait invocation to timeout after 3 hours instead of hanging indefinitely.

We definitely don't want a default overall timeout. A sonobuoy test that is waiting for another sonobuoy test to finish would certainly take more than 3 hours. I'll add a timeout for waiting for other completed resources to be deleted (10 mins). I will also add an optional overall timeout.

ecpullen:
^ 
* Adds `TESTYS_WAIT_TIMEOUT` to optionally prevent hanging indefinitely
* Adds a timeout (10 mins) for the user to fix a resource/test that is blocking test execution
* Moves times to constants.
* Reduced polling time to 10s from 60s
* Renamed `TestRunResults` to `TestRunResult`

ecpullen:
I'm wondering if TestSys should also stream the logs from the agent's it is watching? Maybe we could add that in a follow up pr.

etungsten (approved):

webern (commented):
Is the same timeout being used for these two things?
- Timeout on a long-running test
- Timeout on a resource creation error

What about timeout on a long-running resource creation or deletion?

Have you prevented `--wait` from running for 12 hours when no progress is being made and no errors have been reported?

ecpullen:
> Is the same timeout being used for these two things?
> 
> * Timeout on a long-running test
> * Timeout on a resource creation error
> 
> What about timeout on a long-running resource creation or deletion?
> 
> Have you prevented `--wait` from running for 12 hours when no progress is being made and no errors have been reported?

The timeout for a long running test is the responsibility of the test agent not the run command. A test could take 24 hours to run as its standard time and we shouldn't hard code something that prevents that. Each of the timeouts you are worried about are timeouts that should be implemented elsewhere (mainly the agent itself). Adding any sort of timeout will create a situation where a new test that may take longer than others doesn't work without changing the hard coded timeouts. If you wanted a 3 hour timeout, you can set it with `TESTSYS_WAIT_TIMEOUT`.

webern:
> The timeout for a long running test is the responsibility of the test agent not the run command. 

My original idea, which still stands, is a timeout for the `--wait`... not for the agents. In other words, if testsys is borked, the agent is hung and will never timeout, then I want some timeout on the CLI program that would otherwise wait for it forever. This way we won't have a CI run that hangs forever if testsys isn't behaving as expected.

I also don't think that this command should attempt a cleanup of resources and tests when something bad happens. I don't think there should be this complex and hard to understand timeout window in which you can fix the problem. Instead I would expect the `--wait` command to `FAIL: 1` and leave the cluster in whatever state its in. I can then use that failure to decide whether I want to try to fix it or whether I want to delete all (and CI would likely `testsys delete --all`).

ecpullen:
> > The timeout for a long running test is the responsibility of the test agent not the run command.
> 
> My original idea, which still stands, is a timeout for the `--wait`... not for the agents. In other words, if testsys is borked, the agent is hung and will never timeout, then I want some timeout on the CLI program that would otherwise wait for it forever. This way we won't have a CI run that hangs forever if testsys isn't behaving as expected.

Is the configurable timeout acceptable? I am strongly against a default timeout of the `--wait` for the reasons above. With `TESTSYS_WAIT_TIMEOUT` the desired effect is reached, without potentially making the code unusable with new agents.

The `--wait` does have several `smart` timeouts (10 mins) already.
* Conflicting resource errored
* Conflicting resource won't be automatically cleaned up
* A test is errored or failed and won't be automatically cleaned up

> I also don't think that this command should attempt a cleanup of resources and tests when something bad happens. I don't think there should be this complex and hard to understand timeout window in which you can fix the problem. Instead I would expect the `--wait` command to `FAIL: 1` and leave the cluster in whatever state it's in. I can then use that failure to decide whether I want to try to fix it or whether I want to delete all (and CI would likely `testsys delete --all`).

There currently isn't anything being cleaned up after the test is complete. I have a local commit that will enable automatic clean up, but I think that belongs in a separate pr.

webern (changes requested):
I've asked @cbgbt to take a look.

tools/testsys/src/run.rs
```diff
@@ -435,11 +454,48 @@ impl Run {
...
+                    .await
+                    .context(error::WaitTimeoutSnafu {
+                        what: "test execution",
+                    })??
+            } else {
+                wait_for_crds(&client, &crds).await?
+            };
+
+            debug!("Testing completed writing results");
+            if let Some(output) = self.output_path {
```
cbgbt, line 472: I agree that it seems odd to pair the `output-path` with `--wait`. I wonder if this rendering logic should live in a function that would be callable when we add the same capability to `testsys status`?

tools/testsys/src/run.rs
```diff
@@ -435,11 +454,48 @@ impl Run {
...
 
         debug!("Adding crds to testsys cluster");
-        for crd in crds {
-            let crd = client.create_object(crd).await?;
+        for crd in &crds {
+            let crd = client.create_object(crd.clone()).await?;
             info!("Successfully added '{}'", crd.name().unwrap());
         }
 
+        if self.wait {
```
cbgbt, line 457: This method is too large. I think it could be greatly helped by moving the variant creator above to some kind of `Factory`-style separate method.

I think similarly abstracting out the `wait_for_crds` call to one that takes `timeout: Option<Duration>` and handles that separately would help.

tools/testsys/src/wait.rs
```diff
@@ -0,0 +1,352 @@
...
+use std::collections::HashMap;
+use std::time::Duration;
+use testsys_model::clients::{AllowNotFound, CrdClient};
+use testsys_model::test_manager::TestManager;
+use testsys_model::{
+    Crd, CrdExt, DestructionPolicy, Outcome, Resource, ResourceError, TaskState, Test, TestResults,
+};
+
+const WAIT_TIMEOUT_SECONDS: u64 = 600;
+const WAIT_CHECK_SECONDS: u64 = 10;
```
cbgbt: Why not just store these as `Duration` to avoid needing to encode the unit in the variable name?

tools/testsys/src/wait.rs
```diff
@@ -0,0 +1,352 @@
...
+}
+
+/// Try to fetch the names of all tests that are dependent on this resource.
+async fn fetch_tests_using_resource(
+    client: &TestManager,
+    resource_name: &str,
+) -> Result<Vec<Test>> {
+    let test_client = client.test_client();
+    let mut count = 0;
+    let n_iters = 5;
```
cbgbt: Should this be `const`?

tools/testsys/src/wait.rs
```diff
@@ -0,0 +1,352 @@
...
+    let tests = loop {
+        let tests = test_client.get_all().await;
```
cbgbt: Should we be using jitter/backoff here? It doesn't seem like we wait any interval before requesting the tests again.

tools/testsys/src/wait.rs
```diff
@@ -0,0 +1,352 @@
...
+            resource_result
+                .allow_not_found(|_| warn!("A resource could not be found."))
+                .context(error::TestsysClientSnafu)
+                .transpose()
+        })
+        .collect::<Result<Vec<Resource>>>();
+
+        match resources {
+            resources @ Ok(_) => break resources,
+            Err(_) if count < n_iters => {
```
cbgbt: Similar feedback here to the other fetch loops. We probably should be using retries with backoff and jitter. I also think it would be wise to funnel our client calls through rate-limiting algorithm (I like [governor](https://crates.io/crates/governor)).

cbgbt (changes requested):
In general, it's best to rate-limit requests to external services. Retry loops should use exponential backoff with jitter.

I also agree with @webern that a default (though large) timeout is a wise move. Unbounded waiting on a tool intended to be used in CI can be painful for a variety of reasons.

tools/testsys/src/run.rs
ecpullen: Can we proceed without these changes and I'll create an issue to address this in a followup pr?

tools/testsys/src/run.rs
ecpullen: I'm not sure what you mean.

ecpullen:
I have fixed most of the suggested changes and added a 3 hour default timeout that can be configured with `TESTSYS_WAIT_TIMEOUT`.

tools/testsys/src/error.rs
```diff
@@ -73,6 +75,12 @@ pub enum Error {
     #[snafu(context(false), display("{}", source))]
     PubsysConfig { source: pubsys_config::Error },
 
+    #[snafu(display("Resource '{}' failed to be created: {}", resource_name, error))]
```
webern, line 78: style nit:

```suggestion
    #[snafu(display("Failed to create resource '{}': {}", resource_name, error))]
```

tools/testsys/src/run.rs
webern: > Can we proceed without these changes and I'll create an issue to address this in a followup pr?

I would be Ok with that since the function is already too large before you added 40 lines to it. After refactor it should read more like:

let (foo, bar) = do_one_complex_thing().await?;
let baz = do_another_complex_thing(foo, bar).await?;
do_another_thing(foo, bar, baz).await?;
...etc

tools/testsys/src/run.rs
```diff
@@ -435,11 +449,50 @@ impl Run {
...
+                Duration::from_secs(self.wait_timeout_secs),
+                wait_for_crds(&client, &crds),
+            )
+            .await
+            .context(error::WaitTimeoutSnafu {
+                what: "test execution",
+            })??;
+
+            debug!("Testing completed writing results");
+            if let Some(output) = self.output_path {
```
webern, line 472:
```suggestion
            debug!("Testing completed");
            if let Some(output) = self.output_path {
            debug!("Writing test results");
```

tools/testsys/src/wait.rs
```diff
@@ -0,0 +1,359 @@
...
+    // For each CRD created, add a section to results with the test results.
+    let mut results = HashMap::new();
+    for crd in crds {
+        results.extend(wait_for_crd(client, crd).await?.context()?)
+    }
+
+    Ok(results)
+}
+
+/// Wait until the conflicting resources are deleted.
```
webern, line 65: Can you define what it means for a resource to be conflicting, here, or somewhere? Maybe it exists somewhere already. I think it means a resource that can't be used by more than one test simultaneously? Or one that cannot be created until another is destroyed?

tools/testsys/src/wait.rs
```diff
@@ -0,0 +1,359 @@
...
+        test: String,
+        error: String,
+    },
+    Successful {
+        results: HashMap<String, Vec<TestResults>>,
+    },
+}
+
+impl TestRunResult {
+    fn context(self) -> Result<HashMap<String, Vec<TestResults>>> {
```
webern, line 344: would `results(&self)` be a better name?

webern (approved):

tools/testsys/src/wait.rs
ecpullen: It is described when conflicting resources are identified. Conflicting resources are resources that must not exist before the resource creation can begin.

tools/testsys/src/wait.rs
ecpullen: I was trying to keep consistent with snafu, but I can change it if that's preferred.

cbgbt (approved):

ecpullen:
This pr is in the wrong git repo now that testsys has moved to twoliter. Closing.

PR #3223, opened 2023-06-23
[1.14.x] Pin rust version to 1.69.0

stmcginnis:
**Description of changes:**

We use rust 1.69.0 in the SDK, which does not use the cargo "sparse" protocol by default. In 1.70.0, this protocol does become the default. This should be fine, but the GitHub runner nodes just install the latest stable version of rust. Since some activity happens prior to our build in the SDK, this creates a mismatch where the cache contains sparse data that cannot be correctly consumed by the 1.69.0 rust from the SDK, causing failures.

This is just a temporary issue that currently only impacts the 1.14.x stable branch. For the time being, just pinning the rust version on the runner node on this branch should allow us to get around it.

**Terms of contribution:**


etungsten (approved):

stmcginnis:
Backporting SDK update, so this will not be necessary.

Issue #2373, opened 2022-08-23
Be consistent with command line parsing by standardizing on `clap` or `argh`


stmcginnis:
**What I'd like:**

Most of our CLIs use the `argh` crate for parsing command line input. But some are using `clap` instead.

It would be good to convert these over to `argh` so we are consistent across our code and avoid importing two different crates that provide the same functionality.

**Any alternatives you've considered:**

Could keep them as is since there is no conflict between using the mix of the two, but not a real reason not to just use one or the other.

alok-sin:
I would like to take this. Does this concern with the testsys repo as well? https://github.com/bottlerocket-os/bottlerocket-test-system/issues/535

stmcginnis:
Awesome, that would be great @alok-sin!

I'm not 100% on the `bottlerocket-test-system` changes yet. There has been some talk of removing some of that code, so I don't want to suggest you make changes there only to have it all be removed.

If you could start in this repo that would be ideal. Then we can hopefully have a little better understanding if anything should be updated in the other repo once that is done.

Thanks a lot!

webern:
How did you arrive at standardizing on argh instead of clap? I think clap is becoming the de-facto standard crate for this.

stmcginnis:
@bcressey had expressed a desire to go with argh over clap. If clap is becoming the standard for this, then perhaps we should reevaluate that position. I am fine either way - I would just like to see us using one or the other.

bcressey:
I don't feel as strongly about "only one" but I definitely don't want to add `clap` to the binaries shipped in the OS image, and would prefer that the mix of custom / `structopt` / `argh` that is currently shipped get trimmed down to just `argh`.

That's primarily for reasons of code size. It's not a huge factor for existing builds, but there are some use cases (like a microVM guest) where size is a concern. I don't like the idea of spending 10+ MiB of space for arg parsing that no one will ever see.

webern:
> I don't feel as strongly about "only one"

I also don't feel strongly about "only one" and would hate to see us go from clap -> argh in places where we are using clap now.

Have you checked the size of the compiled code of clap vs argh? Just curious if it's a big contributor to binary size.

stmcginnis:
`structopt` is deprecated, so we should review what we're using and migrate to whichever makes the most sense.

jpculp:
[...]

PR #3228, opened 2023-06-26
Update clap and remove use of structopt in tools

stmcginnis:
**Issue number:**

Related #2373

**Description of changes:**

This updates our `tools/` binaries to use the latest version of the `clap` crate. It also drops the use of `structopt` as that project is now in maintenance mode:

> As clap v3 is now out, and the structopt features are integrated into (almost as-is), structopt is now in maintenance mode: no new feature will be added.

So this standardizes our out-of-distro tools to just use the latest `clap`.

Note: unfortunately we need to pull in two different versions of `clap` due to our use of `generate-readme`. That project is no longer maintained as well, so we will either need to find a replacement or stop using it to generate README files.

**Testing done:**

Built all binaries. Compared `$bin --help` output between the old and new versions to verify there were no non-aesthetic differences.

Ran `cargo make && cargo make ami` to make sure project could still be built and published.

**Terms of contribution:**


tools/pubsys/src/aws/ami/mod.rs
```diff
@@ -30,49 +31,47 @@ use serde::{Deserialize, Serialize};
...
 
 const WARN_SEPARATOR: &str = "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!";
 
 /// Builds Bottlerocket AMIs using latest build artifacts
-#[derive(Debug, StructOpt)]
-#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
+#[derive(Debug, Parser)]
 pub(crate) struct AmiArgs {
     /// Path to the image containing the os volume
-    #[structopt(short = "o", long, parse(from_os_str))]
```
jpculp, line 43: For some reason I thought we had to do something more for the OsStr, but maybe I was overthinking.

stmcginnis: I think that may have been the case in older versions with `structopt`, but it appears `clap` handles things correctly.

tools/infrasys/src/main.rs
```diff
@@ -16,36 +17,33 @@ use std::collections::HashMap;
...
-    // Path to Infra.toml  (NOTE: must be specified before subcommand)
-    #[structopt(long, parse(from_os_str))]
+    // Path to Infra.toml (NOTE: must be specified before subcommand)
+    #[arg(long)]
     infra_config_path: PathBuf,
 
-    #[structopt(subcommand)]
-    subcommand: SubCommand,
+    #[command(subcommand)]
+    subcommand: SubCommands,
```
zmrow: Nit: enums are usually named with the singular form of the word, as an enum can only be one of it's contained variants and is not a collection.

tools/infrasys/src/main.rs
```diff
@@ -16,36 +17,33 @@ use std::collections::HashMap;
...
 }
 
-#[derive(Debug, StructOpt)]
-#[structopt(setting = clap::AppSettings::DeriveDisplayOrder)]
```
zmrow, line 40: I had a bit of trouble looking up `DeriveDisplayOrder`.  I'm assuming it doesn't exist now, but are we sure removing it doesn't change behavior?

tools/pubsys-setup/src/main.rs
```diff
@@ -14,33 +15,33 @@ use std::fs;
...
 /// Helps you get started with credentials to make Bottlerocket images and repos.
-#[derive(Debug, StructOpt)]
+#[derive(Debug, Parser)]
 struct Args {
-    #[structopt(global = true, long, default_value = "INFO")]
+    #[arg(global = true, long, default_value = "INFO")]
     /// How much detail to log; from least to most: ERROR, WARN, INFO, DEBUG, TRACE
     log_level: LevelFilter,
 
-    #[structopt(long, parse(from_os_str))]
```
zmrow, line 28: Does clap handle this properly?  We previously had to use the `parse` argument to do this.

tools/pubsys/src/aws/publish_ami/mod.rs
```diff
@@ -23,52 +24,49 @@ use std::collections::{HashMap, HashSet};
...
-    #[structopt(long, group = "mode")]
+    #[arg(long, group = "mode")]
     grant: bool,
     /// Revoke access from the given users/groups
-    #[structopt(long, group = "mode")]
+    #[arg(long, group = "mode")]
     revoke: bool,
 
-    #[structopt(flatten)]
+    #[command(flatten)]
```
zmrow, line 64: Is `command` used here to get access to `flatten`?

tools/pubsys/src/aws/validate_ami/mod.rs
```diff
@@ -10,34 +10,33 @@ use crate::aws::client::build_client_config;
...
+    #[arg(long)]
     expected_amis_path: PathBuf,
 
     /// Optional path where the validation results should be written
-    #[structopt(long, parse(from_os_str))]
+    #[arg(long)]
     write_results_path: Option<PathBuf>,
 
-    #[structopt(long, requires = "write-results-path")]
+    #[arg(long, requires = "write_results_path")]
```
zmrow, line 34: Why the change from dashes to underscores?  Does this still work properly?

tools/infrasys/src/main.rs
stmcginnis: Yeah, docs are really light on this, especially when using the `derive` approach we do.

It turns out `clap` has made [`DeriveDisplayOrder` the default](https://github.com/clap-rs/clap/pull/3975), so we no longer need to set this for each arg struct.

tools/pubsys-setup/src/main.rs
stmcginnis: Yes, it appears `clap` handles this correctly.

tools/pubsys/src/aws/publish_ami/mod.rs
stmcginnis: Yes, `clap` uses more specific attribute names to differentiate some of the options.

tools/pubsys/src/aws/validate_ami/mod.rs
stmcginnis: `clap` compares this value to the actual struct property, not the rendered argument name. So with the dashes it would error because there was no property found with that name. It needs to match the actual name on line 32.

tools/infrasys/src/main.rs
stmcginnis: Done!

tools/deny.toml
```diff
@@ -86,6 +83,9 @@ skip-tree = [
     # TestSys uses a newer version of base64 and serde_yaml
     { name = "testsys-model", version = "=0.0.8" },
     { name = "bottlerocket-types", version = "=0.0.8" },
+
+    # generate-readme uses an old version of clap
+    { name = "clap", version = "2.34.0" }
```
jpculp: Rather than skipping `clap`'s tree, I would just skip the tree for `generate-readme` itself.

stmcginnis: True, good call!

ecpullen (approved):
LGTM

zmrow (approved):
:moon_cake:

jpculp (approved):

etungsten (approved):

PR #3233, opened 2023-06-27
Updated EKS-D to 1.23-25

kschumy:
**Issue number:**

Closes #

**Description of changes:**

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.23/Cargo.toml
+++ b/packages/kubernetes-1.23/Cargo.toml
@@ -14,7 +14,7 @@ path = "/dev/null"
 package-name = "kubernetes-1.23"
 
 [[package.metadata.build-package.external-files]]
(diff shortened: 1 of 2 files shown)


kschumy:
Just realized the sha is the same. Apparently there were no changes to Kubernetes itself since the release you are using, which doesn't seem unreasonable to me. I'm guessing you probably don't care about updating to a newer version of EKS-D, but I'll leave this open for you to decide.

stmcginnis:
Hah, good catch. I was wondering why it was already in the lookaside cache.

If the sha didn't even change, then the end result in a Bottlerocket image won't be any different before or after this change. So I think we can probably just close this and wait for any more changes.

Thanks!

PR #3126, opened 2023-05-18
kernel: cherry-pick fixes for CVE-2023-32233

markusboehme:
**Issue number:** n/a

**Description of changes:** Cherry-pick the fixes for [CVE 2023-32233](https://seclists.org/oss-sec/2023/q2/133) in the Linux kernel from the respective kernel.org upstream stable trees.

**Testing done:**

* [x] Compiles
* [x] Boots
* [x] Reproducer errors out, does not escalate privileges, and does not destabilize the system

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

Issue #2289, opened 2022-07-21
kernel config: tidy up config fragments


foersleo:
There is some things to tidy up for the parts we set for our kernel configs:

- [ ] Usage of `CONFIG_FOO=n` instead of `# CONFIG_FOO is not set` to deactivate a setting
- [ ] Check if there is any conflicts that merge-config brings up (e.g. intended to have something build in, but dependency options are build as module; merge-config will complain about those)

foersleo:
There is some differences in the 5.10 kernel between the configs for x86_64 and aarch64 configurations we include from our upstream.

[ ] `aarch64` includes a bunch of IR drivers we do not need that are not in `x86_64`, see chain `RC_CORE` -> `RC_DECODERS` -> `IR_*_DECODER`
[ ] `aarch64` has some USB host controller drivers built-in vs. built as module for `x86_64`, see `USB_*HCI*`

markusboehme:
* [ ] ixgbe driver is only needed for metal variants
* [ ] smartpqi driver is only needed for metal variants
* [ ] become more consistent with selection of EDAC drivers, e.g. we disable `EDAC_I10NM` but have drivers for other Intel micro-archs enabled

PR #1862, opened 2021-12-09
Add IPMI kernel module to 5.10 kernel

arnaldo2792:
**Issue number:**
N / A

**Description of changes:**

Related to #1799 , that PR is quite big. I'm taking some of the commits in that PR as their individual PR. The commit in this PR is variant independent, since it changes a package used among all variants.

```
07d6bcf7 Add IPMI kernel module to 5.10 kernel
```

**Testing done:**
- Build aws-dev and launched an instance
- Confirmed that the kernel module can be loaded:

```shell
bash-5.0# modinfo ipmi_msghandler
filename:       /lib/modules/5.10.75/kernel/drivers/char/ipmi/ipmi_msghandler.ko
softdep:        post: ipmi_devintf
version:        39.2
description:    Incoming and outgoing message routing for an IPMI interface.
author:         Corey Minyard <minyard@mvista.com>
license:        GPL
srcversion:     6A1B2902B9322AFBA64E5DC
depends:
retpoline:      Y
intree:         Y
name:           ipmi_msghandler
vermagic:       5.10.75 SMP mod_unload modversions
sig_id:         PKCS#7
signer:
sig_key:
sig_hashalgo:   unknown
signature:
parm:           panic_op:Sets if the IPMI driver will attempt to store panic information in the event log in the event of a panic.  Set to 'none' for no, 'event' for a.
parm:           maintenance_mode_timeout_ms:The time (milliseconds) after the last maintenance message that the connection stays in maintenance mode. (ulong)
parm:           default_retry_ms:The time (milliseconds) between retry sends (ulong)
parm:           default_maintenance_retry_ms:The time (milliseconds) between retry sends in maintenance mode (ulong)
parm:           default_max_retries:The time (milliseconds) between retry sends in maintenance mode (uint)

```

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #3238, opened 2023-06-28
kernel: Update to 5.10.184, 5.15.117, and 6.1.29

foersleo:
**Issue number:** -

**Description of changes:**

Update kernels to latest AL kernels available in the repositories. With those updates we ingest upstream variants for nftables fix from https://github.com/bottlerocket-os/bottlerocket/pull/3126 , so drop the downstream backports.

Through AL we also got some additional config changes that I went ahead and reverted for us. Some of those are enabling quite substantial code changes that AL has backported (newer version of smartpqi driver and bbr2 congestion control algorithm). I do not think we want to enable those mid-series.

**Testing done:**

- [x] Compiles
- [x] Sonobuoy quick test

Sonobuoy quick test results. (No test for the 6.1 kernel update as 6.1 currently is only on `*-dev` variants):

```
> kubectl get nodes -o wide
NAME                                             STATUS   ROLES    AGE   VERSION                INTERNAL-IP     EXTERNAL-IP     OS-IMAGE                                KERNEL-VERSION   CONTAINER-RUNTIME
ip-192-168-8-237.eu-central-1.compute.internal   Ready    <none>   32m   v1.27.1-eks-61789d8    192.168.8.237   18.195.80.186   Bottlerocket OS 1.15.0 (aws-k8s-1.27)   5.15.117         containerd://1.6.20+bottlerocket
ip-192-168-93-82.eu-central-1.compute.internal   Ready    <none>   33m   v1.23.17-eks-1c2761d   192.168.93.82   3.64.127.141    Bottlerocket OS 1.15.0 (aws-k8s-1.23)   5.10.184         containerd://1.6.20+bottlerocket

> sonobuoy run --mode=quick --wait
[...]
19:06:34    systemd-logs   ip-192-168-8-237.eu-central-1.compute.internal   complete   passed                                        
19:06:34    systemd-logs   ip-192-168-93-82.eu-central-1.compute.internal   complete   passed                                        
19:06:34             e2e                                           global   complete   passed   Passed:  1, Failed:  0, Remaining:  0
```

Config diff reports quite some changes:

```
config-aarch64-aws-dev-diff:              1 removed,  19 added,   5 changed
config-aarch64-aws-k8s-1.23-diff:         2 removed,   3 added,   2 changed
config-aarch64-aws-k8s-1.26-diff:         2 removed,   2 added,   0 changed
config-aarch64-metal-dev-diff:            1 removed,  20 added,   5 changed
config-x86_64-aws-dev-diff:               1 removed,  11 added,   7 changed
config-x86_64-aws-k8s-1.23-diff:          1 removed,   6 added,   2 changed
config-x86_64-aws-k8s-1.26-diff:          2 removed,   1 added,   1 changed
config-x86_64-metal-dev-diff:             1 removed,  12 added,   7 changed
config-x86_64-metal-k8s-1.23-diff:        1 removed,   1 added,   0 changed
config-x86_64-metal-k8s-1.26-diff:        2 removed,   1 added,   1 changed
```

The full diff report can be found on [Gist](https://gist.github.com/foersleo/eea6512f5df8c97a535cbdbff4395bc6).

Summary of changes per kernel series:

**5.10**

- `BLK_DEV_SX8` - Driver has been removed upstream in v5.10.184
- `DRM_RCAR_LVDS` and `DRM_RCAR_USE_LVDS` - clean up of drm rcar config options in v5.10.183
- `INFINIBAND_I40W` - removed from aarch64 for consistency with x86_64
- `SCSI_DPT_I2O` - unclear why this one comes up now. The functionality is not new and it has been removed upstream, but the removal has not made it to 5.10 tree.
- `IXGBE*` and `MDIO` - Add the ixgbe driver to x86_64 for consistency with aarch64 - MDIO gets selected by IXGBE

**5.15**

- `BLK_DEV_SX8` - Driver has been removed upstream in v5.15.117
- `DRM_RCAR_LVDS` and `DRM_RCAR_USE_LVDS` - clean up of drm rcar config options in v5.15.116
- `SCSI_DPT_I2O` - unclear why this one comes up now. The functionality is not new and it has been removed upstream, but the removal has not made it to 5.10 tree.
- `HP_ACCEL` and `X86_PLATFORM_DRIVERS_HP` - move HP drivers into central location, renaming options in v5.15.113
- `X86_CPUID` - enable cpuid interface for consistency between kernel series

**6.1**

- `GCC12_NO_ARRAY_BOUNDS` - option removed to use version agnostic option `CC_NO_ARRAY_BOUNDS` in v6.1.26
- `CRYPTO_FIPS_NAME` - set to be AL specific. I am not quite sure if we want to carry this forward
- `PTP*` and `PPS*` - these are needed for functionality in the newer ENA drivers and a dependency for loading newer ena
- `SFC_SIENA` - depends on `PTP_1588_CLOCK` and is thus now visible
- `RT_GROUP_SCHED` - disable, as allocating real CPU bandwidth for task groups can be problematic with systemd (https://github.com/systemd/systemd/blob/main/README#L144-L150)
- `DP83640_PHY` - new driver for ultra-low-power Ethernet phy - not needed
- `HAVE_*`, `LIVEPATCH`, `OBJTOOL`, `STACK_VALIDATION`, `UNWINDER_FRAME_POINTER` - port livepatch capabilities to aarch64, we have these available on previous kernel series as well
- `X86_CPUID` and `X86_MSR` - enable cpuid and msr interface for consistency between kernel series

**Terms of contribution:**


packages/kernel-5.10/config-bottlerocket
```diff
@@ -152,3 +152,7 @@ CONFIG_EXT4_USE_FOR_EXT2=y
 
 # Disable unused MPI3MR driver AL carries as a backport
 # CONFIG_SCSI_MPI3MR is not set
+
+# Work with the previously used in-tree version of SMARTPQI instead of AL backport
+# CONFIG_AMAZON_SCSI_SMARTPQI is not set
+CONFIG_SCSI_SMARTPQI=m
```
markusboehme, line 167: For later: Since we have `CONFIG_SCSI_SMARTPQI=y` for the metal variants, we might as well disable the driver completely for the cloud and VMware variants. I agree with just retaining the status quo for this rebase, though.

packages/kernel-5.10/config-bottlerocket
```diff
@@ -155,4 +155,6 @@ CONFIG_EXT4_USE_FOR_EXT2=y
 
 # Work with the previously used in-tree version of SMARTPQI instead of AL backport
 # CONFIG_AMAZON_SCSI_SMARTPQI is not set
-CONFIG_SCSI_SMARTPQI=m
```
markusboehme: Dropped this by accident?

packages/kernel-5.15/config-bottlerocket
```diff
@@ -200,3 +200,6 @@ CONFIG_ISCSI_TARGET=m
 # CONFIG_BOOTPARAM_SOFTLOCKUP_PANIC is not set
 # CONFIG_BOOTPARAM_HARDLOCKUP_PANIC is not set
 # CONFIG_BOOTPARAM_HUNG_TASK_PANIC is not set
+
+# Disable edac driver for intel 10nm memory controllers                       │mdio-cavium.c      mdio-moxart.c      mdio-mux-multiplexer.c
+# CONFIG_EDAC_I10NM is not set
```
markusboehme: * Editing artifacts in line 204. 
* We do build EDAC driver modules for other micro architectures. I believe consistency here would result in a better user experience. If this is about minimizing changes for Bottlerocket 1.14.2, this patch is perhaps better placed in the 1.14.x branch (or we open an issue to be consistent for Bottlerocket 1.15.0).

packages/kernel-5.10/config-bottlerocket
```diff
@@ -103,6 +103,13 @@ CONFIG_MOUSE_PS2=m
 
 # Add support for IPMI drivers
 CONFIG_IPMI_HANDLER=m
+# Disable more specialized IPMI drivers that are not relevant for our use-cases
+# CONFIG_IPMI_DEVICE_INTERFACE is not set
+# CONFIG_IPMI_PANIC_EVENT is not set
+# CONFIG_IPMI_POWEROFF is not set
+# CONFIG_IPMI_SI is not set
+# CONFIG_IPMI_WATCHDOG is not set
+# CONFIG_ACPI_IPMI is not set
```
markusboehme: That configuration initially confused me. Nothing is exposed to the user space (e.g. no support for `ipmitool`) and there's no system interface drivers (`CONFIG_IPMI_SI`, `CONFIG_IPMI_SSIF` are deselected) to talk to a BMC, so the IPMI message handler would seem to be quite useless. However, `CONFIG_IPMI_HANDLER=m` was set in #1862 to support NVIDIA variants, so I assume NVIDIA drivers provide a bespoke driver. Would you mind adding that to the comment about adding support for IPMI drivers in case you send another revision? It's fine if not, I can do it later.

markusboehme:
* `CONFIG_IXGBE*`: I think going the opposite direction for achieving consistency makes more sense for Bottlerocket (i.e. disable it for all non-metal variants on both architectures). For Enhanced Networking with the Intel 82599 VF on EC2 there will still be the `ixgbevf` driver which is independent of the host's `ixgbe`.
* `CRYPTO_FIPS_NAME`: If we want to base FIPS certification off Amazon Linux' eventually, it may make sense to keep it. Not sure whether now is the right time. What do you think, @bcressey?

packages/kernel-5.10/config-bottlerocket
foersleo: Oops, I totally messed this one up :(

Originally I had a commit in the series to drop ixgbe as you had pointed out above, but then I felt like dropping the driver on the other arch would potentially incur more issues than adding it to another arch. Looks like this line went under when I did the contextual changes when dropping that one commit.

packages/kernel-5.15/config-bottlerocket
foersleo: Ugh, that additional data on the line... Sorry for that.

Regarding the EDAC driver:

We did not have it enabled even though it was available in the past. AL enabled it. I disabled it to keep consistency with what we had before, given that I am not aware of requests for us to support these. I do not have strong feelings to either way, but agreed probably we want to find a consistent way that works best towards user experience.

packages/kernel-5.10/config-bottlerocket
foersleo: Hmm, maybe the ixgbe driver belongs into the same bucket. Needs some general level clean up, but is setup correctly for metal already.

Possibly something I should put under https://github.com/bottlerocket-os/bottlerocket/issues/2289

foersleo:
Update to latest minor version of AL for 5.10 and 5.15.
Leaving an extra update of 6.1 series out as it incurs a bigger update and 6.1 is currently only in dev variants. Going to do that update independently from this PR.

markusboehme (approved):

stmcginnis (approved):
Things look good to me.

PR #3239, opened 2023-06-28
Add CIS benchmark checks for generating a Kubernetes compliance report

### LINKED ISSUE #2852: Consider adding Kubernetes CIS Benchmark checks
**What I'd like:**

As a follow on to #2731 we could potentially add checks for the CIS Kubernetes Benchmark for the `*-k8s-*` Bottlerocket variants.

Kubernetes benchmarks can still be checked manually or using custom tools by the user. Adding checks to the Bottlerocket CIS agent would just be a convenience.

stmcginnis:
**Issue number:**

Closes #2852 

**Description of changes:**

This adds a set of CIS benchmark checks to `bloodhound` for the Kubernetes CIS benchmark.

Still needs exposure through `apiserver` and `apiclient`. The plumbing for CIS reports in general are still being worked on for the CIS Bottlerocket benchmark. Depending on timing, the wiring up of this benchmark will either be added to this PR in additional commits, or added as a follow up PR after this one merges.

**Testing done:**

Ran new checks on `aws-k8s-1.26` node:

```
[ssm-user@control]$ apiclient report cis-k8s
Benchmark name:  CIS Kubernetes Benchmark (Worker Node)
Version:         v1.7.1
Reference:       https://www.cisecurity.org/benchmark/kubernetes
Benchmark level: 1
Start time:      2023-07-18T15:09:51.451720655Z

[PASS] 4.1.1     Ensure that the kubelet service file permissions are set to 644 or more restrictive (Automatic)
[PASS] 4.1.2     Ensure that the kubelet service file ownership is set to root:root (Automatic)
[SKIP] 4.1.3     If proxy kubeconfig file exists ensure permissions are set to 644 or more restrictive (Manual)
[SKIP] 4.1.4     If proxy kubeconfig file exists ensure ownership is set to root:root (Manual)
[PASS] 4.1.5     Ensure that the --kubeconfig kubelet.conf file permissions are set to 644 or more restrictive (Automatic)
[PASS] 4.1.6     Ensure that the --kubeconfig kubelet.conf file ownership is set to root:root (Automatic)
[PASS] 4.1.7     Ensure that the certificate authorities file permissions are set to 600 or more restrictive (Automatic)
[PASS] 4.1.8     Ensure that the client certificate authorities file ownership is set to root:root (Automatic)
[PASS] 4.1.9     If the kubelet config.yaml configuration file is being used validate permissions set to 600 or more restrictive (Automatic)
[PASS] 4.1.10    If the kubelet config.yaml configuration file is being used validate file ownership is set to root:root (Automatic)
[PASS] 4.2.1     Ensure that the --anonymous-auth argument is set to false (Automatic)
[PASS] 4.2.2     Ensure that the --authorization-mode argument is not set to AlwaysAllow (Automatic)
[PASS] 4.2.3     Ensure that the --client-ca-file argument is set as appropriate (Automatic)
[PASS] 4.2.4     Verify that the --read-only-port argument is set to 0 (Automatic)
[PASS] 4.2.5     Ensure that the --streaming-connection-idle-timeout argument is not set to 0 (Automatic)
[PASS] 4.2.6     Ensure that the --make-iptables-util-chains argument is set to true (Automatic)
[SKIP] 4.2.7     Ensure that the --hostname-override argument is not set (not valid for Bottlerocket) (Manual)
[PASS] 4.2.9     Ensure that the --tls-cert-file and --tls-private-key-file arguments are set as appropriate (Automatic)
[SKIP] 4.2.10    Ensure that the --rotate-certificates argument is not set to false (not valid for Bottlerocket) (Manual)
[PASS] 4.2.11    Verify that the RotateKubeletServerCertificate argument is set to true (Automatic)
[PASS] 4.2.12    Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers (Automatic)
[PASS] 4.2.13    Ensure that a limit is set on pod PIDs (Automatic)

Passed:          18
Failed:          0
Skipped:         4
Total checks:    22

Compliance check result: PASS
```
```
[ssm-user@control]$ apiclient report cis-k8s -l 2
Benchmark name:  CIS Kubernetes Benchmark (Worker Node)
Version:         v1.7.1
Reference:       https://www.cisecurity.org/benchmark/kubernetes
Benchmark level: 2
Start time:      2023-07-18T15:10:15.191556058Z

[PASS] 4.1.1     Ensure that the kubelet service file permissions are set to 644 or more restrictive (Automatic)
...
```

```
[ssm-user@control]$ apiclient report cis-k8sh
Missing or unknown subcommand for 'report'
...
```

```json
apiclient report cis-k8s -l 2 -f json
{
  "level": 2,
  "total": 23,
  "passed": 18,
  "skipped": 5,
  "failed": 0,
  "status": "PASS",
  "timestamp": "2023-07-18T15:10:32.352887194Z",
  "name": "CIS Kubernetes Benchmark (Worker Node)",
  "version": "v1.7.1",
  "url": "https://www.cisecurity.org/benchmark/kubernetes",
  "results": {
    "k8s04010100": {
      "name": "k8s04010100",
      "id": "4.1.1",
      "level": 1,
      "title": "Ensure that the kubelet service file permissions are set to 644 or more restrictive",
      "mode": "Automatic",
      "status": "PASS",
      "error": ""
    },
   ...
}
```

**Terms of contribution:**


maxres-ch:
This is great to see! I wrote a daemonset to run the bash script outlined here: https://github.com/aws-samples/containers-blog-maelstrom/tree/main/cis-bottlerocket-benchmark-eks/

Will there be an operator to accompany this or will it be up to users to invoke this client and report the status individually?

stmcginnis:
> Will there be an operator to accompany this or will it be up to users to invoke this client and report the status individually?

Hey! Great questions. The plan is for the Bottlerocket and Kubernetes benchmarks to ultimately be exposed via `apiclient`. That work is still in progress, with the `apiserver` updates happening [here](https://github.com/bottlerocket-os/bottlerocket/pull/3219), and the last step of wiring in to `apiclient` coming once that is done.

So while it's possible to execute this command directly to get the output, the actual end user experience should be a little cleaner through `apiclient report cis-k8s` (or something like that).

sources/bloodhound/src/bin/kubernetes-checks/main.rs
```diff
@@ -0,0 +1,54 @@
...
+            title: "If proxy kubeconfig file exists ensure ownership is set to root:root".to_string(),
+            id: "4.1.4".to_string(),
+            level: 1,
+        }),
+        "k8s04020800" => Box::new(ManualChecker {
+            name: cmd_name.to_string(),
+            title: "Ensure that the --hostname-override argument is not set".to_string(),
+            id: "4.2.8".to_string(),
+            level: 1,
+        }),
```
bcressey: For these manual checks, is the logic that we know that they'll potentially fail on Bottlerocket, so we turn them into "SKIP" tests? E.g. we always set `hostname-override` now to placate the external AWS cloud provider.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -0,0 +1,27 @@
...
+// Bottlerocket doesn't use the standard path for most of these files ¯\_(ツ)_/¯
+const KUBELET_SERVICE_FILE: &str = "/etc/systemd/system/kubelet.service.d/exec-start.conf";
+
+// =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<= =>o.o<=
+
+pub struct K8S04010100Checker {}
+
+impl Checker for K8S04010100Checker {
+    fn execute(&self) -> CheckerResult {
+        check_file_not_mode(KUBELET_SERVICE_FILE, 0b111111)
```
bcressey: It looks like this ends up only checking `0o77` rather than `0o177`, so we wouldn't detect a `744` file.

It might be better to compose it out of libc constants, e.g.
```
use libc::{S_IRUSR, S_IWUSR};

let u_rw_g_o_ = S_IRUSR | S_IWUSR;
let mask = !u_rw_g_o_ & 0o777;
```

sources/bloodhound/src/lib.rs
```diff
@@ -151,6 +153,24 @@ pub fn check_file_not_mode(file_path: &str, mode: u32) -> CheckerResult {
     result
 }
 
+/// Verifies the file at the given path is owned by root:root, returning a `Results` based on the results.
+pub fn ensure_file_owner_and_group_root(file_path: &str) -> CheckerResult {
```
bcressey, line 163: nit: why not take a desired uid and gid? That seems more consistent with the mode checker which takes a mask rather than hard-coding the specific check.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -0,0 +1,714 @@
...
+    }
+
+    fn metadata(&self) -> CheckerMetadata {
+        CheckerMetadata {
+            title: "Ensure that the kubelet service file permissions are set to 644 or more restrictive".to_string(),
```
bcressey, line 28: The v1.7.0 benchmark seems to use this for all permissions:
```suggestion
            title: "Ensure that the kubelet service file permissions are set to 600 or more restrictive".to_string(),
```

sources/Cargo.lock
```diff
@@ -4297,6 +4333,12 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e1766d682d402817b5ac4490b3c3002d91dfa0d22812f341609f97b08757359c"
 
+[[package]]
+name = "unsafe-libyaml"
+version = "0.2.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1865806a559042e51ab5414598446a5871b561d21b6764f2eabb0dd481d880a6"
```
bcressey, line 4371: I'm not very excited about adding libyaml to the distro given its security track record. Do we have other options?

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -228,3 +228,51 @@ impl Checker for K8S04020100Checker {
...
+    fn execute(&self) -> CheckerResult {
+        #[derive(Deserialize)]
+        struct Authorization {
+            mode: String,
+        }
+
+        #[derive(Deserialize)]
+        struct KubeletConfig {
+            authorization: Authorization,
+        }
```
bcressey, line 250: I don't mind this very much, but there'd be less duplicated code to have a separate `KubeletConfig` struct with all the values for all the checks. Maybe a final refactor commit on top of these individual ones to pull it out.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -419,3 +419,46 @@ impl Checker for K8S04020500Checker {
...
+
+        result
+    }
+
+    fn metadata(&self) -> CheckerMetadata {
+        CheckerMetadata {
+            title: "Ensure that the --protect-kernel-defaults argument is set to true".to_string(),
+            id: "4.2.6".to_string(),
+            level: 1,
+            name: "k8s04020600".to_string(),
```
bcressey: In the 1.7.0 benchmark, the `--protect-kernel-defaults` recommendation no longer exists, and the numbering of subsequent recommendations shifts upward.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -555,3 +555,47 @@ impl Checker for K8S04021000Checker {
...
+            if let Ok(config) = serde_yaml::from_reader::<_, KubeletConfig>(kubelet_file) {
+                if !config.rotate_certificates {
+                    result.error = "Kubelet rotateCertificates is disabled".to_string();
+                    result.status = CheckStatus::FAIL;
+                } else {
+                    result.status = CheckStatus::PASS;
+                }
+            } else {
+                // Normally this value should not be present in the config file, so deserialization is expected to fail.
+                result.status = CheckStatus::PASS;
```
bcressey: `rotateCertificates` defaults to false (disabled) - so if it's not present in the config then it's a failed check.

This seems like the hostname override recommendation where it doesn't really apply to EKS worker nodes - but might apply for other cloud providers.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -649,3 +649,66 @@ impl Checker for K8S04021200Checker {
...
+        if let Ok(kubelet_file) = File::open(KUBELET_CONF_FILE) {
+            if let Ok(config) = serde_yaml::from_reader::<_, KubeletConfig>(kubelet_file) {
+                let mut allowed = true;
+                for suite in config.tls_cipher_suites {
+                    if !allowed_suites.iter().any(|&i| i == suite) {
+                        allowed = false;
+                        result.error = format!("'{}' not an allowed cipher suite", suite);
+                        break;
+                    }
+                }
```
bcressey: the lists are small, but this would avoid the O(nm) algorithm:
```
let configured_suites = HashSet::from(config.tls_cipher_suites);
if !configured_suites.is_subset(allowed_suites) {
    result.error = format!("Found disallowed cipher suites");
}
```

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -0,0 +1,714 @@
...
+    fn metadata(&self) -> CheckerMetadata {
+        CheckerMetadata {
+            title: "Ensure that the Kubelet only makes use of Strong Cryptographic Ciphers"
+                .to_string(),
+            id: "4.2.13".to_string(),
+            level: 1,
+            name: "k8s04021300".to_string(),
+            mode: Mode::Automatic,
+        }
+    }
```
bcressey: We are missing the current `4.2.13` check - "Ensure that a limit is set on pod PIDs (Manual)".

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: This was changed in the 1.7.1 benchmark.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: Well that's... unfortunate.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: Double checked documentation. There is a page that states `kubeadm` sets this to true by default. That mis-led me. Verified in the `kubelet` config docs this is indeed `false` by default. Good catch!

sources/bloodhound/src/bin/kubernetes-checks/main.rs
stmcginnis: Hmm... for the most part, these manual checks are supposed to be things where the benchmark defines them as manual and/or we do not have a clear way of verifying things in an automated way. So it's an indication that the user will have to perform the validation themselves to make sure it is correct for their given environment and needs.

For this particular one, I can't recall now why it is manual. We have a mismatch here where the K8s CIS benchmark says this should not be overridden, but in Bottlerocket's case we have unique constraints where it absolutely _should_ be overridden.

I see a few ways we could handle this:

1. Add something like "(not applicable for Bottlerocket)" to the title to show it can be ignored.
2. Just have the checker return `PASS`. Seems like lying, but we know in our case it is OK to be doing it this way.
3. Exclude completely from the set of checks. Could lead to questions of why the built in checkers are incomplete and cause some doubt about trustworthiness of our results, but would side step the fact that it does not apply to Bottlerocket.

Any preferences or other ideas?

sources/bloodhound/src/lib.rs
stmcginnis: Mainly because all we ever care about with the benchmarks is root ownership, and there wouldn't ever be a need to check any other arbitrary uid/gid. So this keeps it simple and doesn't add extra parameters that would never be used.

If you feel strongly about enabling that ability, let me know and I can change it. Or we can also wait if/until it ever comes up that we have multiple uid/gid's to deal with on Bottlerocket.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: Done!

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: Yeah... the tricky part here is that we actual do expect some of these to fail to deserialize in most cases. So that makes it a challenge to have one unifying struct defined. A missing property only relevant to one check would cause a deserialization error for another check.

We could do some custom deserialization and use things like `Option<x>`, but then that does add extra complexity where the result would probably end up being more code rather than less. It's a little messier, but I do like how now it is very localized and clear what is being looked at in each case.

But if you feel strongly about it, let me know and I can spend some time looking at how to handle a single definition to use for all `KubeletConfig` settings.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: Updated - that does make it a little more explicitly clear what we are looking for. 👍

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: Added! It's in the benchmark as a manual check, but easy enough for us to validation automatically for Bottlerocket. 

Note: this check will fail until https://github.com/bottlerocket-os/bottlerocket/pull/3263 is merged.

sources/Cargo.lock
stmcginnis: Not that I have found, but I will continue to look in to any alternatives. Really don't want to manually parse the yaml config file, but that's always an option too if this becomes problematic.

stmcginnis:
Added `apiserver` and `apiclient` parts of the complete implementation and rebased on `develop` to resolve conflicts.

sources/Cargo.lock
stmcginnis: The only other reasonable yaml library I could find was `yaml_rust`, but that appears to be no longer maintained.

sources/api/apiclient/README.md
```diff
@@ -193,13 +193,9 @@ For example, if you want the name "FOO", you can `PATCH` to `/settings?tx=FOO` a
...
 
-#### Bottlerocket CIS Benchmark report
+#### CIS Benchmark Reports
 
-This command can be used to evaluate the current system state and settings for compliance with the [Bottlerocket CIS Benchmark].
-
-```shell
-apiclient report cis
-```
+These set of reports can be used to evaluate the current system state and settings for compliance with the the [Center for Internet Security (CIS) Benchmarks].
```
arnaldo2792:
```suggestion
These set of reports can be used to evaluate the current system state and settings for compliance with the [Center for Internet Security (CIS) Benchmarks].
```

sources/api/apiclient/README.md
```diff
@@ -211,31 +207,68 @@ By default, Bottlerocket instances meet level 1 compliance unless specific setti
 The "Level 2" compliance are things that provide more defense in depth and include settings that may be specific to the deployed environment and usage.
 While these settings are considered more secure, there may be a trade off with inhibiting utility or performance.
 
+**Format**
+
+The default format of the CIS reports are a human readable text report.
```
arnaldo2792:
```suggestion
The default format of the CIS reports is a human readable text report.
```

sources/api/apiclient/README.tpl
```diff
@@ -193,13 +193,9 @@ For example, if you want the name "FOO", you can `PATCH` to `/settings?tx=FOO` a
...
 
-#### Bottlerocket CIS Benchmark report
+#### CIS Benchmark Reports
 
-This command can be used to evaluate the current system state and settings for compliance with the [Bottlerocket CIS Benchmark].
-
-```shell
-apiclient report cis
-```
+These set of reports can be used to evaluate the current system state and settings for compliance with the the [Center for Internet Security (CIS) Benchmarks].
```
arnaldo2792:
```suggestion
These set of reports can be used to evaluate the current system state and settings for compliance with the [Center for Internet Security (CIS) Benchmarks].
```

sources/api/apiclient/src/main.rs
```diff
@@ -151,6 +152,7 @@ fn usage() -> ! {
             reboot                     Reboots the host.
             exec                       Execute a command in a host container.
             report cis                 Retrieve a Bottlerocket CIS benchmark compliance report.
+            report cis-k8s             Retrieve a Kubernetes CIS benchmark compliance report.
```
arnaldo2792, line 155: Shouldn't this command be “conditionality”-compiled based on the variant?

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -0,0 +1,724 @@
...
+
+impl Checker for K8S04010900Checker {
+    fn execute(&self) -> CheckerResult {
+        let no_x_xwr_xwr = S_IXUSR | S_IRWXG | S_IRWXO;
+        check_file_not_mode(KUBELET_CONF_FILE, no_x_xwr_xwr)
+    }
+
+    fn metadata(&self) -> CheckerMetadata {
+        CheckerMetadata {
+            title: "If the kubelet config.yaml configuration file is being used validate permissions set to 600 or more restrictive".to_string(),
```
arnaldo2792, line 154: It seems like all the `kubelet` systemd services do pass `--config <path>`, so could this be?:

```suggestion
            title: "Ensure --config config.yaml file permissions are set to 600 or more restrictive".to_string(),
```

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -0,0 +1,724 @@
...
+pub struct K8S04011000Checker {}
+
+impl Checker for K8S04011000Checker {
+    fn execute(&self) -> CheckerResult {
+        ensure_file_owner_and_group_root(KUBELET_CONF_FILE)
+    }
+
+    fn metadata(&self) -> CheckerMetadata {
+        CheckerMetadata {
+            title: "If the kubelet config.yaml configuration file is being used validate file ownership is set to root:root"
```
arnaldo2792, line 174: Same as in check `K8S04010900Checker`

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -0,0 +1,724 @@
...
+
+impl Checker for K8S04010100Checker {
+    fn execute(&self) -> CheckerResult {
+        let no_x_xw_xw = S_IXUSR | S_IXGRP | S_IWGRP | S_IXOTH | S_IWOTH;
+        check_file_not_mode(KUBELET_SERVICE_FILE, no_x_xw_xw)
+    }
+
+    fn metadata(&self) -> CheckerMetadata {
+        CheckerMetadata {
+            title: "Ensure that the kubelet service file permissions are set to 644 or more restrictive".to_string(),
```
arnaldo2792, line 28: Is the `title` intended to be use as an instruction for the user? Or as a description for the checker? If the latter, then shouldn't `Ensure` be `Ensures`? Otherwise, it sounds like we are instructing the users they should set the correct permissions.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -0,0 +1,724 @@
...
+
+        if let Ok(kubelet_file) = File::open(KUBELET_CONF_FILE) {
+            if let Ok(config) = serde_yaml::from_reader::<_, KubeletConfig>(kubelet_file) {
+                if (!config.tls_cert_file.is_empty() && Path::new(&config.tls_cert_file).exists())
+                    && (!config.tls_private_key_file.is_empty()
+                        && Path::new(&config.tls_private_key_file).exists())
+                {
+                    result.status = CheckStatus::PASS;
+                } else {
+                    result.error = "TLS files not set to expected path".to_string();
```
arnaldo2792, line 496: Should the expected path be included in this error?

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
```diff
@@ -0,0 +1,724 @@
...
+
+        if let Ok(kubelet_file) = File::open(KUBELET_CONF_FILE) {
+            if let Ok(config) = serde_yaml::from_reader::<_, KubeletConfig>(kubelet_file) {
+                let configured_suites: HashSet<&str> = config
+                    .tls_cipher_suites
+                    .iter()
+                    .map(|s| s.as_str())
+                    .collect();
+                if !configured_suites.is_subset(&allowed_suites) {
+                    result.error = "Found disallowed cipher suites".to_string();
```
arnaldo2792, line 654: Should the culprits be included in this error?

sources/api/apiclient/README.md
stmcginnis: Oh right, singular "default format". ;)

sources/api/apiclient/src/main.rs
stmcginnis: I did think about that and decided to just include it. If someone were to try to run the report on an `aws-ecs-*` instance, it would just report back SKIPPED for every assertion.

Conditionally including it makes things slightly more complicated with needing conditional inclusion, especially here with things like documentation. It also changes this CLI "api" by having some commands there on one instance, but not there on others.

It could be done, but I opted to try to keep things as simple as possible. Willing to hear strong opinions though.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: Neither? The title is just the title of the check from the benchmark.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: See above, this is the title of the check.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: It is dependent on what is in the config file, which may not contain anything. So this is a general message to point at where the issue is, and the benchmark contains the steps necessary to manually verify if these values are not found in the config file where they are expected to be.

sources/bloodhound/src/bin/kubernetes-checks/checks.rs
stmcginnis: This is another case where we would want them to refer to the benchmark to evaluate things. This would need to be something they are doing in a custom build, since we know the official Bottlerocket variants do not include any cipher suites that are not acceptable.

sources/api/apiclient/README.tpl
```diff
@@ -193,13 +193,9 @@ For example, if you want the name "FOO", you can `PATCH` to `/settings?tx=FOO` a
...
 
-#### Bottlerocket CIS Benchmark report
+#### CIS Benchmark Reports
 
-This command can be used to evaluate the current system state and settings for compliance with the [Bottlerocket CIS Benchmark].
-
-```shell
-apiclient report cis
-```
+These set of reports can be used to evaluate the current system state and settings for compliance with the [Center for Internet Security (CIS) Benchmarks].
```
bcressey:
```suggestion
These reports can be used to evaluate the current system state and settings for compliance with the [Center for Internet Security (CIS) Benchmarks].
```

(Or "This set of reports")

sources/bloodhound/src/lib.rs
```diff
@@ -288,4 +340,70 @@ mod test_utils {
...
+        // Generate a temp file that will be dropped as unused
+        let test_file_path = NamedTempFile::new()
+            .unwrap()
+            .into_temp_path()
+            .canonicalize()
+            .unwrap()
+            .display()
+            .to_string();
+
+        // Then us that name to create test file with the expected permission mode
```
bcressey: nit:
```suggestion
        // Then use that name to create test file with the expected permission mode
```

sources/bloodhound/src/lib.rs
```diff
@@ -288,4 +340,70 @@ mod test_utils {
...
+        let result = check_file_not_mode(test_file_path.as_str(), 0b111111);
+        assert_eq!(result.status, CheckStatus::FAIL);
+
+        // Clean up, fine to ignore errors
+        let _ = fs::remove_file(test_file_path);
+    }
+
+    #[test]
+    fn test_check_file_not_mode_file_not_found() {
+        // Generate a temp file that will be dropped as unused
```
bcressey: nit: comment doesn't apply
```suggestion
```

bcressey (approved):

sources/Cargo.lock
bcressey: On closer inspection, `unsafe-libyaml` is a fork of libyaml translated into unsafe Rust.

We can keep an eye on it, but it seems OK for now.

sources/bloodhound/src/bin/kubernetes-checks/main.rs
bcressey: The "(not valid for Bottlerocket)" notes look good to me.

etungsten (approved):

grodriguezl:
This is beautiful. Thanks!

PR #3245, opened 2023-06-30
testsys: Allow not-strict handlebars for clusters

ecpullen:
**Issue number:**

N/A

**Description of changes:**

Some eksa cluster configs contain handlebars templating. In order to allow those configs to be used with testsys we need to remove strict mode.

**Testing done:**

**Terms of contribution:**

### DIFF
--- a/tools/testsys/src/crds.rs
+++ b/tools/testsys/src/crds.rs
@@ -317,8 +317,7 @@ impl<'a> CrdInput<'a> {
         fields.insert("namespace".to_string(), NAMESPACE.to_string());
         fields.append(additional_fields);
 
-        let mut handlebars = Handlebars::new();
-        handlebars.set_strict_mode(true);
+        let handlebars = Handlebars::new();
         let rendered_config = handlebars.render_template(&config, &fields)?;
 
         Ok(Some(rendered_config))


stmcginnis (approved):

cbgbt (commented):
> Some eksa cluster configs contain handlebars templating. In order to allow those configs to be used with testsys we need to remove strict mode.

Can you give an example of one of the problematic cases?

According to the docs, removing strict mode causes unset values to write an empty string. Is that behavior correct in these cases?

webern:
> > Some eksa cluster configs contain handlebars templating. In order to allow those configs to be used with testsys we need to remove strict mode.
> 
> Can you give an example of one of the problematic cases?
> 
> According to the docs, removing strict mode causes unset values to write an empty string. Is that behavior correct in these cases?

+1, perhaps a comment is in order at the site of the handlebars library construction that says why strict mode will break use cases and that we are not using it.

webern (approved):

ecpullen:
After reviewing the docs, we can escape the handlebar strings with:
```
{{{{raw}}}} {{ignored}} {{{{/raw}}}}
```

Closing this pr.

PR #3248, opened 2023-07-05
Actions: add random delay to matrix start

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

For our main "build" GitHub Action we build a matrix execution for each variant supported. This causes many jobs to start at the same time, and could lead to triggering some protections from the external resources that they need to pull from (GitHub, crates.io, etc).

This adds a random delay up to 30 seconds to the individual variant runs so they don't all kick off the same actions at the same time, hopefully spreading some of the load a little better and avoiding some of these backoff protections.

**Testing done:**

To be seen in resulting action runs.

**Terms of contribution:**

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,11 @@ jobs:
       fail-fast: false
     name: "Build ${{ matrix.variant }}-${{ matrix.arch }}"
     steps:
+      - name: Random delay
+        run: |
+          delay=$((1 + $RANDOM % 32))
+          echo "Waiting ${delay} seconds before execution"
+          sleep $delay
       - uses: actions/checkout@v3
       - name: Preflight step to set up the runner
         uses: ./.github/actions/setup-node


.github/workflows/build.yml
```diff
@@ -51,6 +51,11 @@ jobs:
       fail-fast: false
     name: "Build ${{ matrix.variant }}-${{ matrix.arch }}"
     steps:
+      - name: Random delay
+        run: |
+          delay=$((1 + $RANDOM % 30))
```
bcressey: nit: perhaps this to avoid bias:
```suggestion
          delay=$((1 + $RANDOM % 32))
```

bcressey (approved):
🎲

etungsten (approved):

etungsten (approved):
Commit message is no longer accurate about the 30 seconds, but it's inconsequential

Issue #2404, opened 2022-09-08
Allow overrides to the default OCI spec capabilities for k8s pods


arnaldo2792:
**What I'd like:**
With #1703, Bottlerocket will support overrides to RLIMITS in the OCI spec. A follow up feature will be to allow overrides to the OCI spec capabilities through the API.

**Any alternatives you've considered:**
Use k8s directives to override the default capabilities per pod, but this method requires the user to override the capabilities for each pod instead of all of them at once.

mchaker:
Closed in #2697

PR #3259, opened 2023-07-12
Support settings oci defaults on ecs variants

### LINKED ISSUE #2759: support settings.oci-defaults on ecs variants
**What I'd like:**

In #2697 we added settings support for setting OCI defaults in orchestrator-launched pods (see also #1703 #2404). The settings we added for this are:
- `settings.oci-defaults.resource-limits`
- `settings.oci-defaults.capabilities`

We want these settings to be available and work the same way on ECS variants. The ECS variants use Docker, so the implementation for these Bottlerocket settings may be non-trivial.

Here are a couple of relevant links:
- `rlimits` might be simple: https://github.com/moby/moby/blob/dd61302f5bd87c887ec4cba22479a37ef78914c8/daemon/oci_linux.go#L40
- Capabilities is more involved because we would need a patch to read capabilities from a file instead of using this hardcoded list: https://github.com/moby/moby/blob/dd61302f5bd87c887ec4cba22479a37ef78914c8/daemon/oci_linux.go#L40

N/A

vyaghras:
**Issue number:** 

Closes #2759

**Description of changes:**
Currently in bottlerocket we have support for setting OCI defaults in orchestrator-launched pods using api. (Refer [#1703](https://github.com/bottlerocket-os/bottlerocket/issues/1703) [#2404](https://github.com/bottlerocket-os/bottlerocket/issues/2404) [#2697](https://github.com/bottlerocket-os/bottlerocket/pull/2697) ).  The settings we added in api are:

* settings.oci-defaults.resource-limits
* settings.oci-defaults.capabilities

But currently in bottlerocket these setting do not work for ECS variants in same way as EKS variants because ECS variants uses docker.([Moby package](https://github.com/moby/moby/tree/6b7974cf1662d56a9aadc2b5c2ad12a80b448671))
This PR will enable the support for these setting for ecs variants.

The priority for these settings are daemon.json => Hostconfig, meaning if any configuration change has been done in HostConfig for Capabilities and Resource limits, it should get preference over the setting done by daemon.json.

**Testing done:**
- [x] Set a capability true using api and create container using docker, that capability should be added in that container. 
- [x] Change the soft limit and hard limit of already set rlimit and create a container, the limit should be the changed one.
- [x] Set the defaults limits using daemon.json file, Create the container, it should use the one set by default ulimit. 
- [x] Change something in host config and create a container, that container should have the requested settings from host config.
- [x] Migration testing to and from 1.15.0
- [x] If nothing set in host config for rlimit then it should use the set in daemon.json using api client
- [x] Perform above testing for ecs-1-nvidia instance.
- [x] Task definition should override the daemon config ulimits

**Terms of contribution:**


packages/docker-engine/0001-read-default-oci-spec-from-file.patch
```diff
@@ -0,0 +1,170 @@
...
++			defCaps,
+ 			c.HostConfig.CapAdd,
+ 			c.HostConfig.CapDrop,
+ 			c.HostConfig.Privileged,
+@@ -171,6 +182,19 @@ func WithCapabilities(c *container.Container) coci.SpecOpts {
+ 	}
+ }
+
++func getUniqueCapabilities(defCaps []string) []string {
++	occurred := map[string]bool{}
```
yeazelm: nit: I've often used `seen` or `exists` for this variable when using it for deduplication since it reads nicely as "if seen, then..."

stmcginnis: A potential optimization since you know there will never be more elements than what is in `defCaps` would be to make this:

```go
occurred := make(map[string]bool, len(defCaps))
```

packages/docker-engine/0001-read-default-oci-spec-from-file.patch
```diff
@@ -0,0 +1,170 @@
...
++	var uniqueCaps []string
++	for c := range defCaps {
++		if occurred[defCaps[c]] != true {
```
stmcginnis: It would be cleaner to do:

```go
if !occurred[defCaps[c]] {
```

packages/docker-engine/0001-read-default-oci-spec-from-file.patch
```diff
@@ -0,0 +1,170 @@
...
+ 	if err != nil {
+@@ -1014,6 +1038,23 @@ func (daemon *Daemon) createSpec(c *container.Container) (retSpec *specs.Spec, e
+ 		opts []coci.SpecOpts
+ 		s    = oci.DefaultSpec()
+ 	)
++
++	if daemon.configStore.BaseRuntimeSpec != "" {
++		baseSpec, errReadfile := os.ReadFile(daemon.configStore.BaseRuntimeSpec)
++		if errReadfile != nil {
++			fmt.Print("Can't find base OCI spec %q", daemon.configStore.BaseRuntimeSpec)
```
stmcginnis: Need to be able to break out from here if there was an error. Otherwise this just prints the message and continues on where it tries to use `baseSpec` on line 130. Though I suppose that will just fail there, so technically this does work.

Release.toml
```diff
@@ -215,4 +215,7 @@ version = "1.15.0"
 "(1.14.1, 1.14.2)" = [
     "migrate_v1.14.2_ecs-images-cleanup.lz4",
 ]
-"(1.14.2, 1.15.0)" = []
+"(1.14.2, 1.15.0)" = [
+"migrate_v1.15.0_oci-defaults-docker-setting.lz4",
```
stmcginnis: Please indent these array values.

packages/docker-engine/daemon-json
```diff
@@ -8,6 +8,7 @@
   "default-runtime": "shimpei",
   "runtimes": { "shimpei": { "path": "shimpei" } },
   "selinux-enabled": true,
+   "base-runtime-spec": "/etc/docker/cri-base.json",
```
stmcginnis: Remove leading space on this line.

packages/docker-engine/daemon-nvidia-json
```diff
@@ -8,6 +8,7 @@
   "default-runtime": "shimpei",
   "runtimes": { "shimpei": { "path": "shimpei" }, "nvidia": { "path": "nvidia-oci" } },
   "selinux-enabled": true,
+   "base-runtime-spec": "/etc/docker/cri-base.json",
```
stmcginnis: Leading space.

stmcginnis (changes requested):
A few minor things, but looks pretty good overall.

packages/docker-engine/0001-read-default-oci-spec-from-file.patch
```diff
@@ -0,0 +1,170 @@
...
++		}
```
yeazelm: This function returns an error, you should not print but instead return the error so the daemon can properly handle this situation.

packages/docker-engine/0001-read-default-oci-spec-from-file.patch
```diff
@@ -0,0 +1,170 @@
...
++
++		var spec specs.Spec
++
++		errUnmarshall := json.Unmarshal(baseSpec, &spec)
++		if errUnmarshall != nil {
++			fmt.Print("Can not unmarshal the Spec file %q", errUnmarshall)
++		}
```
yeazelm: Same as above, you need to be returning an error here and not printing.

packages/docker-engine/0001-read-default-oci-spec-from-file.patch
```diff
@@ -0,0 +1,170 @@
...
++
+ func resourcePath(c *container.Container, getPath func() (string, error)) (string, error) {
+ 	p, err := getPath()
+ 	if err != nil {
+@@ -1014,6 +1038,23 @@ func (daemon *Daemon) createSpec(c *container.Container) (retSpec *specs.Spec, e
+ 		opts []coci.SpecOpts
+ 		s    = oci.DefaultSpec()
+ 	)
++
++	if daemon.configStore.BaseRuntimeSpec != "" {
```
yeazelm: You are essentially changing default behavior to overwrite the function `oci.DefaultSpec()`. It's probably a good idea to call out what you expect to happen here with a comment so future work around this can attempt to keep the same functionality if refactored.

packages/docker-engine/0001-read-default-oci-spec-from-file.patch
```diff
@@ -0,0 +1,170 @@
...
++
+ 	return &s, coci.ApplyOpts(context.Background(), nil, &containers.Container{
+ 		ID: c.ID,
+ 	}, &s, opts...)
+@@ -1075,3 +1117,18 @@ func (daemon *Daemon) mergeUlimits(c *containertypes.HostConfig) {
+ 	}
+ 	c.Ulimits = ulimits
+ }
++
++// mergeBaseSpecUlimits merge the Ulimits from Base Spec file in Daemon and update daemon defaults
```
yeazelm:
```suggestion
+// mergeBaseSpecUlimits merge the Ulimits from Base Spec file and update daemon defaults
```
This stays consistent with the other comments in the file.

sources/models/shared-defaults/oci-defaults-docker-cri-resource-limits.toml
```diff
@@ -0,0 +1,3 @@
+[settings.oci-defaults.resource-limits.max-open-files]
+hard-limit = 4086
+soft-limit = 1024
```
yeazelm: These are what the current settings are right? Do we want to bring them in line with the defaults for our other variants?

yeazelm (commented):
I'd like to clean up the commit messages and order a bit as well.

packages/docker-engine/0001-read-default-oci-spec-from-file.patch
vyaghras: Ohh yeah good point, or may be break from if loop and use default spec so atleast it will not break and create the containers, it could be a little hard to find later what went wrong that it does not shows the settings in Base-spec-file. So correct if we are ok for error we can leave it as is, else break if and use default.

sources/models/shared-defaults/oci-defaults-docker-cri-resource-limits.toml
vyaghras: Yeah these are as per the current settings for ecs variants. Ok I will update them same as k8s variants.

packages/docker-engine/0001-Add-ability-to-read-default-OCI-Spec-from-a-file.patch
```diff
@@ -0,0 +1,171 @@
...
+ 		opts []coci.SpecOpts
+ 		s    = oci.DefaultSpec()
+ 	)
++
++	// Change the default spec as per base-spec-file in Daemon.json
++	if daemon.configStore.BaseRuntimeSpec != "" {
++		baseSpec, errReadfile := os.ReadFile(daemon.configStore.BaseRuntimeSpec)
++		// Fail when unable to read the base spec file
++		if errReadfile != nil {
++			return nil ,fmt.Errorf("unable to read the base spec file provided in Daemon: %v", errReadfile)
```
stmcginnis: s/nil ,fmt/nil, fmt/

stmcginnis (approved):

packages/docker-engine/0001-Add-ability-to-read-default-OCI-Spec-from-a-file.patch
```diff
@@ -0,0 +1,212 @@
+From b8e869f73b176eb80f9f5e3a8a137b679af0f10b Mon Sep 17 00:00:00 2001
+From: Shikha Vyaghra <vyaghras@amazon.com>
+Date: Thu, 6 Jul 2023 17:26:45 +0000
+Subject: [PATCH] Add ability to read default OCI Spec from a file
+
+Add the capability to read the default spec from a file. If a file is not provided, the
+default spec embedded in code is used. There is a parameter "base-spec-file" in
+daemon configuration to provide the default spec file path. Starting containers is
+expected to fail when the daemon is unable to read or parse the requested base-spec-file
+in the daemon config.
```
arnaldo2792: You need to update this to mention that each runtime could get its own base-spec-file.

packages/docker-engine/0001-Add-ability-to-read-default-OCI-Spec-from-a-file.patch
```diff
@@ -0,0 +1,212 @@
...
+ 	"github.com/moby/sys/mount"
+ 	"github.com/moby/sys/mountinfo"
+ 	"github.com/opencontainers/runc/libcontainer/apparmor"
+@@ -45,8 +47,13 @@ func WithRlimits(daemon *Daemon, c *container.Container) coci.SpecOpts {
+ 
+ 		// We want to leave the original HostConfig alone so make a copy here
+ 		hostConfig := *c.HostConfig
++
++		//Merge defaults in Base Spec file with Daemon defaults
++		daemon.mergeBaseSpecUlimits(s)
```
arnaldo2792: ~~I think what we want is to use the default spec values, only when both the daemon and the `HostConfig` for the container are empty. So, you can skip this method, and let `daemon.mergeUlimits` do its thing. Then, if `hostConfig.Ulimits` is empty, `rlimits` will be. You can check if `len(rlimits) == 0`, it so don't override the spec. Otherwise, override the entire value.~~

I misread what contained does. So, I think what should happen is `daemon.mergeUlimits()` remains the same, no changes to it. Then, after `rlmits` is created check if `s.Procces.Rlimits` was already set. If it is, then "upsert" the values in `rlimits`, otherwise, just set it to `rlimits`.

This will be closer to what containerd does [here] to override parts of the OCI spec. In the linked example, if no capabilities were set, then the default value of the spec will be used. ~~Otherwise, the default value will be overridden with what was configured for the container.~~ Otherwise, the existing value will be extended with what was configured for the container.

The changes that you have for the capabilities overrides should follow the same pattern.

FYI, I found that `daemon.mergeUlimits` is used in [this file]. The docker developers specified in the GH issue that I created that `docker info` could show the configurations used for the container. As you can see in the linked file, the OCI spec wasn't used to construct `HostConfig`. You may have to use the base spec file for the runtime configured for the container to show the correct configurations.

[here]: https://github.com/containerd/containerd/blob/4d2c8879908285454a4006534cb0af82bb58a406/pkg/cri/opts/spec_opts.go#L171

[this file]: https://github.com/moby/moby/blob/3056208812eb5e792fa99736c9167d1e10f4ab49/daemon/inspect.go#L138

packages/docker-engine/0001-Add-ability-to-read-default-OCI-Spec-from-a-file.patch
```diff
@@ -0,0 +1,212 @@
...
++	units "github.com/docker/go-units"
+ 	"github.com/moby/sys/mount"
+ 	"github.com/moby/sys/mountinfo"
+ 	"github.com/opencontainers/runc/libcontainer/apparmor"
+@@ -45,8 +47,13 @@ func WithRlimits(daemon *Daemon, c *container.Container) coci.SpecOpts {
+ 
+ 		// We want to leave the original HostConfig alone so make a copy here
+ 		hostConfig := *c.HostConfig
++
++		//Merge defaults in Base Spec file with Daemon defaults
```
bcressey:
```suggestion
+		// Merge defaults in Base Spec file with Daemon defaults
```

packages/docker-engine/0001-Add-ability-to-read-default-OCI-Spec-from-a-file.patch
```diff
@@ -0,0 +1,212 @@
...
++
+ func (daemon *Daemon) createSpec(c *container.Container) (retSpec *specs.Spec, err error) {
+ 	var (
+ 		opts []coci.SpecOpts
+ 		s    = oci.DefaultSpec()
+ 	)
++
++	var filename string
++	runtime := daemon.configStore.GetRuntime(c.HostConfig.Runtime)
++	if runtime.BaseRuntimeSpec != "" {
```
arnaldo2792: I think you should cache the parsed specs in `daemon.configStore`, otherwise every time a container is created the spec will have to be parsed and that adds additional overhead. This will be more aligned to what containerd does.

packages/docker-engine/daemon-json
```diff
@@ -6,9 +6,9 @@
   "exec-opts": ["native.cgroupdriver=cgroupfs"],
   "data-root": "/var/lib/docker",
   "default-runtime": "shimpei",
-  "runtimes": { "shimpei": { "path": "shimpei" } },
+  "runtimes": { "shimpei": { "path": "shimpei", "base-runtime-spec": "/etc/docker/base-spec.json"} },
```
arnaldo2792: I think for our case, it is OK to just specify the `daemon` configuration instead of setting it for each runtime.

sources/models/shared-defaults/oci-defaults-docker-cri.toml
```diff
@@ -0,0 +1,10 @@
+[metadata.settings.oci-defaults]
+affected-services = ["oci-defaults", "docker"]
+
+[services.oci-defaults]
+configuration-files = ["oci-defaults"]
+restart-commands = []
+
+[configuration-files.oci-defaults]
+path = "/etc/docker/base-spec.json"
```
arnaldo2792: In the commit message, you refer to this file as `cri-base.json`. Please adjust.

sources/models/shared-defaults/oci-defaults-docker-cri.toml
```diff
@@ -0,0 +1,10 @@
+[metadata.settings.oci-defaults]
```
arnaldo2792: This file is named `oci-defaults-docker-cri`, but `cri` only applies to Kubernetes. Please drop that part of the name.

arnaldo2792 (changes requested):

packages/docker-engine/0001-Add-ability-to-read-default-OCI-Spec-from-a-file.patch
```diff
@@ -0,0 +1,212 @@
...
+ 		for _, ul := range hostConfig.Ulimits {
+ 			rlimits = append(rlimits, specs.POSIXRlimit{
+ 				Type: "RLIMIT_" + strings.ToUpper(ul.Name),
+@@ -158,8 +165,12 @@ func WithApparmor(c *container.Container) coci.SpecOpts {
+ // WithCapabilities sets the container's capabilties
+ func WithCapabilities(c *container.Container) coci.SpecOpts {
+ 	return func(ctx context.Context, _ coci.Client, _ *containers.Container, s *coci.Spec) error {
++		defCaps := caps.DefaultCapabilities()
++		defCaps = append(defCaps, s.Process.Capabilities.Bounding...)
++		defCaps = getUniqueCapabilities(defCaps)
```
bcressey: This doesn't seem correct to me. It looks like we'll always end up with the union of `caps.DefaultCapabilities()` and `s.Process.Capabilities.Bounding`, meaning that it wouldn't be possible to drop `CAP_NET_RAW` or similar.

I'd expect you'd want to use the capabilities from the spec, if defined, and fall back to `caps.DefaultCapabilities()` otherwise. `caps.TweakCapabilities()` makes some changes as well.

`oci.SetCapabilities()` will unconditionally set bounding / effective / permitted to the same list for root users, and drop effective and permitted for non-root. To avoid surprises, it might make sense to assert the same thing about the incoming spec: either all lists are the same, or only the bounding list is present.

packages/docker-engine/0001-Add-ability-to-read-default-OCI-Spec-from-a-file.patch
```diff
@@ -0,0 +1,212 @@
...
++			ulIdx[rlimit] = &units.Ulimit{
++				Name: rlimit,
++				Hard: int64(ul.Hard),
++				Soft: int64(ul.Soft),
++			}
++		}
++
++	}
++	daemon.configStore.Ulimits = ulIdx
++}
```
bcressey: I'd prefer to do this work in `mergeUlimits` - it's hard for me to follow what's meant to happen here, with two different function calls that both treat `daemon.configStore.Ulimits` as a source of truth.

Comments would be very helpful. I'm not really sure that the logic is correct. Unit tests would help too; I doubt this could go upstream without them because the three-way merge of defaults for both rlimits and capabilities is complicated.

vyaghras:
Change patch to add "default-capabilities" in the daemon.json and update these default-capabilities  and default-ulimits using apiclient settings.

packages/docker-engine/0001-Change-default-capabilities-using-daemon-config.patch
```diff
@@ -0,0 +1,100 @@
...
++++ b/daemon/oci_linux.go
+@@ -156,10 +156,15 @@ func WithApparmor(c *container.Container) coci.SpecOpts {
+ }
+ 
+ // WithCapabilities sets the container's capabilties
+-func WithCapabilities(c *container.Container) coci.SpecOpts {
++func WithCapabilities(daemon *Daemon, c *container.Container) coci.SpecOpts {
+ 	return func(ctx context.Context, _ coci.Client, _ *containers.Container, s *coci.Spec) error {
++		defCaps := caps.DefaultCapabilities()
++		if len(daemon.configStore.Capabilities) != 0 {
```
arnaldo2792: Do you have to check if `daemon.configStore.Capabilities` isn't `nil` as well?

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1393,38 +1393,34 @@ pub fn oci_defaults(
...
-    let oci_defaults_values = get_param(helper, 0)?;
-    // We want the settings path so we know which OCI spec section we are rendering.
-    // e.g. settings.oci-defaults.resource-limits
-    let settings_path = get_param_key_name(helper, 0)?;
-    // Extract the last part of the settings path, which is the OCI spec section we want to render.
-    let oci_spec_section = settings_path
-        .split('.')
-        .last()
-        .expect("did not find (got None for) an oci_spec_section");
+    render_oci_defaults(
```
arnaldo2792: Instead of having a new helper, what if you pass another argument to the `oci_defaults` helper, that is used to determine the “format” of the settings?

```handlebars
{{~oci_defaults <settings> docker~}}
```

By default, `oci_defaults` will render the settings for `containerd`. Then in the code that generates the `result_lines` you can have a trait that is implemented for both Docker and containerd:

```rust
# Somewhere define something similar to this trait
trait OciDefaultsFormatter {
   fn get_capabilities(caps: String);
   fn get_resource_limits(limits: String);
}
```

You can pass an instance of a struct that implements this trait to `oci_spec_capabilities` and `oci_spec_resource_limits`, and call either method as needed.

arnaldo2792 (changes requested):

packages/docker-engine/0001-Change-default-capabilities-using-daemon-config.patch
vyaghras: len(nil) == 0 if the given array will be null.

packages/containerd/containerd-cri-base-json
```diff
@@ -8,12 +8,12 @@
...
         {{~#if settings.oci-defaults.capabilities~}}
         "capabilities": {
-            {{~oci_defaults settings.oci-defaults.capabilities~}}
+            {{~oci_defaults settings.oci-defaults.capabilities "containerd"~}}
         },
         {{~/if~}}
         {{~#if settings.oci-defaults.resource-limits~}}
         "rlimits": [
-            {{~oci_defaults settings.oci-defaults.resource-limits~}}
+            {{~oci_defaults settings.oci-defaults.resource-limits "containerd"~}}
```
bcressey: nit: I would probably reverse the argument order, so it's immediately clear what "mode" the helper is called with

sources/api/schnauzer/src/helpers.rs
```diff
@@ -307,6 +307,9 @@ mod error {
             number: usize,
             source: std::num::TryFromIntError,
         },
+
+        #[snafu(display("Variant {} is invalid, use docker or k8s", var))]
```
bcressey:
```suggestion
        #[snafu(display("Invalid output type '{}', expected 'docker' or 'containerd'", var))]
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -307,6 +307,9 @@ mod error {
...
+        InvalidVariant { var: String },
```
bcressey: I would avoid using the term "Variant" since it's overloaded in the Bottlerocket context 😀 
```suggestion
        InvalidOutputType { var: String },
```

sources/models/shared-defaults/oci-defaults-docker-capabilities.toml
```diff
@@ -0,0 +1,17 @@
...
+fsetid = true
+kill = true
+mknod = true
+net-bind-service = true
+net-raw = true
+setgid = true
+setfcap = true
+setpcap = true
+setuid = true
+sys-chroot = true
```
bcressey: If this is identical to `oci-defaults-containerd-cri-capabilities.toml`, we should just rename that one and have one source of truth.

sources/models/shared-defaults/oci-defaults-docker.toml
```diff
@@ -0,0 +1,10 @@
+[metadata.settings.oci-defaults]
+affected-services = ["oci-defaults", "docker"]
+
+[services.oci-defaults]
+configuration-files = ["oci-defaults"]
+restart-commands = []
+
+[configuration-files.oci-defaults]
+path = "/etc/docker/daemon.json"
+template-path = "/usr/share/templates/docker-daemon-json"
```
bcressey: The docker service already owns this template; we just need to restart it. Otherwise we'd render the same template twice.

```suggestion
[metadata.settings.oci-defaults]
affected-services = ["docker"]
```

sources/models/src/aws-ecs-1/mod.rs
```diff
@@ -6,7 +6,7 @@ use crate::modeled_types::Identifier;
 use crate::{
     AutoScalingSettings, AwsSettings, BootstrapContainer, CloudFormationSettings, DnsSettings,
     ECSSettings, HostContainer, KernelSettings, MetricsSettings, NetworkSettings, NtpSettings,
-    OciHooks, PemCertificate, RegistrySettings, UpdatesSettings,
+    OciDefaults, OciHooks, PemCertificate, RegistrySettings, UpdatesSettings,
 };
```
bcressey, line 10: This PR should get rebased onto develop so we can add `aws-ecs-2` and `aws-ecs-2-nvidia` also.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1509,6 +1559,18 @@ fn oci_spec_resource_limits(value: &Value) -> Result<String, RenderError> {
...
+    match runtime {
+        "docker" => Ok(Runtime::Docker),
+        "containerd" => Ok(Runtime::Containerd),
+        _ => Err(RenderError::from(
+            error::TemplateHelperError::InvalidVariant {
+                var: runtime.to_owned(),
+            },
+        )),
+    }
+}
```
bcressey: You should be able to deserialize directly from the input into this type; `serde_plain` can help.

arnaldo2792 (approved):

packages/docker-engine/0001-Change-default-capabilities-using-daemon-config.patch
vyaghras: This was resolved offline.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1416,8 +1488,8 @@ pub fn oci_defaults(
     let section =
         OciSpecSection::from_str(oci_spec_section).context(error::InvalidOciSpecSectionSnafu)?;
     let result_lines = match section {
-        OciSpecSection::Capabilities => oci_spec_capabilities(oci_defaults_values)?,
-        OciSpecSection::ResourceLimits => oci_spec_resource_limits(oci_defaults_values)?,
+        OciSpecSection::Capabilities => oci_spec_capabilities(oci_defaults_values, runtime)?,
+        OciSpecSection::ResourceLimits => oci_spec_resource_limits(oci_defaults_values, runtime)?,
```
bcressey: This seems like a lot of indirection; consider just calling the associated method directly. That way all of the output is generated in the same place.

```suggestion
        (Runtime::Docker(r), OciSpecSection::Capabilities) => r.capabilities(oci_defaults_values)?,
        (Runtime::Docker(r), OciSpecSection::ResourceLimits) => r.resource_limits(oci_defaults_values)?,
        (Runtime::Containerd(r), OciSpecSection::Capabilities) => r.capabilities(oci_defaults_values)?,
        (Runtime::Containerd(r), OciSpecSection::ResourceLimits) => r.resource_limits(oci_defaults_values)?,
```

Might also be better to do this part before the call so the methods get the preferred form of data:
```
let oci_default_rlimits: HashMap<OciDefaultsResourceLimitType, OciDefaultsResourceLimit> = ...;
let oci_default_capabilities: HashMap<OciDefaultsCapability, bool> = ...;
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1369,6 +1372,62 @@ enum OciSpecSection {
...
+}
+
+impl Runtime {
+    /// Formats capabilities for different runtimes
+    fn get_capabilities(&self, caps: String) -> String {
+        match self {
+            Self::Docker => format!(
+                concat!(r#"["#, "{capabilities}", "],\n",),
+                capabilities = caps,
+            ),
```
bcressey: Per my other suggestion, I think it'd be better to have separate structs with different impls.

packages/docker-engine/0001-Change-default-capabilities-using-daemon-config.patch
```diff
@@ -0,0 +1,100 @@
...
+--- a/daemon/oci_linux.go
++++ b/daemon/oci_linux.go
+@@ -156,10 +156,15 @@ func WithApparmor(c *container.Container) coci.SpecOpts {
+ }
+ 
+ // WithCapabilities sets the container's capabilties
+-func WithCapabilities(c *container.Container) coci.SpecOpts {
++func WithCapabilities(daemon *Daemon, c *container.Container) coci.SpecOpts {
+ 	return func(ctx context.Context, _ coci.Client, _ *containers.Container, s *coci.Spec) error {
++		defCaps := caps.DefaultCapabilities()
```
bcressey: We should check whether the daemon has capabilities first to avoid calling this function and generating a value that will be immediately thrown away.

sources/api/schnauzer/src/helpers.rs
vyaghras: Implemented Runtime having get_capabilities and get_resource_limits as serde_plain complains about deserializing to enum as:
enum {
Docker(Docker),
Containerd(Containerd)
}

sources/models/src/aws-ecs-1/mod.rs
vyaghras: Rebased and added and tested the OCI defaults for aws-ecs-2

packages/docker-engine/docker-engine.spec
```diff
@@ -27,10 +27,13 @@ Source5: daemon-nvidia-json
 # Create container storage mount point.
 Source100: prepare-var-lib-docker.service
 
+
+
```
yeazelm: nit: we can avoid adding the whitespace here.

yeazelm (approved):

sources/models/shared-defaults/oci-defaults-docker-resource-limits.toml
```diff
@@ -0,0 +1,3 @@
+[settings.oci-defaults.resource-limits.max-open-files]
```
yeazelm, line 1: These limits seem really low. We set them significantly higher for [containerd](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/shared-defaults/oci-defaults-containerd-cri-resource-limits.toml). I think this is what the current limits are for `docker` but it feels strange to have wildly different limits between the two. We have three options in my mind (in order of my preference):
* Move these limits to align to `containerd` for all docker-based variants to remain consistent across Bottlerocket
* Move only new variants (like `aws-ecs-2`) to higher settings to avoid creating surprises for end users in case they are depending on these very low limits
* Keep `docker` limits lower than `containerd` by default
We are essentially taking the third option with this change as-is. I'd like to at least acknowledge we are doing that if we end up taking that route.

vyaghras: Yeah, the limits that we are using now, aligns with what we had for old Bottlerocket versions. But I think so too we can at least raise it for ecs-2 variants keeping same for ecs-1 variant.

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1368,9 +1374,106 @@ enum OciSpecSection {
     Capabilities,
     ResourceLimits,
 }
-
 derive_fromstr_from_deserialize!(OciSpecSection);
```
webern, line 1378: nit: do not delete this whitespace.

```suggestion

derive_fromstr_from_deserialize!(OciSpecSection);
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1368,9 +1374,106 @@ enum OciSpecSection {
...
 
+#[derive(Deserialize, Debug, Clone, Copy)]
+#[serde(rename_all = "kebab-case")]
+enum Runtime {
+    Docker,
+    Containerd,
+}
+derive_fromstr_from_deserialize!(Runtime);
```
webern: nit follow style seen elsewhere.

```suggestion
}

derive_fromstr_from_deserialize!(Runtime);
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -1486,30 +1596,14 @@ fn oci_spec_capabilities(value: &Value) -> Result<String, RenderError> {
...
+fn oci_spec_resource_limits(
+    value: &Value,
+) -> Result<HashMap<OciDefaultsResourceLimitType, OciDefaultsResourceLimit>, RenderError> {
+    Ok(serde_json::from_value(value.clone())?)
 }
 
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 // helpers to the helpers
-
 /// Gets the key name (path) of the parameter at `index`. Returns an error if the param cannot be extracted.
```
webern: nit: try not to make unnecessary style changes.

```suggestion

/// Gets the key name (path) of the parameter at `index`. Returns an error if the param cannot be extracted.
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -2701,6 +2795,8 @@ mod test_any_enabled {
 
 #[cfg(test)]
 mod test_oci_spec {
+    use super::Containerd;
+    use super::Docker;
```
webern: It's common to do this. But either way is fine. Though style-wise we would usually see multiple symbols from the same mod like this `use super::{Containerd, Docker};`

```suggestion
    use super::*
```

sources/api/schnauzer/src/helpers.rs
```diff
@@ -2701,6 +2795,8 @@ mod test_any_enabled {
...
     use crate::helpers::*;
     use serde_json::json;
     use OciDefaultsResourceLimitType::*;
```
webern, line 2810: It was already like this, so don't change it, but it is unusual to import an Enum this way. Normally we want enums as `use OciDefaultsResourceLimitType` and later `OciDefaultsResourceLimitType::MaxLockedMemory`.

sources/models/shared-defaults/oci-defaults-docker-resource-limits.toml
webern: Keeping the defaults unchanged on ecs-1 variants seems... important.

webern (approved):
Looks good.

sources/api/schnauzer/src/helpers.rs
vyaghras: This has ben done this way because we are using all the parms of enum like OciDefaultsResourceLimitType::MaxLockedMemory for all 16 RLimit Types that made it look redundant.

sources/models/shared-defaults/oci-defaults-docker.toml
```diff
@@ -0,0 +1,3 @@
+[metadata.settings.oci-defaults]
+affected-services = ["docker"]
+
```
bcressey: nit: don't need this extra newline
```suggestion
```

bcressey (approved):

sources/models/shared-defaults/oci-defaults-docker-resource-limits.toml
vyaghras: > * Keep `docker` limits lower than `containerd` by default
As discussed offline we are going to keep docker limits lower than Containerd as  it is possible to raise the default limits via task definition for ecs variants.

vyaghras: > * Keep `docker` limits lower than `containerd` by default
As discussed offline we are going to keep docker limits lower than Containerd as  it is possible to raise the default limits via task definition for ecs variants.

yeazelm (approved):

PR #3266, opened 2023-07-14
netdog: Add systemd-networkd config generation

zmrow:
**Issue number:**
Related to bottlerocket-os/bottlerocket#2449 

**Description of changes:**
*This PR is probably best read in commit order :)*

This is the grand finale, yes, the PR that puts all the pieces together for `systemd-networkd` config generation!

This PR drives the [previously added](https://github.com/bottlerocket-os/bottlerocket/pull/3220) network/netdev config builders via the `systemd-networkd` devices structs to create `systemd-networkd` config files.  

It also adds a new top-level type: `NetworkDConfig`.  This is the main type that gets created from net config; it contains the devices and a map of VLANs.  The devices are created via conversions of the various versions of net config; these conversions are contained in the newly added `conversions.rs`.  `NetworkDConfig` has a method `create_files()` that outputs all of the necessary configuration files for all configured devices.  Under the hood, each of the contained devices is driven to create the appropriate files for the device.  The device structs implement traits depending on their need for `.network` or `.netdev` files.  Inside the trait implementations, the devices drive the builders.

Conditional compilation bits are also tidied up, ensuring wicked stuff isn't included anywhere in builds that don't include it.

To update the diagram from bottlerocket-os/bottlerocket#3220 , this PR implements the left side of the diagram: the conversion of netconfig to `systemd-networkd` devices, and driving the builders to create config.

```
          This PR
 -------------------------------------           MERGED bottlerocket-os/bottlerocket#3220                            MERGED bottlerocket-os/bottlerocket#3134
           NET CONFIG                   -----------------------------------    --------------------
+---------------+                               NETWORKD DEVICES                 NETWORKD CONFIG
|               |
| InterfaceV1   +-----------------+   +-----------------+      Builder
|               |                 |   |                 +----------------+
+---------------+                 +-> |NetworkdInterface|                |
+---------------+                 +-> |                 |                |     +-----------------+
|               |                 |   |                 |                +---> |                 |
| InterfaceV2   +-----------------+   +-----------------+                      |                 |
|               |                                                 +----------> | .network        |
+---------------+                     +-----------------+ Builder |            |                 |
                                      |                 +---------+            |                 |
+---------------+               +---> | NetworkdBond    |              +-----> +-----------------+
|               |               |     |                 +---------+    |
|               +---------------+     |                 | Builder |    |       +-----------------+
| BondV1        |                     +-----------------+         |    |       |                 |
|               |                                                 +----+-----> |                 |
+---------------+                     +-----------------+              |       | .netdev         |
                                      |                 +--------------+       |                 |
+---------------+                     | NetworkdVlan    |                 +->  |                 |
|               |                +--> |                 |   Builder       |    +-----------------+
|               |                |    |                 +-----------------+
| VlanV1        +----------------+    +-----------------+
|               |
+---------------+
```

**Testing done:**
* Build an `aws-k8s-1.24` (with wicked) image to ensure nothing is broken over there: all good
* Boot an `aws-dev` (with systemd-networkd) image configured for early console in `preconfigured.target`, and observe the proper files being generated and network online!
```
bash-5.1# cat /run/systemd/network/10-eth0.network 
[Match]
Name=eth0
[Link]
RequiredForOnline=true
RequiredFamilyForOnline=ipv4
[Network]
DHCP=yes
IPv6AcceptRA=true
[DHCPv4]
UseDNS=true
UseDomains=true
[DHCPv6]
UseDNS=true
UseDomains=true

bash-5.1# networkctl status eth0
● 2: eth0                                                                                                      
                     Link File: /x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/network/80-release.link
                  Network File: /run/systemd/network/10-eth0.network
                          Type: ether
                         State: routable (configured)
                  Online state: online 
...
```
* Built and booted a `metal-dev` image with the following `net.toml` and observed the box boot, not hang waiting for optional DHCP6, and both interface come up and get DHCP4 addresses.  Found the proper config files in `/run/systemd/network/10-{eno1,eno2}.network`.  I also pulled the network cable from eno1, and observed the box wait for it to become online and `systemd-networkd-wait-online.service` fail, proving the `RequiredForOnline` bit is working.

```
bash-5.1# cat /var/lib/bottlerocket/net.toml
version=3
[eno1]
dhcp4=true
primary=true
[eno2]
dhcp4=true
[eno2.dhcp6]
enabled=true
optional=true

bash-5.1# cat /run/systemd/network/10-eno1.network
[Match]
Name=eno1
[Link]
RequiredForOnline=true
[Network]
DHCP=ipv4
[DHCPv4]
UseDNS=true
UseDomains=true

bash-5.1# cat /run/systemd/network/10-eno2.network
[Match]
Name=eno2
[Link]
RequiredForOnline=true
RequiredFamilyForOnline=ipv4
[Network]
DHCP=yes
[DHCPv4]
UseDNS=true
UseDomains=true
[DHCPv6]
UseDNS=true
UseDomains=true
```
* Using the same metal image, tested that configuration via MAC address (static and DHCP) functions properly.
* Using the same metal image, ensure simple bond configuration works properly with 2 devices.
* Ensure the `/etc/systemd/network` directory gets the proper context:
```
bash-5.1# ls -lahZ /etc/systemd/
...
drwxr-xr-x.  2 root root system_u:object_r:lease_t:s0  60 Aug  3 20:15 network
...
```

**Terms of contribution:**

### DIFF
--- a/packages/os/etc-systemd-network.mount
+++ b/packages/os/etc-systemd-network.mount
@@ -1,0 +1,16 @@
+[Unit]
+Description=systemd-networkd configuration directory
+DefaultDependencies=no
+Conflicts=umount.target
+Before=local-fs.target umount.target
+After=selinux-policy-files.service
+Wants=selinux-policy-files.service
+
+[Mount]
+What=tmpfs
+Where=/etc/systemd/network
+Type=tmpfs
+Options=nosuid,nodev,noexec,noatime,context=system_u:object_r:lease_t:s0,mode=0755
+
+[Install]
+WantedBy=systemd-networkd.service
--- a/sources/api/netdog/src/cli/generate_net_config.rs
+++ b/sources/api/netdog/src/cli/generate_net_config.rs
@@ -1,6 +1,6 @@
 use super::{error, Result};
 use crate::interface_id::InterfaceId;
-use crate::net_config;
+use crate::net_config::{self, Interfaces};
 use crate::{
     DEFAULT_NET_CONFIG_FILE, KERNEL_CMDLINE, OVERRIDE_NET_CONFIG_FILE, PRIMARY_INTERFACE,
     PRIMARY_MAC_ADDRESS,
@@ -50,20 +50,7 @@ pub(crate) fn run() -> Result<()> {
     remove_old_primary_interface()?;
     write_primary_interface(&primary_interface)?;
 
-    let mut wicked_interfaces = net_config.as_wicked_interfaces();
-    for interface in &mut wicked_interfaces {
-        // The kernel command line is too limited to fully specify an interface's configuration;
-        // fix some defaults to match legacy behavior.
-        // Note: we only allow 1 interface to be listed via kernel command line, so this will only
-        // be added to a single interface
-        if from_cmd_line {
-            interface.accept_ra();
-        }
-
-        interface
-            .write_config_file()
-            .context(error::InterfaceConfigWriteSnafu)?;
-    }
+    write_network_config_files(net_config, from_cmd_line)?;
     Ok(())
 }
 
@@ -86,4 +73,47 @@ fn write_primary_interface(interface_id: &InterfaceId) -> Result<()> {
     .context(error::PrimaryInterfaceWriteSnafu {
         path: PRIMARY_INTERFACE,
     })
+}
+
+#[cfg(net_backend = "wicked")]
+fn write_network_config_files(net_config: Box<dyn Interfaces>, from_cmd_line: bool) -> Result<()> {
+    let mut wicked_interfaces = net_config.as_wicked_interfaces();
+    for interface in &mut wicked_interfaces {
+        // The kernel command line is too limited to fully specify an interface's configuration;
+        // fix some defaults to match legacy behavior.
+        // Note: we only allow 1 interface to be listed via kernel command line, so this will only
+        // be added to a single interface
+        if from_cmd_line {
+            interface.accept_ra();
+        }
+
+        interface
+            .write_config_file()
+            .context(error::InterfaceConfigWriteSnafu)?;
+    }
+    Ok(())
+}
+
+#[cfg(net_backend = "systemd-networkd")]
+fn write_network_config_files(net_config: Box<dyn Interfaces>, from_cmd_line: bool) -> Result<()> {
+    use crate::networkd::config::NetworkDConfigFile;
+    let networkd_config = net_config
+        .as_networkd_config()
+        .context(error::NetworkDConfigCreateSnafu)?;
+    for mut config in networkd_config.create_files() {
+        // The kernel command line is too limited to fully specify an interface's configuration;
+        // fix some defaults to match legacy behavior.
+        // Note: we only allow 1 interface to be listed via kernel command line, so this will only
+        // be added to a single interface
+        if from_cmd_line {
+            if let NetworkDConfigFile::Network(ref mut n) = config {
+                n.accept_ra();
+            }
+        }
+
+        config
+            .write_config_file()
+            .context(error::NetworkDConfigWriteSnafu)?;
+    }
+    Ok(())
 }
--- a/sources/api/netdog/src/cli/mod.rs
+++ b/sources/api/netdog/src/cli/mod.rs
@@ -1,8 +1,6 @@
 pub(crate) mod generate_hostname;
 pub(crate) mod generate_net_config;
 pub(crate) mod node_ip;
-#[cfg(net_backend = "systemd-networkd")]
-pub(crate) mod primary_interface;
 pub(crate) mod set_hostname;
 pub(crate) mod write_resolv_conf;
 
@@ -11,6 +9,8 @@ pub(crate) mod install;
 #[cfg(net_backend = "wicked")]
 pub(crate) mod remove;
 
+#[cfg(net_backend = "systemd-networkd")]
+pub(crate) mod primary_interface;
 #[cfg(net_backend = "systemd-networkd")]
 pub(crate) mod write_primary_interface_status;
 
@@ -21,8 +21,6 @@ use crate::{
 pub(crate) use generate_hostname::GenerateHostnameArgs;
 pub(crate) use generate_net_config::GenerateNetConfigArgs;
 pub(crate) use node_ip::NodeIpArgs;
-#[cfg(net_backend = "systemd-networkd")]
-pub(crate) use primary_interface::PrimaryInterfaceArgs;
 use serde::{Deserialize, Serialize};
 pub(crate) use set_hostname::SetHostnameArgs;
 use snafu::{ensure, OptionExt, ResultExt};
@@ -37,6 +35,8 @@ pub(crate) use install::InstallArgs;
 #[cfg(net_backend = "wicked")]
 pub(crate) use remove::RemoveArgs;
 
+#[cfg(net_backend = "systemd-networkd")]
+pub(crate) use primary_interface::PrimaryInterfaceArgs;
 #[cfg(net_backend = "systemd-networkd")]
 pub(crate) use write_primary_interface_status::WritePrimaryInterfaceStatusArgs;
 
@@ -181,15 +181,17 @@ mod tests {
 
 /// Potential errors during netdog execution
 mod error {
-    #[cfg(net_backend = "wicked")]
-    use crate::lease;
-    use crate::{dns, interface_id, net_config, wicked};
-    #[cfg(net_backend = "systemd-networkd")]
-    use crate::{networkd, networkd_status};
+    use crate::{dns, interface_id, net_config};
     use snafu::Snafu;
     use std::ffi::OsString;
     use std::io;
     use std::path::PathBuf;
+
+    #[cfg(net_backend = "wicked")]
+    use crate::{lease, wicked};
+
+    #[cfg(net_backend = "systemd-networkd")]
+    use crate::{networkd, networkd_status};
 
     #[derive(Debug, Snafu)]
     #[snafu(visibility(pub(crate)))]
@@ -219,6 +221,7 @@ mod error {
         #[snafu(display("Failed to write hostname to '{}': {}", path.display(), source))]
         HostnameWriteFailed { path: PathBuf, source: io::Error },
 
+        #[cfg(net_backend = "wicked")]
         #[snafu(display("Failed to write network interface configuration: {}", source))]
         InterfaceConfigWrite { source: wicked::Error },
 
--- a/sources/api/netdog/src/cli/mod.rs
+++ b/sources/api/netdog/src/cli/mod.rs
@@ -183,9 +183,9 @@ mod tests {
 mod error {
     #[cfg(net_backend = "wicked")]
     use crate::lease;
-    #[cfg(net_backend = "systemd-networkd")]
-    use crate::networkd_status;
     use crate::{dns, interface_id, net_config, wicked};
+    #[cfg(net_backend = "systemd-networkd")]
+    use crate::{networkd, networkd_status};
     use snafu::Snafu;
     use std::ffi::OsString;
     use std::io;
@@ -255,6 +255,14 @@ mod error {
 
         #[snafu(display("Unable to find an interface with MAC address '{}'", mac))]
         NonExistentMac { mac: String },
+
+        #[cfg(net_backend = "systemd-networkd")]
+        #[snafu(display("Unable to create systemd-networkd config: {}", source))]
+        NetworkDConfigCreate { source: net_config::Error },
+
+        #[cfg(net_backend = "systemd-networkd")]
+        #[snafu(display("Failed to write network interface configuration: {}", source))]
+        NetworkDConfigWrite { source: networkd::Error },
 
         #[snafu(display("Unable to read '{}': {}", path.display(), source))]
         PathRead {
--- a/sources/api/netdog/src/net_config/v1.rs
+++ b/sources/api/netdog/src/net_config/v1.rs
@@ -14,6 +14,9 @@ use snafu::{ensure, OptionExt, ResultExt};
 use std::{collections::HashSet, str::FromStr};
 use std::{convert::TryInto, ops::Deref};
 
+#[cfg(net_backend = "systemd-networkd")]
+use crate::networkd::NetworkDConfig;
+
 #[derive(Debug, Deserialize)]
 pub(crate) struct NetConfigV1 {
     // Use an IndexMap to preserve the order of the devices defined in the net.toml.  The TOML
@@ -28,7 +31,7 @@ pub(crate) struct NetConfigV1 {
     pub(crate) interfaces: IndexMap<InterfaceName, NetInterfaceV1>,
 }
 
-#[derive(Debug, Deserialize)]
+#[derive(Debug, Deserialize, Clone)]
 #[serde(deny_unknown_fields)]
 pub(crate) struct NetInterfaceV1 {
     // Use this interface as the primary interface for the system
@@ -50,6 +53,7 @@ impl Interfaces for NetConfigV1 {
         !self.interfaces.is_empty()
     }
 
+    #[cfg(net_backend = "wicked")]
     fn as_wicked_interfaces(&self) -> Vec<WickedInterface> {
         let mut wicked_interfaces = Vec::with_capacity(self.interfaces.len());
         for (name, config) in &self.interfaces {
@@ -63,6 +67,12 @@ impl Interfaces for NetConfigV1 {
         }
 
         wicked_interfaces
+    }
+
+    #[cfg(net_backend = "systemd-networkd")]
+    fn as_networkd_config(&self) -> Result<NetworkDConfig> {
+        let devices = self.interfaces.clone().into_iter().collect();
+        NetworkDConfig::new(devices).context(error::NetworkDConfigCreateSnafu)
     }
 }
 
--- a/sources/api/netdog/src/net_config/v1.rs
+++ b/sources/api/netdog/src/net_config/v1.rs
@@ -3,16 +3,16 @@
 
 use super::{error, Error, Interfaces, Result, Validate};
 use crate::addressing::{Dhcp4ConfigV1, Dhcp4OptionsV1, Dhcp6ConfigV1, Dhcp6OptionsV1};
-use crate::{
-    interface_id::{InterfaceId, InterfaceName},
-    wicked::{WickedDhcp4, WickedDhcp6, WickedInterface},
-};
+use crate::interface_id::{InterfaceId, InterfaceName};
 use indexmap::indexmap;
 use indexmap::IndexMap;
 use serde::Deserialize;
 use snafu::{ensure, OptionExt, ResultExt};
 use std::{collections::HashSet, str::FromStr};
 use std::{convert::TryInto, ops::Deref};
+
+#[cfg(net_backend = "wicked")]
+use crate::wicked::{WickedDhcp4, WickedDhcp6, WickedInterface};
 
 #[cfg(net_backend = "systemd-networkd")]
 use crate::networkd::NetworkDConfig;
--- a/sources/api/netdog/src/networkd/config/mod.rs
+++ b/sources/api/netdog/src/networkd/config/mod.rs
@@ -39,7 +39,11 @@ mod private {
     pub enum Bond {}
     pub enum Interface {}
     pub enum Vlan {}
-    pub enum BondWorker {} // interfaces that are bound to a bond
+    // Interfaces that are bound to a bond
+    pub enum BondWorker {}
+    // Interfaces without config, used as the link for a VLAN: typically
+    // "tagged-only" setups
+    pub enum VlanLink {}
 
     // The devices for which we are generating a configuration file.  All device types should
     // implement this trait.
@@ -48,12 +52,19 @@ mod private {
     impl Device for Interface {}
     impl Device for Vlan {}
     impl Device for BondWorker {}
+    impl Device for VlanLink {}
 
     // Devices not bound to a bond, i.e. everything EXCEPT BondWorker(s)
     pub trait NotBonded {}
     impl NotBonded for Bond {}
     impl NotBonded for Interface {}
     impl NotBonded for Vlan {}
+
+    // Devices able to be members of VLANs
+    pub trait CanHaveVlans {}
+    impl CanHaveVlans for Bond {}
+    impl CanHaveVlans for Interface {}
+    impl CanHaveVlans for VlanLink {}
 }
 
 #[cfg(test)]
--- a/sources/api/netdog/src/networkd/config/mod.rs
+++ b/sources/api/netdog/src/networkd/config/mod.rs
@@ -71,8 +71,7 @@ mod private {
 mod tests {
     use crate::networkd::devices::{NetworkDBond, NetworkDInterface, NetworkDVlan};
     use serde::Deserialize;
-    use std::fs;
-    use std::path::{Path, PathBuf};
+    use std::path::PathBuf;
 
     pub(super) const BUILDER_DATA: &str = include_str!("../../../test_data/networkd/builder.toml");
 
--- a/sources/api/netdog/src/networkd/config/netdev.rs
+++ b/sources/api/netdog/src/networkd/config/netdev.rs
@@ -1,5 +1,5 @@
 use super::private::{Bond, Device, Vlan};
-use super::{CONFIG_FILE_PREFIX, NETWORKD_CONFIG_DIR};
+use super::CONFIG_FILE_PREFIX;
 use crate::bonding::{ArpMonitoringConfigV1, ArpValidateV1, BondModeV1, MiiMonitoringConfigV1};
 use crate::interface_id::InterfaceName;
 use crate::networkd::{error, Result};
@@ -105,8 +105,10 @@ impl Display for ArpValidate {
     }
 }
 
+// The `All` variant isn't currently used, but is valid
 #[derive(Debug)]
 enum ArpAllTargets {
+    #[allow(dead_code)]
     All,
     Any,
 }
--- a/sources/api/netdog/src/networkd/config/network.rs
+++ b/sources/api/netdog/src/networkd/config/network.rs
@@ -164,6 +164,23 @@ impl NetworkConfig {
         }
     }
 
+    /// The name of the device that corresponds to this config
+    // This method is useful but is only used in tests so far
+    #[cfg(test)]
+    pub(crate) fn name(&self) -> Option<InterfaceId> {
+        let maybe_name = self.r#match.as_ref().and_then(|m| m.name.as_ref());
+        let maybe_mac = self
+            .r#match
+            .as_ref()
+            .and_then(|m| m.permanent_mac_address.first());
+
+        match (maybe_name, maybe_mac) {
+            (Some(name), _) => Some(InterfaceId::from(name.clone())),
+            (None, Some(mac)) => Some(InterfaceId::from(mac.clone())),
+            (None, None) => None,
+        }
+    }
+
     /// Add config to accept IPv6 router advertisements
     // TODO: expose a network config option for this
     pub(crate) fn accept_ra(&mut self) {
--- a/sources/api/netdog/src/networkd/config/network.rs
+++ b/sources/api/netdog/src/networkd/config/network.rs
@@ -1,7 +1,7 @@
 use super::private::{
     Bond, BondWorker, CanHaveVlans, Device, Interface, NotBonded, Vlan, VlanLink,
 };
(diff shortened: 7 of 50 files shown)


sources/api/netdog/src/networkd/mod.rs
```diff
@@ -1,6 +1,25 @@
...
+pub(self) type Vlans = HashMap<InterfaceName, Vec<InterfaceName>>;
+
+/// Devices implement this trait if they require a .netdev file
+trait NetDevCreator {
+    fn create_netdev(&self) -> NetDevConfig;
+}
+
+/// Devices implement this trait if they require one or more .network files (bonds, for example,
+/// create multiple .network files for the bond and it's workers)
+trait NetworkCreator {
```
yeazelm: nit: Its wordy but `NetworkFileCreator` feels better, without seeing the context of other struct names, this reads like something that creates networks, not `.network` files.

sources/api/netdog/src/networkd/devices/interface.rs
```diff
@@ -16,3 +19,32 @@ pub(crate) struct NetworkDInterface {
     pub(crate) static6: Option<StaticConfigV1>,
     pub(crate) routes: Option<Vec<RouteV1>>,
 }
+
+impl NetworkCreator for NetworkDInterface {
+    fn create_networks(&self, vlans: &Vlans) -> Vec<NetworkConfig> {
```
yeazelm, line 45: I'm not sure I'm convinced that adding this parameter directly is future proof, there may be others in the future that we need to do this work with, should it be something like a container of network configurations we pass and right now the only one we know about is `Vlans`? I see this persist through the calls everywhere and feels awkward to call out vlans separately. It might be worth confirming if there are other composite network device types we might add in the future and just leave the door open for them vs making this interface explicit to worry about Vlans for all device file creation.

sources/api/netdog/src/networkd/mod.rs
```diff
@@ -1,14 +1,56 @@
...
+
+        let mut vlans = HashMap::new();
+        for device in &devices {
+            // VLANs are typically attached to a network device (bond/interface). Net config
+            // specifies the bond or interface in the VLAN config. In systemd-networkd config, the
+            // VLAN is specified in the device config. Create a map of interface/bond name to list
+            // of attached VLANs to assist with config generation later
+            if let NetworkDDevice::Vlan(vlan) = device {
+                vlans
+                    .entry(vlan.device.clone())
```
yeazelm, line 39: Do we support vlan on vlan? This currently doesn't check that the underlying device isn't itself a vlan. In theory this can be done I believe but we should validate this is not circular at the least. It might be simpler to just not support stacking vlans for now. For what its worth, most of the time, users would create multiple vlans on a single network device rather than stacking, so I think its reasonable to enforce this until someone explicitly asks for it.

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -162,6 +162,12 @@ impl NetworkConfig {
         }
     }
 
+    /// Add config to accept IPv6 router advertisements
+    // TODO: expose a network config option for this
```
yeazelm, line 187: Do we really intend to allow users to turn this off? If so, we might want to cut an issue to track it and watch for folks that actually ask for this.

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -162,6 +164,29 @@ impl NetworkConfig {
         }
     }
 
+    /// The name of the device this configuration is for
```
yeazelm:
```suggestion
    /// The name of the device that corresponds to this configuration
```

sources/api/netdog/src/networkd/mod.rs
```diff
@@ -20,4 +88,254 @@ mod error {
...
+                        path.display()
+                    )
+                }
+            }
+        }
+    }
+
+    #[test]
+    fn net_config_to_networkd_config() {
+        // let net_config_path = networkd_data().join("net_config.toml");
```
yeazelm: Is this needed?

yeazelm (commented):
With all the extra conditional compilation, I just want to confirm we have run `clippy` and others for both sides of netdog with and without systemd-networkd enabled to ensure the wicked stuff still builds how we expect. It was hard to track all those changes and hopefully our tools will help us catch any missing bits.

sources/api/netdog/src/networkd/config/network.rs
zmrow: The intent is to make it configurable.  We've got an issue for it! https://github.com/bottlerocket-os/bottlerocket/issues/2615

sources/api/netdog/src/networkd/mod.rs
zmrow: Oops - leftover from dev.  Will remove.

sources/api/netdog/src/networkd/devices/interface.rs
zmrow: :thinking:   That's probably a good callout.  

The only reason this is a bit odd, is because in net config, VLANs specify the device they're attached to.  But in networkd, the _device's_ config must contain the VLAN... 

I'm not sure if there is anything else like this.

sources/api/netdog/src/networkd/mod.rs
zmrow: :laughing:   I can see that.  There's lots of "network"-related names floating around in here.

sources/api/netdog/src/networkd/mod.rs
zmrow: I think a better place to do that type of validation is on the net config side of things, before we get to this point.  I'd rather not have validation in multiple places.  If the net config is "invalid" and we don't catch it networkd is going to choke on it; that feels like a decent balance.

sources/api/netdog/src/networkd/devices/interface.rs
zmrow: After poking around a bit more, I didn't find other situations where a devices config requires additional "things" from the overall net config.  I'm feeling more apt to leave this as-is until we come into a situation where we need additional params. (something something YAGNI)

zmrow:
^ Addresses @yeazelm 's comments.  Thanks!

zmrow:
Also, I confirmed `clippy` is happy and all the tests continue to pass.

sources/api/netdog/src/networkd/mod.rs
yeazelm: I think this is fair.

sources/api/netdog/src/networkd/devices/interface.rs
yeazelm: This makes sense, I just wanted to make sure we weren't ignoring future work but it sounds like this might remain intact for the foreseeable future so I'm happy to leave it even if it does look a bit odd.

yeazelm (approved):

etungsten (approved):
:rocket:

sources/api/netdog/src/networkd/devices/vlan.rs
```diff
@@ -19,3 +22,48 @@ pub(crate) struct NetworkDVlan {
...
+        let Self {
+            name,
+            dhcp4: _,
+            dhcp6: _,
+            static4: _,
+            static6: _,
+            routes: _,
+            device: _,
+            id,
+        } = self;
```
bcressey: The comment is "what" rather than "why" - it's obvious that some fields are skipped but it's not clear why.

For example, we never use the `device` field for a `NetworkDVlan` but I don't know why. If we never use it, why do we even have that field?

What failure case are you actually concerned about defending against here and is there a different mechanism we can use instead? 

The destructuring itself is OK, at least for functions where we use more of the fields - it's kind of excessive here - but the comment implies something is being ensured and I don't follow what that is.

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -162,6 +162,12 @@ impl NetworkConfig {
         }
     }
 
+    /// Add config to accept IPv6 router advertisements
+    // TODO: expose a network config option for this
+    pub(crate) fn accept_ra(&mut self) {
+        self.network_mut().ipv6_accept_ra = Some(true)
+    }
```
bcressey, line 190: nit: commit message should link to a specific commit, rather than saying "similar to the earlier change"

sources/api/netdog/src/networkd/config/network.rs
```diff
@@ -162,6 +162,23 @@ impl NetworkConfig {
         }
     }
 
+    /// The name of the device that corresponds to this config
+    // This method is useful but is only used in tests so far
+    #[allow(dead_code)]
```
bcressey: `#[cfg(test)]` would be clearer for code that's only used for tests - if it's useful later in some other context we can remove the conditional logic.

sources/api/netdog/test_data/networkd/net_config.toml
```diff
@@ -0,0 +1,215 @@
+version = {{version}}
+
+[eno1]
+dhcp4 = true
```
bcressey: Why is this `net_config.toml` different from the one for `wicked` tests? It seems like we'd want to validate the same input for both.

sources/api/netdog/test_data/networkd/network/eno1-ra.network
```diff
@@ -0,0 +1,10 @@
+[Match]
+Name=eno1
+[Link]
+RequiredForOnline=true
+[Network]
+DHCP=ipv4
+IPv6AcceptRA=true
+[DHCPv4]
+UseDNS=true
+UseDomains=true
```
bcressey, line 10: nit: it'd be great if we could separate sections with empty lines

sources/api/netdog/src/cli/mod.rs
```diff
@@ -181,16 +181,20 @@ mod tests {
...
     use std::io;
     use std::path::PathBuf;
 
+    #[cfg(net_backend = "wicked")]
+    use crate::{lease, wicked};
+
+    #[cfg(net_backend = "systemd-networkd")]
+    use crate::networkd;
+    #[cfg(net_backend = "systemd-networkd")]
+    use crate::networkd_status;
```
bcressey:
```suggestion
    #[cfg(net_backend = "systemd-networkd")]
    use crate::{networkd, networkd_status};
```

sources/api/netdog/src/cli/generate_net_config.rs
```diff
@@ -50,20 +50,47 @@ pub(crate) fn run() -> Result<()> {
...
 
-    let mut wicked_interfaces = net_config.as_wicked_interfaces();
-    for interface in &mut wicked_interfaces {
-        // The kernel command line is too limited to fully specify an interface's configuration;
-        // fix some defaults to match legacy behavior.
-        // Note: we only allow 1 interface to be listed via kernel command line, so this will only
-        // be added to a single interface
-        if from_cmd_line {
-            interface.accept_ra();
+    #[cfg(net_backend = "wicked")]
```
cbgbt: In general I think it's best to prefer to conditionally compile entire functions rather than code blocks within a function. Can we abstract this away into a call to something like `write_interface_config_files()` which we alter the implementation of instead?

sources/api/netdog/src/networkd/devices/mod.rs
```diff
@@ -13,3 +15,53 @@ pub(crate) enum NetworkDDevice {
...
+}
+
+/// A tiny macro to ease calling builder methods only if an `Option` is `Some()`.  NetworkDDevices
+/// contain quite a few `Option`s (and probably more in the future), so when driving the config
+/// builders there becomes a bit of `if let Some()` boilerplate.  This macro reduces that
+/// boilerplate to a single line.
+///
+/// The first argument is the builder, the second is the builder method, and the third is the
+/// Option from which you need to add.
+macro_rules! maybe_add_some {
```
cbgbt, line 69: I'm ok with using macros to eliminate boilerplate, but as a tip this could also be succinctly written as e.g.:

```rust
static4.map(|static4| network.with_static_config(static4));
```

sources/api/netdog/src/networkd/devices/interface.rs
```diff
@@ -16,3 +19,32 @@ pub(crate) struct NetworkDInterface {
...
+            routes,
+        } = self;
+
+        let mut network = NetworkBuilder::new_interface(name.clone());
+        network.with_dhcp(dhcp4.clone(), dhcp6.clone());
+        maybe_add_some!(network, with_static_config, static4);
+        maybe_add_some!(network, with_static_config, static6);
+        maybe_add_some!(network, with_routes, routes);
+
+        // Bottlerocket currently only supports attaching VLANs to interfaces via name (not MAC)
```
cbgbt: Do attempts to attach VLANs otherwise get caught and result in an error somewhere? Otherwise it seems like we might not appropriately handle user intent here.

sources/api/netdog/src/net_config/v1.rs
```diff
@@ -64,6 +68,12 @@ impl Interfaces for NetConfigV1 {
 
         wicked_interfaces
     }
+
+    #[cfg(net_backend = "systemd-networkd")]
+    fn as_networkd_config(&self) -> Result<NetworkDConfig> {
+        let devices = Vec::from_iter(self.interfaces.iter());
```
cbgbt: *hyper* nit, not gonna hold you up on this, but I think the compiler can work out what to do here and [`collect` is mentioned in the docs as being preferred](https://doc.rust-lang.org/std/iter/trait.FromIterator.html) unless you need a turbofish, but I don't think you need one here.

```suggestion
        let devices = self.interfaces.iter().collect();
```

sources/api/netdog/src/networkd/conversions.rs
```diff
@@ -0,0 +1,126 @@
...
+//! config structures to their corresponding networkd device structures
+use super::devices::{NetworkDBond, NetworkDDevice, NetworkDInterface, NetworkDVlan};
+use super::error;
+use crate::interface_id::{InterfaceId, InterfaceName};
+use crate::net_config::devices::bond::NetBondV1;
+use crate::net_config::devices::vlan::NetVlanV1;
+use crate::net_config::devices::{interface::NetInterfaceV2, NetworkDeviceV1};
+use crate::net_config::NetInterfaceV1;
+
+impl TryFrom<(&InterfaceId, &NetworkDeviceV1)> for NetworkDDevice {
```
cbgbt: It feels strange to me to implement this on references to `InterfaceId` and `NetworkdDeviceV1`, etc. My expectation around `From/Into` is that they take ownership of the input data and provide ownership to something new.

It seems like maybe this is fallout from using:

```rust
    let devices = Vec::from_iter(self.interfaces.iter());
    NetworkDConfig::new(devices);
```

since this will create an iterator over references to the interfaces. You could possibly align this with expectations a bit more cleanly by doing something like:

```rust
    NetworkDConfig::new(self.interfaces.iter().cloned().collect());
```

sources/api/netdog/src/networkd/devices/vlan.rs
zmrow: I'll update the comment.

>For example, we never use the device field for a NetworkDVlan but I don't know why. If we never use it, why do we even have that field?

The device field gets used to create the map of devices -> VLANs.  It doesn't directly get used to create files for the VLAN itself.  I'll add a comment explaining that.

> What failure case are you actually concerned about defending against here and is there a different mechanism we can use instead?

Basically I was trying to defend against the future addition of fields to the `NetworkDVlan` struct that accidentally don't get used to create files.  If we destructure, the compiler would yell about a specific field not being used, at which point the dev could decide if it's needed or not.

sources/api/netdog/test_data/networkd/network/eno1-ra.network
zmrow: I believe that should be doable via our macro - will look into it

sources/api/netdog/src/networkd/conversions.rs
```diff
@@ -0,0 +1,126 @@
+//! The conversions module contains all of the trait implementations necessary to convert net
+//! config structures to their corresponding networkd device structures
```
cbgbt, line 2: Possibly entering bike shedding territory, but it feels to me like it would be easier to maintain the conversions at the site at which the devices are defined, that way you could implement a new device without having to hop around the codebase too much.

sources/api/netdog/src/networkd/config/netdev.rs
```diff
@@ -105,6 +105,8 @@ impl Display for ArpValidate {
     }
 }
 
+// The `All` variant isn't currently used, but is valid
+#[allow(dead_code)]
 #[derive(Debug)]
 enum ArpAllTargets {
     All,
```
cbgbt:
```suggestion
#[derive(Debug)]
enum ArpAllTargets {
    #[allow(dead_code)]
    All,
```

sources/api/netdog/src/networkd/devices/mod.rs
zmrow: I originally had that exact code, but clippy was unhappy with it since the closure returns the unit type/nothing.   Clippy wanted me to go the `if let Some()` route which made this code three times as long.

The warning was: https://rust-lang.github.io/rust-clippy/master/index.html#/option_map_unit_fn

sources/api/netdog/src/networkd/conversions.rs
zmrow: I thought about that as well, but it required the net config devices modules to import networkd things, and vice versa, and that sort of circular import structure irked me.

I decided on the direction in the implementation so networkd imports everything, and doesn't need to export much.

sources/api/netdog/src/networkd/devices/interface.rs
zmrow: Yes - this gets caught and results in an error when the net config is parsed.  I think the comment wording implies that we should be doing validation here when I really just wanted to inform the reader that we would attach VLANs if the current interface's has a name (not a MAC).

zmrow: I'll update the comment wording

zmrow:
^ The above addresses all of @cbgbt and @bcressey 's comments except for one:

> Why is this net_config.toml different from the one for wicked tests? It seems like we'd want to validate the same input for both.

@bcressey's comment caused me to re-examine why i created a different `net_config.toml` for networkd.  This in turn, made me realize that I had missed implementing a subtle piece of the puzzle for VLANs.  With `wicked`, if an interface isn't otherwise configured, but is used by a VLAN, `wicked` will generate a "dummy" config for the interface.  This config basically brings the interface up without any DHCP.

With `systemd-networkd`, we need to generate this same dummy config for the interface, _but only if_ it is not otherwise configured.  I'm working on implementing that and will include include it in this PR, but it's tricky as it subtly breaks my current model since it requires knowledge of the overall config.

sources/api/netdog/test_data/networkd/network/eno1-ra.network
zmrow: I could blindly emit a newline before new section headings in the macro.  That would leave a newline at the beginning of the file.

I haven't figured out a "smarter" way, given the macro uses repetition to generate all the sections in order.  AFAIK, there's not really a way to enumerate/index during that repetition.

zmrow:
^ The above push addresses the couple issues we had identified in my last comment: 
* the ability to have an interface that is managed, but not otherwise configured - this interface is used only as the link for a VLAN.  
* Use the same `net_config.toml` for both wicked/networkd

This push adds an additional commit, which includes a new type parameter for the builder for `.network` files.  This type parameter is called `VlanLink` and it's constructor turns off addressing autoconfiguration.  The builder with this type parameter only allows a single method that adds VLANs to the link.

sources/api/netdog/test_data/net_config.toml
```diff
@@ -143,7 +143,7 @@ via = "2001:beef:beef::1"
 {{#if (eq version 3)}}
 [myvlan]
 kind = "vlan"
-device = "eno1"
+device = "eno100"
```
zmrow, line 146: This device was changed to eno100 because for `systemd-networkd` we expect to generate a file for this link.  Since we already have (and validate) an `eno1` in this config, we need to change the name in order to generate and validate a different file.

sources/api/netdog/test_data/networkd/network/bond1.network
```diff
@@ -2,13 +2,9 @@
 Name=bond1
 [Link]
 RequiredForOnline=true
-RequiredFamilyForOnline=both
 [Network]
 ConfigureWithoutCarrier=true
-DHCP=yes
+DHCP=ipv4
```
zmrow, line 7: These changes correct a typo I made in the config in an earlier commit.  This bond is meant to only have DHCP4.

zmrow:
The above commit adds an additional integration test to handle the case where a device is a member of a VLAN, but also has it's own addressing config.  Our current set of integration tests only covered the case where the underlying interface was _only_ used for the VLAN.

zmrow:
^ Oops - forgot to add the wicked reference files for the additional integration test.  Added those

sources/api/netdog/src/networkd/mod.rs
```diff
@@ -1,6 +1,88 @@
...
+            }
+        }
+
+        // If the VLANs map contains a device we don't otherwise have config for, it means the
+        // device is used only as a link for the VLAN.  This is used in VLAN "tagged only" type
+        // setups.  Add an empty NetworkDInterface to the list of config to be generated.  An empty
+        // device will get a .network file so it is managed and becomes a memeber of the VLAN, but
+        // will otherwise have all DHCP and addressing config turned off.
+        for vlan_device in vlans.keys() {
+            if !device_names.contains(&InterfaceId::from(vlan_device.clone())) {
```
yeazelm, line 51: I'm fine either way, but figured I should bring this up now since it occurred to me while reading it. Do we plan to use `device_names` anywhere else later? This logic modified `devices` which is what generates `device_names` but the set isn't re-generated. I wonder if it's worth adding newly created devices to this `Vec` just in case future changes want to reuse this, otherwise we have a potential gotcha for those following afterwards.

sources/api/netdog/src/networkd/devices/interface.rs
```diff
@@ -16,3 +19,70 @@ pub(crate) struct NetworkDInterface {
     pub(crate) static6: Option<StaticConfigV1>,
     pub(crate) routes: Option<Vec<RouteV1>>,
 }
+
+impl NetworkDInterface {
+    /// Does this interface contain any config?
```
bcressey: nit: this reads as a question from the author, like "What does this code do??"

`is_unconfigured()` might be better as a function name, and then the comment wouldn't be needed.

sources/api/netdog/src/networkd/mod.rs
```diff
@@ -1,6 +1,88 @@
...
+                    .entry(vlan.device.clone())
+                    .or_insert_with(Vec::new)
+                    .push(vlan.name.clone())
+            }
+        }
+
+        // If the VLANs map contains a device we don't otherwise have config for, it means the
+        // device is used only as a link for the VLAN.  This is used in VLAN "tagged only" type
+        // setups.  Add an empty NetworkDInterface to the list of config to be generated.  An empty
+        // device will get a .network file so it is managed and becomes a memeber of the VLAN, but
```
bcressey:
```suggestion
        // device will get a .network file so it is managed and becomes a member of the VLAN, but
```

sources/api/netdog/src/networkd/config/mod.rs
```diff
@@ -7,7 +7,7 @@ use super::Result;
 pub(crate) use netdev::{NetDevBuilder, NetDevConfig};
 pub(crate) use network::{NetworkBuilder, NetworkConfig};
 
-const NETWORKD_CONFIG_DIR: &str = "/etc/systemd/network";
+const NETWORKD_CONFIG_DIR: &str = "/run/systemd/network";
```
bcressey: I don't love this change - there are a lot more protections on `/etc`. Why is this needed?

sources/api/netdog/test_data/net_config.toml
```diff
@@ -155,6 +155,17 @@ id = 42
 [mystaticvlan.static4]
 addresses = ["192.168.1.100/24"]
 
+# This interface is a member of a VLAN but also has it's own addressing
+[eno1001]
+dhcp4 = true
+
+# This VLAN uses a device that also has it's own addressing config
+[vlancfgdev]
+kind = "vlan"
```
bcressey: (commit message has the same typo)

```suggestion
# This interface is a member of a VLAN but also has its own addressing
[eno1001]
dhcp4 = true

# This VLAN uses a device that also has its own addressing config
[vlancfgdev]
kind = "vlan"
```

sources/api/netdog/test_data/net_config.toml
bcressey: Doesn't this point to a bug in the `wicked` backend rendering? It seems like an error to generate `eno1.xml` and `myvlan.xml` that both reference the same device.

sources/api/netdog/src/networkd/mod.rs
zmrow: It _is_ possible that `device_names` could be useful later, but I'd rather not add any additional loops to this code to regenerate a list we don't currently use anywhere else after this runs.

zmrow:
^ The above push addresses @bcressey 's comments!

sources/api/netdog/src/networkd/config/mod.rs
zmrow: Rather than writing to `/run`, I've changed it back to `/etc/systemd/networkd`, and am creating that directory via mount unit with the correct context attached.

cbgbt: I haven't done much yet with Bottlerocket's mounts, but isn't `/etc` already mounted as tmpfs? Why do we need an additional mount?

zmrow: Indeed `/etc` is mounted as a tmpfs.  We need the directory `/etc/systemd/network` to be created, but because `systemd-networkd` runs with `DefaultDependencies=no` (which includes local-fs), we can't use tmpfiles.d to help us there.  A mount unit also allows us to create the directory with the proper SELinux label.

I'm open to suggestions if there's a better way that handles the timing issue we have!

sources/api/netdog/test_data/net_config.toml
zmrow: `wicked` is the opposite of `systemd-networkd` in that the interface's config file doesn't include the VLAN.  The VLAN's config file includes the interface name.  Since we're now using the same `net_config.toml`, we have to use a different interface name since there's already `eno1` in the file that _is not_ a member of the VLAN.

cbgbt (approved):

zmrow:
^ Changes the `etc-systemd-network.mount` ordering to `Before=generate-network-config.service`, rather than `Before=network-pre.target` since `generate-network-config.service` depends on the directory being present.  (Both will be run before `network-pre.target` based on the ordering in `generate-network-config.service`)

packages/os/etc-systemd-network.mount
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=systemd-networkd configuration directory
+DefaultDependencies=no
+Conflicts=umount.target
+Before=generate-network-config.service
```
bcressey: Need to add the standard mount unit dependencies:
```suggestion
Before=local-fs.target umount.target generate-network-config.service
```

But I'd really prefer that we used `RequiresMountsFor` in `generate-network-config.service` instead. Consider doing that in a drop-in for that unit that's added for `%{with systemd_networkd}`.

packages/os/etc-systemd-network.mount
```diff
@@ -0,0 +1,16 @@
...
+Wants=selinux-policy-files.service
+
+[Mount]
+What=tmpfs
+Where=/etc/systemd/network
+Type=tmpfs
+Options=nosuid,nodev,noexec,noatime,context=system_u:object_r:lease_t:s0,mode=0755
+
+[Install]
+RequiredBy=systemd-networkd.service
```
bcressey: A judgment call but I prefer "WantedBy" unless the other service can't possibly work without this.

```suggestion
WantedBy=systemd-networkd.service
```

bcressey (approved):
LGTM overall, some minor quibbles with the new unit's dependencies.

zmrow:
^ Addresses @bcressey 's comments re: the mount unit / drop-in unit.

Proof of the drop-in working:
```
bash-5.1# systemctl show generate-network-config.service
...
DropInPaths=/x86_64-bottlerocket-linux-gnu/sys-root/usr/lib/systemd/system/generate-network-config.service.d/requires-mounts-network-config.conf
...
RequiresMountsFor=/etc/systemd/network
...
```

bcressey (approved):

yeazelm (approved):

PR #3273, opened 2023-07-19
Add aws-ecs-2 variants

### LINKED ISSUE #2164: add a new aws-ecs variant for kernel 6.1
**What I'd like:**
I'd like a new variant for ECS that includes the new 6.1 kernel.

The ECS Agent works to preserve compatibility across its releases, so there's no upstream version to synchronize with in this case; it's still on 1.x versions.

With that in mind, possible options for a successor to `aws-ecs-1` include:
* `aws-ecs-1.1` - add a ".1" to reflect a new revision
* `aws-ecs-2` - increment "1" to reflect a significant difference in the new variant
* `aws-ecs-2022` - switch to year-based numbering to avoid version confusion with the agent

Introducing the 5.15 kernel in a future update to the `aws-ecs-1` variant, but this breaks with the compatibility promise we make for in-place updates.

arnaldo2792:
**Issue number:**

Closes #2164

**Description of changes:**

This adds the `aws-ecs-2` variants family. Features included in these variants include:
- Secure Boot enabled
- Kernel 6.1
- Use XFS for the data partition
- Support for cgroups v2 by default

**Testing done:**
- [X] Internal ECS testing suite ran successfully for `aws-ecs-2`
- [x] Internal ECS testing suite ran successfully for `aws-ecs-2-nvidia`

I confirmed all the mentioned features are enabled:

**`aws-ecs-2`**
```bash
bash-5.1# apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "24877a20",
    "pretty_name": "Bottlerocket OS 1.15.0 (aws-ecs-2)",
    "variant_id": "aws-ecs-2",
    "version_id": "1.15.0"
  }

### New kernel
bash-5.1# uname -r
6.1.29

### Secure Boot enabled
bash-5.1# journalctl -k | grep Secure
Jul 19 01:13:09 localhost kernel: Secure boot enabled

### cgroupsV2
bash-5.1# stat -fc %T /sys/fs/cgroup/
cgroup2fs

### Boot configs
[ssm-user@control]$ apiclient get settings.boot
{
  "settings": {
    "boot": {
      "init": {
        "systemd.log_level": [
          "info"
        ]
      },
      "kernel": {
        "console": [
          "tty0",
          "ttyS1,115200n8"
        ]
      }
    }
  }
}
[ssm-user@control]$ cat /proc/bootconfig
kernel.console = "tty0", "ttyS1,115200n8"
init.systemd.log_level = "info"


### XFS for data partition
bash-5.1# xfs_spaceman -c info /local
meta-data=/dev/nvme1n1p1         isize=512    agcount=41, agsize=130816 blks
         =                       reflink=1    bigtime=1 inobtcount=1 nrext64=0
data     =                       bsize=4096   blocks=5242368, imaxpct=25
         =                       sunit=128    swidth=128 blks
naming   =version 2              bsize=16384  ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=16384, version=2

```

**`aws-ecs-2-nvidia`**
```bash
bash-5.1# apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "24877a20",
    "pretty_name": "Bottlerocket OS 1.15.0 (aws-ecs-2-nvidia)",
    "variant_id": "aws-ecs-2-nvidia",
    "version_id": "1.15.0"
  }
}

### New kernel
bash-5.1# uname -r
6.1.29

### Secure Boot enabled
bash-5.1# journalctl -k | grep Secure
Jul 19 16:38:26 localhost kernel: Secure boot enabled

### cgroupsV2
bash-5.1# stat -fc %T /sys/fs/cgroup/
cgroup2fs

### Boot configs
[ssm-user@control]$ apiclient get settings.boot
{
  "settings": {
    "boot": {
      "init": {
        "systemd.log_level": [
          "info"
        ]
      },
      "kernel": {
        "console": [
          "tty0",
          "ttyS1,115200n8"
        ]
      }
    }
  }
}
[ssm-user@control]$ cat /proc/bootconfig
kernel.console = "tty0", "ttyS1,115200n8"
init.systemd.log_level = "info"


### XFS for data partition
bash-5.1# xfs_spaceman -c info /local
meta-data=/dev/nvme1n1p1         isize=512    agcount=37, agsize=130816 blks
         =                       reflink=1    bigtime=1 inobtcount=1 nrext64=0
data     =                       bsize=4096   blocks=4718080, imaxpct=25
         =                       sunit=128    swidth=128 blks
naming   =version 2              bsize=16384  ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=16384, version=2

### NVIDIA driver working:
bash-5.1# docker exec 6443448d5e32 nvidia-smi -L
GPU 0: Tesla T4 (UUID: GPU-69f2de54-cb1b-153e-bfdf-81bf722ad6ed)
```

**Terms of contribution:**

### DIFF
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -74,6 +74,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/aws-ecs-1/mod.rs)
 * [Default settings](src/aws-ecs-1/defaults.d/)
 
+## aws-ecs-1-nvidia: Amazon ECS NVIDIA
+
+* [Model](src/aws-ecs-1-nvidia/mod.rs)
+* [Default settings](src/aws-ecs-1-nvidia/defaults.d/)
+
 ## aws-dev: AWS development build
 
 * [Model](src/aws-dev/mod.rs)
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -79,6 +79,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/aws-ecs-1-nvidia/mod.rs)
 * [Default settings](src/aws-ecs-1-nvidia/defaults.d/)
 
+## aws-ecs-2: Amazon ECS
+
+* [Model](src/aws-ecs-1/mod.rs)
+* [Default settings](src/aws-ecs-1/defaults.d/)
+
 ## aws-dev: AWS development build
 
 * [Model](src/aws-dev/mod.rs)
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -84,6 +84,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/aws-ecs-1/mod.rs)
 * [Default settings](src/aws-ecs-1/defaults.d/)
 
+## aws-ecs-2-nvidia: Amazon ECS NVIDIA
+
+* [Model](src/aws-ecs-1-nvidia/mod.rs)
+* [Default settings](src/aws-ecs-1-nvidia/defaults.d/)
+
 ## aws-dev: AWS development build
 
 * [Model](src/aws-dev/mod.rs)
--- a/variants/aws-ecs-2-nvidia/Cargo.toml
+++ b/variants/aws-ecs-2-nvidia/Cargo.toml
@@ -1,0 +1,60 @@
+[package]
+name = "aws-ecs-2-nvidia"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[package.metadata.build-variant.image-features]
+grub-set-private-var = true
+unified-cgroup-hierarchy = true
+uefi-secure-boot = true
+xfs-data-partition = true
+
+[package.metadata.build-variant.image-layout]
+os-image-size-gib = 4
+
+[package.metadata.build-variant]
+included-packages = [
+# core
+    "release",
+    "kernel-6.1",
+# docker
+    "docker-cli",
+    "docker-engine",
+    "docker-init",
+    "docker-proxy",
+# ecs
+    "ecs-agent",
+# NVIDIA support
+    "ecs-gpu-init",
+    "nvidia-container-toolkit",
+    "kmod-6.1-nvidia-tesla-535",
+]
+
+kernel-parameters = [
+    "console=tty0",
+    "console=ttyS0,115200n8",
+    "net.ifnames=0",
+    "netdog.default-interface=eth0:dhcp4,dhcp6?",
+    "quiet",
+]
+
+[lib]
+path = "../variants.rs"
+
+[build-dependencies]
+# core
+release = { path = "../../packages/release" }
+kernel-6_1 = { path = "../../packages/kernel-6.1" }
+# docker
+docker-cli = { path = "../../packages/docker-cli" }
+docker-engine = { path = "../../packages/docker-engine" }
+docker-init = { path = "../../packages/docker-init" }
+docker-proxy = { path = "../../packages/docker-proxy" }
+# ecs
+ecs-agent = { path = "../../packages/ecs-agent" }
+# NVIDIA
+ecs-gpu-init = { path = "../../packages/ecs-gpu-init" }
+nvidia-container-toolkit = { path = "../../packages/nvidia-container-toolkit" }
+kmod-6_1-nvidia = { path = "../../packages/kmod-6.1-nvidia" }
--- a/variants/aws-ecs-2/Cargo.toml
+++ b/variants/aws-ecs-2/Cargo.toml
@@ -1,0 +1,50 @@
+[package]
+name = "aws-ecs-2"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[package.metadata.build-variant.image-features]
+grub-set-private-var = true
+unified-cgroup-hierarchy = true
+uefi-secure-boot = true
+xfs-data-partition = true
+
+[package.metadata.build-variant]
+included-packages = [
+# core
+    "release",
+    "kernel-6.1",
+# docker
+    "docker-cli",
+    "docker-engine",
+    "docker-init",
+    "docker-proxy",
+# ecs
+    "ecs-agent",
+]
+kernel-parameters = [
+    "console=tty0",
+    "console=ttyS0,115200n8",
+    "net.ifnames=0",
+    "netdog.default-interface=eth0:dhcp4,dhcp6?",
+    "quiet",
+]
+
+[lib]
+path = "../variants.rs"
+
+[build-dependencies]
+# core
+release = { path = "../../packages/release" }
+kernel-6_1 = { path = "../../packages/kernel-6.1" }
+# docker
+docker-cli = { path = "../../packages/docker-cli" }
+docker-engine = { path = "../../packages/docker-engine" }
+docker-init = { path = "../../packages/docker-init" }
+docker-proxy = { path = "../../packages/docker-proxy" }
+# ecs
+ecs-agent = { path = "../../packages/ecs-agent" }
--- a/packages/docker-engine/daemon-json
+++ b/packages/docker-engine/daemon-json
@@ -3,7 +3,6 @@
   "live-restore": true,
   "max-concurrent-downloads": 10,
   "storage-driver": "overlay2",
-  "exec-opts": ["native.cgroupdriver=cgroupfs"],
   "data-root": "/var/lib/docker",
   "default-runtime": "shimpei",
   "runtimes": { "shimpei": { "path": "shimpei" } },
--- a/packages/docker-engine/daemon-nvidia-json
+++ b/packages/docker-engine/daemon-nvidia-json
@@ -3,7 +3,6 @@
   "live-restore": true,
   "max-concurrent-downloads": 10,
   "storage-driver": "overlay2",
-  "exec-opts": ["native.cgroupdriver=cgroupfs"],
   "data-root": "/var/lib/docker",
   "default-runtime": "shimpei",
   "runtimes": { "shimpei": { "path": "shimpei" }, "nvidia": { "path": "nvidia-oci" } },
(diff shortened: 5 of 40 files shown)


sources/models/src/lib.rs
```diff
@@ -79,6 +79,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/aws-ecs-2/mod.rs)
 * [Default settings](src/aws-ecs-2/defaults.d/)
 
+## aws-ecs-2: Amazon ECS
+
+* [Model](src/aws-ecs-2/mod.rs)
```
stmcginnis: This will need to reference the `ecs-1` path. When this is in the rendered html view on GitHub, it is not able to follow the symlink from `ecs-2` to `ecs-1` where the actual file exists.

sources/models/src/lib.rs
```diff
@@ -74,6 +74,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/aws-ecs-1/mod.rs)
 * [Default settings](src/aws-ecs-1/defaults.d/)
 
+## aws-ecs-1-nvidia: Amazon ECS NVIDIA
+
+* [Model](src/aws-ecs-2/mod.rs)
```
stmcginnis: Should point to `ecs-1`.

sources/models/src/lib.rs
```diff
@@ -84,6 +84,11 @@ The `#[model]` attribute on Settings and its sub-structs reduces duplication and
 * [Model](src/aws-ecs-2/mod.rs)
 * [Default settings](src/aws-ecs-2/defaults.d/)
 
+## aws-ecs-2-nvidia: Amazon ECS NVIDIA
+
+* [Model](src/aws-ecs-2-nvidia/mod.rs)
```
stmcginnis: Same as others - needs to point to the `ecs-1` path due to symlinks.

stmcginnis (commented):
Looks good so far, just some minor issues pointed out.

arnaldo2792:
Forced push includes:
- Fix references, and point to the folder instead of symlinks
- Remove `cgroupdriver=cgroupfs` from docker config for NVIDIA

yeazelm (commented):
We might also want to update https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-ECS.md to discuss why one might choose `aws-ecs-1` vs `aws-ecs-2` and once we release we will want to update this guide to use latest. Otherwise, looks like a good start!

variants/aws-ecs-2/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+name = "aws-ecs-2"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[package.metadata.build-variant.image-features]
+grub-set-private-var = true
```
bcressey, line 11: This means we can safely add `BootSettings` / `settings.boot` support to these  new variants.

arnaldo2792:
(Forced push adds Boot Config APIs for both variants)

arnaldo2792:
>We might also want to update https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-ECS.md to discuss why one might choose aws-ecs-1 vs aws-ecs-2 and once we release we will want to update this guide to use latest. Otherwise, looks like a good start!

@yeazelm probably the QUICKSTART docs wouldn't be a good place to compare variants. AFAIK, we don't have any docs where we explicitly call out why users should favor specific variants. I do think I can update the docs to point to the `aws-ecs-2` variant once it is released, to begin steering people to use it.

yeazelm:
> > We might also want to update https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-ECS.md to discuss why one might choose aws-ecs-1 vs aws-ecs-2 and once we release we will want to update this guide to use latest. Otherwise, looks like a good start!
> 
> @yeazelm probably the QUICKSTART docs wouldn't be a good place to compare variants. AFAIK, we don't have any docs where we explicitly call out why users should favor specific variants. I do think I can update the docs to point to the `aws-ecs-2` variant once it is released, to begin steering people to use it.

I think at least calling out that we recommend `aws-ecs-2` for new launches in the Quickstart would be good. Currently, variants are more or less obvious via naming for which one a user should choose. But for ECS, it might not be as clear that there is no "ECS version 2" orchestrator like one might expect when looking the differences between `aws-k8s.1.26` and `aws-k8s-1.27`. I think we need to have the information captured _somewhere_ and the Quickstart isn't really the right place, but I don't know if there is anywhere better to help a user through the decision for which variant to choose when faced with `aws-ecs-1` vs `aws-ecs-2`. Maybe the right place is the website rather than this repo, but I wanted to address it now so we can have that documentation ready before these variants end up being published.

stmcginnis:
Maybe in the [list of variants](https://github.com/bottlerocket-os/bottlerocket/blob/develop/README.md#variants) we should just go ahead and put something like "(deprecated)" next to the `aws-ecs-1` variants so there is an indication that they can use it, but it will eventually be going away so they probably want to use the `aws-ecs-2` variants.

But good point that we kind of need something explaining what the differences are between 1 and 2. Maybe something for the project website?

arnaldo2792:
@stockholmux, is there a place on the website where the difference among the different variants for an orchestrator is documented?

stockholmux:
@arnaldo2792 It's mostly automatic. As part of the build process, the scripts pick  the all the variants in a given minor release. 

However, `ecs-1` -> `ecs-2` is a special case since it's the sole variant for that orchestrator, there are probably some assumptions made in the writing. Would you mind throwing an issue on bottlerocket-os/bottlerocket-project-website so I can track this update?

variants/aws-ecs-2-nvidia/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+kernel-parameters = [
+    "console=tty0",
+    "console=ttyS0,115200n8",
+    "net.ifnames=0",
+    "netdog.default-interface=eth0:dhcp4,dhcp6?",
+    "quiet",
+]
+
+[lib]
+path = "/dev/null"
```
webern: After rebasing #3288, please make this change. The rebase will not pick this up, nor will the build fail as a result of the rebase. Thanks!

```suggestion
path = "../variants.rs"
```

variants/aws-ecs-2/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+kernel-parameters = [
+    "console=tty0",
+    "console=ttyS0,115200n8",
+    "net.ifnames=0",
+    "netdog.default-interface=eth0:dhcp4,dhcp6?",
+    "quiet",
+]
+
+[lib]
+path = "/dev/null"
```
webern: After rebasing #3288, please make this change. The rebase will not pick this up, nor will the build fail as a result of the rebase. Thanks!

```suggestion
path = "../variants.rs"
```

arnaldo2792:
@stockholmux I created the issue: https://github.com/bottlerocket-os/bottlerocket-project-website/issues/198

@yeazelm, it seems like there is already an automated way to describe what is included in each variant but has to be “adapted” for ECS variants (see discussion above). Does that ease your concern?

yeazelm:
> @yeazelm, it seems like there is already an automated way to describe what is included in each variant but has to be “adapted” for ECS variants (see discussion above). Does that ease your concern?

Yes, that works for me!

stockholmux:
@arnaldo2792 With regards to the `QUICKSTART-ECS.md` on this repo, perhaps this is right time to transition to the [website quickstart for ECS](https://bottlerocket.dev/en/os/1.14.x/install/quickstart/aws/ecs/)? 

The website quickstart shouldn't matter if it's ecs-1 or ecs-2 (grabs the latest from SSM after switching one character). Otherwise, there seems to be only some missing info about regions and nvidia variants, both of which should probably exist on the website eventually, but they seem like a weird fit for a quickstart. 

What I would propose is to cut down the `QUICKSTART-ECS.md` to the places where it provides specific, unique value and point users to bottlerocket.dev. This is similar to what I'm proposing for EKS in #3285, so it seems like good timing.

Does this light anyone's hair aflame?

arnaldo2792:
(rebase onto develop to include the changes suggested by @webern)

arnaldo2792:
(forced push includes changes to reference `../variants.rs` instead of `/dev/null`)

arnaldo2792:
> @arnaldo2792 With regards to the `QUICKSTART-ECS.md` on this repo, perhaps this is right time to transition to the [website quickstart for ECS](https://bottlerocket.dev/en/os/1.14.x/install/quickstart/aws/ecs/)?
> 
> The website quickstart shouldn't matter if it's ecs-1 or ecs-2 (grabs the latest from SSM after switching one character). Otherwise, there seems to be only some missing info about regions and nvidia variants, both of which should probably exist on the website eventually, but they seem like a weird fit for a quickstart.
> 
> What I would propose is to cut down the `QUICKSTART-ECS.md` to the places where it provides specific, unique value and point users to bottlerocket.dev. This is similar to what I'm proposing for EKS in #3285, so it seems like good timing.
> 
> Does this light anyone's hair aflame?

I don't disagree, but I would like to keep that as a separate PR

variants/aws-ecs-2/Cargo.toml
```diff
@@ -0,0 +1,50 @@
+[package]
+name = "aws-ecs-2"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
```
bcressey, line 8: nit: it's a bit inconsistent to add this to `aws-ecs-2` but not `aws-ecs-2-nvidia` (or `aws-ecs-1` as well). Doesn't actually matter since none of these have a `README.md`.

bcressey (approved):
🚀

variants/aws-ecs-2/Cargo.toml
arnaldo2792: Oh, I saw that we have the same line for most of the variants. I can clean up all of them in a separate PR.

yeazelm (approved):

stmcginnis (approved):
Able to build, and looks fine to me.

PR #3282, opened 2023-07-21
Remove indirect dependencies from tools

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

Several tools had indirect dependencies included in their Cargo.toml files. These are pulled in by direct dependencies, so they should not be explicitly included in case those dependencies change and they are no longer needed, or require different versions.

**Testing done:**

Built each tool and verified no compilation errors.

**Terms of contribution:**

### DIFF
--- a/tools/Cargo.lock
+++ b/tools/Cargo.lock
@@ -2507,7 +2507,6 @@ dependencies = [
  "simplelog",
  "snafu",
  "tempfile",
- "toml",
  "url",
 ]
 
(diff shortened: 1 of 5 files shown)


jpculp (approved):

webern (approved):
This is great. Did you use `cargo udeps`?

stmcginnis:
> This is great. Did you use `cargo udeps`?

I actually used [`cargo machete`](https://github.com/bnjbvr/cargo-machete). I was reading up on some build time optimization and the author called out both tools, but recommended `machete` since it does not require the nightly toolchain to work. I didn't do a side-by-side comparison of results, but other than some false positives on our relatively linked modules (`model`, `bottlerocket-release`) it seemed to catch everything that wasn't being directly used.

Issue #3133, opened 2023-05-19
Re-implement Bottlerocket template system for Out-of-tree Builds


cbgbt:
Related to #2669

**What I'd Like**:
We need to be able to express settings dependencies for templates as described by the design document in #2834. This would allow the settings system to determine what settings and at what versions the API should render templates.

As an example, a template might look something like this:

```
[required-extensions]
motd = { version = "v1', helpers = ["boldface"] }
+++
Today's motd is {{ settings.motd }}

Let's pretend that the `motd` extension gives us a custom helper for boldfacing text {{ boldface "like this" }}.
```

Where the settings required are specified in a header called "frontmatter" and the template body is given as a string below.

To do:
* [x] Implement template parser (#3284)
* [x] Implement template renderer (#3302)
* [x] Implement settings generator binary (#3316)
* [x] Migrate `thar-be-settings` and existing config templates to the new renderer library (https://github.com/bottlerocket-os/bottlerocket/pull/3346)
* [x] Migrate `sundog` and settings-generators to the new binary (#3316)
* [ ] Replace [`ReplaceTemplateMigration`](https://github.com/bottlerocket-os/bottlerocket/blob/ef2fe106efd55728596c0d89cbfce53a58f79e61/sources/api/migration/migration-helpers/src/common_migrations.rs#L522)
* [x] Write a migration guide to help users port existing templates (#3488)

PR #3284, opened 2023-07-21
schnauzer: implement v2 template parser

cbgbt:
**Issue number:**
#3133

**Description of changes:**
The v2 template format requires all templates to include a frontmatter section which explicitly specifies each setting and helper that is used to render the template.

We use `pest` to separate the frontmatter of the document from its body, then serde to parse the frontmatter document.

**Testing done:**
The included tests pass.

**Terms of contribution:**


stmcginnis (approved):
Looks fine to me, and test coverage is good!

sources/api/schnauzer/src/v2/grammars/toml.pest
```diff
@@ -0,0 +1,74 @@
+// pest. The Elegant Parser
+// Copyright (c) 2018 Dragoș Tiselice
+//
+// Licensed under the Apache License, Version 2.0
+// <LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0> or the MIT
+// license <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. All files in the project carrying such notice may not be copied,
+// modified, or distributed except according to those terms.
```
bcressey, line 8: There is no way to nest parsers and use [TomlParser](https://docs.rs/pest_grammars/latest/pest_grammars/toml/struct.TomlParser.html) for this?

If not, let's add a link to the source of this file.

sources/api/schnauzer/tests/templates/succeeds/04_comments.template
```diff
@@ -0,0 +1,7 @@
++++
+# comments are ignored in toml
+[required-extensions]
+labrador = { version = "v1", helpers = [ "woof"] }
```
bcressey: nit:
```suggestion
labrador = { version = "v1", helpers = [ "woof" ] }
```

sources/api/schnauzer/tests/templates/succeeds/04_comments.template
```diff
@@ -0,0 +1,7 @@
++++
+# comments are ignored in toml
```
bcressey: Do we want to allow comments outside of the front matter?

I could see someone adding a comment to the start of the doc, given that they're usually ignored, and we work with a lot of YAML and TOML files.

```
# FIXME: clean this up later
+++
# is this needed?
[required-extensions]
foo = "v1"
```

sources/api/schnauzer/src/v2/grammars/template.pest
```diff
@@ -0,0 +1,10 @@
+template = ${ SOI ~ frontmatter ~ body ~ EOI }
+
+// Allow loose whitespace inside frontmatter
+frontmatter = !{ WHITESPACE* ~ frontmatter_delim ~ toml_document ~ frontmatter_delim }
+frontmatter_delim = @{ "+++" ~ "+"* ~ NEWLINE }
```
bcressey: As I understand it, the purpose of the TOML parsing - over just looking for the delimiters - is that otherwise we could be confused by input like:
```
[required-extensions]
labrador = """
+++
"""
```

If that's the case, let's add a test case for that, assuming some form of that input could be valid.

sources/api/schnauzer/tests/v2_parse_fail_templates.rs
```diff
@@ -0,0 +1,50 @@
...
+#[test]
+fn fails_00_invalid_toml() {
+    // Given a template with invalid toml in frontmatter,
+    // when the template is parsed,
+    // then a GrammarParse error is returned.
+    match include_str!("./templates/fails/00_invalid_toml.template").parse::<Template>() {
+        Err(template::error::Error::GrammarParse { .. }) => (),
+        _ => panic!("Did not encounter expected error while parsing template."),
+    };
+}
```
bcressey: nit: this could potentially be turned into a macro:
```
assert_parse_error!(GrammarParse, "./templates/fails/00_invalid_toml.template");
```

bcressey (approved):

sources/api/schnauzer/src/v2/grammars/toml.pest
cbgbt: Sadly, no. We need references to some of the grammatical rules which are no longer present in the parser. I'll add a link.

sources/api/schnauzer/tests/templates/succeeds/04_comments.template
cbgbt: I'd be okay with allowing comments prior to the frontmatter, or even just removing the starting `+` signs.

sources/api/schnauzer/src/v2/grammars/template.pest
cbgbt: Yes, that is the reason. I'll add a test.

cbgbt:
^ Addresses feedback from @bcressey

bcressey (approved):

sources/api/schnauzer/Cargo.toml
```diff
@@ -19,14 +19,18 @@ handlebars = "4"
 http = "0.2"
 lazy_static = "1"
 log = "0.4"
+maplit = "1.0"
```
zmrow, line 22: Perhaps I missed it being used outside of tests, but if `maplit` is only used in tests, can we move it to `[dev-dependencies]` section?

sources/api/schnauzer/src/v2/template.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    fn from_str(input_str: &str) -> Result<Template> {
+        // Potential improvement here would be to use `ariadne` crate to emit more helpful error messages
+        let mut parsed_pairs = TemplateParser::parse(Rule::template, input_str)
+            .context(error::GrammarParseSnafu)?
+            .flatten();
+
+        let frontmatter_pair = parsed_pairs
+            .clone()
+            .find(|p| p.as_rule() == Rule::toml_document)
```
zmrow, line 122: I know that `find()` short-circuits when it finds a match; is there a reason we would need to ensure no more matches exist?  That is: should there only be a single match?  Do we need to throw an error if there are more?

sources/api/schnauzer/src/v2/template.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub))]
+    pub enum Error {
+        #[snafu(display("Error when parsing template grammar: '{}'\n\nThis is usually due to errors in frontmatter TOML formatting.", source))]
+        GrammarParse {
+            source: pest::error::Error<super::Rule>,
+        },
+        #[snafu(display("Error when parsing template frontmatter: '{}'", source))]
+        FrontmatterParse { source: toml::de::Error },
```
zmrow, line 157: Nit: Elsewhere in our code we've added a newline between error enum variants, just to make reading through them a bit easier/clearer.

sources/api/schnauzer/Cargo.toml
cbgbt: I'm trying to PR this in smaller chunks, but the next commit in the stack uses `maplit` as a dependency in the library, so rather than juggle it about I just left it here.

cbgbt:

zmrow (approved):
:cake:

sources/api/schnauzer/tests/templates/succeeds/05_unambiguous_delims.template
```diff
@@ -0,0 +1,6 @@
+[required-extensions]
+beagle = """
++++
```
webern, line 3: Nice test.

sources/api/schnauzer/tests/v2_parse_fail_templates.rs
```diff
@@ -0,0 +1,51 @@
...
+    ($error:ident, $template:expr) => {
+        match include_str!($template).parse::<Template>() {
+            Err(template::error::Error::$error { .. }) => (),
+            _ => panic!("Did not encounter expected error while parsing template."),
+        };
+    };
+}
+
+#[test]
+fn fails_00_invalid_toml() {
```
webern, line 13: I appreciate that you have written these as separate tests instead of cycling through directory templates. Separate tests are much better!

sources/api/schnauzer/src/v2/template.rs
```diff
@@ -0,0 +1,162 @@
...
+impl FromStr for Template {
+    type Err = error::Error;
+
+    fn from_str(input_str: &str) -> Result<Template> {
+        // Potential improvement here would be to use `ariadne` crate to emit more helpful error messages
+        let mut parsed_pairs = TemplateParser::parse(Rule::template, input_str)
+            .context(error::GrammarParseSnafu)?
+            .flatten();
+
+        // If the template succesfully parses, we know there's always exactly one TOML document and one template body.
```
webern, line 119: Is this true even if front matter is not present? Is front matter always required in a template, or are there cases where front matter would be optional?

sources/api/schnauzer/src/v2/template.rs
```diff
@@ -0,0 +1,162 @@
...
+        },
+
+        #[snafu(display("Error when parsing template frontmatter: '{}'", source))]
+        FrontmatterParse { source: toml::de::Error },
+
+        #[snafu(display("Error in template parser: '{}'", message))]
+        TemplateParserLogic { message: &'static str },
+    }
+}
+pub use error::Error;
```
webern: nit

```suggestion
}

pub use error::Error;
```

webern (approved):
Looks good.

sources/api/schnauzer/src/v2/template.rs
cbgbt: As written, the frontmatter is required, but it can be an empty document. In these cases, the template is like this:

```
+++
template body
```

We actually have a template in Bottlerocket right now which is a static file, and looks just like this (I can't recall which offhand).

cbgbt: I will add a testcase that does this.

cbgbt:
Comments from @webern had me realize that there was a mismatch in my own understanding of static-text templates and my implementation. The two changes above fix that, and also address a stylistic nit.

For what it's worth, we must *always* require the `+++` delimiter at the start of a template, otherwise it is impossible for the parser to appropriately disambiguate a template body with `+++` in it, especially if that body is a TOML document.

webern (approved):

Issue #3057, opened 2023-04-26
High memory utilization on 1.13.4 release 


yeazelm:
**Image I'm using:**
`ami-035c9bfb9c905837c` - but I believe this is with any AWS K8S variant. I've reproduced this on x86_64. 

**What I expected to happen:**
The cluster nodes to work without filling up on memory and errors to come in the logs about "Timed out while waiting for systemd to remove" in the `kubelet` logs.

**What actually happened:**

```
Apr 26 21:38:55 ip-192-168-81-168.us-west-2.compute.internal kubelet[1373]: I0426 21:38:55.258940    1373 pod_container_manager_linux.go:192] "Failed to delete cgroup paths" cgroupName=[kubepods besteffort pod482dc361-4c7d-4502-b54d-4ce53af7089e] err="unable to destroy cgroup paths for cgroup [kubepods besteffort pod482dc361-4c7d-4502-b54d-4ce53af7089e] : Timed out while waiting for systemd to remove kubepods-besteffort-pod482dc361_4c7d_4502_b54d_4ce53af7089e.slice"
```

**How to reproduce the problem:**
I launch a cluster with 2 nodes, add in a some load (in this case simulated by a webserver and some pods calling the webserver) and add in a CronJob as follows:
```
apiVersion: batch/v1
kind: CronJob
metadata:
  name: hello
spec:
  schedule: "* * * * *"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: hello
            image: busybox:1.28
            imagePullPolicy: IfNotPresent
            command:
            - /bin/sh
            - -c
            - date; echo Hello from the Kubernetes cluster; sleep $(( ( RANDOM % 10 )  + 1 ));
          restartPolicy: OnFailure
          hostNetwork: true
      parallelism: 60
```
After some time, the memory on the node fills up with `kubelet` taking up GB of ram and the logs contain logs of errors with failing to clean up the cgroups.

This seems to be a problem when lots of pods are cycled through. It may also have to do with the CronJob terminating the container since just simulating load and deleting the running containers manually doesn't seem to trigger it.

This does not happen on 1.13.3.

bcressey:
[...]

PR #3290, opened 2023-07-24
Systemd update to 252

### LINKED ISSUE #2410: Update to systemd 251/252/253
https://github.com/systemd/systemd/blob/main/NEWS

Try updating systemd to 251/252/253. 

252.6 is latest stable version https://github.com/systemd/systemd-stable/releases/tag/v252.6

arnaldo2792:
**Description of changes:**
This updates systemd to v252. I'm using the latest version (to this day) of the series, since we need to include [this commit], to prevent [this problem]. I only saw this issue manifesting with the 5.10 kernel and with sysmted <=252.4. 

There are two commits in this PR. The first commit removes all the numbers from the existing local patches, since a few commits had to be reworked, and it was hard to differentiate an actual reworked patch. The second commit includes the changes for the update. All the back-ported patches were deleted since the new version includes all the changes in the patches.

A new feature to compress the journal was introduced [in v252], which isn't backwards-compatible. This feature is disabled with an environment variable set in the `systemd-journald` service.

[this commit]: https://github.com/systemd/systemd-stable/commit/de8409ac43f6e4596de4cecce8dbbb5f1f2a18b1
[this problem]: https://github.com/systemd/systemd/issues/25811
[in v252]: https://github.com/systemd/systemd-stable/blob/ca7787a9f29ff836a300b782a73f9c5926ee0d14/NEWS#L1070

**Testing done:**
- [x] Heavy workload test to verify that the kubelet doesn't cause OOM, like in https://github.com/bottlerocket-os/bottlerocket/issues/3057
- [x] Test the journal can be read with `journalctl` in the admin container
- [x] Test the journal can be read after a migration to the new systemd version
- [x] Test the journal lists all the boots
- [x] Test the journal can be read after a rollback
- [x] Test instances boot with unified partitioning
- 10,000 instance launches for these instance types:
  - [x] `m5.large`
  - [x] `m6g.medium`
  - [x] `t3.nano`
  - [x] `t4g.medium`
  - [x] `m5.16xlarge`
  - [x] `m6a.12xlarge`
  - [x] `m6g.12xlarge`

## Systemd 252/251 NEWS summary

This summary only includes what caught my attention and I thought was worth calling out. Refer to the full [NEWS] for greater detail.

[NEWS]: https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS

### Call out

* They said that in future versions of systemd (>252), support for cgroups v1 will be dropped. Thus, we can’t move to any newer version until all the existing variants that use cgroups v1 are removed.

### Compatibility Breaks

* For [this](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L23), We don’t use `ConditionKernelVersion`
* For [this](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L34), We don’t have SELinux labels in service units
* For [this](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L619), we are moving out of 250.11, so technically that version already included the changes to disable the feature described in the link.

### New features

* Nothing particularly appealing. Most of the new tools are related to Unified Kernel Images or other services like `systemd-homed`, which we don't use.

### Changes in systemd itself

* Systemd added a [new compile option](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L90C11-L94C19) to do a “full preset” in the first boot. This option defaults to `false` in this release, but will be enabled by default in upcoming releases. It doesn’t affect us for this update, but we need to keep an eye for future updates.

### Changes in systemd-networkd

* There are new features in [v252](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L262) and [v251](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L1034). I’d like folks working with systemd-networkd to test if their current work doesn’t break with the update, or if they could benefit from any of the new features (cc @zmrow / @yeazelm)

### Changes in other components

* Journalctl has a [new feature](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L412) that ins’t backwards-compatible with the journalctl version in the admin container. It was disabled as suggested in the linked section.
* We could benefit from [this feature](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L454) in `systemd-repart` while building the disk partitions(?). Something to consider.
* Folks working with `systemd-networkd` and the network stack should keep an eye [on this](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L494)
* There is a [new flag](https://github.com/systemd/systemd-stable/blob/b5ac1d45ea6ddbcc1fe996e81585d186b8719e43/NEWS#L1144) to set the default compression for the journal. Nothing is set by default, but could be nice to use one of the supported algorithms to save a few extra bytes of space. We will have to test how existing tools that collect logs behave if the logs are compressed.

**Terms of contribution:**


stmcginnis (approved):
Seems fine from a basic operations perspective.

I built and launched an instance. Everything came up OK and it was able to join the cluster.

Ran the following successfully:

```
systemctl status kubelet
systemctl restart kubelet
journalctl
journalctl -u kubelet
```

Ran `sonobuoy run`:

```
...
21:38:27             e2e                                         global   complete   passed   Passed:366, Failed:  0
21:38:27    systemd-logs   ip-192-168-24-231.us-east-2.compute.internal   complete   passed
```

Ran:

```
apiclient apply <EOF
[settings.kubernetes]
image-gc-high-threshold-percent = 85
image-gc-low-threshold-percent = 80
EOF
```

to trigger a kubelet config change. Then ran `systemctl status kubelet` again to make sure the service was restarted successfully.

```
Active: active (running) since Tue 2023-07-25 21:44:54 UTC; 5s ago
```

packages/systemd/systemd-journald.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+Environment=SYSTEMD_JOURNAL_COMPACT=0
```
bcressey:
```suggestion
[Service]
Environment=SYSTEMD_JOURNAL_KEYED_HASH=0
Environment=SYSTEMD_JOURNAL_COMPACT=0
```

To drop patch 9007.

FWIW the reason I implemented it as a patch was because I wanted the patch to fail if the `SYSTEMD_JOURNAL_KEYED_HASH` option was ever removed upstream. That would be more obvious to a developer than an environment variable silently being ignored, though either regression should be caught by testing.

I'm fine with either approach but we should be consistent.

packages/systemd/systemd.spec
```diff
@@ -388,6 +365,7 @@ install -p -m 0644 %{S:4} %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/i
 %exclude %{_cross_factorydir}%{_cross_sysconfdir}/pam.d
 %exclude %{_cross_factorydir}%{_cross_sysconfdir}/pam.d/other
 %exclude %{_cross_factorydir}%{_cross_sysconfdir}/pam.d/system-auth
+%exclude %{_cross_factorydir}%{_cross_sysconfdir}/locale.conf
```
bcressey: This might be worth keeping, assuming it just contains `LANG=C.UTF-8`.

bcressey (commented):
Can you diff the features and options output for this with systemd 250? The build log will dump out a config like this:
```
#20 14.08 systemd 250  
#20 14.08·  
#20 14.08     build mode                     : release  
#20 14.08     split /usr                     : False  
#20 14.08     split bin-sbin                 : True 
...
```

I'd also recommend going through `meson_options.txt` and looking for new options that we may want to set explicitly, usually to turn off.

bcressey:
From v252 release notes:
> Changes in systemd-networkd: the RapidCommit= is (re-)introduced to enable faster configuration via DHCPv6 (RFC 3315).
@zmrow this could be of interest in the move to networkd, but we should dig into why it was withdrawn.

arnaldo2792:
I updated the PR description and included a summary of what I thought it was worth calling out from the NEWS file.

arnaldo2792:
@bcressey: this is the diff:

```diff
--- 250 2023-08-02 23:03:30.172333928 +0000
+++ 252 2023-08-02 23:03:20.900340189 +0000
@@ -1,4 +1,4 @@
-   systemd 250
+   systemd 252

        build mode                     : release
        split /usr                     : False
@@ -19,8 +19,10 @@
        D-Bus policy directory         : /x86_64-bottlerocket-linux-gnu/sys-root/usr/share/dbus-1/system.d
        D-Bus session directory        : /x86_64-bottlerocket-linux-gnu/sys-root/usr/share/dbus-1/services
        D-Bus system directory         : /x86_64-bottlerocket-linux-gnu/sys-root/usr/share/dbus-1/system-services
+       D-Bus interfaces directory     : no
        bash completions directory     : no
        zsh completions directory      : no
+       private shared lib version tag : 252
        extra start script             : /etc/rc.local
        debug shell                    : /bin/sh @ /dev/tty9
        system UIDs                    : <=999 (alloc >=201)
@@ -35,6 +37,7 @@
        nobody user name               : nobody
        nobody group name              : nobody
        fallback hostname              : localhost
+       default compression method     : none
        default DNSSEC mode            : no
        default DNS-over-TLS mode      : no
        default mDNS mode              : no
@@ -55,13 +58,15 @@
        default net.naming-scheme value: latest
        default KillUserProcesses value: True
        default locale                 : C.UTF-8
+       default nspawn locale          : C.UTF-8
+       default status unit format     : description
        default user $PATH             : (same as system services)
        systemd service watchdog       : 3min
-       time epoch                     : 1676581145 (2023-02-16T20:59:05+00:00)
+       time epoch                     : 1689671493 (2023-07-18T09:11:33+00:00)

      Features
-       enabled                        : ACL, SECCOMP, SELinux, blkid, libfdisk, efi, networkd, pstore, randomseed, repart, systemd-analyze, sysusers, tmpfiles, kmod, ldconfig, gshadow, link-udev-shared, link-systemctl-shared
-       disabled                       : AUDIT, AppArmor, IMA, PAM, SMACK, elfutils, gcrypt, gnutls, libbpf, libcryptsetup, libcryptsetup-plugins, libcurl, libfido2, libidn, libidn2, libiptc, microhttpd, openssl, p11kit, pcre2, pwquality, qrencode, tpm2, xkbcommon, zstd, lz4, xz, zlib, bzip2, backlight, binfmt, bpf-framework, coredump, environment.d, gnu-efi, firstboot, hibernate, homed, hostnamed, hwdb, importd, initrd, kernel-install, localed, logind, machined, nss-myhostname, nss-mymachines, nss-resolve, nss-systemd, oomd, portabled, quotacheck, resolve, rfkill, sysext, timedated, timesyncd, userdb, vconsole, xdg-autostart, idn, polkit, nscd, legacy-pkla, dbus, glib, tpm, man pages, html pages, man page indices, SysV compat, compat-mutable-uid-boundaries, utmp, adm group, wheel group, debug hashmap, debug mmap cache, debug siphash, valgrind, trace logging, install tests, link-networkd-shared, link-timesyncd-shared, link-boot-shared, fexecve, standalone-binaries, static-libsystemd, static-libudev, cryptolib, DNS-over-TLS
+       enabled                        : ACL, SECCOMP, SELinux, blkid, libfdisk, efi, networkd, pstore, randomseed, repart, systemd-analyze, sysusers, tmpfiles, kmod, ldconfig, gshadow, link-udev-shared, link-systemctl-shared, link-journalctl-shared
+       disabled                       : AUDIT, AppArmor, IMA, PAM, SMACK, elfutils, gcrypt, gnutls, libbpf, libcryptsetup, libcryptsetup-plugins, libcurl, libfido2, libidn, libidn2, libiptc, microhttpd, openssl, p11kit, pcre2, pwquality, qrencode, tpm2, xkbcommon, zstd, lz4, xz, zlib, bzip2, backlight, binfmt, bpf-framework, coredump, environment.d, gnu-efi, firstboot, hibernate, homed, hostnamed, hwdb, importd, initrd, kernel-install, localed, logind, machined, nss-myhostname, nss-mymachines, nss-resolve, nss-systemd, oomd, portabled, quotacheck, resolve, rfkill, sysext, sysupdate, timedated, timesyncd, userdb, vconsole, xdg-autostart, idn, polkit, nscd, legacy-pkla, dbus, glib, tpm, man pages, html pages, man page indices, SysV compat, compat-mutable-uid-boundaries, utmp, adm group, wheel group, debug hashmap, debug mmap cache, debug siphash, valgrind, trace logging, install tests, link-networkd-shared, link-timesyncd-shared, link-boot-shared, first-boot-full-preset, fexecve, standalone-binaries, coverage, static-libsystemd, static-libudev, cryptolib, DNS-over-TLS

      User defined options
        Cross files                    : ./cross-compilation.conf

```

arnaldo2792:
I'm working on a change to explicitly disable a few features that are disabled by default, but we don't want any surprises

arnaldo2792:
Forced push includes:

- Drop patch 9007, and use environment variable isntead
- Include `locale.conf`
- Explicitly disable these features: `sysupdate`, `log-message-verification`, `efi-tpm-pcr-compat`

etungsten (approved):
Thanks for testing at scale.

bcressey (approved):
🚀

yeazelm (approved):

PR #3295, opened 2023-07-25
docker-go: change docker run mounts

### LINKED ISSUE #3294: ootb: docker-go mounts need to match host mounts
Similar to what we did in #3287, we need to update the docker-go script so that the docker run mount commands use paths inside the container matching those outside the container. This is because twoliter will be calling this from within a container that has the host's docker socket mounted. As currently written, the current GO_MOD_CACHE gets lost because it is written to the twoliter container's `/tmp`

https://github.com/bottlerocket-os/bottlerocket/blob/332ce730ef1c923057788de83af0b8fc6d4d05ce/tools/docker-go#L87C41-L87C41

Ben caught this here https://github.com/bottlerocket-os/bottlerocket/pull/3287#pullrequestreview-1544414385

This is for #2669 and https://github.com/bottlerocket-os/twoliter/issues/14

webern:
**Issue number:**

Closes #3294

**Description of changes:**

Twoliter will be calling this script from within a container that has the host's docker socket mounted. It is necessary to match internal and external paths for any state that we wish to preserve on the host.

Because the data we want to cache is at $GOPATH/pkg/go, it was necessary to create this as a subdirectory under .gomodcache and mount that so that we could point GOPATH to .gomodcache.

**Testing done:**

`cargo make unit-test` works and the cached items are found where expected.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -98,7 +98,8 @@ BUILDSYS_UPSTREAM_LICENSE_FETCH= "false"
 BUILDSYS_JOBS = "8"
 
 CARGO_HOME = "${BUILDSYS_ROOT_DIR}/.cargo"
-GO_MOD_CACHE = "${BUILDSYS_ROOT_DIR}/.gomodcache"
+# This needs to end with pkg/mod so that we can mount the parent of pkg/mod as GOPATH.
+GO_MOD_CACHE = "${BUILDSYS_ROOT_DIR}/.gomodcache/pkg/mod"
 GO_MODULES = "ecs-gpu-init host-ctr"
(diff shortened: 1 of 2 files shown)


stmcginnis (approved):
Looks good to me!

webern:
> https://github.com/bottlerocket-os/bottlerocket/compare/fe4f6fb504f9aa959e15ecdbef136cecb1ae737a..2a337ef928ed01b77a66129ec79fa685638db4fe

Added a comment to explain GO_MOD_CACHE sensitivity to ending with `pkg/mod`.

bcressey (approved):

PR #3296, opened 2023-07-26
Add package linux-firmware and use it on metal-* variants

### LINKED ISSUE #3246: Bare Metal image: "qed" network driver exists in image but driver firmware file is missing
**Image I'm using:**
Bare Metal Bottlerocket variant (bottlerocket-metal-k8s-1.27-x86_64-v1.14.0.img.gz)

When booting Bottlerocket it shall be possible to use the "qed" network driver which is embedded in the image.

**What actually happened:**
When Bottlerocket boots on a HP DL360 Gen10 the network is not started although the required driver is contained in the image. The required driver (QLogic "qed") is loaded, but the firmware file is missing in the image. See the 2 screenshots below:

bottlerocket_qede_fw_error1

bottlerocket_qede_fw_error2

The driver firmware file is contained in the Linux kernel tree here: https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/plain/qed/qed_init_values_zipped-8.42.2.0.bin. To fix the issue this file must be placed in `/lib/firmware/qed/qed_init_values_zipped-8.42.2.0.bin`.

The network adapter model in the DL360 server is an "HPE Ethernet 10/25Gb 2-port 640FLR-SFP28 Adapter".

Boot on hardware with a network device which needs the "QLogic Everest FastLinQ Linux Driver" (qed).

foersleo:
**Issue number:**

Closes #3246

**Description of changes:**

With introduction of Metal variants we opened up Bottlerocket to a wider variety of Hardware configurations. In the past we have added driver support for some server grade hardware, mostly in the 10G+ NIC realm. While adding the drivers, we overlooked that some of these devices may need firmware binaries to be available in order to update firmware and function with the driver version shipped with the kernel.

Add firmware from [linux-firmware](https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/) to our metal variants for drivers we support in Bottlerocket. I have made an effort to cut down the list of binaries to only the ones we currently supply drivers for, as the linux-firmware package is quite large in full. I was able to cut the size of the installed size on disk down from 426 MB for the full package to 29 MB for the stripped down version carrying only the needed firmware.

We could strip the size on disk down even further when using compressed firmware images. Currently ZSTD and XZ compression are supported. While ZSTD does not change the size significantly XZ compression can further reduce the size from 29 MB (uncompressed) to 25 MB (xz compressed). However, doing that requires us to enable additional kernel config options `CONFIG_FW_LOADER_COMPRESS` and `CONFIG_FW_LOADER_COMPRESS_XZ`. That optimization can probably be part of a separate **commit.**

The resulting firmware package carries firmware for the following list of drivers and their corresponding config options we have configured for our metal kernels:

```
cxgb4 - CONFIG_CHELSIO_T4
tg3 - CONFIG_TIGON3
bnx2x - CONFIG_BNX2X
netxen_nic - CONFIG_NETXEN_NIC
qed - CONFIG_QED
myri10ge - CONFIG_MYRI10GE
i915 - CONFIG_DRM_I915
ice - CONFIG_ICE
```

**Testing done:**

I have done some build testing. Unfortunately assessing if all the firmware is in the correct place and can be reached by the drivers is beyond what I can do as I do not have access to these devices.

**Terms of contribution:**

### DIFF
--- a/packages/linux-firmware/Cargo.toml
+++ b/packages/linux-firmware/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "linux-firmware"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[package.metadata.build-package]
+package-name = "linux-firmware"
+
+[lib]
+path = "../packages.rs"
+
+[[package.metadata.build-package.external-files]]
+url = "https://www.kernel.org/pub/linux/kernel/firmware/linux-firmware-20230625.tar.xz"
+sha512 = "0e48aa7f63495485426d37491c7cb61843165625bd47f912c5d83628c6de871759f1a78be3af3d651f7c396bd87dff07e21ba7afc47896c1c143106d5f16d351"
--- a/packages/linux-firmware/linux-firmware.spec
+++ b/packages/linux-firmware/linux-firmware.spec
@@ -11,7 +11,16 @@ Name: %{_cross_os}linux-firmware
 Version: 20230625
 Release: 1%{?dist}
 Summary: Firmware files used by the Linux kernel
-License: GPL+ and GPLv2+ and MIT and Redistributable, no modification permitted
+# The following list of SPDX identifiers was constructed with help of scancode
+# tooling and has turned up the following licenses for different drivers by
+# checking the different LICENCE/LICENSE files and the licenses in WHENCE:
+# * BSD-Source-Code - myri10ge
+# * LicenseRef-scancode-chelsio-linux-firmware - cxgb4
+# * LicenseRef-scancode-qlogic-firmware - netxen_nic
+# * LicenseRef-scancode-intel - i915, ice
+# * LicenseRef-scancode-proprietary-license - bnx2x, qed
+# * LicenseRef-scancode-free-unknown - tg3
+License: GPL-1.0-or-later AND GPL-2.0-or-later AND BSD-Source-Code AND LicenseRef-scancode-chelsio-linux-firmware AND LicenseRef-scancode-qlogic-firmware AND LicenseRef-scancode-intel AND LicenseRef-scancode-proprietary-license AND LicenseRef-scancode-free-unknown
 URL: https://www.kernel.org/
 
 Source0: https://www.kernel.org/pub/linux/kernel/firmware/linux-firmware-%{version}.tar.xz
--- a/packages/linux-firmware/linux-firmware.spec
+++ b/packages/linux-firmware/linux-firmware.spec
@@ -16,6 +16,17 @@ URL: https://www.kernel.org/
 
 Source0: https://www.kernel.org/pub/linux/kernel/firmware/linux-firmware-%{version}.tar.xz
 
+Patch0001: 0001-linux-firmware-snd-remove-firmware-for-snd-audio-dev.patch
+Patch0002: 0002-linux-firmware-video-Remove-firmware-for-video-broad.patch
+Patch0003: 0003-linux-firmware-bt-wifi-Remove-firmware-for-Bluetooth.patch
+Patch0004: 0004-linux-firmware-scsi-Remove-firmware-for-SCSI-devices.patch
+Patch0005: 0005-linux-firmware-usb-remove-firmware-for-USB-Serial-PC.patch
+Patch0006: 0006-linux-firmware-ethernet-Remove-firmware-for-ethernet.patch
+Patch0007: 0007-linux-firmware-Remove-firmware-for-Accelarator-devic.patch
+Patch0008: 0008-linux-firmware-gpu-Remove-firmware-for-GPU-devices.patch
+Patch0009: 0009-linux-firmware-various-Remove-firmware-for-various-d.patch
+Patch0010: 0010-linux-firmware-amd-ucode-Remove-amd-microcode.patch
+
 %description
 %{summary}.
 
--- a/packages/linux-firmware/linux-firmware.spec
+++ b/packages/linux-firmware/linux-firmware.spec
@@ -1,0 +1,47 @@
+%global debug_package %{nil}
+
+%global fwdir %{_cross_libdir}/firmware
+
+# Many of the firmware files have specialized binary formats that are not supported
+# by the strip binary used in __spec_install_post macro. Work around build failures
+# by skipping striping.
+%global __strip /usr/bin/true
+
+Name: %{_cross_os}linux-firmware
+Version: 20230625
+Release: 1%{?dist}
+Summary: Firmware files used by the Linux kernel
+License: GPL+ and GPLv2+ and MIT and Redistributable, no modification permitted
+URL: https://www.kernel.org/
+
+Source0: https://www.kernel.org/pub/linux/kernel/firmware/linux-firmware-%{version}.tar.xz
+
+%description
+%{summary}.
+
+%prep
+%autosetup -n linux-firmware-%{version} -p1
+
+%build
+
+%install
+mkdir -p %{buildroot}/%{fwdir}
+mkdir -p %{buildroot}/%{fwdir}/updates
+
+# Here we have potential to shave off some extra space by using `install-xz` of
+# `install-zst` to compress firmware images on disk. However, that functionality
+# relies on kernels being configured with `CONFIG_FW_LOADER_COMPRESS_[ZSTD|XZ]`
+# which we currently do not have.
+make DESTDIR=%{buildroot}/ FIRMWAREDIR=%{fwdir} install
+
+
+# Remove executable bits from random firmware
+pushd %{buildroot}/%{fwdir}
+find . -type f -executable -exec chmod -x {} \;
+popd
+
+%files
+%dir %{fwdir}
+%{fwdir}/*
+%license LICENCE.* LICENSE.* GPL* WHENCE
+%{_cross_attribution_file}
--- a/packages/linux-firmware/0001-linux-firmware-snd-remove-firmware-for-snd-audio-dev.patch
+++ b/packages/linux-firmware/0001-linux-firmware-snd-remove-firmware-for-snd-audio-dev.patch
@@ -1,0 +1,1979 @@
+From c5e85907233a5ddc7df414744561f0fd848e907f Mon Sep 17 00:00:00 2001
+From: Leonard Foerster <foersleo@amazon.com>
+Date: Tue, 25 Jul 2023 09:22:51 +0000
+Subject: [PATCH] linux-firmware: snd: remove firmware for snd/audio devices
+
+Bottlerocket does not configure drivers for audio devices for any of its
+kernels. Hence, we do not need to ship firmware for these devices. The
+following list maps the drivers as they are named in WHENCE to the
+kernel config option for reference and easy searchability should we need
+to find the right firmware to add when adding a driver to our kernel
+configuation:
+
+* snd-korg1212 - CONFIG_SND_KORG1212
+* snd-maestro3 - CONFIG_SND_MAESTRO3
+* snd-ymfpci - CONFIG_SND_YMFPCI
+* snd-sb16_csp - CONFIG_SND_SB16_CSP
+* snd-wavefront - CONFIG_SND_WAVEFRONT
+* mtk-sof - CONFIG_SND_SOC_SOF_MTK_TOPLEVEL
+* snd-hda-codex-ca0132 - CONFIG_SND_HDA_INTEL && CONFIG_SND_HDA_CODEC_CA0132
+* snd_soc_sst_acpi - CONFIG_SND_SOC_INTEL_SST_ACPI
+* snd_intel_sst_core - CONFIG_SND_SOC_INTEL_SST
+* snd_soc_catpt - CONFIG_SND_SOC_INTEL_CATPT
+* snd_soc_avs - CONFIG_SND_SOC_INTEL_AVS
+* snd_soc_skl - CONFIG_SND_SOC_INTEL_SKL
+* cs35l41_hda - CONFIG_SND_SOC_C35L41* && SND_HDA_SCODEC_CS35L41*
+
+Signed-off-by: Leonard Foerster <foersleo@amazon.com>
+---
+ LICENCE.IntcSST2    |  39 --
+ LICENCE.adsp_sst    | 999 --------------------------------------------
+ LICENCE.ca0132      |  47 ---
+ LICENCE.fw_sst_0f28 |  40 --
+ LICENSE.cirrus      | 182 --------
+ WHENCE              | 545 ------------------------
+ 6 files changed, 1852 deletions(-)
+ delete mode 100644 LICENCE.IntcSST2
+ delete mode 100644 LICENCE.adsp_sst
+ delete mode 100644 LICENCE.ca0132
+ delete mode 100644 LICENCE.fw_sst_0f28
+ delete mode 100644 LICENSE.cirrus
+
+diff --git a/LICENCE.IntcSST2 b/LICENCE.IntcSST2
+deleted file mode 100644
+index d4f1609..0000000
+--- a/LICENCE.IntcSST2
++++ /dev/null
+@@ -1,39 +0,0 @@
+-Copyright (c) 2014, Intel Corporation.
+-All rights reserved.
+-
+-Redistribution.  Redistribution and use in binary form, without
+-modification, are permitted provided that the following conditions are
+-met:
+-
+-* Redistributions must reproduce the above copyright notice and the
+-  following disclaimer in the documentation and/or other materials
+-  provided with the distribution.
+-* Neither the name of Intel Corporation nor the names of its suppliers
+-  may be used to endorse or promote products derived from this software
+-  without specific prior written permission.
+-* No reverse engineering, decompilation, or disassembly of this software
+-  is permitted.
+-
+-Limited patent license.  Intel Corporation grants a world-wide,
+-royalty-free, non-exclusive license under patents it now or hereafter
+-owns or controls to make, have made, use, import, offer to sell and
+-sell ("Utilize") this software, but solely to the extent that any
+-such patent is necessary to Utilize the software alone, or in
+-combination with an operating system licensed under an approved Open
+-Source license as listed by the Open Source Initiative at
+-http://opensource.org/licenses.  The patent license shall not apply to
+-any other combinations which include this software.  No hardware per
+-se is licensed hereunder.
+-
+-DISCLAIMER.  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND
+-CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+-BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+-FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+-COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+-INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+-BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS
+-OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+-ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+-TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+-USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+-DAMAGE.
+diff --git a/LICENCE.adsp_sst b/LICENCE.adsp_sst
+deleted file mode 100644
+index c66b1b2..0000000
+--- a/LICENCE.adsp_sst
++++ /dev/null
+@@ -1,999 +0,0 @@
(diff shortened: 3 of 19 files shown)


foersleo:
I have not put the source tar in the lookaside cache yet. Hence, the failing test runs. I want to check the licensing on these firmware blobs we ship first. The linux-firmware package is an interesting place with regards to licenses.

Edit: So all the binaries that will currently go into this package are marked as Redistributable with no modification permitted.

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,89 @@
...
+# Many of the firmware files have specialized binary formats that are not supported
+# by the strip binary used in __spec_install_post macro. Work around build failures
+# by skipping striping.
+%global __strip /usr/bin/true
+
+Name:       %{_cross_os}linux-firmware
+Version:    20230625
+Release:    1%{?dist}
+Summary:    Firmware files used by the Linux kernel
+License:    GPL+ and GPLv2+ and MIT and Redistributable, no modification permitted
```
bcressey: This needs to be a valid SPDX expression with SPDX license identifiers.

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,89 @@
...
+URL:        https://www.kernel.org/
+
+Source0:    https://www.kernel.org/pub/linux/kernel/firmware/linux-firmware-%{version}.tar.xz
```
bcressey: Are there any dependencies between these releases and the particular kernel versions?

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,89 @@
+%global debug_package %{nil}
+
+%global _firmwarepath   /usr/lib/firmware
```
bcressey: This should be under the sysroot path, otherwise it will conflict with our `/usr` symlink at build time.

```suggestion
%global fwdir %{_cross_libdir}/firmware
```

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,89 @@
...
+
+# Many of the firmware files have specialized binary formats that are not supported
+# by the strip binary used in __spec_install_post macro. Work around build failures
+# by skipping striping.
+%global __strip /usr/bin/true
+
+Name:       %{_cross_os}linux-firmware
+Version:    20230625
+Release:    1%{?dist}
+Summary:    Firmware files used by the Linux kernel
```
bcressey: nit: I try to avoid the column-aligned layouts since those can lead to whitespace-related churn in future diffs, where other lines have to be adjusted to preserve the column

```suggestion
Name: %{_cross_os}linux-firmware
Version: 20230625
Release: 1%{?dist}
Summary: Firmware files used by the Linux kernel
```

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,89 @@
...
+Patch0007: 0007-linux-firmware-Remove-firmware-for-Accelarator-devic.patch
+Patch0008: 0008-linux-firmware-gpu-Remove-firmware-for-GPU-devices.patch
+Patch0009: 0009-linux-firmware-various-Remove-firmware-for-various-d.patch
+Patch0010: 0010-linux-firmware-amd-ucode-Remove-amd-microcode.patch
+
+BuildRequires:  make
+Requires:   linux-firmware-whence
+Provides:   kernel-firmware = %{version}
+Obsoletes:  kernel-firmware < %{version}
+
```
bcressey: We shouldn't need any of these - just put the "whence" related files in the same package as the others.
```suggestion
```

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,89 @@
...
+# relies on kernels being configured with `CONFIG_FW_LOADER_COMPRESS_[ZSTD|XZ]`
+# which we currently do not have.
+make DESTDIR=%{buildroot}/ FIRMWAREDIR=%{_firmwarepath} install
+
+# Cleanup unnecessary files
+pushd %{buildroot}/%{_firmwarepath}
+
+rm -f check_whence.py configure Makefile README
+
+#Remove executable bits from random firmware
```
bcressey:
```suggestion
# Remove executable bits from random firmware
```

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,89 @@
...
+# Create file list
+FILEDIR=`pwd`
+pushd %{buildroot}/%{_firmwarepath}
+find . \! -type d > $FILEDIR/linux-firmware.files
+find . -type d | sed -e '/^.$/d' > $FILEDIR/linux-firmware.dirs
+popd
+sed -i -e 's:^./::' linux-firmware.{files,dirs}
+sed -i -e 's!^!/usr/lib/firmware/!' linux-firmware.{files,dirs}
+sed -i -e 's/^/"/;s/$/"/' linux-firmware.files
+sed -e 's/^/%%dir /' linux-firmware.dirs >> linux-firmware.files
```
bcressey: I would just bulk-include everything in the firmware directory, possibly splitting it up by device or directory if it's not an overwhelming number. (Like the headers subpackage in the kernel packages, with `asm`, `xen` includes.)

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,89 @@
...
+# Here we have potential to shave off some extra space by using `install-xz` of
+# `install-zst` to compress firmware images on disk. However, that functionality
+# relies on kernels being configured with `CONFIG_FW_LOADER_COMPRESS_[ZSTD|XZ]`
+# which we currently do not have.
+make DESTDIR=%{buildroot}/ FIRMWAREDIR=%{_firmwarepath} install
+
+# Cleanup unnecessary files
+pushd %{buildroot}/%{_firmwarepath}
+
+rm -f check_whence.py configure Makefile README
```
bcressey: We could potentially just exclude these four files in the `%files` section.

packages/linux-firmware/linux-firmware.spec
foersleo: So if I am reading the SPDX grammar and License list correctly this would be something like this:

```
GPL-1.0-or-later AND GPL-2.0-or-later AND MIT AND ?
```

A lot of the firmware is licensed as "Redistributable, no modification permitted" (or that is at least the name for that bucket that fedora uses. All the firmware we currently have left to include falls into that group. If I understand correctly there is currently no solution for these in the SPDX guidelines. The [Fedora licensing guide](https://docs.fedoraproject.org/en-US/legal/update-existing-packages/#_redistributable_no_modification_permitted) marks usage of that with SPDX as unknown.

What do can we make of this?

I can see that for some of the binaries we include (if we take this PR) we might be able to do something like `LicenseRef-scancode-chelsio-linux-firmware` (according to https://scancode-licensedb.aboutcode.org/). I will see if I can find that for all of them.

foersleo:
:arrow_up: Force push was just rebasing on top of develop before tending comments.

foersleo:
:arrow_up: Force push tended most comments by Ben. The only one I have not fixed yet is the one about the SPDX license specs because of the above mentioned reasons.

I went also ahead and removed unneeded license files in the patches in commit 2 of the series. That helps uncluttering the license directory in the installation and only ships the license files for firmware we actually ship.

packages/linux-firmware/linux-firmware.spec
foersleo: These files are actually not there I found. I had oriented myself on Fedora/AL2023 spec file for linux-firmware. Removed.

packages/linux-firmware/linux-firmware.spec
foersleo: I went for a blanket bulk include as suggested. Unfortunately the directory structure of linux-firmware is only existent for some vendors/firmware. For others it is a bit of a wild west setting. Going the bulk-include rule here puts more pressure on being clean in the patches, while keeping the rpm brief.

packages/linux-firmware/linux-firmware.spec
foersleo: I was wondering about that when looking at other specs, but muscle memory got the better of me. Fixed.

packages/linux-firmware/linux-firmware.spec
foersleo: That is a good question. There should not be a dependency of a certain firmware package state requiring a certain kernel version. The worst that could happen there would be additional firmware files the kernel does not know about. Or installing compressed firmware the kernel does not know how to decompress (See comment in spec file).

The other way around we may get into situations where a driver expects certain firmware files which are not there in which case we need to update linux-firmware.

If we want/need a hard dependency between firmware packages is debatable I guess. Maybe make it a rule to just update it in lockstep with major kernel updates is an option?

packages/linux-firmware/linux-firmware.spec
foersleo: So I looked at the `LICEN[CS]E` files we ship with what is proposed here. That is the following list with a mapping of what I think is the right License to specify if I have found a fitting specifier or leaving a comment on the state.

```
GPL-2 - clear what to do
GPL-3 - clear what to do
LICENCE.chelsio_firmware - LicenseRef-scancode-chelsio-linux-firmware 
LICENCE.myri10ge_firmware - seems to be no available
LICENCE.phanfw - LicenseRef-scancode-qlogic-firmware is identical
LICENSE.i915 - seems to be not available and not matching any other intel license
LICENSE.ice - seems to be not available and not matching any other intel license
LICENSE.ice_enhanced - seems to be not available and not matching any other intel license
```

Refering to the lists of Licenses in https://spdx.org/licenses/ and https://scancode-licensedb.aboutcode.org/

zmrow (approved):
From the variant perspective this looks good - I don't have any additional comments about the package or patches!

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,62 @@
...
+find . -type f -executable -exec chmod -x {} \;
+popd
+
+%files
+%dir %{fwdir}
+%{fwdir}/*
+%license LICENCE.* LICENSE.* GPL* WHENCE
+%{_cross_attribution_file}
+
+%changelog
```
yeazelm: We recently noticed we don't need to include this empty `%changelog` in our specfiles. So you can drop it if you'd like.

yeazelm (commented):
I'm happy with the outcome of trimming this size down. I am curious if we know how difficult it will be to keep up with these updates as things are added to the `linux-firmware` bundle upstream. Is there anything special we need to track to ensure updates don't accidentally start bloating with more things or will we just need to be diligent to check each update doesn't include things we don't need at the time of reviewing the update?

foersleo:
> I'm happy with the outcome of trimming this size down. I am curious if we know how difficult it will be to keep up with these updates as things are added to the `linux-firmware` bundle upstream. Is there anything special we need to track to ensure updates don't accidentally start bloating with more things or will we just need to be diligent to check each update doesn't include things we don't need at the time of reviewing the update?

That is a great question and I am a bit concerned about that myself. In all honesty I do not know yet. It looks like there is a release for linux-firmware about every 4-6 weeks (roughly once a month). I will have a stab at what it takes to update once the next release is out.

In the extreme case where these updates are too burdensome to do every month, we could just go with a linux-firmware update when we introduce a new major kernel. As discussed in a comment to Ben's question about dependency to specific kernel versions, the firmware images would be fine for the kernels we currently support. I do not expect major driver updates that would rely on newer firmware files (The file names are usually hard-coded in the drivers) within a stable series.

In an ideal world we would spend some time to build tooling to automatically strip down the files according to some kernel config file. Given the structure of the firmware repo I'd assume this to be quite challenging with lots of corner cases. But it is something we might want to try our hands on when we find some time

foersleo:
Firmware was confirmed to be at the right spot and the driver finding it properly by Max Böhm https://github.com/bottlerocket-os/bottlerocket/issues/3246#issuecomment-1656523329

Thanks a lot Max for helping out here!

packages/linux-firmware/linux-firmware.spec
markusboehme: It would be drivers checking for firmware versions and features, so using newer firmware with older drivers would work. In other cases, drivers would explicitly request a particular firmware version to be loaded (and older firmware files be kept around in new releases). Leo's suggestion of updating with the introduction of a new kernel series at the latest seems sound to me, but more frequently would be fine, too.

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,51 @@
...
+
+Source0: https://www.kernel.org/pub/linux/kernel/firmware/linux-firmware-%{version}.tar.xz
+
+%description
+This package includes firmware files required for some devices to operate.
+
+%prep
+tar -xof %{SOURCE0}
+%setup -TDn linux-firmware-%{version}
+%autopatch -p1
```
markusboehme: Nit: The sequence is a bit peculiar (telling `%setup` not to extract Source0 only to extract it manually before). This could become `%setup -n linux-firmware-%{version}`, or to also squash the patching, just `%autosetup -n linux-firmware-%{version} -p1`.

markusboehme:
I wanted to review those patches as applied to linux-firmware and there's a bit of an interesting problem there: They do apply fine for me via `patch` and `git apply`, but then I lose their history and commentary. When trying to apply them via `git am`, at least `0003-linux-firmware-bt-wifi-Remove-firmware-for-Bluetooth.patch` fails. The first failure is for the first line of the deleted `LICENSE.nxp` which does seem to use CRLF line endings. The patch seems to have converted those to just LF, and hence doesn't match. 

I know Git can perform conversion of line endings, but it seems unexpected here. Can you check whether the original patches as generated by `git format-patch` were correct and this only happened when committing the patch files to the Bottlerocket repo? If so, I wonder whether disabling EOL conversion for `*.patch` files via `.gitattributes` (by treating those as binary files) would be desirable.

For the review, I just pass `--ignore-whitespace` to `git am`.

markusboehme (commented):
Thanks for the diligent work combing through all this! I hope it all plays well with updates and won't need to be redone. :smiling_face_with_tear: 

I'm happy to approve once we reach a conclusion about the SPDX identifiers and the line endings (not caring at this point could be one).

foersleo:
> I wanted to review those patches as applied to linux-firmware and there's a bit of an interesting problem there: They do apply fine for me via `patch` and `git apply`, but then I lose their history and commentary. When trying to apply them via `git am`, at least `0003-linux-firmware-bt-wifi-Remove-firmware-for-Bluetooth.patch` fails. The first failure is for the first line of the deleted `LICENSE.nxp` which does seem to use CRLF line endings. The patch seems to have converted those to just LF, and hence doesn't match.
> 
> I know Git can perform conversion of line endings, but it seems unexpected here. Can you check whether the original patches as generated by `git format-patch` were correct and this only happened when committing the patch files to the Bottlerocket repo? If so, I wonder whether disabling EOL conversion for `*.patch` files via `.gitattributes` (by treating those as binary files) would be desirable.
> 
> For the review, I just pass `--ignore-whitespace` to `git am`.

The line endings seems to be an interesting issue with `git format-patch` and `git am`. This is **not** an artifact of adding patches to a separate repository. The original patches do not apply to a clean checkout/unpack of linux firmware when trying to apply via `git am` without extra flags. We can work around this by passing flags `--ignore-whitespace` (as pointed out by you), or `--keep-cr` which will apply the patches, but create a warning about quoted CRLF:

```
$ for p in ../../linux-firmware-20230625/*.patch; do git am --keep-cr $p; done
Applying: linux-firmware: snd: remove firmware for snd/audio devices
Applying: linux-firmware: video: Remove firmware for video/broadcast devices
warning: quoted CRLF detected
Applying: linux-firmware: bt/wifi: Remove firmware for Bluetooth/WiFi devices
Applying: linux-firmware: scsi: Remove firmware for SCSI devices
Applying: linux-firmware: usb: remove firmware for USB/Serial/PCMCIA devices
Applying: linux-firmware: ethernet: Remove firmware for ethernet/IB devices
Applying: linux-firmware: Remove firmware for Accelarator devices
Applying: linux-firmware: gpu: Remove firmware for GPU devices
Applying: linux-firmware: various: Remove firmware for various devices
Applying: linux-firmware: amd-ucode: Remove amd microcode
```

[Stackoverflow](https://stackoverflow.com/questions/6289001/git-am-format-patch-control-format-of-line-endings) has some more detail about what is happening and what git tries to do to work around this. The quoted commit message in [one of the answers](https://stackoverflow.com/a/28257267) is particularly enlightening about the odds and ins of CRLF. The gist of which is: `git-am` by default does not trust line endings coming in through mail because SMTP transport is CRLF-unsafe.

Unfortunately the proposed `--transfer-encoding` option seems to be not available with `git format-patch` so it seems there is not really anything we can do about this on the patch creation side.

packages/linux-firmware/linux-firmware.spec
foersleo: It is, and I believe I had a good reason when I first started down this journey of packaging linux-firmware. Not it is just lingering around from then without any reason. I confirmed that it can be simplified to the proposed shorter versions and creates an identical package (apart from expected differences like timestamps and buildids). I will fix that with a next iteration.

foersleo:
:arrow_up: Force push simplified `%prep` as suggested by @markusboehme and dropped `%changelog` as suggested by @yeazelm

packages/linux-firmware/linux-firmware.spec
foersleo: Done.

packages/linux-firmware/linux-firmware.spec
foersleo: I have spend some more time on this and found some scancode licenses that seem to match the ones for Intel (i915, ice, ice_enhanced). For the Myricom license there is have still not found a proper License identifier to put here without hesitancy.

Talked to Ben about this and will reach out to licensing experts in our OSPO to help resolve these questions and be on the safe side.

markusboehme:
@foersleo Thanks for looking closer into it! TIL about `git am --keep-cr` which seems to be the better alternative.

markusboehme (approved):

foersleo:
Converting this one to draft for now to reduce risk of accidental merge while I am sorting out the questions regarding the Licenses.

foersleo:
:arrow_up: Force push included fixing up the list of SPDX identifiers as discussed with OSPO and added the `force-upstream` flag to the package as introduced in #3333 .
Unfortunately I did mix in some catch up with the upstream develop branch and so the comparison to the previous state is littered with a lot of other changes.

Will do an additional rebase shortly to base it on top of current  develop branch.

packages/linux-firmware/linux-firmware.spec
foersleo: After reaching out to ospo and playing around some more with the scancode tooling I have come up with a list of SPDX license reference codes. I think we are fine with those, although I am not super happy about `LicenseRef-scancode-free-unknown` for the license of `tg3` as specified in `WHENCE`.

markusboehme (approved):
Still looking good to me with the updated SPDX license specifiers.

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,67 @@
...
+Patch0004: 0004-linux-firmware-scsi-Remove-firmware-for-SCSI-devices.patch
+Patch0005: 0005-linux-firmware-usb-remove-firmware-for-USB-Serial-PC.patch
+Patch0006: 0006-linux-firmware-ethernet-Remove-firmware-for-ethernet.patch
+Patch0007: 0007-linux-firmware-Remove-firmware-for-Accelarator-devic.patch
+Patch0008: 0008-linux-firmware-gpu-Remove-firmware-for-GPU-devices.patch
+Patch0009: 0009-linux-firmware-various-Remove-firmware-for-various-d.patch
+Patch0010: 0010-linux-firmware-amd-ucode-Remove-amd-microcode.patch
+
+%description
+This package includes firmware files required for some devices to operate.
```
arnaldo2792: nit: for consistency with other packages, use `%{summary}.`

packages/linux-firmware/linux-firmware.spec
```diff
@@ -0,0 +1,67 @@
...
+
+%build
+
+%install
+mkdir -p %{buildroot}/%{fwdir}
+mkdir -p %{buildroot}/%{fwdir}/updates
+
+# Here we have potential to shave off some extra space by using `install-xz` of
+# `install-zst` to compress firmware images on disk. However, that functionality
+# relies on kernels being configured with `CONFIG_FW_LOADER_COMPRESS_[ZSTD|XZ]`
```
arnaldo2792, line 53: Should we have an issue to track enabling this config?

packages/linux-firmware/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[package.metadata.build-package]
+package-name = "linux-firmware"
+
+[lib]
+path = "/dev/null"
```
arnaldo2792: We are not supposed to use this path, as explained in https://github.com/bottlerocket-os/bottlerocket/pull/3288

foersleo:
:arrow_up: tended the comments mentioned by Arnaldo.

In addition I removed the force-upstream falg from the linux-firmware package. Discussing it with Ben we should be fine hosting a copy in the look-aside cache and we want to do that for as many packages as possible.

packages/linux-firmware/Cargo.toml
foersleo: Thanks, I had not seen that one. Fixed.

packages/linux-firmware/linux-firmware.spec
foersleo: Fixed.

packages/linux-firmware/linux-firmware.spec
foersleo: https://github.com/bottlerocket-os/bottlerocket/issues/3359

foersleo:
Uploaded the source artifacts to lookaside cache and checked the automated test builds succeed. Merging.

PR #3299, opened 2023-07-27
tools/diff-kernel-config: various small improvements

foersleo:
**Issue number:** -

**Description of changes:**

Add three improvements (two minor, one bigger) to the diff-kernel-config tool.
The big interesting change enables the script to be resumed, which can save the time one would need for rebuilding things that already had been built successfully. This is particularly helpful when using the script for multiple variants where one variant may not build successfully, but others do. We can fix whatever fails that particular build and resume, keeping the state that finished successful previously.

**Testing done:**

I have used unpolished versions of these changes for previous kernel updates to cut down on time to redo builds.

**Terms of contribution:**

### DIFF
--- a/tools/diff-kernel-config
+++ b/tools/diff-kernel-config
@@ -31,7 +31,7 @@ bail() {
 
 usage() {
     cat <<EOF
-Usage: $0 -b GITREV_BEFORE -a GITREV_AFTER -o OUTPUT_DIR [-v VARIANT] [-h]
+Usage: $0 -b GITREV_BEFORE -a GITREV_AFTER -o OUTPUT_DIR [-v VARIANT] [-r] [-h]
 Compare kernel configurations before and after a series of commits.
 
     -a, --after         new Git revision to compare from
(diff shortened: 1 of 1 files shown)


tools/diff-kernel-config
```diff
@@ -206,7 +206,7 @@ for state in after before; do
                     ;;
             esac
 
-            kver_full=$(rpm --query --queryformat '%{VERSION}' "${kernel_rpm}")
+            kver_full=$(rpm --query --queryformat '%{VERSION}-%{RELEASE}' "${kernel_rpm}")
```
markusboehme, line 219: This makes sense for completeness. Only wondering out loud: This will report the release field of Bottlerocket's RPM and not the upstream one's, right? We would have to bump the release ourselves. I've seen we've at least done that for the 5.10 series at least once, but probably have not followed it closely (or the case hasn't come up yet for the 5.15 and 6.1 kernel series).

markusboehme (approved):

tools/diff-kernel-config
foersleo: That is correct. It probably does not make sense then as it is right now.
I will probably pull that commit from the PR then and think how we want to do this properly.

markusboehme: I'd still leave it in. We have bumped the release in the past and in the best case we'll be vigilant and do so again in the future when warranted. As an aside, I've been thinking about changes to our kernel version numbering scheme recently. I'll take this case into consideration.

bitcoinfinancier (approved):

bcressey (approved):

yeazelm (approved):

PR #3302, opened 2023-07-27
schnauzer: Implement v2 template renderer

cbgbt:
**Issue number:**
#3133

**Description of changes:**
This set of commits serves to implement the V2 schnauzer template renderer.

The renderer must dynamically fetch only a subset of available settings and helpers from the system prior to rendering any given template. This has been accomplished by creating `SettingsResolver` and `HelperResolver` interfaces which can provide the template renderer the needed information from the Bottlerocket API.

For now these interfaces
* use a static map of helpers from schnauzer (until they are implemented in settings extensions)
* fetch the full set of unversioned settings from the Bottlerocket API, then pare it down in-memory to the set of requested settings.

**Testing done:**
* All provided tests pass
* I also have a working `aws-k8s-1.24` variant which uses schnauzer v2 as the template renderer, which has passed basic testing and inspection.

**Terms of contribution:**


sources/api/schnauzer/src/v2/import/settings.rs
```diff
@@ -0,0 +1,327 @@
+//! This module contains the interface `SettingsResolver` for fetching settings from the Bottlerocket API.
+//!
+//! To do so we thinly wrap the apiclient to simplify retrieving JSON settings values.
+use super::as_std_err;
+use crate::v2::ExtensionRequirement;
+use async_trait::async_trait;
+use constants::API_SETTINGS_URI;
+use serde_json::{json, Map, Value};
+use snafu::OptionExt;
+use snafu::ResultExt;
```
zmrow: Can be combined: `use snafu::{OptionExt, ResultExt};`

sources/api/schnauzer/src/v2/import/mod.rs
```diff
@@ -0,0 +1,8 @@
+pub mod settings;
+
+/// Utility that Boxes an error type to be returned by a generic trait interface.
+///
+/// Intended to be called as e.g. `fallible().map_err(as_std_err)`
+fn as_std_err<'a, E: std::error::Error + 'a>(err: E) -> Box<dyn std::error::Error + 'a> {
```
zmrow, line 64: Nice!

sources/api/schnauzer/src/v2/import/settings.rs
```diff
@@ -0,0 +1,327 @@
...
+
+        if !code.is_success() {
+            return error::APIResponseSnafu {
+                method,
+                uri: SETTINGS_API_URI.to_string(),
+                code,
+                response_body,
+            }
+            .fail();
+        }
```
zmrow: Nit:  using `ensure!()` allows you to write a positive expression and avoid the `return...fail()` bit. 
```
ensure!(
    code.is_success(),
    error::...
);
```

sources/api/schnauzer/src/v2/import/settings.rs
```diff
@@ -0,0 +1,327 @@
...
+        },
+
+        #[snafu(display("Error {} when {}ing to {}: {}", code, method, uri, response_body))]
+        APIResponse {
+            method: String,
+            uri: String,
+            code: StatusCode,
+            response_body: String,
+        },
+        #[snafu(display(
```
zmrow: Nit: missing newline

sources/api/schnauzer/src/v2/import/settings.rs
```diff
@@ -0,0 +1,327 @@
...
+            "Error when fetching value for Bottlerocket setting '{}'",
+            setting_name
+        ))]
+        FetchSetting { setting_name: String },
```
zmrow: I don't think this variant is used anywhere?

sources/api/schnauzer/src/v2/import/settings.rs
```diff
@@ -0,0 +1,327 @@
...
+                &all_settings.as_object().unwrap(),
+                extension_requirements.into_iter(),
+            );
+            assert_eq!(
+                minimized_settings,
+                expected_settings.as_object().unwrap().clone()
+            );
+        }
+
+        // unimplemented!()
```
zmrow: I think this is leftover from dev time. :)

sources/api/schnauzer/src/v2/import/helpers.rs
```diff
@@ -0,0 +1,355 @@
...
+//! Currently the interface simply provides a static map of all existing helpers, but in the future this will be
+//! modified to query the settings extensions on the host.
+use super::as_std_err;
+use crate::{helpers as handlebars_helpers, v2::ExtensionRequirement};
+use async_trait::async_trait;
+use handlebars::{Handlebars, HelperDef};
+use maplit::hashmap;
+use snafu::OptionExt;
+use std::collections::HashMap;
+use std::collections::HashSet;
```
zmrow: Combine these imports

sources/api/schnauzer/src/v2/import/helpers.rs
```diff
@@ -0,0 +1,355 @@
...
+        let requested_helpers = &extension_requirement.helpers;
+
+        if requested_helpers.is_empty() {
+            return Ok(());
+        }
+
+        let existing_helpers = Self::fetch_helper_names_for_extension(setting_extension, version)?;
+        let existing_helpers: HashSet<_> = existing_helpers.iter().collect();
+
+        let requested_helpers: HashSet<_> = requested_helpers.iter().collect();
```
zmrow: Nit: I might convert this to a HashSet above rather than shadow the variable again down here.  The re-assign to the same variable name made me a bit confused and jump around.

sources/api/schnauzer/src/v2/import/helpers.rs
```diff
@@ -0,0 +1,355 @@
...
+        let missing_settings: Vec<_> = requested_helpers.difference(&existing_helpers).collect();
+        snafu::ensure!(
```
zmrow: _Super_ nit: import `ensure!()` rather than have the long name here.

sources/api/schnauzer/src/v2/import/helpers.rs
```diff
@@ -0,0 +1,355 @@
...
+
+pub mod error {
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub))]
+    pub enum Error {
+        #[snafu(display("No such setting extension '{}'", setting_extension))]
+        NoSuchSettingExtension { setting_extension: String },
+        #[snafu(display(
```
zmrow, line 228: Nit: newline. :)

sources/api/schnauzer/src/v2/template.rs
```diff
@@ -67,15 +74,52 @@ impl TemplateFrontmatter {
...
+    /// Performs parse-time validations of a `TemplateFrontmatter`.
+    fn validate(&self) -> Result<()> {
+        self.ensure_no_helper_names_collide()?;
+        Ok(())
+    }
+
+    /// Ensures imported helper names do not collide.
+    ///
+    /// While settings are namespaced in the templates (`settings.extension.key_of_interest`), helpers cannot be
+    /// similarly namespaced due to a bug in the `handlebars` crate which fails to parse helpers containing a `.`
```
zmrow, line 96: Is there an upstream issue we can link to here?

sources/api/schnauzer/src/v2/import/fake.rs
```diff
@@ -0,0 +1,81 @@
+//! Provides a helpful "fake" implementation of TemplateImporter to use in test scenarios.
+use async_trait::async_trait;
+use handlebars::Handlebars;
+
+use crate::v2::ExtensionRequirement;
+
+use super::{HelperResolver, SettingsResolver};
```
zmrow: Nit: We've typically been grouping all our imports together.  (`rust-analyzer` has a setting for this; it likes to split them up)

sources/api/schnauzer/src/v2/registry.rs
```diff
@@ -0,0 +1,62 @@
+//! This module contains utilities for constructing a Handlebars registry to render a given template.
+//!
+//! The registry is after inspecting the template frontmatter to determine which settings extensions and handlebar
```
zmrow: Should this read `The registry is built...`?

sources/api/schnauzer/src/v2/import/helpers.rs
cbgbt: This is so-written because the owned vector computed in the first line cannot be guaranteed to live long enough unless we assign it to a variable.

sources/api/schnauzer/src/v2/template.rs
cbgbt: So far as I know, I discovered this issue. Planning to open an issue but I also kinda want to see if I can submit a PR for it.

cbgbt:

sources/api/schnauzer/tests/v2_parse_fail_templates.rs
```diff
@@ -2,7 +2,9 @@ use schnauzer::v2::{template, Template};
 
 macro_rules! assert_parse_error {
     ($error:ident, $template:expr) => {
-        match include_str!($template).parse::<Template>() {
+        let parsed = include_str!($template).parse::<Template>();
```
webern, line 5: Have you checked that the panic stack produced by a failed test is easy to understand and trace to the right test?

sources/api/schnauzer/tests/v2_parse_success_templates.rs
```diff
@@ -164,9 +163,14 @@ fn succeeds_05_unambiguous_delims() {
...
-        },
+        }, ExtensionRequirement {
+            name: "helperutil".to_string(),
+            version: "v1".to_string(),
+            helpers: vec!["join_map".to_string()],
+        }
     };
-    let expected_body = "{{ settings.beagle.howl }}\n".to_string();
+    let expected_body =
+        "{{ join_map \"=\" \",\" \"fail-if-missing\" settings.beagle }}\n".to_string();
```
webern: Use Rust raw strings to make these more legible (though it get's a little weird with the newline).

```suggestion
       r# "{{ join_map "=" "," "fail-if-missing" settings.beagle }}
"#r.to_string();
```

sources/api/schnauzer/src/v2/import/helpers.rs
cbgbt: Oh, I'm realizing you might be talking about `requested_helpers` and not `existing_helpers`. Sure, I can change that.

sources/api/schnauzer/src/v2/import/helpers.rs
```diff
@@ -0,0 +1,355 @@
...
+        },
+
+        // globally helpful helpers will be included in a null extension called "helperutil"
+        "helperutil" => hashmap! {
+            "base64_decode" => helper!(handlebars_helpers::base64_decode),
+            "default" => helper!(handlebars_helpers::default),
+            "join_array" => helper!(handlebars_helpers::join_array),
+            "join_map" => helper!(handlebars_helpers::join_map),
+            "goarch" => helper!(handlebars_helpers::goarch),
+        },
```
bcressey: nit / bikeshed: I might like something stronger like `std` or `core` since we end up reserving this part of the settings namespace

sources/api/schnauzer/Cargo.toml
```diff
@@ -9,13 +9,18 @@ build = "build.rs"
...
+
 [dependencies]
 apiclient = { path = "../apiclient", version = "0.1" }
+async-trait = "0.1"
 base64 = "0.13"
 constants = { path = "../../constants", version = "0.1" }
 bottlerocket-release = { path = "../../bottlerocket-release", version = "0.1" }
 dns-lookup = "1"
-handlebars = "4"
+handlebars = "4.3"
```
webern: Why change this? You can update the version with `cargo update --package handlebars` instead if you needed a newer feature.

webern (commented):
testing looks great

sources/api/schnauzer/tests/v2_parse_fail_templates.rs
cbgbt: Yes! I it includes the test name and a printed version of the parse results, which typically include an error message with spans referring to the input.

sources/api/schnauzer/tests/v2_parse_success_templates.rs
cbgbt: I originally did this with raw strings and converted it to this because of the newline, though I was trying to use \n. I'll rewrite it to this.

sources/api/schnauzer/Cargo.toml
cbgbt: My understanding is that this would be the minimum allowable version, and this is the one that I developed against.

sources/api/schnauzer/src/v2/import/settings.rs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        // TODO: Modify this to use per-setting requests in the future.
+        let all_settings = Self::get_settings_json(&self.api_socket)
+            .await?
+            .as_object()
+            .cloned()
+            .context(error::NonJSONObjectSnafu {
+                key: "settings".to_string(),
+            })
+            .map_err(as_std_err)?;
```
bcressey: How expensive is this query and how many times do we invoke it on each boot? Can we safely cache the response and use it for multiple templates?

sources/api/schnauzer/src/v2/template.rs
cbgbt: https://github.com/sunng87/handlebars-rust/issues/595

sources/api/schnauzer/src/v2/import/settings.rs
cbgbt: It's definitely a significant increase in API calls compared to what we're doing now (once-per-thar-be-settings, instead of once-per-template). I'll modify this to cache the response, although we'll need to put some effort into parallelizing the apiserver and performing these queries in parallel when settings extensions are merged, since we can't fetch all values at once in that world.

cbgbt:
^ Addresses currently outstanding comments from @bcressey, @zmrow, and @webern.

cbgbt:
I didn't notice that `cargo deny` was failing in my workspace, then I mucked up and accidentally pushed the wrong branch to the PR. In any case, the above change adds the `cached` crate to `sources/clarify.toml`

bcressey (approved):

sources/api/schnauzer/Cargo.toml
webern: In general, my preference is to avoid unnecessary restrictions in the Cargo.toml versions because it can break a build (if, for example, a transient dependency had some reason that they needed to stay on 4.2).

sources/api/schnauzer/src/v2/import/mod.rs
```diff
@@ -0,0 +1,66 @@
+#[cfg(feature = "testfakes")]
+pub mod fake;
```
webern, line 2: Is it possible to use `cfg(test)`? I would want all of the tests to run all of the time. If some tests are not running without the feature flag then they will effectively be exempt from CI.

webern (approved):
I'm sorry I didn't have time for a closer reading of all the rust code, but this looks good.

sources/api/schnauzer/src/v2/import/mod.rs
cbgbt: None of the tests are hidden behind the feature flag, we enable the flag during testing. This feature flag is meant to allow external crates to use the fakes if they need to render templates during test-time as well.

sources/api/schnauzer/Cargo.toml
cbgbt: That's a good point, I've changed this.

cbgbt:
^ Moves `handlebars` requirement from `4.2` to `4`

It's hard to see because the change also rebases atop develop.

PR #3307, opened 2023-07-31
Update host containers for v1.14.3

markusboehme:
**Issue number:** n/a

**Description of changes:** Update the default admin and control container images to [v0.10.2](https://github.com/bottlerocket-os/bottlerocket-admin-container/pull/83) and [v0.7.3](https://github.com/bottlerocket-os/bottlerocket-control-container/pull/47), respectively.

**Testing done:** Container images have been tested as part of their respective release process. ~I will test the migrations and new defaults once the new images have been released. I'm marking this as a draft PR until then. Testing so far:~ The container images have been released and I got to test them via aws-k8s-1.23:

* Temporarily flipping `Release.toml`'s version to v1.14.3 and running `cargo make check-migrations`: all good (i.e. no obvious mistakes)
* `cargo make`: worked (i.e. TOML isn't broken, migrations do compile)
* An instance booting using a v1.15.0 AMI has the new host container images configured by default.
* A v1.14.2 AMI can be upgraded to v1.15.0 and then uses the new host container images.
* An updated v1.15.0 can be rolled back to v1.14.2 and then uses the old host container images.

I will cherry-pick this change to the `1.14.x` branch via a separate PR once this one is merged. We'll still need the 1.14.3 migrations in `develop`.

**Terms of contribution:**

By submitting this pull request, I agree that this contribution is dual-licensed under the terms of both the Apache License, version 2.0, and the MIT license.

PR #3309, opened 2023-07-31
[1.14.x] Add empty migration for 1.14.3

stmcginnis:
**Description of changes:**

Adds an empty migration definition to Release.toml to allow for a 1.14.3 release.

(cherry picked from commit f0ba96abb4ca5401ffad673b24e3b7864a549e3e)

**Terms of contribution:**

### DIFF
--- a/Release.toml
+++ b/Release.toml
@@ -1,4 +1,4 @@
-version = "1.14.2"
+version = "1.14.3"
 
 [migrations]
 "(0.3.1, 0.3.2)" = ["migrate_v0.3.2_admin-container-v0-5-0.lz4"]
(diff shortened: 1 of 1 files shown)


cbgbt (approved):

bcressey:
Do we still want to merge this? It won't be an empty migration after #3307 is cherry-picked over.

stmcginnis:
> Do we still want to merge this? It won't be an empty migration after #3307 is cherry-picked over.

Yeah, if we're backporting that change, then this one shouldn't be necessary. We'll just need to pick up the version number change.

I'll leave this for now and close once we get #3307 merged and backported.

markusboehme (approved):
I dropped the corresponding migration chain commit from #3307. I'm fine with merging this as-is. I think this should make things easier for everyone: The cherry-pick of #3307 requires no fixing, and this PR doesn't require any changes to still bump the version number.

PR #3312, opened 2023-08-03
Patch Docker host header fix for Go 1.20.6+ changes

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

Starting with Go 1.20.6, there was a security fix that changed the behavior when making local socket connections. It now requires a valid host header to be present, causing an error when connecting. This has since been [fixed upstream](https://github.com/moby/moby/issues/45935), but only in the newer versions of Docker.

This change backports the fix to our version of docker-engine and docker-cli.

**Testing done:**

Prior to applying this change, troubleshooting with the `docker` CLI on the console would give the error:

```sh
bash-5.1# docker run -it --rm --name testing fedora bash
http: invalid Host header
```

After applying, able to successfully using `docker` to run a container:

```sh
bash-5.1# docker run --rm -it fedora bash
[root@0f253dcfff6a /]#
```

Full ECS testing run by @arnaldo2792 and passed (thanks!!)

**Terms of contribution:**

### DIFF
--- a/packages/docker-cli/0001-non-tcp-host-header.patch
+++ b/packages/docker-cli/0001-non-tcp-host-header.patch
@@ -1,0 +1,121 @@
+From c85efebc0c9321fecc536485b47857dbcbe34ddf Mon Sep 17 00:00:00 2001
+From: Sebastiaan van Stijn <github@gone.nl>
+Date: Wed, 12 Jul 2023 15:07:59 +0200
+Subject: [PATCH 1/2] pkg/plugins: use a dummy hostname for local connections
+
+For local communications (npipe://, unix://), the hostname is not used,
+but we need valid and meaningful hostname.
+
+The current code used the socket path as hostname, which gets rejected by
+go1.20.6 and go1.19.11 because of a security fix for [CVE-2023-29406 ][1],
+which was implemented in  https://go.dev/issue/60374.
+
+Prior versions go Go would clean the host header, and strip slashes in the
+process, but go1.20.6 and go1.19.11 no longer do, and reject the host
+header.
+
+Before this patch, tests would fail on go1.20.6:
+
+    === FAIL: pkg/authorization TestAuthZRequestPlugin (15.01s)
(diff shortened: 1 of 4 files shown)


stmcginnis:
Not directly related, but also ran `sonobuoy run --mode certified-conformance` on an `aws-k8s-1.27` cluster to verify no issues having been built with Go 1.20.6.

```
20:51:51             e2e                                         global   complete   passed   Passed:378, Failed:  0, Remaining:  0
20:51:51    systemd-logs     ip-192-168-3-13.us-east-2.compute.internal   complete   passed
```

arnaldo2792 (approved):
ECS internal testing passed

etungsten (approved):

packages/docker-cli/0001-non-tcp-host-header.patch
```diff
@@ -0,0 +1,98 @@
+From c85efebc0c9321fecc536485b47857dbcbe34ddf Mon Sep 17 00:00:00 2001
+From: Sean McGinnis <stmcg@amazon.com>
+Date: Thu, 3 Aug 2023 13:32:10 +0000
+Subject: [PATCH] Backport https://github.com/moby/moby/pull/45942 to vendored
+ docker client
+
+---
```
yeazelm, line 30: nit: It would be good to include a bit more info in the commit messages about these changes. Just as much to just re-state the problem and solution so one doesn't need to read another GitHub url to understand our git message. 

```
Go 1.20.6 fixed an issue with headers which broke the client code since it didn't set a properly formatted
hostname. This adds a DummyHost for the headers on local communication which fixes local access on builds 
later than Go 1.20.6.
```
This way one knows immediately that we are unlikely to drop this patch moving forward just by reading the patch itself.

yeazelm (approved):

packages/docker-cli/0001-non-tcp-host-header.patch
stmcginnis: I copied over the upstream commit's description so we have the full context captured in the patch.

yeazelm (approved):

PR #3315, opened 2023-08-03
Migrate to schnauzer v2 to render configuration-files and generated settings

cbgbt:
---

**Issue number:**
#3133

**Description of changes:**
This change uses the new `schnauzer::v2` template renderer to replace all current uses of `schnauzer` in Bottlerocket.

Specifically, it:
* Implements `schnauzer-v2` settings generator binary
* Migrates `thar-be-settings` to use `schnauzer::v2`
* Migrates existing config templates to the new template syntax
* Modifies sundog to use a POSIX-shell-compatible lexer to tokenize settings generators
* Modifies `schnauzer`-based settings generators to use `schnauzer-v2`

**Thoughts on Migrations**
We overwrite the existing `setting-generator` for `schnauzer`-based generators, but leave the `templates` intact in the API for existing customers. This should enable rollbacks to succeed. Since the new setting generator is semantically identical to the previous one, we don't need to re-render the settings.

**What's Still Missing**
See https://github.com/bottlerocket-os/bottlerocket/issues/3133 for the list of tasks to complete this migration. This PR is missing:
* Replace [`ReplaceTemplateMigration`](https://github.com/bottlerocket-os/bottlerocket/blob/ef2fe106efd55728596c0d89cbfce53a58f79e61/sources/api/migration/migration-helpers/src/common_migrations.rs#L522)
* Write a migration guide to help users port existing templates

**Testing done:**
* [x] All unit/integration tests pass
* [x] Preliminary testing of `aws-k8s` and `aws-ecs` variants
* [ ] Test all variants
* [ ] Fuzz test new templates compared to old templates
* [ ] Migration testing

**Terms of contribution:**


cbgbt:
Sorry to immediately add churn here, but I'm going to split this into two smaller PRs.

cbgbt:
The first of the smaller PRs is available in https://github.com/bottlerocket-os/bottlerocket/pull/3316

PR #3316, opened 2023-08-03
Migrate to schnauzer v2 for generated settings

cbgbt:
---

**Issue number:**
#3133

**Description of changes:**
This change uses the new `schnauzer::v2` template renderer to replace all template-based settings generators in Bottlerocket.

Specifically, it:
* Implements `schnauzer-v2` settings generator binary
* Modifies sundog to use a POSIX-shell-compatible lexer to tokenize settings generators
* Modifies `schnauzer`-based settings generators to use `schnauzer-v2`

**Thoughts on Migrations**
We overwrite the existing `setting-generator` for `schnauzer`-based generators, but leave the `templates` intact in the API for existing customers. This should enable rollbacks to succeed. Since the new setting generator is semantically identical to the previous one, we don't need to re-render the settings.

**What's Still Missing**
See https://github.com/bottlerocket-os/bottlerocket/issues/3133 for the list of tasks to complete this migration.

**Testing done:**
* [x] All unit/integration tests pass
* [x] Preliminary testing of `aws-k8s` and `aws-ecs` variants
* [x] Migration testing (up & down)

**Terms of contribution:**


cbgbt:
^ I added a commit to this feature branch to enable GitHub workflows.

sources/api/schnauzer/src/bin/schnauzer-v2/clirequirements.rs
```diff
@@ -0,0 +1,172 @@
...
+                    name: "std".to_string(),
+                    version: "v1".to_string(),
+                    helpers: vec!["base64_decode", "join_array"]
+                        .into_iter()
+                        .map(String::from)
+                        .collect(),
+                },
+            ),
+            (
+                "weird-extension@but_valid1.23(1, 2,   3)",
```
bcressey: I don't feel strongly about this, but for CLI input I tend to avoid anything that might get mangled or misinterpreted if pasted into a prompt or script. Parens and whitespace jump out at me as slightly problematic.

Would something like `weird-extension@but_valid1.23+1,2,3` ("add the 1, 2, and 3 helpers from this version of this extension") support the requirements?

sources/api/schnauzer/src/lib.rs
```diff
@@ -1,3 +1,55 @@
...
+//! }
+//! ```
+//!
+//! ## The schnauzer Library
+//! The primary user interface is provided via `schnauzer::render_template` and `schnauzer::render_template_file`.
+//! These functions require the user to pass the template, as well as a `TemplateImporter`, which tells schnauzer how
+//! to resolve references to settings extensions and the requested helpers.
+//!
+//! Most users will want to use `schnauzer::BottlerocketTemplateImporter`, which uses settings extensions to resolve
+//! settings and helper data; however, custom `TemplateImporter` implementrations can be used as well.
```
bcressey: nit: "implementations"

sources/api/schnauzer/src/lib.rs
```diff
@@ -1,3 +1,55 @@
...
+//! For static datasets to be used for tests, enable the `testfakes` feature in `Cargo.toml`.
+//!
+//! ## The schnauzer Settings Generator
+//!
+//!
+//! ## schnauzer v1
+//! schnauzer was originally written to render simpler templates which always had the full scope of Bottlerocket
+//! settings and helper functions available to them, making it incompatible with the concept of Out-of-Tree Builds.
+//!
+//! The original schnauzer library deprecated, but continues to be made available under `schnauzer::v1` until it can
```
bcressey: nit: "is deprecated"

bcressey (approved):

sources/api/schnauzer/src/bin/schnauzer-v2/clirequirements.rs
cbgbt: It should, though I'm conflicted. I've been thinking about this syntax since I submitted this PR.

The `requires` line is meant to basically replicate the functionality of frontmatter. One thing it doesn't do here is allow the `optional = true` field, which is probably fine; however, I don't want to be trapped in a corner of never being able to support it.

I've been thinking we may need something like `extension@version(helpers=[helper1, helper2, ...])` so that we can add other named parameters if needed later. Of course, if the time comes, a regex would make for a terrible parser, but that's a bridge for another time.

In spite of all of this, we could probably still use different separators to avoid parens and whitespace, maybe something like:

```
extension@version+helpers:1,2,3+optional:true
```

What do you think?

cbgbt: Hmm, on the other hand, it's extremely unlikely that the templates themselves will be CLI-friendly, they almost always contain whitespace and curly brace characters.

I'm erring on continuing to use the parens (I personally find it more readable) but adding the named arguments.

cbgbt:
* Modified the CLI helper specification syntax to be more future-proof
* Fixed an issue where migrations would be built for all variants when they actually only apply to a subset.

cbgbt:
* Fixes README issues pointed out by @bcressey

etungsten (approved):
:rocket:

PR #3334, opened 2023-08-09
Add support for SeccompDefault setting for k8s 1.25+

### LINKED ISSUE #2742: kubelet >=v1.25: expose setting to enable the use of `RuntimeDefault` as the default seccomp profile for all workloads

cartrius-a:
**Issue number:**

Closes #2742 

**Description of changes:**

Enable **RuntimeDefault** as the default seccomp profile for all workloads via kubelet-configuration. This is disabled by default.

**Testing done:**

I launched a 1.14.x Bottlerocket AMI and verified `kubernetes.seccomp-default` setting was unavailable. After updating to a custom built Bottlerocket image with my changes, the setting could be set using `apiclient set settings.kubernetes.seccomp-default=true`. I verified **SeccompDefault** was equal to true using `kubectl get --raw /api/v1/nodes/<node>/proxy/configz`, meaning that **RuntimeDefault** was being used as the default seccomp profile for the workload. I also rolled back to the older Bottlerocket image through the Admin container and verified the setting was no longer available.

**Terms of contribution:**


jpculp (approved):

rpkelly (approved):

README.md
```diff
@@ -559,6 +559,7 @@ The following settings are optional and allow you to further configure your clus
 * `settings.kubernetes.provider-id`: This sets the unique ID of the instance that an external provider (i.e. cloudprovider) can use to identify a specific node.
 * `settings.kubernetes.registry-burst`: The maximum size of bursty pulls.
 * `settings.kubernetes.registry-qps`: The registry pull QPS.
+* `settings.kubernetes.seccomp-default`: Exposes the ability to enable RuntimeDefault as the default seccomp profile for all workloads via kubelet-configuration.
```
```suggestion
* `settings.kubernetes.seccomp-default`: Enable RuntimeDefault as the default seccomp profile for all workloads via kubelet-configuration. This is disabled by default.
```

packages/kubernetes-1.25/kubelet-config
```diff
@@ -131,6 +131,7 @@ featureGates:
   CSIMigrationvSphere: true
 protectKernelDefaults: true
 serializeImagePulls: false
+seccompDefault: {{default false settings.kubernetes.seccomp-default}}
```
etungsten, line 134: I'm somewhat conflicted on this. On one hand, not storing anything for `seccomp-default` by default and letting the kubelet config field be rendered out to false via the `default` handlebar helper is definitely simpler. But it's also nice to be able to query the Bottlerocket API to see whether `seccomp-default` is actually disabled or enabled. The API settings should probably be treated as declarative.

I'm leaning towards the `seccomp-default` be set to false by default explicitly and just render the value out as is here. Definitely would like more opinions on this.

etungsten (approved):
Great job! Just a few non-blocking comments.

packages/kubernetes-1.25/kubelet-config
jpculp: I would like to think that the Kubernetes default won't switch to _true_ in the middle of a release cycle. If for example they decide to make it _true_ for **1.30** we would already need to make a PR to make Bottlerocket's default _true_ as well. We could then set previous k8s variants explicitly to false in that same PR, but I could go either way if anyone has a strong feeling about doing it now.

cartrius-a:
Updated README.md per @etungsten's suggestion

packages/kubernetes-1.25/kubelet-config
stmcginnis: Very unlikely it would switch for a given release, more likely on a major.minor boundary.

This does mean we will need to watch out for that switching and update our default accordingly. It would be safer to not include the line unless it is explicitly set by the user. But I don't feel too strongly about it in this case.

jpculp: I went ahead and added a commit to explicitly set `settings.kubernetes.seccomp-default` to `false` in the datastore to make this more clear. Whenever we want to switch to `true` for the default, we can make a **kubernetes-seccomp-default-true.toml** and point newer variants to that instead.

etungsten (approved):
:1st_place_medal: 

Diff in models:
```
$ diff -r aws-k8s-1.24/ aws-k8s-1.25/
Only in aws-k8s-1.25/defaults.d: 54-kubernetes-seccomp-default-false.toml
[etung] in sources/models/src [⎇  HEAD](d73a991) [$?]via 🦀 v1.71.0 
$ diff -r aws-k8s-1.25/ aws-k8s-1.26/
Only in aws-k8s-1.26/defaults.d: 54-kubernetes-aws-external-cloud-provider.toml
[etung] in sources/models/src [⎇  HEAD](d73a991) [$?]via 🦀 v1.71.0 
$ diff -r aws-k8s-1.25/ aws-k8s-1.27/
Only in aws-k8s-1.27/defaults.d: 54-kubernetes-aws-external-cloud-provider.toml
Only in aws-k8s-1.27/defaults.d: 55-kubernetes-aws-credential-provider.toml
```

cbgbt (approved):

PR #3340, opened 2023-08-11
microcode: Update amd and intel ucode and remove indirection

foersleo:
**Issue number:** -

**Description of changes:**

```
Update microcode for Intel and AMD processors in face of recent
processor vulnerabilities. In the process cut out the indirection step
through Amazon Linux. They do not change the microcode between getting
it from upstream and us consuming it, so it is an extra step of
indirection.
```

There is one caveat to this change, and that is increased space needed in the images. Amazon Linux reduces the
provided Intel microcode down to the platforms it will encounter in EC2. We had carried that gap of not shipping
microcode for potential target platforms for a while now since we introduced metal variants and were in need to fix
that gap. However, now that we take the upstream Intel package completely we carry a lot more microcode binaries,
which leads to a quite steep increase in space needed (11M vs 235K for the intel microcode rpm size, see below).

But given that this is cutting out an indirection that can introduce additional head ache I think this is a worthwhile
tradeoff for closing the coverage gap. If we have to, we can later thin out the microcode we ship to not include super
binaries for obsolete technology or platforms that are unlikely to run Bottlerocket (Intel Atom platforms come to mind).

```
$ ls -ahl before/*x86*
-rw-r--r--. 1 fedora fedora 6.3K Aug 11 14:57 before/bottlerocket-microcode-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora  29K Aug 11 14:57 before/bottlerocket-microcode-amd-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora 9.1K Aug 11 14:57 before/bottlerocket-microcode-amd-license-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora 235K Aug 11 14:57 before/bottlerocket-microcode-intel-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora 8.2K Aug 11 14:57 before/bottlerocket-microcode-intel-license-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora 7.1K Aug 11 14:57 before/bottlerocket-microcode-licenses-0.0-1.x86_64.rpm
$ ls -ahl after/*x86*
-rw-r--r--. 1 fedora fedora 6.3K Aug 11 14:56 after/bottlerocket-microcode-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora  45K Aug 11 14:56 after/bottlerocket-microcode-amd-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora 9.2K Aug 11 14:56 after/bottlerocket-microcode-amd-license-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora  11M Aug 11 14:56 after/bottlerocket-microcode-intel-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora 8.2K Aug 11 14:56 after/bottlerocket-microcode-intel-license-0.0-1.x86_64.rpm
-rw-r--r--. 1 fedora fedora 7.1K Aug 11 14:56 after/bottlerocket-microcode-licenses-0.0-1.x86_64.rpm
```

**Testing done:**

Booted `aws-k8s-1.27` with the new microcode package on c5.metal host and checked microcode version. Expected according to [release notes](https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files/releases/tag/microcode-20230808) is `05003604` (Search for Processor `CLX-SP` for that platform):

```
bash-5.1# grep microcode /proc/cpuinfo 
microcode	: 0x5003604
microcode	: 0x5003604
microcode	: 0x5003604
microcode	: 0x5003604
microcode	: 0x5003604
[...]
```

**Terms of contribution:**

### DIFF
--- a/packages/microcode/Cargo.toml
+++ b/packages/microcode/Cargo.toml
@@ -11,9 +11,9 @@ path = "../packages.rs"
 # Use latest-srpm-urls.sh to get these.
 
 [[package.metadata.build-package.external-files]]
-url = "https://cdn.amazonlinux.com/blobstore/6d7f707779f6aff41c89bad00f7abe69dc70919cee29a8d3e5060f8070efe71d/linux-firmware-20200421-79.git78c0348.amzn2.src.rpm"
-sha512 = "d5a62eca6ddd7ff322574f17359681d03a733acc51c334127f291af5d5e39fcdf821c073ddcd977b2ca088cd95d35dc31db2001ca4c312a62dcbd4ea935434fd"
+url = "https://www.kernel.org/pub/linux/kernel/firmware/linux-firmware-20230804.tar.xz"
+sha512 = "b7fdffd49530223394a0896608a746395cbe9d1a3ca7e4e744bc8381e937845e085f08e2b56854a233426164072f4c365b281db2f0dbb47192a97a94ada8fae6"
 
 [[package.metadata.build-package.external-files]]
-url = "https://cdn.amazonlinux.com/blobstore/76e8f9f15ec2b27c70aff3ca15a28df51790b25c73fc8dc1bf1f28a9069b15e8/microcode_ctl-2.1-47.amzn2.0.9.src.rpm"
-sha512 = "e1347139d1edbd52d2619d970ba0f03500ba7367d071bb30ab3d209e44b3ff63000fcaa681f7352c79f7d5d2f0753130161b42b0eab7aab97b5b4fc4bfaa1b3b"
+url = "https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files/archive/refs/tags/microcode-20230808.tar.gz"
+sha512 = "8316eb9d35b315e630c6c9fab1ba601b91e72cc42926ef14e7c2b77e7025d276ae06c143060f44cd1a873d3879c067d11ad82e1886c796e6be6bf466243ad85b"
--- a/packages/microcode/Cargo.toml
+++ b/packages/microcode/Cargo.toml
@@ -8,7 +8,7 @@ build = "../build.rs"
 [lib]
 path = "../packages.rs"
 
-# Use latest-srpm-urls.sh to get these.
+# Check the two upstream repositories for the latest releases
 
 [[package.metadata.build-package.external-files]]
 url = "https://www.kernel.org/pub/linux/kernel/firmware/linux-firmware-20230804.tar.xz"
--- a/packages/microcode/microcode.spec
+++ b/packages/microcode/microcode.spec
@@ -3,8 +3,8 @@
 
 # These are specific to the upstream source RPM, and will likely need to be
 # updated for each new version.
-%global amd_ucode_archive linux-firmware-20200421.tar.gz
-%global intel_ucode_archive microcode-20210608-1-amzn.tgz
+%global amd_ucode_version 20230804
+%global intel_ucode_version 20230808
 
 Name: %{_cross_os}microcode
 Version: 0.0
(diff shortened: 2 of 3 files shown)


foersleo:
Cancelled the checks workflow. I have not uploaded the new sources to the lookaside cache for now. Given that we are changing who our upstream is, I did not want to go ahead and push it to the cache, just in case this gets vetoed.

packages/microcode/Cargo.toml
```diff
@@ -11,9 +11,9 @@ path = "../packages.rs"
 # Use latest-srpm-urls.sh to get these.
```
markusboehme: This comment is not accurate anymore, as is `latest-srpm-urls.sh`. I'd drop both for now.

packages/microcode/microcode.spec
```diff
@@ -110,6 +107,7 @@ install -p -m 0644 intel/intel-ucode/* %{buildroot}%{_cross_libdir}/firmware/int
 %dir %{_cross_libdir}/firmware
 %dir %{_cross_libdir}/firmware/intel-ucode
 %{_cross_libdir}/firmware/intel-ucode/??-??-??
+%{_cross_libdir}/firmware/intel-ucode/??-??-??_DUPLICATE
```
markusboehme: Those made me curious. According to [this](https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files/issues/68) and [this](https://github.com/intel/Intel-Linux-Processor-Microcode-Data-Files/issues/69) issue, they were (accidental?) commits of a previous microcode revision and to be ignored. We can do that by adding an `%exclude`.

markusboehme (commented):
I agree with the move to consume microcode straight from the upstream sources for recency and completeness. Since microcode updates are vendor-signed blobs, I don't see a benefit of using an intermediary as our upstream here.

Fun fact about the size considerations: Some Intel family/model/stepping combinations use the same microcode files:

```
$ sha256sum * | awk '$0 !~ /DUPLICATE/ { ucode[$1] += 1 } END { for (hash in ucode) { if (ucode[hash] > 1) { print ucode[hash] " " hash } } }'
2 5ea325315da51857c900a658590653a51e4ccf0c7a6d542a07459320f1e592d4
4 5d8d4a4d5456c43b7cc04937c80aec094ccbf3bd89f34ffa5182913ef944a9f9
4 3ecd7fa0f93cf78e3e3cac5af3225402d21fb528b1480a5363c556ffbd76b7b2
```

These turn out to be the chunkier ones even:

```
$ sha256sum * | grep -e 5d8d4a4d5456c43b7cc04937c80aec094ccbf3bd89f34ffa5182913ef944a9f9 -e 3ecd7fa0f93cf78e3e3cac5af3225402d21fb528b1480a5363c556ffbd76b7b2 -e 5ea325315da51857c900a658590653a51e4ccf0c7a6d542a07459320f1e592d4 | awk '!ucode[$1] { ucode[$1] = $2; system("ls -lh " $2) }'
-rw-rw-r-- 1 markus users 1.2M Aug  8 20:04 06-8f-04
-rw-rw-r-- 1 markus users 215K Aug  8 20:04 06-97-02
-rw-rw-r-- 1 markus users 211K Aug  8 20:04 06-ba-02
```

The size increase could therefore be counteracted somewhat (~ 5 MiB) by symlinking true duplicates. Another day... :-)

foersleo:
:arrow_up: force push fixed the two comments by Markus.

foersleo:
:arrow_up: force push rebased on top of latest develop branch.

Going to abort the checks running because I have not yet uploaded the artifacts to the lookaside cache.

packages/microcode/microcode.spec
foersleo: Ah thanks for digging up the context for these. When I created the commit I was somewhat pressed for time and only did the high level check and found them not to be identical so I took them in. Fixed now.

markusboehme (approved):

arnaldo2792 (approved):

foersleo:
Uploaded the source artifacts to lookaside cache and checked the automated test builds succeed. Merging.

PR #3343, opened 2023-08-12
Only limit concurrency while building RPMs

cbgbt:
**Description of changes:**
cargo make build-variant and build-package typically run several RPM builds concurrently, and those allow for their own levels of parallelism. 8 concurrent jobs (though unscientifically decided) still seems like a reasonable limitation there.

On the other hand, the rest of the build commands mostly focus on building a directory of rust sources. It shouldn't be a problem to let cargo use as many cores as it needs for those.

**Testing done:**
This seems to make `cargo make tuftool` and `cargo make publish-setup` faster when you don't have any of their dependencies' builds cached.

On my (admittedly beefy) machine:

Without change:
* `cargo make tuftool` 152 seconds
* `cargo make publish-setup-tools` 60 seconds

With change:
* `cargo make tuftool` 131 seconds
* `cargo make publish-setup-tools` 38.72 seconds

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -143,7 +143,8 @@ BUILDSYS_SDK_IMAGE = { script = [ "echo ${BUILDSYS_REGISTRY}/${BUILDSYS_SDK_NAME
 BUILDSYS_TOOLCHAIN = { script = [ "echo ${BUILDSYS_REGISTRY}/${BUILDSYS_SDK_NAME}-toolchain-${BUILDSYS_ARCH}:${BUILDSYS_SDK_VERSION}" ] }
 
 # Depends on ${BUILDSYS_JOBS}.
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):

yeazelm (approved):

zmrow (approved):

bcressey:
> On the other hand, the rest of the build commands are either focused on a single RPM, or on a directory of rust sources. It shouldn't be a problem to let cargo use as many cores as it needs.

I still think it makes sense to limit concurrency for `build-package` - building certain RPMs (glibc, util-linux) can drive a lot of subsequent rebuilds.

cbgbt:
* Kept the limitation on `cargo make build-package`
* Fixed the commit message

bcressey (approved):

PR #3353, opened 2023-08-21
avoid unmerged-usr taint for newer systemd

bcressey:
**Issue number:**
N/A

**Description of changes:**
systemd 252 expects these symlinks; otherwise, it sets the taint for "unmerged-usr".

**Testing done:**
Before the change, the taint is listed when the system is in a degraded state:
```
# systemctl status
● ip-10-0-60-129.us-west-2.compute.internal
    State: degraded
    Units: 314 loaded (incl. loaded aliases)
     Jobs: 0 queued
   Failed: 2 units
    Since: Mon 2023-08-21 22:30:39 UTC; 1min 13s ago
  systemd: 252
  Tainted: unmerged-usr
```

After, it is not listed.

**Terms of contribution:**

### DIFF
--- a/packages/filesystem/filesystem.spec
+++ b/packages/filesystem/filesystem.spec
@@ -34,11 +34,13 @@ mkdir -p %{buildroot}/media/cdrom
 mkdir -p %{buildroot}/root/.aws
 
 ln -s .%{_cross_prefix} %{buildroot}%{_prefix}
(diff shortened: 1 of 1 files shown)


yeazelm (approved):
🚢

cbgbt (approved):

PR #3355, opened 2023-08-22
update glibc to 2.38

bcressey:
**Issue number:**
#3170

**Description of changes:**

Housekeeping changes:
* Drop post-release patches for glibc 2.37, and add patches for 2.38.
* Keep the `localedef` hack working with one more hack.
* Remove unsupported configure options, and add new ones to disable
  building `nscd` rather than just excluding it from packaged files.
* Include libmvec for all architectures, since it's now supported on
  aarch64.

Functional changes:
* `--enable-multi-arch` to include microarch-optimized functions in
  the same DSO. This trades off a small amount of disk space and RAM
  to provide optimized versions of some string and memory functions.
  (This feature was disabled by accident in the original packaging.)
* `--enable-fortify-source` to build with the maximum supported level
  of `_FORTIFY_SOURCE` when possible. Currently, this is "2", but
  will be "3" when we update to GCC 12.

**Testing done:**
Tested AWS and VMware variants, with this change alone and in combination with other recent commits. Nodes all came up healthy and ran containers.

**Terms of contribution:**


stmcginnis (approved):
Did a quick sanity check. Built an `aws-k8s-1.27` variant and joined to an EKS cluster. Node joined and reported healthy. All `kube-system` pods looked good and were in a `Running` state. Ran sonobuoy quick test which passed.

Nothing stood out as odd or obviously wrong in the updated patches. Looks good to me.

packages/glibc/HACK-only-build-and-install-localedef.patch
```diff
@@ -37,8 +40,21 @@ index ac9455d6a7..da4e76523d 100644
...
++++ b/include/features.h
+@@ -471,7 +471,7 @@
+    to be disabled when building those functions, despite _GNU_SOURCE
+    being defined.  */
+ #if __GLIBC_USE (ISOC2X)
+-# define __GLIBC_USE_C2X_STRTOL 1
++# define __GLIBC_USE_C2X_STRTOL 0
+ #else
+ # define __GLIBC_USE_C2X_STRTOL 0
+ #endif
```
markusboehme, line 55: It's wondering whether this is the right fix that gave me pause. Diving into this was worth it as I learned about this peculiarity of our glibc build and cleared my doubts.

At first glance I thought this would mess with the entire resulting glibc, but from the logic in the spec file I gleaned that it only affects the build of `localedef` for the host architecture. Care is taken to revert the patch for the actual glibc build, so this is the right place to fix things up.

To check my understanding of the build failure this is working around: Since we compile `localedef` for execution on the build host, this is going to be linked against the host's/SDK's version of glibc. If this isn't glibc 2.38 yet, the linker will notice the missing `__isoc23` symbols. Redirecting to the pre-C23 versions of the `strtol`-class functions just for the `localedef` build means we get away with mixing glibc 2.38's `localedef` with a non-2.38 glibc in the SDK. Does that sound right?

packages/glibc/glibc.spec
```diff
@@ -13,19 +13,17 @@ Source3: ldconfig-service.conf
 # applied and reverted during the build.
 Source99: HACK-only-build-and-install-localedef.patch
 
-# Upstream patches from 2.36 release branch:
+# Upstream patches from 2.38 release branch:
```
markusboehme, line 16: Nit: In the meantime there have been a couple additional patches for malloc. They seem to be optimizations, though, so not critical to pick up right now.

markusboehme (approved):

packages/glibc/0003-nscd-Do-not-rebuild-getaddrinfo-bug-30709.patch
```diff
@@ -0,0 +1,185 @@
+From 6b99458d197ab779ebb6ff632c168e2cbfa4f543 Mon Sep 17 00:00:00 2001
+From: Florian Weimer <fweimer@redhat.com>
+Date: Fri, 11 Aug 2023 10:10:16 +0200
+Subject: [PATCH 3/6] nscd: Do not rebuild getaddrinfo (bug 30709)
+
+The nscd daemon caches hosts data from NSS modules verbatim, without
```
yeazelm, line 6: If we are not building with `nscd`, do we still need this patch?

bcressey: It's easier to maintain the patch series if we just take the recent upstream commits - per spec file instructions - unless there's something of specific concern.

bcressey:
⬆️ force pushes add recent patches (and fix a blunder with the local 9001 patch)

packages/glibc/glibc.spec
bcressey: I've added these to the stack.

packages/glibc/HACK-only-build-and-install-localedef.patch
bcressey: Yes, that's my understanding as well.

yeazelm (approved):

markusboehme (approved):

Issue #2823, opened 2023-02-23
`aws-iam-authenticator` for baremetal variant


rpkelly:
### Discussed in https://github.com/bottlerocket-os/bottlerocket/discussions/2809

<div type='discussions-op-text'>

<sup>Originally posted by **StefanTheWiz** February 15, 2023</sup>
Hi!

I'm trying to use AWS EKS as the control plane with bottlerocket baremetal nodes.
I was wondering if there was a particular reason why `aws-iam-authenticator` is not included in baremetal variant?

Based on the [README](https://github.com/bottlerocket-os/bottlerocket/blob/develop/README.md), I was kinda expecting that it would be included (for instance for using IAM Roles Anywhere).
> To start, we're focusing on the use of Bottlerocket as a host OS in AWS EKS Kubernetes clusters and Amazon ECS clusters.

As an experiment, I've built a version of baremetal that has the `aws-iam-authenticator` (diff [here](https://github.com/StefanTheWiz/bottlerocket/commit/65f1737fd37cbe0d12cb33a59de65a021cc37cc1)).
However, the update process might remove the authenticator, which might mean maintaining our own TUF repo and that would be a tough sell.</div>

rpkelly:
Documentation currently says this concerning authentication:

> settings.kubernetes.authentication-mode: Which authentication method the kubelet should use to connect to the API server, and for incoming requests. Defaults to aws for AWS variants, and tls for other variants.

It does not currently mention if  `aws` is not supported in those other variants, just that it's not the default.

bcressey:
The mechanics of adding the package are straightforward, but I need a better understanding of the security implications of this to form an opinion on whether this is a good idea. I'll reach out to some folks internally.

StefanTheWiz:
> The mechanics of adding the package are straightforward, but I need a better understanding of the security implications of this to form an opinion on whether this is a good idea. I'll reach out to some folks internally.

Hi. Any updates on this? :)

StefanTheWiz:
Still a blocker... 👎

stmcginnis:
Sorry for the delay on updates here. I know there are some possible security concerns with including `aws-iam-authenticator` for metal variants. I know there were some reviews being requested with some AWS security folks to provide more details, but unfortunately I don't have an update on where that process stands. I will reach out to some folks and see if we can get any updates.

cbgbt:
[...]

PR #3329, opened 2023-08-04
Add kubernetes-1.28 package and variants

etungsten:
**Issue number:**

Resolves https://github.com/bottlerocket-os/bottlerocket/issues/3274

**Description of changes:**
This adds the package for K8s 1.28 for building kubelet and new 1.28 Bottlerocket variants.

The new variants will use 6.1 kernel, include secure boot, and use XFS for the data volume filesystem.

**Testing done:**
- [x] aarch64 IPv4 conformance tests
- [x] aarch64 IPv6 conformance tests
- [x] aarch64 NVIDIA conformance tests
- [x] aarch64 NVIDIA smoke tests
- [x] x86_64 IPv4 conformance tests
- [x] x86_64 IPv6 conformance tests
- [x] x86_64 NVIDIA conformance tests
- [x] x86_64 NVIDIA smoke tests
```
$ testsys status 
 NAME                                                     TYPE   STATE     PASSED   SKIPPED   FAILED 
 beta-aarch64-aws-k8s-128-conformance-test                Test       passed      382      7007      2      
 beta-aarch64-aws-k8s-128-conformance-test-retry-1        Test       passed      4        7387      0      
 beta-aarch64-aws-k8s-128-instances                       Resource   completed                             
 beta-aarch64-aws-k8s-128-ipv6-conformance-test           Test       passed      381      7007      3      
 beta-aarch64-aws-k8s-128-ipv6-conformance-test-retry-1   Test       passed      4        7387      0      
 beta-aarch64-aws-k8s-128-ipv6-instances                  Resource   completed                             
 beta-aarch64-aws-k8s-128-nvidia-ipv6-conformance-test    Test       passed      384      7007      0      
 beta-aarch64-aws-k8s-128-nvidia-ipv6-instances           Resource   completed                             
 beta-aarch64-aws-k8s-128-nvidia-ipv6-nvidia-smoke-test   Test       passed      11       0         0      
 beta-x86-64-aws-k8s-128-conformance-test                 Test       passed      384      7007      0      
 beta-x86-64-aws-k8s-128-instances                        Resource   completed                             
 beta-x86-64-aws-k8s-128-ipv6-conformance-test            Test       passed      384      7007      0      
 beta-x86-64-aws-k8s-128-ipv6-instances                   Resource   completed                             
 beta-x86-64-aws-k8s-128-nvidia-ipv6-conformance-test     Test       passed      384      7007      0      
[...]

PR #3357, opened 2023-08-22
variants: add aws-iam-authenticator to metal variants

yeazelm:
**Issue number:**

Closes # [2823](https://github.com/bottlerocket-os/bottlerocket/issues/2823)

**Description of changes:**
AWS IAM Authentication can be set in metal variants right now but the binary is not included in the image causing it to fail. This adds the binary to the variants so that one can use AWS IAM Authentication instead of TLS.

**Testing done:**
Built a metal-k8s-1.23 image and confirmed the binary prints out usage:
```
/usr/bin/aws-iam-authenticator
A tool to authenticate to Kubernetes using AWS IAM credentials

Usage:
  aws-iam-authenticator [command]

Available Commands:
  add         add IAM entity to an existing aws-auth configmap
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  init        Pre-generate certificate, private key, and kubeconfig files for the server.
  server      Run a webhook validation server suitable that validates tokens using AWS IAM
  token       Authenticate using AWS IAM and get token for Kubernetes
  verify      Verify a token for debugging purpose
  version     Version will output the current build information

Flags:
  -i, --cluster-id ID                 Specify the cluster ID, a unique-per-cluster identifier for your aws-iam-authenticator installation.
  -c, --config filename               Load configuration from filename
      --feature-gates mapStringBool   A set of key=value pairs that describe feature gates for alpha/experimental features. Options are:
                                      AllAlpha=true|false (ALPHA - default=false)
                                      AllBeta=true|false (BETA - default=false)
  1 variants: add aws-iam-authenticator to metal variants
                                      ConfiguredInitDirectories=true|false (ALPHA - default=false)
                                      IAMIdentityMappingCRD=true|false (ALPHA - default=false)
  -h, --help                          help for aws-iam-authenticator
  -l, --log-format string             Specify log format to use when logging to stderr [text or json] (default "text")

Use "aws-iam-authenticator [command] --help" for more information about a command.
```
I don't have a cluster for metal where I could test that it works as intended but it should function the same as the aws variants today.

**Terms of contribution:**

### DIFF
--- a/variants/Cargo.lock
+++ b/variants/Cargo.lock
@@ -898,6 +898,7 @@ dependencies = [
 name = "metal-k8s-1_23"
 version = "0.1.0"
 dependencies = [
+ "aws-iam-authenticator",
  "cni",
  "cni-plugins",
  "kernel-5_10",
@@ -909,6 +910,7 @@ dependencies = [
 name = "metal-k8s-1_24"
 version = "0.1.0"
 dependencies = [
+ "aws-iam-authenticator",
  "cni",
  "cni-plugins",
  "kernel-5_15",
@@ -920,6 +922,7 @@ dependencies = [
 name = "metal-k8s-1_25"
 version = "0.1.0"
 dependencies = [
+ "aws-iam-authenticator",
  "cni",
  "cni-plugins",
  "kernel-5_15",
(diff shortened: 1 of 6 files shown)


arnaldo2792 (approved):

stmcginnis (approved):

etungsten (approved):
note to self: add this for `metal-k8s-1.28` in https://github.com/bottlerocket-os/bottlerocket/pull/3329

PR #3364, opened 2023-08-25
pluto: retry on empty private-dns-name from EC2

### LINKED ISSUE #3363: Add retries in `pluto` to handle eventual-consistent EC2 private DNS names
In rare cases, calls to EC2 DescribeInstance will return an empty string for the instance's private DNS name when the instance is newly launched. The API response eventually settles as the private DNS name of the instance becomes consistent. This poses a problem when `pluto` is determining the node name based on the private DNS name of the instance. 

This issue is described in more detail in  https://github.com/kubernetes/cloud-provider-aws/issues/635. The fix for the in-tree cloud provider is in https://github.com/kubernetes/kubernetes/pull/118421 which would cover all pre-1.27 clusters. 

However for 1.27+ clusters, we no longer use the in-tree cloud provider and depend on the external cloud-provider (See https://github.com/bottlerocket-os/bottlerocket/pull/3033). Therefore we need to add similar retry logic to fix this issue in `pluto`.

See https://github.com/awslabs/amazon-eks-ami/pull/1383 for the corresponding fix on the EKS optimized AMI side.

etungsten:
**Issue number:**

Resolves https://github.com/bottlerocket-os/bottlerocket/issues/3363

**Description of changes:**
```
    pluto: retry on empty private-dns-name from EC2
    
    Use fibonacci backoff on requests to EC2 for fetching the private DNS
    name of the instance. Retry on both API failures and on when the private
    DNS name is empty.

```

**Testing done:**
Instance comes up fine. Directly calling pluto works:
On an instance where hostname is resource-based in the subnet:
```
bash-5.1# pluto private-dns-name
"i-088c99da0374c92a5.us-west-2.compute.internal"
```

On an instance where hostname is IP based:
```
bash-5.1# pluto private-dns-name
"ip-192-168-21-104.us-west-2.compute.internal"
```

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/ec2.rs
+++ b/sources/api/pluto/src/ec2.rs
@@ -3,9 +3,15 @@ use crate::{aws, proxy};
 use aws_smithy_types::error::display::DisplayErrorContext;
 use snafu::{OptionExt, ResultExt, Snafu};
 use std::time::Duration;
+use tokio_retry::{
+    strategy::{jitter, FibonacciBackoff},
+    Retry,
+};
 
-// Limit the timeout for the EC2 describe-instances API call to 5 minutes
-const EC2_DESCRIBE_INSTANCES_TIMEOUT: Duration = Duration::from_secs(300);
+// Limit the timeout for fetching the private DNS name of the EC2 instance to 5 minutes.
+const FETCH_PRIVATE_DNS_NAME_TIMEOUT: Duration = Duration::from_secs(300);
+// Fibonacci backoff base duration when retrying requests
+const FIBONACCI_BACKOFF_BASE_DURATION_MILLIS: u64 = 200;
 
 #[derive(Debug, Snafu)]
 pub(super) enum Error {
@@ -21,8 +27,8 @@ pub(super) enum Error {
         >,
     },
 
-    #[snafu(display("Timed-out waiting for EC2 DescribeInstances API response: {}", source))]
-    DescribeInstancesTimeout { source: tokio::time::error::Elapsed },
+    #[snafu(display("Timed out retrieving private DNS name from EC2: {}", source))]
+    FetchPrivateDnsNameTimeout { source: tokio::time::error::Elapsed },
 
     #[snafu(display("Missing field '{}' in EC2 response", field))]
     Missing { field: &'static str },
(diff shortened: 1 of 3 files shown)


stmcginnis (approved):
Looks good!

sources/api/pluto/src/ec2.rs
```diff
@@ -51,26 +57,33 @@ pub(super) async fn get_private_dns_name(region: &str, instance_id: &str) -> Res
...
+                                    .first()
+                                    .and_then(|i| i.private_dns_name().map(|s| s.to_string()))
+                            })
+                        })
+                    })
+                    .filter(|private_dns_name| !private_dns_name.is_empty())
+                    .context(MissingSnafu {
+                        field: "Reservation.Instance.PrivateDNSName",
+                    })
+            },
```
bcressey, line 86: If there are valid scenarios where the EC2 PrivateDNSName might be empty and the node can still join a cluster, then I can't really endorse this since all those nodes would now take 5 minutes longer to start on first boot.

yeazelm (approved):

sources/api/pluto/src/ec2.rs
yeazelm: Can a node actually still join the cluster if this fails? If so, we likely need to re-evaluate our timeout and logic to not sit here for 5 minutes. I assumed this 5 minute timeout was more or less just there to let the cluster realize the node is not going to be healthy enough to connect to the cluster. Either way, 5 minutes still feels too long for this timeout.

etungsten:
Re-evaluating approach since there is potentially a valid combination of VPC configuration and kubelet/aws-iam-authenticator configurations where a node can still join a cluster where private DNS name is empty.

sources/api/pluto/src/ec2.rs
stmcginnis: I don't believe it can still join. The node's boot fails if this fails, so there doesn't seem to be much harm in waiting and giving things a chance to resolve.

I tried joining a node to a cluster in a VPC with DNS disabled. The instance was created and started to boot, but system console ended with:

```
[  312.455248] sundog[983]: Setting generator 'pluto private-dns-name' failed with exit code 1 - stderr: Timed out retrieving private DNS name from EC2: deadline has elapsed
[FAILED] Failed to start User-specified setting generators.
See 'systemctl status sundog.service' for details.
[DEPEND] Dependency failed for Send signal to CloudFormation Stack.
```

The control container never runs, so no chance to get in to the console to explore more. But the node is basically DOA at this point.

Interestingly, the `eksctl create nodegroup` timeout appears to be 25 minutes. So even after this node failed, without specifying a different timeout to the `eksctl` command it waited for ~20 more minutes before failing.

As a test, I provided a `settings.kubernetes.host-name` value in my user data. Since this was a random string, it still failed to join the k8s cluster. But the system log showed that it got past this point in initialization with no delay or other errors.

bcressey:
> Re-evaluating approach since there is potentially a valid combination of VPC configuration and kubelet/aws-iam-authenticator configurations where a node can still join a cluster where private DNS name is empty.

@etungsten did you determine that no such valid combination exists or would ever have worked on Bottlerocket previously? That's my current understanding.

bcressey (approved):

yeazelm (approved):

etungsten:
There is a very fringe case where a user can potentially set up a bootstrap container to set their own `hostname-override` setting after pluto has set the `hostname-override` setting to `""`. After merging this, to maintain that same workflow, a user would just have to set `""` or (a random string) for the `hostname-override` setting in userdata and let their bootstrap container do the overwrite,

PR #3366, opened 2023-08-25
Integrate `systemd-resolved` when using `systemd-networkd`

zmrow:
**Issue number:**
Related to #2449 

**Description of changes:**
This PR is one of, if not the final pieces of the puzzle to integrate `systemd-networkd` into Bottlerocket!

The set of commits here represent a slight change in direction for DNS resolution when using `systemd-networkd`.  Rather than use a custom DBUS listener ([whippet](#3270)) to react to interface changes and write a `resolv.conf`, we can rely on `systemd-resolved` to handle that.  These commits represent the steps needed to integrate `systemd-resolved` into Bottlerocket.

The first few commits bring in the `systemd-resolved` package and integrate it into the OS when using `systemd-resolved`.  There are a few additional mount units added, for directories needed early in boot for configuration files/drop-ins.

The next notable change (which can be updated when `wicked` is removed), is kubelet's configured `resolveConf` path.  Since we can't conditionalize this path based on the network backend, it now points at a symlink that `netdog` will manage and write early in boot.  For reasons described in [this k8s documentation](https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/#known-issues) we are unable to continue to point kubelet at `/etc/resolv.conf` when using **`systemd-resolved`.**
```
bash-5.1# ls -lah /run/netdog/                                                                                                                      
...                                                                                  
lrwxrwxrwx.  1 root root  32 Aug 25 19:40 resolv.conf -> /run/systemd/resolve/resolv.conf 
```

Another notable change is the addition of a oneshot systemd unit that runs `netdog write-network-status` after the network is up, but before other services.  This unit functions similarly as the existing `netdog install` hook that wicked calls, writing the current IP to file as well as handling the aforementioned `resolv.conf` symlink.  Since `systemd-networkd` does not provide hooks into the interface bringup process, we use this oneshot unit instead.

Finally, the last significant change (which probably contains the most code changes) is the handling of custom API settings for DNS name servers and search domains.  Bottlerocket calls a "restart command" `netdog write-resolv-conf`  every single boot when generating DNS configuration files, and calls the same command each time those settings change.  In this PR, that code has been updated when using `systemd-networkd`.  If there aren't any custom DNS settings, `netdog` does nothing and leaves all DNS handling to `systemd-resolved`, who gets DNS information from the DHCP lease.

If there are custom DNS settings, `netdog` will write drop-ins for each configured interface. 
 The drop-in will turn off the `UseDNS` or `UseDomains` setting for the interface, depending on the custom DNS settings.  If, for example, there are custom search domain settings, the drop-in will contain `UseDomains=false` for the interface, meaning it won't use domains acquired via DHCP lease.  Finally, `netdog` will drop a drop-in for `systemd-resolved`, using the custom DNS setting.
 
 An example of the interface drop-in when setting custom name servers and the corresponding `resolvectl` output showing overrides:
 ```
bash-5.1# cat /etc/systemd/network/10-eth0.network
[Match]
Name=eth0
[Link]
RequiredForOnline=true
RequiredFamilyForOnline=ipv4
[Network]
DHCP=yes
IPv6AcceptRA=true
IPv6DuplicateAddressDetection=0
[DHCPv4]
UseMTU=true
[DHCPv6]
WithoutRA=solicit
[IPv6AcceptRA]
UseMTU=true

bash-5.1# cat /etc/systemd/network/10-eth0.network.d/10-dns.conf 
[Network]
DNSDefaultRoute=false
[DHCPv4]
UseDNS=false
UseDomains=true
[DHCPv6]
UseDNS=false
UseDomains=true
[IPv6AcceptRA]
UseDNS=false
UseDomains=true

bash-5.1# resolvectl                                                             
Global
         Protocols: -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
  resolv.conf mode: stub
Current DNS Server: 1.2.3.4
        DNS Servers 1.2.3.4 192.168.0.2

Link 2 (eth0)
Current Scopes: none
     Protocols: -DefaultRoute +LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
    DNS Domain: us-west-2.compute.internal
...

bash-5.1# cat /etc/systemd/resolved.conf.d/10-resolv.conf 
[Resolve]
DNS=1.2.3.4 192.168.0.2
```

**Testing done:**
* Boot an aws-k8s-1.24 instance *using systemd-networkd* and validate the following:
  * The instance comes up and joins the cluster properly and can run pods
  * The resolveConf symlink is correct `resolv.conf -> /run/systemd/resolve/resolv.conf`
  * Set custom DNS settings for _just_ domains and servers, and also both, and observe the proper drop-ins being written for each interface as well as `systemd-resolved`
  * Ensure pods are able to be run before and after setting DNS settings via API (set to the exact same values) in order to test that the global resolved DNS information is used
* Boot an aws-k8s-1.24 instance *using wicked* and validate the same as above, except making sure that custom DNS settings end up in `/etc/resolv.conf`
  * Ensure `libnss_resolve.so` is not present on an instance using wicked and host resolution works ok with the new `nsswitch.conf`
 
* Following the ["Debugging DNS Resolution"](https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/#known-issues) instructions in the k8s docs, ensure DNS resolution is happy:
```
$ kubectl exec -i -t dnsutils -- nslookup kubernetes.default
Server:         10.100.0.10
Address:        10.100.0.10#53

Name:   kubernetes.default.svc.cluster.local
Address: 10.100.0.1
```

**Terms of contribution:**

### CODE REVIEW

**Comment by zmrow** [comment]
^ Rebase onto `develop`

**Comment by zmrow** [comment]
^ Fix clippy sadness

File: packages/os/etc-systemd-network.mount [hunk 1]
```diff
@@ -10,7 +10,7 @@ Wants=selinux-policy-files.service
 What=tmpfs
 Where=/etc/systemd/network
 Type=tmpfs
-Options=nosuid,nodev,noexec,noatime,context=system_u:object_r:lease_t:s0,mode=0755
+Options=nosuid,nodev,noexec,noatime,context=system_u:object_r:etc_t:s0,mode=0755
```
bcressey: The default label for everything under `/etc` is `etc_t`, so at this point it seems like this unit isn't needed.

packages/os/etc-systemd-resolved.conf.d.mount
```diff
@@ -0,0 +1,16 @@
...
+Conflicts=umount.target
+Before=local-fs.target umount.target
+After=selinux-policy-files.service
+Wants=selinux-policy-files.service
+
+[Mount]
+What=tmpfs
+Where=/etc/systemd/resolved.conf.d
+Type=tmpfs
+Options=nosuid,nodev,noexec,noatime,context=system_u:object_r:etc_t:s0,mode=0755
```
bcressey: Same for this unit - I don't think it's actually needed for overriding the SELinux context.

If it's really about racing with tmpfilesd for directory creation, then you can add `ExecStartPre=/usr/bin/mkdir ...` in a drop-in instead. That avoids the overhead that comes with adding a mount.

sources/api/netdog/systemd-derive/src/lib.rs
```diff
@@ -32,6 +32,9 @@ The `SystemdUnitSection` macro requires the following input parameters.
 - `section`: The name of the section this struct represents.  This parameter is set on the struct.
 - `entry`: The configuration entry name (which may be different than the struct member name).  This parameter must be set on each struct member.
 
+The `SystemdUnitSection` macro has the folowing optional input parameters:
```
bcressey: nit:
```suggestion
The `SystemdUnitSection` macro has the following optional input parameters:
```

sources/api/netdog/systemd-derive/src/lib.rs
```diff
@@ -32,6 +32,9 @@ The `SystemdUnitSection` macro requires the following input parameters.
...
+- `space_separated`: Only for use with `Vec`s, this boolean parameter will join the items in the `Vec` with a a space.
```
bcressey: nit:
```suggestion
- `space_separated`: Only for use with `Vec`s, this boolean parameter will join the items in the `Vec` with a space.
```

sources/api/netdog/systemd-derive/src/lib.rs
```diff
@@ -248,13 +253,23 @@ impl ToTokens for SystemdUnitSectionField {
...
+        let field = if self.space_separated {
+            quote! {
+                let joined = self.#struct_field_name.join(" ");
+                if !joined.is_empty() {
+                    write!(f, "{}={}\n", #systemd_entry_name, joined)?;
+                }
+            }
+        } else {
+            quote! {
+                // `Vec`s and `Option`s are both iterators, allowing us to write the same code for both
```
bcressey: Is there a way to do the right thing for `Option` as well? Otherwise it doesn't appear that the `space_separated` flag is actually restricted to the `Vec` case.

sources/api/netdog/systemd-derive/tests/tests.rs
```diff
@@ -76,3 +90,18 @@ Destination=11.0.0.1
 
     assert_eq!(n.to_string(), expected)
 }
+
+#[test]
+fn space_separated() {
```
bcressey: nit: add test for domain suffix spacing also since it's an important part of the external contract

sources/api/netdog/src/net_config/v1.rs
```diff
@@ -53,6 +53,14 @@ impl Interfaces for NetConfigV1 {
         !self.interfaces.is_empty()
     }
 
+    fn interfaces(&self) -> Vec<InterfaceId> {
+        let mut interfaces = Vec::with_capacity(self.interfaces.len());
+        for (name, _) in &self.interfaces {
+            interfaces.push(name.clone().into())
+        }
+        interfaces
```
bcressey: nit: would this be equivalent?

```suggestion
        self.interfaces.to_vec()
```

sources/api/netdog/src/main.rs
```diff
@@ -64,6 +63,17 @@ static SYSCTL_MARKER_FILE: &str = "/run/netdog/primary_sysctls_set";
 static LEASE_DIR: &str = "/run/wicked";
 static SYS_CLASS_NET: &str = "/sys/class/net";
 static SYSTEMD_SYSCTL: &str = "/usr/lib/systemd/systemd-sysctl";
+static RESOLV_CONF_SYMLINK: &str = "/run/netdog/resolv.conf";
+
+#[cfg(net_backend = "wicked")]
+static RESOLV_CONF: &str = "/etc/resolv.conf";
```
bcressey: I'd prefer these had more obvious names since from the name alone I'd assume `RESOLV_CONF` always meant `/etc/resolv.conf`.

Maybe:
* `REAL_RESOLV_CONF`:
  * -> `/etc/resolv.conf` (wicked)
  * -> `/run/systemd/resolve/resolv.conf` (systemd)
* `NETDOG_RESOLV_CONF` -> `/run/netdog/resolv.conf`

sources/api/netdog/src/cli/install.rs
```diff
@@ -53,10 +53,13 @@ pub(crate) fn run(args: InstallArgs) -> Result<()> {
...
         ) => {
             let lease = fetch_lease(&primary_interface, interface_type, args.data_file)?;
-            write_resolv_conf(&lease)?;
             write_current_ip(&lease.ip_address.addr())?;
+            write_primary_interface_sysctl(primary_interface)?;
 
-            write_primary_interface_sysctl(primary_interface)?
+            write_resolv_conf(&lease)?;
+            // Symlink resolv.conf to a common path, for compatibility reasons during
+            // systemd-networkd development
```
bcressey: I think this compat stuff ends up being explained too many times; it seems normal and even a desirable OS interface to just have `netdog` manage a `resolv.conf` symlink for any in-tree or out-of-tree unit that needs to refer to the "real" file.

sources/api/netdog/src/cli/install.rs
```diff
@@ -53,10 +53,13 @@ pub(crate) fn run(args: InstallArgs) -> Result<()> {
...
+            symlink_force(RESOLV_CONF, RESOLV_CONF_SYMLINK)?
```
bcressey: nit: to match the imperative naming of other functions called here:
```suggestion
            force_symlink(RESOLV_CONF, RESOLV_CONF_SYMLINK)?
```

packages/release/nsswitch.conf
```diff
@@ -1,4 +1,4 @@
 passwd: files
 group: files
 shadow: files
-hosts: files dns
+hosts: files resolve [!UNAVAIL=return] dns
```
bcressey, line 4: From the commit message:
> since we include libnss_resolve.so in the main systemd package

But we don't, right? It should only be in the resolved subpackage.

I would try to replace some of the "probably" language with more precise claims - it's either fine or it isn't, and it should be trivial to test.

sources/api/netdog/src/cli/mod.rs
```diff
@@ -165,6 +167,50 @@ where
     Ok(())
 }
 
+// Gather net config from possible sources, returning both the config and the source
+fn fetch_net_config() -> Result<(Option<Box<dyn Interfaces>>, PathBuf)> {
+    let over_ride = Path::new(OVERRIDE_NET_CONFIG_FILE);
```
bcressey: nit:
```suggestion
    let override = Path::new(OVERRIDE_NET_CONFIG_FILE);
```

sources/api/netdog/src/cli/mod.rs
```diff
@@ -165,6 +167,50 @@ where
...
+    let default = Path::new(DEFAULT_NET_CONFIG_FILE);
+    let cmdline = Path::new(KERNEL_CMDLINE);
```
bcressey: nit: this might eliminate the need for `to_path_buf()` calls later:
```suggestion
    let default = PathBuf::from(DEFAULT_NET_CONFIG_FILE);
    let cmdline = PathBuf::from(KERNEL_CMDLINE);
```

sources/api/netdog/src/cli/mod.rs
```diff
@@ -165,6 +167,50 @@ where
...
+
+    if Path::exists(over_ride) {
+        Ok((
+            net_config::from_path(over_ride)
+                .context(error::NetConfigParseSnafu { path: over_ride })?,
```
bcressey: nit: you could iterate over the first two paths to de-dupe / de-branch the code since the body is identical

sources/api/netdog/src/cli/write_network_status.rs
```diff
@@ -1,18 +1,16 @@
-use super::{error, primary_interface_name, write_primary_interface_sysctl, Result};
-use crate::dns::DnsSettings;
+use super::{error, primary_interface_name, symlink_force, write_primary_interface_sysctl, Result};
 use crate::networkd_status::NetworkDInterfaceStatus;
-use crate::CURRENT_IP;
+use crate::{CURRENT_IP, RESOLV_CONF, RESOLV_CONF_SYMLINK};
```
bcressey: If these constants are only used in this module, it might make sense to just define them here.

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+    let restart_resolved = Command::new(SYSTEMCTL)
+        .args(["restart", "systemd-resolved"])
+        .output()
+        .context(error::SystemctlExecutionSnafu)?;
+    ensure!(
+        restart_resolved.status.success(),
+        error::FailedSystemctlSnafu {
+            stderr: String::from_utf8_lossy(&restart_resolved.stderr)
+        }
+    );
```
bcressey: It looks like `networkd` supports a reload command, and we should use the "try" form in case the services aren't yet running (for whatever reason), and the `--no-block` flag unless it's really important that we synchronize the restart with the running lifetime of the `netdog` invocation for some reason.
```
systemctl try-reload-or-restart --no-block ...
```

sources/api/netdog/src/dns.rs
```diff
@@ -53,15 +60,17 @@ impl DnsSettings {
 
     /// Create a DnsSettings from TOML config file, supplementing missing settings from data in
     /// the NetworkDInterfaceStatus.
+    // TODO: Get rid of this constructor when wicked is gone.  The reason to have it is because in
+    // systemd-networkd land, it is significant and drives certain behavior if the DNS config is
+    // empty.  It is also important to have control over when we merge missing settings from the
+    // DHCP lease.
```
bcressey: This comment is extremely cryptic when attached to a function chain like this:
```
Self::from_config() -> Self::new_from_config() -> Self::from_config::impl(...)
```

It's not really clear to me why all these levels of indirection are necessary. Can `from_config` and `new_from_config` be merged? Can we make one of them unconditionally compiled?

Overall the call graph seems really hard to follow. Maybe having an inner impl function that the outer one calls into would be more clear than a set of sibling methods that are conditionally compiled.

sources/api/netdog/src/dns.rs
```diff
@@ -141,6 +152,79 @@ impl DnsSettings {
...
+}
+
+#[cfg(net_backend = "systemd-networkd")]
+impl ResolvedConfDropin {
+    fn from_dns_settings(dns: &DnsSettings) -> Self {
+        let domains = if let Some(domains) = &dns.search {
+            domains.clone()
+        } else {
+            Vec::new()
+        };
```
bcressey, line 187: Can we make the vec optional instead of passing around an empty vec?

sources/api/netdog/src/dns.rs
```diff
@@ -230,6 +314,26 @@ mod tests {
         }
     }
 
+    #[cfg(net_backend = "systemd-networkd")]
+    fn generate_networkd_interface_status_multiple_domains() -> NetworkDInterfaceStatus {
```
bcressey: nit: maybe a macro for this since it's pretty unwieldy to have dedicated functions to set up a config for a unit test that comes much later - it'd be easier to follow if the interface were defined via helper macro closer to where its properties were tested.

sources/api/netdog/src/dns.rs
```diff
@@ -141,6 +152,79 @@ impl DnsSettings {
...
+}
+
+#[cfg(net_backend = "systemd-networkd")]
+#[derive(Debug, SystemdUnitSection)]
+#[systemd(section = "Resolve")]
+struct ResolveSection {
+    #[systemd(entry = "DNS", space_separated)]
+    dns: Vec<String>,
+    #[systemd(entry = "Domains", space_separated)]
+    domains: Vec<String>,
```
bcressey: Are there types other than `String` we could use here?

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+use systemd_derive::{SystemdUnit, SystemdUnitSection};
 
 #[derive(FromArgs, PartialEq, Debug)]
 #[argh(subcommand, name = "write-resolv-conf")]
 /// Writes /etc/resolv.conf, using DNS API settings if they exist
 pub(crate) struct WriteResolvConfArgs {}
 
+/// A struct representing an interface drop-in that overrides DNS-via-DHCP settings
+// TODO: if we end up creating more interface drop-ins that use network settings, it's probably
+// worth making the notion of a drop-in a first class citizen in the networkd config module.
```
bcressey: I tend to dislike TODOs like this that try to direct future maintainers to do some refactoring that wasn't worth it at the time. The exception might be when it's obvious and worth doing but is blocked on some development elsewhere.

Arguably you have two drop-in cases now - one for `UseDNS` and another for `UseDomains`  - since they can be specified independently and could be maintained separately by this code.

(I don't think that refactoring the code this way would be helpful, though.)

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+        // If there are name server or search domain API settings, turn off the option to get these
+        // from the DHCP server.
+        let use_dns = match settings.has_name_servers() {
+            true => Some(false),
+            false => Some(true),
+        };
+        let use_domains = match settings.has_search_domains() {
+            true => Some(false),
+            false => Some(true),
+        };
```
bcressey:
```suggestion
        let should_use_dns_from_dhcp = match settings.has_name_servers() {
            true => Some(false),
            false => Some(true),
        };
        let should_use_domains_from_dhcp = match settings.has_search_domains() {
            true => Some(false),
            false => Some(true),
        };
```

bcressey: The `match` on a boolean is also a bit strange.

```suggestion
        let should_use_domains_from_dhcp = Some(!settings.has_search_domains());
        let should_use_dns_from_dhcp = Some(!settings.has_name_servers());
```
 
```

File: sources/api/netdog/src/cli/write_resolv_conf.rs [hunk 23]
```diff
@@ -1,28 +1,177 @@
...
+                use_dns,
+                use_domains,
+            }),
+        })
+    }
+}
+
+// Gather the DNS settings if they exist, and if so create interface level drop-ins to turn off
+// DNS-via-DHCP for either name servers or search domains.  Then write a systemd-resolved
+// configuration drop-in with the global DNS settings.
```
bcressey: This comment explains too much about what the code does in prose, and not enough about why.
```suggestion
```

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+//
+// For example, if we have DNS search domain settings from the API, we don't need to get them from
+// DHCP, so the interface drop-in will turn them off and systemd-resolved will use the global
+// setting from the API which will be set in the systemd-resolved configuration file.
```
bcressey:
```suggestion
// If we have DNS name servers or search domain settings from the API, we want to ignore the corresponding values in the DHCP lease.
// If we don't, then we want to use the values from DHCP.
// Toggle this functionality via a networkd interface drop-in.
```

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+#[cfg(net_backend = "systemd-networkd")]
+fn handle_dns_settings(_primary_interface: String) -> Result<()> {
+    // The DNS settings config file is always created since it is a configuration file in the API.
+    // If it is empty, don't create any interface drop-ins
+    let dns_settings = DnsSettings::from_config().context(error::GetDnsSettingsSnafu)?;
```
bcressey: The first comment is confusing, since it says the config file is always created - implying that this function will do it - but then we return early without creating any files if the settings are empty.

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+}
+
+#[cfg(net_backend = "systemd-networkd")]
+impl InterfaceDNSDropIn {
+    /// Using API DNS settings, create an appropriate drop-in for a network interface.
+    fn from_dns_settings(settings: &DnsSettings) -> Option<Self> {
+        // If there aren't any settings, don't write a drop-in
+        if !settings.has_name_servers() && !settings.has_search_domains() {
+            return None;
+        }
```
bcressey: It would be cleaner to always write a drop-in, and for this function to always return `Self` rather than an option.

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+    // The DNS settings config file is always created since it is a configuration file in the API.
+    // If it is empty, don't create any interface drop-ins
+    let dns_settings = DnsSettings::from_config().context(error::GetDnsSettingsSnafu)?;
+    let interface_drop_in = match InterfaceDNSDropIn::from_dns_settings(&dns_settings) {
+        Some(d) => d,
+        None => {
+            eprintln!("No DNS configuration to write");
+            return Ok(());
+        }
+    };
```
bcressey: It seems like a mistake to not always rewrite the drop-in with up-to-date values.

This logic would be broken if it were possible to delete settings, since then we could have a drop-in written by a previous iteration.

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+
+    // For each configured interface, create the drop-in directory and file It is possible for net
+    // config to be empty/None; if that is the case we need to fail here since we won't know which
+    // interfaces for which to write drop-ins.
```
bcressey: I am unhappy about failing in the empty / None case; ideally `netdog` would be fine with not configuring any interfaces.

Or put differently, if we're responsible for writing out a `resolv.conf` (because we have API settings) then we should always do that even if there are zero interfaces that need drop-ins.

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+    let (maybe_net_config, _) = fetch_net_config()?;
+    let net_config = maybe_net_config.context(error::MissingNetworkConfigSnafu)?;
+    for interface in net_config.interfaces() {
+        // Remove the colons since the ID might be a MAC address
+        let name = interface.to_string().replace(':', "");
+        let dropin_dirname = format!("10-{}.network.d", name);
```
bcressey: nit:
```suggestion
        let dropin_dir_name = format!("10-{}.network.d", name);
```

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+            return Ok(());
+        }
+    };
+
+    // For each configured interface, create the drop-in directory and file It is possible for net
+    // config to be empty/None; if that is the case we need to fail here since we won't know which
+    // interfaces for which to write drop-ins.
+    let (maybe_net_config, _) = fetch_net_config()?;
+    let net_config = maybe_net_config.context(error::MissingNetworkConfigSnafu)?;
+    for interface in net_config.interfaces() {
```
bcressey: Are `DHCPv4` and `DHCPv6` sections supported for all types of interfaces?

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+
+    // Write the systemd-resolved drop-in which will contain the DNS settings from the API
+    dns_settings
+        .write_resolved_dropin()
+        .context(error::ResolvConfWriteFailedSnafu)?;
+
+    //** Note: order of operations is important here!
+    // After all the above file writes have completed successfully, restart systemd-networkd and
+    // then systemd-resolved.  Restarting networkd first ensures that desired interface-level DNS
+    // settings are set before we restart systemd-resolved to pick up the global settings.
```
bcressey: Is the order really important? It seems like resolved and networkd should end up in agreement either way, and if they don't then there might be a race condition.

Just restarting the services one after another doesn't necessarily mean they are ready and will reach a consistent state; for example, if networkd is sending per-interface updates to resolved when it starts and then resolved goes away during that procedure, then networkd may not retry or continue for other interfaces.

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+
+// Gather the DNS settings if they exist, and if so create interface level drop-ins to turn off
+// DNS-via-DHCP for either name servers or search domains.  Then write a systemd-resolved
+// configuration drop-in with the global DNS settings.
+//
+// For example, if we have DNS search domain settings from the API, we don't need to get them from
+// DHCP, so the interface drop-in will turn them off and systemd-resolved will use the global
+// setting from the API which will be set in the systemd-resolved configuration file.
+#[cfg(net_backend = "systemd-networkd")]
+fn handle_dns_settings(_primary_interface: String) -> Result<()> {
```
bcressey: Somewhere it might be good to add a comment that this is a behavior change relative to the wicked backend, since we end up using resolved's per-interface DNS settings (if we don't have API settings) rather than just the primary interface's.

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,177 @@
...
+    // Write the systemd-resolved drop-in which will contain the DNS settings from the API
+    dns_settings
+        .write_resolved_dropin()
+        .context(error::ResolvConfWriteFailedSnafu)?;
+
+    //** Note: order of operations is important here!
+    // After all the above file writes have completed successfully, restart systemd-networkd and
+    // then systemd-resolved.  Restarting networkd first ensures that desired interface-level DNS
+    // settings are set before we restart systemd-resolved to pick up the global settings.
+    let restart_networkd = Command::new(SYSTEMCTL)
```
bcressey, line 156: nit: could use a macro

packages/os/write-network-status.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Write network status
+# Block manual interactions with this service, since it could leave the system in an
+# unexpected state
+RefuseManualStart=true
+RefuseManualStop=true
+Before=set-hostname.service
+After=network-online.target systemd-resolved.service
+Wants=network-online.target systemd-resolved.service
```
bcressey: This ends up running concurrently with `early-boot-config` and - if it took longer to finish - might not be complete before settings generators ran.

```
[    4.228716] localhost systemd[1]: Starting Bottlerocket userdata configuration system...
[    4.230046] localhost systemd[1]: Starting Write network status...
[    4.240167] localhost early-boot-config[1042]: 15:38:52 [INFO] early-boot-config started
[    4.240167] localhost early-boot-config[1042]: 15:38:52 [INFO] Retrieving platform-specific data
[    4.251987] localhost early-boot-config[1042]: 15:38:52 [WARN] No user data found via local file: /var/lib/bottlerocket/user-data.toml
[    4.260923] localhost early-boot-config[1042]: 15:38:52 [INFO] Received dynamic/instance-identity/document
[    4.261748] localhost early-boot-config[1042]: 15:38:52 [WARN] No user data found.
[    4.261748] localhost early-boot-config[1042]: 15:38:52 [INFO] Sending instance identity document to API
[    4.263129] localhost systemd[1]: Finished Write network status.
[    4.266859] localhost systemd[1]: Finished Bottlerocket userdata configuration system.
[    4.311108] localhost systemd[1]: Starting User-specified setting generators...
```

sources/api/netdog/src/cli/write_resolv_conf.rs
bcressey: To match the wicked behavior, we should set:
* if primary:
  * `UseDNS=yes` and `DNSDefaultRoute=yes` if no API name servers
  * `UseDomains=yes` if no API search domains
* otherwise:
  * `UseDNS=no`
  * `UseDomains=no`
  * `DNSDefaultRoute=no`

We should also set `UseDomains` / `UseDNS` in the `[IPv6AcceptRA]` section to the same values, in case the IPv6 RA client is used.

bcressey (commented):
With the other feedback addressed, relative to [the AL23 config](https://gist.github.com/bcressey/9d9bb6b6c085f0557c4523ffc20e52fa), we're missing these sections in the generated network unit.

```
# We're defaulting to a 1500 MTU otherwise, which is a serious problem.
# UseMTU=yes in [DHCPv4] and [IPv6AcceptRA] might be equivalent.
[Link]
MTUBytes=9001

# Speeds up DHCPv6 configuration
# see https://github.com/amazonlinux/amazon-ec2-net-utils/pull/34
[DHCPv6]
WithoutRA=solicit

# Not sure if this is needed.
# It looks like default routes are added for both DHCPv4 and IPv6 RA.
[Route]
Gateway=_ipv6ra
```

Here are [some very lightly tested patches](https://gist.github.com/bcressey/d1ddf280e0ca857ad4eaf4c752b7ecba).

The MTU-via-DHCP fix seems OK for Nitro instance types using ENA, but not sure if it does the right thing for instance types that use sriov (e.g. c3.large) and Xen netfront / vif (e.g. c1.xlarge). Please confirm that wicked also sets MTU based on DHCP, but assuming it does, this is the logical behavior until we make MTU configurable.

I confirmed that `WithoutRA=solicit` speeds up IPv6 configuration by 1-2s.

Disabling IPv6 DAD is also very good since without it we spend 1-2s extra waiting for eth0 to come online even when no IPv6 address is assigned and a IPv6 link-local address is used instead.

sources/api/netdog/src/cli/mod.rs
zmrow: `override` is a reserved Rust keyword so I had to add the underscore (and I didn't particularly like using a different word or the `r#` syntax)

packages/os/etc-systemd-resolved.conf.d.mount
zmrow: We found that using `ExecStartPre` to do the `mkdir` isn't tenable because `systemd-resolved` has `ProtectSystem=strict` in its unit file.  This setting mounts the entire filesystem read-only, preventing the `mkdir`.

We will opt to create these directories in `netdog` and avoid the units altogether.

sources/api/netdog/src/dns.rs
zmrow: Yes, I'll do that, thanks for catching that.  It's leftover from development.

sources/api/netdog/src/cli/write_resolv_conf.rs
zmrow: That makes sense.  Is there any specific place you'd like to see that?

I can add it to the docs around DNS settings, but it seems like there should be something more surfaceable.

Perhaps an announcement as a pinned issue whenever this goes out the door?

packages/os/write-network-status.service
zmrow: I'll add a `Before=early-boot-config.service` to avoid this.

sources/api/netdog/src/cli/write_network_status.rs
zmrow: The constants are currently used here as well as the `dns` module, so I'll keep them in `main` for now.

sources/api/netdog/src/net_config/v1.rs
zmrow: `interfaces` is a map, so I still need to iterate through them and grab just the key.  That being said, I'll use an iterator chain and make this a bit simpler and shorter

sources/api/netdog/src/cli/write_resolv_conf.rs
bcressey: If you make the change [I suggested below](https://github.com/bottlerocket-os/bottlerocket/pull/3366#discussion_r1306747957) then I think we'll match the current behavior.

sources/api/netdog/src/dns.rs
zmrow: We can use `IpAddr` for the name servers.  The search domains are deserialized from the config file into `DnsSettings` as a `Vec<String>` so that's what we have to work with for this type.

sources/api/netdog/src/dns.rs
zmrow: Not really in this case.  The derive macro operates on `Vec`s and `Option`s with values, not nested objects like `Option<Vec<_>>`. 

This was purposeful for these reason: I don't love creating an empty Vec, but I also feel that an empty `Vec` has the same significance in this case as `None`, and far easier to handle in code than a `Some` with an empty `Vec` inside of it.  (You don't need to check the option is `Some`, and also if the `Vec` is empty.)

zmrow: Oh - or were you suggesting that the field in `ResolveSection` be an `Option`?  Like:
```
ResolveSection {
    dns: Option<...>
}
```

Even then, I'd still prefer a `Vec` since the field needs to be iterated upon and space separated and `Option<String>` becomes hard to disambiguate.

But perhaps I'm missing your point.

sources/api/netdog/src/cli/write_resolv_conf.rs
zmrow: The sections are present for all interfaces - if that particular type of interface _should not_ have DHCP, it will have `DHCP=no` in the config and these drop-ins will not have any effect.

packages/os/etc-systemd-network.mount
zmrow: We will opt to create these directories in netdog and avoid the units altogether.

zmrow:
^ Addresses the first set of feedback from @bcressey .  I had comments or replies on a few of them; I left those open.  I also have not yet addressed the couple nits re: macros.

I'll follow up with the suggested network configuration changes from the second set of feedback ASAP.

sources/api/netdog/src/dns.rs
zmrow: I've made this much simpler and removed the indirection.  I think some of that was leftover from development time so I appreciate the callout.

sources/api/netdog/src/dns.rs
zmrow: These functions are actually going away.  We don't merge the interface DNS settings into the DNS settings anymore, so we don't need to generate this in tests at all.

etungsten (approved):

packages/release/nsswitch.conf
zmrow: Tested this on a `wicked` variant using `getent` - works like a charm!  Updated the commit message.

zmrow:
^ This addresses the remaining feedback from @bcressey .

* The code now always writes drop-ins for the interfaces, turning off DNS-via-DHCP for all interfaces except the primary iterface.  This mirrors the existing behavior when using `wicked`
* Integrates the missing settings from the AL2023 config (added via patches from @bcressey: thanks! )
* Updates all the reference test files to reflect the new settings

packages/os/os.spec
```diff
@@ -488,9 +489,9 @@ install -p -m 0644 %{S:121} %{buildroot}%{_cross_unitdir}
 %{_cross_attribution_vendor_dir}
 
 %if %{with systemd_networkd}
-%dir %{_cross_unitdir}/generate-network-config.service.d
-%{_cross_unitdir}/generate-network-config.service.d/requires-mounts-network-config.conf
-%{_cross_unitdir}/etc-systemd-network.mount
+%{_cross_unitdir}/write-network-status.service
+%dir %{_cross_libdir}/systemd/resolved.conf.d
+%{_cross_libdir}/systemd/resolved.conf.d/00-resolved.conf
```
bcressey: Can we put these in the `netdog` sub-package instead? It's weird to use the otherwise mostly-empty `os` package for this.

packages/os/write-network-status.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Write network status
+# Block manual interactions with this service, since it could leave the system in an
+# unexpected state
+RefuseManualStart=true
+RefuseManualStop=true
+Before=early-boot-config.service
+After=network-online.target systemd-resolved.service
+Wants=network-online.target systemd-resolved.service
```
bcressey: This still seems like a mistake - the network should be fully configured before we reach `network-online.target` since that is its contract.

IMO:
```
[Unit]
After=systemd-networkd-wait-online.service systemd-resolved.service
Wants=systemd-networkd-wait-online.service systemd-resolved.service

...
[Install]
WantedBy=network-online.target
```

Why do we need this to run after resolved though? Can you add a comment or drop the dependency if it's not really needed?

sources/api/netdog/test_data/networkd/network/vlancfgdev.network
```diff
@@ -8,3 +8,4 @@ DHCP=ipv4
 [DHCPv4]
 UseDNS=true
 UseDomains=true
```
bcressey: nit: can we drop this from the default generated `.network` files and leave it up to the drop-ins to write? I'd rather not set the value in one place and override it in another.

sources/api/netdog/src/cli/write_network_status.rs
```diff
@@ -24,12 +22,11 @@ pub(crate) fn run() -> Result<()> {
...
         .context(error::PrimaryInterfaceAddressSnafu {})?;
     write_current_ip(primary_ip)?;
-
-    // Write out resolv.conf
-    write_resolv_conf(&primary_link_status)?;
-
     write_primary_interface_sysctl(primary_interface)?;
 
+    // Symlink resolv.conf to a common path
+    force_symlink(REAL_RESOLV_CONF, NETDOG_RESOLV_CONF)?;
```
bcressey: I lost the thread here for a few minutes while comparing this to the wicked helper path.

Maybe a comment:
```suggestion
    // Symlink resolv.conf to a common path.
    // We don't ever write a resolv.conf when using networkd with resolved;
    // instead we write a drop-in configuration for resolved when the API
    // settings change.
    force_symlink(REAL_RESOLV_CONF, NETDOG_RESOLV_CONF)?;
```

sources/api/netdog/src/cli/mod.rs
```diff
@@ -165,6 +167,46 @@ where
     Ok(())
 }
 
+// Gather net config from possible sources, returning both the config and the source
+fn fetch_net_config() -> Result<(Option<Box<dyn Interfaces>>, PathBuf)> {
+    let over_ride = PathBuf::from(OVERRIDE_NET_CONFIG_FILE);
+    let default = PathBuf::from(DEFAULT_NET_CONFIG_FILE);
```
bcressey:
```suggestion
    let override_path = PathBuf::from(OVERRIDE_NET_CONFIG_FILE);
    let default_path = PathBuf::from(DEFAULT_NET_CONFIG_FILE);
```

`default` is also getting keyword highlighting in this view

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,187 @@
...
+        // existing wicked behavior by ensuring DNS configuration is sourced from settings and the
+        // primary interface only.
+        let mut should_use_dns_from_dhcp = Some(false);
+        let mut should_use_domains_from_dhcp = Some(false);
+        let mut should_be_dns_default_route = Some(false);
+
+        if is_primary {
+            should_use_dns_from_dhcp = Some(!settings.has_name_servers());
+            should_use_domains_from_dhcp = Some(!settings.has_search_domains());
+            should_be_dns_default_route = Some(!settings.has_name_servers());
```
bcressey: nit: separate bools are fine but maybe do this to indicate that they are intended to be the same value in this path:
```suggestion
            should_be_dns_default_route = should_use_dns_from_dhcp;
```

sources/api/netdog/src/cli/write_resolv_conf.rs
```diff
@@ -1,28 +1,187 @@
...
+    // For each configured interface, create the drop-in directory and file
+    let (maybe_net_config, _) = fetch_net_config()?;
+    if let Some(net_config) = maybe_net_config {
+        for interface in net_config.interfaces() {
+            let interface_drop_in =
+                InterfaceDNSDropIn::new(&dns_settings, interface.to_string() == primary_interface);
+
+            // Remove the colons since the ID might be a MAC address
+            let name = interface.to_string().replace(':', "");
+            let dropin_dir_name = format!("10-{}.network.d", name);
```
bcressey, line 133: We define our own convention here - removing the `:` from MAC-address network units? Just checking that we don't need to `systemd-escape` the name instead.

bcressey (approved):
Nice. Looks good except for the `write-network-status.service` dependencies which still make me nervous.

yeazelm (approved):
Looks like there needs to be another `cargo fmt` run to pass checks but otherwise looks good.

sources/api/netdog/src/cli/write_resolv_conf.rs
zmrow: We shouldn't need to keep the colons in the names - as long as the unit file and drop-in dir have the same name, networkd doesn't really care what they are called.

zmrow:
^ The above addressed the last bit of feedback from @bcressey .

It also adds an additional commit to remove `UseDNS` and `UseDomains` from the default generated `.network` files.

bcressey (approved):
🚀

yeazelm (approved):

PR #3375, opened 2023-08-29
readme: possible values for OCI-defaults rlimits

vyaghras:
**Issue number:**

Closes #

**Description of changes:**
- Update possible values of OCI defaults resource limits.
-  Update getrlimit link in readme

**Testing done:**
NA

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -752,9 +752,9 @@ capability | setting | default value
 
 ##### OCI Defaults: Resource Limits
 
-Each of the `resource-limits` settings below contain two numeric fields: `hard-limit` and `soft-limit`, which are **32-bit unsigned integers**.
+Each of the `resource-limits` settings below contain two fields: `hard-limit` and `soft-limit`.
 
-Please see the [`getrlimit` linux manpage](https://man7.org/linux/man-pages/man7/capabilities.7.html) for meanings of `hard-limit` and `soft-limit`.
+Please see the [`getrlimit` linux manpage](https://man7.org/linux/man-pages/man2/getrlimit.2.html) for meanings of `hard-limit` and `soft-limit`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -752,9 +752,9 @@ capability | setting | default value
 
 ##### OCI Defaults: Resource Limits
 
-Each of the `resource-limits` settings below contain two numeric fields: `hard-limit` and `soft-limit`, which are **32-bit unsigned integers**.
+Each of the `resource-limits` settings below contain two fields: `hard-limit` and `soft-limit`, which are **integers between -1 to I64::MAX inclusive** or **unlimited**.
```
stockholmux: Is this information repeated in [929](https://github.com/bottlerocket-os/bottlerocket/blob/fb831a78a6a794be82f4930628bd86dc374c3ebe/README.md?plain=1#L929)?

stockholmux (commented):
It looks like we kind of repeat the information on 755 on 929. Maybe we should just have 755?

It is after an extremely long table, so maybe we want both?

README.md
vyaghras: Yes, we are repeating  the same information, but line 929 make sense because after that line we are explaining the usage of those values. May be this line chan be changed to just Each of the `resource-limits` settings below contain two fields: `hard-limit` and `soft-limit`

stockholmux: > Each of the resource-limits settings below contain two fields: hard-limit and soft-limit

I like it.

```suggestion
Each of the `resource-limits` settings below contain two fields: `hard-limit` and `soft-limit`, which can be **integers between -1 to I64::MAX inclusive** or **the string "unlimited"**.
```

stockholmux (approved):
LGTM

bcressey (approved):

PR #3401, opened 2023-09-01
Deprecate log4j-hotpatch OCI hook

### LINKED ISSUE #3399: remove log4j hotpatch and deprecate the setting
**What I'd like:**
I’d like to remove the log4j hotpatch functionality from the distro. It’s been a while since CVE-2021-44228. The setting has always been disabled by default, and it’s not clear that anyone uses it. Getting rid of it means one less component to patch. It also lets us remove Java and maven from the SDK since we won’t have a JAR artifact to build.

Removing the actual setting would break in-place updates and launches for anyone that specifies this setting, so let’s not do that. Instead, it can be a no-op that logs a deprecation warning.

Keep it around forever. This is bad. It increases the surface area of the distro which means there’s more to patch. It also complicates testing since there are multiple “modes” that can be used to run containers.

arnaldo2792:
**Issue number:**

Closes #3399

**Description of changes:**

Remove the log4j hotpatch functionality and deprecate the setting. Clean up unused packages.

**Testing done:**

In `aws-ecs-1`:

- Migration testing from an adhoc build of 1.14.3 to 1.15, I validated that the deprecation message was logged, and on downgrade, I validated the hotpatch was still applied
- In a 1.15 build, I validated the deprecation message was logged, and the hotpatch wasn't applied
- In a running system, I validated the deprecation message was logged after the `log4j-hotpatch-enabled` settings was set to `true`

**Terms of contribution:**


README.md
```diff
@@ -696,7 +696,7 @@ For AWS variants, these settings allow you to set up CloudFormation signaling to
 Bottlerocket allows you to opt-in to use additional [OCI hooks](https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle) for your orchestrated containers.
 Once you opt-in to use additional OCI hooks, any new orchestrated containers will be configured with them, but existing containers won't be changed.
 
-* `settings.oci-hooks.log4j-hotpatch-enabled`: Enables the [hotdog OCI hooks](https://github.com/bottlerocket-os/hotdog), which are used to inject the [Log4j Hot Patch](https://github.com/corretto/hotpatch-for-apache-log4j2) into containers. Defaults to `false`.
+* `settings.oci-hooks.log4j-hotpatch-enabled`: Deprecated. This setting is no longer supported by Bottlerocket starting from v1.15.0. Though it is still available for backwards compatibility, enabling it has no effect beyond printing a deprecation warning to the system logs.
```
bcressey: nit: maybe make this bold
```suggestion
* `settings.oci-hooks.log4j-hotpatch-enabled`: **Deprecated**. This setting is no longer supported by Bottlerocket starting from v1.15.0. Though it is still available for backwards compatibility, enabling it has no effect beyond printing a deprecation warning to the system logs.
```

packages/release/release.spec
```diff
@@ -181,6 +185,7 @@ install -p -m 0644 %{S:204} %{buildroot}%{_cross_templatedir}/modprobe-conf
 install -p -m 0644 %{S:205} %{buildroot}%{_cross_templatedir}/netdog-toml
 install -p -m 0644 %{S:206} %{buildroot}%{_cross_templatedir}/aws-config
 install -p -m 0644 %{S:207} %{buildroot}%{_cross_templatedir}/aws-credentials
+install -p -m 0644 %{S:1302} %{buildroot}%{_cross_templatedir}/log4j-hotpatch-enabled
```
bcressey, line 188: It's a bit weird to have this in `release` for variants that may not have ever included the setting, but I can't suggest a better place offhand.

sources/api/migration/migrations/v1.15.0/deprecate-log4j-hotpatch-enabled/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2021"
+authors = ["Arnaldo Garcia Rincon <agarrcia@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers", version = "0.1.0"}
+
```
bcressey: nit: stray newline
```suggestion
```

sources/models/shared-defaults/oci-hooks.toml
```diff
@@ -1,13 +1,13 @@
...
-[services.oci-hooks]
-configuration-files = ["oci-hooks"]
-restart-commands = []
+[services.log4j-hotpatch-enabled]
+configuration-files = ["log4j-hotpatch-enabled-log-message"]
+restart-commands = ["systemctl restart deprecation-warning@log4j-hotpatch-enabled.timer"]
 
-[configuration-files.oci-hooks]
-path = "/etc/shimpei/shimpei-hooks.json"
-template-path = "/usr/share/templates/oci-default-hooks-json"
```
bcressey: Can we keep these defaults around for the top-level `oci-hooks` settings? Just in case we add another hook down the road. That seems like it'll be simpler.

If we drop these defaults, then I think we'd need a downward migration to put them back for older versions. Otherwise, the hooks won't work properly on downgrade because the new version won't have set these up in the datastore.

sources/models/shared-defaults/oci-hooks.toml
```diff
@@ -1,13 +1,13 @@
...
+[configuration-files.log4j-hotpatch-enabled-log-message]
+path = "/etc/deprecated-settings/log4j-hotpatch-enabled"
+template-path = "/usr/share/templates/log4j-hotpatch-enabled"
```
bcressey: These changes look good to me since they're isolated to the specific `log4j-hotpatch-enabled` setting.

sources/api/migration/migrations/v1.15.0/log4j-hotpatch-enabled-metadata/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2021"
+authors = ["Arnaldo Garcia Rincon <agarrcia@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers", version = "0.1.0"}
+
```
bcressey: nit: same extra newline at end

sources/models/shared-defaults/oci-hooks.toml
bcressey: The downgrade issue I'm thinking of would happen when launching a brand-new 1.15.0 instance and downgrading to 1.14.3.

An upgrade/downgrade cycle from 1.14.3 wouldn't trigger it, which is what it looks like you tested.

arnaldo2792:
Forced push includes:

- Bold format in README as suggested
- Removed new lines from `Cargo.toml`
- Keep `oci-hooks` defaults to generate ` `shimpei`s hooks

sources/models/shared-defaults/oci-hooks.toml
arnaldo2792: I'll test to confirm the downgrade cycle you discussed.

bcressey:
Tested the downgrade path for `aws-ecs-1` after launching a new image with the setting enabled.

Downgrade to 1.14.3:
```
Welcome to Bottlerocket OS 1.14.3 (aws-ecs-1)!
...
[    4.221136] migrator[444]: 00:51:16 [INFO] Running migration 'migrate_v1.15.0_deprecate-log4j-hotpatch-enabled.lz4'
[    4.346781] migrator[444]: 00:51:16 [INFO] Running migration 'migrate_v1.15.0_log4j-hotpatch-enabled-metadata.lz4'
...

bash-5.1# find /var/lib/bottlerocket/datastore/current/ -name '*log4j*'
/var/lib/bottlerocket/datastore/current/live/settings/oci-hooks/log4j-hotpatch-enabled

bash-5.1# cat /etc/shimpei/shimpei-hooks.json
{
  "hooks": {
    "prestart": [
      { "path": "/usr/libexec/hotdog/hotdog-cc-hook" }
    ],
    "poststart": [
      { "path": "/usr/libexec/hotdog/hotdog-poststart-hook" }
    ]
  }
}
```

Upgrade to 1.15.0 again:
```
Welcome to Bottlerocket OS 1.15.0 (aws-ecs-1)!
...
[    4.626589] migrator[443]: 00:54:45 [INFO] Running migration 'migrate_v1.15.0_log4j-hotpatch-enabled-metadata.lz4'
[    4.723218] migrator[443]: 00:54:45 [INFO] Running migration 'migrate_v1.15.0_deprecate-log4j-hotpatch-enabled.lz4'
...

bash-5.1# find /var/lib/bottlerocket/datastore/current/ -name '*log4j*'
/var/lib/bottlerocket/datastore/current/live/services/log4j-hotpatch-enabled
/var/lib/bottlerocket/datastore/current/live/configuration-files/log4j-hotpatch-enabled-log-message
/var/lib/bottlerocket/datastore/current/live/settings/oci-hooks/log4j-hotpatch-enabled.affected-services
/var/lib/bottlerocket/datastore/current/live/settings/oci-hooks/log4j-hotpatch-enabled

bash-5.1# cat /etc/shimpei/shimpei-hooks.json
{
  "hooks": {
    "prestart": [
    ],
    "poststart": [
    ]
  }
}

bash-5.1# cat /etc/deprecated-settings/log4j-hotpatch-enabled
DEPRECATED_SETTING=oci-hooks.log4j-hotpatch-enabled
```

bcressey (approved):

yeazelm (approved):

PR #3419, opened 2023-09-06
kernel: enable Intel VMD driver for metal variants

markusboehme:
**Issue number:** n/a

**Description of changes:** Enabling the Intel Volume Management Device driver for metal variants lets Bottlerocket boot on hosts that have a root disk in a separate PCI domain.

**Testing done:** A Siemens SIMATIC IPC BX-39A with an NVMe drive failed to boot without this, being unable to discover the partitions needed to assemble the dm-verity root. With this change the drive was found and the host booted.

`diff-kernel-config`:

```
config-aarch64-aws-dev-diff:              0 removed,   0 added,   0 changed
config-aarch64-aws-k8s-1.23-diff:         0 removed,   0 added,   0 changed
config-aarch64-aws-k8s-1.27-diff:         0 removed,   0 added,   0 changed
config-aarch64-metal-dev-diff:            0 removed,   0 added,   0 changed
config-x86_64-aws-dev-diff:               0 removed,   0 added,   0 changed
config-x86_64-aws-k8s-1.23-diff:          0 removed,   0 added,   0 changed
config-x86_64-aws-k8s-1.27-diff:          0 removed,   0 added,   0 changed
config-x86_64-metal-dev-diff:             0 removed,   0 added,   1 changed
config-x86_64-metal-k8s-1.23-diff:        0 removed,   0 added,   1 changed
config-x86_64-metal-k8s-1.27-diff:        0 removed,   0 added,   1 changed
```

```
[...removed variants without changes...]
==> /home/fedora/kernel-diff-vmd/config-x86_64-metal-dev-diff <==
 VMD n -> y

==> /home/fedora/kernel-diff-vmd/config-x86_64-metal-k8s-1.23-diff <==
 VMD n -> y

==> /home/fedora/kernel-diff-vmd/config-x86_64-metal-k8s-1.27-diff <==
 VMD n -> y
```

**Terms of contribution:**

### DIFF
--- a/packages/kernel-5.10/config-bottlerocket-metal
+++ b/packages/kernel-5.10/config-bottlerocket-metal
@@ -106,3 +106,7 @@ CONFIG_SCSI_SMARTPQI=y
 
 # Support for virtio scsi boot devices for other cloud providers
 CONFIG_SCSI_VIRTIO=y
+
+# Intel Volume Management Device driver, to support boot disks in a separate
+# PCI domain.
+CONFIG_VMD=y
--- a/packages/kernel-5.15/config-bottlerocket-metal
+++ b/packages/kernel-5.15/config-bottlerocket-metal
@@ -143,3 +143,7 @@ CONFIG_MOUSE_PS2=m
 # CONFIG_MOUSE_PS2_SENTELIC is not set
 # CONFIG_MOUSE_PS2_TOUCHKIT is not set
 # CONFIG_MOUSE_PS2_FOCALTECH is not set
+
+# Intel Volume Management Device driver, to support boot disks in a separate
+# PCI domain.
+CONFIG_VMD=y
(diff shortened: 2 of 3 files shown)


stmcginnis (approved):
Looks good to me. In case anyone else is curious:

```
Intel® VMD is the new way to configure 11th Generation and greater Intel® Core™ Processor-based platforms for Intel® RST management of RAID and Intel® Optane™ memory volumes.
```

I wonder if we need to also include their [RST driver](https://www.intel.com/content/www/us/en/support/articles/000057787/memory-and-storage/intel-optane-memory.html), but based on the test results it looks like this at least addresses the current use case. So a definite improvement!

foersleo (approved):

foersleo:
> I wonder if we need to also include their [RST driver](https://www.intel.com/content/www/us/en/support/articles/000057787/memory-and-storage/intel-optane-memory.html), but based on the test results it looks like this at least addresses the current use case. So a definite improvement!

I am not quite sure if Intel has released different technologies under the name RST. Rapid Storage Technology (as in the linked document) and Rapid Start Technology (as the `INTEL_RST` driver that is available in Linux since 5.15). From a glance they do not seem to be the same, with one being the firmware based RAID solution also covered by VMD and one being a hybrid sleep system to do firmware assisted hibernation like sleep (?).

Either way the document linked claims that the Rapid Storage Technology one is not available for Linux. So probably something we should have a closer look at to confirm whether we can and should or can not add that to Bottlerocket. Agreed that is no blocker for this one change here.

yeazelm (approved):

arnaldo2792 (approved):

PR #3420, opened 2023-09-06
fix clippy lints for netdog

bcressey:
**Issue number:**
N/A

**Description of changes:**
Fixes all current lints for `netdog`.

**Testing done:**
```
❯ cargo clippy
   Compiling imdsclient v0.1.0 (/home/fedora/brdev/sources/imdsclient)
    Checking netdog v0.1.0 (/home/fedora/brdev/sources/api/netdog)
    Finished dev [unoptimized + debuginfo] target(s) in 4.08s

❯ cargo clippy --tests
    Checking netdog v0.1.0 (/home/fedora/brdev/sources/api/netdog)
    Finished dev [unoptimized + debuginfo] target(s) in 3.92s
```

**Terms of contribution:**


bcressey:
Unfortunately, we don't run `cargo clippy --tests` in CI, and we also don't run `cargo clippy` for non-default features like the systemd-networkd support. That's why these made it through the earlier PRs.

stmcginnis (approved):
Great cleanup!

zmrow (approved):
Thanks!

yeazelm (approved):
This removes some of the code we no longer need since we decided to go with resolved but we also don't need `sources/api/netdog/src/cli/primary_interface.rs` and the code it brings in either so we should probably clean that up as well.

yeazelm (approved):
Nice!

zmrow (approved):
Thanks!

PR #3432, opened 2023-09-08
nvidia-variants: fix deleted NVIDIA char devices

### LINKED ISSUE #3072: NVIDIA char devices deleted after a while or after systemctl reload
**Image I'm using:**
Any NVIDIA variant

The NVIDIA char devices shouldn't be deleted after `systemctl reload`. This is related to a known bug in NVIDIA container toolkit (see https://github.com/NVIDIA/nvidia-docker/issues/1671). 

**What actually happened:**
I wasn't able to execute `nvidia-smi` after I ran `systemctl daemon-reload` 

```bash
bash-4.2# nvidia-smi
Failed to initialize NVML: Unknown Error
bash-4.2#
```

In any NVIDIA variant, deploy a pod that requests NVIDIA GPUs. Run `systemctl daemon-reload`, and then attempt to run `nvidia-smi` from the pod. You will see the error message above.

As mentioned, this is a known issue. The solution is to run `nvidia-ctx`, provided by the `nvidia-container-toolkit`, after boot or with a udev rule. Additionally, the k8s-device-plugin must pass `--pass-device-specs`, as stated [here]

[here]: https://github.com/NVIDIA/nvidia-docker/issues/1671#issuecomment-1532305701

arnaldo2792:
**Issue number:**

Closes #3072

**Description of changes:**

After some time, or by executing `systemctl daemon-reload`, the NVIDIA char devices are deleted from the containers, and tools like `nvidia-smi` stop working. In #3072, it was suggested that the issue should be gone if systemd `240` was used in conjunction with runc >= `1.1.7`. I built Bottlerocket from the tip of develop (since now we include systemd `252` and runc `1.1.9`) to confirm if the problem was fixed, but it wasn't. I even tried with a variant with cgroups v2 enabled, since according to [this comment], that could fix the problem, but it didn't.

I followed what was suggested in [this NOTICE], specifically, the steps under “For K8s environments”, which include:

- Create a `udev` rule that calls `nvidia-ctx` to create the char devices
- Per [this suggestion], pass `pass-device-specs` to the NVIDIA device plugin

According to the NVIDIA device plugin documentation, `pass-device-specs` “exists for the sole purpose of allowing the device plugin to interoperate with the CPUManager in Kubernetes. Setting this flag also requires one to deploy the daemonset with elevated privileges, so only do so if you know you need to interoperate with the CPUManager.”

We don't deploy the NVIDIA device plugin as a daemonset, rather it is executed as a system service. Thus, the service doesn't need additional privileges to work with the new flag.

[this comment]: https://github.com/NVIDIA/nvidia-docker/issues/1671#issuecomment-1236824748 
[this NOTICE]: https://github.com/NVIDIA/nvidia-docker/issues/1730
[this suggestion]: https://github.com/NVIDIA/nvidia-docker/issues/1671#issuecomment-1532305701

**Testing done:**

- In `aws-k8s-1.27`, I deployed a pod with one GPU, executed `systemctl daemon-reload`, and confirmed that `nvidia-smi` from the pod still worked
- In `aws-k8s-1.28`, I deployed a pod with one GPU, executed `systemctl daemon-reload`, and confirmed that `nvidia-smi` from the pod still worked

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
+++ b/packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
@@ -16,6 +16,7 @@ Source0: https://%{goimport}/archive/v%{gover}/nvidia-container-toolkit-%{gover}
 Source1: nvidia-container-toolkit-config.toml
 Source2: nvidia-container-toolkit-tmpfiles.conf
 Source3: nvidia-oci-hooks-json
+Source4: nvidia-gpu-devices.rules
 
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{_cross_os}libnvidia-container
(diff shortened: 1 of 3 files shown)


bcressey (approved):

yeazelm (approved):

arnaldo2792:
Just for my own sanity and for prosperity, I checked what the GPU operator does, and in the latest manifests it passes the flag as well:

https://github.com/NVIDIA/gpu-operator/blob/927043abb9d594f278e748048cdf42e9a72d5f0b/bundle/1.6.2/manifests/gpu-operator.clusterserviceversion.yaml#L33

So there shouldn't be any unexpected surprises.

kolyshkin:
@arnaldo2792 please file a runc bug.

PR #3434, opened 2023-09-08
start-local-vm: allow resizing the extracted images before boot

markusboehme:
**Issue number:** n/a

**Description of changes:** Sometimes it is helpful to not go with the default image sizes for testing purposes--in practice, this would mean testing with bigger images. Introduce the `--boot-image-size` and `--data-image-size` options to `start-local-vm` that allow for resizing the respective images after they have been extracted, but before they are booted.

**Testing done:**

Without `--boot-image-size`, the regular metal-dev disk image size is used:

```
bash-5.1# df -h /local
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda13      961M  657M  305M  69% /local
```

With `--boot-image-size 20G`, the boot disk image is resized to 20 GiB before boot, and Bottlerocket extends the data partition accordingly:

```
bash-5.1# df -h /local
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda13       18G  781M   18G   5% /local
```

**Terms of contribution:**

### DIFF
--- a/tools/start-local-vm
+++ b/tools/start-local-vm
@@ -54,6 +54,7 @@ usage: ${0##*/} [--arch BUILDSYS_ARCH] [--variant BUILDSYS_VARIANT]
                       [--force-extract]
                       [--inject-file LOCAL_PATH[:IMAGE_PATH]]...
                       [--firmware-code PATH] [--firmware-vars PATH]
+                      [--os-image-size SIZE] [--data-image-size SIZE]
 
 Launch a local virtual machine from a Bottlerocket image.
 
@@ -82,6 +83,8 @@ Options:
                         private partition will be lost
     --firmware-code     override the default firmware executable file
     --firmware-vars     override the initial firmware variable storage file
+    --os-image-size     resize the OS disk image to the given size (e.g. 4096M)
+    --data-image-size   resize the data disk image to the given size (e.g. 20G)
     --help              shows this usage text
 
 By default, the virtual machine's port 22 (SSH) will be exposed via the local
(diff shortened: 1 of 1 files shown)


tools/start-local-vm
```diff
@@ -80,6 +83,8 @@ Options:
                         private partition will be lost
     --firmware-code     override the default firmware executable file
     --firmware-vars     override the initial firmware variable storage file
+    --boot-image-size   resize the boot disk image to the given size (e.g. 4096M)
```
bcressey: nit: in `buildsys` we refer to this as the [os image](https://github.com/bottlerocket-os/bottlerocket/blob/baef0849029ba7708033f8b350ef91780d3eff91/tools/buildsys/src/manifest.rs#L144):
```suggestion
    --os-image-size   resize the OS disk image to the given size (e.g. 4096M)
```

tools/start-local-vm
```diff
@@ -80,6 +83,8 @@ Options:
...
+    --data-image-size   reisze the data disk image to the given size (e.g. 20G)
```
bcressey:
```suggestion
    --data-image-size   resize the data disk image to the given size (e.g. 20G)
```

bcressey (approved):

tools/start-local-vm
markusboehme: Ran a spell checker on the script and fixed another typo

tools/start-local-vm
markusboehme: Good point! Makes a lot of sense to stay consistent here. I added a commit that adopts the existing terminology for the `boot_image` variable names while I'm at it.

bcressey (approved):

foersleo (approved):
Great improvement. Also happy that you took the time to do some code hygiene.

PR #3394, opened 2023-08-31
Use `systemd-networkd` for `aws-ecs-2` and k8s 1.28 variants

zmrow:
**Issue number:**
Related to #2449 

**Description of changes:**
*Draft while testing is in progress*

This PR moves the `aws-ecs-2`, Kubernetes 1.28, and `*-dev` variants to `systemd-networkd` as the network backend.

**Testing done:**
WIP - will update as testing is completed

- [x] Build each of the variants
- [x] `aws-k8s-1.28` conformance testing in IPv6-only cluster 
```
{
"plugins": [
{
"plugin": "e2e",
"node": "global",
"status": "complete",
"result-status": "passed",
"result-counts": {
"passed": 384,
"skipped": 7007
},
```

- [x] `aws-k8s-1.28` conformance testing in IPv4 cluster 
```
 e2e   global   complete   passed   Passed:380, Failed:  0, Remaining:  0
```
- [x] `metal-k8s-1.28` conformance testing
```
e2e   complete   passed       1   Passed:380, Failed:  0, Remaining:  0
```
- [x] `vmware-k8s-1.28` conformance testing 
```
e2e   global   complete   passed   Passed:380, Failed:  0, Remaining:  0
```
- [x] `aws-ecs-2` internal ECS testing
- [x] Given custom (correct) DNS settings, instance continues to function properly in a cluster and proper interface/resolved configuration is set (set custom settings identical to what comes in via DHCP, observe the settings as `Global` via `resolvectl`.)
```
bash-5.1# cat /etc/systemd/resolved.conf.d/10-resolv.conf 
[Resolve]
DNS=192.168.0.2
Domains=us-west-2.compute.internal

bash-5.1# resolvectl 
Global
       Protocols: -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
resolv.conf mode: stub
      DNS Servers 192.168.0.2
       DNS Domain us-west-2.compute.internal

Link 2 (eth0)
Current Scopes: none
     Protocols: -DefaultRoute +LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported

bash-5.1# cat /etc/systemd/network/10-eth0.network.d/10-dns.conf 
[Network]
DNSDefaultRoute=false
[DHCPv4]
UseDNS=false
UseDomains=false
[DHCPv6]
UseDNS=false
UseDomains=false
[IPv6AcceptRA]
UseDNS=false
UseDomains=false
```
[...]

PR #3438, opened 2023-09-08
disable IPv6 RA sysctls for networkd

### LINKED ISSUE #3411: Repeated `ndisc_router_discovery failed` messages in the journal when using `systemd-networkd`
**Image I'm using:**
`aws-k8s-1.28` with `systemd-networkd` built from #3394 

**What actually happened:**
While testing for #3394 , I noticed in an IPv6-enabled cluster the nodes join the cluster and run sonobuoy just fine, but the journal is being spammed with the following messages:
```
Sep 01 22:37:25 ... kernel: ICMPv6: RA: ndisc_router_discovery failed to add default route
Sep 01 22:37:35 ... kernel: ICMPv6: RA: ndisc_router_discovery failed to add default route
```

The messages start ~10 seconds after the interfaces are brought up, and continue to be logged every 10 seconds forever.

Amazon Linux 2023 has the following in their config, perhaps this is the reason why:
```
[Route]
Gateway=_ipv6ra
```

Build an image and join it to an IPv6-only cluster

bcressey:
**Issue number:**

Fixes #3411

**Description of changes:**
Try to keep the kernel's IPv6 RA client from changing anything when using systemd-networkd's client, since that could put the interface into an unexpected state that requires a service restart to fix.

This disables all sysctls where the [upstream docs](https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt) have this note:
> Functional default: enabled if accept_ra is enabled.

These sysctls are used to selectively enable or disable parts of the kernel's `accept_ra` behavior, so disabling them all gets us as close as we can to having the entire feature disabled.

**Testing done:**

Before:
```
# cat /etc/sysctl.d/90-primary_interface.conf

# head /proc/sys/net/ipv6/conf/eth0/accept_ra{,_defrtr,_pinfo,_rtr_pref,_mtu}
==> /proc/sys/net/ipv6/conf/eth0/accept_ra <==
0

==> /proc/sys/net/ipv6/conf/eth0/accept_ra_defrtr <==
1

==> /proc/sys/net/ipv6/conf/eth0/accept_ra_pinfo <==
1

==> /proc/sys/net/ipv6/conf/eth0/accept_ra_rtr_pref <==
1

==> /proc/sys/net/ipv6/conf/eth0/accept_ra_mtu <==
1

# sysctl -w net.ipv6.conf.eth0.accept_ra=2

# dmesg|grep ICMPv6
[  402.503946] ICMPv6: RA: ndisc_router_discovery failed to add default route
```

After:
```
# cat /etc/sysctl.d/90-primary_interface.conf
-net.ipv6.conf.eth0.accept_ra = 0
-net.ipv6.conf.eth0.accept_ra_defrtr = 0
-net.ipv6.conf.eth0.accept_ra_pinfo = 0
-net.ipv6.conf.eth0.accept_ra_rtr_pref = 0
-net.ipv6.conf.eth0.accept_ra_mtu = 0

# head /proc/sys/net/ipv6/conf/eth0/accept_ra{,_defrtr,_pinfo,_rtr_pref,_mtu}
==> /proc/sys/net/ipv6/conf/eth0/accept_ra <==
0

==> /proc/sys/net/ipv6/conf/eth0/accept_ra_defrtr <==
0

==> /proc/sys/net/ipv6/conf/eth0/accept_ra_pinfo <==
0

==> /proc/sys/net/ipv6/conf/eth0/accept_ra_rtr_pref <==
0

==> /proc/sys/net/ipv6/conf/eth0/accept_ra_mtu <==
0

# sysctl -w net.ipv6.conf.eth0.accept_ra=2

# dmesg|grep ICMPv6
<nothing>
```

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/cli/mod.rs
+++ b/sources/api/netdog/src/cli/mod.rs
@@ -159,6 +159,21 @@ where
     let mut output = String::new();
     writeln!(output, "{}", ipv4_rp_filter).context(error::SysctlConfBuildSnafu)?;
 
+    #[cfg(net_backend = "systemd-networkd")]
+    // systemd-networkd implements its own RA client, and expects the kernel implementation to be
+    // unused. However, various solutions that run in EC2 might "helpfully" turn it on since it's
+    // required for most non-systemd-networkd systems. Guard against this by explicitly disabling
+    // all of the sysctls that default to enabled when "accept_ra" is enabled.
+    for ipv6_sysctl in [
+        format!("-net.ipv6.conf.{}.accept_ra = 0", interface),
+        format!("-net.ipv6.conf.{}.accept_ra_defrtr = 0", interface),
+        format!("-net.ipv6.conf.{}.accept_ra_pinfo = 0", interface),
+        format!("-net.ipv6.conf.{}.accept_ra_rtr_pref = 0", interface),
+        format!("-net.ipv6.conf.{}.accept_ra_mtu = 0", interface),
+    ] {
+        writeln!(output, "{}", ipv6_sysctl).context(error::SysctlConfBuildSnafu)?;
+    }
+
     fs::write(path, output).context(error::SysctlConfWriteSnafu { path })?;
     Ok(())
 }


zmrow:
Related to #3394

zmrow (approved):
:ship:

yeazelm (approved):

Issue #3272, opened 2023-07-19
Rewrite shells scripts from xfsprogs in Rust


yeazelm:
**What I'd like:** 
Currently, the shells scripts normally found in xfsprogs are excluded from the build. These won't work as is in Bottlerocket since they rely upon a shell. I would still like this functionality in xfsprogs, so we should rewrite them in something like Rust to provide the same experience. Most of these scripts are simple argument parsing and passing to another xfs program so it shouldn't be a ton of work.

The priority in my opinion of these scripts is:
```
fsck.xfs
xfs_admin
xfs_info
xfs_mkfile
xfs_ncheck
xfs_bmap
xfs_freeze
xfs_metadump
```

**Any alternatives you've considered:**
Running the commands manually, for instance, `xfs_info /local` can be accomplished via `xfs_spaceman -p xfs_info -c info /local` but most users won't know this workaround and its a lot more work. I don't know if there are as great alternatives for things like `fsck.xfs`.

arnaldo2792:
One of the concerns that folks had about the missing scripts were failures while using Persistent Volumes, in particular, calls to `fsck`. Depending on how the Persistent Volumes are configured, there is an execution path in the `kubelet` that executes `fsck` on a block device. I tried different scenarios to generate failures when using the `xfs` utils provided by the host. These include:

1. Disk is formatted as `xfs` and the PV is configured as `ext4`, a failure occurs since `fsck -a ext4` fails to run on the `xfs` partition
2. Disk is formatted as `xfs` and the PV is configured as `xfs`, nothing fails, `fsck -a xfs` runs
3. Disk isn't formatted, and the PV is configured with an unsupported format (e.g., `btrfs`), the mount fails
4. Path to the block device is used instead of the path to the `xfs` partition, and the PV is configured as `xfs`, `fsck -a xfs` fails since command fails to identify where is the partition.

This is the configuration file I used for my testing, the only value that I changed was `fsType`, to trigger specific failures:

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: example-pv
spec:
  capacity:
    storage: 100Gi
  volumeMode: Filesystem
  accessModes:
  - ReadWriteMany
  persistentVolumeReclaimPolicy: Delete
  storageClassName: local-storage
  local:
    path: /dev/disk/by-label/LOCAL_VOLUME
    fsType: xfs
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
[...]

PR #3444, opened 2023-09-12
Xfsprogs cli

vyaghras:
**Issue number:**

Completes parts of # [3272](https://github.com/bottlerocket-os/bottlerocket/issues/3272) 

**Description of changes:**
Rewrite xfs_info, xfs_admin and fsck.xfs cli for bottlerocket as Bottlerocket does not have a shell.
- xfs_info cli: This works for both mounted and unmounted target. It fetches data using
xfs_spaceman when the target is mounted and use xfs_db in case of
unmounted target.
Param "-t" for mtab is not included in this cli as
xfs_spaceman -p xfs_info -c "info" -t "$mtab" "${target}"
do not produce expected output. The cli can be used as:
     - xfs_info -V
     - xfs_info $target where target is mount_point or block_device
- xfs_admin cli : All the commands for xfs progs that are listed on
https://man7.org/linux/man-pages/man8/xfs_admin.8.html
can be executed using this CLI.
- fsck.xfs: We are keeping it open for interactive sesssion too, though it is
recommended to invoke xfs_repair for Normal administrative filesystem
repairs directly.

**Testing done:**

-  xfs_admin
- [x] xfs_admin -u $device - For unmounted device
```
sudo xfs_admin -u /dev/nvme3n1p1
UUID = 405fabf4-0e29-40e9-a291-61400aac4fbb
______________________________________________
cargo run --bin xfs_admin --  -u /dev/nvme3n1p1
UUID = 405fabf4-0e29-40e9-a291-61400aac4fbb
```
- [x]  xfs_admin -l /dev/nvme3n1p1 - For unmounted device
```
sudo xfs_admin -l /dev/nvme3n1p1
label = ""
______________________________________________
cargo run --bin xfs_admin --  -l /dev/nvme3n1p1
label = ""
```
- [x] xfs_admin -u $device - For mounted device
```
xfs_admin -u /dev/nvme1n1p1
UUID = 03837d91-7a4a-4d9e-ab0c-176502893af4
```
- [x]  xfs_admin -l /dev/nvme3n1p1 - For mounted device
```
xfs_admin -l /dev/nvme1n1p1
label = " "
```
- [x] xfs_admin -V
```
sudo xfs_admin -V
xfs_admin version 5.0.0
_____________________________________________
 cargo run --bin xfs_admin --  -V
xfs_admin version 5.0.0
``` 
- [x]  xfs_admin --  -e  $device
```
sudo xfs_admin -e /dev/nvme3n1p1
unwritten extents always enabled for v5 superblocks.
versionnum [0xbca5+0x18a] = V5,NLINK,DIRV2,ALIGN,LOGV2,EXTFLG,SECTOR,MOREBITS,ATTR2,LAZYSBCOUNT,PROJID32BIT,CRC,FTYPE,FINOBT,SPARSE_INODES,REFLINK
______________________________________________
cargo run --bin xfs_admin --  -e /dev/nvme3n1p1
unwritten extents always enabled for v5 superblocks.
versionnum [0xbca5+0x18a] = V5,NLINK,DIRV2,ALIGN,LOGV2,EXTFLG,SECTOR,MOREBITS,ATTR2,LAZYSBCOUNT,PROJID32BIT,CRC,FTYPE,FINOBT,SPARSE_INODES,REFLINK
```
- [x] xfs_admin -L new_label $device
```
sudo xfs_admin -L new_label /dev/nvme3n1p1
writing all SBs
new label = "new_label"
sudo xfs_admin -L -- /dev/nvme3n1p1
writing all SBs
new label = ""
_____________________________________________
cargo run --bin xfs_admin --  -L new_label /dev/nvme3n1p1
writing all SBs
new label = "new_label"
 cargo run --bin xfs_admin --  -L -- /dev/nvme3n1p1
writing all SBs
new label = ""

Mounted device
xfs_admin -L new_label /dev/nvme1n1p1
label = "new_label"
xfs_admin -L -- /dev/nvme1n1p1
label = " "
```
- [x] xfs_admin -c 1 -O inobtcount=1 $device
- [x] xfs_admin -f $device
- [x] xfs_admin -j $device
```
sudo xfs_admin -j /dev/nvme3n1p1
Version 2 logs always enabled for v5 superblocks.
versionnum [0xbca5+0x18a] = V5,NLINK,DIRV2,ALIGN,LOGV2,EXTFLG,SECTOR,MOREBITS,ATTR2,LAZYSBCOUNT,PROJID32BIT,CRC,FTYPE,FINOBT,SPARSE_INODES,REFLINK
________________________________________________
cargo run --bin xfs_admin --  -j /dev/nvme3n1p1
Version 2 logs always enabled for v5 superblocks.
versionnum [0xbca5+0x18a] = V5,NLINK,DIRV2,ALIGN,LOGV2,EXTFLG,SECTOR,MOREBITS,ATTR2,LAZYSBCOUNT,PROJID32BIT,CRC,FTYPE,FINOBT,SPARSE_INODES,REFLINK
```
- [x] xfs_admin -p $device
```
// Unmounted filesystem
sudo xfs_admin -p /dev/nvme3n1p1
xfs_admin: Cannot change projid32bit on v5 superblocks.
_________________________________________________
cargo run --bin xfs_admin --  -p /dev/nvme3n1p1
xfs_admin: Cannot change projid32bit on v5 superblocks

// Mounted filesystem
// Error message and error code is different because of xfs_progs version
sudo xfs_admin -p /dev/nvme0n1p1
xfs_admin: /dev/nvme0n1p1 contains a mounted filesystem
echo $?
1
__________________________________________________
cargo run --bin xfs_admin --  -p /dev/nvme0n1p1
filesystem is mounted /
echo $?
2
```
- [x] xfs_admin -U c1b9d5a2-f162-11cf-9ece-0020afc76f16 $device
```
xfs_admin -U c1b9d5a2-f162-11cf-9ece-0020afc76f16 /dev/nvme3n1p1
Clearing log and setting UUID
writing all SBs
new UUID = c1b9d5a2-f162-11cf-9ece-0020afc76f16
sudo xfs_admin -U 405fabf4-0e29-40e9-a291-61400aac4fbb /dev/nvme3n1p1
Clearing log and setting UUID
writing all SBs
new UUID = 405fabf4-0e29-40e9-a291-61400aac4fbb
__________________________________________________
cargo run --bin xfs_admin --  -U c1b9d5a2-f162-11cf-9ece-0020afc76f16 /dev/nvme3n1p1
Clearing log and setting UUID
writing all SBs
new UUID = c1b9d5a2-f162-11cf-9ece-0020afc76f16
cargo run --bin xfs_admin --  -U 405fabf4-0e29-40e9-a291-61400aac4fbb /dev/nvme3n1p1
Clearing log and setting UUID
writing all SBs
new UUID = 405fabf4-0e29-40e9-a291-61400aac4fbb
```
- [x] Fail when device is not provided
```
sudo xfs_admin -u
Usage: xfs_admin [-efjlpuV] [-c 0|1] [-L label] [-U uuid] device
echo $?
2
__________________________________________________
cargo run --bin xfs_admin --  -u
Couldn't parse target block-device
echo $?
2
```
- xfs_info
- [x] xfs_info  -V
```
sudo xfs_info  -V
xfs_info version 5.0.0
 echo $?
0
__________________________________________________
cargo run --bin xfs_info --  -V
xfs_info version 5.0.0
echo $?
0
```
- [x] Fail when device is not provided
```
sudo xfs_info 
Usage: xfs_info [-V] [-t mtab] [mountpoint|device|file]
echo $?
2
__________________________________________________
cargo run --bin xfs_info --
Couldn't parse target  mount-point | block-device | file-image
[ec2-user@ip-172-31-46-161 xfscli]$ echo $?
2
```
- [x] xfs_info $device
```
// Unmounted device
sudo xfs_info /dev/nvme3n1p1
meta-data=/dev/nvme3n1p1         isize=512    agcount=4, agsize=6553536 blks
         =                       reflink=1    bigtime=0 inobtcount=0
data     =                       bsize=4096   blocks=26214144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=12799, version=2
________________________________________________________
cargo run --bin xfs_info --  /dev/nvme3n1p1
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `/home/ec2-user/bottlerocket/sources/target/debug/xfs_info /dev/nvme3n1p1`
meta-data=/dev/nvme3n1p1         isize=512    agcount=4, agsize=6553536 blks
         =                       reflink=1    bigtime=0 inobtcount=0
data     =                       bsize=4096   blocks=26214144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=12799, version=2

// Mounted device
findmnt /dev/nvme4n1p1
TARGET   SOURCE         FSTYPE
/volume2 /dev/nvme4n1p1 xfs 

xfs_info /dev/nvme4n1p1
meta-data=/dev/nvme4n1p1         isize=512    agcount=4, agsize=6553536 blks
         =                       reflink=1    bigtime=0 inobtcount=0
data     =                       bsize=4096   blocks=26214144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=12799, version=2
_______________________________________________________
cargo run --bin xfs_info --  /dev/nvme4n1p1
meta-data=/dev/nvme4n1p1         isize=512    agcount=4, agsize=6553536 blks
         =                       reflink=1    bigtime=0 inobtcount=0
data     =                       bsize=4096   blocks=26214144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=12799, version=2
```
- fsck.xfs
- [x] fsck.xfs  -a /dev/nvme3n1p1 -- Do nothing, successfully
```
sudo fsck.xfs  -a /dev/nvme3n1p1
/sbin/fsck.xfs: XFS file system.
echo $?
0
_________________________________________________
cargo run --bin fsck_xfs --  -a /dev/nvme3n1p1
xfs fsck : XFS file system.
echo $?
0
```

**Terms of contribution:**

### DIFF
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -90,6 +90,7 @@ Requires: %{_cross_os}shimpei
 Requires: %{_cross_os}signpost
 Requires: %{_cross_os}storewolf
 Requires: %{_cross_os}sundog
+Requires: %{_cross_os}xfscli
 Requires: %{_cross_os}thar-be-settings
 Requires: %{_cross_os}thar-be-updates
 Requires: %{_cross_os}updog
@@ -293,6 +294,11 @@ Summary: Compliance check framework
 %description -n %{_cross_os}bloodhound
 %{summary}.
 
+%package -n %{_cross_os}xfscli
+Summary: XFS progs cli
+%description -n %{_cross_os}xfscli
+%{summary}.
+
 %prep
 %setup -T -c
 %cargo_prep
@@ -352,6 +358,7 @@ echo "** Output from non-static builds:"
     -p certdog \
     -p shimpei \
     -p bloodhound \
+    -p xfscli \
     %{?with_ecs_runtime: -p ecs-settings-applier} \
     %{?with_aws_platform: -p shibaken -p cfsignal} \
     %{?with_aws_k8s_family: -p pluto} \
@@ -389,6 +396,14 @@ done
 %if %{with k8s_runtime}
 install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/kubernetes-checks %{buildroot}%{_cross_bindir}
 %endif
+install -d %{buildroot}%{_cross_sbindir}
+for p in \
+  xfs_admin xfs_info \
+; do
+  install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/${p} %{buildroot}%{_cross_sbindir}/
+done
+# Rename fsck_xfs binary to fsck.xfs
+install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/fsck_xfs %{buildroot}%{_cross_sbindir}/fsck.xfs
 
 # Add the bloodhound checker symlinks
 mkdir -p %{buildroot}%{_cross_libexecdir}/cis-checks/bottlerocket
@@ -666,6 +681,12 @@ install -p -m 0644 %{S:121} %{buildroot}%{_cross_unitdir}
 %{_cross_bindir}/bloodhound
 %{_cross_bindir}/bottlerocket-checks
 %{_cross_libexecdir}/cis-checks/bottlerocket
+
+%files -n %{_cross_os}xfscli
+%{_cross_sbindir}/xfs_admin
+%{_cross_sbindir}/xfs_info
+%{_cross_sbindir}/fsck.xfs
+
 %if %{with k8s_runtime}
 %{_cross_bindir}/kubernetes-checks
 %{_cross_libexecdir}/cis-checks/kubernetes
--- a/sources/xfscli/Cargo.toml
+++ b/sources/xfscli/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "xfscli"
+version = "0.1.0"
+authors = ["Shikha Vyaghra <vyaghras@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+
+[[bin]]
+name = "xfs_info"
+
+[[bin]]
+name = "xfs_admin"
+
+[[bin]]
+name = "fsck_xfs"
+path = "src/bin/fsck_xfs/main.rs"
+
+[dependencies]
+argh = "0.1"
+tempfile = "3"
+snafu = "0.7"
--- a/sources/xfscli/src/bin/xfs_admin.rs
+++ b/sources/xfscli/src/bin/xfs_admin.rs
@@ -1,0 +1,246 @@
+use snafu::{OptionExt, ResultExt};
+use std::process::Command;
+use std::process::ExitCode;
+use std::str;
+
+use argh::FromArgs;
+
+use xfscli::error::{self, Error, Result};
+use xfscli::{get_mount_point, perform_xfs_repair, query_xfs_db, XFS_IO};
+
+// This CLI is rewritten in Rust for xfs_admin shell script
+// https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/db/xfs_admin.sh
+// since there isn't a shell on Bottlerocket hosts.
+
+/// change parameters of an XFS filesystem
+#[derive(FromArgs)]
+struct Args {
+    /// enables unwritten extent support on a filesystem that is disabled
+    #[argh(switch, short = 'e')]
+    extent_flag: bool,
+
+    /// specifies that the filesystem image to be processed is stored in a regular file at device
+    #[argh(switch, short = 'f')]
+    filesystem_image: bool,
+
+    /// prints the current filesystem label
+    #[argh(switch, short = 'l')]
+    filesystem_label: bool,
+
+    /// prints the current filesystem UUID
+    #[argh(switch, short = 'u')]
+    filesystem_uuid: bool,
+
+    /// set the filesystem label to label
+    #[argh(option, short = 'L')]
+    label: Option<String>,
+
+    /// enable (1) or disable (0) lazy-counters in the filesystem
+    #[argh(option, short = 'c')]
+    lazy_counters: Option<usize>,
+
+    /// enables version 2 log format
+    #[argh(switch, short = 'j')]
+    log_version_2: bool,
+
+    /// filesystem's external log location
+    #[argh(option)]
+    log_location: Option<String>,
+
+    /// enable 32bit project identifier support
+    #[argh(switch, short = 'p')]
+    project_identifier_32_bit: bool,
+
+    /// specifies the device special file where the filesystem's realtime section resides
+    #[argh(option, short = 'r')]
+    realtime_device: Option<String>,
+
+    /// set the UUID of the filesystem to uuid
+    #[argh(option, short = 'U')]
+    uuid: Option<String>,
+
+    /// add or remove features on an existing V5 filesystem
+    #[argh(option, short = 'O')]
+    v5_feature: Option<String>,
+
+    /// prints the version number and exits.
+    #[argh(switch, short = 'V')]
+    version: bool,
+
+    /// block-device
+    #[argh(positional)]
+    device: Option<String>,
+}
+
+impl Args {
+    // Filesystem should not be mounted with args in this function
+    // For more reference check require_offline variable in
+    // https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/db/xfs_admin.sh
+    fn is_required_offline(&self) -> bool {
+        self.extent_flag
+            || self.filesystem_image
+            || self.log_version_2
+            || self.project_identifier_32_bit
+            || self.uuid.is_some()
+            || self.lazy_counters.is_some()
+            || self.v5_feature.is_some()
+            || self.realtime_device.is_some()
+    }
+}
+
+fn main() -> ExitCode {
+    let args: Args = argh::from_env();
+    match run(args) {
+        Ok(exit_code) => exit_code,
+        Err(e) => {
+            eprintln!("{}", e);
+            e.exit_code()
+        }
+    }
+}
+
+fn run(args: Args) -> Result<ExitCode> {
+    // Return the version for arg version as true
+    if args.version {
+        let xfs_db_response = query_xfs_db(vec!["-p", "xfs_admin", "-V"])?;
+        return Ok(ExitCode::from(xfs_db_response as u8));
+    }
+
+    let device = &args.device.as_ref().context(error::ParseTargetSnafu)?;
+
+    // Collect all the arguments for XFS_DB, XFS_IO and XFS_REPAIR commands
+    // by parsing the CLI inputs.
+    let mut db_opts: Vec<&str> = vec![];
+    let mut io_opts: Vec<&str> = vec![];
+    let mut repair_opts: Vec<&str> = vec![];
+
+    let lazycount_arg;
+    if let Some(lazy_counters) = &args.lazy_counters {
+        lazycount_arg = format!("lazycount={}", lazy_counters);
+        repair_opts.append(&mut vec!["-c", lazycount_arg.as_str()]);
+    }
+
+    if args.extent_flag {
+        db_opts.append(&mut vec!["-c", "version extflg"]);
+    }
+
+    if args.filesystem_image {
+        db_opts.push("-f");
+    }
+
+    if args.log_version_2 {
+        db_opts.append(&mut vec!["-c", "version log2"]);
+    }
+
+    if args.filesystem_label {
+        db_opts.append(&mut vec!["-r", "-c", "label"]);
+        io_opts.append(&mut vec!["-r", "-c", "label"]);
+    }
+
+    let label_arg_db;
+    let label_arg_io;
+    if let Some(label) = &args.label {
+        label_arg_db = format!("label {}", label);
+        db_opts.append(&mut vec!["-c", label_arg_db.as_str()]);
+
+        label_arg_io = match label.as_str() {
+            "--" => format!("label {}", label),
+            _ => {
+                format!("label -s {}", label)
+            }
+        };
+        io_opts.append(&mut vec!["-c", label_arg_io.as_str()]);
+    }
+
+    if let Some(v5_feature) = &args.v5_feature {
+        repair_opts.append(&mut vec!["-c", v5_feature.as_str()]);
+    }
+
+    if args.project_identifier_32_bit {
+        db_opts.append(&mut vec!["-c", "version projid32bit"]);
+    }
+
+    if args.filesystem_uuid {
+        db_opts.append(&mut vec!["-r", "-c", "uuid"]);
+        io_opts.append(&mut vec!["-r", "-c", "fsuuid"]);
+    }
+
+    let uuid_arg;
+    if let Some(uuid) = &args.uuid {
+        uuid_arg = format!("uuid {}", uuid);
+        db_opts.append(&mut vec!["-c", uuid_arg.as_str()]);
+    }
+
+    // Exit if filesystem is mounted and require offline for queried arguments
+    if let Ok(mpt) = get_mount_point(device) {
+        // filesystem is mounted
+        if args.is_required_offline() {
+            return Err(Error::MountedFilesystem { mount_point: mpt });
+        }
+
+        if !io_opts.is_empty() {
+            let mut xfs_io = Command::new(XFS_IO);
+            xfs_io.args(["-p", "xfs_admin"]);
+            xfs_io.args(io_opts);
+            xfs_io.arg(mpt.as_str().trim());
+
+            let xfs_io_result = xfs_io.output().context(error::CommandFailureSnafu {
+                command: "xfs_io".to_string(),
+            })?;
+
+            print!(
+                "{}",
+                String::from_utf8(xfs_io_result.stdout).context(error::FromUtf8Snafu {
+                    command: "xfs io".to_string(),
+                })?,
+            );
+            let xfs_io_result =
+                xfs_io_result
+                    .status
+                    .code()
+                    .context(error::ParseStatusCodeSnafu {
+                        command: "xfs_io".to_string(),
+                    })?;
+            return Ok(ExitCode::from(xfs_io_result as u8));
+        }
+    }
+
+    let mut status = 0;
+    if !db_opts.is_empty() {
+        let mut xfs_db_args = vec!["-x", "-p", "xfs_admin"];
+
+        let log_location_arg;
+        if let Some(log_location) = &args.log_location {
+            log_location_arg = format!("-l {}", log_location);
+            xfs_db_args.push(log_location_arg.as_str());
+        }
+
+        xfs_db_args.append(&mut db_opts);
+        xfs_db_args.push(device);
+
+        status = query_xfs_db(xfs_db_args)?;
+    }
+
+    if !repair_opts.is_empty() {
+        print!("Running xfs_repair to upgrade filesystem.");
+        let mut repair_args: Vec<&str> = vec![];
+
+        let log_location_arg;
+        if let Some(log_location) = &args.log_location {
+            log_location_arg = format!("-l {}", log_location);
+            repair_args.push(log_location_arg.as_str());
+        }
+
+        let realtime_device_arg;
+        if let Some(realtime_device) = &args.realtime_device {
+            realtime_device_arg = format!("-r {}", realtime_device);
+            repair_args.push(realtime_device_arg.as_str());
+        }
+
+        repair_args.append(&mut repair_opts);
+        repair_args.push(device.as_str());
+
+        status += perform_xfs_repair(repair_args)?.exit_code();
+    }
+    Ok(ExitCode::from(status as u8))
+}
--- a/sources/xfscli/src/bin/xfs_info.rs
+++ b/sources/xfscli/src/bin/xfs_info.rs
@@ -1,0 +1,53 @@
+use snafu::OptionExt;
+use std::process::ExitCode;
+use std::str;
+
+use argh::FromArgs;
+
+use xfscli::error::{self, Result};
+use xfscli::{get_mount_point, query_xfs_db, query_xfs_spaceman};
+
+// This CLI is rewritten in Rust for xfs_info shell script
+// https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/spaceman/xfs_info.sh
+// since there isn't a shell on Bottlerocket hosts.
+
+/// display XFS filesystem geometry information
+#[derive(FromArgs)]
+struct Args {
+    /// print the version
+    #[argh(switch, short = 'V')]
+    version: bool,
+
+    /// mount-point | block-device | file-image
+    #[argh(positional)]
+    target: Option<String>,
+}
+
+fn main() -> ExitCode {
+    let args: Args = argh::from_env();
+    match run(args) {
+        Ok(exit_code) => exit_code,
+        Err(e) => {
+            eprintln!("{}", e);
+            e.exit_code()
+        }
+    }
+}
+
+fn run(args: Args) -> Result<ExitCode> {
+    // Return the version for arg version as true
+    if args.version {
+        let xfs_spaceman_response = query_xfs_spaceman(vec!["-p", "xfs_info", "-V"])?;
+        return Ok(ExitCode::from(xfs_spaceman_response as u8));
+    }
+
+    let target = &args.target.as_ref().context(error::ParseTargetSnafu)?;
+
+    // Get info using XFS_SPACEMAN if mounted else from XFS_DB
+    let response = if let Ok(mpt) = get_mount_point(target) {
+        query_xfs_spaceman(vec!["-p", "xfs_info", "-c", "info", mpt.as_str().trim()])?
+    } else {
+        query_xfs_db(vec!["-p", "xfs_info", "-c", "info", target.as_str()])?
+    };
+    Ok(ExitCode::from(response as u8))
+}
--- a/sources/xfscli/src/error.rs
+++ b/sources/xfscli/src/error.rs
@@ -1,0 +1,46 @@
+use snafu::Snafu;
+use std::process::ExitCode;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub))]
+#[repr(i32)]
+pub enum Error {
+    #[snafu(display("Failed to run '{}' successfully {}", command, source))]
+    CommandFailure {
+        command: String,
+        source: std::io::Error,
+    },
+
+    #[snafu(display("Unable to find the mount point for device: {}", target))]
+    FindMount { target: String },
+
+    #[snafu(display("Failed to parse '{}' output: {}", command, source))]
+    FromUtf8 {
+        command: String,
+        source: std::string::FromUtf8Error,
+    },
+
+    #[snafu(display("{} filesystem is mounted", mount_point))]
+    MountedFilesystem { mount_point: String },
+
+    #[snafu(display("Failed to read '{}' status code", command))]
+    ParseStatusCode { command: String },
+
+    #[snafu(display("Could not parse target block device",))]
+    ParseTarget,
+}
+
+impl Error {
+    pub fn exit_code(&self) -> ExitCode {
+        match self {
+            Error::CommandFailure { .. } => ExitCode::from(8),
+            Error::FindMount { .. } => ExitCode::from(8),
+            Error::FromUtf8 { .. } => ExitCode::from(8),
+            Error::MountedFilesystem { .. } => ExitCode::from(2),
+            Error::ParseStatusCode { .. } => ExitCode::from(8),
+            Error::ParseTarget => ExitCode::from(2),
+        }
+    }
+}
--- a/sources/xfscli/src/lib.rs
+++ b/sources/xfscli/src/lib.rs
@@ -1,0 +1,142 @@
+use error::Result;
+use snafu::{OptionExt, ResultExt};
+use std::process::Command;
+
+pub mod error;
+pub static BLKID: &str = "/usr/sbin/blkid";
+pub static FINDMNT: &str = "/usr/bin/findmnt";
+pub static MOUNT: &str = "/usr/bin/mount";
+pub static UMOUNT: &str = "/usr/bin/umount";
+pub static XFS_DB: &str = "/usr/sbin/xfs_db";
+pub static XFS_IO: &str = "/usr/sbin/xfs_io";
+pub static XFS_REPAIR: &str = "/usr/sbin/xfs_repair";
+pub static XFS_SPACEMAN: &str = "/usr/sbin/xfs_spaceman";
+
+/// Check if target is mounted and return mount point
+pub fn get_mount_point(target: &String) -> Result<String> {
+    let mountpt = Command::new(FINDMNT)
+        .args(["-t", "xfs", "-f", "-n", "-o", "TARGET", target])
+        .output()
+        .context(error::CommandFailureSnafu {
+            command: "findmnt".to_string(),
+        })?;
+
+    // All errors will be interpreted as failure in finding the mount point
+    if mountpt.status.success() {
+        Ok(
+            String::from_utf8(mountpt.stdout).context(error::FromUtf8Snafu {
+                command: "mount point".to_string(),
+            })?,
+        )
+    } else {
+        Err(error::Error::FindMount {
+            target: String::from(target),
+        })
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+
+pub enum XfsRepairResponseCode {
+    CommandUnavailable,
+    DirtyLogs,
+    MetadataRepair,
+    Ok,
+    RepairFailure,
+    Unknown(i32),
+}
+
+impl XfsRepairResponseCode {
+    pub fn exit_code(self) -> i32 {
+        match self {
+            XfsRepairResponseCode::CommandUnavailable => 127,
+            XfsRepairResponseCode::DirtyLogs => 2,
+            XfsRepairResponseCode::MetadataRepair => 4,
+            XfsRepairResponseCode::Ok => 0,
+            XfsRepairResponseCode::RepairFailure => 1,
+            XfsRepairResponseCode::Unknown(code) => code,
+        }
+    }
+}
+
+impl From<i32> for XfsRepairResponseCode {
+    fn from(code: i32) -> Self {
+        match code {
+            0 => XfsRepairResponseCode::Ok,
+            1 => XfsRepairResponseCode::RepairFailure,
+            2 => XfsRepairResponseCode::DirtyLogs,
+            4 => XfsRepairResponseCode::MetadataRepair,
+            127 => XfsRepairResponseCode::CommandUnavailable,
+            code => XfsRepairResponseCode::Unknown(code),
+        }
+    }
+}
+
+/// Perform xfs_repair for the given arguments
+pub fn perform_xfs_repair(args: Vec<&str>) -> Result<XfsRepairResponseCode> {
+    let xfs_repair_result =
+        Command::new(XFS_REPAIR)
+            .args(args)
+            .status()
+            .context(error::CommandFailureSnafu {
+                command: "xfs_repair".to_string(),
+            })?;
+
+    let code = xfs_repair_result
+        .code()
+        .context(error::ParseStatusCodeSnafu {
+            command: "xfs_repair".to_string(),
+        })?;
+
+    Ok(XfsRepairResponseCode::from(code))
+}
+
+/// Query xfs db to get required information
+pub fn query_xfs_db(args: Vec<&str>) -> Result<i32> {
+    let xfs_db_result =
+        Command::new(XFS_DB)
+            .args(args)
+            .output()
+            .context(error::CommandFailureSnafu {
+                command: "xfs db".to_string(),
+            })?;
+
+    print!(
+        "{}",
+        String::from_utf8(xfs_db_result.stdout).context(error::FromUtf8Snafu {
+            command: "xfs db".to_string(),
+        })?
+    );
+
+    xfs_db_result
+        .status
+        .code()
+        .context(error::ParseStatusCodeSnafu {
+            command: "xfs_db".to_string(),
+        })
+}
+
+/// Query xfs spaceman to get xfs information
+pub fn query_xfs_spaceman(args: Vec<&str>) -> Result<i32> {
+    let xfs_spaceman_result =
+        Command::new(XFS_SPACEMAN)
+            .args(args)
+            .output()
+            .context(error::CommandFailureSnafu {
+                command: "xfs spaceman".to_string(),
+            })?;
+
+    print!(
+        "{}",
+        String::from_utf8(xfs_spaceman_result.stdout).context(error::FromUtf8Snafu {
+            command: "xfs spaceman".to_string(),
+        })?
+    );
+
+    xfs_spaceman_result
+        .status
+        .code()
+        .context(error::ParseStatusCodeSnafu {
+            command: "xfs_spaceman".to_string(),
+        })
+}
--- a/packages/os/Cargo.toml
+++ b/packages/os/Cargo.toml
@@ -22,6 +22,7 @@ source-groups = [
     "driverdog",
     "cfsignal",
     "bloodhound",
+    "xfscli",
 ]
 package-features = [ "systemd-networkd" ]
 
--- a/sources/Cargo.lock
+++ b/sources/Cargo.lock
@@ -4807,6 +4807,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "xfscli"
+version = "0.1.0"
+dependencies = [
+ "argh",
+ "snafu",
+ "tempfile",
+]
+
+[[package]]
 name = "xml-rs"
 version = "0.8.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/sources/Cargo.toml
+++ b/sources/Cargo.toml
@@ -103,7 +103,9 @@ members = [
 
     "constants",
 
-    "shimpei"
+    "shimpei",
+
+    "xfscli"
 ]
 
 [profile.release]
--- a/sources/xfscli/src/bin/fsck_xfs/error.rs
+++ b/sources/xfscli/src/bin/fsck_xfs/error.rs
@@ -1,0 +1,107 @@
+use snafu::Snafu;
+use std::process::ExitCode;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub))]
+#[repr(i32)]
+pub enum Error {
+    // Errors to map the errors defined in
+    // https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/fsck/xfs_fsck.sh
+    #[snafu(display("xfs {} errror: {} was not found!", cli, command))]
+    CommandUnavailable { cli: String, command: String },
+
+    #[snafu(display("{} does not exist", target))]
+    DeviceNotFound { target: String },
+
+    #[snafu(display(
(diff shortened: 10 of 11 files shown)


packages/os/os.spec
```diff
@@ -389,6 +390,13 @@ done
 %if %{with k8s_runtime}
 install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/kubernetes-checks %{buildroot}%{_cross_bindir}
 %endif
+install -d %{buildroot}%{_cross_sbindir}
+for p in \
+  xfs_admin xfs_info \
+; do
+  install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/${p} %{buildroot}%{_cross_sbindir}/
+done
+install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/fsck_xfs %{buildroot}%{_cross_sbindir}/fsck.xfs
```
yeazelm: I would call out fsck_xfs needs to be renamed. There might be more than need it going forward.

packages/os/os.spec
```diff
@@ -666,6 +674,9 @@ install -p -m 0644 %{S:121} %{buildroot}%{_cross_unitdir}
 %{_cross_bindir}/bloodhound
 %{_cross_bindir}/bottlerocket-checks
 %{_cross_libexecdir}/cis-checks/bottlerocket
+%{_cross_sbindir}/xfs_admin
+%{_cross_sbindir}/xfs_info
+%{_cross_sbindir}/fsck.xfs
```
yeazelm, line 688: Please break these out in their own package, you can build them at the same time as `bloodhound` but they should be their own sub-package.

sources/xfscli/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[[bin]]
+name = "xfs_info"
+
+[[bin]]
+name = "xfs_admin"
+
+[[bin]]
+name = "fsck_xfs"
+path = "src/bin/xfs_fsck.rs"
```
yeazelm: Does `fsck.xfs` not work for this name?

sources/xfscli/src/bin/xfs_info.rs
```diff
@@ -0,0 +1,61 @@
...
+        if mountpt.status.success() {
+            let mpt = String::from_utf8(mountpt.stdout).unwrap();
+            let xfs_spaceman_result = Command::new(XFS_SPACEMAN)
+                .args(["-p", "xfs_info", "-c", "info", mpt.as_str().trim()])
+                .output()
+                .expect("Couldn't run xfs info for mounted point");
+
+            print!("{}", String::from_utf8(xfs_spaceman_result.stdout).unwrap());
+        } else {
+            let xfs_spaceman_result = Command::new(XFS_DB)
```
yeazelm: This is should probably be `xfs_db_result`.

sources/xfscli/src/bin/xfs_info.rs
```diff
@@ -0,0 +1,61 @@
...
+                .args(["-p", "xfs_info", "-c", "info", target.as_str()])
+                .output()
+                .unwrap();
+            print!("{}", String::from_utf8(xfs_spaceman_result.stdout).unwrap());
+        }
+    } else {
+        println!("Couldn't parse target  mount-point | block-device | file-image");
```
yeazelm: This should exit 2 to stay consistent with https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/spaceman/xfs_info.sh

sources/xfscli/src/bin/xfs_info.rs
```diff
@@ -0,0 +1,61 @@
...
+        print!("{}", String::from_utf8(xfs_spaceman_result.stdout).unwrap());
+        exit(0);
+    }
+
+    if let Some(target) = &args.target {
+        // Check if target is mounted
+        let mountpt = Command::new(FINDMNT)
+            .args(["-t", "xfs", "-f", "-n", "-o", "TARGET", target.as_str()])
+            .output()
+            .expect("Couldn't find the mount point");
```
yeazelm: nit: Can we get this information from https://docs.rs/mnt/latest/mnt/ or another library? I'd like to avoid using `Command` to remove specific dependencies on this command returning input and accepting parameters exactly as they do right now. Not that `findmnt` is likely to change, its more of a general stance.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,211 @@
...
+    fs_label: bool,
+    /// enable 32bit project identifier support
+    #[argh(switch, short = 'p')]
+    projid32bit: bool,
+    /// prints the current filesystem UUID
+    #[argh(switch, short = 'u')]
+    fs_uuid: bool,
+    /// enable (1) or disable (0) lazy-counters in the filesystem
+    #[argh(option, short = 'c')]
+    lazy_counters: Option<usize>,
```
yeazelm: nit: Please sort these in lexicographical order.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,211 @@
+use std::ops::Add;
+use std::process::exit;
+use std::process::Command;
+use std::str;
+
+use argh::FromArgs;
+
```
yeazelm, line 10: Its probably worth linking to the upstream shell script it aims to replace, for instance: https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/db/xfs_admin.sh It took me a bit of time to find this and start comparing.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,211 @@
...
+            let xfs_repair_output = xfs_repair
+                .output()
+                .expect("Couldn't run xfs repair for the given device");
+            print!("{}", String::from_utf8(xfs_repair_output.stdout).unwrap());
+            status += xfs_repair_output.status.code().unwrap();
+            exit(status);
+        }
+    } else {
+        println!("Couldn't parse target block-device");
```
yeazelm: exit 2 if we fail the command.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// set the UUID of the filesystem to uuid
+    #[argh(option, short = 'U')]
+    uuid: Option<String>,
+    /// specifies the device special file where the filesystem's realtime section resides
+    #[argh(option, short = 'r')]
+    rtdev: Option<String>,
+    /// block-device
+    #[argh(positional)]
+    device: Option<String>,
+    /// filesystem's external log location
```
yeazelm: nit: Its common practice to separate positional from options so that the reader can find those easily. Please move this to the bottom and separate with a space.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,211 @@
...
+            .expect("Couldn't run xfs db for version info");
+        print!("{}", String::from_utf8(xfs_db_result.stdout).unwrap());
+        exit(xfs_db_result.status.code().unwrap());
+    }
+
+    let mut db_opts: Vec<&str> = vec![];
+    let mut io_opts: Vec<&str> = vec![];
+    let mut repair_opts: Vec<&str> = vec![];
+
+    let mut lazycount_arg = "lazycount=".to_string();
```
yeazelm: Can you add comments on what they are intended to be used for? I think I understand because I've read the other shell script but a reader might not know immediately what they are for and why there are so many being created at once.

sources/xfscli/src/bin/xfs_fsck.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        if auto {
+            println!("xfs fsck : XFS file system.");
+            exit(0);
+        } else {
+            println!(
+                "If you wish to check the consistency of an XFS filesystem or \
+            repair a damaged filesystem, see xfs_repair(8)."
+            );
+        }
```
yeazelm: This script has very particular exit statuses that are relied upon by other tools to inform their own decisions. We should be very explicit about keeping those exit codes consistent.

sources/xfscli/Cargo.toml
vyaghras: No, we can do as following in the newer version of Cargo. But it is not possible with the Cargo version that we have in Bottlerocket. 
```
[[bin]]
name = "hello-rust"
filename = "hello-rust.fap"
```

sources/xfscli/src/bin/xfs_fsck.rs
vyaghras: Resolved all the exit statuses.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,215 @@
+use std::process::exit;
+use std::process::Command;
+use std::str;
+
+use argh::FromArgs;
+
+// This CLI is rewritten in Rust for xfs_admin shell script
+// https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/db/xfs_admin.sh
+// as Shell is not available on a Bottlerocket host.
```
yeazelm:
```suggestion
// This CLI is rewritten in Rust for xfs_admin shell script
// https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/db/xfs_admin.sh
// since there isn't a shell on Bottlerocket hosts.
```

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,215 @@
...
+        db_opts.append(&mut vec!["-c", uuid_arg.as_str()]);
+    }
+
+    if let Some(device) = &args.device {
+        // Check if device is mounted
+        let mountpt = Command::new(FINDMNT)
+            .args(["-t", "xfs", "-f", "-n", "-o", "TARGET", device])
+            .output()
+            .expect("Couldn't find the mount point");
+        if mountpt.status.success() {
```
yeazelm: It looks like we use this in multiple places to switch logic between mounted and non-mounted targets. We should move this to a central place and import it where needed.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,215 @@
...
+            status += xfs_repair_output.status.code().unwrap();
+        }
+        exit(status);
+    } else {
+        println!("Couldn't parse target block-device");
+        exit(2);
+    }
+}
+
+fn is_required_offline(args: &Args) -> bool {
```
yeazelm: nit: I'd like a comment for this function - specifically how one can tell if the arg should be in that list.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,215 @@
...
+                xfs_db.args(["-l", logdev.as_str()]);
+            }
+            xfs_db.args(db_opts);
+            xfs_db.arg(device);
+
+            let xfs_db_output = xfs_db
+                .output()
+                .expect("Couldn't run xfs_db for given device.");
+            print!("{}", String::from_utf8(xfs_db_output.stdout).unwrap());
+            status = xfs_db_output.status.code().unwrap();
```
yeazelm: There are a lot of these sprinkled around where you do basically the same thing but with only a few slight differences. It might be worth creating a helper function for these clis that calls these things for you. It probably just needs the binary to call, the args to pass it, and an error message format string if it fails. Then it can print the output for you and return the status code. Its right on the borderline of over-doing it, but there does seem to be a lot of the same few lines repeated with nearly identical code.

sources/xfscli/src/bin/xfs_fsck.rs
```diff
@@ -0,0 +1,179 @@
...
+                    .output()
+                    .expect("Couldn't run xfs_repair again");
+                error = xfs_repair.status.code().unwrap();
+                fs::remove_dir("/tmp/repair_mnt").expect("could not delete");
+            }
+            exit(repair2fsck_code(error));
+        }
+
+        if auto {
+            println!("xfs fsck : XFS file system.");
```
yeazelm:
```suggestion
            println!("fsck.xfs : XFS file system.");
```
I think the original uses $0 which would work too, but I don't think `xfs fsck` is quite right?

sources/xfscli/src/bin/xfs_fsck.rs
```diff
@@ -0,0 +1,179 @@
...
+            exit(8);
+        }
+
+        if force {
+            let xfs_repair_result = Command::new(XFS_REPAIR)
+                .args(["-e", target])
+                .output()
+                .expect("Couldn't run xfs_repair for device");
+            println!("Performing XFS repair..");
+            let mut error = xfs_repair_result.status.code().unwrap();
```
yeazelm:
```suggestion
            let mut repair_exit_code = xfs_repair_result.status.code().unwrap();
```
Is this an error or simply the exit status from running xfs_repair. Error implies something bad happened.

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+use std::process::exit;
+use std::process::Command;
+use std::str;
+
+use argh::FromArgs;
+
+use xfscli::perform_xfs_repair;
+use xfscli::{BLKID, MOUNT, UMOUNT};
+
+// This CLI is rewritten in Rust for fsck.xfs shell script
```
arnaldo2792: Rather than explaining the why in this comment, explain what's the CLI for.

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// check the root filesystem as well when -A flag is set
+    #[argh(switch, short = 'p')]
+    check_root_fs: bool,
+    /// force repair filesystem
+    #[argh(switch, short = 'f')]
+    force: bool,
+    /// repair filesystem automatically
+    #[argh(switch, short = 'y')]
+    repair: bool,
+
```
arnaldo2792: nit: extra space

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        if force {
+            // Preform xfs_repair
+            println!("Performing XFS repair..");
+            let mut repair_exit_code = perform_xfs_repair(vec!["-e", target]);
+
+            // Clear the log, mount and unmount the XFS file system
+            // before xfs_repair
+            if repair_exit_code == 2 && repair {
```
arnaldo2792: You could give a name to exit code `2`.

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+                println!("Replaying log for {}", target);
+                let mkdir_result = fs::create_dir_all("/tmp/repair_mnt");
+                if mkdir_result.is_err() {
+                    exit(1);
```
arnaldo2792: Do you need to print out or give more details about the error? You could return a readable error and `Err` and the exit code will still be `1`

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+        if force {
+            // Preform xfs_repair
+            println!("Performing XFS repair..");
+            let mut repair_exit_code = perform_xfs_repair(vec!["-e", target]);
+
+            // Clear the log, mount and unmount the XFS file system
+            // before xfs_repair
+            if repair_exit_code == 2 && repair {
+                println!("Replaying log for {}", target);
+                let mkdir_result = fs::create_dir_all("/tmp/repair_mnt");
```
arnaldo2792: Will it be better to use a random directory instead of a targeted directory?

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+                let path = Path::new(&root);
+                let metadata = path.metadata().expect("Could not fetch metadata");
+                let file_type = metadata.file_type();
+
+                if !file_type.is_block_device() {
+                    let blkid_result = Command::new(BLKID)
+                        .args(["-t", &root, "-o", target])
+                        .output()
+                        .expect("Couldn't run blkid");
+                    root = String::from_utf8(blkid_result.stdout).unwrap();
```
arnaldo2792: All this code could be a separate function, that returns the actual root `Path` and root flags. Another note, remove all usage of `unwrap`and use `expect` instead if you don't want to use an error handling library.

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+                if !mount_result.status.success() {
+                    exit(1);
+                }
+
+                Command::new(UMOUNT)
+                    .args(["/tmp/repair_mnt"])
+                    .output()
+                    .expect("Couldn't run unmount");
+
+                println!("Performing XFS repair again after cleaning the log..");
```
arnaldo2792: Is the log cleared basically by mounting/unmounting?

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+                repair_exit_code = perform_xfs_repair(vec!["-e", target]);
+                fs::remove_dir("/tmp/repair_mnt")
+                    .expect("could not delete the /tmp/repair_mnt dir");
+            }
+            exit(repair2fsck_code(repair_exit_code));
+        }
+
+        if auto {
+            println!("fsck.xfs : XFS file system.");
```
arnaldo2792: Does `auto` basically mean `NOOP`? Or why is this print required?

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+        } else {
+            println!(
+                "If you wish to check the consistency of an XFS filesystem or \
+            repair a damaged filesystem, see xfs_repair(8)."
+            );
+        }
+        exit(0);
+    } else {
+        println!("Couldn't parse target block-device ");
```
arnaldo2792: Shouldn't this branch exit non-zero?

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+fn repair2fsck_code(error: i32) -> i32 {
+    match error {
+        0 => 0, // everything is ok
```
arnaldo2792: It would be better to give meaningful names to each exit code, probably with an ENUM. And the same for the returned values.

sources/xfscli/src/bin/xfs_info.rs
```diff
@@ -0,0 +1,62 @@
...
+fn main() {
+    let args: Args = argh::from_env();
+
+    // Return the version for arg version as true
+    if args.version {
+        let xfs_spaceman_result = Command::new(XFS_SPACEMAN)
+            .args(["-p", "xfs_info", "-V"])
+            .output()
+            .expect("Couldn't run xfs_spaceman for version info");
+        print!("{}", String::from_utf8(xfs_spaceman_result.stdout).unwrap());
```
arnaldo2792: Same as all other places, `expect` over `unwrap`.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,212 @@
+use std::process::exit;
```
arnaldo2792: nit: probably this commit should be before the other two commits, since in this commit `query_xfs_db` and `get_mount_point` were defined.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// set the UUID of the filesystem to uuid
+    #[argh(option, short = 'U')]
+    uuid: Option<String>,
+    /// add or remove features on an existing V5 filesystem
+    #[argh(option, short = 'O')]
+    v5_feature: Option<String>,
+    /// prints the version number and exits.
+    #[argh(switch, short = 'V')]
+    version: bool,
+
```
arnaldo2792, line 69: nit: extra space.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,212 @@
...
+    } else {
+        println!("Couldn't parse target block-device");
+        exit(2);
+    }
+}
+
+// Filesystem should not be mounted with args in this function
+// For more reference check require_offline variable in
+// https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/db/xfs_admin.sh
+fn is_required_offline(args: &Args) -> bool {
```
arnaldo2792: Why not making this a function of `Args`? so that you can call `args.is_required_offline`?

sources/xfscli/src/lib.rs
```diff
@@ -0,0 +1,44 @@
+use std::process::Command;
+
+pub static BLKID: &str = "/usr/sbin/blkid";
+pub static FINDMNT: &str = "findmnt";
```
arnaldo2792:
```suggestion
pub static FINDMNT: &str = "/usr/bin/findmnt";
```

sources/xfscli/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+pub static XFS_SPACEMAN: &str = "/usr/sbin/xfs_spaceman";
+
+// Check if target is mounted and return mount point
+pub fn get_mount_point(target: &String) -> Option<String> {
+    let mountpt = Command::new(FINDMNT)
+        .args(["-t", "xfs", "-f", "-n", "-o", "TARGET", target])
+        .output()
+        .expect("Couldn't find the mount point");
+
+    if mountpt.status.success() {
```
arnaldo2792, line 25: A comment explaining that any error is interpreted as if the mount is missing?

sources/xfscli/src/bin/fsck_xfs.rs
vyaghras: Resolved offline

sources/xfscli/src/bin/fsck_xfs.rs
vyaghras: As per my understanding the [script](https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/fsck/xfs_fsck.sh) is replaying the logs by mounting and unmounting the device when repair is unsuccessful . @markusboehme / @foersleo  Will it be possible for you to share your thoughts on this?

vyaghras: As per my understanding the [script](https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/fsck/xfs_fsck.sh) is replaying the logs by mounting and unmounting the device when repair is unsuccessful . @markusboehme / @foersleo  Will it be possible for you to share your thoughts on this?

sources/xfscli/src/bin/fsck_xfs.rs
vyaghras: Yes, in fsck.xfs with auto=true, means NOOP. This print is required to replicate the behavior in  xfs shell [script](https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/fsck/xfs_fsck.sh)

sources/xfscli/src/bin/fsck_xfs.rs
markusboehme: Yes, a mount/unmount cycle is what's needed here. XFS is a journaling file system, i.e. it records intended changes to the file system before actually performing them. Should the IO be interrupted (power loss, machine crashes, disk is unresponsive, etc.) the file system can still be brought into a consistent state by performing the intended actions on the next mount ("replaying/clearing the journal"). The actual work here is performed by the mount; the unmount just needs to happen eventually so the device can be touched directly without interfering with normal accesses.

sources/xfscli/src/error.rs
```diff
@@ -0,0 +1,55 @@
+use snafu::Snafu;
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub))]
+#[repr(i32)]
+pub enum Error {
+    #[snafu(display("xfs fsck errror: xfs_repair was not found!"))]
+    CommandUnavailable,
```
yeazelm: nit: you could make this more generic and take an argument which is the command that was not found. This name implies its generic but only works for `xfs_repair`. In this case, it looks like that is the only place its used so its a nit.

sources/xfscli/src/error.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    #[snafu(display("{} does not exist", target))]
+    DeviceNotFound { target: String },
+
+    #[snafu(display(
+        "xfs fsck errror: The filesystem log is dirty, mount it to recover \
+                the log. If that fails, refer to the section DIRTY LOGS in the \
+                xfs_repair manual page."
```
yeazelm: This doesn't do what you think it does, I think the `\` is going to be included in the long string. Can you confirm the correct output shows up for this error?

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,198 @@
...
+        // Preform xfs_repair
+        println!("Performing XFS repair..");
+        let mut repair_exit_code = perform_xfs_repair(vec!["-e", target]);
+
+        // Clear the log, mount and unmount the XFS file system
+        // before xfs_repair
+        if repair_exit_code == XfsRepairResponseCode::DirtyLogs && repair {
+            println!("Replaying log for {}", target);
+
+            // Using random directory (/tmp/repair_mntxxxx) instead of a targeted directory(/tmp/repair_mnt)
```
yeazelm: nit, add a space here
```suggestion
            // Using random directory (/tmp/repair_mntxxxx) instead of a targeted directory (/tmp/repair_mnt)
```

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,198 @@
...
+                .context(error::MakeDirSnafu)?;
+
+            let mnt_dir_path = mnt_dir
+                .path()
+                .to_str()
+                .expect("Unable to get path of temporary dir");
+
+            let kernel_params = get_root_and_rootflags(target);
+            let root = kernel_params.root;
+            let root_flags = kernel_params.root_flags;
```
yeazelm: nit: you could use these directly since the names are consistent to avoid an additional variable declaration

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,239 @@
...
+        db_opts.append(&mut vec!["-c", uuid_arg.as_str()]);
+    }
+
+    let device = &args.device.as_ref().context(error::ParseTargetSnafu)?;
+
+    // Exit if filesystem is mounted and require offline for queried arguments
+    if let Some(mpt) = get_mount_point(device) {
+        // filesystem is mounted
+        if args.is_required_offline() {
+            return Err(Error::MountedFilesystem { mount_point: mpt });
```
yeazelm, line 178: I expected this to be `MountedFilesystemSnafu`?

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,239 @@
...
+
+        if !io_opts.is_empty() {
+            let mut xfs_io = Command::new(XFS_IO);
+            xfs_io.args(["-p", "xfs_admin"]);
+            xfs_io.args(io_opts);
+            xfs_io.arg(mpt.as_str().trim());
+
+            let xfs_io_result = xfs_io
+                .output()
+                .expect("Couldn't run xfs_io for given device");
```
yeazelm: Do we want to remain consistent and have all of the errors be Snafu? I think it would be more around completeness, Snafu isn't probably offering a ton of value in these specifically, but we ended up adding it most other places. I can be convinced its not worth it.

sources/xfscli/src/bin/xfs_admin.rs
vyaghras: I think Snafu is needed when we use context?

sources/xfscli/src/error.rs
vyaghras: Validated that '\' is not the part of output message :`xfs fsck errror: The filesystem log is dirty, mount it to recover the log. If that fails, refer to the section DIRTY LOGS in the xfs_repair manual page.`

sources/xfscli/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+// Check if target is mounted and return mount point
+pub fn get_mount_point(target: &String) -> Option<String> {
+    let mountpt = Command::new(FINDMNT)
+        .args(["-t", "xfs", "-f", "-n", "-o", "TARGET", target])
+        .output()
+        .expect("Couldn't find the mount point");
+
+    // All errors will be interpreted as failure in finding the mount point
+    if mountpt.status.success() {
+        Some(String::from_utf8(mountpt.stdout).expect("Could't read the mount point"))
```
cbgbt: I think this is generally where you'd rather have a return type like `Result<Option<String>>`, especially since we are already defining our own crate-wide error type.

sources/xfscli/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+        .expect("Unable to read xfs_repair exit status code");
+
+    match code {
+        0 => XfsRepairResponseCode::Ok,
+        1 => XfsRepairResponseCode::RepairFailure,
+        2 => XfsRepairResponseCode::DirtyLogs,
+        4 => XfsRepairResponseCode::MetadataRepair,
+        127 => XfsRepairResponseCode::CommandUnavailable,
+        code => XfsRepairResponseCode::Unknown(code),
+    }
```
cbgbt: This might be more idiomatically expressed using `impl From<i32> for XfsRepairResponseCode`, or even `impl From<ExitStatus>`

sources/xfscli/src/bin/xfs_admin.rs
cbgbt: It's generally best to avoid panicking, and you usually should be returning some kind of `Result` type instead.

*Usually* code that panics is best only if you can logically prove that it could never happen, e.g. by checking the preconditions of the panic. Consider:

```rust
let my_option = Some("hello");
if my_option.is_some() {
    let inner_string = my_option.unwrap(); // We logically know this can't panic.
}
```

Even in these cases, rewriting the code to avoid calling a function that *could theoretically* panic is much more future-proof:

```rust
if let Some(inner_string) = my_option {
    // now do the thing
}
```

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    let device = &args.device.as_ref().context(error::ParseTargetSnafu)?;
+
+    // Collect all the arguments for XFS_DB, XFS_IO and XFS_REPAIR commands
+    // by parsing the CLI inputs.
+    let mut db_opts: Vec<&str> = vec![];
+    let mut io_opts: Vec<&str> = vec![];
+    let mut repair_opts: Vec<&str> = vec![];
+
+    let mut lazycount_arg = "lazycount=".to_string();
```
cbgbt: These mutable strings are a little unusual to me, but I guess it's like this so that you can mutate the value while having it guaranteed to live long enough to keep an `&str` reference to it?

sources/xfscli/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[[bin]]
+name = "xfs_info"
+
+[[bin]]
+name = "xfs_admin"
+
+[[bin]]
+name = "fsck_xfs"
+path = "src/bin/fsck_xfs.rs"
```
cbgbt: I don't think these `[[bin]]` entries are required.

EDIT: I am seeing that `fsck_xfs` is renamed. Nevermind, we probably need that one.

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,199 @@
...
+
+            println!("Performing XFS repair again after cleaning the log..");
+            repair_exit_code = perform_xfs_repair(vec!["-e", target]);
+
+            let mnt_dir_close_msg = format!("Could not delete the {} dir", mnt_dir_path);
+            mnt_dir.close().expect(&mnt_dir_close_msg);
+        }
+
+        let metadata_repaired = match repair_exit_code {
+            XfsRepairResponseCode::Ok => Ok(false),
```
cbgbt: I don't know the details of the underlying software, but I'm a bit surprised that `Ok` translates to `false`.

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+fn main() {
+    let args: Args = argh::from_env();
+    match run(args) {
+        Ok(exit_code) => exit(exit_code),
+        Err(e) => {
+            eprintln!("{}", e);
+            exit(e.exit_code());
```
cbgbt: nit: For this program it doesn't matter, but just as a tip, check out this part of the docs on [`std::process::exit`](https://doc.rust-lang.org/std/process/fn.exit.html):

> Note that because this function never returns, and that it terminates the process, no destructors on the current stack or any other thread’s stack will be run. If a clean shutdown is needed it is recommended to only call this function at a known point where there are no more destructors left to run; or, preferably, simply return a type implementing Termination (such as ExitCode or Result) from the main function and avoid this function altogether

I think given all this, it's probably more hygienic to do something like [this example](https://doc.rust-lang.org/std/process/struct.ExitCode.html#examples) and return an `ExitCode` from `main`.

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,199 @@
...
+            XfsRepairResponseCode::DirtyLogs => Err(Error::DirtyLogs),
+            XfsRepairResponseCode::MetadataRepair => Ok(true),
+            XfsRepairResponseCode::CommandUnavailable => Err(Error::CommandUnavailable {
+                cli: "fsck".to_string(),
+                command: "xfs_repair".to_string(),
+            }),
+            XfsRepairResponseCode::Unknown(code) => Err(Error::UnrecognizedExitCode { code }),
+        }?;
+
+        return Ok(if metadata_repaired { 1 } else { 0 });
```
cbgbt: If you do move to using `ExitStatus` (which is probably a bit more clear than using `Result<i32>`), this could be a good spot to use `ExitCode::SUCCESS` and `ExitCode::FAILURE`.

sources/xfscli/src/bin/xfs_admin.rs
vyaghras: yes, we need these values to live long enough in order to pass as an argument.

sources/xfscli/src/bin/fsck_xfs.rs
vyaghras: This function return success in 2 cases after running xfs_repair. 
1. When filesysten itself is repaired 
2. When metadata has been repaired.

This match is to understand if the success is because of metadata repair or filesystem repair. Then later `if metadata_repaired { 1 } else { 0 }` this returns the exact result with the expected code.

sources/xfscli/src/bin/fsck_xfs.rs
vyaghras: We need to return 1 as success code following https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git/tree/fsck/xfs_fsck.sh, so I am not sure if we should return as ExitCode::FAILURE.

cbgbt: Ah wow, I wasn't expecting that interface.

I'd personally still prefer to return `ExitCode`, but manually provide the value yourself, e.g.

```rust
ExitCode::from(1)
```

sources/xfscli/src/bin/xfs_admin.rs
```diff
@@ -0,0 +1,239 @@
...
+    // by parsing the CLI inputs.
+    let mut db_opts: Vec<&str> = vec![];
+    let mut io_opts: Vec<&str> = vec![];
+    let mut repair_opts: Vec<&str> = vec![];
+
+    let mut lazycount_arg = "lazycount=".to_string();
+    if let Some(lazy_counters) = &args.lazy_counters {
+        lazycount_arg.push_str(lazy_counters.to_string().as_str());
+        repair_opts.append(&mut vec!["-c", lazycount_arg.as_str()]);
+    }
```
cbgbt: nit: I might recommend (for this value and others like it) that you compute the entire value inside the conditional.

```suggestion
    let lazycount_arg;
    if let Some(lazy_counters) = &args.lazy_counters {
        lazycount_arg = format!("lazycount={}", lazy_counters);
        repair_opts.append(&mut vec!["-c", lazycount_arg.as_str()]);
    }
```

This removes the need for mutability, and I also find these to be a little confusing, especially the one below:

```rust
    let mut label_arg_db = "label ".to_string();
    let mut label_arg_io = "label ".to_string();
```

vyaghras:
Update in ⬆️  push:
1. Changing all the expect errors in to errors using snafu.
2. Moving all the fsck related errors in a separate file.(To keep it clean as that command has specific error return codes). 
3.  Change error codes for all the general errors like Command unsuccessful, Unable to read data to 8( Operational error).

sources/xfscli/src/bin/fsck_xfs.rs
vyaghras: Changed all the i32 return types to ExitCode and return Exitcode from main function.

sources/xfscli/src/bin/xfs_admin.rs
vyaghras: Handled all the error using Snafu.

sources/xfscli/src/bin/fsck_xfs.rs
```diff
@@ -0,0 +1,219 @@
...
+            )?;
+
+            let mut mount_args = vec![target.as_str(), mnt_dir_path];
+            if basename_dev == basename_root && !kernel_params.root_flags.is_empty() {
+                mount_args.push(kernel_params.root_flags.as_str());
+            }
+
+            let mount_status = Command::new(MOUNT).args(mount_args).status().context(
+                fsckerror::CommandFailureSnafu {
+                    command: "Mount".to_string(),
```
cbgbt:
```suggestion
                    command: "mount".to_string(),
```

sources/xfscli/src/fsckerror.rs
cbgbt, line 1: I would make sure this is a submodule of `fsck_xfs`, either by adding a `mod fsckerror { ... }` within the file, or using a multi-file binary like so:

```
bin
└── fsck_xfs
    ├── error.rs
    └── main.rs
```

vyaghras: Moved fsck cli and error file in a module.

vyaghras:
Update in ⬆️  push:
Moved fsck cli(main.rs) and error file in a module fsck_xfs.

sources/xfscli/src/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    xfs_db_result
+        .status
+        .code()
+        .context(error::ParseStatusCodeSnafu {
+            command: "xfs_db".to_string(),
+        })
+}
+
+// Query xfs spaceman to get xfs information
```
cbgbt: nit: I'm sorry to have missed this in a previous pass, but these should be docstring comments.

cbgbt (approved):
Nice work! 🚀

sources/xfscli/src/lib.rs
vyaghras: Changed these comments to Docstring comments

vyaghras:
Update in ⬆️ push:
Changed all comments in lib.rs to Docstring comments

cbgbt (approved):

yeazelm (approved):
LGTM! 🚢

PR #3451, opened 2023-09-13
disable synthesized hostname in systemd-resolved

bcressey:
**Issue number:**
N/A - reported by @etungsten 

**Description of changes:**
Disable the synthesized hostname feature in `systemd-resolved`.

Otherwise, the reverse lookup in `netdog` will return "linux" if no other name can be resolved by DNS. This causes problems when trying to join several nodes to a Kubernetes cluster, since they all attempt to register as "linux".

The expected (historical) behavior is that the system hostname winds up as the primary interface's IP address in this case. This is not ideal and causes other headaches: for example, the short name of the host is the first part of the IP in dotted decimal notation. But in the absence of a setting or reverse DNS, it's at least a potentially predictable and unique identifier.

**Testing done:**

Before:
```
bash-5.1# resolvectl query 198.19.33.182
198.19.33.182: linux                           -- link: eth0
               linux.local                     -- link: eth0

-- Information acquired via protocol DNS in 8.5ms.
-- Data is authenticated: yes; Data was acquired via local or encrypted transport: yes
-- Data from: synthetic
```

After:
```
bash-5.1# resolvectl query 198.19.16.218
198.19.16.218: resolve call failed: Name '218.16.19.198.in-addr.arpa' not found
```

**Terms of contribution:**


etungsten (approved):

yeazelm (approved):
Nice fix!

zmrow (approved):

PR #3455, opened 2023-09-14
Revert "release: consult resolved for host lookups"

### LINKED ISSUE #3457: kubelet now uses cgo for all DNS queries
**Image I'm using:**
aws-k8s-1.28

When bootstrapping a cluster, `kubelet` should try multiple entries for the same host in `/etc/hosts`:
```
10.0.10.1 asdfasdfasfd.aws.dev
10.0.10.2 asdfasdfasfd.aws.dev
10.0.10.3 asdfasdfasfd.aws.dev
```

**What actually happened:**
`kubelet` tries to connect to the first IP repeatedly. 

Using `GODEBUG=netdns=2` we can see that `cgo` rather than `netgo` is now used.

Sample program:
```go
package main

import (
	"fmt"
	"net"
	"os"
)

func main() {
	addrs, err := net.LookupHost(os.Args[1])
	if err != nil {
		fmt.Println(err)
	} else {
		fmt.Println(addrs)
	}
}
```

List multiple IPs for the same host:
```
apiclient apply <<EOF
[settings.network]
hosts = [
 ["10.0.10.1", ["asdfasdfasfd.aws.dev"]],
 ["10.0.10.2", ["asdfasdfasfd.aws.dev"]],
 ["10.0.10.3", ["asdfasdfasfd.aws.dev"]],
]
EOF
```

Query the host:
```
# GODEBUG=netdns=2 dns-test asdfasdfasfd.aws.dev
go package net: confVal.netCgo = false  netGo = false
go package net: dynamic selection of DNS resolver
go package net: hostLookupOrder(asdfasdfasfd.aws.dev) = cgo
[10.0.10.1]
```

This happens because `netgo` sees the new "resolve" directive in `nsswitch.conf`, and knows it cannot use that NSS plugin so it falls back to `cgo`.

zmrow:
**Issue number:**
Fixes #3457

**Description of changes:**
```    
    This reverts commit dfbf90b8615d237f6ccf774ba8af847cae2c4ae8.
    
    Go's netgo resolver doesn't understand the "resolve" directive in
    nsswitch.conf, so it falls back to cgo, which uses glibc to look up
    hosts.  The glibc lookup behaves differently in certain cases, such as
    multiple records for the same host in `/etc/hosts`.  Avoid the behavior
    change by using the previous order.  This still results in queries going
    through systemd-resolved, just not via the nss plugin.
```

**Testing done:**
Before the change:
```
bash-5.1# resolvectl statistics
DNSSEC supported by current servers: no

Transactions             
Current Transactions: 0
  Total Transactions: 756
                         
Cache                    
  Current Cache Size: 2
          Cache Hits: 305
        Cache Misses: 451
                         
DNSSEC Verdicts          
              Secure: 0
            Insecure: 0
               Bogus: 0
       Indeterminate: 0
```

After the change:
```
bash-5.1# resolvectl statistics
DNSSEC supported by current servers: no

Transactions             
Current Transactions: 0
  Total Transactions: 236
                         
Cache                    
  Current Cache Size: 10
          Cache Hits: 144
        Cache Misses: 92
                         
DNSSEC Verdicts          
              Secure: 0
            Insecure: 0
               Bogus: 0
       Indeterminate: 0
```

**Terms of contribution:**

### DIFF
--- a/packages/release/nsswitch.conf
+++ b/packages/release/nsswitch.conf
@@ -1,4 +1,4 @@
 passwd: files
 group: files
 shadow: files
-hosts: files resolve [!UNAVAIL=return] dns
+hosts: files dns


stmcginnis (approved):

zmrow:
^ Update commit message (and fix a typo in said message)

bcressey:
If I repeat the test case in #3457 with this change, I get the expected output:
```
# GODEBUG=netdns=2 dns-test asdfasdfasfd.aws.dev
go package net: confVal.netCgo = false  netGo = false
go package net: dynamic selection of DNS resolver
go package net: hostLookupOrder(asdfasdfasfd.aws.dev) = files,dns
[10.0.10.1 10.0.10.2 10.0.10.3]
```

bcressey (approved):

yeazelm (approved):

Issue #2409, opened 2022-09-12
Add additional IPVS modules like `ip_vs_lc`


etungsten:
> Not that it's part of this thread specifically - but it would be nice to add a few more of the reecommended IPVS modules like `ip_vs_sh`.

_Originally posted by @diranged in https://github.com/bottlerocket-os/bottlerocket/issues/1726#issuecomment-1242881788_

etungsten:
Hi @diranged, I just checked our 5.10 kernel config and it does seem like we have most of the IPVS modules enabled (including `ip_vs_sh`):
```
 # IPVS transport protocol load balancing support
 #
 CONFIG_IP_VS_PROTO_TCP=y
 CONFIG_IP_VS_PROTO_UDP=y
 CONFIG_IP_VS_PROTO_AH_ESP=y
 CONFIG_IP_VS_PROTO_ESP=y
 CONFIG_IP_VS_PROTO_AH=y
 CONFIG_IP_VS_PROTO_SCTP=y
 
  #
  # IPVS scheduler
  #
  CONFIG_IP_VS_RR=m
  CONFIG_IP_VS_WRR=m
  CONFIG_IP_VS_LC=m
  CONFIG_IP_VS_WLC=m
  CONFIG_IP_VS_FO=m
  CONFIG_IP_VS_OVF=m
  CONFIG_IP_VS_LBLC=m
  CONFIG_IP_VS_LBLCR=m
  CONFIG_IP_VS_DH=m
  CONFIG_IP_VS_SH=m
  CONFIG_IP_VS_MH=m
  CONFIG_IP_VS_SED=m
  CONFIG_IP_VS_NQ=m
  
  #
  # IPVS SH scheduler
  #
  CONFIG_IP_VS_SH_TAB_BITS=8
  
  #
  # IPVS MH scheduler
  #
  CONFIG_IP_VS_MH_TAB_INDEX=12
  
  #
  # IPVS application helper
  #
  CONFIG_IP_VS_FTP=m
  CONFIG_IP_VS_NFCT=y
  CONFIG_IP_VS_PE_SIP=m
```

Please feel free to re-open this issue if you see we're missing anything!

diranged:
@etungsten You're right ... the one that fails for me is `lc`:

```
E0913 00:17:27.243643       1 server_others.go:107] "Can't use the IPVS proxier" err="IPVS proxier will not be used because the following required kernel modules are not loaded: [ip_vs_lc]"
```

[...]

Issue #3052, opened 2023-04-26
br_netfilter kernel module needed for kubeadm - change in bridge kernel module


tommie:
**Image I'm using:** metal-k8s-1.26-v1.13.2

**What I expected to happen:** `kubeadm init` runs happily.

**What actually happened:** The preflight check `bridge-nf-call-iptables` fails with `/proc/sys/net/bridge/bridge-nf-call-iptables does not exist`

**How to reproduce the problem:** Pretty much any invocation of `kubeadm init` should cause this.

The check was added here: https://github.com/kubernetes/kubernetes/commit/fb4ee4debafce916e3b17ac20707e1fa1bb94630

IIRC, kubeadm does this check because Kubelet uses that proc entry, but I'm not sure.

The problem is that the `bridge` module requires `br_netfilter`: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/bridge/br.c#n424

Confirmed the kmod kit says [`CONFIG_BRIDGE_NETFILTER=m`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/Kconfig#n188) for v1.13.3. I can't say if this module should be loaded by default, or simply modified to `y`.

(I'm also surprised no one else has hit this preflight check before me, so I might have missed something elsewhere...)

stmcginnis:
Thanks for filing this @tommie. Typically Bottlerocket is only used as a worker node, so I don't think it is common for someone to be running `kubeadm init`. Just out of curiosity, can you describe a little more how you are using Bottlerocket as the control plane? Are you spinning up your own cluster manually?

tommie:
I should really try to figure out why kubeadm wants this. IIRC, it was actually for Kubelet, but it's possible it's for the control plane. My notes don't say, sadly.

I'm building a script to bootstrap a cluster from scratch on Hetzner Cloud/Robot. K8s is complicated enough that I want to document it in a script, instead of just following someone's how-to. My use-case is a really small cluster with just two nodes or so. Pretty much, I just want to move away from my servers being "precious", and orchestrate everything with K8s, even if it's a small workload. The idea is to create a system that is easy to bootstrap, but also easily extensible.

I recently worked with a client that was running a Windows server with MS SQL and some pretty horrible DevOps practices. It would have been so awesome to just say "here is a cheap two-node K8s cluster, now let's focus on CI/CD instead of petting the Windows server for every release."

[...]

PR #3460, opened 2023-09-15
models: add new setting to auto-load kernel modules

foersleo:
**Issue number:** https://github.com/bottlerocket-os/bottlerocket/discussions/3050, https://github.com/bottlerocket-os/bottlerocket/issues/2409

**Description of changes:**

Loading specific kernel modules can be necessary to use specific features on your node. One example would be loading the correct module for your choice of scheduling algorithm for ipvs.

Loading kernel modules on boot currently requires the use and maintenance of an init-container image to load the desired kernel modules on boot. We can simplify that by adding the additional setting `load-on-boot` for kernel modules and utilizing the already available systemd-modules-load service.

Module auto-loading is in conflict with blocking modules from loading through the sibling setting `allowed`. Hence, do not auto-load a module on boot if the module is not allowed at the same time. Blocking takes precedence, as it is the more prohibitive operation.

**Testing done:**

I have run tests with some variations of module specific options on a local vm in my dev-system and results were as expected:

* autoloading and allowing the module to be loaded:
  ```
  [settings.kernel.modules.ip_vs_lc]
  allowed = true
  load-on-boot = true
  ```
  Results in autoloading the required module and dependency modules when re-triggering systemd-modules-load during boot:
  ```
  [  OK  ] Stopped Load Kernel Modules.
           Stopping Load Kernel Modules...
           Starting Load Kernel Modules...
  [    5.244013] IPVS: Registered protocols (TCP, UDP, SCTP, AH, ESP)
  [    5.244557] IPVS: Connection hash table configured (size=4096, memory=32Kbytes)
  [    5.245233] IPVS: ipvs loaded.
  [    5.250915] IPVS: [lc] scheduler registered.
  [  OK  ] Finished Load Kernel Modules.
  [...]
  bash-5.1# lsmod | grep ip_vs
  ip_vs_lc               16384  0
  ip_vs                 192512  2 ip_vs_lc
  bash-5.1# cat /etc/modules-load.d/modules-load.conf 
  ip_vs_lc
  ```
* conflict of blocking and auto-loading does not add the module to the config file and does not load the module:
  ```
  [settings.kernel.modules.ip_vs_lc]
  allowed = false
  load-on-boot = true
  ```
  Results in not autoloading the and not adding it to the config file:
  ```
  [  OK  ] Stopped Load Kernel Modules.
           Stopping Load Kernel Modules...
           Starting Load Kernel Modules...
  [  OK  ] Finished Load Kernel Modules.
  [...]
  bash-5.1# lsmod | grep ip_vs
  bash-5.1# cat /etc/modules-load.d/modules-load.conf
  ```

* Accidental overriding of a blocking setting is ruled out by toml not allowing to define the same table twice. The userdata configuration system will be derailed with a parser error for providing invalid user-data.

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1164,6 +1164,17 @@ Here are the metrics settings:
   allowed = false
   ```
 
+* `settings.kernel.modules.<name>.autoload`: Whether the named kernel modules shall be loaded automatically.
+  **Important note:** this setting needs to be used in conjunction with the `allowed` setting for the same module to ensure we are not auto-loading a module that is blocked.
+
+  Example user data for auto-loading a kernel module on boot:
+
+  ```toml
+  [settings.kernel.modules.ip_vs_lc]
+  allowed = true
+  autoload = true
+  ```
+
 * `settings.kernel.sysctl`: Key/value pairs representing Linux kernel parameters.
   Remember to quote keys (since they often contain ".") and to quote all values.
 
--- a/sources/api/migration/migrations/v1.16.0/kernel-modules-autoload-restart/src/main.rs
+++ b/sources/api/migration/migrations/v1.16.0/kernel-modules-autoload-restart/src/main.rs
@@ -1,0 +1,23 @@
+use migration_helpers::common_migrations::{ListReplacement, ReplaceListsMigration};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new `autoload` setting to `settings.kernel.modules`, which needs
+/// re restart of `systemd-modules-load.services`.
+fn run() -> Result<()> {
+    migrate(ReplaceListsMigration(vec![ListReplacement {
+        setting: "services.kernel-modules.restart-commands",
+        old_vals: &[],
+        new_vals: &["/usr/bin/systemctl try-restart systemd-modules-load"],
+    }]))
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/models/shared-defaults/defaults.toml
+++ b/sources/models/shared-defaults/defaults.toml
@@ -129,12 +129,16 @@ restart-commands = ["/usr/bin/corndog sysctl"]
 affected-services = ["sysctl"]
 
 [services.kernel-modules]
-configuration-files = ["modprobe-conf"]
-restart-commands = []
+configuration-files = ["modprobe-conf", "modules-load"]
+restart-commands = ["/usr/bin/systemctl try-restart systemd-modules-load"]
 
 [configuration-files.modprobe-conf]
 path = "/etc/modprobe.d/modprobe.conf"
 template-path = "/usr/share/templates/modprobe-conf"
+
+[configuration-files.modules-load]
+path = "/etc/modules-load.d/modules-load.conf"
+template-path = "/usr/share/templates/modules-load"
 
 [metadata.settings.kernel.modules]
 affected-services = ["kernel-modules"]
--- a/Release.toml
+++ b/Release.toml
@@ -1,4 +1,4 @@
-version = "1.15.0"
+version = "1.16.0"
 
 [migrations]
 "(0.3.1, 0.3.2)" = ["migrate_v0.3.2_admin-container-v0-5-0.lz4"]
@@ -233,4 +233,6 @@ version = "1.15.0"
     "migrate_v1.15.0_public-control-container-v0-7-4.lz4",
     "migrate_v1.15.0_log4j-hotpatch-enabled-metadata.lz4",
     "migrate_v1.15.0_deprecate-log4j-hotpatch-enabled.lz4",
+]
+"(1.15.0, 1.16.0)" = [
 ]
--- a/Release.toml
+++ b/Release.toml
@@ -235,4 +235,8 @@ version = "1.16.0"
     "migrate_v1.15.0_deprecate-log4j-hotpatch-enabled.lz4",
 ]
 "(1.15.0, 1.16.0)" = [
+    "migrate_v1.16.0_kernel-modules-autoload-configs.lz4",
+    "migrate_v1.16.0_kernel-modules-autoload-files.lz4",
+    "migrate_v1.16.0_kernel-modules-autoload-restart.lz4",
+    "migrate_v1.16.0_kernel-modules-autoload-settings.lz4",
 ]
--- a/packages/release/modules-load.template
+++ b/packages/release/modules-load.template
@@ -1,0 +1,9 @@
+{{#if settings.kernel.modules}}
+{{#each settings.kernel.modules}}
+{{#if this.allowed}}
+{{#if this.autoload}}
+{{@key}}
+{{/if}}
+{{/if}}
+{{/each}}
+{{/if}}
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -23,6 +23,7 @@ Source204: modprobe-conf.template
 Source205: netdog.template
 Source206: aws-config
 Source207: aws-credentials
+Source208: modules-load.template
 
 Source1001: multi-user.target
 Source1002: configured.target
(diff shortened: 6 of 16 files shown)


stmcginnis (changes requested):
Awesome, this looks like a big usability improvement for anyone needing to load additional modules.

With the new setting, you will need to add a migration to add/remove that setting on upgrade/downgrade. There is an example of doing something similar here: https://github.com/bottlerocket-os/bottlerocket/pull/2930/commits/ac1e6e98a038de2a25d3e5c6a0fdedb89abd3142

Other than the missing migration, this looks great!

README.md
```diff
@@ -1164,6 +1164,17 @@ Here are the metrics settings:
   allowed = false
   ```
 
+* `settings.kernel.modules.<name>.load-on-boot`: Whether the named kernel modules shall be loaded on boot.
```
markusboehme: This could call out that the module will be loaded once this setting is applied, not just on next boot. Which brings me to my nitpick with the name: "load-on-boot" is not quite accurate. The PR's title calls it "auto-load", and I think this is the better name.

foersleo: Yeah, as should be evident I struggled with this a bit myself. With changes now triggering the reload that is right. Originally I did not have the restart-command and was hoping for it to be early enough to be just caught by the usual run of `systemd-modules-load` during boot. Will do the name cleanups accordingly.

foersleo:
I will have to have a look at the migrations part. I have not done that before.

foersleo:
:arrow_up: tended the renaming to `autoload` as pointed out by Markus.

I will add the migrations part next week as this seems to be a little bit more specialized with a field in a variably named table. Will have to do some reading.

foersleo:
:arrow_up: Added the migrations and bumped the Release version before, now that 1.15.0 is released and these migrations will be relevant when moving to the next version.

I ended up needing 4 separate migrations. I was contemplating if it would make more sense to have one big custom migration or not, but maybe those with more experience writing migrations can chime in here. I was also wondering if it would be possible to do multiple `migrate` calls the `run` function of one migration and combine their results, although I did not find precedent for that in existing migrations. So probably it is prohibitive to do that? I would be interested in getting some insight on that from people that know more about the migrations system.

I did test the migrations, the following is the gist of what happened:

**Upgrade:**
```
bash-5.1# updog update -i 1.16.0 -r -n
Starting update to 1.16.0
Reboot scheduled for Tue 2023-09-19 09:58:57 UTC, use 'shutdown -c' to cancel.
Update applied: aws-k8s-1.27 1.16.0
[...]
bash-5.1# cd /var/lib/bottlerocket/datastore/current/live/
bash-5.1# cd settings/kernel/
bash-5.1# cat modules.affected-services 
["kernel-modules"]
bash-5.1# cd ../../
bash-5.1# cd configuration-files/
bash-5.1# cat modules-load/path 
"/etc/modules-load.d/modules-load.conf"
bash-5.1# cat modules-load/template-path 
"/usr/share/templates/modules-load"
bash-5.1# cd ../
bash-5.1# cd services/kernel-modules/
bash-5.1# cat configuration-files 
["modprobe-conf","modules-load"]
bash-5.1# cat restart-commands 
["/usr/bin/systemctl restart systemd-modules-load"]
```
**Downgrade (setup one module for autoload previously to see that these keys get cleaned up correctly as well):**
```
bash-5.1# apiclient set kernel.modules.ip_vs_lc.allowed=true kernel.modules.ip_vs_lc.autoload=true
bash-5.1# cd settings/kernel/modules
bash-5.1# ls
ip_vs_lc
bash-5.1# cd ip_vs_lc/
bash-5.1# ls
allowed  autoload
bash-5.1# cat *
true
true
bash-5.1# lsmod | grep ip_vs
ip_vs_lc               16384  0
ip_vs                 192512  2 ip_vs_lc
bash-5.1# signpost rollback-to-inactive
bash-5.1# reboot
[...]
bash-5.1# cd /var/lib/bottlerocket/datastore/current/live/
bash-5.1# cd settings/kernel/modules/ip_vs_lc/ 
bash-5.1# ls
allowed
bash-5.1# lsmod | grep ip_vs
bash-5.1# cd ../../../../
bash-5.1# cd services/kernel-modules/
bash-5.1# cat configuration-files 
["modprobe-conf"]
bash-5.1# cat restart-commands 
[]
bash-5.1# cd ../../
bash-5.1# cd configuration-files/
bash-5.1# ls modules-load
ls: cannot access 'modules-load': No such file or directory

```

foersleo:
:arrow_up: force push fixed some formatting issues.

foersleo:
Last force pushed fixed additional clippy warnings.

bcressey:
> One example would be loading the correct module for your choice of scheduling algorithm for ipvs.

Do we have other examples? I noted in #3391 that the kernel seems to auto-load the IPVS modules.

sources/api/migration/migrations/v1.16.0/kernel-modules-autoload-restart/src/main.rs
```diff
@@ -0,0 +1,23 @@
...
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// We added a new `autoload` setting to `settings.kernel.modules`, which needs
+/// re restart of `systemd-modules-load.services`.
+fn run() -> Result<()> {
+    migrate(ReplaceListsMigration(vec![ListReplacement {
+        setting: "services.kernel-modules.restart-commands",
+        old_vals: &[],
+        new_vals: &["/usr/bin/systemctl restart systemd-modules-load"],
```
bcressey: `try-restart` is safer since it won't cause services to be queued for starting unless they're already started, which could create issues during boot. It's not specifically a problem here since `systemd-modules-load.service` starts very early, but I'd prefer to stick to the "safer" usage everywhere.

```suggestion
        new_vals: &["/usr/bin/systemctl try-restart systemd-modules-load"],
```

sources/models/shared-defaults/defaults.toml
```diff
@@ -129,13 +129,17 @@ restart-commands = ["/usr/bin/corndog sysctl"]
 affected-services = ["sysctl"]
 
 [services.kernel-modules]
-configuration-files = ["modprobe-conf"]
-restart-commands = []
+configuration-files = ["modprobe-conf", "modules-load"]
+restart-commands = ["/usr/bin/systemctl restart systemd-modules-load"]
```
bcressey:
```suggestion
restart-commands = ["/usr/bin/systemctl try-restart systemd-modules-load"]
```

bcressey (approved):
Quite the crash course in migrations, but they look OK to me.

foersleo:
> > One example would be loading the correct module for your choice of scheduling algorithm for ipvs.
> 
> Do we have other examples? I noted in #3391 that the kernel seems to auto-load the IPVS modules.

I missed the part that IPVS modules are now actually working to be autoloaded. I remember back when I was looking at the original ticket that the overhaul of the kube-proxy code indicated that starting in k8s 1.27 the modules could be loaded automatically. But I did not revisit this until now.

Another instance of people loading modules manually would be `br_netfilter` as raised in https://github.com/bottlerocket-os/bottlerocket/issues/3052 . But that again seems to be just a different facet of the same "kube-proxy needs different modules for different settings".

Even if the autoloading of ipvs modules works with k8s-1.27 plus, this is probably a good change for people still using older versions and those use-cases we do not know about for comparably low cost.

foersleo:
:arrow_up: force-push changed the `restart-command` to use `try-restart` instead of `restart` as suggested by Ben.

sources/api/migration/migrations/v1.16.0/kernel-modules-autoload-restart/src/main.rs
foersleo: systemd activation states and lingo around these is somewhat confusing or overloaded.

The man page says `try-restart` is only doing its work when the service is running (`This does nothing if units are not running.`). The activation state for `systemd-modules-load` is `active (exited)` once it ran (in contrast to `active (running)`).

I confirmed that it is still considered what is called running in the manpage by manually triggering a `systemctl try-reload systemd-modules-load` and checking the time the service last ran. Added the changes accordingly.

bcressey (approved):

markusboehme (approved):
Nice work, especially with the migrations--learned a thing or two! :mage:

stmcginnis (approved):
🎉

PR #3462, opened 2023-09-15
actions: Enable list-variants to filter on changes

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This teaches the list-variants GitHub action to be able to return only a list of variants that are impacted by the changes in the PR.

It switches the list-variants action to be a JavaScript action. This allows us to have more control over what we can do and how we process the pull request information.

By default it will not filter any of the variants and will just return whatever is found under the variants directory. This is useful for the current nightly workflow and if we add anything else that needs to do anything for all variants included.

There is now an input called filter-variants that enables the new behavior. If that is set to true, the action will get the list of variants from the repo, but also try to determine all paths in the repo that make up each variant. With that information, it is able to look at the files being modified as part of the PR and determine which variants are actually impacted by the change. Any variants that do not appear to be modified will be filtered out of the returned list, so that the build workflow will only include affected variants as part of the matrix of tests to run.

**Note**: This will reduce the number of builds run with each PR, so there is a chance something could be missed if the determination of important paths is not quite right. The `nightly` job will still build all variants, so we should still find out quickly if we somehow merge a breaking change.

We can also improve the logic over time if we identify any gaps in this logic. So while this does add a slight risk that we may miss something that the current workflows would catch, it drastically reduces our runner usage and has a low risk of causing issues.

**Testing done:**

I set up a mirror personal repo to be able to make test changes and verify behavior. After merging these changes to that repo, I was able to create different pull requests making different changes to make sure things behaved as expected.

- Changed `README.md` file, verified only `aws-dev` variant was included (bare minimum if nothing appears to be impacted)
- Changed `packages/kernel-6.1/kernel-6.1.spec` and verified all of the recently added variants using this kernel were returned
- Changed `packages/kubernetes-1.26/kubelet.service` and verified all `*-k8s-1.26-*` variants were included
- Changes `sources/bloodhound/build.rs` file and verified all variants were included
- Flipped filter flag to `false` and verified all variants were returned regardless of modified files

**Terms of contribution:**


stmcginnis:
~Tested with PRs on top of this change and they worked as expected, but didn't look at this change itself. I expected only the dev variants to be returned, but all were. Going to put in some temporary extra debug output to understand what is going on.~

Found typo, now working as expected.

.github/actions/list-variants/action.yml
```diff
@@ -8,17 +16,6 @@ outputs:
...
-      run: |
-        cd variants
-        output="variants=$(ls -d */ | cut -d'/' -f 1 | grep -vE '^(shared|target)$' | jq -R -s -c 'split("\n")[:-1]')"
-        echo $output
-        echo $output >> $GITHUB_OUTPUT
-        output="aarch-enemies=$(ls -d */ | cut -d'/' -f 1 | grep -E '(^(metal|vmware)|\-dev$)' | jq -R -s -c 'split("\n")[:-1] | [ .[] | {"variant": ., "arch": "aarch64"}]')"
-        echo $output
-        echo $output >> $GITHUB_OUTPUT
+  using: "node20"
+  main: 'dist/index.js'
```
bcressey, line 20: I'm quite uncomfortable taking this dependency.

stmcginnis: This is the configuration setting for the action run to tell it to use node20. We are not actually importing any code into the project itself.

Is that what you are referring to?

.github/actions/list-variants/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "list-variants",
+  "version": "1.0.0",
+  "private": true,
+  "main": "index.js",
+  "scripts": {
+    "lint": "eslint .",
+    "prepare": "ncc build index.js -o dist --source-map --license licenses.txt"
+  },
+  "dependencies": {
```
stmcginnis, line 10: Just FYI, `toml` is the only dependency we're using that isn't being used by the [other actions](https://github.com/actions/checkout/blob/main/package.json#L30) we have been using,

.github/actions/list-variants/action.yml
arnaldo2792, line 20: Have you tried not compiling the app, and just do what [this example] does?. I would like us to avoid including any "compiled" javascript file since they tend to be gigantic.

[this example]: https://github.com/duckduckgo/privacy-configuration

stmcginnis: Yeah, the alternative is to include the `node_modules` directory, at least from what I understand. That would be much more than the compiled output that gets somewhat minified in `dist/`.

stmcginnis:
No longer pursing, and I believe the plan is to take a different approach.

PR #3468, opened 2023-09-19
restore check shell

### LINKED ISSUE #3464: restore shell-check after twoliter implementation
When we moved the build system to Twoliter, we also moved the `Makefile.toml` `tasks.shell-check` task. This is problematic because the existence of, location of, etc scripts that need to be linted by shellcheck should not be a Twoliter concern.

Restore the shellcheck script to the Bottlerocket repo and do not use Twoliter for checking shell scripts.

webern:
**Issue number:**

Closes #3464 

**Description of changes:**

An inadvertent change with Twoliter is that shellcheck linting disappeared. This feels like something more at the project level than the build system level so I have restored the shellcheck here. I cherry-picked the original commits to retain authorship. 

**Testing done:**

Used it. Messed up a script and used it again to prove it was working.

**Terms of contribution:**

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -342,19 +342,56 @@ run_task = "run-twoliter"
 
 # A top level target for devs to ensure review and patch readiness
 [tasks.check]
+dependencies = ["check-shell"]
 run_task = "run-twoliter"
 
 [tasks.check-fmt]
 run_task = "run-twoliter"
 
 [tasks.check-lints]
+dependencies = ["check-shell"]
 run_task = "run-twoliter"
 
 [tasks.check-clippy]
 run_task = "run-twoliter"
 
 [tasks.check-shell]
-run_task = "run-twoliter"
+script = [
+'''
+rc=0
+
+# For bash first-party shell code
+if ! docker run --rm \
+  --network=none \
+  --user "$(id -u):$(id -g)" \
+  --security-opt label:disable \
+  -v "${BUILDSYS_TOOLS_DIR}":/tmp/tools \
+  "${BUILDSYS_SDK_IMAGE}" \
+  bash -c \
+    'flagged_scripts=0 && \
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):

yeazelm (approved):

zmrow (approved):

Makefile.toml
```diff
@@ -342,19 +342,56 @@ run_task = "run-twoliter"
...
+rc=0
+
+# For bash first-party shell code
+if ! docker run --rm \
+  --network=none \
+  --user "$(id -u):$(id -g)" \
+  --security-opt="label:disable" \
+  -v "${BUILDSYS_TOOLS_DIR}":/tmp/tools \
+  "${BUILDSYS_SDK_IMAGE}" \
+  bash -c \
```
bcressey, line 370: This doesn't feel like the right interface given that `twoliter` is going to need to manage the SDK install and execution. `twoliter` also needs to do its own shell-checking for `rpm2img` etc, though that doesn't need to be delivered as a Makefile.toml task.

What do you think about exposing a `twoliter lint <path>` command that will run all the lints for the target path? Potentially to include Rust, Go, shell, rpm spec lints.

In the interim this could just be done in CI. We might want to consider a `Makefile` or some other build interface for the non-twoliter tasks, since it is very confusing to blend them together in the same `Makefile.toml`.

webern: Should we consider adding that to the Twoliter backlog and putting this back here for now?

bcressey: I'm OK with the Twoliter backlog part, but I'd rather this didn't come back here.

Losing `shellcheck` for a couple of developer-oriented scripts feels like the lesser evil compared to prolonging the existence of this `Makefile.toml` and adding new non-twoliter functionality.

PR #3479, opened 2023-09-20
skip reinstillation of twoliter from source

### LINKED ISSUE #3478: twoliter reinstalls for every command when installing from git
When we are testing changes to Twoliter, we need to install from git. In that case the sha is not going to match the version output from `twoliter --version`. This means that we have to rebuild and install Twoliter each time we call `cargo make`. It takes 10 minutes to compile Twoliter.

Instead, if `TWOLITER_ALLOW_BINARY_INSTALL=false`, we should skip checking the version.

webern:
**Issue number:**

Closes #3478 

**Description of changes:**

When we are testing changes to Twoliter, we need to install from git. In that case the sha is not going to match the version output from Twoliter --version. This means that we have to rebuild and install Twoliter each time we call cargo make. It takes 10 minutes to compile Twoliter.

Fix this such that TWOLITER_ALLOW_BINARY_INSTALL=false skips checking the version when any `twoliter` is installed.

**Testing done:**

Used this locally and it fixed the problem.

**Terms of contribution:**

### DIFF
--- a/tools/install-twoliter.sh
+++ b/tools/install-twoliter.sh
@@ -110,6 +110,10 @@ on_exit "rm -rf ${workdir}"
 
 if [ "${reuse_existing}" = "true" ] ; then
    if [ -x "${dir}/twoliter" ] ; then
+      if [ "${allow_bin}" != "true" ]; then
+        echo "Twoliter binary found and --allow-binary-install is false. Skipping install."
+        exit 0
+      fi
       version_output="$("${dir}/twoliter" --version)"
       found_version=v$(echo $version_output | awk '{print $2}')
       echo "Found twoliter ${found_version} installed."


etungsten (approved):

stmcginnis (approved):

tools/install-twoliter.sh
```diff
@@ -110,6 +110,10 @@ on_exit "rm -rf ${workdir}"
 
 if [ "${reuse_existing}" = "true" ] ; then
    if [ -x "${dir}/twoliter" ] ; then
+      if [ "${allow_bin}" != "true" ]; then
+        echo "Twoliter binary found and --allow-binary-install is false. Skipping install."
+        exit 0
```
bcressey, line 115: I don't agree with this change and would like to see it reverted.

The version check is pretty important as otherwise there's no guarantee that the build will happen with the pinned version rather than whatever happens to be present. This potentially affects the correctness of any build where the binary is not available and the source build must be used, which can happen for other reasons than a developer working locally on twoliter.

It might be better for the twoliter version to not include the Git SHA, or to allow the path to the binary to be overridden (e.g. to `~/twoliter/target/release/twoliter`) and skip the version check if the override is set - with a loud warning that the override is preventing the version check.

bcressey: > In that case the sha is not going to match the version output from Twoliter --version.

```
❯ ./tools/twoliter/twoliter -V
twoliter 0.0.3
```

I don't understand what sha we are or aren't matching since `twoliter -V` doesn't print a SHA.

PR #3482, opened 2023-09-21
readme: clarify host lifecycle behaviour

jhart0:
Doc only update clarifying lifecycle behavior for host containers

### DIFF
--- a/README.md
+++ b/README.md
@@ -1306,6 +1306,7 @@ There are a few important caveats to understand about host containers:
 * They run in a separate instance of containerd than the one used for orchestrated containers like Kubernetes pods.
 * They're not updated automatically. You need to update the `source` and commit those changes.
 * If you set `superpowered` to true, they'll essentially have root access to the host.
+* If the container exits for any reason, Bottlerocket will attempt to restart it after 45 seconds.
 
 Because of these caveats, host containers are only intended for special use cases.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1306,6 +1306,7 @@ There are a few important caveats to understand about host containers:
 * They run in a separate instance of containerd than the one used for orchestrated containers like Kubernetes pods.
 * They're not updated automatically. You need to update the `source` and commit those changes.
 * If you set `superpowered` to true, they'll essentially have root access to the host.
+* If the container exits (whether successfully or not), it will always restart after 45 seconds
```
stmcginnis: Thanks, this is good clarification. A minor suggestion on the wording, but looks good!

```suggestion
* If the container exits for any reason, Bottlerocket will attempt to restart it after 45 seconds.
```

stmcginnis (approved):
Thanks @jhart0! 🎉

jpculp:
Thanks for this! Before we merge, would you be willing to squash the commits under something like "readme: clarify host lifecycle behaviour"? If you would prefer we can also do it on our end.

jhart0:
> would you be willing to squash the commits under something like "readme: clarify host lifecycle behaviour"

done, thanks!

jpculp (approved):

PR #3485, opened 2023-09-21
Actions: prebuild common packages to cache for full builds

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This builds some of the common packages in a step before the normal full build. The build output from these package builds will be captured in the cache, so later on when there are multiple variants being built that all use the same packages, it should be able to get the artifacts from the cache and perform an incremental build rather than each building the same package repeatedly.

**Testing done:**

Observing CI run to make sure it is behaving as expected.

**Terms of contribution:**

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -18,7 +18,7 @@ runs:
       with:
         path: |
           ~/.cargo
-        key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}
+        key: ${{ hashFiles('.github/**/*.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}
     - uses: actions/cache@v3
       # Cache first-party code dependencies
       with:
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -6,11 +6,25 @@ inputs:
     required: false
     default: false
     type: boolean
+  read-only-cache:
+    description: "Whether to update the cache (false) or not (true)"
+    required: false
+    default: false
+    type: boolean
+  cache-arch:
+    description: "The platform architecture for this cache"
+    required: false
+    default: ""
 runs:
   using: "composite"
   steps:
     - run: |
         echo "OS_ARCH=`uname -m`" >> $GITHUB_ENV
+        if [[ "${{ inputs.cache-arch }}" == "" ]]; then
+          echo "CACHE_ARCH=`uname -m`" >> $GITHUB_ENV
+        else
+          echo "CACHE_ARCH=${{ inputs.cache-arch }}" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -37,12 +37,39 @@ jobs:
...
+      # TODO: Make some of this dynamic like we do with discovering variants
+      - run: |
+          cargo make -e BUILDSYS_ARCH=${{ matrix.arch }} -e PACKAGE=kernel-5_10 build-package
+        shell: bash
+      - run: |
+          cargo make -e BUILDSYS_ARCH=${{ matrix.arch }} -e PACKAGE=kernel-5_15 build-package
+        shell: bash
+      - run: |
+          cargo make -e BUILDSYS_ARCH=${{ matrix.arch }} -e PACKAGE=kernel-6_1 build-package
+        shell: bash
```
bcressey, line 67: These are variant-specific and need to be rebuilt for vmware and metal, currently.

bcressey:
How much cache space do we have available and how much does this use?

.github/actions/setup-node/action.yml
```diff
@@ -18,13 +23,43 @@ runs:
...
+        path: |
+          .cargo
+        key: cargo-${{ hashFiles('.github/**/*.yml') }}${{ hashFiles('sources/Cargo.lock') }}
+      if: ${{ inputs.read-only-cache == 'true' }}
+    - uses: actions/cache/restore@v3
+      # Cache build artifacts in 'variants/target/' to reduce rebuilds
+      with:
+        path: |
+          variants/target
+        key: target-${{ hashFiles('.github/**/*.yml') }}${{ hashFiles('packages/**/Cargo.toml', 'packages/**/*.spec') }}
```
bcressey: If we retain the `cargo` metadata under `variants/target`, but lose the RPMs under `build/rpms` because they're evicted from the cache for size reasons, then the builds will break.

stmcginnis:
> How much cache space do we have available and how much does this use?

From the docs:
> There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB.

Based on the last run, this added ~5MB compressed to the cache. But it doesn't appear to be working right. The subsequent builds were not any faster, and just based on the timestamps, it doesn't look like it is picking up the artifacts right so it is not speeding up the build at all.

.github/actions/setup-node/action.yml
stmcginnis: Cache eviction works by evicting the older cache items first, so it would need to be a pretty severe cache issue to lost part of the same build artifacts.

webern (commented):
This makes me a bit nervous on principle. I think of CI builds being "from clean" (as much as possible).

stmcginnis:
This may be an interesting avenue to explore more, but this current attempt does not appear to bring any benefit. Closing out since I don't think we want to pursue this particular approach further.

PR #3487, opened 2023-09-25
actions: Use latest cargo-make

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This updates the version of cargo-make used in the GitHub Action runs. This is only part of the workflow since things are delegated off to twoliter and the SDK, but we should keep the initial `cargo make` calls updated.

**Testing done:**

Local usage.

**Terms of contribution:**

### DIFF
--- a/.github/actions/setup-node/action.yml
+++ b/.github/actions/setup-node/action.yml
@@ -25,8 +25,8 @@ runs:
         path: |
           .cargo
         key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}
-    - run: cargo install --locked --version 0.36.0 cargo-make
+    - run: cargo install cargo-make
       shell: bash
     - if: ${{ inputs.perform-cache-cleanup }}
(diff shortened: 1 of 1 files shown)


stmcginnis:
Cancelled workflow to save some run time. The jobs got past the part that would catch any error, so no reason to run all the builds.

.github/actions/setup-node/action.yml
```diff
@@ -25,7 +25,7 @@ runs:
         path: |
           .cargo
         key: ${{ hashFiles('.github/workflows/cache.yml') }}-${{ runner.os }}-${{ env.OS_ARCH }}-${{ hashFiles('sources/Cargo.lock') }}
-    - run: cargo install --locked --version 0.36.0 cargo-make
+    - run: cargo install --locked --version 0.37.1 cargo-make
```
jpculp:
```suggestion
    - run: cargo install --locked cargo-make
```
I wonder if it's about time we unpinned `cargo-make` and others. It is annoying on the rare occurrence that latest breaks something, but is it as annoying or as frequent as bumping versions?

jpculp (approved):

etungsten (approved):

webern (approved):

PR #3490, opened 2023-09-27
POC: Filter CI build variants based on actual changes (Rust Edition)

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

This uses a combination of the get-changed-files action and bash
scripting to determine what variants need testing based on the changes.
This would require updates to buildsys or another tool that could tell
us what paths in the repo are relevant to each variant to be able to
make the decision if it needs to be tested.

**Testing done:**

Testing with personal repo and now with main repo.

**Terms of contribution:**


stmcginnis:
Cancelled workflow because we just needed the initial "Determine variants" job to run to verify it picks up the right variants.

Next step would be to determine where we want to add the capability to get the set of paths that feed into each variant. Buildsys might be the best place for this since it already does some variant metadata parsing, but it could also be another tool specifically for this purpose. Will continue to explore that.

bcressey:
This evolved into https://github.com/bottlerocket-os/twoliter/pull/107, which is more along the lines of what we'd take forward. Closing this.

PR #3491, opened 2023-09-28
apiclient: update tungstenite to v0.20.1

arnaldo2792:
**Description of changes:**

Newer versions of `tungstenite` implemented the v13 version of the Web Sockets protocol (described in [rfc6455]). In this version, new headers are required in the upgrade request. With this update, now the upgrade request sent by `apiclient` include the missing headers.

In the same version, the 'frame' message was added. Neither `apiserver` nor `apiclient` send this type of message, so they are ignored by `apiclient`.

I found the required headers [here]. For the values I followed the RFC:

> The request MUST contain a |Host| header field whose value contains /host/ plus optionally ":" followed by /port/ (when not using the default port).

> The request MUST contain an |Upgrade| header field whose value MUST include the "websocket" keyword

> The request MUST contain a |Connection| header field whose value MUST include the "Upgrade" token.

>The request MUST include a header field with the name |Sec-WebSocket-Key|.  The value of this header field MUST be a nonce consisting of a randomly selected 16-byte value that has been base64-encoded (see Section 4 of [RFC4648]).  The nonce MUST be selected randomly for each connection.

> The request MUST include a header field with the name |Sec-WebSocket-Version|.  The value of this header field MUST be 13.

[rfc6455]: https://www.rfc-editor.org/rfc/rfc6455
[here]: https://github.com/snapview/tungstenite-rs/blob/8b3ecd3cc0008145ab4bc8d0657c39d09db8c7e2/src/handshake/client.rs#L115 

**Testing done:**
I built the `aws-ecs-1` and `aws-ecs-2` variants, and confirmed I could exec commands through the admin container:

```bash
[ssm-user@control]$ apiclient get os
{
  "os": {
    "arch": "x86_64",
    "build_id": "14d6c7d1",
    "pretty_name": "Bottlerocket OS 1.16.0 (aws-ecs-2)",
    "variant_id": "aws-ecs-2",
    "version_id": "1.16.0"
  }
}
[ssm-user@control]$ apiclient exec admin sheltie cat /etc/os-release
NAME=Bottlerocket
ID=bottlerocket
VERSION="1.16.0 (aws-ecs-2)"
PRETTY_NAME="Bottlerocket OS 1.16.0 (aws-ecs-2)"
VARIANT_ID=aws-ecs-2
VERSION_ID=1.16.0
BUILD_ID=14d6c7d1
HOME_URL="https://github.com/bottlerocket-os/bottlerocket"
SUPPORT_URL="https://github.com/bottlerocket-os/bottlerocket/discussions"
BUG_REPORT_URL="https://github.com/bottlerocket-os/bottlerocket/issues"
DOCUMENTATION_URL="https://bottlerocket.dev"
```

I also confirmed that I can enter the admin container:

```bash
[ssm-user@control]$ enter-admin-container
Confirming admin container is enabled...
Waiting for admin container to start...
Entering admin container
          Welcome to Bottlerocket's admin container!
    ╱╲
   ╱┄┄╲   This container provides access to the Bottlerocket host
   │▗▖│   filesystems (see /.bottlerocket/rootfs) and contains common
  ╱│  │╲  tools for inspection and troubleshooting.  It is based on
  │╰╮╭╯│  Amazon Linux 2, and most things are in the same places you
    ╹╹    would find them on an AL2 host.

To permit more intrusive troubleshooting, including actions that mutate the
running state of the Bottlerocket host, we provide a tool called "sheltie"
(`sudo sheltie`).  When run, this tool drops you into a root shell in the
Bottlerocket host's root filesystem.
[root@admin]#

```

**Terms of contribution:**

### DIFF
--- a/sources/api/apiclient/Cargo.toml
+++ b/sources/api/apiclient/Cargo.toml
@@ -10,6 +10,7 @@ build = "build.rs"
 exclude = ["README.md"]
 
 [dependencies]
+base64 = "0.21"
 constants = { path = "../../constants", version = "0.1" }
 datastore = { path = "../datastore", version = "0.1" }
 futures = { version = "0.3", default-features = false }
@@ -31,7 +32,7 @@ signal-hook = "0.3"
 simplelog = "0.12"
 snafu = { version = "0.7", features = ["futures"] }
 tokio = { version = "~1.25", default-features = false, features = ["fs", "io-std", "io-util", "macros", "rt-multi-thread", "time"] }  # LTS
-tokio-tungstenite = { version = "0.16", default-features = false, features = ["connect"] }
+tokio-tungstenite = { version = "0.20", default-features = false, features = ["connect"] }
 toml = "0.5"
 unindent = "0.1"
 url = "2"
--- a/sources/api/apiclient/src/exec/connect.rs
+++ b/sources/api/apiclient/src/exec/connect.rs
@@ -1,9 +1,13 @@
 //! The 'connect' module provides a function for connecting to a WebSocket over a Unix-domain
 //! socket, which is a bit more finicky than normal.
 
+use base64::{engine, Engine as _};
+use http::header::{CONNECTION, HOST, SEC_WEBSOCKET_KEY, SEC_WEBSOCKET_VERSION, UPGRADE};
+use httparse::Header;
 use hyper::service::Service;
 use hyper_unix_connector::{UnixClient, Uri, UDS};
 use log::debug;
+use rand::{thread_rng, Rng};
 use snafu::{ensure, ResultExt};
 use std::path::Path;
 use tokio_tungstenite::{client_async, tungstenite::http::StatusCode, WebSocketStream};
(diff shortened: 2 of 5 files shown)


sources/deny.toml
```diff
@@ -78,7 +76,10 @@ skip-tree = [
     # serde_yaml actually uses a newer hashbrown dependency,
     # but several crates in the graph pull in an older one, so skip for now
     # until they can be updated
-    { name = "hashbrown", version = "=0.13.2" }
+    { name = "hashbrown", version = "=0.13.2" },
+    # tungstenite uses a newer version of base64, but must of the
+    # first party crates are still in this version
+    { name = "base64", version = "=0.13.1" }
```
jpculp: After the next `tough` release, we can switch everything to `0.21`.

sources/api/apiclient/Cargo.toml
```diff
@@ -31,7 +32,7 @@ signal-hook = "0.3"
 simplelog = "0.12"
 snafu = { version = "0.7", features = ["futures"] }
 tokio = { version = "~1.25", default-features = false, features = ["fs", "io-std", "io-util", "macros", "rt-multi-thread", "time"] }  # LTS
-tokio-tungstenite = { version = "0.16", default-features = false, features = ["connect"] }
+tokio-tungstenite = { version = "0.20.1", default-features = false, features = ["connect"] }
```
jpculp:
```suggestion
tokio-tungstenite = { version = "0.20", default-features = false, features = ["connect"] }
```

webern (commented):
I think in the past we had a case where apiclient processes became zombies and maybe tungstenite was to blame? Anyone remember? Should we check for zombies?

sources/api/apiclient/src/exec/connect.rs
```diff
@@ -1,9 +1,13 @@
 //! The 'connect' module provides a function for connecting to a WebSocket over a Unix-domain
 //! socket, which is a bit more finicky than normal.
 
+use base64::{engine, Engine as _};
```
bcressey, line 4: Do we need `Engine` in scope for a trait or something? Not sure I follow what this does.

bcressey (approved):

bcressey:
Zombies showed up in #1384 and #1507 but I think that was before `apiclient exec`.

sources/api/apiclient/src/exec/connect.rs
arnaldo2792: Yes, we need `Engine` in scope otherwise we can't call the methods to generate the base64 strings.

sources/deny.toml
```diff
@@ -78,7 +76,10 @@ skip-tree = [
     # serde_yaml actually uses a newer hashbrown dependency,
     # but several crates in the graph pull in an older one, so skip for now
     # until they can be updated
-    { name = "hashbrown", version = "=0.13.2" }
+    { name = "hashbrown", version = "=0.13.2" },
+    # tungstenite uses a newer version of base64, but must of the
```
yeazelm:
```suggestion
    # tungstenite uses a newer version of base64, but many of the
```

yeazelm (approved):

arnaldo2792:
(Forced push fixes the comments in the last revision)

PR #3520, opened 2023-10-11
Backport systemd patches and configure netlink timeout

zmrow:
**Issue number:**
N/A

**Description of changes:**
This PR backports a [few patches](https://github.com/systemd/systemd/pull/29390/commits) from a [recent `systemd` issue](https://github.com/systemd/systemd/issues/25441).  Further details are in the issue; the tl;dr is that on a heavily loaded system (such as one running many containers/pods), it is possible for the kernel to not respond to `RTM_NEWROUTE` requests within the current default 25 second timeout.  This results in network links becoming unusable.  Bottlerocket now sets this timeout to `infinity` since we require all links to be usable, especially the main link, even if the system is heavily loaded.

```
    systemd: Backport patches that allow netlink timeout to be configured

    Backport 2 patches that allow the configurability of the default timeout
    for waiting for replies for netlink messages from the kernel.  The
    default timeout is 25 seconds and it is possible to hit this timeout if
    the system is under load.  If the timeout is hit, network links can
    become unusable without intervention.

    Reference issue: github.com/systemd/systemd/issues/25441
```
```
    release: Configure default SYSTEMD_NETLINK_DEFAULT_TIMEOUT

    Write a drop-in for the systemd-networkd service that sets the
    SYSTEMD_NETLINK_DEFAULT_TIMEOUT environment variable to "infinity",
    rather than use the default timeout of 25 seconds.  This ensures that
    systemd-networkd won't put links into an inoperable state if the
    kernel/system is heavily loaded and doesn't respond right away.
```

**Testing done:**
`systemd-networkd`'s environment shows the appropriate variable after this change:
```
bash-5.1# systemctl status systemd-networkd
...
   Main PID: 2767 (systemd-network)
...

bash-5.1# cat /proc/2767/environ                                                                                                
...SYSTEMD_NETLINK_DEFAULT_TIMEOUT=infinity
```

Also set a super short timeout of `1ms` and watched the network fail to come up:
```
bash-5.1# journalctl -u systemd-networkd
...
Oct 11 22:29:13 localhost systemd-networkd[1004]: Enumeration completed                                                                        
Oct 11 22:29:13 localhost systemd-networkd[1004]: eth0: Failed to wait for the interface to be initialized: Connection timed out
Oct 11 22:29:13 localhost systemd-networkd[1004]: eth0: Failed                                                                                 
```

**Terms of contribution:**

### DIFF
--- a/packages/systemd/systemd.spec
+++ b/packages/systemd/systemd.spec
@@ -14,6 +14,15 @@ Source3: journald.conf
 Source4: issue
 Source5: systemd-journald.conf
 
+# Backport of upstream patches that make the netlink default timeout
+# configurable.  Bottlerocket carries this patch and configures the timeout in
+# an effort to avoid a situation where a network link becomes unusable if the
+# system is under load and doesn't process the RTM_NEWROUTE acknowledgement
+# within the default timeout of 25 seconds.
+# Reference issue: github.com/systemd/systemd/issues/25441
+Patch1001: 1001-sd-netlink-make-calc_elapse-return-USEC_INFINITY-whe.patch
+Patch1002: 1002-sd-netlink-make-the-default-timeout-configurable-by-.patch
+
 # Local patch to work around the fact that /var is a bind mount from
 # /local/var, and we want the /local/var/run symlink to point to /run.
 Patch9001: 9001-use-absolute-path-for-var-run-symlink.patch
--- a/packages/release/release.spec
+++ b/packages/release/release.spec
@@ -75,6 +75,7 @@ Source1080: runtime.slice
 # Drop-in units to override defaults
 Source1100: systemd-tmpfiles-setup-service-debug.conf
 Source1101: systemd-resolved-service-env.conf
+Source1102: systemd-networkd-service-env.conf
 
 # systemd-udevd default link
 Source1200: 80-release.link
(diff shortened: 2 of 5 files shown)


packages/systemd/9015-sd-netlink-make-the-default-timeout-configurable-by-.patch
```diff
@@ -0,0 +1,73 @@
...
+@@ -108,7 +108,11 @@ All tools:
+   for example in `systemd-nspawn`, will be logged to the audit log, if the
+   kernel supports this.
+ 
+-`systemctl`:
++* `$SYSTEMD_ENABLE_LOG_CONTEXT` — if set, extra fields will always be logged to
++  the journal instead of only when logging in debug mode.
++
++* `$SYSTEMD_NETLINK_DEFAULT_TIMEOUT` — specifies the default timeout of waiting
++  replies for netlink messages from the kernel. Defaults to 25 seconds.
```
bcressey: This part isn't quite right:
* The `$SYSTEMD_ENABLE_LOG_CONTEXT` variable support didn't get backported so you can just omit that
* The `systemctl:` section header should not go away and should continue to start just before the list with `$SYSTEMCTL_FORCE_BUS=1`

packages/systemd/systemd.spec
```diff
@@ -61,6 +61,15 @@ Patch9012: 9012-core-mount-increase-mount-rate-limit-burst-to-25.patch
...
 Patch9013: 9013-sd-dhcp-lease-parse-multiple-domains-in-option-15.patch
 
+# Backport of upstream patches that make the netlink default timeout
+# configurable.  Bottlerocket carries this patch and configures the timeout in
+# an effort to avoid a situation where a network link becomes unusable if the
+# system is under load and doesn't process the RTM_NEWROUTE acknowledgement
+# within the default timeout of 25 seconds.
+# Reference issue: github.com/systemd/systemd/issues/25441
+Patch9014: 9014-sd-netlink-make-calc_elapse-return-USEC_INFINITY-whe.patch
+Patch9015: 9015-sd-netlink-make-the-default-timeout-configurable-by-.patch
```
bcressey: These should be numbered differently - say 1001 and 1002 - since we expect to be able to drop them after a future rebase.

Local patches like the 9000 series should always be applied after any upstream patches since we want to rebase local changes immediately and not have to adjust backported patches to work around our changes.

zmrow:
^ Addresses @bcressey 's feedback

yeazelm (approved):

bcressey (approved):

PR #3522, opened 2023-10-12
ootb: Copy licenses via rpm

### LINKED ISSUE #3177: kits: use an rpm package for bottlerocket licenses
### Description

The script, `rpm2img` assumes that we are running in the Bottlerocket monorepo.
During the script execution it copies `LICENSE-MIT` directly from the git repo.
This is problematic because an out-of-tree build may not have these files.

Instead, we should package licenses into an RPM file during the package build step and install them from the RPM package during rpm2img.

In other words, we can no longer do this:
https://github.com/bottlerocket-os/bottlerocket/blob/08497fd1503397a0e02858be10917c34965b7ca8/tools/rpm2img#L233

### Definition of Done

When the build no longer relies on Bottlerocket's license files being being present during rpm2img.
```
id: br_licenses
epic: twoliter_exec
```

Related to https://github.com/bottlerocket-os/bottlerocket/issues/2669

ecpullen:
**Issue number:**

Closes #3177 

**Description of changes:**

Instead of copying Bottlerocket licenses (COPYRIGHT, LICENSE-APACHE, LICENSE-MIT) as part of rpm2img, the licenses will instead be added by a package.

**Testing done:**

Added an additional license - `LICENSE-ETHAN`
Built a custom variant of bottlerocket to see - 
```
# ls /usr/share/licenses/
COPYRIGHT        grep         libinih                 libxcrypt
LICENSE-APACHE   grub         liblzma                 libz
LICENSE-ETHAN    host-ctr     libmnl                  libzstd
LICENSE-MIT      iproute      libmount                makedumpfile
acpid            iptables     libncurses              microcode
bash             iputils      libnetfilter_conntrack  oci-add-hooks
ca-certificates  kernel-6.1   libnetfilter_cthelper   os
chrony           kexec-tools  libnetfilter_cttimeout  pigz
conntrack-tools  keyutils     libnetfilter_queue      policycoreutils
containerd       kmod         libnfnetlink            procps
coreutils        libacl       libnftnl                readline
dbus-broker      libattr      libpcre                 runc
docker-cli       libaudit     libseccomp              shim
docker-engine    libblkid     libselinux              strace
docker-init      libbzip2     libsemanage             systemd
docker-proxy     libcap       libsepol                util-linux
e2fsprogs        libelf       libsmartcols            xfsprogs
ethtool          libexpat     libstd-rust
findutils        libfdisk     liburcu
glibc            libgcc       libuuid
```

The additional license is also copied.

I have also built bottlerocket with this package and the existing `rpm2img` implementation to ensure that the build succeeds (even though the current rpm2img implementation will overwrite the 3 license files)

**Terms of contribution:**

### DIFF
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -68,6 +68,11 @@ Source300: ephemeral-storage.rules
 Source301: ebs-volumes.rules
 Source302: supplemental-storage.rules
 
+# 4xx sources: Bottlerocket licenses
+Source400: COPYRIGHT
+Source401: LICENSE-MIT
+Source402: LICENSE-APACHE
+
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{_cross_os}apiclient
 Requires: %{_cross_os}apiserver
@@ -520,8 +525,15 @@ install -p -m 0644 %{S:121} %{buildroot}%{_cross_unitdir}
 %cross_scan_attribution --clarify %{_builddir}/sources/clarify.toml \
     cargo --offline --locked %{_builddir}/sources/Cargo.toml
 
+# Install licenses
+install -d %{buildroot}%{_cross_licensedir}
+install -p -m 0644 %{S:400} %{S:401} %{S:402} %{buildroot}%{_cross_licensedir}
+
 %files
 %{_cross_attribution_vendor_dir}
+%{_cross_licensedir}/COPYRIGHT
+%{_cross_licensedir}/LICENSE-MIT
+%{_cross_licensedir}/LICENSE-APACHE
 
 %files -n %{_cross_os}apiserver
 %{_cross_bindir}/apiserver
--- a/packages/os/COPYRIGHT
+++ b/packages/os/COPYRIGHT
@@ -1,0 +1,19 @@
+Copyright Amazon.com, Inc., its affiliates, or other contributors. All Rights Reserved.
+
+Except as otherwise noted (below and/or in individual files), Bottlerocket is dual-licensed under
+the Apache License, version 2.0 <LICENSE-APACHE> or the MIT license <LICENSE-MIT>, at your option.
+
+Copyrights in Bottlerocket are retained by their contributors. No copyright assignment is required
+to contribute to Bottlerocket. Contributions to Bottlerocket are explicitly made under both the
+Apache License, version 2.0, and the MIT license. For full authorship information, see the version
+control history.
+
+Bottlerocket operating system images include packages written by third parties, which may carry
+their own copyright notices and license terms. These are available in /usr/share/licenses on the
+operating system images.
+
+=^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
+
+Contains modified hyper-proxy files [mod.rs, stream.rs, tunnel.rs] from
+https://github.com/tafia/hyper-proxy 2021-09-20.
+Copyright (c) 2017 Johann Tuffe. Licensed under the MIT License.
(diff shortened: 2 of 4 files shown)


packages/bottlerocket-licenses/bottlerocket-licenses.spec
```diff
@@ -0,0 +1,32 @@
...
+%description
+%{summary}.
+
+%prep
+%setup -T -c
+
+%build
+
+%install
+mkdir -p %{buildroot}/usr/share/bottlerocket/
```
cbgbt: Looks like `bottlerocket` may be a mistake here since we're actually putting licenses in `/usr/share/licenses`?

In any case, this directory isn't quite right, `/usr/share/licenses` is mounted by a [systemd unit set up in `release`](https://github.com/bottlerocket-os/bottlerocket/blob/6a66bb2170c56a08dca4ac8ed279c78cc18017fa/packages/release/release.spec#L184)

```suggestion
mkdir -p %{buildroot}/%{_cross_licensedir}
```

You can generally just find/replace `/usr/share/licenses` with `%{_cross_licensedir}`.

cbgbt (commented):
I suspect we want to add this package as a dependency to `release`.

packages/bottlerocket-licenses/bottlerocket-licenses.spec
```diff
@@ -0,0 +1,32 @@
...
+Name: %{_cross_os}bottlerocket-licenses
+Version: 0.0
+Release: 0%{?dist}
+Summary: bottlerocket-licenses
+License: Apache-2.0 OR MIT
+URL: https://github.com/bottlerocket-os/bottlerocket
+
+Source100: COPYRIGHT
+Source101: LICENSE-MIT
+Source102: LICENSE-APACHE
```
bcressey: I would just put these files in the same directory as `os.spec` and include them in that package.

packages/os/os.spec
```diff
@@ -299,6 +304,11 @@ Summary: XFS progs cli
 %description -n %{_cross_os}xfscli
 %{summary}.
 
+%package -n %{_cross_os}licenses
```
cbgbt: You would need to add a `Requires` on `%{_cross_os}licenses` to have this package actually be installed (in the same way that we require e.g. `%{_cross_os}xfscli`)

I see now that we handle attributions in the top-level packaging below, I think it makes sense to add our licenses there rather than create a subpackage.

ecpullen:
Rebuilt the ami with an additional license file and verified that it was added to licenses.
```
# ls /usr/share/licenses/
COPYRIGHT        grep         libinih                 libxcrypt
LICENSE-APACHE   grub         liblzma                 libz
LICENSE-ETHAN    host-ctr     libmnl                  libzstd
LICENSE-MIT      iproute      libmount                makedumpfile
acpid            iptables     libncurses              microcode
bash             iputils      libnetfilter_conntrack  oci-add-hooks
ca-certificates  kernel-6.1   libnetfilter_cthelper   os
chrony           kexec-tools  libnetfilter_cttimeout  pigz
conntrack-tools  keyutils     libnetfilter_queue      policycoreutils
containerd       kmod         libnfnetlink            procps
coreutils        libacl       libnftnl                readline
dbus-broker      libattr      libpcre                 runc
docker-cli       libaudit     libseccomp              shim
docker-engine    libblkid     libselinux              strace
docker-init      libbzip2     libsemanage             systemd
docker-proxy     libcap       libsepol                util-linux
e2fsprogs        libelf       libsmartcols            xfsprogs
ethtool          libexpat     libstd-rust
findutils        libfdisk     liburcu
glibc            libgcc       libuuid
```

cbgbt (approved):

bcressey (approved):

stmcginnis (approved):

PR #3529, opened 2023-10-16
k8s: update eni-max-pods mapping for latest

### LINKED ISSUE #3474: v1.16.0 update eni-max-pods mapping file

stmcginnis:
**Issue number:**

Closes #3474

**Description of changes:**

This pulls in the new instance types to the eni-max-pods mapping file. It adds the max pod values for a new set of published instance types.

Generated content by using the [update script](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/gen_vpc_ip_limits.go).

**Testing done:**

Build image, verified file able to be included.

**Terms of contribution:**


etungsten (approved):

packages/os/eni-max-pods
```diff
@@ -211,8 +232,6 @@ dl1.24xlarge 737
 f1.16xlarge 394
 f1.2xlarge 58
 f1.4xlarge 234
-g2.2xlarge 58
-g2.8xlarge 234
```
zmrow, line 215: Are these instance types not available any more?

stmcginnis: Yeah, it looks like these must have been retired. The script that generates these files queries EC2 directly for information, so it's pretty safe.

zmrow (approved):

PR #3533, opened 2023-10-17
systemd: Update to latest stable patch version 252.18

zmrow:
**Issue number:**
N/A

**Description of changes:**
Updates systemd to the latest stable patch version in the 252 chain: `252.18`.

**Testing done:**
WIP

- [x] Ensure the build log contains no relevant warnings or warnings about patch offsets
- [x] Test spinning up large amounts (3k+) of pods in an ipv6-enabled cluster with systemd-networkd via `aws-k8s-1.28` to regression test the issue fixed in #3520 (instance loses ipv6 address because of timeout).  **Used 80 nodes and 5000 pods.**
- [x] Test launches at scale (2000 nodes) for multiple instance types and network backends (wicked / systemd-networkd), ensure nodes are schedule-able and run pods
  - [x] m5.large -> systemd-networkd
  - [x] m5.large -> wicked
  - [x] t3.medium -> systemd-networkd
  - [x] t3.medium -> wicked
  - [x] m6a.12xlarge -> systemd-networkd (could only get 800 of this instance type in my region)
  - [x] m6a.12xlarge -> wicked (same here)

**Terms of contribution:**

### DIFF
--- a/packages/systemd/Cargo.toml
+++ b/packages/systemd/Cargo.toml
@@ -13,8 +13,8 @@ releases-url = "https://github.com/systemd/systemd-stable/releases"
 package-features = ["unified-cgroup-hierarchy"]
 
 [[package.metadata.build-package.external-files]]
-url = "https://github.com/systemd/systemd-stable/archive/v252.13/systemd-stable-252.13.tar.gz"
-sha512 = "db96a49a8819abbd68606c4063b2f8ef56d2fa07a62733e82c31de12f8e6d33f40bc85d162edc56bd77b23904a017b3b7f7050d281613b78ac39d25c0f1f70ad"
+url = "https://github.com/systemd/systemd-stable/archive/v252.18/systemd-stable-252.18.tar.gz"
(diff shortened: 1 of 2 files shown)


packages/systemd/Cargo.toml
```diff
@@ -26,3 +26,4 @@ libseccomp = { path = "../libseccomp" }
 libselinux = { path = "../libselinux" }
 libxcrypt = { path = "../libxcrypt" }
 util-linux = { path = "../util-linux" }
+
```
bcressey: nit: I don't think we want to add a newline to the end of the file. Can you check your editor settings?

bcressey (commented):
Can you check the build output for the systemd package and make sure there are no warnings about patch offsets or warnings during the build?

packages/systemd/Cargo.toml
zmrow: :/  That's weird - will fix

zmrow:
^ Fixed the added newline in the `Cargo.toml`

zmrow:
>Can you check the build output for the systemd package and make sure there are no warnings about patch offsets or warnings during the build?

Checked the build log and didn't find any relevent warnings or references to patch offsets!  Also update the description

bcressey (approved):

stmcginnis (approved):

Issue #3514, opened 2023-10-10
logdog doesn't capture network configuration for networkd


bcressey:
**Image I'm using:**
aws-k8s-1.28 1.15.0

**What I expected to happen:**
`logdog` should capture network information.

**What actually happened:**
`logdog` creates a `logdog.errors` file with this output:
```
Error running command 'exec wicked wicked show all': 'Error starting command 'exec wicked wicked show all': No such file or directory (os error 2)'
```

**How to reproduce the problem:**
Run `logdog`.

PR #3550, opened 2023-10-26
logdog: Dynamically load configuration files

### LINKED ISSUE #87: logdog and models directories are expected by the build
## Description

Specific directories are mounted in the variant-building Dockerfile. R related to #58. We need to eliminate these mounts from the `Dockerfile`.

Here's the problem:

https://github.com/bottlerocket-os/twoliter/blob/fcdf2c9977f1eba29cf78db6e9edd461b43cddcc/twoliter/embedded/Dockerfile#L146..L147

## Definition of Done

When the Dockerfile no-longer expects Bottlerocket-project-specific directories to exist.

ecpullen:
**Issue number:**

Partially Closes https://github.com/bottlerocket-os/twoliter/issues/87

**Description of changes:**

Walks the directory `/usr/share/logdog.d` to dynamically load all logdog configuration files.

Note: We need to keep `logdog/current/conf` until the necessary changes are made to twoliter.

**Testing done:**

Built aws-dev and verified that `logdog.conf.common` was present. Called `logdog` to ensure the command still works. Created a custom variant with an additional log config. Verified that both log conf files commands were executed.

**Terms of contribution:**


packages/kubernetes-1.23/kubernetes-1.23.spec
```diff
@@ -39,6 +39,7 @@ Source11: kubelet-server-crt
 Source12: kubelet-server-key
 Source13: etc-kubernetes-pki-private.mount
 Source14: credential-provider-config-yaml
+Source15: logdog.kubernetes-1.23.conf
```
bcressey: Let's keep the file name the same across packages. That way it's easier to compare the directories via `git diff --no-index`:
```suggestion
Source15: logdog.kubelet.conf
```

packages/release/logdog.common.conf
```diff
@@ -17,4 +17,4 @@ exec signpost signpost status
 exec wicked wicked show all
 file os-release /etc/os-release
 glob /var/log/kdump/*
-settings settings.json
+settings settings.json
```
bcressey: Can we move the `exec wicked` line to the wicked package? Then we could add a separate `logdog.conf` drop-in to the networkd package and fix #3514.

bcressey: Also, your editor is not saving a single newline at the end of the file - can you update your config?

sources/logdog/src/log_request.rs
```diff
@@ -44,12 +42,22 @@ const SENSITIVE_SETTINGS_PATTERNS: &[&str] = &[
...
-        .lines()
-        .chain(VARIANT_REQUESTS.lines())
-        .filter(|&command| !command.is_empty() && !command.trim_start().starts_with('#'))
-        .collect()
+pub(crate) fn log_requests() -> Result<Vec<String>> {
+    Ok(WalkDir::new(REQUESTS_DIR)
+        .into_iter()
+        .collect::<std::result::Result<Vec<_>, _>>()?
+        .into_iter()
+        .filter(|entry| entry.path().is_file())
```
bcressey: Typically for these `.conf.d` directories, the snippets are added in order, e.g.:
```
00-release.conf
10-networkd.conf
20-k8s.conf
```

That doesn't matter a whole lot here, since entries can't really conflict or overwrite each other, but it might still be useful so that all the commands are run in a predictable order. And it 

Also, I would set min and max depth to 1 since we don't expect to recurse into nested dirs.

webern (approved):
Nice solution. When we initially created logdog we wanted its behavior to be static, but that makes no sense now with ootb.

bcressey (approved):

zmrow (approved):

PR #3552, opened 2023-10-28
bloodhound: Handle chronyd is-active failures

### LINKED ISSUE #3551: Bottlerocket CIS benchmark reporting may report manual checking is required if time sync is disabled

The `apiclient report cis` level 1 benchmark checks to make sure chronyd is configured and active to ensure time sync is happening. To verify the service status of chronyd, the check runs `systemctl is-active chronyd`. If everything is good, this command returns success and the output is "active".

**What actually happened:**

If this service gets disabled, the `systemctl is-active` command actually has a return code of 3, with the output of "inactive". There is a problem in the current check logic where it still expects the command to return success, with just the stdout string indicating it is not active.

This causes the benchmark to report the check as `SKIP`'d which causes the user to need to manually check on the service status.

This is sort of mostly harmless, since the user would need to check on the status anyway to remediate the situation and restart the chronyd service. But the report should be able to properly detect that the service is not active and report it as a `FAIL` state to be accurate.

To be very clear, this service is enabled by default and there is not an exposed user data setting that could lead to this state. It requires manual intervention to cause a failure condition.

To reproduce the failure, someone would need to launch and instance and connect to the control container. From there, they would need to `enter-admin-container` to get to the admin container and `sheltie` to get system access. Once in `sheltie`, they would then need to run `systemctl stop chronyd` to stop the service manually.

stmcginnis:
**Issue number:**

Closes #3551 

**Description of changes:**

Bottlerocket CIS Benchmark control 2.1.1.1 verifies time sync is enabled. For our check for chronyd service status, we run `systemctl is-active chronyd` to verify its active state.

The `systemctl is-active` command doesn't just report the state string though. If the service is not active, the return code from running this command is a 3 and the command execution from rust therefore does not report success.

This changes the way the check is performed by calling `systemctl show --property ActiveState chronyd`. This command will just report the state, and its exit status will be 0 as long as the state can be retrieved. It is not affected by what that state actually is. The string check of the output is also improved to explicitly look for the full `ActiveState` output.

**Testing done:**

Built and deployed instance prefix. Ran:

```sh
systemctl stop chronyd
apiclient report cis
```

Noted the 2.1.1.1 check reported `SKIP`, indicating manual checking was required.

Built and deployed instance with this fix included. Ran:

```sh
systemctl stop chronyd
apiclient report cis
```

Verified check now reported as `FAIL`. Then ran:

```sh
systemctl start chronyd
apiclient report cis
```

Verified reported `PASS` with service running.

**Terms of contribution:**

### DIFF
--- a/sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
+++ b/sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
@@ -287,8 +287,8 @@ impl Checker for BR02010101Checker {
 
         check_output_contains!(
             SYSTEMCTL_CMD,
-            ["is-active", "chronyd"],
-            &["active"],
+            ["show", "--property", "ActiveState", "chronyd"],
+            &["ActiveState=active"],
             "unable to verify chronyd service enabled",
             "chronyd NTP service is not enabled"
         )


sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -285,10 +285,10 @@ impl Checker for BR02010101Checker {
             return result;
         }
 
-        check_output_contains!(
+        check_output_not_contains!(
             SYSTEMCTL_CMD,
             ["is-active", "chronyd"],
-            &["active"],
+            &["inactive"],
```
bcressey: This would miss the other non-active values in `ActiveState`, like "failed".

You could do the equivalent of:
```
systemctl show --property ActiveState chronyd
```

And check that the output contains `ActiveState=active`.

stmcginnis: Good call!

yeazelm (approved):

bcressey (approved):

PR #3555, opened 2023-10-30
settings: implement motd settings extension

cbgbt:
**Description of changes:**
This uses the `bottlerocket-settings-sdk` to add the first Bottlerocket settings extension for the `motd` setting.

The binary exists now, but we are currently just pulling in the crate as a library into the monolithic models until we can replace more of the settings with extensions.

RPMs to install the setting will be provided at a later time.

**Testing done:**
* Launched a aws-k8s-1.2{5,6,7} Bottlerocket images and checked `settings.motd`

**Terms of contribution:**


cbgbt:
I must have made some mistake locally, but I'm getting an entirely different error locally now. This needs more work. I'll re-open this when I'm more confident that it's ready.

PR #3560, opened 2023-10-31
settings: implement motd settings extension

cbgbt:
**Description of changes:**
This uses the `bottlerocket-settings-sdk` to add the first Bottlerocket settings extension for the `motd` setting.

The binary exists now, but we are currently just pulling in the crate as a library into the monolithic models until we can replace more of the settings with extensions.

**Testing done:**
* Launched a aws-k8s-1.2{5,6,7} Bottlerocket images and checked `settings.motd`
* Added the settings extension binary to `aws-k8s-1.27` and ran  `sudo sheltie settings-extension-motd`

**Terms of contribution:**


cbgbt:
^ Forgot to run the non-unit-test checks in my local workspace.

packages/settings-motd/settings-motd.spec
```diff
@@ -0,0 +1,30 @@
...
+%cargo_prep
+
+%build
+mkdir bin
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p settings-extension-motd
+
+%install
+install -d %{buildroot}%{_cross_bindir}
+install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/settings-extension-motd %{buildroot}%{_cross_bindir}
```
cbgbt: I need to change the install paths to be in-line with the disk layout [called out in the design doc](https://github.com/cbgbt/bottlerocket/blob/motd-extension-again/docs/design/settings/extensions/settings-extension-example.md#settings-extension-binaries)

packages/settings-motd/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+source-groups = [
+    "models/settings-extensions/motd",
+]
```
bcressey: The `os` package watches the top-level "models" source group, so changes to these files will result in both packages rebuilding.

packages/settings-motd/settings-motd.spec
```diff
@@ -0,0 +1,30 @@
...
+
+%description
+%{summary}.
+
+%prep
+%setup -T -c
+%cargo_prep
+
+%build
+mkdir bin
```
bcressey: Why is this `mkdir` needed?

packages/settings-motd/settings-motd.spec
```diff
@@ -0,0 +1,30 @@
...
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p settings-extension-motd
```
bcressey: How large is this file? Wondering if it is useful to link with the settings SDK as a dylib to reduce the footprint.

I'm also concerned about the potential hit to build times if we have multiple extensions in different packages, rather than collecting them in one and building the shared dependencies once (as `os` does).

sources/deny.toml
```diff
@@ -58,7 +58,11 @@ wildcards = "deny"
 
 deny = [
     { name = "structopt" },
-    { name = "clap", wrappers = [ "cargo-readme" ] },
+
+    # The bottlerocket-settings-sdk uses clap, but plans to migrate to argh
+    # We will provide an exception until the following is resolved:
+    # https://github.com/bottlerocket-os/bottlerocket-settings-sdk/issues/23
+    { name = "clap", wrappers = [ "cargo-readme", "bottlerocket-settings-sdk" ] },
```
bcressey: Can we prioritize this? If this exception goes in then inevitably it will be used to justify another exception.

I don't have a problem with clap as such but I don't think we need more than one argument parsing crate in the distro.

sources/models/settings-extensions/motd/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "settings-extension-motd"
+version = "0.1.0"
+authors = ["Sean P. Kelly <seankell@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
bcressey:
```suggestion
```

sources/models/settings-extensions/motd/motd.toml
```diff
@@ -0,0 +1,14 @@
...
+default-version = "v1"
+
+
+[v1]
+    [v1.validation.cross-validates]
+
+    [v1.templating]
+        helpers = []
+
+    [v1.generation.requires]
```
bcressey: nit: I kind of hate the indented style for TOML

sources/models/src/metal-k8s-1.28/mod.rs
```diff
@@ -13,7 +13,7 @@ use crate::{
 // that uses its name in serialization; internal structures use the field name that points to it
 #[model(rename = "settings", impl_default = true)]
 struct Settings {
-    motd: String,
+    motd: settings_extension_motd::MotdV1,
```
bcressey, line 16: Do we need to convert all other top-level fields in these models to settings extensions before we can get rid of the overarching model?

sources/models/string_impls_for/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "string_impls_for"
+version = "0.1.0"
+authors = ["Sean P. Kelly <seankell@amazon.com>"]
```
bcressey: Let's add the original authors here too since it's refactoring out of an existing crate.

sources/models/src/metal-k8s-1.28/mod.rs
cbgbt: Yes.

packages/settings-motd/Cargo.toml
cbgbt: Do you think it's reasonable to hoist `settings-extensions` out of `models` to resolve this?

cbgbt: After tossing this around, we ultimately want to remove `models`. I placed `settings-extensions` there given that it's relevant in the current design, but I think it should be moved.

packages/settings-motd/settings-motd.spec
cbgbt: The file is ~350k. I haven't attempted to compile rust dylibs, but I can do a PoC to see how the size changes. This could be worth it given we anticipate ~20 extensions to be installed in the k8s variants.

cbgbt: > I'm also concerned about the potential hit to build times if we have multiple extensions in different packages

I added this as an individual package since it will need to be this way for out-of-tree builds to pick and choose their extension sets. We can join extension packages and opt to split them later, though.

cbgbt:
^ Responds to feedback from @bcressey and myself:
* Installs the extension binary in `/usr/libexec` and succinct symlinks in `/usr/libexec/settings/`
* Moves `settings-extensions` to a top-level directory in `sources` to allow independent rebuild logic between `os` and extension packages.
* Eliminates `clap` from the settings SDK.
* Adds author attributions for `string_impls_for!`
* Modifies whitespace in `motd.toml`

cbgbt:
^ The above rectifies a mistake in which code is included in which commits. I noticed that the removal of the `clap` exception from `deny.toml` was in the wrong commit.

sources/settings-extensions/motd/motd.toml
```diff
@@ -0,0 +1,13 @@
+[extension]
+supported-versions = [
+    "v1"
+]
```
bcressey, line 4: Does this file need to be installed in the spec?

bcressey (approved):

zmrow (approved):
Provided @bcressey 's comment re: `sources/settings-extensions/motd/motd.toml` is addressed - this looks good!

sources/settings-extensions/motd/motd.toml
cbgbt: It does need to be installed, but I would like to do so in a followup after completing https://github.com/bottlerocket-os/twoliter/issues/111. As it is now, I'd probably need to move the config to the RPM, but I consider it to be source code of the extension and think that it belongs in the crate.

PR #3562, opened 2023-11-02
Improve CIS UDF kernel module check

stmcginnis:
**Issue number:**

N/A

**Description of changes:**

The level 2 check 1.1.1.1 verifies mounting UDF filesystems is disabled. The current check for whether it is already loaded was not correct. Luckily there is a second check as part of this control to make sure loading is disabled. If the setting for loading is to not allow it, but the module is already loaded, the check does not return the expected output. So it would still report failure, but it is less than ideal for reporting the actual issue.

This changes the check for whether the module is loaded to correctly identify if the module is loaded or not before checking whether the ability has been disabled.

**Testing done:**

Ran the level 2 report on a default deployment and verified it reported `FAIL` for this check.

Checked the loaded but disabled scenario:

```sh
bash-5.1# lsmod | grep udf

bash-5.1# modprobe -v -n udf
insmod /lib/modules/5.15.134/kernel/drivers/cdrom/cdrom.ko.xz 
insmod /lib/modules/5.15.134/kernel/lib/crc-itu-t.ko.xz 
insmod /lib/modules/5.15.134/kernel/fs/udf/udf.ko.xz

bash-5.1# insmod /lib/modules/5.15.134/kernel/drivers/cdrom/cdrom.ko.xz

bash-5.1# insmod /lib/modules/5.15.134/kernel/lib/crc-itu-t.ko.xz

bash-5.1# insmod /lib/modules/5.15.134/kernel/fs/udf/udf.ko.xz

bash-5.1# grep udf /proc/modules 
udf 139264 0 - Live 0xffffffffc0636000
crc_itu_t 16384 1 udf, Live 0xffffffffc0631000
cdrom 77824 1 udf, Live 0xffffffffc061d000

bash-5.1# apiclient set kernel.modules.udf.allowed=false

bash-5.1# modprobe -v -n udf                             

bash-5.1# apiclient report cis -l 2
...
[FAIL] 1.1.1.1   Ensure mounting of udf filesystems is disabled (Automatic)
...
```

Then unloaded the module and made sure it correctly identifies the module is not loaded and loading is disabled:

```sh
bash-5.1# rmmod udf

bash-5.1# modprobe -v -n udf                             
install /bin/true 

bash-5.1# apiclient report cis -l 2
...
[PASS] 1.1.1.1   Ensure mounting of udf filesystems is disabled (Automatic)
...
```

Then unload

**Terms of contribution:**

### DIFF
--- a/sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
+++ b/sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
@@ -22,15 +22,22 @@ pub struct BR01010101Checker {}
 
 impl Checker for BR01010101Checker {
     fn execute(&self) -> CheckerResult {
-        let module_result = check_file_contains!(
-            PROC_MODULES_FILE,
-            &[" udf,"],
-            "unable to parse modules to check for udf",
-            "udf is currently loaded"
-        );
-        if module_result.status != CheckStatus::PASS {
+        let mut module_result = CheckerResult::default();
+
+        // Make sure UDF isn't already loaded
+        if let Ok(found) = look_for_word_in_file(PROC_MODULES_FILE, "udf") {
+            if found {
+                module_result.error = "udf is currently loaded".to_string();
+                module_result.status = CheckStatus::FAIL;
+                return module_result;
+            }
+        } else {
+            module_result.error =
+                "unable to parse modprobe output to check if udf is enabled".to_string();
             return module_result;
         }
+
+        // Make sure the ability to load UDF is disabled
         check_output_contains!(
             MODPROBE_CMD,
             ["-n", "-v", "udf"],
--- a/sources/bloodhound/src/lib.rs
+++ b/sources/bloodhound/src/lib.rs
@@ -9,6 +9,16 @@ pub mod args;
 pub mod output;
 pub mod results;
 
+/// Reads a file and checks if the given `search_word` is present in its contents.
+pub fn look_for_word_in_file(path: &str, search_word: &str) -> Result<bool, io::Error> {
+    let reader = BufReader::new(File::open(path)?);
+    Ok(reader.lines().any(|line| {
+        line.unwrap_or_default()
+            .split_ascii_whitespace()
+            .any(|word| word == search_word)
+    }))
+}
+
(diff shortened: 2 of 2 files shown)


sources/bloodhound/src/bin/bottlerocket-checks/checks.rs
```diff
@@ -22,15 +22,22 @@ pub struct BR01010101Checker {}
...
-            PROC_MODULES_FILE,
-            &[" udf,"],
-            "unable to parse modules to check for udf",
-            "udf is currently loaded"
-        );
-        if module_result.status != CheckStatus::PASS {
+        let mut module_result = CheckerResult::default();
+
+        // Make sure UDF isn't already loaded
+        if let Ok(found) = look_for_string_in_file(PROC_MODULES_FILE, " udf,") {
```
bcressey: It would be slightly better to check for `"udf "` - `" udf,"` is matching a line for a module that's a dependency of udf, and only matches if udf is the first module in that list.

However, it seems worth adding a different function, like `look_for_word_in_file` that calls `split_ascii_whitespace` on the input and checks for an exact token match instead of using `contains`, to avoid false positives for short inputs like these.

sources/bloodhound/src/lib.rs
```diff
@@ -197,6 +207,46 @@ mod test_utils {
...
+    #[test]
+    fn test_look_for_word_in_file_not_found() {
+        let mut test_file = NamedTempFile::new().unwrap();
+        writeln!(
+            test_file,
+            concat!(
+                "crypto_simd 16384 1 aesni_intel, Live 0xffffffffc034f000\n",
+                "cryptd 28672 2 ghash_clmulni_intel,crypto_simd, Live 0xffffffffc0335000\n",
+                "configfs 57344 1 - Live 0xffffffffc0320000\n"
+            )
```
bcressey, line 236: Let's add a test case that a partial match like "my-udf" isn't found.

bcressey (approved):
Nice!

sources/bloodhound/src/lib.rs
stmcginnis: Done!

yeazelm (approved):
LGTM!

PR #3563, opened 2023-11-02
models: move modeled_types to its own crate

cbgbt:
**Description of changes:**

```
models: move modeled_types to its own crate

This allows settings extensions to be added to the Bottlerocket repository that use our existing modeled types without introducing a circular dependency between `models` and the extension.
```

**Testing done:**
* Smoke testing of an existing Bottlerocket variant
* No functional changes. If all CI builds succeed, this should be acceptable.

**Terms of contribution:**


sources/models/modeled_types/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "modeled_types"
+version = "0.1.0"
+authors = []
```
zmrow: No authors?

zmrow (approved):

sources/models/modeled_types/Cargo.toml
cbgbt: I think the models have been contributed to by the majority of Bottlerocket developers. I'm not sure I could fairly construct an authors list!

sources/models/Cargo.toml
```diff
@@ -10,24 +10,12 @@ build = "build.rs"
...
 bottlerocket-release = { path = "../bottlerocket-release", version = "0.1" }
-indexmap = { version = "1", features = ["serde"] }
-lazy_static = "1"
 libc = "0.2"
 model-derive = { path = "model-derive", version = "0.1" }
-regex = "1"
-scalar = { path = "scalar", version = "0.1" }
-scalar-derive = { path = "scalar-derive", version = "0.1" }
-semver = "1"
+modeled_types = { path = "modeled_types", version = "0.1" }
```
bcressey: For consistency? We tend to use the kebab form elsewhere.

```suggestion
modeled-types = { path = "modeled-types", version = "0.1" }
```

bcressey (approved):

cbgbt:
^ Renames `modeled_types` to `modeled-types` per suggestion from @bcressey

etungsten (approved):

PR #3569, opened 2023-11-03
adds ecs-2 to the readme

stockholmux:
**Issue number:**

Closes # n/a

**Description of changes:**
I noticed that the variant list didn't have `ecs-2` (available since 1.15.0) and that we reference `ecs-1` as an example. Since `ecs-2` is the most recent, this PR makes it more prominent than `ecs-1`.

**Testing done:**

👁️ 👁️ 

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ The following variants support ECS:
 
 * `aws-ecs-1`
 * `aws-ecs-1-nvidia`
+* `aws-ecs-2`
+* `aws-ecs-2-nvidia`
 
 We also have variants that are designed to be Kubernetes worker nodes in VMware:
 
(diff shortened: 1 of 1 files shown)


stmcginnis (approved):
Good catch!

stockholmux:
One additional note: we have to manually update this. Bottlerocket.dev has a script that runs on every release to pull this info. For example:

https://bottlerocket.dev/en/os/1.16.x/version-information/packages-by-variant/

That page is not exactly the same as what is presented here, but I'm glad to create a simpler page on the website that lists the information in a format just like this, then we can refer people to bottlerocket.dev and remove this section.  This eliminates the need to update the readme.

bcressey (approved):

PR #3592, opened 2023-11-10
Fix for invalid chars when returning IPv6 hostname

### LINKED ISSUE #3571: Fail to get ValidLinuxHostname with ipv6 link-local address
**Image I'm using:**
I'm building a custom 1.15.1 metal-dev variant build with additional drivers turned on but I speculate this is a core issue and not specific to my build.  I've seen it with previous version builds as well but believe it has only ever happened on networkd based builds not wicked.

When booting up I'm expecting to generate initial settings correctly, specifically for the hostname.

**What actually happened:**
sundog[1606]: Error deserializing hashMap to Settings: Error deserializing scaler value: Unable to deserialize into ValidLinuxHostname: 'fe80::dea6:32ff:fea9:513c' must only be [0-9a-z.-], and 1-253 chars long

It doesn't happen on every initial build/boot but fairly regularly I can't generate initial settings on a clean build.  Just boot and get this error and forward progress halts and my config settings seem to be corrupted from then on and I can't boot.  Occasionally I can reflash/purge the config directory and beat the race condition and it gives me a hostname based on a valid IPv4 address.  I don't currently have IPv6 turned on for this lan segment so I've never seen it fail with a non link-local IPv6.

gregular:
**Issue number:**

Closes #3571 

**Description of changes:**
Netdog will return an IPv6 address string in some corner cases.  Replace colons with dashes to ensure the string will validate as ValidLinuxHostname.

**Testing done:**
Validated locally when no valid IPv4 is assigned by `systemd-networkd` _but_ a valid IPv6 is assigned when `systemd-networkd-wait-online.service` times out that the string netdog returns can be validated as ValidLinuxHostname

**Terms of contribution:**

### DIFF
--- a/sources/api/netdog/src/cli/generate_hostname.rs
+++ b/sources/api/netdog/src/cli/generate_hostname.rs
@@ -54,8 +54,8 @@ pub(crate) async fn run() -> Result<()> {
     } else {
         hostname
     }
-    // If no hostname has been determined we return the IP address of the host.
-    .unwrap_or(ip_string);
+    // If no hostname has been determined we return the IP address of the host, replacing invalid ipv6 chars.
+    .unwrap_or(ip_string.replace(":", "-"));
 
     // sundog expects JSON-serialized output
     print_json(hostname)


yeazelm (approved):
Very nice! Thanks for the PR!

webern (approved):
Makes sense to me.

zmrow (approved):
Thanks for this! :partying_face:

PR #3595, opened 2023-11-13
settings: add ntp settings extension

sam-berning:
**Description of changes:**

Implements a settings extension for NTP settings using the `bottlerocket-settings-sdk`.

Because it uses the SDK, this will create a settings binary which can be installed on a variant via the `settings-ntp` package. But for now, we are only using the settings model from the extension.

**Testing done:**

- ran a build of `aws-dev` on an ec2 instance and checked that the NTP settings were set correctly on boot and would update `/etc/chrony.conf` when set to a new value
- included the settings extension binary on a build of `aws-dev`, and made sure the binary gave the expected responses.

**Terms of contribution:**

### DIFF
--- a/sources/settings-extensions/ntp/Cargo.toml
+++ b/sources/settings-extensions/ntp/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "settings-extension-ntp"
+version = "0.1.0"
+authors = ["Sam Berning <bernings@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+
+[dependencies]
+modeled-types = { path = "../../models/modeled-types", version = "0.1" }
+model-derive = { path = "../../models/model-derive", version = "0.1" }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+string_impls_for = { version = "0.1", path = "../../models/string_impls_for" }
+
+[dependencies.bottlerocket-settings-sdk]
+git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
+tag = "bottlerocket-settings-sdk-v0.1.0-alpha.1"
+version = "0.1.0-alpha"
--- a/sources/settings-extensions/ntp/Cargo.toml
+++ b/sources/settings-extensions/ntp/Cargo.toml
@@ -7,11 +7,11 @@ edition = "2021"
 publish = false
 
 [dependencies]
+env_logger = "0.10"
 modeled-types = { path = "../../models/modeled-types", version = "0.1" }
 model-derive = { path = "../../models/model-derive", version = "0.1" }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-string_impls_for = { version = "0.1", path = "../../models/string_impls_for" }
 
 [dependencies.bottlerocket-settings-sdk]
 git = "https://github.com/bottlerocket-os/bottlerocket-settings-sdk"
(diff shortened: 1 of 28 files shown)


sources/settings-extensions/ntp/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+
+[dependencies]
+modeled-types = { path = "../../models/modeled-types", version = "0.1" }
+model-derive = { path = "../../models/model-derive", version = "0.1" }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+string_impls_for = { version = "0.1", path = "../../models/string_impls_for" }
```
cbgbt: Looks like we don't use this dependency, you can remove it.

sources/settings-extensions/ntp/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "settings-extension-ntp"
+version = "0.1.0"
+authors = ["Sam Berning <bernings@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+
+[dependencies]
+modeled-types = { path = "../../models/modeled-types", version = "0.1" }
```
cbgbt, line 11: This appears to be unused, and can be removed.

I tend to like [`cargo-machete`](https://github.com/bnjbvr/cargo-machete) for finding cases like these, though it is imprecise.

sources/settings-extensions/ntp/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "settings-extension-ntp"
+version = "0.1.0"
+authors = ["Sam Berning <bernings@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+
+[dependencies]
```
cbgbt, line 9: The `settings-sdk` emits logs to `stderr`, and I realized after talking with @zmrow that I never initialized the logger in the `motd` extension.

I would take a dependency on [`env_logger`](https://crates.io/crates/env_logger) and initialize the logger just before initializing the extension with `env_logger::init()`.

sources/settings-extensions/ntp/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+        _dependent_settings: Option<serde_json::Value>,
+    ) -> Result<GenerateResult<Self::PartialKind, Self>> {
+        Ok(GenerateResult::Complete(
+            existing_partial.unwrap_or_default(),
+        ))
+    }
+
+    fn validate(_value: Self, _validated_settings: Option<serde_json::Value>) -> Result<()> {
+        // TODO: add additional validation -- legal for this to be set to any URL or do we need to
+        // check that it is a valid time server?
```
cbgbt: We tend not to rely on external services when validating settings as long as they are the same shape, and I think that's the right move. It would be too easy for a transient error to cause unnecessary failure at scale.

cbgbt (commented):
Nice work! Just a few suggestions.

sources/settings-extensions/ntp/Cargo.toml
stmcginnis: +1 for `cargo machete`, though I've noticed it doesn't recognize relative imports like these. So just a warning.

sam-berning: I do use this one for the `Url` type in [lib.rs](https://github.com/bottlerocket-os/bottlerocket/pull/3595/files#diff-34509cee9ac5e2fdc3b51e71de432bbf936201fc7e22a3281f38ef87306f7d2cR5)

cbgbt: Whoops! Sorry, I missed that.

cbgbt (approved):

zmrow (approved):

PR #3605, opened 2023-11-17
docs: Update VERSION in metal readme

yeazelm:
The version has fallen out of date as time moves on and users often get confused on what version they should choose. This changes the logic to lookup the latest release and use that instead of a hardcoded version.

**Terms of contribution:**

### DIFF
--- a/PROVISIONING-METAL.md
+++ b/PROVISIONING-METAL.md
@@ -34,11 +34,13 @@ sha512sum -c <<<"a3c58bc73999264f6f28f3ed9bfcb325a5be943a782852c7d53e803881968e0
 
 Next, set your desired version and variant, and use `tuftool` to download the image:
 To install `tuftool` you'll need to install Rust (via [rustup](https://rustup.rs/) or the official site), and then you can run `cargo install tuftool`.
+The VERSION corresponds to the [Bottlerocket version](https://github.com/bottlerocket-os/bottlerocket/releases/latest), the latest release is almost always what you want.
+You might need to install `jq` to fetch the VERSION.
 
 ```shell
 ARCH="x86_64"
-VERSION="v1.12.0"
-VARIANT="metal-k8s-1.23"
+VERSION=$(curl -s https://api.github.com/repos/bottlerocket-os/bottlerocket/releases/latest | jq -r '.name')
+VARIANT="metal-k8s-1.28"
 IMAGE="bottlerocket-${VARIANT}-${ARCH}-${VERSION}.img.lz4"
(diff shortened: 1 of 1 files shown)


PROVISIONING-METAL.md
```diff
@@ -34,11 +34,12 @@ sha512sum -c <<<"a3c58bc73999264f6f28f3ed9bfcb325a5be943a782852c7d53e803881968e0
 
 Next, set your desired version and variant, and use `tuftool` to download the image:
 To install `tuftool` you'll need to install Rust (via [rustup](https://rustup.rs/) or the official site), and then you can run `cargo install tuftool`.
+The VERSION corresponds to the [Bottlerocket version](https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.16.1), the latest release is almost always what you want.
```
stmcginnis: Rather than linking to v1.16.1, I think it might be better to just go the [releases page](https://github.com/bottlerocket-os/bottlerocket/releases) and let them find it. Your comment about latest release kind of points to that anyway.

PROVISIONING-METAL.md
```diff
@@ -34,11 +34,12 @@ sha512sum -c <<<"a3c58bc73999264f6f28f3ed9bfcb325a5be943a782852c7d53e803881968e0
...
 
 ```shell
 ARCH="x86_64"
-VERSION="v1.12.0"
-VARIANT="metal-k8s-1.23"
+VERSION=$(curl -s https://api.github.com/repos/bottlerocket-os/bottlerocket/releases/latest | jq -r '.name')
+VARIANT="metal-k8s-1.28"
```
stmcginnis, line 43: Use of `jq` implies a non-standard prereq, but I think for this audience that would be OK.

PROVISIONING-METAL.md
yeazelm: Oh, good catch, I tried to copy the `latest` url which ended up redirecting me to 1.16.1 and that is the url I copied. `https://github.com/bottlerocket-os/bottlerocket/releases/latest` is what I typed to get there :) I'll update to that which should land the user on the latest release.

PROVISIONING-METAL.md
yeazelm: This is a good callout. We mention installing `jq` for the quickstarts and I assumed we had called that out here too. I can add a line saying you might need to install it to use that one-liner.

stmcginnis (approved):
Looks good!

cbgbt (approved):

PR #3633, opened 2023-11-30
Move template migrations to schnauzer v2

### LINKED ISSUE #3133: Re-implement Bottlerocket template system for Out-of-tree Builds

cbgbt:
**Issue number:**

Closes #3133

**Description of changes:**
The template migrator takes preventative steps to avoid overwriting user-provided settings or setting generators when they were initially generated by a setting generator. It does this by running the old generator and checking to see if the current value is a match (implying it has not been overwritten) before taking action.

The concept of "template metadata" disappears with schnauzer-v2, leaving behind a binary invocation which includes template information. This PR changes the schnauzer-v2 CLI to become a library that we can link against and invoke in migrations.

The previous schnauzer migration lives on (but deprecated) until such a time that we can archive all migrations that depend on it.

**Testing done:**
* [x] Provided unit tests pass
* [x] Migration testing new host containers passes forwards & backwards checks

**Terms of contribution:**


etungsten (approved):
:rocket:

zmrow (approved):

PR #3637, opened 2023-12-02
Remove conditional compilation from `early-boot-config`

### LINKED ISSUE #3616: OOTB: Remove conditional compilation from `early-boot-config`
Eliminate conditional compilation from `early-boot-config` by getting rid of platform-specific logic.  User data “providers” will become small discreet binary packages that `early-boot-config` will discover at runtime and run in order from a `.d`-style directory
* User data provider code will continue to live in the `early-boot-config` crate.  They will be refactored into discrete binaries that consume modules where each data source implements a trait.
  * Binaries will implement a single user data source, as in a single file, or IMDS, etc.
  * Binaries will return an `ExitCode` signifying success or failure, and if user data is found, return a `SettingsJson` object on stdout.
* Each binary will have it’s own package in the `packages/` directory.
  * The package will contain 2 subpackages, a `-client` package and a `-provider` package. 
  * The `-client` package will contain just the binary installed to `/usr/libexec/early-boot-config/bin/`
  * The `-provider` package will symlink the binary to `/usr/share/early-boot-config/providers.d` with the appropriate prefix (10-, 20-, etc)
* Another metapackage `user-data-providers` will be created that will have subpackages for each variant type (AWS, VMware, metal).  The subpackages will pull in the correct user data provider packages.
* Variants will consume the user-data-provider subpackage as necessary

zmrow:
**Issue number:**

Closes #3616 

**Description of changes:**
*This PR is best read in commit order!*  Commits are fairly small and self-contained, but there are a few of them. :)

The strategy for this PR is largely laid out in #3616, but the changes in this PR remove all conditional compilation from `early-boot-config` by using discrete "user data provider" binaries for each applicable user data source.

Logic for fetching user data was largely "lifted and shifted" to discrete modules for each source.  A new trait `UserDataProvider` was added, with each user data source implementing said trait (local file, IMDS, guestinfo, etc). 

Each user data source is bundled into it's own binary within the `early-boot-config` package.  Each binary is packaged separately with its own package in the `packages/` directory.  Each package includes two subpackages, a `-client` package which includes _just_ the binary, and a `-provider` package that symlinks the binary into a `.d`-style directory in the proper order.  Ordering of sources stays identical to the ordering we have today.

All of the user data provider packages are bundled into an `early-boot-config-providers` metapackage that contains subpackages for each variant type: AWS, VMware, metal.  The variant type determines the proper providers. 
 Each variant then consumes the corresponding metapackage, i.e `aws-k8s-1.28` will consume the `aws-data-providers` package.

At runtime, `early-boot-config` consults said `.d` directory (`/usr/share/early-boot-config/data-providers.d`) and runs the binaries in the proper order.  If user data is found at the source, the binary returns a serialized `SettingsJson` object on stdout, along with a successful exit code.  If no user data is found, but no problems were encountered, it exits successfully, but outputs nothing.  If any errors are encountered, the binary exits unsuccessful and prints its error.

Similar to today's logic, `early-boot-config` submits the user data to the API in order.

Example output, showing the proper order and provider log output on an `aws-dev` variant:
```
bash-5.1# journalctl -u early-boot-config
Dec 04 18:41:07 localhost systemd[1]: Starting Bottlerocket userdata configuration system...
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] early-boot-config started
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Gathering user data providers
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Provider '10-local-defaults': [2023-12-04T18:41:07Z INFO  early_boot_config::provider] /local/user-data-defaults.toml does not exist, not using it
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] No user data found via /usr/share/early-boot-config/data-providers.d/10-local-defaults
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Provider '20-local': [2023-12-04T18:41:07Z INFO  early_boot_config::provider] /var/lib/bottlerocket/user-data.toml does not exist, not using it
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] No user data found via /usr/share/early-boot-config/data-providers.d/20-local
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Provider '30-ec2-identity-doc': [2023-12-04T18:41:07Z INFO  early_boot_config::provider::ec2_identity_doc] Using IMDS for region
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Provider '30-ec2-identity-doc': [2023-12-04T18:41:07Z INFO  imdsclient] Received dynamic/instance-identity/document
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Found user data via EC2 instance identity document, sending to API
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Provider '40-ec2-imds': [2023-12-04T18:41:07Z INFO  early_boot_config::provider::ec2_imds] Fetching user data from IMDS
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Provider '40-ec2-imds': [2023-12-04T18:41:07Z INFO  imdsclient] Received user-data
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Found user data via EC2 IMDS, sending to API
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] Provider '50-local-overrides': [2023-12-04T18:41:07Z INFO  early_boot_config::provider] /local/user-data-overrides.toml does not exist, not using it
Dec 04 18:41:07 localhost early-boot-config[1017]: [2023-12-04T18:41:07Z INFO  early_boot_config] No user data found via /usr/share/early-boot-config/data-providers.d/50-local-overrides
Dec 04 18:41:07 localhost systemd[1]: Finished Bottlerocket userdata configuration system.
```

An example from VMware showing proper sources (CD-ROM and guestinfo both include user data):
```
bash-5.1# journalctl -u early-boot-config
Dec 07 20:38:51 localhost systemd[1]: Starting Bottlerocket userdata configuration system...
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] early-boot-config started
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Gathering user data providers
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Provider '10-local-defaults': [2023-12-07T20:38:51Z INFO  early_boot_config::provider] /local/user-data-defaults.toml does not exist, not using it
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] No user data found via /usr/share/early-boot-config/data-providers.d/10-local-defaults
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Provider '20-local': [2023-12-07T20:38:51Z INFO  early_boot_config::provider] /var/lib/bottlerocket/user-data.toml does not exist, not using it
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] No user data found via /usr/share/early-boot-config/data-providers.d/20-local
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Provider '30-vmware-cd-rom': [2023-12-07T20:38:51Z INFO  early_boot_config::provider::vmware_cd_rom] Attempting to retrieve user data from mounted CD-ROM
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Provider '30-vmware-cd-rom': [2023-12-07T20:38:51Z INFO  early_boot_config::provider::vmware_cd_rom] '/media/cdrom/ovf_env.xml' exists, using it
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Found user data via CD-ROM, sending to API
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Provider '40-vmware-guestinfo': [2023-12-07T20:38:51Z INFO  early_boot_config::provider::vmware_guestinfo] Attempting to retrieve user data via guestinfo interface
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Provider '40-vmware-guestinfo': [2023-12-07T20:38:51Z INFO  early_boot_config::provider::vmware_guestinfo] Found user data encoding: base64
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Provider '40-vmware-guestinfo': [2023-12-07T20:38:51Z INFO  early_boot_config::provider::vmware_guestinfo] Decoding user data
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Found user data via guestinfo, sending to API
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] Provider '50-local-overrides': [2023-12-07T20:38:51Z INFO  early_boot_config::provider] /local/user-data-overrides.toml does not exist, not using it
Dec 07 20:38:51 localhost early-boot-config[1697]: [2023-12-07T20:38:51Z INFO  early_boot_config] No user data found via /usr/share/early-boot-config/data-providers.d/50-local-overrides
Dec 07 20:38:51 localhost systemd[1]: Finished Bottlerocket userdata configuration system.
```

**Testing done:**

- [x] Ensure ordering is correct for each variant type (AWS, VMware, metal)
  - [x] AWS (`aws-dev`, `aws-k8s-1.27`, `aws-k8s-1.28`)
  - [x] VMware (`vmware-k8s-1.28`)
  - [x] metal (`metal-dev`)
- [x] Test with multiple sources present with distinct settings (`aws-k8s-1.28`): Observe settings from EC2 instance identity doc, and user data from IMDS both populated in the API. 
- [x] Test with multiple sources present with setting overrides (`metal-dev`): Set `settings.motd` via user-data, but overrode it with `/local/user-data-overrides.toml`.  Observe the overriden value present in the API settings.
- [x] Test all providers
  - [x] Files
  - [x] IMDS
  - [x] Identity doc
  - [x] CD-ROM (used an `ovf_env.xml` file inside an ISO attached to VMware VM)
  - [x] Guestinfo 

**Terms of contribution:**


zmrow:
^ Added additional `build-dependencies` to the variant `Cargo.toml`'s.

zmrow:
^ Remove offending `--bin` from `early-boot-config`'s `%build` line in `os.spec` that was causing builds to fail

zmrow:
^ In the `os` package, build `early-boot-config` separately with the `--bin` flag.  This ensures only that binary is built, and not the user data providers which may have architecture-dependent components.

zmrow:
^ Ensure VMware guestinfo provider is only included for x86_64

zmrow:
^ Added an `early-boot-config` package, moving it out of the `os` package.  Renamed all the related user data provider packages with the prefix `early-boot-config`

sources/api/early-boot-config/src/provider.rs
```diff
@@ -32,3 +37,95 @@ pub trait PlatformDataProvider {
...
+    let path = path.as_ref();
+
+    if !path.exists() {
+        info!("{} does not exist, not using it", path.display());
+        return Ok(None);
+    }
+    info!("'{}' exists, using it", path.display());
+
+    // Read the file, decompressing it if compressed.
+    let user_data_str = expand_file_maybe(path).context(error::InputFileReadSnafu { path })?;
```
cbgbt: This is cool

packages/early-boot-config/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-package]
+source-groups = [
+    "api/early-boot-config",
```
cbgbt: One unfortunate side effect of this is that changes to `early-boot-config` will still trigger rebuilds of the `os` package.

Some ideas to solve this:
1. Move `early-boot-config` outside of `api` module
2. Add options to buildsys to have some sort of "ignored source-groups" which remove anything that should not be included
3. Disregard the rebuilds in this particular scenario.


I personally like option 2. Maybe we cut an issue for that here.

sources/api/early-boot-config/src/provider/ec2_identity_doc.rs
```diff
@@ -0,0 +1,108 @@
+/// EC2 Identity Document
```
cbgbt: I'm a little surprised that this PR doesn't remove any code. How were we doing this in the past? Does that code need to change or go away?

sources/api/early-boot-config/build.rs
```diff
@@ -1,19 +1,3 @@
...
-                "For local builds, you must set the '{}' environment variable so we know \
-                which data provider to build. Valid values are the directories in \
-                models/src/variants/, for example 'aws-ecs-1': {}",
-                VARIANT_ENV, e,
-            );
-            std::process::exit(1);
-        }
-    };
-    variant.emit_cfgs();
-
```
cbgbt, line 17: 😎 🚀

variants/aws-dev/Cargo.toml
```diff
@@ -47,6 +48,12 @@ path = "../variants.rs"
 [build-dependencies]
 # core
 early-boot-config = { path = "../../packages/early-boot-config" }
+early-boot-config-ec2-identity-doc = { path = "../../packages/early-boot-config-ec2-identity-doc" }
+early-boot-config-ec2-imds = { path = "../../packages/early-boot-config-ec2-imds" }
+early-boot-config-local-defaults = { path = "../../packages/early-boot-config-local-defaults" }
+early-boot-config-local-overrides = { path = "../../packages/early-boot-config-local-overrides" }
+early-boot-config-local = { path = "../../packages/early-boot-config-local" }
+early-boot-config-providers = { path = "../../packages/early-boot-config-providers" }
```
cbgbt: I'm a bit surprised by this. Couldn't these just refer to the appropriate metapackage?

sources/api/early-boot-config/src/bin/early-boot-config.rs
```diff
@@ -91,6 +95,85 @@ fn parse_args(args: env::Args) -> Args {
...
 }
 
+/// Gather user data providers to run in order
+fn gather_providers() -> Result<Vec<PathBuf>> {
+    Ok(WalkDir::new(PROVIDERS_DIR)
+        .max_depth(1)
+        .min_depth(1)
+        .sort_by_file_name()
+        .into_iter()
+        .collect::<std::result::Result<Vec<_>, _>>()?
```
cbgbt: There's no problem here, I'm just a bit surprised that this error type doesn't need to be coaxed into a different one, but it's hard to tell in the diff view what the Error type of the returned Result is.

I would anticipate the iterator's error type to be `std::io::Error`.

sources/api/early-boot-config/src/bin/early-boot-config.rs
```diff
@@ -104,39 +187,34 @@ async fn run() -> Result<()> {
...
-        .context(error::ProviderSnafu)?
-    {
-        // Don't send an empty request to the API
-        if settings_json.json.is_empty() {
-            warn!("{} was empty", settings_json.desc);
+    info!("Gathering user data providers");
+    let providers = gather_providers()?;
+    for provider in providers {
+        debug!("Found '{}', running it...", provider.display());
+        let result = run_provider(&args.log_level, &provider).await?;
```
cbgbt: I think we should probably run the providers in parallel, but then apply the results in-order.

sources/api/early-boot-config/src/bin/early-boot-config.rs
```diff
@@ -91,6 +95,85 @@ fn parse_args(args: env::Args) -> Args {
...
+    let provider_logs = String::from_utf8_lossy(&result.stderr);
+    for line in provider_logs.lines() {
+        info!("Provider '{}': {}", provider_name, line);
+    }
+
+    ensure!(
+        result.status.success(),
+        error::ProviderFailureSnafu {
+            provider: &provider,
+        }
```
cbgbt: For error cases, the providers are emitting a string to stdout describing the error. We should probably include it here as part of the error message.

cbgbt (commented):
Excellent work!

sources/api/early-boot-config/src/provider/ec2_identity_doc.rs
cbgbt: I was reading this commit-by-commit. The removal comes in a later commit.

packages/early-boot-config-providers/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+source-groups = [
+    "api/early-boot-config",
+]
+
+# RPM BuildRequires
+[build-dependencies]
+glibc = { path = "../glibc" }
+
+# RPM Requires
+[dependencies]
```
cbgbt: After looking at this with @zmrow we realized we need to add the data provider client/provider packages as dependencies here

bcressey:
We need to wind all of this into a single early-boot-config package with sub-packages.

Some of the tricks that @jmt-lab is using in #3700 should help, like `Supplements` and using a Lua scriptlet to create a symlink.

sources/api/early-boot-config/src/bin/early-boot-config.rs
sam-berning: I was surprised as well, so I played around with this code a bit.

So what's going on here:
1. the iterator collects to a `Result<Vec<walkdir::DirEntry>, walkdir::Error>`
2. snafu implements `From<walkdir::Error>` on `error::Error` due to the `WalkDirSnafu` variant
3. the `?` operator automatically converts the `walkdir::Error` to `error::Error`

I guess I've been using `.context()` more than I need to, lol

webern:
@sam-berning you could get this rebased as a starting point!

webern:
> We need to wind all of this into a single early-boot-config package with sub-packages.
> 
> Some of the tricks that @jmt-lab is using in #3700 should help, like `Supplements` and using a Lua scriptlet to create a symlink.

@bcressey we want to make sure that we compile all of the code (barring VMWare bits when building arm) when we `cd sources && cargo build` and also when we `cargo make` for any variant. Just making sure your packaging suggestion doesn't reintroduce conditional compilation in any way. In #3700 we ended up building twice in the RPM.spec because we didn't actually remove conditional compilation! :sweat_smile: , but with `early-boot-config`, as written in this PR currently, we have. We want to make sure not to lose sight of that prize when contemplating packaging strategy.

bcressey:
@webern - absolutely, I just don't want the packaging scattered across nine spec files.

In some sense each spec is "free" but only if future changes are largely decoupled, and if they don't all need to be understood together to form a mental model of the behavior on the installed system.

One of my packaging regrets is splitting Docker into 4 packages (now down to 3) when it really should have been 1 or 2 all along.

The current scattering of systemd-related functionality across release, os, and host-ctr is another concern I have. That's a symptom of the larger problem is that "os" is too large and monolithic, and ends up containing things because of Rust compile times rather than sensible modular boundaries.

We're on a path to fixing that, and this work is part of the fix, but I don't want to go to a different extreme of micro-packages in the process.

webern:
> We're on a path to fixing that, and this work is part of the fix, but I don't want to go to a different extreme of micro-packages in the process.


One thing to think about is kit design and whether or not you would want sub-packages to belong to different kits.

sam-berning:
^ rebased on latest `develop`. Zac's original branch state is backed up to https://github.com/sam-berning/bottlerocket/tree/zmrow-ebc-src-bin

sam-berning:
^ updated ebc to run providers in parallel but apply their results in order, and packaged everything all in one rpm

sam-berning:
Looks like the `vmw_backdoor` crate only builds for x86, so the `vmware_guestinfo` provider at least will need to be split out into a different rpm. I think I'll extract the `early-boot-config-vmware` package out to its own rpm.

sam-berning:
^ actually ended up creating a dummy version of the `vmware-guestinfo` provider for arm so that we can keep everything in one package for now.

this is technically adding more conditional compilation, but I think it's still in the right direction. I'll mess around with pulling the user data providers into separate crates from each other and from `early-boot-config` so their `Cargo.toml`s aren't coupled and hopefully then we can remove conditional compilation altogether

sam-berning:
^ rebased on latest `develop` and separated the user data provider interface from `early-boot-config` and from the data providers themselves. also conditionally packaged the vmware data providers only if we're building for x86.

sam-berning:
the commits probably need to be squashed a bit before we merge but I wanted to keep as much of the history on this PR as possible while we're reviewing. should be still readable commit-by-commit.

sources/user-data-providers/ec2-identity-doc/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "ec2-identity-doc-user-data-provider"
+version = "0.1.0"
+authors = ["Zac Mrowicki <mrowicki@amazon.com>", "Sam Berning <bernings@amazon.com>"]
```
bcressey: I'd like to add a `generate_readme` build script to each of these providers, along with a top-level README.md in `sources/user-data-providers` describing what these crates are.

sources/user-data-providers/local-defaults/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "local-defaults-user-data-provider"
+version = "0.1.0"
+authors = ["Zac Mrowicki <mrowicki@amazon.com>", "Sam Berning <bernings@amazon.com>"]
+edition = "2021"
+license = "Apache-2.0 OR MIT"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
bcressey: nit: we can delete these boilerplate comments

sources/user-data-providers/local-defaults/src/lib.rs
```diff
@@ -0,0 +1,15 @@
...
+
+const LOCAL_DEFAULTS_FILE: &str = "/local/user-data-defaults.toml";
+
+pub struct LocalDefaults;
+
+#[async_trait]
+impl UserDataProvider for LocalDefaults {
+    async fn user_data(&self) -> Result<Option<SettingsJson>, Box<dyn std::error::Error>> {
+        user_data_from_file(LOCAL_DEFAULTS_FILE)
+    }
```
bcressey: What do you think about having `AsyncUserDataProvider` and `UserDataProvider` traits instead? Or even removing the async trait and forcing providers to provide a non-async interface?

It's just the IMDS provider that needs to be async. Most of the other providers are very much "sync" - they use `std::fs` for I/O, etc.

It'd be nice to not pay for tokio as a dependency for every provider.

sources/api/user-data-provider/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+serde_plain = "1"
+serde-xml-rs = "0.6"
+snafu = "0.7"
+tokio = { version = "~1.32", default-features = false, features = ["process", "macros", "rt-multi-thread"] }  # LTS
+toml = "0.5"
+walkdir = "2.4"
+
+[target.'cfg(target_arch = "x86_64")'.dependencies]
+# vmw_backdoor includes x86_64 assembly, prevent it from building for ARM
+vmw_backdoor = "0.2"
```
bcressey: We don't need this as a dependency here, right? Just in the guestinfo provider?

sources/api/user-data-provider/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+log = "0.4"
+retry-read = { path = "../../retry-read", version = "0.1" }
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+serde_plain = "1"
+serde-xml-rs = "0.6"
+snafu = "0.7"
+tokio = { version = "~1.32", default-features = false, features = ["process", "macros", "rt-multi-thread"] }  # LTS
+toml = "0.5"
+walkdir = "2.4"
```
bcressey: Can we clean up some of these, like `imdsclient` and `serde-xml-rs`?

packages/early-boot-config/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
```
bcressey: To make rebuild tracking work properly, I'd like us to move all of `early-boot-config` out of `api` - Sean's first option. 

Something like this:
```
# main CLI
sources/early-boot-config/early-boot-config

# interface trait
sources/early-boot-config/early-boot-config-provider

# provider implementations
sources/early-boot-config/user-data-providers/...
```

packages/early-boot-config/early-boot-config.spec
```diff
@@ -0,0 +1,152 @@
...
+
+BuildRequires: %{_cross_os}glibc-devel
+
+%description
+%{summary}.
+
+%package -n %{_cross_os}early-boot-config-common
+Summary: early-boot-config binary
+%description -n %{_cross_os}early-boot-config-common
+%{summary}.
```
bcressey: nit: I like to separate the description from the definition:
```suggestion
%package -n %{_cross_os}early-boot-config-common
Summary: early-boot-config binary

%description -n %{_cross_os}early-boot-config-common
%{summary}.
```

(Essentially my style rule is "have an empty line before any of the built-in rpm sections".)

packages/early-boot-config/early-boot-config.spec
```diff
@@ -0,0 +1,152 @@
...
+
+%package -n %{_cross_os}early-boot-config-data-providers
+Summary: early-boot-config data providers directory
+%description -n %{_cross_os}early-boot-config-data-providers
+%{summary}.
```
bcressey: This subpackage can be merged into the common one.

packages/early-boot-config/early-boot-config.spec
```diff
@@ -0,0 +1,152 @@
...
+URL: https://github.com/bottlerocket-os/bottlerocket
+
+Source100: early-boot-config.service
+
+BuildRequires: %{_cross_os}glibc-devel
+
+%description
+%{summary}.
+
+%package -n %{_cross_os}early-boot-config-common
```
bcressey: If every variant recipe is referring to a subpackage, like `early-boot-config-vmware` or `early-boot-config-aws`, then it seems better to use the main, top-level package as the "common"package.

packages/early-boot-config/early-boot-config.spec
```diff
@@ -0,0 +1,152 @@
...
+%package -n %{_cross_os}early-boot-config-local
+Summary: local-provider
+Requires: %{_cross_os}early-boot-config-data-providers
+%description -n %{_cross_os}early-boot-config-local
+%{summary}.
+
+%package -n %{_cross_os}early-boot-config-aws
+Summary: early-boot-config package for AWS
+Provides: %{_cross_os}early-boot-config
+Requires: %{_cross_os}early-boot-config-common
```
bcressey: If you take my suggestion to factor the "common" subpackage back into the top-level one:
```suggestion
Requires: %{name}
```

packages/early-boot-config/early-boot-config.spec
```diff
@@ -0,0 +1,152 @@
...
+%{_cross_libexecdir}/early-boot-config/bin/ec2-identity-doc-user-data-provider
+%{_cross_libexecdir}/early-boot-config/bin/ec2-imds-user-data-provider
+
+%ifarch x86_64
+%files -n %{_cross_os}early-boot-config-vmware
+%{_cross_libexecdir}/early-boot-config/bin/vmware-cd-rom-user-data-provider
+%{_cross_libexecdir}/early-boot-config/bin/vmware-guestinfo-user-data-provider
+%endif
+
+%files -n %{_cross_os}early-boot-config-metal
```
bcressey, line 155: nit: can add a comment to clarify that this is intentional:
```suggestion
# There are no metal-specific providers, just dependencies like the local file providers.
%files -n %{_cross_os}early-boot-config-metal
```

packages/early-boot-config/early-boot-config.spec
```diff
@@ -0,0 +1,152 @@
...
+%install
+install -d %{buildroot}%{_cross_bindir}
+install -p -m 0755 ${HOME}/.cache/%{__cargo_target}/release/early-boot-config %{buildroot}%{_cross_bindir}
+
+install -d %{buildroot}%{_cross_unitdir}
+install -p -m 0644 %{S:100} %{buildroot}%{_cross_unitdir}
+
+install -d %{buildroot}%{_cross_libexecdir}/early-boot-config/bin
+install -p -m 0755 \
+    ${HOME}/.cache/%{__cargo_target}/release/ec2-identity-doc-user-data-provider \
```
bcressey: Just before the install section, I'd add a macro like this:
```
%global cargo_outdir %{getenv:CARGO_TARGET_DIR}/%{__cargo_target}/release
```

Then you can refer to files like this:
```
%{cargo_outdir}/ec2-imds-user-data-provider
```

(Part of the "style" of RPM spec writing is finding places to condense repeated patterns with macros, because there is quite a lot of repetition and boilerplate when building and packaging software.)

packages/early-boot-config/early-boot-config.spec
```diff
@@ -0,0 +1,152 @@
...
+    ${HOME}/.cache/%{__cargo_target}/release/vmware-cd-rom-user-data-provider \
+    ${HOME}/.cache/%{__cargo_target}/release/vmware-guestinfo-user-data-provider \
+    %{buildroot}%{_cross_libexecdir}/early-boot-config/bin
+%endif
+
+install -d %{buildroot}%{_cross_datadir}/early-boot-config/data-providers.d
+
+%post -n %{_cross_os}early-boot-config-aws -p <lua>
+posix.symlink("../../../libexec/early-boot-config/bin/ec2-identity-doc-user-data-provider", "%{_cross_datadir}/early-boot-config/data-providers.d/30-ec2-identity-doc")
+posix.symlink("../../../libexec/early-boot-config/bin/ec2-imds-user-data-provider", "%{_cross_datadir}/early-boot-config/data-providers.d/40-ec2-imds")
```
bcressey: There's no conflict between any of these symlinks from what I can see, so I would skip the `%post` scriptlets and just create and package the symlinks:
```
# before %install
# convention is not to include %{buildroot} - then you can reuse these macros in the %files sections.

%global early_boot_config_bindir %{_cross_libexecdir}/early-boot-config/bin
%global early_boot_config_provider_dir %{_cross_datadir}/early-boot-config/data-providers.d

# instead of %post
# (`ln -r` will take care of the relative symlink for you.)

install -d %{buildroot}%{early_boot_config_provider_dir}
ln -rs \
  %{buildroot}%{early_boot_config_bindir}/ec2-identity-doc-user-data-provider 
  %{buildroot}%{early_boot_config_provider_dir}/40-ec2-imds
```

sources/api/early-boot-config/src/main.rs
```diff
@@ -31,6 +30,8 @@ use crate::provider::{Platform, PlatformDataProvider};
 // ConditionPathExists that will prevent it from running again if this file exists.
 // We create it after running successfully.
 const MARKER_FILE: &str = "/var/lib/bottlerocket/early-boot-config.ran";
+/// The directory containing user data provider binaries
+const PROVIDERS_DIR: &str = "/usr/share/early-boot-config/data-providers.d";
```
bcressey: This will affect the packaging, but I'd recommend this for binaries:
```suggestion
/// The directory containing user data provider binaries
const PROVIDERS_DIR: &str = "/usr/libexec/early-boot-config/data-providers.d";
```

It's slightly weird to put binaries in `/usr/share` - the [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrshareArchitectureindependentData) says that's for "architecture-independent data".

(This wouldn't Bottlerocket's only crime against the FHS, but still.)

(Need to fix the docstring if you update this.)

packages/early-boot-config/early-boot-config.spec
```diff
@@ -0,0 +1,152 @@
...
+
+%post -n %{_cross_os}early-boot-config-local -p <lua>
+posix.symlink("../../../libexec/early-boot-config/bin/local-defaults-user-data-provider", "%{_cross_datadir}/early-boot-config/data-providers.d/10-local-defaults")
+posix.symlink("../../../libexec/early-boot-config/bin/local-file-user-data-provider", "%{_cross_datadir}/early-boot-config/data-providers.d/20-local-user-data")
+posix.symlink("../../../libexec/early-boot-config/bin/local-overrides-user-data-provider", "%{_cross_datadir}/early-boot-config/data-providers.d/50-local-overrides")
```
bcressey: `local-overrides` should probably be `99-` - it's very important that it comes last or it won't have the "override" behavior.

sources/api/early-boot-config/src/main.rs
```diff
@@ -94,48 +95,144 @@ fn parse_args(args: env::Args) -> Args {
...
+    info!("early-boot-config started");
+
+    info!("Gathering user data providers");
+    let mut threads = Vec::new();
+    let providers = gather_providers()?;
+    for provider in providers {
+        threads.push((
+            provider.clone(),
+            tokio::spawn(async move { run_provider(&args.log_level, &provider).await }),
+        ));
```
bcressey: Might be better to use [FuturesOrdered](https://docs.rs/futures/latest/futures/stream/struct.FuturesOrdered.html) instead:
```
let mut s = stream::FuturesOrdered::new();
let providers = gather_providers()?;
for provider in providers {
    s.push_back(...);
}

while Some(result) = s.next().await? {
    // ...
}
```

sources/user-data-providers/local-defaults/src/lib.rs
sam-berning: > What do you think about having `AsyncUserDataProvider` and `UserDataProvider` traits instead? Or even removing the async trait and forcing providers to provide a non-async interface?

I think I'll play around with the non-async interface. It really is just for IMDS so it's not worth forcing the others to use tokio. The implementation of `ec2-imds-user-data-provider` will be quite different from the others but I think it's worth it

sources/early-boot-config/user-data-providers/vmware-guestinfo/README.md
```diff
@@ -0,0 +1,9 @@
+# vmware-guestinfo-user-data-provider
+
+Current version: 0.1.0
+
+
```
webern: Now you need to add a doc comment at the top of each of these `main.rs` and/or `lib.rs` files to make sure something gets populated into the README!

sources/early-boot-config/user-data-providers/ec2-identity-doc/README.md
```diff
@@ -0,0 +1,9 @@
+# ec2-identity-doc-user-data-provider
+
+Current version: 0.1.0
+
+
```
webern: No info :(

sources/early-boot-config/user-data-providers/ec2-imds/Cargo.toml
```diff
@@ -5,13 +5,17 @@ authors = ["Zac Mrowicki <mrowicki@amazon.com>", "Sam Berning <bernings@amazon.c
 edition = "2021"
 license = "Apache-2.0 OR MIT"
 publish = false
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+build = "build.rs"
```
webern: Not necessary when following the naming convention.

```suggestion
```

sources/early-boot-config/user-data-providers/ec2-imds/README.md
```diff
@@ -0,0 +1,9 @@
+# ec2-imds-user-data-provider
+
+Current version: 0.1.0
+
```
webern, line 4: No info :(

packages/early-boot-config/early-boot-config.spec
```diff
@@ -87,66 +73,91 @@ Requires: %{_cross_os}early-boot-config-local
     -p vmware-guestinfo-user-data-provider
 %endif
 
+%global cargo_outdir %{getenv:HOME}/.cache/%{__cargo_target}/release
```
webern, line 68: Is there no macro for this? Should we rely on it being in the HOME directory?
An alternative might be to build with `cargo install` and specify the `--root /where/i/want/it`.

packages/early-boot-config/early-boot-config.spec
```diff
@@ -87,66 +73,91 @@ Requires: %{_cross_os}early-boot-config-local
...
+  %{buildroot}%{early_boot_config_bindir}/local-overrides-user-data-provider \
+  %{buildroot}%{early_boot_config_provider_dir}/99-local-overrides
 
 %ifarch x86_64
-%post -n %{_cross_os}early-boot-config-vmware -p <lua>
-posix.symlink("../../../libexec/early-boot-config/bin/vmware-cd-rom-user-data-provider", "%{_cross_datadir}/early-boot-config/data-providers.d/30-vmware-cd-rom")
-posix.symlink("../../../libexec/early-boot-config/bin/vmware-guestinfo-user-data-provider", "%{_cross_datadir}/early-boot-config/data-providers.d/40-vmware-guestinfo")
+ln -rs \
+  %{buildroot}%{early_boot_config_bindir}/vmware-cd-rom-user-data-provider \
+  %{buildroot}%{early_boot_config_provider_dir}/30-vmware-cd-rom
```
webern, line 120: Are all of these symlinks always going to be there, even on platforms where the particular user-data-provider is not relevant? I don't follow how we choose the right one for the variant.

sources/api/early-boot-config/src/main.rs
sam-berning: Been trying a few approaches to use this, but it keeps getting messy because we need the provider name for logging and errors when we join the threads. I tried having the thread return `(provider_name, result)` as well as creating a custom struct that impls `Future`, but I think both ended up less readable than the `Vec<(provider, JoinHandle)>` approach -- the return from `.next()` always needed to be converted into multiple other types, and

```
while let Some(Ok(provider, result)) = s.next().await {
...
}
```

doesn't handle errors correctly.

I think we should stick with the current solution unless you feel there's a compelling performance reason to use `FuturesOrdered`.

variants/vmware-k8s-1.28/Cargo.toml
```diff
@@ -32,6 +32,7 @@ kernel-parameters = [
 ]
 included-packages = [
     # core
+    "early-boot-config-vmware",
```
webern, line 35: @bcressey Your potential solution might change this, but is this the correct approach for now?

sources/early-boot-config/early-boot-config-provider/src/settings.rs
```diff
@@ -1,14 +1,14 @@
 //! The settings module owns the `SettingsJson` struct which contains the JSON settings data being
 //! sent to the API.
 
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use snafu::{OptionExt, ResultExt};
 
 /// SettingsJson represents a change that a provider would like to make in the API.
-#[derive(Debug)]
```
webern, line 8: Just to help me understand, how was this used before if it did not derive `Serialize` or `Deserialize`?

webern (commented):
Re-testing is going to be some work, but this is looking good to me.

variants/vmware-k8s-1.28/Cargo.toml
bcressey: > @bcressey Your potential solution might change this, but is this the correct approach for now?

Yes, this is the best way for now.

packages/early-boot-config/early-boot-config.spec
bcressey: As long as we're using `%cargo_build` then this will be the right path, since we set the target directory in the environment.

I opened https://github.com/bottlerocket-os/bottlerocket-sdk/issues/159 to add standard versions of these macros.

packages/early-boot-config/early-boot-config.spec
bcressey: > Are all of these symlinks always going to be there, even on platforms where the particular user-data-provider is not relevant? 

The symlinks are all created in the buildroot, but are only included in the specific subpackages with the corresponding provider.

packages/early-boot-config/early-boot-config.spec
```diff
@@ -0,0 +1,163 @@
...
+    -p local-defaults-user-data-provider \
+    -p local-file-user-data-provider \
+    -p local-overrides-user-data-provider
+
+%ifarch x86_64
+# build vmware user data providers
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p vmware-cd-rom-user-data-provider \
+    -p vmware-guestinfo-user-data-provider
+%endif
```
bcressey: I'd expect the build to be faster if we run them all in parallel -
```suggestion
%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
    -p early-boot-config \
    -p ec2-identity-doc-user-data-provider \
    -p ec2-imds-user-data-provider \
    -p local-defaults-user-data-provider \
    -p local-file-user-data-provider \
    -p local-overrides-user-data-provider \
%ifarch x86_64
    -p vmware-cd-rom-user-data-provider \
    -p vmware-guestinfo-user-data-provider \
%endif
  %{nil}
```

sources/early-boot-config/user-data-providers/vmware-guestinfo/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+[dependencies]
+async-trait = "0.1"
+base64 = "0.13"
+log = "0.4"
+serde = { version = "1", features = ["derive"] }
+serde_plain = "1"
+snafu = "0.7"
+tokio = { version = "~1.32", default-features = false, features = ["process", "macros", "rt-multi-thread"] }  # LTS
+early-boot-config-provider = { path = "../../early-boot-config-provider", version = "0.1" }
+vmw_backdoor = "0.2"
```
bcressey, line 21: I expect we have to do this so that interactions with the sources tree on an aarch64 build host, like `cargo clippy`, will not fail:

```suggestion
[target.'cfg(target_arch = "x86_64")'.dependencies]
# vmw_backdoor includes x86_64 assembly, prevent it from building for ARM
vmw_backdoor = "0.2"
```

sources/early-boot-config/user-data-providers/vmware-guestinfo/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    /// Request a key's value from guestinfo
+    fn backdoor_get_bytes(key: &str) -> Result<Option<Vec<u8>>> {
+        // Probe and access the VMware backdoor.  `kernel lockdown(7)` may block "privileged"
+        // mode because of its use of `iopl()`; the 5.15 kernels have it disabled regardless
+        // of lockdown mode. If this fails, fall back to "unprivileged" access without first
+        // requesting access to the relevant IO ports. KVM and VMware both have them special-
+        // cased in their emulation to not raise an exception to the guest OS and things
+        // should work out.
+        let mut backdoor = vmw_backdoor::probe_backdoor_privileged()
```
bcressey: Need to conditionalize this so that the `vmw_backdoor` path isn't taken when building on an aarch64 host.

sources/early-boot-config/early-boot-config/src/main.rs
```diff
@@ -0,0 +1,326 @@
+/*!
+# Introduction
+
+early-boot-config sends user data to the Bottlerocket API.
+
+Variants include their required user data provider binaries via packages.  early-boot-config discovers these binaries at runtime in /usr/libexec/early-boot-config/data-providers.d and runs them in order, sending any user data found to the API.
```
bcressey:
```suggestion
Variants include their required user data provider binaries via packages.  early-boot-config discovers these binaries at runtime in `/usr/libexec/early-boot-config/data-providers.d` and runs them in order, sending any user data found to the API.
```

sources/early-boot-config/early-boot-config/src/main.rs
```diff
@@ -0,0 +1,326 @@
...
+use snafu::{ensure, ResultExt};
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::str::{self, FromStr};
+use std::{env, io, process};
+use tokio::process::Command as AsyncCommand;
+use walkdir::WalkDir;
+
+// TODO
+// Tests!
```
bcressey: not that it doesn't need tests, but the TODO comment isn't motivating that
```suggestion
```

bcressey (commented):
Reviewed it all - nice! Everything LGTM except the async trait for `UserDataProvider` which I'd still like to make sync. If nothing else so the providers are smaller on disk and build faster.

sources/api/early-boot-config/src/main.rs
bcressey: > I think we should stick with the current solution unless you feel there's a compelling performance reason to use FuturesOrdered.

Sounds good!

sources/user-data-providers/local-defaults/src/lib.rs
bcressey: > I think I'll play around with the non-async interface.

Did you end up trying this? How'd it go?

sources/early-boot-config/early-boot-config-provider/src/settings.rs
sam-berning: when providers existed in the same crate as `early-boot-config`, they could just return this struct, with the settings changes pre-serialized in the `json` field, which at the time was a String. we only need `Serialize` and `Deserialize` now because we're using this as the return format of a provider binary

sources/early-boot-config/user-data-providers/vmware-guestinfo/README.md
sam-berning: oh duh 🤦‍♂️ will add

sam-berning:
^ addressed comments

sources/user-data-providers/local-defaults/src/lib.rs
sam-berning: I couldn't figure out how to make it work with the EC2 providers so I ended up just creating `AsyncUserDataProvider` and `UserDataProvider` as separate traits.

webern (approved):

sources/early-boot-config/user-data-providers/vmware-cd-rom/README.md
```diff
@@ -0,0 +1,11 @@
+# vmware-cd-rom-user-data-provider
+
+Current version: 0.1.0
+
+## Introduction
+
+User data provider binary that fetches user data provided via CD-ROM to a VMWare VM.
```
bcressey: need to globally change `VMWare` to `VMware` - `git grep VMWare` should not find any matches.

sources/early-boot-config/user-data-providers/vmware-guestinfo/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+use std::io::{Cursor, Read};
+
+// The fields in which user data and its encoding are stored in guestinfo
+const GUESTINFO_USERDATA: &str = "guestinfo.userdata";
+const GUESTINFO_USERDATA_ENCODING: &str = "guestinfo.userdata.encoding";
+
+pub struct VmwareGuestinfo;
+
+#[cfg(target_arch = "x86_64")]
+impl VmwareGuestinfo {
```
bcressey: This still triggers [clippy failures](https://gist.github.com/bcressey/7997b8eb11d346cf0ea0e0e0faa62c56) on an aarch64 host.

Maybe try moving `lib.rs` to `x86_64.rs`, and having this become `lib.rs`:
```
 #[cfg(target_arch = "x86_64")]
mod x86_64;

#[cfg(not(target_arch = "x86_64"))]
mod not_x86_64;

#[cfg(target_arch = "x86_64")]
pub use x86_64::*;

#[cfg(not(target_arch = "x86_64"))]
pub use not_x86_64::*;
```

With `not_x86_64.rs`:
```
use early_boot_config_provider::settings::SettingsJson;
use early_boot_config_provider::provider::UserDataProvider;

pub struct VmwareGuestinfo;

impl UserDataProvider for VmwareGuestinfo {
    #[allow(dead_code)]
    fn user_data(&self) -> std::result::Result<Option<SettingsJson>, Box<dyn std::error::Error>> {
        unimplemented!()
    }
}
```

This seems to correctly no-op stuff on aarch64, but I haven't tested on x86_64. Then you can get rid of the `#[cfg(target_arch = "x86_64")]` conditionals in this file.

bcressey (approved):
Whew! This looks really good. I have two small concerns I'd like to see fixed, but they shouldn't affect any testing done.

sam-berning:
^ fixed merge conflicts and addressed @bcressey's comments.

Tested all sets of user data providers by building `aws-dev`, `vmware-dev`, and `metal-dev`. For each variant, I injected local files with a custom `rpm2img` script. Verified that the user data looked as expected on all variants, and checked the `journalctl` logs to see that all providers picked up the provided user data and applied it in the correct order.

Testing report:

### metal:
```
bash-5.1# journalctl | grep early-boot-config
Apr 03 23:00:23 localhost early-boot-config[1219]: [2024-04-03T23:00:23Z INFO  early_boot_config] early-boot-config started
Apr 03 23:00:23 localhost early-boot-config[1219]: [2024-04-03T23:00:23Z INFO  early_boot_config] Gathering user data providers
Apr 03 23:00:23 localhost early-boot-config[1219]: [2024-04-03T23:00:23Z INFO  early_boot_config] Provider '10-local-defaults': [2024-04-03T23:00:23Z INFO  early_boot_config_provider::provider] '/local/user-data-defaults.toml' exists, using it
Apr 03 23:00:23 localhost early-boot-config[1219]: [2024-04-03T23:00:23Z INFO  early_boot_config] Found user data via user data from /local/user-data-defaults.toml, sending to API
Apr 03 23:00:23 localhost early-boot-config[1219]: [2024-04-03T23:00:23Z INFO  early_boot_config] Provider '20-local-user-data': [2024-04-03T23:00:23Z INFO  early_boot_config_provider::provider] '/var/lib/bottlerocket/user-data.toml' exists, using it
Apr 03 23:00:23 localhost early-boot-config[1219]: [2024-04-03T23:00:23Z INFO  early_boot_config] Found user data via user data from /var/lib/bottlerocket/user-data.toml, sending to API
Apr 03 23:00:23 localhost early-boot-config[1219]: [2024-04-03T23:00:23Z INFO  early_boot_config] Provider '99-local-overrides': [2024-04-03T23:00:23Z INFO  early_boot_config_provider::provider] '/local/user-data-overrides.toml' exists, using it
Apr 03 23:00:23 localhost early-boot-config[1219]: [2024-04-03T23:00:23Z INFO  early_boot_config] Found user data via user data from /local/user-data-overrides.toml, sending to API
```

### VMware:
```
bash-5.1# journalctl | grep early-boot-config
Apr 04 00:18:16 localhost early-boot-config[1690]: [2024-04-04T00:18:16Z INFO  early_boot_config] early-boot-config started
Apr 04 00:18:16 localhost early-boot-config[1690]: [2024-04-04T00:18:16Z INFO  early_boot_config] Gathering user data providers
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Provider '10-local-defaults': [2024-04-04T00:18:17Z INFO  early_boot_config_provider::provider] '/local/user-data-defaults.toml' exists, using it
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Found user data via user data from /local/user-data-defaults.toml, sending to API
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Provider '20-local-user-data': [2024-04-04T00:18:17Z INFO  early_boot_config_provider::provider] '/var/lib/bottlerocket/user-data.toml' exists, using it
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Found user data via user data from /var/lib/bottlerocket/user-data.toml, sending to API
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Provider '30-vmware-cd-rom': [2024-04-04T00:18:17Z INFO  vmware_cd_rom_user_data_provider] Attempting to retrieve user data from mounted CD-ROM
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Provider '30-vmware-cd-rom': [2024-04-04T00:18:17Z INFO  vmware_cd_rom_user_data_provider] '/media/cdrom/user-data' exists, using it
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Found user data via CD-ROM, sending to API
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Provider '40-vmware-guestinfo': [2024-04-04T00:18:17Z INFO  vmware_guestinfo_user_data_provider::x86_64] Attempting to retrieve user data via guestinfo interface
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Provider '40-vmware-guestinfo': [2024-04-04T00:18:17Z INFO  vmware_guestinfo_user_data_provider::x86_64] Found user data encoding: base64
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Provider '40-vmware-guestinfo': [2024-04-04T00:18:17Z INFO  vmware_guestinfo_user_data_provider::x86_64] Decoding user data
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Found user data via guestinfo, sending to API
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Provider '99-local-overrides': [2024-04-04T00:18:17Z INFO  early_boot_config_provider::provider] '/local/user-data-overrides.toml' exists, using it
Apr 04 00:18:17 localhost early-boot-config[1690]: [2024-04-04T00:18:17Z INFO  early_boot_config] Found user data via user data from /local/user-data-overrides.toml, sending to API
```

### AWS:
```
bash-5.1#  journalctl | grep early-boot-config
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] early-boot-config started
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Gathering user data providers
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Provider '10-local-defaults': [2024-04-03T22:30:51Z INFO  early_boot_config_provider::provider] '/local/user-data-defaults.toml' exists, using it
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Found user data via user data from /local/user-data-defaults.toml, sending to API
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Provider '20-local-user-data': [2024-04-03T22:30:51Z INFO  early_boot_config_provider::provider] '/var/lib/bottlerocket/user-data.toml' exists, using it
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Found user data via user data from /var/lib/bottlerocket/user-data.toml, sending to API
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Provider '30-ec2-identity-doc': [2024-04-03T22:30:51Z INFO  ec2_identity_doc_user_data_provider] Using IMDS for region
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Provider '30-ec2-identity-doc': [2024-04-03T22:30:51Z INFO  imdsclient] Received dynamic/instance-identity/document
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Found user data via EC2 instance identity document, sending to API
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Provider '40-ec2-imds': [2024-04-03T22:30:51Z INFO  ec2_imds_user_data_provider] Fetching user data from IMDS
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Provider '40-ec2-imds': [2024-04-03T22:30:51Z INFO  imdsclient] Received user-data
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Found user data via EC2 IMDS, sending to API
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Provider '99-local-overrides': [2024-04-03T22:30:51Z INFO  early_boot_config_provider::provider] '/local/user-data-overrides.toml' exists, using it
Apr 03 22:30:51 localhost early-boot-config[1234]: [2024-04-03T22:30:51Z INFO  early_boot_config] Found user data via user data from /local/user-data-overrides.toml, sending to API
```

bcressey (approved):
Very nice!

sam-berning:
^ rebased once more

webern (approved):
Nice!

PR #3639, opened 2023-12-05
pluto: handle auth info in https_proxy

### LINKED ISSUE #3525: pluto doesn't handle proxy credentials in `https_proxy`
We've been happily running on Kubernetes 1.25 for several months using bottlerocket images but we are now in the process of upgrading our clusters. However, after a suuccessful upgrade of the EKS managed control plane to 1.26, new nodes are failing to join the node goup.

This seems similar to  [3064](https://github.com/bottlerocket-os/bottlerocket/issues/3064) to which my colleague Paul asked a question on -a couple of days ago  but not quite the same problem.

Any advice on what the issue may be and how we can fix it please?

**Image I'm using:**
bottlerocket-aws-k8s-1.26-x86_64-v1.15.1-264e294c

Instances launch and join the EKS nodegroup

**What actually happened:**
Instances launch but fail to join the nodegroup, The System log shows the following error:

[  OK  ] Finished Bottlerocket userdata configuration system.
         Starting User-specified setting generators...
         Unmounting EFI System Partition Automount...
[  OK  ] Unmounted EFI System Partition Automount.
[  303.392308] sundog[1343]: Setting generator 'pluto private-dns-name' failed with exit code 1 - stderr: Timed out retrieving private DNS name from EC2: deadline has elapsed
[FAILED] Failed to start User-specified setting generators

We do also see this as well in the console log but not sure if this is relevant to our issue - but it is different to what we see on a 1.25 control plane cluster:

[  OK  ] Finished Generate network configuration.
[    1.708363] migrator[1266]: Data store does not exist at given path, exiting (/var/lib/bottlerocket/datastore/current)
[  OK  ] Finished Bottlerocket data store migrator.

Our VPC has two CIDRs, a secondary CIDR is used for our pod network

etungsten:
**Issue number:**
Resolves https://github.com/bottlerocket-os/bottlerocket/issues/3525

**Description of changes:**
```
    pluto: handle auth info in https_proxy
    
    Need to parse out auth info in `https_proxy` and set to proxy header.
```

**Testing done:**
Unit tests pass.

Set up my own tiny proxy host and launched new Bottlerocket host with the following settings:
```toml
[settings.network]
https-proxy = "http://user:pass@my_proxy:9898"
```

The boot succeeds, pluto successfully fetches instance information through AWS EC2 API.
In cloudtrail, i can see the source of the DescribeInstance event as being my proxy:
```
{
    "userIdentity": {
        "type": "AssumedRole",
        "principalId": ".....i-0d2d921de0b0a9aad",
        "arn": "..../i-0d2d921de0b0a9aad",
...
    "eventTime": "2023-12-05T23:51:56Z",
    "eventSource": "ec2.amazonaws.com",
    "eventName": "DescribeInstances",
    "awsRegion": "us-west-2",
    "sourceIPAddress": "<my_proxy_url>",
    "userAgent": "aws-sdk-rust/0.55.3 os/linux lang/rust/1.73.0",
```

**Terms of contribution:**

### DIFF
--- a/sources/api/pluto/src/hyper_proxy/mod.rs
+++ b/sources/api/pluto/src/hyper_proxy/mod.rs
@@ -5,10 +5,10 @@
 
 mod stream;
 mod tunnel;
+use futures_util::future::TryFutureExt;
+use headers::{authorization::Credentials, Authorization, HeaderMapExt, ProxyAuthorization};
 use http::header::HeaderMap;
 use hyper::{service::Service, Uri};
-
-use futures_util::future::TryFutureExt;
 use std::{fmt, io, sync::Arc};
 use std::{
     future::Future,
@@ -27,7 +27,12 @@ type BoxError = Box<dyn std::error::Error + Send + Sync>;
 
 /// The Intercept enum to filter connections
 #[derive(Debug, Clone)]
+#[allow(dead_code)]
 pub enum Intercept {
+    /// Only https connections will go through proxy
+    Https,
+    /// No connection will go through this proxy
+    None,
     /// A custom intercept
     Custom(Custom),
 }
@@ -84,7 +89,9 @@ impl Intercept {
     /// A function to check if given `Uri` is proxied
     pub fn matches<D: Dst>(&self, uri: &D) -> bool {
         match (self, uri.scheme()) {
+            (&Intercept::Https, Some("https")) => true,
             (&Intercept::Custom(Custom(ref f)), _) => f(uri.scheme(), uri.host(), uri.port()),
+            _ => false,
         }
     }
 }
@@ -114,6 +121,20 @@ impl Proxy {
             uri,
             headers: HeaderMap::new(),
             force_connect: false,
+        }
+    }
+
+    /// Set `Proxy` authorization
+    pub fn set_authorization<C: Credentials + Clone>(&mut self, credentials: Authorization<C>) {
+        match self.intercept {
+            Intercept::Https => {
+                self.headers.typed_insert(ProxyAuthorization(credentials.0));
+            }
+            _ => {
+                self.headers
+                    .typed_insert(Authorization(credentials.0.clone()));
+                self.headers.typed_insert(ProxyAuthorization(credentials.0));
+            }
         }
     }
 }
--- a/sources/api/pluto/src/proxy.rs
+++ b/sources/api/pluto/src/proxy.rs
@@ -1,8 +1,10 @@
 use crate::hyper_proxy::{Proxy, ProxyConnector};
+use headers::Authorization;
 use hyper::Uri;
 use hyper_rustls::HttpsConnectorBuilder;
 use snafu::{ResultExt, Snafu};
 use std::env;
+use url::Url;
 
 #[derive(Debug, Snafu)]
 pub(super) enum Error {
@@ -10,6 +12,12 @@ pub(super) enum Error {
     UriParse {
         input: String,
         source: hyper::http::uri::InvalidUri,
+    },
+
+    #[snafu(display("Unable to parse '{}' as URL: {}", input, source))]
+    UrlParse {
+        input: String,
+        source: url::ParseError,
     },
 
     #[snafu(display("Failed to create proxy creator: {}", source))]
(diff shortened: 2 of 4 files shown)


jpculp (approved):

sources/api/pluto/src/proxy.rs
```diff
@@ -33,6 +34,42 @@ pub(crate) fn fetch_proxy_env() -> (Option<String>, Option<String>) {
...
+    password: Option<String>,
+}
+
+// Check for presence of proxy auth info to add to header, reformat https_proxy to be proxy URL
+fn parse_proxy_info(https_proxy: &str) -> ProxyInfo {
+    match https_proxy.rsplit_once('@') {
+        Some((p, s)) => {
+            // Check for scheme
+            if let Some((scheme, auth_info)) = p.split_once("://") {
+                let userpass = auth_info.split_once(':');
```
bcressey: Is there a URL parsing crate we should use here instead? What if the username or password contains `:` or `@`? I assume this would be invalid (not URL encoded), but the type (`&str`) doesn't guarantee it.

sources/api/pluto/src/hyper_proxy/mod.rs
```diff
@@ -27,7 +27,16 @@ type BoxError = Box<dyn std::error::Error + Send + Sync>;
 
 /// The Intercept enum to filter connections
 #[derive(Debug, Clone)]
+#[allow(dead_code)]
 pub enum Intercept {
+    /// All incoming connection will go through proxy
+    All,
+    /// Only http connections will go through proxy
+    Http,
+    /// Only https connections will go through proxy
```
bcressey: Why do we need to support `All` or `Http` - we should only have `https` URLs.

etungsten:
Push above addresses @bcressey's comments.

Retested, same behavior as observed in cloudtrail:
```
    "userIdentity": {
        "type": "AssumedRole",
        "principalId": "...:i-03de3baa30b4d01f0",
        "arn": "...i-03de3baa30b4d01f0",
...
        }
    },
    "eventSource": "ec2.amazonaws.com",
    "eventName": "DescribeInstances",
    "awsRegion": "us-west-2",
    "sourceIPAddress": "my_proxy",
    "userAgent": "aws-sdk-rust/0.55.3 os/linux lang/rust/1.73.0",
```

jpculp (approved):

sources/api/pluto/src/hyper_proxy/mod.rs
```diff
@@ -116,6 +123,20 @@ impl Proxy {
...
+            Intercept::Https => {
+                self.headers.typed_insert(ProxyAuthorization(credentials.0));
+            }
+            _ => {
+                self.headers
+                    .typed_insert(Authorization(credentials.0.clone()));
+                self.headers.typed_insert(ProxyAuthorization(credentials.0));
+            }
+        }
+    }
```
bcressey, line 139: Why do we need to set the authorization headers in the `_` (specifically, `None`) case where connections aren't going through the proxy?

yeazelm (approved):
LGTM!

jpculp (approved):

sources/api/pluto/src/proxy.rs
```diff
@@ -75,7 +91,18 @@ pub(crate) fn setup_http_client(
...
-    let proxy = Proxy::new(intercept, proxy_uri);
+    let mut proxy = Proxy::new(intercept, proxy_uri);
+    // Parse https_proxy as URL to extract out auth information if any
+    let proxy_url = Url::parse(&https_proxy).context(UrlParseSnafu {
+        input: &https_proxy,
+    })?;
+    if !proxy_url.username().is_empty() {
+        proxy.set_authorization(Authorization::basic(
+            proxy_url.username(),
+            proxy_url.password().unwrap_or_default(),
```
cbgbt: I haven't done much with [`Proxy-Authorization`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization) but my read of the header documentation and my understanding of URL username/password encoding makes me think that it's valid to provide a passphrase without a username.

Should we be inserting the header if a password is given without a username?

etungsten:
Push above addresses @cbgbt 's comment

sources/api/pluto/src/proxy.rs
etungsten: I believe so. Good catch

yeazelm (approved):

webern (approved):

cbgbt (approved):

sources/api/pluto/src/hyper_proxy/mod.rs
etungsten: This was lifted from [`hyper_proxy` source](https://docs.rs/hyper-proxy/latest/src/hyper_proxy/lib.rs.html#196-210) and it's not super clear to me either. In any case, we shouldn't be reaching this path since we're only proxying HTTPS traffic.

etungsten: After some digging, we've determined that the Auth header here will never get sent if the proxy doesn't "match" with the destination URL. We'll go ahead and simplify this logic here so it only contains the case we care about, i.e. proxying HTTPS traffic.

PR #3673, opened 2023-12-15
selinux-policy: remove unused entry

yeazelm:
We don't invoke `runc` directly but rather through `host-ctr`, `containerd`, or `docker` so this line isn't doing anything. This cleans up the line to be more clear that we don't actually need this rule. In fact, it was pointing to a different location than where `runc` resides.

**Testing done:**
Built an aws-k8s-1.28 image and confirmed it could launch host containers and orchestrated containers.

**Terms of contribution:**


arnaldo2792:
~Have you tried upgrade/downgrade testing?~

I was pointed out that this path never existed

arnaldo2792 (approved):

webern (approved):

bcressey (approved):

PR #3700, opened 2024-01-06
netdog: remove conditional compilation

### LINKED ISSUE #3618: OOTB: Remove conditional compilation from `netdog`
Removing conditional compilation from `netdog` may require a bit more detailed design.  `netdog` **currently** needs to be aware of the network backend so it generates the proper configuration file format, and also needs to be aware of the platform (AWS/non-AWS) to understand how to generate the proper hostname.

Moving forward into an OOTB world, we do not want platform concepts to live at this layer of software.  Ideally these mechanisms would be abstracted away and `netdog` would not need to understand the mechanisms, just discover and drive them.

A few options are lined out below:

## Helper binaries (preferred option so far)
Similar design pattern to #3616 and settings extensions 
* The parts of the program that are currently conditionally compiled will become their own discrete binaries and will be packaged as such.
* `netdog` will run discover and run those binaries to generate network config, generate hostname, etc.
* Create a metapackage bundling the proper set of helpers for each variant, the variant can then pull in the bundled set of helpers from this metapackage.

**Pros**
* Follows an already established pattern as settings extensions and `early-boot-config`.
* Flexible, allowing for variants to choose applicable bits.
* Extensible, different mechanisms can be packaged and implemented without changes to core `netdog`.  For example, a different network backend could be supported through a new "config generation" binary.

**Cons**
* Additional binaries add complexity to packaging
* (Probably negligable) Additional binary size adds to total image size
* (Potentially negligable) Calling additional binaries could have time penalties at runtime

**Other options that may be less desireable:**

### Configuration file to direct behavior
* Package a configuration file which `netdog` will read at runtime to determine its behavior.
* Remove all conditional compilation from `netdog` and compile all code, even if it is not applicable.

**Pros**
* Changes to `netdog` aren't extensive, mostly deletions of `cfg` directives

**Cons**
* Some `netdog` subcommands aren't applicable depending on the network backend.
* Raises the complexity of creating an OOTB as the user is required to create these files, be aware of their structure, etc.
* Does not allow for extensibility of `netdog` without interfering with the core program.

### Multiple `netdog` packages
* Split `netdog` into multiple crates (`netdog-lib`, `netdog-wicked`, `netdog-networkd`).
* Create 2 packages `netdog-wicked`, `netdog-networkd`
* Variants will include the correct package for their needs

**Pros**
* Very deterministic behavior

**Cons**
* Not as extensible in the future, and future changes for OOTB may not fit neatly into a `wicked` or `networkd` package.

jmt-lab:
**Issue number:** 
Closes #3618

Removes conditional compilation of netdog for wicked and systemd networkd. This change moves our build system to build two variants of the netdog rpm: one for wicked and one for systemd-networkd. To facilitate this change the following was done:

* Migrate `#[cfg(net_backend = ]` to `#[cfg(feature = `
* Add two cargo features to sources/api/netdgo: 'wicked' and 'systemd-networkd'
* Extract netdog to its own package folder (packages/netdog)
* Extract netdog rpm building to netdog.spec
* Create two subpackages one netdog-wicked  and netdog-systemd-networkd  which are selected via image-features
* Migrated netdog to utilize the dogtag hostname detectors

**Testing done:**
Built and published aws-ecs-1 (wicked) and aws-ecs-2 (systemd-networkd) and verified network accessibility and use in a ecs cluster.

**Terms of contribution:**


packages/netdog/netdog.spec
```diff
@@ -0,0 +1,96 @@
...
+Requires: %{_cross_os}wicked
+%description -n %{_cross_os}netdog-wicked
+%{summary}.
+
+%prep
+%setup -T -c
+%cargo_prep
+
+%build
+mkdir bin
```
sam-berning, line 61: nit: is this directory used anywhere?

sources/api/netdog/Cargo.toml
```diff
@@ -1,13 +1,18 @@
 [package]
 name = "netdog"
-version = "0.1.0"
+version = "0.1.1"
```
cbgbt: We've historically not done much management of subcomponent versioning. I'm supportive of using meaningful versioning for these changes, especially for subcomponents that will get their own RPM.

packages/netdog/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+# RPM BuildRequires
+[build-dependencies]
+glibc = { path = "../glibc" }
```
cbgbt: We prefer to add newlines at EOF of text files per posix standards. This is configurable in most editors.

I'd suggest doing an interactive re-base to re-add missing newlines.

sources/api/netdog/Cargo.toml
```diff
@@ -9,7 +9,7 @@ publish = false
 exclude = ["README.md"]
 
 [features]
-default = ["systemd-networkd"]
+default = ["wicked"]
 wicked = []
```
cbgbt: Are the unit tests something we can fix here? I think advocating for `systemd-networkd` as the default makes sense.

sources/api/netdog/Cargo.toml
```diff
@@ -1,13 +1,18 @@
...
+version = "0.1.1"
 authors = ["Ben Cressey <bcressey@amazon.com>"]
 license = "Apache-2.0 OR MIT"
 edition = "2021"
 publish = false
 # Don't rebuild crate just because of changes to README.
 exclude = ["README.md"]
 
+[features]
+default = ["wicked"]
```
sam-berning: it feels wrong to me that wicked is the default feature flag, but we're using systemd-networkd as the default version of netdog otherwise.

I see the change was made to pass the unit tests, is it possible to rewrite the tests to either work with systemd-networkd or to specifically use the wicked features?

variants/aws-ecs-2/Cargo.toml
```diff
@@ -42,6 +43,7 @@ path = "../variants.rs"
 # core
 release = { path = "../../packages/release" }
 kernel-6_1 = { path = "../../packages/kernel-6.1" }
+netdog = {path = "../../packages/netdog" }
```
sam-berning: why is netdog added as a build dependency only on this variant?

variants/aws-ecs-1-nvidia/Cargo.toml
```diff
@@ -20,6 +20,7 @@ included-packages = [
 # core
     "release",
     "kernel-5.10",
+    "netdog-wicked",
```
sam-berning: checking my understanding, when a variant directly includes this subpackage, it overwrites the link created by `netdog-systemd-networkd`, right? are we sure that the packages are always installed in the correct order?

packages/netdog/Cargo.lock
```diff
@@ -0,0 +1,14 @@
+# This file is automatically @generated by Cargo.
```
cbgbt: I don't believe that we tend to commit lockfiles for individual packages.

packages/settings-motd/Cargo.lock
```diff
@@ -0,0 +1,14 @@
+# This file is automatically @generated by Cargo.
```
cbgbt: This similarly seems like it shouldn't be here.

packages/netdog/netdog.spec
jmt-lab: For the netdog binary itself i believe, needs to be created before hand

sources/api/netdog/Cargo.toml
jmt-lab: Good catch, I forgot to add the unit test change, I can modify the unit test to work with both, as systemd-networkd only adds extra lines (if i change the test to only verify that the first line is the expected it should pass on both and we can flip the default)

variants/aws-ecs-2/Cargo.toml
jmt-lab: Somehow this escaped my remove all

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,96 @@
...
+Source2: run-netdog.mount
+Source3: write-network-status.service
+Source4: netdog-tmpfiles.conf
+Source5: disable-udp-offload.service
+Source6: 00-resolved.conf
+Source7: generate-network-config.service
+
+BuildRequires: %{_cross_os}glibc-devel
+Requires: %{_cross_os}netdog
+%description
```
cbgbt: nit: mind adding whitespace between these?

variants/aws-ecs-1-nvidia/Cargo.toml
jmt-lab: Yes essentially, so netdog defines two subpackages:

netdog-systemd-networkd which provides netdog:1 (epoch 1)
netdog-wicked which provides netdog:0 (epoch 0)

So way to think of the above without it the pseudo dnf install resolution would go dnf install netdog (find the default latest epoch) this would auto resolve to netdog-systemd-networkd since it provides netdog with highest epoch

However if you manually say install netdog-wicked it will ensure that netdog is provided by netdog:0 or netdog-wicked

packages/os/Cargo.lock
```diff
@@ -0,0 +1,14 @@
+# This file is automatically @generated by Cargo.
```
cbgbt: Similar extraneous lockfile here.

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,96 @@
...
+%install
+install -d %{buildroot}%{_cross_bindir}
+install -d %{buildroot}%{_cross_tmpfilesdir}
+install -d %{buildroot}%{_cross_unitdir}
+install -d %{buildroot}%{_cross_libdir}
+install -p -m 0644 %{S:4} %{buildroot}%{_cross_tmpfilesdir}/netdog.conf
+install -p -m 0644 %{S:2} %{S:3} %{S:7} %{buildroot}%{_cross_unitdir}
+install -d %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
+install -p -m 0644 %{S:6} %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
+%if %{with vmware_platform}
```
cbgbt: I think we may want this file to be included in yet another package which we include in `vmware` variants. Right now this package still requires conditional compilation on vmware.

variants/aws-dev/Cargo.toml
```diff
@@ -28,6 +27,7 @@ included-packages = [
 # core
     "release",
     "kernel-6.1",
+
```
cbgbt: nit: I'd suggest pruning (or standardizing) the added whitespace, since it's not consistent between files.

packages/netdog/netdog.spec
```diff
@@ -90,7 +90,7 @@ install -p -m 0755 ${HOME}/.cache/wicked/%{__cargo_target}/release/netdog %{buil
 %endif
 
 %post -n %{_cross_os}netdog-wicked -p <lua>
-fs.link("%{_cross_bindir}/netdog-wicked", "%{_cross_bindir}/netdog")
+posix.link("%{_cross_bindir}/netdog-wicked", "%{_cross_bindir}/netdog")
```
cbgbt: stylistic nit that might appear strangely in the GitHub UI. I'd suggest rebasing this change to the one that introduces the new RPM, since it's otherwise irrelevant here.

packages/release/Cargo.lock
```diff
@@ -0,0 +1,570 @@
+# This file is automatically @generated by Cargo.
```
cbgbt: I wonder if these are being generated by an editor plugin like rust-analyzer? We may want to concoct a .gitignore for these entries.

cbgbt (commented):
Nice work! Please pardon any midair collisions with @sam-berning. I think we reviewed concurrently.

variants/aws-ecs-1-nvidia/Cargo.toml
sam-berning: got it, thanks for the explanation

packages/netdog/Cargo.toml
jmt-lab: You got it

sources/api/netdog/Cargo.toml
jmt-lab: will fix

packages/netdog/netdog.spec
jmt-lab: Sure

packages/netdog/netdog.spec
jmt-lab: LIke a third subpackage netdog-vmware which we then require in the vmware variants ontop of the normal

packages/release/Cargo.lock
jmt-lab: Thats exactly what is happening

packages/netdog/netdog.spec
cbgbt: After looking at the specific service, while it is networking related, I wouldn't claim that it's necessarily part of netdog.

I think what we can do for now is move this back into `os` and opt to include it in a variant-specific `vmware` package

jmt-lab: like adding a vmware-config subpackage in os?

sources/api/netdog/src/cli/mod.rs
```diff
@@ -228,7 +228,9 @@ mod tests {
         let fake_file = tempfile::NamedTempFile::new().unwrap();
         let expected = "-net.ipv4.conf.eno1.rp_filter = 2\n";
         write_interface_sysctl(interface, &fake_file).unwrap();
-        assert_eq!(std::fs::read_to_string(&fake_file).unwrap(), expected);
+        assert!(std::fs::read_to_string(&fake_file)
+            .unwrap()
+            .starts_with(expected));
```
yeazelm, line 233: This actually changes the semantics a bit no? We are checking for precisely the content listed above, did you need to make the `starts_with` change to get it to pass?

variants/aws-dev/Cargo.toml
```diff
@@ -12,7 +12,6 @@ grub-set-private-var = true
 unified-cgroup-hierarchy = true
 xfs-data-partition = true
 uefi-secure-boot = true
-systemd-networkd = true
```
yeazelm: So if I understand correctly, we are shifting the dependency to default into `systemd-networkd`? And we rely upon the netdog-wicked to bring in wicked.

packages/netdog/netdog.spec
yeazelm: Although it bloats the number of rpms a tad, I think I like the idea of a `netdog-vmware-config` rpm and possibly moving the rest of the shared config into its own rpm and only basically swapping the resulting binary and `Requires` directives for the resulting rpms. That would allow others to "bring their own" netdog binary without having to remember to find and collect all the right configuration files needed. This is more preference and can be persuaded either way on the approach. I don't know if we really expect anyone to ever mess with and provide their own netdog but I think its more about setting a pattern for how we group config files that might be shared among mutually exclusive binaries. If I'm hijacking this thread we can move it away from here too.

packages/.gitignore
```diff
@@ -1,3 +1,4 @@
 *.patch.bz2
 *.src.rpm
 *.zip
+**/Cargo.lock
```
yeazelm: Was this intentional? I thought we actually did want to track `Cargo.lock` files?

sources/api/netdog/src/cli/mod.rs
jmt-lab: systemd-networkd generates the correct line however, it also generates extra lines for ipv6 settings. Considering this test was looking for the first line i assumed keeping it the same would be preferable

variants/aws-dev/Cargo.toml
jmt-lab: correct

packages/.gitignore
jmt-lab: This was in response to a preivous comment where my editor kept adding .lock files where not needed, sean suggested maybe we should setup a gitignore just for this level of Cargo.lock files

sources/api/netdog/Cargo.toml
```diff
@@ -1,13 +1,18 @@
...
 license = "Apache-2.0 OR MIT"
 edition = "2021"
 publish = false
 # Don't rebuild crate just because of changes to README.
 exclude = ["README.md"]
 
+[features]
+default = ["systemd-networkd"]
+wicked = []
+systemd-networkd = []
```
webern: It is not a great idea to have mutually exclusive features. It would be better to have a `wicked` feature, and to use `cfg(wicked)` and `cfg(not(wicked))`, otherwise `--all-features` will not compile.

But even so, this doesn't actually solve the problem of removing conditional compilation. We are still compiling conditionally. I would prefer it if the Rust package contained `bin/netdog` and `bin/netdog-wicked` which can both be compiled at the same time.

jmt-lab: I went this approach to minimize the risk to the code and utilize standard rust/cargo approach for this. While it is true that generally you want to avoid mutually exclusive features, there is also not a valid case to compile with --all-features as a variant is not going to have BOTH wicked and systemd installed on it. Also the cfg(wicket) cfg(not(feature=wicked)) approach limits us to just these two and means if we add a third we'd have to readdress the feature flags entirely instead of adding a third feature. 

I do think I need to add a compile_error if a user tries to compile this with both feature flags, that is definitely a miss. I do understand where you are coming from on the not removing conditional compilation you are true the cargo crate still has conditional compilation. However, bottlerocket as a whole no longer does here as it compiles both netdog-systemd-networkd and netdog-wicked thus ensuring we build both versions.

jmt-lab:
Rolled back to handling the binary split in the rpm spec. It is more than sufficient and attempts to split the binary at cargo level fell upon issues with how netdog is structured that were not minimal to address

packages/.gitignore
bcressey: > This was in response to a preivous comment where my editor kept adding .lock files where not needed, sean suggested maybe we should setup a gitignore just for this level of Cargo.lock files

That sounds like weird editor behavior - is it something you can change locally? I haven't seen spurious `Cargo.lock` files show up in past pull requests.

packages/netdog/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "netdog"
+version = "0.1.0"
+edition = "2021"
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
```
bcressey, line 9: You will need to use the [source groups](https://github.com/bottlerocket-os/twoliter/blob/40e21e03aecc18de092f9cd94bb045e6c3028d8d/tools/buildsys/src/manifest.rs#L12) feature to track changes under `sources`.

To avoid also triggering rebuilds of `os` when `netdog` changes, you'll need to relocate `sources/api/netdog` to `sources/netdog`.

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,85 @@
+%global _cross_first_party 1
+%undefine _debugsource_packages
+
+Name: %{_cross_os}netdog
+Version: 0.1.1
```
bcressey, line 5: I'd prefer to keep a `0.0` version here - I don't want people thinking they have to increment this version whenever they make a change to `netdog`.

packages/os/Cargo.toml
```diff
@@ -35,6 +34,7 @@ glibc = { path = "../glibc" }
 
 # RPM Requires
 [dependencies]
+netdog = { path = "../netdog" }
```
bcressey: The wicked and systemd-networkd dependencies down below this line should be cleaned up.

What's actually left in `os.spec` that requires `netdog`? I.e. is this a true dependency? If not we shouldn't list it.

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,85 @@
...
+mkdir bin
+
+echo "** Build Netdog Binaries"
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p netdog \
+    --target-dir=${HOME}/.cache/networkd
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p netdog \
+    --features wicked \
+    --target-dir=${HOME}/.cache/wicked
```
bcressey: I'd like to be explicit in the first branch here and not default to one feature or the other:
```suggestion
%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
    -p netdog \
    --features networkd \
    --target-dir=${HOME}/.cache/networkd
%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
    -p netdog \
    --features wicked \
    --target-dir=${HOME}/.cache/wicked
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,85 @@
...
+install -d %{buildroot}%{_cross_bindir}
+install -d %{buildroot}%{_cross_tmpfilesdir}
+install -d %{buildroot}%{_cross_unitdir}
+install -d %{buildroot}%{_cross_libdir}
+install -p -m 0644 %{S:4} %{buildroot}%{_cross_tmpfilesdir}/netdog.conf
+install -p -m 0644 %{S:2} %{S:3} %{S:7} %{buildroot}%{_cross_unitdir}
+install -d %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
+install -p -m 0644 %{S:6} %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
+install -p -m 0755 ${HOME}/.cache/networkd/%{__cargo_target}/release/netdog %{buildroot}%{_cross_bindir}/netdog-systemd-networkd
+install -p -m 0755 ${HOME}/.cache/wicked/%{__cargo_target}/release/netdog %{buildroot}%{_cross_bindir}/netdog-wicked
```
bcressey: Style-wise, I would prefer to create a directory and populate it before moving to the next:
```
install -d %{buildroot}%{_cross_bindir}
install -p -m 0755 ${HOME}/.cache/networkd/%{__cargo_target}/release/netdog %{buildroot}%{_cross_bindir}/netdog-systemd-networkd
install -p -m 0755 ${HOME}/.cache/networkd/%{__cargo_target}/release/netdog %{buildroot}%{_cross_bindir}/netdog-systemd-networkd

install -d %{buildroot}%{_cross_tmpfilesdir}
install -p -m 0644 %{S:4} %{buildroot}%{_cross_tmpfilesdir}/netdog.conf

...
```

Also it would be good to number the sources so that within a "type" (systemd unit, template) the numbers are consecutive. That pairs with the above advice because it makes the lines more obviously correct - if all units are in the range from `%{S:10}` to `{S:19}` then the reviewer doesn't have to do a lot of scrolling back and forth to figure out which source is being installed.

(`release.spec` is structured this way for this reason.)

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,85 @@
...
+
+%files -n %{_cross_os}netdog-systemd-networkd
+%{_cross_bindir}/netdog-systemd-networkd
+%{_cross_tmpfilesdir}/netdog.conf
```
bcressey, line 105: Some of these files are needed for both `wicked` and `networkd` backends. Convention is to place these in a package like `%{_cross_os}netdog-common` that both packages require. Then you don't have "files listed twice" RPM warnings.

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,85 @@
...
+%{_cross_bindir}/netdog-wicked
+%{_cross_tmpfilesdir}/netdog.conf
+%{_cross_unitdir}/generate-network-config.service
+%{_cross_unitdir}/run-netdog.mount
+
+%post -n %{_cross_os}netdog-wicked -p <lua>
+posix.link("%{_cross_bindir}/netdog-wicked", "%{_cross_bindir}/netdog")
+
+%post -n %{_cross_os}netdog-systemd-networkd -p <lua>
+posix.link("%{_cross_bindir}/netdog-systemd-networkd", "%{_cross_bindir}/netdog")
```
bcressey: scriptlets come between `%install` and the first `%files` section

jmt-lab:
Reason for return of using rpmspec:

The reason for this was a bunch of roadblock updates i ran into with splitting the binaries out:

* I first attempted a pretty straight forward split where i copied [main.rs](http://main.rs/) to two bin/networkd.rs and bin/wicked.rs files, this worked fine however leads to very large duplicate files since netdog currently re-uses its [main.rs](http://main.rs/) to set a large amount of static consts used throughout the crate
* I then tried a hack by just symlinking [main.rs](http://main.rs/) twice to avoid cargo complaining about two targets using the same [main.rs](http://main.rs/) This did not propagate well into git and our ci cd
* I then tried to do a lib and bin split where i took all the static const and most of the shared code to be in [lib.rs](http://lib.rs/) and only moved the async main to the separated files, however this led to a collision with how netdog's multiple modules use private error messages with no shared top-type (meaning build error on visibility) this is fixable but would require alot ot making things public that i wasn't sure we wanted as such

As you can see above each approach i took to do the binary split at cargo time was bringing with it a bunch of more work when we had a functional solution. I decided to go back to a functional solution  and file a github issue (which i need to do) for us to refactor netdog in the future to handle both this and how the cfgs are too integrated into too many source files

packages/os/os.spec
```diff
@@ -48,20 +47,16 @@ Source114: bootstrap-containers@.service
...
 Source116: load-kernel-modules.service.in
 Source117: cfsignal.service
-Source118: generate-network-config.service
 Source119: reboot-if-required.service
 Source120: warm-pool-wait.service
-Source121: disable-udp-offload.service
 Source122: has-boot-ever-succeeded.service
-Source123: run-netdog.mount
-Source124: write-network-status.service
+Source123: disable-udp-offload.service
```
bcressey: Let's move this to `netdog.spec` as well since it's networking related, or else to `release.spec`. `release.spec` is slightly more correct since that is where the `ethtool` requirement comes in.

packages/os/os.spec
```diff
@@ -99,6 +93,7 @@ Requires: %{_cross_os}xfscli
 Requires: %{_cross_os}thar-be-settings
 Requires: %{_cross_os}thar-be-updates
 Requires: %{_cross_os}updog
+Requires: %{_cross_os}netdog
```
bcressey: This dependency should move to `release.spec`.

packages/os/os.spec
```diff
@@ -242,6 +226,11 @@ Summary: Bottlerocket certificates handler
 %description -n %{_cross_os}certdog
 %{summary}.
 
+%package -n %{_cross_os}vmware-netconfig
+Summary: Bottlerocket vmware specific network configuration
+%description -n %{_cross_os}vmware-netconfig
+%{summary}.
```
bcressey: This isn't necessary. `disable-udp-offload.service` already has `ConditionVirtualization=vmware` and won't run on other platforms, so we should unconditionally include it somewhere.

packages/os/os.spec
```diff
@@ -478,15 +466,10 @@ install -d %{buildroot}%{_cross_unitdir}
...
-  %{S:123} \
+  %{S:113} %{S:114} %{S:119} %{S:122} %{S:123} \
   %{buildroot}%{_cross_unitdir}
 
-%if %{with systemd_networkd}
-install -p -m 0644 %{S:124} %{buildroot}%{_cross_unitdir}
-install -d %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
-install -p -m 0644 %{S:12} %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
-%endif
+
```
bcressey: nit: extra newline
```suggestion
```

packages/os/os.spec
```diff
@@ -511,17 +494,13 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_tmpfilesdir}/migration.conf
 install -p -m 0644 %{S:201} %{buildroot}%{_cross_tmpfilesdir}/host-containers.conf
 install -p -m 0644 %{S:202} %{buildroot}%{_cross_tmpfilesdir}/thar-be-updates.conf
 install -p -m 0644 %{S:203} %{buildroot}%{_cross_tmpfilesdir}/bootstrap-containers.conf
-install -p -m 0644 %{S:204} %{buildroot}%{_cross_tmpfilesdir}/netdog.conf
+
```
bcressey, line 533: nit: remove added newline - we don't want extra whitespace between the different install blocks
```suggestion
```

sources/api/netdog/Cargo.toml
```diff
@@ -1,14 +1,19 @@
...
 authors = ["Ben Cressey <bcressey@amazon.com>"]
 license = "Apache-2.0 OR MIT"
 edition = "2021"
 publish = false
 # Don't rebuild crate just because of changes to README.
 exclude = ["README.md"]
 
+[features]
+default = []
+wicked = []
```
bcressey, line 13: So that we can be explicit about the feature selection in `netdog.spec`, but still have CI test one of these paths:
```suggestion
default = ["networkd"]
networkd = []
wicked = []
```

But somewhere it should be an error to define both features.

sources/api/netdog/src/cli/generate_net_config.rs
```diff
@@ -7,7 +7,7 @@ use snafu::{OptionExt, ResultExt};
 use std::fs;
 use std::path::Path;
 
-#[cfg(net_backend = "systemd-networkd")]
+#[cfg(not(feature = "wicked"))]
```
bcressey, line 10:
```suggestion
#[cfg(feature = "networkd")]
```

sources/api/netdog/src/main.rs
```diff
@@ -24,7 +24,7 @@ valid example: `netdog.default-interface=eno1:dhcp4,dhcp6?`.
 The subcommand `write-resolv-conf` writes the resolv.conf, favoring DNS API settings and
 supplementing any missing settings with DNS settings from the primary interface's DHCP lease.  It
 is meant to be used as a restart command for DNS API settings.
-*/
+ */
```
bcressey: Is this happening from `cargo fmt`? If not I wouldn't include the change:
```suggestion
*/
```

sources/api/netdog/src/networkd/config/netdev.rs
```diff
@@ -142,7 +142,7 @@ impl Display for NetDevMacAddress {
 }
 
 impl NetDevConfig {
-    const FILE_EXT: &str = "netdev";
+    const FILE_EXT: &'static str = "netdev";
```
bcressey, line 145: Why did this change?

sources/api/netdog/src/cli/mod.rs
```diff
@@ -366,17 +368,17 @@ mod error {
...
 
-        #[cfg(net_backend = "systemd-networkd")]
+        #[cfg(not(feature = "wicked"))]
         #[snafu(display("Failed to retrieve networkctl status: {}", source))]
         NetworkDInterfaceStatus {
             source: networkd_status::NetworkDStatusError,
         },
 
-        #[cfg(net_backend = "systemd-networkd")]
+        #[cfg(not(feature = "wicked"))]
```
bcressey, line 384: Switching from conditional compilation to a feature is a pretty big change - might be better to do it in its own commit.

sources/api/netdog/src/cli/mod.rs
```diff
@@ -228,7 +228,9 @@ mod tests {
         let fake_file = tempfile::NamedTempFile::new().unwrap();
         let expected = "-net.ipv4.conf.eno1.rp_filter = 2\n";
         write_interface_sysctl(interface, &fake_file).unwrap();
-        assert_eq!(std::fs::read_to_string(&fake_file).unwrap(), expected);
+        assert!(std::fs::read_to_string(&fake_file)
```
bcressey, line 231: Was this an existing unit test failure that wasn't caught because we don't test the systemd-networkd path?

variants/aws-ecs-2/Cargo.toml
```diff
@@ -12,7 +12,6 @@ grub-set-private-var = true
 unified-cgroup-hierarchy = true
 uefi-secure-boot = true
 xfs-data-partition = true
-systemd-networkd = true
```
bcressey: Can you open an issue in twoliter to remove this feature flag if we're not using it? We should cut it from `buildsys` and `Dockerfile`.

packages/netdog/Cargo.toml
jmt-lab: 👍

packages/netdog/netdog.spec
jmt-lab: So you would like me to return the netdog package to having two features instead of one, Matt's main argument for having the one is that --all-features would still work (albeit will only test the wicked branch)

sources/api/netdog/src/networkd/config/netdev.rs
jmt-lab: This is a requirement of upgrated rust sdk. I'm surprised we aren't being riddled with more of these errors, but when the rust update landed as part of 1.18 into develop and i upgraded the PR it started flagging this as an error (previously was a warning) All module global consts need to have a static lifetime if using a type that requires lifetimes such as &str. It is no longer implicit

sources/api/netdog/src/cli/mod.rs
jmt-lab: correct

variants/aws-ecs-2/Cargo.toml
jmt-lab: yup will do

jmt-lab:
fix version number and remove accidental file

variants/aws-ecs-2/Cargo.toml
webern: I opened https://github.com/bottlerocket-os/twoliter/issues/142

packages/systemd/systemd.spec
```diff
@@ -125,7 +125,6 @@ Summary: Files for networkd
 
 %package resolved
 Summary: Files for resolved
-
```
webern: This feels unnecessary and would take the PR down to 254 files :sweat_smile:

sources/netdog/build.rs
```diff
@@ -0,0 +1,19 @@
+use bottlerocket_variant::{Variant, VARIANT_ENV};
```
webern: When moving files I like to do so in an isolated commit before making changes to the files. That way we can be sure git understands that the file was moved so that changes can be tracked across the move. I was able to find at least one person on the Internet that agrees with me: [Always-move-and-rename-Git-files-in-an-isolated-commit](https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/Always-move-and-rename-Git-files-in-an-isolated-commit)

sources/netdog/README.md
```diff
@@ -1,6 +1,6 @@
 # netdog
 
-Current version: 0.1.0
+Current version: 0.1.1
```
webern: If you reverted from `0.1.1` back to `0.1.0` then this needs to be regenerated/reverted as well.

sources/netdog/src/cli/mod.rs
```diff
@@ -4,12 +4,12 @@ pub(crate) mod node_ip;
...
 
-#[cfg(net_backend = "wicked")]
+#[cfg(feature = "wicked")]
 pub(crate) mod install;
-#[cfg(net_backend = "wicked")]
+#[cfg(feature = "wicked")]
 pub(crate) mod remove;
 
-#[cfg(net_backend = "systemd-networkd")]
+#[cfg(not(feature = "wicked"))]
```
webern, line 12: Thank you for making this change. I think this is better instead of having mutually-exclusive features like before.

sources/netdog/src/networkd/config/netdev.rs
```diff
@@ -142,7 +142,7 @@ impl Display for NetDevMacAddress {
 }
 
 impl NetDevConfig {
-    const FILE_EXT: &str = "netdev";
+    const FILE_EXT: &'static str = "netdev";
```
webern, line 145: Unnecessary, literals/consts are always understood to be static.

```suggestion
    const FILE_EXT: &str = "netdev";
```

packages/netdog/Cargo.toml
webern: > To avoid also triggering rebuilds of os when netdog changes, you'll need to relocate sources/api/netdog to sources/netdog.

Why is this, do we have a watch on all of `sources/api`? Incidentally, I don't think I've ever seen a variant build in which os is not rebuilt.

packages/netdog/netdog.spec
webern: I have a preference for non-mutually-exclusive features (the way you have it now), but perhaps you can use `--features default` or `--default-features=true` to make this more self documenting.

```sugestion
# Build with systemd-networkd as the backend
%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
    -p netdog \
    --features default \
    --default-features=true \
    --target-dir=${HOME}/.cache/networkd

# Build with wicked as the backend
%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
    -p netdog \
    --features wicked \
    --default-features=false
    --target-dir=${HOME}/.cache/wicked
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,92 @@
...
+mkdir bin
+
+echo "** Build Netdog Binaries"
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p netdog \
+    --target-dir=${HOME}/.cache/networkd
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p netdog \
+    --features wicked \
+    --target-dir=${HOME}/.cache/wicked
```
webern, line 77: Maybe this helps with self-documenting code: 

```suggestion
    --features wicked \
    --default-features=false \
    --target-dir=${HOME}/.cache/wicked
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,92 @@
...
+%{_cross_bindir}/netdog-wicked
+%{_cross_tmpfilesdir}/netdog.conf
+%{_cross_unitdir}/generate-network-config.service
+%{_cross_unitdir}/run-netdog.mount
+
+%post -n %{_cross_os}netdog-wicked -p <lua>
+posix.link("%{_cross_bindir}/netdog-wicked", "%{_cross_bindir}/netdog")
+
+%post -n %{_cross_os}netdog-systemd-networkd -p <lua>
+posix.link("%{_cross_bindir}/netdog-systemd-networkd", "%{_cross_bindir}/netdog")
```
webern: Newline! :)
```suggestion
posix.link("%{_cross_bindir}/netdog-systemd-networkd", "%{_cross_bindir}/netdog")

```

packages/netdog/Cargo.toml
jmt-lab: This was done

sources/netdog/build.rs
jmt-lab: Will do next time, right now undoing and fixing this would be really painful

jmt-lab:
Addressing comments and rebase on latest

webern (dismissed):
Approved assuming that Ben's packaging suggestions are addressed.

variants/vmware-k8s-1.24/Cargo.toml
sam-berning, line 1: Looks like the rebase resurrected this variant and metal-k8s-1.24 which were removed [here](https://github.com/bottlerocket-os/bottlerocket/commit/f1be358629beccaf3dc4bd18d4939ffec04a2de8)

jmt-lab: yeah noticed cleaning this up

jmt-lab:
Addressed Ben's concerns on rpm adjustments
Rebased off develop
Restructured commits to isolate the move from edit and grouped changes by category of work

variants/vmware-dev/Cargo.toml
```diff
@@ -33,6 +32,7 @@ included-packages = [
     "release",
     "kernel-6.1",
     "open-vm-tools",
+    "netdog-vmware-config",
```
sam-berning: this package doesn't exist anymore, should be removed

sources/netdog/build.rs
jmt-lab: Fixed this

variants/vmware-dev/Cargo.toml
jmt-lab: oop you are right

jmt-lab:
Removed redundant package in vmware variants

sam-berning (approved):
alright, lgtm pending testing. nice work!

jmt-lab:
Testing result on changes: Everything looks good and is functioning as expected

jmt-lab:
rebased off latest

sources/Cargo.lock
```diff
@@ -2661,6 +2661,7 @@ dependencies = [
 name = "netdog"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
```
bcressey: Please stick with `snafu` for OS components. Also, I don't think this should be part of this commit, but that's irrelevant if it goes away.

sources/netdog/Cargo.toml
```diff
@@ -8,11 +8,16 @@ publish = false
...
+default = []
+wicked = []
+
 [dependencies]
+anyhow = "1.0"
 argh = "0.1"
 dns-lookup = "2"
 ipnet = { version = "2", features = ["serde"] }
-imdsclient = { path = "../../imdsclient", version = "0.1" }
+imdsclient = { path = "../imdsclient", version = "0.1" }
```
bcressey: There's another source of conditional compilation implied by this - `imdsclient` is used in `generate_hostname.rs` which has these:
```
#[cfg(variant_platform = "aws")]
#[cfg(not(variant_platform = "aws"))]
```

sources/netdog/build.rs
```diff
@@ -25,19 +14,6 @@ fn main() {
         }
     };
     variant.emit_cfgs();
-    emit_net_backend_cfgs();
```
bcressey: `variant.emit_cfgs();` above this line is also problematic.

packages/netdog/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
```
bcressey, line 1: Please wrap the summary line of your commit messages around 50 characters. The ellipsis in the GitHub web view is an indication that it's too long.

I tend to follow [this style guide](https://cbea.ms/git-commit/) with the addition of a "subsystem: ..." prefix to the summary.

packages/netdog/Cargo.toml
bcressey, line 9: Still need to add something like:
```
source-groups = [
    "netdog"
]
```

packages/os/os.spec
```diff
@@ -511,17 +487,13 @@ install -p -m 0644 %{S:200} %{buildroot}%{_cross_tmpfilesdir}/migration.conf
...
```
bcressey, line 533: This extra newline still shows up in this commit.

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,101 @@
...
+URL: https://github.com/bottlerocket-os/bottlerocket
+
+Source2: netdog-tmpfiles.conf
+
+Source10: run-netdog.mount
+Source11: write-network-status.service
+Source12: generate-network-config.service
+Source13: disable-udp-offload.service
+
+Source20: 00-resolved.conf
```
bcressey: Why start with `Source2`?

Headers in a comment might help enforce the grouping, e.g.:
```
# drop-in configs
...

# units
...

# service configs
...
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,101 @@
...
+
+install -d %{buildroot}%{_cross_libdir}
+install -d %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
+install -p -m 0644 %{S:20} %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
+
+%post -n %{_cross_os}netdog-wicked -p <lua>
+posix.link("%{_cross_bindir}/netdog-wicked", "%{_cross_bindir}/netdog")
+
+%post -n %{_cross_os}netdog-systemd-networkd -p <lua>
+posix.link("%{_cross_bindir}/netdog-systemd-networkd", "%{_cross_bindir}/netdog")
```
bcressey: Avoid creating absolute symlinks. They cause problems when working with different roots - e.g. when the host path is mounted in a container, or when extracting the RPM contents - since they can point outside the root to another unpackaged file.

```
%post -n %{_cross_os}netdog-wicked -p <lua>
posix.link("netdog-wicked", "%{_cross_bindir}/netdog")

%post -n %{_cross_os}netdog-systemd-networkd -p <lua>
posix.link("netdog-systemd-networkd", "%{_cross_bindir}/netdog")
```

packages/os/os.spec
```diff
@@ -478,15 +459,10 @@ install -d %{buildroot}%{_cross_unitdir}
...
-  %{S:123} \
+  %{S:113} %{S:114} %{S:119} %{S:122} \
   %{buildroot}%{_cross_unitdir}
 
-%if %{with systemd_networkd}
-install -p -m 0644 %{S:124} %{buildroot}%{_cross_unitdir}
-install -d %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
-install -p -m 0644 %{S:12} %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
-%endif
+
```
bcressey: The extra newline still shows up in this commit.

packages/release/Cargo.toml
```diff
@@ -38,6 +38,7 @@ libaudit = { path = "../libaudit" }
 libgcc = { path = "../libgcc" }
 libstd-rust = { path = "../libstd-rust" }
 makedumpfile = { path = "../../packages/makedumpfile" }
+netdog = {path = "../netdog" }
```
bcressey, line 42: `release` should no longer depend on `wicked` - right now this dependency is forcing it to be built for all variants, but it should only be built for variants that include it.

variants/aws-ecs-1-nvidia/Cargo.toml
```diff
@@ -20,6 +20,7 @@ included-packages = [
 # core
     "release",
     "kernel-5.10",
+    "wicked",
```
bcressey: The wicked variants should also add a dependency on wicked in the `build-dependencies` section below.

jmt-lab:
* Changed to utilize image-features
* Integrate use of dogtag hostname helpers

jmt-lab:
For testing I did the following:

- [x] aws-ecs-2 test
   - [x] verify network connectivity
   - [x] verify netdog-systemd-networkd is installed and symlinked
   - [x] verify 10-reverse-dns is installed
   - [x] verify 20-imds is installed
   - [x] verify hostname got set correctly and it did not fall through to setting ip
- [x] aws-ecs-1 test
   - [x] verify network connectivity
   - [x] verify netdog-wicked is installed and symlinked
   - [x] verify 10-reverse-dns is installed
   - [x] verify 20-imds is installed
   - [x] verify hostname got set correctly and it did not fallthrough to setting ip

variants/aws-k8s-1.24/Cargo.toml
```diff
@@ -39,3 +39,4 @@ cni-plugins = { path = "../../packages/cni-plugins" }
 kernel-5_15 = { path = "../../packages/kernel-5.15" }
 kubernetes-1_24= { path = "../../packages/kubernetes-1.24" }
 release = { path = "../../packages/release" }
+wicked = { path = "../../packages/wicked" }
```
bcressey: We shouldn't need this change.

packages/netdog/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+    "netdog",
+    "dogtag",
+]
+
+# RPM BuildRequires
+[build-dependencies]
+glibc = { path = "../glibc" }
+
+[dependencies]
+wicked = { path = "../wicked" }
```
bcressey, line 22: Just because it's weird to only depend on one of the two backend daemons:
```suggestion
[dependencies]
wicked = { path = "../wicked" }
systemd = { path = "../systemd" }
```

sources/dogtag/src/lib.rs
```diff
@@ -28,7 +29,7 @@ pub struct Cli {
 pub type Result<T> = std::result::Result<T, error::Error>;
 
 /// find_hostname will utilize the helpers located in /var/bottlerocket/dogtag/ to try and discover the hostname
-pub async fn find_hostname() -> Result<String> {
+pub async fn find_hostname(ip_addr: IpAddr) -> Result<String> {
```
bcressey: nit: comment still refers to `/var/bottlerocket/dogtag`.

sources/netdog/src/cli/generate_hostname.rs
```diff
@@ -0,0 +1,30 @@
...
+/// The result is returned as JSON. (intended for use as a settings generator)
+pub(crate) async fn run() -> Result<()> {
+    let ip_string = fs::read_to_string(CURRENT_IP)
+        .context(error::CurrentIpReadFailedSnafu { path: CURRENT_IP })?;
+    let ip = IpAddr::from_str(&ip_string).context(error::IpFromStringSnafu { ip: &ip_string })?;
+
+    // If no hostname has been determined we return the IP address of the host, replacing invalid ipv6 chars.
+    let hostname = find_hostname(ip)
+        .await
+        .unwrap_or(ip_string.replace(':', "-"));
```
bcressey: Might be better to handle this in `dogtag`.

The other potential edge case is where we "find" a hostname, but it's the same as the IPv6 address we passed in, in which case we'd still want to do the replacement.

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+%package -n %{_cross_os}netdog-common
+Summary: Common configuration for Bottlerocket's network configuration helper
+Requires: %{_cross_os}hostname-reverse-dns
+Requires: (%{_cross_os}hostname-imds if %{_cross_os}variant-platform(aws))
+%description -n %{_cross_os}netdog-common
+%{summary}.
+
+%package -n %{_cross_os}netdog-systemd-networkd
+Summary: Bottlerocket network configuration helper
+Provides: %{_cross_os}netdog = 2
```
bcressey: We don't need this.

```suggestion
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+Requires: %{_cross_os}systemd-networkd
+Requires: %{_cross_os}systemd-resolved
+Requires: %{_cross_os}netdog
+Conflicts: (%{_cross_os}netdog-wicked or %{_cross_os}image-feature(no-systemd-networkd))
+%description -n %{_cross_os}netdog-systemd-networkd
+%{summary}.
+
+%package -n %{_cross_os}netdog-wicked
+Summary: Bottlerocket network configuration helper
+Provides: %{_cross_os}netdog = 1
```
bcressey:
```suggestion
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+
+BuildRequires: %{_cross_os}glibc-devel
+Requires: %{_cross_os}netdog
+Requires: (%{_cross_os}netdog-systemd-networkd if %{_cross_os}image-feature(systemd-networkd))
+Requires: (%{_cross_os}netdog-wicked if %{_cross_os}image-feature(no-systemd-networkd))
+
+%description
+%{summary}.
+
+%package -n %{_cross_os}netdog-common
```
bcressey: Prefer the shorthand syntax for subpackages that suffix the main package name:
```suggestion
%package common
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+
+Source10: run-netdog.mount
+Source11: write-network-status.service
+Source12: generate-network-config.service
+Source13: disable-udp-offload.service
+
+Source20: 00-resolved.conf
+
+BuildRequires: %{_cross_os}glibc-devel
+Requires: %{_cross_os}netdog
```
bcressey: This is a little weird - package requiring itself - and isn't needed when we drop the epoch provides hack:
```suggestion
```

The two requirements below are a nicer way of doing that.

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+Requires: (%{_cross_os}netdog-systemd-networkd if %{_cross_os}image-feature(systemd-networkd))
+Requires: (%{_cross_os}netdog-wicked if %{_cross_os}image-feature(no-systemd-networkd))
```
bcressey, line 24: shorthand syntax:
```suggestion
Requires: (%{name}-systemd-networkd if %{_cross_os}image-feature(systemd-networkd))
Requires: (%{name}-wicked if %{_cross_os}image-feature(no-systemd-networkd))
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+
+%description
+%{summary}.
+
+%package -n %{_cross_os}netdog-common
+Summary: Common configuration for Bottlerocket's network configuration helper
+Requires: %{_cross_os}hostname-reverse-dns
+Requires: (%{_cross_os}hostname-imds if %{_cross_os}variant-platform(aws))
+%description -n %{_cross_os}netdog-common
```
bcressey:
```suggestion
%description common
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+%{summary}.
+
+%package -n %{_cross_os}netdog-systemd-networkd
```
bcressey:
```suggestion
%package systemd-networkd
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+Summary: Bottlerocket network configuration helper
+Provides: %{_cross_os}netdog = 2
+Requires: %{_cross_os}netdog-common
```
bcressey:
```suggestion
Requires: %{name}-common
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+Requires: %{_cross_os}systemd-networkd
+Requires: %{_cross_os}systemd-resolved
+Requires: %{_cross_os}netdog
+Conflicts: (%{_cross_os}netdog-wicked or %{_cross_os}image-feature(no-systemd-networkd))
+%description -n %{_cross_os}netdog-systemd-networkd
```
bcressey, line 35:
```suggestion
Conflicts: (%{name}-wicked or %{_cross_os}image-feature(no-systemd-networkd))
%description systemd-networkd
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+%{summary}.
+
+%package -n %{_cross_os}netdog-wicked
```
bcressey:
```suggestion
%package wicked
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+Summary: Bottlerocket network configuration helper
+Provides: %{_cross_os}netdog = 1
+Requires: %{_cross_os}netdog-common
```
bcressey:
```suggestion
Requires: %{name}-common
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+%description -n %{_cross_os}netdog-common
+%{summary}.
+
+%package -n %{_cross_os}netdog-systemd-networkd
+Summary: Bottlerocket network configuration helper
+Provides: %{_cross_os}netdog = 2
+Requires: %{_cross_os}netdog-common
+Requires: %{_cross_os}systemd-networkd
+Requires: %{_cross_os}systemd-resolved
+Requires: %{_cross_os}netdog
```
bcressey:
```suggestion
Requires: %{name}
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+Conflicts: (%{_cross_os}netdog-wicked or %{_cross_os}image-feature(no-systemd-networkd))
+%description -n %{_cross_os}netdog-systemd-networkd
+%{summary}.
+
+%package -n %{_cross_os}netdog-wicked
+Summary: Bottlerocket network configuration helper
+Provides: %{_cross_os}netdog = 1
+Requires: %{_cross_os}netdog-common
+Requires: %{_cross_os}wicked
+Requires: %{_cross_os}netdog
```
bcressey:
```suggestion
Requires: %{name}
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+Conflicts: (%{_cross_os}netdog-systemd-networkd or %{_cross_os}image-feature(systemd-networkd))
```
bcressey, line 42:
```suggestion
Conflicts: (%{name}-systemd-networkd or %{_cross_os}image-feature(systemd-networkd))
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+%description -n %{_cross_os}netdog-wicked
+%{summary}.
```
bcressey, line 44:
```suggestion
%description wicked
%{summary}.
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+    -p netdog \
+    --features default \
+    --target-dir=${HOME}/.cache/networkd
+%cargo_build --manifest-path %{_builddir}/sources/Cargo.toml \
+    -p netdog \
+    --features wicked \
+    --target-dir=${HOME}/.cache/wicked
+
+%install
+install -d %{buildroot}%{_cross_libexecdir}/hostname-resolvers
```
bcressey: I would pick a consistent phrase - either "hostname resolvers" or "hostname detectors" ("detectors" is in the package description and would be easiest to change).

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+
+install -d %{buildroot}%{_cross_libdir}
+install -d %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
+install -p -m 0644 %{S:20} %{buildroot}%{_cross_libdir}/systemd/resolved.conf.d
+
+%post -n %{_cross_os}netdog-wicked -p <lua>
+posix.symlink("netdog-wicked", "%{_cross_bindir}/netdog")
+
+%post -n %{_cross_os}netdog-systemd-networkd -p <lua>
+posix.symlink("netdog-systemd-networkd", "%{_cross_bindir}/netdog")
```
bcressey:
```suggestion
%post wicked -p <lua>
posix.symlink("netdog-wicked", "%{_cross_bindir}/netdog")

%post systemd-networkd -p <lua>
posix.symlink("netdog-systemd-networkd", "%{_cross_bindir}/netdog")
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+
+%files -n %{_cross_os}hostname-reverse-dns
+%{_cross_libexecdir}/hostname-resolvers/10-reverse-dns
+
+%files -n %{_cross_os}hostname-imds
+%{_cross_libexecdir}/hostname-resolvers/20-imds
+
+%files -n %{_cross_os}netdog-common
```
bcressey:
```suggestion
%files common
```

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,133 @@
...
+%{_cross_unitdir}/run-netdog.mount
+
+%files -n %{_cross_os}netdog-systemd-networkd
+%{_cross_bindir}/netdog-systemd-networkd
+%{_cross_unitdir}/write-network-status.service
+%dir %{_cross_libdir}/systemd/resolved.conf.d
+%{_cross_libdir}/systemd/resolved.conf.d/00-resolved.conf
+
+%files -n %{_cross_os}netdog-wicked
+%{_cross_bindir}/netdog-wicked
```
bcressey:
```suggestion
%files systemd-networkd
%{_cross_bindir}/netdog-systemd-networkd
%{_cross_unitdir}/write-network-status.service
%dir %{_cross_libdir}/systemd/resolved.conf.d
%{_cross_libdir}/systemd/resolved.conf.d/00-resolved.conf

%files wicked
%{_cross_bindir}/netdog-wicked
```

packages/netdog/netdog.spec
bcressey: Actually, you don't have anything in the `netdog` package now, and we don't need to play the weird game with epochs, so I would just drop the "common" subpackage and move this stuff into `%files` (for the main netdog package).

sources/netdog/src/cli/generate_hostname.rs
jmt-lab: So the dogtag lib accepts a full IpAddr object instead of the ip_string itself. I imagined this replacement was more needed for storage of the hostname in the settings or on disk configuration?

jmt-lab: We talked about it going to move the ip-address fallback in dogtag find_hostname

jmt-lab:
Updated dogtag to handle fallback to ip address

variants/metal-k8s-1.27/Cargo.toml
```diff
@@ -16,6 +16,7 @@ partition-plan = "unified"
 [package.metadata.build-variant.image-features]
 grub-set-private-var = true
 unified-cgroup-hierarchy = true
+systemd-networkd = true
```
sam-berning: why is this present in only this one variant?

jmt-lab: good catch i've removed it

jmt-lab:
Fix formatting

packages/netdog/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+systemd = { path = "../systemd" }
```
bcressey: Should be a newline at the end of this file - can you verify that your editor is configured to do this?

packages/netdog/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+publish = false
+build = "../build.rs"
+
+[lib]
+path = "../packages.rs"
+
+[package.metadata.build-packages]
+source-groups = [
+    "netdog",
+    "dogtag",
```
bcressey, line 14: nit: we could put `dogtag` under `netdog` (like how various update-related crates are under `sources/updater`) but that can be done later

sources/netdog/src/cli/generate_hostname.rs
```diff
@@ -0,0 +1,30 @@
...
+/// The result is returned as JSON. (intended for use as a settings generator)
+pub(crate) async fn run() -> Result<()> {
+    let ip_string = fs::read_to_string(CURRENT_IP)
+        .context(error::CurrentIpReadFailedSnafu { path: CURRENT_IP })?;
+    let ip = IpAddr::from_str(&ip_string).context(error::IpFromStringSnafu { ip: &ip_string })?;
+
+    // If no hostname has been determined we return the IP address of the host, replacing invalid ipv6 chars.
+    let hostname = find_hostname(ip)
+        .await
+        .context(error::HostnameDetectionSnafu)?;
```
bcressey: nit: comments are out-of-date since they refer to a `dogtag` implementation detail:
```suggestion
/// Attempt to convert the assigned IP address to a valid hostname.
///
/// The result is returned as JSON. (intended for use as a settings generator)
pub(crate) async fn run() -> Result<()> {
    let ip_string = fs::read_to_string(CURRENT_IP)
        .context(error::CurrentIpReadFailedSnafu { path: CURRENT_IP })?;
    let ip = IpAddr::from_str(&ip_string).context(error::IpFromStringSnafu { ip: &ip_string })?;
    let hostname = find_hostname(ip)
        .await
        .context(error::HostnameDetectionSnafu)?;
```

packages/os/os.spec
bcressey: Extra newline is still here!

packages/netdog/netdog.spec
```diff
@@ -0,0 +1,123 @@
...
+Requires: (%{_cross_os}hostname-imds if %{_cross_os}variant-platform(aws))
+Requires: (%{_cross_os}netdog-systemd-networkd if %{_cross_os}image-feature(systemd-networkd))
+Requires: (%{_cross_os}netdog-wicked if %{_cross_os}image-feature(no-systemd-networkd))
+
+%description
+%{summary}.
+
+%package systemd-networkd
+Summary: Bottlerocket network configuration helper
+Requires: %{_cross_os}netdog
```
bcressey:
```suggestion
Requires: %{name}
```

jmt-lab:
* Rebased off develop
* Bump netdog version
* Adjust Requires
* Add missing newline
* Remove extra newline
* Adjust comment

bcressey (approved):
🚢 🚢 🚢 🥅 🐶

PR #3701, opened 2024-01-08
Fine-tune the Ordering for the atomic usages

wang384670111:
https://github.com/bottlerocket-os/bottlerocket/blob/147a019a57f3575b04669277da033c5b0c67b30e/sources/api/apiclient/src/exec.rs#L307-L311
https://github.com/bottlerocket-os/bottlerocket/blob/147a019a57f3575b04669277da033c5b0c67b30e/sources/api/apiclient/src/exec.rs#L476-L477

Here, `max_messages_outstanding` and `messages_written` are simply used as markers in a multithreaded context and do not synchronize with other locals. Therefore, I think using `Relaxed` ordering should be adequate to ensure the program's correctness.

### DIFF
--- a/sources/api/apiclient/src/exec.rs
+++ b/sources/api/apiclient/src/exec.rs
@@ -306,10 +306,10 @@ impl ReadFromServer {
                                     );
                                     capacity
                                         .max_messages_outstanding
-                                        .store(new.max_messages_outstanding, Ordering::SeqCst);
+                                        .store(new.max_messages_outstanding, Ordering::Relaxed);
                                     capacity
                                         .messages_written
-                                        .store(new.messages_written, Ordering::SeqCst);
+                                        .store(new.messages_written, Ordering::Relaxed);
                                 }
                             }
                         }
@@ -473,8 +473,8 @@ impl ReadFromUser {
     fn wait_for_capacity(messages_read: u64, capacity: &Arc<AtomicCapacity>) -> Result<()> {
         let mut waited = 0u64;
         loop {
-            let max_outstanding = capacity.max_messages_outstanding.load(Ordering::SeqCst);
-            let messages_written = capacity.messages_written.load(Ordering::SeqCst);
+            let max_outstanding = capacity.max_messages_outstanding.load(Ordering::Relaxed);
+            let messages_written = capacity.messages_written.load(Ordering::Relaxed);
 
             // Check how many messages are currently waiting to be written; read - written.
             // If the server has written more than we've read, something is quite wrong!


sources/api/apiclient/src/exec.rs
```diff
@@ -473,8 +473,8 @@ impl ReadFromUser {
     fn wait_for_capacity(messages_read: u64, capacity: &Arc<AtomicCapacity>) -> Result<()> {
         let mut waited = 0u64;
         loop {
-            let max_outstanding = capacity.max_messages_outstanding.load(Ordering::SeqCst);
-            let messages_written = capacity.messages_written.load(Ordering::SeqCst);
+            let max_outstanding = capacity.max_messages_outstanding.load(Ordering::Relaxed);
+            let messages_written = capacity.messages_written.load(Ordering::Relaxed);
```
cbgbt, line 477: I have some reservations about the `messages_written` variable. It looks like we perform subtraction of this from `messages_read` later. Couldn't this ultimately suffer without synchronization?

While relaxing the ordering here *may* be correct (I actually can't tell!), `apiclient` is ultimately IO-bound. I think my general inclination is that removing sequential consistency wouldn't result in any noticeable performance improvement and introduces correctness risks here.

cbgbt (commented):
Thanks for submitting this! I have some reservations about this change, though.

sources/api/apiclient/src/exec.rs
wang384670111: > I have some reservations about the `messages_written` variable. It looks like we perform subtraction of this from `messages_read` later. Couldn't this ultimately suffer without synchronization?
> 
> While relaxing the ordering here _may_ be correct (I actually can't tell!), `apiclient` is ultimately IO-bound. I think my general inclination is that removing sequential consistency wouldn't result in any noticeable performance improvement and introduces correctness risks here.

I understand your point.

you're suggesting that using `Relaxed` might lead to `messages_written` being read as an `outdated value`, which could result in incorrect subtraction by `messages_read`. However, this is a common `misconception` about memory ordering. My clarification is:

Each atomic operation is inherently atomic, ensuring that every thread observes a `consistent modification sequence` for `an atomic`, which is distinct from `sequential consistency`. Therefore, regardless of the memory ordering used (`Relaxed`, `Acquire`, `SeqCst`), each `load` retrieves the `latest value` of the atomic variable at that moment. The purpose of using `Acquire`/`Release` is to conditionally pass knowledge about `other locations`, not about the location that is being stored to and loaded from. Therefore, in this case, since `messages_written` does not synchronize access to other variables, and because each load operation retrieves the current atomic value—even if it is followed by a subtraction—the behavior aligns with the program's expectations.

If you have any further queries or insights, I would be very interested to hear your thoughts and am more than willing to provide any necessary explanations.

Thank you very much for taking the time to consider this.

wang384670111:
Certainly, I'd like to add that I fully understand your cautious approach in opting for `SeqCst`. If loosening the `memory ordering` doesn't significantly boost performance, then opting for the simpler `SeqCst` can indeed be a wise choice. Using `Relaxed` does require a great deal of courage.

I've recently noticed that this cautious approach towards `memory ordering` is a common sentiment in the community. However, I believe that `SeqCst` is often unnecessary, and that `Acquire`/`Release` is correct in most cases. Therefore, a deeper understanding of memory ordering is essential to employ it correctly.

cbgbt:
I think my fear with the move to `Relaxed` is that the assumptions required for correct behavior are easy to accidentally invalidate in the future, so I'm going to close this PR. But I really do appreciate the PR and the discussion -- thank you!

PR #3706, opened 2024-01-09
certdog: change certdog to use a generated certdog.toml file

### LINKED ISSUE #3623: OOTB: Remove model dependency from `certdog`
`certdog` queries the API for all settings, deserializing into the model. It should probably read from a template rendered by a settings extension.

jmt-lab:
**Issue number:** #3623 

Closes #3623 

**Description of changes:**
Changed certdog binary to accept an argument -c instead of -s that defaults to the location /etc/certdog.toml. This toml file will be generated by schnauzer from settings.pki. Certdog now will read the custom certificates from this generated template instead of calling back to the settings api.

**Testing done:**
Tested via ecs using the aws-ecs-2 variant and verified that the file is generated correctly and certdog runs with no errors

**Terms of contribution:**

### DIFF
--- a/Release.toml
+++ b/Release.toml
@@ -270,4 +270,7 @@ version = "1.19.2"
     "migrate_v1.19.1_aws-control-container-v0-7-8.lz4",
     "migrate_v1.19.1_public-control-container-v0-7-8.lz4",
 ]
-"(1.19.1, 1.19.2)" = []
+"(1.19.1, 1.19.2)" = [
+    "migrate_v1.19.2_certdog-config-file-v0-1-0.lz4",
+    "migrate_v1.19.2_certdog-service-cfg-v0-1-0.lz4",
+]
--- a/packages/os/certdog-toml
+++ b/packages/os/certdog-toml
@@ -1,0 +1,10 @@
+[required-extensions]
+pki = "v1"
++++
+{{#if settings.pki}}
+{{#each settings.pki}}
+["{{@key}}"]
+trusted = {{this.trusted}}
+data = "{{this.data}}"
+{{/each}}
+{{/if}}
--- a/sources/api/certdog/src/main.rs
+++ b/sources/api/certdog/src/main.rs
@@ -15,11 +15,14 @@ use std::fmt::Write;
 use std::fs;
 use std::io::BufReader;
 use std::io::{BufRead, Seek};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process;
 
 use model::modeled_types::Identifier;
+use model::PemCertificate;
 
+// Default location of the config file
+const DEFAULT_CONFIG_FILE: &str = "/etc/certdog.toml";
 // Read from the source in `/usr/share/factory` not the copy in `/etc`
 const DEFAULT_SOURCE_BUNDLE: &str = "/usr/share/factory/etc/pki/tls/certs/ca-bundle.crt";
 // This file is first created with tmpfilesd configurations
@@ -36,9 +39,9 @@ struct Args {
     #[argh(option, default = "LevelFilter::Info", short = 'l')]
     /// log-level trace|debug|info|warn|error
     log_level: LevelFilter,
-    #[argh(option, default = "constants::API_SOCKET.to_string()", short = 's')]
-    /// socket-path path to apiserver socket
-    socket_path: String,
+    #[argh(option, default = "PathBuf::from(DEFAULT_CONFIG_FILE)", short = 'c')]
+    /// config-path path to certdog toml
+    config_path: PathBuf,
     #[argh(option, default = "DEFAULT_TRUSTED_STORE.to_string()", short = 't')]
     /// trusted-store path to the trusted store
     trusted_store: String,
@@ -47,6 +50,7 @@ struct Args {
     source_bundle: String,
 }
 
+#[derive(Default)]
 struct CertBundle {
     trusted_certs: Vec<x509_parser::pem::Pem>,
     distrusted_certs: Vec<x509_parser::pem::Pem>,
@@ -54,30 +58,29 @@ struct CertBundle {
 
 /// Query the API for the certificate bundles, returns a tuple with trusted
 /// and distrusted PEM certificates
-async fn get_certificate_bundles<P>(socket_path: P) -> Result<CertBundle>
+async fn get_certificate_bundles<P>(config_path: P) -> Result<CertBundle>
 where
     P: AsRef<Path>,
 {
-    debug!("Querying the API for settings");
+    debug!("Loading certdog configuration");
+    let config_path = config_path.as_ref();
+    if !config_path.exists() {
+        return Ok(CertBundle::default());
+    }
 
-    let method = "GET";
-    let uri = constants::API_SETTINGS_URI;
-    let (_code, response_body) = apiclient::raw_request(&socket_path, uri, method, None)
-        .await
-        .context(error::APIRequestSnafu { method, uri })?;
+    let config_str = fs::read_to_string(config_path).context(error::ConfigReadSnafu {
+        config: format!("{:?}", config_path),
+    })?;
+    let pki: HashMap<Identifier, PemCertificate> =
+        toml::from_str(config_str.as_str()).context(error::ConfigDeserializationSnafu {
+            config: format!("{:?}", config_path),
+        })?;
 
-    // Build a Settings struct from the response string
-    debug!("Deserializing response");
-    let settings: model::Settings =
-        serde_json::from_str(&response_body).context(error::ResponseJsonSnafu { uri })?;
-
-    split_bundles(settings.pki.unwrap_or_default())
+    split_bundles(pki)
 }
 
 /// Returns a tuple with two lists, for trusted and distrusted certificates
-fn split_bundles(
-    certificates_bundle: HashMap<Identifier, model::PemCertificate>,
-) -> Result<CertBundle> {
+fn split_bundles(certificates_bundle: HashMap<Identifier, PemCertificate>) -> Result<CertBundle> {
     let mut trusted_certs: Vec<x509_parser::pem::Pem> = Vec::new();
     let mut distrusted_certs: Vec<x509_parser::pem::Pem> = Vec::new();
 
@@ -210,8 +213,8 @@ async fn run() -> Result<()> {
     SimpleLogger::init(args.log_level, LogConfig::default()).context(error::LoggerSnafu)?;
 
     info!("certdog started");
-    let certificate_bundles = get_certificate_bundles(&args.socket_path).await?;
-    info!("Got certificate bundles from API");
+    let certificate_bundles = get_certificate_bundles(&args.config_path).await?;
+    info!("Got certificate bundles from configuration file");
     update_trusted_store(certificate_bundles, args.trusted_store, args.source_bundle)?;
     info!("Updated trusted store");
 
@@ -238,12 +241,18 @@ mod error {
     #[derive(Debug, Snafu)]
     #[snafu(visibility(pub(super)))]
     pub(super) enum Error {
-        #[snafu(display("Error sending {} to {}: {}", method, uri, source))]
-        APIRequest {
-            method: String,
-            uri: String,
-            #[snafu(source(from(apiclient::Error, Box::new)))]
-            source: Box<apiclient::Error>,
+        #[snafu(display("Error reading data from configuration at {}: {}", config, source))]
+        ConfigRead {
+            config: String,
+            #[snafu(source(from(std::io::Error, Box::new)))]
+            source: Box<std::io::Error>,
+        },
+
+        #[snafu(display("Error deserializing pki from configuration at {}: {}", config, source))]
+        ConfigDeserialization {
+            config: String,
+            #[snafu(source(from(toml::de::Error, Box::new)))]
+            source: Box<toml::de::Error>,
         },
 
         #[snafu(display("Unable to decode base64 from certificate '{}': {}", name, source))]
@@ -273,12 +282,6 @@ mod error {
 
         #[snafu(display("Error while reading bundle from file '{}': {}", path.display(), source))]
         ReadSourceBundle { path: PathBuf, source: io::Error },
-
-        #[snafu(display("Error deserializing response from '{}': {}", uri, source))]
-        ResponseJson {
-            uri: String,
-            source: serde_json::Error,
-        },
 
         #[snafu(display("Failed to update trust store: {}", source))]
         UpdateTrustedStore { source: io::Error },
--- a/sources/api/migration/migrations/v1.19.2/certdog-config-file-v0-1-0/src/main.rs
+++ b/sources/api/migration/migrations/v1.19.2/certdog-config-file-v0-1-0/src/main.rs
@@ -1,0 +1,17 @@
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// Add settings for the new certdog-toml config file
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec![
+        "configuration-files.certdog-toml",
+    ]))
+}
+
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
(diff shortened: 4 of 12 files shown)


packages/os/certdog-toml
```diff
@@ -0,0 +1,11 @@
+[required-extensions]
+std = { version = "v1", helpers = ["default"]}
+pki = "v1"
++++
+{{#if settings.pki}}
+{{#each settings.pki}}
+[{{@key}}]
```
bcressey: PKI bundles have the `Identifier` type that doesn't allow characters that aren't valid for TOML keys. However, some of the other types (like `BootConfigKey`) need to be quoted. So as a defensive measure, you could consider doing that here. Otherwise, just something to watch out for in future conversions.
```suggestion
["{{@key}}"]
```

packages/os/certdog-toml
```diff
@@ -0,0 +1,11 @@
...
+trusted = {{this.trusted}}
+data = "{{this.data}}"
+{{/each}}
+{{/if}}
```
bcressey: Can you configure your editor to insert a final newline?

sources/api/certdog/src/main.rs
```diff
@@ -54,30 +57,29 @@ struct CertBundle {
...
+            })?;
+        let pki: HashMap<Identifier, PemCertificate> = toml::from_str(config_str.as_str())
+            .context(error::ConfigDeserializationSnafu {
+                config: config_path.as_ref().to_str().unwrap(),
+            })?;
+
+        split_bundles(pki)
+    } else {
+        split_bundles(HashMap::new())
+    }
```
bcressey: nit: I prefer the early return approach, e.g.:
```
if !config_path.as_ref().exists() {
    return split_bundles(HashMap::new());
}
let config_str = ...
...
split_bundles(pki)
```

That helps avoid some of the rightward drift and keeps the branch logic close to its condition.

sources/models/shared-defaults/defaults.toml
```diff
@@ -162,9 +162,12 @@ affected-services = ["bootstrap-containers"]
 # Certdog
 
 [services.pki]
-configuration-files = []
+configuration-files = ["certdog-toml"]
```
bcressey, line 165: One rule of thumb is that every single edit or addition to a defaults file under `models` requires a corresponding migration.

Here you'll end up needing them for:
* `services.pki.configuration-files` replacement
* `configuration-files.certdog-toml` addition
* `metadata.settings.pki.affected-services` replacement

sources/models/shared-defaults/defaults.toml
```diff
@@ -162,9 +162,12 @@ affected-services = ["bootstrap-containers"]
...
 restart-commands = ["/usr/bin/certdog"]
 
+[metadata.settings.pki]
+affected-services = ["pki"]
```
bcressey: This is already defined in:
* `containerd-cri-pki.toml`
* `docker-pki.toml`

So you'll need to update both of those instead, and conditionalize the migration.

packages/os/certdog-toml
jmt-lab: Sure thing

packages/os/certdog-toml
jmt-lab: Good call out i'll adjust this to be safe

sources/api/certdog/src/main.rs
jmt-lab: Sure can do

sources/models/shared-defaults/defaults.toml
jmt-lab: Should i just remove this from defaults.toml and leave the migration to simply the add of the tempalte file?

bcressey: > Should i just remove this from defaults.toml and leave the migration to simply the add of the tempalte file?

No - you need to update `defaults.toml` so that freshly launched instances with an unpopulated datastore get the change. And you need migrations so that instances that upgrade or downgrade get the new values on upgrade, and the old values on downgrade.

sources/api/certdog/src/main.rs
```diff
@@ -54,30 +57,29 @@ struct CertBundle {
...
-    split_bundles(settings.pki.unwrap_or_default())
+    debug!("Loading certdog configuration");
+    if config_path.as_ref().exists() {
+        let config_str =
+            fs::read_to_string(config_path.as_ref()).context(error::ConfigReadSnafu {
+                config: config_path.as_ref().to_str().unwrap(),
+            })?;
+        let pki: HashMap<Identifier, PemCertificate> = toml::from_str(config_str.as_str())
+            .context(error::ConfigDeserializationSnafu {
+                config: config_path.as_ref().to_str().unwrap(),
```
arnaldo2792: If what you want is the path string, you can try:

```suggestion
                config: format!("{:?}", config_path),
```

With:

```rust
println!("{:?}", &Path::new("My").join("Cool").join("Path"));
```

You will see:

```
"My/Cool/Path"
```

sources/api/certdog/src/main.rs
arnaldo2792: nit: and probably modify `split_bundles` to receive `Option<HashMap<Identifier, PemCertificate>>`, and avoid the extra `HashMap::new` initialization by passing `None`.

sources/models/shared-defaults/defaults.toml
bcressey: Oh, you're right - you can drop this since `certdog` is already invoked as the restart command, and the affected services aren't changing.

sources/api/certdog/src/main.rs
```diff
@@ -36,9 +39,9 @@ struct Args {
     #[argh(option, default = "LevelFilter::Info", short = 'l')]
     /// log-level trace|debug|info|warn|error
     log_level: LevelFilter,
-    #[argh(option, default = "constants::API_SOCKET.to_string()", short = 's')]
-    /// socket-path path to apiserver socket
-    socket_path: String,
+    #[argh(option, default = "DEFAULT_CONFIG_FILE.to_string()", short = 'c')]
+    /// config-path path to apiserver socket
+    config_path: String,
```
bcressey:
```suggestion
    #[argh(option, default = "DEFAULT_CONFIG_FILE.to_string()", short = 'c')]
    /// config-path path to config file
    config_path: String,
```

Can we use an actual `PathBuf` to store the path or is the `String` helpful for some reason?

sources/api/certdog/src/main.rs
```diff
@@ -54,34 +57,34 @@ struct CertBundle {
...
 {
-    debug!("Querying the API for settings");
-
-    let method = "GET";
-    let uri = constants::API_SETTINGS_URI;
-    let (_code, response_body) = apiclient::raw_request(&socket_path, uri, method, None)
-        .await
-        .context(error::APIRequestSnafu { method, uri })?;
+    debug!("Loading certdog configuration");
+    if !config_path.as_ref().exists() {
```
bcressey: nit: can remove some duplication by doing the `as_ref()` call once to get a `Path`:
```suggestion
    debug!("Loading certdog configuration");
    let config_path = config_path.as_ref();
    if !config_path.exists() {
```

sources/api/certdog/src/main.rs
```diff
@@ -54,34 +57,34 @@ struct CertBundle {
...
+        })?;
 
-    split_bundles(settings.pki.unwrap_or_default())
+    split_bundles(Some(pki))
 }
 
 /// Returns a tuple with two lists, for trusted and distrusted certificates
 fn split_bundles(
-    certificates_bundle: HashMap<Identifier, model::PemCertificate>,
+    certificates_bundle: Option<HashMap<Identifier, PemCertificate>>,
```
bcressey: It seems weird to take an `Option` here - why would we call `split_bundles()` at all if we know we have nothing to split?

Maybe have `CertBundle` implement default and then have `get_certificate_bundles()` return `Ok(CertBundle::default())` in the no config case?

sources/api/migration/migrations/v1.18.1/certdog-v0-1-1/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.1"
+authors = ["Jarrett Tierney <jmt@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+# Don't rebuild crate just because of changes to README.
+exclude = ["README.md"]
+
+[dependencies]
+migration-helpers = { path = "../../../migration-helpers", version = "0.1.0" }
```
bcressey: file should end in a newline

sources/api/migration/migrations/v1.18.1/certdog-v0-1-1/src/main.rs
```diff
@@ -0,0 +1,25 @@
...
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec![
+        "services.pki",
+        "configuration-files.certdog-toml",
+    ]))?;
+    migrate(ReplaceListsMigration(vec![ListReplacement {
+        setting: "services.pki.configuration-files",
+        old_vals: &[],
+        new_vals: &["certdog-toml"],
+    }]))
```
bcressey: This migration will need to be split up to fix the multiple migrate calls.

But in case it helps avoid the other extreme (more migrations than strictly necessary) - some of the migration helpers, like `ReplaceListMigrations` here, take a list of inputs rather than only one. That helps to avoid some duplication - you can do all migrations of a specific kind in the same binary, but not two different kinds.

packages/os/certdog-toml
```diff
@@ -0,0 +1,11 @@
+[required-extensions]
+std = { version = "v1", helpers = ["default"]}
+pki = "v1"
```
webern, line 2: What is this extension? Does it need to be modeled as a dependency in RPM spec?

sources/api/certdog/src/main.rs
```diff
@@ -36,9 +39,9 @@ struct Args {
     #[argh(option, default = "LevelFilter::Info", short = 'l')]
     /// log-level trace|debug|info|warn|error
     log_level: LevelFilter,
-    #[argh(option, default = "constants::API_SOCKET.to_string()", short = 's')]
-    /// socket-path path to apiserver socket
-    socket_path: String,
+    #[argh(option, default = "DEFAULT_CONFIG_FILE.to_string()", short = 'c')]
```
webern: This looks weird. I guess this is correctly resolving to the constant+function call and not setting the default to the string literal `DEFAULT_CONFIG_FILE.to_string()`? Did you check this? I don't know if the Bottlerocket call sites are specifying the path or relying on this default.

sources/api/certdog/src/main.rs
```diff
@@ -54,34 +57,34 @@ struct CertBundle {
...
-    debug!("Querying the API for settings");
-
-    let method = "GET";
-    let uri = constants::API_SETTINGS_URI;
-    let (_code, response_body) = apiclient::raw_request(&socket_path, uri, method, None)
-        .await
-        .context(error::APIRequestSnafu { method, uri })?;
+    debug!("Loading certdog configuration");
+    if !config_path.as_ref().exists() {
+        return split_bundles(None);
```
webern: Seems strange to announce the loading of the configuration if it can be skipped in the next line. (I don't know if this is worthy of a warning or if it is fine. If it's a problem should this actually be an error condition instead?)

```suggestion
        warn!("Configuration file does not exist, skipping")
        return split_bundles(None);
```

sources/api/certdog/src/main.rs
```diff
@@ -246,6 +249,20 @@ mod error {
             source: Box<apiclient::Error>,
```
webern: I can't comment on line 245: But the `APIRequest` error is no longer used. One quirk of Snafu's macros is that the compiler can't detect unused Error variants because Snafu is generating code in which they are used. Same with the `ResponseJson` Error variant.

sources/api/migration/migrations/v1.18.1/certdog-v0-1-1/src/main.rs
webern: Yikes, nice catch. I had forgotten about this limitation.

jmt-lab:
Updated per comments
Rebased commits and consolidated work
Split migrations

jmt-lab:
Fix build issue

webern:
> Fix build 

Still has unused imports.

jmt-lab:
Removing unused imports

jmt-lab:
Rebased on latest

Release.toml
```diff
@@ -270,3 +270,7 @@ version = "1.19.1"
     "migrate_v1.19.1_aws-control-container-v0-7-8.lz4",
     "migrate_v1.19.1_public-control-container-v0-7-8.lz4",
 ]
+"(1.19.1, 1.20.0)" = [
+    "migrate_v1.20.0_certdog-config-file-v0-1-0.lz4",
+    "migrate_v1.20.0_certdog-service-cfg-v0-1-0.lz4",
+]
```
webern: Nit: missing newline.

```suggestion
]

```

webern (approved):
We don't actually want to merge this PR with it doing the version bump. The release  should yeild a PR that bumps to 1.20. After that, rebase, then it's ready .

webern:

sources/api/migration/migrations/v1.19.2/certdog-service-cfg-v0-1-0/src/main.rs
```diff
@@ -0,0 +1,19 @@
+use migration_helpers::common_migrations::{ListReplacement, ReplaceListsMigration};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// Add settings for the new certdog-toml config file
+fn run() -> Result<()> {
+    migrate(ReplaceListsMigration(vec![ListReplacement {
+        setting: "services.pki.configuration-files",
```
webern, line 8: Wait. You're adding `services.pki.configuration-files` with an `AddPrefixesMigration` then adding a value here? Isn't this done with Defaults? @bcressey?

sources/api/certdog/src/main.rs
```diff
@@ -210,7 +213,7 @@ async fn run() -> Result<()> {
     SimpleLogger::init(args.log_level, LogConfig::default()).context(error::LoggerSnafu)?;
 
     info!("certdog started");
-    let certificate_bundles = get_certificate_bundles(&args.socket_path).await?;
+    let certificate_bundles = get_certificate_bundles(&args.config_path).await?;
```
bcressey, line 216: nit: the log message below this line still refers to the API

sources/api/migration/migrations/v1.19.2/certdog-config-file-v0-1-0/src/main.rs
```diff
@@ -0,0 +1,18 @@
+use migration_helpers::common_migrations::AddPrefixesMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// Add settings for the new certdog-toml config file
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec![
+        "services.pki",
```
bcressey: This isn't correct - `services.pki` was added a long time ago (in v1.2.0).

`AddPrefixesMigration` could also be written as `RemovePrefixesOnDowngradeMigration` - it's a no-op on upgrade but does the equivalent of `rm -rf` on downgrade. If used incorrectly then previously valid settings may go missing.

sources/api/migration/migrations/v1.19.2/certdog-service-cfg-v0-1-0/src/main.rs
bcressey: @webern this migration looks correct to me but the other one doesn't.

`services.pki.configuration-files` is going from an empty list to a list with one entry in `defaults.toml` - so we need a migration to cover that same change as `storewolf` will not clobber existing settings.

jmt-lab:
Rebased off latest
Removed unneded extension in template
Removed incorrect migration in AddPrefixes
Updated log message to be accurate

bcressey (approved):

webern (approved):

sources/api/certdog/src/main.rs
```diff
@@ -36,9 +39,9 @@ struct Args {
     #[argh(option, default = "LevelFilter::Info", short = 'l')]
     /// log-level trace|debug|info|warn|error
     log_level: LevelFilter,
-    #[argh(option, default = "constants::API_SOCKET.to_string()", short = 's')]
-    /// socket-path path to apiserver socket
-    socket_path: String,
+    #[argh(option, default = "PathBuf::from(DEFAULT_CONFIG_FILE)", short = 'c')]
+    /// config-path path to apiserver socket
```
arnaldo2792: This comment isn't accurate, this path is the path for the configuration not the socket.

PR #3713, opened 2024-01-17
prairiedog: migrate to config file

### LINKED ISSUE #3622: OOTB: Remove model dependency from `prairiedog`
`prariedog` queries the API for all settings, deserializing into the model.  It should probably read from a template rendered by a settings extension.

jmt-lab:
**Issue number:** #3622 

Closes #3622 

**Description of changes:**
* Migrates prairiedog from fetching BootSettings from settings api to fetching this from /etc/prairiedog.toml or config file provided by -c/--config-path
* Adds template prairiedog-toml and migration

**Testing done:**
Fully tested migration upgrade and the reboot logic for the reconcile works. Also made sure kernel settings synced to the file

**Terms of contribution:**

### DIFF
--- a/packages/os/prairiedog-toml
+++ b/packages/os/prairiedog-toml
@@ -1,0 +1,20 @@
+[required-extensions]
+boot = "v1"
++++
+{{#if settings.boot}}
+{{#if settings.boot.reboot-to-reconcile}}
+reboot-to-reconcile = {{settings.boot.reboot-to-reconcile}}
+{{/if}}
+{{#if settings.boot.kernel}}
+[kernel]
+{{#each settings.boot.kernel}}
+"{{@key}}" = [ {{#each this}}"{{{this}}}",{{/each}} ]
+{{/each}}
+{{/if}}
+{{#if settings.boot.init}}
+[init]
+{{#each settings.boot.init}}
+"{{@key}}" = [ {{#each this}}"{{{this}}}",{{/each}} ]
+{{/each}}
+{{/if}}
+{{/if}}
--- a/sources/api/prairiedog/Cargo.toml
+++ b/sources/api/prairiedog/Cargo.toml
@@ -14,13 +14,14 @@ bytes = "1"
 constants = { path = "../../constants", version = "0.1" }
 log = "0.4"
 nix = "0.26"
-models =  { path = "../../models", version = "0.1" }
+modeled-types = { path = "../../models/modeled-types", version = "0.1"}
 schnauzer = { path = "../schnauzer", version = "0.1" }
 signpost = { path = "../../updater/signpost", version = "0.1" }
 simplelog = "0.12"
 snafu = "0.7"
-serde_json = "1"
-tokio = { version = "~1.32", default-features = false, features = ["macros", "rt-multi-thread"] } # LTS
+serde = { version = "1.0", features = ["derive"]}
+serde_json = "1.0"
+toml = "0.8"
 
 [dev-dependencies]
 maplit = "1"
--- a/sources/api/prairiedog/src/bootconfig.rs
+++ b/sources/api/prairiedog/src/bootconfig.rs
@@ -1,13 +1,13 @@
 use crate::error;
 use crate::error::Result;
 use crate::initrd::generate_initrd;
-use model::modeled_types::{BootConfigKey, BootConfigValue};
-use model::BootSettings;
+use modeled_types::{BootConfigKey, BootConfigValue};
+use serde::{Deserialize, Serialize};
 use snafu::{ensure, ResultExt};
 use std::collections::HashMap;
 use std::convert::TryInto;
 use std::path::Path;
-use tokio::io;
+use std::{fs, io};
 
 // Boot config related consts
 const BOOTCONFIG_INITRD_PATH: &str = "/var/lib/bottlerocket/bootconfig.data";
@@ -21,6 +21,27 @@ const DEFAULT_BOOT_SETTINGS: BootSettings = BootSettings {
     kernel_parameters: None,
     init_parameters: None,
 };
+
+#[derive(Deserialize, Serialize, Debug)]
+#[serde(rename_all = "kebab-case")]
+struct BootSettings {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    reboot_to_reconcile: Option<bool>,
+    #[serde(
+        alias = "kernel",
+        rename(serialize = "kernel"),
+        default,
+        skip_serializing_if = "Option::is_none"
+    )]
+    kernel_parameters: Option<HashMap<BootConfigKey, Vec<BootConfigValue>>>,
+    #[serde(
+        alias = "init",
+        rename(serialize = "init"),
+        default,
+        skip_serializing_if = "Option::is_none"
+    )]
+    init_parameters: Option<HashMap<BootConfigKey, Vec<BootConfigValue>>>,
+}
 
 fn append_boot_config_value_list(values: &[BootConfigValue], output: &mut String) {
     for (i, v) in values.iter().enumerate() {
@@ -66,11 +87,11 @@ fn serialize_boot_settings_to_boot_config(boot_settings: &BootSettings) -> Resul
 }
 
 /// Queries Bottlerocket boot settings and generates initrd image file with boot config as the only data
-pub(crate) async fn generate_boot_config<P>(socket_path: P) -> Result<()>
+pub(crate) fn generate_boot_config<P>(config_path: P) -> Result<()>
 where
     P: AsRef<Path>,
 {
-    let bootconfig_bytes = match get_boot_config_settings(socket_path).await? {
+    let bootconfig_bytes = match get_boot_config_settings(config_path)? {
         Some(boot_settings) => {
             info!("Generating initrd boot config from boot settings");
             trace!("Boot settings: {:?}", boot_settings);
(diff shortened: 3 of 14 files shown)


sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -94,23 +95,14 @@ where
...
-async fn get_boot_config_settings<P>(socket_path: P) -> Result<Option<BootSettings>>
+async fn get_boot_config_settings<P>(config_path: P) -> Result<Option<BootSettings>>
 where
     P: AsRef<Path>,
 {
-    let uri = "/settings";
-    let settings: serde_json::Value =
-        schnauzer::v1::get_json(socket_path, uri, Some(("prefix", "boot")))
-            .await
-            .context(error::RetrieveSettingsSnafu)?;
```
cbgbt, line 105: Mind removing the enum variants that aren't used anymore?

packages/os/prairiedog-toml
```diff
@@ -0,0 +1,23 @@
+[required-extensions]
+std = { version = "v1", helpers = ["default"]}
```
cbgbt: Looks like we don't actually use `default`. I think you can remove the dependency on `std`.

sources/api/migration/migrations/v1.18.1/prairiedog-config-file-v0-1-0/src/main.rs
```diff
@@ -0,0 +1,23 @@
+use migration_helpers::common_migrations::{
+    AddPrefixesMigration, ListReplacement, ReplaceListsMigration,
+};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec![
+        "configuration-files.prairiedog-toml",
+    ]))?;
```
cbgbt: Similar to the prairiedog PR, I don't think `AddPrefixesMigration` does the right thing, which means that I think the upgrade case will not actually begin rendering a new config file.

I think we need a new type of Migration (`AddConfigFileMigration`) which allows you to specify the values of the config file in the migration. On upgrade, the new migration is written as specified, and on downgrade it is removed.

cbgbt (commented):
Cool! For some feedback on git style:
* I would squash `prairiedog: fix source number to match with other templates` to be included in the same commit that modified the RPM.
* We tend to leave PR/fixup style comments to the PRs themselves, and the commit messages just focus on the unique change made in that commit. For consistency, it would be best to remove the `fixup!` lines.

sources/api/prairiedog/src/bootconfig.rs
jmt-lab: sure thing

sources/api/migration/migrations/v1.18.1/prairiedog-config-file-v0-1-0/src/main.rs
jmt-lab: This seems to work just fine from my testing

jmt-lab:
squashed commits and updated changes via comments

sources/api/migration/migrations/v1.18.1/prairiedog-config-file-v0-1-0/src/main.rs
webern: Sean's comment concerns me. Can you do some additional testing?
- Use apiclient to change the relevant settings and confirm that the config file is re-rended with the changed settings.
- Confirm that the datastore is in the correct state after downgrade and the config file has been removed after downgrade.

packages/os/prairiedog-toml
```diff
@@ -0,0 +1,22 @@
...
+{{/each}}
+{{/if}}
+
+{{#if settings.boot.init}}
+[init]
+{{#each settings.boot.init}}
+"{{@key}}" = [ {{#each this}}"{{this}}",{{/each}} ]
+{{/each}}
+{{/if}}
+{{/if}}
```
bcressey: nit: can you structure templates like these so they render a completely empty file if there are no corresponding settings, and (ideally) with the correct spacing between the logical sections?

Whitespace issues like that are among my pet peeves when dealing with the output of templating systems. I'd like the output to match what I'd write if I created the file by hand, and as you may have noticed I have feelings about missing or extra newlines.

packages/os/prairiedog-toml
```diff
@@ -0,0 +1,22 @@
...
+{{#if settings.boot}}
+{{#if settings.boot.reboot-to-reconcile}}
+reboot-to-reconcile = {{settings.boot.reboot-to-reconcile}}
+{{/if}}
+
+{{#if settings.boot.kernel}}
+[kernel]
+{{#each settings.boot.kernel}}
+"{{@key}}" = [ {{#each this}}"{{this}}",{{/each}} ]
+{{/each}}
```
bcressey: Does this render correctly when the values contain quotes or spaces?
```
$ apiclient apply <<EOF
settings.boot.kernel-parameters]
param = ["\"with quotes\""]
EOF
```

sources/api/prairiedog/src/main.rs
```diff
@@ -60,9 +62,9 @@ struct Args {
     #[argh(option, default = "LevelFilter::Info", short = 'l')]
     /// log-level trace|debug|info|warn|error
     log_level: LevelFilter,
-    #[argh(option, default = "constants::API_SOCKET.to_string()", short = 's')]
-    /// socket-path path to apiserver socket
-    socket_path: String,
+    #[argh(option, default = "DEFAULT_CONFIG_FILE.to_string()", short = 'c')]
+    /// config-path to provide alternate config file path
+    config_path: String,
```
bcressey: same question as elsewhere: can we use the `PathBuf` type for this?

sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -94,23 +95,14 @@ where
...
-        None => Ok(None),
-        Some(boot_settings_val) => Ok(Some(
-            serde_json::from_value(boot_settings_val.to_owned())
-                .context(error::BootSettingsFromJsonValueSnafu)?,
-        )),
-    }
+    let config_str = fs::read_to_string(config_path.as_ref()).context(error::ReadFileSnafu {
+        path: config_path.as_ref().to_path_buf(),
+    })?;
+    toml::from_str(config_str.as_str()).context(error::InputTomlSnafu)
```
bcressey, line 123: If we're not using await, then the function probably shouldn't be async and we maybe shouldn't even use tokio. On the other hand, if we're using async, we should use `tokio::fs` and await.

Right now this sort of splits the difference and ends up making blocking I/O calls in async functions, which is something we need to be vigilant about.

sources/api/migration/migrations/v1.18.1/prairiedog-config-file-v0-1-0/src/main.rs
```diff
@@ -0,0 +1,23 @@
...
+    migrate(ReplaceListsMigration(vec![ListReplacement {
```
webern: I believe, per Ben in a different PR, we cannot have multiple migrations in a single migration binary.

webern (changes requested):

sources/api/prairiedog/src/bootconfig.rs
jmt-lab: Most of the methods pre-existing in this tool are written with tokio async but without apiserver they technically don't seem to need to be, so the question really here is, would you like me to de-async this or just update my fs to use tokio::fs?

sources/api/migration/migrations/v1.18.1/prairiedog-config-file-v0-1-0/src/main.rs
jmt-lab: Performed the first test (first bullet point) it does indeed update the config file accordingly with the changed settings.

jmt-lab: Performed the upgrade/downgrade test, downgrade works as expected

jmt-lab:
Rebased off latest
Split migrations

jmt-lab:
Adjust spaces in template

packages/os/prairiedog-toml
jmt-lab: Isn't this more of a question for schnauzer than necessarily the template file, or is there a helper method I need to be using for the above?

sources/api/prairiedog/src/bootconfig.rs
jmt-lab: De-asyncified prairiedog

webern (changes requested):
Some lints are failing.

Also, we do not want to bump Release.toml and Twoliter.toml versions as part of this PR. That should be done separately then these PRs should be rebased.

webern:

packages/os/prairiedog-toml
jmt-lab: Yeah i need to change {{this}} to {{{this}}} otherwise it html escapes

jmt-lab:
Rebased off latest
Fixed nested quotes in template
Fixed formatting

sources/api/prairiedog/Cargo.toml
```diff
@@ -20,7 +20,7 @@ signpost = { path = "../../updater/signpost", version = "0.1" }
 simplelog = "0.12"
 snafu = "0.7"
 serde_json = "1"
-tokio = { version = "~1.32", default-features = false, features = ["macros", "rt-multi-thread"] } # LTS
```
webern, line 23: One of the primary purposes of this work is to remove the dependency on `models`. Can you please remove it?

webern (changes requested):
Ben's comments are blocking as well as my request that you remove the dependency on `models`. For future reference, removing the dependency on `models` is the reason we are doing this. If we forgot to do that on the previously merged PR, please do a follow up. Thanks!

jmt-lab:
Models dependency removed

sources/api/prairiedog/Cargo.toml
```diff
@@ -15,12 +15,14 @@ constants = { path = "../../constants", version = "0.1" }
 log = "0.4"
 nix = "0.26"
 models =  { path = "../../models", version = "0.1" }
```
webern: This is the one that needs to be removed.

```suggestion
```

webern (changes requested):
Close, but it still has the models dependency.

sources/api/prairiedog/Cargo.toml
jmt-lab: Fixed

webern (approved):

jmt-lab:
rebased off latest

packages/os/prairiedog-toml
```diff
@@ -0,0 +1,22 @@
...
+reboot-to-reconcile = {{settings.boot.reboot-to-reconcile}}
+
+{{/if}}
+{{#if settings.boot.kernel}}
+[kernel]
+{{#each settings.boot.kernel}}
+"{{@key}}" = [ {{#each this}}"{{{this}}}",{{/each}} ]
+{{/each}}
+
+{{/if}}
```
bcressey: Does this render a config file with two empty lines if none of the settings are present?

bcressey (approved):

PR #3720, opened 2024-01-19
systemd-sysusers: change execution order

arnaldo2792:
**Description of changes:**

`systemd-sysusers` might run while the `/etc` `tmpfs` filesystem is still being set up, which results on failures on the service when it tries to read SELinux configuration files. With this change, `systemd-sysusers` is forced to run after the required SELinux configuration files are in place.

**Testing done:**

- I confirmed `systemd-sysusers` runs after `selinux-policy-files`. 
- There weren't any new warnings when I ran `journalctl -p4`. 
- `systemctl status` shows the system is `running` 

**Terms of contribution:**

### DIFF
--- a/packages/systemd/systemd-sysusers.conf
+++ b/packages/systemd/systemd-sysusers.conf
@@ -1,0 +1,2 @@
+[Unit]
+After=selinux-policy-files.service
--- a/packages/systemd/systemd.spec
+++ b/packages/systemd/systemd.spec
@@ -13,6 +13,7 @@ Source2: systemd-modules-load.conf
 Source3: journald.conf
 Source4: issue
 Source5: systemd-journald.conf
+Source6: systemd-sysusers.conf
 
 # Backport of upstream patches that make the netlink default timeout
 # configurable.  Bottlerocket carries this patch and configures the timeout in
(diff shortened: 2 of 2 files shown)


yeazelm (approved):
LGTM

cbgbt (approved):

packages/systemd/systemd-sysusers.conf
```diff
@@ -0,0 +1,5 @@
+[Unit]
+After=selinux-policy-files.service
+
+[Install]
+WantedBy=local-fs.target
```
bcressey: Do we need to include this? It's not part of the default unit. `local-fs.target` is also pretty late considering that we'll have started all the services that need users created before then.

arnaldo2792: Maybe not? I was confused by when this unit starts if it doesn't have an `Install` section, but probably it isn't needed. I'll test and confirm.

arnaldo2792:
(Forced push removes the unnecessary section)

packages/systemd/systemd-sysusers.conf
arnaldo2792: I confirmed the `Install` section isn't needed, I removed it and I still saw `systemd-sysusers` running after the SELinux policy files were copied.

yeazelm (approved):

bcressey (approved):

PR #3715, opened 2024-01-18
Migrate corndog to use its own config file

jmt-lab:
**Issue number:** #3624 

Closes #3624 

**Description of changes:**
Migrated corndog to read its settings from /etc/corndog.toml which is generated
Added corndog-toml template
Added migrations to ensure configuration file is seen both by services.sysctl and services.lockdown

**Testing done:**

- [x] Ran `cargo +nightly udeps` to ensure all unused dependencies have been removed
- [x] Pass the settings on first boot via userdata. Verify config.
- [x] cat /etc/os-release record your build ID include commit sha, record this
- [x] Use apiclient to update relevant settings at runtime. Verify config was updated correctly.
- [x] Downgrade to older version. 
  - [x] verify connectivity
  - [x] cat /etc/os-release to verify lower version
  - [x] verify that migrations have changed the metadata/settings as expected (i.e. they are gone from the datastore, or lists have been changed back)
- [x] Upgrade to newer version.
  - [x] verify connectivity
  - [x] cat /etc/os-release to verify version
  - [x] verify config file

<details>
<summary>Test Report</summary>

### Test Report:

**Initial_Build_ID.Output**

```json
{
  "os": {
    "arch": "aarch64",
    "build_id": "f6cd87be",
    "pretty_name": "Bottlerocket OS 1.19.3 (aws-ecs-2)",
    "variant_id": "aws-ecs-2",
    "version_id": "1.19.3"
  }
}
```

**Initial_Config.Output**

```toml
lockdown = "integrity"
[sysctl]
"net.ipv4.conf.default.arp_accept" = "0"

```

configuration-files.corndog-toml

```json
{
  "configuration-files": {
    "corndog-toml": {
      "path": "/etc/corndog.toml",
      "template-path": "/usr/share/templates/corndog-toml"
    }
  }
}
```

services.sysctl

```json
{
  "services": {
    "sysctl": {
      "configuration-files": [
        "corndog-toml"
      ],
      "restart-commands": [
        "/usr/bin/corndog sysctl"
      ]
    }
  }
}
```

**Downgrade.Output**

```json
{
  "active_partition": {
    "image": {
      "arch": "aarch64",
      "variant": "aws-ecs-2",
      "version": "1.19.3"
    },
    "next_to_boot": true
  },
  "available_updates": [
    "1.19.3",
    "1.19.2"
  ],
  "chosen_update": {
    "arch": "aarch64",
    "variant": "aws-ecs-2",
    "version": "1.19.2"
  },
[...]

PR #3724, opened 2024-01-22
bootstrap-containers: migrate to using configuration file

### LINKED ISSUE #3626: OOTB: Remove model dependency from `bootstrap-containers`
`bootstrap-containers` queries the API for all settings, deserializing into the model. It should probably read from a template rendered by a settings extension.

jmt-lab:
**Issue number:** #3626 

Closes #3626 

**Description of changes:**
This migrates the bootstrap-containers tool to fetch its settings via a generated toml file.

log:
* add migrations and package changes
* migrate bootstrap-containers to load from config file

**Testing done:**

- [x] Use apiclient to update relevant settings at runtime. Verify config was updated correctly.
  - [x] verify connectivity
  - [x] verify connectivity
  - [x] verify config file

<details>
<summary>Test Report</summary>

**Initial_Build_ID.Output**

```json
{
  "os": {
    "arch": "aarch64",
    "build_id": "bcfb7156",
    "pretty_name": "Bottlerocket OS 1.20.0 (aws-ecs-2)",
    "variant_id": "aws-ecs-2",
    "version_id": "1.20.0"
  }
}
```

**Initial_Config.Output**

```toml
["bear"]
source = "public.ecr.aws/h1i5r3v0/jmt-bottletest:latest"
mode = "off"
user-data = "ypXCt82h4bSlwrfKlA=="

```

configuration-files.bootstrap-containers-toml

```json
{
  "configuration-files": {
    "bootstrap-containers-toml": {
      "path": "/etc/bootstrap-containers/bootstrap-containers.toml",
      "template-path": "/usr/share/templates/bootstrap-containers-toml"
    }
  }
}
```

services.bootstrap-containers

```json
{
  "services": {
    "bootstrap-containers": {
      "configuration-files": [
        "host-ctr-toml",
        "bootstrap-containers-toml"
      ],
      "restart-commands": [
        "/usr/bin/bootstrap-containers create-containers"
      ]
    }
  }
}
```

**Downgrade.Output**

```json
{
  "active_partition": {
    "image": {
      "arch": "aarch64",
      "variant": "aws-ecs-2",
      "version": "1.19.3"
    },
    "next_to_boot": true
  },
  "available_updates": [
    "1.19.3",
    "1.19.2"
  ],
  "chosen_update": {
    "arch": "aarch64",
    "variant": "aws-ecs-2",
    "version": "1.19.2"
  },
  "most_recent_command": {
    "cmd_status": "Success",
    "cmd_type": "refresh",
    "exit_status": 0,
    "stderr": "",
    "timestamp": "2024-03-18T20:24:18.772874378Z"
  },
  "staging_partition": null,
  "update_state": "Available"
}
```

```
20:24:17 [INFO] Refreshing updates...
20:24:18 [INFO] Downloading and applying update to disk...
20:24:23 [INFO] Still waiting for updated status, will wait up to 594.5s longer...
20:24:28 [INFO] Still waiting for updated status, will wait up to 589.5s longer...
20:24:33 [INFO] Still waiting for updated status, will wait up to 584.5s longer...
20:24:36 [INFO] Setting the update active so it will apply on the next reboot...
20:24:37 [INFO] Rebooting, goodbye...
```

**Downgrade_Build_ID.Output**

```json
{
  "os": {
    "arch": "aarch64",
    "build_id": "f097c617",
    "pretty_name": "Bottlerocket OS 1.19.3 (aws-ecs-2)",
    "variant_id": "aws-ecs-2",
    "version_id": "1.19.3"
  }
}
```

**Downgrade_Config.Output**

```
cat: /etc/bootstrap-containers/bootstrap-containers.toml: No such file or directory
```

configuration-files.bootstrap-containers-toml

```
{}
```

services.bootstrap-containers

```json
{
  "services": {
    "bootstrap-containers": {
      "configuration-files": [
        "host-ctr-toml"
      ],
      "restart-commands": [
        "/usr/bin/bootstrap-containers create-containers"
      ]
    }
  }
}
```

**Upgrade.Output**

```json
{
  "active_partition": {
    "image": {
      "arch": "aarch64",
      "variant": "aws-ecs-2",
      "version": "1.19.3"
    },
    "next_to_boot": true
  },
  "available_updates": [
    "1.20.0",
    "1.19.3"
  ],
  "chosen_update": {
    "arch": "aarch64",
    "variant": "aws-ecs-2",
    "version": "1.20.0"
  },
  "most_recent_command": {
    "cmd_status": "Success",
    "cmd_type": "refresh",
    "exit_status": 0,
    "stderr": "",
    "timestamp": "2024-03-28T20:08:53.656541162Z"
  },
  "staging_partition": null,
  "update_state": "Available"
}
```

```
20:08:52 [INFO] Refreshing updates...
20:08:53 [INFO] Downloading and applying update to disk...
20:08:58 [INFO] Still waiting for updated status, will wait up to 594.5s longer...
20:09:03 [INFO] Still waiting for updated status, will wait up to 589.5s longer...
20:09:08 [INFO] Still waiting for updated status, will wait up to 584.5s longer...
20:09:13 [INFO] Still waiting for updated status, will wait up to 579.5s longer...
20:09:14 [INFO] Setting the update active so it will apply on the next reboot...
20:09:15 [INFO] Rebooting, goodbye...
```

**Upgrade_Build_ID.Output**

```json
{
  "os": {
    "arch": "aarch64",
    "build_id": "bcfb7156",
    "pretty_name": "Bottlerocket OS 1.20.0 (aws-ecs-2)",
    "variant_id": "aws-ecs-2",
    "version_id": "1.20.0"
  }
}
```

**Upgrade_Config.Output**

```toml
["bear"]
source = "public.ecr.aws/h1i5r3v0/jmt-bottletest:latest"
mode = "off"
user-data = "ypXCt82h4bSlwrfKlA=="
```

configuration-files.bootstrap-containers-toml

```json
{
  "configuration-files": {
    "bootstrap-containers-toml": {
      "path": "/etc/bootstrap-containers/bootstrap-containers.toml",
      "template-path": "/usr/share/templates/bootstrap-containers-toml"
    }
  }
}
```

services.bootstrap-containers

```json
{
  "services": {
    "bootstrap-containers": {
      "configuration-files": [
        "host-ctr-toml",
        "bootstrap-containers-toml"
      ],
      "restart-commands": [
        "/usr/bin/bootstrap-containers create-containers"
      ]
    }
  }
}
```

**Change_Settings.Output**

```bash
apiclient set --json '{"bootstrap-containers": { "bear": { "user-data": "SGVsbG8gZnJvbSBib290c3RyYXAtY29udGFpbmVycw==" } } }'
```

```toml
["bear"]
source = "public.ecr.aws/h1i5r3v0/jmt-bottletest:latest"
mode = "off"
user-data = "SGVsbG8gZnJvbSBib290c3RyYXAtY29udGFpbmVycw=="

```

</details>

**Terms of contribution:**

### DIFF
--- a/packages/os/bootstrap-containers-toml
+++ b/packages/os/bootstrap-containers-toml
@@ -1,0 +1,17 @@
+[required-extensions]
+bootstrap-containers = "v1"
+std = { version = "v1", helpers = ["if_not_null"] }
++++
+{{#if_not_null settings.bootstrap-containers}}
+{{#each settings.bootstrap-containers}}
+["{{@key}}"]
+source = "{{{this.source}}}"
+mode = "{{{this.mode}}}"
+{{#if_not_null this.user-data}}
+user-data = "{{{this.user-data}}}"
+{{/if_not_null}}
+{{#if_not_null this.essential}}
+essential = {{this.essential}}
+{{/if_not_null}}
+{{/each}}
+{{/if_not_null}}
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -35,6 +35,7 @@ Source14: certdog-toml
 Source15: prairiedog-toml
 Source16: thar-be-updates-toml
 Source17: corndog-toml
+Source18: bootstrap-containers-toml
 Source19: host-containers-toml
 
 # 1xx sources: systemd units
@@ -477,7 +478,7 @@ install -d %{buildroot}%{_cross_datadir}/updog
 install -p -m 0644 %{_cross_repo_root_json} %{buildroot}%{_cross_datadir}/updog
 
 install -d %{buildroot}%{_cross_templatedir}
-install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{S:14} %{S:15} %{S:16} %{S:17} %{S:19} \
+install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{S:14} %{S:15} %{S:16} %{S:17} %{S:18} %{S:19} \
   %{buildroot}%{_cross_templatedir}
 
 install -d %{buildroot}%{_cross_unitdir}
@@ -700,6 +701,7 @@ install -p -m 0644 %{S:400} %{S:401} %{S:402} %{buildroot}%{_cross_licensedir}
 %{_cross_bindir}/bootstrap-containers
 %{_cross_unitdir}/bootstrap-containers@.service
 %{_cross_tmpfilesdir}/bootstrap-containers.conf
+%{_cross_templatedir}/bootstrap-containers-toml
 
 %files -n %{_cross_os}bloodhound
 %{_cross_bindir}/bloodhound
--- a/sources/api/bootstrap-containers/src/main.rs
+++ b/sources/api/bootstrap-containers/src/main.rs
@@ -74,7 +74,7 @@ journalctl -u bootstrap-containers@bear.service
 extern crate log;
 
 use base64::Engine;
-use datastore::{serialize_scalar, Key, KeyType};
+use serde::{Deserialize, Serialize};
 use simplelog::{Config as LogConfig, LevelFilter, SimpleLogger};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::HashMap;
@@ -83,29 +83,43 @@ use std::env;
 use std::ffi::OsStr;
 use std::fmt::Write;
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::{self, Command};
 use std::str::FromStr;
 
-use model::modeled_types::{BootstrapContainerMode, Identifier};
+use modeled_types::{BootstrapContainerMode, Identifier, Url, ValidBase64};
 
 const ENV_FILE_DIR: &str = "/etc/bootstrap-containers";
 const DROPIN_FILE_DIR: &str = "/etc/systemd/system";
 const PERSISTENT_STORAGE_DIR: &str = "/local/bootstrap-containers";
 const DROP_IN_FILENAME: &str = "overrides.conf";
+const DEFAULT_CONFIG_PATH: &str = "/etc/bootstrap-containers/bootstrap-containers.toml";
+
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+struct BootstrapContainer {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    source: Option<Url>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    mode: Option<BootstrapContainerMode>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    user_data: Option<ValidBase64>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    essential: Option<bool>,
+}
 
 /// Stores user-supplied global arguments
 #[derive(Debug)]
 struct Args {
     log_level: LevelFilter,
-    socket_path: String,
+    config_path: PathBuf,
 }
 
 impl Default for Args {
     fn default() -> Self {
         Self {
             log_level: LevelFilter::Info,
-            socket_path: constants::API_SOCKET.to_string(),
+            config_path: PathBuf::from_str(DEFAULT_CONFIG_PATH).unwrap(),
         }
     }
 }
@@ -134,16 +148,15 @@ fn usage() {
         mark-bootstrap
 
     Global arguments:
-        [ --socket-path PATH ]
+        [ --config-path PATH ]
         [ --log-level trace|debug|info|warn|error ]
 
     Mark bootstrap arguments:
         --container-id CONTAINER-ID
         --mode MODE
 
-    Socket path defaults to {}",
-        program_name,
-        constants::API_SOCKET,
+    Config path defaults to {}",
+        program_name, DEFAULT_CONFIG_PATH,
     );
 }
 
@@ -165,10 +178,11 @@ fn parse_args(args: env::Args) -> Result<(Args, Subcommand)> {
                     .context(error::LogLevelSnafu { log_level })?;
             }
 
-            "-s" | "--socket-path" => {
-                global_args.socket_path = iter.next().context(error::UsageSnafu {
-                    message: "Did not give argument to --socket-path",
-                })?
+            "-c" | "--config-path" => {
+                let config_str = iter.next().context(error::UsageSnafu {
+                    message: "Did not give argument to --config-path",
+                })?;
+                global_args.config_path = PathBuf::from(config_str.as_str());
             }
 
             // Subcommands
@@ -242,10 +256,7 @@ fn parse_mark_bootstrap_args(args: Vec<String>) -> Result<Subcommand> {
 }
 
 /// Handles how the bootstrap containers' systemd units are created
-fn handle_bootstrap_container<S>(
-    name: S,
-    container_details: &model::BootstrapContainer,
-) -> Result<()>
+fn handle_bootstrap_container<S>(name: S, container_details: &BootstrapContainer) -> Result<()>
 where
     S: AsRef<str>,
 {
@@ -382,27 +393,17 @@ where
 }
 
 /// Query the API for the currently defined bootstrap containers
-async fn get_bootstrap_containers<P>(
-    socket_path: P,
-) -> Result<HashMap<Identifier, model::BootstrapContainer>>
+fn get_bootstrap_containers<P>(config_path: P) -> Result<HashMap<Identifier, BootstrapContainer>>
 where
     P: AsRef<Path>,
 {
-    debug!("Querying the API for settings");
+    let config_str = fs::read_to_string(config_path.as_ref()).context(error::ReadConfigSnafu {
+        config_path: config_path.as_ref(),
+    })?;
 
-    let method = "GET";
-    let uri = constants::API_SETTINGS_URI;
-    let (_code, response_body) = apiclient::raw_request(&socket_path, uri, method, None)
-        .await
-        .context(error::APIRequestSnafu { method, uri })?;
-
-    // Build a Settings struct from the response string
-    debug!("Deserializing response");
-    let settings: model::Settings =
-        serde_json::from_str(&response_body).context(error::ResponseJsonSnafu { method, uri })?;
-
-    // If bootstrap containers aren't defined, return an empty map
-    Ok(settings.bootstrap_containers.unwrap_or_default())
+    toml::from_str(&config_str).context(error::DeserializationSnafu {
+        config_path: config_path.as_ref(),
+    })
 }
 
 /// SystemdUnit stores the systemd unit being manipulated
@@ -492,12 +493,12 @@ where
 }
 
 /// Handles the `create-containers` subcommand
-async fn create_containers<P>(socket_path: P) -> Result<()>
+fn create_containers<P>(config_path: P) -> Result<()>
 where
     P: AsRef<Path>,
 {
     let mut failed = 0usize;
-    let bootstrap_containers = get_bootstrap_containers(socket_path).await?;
+    let bootstrap_containers = get_bootstrap_containers(config_path)?;
     for (name, container_details) in bootstrap_containers.iter() {
         // Continue to handle other bootstrap containers if we fail one
         if let Err(e) = handle_bootstrap_container(name, container_details) {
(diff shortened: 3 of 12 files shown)


sources/api/bootstrap-containers/src/main.rs
```diff
@@ -381,26 +393,18 @@ where
...
-    let uri = constants::API_SETTINGS_URI;
-    let (_code, response_body) = apiclient::raw_request(&socket_path, uri, method, None)
-        .await
-        .context(error::APIRequestSnafu { method, uri })?;
-
-    // Build a Settings struct from the response string
-    debug!("Deserializing response");
-    let settings: model::Settings =
-        serde_json::from_str(&response_body).context(error::ResponseJsonSnafu { method, uri })?;
+    let config_str = fs::read_to_string(config_path.as_ref()).context(error::ReadConfigSnafu {
```
arnaldo2792, line 400: You could read a `slice` and save yourself the extra `String` transformation.

sources/api/bootstrap-containers/src/main.rs
```diff
@@ -133,16 +136,19 @@ fn usage() {
         mark-bootstrap
 
     Global arguments:
+        [ --config-path PATH ]
         [ --socket-path PATH ]
```
webern: Is the --socket-path still required? If not we should either 

a) note that it is deprecated and give a warning if the command line arg is used

or, preferably

b) remove the argument from the cli

Edit: Oh I see it still used. Probably to mount it into the container it is starting.

webern (commented):
Isn't there is systemd unit that needs to change in order to pass the new argument?

packages/os/os.spec
```diff
@@ -31,6 +31,7 @@ Source12: 00-resolved.conf
 %if %{with k8s_runtime}
 Source13: cis-checks-k8s-metadata-json
 %endif
+Source18: bootstrap-containers-toml
```
bcressey, line 38: The weird numbering here is because of additional configuration file pull requests in flight?

sources/api/bootstrap-containers/src/main.rs
```diff
@@ -92,18 +92,21 @@ const ENV_FILE_DIR: &str = "/etc/bootstrap-containers";
 const DROPIN_FILE_DIR: &str = "/etc/systemd/system";
 const PERSISTENT_STORAGE_DIR: &str = "/local/bootstrap-containers";
 const DROP_IN_FILENAME: &str = "overrides.conf";
+const DEFAULT_CONFIG_PATH: &str = "/etc/bootstrap-containers.toml";
```
bcressey:
```suggestion
const DEFAULT_CONFIG_PATH: &str = "/etc/bootstrap-containers/bootstrap-containers.toml";
```

And then add a mount unit like `packages/host-ctr/etc-host-containers.mount.in` to ensure that `/etc/bootstrap-containers` is labeled as `secret_t`.

packages/os/bootstrap-containers-toml
```diff
@@ -0,0 +1,16 @@
...
+bootstrap-containers = "v1"
++++
+{{#if settings.bootstrap-containers}}
+{{#each settings.bootstrap-containers}}
+["{{@key}}"]
+source = "{{this.source}}"
+mode = "{{this.mode}}"
+{{#if this.user-data}}
+user-data = "{{this.user-data}}"
+{{/if}}
```
bcressey: User data is opaque so we assume it may contain secrets; it must be rendered to a part of `/etc` with a special label that doesn't allow read access to unprivileged containers.

sources/api/migration/migrations/v1.18.1/bootstrap-containers-config-file-v0-1-0/src/main.rs
```diff
@@ -0,0 +1,23 @@
...
+fn run() -> Result<()> {
+    migrate(AddPrefixesMigration(vec![
+        "configuration-files.bootstrap-containers-toml",
+    ]))?;
+    migrate(ReplaceListsMigration(vec![ListReplacement {
+        setting: "services.bootstrap-containers.configuration-files",
+        old_vals: &["host-ctr-toml"],
+        new_vals: &["host-ctr-toml", "bootstrap-containers-toml"],
+    }]))
+}
```
bcressey: Need to split this into two migrations to avoid a double `migrate` call (per https://github.com/bottlerocket-os/bottlerocket/pull/3715#discussion_r1466525851).

sources/models/shared-defaults/defaults.toml
```diff
@@ -153,12 +153,16 @@ affected-services = ["lockdown"]
...
 [services.bootstrap-containers]
-configuration-files = ["host-ctr-toml"]
+configuration-files = ["host-ctr-toml", "bootstrap-containers-toml"]
 restart-commands = ["/usr/bin/bootstrap-containers create-containers"]
 
 [metadata.settings.bootstrap-containers]
 affected-services = ["bootstrap-containers"]
 
+[configuration-files.bootstrap-containers-toml]
+path = "/etc/bootstrap-containers.toml"
```
bcressey:
```suggestion
path = "/etc/bootstrap-containers/bootstrap-containers.toml"
```

packages/os/os.spec
jmt-lab: Correct

webern:

packages/os/bootstrap-containers-toml
```diff
@@ -0,0 +1,16 @@
+[required-extensions]
+bootstrap-containers = "v1"
++++
+{{#if settings.bootstrap-containers}}
+{{#each settings.bootstrap-containers}}
+["{{@key}}"]
+source = "{{this.source}}"
+mode = "{{this.mode}}"
+{{#if this.user-data}}
+user-data = "{{this.user-data}}"
```
sam-berning: this will URL-escape the special characters, most importantly the `==` at the end of the base64 string. In order to avoid this, you need to format it with triple curly braces: https://handlebarsjs.com/guide/#html-escaping

(ran into this while testing host-containers 😄)

```suggestion
user-data = "{{{this.user-data}}}"
```

edit: fixed bad doc link

jmt-lab:
rebase on latest

jmt-lab:
Finished changes and testing run

webern (approved):
Verified this commit from the testing report `65bc4fc7`. :+1:

rpkelly (approved):

packages/os/bootstrap-containers-toml
```diff
@@ -0,0 +1,16 @@
...
+{{#each settings.bootstrap-containers}}
+["{{@key}}"]
+source = "{{this.source}}"
+mode = "{{this.mode}}"
+{{#if this.user-data}}
+user-data = "{{{this.user-data}}}"
+{{/if}}
+{{#if this.essential}}
+essential = {{this.essential}}
+{{/if}}
```
bcressey: Boolean value. Not sure if we can use `{{ default false this.essential }}` - if not then then let's add an `else` block.

sources/api/bootstrap-containers/src/main.rs
```diff
@@ -92,18 +92,21 @@ const ENV_FILE_DIR: &str = "/etc/bootstrap-containers";
...
     log_level: LevelFilter,
+    config_path: PathBuf,
     socket_path: String,
 }
 
 impl Default for Args {
     fn default() -> Self {
         Self {
             log_level: LevelFilter::Info,
+            config_path: PathBuf::from_str(DEFAULT_CONFIG_PATH).unwrap(),
```
bcressey, line 122: We need to clean up the other use of `socket_path` in `mark_bootstrap` - the larger goal of this series of changes is to remove the datastore dependency from these programs entirely.

It's going to be a bit more involved. What I'd suggest trying is calling `apiclient set bootstrap-containers.%i=off` instead. Not sure that will work but maybe.

sources/api/bootstrap-containers/src/main.rs
```diff
@@ -381,26 +394,20 @@ where
...
 ) -> Result<HashMap<Identifier, model::BootstrapContainer>>
 where
     P: AsRef<Path>,
 {
-    debug!("Querying the API for settings");
-
-    let method = "GET";
-    let uri = constants::API_SETTINGS_URI;
-    let (_code, response_body) = apiclient::raw_request(&socket_path, uri, method, None)
+    let config_str = tokio::fs::read_to_string(config_path.as_ref())
```
bcressey: When the API dependency is fully removed you should be able to drop the tokio dependency and all uses of async/await.

sources/api/bootstrap-containers/src/main.rs
jmt-lab: Doesn't this still not solve the problem? Since the apiclient call and apiclient set commands would essentially be identical operations and dependency only difference is calling the apiclient binary instead of direct sending the request?

jmt-lab:
Addressed comments

packages/os/bootstrap-containers-toml
```diff
@@ -0,0 +1,16 @@
+[required-extensions]
+bootstrap-containers = "v1"
++++
+{{#if settings.bootstrap-containers}}
+{{#each settings.bootstrap-containers}}
```
arnaldo2792, line 6: Wouldn't it be more readable (and probably more maintainable in the long run) to have this rendered by a handlebars helper? e.g.

```hbs
{{bootstrap-containers settings.bootstrap-containers}}
```

Then you can just use Rust and strong typing to generate the toml tables. . I want us to avoid another "generate on-the-flight complex structures" and use the Rust to guarantee the correctness of the serialized format. (@bcressey what do you think?)

sources/api/bootstrap-containers/src/main.rs
```diff
@@ -518,38 +519,33 @@ where
...
-        let mut map = HashMap::new();
-        map.insert(key, value);
-        let settings: model::Settings = datastore::deserialization::from_map(&map)
-            .context(error::SettingsDeserializeSnafu { settings: map })?;
-
+        let formatted = format!("settings.bootstrap-containers.{}.mode=off", container_id);
         info!("Turning off container '{}'", container_id);
-        apiclient::set::set(socket_path, &settings)
-            .await
+        let status = Command::new("apiclient")
```
arnaldo2792: There is a helper method in this crate `command`, wouldn't it work for your case?

packages/os/bootstrap-containers-toml
sam-berning: I like this approach, this could just be a generic `settings-to-toml` helper which we could use for all of these PRs where we substitute reading from the API for reading from a config file

webern (changes requested):
Remove bear.toml

jmt-lab:
* Adjusted template to use new if_not_null helper
* Adjusted call to apiclient to use command() helper

Still need to remove tokio

jmt-lab:
removed tokio

jmt-lab:
Rebased off latest and moved migrations to 1.20.0 will rerun the tests to be safe

jmt-lab:
Release.toml was wrong resulting in migrations not occuring, fixed this

packages/os/bootstrap-containers-toml
bcressey: > Wouldn't it be more readable (and probably more maintainable in the long run) to have this rendered by a handlebars helper?

It doesn't really help in the long run where the goal is to decouple templates and settings extensions so they can be maintained independently.

> I like this approach, this could just be a generic settings-to-toml helper which we could use for all of these PRs where we substitute reading from the API for reading from a config file

I wasn't too keen on `settings-to-json` when [that put in an appearance](https://github.com/bottlerocket-os/bottlerocket/pull/3834#discussion_r1534882924) recently.

I am a little more positively inclined on something like `settings-to-toml` if the helper is producing a 1:1 match with a subset of the external system settings API, if only because that API is covered by a strong covenant with the end user that it won't change.

However I'm still not very enthusiastic. I'd like most of these agents to be able to be configured in ways that aren't 100% reflected in the external API - like `kubelet` and `ecs-agent` are not - since some behaviors may make sense to hard-code in certain variants.

Ultimately an agent whose config is nothing but the external API is in a precarious position: it makes more sense to me to have use a combination of Bottlerocket features and systemd functionality to dispense with that agent entirely (as with `ecs-settings-applier` in the linked PR) where we can, than to add helpers which special-case just that agent rather than answering a more general need.

sources/api/bootstrap-containers/src/main.rs
```diff
@@ -520,38 +521,24 @@ where
...
-        map.insert(key, value);
-        let settings: model::Settings = datastore::deserialization::from_map(&map)
-            .context(error::SettingsDeserializeSnafu { settings: map })?;
-
+        let formatted = format!("settings.bootstrap-containers.{}.mode=off", container_id);
         info!("Turning off container '{}'", container_id);
-        apiclient::set::set(socket_path, &settings)
-            .await
-            .context(error::SetSnafu)?;
+        command("apiclient", ["set", formatted.as_str()])?;
```
bcressey, line 535: Can you verify that this actually works as expected? I.e. a bootstrap container that is set to "once" mode only runs once on first boot and not multiple times - either on the same boot (which can happen when the systemd target changes) or on subsequent boots.

I'd suggest a bootstrap container that runs something like:
```
#!/bin/bash
TS="$(date +%s)"
apiclient get settings.bootstrap-containers >
  /.bottlerocket/rootfs/local/${TS}.txt
```

And confirm that only one file is created and it shows the expected value.

bcressey (approved):
LGTM with one additional testing callout.

webern (approved):

sources/api/bootstrap-containers/src/main.rs
jmt-lab: Tested and it only returned one file, logs also show correct behavior

Issue #3745, opened 2024-01-31
Sandbox container image being GC'd in 1.29


z0rc:
**Image I'm using:**
Bottlerocket AMI, version 1.18.0, variant aws-k8s-1.29

**What I expected to happen:**
Pods should be able to run on long running nodes (uptime > 12h).

**What actually happened:**
Pods fail to start. Kubernetes Event:
```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox image "602401143452.dkr.ecr.us-east-2.amazonaws.com/eks/pause:3.1-eksbuild.1": failed to pull image "602401143452.dkr.ecr.us-east-2.amazonaws.com/eks/pause:3.1-eksbuild.1": failed to pull and unpack image "602401143452.dkr.ecr.us-east-2.amazonaws.com/eks/pause:3.1-eksbuild.1": failed to resolve reference "602401143452.dkr.ecr.us-east-2.amazonaws.com/eks/pause:3.1-eksbuild.1": pull access denied, repository does not exist or may require authorization: authorization failed: no basic auth credentials
```

**How to reproduce the problem:**
Launch EKS 1.29 cluster with bottlerocket nodes. Let it run for some time, until kubelet garbage collects `pause` image. Time can be reduced by increased image churn, where a lot of pods launched and stopped using different images, like CI runners for example.

Effectively it's same as https://github.com/awslabs/amazon-eks-ami/issues/1597. Reproducible in the same way on Bottlerocket nodes.

gthao313:
@z0rc Thanks for reporting this issue! We're able to reproduce the issue and currently we are working on the fix.

ecpullen:
Thanks for bringing us this issue. We have been able to reproduce this error and we are working to develop a fix.

gthao313:
@z0rc Thanks for opening this issue! we're able to reproduce the issue and work on the fix now!

bcressey:
For anyone affected, @cartermckinnon shared a [workaround](https://github.com/awslabs/amazon-eks-ami/issues/1597#issuecomment-1919735779) of overriding the sandbox image URI to one that doesn't require credentials.

This can be done through user-data settings:
```
[settings.kubernetes]
pod-infra-container-image = "registry.k8s.io/pause:3.9"
```

Or this:
```
[settings.kubernetes]
pod-infra-container-image = "public.ecr.aws/eks-distro/kubernetes/pause:v1.29.0-eks-1-29-latest"
```

[...]

PR #3757, opened 2024-02-05
host-ctr: add label flag to host-ctr pull-image.

gthao313:
**Issue number:**
#3745 
Closes #

**Description of changes:**
Sandbox container image on k8s 1.29 isn't pinned and causea the Sandbox container image would be GC. we add label flag to host-ctr pull-image and pass "io.cri-containerd.pinned=pinned" label to all 1.29 varaints. Then the Sandbox contianer image will be pinned and avoid GC issue.

**Testing done:**

- [x] Launch the 1.29 nodes to 1.29 cluster and check if the pause image is pinned
```
602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause             3.1-eksbuild.1       106a8e54d5eb3       299kB
[root@admin]# /crictl --runtime-endpoint unix:///.bottlerocket/rootfs/run/containerd/containerd.sock inspecti 106a8e54d5eb3 | grep pinned
    "pinned": true
```
- [x] conformance test for 1.29 nodes
AWS
```
[cargo-make][1] INFO - Running Task: testsys
 NAME                                           TYPE                STATE                         PASSED            FAILED            SKIPPED   BUILD ID            LAST UPDATE
 x86-64-aws-k8s-129-conformance-test            Test                passed                           392                 0               7019                       2024-02-05T03:18:41Z
 x86-64-aws-k8s-129-instances                   Resource            completed                                                                                       2024-02-05T01:25:21Z
```
VMWARE - a Known kubernetes flake on conformance test.
```
Plugin: e2e
Status: failed
Total: 7411
Passed: 391
Failed: 1
Skipped: 7019

Failed tests:
 [sig-network] Services should serve endpoints on same port and different protocols [Conformance]
```

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.29/prestart-pull-pause-ctr.conf
+++ b/packages/kubernetes-1.29/prestart-pull-pause-ctr.conf
@@ -6,4 +6,5 @@ ExecStartPre=/usr/bin/host-ctr \
     pull-image \
     --source=${POD_INFRA_CONTAINER_IMAGE} \
     --registry-config=/etc/host-containers/host-ctr.toml \
-    --skip-if-image-exists=true
+    --skip-if-image-exists=true \
+    --label="io.cri-containerd.pinned=pinned"
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -39,6 +39,11 @@ import (
 // Example 2: 777777777777.dkr.ecr.cn-north-1.amazonaws.com.cn/my_image:latest
 var ecrRegex = regexp.MustCompile(`(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?.*`)
 
+const (
+	// The maximum size of an	image label.
+	imageLabelMaxSize = 4096
+)
+
 func init() {
 	rand.New(rand.NewSource(time.Now().UnixNano()))
 	// Dispatch logging output instead of writing all levels' messages to
@@ -161,9 +166,18 @@ func App() *cli.App {
 					Destination: &useCachedImage,
 					Value:       false,
 				},
+				&cli.StringSliceFlag{
+					Name:  "label",
+					Usage: "label to add to the pulled image in `key=value` format",
+				},
 			},
 			Action: func(c *cli.Context) error {
-				return pullImageOnly(containerdSocket, namespace, source, registryConfig, useCachedImage)
+				labels := c.StringSlice("label")
+				labelsMap, err := convertLabels(labels)
+				if err != nil {
+					return err
+				}
+				return pullImageOnly(containerdSocket, namespace, source, registryConfig, useCachedImage, labelsMap)
 			},
 		},
 		{
@@ -274,13 +288,15 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 	// Check if the image source is an ECR image. If it is, then we need to handle it with the ECR resolver.
 	isECRImage := ecrRegex.MatchString(source)
 	var img containerd.Image
+	emptyLabels := make(map[string]string)
+
 	if isECRImage {
-		img, err = fetchECRImage(ctx, source, client, registryConfigPath, useCachedImage)
+		img, err = fetchECRImage(ctx, source, client, registryConfigPath, useCachedImage, emptyLabels)
 		if err != nil {
 			return err
 		}
 	} else {
-		img, err = fetchImage(ctx, source, client, registryConfigPath, useCachedImage)
+		img, err = fetchImage(ctx, source, client, registryConfigPath, useCachedImage, emptyLabels)
 		if err != nil {
 			log.G(ctx).WithField("ref", source).Error(err)
 			return err
@@ -494,7 +510,7 @@ func runCtr(containerdSocket string, namespace string, containerID string, sourc
 }
 
 // pullImageOnly pulls the specified container image
-func pullImageOnly(containerdSocket string, namespace string, source string, registryConfigPath string, useCachedImage bool) error {
+func pullImageOnly(containerdSocket string, namespace string, source string, registryConfigPath string, useCachedImage bool, labels map[string]string) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	ctx = namespaces.WithNamespace(ctx, namespace)
@@ -508,12 +524,12 @@ func pullImageOnly(containerdSocket string, namespace string, source string, reg
 	// Check if the image source is an ECR image. If it is, then we need to handle it with the ECR resolver.
 	isECRImage := ecrRegex.MatchString(source)
 	if isECRImage {
-		_, err = fetchECRImage(ctx, source, client, registryConfigPath, useCachedImage)
+		_, err = fetchECRImage(ctx, source, client, registryConfigPath, useCachedImage, labels)
 		if err != nil {
 			return err
 		}
 	} else {
-		_, err = fetchImage(ctx, source, client, registryConfigPath, useCachedImage)
+		_, err = fetchImage(ctx, source, client, registryConfigPath, useCachedImage, labels)
 		if err != nil {
 			log.G(ctx).WithField("ref", source).Error(err)
 			return err
@@ -646,7 +662,7 @@ func fetchECRRef(ctx context.Context, input string) (ecr.ECRSpec, error) {
 }
 
 // fetchECRImage does some additional conversions before resolving the image reference and fetches the image.
-func fetchECRImage(ctx context.Context, source string, client *containerd.Client, registryConfigPath string, fetchCachedImageIfExist bool) (containerd.Image, error) {
+func fetchECRImage(ctx context.Context, source string, client *containerd.Client, registryConfigPath string, fetchCachedImageIfExist bool, labels map[string]string) (containerd.Image, error) {
 	ecrRef, err := fetchECRRef(ctx, source)
 	if err != nil {
 		return nil, err
@@ -658,7 +674,7 @@ func fetchECRImage(ctx context.Context, source string, client *containerd.Client
 		WithField("source", source).
 		Debug("parsed ECR reference from URI")
 
-	img, err := fetchImage(ctx, ref, client, registryConfigPath, fetchCachedImageIfExist)
+	img, err := fetchImage(ctx, ref, client, registryConfigPath, fetchCachedImageIfExist, labels)
 	if err != nil {
 		log.G(ctx).WithField("ref", ref).Error(err)
 		return nil, err
@@ -971,7 +987,7 @@ func withProxyEnv() oci.SpecOpts {
 }
 
 // fetchImage returns a `containerd.Image` given an image source.
-func fetchImage(ctx context.Context, source string, client *containerd.Client, registryConfigPath string, useCachedImage bool) (containerd.Image, error) {
+func fetchImage(ctx context.Context, source string, client *containerd.Client, registryConfigPath string, useCachedImage bool, labels map[string]string) (containerd.Image, error) {
 	// Check the containerd image store to see if image exists
 	img, err := client.GetImage(ctx, source)
 	if err != nil {
@@ -986,11 +1002,11 @@ func fetchImage(ctx context.Context, source string, client *containerd.Client, r
 		log.G(ctx).WithField("ref", source).Info("Image exists, fetching cached image from image store")
 		return img, err
 	}
-	return pullImage(ctx, source, client, registryConfigPath)
+	return pullImage(ctx, source, client, registryConfigPath, labels)
 }
 
 // pullImage pulls an image from the specified source.
-func pullImage(ctx context.Context, source string, client *containerd.Client, registryConfigPath string) (containerd.Image, error) {
+func pullImage(ctx context.Context, source string, client *containerd.Client, registryConfigPath string, labels map[string]string) (containerd.Image, error) {
 	// Handle registry config
 	var registryConfig *RegistryConfig
 	if registryConfigPath != "" {
@@ -1017,9 +1033,18 @@ func pullImage(ctx context.Context, source string, client *containerd.Client, re
 	var img containerd.Image
 	for {
 		var err error
-		img, err = client.Pull(ctx, source,
+
+		pullOpts := []containerd.RemoteOpt{
 			withDynamicResolver(ctx, source, registryConfig),
-			containerd.WithSchema1Conversion)
+			containerd.WithSchema1Conversion,
+		}
+
+		if len(labels) != 0 {
+			pullOpts = append(pullOpts, containerd.WithPullLabels(labels))
+		}
+
+		img, err = client.Pull(ctx, source, pullOpts...)
+
 		if err == nil {
 			log.G(ctx).WithField("img", img.Name()).Info("pulled image successfully")
 			break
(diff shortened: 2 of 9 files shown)


packages/kubernetes-1.29/prestart-pull-pause-ctr.conf
```diff
@@ -6,4 +6,5 @@ ExecStartPre=/usr/bin/host-ctr \
     pull-image \
     --source=${POD_INFRA_CONTAINER_IMAGE} \
     --registry-config=/etc/host-containers/host-ctr.toml \
-    --skip-if-image-exists=true
+    --skip-if-image-exists=true \
+    --labels="io.cri-containerd.pinned=pinned"
```
yeazelm: Do we need a migration for this?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1184,3 +1201,15 @@ func withUnmaskedPaths(unmaskPaths []string) oci.SpecOpts {
 		return nil
 	}
 }
+
+// Convert labels to map[string]string for containerd.WithPullLabels.
+// Labels are in the format of "key=value,key2=value2".
+func convertLabels(labels string) map[string]string {
+	labelsMap := make(map[string]string)
```
yeazelm, line 1216: This could use some error handling, it is possible a user will put malformed input here and this function shouldn't provide undefined values if that is the case. I think it should validate its input by checking that `labels` is not empty (and if so just return the empty map). It should also ensure it gets "good" splits on the `,` and `=`. I'm sort of focused on someone forgetting a value and doing `--labels="foo=bar,baz="` or `--labels="foo=bar,baz,fuzz"` and such.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1184,3 +1201,15 @@ func withUnmaskedPaths(unmaskPaths []string) oci.SpecOpts {
 		return nil
 	}
 }
+
+// Convert labels to map[string]string for containerd.WithPullLabels.
+// Labels are in the format of "key=value,key2=value2".
+func convertLabels(labels string) map[string]string {
```
yeazelm: I'd like to see some unit tests for this since we are taking input from the cli which should have some reasonable validation that we are transforming this correctly into the right map.

yeazelm (commented):
You included a binary in the git commit that we should ensure we strip out. 

I'd also like to sort out the lint failures to see if its the code or if its something else? I ran the lint manually and it did find something worth fixing while we are updating the code:
```
cmd/host-ctr/main.go:43:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: As of Go 1.20 there is no reason to call Seed with a random value. Programs that call Seed with a known value to get a specific sequence of results should use New(NewSource(seed)) to obtain a local random generator. (staticcheck)
        rand.Seed(time.Now().UnixNano())
```

packages/kubernetes-1.29/prestart-pull-pause-ctr.conf
arnaldo2792: No, since the PR isn't changing the API shape nor the API store values.

yeazelm: I realized this after I hit the wrong button and submitted the comment out of band. But left it in case I was wrong on the second guess.

bcressey: Unless it's going to cause some other issue, we should set this for all k8s versions, not just 1.29.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -161,9 +162,15 @@ func App() *cli.App {
 					Destination: &useCachedImage,
 					Value:       false,
 				},
+				&cli.StringFlag{
+					Name:        "labels",
+					Usage:       "add labels to the pulled image. Each label should be in the format of `key=value` and separated by a comma",
```
bcressey: What characters are allowed in a label? If either comma or whitespace are allowed then this might not be the right interface.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -161,9 +162,15 @@ func App() *cli.App {
 					Destination: &useCachedImage,
 					Value:       false,
 				},
+				&cli.StringFlag{
```
bcressey: A [StringSliceFlag](https://cli.urfave.org/v2/examples/flags/#multiple-values-per-single-flag) that lets you pass one label per instance of `--label` might be better here.

sources/host-ctr/cmd/host-ctr/main.go
bcressey: In general, input conversion and validation should happen as close to where data enters the system as possible, so that you're not passing around potentially invalid or loosely typed data.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1184,3 +1201,15 @@ func withUnmaskedPaths(unmaskPaths []string) oci.SpecOpts {
...
 }
+
+// Convert labels to map[string]string for containerd.WithPullLabels.
+// Labels are in the format of "key=value,key2=value2".
+func convertLabels(labels string) map[string]string {
+	labelsMap := make(map[string]string)
+	labelsList := strings.Split(labels, ",")
+	for _, label := range labelsList {
+		labelKeyValue := strings.Split(label, "=")
+		labelsMap[labelKeyValue[0]] = labelKeyValue[1]
```
bcressey: Are labels required to have an `=` separate or can raw keys be used?

Linking to the label "spec" for containerd would help provide the right context for a thorough review.

dims: I believe it is here: https://github.com/opencontainers/image-spec/blob/main/config.md#:~:text=creating%20a%20container.-,Labels,-object%2C%20OPTIONAL

which links to https://github.com/opencontainers/image-spec/blob/365fa41472a19107b0b4cbc17e141a9b97f9b66e/annotations.md#rules

packages/kubernetes-1.29/prestart-pull-pause-ctr.conf
dims: Yes, this can be set for all versions

sources/host-ctr/cmd/host-ctr/main.go
dims: Added here - https://github.com/bottlerocket-os/bottlerocket/pull/3757/files#r1478741201

sources/host-ctr/cmd/host-ctr/main.go
gthao313: containerd expects a map[string]string as labels input. Otherwise, all it cares about is that the key and value are under 4096 bytes combined. https://github.com/containerd/containerd/blob/f5e7fe0cb6286333a597a407029fe7c760626669/pkg/labels/validate.go#L32-L41

I'll add this validation and meanwhile I'll validate `key=value` format.

gthao313:
^ addressed comments
1. use `StringSliceFlag` which can allow you to pass multiple values for a single flag, and joins all inputs with `,`. Therefore, we don't need validate if the input is seperated by `,`
Xxample:
```
... \
--label "io.cri-containerd.pinned=pinned \
--label "io.cri-containerd.example=example \
```
2. add validation and unit-test for `converLables` functions. [label rules](https://github.com/opencontainers/image-spec/blob/365fa41472a19107b0b4cbc17e141a9b97f9b66e/annotations.md#rules)
 - The single label must be None empty string
 - The single lable must be `key=value` format
 - The key and value are under 4096 bytes combined.

3. set `pinned` label for all k8s versions

ginglis13:
> I'd also like to sort out the lint failures to see if its the code or if its something else?

probably need a `go mod tidy` and/or bump the version in the workflow for golangci-lint to match host-ctr's (1.21):  
https://github.com/bottlerocket-os/bottlerocket/blob/2b1a787273cbbbf7a5f97ac063151c509c29d15b/.github/workflows/golangci-lint.yaml#L19

ginglis13 (approved):
LGTM on green CI

gthao313:
`golangci-lint` was broken on go 1.19 version. I'm going to bump bump the version in the workflow for golangci-lint to match 1.21.

gthao313:
rebase to use go 1.21 version on `golangci-lint`

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1184,3 +1210,38 @@ func withUnmaskedPaths(unmaskPaths []string) oci.SpecOpts {
...
+	}
+	return labelsMap, nil
+}
+
+// Validate a label's key and value are under 4096 bytes
+//
+// validation logic derived from:
+// https://github.com/containerd/containerd/blob/f5e7fe0cb6286333a597a407029fe7c760626669/pkg/labels/validate.go#L32-L41
+func Validate(k, v string) error {
```
arnaldo2792: nit: this is private so it should be all lowercase.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -971,7 +984,7 @@ func withProxyEnv() oci.SpecOpts {
 }
 
 // fetchImage returns a `containerd.Image` given an image source.
-func fetchImage(ctx context.Context, source string, client *containerd.Client, registryConfigPath string, useCachedImage bool) (containerd.Image, error) {
+func fetchImage(ctx context.Context, source string, client *containerd.Client, registryConfigPath string, useCachedImage bool, labels string) (containerd.Image, error) {
```
arnaldo2792: Currently with this function as it is, the image will be labeled only when it doesn't exist, but if it does, the image won't be labeled. I think it doesn't hurt to always apply the labels to the image, regardless if they are already present (provided that containerd doesn't fail when the labels already exist for the image).

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -39,6 +39,13 @@ import (
 // Example 2: 777777777777.dkr.ecr.cn-north-1.amazonaws.com.cn/my_image:latest
 var ecrRegex = regexp.MustCompile(`(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?.*`)
 
+// derived from: https://github.com/containerd/containerd/blob/f5e7fe0cb6286333a597a407029fe7c760626669/pkg/labels/validate.go#L25-L28
+const (
+	maxSize = 4096
+	// maximum length of key portion of error message if len of key + len of value > maxSize
+	keyMaxLen = 64
+)
+
```
bcressey: Not so much derived as copy/pasted, which is generally not what you want to do as it can trigger licensing changes.

Here it would be better to give the constant we need a more specific name:
```suggestion
const (
	imageLabelMaxSize = 4096
)
```

I don't think `keyMaxLen` is necessary; we aren't expecting `host-ctr` to be invoked with very long keys.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -161,9 +168,14 @@ func App() *cli.App {
...
 					Value:       false,
 				},
+				&cli.StringSliceFlag{
+					Name:  "label",
+					Usage: "add multiple labels to the pulled image. The label should be in the format of `key=value`",
+				},
 			},
 			Action: func(c *cli.Context) error {
-				return pullImageOnly(containerdSocket, namespace, source, registryConfig, useCachedImage)
+				labels := strings.Join(c.StringSlice("label"), `, `)
```
bcressey: Why take a slice of strings and turn it into a single string, only to split it back out later? Can you just pass around the slice and avoid the `strings.Join` and `strings.Split` calls?

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -1184,3 +1210,38 @@ func withUnmaskedPaths(unmaskPaths []string) oci.SpecOpts {
...
+// Label are in the format of "key=value".
+func convertLabels(labels string) (map[string]string, error) {
+	labelsMap := make(map[string]string)
+	labelsList := strings.Split(labels, ",")
+	for _, label := range labelsList {
+		labelKeyValue := strings.Split(label, "=")
+
+		if len(labelKeyValue) != 2 {
+			return labelsMap, fmt.Errorf("invalid label format: %s. It should be key=value", label)
+		}
```
bcressey: You could also tolerate input without an `=` and just assume it's a key where the value is the empty string.

sources/host-ctr/cmd/host-ctr/main.go
bcressey: The function should be named `validateLabel` or something to identify what it's validating.

I'd say just do the length check in `convertLabels` prior to the split on `=` and omit the `keyMaxLen` logic.

The main point from an interface standpoint is that the total length has to be under 4096 or it's not valid. We don't need to "derive" code from another project to check that.

gthao313: The function has been removed.

sources/host-ctr/cmd/host-ctr/main.go
gthao313: Yeah, good catch. I change to use the slice.

sources/host-ctr/cmd/host-ctr/main.go
gthao313: yeah, convertLabels function actually handle this situation and consider this is correct. I have unit test [here](https://github.com/bottlerocket-os/bottlerocket/pull/3757/files#diff-46e7ce9ff30b890e45ef5c58083d8fd181908851aa23c5c7db04d12bc5833ddcR283-R288)

sources/host-ctr/cmd/host-ctr/main.go
gthao313: We don't need "derive". I already added this to `convertLable` function.

sources/host-ctr/cmd/host-ctr/main.go
gthao313: I think we are fine. The current strategy can make sure we label all the images which needs to be labeled. I think it would be the same as label all images even they are already labeled.

sources/host-ctr/cmd/host-ctr/main.go
gthao313: I think if we support like that, we might allow the host-ctr to accept incorrect label. If users specify the label like `value-key`. The label doesn't have `=`, and if we assume it's a key where the value is the empty string, I think this would misinterpret the user's request. We should strictly ask users to input like "key=value". For empty value, it could be `key=` and this would be accepted by host-ctr. Unit-test [here](https://github.com/bottlerocket-os/bottlerocket/pull/3757/files#diff-46e7ce9ff30b890e45ef5c58083d8fd181908851aa23c5c7db04d12bc5833ddcR283-R289)

gthao313:
^ addressed comments.
1. remove Validate function and add label max size check to `convertLabel`.
2. move 'covertLabel' before `pullImageOnly`. convert and validate the input first.
3. add more unit tests.
4. Use slice

Test
```
[root@admin]# /crictl --runtime-endpoint unix:///.bottlerocket/rootfs/run/containerd/containerd.sock inspecti 106a8e54d5eb3 | grep pinned
    "pinned": true
[root@admin]# sudo sheltie
bash-5.1# ctr -a /run/containerd/containerd.sock -n k8s.io images list | grep pause
602401143452.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.1-eksbuild.1                                                                                application/vnd.docker.distribution.manifest.list.v2+json sha256:1cb4ab85a3480446f9243178395e6bee7350f0d71296daeb6a9fdd221e23aea6 292.4 KiB linux/amd64,linux/arm64 io.cri-containerd.image=managed,io.cri-containerd.pinned=pinned
ecr.aws/ar
```

sources/host-ctr/cmd/host-ctr/main.go
yeazelm: Is it valid to have the key be an empty string? If so, then I think that is probably a better approach. I wasn't sure if the containerd spec allows for that but we shouldn't enforce something more than what the upstream system can handle.

yeazelm: https://github.com/opencontainers/image-spec/blob/365fa41472a19107b0b4cbc17e141a9b97f9b66e/annotations.md#rules the second point:
`While the value MUST be present, it MAY be an empty string.` so I think we should accept empty strings, my earlier comment was overly strict on what is valid.

gthao313: >While the value MUST be present, it MAY be an empty string.

Yeah, they support the value can be empty string. The reason why I think we need `=` here because containerD directly receives `map` and we need convert the slice of strings  to `map`.

Ok, I think I can allow the labels to accept the label format without an `=`.

source from [rules](https://github.com/opencontainers/image-spec/blob/365fa41472a19107b0b4cbc17e141a9b97f9b66e/annotations.md#rules).

gthao313:
^addressed comments

tolerate input without an = and just assume it's a key where the value is the empty string.
If the provided label contains `=`, split them into key an value; otherwise, assume assume it's a key where the value is the empty string.

yeazelm (approved):
This looks correct to me 🚢

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -161,9 +166,18 @@ func App() *cli.App {
 					Destination: &useCachedImage,
 					Value:       false,
 				},
+				&cli.StringSliceFlag{
+					Name:  "label",
+					Usage: "add multiple labels to the pulled image. The label should be in the format of `key=value`",
```
bcressey: nit: could condense this a bit:
```suggestion
					Usage: "label to add to the pulled image in `key=value` format",
```

bcressey (approved):

PR #3761, opened 2024-02-05
actions-workflow: bump action go version to 1.21

gthao313:
**Description of changes:**
golangci-lint was broken on go 1.19 version. bump the version in the workflow for golangci-lint to match 1.21.
```
* can't run linter goanalysis_metalinter: buildir: failed to load package sysenc: could not load export data: no export data for "github.com/cilium/ebpf/internal/sysenc"
```

**Testing done:**
See Actions workflow

**Terms of contribution:**

### DIFF
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -40,7 +40,7 @@ import (
 var ecrRegex = regexp.MustCompile(`(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?.*`)
 
 func init() {
-	rand.Seed(time.Now().UnixNano())
+	rand.New(rand.NewSource(time.Now().UnixNano()))
 	// Dispatch logging output instead of writing all levels' messages to
 	// stderr.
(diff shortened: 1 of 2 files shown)


ginglis13 (approved):

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -40,7 +40,7 @@ import (
 var ecrRegex = regexp.MustCompile(`(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?.*`)
 
 func init() {
-	rand.Seed(time.Now().UnixNano())
+	rand.New(rand.NewSource(time.Now().UnixNano()))
```
yeazelm, line 43: I think we should just not seed correct? I don't think its needed anymore. Can you confirm either way with the docs for this lint?

gthao313, line 43: I have tested this change on host-ctr [PR](https://github.com/bottlerocket-os/bottlerocket/pull/3757)

yeazelm: I just re-read the docs, you have it correct now.

yeazelm (approved):

PR #3772, opened 2024-02-10
settings: add host-containers settings extension

### LINKED ISSUE #3650: OOTB: Port `host_containers` settings model to settings extensions
Create a settings extension for `host_containers`.

### Definition of done
A new `host-containers` crate is created under `sources/settings-extensions`, and all variants use the model defined in that crate in place of the existing settings. Additionally, a basic `settings-host-containers` RPM package is created that installs the extension binary.

sam-berning:
**Issue number:**

Closes #3650 

**Description of changes:**

Creates a host-containers settings extension.

Because host-containers still depends on the API to get settings, there are some changes in here to use the settings extension model within `host-containers`. This is definitely not best practice in the long run, but it should be stable for now and will be removed once we migrate `host-containers` to read settings from a config file instead. We can discuss whether we should wait to merge this PR until that change is in.

**Testing done:**

Built an `aws-dev` AMI and ran an instance with it. I tested both that `host-containers` worked as expected (considering I got into the instance via the admin container, it worked just fine) and that the settings extension gave responses as expected.

**Terms of contribution:**

### DIFF
--- a/sources/api/host-containers/src/main.rs
+++ b/sources/api/host-containers/src/main.rs
@@ -135,7 +135,9 @@ mod error {
 type Result<T> = std::result::Result<T, error::Error>;
 
 /// Query the API for the currently defined host containers
-async fn get_host_containers<P>(socket_path: P) -> Result<HashMap<Identifier, model::HostContainer>>
+async fn get_host_containers<P>(
+    socket_path: P,
+) -> Result<HashMap<Identifier, settings_extension_host_containers::HostContainer>>
 where
     P: AsRef<Path>,
 {
@@ -161,7 +163,7 @@ where
         serde_json::from_str(&response_body).context(error::ResponseJsonSnafu { method, uri })?;
 
     // If host containers aren't defined, return an empty map
-    Ok(settings.host_containers.unwrap_or_default())
+    Ok(settings.host_containers.unwrap_or_default().host_containers)
 }
 
 /// SystemdUnit stores the systemd unit being manipulated
@@ -347,7 +349,10 @@ fn parse_args(args: env::Args) -> Args {
     }
 }
 
-fn handle_host_container<S>(name: S, image_details: &model::HostContainer) -> Result<()>
+fn handle_host_container<S>(
+    name: S,
+    image_details: &settings_extension_host_containers::HostContainer,
+) -> Result<()>
 where
     S: AsRef<str>,
 {
--- a/sources/settings-extensions/host-containers/src/lib.rs
+++ b/sources/settings-extensions/host-containers/src/lib.rs
@@ -1,0 +1,119 @@
+/// host-containers settings allow users to configure multiple host containers
+use bottlerocket_settings_sdk::{GenerateResult, SettingsModel};
+use model_derive::model;
+use modeled_types::{Identifier, Url, ValidBase64};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::collections::HashMap;
+use std::convert::Infallible;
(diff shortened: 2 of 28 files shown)


sources/api/host-containers/src/main.rs
```diff
@@ -135,7 +135,9 @@ mod error {
 type Result<T> = std::result::Result<T, error::Error>;
 
 /// Query the API for the currently defined host containers
-async fn get_host_containers<P>(socket_path: P) -> Result<HashMap<Identifier, model::HostContainer>>
+async fn get_host_containers<P>(
+    socket_path: P,
+) -> Result<HashMap<Identifier, settings_extension_host_containers::HostContainer>>
 where
     P: AsRef<Path>,
 {
```
webern, line 143: Below this we are calling API Server. In other other PRs, e.g. corndog, prariedog, et. al., we are eliminating reliance on the API Server by changing to a config. We will be doing that here as well right? Are we deciding to make such changes in two steps instead of one PR?

If we do, it raises the question of whether the config should be modeled as a map of `settings_extension_host_containers::HostContainer` structures. In other cases I don't think we have re-used the settings extension model directly in the config file.

sam-berning: I was imagining that the setting extension and the config file change would be two PRs, I think they should be relatively separate changes. I can  take the config file work and get a PR out for that first.

> If we do, it raises the question of whether the config should be modeled as a map of `settings_extension_host_containers::HostContainer` structures. In other cases I don't think we have re-used the settings extension model directly in the config file.

The dogs should define their own structs to model the config files. We shouldn't depend on the settings model structs because those are no longer the source of truth for the setting shape, and we shouldn't depend on settings extensions because we can't guarantee that they'll be there

sam-berning:
^ fixed fmt errors

sources/api/host-containers/src/main.rs
webern: 2 PRs seems fine. If I understand correctly (and this is how I would want it): `host-containers` will no longer depend on the settings-extension once it has a config file, right?

sam-berning: yep, that's what I was thinking

sources/settings-extensions/host-containers/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+
+impl<'de> Deserialize<'de> for HostContainersSettingsV1 {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let host_containers = HashMap::deserialize(deserializer)?;
+        Ok(Self { host_containers })
+    }
+}
```
webern, line 33: Are these necessary? Can you derive `Serialize` and `Deserialize`?

sources/settings-extensions/host-containers/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+
+#[model(impl_default = true)]
```
webern, line 35: Same question I had before. What are we using the `model` macro for? Can we do whatever it's doing manually instead. I think all it is doing is making these `Option`s and implementing `Serialize` and `Deserialize`. This would be better for readability if it were done manually.

webern (approved):

sam-berning:
Looks like the `vmw_backdoor` crate only builds for x86, so I'll need to split the vmware user data providers out into a separate rpm to avoid conditional compilation

PR #3775, opened 2024-02-14
update grub and shim

bcressey:
**Issue number:**
N/A

**Description of changes:**
Update `shim` to 15.8 which includes recent CVE fixes.

Update `grub` to the latest version from AL23, and revert two patches that aren't required for Bottlerocket's Secure Boot implementation. This includes fixes for CVE-2023-4692 and CVE-2023-4693, which don't apply to Bottlerocket since the NTFS module isn't built. Overall, this update is a bit of a no-op but I wanted to get it in to record the decision to revert the patches, rather than leaving that for a future update.

I also added a patch from Red Hat that fixes CVE-2023-4001. That doesn't apply to Bottlerocket because the `search` directive isn't used to locate `grub.cfg`, so there is no ability to bypass a password that might be set. On variants that support Secure Boot there's also no way to modify `grub.cfg` to set a password. However, the functionality is useful to ensure that we read the expected boot config file for extending the kernel command line.

**Testing done:**
Confirmed that Secure Boot works for the following variants:
* `aws-k8s-1.28` - x86_64, aarch64
* `vmware-dev` - x86_64
* `metal-dev` - x86_64 (under QEMU)

I also tested legacy BIOS boot on c3.large to confirm the "uefi-preferred" functionality works as expected.

Verified that in-place upgrades and downgrades worked for `aws-k8s-1.28` (x86_64) and `aws-k8s-1.28-nvidia` (aarch64).

**Terms of contribution:**

### DIFF
--- a/packages/grub/0047-Revert-UBUNTU-Move-verifiers-after-decompressors.patch
+++ b/packages/grub/0047-Revert-UBUNTU-Move-verifiers-after-decompressors.patch
@@ -1,0 +1,48 @@
+From 95cafe6cf7dd2a02bd33ddee624bb9b7c3a931ae Mon Sep 17 00:00:00 2001
+From: Ben Cressey <bcressey@amazon.com>
+Date: Tue, 13 Feb 2024 22:21:41 +0000
+Subject: [PATCH] Revert "UBUNTU: Move verifiers after decompressors"
+
+We use the PGP verifier to validate the signature of grub.cfg, and do
+not want to expose the decompressors to untrusted input.
+
+Signed-off-by: Ben Cressey <bcressey@amazon.com>
+---
+ include/grub/file.h        | 4 ++--
+ tests/file_filter/test.cfg | 2 +-
+ 2 files changed, 3 insertions(+), 3 deletions(-)
+
+diff --git a/include/grub/file.h b/include/grub/file.h
+index fa23688..96827a4 100644
+--- a/include/grub/file.h
++++ b/include/grub/file.h
+@@ -180,13 +180,13 @@ extern grub_disk_read_hook_t EXPORT_VAR(grub_file_progress_hook);
+ /* Filters with lower ID are executed first.  */
+ typedef enum grub_file_filter_id
+   {
++    GRUB_FILE_FILTER_VERIFY,
+     GRUB_FILE_FILTER_GZIO,
+     GRUB_FILE_FILTER_XZIO,
+     GRUB_FILE_FILTER_LZOPIO,
++    GRUB_FILE_FILTER_MAX,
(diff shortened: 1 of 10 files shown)


packages/grub/0047-Revert-UBUNTU-Move-verifiers-after-decompressors.patch
```diff
@@ -0,0 +1,48 @@
...
+   {
++    GRUB_FILE_FILTER_VERIFY,
+     GRUB_FILE_FILTER_GZIO,
+     GRUB_FILE_FILTER_XZIO,
+     GRUB_FILE_FILTER_LZOPIO,
++    GRUB_FILE_FILTER_MAX,
+     GRUB_FILE_FILTER_COMPRESSION_FIRST = GRUB_FILE_FILTER_GZIO,
+     GRUB_FILE_FILTER_COMPRESSION_LAST = GRUB_FILE_FILTER_LZOPIO,
+-    GRUB_FILE_FILTER_VERIFY,
+-    GRUB_FILE_FILTER_MAX,
```
webern, line 31: unnecessary sort?

webern (approved):

packages/grub/0047-Revert-UBUNTU-Move-verifiers-after-decompressors.patch
markusboehme: No, the order does matter here. Enumerations in C are numeric, not symbolic, i.e. the order changes their assigned values. It is idiomatic to loop e.g. from 0 to some sentinel value (like `GRUB_FILE_FILTER_MAX` here). For this particular case, the ordering of filters in the code determines their numeric value and therefore in which order they're applied. In any case, those two patches are reverts of downstream patches packaged in the SRPM and hence should match what's in there.

packages/grub/0048-add-flag-to-only-search-root-dev.patch
```diff
@@ -0,0 +1,163 @@
...
++    }
++    *p = '\0';
++  
++    if (*t != '\0')
++    {
++      t++;
++      while ((*r++ = *t++) != '\0')
++        ;
++      *r = '\0';
++    }
```
markusboehme, line 125: The termination of `r` should probably be outside the conditional to guarantee sane output in case there's nothing following the delimiter. That's not a problem for the function's use case in this patch for now, since the remainder is ignored anyway, but curious whether you know the proper process to get this fixed in Fedora?

markusboehme:
Confirmed patches match their upstream counterparts and the shim update included revokes the previous `grub` generation number. Ready to approve once test results are in. :racehorse:

yeazelm (approved):
The changes look good to me. 🚢

markusboehme (approved):
:lock: :heavy_check_mark:

PR #3777, opened 2024-02-14
host-containers: migrate to use config file

### LINKED ISSUE #3627: OOTB: Remove model dependency from `host-containers`
`host-containers` queries the API for all settings, deserializing into the model. It should probably read from a template rendered by a settings extension.

sam-berning:
**Issue number:**

Closes #3627 

**Description of changes:**

Removes dependency on the API model from host-containers by changing it to use a config file rendered from the settings.

**Testing done:**

- [x] Use apiclient to update relevant settings at runtime. Verify config was updated correctly.
  - [x] verify connectivity
  - [x] verify connectivity
  - [x] verify config file

**Terms of contribution:**

### DIFF
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -34,6 +34,7 @@ Source13: cis-checks-k8s-metadata-json
 Source14: certdog-toml
 Source15: prairiedog-toml
 Source16: thar-be-updates-toml
+Source19: host-containers-toml
 
 # 1xx sources: systemd units
 Source100: apiserver.service
@@ -475,7 +476,8 @@ install -d %{buildroot}%{_cross_datadir}/updog
 install -p -m 0644 %{_cross_repo_root_json} %{buildroot}%{_cross_datadir}/updog
 
 install -d %{buildroot}%{_cross_templatedir}
-install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{S:14} %{S:15} %{S:16} %{buildroot}%{_cross_templatedir}
+install -p -m 0644 %{S:5} %{S:6} %{S:7} %{S:8} %{S:14} %{S:15} %{S:16} %{S:19} \
+  %{buildroot}%{_cross_templatedir}
 
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 \
@@ -594,6 +596,7 @@ install -p -m 0644 %{S:400} %{S:401} %{S:402} %{buildroot}%{_cross_licensedir}
 %{_cross_tmpfilesdir}/host-containers.conf
 %dir %{_cross_templatedir}
 %{_cross_templatedir}/host-ctr-toml
+%{_cross_templatedir}/host-containers-toml
 
 %files -n %{_cross_os}storewolf
 %{_cross_bindir}/storewolf
--- a/sources/api/host-containers/Cargo.toml
+++ b/sources/api/host-containers/Cargo.toml
@@ -10,16 +10,17 @@ build = "build.rs"
 exclude = ["README.md"]
 
 [dependencies]
-apiclient = { path = "../apiclient", version = "0.1" }
 base64 = "0.13"
 constants = { path = "../../constants", version = "0.1" }
-http = "0.2"
 log = "0.4"
-models = { path = "../../models", version = "0.1" }
-serde_json = "1"
+modeled-types = { path = "../../models/modeled-types", version = "0.1" }
+serde = { version = "1", features = ["derive"] }
 simplelog = "0.12"
 snafu = "0.7"
-tokio = { version = "~1.32", default-features = false, features = ["macros", "rt-multi-thread"] }  # LTS
+toml = "0.8"
+
+[dev-dependencies]
+tempfile = "3"
 
 [build-dependencies]
 generate-readme = { version = "0.1", path = "../../generate-readme" }
--- a/sources/api/host-containers/src/config.rs
+++ b/sources/api/host-containers/src/config.rs
@@ -1,0 +1,18 @@
+use modeled_types::{Identifier, Url, ValidBase64};
+use serde::Deserialize;
+use std::collections::HashMap;
+
+#[derive(Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) struct HostContainersConfig {
+    pub(crate) host_containers: Option<HashMap<Identifier, HostContainer>>,
+}
+
+#[derive(Deserialize, Debug, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+pub(crate) struct HostContainer {
+    pub(crate) source: Option<Url>,
+    pub(crate) enabled: Option<bool>,
+    pub(crate) superpowered: Option<bool>,
+    pub(crate) user_data: Option<ValidBase64>,
+}
--- a/sources/api/host-containers/src/main.rs
+++ b/sources/api/host-containers/src/main.rs
@@ -3,7 +3,7 @@
 
 host-containers ensures that host containers are running as defined in system settings.
 
-It queries the API for their settings, then configures the system by:
+It reads the currently configured containers from its config file, then configures the system by:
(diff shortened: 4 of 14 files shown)


sam-berning:
^ update template to `{{{` everything to keep it consistent with what we expected when reading directly from the API

sources/api/host-containers/src/config.rs
```diff
@@ -0,0 +1,16 @@
+use model_derive::model;
+use modeled_types::{Identifier, Url, ValidBase64};
+use std::collections::HashMap;
+
+#[model]
```
webern: Using  `model` makes me a little squeamish, perhaps irrationally. What is it that this macro is doing for you? Is it just turning the fields into `Options` and implementing `Deserialize`? If so, I would prefer that be explicitly done rather than hiding information from the programmer in a fairly anemic procedural macro.

I feel the same way about `modeled-types` just due to its associate the "The Model", however I see that you are using `ValidBase64` which is real/desirable code re-use.

packages/os/os.spec
```diff
@@ -591,6 +592,7 @@ install -p -m 0644 %{S:400} %{S:401} %{S:402} %{buildroot}%{_cross_licensedir}
 %{_cross_tmpfilesdir}/host-containers.conf
 %dir %{_cross_templatedir}
 %{_cross_templatedir}/host-ctr-toml
+%{_cross_templatedir}/host-containers-toml
```
webern, line 599: These config file names are a little bit confusing next to each other, but I think you've named the new one correctly.

sources/api/host-containers/Cargo.toml
```diff
@@ -10,16 +10,15 @@ build = "build.rs"
 exclude = ["README.md"]
 
 [dependencies]
-apiclient = { path = "../apiclient", version = "0.1" }
 base64 = "0.13"
 constants = { path = "../../constants", version = "0.1" }
-http = "0.2"
 log = "0.4"
-models = { path = "../../models", version = "0.1" }
```
webern, line 18: :tada:

sources/api/host-containers/src/main.rs
```diff
@@ -134,34 +120,23 @@ mod error {
 
 type Result<T> = std::result::Result<T, error::Error>;
 
-/// Query the API for the currently defined host containers
-async fn get_host_containers<P>(socket_path: P) -> Result<HashMap<Identifier, model::HostContainer>>
+/// Read the currently defined host containers from the config file
+fn get_host_containers<P>(config_path: P) -> Result<HashMap<Identifier, config::HostContainer>>
```
webern, line 124: Can you add a test for this function by creating a config file, perhaps inline? One benefit of the new paradigm is that we can test since we don't need the API server.

sources/api/host-containers/src/main.rs
```diff
@@ -134,34 +120,23 @@ mod error {
...
-            code,
-            response_body,
-        }
-    );
-
-    // Build a Settings struct from the response string
-    let settings: model::Settings =
-        serde_json::from_str(&response_body).context(error::ResponseJsonSnafu { method, uri })?;
+    debug!("Reading containers from the config file!");
+    let config_path = config_path.as_ref();
```
webern: More information is better when debugging.

```suggestion
    let config_path = config_path.as_ref();
    debug!("Reading containers from the config file: {}", config_path.display());
```

sources/api/host-containers/src/main.rs
```diff
@@ -469,7 +443,7 @@ fn is_container_affected(settings: &[&str], container_name: &str) -> bool {
     false
 }
 
-async fn run() -> Result<()> {
```
webern, line 472: :tada:

webern (approved):

sources/api/host-containers/src/config.rs
sam-berning: > Is it just turning the fields into Options and implementing Deserialize?

Yeah, this and making the struct and all fields `pub`. I was also implementing `Debug` during development but it's not necessary. I think you're right that we shouldn't use `model-derive` here, I just noticed that it did all of what I wanted and it looked prettier 😆

packages/os/os.spec
sam-berning: yeah agreed. I also considered combining the two config files but I think that would be even more confusing

sources/api/host-containers/src/config.rs
webern: Prettier is all relative. To me, knowing that something is `pub` without looking at a derive macro is... prettier.

sam-berning:
^ address @webern's comments

jmt-lab (approved):

webern (approved):

webern (changes requested):
Switching my review status as it needs a rebase and then needs testing. Here is the test plan that we have suggested, this should be pasted into the testing section of the PR description

```
- [ ] Ran `cargo +nightly udeps` to ensure all unused dependencies have been removed
- [ ] Pass the settings on first boot via userdata. Verify config.
- [ ] cat /etc/os-release record your build ID include commit sha, record this
- [ ] Use apiclient to update relevant settings at runtime. Verify config was updated correctly.
- [ ] Downgrade to older version. 
  - [ ] verify connectivity
  - [ ] cat /etc/os-release to verify lower version
  - [ ] verify that migrations have changed the metadata/settings as expected (i.e. they are gone from the datastore, or lists have been changed back)
- [ ] Upgrade to newer version.
  - [ ] verify connectivity
  - [ ] cat /etc/os-release to verify version
  - [ ] verify config file
 ```

**Comment by sam-berning** [comment]
^ testing summary:

    Enabled and configured the admin container in user data. `cat`'d the config file at `/etc/host-containers/host-containers.toml` and it contained the user data.

```
bash-5.1# cat /etc/os-release
...
VERSION="1.19.3 (aws-dev)"
...
BUILD_ID=1e9e8f3a
...
```

- [x] Use apiclient to update relevant settings at runtime. Verify config was updated correctly.

```
bash-5.1# apiclient set --json '{"settings": {"host-containers": {"foo": {"source": "public.ecr.aws/bottlerocket/bottlerocket-admin:v0.11.5", "enabled": true, "superpowered": true}}}}'
bash-5.1# cat /etc/host-containers/host-containers.toml
...
[host-containers."foo"]
source = "public.ecr.aws/bottlerocket/bottlerocket-admin:v0.11.5"
enabled = true
superpowered = true
```

  - [x] verify connectivity

```
bash-5.1# updog update -i 1.19.2 -r -n
Starting update to 1.19.2
Reboot scheduled for Wed 2024-03-06 01:53:29 UTC, use 'shutdown -c' to cancel.
Update applied: aws-dev 1.19.2
# after reboot, was able to ssh successfully
bash-5.1# cat /etc/os-release
...
VERSION="1.19.2 (aws-dev)"
...
# no host-containers-toml config file, and it is removed from host-containers config files
bash-5.1# ls /var/lib/bottlerocket/datastore/current/live/configuration-files/ 
certdog-toml   containerd-config-toml  hostname				   metricdog-toml  motd		proxy-env
cfsignal-toml  docker-daemon-config    hosts				   modprobe-conf   netdog-toml	updog-toml
chrony-conf    host-ctr-toml	       log4j-hotpatch-enabled-log-message  modules-load    oci-hooks
bash-5.1# cat /var/lib/bottlerocket/datastore/current/live/services/host-containers/configuration-files 
["host-ctr-toml"]
```

  - [x] verify connectivity
  - [x] verify config file

```
bash-5.1# updog update -i 1.19.3 -r -n
Starting update to 1.19.3
Reboot scheduled for Wed 2024-03-06 01:58:04 UTC, use 'shutdown -c' to cancel.
Update applied: aws-dev 1.19.3
# after reboot, was able to ssh successfully
bash-5.1# cat /etc/os-release
...
VERSION="1.19.3 (aws-dev)"
...
bash-5.1# cat /etc/host-containers/host-containers.toml
# looked as expected!
```

webern (approved):
Looks good.

jmt-lab:
Looks good to me as well

PR #3782, opened 2024-02-16
Support for ECS_ENABLE_CONTAINER_METADATA

### LINKED ISSUE #3717: Add additional ECS configuration values to `settings.ecs`
**What I'd like:**
Add support for configuring the following values for ecs-agent:
* [ECS_BACKEND_HOST](https://github.com/aws/amazon-ecs-agent/blob/61c8a8c5b48e2efb7193fe61c4bb3cae816f5821/agent/config/config.go#L528)
* [ECS_AWSVPC_BLOCK_IMDS](https://github.com/aws/amazon-ecs-agent/blob/61c8a8c5b48e2efb7193fe61c4bb3cae816f5821/agent/config/config.go#L570)

n/a

arnaldo2792:
**Issue number:**

Closes #3717

**Description of changes:**

As described in the #3717, the ECS agent will try to mount a file into the task using both `ECS_HOST_DATA_DIR` and `ECS_DATADIR` to create the path of the source file for the mount. This is a problem in Bottlerocket because both paths include `/var/lib/ecs` so the final source path ends up including that path twice, e.g.: `/var/lib/ecs/var/lib/ecs/data`. 

The source file is created in the correct location (that is `/var/lib/ecs/data`), because the ECS agent only uses `ECS_DATADIR` to write to the metadata file for the task. The problem is on the source file used for the mount, as described above.

With this change, the ECS agent is patched to ignore `ECS_HOST_DATA_DIR` while creating the source path for the mount, and only `ECS_DATADIR` is used instead. 

**Testing done:**
- [X] In `aws-ecs-2`, I tested that an ECS task can read from the file at `${ECS_CONTAINER_METADATA_FILE}`
- [X] Ran migration testing
- [x] There isn't a test that covers this case in the internal testing suite, but still I'll run it to confirm nothing breaks due to the patch

**Terms of contribution:**

### DIFF
--- a/packages/ecs-agent/ecs.config
+++ b/packages/ecs-agent/ecs.config
@@ -39,3 +39,6 @@ ECS_BACKEND_HOST="{{settings.ecs.backend-host}}"
 {{#if settings.ecs.awsvpc-block-imds}}
 ECS_AWSVPC_BLOCK_IMDS="{{settings.ecs.awsvpc-block-imds}}"
 {{/if}}
+{{#if settings.ecs.enable-container-metadata}}
+ECS_ENABLE_CONTAINER_METADATA="{{settings.ecs.enable-container-metadata}}"
+{{/if}}
--- a/Release.toml
+++ b/Release.toml
@@ -273,4 +273,5 @@ version = "1.19.2"
 "(1.19.1, 1.19.2)" = [
     "migrate_v1.19.2_certdog-config-file-v0-1-0.lz4",
     "migrate_v1.19.2_certdog-service-cfg-v0-1-0.lz4",
+    "migrate_v1.19.2_add-ecs-enable-container-metadata.lz4",
 ]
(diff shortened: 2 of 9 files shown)


yeazelm (approved):
LGTM!

ginglis13 (approved):

packages/ecs-agent/ecs.config
```diff
@@ -39,3 +39,6 @@ ECS_BACKEND_HOST="{{settings.ecs.backend-host}}"
 {{#if settings.ecs.awsvpc-block-imds}}
 ECS_AWSVPC_BLOCK_IMDS="{{settings.ecs.awsvpc-block-imds}}"
 {{/if}}
+{{#if settings.ecs.enable-container-metadata}}
+ECS_ENABLE_CONTAINER_METADATA="{{settings.ecs.enable-container-metadata}}"
```
stockholmux, line 43: @arnaldo2792 Since this adds a new setting, don't forget to make an issue on bottlerocket/bottlerocket-project-website once meged.

webern (approved):

PR #3801, opened 2024-02-26
Update github checkout and cache actions to v4

### LINKED ISSUE #3800: Github: update workflows to use node.js 20
**Platform I'm building on:**

Github (via Actions). I will have a PR in a few seconds.

Actions run without deprecation warnings.

**What actually happened:**

50 or so annotations. See https://github.com/bottlerocket-os/bottlerocket/actions/runs/8042795626

Run a workflow on Github.

larvacea:
Github has deprecated node 16, so we should be using the @v4 versions of the checkout and cache actions.

**Issue number:**

Closes #3800

**Description of changes:**

We had deprecated versions of cache and checkout in two files.

**Testing done:**

With any luck, this pull request will launch a workflow.

**Terms of contribution:**

### DIFF
--- a/.github/actions/list-variants/action.yml
+++ b/.github/actions/list-variants/action.yml
@@ -10,7 +10,7 @@ outputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - id: get-variants
       name: Determine variants
       shell: bash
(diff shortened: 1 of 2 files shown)


larvacea:
I verified that the workflow run for this pull request did not generate 50 annotations. The Determine variants job was calling checkout@v3 in the list-variants action, and the setup-node action in the other 49 jobs was calling cache@v3. Now, both are @v4 and the run is deprecation-free.

ginglis13 (approved):

markusboehme (approved):

PR #3804, opened 2024-03-01
fix: provide runtime cgroups to kubelet

### LINKED ISSUE #3776: Missing runtime metrics from cAdvisor
**Platform I'm building on:**

EKS

I'd expect to see the cAdvisor runtime values as part of the kubelet metrics.

**What actually happened:**

The kubelet cAdvisor metrics are missing the runtime values meaning we can't see the node usage.

Run `kubectl get --raw "/api/v1/nodes/${NODE_NAME}/proxy/stats/summary" | jq -r '.node.systemContainers[] | .name'` and you'll see the following response.

```shell
kubelet
pods
```

This the same as https://github.com/awslabs/amazon-eks-ami/issues/1667; the solution I added there would also seem to be relevant in the context of Bottlerocket.

ginglis13:
**Issue number:**

Closes #3776 

**Description of changes:**

Pass the cgroup where containerd is running to kubelet to enable container resource and performance monitoring via tools like cadvisor.

**Testing done:**

Tested on `aws-k8s-1.23-aarch64` and `aws-k8s-1.29-aarch64`:

```
$ kubectl get --raw "/api/v1/nodes/${NODE_NAME}/proxy/stats/summary" | jq -r '.node.systemContainers[] | .name'
pods
kubelet
runtime
```

**Terms of contribution:**

### DIFF
--- a/packages/kubernetes-1.23/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.23/kubelet-exec-start-conf
@@ -38,4 +38,5 @@ ExecStart=/usr/bin/kubelet \
 {{#if settings.kubernetes.log-level includeZero=true}}
     -v {{settings.kubernetes.log-level}} \
 {{/if}}
-    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE}
+    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE} \
+    --runtime-cgroups=/runtime.slice/containerd.service
--- a/packages/kubernetes-1.24/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.24/kubelet-exec-start-conf
@@ -37,4 +37,5 @@ ExecStart=/usr/bin/kubelet \
 {{#if settings.kubernetes.log-level includeZero=true}}
     -v {{settings.kubernetes.log-level}} \
 {{/if}}
-    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE}
+    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE} \
+    --runtime-cgroups=/runtime.slice/containerd.service
--- a/packages/kubernetes-1.25/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.25/kubelet-exec-start-conf
@@ -37,4 +37,5 @@ ExecStart=/usr/bin/kubelet \
 {{#if settings.kubernetes.log-level includeZero=true}}
     -v {{settings.kubernetes.log-level}} \
 {{/if}}
-    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE}
+    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE} \
+    --runtime-cgroups=/runtime.slice/containerd.service
--- a/packages/kubernetes-1.26/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.26/kubelet-exec-start-conf
@@ -37,4 +37,5 @@ ExecStart=/usr/bin/kubelet \
 {{#if settings.kubernetes.log-level includeZero=true}}
     -v {{settings.kubernetes.log-level}} \
 {{/if}}
-    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE}
+    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE} \
+    --runtime-cgroups=/runtime.slice/containerd.service
--- a/packages/kubernetes-1.27/kubelet-exec-start-conf
+++ b/packages/kubernetes-1.27/kubelet-exec-start-conf
@@ -34,4 +34,5 @@ ExecStart=/usr/bin/kubelet \
 {{#if settings.kubernetes.log-level includeZero=true}}
     -v {{settings.kubernetes.log-level}} \
 {{/if}}
-    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE}
+    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE} \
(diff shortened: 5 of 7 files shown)


packages/kubernetes-1.23/kubelet-exec-start-conf
```diff
@@ -38,4 +38,5 @@ ExecStart=/usr/bin/kubelet \
 {{#if settings.kubernetes.log-level includeZero=true}}
     -v {{settings.kubernetes.log-level}} \
 {{/if}}
-    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE}
+    --pod-infra-container-image ${POD_INFRA_CONTAINER_IMAGE} \
+    --runtime-cgroups=/runtime.slice/containerd.service
```
bcressey, line 42: This doesn't feel quite right to me. `runtime.slice` is a systemd-managed cgroup and should not be touched by any process without setting up a delegation.

The kubelet args reference says this is for:
> Optional absolute name of cgroups to create and run the runtime in.

But kubelet should be using the systemd cgroup driver and shouldn't be creating any cgroups directly.

Possibly the option has evolved and does more than what's described?

Donnerdiebel:
Hi all!

Any chances that this will get merged in the near future?

We just ran into this problem and deploying additional cAdvisors to be able to collect runtime metrics when using bottlerocket amis on EKS is not exactly outr preferred solution.
Anyways, thanks for taking a look into this and creating the fix.

All the best!

ginglis13:
> Hi all!
> 
> Any chances that this will get merged in the near future?
> 
> We just ran into this problem and deploying additional cAdvisors to be able to collect runtime metrics when using bottlerocket amis on EKS is not exactly outr preferred solution. Anyways, thanks for taking a look into this and creating the fix.
> 
> All the best!

Hey @Donnerdiebel , thanks for the bump. I've got some additional testing to do for this change and haven't gotten around to it... I'll try to have an update by the end of the week 😄

packages/kubernetes-1.23/kubelet-exec-start-conf
ginglis13: The wording of the kubelet args ref does sound  like there’s changes being made at a system level here:

> Optional absolute name of cgroups to create and run the runtime in.

but after reading more and more issues/links related to this, it seems to me that this is meant to convey the name of the cgroup that the runtime is running in and to collect stats from. I traced this flag bag to its inception, across its old name `runtime-container`, to the commit where it all began: https://sourcegraph.com/github.com/kubernetes/kubernetes/-/commit/38efc837b906d475daf683beb559b90b9a03f5af. Here, the flag used to have the description 

> "Absolute name of the cgroups to create (if required) and run the runtime in (Default: /docker-daemon)."

Maybe that “if required” bit would make the helptext on the flag today sound less intrusive than it really is because it really will not create the cgroup if it already exists. It looks to me like a vestige of the past, when this option may actually have been creating cgroups. Even as far back as k8s 1.23, the only context in which the value(s) set by this flag are used by cadvisor: https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/kubernetes/kubernetes%24%40v1.23.0+cgrouproots&patternType=keyword&sm=0 (`cgrouproots` is populated with, among other things, the value of `--runtime-cgroups`: [here](https://sourcegraph.com/github.com/kubernetes/kubernetes@ab69524f795c42094a6630298ff53f3c3ebab7f4/-/blob/pkg/kubelet/config/flags.go?L90:18-90:32#tab=references) it is set and [here](https://sourcegraph.com/github.com/kubernetes/kubernetes@ab69524f795c42094a6630298ff53f3c3ebab7f4/-/blob/cmd/kubelet/app/server.go?L667-672) it is appended to that list)

This will be the case given the default/recommended systemd cgroup driver is used. I suppose someone out there could be using [cgroupfs](https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/#using-the-cgroupfs-driver) but I’m not sure why they would since kubernetes does not recommend it.

I did some additional testing just to be more confident here. I created two clusters, one for k8s 1.23-1.25 nodes and one for k8s 1.26-1.29 nodes. For each node, I 1/ ensured that the change worked in viewing the runtime stats group via `kubectl get —raw "/api/v1/nodes/${NODE_NAME}/proxy/stats/summary" | jq -r '.node.systemContainers[] | .name'` and 2/ that the cgroup hierarchy via `systemd-cgls -u runtime.slice` and `systemd-cgls -u system.slice` between nodes with the change and without the change passed the “eyeball test”, aka no abnormalities.

bcressey: Great! Thank you for the extensive research and testing here. That puts my mind at ease.

bcressey (approved):

gthao313:
The change and test looks good to me.
Just have a nit for future commit message. It would be recommended to give info like what package you have changes on. 
For example, `fix: provide runtime cgroups to kubelet` can be `kubernetes:  provide runtime cgroups to kubelet`. : )

gthao313 (approved):

ginglis13:
^ reword commit message

dhumphries-sainsburys:
Do we know when this is going to be released to a tag as doesn't seem to be there yet?

ginglis13:
Hi @dhumphries-sainsburys, you are correct that this fix has not yet made it into an official release of Bottlerocket. It will be included in the next release, which the team is still working to plan and prioritize.

saikumarch7548:
when is the next release planned @ginglis13

ginglis13:
Hi @saikumarch7548, this change has been included in the latest release of Bottlerocket, v1.19.5: https://github.com/bottlerocket-os/bottlerocket/releases/tag/v1.19.5

saikumarch7548:
@ginglis13 , Thank you , we have tested it in our product and it works

PR #3033, opened 2023-04-19
aws-k8s: pass `--hostname-override` to kubelet, set it to inst. `PrivateDnsName` for k8s-1.26

etungsten:
**Issue number:**

Resolves https://github.com/bottlerocket-os/bottlerocket/issues/3028

**Description of changes:**

```
    models,packages: add 'hostname-override' for kubelet, add kubelet option
    
    This adds the 'kubernetes.hostname-override' setting for kubelet's
    '--hostname-override' option.
```
```
    pluto: add 'private-dns-name' subcommand
    
    Adds a new command for retrieving the instance's PrivateDnsName.
    
    Refactors out code for setting up proxy for the AWS API clients so it
    can be shared between the EKS module and EC2 module.
```
```
    models: generate 'hostname-override' and set cp to external in aws-k8s-1.26
    
    This separates the model for aws-k8s-1.26 so that we generate the
    'hostname-override' setting and set the cloud-provider to 'external' for
    aws-k8s-1.26 variants only.
```
```
    packages: set '--cloud-provider' in 1.26 kubelet according to setting
    
    This lets the 'settings.kubernetes.cloud-provider' setting control what
    gets passed to '--cloud-provider' in kubelet options
```
```
    migrations: add migrations for new 'hostname-override' setting
```

**Testing done:**
#### On `aws-k8s-1.26`:
Launched `aws-k8s-1.26` node into 1.26 cluster in us-west-2 and the node registers itself fine:
```
$ kubectl get nodes -o wide
NAME                                             STATUS   ROLES    AGE     VERSION               INTERNAL-IP      EXTERNAL-IP     OS-IMAGE                                KERNEL-VERSION   CONTAINER-RUNTIME
i-044decb26a95a5253.us-west-2.compute.internal   Ready    <none>   2m29s   v1.26.2-eks-b106822   192.168.21.148   52.38.212.231   Bottlerocket OS 1.14.0 (aws-k8s-1.26)   5.15.102         containerd://1.6.19+bottlerocket
i-0beac6e5f9da1f6e2.us-west-2.compute.internal   Ready    <none>   2m27s   v1.26.2-eks-b106822   192.168.10.93    52.35.104.85    Bottlerocket OS 1.14.0 (aws-k8s-1.26)   5.15.102         containerd://1.6.19+bottlerocket
```

Checking on the host, the kubelet command line arguments has `--hostname-override` as expected:
```
# /etc/systemd/system/kubelet.service.d/exec-start.conf
[Service]
ExecStart=
[...]

PR #3828, opened 2024-03-18
Jmt/ootb/pluto

### LINKED ISSUE #3620: OOTB: Remove conditional compilation and model dependency from `pluto`
Currently the program is conditionally compiled based on the platform (AWS).  It also makes calls to the API and pulls the entire set of settings from the model - which will need to change for OOTB.

- [ ] Investigate if some of `pluto`'s behavior could be rolled into the Kubernetes settings extension  (perhaps kubernetes setting generator could call out to other binaries)
- [ ] Remove the conditional compilation directives and exclude the package from non-AWS variants
- [ ] Remove the model dependency (probably read from a file rendered by schnauzer)

jmt-lab:
**Issue number:** #3620

Closes #3620 

**Description of changes:**

Removes conditional compilation from pluto. 

* Changes pluto to fetch the necessary settings it needs via apiclient get cli
* Removes pluto from use as setting-generator calls
* Introduces pluto systemd unit to run to generate its settings after sundog and before settings-applier
* Uses apiclient set to set its generated settings
* New migration helper RemoveMetadata introduced along with unittests

**Testing done:**

	- [x] verify connectivity 
	- [x] verify connectivity 

Tested with downgrade -> upgrade 

Initial Build ID 1.20.0 0bb29776
Downgrade Build ID 1.19.5 64049ba
Upgrade Build ID 1.20.0 0bb29776

Verified that pluto in both situations generated correct settings
Verified that the new remove metadata does a noop on downgrade and setting-generators return
Verified that the new remove metadata removes setting-generator settings on upgrade

**Terms of contribution:**


sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,59 @@
...
+/// before template files are fully rendered. (i.e. Pluto self resolves some of the values that
+/// are used in the configuration file). To keep this behavior sane if the configuration file
+/// does not exist, return None and auto populate the region off IMDS
+pub(crate) async fn get_aws_k8s_info() -> Result<AwsK8sInfo> {
+    let path = Path::new(DEFAULT_CONFIG_PATH);
+    if path.exists() {
+        let config = tokio::fs::read_to_string(DEFAULT_CONFIG_PATH).await.context(ReadSnafu)?;
+        toml::from_str(config.as_str()).context(DeserializeTomlSnafu)
+    } else {
+        // Attempt to fetch the region from imds since AWS_REGION is not set
```
jmt-lab: The below change had to be done as there is a circular dependency with pluto and AWS region.

* Pluto requires the template file to exist
* However it gets called before said template file is generated by the system and is used to auto populate its own settings. 

@bcressey If you have a more elegant solution to this problem please let me know

jmt-lab:
Fix formatting

jmt-lab:
Rebased, build fixes, tests ran

packages/pluto/build.rs
```diff
@@ -0,0 +1,13 @@
+use std::path::PathBuf;
+use std::process::{exit, Command};
+
+fn main() -> Result<(), std::io::Error> {
+    let root_json_path = PathBuf::from(env!("PUBLISH_REPO_ROOT_JSON"));
+    println!("cargo:rerun-if-changed={}", root_json_path.display());
```
webern: This should not be a build time dependency. What is going on here? The package should use the same `build.rs` as all the other packages.

sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,61 @@
...
-        WrongVariant,
-    }
-
-    pub(crate) async fn get_aws_k8s_info() -> Result<AwsK8sInfo> {
-        WrongVariantSnafu.fail()
+/// Gets the info that we need to know about the EKS cluster from the configuration file.
+/// Note that we cannot rely on the configuration file being generated as pluto gets called
+/// before template files are fully rendered. (i.e. Pluto self resolves some of the values that
+/// are used in the configuration file). To keep this behavior sane if the configuration file
+/// does not exist, return None and auto populate the region off IMDS
```
webern: How do we know that we are getting the right behavior? i.e. how do we know that `cluster_dns_ip` is not actually set in the API if the configuration file is not there?

sources/api/settings-committer/src/main.rs
```diff
@@ -191,7 +191,7 @@ fn parse_args(args: env::Args) -> Args {
 
     Args {
         transaction: transaction.unwrap_or_else(|| constants::LAUNCH_TRANSACTION.to_string()),
-        log_level: log_level.unwrap_or(LevelFilter::Info),
+        log_level: log_level.unwrap_or(LevelFilter::Trace),
```
webern: Looks like a debugging/testing change to be undone.

webern (dismissed):
You need a clippy fix

sources/api/settings-committer/src/main.rs
jmt-lab: yup good catch

sources/api/pluto/src/api.rs
jmt-lab: I compared the resulting values between release bottlerocket and this branch and saw the same resolution of cluster_dns_ip occuring, if you have any idea of a better way to validate this do let me know

packages/pluto/build.rs
jmt-lab: I am not sure why i copied this out of os, it is a mistake, i shall strike it from the record haha

jmt-lab:
* Rebased on latest
* Removed redundant build.rs
* Undid debug change
* Removed unnecessary variable

bcressey:
> Migrates pluto out of os package to its own package

Why? Generally speaking the whimsical names shouldn't intrude into the visible surface area of the distro. For OOTBs that will also include package names.

Historical aside: `pluto` is so-named because it's a settings generator, and the program that invokes settings generators is called `sundog`, because it's the counterpart to `early-boot-config`, which used to be called `moondog`, which is the source of non-generated settings. `moondog` was one of the earliest names, back when the desert code name was "Atacama" instead of "Thar"; the Atacama Desert has the Very Large Telescope and hence there was an astronomical theme early on, before Thar let us unlock the pirate pun naming achievement. Bottlerocket was started in part by Amazon Linux veterans and observed the Amazon Linux tradition of using deserts as code names for major releases.

Back on topic: I am more OK with the generic ones like `updog` (for updates) and `netdog` (for networks) but would still prefer that they stay in the `os` aggregate package or else move to the appropriate `settings-*` package once those get rolling. For `netdog` in particular, splitting it out makes sense because there's a lot going on there and a separation of concerns is useful.

Aside from not wanting to export whimsy or to explain `pluto` ever again, I'm concerned about compile times exploding if we start compiling serde a hundred times in parallel.

packages/pluto/pluto-toml
```diff
@@ -0,0 +1,14 @@
...
++++
+{{#if_not_null settings.aws.region}}
+region = "{{{settings.aws.region}}}"
+{{/if_not_null}}
+{{#if_not_null settings.kubernetes.cluster-name}}
+cluster-name = "{{{settings.kubernetes.cluster-name}}}"
+{{/if_not_null}}
+{{#if_not_null settings.kubernetes.cluster-dns-ip}}
+cluster-dns-ip = "{{{settings.kubernetes.cluster-dns-ip}}}"
+{{/if_not_null}}
```
bcressey: For `pluto` we rely on the fact that it's executed by `sundog` which does its own API query for `settings.network` proxy settings and then sets them in the environment.

To keep the results consistent with `pluto` invoked manually vs. `pluto` invoked by `sundog`, it might be good to also include the proxy settings here and read them that way rather than from the environment.

There's actually a kind of a lot of `sundog` that caters to the needs of `pluto` so it'd be good to make it more independent. That will also help when we move to the newer approach for managing and generating settings.

variants/aws-k8s-1.23-nvidia/Cargo.toml
```diff
@@ -26,6 +26,7 @@ included-packages = [
     "nvidia-container-toolkit-k8s",
     "nvidia-k8s-device-plugin",
     "kmod-5.10-nvidia-tesla-470",
+    "pluto",
```
bcressey: If we must do this then let's start the `settings-kubernetes` package and put `pluto` there until it can be joined by its friend the settings extension.

packages/pluto/pluto-toml
```diff
@@ -0,0 +1,14 @@
+[required-extensions]
+aws = "v1"
+kubernetes = "v1"
+std = { version = "v1", helpers = ["if_not_null"]}
++++
+{{#if_not_null settings.aws.region}}
+region = "{{{settings.aws.region}}}"
```
bcressey: Does this work as expected if `pluto` runs on a system that doesn't know about `settings.aws`?

Another way of asking (or testing) is if this would blow up:
```
{{#if_not_null settings.xyzzy.blorb}}
blorb = "{{{settings.xyzzy.blorb}}}"
{{/if_not_null}}
```

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -26,9 +26,20 @@ affected-services = ["containerd", "kubernetes", "host-containerd", "host-contai
 configuration-files = ["warm-pool-wait-toml"]
 restart-commands = []
 
+[services.pluto]
+configuration-files = ["pluto-toml"]
+restart-commands = []
```
bcressey: This is a little scary in that if we had a restart command here, it could be invoked in an infinite loop since `pluto` generates the same settings that it's affected by.

Is this the first time we've touched a settings generator in this conversion effort? I'm not confident this is the right approach architecturally; settings generators have a more complex producer/consumer relationship with the settings API.

I wonder if the `schnauzer-v2 render ...` machinery is really the better way to go here.

sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,60 @@
...
+pub(crate) async fn get_aws_k8s_info() -> Result<AwsK8sInfo> {
+    match tokio::fs::read_to_string(DEFAULT_CONFIG_PATH).await {
+        Ok(config_str) => toml::from_str(config_str.as_str()).context(DeserializeTomlSnafu),
+        Err(e) => {
+            if e.kind() == ErrorKind::NotFound {
+                let mut imds = imdsclient::ImdsClient::new();
+                let region = imds.fetch_region().await.unwrap();
```
bcressey: This seems like a potentially incompatible change from the previous logic. Before we would just return `None` for all three values in `AwsK8sInfo` if the setting wasn't present; now we generated one of the fields from IMDS. Why?

jmt-lab: AWS Region is required at the first time pluto is called. In my testing if Region was None it would fail to boot

packages/pluto/pluto-toml
jmt-lab: Pluto is only ever included on aws k8s variants as far as I'm aware.

jmt-lab:
> > Migrates pluto out of os package to its own package
> 
> Why? Generally speaking the whimsical names shouldn't intrude into the visible surface area of the distro. For OOTBs that will also include package names.
> 
> Historical aside: `pluto` is so-named because it's a settings generator, and the program that invokes settings generators is called `sundog`, because it's the counterpart to `early-boot-config`, which used to be called `moondog`, which is the source of non-generated settings. `moondog` was one of the earliest names, back when the desert code name was "Atacama" instead of "Thar"; the Atacama Desert has the Very Large Telescope and hence there was an astronomical theme early on, before Thar let us unlock the pirate pun naming achievement. Bottlerocket was started in part by Amazon Linux veterans and observed the Amazon Linux tradition of using deserts as code names for major releases.
> 
> Back on topic: I am more OK with the generic ones like `updog` (for updates) and `netdog` (for networks) but would still prefer that they stay in the `os` aggregate package or else move to the appropriate `settings-*` package once those get rolling. For `netdog` in particular, splitting it out makes sense because there's a lot going on there and a separation of concerns is useful.
> 
> Aside from not wanting to export whimsy or to explain `pluto` ever again, I'm concerned about compile times exploding if we start compiling serde a hundred times in parallel.

I was just following netdog since we pulled it out when doing a similar conditional removement, I'm okay sliding this back into os.spec if that is preferred

jmt-lab:
Updated to ensure the compilation changes with pluto & cfsignal work well. Note I will be addressing shibaken and other remaining if blocks in another PR

packages/os/os.spec
```diff
@@ -214,24 +212,22 @@ Summary: Bottlerocket certificates handler
...
 %description -n %{_cross_os}pluto
 %{summary}.
-%endif
 
 %if %{with aws_platform}
 %package -n %{_cross_os}shibaken
 Summary: Run tasks reliant on IMDS
 %description -n %{_cross_os}shibaken
 %{summary}.
+%endif
```
bcressey: You may as well give `shibaken` and `driverdog` the same treatment as part of the wider conditional compilation removal effort. You have done most of that work here.

You can group those changes (plus the ones for pluto) into an initial "os: replace bconds with boolean deps" commit.

packages/os/os.spec
```diff
@@ -641,13 +630,13 @@ install -p -m 0644 %{S:400} %{S:401} %{S:402} %{buildroot}%{_cross_licensedir}
 %{_cross_templatedir}/warm-pool-wait-toml
 %{_cross_unitdir}/warm-pool-wait.service
 %endif
+%endif
```
bcressey: `warm-pool-wait.service` is an interesting conversion from conditional compilation:
* requires shibaken
* should be in a separate package
* only installed for aws and k8s runtime

Top-level OS requirement would be this:
```
Requires: (%{_cross_os}warm-pool-wait if %{_cross_os}variant-family(aws-k8s))
```

`warm-pool-wait` package requirements:
```
Requires: %{_cross_os}shibaken
```

packages/os/pluto.service
```diff
@@ -0,0 +1,14 @@
+# pluto.service
```
bcressey: omit - files don't need a comment about their filename

packages/os/pluto.service
```diff
@@ -0,0 +1,14 @@
...
+[Unit]
+Description=Generate additional settings for Kubernetes
+After=sundog.service
+Requires=sundog.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/pluto
```
bcressey, line 15: These early services should commit settings first if they might depend on them being available:
```suggestion
ExecStartPre=/usr/bin/settings-committer
ExecStart=/usr/bin/pluto
```

packages/os/pluto.service
```diff
@@ -0,0 +1,14 @@
+# pluto.service
+[Unit]
+Description=Generate additional settings for Kubernetes
+After=sundog.service
+Requires=sundog.service
```
bcressey: From a systemd unit correctness perspective, `pluto` should include more of its own dependencies, it's sort of drafting off of `sundog.service` here.

It expects the network to be up and the API server to be available:
```suggestion
After=network-online.target apiserver.service sundog.service
Requires=sundog.service
# We don't want to restart the unit if the network goes offline or apiserver restarts
Wants=network-online.target apiserver.service
# Block manual interactions with this service, since it could leave the system in an
# unexpected state
RefuseManualStart=true
RefuseManualStop=true
```

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -1260,6 +1260,136 @@ mod test_add_metadata {
...
+        .backward(data)
+        .unwrap();
+        assert_eq!(
+            result.metadata,
+            hashmap! {
+                "hi".into() => hashmap!{"sup".into() => "wassup".into()},
+            }
+        );
+    }
+}
```
bcressey: I don't think you need this test unless I'm missing something subtle, since backwards is always a no-op.

packages/os/pluto-toml
bcressey, line 1: Delete this empty file?

sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,101 @@
...
+        serde_json::from_slice(view_str.as_slice()).context(DeserializeSnafu)?;
+    Ok(AwsK8sInfo {
+        region: view.settings.aws.and_then(|a| a.region),
+        cluster_name: view
+            .settings
+            .kubernetes
+            .as_ref()
+            .and_then(|k| k.cluster_name.clone()),
+        cluster_dns_ip: view.settings.kubernetes.and_then(|k| k.cluster_dns_ip),
+    })
```
bcressey: One of the nice things about this refactor is that we can generate all the settings in one invocation, instead of calling `pluto` four times. That allows a more efficient implementation.

We can call `get_aws_k8s_info()` once and then pass `AwsK8sInfo` into each of the functions that needs it:
1. `get_cluster_dns_ip`
2. `get_node_ip` (also needs the cluster DNS IP, and can receive that as a parameter vs calling `get_cluster_dns_ip` again)
3. `get_private_dns_name`

sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,101 @@
...
-                .and_then(|k| k.cluster_name.clone())
-                .map(|s| s.into()),
-            cluster_dns_ip: settings.kubernetes.and_then(|k| k.cluster_dns_ip),
-        })
-    }
+#[derive(Debug, Snafu)]
+pub(crate) enum Error {
+    #[snafu(display("Failed to call apiclient: {}", source))]
+    CommandFailure { source: std::io::Error },
+    #[snafu(display("APIClient execution failed: {}", reason))]
```
bcressey:
```suggestion
    #[snafu(display("apiclient execution failed: {}", reason))]
```

sources/api/pluto/src/main.rs
```diff
@@ -332,55 +319,42 @@ async fn get_private_dns_name(client: &mut ImdsClient) -> Result<String> {
...
-        println!("{}", max_pods);
+    // Generate the node-ip
+    let node_ip = get_node_ip(&mut client).await?;
+    // Generate the cluster-dns-ip
+    let cluster_dns_ip = get_cluster_dns_ip(&mut client).await?;
+
+    // Generate max-pods
+    // Skip the set if this fails
+    let max_pods = if let Ok(max_pods) = get_max_pods(&mut client).await {
+        Some(max_pods)
```
bcressey: Why skip the set? This seems to be new behavior - before we would error out if we couldn't get a max pods value. The conditional was just to deal with sundog's weird output requirements; we could print the number directly but needed to JSON-format the other values.

jmt-lab:
* Fixed unwrap in get_max_pods
* Remove build_variant

packages/os/os.spec
```diff
@@ -96,15 +97,12 @@ Requires: %{_cross_os}thar-be-settings
...
-%if %{with aws_k8s_family}
-Requires: %{_cross_os}pluto
-%endif
-
 %if %{with aws_platform}
 Requires: %{_cross_os}shibaken
-Requires: %{_cross_os}cfsignal
 %endif
 
+Requires: (%{_cross_os}cfsignal if %{_cross_os}variant-platform(aws))
```
webern, line 102: @bcressey Is not our permanent solution, is it? I guess if the variant name is different `weird-k8s-1.27` the user can still include these things by explicitly placing them in variant packages list? My concern is just that the `aws` comes from the variant name. There's a lot of magic in the variant name.

sources/api/pluto/src/main.rs
jmt-lab: Discussed offline, we should be good here.

sources/api/migration/migration-helpers/src/common_migrations.rs
webern: > I don't think you need this test unless I'm missing something subtle, since backwards is always a no-op.

Since this migration is removing something, backwards should put it back. Right?

sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,101 @@
...
+    },
 }
 
-/// This dummy code is compiled when the `sources` workspace is being compiled for non `aws-k8s-*`
-/// variants.
-#[cfg(not(variant_family = "aws-k8s"))]
-mod inner {
-    use super::*;
-    use snafu::Snafu;
+pub(crate) async fn client_command<I, S>(args: I) -> Result<Vec<u8>>
```
webern, line 97: Could you do the json parse in this function and return a `Result<serde_json::Value>`?

sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,101 @@
...
+    .await?;
+    let view: SettingsView =
+        serde_json::from_slice(view_str.as_slice()).context(DeserializeSnafu)?;
+    Ok(AwsK8sInfo {
+        region: view.settings.aws.and_then(|a| a.region),
+        cluster_name: view
+            .settings
+            .kubernetes
+            .as_ref()
+            .and_then(|k| k.cluster_name.clone()),
```
webern, line 154: Isn't the cluster_name eventually required? Can it be required here instead of being an `Option`?

webern (commented):
Is the empty file `packages/os/pluto-toml` intentional? If so can you explain why you need an empty file?

sources/api/migration/migration-helpers/src/common_migrations.rs
bcressey: > Since this migration is removing something, backwards should put it back. Right?

`storewolf` will put it back

packages/os/os.spec
bcressey: > the user can still include these things by explicitly placing them in variant packages list?

That, or by adding this to one of their packages that's explicitly installed:
```
Provides: %{_cross_os}variant-platform(aws)
```

Having a `<ootb>-release` package to collect things like this will be a good convention.

sources/api/migration/migration-helpers/src/common_migrations.rs
jmt-lab: Yeah I validated in my testing that storewolf indeed put them back

sources/api/pluto/src/api.rs
jmt-lab: I'll need to validate that it doesn't break anything to require it and not use an Option. Generally I followed our convention that #[model] does

sources/api/pluto/src/api.rs
jmt-lab: I can move it here, but let me double check cause i believe i also use this for apiclient set command which doesn't return json

jmt-lab: Yeah not sure its a good idea to change to serde_json since it is used for set as well which does not return valid json afaik

sources/api/pluto/src/api.rs
jmt-lab: While I think you are probably right, talked offline with ben to see if there are cases where it isn't set and don't think there is need to really to take the risk of being surprised by it if it fails

jmt-lab:
Removed unnecessary test

jmt-lab:
Interactive rebase done to better arrange changes

jmt-lab:
Rebased to squash commits and separate migration
Rebased off latest develop

sources/api/pluto/src/main.rs
```diff
@@ -327,60 +310,54 @@ async fn get_private_dns_name(client: &mut ImdsClient) -> Result<String> {
...
+    // Generate private-dns-name
+    let private_dns_name = get_private_dns_name(&mut client, &aws_k8s_info).await?;
+
+    let generated_settings: serde_json::Value = serde_json::json!({
+        "kubernetes": {
+            "node-ip": node_ip,
+            "cluster-dns-ip": cluster_dns_ip,
+            "max-pods": max_pods,
+            "provider-id": provider_id,
+            "hostname-override": private_dns_name
```
bcressey: [KubernetesSettings](https://github.com/bottlerocket-os/bottlerocket/blob/64eff582e8738ce9e4dbf21f485791b225bac319/sources/models/src/lib.rs#L316) says:
```
// Generated in `aws-k8s-1.26*` variants only
```

This is a problem the new `pluto` will have to handle somehow. At the time in #3033, we felt the risk of a breaking behavior change from generating this as a new default was too great, so it only applied to 1.26 and beyond.

One option would be a `pluto.toml` that we can package in the different Kubernetes packages, but that's going to be weird since they're not currently platform-specific.

Another would be to pull the variant ID out of `apiclient get os.variant_id` and see if it matches the problematic prefixes - `aws-k8s-1.23` through `aws-k8s-1.25`, and if so don't include a hostname override.

sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,101 @@
...
-        WrongVariantSnafu.fail()
-    }
+/// Gets the info that we need to know about the EKS cluster from the Bottlerocket API.
+pub(crate) async fn get_aws_k8s_info() -> Result<AwsK8sInfo> {
+    let view_str = client_command(&[
+        "get",
+        "settings.aws.region",
+        "settings.kubernetes.cluster-name",
+        "settings.kubernetes.cluster-dns-ip",
+    ])
```
bcressey: This is a logic error I missed in earlier reviews - we also need to retrieve any settings that we might decide to set, and if we find a value, we must not clobber it in the later set call.

bcressey (changes requested):

sources/api/migration/migration-helpers/src/common_migrations.rs
```diff
@@ -1260,6 +1260,114 @@ mod test_add_metadata {
...
+                    setting_metadata.setting
+                );
+            }
+        }
+        Ok(input)
+    }
+
+    fn backward(&mut self, input: MigrationData) -> Result<MigrationData> {
+        println!(
+            "RemoveMetadataMigration({:?}) has no work to do on downgrade.",
```
webern: Double-checking this is right, @bcressey. We don't need to do anything on downgrade because the defaults will cause the metadata to be put back into place on downgrade, right?

webern (approved):
- The Pluto code looks correct to me, to the best of my ability with the caveat that I can't quite tell if there could be a subtle behavior change based on the fact that Pluto used to be called at different times for these things and may have had different settings available when that occurred.
- The migration seems probably correct, though I have a question about "nothing to do on downgrade".
- I mostly deferring to Ben on packaging concerns.

sources/api/migration/migration-helpers/src/common_migrations.rs
bcressey: > Double-checking this is right, @bcressey. We don't need to do anything on downgrade because the defaults will cause the metadata to be put back into place on downgrade, right?

Yes. `migrator.service` runs before `storewolf.service` and (simplifying) is responsible for fixing things up so the datastore that `storewolf` deserializes matches what the model expects.

`storewolf` will add any new defaults to an existing data store (since 036be7876b7e92f8f50063b5faded8e799912ab5) but does not touch existing data.

One conclusion I've drawn from this - frankly downright weird! - migration is that there's probably something much smarter we could do for the datastore files that cannot be set via the API, such as "remove all existing metadata so that `storewolf` puts it back".

But that would fall outside the scope of what we can cover in any one set of up/down migrations; we'd need the OS itself to be smart enough to do that, which means doing the repo pivot trick to move to the new world where we can treat it as a given.

If we did that, we could expand the scope to include configuration-files and services too. Might as well throw in unrecognized prefixes and individual files while we're at it, and the ability to set a field to `{{ default }}` to force it to be re-evaluated to its current default value. Potentially we could eliminate almost all of what we use migrations for even in the current settings system.

sources/api/pluto/src/main.rs
```diff
@@ -327,60 +315,74 @@ async fn get_private_dns_name(client: &mut ImdsClient) -> Result<String> {
...
-        )
-        .context(error::OutputJsonSnafu { output: &setting })?;
-        println!("{}", max_pods);
+    // Generate the cluster-dns-ip
+    let cluster_dns_ip = if let Some(ip) = aws_k8s_info
+        .cluster_dns_ip
+        .as_ref()
+        .and_then(|x| x.iter().next())
+    {
+        *ip
```
bcressey: I would prefer that `get_cluster_dns_ip` handle this logic - check the passed settings struct for the value, and if so return it immediately.

sources/api/pluto/src/main.rs
```diff
@@ -327,60 +315,74 @@ async fn get_private_dns_name(client: &mut ImdsClient) -> Result<String> {
...
     } else {
-        let output =
-            serde_json::to_string(&setting).context(error::OutputJsonSnafu { output: &setting })?;
-        println!("{}", output);
-    }
+        let ip = get_cluster_dns_ip(&mut client, &aws_k8s_info).await?;
+        IpAddr::from_str(&ip).context(error::BadIpSnafu { ip: ip.clone() })?
```
bcressey: This `IpAddr::from_str` check should happen within the function - it shouldn't return `Ok` for an invalid IP address.

sources/api/pluto/src/main.rs
```diff
@@ -327,60 +315,74 @@ async fn get_private_dns_name(client: &mut ImdsClient) -> Result<String> {
...
+    };
+
+    // Generate the node-ip
+    let node_ip = if let Some(node_ip) = aws_k8s_info.node_ip.as_ref() {
+        node_ip.clone()
+    } else {
+        get_node_ip(&mut client, cluster_dns_ip).await?
+    };
```
bcressey: Similarly, `get_node_ip` can take the `AwsK8sInfo` parameter and just return the value from it if present. Then we avoiding splitting the logic about what value to set across functions.

Might even be better to rename these functions to `generate_<setting>` and pass `&mut AwsK8sInfo` and just let the function take care of it.

sources/api/pluto/src/main.rs
```diff
@@ -327,60 +315,74 @@ async fn get_private_dns_name(client: &mut ImdsClient) -> Result<String> {
...
+
+    // Generate private-dns-name
+    let private_dns_name = if NO_HOSTNAME_VARIANTS.contains(&aws_k8s_info.variant_id.as_str()) {
+        None
+    } else if let Some(hostname_override) = aws_k8s_info.hostname_override {
+        Some(hostname_override.clone())
+    } else {
+        Some(get_private_dns_name(&mut client, &aws_k8s_info).await?)
+    };
```
bcressey: These chained `if` statements make it hard to follow which values we actually have and when.

Refactoring all of this along these lines could make it more clear and be more straightforward to extend in the future:
```
generate_cluster_dns_ip(&mut client, &mut aws_k8s_info).await?;
generate_node_ip(&mut client, &mut aws_k8s_info).await?;
generate_max_pods(&mut client, &mut aws_k8s_info).await?;
generate_private_dns_name(&mut client, &mut aws_k8s_info).await?;
```

Even better would be to keep `aws_info` and `k8s_info` separate, since we don't expect to mutate the AWS-related fields. Then when you're done calling the generate functions, you can serialize the k8s struct and return that.

There's an ordering dependency where we must run `generate_cluster_dns_ip` first, before we run `generate_node_ip`. We could enforce that by having `generate_node_ip` call `generate_cluster_dns_ip` right away - if it does an early check / return for an existing value, that will be quick, and then the main function doesn't have to risk getting it wrong.

sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,151 @@
...
 
-    pub(crate) async fn get_aws_k8s_info() -> Result<AwsK8sInfo> {
-        WrongVariantSnafu.fail()
-    }
+/// Gets the info that we need to know about the EKS cluster from the Bottlerocket API.
+pub(crate) async fn get_aws_k8s_info() -> Result<AwsK8sInfo> {
+    let view_str = client_command(&[
+        "get",
+        "os.variant-id",
+        "settings.aws.region",
```
bcressey: You would also need to retrieve the proxy server settings:
```
settings.network.https-proxy
settings.network.no-proxy
```

And ensure that they are set in the environment where they can be found by `proxy::fetch_proxy_env`.

This is something `sundog` does on behalf of settings generators.

Note that `sundog` has [k8s-specific no_proxy logic](https://github.com/bottlerocket-os/bottlerocket/blob/fcf71a47c0ff005327ecde870d8a70877fc89196/sources/api/sundog/src/main.rs#L281), but it appears this is not needed since we don't connect to those hosts.

jmt-lab:
* Rebased on latest
* Refactored to use generate methods that pass aws_k8s_info
* Triggered node_ip off cluster_dns
* Retested everything functioned as expected

webern (approved):

sources/api/pluto/src/main.rs
```diff
@@ -176,27 +173,51 @@ async fn get_max_pods(client: &mut ImdsClient) -> Result<String> {
...
 /// `10` for the last octet. If the EKS call is not successful, it falls back to using IMDS MAC CIDR
 /// blocks to return one of two default addresses.
-async fn get_cluster_dns_ip(client: &mut ImdsClient) -> Result<String> {
+async fn generate_cluster_dns_ip(
+    client: &mut ImdsClient,
+    aws_k8s_info: &mut AwsK8sInfo,
+) -> Result<()> {
+    if aws_k8s_info.cluster_dns_ip.is_some() {
+        return generate_node_ip(client, aws_k8s_info).await;
+    }
```
bcressey: It's weird to overload `generate_cluster_dns_ip` to also run `generate_node_ip`.

```suggestion
async fn generate_cluster_dns_ip(
    client: &mut ImdsClient,
    aws_k8s_info: &mut AwsK8sInfo,
) -> Result<()> {
    if aws_k8s_info.cluster_dns_ip.is_some() {
        return Ok(());
    }
```

sources/api/pluto/src/main.rs
```diff
@@ -176,27 +173,51 @@ async fn get_max_pods(client: &mut ImdsClient) -> Result<String> {
...
+        // values based on the IPv4 cidr range of our primary network interface
+        get_ipv4_cluster_dns_ip_from_imds_mac(client).await?
+    };
+
+    aws_k8s_info.cluster_dns_ip = Some(KubernetesClusterDnsIp::Scalar(
+        IpAddr::from_str(ip_addr.as_str()).context(error::BadIpSnafu {
+            ip: ip_addr.clone(),
+        })?,
+    ));
+    generate_node_ip(client, aws_k8s_info).await
```
bcressey:
```suggestion
    Ok(())
```

sources/api/pluto/src/main.rs
```diff
@@ -260,24 +281,18 @@ async fn get_ipv4_cluster_dns_ip_from_imds_mac(client: &mut ImdsClient) -> Resul
...
-async fn get_node_ip(client: &mut ImdsClient) -> Result<String> {
-    // Retrieve the user specified cluster DNS IP if it's specified, otherwise use the pluto-generated
-    // cluster DNS IP
-    let aws_k8s_info = api::get_aws_k8s_info().await.context(error::AwsInfoSnafu)?;
-    let configured_cluster_dns_ip = aws_k8s_info
+async fn generate_node_ip(client: &mut ImdsClient, aws_k8s_info: &mut AwsK8sInfo) -> Result<()> {
+    if aws_k8s_info.node_ip.is_some() {
+        return Ok(());
+    }
+    let cluster_dns_ip = aws_k8s_info
```
bcressey, line 295: once the recursion is avoided:
```suggestion
    generate_cluster_dns_ip()?;
    let cluster_dns_ip = aws_k8s_info
```

sources/api/pluto/src/main.rs
```diff
@@ -327,60 +360,32 @@ async fn get_private_dns_name(client: &mut ImdsClient) -> Result<String> {
...
-    args.nth(1).unwrap_or_else(|| usage())
+    Ok(())
 }
 
 async fn run() -> Result<()> {
-    let setting_name = parse_args(env::args());
     let mut client = ImdsClient::new();
+    let mut aws_k8s_info = api::get_aws_k8s_info().await.context(error::AwsInfoSnafu)?;
+
+    generate_cluster_dns_ip(&mut client, &mut aws_k8s_info).await?;
```
bcressey, line 387:
```suggestion
    generate_cluster_dns_ip(&mut client, &mut aws_k8s_info).await?;
    generate_node_ip(&mut client, &mut aws_k8s_info).await?;
```

sources/api/pluto/src/main.rs
```diff
@@ -141,7 +129,15 @@ use error::PlutoError;
...
-async fn get_max_pods(client: &mut ImdsClient) -> Result<String> {
+async fn generate_max_pods(client: &mut ImdsClient, aws_k8s_info: &mut AwsK8sInfo) -> Result<()> {
+    if aws_k8s_info.max_pods.is_some() {
+        return Ok(());
+    }
+    aws_k8s_info.max_pods = get_max_pods(client).await.ok();
+    Ok(())
+}
+
+async fn get_max_pods(client: &mut ImdsClient) -> Result<u32> {
```
bcressey, line 140: Why split max pods into "generate" and "get" but not the others? Might be cleaner to do the same for all.

sources/api/pluto/src/api.rs
```diff
@@ -1,76 +1,183 @@
...
+        env::set_var("HTTPS_PROXY", https_proxy.clone());
     }
+    if let Some(no_proxy) = view
+        .settings
+        .network
+        .as_ref()
+        .and_then(|x| x.no_proxy.as_ref())
+    {
+        env::set_var("no_proxy", no_proxy.clone());
+        env::set_var("NO_PROXY", no_proxy.clone());
```
bcressey: `env::set_var` is unsafe in a multi-threaded context - which we may be here in because we're running in an async runtime.

It's probably OK here because we know that no other tasks are running, but we could avoid the problem by keeping these settings in the struct and passing them to the ec2 / eks functions rather than reading from the environment.

sources/api/pluto/src/main.rs
jmt-lab: This was to maintain the approach of not setting it should get_max_pods fails which is unique to this.

sources/api/pluto/src/main.rs
jmt-lab: This was in response to a previous comment you have made when proposing the generate change :D but I can undo it no problem.

bcressey (approved):
🪐

PR #3836, opened 2024-03-20
prairiedog: if the variant doesn't support boot settings be okay with…

jmt-lab:
… None

**Description of changes:**

* Made the reading method to return None should the file not exist.

**Testing done:**

* Launched aws-ecs-1 and ensured it properly joined the ecs cluster

**Terms of contribution:**

### DIFF
--- a/sources/api/prairiedog/src/bootconfig.rs
+++ b/sources/api/prairiedog/src/bootconfig.rs
@@ -6,6 +6,7 @@ use serde::{Deserialize, Serialize};
 use snafu::{ensure, ResultExt};
 use std::collections::HashMap;
 use std::convert::TryInto;
+use std::io::ErrorKind;
 use std::path::Path;
 use std::{fs, io};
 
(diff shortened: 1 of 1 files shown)


sources/api/prairiedog/src/bootconfig.rs
```diff
@@ -117,8 +117,14 @@ fn get_boot_config_settings<P>(config_path: P) -> Result<Option<BootSettings>>
...
-    let config_str = fs::read_to_string(config_path.as_ref()).context(error::ReadFileSnafu {
-        path: config_path.as_ref().to_path_buf(),
+    let config_path = config_path.as_ref();
+    if !config_path.try_exists().context(error::ReadFileSnafu {
+        path: config_path.to_path_buf(),
+    })? {
+        return Ok(None);
+    }
+    let config_str = fs::read_to_string(config_path).context(error::ReadFileSnafu {
+        path: config_path.to_path_buf(),
```
bcressey: Commit message could be reworded, it's a bit hard to parse right now.

There's a TOCTOU issue here where `fs::read_to_string` could fail if the file is removed before it runs and after the `try_exists` check. To avoid this, and the redundant check, it might be better not to use `fs::read_to_string` at all.

jmt-lab: For some context, toml deprecated its from_slice parser, also part of the reason to utilize try_exists above instead of exists() is this very problem, exists is prone to failure on the exact situation you described where try_exists can validate this. Though I wonder it may be better to treat a failure of config_path.try_exists() as equivalent to no file found.

jmt-lab: Updated per discussion

webern (approved):
I'm not clear on why we are hitting these cases where a template file isn't rendered yet. I will need help creating a mental model of what is happening during template rendering (i.e. order of ops).

arnaldo2792:
@webern, the case that I saw was that in `aws-ecs-1`, the `bootconfig.toml` defaults aren't used. With the ongoing work to use configuration files instead of API calls, the generation of the configuration for prairiedog was moved to the API. Since the service that creates the configuration file is added by the `bootconfig.toml` defaults, the file is never rendered for `aws-ecs-1`.

arnaldo2792 (approved):

PR #3838, opened 2024-03-20
schnauzer: add if_not_null template helper

sam-berning:
**Issue number:**

N/A

**Description of changes:**

There's currently not a safe way to render potentially-null falsy values in a schnauzer template. Gating with `{{#if falsy}}` will skip rendering, and `{{default false falsy}}` will always render something, even if `falsy` is unset.

`{{#if-not-null falsy}}` checks only that the value is not null, it doesn't care if it's falsy. This allows us to conditionally render template portions that may contain falsy values.

So for example, in `packages/os/host-containers-toml`, we currently render:

```
{{#each settings.host-containers}}
[host-containers."{{{@key}}}"]
...
{{#if this.enabled}}
enabled = {{{this.enabled}}}
{{/if}}
{{#if this.superpowered}}
superpowered = {{{this.superpowered}}}
{{/if}}
...
{{/each}}
```

where `this.enabled` and `this.superpowered` are booleans. If they are explicitly set to `false`, they will not be rendered. In this particular case it doesn't cause a behavior issue because those settings default to `false`, but it could cause problems in a case where the default is `true`.

With this helper, we can instead write:
```
{{#each settings.host-containers}}
[host-containers."{{{@key}}}"]
...
{{#if-not-null this.enabled}}
enabled = {{{this.enabled}}}
{{/if-not-null}}
{{#if-not-null this.superpowered}}
superpowered = {{{this.superpowered}}}
{{/if-not-null}}
...
{{/each}}
```

This will correctly render `enabled = false` and `superpowered = false`.

Related: https://github.com/bottlerocket-os/bottlerocket/pull/3724, https://github.com/bottlerocket-os/bottlerocket/pull/3777

**Testing done:**

updated the `host-containers-toml` template using the `{{#if-not-null}}` helper, and configured settings to reflect the three different states we care about (set to a truthy value, set to a falsy value, and unset) and confirmed that the template rendered as expected.

**Terms of contribution:**

### DIFF
--- a/sources/api/schnauzer/src/helpers.rs
+++ b/sources/api/schnauzer/src/helpers.rs
@@ -4,7 +4,8 @@
 
 use dns_lookup::lookup_host;
 use handlebars::{
-    handlebars_helper, Context, Handlebars, Helper, Output, RenderContext, RenderError,
+    handlebars_helper, Context, Handlebars, Helper, HelperDef, Output, RenderContext, RenderError,
+    Renderable,
 };
 use lazy_static::lazy_static;
 use model::modeled_types::{OciDefaultsCapability, OciDefaultsResourceLimitType};
@@ -675,6 +676,42 @@ pub fn default(
         template: template_name.to_owned(),
     })?;
     Ok(())
+}
+
+/// The `if_not_null` helper is used to check when a value is not null. This is
+/// useful especially for falsy values such as `false`, `0`, or `""` to
+/// distinguish between "not set" and "false".
+#[derive(Clone, Copy)]
+pub struct IfNotNullHelper;
+
+impl HelperDef for IfNotNullHelper {
+    fn call<'reg: 'rc, 'rc>(
+        &self,
+        helper: &Helper<'reg, 'rc>,
+        registry: &'reg Handlebars<'reg>,
+        ctx: &'rc Context,
+        renderctx: &mut RenderContext<'reg, 'rc>,
+        out: &mut dyn Output,
+    ) -> Result<(), RenderError> {
+        trace!("Starting if_not_null helper");
+        let template_name = template_name(renderctx);
+        trace!("Template name: {}", &template_name);
+
+        trace!("Number of params: {}", helper.params().len());
+        check_param_count(helper, template_name, 1)?;
+
+        let param = get_param(helper, 0)?;
+
+        let tmpl = if !param.is_null() {
+            helper.template()
+        } else {
+            helper.inverse()
+        };
+        match tmpl {
+            Some(t) => t.render(registry, ctx, renderctx, out),
+            None => Ok(()),
+        }
+    }
 }
 
 /// The `ecr-prefix` helper is used to map an AWS region to the correct ECR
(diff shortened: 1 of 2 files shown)


jmt-lab (approved):
LGTM

sam-berning:
^ fixed doc comment

sources/api/schnauzer/src/helpers.rs
```diff
@@ -677,6 +678,42 @@ pub fn default(
...
+impl HelperDef for IfNotNullHelper {
+    fn call<'reg: 'rc, 'rc>(
+        &self,
+        helper: &Helper<'reg, 'rc>,
+        registry: &'reg Handlebars<'reg>,
+        ctx: &'rc Context,
+        renderctx: &mut RenderContext<'reg, 'rc>,
+        out: &mut dyn Output,
+    ) -> Result<(), RenderError> {
+        trace!("Starting default helper");
```
arnaldo2792: This is not the default helper

sources/api/schnauzer/src/helpers.rs
```diff
@@ -2080,6 +2117,86 @@ mod test_default {
...
+        registry.register_helper("if-not-null", Box::new(IfNotNullHelper));
+
+        registry.render_template(tmpl, data)
+    }
+
+    #[test]
+    fn null_value() {
+        let result = setup_and_render_template(
+            "{{#if-not-null setting}}foo{{/if-not-null}}",
+            &json!({"not-the-setting": 42.42}),
```
webern: Why is 42.42 the same as null?

sources/api/schnauzer/src/v2/import/helpers.rs
```diff
@@ -66,6 +66,7 @@ fn all_helpers() -> HashMap<ExtensionName, HashMap<HelperName, Box<dyn HelperDef
             "default" => helper!(handlebars_helpers::default),
             "join_array" => helper!(handlebars_helpers::join_array),
             "join_map" => helper!(handlebars_helpers::join_map),
+            "if-not-null" => Box::new(handlebars_helpers::IfNotNullHelper),
```
webern: It looks like the naming convention is underscore:

```suggestion
            "if_not_null" => Box::new(handlebars_helpers::IfNotNullHelper),
```

arnaldo2792 (commented):
Could you please add in the description how this helper improves or changes how we render settings today? I didn't review the other PRs so I don't fully understand how this will affect templates in the future (e.g. I had a similar problem in another PR and we could probably solve both problems with one helper)

sources/api/schnauzer/src/helpers.rs
```diff
@@ -2080,6 +2117,86 @@ mod test_default {
...
+        let mut registry = Handlebars::new();
+        registry.register_helper("if-not-null", Box::new(IfNotNullHelper));
+
+        registry.render_template(tmpl, data)
+    }
+
+    #[test]
+    fn null_value() {
+        let result = setup_and_render_template(
+            "{{#if-not-null setting}}foo{{/if-not-null}}",
```
webern: Is it possible to create this so that the natural/built-in conditionals work as expected, i.e. the native `#if`?

```suggestion
            "{{#if not_null setting}}foo{{/if}}",
```

sam-berning: I couldn't get this to work. for some reason handlebars treats a boolean return from a helper as a string, so the `#if` helper resolves to `true` when we return `false` from `not_null`.

it could work if I returned an empty string on false, because handlebars treats empty string as falsy, but it felt hackier to me and essentially limits us to using it in the same scenarios as `#if-not-null`

sam-berning:
> Could you please add in the description how this helper improves or changes how we render settings today? I didn't review the other PRs so I don't fully understand how this will affect templates in the future (e.g. I had a similar problem in another PR and we could probably solve both problems with one helper)

updated description with an example

sources/api/schnauzer/src/v2/import/helpers.rs
sam-berning: the naming convention is a bit all over the place (see `ecr-prefix`, `tuf-prefix` above), but I'm good to stick with the convention within `std`

sources/api/schnauzer/src/helpers.rs
sam-berning: it's not, I'm just specifying another setting with a dummy value. I suppose this test would be the same if I just passed `{}` as the settings json, so maybe that's more readable. was just following the convention of other tests in this file

sources/api/schnauzer/src/helpers.rs
arnaldo2792: > I couldn't get this to work. for some reason handlebars treats a boolean return from a helper as a string

That's because the value returned by the helpers is a string, not a boolean. I had the same issue.

arnaldo2792:
For your use case, wouldn't it be simpler to just  do:

```hbs
{{#each settings.host-containers}}
[host-containers."{{{@key}}}"]
...
enabled = {{default false this.enabled}}
superpowered = {{default true this.superpowered}}
...
{{/each}}
```

We know that by default, those two values are `false`, it should be OK to just default to those values.

bcressey (approved):
Nice!

sam-berning:
> For your use case, wouldn't it be simpler to just do:
> 
> ```
> {{#each settings.host-containers}}
> [host-containers."{{{@key}}}"]
> ...
> enabled = {{default false this.enabled}}
> superpowered = {{default true this.superpowered}}
> ...
> {{/each}}
> ```
> 
> We know that by default, those two values are `false`, it should be OK to just default to those values.

Yeah that would work behavior-wise in this case. But I don't think we want both `host-containers` and the template to have to be aware and manage default values for the settings. This approach leans towards giving that responsibility fully to `host-containers`, but @bcressey also proposed that we could set defaults in the config file and make it an error for a setting to be unset in `host-containers`.

bcressey:
I am sympathetic toward @sam-berning's view that `if_not_null` lets us directly reuse the existing logic from "read settings from API, deal with None values somehow" to "read settings from config, deal with None values somehow".

We can evolve the client programs and the templates over time but I like `if_not_null` because it gives us a simple rule to follow in the code reviews ("always use `if_not_null`, not `if`, for these conversions") that lets us avoid the surprises (empty strings and other falsy values).

sources/api/schnauzer/src/v2/import/helpers.rs
webern: There used to be a comment to help us remember which way we are supposed to name them. I know because I wrote it because Tom told me to! Maybe this comment should be restored!

https://github.com/bottlerocket-os/bottlerocket/blob/b530f30850d58ad84a8b0ded2e08fa91d554c7a8/sources/api/schnauzer/src/lib.rs#L121

webern (approved):

sources/api/schnauzer/src/v2/import/helpers.rs
sam-berning: I need to fix the commit title to be snake case too so I'll add this back

webern (approved):

Issue #3883, opened 2024-04-09
OOTB: package builds should not be aware of image features


webern:
Some packages in Bottlerocket are built differently based on what image features are found in the variant Cargo.toml. We need packages to be unaware of variant and built the same for all variants.

In concert with https://github.com/bottlerocket-os/twoliter/issues/193

These are the Bottlerocket packages affected:

```
packages/kernel-5.10/Cargo.toml
9:variant-sensitive = "platform"

packages/systemd/Cargo.toml
13:package-features = ["unified-cgroup-hierarchy"]

packages/kernel-5.15/Cargo.toml
9:variant-sensitive = "platform"

packages/os/Cargo.toml
9:variant-sensitive = true
27:package-features = [ "systemd-networkd" ]

packages/kernel-6.1/Cargo.toml
9:variant-sensitive = "platform"
```

webern:
We might have this done.

grep spec files for image-features in packages. If we don't find any we should try building without passing image features to the package builds.

bcressey:
Done in #4008.

webern:
But... we should still delete the deprecated loading of the variant manifest here.

PR #3886, opened 2024-04-10
systemd: use bootconfig for unified cgroup hierarchy

bcressey:
**Issue number:**
Related: #3883

**Description of changes:**
This takes advantage of the new Twoliter support for [bootconfig drop-ins](https://github.com/bottlerocket-os/twoliter/commit/b4f16204c4bcf831fcbc703b6d48439d1f67eed1) and [variant metadata](https://github.com/bottlerocket-os/twoliter/commit/28b64bb180ae51e7fe56f2d4125873b35e39958d).

Variants with the `unified-cgroup-hierarchy` feature flag set will end up installing the new subpackage when the image is built, which will populate the default bootconfig with the right systemd option.

Default to "hybrid" at build time rather than "unified" since the two variants that don't support boot config (`aws-ecs-1` and `aws-ecs-1-nvidia`) also require "hybrid" mode.

**Testing done:**
When building `aws-dev`, the following capability is logged for the metadata rpm:
```
  #13 0.582 bottlerocket-image-feature(unified-cgroup-hierarchy)
```

Later, this triggers the expected installation of the new subpackage:
```
  #15 1.175 Installing dependencies:
...
  #15 1.175  bottlerocket-systemd                          aarch64 252.22-1      repo 4.4 M
  #15 1.175  bottlerocket-systemd-console                  aarch64 252.22-1      repo  49 k
  #15 1.175  bottlerocket-systemd-networkd                 aarch64 252.22-1      repo 1.6 M
  #15 1.175  bottlerocket-systemd-resolved                 aarch64 252.22-1      repo 416 k
  #15 1.175  bottlerocket-systemd-unified-cgroup-hierarchy aarch64 252.22-1      repo 7.2 k
```

After boot, the expected hierarchy and settings are present:
```
bash-5.2# df /sys/fs/cgroup
Filesystem     1K-blocks  Used Available Use% Mounted on
cgroup2                0     0         0    - /sys/fs/cgroup

bash-5.2# apiclient get settings.boot.init
{
  "settings": {
    "boot": {
      "init": {
        "systemd.unified_cgroup_hierarchy": [
          "1"
        ]
      }
    }
  }
}
```

I also verified I could override the setting and the system would come up in hybrid mode.

For a variant like `aws-k8s-1.24`, the hybrid subpackage is installed and the system comes up in hybrid mode as expected.

```
  #15 1.191 Installing dependencies:
...
  #15 1.191  bottlerocket-systemd                         aarch64 252.22-1       repo 4.4 M
  #15 1.191  bottlerocket-systemd-hybrid-cgroup-hierarchy aarch64 252.22-1       repo 6.6 k
```

**Terms of contribution:**

### DIFF
--- a/packages/systemd/systemd.spec
+++ b/packages/systemd/systemd.spec
@@ -14,6 +14,7 @@ Source3: journald.conf
 Source4: issue
 Source5: systemd-journald.conf
 Source6: systemd-sysusers.conf
+Source7: bootconfig-unified-cgroup-hierarchy.conf
 
 # Backport of upstream patches that make the netlink default timeout
 # configurable.  Bottlerocket carries this patch and configures the timeout in
@@ -102,6 +103,10 @@ Requires: %{_cross_os}libselinux
 Requires: %{_cross_os}libuuid
 Requires: %{_cross_os}libxcrypt
 
+# Only require a cgroup hierarchy package when building an image, not
+# for packages that need systemd-devel as a build dependency.
+Requires: (%{name}(cgroup-hierarchy) if %{_cross_os}metadata)
+
 %description
 %{summary}.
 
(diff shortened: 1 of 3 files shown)


packages/systemd/systemd.spec
```diff
@@ -130,6 +131,15 @@ Summary: Files for resolved
 %description resolved
 %{summary}.
 
+%package unified-cgroup-hierarchy
+Summary: Bootconfig snippet for unified cgroup hierarchy
+Requires: (%{_cross_os}image-feature(unified-cgroup-hierarchy) and %{name})
+Supplements: (%{_cross_os}image-feature(unified-cgroup-hierarchy) and %{name})
```
yeazelm: I haven't run across `Supplements` in the wild much, is this a nice to have or do we need this weak backward dependency for this to work?

yeazelm (approved):

packages/systemd/systemd.spec
bcressey: > I haven't run across `Supplements` in the wild much, is this a nice to have or do we need this weak backward dependency for this to work?

The main drawback of weak dependencies - which otherwise have the nice property of affinitizing a package install to certain features - is that the package can be dropped under [certain conditions](https://rpm-software-management.github.io/rpm/manual/dependencies.html):
> On the other hand, if a package that is named foo or that Provides: foo is selected, and a package bar containing “Supplements: foo” exists, then bar is also selected as long as doing so does not lead to unresolvable dependencies.

I can work around this with strong forward dependencies with a bit of effort.

bcressey:
⬆️ force push removes the use of `Supplements` and weak dependencies. It's not quite as elegant but will avoid the risk of subtle failures.

bcressey:
⬆️ force push fixes parens

yeazelm (approved):
Nice!

arnaldo2792 (approved):

Issue #1667, opened 2021-07-23
FIPS Support / Validation


vennemp:
This is related to my comment in #1297 

Will you adding support for FIPS 140-2 standards? Please note: I am NOT asking about FIPS validation with regards to the containers themselves - I understand that is handled, at least technically speaking, during the build phase of the container image.  FIPS 140-2 compliance for other OSes (Amazon Linux 2, RHEL) configures the crypto module to operate in compliance (proper crypto algorithms, entropy, etc) - these would be done in container image; however, there are other changes at the kernel-level that occur as well.  I don't believe full on validation is in scope for Bottlerocket since crypto modules themselves get validated (not necessarily an OS) but perhaps guidance and/or documentation on meeting compliance requirements would be helpful.

zmrow:
Thanks for opening this!  We'll look into it.

technotaff-nbs:
A lack of compliance statement for CIS is preventing our company (and i'm sure many others) from adopting Bottlerocket.

JohnTrevorBurke:
Any updates around FIPS compliance and the usage of Bottlerocket OS? It's a blocker for our team to be able to use for Fedramp compliant EKS nodes. Any updates would be very much appreciated!

vennemp:
Hey @JohnTrevorBurke i believe I just added you in linked in. I may be able to help.

geekmuse:
In the same boat here re: FedRAMP.  Any info/updates here?

FringeCase:
Bumping @geekmuse comment; FedRAMP path

walesb:
Hi Team, I have the same query here, I just want to confirm to know if AWS BottleRocket AMI's are FIPS-140-2 compliant by default or not? 

Looking forward to your reply.

stevehipwell:
Is there any news on this?

bcressey:
FIPS compliance is our second most requested feature, behind CIS (which is in progress), and I'm planning to focus on it once the CIS benchmark is complete.

There are (at least) two interpretations of what this means, and different auditors may have different takes:
1. Bottlerocket's kernel uses compliant algorithms when `fips=1` is passed on the kernel command line, and its userspace tries to arrange for containers to do the same (by mounting in `/etc/system-fips` and such).
2. In addition to the above, Bottlerocket's kernel and userspace are themselves FIPS certified, meaning that they only use validated, FIPS-compliant crypto.

[...]

PR #3887, opened 2024-04-10
packages: build Go binaries for FIPS and non-FIPS

bcressey:
**Issue number:**
Related: #1667

**Description of changes:**
For every Go package that ran afoul of the new [check-fips](https://github.com/bottlerocket-os/bottlerocket-sdk/blob/develop/macros/check-fips) script, I added "bin" and "fips-bin" subpackages with the corresponding binaries. Apart from the packaging metadata, the core change is very simple: build each binary twice, first with `go build` and then with `gofips build`.

For cases where build arguments would lead to excessive duplication or very long lines, I moved them into arrays so the copy/paste wasn't quite as offensive.

I also touched a handful of Go packages that didn't need to be built twice, to drop arguments like `-buildmode=pie` that come from `GOFLAGS`.

**Testing done:**
Ran a variety of variants - `aws-dev`, `aws-k8s-1.24`, `aws-k8s-1.28-nvidia`, `aws-ecs-1`, `aws-ecs-2` - built with and without the FIPS flag set. The nodes were able to join clusters and run pods.

**Terms of contribution:**

### DIFF
--- a/packages/amazon-ssm-agent/amazon-ssm-agent.spec
+++ b/packages/amazon-ssm-agent/amazon-ssm-agent.spec
@@ -1,8 +1,3 @@
-# Don't generate debug packages because we are compiling without CGO,
-# and the `gc` compiler doesn't append the  the ".note.gnu.build-id" section
-# https://fedoraproject.org/wiki/PackagingDrafts/Go#Build_ID
-%global debug_package %{nil}
-
 %global goproject github.com/aws
 %global gorepo amazon-ssm-agent
 %global goimport %{goproject}/%{gorepo}
@@ -17,8 +12,27 @@ Source0: %{gorepo}-%{version}.tar.gz
 Source1000: clarify.toml
 
 BuildRequires: %{_cross_os}glibc-devel
+Requires: %{name}(binaries)
 
 %description
+%{summary}.
+
+%package bin
+Summary: Remote management agent binaries
+Provides: %{name}(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{name})
+Conflicts: (%{_cross_os}image-feature(fips) or %{name}-fips-bin)
+
+%description bin
+%{summary}.
+
+%package fips-bin
+Summary: Remote management agent binaries, FIPS edition
+Provides: %{name}(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{name})
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
+
+%description fips-bin
 %{summary}.
 
 %prep
@@ -30,19 +44,28 @@ BuildRequires: %{_cross_os}glibc-devel
 go build -ldflags "${GOLDFLAGS}" -o amazon-ssm-agent \
   ./core/agent.go ./core/agent_unix.go ./core/agent_parser.go
 
+gofips build -ldflags "${GOLDFLAGS}" -o fips/amazon-ssm-agent \
+  ./core/agent.go ./core/agent_unix.go ./core/agent_parser.go
+
 go build -ldflags "${GOLDFLAGS}" -o ssm-agent-worker \
+  ./agent/agent.go ./agent/agent_unix.go ./agent/agent_parser.go
+
+gofips build -ldflags "${GOLDFLAGS}" -o fips/ssm-agent-worker \
   ./agent/agent.go ./agent/agent_unix.go ./agent/agent_parser.go
 
 go build -ldflags "${GOLDFLAGS}" -o ssm-session-worker \
   ./agent/framework/processor/executer/outofproc/sessionworker/main.go
 
+gofips build -ldflags "${GOLDFLAGS}" -o fips/ssm-session-worker \
+  ./agent/framework/processor/executer/outofproc/sessionworker/main.go
+
 %install
 # Install the SSM agent under 'libexecdir', since it is meant to be used by other programs
-install -d %{buildroot}%{_cross_libexecdir}/amazon-ssm-agent/bin/%{version}
+install -d %{buildroot}{%{_cross_libexecdir},%{_cross_fips_libexecdir}}/amazon-ssm-agent/bin/%{version}
 for b in amazon-ssm-agent ssm-agent-worker ssm-session-worker; do
-  install -D -p -m 0755 ${b} %{buildroot}%{_cross_libexecdir}/amazon-ssm-agent/bin/%{version}
+  install -p -m 0755 ${b} %{buildroot}%{_cross_libexecdir}/amazon-ssm-agent/bin/%{version}
+  install -p -m 0755 fips/${b} %{buildroot}%{_cross_fips_libexecdir}/amazon-ssm-agent/bin/%{version}
 done
-
 
 %cross_scan_attribution --clarify %{S:1000} go-vendor vendor
 
@@ -50,5 +73,15 @@ done
 %license LICENSE
 %{_cross_attribution_file}
 %{_cross_attribution_vendor_dir}
+
+%files bin
 %dir %{_cross_libexecdir}/amazon-ssm-agent
-%{_cross_libexecdir}/amazon-ssm-agent
+%{_cross_libexecdir}/amazon-ssm-agent/bin/%{version}/amazon-ssm-agent
+%{_cross_libexecdir}/amazon-ssm-agent/bin/%{version}/ssm-agent-worker
+%{_cross_libexecdir}/amazon-ssm-agent/bin/%{version}/ssm-session-worker
+
+%files fips-bin
+%dir %{_cross_fips_libexecdir}/amazon-ssm-agent
+%{_cross_fips_libexecdir}/amazon-ssm-agent/bin/%{version}/amazon-ssm-agent
+%{_cross_fips_libexecdir}/amazon-ssm-agent/bin/%{version}/ssm-agent-worker
+%{_cross_fips_libexecdir}/amazon-ssm-agent/bin/%{version}/ssm-session-worker
--- a/packages/cni/cni.spec
+++ b/packages/cni/cni.spec
@@ -29,7 +29,7 @@ Requires: %{_cross_os}iptables
 %build
 %set_cross_go_flags
 
-go build -buildmode=pie -ldflags="${GOLDFLAGS}" -o "bin/cnitool" %{goimport}/cnitool
+go build -ldflags="${GOLDFLAGS}" -o "bin/cnitool" %{goimport}/cnitool
 
 %install
 install -d %{buildroot}%{_cross_libexecdir}/cni/bin
--- a/packages/kubernetes-1.23/kubernetes-1.23.spec
+++ b/packages/kubernetes-1.23/kubernetes-1.23.spec
@@ -70,8 +70,27 @@ Requires: %{_cross_os}findutils
 Requires: %{_cross_os}ecr-credential-provider
 Requires: %{_cross_os}aws-signing-helper
 Requires: %{_cross_os}static-pods
+Requires: %{_cross_os}kubelet-1.23(binaries)
 
 %description -n %{_cross_os}kubelet-1.23
+%{summary}.
+
+%package -n %{_cross_os}kubelet-1.23-bin
+Summary: Container cluster node agent binaries
+Provides: %{_cross_os}kubelet-1.23(binaries)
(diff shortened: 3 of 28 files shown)


packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -17,37 +12,73 @@ Source0: %{gorepo}-%{version}.tar.gz
 Source1000: clarify.toml
 
 BuildRequires: %{_cross_os}glibc-devel
+Requires: %{name}(binaries)
```
bcressey, line 15: Each top-level package now depends on a capability like this to ensure that its binaries are installed - since the package wouldn't be very useful without those!

packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -17,37 +12,73 @@ Source0: %{gorepo}-%{version}.tar.gz
...
 
 %description
 %{summary}.
 
+%package bin
+Summary: Remote management agent binaries
+Provides: %{name}(binaries)
+Requires: (%{_cross_os}image-feature(no-fips) and %{name})
+Conflicts: (%{_cross_os}image-feature(fips) or %{name}-fips-bin)
```
bcressey, line 24: Each binary package provides the `%{name}(binaries)` capability that the top-level package needs.

It requires the top-level package (so that `dnf install bottlerocket-runc-bin` also installs `bottlerocket-runc`) as well as the corresponding image feature capability from the metadata package. This will either be "fips" or "no-fips" depending on the variant.

Finally, it conflicts with the opposite image feature or the other binary package. The conflict helps the package manager eliminate the "wrong" package from consideration, and also prevents `bottlerocket-runc-bin` and `bottlerocket-runc-fips-bin` from being installed together.

packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -17,37 +12,73 @@ Source0: %{gorepo}-%{version}.tar.gz
...
+
+%description fips-bin
+%{summary}.
+
 %prep
 %setup -n %{gorepo}-%{version}
 
 %build
-%set_cross_go_flags
+%set_cross_go_flags_static
```
bcressey: This spec needed one of the largest changes which led to this new macro (discussed in https://github.com/bottlerocket-os/bottlerocket-sdk/pull/162).

Basically the old `CGO_ENABLED=0` trick to get a static binary doesn't work for FIPS builds that need AWS-LC, and this macro sets up an environment that leaves CGO enabled and uses our musl-targeted GCC as the external linker, since musl supports a usable form static linking and glibc does not.

packages/kubernetes-1.23/kubernetes-1.23.spec
```diff
@@ -89,17 +108,27 @@ export FORCE_HOST_GO=1
 make generated_files
 
 # Build kubelet with the target toolchain.
+%set_cross_go_flags
+unset CC
```
bcressey, line 113: `%set_cross_go_flags` brings the standard `GOFLAGS` and `GOLDFLAGS` into the environment. Afterwards, we need to unset the `CC` variable or else it'll fail to build the host codegen programs.

The `export %{kube_cc}` line below (unchanged) is what sets the cross-compiler in the way that the Makefile expects.

packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
```diff
@@ -49,8 +49,8 @@ Conflicts: %{name}-ecs
 
 %build
 %cross_go_configure %{goimport}
-go build -buildmode=pie -ldflags="${GOLDFLAGS}" -o nvidia-container-runtime-hook ./cmd/nvidia-container-runtime-hook
-go build -buildmode=pie -ldflags="${GOLDFLAGS}" -o nvidia-ctk ./cmd/nvidia-ctk
+go build -ldflags="${GOLDFLAGS}" -o nvidia-container-runtime-hook ./cmd/nvidia-container-runtime-hook
```
yeazelm, line 52: I'm not finding a great place to ask this, so this line is as good as any. Do we have a general idea on what *should* have FIPS and what shouldn't? I think we are finding them all in this PR but I'm unsure how we document what one needs to consider when the next Go package needs to be added to the repo. Is it just "always do these things for all Go packages?" If so, I think I'm fine with that but it might be nice to consider a way for us to avoid having so much repetitive logic for non-FIPS and FIPS builds and to update our guidance for packaging Go-based packages so we don't collectively forget what thought went into this PR later down the line.

yeazelm (approved):
This looks right to me. The one thing that has me scratching my head is that I noticed the very repetitive and predictable changes made to all the spec files. It is really noticeable in a PR such as this one where we have basically touched all go builds at once. I haven't thought of a simpler way to accomplish this so I'll approve :) But I do wonder if we could code gen or template some of these things more deeply to make edits to something like this easier to do across all the packages at once.

bcressey:
> But I do wonder if we could code gen or template some of these things more deeply to make edits to something like this easier to do across all the packages at once.

I considered having a macro along the lines of `%debug_package` since the metadata is essentially mechanical, as you say.

I didn't like the idea much because it felt too magical or potentially confusing. It works for debug info because that's a transparent background process that usually doesn't intrude on the actual process of building and installing the software - you don't have to refer to the paths in `%install`. The packages are defined, the file lists are generated automatically, and the `%files` section for the package just refers to the list.

For the FIPS / non-FIPS distinction, we could automate the definitions and the file section, but we'd be left with somewhat magical incantations like:
```
%global fips_package 1

Name: blah

%build
go build ....
gofips build ....

%install
install -d %{buildroot}%{_cross_bindir}
install -p -m 0755 go-bin %{buildroot}%{_cross_bindir}

install -d %{buildroot}%{_cross_fips_bindir}
install -p -m 0755 go-fips-bin %{buildroot}%{_cross_fips_bindir}

%files
<everything except the binaries>
```

It just felt too weird to populate the buildroot with files - especially the key binaries, like `runc` for the runc package - and then not list them in `%files`. Worse, it's actually a packaging error to list them there: they'd be included twice, and potentially both versions would end up on the system at runtime.

And if we don't automate `%files` for the subpackages, then we end up with sections referring to subpackages that aren't visible in the spec file.

In my mind, that only left a macro to generate the subpackage definitions, like `%generate_fips_and_non_fips_subpackages`. I suppose that could be OK but I tend to prefer the explicit definitions. They are annoying to add the first time but after that it's just something I skip past unless I'm trying to understand or adjust requirements, at which point it's nice to just have them in plain sight.

packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
bcressey: > I'm not finding a great place to ask this, so this line is as good as any. Do we have a general idea on what _should_ have FIPS and what shouldn't?

The [check-fips](https://github.com/bottlerocket-os/bottlerocket-sdk/blob/develop/macros/check-fips) script handles this pretty well. It becomes a build error to include a Go binary that imports `crypto/tls` without also building a FIPS-enabled version of that binary. At that point the packager has to address it. Some packages, like `oci-add-hooks`, are fine and don't need any special treatment.

I didn't want to gate the use of a new SDK on a PR like this, so the check is currently off by default. Once this merges I can flip that around in the next SDK release.

larvacea: The lede cites [check-fips](https://github.com/bottlerocket-os/bottlerocket-sdk/blob/develop/macros/check-fips) as the motivation. Is the criterion "if the binary uses crypto/tls, it needs a FIPS-equivalent using goboringcrypto"?

In the check-fips script, the test applied to the FIPS binary is:

```bash
      if ! uses_go_boring_crypto "${f}"; then
```

Is this sufficient, or should the script test:

```bash
      if uses_go_crypto_tls "${f}" then
        echo "${f} is built with crypto/tls" >&2
        (( miscompiled+=1 ))
      elif ! uses_go_boring_crypto "${f}"; then
        echo "${f} is not built with FIPS crypto (but ${b} is built with crypto/tls)" >&2
        (( miscompiled+=1 ))
      fi
```

I may be overcomplicating this, but I have encountered the "we linked library good-fred, but we also linked library fred-evil-twin, and the TLA customers DO NOT WANT" problem, so I worry.

yeazelm:
>  I suppose that could be OK but I tend to prefer the explicit definitions.

I agree with your all your comments but this one rings true the most to me. The verbosity is annoying at first but it does seem like the spec files have just the right amount of context vs opaque magic that might be hard to understand. I went through several of these general approaches in my head when writing my comment but failed to enumerate them so well so thanks for doing that, it really helps solidify that we arrived at the ideal approach!

packages/nvidia-container-toolkit/nvidia-container-toolkit.spec
yeazelm: > I didn't want to gate the use of a new SDK on a PR like this, so the check is currently off by default. Once this merges I can flip that around in the next SDK release.

This is great, I think this is what I was hoping to see to make sure we don't have drift from the good state after this PR as we make changes to the Go packages.

bcressey: > I may be overcomplicating this, but I have encountered the "we linked library good-fred, but we also linked library fred-evil-twin, and the TLA customers DO NOT WANT" problem, so I worry.

The Go upstream uses (such as [boring.go](https://github.com/golang/go/blob/master/src/crypto/tls/boring.go) and [notboring.go](https://github.com/golang/go/blob/master/src/crypto/tls/notboring.go)) are gated behind conditionals like:
```
//go:build boringcrypto
//go:build !boringcrypto
```

That is pretty fundamental to how `GOEXPERIMENT=boringcrypto` is implemented so I am not very worried about including both - the build would fail because of multiple definitions for the same function, for example.

packages/cni/cni.spec
```diff
@@ -29,7 +29,7 @@ Requires: %{_cross_os}iptables
 %build
 %set_cross_go_flags
 
-go build -buildmode=pie -ldflags="${GOLDFLAGS}" -o "bin/cnitool" %{goimport}/cnitool
+go build -ldflags="${GOLDFLAGS}" -o "bin/cnitool" %{goimport}/cnitool
```
arnaldo2792, line 32: No fips version for this package?

arnaldo2792 (approved):

packages/cni/cni.spec
bcressey: > No fips version for this package?

It doesn't import `crypto/tls`, so it doesn't need it. `oci-add-hooks` and `ecs-gpu-init` were the same way.

bcressey:
⬆️ force push rebases on develop, fixing conflicts and dropping changes that were merged in separate commits (7bbd73fae5570dd93a243da787da214f8b2cbf83, 30aeecd4e3c644c6c7f8a8097306bbdf539e965e, 35c2432a4cbdb076313c4b5ac54f8311754b1959).

yeazelm (approved):

packages/amazon-ssm-agent/amazon-ssm-agent.spec
```diff
@@ -17,10 +12,29 @@ Source0: %{gorepo}-%{version}.tar.gz
...
+Conflicts: (%{_cross_os}image-feature(fips) or %{name}-fips-bin)
+
+%description bin
+%{summary}.
+
+%package fips-bin
+Summary: Remote management agent binaries, FIPS edition
+Provides: %{name}(binaries)
+Requires: (%{_cross_os}image-feature(fips) and %{name})
+Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
```
webern, line 33: What does this part prevent?

```
or %{name}-bin
```

webern (approved):

packages/amazon-ssm-agent/amazon-ssm-agent.spec
bcressey: > What does this part prevent?
> 
> 
> 
> ```
> 
> or %{name}-bin
> 
> ```

It's meant to prevent the bin and fips-bin packages from both being installed, since only one set of binaries would end up in use.

It's a bit redundant but hypothetically nothing stops someone from adding `Provides: %{_cross_os}image-feature(fips)` to a random package in their variant that doesn't have that feature set. So even if that happened, this would force only one of the two packages to be installed.

PR #3892, opened 2024-04-11
pessimize pluto and cfsignal builds

bcressey:
**Issue number:**
N/A

**Description of changes:**
On a high-end build machine, `pluto` takes 12 minutes to compile when using one codegen unit, and around 3 minutes to compile with the default of 16 codegen units. This is because the aws-sdk-ec2 crate contains a massive amount of code.

Sacrifice a little code quality and spec file clarity to bring build times for this package back under control by setting up a new target directory and another background job for `pluto`. `cfsignal` is given the same treatment, even though aws-sdk-cloudformation is not as much of a problem, because they share crate dependencies that aren't used by other crates.

Refactor the handling of stdout and stderr file descriptors to ensure that the `cargo build` invocation is shown when printing the results, instead of when the job is first started. Otherwise, it overlaps with the non-static, non-AWS-SDK output in a confusing way.

**Testing done:**
Built `vmware-dev`, `aws-dev`, `aws-k8s-1.28` to confirm that the expected binaries were built.

Launched the `aws-k8s-1.28` node which joined the cluster after `pluto` generated the usual settings.

Binary size before this change:
```
bash-5.1# ls -latr /usr/bin/pluto /usr/bin/cfsignal
-rwxr-xr-x. 1 root root  8903560 Apr  5 22:52 /usr/bin/cfsignal
-rwxr-xr-x. 1 root root 12931096 Apr  5 22:52 /usr/bin/pluto
```

Binary size after this change:
```
bash-5.1# ls -latr /usr/bin/pluto /usr/bin/cfsignal
-rwxr-xr-x. 1 root root 11617128 Apr 11 22:48 /usr/bin/cfsignal
-rwxr-xr-x. 1 root root 17133856 Apr 11 22:48 /usr/bin/pluto
```

Definitely worse, but acceptable if this improves build times as expected.

**Terms of contribution:**

### DIFF
--- a/packages/os/os.spec
+++ b/packages/os/os.spec
@@ -315,15 +315,47 @@ migrations=()
 for migration in $(find %{_builddir}/sources/api/migration/migrations/v[0-9]* -mindepth 1 -maxdepth 1 -type d); do
     migrations+=("-p $(basename ${migration})")
 done
+
+# Since RPM automatically logs the commands that run, and since we want to display those commands
+# along with the output from the background job, we do some file descriptor juggling below.
+#  exec 3>&1 4>&2           - save stdout and stderr to fds 3 and 4
+#  exec 1>"${output}" 2>&1  - redirect stdout and stderr to job-specific log
+#  exec 1>&3 2>&4           - restore stdout and stderr from fds 3 and 4
+
 # Store the output so we can print it after waiting for the backgrounded job.
+exec 3>&1 4>&2
 static_output="$(mktemp)"
+exec 1>"${static_output}" 2>&1
 # Build static binaries in the background.
 %cargo_build_static --manifest-path %{_builddir}/sources/Cargo.toml \
     -p apiclient \
     ${migrations[*]} \
-    >> ${static_output} 2>&1 &
+    &
 # Save the PID so we can wait for it later.
 static_pid="$!"
+exec 1>&3 2>&4
+
+%if %{with aws_platform} || %{with aws_k8s_family}
+# The AWS SDK crates are extremely slow to build with only one codegen unit.
+# Pessimize the release build for just the crates that depend on them.
+# Store the output so we can print it after waiting for the backgrounded job.
+exec 3>&1 4>&2
+aws_sdk_output="$(mktemp)"
+exec 1>"${aws_sdk_output}" 2>&1
+RUSTFLAGS="$(echo "%{__global_rustflags_shared}" | sed -e 's,-Ccodegen-units=1,,g')" \
+%{__cargo_cross_env} \
+CARGO_TARGET_DIR="${HOME}/.cache/.aws-sdk" \
+%{__cargo} build \
+    %{__cargo_cross_opts} \
+    --release \
+    --manifest-path %{_builddir}/sources/Cargo.toml \
+    %{?with_aws_platform: -p cfsignal} \
+    %{?with_aws_k8s_family: -p pluto} \
+    &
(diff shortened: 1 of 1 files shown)


jpculp (approved):

packages/os/os.spec
```diff
@@ -315,15 +315,41 @@ migrations=()
...
 %cargo_build_static --manifest-path %{_builddir}/sources/Cargo.toml \
     -p apiclient \
     ${migrations[*]} \
-    >> ${static_output} 2>&1 &
+    &
 # Save the PID so we can wait for it later.
 static_pid="$!"
+exec 1>&3 2>&4
+
+%if %{with aws_platform} || %{with aws_k8s_family}
```
arnaldo2792, line 338: Is this one of those few exceptions where it is OK to make the spec file platform or family-aware?

packages/os/os.spec
```diff
@@ -315,15 +315,41 @@ migrations=()
...
+# The AWS SDK crates are extremely slow to build with only one codegen unit.
+# Pessimize the release build for just the crates that depend on them.
+# Store the output so we can print it after waiting for the backgrounded job.
+exec 3>&1 4>&2
+aws_sdk_output="$(mktemp)"
+exec 1>"${aws_sdk_output}" 2>&1
+RUSTFLAGS="$(echo "%{__global_rustflags_shared}" | sed -e 's,-Ccodegen-units=1,,g')" \
```
arnaldo2792, line 345: Just for my own understanding, what's the penalty of building all the first-party crates with more than one code unit?

arnaldo2792 (approved):

packages/os/os.spec
bcressey: > Just for my own understanding, what's the penalty of building all the first-party crates with more than one code unit?

I expect it'd be in the realm of 10-20% binary size increase and an unknown performance hit. It'd push us pretty close to the current filesystem size limit for existing variants.

packages/os/os.spec
bcressey: > Is this one of those few exceptions where it is OK to make the spec file platform or family-aware?

This is just to keep the existing conditional compile logic in effect.

When these crates don't need to be conditionally compiled any more, we can just drop the conditions and always do the AWS SDK consumer crates in a separate pass (like we do with static builds).

packages/os/os.spec
```diff
@@ -315,15 +315,41 @@ migrations=()
 for migration in $(find %{_builddir}/sources/api/migration/migrations/v[0-9]* -mindepth 1 -maxdepth 1 -type d); do
     migrations+=("-p $(basename ${migration})")
 done
+
 # Store the output so we can print it after waiting for the backgrounded job.
+exec 3>&1 4>&2
```
larvacea, line 326: nit: a comment explaining the purpose (line 330) of fd 3 and fd 4 here would have helped me out. I experienced seconds and seconds of bemusement, tbh ("fd 3: what? Non-standard-wrong-error with bonus Hadamard matrix?"). Live and learn. This may just be me.

larvacea (approved):
Looks good to me.

packages/os/os.spec
bcressey: > nit: a comment explaining the purpose (line 330) of fd 3 and fd 4 here would have helped me out.

Added a comment before the file descriptor juggling starts to explain what's going on.

larvacea (approved):
Thanks!

PR #3898, opened 2024-04-15
dogtag: implement imds and reverse dns hostname tools

### LINKED ISSUE #3871: Dogtag Hostname tool
This is needed in support of removing conditional compilation from netdog

Supports/precedes #3618
Related to #3700

jmt-lab:
**Issue number:** 3871

Closes #3871 

**Description of changes:**

Implements two hostname discovery helpers imds and reverse dns. These small programs will be used by netdog to discover a node's hostname. This PR only covers the small implementation of the dogtag helpers. Once this is merged I can adjust the netdog PR to utilize them.

**Testing done:**

Built both helpers currently and tested on instances to validate they have returned the appropriate and correct hostnames.

**Terms of contribution:**

### DIFF
--- a/sources/dogtag/Cargo.toml
+++ b/sources/dogtag/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "dogtag"
+version = "0.1.0"
+authors = ["Jarrett Tierney <jmt@amazon.com>"]
+license = "Apache-2.0 OR MIT"
+edition = "2021"
+publish = false
+exclude = ["README.md"]
+
+[[bin]]
+name = "20-imds"
+path = "bin/imds.rs"
+
+[[bin]]
+name = "10-reverse-dns"
+path = "bin/reverse.rs"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+argh = "0.1"
+dns-lookup = "2"
+imdsclient = { version = "0.1", path = "../imdsclient" }
+log = "0.4"
+snafu = "0.8"
+tokio = { version = "~1.32", features = ["macros"]}  # LTS
+walkdir = "2"
+
+[build-dependencies]
+generate-readme = { version = "0.1", path = "../generate-readme" }
--- a/sources/dogtag/README.md
+++ b/sources/dogtag/README.md
@@ -1,0 +1,16 @@
+# dogtag
+
+Current version: 0.1.0
+
+dogtag resolves the hostname of a bottlerocket server/instance. It's used to generate settings.network.hostname. To accomplish this, it uses a set of standalone binaries in /var/bottlerocket/dogtag that resolve the hostname via different methods.
+
+Currently, bottlerocket ships with two hostname resolver binaries:
+
+20-imds - Fetches hostname from EC2 Instance Metadata Service
+10-reverse-dns - Uses reverse DNS lookup to resolve the hostname
+
+dogtag runs the resolvers in /var/bottlerocket/dogtag in reverse alphanumerical order until one of them returns a hostname, at which point it will exit early and print the returned hostname to stdout.
+
+## Colophon
+
+This text was generated from `README.tpl` using [cargo-readme](https://crates.io/crates/cargo-readme), and includes the rustdoc from `src/main.rs`.
--- a/sources/dogtag/bin/imds.rs
+++ b/sources/dogtag/bin/imds.rs
@@ -1,0 +1,39 @@
+use dogtag::Cli;
+use imdsclient::ImdsClient;
+use snafu::{OptionExt, ResultExt};
+
+type Result<T> = std::result::Result<T, error::Error>;
+
+/// Implements a hostname lookup tool by fetching the public hostname
+/// from the instance metadata via IMDS. It will interface with IMDS
+/// via:
+///
+/// * Check for IPv6, default to IPv4 if not available
+/// * Check for IMDSv2, fallback to IMDSv1 if not enabled
+#[tokio::main]
+async fn main() -> Result<()> {
+    // Even though for this helper we do not need any arguments
+    // still validate to ensure the helper follows standards.
+    let _: Cli = argh::from_env();
+    let mut imds = ImdsClient::new();
+    let hostname = imds
+        .fetch_hostname()
+        .await
+        .context(error::ImdsSnafu)?
+        .context(error::NoHostnameSnafu)?;
+    println!("{}", hostname);
+    Ok(())
+}
+
+mod error {
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub(super) enum Error {
+        #[snafu(display("failed to fetch hostname from IMDS: {}", source))]
+        Imds { source: imdsclient::Error },
+        #[snafu(display("no hostname returned by imds"))]
+        NoHostname,
+    }
+}
--- a/sources/dogtag/bin/reverse.rs
+++ b/sources/dogtag/bin/reverse.rs
@@ -1,0 +1,34 @@
+use dns_lookup::lookup_addr;
+use dogtag::Cli;
+use snafu::ResultExt;
+
+type Result<T> = std::result::Result<T, error::Error>;
+
+/// Looks up the public hostname by using dns-lookup to
+/// resolve it from the ip address provided
+fn main() -> Result<()> {
+    let cli: Cli = argh::from_env();
+    let ip: std::net::IpAddr = cli.ip_address.parse().context(error::InvalidIpSnafu)?;
+    let hostname = lookup_addr(&ip).context(error::LookupSnafu)?;
+    println!("{}", hostname);
+    Ok(())
+}
+
+mod error {
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub(super) enum Error {
+        #[snafu(display("Invalid ip address passed to tool {}", source))]
+        InvalidIp {
+            #[snafu(source(from(std::net::AddrParseError, Box::new)))]
+            source: Box<std::net::AddrParseError>,
+        },
+        #[snafu(display("Failed to lookup hostname via dns {}", source))]
+        Lookup {
+            #[snafu(source(from(std::io::Error, Box::new)))]
+            source: Box<std::io::Error>,
+        },
+    }
+}
--- a/sources/dogtag/src/lib.rs
+++ b/sources/dogtag/src/lib.rs
@@ -1,0 +1,84 @@
+/*!
+dogtag resolves the hostname of a bottlerocket server/instance. It's used to generate settings.network.hostname. To accomplish this, it uses a set of standalone binaries in /var/bottlerocket/dogtag that resolve the hostname via different methods.
+
+Currently, bottlerocket ships with two hostname resolver binaries:
+
+20-imds - Fetches hostname from EC2 Instance Metadata Service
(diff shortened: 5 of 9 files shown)


jmt-lab:
Removed unused mockito

sources/dogtag/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+
+* 01-imds - Fetches hostname from the Instance Metadata via IMDS
+* 00-reverse-dns - Uses reverse dns lookup to resolve the hostname
+ */
+use argh::FromArgs;
+use log::debug;
+use snafu::ResultExt;
+use std::{fs, process};
+
+const DOGTAG_BIN_PATH: &str = "/var/bottlerocket/dogtag";
```
sam-berning, line 17: for the similar functionality in `early-boot-config`, we use `/usr/libexec/early-boot-config/data-providers.d/`, not sure if it would be more coherent for these lookup binaries to live under `/usr/libexec/dogtag.d/`?

sources/dogtag/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+    );
+    let walker = fs::read_dir(DOGTAG_BIN_PATH).context(error::IoSnafu)?;
+    let mut hostname_helpers = Vec::new();
+    for entry in walker.flatten() {
+        let path = entry.path();
+        if path.is_file() {
+            hostname_helpers.push(path.clone());
+        }
+    }
+    // We want to do reverse sort as we want to prioritize higher numbers first
```
sam-berning, line 36: should we allow for a usecase where an OOTBuilder would want to add a dogtag resolver at a lower priority than the default ones? Like perhaps put the two defaults on `10-` and `20-`? with a 2-digit prefix to the binaries, there should be plenty of wiggle room -- hard to imagine a distro with >99 ways to resolve a hostname.

sources/dogtag/bin/reverse.rs
```diff
@@ -0,0 +1,43 @@
+/*!
```
sam-berning: I don't think we need the full `dogtag` docs on each resolver, maybe just some info on the resolver itself?

sources/dogtag/bin/imds.rs
```diff
@@ -0,0 +1,44 @@
...
+/// * Check for IPv6, default to IPv4 if not available
+/// * Check for IMDSv2, fallback to IMDSv1 if not enabled
+#[tokio::main]
+async fn main() -> Result<()> {
+    // Even though for this helper we do not need any arguments
+    // still validate to ensure the helper follows standards.
+    let _: Cli = argh::from_env();
+    let mut imds = ImdsClient::new();
+    let hostname = imds.fetch_hostname().await.context(error::ImdsSnafu)?;
+    println!("{}", hostname.context(error::NoHostnameSnafu)?);
```
sam-berning: nit: I think I'd prefer we handle all the errors right away, e.g.:

```suggestion
    let hostname = imds.fetch_hostname().await
            .context(error::ImdsSnafu)?
            .context(error::NoHostnameSnafu)?;
    println!("{}", hostname);
```

sources/dogtag/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+* 00-reverse-dns - Uses reverse dns lookup to resolve the hostname
+ */
+use argh::FromArgs;
+use log::debug;
+use snafu::ResultExt;
+use std::{fs, process};
+
+const DOGTAG_BIN_PATH: &str = "/var/bottlerocket/dogtag";
+
+/// CLi defines the standard cmdline interface for all hostname handlers
```
sam-berning:
```suggestion
/// Cli defines the standard cmdline interface for all hostname handlers
```

sources/dogtag/src/lib.rs
jmt-lab: I think that makes more sense, will adjust this

sources/dogtag/src/lib.rs
jmt-lab: Yeah i think thats a good idea, i'll bump them to 10- and 20-

jmt-lab:
Addressed PR comments

sources/dogtag/src/lib.rs
sam-berning: I'm realizing that there is actually a bit of a distinction here -- `early-boot-config` merges all of the results of its data providers into one, much more in keeping with what you'd expect from a `.d` directory, whereas after any of these hostname resolvers return a result, we return it. I'm ok to stick with the current name

sources/dogtag/src/lib.rs
```diff
@@ -0,0 +1,79 @@
+/*!
+dogtag is a set of tools that detect the hostname of a bottlerocket server/instance and prints it to stdout.
+if the tool is called in an environment it cannot resolve the hostname it will error out.
+
+Currently the following hostname tools are implemented:
+
+* 01-imds - Fetches hostname from the Instance Metadata via IMDS
+* 00-reverse-dns - Uses reverse dns lookup to resolve the hostname
```
sam-berning: names need updating

sources/dogtag/src/lib.rs
```diff
@@ -0,0 +1,79 @@
+/*!
+dogtag is a set of tools that detect the hostname of a bottlerocket server/instance and prints it to stdout.
```
sam-berning: this is an extreme nit but I think it'd be clearer to describe `dogtag` as just a "hostname resolver" rather than a "set of tools". I'd also like to see an explanation of how it uses the resolvers here as well. Something like this:


> `dogtag` resolves the hostname of a bottlerocket server/instance. It's used to generate `settings.network.hostname`. To accomplish this, it uses a set of standalone binaries in `/var/bottlerocket/dogtag` that resolve the hostname via different methods.
>
> Currently, bottlerocket ships with two hostname resolver binaries:
>
> * `20-imds` - Fetches hostname from EC2 Instance Metadata Service
> * `10-reverse-dns` - Uses reverse DNS lookup to resolve the hostname
>
> `dogtag` runs the resolvers in `/var/bottlerocket/dogtag` in reverse alphanumerical order until one of them returns a hostname, at which point it will exit early and print the returned hostname to `stdout`.

sources/dogtag/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+
+pub type Result<T> = std::result::Result<T, error::Error>;
+
+/// find_hostname will utilize the helpers located in /var/bottlerocket/dogtag/ to try and discover the hostname
+pub async fn find_hostname() -> Result<String> {
+    debug!(
+        "attempting to discover hostname helpers in {}",
+        DOGTAG_BIN_PATH
+    );
+    let walker = fs::read_dir(DOGTAG_BIN_PATH).context(error::IoSnafu)?;
```
sam-berning: you may be able to use `walkdir` in order to gather and sort the binaries all in one step, e.g. https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/early-boot-config/early-boot-config/src/main.rs#L96-L106

jmt-lab:
Updated to use walkdir and fixed documentation

jmt-lab:
Fix embarassing build issue

sources/dogtag/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+argh = "0.1"
+dns-lookup = "2"
+imdsclient = { version = "0.1", path = "../imdsclient" }
+log = "0.4"
+snafu = "0.8"
+tokio = { version = "1.32", features = ["full"]}
+walkdir = "2.4"
```
jpculp:
```suggestion
tokio = { version = "~1.32", features = ["full"]}  # LTS
walkdir = "2"
```

Also, do you need the full feature set here?

jmt-lab: No but we use full everywhere else

jpculp: That's not the case for crates under `/sources`.

sam-berning (approved):

jpculp (approved):

sources/dogtag/README.md
```diff
@@ -0,0 +1,16 @@
+# dogtag
+
+Current version: 0.1.0
+
+dogtag resolves the hostname of a bottlerocket server/instance. It's used to generate settings.network.hostname. To accomplish this, it uses a set of standalone binaries in /var/bottlerocket/dogtag that resolve the hostname via different methods.
+
+Currently, bottlerocket ships with two hostname resolver binaries:
```
bcressey, line 7: We should not ship the IMDS resolver on non-aws variants.

sources/dogtag/bin/imds.rs
```diff
@@ -0,0 +1,39 @@
...
+use snafu::{OptionExt, ResultExt};
+
+type Result<T> = std::result::Result<T, error::Error>;
+
+/// Implements a hostname lookup tool by fetching the public hostname
+/// from the instance metadata via IMDS. It will interface with IMDS
+/// via:
+///
+/// * Check for IPv6, default to IPv4 if not available
+/// * Check for IMDSv2, fallback to IMDSv1 if not enabled
```
bcressey, line 12: I don't think these comments apply; `imdsclient` always uses IMDSv2 and the IPv4 address (which is still valid on IPV6-only ENIs).

sources/dogtag/bin/imds.rs
```diff
@@ -0,0 +1,39 @@
...
+
+mod error {
+    use snafu::Snafu;
+
+    #[derive(Debug, Snafu)]
+    #[snafu(visibility(pub(super)))]
+    pub(super) enum Error {
+        #[snafu(display("failed to fetch hostname from IMDS: {}", source))]
+        Imds { source: imdsclient::Error },
+        #[snafu(display("no hostname returned by imds"))]
```
bcressey, line 36: nit: "IMDS"

sources/dogtag/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+            .unwrap_or(None);
+        if let Some(output) = output.as_ref() {
+            // Read the std output
+            if output.status.success() {
+                let hostname = String::from_utf8_lossy(output.stdout.as_slice()).to_string();
+                return Ok(hostname.trim().to_string());
+            }
+        }
+    }
+    Err(error::Error::NoHelper {})
```
bcressey, line 64: The original logic in ab3b3e6373637448d169968e6e5760f9df66a411 had a fallback where the host's IP address was used in a slightly cleaned up way (dots replaced with dashes) in these two cases:
1. `lookup_addr()` returns the IP address rather than a name
2. no host is found at all

That logic was later removed in 0de2d11f8bb88055a4928f9c46a6bdad4b70ee01 since in some cases (EKS-A clusters) we want the hostname to exactly match the IP.

0b9658facc9734e0eea279e0b97ee5923cef0988 fixed a different bug related to direct use of an IPv6 address, where colons needed to be replaced with dashes.

Is that handled in `netdog`? It might be cleaner to do it here for separation of concerns, so netdog only has to put in an IP address and it gets back a hostname (from IMDS, reverse DNS, IPv4 address, or sanitized IPv6 address).

sources/dogtag/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+        .into_iter()
+        .map(|x| x.into_path())
+        .collect();
+    hostname_helpers.reverse();
+
+    for helper in hostname_helpers.iter() {
+        let output = process::Command::new(helper)
+            .output()
+            .map(Some)
+            .unwrap_or(None);
```
bcressey, line 55: I don't see where we are passing the `ip_address` argument, which at least the reverse DNS helper needs or else it will immediately error out.

Issue #3459, opened 2023-09-14
No LustreFS support in newer variants (k8s-1.28, ecs-2)


yeazelm:
The 6.1 kernel currently doesn't support [LustreFS](https://aws.amazon.com/fsx/lustre/) but the 5.10 and 5.15 kernels do. There seems to be some work ongoing upstream to bring this support back into the 6.1 kernel. We should add this support once it is ready for the 6.1 kernels.

bcressey:
This affects all the new variants using 6.1:
* `aws-ecs-2` / `aws-ecs-2-nvidia`
* `aws-k8s-1.28` / `aws-k8s-1.28-nvidia`
* `vmware-k8s-1.28`
* `metal-k8s-1.28`

Other variants are not affected.

autarchprinceps:
What options does one have if one needs continued FSx support? Not upgrade to EKS 1.28? Because you can't downgrade, and you won't know this is an issue, until you upgrade. Just ran into this issue in one of our clusters

stmcginnis:
That's a good question @autarchprinceps. Assuming you are not relying on any Kubernetes 1.28 functionality, you could deploy new worker nodes using a Bottlerocket 1.27 AMI. That is a [supported Kubernetes configuration](https://kubernetes.io/releases/version-skew-policy/) that would allow you to still use Lustre, even after the EKS cluster has been upgraded to 1.28.

kamirendawkins:
Another workaround that we have chosen, to avoid version skew, is to leverage Karpenter to force pods that require FSX onto AL2(Kernel 5.10) until this is resolved and we can move back to BottleRocket for those services.

bcressey:
This *should* be fixed by the kernel 6.1 update in #3853, but needs to be verified. (cc: @larvacea)

larvacea:
A (very) quick test pre-release showed that 6.1 had the LustreFS configuration and patch from upstream, and that the pieces appeared to be able to speak to a Lustre file system. If you feel confident in upstream testing, this may be sufficient. If not, let me know and I can add Lustre testing to our loop.

bcressey:
I'd like to see positive proof that the Lustre CSI driver works on Bottlerocket with the 6.1 kernel before we resolve this and add a release note about it.

natehudson:
[...]

PR #3905, opened 2024-04-22
remove variant sensitivity from 5.15 kernel

bcressey:
**Issue number:**
Related: #3883

**Description of changes:**
Move non-aws variants from the 5.15 to the 6.1 kernel. After the only in-tree users of the 5.15 kernel are aws variants, merge the aws-specific config into the main config and remove the variant sensitive marker from the package.

Although changing the kernel for a variant mid-stream is unusual, the [update policy](https://github.com/bottlerocket-os/bottlerocket/blob/develop/SECURITY_FEATURES.md#update-policy) permits it for cases where the kernel will reach end-of-life before the variant.

Removing conditional compilation from the 6.1 kernel will be quite a bit more complicated. Landing this change first will limit that complexity to just the 6.1 kernel, rather than requiring a separate set of complicated changes for the 5.15 kernel shortly before it reaches end-of-life.

(A future change will move all the `aws-k8s-*` variants from the 5.15 to the 6.1 kernel. Currently that's blocked by #3459 and full confirmation that Lustre works as expected.)

**Testing done:**
```
❯ ./tools/diff-kernel-config -b HEAD^ -a HEAD -v aws-k8s-1.24 -o configs
...

config-aarch64-aws-k8s-1.24-diff:         0 removed,   0 added,   0 changed
config-x86_64-aws-k8s-1.24-diff:          0 removed,   0 added,   0 changed
```

**Terms of contribution:**


larvacea (approved):
Looks good to me. Thanks!

yeazelm (approved):

PR #3918, opened 2024-04-26
Allow downstream users to use a different containerd templated configuration

vyaghras:
**Issue number:**

Closes #

**Description of changes:**
 This will allow downstream users to provide their own containerd templated configuration, so that they can provide more configurations: 
- The API system uses the templated configuration provided by the users.
- The existing variants use the existing templated containerd configurations using links. 
- Migration to use new template on upgrade and old template on downgrade.

**Testing done:**
- Migration from old version to new version k8s
  - [x] /etc/containerd/config.toml remains same.
  - [x] ls -ltr usr/share/templates conatins following files before migration
  ```
  ls -ltr usr/share/templates
  -rw-r--r--. 1 root root 2564 Mar 18 21:23 containerd-config-toml_k8s_nvidia_containerd_sock
  -rw-r--r--. 1 root root 2563 Mar 18 21:23 containerd-config-toml_k8s_containerd_sock
  -rw-r--r--. 1 root root  386 Mar 18 21:23 containerd-config-toml_basic
  ```
  And following files after migration
  ```
  -rw-r--r--. 1 root root 2563 Mar 18 21:23 containerd-config-toml_k8s_containerd_sock
  lrwxrwxrwx. 1 root root   42 Apr 25 22:49 containerd-config -> containerd-config-toml_k8s_containerd_sock
  ```
  - [x] Previously running pods keeps running successfully.
  - [x] Downgrade to old version bring the system in the same state as earlier.
- Migration from old Bottlerocket version for k8s Nvidia
  - [x] /etc/containerd/config.toml remains same.
  - [x] ls -ltr usr/share/templates conatins following files before migration
  ```
  ls -ltr usr/share/templates
  -rw-r--r--. 1 root root 2564 Mar 18 21:23 containerd-config-toml_k8s_nvidia_containerd_sock
  -rw-r--r--. 1 root root 2563 Mar 18 21:23 containerd-config-toml_k8s_containerd_sock
  -rw-r--r--. 1 root root  386 Mar 18 21:23 containerd-config-toml_basic
  ```
  And following files after migration
  ```
  ls -ltr usr/share/templates/
  -rw-r--r--. 1 root root 2564 Mar 18 21:23 containerd-config-toml_k8s_nvidia_containerd_sock
   lrwxrwxrwx. 1 root root   49 Apr 26 16:09 containerd-config -> containerd-config-toml_k8s_nvidia_containerd_sock
  ```
  - [x] Previously running pods keeps running successfully.
  - [x] Downgrade to old version bring the system in the same state as earlier.
- Migration from old Bottlerocket version for ECS
- Migration from old Bottlerocket version for Vmware
- Migration from old Bottlerocket version for Metal 

**Terms of contribution:**

### DIFF
--- a/packages/containerd/containerd.spec
+++ b/packages/containerd/containerd.spec
@@ -55,6 +55,26 @@ Requires: (%{_cross_os}image-feature(fips) and %{name})
 Conflicts: (%{_cross_os}image-feature(no-fips) or %{name}-bin)
 
 %description fips-bin
+
+%package basic
+Summary: Basic containerd configuration
+Requires: %{name}
+
+%description basic
+%{summary}.
+
+%package cri
+Summary: Containerd configuration for CRI
+Requires: %{name}
+
+%description cri
+%{summary}.
+
+%package cri-nvidia
+Summary: Containerd configuration for CRI-nvidia
+Requires: %{name}
+
+%description cri-nvidia
 %{summary}.
 
 %prep
@@ -109,6 +129,15 @@ install -p -m 0644 %{S:5} %{buildroot}%{_cross_tmpfilesdir}/containerd.conf
 
 %cross_scan_attribution --clarify %{S:1000} go-vendor vendor
 
+%post basic -p <lua>
+posix.symlink("containerd-config-toml_basic", "%{_cross_templatedir}/containerd-config")
+
+%post cri -p <lua>
+posix.symlink("containerd-config-toml_k8s_containerd_sock", "%{_cross_templatedir}/containerd-config")
+
+%post cri-nvidia -p <lua>
+posix.symlink("containerd-config-toml_k8s_nvidia_containerd_sock", "%{_cross_templatedir}/containerd-config")
+
 %files
 %license LICENSE NOTICE
 %{_cross_attribution_file}
@@ -117,7 +146,6 @@ install -p -m 0644 %{S:5} %{buildroot}%{_cross_tmpfilesdir}/containerd.conf
 %{_cross_unitdir}/etc-containerd.mount
 %{_cross_unitdir}/prepare-var-lib-containerd.service
 %dir %{_cross_factorydir}%{_cross_sysconfdir}/containerd
-%{_cross_templatedir}/containerd-config-toml*
 %{_cross_templatedir}/containerd-cri-base-json
 %{_cross_tmpfilesdir}/containerd.conf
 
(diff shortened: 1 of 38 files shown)


packages/containerd/containerd.spec
```diff
@@ -38,6 +38,36 @@ Requires: %{_cross_os}pigz
...
+Requires: %{name}
+
+%description basic
+%{summary}.
+
+%post basic -p <lua>
+posix.symlink("containerd-config-toml_basic", "%{_cross_templatedir}/containerd-config")
+
+%package cri
+Summary: Containerd configuration for Kubernetes
```
arnaldo2792:
```suggestion
Summary: Containerd configuration for CRI
```

packages/containerd/containerd.spec
```diff
@@ -38,6 +38,36 @@ Requires: %{_cross_os}pigz
...
 %{summary}.
 
+%package basic
+Summary: Basic containerd configuration
+Requires: %{name}
+
+%description basic
+%{summary}.
+
+%post basic -p <lua>
```
arnaldo2792: Lets move the `post` install scripts after `%install`

packages/containerd/containerd.spec
```diff
@@ -38,6 +38,36 @@ Requires: %{_cross_os}pigz
...
+Requires: %{name}
+
+%description cri
+%{summary}.
+
+%post cri -p <lua>
+posix.symlink("containerd-config-toml_k8s_containerd_sock", "%{_cross_templatedir}/containerd-config")
+
+%package cri-nvidia
+Summary: Containerd configuration for Kubernetes-nvidia
```
arnaldo2792: Same note as elsewhere `s/Kubernetes/CRI`

sources/api/migration/migrations/v1.20.0/update-containerd-config-template-path/src/main.rs
```diff
@@ -0,0 +1,37 @@
+use migration_helpers::common_migrations::{NoOpMigration, ReplaceStringMigration};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+///
```
arnaldo2792: Empty comment here. We usually add a description about why this is needed.

sources/api/migration/migrations/v1.20.0/update-containerd-config-template-path/src/main.rs
```diff
@@ -0,0 +1,37 @@
...
+fn run() -> Result<()> {
+    if cfg!(variant_runtime = "ecs") {
```
arnaldo2792, line 7: Could you try assigning the migration to a variable, and run that at the end? e.g.

```rust
let migration = if cfg!(variant_runtime = "ecs") {
    ReplaceStringMigration {...}
} else if cfg!(variant_runtime = "k8s") {
     ReplaceStringMigration {...}
} else {
    NoOpMigration
};

migrate(migration)?
```

vyaghras: Resolved offline.

vyaghras:
We are closing this for now and will explore alternative approaches in the future.

PR #3927, opened 2024-04-29
bump package versions for static-pods and early-boot-config

jmt-lab:
**Issue number:** 3926

**Description of changes:**

* Bump rpm version of static-pods to address build failures
* Revert static-pods build dependency change
* Bump rpm version of early-boot-config

**Terms of contribution:**

### DIFF
--- a/packages/early-boot-config/early-boot-config.spec
+++ b/packages/early-boot-config/early-boot-config.spec
@@ -2,7 +2,7 @@
 %undefine _debugsource_packages
 
 Name: %{_cross_os}early-boot-config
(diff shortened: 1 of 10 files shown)


sumukhballal (approved):

bcressey:
Can you also revert https://github.com/bottlerocket-os/bottlerocket/commit/9a6b352189941c9c9eb695d81f459680c752d147 since it didn't help? It has a side effect of slowing down the build, because the kubelet package can't start until the static-pods package is finished.

bcressey (approved):

PR #3929, opened 2024-04-30
Add instance type matching to shibaken.

larvacea:
We would like to query and match the instance type for the benefit of systemd unit files. Returning success/failure is useful for conditional execution in systemd unit files.

**Description of changes:**

Adds a subcommand to `shibaken`:

```
shibaken instance-type --prefix m5.x --prefix some-thing
```

Exit code 0 (success or true) if the IMDS instance type starts with one (or more) of the `--prefix` parameters. Otherwise exit code 1 (failure or false). Failure is also returned if we do not get a non-empty instance type, or if the command has no `--prefix` options (no instance is a member of the empty set).

This iterates through the supplied prefixes looking for a match. This is reasonable for a few dozen prefixes, and I expect we will have at most a half dozen. Should we ever have a requirement for thousands of prefixes, we should revisit that search-by-iteration design choice.

**Testing done:**

Manual testing on an EC2 instance running aws-dev:

- `shibaken` returns 0 (true) if the instance type matches at least one `--prefix`.
- `shibaken` returns 1 (false) if the instance type matches no `--prefix`.
- `shibaken` returns 1 (false) if there are no `--prefix` parameters.
- `shibaken` works as expected with 0, 1, or 2 `--prefix` parameters.

Not tested:

- behavior when the IMDS fetcher returns an error.
- behavior when the IMDS fetcher returns an empty string for instance type.

**Terms of contribution:**

### DIFF
--- a/sources/api/shibaken/src/instance.rs
+++ b/sources/api/shibaken/src/instance.rs
@@ -1,0 +1,49 @@
+use std::process::ExitCode;
+
+use argh::FromArgs;
+use imdsclient::ImdsClient;
+use snafu::ResultExt;
+
+use crate::error::{self, Result};
+
+#[derive(FromArgs, Debug)]
+#[argh(subcommand, name = "instance-type")]
+/// Fetch instance type from IMDS
+pub(crate) struct InstanceType {
+    #[argh(option)]
+    /// prefix string(s) to match against instance type
+    prefix: Vec<String>,
+}
+
+impl InstanceType {
+    pub(crate) async fn run(self) -> Result<ExitCode> {
+        if self.prefix.is_empty() {
+            // Is it success or failure to match against the empty set?
+            // Pretend this is the element-of operator. Nothing is an
+            // element of the empty set.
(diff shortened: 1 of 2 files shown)


jpculp:
This feels a bit weird to me since `shibaken` is a settings generator.

larvacea:
shibaken is apparently already multipurpose. It has, for instance, a subcommand to match partition against a list of strings, and a please-wait-for-warm-start subcommand. That's not a great argument for adding yet another thing to shibaken's arsenal of apparently unrelated abilities, so if the consensus is "make this its own thing, don't make shibaken even more cluttered," I will cheerfully go do that.

jpculp:
Hm. I was under the impression warmpool worked differently, but now that I look at it it does break the settings generator precedent. I do think it makes sense to consolidate the IMDS things to `shibaken`, but I guess I'm just not crazy about the exit code approach as is.

larvacea:
Based on feedback--will not alter shibaken to do this.

PR #3932, opened 2024-05-01
remove variant sensitivity from 6.1 kernel

bcressey:
**Issue number:**
Related: #3883 

**Description of changes:**
Merge all the platform-specific kernel config snippets back together, and remove variant sensitivity from the `kernel-6.1` package.

By far the largest (and least interesting) part of this change is the switch to explicit listing of all kernel modules. My goal here is to turn some of the kernel config changes - enabling a new module, making a module built-in, or removing a module - into code changes that can be reviewed in the usual way. My hope is that it can drive some cleanup over time, if not in this kernel then in kernels still to come: adding a `modules-aws` and `modules-vmware` package, removing unnecessary variation between x86_64 and aarch64, and so forth. `modules-metal` is a first step in that direction.

The bootconfig snippets are a potentially contentious addition. They don't reduce the kernel's increased size or surface area, though they do make it a no-op by default. They won't take effect for in-place upgrades, only for newly provisioned nodes. However, I do see value in making explicit which functionality is not needed on which platforms, to assist future refactoring.

As a free bonus change, switch to zstd for firmware compression to remove another use of xz, and to avoid adding it to `aws` and `vmware` variants.

**Testing done:**
No removals, only additions and changes. For the most part, changes are storage drivers (+ dependencies) moving from modules to built-in, and network drivers (+ dependencies) moving from off to modules.

```
❯ ./tools/diff-kernel-config -b develop -a HEAD -v aws-dev -v vmware-dev -v metal-dev -o configs
...
config-aarch64-aws-dev-diff:      0 removed,  90 added,  29 changed
config-aarch64-metal-dev-diff:    0 removed,  15 added,   3 changed
config-x86_64-aws-dev-diff:       0 removed,  98 added,  31 changed
config-x86_64-metal-dev-diff:     0 removed,   1 added,   4 changed
config-x86_64-vmware-dev-diff:    0 removed,  92 added,  38 changed
```

`diff-report` and `diff-summary` are available [here](https://gist.github.com/bcressey/61b9162a14c982e4aaf79dc113444c25).

**Terms of contribution:**


larvacea (approved):
Are the storage driver configs y because it would be challenging to boot without a driver for the boot device, and the network drivers are m because it's fine to load them as needed (using those storage drivers) when systemd is bringing up networking?

yeazelm (approved):
This is a lot of change but after thinking about it and some offline conversation, I am on board with this change.

bcressey:
> Are the storage driver configs y because it would be challenging to boot without a driver for the boot device, and the network drivers are m because it's fine to load them as needed (using those storage drivers) when systemd is bringing up networking?

Yes, exactly. Storage drivers need to be "y" because the combination of [dm-mod.create](https://github.com/bottlerocket-os/twoliter/blob/a8554c3316b652f673ad9f373b8846af0e0ef9f2/twoliter/embedded/rpm2img#L586) and `root=/dev/dm-0`  to mount the root filesystem requires any necessary storage drivers to be built into the kernel, as there is no initramfs or userspace available to load them from modules. Network drivers aren't necessary until later in the boot so they can be modules.

PR #3940, opened 2024-05-03
build the pause image from upstream source and use as the default

tzneal:
**Description of changes:**

Builds the pause image from the Kubernetes source into the image.

**Testing done:**

Ran an instance with no pod-infra-container-image specified, and with one specified:

```
bash-5.1# journalctl --facility=1
May 06 18:59:42 localhost root[5061]: The setting settings.kubernetes.pod-infra-container-image is deprecated and has no effect. It will be removed in future versions of Bottlerocket.
May 06 19:05:51 i-00e97c60c2abba780.us-west-2.compute.internal root[34230]: The setting settings.kubernetes.pod-infra-container-image is deprecated and has no effect. It will be removed in future versions of Bottlerocket.
```

```
bash-5.1# journalctl  -u kubelet | head
May 06 18:59:43 i-00e97c60c2abba780.us-west-2.compute.internal systemd[1]: Starting Kubelet...
May 06 18:59:43 i-00e97c60c2abba780.us-west-2.compute.internal ctr[5426]: unpacking localhost/kubernetes/pause:0.1.0 (sha256:dc510f415dfb132b9359fc451408733862c89b0e10e3ffbdeb6e36be50148a54)...done
May 06 18:59:43 i-00e97c60c2abba780.us-west-2.compute.internal ctr[5478]: io.cri-containerd.pinned=pinned,io.cri-containerd.image=managed
May 06 18:59:44 i-00e97c60c2abba780.us-west-2.compute.internal kubelet[5492]: Flag --container-runtime-endpoint has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
May 06 18:59:44 i-00e97c60c2abba780.us-west-2.compute.internal kubelet[5492]: Flag --containerd has been deprecated, This is a cadvisor flag that was mistakenly registered with the Kubelet. Due to legacy concerns, it will follow the standard CLI deprecation timeline before being removed.
May 06 18:59:44 i-00e97c60c2abba780.us-west-2.compute.internal kubelet[5492]: Flag --register-with-taints has been deprecated, This parameter should be set via the config file specified by the Kubelet's --config flag. See https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/ for more information.
May 06 18:59:44 i-00e97c60c2abba780.us-west-2.compute.internal kubelet[5492]: Flag --pod-infra-container-image has been deprecated, will be removed in a future release. Image garbage collector will get sandbox image information from CRI.
May 06 18:59:44 i-00e97c60c2abba780.us-west-2.compute.internal kubelet[5492]: I0506 18:59:44.811726    5492 server.go:204] "--pod-infra-container-image will not be pruned by the image garbage collector in kubelet and should also be set in the remote runtime"
May 06 18:59:44 i-00e97c60c2abba780.us-west-2.compute.internal kubelet[5492]: I0506 18:59:44.821784    5492 server.go:487] "Kubelet version" kubeletVersion="v1.29.1-eks-61c0bbb"
May 06 18:59:44 i-00e97c60c2abba780.us-west-2.compute.internal kubelet[5492]: I0506 18:59:44.821820    5492 server.go:489] "Golang settings" GOGC="" GOMAXPROCS="" GOTRACEBACK="
```

**Terms of contribution:**


tzneal:
Putting up to get some thoughts on the approach before I replicate this to the other K8s versions.

packages/kubernetes-1.29/kubernetes-1.29.spec
```diff
@@ -122,6 +131,23 @@ export KUBE_OUTPUT_SUBPATH="_fips_output/local"
 export GOEXPERIMENT="boringcrypto"
 make WHAT="cmd/kubelet"
 
+# build the pause container
+cd build/pause/linux/
+
+# Build static pause executable for container image.
+mkdir -p rootfs/usr/bin
+%{_cross_triple}-musl-gcc ${_cross_cflags} -static pause.c -o rootfs/pause
```
bcressey: To apply all the hardening flags:
```suggestion
%{_cross_triple}-musl-gcc %{_cross_cflags} %{_cross_ldflags} -static-pie pause.c -o rootfs/pause
```

packages/kubernetes-1.29/kubernetes-1.29.spec
```diff
@@ -122,6 +131,23 @@ export KUBE_OUTPUT_SUBPATH="_fips_output/local"
...
+
+# Construct container image.
+mkdir -p image/rootfs
+%tar_cf image/rootfs/layer.tar -C rootfs .
+DIGEST=$(sha256sum image/rootfs/layer.tar | sed -e 's/ .*//')
+install -m 0644 %{S:100} image/config.json
+sed -i "s/~~digest~~/${DIGEST}/" image/config.json
+install -m 0644 %{S:101} image/manifest.json
+
+%tar_cf ../../../_output/local/bin/linux/%{_cross_go_arch}/amazon-eks-pause.tar -C image .
```
bcressey: To make it more vendor-neutral:
```suggestion
%tar_cf ../../../_output/local/bin/linux/%{_cross_go_arch}/kubernetes-pause.tar -C image .
```

packages/kubernetes-1.29/kubernetes-1.29.spec
```diff
@@ -190,6 +219,7 @@ install -p -m 0644 %{S:15} %{buildroot}%{_cross_datadir}/logdog.d
 %dir %{_cross_libexecdir}/kubernetes
 %{_cross_libexecdir}/kubernetes/kubelet-plugins
 %{_cross_datadir}/logdog.d/logdog.kubelet.conf
+%{_cross_datadir}/eks/amazon-eks-pause.tar
```
bcressey: since it's an arch-specific binary rather than human-readable:
```suggestion
%{_cross_libexecdir}/kubernetes/kubernetes-pause.tar
```

packages/kubernetes-1.29/pause-config.json
```diff
@@ -0,0 +1 @@
+{"author":"Amazon Web Services, Inc.","config":{"Cmd":["/pause"],"ArgsEscaped":true},"created":"2014-12-12T01:12:53.332832423Z","history":[{"created":"2014-12-12T01:12:53.332832423Z","author":"Amazon Web Services, Inc.","created_by":"Bottlerocket","empty_layer":true}],"os":"linux","rootfs":{"type":"layers","diff_ids":["sha256:~~digest~~"]}}
```
bcressey:
```suggestion
{"author":"Bottlerocket","config":{"Cmd":["/pause"],"ArgsEscaped":true},"created":"2014-12-12T01:12:53.332832423Z","history":[{"created":"2014-12-12T01:12:53.332832423Z","author":"Bottlerocket","created_by":"Bottlerocket","empty_layer":true}],"os":"linux","rootfs":{"type":"layers","diff_ids":["sha256:~~digest~~"]}}
```

packages/kubernetes-1.29/pause-manifest.json
```diff
@@ -0,0 +1 @@
+[{"Config":"config.json","RepoTags":["localhost/amazon/pause-bottlerocket:0.1.0"],"Layers":["rootfs/layer.tar"]}]
```
bcressey:
```suggestion
[{"Config":"config.json","RepoTags":["localhost/kubernetes/pause:0.1.0"],"Layers":["rootfs/layer.tar"]}]
```

packages/kubernetes-1.29/prestart-load-pause-ctr.conf
```diff
@@ -1,4 +1,18 @@
 [Service]
+# load the built-in pause image
+ExecStartPre=/usr/bin/ctr \
+    --namespace=k8s.io \
+    image import \
+    --all-platforms \
+    /usr/share/eks/amazon-eks-pause.tar
```
bcressey:
```suggestion
    /usr/libexec/kubernetes/kubernetes-pause.tar
```

packages/kubernetes-1.29/prestart-load-pause-ctr.conf
```diff
@@ -1,4 +1,18 @@
...
+
+# label it to prevent it from being removed
+ExecStartPre=/usr/bin/ctr \
+    --namespace=k8s.io \
+    image label \
+    localhost/amazon/pause-bottlerocket:0.1.0 \
```
bcressey:
```suggestion
    localhost/kubernetes:0.1.0 \
```

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -4,6 +4,7 @@ standalone-mode = false
 authentication-mode = "aws"
 server-tls-bootstrap = true
 cloud-provider = "aws"
+pod-infra-container-image = "localhost/amazon/pause-bottlerocket:0.1.0"
```
bcressey: We can just drop this in addition to the settings generator going away.

And then either:
* set `POD_INFRA_CONTAINER_IMAGE` in the `kubelet-env` template to the hardcoded value
* or else set the hardcoded value in `kubelet-exec-start-conf`

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -12,10 +13,6 @@ node-ip.setting-generator = "pluto node-ip"
 provider-id.setting-generator = "pluto provider-id"
 affected-services = ["kubernetes"]
 
-[metadata.settings.kubernetes.pod-infra-container-image]
-setting-generator = "schnauzer-v2 render --requires 'aws@v1' --requires 'kubernetes@v1(helpers=[pause-prefix])' --template '{{ pause-prefix settings.aws.region }}/eks/pause:3.1-eksbuild.1'"
-affected-services = ["kubernetes", "containerd"]
-
```
bcressey: e3bc71dad7734070f62c7541626edc12c76e5c0d is an example of where we've deprecated a setting and turned it into a no-op. Metadata here could be:
```
[metadata.settings.kubernetes.pod-infra-container-image]
affected-services = ["pod-infra-container-image"]

[services.pod-infra-container-image]
configuration-files = ["pod-infra-container-image-log-message"]
restart-commands = ["systemctl restart deprecation-warning@pod-infra-container-image.timer"]

[configuration-files.pod-infra-container-image-log-message]
path = "/etc/deprecated-settings/pod-infra-container-image"
template-path = "/usr/share/templates/pod-infra-container-image"
```

Then you should just need to add a `packages/kubernetes-*/pod-infra-container-image` template:
```
[required-extensions]
kubernetes = { version = "v1", optional = true }
+++
{{~#if settings.kubernetes.pod-infra-container-image~}}
DEPRECATED_SETTING=settings.kubernetes.pod-infra-container-image
{{~/if~}}
```

And in the spec:
```
SourceXYZ: pod-infra-container-image

%install
...
install -p -m 0644 %{S:XYZ} %{buildroot}%{_cross_templatedir}/pod-infra-container-image

%files -n %{_cross_os}kubelet-1.XY
...
%{_cross_templatedir}/pod-infra-container-image
```

This will leverage the common `deprecated-warning@service` and `deprecated-warning@timer` implementation in `release`.

sources/models/shared-defaults/kubernetes-aws.toml
bcressey: Oh, we'd also need to update these two files with the hardcoded value:
```
packages/containerd/containerd-config-toml_k8s_containerd_sock
packages/containerd/containerd-config-toml_k8s_nvidia_containerd_sock
```

tzneal: I thought about just always using the built in one, but thought there might be some weird use case where someone runs a specific pod infra container image.  I'm not sure what that would be, WDYT?

bcressey: > I thought about just always using the built in one, but thought there might be some weird use case where someone runs a specific pod infra container image. I'm not sure what that would be, WDYT?

I haven't heard of a use case like that. Only the obvious one, where someone might want to override the image so it's pulled from a different registry in a restricted or air-gapped environment. Building it solves that problem nicely!

tzneal:
I left the `pause-prefix` code in for now as its referenced in migrations:

```
sources/api/migration/migrations/v1.16.0/schnauzer-v2-generators/src/main.rs
58:                new_val: "schnauzer-v2 render --requires 'aws@v1' --requires 'kubernetes@v1(helpers=[pause-prefix])' --template '{{ pause-prefix settings.aws.region }}/eks/pause:3.1-eksbuild.1'",

sources/api/migration/migrations/archived/v1.1.0/schnauzer-paws/src/main.rs
8:    "{{ pause-prefix settings.aws.region }}/eks/pause-{{ goarch os.arch }}:3.1";
```

But maybe its good to be removed from there as well?

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -13,8 +13,15 @@ provider-id.setting-generator = "pluto provider-id"
 affected-services = ["kubernetes"]
 
 [metadata.settings.kubernetes.pod-infra-container-image]
-setting-generator = "schnauzer-v2 render --requires 'aws@v1' --requires 'kubernetes@v1(helpers=[pause-prefix])' --template '{{ pause-prefix settings.aws.region }}/eks/pause:3.1-eksbuild.1'"
-affected-services = ["kubernetes", "containerd"]
+affected-services = ["pod-infra-container-image"]
+
```
bcressey, line 13: The following applies to all migrations - 

https://github.com/bottlerocket-os/bottlerocket/commit/b46c719afa4998e77fc6617d011f06fd30cc2251 is a fairly clean example of adding a migration, which requires edits in several places:
* `Release.toml` (add these to a new 1.20.0 -> 1.21.0 section
* `sources/Cargo.toml`
* new crates under `sources/api/migration/migrations/v1.21.0` (copy an existing migration crate then edit its Cargo.toml and main.rs)

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -13,8 +13,15 @@ provider-id.setting-generator = "pluto provider-id"
 affected-services = ["kubernetes"]
 
 [metadata.settings.kubernetes.pod-infra-container-image]
-setting-generator = "schnauzer-v2 render --requires 'aws@v1' --requires 'kubernetes@v1(helpers=[pause-prefix])' --template '{{ pause-prefix settings.aws.region }}/eks/pause:3.1-eksbuild.1'"
-affected-services = ["kubernetes", "containerd"]
+affected-services = ["pod-infra-container-image"]
```
bcressey, line 12: Here we're changing a metadata list, `affected-services` - going from the two previous entries to one new one.

The closest migration is [this one](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/migration/migrations/v1.20.0/thar-be-updates-affected-services-v0-1-0/src/main.rs).

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -13,8 +13,15 @@ provider-id.setting-generator = "pluto provider-id"
 affected-services = ["kubernetes"]
 
 [metadata.settings.kubernetes.pod-infra-container-image]
-setting-generator = "schnauzer-v2 render --requires 'aws@v1' --requires 'kubernetes@v1(helpers=[pause-prefix])' --template '{{ pause-prefix settings.aws.region }}/eks/pause:3.1-eksbuild.1'"
```
bcressey, line 12: Here we're removing a metadata element, which is unusual. @jmt-lab can you please split out your new migration helper from #3828 and post a separate pull request for it?

sources/models/shared-defaults/kubernetes-aws.toml
```diff
@@ -13,8 +13,15 @@ provider-id.setting-generator = "pluto provider-id"
...
-affected-services = ["kubernetes", "containerd"]
+affected-services = ["pod-infra-container-image"]
+
+[services.pod-infra-container-image]
+configuration-files = ["pod-infra-container-image-log-message"]
+restart-commands = ["systemctl restart deprecation-warning@pod-infra-container-image.timer"]
+
+[configuration-files.pod-infra-container-image-log-message]
+path = "/etc/deprecated-settings/pod-infra-container-image"
+template-path = "/usr/share/templates/pod-infra-container-image"
```
bcressey, line 20: Here we're creating new "prefixes" in the datastore - which roughly correspond to maps with some number of (possibly nested) key / value pairs, or (equivalently) directories with files (keys), file contents (values), and subdirectories (nested maps).

These end up as directories in the data store:
```
/var/lib/bottlerocket/datastore/current/live/services/pod-infra-container-image/...
/var/lib/bottlerocket/datastore/current/live/configuration-files/pod-infra-container-image-log-message/...
```

Those directories will be auto-populated by a program that uses this file on new launches and upgrades, but need to be removed on downgrade.

`AddPrefixesMigration` (as shown [here](https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/api/migration/migrations/v1.15.0/deprecate-log4j-hotpatch-enabled/src/main.rs)) is the migration helper for this. It's roughly like an `rm -rf` on those directories.

```
    migrate(AddPrefixesMigration(vec![
        "services.pod-infra-container-image",
        "configuration-files.pod-infra-container-image-log-message",
    ]))
```

bcressey (commented):
All the code and config changes look good to me. The remaining piece is the migrations, which you're welcome to tackle or punt to one of the developers more familiar with this dark corner of Bottlerocket.

bcressey:
> I left the pause-prefix code in for now as its referenced in migrations:

We're overdue for a migrations archival - relocating some of the newer migrations to `sources/migration/migrations/archived` where they'll no longer be built. After that we can clean up this part of `schnauzer`. Agreed that it's probably better as a backlog issue than worrying about it here.

sources/models/shared-defaults/kubernetes-aws.toml
jmt-lab: Opened #3958

sources/api/migration/migrations/v1.21.0/build-kubernetes-pause-container/src/main.rs
```diff
@@ -0,0 +1,35 @@
+use migration_helpers::common_migrations::{
+    AddPrefixesMigration, MetadataListReplacement, RemoveMetadataMigration,
+    ReplaceMetadataListsMigration, SettingMetadata,
+};
+use migration_helpers::{migrate, Result};
+use std::process;
+
+fn run() -> Result<()> {
+    migrate(RemoveMetadataMigration(&[SettingMetadata {
```
tzneal: I couldn't find any other migrations with multiple independent migrate calls.  Presumably we get into a bad state if the first succeeds and second fails.

bcressey: > Presumably we get into a bad state if the first succeeds and second fails.

It is worse than that. 😀 

Each migration starts from the snapshot of the datastore taken before the program is run, so the second and subsequent migrations within the same binary will silently discard the results of the previous one.

The fix is essentially to split this into three separate migrations:
```
# RemoveMetadataMigration
pod-infra-container-image-remove-settings-generator

# ReplaceMetadataListMigration
pod-infra-container-image-affected-services

# AddPrefixesMigration
pod-infra-container-image-services
```

tzneal: SGTM, thanks, and fixed!

bcressey (approved):
LGTM. Might be good to tweak the commit titles a bit:
```
settings: use built-in pause container image for k8s
kubernetes-*: switch to built-in pause container image
```

(The k8s commits can be kept separate or squashed into one, up to you.)

bcressey (approved):

arnaldo2792 (approved):

larvacea (approved):
lgtm. I am happy to get a second (or so) back from pulling this image from ECR.

bcressey (approved):

PR #3944, opened 2024-05-04
remove the libbzip2 package

bcressey:
**Issue number:**
N/A

**Description of changes:**
In the wake of the xz event, I'd like to continue removing older, rarely updated compression libraries. bzip2 is the obvious next candidate since it's only depended on by two packages.

In the case of `makedumpfile` it's not actually even required; it's just specified in the default link flags under the assumption that `libelf` and `libdw` will depend on that library.

Removing the `libsemanage` dependency is a bit more involved. The resulting patch is not upstreamable in this form, but suffices for Bottlerocket with its tiny SELinux policy files. It shouldn't be difficult to maintain given that `compressed_file.c` hasn't changed in the past two years.

**Testing done:**
Verified that `semodule` works with the uncompressed files:
```
bash-5.2# semodule -B
bash-5.2# dmesg|tail
[   12.595054] SELinux:  Converting 29 SID table entries...
[   12.598671] SELinux:  policy capability network_peer_controls=1
[   12.598675] SELinux:  policy capability open_perms=1
[   12.598677] SELinux:  policy capability extended_socket_class=1
[   12.598678] SELinux:  policy capability always_check_network=0
[   12.598679] SELinux:  policy capability cgroup_seclabel=1
[   12.598680] SELinux:  policy capability nnp_nosuid_transition=1
[   12.598680] SELinux:  policy capability genfs_seclabel_symlinks=1
[   12.598681] SELinux:  policy capability ioctl_skip_cloexec=0
[   12.652009] audit: type=1403 audit(1714852856.659:3): auid=4294967295 ses=4294967295 lsm=selinux res=1
```

Verified that `makedumpfile` works:
```
bash-5.2# echo c >/proc/sysrq-trigger
[   28.567910] sysrq: Trigger a crash
[   28.568274] Kernel panic - not syncing: sysrq triggered crash
...

bash-5.2# tail -n20 /var/log/kdump/dmesg.log 
[   28.570155] Call trace:
[   28.570392]  dump_backtrace+0xe4/0x140
[   28.570759]  show_stack+0x20/0x30
[   28.571082]  dump_stack_lvl+0x68/0x84
[   28.571438]  dump_stack+0x18/0x34
[   28.571763]  panic+0x12c/0x314
[   28.572064]  sysrq_reset_seq_param_set+0x0/0xa0
[   28.572503]  __handle_sysrq+0x90/0x1a0
[   28.572869]  write_sysrq_trigger+0xe4/0x1b0
[   28.573272]  proc_reg_write+0xac/0x100
[   28.573636]  vfs_write+0xdc/0x360
[   28.573962]  ksys_write+0x74/0x110
[   28.574288]  __arm64_sys_write+0x24/0x30
[   28.574665]  invoke_syscall+0x50/0x120
[   28.575030]  el0_svc_common.constprop.0+0x54/0x190
[   28.575488]  do_el0_svc+0x38/0x50
[   28.575810]  el0_svc+0x34/0x110
[   28.576114]  el0t_64_sync_handler+0xbc/0x140
[   28.584885] Starting crashdump kernel...
[   28.585268] Bye!
```

**Terms of contribution:**


yeazelm (approved):

packages/selinux-policy/selinux-policy.spec
```diff
@@ -64,7 +64,7 @@ install -d "${moddir}"
 for m in *.cil ; do
   mod="${m%.*}"
   install -d "${moddir}/${mod}"
-  bzip2 -c "${m}" > "${moddir}/${mod}/cil"
+  install -p -m 0644 "${m}" "${moddir}/${mod}/cil"
```
arnaldo2792, line 67: Will we consider using ZSTD?

packages/libsemanage/0001-remove-bzip2-dependency.patch
```diff
@@ -0,0 +1,268 @@
...
+ $(LIBPC): $(LIBPC).in ../VERSION
+diff --git a/src/compressed_file.c b/src/compressed_file.c
+index 5546b83..3632d79 100644
+--- a/src/compressed_file.c
++++ b/src/compressed_file.c
+@@ -27,148 +27,14 @@
+ #include <unistd.h>
+ #include <fcntl.h>
+ 
+-#include <bzlib.h>
```
arnaldo2792, line 35: I wonder if upstream will be interested in a patch that includes conditionals for compressed modules. Are you aware of any discussion about other compression libraries?

packages/libbzip2/Cargo.toml
```diff
@@ -1,19 +0,0 @@
-[package]
-name = "libbzip2"
```
arnaldo2792, line 2: Lovely! One less package to update!

arnaldo2792 (approved):

packages/selinux-policy/selinux-policy.spec
bcressey: > Will we consider using ZSTD?

I'd like to, but probably only if it ends up as an upstream feature. The benefits from compression here are so modest that I don't think it's worth carrying an even larger patch to implement that.

packages/libsemanage/0001-remove-bzip2-dependency.patch
bcressey: I haven't seen any recent discussion; in [past threads](https://lore.kernel.org/selinux/?q=libsemanage+compression), compression tended to be challenged with "is this really necessary?" which strikes me as totally correct and fair.

If we really needed compression, we could work around this in Bottlerocket with a combination of squashfs/erofs (as we do with licenses and kernel devel files) and overlayfs (to add a writable upper layer). But that seems like overkill here.

PR #3945, opened 2024-05-05
settings: Add container-runtime settings to AWS K8s NVIDIA variants

yeazelm:
**Description of changes:**
The settings were originally added for 1.10.1 but nvidia flavors were excluded. This adds those in for those variants and creates migrations for them.

**Testing done:**
Tested by building 1.19.5 and doing upgrade testing for both `aws-k8s-1.28` and `aws-k8s-1.28-nvidia`.
Confirmed for `aws-k8s-1.28`
[x] Settings apply fine on 1.19.5
[x] Settings are untouched on upgrade to 1.20.0
[x] Settings are untouched on downgrade to 1.19.5 
Confirmed for `aws-k8s-1.28-nvidia`
[x] Settings don't work on 1.19.5
[x] Upgrade to 1.20.0 allows settings the settings
[x] Downgrade removes the settings
[x] Confirmed the setting makes it into `/etc/containerd/config.yaml`

After running `apiclient set settings.container-runtime.max-concurrent-downloads=4` on upgraded node:
```
[root@admin]# sheltie
bash-5.1# cat /etc/containerd/
config.toml    cri-base.json  nvidia.env
bash-5.1# cat /etc/containerd/config.toml
version = 2
root = "/var/lib/containerd"
state = "/run/containerd"
disabled_plugins = [
    "io.containerd.internal.v1.opt",
    "io.containerd.snapshotter.v1.aufs",
    "io.containerd.snapshotter.v1.devmapper",
    "io.containerd.snapshotter.v1.native",
    "io.containerd.snapshotter.v1.zfs",
]

[grpc]
address = "/run/containerd/containerd.sock"

[plugins."io.containerd.grpc.v1.cri"]
enable_selinux = true
# Pause container image is specified here, shares the same image as kubelet's pod-infra-container-image
sandbox_image = "XXXXXXX.dkr.ecr.us-west-2.amazonaws.com/eks/pause:3.1-eksbuild.1"
max_concurrent_downloads = 4

[plugins."io.containerd.grpc.v1.cri".containerd]
default_runtime_name = "nvidia"

[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia]
runtime_type = "io.containerd.runc.v2"
base_runtime_spec = "/etc/containerd/cri-base.json"

[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia.options]
SystemdCgroup = true
BinaryName = "nvidia-oci"

[plugins."io.containerd.grpc.v1.cri".cni]
bin_dir = "/opt/cni/bin"
conf_dir = "/etc/cni/net.d"
```

**Terms of contribution:**

### DIFF
--- a/sources/api/migration/migrations/v1.20.0/container-runtime-metadata-nvidia/src/main.rs
+++ b/sources/api/migration/migrations/v1.20.0/container-runtime-metadata-nvidia/src/main.rs
@@ -1,0 +1,26 @@
+use migration_helpers::common_migrations::{AddMetadataMigration, NoOpMigration, SettingMetadata};
+use migration_helpers::migrate;
+use migration_helpers::Result;
+use std::process;
+
+/// We added a new setting for configuring container runtime (containerd) settings only for NVIDIA k8s variants.
+fn run() -> Result<()> {
+    if cfg!(variant_family = "aws-k8s") && cfg!(variant_flavor = "nvidia") {
+        migrate(AddMetadataMigration(&[SettingMetadata {
+            metadata: &["affected-services"],
+            setting: "settings.container-runtime",
+        }]))
+    } else {
+        migrate(NoOpMigration)
+    }
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/Release.toml
+++ b/Release.toml
@@ -305,4 +305,6 @@ version = "1.20.0"
     "migrate_v1.20.0_add-ntp-default-options-v0-1-0.lz4",
     "migrate_v1.20.0_static-pods-add-prefix-v0-1-0.lz4",
     "migrate_v1.20.0_static-pods-services-cfg-v0-1-0.lz4",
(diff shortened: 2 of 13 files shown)


sources/api/migration/migrations/v1.20.0/container-runtime-metadata-nvidia/src/main.rs
```diff
@@ -0,0 +1,26 @@
...
+fn run() -> Result<()> {
+    if cfg!(variant_family = "aws-k8s") && cfg!(variant_flavor = "nvidia") {
+        migrate(AddMetadataMigration(&[SettingMetadata {
+            metadata: &["affected-services"],
+            setting: "settings.container-runtime",
+        }]))
+    } else {
+        migrate(NoOpMigration)
+    }
+}
```
bcressey, line 16: As a rule of thumb, every change to the TOML files under "models" requires a migration, and vice versa.

I'd expect you need to add "affected-services" metadata to the corresponding toml, or the settings won't work as expected on 1.20.

bcressey: Ah, I see - the "affected-services" metadata is already present in older versions, it just doesn't do anything because the settings are unknown.

bcressey (approved):

yeazelm:
^ pushed a missing variant (`aws-k8s-1.25-nvidia`). I've triple checked we have them all now 😄

arnaldo2792 (approved):

PR #3946, opened 2024-05-06
logdog: switch to use apiclient cli to get settings

### LINKED ISSUE #3784: OOTB: Remove conditional compilation from logdog
**What I'd like:**

- [ ] Check if there is model dependency on api server, if so remove it via transfer to configuration file
- [ ] Check if there is conditional compilation #[cfg(..)] flags and if so remove them

jmt-lab:
**Issue number:** 3784

Closes #3784

**Description of changes:**

* Adjusted logdog to get settings via cli apiclient get
* Changed shape to be managed with serde_json::Value
* Used a normalize, denormalize methodology to flatten the object to maintain stripping of secrets

**Testing done:**

* Tested and launched, validated that all secrets were removed after generating bottlerocket-logs.tar.gz

**Terms of contribution:**

### DIFF
--- a/sources/logdog/src/error.rs
+++ b/sources/logdog/src/error.rs
@@ -4,7 +4,6 @@ use std::io;
 use std::path::PathBuf;
 
 use crate::log_request::REQUESTS_DIR;
-use datastore::{deserialization, serialization};
 use reqwest::Url;
 use snafu::{Backtrace, Snafu};
 
@@ -12,12 +11,14 @@ use snafu::{Backtrace, Snafu};
 #[snafu(visibility(pub(crate)))]
 #[allow(clippy::enum_variant_names)]
 pub(crate) enum Error {
-    #[snafu(display("Error calling Bottlerocket API '{}': {}", uri, source))]
-    ApiClient {
-        #[snafu(source(from(apiclient::Error, Box::new)))]
-        source: Box<apiclient::Error>,
-        uri: String,
+    #[snafu(display("Error calling apiclient: {}", source))]
+    ApiCommandFailure {
+        #[snafu(source(from(std::io::Error, Box::new)))]
+        source: Box<std::io::Error>,
     },
+
+    #[snafu(display("Error fetching settings via apiclient: {}", reason))]
+    ApiExecutionFailure { reason: String },
 
     #[snafu(display("Error creating the command stderr file '{}': {}", path.display(), source))]
     CommandErrFile {
@@ -66,9 +67,6 @@ pub(crate) enum Error {
         path: PathBuf,
         backtrace: Backtrace,
     },
-
-    #[snafu(display("Error deserializing Settings: {} ", source))]
-    DeserializeSettings { source: deserialization::Error },
 
     #[snafu(display("Error creating the error file '{}': {}", path.display(), source))]
     ErrorFile {
@@ -166,9 +164,6 @@ pub(crate) enum Error {
 
     #[snafu(display("Cannot write to / as a file."))]
     RootAsFile { backtrace: Backtrace },
-
-    #[snafu(display("Error serializing Settings: {} ", source))]
-    SerializeSettings { source: serialization::Error },
 
     #[snafu(display("Unable to deserialize Bottlerocket settings: {}", source))]
     SettingsJson { source: serde_json::Error },
--- a/sources/logdog/src/log_request.rs
+++ b/sources/logdog/src/log_request.rs
@@ -9,9 +9,7 @@
 //! these provide the list of log requests that `logdog` will run.
 
 use crate::error::{self, Result};
-use datastore::deserialization::from_map;
-use datastore::serialization::to_pairs;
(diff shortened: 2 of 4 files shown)


sources/logdog/src/error.rs
```diff
@@ -4,20 +4,21 @@ use std::io;
...
-    ApiClient {
-        #[snafu(source(from(apiclient::Error, Box::new)))]
-        source: Box<apiclient::Error>,
-        uri: String,
+    #[snafu(display("Error calling apiclient: {}", source))]
+    ApiCommandFailure {
+        #[snafu(source(from(std::io::Error, Box::new)))]
+        source: Box<std::io::Error>,
     },
+    #[snafu(display("Error fetching settings via apiclient: {}", reason))]
```
bcressey, line 20: nit: leave a newline between each enum variant

sources/logdog/src/log_request.rs
```diff
@@ -166,24 +164,73 @@ where
...
+    match json {
+        serde_json::Value::Object(map) => {
+            for (k, v) in map {
+                let new_key = if prefix.is_empty() {
+                    k.clone()
+                } else {
+                    format!("{}.{}", prefix, k)
+                };
+                normalize(v, new_key, flattened)?;
+            }
```
bcressey: The use of `.` as a separator is ambiguous because some keys have dots in them. `settings.kernel.sysctl` is one but there are many others. This approach would corrupt the structure when unflattened.

I would look instead at `serde_json::Value` and [pointer_mut](https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html#method.pointer_mut) which takes a JSONPointer query. Then you could replace the value with "REDACTED" if it is a string.

sources/logdog/src/log_request.rs
```diff
@@ -192,13 +239,20 @@ where
 }
 
 /// Uses `apiclient` to request all settings from the apiserver and deserializes into a `Settings`
```
webern:
```suggestion
/// Uses `apiclient` to request all settings from the apiserver and deserializes into a `json::Value`
```

sources/logdog/src/log_request.rs
```diff
@@ -166,24 +164,73 @@ where
...
 
     // Filter all settings that match any of the "sensitive" patterns
     for pattern in SENSITIVE_SETTINGS_PATTERNS {
         let pattern =
             Pattern::new(pattern).context(error::ParseGlobPatternSnafu { pattern: *pattern })?;
-        settings_map.retain(|k, _| !pattern.matches(k.name().as_str()))
+        settings_map.retain(|k, _| !pattern.matches(k.as_str()))
     }
 
     // Serialize the map back to a `Settings` to remove the escaping so it writes nicely to file
```
webern:
```suggestion
    // Serialize the map back to a `json::Value` to remove the escaping so it writes nicely to file
```

sources/logdog/src/log_request.rs
```diff
@@ -166,39 +163,66 @@ where
...
-    let uri = constants::API_SETTINGS_URI;
-    let (_status, response_body) = apiclient::raw_request(constants::API_SOCKET, uri, "GET", None)
-        .await
-        .context(error::ApiClientSnafu { uri })?;
+/// Uses `apiclient` to request all settings from the apiserver and deserializes into a `json::Value`
+async fn get_settings() -> Result<serde_json::Value> {
+    let result = Command::new("/usr/bin/apiclient")
+        .arg("get")
+        .output()
+        .context(error::ApiCommandFailureSnafu)?;
```
bcressey, line 216: This changes the output shape to also include `os`, and the settings are wrapped in a `settings` container.

I kind of prefer this - strictly speaking, it's a breaking change, but it's more complete with the "os" structure.

Please make sure to note this sort of change in the commit message and in the description of the change.

bcressey: In the commit message you can add rationale as to why a breaking change is OK - this tool is intended to collect logs that are forwarded to support, and there's no specific contract around the format of the archive or any of its contents.

jmt-lab:
Updated commit message with description of the breaking change

sources/logdog/src/log_request.rs
```diff
@@ -9,9 +9,7 @@
 //! these provide the list of log requests that `logdog` will run.
 
 use crate::error::{self, Result};
-use datastore::deserialization::from_map;
-use datastore::serialization::to_pairs;
-use glob::{glob, Pattern};
+use glob::glob;
```
bcressey, line 12: We can drop the `glob` dependency it looks like?

bcressey (approved):

sources/logdog/src/log_request.rs
jmt-lab: I wish but it is used in another method i used udeps to validate

jmt-lab: its used by handle_glob_request

webern (approved):

PR #3964, opened 2024-05-16
Enable k8s reserved cpus

james-masson:
**Issue number:**

As per discussions with @yeazelm 

**Description of changes:**

Adds support for K8s reserved-cpus functionality

https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#explicitly-reserved-cpu-list

**Testing done:**

~~Warning - not functional.~~

~~Seems I've added a new package to Cargo for the migrations, but can't figure out how to only update the Cargo.lock with only my package changes.~~

Local build, unit-test, deployed AMI using new config.

**Terms of contribution:**

### DIFF
--- a/Release.toml
+++ b/Release.toml
@@ -319,4 +319,5 @@ version = "1.21.0"
     "migrate_v1.21.0_pod-infra-container-image-remove-settings-generator.lz4",
     "migrate_v1.21.0_pod-infra-container-image-affected-services.lz4",
     "migrate_v1.21.0_pod-infra-container-image-services.lz4",
+    "migrate_v1.21.0_k8s-reserved-cpus-v0-1-0.lz4",
 ]
--- a/packages/kubernetes-1.29/kubelet-config
+++ b/packages/kubernetes-1.29/kubelet-config
@@ -91,13 +91,17 @@ kubeReserved:
   {{/if}}
   {{/if}}
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
+{{#unless settings.kubernetes.reserved-cpus}}
 kubeReservedCgroup: "/runtime"
+{{/unless}}
 {{#if settings.kubernetes.system-reserved}}
 systemReserved:
   {{#each settings.kubernetes.system-reserved}}
   {{@key}}: "{{this}}"
   {{/each}}
+{{#unless settings.kubernetes.reserved-cpus}}
 systemReservedCgroup: "/system"
+{{/unless}}
 {{/if}}
 cpuCFSQuota: {{default true settings.kubernetes.cpu-cfs-quota-enforced}}
 cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
@@ -180,4 +184,7 @@ reservedMemory:
 {{/if}}
 {{/each}}
 {{/if}}
+{{/if}}
+{{#if settings.kubernetes.reserved-cpus}}
+reservedSystemCPUs: {{settings.kubernetes.reserved-cpus}}
 {{/if}}
--- a/sources/api/migration/migrations/v1.21.0/k8s-reserved-cpus-v0-1-0/src/main.rs
+++ b/sources/api/migration/migrations/v1.21.0/k8s-reserved-cpus-v0-1-0/src/main.rs
@@ -1,0 +1,18 @@
+use migration_helpers::common_migrations::AddSettingsMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// Add the option to set Kubernetes reserved-cpus
+fn run() -> Result<()> {
+    migrate(AddSettingsMigration(&["settings.kubernetes.reserved-cpus"]))
+}
+
+// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
+// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
+// https://github.com/shepmaster/snafu/issues/110
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("{}", e);
+        process::exit(1);
+    }
+}
--- a/sources/models/src/lib.rs
+++ b/sources/models/src/lib.rs
@@ -48,13 +48,13 @@ use std::net::IpAddr;
 use crate::de::{deserialize_limit, deserialize_mirrors, deserialize_node_taints};
 use modeled_types::{
     BootConfigKey, BootConfigValue, BootstrapContainerMode, CpuManagerPolicy, CredentialProvider,
-    DNSDomain, EtcHostsEntries, Identifier, IntegerPercent, KubernetesAuthenticationMode,
-    KubernetesBootstrapToken, KubernetesCloudProvider, KubernetesClusterDnsIp,
-    KubernetesClusterName, KubernetesDurationValue, KubernetesLabelKey, KubernetesLabelValue,
-    KubernetesQuantityValue, KubernetesReservedResourceKey, KubernetesTaintValue,
-    KubernetesThresholdValue, OciDefaultsCapability, OciDefaultsResourceLimitType,
-    PemCertificateString, SingleLineString, TopologyManagerPolicy, TopologyManagerScope, Url,
-    ValidBase64, ValidLinuxHostname,
+    DNSDomain, EtcHostsEntries, Identifier, IntegerPercent, KernelCpuSetValue,
+    KubernetesAuthenticationMode, KubernetesBootstrapToken, KubernetesCloudProvider,
+    KubernetesClusterDnsIp, KubernetesClusterName, KubernetesDurationValue, KubernetesLabelKey,
+    KubernetesLabelValue, KubernetesQuantityValue, KubernetesReservedResourceKey,
+    KubernetesTaintValue, KubernetesThresholdValue, OciDefaultsCapability,
+    OciDefaultsResourceLimitType, PemCertificateString, SingleLineString, TopologyManagerPolicy,
+    TopologyManagerScope, Url, ValidBase64, ValidLinuxHostname,
 };
 
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
@@ -137,6 +137,7 @@ struct KubernetesSettings {
     shutdown_grace_period_for_critical_pods: KubernetesDurationValue,
     memory_manager_reserved_memory: HashMap<Identifier, KubernetesMemoryReservation>,
     memory_manager_policy: KubernetesMemoryManagerPolicy,
+    reserved_cpus: KernelCpuSetValue,
 
     // Settings where we generate a value based on the runtime environment.  The user can specify a
     // value to override the generated one, but typically would not.
--- a/packages/kubernetes-1.23/kubelet-config
+++ b/packages/kubernetes-1.23/kubelet-config
@@ -87,13 +87,17 @@ kubeReserved:
   {{/if}}
   {{/if}}
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
+{{#unless settings.kubernetes.reserved-cpus}}
 kubeReservedCgroup: "/runtime"
+{{/unless}}
 {{#if settings.kubernetes.system-reserved}}
 systemReserved:
   {{#each settings.kubernetes.system-reserved}}
   {{@key}}: "{{this}}"
   {{/each}}
+{{#unless settings.kubernetes.reserved-cpus}}
 systemReservedCgroup: "/system"
+{{/unless}}
 {{/if}}
 cpuCFSQuota: {{default true settings.kubernetes.cpu-cfs-quota-enforced}}
 cpuManagerPolicy: {{default "none" settings.kubernetes.cpu-manager-policy}}
@@ -181,4 +185,7 @@ reservedMemory:
 {{/if}}
 {{/each}}
 {{/if}}
+{{/if}}
+{{#if settings.kubernetes.reserved-cpus}}
+reservedSystemCPUs: {{settings.kubernetes.reserved-cpus}}
 {{/if}}
--- a/packages/kubernetes-1.24/kubelet-config
+++ b/packages/kubernetes-1.24/kubelet-config
@@ -87,13 +87,17 @@ kubeReserved:
   {{/if}}
   {{/if}}
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
+{{#unless settings.kubernetes.reserved-cpus}}
 kubeReservedCgroup: "/runtime"
+{{/unless}}
 {{#if settings.kubernetes.system-reserved}}
 systemReserved:
   {{#each settings.kubernetes.system-reserved}}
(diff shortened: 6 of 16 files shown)


sources/api/migration/migrations/v1.20.0/k8s-reserved-cpus-v0-1-0/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
```
yeazelm: These will end up in a new directory. Right now it is pointing at `1.21.0` from https://github.com/bottlerocket-os/bottlerocket/blob/develop/Release.toml#L1.

Release.toml
```diff
@@ -311,5 +311,6 @@ version = "1.21.0"
     "migrate_v1.20.0_public-admin-container-v0-11-8.lz4",
     "migrate_v1.20.0_aws-control-container-v0-7-12.lz4",
     "migrate_v1.20.0_public-control-container-v0-7-12.lz4",
+    "migrate_v1.20.0_k8s-reserved-cpus-v0-1-0.lz4",
```
yeazelm: You'll want to put this in the next release. The list is currently empty but this would be the first one.

```suggestion
]
 "(1.20.0, 1.21.0)" = [
       "migrate_v1.21.0_k8s-reserved-cpus-v0-1-0.lz4",
]
```

yeazelm (commented):
This is looking like a reasonable addition, thanks for the PR! I'd like to note the migration has to go into the next release so I've put some comments in the PR to shift you there. This is the first PR that has a migration for 1.21.0 (which may become 1.20.1 but the maintainers can sort that out if it changes) so that is why there are no directories showing you the way. You'll just want to shift the path for the actual migration crate to `v1.21.0` and rename them. Otherwise the migration looks right from reading the code. If you can do the rename I'll kick off some testing to ensure it works.

Release.toml
james-masson: This should be fixed now.

sources/api/migration/migrations/v1.20.0/k8s-reserved-cpus-v0-1-0/Cargo.toml
james-masson: Should be fixed

Release.toml
```diff
@@ -312,4 +312,6 @@ version = "1.21.0"
     "migrate_v1.20.0_aws-control-container-v0-7-12.lz4",
     "migrate_v1.20.0_public-control-container-v0-7-12.lz4",
 ]
-"(1.20.0, 1.21.0)" = []
+"(1.20.0, 1.21.0)" = [
+    "migrate_v1.21.0_k8s-reserved-cpus-v0-1-0.lz4",
```
yeazelm:
```suggestion
"(1.20.1, 1.21.0)" = [
    "migrate_v1.21.0_k8s-reserved-cpus-v0-1-0.lz4",
```

This has shifted in `develop` since you published but otherwise this is looking right.

yeazelm (commented):
I was able to pull this branch and validate that the settings take. I couldn't verify that the reserved CPUs were actually reserved. Do you have some commands I we can document to use for validation of this change?

yeazelm:
I've tested the migration works as well. We are pretty close, as a tidiness ask, can you squash the commits into two commits?
* one to cover adding the setting and templates 
* and another to add the migration

If you can provide the commands to validate the actual functionality and clean up the commits, we should be in a good spot to merge! Thanks for the work on this @james-masson

packages/kubernetes-1.29/kubelet-config
```diff
@@ -91,13 +91,17 @@ kubeReserved:
   {{/if}}
   {{/if}}
   ephemeral-storage: "{{default "1Gi" settings.kubernetes.kube-reserved.ephemeral-storage}}"
+{{#unless settings.kubernetes.reserved-cpus}}
```
yeazelm, line 94: We should add this for all k8s versions that support this. I believe this is all of ours we currently support - 1.23 through 1.30.

james-masson:
> I was able to pull this branch and validate that the settings take. I couldn't verify that the reserved CPUs were actually reserved. Do you have some commands I we can document to use for validation of this change?

Given an 8 CPU core system as an example

1. Configure the system
```
[settings.kubernetes]
cpu-manager-policy = "static"
reserved-cpus = "0-1" # don't use these CPUs for any high-priority k8s workloads
```

2. Inject a guaranteed class pod into the cluster...


... with an integer number of CPUs that matches the number of CPUs that are *not* in the "reserved-cpus" mask - in this case 6

This assumes the node is otherwise empty, apart from daemonsets - otherwise you'll have to reduce the number of cpus the pod requests to get it to fit.

eg.

```
---
apiVersion: v1
kind: Pod
metadata:
  name: guarenteed-qos
spec:
  containers:
    - name: qos-demo
      image: nginx
      resources:
        limits:
          memory: "200Mi"
          cpu: "6"
```


3. Look at the cpuset that is assigned to the guaranteed class pod.

Assuming cgroup v2

```
kubectl exec -it <guaranteed pod> -- cat /sys/fs/cgroup/cpuset.cpus.effective
```
The set *should not* include the CPU numbers set in the "reserved-cpus" mask in step 1

4. Look at the cpuset that is assigned to a random daemonset

```
kubectl -n kube-system exec -it <daemonset pod> -- cat /sys/fs/cgroup/cpuset.cpus.effective
```

The set *should* include the CPU numbers set in the "reserved-cpus" mask in step 1, and also *should not* overlap with the guaranteed-class pod cpuset values.

packages/kubernetes-1.29/kubelet-config
james-masson: Done

Release.toml
james-masson: Updated to match current develop

james-masson:
@yeazelm - This has now been built and tested, and should have all the changes you asked for.

sources/models/src/lib.rs
```diff
@@ -295,6 +295,7 @@ struct KubernetesSettings {
     shutdown_grace_period_for_critical_pods: KubernetesDurationValue,
     memory_manager_reserved_memory: HashMap<Identifier, KubernetesMemoryReservation>,
     memory_manager_policy: KubernetesMemoryManagerPolicy,
+    reserved_cpus: SingleLineString,
```
arnaldo2792: I read the code of how the Kubernetes parses this line, and I think we should implement the same parsing logic in a struct. We already do something similar for other values that kubernetes expects in other places (see [KubernetesDurationValue]). This will prevent users from setting a random, since the API will refuse to apply the value, which provides a better experience than chasing errors in the kubelet logs. The new type could be `KubernetesCpuSet`, and implement what the kubernetes folks did [here].

[KubernetesDurationValue]: https://github.com/bottlerocket-os/bottlerocket/blob/d47e652b701c5b5e150146cb250971d263584f68/sources/models/modeled-types/src/kubernetes.rs#L820
[here]: https://github.com/kubernetes/utils/blob/d93618cff8a22d3aea7bf78d9d528fd859720c2d/cpuset/cpuset.go#L203

sources/api/migration/migrations/v1.21.0/k8s-reserved-cpus-v0-1-0/src/main.rs
```diff
@@ -0,0 +1,18 @@
+use migration_helpers::common_migrations::AddSettingsMigration;
+use migration_helpers::{migrate, Result};
+use std::process;
+
+/// Add the option to set Kubernetes reserved-cpus
+fn run() -> Result<()> {
+    migrate(AddSettingsMigration(&["settings.kubernetes.reserved-cpus"]))
```
arnaldo2792, line 7: Don't we need to limit this for k8s only? And have a conditional here to perform a `Noop` migration?

sources/models/src/lib.rs
james-masson: Arguably this is not a kubernetes specific string - it's a kernel concept,  "cpuset string"  is effectively a standard.

https://docs.kernel.org/admin-guide/cgroup-v2.html#cpuset-interface-files

It is probably a useful concept to have in Bottlerocket, as my future PRs will include a lot more use of "cpuset string" that are not tied to Kubernetes.

So, if you want this, best it's `KernelCpuSet` - not `KubernetesCpuSet`

But it's almost academic - I lack the skills to do this in Rust ;-)

yeazelm: https://github.com/bottlerocket-os/bottlerocket/commit/0c9d63c7b55fe8ce977b3cd54ca4c8ba21c550d2 I wrote up a quick approach to this to assist with getting you there. This should work for your needs but I didn't test the integration myself. I'm happy to help get you the rest of the way if needed on this change!

sources/api/migration/migrations/v1.21.0/k8s-reserved-cpus-v0-1-0/src/main.rs
james-masson: I couldn't find any example of that for an "Add a new setting" migration.

I found

https://github.com/bottlerocket-os/bottlerocket/blob/1.20.x/sources/api/migration/migrations/v1.20.0/add-ntp-default-options-v0-1-0/src/main.rs

and

https://github.com/bottlerocket-os/bottlerocket/blob/1.19.x/sources/api/migration/migrations/v1.19.2/add-ecs-enable-container-metadata/src/main.rs

and

https://github.com/bottlerocket-os/bottlerocket/blob/1.19.x/sources/api/migration/migrations/v1.19.0/add-additional-ecs-settings/src/main.rs

all of which do the same as mine...

sources/models/src/lib.rs
james-masson: thanks @yeazelm - I cherry-picked your commit and squashed it in here.
Tested and working.

james-masson: negative case correctly gives.

```
[    4.753115] early-boot-config[1198]: Error PATCHing '/settings?tx=bottlerocket-launch': Status 400 when PATCHing /settings?tx=bottlerocket-launch: Json deserialize error: Unable to deserialize into KernelCpuSetValue: Invalid Kernel CpuSet value 'abc' at line 1 column 1698
```

sources/api/migration/migrations/v1.21.0/k8s-reserved-cpus-v0-1-0/src/main.rs
yeazelm: https://github.com/bottlerocket-os/bottlerocket/blob/a1bcca5556c9f0053df16aabf36920adae4e6944/sources/api/migration/migrations/archived/v1.15.0/oci-defaults-docker-setting/src/main.rs#L2 is the inverse of what you need here.

```suggestion
     if cfg!(variant_runtime = "k8s") {
         migrate(AddSettingsMigration(&["settings.kubernetes.reserved-cpus"]))
    } else {
        migrate(NoOpMigration);
    }
```

The example I linked to used `?` but I don't think it is required.

james-masson: I tried a few conbinations of this and the other change you linked to.

I'm stuck on...

```
  #19 0.466 7  | /     if cfg!(variant_runtime = "k8s") {                                                                                                             
  #19 0.466 8  | |         migrate(AddSettingsMigration(&["settings.kubernetes.reserved-cpus"]))
  #19 0.466    | |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `()`, found `Result<(), Error>`                             
  #19 0.466 9  | |     } else {                                                                                                                                       
  #19 0.466 10 | |        migrate(NoOpMigration);                                                                                                                     
  #19 0.466 11 | |     }                                                                                                                                              
  #19 0.466    | |_____- expected this to be `()`                                                                                                                     
  #19 0.466    |
  #19 0.466    = note: expected unit type `()`
  #19 0.466                    found enum `Result<(), migration_helpers::error::Error>`
  #19 0.466 help: consider using a semicolon here
  #19 0.466    |
  #19 0.466 8  |         migrate(AddSettingsMigration(&["settings.kubernetes.reserved-cpus"]));
  #19 0.466    |                                                                              +
  #19 0.466 help: consider using a semicolon here
  #19 0.466    |
```

... any hints?

yeazelm: Oh, sorry, there was a typo in my suggestion, there should not be a ; at the end of the first migration call to end the statement.
```suggestion
 if cfg!(variant_runtime = "k8s") {
         migrate(AddSettingsMigration(&["settings.kubernetes.reserved-cpus"]))
    } else {
        migrate(NoOpMigration)
    }
```
The error is complaining about mixing and matching statements vs an expression. In this case, if you leave off the `;` it will return the results of the function call, in this case `migration()` so want both statements to not have `;` but then you also don't want `Ok(())` at the end of your `run()` function. So the code I used to get this to compile looks like this for the entire file:
```
use migration_helpers::common_migrations::{AddSettingsMigration, NoOpMigration};
use migration_helpers::{migrate, Result};
use std::process;

/// Add the option to set Kubernetes reserved-cpus
fn run() -> Result<()> {
     if cfg!(variant_runtime = "k8s") {
         migrate(AddSettingsMigration(&["settings.kubernetes.reserved-cpus"]))
    } else {
        migrate(NoOpMigration)
    }
}

// Returning a Result from main makes it print a Debug representation of the error, but with Snafu
// we have nice Display representations of the error, so we wrap "main" (run) and print any error.
// https://github.com/shepmaster/snafu/issues/110
fn main() {
    if let Err(e) = run() {
        eprintln!("{}", e);
        process::exit(1);
    }
}
```

arnaldo2792: Either this suggestion, or adding `?;` and `Ok(())` at the end should be enough.

bcressey, line 7: This migration looks fine to me - it doesn't matter that `aws-ecs-*` doesn't have these settings, it'll just be a no-op on downgrade because the paths won't exist.

yeazelm (approved):
After working through @bcressey's comment I agree the migration is fine as is. So approving.

sources/api/migration/migrations/v1.21.0/k8s-reserved-cpus-v0-1-0/src/main.rs
bcressey: We only need a conditional migration in the case where settings are added in one version for some variants (`aws-k8s-*`) and then in a subsequent version for other variants (`aws-ecs-*`). These settings won't ever be added for `aws-ecs-*` so it's fine.

arnaldo2792 (approved):

bcressey (approved):
Looks good, thanks!

bcressey (approved):
Looks good - rebase just fixed the merge conflict in the `use modeled_types::...` lines at the top of `sourcse/models/src.lib.rs`.

bcressey (approved):
`cargo fmt` appeased!

PR #3971, opened 2024-05-17
Pki settings extension

### LINKED ISSUE #3657: OOTB: Port `pki` settings model to settings extensions
Create a settings extension for `pki`.

### Definition of done
A new `pki` crate is created under `sources/settings-extensions`, and all variants use the model defined in that crate in place of the existing settings. Additionally, a basic `settings-pki` RPM package is created that installs the extension binary.

mgsharm:
**Issue number:**

Closes #[3657](https://github.com/bottlerocket-os/bottlerocket/issues/3657)

**Description of changes:**

- Creates `pki` settings extension and uses it in every variant's settings model. 
- Creates `settings-pki` RPM package that installs the extension binary.

**Testing done:**

- Built `aws-dev` variant with the `settings-pki` installed. Launched `ec2` instance with the `aws-dev` variant ami. Connected with the instance via `SSM` to run `apiclient` commands.
- Called apiclient to verify the `settings-pki` worked as expected.
```
[ssm-user@control]$ apiclient set pki.my-trusted-bundle.data="LS0tLS1..." \
> pki.my-trusted-bundle.trusted=true
```

```
[ssm-user@control]$ apiclient get settings.pki
{  
   "settings":  {
       "pki": {      
           "my-trusted-bundle": {       
                "data": "LS0tLS1...",
                 "trusted": true
            }
       }
   }
}
```

- Also tested by building locally.

```
> cert_content=$(cat test-pem)
> json_value="{\"foo\":{\"data\":\"$cert_content\",\"trusted\":true}}"
> cargo run proto1 set --setting-version v1 --value "$json_value"
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.24s
     Running `/home/fedora/bottlerocket/sources/target/debug/settings-extension-pki proto1 set --setting-version v1 --value '{"foo":{"data":"LS0tLS1...,"trusted":true}}'`
```  

**Terms of contribution:**


packages/settings-pki/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+source-groups = [
+    "settings-extensions/pki"
+]
+
+# RPM BuildRequires
+[build-dependencies]
+glibc = { path = "../glibc" }
+
+# RPM Requires
+[dependencies]
```
jmt-lab: nit: new line at end of files, probably good idea to set this as a setting in your editor of choice

sources/settings-extensions/pki/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+        }))
+    }
+
+    fn validate(_value: Self, _validated_settings: Option<serde_json::Value>) -> Result<()> {
+        // Validate anything that parses as PkiSettingsV1.
+        Ok(())
+    }
+}
+
+impl LinearlyMigrateable for PkiSettingsV1 {
```
sam-berning: this trait is not needed for the `NullMigrator`, it's only used to provide an interface to the `LinearMigrator`. if we later need migrations for this setting, we can implement this, but for now let's remove it.

mgsharm: Removed it.

packages/settings-pki/Cargo.toml
mgsharm: Done.

jmt-lab (approved):

sources/settings-extensions/pki/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+        let mut expected_pki: HashMap<Identifier, PemCertificate> = HashMap::new();
+        expected_pki.insert(
+            Identifier::try_from("foo").unwrap(),
+            PemCertificate {
+                data: Some(PemCertificateString::try_from(VALID_PEM).unwrap()),
+                trusted: Some(true),
+            },
+        );
+
+        assert_eq!(pki, PkiSettingsV1 { pki: expected_pki });
```
sam-berning, line 107: I'm realizing that many of the other settings extensions don't have this, but we should test the serialize path as well, e.g https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/settings-extensions/ntp/src/lib.rs#L87-L88

sources/settings-extensions/pki/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+            Ok(GenerateResult::Complete(PkiSettingsV1 {
+                pki: HashMap::new(),
+            }))
+        )
+    }
+
+    #[test]
+    fn test_serde_pki() {
+        let test_json = format!(
+            r#"{{
```
sam-berning: why the double `{` here?

mgsharm: Double braces in `format!` are used to produce a JSON string that includes curly braces, and without escaping, this will give errors.

sources/settings-extensions/pki/src/lib.rs
mgsharm: Done, updated the test.

sam-berning (approved):

Issue #3060, opened 2023-04-27
create a local NVMe RAID-0


FernandoMiguel:
**What I'd like:**
AL now supports creating RAID of instance store volumes
That work was done in this PR 
https://github.com/awslabs/amazon-eks-ami/pull/1171

Could bottlerocket implement the same or similar approach so that EKS nodes benefit from that feature?

arnaldo2792:
Hey @FernandoMiguel, could you please have a look at [this discussion]? There are instructions there to create a RAID-0 array for containerd, using bootstrap containers. Will this be sufficient to support your use case?

[this discussion]: https://github.com/bottlerocket-os/bottlerocket/discussions/1991#discussioncomment-3265188

FernandoMiguel:
@arnaldo2792 i rather prefer the native way of al2 to having to modify bottlerocket boot myself with no future garantes it will work.
Hence looking for a similar approach to be implemented into bottlerocket so we can take  benefit of instance store disks.

stevehipwell:
@arnaldo2792 I think having an OCI image able to run arbitrary scripts, as requested in https://github.com/bottlerocket-os/bottlerocket/issues/2010, would make it easier to manually implement this as a short term solution. But ideally an API configuration to enable this would be my preference.

CC @bwagner5

zmrow:
Thanks for bringing this to our attention @stevehipwell and @FernandoMiguel !  We aren't currently working on this, but will follow up here if/when we start in on it.

denraf:
The problem with the boot container is that karpenter is not aware of the ephemeral ssd storage on an instance type like c6id.

bryantbiggs:
> The problem with the boot container is that karpenter is not aware of the ephemeral ssd storage on an instance type like c6id.

Sure it is, you can query that from the AWS EC2 API

denraf:
> > The problem with the boot container is that karpenter is not aware of the ephemeral ssd storage on an instance type like c6id.
> 
> Sure it is, you can query that from the AWS EC2 API

Unfortunately not.
E.g. c6id.2xlarge
https://karpenter.sh/preview/concepts/instance-types/#c6id2xlarge
ephemeral-storage: 17G

Draft PR to resolve it for AL2: https://github.com/aws/karpenter/pull/3827
In that PR it is stated work is needed in bottlerocket to support that.
Which brings us back to this issue.

bryantbiggs:
[...]

PR #3977, opened 2024-05-21
Add udev rule to create symlinks using EBS volumes' device names

### LINKED ISSUE #3975: Create symlinks to devices using the device name configured for EBS volumes
**What I'd like:**

A udev rule to create symlinks using the device name configured for EBS volumes, so that users have a predictable name they can use to access their volumes

arnaldo2792:
**Issue number:**

Closes #3975

**Description of changes:**

---
#### V2

The first two commits in the series add tools to interact with NVMe devices. They were used to query the device info instead of issuing `ioctl` syscalls, like in the first version of the PR.

The third commit extends `ghostdog` and adds the `ebs-device-name` command to emit the device name from the vendor data returned by `nvme-cli`.

The last commit adds two `udev` rules that use the value emitted by `ghostdog` to create the required symlinks.

---

#### V1

The first commit in the series extends `ghostdog` with a command to query the device using `ioctl`. The implementation was inspired by the [script used in Amazon Linux] and [`libnvme`]. Since the `ioctl` syscall is an actual `C` API, the structs used to interact with it include `repr(C)` so that the struct is in a compatible format for `C`. The [Rustonomicon] suggests the usage of `rust-bindgen` to generate rust bindings to interact with `C` APIs. However, adding the entire `libnvme` would be overkill so I added just the structures that are needed to query the device information we want.

For the created structures, I used [`kvm-bindings`] as a reference to understand how the structures should be configured, as well as how the Rust types are mapped to `C` types.

The second commit in the series updates the existing udev rules to create symlinks to the EBS volumes, for both the devices and the partitions, using the device name configured for the volume.

[script used in Amazon Linux]: (https://github.com/amazonlinux/amazon-ec2-utils/blob/main/ebsnvme-id)
[`libnvme`]: https://github.com/linux-nvme/libnvme
[Rustonomicon]: https://doc.rust-lang.org/nomicon/other-reprs.html
[`kvm-bindings`]: https://github.com/rust-vmm/kvm-bindings

**Testing done:**

With `aws-ecs-2`, x86_64, I confirmed that the symlinks are created pointing at the correct locations. In my instance, `xvda` corresponds to the root volume (`nvme0n1`) and `xvdb` corresponds to the data volume (`nvme1n1`):

```
bash-5.1# ls -la /dev/xv*
lrwxrwxrwx. 1 root root  7 May 21 15:16 /dev/xvda -> nvme0n1
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvda1 -> nvme0n1p1
lrwxrwxrwx. 1 root root 10 May 21 15:16 /dev/xvda10 -> nvme0n1p10
lrwxrwxrwx. 1 root root 10 May 21 15:16 /dev/xvda11 -> nvme0n1p11
lrwxrwxrwx. 1 root root 10 May 21 15:16 /dev/xvda12 -> nvme0n1p12
lrwxrwxrwx. 1 root root 10 May 21 15:16 /dev/xvda13 -> nvme0n1p13
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvda2 -> nvme0n1p2
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvda3 -> nvme0n1p3
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvda4 -> nvme0n1p4
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvda5 -> nvme0n1p5
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvda6 -> nvme0n1p6
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvda7 -> nvme0n1p7
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvda8 -> nvme0n1p8
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvda9 -> nvme0n1p9
lrwxrwxrwx. 1 root root  7 May 21 15:16 /dev/xvdb -> nvme1n1
lrwxrwxrwx. 1 root root  9 May 21 15:16 /dev/xvdb1 -> nvme1n1p1
```

**Terms of contribution:**

### DIFF
--- a/packages/os/ebs-volumes.rules
+++ b/packages/os/ebs-volumes.rules
@@ -1,8 +1,15 @@
 ACTION=="remove", GOTO="ebs_volumes_end"
+KERNEL!="nvme*", GOTO="ebs_volumes_end"
 SUBSYSTEM!="block", GOTO="ebs_volumes_end"
-ENV{DEVTYPE}!="disk", GOTO="ebs_volumes_end"
+ATTRS{model}!="Amazon Elastic Block Store", GOTO="ebs_volumes_end"
 
 # Follow AWS recommendation of never timing out IO on EBS volumes attached via NVMe
-KERNEL=="nvme*", ATTRS{model}=="Amazon Elastic Block Store", ATTR{queue/io_timeout}="4294967295"
+ENV{DEVTYPE}=="disk", ATTR{queue/io_timeout}="4294967295"
+
+# Add symlink for disk
+KERNEL=="nvme[0-9]*n[0-9]*", ENV{DEVTYPE}=="disk", IMPORT{program}="/usr/bin/ghostdog ebs-device-name $devnode", SYMLINK+="$env{XVD_DEVICE_NAME}"
+
+# Add symlink for partition
+KERNEL=="nvme[0-9]*n[0-9]*p[0-9]*", ENV{DEVTYPE}=="partition", IMPORT{parent}="XVD_DEVICE_NAME", SYMLINK+="$env{XVD_DEVICE_NAME}$number"
 
 LABEL="ebs_volumes_end"
--- a/sources/ghostdog/src/main.rs
+++ b/sources/ghostdog/src/main.rs
@@ -8,11 +8,15 @@ use gptman::GPT;
 use hex_literal::hex;
 use lazy_static::lazy_static;
 use signpost::uuid_to_guid;
-use snafu::ResultExt;
+use snafu::{ensure, ResultExt};
 use std::collections::HashSet;
 use std::fs;
 use std::io::{Read, Seek};
 use std::path::PathBuf;
+use std::process::Command;
+
+const NVME_CLI_PATH: &str = "/sbin/nvme";
+const NVME_IDENTIFY_DATA_SIZE: usize = 4096;
 
 #[derive(FromArgs, PartialEq, Debug)]
 /// Manage ephemeral disks.
@@ -25,12 +29,21 @@ struct Args {
 #[argh(subcommand)]
 enum SubCommand {
     Scan(ScanArgs),
+    EbsDeviceName(EbsDeviceNameArgs),
 }
 
 #[derive(FromArgs, PartialEq, Debug)]
 #[argh(subcommand, name = "scan")]
 /// Scan a device to see if it is an ephemeral disk.
 struct ScanArgs {
+    #[argh(positional)]
+    device: PathBuf,
+}
+
+#[derive(FromArgs, PartialEq, Debug)]
+#[argh(subcommand, name = "ebs-device-name")]
+/// Returns the device name used for the EBS device
+struct EbsDeviceNameArgs {
     #[argh(positional)]
     device: PathBuf,
 }
@@ -44,6 +57,11 @@ fn run() -> Result<()> {
             let mut f = fs::File::open(&path).context(error::DeviceOpenSnafu { path })?;
             let device_type = find_device_type(&mut f)?;
             emit_device_type(&device_type);
+        }
+        SubCommand::EbsDeviceName(ebs_device_name) => {
+            let path = ebs_device_name.device;
+            let device_name = find_device_name(format!("{}", path.display()))?;
+            emit_device_name(&device_name);
         }
     }
     Ok(())
@@ -75,9 +93,41 @@ where
     Ok(device_type.to_string())
 }
 
+/// Finds the device name using the nvme-cli
+fn find_device_name(path: String) -> Result<String> {
+    // nvme-cli writes the binary data to STDOUT
+    let output = Command::new(NVME_CLI_PATH)
+        .args(["id-ctrl", &path, "-b"])
+        .output()
+        .context(error::NvmeCommandSnafu { path: path.clone() })?;
+
+    parse_device_name(&output.stdout, path)
+}
+
+/// Parses the device name from the binary data returned by nvme-cli
+fn parse_device_name(device_info: &[u8], path: String) -> Result<String> {
+    // Bail out if the data returned isn't complete
+    ensure!(
+        device_info.len() == NVME_IDENTIFY_DATA_SIZE,
+        error::InvalidDeviceInfoSnafu { path }
+    );
+
+    // The vendor data is stored at the last 1024 bytes
+    // The device name is stored at the first 32 bytes of the vendor data
+    let offset = NVME_IDENTIFY_DATA_SIZE - 1024;
+    let device_name = &device_info[offset..offset + 32];
+
+    Ok(String::from_utf8_lossy(device_name).trim_end().to_string())
+}
+
 /// Print the device type in the environment key format udev expects.
 fn emit_device_type(device_type: &str) {
     println!("BOTTLEROCKET_DEVICE_TYPE={}", device_type);
+}
+
+/// Print the device name in the environment key format udev expects.
+fn emit_device_name(device_name: &str) {
+    println!("XVD_DEVICE_NAME={}", device_name)
 }
 
 // Known system partition types for Bottlerocket.
@@ -114,6 +164,13 @@ mod error {
             path: std::path::PathBuf,
             source: std::io::Error,
         },
+        #[snafu(display("Failed to execute NVMe command for device '{}': {}", path.display(), source))]
+        NvmeCommand {
+            path: std::path::PathBuf,
+            source: std::io::Error,
+        },
+        #[snafu(display("Invalid device info for device '{}'", path.display()))]
+        InvalidDeviceInfo { path: std::path::PathBuf },
     }
 }
 
--- a/packages/libnvme/0001-linux-Fix-uninitialized-variables.patch
+++ b/packages/libnvme/0001-linux-Fix-uninitialized-variables.patch
@@ -1,0 +1,40 @@
+From a0803d1d6fb6d899cd22bd7b37ee27859bfdf155 Mon Sep 17 00:00:00 2001
+From: Tomas Bzatek <tbzatek@redhat.com>
+Date: Fri, 3 May 2024 17:19:39 +0200
+Subject: [PATCH] linux: Fix uninitialized variables
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+In file included from ../src/nvme/linux.c:40:
+In function ‘freep’,
+    inlined from ‘nvme_get_telemetry_log’ at ../src/nvme/linux.c:169:23:
+../src/nvme/cleanup.h:24:9: warning: ‘log’ may be used uninitialized [-Wmaybe-uninitialized]
+   24 |         free(*(void **)p);
+      |         ^~~~~~~~~~~~~~~~~
+../src/nvme/linux.c: In function ‘nvme_get_telemetry_log’:
+../src/nvme/linux.c:169:30: note: ‘log’ was declared here
+  169 |         _cleanup_free_ void *log;
+      |                              ^~~
+
+Signed-off-by: Tomas Bzatek <tbzatek@redhat.com>
+---
+ src/nvme/linux.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/src/nvme/linux.c b/src/nvme/linux.c
+index 25196fd5..35976011 100644
+--- a/src/nvme/linux.c
++++ b/src/nvme/linux.c
+@@ -166,7 +166,7 @@ int nvme_get_telemetry_log(int fd, bool create, bool ctrl, bool rae, size_t max_
+ 
+ 	struct nvme_telemetry_log *telem;
+ 	enum nvme_cmd_get_log_lid lid;
+-	_cleanup_free_ void *log;
++	_cleanup_free_ void *log = NULL;
+ 	void *tmp;
+ 	int err;
+ 	size_t dalb;
+-- 
+2.44.0
+
--- a/packages/libnvme/Cargo.toml
+++ b/packages/libnvme/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "libnvme"
+version = "0.1.0"
+edition = "2021"
+publish = false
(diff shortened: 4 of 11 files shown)


packages/os/ebs-volumes.rules
```diff
@@ -1,8 +1,10 @@
 ACTION=="remove", GOTO="ebs_volumes_end"
-SUBSYSTEM!="block", GOTO="ebs_volumes_end"
-ENV{DEVTYPE}!="disk", GOTO="ebs_volumes_end"
+
+# Create symlinks for the device name configured in EC2, for both block devices and partitions
+KERNEL=="nvme[0-9]*n[0-9]*", ENV{DEVTYPE}=="disk", ATTRS{model}=="Amazon Elastic Block Store", PROGRAM="/usr/bin/ghostdog ebs-device-name /dev/$kernel", SYMLINK+="$result"
```
bcressey, line 4: If you set `ENV{XVD_DEVICE_NAME}` in the parent, you should be able to import it here via `IMPORT{parent}="XVD_DEVICE_NAME"` to avoid the second binary invocation.

packages/os/ebs-volumes.rules
```diff
@@ -1,8 +1,10 @@
...
+KERNEL=="nvme[0-9]*n[0-9]*p[0-9]*", ENV{DEVTYPE}=="partition", ATTRS{model}=="Amazon Elastic Block Store", PROGRAM="/usr/bin/ghostdog ebs-device-name /dev/$kernel", SYMLINK+="$result$number"
```
bcressey: If you set `ENV{XVD_DEVICE_NAME}` in the parent, you should be able to import it here via `IMPORT{parent}="XVD_DEVICE_NAME"` to avoid the second binary invocation.

packages/os/ebs-volumes.rules
bcressey: I'd sort of prefer to do this as part of the `ghostdog scan` operation, but it's OK to have a different subcommand.

I'd like it to take similar arguments though. It should also be set up with similar guards, and run after the io_timeout rule:
```
# add these guards to the top, after ACTION=="remove"
KERNEL!="nvme*", GOTO="ebs_volumes_end"
ATTRS{model}!="Amazon Elastic Block Store", GOTO="ebs_volumes_end"
...

# Set XVD_DEVICE_NAME for disk
KERNEL=="nvme[0-9]*n[0-9]*", ENV{DEVTYPE}=="disk", IMPORT{program}="/usr/bin/ghostdog ebs-device-name $devnode"

# Set XVD_DEVICE_NAME for partition
KERNEL=="nvme[0-9]*n[0-9]*p[0-9]*", ENV{DEVTYPE}=="partition", IMPORT{parent}="{XVD_DEVICE_NAME}"

# Use XVD_DEVICE_NAME
XVD_DEVICE_NAME=="xvd*", SYMLINK+="$env{XVD_DEVICE_NAME}"
```
```

File: sources/ghostdog/src/nvme.rs [hunk 3]
```diff
@@ -0,0 +1,362 @@
...
+    actp: __le16,
+    apws: __u8,
+    rsvd23: [__u8; 9],
+}
+
+#[repr(C)]
+#[derive(Debug)]
+/// Structure returned by the NVME_IDENTIFY_CTRL command
+/// See: https://github.com/linux-nvme/libnvme/blob/1ec2432ad0f71a8fc7e37548405d52923dd3dc6e/src/nvme/types.h#L1365
+struct NvmeIdCtrl {
```
larvacea: Hooray, this struct has the same number of members as the types.h struct.

sources/ghostdog/src/nvme.rs
```diff
@@ -0,0 +1,362 @@
...
+
+#[derive(Debug)]
+/// Represents an EBS NVME volume
+pub(crate) struct EBSNvmeDevice {
+    pub(crate) device_name: String,
+}
+
+impl From<NvmeIdCtrl> for EBSNvmeDevice {
+    fn from(value: NvmeIdCtrl) -> Self {
+        // The device name is within the first 32 bytes of the vendor specific data, which
```
larvacea: Does the name start at byte 0 of vendor specific data? If the name is shorter than 32 bytes, are the remaining bytes guaranteed to be whitespace (including, as Rust helpfully points out, zero bytes)?
Nit: if the name starts at byte 0, one could use trim_end() rather than trim(). Is it worth testing for a non-blank device name here (just to make for a more informative error message)?

sources/ghostdog/src/nvme.rs
```diff
@@ -0,0 +1,362 @@
...
+    // to this struct will be send in the ioctl command.
+    let mut ctrl = NvmeIdCtrl::default();
+    let mut cmd = NvmePassthruCmd {
+        opcode: NVME_ADMIN_IDENTIFY,
+        data_len: std::mem::size_of::<NvmeIdCtrl>() as u32,
+        addr: addr_of_mut!(ctrl) as u64,
+        // This is required, otherwise the syscall always returns the same device; this
+        // value corresponds to what Amazon Linux does.
+        // See: https://github.com/amazonlinux/amazon-ec2-utils/blob/09b53f2a20c3e4d227acfb452138fdb672baaa1a/ebsnvme-id#L114
+        cdw10: 1,
```
larvacea: This is defined in [NVME 1.1](https://www.nvmexpress.org/wp-content/uploads/NVM-Express-1_1a.pdf), Figure 81 (page 84). The value 1 means "The Identify Controller data structure is returned to
the host."

sources/ghostdog/src/nvme.rs
```diff
@@ -0,0 +1,362 @@
...
+mod test {
+    use super::*;
+    /// The size of the NvmeIdCtrl struct
+    const NVME_IDENTIFY_DATA_SIZE: u32 = 4096;
+
+    #[test]
+    // Test that the size of the structure matches the expected size, in case the
+    // structure shape is changed. This is forces developers to update the expected
+    // size of the struct.
+    fn test_nvme_id_ctrl_size() {
```
larvacea: Thank you! I was so relieved when I saw this test. Yes, by all means, let us validate that the huge lump of repr(C) definition actually added up to 4k bytes.

larvacea (approved):
Looks correct to me.

sources/ghostdog/src/main.rs
```diff
@@ -75,11 +93,43 @@ where
...
+}
+
 /// Print the device type in the environment key format udev expects.
 fn emit_device_type(device_type: &str) {
     println!("BOTTLEROCKET_DEVICE_TYPE={}", device_type);
 }
 
+/// Print the device name in the environment key format udev expects.
+fn emit_device_name(device_name: &str) {
+    println!("XVD_DEVICE_NAME={}", device_name)
```
larvacea, line 130: Per https://docs.aws.amazon.com/ebs/latest/userguide/nvme-ebs-volumes.html I believe this is fine because:

* Device names are at most 32 bytes long, and unused trailing bytes are space characters.
* Device names do not contain any characters that would require quoting or escaping.

Therefore this will give us a valid environment key assignment of the form XVD_DEVICE_NAME=/some/stuff and udev will be happy.

packages/os/ebs-volumes.rules
```diff
@@ -1,8 +1,14 @@
...
 
 # Follow AWS recommendation of never timing out IO on EBS volumes attached via NVMe
-KERNEL=="nvme*", ATTRS{model}=="Amazon Elastic Block Store", ATTR{queue/io_timeout}="4294967295"
+KERNEL=="nvme*", ENV{DEVTYPE}=="disk", ATTRS{model}=="Amazon Elastic Block Store", ATTR{queue/io_timeout}="4294967295"
+
+# Add symlink for disk
+KERNEL=="nvme[0-9]*n[0-9]*", ENV{DEVTYPE}=="disk", IMPORT{program}="/usr/bin/ghostdog ebs-device-name $devnode", SYMLINK+="$env{XVD_DEVICE_NAME}"
+
+# Add symlink for partition
+KERNEL=="nvme[0-9]*n[0-9]*p[0-9]*", ENV{DEVTYPE}=="partition", IMPORT{parent}="XVD_DEVICE_NAME", SYMLINK+="$env{XVD_DEVICE_NAME}$number"
```
larvacea, line 13: What form do you expect XVD_DEVICE_NAME to take? Is appending the partition number to that going to be unambiguous? I ask because if XVD_DEVICE_NAME is something like /dev/sda1, you could get a symlink for /dev/sda11 here.

larvacea (approved):
You might wish to update the lookaside cache and rerun the Github CI. If you're happy with appending the partition number to XVD_DEVICE_NAME, I'm happy (see comment on ebs-volumes.rules).

packages/os/ebs-volumes.rules
larvacea: I see that this is exactly what AL2023 does, so That Will Be Fine (probably, not provably, but what do you want). The XVD_DEVICE_NAME from the controller is of the form xvda, and the partitions get named xvda1, xvda128, and so on.

larvacea: For attached-after-boot EBS, /dev/sdb and /dev/sdb1 appear. Also fine.

packages/release/release.spec
```diff
@@ -122,6 +122,7 @@ Requires: %{_cross_os}kexec-tools
 Requires: %{_cross_os}keyutils
 Requires: %{_cross_os}makedumpfile
 Requires: %{_cross_os}netdog
+Requires: %{_cross_os}nvme-cli
```
bcressey: nit: we can omit the dependency in "release" since it will come in via `ghostdog`

sources/ghostdog/src/main.rs
```diff
@@ -75,11 +93,43 @@ where
...
+    // Bail out if the data returned isn't complete
+    ensure!(
+        device_info.len() == NVME_IDENTIFY_DATA_SIZE,
+        error::InvalidDeviceInfoSnafu { path }
+    );
+
+    // The vendor data is stored at the last 1024 bytes
+    // The device name is stored at the first 32 bytes of the vendor data
+    let device_name =
+        &device_info[NVME_IDENTIFY_DATA_SIZE - 1024..NVME_IDENTIFY_DATA_SIZE - 1024 + 32];
```
bcressey:
```suggestion
    let offset = NVME_IDENTIFY_DATA_SIZE - 1024;
    let device_name = &device_info[offset..offset + 32];
```

packages/os/ebs-volumes.rules
```diff
@@ -1,8 +1,14 @@
 ACTION=="remove", GOTO="ebs_volumes_end"
-SUBSYSTEM!="block", GOTO="ebs_volumes_end"
-ENV{DEVTYPE}!="disk", GOTO="ebs_volumes_end"
+KERNEL!="nvme*", GOTO="ebs_volumes_end"
+ATTRS{model}!="Amazon Elastic Block Store", GOTO="ebs_volumes_end"
 
 # Follow AWS recommendation of never timing out IO on EBS volumes attached via NVMe
-KERNEL=="nvme*", ATTRS{model}=="Amazon Elastic Block Store", ATTR{queue/io_timeout}="4294967295"
+KERNEL=="nvme*", ENV{DEVTYPE}=="disk", ATTRS{model}=="Amazon Elastic Block Store", ATTR{queue/io_timeout}="4294967295"
```
bcressey: nit: restore the previous guard and omit the redundant `KERNEL` and `ATTRS{model}` checks:
```suggestion
SUBSYSTEM=="block", ENV{DEVTYPE}=="disk", ATTR{queue/io_timeout}="4294967295"
```

(Is it OK to keep the `SUBSYSTEM!="block" guard at the top? Or did that cause partitions to be skipped? Not sure why you removed it. If it works it's better to have it at the top and omit it here also.)

bcressey (approved):
LGTM overall. The udev rules can be tweaked a bit for simplicity.

For readability, I prefer udev rules that bail out and skip to the end immediately if they don't match what the rule is trying to cover. And any conditions checked at top don't need to be checked again in the body of the rules file.

arnaldo2792:
(Forced push addresses nits left in previous reviews.)

bcressey (approved):

apexlnc:
Hey @bcressey @arnaldo2792, 

Thanks for working on this!

Just so I fully understand, this introduces native support for instance store local NVME volumes without having to build a custom bootstrap container, right? 

Will this work with Karpenter's RAID0 feature out of the box?
https://github.com/aws/karpenter-provider-aws/pull/4735

Even with the bootstrap container workaround, I believe Karpenter had issues allocating the volumes for use, as described here: 
https://github.com/bottlerocket-os/bottlerocket/issues/3060

arnaldo2792:
Hey @ikevinc thanks for the kind words! Just a clarification, this does not solve the problem you are pointing out. This PR introduced a udev rule that allows users to identify NVME devices using EBS device mappings (see [this]). The feature that you want is currently being worked in https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/15 and https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/62. With these two changes, you will be able to use ephemeral storage and configure a RAID 0 array, without requiring you to have bootstrap containers.

[this]: https://docs.aws.amazon.com/ebs/latest/userguide/nvme-ebs-volumes.html#identify-nvme-ebs-device

kcao-nydig:
> Hey @ikevinc thanks for the kind words! Just a clarification, this does not solve the problem you are pointing out. This PR introduced a udev rule that allows users to identify NVME devices using EBS device mappings (see [this]). The feature that you want is currently being worked in https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/15 and https://github.com/bottlerocket-os/bottlerocket-core-kit/pull/62. With these two changes, you will be able to use ephemeral storage and configure a RAID 0 array, without requiring you to have bootstrap containers.
> 
> 
> 
> [this]: https://docs.aws.amazon.com/ebs/latest/userguide/nvme-ebs-volumes.html#identify-nvme-ebs-device

I see. Thanks for the clarifications. Will those make it into the 1.21.0 release?

PR #3985, opened 2024-05-24
Add Nvidia GPU Time-slicing support

monirul:
**Issue number:**

Closes #

**Description of changes:**
NVIDIA GPUs supports Time Slicing feature which allows user to share a GPU among a larger number of workload by dividing the GPU’s time into slices. Each workload gets a turn to use the GPU resources within its allocated time slice. This is similar to how a CPU might time-slice between different processes, ensuring that the GPU is used efficiently and not sitting idle. 
This PR contains the changes required for bottlerocket to enable Timeslicing for kubernetes.

This PR introduces two bottlerocket settings API:

Bottlerocket Settings | Impact | Value | What it means?
-- | -- | -- | --
`settings.kubernetes.nvidia.device-plugin.max-sharing-per-gpu` | sets the value of the `replicas` settings of the device plugin for the timesliced resources | integer default: `0` | When the value is greater than `0`. the timeslicing will be enabled.
`settings.kubernetes.nvidia.device-plugin.rename-shared-gpu` | sets the value of the `renameByDefault` settings of the device plugin for the timesliced resources | `true` \| `false` default: `false` | When this setting is set to `false`, it does not change the shared gpu's resource name. if set to `true`, it renames the gpus and append `.shared` in the gpu name. for example, if the value is set to `true`, the gpu name of `nvidia.com/gpu` will be changed to `nvidia.com/gpu.shared`

**Testing done:**

```
bash-5.1# apiclient set settings.kubernetes.nvidia.device-plugin.max-sharing-per-gpu=10
[root@admin]# cat .bottlerocket/rootfs/etc/nvidia-k8s-device-plugin/settings.yaml
version: v1
flags:
  migStrategy: "none"
  failOnInitError: true
  plugin:
    passDeviceSpecs: true
    deviceListStrategy: "volume-mounts"
    deviceIDStrategy: "index"
sharing:
  timeSlicing:
    renameByDefault: true
    resources:
    - name: "nvidia.com/gpu"
      replicas: 10

$ kubectl describe node ip-192-168-68-216.us-west-2.compute.internal          
Name:               ip-192-168-68-216.us-west-2.compute.internal
...
Capacity:
  cpu:                    8
  ephemeral-storage:      18366Mi
  hugepages-1Gi:          0
  hugepages-2Mi:          0
  memory:                 32458088Ki
  nvidia.com/gpu.shared:  10
  pods:                   58

```

Note: Migration test is still in progress. I will update once the test is complete.

**Terms of contribution:**

### DIFF
--- a/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
+++ b/packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
@@ -13,6 +13,8 @@ License: Apache-2.0
 URL: https://github.com/NVIDIA/k8s-device-plugin
 Source0: https://%{goimport}/archive/v%{gover}/v%{gover}.tar.gz#/k8s-device-plugin-%{gover}.tar.gz
 Source1: nvidia-k8s-device-plugin.service
+Source2: nvidia-k8s-device-plugin-conf
+Source3: nvidia-k8s-device-plugin-tmpfiles.conf
 
 BuildRequires: %{_cross_os}glibc-devel
 Requires: %{name}(binaries)
@@ -61,11 +63,18 @@ install -p -m 0755 fips/nvidia-device-plugin %{buildroot}%{_cross_fips_bindir}
 
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 %{S:1} %{buildroot}%{_cross_unitdir}
+install -d %{buildroot}%{_cross_tmpfilesdir}
+install -D -p -m 0644 %{S:2} %{buildroot}%{_cross_templatedir}/nvidia-k8s-device-plugin-conf
+install -m 0644 %{S:3} %{buildroot}%{_cross_tmpfilesdir}/nvidia-k8s-device-plugin-tmpfiles.conf
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/nvidia-k8s-device-plugin
 
 %files
 %license LICENSE
 %{_cross_attribution_file}
 %{_cross_unitdir}/nvidia-k8s-device-plugin.service
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/nvidia-k8s-device-plugin
+%{_cross_templatedir}/nvidia-k8s-device-plugin-conf
+%{_cross_tmpfilesdir}/nvidia-k8s-device-plugin-tmpfiles.conf
 
 %files bin
 %{_cross_bindir}/nvidia-device-plugin
--- a/sources/models/shared-defaults/nvidia-k8s-device-plugin.toml
+++ b/sources/models/shared-defaults/nvidia-k8s-device-plugin.toml
@@ -1,0 +1,20 @@
+# nvidia device plugin service
+[services.nvidia-device-plugin]
+restart-commands = ["/bin/systemctl try-reload-or-restart nvidia-k8s-device-plugin.service"]
+configuration-files = ["nvidia-device-plugin"]
+
+[configuration-files.nvidia-device-plugin]
+path = "/etc/nvidia-k8s-device-plugin/settings.yaml"
+template-path = "/usr/share/templates/nvidia-k8s-device-plugin-conf"
+
+[metadata.settings.kubernetes.nvidia.device-plugin]
+affected-services = ["nvidia-device-plugin"]
+
+[settings.kubernetes.nvidia.device-plugin]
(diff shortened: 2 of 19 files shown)


packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
```diff
@@ -61,14 +63,22 @@ install -p -m 0755 fips/nvidia-device-plugin %{buildroot}%{_cross_fips_bindir}
...
 %files
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/nvidia-k8s-device-plugin
 %license LICENSE
 %{_cross_attribution_file}
 %{_cross_unitdir}/nvidia-k8s-device-plugin.service
 
 %files bin
 %{_cross_bindir}/nvidia-device-plugin
+%{_cross_templatedir}/nvidia-k8s-device-plugin-conf
+%{_cross_tmpfilesdir}/nvidia-k8s-device-plugin-tmpfiles.conf
```
arnaldo2792: These should be part of the main package, not the `bin` package. Otherwise, when `fips` is enabled they won't be installed.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
```diff
@@ -61,14 +63,22 @@ install -p -m 0755 fips/nvidia-device-plugin %{buildroot}%{_cross_fips_bindir}
 
 install -d %{buildroot}%{_cross_unitdir}
 install -p -m 0644 %{S:1} %{buildroot}%{_cross_unitdir}
+install -d %{buildroot}%{_cross_tmpfilesdir}
+install -D -p -m 0644 %{S:2} %{buildroot}%{_cross_templatedir}/nvidia-k8s-device-plugin-conf
+install -m 0644 %{S:3} %{buildroot}%{_cross_tmpfilesdir}/nvidia-k8s-device-plugin-tmpfiles.conf
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}
+install -d %{buildroot}%{_cross_factorydir}%{_cross_sysconfdir}/nvidia-k8s-device-plugin
```
arnaldo2792: You can do this in a single line, there are other specs where already do something similar.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
```diff
@@ -61,14 +63,22 @@ install -p -m 0755 fips/nvidia-device-plugin %{buildroot}%{_cross_fips_bindir}
...
 
 %files
+%dir %{_cross_factorydir}%{_cross_sysconfdir}/nvidia-k8s-device-plugin
```
arnaldo2792: This should be after the `%license` and `%{_cross_attribution_file}`

sources/models/shared-defaults/nvidia-k8s-device-plugin.toml
```diff
@@ -0,0 +1,15 @@
...
+configuration-files = ["nvidia-device-plugin"]
+
+[configuration-files.nvidia-device-plugin]
+path = "/etc/nvidia-k8s-device-plugin/settings.yaml"
+template-path = "/usr/share/templates/nvidia-k8s-device-plugin-conf"
+
+[metadata.settings.kubernetes.nvidia.device-plugin]
+affected-services = ["nvidia-device-plugin"]
+
+[settings.kubernetes.nvidia.device-plugin]
```
arnaldo2792, line 13: Does this mean that by default, we don't want to share GPUs? If so, a comment can help.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
monirul: good catch. done.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
monirul: done.

packages/nvidia-k8s-device-plugin/nvidia-k8s-device-plugin.spec
monirul: changed as suggested.

sources/models/shared-defaults/nvidia-k8s-device-plugin.toml
monirul: added comments

PR #3987, opened 2024-05-24
Remove APIClient dependency on the Settings model

### LINKED ISSUE #3949: ootb: apiclient needs to be model agnostic
Migrate apiclient to a server client model, to allow the client to be model agnostic

apiclient needs to no longer depend on the `models` crate. To do this it will rely on responses from apiserver and heuristics on those responses to maintain its current functionality.

sumukhballal:
**Issue number:**

Closes 

**Description of changes:**

We removed the APIClient's dependency on the Settings model. We did this for two dependent subcommands, `set` and `apply`. 

1. For the `apply` command, client-side model verification before serialization was removed. 
2. For the `set` command, for key-value pair inputs a new APIserver endpoint /settings/keypair was introduced. This had its benefits as it allowed the current JSON inputs to work as is, but allowed the transfer of key-pair specific logic from the Client to the Server without any major changes i.e no real change in the data-format for the APIServer to be input (key-value or json) aware.

**Testing done:**

Testing was done `locally` & on an aws-ecs-1 AMI instance. Additionally, Testing between existing AMI and new AMI;s with the code changes was done to prove consistent behavior.

1. Single Set (Map type) works as is.
```
[ssm-user@control]$ apiclient set motd=test
[ssm-user@control]$ apiclient get settings.motd
{
  "settings": {
    "motd": "test"
  }
}
```
4. Multiple Set (Map type) works as is.
```
[ssm-user@control]$ apiclient set motd=random kernel.lockdown=integrity
[ssm-user@control]$ apiclient get settings.kernel.lockdown
{
  "settings": {
    "kernel": {
      "lockdown": "integrity"
    }
  }
}
[ssm-user@control]$ apiclient get settings.motd
{
  "settings": {
    "motd": "random"
  }
}
```
5. Invalid Set (Map type).
```
[ssm-user@control]$  apiclient set motd=random kernel.lockdown=random

Failed to change settings: Failed PATCH request to '/settings/keypair?tx=apiclient-set-tTZRnB3Bi2B8ksZp': Status 400 when PATCHing /settings/keypair?tx=apiclient-set-tTZRnB3Bi2B8ksZp: Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: Error deserializing scalar value: Unable to deserialize into Lockdown: Invalid Linux lockdown mode 'random'
```
6. Passing json and key value pairs should not work.
```
[ssm-user@control]$ apiclient set kernel.lockdown=integrity --json '{"motd": "test"}'

Cannot specify key=value pairs and --json settings with 'set'
```
7. JSON form, with multiple values, internal dots, and quotes/types working as expected:
```
[ssm-user@control]$ apiclient set --json '{"kernel": {"sysctl": {"vm.max_map_count": "888888", "user.max_user_namespaces": "9999"}}}'
[ssm-user@control]$
[ssm-user@control]$ apiclient get settings.kernel
{
  "settings": {
    "kernel": {
      "lockdown": "integrity",
      "sysctl": {
        "user.max_user_namespaces": "9999",
        "vm.max_map_count": "888888"
      }
    }
  }
}
```
8. 'Apply' works as is. 
```
[ssm-user@control]$ apiclient apply
[settings]
motd = "from stdin, TOML"

[ssm-user@control]$ apiclient get settings.motd
{
  "settings": {
    "motd": "from stdin, TOML"
  }
}
```
**Terms of contribution:**

### DIFF
--- a/sources/api/apiclient/src/main.rs
+++ b/sources/api/apiclient/src/main.rs
@@ -6,14 +6,13 @@
 // library calls based on the given flags, etc.)  The library modules contain the code for talking
 // to the API, which is intended to be reusable by other crates.
 
-use apiclient::{apply, exec, get, reboot, report, set, update};
-use datastore::{serialize_scalar, Key, KeyType};
+use apiclient::{apply, exec, get, reboot, report, set, update, SettingsInput};
 use log::{info, log_enabled, trace, warn};
+use serde::{Deserialize, Serialize};
 use simplelog::{
     ColorChoice, ConfigBuilder as LogConfigBuilder, LevelFilter, TermLogger, TerminalMode,
 };
 use snafu::ResultExt;
-use std::collections::HashMap;
 use std::env;
 use std::ffi::OsString;
 use std::process;
@@ -84,10 +83,16 @@ struct RawArgs {
 #[derive(Debug)]
 struct RebootArgs {}
 
+/// Stores a vector of user-supplied key-value pairs for the 'set' subcommand.
+#[derive(Serialize, Deserialize)]
+pub struct SetKeyPairSettings {
+    request_payload: Vec<String>,
+}
+
 /// Stores user-supplied arguments for the 'set' subcommand.
 #[derive(Debug)]
 enum SetArgs {
-    Simple(HashMap<Key, String>),
+    Simple(Vec<String>),
     Json(serde_json::Value),
 }
 
@@ -449,7 +454,7 @@ fn parse_reboot_args(args: Vec<String>) -> Subcommand {
 // what formats to accept.  This code currently makes it as convenient as possible to set settings,
 // by adding/removing a "settings" prefix as necessary.
 fn parse_set_args(args: Vec<String>) -> Subcommand {
-    let mut simple = HashMap::new();
+    let mut simple = Vec::new();
     let mut json = None;
 
     let mut iter = args.into_iter();
@@ -491,22 +496,8 @@ fn parse_set_args(args: Vec<String>) -> Subcommand {
             }
 
             x if x.contains('=') => {
-                let (raw_key, value) = x.split_once('=').unwrap();
-
-                let mut key = Key::new(KeyType::Data, raw_key).unwrap_or_else(|_| {
-                    usage_msg(format!("Given key '{}' is not a valid format", raw_key))
-                });
-
-                // Add "settings" prefix if the user didn't give a known prefix, to ease usage
-                let key_prefix = &key.segments()[0];
-                if key_prefix != "settings" {
-                    let mut segments = key.segments().clone();
-                    segments.insert(0, "settings".to_string());
-                    key = Key::from_segments(KeyType::Data, &segments)
-                        .expect("Adding prefix to key resulted in invalid key?!");
-                }
-
-                simple.insert(key, value.to_string());
+                // Push each key=value pair to vector.
+                simple.push(x.to_string());
             }
 
             x => usage_msg(format!("Unknown argument '{}'", x)),
@@ -698,41 +689,6 @@ async fn check(args: &Args) -> Result<String> {
     Ok(output)
 }
 
-/// We want the key=val form of 'set' to be as simple as possible; we don't want users to have to
-/// annotate or structure their input too much just to tell us the data type, but unfortunately
-/// knowledge of the data type is required to deserialize with the current datastore ser/de code.
-///
-/// To simplify usage, we use some heuristics to determine the type of each input.  We try to parse
-/// each value as a number and boolean, and if those fail, we assume a string.  (API communication
-/// is in JSON form, limiting the set of types; the API doesn't allow arrays or null, and "objects"
-/// (maps) are represented natively through our nested tree-like settings structure.)
-///
-/// If this goes wrong -- for example the user wants a string "42" -- we'll get a deserialization
-/// error, and can print a clear error and request the user use JSON input form to handle
-/// situations with more complex types.
-///
-/// If you have an idea for how to improve deserialization so we don't have to do this, please say!
-fn massage_set_input(input_map: HashMap<Key, String>) -> Result<HashMap<Key, String>> {
-    // Deserialize the given value into the matching Rust type.  When we find a matching type, we
-    // serialize back out to the data store format, which is required to build a Settings object
-    // through the data store deserialization code.
-    let mut massaged_map = HashMap::with_capacity(input_map.len());
-    for (key, in_val) in input_map {
-        let serialized = if let Ok(b) = serde_json::from_str::<bool>(&in_val) {
-            serialize_scalar(&b).context(error::SerializeSnafu)?
-        } else if let Ok(u) = serde_json::from_str::<u64>(&in_val) {
-            serialize_scalar(&u).context(error::SerializeSnafu)?
-        } else if let Ok(f) = serde_json::from_str::<f64>(&in_val) {
-            serialize_scalar(&f).context(error::SerializeSnafu)?
-        } else {
-            // No deserialization, already a string, just serialize
-            serialize_scalar(&in_val).context(error::SerializeSnafu)?
-        };
-        massaged_map.insert(key, serialized);
-    }
-    Ok(massaged_map)
-}
-
 // =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=   =^..^=
 // Main dispatch
 
@@ -803,25 +759,24 @@ async fn run() -> Result<()> {
 
         Subcommand::Set(set) => {
             let settings = match set {
-                SetArgs::Simple(input_map) => {
-                    // For key=val, we need some type information to deserialize into a Settings.
-                    trace!("Original key=value input: {:#?}", input_map);
-                    let massaged_map = massage_set_input(input_map)?;
-                    trace!("Massaged key=value input: {:#?}", massaged_map);
-
-                    // The data store deserialization code understands how to turn the key names
-                    // (a.b.c) and serialized values into the nested Settings structure.
-                    datastore::deserialization::from_map(&massaged_map)
-                        .context(error::DeserializeMapSnafu)?
+                SetArgs::Simple(simple) => {
+                    trace!("User supplied Key Value settings {:#?}", simple);
+                    // Construct the Key Pair struct.
+                    let set_key_pair = SetKeyPairSettings {
+                        request_payload: simple,
+                    };
+                    let settings_string =
+                        serde_json::to_string(&set_key_pair).context(error::SerializeSnafu)?;
+                    SettingsInput::KeyPair(settings_string)
                 }
                 SetArgs::Json(json) => {
-                    // No processing to do on JSON input; the format determines the types.  serde
-                    // can turn a Value into the nested Settings structure itself.
-                    serde_json::from_value(json).context(error::DeserializeJsonSnafu)?
+                    trace!("User supplied Json settings {:#?}", json);
+                    // Convert JSON Value to a string.
+                    SettingsInput::Json(json.to_string())
                 }
             };
 
-            set::set(&args.socket_path, &settings)
+            set::set(&args.socket_path, settings)
                 .await
                 .context(error::SetSnafu)?;
         }
@@ -929,17 +884,6 @@ mod error {
     pub(crate) enum Error {
         #[snafu(display("Failed to apply settings: {}", source))]
         Apply { source: apply::Error },
-
-        #[snafu(display("Unable to deserialize input JSON into model: {}", source))]
-        DeserializeJson { source: serde_json::Error },
-
-        // This is an important error, it's shown when the user uses 'apiclient set' with the
-        // key=value form and we don't have enough data to deserialize the value.  It's not the
-        // user's fault and so we want to be very clear and give an alternative.
-        #[snafu(display("Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: {}", source))]
-        DeserializeMap {
-            source: datastore::deserialization::Error,
-        },
 
         #[snafu(display("Failed to exec: {}", source))]
         Exec { source: exec::Error },
--- a/sources/api/apiclient/src/set.rs
+++ b/sources/api/apiclient/src/set.rs
@@ -1,4 +1,4 @@
-use crate::rando;
+use crate::{rando, SettingsInput};
 use snafu::ResultExt;
 use std::path::Path;
 
@@ -6,7 +6,7 @@ use std::path::Path;
 /// containing those changes.  The given Settings only has to be populated (i.e. Option::Some) with
 /// the settings you want to change.  If you're deserializing a request from a user, for example,
 /// the created Settings will only have the requested keys populated.
-pub async fn set<P>(socket_path: P, settings: &model::Settings) -> Result<()>
+pub async fn set<P>(socket_path: P, settings: SettingsInput) -> Result<()>
 where
     P: AsRef<Path>,
 {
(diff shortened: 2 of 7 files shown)


sumukhballal:
Added a clippy fix!

bcressey:
When we're in "CLI mode" (not used as a library), can we clean up this error?
```
Failed to change settings: Failed PATCH request to '/settings/keypair?tx=apiclient-set-tTZRnB3Bi2B8ksZp': Status 400 when PATCHing /settings/keypair?tx=apiclient-set-tTZRnB3Bi2B8ksZp: Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: Error deserializing scalar value: Unable to deserialize into Lockdown: Invalid Linux lockdown mode 'random'

# can be just this:
Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: Error deserializing scalar value: Unable to deserialize into Lockdown: Invalid Linux lockdown mode 'random'
```

In general, I'd like to keep the user-facing output the same before and after this change. Scripts and what not can take dependencies on specific output messages.

bcressey:
Same for `apiclient apply` - we should get the cleaned-up errors like this:
```
$ apiclient apply <<EOF
> [settings.kernel]
> lockdown = "random"
> EOF
Failed to apply settings: Failed to deserialize settings from '-' into this variant's model: Unable to deserialize into Lockdown: Invalid Linux lockdown mode 'random'
```

sources/api/apiclient/src/main.rs
```diff
@@ -84,10 +83,15 @@ struct RawArgs {
 #[derive(Debug)]
 struct RebootArgs {}
 
+/// Stores a vector of user-supplied key-value pairs for the 'set' subcommand.
+#[derive(Serialize, Deserialize)]
+pub struct SetKeyPairSettings {
+    data: Vec<String>,
```
bcressey: nit: maybe something less generic? (this admittedly isn't much better)
```suggestion
    request_payload: Vec<String>,
```

sources/api/apiclient/src/set.rs
```diff
@@ -31,6 +31,19 @@ where
...
+/// If its of type json, we send it to /settings.
+/// See the API model for information about these endpoints.
+fn get_uri(settings: &str, transaction: &str) -> String {
+    if let Ok(json_value) = serde_json::from_str::<Value>(settings) {
+        if let Some(_value) = json_value.get("data") {
+            return format!("/settings/keypair?tx={}", transaction);
+        }
+    }
+    format!("/settings?tx={}", transaction)
+}
```
bcressey: It would be somewhat more idiomatic Rust to have a `SettingsInput` enum (with an associated `String` value and a `Display` impl) and then use a `match` statement here. But that has the potential to touch a lot of other callers if `apiclient::set` is widely used (is it? might not be).

sources/api/openapi.yaml
```diff
@@ -270,7 +270,30 @@ paths:
           description: "Invalid body"
         500:
           description: "Server error"
-
+  /settings/keypair/:
+      patch:
+      summary: "Update settings which come in key value form"
+      operationId: "set_key_pair_settings"
+      parameters:
```
bcressey: indentation seems off here?

sources/api/apiserver/src/server/mod.rs
```diff
@@ -54,6 +56,45 @@ fn notify_unix_socket_ready() -> Result<()> {
     Ok(())
 }
 
+/// We want the key=val form of 'set' to be as simple as possible; we don't want users to have to
+/// annotate or structure their input too much just to tell us the data type, but unfortunately
+/// knowledge of the data type is required to deserialize with the current datastore ser/de code.
```
bcressey: nit: this maybe belongs at the end of the file or in a different module; it's surprising to find it up here with the server setup and sd notify

sources/api/apiserver/src/server/mod.rs
```diff
@@ -54,6 +56,45 @@ fn notify_unix_socket_ready() -> Result<()> {
...
+///
+/// To simplify usage, we use some heuristics to determine the type of each input.  We try to parse
+/// each value as a number and boolean, and if those fail, we assume a string.  (API communication
+/// is in JSON form, limiting the set of types; the API doesn't allow arrays or null, and "objects"
+/// (maps) are represented natively through our nested tree-like settings structure.)
+///
+/// If this oes wrong -- for example the user wants a string "42" -- we'll get a deserialization
+/// error, and can print a clear error and request the user use JSON input form to handle
+/// situations with more complex types.g
```
bcressey:
```suggestion
/// If this goes wrong -- for example the user wants a string "42" -- we'll get a deserialization
/// error, and can print a clear error and request the user use JSON input form to handle
/// situations with more complex types.
```

sources/api/apiserver/src/server/mod.rs
```diff
@@ -89,7 +130,8 @@ where
             .service(
                 web::scope("/settings")
                     .route("", web::get().to(get_settings))
-                    .route("", web::patch().to(patch_settings)),
+                    .route("", web::patch().to(patch_settings))
+                    .route("/keypair", web::patch().to(patch_settings_key_pair)),
```
bcressey, line 94: Just checking - if we had a model with something like `settings.keypair.foo`, we wouldn't want this:
```
GET /settings/motd # returns value of the "motd" setting
GET /settings/keypair # returns value of the "keypair" setting
PATCH /settings/keypair # server-side route, special cased
PATCH /settings/motd # error
```

From a quick glance at the code, we don't support using the URI path to select a particular subset of settings.

If we did, it might be cleaner to have a top-level `settings-keypair` route instead to avoid ambiguity.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -10,20 +10,22 @@ pub use error::Error;
...
 use nix::unistd::{chown, Gid};
+use serde::{Deserialize, Serialize};
 use snafu::{ensure, OptionExt, ResultExt};
 use std::collections::{HashMap, HashSet};
 use std::env;
 use std::fs::{set_permissions, File, Permissions};
 use std::os::unix::fs::PermissionsExt;
 use std::os::unix::process::ExitStatusExt;
 use std::path::{Path, PathBuf};
+use std::process;
```
bcressey: change the line below to `use std::process::{self, Command};`

sources/api/apiserver/src/server/mod.rs
```diff
@@ -277,6 +319,37 @@ async fn get_settings(
     Ok(SettingsResponse(settings))
 }
 
+/// Prints a more specific message before exiting through usage().
+fn usage_msg<S: AsRef<str>>(msg: S) -> ! {
+    eprintln!("{}\n", msg.as_ref());
+    process::exit(2);
+}
```
bcressey: We don't want to tear down the server in response to a bad client request. This needs to go back to the client as a bad request error.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -277,6 +319,37 @@ async fn get_settings(
...
+
+/// Constructs a Key Pair Map from the
+fn construct_key_pair_map(settings_key_pair_vec: &Vec<String>) -> HashMap<Key, String> {
+    let mut settings_key_pair_map = HashMap::new();
+    for settings_key_pair in settings_key_pair_vec {
+        let (raw_key, value) = settings_key_pair.split_once('=').unwrap();
```
bcressey: can't `unwrap` here (or in any of the transplanted code) or we end up panicking the server for bad input

sources/api/apiserver/src/server/mod.rs
```diff
@@ -277,6 +319,37 @@ async fn get_settings(
...
+            usage_msg(format!("Given key '{}' is not a valid format", raw_key))
+        });
+
+        // Add "settings" prefix if the user didn't give a known prefix, to ease usage
+        let key_prefix = &key.segments()[0];
+        if key_prefix != "settings" {
+            let mut segments = key.segments().clone();
+            segments.insert(0, "settings".to_string());
+            key = Key::from_segments(KeyType::Data, &segments)
+                .expect("Adding prefix to key resulted in invalid key?!");
```
bcressey: need to fix this panic too

sources/api/apiserver/src/server/mod.rs
```diff
@@ -289,6 +362,26 @@ async fn patch_settings(
...
+    data: web::Data<SharedData>,
+) -> Result<HttpResponse> {
+    // Convert to a Map of Key Value pairs.
+    let settings_key_pair_map = construct_key_pair_map(&settings.data);
+    let transaction = transaction_name(&query);
+    let mut datastore = data.ds.write().ok().context(error::DataStoreLockSnafu)?;
+    // We massage the values in the input key pair map.
+    let settings_key_pair_massaged_map = massage_set_input(settings_key_pair_map)?;
+    // The data store deserialization code understands how to turn the key names
+    // (a.b.c) and serialized values into the nested Settings structure.
```
bcressey: We should do this part of input validation before taking the datastore lock.

sources/api/apiserver/src/server/mod.rs
sumukhballal: Yes that is true. 

I used keypair as a path after checking the current GET for specific settings is part of the query: `/settings?keys=foo,bar,baz`  .  I preferred to just add a Path `/keypair` to `/settings` since it falls under the settings umbrella while still being specific enough only for "keypair" type inputs. 

I do like the approach of using a top level `/settings-keypair`. But since we wouldnt really support GET's from this endpoint, I preferred to keep the more specific `keypair` under the `/settings` umbrella.

That way 

```
GET /settings.    # Get Settings in JSON format
PATCH /settings # Only JSON inputs of Model type
PATCH /setting/keypair # Only JSON inputs of SetKeyPair type
GET /settings/{SETTINGS_TYPES} # No support
```

But yeah in the future, if we want to use Keys as queries, we should update the `settings/keypair` to be its standalone `/settings-keypair` type as its would cause confusion.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -54,6 +56,45 @@ fn notify_unix_socket_ready() -> Result<()> {
...
+/// We want the key=val form of 'set' to be as simple as possible; we don't want users to have to
+/// annotate or structure their input too much just to tell us the data type, but unfortunately
+/// knowledge of the data type is required to deserialize with the current datastore ser/de code.
+///
+/// To simplify usage, we use some heuristics to determine the type of each input.  We try to parse
+/// each value as a number and boolean, and if those fail, we assume a string.  (API communication
+/// is in JSON form, limiting the set of types; the API doesn't allow arrays or null, and "objects"
+/// (maps) are represented natively through our nested tree-like settings structure.)
+///
+/// If this oes wrong -- for example the user wants a string "42" -- we'll get a deserialization
```
mgsharm: Typo `oes` -> `goes`.

sources/api/apiserver/src/server/mod.rs
```diff
@@ -277,6 +319,37 @@ async fn get_settings(
     Ok(SettingsResponse(settings))
 }
 
+/// Prints a more specific message before exiting through usage().
+fn usage_msg<S: AsRef<str>>(msg: S) -> ! {
+    eprintln!("{}\n", msg.as_ref());
+    process::exit(2);
+}
+
+/// Constructs a Key Pair Map from the
```
mgsharm: This comment seems incomplete.

mgsharm: Also, do we even need this comment ? The function name is descriptive enough IMO.

sumukhballal:
Added changes to address @bcressey & @mgsharm's comments.

sources/api/apiclient/src/main.rs
```diff
@@ -84,10 +83,15 @@ struct RawArgs {
 #[derive(Debug)]
 struct RebootArgs {}
 
+/// Stores a vector of user-supplied key-value pairs for the 'set' subcommand.
+#[derive(Serialize, Deserialize)]
+pub struct SetKeyPairSettings {
+    request_payload: Vec<String>,
+}
 /// Stores user-supplied arguments for the 'set' subcommand.
```
webern: nit

```suggestion
}

/// Stores user-supplied arguments for the 'set' subcommand.
```

sources/api/openapi.yaml
```diff
@@ -270,7 +270,30 @@ paths:
...
+          description: "Transaction in which to update settings; defaults to user 'default' transaction"
+          schema:
+            type: string
+          required: false
+      requestBody:
+        required: true
+        content:
+          text/plain:
+            schema:
+              type: string
```
webern: It this right? Isn't it a JSON of key/value pairs?

sources/api/apiserver/src/server/mod.rs
```diff
@@ -712,6 +808,11 @@ impl ResponseError for error::Error {
     }
 }
 
+/// A struct used by actix for the 'set' subcommand when user-supplied values are key-pairs.
+#[derive(Serialize, Deserialize)]
+struct SetKeyPairSettings {
+    request_payload: Vec<String>,
+}
 /// SharedData is responsible for any data needed by web handlers that isn't provided by the client
```
webern: nit

```suggestion
}

/// SharedData is responsible for any data needed by web handlers that isn't provided by the client
```

sources/api/apiserver/src/server/mod.rs
webern: Are we assuming then that an OOTB user would never create a setting named `keypair`? i.e. if I made `settings.keypair` a part of my variant's schema, this would break?

sources/api/apiserver/src/server/mod.rs
```diff
@@ -646,6 +667,76 @@ fn transaction_name(query: &web::Query<HashMap<String, String>>) -> &str {
...
+/// If you have an idea for how to improve deserialization so we don't have to do this, please say!
+fn massage_set_input(input_map: HashMap<Key, String>) -> Result<HashMap<Key, String>> {
+    // Deserialize the given value into the matching Rust type.  When we find a matching type, we
+    // serialize back out to the data store format, which is required to build a Settings object
+    // through the data store deserialization code.
+    let mut massaged_map = HashMap::with_capacity(input_map.len());
+    for (key, in_val) in input_map {
+        let serialized = if let Ok(b) = serde_json::from_str::<bool>(&in_val) {
+            serialize_scalar(&b).context(error::SerializeSnafu)?
+        } else if let Ok(u) = serde_json::from_str::<u64>(&in_val) {
```
webern, line 723: Do we have negative numbers anywhere in settings?

Edit: Oh, I guess this code was moved as-is, so I guess not!?

```suggestion
        } else if let Ok(u) = serde_json::from_str::<i64>(&in_val) {
```

sources/api/apiserver/src/server/mod.rs
sumukhballal: If a user creates a settings with the name `keypair`, it would work as is. For example, 

```
apiclient set keypair=test # works
```
because it would call the /settings/keypair endpoint but with this payload

```
{"request_payload": "{\"keypair\": \"test\"}"}
```

So this should work as is. Ben was hinting at the fact that if we ever introduce a /get with a path based structure like GET /settings/{SETTINGS_NAME} like /settings/motd or in this case, /settings/keypair , it would lead to confusion. (It would still work, since the type is PATCH)

sources/api/apiserver/src/server/mod.rs
sumukhballal: Yes, copied as is.

sources/api/openapi.yaml
sumukhballal: Earlier we would deserialize directly into Settings Model and just serialize that to send over the network. Now we changed the Model to be a simple json string for /settings endpoint and a 

```
{
"request_payload": Vec<String>
}
```
json string for the /settings/keypair endpoint.

But youre right, I should update this with a new Schema and it should be JSON.

bcressey (approved):
LGTM.

In order to remove the actual model dependency from `Cargo.toml`, it looks like we need `model::exec` to move somewhere else. Are you planning to tackle that next?

It's not strictly required here but it would be easy to make the mistake of using more of the model than that, including `Settings` which would be broken.

sumukhballal:
> LGTM.
> 
> In order to remove the actual model dependency from `Cargo.toml`, it looks like we need `model::exec` to move somewhere else. Are you planning to tackle that next?
> 
> It's not strictly required here but it would be easy to make the mistake of using more of the model than that, including `Settings` which would be broken.

@bcressey  Yeah, Ill add that as part of another PR. I was wondering if it would be a good idea to add those types to the 'modeled-types' package & update any of its references. Like here: https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/modeled-types/src/shared.rs

or just add them into APIClient?

bcressey:
> I was wondering if it would be a good idea to add those types to the 'modeled-types' package & update any of its references.

Maybe a shared api-types crate that is just for apiserver & apiclient alignment? modeled-types is somewhat bound up in settings concerns currently, and I foresee it moving to the settings SDK for ease of out-of-tree use.

sumukhballal:
> > I was wondering if it would be a good idea to add those types to the 'modeled-types' package & update any of its references.
> 
> Maybe a shared api-types crate that is just for apiserver & apiclient alignment? modeled-types is somewhat bound up in settings concerns currently, and I foresee it moving to the settings SDK for ease of out-of-tree use.

Sounds good. Will raise a PR shortly!

webern:
> > LGTM.
> > In order to remove the actual model dependency from `Cargo.toml`, it looks like we need `model::exec` to move somewhere else. Are you planning to tackle that next?
> > It's not strictly required here but it would be easy to make the mistake of using more of the model than that, including `Settings` which would be broken.
> 
> @bcressey Yeah, Ill add that as part of another PR. I was wondering if it would be a good idea to add those types to the 'modeled-types' package & update any of its references. Like here: https://github.com/bottlerocket-os/bottlerocket/blob/develop/sources/models/modeled-types/src/shared.rs
> 
> or just add them into APIClient?

So, #3949 isn't actually closed until that subsequent PR merges...

webern (approved):

PR #3993, opened 2024-05-28
README: Add k8s cmd to retrieve log archive

### LINKED ISSUE #3973: Collecting logs from EKS Worker Nodes running Bottlerocket AMI when no SSH is enabled
The documentation on the link says that we can use [SSH over SSM] when no SSH is enabled:
https://github.com/bottlerocket-os/bottlerocket/blob/1353ea1c35a14c3c9f4d374c6781059ea677e2ef/README.md?plain=1#L483

however I tried following those steps and this does not seem to work on Bottlerocket
when trying to edit the ssh config file it says:
```
"~/.ssh/config" E212: Can't open file for writing 
```
even from inside the admin container.

I was able to find a solution which works, only SSM access is required:

Steps are:

```
1. SSM to node
2. enter-admin-container
3. sudo sheltie
4. logdog
```

logs will be generated at: /var/log/support/bottlerocket-logs.tar.gz 

then we can run

`5. kubectl get --raw "/api/v1/nodes/<node>/proxy/logs/support/bottlerocket-logs.tar.gz" > bottlerocket-logs.tar.gz `

and it will transfer the bottlerocket-logs.tar.gz to the local machine.

If  Bottlerocket team believes this is an approved solution and the README needs to be updated with above solution feel free to do so or let me know, and I can create a PR.
Thanks.

koooosh:
**Issue number:**

Closes #3973

**Description of changes:**

Customer @maiconrocha found an alternative command to retrieve the log archive without using SSH for `aws-k8s` variants.

Updating the README with this command.

**Testing done:**

AMI: `bottlerocket-aws-k8s-1.24-x86_64-v1.20.0-fcf71a47`

*Connected to instance via SSM*

On instance:
1. `enter-admin-container`
2. `sudo sheltie`
3. `logdog`

On local machine:
1. `kubectl get --raw "/api/v1/nodes/NODE_NAME/proxy/logs/support/bottlerocket-logs.tar.gz" > bottlerocket-logs.tar.gz`
2. Verified that log archive was fetched

**Terms of contribution:**

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,6 +462,11 @@ See the [`settings.aws.*` reference](https://bottlerocket.dev/en/os/latest/#/api
 ### Logs
 
 You can use `logdog` through the [admin container](#admin-container) to obtain an archive of log files from your Bottlerocket host.
+
+For a list of what is collected, see the logdog [command list](sources/logdog/src/log_request.rs).
+
+#### Generating logs
+
 SSH to the Bottlerocket host or `apiclient exec admin bash` to access the admin container, then run:
 
 ```shell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -480,6 +480,13 @@ ssh -i YOUR_KEY_FILE \
   "cat /.bottlerocket/support/bottlerocket-logs.tar.gz" > bottlerocket-logs.tar.gz
 ```
 
+Or, if you are running Kubernetes, you can retrieve the file wihout using SSH like:
```
yeazelm: Can you break this into sections for generating the logs and fetching the logs. It would be helpful to show that generating the logs is always consistent, but one can fetch the logs in at least 3 different ways which are enumerated here:
* Via SSH if already enabled
* Using `kubectl get`
* Using SSH over SSM 
This will help users understand that they don't need to do all these steps, they can choose the one that works for them.

ginglis13 (approved):

yeazelm (approved):

PR #3994, opened 2024-05-28
Nvidia settings API for container runtime

monirul:
**Issue number:**

Closes #

**Description of changes:**
This PR will expose two new APIs that will allow customer to configure value of `accept-nvidia-visible-devices-as-volume-mounts` and `accept-nvidia-visible-devices-envvar-when-unprivileged` for nvidia container runtime.

Bottlerocket Settings | Impact | Value | What it means?
-- | -- | -- | --
`settings.kubernetes.nvidia.container-runtime.visible-devices-as-volume-mounts` | allows to change the  `accept-nvidia-visible-devices-as-volume-mounts` value for k8s container-toolkit | `true` \| `false` default: `true` | Adjusting the `visible-devices-as-volume-mounts` settings will alters the method of GPU detection and integration within container environments. Setting this parameter to `true` enables the NVIDIA runtime to recognize GPU devices listed in the `NVIDIA_VISIBLE_DEVICES` environment variable and mount them as volumes, which permits applications within the container to interact with and leverage the GPUs as if they were local resources.
`settings.kubernetes.nvidia.container-runtime.visible-devices-envvar-when-unprivileged` | allows to set value of `accept-nvidia-visible-devices-envvar-when-unprivileged` settings of nvidia container runtime for k8s varient | `true` \| `false` default: `false` | When this setting is set to `false`, it prevents unprivileged containers from accessing all GPU devices on the host by default. If `NVIDIA_VISIBLE_DEVICES` is set to all within the container images and `visible-devices-envvar-when-unprivileged` is set to true, all GPUs on the host will be accessible to the containers, regardless of the limits set via nvidia.com/gpu. This could lead to situations where more GPUs are allocated to a pod than intended, which can affect resource scheduling and isolation.

**Testing done:**
Yes. 
```
$ apiclient set settings.kubernetes.nvidia.container-runtime.visible-devices-as-volume-mounts = true
$ apiclient set settings.kubernetes.nvidia.container-runtime.visible-devices-envvar-when-unprivileged = false
$ apiclient get settings.kubernetes.nvidia.container-runtime
{
  "settings": {
    "kubernetes": {
      "nvidia": {
        "container-runtime": {
          "visible-devices-as-volume-mounts": true,
          "visible-devices-envvar-when-unprivileged": false
        }
      }
    }
  }
}

$ cat /etc/nvidia-container-runtime/config.toml
accept-nvidia-visible-devices-as-volume-mounts = true
accept-nvidia-visible-devices-envvar-when-unprivileged = false

[nvidia-container-cli]
root = "/"
path = "/usr/bin/nvidia-container-cli"
environment = []
ldconfig = "@/sbin/ldconfig"
```
**Terms of contribution:**

### DIFF
--- a/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
+++ b/packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
@@ -1,0 +1,12 @@
+[required-extensions]
+kubernetes = "v1"
+
++++
+accept-nvidia-visible-devices-as-volume-mounts = {{settings.kubernetes.nvidia.container-runtime.visible-devices-as-volume-mounts}}
+accept-nvidia-visible-devices-envvar-when-unprivileged = {{settings.kubernetes.nvidia.container-runtime.visible-devices-envvar-when-unprivileged}}
+
+[nvidia-container-cli]
+root = "/"
+path = "/usr/bin/nvidia-container-cli"
+environment = []
+ldconfig = "@/sbin/ldconfig"
--- a/sources/models/shared-defaults/nvidia-k8s-container-toolkit.toml
+++ b/sources/models/shared-defaults/nvidia-k8s-container-toolkit.toml
@@ -1,0 +1,14 @@
+[settings.kubernetes.nvidia.container-runtime]
+visible-devices-as-volume-mounts = false
+visible-devices-envvar-when-unprivileged = true
+
+[metadata.settings.kubernetes.nvidia.container-runtime]
+affected-services = ["nvidia-container-toolkit"]
+
+[services.nvidia-container-toolkit]
+configuration-files = ["nvidia-container-toolkit"]
+restart-commands = []
+
+[configuration-files.nvidia-container-toolkit]
+path = "/etc/nvidia-container-runtime/config.toml"
+template-path = "/usr/share/factory/nvidia-container-runtime/nvidia-container-toolkit-config-k8s"
--- a/sources/models/src/aws-k8s-1.30-nvidia/defaults.d/81-nvidia-k8s-container-toolkit.toml
+++ b/sources/models/src/aws-k8s-1.30-nvidia/defaults.d/81-nvidia-k8s-container-toolkit.toml
@@ -1,0 +1,1 @@
+../../../shared-defaults/nvidia-k8s-container-toolkit.toml
--- a/Release.toml
+++ b/Release.toml
@@ -314,4 +314,5 @@ version = "1.21.0"
 ]
 "(1.20.0, 1.21.0)" = [
     "migrate_v1.21.0_pluto-remove-generators-v0-1-0.lz4",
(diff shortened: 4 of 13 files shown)


sources/models/shared-defaults/nvidia-k8s-container-toolkit.toml
```diff
@@ -0,0 +1,14 @@
+[settings.kubernetes.nvidia.container-runtime]
+visible-devices-as-volume-mounts = false
+visible-devices-envvar-when-unprivileged = true
+
+[metadata.settings.kubernetes.nvidia.container-runtime]
```
arnaldo2792, line 5: This PR is missing a migration to remove the affected services on a downgrade.

packages/nvidia-container-toolkit/nvidia-container-toolkit-config-k8s
```diff
@@ -0,0 +1,12 @@
+[required-extensions]
+kubernetes = "v1"
+
++++
+accept-nvidia-visible-devices-as-volume-mounts = {{settings.kubernetes.nvidia.container-runtime.visible-devices-as-volume-mounts}}
+accept-nvidia-visible-devices-envvar-when-unprivileged = {{settings.kubernetes.nvidia.container-runtime.visible-devices-envvar-when-unprivileged}}
```
arnaldo2792, line 6: Let's be safe, and use the `{{default}}` helper, otherwise if `settings.kubernetes.nvidia.container-runtime.visible-devices-as-volume-mounts` isn't present, the render will fail.

sources/models/shared-defaults/nvidia-k8s-container-toolkit.toml
```diff
@@ -0,0 +1,14 @@
+[settings.kubernetes.nvidia.container-runtime]
+visible-devices-as-volume-mounts = false
+visible-devices-envvar-when-unprivileged = true
```
arnaldo2792, line 3: Lets use the default values to prevent unprivileged pods from accessing all the devices:

```toml
accept-nvidia-visible-devices-envvar-when-unprivileged = false
```

sources/models/src/aws-k8s-1.30-nvidia/defaults.d/81-nvidia-k8s-container-toolkit.toml
```diff
@@ -0,0 +1 @@
+../../../shared-defaults/nvidia-k8s-container-toolkit.toml
```
arnaldo2792, line 1: The PR is missing symlinks for other variants, we need symlinks for:

- aws-k8s-1.24-nvidia
- aws-k8s-1.25-nvidia
- aws-k8s-1.26-nvidia

arnaldo2792 (changes requested):

ytsssun:
Opened https://github.com/bottlerocket-os/bottlerocket/pull/4052 instead.

PR #3981, opened 2024-05-23
Build a statically linked version of kmod

vigh-m:
**Issue number:**

Updates #3968

**Description of changes:**
This change builds a statically linked version of `kmod`. ~This new `kmod` is unlinked from the existing `/bin/kmod` and its symlinks. It is installed in `/usr/libexec/kmod`~

This new version of `kmod` replaces the existing one. `.so` files are still provided for dependencies

By providing a statically linked kmod, containers can mount it and load kernel modules without compatibility issues. This is a first step to unblocking #3968

**Testing done:**

_Tested on an admin container_

- [x] Validating the linking status
```
# ldd /.bottlerocket/rootfs/usr/bin/kmod
	not a dynamic executable

# readelf --program-headers --wide /.bottlerocket/rootfs/usr/bin/kmod

Elf file type is EXEC (Executable file)
Entry point 0x401cf0
There are 10 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x000000 0x0000000000400000 0x0000000000400000 0x0005e0 0x0005e0 R   0x1000
  LOAD           0x001000 0x0000000000401000 0x0000000000401000 0x10e472 0x10e472 R E 0x1000
  LOAD           0x110000 0x0000000000510000 0x0000000000510000 0x043b13 0x043b13 R   0x1000
  LOAD           0x154a60 0x0000000000555a60 0x0000000000555a60 0x007110 0x00ceb0 RW  0x1000
  NOTE           0x000270 0x0000000000400270 0x0000000000400270 0x000040 0x000040 R   0x8
  NOTE           0x0002b0 0x00000000004002b0 0x00000000004002b0 0x000044 0x000044 R   0x4
  TLS            0x154a60 0x0000000000555a60 0x0000000000555a60 0x000020 0x000060 R   0x8
  LOOS+0x474e553 0x000270 0x0000000000400270 0x0000000000400270 0x000040 0x000040 R   0x8
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x10
  GNU_RELRO      0x154a60 0x0000000000555a60 0x0000000000555a60 0x0055a0 0x0055a0 R   0x1

 Section to Segment mapping:
  Segment Sections...
   00     .note.gnu.property .note.gnu.build-id .note.ABI-tag .rela.plt
   01     .init .plt .text .fini
[...]

PR #4007, opened 2024-05-30
Mount static kmod as /sbin/modprobe

### LINKED ISSUE #3968: Changes to kernel module compression can break certain workflows
**Image I'm using:**
All Bottlerocket OS 1.20.0 variants

Updating to Bottlerocket v1.20.0 should not have broken any existing container workflows.

**What actually happened:**
Launching Bottlerocket v1.20.0 can result in unexpected errors like:
```
level=warning msg="iptables modules could not be initialized. It probably means that iptables is not available on this system" error="could not load module iptable_raw: exit status
```
or
```
Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process:
```

**Root Cause:**
Following [changes](https://github.com/bottlerocket-os/bottlerocket/pull/3861) in Bottlerocket v1.20.0 we dropped support for `xz` compressed modules in favour of `gz` compression. We're seeing failures to launch certain containers which require kernel modules and have `modprobe` built without `libz (gzip)` support. 

Known Affected systems:
- Cilium

**Workaround:**
A workaround would be to use user data on new launches or `apiclient` in running hosts to setup `kernel.modules.<name>.autoload` to load the required modules as described [here](https://bottlerocket.dev/en/os/1.19.x/api/settings/kernel/#modules_autoload).

In the case of Cilium the commands would like the following when using the apiclient:
```
apiclient set \
  settings.kernel.modules.ip_tables.allowed=true \
  settings.kernel.modules.ip_tables.autoload=true \
  settings.kernel.modules.iptable_nat.allowed=true \
  settings.kernel.modules.iptable_nat.autoload=true \
  settings.kernel.modules.iptable_mangle.allowed=true \
  settings.kernel.modules.iptable_mangle.autoload=true \
  settings.kernel.modules.iptable_raw.allowed=true \
  settings.kernel.modules.iptable_raw.autoload=true \
  settings.kernel.modules.iptable_filter.allowed=true \
  settings.kernel.modules.iptable_filter.autoload=true \
  settings.kernel.modules.ip6table_mangle.allowed=true \
  settings.kernel.modules.ip6table_mangle.autoload=true \
  settings.kernel.modules.ip6table_raw.allowed=true \
  settings.kernel.modules.ip6table_raw.autoload=true \
  settings.kernel.modules.ip6table_filter.allowed=true \
  settings.kernel.modules.ip6table_filter.autoload=true
```
OR via `user-data.toml`:
```
[settings.kernel.modules.ip_tables]
allowed = true
autoload = true

[settings.kernel.modules.iptable_nat]
allowed = true
autoload = true

[settings.kernel.modules.iptable_mangle]
allowed = true
autoload = true

[settings.kernel.modules.iptable_raw]
allowed = true
autoload = true

[settings.kernel.modules.iptable_filter]
allowed = true
autoload = true

[settings.kernel.modules.ip6table_mangle]
allowed = true
autoload = true

[settings.kernel.modules.ip6table_raw]
allowed = true
autoload = true

[settings.kernel.modules.ip6table_filter]
allowed = true
autoload = true
```

vigh-m:
**Issue number:**

Closes #3968 

**Description of changes:**
This change mounts the static kmod built in #3981 into a customer container. 

**Testing done:**
Tested on the host-ctr:
```
[ssm-user@control]$ enter-admin-container
Confirming admin container is enabled...
Waiting for admin container to start........
Entering admin container
          Welcome to Bottlerocket's admin container!
    ╱╲
   ╱┄┄╲   This container provides access to the Bottlerocket host
   │▗▖│   filesystems (see /.bottlerocket/rootfs) and contains common
  ╱│  │╲  tools for inspection and troubleshooting.  It is based on
  │╰╮╭╯│  Amazon Linux 2, and most things are in the same places you
    ╹╹    would find them on an AL2 host.

To permit more intrusive troubleshooting, including actions that mutate the
running state of the Bottlerocket host, we provide a tool called "sheltie"
(`sudo sheltie`).  When run, this tool drops you into a root shell in the
Bottlerocket host's root filesystem.
[root@admin]# ldd /usr/sbin/modprobe
  not a dynamic executable
```

Tested on a kubernetes container:
```
root@my-pod:/# ldd /usr/sbin/modprobe
  not a dynamic executable
root@my-pod:/# ldd /sbin/modprobe
  not a dynamic executable
root@my-pod:/# ldd /usr/bin/kmod
ldd: /usr/bin/kmod: No such file or directory
```

Pod Spec used:
```
apiVersion: v1
kind: Pod
metadata:
  name: my-pod
spec:
  containers:
  - name: my-container
    image: ubuntu:noble
    command: ["/bin/sh", "-c", "while true; do echo 'Running...'; sleep 5; done"]
    securityContext:
      capabilities:
        add: ["SYS_MODULE"]
    volumeMounts:
    - name: kernel-modules
      mountPath: /lib/modules/
      readOnly: true
  volumes:
  - name: kernel-modules
    hostPath:
      path: /lib/modules/
```

**Terms of contribution:**

### DIFF
--- a/packages/containerd/containerd-cri-base-json
+++ b/packages/containerd/containerd-cri-base-json
@@ -102,6 +102,15 @@ oci-defaults = { version = "v1", helpers = ["oci_defaults"] }
                 "mode=755",
                 "size=65536k"
             ]
+        },
+        {
+            "destination": "/sbin/modprobe",
+            "source": "/usr/bin/kmod",
+            "options": [
+                "exec",
+		"bind",
+                "ro"
+            ]
         }
     ],
     "linux": {
--- a/sources/host-ctr/cmd/host-ctr/main.go
+++ b/sources/host-ctr/cmd/host-ctr/main.go
@@ -896,6 +896,12 @@ func withPrivilegedMounts() oci.SpecOpts {
 			Source:      "/mnt",
 			Type:        "bind",
 		},
+		{
+			Options:     []string{"bind", "ro"},
+			Destination: "/sbin/modprobe",
+			Source:      "/usr/bin/kmod",
+			Type:        "bind",
+		},
 	})
 }
 
--- a/packages/docker-engine/0002-oci-inject-kmod-in-all-containers.patch
+++ b/packages/docker-engine/0002-oci-inject-kmod-in-all-containers.patch
@@ -1,0 +1,31 @@
+From e35f5eeeaa4c7b9ec1ae0720fc7de0fc4d43e02f Mon Sep 17 00:00:00 2001
+From: Arnaldo Garcia Rincon <agarrcia@amazon.com>
+Date: Thu, 30 May 2024 14:38:33 +0000
+Subject: [PATCH] oci: inject kmod in all containers
+
+Append a new mount to the default spec created for Linux containers
+
+Signed-off-by: Arnaldo Garcia Rincon <agarrcia@amazon.com>
+---
+ oci/defaults.go | 6 ++++++
+ 1 file changed, 6 insertions(+)
+
+diff --git a/oci/defaults.go b/oci/defaults.go
+index c3dae8b..2e90cfa 100644
+--- a/oci/defaults.go
++++ b/oci/defaults.go
+@@ -100,6 +100,12 @@ func DefaultLinuxSpec() specs.Spec {
+ 				Source:      "shm",
+ 				Options:     []string{"nosuid", "noexec", "nodev", "mode=1777"},
+ 			},
++			{
++				Destination: "/sbin/modprobe",
++				Type:        "bind",
++				Source:      "/usr/bin/kmod",
++				Options:     []string{"exec", "bind", "ro"},
++			},
+ 		},
+ 		Linux: &specs.Linux{
+ 			MaskedPaths: []string{
+-- 
+2.44.0
--- a/packages/docker-engine/docker-engine.spec
+++ b/packages/docker-engine/docker-engine.spec
@@ -30,6 +30,7 @@ Source100: prepare-var-lib-docker.service
 Source1000: clarify.toml
 
 Patch0001: 0001-Change-default-capabilities-using-daemon-config.patch
+Patch0002: 0002-oci-inject-kmod-in-all-containers.patch
 
 BuildRequires: git
 BuildRequires: %{_cross_os}glibc-devel


bcressey:
Code changes LGTM, though I would split them into one for host-ctr and one for containerd.

Does it work for Cilium? For the example pod spec, can you load a module via `modprobe`?

yeazelm:
The code looks good, it would be awesome to include a before and after on that pod test to show the change, even better showing it loading a module after the fix to prove this fixes the issue.

vigh-m:
Yeah. It loads the modules like expected. 
```
root@my-pod:/# ln -s /usr/sbin/modprobe /usr/bin/lsmod
root@my-pod:/# lsmod | grep tables
nf_tables             307200  0
nfnetlink              20480  2 nf_conntrack_netlink,nf_tables
root@my-pod:/# modprobe -r nf_tables
root@my-pod:/# lsmod | grep tables
root@my-pod:/# modprobe nf_tables
root@my-pod:/# lsmod | grep tables
nf_tables             307200  0
nfnetlink              20480  2 nf_conntrack_netlink,nf_tables
```

I had validated but didn't add it to the initial notes

vigh-m:
Validated via SHA512 that it's the same modprobe

- On admin container
```
[root@admin]# sha512sum /usr/bin/kmod
b2f5451a89fc2e21a971b23086380ea3c8916b5ad753c31812b4419992e6f63687da3b35723ec1d1545039c4e946e54a01ff2555420966bcceda6bb8d0569404  /usr/bin/kmod
```
- On rootfs
```
[root@admin]# sha512sum /.bottlerocket/rootfs/bin/kmod
b2f5451a89fc2e21a971b23086380ea3c8916b5ad753c31812b4419992e6f63687da3b35723ec1d1545039c4e946e54a01ff2555420966bcceda6bb8d0569404  /.bottlerocket/rootfs/bin/kmod
```
- On container
```
root@my-pod:/# sha512sum /usr/sbin/modprobe
b2f5451a89fc2e21a971b23086380ea3c8916b5ad753c31812b4419992e6f63687da3b35723ec1d1545039c4e946e54a01ff2555420966bcceda6bb8d0569404  /usr/sbin/modprobe
```

yeazelm (approved):

arnaldo2792:
For the validation, you could just have checked that the `modprobe` binary is coming from the host via `/proc/self/mountinfo`.

sources/host-ctr/cmd/host-ctr/main.go
```diff
@@ -896,6 +896,12 @@ func withPrivilegedMounts() oci.SpecOpts {
 			Source:      "/mnt",
 			Type:        "bind",
 		},
+		{
+			Options:     []string{"rbind", "ro"},
```
arnaldo2792: To be consistent with the other mount, this should be `bind` not `rbind`

packages/containerd/containerd-cri-base-json
```diff
@@ -102,6 +102,15 @@ oci-defaults = { version = "v1", helpers = ["oci_defaults"] }
                 "mode=755",
                 "size=65536k"
             ]
+        },
+        {
+            "destination": "/sbin/modprobe",
```
arnaldo2792, line 107: One risk of this implementation is that if `PATH` is `/usr/bin:/sbin`, the `modprobe` binary in the container will be used instead of the injected `modprobe`. I guess that's a calculated risk to take, since we can't inject `modprobe` in every single path in `PATH`. Probably something to follow up for a later release is checking `PATH` and inject `modprobe` to the first path in that list.

bcressey (approved):
It would be good to validate that Cilium specifically works as expected with this change, in case there are more paths we need to bind to.

arnaldo2792 (approved):

vigh-m:
New changes include:
- Updates to docker to inject kmod by default
- Fix based on comments
- Typo fixes

vigh-m:
Fixing typo in the commit message

bcressey (approved):

yeazelm (approved):

vigh-m:
Validated by running the `cilium/cilium:v1.15.5` image from DockerHub:

Locally launched container:
```
$ docker run -it --rm cilium/cilium:v1.15.5 sh
# ldd /usr/sbin/modprobe
	linux-vdso.so.1 (0x00007fffc838c000)
	libzstd.so.1 => /lib/x86_64-linux-gnu/libzstd.so.1 (0x00007f6632d16000)
	liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x00007f6632ceb000)
	libcrypto.so.3 => /lib/x86_64-linux-gnu/libcrypto.so.3 (0x00007f66328a7000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f663267e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f6632e14000)
# ls -la /usr/sbin/modprobe
lrwxrwxrwx. 1 root root 9 Aug 17  2021 /usr/sbin/modprobe -> /bin/kmod
# cat /proc/self/mountinfo  | grep kmod
#
```

VS container launched on bottlerocket
```
$ kubectl exec -it my-pod -- bash
root@my-pod:/home/cilium#
root@my-pod:/home/cilium# ldd /usr/sbin/modprobe
	not a dynamic executable
root@my-pod:/home/cilium# ls -la /usr/sbin/modprobe
lrwxrwxrwx. 1 root root 9 Aug 17  2021 /usr/sbin/modprobe -> /bin/kmod
root@my-pod:/home/cilium# ldd /bin/kmod
	not a dynamic executable
root@my-pod:/home/cilium# cat /proc/self/mountinfo | grep kmod
1281 1275 252:0 /x86_64-bottlerocket-linux-gnu/sys-root/usr/bin/kmod /usr/bin/kmod ro,relatime master:1 - ext4 /dev/root ro,seclabel,stripe=1024
```

I'm able to load and unload modules
```
root@my-pod:/home/cilium# lsmod | grep tables
nf_tables             307200  0
nfnetlink              20480  2 nf_conntrack_netlink,nf_tables
root@my-pod:/home/cilium# modprobe -r nf_tables
root@my-pod:/home/cilium# lsmod | grep tables
root@my-pod:/home/cilium# modprobe nf_tables
root@my-pod:/home/cilium# lsmod | grep tables
nf_tables             307200  0
nfnetlink              20480  2 nf_conntrack_netlink,nf_tables
```

PR #4014, opened 2024-05-31
migrations: archive previous releases

bcressey:
**Issue number:**
Previously: #3633, #2800, #2606, ...

**Description of changes:**
This avoids rebuilding existing migrations and reduces build times.

**Testing done:**
Builds work.

**Terms of contribution:**


webern (commented):
Are there consequences to upgrade/downgrade workflows of archiving, e.g. 1.20 so soon?

bcressey:
> Are there consequences to upgrade/downgrade workflows of archiving, e.g. 1.20 so soon?

Not necessarily but it depends on how the upgrade/downgrade workflow is handled.

As I understand it, the problem is with a sequence like this:
```
build old version
build new version
sync repo
```

The old version will create a migrations RPM during `os` build with all the older migration binaries. However, if that's not synced up to the repo, then when the new version is built, a new migrations RPM will be created with only the subset of current migration binaries.

I expect a sequence like this would be fine:
```
build old version
sync repo
build new version
sync repo
```

webern (approved):

arnaldo2792 (approved):

cbgbt (approved):

PR #4023, opened 2024-06-03
Revert "docker-engine: inject kmod in all containers"

yeazelm:
**Description of changes:**
In testing ECS variants that use Docker, we found issues with use cases that have volume mounts. This seems to cause failures for the containers to launch. Reverting this for now until we can find a better solution that doesn't break these workflows. 

**Testing done:**
When running some internal tests that also use mounts, Docker would fail to run the container and we got an exit of 254. Removing this patch brings the functionality back.

**Terms of contribution:**


bcressey (approved):

larvacea (approved):
Yes, this does precisely revert the change to packages/docker-engine, no more, no less.

PR #4026, opened 2024-06-04
Revert mounting static kmod as /sbin/modprobe 

yeazelm:
**Description of changes:**
We found this change to be unsafe for alpine containers in our testing. Reverting this change for the time being until we can find a way to do this safely for all containers. 

**Testing done:**
Attempt to launch an image based upon alpine, the process dies with exit 254 when trying to run alpine images.

**Terms of contribution:**


arnaldo2792 (approved):

larvacea:
I believe the lint failure does not need to be fixed immediately.

vigh-m (approved):

PR #4035, opened 2024-06-05
packages: add mdadm package for software RAID support

tzneal:
**Issue number:**

Closes #

**Description of changes:**

Adds the mdadm package to support future software raid configuration of local ephemeral storage.

**Testing done:**

- Built/ran an AMI and used an admin container to run mdadm and configure local storage.
- `cargo make check`

**Terms of contribution:**

### DIFF
--- a/packages/mdadm/0001-report-monitor-output-to-syslog.patch
+++ b/packages/mdadm/0001-report-monitor-output-to-syslog.patch
@@ -1,0 +1,39 @@
+From a277a7b7d171dbc7cee691ee20a9990fec524581 Mon Sep 17 00:00:00 2001
+From: Todd Neal <tnealt@amazon.com>
+Date: Fri, 7 Jun 2024 18:22:52 +0000
+Subject: [PATCH] report monitor output to syslog and remove env script
+
+---
+ systemd/mdmonitor-oneshot.service | 3 +--
+ systemd/mdmonitor.service         | 3 +--
+ 2 files changed, 2 insertions(+), 4 deletions(-)
+
+diff --git a/systemd/mdmonitor-oneshot.service b/systemd/mdmonitor-oneshot.service
+index ba86b44..c54f080 100644
+--- a/systemd/mdmonitor-oneshot.service
++++ b/systemd/mdmonitor-oneshot.service
+@@ -10,7 +10,6 @@ Description=Reminder for degraded MD arrays
+ Documentation=man:mdadm(8)
+ 
+ [Service]
+-Environment=MDADM_MONITOR_ARGS=--scan
++Environment=MDADM_MONITOR_ARGS="--scan --syslog"
+ EnvironmentFile=-/run/sysconfig/mdadm
+-ExecStartPre=-/usr/lib/mdadm/mdadm_env.sh
+ ExecStart=BINDIR/mdadm --monitor --oneshot $MDADM_MONITOR_ARGS
+diff --git a/systemd/mdmonitor.service b/systemd/mdmonitor.service
+index 9c36478..fcf895b 100644
+--- a/systemd/mdmonitor.service
++++ b/systemd/mdmonitor.service
+@@ -11,7 +11,6 @@ DefaultDependencies=no
+ Documentation=man:mdadm(8)
+ 
+ [Service]
+-Environment=  MDADM_MONITOR_ARGS=--scan
++Environment=  MDADM_MONITOR_ARGS="--scan --syslog"
+ EnvironmentFile=-/run/sysconfig/mdadm
+-ExecStartPre=-/usr/lib/mdadm/mdadm_env.sh
+ ExecStart=BINDIR/mdadm --monitor $MDADM_MONITOR_ARGS
+-- 
+2.40.1
+
--- a/packages/mdadm/mdadm-tmpfiles.conf
+++ b/packages/mdadm/mdadm-tmpfiles.conf
@@ -1,0 +1,2 @@
+d /var/run/mdadm 0700 root root -
+Z /var/run/mdadm 0700 root root -
--- a/packages/mdadm/mdadm.spec
+++ b/packages/mdadm/mdadm.spec
@@ -1,0 +1,68 @@
+Name: %{_cross_os}mdadm
+Version: 4.3
+Release: 1%{?dist}
+Summary: mdadm is used for controlling Linux md devices (aka RAID arrays)
+License: GPL-2.0-only
+URL: https://cdn.kernel.org/pub/linux/utils/raid/mdadm/
+Source0: https://cdn.kernel.org/pub/linux/utils/raid/mdadm/mdadm-%{version}.tar.xz
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: %{_cross_os}systemd-devel
+
+Source100: mdadm-tmpfiles.conf
+Patch100: 0001-report-monitor-output-to-syslog.patch
+
+%description
+%{summary}.
+
+%global set_env \
+%set_cross_build_flags \\\
+export CC=%{_cross_target}-gcc \\\
+CXFLAGS="%{_cross_cflags} -DNO_COROSYNC -DNO_DLM" \
+%{nil}
+
+%prep
+%autosetup -n mdadm-%{version} -p1
+
+%build
+%set_env
+make LDFLAGS="%{_cross_ldflags}"
+
+%install
+%set_env
+make install-bin DESTDIR=%{buildroot}%{_cross_rootdir}/usr
+make install-udev DESTDIR=%{buildroot}
+make install-systemd DESTDIR= SYSTEMD_DIR=%{buildroot}%{_cross_unitdir}
+
+install -d %{buildroot}%{_cross_tmpfilesdir}
+install -p -m 0644 %{S:100} %{buildroot}%{_cross_tmpfilesdir}/mdadm.conf
+
+%files
+%license COPYING
+%{_cross_attribution_file}
+
+%{_cross_sbindir}/mdadm
+%{_cross_sbindir}/mdmon
+
+%{_cross_tmpfilesdir}/mdadm.conf
+
+%{_cross_udevrulesdir}/01-md-raid-creating.rules
+%{_cross_udevrulesdir}/63-md-raid-arrays.rules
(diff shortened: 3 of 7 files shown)


packages/mdadm/mdadm.spec
```diff
@@ -0,0 +1,49 @@
...
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: %{_cross_os}systemd-devel
+
+# local patches
+Patch0001: 0001-rename-MAP_FILE-FMAP_FILE-to-avoid-issue-with-MAP_FI.patch
+
+%description
+%{summary}.
+
+%global debug_package %{nil}
```
yeazelm: nit: we usually put this at the top of the file to get it out of the way.

packages/mdadm/mdadm.spec
```diff
@@ -0,0 +1,49 @@
...
+%{_cross_attribution_file}
+%{_cross_sbindir}/mdadm
+%{_cross_sbindir}/mdmon
+%{_cross_udevrulesdir}/01-md-raid-creating.rules
+%{_cross_udevrulesdir}/63-md-raid-arrays.rules
+%{_cross_udevrulesdir}/64-md-raid-assembly.rules
+%{_cross_udevrulesdir}/69-md-clustered-confirm-device.rules
+
+%changelog
+
```
yeazelm, line 59: nit: we usually don't have empty newlines at the end of files.

packages/mdadm/0001-rename-MAP_FILE-FMAP_FILE-to-avoid-issue-with-MAP_FI.patch
```diff
@@ -0,0 +1,86 @@
+From 1f0adcdadac76b8ef1f52cc19ce28c8af7f70488 Mon Sep 17 00:00:00 2001
+From: Todd Neal <tnealt@amazon.com>
+Date: Thu, 9 May 2024 20:53:55 +0000
+Subject: [PATCH] rename MAP_FILE -> FMAP_FILE to avoid issue with MAP_FILE
```
yeazelm: I don't have context on why this was needed if you can provide a bit more background on the issue this works around that would be awesome!

tzneal: When I built it last month, I was having an issue with a conflict between a MAP_FILE dependency and one in the system headers.  I just removed the patch and its gone now, so 🤷‍♂️ .

packages/release/Cargo.lock
```diff
@@ -0,0 +1,591 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 3
```
bcressey: This file should be removed - it gets created if you accidentally run `cargo make` in a package directory.

packages/mdadm/Cargo.lock
```diff
@@ -0,0 +1,131 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
+version = 3
```
bcressey: remove this `Cargo.lock`

packages/mdadm/mdadm.spec
```diff
@@ -0,0 +1,44 @@
...
+BuildRequires: %{_cross_os}glibc-devel
+BuildRequires: %{_cross_os}systemd-devel
+
+%description
+%{summary}.
+
+%global set_env \
+%set_cross_build_flags \\\
+export CC=%{_cross_target}-gcc \\\
+CFLAGS="%{_cross_cflags} -DNO_COROSYNC -DNO_DLM" \
```
bcressey: It looks like this Makefile is set up to use `CXFLAGS` to extend the flags.

Otherwise you'd need to pass CFLAGS on the `make` command line to override the defaults. (This is most likely why debuginfo isn't being generated.)

packages/mdadm/mdadm.spec
```diff
@@ -0,0 +1,44 @@
+%global debug_package %{nil}
```
bcressey: You shouldn't need `%global debug_package %{nil}` at the top if everything's working correctly. We have that in some specs but that's only "fine" in cases where it truly isn't needed, for packages like `release` with no compiled code. Everywhere else it's just tech debt that we'll have to pay off when we start doing something useful with debuginfo.

packages/mdadm/mdadm.spec
```diff
@@ -0,0 +1,44 @@
...
+BuildRequires: %{_cross_os}systemd-devel
+
+%description
+%{summary}.
+
+%global set_env \
+%set_cross_build_flags \\\
+export CC=%{_cross_target}-gcc \\\
+CFLAGS="%{_cross_cflags} -DNO_COROSYNC -DNO_DLM" \
+LDFLAGS="%{_cross_ldflags}" \
```
bcressey: Looks like this needs to be set on the `make` command line instead of in the environment.

packages/mdadm/mdadm.spec
```diff
@@ -0,0 +1,44 @@
...
+make install-bin DESTDIR=%{buildroot}%{_cross_rootdir}/usr
+make install-udev DESTDIR=%{buildroot}
+
+%files
+%license COPYING
+%{_cross_attribution_file}
+%{_cross_sbindir}/mdadm
+%{_cross_sbindir}/mdmon
+%{_cross_udevrulesdir}/01-md-raid-creating.rules
+%{_cross_udevrulesdir}/63-md-raid-arrays.rules
```
bcressey, line 49: These rules reference a few different systemd units:
* `mdmonitor.service`
* `mdmon@.service`
* `mdadm-grow-continue@.service`

It seems like all of those just call `mdadm` or `mdmon` so they should be fine to include. The first two aren't used for RAID-1 and it's doubtful that we'll need to grow a RAID-0 array, but better to have them unless we patch the udev rules to not need them.

packages/mdadm/mdadm.spec
```diff
@@ -0,0 +1,44 @@
...
+%{_cross_udevrulesdir}/64-md-raid-assembly.rules
```
bcressey, line 50: references `mdadm-last-resort@.service` so let's include that

packages/mdadm/mdadm.spec
```diff
@@ -0,0 +1,44 @@
...
+%{_cross_udevrulesdir}/69-md-clustered-confirm-device.rules
```
bcressey, line 51: Just needs `blkid` and `mdadm`.

packages/mdadm/mdadm.spec
```diff
@@ -0,0 +1,44 @@
...
+
+%changelog
```
bcressey: nit: we discovered we can omit `%changelog` now, so it's fine to drop it

bcressey (commented):
Looks good overall.

As a general packaging note: udev rules and systemd units are places where unexpected dependencies can hide, so they're worth a close look.

packages/mdadm/Cargo.lock
tzneal: Without this, `cargo make -e PACKAGE=mdadm build-package` fails.  What am I missing?

packages/mdadm/mdadm.spec
tzneal: Ah, yeah, fixing the compile flags eliminates the need for this. Thanks!

packages/mdadm/Cargo.lock
larvacea: Good question, and I too have this problem with `cargo make build-package`. It's not just you, and it's not just this package. I'll let you know as soon as I have an answer. Or, possibly, @bcressey will answer us both before I figure it out, as he does.

packages/mdadm/mdadm.spec
tzneal: Added, required a few more changes for mdmonitor to actually work (specifying output goes to syslog and creating /var/run/mdadm).

packages/mdadm/Cargo.lock
larvacea: Good news: thanks to @yeazelm I know when we introduced this misbehavior in our build system. The somewhat good news is that there is a workaround: if you build a variant that includes this package (e.g., you add it to the packages in the `aws-dev` variant and build that variant), the build system will happily build your package without a `Cargo.lock` file, as it builds the variant you have selected. I will see what we can do about making `cargo make build-package` work as intended, and update this pull request.

packages/mdadm/0001-report-monitor-output-to-syslog.patch
```diff
@@ -0,0 +1,39 @@
...
+--- a/systemd/mdmonitor-oneshot.service
++++ b/systemd/mdmonitor-oneshot.service
+@@ -10,7 +10,7 @@ Description=Reminder for degraded MD arrays
+ Documentation=man:mdadm(8)
+ 
+ [Service]
+-Environment=MDADM_MONITOR_ARGS=--scan
++Environment=MDADM_MONITOR_ARGS="--scan --syslog"
+ EnvironmentFile=-/run/sysconfig/mdadm
+ ExecStartPre=-/usr/lib/mdadm/mdadm_env.sh
```
bcressey: Seems like we should also patch out the ExecStartPre shell script (which will always fail).

Does it do anything useful?

packages/mdadm/mdadm-tmpfiles.conf
```diff
@@ -0,0 +1,2 @@
+d /var/run/mdadm 0700 chrony chrony -
```
bcressey: This should be an mdadm user account (or whatever the standard is). You can create that via a sysusers.conf drop-in like in the chrony package.

packages/mdadm/0001-report-monitor-output-to-syslog.patch
tzneal: I thought someone might want to provide those settings, but I've got no use and we can always add it back if needed. Removed.

packages/mdadm/mdadm-tmpfiles.conf
tzneal: I checked and standard seems to be root/root, updated.

bcressey (approved):

yeazelm (approved):

Issue #260, opened 2019-09-19
API can't handle numeric settings sourced from settings generators


sam-aws:
From https://github.com/amazonlinux/PRIVATE-thar/pull/255 we found that Sundog can't interpret a setting from a setting generator's output if the type in the model is numeric. We probably need some serde magic here

PR #4038, opened 2024-06-06
migrate to settings plugins and eliminate variant-based conditional compilation

### LINKED ISSUE #3615: OOTB: Remove conditional compilation from Bottlerocket
**What I'd like:**
For [out of tree builds](https://github.com/bottlerocket-os/bottlerocket/issues/2669), Bottlerocket's binaries will be re-used for multiple variants, rather than built from source on a per-variant basis. This means we must remove conditional compilation from the distribution.

This issue is open to track the overall effort to remove conditional compilation.

Includes
- [x] #3616 
- [x] #3617
- [x] #3618 
- [x] #3783 
- [x] #3784 
- [x] #3621 
- [x] #3620 
- [x] #3619
- [x] #3883

bcressey:
**Issue number:**
Closes: #3615 

**Description of changes:**

Some major long-term efforts culminate here:
1. Move all the distinct settings models into settings plugins.
2. Make apiserver / datastore / models work with settings plugins.
3. Move all the distinct variant defaults into settings defaults crates.
4. Teach storewolf to load defaults at runtime.
5.  Add new packages and dependencies to pull in the correct plugin and defaults for each variant.

The end result of this is that `os` is no longer compiled differently based on the variant being built, and everything still works.

**Testing done:**
Built `aws-dev` and `aws-k8s-1.28-nvidia`. Verified that they used the expected defaults and supported the correct settings.

**Terms of contribution:**


sources/api/apiserver/src/server/controller.rs
```diff
@@ -366,8 +366,8 @@ pub(crate) fn set_settings<D: DataStore>(
     transaction: &str,
 ) -> Result<()> {
     trace!("Serializing Settings to write to data store");
-    let pairs =
-        to_pairs(settings).context(error::DataStoreSerializationSnafu { given: "Settings" })?;
+    let pairs = to_pairs_with_prefix("settings", settings)
+        .context(error::DataStoreSerializationSnafu { given: "Settings" })?;
```
cbgbt, line 370: Just making sure I understand -- this is required because the models were previously called `Settings` and the serializer would subsequently add that as part of each K/V name?

sources/api/datastore/src/serialization/pairs.rs
```diff
@@ -214,11 +218,6 @@ impl<'a> ser::Serializer for Serializer<'a> {
     }
 
     // Types we can't (or don't want to) represent.
-    // Can't fit u64 into signed 64-bit range.
-    fn serialize_u64(self, _v: u64) -> Result<()> {
-        bad_type("u64")
-    }
-
```
cbgbt, line 221: This is a bit concerning, but I probably don't understand the source of this constraint.

My assumption is that `u64` is disallowed due to a constraint in `serde_json` (indeed, Javascript cannot store integers larger than 53bits). Looking at [`serde_json` docs](https://docs.rs/serde_json/latest/serde_json/value/struct.Number.html#method.as_u64), it seems like maybe it does support `u64` though. Are we making some change to the datastore which makes this more flexible?

packages/settings-plugins/settings-plugins.spec
```diff
@@ -0,0 +1,205 @@
...
+Name: %{_cross_os}settings-plugins
+Version: 0.0
+Release: 0%{?dist}
+Summary: Settings plugins
+License: Apache-2.0 OR MIT
+URL: https://github.com/bottlerocket-os/bottlerocket
+BuildRequires: %{_cross_os}glibc-devel
+Requires: %{_cross_os}glibc-devel
+
+Requires: %{_cross_os}settings-plugin(any)
```
cbgbt, line 14: I assume we'll need hoist this or something like it into `os` to force kits to provide something?

I suppose alternatively (and maybe correctly) we provide a null plugin.

packages/settings-defaults/settings-defaults.spec
```diff
@@ -0,0 +1,360 @@
...
+Summary: Settings defaults for the aws-ecs-2-nvidia variant
+Provides: %{_cross_os}settings-defaults(any)
+Provides: %{_cross_os}settings-defaults(aws-ecs-2-nvidia)
+Conflicts: %{_cross_os}settings-defaults(any)
+
+%description aws-ecs-2-nvidia
+%{summary}.
+
+%package aws-k8s-1.24
+Summary: Settings defaults for the aws-k8s-1.24 variant
```
cbgbt: Worth mentioning `1.23` in the summary? (Same Q for the nvidia defaults for these versions)

cbgbt (commented):
Clever use of dynamic libraries to solve this problem. Nice!

webern (approved):
I'll keep looking through this but don't want to be a blocker 😅

This is exciting!

sources/api/datastore/src/serialization/pairs.rs
bcressey: > Are we making some change to the datastore which makes this more flexible?

On the datastore side, the "no u64" constraint seems to have come out of [this discussion](https://github.com/bottlerocket-os/bottlerocket/issues/260#issuecomment-543314276). Not sure if `serde_json` didn't support `u64` at the time, or if Tom was concerned about not tying persisted data to behavior outside the actual JSON spec.

It's still the case that we don't want to actually persist or support u64s in the datastore - if nothing else because the TOML serde implementation doesn't support them.

The problem being fixed here is that we need to bounce conversions through a serde_json Value, so that our custom serializer gets structured data to work with rather than the serde_json RawValue that comes out of the plugins. And when the serializer sees a `Value::Number(n)` it takes the `serialize_u64` code path.

I've confirmed that the API still rejects larger numbers, so we're not actually losing the "no u64" property:
```
# apiclient set updates.seed=5000000000
Failed to change settings: Failed PATCH request to '/settings/keypair?tx=apiclient-set-IzntMxvwbPebZ0GV': Status 400 when PATCHing /settings/keypair?tx=apiclient-set-IzntMxvwbPebZ0GV: Unable to match your input to the data model.  We may not have enough type information.  Please try the --json input form.  Cause: Error during deserialization: invalid value: integer `5000000000`, expected u32 at line 1 column 29

# apiclient apply <<EOF
> [settings.updates]
> seed = 5000000000
> EOF
Failed to apply settings: Failed to PATCH settings from '-' to '/settings?tx=apiclient-apply-WEVNJ04u1F1d51xU': Status 400 when PATCHing /settings?tx=apiclient-apply-WEVNJ04u1F1d51xU: Json deserialize error: invalid value: integer `5000000000`, expected u32 at line 1 column 29
```

packages/settings-plugins/settings-plugins.spec
bcressey: > I assume we'll need hoist this or something like it into os to force kits to provide something?

I added the dependency to `os.spec` in a later. However, your comment made me realize that in the absence of a "preferred" choice defined by requires in the top-level package, the approach here would allow `dnf` to pick a settings plugin at random to satisfy the `%{_cross_os}settings-plugin(any)` dependency. (And same for defaults.)

I'll tighten that up in the next revision.

sources/settings-plugins/dummy/Cargo.toml
```diff
@@ -0,0 +1,34 @@
+[package]
+name = "settings-plugin-dummy"
```
webern: This could use a readme or something to explain its existence.

sources/settings-defaults/metal-k8s-1.30/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    bottlerocket_defaults_helper::generate_defaults_toml().unwrap();
```
webern: It seems like this could be deduplicated by placing it at `../build.rs`.

sources/settings-defaults/metal-dev/src/lib.rs
```diff
@@ -0,0 +1 @@
+// This "crate" only generates a defaults.toml as a side effect of the build.
```
webern: It seems like this could be deduplicated by placing it at `../lib.rs`.

sources/models/src/variant/.keep
```diff
@@ -1,3 +1,2 @@
-We want git to track this directory so that it exists for end-user builds.
-It's needed by build.rs and the symlinks described in README.md.
-It's clearer what's going on if the directory exists already, rather than build.rs creating it.
+FIXME: this directory is no longer needed for builds, but can't be removed yet
+because Twoliter expects it to exist.
```
webern, line 2: Yeah, it's `build.Dockerfile` that needs it.

sources/models/build.rs
```diff
@@ -1,130 +1,3 @@
...
-use filetime::{set_symlink_file_times, FileTime};
-use rand::{distributions::Alphanumeric, thread_rng, Rng};
-use std::fs;
-use std::io;
-use std::os::unix::fs::symlink;
-use std::path::Path;
-use std::process;
-
-/// We create a link from 'current' to the variant selected by the environment variable above.
-const VARIANT_LINK: &str = "src/variant/current";
```
webern, line 17: :tada:

webern (approved):

bcressey:
I still need to rebase, but the above force push uses the tagged `v0.1.0` release of `bottlerocket-settings-sdk`, and addresses feedback so far:
* drop `authors` field from the new crate manifests (@sam-berning)
* use shared sources for the `shared-defaults` crates (@webern )
* rename `settings-plugin-dummy` to `simple-settings-plugin`, move under `api`, add README (@webern)
* revamp dependencies in the `settings-defaults` package to force `dnf` to pick an exact match (@cbgbt)
* drop `settings-extensions` packages since they aren't used and don't have a clear transition path to out-of-tree

cbgbt (approved):

webern (approved):

mgsharm (approved):

webern (approved):

sources/api/apiserver/src/server/controller.rs
bcressey: Yes - now the name of the struct is not visible to the serializer, so we have to provide it. I feel like there's probably a way to make the old behavior work, but I couldn't find it in the time I spent trying.

